PR #3780, opened 2015-01-11
FIX: Expands the CMS' centre-pane when collapsed and it's clicked.

phptek:


admin/javascript/LeftAndMain.Panel.js
```diff
@@ -134,5 +134,14 @@
 				return false;
 			}
 		});
+
+		$('.cms-content-tools').entwine({
```
kinglozzer: Any reason why you’re using the `if` statement instead of just binding to `.cms-content-tools.collapsed` here?

phptek: Haha good spot. I think I had originally placed the code in a different file and then copy/pasted it. I'll refactor and re-push.

kinglozzer:
Tested and working perfectly. Thanks @phptek

PR #3782, opened 2015-01-12
fix to grid field loading wrong current page id when using multiple tabs

pitchandtone:


pitchandtone:
Usecase:
Page has many products (or any has many relationship)

1) Open page 1
2) Open page 1's product
3) Open second tab with page 2
4) Go back to first tab
5) Hit save.

Currently you get redirected. Or if you use versioned girdfield the product you were editing is now owned by page 2 (wha?). See icecaster/silverstripe-versioned-gridfield#18

The reason is that it loads the currentPage from session rather than request and session is not distinct to tabs obviously.

This fix means that the page that it currently belongs to is sent in the POST and picked up.

dhensby:
Nice, I like this. It'll need to go against 3 branch, though.

There are no doubt places in the CMS where the session is relied upon? Can you take a look and add this to those forms so it's not relying on session state?

pitchandtone:
@dhensby sorry not following, this will still default back to session if this parameter isn't present.

dhensby:
I mean there are likely other forms that just jump straight to inspecting the session instead of using the controller to get the page ID - can you check this and stop that happening if it is the case?

This PR needs opening against `3` branch, please

admin/code/LeftAndMain.php
```diff
@@ -1442,6 +1442,8 @@ public function getSilverStripeNavigator() {
 	public function currentPageID() {
 		if($this->request->requestVar('ID') && is_numeric($this->request->requestVar('ID')))	{
 			return $this->request->requestVar('ID');
+		} elseif ($this->request->requestVar('CMSMainCurrentPageID') && is_numeric($this->request->requestVar('CMSMainCurrentPageID'))) {
```
tractorcow: I think this line needs a comment to help future maintainers know WHY it's there. Maybe a `// {@see GridFieldDetailForm} for use case.`

forms/gridfield/GridFieldDetailForm.php
```diff
@@ -403,6 +403,9 @@ public function ItemEditForm() {
 			}
 		}
 
+		if($this->controller->hasMethod('currentPageID') && $this->controller->currentPageID()) {
```
tractorcow: Comment what this is doing here too please. :)

PR #3008, opened 2014-04-04
BUG Fixed handling of numbers in certain locales.

tractorcow:
Fixes #2161

@lekoala, @jurriaan, @andrelohmann 

Do you like this solution?

Credit to @andrelohmann for beginning this solution.

PR #3786, opened 2015-01-13
Master

andrelohmann:
I have solved to issues here, that I came across a while ago


tractorcow:
Does this happen to cover https://github.com/silverstripe/silverstripe-framework/pull/3008 as well? We are looking to maybe fix this issue in 3.1

wilr:
Dup.

PR #3789, opened 2015-01-13
Implemented naturalPrev function in Hierarchy class

jedateach:
This has been marked as "todo" since forever. I just need it for a project.

`naturalNext` function doesn't have unit tests, so they both need unit tests at some stage.

Only tested with parent->children relationship so far, rather than deep nesting.

### DIFF
--- a/model/Hierarchy.php
+++ b/model/Hierarchy.php
@@ -701,12 +701,57 @@ class Hierarchy extends DataExtension {
 	}
 	
 	/**
-	 * Get the next node in the tree of the type. If there is no instance of the className descended from this node,
+	 * Get the previous node in the tree of the type. If there is no instance of the className descended from this node,
 	 * then search the parents.
-	 * 
-	 * @todo Write!
+	 *
+	 * @param string $className Class name of the node to find.
+	 * @param string|int $root ID/ClassName of the node to limit the search to
+	 * @param DataObject afterNode Used for recursive calls to this function
+	 * @return DataObject
 	 */
-	public function naturalPrev( $className, $afterNode = null ) {
+	public function naturalPrev( $className = null, $root = 0, $afterNode = null ) {
+		// If this node is not the node we are searching from, then we can possibly return this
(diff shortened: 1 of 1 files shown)


model/Hierarchy.php
```diff
@@ -701,12 +701,57 @@ public function getBreadcrumbs($separator = ' &raquo; ') {
...
+				}
+			}
+			
+			if($prevNode) {
+				return $prevNode;
+			}
+		}
+		
+		// if this is not an instance of the root class or has the root id, search the parent
+		if(!(is_numeric($root) && $root == $this->owner->ID || $root == $this->owner->class)
```
dhensby: You need some more brackets here to group the conditions.

dhensby:
I think this needs to go against 3 branch

model/Hierarchy.php
tractorcow: Is this enough?

```php
// if this is not an instance of the root class or has the root id, search the parent
if($root != $this->owner->ID && $root != $this->owner->class && ($parent = $this->owner->Parent())) {
	// ....
}
```

willmorgan: `&& $parent->exists()`

dhensby:
This also needs tests

dhensby:
This actually needs to go against master as it's a breaking API change... even though it was never implemented :(

model/Hierarchy.php
```diff
@@ -701,12 +701,57 @@ public function getBreadcrumbs($separator = ' &raquo; ') {
...
+				&& ($parent = $this->owner->Parent())) {
+			
+			return $parent->naturalPrev( $className, $root, $this->owner );
```
dhensby: spacing inside parenthesis shouldn't be there

dhensby:
@tractorcow is there any value in someone picking this up? I don't even know why there's a function stub for this

tractorcow:
We should just remove the stub in 4.x.

PR #3790, opened 2015-01-14
Extending site - RSS correction

ProzacJellybeans:
Instructions needed to be given on flushing more frequently, not just on template editing


dhensby:
Thanks, please can you explain why (because the allowed_actions have been changed)

PR #3793, opened 2015-01-14
API Adding default_classes config option for Form and FormField

### LINKED ISSUE #3688: Allow Form to be extended
Using HTML/CSS Frameworks like Bootstrap or Foundation you have to transform the Forms to be rendered properly. A current solution is to subclass Form and do changes. This is quite un-handy when you have already some Forms (e.g. MemberLoginForm) which you also have to subclass and use instead.

The Foundation solution is quite easy and just calles some stuff in `forTemplate()`. See https://github.com/ryanwachtl/silverstripe-foundation-forms/blob/master/code/FoundationForm.php

I think it might be quite handy if at least `Form::forTemplate()` would have an extension point where you can plug in some functionality for all forms globally (you'll have to check if you're in frontend manually).

dhensby:
This adds the ability to add default classes to `Form` and `FormField` which is useful for CSS frameworks to ensure fields and forms can have classes which are relevant.

Fixes #3688

### DIFF
--- a/forms/Form.php
+++ b/forms/Form.php
@@ -142,6 +142,12 @@ class Form extends RequestHandler {
 	protected $extraClasses = array();
 
 	/**
+	 * @config
+	 * @var array $default_classes The default classes to apply to the Form
+	 */
+	private static $default_classes = array();
+
+	/**
 	 * @var string
 	 */
 	protected $encType;
@@ -224,6 +230,8 @@ class Form extends RequestHandler {
 		}
 
 		$this->securityToken = ($securityEnabled) ? new SecurityToken() : new NullSecurityToken();
+
+		$this->setupDefaultClasses();
 	}
 
 	private static $url_handlers = array(
@@ -261,6 +269,19 @@ class Form extends RequestHandler {
 		}
 
 		return $this;
+	}
+
+	/**
+	 * set up the default classes for the form. This is done on construct so that the default classes can be removed
+	 * after instantiation
+	 */
+	protected function setupDefaultClasses() {
+		$defaultClasses = self::config()->get('default_classes');
+		if ($defaultClasses) {
+			foreach ($defaultClasses as $class) {
+				$this->addExtraClass($class);
+			}
+		}
 	}
 
 	/**
--- a/tests/forms/FormFieldTest.php
+++ b/tests/forms/FormFieldTest.php
@@ -3,7 +3,52 @@
  * @package framework
  * @subpackage tests
  */
-class FormFieldTest extends SapphireTest implements TestOnly {
+class FormFieldTest extends SapphireTest {
+
+	public function testDefaultClasses() {
+		Config::nest();
+
+		Config::inst()->update('FormField', 'default_classes', array(
+			'class1',
+		));
+
+		$field = new FormField('MyField');
+
+		$this->assertContains('class1', $field->extraClass(), 'Class list does not contain expected class');
+
+		Config::inst()->update('FormField', 'default_classes', array(
+			'class1',
+			'class2',
+		));
+
+		$field = new FormField('MyField');
+
+		$this->assertContains('class1 class2', $field->extraClass(), 'Class list does not contain expected class');
+
+		Config::inst()->update('FormField', 'default_classes', array(
+			'class3',
+		));
+
+		$field = new FormField('MyField');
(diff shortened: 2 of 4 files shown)


tests/forms/FormFieldTest.php
```diff
@@ -3,7 +3,52 @@
...
+		Config::nest();
+
+		Config::inst()->update('FormField', 'default_classes', array(
+			'class1',
+		));
+
+		$field = new FormField('MyField');
+
+		$this->assertContains('class1', $field->extraClass(), 'Default class not added to class list');
+
```
stevie-mayhew: Shouldn't this be "default classes added to class list"?

dhensby: The message in the final parameter is shown if the assertion fails.

So, in this example, we would state the message if the `extraClass` method didn't contain "class1" and therefore is a failure because the default class wasn't added.

stevie-mayhew: I knew I was going mental. Thanks @dhensby

tractorcow: Aren't they normally written in the positive sense? https://github.com/silverstripe/silverstripe-framework/blob/3.1/tests/model/DataObjectTest.php#L1205

forms/Form.php
```diff
@@ -142,6 +142,11 @@ class Form extends RequestHandler {
 	protected $extraClasses = array();
 
 	/**
+	 * @var array $default_classes The default classes to apply to the Form
```
tractorcow: Can you pop a `@config` notifier here if it's configurable?

We haven't, but probably should add this to the coding conventions.

forms/Form.php
```diff
@@ -264,6 +271,19 @@ public function setupFormErrors() {
...
 
 	/**
+	 * set up the default classes for the form. This is done on construct so that the default classes can be removed
+	 * after instantiation
+	 */
+	protected function setupDefaultClasses() {
+		$defaultClasses = self::config()->get('default_classes');
+		if ($defaultClasses) {
+			foreach ($defaultClasses as $class) {
+				$this->addExtraClass($class);
```
chillu: Indentation ;)

chillu:
Awesome feature, this will make theming a lot easier

tests/forms/FormFieldTest.php
dhensby: Erm, It's going to be hard to write this failure positively. I could write "Class list does not contain expected default class"/

forms/Form.php
dhensby: What's up with the indentation? :(

tests/forms/FormFieldTest.php
kinglozzer: You’re writing what you expect to happen, so it’d just be: “Default class is added to class list”

dhensby:
Ok, I've addressed the comments by:

- Rewording the test assertion messages
- Adding `@config` to the static vars

@chillu if you can tell me what's up with the indentation, then I'll sort that out. Otherwise, I think we're ready.

tests/forms/FormFieldTest.php
dhensby: The final parameter is the message that is shown when the assertion fails. In this example that would be when the expected class "class1" is no in the string returned by the function.

If the assertion failed, you're suggesting the message should be "Default class is added to class list", which is not why the test failed.

If no final argument is provided, then it would use the default which would be: "Failed to assert that string "[insert returned string]" contained "class1".

kinglozzer: @dhensby I agree with you: the negative version makes more sense than the positive when displayed in the error output, but every other core test is written in the positive sense (I was just pointing out “our” convention for those messages).

We’re [not alone](https://github.com/symfony/DependencyInjection/blob/master/Tests/ContainerTest.php#L99) in messages being written this way, but the wording on [PHPUnit assertion docs](https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertArrayHasKey) always starts:

> Reports an error identified by $message... 

Our messages don’t really identify the error, so perhaps they should all be rewritten the other way around?

dhensby: Maybe we just shouldn't customise the messages ;)

PR #3795, opened 2015-01-15
FIX. Summary fields can't be translated

uniun:
fieldLabels() now can find these fields and translate them.

### DIFF
--- a/security/Member.php
+++ b/security/Member.php
@@ -99,9 +99,9 @@ class Member extends DataObject implements TemplateGlobalProvider {
 	);
 	
 	private static $summary_fields = array(
-		'FirstName' => 'First Name',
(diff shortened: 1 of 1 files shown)


tractorcow:
Yes, but now the summary fields are incorrect. Will the correct gridfield titles be shown in the member listing? Couldn't you fix the error in fieldLabels, if this is the part that's unable to perform the translation? It would be really awesome if you could please demonstrate the error, and that the fix works, with a unit test.

tractorcow:
Ok had a chat with @zauberfish and we decided to merge this in. :)

tractorcow:
An issue still exists at https://github.com/silverstripe/silverstripe-framework/issues/3903

PR #3415, opened 2014-08-21
FIX: belongs_many_many back to the same class type

theorytank:
Currently SS get's confused when referencing multiple of the same DataObject type in a belongs_many_many definition using the dot notation.  It doesn't recognize the dot notation at all, and leaving it off means it will just choose the first match in the many_many definition which isn't correct.  This adds in the dot check and reverse look-up functionality allowing for multiple definitions.

Updates from PULL Request #3409
* formatting fixed (spaces to tabs)
* restored summaryFields function to original format (couldn't figure out why I changed that, might have been going down a rabbit hole at the time)

PR #3799, opened 2015-01-18
NEW: Support multiple many_manys between the same classes (closes #1377)

### LINKED ISSUE #1377: [2009-09-04] Support multiple many_many relationships between the same classes
created by: @MarcusDalgren (smurkas)
created at: 2009-09-04
original ticket: http://open.silverstripe.org/ticket/4546

---

Currently the code for figuring out the table name from the belongs_many_many side (line 26 to 31) returns an incorrect table name.

Since the name of the joining table is defined by the many_many class the belongs_many_many class will have to look up it's related class and then check the relationship again from that end to determine the table name.

This is of course doable but a simpler solution would be to change the naming convention for joining tables. Right now a joining table gets the name of the many_many class + _ + whatever the developer decides to call the connection to the belongs_many_many class. The simple solution here would be to simply name the joining table to many_many class + _ + belongs_many_many class.

That way figuring out the name of the joining table would be less of a hassle. If this breaks to much/isn't feasible then I have a patch for the first solution I proposed.

kinglozzer:


dhensby:
@kinglozzer this is a very nice addition indeed. We still don't have passing builds, though.

Have you run the test suite locally?

kinglozzer:
Yep, tests are passing locally.

I have to thank @theorytank for starting this work in #3415. In fact, I think I’ll break this up into two commits to give proper credit in the commit history. @theorytank, would you be okay with that?

theorytank:
Thanks so much for finishing that up.  I've been moving that code from version to version locally for about a year now without any issues.  I just couldn't figure out how to write the unit tests.

Thanks everyone.  I promise to figure out this process soon.

kinglozzer:
Okay, I’ve broken this apart into two commits: one is @theorytank’s original commit from #3415 and one is a slight refactor and tests/docs. Shame we still don’t have builds working, but I’ve re-run all framework tests locally and they’re passing :)

model/DataObject.php
```diff
@@ -1926,13 +1926,21 @@ public function many_many_extraFields($component = null) {
...
+					}
+
 					$SNG_candidate = singleton($candidate);
 					$candidateManyMany = $SNG_candidate->stat('many_many');
 
 					// Find the relation given the class
-					if($candidateManyMany) foreach($candidateManyMany as $relation => $relatedClass) {
-						if($relatedClass == $class) {
-							$relationName = $relation;
+					if( ! $relationName && $candidateManyMany) {
```
dhensby: spacing around `!`

model/DataObject.php
```diff
@@ -1926,13 +1926,21 @@ public function many_many_extraFields($component = null) {
...
+						foreach($candidateManyMany as $relation => $relatedClass) {
+							if($relatedClass == $class) {
```
dhensby: On line 2022 you're using: `$candidateClass == $class || is_subclass_of($class, $candidateClass)`, don't you need to do the same here?

tests/model/DataObjectTest.php
```diff
@@ -1660,6 +1703,19 @@ class DataObjectTest_Company extends DataObject {
 	);
 }
 
+class DataObjectTest_EquipmentCompany extends DataObjectTest_Company {
```
dhensby: `implements TestOnly`

dhensby:
@kinglozzer can you add a test where a child class of the relations are used to ensure it works with subclasses as well as the explicitly related classes?

model/DataObject.php
kinglozzer: Well, both occurrences are existing code that I’ve just wrapped in `if`/`else` statements.

However, I’m trying to wrap my head around exactly why the `is_subclass_of()` is present _at all_... As far as I can make out, this is the only data structure that it would apply to:

```php
class Team extends DataObject {
    private static $many_many = array('Players' => 'SubPlayer');
}

class Player extends DataObject {
    // Implicit belongs_many_many?
}

class SubPlayer extends Player {
    private static $belongs_many_many = array('Teams' => 'Team');
}

singleton('Player')->many_many('Teams'); // null
```

But I’m sure this would never get past line 2003... I’m so confused!

kinglozzer:
@dhensby I’ve added a couple of assertions for subclasses (both from `many_many` and `belongs_many_many` sides), let me know if you think I’ve missed something :)

model/DataObject.php
```diff
@@ -1926,13 +1926,21 @@ public function many_many_extraFields($component = null) {
 				$manyMany = $SNG_class->stat('belongs_many_many');
 				$candidate = (isset($manyMany[$component])) ? $manyMany[$component] : null;
 				if($candidate) {
+					$relationName = null;
+					// Extract class and relation name from dot-notation
+					if(strpos($candidate, '.') !== false) {
+						list($candidate, $relationName) = explode('.', $candidate);
```
dhensby: You need to explicitly set the 3rd argument of explode to be `2` as you're using list.

```php
explode('.', $candidate, 2);
```

dhensby:
The test fixture .yml file needs to be spaces not tabs

dhensby:
and please can we stop using positive messages for the assertion failure message :/

Better not to have a message than the wrong one.

kinglozzer:
Done, done and done!

dhensby:
We're seeing lots of errors!

dhensby:
You need to make sure the object classes are in the `extraDataObjects` for any tests using the fixture you changed

kinglozzer:
Ohhh, the fixtures are shared with other tests - missed that, yay for Travis!

dhensby:
I'm seeing GREEN LIGHTS!

dhensby:
I'm so upset this isn't in 3.1! :(

dhensby:

tractorcow:
New feature means it must go into 3 branch sorry :) It's in the right place.

tractorcow:
> and please can we stop using positive messages for the assertion failure message :/

I think it's worth documenting how exactly we should be writing our test cases, since the practice up until now is about 30% negative 70% positive. I think negative makes more sense though, but we should at least be consistent.

tractorcow:
Do you think we could break down some of these god-methods? Could you look at some of the larger methods and refactor out key functions. This would make it a lot easier to peer review...

It's a good chance to do cleanup, phpdoc things, remove deprecated api, and simplify some of the code. The problem with PHP is that array ends up being the default data structure, but it can be confusing to reverse engineer meaning from array manipulation code.

dhensby:
> I think it's worth documenting how exactly we should be writing our test cases, since the practice up until now is about 30% negative 70% positive. I think negative makes more sense though, but we should at least be consistent.

New code should be right (IMO) and consistency should be the next aim.

The messages are to supplement the default failure message. A lot of the time I'd say a message isn't required:

```php
$this->assertInstanceOf('DataObject', new stdClass);
```

This would give "failed to assert object of type stdClass is DataObject". From the context of a test it should be pretty obvious what is going on and why it's needed. We could add something like "DataObject::get() failed to return a DataObject", which gives context without looking at the code (but generally we'd have to look at the test code anyway to investigate the circumstances of the failure and so on, so is nice but non-essential).

```php
$this->assertTrue(is_object('string'));
```

This on the other hand would need a message as the default error would be "failed to assert that false is true". Not uesful. So adding "function setHeader failed to return an object" is pretty much essential.

Note that the importance here is to use the correct assertions where possible. If there isn't an appropriate assertion then a message is going to be required and it should explain what was wrong with the assertion not what it should have been to pass.

kinglozzer:
@tractorcow I’ve broken up the methods in f9f98660c1c63b668d38dc7400f2ef458ddd1b6e. I didn’t intend to rename them, but I didn’t want to add more methods that followed the “wrong” convention - it would’ve been weird to have matching methods named differently like `many_many` and `manyManyComponent`. I need to replace core uses of these methods too, figured I’d push it now for feedback though :)

model/DataObject.php
```diff
@@ -1744,44 +1745,79 @@ public function getManyManyComponents($componentName, $filter = "", $sort = "",
...
 	 * @param string $component Name of component
-	 *
 	 * @return string|array The class of the one-to-one component, or an array of all one-to-one components and their
 	 *                      classes.
 	 */
-	public function has_one($component = null) {
-		$classes = ClassInfo::ancestry($this);
+	public function hasOne($component = null) {
+		if($component) {
+			return $this->hasOneComponent($component);
```
tractorcow: This is very good.

model/DataObject.php
```diff
@@ -1717,7 +1717,8 @@ public function getRemoteJoinField($component, $type = 'has_many', &$polymorphic
 	 * @todo Implement query-params
 	 */
 	public function getManyManyComponents($componentName, $filter = "", $sort = "", $join = "", $limit = "") {
```
dhensby: It looks like $join is ignored here.

I'd favour deprecating ALL the optional params and telling people to manipulate the list directly

model/DataObject.php
```diff
@@ -1744,44 +1745,79 @@ public function getManyManyComponents($componentName, $filter = "", $sort = "",
 	}
 
 	/**
+	 * @deprecated 4.0 Method has been renamed to hasOne()
+	 * @param string $component
+	 * @return array|null
+	 */
+	public function has_one($component = null) {
+		Deprecation::notice('4.0', 'Method has been renamed to hasOne()');
```
dhensby: Can our deprecation notices be a bit more verbose? eg: "belongs_to has been deprecated, please use belongsTo instead"

model/DataObject.php
```diff
@@ -1744,44 +1745,79 @@ public function getManyManyComponents($componentName, $filter = "", $sort = "",
...
-			if($component) {
-				$hasOne = Config::inst()->get($class, 'has_one', Config::UNINHERITED);
-
-				if(isset($hasOne[$component])) {
-					return $hasOne[$component];
+			$newItems = (array)Config::inst()->get($class, 'has_one', Config::UNINHERITED);
+			// Validate the data
+			foreach($newItems as $k => $v) {
+				if(!is_string($k) || is_numeric($k) || !is_string($v)) {
+					user_error("$class::\$has_one has a bad entry: "
```
dhensby: Throw some kind of exception instead of user_error?

model/DataObject.php
```diff
@@ -1744,44 +1745,79 @@ public function getManyManyComponents($componentName, $filter = "", $sort = "",
...
+		}
+	}
+
+	/**
+	 * @deprecated 4.0 Method has been renamed to belongsTo()
+	 * @param string $component
+	 * @param bool $classOnly
+	 * @return array|null
+	 */
+	public function belongs_to($component = null, $classOnly = true) {
```
dhensby: I think this is a strange API where `belongsTo` is the only one with a "classonly" parameter.

What is it's special case for needing it?

--

Edit: Ah, I see it's for the relations that have a capability of being ambiguous, something you added as part of this PR? I don't think it should really be a param myself...

model/DataObject.php
```diff
@@ -1790,17 +1826,12 @@ public function has_one($component = null) {
...
 		if($component) {
-			if($belongsTo && array_key_exists($component, $belongsTo)) {
-				$belongsTo = $belongsTo[$component];
-			} else {
-				return false;
-			}
+			return $this->belongsToComponent($component, $classOnly);
 		}
 
+		$belongsTo = $this->stat('belongs_to');
```
dhensby: Are we meant to be using `stat`? I know it uses `Config` but isn't that just for legacy? Other places in this refactor are using uninherited config settings, which seems more appropriate?

model/DataObject.php
```diff
@@ -1876,164 +1932,256 @@ public function has_many($component = null, $classOnly = true) {
...
 		foreach($classes as $class) {
 			if(in_array($class, array('ViewableData', 'Object', 'DataObject'))) continue;
-			$relationName = null;
 
-			// Find extra fields for one component
-			if($component) {
-				$SNG_class = singleton($class);
-				$extraFields = $SNG_class->stat('many_many_extraFields');
+			// Find all the extra fields for all components
+			$newItems = eval("return (array){$class}::\$many_many_extraFields;");
```
dhensby: I thought we'd got rid of this?

model/DataObject.php
```diff
@@ -1876,164 +1932,256 @@ public function has_many($component = null, $classOnly = true) {
...
+	 * @param string $component
+	 * @return array|null
+	 */
+	public function manyManyExtraFieldsForComponent($component) {
+		$classes = ClassInfo::ancestry($this);
+		foreach($classes as $class) {
+			if(in_array($class, array('ViewableData', 'Object', 'DataObject'))) continue;
+			$relationName = null;
+			$SNG_class = singleton($class);
+			$extraFields = $SNG_class->stat('many_many_extraFields');
```
dhensby: with config we don't need to instantiate a singleton to access these settings any longer

dhensby:
Overall this is looking like a really sweet set of changes.

My main concerns are:

1. Use of `->stat` to access config vars (especially instantiating singletons to do it)
2. Deprecation notices not being explicit enough
3. There seems to be a lot of repeated code around the looping over class inheritance to work out relations. Can this not be pulled out to a single function with parameters about the config vars to look at?

model/DataObject.php
kinglozzer: I agree with that proposal, perhaps one for a separate PR - we’ve already expanded this one a fair bit! :)

model/DataObject.php
kinglozzer: Sure, will update the messages

model/DataObject.php
kinglozzer: Just keeping the existing behaviour. I think `user_error` makes sense anyway - would you ever actually want to catch this and try to carry on?

model/DataObject.php
kinglozzer: Yep, `has_many` has the same behaviour - introduced before this PR (in fact, blame view says it’s from 2009!). I can’t see anywhere that `$classOnly` is ever set to false in core code, so perhaps the second argument could be deprecated

model/DataObject.php
kinglozzer: I think you’re right - I was just blindly following the existing code, but it looks like accessing `Config` directly is probably better. Will replace these

model/DataObject.php
kinglozzer: In 3.1, I wasn’t sure whether to include that fix in this PR or let it get merged up. Would including it here be more likely to cause a merge error than not? Whichever is easiest!

model/DataObject.php
kinglozzer: Good spot!

kinglozzer: What do you think of the name of this method, `manyManyExtraFieldsForComponent`?  I was torn between that and `manyManyComponentExtraFields`

kinglozzer:
Regarding deprecation: these changes will trigger notices in 4.0 (or earlier if devs manually set the notice version to 4.0), with a view to finally removing the deprecated methods in 5.0. Is this acceptable? It sounds so far in the future, but I think it’s probably the best (least annoying) approach

> repeated code around the looping over class inheritance to work out relations

I’m assuming you mean the `ClassInfo::ancestry()` stuff? It might even be worth caching the result if those lookups are expensive

dhensby:
> I’m assuming you mean the ClassInfo::ancestry() stuff? It might even be worth caching the result if those lookups are expensive

Yes and the loops involved in that

model/DataObject.php
dhensby: You're right... but given you've done so much it seems like something pretty simple to fix.

Just change the params to `null` and if any of them `!== null` then throw a deprecation error.

model/DataObject.php
dhensby: thankies

model/DataObject.php
dhensby: by throwing errors then in the future it'll make handling errors further up the tree much easier.

model/DataObject.php
dhensby: I think so... needing the full string would be a very edge case and I suppose could be left to another function?

model/DataObject.php
dhensby: I think given all this work the merging will be a nightmare and the other fix is likely to NOT be merged in if you don't do it now.

model/DataObject.php
dhensby: Leave it as it is, I think :)

kinglozzer:
Do you think it might be worthwhile deprecating passing the `$component` parameter, now that we’ve got separate methods for these? I.e.:

```php
if($component) {
	Deprecation::notice('4.0', 'Please use hasOneComponent() instead of passing a component name to hasOne()');
	return $this->hasOneComponent($component);
}
```

model/DataObject.php
kinglozzer: Have updated these now

model/DataObject.php
kinglozzer: Makes sense. Have changed these to `LogicException`s, thought that was the best fit :)

model/DataObject.php
kinglozzer: I’ve implemented the fix & tests from 763173d6c60d462e1647152bf1b7e5c1a5ed2510

kinglozzer:
@dhensby I’ve replaced the singleton instantiation with `Config` getters where I can. Just need to look at refactoring those pesky loops now!

model/DataObject.php
kinglozzer: ... and I’ve shortened them again. When thrown with `Deprecation::SCOPE_METHOD` (default 2nd arg), it already includes the name of the original method. So the output of the above is:

`DataObject:: has_one is deprecated. Method has been renamed to hasOne()`

kinglozzer:
I’ve managed to ditch the loops over the class ancestry: it now just uses “inherited” config values instead (except for `manyManyComponent()` as this needs to know the specific class that the component is defined in for the join table).

dhensby:
> Do you think it might be worthwhile deprecating passing the $component parameter, now that we’ve got separate methods for these? I.e.:

Yes

> I’ve managed to ditch the loops over the class ancestry

awesome. We'll probably see some efficiency gains then!

Looks like this now needs rebasing  - I'll have another look over the diff and add anything else I see - but I think we're really close now

model/DataObject.php
```diff
@@ -1575,6 +1575,11 @@ public function getComponents($componentName, $filter = "", $sort = "", $join =
 			);
 		}
 
+		if($filter !== null || $sort !== null || $limit !== null) {
```
dhensby: You need to check join too, even though it had previously been discarded we should really not discard it now.

model/DataObject.php
```diff
@@ -1876,168 +1931,241 @@ public function has_many($component = null, $classOnly = true) {
...
+		}
 
-					// Find the relation given the class
-					if($candidateManyMany) foreach($candidateManyMany as $relation => $relatedClass) {
-						if($relatedClass == $class) {
-							$relationName = $relation;
-							break;
-						}
-					}
+		return count($items) ? $items : null;
```
dhensby: `!empty()` is better than `count()`.

personally, I think we should always return a consistent data type. `array() == false` just as `null == false` so shouldn't be a problem to do that

model/DataObject.php
```diff
@@ -1876,168 +1931,241 @@ public function has_many($component = null, $classOnly = true) {
...
-						if(isset($extraFields[$relationName])) {
-							return $extraFields[$relationName];
-						}
-					}
+	/**
+	 * Return the many-to-many extra fields specification for a specific component.
+	 * @param string $component
+	 * @return array|null
+	 */
+	public function manyManyExtraFieldsForComponent($component) {
```
dhensby: I think this method is taking on too much responsibility. Each bit where you're looping can probably be it's own function.

Also, isn't the logic for extracting the reverse relation via dot notation a share-able set of logic that could be a function too?

The fact you're returning in many different places under different conditions shows that it's a good candidate for being split into component parts

dhensby:
Right, my only other comment is regarding the loops over the config values and throwing errors.

Before, when we were having to loop anyway, it was acceptable to check the config values were well formed, but now we're introducing a loop essentially just to validate the config settings. This doesn't need to happen every time someone accesses `hasOne()` or `hasMany()` it only needs to happen on DB build or possibly manifest build.

I guess that's very much out of scope of this PR if it means amending the builder or the config parser, but probably something to be addressed soon after this.

model/DataObject.php
kinglozzer: `$join` already throws an `InvalidArgumentException` just above here, so I figured I’d leave that untouched!

model/DataObject.php
kinglozzer: I was reluctant to change any return types in a minor release, though - could cause issues for anyone doing strict comparisons. Will change to `!empty()`

kinglozzer:
> Right, my only other comment is regarding the loops over the config values and throwing errors.

You’re right about that, we could perhaps move it to `DataObject::requireTable()`?

model/DataObject.php
dhensby: Hm, is `$classOnly = false` used by functions to resolve ambiguous relationships where dot notation is used to find the reverse relation?

model/DataObject.php
dhensby: Ah, didn't spot that!

kinglozzer:
I’ve added a commit which moves where the data validation happens: https://github.com/kinglozzer/sapphire/commit/dc9e094483013b7529aef0fa0a3f6930a3c8d53d. I know it probably should be out of the scope of this PR, but the reason I did this is that my changes (changing the loops to inherited config) mean that it may show the wrong class name when throwing those exceptions. Moving it now means that it will show the correct class names :)

dhensby:
Great. Looks good :)

model/DataObject.php
```diff
@@ -1881,168 +1940,203 @@ public function has_many($component = null, $classOnly = true) {
...
-						}
-					}
-
-					if($relationName) {
-						$extraFields = $SNG_candidate->stat('many_many_extraFields');
-						if(isset($extraFields[$relationName])) {
-							return $extraFields[$relationName];
-						}
-					}
-				}
```
kinglozzer: For reference: I’ve removed this entire block. This section scans through the `belongs_many_many` side of the relation and looks for extra fields on that side. The reason I’ve removed it is that it’s not supported - `DataObject::requireTable()` only looks for extra fields on the `many_many` side, not `belongs_many_many`, so any extra fields added on the `belongs_many_many` side are never added to the database.

dhensby: Can you confirm this isn't a documented feature so we can remove it as redundant rather than this being strictly a bug?

kinglozzer: The best I can do is say:

1. There’s no documentation that suggests you can define extra fields on the `belongs_many_many` side (mainly because there’s almost no documentation on extra fields at all!)
2. The [code responsible](https://github.com/silverstripe/silverstripe-framework/blame/3a7e24a2208f58faa7ae8f1b5e575d39a65c918a/model/DataObject.php#L3096) for fetching extra fields to build the join table hasn’t changed since 2007 - it only checks for `many_many_extraFields` if it encounters `many_many` - it doesn’t even look at `belongs_many_many`.

dhensby: then lets get rid of it :)

model/DataObject.php
```diff
@@ -1744,63 +1756,80 @@ public function getManyManyComponents($componentName, $filter = "", $sort = "",
 	}
 
 	/**
+	 * @deprecated 4.0 Method has been renamed to hasOne()
+	 * @param string $component
+	 * @return array|null
+	 */
+	public function has_one($component = null) {
+		Deprecation::notice('4.0', 'Please use hasOne() instead');
```
dhensby: to save getting multiple deprecation notices, perhaps ths should check `if $compontent !== null` and throwing a "use hasOneComponent($component)" notice instead?

At the moment a dev would see two contradictory notices otherwise...

model/DataObject.php
```diff
@@ -1744,63 +1756,80 @@ public function getManyManyComponents($componentName, $filter = "", $sort = "",
...
-	 * @return string|array The class of the one-to-one component, or an array of all one-to-one components and their
-	 *                      classes.
+	 * @param string $component Deprecated - Name of component
+	 * @return string|array The class of the one-to-one component, or an array of all one-to-one components and
+	 * 							their classes.
 	 */
-	public function has_one($component = null) {
-		$classes = ClassInfo::ancestry($this);
+	public function hasOne($component = null) {
+		if($component) {
```
dhensby: strict type check? I guess is doesn't matter...

model/DataObject.php
kinglozzer: Good point, fixed

model/DataObject.php
```diff
@@ -1744,63 +1756,90 @@ public function getManyManyComponents($componentName, $filter = "", $sort = "",
...
 
 	/**
+	 * @deprecated 4.0 Method has been replaced by hasOne() and hasOneComponent()
+	 * @param string $component
+	 * @return array|null
+	 */
+	public function has_one($component = null) {
+		if($component) {
+			Deprecation::notice('4.0', 'Please use hasOneComponent() instead');
+			return $this->hasOne($component);
```
dhensby: if you're telling them to use hasOneComponent, then you need to here. there should be an else wrapping the rest

kinglozzer: Oops, I did that on the other methods, just missed this one! I can wrap the rest of the function in an else if you like, personal preference I guess :)

dhensby: oh, derp, yes, don't need to wrap if you're returning.

tractorcow: Returning is nearly always preferable to a massive pyramid of nested conditions. Makes for much simpler to read code.

dhensby:
Right, where did we get to with this? unfortunately it looks like it now needs rebasing, but we must be really close?

kinglozzer:
Yep, I need to rebase and also move the `DataObject::$db` “validation” to the new method I created (missed that first time around) and then I think we’re all set!

model/DataObject.php
```diff
@@ -1575,6 +1575,11 @@ public function getComponents($componentName, $filter = "", $sort = "", $join =
 			);
 		}
 
+		if($filter !== null || $sort !== null || $limit !== null) {
+			Deprecation::notice('4.0', 'The $filter, $sort and $limit parameters for DataObject::getComponents()
```
dhensby: This needs to be `'3.2'` not `4.0` we'd like to remove it in 4 so the warning should show sooner than that

dhensby:
> This needs to be '3.2' not 4.0 we'd like to remove it in 4 so the warning should show sooner than that

Ok, I think that's my final comment before I'll pull it.

kinglozzer:
Deprecation notices changed to 3.2, now we just need those green ticks from Mr Travis...

waiting

dhensby:
Is that a man in a suit pretending to be the cookie monster?!

dhensby:
Not looking good :/

Looks like some tests need updating?

kinglozzer:
Fixed, I’d missed a few places when updating core code to use the new methods!

([CMS updates ready to be made into a PR](https://github.com/kinglozzer/silverstripe-cms/commit/23383d4f882b4bff675d234e47bc53796e8e1ff4))

dhensby:
Great stuff all round!

tractorcow:
Darn I'm too late

> This needs to be '3.2' not 4.0 we'd like to remove it in 4 so the warning should show sooner than that

If it's introduced into 3.2 that's fine, but it can't break anything or start throwing new errors in 4.0. Then in 5.0 (or some time in 4.x) we can physically remove the deprecated method. I think upgrading a minor version should not introduce new errors, even if they are deprecation notices. A minor level upgrade shouldn't require additional code changes to remove notices.

I had this discussion a while back, and that's the conclusion we came up with.

kinglozzer:
Woohoo! Thanks for the peer reviews guys :)

PR #3623, opened 2014-11-10
FIX: DataObject::db() doesn't respect overloaded db types (fixes #3620)

lozcalver:

Issue #3620, opened 2014-11-10
Overridden DB fields inherit incorrect casting


lozcalver:
I haven’t had time to fully investigate this, but what I’m doing is overriding 'Content' in a subclass of Page:

``` php
class MyPage extends Page {
    private static $db = array(
        'Content' => 'Text'
    );
}
```

When rendered in the template, the value is treated as HTMLText (so no nl2br for me :cry:). Tried `private static $casting` to no avail, looks like it may be being cached [here](https://github.com/silverstripe/silverstripe-framework/blob/3.1/model/DataObject.php#L1680) by a previous call to `SiteTree->$field`.

PR #3803, opened 2015-01-19
Re-reverse order of scaffolded db fields.

### LINKED ISSUE #3802: Form Fields scaffolded from $db are now reversed in order
After changes released in 3.1.9, (see #3620) DataObject->db() now reverses the order of $db fields so child DataObjects can override their parents $db fields. But because FormScaffolder->getFieldList() uses DataObject->db() to scaffold form fields, we're getting scaffolded form fields in reverse order too.

sheadawson:
Fixes #3802

### DIFF
--- a/forms/FormScaffolder.php
+++ b/forms/FormScaffolder.php
@@ -74,7 +74,7 @@ class FormScaffolder extends Object {
 		}
 		
 		// add database fields
-		foreach($this->obj->db() as $fieldName => $fieldType) {
+		foreach(array_reverse($this->obj->db()) as $fieldName => $fieldType) {
 			if($this->restrictFields && !in_array($fieldName, $this->restrictFields)) continue;
 			
 			// @todo Pass localized title


camfindlay:
@tractorcow can you weigh in on this post 3.1.9 release? :) Wonder if we need a test in place for this to avoid any future regressions?

kinglozzer:
@sheadawson 85b4ba15fcfa83ef11569121020aa3b18840c70d reversed the order that classes (but not fields) were looped over, so I’m not sure whether this will entirely fix the issue. Take the following example:

```php
class MyObject extends DataObject {
  private static $db = array(
    'Title' => 'Varchar',
    'Description' => 'Text'
  );
}

class MyChildObject extends MyObject {
  private static $db = array(
    'Location' => 'Varchar',
    'Price' => 'Varchar'
  );
}
```

The expected order is: Title, Description, Location, Price. I think this change would result in Description, Title, Price, Location.

In the short term, I think we could change `DataObject::db()` to only reverse the classes array if `$fieldName` is present. That solves both this issues and the bug I originally was trying to fix.

This has pointed out another issue: if you don’t provide an argument to `DataObject::db()`, you can’t overload the field type (basically the same bug as #3620) - but that bug already exists independent of my suggested change.

tractorcow:
Hi @kinglozzer , there's also a bug I can see with your pull request.

Because you reverse the order you check classes, checkse for db('FieldName') will now correctly inherit, but db() will give the incorrect field types. E.g. if you override 'FieldName' with a new type, checking that field by itself will return your new type, but in the db() list the old type will remain.

My suggestion is to fix both issues in db method, rather than trying to fix it everywhere the method is called.

I would expect the following:
- Requests for a single field name should return the type specified by the highest class which defines it
- Requests for all DB fields should return the most highly defined specification for each field
- Requests for all DB fields should return the fields in the same order as per 3.1.8, regardless of the field type.

I think the original commit is fine, but a new array_merge will be needed to merge fields together to respect the above logic.

tractorcow:
See my comment on https://github.com/silverstripe/silverstripe-framework/pull/3623/files#r23183303

kinglozzer:
@tractorcow yes, that’s the bug I was talking about in my above comment. We need to remove the `array_reverse` from `DataObject::db()` if no argument is given, or calls to it will return child class db fields _before_ their parent classes (see my above example). I’ll put together a PR

tractorcow:
That would work. I wonder though, at what point do you have a separate method for each behaviour? A single method should have a single purpose, IMO.

tractorcow:
Closing in favour of #3804

PR #3804, opened 2015-01-19
FIX: DataObject::db returned fields in incorrect order, with incorrect data types

### LINKED ISSUE #3802: Form Fields scaffolded from $db are now reversed in order

kinglozzer:
fixes #3802

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -1666,9 +1666,14 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	 */
 	public function db($fieldName = null) {
 		$classes = ClassInfo::ancestry($this, true);
-		$items = array();
 
-		foreach(array_reverse($classes) as $class) {
+		// If we're looking for a specific field, we want to hit subclasses first as they may override field types
+		if($fieldName) {
+			$classes = array_reverse($classes);
+		}
+
+		$items = array();
+		foreach($classes as $class) {
 			if(isset(self::$_cache_db[$class])) {
 				$dbItems = self::$_cache_db[$class];
 			} else {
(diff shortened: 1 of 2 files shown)


kinglozzer:
@tractorcow or as an alternative, refactored into two methods as you suggested: https://github.com/kinglozzer/sapphire/commit/ed07bfadb3ddf53c118a3a65bb11db67cc8091a8

tractorcow:
For 3.1 this is going to be fine. :)

@sheadawson  can you check this fixes your issue please?

model/DataObject.php
```diff
@@ -1666,9 +1666,14 @@ public function belongs_to($component = null, $classOnly = true) {
...
-		$items = array();
 
-		foreach(array_reverse($classes) as $class) {
+		// If we're looking for a specific field, we want to hit subclasses first as they may override field types
+		if($fieldName) {
+			$classes = array_reverse($classes);
+		}
+
+		$items = array();
+		foreach($classes as $class) {
```
tractorcow: Don't forget to re-instate the exception for `$class === 'DataObject'`

tractorcow:
Thanks @kinglozzer really appreciate your prompt response in this. :)

sheadawson:
Yeap fixes my issue thanks guys!

tractorcow:
Thanks both @sheadawson and @kinglozzer

PR #3805, opened 2015-01-20
Fix #3794

### LINKED ISSUE #3794: CGI mode detection breaks install
Using the 3.1.9 release, I just tried to do a fresh install twice on our live server, but the install stops and fails when trying to detect URL rewriting "Friendly URLs are not working. This is most.....". The server runs PHP Version 5.4.35 with CGI/FastCGI enabled.

All seems installed but there is no way to access `/dev`, `/admin` or anything, getting a 404.

The errors comes down to the new basic auth rewrite rule added for CGI/FastCGI system. Added here https://github.com/silverstripe/silverstripe-framework/blob/b2ace2b76cf65f97d22e9591e2d393388ee86044/dev/install/install.php5#L1497
by this commit fbebf96d66d71f32766f84ef73a75d8e69910d8c

When removing this new RewriteRule, all goes back to normal. Maybe that rule should not be added automatically for all systems? Especially since this seems to be useful for test systems mainly?

cjsewell:
Fix #3794

### DIFF
--- a/dev/install/install.php5
+++ b/dev/install/install.php5
@@ -1494,7 +1494,7 @@ HTML;
 
 		if($base != '.') $baseClause = "RewriteBase '$base'\n";
 		else $baseClause = "";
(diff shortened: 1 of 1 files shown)


tractorcow:
Thanks @cjsewell for this, I hope this finally solves the noted issues!

PR #3809, opened 2015-01-21
DOCS Fix missing changelogs

camfindlay:
Reimplements some missing changelog pages after docs overhaul.

Thanks for Chris Darling for reporting.


camfindlay:
@willmorgan if you are up on your side of the world, fancy checking and merging this. Seems some of the changelogs went missing in the great docs migration. This corrects that and reinstates those missing based on the state of the 3.1 branch prior to the first commit introducing the restructure. :+1:

PR #3813, opened 2015-01-22
Flush should be off by default - consistant

nfauchelle:
Flushing should be off by default, just like it is when you are on the site normally.

It's not obvious that tests always flush when they run and you have to pass flush=0 to turn them off. People expect to have to flush to clear the cache, and the documentation already covers mentions about flushing.

Flushing all the time is slow. It adds 5-6 seconds for the tests even with sqlite. When your doing them a couple of times a minute it's pretty painful.

-> http://doc.silverstripe.org/en/developer_guides/testing/unit_testing/
*phpunit framework/tests '' flush=all *
*Run tests with optional `$_GET` parameters (you need an empty second argument)*

### DIFF
--- a/dev/TestRunner.php
+++ b/dev/TestRunner.php
@@ -78,9 +78,9 @@ class TestRunner extends Controller {
 	 * top of the loader stacks.
 	 */
 	public static function use_test_manifest() {
-		$flush = true;
-		if(isset($_GET['flush']) && $_GET['flush'] === '0') {
-			$flush = false;
+		$flush = false;
+		if(isset($_GET['flush']) && ($_GET['flush'] === '1' || $_GET['flush'] == 'all')) {
+			$flush = true;
 		}
 
 		$classManifest = new SS_ClassManifest(


kinglozzer:
Looks like this was changed intentionally in #3393. If that’s how things are going to stay, the docs should be updated to reflect that (i.e. remove references to `flush=1` when running tests).

nfauchelle:
I've found that I don't flush much at all, and feel it makes more sense to have the default as off.

It'd be like having dev mode as always flushing and having to turn that off by default.

dhensby:
How does have flushing off affect tests when they are run repeatedly especially after changes?

nfauchelle:
It depends on the change of course. If you edit something which needs a flush then you'd have to flush - just like in the browser normally. However code inside a function or html code (generally) doesn't need a flush.

Depends on your testing/development workflow - I would run a test, tweak some code, run the test, tweak some code etc. Some people might just run the tests once at the end of the day and always need to flush.

Regardless, I think it makes it consistent across the board - just like using the cli-script that to flush you pass it in - not have it on by default.

dhensby:
I do tend to agree, to be honest.

@halkyon I think we need you to weigh in on your original thinking here

chillu:
@halkyon original rationale: "We also flush the cache in any case when TestRunner::use_test_manifest()
is called. I don't see any reason why you wouldn't want to flush the cache when running tests."

I think that's probably a bit overshooting the goal. Auto-flush on test execution basically prevents you from doing TDD, at least in the way flush is currently implemented in SilverStripe (`*cough*` PSR4).

So +1 from me on merging this, pending insight from @halkyon

nfauchelle:
@chillu @dhensby 
Any update on this?

nfauchelle:
@halkyon Can we go with a merge on this?

tractorcow:
Documentation in docs/en/howto/phpunit-configuration.md needs to be updated, since it's now incorrect. Flushing is off by default for tests so we need to declare that.

halkyon:
@tractorcow indeed. I just saw that. I'll open a PR to fix it.

tractorcow:
Damn, since the docs have all been moved around I have no idea where that page is now. =(

halkyon:
Hrm. Yeah. Looks like `flush=0` is non-existent in the new docs too.

tractorcow:
Probably needs to be re-added to https://github.com/silverstripe/silverstripe-framework/blob/3.1/docs/en/02_Developer_Guides/06_Testing/00_Unit_Testing.md

halkyon:
Here ya go: https://github.com/silverstripe/silverstripe-framework/pull/4554

tractorcow:
Somehow auto-flush got re-introduced at https://github.com/silverstripe/silverstripe-framework/pull/4404. Doh!

PR #3814, opened 2015-01-22
FIX: dev/build not flushing manifests if site is in a subfolder

kinglozzer:
For example: http://localhost/silverstripe/dev/build would result in `$requestURL` being `silverstripe/dev/build`, but `BASE_URL . '/dev/build'` would be `/silverstripe/dev/build` (note leading slash).


kinglozzer:
Reopened in #3815, without creating a branch on this repo this time :see_no_evil:

PR #3815, opened 2015-01-22
FIX: dev/build not flushing manifests if site is in a subfolder

kinglozzer:
For example: http://localhost/silverstripe/dev/build would result in `$requestURL` being `silverstripe/dev/build`, but `BASE_URL . '/dev/build'` would be `/silverstripe/dev/build` (note leading slash).

### DIFF
--- a/core/Core.php
+++ b/core/Core.php
@@ -96,7 +96,7 @@ Injector::set_inst($injector);
 // The coupling is a hack, but it removes an annoying bug where new classes
 // referenced in _config.php files can be referenced during the build process.
 $requestURL = isset($_REQUEST['url']) ? trim($_REQUEST['url'], '/') : false;
-$flush = (isset($_GET['flush']) || $requestURL == 'dev/build' || $requestURL == BASE_URL . '/dev/build');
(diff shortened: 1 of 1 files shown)


dhensby:
Is this unit-testable?

kinglozzer:
Not sure, I have no idea how you’d approach testing something like this!

kinglozzer:
Bump :)

dhensby:
On sites not in a subdir, wouldn't the `BASE_URL` be an empty string? if so, the 2nd of the 3 conditions is redundant, no?

kinglozzer:
@dhensby Good spot, I’ve removed that redundant check now

dhensby:
Ok, have you tested this manually to ensure it works as expected?

dhensby:
I do find this quite a hacky solution, don't the other dev functions need this kind of fix too?

kinglozzer:
Yep, I’ve tested both in and not in a subdirectory, with and without trailing slashes and it works as expected.

It does feel hacky, but I’m just trying to fix a regression from #3677 rather than refactor anything :)

dhensby:
agreed

PR #3817, opened 2015-01-22
mysite/code/Page.php to mysite/code/Product.php

anjanawijesundara:
"mysite/code/Page.php" line in Displaying Results Section should be "mysite/code/Product.php"


kinglozzer:
Thanks @anjanawijesundara

PR #3818, opened 2015-01-22
FIX DataDifferencer sanitisation

micmania1:


model/DataDifferencer.php
```diff
@@ -160,38 +165,33 @@ public function ChangedFields() {
...
 			$fields = array_keys($this->toRecord->toMap());
 		}
-		
+
 		foreach($fields as $field) {
 			if(in_array($field, $this->ignoredFields)) continue;
 
-			if(!$this->fromRecord || $this->fromRecord->$field != $this->toRecord->$field) {
+			if(!$this->fromRecord
+				|| $this->fromRecord->obj($field)->getValue() != $this->toRecord->obj($field)->getValue()
```
tractorcow: Is obj necessary here? ->$field should just give the raw value.

tractorcow:
Can you please add a test case?

micmania1:
Yep, i'll add some tests now.

dhensby:
@micmania1 potentially fixed by #3894 ?

micmania1:
I'd be surprised if it was. I got stuck on this, primarily because there was a test failing (as seen) which is not a real use-case. many_many is not supported in DataObject diffs which is the opposite of what the test tries to prove. 

The test actually points out that things are broken and that its a workaround which boggles my mind - it makes the test is completely pointless.

I spent the best part of a day trying to add support for many_many fields without breaking the test and fixing the security issue, however I wasn't able to do both. 

There is currently too many ways to access a value:
$this->obj('MyField'); // goes off and does something
$this->getField('MyField'); // goes off and does something
$this->MyField; // goes off and does something.

All of these things do different things. Some of them return objects and some of them return DBField subclassed instances. Some of them vary depending on when you call them. 

There is no way that I could find to reliably call a many_many field, sanitise it and make the current test pass.

My advice would be to delete the current test, fix this, decide on a consistent way to access field values then support many_many field comparison in a future version.

dhensby:
`relObject` is probably what you're looking for to get a dot notation access to an object - though it doesn't look like it'd work to access `Parent.Child.Child.MethodName` just DB fields.

If there's a test in there proving a bug (and passing!?) then that's wrong and really should be marked as a skipped or incomplete test. Which test is it?

If you think you have a solution but the tests are broken, then it's fine to fix the tests!

dhensby:
How would one do a diff on many_many? Given that the relations aren't versioned... or are they??

tractorcow:
Yikes, sorry if I've forgotten about you and re-implemented an attempted fix over the top of your work. https://github.com/silverstripe/silverstripe-framework/pull/3894 is actually a fix @dhensby and I had worked on independently.

You are right though, we hadn't invested much effort in implementing has_many relations. Hope you can come up with a better solution.

dhensby:
@micmania1 you're call if you want to close this or try re-basing and improving what there is

micmania1:
I'll test the the RC today and make sure it hasn't missed anything. My fix seems to have changed DataDifferencer a lot more so I'll see if I can figure out why.

sminnee:
@micmania1 how was that fateful February?

micmania1:
I vaguely remember this but I wish I didn't. I think this is good to close.

PR #3828, opened 2015-01-26
API `byIDs` added to `ArrayList`

dhensby:
Added `byIDs` for consistent API with `DataList`

### DIFF
--- a/model/ArrayList.php
+++ b/model/ArrayList.php
@@ -56,7 +56,7 @@ class ArrayList extends ViewableData implements SS_List, SS_Filterable, SS_Sorta
 	 * @return bool
 	 */
 	public function exists() {
-		return (bool) count($this);
+		return !empty($this->items);
 	}
 
 	/**
@@ -477,6 +477,79 @@ class ArrayList extends ViewableData implements SS_List, SS_Filterable, SS_Sorta
 	 *          // aziz with the age 21 or 43 and bob with the Age 21 or 43
 	 */
 	public function filter() {
+
+		$keepUs = call_user_func_array(array($this, 'normaliseFilterArgs'), func_get_args());
+
+		$itemsToKeep = array();
+		foreach($this->items as $item){
+			$keepItem = true;
+			foreach ($keepUs as $column => $value) {
+				if ((is_array($value) && !in_array($this->extractValue($item, $column), $value))
+					|| (!is_array($value) && $this->extractValue($item, $column) != $value)
+				) {
+					$keepItem = false;
+					break;
+				}
+			}
+			if($keepItem) {
+				$itemsToKeep[] = $item;
+			}
+		}
+
+		$list = clone $this;
+		$list->items = $itemsToKeep;
+		return $list;
+	}
+
+	/**
+	 * Return a copy of this list which contains items matching any of these charactaristics.
+	 *
+	 * @example // only bob in the list
+	 *          $list = $list->filterAny('Name', 'bob');
+	 * @example // azis or bob in the list
+	 *          $list = $list->filterAny('Name', array('aziz', 'bob');
+	 * @example // bob or anyone aged 21 in the list
+	 *          $list = $list->filterAny(array('Name'=>'bob, 'Age'=>21));
+	 * @example // bob or anyone aged 21 or 43 in the list
+	 *          $list = $list->filterAny(array('Name'=>'bob, 'Age'=>array(21, 43)));
+	 * @example // all bobs, phils or anyone aged 21 or 43 in the list
+	 *          $list = $list->filterAny(array('Name'=>array('bob','phil'), 'Age'=>array(21, 43)));
+	 *
+	 * @param string|array See {@link filter()}
+	 * @return DataList
+	 */
+	public function filterAny() {
+		$keepUs = call_user_func_array(array($this, 'normaliseFilterArgs'), func_get_args());
+
+		$itemsToKeep = array();
+
+		foreach ($this->items as $item) {
+			foreach ($keepUs as $column => $value) {
+				$extractedValue = $this->extractValue($item, $column);
+				$matches = is_array($value) ? in_array($extractedValue, $value) : $extractedValue == $value;
+				if ($matches) {
+					$itemsToKeep[] = $item;
+					break;
+				}
+			}
+		}
+
+		$list = clone $this;
+		$list->items = array_unique($itemsToKeep, SORT_REGULAR);
+		return $list;
+
+	}
+
+	/**
+	 * Take the "standard" arguments that the filter/exclude functions take and return a single array with
+	 * 'colum' => 'value'
+	 *
+	 * @param $column array|string The column name to filter OR an assosicative array of column => value
+	 * @param $value array|string|null The values to filter the $column against
+	 *
+	 * @return array The normalised keyed array
+	 */
+	protected function normaliseFilterArgs($column, $value = null) {
 		if(count(func_get_args())>2){
 			throw new InvalidArgumentException('filter takes one array or two arguments');
 		}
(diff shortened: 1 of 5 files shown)


dhensby:
@silverstripe/core-team to be honest, I'm thinking that this function `byIDs` on DataList/ArrayList should just be deprecated.

It's a function that is effectively a hard coded filter which doesn't add any real benefit over doing `$this->filter('ID', $ids);`

Having it as part of the code base increases code maintenance and doesn't really provide any special flexibility.

tractorcow:
It's probably ok, since filtering by ID (in `DataList`) negates the necessity to do a classwalk to find the table for the filtered field. It doesn't really give the same benefit in ArrayList as it does there though.

tests/model/ArrayListTest.php
```diff
@@ -746,7 +746,8 @@ public function testCanFilterByEmpty() {
 		$this->assertFalse($list->canFilterBy('Age'));
 	}
 
-	public function testByID() {
+	public function testByID()
+	{
```
halkyon: *super nitpicky mode* brace shouldn't be on a newline ;-)

tractorcow: I had `setenv nitpickymode 0`, but yeah I saw that. :)

Welcome back @halkyon

dhensby: PHPStorm! Rage! Can I get away with saying.... PSR?

stevie-mayhew: Wasn't it you who added the .editorconfig files @dhensby ? :D

tractorcow:
The issue with travis is caused by travis-support generating the incorrect `version` tags for framework on the 3 branch (it should be 3.x-dev not dev-3). I'll look into this later. :D

tests/model/ArrayListTest.php
dhensby: Unfortunately you can't dictate code style like this, it's more white space based. :(

dhensby:
Ok, great. I've fixed that naughty curly brace going to the wrong line and I've also squashed the tests

dhensby:
Ok, so I noticed that the `SS_Filterable` interface has diverged a bit from what's implemented on `DataList` and `ArrayList` so I've updated the interface and implemented the method(s) as needed on `ArrayList`.

I've also opened issue #3847 so we can have some clarity on what should be returned by the `map` method.

As I've now edited the interface, I think that either this needs to go against master OR I need to pull out the interface changes and add them to master.

dhensby:
As a compromise we could put all these changes into 3 (except the interface changes) and then put the interface changes into 4.

Therefore we get a consistent API which is BC for 3.2 and then enforce the interface properly (non-BC) in 4?

dhensby:
bump @silverstripe/core-team

camspiers:
As mentioned on #4557 I think there is a mistake in ArrayList::filterAny, and I think tests should be added for it.

model/ArrayList.php
```diff
@@ -477,6 +477,72 @@ public function canFilterBy($by) {
...
+			}
+		}
+
+		$list = clone $this;
+		$list->items = $itemsToKeep;
+		return $list;
+
+	}
+
+	protected function normaliseFilterArgs() {
```
tractorcow: phpdoc please.

model/ArrayList.php
```diff
@@ -477,6 +477,72 @@ public function canFilterBy($by) {
...
+	 * @example // bob or anyone aged 21 or 43 in the list
+	 *          $list = $list->filterAny(array('Name'=>'bob, 'Age'=>array(21, 43)));
+	 * @example // all bobs, phils or anyone aged 21 or 43 in the list
+	 *          $list = $list->filterAny(array('Name'=>array('bob','phil'), 'Age'=>array(21, 43)));
+	 *
+	 * @param string|array See {@link filter()}
+	 * @return DataList
+	 */
+	public function filterAny() {
+		$keepUs = call_user_func_array(array($this, 'normaliseFilterArgs'), func_get_args());
```
tractorcow: Why not just make normaliseFilterArgs take a single $args parameter? code inspection tools hate code linked with call_user_func_array(), and it's a bit ugly.

model/ArrayList.php
```diff
@@ -477,6 +477,72 @@ public function canFilterBy($by) {
...
+
+		$itemsToKeep = array();
+
+		foreach ($this->items as $column => $value) {
+			$keepItem = false;
+			if ((is_array($value) && in_array($this->extractValue($item, $column), $value))
```
tractorcow:
```php
$extractedValue = $this->extractValue($item, $column);
$matches = is_array($value) ? in_array($extractedValue, $value) : $extractedValue == $value;
if($matches) {...}
```

Cut down on duplicate code. ;)

model/UnsavedRelationList.php
```diff
@@ -299,7 +299,7 @@ public function avg() {
 		throw new LogicException(__FUNCTION__ . " can't be called on an UnsavedRelationList.");
 	}
 
-	public function byIDs() {
+	public function byIDs(array $ids) {
```
tractorcow: Is this going to break semver for subclasses?

model/ArrayList.php
tractorcow: you may as well put it in a method for checking matches, then having all the filter methods call it. It seems to be duplicated logic in each? (exclude, filter, filterAny, etc).

model/UnsavedRelationList.php
dhensby: If someone's subclassed `UnsavedRelationList` and overloaded `byIDs` and NOT accepted an ID parameter... well... then I don't know what kind of world we are living in.

I suppose I could remove the type hinting if we thought there was a real risk of regression here; personally, I don't

tractorcow: Yeah, we probably should punish poor coding. :)

model/ArrayList.php
dhensby: Because the number of arguments changes the behaviour of these functions. the point is to turn that unknown number of args into a normalised array

model/ArrayList.php
dhensby: The matching changes slightly if you want to match any or all, though.

It would be better to fix that, but at the moment I just want to add these APIs in rather than refactor large amounts of `ArrayList`

dhensby:
feedback dealt with

model/ArrayList.php
```diff
@@ -477,6 +477,80 @@ public function canFilterBy($by) {
...
+	 * @param string|array See {@link filter()}
+	 * @return DataList
+	 */
+	public function filterAny() {
+		$keepUs = call_user_func_array(array($this, 'normaliseFilterArgs'), func_get_args());
+
+		$itemsToKeep = array();
+
+		foreach ($this->items as $item) {
+			$keepItem = false;
```
camspiers: looks like you can remove this now.

dhensby: removed, thanks :)

model/ArrayList.php
tractorcow: Yeah, and that number will just be the size of the array that's passed in.

```php
$keepUs = $this->normaliseFilterArgs(func_get_args());
```

and
```php
function normaliseFilterArgs($args) {}
```

tractorcow:
Dealings-with with fed-back

model/ArrayList.php
dhensby: hmmm. *personally* I don't like that, it's yet another method api convention that's different...

I suppose my personal preference here shouldn't make a big difference; if it makes a difference to code inspectio... i suppose it's worth it.

sminnee:
Hey @dhensby @tractorcow where are we at with this?

model/ArrayList.php
tractorcow: ok, fine. :D

tractorcow:
We are at merged.

dhensby:
yay, thanks

camspiers:
Nice work @dhensby!

PR #3829, opened 2015-01-27
Fix issue where TreeDropdownField loses value on show/hide

dhensby:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1157


dhensby:
I've also tidied up the template so that we are using templating functions a bit more

PR #3833, opened 2015-01-29
added feature for excluding page types from heireichy

pitchandtone:
Very useful if you have a large set of pages and the CMS will only show "Too many pages, show as list". Used on wellingtonnz.com for their listings (accomodation, events, site & activities etc)

### DIFF
--- a/forms/TreeDropdownField.php
+++ b/forms/TreeDropdownField.php
@@ -213,8 +213,7 @@ class TreeDropdownField extends FormField {
 			$properties,
 			array(
 				'Title' => $title,
-				'TitleURLEncoded' => rawurlencode($title),
-				'Metadata' => ($metadata) ? Convert::raw2att(Convert::raw2json($metadata)) : null
+				'Metadata' => ($metadata) ? Convert::raw2json($metadata) : null,
 			)
 		);
 
--- a/javascript/TreeDropdownField.js
+++ b/javascript/TreeDropdownField.js
@@ -56,7 +56,7 @@
 			
 				var linkTitle = strings.openLink;
 				if(linkTitle) this.find("treedropdownfield-toggle-panel-link a").attr('title', linkTitle);
-				if(this.data('title')) this.setTitle(decodeURIComponent(this.data('title')));
+				if(this.data('title')) this.setTitle(this.data('title'));
 				
 				this.getPanel().hide();
 				this._super();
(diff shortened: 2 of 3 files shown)


dhensby:
Can we have some tests/docs please.

Also, this needs to go against 3 branch

model/Hierarchy.php
```diff
@@ -617,9 +617,14 @@ public function numChildren($cache = true) {
 	 */
 	public function stageChildren($showAll = false) {
 		$baseClass = ClassInfo::baseDataClass($this->owner->class);
+		$hidden = array();
+		if ($baseClass::config()->get('hideFromHierarchy')) {
+			$hidden = $baseClass::config()->get('hideFromHierarchy');
```
tractorcow: Could you please document this feature? I suggest the following in Hierachy.php itself

```php

/**
 * A list of classnames to exclude from display in the hierarchy.
 * Note: Does not filter subclasses (non-inheriting)
 *
 * @var array
 * @config
 */
private static $hide_from_hierarchy = array();
```

I suggest to name the variable hide_from_hierarchy as it meets the static variable naming convention.

model/Hierarchy.php
```diff
@@ -641,9 +646,16 @@ public function liveChildren($showAll = false, $onlyDeletedFromStage = false) {
 		}
 
 		$baseClass = ClassInfo::baseDataClass($this->owner->class);
+
+		$hidden = array();
+		if ($baseClass::config()->get('hideFromHierarchy')) {
```
tractorcow: To be sneaky and/or lazy, if you add the config I suggested above you can do `$this->owner->config()->hide_from_hierarchy` and you don't need to check to see if the config exists.

model/Hierarchy.php
```diff
@@ -641,9 +646,16 @@ public function liveChildren($showAll = false, $onlyDeletedFromStage = false) {
...
+			$hidden = $baseClass::config()->get('hideFromHierarchy');
+		}
+
 		$children = $baseClass::get()
 			->filter('ParentID', (int)$this->owner->ID)
 			->exclude('ID', (int)$this->owner->ID)
+			->exclude('ClassName', $hidden)
```
tractorcow: I would ensure this filter isn't invoked if $hidden is empty.

model/Hierarchy.php
```diff
@@ -617,9 +617,14 @@ public function numChildren($cache = true) {
...
 		$baseClass = ClassInfo::baseDataClass($this->owner->class);
+		$hidden = array();
+		if ($baseClass::config()->get('hideFromHierarchy')) {
+			$hidden = $baseClass::config()->get('hideFromHierarchy');
+		}
 		$staged = $baseClass::get()
 			->filter('ParentID', (int)$this->owner->ID)
-			->exclude('ID', (int)$this->owner->ID);
+			->exclude('ID', (int)$this->owner->ID)
+			->exclude('ClassName', $hidden);
```
tractorcow: I would ensure this filter isn't invoked if $hidden is empty.

tractorcow:
This actually is a novel solution to the problem of too many pages. I hope this works well. :D

PR #3835, opened 2015-01-29
adding hide_from_hierachy

pitchandtone:
Very useful if you have a large set of pages and the CMS will only show "Too many pages, show as list". Used on wellingtonnz.com for their listings (accomodation, events, site & activities etc)

### DIFF
--- a/model/Hierarchy.php
+++ b/model/Hierarchy.php
@@ -38,6 +38,32 @@ class Hierarchy extends DataExtension {
 	 */
 	private static $node_threshold_leaf = 250;
 
+	/**
+	 * A list of classnames to exclude from display in both the CMS and front end
+	 * displays. ->Children() and ->AllChildren affected.
+	 * Especially useful for big sets of pages like listings
+	 * If you use this, and still need the classes to be editable
+	 * then add a model admin for the class
+	 * Note: Does not filter subclasses (non-inheriting)
+	 *
+	 * @var array
+	 * @config
+	 */
+	private static $hide_from_hierarchy = array();
+
+	/**
+	 * A list of classnames to exclude from display in the page tree views of the CMS,
+	 * unlike $hide_from_hierarchy above which effects both CMS and front end.
+	 * Especially useful for big sets of pages like listings
+	 * If you use this, and still need the classes to be editable
+	 * then add a model admin for the class
+	 * Note: Does not filter subclasses (non-inheriting)
+	 *
+	 * @var array
+	 * @config
+	 */
+	private static $hide_from_cms_tree = array();
+
 	public static function get_extra_config($class, $extension, $args) {
 		return array(
 			'has_one' => array('Parent' => $class)
@@ -606,6 +632,18 @@ class Hierarchy extends DataExtension {
 	}
 
 	/**
+	 * Checks if we're on a controller where we should filter. ie. Are we loading the SiteTree?
+	 *
+	 * @return bool
+	 */
+	public function showingCMSTree() {
+		if (!Controller::has_curr()) return false;
+		$controller = Controller::curr();
+		return $controller instanceof LeftAndMain
+			&& in_array($controller->getAction(), array("treeview", "listview", "getsubtree"));
+	}
+
+	/**
 	 * Return children from the stage site
 	 *
 	 * @param showAll Inlcude all of the elements, even those not shown in the menus.
@@ -614,9 +652,17 @@ class Hierarchy extends DataExtension {
 	 */
 	public function stageChildren($showAll = false) {
 		$baseClass = ClassInfo::baseDataClass($this->owner->class);
+		$hide_from_hierarchy = $this->owner->config()->hide_from_hierarchy;
+		$hide_from_cms_tree = $this->owner->config()->hide_from_cms_tree;
 		$staged = $baseClass::get()
-			->filter('ParentID', (int)$this->owner->ID)
-			->exclude('ID', (int)$this->owner->ID);
+				->filter('ParentID', (int)$this->owner->ID)
+				->exclude('ID', (int)$this->owner->ID);
+		if ($hide_from_hierarchy) {
+			$staged = $staged->exclude('ClassName', $hide_from_hierarchy);
+		}
+		if ($hide_from_cms_tree && $this->showingCMSTree()) {
+			$staged = $staged->exclude('ClassName', $hide_from_cms_tree);
+		}
 		if (!$showAll && $this->owner->db('ShowInMenus')) {
 			$staged = $staged->filter('ShowInMenus', 1);
 		}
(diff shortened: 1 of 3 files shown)


pitchandtone:
Moved from https://github.com/silverstripe/silverstripe-framework/pull/3833

micmania1:
The problem I see with this PR is that front-end functionality would be affected which may not be a desired affect.

*>>begin shameless plug*
https://github.com/micmania1/silverstripe-lumberjack
*>>end shameless plug*

wilr:
See also https://github.com/micschk/silverstripe-excludechildren

pitchandtone:
Above are good outcomes.

dhensby:
I'd say that this probably should still be a core feature. At the moment those modules both still show the currently edited page in the sitetree when editing even if it's meant to be hidden.

So a more complete integration would be better.

pitchandtone:
OK I need to add some tests to it, which is going to be a bit of a bitch.

dhensby:
@pitchandtone In my opinion it also will need an update to the `LeftAndMain` functions that update the site tree (see #3837)

model/Hierarchy.php
```diff
@@ -610,9 +622,13 @@ public function numChildren($cache = true) {
 	 */
 	public function stageChildren($showAll = false) {
 		$baseClass = ClassInfo::baseDataClass($this->owner->class);
+		$hidden = $this->owner->config()->hide_from_hierarchy;
 		$staged = $baseClass::get()
 			->filter('ParentID', (int)$this->owner->ID)
 			->exclude('ID', (int)$this->owner->ID);
+		if (count($hidden) > 0) {
```
dhensby: `!empty()` is slightly better here as we aren't interested in how many there are, just that there are some

model/Hierarchy.php
```diff
@@ -641,7 +658,9 @@ public function liveChildren($showAll = false, $onlyDeletedFromStage = false) {
 				'Versioned.mode' => $onlyDeletedFromStage ? 'stage_unique' : 'stage',
 				'Versioned.stage' => 'Live'
 			));
-
+		if (count($hidden) > 0) {
```
dhensby: see https://github.com/silverstripe/silverstripe-framework/pull/3835/files#r24806443

dhensby:
@pitchandtone any progress on this? It'd be a lovely feature IMO

pitchandtone:
What do y'all reckon? Was tossing up adding an extra argument to both stageChildren and LiveChildren which is $includeHidden = false.

dhensby:
There's already "showAll" and I don't think I fully see the need to have an option to show items in the sitetree that have been explicitly hidden for reasons of organisation.

pitchandtone:
Is anyone going to merge this?

model/Hierarchy.php
```diff
@@ -610,9 +622,18 @@ public function numChildren($cache = true) {
 	 */
 	public function stageChildren($showAll = false) {
 		$baseClass = ClassInfo::baseDataClass($this->owner->class);
-		$staged = $baseClass::get()
-			->filter('ParentID', (int)$this->owner->ID)
-			->exclude('ID', (int)$this->owner->ID);
+		$hidden = $this->owner->config()->hide_from_hierarchy;
+		if (empty($hidden)) {
+			$staged = $baseClass::get()
```
tractorcow:
```php
$staged = $baseClass::get()
	->filter('ParentID', (int)$this->owner->ID)
	->exclude('ID', (int)$this->owner->ID);
if($hidden) {
	$hidden = $hidden->exclude('ClassName', $hidden);
}
```

Also, your double `->exclude(` and `:not` filters negate each other don't they?

model/Hierarchy.php
```diff
@@ -634,15 +655,26 @@ public function liveChildren($showAll = false, $onlyDeletedFromStage = false) {
...
-			));
-
-		if(!$showAll) $children = $children->filter('ShowInMenus', 1);
+		$hidden = $this->owner->config()->hide_from_hierarchy;
+		if (empty($hidden)) {
+			$children = $baseClass::get()
+				->filter('ParentID', (int)$this->owner->ID)
+				->exclude('ID', (int)$this->owner->ID)
+				->setDataQueryParam(array(
+					'Versioned.mode' => $onlyDeletedFromStage ? 'stage_unique' : 'stage',
```
tractorcow: Same as above; Don't duplicate all of the logic if only a small part of your code changes depending on `$hidden` :)

tractorcow:
I have another question; Don't you think that the `AllChildren` should include filtered children? Looking at the API it feels a bit weird that `Children` is filtered (ok) but `AllChildren` is also filtered? Do we need an `ActuallyAllChildren`? :P

pitchandtone:
I have added an extra static hide_from_cms_tree to cater to both use cases. 

I have also tidied up logic to be more streamlined after @tractorcow's comments.

dhensby:
This needs reading, please

tractorcow:
> I have added an extra static hide_from_cms_tree to cater to both use cases.

Nice, thanks that's all I was worried about really. :)

model/Hierarchy.php
```diff
@@ -38,6 +38,19 @@ class Hierarchy extends DataExtension {
...
+	 * Especially useful for big sets of pages like listings
+	 * If you use this, and still need the classes to be editable
+	 * then add a model admin for the class
+	 * Note: Does not filter subclasses (non-inheriting)
+	 *
+	 * @var array
+	 * @config
+	 */
+	private static $hide_from_hierarchy = array();
+	private static $hide_from_cms_tree = array();
```
tractorcow: Please split PHPdocs for these. The second var will end up not actually documented.

model/Hierarchy.php
```diff
@@ -602,6 +615,17 @@ public function numChildren($cache = true) {
 	}
 
 	/**
+	 * Checks if we're on a controller where we should filter. ie. Are we loading the SiteTree?
+	 *
+	 * @return bool
+	 */
+	public function showingCMSTree() {
+		$controller = Controller::curr();
```
tractorcow: `Controller::curr()` throws an error if none is assigned. You can add a `Controller::has_curr()` check to protect against that. :)

tractorcow:
Feedback left. :D

I still feel a bit weird about $allChildren not really being all children, but in most cases it's up to the dev to implement the correct config that's appropriate for their project, and realise the implications, so it's ok.

pitchandtone:
If you're all happy with the above, I'll squash the commits into one.

tractorcow:
Perfect, squash rebase and fix tests please. ;D

pitchandtone:
@tractorcow any idea why we're not getting a result from Travis? I can't see which tests aren't working.

pitchandtone:
Just rebased, that might be it.

pitchandtone:
@tractorcow @dhensby we good to merge?

dhensby:
@pitchandtone it's still saying conflicts!



I'll try to merge by hand

dhensby:
merged by hand

micschk:
I've marked micschk/silverstripe-excludechildren as incompatible from 3.5 upwards & added instructions to the README to use this core functionality instead.

I know this PR is 2 years old by now, but leaving this remark here anyway; I still think the best solution to the 'too many children' problem would be if we could somehow force a page type to use the CMS built-in 'show as list' interface (eg in a 'Children' tab). I don't think this is currently possible so we have to resort to Gridfield/ModelAdmin. But list-view would imo result in a more consistent user experience for this situation.

PR #3836, opened 2015-01-29
ExistingAutocompleter should check for canEdit() permission

uniun:
GridFieldAddExistingAutocompleter should check for canEdit() permission.

### DIFF
--- a/tests/forms/gridfield/GridFieldAddExistingAutocompleterTest.php
+++ b/tests/forms/gridfield/GridFieldAddExistingAutocompleterTest.php
@@ -1,15 +1,19 @@
 <?php
 class GridFieldAddExistingAutocompleterTest extends FunctionalTest {
 
-	protected static $fixture_file = 'GridFieldTest.yml';
+	protected static $fixture_file = 'GridFieldAddExistingAutocompleterTest.yml';
 
-	protected $extraDataObjects = array('GridFieldTest_Team', 'GridFieldTest_Player', 'GridFieldTest_Cheerleader');
+	protected $extraDataObjects = array(
+		'GridFieldAddExistingAutocompleterTest_Team',
+		'GridFieldAddExistingAutocompleterTest_Player',
+		'GridFieldAddExistingAutocompleterTest_Cheerleader'
+	);
 	
 	function testScaffoldSearchFields() {
 		$autoCompleter = new GridFieldAddExistingAutocompleter($targetFragment = 'before', array('Test'));
-		$gridFieldTest_Team = singleton('GridFieldTest_Team');
+		$gridFieldTest_Team = singleton('GridFieldAddExistingAutocompleterTest_Team');
 		$this->assertEquals(
-			$autoCompleter->scaffoldSearchFields('GridFieldTest_Team'), 
+			$autoCompleter->scaffoldSearchFields('GridFieldAddExistingAutocompleterTest_Team'), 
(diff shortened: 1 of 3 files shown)


dhensby:
There's now a test failure, please can you look into it and provide a test for this? thanks

dhensby:
@uniun are you likely to make any progress on the PR?

Also, should the canEdit permission dictate if you can add the item as a relation? It's not really an "edit" as such just an association.

uniun:
Sorry, I've been quite busy lately. I'll look on it.
And yes, I agree this is not an "edit". But it should check for canView at least I think.

dhensby:
No problem

Definitely canView... but typically that'll be true because it should be used to affect if it is shown on the front end of the site :/

Really tricky as actually there's probably a new permission required for this kind of thing. Maybe a config option needed to say whether to check a permission at all and, if so, which one?

uniun:
Maybe. But how about GridFieldViewButton then? This component also checks for canView.

I understand typically that'll true and it should be used on frontend of the site. But sometimes this permission is usefull on backend too.
For example, if you want to "hide" members of "Administrator" group (or any other) from other members it can be achieved with canView.

dhensby:
> Maybe. But how about GridFieldViewButton then? This component also checks for canView.

Ok, lets use `canView` then

> For example, if you want to "hide" members of "Administrator" group (or any other) from other members it can be achieved with canView.

good point and this is an example where it would be very useful

uniun:
Sorry, but the test still fails. This is the first time I write/edit the test :)

uniun:
So what what should I do now?
1. add canView() to the GridFieldTest_Team object in GridFieldTest.php
2. change team object in GridFieldAddExistingAutocompleterTest to a new object, e.g. GridFieldAddExistingAutocompleterTest _Team?

tests/forms/gridfield/GridFieldAddExistingAutocompleterTest.php
```diff
@@ -111,3 +111,16 @@ public function Form() {
 		return new Form($this, 'Form', new FieldList($field), new FieldList());
 	}
 }
+
+class GridFieldTest_Team extends DataObject implements TestOnly {
```
dhensby: This class already exists, it seems. Either you need to pick a new name or you should just amend the existing class to have the canView function you need (assuming that doesn't break other tests)!

uniun:
I've decided to create new DataObject for this test. I don't want to change default team object because it may brake other test. Also any other test (e.g. GridFieldExportButtonTest) that checks canView permission, do not use default team object.

uniun:
Damn, test still fails... But now there is something GridFieldPaginatorTest? I didn't change it.

dhensby:
That test still needs the `GridFieldTest_Cheerleader` object and that needs to be defined in the `$extraDataObjects` array for the `GridFieldPaginatorTest` test

dhensby:
Ah so, you've renamed classes that were relied upon in another test. Those classes need to remain really, you only need to create one new class for this test rather than rename all of the existing ones.

uniun:
Hmm. I think I need to create at least two of them: GridFieldTest_Team and GridFieldTest_Player. 
$field = new GridField('testfield', 'testfield', $player->Teams(), $config);
Gridfield uses player's teams, so I need to change Teams relation model to new one.

PR #3837, opened 2015-01-29
Add hook to updatetreenodes for customisation of tree

dhensby:
Somewhat related to #3835

This adds a hook to updatetreenodes to allow plugs like https://github.com/micmania1/silverstripe-lumberjack and https://github.com/micschk/silverstripe-excludechildren to enforce hidden children from the tree even through AJAX requests.

Currently, a page being edited will appear in the sitetree as the response to the ajax request can't be customised to stop pages being shown.

### DIFF
--- a/admin/code/LeftAndMain.php
+++ b/admin/code/LeftAndMain.php
@@ -977,6 +977,7 @@ class LeftAndMain extends Controller implements PermissionProvider {
 				'PrevID' => $prev ? $prev->ID : null
 			);
 		}
+		$this->extend('updateUpdatedTreeNodes', $data, $request);
 		$this->response->addHeader('Content-Type', 'text/json');
 		return Convert::raw2json($data);
 	}


dhensby:
cc @micmania1 @micschk @pitchandtone

micmania1:
Thanks for looking into this. Although its unintended, its actually quite useful to be able to flick between recently edited posts so i've just avoided it and nobody has felt strongly enough to look into it until now (or even report it).

Looking at the way the whole method works i'd be keen to rewrite it. With the hook where it is, you'd have to loop through $data, create the relevant DataObject and then check if it should display or not. Its already looping and executing multiple queries per turn so I would be hesitant in adding the extra overhead.

Ideally, you'd be able to hook into a DataList or DataQuery to filter out any records that you don't need.

micschk:
@dhensby, I agree the list returned should be able to be extended/filtered.
However, I actually think this behaviour could be considered a 'feature not a bug', user experience wise. I'm overlaying a 'this page is actually hidden under its parent' icon on gridfieldpages (pictures worth a thousand words, icons at least 8)

dhensby:
@micmania1 I agree. this PR is more for a discussion point than final solution. (hence no docs or tests).

@micschk I'm not so sure. Both devs I've spoken to think it's a bug and it's confusing to a user why it would disappear after a refresh.

It's definitely a side-effect and not an intended feature. If a user added several pages, then saw them in the tree and tried to re-order them, they'd break the order of their child pages quite badly.

sminnee:
> @micmania1 I agree. this PR is more for a discussion point than final solution. (hence no docs or tests).

I'm gonna close it off, then, @dhensby, since discussion appears to have stalled...

PR #3841, opened 2015-01-31
Fix large foreign key dropdowns being changed to text fields

textagroup:
The following commit disabled using datalists with over a 100 records due to php memory limits.
Would like to consider a different approach by using a hash instead of a object list.

4b8895c8
BUGFIX Don't scaffold has_one relations into a DropdownField in ForeignKey->scaffoldFormField() if more than 100 records would be created, to avoid exceeding PHP memory (fixes #6776)

### DIFF
--- a/model/fieldtypes/ForeignKey.php
+++ b/model/fieldtypes/ForeignKey.php
@@ -41,11 +41,14 @@ class ForeignKey extends Int {
 			// Don't scaffold a dropdown for large tables, as making the list concrete
 			// might exceed the available PHP memory in creating too many DataObject instances
 			if($list->count() < 100) {
-				$field = new DropdownField($this->name, $title, $list->map('ID', $titleField));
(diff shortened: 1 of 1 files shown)


kinglozzer:
I’m not entirely sure that this will actually make any difference to how much memory is used: calling `->toArray()` on the map will still loop over every `DataObject` in the same way that `DropdownField` will loop over the source. It’s doing the same work, just at a different time. Have you tested this to see how much difference it makes?

dhensby:
If the source items weren't pulled from the DB until render time it'd allow instantiation of dropdowns that could then never be used and that'd make a difference.

Perhaps `DropdownField`'s use of `DataList`s needs to be a bit more clever to save loading objects it doesn't need to.

sminnee:
It looks like this has stalled so I'm closing this PR. Feel free to reopen if you think it's appropriate.

PR #3851, opened 2015-02-04
fixes #3850: RestfulService not using the cache

### LINKED ISSUE #3850: RestfulService never using the cache
Hello,
from my understanding at in [framework/api/RestfulService.php, line 187](https://github.com/silverstripe/silverstripe-framework/blob/aec84303956d31f8db9f288b4aa5223f2de4dda4/api/RestfulService.php#L187), you're checking if flush is set to ignore the cache, but you're missing a negation in the if:

`if($this->cache_expire > 0 && self::$flush` ...

should be

`if($this->cache_expire > 0 && !self::$flush` ...

Am I right? (the documentation says "unless" too)

To test this you can add the following line after the else:

`Debug::log('API request: '.$url.' -- '.$this->cache_expire. ' -- flush: '. (self::$flush ? 'Y' : 'N'));`

Thanks.

giovdk21:


dhensby:
good spot. thanks

PR #3854, opened 2015-02-04
BUG Fix developer output in redirection script

tractorcow:


tractorcow:
No idea what's going on with travis... it looks like it's failing to install phpunit 3.7.x-dev?

tractorcow:
https://github.com/minkphp/Mink/commit/87a18d72e4aea93c4f46f21083f2987fe8af87d0

Updated behat-extension to use 1.6@stable instead and that should fix it.

kinglozzer:
@tractorcow Another build failure [here](https://travis-ci.org/silverstripe/silverstripe-framework/jobs/49716398), looks like the same issue but for behat/mink-selenium2-driver 1.2.\*-dev vs 1.2.*@stable? I’d correct it myself but wanted to double-check as I’m not familiar with Behat :)

tractorcow:
Yeah It looks like the behat guys shot themselves in their own foot there :) https://github.com/Behat/MinkSelenium2Driver/blob/master/composer.json still points to the unversionable 1.6-dev

tractorcow:
I see, they've moved to https://github.com/minkphp/MinkSelenium2Driver/blob/master/composer.json. We need to update our reference to 1.7

tractorcow:
This should fix it https://github.com/silverstripe-labs/silverstripe-behat-extension/commit/5bf9eb4961b7073698ef5702772d238ee1f860bb

tractorcow:
Yay that fixed it. ;)

PR #3858, opened 2015-02-06
FIX issue where empty composite fields created a fieldlist with empty items

dhensby:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1156


dhensby:
:)

PR #2230, opened 2013-07-15
Improved RestfulService

dhensby:
After quite a bit of usage of the RestfulService, we've found a fair few issues and nice-to-haves that we've implemented.

We've taken feedback from this #2157

PR #3471, opened 2014-09-04
MINOR Uses SS_Cache to cache RestfulService responses instead of filesys...

patbolo:
...tem

xmlresponse_* files are saved in the temp folder, are not garbage collected (they can fill up a disk if not actively monitored) and are not in line with the usual approach for managing cached objects in the framework. Using APC or similar other backend is then impossible for these files. This PR porposes to use SS_Cache for these files as well. Also the current code will use the cached object if available if the live call failed, the new code doesn't provide this fallback mechanism, but whether this fallback is a good thing is really debatable, as the returned cached object is stalled, and this error can be part of the spec for the API (ie 404 resource not found anymore)

PR #3864, opened 2015-02-09
Improved RestfulService - 3 branch

dhensby:
See #2157 and #2230

### DIFF
--- a/api/RestfulService.php
+++ b/api/RestfulService.php
@@ -28,6 +28,16 @@ class RestfulService extends ViewableData implements Flushable {
 	 * @config
 	 * @var array
 	 */
+	private static $cachable_methods = array(
+		'GET',
+		'HEAD',
+		'OPTIONS'
+	);
+
+	/**
+	 * @config
+	 * @var array
+	 */
 	private static $default_proxy;
 
 	/**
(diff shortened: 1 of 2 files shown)


dhensby:
Although I think this should still get into core, I think we should aim to either remove RestfulService or make it a wrapper from `Guzzle` as there are a lot of parts of this class I'd like to re-write.

I know @chillu was in favour of taking in Guzzle instead.

In fact, I think we should probably pull `RestfulService` out of core. It's only used ONCE in the entire framework (and CMS) and that's for requesting OEMBED responses.

It's a lot of code (and tests) to maintain for one function

wilr:
+1 for using Guzzle rather than maintaining our own. @dhensby something for master I guess so happy for this to be included for 3.2? 4.0 I'd imagine deprecating the class is acceptable.

dhensby:
Ok, rebased to make this mergable

api/RestfulService.php
```diff
@@ -164,25 +243,38 @@ protected function constructURL(){
...
+		else {
+			$cacheExpire = 0;
+		}
 
 		// Check for unexpired cached feed (unless flush is set)
 		//assume any cache_expire that is 0 or less means that we dont want to
 		// cache
-		if($this->cache_expire > 0 && !self::$flush
+		if ($canCache && $this->cache_expire > 0 && !self::$flush
 				&& @file_exists($cache_path)
```
tractorcow: Why don't we use the SS_Cache api here?

tractorcow:
I've had my eye on guzzle replacement :) One day someone will have time I hope.

Even better would be to move this out of core into a module.

api/RestfulService.php
dhensby: This point was made previously and some of the work has been done here: https://github.com/silverstripe/silverstripe-framework/pull/3471

It wasn't something I was focussing on when I initially made these improvements and this is just a rebase of old work.

My view has shifted to getting people to use guzzle over this and adding a RestfulService poly-fill for those that want legacy support.

Core only uses RestfulService once in Oembed.php and that could be done with a raw curl request instead (as it's so trivial).

I'd therefore opt to remove any kind of client from core and users can require guzzle as needed

stevie-mayhew:
Have you seen https://github.com/colymba/silverstripe-restfulapi - we've been using it in favour of the default RestfulService as it seems to work a bit better :smile:

dhensby:
@stevie-mayhew do you mean instead of `RestfulServer`?

stevie-mayhew:
Bah. Sorry. Continue.

tractorcow:
Is it ok if we address this for 3.3? Is this PR semver?

dhensby:
Yes, this is semver IMO

api/RestfulService.php
```diff
@@ -139,9 +139,62 @@ public function basicAuth($username, $password) {
 
 	/**
 	 * Set a custom HTTP header
+	 * @deprecated
 	 */
 	public function httpHeader($header) {
-		$this->customHeaders[] = $header;
+		Deprecation::notice('3.2', 'Use "addHeader" instead');
```
sminnee: Should probably be 3.4 now...

sminnee: Or 4.0? I can't remember how we decided to mark that...

sminnee:
@dhensby did you want to incorporate these changes or move to deprecated `RestfulService` entirely in SS4? I think we should do 1 or the other...

api/RestfulService.php
tractorcow: 4.0, as that is the version this deprecated api will be removed in.

tractorcow:
+1 for deprecating. We should add the deprecation notice in 3 branch, and physically remove it in 4.0.

dhensby:
deprecate

Issue #3354, opened 2014-08-02
FormFields don't have a way to load a set a dataValue vs presentation value


dhensby:
At the moment, there is only one way to set a value on a FormField (using `setValue`)

This is used by both `Form`s to load submitted data as well as DB data, so if you want to manipulate your data for presentation vs storage, it's very difficult.

For example, storing currency as cents/pence but wanting to present it as pounds/dollars:

``` php

class MoneyPenceField extends NumericField {
    /**
     * for the template
     */
    public function Value() {
        return $this->value / 100;
    }

    /**
     * for the DB
     */
    public function dataValue() {
        return $this->value;
    }

    /**
     * This could be a submitted value from a form (so 200) or a value from a DB (20000)
     * as it is set either as $field->setValue($obj->Money); or `loadDataFrom($request)` on
     * the form
     */
    public function setValue($val) {
    }
}
```

Really there needs to be a "set value from submission" or "set raw value" function so presentation and submissions can be handled distinctly

dhensby:
A bit of a "hack" is being used here: https://github.com/silverstripe/silverstripe-framework/pull/3008/files#r22024977

But I don't like that.

GuySartorelli:
There's a lot of conversation here about jQuery and not much about the original issue - but in any case this is a pretty old issue which hasn't seen any movement.
If this is a genuine problem for someone and that someone is willing to put in some work to come up with an approach and implement a change, we can revisit it - but until then I'm closing this issue.

Issue #2161, opened 2013-06-26
I18n NumericField 


zigterman:
When you use the NumericField and set your local to a country that uses a , instead of a . you get some problems.

The validator works, there is a check with Zend_Locale_Format and your local wil be used. But when you save the value a 0 is send to the database. The problem is in 
```php
public function dataValue() {
	return (is_numeric($this->value)) ? $this->value : 0;
}
```
the function expects a . instead of a ,. So the is_numeric returns false, you can use the check from Zend_Local_Format, but if your database expects a numer you need a . instead of a ,. I fixed this by making the function look like this:

```php
public function dataValue() {
	return Zend_Locale_Format::isNumber(trim($this->value), array('locale' => i18n::get_locale())) ? str_replace(',', '.', $this->value) : 0;
}
```

There still is one little problem now, when you reload your data in the field, it wil automatically use a . instead of ,. So when you save the data, you get een a validation error.

To fix this I override the setValue function to something like this:
```php
	public function setValue($value) {
		return parent::setValue(Zend_Locale_Format::toNumber($value));
	}
```

You have to do the first replace, when you load the data from the database there is a . instead of the expected ,.

I'm not working in php a lot, so maybe there is a better way to fix this. That's why I send a bug report and not a pull request.

smindel:
This is an important issue for users of non-EN locales and I am concerned that we are about to cut corners again. I appreciate your taking on this matter and I want to put in my two cents:

Short:
- values stored in FormFields must not be localised but raw
- a dedicated setter for localised values must be added to the FormField API and this setter has to be used throughout the framework when setting values on FormFields to user supplied input
- alternatively the i18n class can be promoted to a full tier in the stack that form data has to pass through

Long:

Localization is supposed to help overcome differences in the representation of values for different locales in the view. While the user has every right to enter a VAT rate as "0,19" instead of "0.19", as developers we don't want to be bothered with all the babel in the business logic.

1. This being said, generally the API of a FormField has to provide dedicated setters and getters for raw data (which i prefer to be the primary representation of the data, because this is what we deal with in the business logic) as well as for the view (localise values for the rendered forms and *de*localise user provided localised input). This requires a bidirectonal lossless conversion between the representations. Basically the FormFields need the same mechanics as we have in the ORM, which translates back and forth between the raw data representation (php types) and the persistency representation (types of the storage engine)

[...]

PR #3866, opened 2015-02-09
Fix broken localised NumericField

smindel:
- Values stored in FormFields must not be localised but raw.
- A dedicated setter for localised values must be added to the FormField API and this setter has to be used throughout the framework when setting values on FormFields to user supplied input
- See #2161, #3354

### DIFF
--- a/forms/NumericField.php
+++ b/forms/NumericField.php
@@ -18,14 +18,7 @@ class NumericField extends TextField {
 			return true;
 		}
 		
-		require_once THIRDPARTY_PATH."/Zend/Locale/Format.php";
-
-		$valid = Zend_Locale_Format::isNumber(
-			trim($this->value), 
-			array('locale' => i18n::get_locale())
-		);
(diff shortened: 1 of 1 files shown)


dhensby:
This needs to be against 3 branch at the minimum (if it's backwards compatible)

dhensby:
Also, this is an API that really should be added to the `FormField` not just numeric field. We'll need tests too.

tractorcow:
Yet again another dupe of an earlier solution I made at https://github.com/silverstripe/silverstripe-framework/pull/3008

Did you want to try out my fix on your client's code and see if that fixes it for them?

chillu:
Hey Andi, how are things over in Hamburg? :) Closing this as a duplicate of #3008, which has had a fair bit of peer review already. The `getViewValue()` is an interesting idea, but I don't think we should introduce it in one FormField only (or in a stable release line).

PR #3869, opened 2015-02-09
MINOR SS_Query properties now protected

dhensby:
Changing the properties' visibility to protected allows better extension of SS_Query.


willmorgan:
Hmm, this is technically opening up an API in some ways, is it not?

dhensby:
Semver is for public apis only

Issue #3173, opened 2014-05-29
Folder TreeDropdownField


nglasl:
This issue is in relation to a TreeDropdownField of type folder, and regards the search functionality throwing errors (also an incorrect result set being returned once these errors have been corrected). An example can be seen when either viewing a file from the AssetAdmin, or when selecting a file using an UploadField.

Basically when you trigger a search for a child folder's name, it attempts a database query to determine each parent folder.

https://github.com/silverstripe/silverstripe-framework/blob/3.1/forms/TreeDropdownField.php#L464

The problem here being that by default, a folder table does not exist. This in turn causes a server error. I have currently implemented a patch for our client project which corrects this issue as follows:

```php
$table = $this->sourceObject;
while(!ClassInfo::hasTable($table)) {
	$table = get_parent_class($table);
}
```

However this reveals a larger issue when using the **AssetAdmin**. The search no longer raises an error, but returns an incorrect result set (one level too deep), which actually looks to have been incorrect to start with if you search for the root folder name without the patch in place! If you remove the following line, you'll also begin to have file elements returned.

https://github.com/silverstripe/silverstripe-framework/blob/3.1/filesystem/File.php#L393

Therefore it looks as though this problem goes a little deeper than expected. I've managed to narrow it down to the location in which this TreeDropdownField has been created from **File.php**, which must conflict with something else (perhaps from the previous file selection screen). The field definition and configuration are identical to that found from an UploadField, however that one functions correctly. Perhaps some kind of cached data is being pulled in at the last second?

NGlasl:
This has mostly been addressed through the previous commits, however running through some brief testing still shows a couple of minor issues.

1. Using the **AssetAdmin** Folder **TreeDropdownField**, the user defined filter doesn't end up being applied, as the full result set is always returned.

2. When using a Folder **TreeDropdownField** elsewhere, it seems to think that folders with files under them are expandable (using the little arrow symbol), however these don't actually expand when selected.

nglasl:
Having another quick test with 3.3, it seems that **1** is still an issue, however **2** has been resolved. I don't think this issue is very important among others, therefore I'm going to close this for now and come back to it when I see the need.

PR #3663, opened 2014-11-21
BUG: Search for Folders in TreeDropdownField

jonom:
Fixes silverstripe/silverstripe-framework#3173
Fixes silverstripe/silverstripe-framework#3035

PR #3871, opened 2015-02-09
BUGFIX Use correct query when searching for items managed by a tree drop...

patbolo:
...down field #3173

### DIFF
--- a/forms/TreeDropdownField.php
+++ b/forms/TreeDropdownField.php
@@ -460,15 +460,18 @@ class TreeDropdownField extends FormField {
 				if ($row->ParentID) $parents[$row->ParentID] = true;
 				$this->searchIds[$row->ID] = true;
 			}
+			
+			$sourceObject = $this->sourceObject;
+			
 			while (!empty($parents)) {
-				$res = DB::query('SELECT "ParentID", "ID" FROM "' . $this->sourceObject
-					. '" WHERE "ID" in ('.implode(',',array_keys($parents)).')');
+				$items = $sourceObject::get()
+					->filter("ID",array_keys($parents));
 				$parents = array();
 
-				foreach($res as $row) {
-					if ($row['ParentID']) $parents[$row['ParentID']] = true;
-					$this->searchIds[$row['ID']] = true;
-					$this->searchExpanded[$row['ID']] = true;
+				foreach($items as $item) {
+					if ($item->ParentID) $parents[$item->ParentID] = true;
+					$this->searchIds[$item->ID] = true;
+					$this->searchExpanded[$item->ID] = true;
 				}
 			}
 		}
--- a/forms/TreeDropdownField.php
+++ b/forms/TreeDropdownField.php
@@ -431,27 +431,25 @@ class TreeDropdownField extends FormField {
 			$res = call_user_func($this->searchCallback, $this->sourceObject, $this->labelField, $this->search);
 		} else {
 			$sourceObject = $this->sourceObject;
-			$wheres = array();
+			$filters = array();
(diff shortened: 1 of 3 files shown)


forms/TreeDropdownField.php
```diff
@@ -460,15 +460,26 @@ protected function populateIDs() {
 				if ($row->ParentID) $parents[$row->ParentID] = true;
 				$this->searchIds[$row->ID] = true;
 			}
+			
+			$sourceObject = $this->sourceObject;
+			$qSuffix = '';
+			// The parent of a file can also be a file (and not necessarily a folder, see File::has_one()) but we are 
+			// only interested in displaying folders in the dropdown field, when searching for folders.
+			if ($sourceObject == 'Folder') {
+				$qSuffix .= '"ClassName" IN (\''.implode('\',\'',Convert::raw2sql(ClassInfo::subclassesFor('Folder'))).'\')';
```
dhensby: What is the need for this argument? Why is `Folder` special?

The query builder will automatically do this for `Folder::get()` just like it would for `Page::get()`.

patbolo: True this is not needed.
When fixing the issue I had some files returned as part of the list of parent folders, but I can't reproduce it now. Maybe I added the check against ClassName before I turned the query from raw to ORM based.

dhensby:
Ok, can you elaborate a bit more on what the bug was and how this fixes the issue?

Is it something that can get some test coverage?

dhensby:
how strange this only fails in postgres - @tractorcow any ideas?

patbolo:
My gut feeling is that it is related to some case sensitiveness during search, unrelated to the fix for that issue, but surfaced by the new test. I'll investigate furthermore next week

tests/forms/TreeDropdownFieldTest.php
```diff
@@ -0,0 +1,97 @@
...
+			'FileTest-folder1-subfolder1 is found, nested under folder1'
+		);
+
+		// Looking for two files with 'sub' in their name, both under the same folder
+		$file1 = $this->objFromFixture('File','subfolderfile1');
+		$file2 = $this->objFromFixture('File','subfolderfile2');
+		$cssPath = 'ul.tree li#selector-TestTree-'.$subfolder->ID.' li#selector-TestTree-'.$file1->ID.' a';
+		$firstResult = $parser->getBySelector($cssPath);
+		$this->assertEquals(
+			(string)$firstResult[0],
```
tractorcow: Perhaps add another assertion prior to this that count($firstResult) > 0? you could include the search data (folder, file id, etc) in the assertion description.

tractorcow:
I'm not sure what's going on... I might need to run these tests locally with a debugger and postgres.

patbolo:
The original issue is explained in the ticket I linked to, #3173. It boils down to the raw query which assumes that the source object has a dedicated table, while it is not true necessarily, ie Folder.

I have written a test for the fix, but as there isn't any tests for the whole class, the test is for the search function, not only for that particular issue. That also surfaced an issue with case insensitive search in postgres, so I have grabbed the logic in master and backported it into this branch, so that test can pass.

dhensby:
is this related to #3663?

patbolo:
@dhensby yes and also https://github.com/oilee80/silverstripe-framework/commit/986df31a680fd1ad5399826d15754ae9eb47f744

jonom:
Closed mine as this is clearly a better fix - thanks @patbolo!

chillu:
Looks good to me, using the ORM properly in core classes is always a nice improvement - and in this case also a bug fix.

PR #3874, opened 2015-02-10
Private visibility is toxic to extensibility

dhensby:
I've removed declarations of private vars and functions for better extensibility of classes

### DIFF
--- a/control/RequestProcessor.php
+++ b/control/RequestProcessor.php
@@ -13,7 +13,7 @@ class RequestProcessor implements RequestFilter {
 	 *
 	 * @var array
 	 */
-	private $filters = array();
+	protected $filters = array();
 
 	public function __construct($filters = array()) {
 		$this->filters = $filters;
--- a/control/injector/Injector.php
+++ b/control/injector/Injector.php
@@ -130,27 +130,27 @@ class Injector {
 	 *
 	 * @var array
 	 */
-	private $serviceCache;
+	protected $serviceCache;
 
 	/**
 	 * Cache of items that need to be mapped for each service that gets injected
 	 *
 	 * @var array
 	 */
-	private $injectMap;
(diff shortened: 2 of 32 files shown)


willmorgan:
cc: @silverstripe/core-team

tractorcow:
I would LIKE this to go into 3.1, but I'd consider this an enhancement, so it'll need to go into 3.

jonom:
Would it also be fair to say that "private and protected visibility is toxic to horizontal extensibility" (via Extensions / DataExtensions)?

tractorcow:
That's true to an extend, @jonom

dhensby:
PUN SO INTENDED!

jonom:
:+1: !

jonom:
Seriously though, if there are protected helper functions doing the heavy lifting in SS classes, isn't that closing things off a bit?

jonom:
Just encountered this issue for the first time yesterday when using an extension hook on a form

dhensby:
`protected` can be overridden (and importantly called) by descendant classes, `private` cannot (which is why `protected` is more suitable).

I suppose if we're allowing decorators on objects.. they really need to have access to these functions too (to be able to call them). But we can't just make every function public, it's massively polluting to the API.

This is a limitation in the way we "decorate" objects I suppose :/

@jonom what are you trying to do with an extension on form?

jonom:
Yeah it's in decorating that's the issue.

I was trying to modify the Insert Media form in the CMS and having a lot of trouble (dev list post incoming about that) and at one point tried to recreate parts of the form, but couldn't do it efficiently because the helper functions were protected.

It might not be a significant issue, it's just one example I encountered and wondered if it's a wide problem that's interfering with that ideal of being able to override any single function in the framework.

PR #3876, opened 2015-02-11
Allow the paddedresize to take another hex value to specify a transparency on the padded color

nfauchelle:
When using the paddedresize sometimes it'd be useful to have the padding transparent as the color behind the image can change.

This pull allows you to add another hex value to the color. Using E7 on the end you can make it fully transparent, or use a hex value for an integer between 0 or 127 for a different level.

Also fixed the doc where it used a shortcode color which doesn't work.

Update the doc to explain about the padded image color a little.

### DIFF
--- a/filesystem/GD.php
+++ b/filesystem/GD.php
@@ -347,9 +347,11 @@ class GDBackend extends Object implements Image_Backend {
 		$r = hexdec(substr($webColor,0,2));
 		$g = hexdec(substr($webColor,2,2));
 		$b = hexdec(substr($webColor,4,2));
-		
+		if(strlen($webColor) == 8) {
+			$a = hexdec(substr($webColor,6,2));
+			return imagecolorallocatealpha($image, $r, $g, $b, $a);
+		}
 		return imagecolorallocate($image, $r, $g, $b);
-		
 	}
 
 	/**
--- a/filesystem/GD.php
+++ b/filesystem/GD.php
@@ -342,13 +342,17 @@ class GDBackend extends Object implements Image_Backend {
 			return $useAsMinimum ? $this->resizeByWidth( $maxWidth ) : $this->resizeByHeight( $maxHeight );
 	}
 	
-	public static function color_web2gd($image, $webColor) {
+	public static function color_web2gd($image, $webColor, $transparencyPercent=0) {
 		if(substr($webColor,0,1) == "#") $webColor = substr($webColor,1);
 		$r = hexdec(substr($webColor,0,2));
 		$g = hexdec(substr($webColor,2,2));
 		$b = hexdec(substr($webColor,4,2));
-		if(strlen($webColor) == 8) {
-			$a = hexdec(substr($webColor,6,2));
+
+		if($transparencyPercent) {
+			if($transparencyPercent > 100 || $transparencyPercent === 'true') {
+				$transparencyPercent = 100;
+			}
+			$a = 127 * bcdiv($transparencyPercent, 100, 2);
 			return imagecolorallocatealpha($image, $r, $g, $b, $a);
 		}
 		return imagecolorallocate($image, $r, $g, $b);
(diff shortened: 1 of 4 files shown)


dhensby:
Nice addition. What was your reasoning with adding it to the 2nd param over just adding a 3rd param?

this is a new feature, so should be against 3 branch.

nfauchelle:
Not sure what you mean about the 2nd param. The color is the third param in the function. It would be nice to have separated it out and have another param for the transparency percent but you can't have more than 3 params when calling from template.

    FooImage.paddedResize(100,200,FFFFFF,100) 

would be nice, but that 4th param isn't picked up.

nfauchelle:
>> this is a new feature, so should be against 3 branch.

Since it was backwards compatible / non breaking I thought it would go in 3.1
I've only just been told that 3.1 is now on feature freeze.....

dhensby:
Fourth parameter, yes, sorry. Is that limitation still in the templating language :(

It'd have to go against 3 at least

nfauchelle:
The more I look at it, I think I should also allow setting the color to 'transparent' which would set it to FFFFFFE7 so people can set a level as well if they want, but I think having transparent is probably the more popular use case.

I will sent through another PR and you can merge into 3?

kinglozzer:
I just tested adding a fourth parameter and it seems to be picked up okay

nfauchelle:
Wow. You are right! I didn't check it in the right place before.
I will change this so it can take in a transparency value as a 4th argument - it's much more readable.

dhensby:
@nfauchelle any progress on this? would be a great feature to get in.

Please can you open a PR against 3 branch with the fourth param and we'll get it in :)

nfauchelle:
@dhensby 
Just working on the patch now, and added ImagickBackend support... pull req coming shortly.

dhensby:
<3

nfauchelle:
@dhensby 
Patch added :)

filesystem/GD.php
```diff
@@ -358,8 +364,9 @@ public static function color_web2gd($image, $webColor) {
 	 * @param width
 	 * @param height
 	 * @param backgroundColour
+	 * @param transparencyPercent
 	 */
-	public function paddedResize($width, $height, $backgroundColor = "FFFFFF") {
+	public function paddedResize($width, $height, $backgroundColor = "FFFFFF", $transparencyPercent=0) {
```
dhensby: There should be spacing around the `=`

`$transparencyPercent = 0`

filesystem/ImagickBackend.php
```diff
@@ -183,17 +183,42 @@ public function resizeByHeight($height) {
 	 *
 	 * @param int $width
 	 * @param int $height
+	 * @param int $transparencyPercent
 	 * @return Image_Backend
 	 */
-	public function paddedResize($width, $height, $backgroundColor = "FFFFFF") {
+	public function paddedResize($width, $height, $backgroundColor = "FFFFFF", $transparencyPercent=0) {
```
dhensby: same as above: https://github.com/silverstripe/silverstripe-framework/pull/3876/files#r24853051

filesystem/ImagickBackend.php
```diff
@@ -183,17 +183,42 @@ public function resizeByHeight($height) {
...
 	}
+
+	/**
+	 * Convert a percentage (or 'true') to a two char hex code to signifiy the level of an alpha channel
+	 *
+	 * @param $percent
+	 * @return string
+	 */
+	public function calculateAlphaHex($percent) {
+		if($percent > 100 || $percent == 'true') {
```
dhensby: What are we accepting a string of 'true'? I think this kind of mixed variable makes the code less consistent and more error prone.

nfauchelle: It means in the template you can then say 

    PaddedResize(100,100,FFFFFF,true)

and then it will be transparent, just like doing 100. It's a string check for true since the template parser gives it as a string and not boolean. Happy to remove thou, just thought was a nice addition :)

dhensby: Yer, I definitely think we need to reduce this kind of mixed param type. "true" is no easier than "100" in my opinion (it's even 1 extra char to right "true" in the template!) it helps reduce complexity by keeping the param type consistent.

nfauchelle: Fair call :) will remove and cleanup whitespace and send through.

dhensby: thanks a bunch. remember the docs too :)

nfauchelle:
@dhensby
Added patch for those changes

dhensby:
thanks @nfauchelle

Issue #3481, opened 2014-09-09
js error in lib.js


sb-relaxt-at:
There is a script error in the isExternal check in admin/javascript/lib.js (https://github.com/silverstripe/silverstripe-framework/blob/master/admin/javascript/lib.js#L247).
documentUrl is undefined, I assume it should be document. It can be reproduced by selecting menu point "help" in the backend. The issue has been first mentioned in #2635 and seems to be a minor one.

PR #3877, opened 2015-02-11
Fixed " js error in lib.js #3481"

AdenFraser:


kinglozzer:
Thanks @AdenFraser. Could you also replace the references to `documentBase`? I think that’s suffering from the same issue

AdenFraser:
There we go :)

kinglozzer:
Squashed and merged in d555b5b26ee97f574c862e62c040d33a965a8736, thanks @AdenFraser :)

Issue #3477, opened 2014-09-08
Image_Backend interface constructor needs updating (master)


kinglozzer:
As pointed out by @jedateach, #2859 added an extra argument to [`GD::__construct()`](https://github.com/kinglozzer/sapphire/blob/791ee7171b20ffd66f2d6d0c18589adfc22b6680/filesystem/GD.php#L38) that we should add to `Image_Backend::__construct()`. It’ll also mean that `Imagick_Backend::construct()` needs to have the argument added.

kinglozzer:
Closed via 3b5a7c22b32438da70873600f55f9a243cae7cf1

PR #3879, opened 2015-02-11
Fixed "Image_Backend interface constructor needs updating (master) #3477"

AdenFraser:
For both the Image_Backend Model and Imagick_Backend filesystem class, as well as updated / added Docblocks for the aforementioned classes and also GDBackend.


kinglozzer:
Thanks @AdenFraser

PR #3881, opened 2015-02-11
Patch for issue #2986

### LINKED ISSUE #2986: Upgrade php 5.3 -> 5.4 memory issue (cache related)
After upgrading from PHP 5.3 to PHP 5.4 vice versa, SilverStripe generates an out-of-memory error, that can be fixed by doing a /dev/build?flush=1 or by emptying the cache by hand. Versions 3.1.2 and 3.1.3. Happened on a nix hosting platform and can be reproduced on MAMP.

Not so much of an issue if only you know in advance. Unfortunately I'm not fluent enough to come up with an answer nor where it comes from, SilverStripe or Zend.

JorisDebonnet:
This fixes #2986 by tying the automatically generated cache folder to the php (minor) version. 

Generates folder names such as: silverstripe-cache-php506-home-user-domains-...

### DIFF
--- a/core/TempPath.php
+++ b/core/TempPath.php
@@ -64,8 +64,9 @@ function getTempParentFolder($base = null) {
 	}
 
 	// failing the above, try finding a namespaced silverstripe-cache dir in the system temp
-	$cacheFolder = DIRECTORY_SEPARATOR . 'silverstripe-cache' . str_replace(array(' ', '/', ':', '\\'), '-', $base);
-	$tempPath = sys_get_temp_dir() . $cacheFolder;
+	$tempPath = sys_get_temp_dir() . DIRECTORY_SEPARATOR . 
+	            'silverstripe-cache-php' . (PHP_MAJOR_VERSION * 100 + PHP_MINOR_VERSION) . 
+	            str_replace(array(' ', '/', ':', '\\'), '-', $base);
 	if(!@file_exists($tempPath)) {
 		$oldUMask = umask(0);
 		$worked = @mkdir($tempPath, 0777);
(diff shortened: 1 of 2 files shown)


JorisDebonnet:
Squashing commits is still not possible in the github.com interface, so if that's required, I'm afraid someone will have to do it for me (somehow)!

dhensby:
Heh, nice. Why not just use `PHP_VERSION`? You're ignoring the "release" part of the PHP_VERSION and also the "EXTRA" which *could* affect the manifest's ability.

dhensby:
Plus, this could go against 3.1 branch as it's not changing any public APIs

JorisDebonnet:
I was currently assuming that only minor php versions caused the issue, although I admit I have no proof for that. Suppose we could just put the full version in there then! 

As for going against the 3.1 branch, then I should just close this pull request and create a new one right?

tractorcow:
> Heh, nice. Why not just use PHP_VERSION? You're ignoring the "release" part of the PHP_VERSION and also the "EXTRA" which could affect the manifest's ability.
> Plus, this could go against 3.1 branch as it's not changing any public APIs

I concur with this. I think that if you have multiple release versions of PHP running concurrently you have other issues on your plate. :) I think the simpler the better.

To change branches you'll need to make a new PR, sorry. :D

halkyon:
Keep in mind that if you use `PHP_VERSION`, Debian etc include some extra bits at the end, e.g. 5.3.3-7+squeeze25 etc. I would make sure that values like this don't break the path, but apart from that seems like a good idea.

JorisDebonnet:
Aight, thanks all for the feedback, a new pull request #3882 has been created!

PR #3888, opened 2015-02-12
Private visibility is toxic to extensibility

dhensby:
See #3874

### DIFF
--- a/control/RequestProcessor.php
+++ b/control/RequestProcessor.php
@@ -13,7 +13,7 @@ class RequestProcessor implements RequestFilter {
 	 *
 	 * @var array
 	 */
-	private $filters = array();
+	protected $filters = array();
 
 	public function __construct($filters = array()) {
 		$this->filters = $filters;
--- a/control/injector/Injector.php
+++ b/control/injector/Injector.php
@@ -130,27 +130,27 @@ class Injector {
 	 *
 	 * @var array
 	 */
-	private $serviceCache;
+	protected $serviceCache;
 
 	/**
 	 * Cache of items that need to be mapped for each service that gets injected
 	 *
 	 * @var array
 	 */
-	private $injectMap;
+	protected $injectMap;
 
 	/**
 	 * A store of all the service configurations that have been defined.
 	 *
 	 * @var array
 	 */
-	private $specs;
+	protected $specs;
 
 	/**
 	 * A map of all the properties that should be automagically set on all
 	 * objects instantiated by the injector
 	 */
-	private $autoProperties;
+	protected $autoProperties;
 
 	/**
 	 * A singleton if you want to use it that way
@@ -165,7 +165,7 @@ class Injector {
 	 *
 	 * @var boolean
 	 */
-	private $autoScanProperties = false;
+	protected $autoScanProperties = false;
 
 	/**
 	 * The default factory used to create new instances.
--- a/core/Config.php
+++ b/core/Config.php
@@ -751,8 +751,8 @@ class Config_LRU {
 		$obj->c = ++$this->c;
 	}
 
-	private $hit = 0;
-	private $miss = 0;
+	protected $hit = 0;
+	protected $miss = 0;
 
 	public function stats() {
 		return $this->miss ? ($this->hit / $this->miss) : 0;
--- a/core/Extension.php
+++ b/core/Extension.php
@@ -35,7 +35,7 @@ abstract class Extension {
 	 * Reference counter to ensure that the owner isn't cleared until clearOwner() has
 	 * been called as many times as setOwner()
 	 */
-	private $ownerRefs = 0;
+	protected $ownerRefs = 0;
 
 	public $class;
 
--- a/dev/SapphireTest.php
+++ b/dev/SapphireTest.php
@@ -768,7 +768,7 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 	/**
 	 * Helper function for the DOS matchers
 	 */
-	private function dataObjectArrayMatch($item, $match) {
+	protected function dataObjectArrayMatch($item, $match) {
 		foreach($match as $k => $v) {
 			if(!array_key_exists($k, $item) || $item[$k] != $v) return false;
 		}
@@ -778,7 +778,7 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 	/**
 	 * Helper function for the DOS matchers
 	 */
-	private function DOSSummaryForMatch($dataObjectSet, $match) {
+	protected function DOSSummaryForMatch($dataObjectSet, $match) {
 		$extracted = array();
 		foreach($dataObjectSet as $item) $extracted[] = array_intersect_key($item->toMap(), $match);
 		return var_export($extracted, true);
--- a/dev/SapphireTest.php
+++ b/dev/SapphireTest.php
@@ -116,7 +116,8 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 	/** 
 	 * Helper arrays for illegalExtensions/requiredExtensions code
 	 */
-	private $extensionsToReapply = array(), $extensionsToRemove = array();
+	protected $extensionsToReapply = array(), $extensionsToRemove = array();
+
 
 	
 	/**
--- a/dev/SapphireTestReporter.php
+++ b/dev/SapphireTestReporter.php
@@ -321,7 +321,7 @@ class SapphireTestReporter implements PHPUnit_Framework_TestListener {
 	 * @param obj PHPUnit_Framework_AssertationFailedError, PHPUnit error
 	 * @return array
 	 */
-	private function getTestException(PHPUnit_Framework_Test $test, Exception $e) {
+	protected function getTestException(PHPUnit_Framework_Test $test, Exception $e) {
 		// get the name of the testFile from the test
 		$testName = preg_replace('/(.*)\((.*[^)])\)/', '\\2', $test->toString());
 		$trace = $e->getTrace();
--- a/dev/TeamCityListener.php
+++ b/dev/TeamCityListener.php
@@ -7,7 +7,7 @@
  */
 class TeamCityListener implements PHPUnit_Framework_TestListener {
 
-	private function escape($str) {
+	protected function escape($str) {
 		return strtr($str, array(
 			"\n" => '|n',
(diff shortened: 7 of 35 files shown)


MatthewBonner:
I can't believe you have done this, surely there should be some sort of discussion around what kind of re factoring needs to take place? For instance, the properties you have made protected should be private, and you should either have protected or public accessors and mutators so that the class cannot be changed or the type if the property contains a scalar value.

micmania1:
I think more consideration needs to be given to what is being changed and why. I don't think changing every single instance is beneficial.

For example, I'm not sure its even possible to extend Config_LRU or Injector.

Also, because they're private, its much safer to remove or change their function in future versions because there is no chance that they're being used elsewhere in bespoke code or modules (or even in the framework).

dhensby:
> there should be some sort of discussion around what kind of re factoring needs to take place?

Yes, this PR is to instigate that

> the properties you have made protected should be private, and you should either have protected or public accessors and mutators so that the class cannot be changed or the type if the property contains a scalar value

Can you give a specific example? Why should they be private specifically and other properties not? What's so special about the properties I've changed?

@micmania1 I agree with you in general terms. In fact, [Symfony 2 recently changed all their protected properties to private](http://fabien.potencier.org/article/47/pragmatism-over-theory-protected-vs-private) but, as that article states:

> Having a few well defined extension points force the developer to extend your library the right way instead of hacking your code.


My main aim with this PR is to bring in some consistency - a better solution can be found for master or when we're somewhere closer to a level that Symfony is.

dhensby:
> I'm not sure its even possible to extend Config_LRU or Injector

Injector definitely can: `Injector::set_inst(new MyExtendedInjector());`

MatthewBonner:
There are a few bad practises with the following example, but it highlights that complex code can lead to undesirable results, which may be hard to debug, this is only a simple example, if I was to replace it with some code from SS then the debugging would become a real nightmare.

```php
<?php

class Task {
  private $ran = FALSE;

  public function hasRan () {
    return $this->ran;
  }

  public function init () {
    set_memory_limit(-1);
  }

  public function run () {
    // run task
  }
}

class TaskWithNoInit {
  private $ran = FALSE;

  public function hasRan () {
    return $this->ran;
  }

  public function run () {
    // run task
  }
}

class Parent {
  protected $task;

  public function __construct () {
    $this->task = new Task;
  }

  protected function ensureInitLogic () {
    if ($this->task->hasRan()) {
      $this->task->init();
    }
  }

  public function doLogic () {
    $this->task->run();
  }
}

class Child {
  protected function ensureInitLogic () {
    if ($this->task->hasRan()) {
      $this->task = new TaskWithNoInit;
      parent::ensureInitLogic();
    }
  }
}

?>
```

MatthewBonner:
The following shows a way to solve the problem, making it easier to debug, and prevent programmers from making silly mistakes...

```php
<?php

interface ITask {
  function hasRan();
  function run();
}

class Task implements ITask {
  private $ran = FALSE;

  public function hasRan () {
    return $this->ran;
  }

  public function init () {
    set_memory_limit(-1);
  }

  public function run () {
    // run task
  }
}

class TaskWithNoInit implements ITask {
  private $ran = FALSE;

  public function hasRan () {
    return $this->ran;
  }

  public function run () {
    // run task
  }
}

class Parent {
  private $task;

  public function __construct () {
    // This poorly designed class requires the specific Task class, opposed to a class implementing the ITask interface
    $this->task = new Task;
  }

  protected function getTask () {
    return $this->task;
  }

  protected function ensureInitLogic () {
    if ($this->task->hasRan()) {
      $this->task->init();
    }
  }

  public function doLogic () {
    $this->task->run();
  }
}

class Child {
  protected function ensureInitLogic () {
    if ($this->getTask()->hasRan()) {
      // This won't work, because $this->task does not exist
      $this->task = new TaskWithNoInit;
      parent::ensureInitLogic();
    }
  }
}

?>

**Comment by MatthewBonner** [comment]
To answer the following question:
"Why should they be private specifically and other properties not? What's so special about the properties I've changed?"

Realistically, in my view, all properties should be private due to PHP's lack of features in making properties read-only, and due to the type juggling. If we enforce stricter rules then the code as a result will naturally become cleaner and more robust.

**Comment by dhensby** [comment]
@MatthewBonner your example code is very hard to follow because you aren't explicitly defining inheritance and there are no comments to highlight in what way this relates to any of the changes in my PR.

For arguments sake I present #3869 

If you want to extend SS_Query to query against a standard array then the access to the private properties is ESSENTIAL. A simple example below:

```php

class ArrayQuery extends SS_Query {

    protected $record = array();

    public function nextRecord() {
        return isset($this->record[$this->rowNumber + 1]) ? $this->record[$this->rowNumber + 1] : false;
    }

}

```

This could be done with protected or public accessor mthods but that's just adding another layer... Ok, if the class had those methods, then I'd not need to change the visibility - but I don't see an inherent advantage in making them private.

I do suppose it means the ability to more easily remove those properties late for some other ones, but I guess that's a massive refactor job for the future and not inline with current conventions

dhensby:
> To answer the following question:
> "Why should they be private specifically and other properties not? What's so special about the properties I've changed?"
> 
> Realistically, in my view, all properties should be private due to PHP's lack of features in making properties read-only, and due to the type juggling. If we enforce stricter rules then the code as a result will naturally become cleaner and more robust.

Ok, so the real crux here is you think all properties should be `private`? It's not that I'm changing specific properties that you feel are somewhat special.

If that's the case, then you're welcome to take on that responsibility and start putting that against master.

But, in the absence of a convention that says all properties should be private, this is a solution that brings the few exceptions within the framework in-line with the rest of the code base until a better and more robust solution is implemented.

MatthewBonner:
What is wrong with...

```php
<?php

class SS_Query {
  private $record;

  protected function getRecord () {
    return $this->record;
  }
}

?>
```

dhensby:
Well, your solution misses the point and also `$record` shouldn't be inherently on an `SS_Query`... it's meant to be abstract to the storage method (array, DB query, API result set, etc)

With your example accessing `$this->rowNumber` still throws a fatal error for accessing a private property on a parent class...

I don't disagree that adding a layer of functions to get the property value is a better solution, but it's just a significantly more difficult and disruptive change than this one

MatthewBonner:
The example was based on the code in this context, not in the context of SilverStripe, for the sake of simplicity.

You have to weigh up the breaking changes in comparison with making a change that requires more effort in the short-term.

Just because it is easier to make private properties protected, it doesn't mean it is the right way to solve the problem of extending the API.

dhensby:
This PR is in the context of SS, so we should keep our discussion of the PR in that context too.

> Just because it is easier to make private properties protected, it doesn't mean it is the right way to solve the problem of extending the API.

Agreed, but this is a fix that fits with the current conventions and consistency is better than inconsistency.

MatthewBonner:
Well at some point I've got to lead the upgrade of a SS 3.1 site to SS 3.2 when it is released, due to the size of the website it has taken over a year to upgrade from SS 2 to SS 3 and then implement a new theme. We've had countless problems with sloppy code, causing hours of debugging nightmares and causing progress to slow, when realistically the upgrade process should have been relatively smooth. Your proposing making a change that is eventually going to lead to having to make breaking changes, and it is this disregard to breaking changes in SS that I really despise, it makes my job, and the job of others incredibly tedious when it doesn't need to be. All we have to do is follow some simple rules and people will be amazed at the quality we are churning out.

I wouldn't argue about one or two protected properties per class, but making them all protected is a mistake I made some time ago, and rather than you having to go through the same learning curve on why this is a bad idea, I'd prefer to encourage you to avoid doing such a thing in the hope that I can save you and others the same pain.

MatthewBonner:
I've updated my example to keep it in context with SilverStripe.

camfindlay:
@dhensby probably something to raise at the next core committers hangout too after a good discussion in the dev list perhaps to thrash out the pros/cons of the change.

sunnysideup:
+1 on the approach below (and above)  - BUT where it gets really powerful is shown in method doSomethingElse, because this is where you can influence every use of records within a framework of classes in a pretty efficient way (by overriding the parent getRecords methods in a child class): 
```php


<?php

class SS_Query {
  private $record;
  private $rowNumber ;

  protected function getRecord() {
    return $this->record;
  }

  protected function getRowNumber() {
    return $this->rowNumber;
  }
 
 private function doSomethingElse(){
  //NOTE - we are NOT accessing the variable directly here.
  $myRecords = $this->getRecord();
 }
}

class ArrayQuery extends SS_Query {
    public function nextRecord() {
        $record = $this->getRecord();
        $rowNumber = $this->getRowNumber();
        return isset($record[$rowNumber + 1]) ? $record[$rowNumber + 1] : false;
    }
}

?>

tractorcow:
The point I'd like to make that, any class should be extendible, either vertically (by subclassing) or to a more restricted degree, horizontally (by adding extensions). A good framework should respect the open closed principle, meaning that it should be able to have its behaviour overridden without having to edit internal code.

Either of the above approaches are reasonable here, although we could definitely have another discussion about which getters are appropriate as protected or public.

I personally prefer having getters, and enforcing usage of these over directly accessing private variables, so I'm leaning towards the solution put forward by Matt, but we would have to be more thorough in our review of each usage of private variables in each class.

Whatever solution we take needs to support the mechanisms and coding practices we've developed for SilverStripe that might not apply to other frameworks (e.g. extensions).

So, let's take it to the devlist and see what the community prefer.

tractorcow:
Did someone add a reply then delete it? :P

sminnee:
Hey @dhensby do you think there is value in keeping this PR open? I'm not sure if it's mergeable as-is and I'm not sure what the path forward would be. It's stalled for a year...

tractorcow:
I think we can close this, but only if @dhensby is happy for us to do so.

In future, maybe we just add accessors where needed, rather than converting member visibility.

sminnee:
Yeah my main concern with "everything protected" is that it substantially increases our public API, which limits what we can do in minor versions. I guess we can say "protected methods are not part of the public API" but then we're basically saying that anyone who uses this extension-point is in no-mans land, which isn't great either.

dhensby:
I think either we make them all private and have a well defined API or we open them up and don't.

As opening them up and having a badly defined API is only going to cause us problems, I think a better solution is to add setter/getter methods for these

PR #3889, opened 2015-02-12
Several changes that deprecate the old written tutorials in favour of Lessons section

unclecheese:


chillu:
Good stuff! Do you also want to replace the "tutorial" link in the main doc menu with a "lessons" one? http://take.ms/L1V2g. On that note: Why is the "Lessons" navigation item in the main top menu highlighted with a blue border even if I'm not on that page? This might have been a measure to highlight the existing "Training" item above the others (questionable as well), but that only works IMHO when highlighting a single entry, not multiple ones. My suggestion is to remove the highlights on both entries. /cc @camfindlay

camfindlay:
Yeah I'd be open to making the language consistent. "Lessons" is what we are calling things and it is easy enough to change the folder name to rejig the nav. Just make sure to put in a 301 redir too.

unclecheese:
The idea is that the "Tutorials" section of the docs will no longer exist, and we'll set up 301s from all the old links to go to the appropriate lesson. This pull request is a bit of an interim measure. It deprecates most of them, but leaves them available for users who may want to still use them for whatever reason. More importantly, not all of the topics have been covered by Lessons at this point, so we can't do a full hatchet job quite yet. Sig is keen to get to a mostly-deprecated state for the UK trainings, which is why this is getting ushered through a bit prematurely.

unclecheese:
Where are we at with this? Good to merge?

chillu:
Merged - sorry for the delay, and thanks for your hard work on improving docs, Aaron :)
Does anybody know what the deal is with the double highlight in the menu though? Looks odd.

camfindlay:
@chillu highlight I think gets set in the CMS, can be changed, what are you proposing?

chillu:
@camfindlay As I said above: "My suggestion is to remove the highlights on both entries."

camfindlay:
I agree only having 1 highlight makes more sense, as for dropping both... I think that was @clarkepaul 's design call... happy to only highlight lessons (as this is legitimately something we want to get people into) :)

clarkepaul:
Lessons was added after the IA was made that I did so I haven't been involved how we present these areas. I think the idea was that lessons would be different enough from tutorials to complement each other rather than compete. Happy to discuss with someone if need be.

camfindlay:
Cheers @clarkepaul the new lessons are set to replace the old tutorials. We also had that training link in the same space. I'd be keen to drop the highlight on training and keep on lessons as really only one call to action should be highlighted and I think lessons adds the most value for new community members so should be easy to find (I'd actually be keen to swap the homepage call to action from "demo" to the lessons).

PR #3890, opened 2015-02-12
Cleaning up Test DataObjects to ensure TestOnly is implemented

dhensby:
Pretty self-explanatory.

Enforcing that the framework implements TestOnly on DataObjects defined for test purposes only and adds them to `$extraDataObjects` on tests that need them where they haven't been included.


dhensby:
thanks @wilr

Issue #3891, opened 2015-02-12
Wrong temp folder in tiny_mce_gzip


sb-relaxt-at:
The cache_dir of tiny_mce_gzip.php does not respect TEMP_FOLDER if it is set, instead it uses getTempFolder(). This is an issue if someone wants to use a different TEMP_FOLDER path. 

Steps to reproduce
- Silverstripe 3.1 or master branch
- Set TEMP_FOLDER via _ss_environment.php, do not create a silverstripe-cache in the webroot
- Login and edit a page
- The TinyMCE won't load due to the missing cache directory (in case it can not create 'silverstripe-cache' in the webroot)

I will suggest a fix for this in a PR.

sminnee:
Fixed by #3892

PR #3892, opened 2015-02-12
Changed cache_dir to respect TEMP_FOLDER

sb-relaxt-at:
Fixes issue #3891 by using TEMP_FOLDER as cache_dir


dhensby:
Ok, I think this is a bug fix.

Thanks for this

PR #3894, opened 2015-02-13
BUG Lots of encoding fixes

tractorcow:


forms/FormAction.php
```diff
@@ -17,31 +17,47 @@
...
 	protected $buttonContent = null;
 	
 	/**
 	 * Create a new action button.
 	 *
-	 * @param action The method to call when the button is clicked
-	 * @param title The label on the button
-	 * @param form The parent form, auto-set when the field is placed inside a form 
+	 * @param string $action The method to call when the button is clicked
+	 * @param string $title The label on the button. This should not plain text, not escaped as HTML.
```
hafriedlander: Makes sense wibble

view/ViewableData.php
```diff
@@ -273,13 +273,11 @@ public function castingClass($field) {
 	 * @return string 'xml'|'raw'
 	 */
 	public function escapeTypeForField($field) {
-		if(!$class = $this->castingClass($field)) {
-			$class = self::$default_cast;
-		}
-		
+		$class = $this->castingClass($field) ?: $this->config()->get('default_cast');
```
hafriedlander: Everywhere else uses `Config::inst()->get('ViewableData', 'default_cast', Config::FIRST_SET);`, but that might be wrong as `$class` might end up being null

tractorcow: That's been changed across ViewableData.php to properly respect config.

PR #3896, opened 2015-02-13
Use double quotes to escape quotes in CSV export

### LINKED ISSUE #3895: Escaping quotes in the CSV export for Excel
I am having some issues with the CSV export and Excel. 
On line [152 of GridFieldExportButton](https://github.com/silverstripe/silverstripe-framework/blob/3.1/forms/gridfield/GridFieldExportButton.php#L152) quotes are escaped with `\"` but it is [recommended](http://stackoverflow.com/a/216623/908257) to use a double quote `""`. Is there any reason that the slash quote is used here?

Happy to submit a PR but just wanted to check there was no specific reason as it seems to work fine in Openoffice etc just as well.

Ben-Speakman:
Resolves #3895


dhensby:
thanks

PR #3897, opened 2015-02-13
FIX: Misleading error message in SSViewer

kinglozzer:
If `SSViewer.theme` is set, this error message implies that it’s searching in the theme even when `SSViewer.theme_enabled` is set to false. I’ve also made it more clear in the error message that the theme is disabled.

### DIFF
--- a/view/SSViewer.php
+++ b/view/SSViewer.php
@@ -778,12 +778,14 @@ class SSViewer implements Flushable {
 		if(!$this->chosenTemplates) {
 			$templateList = (is_array($templateList)) ? $templateList : array($templateList);
 
-			user_error(
-				"None of these templates can be found in theme '"
-				. Config::inst()->get('SSViewer', 'theme') . "': "
-				. implode(".ss, ", $templateList) . ".ss", 
-				E_USER_WARNING
-			);
+			$message = 'None of the following templates could be found';
+			if(!$theme) {
+				$message .= ' (no theme in use)';
+			} else {
+				$message .= ' in theme "' . $theme . '"';
+			}
+
+			user_error($message . ': ' . implode(".ss, ", $templateList) . ".ss", E_USER_WARNING);
 		}
 	}
 


kinglozzer:
Come to think of it, it might be more useful to set a different message if the theme is disabled? Something like:

```php
if(Config::inst()->get('SSViewer', 'theme_enabled')) {
	user_error(
		"None of these templates can be found in theme '"
		. $theme . "': "
		. implode(".ss, ", $templateList) . ".ss", 
		E_USER_WARNING
	);
} else {
	user_error(
		"None of these templates can be found (theme is disabled): "
		. implode(".ss, ", $templateList) . ".ss", 
		E_USER_WARNING
	);
}
```

dhensby:
Great bug to fix, this has caught me out many times.

dhensby:
I think your second solution is best. The message needs to be as unambiguous as possible.

Also, a unit test would be good :)

```php
$message = 'None of the following templates could be found';
if (!$theme) {
    $message .= ' (no theme in use)';
}
else {
    $message = ' (using theme "' . $theme . '")';
}
user_error($message . ': ' . impode('.ss', $templateList) . '.ss', E_USER_WARNING);
```

dhensby:
whoops, looks like I was a bit too slow with my reply ;)

I think it needs to be a bit more DRY :)

kinglozzer:
Heh, only a few seconds in it! Fixed that up now :)

view/SSViewer.php
```diff
@@ -775,15 +775,17 @@ public function __construct($templateList, TemplateParser $parser = null) {
 			);
 		}
 
-		if(!$this->chosenTemplates) {
+		if( ! $this->chosenTemplates) {
```
micmania1: TOO MANY SPACES :(

dhensby: in the wrong place :p

`if (!$this->chosenTemplates) {`

kinglozzer: Oops, completely misread our coding conventions :persevere:. Fixed now!

PR #3899, opened 2015-02-13
FIX Imported namespaces now correctly used to determine class inheritance

### LINKED ISSUE #3707: Namespaced PermissionProvider implementors are not detected
If you namespace an implementor of PermissionProvider, like so:

```php
<?php namespace SomeCompany\Permissions;

use PermissionProvider;

class CustomProvider implements PermissionProvider {
    public function providePermissions() {
        return array(
            'DESTROY_WORLD' => 'Destroy the world',
        );
    }
}
```

Changes won't be picked up, so the permission codes will not be shown when creating roles.

dhensby:
Fixes #3707

This fix allows the manifest builder to recognise Classes and Interfaces that have been used within namespaced classes or where namespaced classes have been imported with a `use` keyword.

Previously a classname or interface in the global scope explicitly needed to have a `\` prepended to the name in the `extends` or `implements` declarations for the manifest builder to recognise it.

Some examples of previously failing declarations that now pass are correctly recognised

```php
<?php
namespace SilverStripe\Test\Space;

use ModelAdmin as Admin;
use PermissionProvider, TestOnly as Test,HiddenClass;
use OtherInterface;
use \InterfaceName;

class MyExample extends Admin implements PermissionProvider, Test, HiddenClass, OtherInterface, InterfaceName {

    ...

}
```

This class would now (correctly) be picked up as a `ModelAdmin` subclass and an implementor of `PermissionProvider`;

Other declarations that were correctly interpreted remain so:

```php
<?php
namespace SilverStripe\Test\Space;

class MyClass extends \ModelAdmin {
    ...
}
```

### DIFF
--- a/core/manifest/ClassManifest.php
+++ b/core/manifest/ClassManifest.php
@@ -107,6 +107,33 @@ class SS_ClassManifest {
 	}
 
 	/**
+	 * Create a {@link TokenisedRegularExpression} that extracts the namespaces imported with the 'use' keyword
+	 *
+	 * This searches symbols for a `use` followed by 1 or more namespaces which are optionally aliased using the `as`
+	 * keyword. The relevant matching tokens are added one-by-one into an array (using `save_to` param).
+	 *
+	 * eg: use Namespace\ClassName as Alias, OtherNamespace\ClassName;
+	 *
+	 * @return TokenisedRegularExpression
+	 */
+	public static function get_imported_namespace_parser() {
+		return new TokenisedRegularExpression(array(
+			0 => T_USE,
+			1 => T_WHITESPACE,
+			2 => array(T_NS_SEPARATOR, 'save_to' => 'importString[]', 'optional' => true),
+			3 => array(T_STRING, 'save_to' => 'importString[]', 'can_jump_to' => array(2, 8)),
+			4 => array(T_WHITESPACE, 'save_to' => 'importString[]'),
+			5 => array(T_AS, 'save_to' => 'importString[]'),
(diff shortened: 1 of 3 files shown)


dhensby:
It broke all the things!

core/manifest/ClassLoader.php
```diff
@@ -101,7 +101,7 @@ public function getItemPath($class) {
 	 * @return bool
 	 */
 	public function classExists($class) {
```
willmorgan: Should this fn be renamed? (I know it would break semver but still)

dhensby: It should be, but not as part of this bug fix

dhensby: Ok, I moved this to it's own PR for better discussion #3904

core/manifest/ClassManifest.php
```diff
@@ -333,6 +360,123 @@ public function handleDir($basename, $pathname, $depth) {
...
+					//normalised class name (PHP is case insensitive for symbols/namespaces
+					$lClass = strtolower($class);
+
+					//go through all the imports and see if the class exists within one of them
+					foreach ($imports as $alias => $import) {
+						//normalise import
+						$import = trim($import, '\\');
+
+						//if there is no string key, then there was no declared alias - we'll use the main declaration
+						if (is_int($alias)) {
```
tractorcow: pyramid

tractorcow: Maybe some more continue, break, or break out some of these inner loops into functions for readability?

PR #3900, opened 2015-02-14
Behat: Wait until CMS loading overlay is removed

chillu:
This makes tests more robust at the expense of an additional
check with every step. The "ajax steps" logic works similarly,
but does not respect the loading indicator. While this logic
waits for a short time after ajax requests are finished,
in some cases that isn't enough time for the browser to
process the response and remove the loading indicator.
Since the "ajax steps" logic applies to operations outside of the CMS
as well, we can't add the loading indicator check there.

Note: It would be cleaner to create a JS API for checking the loading
state of the CMS, but I'd consider that scope creep here ;)

### DIFF
--- a/tests/behat/features/bootstrap/SilverStripe/Framework/Test/Behaviour/CmsUiContext.php
+++ b/tests/behat/features/bootstrap/SilverStripe/Framework/Test/Behaviour/CmsUiContext.php
@@ -6,6 +6,7 @@ use Behat\Behat\Context\ClosuredContextInterface,
 	Behat\Behat\Context\TranslatedContextInterface,
 	Behat\Behat\Context\BehatContext,
 	Behat\Behat\Context\Step,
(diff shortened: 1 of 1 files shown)


chillu:
@tractorcow Any idea why the branch alias isn't picked up here? Looks like travis-support creates it

tractorcow:
Are you referring to the framework appearing as 3.x-dev instead of 3.2.x-dev? I think that could be a feature of the framework repo being the one that is actually tested. That being said, is there any harm in how it's working currently? :D

chillu:
@tractorcow Re-ran tests, and they're still passing. I think the branch alias problem resolved itself in the meantime. Wanna merge?

tractorcow:
Small bugfix for cms errors at https://github.com/silverstripe/silverstripe-framework/pull/5370

PR #3901, opened 2015-02-15
Improve styling on modal dialogs

flashbackzoo:
Removed the thick black border around the dialog and close button. Added a 1px border with dropshadow and updated the close button icon.


dhensby:
Can we see some screenshots of before and after, please?

flashbackzoo:
Before:
before

After:
after

nzjoel:
Will be so good to have this change in place. This has to be my biggest niggly design issue with the CMS since ever.

dhensby:
This does look nice :)

chillu:
Looking great, thanks guys!

PR #3904, opened 2015-02-16
FIX make class loader classExists check interface_exists as per docs

dhensby:
The PHP comment says that it checks for classes and interfaces.

If that's the case, we should use the built in `interface_exists` function with the `class_exists` one for speed.

Should this method be renamed?

Should this method even use the php functions at all? Currently it doesn't depend on the manifest being used - though removing those checks would miss PHP classes/interfaces. Though the PHPDoc says "... in the manifest"

Should we have functions to check existence of classes/interfaces separately?

### DIFF
--- a/core/manifest/ClassLoader.php
+++ b/core/manifest/ClassLoader.php
@@ -95,13 +95,13 @@ class SS_ClassLoader {
 	}
 
 	/**
-	 * Returns true if a class name exists in the manifest.
+	 * Returns true if a class or interface name exists in the manifest.
 	 *
 	 * @param  string $class
 	 * @return bool
 	 */
 	public function classExists($class) {
-		return class_exists($class, false) || $this->getItemPath($class);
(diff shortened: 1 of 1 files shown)


wilr:
> Should we have functions to check existence of classes/interfaces separately?

Yep appears it's only used for ClassLoader::exists() so creating an interfaceExists() method would make the most sense.

dhensby:
@wilr if there's currently no use for them outside the internal exists() then maybe it isn't needed?

dhensby:
@wilr that phpdoc is now wrong as it does return if an interface exists... The class manifest check makes no distinction, I believe.

dhensby:
Now reverts ddec797 too

dhensby:
@wilr either this change needs to go in, or (to fix it so it only checks classes) the manifest needs a big refactor!

PR #3905, opened 2015-02-16
Update Common Problems with mod_rewrite issue

muskie9:
[Reference for issue #58 in silverstripe-installer](https://github.com/silverstripe/silverstripe-installer/issues/58#issuecomment-74509529)

### DIFF
--- a/docs/en/00_Getting_Started/01_Installation/05_Common_Problems.md
+++ b/docs/en/00_Getting_Started/01_Installation/05_Common_Problems.md
@@ -25,6 +25,20 @@ On "live" environments, the `?isDev=1` solution is preferred, as it means that y
 (and potentially security sensitive) PHP errors as well.
 </div>
 
+## mod_rewrite isn't working but it's installed (prior to SilverStripe 3.1.11)
+
+Due to some changes to `mod_dir` in [Apache 2.4](http://httpd.apache.org/docs/current/mod/mod_dir.html#DirectoryCheckHandler) (precedence of handlers), index.php gets added to the URLs as soon as you navigate to the homepage of your site. Further requests are then handled by index.php rather than `mod_rewrite` (framework/main.php). To fix this place the following within the `mod_rewrite` section of your .htaccess file:
+
+```
+<IfModule mod_rewrite.c>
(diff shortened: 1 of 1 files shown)


dhensby:
Thanks for this. I'd like to get a patch in this for 3.1.11 or 3.2 - could you update the docs to say "prior to 3.1.10"...

muskie9:
All set @dhensby

dhensby:
Thanks @muskie9 - I'm just waiting on likelihood that the fix will make 3.1.10 (I doubt it will) so this may need to say "before 3.1.11" (sorry, I know my original suggestion was 3.1.10, but looking at it it's ambiguous in it's current form if the fix isn't released in 3.1.10)

dhensby:
You can make PRs safely against the 3.0 branch which doesn't need to say anything about prior versions

muskie9:
Makes sense. I can update this once 3.1.10 stable is out and we know whether the fix made it in, unless you'd rather have it against the 3.0 branch.

dhensby:
@muskie9 It should go against all previous docs branches (2.4, 3.0 and 3.1) as it's a problem for all versions until it's fixed

muskie9:
@dhensby am I correct in thinking this would need a separate PR against 3.0 branch since the docs structure and files have changed from 3.0 to 3.1?

dhensby:
That's right

dhensby:
ok, can you update it to 3.1.11 please

dhensby:
<3

dhensby:
ARGH DANG I just noticed you said to put it *before* the `<IfModule mod_rewrite.c>` but it needs to go inside that clause...

Because you don't want to turn off the index.php interception if mod_rewrite is not enabled!!

muskie9:
I looked at your PR at some point, I should've caught that. Updated. Guessing that should get updated in 3.0 as well.

dhensby:
> Guessing that should get updated in 3.0 as well.

yes, please, sorry. I should have read better

dhensby:
Thanks a lot, really appreciated - sorry for the tedious revisions

muskie9:
No worries, wanted to make sure I got it right ;)

PR #3907, opened 2015-02-16
Update CheckboxSetField.php

phillprice:
Updated example to work in api - as DropdownField does.


dhensby:
fixed via 89aae54cf8c8557d3d330f6c20dddca5e2317dd3

dhensby:
thanks

PR #3916, opened 2015-02-18
NEW Allow the paddedresize to take another hex value to specify a transparency on the padded color

dhensby:
#3876 against branch 3

Thoughts: we handle if the % is > 100 but not if it's less than 0.

### DIFF
--- a/filesystem/GD.php
+++ b/filesystem/GD.php
@@ -417,14 +417,20 @@ class GDBackend extends Object implements Image_Backend {
 			return $useAsMinimum ? $this->resizeByWidth( $maxWidth ) : $this->resizeByHeight( $maxHeight );
 	}
 
-	public static function color_web2gd($image, $webColor) {
+	public static function color_web2gd($image, $webColor, $transparencyPercent = 0) {
 		if(substr($webColor,0,1) == "#") $webColor = substr($webColor,1);
 		$r = hexdec(substr($webColor,0,2));
 		$g = hexdec(substr($webColor,2,2));
 		$b = hexdec(substr($webColor,4,2));
 
+		if($transparencyPercent) {
+			if($transparencyPercent > 100) {
+				$transparencyPercent = 100;
+			}
+			$a = 127 * bcdiv($transparencyPercent, 100, 2);
+			return imagecolorallocatealpha($image, $r, $g, $b, $a);
+		}
 		return imagecolorallocate($image, $r, $g, $b);
-
 	}
 
 	/**
@@ -433,8 +439,11 @@ class GDBackend extends Object implements Image_Backend {
 	 * @param width
 	 * @param height
 	 * @param backgroundColour
+	 * @param transparencyPercent
 	 */
-	public function paddedResize($width, $height, $backgroundColor = "FFFFFF") {
+	public function paddedResize($width, $height, $backgroundColor = "FFFFFF", $transparencyPercent = 0) {
+		//keep the % within bounds of 0-100
+		$transparencyPercent = min(100, max(0, $transparencyPercent));
 		if(!$this->gd) return;
 		$width = round($width);
 		$height = round($height);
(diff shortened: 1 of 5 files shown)


nfauchelle:
Yea I guess we probably shouldn't bother with the >100 check. We don't do a lot of checking on other params and just assume the user is going to write the correct thing.

dhensby:
I think it's sensible to check and to be lenient with parameter input. If we can perform a simple check, we probably should. I'm just thinking if it should be extended to <0 as well.

There are 2 approaches:

1. Throw an InvalidArguementExcpetion if the param is out of range
2. Ignore or reset the argument if it's out of range

The first is stricter and could lead to failures in production, the second is kinder and more forgiving. I suppose we should do the second. Perhaps ignoring the transparency if it's <0 ?

However, the second also makes API changes later more difficult. For example if you accepted -1 as a valid param then if you wanted it to have a special meaning later, then you'd make that transition harder.

nfauchelle:
@dhensby 
Want this to be changed per your last comment, or just merged? Seems you were in two minds on change here.

I think either throw an exception on both over 100 or under 0 or just pass them through and let whatever happen happen, but I think exception would be bette.. if we were to change.

sminnee:
@dhensby do you think this is good as-is? if it's a 50/50 call I'd err on the side of getting features out to people as-is rather than blocking PRs.

nfauchelle:
@sminnee 
tbh I would consider removing it all and using something like http://image.intervention.io/
similar concept to file system changes... just get something else to handle it.

sminnee:
@nfauchelle yeah I'd tend to agree. That'd be an SS4 change.

dhensby:
I already started some work on intergrating that, but alas time is not on my side at the moment

dhensby:
I'd accept this with "Ignore or reset the argument if it's out of range" - i know this is in my branch so I need to do that work and rebase this... so it's on me to sort it out now unless @nfauchelle wants to take it back.

filesystem/GD.php
```diff
@@ -390,14 +390,20 @@ public function resizeRatio( $maxWidth, $maxHeight, $useAsMinimum = false ) {
...
 		$r = hexdec(substr($webColor,0,2));
 		$g = hexdec(substr($webColor,2,2));
 		$b = hexdec(substr($webColor,4,2));
 
+		if($transparencyPercent) {
+			if($transparencyPercent > 100) {
+				$transparencyPercent = 100;
+			}
+			$a = 127 * bcdiv($transparencyPercent, 100, 2);
+			return imagecolorallocatealpha($image, $r, $g, $b, $a);
```
tractorcow: Actually, you have your scale backwards... imagecolorallocatealpha()'s last parameter is alpha, which denotes opacity, not transparency. A value of 0 transparency should mean opacity (alpha) of 127. Should you rename your method argument to $opacityPercent and set default to 100?

You can also simplify your logic to `if($opacityPercent < 100)` ;p

filesystem/ImagickBackend.php
```diff
@@ -193,19 +193,44 @@ public function resizeByHeight($height) {
...
+	 * Convert a percentage (or 'true') to a two char hex code to signifiy the level of an alpha channel
+	 *
+	 * @param $percent
+	 * @return string
+	 */
+	public function calculateAlphaHex($percent) {
+		if($percent > 100) {
+			$percent = 100;
+		}
+		// unlike GD, this uses 255 instead of 127, and is reversed. Lower = more transparent
```
tractorcow: See my comment above... it's not reversed in GD it's backwards. :P

tractorcow:
Sheesh, I read the docs again and I was wrong. :D Sorry @dhensby you were right lol.

nfauchelle:
@dhensby sorry I am pretty busy atm. Maybe @tractorcow could as penance for his incorrect critique :-)

tractorcow:
That would be totally fair too. :D Maybe I'll find some time soon.

So we are going with the "Ignore or reset the argument if it's out of range" solution?

dhensby:
OK - I've rebased this, updated the docs and added the "reset argument if out of range" approach to dealing with the transparency %

tractorcow:
Merged finally.

dhensby:
Unfortunately this seems to have been dropped from master as it didn't merge properly. Anyone want to volunteer to PR against master?

tractorcow:
Not really!

nfauchelle:
tbh, I would consider dropping this and look at using 
http://image.intervention.io/ && https://github.com/thephpleague/glide with v4.

dhensby:
@nfauchelle yes, that's something I want to do and had kind of started on but - alas - time is lacking

PR #3917, opened 2015-02-18
Correct return statements in SS_HTTPResponse

oddnoc:
- setBody failed to return a value; it now returns $this as related methods do
- getHeader had an extra, unreachable return statement; removed


dhensby:
Thanks

PR #3920, opened 2015-02-19
Fix localStorage for Windows 8.1 IE11 desktop mode

jamesbolitho:
In IE11 windows 8.1 call to window.localStorage was throwing out an access denied error.  Using try and catch manages the issue and allows the script to execute in IE 11 in desktop mode.

### DIFF
--- a/admin/javascript/LeftAndMain.Preview.js
+++ b/admin/javascript/LeftAndMain.Preview.js
@@ -179,18 +179,22 @@
 			 * Store the preview options for this page.
 			 */
 			saveState : function(name, value) {
-				if(!window.localStorage) return;
-				
-				window.localStorage.setItem('cms-preview-state-' + name, value);
+				try {
+					window.localStorage.setItem('cms-preview-state-' + name, value);
+				} catch (exception) {
+					return false;
+				}
 			},
 
 			/**
 			 * Load previously stored preferences
 			 */
 			loadState : function(name) {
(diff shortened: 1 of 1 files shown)


dhensby:
Would a `localStorage` poly-fill not be a better long-term solution here?

dhensby:
Hmm, or is this caused by the iframe being the item that tries to write to the localStorage and so IE is being disallowing it? If so, perhaps there is a fix to this problem rather than (I assume) allowing a feature to break silently?

chillu:
I can't reproduce this issue in Browserstack on Win 8.1 with IE11 in desktop mode. But it might just be a specific security policy which isn't set by default, so we should follow best practice for accessing localstorage either way. Although a targeted Modernizr polyfill (http://modernizr.com/download/#-localstorage) wouldn't have a large impact on the codebase, I'm happy with the one off check since its only a few lines of code. @jamesbolitho Can you please document why its necessary in the JS method header, just in case that link goes dark at some point? Also, please rename to `_supportsLocalStorage()`.

dhensby:
ping @jamesbolitho

jamesbolitho:
Ah sorry didn't realise checks had failed.  Re-submitted pull request.

dhensby:
@jamesbolitho are you able to squash you commits into 1, please?

willmorgan:
Closing - squashed version available: #4560

PR #3921, opened 2015-02-21
Remove GroupedDropdownField::validate() (regression)

chillu:
Regression caused by 41ea83b,
which introduced field validation in the parent class without
also adapting this subclass.

Fixing the immediate issue (can't save this field type),
as a first step to the actual fix (implementing validate).

See silverstripe/silverstripe-translatable#179


stevie-mayhew:
Ah crap, sorry about this one. We should write a few more tests for the these fields :laughing:

tractorcow:
I've merged and added #3923 to ensure this @todo doesn't get lost in the aether.

PR #3924, opened 2015-02-22
NEW Always use default_cast

dhensby:
This is the start of some work to make `ViewableData` always use the `defaul_cast` to ensure that template escaping is secure by default.

At the moment this solution hasn't addressed the following issues specifically:

1. Setting cast of fields on object instances (eg: creating an ArrayData and setting a custom cast)
2. Dealing with setting default_cast higher up the inheritance chain

This patch effectively deprecates the `$forceObject` parameter of ViewableData::obj() as an object is now permanently returned under the `default_cast` if another casting rule is not specified.

Decision that need to be made:

0. Should this behaviour even happen? Is it desirable? Most importantly, is it any safer?
1. Should default_cast be overridden by definitions on descendant classes?
2. Should default_cast on descendent classes only override the setting for the fields/functions on THAT class? (eg: a default_cast on `MyClass` can only override a function on `MyClass` - similar to `allowed_actions`)

Todo:

- Re-write / fix broken tests (in `ViewableDataTest`)
- Fix test failures everywhere else

### DIFF
--- a/view/ViewableData.php
+++ b/view/ViewableData.php
@@ -278,7 +278,7 @@ class ViewableData extends Object implements IteratorAggregate {
 	 */
 	public function castingClass($field) {
 		$spec = $this->castingHelper($field);
-		if(!$spec) return null;
+		if(!$spec) return $this->config()->get('default_cast');
 
 		$bPos = strpos($spec,'(');
 		if($bPos === false) return $spec;
--- a/view/ViewableData.php
+++ b/view/ViewableData.php
@@ -418,8 +418,8 @@ class ViewableData extends Object implements IteratorAggregate {
 				$value = $this->$fieldName;
 			}
 
-			if(!is_object($value) && ($this->castingClass($fieldName) || $forceReturnedObject)) {
-				if(!$castConstructor = $this->castingHelper($fieldName)) {
+			if(!is_object($value) && ($this->castingClass($fieldName))) {
+				if (!$castConstructor = $this->castingHelper($fieldName)) {
 					$castConstructor = $this->config()->default_cast;
 				}
 
@@ -430,13 +430,6 @@ class ViewableData extends Object implements IteratorAggregate {
 			}
 
 			if($cache) $this->objCacheSet($cacheName, $value);
-		}
-
-		if(!is_object($value) && $forceReturnedObject) {
-			$default = $this->config()->default_cast;
-			$castedValue = new $default($fieldName);
-			$castedValue->setValue($value);
-			$value = $castedValue;
 		}
 
 		return $value;
--- a/forms/Form.php
+++ b/forms/Form.php
@@ -206,6 +206,15 @@ class Form extends RequestHandler {
 		'forTemplate',
 	);
 
+	private static $casting = array(
+		'AttributesHTML' => 'HTMLText',
+		'FormAttributes' => 'HTMLText',
+		'MessageType' => 'Text',
+		'Message' => 'HTMLText',
+		'FormName' => 'Text',
+		'Legend' => 'HTMLText',
+	);
+
 	/**
 	 * @var FormTemplateHelper
 	 */
--- a/forms/FormField.php
+++ b/forms/FormField.php
@@ -255,6 +255,22 @@ class FormField extends RequestHandler {
 	 */
 	protected $schemaData = [];
 
+	private static $casting = array(
+		'FieldHolder' => 'HTMLText',
+		'Field' => 'HTMLText',
+		'AttributesHTML' => 'HTMLText',
+		'Value' => 'HTMLText',
+		'extraClass' => 'Text',
+		'ID' => 'Text',
+		'isReadOnly' => 'Boolean',
+		'HolderID' => 'Text',
+		'Title' => 'HTMLText',
+		'RightTitle' => 'HTMLText',
+		'MessageType' => 'Text',
+		'Message' => 'HTMLText',
+		'Description' => 'HTMLText',
+	);
+
 	/**
 	 * Structured schema state representing the FormField's current data and validation.
 	 * Used to render the FormField as a ReactJS Component on the front-end.
--- a/model/FieldType/DBField.php
+++ b/model/FieldType/DBField.php
@@ -72,6 +72,7 @@ abstract class DBField extends ViewableData {
 
 	private static $casting = array(
 		'ATT' => 'HTMLText',
+		'XML' => 'HTMLText',
 	);
 
 	/**
--- a/model/FieldType/DBField.php
+++ b/model/FieldType/DBField.php
@@ -68,6 +68,12 @@ abstract class DBField extends ViewableData {
 	 */
 	private static $default_search_filter_class = 'PartialMatchFilter';
 
+	//private static $default_cast = 'HTMLText';
+
+	private static $casting = array(
+		'ATT' => 'Text',
+	);
+
 	/**
 	 * @var $default mixed Default-value in the database.
 	 * Might be overridden on DataObject-level, but still useful for setting defaults on
--- a/model/FieldType/DBField.php
+++ b/model/FieldType/DBField.php
@@ -71,7 +71,7 @@ abstract class DBField extends ViewableData {
 	//private static $default_cast = 'HTMLText';
 
 	private static $casting = array(
-		'ATT' => 'Text',
+		'ATT' => 'HTMLText',
 	);
 
 	/**
--- a/model/FieldType/DBField.php
+++ b/model/FieldType/DBField.php
@@ -68,10 +68,16 @@ abstract class DBField extends ViewableData {
 	 */
 	private static $default_search_filter_class = 'PartialMatchFilter';
 
-	//private static $default_cast = 'HTMLText';
-
 	private static $casting = array(
 		'ATT' => 'HTMLText',
+		'XML' => 'HTMLText',
+		'HTMLATT' => 'HTMLText',
+		'URLATT' => 'HTMLText',
+		'RAWURLATT' => 'HTMLText',
+		'ATT' => 'HTMLText',
+		'RAW' => 'HTMLText',
+		'JS' => 'HTMLText',
+		'HTML' => 'HTMLText',
 		'XML' => 'HTMLText',
 	);
 
--- a/model/HTMLValue.php
+++ b/model/HTMLValue.php
@@ -163,6 +163,7 @@ class SS_HTML4Value extends SS_HTMLValue {
 		// Ensure that \r (carriage return) characters don't get replaced with "&#13;" entity by DOMDocument
 		// This behaviour is apparently XML spec, but we don't want this because it messes up the HTML
 		$content = str_replace(chr(13), '', $content);
+		$content = str_replace('&', 'SS_ENTITY_AMERSAND', $content);
 
 		// Reset the document if we're in an invalid state for some reason
 		if (!$this->isValid()) $this->setDocument(null);
@@ -175,5 +176,9 @@ class SS_HTML4Value extends SS_HTMLValue {
 		libxml_clear_errors();
 		libxml_use_internal_errors($errorState);
 		return $result;
+	}
+
+	public function getContent() {
+		return str_replace('SS_ENTITY_AMERSAND', '&', parent::getContent());
 	}
 }
--- a/tests/view/ViewableDataTest.php
+++ b/tests/view/ViewableDataTest.php
@@ -8,14 +8,42 @@
  */
 class ViewableDataTest extends SapphireTest {
 
+	public function testCasting() {
+		$htmlString = "&quot;";
+		$textString = '"';
+
+		$htmlField = DBField::create_field('HTMLText', $textString);
+
+		$this->assertEquals($textString, $htmlField->forTemplate());
+		$this->assertEquals($htmlString, $htmlField->obj('HTMLATT')->forTemplate());
+		$this->assertEquals('%22', $htmlField->obj('URLATT')->forTemplate());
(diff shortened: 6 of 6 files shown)


sminnee:
In general, I think this is the right way to go, and frankly is probably a bugfix to the the attempts we (I?) made in 3.1 to make it easier to avoid XSS issues. This is the complete solution.

I'm a little worried about the performance impact of this change, but that's nothing a profile can't check. We should also look to see how many HTMLText objects get instantiated: if it's a large number of them (100+ on a complex page) we may want to look at including an optimisation path where HTMLText values aren't instantiated. Of course, that's only applicable when shortcodes are disabled so it might end up making the code confusing and ugly.

Note that HTMLText by default includes shortcode parsing. All of those elements you've marked at HTMLText are now calling the shortcode parser. I'm not sure that's desirable. The return value of SSViewer::process() shows how you can disable this, but that kind of approach can't be easily wedged into the casting syntax. The casting syntax does use `Object::create_from_string` so it may be possible to set default_cast to `"HTMLText(array('shortcodes' => false))"`. create_from_string will cache the parsing of that string so it shouldn't be too inefficient. `ViewableData::escapeTypeForField()` may need to be fixed if we do this, it looks like it assumes the casting value is a simple classname. It's not called in very many places so any small additional call (e.g. an object instantiation) isn't too big a deal.

As for the comments about overriding default_cast: I would't worry too much about its overriding semantics. It's not something that most developers should do, and trying to make it smart will probably slow down the runtime, which is not a good trade-off. I would say that if you override default_cast, it overrides it for all methods, we advise against it unless you really know what you're doing (e.g. if you're rendering non-HTML templates or something). Frankly, it's a little strange that it's even able to be modified. ;-)

dhensby:
> In general, I think this is the right way to go, and frankly is probably a bugfix to the the attempts we (I?) made in 3.1 to make it easier to avoid XSS issues. This is the complete solution.

Yes, I think that's right, this is a more robust approach

> I'm a little worried about the performance impact of this change

yes, me too.

Perhaps if this is a 4.0 change the approach isn't radical enough? Perhaps we need some kind of rendering helper class for objects that can be used to store the meta data on the values and how they should be rendered. Splitting escaping away from `DBField` and `ViewableData`? Though this may be a big piece of work to bite off and would really need an RFC process and a lot of thought.

> Note that HTMLText by default includes shortcode parsing. All of those elements you've marked at HTMLText are now calling the shortcode parser. I'm not sure that's desirable. ... The casting syntax does use Object::create_from_string so it may be possible to set default_cast to "HTMLText(array('shortcodes' => false))".

Yes, this is a problem - if we could cast it as off in this way, then that would be great.

> As for the comments about overriding default_cast: I would't worry too much about its overriding semantics

Good point RE non HTML rendering, that's a pretty solid case for when you'd want to be able to override it for all parent classes too.

I guess it depends whether the security impact of ensuring core data types remain under a "default_cast" is worth the performance impact. I suppose the way around this being a problem is to always explicitly cast and then the default cast is only for the un-cast fields that a developer is adding in their own code.

sminnee:
@tractorcow @hafriedlander @dhensby I think that we need to get this one ready for master as a priority, in order to better guard against XSS issues.

Re: "Perhaps if this is a 4.0 change the approach isn't radical enough?" I think that it's better to do something now, then not do anything while waiting for something better. So I'd suggest we proceed with this change, and if we proceed with an RFC, well that's nice too. This is not a massively invasive change, it applies the same general principles that SSViewer had always had, it's more just a bit-too-invasive-for-a-minor-release, because it will fix "bugs" that developers will likely have relied on. ;-)

tractorcow:
@dhensby @sminnee apologies for not giving this PR much attention lately, but the truth is that I'm extremely keen to get this reviewed and merged into 4.0. I haven't seen much other feedback from the other developers with regards to this feature, so I'll see if I can drum up some more support to get this reviewed from other developers. :)

Specific things:

- Why not implement it at castingHelper? Shouldn't we use default-cast here?
- I'm not the fan of `if($this->hasMethod('db'))` in code. Shall we override this method in DataObject instead?

view/ViewableData.php
```diff
@@ -358,44 +358,37 @@ public function renderWith($template, $customFields = null) {
...
 			if($methodIsAllowed && $this->hasMethod($fieldName)) {
 				$value = $arguments ? call_user_func_array(array($this, $fieldName), $arguments) : $this->$fieldName();
 			} else {
 				$value = $this->$fieldName;
 			}
-			
-			if(!is_object($value) && ($this->castingClass($fieldName) || $forceReturnedObject)) {
-				if(!$castConstructor = $this->castingHelper($fieldName)) {
+
+			if(!is_object($value) && ($this->castingClass($fieldName))) {
```
tractorcow: Now castingClass and castingHelper always return a value, so it's no longer necessary to check for this.

view/ViewableData.php
```diff
@@ -358,44 +358,37 @@ public function renderWith($template, $customFields = null) {
 	 */
 	public function obj($fieldName, $arguments = null, $forceReturnedObject = true, $cache = false, $cacheName = null) {
 		if(!$cacheName) $cacheName = $arguments ? $fieldName . implode(',', $arguments) : $fieldName;
-		
+
 		if(!isset($this->objCache[$cacheName])) {
 			// HACK: Don't call the deprecated FormField::Name() method
```
tractorcow: I think we can finally remove this, as FormField::Name is gone. now `$methodIsAllowed` is always true.

tractorcow:
Just to add some support and reasoning to this change,

Having default casting applied to all templates is important in the same way that having parameterised queries is important to the ORM. By defining a single and mandatory place where any value (in this case, text) is forcibly passed through, we can guarantee protection at the template level.

Under the old code, it would be otherwise necessary to use DBField::create_field in order to force a casting. While this still works with this code, I feel this is bad style in the same way that mandating `Convert::raw2sql` promotes bad coding style:

- It promotes 'weakest link' coding; It requires that the style is adhered to in every instance, and one missing cast (such as in a string manipulation) can render the whole chain insecure.
- It requires a coding convention that is awkward un unweildly to developers, which is the biggest red flag. If you have a method whch takes the result of another method and (for instance) wraps the value in a `<p></p>`, then the difference is between these two styles:

```php
class MyClass extends ViewableData {
    public function getData() {
        return DBField::create_field('HTMLText', $this->getSomeHtml());
     }
    public function getFormattedData() {
        $value = $this->getData();
        return DBField::create_field('HTMLText', '<p>' . $value->getValue() . '</p>');
    }
}
```

and

```php
class MyClass extends ViewableData {
    private static $casting = array(
        'Data' => 'HTMLText',
        'FormattedData' => 'HTMLText'
    );
    public function getData() {
         $this->getSomeHtml();
     }
    public function getFormattedData() {
        return '<p>' . $this->getData() . '</p>';
    }
}
```

In the above example, a fatal mistake would be to forget the DBField::create on the second method. The consequence of this is a severe XSS vulnerability.

In the second example, the worst mistake that could be made is forgetting to add the casting. The consequence of this is that the text would be double encoded. This is NOT a vulnerability, but is instead a simple functionality degradation.

tractorcow:
Also $forceReturnedObject is redundant. Let's remove it? See comment at https://github.com/silverstripe/silverstripe-framework/pull/4628

sminnee:
Shall we get this merged in, guys? @dhensby looks like it needs a rebase.

dhensby:
This, unfortunately, is a living PR and is not complete and had significant impacts throughout core which I need to update...

I'll try to look at it, hack day maybe...

tractorcow:
You didn't need TOO many `cast` configs... which was my concern.

I'd look at this at alpha2. Let's put this in the backlog, maybe, @sminnee to look at?

sminnee:
It would make for a great hackday project. :-)

tractorcow:
@dhensby can you give me a bit of background on https://github.com/silverstripe/silverstripe-framework/pull/3924/commits/726b275289e7860e79e1e7462ba20f120d77bb67?

dhensby:
Oh gawd - I can't remember the specifics of that now... I think it was a WIP hence the commit called "stash".

Basically there is some issue with `DOMDocument` incorrectly encoding `&` so it had to be pulled out.

Here's my search history from the time:

tractorcow:
Oh, is this the problem I had in 4.x earlier when updating tinymce?
I fixed it by adding the below code to SS_HTMLValue::getContent()


```php
// Prevent &nbsp; being encoded as literal utf-8 characters
// Possible alternative solution: http://stackoverflow.com/questions/2142120/php-encoding-with-domdocument
$from = mb_convert_encoding('&nbsp;', 'utf-8', 'html-entities');
$res = str_replace($from, '&nbsp;', $res);
```

dhensby:
I think I had the problem with & that shouldn't be encoded... I can't remember the very specifics. Maybe running the test suite will make it more obvious?

tractorcow:
Not really passing :) I'm testing this locally.

tractorcow:
I've rebased this locally, and did a bit of code review.

@dhensby what do you think about my suggestions at https://github.com/tractorcow/silverstripe-framework/commits/pulls/default-cast?

I've ditched $forceReturnedObject parameter, and removed any code that had to handle a non-object return type.

I'm not sure about the `XML_val` => `Template_val`, but it seems better to make it a bit more explicit? I'm looking at some of the string classes and XML !== forTemplate() in a few cases. I'm going to look at how things like shortcodes and newlines are going to affect this. :D

tractorcow:
I think I'm going to revert that part of my change... ;D

tractorcow:
> Note that HTMLText by default includes shortcode parsing. All of those elements you've marked at HTMLText are now calling the shortcode parser. I'm not sure that's desirable. 

This is a hard problem, but it might be safer to turn off shortcode passing by default, and turn it on on a case by case basis.

sminnee:
We probably want some way of defining a type "HTML text with shortcodes" and "HTML text without shortcodes"

It could be a constructor property. Shortcodes should only be added by default on database-backed HTMLtext fields intended to have WYSIWYG editors attached to them.

We may find that shortcode parsing is only one such processor, and so an array $processors argument might be better than a Boolean $hasShortcodes one.

sminnee:
We can update the service definition of "HTMLText" to include the shortcode processing - then DB field definitions will remain as-is :-)

tractorcow:
It won't really work if we are relying on existing `casting => 'HTMLText'` to NOT include shortcode parsing.

tractorcow:
I propose turning shortcodes off by default, but then making it really easy to turn on.

At the moment HTMLText has a few options, and I've toyed with a few mechanisms for turning it back on.

*  - `HTMLText(['shortcodes=true', 'whitelist=meta,link'])`
*  - `HTMLText('whitelist=meta,link')`
*  - `HTMLText(['shortcodes=true'])`
*  - `HTMLText('shortcodes=true')`

tractorcow:
Ok, so this is what my Page.php looks like


```php
	private static $db = array(
		"URLSegment" => "Varchar(255)",
		"Title" => "Varchar(255)",
		"MenuTitle" => "Varchar(100)",
		"Content" => "HTMLText(['shortcodes=true'])",
		"MetaDescription" => "Text",
		"ExtraMeta" => "HTMLText(['whitelist' => 'meta,link'])",
		"ShowInMenus" => "Boolean",
		"ShowInSearch" => "Boolean",
		"Sort" => "Int",
		"HasBrokenFile" => "Boolean",
		"HasBrokenLink" => "Boolean",
		"ReportClass" => "Varchar",
		"CanViewType" => "Enum('Anyone, LoggedInUsers, OnlyTheseUsers, Inherit', 'Inherit')",
		"CanEditType" => "Enum('LoggedInUsers, OnlyTheseUsers, Inherit', 'Inherit')",
	);
```

sminnee:
It's important that `"Content" => "HTMLText"` or something equally simple still contains shortcodes. Otherwise everyone will ask why links/files have broken.

We could create 2 service definitions:

 * `HTMLText`: HTML fields in the WYSIWYG editor. DBHTMLText with shortcodes
 * `HTMLTemplateFragment`: The type used to render template fragments:  DBHTMLText without shortcodes

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/commit/9f399118db5f5c8acd2b37b02170e144c404090e for proof of concept.

If you like, we can add a bit of code to `DataObject::castingHelper` that converts `HTMLText` to `HTMLText(['shortcodes=true'])` as a bacwards compatibility fix. That will work for `$db` but not affect `$casting`

tractorcow:
Or how about `ShortcodeHTML` specified via Injector?

tractorcow:
We may as well not have shortcodes in HTMLText at all, and simply allow post-processer to be injected as a dependency.

sminnee:
> My reasoning is that 90% of the time where HTMLText is used, the code does not expect shortcodes to be parsed. E.g. SSViewer::render().

This isn't true. 99.9% of the places that HTMLText are used are in WYSIWYG fields on page types. The important measure is how often these things are used across all modules and sites in the ecosystem, not the lines of code in framework.

sminnee:
`SSViewer::render()` should have its own service definition; I like "HTMLTemplateFragment" since that's what is being produced. "SSViewerTemplateFragment" might also be good.

    Injector:
      HTMLText:
        class: SilverStripe\Model\FieldType\DBHTMLText
            properties:
              ProcessShortcodes: true
      HTMLTemplateFragment:
        class: SilverStripe\Model\FieldType\DBHTMLText

tractorcow:
Ok, that's what we'll do. Sounds fine with me. :)

We'll go with `HTMLFragment` as discussed earlier, because I love concise titles.

tractorcow:
OK, I've cleaned this up and re-raised over at https://github.com/silverstripe/silverstripe-framework/pull/5653.

PR #3926, opened 2015-02-23
Cleaning up Controller::handleRequest

dhensby:
This PR reduced the amount of responsiblity taken on by `Controller::handleRequest`. In turn this reduces the code complexity as well.

The handleRequest function has been split into component parts (setup and teardown - the functions that prepare and cleanup the controller before and after the request is _actually_ handled).

The init function has had a similar treatment. The init function is now split into the logic that should actually be happening when init is run and the logic preparing to run the init and enforcing that base init has been run at all.

init is still defined on controllers and executed as expected but instantiators of controllers should be calling `doInit` from now on when preparing their controllers.

Further, I've removed all direct manipulations of class properties and used their accessors where appropriate (and created them where missing).

---

1. Separated responsibility of handleAction so that it no longer bootstraps the controller and cleans up after the request is handled.
2. NEW beforeHandleRequest to take responsibility of bootstrapping the controller
3. NEW afterHandleRequest to take responsibility of cleanup for the controller
4. NEW calling init on controllers removed in favour of doInit() which takes responsibility of enforcing that "base init" is called and the before and after hooks. init() is now protected visibility
5. NEW Added prepareResponse to Controller for dealing with responses from controllers
6. NEW setResponse added to controller for setting response objects on the controller

### DIFF
--- a/control/Controller.php
+++ b/control/Controller.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Controllers are the cornerstone of all site functionality in SilverStripe. The {@link Director}
  * selects a controller to pass control to, and then calls {@link run()}. This method will execute
@@ -86,11 +87,36 @@ class Controller extends RequestHandler implements TemplateGlobalProvider {
 	 *
 	 * @uses BasicAuth::requireLogin()
 	 */
-	public function init() {
+	protected function init() {
 		if($this->basicAuthEnabled) BasicAuth::protect_site_if_necessary();
 
 		// This is used to test that subordinate controllers are actually calling parent::init() - a common bug
 		$this->baseInitCalled = true;
+	}
+
+	/**
+	 * A stand in function to protect the init function from failing to be called as well as providing before and
+	 * after hooks for the init function itself
+	 *
+	 * This should be called on all controllers before handling requests
+	 */
+	public function doInit() {
+		//extension hook
+		$this->extend('onBeforeInit');
+
+		// Safety call
+		$this->baseInitCalled = false;
+		$this->init();
+		if (!$this->baseInitCalled) {
+			user_error(
+				"init() method on class '$this->class' doesn't call Controller::init()."
+				. "Make sure that you have parent::init() included.",
+				E_USER_WARNING
+			);
+		}
+
+		$this->extend('onAfterInit');
+
 	}
 
 	/**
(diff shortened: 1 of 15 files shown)


control/RequestHandler.php
```diff
@@ -495,5 +495,6 @@ public function getRequest() {
 	 */
```
tractorcow: `@return $this`

control/Controller.php
```diff
@@ -215,6 +312,14 @@ public function getResponse() {
 		return $this->response;
 	}
 
+	/**
+	 * @param SS_HTTPResponse $response
+	 */
```
tractorcow: `@return $this`

control/Controller.php
```diff
@@ -196,8 +289,12 @@ protected function handleAction($request, $action) {
 		}
 	}
 
+	/**
+	 * @param array $urlParams
```
tractorcow: `@return $this`

control/Controller.php
```diff
@@ -83,30 +145,59 @@ public function Link() {
 	}
 
 	/**
+	 * {@inheritdoc}
+	 *
+	 * Also set the URLParams
```
tractorcow: `{@inheritdoc}` is probably not used properly here.

Please include `@param` and `@return` phpdoc.

control/Controller.php
```diff
@@ -76,6 +83,61 @@ public function init() {
...
+	}
+
+	/**
+	 * A stand in function to protect the init function from failing to be called as well as providing before and
+	 * after hooks for the init function itself
+	 *
+	 * This should be called on all controllers before handling requests
+	 */
+	public function callInit() {
+
```
tractorcow: No blank lines after the opening braces please. :D

control/Controller.php
```diff
@@ -69,6 +69,13 @@ class Controller extends RequestHandler implements TemplateGlobalProvider {
 	 * @uses BasicAuth::requireLogin()
 	 */
 	public function init() {
```
tractorcow: How about making the method protected? :D

You can put the deprecation notice inside __call

tractorcow:
Like the PR... one functional suggestion given. :D

control/Controller.php
assertchris: If the intention is to add detail to the inherited phpdoc then it is used correctly.

control/Controller.php
```diff
@@ -76,6 +83,61 @@ public function init() {
...
+		return $this->callInitRunning;
+	}
+
+	/**
+	 * A stand in function to protect the init function from failing to be called as well as providing before and
+	 * after hooks for the init function itself
+	 *
+	 * This should be called on all controllers before handling requests
+	 */
+	public function callInit() {
```
tractorcow: I would have gone with `doInit` to be consistent with the other `doMethods`, but that's just a point of artistic difference. :D

control/Controller.php
dhensby: But that is slow and nasty :(

I don't think we should start filling up __call with all our deprecation notices...

control/Controller.php
dhensby: Good idea, I prefer that

control/Controller.php
dhensby: <3

dhensby:
Thinking about it, if this is going into master, then we don't need deprecation warnings... it's just pointless extra code; either this should go into 3.2 with deprecations or it should go in as a refactor to master..

What do you think @tractorcow?

dhensby:
OK, addressed most comments. Need thoughts on if this is 4.0 or 3.x

If it's 3.x then I'd be open to moving deprecation into `__call` otherwise I'd opt to just remove deprecation notices and put it in the upgrade notes...

control/Controller.php
tractorcow: Then just make it protected and NOT have the deprecation notice. ;)

tractorcow:
> If it's 3.x then I'd be open to moving deprecation into __call otherwise I'd opt to just remove deprecation notices and put it in the upgrade notes...

Sounds good. :D

dhensby:
Ok. Removed the depreciation handling which simplified things. 

We can still add a depreciation warning to 3.x that simply renames init to doInit to ease upgrade pain 

I just need to update the docs a bit more

dhensby:
Oh, and to add PRs to our other modules :P

tractorcow:
Sorry I forgot about this PR @dhensby ;_;

dhensby:
rebased

dhensby:
ping @tractorcow

assertchris:
I'm not a huge fan of init → doInit. Would prefer something like init → onInit.

Given the nature of this change (and especially since it is breaking), perhaps it would be good to get feedback from an additional developer? Perhaps @sminnee or @chillu, if they have some time?

dhensby:
It's only had 11 months for feedback ;)

The name of the function I think is neither here-nor-there. I can change it, but it is effort for a fairly subjective issue.

I do disagree with it being `onInit`, though. `on*` functions are usually event based (`onBeforeWrite`, `onAfterDelete`) where as this is the trigger so we are saying "do this thing" - ideally `init` should still be used but I chose not to so that the upgrade path was going to be far more simple this way (modules can keep their init functions and not needlessly rename them).

assertchris:
@dhensby 11 months is a long time, but it is still a breaking change and @tractorcow is the only developer to have given significant feedback on it yet. I made the `on*` comment because it was linked to from a thread, where @sminnee mentioned that as a preferred pattern. In that thread @tractorcow also uses `on*` as an example. It is subjective, sure.

> modules can keep their init functions and not needlessly rename them

Unless I'm mistaken, modules cannot keep their `init` methods because this PR changes the function visibility. Wouldn't they still have to change their `init` methods to `doInit`?

dhensby:
>> so any overloaded instances of this method need to have their visibility changed
> [Edited to replace some gubbins about init vs. doInit with a quote from the docs included in this PR]

I think it turns out that you can increase visibility without PHP throwing a shit-storm but not decrease, so that's actually bad docs by me... but people *should* change their visibility in maintained modules

dhensby:
Tests fail because the CMS test case is not pulling in these changes https://github.com/silverstripe/silverstripe-cms/pull/1169

control/Controller.php
```diff
@@ -103,88 +129,129 @@ public function Link() {
...
 	 *
 	 * @param SS_HTTPRequest $request
 	 * @param DataModel $model
 	 *
 	 * @return SS_HTTPResponse
 	 */
 	public function handleRequest(SS_HTTPRequest $request, DataModel $model) {
-		if(!$request) user_error("Controller::handleRequest() not passed a request!", E_USER_ERROR);
+		if (!$request) {
+			user_error("Controller::handleRequest() not passed a request!", E_USER_ERROR);
```
tractorcow: I prefer InvalidArgumentException, but sure.

control/Controller.php
```diff
@@ -103,88 +129,129 @@ public function Link() {
...
-					. "returning it without modification.");
+	/**
+	 * Prepare the response (we can receive an assortment of response types (strings/objects/HTTPResponses) and
+	 * changes the controller response object appropriately
+	 *
+	 * @param $response
+	 */
+	protected function prepareResponse($response) {
+		if ($response instanceof SS_HTTPResponse) {
+			if (isset($_REQUEST['debug_request'])) {
```
tractorcow: TODO: Remove this debugging code.

PR #3132, opened 2014-05-13
API: Update Requirements to use Injector and Config

kinglozzer:
- Rename `Requirements_Backend` to `RequirementsHandler` - `Requirements_Backend` is now an interface for the required methods
- Switch to the `Config` API for setting various static properties
- Move `Requirements_Backend->combine_js_with_jsmin` and `write_header_comment` to `Config`, use `__set()` to deprecate them
- Update names of static methods to be more consistent with coding conventions (existing template syntax preserved with `TemplateGlobalProvider`)
- Update unit tests and existing uses of `Requirements` in core

All comments are welcome :)

PR #3928, opened 2015-02-23
YML Config Not Compatible with Requirements Class

johndalangin:
After hours of trying to debug why `write_js_to_bottom` wasn't working, I realized that the Requirements class variables were set to `protected` and not `private static` which I believe is necessary for the Config YML variables to be passed correctly to the class.

The documentation was edited to reflect the working, albeit traditional non-YML method of setting class variables.

Best regards,

John

### DIFF
--- a/docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
+++ b/docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
@@ -176,15 +176,17 @@ careful when messing with the order of requirements.
 By default, SilverStripe includes all Javascript files at the bottom of the page body, unless there's another script 
 already loaded, then, it's inserted before the first `<script>` tag. If this causes problems, it can be configured.
 
-**mysite/_config/app.yml**
-
-	:::yml
-	Requirements:
-	  write_js_to_body: true
-	  force_js_to_bottom: true
+	:::php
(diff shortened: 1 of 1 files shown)


dhensby:
I think the correct fix here is not to change the docs to recommend the old style, but instead to change the statics to the new style? 

@kinglozzer weren't you doing some work on requirements?

kinglozzer:
Yep, in #3132 (I need to do some work on that!). Given that work would be a minor release at the earliest, I think for now it makes sense to update the docs :)

johndalangin:
Thanks for acknowledging!

I also think it would be best to edit the docs for the time being so that others can avoid having to troubleshoot needlessly.

camfindlay:
I think updating docs at this stage would be appropriate. @johndalangin can you perhaps review that section of the docs and ensure it is clear the user needs to use the _config.php file and not yaml at this stage?

dhensby:
Ok, I'll take this in and @kinglozzer you'll need to remember to correct the docs BACK (you can do it as part of your PR) once your PR is ready

johndalangin:
@camfindlay Sure, I'll edit the other sections as well to refer to `_config.php` for the time being until a patch is available.

Thanks!

PR #3930, opened 2015-02-24
Improved getIP() detection

axllent:
Ignore invalid `HTTP_CLIENT_IP` or `HTTP_X_FORWARDED_FOR` headers, as well as reserved internal IP addresses (eg: internal networks with squid proxy).

### DIFF
--- a/control/HTTPRequest.php
+++ b/control/HTTPRequest.php
@@ -655,15 +655,20 @@ class SS_HTTPRequest implements ArrayAccess {
 	 * @return string
 	 */
 	public function getIP() {
+		$ip = false;
 		if (!empty($_SERVER['HTTP_CLIENT_IP'])) {
 			//check ip from share internet
-			return $_SERVER['HTTP_CLIENT_IP'];
+			$ip = $_SERVER['HTTP_CLIENT_IP'];
 		} elseif (!empty($_SERVER['HTTP_X_FORWARDED_FOR'])) {
 			//to check ip is pass from proxy
-			return  $_SERVER['HTTP_X_FORWARDED_FOR'];
-		} elseif(isset($_SERVER['REMOTE_ADDR'])) {
-			return $_SERVER['REMOTE_ADDR'];
+			$ip = $_SERVER['HTTP_X_FORWARDED_FOR'];
 		}
+		if ((!$ip || !filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE))
+			&& !empty($_SERVER['REMOTE_ADDR'])) {
+			//if no other forwarding ip is found, invalid, or internal ip address
+			$ip = $_SERVER['REMOTE_ADDR'];
+		}
+		return $ip;
 	}
 
 	/**


dhensby:
@axllent I like this solution, though is it borne from a real-world example where you're getting bad IPs returned instead of the correct one?

This should also have some unit-tests IMO.

control/HTTPRequest.php
```diff
@@ -655,15 +655,19 @@ public function allParsed() {
...
 		} elseif (!empty($_SERVER['HTTP_X_FORWARDED_FOR'])) {
 			//to check ip is pass from proxy
-			return  $_SERVER['HTTP_X_FORWARDED_FOR'];
-		} elseif(isset($_SERVER['REMOTE_ADDR'])) {
-			return $_SERVER['REMOTE_ADDR'];
+			$ip = $_SERVER['HTTP_X_FORWARDED_FOR'];
 		}
+		if(!$ip || !filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE)) {
+			//if no other forwarding ip is found, invalid, or internal ip address
+			$ip = $_SERVER['REMOTE_ADDR'];
```
kinglozzer: This was previously checked with `isset($_SERVER['REMOTE_ADDR'])`, can you guarantee it’s always present?

dhensby: Good point. It'll probably be missing on CLI requests?!

axllent: @dhensby Yes, for instance I have a transparent squid proxy inside my own network I use for testing sometimes. This automatically adds `HTTP_X_FORWARDED_FOR` with my internal IP, and the current getIP() returns that rather than my external one. The other point to mention is that `HTTP_X_FORWARDED_FOR` is easily spoofed, and thus could be set to anything (other than an IP address as far as I know). This simply resorts back to the `REMOTE_ADDR` if it doesn't validate as an external IP address.

@kinglozzer That's a very good point, and yes, I overlooked that!

I'll modify my patch ...

dhensby:
Please squash the commits :) thanks

axllent:
I have literally no idea how to squash those commits without creating a complete new pull request (due to my limited knowledge advances merging etc with git). I have (I think) squashed them successfully & [committed those changes](https://github.com/axllent/silverstripe-framework/commit/cccb6b214c0ff8a79267971cd345ea9e18bd71a1), but it doesn't seem to have any baring on this PR. Is it easier if you simply reject this one and I submit another separate PR @dhensby ?

dhensby:
Easiest way to squash would be:

```bash
$ git reset --HARD b62f09475cd7b17db89ffd3c5e67672fe1c7a438
$ git reset HEAD~2
$ git commit -a --amend --no-edit
$ git push -f

dhensby:
I amended the last comment as I got the commit hash wrong

axllent:
Thanks for the help. Before I do that and we work in circles, was my last attempt not what you were looking for? (I edited the last post referring to b62f094)

dhensby:
Those commands do the following: 

Put your code back to how it was when I asked you to squash (you've added a merge commit and we want rid of that)

Go back to your first commit but keep the fuel changes on disk

Commit all the changes and amend the first of your commits with all the subsequent work and don't ask for a new commit message 

Force the changes into your PR

dhensby:
Yes, please still do it. Your last attempt has just added 2 new commits ;)

dhensby:
What a beaut. Worked like a charm it seems. Thanks.

axllent:
No problem, thanks for the steps, no issues there :) I'll get this all one day! One patch now, yay!

dhensby:
OK, taking this in. thanks

PR #2736, opened 2013-12-20
Change the LRU cache to a simple in-memory mode

simonwelsh:
This dramatically decreases the complexity and overhead of the Config_LRU cache with only a slight increase in memory usage when the cache hits about 2000 items.

As an aside, Config_LRU should probably get renamed now that it's not a LRU cache.

PR #3934, opened 2015-02-24
Change the LRU cache to a simple in-memory model

dhensby:
Slight memory increase when there are lots of additions for much better performance in every case.

---

A re-issue of #2736

### DIFF
--- a/core/Config.php
+++ b/core/Config.php
@@ -250,7 +250,7 @@ class Config {
 	 * leak through to other instances.
 	 */
 	public function __construct() {
-		$this->cache = new Config_LRU();
+		$this->cache = new Config_MemCache();
 	}
 
 	public function __clone() {
@@ -681,6 +681,7 @@ class Config {
 /**
  * @package framework
  * @subpackage core
+ * @deprecated 3.2
  */
 class Config_LRU {
 	const SIZE = 1000;
@@ -692,6 +693,7 @@ class Config_LRU {
 	protected $c = 0;
 
 	public function __construct() {
+		Deprecation::notice('3.2', 'Please use Config_MemCache instead', Deprecation::SCOPE_CLASS);
 		if (version_compare(PHP_VERSION, '5.3.7', '<')) {
 			// SplFixedArray causes seg faults before PHP 5.3.7
 			$this->cache = array();
@@ -783,6 +785,69 @@ class Config_LRU {
 		else {
 			for ($i = 0; $i < self::SIZE; $i++) $this->cache[$i]->key = null;
 			$this->indexing = array();
+		}
+	}
+}
+
+/**
+ * @package framework
+ * @subpackage core
+ */
+class Config_MemCache {
+	protected $cache;
+
+	protected $i = 0;
+	protected $c = 0;
+	protected $tags = array();
+
+	public function __construct() {
+		$this->cache = array();
+	}
+
+	public function set($key, $val, $tags = array()) {
+		foreach($tags as $t) {
+			if(!isset($this->tags[$t])) {
+				$this->tags[$t] = array();
+			}
+			$this->tags[$t][$key] = true;
+		}
+
+		$this->cache[$key] = array($val, $tags);
+	}
+
+	private $hit = 0;
+	private $miss = 0;
+
+	public function stats() {
+		return $this->miss ? ($this->hit / $this->miss) : 0;
+	}
+
+	public function get($key) {
+		if(isset($this->cache[$key])) {
+			++$this->hit;
+			return $this->cache[$key][0];
+		}
+
+		++$this->miss;
+		return false;
+	}
+
+	public function clean($tag = null) {
+		if($tag) {
+			if(isset($this->tags[$tag])) {
+				foreach($this->tags[$tag] as $k => $dud) {
+					// Remove the key from everywhere else it is tagged
+					$ts = $this->cache[$k][1];
+					foreach($ts as $t) {
+						unset($this->tags[$t][$k]);
+					}
+					unset($this->cache[$k]);
+				}
+				unset($this->tags[$tag]);
+			}
+		} else {
+			$this->cache = array();
+			$this->tags = array();
 		}
 	}
 }
--- a/tests/core/ConfigTest.php
+++ b/tests/core/ConfigTest.php
@@ -254,54 +254,92 @@ class ConfigTest extends SapphireTest {
 		$this->markTestIncomplete();
 	}
 
-	public function testLRUDiscarding() {
-		$cache = new ConfigTest_Config_LRU();
+	public function testCacheCleaning() {
+		$cache = new ConfigTest_Config_MemCache();
 
+		for ($i = 0; $i < 1000; $i++) $cache->set($i, $i);
+		$this->assertEquals(1000, count($cache->cache));
+
+		$cache->clean();
+		$this->assertEquals(0, count($cache->cache), 'Clean clears all items');
+		$this->assertFalse($cache->get(1), 'Clean clears all items');
+
+		$cache->set(1, 1, array('Foo'));
+		$this->assertEquals(1, count($cache->cache));
+		$this->assertEquals(1, count($cache->tags));
+
+		$cache->clean('Foo');
+		$this->assertEquals(0, count($cache->tags), 'Clean items with matching tag');
+		$this->assertFalse($cache->get(1), 'Clean items with matching tag');
+
+		$cache->set(1, 1, array('Foo', 'Bar'));
+		$this->assertEquals(2, count($cache->tags));
+		$this->assertEquals(1, count($cache->cache));
+
+		$cache->clean('Bar');
+		$this->assertEquals(1, count($cache->tags));
+		$this->assertEquals(0, count($cache->cache), 'Clean items with any single matching tag');
+		$this->assertFalse($cache->get(1), 'Clean items with any single matching tag');
+	}
+
+	public function testLRUDiscarding() {
+		$depSettings = Deprecation::dump_settings();
+		Deprecation::restore_settings(array(
+			'level' => false,
+			'version' => false,
+			'moduleVersions' => false,
+		));
+		$cache = new ConfigTest_Config_LRU();
 		for ($i = 0; $i < Config_LRU::SIZE*2; $i++) $cache->set($i, $i);
 		$this->assertEquals(
 			Config_LRU::SIZE, count($cache->indexing),
 			'Homogenous usage gives exact discarding'
 		);
-
 		$cache = new ConfigTest_Config_LRU();
-
 		for ($i = 0; $i < Config_LRU::SIZE; $i++) $cache->set($i, $i);
 		for ($i = 0; $i < Config_LRU::SIZE; $i++) $cache->set(-1, -1);
 		$this->assertLessThan(
 			Config_LRU::SIZE, count($cache->indexing),
 			'Heterogenous usage gives sufficient discarding'
 		);
+		Deprecation::restore_settings($depSettings);
 	}
 
 	public function testLRUCleaning() {
+		$depSettings = Deprecation::dump_settings();
+		Deprecation::restore_settings(array(
+			'level' => false,
+			'version' => false,
+			'moduleVersions' => false,
+		));
 		$cache = new ConfigTest_Config_LRU();
-
 		for ($i = 0; $i < Config_LRU::SIZE; $i++) $cache->set($i, $i);
 		$this->assertEquals(Config_LRU::SIZE, count($cache->indexing));
-
 		$cache->clean();
 		$this->assertEquals(0, count($cache->indexing), 'Clean clears all items');
 		$this->assertFalse($cache->get(1), 'Clean clears all items');
-
 		$cache->set(1, 1, array('Foo'));
 		$this->assertEquals(1, count($cache->indexing));
-
 		$cache->clean('Foo');
 		$this->assertEquals(0, count($cache->indexing), 'Clean items with matching tag');
 		$this->assertFalse($cache->get(1), 'Clean items with matching tag');
-
 		$cache->set(1, 1, array('Foo', 'Bar'));
 		$this->assertEquals(1, count($cache->indexing));
-
 		$cache->clean('Bar');
 		$this->assertEquals(0, count($cache->indexing), 'Clean items with any single matching tag');
 		$this->assertFalse($cache->get(1), 'Clean items with any single matching tag');
+		Deprecation::restore_settings($depSettings);
 	}
 }
 
 class ConfigTest_Config_LRU extends Config_LRU implements TestOnly {
-
 	public $cache;
 	public $indexing;
+}
+
+class ConfigTest_Config_MemCache extends Config_MemCache implements TestOnly {
+
+	public $cache;
+	public $tags;
 
 }


nfauchelle:
I was playing this this tonight.

Items below are using php 5.5

Using the 60 tests in my project times went from
avg 1.15 minutes with memory: 96.25Mb
to
avg 59.62 seconds with memory: 99.75Mb

When using my tests + userforms + sortablegrid tests;
avg 1.74 minutes with memory: 104.00Mb
to
avg 1.57 minutes with memory: 108.00Mb

I then started to use the cwp recipe.
$ vendor/bin/phpunit cwp-core/tests/
Time: 17.23 seconds, Memory: 94.75Mb
to
Time: 16.88 seconds, Memory: 100.75Mb

I tried to run all the tests in cwp but this died, I got two thirds though. 

After 70% of the 1676 tests:
Time: 796.3177 seconds, Memory: 192.5
to
Time: 782.0267 Memory: 207.5

So I think this is very worth while. Improved speed for a little more memory.

dhensby:
@nfauchelle for the ease of comparison, can you convert all times into seconds (rather than decimal minutes and sometimes seconds)?

dhensby:
> I tried to run all the tests in cwp but this died, I got two thirds though.

Why did it die? Due to memory usage?

> So I think this is very worth while. Improved speed for a little more memory.

This might not be the case if tests can't pass!

I think @hafriedlander original concerns were about memory usage in real-world usage too. Have you been able to compare something like memory required to load your homepage with and without this as well as pages with heavy lifting?

nfauchelle:
The test died before I had even made the change.
    ERROR [User Error]: Uncaught PHPUnit_Framework_Error: Couldn't run query: 
    ....
    The used table type doesn't support FULLTEXT indexes

I tested hitting the home page 20 times (scripted) and compared the times and memory before and after. This doesn't really effect those at all, you can only really notice a difference with tests since those make a lot of changes to the config system.

dhensby:
So travis on php 5.4 with this patch:

> Time: 344 seconds, Memory: 166.75Mb

travis on php 5.4 the last merge to 3:

> Time: 378 seconds, Memory: 162.25Mb

We have to weigh up is a sacrifice of 4.5Mb worth 35 seconds (on the test suite)? It probably is. But what are the *real world* savings on page loads?

Is the relation between memory consumption and speed improvement linear at all? Or is the memory increase fixed whilst time savings are simply based on how many accesses there are to the Config? Because then we may see fairly significant memory increase (proportionally) for insignificant improvements of real-world usage

dhensby:
> I tested hitting the home page 20 times (scripted) and compared the times and memory before and after. This doesn't really effect those at all, you can only really notice a difference with tests since those make a lot of changes to the config system.

Ok, if this is a bout build speed, then I'm all for it, especially if the memory footprint on real-world sites are practically unaffected - though I would like some real numbers for that.

dhensby:
I suppose one of the most obvious problems here is that this is no longer an LRU and so probably shouldn't be named as such.

There is no limit on how big  the LRU cache can be and so has theoretical infinite limit

This isn't great behaviour as there may be config options that are only used once and so caching them simply consumes memory for no real benefit.

nfauchelle:
Tests were performed on my macbook air. Each time the silverstripe-cache folder was deleted
the site was loaded in the browser to rebuild it all.

The script would hit the site once, then perform 25 additional hits taking the time and memory of each one.
Times were taking from the start of framework/main.php and stopped at the end with memory.

With the times take into account that got things like browser with a dozen tabs open, music playing etc. I ran the PHP 5.4 test 100 times so that'd be the most 'accurate'. 

Before change
-------------

PHP 5.3
Min Time: 0.7834 Max Time: 0.9453 Avg Time: 0.8107 Median Time: 0.8029
Min Memory: 44 Max Memory: 44 Avg Memory: 44.00 Median Memory: 44

PHP 5.4 (100 tests)
Min Time: 0.7573 Max Time: 0.8402 Avg Time: 0.7764 Median Time: 0.7719
Min Memory: 29.5 Max Memory: 29.5 Avg Memory: 29.50 Median Memory: 29.5

PHP 5.5
Min Time: 0.7605 Max Time: 0.9295 Avg Time: 0.7829 Median Time: 0.77545
Min Memory: 29.5 Max Memory: 29.5 Avg Memory: 29.50 Median Memory: 29.5


After change
-------------

PHP 5.3
Min Time: 0.6947 Max Time: 0.8125 Avg Time: 0.7404 Median Time: 0.7287
Min Memory: 44 Max Memory: 44 Avg Memory: 44.00 Median Memory: 44

PHP 5.4 (100 tests)
Min Time: 0.6779 Max Time: 0.8314 Avg Time: 0.7147 Median Time: 0.7047
Min Memory: 29.5 Max Memory: 29.5 Avg Memory: 29.50 Median Memory: 29.5

PHP 5.5
Min Time: 0.6907 Max Time: 0.7176 Avg Time: 0.7017 Median Time: 0.7018
Min Memory: 29.5 Max Memory: 29.5 Avg Memory: 29.50 Median Memory: 29.5

nfauchelle:
@dhensby
I think that is fine though, and like you said a 35 second gain for a few mb is well worth it! Trimming 10 seconds off my unit tests, well worth it!

The problem you say with the cache I don't think is a problem, it's just what happens with caches, some items get used more than others.

dhensby:
Ok, given those numbers it looks pretty straightforward. Your test site basically uses no extra memory. Is the site a fresh 3 site or is it a real site with new page types and config options?

> The problem you say with the cache I don't think is a problem, it's just what happens with caches, some items get used more than others.

well, Caches should generally have a limit to how large they can get; there needs to be a balance between it being too small that we introduce thrashing and too large that it ends up taking all the  resource.

An LRU is a type of cache where the items in the cache that are accessed the least are first to go. At the moment, there's no limit and no discarding of cache items.

At the least it shouldn't be called Config_LRU and perhaps just Config_Store. At best we should be providing some cache size limitations.

nfauchelle:
@dhensby 
I seup the site from this
`composer create-project cwp/cwp-installer cwp-installer --repository-url="https://packages.cwp.govt.nz"`
but added no extra content. There is a lot of extra modules in there!

dhensby:
OK. Great

camfindlay:
If you wanted to test against a CWP site with some content I think this module adds in some demo content https://gitlab.cwp.govt.nz/cwp/samplecontent/tree/master - that should really put this through it's paces for a benchmark.

sunnysideup:
great to see this back on the agenda.... THANK YOU!

nfauchelle:
@camfindlay
I had a quick look at the code in the repo. How/where does it populate the data? I don't mind running the benchmarks just give me a bit more info please.

camfindlay:
@nfauchelle I'll find out for you

nfauchelle:
@camfindlay 
*nudge* Would love to get this merged :)

camfindlay:
@nfauchelle actually appears that module adds form code and doesn't specifically add some demo content. Drat. Perhaps just benchmark with the cwp recipe default or you could use something like https://github.com/silverstripe/demo.silverstripe.org

nfauchelle:
@camfindlay 
Using the cwp default is what I had done.

https://github.com/silverstripe/silverstripe-framework/pull/3934#issuecomment-75744258
was done using
https://github.com/silverstripe/silverstripe-framework/pull/3934#issuecomment-75807298

sunnysideup:
should I run it on www.silverstripe-ecommerce.com and look at RAM use and that sort of thing?

nfauchelle:
@sunnysideup 
Sure can't hurt :)
The more testing the better, include the PHP version as well.

nfauchelle:
@dhensby  @camfindlay 
What do we need to do to get this merged?

camfindlay:
This might be something that can be discussed at the core committers google hangout session. Will remind them about it (I'm not a core committer ;) )

nfauchelle:
@camfindlay 
1) When is that?
2) Can't it be discussed here so it's open to all?

camfindlay:
The google hangout is public and is on in about half an hour at https://plus.google.com/b/106943799421369121040/events/ce712pt122h2j1d4mm7hipk12a0

dhensby:
As I've effectively authored this PR I'll need another member of the core team to review and approve it

nfauchelle:
@camfindlay @dhensby 
Thanks for raising this today.
So I think Sam was saying to merge it in and set a memory limit on travis. 

PHP 5.3 is sitting around 195mb currently for tests, im not sure how much memory the behat testing is using... so maybe look at 250mb - that'd catch any blow outs and still allow room to grow?

dhensby:
I think that setting a memory limit on the test suite isn't the best route to go down. The memory footprint of the test suite is around 200MB and that's WAY over what a normal site needs to handle a request (~50MB) - officially 48MB is required according to the [server requirements](http://doc.silverstripe.org/en/getting_started/server_requirements/).

micmania1:
Why not leave Config_LRU and then add a new class here with the ability to swap them out? That way devs can fine tune their site to what is most important for them.

The only down side to that it would probably have to be done using a constant, however until there's a great deal of work done on refactoring the boot loader if think that's acceptable. Eg. if(defined('SS_CONFIGBACKEND')) //...

nfauchelle:
@dhensby @sminnee 
Sorry... maybe I didn't hear Sam correctly in the Google Hangout yesterday, I thought that was his suggestion? If not then what do we need to do to get this merged, feels like we are going around in circles a little.

kinglozzer:
I’ve been playing around with this, but leaving the LRU logic in-place - I noticed a fair amount of the slowdown comes from the cleaning logic. Could I ask some of you guys to test this if you get the chance? :) https://github.com/kinglozzer/sapphire/commit/5c6a50e81334bf7a0dc809d507f604547d3f9809

These are my results (PHP 5.5.18, aggregated results over 5 XHProf runs):

Old code | LRU Removal (this PR) | My experimental code
----|----|----
**Total Incl. Wall Time (microsec):** 648,039 microsecs | **Total Incl. Wall Time (microsec):** 584,321 microsecs | **Total Incl. Wall Time (microsec):** 582,497 microsecs
**Total Incl. CPU (microsecs):** 556,226 microsecs | **Total Incl. CPU (microsecs):** 543,902 microsecs | **Total Incl. CPU (microsecs):** 540,443 microsecs
**Total Incl. MemUse (bytes):** 28,011,773 bytes | **Total Incl. MemUse (bytes):** 28,050,251 bytes | **Total Incl. MemUse (bytes):** 27,974,685 bytes

It’d be really interesting to see if you guys see the same speedup as me. My only concern with my approach is that the `$tags` array might get pretty large, I need to do a bit more testing to check for memory increases if that happens.

kinglozzer:
Okay, if I fill up the cache and access it to trigger the LRU setting/getting like this...

```php
for($i=0; $i < 3000; $i++) {
	Config::inst()->update('FooBar', 'barbar'.$i, array('fooo' => 'Barrrr'));
	config::inst()->get('FooBar', 'barbar'.$i);
}
```

... the results are:

Old code | LRU Removal (this PR) | My experimental code
----|----|----
**Total Incl. Wall Time (microsec):** 6,241,434 microsecs | **Total Incl. Wall Time (microsec):** 1,019,816 microsecs | **Total Incl. Wall Time (microsec):** 6,300,477 microsecs
**Total Incl. CPU (microsecs):** 6,165,120 microsecs | **Total Incl. CPU (microsecs):** 756,403 microsecs | **Total Incl. CPU (microsecs):** 6,238,755 microsecs
**Total Incl. MemUse (bytes):** 30,115,832 bytes | **Total Incl. MemUse (bytes):** 35,594,901 bytes | **Total Incl. MemUse (bytes):** 30,115,832 bytes

So, on the face of it, my attempt seems to save time (and even a small amount of memory) up until the point at which the cache fills up. After the LRU logic kicks in, it seems pretty-much identical to the existing code (no, I didn’t paste the same memory figures twice - they actually came out identical!).

dhensby:
I suppose given the same LRU limit and same Conig values you'd expect the memory usage to be the same. This test is a little contrived because the point of the LRU is to cache the most used items and this is likely not really testing that.

As I see it we have the following pros and cons of this PR vs keeping or optimising existing logic:

| Pros | Cons |
| ------ | ------ |
| Significantly faster | More memory usage with large Config sets |
| Less complex code to maintain | |

My opinion is we should maintain the most simple code we can whilst allowing the ability to use a different cache for projects where they have special requirements. That, plus the fact we're looking at 6 seconds vs 1 second there just doesn't seem to be any justification for not taking this PR.

kinglozzer:
>This test is a little contrived because the point of the LRU is to cache the most used items and this is likely not really testing that.

Yep, the second test wasn’t intended to be a performance test - simply a check for increased memory usage under huge loads - I should’ve made that clearer.

Given that the main “against” is the potential increase in memory usage it was just a suggestion that offers the same performance improvements as this PR, up to the point where the cache starts replacing items, without that increase.

> That, plus the fact we're looking at 6 seconds vs 1 second there just doesn't seem to be any justification for not taking this PR.

The 6 seconds is just because I’m pushing config values thousands of times in a loop (as above, my bad for not making that clearer!).

Either way, I think I agree with @micmania1’s suggestion of adding a new “backend” and making it configurable instead of just replacing the LRU cache with something that isn’t really an LRU.

nfauchelle:
@kinglozzer 
Im not sure if I agree with having a second backend, I don't think there is enough benefit in doing so. Why would people want to keep using the existing code?

I think the issue with with the memory is really a non issue.... you need to do something pretty freaky to make SS run out of memory by saving to many things to the cache... you can do that in other ways as well, by looping and creating millions of different Page objects, or you can create problems by having crazy cache keys and causing too many items to be saved in the silverstripe-cache (zend cache..?) folder (since those files don't really expire)...

It's not like this config LRU cache is going to magically fill up and use all the ram... it will only fill up through a LOT of set calls, and doing that is pretty abnormal. The unit tests which would call the config cache a lot didn't blow up, and useage with the CWP and a few other sites didn't blow up either.

But back to the point, I think this isn't different enough to warrant another backend, this isn't adding storage and persistence with redis or anything. If think if it was added, it should be removed in the next version once it's been out in a wild for a version.

*early morning long ramblings*

dhensby:
I think we should only be supporting one backend out of the box. The default cache store will invariably be the only one used and so the unused one will grow stale and won't have the usage to uncover any bugs or issues (or improvements).

I just don't see a real world example where we hit 3000 config declarations and that people are happy to trade time to free memory...

@kinglozzer I'm confused, so your benchmarks for this PR don't include 3,000 items but the other two do?

kinglozzer:
@dhensby The benchmarks in [this comment](https://github.com/silverstripe/silverstripe-framework/pull/3934#issuecomment-77595168) are on an in-development site with ~15 modules, the benchmarks in [this comment](https://github.com/silverstripe/silverstripe-framework/pull/3934#issuecomment-77598053) are on the same site, but with the cache being filled just to test memory usage with huge amounts of config declarations.

The reason I put [this code](https://github.com/kinglozzer/sapphire/commit/5c6a50e81334bf7a0dc809d507f604547d3f9809) up for consideration is that it offers the same speedup (unless you hit thousands of items which, as has been said, is very unlikely) without the unbounded memory increase, which was the main concern raised. I’m happy with this PR, I agree that simpler code is better, I was just offering up an alternative that I hoped to be somewhat “best of both worlds” :)

I agree that only one backend will probably ever be used, but I don’t think we should keep the `Config_LRU` name for a cache that isn’t an LRU cache - simply adding this in a new class would solve that, we could then remove the old `Config_LRU` in 4.0. Admittedly it’s unlikely that anyone will ever touch this class in user-code, so if everyone else is happy to leave the name as-is then I’ll back down on that :)

dhensby:
OK. If I understand correctly, then this PR is still significantly faster at assigning a large number of config vars and basically the same with a "normal" amount? 

I like your code and that you've got the best of both for a typical site. Perhaps it is more appropriate...

Does yours only slow once a site goes over 1,000 or is the process of setting and getting slow regardless of the limit? 

TBH, the more I think about it, the more arbitary the limit seems to me... As a developer I would always trade memory for speed, so if I hit the limit and started slowing things down due to thrashing, I'd just increase the cache size.

I'm conscious that I don't want to waste your time as I keep leaning towards the current PRs solution due to the fact the LRU seems completely theoretical and pointless. I can't think of a real world example where you'd genuinely not trade 40kb for 60ms or a theoretical one where you wouldn't trade 5MB for 5 seconds

dhensby:
OK. I think what's happening here is: 

1. we all agree this solution is suitable and does the job.
2. There is a concern that memory usage is  uncapped
3. Code is being produced (thus effort and time is ised) to try to mitigate this concern 
4. The memory usage concern is theoretical and unlikely to cause a real world problem
5. There should only be one solution in core (to reduce code maintenance)

Therefore, in the interest of reducing code complexity and increasing speed, I think we should take this in as-is.

@kinglozzer I like your solution as a fast but memory limited version, perhaps you should great a module with it and mention it in docs?

nfauchelle:
@dhensby 
Agree, would like to see this taken in as-is, and the sooner the better so it has longer to get 'tested' in use by people.

The class name will get renamed when the config system is overhauled so I wouldn't worry about that.

dhensby:
> The class name will get renamed when the config system is overhauled so I wouldn't worry about that.

Whose planning that? ;)

nfauchelle:
;-)

core/Config.php
```diff
@@ -678,77 +678,39 @@ public function remove($class, $name /*,$key = null*/ /*,$value = null*/) {
 
 }
 
+class Config_LRU extends ConfigCache {
+
+	public function __construct() {
+		Deprecation::notice('3.2', ' Config_LRU is deprecated, please use ConfigCache instead', Deprecation::SCOPE_CLASS);
```
tractorcow: Probably should be '4.0' due to semver.

Also, should config_lru implementation be kept in place, rather than extending the new class?

tractorcow:
> Why not leave Config_LRU and then add a new class here with the ability to swap them out? That way devs can fine tune their site to what is most important for them.

I agree with this. :)

core/Config.php
```diff
@@ -678,77 +678,39 @@ public function remove($class, $name /*,$key = null*/ /*,$value = null*/) {
...
+}
+
 /**
  * @package framework
  * @subpackage core
  */
-class Config_LRU {
-	const SIZE = 1000;
-
+class ConfigCache {
```
tractorcow: Naming convention for any class in a file is that any part before the first underscore should match the name of the file.

`Config_MemCache` would be a better classname.

core/Config.php
```diff
@@ -250,7 +250,7 @@ public static function unnest() {
 	 * leak through to other instances.
 	 */
 	public function __construct() {
-		$this->cache = new Config_LRU();
+		$this->cache = new ConfigCache();
```
tractorcow: This class now no longer exists. :)

nfauchelle:
@tractorcow 
Why would you agree to have two which you can toggle?

tractorcow:
Because you can't state that you are deprecating a class, and then replace it's core functionality. At the moment `Config_LRU` is NOT a LRU cache anymore. Semantic versioning says it's ok to add new api, but only in a backwards compatible way.

And they don't need to be swappable in the core, so I don't really care so much about that. I only care that the core functionality of that particular class is left intact, and we simply switch over to use the new better cache. We can remove the class it it's entirety in 4.0

nfauchelle:
> We can remove the class it it's entirety in 4.0.
That makes sense then. From your previous comment it sounded (well, I assumed) that it was planned to support both. I think if there is a plan to require the old one later then I think it's fine to have both for now.

tractorcow:
Yeah I probably should have pointed out the part I agreed with more specifically. :)

dhensby:
Ok. So @tractorcow's view is essentially this (tl;dr: he's right):

We're basically creating a new Config cache class, deleting Config_LRU and either pretending we didn't (keeping the same class name) or pretending we are deprecating (when really, we're removing the class and replacing it whilst pretending we are renaming).

To stay compliant with SemVer we should keep both classes and deprecate LRU. LRU can then be removed in 4.x

*Sigh*

Unfortunately he's right

tractorcow:
> Unfortunately he's right

Well I did relent on the deprecation-being-breaking issue. :P

nfauchelle:
I just wonder, is this a little bit of a slippery slope...

Bigger picture, are we going to have 6 months down the track a lot of flags / backend extra items simply because we can't move forward on things due to semver but want them in now anyway.

Maybe this should just get merged into 4.0 as is, and not worry about 3.x at all?

People that want it in earlier could apply the patch.... I would love to see it in 3.2, but is having both classes and a switch really the right thing for the project overall? If we start doing this to everything the project will get even more bloated..

dhensby:
@nfauchelle I think we're stepping into debating SemVer.

It could go straight to master, yes.

I suppose it depends how much effort we want to put in to get improvements into the 3 release instead of 4.

I'd like to get it into 3 and will put the effort in to do that

nfauchelle:
@dhensby 
I think SemVer is good and we should do it no doubt.

It's more a question around the back porting of things into 3 because 4 might be a long way away yet I guess.... (and, thanks for your work on getting this in).

tractorcow:
it's fine in 3.2

kinglozzer:
@dhensby Yeah you’re right, sorry for introducing more noise to the discussion!

The class renaming was my last concern and that has been solved, so I’m (finally) going to spank that green merge button. Thanks all, especially @dhensby and @nfauchelle for your time testing this!

nfauchelle:
YAY!

dhensby:
Thanks @kinglozzer and for the effort in testing this

tractorcow:
Good job. :D

halkyon:
Nice job guys. Every little performance improvement will help :)

PR #3720, opened 2014-12-10
NEW: CMS site tree status icons

jonom:
Added coloured status 'dots' to overlay page icons and improved visibility of status labels. As discussed in https://github.com/silverstripe/silverstripe-framework/pull/2794

img

PR #3936, opened 2015-02-24
Sitetree status icons (3 branch version)

jonom:
This is a version of https://github.com/silverstripe/silverstripe-framework/pull/3720 created for the 3 branch.

I ran in to issues with the admin/screen.css file - it didn't seem to be in sync with it's related scss files and needed to be re-compiled. It could be a configuration issue at my end though I guess. @clarkepaul can you comment?


dhensby:
It's possible that people have been making changes to the SCSS without compiling and/or manually changing CSS values by hand...

I'd be happy to take this in

dhensby:
*bump* @clarkepaul

clarkepaul:
If that was me sorry?  must have been years ago

dhensby:
Right, lets take this in... what's the worst that can happen?

@jonom can you confirm before I do that you don't notice any serious issues with front end when you do use this PR?

kinglozzer:
> I ran in to issues with the admin/screen.css file - it didn't seem to be in sync with it's related scss files and needed to be re-compiled.

I wouldn’t worry about this too much, it always seems to happen when I submit frontend changes!

I’ve just checked it out and had a click around and everything seems fine :)

jonom:
Sweet, yeah everything looked good to me when I tested it too

dhensby:
THANKS!

jonom:
Yesssssssss! Thanks for the merge

clarkepaul:
Nice one!

scott1702:
Hey @jonom, I've been looking over your work and its really awesome, nice job! One thing I'd like to change is the colours of the badges to be more easily viewable for people with various levels of colour-blindness. The 'modified' orange colour seems too close to the red and yellow of deleted/removed from draft and draft, I changed this to purple but I'm open to discussion about possible alternate colours. I also increased the colour contrast ratio of the current badges to be more visible. What do you think?
image

tractorcow:
I like the purple :D

jonom:
Hey @scott1702, glad you like it and fair point about contrast. @clarkepaul thought the badge colours might need some tweaking too. I'm not sure I'm fond of purple as I think it might clash with the rest of the design, but not sure what would be better. Maybe just a less bright purple?

Have you tested the current and revised colours with a simulator like http://www.adobe.com/ca/accessibility/products/photoshop.html#colorblind ? I haven't yet but I'd just want to be certain we're making a significant improvement with a change like this.

The good news is we're not only using colour to communicate here - users can see the text labels by rolling over links (plus they're always visible in the full site tree) so we do have a fallback at least.

A7DC:
Regarding the colours from an aesthetic point of view, they could do with some work. I agree with @jonom in the sense that, if the text is legible, then that is a sufficient enough way to communicate when users are unable to distinguish between colours.

scott1702:
I readjusted the colours slightly to match each other a bit better, and I made sure all badge colours meet a colour contrast ratio of 3:1. @A7DC also looked at the new colours through a colour-blind filter and they look fine.

image

jonom:
I've just had a quick look at the original colours proofed in Photoshop for colour blindness and I'm able to distinguish okay between red/orange/yellow with both types of colour blindness simulated. I think just increasing the contrast of the current colours might be enough, particularly since the rest of the CMS interface doesn't seem to make any special effort in regard to colour blindness.

Side note: Is there a policy about visual/design changes for the CMS? Community contributions are great but in order to keep a cohesive look throughout I feel like one or two people should have final say about graphic design. Plus when you put aside accessibility concerns there's often no obvious best answer when it comes to visuals, people just have different tastes, so it might be impossible to reach a consensus on some design decisions. For this reason I was imagining @clarkepaul (or someone else at SilverStripe Ltd) might make changes to what I submitted to ensure it matches the overall look of the CMS (what I did is quite pale and flat compared to the rest of the interface).

Ultimately whatever the majority wants will be fine with me but probably need a few more opinions here to determine which way to go, and maybe some input from those "design overseers" if they exist?

clarkepaul:
If there are any decisions that can't be made collectively then I'll jump in and make a call but I currently don't have the time to have a voice in in every matter. We now have a small team of which I spend some time with who will be working on improving the CMS and modules and they will start to define this sort of thing a little bit more for the open source project (like colours, min-contrast, default styles etc). Its going to be a step by step process though as this project doesn't have that consistency yet. At the moment we are going through making sure the issues identified last year the usability test are seen to before going on to new things.

I'll have a little play with the colours and see if there could be a replacement for the purple, the colour isn't too dominating and I could think of a better colour off the top of my head.

clarkepaul:
Ok so I've looked into the colours a bit deeper, and since we are changing delete to archive the red colour isn't as necessary as it once was. I associate red when an error occurs so for archive I have used a dulled down state (delete from draft will become archived). I have also introduced one of brand colours of SilverStripe which is the pink. I added some possible message/notification colours as well which is another part of the UI which didn't really get much love to see how these colours can be used more widely. 

I realise these mockups are quite different from what we had already so happy to hear what others think.
sitetree_states
states

jonom:
@clarkepaul, I like where you're going with the brand colours. The pink is still fairly close to red though so could still have error connotation. I wonder if we could take advantage of that though:

ss-colour-palette

To my mind orange is more appropriate for the modified state as the message is sort of *"Hey, you have some unsaved changes here, but it might not be a big deal".* For an unpublished page though I think the message is more urgent, like *"Hey, no one can see this page yet! You should publish this, or archive it maybe?"*

jonom:
p.s. I lightened the blue background as I thought it looked too solid compared to the other background colours. I also slightly darkened the green text to up the contrast but might need to go a little bit further even.

chillu:
Awesome discussion here! Agree with Jono that Paul's "modified" pink colour looks too much like an error. But I have the same concerns with Jono's "draft" colouring. Why do we need a different colour for "modified" and "draft" in the first place? They're both non-critical UI feedback which warrant the same level of attention from CMS authors ("remember to publish this draft" vs. "remember to publish those changes").

jonom:
Well, that's true. But with more colours it can look like a Site *Christmas* Tree :grinning: 

No but seriously, having orange for draft and modified is probably a good idea. Think Workflow should still have a different colour though.

clarkepaul:
I agree to some point but the little devil on the other shoulder is saying that a published page with changes is very different from a draft page which isn't published. I guess there is still the text to differentiate the two.

scott1702:
I talked with @clarkepaul and one colour for both draft and modified is the right way to go. I updated the colours based on his colours and @jonom's lighter blue for workflow. PR at #4204

jonom:
Just one thing to add: in my experience so far CMS users tend to unpublish pages as a way of archiving them. They know the content won't be visible to visitors but they can get the content back if they ever need to. If people continue to do this, then the modified pages that require action could get lost visually amongst 'faux-archived' pages since they have the same colour 'status dots'.

I think @tractorcow made some big changes to how archiving works, so as long as we're making the 'Archive' action really clear now and educating people about how it works, that shouldn't be a problem. But just wanted to put that user behaviour there.

clarkepaul:
This might be something we specifically need to check with users - how they used it before and after the change. We want to do user testing for everything medium/high risk from the work we deliver. 

FYI we also observed that users didn't know how to view deleted pages, so we have upcoming improvements for accessing archived pages (and the other states).

jonom:
Awesome yeah I didn't even realise myself that you could restore deleted pages in the CMS for a long time - stumbled upon that functionality by chance when playing with filtering the site tree.

purplespider:
I'm sorry I'm a bit late to this party, but I've just had a play with this in the 3.2-beta1 and am concerned about the decision to make Draft and Modified pages the same colour.

It didn't matter so much in 3.1, as the colour was always shown with the label, but in 3.2, there are times when **only** the colour is displayed, e.g. the dots in the editing mode site tree. I can see this causing confusion and reduced clarity over the status of all pages.

It was mentioned that the two page statuses are treated equally by editors, I don't believe this to always be the case. Pages that are marked as Modified are often likely to be pages that have had a minor tweak made to them, so it's no big deal for them to be published. However, Draft pages are often nowhere near ready to publish. Accidently publishing a "Modified" page is not normally a big problem, but accidently publishing a "Draft" page could well be a big problem.

My preference, taking into my experience of dozens of clients who use SilverStripe, would be for different colours.

clarkepaul:
Just an FYI the smaller sitetree area will be increasing in size and the label will be visible with the color, I'm pretty sure that it has been built so that even with a long status you can hover over it to show the full status.

scott1702:
An active/hovered page in the site tree will show a page's status (draft/modified/archived) so theoretically a content editor _should_ see the page state before they decide the make changes to or publish a page. However if you think content editors would prefer two separate colours for modified and draft pages I'd be happy to update them, keeping in mind the separate colours will be shown in the same place that the content editor will see the label.

jonom:
I would actually still prefer two colours too - I think the main issue we had was agreeing on what the colours should be :) think we can put the needs-to-be-high-contrast argument to bed since we decided in the end to use the same colour for both, so what if we just went back to a subtle difference with yellow for draft, orange for modified? Or vice versa. I also wonder in hindsight if a solid colour for the dot might look better.

@clarkepaul now that the sidebar area is going to be wider I wonder if I've made the nav look a bit awkward by making the highlighted item 100% width? I haven't downloaded the beta yet but will tomorrow, some of the changes I made were specifically to account for the narrow sidebar so might not be necessary anymore. But yes the label was designed to animate in on hover when space is limited.

jonom:
How about same colour labels for draft and modified, but the modified one has a 'weaker' dot?

screen shot 2015-06-22 at 11 47 50 am

dots

PR for that here: https://github.com/silverstripe/silverstripe-framework/pull/4335

clarkepaul:
Cool, I like it. Seems like you've thought about which one should be solid and not - I was going to say the "removed from draft" shouldn't be solid if "removed from draft" can be still published.
If the pattern is that the dot gets lighter the closer it is to being published:
Published = Nothing
Modified = Ring
Draft = Dot
Deleted from draft = Ring (could still be published)
Archived = Dot (is both deleted and unpublished) - new state in 3.2

jonom:
I was thinking Modified would be the only ring treatment but good point, Removed from draft could get it too. I think once 3.2 is released it will be pretty rare to see the "Removed from draft" status but makes sense to treat it a little differently. I will update the PR.

clarkepaul:
Based on your comments earlier and @chillu 's comment the full orange dot could be toned down. Something like this?
to-dot-or-not

jonom:
It's tricky to get an orange that still looks punchy - do you think this is okay or still too red?

screen shot 2015-06-22 at 3 38 17 pm

Also I think we're approaching 'Removed from Draft' the wrong way, think maybe we should make it an orange ring exactly like modified, because it represents an unpublished change and should draw more attention. Don't think a grey ring is strong enough for a page that has been removed from draft but is still live.

chillu:
I find it hard to see the difference between your first and second attempt, Jono - they both look red to me (=error). I don't think it needs to be very punchy - its easy to distinguish it from both the green dot (new page) and the unfilled dot (modified). So prefer Paul's orange dot. BTW, we should be safe in terms of colour blindness since we display the labels next to the tree node, right?

chillu:
Oh, we don't *have* a green circle, right? Maybe the red filled circle for "draft" should actually be a green unfilled circle? It's a "new page", right? Am I reopening an already resolved discussion here?

jonom:
Yeah that's an old discussion :) problem is green=success, just like red=error, it's why it's so hard to nail the colours here

jonom:
But I will dial it back so it's a nice ambiguous citrus tone ;)

jonom:
Think I'm having some colour profile issues with this orange - @clarkepaul what is the hex code of the solid orange in your dot?

orange

Also what do you think about the orange ring for 'Removed from Draft' here? Do you prefer it that way or with the grey ring to match the Archived colour?

clarkepaul:
I still prefer the grey ring because of whats happening to the state, I agree that it doesn't stand out as much as it probably should but I would prefer it follows the convention previously mentioned. I can totally see why you have given it the same visual cue as the modified page but it might be more confusing for the user to compare it with the modified state.

My hex for the dot is #ff7f22 but it didn't get it a very good way - it was a pretty random colour I just picked through Photoshop colour adjustments.

jonom:
Fair enough will keep it with the grey ring - #ff7f22 is the colour I used so must just be getting some colour profile discrepancies. Do you know if 'deletedonlive' is a real status that gets used anywhere? If I unpublish a page it gets the 'addedtodraft' status so I'm wondering if I can remove 'deletedonlive'.

clarkepaul:
@flashbackzoo @tractorcow do you guys know the answer to @jonom 's question?

scott1702:
'deletedonlive' is a state which was previously in the bulk actions. @tractorcow removed this in 3.2 as it had the same functionality as 'Unpublish' I believe.

tractorcow:
`deletedonlive` is replaced with `archived` from 3.2 onwards.

PR #3937, opened 2015-02-24
Sminnee use validationexceptions in forms

dhensby:
Note: completion of this work is included in the story #6228.
#2119 merged into master

Need to remerge the docs changes into the new docs structure

@sminnee an eye over this would help.

### DIFF
--- a/admin/code/CMSForm.php
+++ b/admin/code/CMSForm.php
@@ -8,26 +8,34 @@
  * @subpackage admin
  */
 class CMSForm extends Form {
-	
+
+	protected $responseNegotiator;
+
 	/**
 	 * Route validation error responses through response negotiator,
 	 * so they return the correct markup as expected by the requesting client.
 	 */
-	protected function getValidationErrorResponse() {
+	protected function getValidationErrorResponse(ValidationResult $result) {
 		$request = $this->getRequest();
 		$negotiator = $this->getResponseNegotiator();
 
 		if($request->isAjax() && $negotiator) {
-			$this->setupFormErrors();
+			// Load form errors from the result into the form
+			// Also save them to session, in case the negotation returns a 302
+			$this->setupFormErrors($result, $this->getData());
+			$this->saveFormErrorsToSession($result, $this->getData());
+
 			$result = $this->forTemplate();
 
+			$negotiator->setResponse(new SS_HTTPResponse($this));
 			return $negotiator->respond($request, array(
 				'CurrentForm' => function() use($result) {
 					return $result;
-				}
+				},
 			));
+
 		} else {
-			return parent::getValidationErrorResponse();
+			return parent::getValidationErrorResponse($result);
 		}
 	}
 
--- a/docs/en/changelogs/4.0.0.md
+++ b/docs/en/changelogs/4.0.0.md
@@ -5,6 +5,7 @@
 ### Framework
 
  * Deprecate `SQLQuery` in favour `SQLSelect`
+ * ValidationExceptions are now handled by Form, simplifying the use of model validation.
 
 ## Upgrading
 
--- a/forms/Form.php
+++ b/forms/Form.php
@@ -247,33 +247,34 @@ class Form extends RequestHandler {
 	);
 
 	/**
-	 * Set up current form errors in session to
-	 * the current form if appropriate.
+	 * Take errors from a ValidationResult and populate the form with the appropriate message.
+	 *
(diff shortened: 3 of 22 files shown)


dhensby:
Green ticks at last :)

dhensby:
@sminnee thoughts on this and help with docs would be appreciated :)

sminnee:
Thanks for bringing this into 2015. The code still seems to be in line with what I did and so I'm happy to merge it. I feel that pressing the button myself is too close to merging my own work, though. ;-) @tractorcow do you want to have a look, and if you're happy, do the honours?

dhensby:
Docs still need merging.

admin/code/CMSForm.php
```diff
@@ -8,26 +8,34 @@
  * @subpackage admin
  */
 class CMSForm extends Form {
-	
+
+	protected $responseNegotiator;
```
tractorcow: PHPDoc please for members :)

admin/code/CMSForm.php
```diff
@@ -8,26 +8,34 @@
...
 		$negotiator = $this->getResponseNegotiator();
 
 		if($request->isAjax() && $negotiator) {
-			$this->setupFormErrors();
+			// Load form errors from the result into the form
+			// Also save them to session, in case the negotation returns a 302
+			$this->setupFormErrors($result, $this->getData());
+			$this->saveFormErrorsToSession($result, $this->getData());
+
 			$result = $this->forTemplate();
```
tractorcow: Careful; Now $result is re-used with two different types. I would expect to see a new variable used for the actual response body.

forms/Form.php
```diff
@@ -552,13 +578,13 @@ public function getRedirectToFormOnValidationError() {
...
+	 * @deprecated 3.2
 	 */
-	public function addErrorMessage($fieldName, $message, $messageType, $escapeHtml = true) {
-		Session::add_to_array("FormInfo.{$this->FormName()}.errors",  array(
-			'fieldName' => $fieldName,
-			'message' => $escapeHtml ? Convert::raw2xml($message) : $message,
-			'messageType' => $messageType,
-		));
+	public function addErrorMessage($fieldName, $message, $messageType) {
+		Deprecation::notice('3.2', 'Throw a ValidationException instead.');
```
tractorcow: Is this really the only possible option? What if you want to add several errors and then do some custom routing with the resulting form? Couldn't the deprecation notice ask the user to use getSessionValidationResult and add the error there instead?

docs/en/changelogs/4.0.0.md
```diff
@@ -5,6 +5,7 @@
 ### Framework
 
  * Deprecate `SQLQuery` in favour `SQLSelect`
+ * ValidationExceptions are now handled by Form, simplifying the use of model validation.
```
tractorcow: Do we need upgrading notes?

forms/Form.php
```diff
@@ -449,36 +469,42 @@ public function checkAccessAction($action) {
 	 *
 	 * @return SS_HTTPResponse|string
```
tractorcow: I don't think this returns a string anymore

forms/Form.php
```diff
@@ -1181,31 +1189,66 @@ public function setMessage($message, $type, $escapeHtml = true) {
...
+
+	/**
+	 * Returns the ValidationResult stored in the session.
+	 * You can use this to modify messages without throwing a ValidationException.
+	 * If a ValidationResult doesn't yet exist, a new one will be created
+	 *
+	 * @return ValidationResult The ValidationResult object stored in the session
+	 */
+	public function getSessionValidationResult() {
+		$result = Session::get("FormInfo.{$this->FormName()}.result");
```
tractorcow: I know that we can store serialised objects in the session, but couldn't this be a lot more data than we need? (Given ValidationResult extends Object, and all of that data is potentially being serialised along with it).

I'd be happy if ValidationResult implemented Serializable so that I knew my session wasn't being filled up with garbage. :)

tractorcow:
Hi @dhensby @sminnee I've given this a cursory look, but I need to spend some more time (when it comes around) to look at this PR in full soon.

It'd be interesting to see how this works with the new validation rules in 4.0 for formfields merged a few weeks ago.

dhensby:
heh, I think I need to hand this back to @sminnee :p

dhensby:
I'm not realistically going to get round to doing this - my attempted rescue of the original PR ends here

sminnee:
We're going to get open sourcerers to look at this in alpha4, so reopening.

PR #3941, opened 2015-02-25
Minor correction to code for getting row count.

johndalangin:


wilr:
You don't need to call execute before calling value in 3.1.

johndalangin:
Oh, sorry Will. Apparently it was working all along.

PR #3942, opened 2015-02-25
match scaffoldFormField signature on Varchar to DBField

briceburg:
The current code throws STRICT warnings when Varchar is extended. E.g. 

```php
class FlexiChoice extends Varchar
{

    // throws a STRICT warning because signature does not Match DBField
    public function scaffoldFormField($title = null) {
        $field = new FlexiChoiceField($this->name, $title);
        return $field;
    }

    ...
}
```

### DIFF
--- a/model/fieldtypes/Varchar.php
+++ b/model/fieldtypes/Varchar.php
@@ -93,7 +93,7 @@ class Varchar extends StringField {
 	 * (non-PHPdoc)
 	 * @see DBField::scaffoldFormField()
 	 */
-	public function scaffoldFormField($title = null, $params = null) {
+	public function scaffoldFormField($title = null) {
 		if(!$this->nullifyEmpty) {
 			// Allow the user to select if it's null instead of automatically assuming empty string is
 			return new NullableField(new TextField($this->name, $title));


kinglozzer:
@briceburg you just need to update your method signature to match `Varchar`’s signature. I.e.:

```php
class FlexiChoice extends Varchar {
        $field = new FlexiChoiceField($this->name, $title);
        return $field;
    }
}
```

kinglozzer:
Although... the second argument is never used and there are many other field types with this method signature that also don’t use it. I think if we remove the second argument from one field type, we should remove it from all of them (that is, assuming none of them use the `$params` arg).

I’m not sure if that constitutes an API change - I’m leaning towards no because:
- `$params` isn’t used by anything
- Any subclasses that included `$params` will have defined a default value, even if it’s just `null` (unless strict notices are suppressed, in which case they won’t notice anyway!), so removing it won’t affect that

This isn’t really fixing a bug as such, so I’d say this could go against the `3` branch (tidying up this class and other classes that don’t use the `$params` argument) :)

briceburg:
@kinglozzer Hey loz, roger that. My intent was to future proof code against any future tidying you guys make. Matching the method signature of DBField seems like the best approach long term. Thanks!

dhensby:
It's *definitely* an API change to remove it. Though if it's never in use, it should probably go.

PR #3943, opened 2015-02-25
FIX: Ensuring GridFieldDetailForm validator uses Object->hasMethod() to respect e...

patricknelson:
I noticed that while SilverStripe utilizes a decorator pattern with it's extensions, it didn't appear to respect (or see) my custom `->getCMSValidator()` method on one of my extensions. I'm not sure if this was by accident or design, but it appears to have been due to using the native `method_exists()` function instead of calling the `->hasMethod()` object descended from the `Object` parent. 

Now for my particular use case **specifically** with `getCMSValidator()`, I'm fine if that method gets defined on the target object and overrides my decorator method (a workaround would probably be simple within the method itself). What I'm wondering is this: There are many other instances here in the framework that I'm seeing where `method_exists()` is being used where potentially `Object->hasMethod()` could be getting called instead (to help prevent that frustrating hit-or-miss end result that caused me to almost pull my hair out). 

**So:** Would it make sense to drop a `hasMethod($object, $method)` function somewhere into the framework and then use that in place of `method_exists()` and within that function simply rely on `->hasMethod()` (if the object is an `instanceof Object` or otherwise falling back to `method_exists()`? To me it would make sense to centralize and utilize this as much as possible, so that's why I wanted to open this pull request; to open a discussion about that as well.

### DIFF
--- a/forms/gridfield/GridFieldDetailForm.php
+++ b/forms/gridfield/GridFieldDetailForm.php
@@ -95,7 +95,7 @@ class GridFieldDetailForm implements GridField_URLHandler {
 
 		// if no validator has been set on the GridField and the record has a
 		// CMS validator, use that.
-		if(!$this->getValidator() && method_exists($record, 'getCMSValidator')) {
+		if(!$this->getValidator() && (method_exists($record, 'getCMSValidator') || $record instanceof Object && $record->hasMethod('getCMSValidator'))) {
 			$this->setValidator($record->getCMSValidator());
 		}
 


wilr:
Yes, method_exists() should be avoided as much as possible. The other tagent related issue is getCMSValidator is not defined on DataObject and so only one will be called. A related fix but for 4.0  would be to add getCMSValidator() to the base dataobject and include an updateCMSValidator extend hook as pair with getCMSFields.

patricknelson:
Hi will

As a relatively new user and more used to Laravel, I like their use of evens and observers. I've got more experience with events which can be defined statically in closures and triggered in the order they were registered. Easiest to dfine externally for separation of concerns (eg my extension can register it without the affected DataObject having to be tightly bound via hook unless I needed something in the middle of my method, even then, I could trigger a custom event there too). Just food for thought.

patricknelson:
Sorry I'm on my phone.

patricknelson:
I edited my previous comment, since I was on my phone earlier. Sorry for all the typos, I was in a hurry to finish up since I was at an event that was about to start) :D

dhensby:
There is a desire to move to an event based system and something that will likely be looked at.

In respect to the bug / question you raise - as @wilr says `hasMethod()` should always be used where applicable.

forms/gridfield/GridFieldDetailForm.php
```diff
@@ -95,7 +95,7 @@ public function handleItem($gridField, $request) {
 
 		// if no validator has been set on the GridField and the record has a
 		// CMS validator, use that.
-		if(!$this->getValidator() && method_exists($record, 'getCMSValidator')) {
+		if(!$this->getValidator() && $record->hasMethod('getCMSValidator')) {
```
dhensby: I'm not certain that $record is always an `Object` here :/

patricknelson: Yep, you are right, that is my fear. Would there be any objection to the approach of abstracting to a simple function? That way any `instanceof Object` could be easily checked using the proper technique and then everything else would be checked via  `method_exists()`.

dhensby:
defining a global helper function... hmm... I'm not adverse to it myself...

@tractorcow thoughts?

tractorcow:
Something like:

```php
class Object {
    public static object_has_method($object, $method) {
        return method_exists($object, 'hasMethod')
            ? $record->hasMethod($method)
            : method_exists($record, $method);
}
```

And it doesn't have to be a global method. It could be a static method an object.

patricknelson:
Right, it needs to be called statically (i.e. it can't be instance level due to the nature of it). I'll argue for the global helper function for a moment, however. The `Object` object already has the instance level `->hasMethod()` on it and technically `->object_has_method()` can also be called from an instance context (probably initially confusing newer users such as myself with an IDE like PhpStorm when those options pop-up). Also, to me, the syntax of:

```php
if (hasMethod($object, 'myMethod')) $object->myMethod();
```

Seems more intuitive, less verbose and less redundant than:

```php
if (Object::object_has_method($object, 'myMethod')) $object->myMethod();
```
Finally, from a functionality standpoint, we only really know *for sure* that the method `->hasMethod()` will do what we intend it to do when it's on an instance of `Object` (before we call it), so if we're incorporating a framework level method intended for general re-use and abstracting our desired functionality of determining if extension/decorator functionality (or built-in PHP defined method) functionality is setup on an object, I'd suggest only calling `->hasMethod()` on *instances* of `Object` *and then* determining if it has the desired method.

For example, my revised version of the above would then be:

```php
function hasMethod($object, $method) {
	return $object instanceof Object
		? $object->hasMethod($method)
		: method_exists($object, $method);
}
```

dhensby:
I am a fan of duck typing and don't see the need to explicitly check that the object is indeed an `instanceof Object`, however that may not be a shared opinion.

The argument of a helper function does appeal to me over slapping yet another method onto `Object` and having it a mish-mash of helper statics and instance methods. Generally I'd like a class to either be a static utility class OR a instance class, but not both.

Though polluting the global function scope with helpers is nasty too... Though, we already do it.

In an ideal world, we'd have a set of Utility classes that were appropriately named for these types of functions and the methods would sit in there.

patricknelson:
Also ideally we'd be using namespaces, as I too am *very much not* a fan of polluting global scope either. Stuffing static methods onto an object is a doable (but not the best) method of name spacing. However, at least the `hasMethod()` alias follows the nomenclature (albeit not the parameter signature) of the original `Object->hasMethod()` and, given it's global scope, would be useful in a general sense. 

On duck typing, either way is fine with me, as long as we call the same thing regularly and routinely all over the place and don't end up with the same thing implemented 3 different ways and called just as many different ways by the different developers that work on this system. For me, if an object has  a `->hasMethod()` method on it, it will look like a duck and hopefully act like one too but could act in unexpected ways to developers who implement it outside of the `Object` inheritance. That's because you could shake a stick at them for not sticking to good coding standards by not matching the parent method signature (i.e. ensuring that you don't remove parameters, accept the same types, return the same types, etc) because otherwise you'll already intuitive understand that you'll cause errors. So, if it's not a child of `Object` (if you've overridden or not) then I can know that you've potentially not matched the same signature of that method so I shouldn't call it. Looks like a duck, may not walk or quack like one. But if it looks like a duck, i.e. via overriding, and doesn't walk/quack like one, then you've got bad code smell (IMHO).

dhensby:
> However, at least the hasMethod() alias follows the nomenclature (albeit not the parameter signature) of the original Object->hasMethod() and, given it's global scope, would be useful in a general sense

Yes, this is the biggest reason I see to putting it in the global scope

I don't think this is the place to get into duck typing argument; you've covered off the pros and cons of duck typing and it comes down to opinion to make a decision. If anything, it's an argument for using interfaces more liberally and then this wouldn't even be a discussion point

patricknelson:
Ok, no problem. I'm just trying to nail down what we should do to implement this so it's ready for merge when it's completed. So should I go this route instead?

```php
function hasMethod($object, $method) {
	return method_exists($object, 'hasMethod')
		? $object->hasMethod($method)
		: method_exists($object, $method);
}
```

dhensby:
I give @tractorcow the final say on whether this should go on `Object` or not as I see the benefits to both and am not strongly opinionated enough on it to make a big deal about it; but I am leaning towards global function.

Thoughts on proposed code:

1. Naming, shouldn't it be has_method?
2. `method_exists` accepts a string OR an object so we need to either ensure `$objecy` is an object before calling on it or throwing an `InvalidArgumentException` before we run that logic.

tractorcow:
Just in case this short circuits the whole conversation, any new API can't go into 3.1, so if you want to just 'fix' it for the given use case, it would be a protected helper method on GridFieldDetailForm, in advance of some API being created further down the line.

In that case, I'd opt for extending Object to allow to check instances of specific objects, but perhaps by 4.0 we'll have trait-powered extensions, which means we may as well just use method_exists anyway. :)

patricknelson:
Right. In this case, there's just this one instance where it would be getting changed to use `hasMethod()` and that's already being done (save for the quick `instanceof Object` that needs to be thrown in as well), so as far as I can tell, no further abstraction (if only done in that class) will be needed. There are maybe 70 other usages of `method_exists()` through this repository and that sort of change (and I suppose the discussion it would require) is I'm sure well outside the scope of the bug I was trying to fix here, anyway. Albeit, the bug is related to that topic and appears to warrant further investigation.

dhensby:
> Just in case this short circuits the whole conversation, any new API can't go into 3.1

Yes, this needs to go against 3 branch

dhensby:
let's get this fixed either by doing `method_exists($record, 'hasMethod')` or checking `instanceof` (I think the former is more common in the framework?!)

dhensby:
@patricknelson any progress on getting this specific patch amended?

patricknelson:
Ok, made a quick update to the code to perform both `method_exists` and `instanceof Object` since I wasn't really sure which to go with, so I figured "Why not both?" (for now) since we're only going to resolve this directly related to the issue I was having and not address it at a wider API level.

dhensby:
Looks sensible, thanks. Will wait for green ticks

patricknelson:
No problem. I know it's only one line, but I figured we needed to settle on a good solution. That and I got pretty tied up with work recently :)

dhensby:
Thanks

Issue #3944, opened 2015-02-25
Module  sub-theme template inheritance issues


jedateach:
I like to use themes for SS development, more so I like to put module-specific themes into their own folders. (not sure what to call them - theme_module folders?)

eg:
```
themes/foo
themes/foo_shop
themes/foo_blog
```

Secondly, I've also used subfolders of template such as 'email', 'forms', and 'order', which is outside the normal "Includes" and "Layout" format..but not unheard of - just look at `framework/templates`

I've had problems  overriding these custom folders (found in modules) with theme and theme_module folders.

The docs talk about the inheritance order being: mysite, themes, modules, framework
http://doc.silverstripe.org/en/developer_guides/templates/template_inheritance

The docs don't talk about theme_module folders, or the use of folders outside of "Includes" and "Layout". I think they should be at least documented as a result of this issue.

[Line 189 of TemplateManifest](https://github.com/silverstripe/silverstripe-framework/blob/3.1/core/manifest/TemplateManifest.php#L189) looks like the place where theme_module folders are included in manifest. Looks like they are lumped in with the parent theme in the templates manifest array.

jedateach:
I've created a bunch of templates to observe the ordering.

Here's the output of `var_dump($this->templates['testtemplate']);` from TemplateManifest, after it has been regnerated.
```
array (size=5)
  'main' => string '/Users/jeremy/Sites/debugging/mymodule/templates/TestTemplate.ss' (length=64)
  'mysite' => 
    array (size=3)
      'main' => string '/Users/jeremy/Sites/debugging/mysite/templates/TestTemplate.ss' (length=62)
      'Includes' => string '/Users/jeremy/Sites/debugging/mysite/templates/Includes/TestTemplate.ss' (length=71)
      'other' => string '/Users/jeremy/Sites/debugging/mysite/templates/other/TestTemplate.ss' (length=68)
  'Includes' => string '/Users/jeremy/Sites/debugging/mymodule/templates/Includes/TestTemplate.ss' (length=73)
  'other' => string '/Users/jeremy/Sites/debugging/mymodule/templates/other/TestTemplate.ss' (length=70)
  'themes' => 
    array (size=1)
      'simple' => 
        array (size=3)
          'main' => string '/Users/jeremy/Sites/debugging/themes/simple_mymodule/templates/TestTemplate.ss' (length=78)
[...]

PR #3945, opened 2015-02-25
Explain the module specific themes separation capability of Silverstripe

jedateach:
A problem I had recently highlighted the need for some missing docs on how the `theme_module` setup is possible.

 #3944

### DIFF
--- a/docs/en/02_Developer_Guides/01_Templates/05_Template_Inheritance.md
+++ b/docs/en/02_Developer_Guides/01_Templates/05_Template_Inheritance.md
@@ -46,9 +46,10 @@ your entire project for the appropriate `.ss` files located in `template` direct
 It will each and prioritize templates in the following priority:
 
 1. mysite (or other name given to site folder)
-2. themes
-3. modules 
-4. framework.
+2. module-specific themes (e.g. themes/simple_blog)
+3. themes (e.g. themes/simple)
+4. modules  (e.g. blog)
+5. framework
 
 <div class="warning">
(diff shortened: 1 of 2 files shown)


dhensby:
:mind_blown: is this true?! Are there even any tests for this behaviour?

jedateach:
Ya - its been around since 2.3 / 2.4

I think this line is the best proof:
https://github.com/silverstripe/silverstripe-framework/blob/3.1/core/manifest/TemplateManifest.php#L189

Not sure if there are tests.

docs/en/02_Developer_Guides/01_Templates/05_Template_Inheritance.md
```diff
@@ -46,9 +46,10 @@ your entire project for the appropriate `.ss` files located in `template` direct
 It will each and prioritize templates in the following priority:
 
 1. mysite (or other name given to site folder)
-2. themes
-3. modules 
-4. framework.
+2. module-specific themes (e.g. themes/simple_blog)
```
micmania1: Is this definitely the right priority? I answered this a while back: http://stackoverflow.com/questions/24969798/blog-template-isnt-overridden - I'm not sure if I'm right but I'm guessing I would have tested at the time. I'm fairly certain that themes are higher priority then mysite.

dhensby: My understanding is that mysite trumps all

micmania1: Appears so :)

PR #3946, opened 2015-02-26
Test against php 7

dhensby:
Is this worthwhile? I don't see why not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-language: php 
+language: php
 
 php:
   - 5.4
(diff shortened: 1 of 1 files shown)


dhensby:
I'd like to get the HHVM build working and would be interested in making 7 work too (if possible), so I'd like these to go in myself

kinglozzer:
I’d definitely like to see the HHVM build working! I’m less concerned about PHP 7 at this stage as it’s not feature-complete (and still buggy), but I don’t see any reason why it can’t be an allowed failure for now :). Just as long as it’s not left to rot like HHVM has been!

> would be interested in making 7 work too (if possible)

On this topic: a colleague of mine played around with it very briefly last week, something similar to [this issue](https://bugs.php.net/bug.php?id=69124) was the main hurdle we encountered. Calls to `Object::extend()` seem to do something internally to strings which causes the same error - probably something to do with references, we didn’t investigate!

dhensby:
Removing hhvm-nightly as it's no longer supported by Travis (https://github.com/travis-ci/travis-ci/issues/3788)

PR #3947, opened 2015-02-26
DOCS Reinstate a missing upload field doc and give a review.

camfindlay:
Once merged @tractorcow can you check the information for current validity?


tractorcow:
Docs are fine. :D The docs used to be incorrect, but since recent security updates these instructions should be safe to follow again.

Issue #3938, opened 2015-02-25
RFC: DataList::create() is case sensitive


pedro2555:
Creating a file `/mysite/code/MyObject.php` with:

    class MyObject extends DataObject { }

`/dev/build?flush` doesn't cry. And you can get a list of MyObjects using:

    DataList::create('MyObject');

But you can't with:

    DataList::create('myObject');

On Linux this is okish, on Windows it isn't. Overall it makes no sense, since URLs are not case sensitive no website framework should be.

Further more, I can create `/mysite/code/myObject.php` (along side with `MyObject`) with:

    class myObject extends DataObject { }

`/dev/build?flush` complains about having two files containing 'myObject' class and refuses to continue, which makes sense.

I haven't done any deep investigation but I would assume this is a fault on the database configuration, I'm using MySQL 5.5.40, and running `select * from SiteTree;` isn't the same as `select * from siteTree;`.

pedro2555:
Further more this appears to happen only with `DataObject`, extending `Page` and I can query for `Page`, `page` or `pAgE`.

dhensby:
In PHP classes and function names are case insensitive.

The framework *should* be mimicking this. Do you have any test cases that prove the issue? The class manifest usually lower strings everything for it's comparisons.

pedro2555:
Yes I do. As stated, create a `MyObject` class `/mysite/code/MyObject.php`

    class MyObject extends DataObject { }

Then try and create a `DataList` via `DataList::create()` with any case variant of `MyObject` that it is not `MyObject` and it will result in a server error.

For example, in `Page_Controller`:

    static $allowed_actions = array('Test');

    public function Test() {
        echo DataList::create('MyObject`)->count(); // Executes OK
        echo DataList::create('myobject'')->count(); // Results in server error
    }

Accessing `localhost/home/test` should have resulted in `00`, but instead it prints out `0` and then the 500 server error page.

And it is not a `count()` issue, running a foreach also throws a server error.

pedro2555:
If you can point out possible fail points in the source code, I might look into it and send a patch. So far I could not figure out where the problem may reside.

dhensby:
By test cases I meant actual unit tests that show the unexpected behaviour

pedro2555:
Sorry no. Not very familiar with testing in PHP.

tractorcow:
Case insensitivity is a large issue in the framework; It's implemented sparsely across several places, but not in any real consistent way.

[...]

PR #3949, opened 2015-02-27
Ensuring classinfo is case insensitive

dhensby:
potential fix to #3938

### DIFF
--- a/core/ClassInfo.php
+++ b/core/ClassInfo.php
@@ -3,8 +3,8 @@
 /**
  * Provides introspection information about the class tree.
  *
- * It's a cached wrapper around the built-in class functions.  SilverStripe uses 
- * class introspection heavily and without the caching it creates an unfortunate 
+ * It's a cached wrapper around the built-in class functions.  SilverStripe uses
+ * class introspection heavily and without the caching it creates an unfortunate
  * performance hit.
  *
  * @package framework
@@ -35,7 +35,7 @@ class ClassInfo {
 	 * @var Array Cache for {@link ancestry()}.
 	 */
 	private static $_cache_ancestry = array();
-	
+
 	/**
 	 * @todo Move this to SS_Database or DB
 	 */
@@ -52,17 +52,18 @@ class ClassInfo {
 			return false;
 		}
 	}
-	
+
 	public static function reset_db_cache() {
 		self::$_cache_all_tables = null;
 		self::$_cache_ancestry = array();
 	}
-	
+
 	/**
 	 * Returns the manifest of all classes which are present in the database.
 	 * @param string $class Class name to check enum values for ClassName field
 	 */
 	public static function getValidSubClasses($class = 'SiteTree', $includeUnbacked = false) {
+		$class = self::class_name($class);
 		$classes = DB::getConn()->enumValuesForField($class, 'ClassName');
 		if (!$includeUnbacked) $classes = array_filter($classes, array('ClassInfo', 'exists'));
 		return $classes;
@@ -71,7 +72,7 @@ class ClassInfo {
 	/**
 	 * Returns an array of the current class and all its ancestors and children
 	 * which have a DB table.
-	 * 
+	 *
 	 * @param string|object $class
 	 * @todo Move this into data object
 	 * @return array
@@ -79,9 +80,7 @@ class ClassInfo {
 	public static function dataClassesFor($class) {
 		$result = array();
 
-		if (is_object($class)) {
-			$class = get_class($class);
-		}
+		$class = self::class_name($class);
 
 		$classes = array_merge(
 			self::ancestry($class),
@@ -102,7 +101,7 @@ class ClassInfo {
 	 * @return string
 	 */
 	public static function baseDataClass($class) {
-		if (is_object($class)) $class = get_class($class);
+		$class = self::class_name($class);
 
 		if (!is_subclass_of($class, 'DataObject')) {
 			throw new InvalidArgumentException("$class is not a subclass of DataObject");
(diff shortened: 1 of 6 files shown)


pedro2555:
Nice, I will give it a shot.

dhensby:
@pedro2555 this isn't working yet :(

dhensby:
To fix this bug essentially requires an API change (unless we start normalising class names in other places) and so ancestry now returns a lowercase key => correct casting.

I think the `ClassInfo` helper needs a bit of an overhaul to make it less vulnerable to case and to provide lookup functions.

core/ClassInfo.php
```diff
@@ -52,17 +52,20 @@ public static function hasTable($class) {
...
 	}
-	
+
 	/**
 	 * Returns the manifest of all classes which are present in the database.
 	 * @param string $class Class name to check enum values for ClassName field
 	 */
 	public static function getValidSubClasses($class = 'SiteTree', $includeUnbacked = false) {
+		if (self::hasTable($class)) {
+			$class = self::$_cache_all_tables[strtolower($class)];
```
tractorcow: This seems complicated to me. singleton($class)->class is a lot less work, and DataObjects can't be abstract anyway so no chance of that breaking here.

core/ClassInfo.php
```diff
@@ -102,7 +105,14 @@ public static function dataClassesFor($class) {
 	 * @return string
 	 */
 	public static function baseDataClass($class) {
-		if (is_object($class)) $class = get_class($class);
+		if (is_object($class)) {
+			$class = get_class($class);
+		}
+		else {
+			//normalise the class case
+			$reflection = new ReflectionClass($class);
```
tractorcow: I would put this into a separate method, rather than inlining the name normalisation here. That would probably require it to go into 3.2 anyway, however (maybe?). It would be better to have a single place where class normalisation is done, and it may as well be in this class anyway.

core/ClassInfo.php
```diff
@@ -121,23 +131,31 @@ public static function baseDataClass($class) {
...
 	 * @param mixed $class string of the classname or instance of the class
 	 * @return array Names of all subclasses as an associative array.
 	 */
 	public static function subclassesFor($class) {
+		//normalise class case
+		if (is_object($class)) {
+			$className = get_class($class);
+		}
+		else {
+			$reflection = new ReflectionClass($class);
```
tractorcow: Repeated from above; Refactoring into a normalisation method would be awesome. :)

core/ClassInfo.php
```diff
@@ -261,23 +284,27 @@ public static function has_method_from($class, $method, $compclass) {
 	 * @return string
 	 */
 	public static function table_for_object_field($candidateClass, $fieldName) {
-		if(!$candidateClass || !$fieldName) {
+		if(!$candidateClass || !$fieldName || !is_subclass_of($candidateClass, 'DataObject')) {
```
tractorcow: Nice catch.

core/ClassInfo.php
```diff
@@ -261,23 +284,27 @@ public static function has_method_from($class, $method, $compclass) {
...
 			return null;
 		}
 
-		$exists = class_exists($candidateClass);
+		//normalise class name
+		$reflection = new ReflectionClass($candidateClass);
```
tractorcow: As noted above, could be in its own normalisation method.

core/ClassInfo.php
dhensby: The problem is that there's no guarantee that the class is ACTUALLY a sublcass of `DataObject`.

Also, I'm scared of the performance impact of creating singletons for every class just to get a normalised classname. strtolower is probably the least CPU intense way of making them consistent?

core/ClassInfo.php
dhensby: Agreed - I thought that as I repeatedly wrote it out.

dhensby:
The more I look at this work the more I think this really needs to go against master with a bit of refactor across the board to make the entire framework more agnostic to case.

I dislike using the `ReflectionClass` to normalise the class name as the manifest is perfectly capable of storing this itself but that's another API change and refactor

core/ClassInfo.php
tractorcow: Singletons should only be hard to do the first time they're called.

There's nothing wrong with strtolower... it's the hasTable that invokes the database and is my concern. On the other hand, we probably shouldn't be calling db queries on tables that may not exist, so maybe that should be there anyway.

Let's complain about another thing. :) I'm not sure that you should rely on the $_cache_all_tables private static outside of the method that generates the cache, since it makes it harder to read.

How about you use the suggested `normalise_class_name` method here instead?

tractorcow:
Let's suggest this then:
- Move this to 4, and update the manifest to better support caching of classnames in a caseless fashion (perhaps a cache of lowercase => CorrectCase?)
- Use a single method in ClassInfo to invoke the case correction and use it across the board.

tractorcow:
Oh, I found a better method. :)

```php
/**
 * Returns the normalised class name, or false if not a class
 * @param string $class
 * @return string|false
 */
public static function normalise_class_name($class) {
if(!class_exists($class)) return false;
foreach(get_declared_classes() as $caseClass) {
   if(strcasecmp($class, $caseClass) == 0) return $caseClass;
}
return $class;
}
```

pedro2555:
@tractorcow What a sweet piece of code

tractorcow:
Probably slow as hell though. Worth benchmarking vs singleton($class) method.

pedro2555:
Quick and dirt test vs singleton, and it is around 3x faster.

core/ClassInfo.php
dhensby: `hasTable` only calls the DB once and it's a `LIST TABLES` query. I assume this is called at least once every time there's a request and so this isn't adding any extra DB overhead.

I agree accessing that cache is pretty nasty.

An overall comment I'd make is to remember I've tried to do this entire PR in the least intrusive way possible in an attempt to get it into a patch release. In hind-sight I think that there are too many flaws in the Manifest and in `ClassInfo` to make this realistic - so this is probably more a proof of concept now and should be moved to master to allow a greater re-write of the other parts of the system to accommodate case agnosticism.

dhensby:
I do think that a refactor of the major impact points for 4.0 is a good idea.

Is looping over all declared classes faster than just using a reflection class?

tractorcow:
I think just using the reflection class is the simplest way. :)

dhensby:
@tractorcow bump

kinglozzer:
Regression :cry:. Though it’s not really your fault, so I’ll forgive you :wink:.

`DataQuery` sometimes passes classes that don’t exist to `ClassInfo::subclassesFor()` (I’ll log a ticket for this). That used to be fine, it would just return an array only containing the class you pass in, but now it causes a fatal error. We can (and should) fix up `DataQuery`, but I’m a bit concerned that this may break other people’s code if they’re doing anything ~~equally stupid~~ similar. I’m not sure if this only applies to `ClassInfo::subclassesFor()`, or whether all `ClassInfo` methods simply didn’t care whether the class existed.

Perhaps we should amend `ClassInfo::class_name` to this (for now), just in case. Probably easier than trying to patch the individual `ClassInfo` methods:

```php
public static function class_name($nameOrObject) {
	} elseif (!class_exists($nameOrObject)) {
		return $nameOrObject;
	}
	$reflection = new ReflectionClass($nameOrObject);
	return $reflection->getName();
}
```

dhensby:
Interesting. That seems like a sensible fix so we should do it for 3.2 and possibly be stricter with 4?

PR #3950, opened 2015-02-27
DOCS Updated docs

micmania1:
The docs as they were would break websites and no warning was given regarding the security around doing this.


camfindlay:
Add DOCS prefix so core committers can find and merge these quickly :+1:

dhensby:
The key is naming the commit correctly as the change logs will search for the key words. This isn't a FIX its DOCS

micmania1:
Yep, I knew there was something. I just couldn't think what :)

PR #2855, opened 2014-02-17
Issue #2854: Permit ViewableData children to avoid caching 

g4b0:
Useful for frontend Session work. The issue: https://github.com/silverstripe/silverstripe-framework/issues/2854

PR #3955, opened 2015-03-01
Refactor ViewableData::obj caching

tractorcow:
To demonstrate alternative solution to #2855

Without introducing any new external api, this refactor allows user code to customise, disable, or enhance the viewable data obj caching.


dhensby:
Looks good, but as a new feature, shouldn't it be 3 branch? It's not a public API, but it's protected so can be overridden in child classes and thus is essentially a new API.

tractorcow:
Yeah I think so. Will move it to 3.

tractorcow:
Replaced with https://github.com/silverstripe/silverstripe-framework/pull/3985

PR #3956, opened 2015-03-02
FIX: DataQuery::applyRelation using incorrect foreign key (fixes #3954)

### LINKED ISSUE #3954: DataQuery applyRelation has_one join incorrect foreignKey
$has_one = array_flip($model->has_one());

I getting incorrect foreignKey while I have multiple has_one with same class.

e.g.

$has_one = array(
    'Owner' => 'Member',
    'Father' => 'Member',
    'Son' => 'Member'
);

It always getting the "Son" as foreignKey.

kinglozzer:
In the case of `has_one`, the foreign key is always the relation name!

### DIFF
--- a/tests/model/DataQueryTest.php
+++ b/tests/model/DataQueryTest.php
@@ -49,6 +49,19 @@ class DataQueryTest extends SapphireTest {
 			$dq->sql());
 	}
 
+	public function testApplyRelation() {
+		// Test applyRelation with two has_ones pointing to the same class
+		$dq = new DataQuery('DataQueryTest_B');
+		$dq->applyRelation('TestC');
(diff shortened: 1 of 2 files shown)


tests/model/DataQueryTest.php
```diff
@@ -241,7 +254,7 @@ class DataQueryTest_A extends DataObject implements TestOnly {
 	);
 
 	private static $has_one = array(
-		'TestC' => 'DataQueryTest_C',
+		'TestC' => 'DataQueryTest_C'
```
kinglozzer: CONVENTIONS! Will fix :)

dhensby: Trailing commas FTW!

dhensby:
Looks like a good fix!

PR #3959, opened 2015-03-03
DOCS Adding GitHub labels

dhensby:
As per my proposal to change GH labels.

### DIFF
--- a/docs/en/05_Contributing/01_Code.md
+++ b/docs/en/05_Contributing/01_Code.md
@@ -59,7 +59,43 @@ We ask for this so that the ownership in the license is clear and unambiguous, a
 8. Issue pull request on GitHub.  Visit your forked respoistory on GitHub.com and click the "Create Pull Request" button nex tot the new branch.
 
 The core team is then responsible for reviewing patches and deciding if they will make it into core.  If
-there are any problems they will follow up with you, so please ensure they have a way to contact you! 
+there are any problems they will follow up with you, so please ensure they have a way to contact you!
+
+### The Pull Request Process
+
+Once your pull request is issued, it's not the end of the road. A [core committer](/contributing/core_committers/) will most likely have some questions for you and may ask you to make some changes depending on discussions you have.
+If you've been naughty and not adhered to the coding conventions, expect a few requests to make changes so your code is in-line.
+
+If your change is particularly significant, it may be referred to the [mailing list](https://groups.google.com/forum/#!forum/silverstripe-dev) for further community discussion.
+
+A core committer will also "label" your PR using the labels defined in GitHub, these are to correctly classify and help find your work at a later date.
+
+#### GitHub Labels
+
+The current GitHub labels are grouped into 5 sections:
+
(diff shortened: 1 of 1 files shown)


camfindlay:
+1 Nice addition @dhensby :+1:

dhensby:
I'm not sure this is the best place for them to go but couldn't find anywhere more suitable.

It may be worth adding a section about guidelines for core commiters somewhere?

@camfindlay / @wilr

docs/en/05_Contributing/01_Code.md
```diff
@@ -59,7 +59,43 @@ We ask for this so that the ownership in the license is clear and unambiguous, a
...
+| change/patch | A change to go against the next patch release (eg: 3.1.x) |
+| impact/critical | This is a bug which breaks functionality in a way that can not be worked around and must be fixed |
+| impact/high | This is a bug which breaks functionality but can be mitigated by other code changes |
+| impact/medium | This is a bug which has unexpected behaviour but does not break functionality |
+| impact/low | This is a bug which is a nuisance but doesn't break any functionality (typos, etc) |
+| effort/easy | This is a bug that someone with limited SilverStripe experience could resolve |
+| effort/medium | This is a bug that someone with a good understanding of SilverStripe could resolve |
+| effort/hard | This is a bug that only an expert with SilverStripe could resolve |
+| type/docs | This is a docs change |
+| type/frontend | This is a change to front-end (CSS, HTML, etc) |
```
chillu: Could you reduce the repetition here? The "This is a" seems a bit redundant

docs/en/05_Contributing/01_Code.md
```diff
@@ -59,7 +59,43 @@ We ask for this so that the ownership in the license is clear and unambiguous, a
...
+| feedback-required/core-team | This issue is awaiting feedback from core team members |
```
chillu: Isn't that the default? I'd say the "feedback-required/author" is marginally useful, to filter for follow ups. But basically any open request which doesn't have "feedback-required/author" on it will require feedback or merge from the core team, right?

chillu:
Dan, can you confirm that you'll rename the existing labels once this is merged?

docs/en/05_Contributing/01_Code.md
dhensby: Agreed, I'll cut it down

docs/en/05_Contributing/01_Code.md
dhensby: I suppose that it's more for things that need group consideration, rather than a "default state" where some kind of review is needed.

Really, I only invisage them being used on PRs or issues with a long lifetime or where feedback needs to be collated, rather than switching them everytime someone replies

dhensby:
> can you confirm that you'll rename the existing labels once this is merged

Yep :)

dhensby:
Addressed those comments

chillu:
Thanks Dan!

dhensby:
Ok, I've added the labels now (to framework) renaming where appropriate - Colours were chosen by GitHub - I'm not great with colours and such so maybe someone could propose or amend them to be more suitable?

PR #3961, opened 2015-03-03
FIX: UploadField description alignment

kinglozzer:
Before:
screen shot 2015-03-03 at 09 18 37

After:
screen shot 2015-03-03 at 09 26 03


chillu:
Nice work!

PR #3962, opened 2015-03-03
FIX Prevent unnecessary call to config system which doesn't exist yet

micmania1:
By passing the Config instantiation through the Injector and service configuration we're unnecessarily calling Config::inst()->get() on a config system which doesn't yet exist.

More specifically it calls Config::inst()->get('Config', 'dependencies') here: https://github.com/silverstripe/silverstripe-framework/blob/4ae0d90c5565b2245fe7b9ff96d89e4e386f018c/control/injector/Injector.php#L660 which in turn calls Config::inst()->get('Config', 'extensions').

This jumps through hoops trying to find the cached config for this class using the current Config_LRU class. When it doesn't find it, it then tries to look through the manifests which don't get created or brought from cache until later on in Core.php. Eventually this returns null for both Config::inst()->get() calls.

I ran a quick performance test before and after on an ec2 t2.micro box using siege (concurrency 5, 10 repetitions) and saved 400ms on average response times on a default installation - it would be awesome if somebody could confirm this is actually the case?

### DIFF
--- a/core/Core.php
+++ b/core/Core.php
@@ -85,8 +85,6 @@ require_once 'control/injector/Injector.php';
 // Initialise the dependency injector as soon as possible, as it is 
 // subsequently used by some of the following code
 $injector = new Injector(array('locator' => 'SilverStripeServiceConfigurationLocator'));
-$injector->registerService(Config::inst());
-
 Injector::set_inst($injector);
 
 ///////////////////////////////////////////////////////////////////////////////


micmania1:
Before:

	Transactions:                      50 hits
	Availability:                 100.00 %
	Elapsed time:                  76.32 secs
	Data transferred:               0.07 MB
	Response time:                  6.61 secs
	Transaction rate:               0.66 trans/sec
	Throughput:                     0.00 MB/sec
	Concurrency:                    4.33
	Successful transactions:          50
	Failed transactions:               0
	Longest transaction:           10.84
	Shortest transaction:           2.76


After:

	Transactions:                      50 hits
	Availability:                 100.00 %
	Elapsed time:                  71.81 secs
	Data transferred:               0.07 MB
	Response time:                  6.20 secs
	Transaction rate:               0.70 trans/sec
	Throughput:                     0.00 MB/sec
	Concurrency:                    4.32
	Successful transactions:          50
	Failed transactions:               0
	Longest transaction:            8.89
	Shortest transaction:           2.78

dhensby:
It'd be interesting to see how significant the savings are once #3934 is merged

core/Core.php
```diff
@@ -85,10 +85,11 @@
 // Initialise the dependency injector as soon as possible, as it is 
 // subsequently used by some of the following code
 $injector = new Injector(array('locator' => 'SilverStripeServiceConfigurationLocator'));
-$injector->registerService(Config::inst());
-
 Injector::set_inst($injector);
 
+// Initialise Config
+Config::inst();
```
dhensby: I'm not sure why this is required as anywhere else that uses config will have to call this?

micmania1: I guess that's true. It felt a bit dirty anyway :)

dhensby:
Ok, after merging in the config_lru changes and commenting out line 88 from core.php I'm seeing improvements of about 4.5%

dhensby:
My point about the Config_LRU is that the impact of this fix will be less once the LRU changes are in because the slow LRU currently impacts this call

dhensby:
Well, considering this doesn't seem to break anything, I'd be happy to take it in.

nyeholt:
The 'dependencies' static declaration overlaps a fair amount with the config based 'properties' definition, but exists as a bit of syntactic sugar to avoid the need to explicitly declare those via config, and provides the inherited lookup of those. If that call to 'dependencies' is causing slowness, then I'd love to get rid of it :D Of course, that would be a backwards compatible breaker so release version would need to be later.

micmania1:
@nyeholt its not so much an issue with the dependencies static, its just that in this case its being called so early that the configuration has not yet loaded so its impossible to find a value for it. However, it continues to look for it which is quite expensive.

micmania1:
Anybody?

tractorcow:
I'm going to trust that the work you did and benchmarking demonstrates this is an improvement. I think the best course of action is to put this out in the wild and see if any other issues arise.

nyeholt:
This will be BC breaking for anyone that currently relies on binding the config in based on its existence as a named service; however, the likelihood of that being the case is very low given most people refer to the config via a direct static reference anyway.

PR #3964, opened 2015-03-03
Changed comparison from truthy to explicit

mobiusnz:
Changed comparison from truthy (==) to explicit (===) to avoid cases of `null == false`

### DIFF
--- a/forms/OptionsetField.php
+++ b/forms/OptionsetField.php
@@ -68,14 +68,18 @@ class OptionsetField extends DropdownField {
 				$odd = ($odd + 1) % 2;
 				$extraClass = $odd ? 'odd' : 'even';
 				$extraClass .= ' val' . preg_replace('/[^a-zA-Z0-9\-\_]/', '_', $value);
-				
+                if($value) {
+                    $checked = ($value == $this->value);
+                } else {
+                    $checked = ($value === $this->value) || (((string) $value) === ((string) $this->value));
+                }
 				$options[] = new ArrayData(array(
 					'ID' => $itemID,
 					'Class' => $extraClass,
 					'Name' => $this->name,
 					'Value' => $value,
 					'Title' => $title,
-					'isChecked' => $value === $this->value,
+					'isChecked' => $checked,
 					'isDisabled' => $this->disabled || in_array($value, $this->disabledItems),
 				));
 			}
--- a/forms/OptionsetField.php
+++ b/forms/OptionsetField.php
@@ -75,7 +75,7 @@ class OptionsetField extends DropdownField {
 					'Name' => $this->name,
 					'Value' => $value,
 					'Title' => $title,
-					'isChecked' => $value == $this->value,
+					'isChecked' => $value === $this->value,
 					'isDisabled' => $this->disabled || in_array($value, $this->disabledItems),
 				));
 			}


dhensby:
I'm worried by this change as it is quite subtle behaviour change.

I think that this needs test coverage to prove what was breaking before and we need to consider if this is an API change or a bug fix...

mobiusnz:
Hi @dhensby, 

I'm not exactly sure how to write a unit test that demonstrates this fix, but I can definitely show you this is a bug.

Consider this basic form:
````php
    public function Form() {
        return new Form($this, 'Form', new FieldList(array(
            new OptionsetField('Enum', 'Do we like using enums instead of booleans?', array(
                "Yes" => "Yes", "No" => "No")),
            new OptionsetField('Boolean', 'Do we like preselected optionsets?', array(
                true => "Yes", false => "No")))),
            new FieldList()
        );
    }
````

Notice that we haven't specified any defaults for either optionset. You would expect then, that the user should have to choose their preference for both, right? If I wanted one to be selected, I would have set a value.

Here's the result of the above form:

ss optionset bug

Another way to look at this, is can you think of an example where you *would* want non-explicit comparisons? I can't!

**Comment by tractorcow** [comment]
What if you are using IDs as keys, and the source object has (int) type keys but the posted back value treats them as (string)s?

Using false and true as keys are really risky in the first place. true == 'false' by the way.

**Comment by mobiusnz** [comment]
The main goals of this patch is to 

a) ensure true/false or 0/1 type keys don't end up with the negative preselected, and 
b) to be able to populate a `Boolean` data object field (ie so you can go `$form->saveInto($dataObject)`)

I found that the method for checking if an item should be selected is much more robust in `DropdownField`, so have copied that and applied it to `OptionsetField`

This should ensure consistency across SilverStripe Form objects

**Comment by tractorcow** [comment]
Why not just change the condition to

`($this->value !== null && $value == $this->value)`

**Comment by tractorcow** [comment]
You'll also need to include some unit tests here; Does this form work if not attached to a dataobject? Can posting back to itself save and restore it's own state?

What HTML does your above example generate?

**Comment by tractorcow** [comment]
Sorry if I'm sounding skeptical, I'm not, just making sure all possible use cases are covered. :)

**Comment by tractorcow** [comment]
Well I am a bit I guess :) haha

**Comment by mobiusnz** [comment]
I've made the isChecked test the same as in the parent DropdownField (except there it is isSelected). If we were to change to the test @tractorcow suggested we should look at changing the test in DropdownField as well, as inconsistencies within a framework is not something I enjoy.

As for the HTML generated, this is not changed. The only thing that has changed is the test for whether the "isChecked" option is true or false.

Casual testing with a dummy form indicates that the postback values are saved and loaded correctly.

I'm may need a bit of guidance on the tests sorry - this is my first pull request! I've looked at the existing tests for OptionsetField and they're very... sparse. It'd probably make sense to copy the tests from DropdownField given OptionsetField inherits from that, so I'll start there.

**Comment by tractorcow** [comment]
The trouble is that (string)false ends up as an empty string, and your code (using false as a key) is going to run into issues where a validator requires a value to be assigned, or an optional empty option is specified. Even (string)0 ends up a non-zero length string, which is what validators use to handle 0 as a valid key.

Maybe the code should be `$checked = strlen($this->value) && $value == $this->value;` to distinguish between:

 * '' no selection (null and empty strings fail strlen check)
 * '0' false selection ('0' == 0)
 * '1' true selection

Trying to use (boolean)false as a key means there's no distinction between no selection and false, since it evaluates to '' when cast as string.

Just to re-iterate, you should never be using literal true and false booleans as keys since they don't serialise well in forms. I'd go so far as to document that fact in the classdoc.

**Comment by dhensby** [comment]
> Trying to use (boolean)false as a key means there's no distinction between no selection and false, since it evaluates to '' when cast as string.
> 
> Just to re-iterate, you should never be using literal true and false booleans as keys since they don't serialise well in forms. I'd go so far as to document that fact in the classdoc.

^thus

**Comment by mobiusnz** [comment]
[Looking at the PHP documentation](http://nz.php.net/manual/en/language.types.array.php), booleans are casted to integers when used as a key. That will explain why the html generated gives `<input id="Form_Form_Boolean_0" class="radio" name="Boolean" type="radio" value="0">`

So there is no change in the serialisation of the form. The bug also exists if you use 0 and 1 as your keys, eg with:

```php
new OptionsetField('Boolean', 'Do we like preselected optionsets?', array(
                1 => "Yes", 0 => "No"))
```

My patch fixes this issue.

tractorcow:
Ok, I'll accept that internally php is casting it to a useful value. It should give you an indication though that booleans don't make good keys.

In any case, despite the fact that DropdownField logic looks awful, I'll accept it being substituted here over the existing logic. You'd need to ensure you also include the logic for the empty string value.

```php
// There has got to be a better way to do this
if($value === '' && ($this->value === '' || $this->value === null)) {
	$selected = true;
} else {
	// check against value, fallback to a type check comparison when !value
	if($value) {
		$selected = ($value == $this->value);
	} else {
		$selected = ($value === $this->value) || (((string) $value) === ((string) $this->value));
	}

	$this->isSelected = $selected;
}
```

dhensby:
I've been looking at these "option" fields recently and, despite class inheritance, there's very little shared logic, which I think should be improved.

I think that an explicit empty string should be denoted as "empty" (the empty string value" anything else can be a legitimate value but should be a string. Then we can do strict checking while casting as string

dhensby:
Also, you're using spaces and not tabs, that needs to be fixed

tractorcow:
@dhensby I'm rewriting these for 4.0 so it's in progress :)

sminnee:
@tractorcow how did you get on with the optionset updates? has this issue been addressed in 4?

tractorcow:
It was merged a while ago... in fact the final logic matches the one in this pr exactly.

https://github.com/silverstripe/silverstripe-framework/blob/master/forms/SelectField.php#L177

tractorcow:
Fixed in master.

PR #3966, opened 2015-03-04
Update FunctionalTest.php API notes

AntonyThorpe:
submitForm method needs the name of the button (or null) to work


dhensby:
too right, thanks!

PR #3969, opened 2015-03-04
FIX: Corrected padding on non-sortable columns.

sminnee:
span.non-sortable in the GridField header relies on there being other, sortable columns in the list to get the padding right. If there are no such columns (e.g. in a simple header), then the header is shown without padding, which is ugly.

This small change fixes that. Tested on Chrome and Firefox. 1px smaller than a sortable header on Firefox but still looks fine.


dhensby:
Looks like this needs rebasing. Also, any chance of before/after shots?

sminnee:
Rebasing? You mean to 3.2? I figured it's a bugfix so 3.1 is okay, but could do 3.2 if you think that's best.

Before: https://www.evernote.com/shard/s11/sh/78e917b0-b0a9-4a78-a3bf-6a4f60af5903/16cc8a11834d4e97115f81e2eb73a410

After: https://www.evernote.com/shard/s11/sh/7e3913ad-28bc-46dc-9d07-67f0250959a0/f6c8b25cfe37223581b46a10742ea518

dhensby:
No, sorry, against 3.1, We can't merge cleanly at the moment: HELP! No greed button

sminnee:
Doh, I missed that. Fixed now.

dhensby:
thanks :)

PR #3970, opened 2015-03-04
API enable PaginatedList to be disabled by setting page length to 0

tractorcow:
Useful for cases where paging size could be set / disabled via CMS settings


dhensby:
This looks like a very nice solution

PR #3971, opened 2015-03-05
BUG Fix serialised stateid exceeding request length

tractorcow:
cc @chillu 

For some reason it didn't work with `substr($str, 0,7)` but it's fine with 8 for some reason. :)

### DIFF
--- a/forms/gridfield/GridField.php
+++ b/forms/gridfield/GridField.php
@@ -840,7 +840,7 @@ class GridField_FormAction extends FormAction {
 			'args' => $this->args,
 		);
 
-		$id = md5(serialize($state));
+		$id = substr(md5(serialize($state)), 0, 8);
 		Session::set($id, $state);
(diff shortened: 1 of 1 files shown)


chillu:
Peer reviewed together with @madmatt, we think its a good solution. Less unique, but still enough variation, particularly since its for an existing user. Just waiting for Travis to pass.

tractorcow:
That's an obscure mysql error... restarting travis.

tractorcow:
Ah thanks. :)

tractorcow:
We found another problem with this fix.... md5 sometimes generates all-numeric keys, which then crashes when assigning to session::set, since it can't accept numeric indexes.

Fix is to use an md5 mechanism that generates alphabetic keys not alphanumeric.

tractorcow:
fix at https://github.com/silverstripe/silverstripe-framework/pull/3986

PR #3973, opened 2015-03-05
Remove incorrect L flag for Basic Auth RewriteRule

cjsewell:
As per the installer [.htaccess](https://github.com/silverstripe/silverstripe-installer/blob/3.1/.htaccess#L37) the RewriteRule to enable Basic Auth when using CGI/FastCGI with Apache does not have the L flag.


dhensby:
Thanks

Issue #3763, opened 2015-01-07
removeFieldFromTab creates new tab


sunnysideup:
In a project, there was a line 

```php
$fields->removeFieldFromTab("Root.Content", "Foo");
```

Foo did not exist on Root.Content, but it existed as Root.Foo (this was actually an upgrading issue - in 2.4 the line made sense, in 3.1 it did not).

As a result, however, a second "Content" tab was created. Obviously this should not happen.

dhensby:
This is happening because `removeFieldFromTab` uses `findOrMakeTab` to find the tab, which is clearly wrong. We need to just find the tab and to stop if we can't find it.

There needs to be a function for just finding the tab, instead of trying to make one

PR #3974, opened 2015-03-05
Fixes issue #3763; added findTab method to FieldList class

pedro2555:
When calling removeFieldFromTab() or removeFieldsFromTab() on FieldList class
if the tab did not exists it was created; as reported on issue#3763.

### DIFF
--- a/forms/FieldList.php
+++ b/forms/FieldList.php
@@ -270,12 +270,9 @@ class FieldList extends ArrayList {
 	 * @param string $tabName The tab to return, in the form "Tab.Subtab.Subsubtab".
 	 *   Caution: Does not recursively create TabSet instances, you need to make sure everything
 	 *   up until the last tab in the chain exists.
-	 * @param string $title Natural language title of the tab. If {@link $tabName} is passed in dot notation,
-	 *   the title parameter will only apply to the innermost referenced tab.
-	 *   The title is only changed if the tab doesn't exist already.
 	 * @return Tab The found or false if none found
 	 */
-	public function findTab($tabName, $title = null) {
+	public function findTab($tabName) {
 		// Backwards compatibility measure: Allow rewriting of outdated tab paths
 		$tabName = $this->rewriteTabPath($tabName);
 
--- a/forms/FieldList.php
+++ b/forms/FieldList.php
@@ -149,8 +149,9 @@ class FieldList extends ArrayList {
 		$this->flushFieldsCache();
 
 		// Find the tab
-		$tab = $this->findOrMakeTab($tabName);
-		$tab->removeByName($fieldName);
+		$tab = $this->findTab($tabName);
+		if ($tab)
+			$tab->removeByName($fieldName);
 	}
 	
 	/**
@@ -163,7 +164,10 @@ class FieldList extends ArrayList {
 		$this->flushFieldsCache();
 
 		// Find the tab
-		$tab = $this->findOrMakeTab($tabName);
+		$tab = $this->findTab($tabName);
+		
+		if (!$tab)
+			return false;
 		
 		// Add the fields to the end of this set
 		foreach($fields as $field) $tab->removeByName($field);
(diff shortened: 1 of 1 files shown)


pedro2555:
Mention #3763

forms/FieldList.php
```diff
@@ -260,6 +264,38 @@ public function hasTabSet() {
...
+	 * @param string $title Natural language title of the tab. If {@link $tabName} is passed in dot notation,
+	 *   the title parameter will only apply to the innermost referenced tab.
+	 *   The title is only changed if the tab doesn't exist already.
+	 * @return Tab The found or false if none found
+	 */
+	public function findTab($tabName, $title = null) {
+		// Backwards compatibility measure: Allow rewriting of outdated tab paths
+		$tabName = $this->rewriteTabPath($tabName);
+
+		$parts = explode('.',$tabName);
```
dhensby: space after the comma please

forms/FieldList.php
```diff
@@ -149,8 +149,9 @@ public function removeFieldFromTab($tabName, $fieldName) {
 		$this->flushFieldsCache();
 
 		// Find the tab
-		$tab = $this->findOrMakeTab($tabName);
-		$tab->removeByName($fieldName);
+		$tab = $this->findTab($tabName);
+		if ($tab)
+			$tab->removeByName($fieldName);
```
dhensby: this should either be wrapped in curly braces or on one line (I'd prefer the former)

forms/FieldList.php
```diff
@@ -163,7 +164,10 @@ public function removeFieldsFromTab($tabName, $fields) {
 		$this->flushFieldsCache();
 
 		// Find the tab
-		$tab = $this->findOrMakeTab($tabName);
+		$tab = $this->findTab($tabName);
+		
+		if (!$tab)
```
dhensby: Please don't return false here. Can you just wrap the foreach in a positive assertion on `$tab`?

eg: `if ($tab) foreach($fields as $field) $tab->removeByName($field);`

dhensby:
Thanks for this, it's been bugging me for a long time but I'd slightly forgotten about it.

Please can you add a unit test or two to make sure this function has test coverage.

Also, it's a bit of code duplication with `findormaketab` - would it be possible to utilise this function there?

pedro2555:
@dhensby Yeah, I think it is one of those lazy things no one wants to do.

Haven't got around unit testing in PHP yet, I will soon.

Yes, `findTab` is basically a copy-paste of `findOrMakeTab`, probably we could use `findTab` in `findOrMakeTab`, but we could also make `findOrMakeTab` recursive which is suggested in the comments. I haven't really looked at it that much so I will not, for now, take any decision on that matter.

You may leave this PR hanging until the points above are decided/fixed.

dhensby:
A unit test for this will be required for it to get in. If you're not confident writing one, I can do it for you, but I'd recommend you have a look at writing one - it's quite simple to do and a good ability to have.

forms/FieldList.php
```diff
@@ -260,6 +264,38 @@ public function hasTabSet() {
...
+	 *
+	 * @param string $tabName The tab to return, in the form "Tab.Subtab.Subsubtab".
+	 *   Caution: Does not recursively create TabSet instances, you need to make sure everything
+	 *   up until the last tab in the chain exists.
+	 * @param string $title Natural language title of the tab. If {@link $tabName} is passed in dot notation,
+	 *   the title parameter will only apply to the innermost referenced tab.
+	 *   The title is only changed if the tab doesn't exist already.
+	 * @return Tab The found or false if none found
+	 */
+	public function findTab($tabName, $title = null) {
```
dhensby: The $title param isn't used, it should be removed.

pedro2555:
@dhensby Yeah, I know. Just haven't yet had the time to learn it in PHP. I've tried to install PHPUnit without composer, but no success. I don't like composer, although I understand it's advantages.

If you want, you can write one for me, I will appreciate that for sure. And I will eventually look at testing in PHP, it's not a matter of if but when.

sminnee:
I believe that I was looking at a pull request for @patricknelson that also implemented this. Do we have a dupe?

patricknelson:
@sminnee You are really making the rounds! Thank you. Indeed, this is a much in demand feature. Current functionality is whacko. The PR you speak of is here: https://github.com/silverstripe/silverstripe-framework/pull/4122

A comment on this variation: This method should **abssolutely not** be returning anything but a `Tab`, nor nothing (`null`). It really should not change type, and it should document this fact in its `@return`.

To my fault: I should have had the foresight to add unit tests on mine, given I know it's expected (rightfully so) and these PR's have a tendency to sit and age for a bit. Thankfully mine (being a feature) is base on `3`). This branch is waaaay outdated anyway and wouldn't be suitable for a new feature as I have learned in my wise old age.

sminnee:
Cool, in that case I'm going to close this as a duplicate of #4122.

PR #3975, opened 2015-03-05
DOCS: Correct & expand routing docs

oddnoc:
- Correct syntax error in TeamController sample code
- Explain mandatory use case of $url_handlers
- Provide sample code for mandatory use case of $url_handlers

### DIFF
--- a/docs/en/02_Developer_Guides/02_Controllers/02_Routing.md
+++ b/docs/en/02_Developer_Guides/02_Controllers/02_Routing.md
@@ -135,6 +135,14 @@ start parsing variables and the appropriate controller action AFTER the `//`).
 
 ## URL Handlers
 
+<div class="alert" markdown="1">
+You **must** use the **$url_handlers** static array described here if your URL
+pattern does not use the Controller class's default pattern of
+`$Action//$ID/$OtherID`. If you fail to do so, and your pattern has more than
+2 parameters, your controller will throw the error "I can't handle sub-URLs of
+a *class name* object" with HTTP status 404.
+</div>
+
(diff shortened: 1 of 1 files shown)


docs/en/02_Developer_Guides/02_Controllers/02_Routing.md
```diff
@@ -154,12 +162,35 @@ This is useful when you want to provide custom actions for the mapping of `teams
...
+class specifies the URL pattern in `$url_handlers`. Notice that it defines 5
+parameters.
+
+
+	:::php
+	class FeedController extends ContentController {
+
+		private static $allowed_actions = array('Go');
+		private static $url_handlers = array(
+			'Go/$UserName/$AuthToken/$Timestamp/$OutputType/$DeleteMode' => 'Go'
```
dhensby: I like the idea of a "real world" example, but this seems quite obscure and lacking context.

Could you either provide some context (perhaps as simple as adding an action) or use a concocted example which is a bit more obvious?

dhensby:
Thanks. Can you squash these please?

```bash
$ git reset HEAD^
$ git commit --amend --no-edit
$ git push -f
```

oddnoc:
All squashed! (BTW, I do discrete commits to make discussion easier, with the intention of squashing at the end. Let me know if you think that's unnecessary.)

dhensby:
> I do discrete commits to make discussion easier, with the intention of squashing at the end

Completely agree with that approach, I just add the commands to help as I don't know who does and doesn't know how to do this out of the box :)

dhensby:
Thanks!

PR #3978, opened 2015-03-06
API Require length of Varchar to be set fixes #3681

### LINKED ISSUE #3681: Default Varchar size is 50 - should default to 255
Default Varchar size is 50 - should it not default to 255?

Or is it has to be 50 then it needs to be mentioned a lot better in the docs.

dhensby:
The default varchar size of 50 is arbitrary and, for those that don't know about it, confusing.

This change forces users to be explicit and therefore more clear about how large the varchar is.

### DIFF
--- a/model/fieldtypes/Varchar.php
+++ b/model/fieldtypes/Varchar.php
@@ -25,9 +25,11 @@ class Varchar extends StringField {
  	 * @param $size int The maximum size of the field, in terms of characters
  	 * @param $options array Optional parameters, e.g. array("nullifyEmpty"=>false).
  	 *                       See {@link StringField::setOptions()} for information on the available options
- 	 * @return unknown_type
  	 */
-	public function __construct($name = null, $size = 50, $options = array()) {
+	public function __construct($name = null, $size = null, $options = array()) {
+		if ($size === null) {
+			Deprecation::notice('3.2', 'The parameter $size is now required, please declare it');
+		}
 		$this->size = $size ? $size : 50;
 		parent::__construct($name, $options);
 	}


dhensby:
Oh gosh, maybe this is a terrible idea given DBField::create_field('Varchar', $val); will likely throw the deprecation error.

Perhaps we just need to advise people to define it, rather than force a number to be given

dhensby:
Closing this as it's not realistic for now

willmorgan:
@dhensby a good candidate for 4, don't you think?

dhensby:
The problem here is that you should be able to instantiate the abstract notion of a Varchar in code without having to define it's length.

So throwing some kind of error on construct for that is a pain and unhelpful. We'd need to separate out the notion of a string field from a DB varchar type.

sminnee:
We could make the default 255?

dhensby:
I was hoping to make the definition of the length to be mandatory 

Whilst 255 is traditionally the limit for vaechars in DB, its quite a bit larger now (in mysql at least)

sminnee:
You'll still have the problem when, for example, choosing the length of the SiteTree.Title field. I don't think refusing to set a default is going to do users any favours.

Let's pick a default that we can be comfortable with, and put that in instead of 50.

dhensby:
Having no standard means devs have to be explicit - it means removing ambiguity and I'm in favour of that.

If we change the default from 50 to 255 it'll be even more confusing than it is now (in that people will think it's 255 or 50 when it's not, depending on the version of SS they are using).

I'd probably vote to leave it at 50, at least that's so short people are practically forced to define a limit anyway

PR #3979, opened 2015-03-06
Nest and unnest Config and Controller for each test

dhensby:
This stops devs having to nest and unnest for their own tests

### DIFF
--- a/dev/SapphireTest.php
+++ b/dev/SapphireTest.php
@@ -5,12 +5,12 @@ require_once 'TestRunner.php';
  * Test case class for the Sapphire framework.
  * Sapphire unit testing is based on PHPUnit, but provides a number of hooks into our data model that make it easier
  * to work with.
- * 
+ *
  * @package framework
  * @subpackage testing
  */
 class SapphireTest extends PHPUnit_Framework_TestCase {
-	
+
 	/** @config */
 	private static $dependencies = array(
 		'fixtureFactory' => '%$FixtureFactory',
@@ -21,7 +21,7 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 	 * If passed as an array, multiple fixture files will be loaded.
 	 * Please note that you won't be able to refer with "=>" notation
 	 * between the fixtures, they act independent of each other.
-	 * 
+	 *
 	 * @var string|array
 	 */
 	protected static $fixture_file = null;
@@ -30,19 +30,19 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 	 * @var FixtureFactory
 	 */
 	protected $fixtureFactory;
-	
+
 	/**
 	 * @var bool Set whether to include this test in the TestRunner or to skip this.
 	 */
 	protected $skipTest = false;
-	
+
 	/**
 	 * @var Boolean If set to TRUE, this will force a test database to be generated
-	 * in {@link setUp()}. Note that this flag is overruled by the presence of a 
+	 * in {@link setUp()}. Note that this flag is overruled by the presence of a
 	 * {@link $fixture_file}, which always forces a database build.
 	 */
 	protected $usesDatabase = null;
-	
+
 	protected $originalMailer;
 	protected $originalMemberPasswordValidator;
 	protected $originalRequirements;
@@ -50,33 +50,33 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 	protected $originalTheme;
 	protected $originalNestedURLsState;
 	protected $originalMemoryLimit;
-	
+
 	protected $mailer;
-	
+
 	/**
 	 * Pointer to the manifest that isn't a test manifest
 	 */
 	protected static $regular_manifest;
-	
+
 	/**
 	 * @var boolean
 	 */
 	protected static $is_running_test = false;
-	
+
 	protected static $test_class_manifest;
-	
+
 	/**
 	 * By default, setUp() does not require default records. Pass
 	 * class names in here, and the require/augment default records
 	 * function will be called on them.
 	 */
 	protected $requireDefaultRecordsFrom = array();
-	
-	
+
+
 	/**
 	 * A list of extensions that can't be applied during the execution of this run.  If they are
 	 * applied, they will be temporarily removed and a database migration called.
-	 * 
+	 *
 	 * The keys of the are the classes that the extensions can't be applied the extensions to, and
 	 * the values are an array of illegal extensions on that class.
 	 */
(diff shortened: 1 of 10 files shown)


kinglozzer:
:+1:. Looks like the failures are the same config test issue I encountered in #3132 (see 2nd commit) :)

dhensby:
That test is really odd... The test asserts that it is null at the top and then later asserts it doesn't contain a string... I don't get why this breaks with the change I made...

dhensby:
Oh. No it doesn't, I'm just not reading it properly... So it's testing on a config value set in another test? Odd. I'll amend that test, then

dhensby:
Thanks @kinglozzer I cherry-picked your commit out :)

dhensby:
See the diff without whitespace changes: https://github.com/silverstripe/silverstripe-framework/pull/3979/files?w=1

dhensby:
Looks like this improves build times and memory usage for tests.

It also will make the test suite more stable as people won't have to worry about Config leaks.

I think I should update the testing docs to make sure it's clear that each test is run in a "sandbox" for config/injector.

dhensby:
Hmm, should probably nest the tests on the `setUpOnce` and `tearDownOnce` functions too..

Then a dev can set a global config setting for all their tests (via the once).

tractorcow:
I think this needs upgrading notes in the 3.1.13.md file. See (paraphrased)

```markdown
## Upgrading

If you are writing test cases and used nest/unnest in injector and config to restore the testing state, you don't need to anymore!
```

Also, do you think we should do this in 3.2 instead? It seems to be a stability increase, rather than a bugfix, so it's probably more appropriate there.

dhensby:
Adding upgrading notes.

> Also, do you think we should do this in 3.2 instead?

I think it's part of the test suite so not applicable to semver. I think it's also nice to get it in sooner rather than later.

docs/en/04_Changelogs/3.1.13.md
```diff
@@ -0,0 +1,4 @@
+## Upgrading
+
+If you are writing test cases and used nest/unnest in injector and config to restore the testing state, you don't need
```
tractorcow: I was paraphrasing because I was / am lazy and tired. Do you feel like elaborating? Perhaps say what nesting does, why it was necessary and why it's no longer?

tractorcow:
Have you run the cms tests against this?

docs/en/04_Changelogs/3.1.13.md
dhensby: Oh :(

I added lots of docs to the... docs...

Ok, I'll think of something more inspiring to add

dhensby:
> Have you run the cms tests against this?

Good Q, it's been a while so I can't remember - I think so. I'll do it again

tractorcow:
Just in case merging this suddenly breaks those tests. :) I don't think it will but just to be sure.

dhensby:
@tractorcow I can confirm that all CMS tests pass:

```sh
[root@localhost test]# vendor/bin/phpunit cms/tests
PHPUnit 3.7.38 by Sebastian Bergmann.

Configuration read from /var/www/test/phpunit.xml.dist

....I..........................................................  63 / 186 ( 33%)
............................................................... 126 / 186 ( 67%)
............................................................

Time: 5.52 minutes, Memory: 94.75Mb

OK, but incomplete or skipped tests!
Tests: 186, Assertions: 776, Incomplete: 1.
```

dhensby:
Travis is failing on 2 of the builds because composer won't install properly... but it does pass, I promise....

tractorcow:
I want to merge this but I'm a bit paranoid of test performance at the moment, while I'm trying to get 3.2 polished and ready. I think I'd like to test this PR on the current 3.2 build to ensure it doesn't introduce any significant performance regressions.

tractorcow:
True, but it also has the risk of introducing some more significant ones. :D

for instance, 

```php

function setUp() {
parent::setUp();
Config::nest();
}

function tearDown() {
Config::unnest();
}
```

As terrible as the code is (tearDown forgets to call the parent), it runs the risk of having additional levels of nesting. Maybe we should have some kind of `brokenOnWrite` protection for setup and tear down?

dhensby:
I'll run a test with it cherry-picked onto 3.2

Are you saying you're worried someone nest and nests themselves whilst failing to call the `parent::tearDown()` - therefore causing massive nesting?

I don't think it's our place to worry about users failing to do that; we make sure our tests are all legit and if they do it wrong, then they'll get performance issues.

I suppose we need to check that all our tests DO call parent methods, but I assume other issues would have arisen if we weren't doing it properly..

dhensby:
My results from cherry-pick onto 3.2:

```sh
[root@localhost test-3.2]# vendor/bin/phpunit framework/tests
PHPUnit 3.7.38 by Sebastian Bergmann.

Configuration read from /var/www/test-3.2/phpunit.xml.dist

.............................................................   61 / 1581 (  3%)
...................................I.........................  122 / 1581 (  7%)
.........................I...................................  183 / 1581 ( 11%)
..I..........................................................  244 / 1581 ( 15%)
.............................................................  305 / 1581 ( 19%)
.............................................................  366 / 1581 ( 23%)
.............................................................  427 / 1581 ( 27%)
.............................................................  488 / 1581 ( 30%)
.............................................................  549 / 1581 ( 34%)
................................I............................  610 / 1581 ( 38%)
.............................................................  671 / 1581 ( 42%)
.............................................................  732 / 1581 ( 46%)
.............................................................  793 / 1581 ( 50%)
.............................................................  854 / 1581 ( 54%)
.............................................................  915 / 1581 ( 57%)
.............................................................  976 / 1581 ( 61%)
............................................................. 1037 / 1581 ( 65%)
............................................................. 1098 / 1581 ( 69%)
............................................................. 1159 / 1581 ( 73%)
....SSSSSSSSSSSSSSSSS........................................ 1220 / 1581 ( 77%)
.....SS...................................................... 1281 / 1581 ( 81%)
............................................................. 1342 / 1581 ( 84%)
............................................................. 1403 / 1581 ( 88%)
..................................................I.......... 1464 / 1581 ( 92%)
............................................................. 1525 / 1581 ( 96%)
.......................................................

Time: 6.75 minutes, Memory: 144.25Mb

OK, but incomplete or skipped tests!
Tests: 1580, Assertions: 6473, Incomplete: 5, Skipped: 19.
```

However, CMS on 3.2 seg faulted...

```sh
[root@localhost test-3.2]# vendor/bin/phpunit cms/tests
PHPUnit 3.7.38 by Sebastian Bergmann.

Configuration read from /var/www/test-3.2/phpunit.xml.dist

.......I.......................................................  63 / 187 ( 33%)
.....Segmentation fault
```

tractorcow:
Segfault issues are fixed with https://github.com/silverstripe/silverstripe-framework/pull/4288

I would cautiously recommend rebasing on 3.2 and merging it there; I can put it into 3.2 beta 1 if you want.

tractorcow:
Ok, let's see how successful we are with this in 3.1. Tests pass; Code seems ok. :D

dhensby:
whoop. you won't regret this!

tractorcow:
I regret this. Lots of code is breaking and I've traced it to the removal of the code in `tearDownOnce`. It's not enough to unnest() the config; `Object::$extra_methods` needs to be flushed.

PR #3980, opened 2015-03-07
FEATURE: Added global maximum upload file size by type

Turnerj:
Re-submitted my original PR due to a few screw ups on my part (not branching properly before pushing it). The documentation now is in the right spot since the major documentation rework.

Original: https://github.com/silverstripe/silverstripe-framework/pull/3273

From my original PR:

-----
Adds the ability to specify the max file size for a type via YAML config.

```yaml
Upload_Validator:
  default_max_file_size:
    '[image]': '1m'
    '[doc]': '5m'
    'jpeg': 2000
```

These settings can still be updated on the instance variable via the method `setAllowedMaxFileSize`.

You can specify an App Category to limit the max file for rather than each individual file extension. This uses the built-in `app_categories` list. To specify an App Category in the config, just use the square brackets `[` and `]` around the name. This is to differentiate between an extension being limited and an entire category being limited.

With this update, it also adds support for `File::ini2bytes` when setting the allowed max file size.

### DIFF
--- a/filesystem/Upload.php
+++ b/filesystem/Upload.php
@@ -306,6 +306,16 @@ class Upload extends Controller {
 class Upload_Validator {
 
 	/**
+	* Contains a list of the max file sizes shared by
+	* all upload fields. This is then duplicated into the
+	* "allowedMaxFileSize" instance property on construct.
+	*
+	* @config
+	* @var array
+	*/
+	private static $default_max_file_size = array();
+
+	/**
 	 * Information about the temporary file produced
 	 * by the PHP-runtime.
 	 *
@@ -360,22 +370,46 @@ class Upload_Validator {
 	 * @return int Filesize in bytes
 	 */
 	public function getAllowedMaxFileSize($ext = null) {
+		
+		// Check if there is any defined instance max file sizes
+		if (empty($this->allowedMaxFileSize)) {
+			// Set default max file sizes if there isn't
+			$fileSize = Config::inst()->get('Upload_Validator', 'default_max_file_size');
+			if (isset($fileSize)) {
+				$this->setAllowedMaxFileSize($fileSize);
+			} else {
+				// When no default is present, use maximum set by PHP
+				$maxUpload = File::ini2bytes(ini_get('upload_max_filesize'));
+				$maxPost = File::ini2bytes(ini_get('post_max_size'));
+				$this->setAllowedMaxFileSize(min($maxUpload, $maxPost));
+			}
+		}
+		
 		$ext = strtolower($ext);
-		if(isset($ext) && isset($this->allowedMaxFileSize[$ext])) {
-			return $this->allowedMaxFileSize[$ext];   
+		if ($ext) {
+			if (isset($this->allowedMaxFileSize[$ext])) {
+				return $this->allowedMaxFileSize[$ext];
+			}
+			
+			$category = File::get_app_category($ext);
+			if ($category && isset($this->allowedMaxFileSize['[' . $category . ']'])) {
+				return $this->allowedMaxFileSize['[' . $category . ']'];
+			}
+			
+			return false;
 		} else {
 			return (isset($this->allowedMaxFileSize['*'])) ? $this->allowedMaxFileSize['*'] : false;
 		}
 	}
 	
 	/**
-	 * Set filesize maximums (in bytes).
+	 * Set filesize maximums (in bytes or INI format).
 	 * Automatically converts extensions to lowercase
 	 * for easier matching.
 	 * 
 	 * Example: 
 	 * <code>
-	 * array('*' => 200, 'jpg' => 1000)
+	 * array('*' => 200, 'jpg' => 1000, '[doc]' => '5m')
 	 * </code>
 	 *
 	 * @param array|int $rules
@@ -384,7 +418,22 @@ class Upload_Validator {
 		if(is_array($rules) && count($rules)) {
 			// make sure all extensions are lowercase
 			$rules = array_change_key_case($rules, CASE_LOWER);
-			$this->allowedMaxFileSize = $rules;
+			$finalRules = array();
+			$tmpSize = 0;
+			
+			foreach ($rules as $rule => $value) {
+				if (is_numeric($value)) {
+					$tmpSize = $value;
+				} else {
+					$tmpSize = File::ini2bytes($value);
+				}
+			
+				$finalRules[$rule] = (int)$tmpSize;
+			}
+			
+			$this->allowedMaxFileSize = $finalRules;
+		} elseif(is_string($rules)) {
+			$this->allowedMaxFileSize['*'] = File::ini2bytes($rules);
 		} elseif((int) $rules > 0) {
 			$this->allowedMaxFileSize['*'] = (int)$rules;
 		}
--- a/tests/filesystem/UploadTest.php
+++ b/tests/filesystem/UploadTest.php
@@ -84,15 +84,100 @@ class UploadTest extends SapphireTest {
 			'error' => UPLOAD_ERR_OK,
 		);
 		
-		$v = new UploadTest_Validator();
-		$v->setAllowedMaxFileSize(array('txt' => 10));
-		
 		// test upload into default folder
 		$u1 = new Upload();
+		$v = new UploadTest_Validator();
+		
+		$v->setAllowedMaxFileSize(array('txt' => 10));
 		$u1->setValidator($v);
 		$result = $u1->load($tmpFile);
-		
 		$this->assertFalse($result, 'Load failed because size was too big');
+		
+		$v->setAllowedMaxFileSize(array('[doc]' => 10));
+		$u1->setValidator($v);
+		$result = $u1->load($tmpFile);
+		$this->assertFalse($result, 'Load failed because size was too big');
+		
+		$v->setAllowedMaxFileSize(array('txt' => 200000));
+		$u1->setValidator($v);
+		$result = $u1->load($tmpFile);
+		$this->assertTrue($result, 'Load failed with setting max file size');
+		
+		// check max file size set by app category
+		$tmpFileName = 'UploadTest-testUpload.jpg';
+		$tmpFilePath = TEMP_FOLDER . '/' . $tmpFileName;
+		file_put_contents($tmpFilePath, $tmpFileContent . $tmpFileContent);
+		
+		$tmpFile = array(
+			'name' => $tmpFileName,
+			'type' => 'image/jpeg',
+			'size' => filesize($tmpFilePath),
+			'tmp_name' => $tmpFilePath,
+			'extension' => 'jpg',
+			'error' => UPLOAD_ERR_OK,
+		);
+		
+		$v->setAllowedMaxFileSize(array('[image]' => '40k'));
+		$u1->setValidator($v);
+		$result = $u1->load($tmpFile);
+		$this->assertTrue($result, 'Load failed with setting max file size');
+		
+		$v->setAllowedMaxFileSize(array('[image]' => '1k'));
+		$u1->setValidator($v);
+		$result = $u1->load($tmpFile);
+		$this->assertFalse($result, 'Load failed because size was too big');
+		
+		$v->setAllowedMaxFileSize(array('[image]' => 1000));
+		$u1->setValidator($v);
+		$result = $u1->load($tmpFile);
+		$this->assertFalse($result, 'Load failed because size was too big');
+	}
+	
+	public function testGetAllowedMaxFileSize() {
+		Config::nest();
+		
+		// Check the max file size uses the config values
+		$configMaxFileSizes = array(
+			'[image]' => '1k',
+			'txt' => 1000
+		);
+		Config::inst()->update('Upload_Validator', 'default_max_file_size', $configMaxFileSizes);
+		$v = new UploadTest_Validator();
+		
+		$retrievedSize = $v->getAllowedMaxFileSize('[image]');
+		$this->assertEquals(1024, $retrievedSize, 'Max file size check on default values failed (config category set check)');
+		
+		$retrievedSize = $v->getAllowedMaxFileSize('txt');
+		$this->assertEquals(1000, $retrievedSize, 'Max file size check on default values failed (config extension set check)');
+		
+		// Check instance values for max file size
+		$maxFileSizes = array(
+			'[doc]' => 2000,
+			'txt' => '4k'
+		);
+		$v = new UploadTest_Validator();
+		$v->setAllowedMaxFileSize($maxFileSizes);
+		
+		$retrievedSize = $v->getAllowedMaxFileSize('[doc]');
+		$this->assertEquals(2000, $retrievedSize, 'Max file size check on instance values failed (instance category set check)');
+		
+		// Check that the instance values overwrote the default values
+		// ie. The max file size will not exist for [image]
+		$retrievedSize = $v->getAllowedMaxFileSize('[image]');
+		$this->assertFalse($retrievedSize, 'Max file size check on instance values failed (config overridden check)');
+		
+		// Check a category that has not been set before
+		$retrievedSize = $v->getAllowedMaxFileSize('[zip]');
+		$this->assertFalse($retrievedSize, 'Max file size check on instance values failed (category not set check)');
+		
+		// Check a file extension that has not been set before
+		$retrievedSize = $v->getAllowedMaxFileSize('mp3');
+		$this->assertFalse($retrievedSize, 'Max file size check on instance values failed (extension not set check)');
+		
+		$retrievedSize = $v->getAllowedMaxFileSize('txt');
+		$this->assertEquals(4096, $retrievedSize, 'Max file size check on instance values failed (instance extension set check)');
+		
+		Config::unnest();
 	}
 
 	public function testAllowedSizeOnFileWithNoExtension() {
--- a/docs/en/02_Developer_Guides/03_Forms/Field_types/05_UploadField.md
+++ b/docs/en/02_Developer_Guides/03_Forms/Field_types/05_UploadField.md
@@ -148,6 +148,16 @@ NOTE: this only sets the configuration for your UploadField, this does NOT chang
 	$this->getValidator()->setAllowedMaxFileSize($size);
 ```
 
+You can also specify a default global maximum file size setting in your config for different file types. This is overridden when specifying the max allowed file size on the UploadField instance.
+
+```yaml
+	Upload_Validator: 
+	  default_max_file_size: 
+	    '[image]': '1m'
+	    '[doc]': '5m'
+	    'jpeg': 2000
+```
+
 ### Preview dimensions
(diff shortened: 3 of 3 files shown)


dhensby:
Squash them up please :)

```bash
$ git reset HEAD^
$ git commit -a --amend --no-edit
$ git push -f
```

dhensby:
So what happens if the PHP max upload size is lower than the mas size declared in the YML?

dhensby:
Also, unfortunately, as a new feature this will need to go against branch 3.

I'll merge it into that branch to save you the hassle.

Turnerj:
Now squashed :)

PHP throws a warning when the PHP max upload size is lower than what is declared in the YML file. This would be the same behaviour it had before my PR and doesn't seem to be a code way to get around this (as you can't set the PHP max upload size during run time).

If it was possible to override, I would be tempted to do that in my PR.

Thanks for that, I don't think I can change the target branch after already declaring a PR against it.

chillu:
I know this feature went through a bit of core team scrutiny already (thanks for being so thorough @Turnerj!), but can somebody remind me why this should be in core in the first place? I'm failing to come up with a case from my many SilverStripe projects where this would've come in handy. Maybe we should just make it easier to override `Upload_Validator`? /cc @halkyon @dhensby @tractorcow

Turnerj:
You do have a point @chillu and I did actually [have it as a module](https://github.com/Turnerj/silverstripe-formfield-plus/blob/master/code/validators/UploadPlusValidator.php) however it wasn't without its complications. I do understand that there is work to split more of out of the core so there may be a little questioning for new features to be added in.

While saying that, I feel that the specific functionality that this PR adds doesn't make sense in a module. The core provides the Upload field with controls for max file size already. If the Upload field/validator or max file size controls were outside of the core (in a module), it would be worthwhile updating such a module with the changes in this PR.

Perhaps the Upload field itself doesn't belong in the framework anymore?

tractorcow:
I don't think it hurts to have small and useful improvements to the default validation, if it makes sense.

However this will need to go into the 3 branch since it's an API improvement and not a bugfix. :) Good job though.

filesystem/Upload.php
```diff
@@ -306,6 +306,15 @@ public function getErrors() {
 class Upload_Validator {
 
 	/**
+	* Contains a list of the max file sizes shared by
+	* all upload fields. This is then duplicated into the
+	* "allowedMaxFileSize" instance property on construct.
+	*
+	* @var array
+	*/
+	private static $default_max_file_size = array();
```
dhensby: I think this variable naming is confusing. It's actually default max file sizes (by extension)

tests/forms/uploadfield/UploadFieldTest.php
```diff
@@ -167,7 +167,9 @@ public function testAllowedExtensions() {
 			'UploadFieldTest_Controller/Form/field/AllowedExtensionsField/upload',
 			array('AllowedExtensionsField' => $this->getUploadFile($invalidFile))
 		);
-		$this->assertTrue($response->isError());
+		$responseJson = Convert::json2array($response->getBody());
+		
+		$this->assertTrue($response->isError() || isset($responseJson[0]['error']));
```
dhensby: I think this kind of assertion is not a good idea. Putting OR logic in here means that we never have any guarantee that the last condition is ever true because as soon as the first one is the second is ignored.

In a unit test you should know the expected outcome and test that, there shouldn't be any chance about this assertion.

Therefore, either both should be individually asserted or the one that doesn't happen should be removed (and perhaps assertFalse instead)

dhensby:
Right, it wasn't entirely clear to me that this isn't an entire new feature, it's an enhancement to an existing feature.

Given that setting the max file size per extension already exists, being able to do it by type (image, document, etc) is a very welcome enhancement, in my view.

@chillu whilst I agree there is little need to this feature in general, whilst it's in core we should accept sensible enhancements, which this definitely is, and discussion around removing the feature itself should be taken somewhere else.

tests/forms/uploadfield/UploadFieldTest.php
Turnerj: Yeah, that makes sense. Given one of the other files in the commit actually removed setting the error status code (that `$response->isError()` would pick up), in terms of the test the `isset` check is the valid one. Will fix that in another commit shortly.

dhensby: Thanks, same for line 183 :)

dhensby:
@Turnerj ok, if the code is changing responses from being HTTP errors to JSON, then I think this actually has to go in master as it's a breaking change for anyone relying on those HTTP headers.

Can you either set it back to use HTTP errors or we'll need to put it into master

Turnerj:
I think maybe it does need to go against the master then (is that something you can change on your end or need me to open another PR with a different target?). The reason for removing the HTTP errors being set was that broke prior functionality (https://github.com/silverstripe/silverstripe-framework/commit/bdd936d078f685c795de00f354fb26eccf558745) where the CMS was presented with the message "Forbidden" rather than the detailed error.

dhensby:
Ok, patch it up and we'll merge it into master

Turnerj:
Test case has been updated and commits re-squashed. Let me know if there is anything else you need me to do from my end.

tests/filesystem/UploadTest.php
```diff
@@ -93,6 +93,30 @@ public function testAllowedFilesize() {
...
+			'size' => filesize($tmpFilePath),
+			'tmp_name' => $tmpFilePath,
+			'extension' => 'jpg',
+			'error' => UPLOAD_ERR_OK,
+		);
+		
+		$v->setAllowedMaxFileSize(array('[image]' => '40k'));
+		$u1->setValidator($v);
+		$result = $u1->load($tmpFile);
+		$this->assertTrue($result, 'Load passed because file size was under limit');
```
dhensby: The last parameter to `assert` calls should be a message presented when the assertion *fails*.

So it needs a hint about what could have gone wrong ('max filesize setting failed' or something)

dhensby:
Ok, and final comment, your tests don't seem to cover setting non "ini" syntax (ie: raw size values).

filesystem/Upload.php
```diff
@@ -384,7 +405,22 @@ public function setAllowedMaxFileSize($rules) {
 		if(is_array($rules) && count($rules)) {
 			// make sure all extensions are lowercase
 			$rules = array_change_key_case($rules, CASE_LOWER);
-			$this->allowedMaxFileSize = $rules;
+			$finalRules = array();
+			$tmpSize = 0;
+			
+			foreach ($rules as $rule => $value) {
+				if (is_string($value)) {
```
dhensby: I'm worried this type check is too liberal (eg: '1024' could evaluate as true).

Can we do `if (is_numeric(...))` and swap the code round so that numeric strings are allowed?

tests/filesystem/UploadTest.php
Turnerj: Ahhhh, I thought `assertTrue` was for successful assertions (thus a successful message). Should I change to `assertFalse` and/or just change to a negative message?

filesystem/Upload.php
Turnerj: Yep, definitely a valid case. Will shoot through a commit shortly.

Turnerj:
Actually just above my test is a non-INI test case for uploading files. I can add another if you think it is required.

tests/filesystem/UploadTest.php
dhensby: so `assertTrue` is saying "the first parameter MUST be `true` (ie: `$result === true`). it will fail if that is not the case and then the last param will be displayed in the test results.

dhensby:
> Actually just above my test is a non-INI test case for uploading files. I can add another if you think it is required.

I'd like to see it tested with the type definition. Also the above test doesn't seem to test successful uploads :S

chillu:
Hm, I think our definitions of a "sensible enhancement" differ here - for me, this addition makes core more complex without a direct need (feature can be written as an extension). But since two other core devs have spoken out in favour of it, I stand overruled. One more thing though: Mark `$default_max_file_size` with an `@config` annotation :)

dhensby:
@chillu I'm open to having my mind changed as I agree we want to keep core scope creep down to a minimum.

However, my view is:

1. The core allows grouping of extensions into categories
2. The core allows setting upload limits by file extension
3. I think setting upload limits to extension is an edge-case requirement
4. However, if we have the feature, it makes sense to make it less painstaking to use

So it seems sensible to me to add the ability to define upload sizes by the groups that are already defined in core. I see no advantage to having to define the same rule for `jpeg` as `jpg` when the core can take care of that for me by using`[image]`.

chillu:
Ah, damn - `getAllowedMaxFileSize($ext = null)`, I missed that its already a feature in core! In that case, yes it makes sense. Sorry for the confusion here, happens when you try to do too many things at once

dhensby:
Ah yes, that was my initial confusion also!

Turnerj:
I'll get the `@config` annotation in tonight. :) In the future, should I word a little more into the PR explaining why it could/should belong in the core and more what it relates to?

Turnerj:
As discussed on #3991, reverted the HTTP status changes for the Upload Field and updated the tests. @dhensby, do you want to do whatever magic to check against the 3 branch? Should now pass the tests.

dhensby:
@Turnerj I'm not seeing any `isError` checks in your new tests?

Turnerj:
@dhensby , so what I had modified before [was here](https://github.com/silverstripe/silverstripe-framework/pull/3980/files#diff-3763c77ac4434257977ba080fc021a43L170) though I have a feeling that Git sees it as a revert rather than a new change. (And the only reason these were updated before was because of me removing the line that set the HTTP status code - that line is also reverted and Github seems to have removed that file from my commit changes)

The other tests I added [here](https://github.com/silverstripe/silverstripe-framework/pull/3980/files#diff-8c10a99435d40ae645ae097ca97c4665R93) are against `Upload` (not `UploadField`) and return a boolean instead of a HTTP response (so no `isError`).

dhensby:
Ah I see, you just amended an existing test which was failing, none of your tests actually relied upon it. OK,

Turnerj:
Sorry for the confusion about that! Let me know if there are any last minute amendments you need from me. :)

filesystem/Upload.php
```diff
@@ -335,6 +344,11 @@ class Upload_Validator {
 	 */
 	public $allowedExtensions = array();
 
+	public function __construct() {
+		$fileSize = Config::inst()->get('Upload_Validator', 'default_max_file_size');
+		$this->setAllowedMaxFileSize($fileSize);
```
dhensby: This function does a fair bit of work, which may be unnecessary if the defaults are overridden after construct then this is a waste of CPU cycles.

Could there be a way to leave the computation to be "just in time" (ie: only when it's manually set OR when the `getAllowedMaxFileSize` is executed.

dhensby: Perhaps, rather than setting the defaults against the object, the `get` function will check the instance specific settings and then check the defaults after?

dhensby:
Right, it looks like the `setAllowedMaxFileSize` and `getAllowedMaxFileSize` functions don't have any test coverage themselves.

They probably need some fairly in depth tests to make sure that defaults / explicit per instance rules are returned as expected

Turnerj:
> Right, it looks like the setAllowedMaxFileSize and getAllowedMaxFileSize functions don't have any test coverage themselves.

[Here](https://github.com/silverstripe/silverstripe-framework/pull/3980/files#diff-8c10a99435d40ae645ae097ca97c4665R91) in the `UploadTest`, it covers the setting of `setAllowedMaxFileSize` and when it performs `load()`, it internally performs the `getAllowedMaxFileSize` check. I can add a separate test however I think these functions are covered quite well.



> This function does a fair bit of work, which may be unnecessary if the defaults are overridden after construct then this is a waste of CPU cycles.

> Could there be a way to leave the computation to be "just in time" (ie: only when it's manually set OR when the getAllowedMaxFileSize is executed.

> Perhaps, rather than setting the defaults against the object, the get function will check the instance specific settings and then check the defaults after?

Yep, I see what you mean about that. I did it this way initially to keep the overall code changes low though I am aware that the configuration system does have overhead.

I will see about having `getAllowedMaxFileSize` to read the default if it can't find a match on the instance, just need to work out the tidiest way for implementation (I can think of half a dozen different ways to do it).

Turnerj:
Latest commit addresses the issue with the config look up in the constructor. That has moved to the `getAllowedMaxFileSize` function where it checks if there are any instance values currently set and if not, overrides with the default values. If no default values are present in the config, it uses the values set by PHP for file uploads.

I've also added a bunch of new tests [starting here](https://github.com/silverstripe/silverstripe-framework/pull/3980/files#diff-8c10a99435d40ae645ae097ca97c4665R135) which cover `setAllowedMaxFileSize`, `getAllowedMaxFileSize` and also how the configuration defaults are overridden.

tests/filesystem/UploadTest.php
```diff
@@ -84,15 +84,92 @@ public function testAllowedFilesize() {
...
+		// Check the max file size overrides the config
+		$maxFileSizes = array(
+			'[image]' => '1k',
+			'txt' => 1000
+		);
+		Config::inst()->update('Upload_Validator', 'default_max_file_size', $maxFileSizes);
+		$v = new UploadTest_Validator();
+		
+		$retrievedSize = $v->getAllowedMaxFileSize('[image]');
+		$result = $retrievedSize == File::ini2bytes($maxFileSizes['[image]']);
```
dhensby: By using `File::ini2bytes` you're relying on the exact same logic as the function you're testing. this should be a hardcoded 1000

tests/filesystem/UploadTest.php
```diff
@@ -84,15 +84,92 @@ public function testAllowedFilesize() {
...
+		$this->assertTrue($result, 'Max file size check on default values failed');
```
dhensby: You should just use:

```php
$this->assertEquals(1000, $v->getAllowedMaxFileSize('[image]');
```

dhensby:
Awesome, thanks. I think we're really close now.

My comments are:

1. Create test functions per function you test eg: `testGetAllowedMaxFileSize()` for tests around `getAllowedMaxFileSize`
2. Try not to just duplicate functionality from the functions you test (like using `File::ini2bytes`) tests are about known input == known output so hard code it.
3. Use more than just `assertTrue` / `assertFalse` for your tests. `$this->assertEquals('var', $var)` is much better than `$this->assertTrue($var == 'var')`

Turnerj:
Tests have been updated as requested :)

tests/filesystem/UploadTest.php
```diff
@@ -84,15 +84,91 @@ public function testAllowedFilesize() {
...
+		$v = new UploadTest_Validator();
+		$v->setAllowedMaxFileSize($maxFileSizes);
+		
+		$retrievedSize = $v->getAllowedMaxFileSize('[doc]');
+		$this->assertEquals(2000, $retrievedSize, 'Max file size check on instance values failed');
+		
+		// Check that the instance values overwrote the default values
+		// ie. The max file size will not exist for [image]
+		$retrievedSize = $v->getAllowedMaxFileSize('[image]');
+		$this->assertFalse($retrievedSize, 'Max file size check on instance values failed');
```
dhensby: Hmm, I'm not sure this is desired behaviour..

If one sets a global default for `[image]`, that should be retained unless it's overwritten later

tests/filesystem/UploadTest.php
```diff
@@ -84,15 +84,91 @@ public function testAllowedFilesize() {
...
+		
+		$retrievedSize = $v->getAllowedMaxFileSize('txt');
+		$this->assertEquals(4096, $retrievedSize, 'Max file size check on instance values failed');
+		
+		// Cleanup configuration not to affect other tests
+		Config::inst()->remove('Upload_Validator', 'default_max_file_size');
```
dhensby: You can wrap your tests in `Config::nest()` and `Config::unnest()` to "sandbox" your config changes and prevent you having to keep track of changes yourself

tests/filesystem/UploadTest.php
dhensby: Also, shouldn't it respond with PHP's internal max?

tests/filesystem/UploadTest.php
```diff
@@ -84,15 +84,91 @@ public function testAllowedFilesize() {
...
+			'txt' => 1000
+		);
+		Config::inst()->update('Upload_Validator', 'default_max_file_size', $configMaxFileSizes);
+		$v = new UploadTest_Validator();
+		
+		$retrievedSize = $v->getAllowedMaxFileSize('[image]');
+		$this->assertEquals(1024, $retrievedSize, 'Max file size check on default values failed');
+		
+		$retrievedSize = $v->getAllowedMaxFileSize('txt');
+		$this->assertEquals(1000, $retrievedSize, 'Max file size check on default values failed');
```
dhensby: Can you also test getting a value you haven't set and that it responds as expected?

tests/filesystem/UploadTest.php
Turnerj: So the way `getAllowedMaxFileSize` and `setAllowedMaxFileSize` worked before was:
- If you set a new allowed max file size, it overrides anything that use to be set
- If nothing is set, return false

My code makes the allowed max file size have an initial value which is either what is defined in the configuration or what PHP says in the maximum upload/post size (if nothing is defined in the config). Because I am currently using the overall way it worked before, setting a new value override the old (rather than merge) which means `[image]` check actually returns false.

If I introduce the ability to merge the values, I should probably introduce a way to remove items and also clear the entire list. I can do this however my change was designed to be an addition without breaking change.

I do though understand what you mean, it would be nice if the values did merge rather than override. If you are happy with the scope of this PR to increase, I can introduce those changes too.

tests/filesystem/UploadTest.php
Turnerj: I thought that something like that might exist though I didn't originally find it when searching through the testing documentation. Will update code to support that now.

Turnerj:
Any progress on this or the information about merging-vs-overriding the max file size like [what I mentioned here](https://github.com/silverstripe/silverstripe-framework/pull/3980#discussion_r26430316)?

Turnerj:
Can this be merged in or should I just close it?

tractorcow:
Hi @Turnerj this PR is fine but it needs to go into the 3 branch, not the 3.1 one. There's a bit of an explanation about semver at http://doc.silverstripe.org/en/contributing/release_process/.

Turnerj:
@tractorcow, have opened another PR against 3 (#4052) for any further discussion and for you to merge in :)

tests/filesystem/UploadTest.php
dhensby: If it is in line with how the system currently works, then ok

dhensby: Though, if you could open an issue about this, as I don't think it's desirable behaviour.

tractorcow:
Merged there :)

tests/filesystem/UploadTest.php
Turnerj: Done, see #4055.

PR #3982, opened 2015-03-07
Safe unnesting of Config and Injector

dhensby:
Currently this only seriously affects `Injector`.

Injector declares that the parameter to `set_inst` must be an `Injector` and so if you unnest too far, it will break because it attempts to set an injector inst of `null`.

Thoughts:

Perhaps if we unnest too far it should just instantiate a new inst explicitly, rather than returning the current inst... Though I don't feel an arbitary unnest should discard the route config, instead, a dev should explicitly do that if it's their desire to reset the config completely.

### DIFF
--- a/core/Config.php
+++ b/core/Config.php
@@ -10,62 +10,62 @@
  * - An array
  * - A non-array value
  *
- * If the value is an array, each value in the array may also be one of those 
+ * If the value is an array, each value in the array may also be one of those
  * three types.
  *
- * A property can have a value specified in multiple locations, each of which 
- * have a hard coded or explicit priority. We combine all these values together 
- * into a "composite" value using rules that depend on the priority order of 
+ * A property can have a value specified in multiple locations, each of which
+ * have a hard coded or explicit priority. We combine all these values together
+ * into a "composite" value using rules that depend on the priority order of
  * the locations to give the final value, using these rules:
  *
- * - If the value is an array, each array is added to the _beginning_ of the 
- *	composite array in ascending priority order. If a higher priority item has 
+ * - If the value is an array, each array is added to the _beginning_ of the
+ *	composite array in ascending priority order. If a higher priority item has
  *	a non-integer key which is the same as a lower priority item, the value of
- * 	those items  is merged using these same rules, and the result of the merge 
- *	is located in the same location the higher priority item would be if there 
+ * 	those items  is merged using these same rules, and the result of the merge
+ *	is located in the same location the higher priority item would be if there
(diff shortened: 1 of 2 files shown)


micmania1:
So hard to read the modified files btw - too much whitespace changes.

Should we not just throw an error when Config::set_inst() is passed anything other than an instance of Config? I don't see benefit in trying to fix a users issue, at least not without alerting them that something is wrong.

tractorcow:
I think it should crash with a message instead of trying to silently
recover.

dhensby:
@micmania1 sorry, you can see non-whitespace changes by appending `?w=1` to the files url (https://github.com/silverstripe/silverstripe-framework/pull/3982/files?w=1) but I'll stop with my whitespace cleaning in PRs.

I'm still quite torn between whether the framework should be lenient/forgiving or strict.

This way dev's can be a bit lazy and unnest more than they nest and they'll never lose the initial settings for the app (which I think is pretty important to retain). It's not infeasible to imagine a scenario where a complex set of nesting and unnesting is done where keeping track of nesting counts is difficult.

Throwing a logic exception "Can't unnest Config that is not nested" makes sense, but should it really break the system if it's recoverable? The debate would be how is best to recover if we think killing the application is not a good idea here.

Personally, I'm a fan of being strict in my own projects, but that's not really how the framework is in general and I don't think its conducive to encouraging our community to use SS. By being strict, we force our devs to write more code (either try catch or checking if the Config/Injector is nested).

If we threw errors, we'd need to provide an API to check if the config was currently nested and probably a way to access the root `Config`/`Injector`.

Perhaps throwing a warning and then retaining the root `Config`/`Injector` is  the best of both? We can push better code whilst allowing for sloppiness?

dhensby:
@chillu / @hafriedlander / @sminnee - your thoughts on strictness would be appreciated here

tractorcow:
Perhaps a crash on dev/test and silently ignored on live?

dhensby:
Yes, that's how I'm leaning

core/Config.php
```diff
@@ -231,13 +231,16 @@ public static function nest() {
...
 	 * Config object was copied from.
-	 * 
+	 *
 	 * @return Config Reference to new active Config instance
 	 */
 	public static function unnest() {
-		return self::set_instance(self::$instance->nestedFrom);
+		if (self::inst()->nestedFrom) {
+			self::set_instance(self::inst()->nestedFrom);
+		}
```
tractorcow: Please use a warning here in an else so that we can catch these errors during dev.

PR #3983, opened 2015-03-07
Making docs gender agnostic

dhensby:
Pretty straight forward. We shouldn't be using genders in docs when referring to users or developers. Especially when they are so one-sidedly referring to men.

Especially as it's even stated in the [docs on Docs](http://docs.silverstripe.org/en/contributing/documentation/#writing-style)

### DIFF
--- a/admin/javascript/LeftAndMain.js
+++ b/admin/javascript/LeftAndMain.js
@@ -10,7 +10,7 @@ jQuery.noConflict();
 		// Entwine's 'fromWindow::onresize' does not trigger on IE8. Use synthetic event.
 		var cb = function() {$('.cms-container').trigger('windowresize');};
 
-		// Workaround to avoid IE8 infinite loops when elements are resized as a result of this event 
+		// Workaround to avoid IE8 infinite loops when elements are resized as a result of this event
 		if($.browser.msie && parseInt($.browser.version, 10) < 9) {
 			var newWindowWidth = $(window).width(), newWindowHeight = $(window).height();
 			if(newWindowWidth != windowWidth || newWindowHeight != windowHeight) {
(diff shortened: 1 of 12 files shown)


dhensby:
Some of the changelogs had references to "he" which should probably come out, but wasn't sure if I should be editing the changelogs that are driven from commit messages

oddnoc:
This is great, but could you remove the whitespace-only changes (they make it hard to review the substantive changes). Thanks!

chillu:
Hey Dan, thanks a lot - was on my bucket list for ages heh. How much work is it to remove the whitespace? I'd be inclined to merge, but as Fred says it makes it a lot harder to review. Please check that either your IDE doesn't remove trailing whitespace by default, or that you do a line-by-line review before committing (I find SourceTree handy for this)

dhensby:
https://github.com/silverstripe/silverstripe-framework/pull/3983/files?w=1

dhensby:
Until my last few PRs I used to do adding with hunks to keep whitespace changes out. However, cleaning trailing whitespace and adding newlines to EOF should be standard practice so I've decided that this falls into "boy scout rule" (leave something better than when you found it).

I *can* go through this and take out all the whitespace changes if it's really a big deal, but I don't think it should be :)

chillu:
I know about the `w=1` param, but not all community members participating in the discussion on a specific diff will. So while the "boy scout rule" is a good one, I think having a concise and readable code change is more important. If you think trailing whitespace is an issue worth fixing, please submit a separate PR for it, across all files. Its going to be a nuisance in further "git blame" use, but less so than having all this diff noise on further PRs.

Regardless, I'll merge this one - but won't next time ;)

dhensby:
Thanks Ingo, from now on I'll do cleaner commits. 

Annoyingly my IDE is meant to only clean line endings on changed lines, but seems to fail to do that

oddnoc:
I noticed that the "reformat changed lines only" feature in NetBeans doesn't seem to work, so I had to turn it off. Thanks for the w=1 tip. I wasn't aware of it :)

PR #3985, opened 2015-03-08
Refactor ViewableData::obj caching

tractorcow:
Replaces #3955

This feature allows caching of viewable data to be overridden in subclasses.


g4b0:
Nice work tractorcow!

PR #3986, opened 2015-03-08
BUG Fix gridfield generating invalid session keys

tractorcow:
The issue is that md5 could occasionally generate all-numeric keys which would fail on assignment. `$_SESSION` may not have numeric keys.


tractorcow:
This will necessitate an early 3.1.11 release, since this is a high-level issue in the framework which will break some gridfields.

dhensby:
Clearly a safe change, but I do like waiting for the green ticks :)

tractorcow:
ta

PR #3987, opened 2015-03-09
DOCS Update i18n language customisation docs

joshkosmala:
I could be wrong, but pretty sure it's meant to be common_languages instead of common_locales

### DIFF
--- a/docs/en/02_Developer_Guides/13_i18n/index.md
+++ b/docs/en/02_Developer_Guides/13_i18n/index.md
@@ -95,7 +95,7 @@ Similarly, to change an existing language label, you can overwrite one of these 
 
 	:::yml
 	i18n:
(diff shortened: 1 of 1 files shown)


dhensby:
No, I think locales is right

common_locales is country codes like `en_GB` -> United Kingdom

common_languages is language codes like `en` -> English

dhensby:
Hmm, though the one above looks wrong...

This is confusing docs as they both refer to language when one is locale...

dhensby:
If you could amend this PR to make that aspect of the docs more clear and correct the one above, that would be super helpful!

sminnee:
@dhensby in the absence of that improvement, shall we just merge this?

dhensby:
We can make the change when we merge by hand :P

PR #3989, opened 2015-03-09
Update 00_Partial_Caching.md

wernerkrauss:
Adding $ for variable names and quoted parameters as discussed with @kinglozzer on IRC


kinglozzer:
My thinking was:

- Our convention is `$` prefixes for every other control structure, so why not `<% cached %>`?
- Quotes are optional, yes - I guess that’s a personal preference - but I’d prefer to encourage this too :)

dhensby:
> Quotes are optional, yes

Quotes 100% should be used for string literals. $ prefix for variables.

It's my dream that one day the template parser will be strict about types so we can use `false` properly

dhensby:
Thanks

PR #3991, opened 2015-03-09
3980 on 3

dhensby:
Fixes #3980

Conflicts:
	docs/en/02_Developer_Guides/03_Forms/Field_types/05_UploadField.md
	filesystem/Upload.php
	tests/filesystem/UploadTest.php

### DIFF
--- a/docs/en/02_Developer_Guides/03_Forms/Field_types/05_UploadField.md
+++ b/docs/en/02_Developer_Guides/03_Forms/Field_types/05_UploadField.md
@@ -148,6 +148,16 @@ NOTE: this only sets the configuration for your UploadField, this does NOT chang
 	$this->getValidator()->setAllowedMaxFileSize($size);
 ```
 
+You can also specify a default global maximum file size setting in your config for different file types. This is overridden when specifying the max allowed file size on the UploadField instance.
+
+```yaml
+	Upload_Validator: 
+	  default_max_file_size: 
+	    '[image]': '1m'
+	    '[doc]': '5m'
+	    'jpeg': 2000
+```
+
 ### Preview dimensions
 Set the dimensions of the image preview. By default the max width is set to 80 and the max height is set to 60.
 
@@ -296,6 +306,19 @@ Certain default values for the above can be configured using the YAML config sys
 ```
 
 The above settings can also be set on a per-instance basis by using `setConfig` with the appropriate key.
+
+The `Upload_Validator` class has configuration options for setting the `default_max_file_size`.
+
+```yaml
+	Upload_Validator: 
+	  default_max_file_size: 
+	    '[image]': '1m'
+	    '[doc]': '5m'
+	    'jpeg': 2000
+```
+
+You can specify the file extension or the app category (as specified in the `File` class) in square brackets. It supports setting the file size in bytes or using the syntax supported by `File::ini2bytes()`.
+
 
 You can also configure the underlying `[api:Upload]` class, by using the YAML config system.
 
--- a/filesystem/Upload.php
+++ b/filesystem/Upload.php
@@ -306,6 +306,16 @@ class Upload extends Controller {
 class Upload_Validator {
 
 	/**
+	 * Contains a list of the max file sizes shared by
+	 * all upload fields. This is then duplicated into the
+	 * "allowedMaxFileSize" instance property on construct.
+	 *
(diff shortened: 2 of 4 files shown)


Turnerj:
Looks like there was another test added on the 3 branch which uses `$response->isError()` rather than checking the payload directly.

https://github.com/dhensby/silverstripe-framework/blob/added-max-size-by-upload-type-turnerj/tests/forms/uploadfield/UploadFieldTest.php#L199-L208

dhensby:
Hmm, it seems there are a few tests here that are relying on `isError`.. I'm thinking changing the API to use JSON instead of HTTP errors is a bad move :/

Turnerj:
It should only be that one reference I highlighted that needs adjusting to pass the tests as only the `upload` function on `UploadField` has had the throwing of an error removed.

From my previous research, to allow the client side to read the error message when a non-successful HTTP status is used requires a replacement/patch of the Upload JS code and tweaking how errors are handled in the CMS (clientside).

Seems to be a decent use of `isError` and the JSON response checks:

[Example 1](https://github.com/dhensby/silverstripe-framework/blob/added-max-size-by-upload-type-turnerj/tests/forms/uploadfield/UploadFieldTest.php#L361-L363)
[Example 2](https://github.com/dhensby/silverstripe-framework/blob/added-max-size-by-upload-type-turnerj/tests/forms/uploadfield/UploadFieldTest.php#L264-L273)

It does seem that mainly upload tests are using the JSON check and everything are using the `isError` check (eg. delete tests).

Because the user won't know what the error is if we go back to using HTTP errors, I think we should persist with removing it.

dhensby:
Is it not strange that sometimes we use an HTTP error and sometimes a JSON one?

It seems quite inconsistent and will only aid confusion for devs about when to respond with an error and when not to.

I think a much better patch *would* be to get the CMS handling errors properly. the jqXHR object should allow access to the response text?

Turnerj:
Yeah, it is strange in one way that some parts use `isError` and some parts don't. I definitely understand where you are coming from there.

I see a reason to use HTTP status codes for web services and APIs for backend services though less so for an UploadField which its functions are usually invoked by the CMS or other frontend services.

A quick lookup of jqXHR makes it sound like it is possible to read the response text under HTTP errors, will need to test to confirm. If it does, will need to work out how to patch a third party library we are using in the CMS (whether I update the library itself and get a PR to use a completely new version OR whether I just patch our copy, making it slightly more difficult to update in the future).

dhensby:
gosh.. @silverstripe/core-team thoughts?

dhensby:
actually, @Turnerj you know what, keep using the http errors here. we're getting bogged down in a separate issue which is that the http errors aren't handled in the CMS.

The current practice is HTTP errors, lets stick with it and get this feature in.

Turnerj:
OK, will update my PR later today, reversing the removal of the HTTP errors being thrown and updating my tests to use the `isError` check.

Is there are Issue # (or dev mailing list link) for the HTTP errors work? I am willing to lend a hand in bringing those sorts of updates to the CMS if you guys want/need.

tractorcow:
If you like I can merge up 3.1 into 3 so that the documentation page gets restored.

dhensby:
@tractorcow please :)

dhensby:
> Is there are Issue # (or dev mailing list link) for the HTTP errors work? I am willing to lend a hand in bringing those sorts of updates to the CMS if you guys want/need.

I'm not certain, I know that there are comments on the commit that introduced this problem, but not sure if there is an issue (probably is... check CMS?!)

tractorcow:
My understanding is:

HTTP errors for things which a user shouldn't be able to do. E.g. injecting invalid data into requests, cross site scripting attempts, calling forbidden form actions. If a user in the normal course of actions encounters such errors then that indicates a problem in the code.

json errors for errors which should be displayed to a user in the normal course of action. E.g. "You uploaded a file which is too large", "Can't overwrite a file with the same name" etc.

dhensby:
@tractorcow I think we all agree with that, but that's a hefty API change. The field currently returns HTTP errors so it should remain that way as it's a seperate bug to what this is trying to add / change.

tractorcow:
3 has had changes from 3.1 merged in. You'll need to rebase this.

dhensby:
thanks!

PR #3996, opened 2015-03-10
Fixed typo in HtmlType translation key

wernerkrauss:
Small typo, i wonderd why "HTML-File" appears twice on transifex.


dhensby:
thanks

PR #3997, opened 2015-03-10
Performance improvements to ClassInfo::ancestry()

kinglozzer:
Basically just switching from `get_parent_class()` in a loop to using `class_parents()` instead. Not a massive performance increase, but every little helps :)

Benchmarks on PHP 5.5.18, completely blank install of CMS + Framework only, loading home page. Aggregate of 10 runs each:

Before:
screen shot 2015-03-10 at 13 45 37

After:
screen shot 2015-03-10 at 13 45 44

### DIFF
--- a/core/ClassInfo.php
+++ b/core/ClassInfo.php
@@ -157,15 +157,20 @@ class ClassInfo {
 		if (!is_string($class)) $class = get_class($class);
 
 		$cacheKey = $class . '_' . (string)$tablesOnly;
-		$parent = $class;
 		if(!isset(self::$_cache_ancestry[$cacheKey])) {
-			$ancestry = array();
-			do {
-				if (!$tablesOnly || DataObject::has_own_table($parent)) {
-					$ancestry[$parent] = $parent;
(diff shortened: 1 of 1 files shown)


dhensby:
nice - thoughts on why it's not 3.1?

kinglozzer:
Wasn’t sure if improvements to public APIs are allowed in patch releases, reading through the semver spec it only mentions that you MAY increment minor versions for improvements in private APIs. So I guess that means it could go against 3.1 :)

dhensby:
You're not changing any APIs here, you're making a performance patch. The API to call the function and the response of the function are identical before and after this change, so it's acceptable as a patch release.

Slowness is definitely a bug, not a feature ;)

kinglozzer:
I’ll re-open against 3.1, I spotted a possible backwards-incompatibility anyway (if you pass a class that doesn’t exist). I _suppose_ we should keep guarding against that, even though you’ve probably got bigger issues if you’re passing around non-existant classes!

tractorcow:
Nice job @kinglozzer

PR #3998, opened 2015-03-10
Performance improvements to DataObject::is_composite_field()

kinglozzer:
`DataObject::getField()` calls this, so it’s called many times!

Before:
screen shot 2015-03-10 at 17 06 13

After:
screen shot 2015-03-10 at 17 06 17

These benchmarks were conducted on the home page of a site, showing ~8 DataObjects, nothing fancy beyond that. I’m happy to conduct the tests on a blank install w/ Simple theme if anyone would like to see those results :)

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -161,6 +161,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	protected static $_cache_get_one;
 	protected static $_cache_get_class_ancestry;
 	protected static $_cache_composite_fields = array();
+	protected static $_cache_is_composite_field = array();
 	protected static $_cache_custom_database_fields = array();
 	protected static $_cache_field_labels = array();
 
@@ -308,14 +309,25 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	 * Will check all applicable ancestor classes and aggregate results.
 	 */
 	public static function is_composite_field($class, $name, $aggregated = true) {
-		if(!isset(DataObject::$_cache_composite_fields[$class])) self::cache_composite_fields($class);
+		$key = $class . '_' . $name . '_' . (string)$aggregated;
+
+		if(!isset(DataObject::$_cache_is_composite_field[$key])) {
+			$isComposite = null;
+
+			if(!isset(DataObject::$_cache_composite_fields[$class])) {
+				self::cache_composite_fields($class);
+			}
 		
-		if(isset(DataObject::$_cache_composite_fields[$class][$name])) {
-			return DataObject::$_cache_composite_fields[$class][$name];
-			
-		} else if($aggregated && $class != 'DataObject' && ($parentClass=get_parent_class($class)) != 'DataObject') {
-			return self::is_composite_field($parentClass, $name);
+			if(isset(DataObject::$_cache_composite_fields[$class][$name])) {
+				$isComposite = DataObject::$_cache_composite_fields[$class][$name];
+			} elseif($aggregated && $class != 'DataObject' && ($parentClass=get_parent_class($class)) != 'DataObject') {
+				$isComposite = self::is_composite_field($parentClass, $name);
+			}
+
+			DataObject::$_cache_is_composite_field[$key] = ($isComposite) ? $isComposite : false;
 		}
+
+		return DataObject::$_cache_is_composite_field[$key] ?: null;
 	}
 
 	/**
@@ -2960,6 +2972,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 		DataObject::$_cache_db = array();
 		DataObject::$_cache_get_one = array();
 		DataObject::$_cache_composite_fields = array();
+		DataObject::$_cache_is_composite_field = array();
 		DataObject::$_cache_custom_database_fields = array();
 		DataObject::$_cache_get_class_ancestry = array();
 		DataObject::$_cache_field_labels = array();


kinglozzer:
More info: there is such a huge reduction in function calls because the method calls itself on each parent class, every single time it’s executed! In the screenshots below, each nested `DataObject::is_composite_field()` call then has further nested calls beneath it.

Before:
screen shot 2015-03-10 at 17 11 55

After:
screen shot 2015-03-10 at 17 12 06

dhensby:
We're sacrificing memory for speed here (again).

We need to be careful that we don't lose sight of that and start getting complaints about SS being memory heavy (which I've always thought it was).

These caches seem to never get cleaned up.

Question: Why is this a better solution than using `SS_Cache`? I'd have thought this stuff doesn't change between requests?

kinglozzer:
I’ll investigate the memory usage in more detail. It shouldn’t be a significant increase: it’s one array with, at worst, a few hundred keys and string/false values. Plus I imagine removing all those extra function calls mitigates that increase a bit too.

> These caches seem to never get cleaned up.

Do you mean they never get emptied, or never get cleaned up from a code maintenance point of view? These caches are reset in `DataObject::reset()` for tests, but apart from that their benefit comes from them surviving the lifetime of the request so I don’t think they can be cleaned.

> Question: Why is this a better solution than using SS_Cache? I'd have thought this stuff doesn't change between requests?

I thought the same thing while authoring this. There are many things we could persist between requests in `DataObject`, and that’s partly the reason why I didn’t attempt that: it’d be a much bigger changeset than this PR as it’d make more sense to do it all at once.

dhensby:
> Do you mean they never get emptied

Yes. Though perhaps it isn't a problem?

MatthewBonner:
I'm with Dan, this seems a fairly pointless exercise for the reasons Dan has already provided, however we do need to investigate how performance can be improved, the problem though is that we don't want to compromise a clean API in favour of using SS_Cache everywhere either.

The pull request replaces 1 line of unreadable code with 8 lines of unreadable code, this is ultimately going to lead to code that can't be optimised at all because nobody can read it.

If we want better performance for bigger websites, we should be looking at ways in which we can scale out, such as separating out the tables into different databases so we can have a database server for authentication, a database server for sessions, a database server for versioning, etc. This way we can distribute the load across more databases.

kinglozzer:
> this seems a fairly pointless exercise

I disagree: a function that unnecessarily calls itself thousands of times immediately sticks out as a nail that needs to be hammered in my view. Especially so given how simple this change is, and that it offers a noticeable performance benefit. Dan’s raised valid concerns about memory, and that does need to be investigated before this could be merged, but I don’t think that renders this PR immediately pointless.

IMO this is a “performance bug”: if it can be fixed in a point release, without introducing unnecessary complexity (or significantly affecting memory usage), then it should be.

> The pull request replaces 1 line of unreadable code with 8 lines of unreadable code

I’ve not changed any functionality here, I’ve basically just wrapped the existing code in an `if(isset())` conditional statement. We may be able to refactor the code, and look into storing the data between requests, but I believe that’s best done in a separate PR. I don’t think this PR blocks that, or adds anything that makes the code any harder to understand than it already is.

> we don't want to compromise a clean API in favour of using SS_Cache everywhere either

You’re absolutely right, this sort of addition would need careful thought and would probably introduce new APIs, so likely wouldn’t be a candidate for a patch release. As above, though, I don’t think this PR would block/slow down that sort of improvement being made.

dhensby:
> this seems a fairly pointless exercise for the reasons Dan has already provided

For clarity, I'm not saying this is pointless - far from it. I pretty much think a memory for speed trade-off is acceptable. I'm just saying we need to make sure we don't just end up shifting the problem from speed to memory.

> we don't want to compromise a clean API in favour of using SS_Cache everywhere either

I'm not sure using SS_Cache makes the API any more dirty... using it *everywhere* makes the core more consistent and predictable, rather than these protected / private vars knocking around to do volatile in-memory caching.

At the moment we're basically doing `$cache = ClassName::$private_static[$cache_key];` rather than `$cache = SS_Cache::factory('ClassName_Cache)->get($cache_key');` which seems a bit more clean (and testable) to me.

> The pull request replaces 1 line of unreadable code with 8 lines of unreadable code, this is ultimately going to lead to code that can't be optimised at all because nobody can read it.

I don't think it's unreadable, but better in line docs would be a help

> If we want better performance for bigger websites, we should be looking at ways in which we can scale out, such as separating out the tables into different databases so we can have a database server for authentication, a database server for sessions, a database server for versioning, etc. This way we can distribute the load across more databases.

I don't think that's a problem we're trying to mitigate at the moment. It's about 5 orders of magnitude above the kind of performance fixes we need. It's not about making SS scalable for the largest of sites; it's about reducing the TTFB on a "typical" site.

I don't see much merit in an argument of "it's 500ms for your tiny site, but if you used our DB sharding feature you'd be able to server 10m users with the same response time too" when most SS sites are no where near that scale.

dhensby:
@kinglozzer I think the memory usage for this change is pretty trivial and not worth a lot of effort to investigate. as you say, it's a key -> bool store.

If you don't think using `SS_Cache` is in the remit of this PR, then OK - we can consider it as-is.

I agree it can go into a patch release.

willmorgan:
I agree this is a performance bug and should be good to go in to a patch release.

SilverStripe _is_ memory heavy, but we shouldn't be using that as a reason to stop fairly awesome performance improvements like this.

Yes, we should be looking at using `SS_Cache`, but not right now - this is a great candidate for a patch that could help higher traffic sites.

chillu:
Good discussion - agree with Loz and Will that "perfect is the enemy of good" in this case. Yes, we should use SS_Cache more consistently across core, and we should allow more horizontal scaling, but that shouldn't stop us from slightly improving the status quo. DataObject already uses quite a few statics for caching, so its actually more consistent to continue this.

chillu:
BTW, I'm merging this as proposed into 3.1 since it'll improve performance across many sites, and is a low risk non API breaking change. But in general, 3.1 is "bugfix only" now, so we should default these things to 3.2

PR #4006, opened 2015-03-11
FIX: Security::$default_message_set Config value unusable

kinglozzer:
`Security::set_default_message_set()` is marked as deprecated and encourages the use of the `Security.default_message_set` config setting, but that setting isn’t actually used when fetching messages.

I’ve also removed the default empty string value. If the default is a string, setting an array value in config results in a type mismatch exception, whereas if the default isn’t defined you can set the value to either a string or an array.

### DIFF
--- a/security/Security.php
+++ b/security/Security.php
@@ -93,9 +93,10 @@ class Security extends Controller {
 	/**
 	 * Default message set used in permission failures.
 	 *
+	 * @config
 	 * @var array|string
 	 */
-	private static $default_message_set = '';
+	private static $default_message_set;
 
 	/**
 	 * Random secure token, can be used as a crypto key internally.
@@ -176,9 +177,6 @@ class Security extends Controller {
 	 *                                 If you pass an array, you can use the
 	 *                                 following keys:
 	 *                                   - default: The default message
-	 *                                   - logInAgain: The message to show
-	 *                                                 if the user has just
-	 *                                                 logged out and the
 	 *                                   - alreadyLoggedIn: The message to
 	 *                                                      show if the user
 	 *                                                      is already logged
@@ -209,8 +207,8 @@ class Security extends Controller {
 		} else {
 			// Prepare the messageSet provided
 			if(!$messageSet) {
-				if(self::$default_message_set) {
-					$messageSet = self::$default_message_set;
+				if($configMessageSet = static::config()->get('default_message_set')) {
+					$messageSet = $configMessageSet;
 				} else {
 					$messageSet = array(
 						'default' => _t(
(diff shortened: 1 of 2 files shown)


security/Security.php
```diff
@@ -95,7 +95,7 @@ class Security extends Controller {
 	 *
 	 * @var array|string
```
dhensby: Fix the PHPDoc?

kinglozzer: The thing is it _can_ be set to a string, it’s [treated the same as setting `default: 'string'`](https://github.com/kinglozzer/sapphire/blob/patch-1/security/Security.php#L237-L239), so I figured I’d leave it. Though I guess it’s not really a message _set_ if you just set it to a string!

dhensby: I stand corrected

dhensby: Hmm, so why does having it as empty string cause any problems?!

I think this should just be null / undefined, tbh.

Can I ask you to mark it `@config` please?

kinglozzer: I’ve got no idea, but if it’s an empty string then setting the following throws a type mismatch exception in `Config`:

```yml
Security:
  default_message_set:
    'default': 'test'
```

Leaving it undefined seems to work fine with both a string and array though, so I’ll make that change :)

dhensby:
Can we have some tests for this feature, please?

If there was a bug you're fixing, please recreate the bug in a test and prove it's fixed. also, if there are any other bugs or potentially untested scenarios, that are now resolved, it'd be good to give that some test coverage

kinglozzer:
Added tests, also removed one of the messages as it’s never used.

dhensby:
Hmm, is that something that's meant to be in 3.2 but some how made it's way into 3.1?

dhensby:
Either way, it should come out if it's not used, but please take it out of the docblock too?

Did you check it wasn't used in CMS?

tractorcow:
I don't see how it could even possibly be used.

kinglozzer:
I’ve removed 'logInAgain' from the docblock too. The array is a local variable that’s never returned or passed to any other methods, so it can’t be used anywhere else. Looking at [blame view](https://github.com/silverstripe/silverstripe-framework/commit/c4f3bc962f7acd6c749b2d59740351ed1f5bca8f#diff-e69a0586f00e73ade24740217b63eb07R186) it looks like it’s a really old ([removed](https://github.com/silverstripe/silverstripe-framework/commit/60ca784aaef1ee6386ca8b4c29dc03b27ddd7254)) feature that tried to guess if you’d just logged out

dhensby:
Whoop, good riddance!

security/Security.php
```diff
@@ -209,8 +207,8 @@ public static function permissionFailure($controller = null, $messageSet = null)
 		} else {
 			// Prepare the messageSet provided
 			if(!$messageSet) {
-				if(self::$default_message_set) {
-					$messageSet = self::$default_message_set;
+				if($configMessageSet = Security::config()->get('default_message_set')) {
```
dhensby: `static::config()->get(...)`?

dhensby:
And I think we made it! thanks

kinglozzer:
fsa

tractorcow:
Note problem

PR #3952, opened 2015-02-27
Renamed IMagickBackend to ImagicBackend

christopherdarling:
Also updated config:: to Config:: on line #52

PR #4010, opened 2015-03-12
Remove magic quotes

micmania1:
Kill it with fire!

### DIFF
--- a/core/Constants.php
+++ b/core/Constants.php
@@ -74,16 +74,6 @@ foreach ($dirsToCheck as $dir) {
 	}
 }
 
-///////////////////////////////////////////////////////////////////////////////
-// GLOBALS AND DEFINE SETTING
-
-function stripslashes_recursively(&$array) {
-	foreach($array as $k => $v) {
-		if(is_array($v)) stripslashes_recursively($array[$k]);
-		else $array[$k] = stripslashes($v);
-	}
-}
-
 /**
  * A blank HTTP_HOST value is used to detect command-line execution.
  * We update the $_SERVER variable to contain data consistent with the rest of the application.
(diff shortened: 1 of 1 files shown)


tractorcow:
We don't need no water

dhensby:
I'm not sure this should have gone in (or been taken out) yet.

1. There's no update to the upgrade docs (we probably need that)
2. `stripslashes_recursively` if effectively an API that is being removed without warning so shouldn't come out until it's been deprecated properly
3. Whilst we still support 5.3, this should remain IMO despite our warning
4. There's no way of hooking into this part of the bootstrap process to put your own magic_quotes protection if you REALLY need to (unless you push it into your environment file but that has to be seen as a terrible solution)

---

I think we need a process for dealing with deprecated functionality because I also worry that removing deprecated functions from master will have several nasty repercussions:

1. Potentially merging up changes from 3(.1) -> master will be more difficulty because if deprecated functionality is amended to patch a bug, the merge is likely to encounter problem. Fixing deprecated functionality does happen and happened recently with [`ImagickBackend` function](https://github.com/silverstripe/silverstripe-framework/pull/3952)
2. Upgrading to master may be harder as we won't throw deprecation errors.

I think we should be removing deprecated functionality only when we branch to the new major version (ie: set up a **4.0.x-dev** branch)  - as part of that process, removing functionality marked deprecated for 3 should be removed.

Note: it should be the 4.0 branch and not the 4 branch so the upgrading from 3 to 4 is still smooth if using the 4.x-dev branch

tractorcow:
1. Can you write an upgrading doc @micmania1 ?
2. It seems that this feature has been deprecated for a while (albeit incorrectly with a warning rather than a deprecation error) so I'm considering that it's acceptable to remove it.
2. We don't NOT support 5.3, just not with that option turned on.
4. You can change .htaccess to point to your own bootstrap file which then includes the main. This is what was used by the [dynamiccache module](https://github.com/tractorcow/silverstripe-dynamiccache#installation-instructions)

tractorcow:
And merging up from 3.1 to master is always hard :)

halkyon:
It could be worth adding as another check for `get_magic_quotes_gpc `in the installer, if it's not there already.

tractorcow:
I think the installer is probably the right place to put it, even if it remains forever deprecated.

dhensby:
`stripslashes_recursively` has never had a warning and should not come out

I personally think whilst we support 5.3 we should handle magic quotes with the warning. 

From all I can recognise 4 will not support 5.3 so its safe to come out of master, just if we think deprecated code removal should be part of a branching process, then just not yet.

tractorcow:
The good thing is that it's not too late. Just add a deprecation in 3.2.

tractorcow:
See #4012

PR #4011, opened 2015-03-13
Using sigular name for model import form rather than class name

dhensby:
Currently, the import form uses the class name of an object when asking if a user wants to see the "import spec" which is not useful.

Now it'll show the familiar "singular name"


kinglozzer:
:+1:

PR #4014, opened 2015-03-15
FIX: broken link in docs to how_tos/extend_cms_interface

jedateach:
I'm not certain this fixes the issue, as I haven't tested it. It at least points it out.


dhensby:
I think the links need to be absolute? I'm not certain - @wilr?

wilr:
@jedateach Please make sure you use DOC rather than FIX for documentation work.

jedateach:
:+1:

PR #4016, opened 2015-03-16
HTTP basic auth fix

cjsewell:
It appears with some apache/php configurations, the HTTP Basic auth header created by the .htaccess rule from [this fix](https://github.com/silverstripe/silverstripe-installer/pull/78) could exist in REDIRECT_HTTP_AUTHORIZATION or HTTP_AUTHORIZATION. 

Currently only HTTP_AUTHORIZATION is checked, this pull request fixes it to check both.
Also added a little documentation.

See [When setting environment variables in Apache RewriteRule directives, what causes the variable name to be prefixed with “REDIRECT_”?](http://stackoverflow.com/questions/3050444/when-setting-environment-variables-in-apache-rewriterule-directives-what-causes#answer-10128290) for more information about Server vars being prfixed with REDIRECT_

### DIFF
--- a/security/BasicAuth.php
+++ b/security/BasicAuth.php
@@ -50,9 +50,19 @@ class BasicAuth {
 		$isRunningTests = (class_exists('SapphireTest', false) && SapphireTest::is_running_test());
 		if(!Security::database_is_ready() || (Director::is_cli() && !$isRunningTests)) return true;
 
+                /*
+                 * Enable HTTP Basic authentication workaround for PHP running in CGI mode with Apache
+                 * Depending on server configuration the auth header may be in HTTP_AUTHORIZATION or
+                 * REDIRECT_HTTP_AUTHORIZATION
+                 * 
+                 * The follow rewrite rule must be in the sites .htaccess file to enable this workaround
+                 * RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
+                 */
+		$authHeader = (isset($_SERVER['HTTP_AUTHORIZATION']) ? $_SERVER['HTTP_AUTHORIZATION'] :
+			      (isset($_SERVER['REDIRECT_HTTP_AUTHORIZATION']) ? $_SERVER['REDIRECT_HTTP_AUTHORIZATION'] : null));
 		$matches = array();
-		if (isset($_SERVER['HTTP_AUTHORIZATION']) &&
-			preg_match('/Basic\s+(.*)$/i', $_SERVER['HTTP_AUTHORIZATION'], $matches)) {
+		if ($authHeader &&
+                        preg_match('/Basic\s+(.*)$/i', $authHeader, $matches)) {
 			list($name, $password) = explode(':', base64_decode($matches[1]));
 			$_SERVER['PHP_AUTH_USER'] = strip_tags($name);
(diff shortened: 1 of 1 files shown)


security/BasicAuth.php
```diff
@@ -50,9 +50,19 @@ public static function requireLogin($realm, $permissionCode = null, $tryUsingSes
...
 
+                /*
+                 * Enable HTTP Basic authentication workaround for PHP running in CGI mode with Apache
+                 * Depending on server configuration the auth header may be in HTTP_AUTHORIZATION or
+                 * REDIRECT_HTTP_AUTHORIZATION
+                 * 
+                 * The follow rewrite rule must be in the sites .htaccess file to enable this workaround
+                 * RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
+                 */
+		$authHeader = isset($_SERVER['HTTP_AUTHORIZATION']) ? 'HTTP_AUTHORIZATION' :
```
tractorcow: Instead of getting the header name to later retrieve from $_SERVER should we just get the value here?

cjsewell:
Updated as @tractorcow suggested. . . . Far cleaner!

security/BasicAuth.php
cjsewell: Indeed! Don't ask why I implemented it the way I did?

security/BasicAuth.php
```diff
@@ -50,9 +50,19 @@ public static function requireLogin($realm, $permissionCode = null, $tryUsingSes
...
+                /*
+                 * Enable HTTP Basic authentication workaround for PHP running in CGI mode with Apache
+                 * Depending on server configuration the auth header may be in HTTP_AUTHORIZATION or
+                 * REDIRECT_HTTP_AUTHORIZATION
+                 * 
+                 * The follow rewrite rule must be in the sites .htaccess file to enable this workaround
+                 * RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
+                 */
+		$authHeader = isset($_SERVER['HTTP_AUTHORIZATION']) ? $_SERVER['HTTP_AUTHORIZATION'] :
+			      isset($_SERVER['REDIRECT_HTTP_AUTHORIZATION']) ? $_SERVER['REDIRECT_HTTP_AUTHORIZATION'] : null;
```
tractorcow: I'd probably bracket these just to be sure operator precedence doesn't screw the logic up.

http://stackoverflow.com/questions/6224330/understanding-nested-php-ternary-operator

cjsewell: Makes sense. Updated!

cjsewell: Bit odd, I have updated as sugged as you can see [here]https://github.com/guru-digital/silverstripe-framework/blob/REDIRECT_fix/security/BasicAuth.php#L61).

However the pull request diff hasnt updated....?

cjsewell: Seems to have been updated now, so the diff now shows the brackets.

tractorcow:
Mergled.

cjsewell:
Excellent. Thanks. Any idea when the next release will be published?

tractorcow:
Probably 3.1.12 sorry :)

PR #4022, opened 2015-03-18
BUG Fix joins on tables containing "select" being mistaken for sub-selects

### LINKED ISSUE #3965: Versioned tables with "select" in their name cause a bad SQL query to be generated
df6a8b6 / #3282 (released first in 3.1.6) introduced a bug that prevents versioned table name replacement from working for tables that have "select" in their name. Quotes are incorrectly removed from the table name string in [SQLQuery:906](https://github.com/silverstripe/silverstripe-framework/blob/3.1.10/model/SQLQuery.php#L906-L907):

```php
// breaks in the case $join['table'] = 'ProductSelectorPage'; for example
$table = strpos(strtoupper($join['table']), 'SELECT') ? $join['table'] : "\"" . $join['table'] . "\"";
```

It appears that the change in this commit/issue relies entirely on string matching, so the feature either needs to be rethought, or the substring match needs to be made more specific so that it cannot collide with a table name:

```php
$table = strpos(strtoupper($join['table']), '(SELECT') === 0 ? $join['table'] : "\"" . $join['table'] . "\"";
```

cc: @senorgeno

tractorcow:
Fix PHPDoc on SQLQuery::addFrom and SQLQuery::setFrom
Fixes #3965

### DIFF
--- a/model/SQLQuery.php
+++ b/model/SQLQuery.php
@@ -239,10 +239,10 @@ class SQLQuery {
 	/**
 	 * Set table for the SELECT clause.
 	 *
-	 * @example $query->setFrom("MyTable"); // SELECT * FROM MyTable
+	 * @example $query->setFrom('"MyTable"'); // SELECT * FROM "MyTable"
 	 *
-	 * @param string|array $from Escaped SQL statement, usually an unquoted table name
-	 * @return SQLQuery
+	 * @param string|array $from Single, or list of, ANSI quoted table names
+	 * @return self
 	 */
 	public function setFrom($from) {
 		$this->from = array();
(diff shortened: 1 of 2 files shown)


model/SQLQuery.php
```diff
@@ -903,9 +903,13 @@ public function sql() {
...
 
-				$table = strpos(strtoupper($join['table']), 'SELECT') ? $join['table'] : "\"" 
-					. $join['table'] . "\"";
-				$aliasClause = ($alias != $join['table']) ? " AS \"" . Convert::raw2sql($alias) . "\"" : "";
+				// Ensure tables are quoted, unless the table is actually a sub-select
+				$table = preg_match('/\bSELECT\b/i', $join['table'])
+					? $join['table']
+					: "\"{$join['table']}\"";
+				$aliasClause = ($alias != $join['table'])
+						? " AS \"{$alias}\""
```
tractorcow: Changed because raw2sql is not the Ansi standard way to escape SQL identifiers.

http://savage.net.au/SQL/sql-99.bnf.html#delimited%20identifier

dhensby: Well... if the tests pass

dhensby:
Well... if the tests pass

PR #4025, opened 2015-03-20
FIX Don't redirect in CLI when forceWWW or forceSSL is used

micmania1:
At the moment it doesn't check to see if we're running cli before redirecting.

### DIFF
--- a/control/Director.php
+++ b/control/Director.php
@@ -841,6 +841,7 @@ class Director implements TemplateGlobalProvider {
 	 * @return boolean|string String of URL when unit tests running, boolean FALSE if patterns don't match request URI
 	 */
 	public static function forceSSL($patterns = null, $secureDomain = null) {
+		if(Director::is_cli()) return false;
 		if(!isset($_SERVER['REQUEST_URI'])) return false;
 		
 		$matched = false;
@@ -889,7 +890,11 @@ class Director implements TemplateGlobalProvider {
 	 * Force a redirect to a domain starting with "www."
 	 */
 	public static function forceWWW() {
-		if(!Director::isDev() && !Director::isTest() && strpos($_SERVER['HTTP_HOST'], 'www') !== 0) {
+		if(!Director::is_cli() 
+			&& !Director::isDev() 
+			&& !Director::isTest() 
+			&& strpos($_SERVER['HTTP_HOST'], 'www') !== 0
+		) {
 			$destURL = str_replace(Director::protocol(), Director::protocol() . 'www.', 
 				Director::absoluteURL($_SERVER['REQUEST_URI']));
 


control/Director.php
```diff
@@ -889,7 +890,11 @@ public static function forceSSL($patterns = null, $secureDomain = null) {
 	 * Force a redirect to a domain starting with "www."
 	 */
 	public static function forceWWW() {
-		if(!Director::isDev() && !Director::isTest() && strpos($_SERVER['HTTP_HOST'], 'www') !== 0) {
+		if(!Director::is_cli() 
+			&& !Director::isDev()
```
tractorcow: Isn't this the same as saying `Director::isLive()` only?

micmania1: Technically, no. The environment type could be set to 'blah'.

dhensby: > The environment type could be set to 'blah'.

It *could* be, but that would be wrong. `Director::get_environment_type()` will just return false if it's not one of the 3 expected types.

---

I don't see why this function is dictating the environments this should work on. It's not `forceSSLIfNotTestOrDev`. The developer should make the call if they want to force www on the test or live environments.

Having functionality that only works on one stage can hide bugs and cause unexpected and un-testable results.

dhensby:
TBH, I'd just deprecate this function. People should be doing it at a lower level (like .htaccess). If they can't then they can add a custom script to do it.

The protection against redirecting on the CLI should be done in one place (where HTTP redirects are sent to the browser) NOT each time we call a redirect. Fix it at source.

control/Director.php
tractorcow: Also we have some test environments that have ssl :)

sminnee:
We're not removing this functionality  in 3.x, and I disagree with leaving it deprecated & broken in 3. So deprecating it doesn't seem to help the problems we have right now, it will just annoy those who are using it. Let's put a pin in that discussion as it's not relevant to this PR. If someone wants to raise an issue about deprecating/refactoring these features, feel free.

IMO, `forceWWW()` and `forceSSL()` should be thought of as configuration directives implemented as static methods. They should silently no-op on CLI, so I think the change that Michael has produced makes sense.

On the other hand, `Director::force_redirect()` is a method that shouldn't be called on Director::is_cli() and should throw an Exception if called when Director::is_cli() is true. This will ensure that if it's called elsewhere it will be obvious what has happened. Can we add this to the patch?

Can we also get the tests fixed.

control/Director.php
sminnee: > I don't see why this function is dictating the environments this should work on. It's not forceSSLIfNotTestOrDev. The developer should make the call if they want to force www on the test or live environments.

SS4 change.

tractorcow:
@sminnee I think that we should consider the case of is_cli() = true but also is_running_tests() = true.

The return shortcut should probably ONLY short circuit the code if running as CLI AND not running tests. When running unit tests it should behave as you would predict a normal web request would. See the end of forceSSL() for example.

Otherwise, the method is simply not testable.

sminnee:
Right now the behaviour of the test response is to return the destination URL as a string. I guess we can do that for `forceWWW()` as well—as you say it's best not to change that in 3.x, and it's best to have `forceWWW()` and `forceSSL()` operate similarly—but it's a bit hacky approach.

A nicer fix might be to make more use of `SS_HTTPRepsonse_Exception` object embedding the appropriate redirections, in both test and production scenarios. But let's leave that out of scope for this PR.

chillu:
@micmania1 Can you implement the small amount of changes required to get this over the line?

- Sam's suggestion to throw an exception on `Director::force_redirect()`
- Dan/Damian's suggestion not to check for `isDev()` in `forceSSL()`
- Fix tests
- Technically, if 3.3.0 comes out before this is merged, the pull request should be rebased on 3.2 - but it seems like a low risk bugfix for 3.1 either way.

dhensby:
is throwing an exception not a violation of SemVer?

tractorcow:
I disagree with @chillu ; We shouldn't be merging non-critical fixes into 3.1.

@dhensby technically it would be a change in behaviour, so 3 branch at earliest.

3.3.0 is coming out this week, in any case.

dhensby:
Do we have any formal EoL roadmap for versions? Why is 3.1 critical fixes only now? I think it's in wide enough use that bugs should be fixed even if they aren't critical

tractorcow:
Uh, this was to be discussed at the core hangout. :)

control/Director.php
```diff
@@ -841,6 +841,7 @@ protected static function force_redirect($destURL) {
 	 * @return boolean|string String of URL when unit tests running, boolean FALSE if patterns don't match request URI
 	 */
 	public static function forceSSL($patterns = null, $secureDomain = null) {
+		if(Director::is_cli()) return false;
```
tractorcow: This needs to return the $destURL when running tests. You need to make return false ONLY if not running tests.

tractorcow:
I gave this another review to see what we can do to get it over the line; I found a few issues with running during testing.

Maybe you can move the check to force_redirect, rather than returning at the top level?

control/Director.php
tractorcow: Move your check to force_redirect and that's the only change you should need.

dhensby:
this was auto-closed when we deleted the 3.4 branch. Are we interested in getting this in?

tractorcow:
Replaced with https://github.com/silverstripe/silverstripe-framework/pull/7578

PR #4029, opened 2015-03-23
[DO NOT MERGE] API Removed SS_ConfigStaticManifest in favour of Reflection

micmania1:
Apologies for the "do not merge", but this is the best way to show a diff and include as many people as possible.

Anyway, i'm hoping that this PR is the start of a discussion to remove SS_ConfigStaticManifest and its flush process.

Instead of collecting private statics through parsing PHP files during the flush process, this PR uses Reflection to get private static config values at run-time when they're requested. This doesn't remove or affect YAML configuration in any way.

Initial performance tests show a small speed increase and ~0.6mb memory saving:

Before:

	Parameter			Min.		Max.		Avg.
	CPU Time			85.104 ms	180.61 ms	94.278 ms
	Memory Usage		9.97 M		10 M		9.98 M
	Peak Memory Usage	10.07 M		10.1 M		10.07 M
	Dataset Size		50

After:

	Parameter			Min.		Max.		Avg.
	CPU Time			85.052 ms	181.952 ms	92.791 ms
	Memory Usage		9.35 M		9.38 M		9.36 M
	Peak Memory Usage	9.34 M		9.49 M		9.46 M
	Dataset Size		50

These were ran using php 5.5.

This is only a proof of concept, but if nobody raises any issues with this method i'll turn this into a proper PR. Feedback welcome.

### DIFF
--- a/core/Config.php
+++ b/core/Config.php
@@ -489,14 +489,14 @@ class Config {
 			}
 		}
 
-		$value = $nothing = null;
+		$value = null;
 
 		// Then the manifest values
 		foreach($this->manifests as $manifest) {
-			$value = $manifest->get($class, $name, $nothing);
-			if ($value !== $nothing) {
+			$value = $manifest->get($class, $name, null);
+			if ($value !== null) {
 				self::merge_low_into_high($result, $value, $suppress);
-				if ($result !== null && !is_array($result)) return $result;
+				if (!is_null($result) && !is_array($result)) return $result;
 			}
 		}
 
@@ -514,22 +514,23 @@ class Config {
 			if ($extraSources) $sources = array_merge($sources, $extraSources);
 		}
 
-		$value = $nothing = null;
-
 		foreach ($sources as $staticSource) {
-			if (is_array($staticSource)) {
-				$value = isset($staticSource[$name]) ? $staticSource[$name] : $nothing;
-			}
-			else {
-				foreach ($this->staticManifests as $i => $statics) {
-					$value = $statics->get($staticSource, $name, $nothing);
-					if ($value !== $nothing) break;
+			if(is_array($staticSource)) {
+				$value = isset($staticSource[$name]) ? $staticSource[$name] : null;
+			} else if(class_exists($staticSource) && property_exists($staticSource, $name)) {
+				$property = new ReflectionProperty($staticSource, $name);
+				if($property->isStatic() && $property->isPrivate()) {
+					$property->setAccessible(true);
+					$value = $property->getValue();
 				}
 			}
 
-			if ($value !== $nothing) {
+			if (!is_null($value)) {
 				self::merge_low_into_high($result, $value, $suppress);
-				if ($result !== null && !is_array($result)) return $result;
+				if (!is_null($result) && !is_array($result)) return $result;
+
+				// Reset value. If we don't do this, the value array may have double values.
+				$value = null;
 			}
 		}
 
@@ -541,7 +542,6 @@ class Config {
 			$parent = get_parent_class($class);
 			if ($parent) $this->getUncached($parent, $name, $sourceOptions, $result, $suppress, $tags);
 		}
-
 		return $result;
 	}
 
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
 		}
 	},
 	"autoload": {
-		"classmap": ["tests/behat/features/bootstrap"]	
+		"classmap": ["tests/behat/features/bootstrap"]
 	},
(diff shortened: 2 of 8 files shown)


core/Config.php
```diff
@@ -514,16 +514,25 @@ protected function getUncached($class, $name, $sourceOptions, &$result, $suppres
...
-			else {
-				foreach ($this->staticManifests as $i => $statics) {
-					$value = $statics->get($staticSource, $name, $nothing);
-					if ($value !== $nothing) break;
+			} else if(class_exists($staticSource) && property_exists($staticSource, $name)) {
+				$property = new ReflectionProperty($staticSource, $name);
+				if($property->isStatic() && $property->isPrivate()) {
+					$property->setAccessible(true);
+					$value = $property->getValue();
+					$property->setAccessible(false);
```
dhensby: I'd have thought this was unnecessary as it doesn't actually change the visibility of the property for the class... 

citation: http://php.net/manual/en/reflectionproperty.setaccessible.php#110414

dhensby:
We're getting a rather strange test failure and I'd expect we aren't seeing more as the test coverage is probably not as good as it should be for this part of the system.

dhensby:
Please can you elaborate on how you tested this and came to conclude it was faster and used less memory? Did you test on a production site (ie: with lots of functionality) or on a clean install?

micmania1:
I tested using a base install of master on PHP 5.5 using xhprof.

The goal of this wasn't actually to improve performance initially - the test was to ensure that this doesn't introduce any new performance issues. The memory saving was expected but also I would expect the flush process to take less time. 

The long goal is to improve performance but the initial goal is to simplify things - parsing PHP files to get static values seems a bit OTT.

The test failures make sense. This solution is case-sensitive so that's something I'd need to consider in the final solution. I haven't looked into how the case-insensitivity is currently implemented so that'll be next on my list :)

micmania1:
@hafriedlander can you take a look at this when you get a chance?

kinglozzer:
Alternative suggestion (5.4+ only, though): [`Closure::bind()`](http://php.net/manual/en/closure.bind.php).

```php
// Note: actual code would store each $thief (as one is needed for each class)
$thief = function($name) {
	return self::$$name;
};

$thief = Closure::bind($thief, null, $staticSource);
$value = $thief($name);
```

Just had a quick play with it, seems to be slightly (~3%) more performant than `ReflectionProperty`. Uses ~0.3% more memory than this solution (~66kb more), but still uses less memory than the existing static manifests. Only benchmarked on one site - I’ll play around with some that vary in complexity :)

core/Config.php
```diff
@@ -514,22 +514,23 @@ protected function getUncached($class, $name, $sourceOptions, &$result, $suppres
...
-			if (is_array($staticSource)) {
-				$value = isset($staticSource[$name]) ? $staticSource[$name] : $nothing;
-			}
-			else {
-				foreach ($this->staticManifests as $i => $statics) {
-					$value = $statics->get($staticSource, $name, $nothing);
-					if ($value !== $nothing) break;
+			if(is_array($staticSource)) {
+				$value = isset($staticSource[$name]) ? $staticSource[$name] : null;
+			} else if(class_exists($staticSource) && property_exists($staticSource, $name)) {
```
kinglozzer: I think the `class_exists()` check here is redundant: `property_exists()` will return false if the class doesn’t exist.

core/Config.php
```diff
@@ -514,22 +514,23 @@ protected function getUncached($class, $name, $sourceOptions, &$result, $suppres
 			if ($extraSources) $sources = array_merge($sources, $extraSources);
 		}
 
-		$value = $nothing = null;
```
kinglozzer: Another observation: `$value` needs to be reset to `null` inside the `foreach($sources...` loop: it *was* being set back to `null` by `$value = $statics->get($staticSource, $name, $nothing);`, but that has been removed.

chillu:
Given the complexity and inherent brittleness of `SS_ConfigStaticManifest_Parser`, I'm generally in favour of this change. We'd no longer be at the mercy of a correctly implemented regex which can parse all possible syntax variations. It'll also open up more possibilities around annotations.

I have no strong opinions on `Reflection` vs. `Closure::bind()`.

Have you tested on an SSD? I'm not sure how PHP 5.5's built-in opcache works, i.e. if `scandir()` and `file_get_contents()` in `ManifestFileFinder` incur a higher performance penalty on slower disks.

I guess there's no need to test this against PHP versions older than 5.5 (with potentially slower Reflection implementations), since a 4.x release is unlikely to support them.

PR #4030, opened 2015-03-23
FIX Debug::text no longer incorrecty returns "ViewableData_debugger"

dhensby:
Fixes an issue when `Debug::show($var)` would sometimes just return `ViewabledData_Debugger` instead of debug info.

This was because `debug()` on the base `ViewableData` returns an object and not a string.

### DIFF
--- a/dev/Debug.php
+++ b/dev/Debug.php
@@ -1,34 +1,34 @@
 <?php
 /**
  * Supports debugging and core error handling.
- * 
+ *
  * Attaches custom methods to the default error handling hooks
  * in PHP. Currently, two levels of error are supported:
- * 
+ *
  * - Notice
  * - Warning
  * - Error
- * 
+ *
  * Uncaught exceptions are currently passed to the debug
  * reporter as standard PHP errors.
- * 
+ *
  * Errors handled by this class are passed along to {@link SS_Log}.
  * For configuration information, see the {@link SS_Log}
  * class documentation.
- * 
+ *
  * @todo add support for user defined config: Debug::die_on_notice(true | false)
  * @todo better way of figuring out the error context to display in highlighted source
- * 
+ *
  * @package framework
  * @subpackage dev
  */
 class Debug {
-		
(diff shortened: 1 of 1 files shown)


tractorcow:
Instead of adding a single if-case, a more generic solution could be to cast the result to string, and then just implement __toString in ViewableData_Debugger. 

Also, ViewableData doesn't have a forTemplate method. :)

dhensby:
Teaches me for trying to submit a quick patch!

`ViewableData` doesn't have *the* cornerstone of rendering in the Framework? Who'd have thought it...

Ok, will cast as a string and add `__toString()` to the debugger. Will have to go against 3.

dhensby:
Actually, I'll fix this for 3.1 and provide a more robust fix for 3

dhensby:
in fact, it's already cast as a string, which is why it appears as "VieableData_Debugger" when `Debug::show()` is used (the default `__toString` is to return the classname) so I should only need to implement a `__toString` on the debugger to fix this in 3

tractorcow:
I'm going to say it needs to be fixed in 3.1 :)

dhensby:
@tractorcow this fixes it in 3.1.x-dev and the other PR fixes it in 3.x-dev. Best of both?

tractorcow:
I think the 3.2 fix in 3.1 would end up with a less tightly coupled solution.

dhensby:
True, but it's a new feature, so can't go into 3.1

tractorcow:
Then can you generalise this solution so it's not tightly bound to ViewableData_debugger, but it still solves the problem?

how about `if($debug instanceof ViewableData && $debug->hasMethod('forTemplate'))`?

tractorcow:
How about add a simple note then

// Not necessary after 3.2. See https://github.com/silverstripe/silverstripe-framework/pull/4034

Just so devs can see this code can be retired later :)

dhensby:
done

PR #4031, opened 2015-03-23
FIX GridFieldDetailForm::setItemEditFormCalback broke chaining

dhensby:
This is an inconsistent API!

This is a bug as `set` methods should return `$this`


kinglozzer:
Dunno why the Travis notification is stuck on pending, the tests passed!

dhensby:
Sometimes the webhooks are slow.. What is really annoying is when it won't let you merge on mobile because of it ;)

PR #4032, opened 2015-03-24

unclecheese:

```php
});
```

### DIFF
--- a/model/DataQuery.php
+++ b/model/DataQuery.php
@@ -748,7 +748,7 @@ class DataQuery {
 	 * @param $fieldExpression String The field to select (escaped SQL statement)
 	 * @param $alias String The alias of that field (escaped SQL statement)
 	 */
-	protected function selectField($fieldExpression, $alias = null) {
+	public function selectField($fieldExpression, $alias = null) {
 		$this->query->selectField($fieldExpression, $alias);
 	}
 


tractorcow:
I agree, but can we put this into 3 branch?

I suppose there's no harm in this being public if other methods (like selectFromTable) are already public though.

unclecheese:
Will close and resubmit to 3.

micmania1:
Shouldn't API changes like this go into master? They're potentially breaking backwards compatibility.

unclecheese:
I can see that it would be breaking if we were decreasing the visibility, but I can't think of a use case where increasing visibility would break anything.

micmania1:
class MyClass extends DataQuery {
        protected function selectField($fieldExpression, $alias = null) {}
    }

I believe this would now break because you're making a public method protected in a subclass.

unclecheese:
Yeah, I just ran a test, and you're correct. Will wait for @tractorcow 's call on this.

tractorcow:
They don't have to go into master. 3.2 (3 branch) is fine for backwards compatible API.

micmania1:
The framework is supposed to follow semantic versioning so it does need to go into master.

> Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are 
> introduced to the public API.

Its unfair to make assumptions about how others are using this codebase. Semantic versioning was adopted for a reason, with lots of input from the community and I think the core team should remain disciplined in implementing it.

dhensby:
@micmania1 is right, this would break someone's extension.

tractorcow:
It seems extremely unlikely, but I guess we may as well be as strict as we can with semver (at least when the option is afforded us).

unclecheese:
OK, I'm pretty Git illiterate, as evidenced by my duplicate PR. So my apologies, but where should I submit this now? Master? Which means this doesn't ship until 4.0?

dhensby:
@unclecheese - yes, I think that's the general concensus (that it will go into master).

You'd need to do something like this to get it against master:

```sh
$ git checkout patch-9
$ git pull --rebase upstream master
$ git push -f
```

Then when you create the PR you make sure it's against master

unclecheese:
Done! Thanks, @dhensby

PR #4033, opened 2015-03-24

unclecheese:

```php
});
```


tractorcow:
Deja vu?

kinglozzer:
This has been merged in #4038

PR #4034, opened 2015-03-24
NEW ViewableData_Debugger implements __toString

dhensby:
This fixes an issue were ViewableData can be 'debugged' and no information is output - related #4030

### DIFF
--- a/view/ViewableData.php
+++ b/view/ViewableData.php
@@ -700,7 +700,14 @@ class ViewableData_Debugger extends ViewableData {
 		$this->object = $object;
 		parent::__construct();
 	}
-	
+
+	/**
+	 * @return string The rendered debugger
+	 */
+	public function __toString() {
+		return $this->forTemplate();
+	}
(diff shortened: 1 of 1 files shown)


willmorgan:
Are there any precautions that we can take to prevent exceptions/errors being thrown?

IIRC, if an error is thrown during `__toString`, then 'Exception thrown without a stack frame' can potentially be shown to the user.

dhensby:
What kind of exceptions are you expecting, @willmorgan ?

tractorcow:
I think this seems fine. If we encounter errors we can probably re-look at this again in the future.

willmorgan:
Template errors for starters. But hey, merged now...

tractorcow:
Sorry @willmorgan . Do you want to have a go at changing this if you feel you'd like more error checking? If you think it's needed, I'll accept another PR with exception handling if you would like to submit one. :)

PR #4038, opened 2015-03-25

unclecheese:

```php
});
```

### DIFF
--- a/model/DataQuery.php
+++ b/model/DataQuery.php
@@ -789,7 +789,7 @@ class DataQuery {
 	 * @param $fieldExpression String The field to select (escaped SQL statement)
 	 * @param $alias String The alias of that field (escaped SQL statement)
 	 */
-	protected function selectField($fieldExpression, $alias = null) {
+	public function selectField($fieldExpression, $alias = null) {
 		$this->query->selectField($fieldExpression, $alias);
 	}
 


sminnee:
Is changing the visibility of a method from protected to public a BC-breaking API change?

unclecheese:
Yes. In theory, if someone subclassed DataQuery, and overloaded the method
using the parent visibility, that subclass would break.

tractorcow:
In merging this I completely forgot about the prior discussion I had about this. :) Oh well it's merged now!

patricknelson:
It looks like this was applied only to `4.0`. Is there a technique for adding your own custom select field to `DataList` some other way in `3.x`? Currently only way I can see doing this is via:

```php
$list = $list->alterDataQuery(function (DataQuery $dataQuery) {
	// ... protected in 3.x
	$dataQuery->selectField('MyAdvancedColumn');
});
```

😭

patricknelson:
I've overcome this for now by proxy of manually instantiating `DataList` since I only need it for that class. Unfortunately using `Injector` will not work on `DataList` because of this bug:

So my solution is basically create a function to instantiate my custom classes for me:
```php
function getCustomDataList($forObject) {
	$result = new CustomDataList($forObject); // Uses CustomDataQuery
	$result->setDataModel(DataModel::inst()); // TODO: This will probably not work in SS 4.0
	return $result;
}
```

**CustomDataList:**
```php
class CustomDataList extends DataList {
	public function __construct($dataClass) {
		parent::__construct($dataClass);

		// Override data query with our own custom instance.
		$this->dataQuery = new CustomDataQuery($this->dataClass);
	}
}
```

**CustomDataQuery:**

```php
class CustomDataQuery extends DataQuery {
	/**
	 * Overridden just so we can make it public.
	 *
	 * @param String $fieldExpression
	 * @param null $alias
	 */
	public function selectField($fieldExpression, $alias = null) {
		parent::selectField($fieldExpression, $alias);
	}
}
```

And then... finally... you will have access to `->selectField()` without altering core:

```php
$list = getCustomDataList('MyDataObject');`
$list = $list->alterDataQuery(function (CustomDataQuery $dataQuery) {
	// Accessible in the CustomDataQuery class thanks to function above.
	$dataQuery->selectField('MyAdvancedColumn');
});
```

PR #4039, opened 2015-03-25
Add more 3xx status codes to SS_HTTPResponse::isFinished() method

jacobbuck:
This is so `Controller::handleRequest()` will halt processing for more redirect types, including `304 Not Modified` :smile:

### DIFF
--- a/control/HTTPResponse.php
+++ b/control/HTTPResponse.php
@@ -286,7 +286,7 @@ EOT
 	 * @return bool
 	 */
 	public function isFinished() {
-		return in_array($this->statusCode, array(301, 302, 401, 403));
+		return in_array($this->statusCode, array(301, 302, 303, 304, 305, 307, 401, 403));
 	}
 	
 }


willmorgan:
I suspect this should be placed against 3.x as it constitutes a functionality change.

tractorcow:
I'd go one step further and move the 'finished_response_codes' into a config that can be customised on a per-project basis.

mateusz:
Arguably, if you read the rationale for the function, this sounds like a bug. This function is supposed to "return true if a redirect has already been requested", so I'd expect it to return true for any 3xx status code (which are all redirection codes).

jacobbuck:
@mateusz True, but I specifically only added 3xx status codes from the HTTP/1.1 spec.

tractorcow:
Consulting the spec seems to indicate that the old behaviour of the code only supported http 1.0 codes, where this pr upgrades the method to support http 1.1. Perhaps we could say the code is outdated and this fix is necessary to bring behaviour in line with current specs?

willmorgan:
It's changing the underlying behaviour of `isFinished`. Any code that relies on that - which I don't think would be that rare, could break if it's not looking out for 304s in particular.

I still think this should go in to 3, not 3.1. Sorry!

jacobbuck:
@tractorcow except 304 (which was missing) has existed since the original HTTP/0.9 spec :wink: 

@willmorgan I actually made the pull-request because it improves the performance of 304's

dhensby:
I've got to agree that `isFinished` not returning `true` for all valid redirect codes is a bug (sorry @willmorgan).

I agree with @tractorcow that we should have config setting for "finished codes", but that is a new feature for 3.x-dev

mateusz:
:+1:

tractorcow:
I'm happy :)

A piece of feedback for the future, @jacobbuck,  is that any bug fixes should have "BUG" or "FIX" in the commit message, as this helps us to build the changelog. 

http://doc.silverstripe.org/en/contributing/code/#commit-messages

Thanks!

jacobbuck:
Awesome, thanks guys. And thanks for the feedback, will keep that in mind next time :+1:

PR #4041, opened 2015-03-26
BUG: export and print buttons outside button row

adrexia:
Export and print buttons are appearing outside the button row in model admin, meaning that if the add button is removed (say, by removing the create permission), the buttons are flush with the gridfield.

before:
silverstripe_bug

after:
silverstripe_bugafter


adrexia:
This is a really easy one and it involves no css if that's what people are afraid of. Anyone? ;)

tractorcow:
Hi Naomi :) Sure all done!

adrexia:
Thanks! :)

PR #4042, opened 2015-03-26
Hardcoded http:// cause browser warnings in admin panel 

selay:
Two links Help button (http://userhelp.silverstripe.org/framework/en/3.1) and logo (http://www.silverstripe.org/) are hardcoded with protcol defined in admin/code/LeftAndMain.php
private static $application_link = 'http://www.silverstripe.org/';
private static $help_link = 'http://userhelp.silverstripe.org/framework/en/3.1';
Many use https for the admin panel and some browsers (Internet explorer all versions) popup and say connection is not secure. 
In fact, it is nothing but it scares people. 
Changed http:// to //: in both links.

### DIFF
--- a/admin/code/LeftAndMain.php
+++ b/admin/code/LeftAndMain.php
@@ -76,7 +76,7 @@ class LeftAndMain extends Controller implements PermissionProvider {
 	 * @config
 	 * @var string
 	 */
-	private static $help_link = 'http://userhelp.silverstripe.org/framework/en/3.1';
+	private static $help_link = '//userhelp.silverstripe.org/framework/en/3.1';
 
 	/**
 	 * @var array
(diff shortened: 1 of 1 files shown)


tractorcow:
The trouble is there is no https://userhelp.silverstripe.org at the moment, so there's no real point supporting a https link to it. The warning given on that page is probably going to be just as scary as the existing notice.

selay:
I see. If https is not going to be enabled on that server soon, then how about setting a JavaScript to build the link or add the protocol on click :)
Then browser may not see it on normal load. 
Well, I know it is not a good solution but neither is the current situation.

wolfv:
Maybe it would also be an Option to implement a 301 redirect in silverstripe? E.g. the link would be to /admin/help and then redirect to the URL ...

selay:
Silverstripe.org wihtout subdomain has ssl enabled. 
How about redirecting here: https://www.silverstripe.org/community/ or https://www.silverstripe.org/blog/

tractorcow:
www.silverstripe.org is fine as https, but the help link still needs to go to the documentation. We can't send them to the blog or the generic community section.

I think just fix that one link to ss.org and leave the other?

kinglozzer:
@tractorcow HTTPS seems to be up and running for silverstripe.org and userhelp, any objections to merging this now? :)

selay:
Looks good.

PR #4045, opened 2015-03-30
Don't submit when disabled buttons are clicked

flashbackzoo:


javascript/GridField.js
```diff
@@ -136,6 +136,11 @@
 		$('.ss-gridfield .action').entwine({
 			onclick: function(e){
 				var filterState='show'; //filterstate should equal current state.
+
+				// If the button is disabled, do nothing.
+				if (this.button('option', 'disabled')) {
+					return;
```
dhensby: Please use `ev.preventDefault()`

flashbackzoo: Replacing the `return` with `e.preventDefault()` won't work. At the bottom of the handler `this.getGridField().reload()` will still get called unless you return. I could use both but I'm not sure what default action would be cancelled here?

flashbackzoo:
I've added the changes from https://github.com/silverstripe/silverstripe-cms/pull/1180 to this pull request.

javascript/GridField.js
tractorcow: You need to do both then :)

javascript/GridField.js
```diff
@@ -146,6 +152,37 @@
...
+			onunmatch: function () {
+				this.find('input[type="text"]').unbind('keyup');
+			},
+			toggleDisabled: function () {
+				var $button = this.find('button[type="submit"]'),
+					$input = this.find('input[type="text"]'),
+					inputValue = $input.val(),
+					buttonDisabled = $button.prop('disabled');
+
+				if ((inputValue === '' && buttonDisabled === false) ||
```
tractorcow: It would probably work fine without any condition here. Does it save any work if this evaluates to false?

javascript/GridField.js
```diff
@@ -146,6 +152,37 @@
...
+		/**
+		 * Don't allow users to submit empty values in grid field auto complete inputs.
+		 */
+		$('.ss-gridfield .add-existing-autocompleter').entwine({
+			onbuttoncreate: function () {
+				var self = this;
+
+				this.toggleDisabled();
+
+				this.find('input[type="text"]').bind('keyup', function () {
```
tractorcow: I think we should use 'on' instead of 'bind', as it is more ajax friendly.

https://api.jquery.com/on/

javascript/GridField.js
```diff
@@ -146,6 +152,37 @@
...
+				});
+			},
+			onunmatch: function () {
+				this.find('input[type="text"]').unbind('keyup');
+			},
+			toggleDisabled: function () {
+				var $button = this.find('button[type="submit"]'),
+					$input = this.find('input[type="text"]'),
+					inputValue = $input.val(),
+					buttonDisabled = $button.prop('disabled');
```
tractorcow: I like `$button.is(':disabled')` as it's guaranteed to return a boolean value.

javascript/GridField.js
```diff
@@ -146,6 +152,37 @@
 			}
 		});
 
+		/**
+		 * Don't allow users to submit empty values in grid field auto complete inputs.
+		 */
+		$('.ss-gridfield .add-existing-autocompleter').entwine({
```
tractorcow: You have a lot of .find('input') in this class, plus you are having to bind an event to a child element. Why not promote the textbox to the top level.

`$('.ss-gridfield .add-existing-autocompleter input[type="text"]')`

Then you can use `onkeyup` as a top level event. You can still use `onmatch` to do your initial loading, and `$button = this.parents('.add-existing-autocompleter`).find('input[type="submit"]')` to get the button.

javascript/GridField.js
flashbackzoo: yeah fair enough, i'll change it, it is a bit slower though. https://jsperf.com/prop-vs-ischecked/52

tractorcow:
There seems to be something odd going on with the behat in travis. Let's investigate before we finish merging this.

https://travis-ci.org/silverstripe/silverstripe-framework/jobs/56474659

flashbackzoo:
The error can be reproduced as follows:
- Go to /admin/security
- Select the 'Groups' tab
- Click into a group in the grid field

The error happens because the auto completer has two buttons. The visible 'link existing' button and a hidden 'find' button. The 'find' button is not the button we want to be dealing with but is included in the `$button` variable because the selector being used is `button[type="submit"]`.

By using the `.ss-ui-button` selector instead, we make sure we're dealing the the initialised button.

PR #4046, opened 2015-03-30
Updated example code

ehyland:
HomePageTest should extend FunctionalTest.
Removed "#" from example form ID in `FunctionalTest::submitForm()`


dhensby:
Thanks

PR #4051, opened 2015-03-31
API Security better respects BackURL on login

tractorcow:
Also restore missing authentication message not appearing in the login form $Content area (regression from #1807)


tractorcow:
I almost thought about putting this in 3.1, but affecting login redirection is a risk of introducing more regressions.

This can be implemented as an extension if it's needed in 3.1, but I'd like some peer review before doing this.

cc @assertchris

assertchris:
The longest travis run...

tractorcow:
Thanks @dhensby

PR #4052, opened 2015-03-31
FEATURE: Added maximum upload file size by type (on 3)

Turnerj:
See previous chat on #3980 about this change. Now against the 3 branch instead of 3.1 with no merge conflicts.

Changes include document updates and tests (a lot of tests).


dhensby:
great stuff! thanks

Issue #3084, opened 2014-05-01
Parameters passed to templates are always true, even when they are explicitly false


oddnoc:
Consider parts of three files:

1. Page.php (Page_Controller)
```php		
public function Truthiness() {
	return true;
}

public function Falsiness() {
	return false;
}
```

2. Layout/Page.ss
```
<% include Truthiness No=$Falsiness, Yes=$Truthiness %>
```
3. Includes/Truthiness.ss
```
<%-- Truthiness.ss --%>
<p>
<% if $No  %>War is Peace, because "$No" is truthy.<% end_if %><br>
<% if $Yes %>Logic still works, because "$Yes" is truthy.<% end_if %>
</p>
```
The output of this is as follows:
```html
<p>
War is Peace, because "" is truthy.<br>
Logic still works, because "1" is truthy.
</p>
```

oddnoc:
Is there anything to be done about this?

hafriedlander:
This should be fixable. In fact, I'm not sure why it happens just from looking at the code. At https://github.com/silverstripe/silverstripe-framework/blob/3.1/view/SSViewer.php#L519 the boolean cast should make an empty string become false.

tractorcow:
Duplicate of https://github.com/silverstripe/silverstripe-framework/pull/4053

tractorcow:
Also, not trying to be dismissive... I just want to make sure we collect all of the various symptoms of the same issue in one place. :D I'm open to the idea that this might not be a duplicate, but if you think it is I'll re-open it.

tractorcow:
Ok, turns out it was the same problem, but a different part of the code caused it. :)

In any case it's fixed in #4053 . Let's leave this open until the other is merged also.

oddnoc:
Sounds great. FWIW, I can confirm the fix on my test case. Thanks!

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-framework/pull/4053

PR #4053, opened 2015-04-01
BUG Fix default casted (boolean)false evaluating to true in templates

tractorcow:
Because 'Text' is the default_cast on ViewableData, in some cases boolean false could evaluate to 'exists'  in templates.

E.g. where $MyVar type may not be known to the template

```html
<% if $MyVar.exists %>
MyVar = $MyVar
<% end_if %>
```

```php
public function getMyVar() { return false; }
```

Would print out `My Var = ` if internally it's a boolean false.

The bug is caused by the value internally evaluating `false == '0'` to true, because of the loose comparison.

If you omit the .exists above then the SSViewer will internally call .exists on the object, assuming it's given a $casting helper explicitly. A work around in the mean time is to use null instead of booleans to denote empty values.

Also removes the nullifyIfEmpty exception since in no cases should an empty string cause this field to evaluate to exists, otherwise the same issue arises with strings instead of booleans.

The replaced method uses the same logic as RequiredFields to check for valid values, meaning (string)'0' is treated as a non-empty value.

### DIFF
--- a/model/fieldtypes/StringField.php
+++ b/model/fieldtypes/StringField.php
@@ -83,7 +83,9 @@ abstract class StringField extends DBField {
 	 * @see core/model/fieldtypes/DBField#exists()
 	 */
 	public function exists() {
-		return ($this->value || $this->value == '0') || ( !$this->nullifyEmpty && $this->value === '');
+		return $this->getValue() // All truthy values exist
+			|| (is_string($this->getValue()) && strlen($this->getValue())) // non-empty strings exist ('0' but not (int)0)
+			|| (!$this->getNullifyEmpty() && $this->getValue() === ''); // Remove this stupid exemption in 4.0
 	}
 
 	/**
--- a/tests/templates/SSViewerTestIncludeScopeInheritanceInclude.ss
+++ b/tests/templates/SSViewerTestIncludeScopeInheritanceInclude.ss
@@ -1,1 +1,1 @@
-$Title <% if ArgA %>- $ArgA <% end_if %>- <%if First %>First-<% end_if %><% if Last %>Last-<% end_if %><%if MultipleOf(2) %>EVEN<% else %>ODD<% end_if %> top:$Top.Title
+<% if $Title %>$Title<% else %>Untitled<% end_if %> <% if $ArgA %>_ $ArgA <% end_if %>- <% if $First %>First-<% end_if %><% if $Last %>Last-<% end_if %><%if $MultipleOf(2) %>EVEN<% else %>ODD<% end_if %> top:$Top.Title
--- a/view/SSViewer.php
+++ b/view/SSViewer.php
@@ -429,6 +429,15 @@ class SSViewer_DataPresenter extends SSViewer_Scope {
 		}
 	}
 
+	/**
+	 * Get the injected value
+	 * 
+	 * @param string $property Name of property
+	 * @param array $params
+	 * @param bool $cast If true, an object is always returned even if not an object.
+	 * @return array Result array with the keys 'value' for raw value, or 'obj' if contained in an object
+	 * @throws InvalidArgumentException
+	 */
 	public function getInjectedValue($property, $params, $cast = true) {
 		$on = $this->itemIterator ? $this->itemIterator->current() : $this->item;
 
@@ -512,31 +521,24 @@ class SSViewer_DataPresenter extends SSViewer_Scope {
 		if (isset($arguments[1]) && $arguments[1] != null) $params = $arguments[1]; 
 		else $params = array();
 
-		$hasInjected = $res = null;
-
-		if ($name == 'hasValue') {
-			if ($val = $this->getInjectedValue($property, $params, false)) {
-				$hasInjected = true; $res = (bool)$val['value'];
-			}
-		}
-		else { // XML_val
-			if ($val = $this->getInjectedValue($property, $params)) {
-				$hasInjected = true;
-				$obj = $val['obj'];
+		$val = $this->getInjectedValue($property, $params);
+		if ($val) {
+			$obj = $val['obj'];
+			if ($name === 'hasValue') {
+				$res = $obj instanceof Object
+					? $obj->exists()
+					: (bool)$obj;
+			} else {
+				// XML_val
 				$res = $obj->forTemplate();
 			}
-		}
-
-		if ($hasInjected) {
 			$this->resetLocalScope();
 			return $res;
-		}
-		else {
+		} else {
 			return parent::__call($name, $arguments);
 		}
 	}
 }
-
 
 /**
  * Parses a template file with an *.ss file extension.
--- a/tests/model/StringFieldTest.php
+++ b/tests/model/StringFieldTest.php
@@ -36,6 +36,22 @@ class StringFieldTest extends SapphireTest {
 		);
 	}
 
+	public function testExists() {
+		// True exists
+		$this->assertTrue(DBField::create_field('StringFieldTest_MyStringField', true)->exists());
+		$this->assertTrue(DBField::create_field('StringFieldTest_MyStringField', '0')->exists());
+		$this->assertTrue(DBField::create_field('StringFieldTest_MyStringField', '1')->exists());
+		$this->assertTrue(DBField::create_field('StringFieldTest_MyStringField', 1)->exists());
+		$this->assertTrue(DBField::create_field('StringFieldTest_MyStringField', 1.1)->exists());
+
+		// false exists
+		$this->assertFalse(DBField::create_field('StringFieldTest_MyStringField', false)->exists());
+		$this->assertFalse(DBField::create_field('StringFieldTest_MyStringField', '')->exists());
+		$this->assertFalse(DBField::create_field('StringFieldTest_MyStringField', null)->exists());
+		$this->assertFalse(DBField::create_field('StringFieldTest_MyStringField', 0)->exists());
+		$this->assertFalse(DBField::create_field('StringFieldTest_MyStringField', 0.0)->exists());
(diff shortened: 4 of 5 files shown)


sminnee:
I'm comfortable that this is a bugfix and not an API change. Can you clarify which of your test assertions fail without this change, @tractorcow?

sminnee:
Hrm, what happens if an int value 0 is casted to to Text? That probably shouldn't return true for `exists()` either, should it.

The use case of the value being "0" is when someone writes "0", into a free text field. That should be displayed in most cases.

The cases that you describe are when non-string values are passed. false & 0 are the two that come to mind.

Can we think of any time when a StringField field will be set to a numeric value based on user input?

tractorcow:
Hm, that's a good point. For example the below

`<% if $Count %>We have $Count comments<% end_if %>`

Probably should never allow 0 right?

tractorcow:
Updated with some additional checks on numeric values. :)

tractorcow:
Test failures are real; We need to make sure that the correct null/non-null value is written to the DB. I'll fix this shortly.

sminnee:
DBField::testExists() needs to be updated.

tractorcow:
Ok, I've opted to restore the exemption for nullifyIfTrue and this is why.

If it's false (which is NOT the default) this field needs to treat empty strings (not bool false or null) as non-empty. For the sake of this specific issue, this is ok, because the default value for that option suffices.

It's really not worth tampering with the way 3.1 writes strings to the database, so let's leave it until 3.2 to suggest a replacement. Since it's parameterised, it's likely not going to matter so much anyway. :)

model/fieldtypes/StringField.php
```diff
@@ -83,7 +83,9 @@ public function getNullifyEmpty() {
 	 * @see core/model/fieldtypes/DBField#exists()
 	 */
 	public function exists() {
-		return ($this->value || $this->value == '0') || ( !$this->nullifyEmpty && $this->value === '');
+		return $this->value
+			|| (is_string($this->value) && strlen($this->value))
+			|| (!$this->nullifyEmpty && $this->value === '');
```
dhensby: Use the getter method for nullifyEmpty. Also, can there be some docs around what that var is actually intended to do

tractorcow: Yeah I thought about that, but it wasn't used consistently.

I had a look through the code and it looks like it's ment to map '' to nulls internally when saved to the database, but it's not clear what other values are ment to be mapped to (bool)false, true, int, etc....

The main problem I have with the API in this area is that we are taking a database field and using it as a casting helper. In practice a class should be single purpose, and the whole concept of casting for a template should be separate to how it's stored in the database.

I would like to separate these at some point in the future, but for now let's just make them work the best they can without changing the API too drastically.

dhensby: Yes, that would be nicer. But still, can we use the getter function `$this->getNullifyEmpty()` rather than directly accessing the property?

tractorcow: Yeah I'll update.

tractorcow:
Copied from #3084 which has the same issue.

Consider parts of three files:

1. Page.php (Page_Controller)
```php		
public function Truthiness() {
	return true;
}

public function Falsiness() {
	return false;
}
```

2. Layout/Page.ss
```
<% include Truthiness No=$Falsiness, Yes=$Truthiness %>
```
3. Includes/Truthiness.ss
```
<%-- Truthiness.ss --%>
<p>
<% if $No  %>War is Peace, because "$No" is truthy.<% end_if %><br>
<% if $Yes %>Logic still works, because "$Yes" is truthy.<% end_if %>
</p>
```
The output of this is as follows:
```html
<p>
War is Peace, because "" is truthy.<br>
Logic still works, because "1" is truthy.
</p>
```

view/SSViewer.php
```diff
@@ -512,32 +521,25 @@ public function __call($name, $arguments) {
 		if (isset($arguments[1]) && $arguments[1] != null) $params = $arguments[1]; 
 		else $params = array();
 
-		$hasInjected = $res = null;
```
tractorcow: This is the piece of code that was broken from #3084

tractorcow:
Updated based on the feedback and a fix for #3084

tests/templates/SSViewerTestIncludeScopeInheritanceInclude.ss
```diff
@@ -1 +1 @@
-$Title <% if ArgA %>- $ArgA <% end_if %>- <%if First %>First-<% end_if %><% if Last %>Last-<% end_if %><%if MultipleOf(2) %>EVEN<% else %>ODD<% end_if %> top:$Top.Title
+<% if $Title %>$Title<% else %>Untitled<% end_if %> <% if ArgA %>_ $ArgA <% end_if %>- <%if First %>First-<% end_if %><% if Last %>Last-<% end_if %><%if MultipleOf(2) %>EVEN<% else %>ODD<% end_if %> top:$Top.Title
```
tractorcow: I changed the `-` before the `$ArgA` since it was a bit vague.

view/SSViewer.php
```diff
@@ -512,32 +521,25 @@ public function __call($name, $arguments) {
...
-
-		if ($name == 'hasValue') {
-			if ($val = $this->getInjectedValue($property, $params, false)) {
-				$hasInjected = true; $res = (bool)$val['value'];
-			}
-		}
-		else { // XML_val
-			if ($val = $this->getInjectedValue($property, $params)) {
```
tractorcow: The problem is that this line of code didn't properly handle the value already being cast to a DBField, and just did a `!= null` check on it, which of course is always true.

view/SSViewer.php
```diff
@@ -512,32 +521,25 @@ public function __call($name, $arguments) {
 		if (isset($arguments[1]) && $arguments[1] != null) $params = $arguments[1]; 
 		else $params = array();
 
-		$hasInjected = $res = null;
-
-		if ($name == 'hasValue') {
-			if ($val = $this->getInjectedValue($property, $params, false)) {
-				$hasInjected = true; $res = (bool)$val['value'];
```
tractorcow: The problem is that this line of code didn't properly handle the value already being cast to a DBField, and just did a `!= null` check on it, which of course is always true.

tests/templates/SSViewerTestIncludeScopeInheritanceInclude.ss
dhensby: I'm really surprised this doesn't throw themplate parse errors.

`<%if ...` (no space between `%` and `if` always caused problems for me.

Please can you update this template to use `$` prepended vars (eg: `if $First`)

dhensby:
OK, this is looking really good, I can't wait to stop doing `<% if $Bool.exists %>`

Definitely want to get this in asap

view/SSViewer.php
```diff
@@ -512,32 +521,25 @@ public function __call($name, $arguments) {
...
-				$hasInjected = true;
-				$obj = $val['obj'];
-				$res = $obj->forTemplate();
+		$val = $this->getInjectedValue($property, $params);
+		if ($val) {
+			$obj = $val['obj'];
+			if ($name === 'hasValue') {
+				$res = $obj instanceof Object
+					? $obj->exists()
+					: ($obj !== null);
```
dhensby: Will `$obj` ever be `null`, won't it always be an object?

It'll be a casting helper and so shouldn't we check the casting helper's `XML_val()` or similar?

tests/templates/SSViewerTestIncludeScopeInheritanceInclude.ss
tractorcow: Sure, the templates need an update anyway

view/SSViewer.php
tractorcow: I could just change it to 'true' instead of null check

tractorcow: XML_val isn't useful for checking falsiness

dhensby: I think it's confusing to have a comparison that can never be true.

Perhaps just `(bool)$obj` ?

tractorcow: ok :D

tractorcow:
Updated @dhensby

tractorcow:
Wow glad to get this fixed finally. :)

kinglozzer:
Hmm, looks like this has caused a possible regression with the “overlay” logic in `SSViewer` if you do something like `$this->renderWith('Template', $data)` where `$data` is an array:

```php
$data = array('MyBool' => $model->BooleanField); // If BooleanField is "false", this value is a string '0'
echo singleton('ViewableData')->renderWith('MyTemplate', $data);
```

```ss
<%-- MyTemplate.ss --%>
<% if $MyBool %>true<% else %>false<% end_if %>
```

The result will always be “true”. Or a real-world example: [this](https://github.com/kinglozzer/SilverStripe-GMapsObject/blob/a016d5744c5cd5e0a31a110efac9fac8e7a22dc4/code/GMapsObjectExtension.php#L29-L30) is always treated as true.

There are a few workarounds (e.g. wrapping the array in `ArrayData`, or using `$model->dbObject('BooleanField')`), but it’s potentially breaking.

I’m having a brain freeze afternoon and I don’t know if we even *should* fix this (a string of '0' should be shown, but should it evaluate to false in an if statement?), just pointing it out as it’s a behavioural change.

dhensby:
Did this work before or has it always been broken? 

I think when a DO loads a record it should be casting the bools properly to prevent this. 

I guess the real problem is that the dev is passing a strong of 0 to a strictly typed template syntax. Though we shouldn't have a strict typed template language if PHP isn't!

PR #4054, opened 2015-04-01
fixing print functionality if passed arrayData

pitchandtone:


kinglozzer:
Makes sense. Should we apply the same fix to `GridFieldExportButton` too?

dhensby:
> Makes sense. Should we apply the same fix to GridFieldExportButton too?

Why not

PR #4059, opened 2015-04-02
[API] Merge Email send methods & add extension points

danbroooks:
Merged duplicated code in Email.php to use the Send method with argument when sending as plaintext.

Added onBeforeSend & onAfterSend extension points.

If this PR is going to be accepted, point me in the direction as to where documentation needs to be added and I will include it in this request prior to merge.

### DIFF
--- a/email/Email.php
+++ b/email/Email.php
@@ -404,94 +404,32 @@ class Email extends ViewableData {
 		}
 	}
 
-	/**
-	 * Send the email in plaintext.
-	 * 
-	 * @see send() for sending emails with HTML content.
-	 * @uses Mailer->sendPlain()
-	 * 
-	 * @param string $messageID Optional message ID so the message can be identified in bounces etc.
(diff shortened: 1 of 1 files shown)


dhensby:
Good to see email getting some well overdue love. Its been on my radar a while. See #3389

I'd note that this would need to go against 3 branch as its a new api

dhensby:
Also, deprecating send plain may be a good idea

danbroooks:
Is this waiting on action from me now then, sorry?

dhensby:
@danbroooks wow - this has taken a while, sorry.

It needs to be closed and opened against branch 3. If you could deprecate the sendPlain function too, that'd be great.

The other thing worth noting is that in 4 we're hoping to pull mailer out and use SwiftMailer instead, so I wouldn't spend too much effort on this unless you really want to

sminnee:
Since we've replaced the SS4 mailer and there has been no activity for 18 months I'm going close this.

PR #4060, opened 2015-04-02
FIX Files should only report as "existing" if the file actually exists

dhensby:
This fixes an issue where images and files that have models in the DB but no actual file on the filesystem still report as existing.

Fixes this case where there would be no output:

```html
<% if $Image %>
    $image
<% else %>
   <img src="holder.jpg"/>
<% end_if %>
```

### DIFF
--- a/model/Image.php
+++ b/model/Image.php
@@ -123,25 +123,13 @@ class Image extends File implements Flushable {
 	}
 
 	/**
-	 * An image exists if it has a filename.
-	 * Does not do any filesystem checks.
-	 * 
-	 * @return boolean
-	 */
-	public function exists() {
-		if(isset($this->record["Filename"])) {
-			return true;
-		}		
-	}
-	
-	/**
 	 * Return an XHTML img tag for this Image,
 	 * or NULL if the image file doesn't exist on the filesystem.
 	 * 
 	 * @return string
 	 */
 	public function getTag() {
-		if(file_exists(Director::baseFolder() . '/' . $this->Filename)) {
+		if($this->exists()) {
 			$url = $this->getURL();
 			$title = ($this->Title) ? $this->Title : $this->Filename;
 			if($this->Title) {
@@ -618,8 +606,8 @@ class Image extends File implements Flushable {
 	public function getDimensions($dim = "string") {
 		if($this->getField('Filename')) {
 
-			$imagefile = Director::baseFolder() . '/' . $this->getField('Filename');
-			if(file_exists($imagefile)) {
+			$imagefile = $this->getFullPath();
+			if($this->exists()) {
 				$size = getimagesize($imagefile);
 				return ($dim === "string") ? "$size[0]x$size[1]" : $size[$dim];
 			} else {
--- a/filesystem/File.php
+++ b/filesystem/File.php
@@ -222,8 +222,17 @@ class File extends DataObject {
 	}
 
 	/**
-	 * Find a File object by the given filename.
+	 * A file only exists if the file_exists() and is in the DB as a record
 	 * 
+	 * @return bool
+	 */
+	public function exists() {
+		return parent::exists() && file_exists($this->getFullPath());
+	}
+
+	/**
+	 * Find a File object by the given filename.
+	 *
 	 * @param String $filename Matched against the "Name" property.
 	 * @return mixed null if not found, File object of found file
 	 */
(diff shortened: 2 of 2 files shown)


willmorgan:
This should go in to 3, I think - I think I've written code in the past along the lines of:

```php
if($file->exists() && !file_exists($file->Filename))
```

dhensby:
My argument to that is: The `exists()` method returning true in that instance is a bug and you've written code to mitigate that bug.

We'd basically never release a patch ever if our argument to bug fixes was "but someone could have mitigated it so it's a breaking change"

willmorgan:
I'm more concerned about cases like this, that will be broken:

```php
if($file->exists()) {
    if(!file_exists($file->Filename)) {
        // some remedial logic here
    }
    // the main branch here
}
```

I have no problem with the nature of the PR, just the branch it's currently submitted against.

model/Image.php
```diff
@@ -123,25 +123,13 @@ public function getCMSFields() {
 	}
 
 	/**
-	 * An image exists if it has a filename.
-	 * Does not do any filesystem checks.
```
tractorcow: This to me does feel like this is a change in behaviour... I think it's worth doing a bit of digging into this method to see where it was introduced, and why it explicitly doesn't do a filesystem check. If it turns out that use case is no longer valid then maybe it would be ok to merge this into 3.1, but maybe even then it'd be safer to target the 3 branch.

tractorcow: Well, it looks like this function is unchanged since being imported for the first time into open source.

https://github.com/dhensby/silverstripe-framework/commit/4a5d9b03f82d383cef40602aed7353dea17c31b4

The documentation about this strange behaviour seems to have been added since then, rather than at the same time.

tractorcow:
I think the argument is quite easy to argue either way, the method behaves unexpectedly in templates, but the odd behaviour has been at least documented as such. Given that the specific behaviour is documented this way, I guess my vote is to still merge this into 3.2 instead of 3.1.

dhensby:
The question is, is this a bug or expected behaviour?

If it's a bug, it can go in 3.1 - if it's not, then it's a breaking change and goes in master.

The fact that `Image` acts differently to `File` and that neither take into account whether the entity being represented by the object actually exists is, to me, a bug.

There is also no API to check that a file actually exists, so the bug in the template is completely unresolvable without custom code to mitigate this bug.

If the consensus is that this is an intended and expected feature, then I'll put it against 4.x-dev.

willmorgan:
> If it's a bug, it can go in 3.1 - if it's not, then it's a breaking change and goes in master.
> If the consensus is that this is an intended and expected feature, then I'll put it against 4.x-dev.

I'd say it's a bug, but it's not fixed in a BC way - and it's pretty impossible to change that now.

It can go in to 3 - doesn't need to be 4.

tractorcow:
I would argue it's an intended feature that doesn't work very well, and I'd consider it an api enhancement to include the filesystem check. I think that semver requires that it should go into 3.

I can't think of a case where it would break something that wasn't already working properly. :D An empty `` object inside an `<% if $Image %>` would now not be rendered right?

sminnee:
I think that this is pretty clearly a bug, not a feature change, but given the discussion, I'd class this as medium regression risk and commit to 3.2 or master.

model/Image.php
dhensby: It's an interesting case where a function has been documented with no care as to whether it's really right.

It's just "this code does something".

```php

/**
 * This function multiples the two arguments
 */
function divide($numerator, $denominator) {
return $numerator * $denominator;
}
```

Now, if that doc block was just added later to state what the function does, as opposed to document it's intended purpose, then we can say it's a bug, otherwise we'd say it's stupid naming but acts as intended.

The API for "exists" has already been set and shouldn't really change arbitrarally because old code changed it's behaviour; just because someone documented it incorrectly doesn't mean it's expected behaviour

sminnee:
The docs of exists() say

> Returns true if this object "exists", i.e., has a sensible value.

I don't think a File record with a missing file counts as a sensible value.

This is a bug.

dhensby:
> I'd say it's a bug, but it's not fixed in a BC way - and it's pretty impossible to change that now.

Fixing bugs in a non-BC way is impossible by definition. You're changing behaviour because it's wrong and that will break things that relied on that bug.

> I can't think of a case where it would break something that wasn't already working properly. :D An empty  object inside an <% if $Image %> would now not be rendered right?

that's right.

I think that given 3 core-committers have said this should go into 3.x-dev then I'll put it there (even though I think 3.1 deserves this fix!) :P

PR #4061, opened 2015-04-02
removed unnecessary write() from Member::logIn()

christopherdarling:
$this->write() is already called in regenerateTempID()


dhensby:
Hmm, I think the implied write in `regenerateTempID` isn't a good enough reason to change this as it makes the function more vulnerable to regression.

a `write` on an object that hasn't been modified won't touch the DB so this shouldn't be a *big* performance problem but makes the function more resilient

christopherdarling:
fair enough - good point, feel free to close

PR #4069, opened 2015-04-13
DOCS Make a note to ensure people install Git and Composer

camfindlay:
Since Composer uses git and this isn't explicitly documented anywhere at our end I have added a note and link for newcomers to follow and get Git before getting Composer.

### DIFF
--- a/docs/en/00_Getting_Started/02_Composer.md
+++ b/docs/en/00_Getting_Started/02_Composer.md
@@ -9,7 +9,9 @@ We also have separate instructions for [installing modules with Composer](/devel
 
 ## Installing composer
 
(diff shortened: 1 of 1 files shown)


willmorgan:
> Installing on Mac
There are several ways to install Git on a Mac. The easiest is probably to install the Xcode Command Line Tools.

Hmm, are you sure about that? :wink:

stevie-mayhew:
You need xcode to install brew anyway?

chillu:
I don't think we need to explain how to install git itself, there's enough instructions out there for this purpose. The git-scm book is very common, so I trust they keep their install methods current and relevant. Cam's addition doesn't try to do that, so I'm fine to merge as-is.

PR #4070, opened 2015-04-13
Move iframe load handler to separate method

jkRoberts:
When using the DMS module linking to an existing document the edit iframe would fail to remove the loading icon. This change makes it possible for DMS javascript to manually call the iframe load logic.

### DIFF
--- a/javascript/UploadField.js
+++ b/javascript/UploadField.js
@@ -416,7 +416,7 @@
 				return false;
 			} 
 		});
-		$( 'div.ss-upload:not(.disabled):not(.readonly) .ss-uploadfield-item-edit').entwine({
+		$( 'div.ss-upload:not(.disabled):not(.readonly) .ss-uploadfield-item-edit, .ss-assetuploadfield .ss-uploadfield-item-edit').entwine({
 			onclick: function(e) {
 				var self = this,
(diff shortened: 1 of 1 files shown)


dhensby:
I don't see any problems with taking this in. Anyone else on the core team got an opinion?

flamerohr:
How come some of the selectors are being changed?

If it were just splitting out loadIframe functionality, I think this is all good.

javascript/UploadField.js
```diff
@@ -416,7 +416,7 @@
 				return false;
 			} 
 		});
-		$( 'div.ss-upload:not(.disabled):not(.readonly) .ss-uploadfield-item-edit').entwine({
+		$( 'div.ss-upload:not(.disabled):not(.readonly) .ss-uploadfield-item-edit, .ss-assetuploadfield:not(.disabled):not(.readonly) .ss-uploadfield-item-edit').entwine({
```
sminnee: I don't think this should be restricted only to assetuploadfield.

dhensby:
@jkRoberts any plans to address the feedback on this?

maxime-rainville:

maxime-rainville:
Asset management has been completely redone in SilverStripe 4. So there wouldn't be any point rebasing this PR.

PR #4072, opened 2015-04-14
BUG Fix accordion sometimes displaying scrollbars

tractorcow:
To replicate this issue create a HTMLEditorField within a ToggleCompositeField and set the rows to something smaller than 15.

### DIFF
--- a/javascript/ToggleCompositeField.js
+++ b/javascript/ToggleCompositeField.js
@@ -5,6 +5,7 @@
 				this._super();
 
 				this.accordion({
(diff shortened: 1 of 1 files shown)


tractorcow:
Tested on IE 11, firefox, chrome

assertchris:
WHATNOIE7?!

chillu:
Discussed offline with Damian, just checking that there's no CSS heights to document or remove, and that he's taking auto-sizing TinyMCE into account - which he does.

PR #4074, opened 2015-04-15
Improve SQL Format information paragraph

rotassator:
Improve SQL Format information paragraph in the SilverStripe Coding Conventions. It just looked like some missing punctuation was missed when written.


tractorcow:
Thanks @rotassator :D

PR #4079, opened 2015-04-17
Add row and cell generator methods

assertchris:
The reason for this change is to allow row-level and cell-level changes in subclasses.

Also: https://github.com/silverstripe/silverstripe-framework/pull/4079/files#diff-c804f1dc153eaf30afb92594fa774456L362

### DIFF
--- a/forms/gridfield/GridField.php
+++ b/forms/gridfield/GridField.php
@@ -1,25 +1,26 @@
 <?php
+
 /**
  * Displays a {@link SS_List} in a grid format.
- * 
- * GridField is a field that takes an SS_List and displays it in an table with rows 
- * and columns. It reminds of the old TableFields but works with SS_List types 
+ *
+ * GridField is a field that takes an SS_List and displays it in an table with rows
+ * and columns. It reminds of the old TableFields but works with SS_List types
  * and only loads the necessary rows from the list.
- * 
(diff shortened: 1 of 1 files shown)


forms/gridfield/GridField.php
```diff
@@ -83,9 +87,9 @@ class GridField extends FormField {
 	/**
 	 * Creates a new GridField field
 	 *
-	 * @param string $name
-	 * @param string $title
-	 * @param SS_List $dataList
+	 * @param string          $name
```
tractorcow: Can we just use simple spacing for phpdoc? Most of the codebase doesn't follow aligned indentation.

tractorcow:
I guess this doesn't really add much rather than refactor existing code. Do you think this is minor enough to fit into 3.1? cc @silverstripe/core-team ?

tractorcow:
I would say "style sucks, so minor code cleanup" is safe to merge into a patch release.

willmorgan:
I think that it's safe to go in, though I've gotta say all the reformatting makes it hard to review :wink:

Thanks @assertchris!

dhensby:
@willmorgan I'm only 4 months late but you can add `?w=1` to the end of the url to see it without whitespace changes!

PR #3712, opened 2014-12-05
Update 'spinner' to a SilverStripe branded version

nickspiel:
Updated PR based on feedback form #3678

PSD: http://nickspiel.me/uploads/silverstripe-assets/spinner.psd

PR #4080, opened 2015-04-17
Adding loading spinner to ui-autocomplete fields when making ajax reqests

dhensby:
This change now makes it apparent to the use when "work" is still being done and we are waiting for a response.

A nice bit of UI feedback.

loader

### DIFF
--- a/admin/css/screen.css
+++ b/admin/css/screen.css
@@ -140,6 +140,7 @@ body, html { font-size: 12px; line-height: 16px; font-family: Arial, sans-serif;
 .ui-accordion .ui-accordion-content { border: 1px solid #c0c0c2; border-top: none; }
 
 .ui-autocomplete { max-height: 240px; overflow-x: hidden; overflow-y: auto; }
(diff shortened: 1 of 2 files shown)


jonom:
Nice one! I wonder if it should use a small version of that [new SS spinner](https://github.com/silverstripe/silverstripe-framework/pull/3712) though? spinner

willmorgan:
:+1: looks good to me!

I think that we should at least see how it looks with the new loading spinner. But I get the feeling that some convention needs to be established about when we use it, because it might not be for everything.

chillu:
I think the SS spinner is too distracting here. I can understand its merits in the main application loader screen, but that doesn't mean we need to use it everywhere. Particularly in these small dimensions it actually looks like a textual character thats behaving weirdly - people are quite used to a circular loading indicator in all kinds of web apps, my vote goes to keeping it that way here.

dhensby:
I don't think having the SS branded loader is right here; I'm not inclined to do this in HTML when someone can just photoshop it to see what it's like.

Spinner is much more natural here.

willmorgan:
I think this is good to merge now. Personally I think that the SS spinner doesn't really lend itself too well here and that the "standard" loading spinner is much more recognisable.

jonom:
:+1:

PR #4083, opened 2015-04-20
Abstracting GridField row class generation

assertchris:
This further abstracts the classes generated for each row of the GridField. Without it, subclasses need to implement the same first/last/odd/even/base classes, and they also become responsible for the special `data-*` parameters GridField needs.

### DIFF
--- a/forms/gridfield/GridField.php
+++ b/forms/gridfield/GridField.php
@@ -349,7 +349,9 @@ class GridField extends FormField {
 				if($record->hasMethod('canView') && !$record->canView()) {
 					continue;
 				}
+
 				$rowContent = '';
+
 				foreach($this->getColumns() as $column) {
 					$colContent = $this->getColumnContent($record, $column);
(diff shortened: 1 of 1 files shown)


willmorgan:
Hmm, this is now a stealthy API change because of #4079 :wink:

I think we need to be more careful in the future - we've introduced 2 new methods and then changed their argument signature fairly quickly, and taken them in to 3.1.

Does it matter so much that they're new and protected?

@tractorcow, thoughts?

assertchris:
@willmorgan "Does it matter so much that they're new and protected?" and there has been no framework release featuring the changes... :)

tractorcow:
I agree with @willmorgan in that I think it's more a sign we should PR minor changes more carefully, to prevent us having to patch them up like this in the future. ;D I'm as guilty of that as any.

We can probably excuse this one as a replacement PR for the prior.

PR #4085, opened 2015-04-20
BUG Fix users with all cms section access not able to edit files

### LINKED ISSUE #4078: BUG Cannot upload files with CMS_ACCESS_LeftAndMain
Since 3.1.9, actually since https://github.com/silverstripe/silverstripe-framework/pull/3781, I think, the `CMS_ACCESS_LeftAndMain` permission does not work as intended. At least for the assets.

Now, by default, a user can only create a `File`, if he is an `ADMIN` or has the `CMS_ACCESS_AssetAdmin` permission. But if you use the `CMS_ACCESS_LeftAndMain` permission, which until then stood for "allow access to all admins", the user is not able to create files. But if you use this checkbox, the other admin checkboxes are disabeld. So for now there is only one way: you have to unset the `CMS_ACCESS_LeftAndMain` checkbox and set those admin checkboxes (eg. `CMS_ACCESS_AssetAdmin`) one by one.

tractorcow:
Fixes #4078 

If a user has "all cms section" access they should be able to edit files.

### DIFF
--- a/tests/filesystem/FileTest.yml
+++ b/tests/filesystem/FileTest.yml
@@ -35,6 +35,8 @@ Permission:
     Code: CMS_ACCESS_LeftAndMain
   assetadmin:
     Code: CMS_ACCESS_AssetAdmin
+  securityadmin:
+    Code: CMS_ACCESS_SecurityAdmin
 Group:
   admins:
     Title: Administrators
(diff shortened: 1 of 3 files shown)


tests/filesystem/FileTest.yml
```diff
@@ -42,9 +44,12 @@ Group:
   cmsusers:
     Title: 'CMS Users'
     Permissions: =>Permission.cmsmain
+  securityusers:
+    Title: 'Security Users'
+    Permissions: =>Permission.securityadmin
   assetusers:
     Title: 'Asset Users'
-    Permissions: =>Permission.cmsmain, =>Permission.assetadmin
```
tractorcow: This was a bug; Relation between groups and permissions is has_many not many_many, and two groups were attempting to link to the same permission.

tractorcow:
Thanks @chillu

PR #4086, opened 2015-04-20
fix width of remote url field label to allow right-click-paste

xini:
when you right-click into the left half of the field the 'paste' option doesn't show because the label is too wide.


willmorgan:
Hey @xini, you'll need to make the changes to the SCSS file too, in order for this PR to be merged in.

I believe the source file is `admin/scss/_uitheme.scss`

xini:
Hi @willmorgan, done. sorry for that.

dhensby:
@xini sorry, I'm not completely clear as to what part of the CMS this is affecting specifically.

Can you provide a bit more detail in regards to replicating (screen shot would be good)?

xini:
@dhensby, here the situation:
You want to insert a video into the content of a page. You go to YouTube and select and copy a URL.
You open the add media panel and select 'from the web'. To paste the URL you right-click on the field to get the 'paste' option (you could of course use ctrl+v, but my customer doesn't know that). 
When you right-click on the left half of the field, the paste option is not offered. 
screen-left

When you right-click on the right half of the field the 'paste' option is offered.
screen-right

When you inspect the field you'll see that the label covers the field on the left half, hence when right-clicking on the label the 'paste' option is not shown.
inspect

dhensby:
thanks, sorry it took so long

Issue #4089, opened 2015-04-21
Unable to use CURLOPT_* constants in RestfulService YML configuration.


patricknelson:
Right now you cannot specify the standard curl configuration options (e.g. `CURLOPT_CONNECTTIMEOUT`) in the YML configuration. Instead, if you need to change the connect timeout, you'd have to provide the following configuration with the actual value of the constant itself, e.g.:

```yaml
RestfulService:
  default_curl_options:
    78: 10
```

I'd like to be able to do the following instead:

```yaml
RestfulService:
  default_curl_options:
    CURLOPT_CONNECTTIMEOUT: 10
```

I can submit a pull request demonstrating the code that I've written to address this.

dhensby:
Closing because we're dropping support for restfulservice in 4.x and it's not worth the effort to maintain it

PR #4090, opened 2015-04-21
FIX for #4089, facilitating CURLOPT_* constants in RestfulService YML config.

patricknelson:
PR for #4089

### DIFF
--- a/api/RestfulService.php
+++ b/api/RestfulService.php
@@ -53,7 +53,7 @@ class RestfulService extends ViewableData implements Flushable {
 	 */
 	public static function set_default_curl_option($option, $value) {
 		Deprecation::notice('3.2', 'Use the "RestfulService.default_curl_options" config setting instead');
-		Config::inst()->update('RestfulService', 'default_curl_options', array($option => $value));
+		Config::inst()->update(get_called_class(), 'default_curl_options', array($option => $value));
 	}
 
 	/**
@@ -63,7 +63,25 @@ class RestfulService extends ViewableData implements Flushable {
 	 */
 	public static function set_default_curl_options($optionArray) {
 		Deprecation::notice('3.2', 'Use the "RestfulService.default_curl_options" config setting instead');
-		Config::inst()->update('RestfulService', 'default_curl_options', $optionArray);
+		Config::inst()->update(get_called_class(), 'default_curl_options', $optionArray);
+	}
+
+	/**
+	 * Parses cURL options defined in YML configuration.
+	 *
+	 * @return array
+	 */
+	public static function get_default_curl_options() {
+		$rawOptions = static::config()->get('default_curl_options');
+		$options = array();
+		foreach($rawOptions as $key => $value) {
+			// See if this key looks like a constant and, if so, try to get its value and use *that* as the key instead.
+			if (!is_numeric($key) && substr($key, 0, 8) === "CURLOPT_" && defined($key)) {
+				$key = constant($key);
+			}
+			$options[$key] = $value;
+		}
+		return $options;
 	}
 
 	/**
@@ -83,7 +101,7 @@ class RestfulService extends ViewableData implements Flushable {
 			'Use the "RestfulService.default_curl_options" config setting instead, '
 				. 'with direct reference to the CURL_* options'
 		);
-		config::inst()->update('RestfulService', 'default_proxy', array(
+		Config::inst()->update(get_called_class(), 'default_proxy', array(
 			CURLOPT_PROXY => $proxy,
 			CURLOPT_PROXYUSERPWD => "{$user}:{$password}",
 			CURLOPT_PROXYPORT => $port,
@@ -175,7 +193,7 @@ class RestfulService extends ViewableData implements Flushable {
 			$method,
 			$data,
 			array_merge((array)$this->customHeaders, (array)$headers),
-			$curlOptions + (array)$this->config()->default_curl_options,
+			$curlOptions + (array)static::get_default_curl_options(),
 			$this->getBasicAuthString()
 		));
 
@@ -231,7 +249,7 @@ class RestfulService extends ViewableData implements Flushable {
 		$timeout   = 5;
 		$sapphireInfo = new SapphireInfo();
 		$useragent = 'SilverStripe/' . $sapphireInfo->Version();
-		$curlOptions = $curlOptions + (array)$this->config()->default_curl_options;
+		$curlOptions = $curlOptions + (array)static::get_default_curl_options();
 
 		curl_setopt($ch, CURLOPT_URL, $url);
 		curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
@@ -350,6 +368,20 @@ class RestfulService extends ViewableData implements Flushable {
 		$lastHeaders = array_pop($parts);
 		$headers = $this->parseRawHeaders($lastHeaders);
 		//return the response object
+		return $this->createResponse($rawBody, $statusCode, $headers);
+	}
+
+	/**
+	 * Instantiates the RestfulService_Response object -- however, this is only setup to allow easily overriding this
+	 * functionality to generate a different response object (which extends either RestfulService_Response or
+	 * SS_HTTPResponse) by overriding this class as well as this method.
+	 *
+	 * @param	string	$rawBody
+	 * @param	int		$statusCode
+	 * @param	array	$headers
+	 * @return	RestfulService_Response
+	 */
+	protected function createResponse($rawBody, $statusCode, $headers) {
 		return new RestfulService_Response($rawBody, $statusCode, $headers);
 	}
 
@@ -620,7 +652,7 @@ class RestfulService_Response extends SS_HTTPResponse {
 	public function setCachedBody($content) {
 		Deprecation::notice('3.2', 'Setting the response body is now deprecated, set the cached request instead');
 		if (!$this->cachedResponse) {
-			$this->cachedResponse = new RestfulService_Response($content);
+			$this->cachedResponse = new static($content);
 		}
 		else {
 			$this->cachedResponse->setBody($content);


api/RestfulService.php
```diff
@@ -67,6 +67,22 @@ public static function set_default_curl_options($optionArray) {
...
+	 * Parses cURL options defined in YML configuration.
+	 *
+	 * @return array
+	 */
+	public static function get_default_curl_options() {
+		$rawOptions = Config::inst()->get('RestfulService', 'default_curl_options');
+		$options = array();
+		foreach($rawOptions as $key => $value) {
+			// See if this key looks like a constant and, if so, try to get its value and use *that* as the key instead.
+			if (!is_numeric($key) && substr($key, 0, 8) === "CURLOPT_" && defined($key)) $key = constant($key);
```
willmorgan: Could we get this with braces, please?

api/RestfulService.php
```diff
@@ -175,7 +191,7 @@ public function request($subURL = '', $method = "GET", $data = null, $headers =
 			$method,
 			$data,
 			array_merge((array)$this->customHeaders, (array)$headers),
-			$curlOptions + (array)$this->config()->default_curl_options,
+			$curlOptions + (array)$this->get_default_curl_options(),
```
willmorgan: You're calling a static function dynamically here; don't you mean `static::get_default_curl_options()`?

api/RestfulService.php
```diff
@@ -67,6 +67,22 @@ public static function set_default_curl_options($optionArray) {
 	}
 
 	/**
+	 * Parses cURL options defined in YML configuration.
+	 *
+	 * @return array
+	 */
+	public static function get_default_curl_options() {
+		$rawOptions = Config::inst()->get('RestfulService', 'default_curl_options');
```
willmorgan: Should use `get_called_class()` here instead of hardcoding in RestfulService, IMHO. Allows you to extend easier without needing to convert this in to an instance method and lookup using `$this->config()`.

api/RestfulService.php
patricknelson: Nope, I didn't miss that per se. You cannot call an instance method statically but you can certainly call static methods from an instance context, which uses late static binding akin to `static` keyword, by default and it's easy :D It may be a code smell (I.e. Unexpected) but it's based on a solid principle at least. That's all from memory but I'm pretty certain it's fine, unless other parts of the file are calling static methodson itself differently (and I've frequently seen `self` in SS unfortunately).

api/RestfulService.php
patricknelson: On iPad at moment, but if I recall, the other methods nearby use this same method. It may be more significant of a change but I'd be happy to do that in the other methods as well (so long as everyone is ok with that update and the impact it would have, if any, on other branches this would merge into).  If this doesn't become a part of `3` (I suspect it does) we would expect any other additional methods to be tweaked as well, I suppose 

Is this common elsewhere in the code base? I canot check at the moment.

api/RestfulService.php
patricknelson: Easy to change though.

willmorgan: > Easy to change though.

Would you mind?

Static methods should be called statically; dynamic methods dynamically. There's no benefit to invoking it in a different context here.

Yes, we should be using `static` instead of `self`, in general, but I can't see how that principle applies here.

api/RestfulService.php
willmorgan: Pretty sure it's a good precedent to set... paging @silverstripe/core-team

patricknelson: I will try that out here and then elsewhere in the same object utilizing `Config` just to rustle jimmies and see what happens :)

patricknelson:
@willmorgan and @silverstripe/core-team

By the way, while I'm in here (and having had my own issues but unable to decipher what was going on  which ultimately lead me to **this** optimization), I noticed that there is a lot of opportunity to improve error reporting here by using (*cringe*) ... exceptions. I'd be willing to capitulate to using `user_error()` or `trigger_error()` since that is currently the standard set in place (even though really we should be using Exceptions) but, if you look at [this code as it is right now](https://github.com/patricknelson/silverstripe-framework/blob/2bdae810e24b07b62b2b0ba0bc080385416ccc4b/api/RestfulService.php#L361), the `$curlError` variable is just sitting there and not being used at all. In fact, my timeout error is a local error and therefore doesn't result in an actual HTTP code (since HTTP wasn't even possible yet due to a DNS error on my end, lower in the stack) so I had no insight into what was happening without manually weeding right back into the SS core (which I've been finding myself doing quite often, unfortunately since I'm pedantic about getting fairly precise/expected results).

What do you guys think? 

On another note, I'll be updating my commit soon.

api/RestfulService.php
patricknelson: Done, see: https://github.com/patricknelson/silverstripe-framework/commit/cb2c6b98241286bae6743928fc1d2f359e162f6a#diff-e0ab4000778eb49720f320171895f8d3R79

patricknelson: See https://github.com/patricknelson/silverstripe-framework/commit/d9b76296c9927a92aeb0e35f3979eb9028875766#diff-e0ab4000778eb49720f320171895f8d3R79

patricknelson:
I want to actually reply to the line comments here since I've had to rebase so the commits are out of date.

* Setup braces instead of all one line for `if` statement. https://github.com/patricknelson/silverstripe-framework/commit/d9b76296c9927a92aeb0e35f3979eb9028875766#diff-e0ab4000778eb49720f320171895f8d3R79
* Using `static` keyword now instead of `$this->` on static methods: https://github.com/patricknelson/silverstripe-framework/commit/d9b76296c9927a92aeb0e35f3979eb9028875766#diff-e0ab4000778eb49720f320171895f8d3R196
* Using `get_called_class()` here instead of the hard-coded `RestfulService` class name. This should still work since the `Config` abstraction iterates through parents and extensions until it finds a match: https://github.com/patricknelson/silverstripe-framework/commit/d9b76296c9927a92aeb0e35f3979eb9028875766#diff-e0ab4000778eb49720f320171895f8d3R104

Also, new:

* Based on your suggestion to use `get_called_class()` here, realized that overriding `ResfulService` may also imply the need to potentially override `ResfulService_Response` (or both) and the response object wouldn't ever be used unless you had a way to ensure it created an instance of your object instead of the original, therefore, I abstracted the instantiation of the `ResfulService_Response` object to another method (in lieu of a real IoC container as far as I'm aware) which should allow you to ensure that the desired objects would be used without having to edit to much (or any) of the core code: https://github.com/patricknelson/silverstripe-framework/commit/d9b76296c9927a92aeb0e35f3979eb9028875766#diff-e0ab4000778eb49720f320171895f8d3R371 -- That also meant using the `static` keyword here as well (just below):  https://github.com/patricknelson/silverstripe-framework/commit/d9b76296c9927a92aeb0e35f3979eb9028875766#diff-e0ab4000778eb49720f320171895f8d3R655

api/RestfulService.php
tractorcow: I personally prefer `static::config()->default_curl_options` but you can use get_called_class if you like.

patricknelson: Actually I like too. Simpler to read and right as well as providing the same exact benefits and separation of concerns. 


- Patrick Nelson

willmorgan: Sorry, yeah, that would be great.

api/RestfulService.php
```diff
@@ -83,7 +101,7 @@ public static function set_default_proxy($proxy, $port = 80, $user = "", $passwo
 			'Use the "RestfulService.default_curl_options" config setting instead, '
 				. 'with direct reference to the CURL_* options'
 		);
-		config::inst()->update('RestfulService', 'default_proxy', array(
+		config::inst()->update(get_called_class(), 'default_proxy', array(
```
willmorgan: This should really be static::config() too - just noticed that config is lowercase.

patricknelson: Updated: https://github.com/patricknelson/silverstripe-framework/commit/5e8d166b48ba7037f609e822ccbe2554834cf804#diff-e0ab4000778eb49720f320171895f8d3R104

api/RestfulService.php
```diff
@@ -53,7 +53,7 @@ public static function flush() {
 	 */
 	public static function set_default_curl_option($option, $value) {
 		Deprecation::notice('3.2', 'Use the "RestfulService.default_curl_options" config setting instead');
-		Config::inst()->update('RestfulService', 'default_curl_options', array($option => $value));
+		Config::inst()->update(get_called_class(), 'default_curl_options', array($option => $value));
```
willmorgan: This (and other occurrences) can be replaced with `static::config()->update(...` too. Sorry to mess you around.

api/RestfulService.php
patricknelson: All set on that: https://github.com/patricknelson/silverstripe-framework/commit/5e8d166b48ba7037f609e822ccbe2554834cf804#diff-e0ab4000778eb49720f320171895f8d3R75

Also note, in the interest of extensibility, the *other* references which must be calling `Config::inst()->update()` directly (since there is no ability to call `->update()` via `static::config()`) were also updated to use `get_called_class()` here: https://github.com/patricknelson/silverstripe-framework/commit/5e8d166b48ba7037f609e822ccbe2554834cf804#diff-e0ab4000778eb49720f320171895f8d3R56

In the same vogue (extensibility), I abstracted the instantiation of the `RestfulService_Response` class to a method since SS unfortunately doesn't use IoC (as far as I know) here, in case the developer does ever extend the class (i.e. the point to the above changes): https://github.com/patricknelson/silverstripe-framework/commit/5e8d166b48ba7037f609e822ccbe2554834cf804#diff-e0ab4000778eb49720f320171895f8d3R371 -- that also impacted this line as well: https://github.com/patricknelson/silverstripe-framework/commit/5e8d166b48ba7037f609e822ccbe2554834cf804#diff-e0ab4000778eb49720f320171895f8d3R655

api/RestfulService.php
patricknelson: That's ok -- I just commented on that here. I don't think that's possible. Looking in my IDE, at least, I don't see an `->update()` method on the `Config_ForClass` class which is returned by the `::config()` container on `Object`.

patricknelson: I agree though that this API should be more consistent ;) Here's the class/file I'm looking at which is returned by `static::config()` calls: https://github.com/silverstripe/silverstripe-framework/blob/3/core/Config.php#L859

patricknelson: My comment pointing this out: https://github.com/silverstripe/silverstripe-framework/pull/4090#discussion_r28975400

dhensby:
It's a shame this hasn't seen it through - @patricknelson are you able to rebase and we can look at getting this going?

I understand that config_forclass doesn't have an `update` method (crappy)

patricknelson:
Man this got forgotten ... added to my "TODO" as well :dizzy_face:

dhensby:
@patricknelson RestfulService is so deeply flawed that I don't think we should spend any time on it. It's getting pulled from core in 4 in favour of guzzle, so I'll close this to save you any more pain. Sorry

patricknelson:
Not a problem at all! This is a very good thing, I love the Guzzle library.

PR #4091, opened 2015-04-21
FIX: Object::parse_class_spec failed to parse associative arrays

kinglozzer:
Submitted to 3.1 as this enables a documented ability of the framework - [configuring database field options](https://github.com/silverstripe/silverstripe-framework/blob/d5361dedecee991e37b6563918de0aa32c6d0f21/model/fieldtypes/StringField.php#L31). I.e:

```php
private static $db = array(
	'MyField' => 'Varchar(255, array("nullifyEmpty" => false))'
);
```

### DIFF
--- a/core/Object.php
+++ b/core/Object.php
@@ -201,6 +201,8 @@ abstract class Object {
 		$bucket = &$args;
 		$bucketStack = array();
 		$had_ns = false;
+		$isAssociativeArray = false;
+		$lastArrayKey = null;
 		
 		foreach($tokens as $token) {
 			$tName = is_array($token) ? $token[0] : $token;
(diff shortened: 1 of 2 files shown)


kinglozzer:
Screwed this up by force pushing and overwriting my own work with an outdated version... will re-open once I’ve recovered it. :poop:

kinglozzer:
Anyway, I intended to submit a (slightly naughty) alternative proposal: [use `eval()`](https://github.com/kinglozzer/sapphire/commit/af18c0b99f9b408fd8e42672b92878603fb9312a) (thanks  to @stnvh for the regex in that). Obviously `eval()` is evil, and regular expressions can be a pain to maintain, but given the amount of confusing code we could shed I thought it was at least worth proposing. Using `eval()` would also make this much, much more flexible, e.g:

```php
private static $db = array(
    'MyField' => 'CustomField(new CustomFieldDependency())',
    'AnotherField' => 'ClassName(Dependency::SOME_CONST)'
);
```

I’m sure dependency injection like this isn’t the purpose of this code, but you get the idea.

kinglozzer:
Reopened as #4162

Issue #4092, opened 2015-04-21
Issue with multiple parent has_many relations to the same child object.


patricknelson:
Apparently, `DataObject->getRemoteJoinField()` has a flaw in it preventing you from having two separate relations from a parent to the same child object. It appears this may only affect the `3.1` branch, since reviewing the newer code in the `3` branch, I think this bug may have been addressed.

**For example**, say you have a website with contests and you can have multiple grand prize winners but also have multiple runners up. Also, say you have these contests monthly and the same players will play. You can boil this down to the following:

```php
<?php
class Contest extends DataObject {
	private static $has_many = [
		"GrandPrize" => "Person",
		"RunnerUp" => "Person",
	];
}

class Person extends DataObject {
	private static $has_one = [
		"GrandPrize" => "Contest", // This key will match up with the "component" name (the key) on the parent object. This also becomes "GrandPrizeID" to reference back to the parent.
		"RunnerUp" => "Contest",
	];
}
```

It turns out that in `->getRemoteJoinField()`, which is responsible for automatically determining the foreign key (e.g. "GrandPrizeID" pointing back to the `Contest` object) uses `array_flip()` when it gets the names of the unique relations (on the child object, `Person`) pointing back to the parent object (`Contest`) and, in doing so, clobbers the first relation (`GrandPrize`) which means that when you attempt to get the relation for `->GrandPrizeWinners()` it will return a `HasManyList` for `RunnerUp` instead.

**Right now**, it doesn't matter what the relation names are on the parent/child objects. However, due to the inability to determine which relation to use to reference back properly, e.g. that `GrandPrizeID` should be used for the `->GrandPrizeWinners()` relation, it **will** matter what the relation name is, if you happen to have an edge case like mine where you need multiple `has_many` relations from the same parent to the same child (without using `many_many`, which admittedly may be a workaround for this).

My pull request #4093 will prefer the `has_one` relation pointing back that has a matching name in the parent `has_many` and, if no matching name is found, will continue with the existing functionality.

patricknelson:
Please let me know how I should proceed with a pull request on this -- which branch should I base this on? Since it's a bug fix, should this go into `3` branch?

patricknelson:
I'm sorry -- I will correct myself in my code above...

[...]

PR #4093, opened 2015-04-21
FIX for #4092, allowing multiple parent has_many relations to the same child object

patricknelson:
Addresses #4092 which affects `3.1` only.

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -1546,9 +1546,13 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 		if(!is_array($remoteRelations)) {
 			$remoteRelations = array();
 		}
+
+		// See if we happen to already have an exact match pointing back to this object (or a parent object) and, if so, use that as our foreign key.
+		$ancestry = array_reverse(ClassInfo::ancestry($this));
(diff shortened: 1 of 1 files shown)


kinglozzer:
If #4092 can’t be fixed with dot-notation, we’ll need a unit test that fails before this change to cover this issue :)

patricknelson:
It can be fixed with dot-notation. Anyone else reading this and encountering the same problem, instead of:

```php
<?php
class Contest extends DataObject {
    private static $has_many = [
        "GrandPrizeWinners" => "Person",
        "RunnersUp" => "Person",
    ];
}

class Person extends DataObject {
    private static $has_one = [
        "GrandPrize" => "Contest",
        "RunnerUp" => "Contest",
    ];
}
```

Do this instead:

```php
<?php
class Contest extends DataObject {
    private static $has_many = [
        "GrandPrizeWinners" => "Person.GrandPrize",
        "RunnersUp" => "Person.RunnerUp",
    ];
}

class Person extends DataObject {
    private static $has_one = [
        "GrandPrize" => "Contest",
        "RunnerUp" => "Contest",
    ];
}
```

**And make sure** you perform a `dev/build` as well, otherwise it will not work until you do.

patricknelson:
Thanks again @kinglozzer!

PR #4094, opened 2015-04-22
This adds the ability to toggle a field description

flashbackzoo:
If a field requires a long description it can be useful have it hidden until the user needs to read it.

*Description hidden:*
more-info-closed

*Description show:*
more-info-open

I'm looking for feedback on the technical approach here, so I have only pushed the initial show/hide logic at this stage. There is additional styling required to get the layout looking like the screenshots.


chillu:
Interesting! When developing the help text I think we deliberately decided to show it by default. It would be good to make the "info" icons opt-in though (globally and per field), as an easier way to show longer help text.

Please have a read through historical discussions to get context:
https://github.com/silverstripe/silverstripe-framework/pull/969
https://github.com/silverstripe/silverstripe-framework/pull/1029
https://github.com/silverstripe/silverstripe-framework/pull/986
https://github.com/silverstripe/silverstripe-framework/pull/525
https://github.com/silverstripe/silverstripe-framework/pull/979

Make sure the styling works correctly with different field layout combos, like in `FieldGroup` or a `DateField` with a `dmy` configuration option (three separate fields). There's a `frameworktest` module which creates many of these combinations, I've used it for similar sanity checks before: https://github.com/silverstripe-labs/silverstripe-frameworktest

forms/FormField.php
```diff
@@ -138,6 +138,8 @@ public static function name_to_label($fieldName) {
 	 * @return string
 	 */
 	public static function create_tag($tag, $attributes, $content = null) {
+		Requirements::javascript(FRAMEWORK_DIR . '/javascript/FieldToggleDescription.js');
```
tractorcow: This should be in the FromField() function, which is where all other formfields require their javascript.

It should also only be included if necessary. If you don't want to include this check, then you can put it in one of the existing framework/admin/javascript files; likely LeftAndMain.FieldHelp.js, which already has similar field helpers.

javascript/FieldToggleDescription.js
```diff
@@ -0,0 +1,41 @@
+/**
+ * Enable toggling (show/hide) of the field's description.
+ */
+
+(function ($) {
+
+    $.entwine('ss', function ($) {
+
+        $('.toggle-description').entwine({
```
tractorcow: This classname is not documented anywhere is it? You'd need to include some kind of instructions for users.

I think the rule is, is this documentation in a place where someone not explicitly looking for it, but who could benefit from it, able to find it? Somewhere in docs/en/02_Developer_Guides/03_Forms/?

jonom: Should also be referenced (if not documented) here if this gets rolled in: http://doc.silverstripe.org/en/developer_guides/customising_the_admin_interface/how_tos/cms_formfield_help_text

forms/FormField.php
chillu: Also make sure that's included in the main `Requirements::combine_files()` call in `CMSMain`, we don't want to cause more asset loads to slow down the CMS

flashbackzoo: Thanks, I'll move it :)

flashbackzoo:
The naming convention is now more consistent with FieldHelp / Tooltip. And there are some docs.

forms/FormField.php
tractorcow: I think if it's in LeftAndMain we should remove it from here. This is a CMS specific feature not a field specific one?

flashbackzoo: thanks, yup that was an oversight on my part, i forgot to remove it.

flashbackzoo:
The initial styling is there now. Here's a screenshot of the styles applied in the files area.
file-upload

There are a lot of image file changes as a result of adding an icon to the buttons sprite (`admin/images/btn-icon/information.png`) and running a `compass compile`. I've had a good look around the CMS and everything seems to be presenting fine.

I'm going to do some testing on other field types as @chillu suggested, and as a result probably push up another commit :)

flashbackzoo:
Here are some screenshots of various field types.

toggle-desc-checkboxes
toggle-desc-dropdown
toggle-desc-html
toggle-desc-text
toggle-desc-textarea
toggle-desc-upload

jonom:
Hey @flashbackzoo you may have already tested but just wondering if this works for a
>DateField with a dmy configuration option (three separate fields)

as chillu pointed out, and also on individual fields nested within a `FieldGroup`?

flashbackzoo:
Hey - Yup it works on `DateField` with three fields.

For a `FieldGroup` you can add a toggleable description to the group.

tractorcow:
Thanks @flashbackzoo :D

PR #4097, opened 2015-04-22
BUG Fix tinymce errors crashing CMS

tractorcow:
When removing a tinymce field, internal third party errors should be caught and ignored gracefully rather than breaking the whole CMS.

Fixes https://github.com/silverstripe-labs/silverstripe-content-widget/issues/2

### DIFF
--- a/javascript/HtmlEditorField.js
+++ b/javascript/HtmlEditorField.js
@@ -293,8 +293,15 @@ ss.editorWrappers['default'] = ss.editorWrappers.tinyMCE;
 			onremove: function() {
 				var ed = tinyMCE.get(this.attr('id'));
 				if (ed) {
-					ed.remove();
-					ed.destroy();
+					try {
+						ed.remove();
+					} catch(ex) {}
+					try {
+						ed.destroy();
+					} catch(ex) {}
+					
+					// Remove any residual tinyMCE editor element
+					this.next('.mceEditor').remove();
 
 					// TinyMCE leaves behind events. We should really fix TinyMCE, but lets brute force it for now
 					$.each(jQuery.cache, function(){


tractorcow:
Maybe @chillu could have a look at this. Is it acceptable to handle errors in third party libs this way, or should we console.log them?

hafriedlander:
Do you know which of the two functions the errors come from? Or put another way, should remove & destroy be in separate try/catch blocks (so destroy is still called if remove fails). And what happens if remove fails, is TinyMCE just not removed? So many questions.

sminnee:
Although discarding an error is often an anti-pattern, I think it makes sense in this case, because the action is getting rid of something rather than adding something. The consequence of a failure in the relevant code is presumably that a junk TinyMCE object persists in memory, which is tolerable.

If `ed.remove()` fails I'd expect that we'd still want to call `ed.destroy()`. FYI code for both is here: https://github.com/silverstripe/silverstripe-framework/blob/3.1/thirdparty/tinymce/tiny_mce_src.js#L14722

sminnee:
One of the issues that we might run into is id `ed.remove()` fails before the last line, `DOM.remove(self.getContainer())` doesn't actually get called. Does this mean that the TinyMCE widget is still visible?

Do we have any insight as to what kind of error is being called that necessitates this change?

tractorcow:
The message comes from the .remove(), with the actual exception being linked to a message buried deep within automatically generated / minified code.

Here is the entire stacktrace.

image

tractorcow:
I've updated this to indepently catch errors on both remove and destroy, and added a cleanup .remove on the tinyMce element in case it remains afterwards.

PR #4098, opened 2015-04-23
DOCS: formatting fix

phillprice:
the end # removed from titles as they were showing on the page.


camfindlay:
Cheers for that @phillprice , I think there was also a ticket over on the docsviewer module to support closed markdown tags (which is technically part of the spec).

PR #4099, opened 2015-04-23
FIX use config for Security::$login_url

dhensby:
Currently the config settings are ingnored wen setting the `login_url` on `Security`, this fixes that problem


tractorcow:
If this wasn't a config setting, but now is, does that make it a new feature? Should that means it now needs to go into 3.2?

micmania1:
I already made this change to 3.2 as well as adding logout_url: https://github.com/silverstripe/silverstripe-framework/commit/229bea399bf5942455cae43ed55c01640497035a

willmorgan:
I reckon this is a bug fix, so the current PR target is fine IMHO.

PR #4101, opened 2015-04-23
FIX #4100 Setup the ability to overload the ShortcodeParser class and en...

### LINKED ISSUE #4100: Not able to override/overload ShortcodeParser (not a child of Object)
I'm not able to override functionality of the `ShortcodeParser` class because it's not a child of the core `Object` class like the other parsers setup under `framework/parsers`. Ideally I'd like to be able to do the following somewhere in my code (early on in execution):

```php
Object::useCustomClass('ShortcodeParser', 'CustomShortcodeParser', true);
```

I'm considering a quick PR which will tweak this class to make it a little more extensible for these sorts of situations (I'm already having to do this in my own project to address unrelated bugs).

patricknelson:
Setup the ability to overload the ShortcodeParser class and ensuring its methods/properties are extensible via "static" keyword. Fixes issue #4100.

### DIFF
--- a/parsers/ShortcodeParser.php
+++ b/parsers/ShortcodeParser.php
@@ -9,15 +9,15 @@
  * @package framework
  * @subpackage misc
  */
-class ShortcodeParser {
+class ShortcodeParser extends Object {
 	
 	public function img_shortcode($attrs) {
 		return "<img src='".$attrs['src']."'>";
 	}
 	
-	private static $instances = array();
+	protected static $instances = array();
 	
-	private static $active_instance = 'default';
+	protected static $active_instance = 'default';
 	
 	// --------------------------------------------------------------------------------------------------------------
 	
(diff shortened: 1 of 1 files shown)


patricknelson:
Updated commit to change access to local properties/constants (which were changed to protected) to use `self` keyword instead of `static` keyword due to the Travis CI build errors for PHP 5.3.

tests/filesystem/FileTest.php
```diff
@@ -12,7 +12,7 @@ class FileTest extends SapphireTest {
 	public function testLinkShortcodeHandler() {
 		$testFile = $this->objFromFixture('File', 'asdf');
 
-		$parser = new ShortcodeParser();
+		$parser = ShortcodeParser::create();
```
willmorgan: I think we should keep it as it is here, as in the tests we want to be fully sure of what we're testing.

patricknelson: That's a very good point. I've reverted that change.

chillu:
Thanks Patrick! In general, we have to move those enhancements to the 3.2 release line ("3" branch here in framework), but its a low risk change so making an exception here.

patricknelson:
Ok no problem. That being said: Do you think maybe #4140 should then be rebased upon `3` instead of `3.1` like it is now?

PR #4102, opened 2015-04-24
NEW Added Youtube's short URL.

micmania1:
At the moment using the short URL (eg. https://youtu.be/qA0T8WumxT4) defaults to http. This introduces issue when running over https.


patricknelson:
Awesome I'm glad this went into 3.1.

PR #4105, opened 2015-04-25
Update 04_GridField.md docs for typo

AntonyThorpe:
Variable name should be $grid, as defined when creating the GridField just above.  Thanks.


willmorgan:
Good spot

Issue #4104, opened 2015-04-24
Make ListField (possibly other fields) argument error messages more intuitive.


patricknelson:
Just a minor complaint. Here's a copy/paste from https://groups.google.com/forum/#!topic/silverstripe-dev/-LBZExd2exQ

---
At first glance, would any of you know what this means?

    No associative arrays allowed multiple=true

I do now, but if you haven't seen this before it may be a bit confusing (see actual code here). Basically, all you see is that line above with a stack trace. It turns out that this is a very logical error validation, but the message could be improved, in my opinion. 

Basically: When passing an array of current values to a ListboxField() (when you set it up to allow multiple values), it's checking to make sure you don't accidentally give it an associative array. There's one key word missing from that error message: "values". I'd suggest maybe changing this error to something a bit more like:

     No associative arrays are allowed as values (when multiple=true).

That should make it easier to figure out what on earth just happened. It was confusing because I didn't receive the error until after saving my form, so it seemed fairly random.

PR #4106, opened 2015-04-25
FIX for #4104, minor revision of error messages in ListboxField (more intuitive).

patricknelson:
For issue #4104. I figured this would be more intuitive. Explaining why only indexed arrays are accepted, given the possibility of the programmer accidentally providing an array of values as keys with the titles as array values (when only the keys were intended to be saved, even though the object utilizes the array values for retention in the DB).


chillu:
Thanks for caring about this level of error feedback, Patrick!

PR #4108, opened 2015-04-26
Fix/Performance: Combine multiple queries into one for ManyManyList::getExtraData

Turnerj:
Previously when calling `getExtraData`, it would take an individual query per extra field to get the data. This PR adjusts that to combine each field into the single query. Beyond passing an array to the `SQLQuery` and looping over the columns in the result, the functionality hasn't changed.

`GridFieldDetailForm_ItemRequest` is the only reference in Framework or CMS that actually uses the `getExtraData` function.


Turnerj:
*bump* Is this change good to merge in?

tractorcow:
It does look good, and it seems to be still covered by tests so you didn't need any new ones. :)

PR #4112, opened 2015-04-27
Clean up CheckboxField

assertchris:


assertchris:
@tractorcow ping

forms/CheckboxField.php
```diff
@@ -6,59 +7,102 @@
...
+	 */
 	public function getAttributes() {
-		$attrs = parent::getAttributes();
-		$attrs['value'] = 1;
+		$attributes = array(
+			'type' => 'checkbox',
+			'value' => 1,
+		);
+
+		if($value = $this->Value()) {
```
tractorcow: `$value` is unused, so let's remove it?

assertchris:
Fixed

forms/CheckboxField.php
```diff
@@ -6,59 +7,102 @@
...
 	public function getAttributes() {
-		$attrs = parent::getAttributes();
-		$attrs['value'] = 1;
+		$attributes = array(
+			'type' => 'checkbox',
+			'value' => 1,
+		);
+
+		if($this->Value()) {
+			$attributes['checked'] = 'checked';
```
tractorcow: I just noticed; Prior to this change 'checked' would be null if not selected, but now it's not even assigned. Should we assign null in an else?

assertchris:
Like that?

assertchris:
@tractorcow ping

tractorcow:
I think we should start putting these minor annoyance fixes into 3.2 branch; We aren't going to be merging up from 3.1 to 3.2 if we can help it.

PR #4113, opened 2015-04-27
Clean up NumericField

assertchris:


assertchris:
@tractorcow ping

forms/NumericField.php
```diff
@@ -4,138 +4,196 @@
...
+	 * @param array $data
+	 *
+	 * @return $this
+	 *
+	 * @throws Zend_Locale_Exception
+	 */
 	public function setValue($value, $data = array()) {
 		require_once "Zend/Locale/Format.php";
 
-		// If passing in a non-string number, or a value
```
tractorcow: Not sure why this comment was removed; This was here to help explain the logic in this method for future developers and I think it still makes sense to keep that justification in place.

tractorcow: In 4.0 I want to also include setDataValue and loadFrom (to complement dataValue and saveInto), at which point we can move this logic out of here. ;)

assertchris:
Fixed

PR #4114, opened 2015-04-27
Clean up NullableField

assertchris:


forms/NullableField.php
```diff
@@ -1,75 +1,98 @@
...
 	protected $isNullLabel;
-	
 
 	/**
 	 * Create a new nullable field
-	 * @param $valueField
-	 * @return NullableField
+	 *
+	 * @param FormField $valueField
+	 * @param null $isNullLabel
```
tractorcow: boolean instead of null?

forms/NullableField.php
```diff
@@ -77,53 +100,80 @@ public function getIsNullId() {
...
+		);
 	}
 
 	/**
 	 * Value is sometimes an array, and sometimes a single value, so we need to handle both cases
+	 *
+	 * @param mixed $value
+	 * @param null|array $data
+	 *
+	 * @return static
```
tractorcow: Should we use `$this` return type? static implies the same (late bound) type, but not necessarily the same instance, and is probably more useful for static methods.

In this case I think `$this` adds more value. :D

See http://www.phpdoc.org/docs/latest/guides/types.html for reference.

forms/NullableField.php
```diff
@@ -77,53 +100,80 @@ public function getIsNullId() {
...
+		if($this->isReadonly()) {
 			$nullableCheckbox = new CheckboxField_Readonly($this->getIsNullId());
 		} else {
 			$nullableCheckbox = new CheckboxField($this->getIsNullId());
 		}
+
 		$nullableCheckbox->setValue(is_null($this->dataValue()));
 
-		return $this->valueField->Field() . ' ' . $nullableCheckbox->Field()
-			. '&nbsp;<span>' . $this->getIsNullLabel().'</span>';
```
tractorcow: Could the layout be broken by removing the `&nbsp;`? Possibly this is intended to prevent the label wrapping to the next line?

assertchris: [this](https://github.com/silverstripe/silverstripe-framework/pull/4114/files#diff-6e6a6671e3877cda188e546cd9a96624R117)

assertchris:
@tractorcow ping

forms/NullableField.php
```diff
@@ -1,75 +1,98 @@
...
+	 * @param FormField $valueField
+	 * @param null|string $isNullLabel
 	 */
 	public function __construct(FormField $valueField, $isNullLabel = null) {
 		$this->valueField = $valueField;
-		$this->isNullLabel = $isNullLabel;
-		if ( is_null($this->isNullLabel) ) {
-			// Set a default label if one is not provided.
+
+		if($isNullLabel) {
```
tractorcow: We need to support `false` for explicitly empty labels. Can we change this to `if(isset($isNullLabel))` to keep the current behaviour?

forms/NullableField.php
```diff
@@ -77,53 +100,80 @@ public function getIsNullId() {
...
+	 * @param mixed $value
+	 * @param null|array $data
+	 *
+	 * @return $this
 	 */
 	public function setValue($value, $data = null) {
-		if ( is_array($data) && array_key_exists($this->getIsNullId(), $data) && $data[$this->getIsNullId()] ) {
+		$id = $this->getIsNullId();
+
+		if(is_array($data) && array_key_exists($id, $data) && $data[$id]) {
```
tractorcow: or just `if(is_array($data) && !empty($data[$id]))` ? :P

assertchris:
Fixed

PR #4116, opened 2015-04-27
Clean up TextareaField

assertchris:


forms/TextareaField.php
```diff
@@ -14,58 +13,96 @@
...
 	public function setColumns($cols) {
 		$this->cols = $cols;
+
 		return $this;
 	}
 
+	/**
+	 * @return int
+	 */
+	public function getColumns() {
```
tractorcow: If it's a new public API method it probably needs to go into 3.2 (3 branch).

tractorcow:
See my comment on https://github.com/silverstripe/silverstripe-framework/pull/4117 about duplicate PHPDoc in subclasses.

forms/TextareaField.php
assertchris: ARG. Do you want me to remove these getters for this PR?

tractorcow: I don't _want_ you to remove it.

I simply require you to. :D

assertchris:
@tractorcow ping

PR #4118, opened 2015-04-27
Clean up LiteralField

assertchris:


forms/LiteralField.php
```diff
@@ -1,76 +1,103 @@
...
- * 
+ *
  * @package forms
  * @subpackage fields-dataless
  */
 class LiteralField extends DatalessField {
-	
 	/**
-	 * @var string $content
+	 * @var string|FormField
```
tractorcow: I think we are just meant to use string. If you use a FormField as a value, the field internally relies on the `__toString` to cast it safely to HTML safe text. Should we thus limit our expectations to strings only?

tractorcow: When does this object contain another sub-formfield?

assertchris: I don't know - Just documenting the behaviour that exists.

tractorcow: Where did you see this?

assertchris: [here](https://github.com/silverstripe/silverstripe-framework/pull/4118/files#diff-3f1ed6b2e8300f694532ef48e723667bL35)

tractorcow: Oh well, merge it now mark as deprecated later I guess.

forms/LiteralField.php
```diff
@@ -1,76 +1,103 @@
...
 	public function FieldHolder($properties = array()) {
-		if(is_object($this->content)) {
-			$obj = $this->content;
-			if($properties)
-				$obj = $obj->customise($properties);
-			return $obj->forTemplate();
-		} else {
-			return $this->content;
+		if($this->content instanceof FormField) {
+			$context = $this->content;
```
tractorcow: Could the `$this->context` possibly be intended for DBField as well as FormField? Maybe it should be `instanceof ViewableData`? (Although even that method doesn't implement forTemplate, it's the highest common denominator of classes that do).

assertchris:
@tractorcow ping

tractorcow:
It's like, uh, a kind of, literal field, man.

field

PR #4120, opened 2015-04-27
FIX: X-Reload & X-ControllerURL didn't support absolute URLs (fixes #4119)

### LINKED ISSUE #4119: X-Reload header doesn’t allow redirecting to external URLs
Perhaps I’m approaching this the wrong way as I have a fairly specific use-case. I need to submit a CMS form via AJAX and, depending on validation etc, either handle reloading the form as “usual”, or redirect to an external payment gateway.

I’d hoped that adding the `X-Reload` header would allow me to redirect to the URL, but it only redirects to relative URLs: https://github.com/silverstripe/silverstripe-framework/blob/d2a3da22039d6af06f1f1f60d15f563c4fcfbe3f/admin/javascript/LeftAndMain.js#L634-L639

kinglozzer:


tractorcow:
Has this been tested in a site with a non `/` baseurl? e.g. `http://localhost/mysite`

kinglozzer:
Yep. The only “core” use I can find is in the CMS: `CMSProfileController::save()` for forcing a full reload after changing locale. I’ve tested that functionality on a `/` and `/silverstripe` base URL and it still works fine :)

PR #4124, opened 2015-04-28
Removed case insensitivity on First Sentence

cmcramer:
Sentences aren't found if the last word ends in _mrs _ms. Removed case insensitive flag and specified specific cases to skip. Removed "Miss" as this is not an abbreviation and is properly spelled without a .


cmcramer:
Closing this pull request. Will resubmit.

PR #4125, opened 2015-04-28
Clean up FormField

assertchris:
:trollface:

### DIFF
--- a/forms/FormField.php
+++ b/forms/FormField.php
@@ -1,19 +1,25 @@
 <?php
+
 /**
- * Represents a field in a form. 
- *  
+ * Represents a field in a form.
+ *
  * A FieldList contains a number of FormField objects which make up the whole of a form.
- * In addition to single fields, FormField objects can be "composite", for example, the {@link TabSet}
- * field.  Composite fields let us define complex forms without having to resort to custom HTML.
- * 
- * <b>Subclassing</b>
- * 
- * Define a {@link dataValue()} method that returns a value suitable for inserting into a single database field. 
- * For example, you might tidy up the format of a date or currency field.
- * Define {@link saveInto()} to totally customise saving. 
- * For example, data might be saved to the filesystem instead of the data record, 
- * or saved to a component of the data record instead of the data record itself.
- * 
+ *
+ * In addition to single fields, FormField objects can be "composite", for example, the
+ * {@link TabSet} field. Composite fields let us define complex forms without having to resort to
+ * custom HTML.
+ *
+ * To subclass:
+ *
+ * Define a {@link dataValue()} method that returns a value suitable for inserting into a single
+ * database field.
+ *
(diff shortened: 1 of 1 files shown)


forms/FormField.php
```diff
@@ -24,153 +30,244 @@ class FormField extends RequestHandler {
...
 	public static function create_tag($tag, $attributes, $content = null) {
 		$preparedAttributes = '';
-		foreach($attributes as $k => $v) {
-			// Note: as indicated by the $k == value item here; the decisions over what to include in the attributes
-			// can sometimes get finicky
-			if(!empty($v) || $v === '0' || ($k == 'value' && $v !== null) ) {
-				$preparedAttributes .= " $k=\"" . Convert::raw2att($v) . "\"";
+
+		foreach($attributes as $attributeKey => $attributeValue) {
+			if(!empty($attributeValue) || $attributeValue === '0' || $attributeKey == 'value') {
```
tractorcow: You seem to have lost the `$attributeValue !== null` check from the original code. Is that intentional?

assertchris: Yes. `!empty($v)` means `$v !== null`.

tractorcow: Yes but you'll see that if it was null previously and $k was 'value' then it would not be added to the attributes. Now it does. The `!empty()` is on the wrong side of the `||`.

tractorcow:
No other complaints :)

forms/FormField.php
tractorcow: `if(!empty($attributeValue) || $attributeValue === '0' || ($attributeKey == 'value' && isset($attributeValue))) {`

tractorcow: I think the intention is if `$attributeValue === false` we should bypass the !empty check, but we should still hide it if it's `=== null`.

assertchris:
@tractorcow ping

forms/FormField.php
```diff
@@ -310,199 +437,251 @@ public function addExtraClass($class) {
...
 		return Convert::raw2att($this->value);
 	}
-	
+
 	/**
 	 * Set the field value.
-	 * 
+	 *
 	 * @param mixed $value
-	 * @param mixed $data Optional data source passed in by {@see Form::loadDataFrom}
```
tractorcow: I hate to say this but this parameter is super important; We need to add it in formally in 4.0, but we can't remove the informal existence of this variable at this stage.

See my note on https://github.com/silverstripe/silverstripe-framework/pull/4258 where I very nearly did this but decided to wait until a separate PR. :)

tractorcow:
Additional comment is purely not a fault on your part; The API is terrible and it should feel terrible. Let's fix it properly in 4.0 for sure. :D

assertchris:
How's that?

tractorcow:
Glad you got the tests to pass. :D

dhensby:
Weirdest issue ever... Adding the optional parson to setValue broke all the things...

PR #4126, opened 2015-04-28
Clean up HeaderField

assertchris:


forms/HeaderField.php
```diff
@@ -8,52 +9,80 @@
...
-		
-		if($headingLevel) $this->headingLevel = $headingLevel;
-		
+
+			if(isset($args[1])) {
+				$headingLevel = $args[1];
+			}
+
+			if(isset($args[3])) {
+				$this->setForm($args[3]);
```
tractorcow: The old code ignores $form and passing in $form into the field contstructor has been deprecated since 3.x.0 (can't recall exactly). Same argument as https://github.com/silverstripe/silverstripe-framework/pull/4128/files... better to remove the unused variable than try to call setForm where it wasn't called before.

assertchris:
@tractorcow ping

forms/HeaderField.php
```diff
@@ -8,52 +9,76 @@
...
+	 */
 	public function __construct($name, $title = null, $headingLevel = 2) {
-		// legacy handling for old parameters: $title, $heading, ...
-		// instead of new handling: $name, $title, $heading, ...
+		// legacy handling:
+		// $title, $headingLevel...
 		$args = func_get_args();
+
 		if(!isset($args[1]) || is_numeric($args[1])) {
-			$title = (isset($args[0])) ? $args[0] : null;
```
tractorcow: `if it's not null, set it to that value. if it is null, set it to null`. Seems super redundant?

PR #4128, opened 2015-04-28
Clean up LabelField

assertchris:


forms/LabelField.php
```diff
@@ -1,30 +1,38 @@
...
+
+			if(isset($args[0])) {
+				$title = $args[0];
+			}
+
+			// Prefix name to avoid collisions.
+			$name = 'LabelField' . $title;
+
+			if(isset($args[3])) {
+				$this->setForm($args[3]);
```
tractorcow: form parameter in the constructor is actually deprecated (I think in 3.1.0 or 3.0.0). I found this out after merging a "fix" for a field by setting 'setForm' and making someone else angry upstream, I think in FormAction?. I think we should just drop this.

forms/LabelField.php
```diff
@@ -1,30 +1,38 @@
...
-			$form = (isset($args[3])) ? $args[3] : null;
-		} 
-		
+
+			if(isset($args[0])) {
+				$title = $args[0];
+			}
+
+			// Prefix name to avoid collisions.
+			$name = 'LabelField' . $title;
```
tractorcow: I think it's ok for fields to not have names if they aren't datafields. Composite fields (FieldGroup etc) don

assertchris: This would be a breaking change.

forms/LabelField.php
assertchris: This would be a breaking change.

forms/LabelField.php
tractorcow: I'm not saying change it, I'm saying "don't change it". This PR is making the change.

forms/LabelField.php
tractorcow: No it wouldn't be. I'm saying remove the unused `$form` variable instead of now passing it into a new function call.

forms/LabelField.php
assertchris: The name was added before. The change is collision avoidance (by prefix) lifted out of `HeaderField` (https://github.com/silverstripe/silverstripe-framework/pull/4126/files#diff-3bdc3b156f8ea291aba1bd58b45153e3L26).

Issue #4129, opened 2015-04-28
DataObject must return "null" not "false" on has_many()


patricknelson:
I found a bug in `SortableUploadField` ([issue here](https://github.com/bummzack/sortablefile/issues/37)) that resulted from the fact that `has_one()` and `many_many()` both return null but `has_many()` does not, as one would expect for consistency. For example, I should be able to do the following:

```php
$isManyMany = ($record->many_many($relationName) !== null);
$isHasMany = ($record->has_many($relationName) !== null);
```

Instead, based on current functionality, you would have to do this:

```php
$isManyMany = ($record->many_many($relationName) !== null);
$isHasMany = ($record->has_many($relationName) !== false);
```

I recognize part of the reason this is coming up now is because people may be more likely performing a more loose equality test instead of a strict test (i.e. `===`) like I am in this case. Either way, the return types should probably be consistent!

patricknelson:
Merged.

PR #4130, opened 2015-04-28
FIX for #4129: Ensure has_many() returns null instead of false to be consistent with other relation component methods.

patricknelson:
Addresses #4129. Relevant change on line [1721](https://github.com/silverstripe/silverstripe-framework/pull/4130/files#diff-c4a683f386eac1b20ffeca437649698fR1721). I also ensured that the documented return types were consistent with what was actually returned, since this methods can sometimes not return anything at all (or, specifically, can `return null`, effectively similar in PHP to not returning).

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -1598,7 +1598,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	 *
 	 * @param string $component Name of component
 	 *
-	 * @return string|array The class of the one-to-one component, or an array of all one-to-one components and their
+	 * @return string|array|null The class of the one-to-one component, or an array of all one-to-one components and their
 	 *                      classes.
 	 */
 	public function has_one($component = null) {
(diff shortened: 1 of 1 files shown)


kinglozzer:
I agree that this should be addressed, but I don’t think this should be done in a patch release as it’s a potentially breaking change (if someone is using a strict equality check, this would break their code). As semantic versioning dictates that backwards-incompatible changes should be for major releases, and this is a pretty easy issue to work around, I think this should go against master instead of 3.1 :)

patricknelson:
So to verify -- `master` instead of `3`?

kinglozzer:
Yep, master please. Could you also apply the same fix to `belongs_to()` (or `belongsToComponent()` as it is now named in master)? Thanks!

patricknelson:
No problem -- I'll try to rebase this on `master` instead. It looks like I'll need a new PR for this so, hah. https://github.com/isaacs/github/issues/18

patricknelson:
See here: https://github.com/silverstripe/silverstripe-framework/pull/4131

PR #4133, opened 2015-04-28
Check that LastVisited field exists before making it readonly

nimeso:


dhensby:
Seems sensible, but under what circumstances does the field not exist?

dhensby:
Merged into 3.2 because 3.1 is security only fixes now

PR #4136, opened 2015-04-29
Documentation tidy for Requirements.php

nikrolls:
Documentation-only changes, except where a couple of methods have been moved to live next to other related functions.

Among other things, this will fix the annoying errors in IDEs that are expecting variables of type `The` :)


nikrolls:
The build error seems to be an issue with MySQL in Travis rather than the PR itself.

view/Requirements.php
```diff
@@ -152,116 +161,166 @@ public static function css($file, $media = null) {
...
+	 * @param array  $files            Array of filenames relative to docroot
 	 *
-	 * @param string $combinedFileName
-	 * @param array $files
-	 * @param string $media
+	 * @return bool|void
 	 */
-	public static function combine_files($combinedFileName, $files, $media = null) {
-		self::backend()->combine_files($combinedFileName, $files, $media);
+	public static function combine_files($combinedFileName, $files) {
```
tractorcow: This line needs fixing to put back `$media`

Issue #4139, opened 2015-04-29
Not able to override/overload HtmlEditorConfig (not a child of Object)


patricknelson:
Similar to issue #4101, the class `HtmlEditorConfig` is not a child of `Object` and isn't able to utilize the features of the SS injector `Object::create()`.

Also, the `->generateJS()` does a lot more than just generate JS code, since it also is taking a base `$config` (from `$this->settings`) and adding more options to it as a final array of settings before finally converting to JSON for TinyMCE's use. This is generally fine (before) but now that this can be extended, it conflicts somewhat with D.R.Y. since, if I needed to override that method, I'd have to copy/paste all of that code in order to better re-generate my own version of the JS code. The obvious issue here is that the method by which that config may change over time (bug fixes, enhancements, etc) but the overriding method will not change and increases exposure for more bugs over time.

Anyway -- I think I've already got some improvements for this, so a PR is well on its way as usual :)

maxime-rainville:
`HTMLEditorConfig` now uses the `Configurable` and `Injectable` trait in 4. So this is not needed anymore.

PR #4122, opened 2015-04-27
NEW feature #4121 Ability to find existing tab by name without creating a new tab.

patricknelson:
Proof of concept for feature outlined in #4121.

PR #4140, opened 2015-04-29
FIX for #4139 allow HtmlEditorConfig the ability to be overloaded via injector.

patricknelson:
Addresses #4139. Abstracted `->generateJS()` to several methods for better extensibility as well.

**Note:** This is similar (not technically related) to #4101 where the same type of work as done to make it overall more extensible.

### DIFF
--- a/forms/HtmlEditorConfig.php
+++ b/forms/HtmlEditorConfig.php
@@ -11,11 +11,11 @@
  * @package forms
  * @subpackage fields-formattedinput
  */
-class HtmlEditorConfig {
+class HtmlEditorConfig extends Object {
 
-	private static $configs = array();
-	
-	private static $current = null;
+	protected static $configs = array();
+
+	protected static $current = null;
 	
 	/**
 	 * Get the HtmlEditorConfig object for the given identifier. This is a correct way to get an HtmlEditorConfig
@@ -26,8 +26,8 @@ class HtmlEditorConfig {
 	 *                            identifier
 	 */
 	public static function get($identifier = 'default') {
-		if (!array_key_exists($identifier, self::$configs)) self::$configs[$identifier] = new HtmlEditorConfig();
-		return self::$configs[$identifier];
+		if (!array_key_exists($identifier, static::$configs)) static::$configs[$identifier] = HtmlEditorConfig::create();
+		return static::$configs[$identifier];
 	}
 	
 	/**
@@ -36,7 +36,7 @@ class HtmlEditorConfig {
 	 * @return null
 	 */
 	public static function set_active($identifier = null) {
-		self::$current = $identifier;
+		static::$current = $identifier;
 	}
 	
 	/**
@@ -44,8 +44,8 @@ class HtmlEditorConfig {
 	 * @return HtmlEditorConfig - the active configuration object
 	 */
 	public static function get_active() {
-		$identifier = self::$current ? self::$current : 'default';
-		return self::get($identifier);
+		$identifier = static::$current ? static::$current : 'default';
+		return static::get($identifier);
 	}
 	
 	/**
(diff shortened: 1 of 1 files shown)


forms/HtmlEditorConfig.php
```diff
@@ -289,35 +289,78 @@ public function removeButtons() {
...
+
+		return $config;
+	}
+
+	/**
+	 * Returns an array of internal plugins.
+	 *
+	 * @return array
+	 */
+	public function getInternalPlugins() {
```
chillu: Can you please make the new functions protected? We want to keep the HTMLEditorConfig as agnostic to the JS implementation as possible, and the "internal" vs "external" definition is specific to TinyMCe.

patricknelson: Sure, no problem. I thought about that, but forgot about it as I was still using the public methods for my smoke testing.

chillu:
As long as you keep the new methods `protected`, I think 3.1 is fine. That's just a heads up that we'll get more picky with merges as the branch matures, and 3.2 gets closer to release.

dhensby:
@chillu any more feedback?

dhensby:
This should be based on 3, so I'll switch that...

dhensby:
It needs rebasing on 3 now, please :

patricknelson:
Hi @dhensby if you look at the code, you'll see this aught to be pretty frustrating since it sat for so long. I updated my PR almost immediately as soon as @chillu requested but since then it has diverged quite a bit. So it's not going to be a very easy thing to rebase, especially since it was updated since then with more *static* methods which are counter to the extensibility that I was trying to add (via utilization of `::create()` for injector use). That is, if this were merged first, I bet @tractorcow would have found a solution "API Support for multiple HTMLEditorConfig per page" which itself also modifies precisely the came code that my commit modified :disappointed:  See: https://github.com/silverstripe/silverstripe-framework/commit/a8ace7534194b5e1a636c96eca4607d08726dfeb

So... if you look at my code, I set it up so that it can be overridden. If you look at @tractorcow's code, since it's a static method `::require_js()` that replaces `->generateJS()` (which I outlined in my previously submitted issue #4139 that it did more than simply generate JS, it did a bit too much and was thus a bit cumbersome to override). That's why I think we both ended up editing the same code, since that method maybe does a bit too much and should be separated to multiple methods. For the record, those new **instance level** methods (at the time when I wrote it, because I cannot yet take the time to remotely comprehend what @tractorcow did since I am busy) are:

- `->getConfig()`
- `->getInternalPlugins()`
- `->getExternalPlugins()`
- `->getExternalPluginsJS()`

Thanks @dhensby for taking the time to do the footwork on following up on these, but I due to the lag here I won't be able to really dive into this until later. I just wish these PR's didn't get so stale.

It's possible that this can still be saved (conceptually it seems easy) by simply retaining `::require_js()` but set it up so that it then simply uses `::create()` to manufacture the `Injector`-based instance to then call `->generateJs()` which would have all the heavy-lifting/delegation in it again. Reintroducing the method that was once removed. Then, since it was removed, make it protected now instead of public to prevent access? Not sure about that... ideas?

Per my statement here it may be a few weeks before I can really get in to this (sorry again): https://github.com/silverstripe/silverstripe-framework/pull/4122#issuecomment-241556234

dhensby:
Maybe as punishment we should make @tractorcow merge this by hand, rather than making you rebase / do it all over

patricknelson:
_::cracks whip::_ :collision:

To be fair though, I'm sure he's busy too.

tractorcow:
Oh no, I'm in trouble again aren't I? =(

Um, I guess some time I'll need to rebase this for 3.x as well as 4.x then?

patricknelson:
No trouble from me :joy: At this point, I've been so busy and these have grown so old (partly my fault as well) that I've had to give up on it until I reach the mythical (yet enlightening) point of "free time". Which means not any time soon (at my company, this time of year is pretty busy as you'd expect).

maxime-rainville:

robbieaverill:
For what it's worth, this is already done in SilverStripe 4

patricknelson:
That's good news @robbieaverill Yet another reason to upgrade 😄

PR #4141, opened 2015-04-29
UPDATE: change all instances of $this->request to use appropriate getter/setter

stevie-mayhew:
See https://github.com/silverstripe/silverstripe-cms/pull/1194#issuecomment-97573329


stevie-mayhew:
I considered changing `$request` from `protected` to `private` but I can think of instances where I would want to overwrite setRequest on a subclass. The status quo as `protected` is probably best. Let me know if anyone has opinions on this, but that change can be made in master once this has been merged upwards as it is an API change and this PR doesn't change API, only add them so is suitable for the minor branch.

chillu:
Sweet! Also thanks for fixing docs in the same go.

PR #4143, opened 2015-04-30
API Use unbuffered MySQL Prepared statements by default

### LINKED ISSUE #3426: MySQLiConnector: mysqli_stmt::get_result is undefined if not using mysqlnd driver
```
Fatal error: Call to undefined method mysqli_stmt::get_result from context MySQLiConnector in /var/www/ss-master/framework/model/connect/MySQLiConnector.php on line 248
```

This doesn't work on some environments, specifically those that use libmysql instead of mysqlnd (e.g. HHVM). More info on that here: http://php.net/manual/en/mysqli-stmt.get-result.php.

tractorcow:
This solution avoids the usage of any MySQL Native Driver specific API, so it should work with libmysql. 

Using unbuffered SQL Statements should give a general performance increase, but we still use an internal buffer to support seek operations, or in cases where subsequent statements need to be prepared and require existing statements to release resources.

The non-prepared DB::query will still use the existing MySQLQuery object, but it has been stripped down a bit.

Fixes #3426

### DIFF
--- a/model/connect/MySQLStatement.php
+++ b/model/connect/MySQLStatement.php
@@ -1,0 +1,190 @@
+<?php
+
+/**
+ * Provides a record-view for mysqli statements
+ *
+ * By default streams unbuffered data, but seek(), rewind(), or numRecords() will force the statement to
+ * buffer itself and sacrifice any potential performance benefit.
+ */
+class MySQLStatement extends SS_Query {
+
+	/**
+	 * The related mysqli statement object if generated using a prepared query
+	 *
+	 * @var mysqli_stmt
+	 */
+	protected $statement;
+
+	/**
+	 * Metadata result for this statement
+	 *
+	 * @var mysqli_result
+	 */
+	protected $metadata;
+
+	/**
+	 * Is the statement bound to the current resultset?
+	 *
+	 * @var bool
+	 */
+	protected $bound = false;
+
+	/**
+	 * List of column names
+	 *
+	 * @var array
+	 */
+	protected $columns = array();
+
+	/**
+	 * List of bound variables in the current row
+	 *
+	 * @var array
+	 */
+	protected $boundValues = array();
+
+	/**
+	 * Array buffer of all rows in this result set
+	 *
+	 * @var array
+	 */
+	protected $rows = array();
+
+	/**
+	 * If the statement has been closed.
+	 * The statement will automatically close once all rows have been buffered.
+	 *
+	 * @var bool
+	 */
+	protected $closed = false;
+
+	/**
+	 * Force all remaining unbuffered rows to be buffered and close the statement
+	 */
+	public function buffer() {
+		// Force internal pointer to iterate to the end
+		while($this->bufferNext()) {}
+	}
+
+	/**
+	 * Release all resources held by this object
+	 */
+	protected function close() {
+		if($this->isClosed()) {
+			return;
+		}
+		$this->metadata->free();
+		$this->statement->close();
+		$this->closed = true;
+		$this->currentRecord = false;
+	}
+
+	/**
+	 * Have resources been released?
+	 *
+	 * @return boolean
+	 */
+	public function isClosed() {
+		return $this->closed;
+	}
+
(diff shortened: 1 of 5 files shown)


model/connect/MySQLStatement.php
```diff
@@ -0,0 +1,190 @@
...
+	 * Metadata result for this statement
+	 *
+	 * @var mysqli_resulst
+	 */
+	protected $metadata;
+
+	/**
+	 * Is the statement bound to the current resultset?
+	 *
+	 * @var boolean
```
assertchris: `bool` PLOX

model/connect/MySQLStatement.php
```diff
@@ -0,0 +1,190 @@
...
+	 * The related mysqli statement object if generated using a prepared query
+	 *
+	 * @var mysqli_stmt
+	 */
+	protected $statement;
+
+	/**
+	 * Metadata result for this statement
+	 *
+	 * @var mysqli_resulst
```
halkyon: misspelled "result" here :)

model/connect/MySQLStatement.php
tractorcow: why not boolean?

model/connect/MySQLStatement.php
tractorcow: oops

model/connect/MySQLStatement.php
assertchris: `bool` is consistent with more of the codebase, and with the STH RFC that is part of PHP 7.

tractorcow:
Fixed phpdoc

tractorcow:
It looks as though some people have reported out of memory issues with this. Maybe have to go with a "slower / more robust" solution?

camfindlay:
Is the prepared statements work listed on uservoice roadmap? Let me know so I can add it if not :)

tractorcow:
It's technically marked as "done". This is just bugfixing to that.

camfindlay:
thanks :+1:

tractorcow:
Alternative solution at https://github.com/silverstripe/silverstripe-framework/pull/4154. I need to figure out a better way of measuring memory usage.

Other solution should be more stable + use less memory but is slower.

tractorcow:
I've updated this to be less wasteful with memory; MySQLStatement::$rows now only stores row values instead of duplicating all column names.

assertchris:
You _could_ get a count of rows and use `SplFixedArray` instead...

tractorcow:
And I've figured out why that weird code was in the unit tests. Because PostgreSQL can't build MultiEnum values. ;P

tractorcow:
> You could get a count of rows and use SplFixedArray instead...

We could, but then we'd have to pre-buffer on every query, and we may as well use the other implementation then and drop unbuffered querying.

tractorcow:
I've gone ahead and removed the "skipped" tests, since they have been skipped since 3.0 days and no one has rewritten them (nor am I sure exactly how they should be written since it's not 100% clear what it's testing).

tractorcow:
Solutions to high memory usage: 

- Use a higher PHP version (seems to be less resource hungry)
- Use PDOConnector instead (it should be standard honestly)
- Use a higher memory limit

assertchris:
What about multiple connections and a connection pool?

tractorcow:
I wouldn't mind looking at connection pooling one day, but I think that's possibly outside the scope of this fix. All I'm trying to do is find a reasonable solution to users without mysql native driver, since `mysqli_stmt::get_results` doesn't exist.

It could be a huge amount of engineering. :)

assertchris:
Yeah - not a quick change. 

This doesn't feel like a good fix if it replaces one bug (undefined function) for another (out of memory).

tractorcow:
This needs some serious review. Can @silverstripe/core-team please look at this, as this is the last critical fix before we can release 3.2?

We need to choose one of the two solutions, or either come up with a third.

sminnee:
Killing this in favour of #4154

PR #4144, opened 2015-04-30
Postgres case sensitivity

RantyDave:
Deals with case sensitivity when setting up on postgres.

### DIFF
--- a/dev/install/install.php5
+++ b/dev/install/install.php5
@@ -148,6 +148,12 @@ if(isset($_REQUEST['db'])) {
 	);
 }
 
+//Postgres needs lowercase
+if ($databaseConfig['type'] == 'PostgreSQLDatabase') {
+	$databaseConfig['database'] = strtolower($databaseConfig['database']);
+	$databaseConfig['username'] = strtolower($databaseConfig['username']);
+}
+
 if(isset($_REQUEST['admin'])) {
 	// Disabled inputs don't submit anything - we need to use the environment (except the database name)
 	if($usingEnv) {


chillu:
Wouldn't this be better placed in https://github.com/silverstripe/silverstripe-postgresql/blob/master/code/PostgreSQLDatabaseConfigurationHelper.php? Do you have a Postgres reference confirming that lowercasing on both those properties is the default behaviour?

RantyDave:
> On 30/04/2015, at 7:48 pm, Ingo Schommer <notifications@github.com> wrote:
> 
> Wouldn't this be better placed in https://github.com/silverstripe/silverstripe-postgresql/blob/master/code/PostgreSQLDatabaseConfigurationHelper.php <https://github.com/silverstripe/silverstripe-postgresql/blob/master/code/PostgreSQLDatabaseConfigurationHelper.php>?
> 
Yes.

I was really new to Silverstripe at this point and was just trying to fix problems I had in the whole setting up process.
> Do you have a Postgres reference confirming that lowercasing on both those properties is the default behaviour?
> 
> 
This?

http://www.thenextage.com/wordpress/postgresql-case-sensitivity-part-1-the-ddl/ <http://www.thenextage.com/wordpress/postgresql-case-sensitivity-part-1-the-ddl/>

chillu:
Thanks for researching this, Dave! Since you agree that the fix is better placed in the `silverstripe-postgresql` module, I'll close this framework pull request. Feel free to open a new pull request over at the other module.

PR #4145, opened 2015-04-30
fix translation

stee1cat:


chillu:
Sorry, you'll need to do this on transifex, otherwise it'll get overwritten on the next import: http://doc.silverstripe.org/en/3.0/misc/contributing/translation/

PR #3381, opened 2014-08-13
FIX: Class names referenced in Config should be case-insensitive

sminnee:
The config system maps top-level key name to a PHP class. PHP classes are case-insensitive,
but without this patch, the Config system was case-sensitive on this variable. I think this
should be considered a bug.

The following patch makes the "class" key case-insensitive, but leave the "name" key, as
case-sensitive, mirroring the semantics of PHP classes and static variables, respectively.

Arguably, the entire config system could be case-insensitive, but I haven't done that in
this patch.

PR #4148, opened 2015-05-01
NEW Update SS_ConfigStaticManifest to use Reflection

micmania1:
This is the PR as discussed in #4029 

It completely removes the static manifest and replaces the SS_ConfigStaticManifest implementation with Reflection.

Tests on PHP 5.5 showed no speed decrease but saved 0.5mb memory per page load on a base install, loading the home page.

### DIFF
--- a/core/Core.php
+++ b/core/Core.php
@@ -112,7 +112,7 @@ if(file_exists(BASE_PATH . '/vendor/autoload.php')) {
 }
 
 // Now that the class manifest is up, load the static configuration
-$configManifest = new SS_ConfigStaticManifest(BASE_PATH, false, $flush);
+$configManifest = new SS_ConfigStaticManifest();
 Config::inst()->pushConfigStaticManifest($configManifest);
 
 // And then the yaml configuration
--- a/core/manifest/ConfigStaticManifest.php
+++ b/core/manifest/ConfigStaticManifest.php
@@ -1,385 +1,42 @@
 <?php
+
 /**
- * A utility class which builds a manifest of the statics defined in all classes, along with their
- * access levels and values
- *
- * We use this to make the statics that the Config system uses as default values be truely immutable.
- *
- * It has the side effect of allowing Config to avoid private-level access restrictions, so we can
- * optionally catch attempts to modify the config statics (otherwise the modification will appear
- * to work, but won't actually have any effect - the equvilent of failing silently)
+ * Allows access to config values set on classes using private statics.
  *
  * @package framework
  * @subpackage manifest
  */
 class SS_ConfigStaticManifest {
 
-	protected $base;
-	protected $tests;
-
-	protected $cache;
-	protected $key;
-
-	protected $index;
-	protected $statics;
-
-	static protected $initial_classes = array(
-		'Object', 'ViewableData', 'Injector', 'Director'
-	);
-
 	/**
-	 * Constructs and initialises a new config static manifest, either loading the data
-	 * from the cache or re-scanning for classes.
+	 * @param string $class
+	 * @param string $name
+	 * @param null $default
 	 *
-	 * @param string $base The manifest base path.
-	 * @param bool   $includeTests Include the contents of "tests" directories.
-	 * @param bool   $forceRegen Force the manifest to be regenerated.
-	 * @param bool   $cache If the manifest is regenerated, cache it.
+	 * @return mixed|null
 	 */
-	public function __construct($base, $includeTests = false, $forceRegen = false, $cache = true) {
-		$this->base  = $base;
-		$this->tests = $includeTests;
+	public function get($class, $name, $default = null) {
+		if(class_exists($class)) {
 
-		$cacheClass = defined('SS_MANIFESTCACHE') ? SS_MANIFESTCACHE : 'ManifestCache_File';
+			// The config system is case-sensitive so we need to check the exact value
+			$reflection = new ReflectionClass($class);
+			if(strcmp($reflection->name, $class) === 0) {
 
-		$this->cache = new $cacheClass('staticmanifest'.($includeTests ? '_tests' : ''));
-		$this->key = sha1($base);
-
-		if(!$forceRegen) {
-			$this->index = $this->cache->load($this->key);
-		}
-
-		if($this->index) {
-			$this->statics = $this->index['$statics'];
-		}
-		else {
-			$this->regenerate($cache);
-		}
-	}
-
-	public function get($class, $name, $default) {
-		if (!isset($this->statics[$class])) {
-			if (isset($this->index[$class])) {
-				$info = $this->index[$class];
-
-				if (isset($info['key']) && $details = $this->cache->load($this->key.'_'.$info['key'])) {
-					$this->statics += $details;
-				}
-
-				if (!isset($this->statics[$class])) {
-					$this->handleFile(null, $info['path'], null);
-				}
-			}
-			else {
-				$this->statics[$class] = false;
-			}
-		}
-
-		if (isset($this->statics[$class][$name])) {
-			$static = $this->statics[$class][$name];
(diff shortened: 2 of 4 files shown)


tractorcow:
Look at all that no code that's no longer there. :)

micmania1:
Tests failed on PHP 5.3 - damn syntaxes.

micmania1:
Note: If merged then some of this stuff will need to have a deprecation warning in future ss 3 releases.

model/DataList.php
```diff
@@ -51,6 +51,11 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 	 * @param string $dataClass - The DataObject class to query.
 	 */
 	public function __construct($dataClass) {
+		// Make sure we're getting the correct class name - this may be case-sensitive.
+		if(class_exists($dataClass)) {
+			$reflection = new ReflectionClass($dataClass);
+			$dataClass = $reflection->name;
+		}
```
kinglozzer: Why has this been added? Did the old manifest fix case sensitivity issues that the new version doesn’t?

micmania1: I'll get back to you XD (There is a reason but i'm not sure this is the right fix).

micmania1:
I found a bug in submitting this pull request. DataObject::has_own_table() doesn't work with class names which don't exactly match the class name (ie. they're case-sensitive). This is because the "class name" in Config::inst()->get("ClassName', 'blah') is case-sensitive.

The following returns null for incorrectly cased class names:
http://api.silverstripe.org/3.1/source-class-DataObject.html#2483

The thing that highlighted this was the failure of DataObjectTest::testGetCaseInsensitive().

	public function testGetCaseInsensitive() {
		// Test get_one() with bad case on the classname
		// Note: This will succeed only if the underlying DB server supports case-insensitive
		// table names (e.g. such as MySQL, but not SQLite3)
		if(!(DB::get_conn() instanceof MySQLDatabase)) {
			$this->markTestSkipped('MySQL only');
		}

		$subteam1 = DataObject::get_one('dataobjecttest_subteam', array(
			'"DataObjectTest_Team"."Title"' => 'Subteam 1'
		), true);
		$this->assertNotEmpty($subteam1);
		$this->assertEquals($subteam1->Title, "Subteam 1");
	}

This tries to prove that using a lowercase class name won't affect whether or not the result will be returned. The should have been failing, however reason this has passed up until this point is because the query is built incorrectly. 

DataObject::has_own_table()'s failure has a knock-on affect which prevents ClassInfo::ancestry($class, $tableOnly = true) working correctly. When only looking up class ancestors with tables, the lower-case table name will be ignored (DataObject::has_own_table() == false) and no SQL will be added to fetch data from that table.

The given test has passed like so up until now:

 	class MyBaseClass extends DataObejct {
 		private static $db = array('SomeField' => 'Varchar');
 	}

 	class MySubclass extends MyBaseClass {
 		private static $db = array('SomeOtherField' => 'Varchar');
 	}

What happens when you query the subclass is that it gets completely ignored and only the base class is queried:
 	// Tries to detect whether or not 'mysubclass' has its own table. Doesn't find it so excludes from query.
 	// SELECT MyBaseClass.SomeField FROM MyBaseClass
 	DataObject::get_one('mysubclass');


 	// What SHOULD happen
 	// SELECT MyBaseClass.SomeField, MySubclass.SomeOtherField FROM MyBaseClass.SomeField LEFT JOIN MyBaseClass ON (MyBaseClass.ID ON MySubclass.ID)
 	DataObject::get_one('mysubclass');

The fix i've implemented is in DataObject::has_own_table() and in ClassInfo::ancestry() so that only correct case-sensitive class names are used. This fixes the issue because the config system is case-sensitive when looking up class names, which is the root cause of this bug.

If you need me to split this out into 2 commits let me know.

micmania1:
I see case-insensitivity has been suggested in #3381 - it would be good as long as performance doesn't take a hit.

tractorcow:
@micmania1 you probably need to refactor that classname normalisation code you added into ClassInfo into a new method, so that other parts of the codebase are able to access it. Normalising the table name from a class is one such case.

micmania1:
I'd like to wait and see a response in #3381 before this is merged given that this is easier to make case-insensitive.

@tractorcow I do agree, but I plan to tackle ClassInfo at some point so I'll give some proper thought to where that should belong then.

nfauchelle:
@micmania1 
@micmania1 
Really like this commit. I've been using now on two websites (locally) for a couple of days now, and thought I would share some benchmarks.

PHP 5.5.18. No opcode caching. There is numerous things running on my laptop so not best conditions, but still....


_Website A_
SS 3.1.13
betterbuttons, gridfieldextensions
mysite: 8 Pages, 2 DataObjects, 2 DataExtensions.

Accessing / after 25 runs.

Times:
Before: Min: 0.2429 Max: 0.2550 Avg: 0.2474 Median: 0.245
After:    Min: 0.2369 Max: 0.2653 Avg: 0.2539 Median: 0.25435

Memory:
Before:	Min: 25,690,112 Max: 25,690,112 Avg: 25,690,112.00 Median: 25,690,112 
After:	Min: 24,903,680 Max: 24,903,680 Avg: 24,903,680.00 Median: 24,903,680 

Performing a flush on the command line 5 times.

Time:
Before:	Min: 3.0450 Max: 3.0890 Avg: 3.0648 Median: 3.05965
After:	Min: 0.7980 Max: 0.8431 Avg: 0.8171 Median: 0.8107

Memory:
Before:	Min: 19,136,512 Max: 19,398,656 Avg: 19,346,227.20 Median: 19,398,656 
After:	Min: 17,825,792 Max: 17,825,792 Avg: 17,825,792.00 Median: 17,825,792 


_Website B:_
SS 3.1.12
betterbuttons, gridfieldextensions, silverstripe-catalogmanager, silverstripe-lumberjack, silverstripe-menumanager
mysite: 9 Pages, 21 DataObjects, 4 Model Admin, 5 DataExtensions

Accessing / after 25 runs.

Times:
Before:	Min: 0.4160 Max: 0.5906 Avg: 0.4364 Median: 0.42385
After:	Min: 0.4101 Max: 0.5017 Avg: 0.4343 Median: 0.42735

Memory:
Before:	Min: 27,525,120 Max: 27,787,264 Avg: 27,588,034.56 Median: 27,525,120
After:	Min: 26,476,544 Max: 26,476,544 Avg: 26,476,544.00 Median: 26,476,544 


Performing a flush on the command line 5 times.

Time:
Before:	Min: 6.0492 Max: 6.2917 Avg: 6.1396 Median: 6.09
After:	Min: 3.6682 Max: 3.9522 Avg: 3.7599 Median: 3.7007

Memory:
Before: Min: 28,049,408 Max: 28,049,408 Avg: 28,049,408.00 Median: 28,049,408 
After:	Min: 26,738,688 Max: 27,000,832 Avg: 26,948,403.20 Median: 27,000,832 

--

So we can see a good speed up there with the flush! About 2.3~ seconds faster now. On the Website A there is a memory saving of 600k~ on general requests, on the larger site we are just over 1Mb.

There seems to be no real speed change in general use of the site.

Developing Website A with this patch has been great! It's so fast to flush and in many cases don't even need to.

dhensby:
So, do we want to move this along? How about we rebase, please :)

nfauchelle:
Yes please :) Would love to see it merged.

micmania1:
Still waiting on #3381 @sminnee @hafriedlander

tractorcow:
We've already addressed case-sensitivity issues in https://github.com/silverstripe/silverstripe-framework/pull/3949. Do you still need to wait for #3381?

micmania1:
Rebased!!!

nfauchelle:
Woohoo!

kinglozzer:
dance

Issue #3621, opened 2014-11-10
Not possible to filter on DB columns that are explicitly null


dhensby:
If the database column has a `NULL` value stored, it is not currently possible to filter a DataList to pick out null values.

`Page::get()->filter('NullableField', null)` produces `"NullableField" = ''` which doesn't catch null values.

tractorcow:
Yeah but filters don't currently allow for parameter-less filters.

micmania1:
It wouldn't be parameter-less. You'd be defining is null or is not null (boolean). See my example further up.

tractorcow:
Oh, true. That would work.

assertchris:
Opted for @tractorcow's suggestion of a value-type condition on `null`. In other words: if the value is null (on `applyOne` or `excludeOne`) then `IS NULL` or `IS NOT NULL` are applied instead of string handling.

tractorcow:
PR is at https://github.com/silverstripe/silverstripe-framework/pull/4196, and marked against 4.0 since it's potentially breaking API.

kinglozzer:
Closed via 392536cb1738f811b9716428638fb7cb518bf7a5

jonom:
For devs looking for a solution for 3.x you probably need to bypass the filter method and do something like this:

`Page::get()->where("\"NullableField\" IS NULL OR \"NullableField\" = ''")`

jonom:
Should we add something about this limitation to the 3.x docs? No mention of it currently https://docs.silverstripe.org/en/3/developer_guides/model/data_model_and_orm/#filtering-results

PR #4152, opened 2015-05-03
[Feedback, Sweet Feedback, Please] DB filter by NULL

assertchris:
This adds the ability to filter fields by `IS NULL` / `IS NOT NULL`. 

Further discussion at: https://github.com/silverstripe/silverstripe-framework/issues/3621

### DIFF
--- a/search/filters/ExactMatchFilter.php
+++ b/search/filters/ExactMatchFilter.php
@@ -1,20 +1,18 @@
 <?php
-/**
- * @package framework
- * @subpackage search
- */
 
 /**
- * Selects textual content with an exact match between columnname and keyword.
+ * Selects textual content with an exact match between column name and keyword.
  *
  * @todo case sensitivity switch
  * @todo documentation
- * 
+ *
  * @package framework
  * @subpackage search
  */
 class ExactMatchFilter extends SearchFilter {
-
+	/**
+	 * @param array $modifiers
+	 */
 	public function setModifiers(array $modifiers) {
 		if(($extras = array_diff($modifiers, array('not', 'nocase', 'case'))) != array()) {
 			throw new InvalidArgumentException(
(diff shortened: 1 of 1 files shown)


tractorcow:
I don't think we can introduce a new filter behaviour in 3.1, especially one that's not backwards compatible (null could have been used in the past to filter for empty strings).

Perhaps you should look at 3.2 for this fix, since that's the original milestone for #3621?

If you move it to 3.2 it still doesn't change the problem that you're altering current behaviour instead of simply adding a new feature.

What you could do is add another modifier to turn the new behaviour on.

`->filter('Field', null)` // if the field is empty (current behaviour)
`->filter('Field:not', null)` // if field isn't empty string (current behaviour)
`->filter('Field:nullable', null)` // If field is null
`->filter('Field:nullable:not', null)` // if the field is not null

I would argue that in 4.0 it would be fine to drop the `nullable` filter, but for anything 3.x it would be necessary to turn on the new function.

The other (and perhaps better) solution is to simply fix this in 4.0 and not touch 3.x at all?

assertchris:
All good ideas. Whaddya think, @sminnee?

tractorcow:
Have you tested that `filter('Field',  array(null, '', 'value')` correctly evaluates to `"Field" IS NULL OR "Field" in ('', 'value')` or `"Field" IS NULL OR Field = '' OR FIELD = 'value'` depending on the :case :nocase modifiers?

tractorcow:
I can help you with unit tests too later. :)

assertchris:
@tractorcow as I mentioned in the issue (linked above) this filter is only applied to `applyOne` and `excludeOne`. This is an exploratory PR, not a complete solution. We haven't even decided if it's supposed to go in 3.2 or 4.0 yet. :)

tractorcow:
Ah, well we probably shouldn't raise PR then at this stage. :) We wouldn't want someone to come along and merge this if it's not ready? I'll close this for now then, and we can continue to discuss here.

dhensby:
Did we get IS NULL support in?

tractorcow:
Yes, at https://github.com/silverstripe/silverstripe-framework/pull/4196.

Issue #3915, opened 2015-02-17
ClassInfo::implementorsOf() doesn't return all classes that implement an interface


dhensby:
Classes that descend from other classes that implement an interface don't get returned as "implementors of".

eg:

``` php
class MyClass implements HiddenClass {
}

class MyChildClass extends MyClass {
}

ClassInfo::implementorsOf('HiddenClass'); //just returns MyClass
```

sminnee:
In the places that it is used, the current behaviour is probably more appropriate.

For example, if we define`LeftAndMain::providePermissions()` we don't want it calling the identical method on every subclass.

I would suggest that we add a flag to allow for Dan's suggested behaviour rather than just changing it.

As such, marking as enhancement.

sminnee:
#wontfix

michalkleiner:
There's a way how to check if the class has any subclasses and only call the method on the youngest child.
I had to do just that as I don't want to trigger the method on all implementors.

However, I just found that the method doesn't return any classes that implement an interface IntB when looking for implementors of IntA (and IntB extends IntA).

PR #4153, opened 2015-05-04
API Enable tree filter highlighting

tractorcow:
Decoupling of CMS / Framework

This feature allows pages matching a custom search to be highlighted in the left hand tree view. This is handy for distinguishing between pages which actually match your search, and those which are simply displayed in order to hold child results.

The class 'CMSSiteTreeFilter' has been refactored out of the core beacuse it's a CMS class, and instead a framework interface 'LeftAndMain_SearchFilter' is used as an abstract api for a search-filterable class.

See https://github.com/silverstripe/silverstripe-cms/pull/1200 for CMS component

### DIFF
--- a/admin/code/LeftAndMain.php
+++ b/admin/code/LeftAndMain.php
@@ -773,6 +773,29 @@ class LeftAndMain extends Controller implements PermissionProvider {
 	}
 
 	/**
+	 * Gets the current search filter for this request, if available
+	 *
+	 * @throws InvalidArgumentException
+	 * @return LeftAndMain_SearchFilter
+	 */
+	protected function getSearchFilter() {
+		// Check for given FilterClass
+		$params = $this->getRequest()->getVar('q');
+		if(empty($params['FilterClass'])) {
+			return null;
+		}
+
+		// Validate classname
+		$filterClass = $params['FilterClass'];
+		$filterInfo = new ReflectionClass($filterClass);
+		if(!$filterInfo->implementsInterface('LeftAndMain_SearchFilter')) {
+			throw new InvalidArgumentException(sprintf('Invalid filter class passed: %s', $filterClass));
+		}
+		
+		return Injector::inst()->createWithArgs($filterClass, array($params));
+	}
+
+	/**
 	 * Get a site tree HTML listing which displays the nodes under the given criteria.
 	 *
 	 * @param $className The class of the root object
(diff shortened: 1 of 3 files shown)


admin/code/LeftAndMain.php
```diff
@@ -773,6 +773,29 @@ public function SiteTreeAsUL() {
...
+		// Check for given FilterClass
+		$params = $this->getRequest()->getVar('q');
+		if(empty($params['FilterClass'])) {
+			return null;
+		}
+
+		// Validate classname
+		$filterClass = $params['FilterClass'];
+		$filterInfo = new ReflectionClass($filterClass);
+		if(!$filterInfo->implementsInterface('LeftAndMain_SearchFilter')) {
```
assertchris: `is_subclass_of($params['FilterClass'], 'LeftAndMain_SearchFilter')` might be more efficient.

admin/code/LeftAndMain.php
```diff
@@ -1956,3 +2013,42 @@ public function setIsCurrent($bool) {
...
+	public function isPageIncluded($page);
+
+	/**
+	 * Given a page, determine any additional CSS classes to apply to the tree node
+	 *
+	 * @param DataObject $page
+	 * @return array|string
+	 */
+	public function getPageClasses($page);
+}
```
assertchris: GitHub cries whenever it sees `⊘⏎`...

admin/code/LeftAndMain.php
```diff
@@ -1915,16 +1954,34 @@ public function forTemplate() {
...
+	 *
+	 * @return string
+	 */
 	public function getClasses() {
+		// Get classes from object
 		$classes = $this->obj->CMSTreeClasses($this->numChildrenMethod);
-		if($this->isCurrent) $classes .= " current";
-		$flags = $this->obj->hasMethod('getStatusFlags') ? $this->obj->getStatusFlags() : false;
+		if($this->isCurrent) {
+			$classes .= " current";
```
assertchris: Consistent quotes maybe? No biggie...

admin/code/LeftAndMain.php
tractorcow: I can't since that method only works with classes. LeftAndMain_SearchFilter is an interface.

assertchris: screen shot 2015-05-04 at 5 06 09 pm

tractorcow:
Updated based on your feedback @assertchris

admin/code/LeftAndMain.php
assertchris: ...since 5.3.7

tractorcow: We need to support 5.3.3 though =(

kinglozzer: How about...
```php
ClassInfo::classImplements($params['FilterClass'], 'LeftAndMain_SearchFilter')
```
Hopefully that’s more efficient than reflection :)

assertchris: My money's on ReflectionClass being more efficient than the manifests... :)

kinglozzer: Actually I think the manifests might win here (for once) - all the data sitting behind this method is already loaded into memory from a cache as part of our “bootstrap” process!

tractorcow: That doesn't work @kinglozzer ... that only works if the interface is DIRECTLY on the class, not on a parent class. It doesn't respect inheritance. =( I'd have to get implementors then run is_subclass_of on each item.

kinglozzer: doh...

assertchris: I would consider that a bug in `ClassInfo::classImplements`...

tractorcow:
Ta!

admin/code/LeftAndMain.php
dhensby: i've already reported this https://github.com/silverstripe/silverstripe-framework/issues/3915

PR #4154, opened 2015-05-04
API Use mysql buffered statements

### LINKED ISSUE #3426: MySQLiConnector: mysqli_stmt::get_result is undefined if not using mysqlnd driver

tractorcow:
Avoids the usage of any MySQL Native Driver specific API

Alternative solution to https://github.com/silverstripe/silverstripe-framework/pull/4143

Fixes #3426

The difference in this solution is that we do NOT attempt to leave buffering until as late as possible (if necessary) and instead we buffer all statements up front. This has a slight performance cost, but in other cases could result in less memory usage.

When running unit tests I get max memory usage 192.1M.

### DIFF
--- a/model/connect/MySQLStatement.php
+++ b/model/connect/MySQLStatement.php
@@ -1,0 +1,116 @@
+<?php
+
+/**
+ * Provides a record-view for mysqli statements
+ *
+ * By default streams unbuffered data, but seek(), rewind(), or numRecords() will force the statement to
+ * buffer itself and sacrifice any potential performance benefit.
+ */
+class MySQLStatement extends SS_Query {
+
+	/**
+	 * The related mysqli statement object if generated using a prepared query
+	 *
+	 * @var mysqli_stmt
+	 */
+	protected $statement;
+
+	/**
+	 * Metadata result for this statement
+	 *
+	 * @var mysqli_result
+	 */
+	protected $metadata;
+
+	/**
+	 * Is the statement bound to the current resultset?
+	 *
+	 * @var bool
+	 */
+	protected $bound = false;
+
+	/**
+	 * List of column names
+	 *
+	 * @var array
+	 */
+	protected $columns = array();
+
+	/**
+	 * List of bound variables in the current row
+	 *
+	 * @var array
+	 */
+	protected $boundValues = array();
+
+	/**
+	 * Binds this statement to the variables
+	 */
+	protected function bind() {
+		$variables = array();
+
+		// Bind each field
+		while($field = $this->metadata->fetch_field()) {
+			$this->columns[] = $field->name;
+			// Note that while boundValues isn't initialised at this point,
+			// later calls to $this->statement->fetch() Will populate
+			// $this->boundValues later with the next result.
+			$variables[] = &$this->boundValues[$field->name];
+		}
+
+		call_user_func_array(array($this->statement, 'bind_result'), $variables);
+		$this->bound = true;
+		$this->metadata->free();
+
+		// Buffer all results
+		$this->statement->store_result();
+	}
+
+	/**
+	 * Hook the result-set given into a Query class, suitable for use by SilverStripe.
+	 * @param mysqli_stmt $statement The related statement, if present
+	 * @param mysqli_result $metadata The metadata for this statement
+	 */
+	public function __construct($statement, $metadata) {
+		$this->statement = $statement;
+		$this->metadata = $metadata;
+
+		// Immediately bind and buffer
+		$this->bind();
+	}
+
+	public function __destruct() {
+		$this->statement->close();
+		$this->closed = true;
+		$this->currentRecord = false;
+	}
+
+	public function seek($row) {
+		$this->rowNum = $row - 1;
+		$this->statement->data_seek($row);
+		return $this->next();
+	}
+
+	public function numRecords() {
+		return $this->statement->num_rows();
+	}
+
+	public function nextRecord() {
+		// Skip data if out of data
+		if (!$this->statement->fetch()) {
+			return false;
+		}
+
+		// Dereferenced row
+		$row = array();
+		foreach($this->boundValues as $key => $value) {
+			$row[$key] = $value;
+		}
+		return $row;
+	}
+
+	public function rewind() {
+		return $this->seek(0);
+	}
+
+}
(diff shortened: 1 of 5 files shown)


tractorcow:
I think we should simply choose the most stable solution, and propose that people use PDOConnector as the recommended connector going forward.

tractorcow:
As per the other pull request, I have removed the old and skipped tests for MultiEnum.

tractorcow:
This needs some serious review. Can @silverstripe/core-team please look at this, as this is the last critical fix before we can release 3.2?

We need to choose one of the two solutions, or either come up with a third.

model/connect/MySQLiConnector.php
```diff
@@ -235,18 +238,16 @@ public function preparedQuery($sql, $parameters, $errorLevel = E_USER_ERROR) {
...
 			$values = $this->parameterValues($parameters);
 			$this->databaseError($this->getLastError(), $errorLevel, $sql, $values);
 			return null;
 		}
 
-		// May not return result for non-select statements
-		if($result = $lastStatement->get_result()) {
-			return new MySQLQuery($this, $result, $lastStatement);
+		// Non-select queries will have no result data
+		if($lastStatement && ($metaData = $lastStatement->result_metadata())) {
```
assertchris: If a select query isn't a prepared statement, will `result_metadata()` return a value? If not, are we breaking execution for those queries?

model/connect/MySQLStatement.php
```diff
@@ -0,0 +1,113 @@
+<?php
+
+/**
+ * Provides a record-view for mysqli statements
+ *
+ * By default streams unbuffered data, but seek(), rewind(), or numRecords() will force the statement to
+ * buffer itself and sacrifice any potential performance benefit.
+ */
+class MySQLStatement extends SS_Query {
```
assertchris: Maybe `MySQLPreparedStatement` is a more explicit name?

model/connect/MySQLStatement.php
```diff
@@ -0,0 +1,113 @@
...
+	 * @var array
+	 */
+	protected $columns = array();
+
+	/**
+	 * List of bound variables in the current row
+	 *
+	 * @var array
+	 */
+	protected $boundValues = array();
```
assertchris: Where is this initialised?

model/connect/MySQLiConnector.php
tractorcow: No it won't. Only "select" statements return results_metadata. The metadata is the column details of the results. Essentially the metadata of this value is a zero-length resultset containing the same results contained by the statement. It's reasonably safe to assume there is a 1-to-1 correlation between results_metadata and the statement containing results.

model/connect/MySQLStatement.php
tractorcow: All statements are prepared statements, so I think it speaks for itself. ;)

model/connect/MySQLStatement.php
tractorcow: check bind()

model/connect/MySQLStatement.php
```diff
@@ -0,0 +1,113 @@
...
+	/**
+	 * Binds this statement to the variables
+	 */
+	protected function bind() {
+		$variables = array();
+
+		// Bind each field
+		while($field = $this->metadata->fetch_field()) {
+			$this->columns[] = $field->name;
+			$variables[] = &$this->boundValues[$field->name];
```
tractorcow: This part looks weird because at this point, boundValues is empty. By using the reference operator, it doesn't need to be isset, since it simply allows the statement to write to this bound value on `$statement->fetch()`. It's weird code because mysqli relies on results being bound by reference to a statement.

model/connect/MySQLStatement.php
sminnee: All good on the classname but what about the inheritance?

model/connect/MySQLStatement.php
sminnee: Right, now I get it.

Can you please explicitly comment this, then? Eg: `$variables will be populated by `$statement->fetch(), which will in turn populate $this->boundValues for us`

model/connect/MySQLiConnector.php
sminnee: So, are you saying that you use MySQL prepared statements even when there are no variables to pass in? 

What happens if the query doesn't throw an error but also doesn't provide meta-data? Maybe just throw an exception if that happens?

sminnee:
OK I've reviewed this change and I think it's better than #4143

Before merging I suggested an extra comment and an exception above; I think we should also bundle in a patch to the upgrade guide.

model/connect/MySQLiConnector.php
sminnee: Discussed with Damian; we're going to return a stub MySQLQuery just like 3.1 does.

model/connect/MySQLStatement.php
sminnee: Discussed with Damian; changing inheritance will muck up the constructor call signatures. The public 3.1 API where this can be accessed is SQLQuery::execute(), and this is documented as returning an SS_Query, so we think that returning a different subclass of SS_Query is fine.

sminnee:
Note: This change does increase memory usage in situations where you query a very large result set, iterate on the first few items, and then discard the remaining result set. It will use mysql-driver rather than php memory, but memory nonetheless.

First up, if you're doing this with a result set large enough to blow through all your memory, your code should be refactored to use a limit query.

If, however, you are trying to upgrade a large legacy codebase, then your options are to switch to one of the other database drivers:

 * mysqlnd
 * PDO

tractorcow:
Ok I've updated this PR to include the following:

* Both ::query and ::preparedQuery will now always return a SS_Query object, to be in line with 3.1 behaviour. For non-select queries the actual result won't be iterable.
* Revert the constructor of MySQLQuery to include the $database parameter as per the old api in 3.1, although this parameter is ignored.
* Unit tests for affectedRows

tractorcow:
Test cases pass :D

sminnee:
Merging!

tractorcow:
Yay thanks!

PR #4155, opened 2015-05-04
FIX: File::getFileType() was case sensitive (fixes #3631)

### LINKED ISSUE #3631: File edit form “file type” shows “unknown” for uppercase file extensions
E.g. rename a file to have an uppercase extension: `some-image.JPG`. The edit form (both in `UploadField` and in `AssetAdmin` if the CMS module is installed) show the file type as “unknown”.

Not really sure where we should fix this: `getFileType()` or `getExtension()`. `File::validate()` manually transforms `getExtension()` to lower case, so I assume `getExtension()` is supposed to return whatever case is on the disk - therefore I’m going to guess at `getFileType()`.

kinglozzer:


kinglozzer:
bump!

kinglozzer:
:kissing:

PR #4156, opened 2015-05-04
Remove case sensitivity from sentence end

cmcramer:
Removed case sensitivity from FirstSentence() and Summary() end of
sentence exceptions.

Currently sentences aren't recognized if the last word ends in _mrs
_ms, etc. Removed case insensitive flag and specified cases to skip.
Removed "Miss" as this is not an abbreviation and is properly spelled
without a period.

### DIFF
--- a/model/fieldtypes/HTMLText.php
+++ b/model/fieldtypes/HTMLText.php
@@ -141,7 +141,7 @@ class HTMLText extends Text {
 		/* Otherwise work backwards for a looking for a sentence ending (we try to avoid abbreviations, but aren't
 		 * very good at it) */
 		for ($i = $maxWords; $i >= $maxWords - $flex && $i >= 0; $i--) {
-			if (preg_match('/\.$/', $words[$i]) && !preg_match('/(Dr|Mr|Mrs|Ms|Miss|Sr|Jr|No)\.$/i', $words[$i])) {
+			if (parent::is_sentence_end($words[$i])) {
 				return implode(' ', array_slice($words, 0, $i+1));
 			}
 		}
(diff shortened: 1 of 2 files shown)


tractorcow:
Couldn't we solve this issue by using the match start of string '^' instead, or even a word bound '\b'? If we are already splitting on word why would we not?

cmcramer:
Probably good to add ^, but I don't see any advantage to using a case insensitive match, particularly with "No" in the list of exceptions.

tractorcow:
I think if we wanted to change case sensitivity (technically an api change) it would need to be in 3.2 at the earliest. 3.1 can accept bugfixes though, given there is no change in behaviour, as we are following [semantic versioning](http://semver.org/) in core.

Removing "miss" is fine since it's not ment to be followed by a period, and fixing the "ending with" would be an appropriate fix.

cmcramer:
That just leaves the problem with no. Do you want me to close this pull request and resubmit using your recommendations instead of removing "i" flag?

tractorcow:
What er, does the "no" mean (in this context). Number? I'm not sure it's going to be appropriate to assume it's such, since it doesn't seem to match the general context.

I'd probably flag it as a bug, along with "miss" :P

tractorcow:
Also, no need to close this one; Just update it (squash and force push it).

Do you think it's worth putting these regexps in one place? If it's a protected method on Text we can argue it's not a new API, as it's not public (and can go into 3.1).

cmcramer:
"No" is in the exception list because it is representation of the Numero symbol http://en.wikipedia.org/wiki/Numero_sign. One could argue that it is more common to see "no. 1" mid-sentence than "no" as the last word in a sentence.

I agree it would be beter to pull the regexps out of the three functions and call it from a one place. That would also make it easier to make FirstSentence non-English specific in the future.

dhensby:
Bump

tractorcow:
It looks like you meant to remove case insensitivity, but you left it as case insensitive? I think this PR title should have been "add case sensitivity" not "remove" right? 

This is an old pull request, though. Should we just close it? 4.0 code has replaced all of this in any case.

dhensby:
I mean - if this problem is now fixed... we should probably close it. it shouldn't be going into 3.1

sminnee:
Agree; closing.

PR #4157, opened 2015-05-04
NEW: Allow 'null' limit for database queries (closes #3487)

### LINKED ISSUE #3487: $dataList->limit(null, $offset); doesn’t work
```php
$list = SomeDataObject::get(); // Let's say 5 items are in the list
$list = $list->limit(null, 3); // Should return 2 items, actually returns 5.
```

We essentially need to uncomment [this assertion](https://github.com/silverstripe/silverstripe-framework/blob/master/tests/model/DataListTest.php#L120-L122) and make it pass.

kinglozzer:
- [SQLite3 pull request](https://github.com/silverstripe-labs/silverstripe-sqlite3/pull/8)
- [PostgreSQL pull request](https://github.com/silverstripe/silverstripe-postgresql/pull/40)

I spent ages setting up SQL Server in a virtual machine for testing, only to realise that the MSSQL module [doesn’t need any changes](https://github.com/silverstripe/silverstripe-mssql/blob/758fe810a58616da9517913ec3778ab831e54b9c/code/MSSQLQueryBuilder.php#L59)!

### DIFF
--- a/_config/database.yml
+++ b/_config/database.yml
@@ -7,10 +7,10 @@ Injector:
     properties:
       connector: %$PDOConnector
       schemaManager: %$MySQLSchemaManager
-      queryBuilder: %$DBQueryBuilder
+      queryBuilder: %$MySQLQueryBuilder
   MySQLDatabase:
     class: 'MySQLDatabase'
     properties:
       connector: %$MySQLiConnector
       schemaManager: %$MySQLSchemaManager
-      queryBuilder: %$DBQueryBuilder
+      queryBuilder: %$MySQLQueryBuilder
--- a/model/connect/MySQLQueryBuilder.php
+++ b/model/connect/MySQLQueryBuilder.php
@@ -1,0 +1,56 @@
+<?php
+
+/**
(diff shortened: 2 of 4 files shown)


tractorcow:
I tried to check what the Ansi SQL standard was, but it doesn't seem to even use 'limit', so I guess we are stuck with adaptor specific query generators. Good job @kinglozzer . I'll merge the other two PR then restart travis on framework before merging this. Sound good?

tractorcow:
@kinglozzer can you please check the test failures?

tractorcow:
I merged your sqlite fix. Let's hope this resolves any remaining issue.

kinglozzer:
SQLite3 was a screw up (sorry!), the PostgreSQL failure was partly bad test design: with no explicit sort specified, the order of results isn’t guaranteed - so the builds were only passing locally by chance (adding `->sort('ID ASC')` would make them pass). I’ve changed those assertions now, so we should see green ticks soon! :pray:

kinglozzer:
:D thanks @tractorcow

PR #4158, opened 2015-05-05
Change cursor for drag + drop icon on hover/active

scott1702:


jonom:
This is supported by Firefox with the -moz prefix too so should probably include that and the unprefixed version too for longevity. Does this fall back to `cursor:move` otherwise? http://caniuse.com/#search=cursors

scott1702:
Yep good points, updated

PR #4162, opened 2015-05-06
FIX: Object::parse_class_spec failed to parse associative arrays

kinglozzer:
Submitted to 3.1 as this enables a documented ability of the framework - [configuring database field options](https://github.com/silverstripe/silverstripe-framework/blob/d5361dedecee991e37b6563918de0aa32c6d0f21/model/fieldtypes/StringField.php#L31). I.e:

```php
private static $db = array(
	'MyField' => 'Varchar(255, array("nullifyEmpty" => false))'
);
```

[Alternative solution](https://github.com/silverstripe/silverstripe-framework/pull/4091#issuecomment-99526777) using `eval()`.

### DIFF
--- a/core/Object.php
+++ b/core/Object.php
@@ -195,77 +195,113 @@ abstract class Object {
 		$tokens = token_get_all("<?php $classSpec");
 		$class = null;
 		$args = array();
-		$passedBracket = false;
 		
 		// Keep track of the current bucket that we're putting data into
 		$bucket = &$args;
 		$bucketStack = array();
-		$had_ns = false;
+		$hadNamespace = false;
+		$currentKey = null;
 		
 		foreach($tokens as $token) {
-			$tName = is_array($token) ? $token[0] : $token;
-			// Get the class naem
-			if($class == null && is_array($token) && $token[0] == T_STRING) {
+			// $forceResult used to allow null result to be detected
+			$result = $forceResult = null;
+			$tokenName = is_array($token) ? $token[0] : $token;
+
+			// Get the class name
+			if($class === null && is_array($token) && $token[0] === T_STRING) {
 				$class = $token[1];
-			} elseif(is_array($token) && $token[0] == T_NS_SEPARATOR) {
+			} elseif(is_array($token) && $token[0] === T_NS_SEPARATOR) {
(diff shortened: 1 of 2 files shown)


kinglozzer:
*Bump!*

Happy to resubmit this against another branch if it’s decided that it’d be more appropriate.

dhensby:
Well, the tests pass but I don't think this can go in 3.1, really. 

How about using PhpParser ;)

Isn't `eval` what we normally use for this kind of thing? 

@tractorcow, thoughts?

kinglozzer:
> How about using PhpParser ;)

Not sure how it’ll cope as it’s not valid PHP. I guess we could prefix the string with `new` and it’d be valid, though...

> Isn't `eval` what we normally use for this kind of thing?

@dhensby https://github.com/kinglozzer/sapphire/commit/af18c0b99f9b408fd8e42672b92878603fb9312a ;)

dhensby:
The rival code looks a lot more straight forward and maintainable, but it does have the pariah which is eval... 

Which is your preferred?

kinglozzer:
> Which is your preferred?

I have no strong preference, but I’d be anxious about making assumptions about how people use that API (at least in a non-major version): what if someone, somewhere is doing the unthinkable and passing user-generated code into it?

I’d personally be fine with switching to using `eval()` in master, provided we clearly document it. I’d consider going one further too and renaming it to `eval_class_spec()` to be doubly-sure that nobody does something silly with it in future.

tractorcow:
If we're going to go with PhpParser anyway (see https://github.com/silverstripe/silverstripe-framework/pull/5567) then I'd probably recommend to use the same tool for this problem.

I'd be happy to ban `eval` from all lines of code in the framework. :D

dhensby:
PHPParser won't happen any time soon - no objections filed so this is going in

tractorcow:
yikes, ok, please merge up @dhensby . @sminnee and I were BOTH lamenting the lack of ability to do this in master. We need it for the default_cast work lol.

dhensby:
All merged up

tractorcow:
Thanks for this, I used this in my default_cast pr  #5653 and it was really helpful.

PR #4163, opened 2015-05-06
Change deprecation notice version for DataObject component methods

kinglozzer:
As per the discussion from the [core committers hangout](https://www.youtube.com/watch?v=d6ADget_PfY), I got this completely wrong in #3799! :)

My assumption was that we allow a grace period between the deprecation notice first showing and actually breaking backward compatibility: based on the assumption that devs *haven’t* changed their deprecation version, rather than assuming devs *have* preemptively changed their deprecation version. However, considering that we’re now using semantic versioning and devs should expect BC breaks in major versions, I’m happy with this way of doing things!

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -1576,7 +1576,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 		}
 
 		if($filter !== null || $sort !== null || $limit !== null) {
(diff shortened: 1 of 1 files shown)


tractorcow:
Thanks greatly @kinglozzer

dhensby:
:( makes me sad

tractorcow:
Yeah but it makes test cases happy. :)

PR #4164, opened 2015-05-06
FIX: Modifications to GridFieldExportButton to allow ArrayList use in SS_Report

wilr:
- canView is not defined, so if method doesn't exist assume that the user can view it (what GridField does).

- check destroy method exists

- Get value by source number returned null on arraylist, this will fetch by column name

### DIFF
--- a/forms/gridfield/GridFieldExportButton.php
+++ b/forms/gridfield/GridFieldExportButton.php
@@ -132,7 +132,7 @@ class GridFieldExportButton implements GridField_HTMLProvider, GridField_ActionP
 		}
 
 		foreach($items->limit(null) as $item) {
-			if($item->hasMethod('canView') && $item->canView()) {
+			if(!$item->hasMethod('canView') || $item->canView()) {
 				$columnData = array();
 
 				foreach($csvColumns as $columnSource => $columnHeader) {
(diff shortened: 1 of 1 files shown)


wilr:
@tractorcow Probably should go against the 3 branch rather than 3.1 but figured as a bug fix may get in. Can rebase it if it's not suitable.

forms/gridfield/GridFieldExportButton.php
```diff
@@ -132,7 +132,7 @@ public function generateExportFileData($gridField) {
 		}
 
 		foreach($items->limit(null) as $item) {
-			if($item->hasMethod('canView') && $item->canView()) {
+			if(!$item->hasMethod('canView') || $item->canView()) {
```
tractorcow: `$item->hasMethod('hasMethod')` ? :P

wilr: CanView?

tractorcow: I was making a useless joke sorry. It was a "coffee machine is broken" week.

tractorcow:
And I forgot to merge it as well. Yikes sorry @wilr

PR #4167, opened 2015-05-07
Fix CMSMenuItem CSS references for namespaced controllers

nedmas:
The controller class name including namespace is used to create CSS class names for `CMSMenuItem` (specifically for the icon).

Currently this is done without escaping the class name resulting in CSS classes with backslashes (`\`) which are invalid characters for CSS class names.

This PR simply uses the existing `Convert::raw2htmlname()` method to escape the class name before use in `CMSMenuItem`.

### DIFF
--- a/admin/code/CMSMenu.php
+++ b/admin/code/CMSMenu.php
@@ -142,7 +142,7 @@ class CMSMenu extends Object implements IteratorAggregate, i18nEntityProvider {
 			$cmsClasses = self::get_cms_classes();
 			foreach($cmsClasses as $cmsClass) {
 				$menuItem = self::menuitem_for_controller($cmsClass);
-				if($menuItem) $menuItems[$cmsClass] = $menuItem;
+				if($menuItem) $menuItems[Convert::raw2htmlname(str_replace('\\', '-', $cmsClass))] = $menuItem;
 			}
 		}
 		
--- a/admin/code/LeftAndMain.php
+++ b/admin/code/LeftAndMain.php
@@ -556,7 +556,7 @@ class LeftAndMain extends Controller implements PermissionProvider {
 	public static function menu_icon_for_class($class) {
 		$icon = Config::inst()->get($class, 'menu_icon', Config::FIRST_SET);
 		if (!empty($icon)) {
-			$class = strtolower($class);
+			$class = strtolower(Convert::raw2htmlname(str_replace('\\', '-', $class)));
 			return ".icon.icon-16.icon-{$class} { background: url('{$icon}'); } ";
 		}
 		return '';
--- a/admin/javascript/LeftAndMain.Menu.js
+++ b/admin/javascript/LeftAndMain.Menu.js
@@ -92,7 +92,7 @@
 			updateMenuFromResponse: function(xhr) {
 				var controller = xhr.getResponseHeader('X-Controller');
 				if(controller) {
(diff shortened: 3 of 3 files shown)


nedmas:
@tractorcow any chance this can be included in 3.1.14 please?

dhensby:
Please can you tell me what the before and after class names are?

nedmas:
Sure, so `Convert::raw2htmlname()` simply strips out the backslashes (`\`).

Example before:
`Example\Namespace\Controllers\ExampleAdmin`

Example after:
`ExampleNamespaceControllersExampleAdmin`

nedmas:
I've updated the PR to include the same stripping of backslashes (`\`) from the controller name when used to reference the ID and CSS class names when updating the menu. Aside from this I can find no other uses of the class name that need to be updated.

dhensby:
OK, I think that `\` need to be replaced with some character, otherwise a controller named `MyController` would conflict with a controller named `My\Controller`.

nedmas:
Agreed, although I think a conflict is unlikely we should code defensively. What character is best to use? I would suggest underscore (`_`).

dhensby:
I think underscores do make sense... but then PSR0 named classes could ALSO conflict.. gosh.

I reckon we go with an `_` and that'll be close enough

nedmas:
Good point, maybe we should use dash (`-`) instead? Might as well avoid all known conflicts. Can you think of any potential conflicts using dash (`-`)?

dhensby:
That seems sensible :) can't think of anything immediately

nedmas:
I've updated the PR to replace backslashes (`\`) with dashes (`-`). I've extracted it into a helper function called `Convert::fqcn2htmlname()` as it was needed in more than one class. The helper function replaces all backslashes with dashes then runs it through `Convert::raw2htmlname()` for good measure. I've added the same regex used in `Convert::raw2htmlname()` after replacing the backslashes with dashes in the javascript as well.

Is there anything else that I should include or anything I've missed?

dhensby:
If you're going to add new features then it can't make a patch release (so this PR will need to go against 3 (or 3.2)).

If you want it in a 3.1 release, keep this to the minimum required to fix the bug to get it into 3.1.x release

core/Convert.php
```diff
@@ -353,4 +355,19 @@ public static function raw2url($title) {
 	public static function nl2os($data, $nl = PHP_EOL) {
 		return preg_replace('~\R~u', $nl, $data);
 	}
+
+	/*
```
dhensby: PHP doc block should start `/**`

core/Convert.php
```diff
@@ -353,4 +355,19 @@ public static function raw2url($title) {
...
+	 *
+	 * Replaces backslashes (\) with dashes (-) then passes
+	 * the resulting string through Convert::raw2htmlname().
+	 *
+	 * @uses Convert::raw2htmlname()
+	 * @param string $val String to escape
+	 * @return string
+	 */
+	public function fqcn2htmlname($val)
+	{
```
dhensby: This is pedantic, but the coding conventions require the `{` to be on the same line as the `)`

nedmas:
Ok I'll change it to exclude new features and add them in against 3 or 3.2 once this is merged.

dhensby:
Thanks :)

nedmas:
Thanks

PR #4170, opened 2015-05-07
Store current page before marking tree nodes (fixes #4137, fixes silverstripe/silverstripe-cms#1135)

### LINKED ISSUE #4137: Tree view sometimes fails to load nodes, or sorts them incorrectly
I’m having a really difficult time tracking down the exact cause of this issue...

The site in question has ~100 pages, no modules that interfere with any of the tree view or hierarchy logic. The issues I’m experiencing are:

- If I hard-refresh while editing a page that’s 2 levels deep, it appears at the bottom of the list (i.e. last in the sort order, even though it definitely isn’t last).
- If I hard-refresh while editing a page that’s 3 levels deep, the tree view appears briefly, but is completely collapsed, and then it hides itself entirely - collapsing the very top-level “node” that is the site title.

Uncommenting [this line](https://github.com/silverstripe/silverstripe-framework/blob/d6e1c5131c299df549c3c407d9063c5faa7beb54/admin/code/LeftAndMain.php#L824) fixes both issues, I don’t know the `Hierarchy` class well enough to understand why this was commented out originally. According to @stojg in #3309:

> when it tries to find a deleted page, it flushes the DataObject cache and the that chains to hierarchy flush as well. This throws away all work done previously in markPartialTree

Difficult to reproduce. I managed to reproduce it on a blank install by adding around 50 pages (no particular structure, dotted around at different levels), then editing a page that’s 4-levels deep.

kinglozzer:
Well, this was an easier solution than I expected!

Code docs should explain this well enough, but basically:

- Without `$obj->markToExpose()`, the tree view fails to load with the current page highlighted under the conditions described in [this issue](https://github.com/silverstripe/silverstripe-cms/issues/1135). It then tries to load `getsubtree?ID=0&ajax=1` before collapsing _fully_ (all the way down to the root “your site name” node), making it very difficult to then find the current page;
- With `$obj->markToExpose()` uncommented where it was, [this issue](https://github.com/silverstripe/silverstripe-cms/issues/1049) re-occurs;
- By moving `$this->currentPage()` before `$obj->markToExpose()`, the deleted pages in `Hierarchy::$marked` aren’t wiped, fixing both issues

### DIFF
--- a/admin/code/LeftAndMain.php
+++ b/admin/code/LeftAndMain.php
@@ -811,6 +811,12 @@ class LeftAndMain extends Controller implements PermissionProvider {
 		// Get the tree root
 		$record = ($rootID) ? $this->getRecord($rootID) : null;
 		$obj = $record ? $record : singleton($className);
+
+		// Get the current page
+		// NOTE: This *must* be fetched before markPartialTree() is called, as this
+		// causes the Hierarchy::$marked cache to be flushed (@see CMSMain::getRecord)
+		// which means that deleted pages stored in the marked tree would be removed
+		$currentPage = $this->currentPage();
 		
 		// Mark the nodes of the tree to return
 		if ($filterFunction) $obj->setMarkingFilterFunction($filterFunction);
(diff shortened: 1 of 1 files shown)


kinglozzer:
@silverstripe/core-team Can anyone peer review this? It causes the issues in #4137 to occur on *all* sites with more than ~60 pages

tractorcow:
Tested and it seems ok. I can view children of deleted pages so the issue in the original solution didn't come back.

jonom:
Thanks Loz! :+1:

PR #4171, opened 2015-05-07
SQLQuery::Count errors with having clause

dhensby:
see #4147


dhensby:
thanks @arambalakjian / @tractorcow

PR #4172, opened 2015-05-08
Better Zend references

assertchris:
Good ol' Composer...

### DIFF
--- a/cache/Cache.php
+++ b/cache/Cache.php
@@ -183,8 +183,6 @@ class SS_Cache {
 		}
 
 		$frontendOptions = $frontendOptions ? array_merge($basicOptions, $frontendOptions) : $basicOptions;
-		
-		require_once 'Zend/Cache.php';
 
 		return Zend_Cache::factory(
 			$frontend, $backend[0], $frontendOptions, $backend[1]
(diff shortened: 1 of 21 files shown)


kinglozzer:
This will break things for people who install without using composer (i.e. zip archive), won’t it? Unfortunately we still need to support that :(

assertchris:
Ah, no ways. That's a good point and makes me sad.

chillu:
Shouldn't we just include `vendor/composer/autoload.php` stuff in the ZIP download by default?

kinglozzer:
@chillu I thought that was something that had only been proposed, but looking at the zip archive from http://www.silverstripe.org/software/download/ it’s already in place!

assertchris:
Hey, wait...Does that mean this change could actually happen?

kinglozzer:
@assertchris the problem is that it’s not currently _required_, so this would be a breaking change for anyone who didn’t upload `vendor/` to their server and dropped in the new `framework/` expecting it to work.

I’d be in favour of _requiring_ this (i.e. requiring people without composer to download from silverstripe.org) in future SS versions so we can take advantage of stuff like this, though that’s probably a dev-list discussion of its own!

assertchris:
Yip - perhaps a 4.0 thing, if errybody likes the idea of classmap over `require`...

tractorcow:
I don't think this is a breaking change, but it wouldn't hurt to do it in 4.0 anyway.

PR #4173, opened 2015-05-09
Update 06_Tabbed_Forms.md

AntonyThorpe:
Typo.  Adding a field to a tab, not from.  Thanks.


dhensby:
It's never too late to improve the docs... sorry for the delay :/

PR #4175, opened 2015-05-11
FIX: remove validation type constraint from form fields for 3.2 release

### LINKED ISSUE #4165: FormField validation BC-break (3.2)
https://github.com/silverstripe/silverstripe-framework/pull/3636 added `Validator` type hinting to a number of form field `validate()` methods which causes errors to crop up if any of the fields have been subclassed (i.e. `validate() should be compatible with FormField::validate(Validator $validator)`). Looks like this was done before 3 was branched off master.

stevie-mayhew:
Fixes #4165

Do we need to document that this should be re-added to master when merges are pulled back to that branch? Not quite sure where I should put that - perhaps just an issue against master?


kinglozzer:
Looks good, I’ll await @tractorcow’s input on the reminder to re-add them as it’ll likely be him that does the merge ;)

tractorcow:
I'm sure we won't forget :)

stevie-mayhew:
Never Forget

PR #4177, opened 2015-05-11
Adding dependency badge and reference badge

reiz:
Your dependencies are all up-to-date. And more than 1000 packages are referencing this package. Keep up the good work.


wilr:
Fixed in 747e87b

PR #4184, opened 2015-05-13
has_one cache

pitchandtone:
massively reduces DB calls if objects call common has_one

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -1417,9 +1417,9 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 		if($class = $this->has_one($componentName)) {
 			$joinField = $componentName . 'ID';
 			$joinID    = $this->getField($joinField);
-			
+
 			if($joinID) {
-				$component = $this->model->$class->byID($joinID);
+				$component = DataObject::get_one($class, 'ID = ' . $joinID);
 			}
 
 			if(!isset($component) || !$component) {


chillu:
That's an API change which we can't merge into 3.1 - introducing caches in such a central place of the framework can have all kinds of nasty side effects in application code. Even for 4.0, we have to be quite careful. I'd rather see this solved on a larger scale through an identity map across the ORM. Just one example off the top of my head:

```php
$obj = $hostObj->MyHasOne();
$hostObj->MyHasOneID = 99;
$obj = $hostObj->MyHasOne(); // oops, outdated data
```

pitchandtone:
In your example the code currently would go back to the DB and get it again. Whereas mine would have a reference to the object, so unless people are relying on outdated data....

I absolutely agree the ORM should have a proper persistence/indentity map built in, the performance benefit would not be small. 

I have an object which has 14 children who all rederence the parent in a array of ways. The queries on the page dropped from ~1000 to 92.

pitchandtone:
$member = $this->Member(); // aaa
$member->FirstName = 'bbb';
$w2 = Watchlist::get_by_id('Watchlist', $this->ID);
$member2 = $w2->Member();
return $member->FirstName . ' -- ' . $member2->FirstName;
// current code: bbb --- aaaa
// new code: bbb --- bbb

pitchandtone:
Also realise that this is very very similar to the get_one cache that's already in place, except it's done by ID, rather than filter.

chillu:
> In your example the code currently would go back to the DB and get it again. Whereas mine would have a reference to the object, so unless people are relying on outdated data

My example would implicitly rely on getting an *updated* (uncached) object, which your caching proposal would change: It wrongly returns you an old object, not the correct one with ID=99. In isolation, that's an unusual example, but I've wasted too many hours with obscure caching bugs to take such a core change lightly.

`DataObject::get_one()` is cached, but `DataList::get()->byId()` isn't. `hasMany()` and `manyMany()` aren't cached either. You could argue that only caching `hasOne()` makes this inconsistent. Caching the other relations gets more tricky, but that's not a good excuse to create this inconsistency.

I've had a look in other frameworks:
- Rails caches associations by default: http://guides.rubyonrails.org/association_basics.html#controlling-caching
- Rails 4.0 has removed its identity map for ActiveRecord associations due to inconsistencies: http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-3-2-to-rails-4-0-active-record
- Laravel doesn't cache (AFAICT): http://laravel.com/docs/5.0/eloquent#relationships

I can't follow your `$member` example above, sorry.

pitchandtone:
Actually the code example you've given is currently broken due to the component cache on the instance, I've resolved this in #4188. Your example won't even touch the new global cache, that will only be used if two different objects are asking the same object.

I've updated my example above to be more clear, let me know if this still doens't make sense.

I'm happy enough that SS uses an active record pattern rather than a data mapper pattern which is where identitiy maps normally reside.

I agree that creating an identity map which tracks list interaction across lists is difficult to imposssible to implement. Hence rails dropping it.

What I don't agree on is why get_one can have a cache, whereas has_one cannot. The question that comes out of that for me is why doesn't has_one use the get_one functionality and thereby leverage the get_one cache.

While your argument groups has_one with has_many and many_many, I think a more pragmatic grouping would be lists vs single objects.

Yes lists are hard. Loading single objects from the db is not, hence why a cache already exists in SS for that kind of procedure.

model/DataObject.php
```diff
@@ -1417,7 +1418,12 @@ public function getComponent($componentName) {
...
 			$joinID    = $this->getField($joinField);
-			
+
+			if(isset(DataObject::$_cache_has_one[$class][$joinID])) {
+				$this->components[$componentName] = DataObject::$_cache_has_one[$class][$joinID];
+				return $this->components[$componentName];
+			}
+
 			if($joinID) {
 				$component = $this->model->$class->byID($joinID);
```
tractorcow: We could use `DataObject::get_by_id($class, $joinID, true)` to use the caching in `DataObject::get_by_id`

I'm still a bit fuzzy on why we have DataModel. @chillu do you have a better idea?

chillu:
Sorry for the late response, and thanks for taking the time to elaborate. You said:

> While your argument groups has_one with has_many and many_many, I think a more pragmatic grouping would be lists vs single objects.

After thinking a bit more about the scope of the change vs. the developer impact, I agree with you on this - it's the pragmatic solution. Identity maps are hard or even impossible, caching `has_one()` on the other hand is a pragmatic fix. I still believe this change will trip up developers with edge cases though (even with #4188 in place), so should be treated as an API change for 4.0. We could make it opt-in for 3.2? @tractorcow What do you think?

On DataModel, that's an attempt at DI which predates `Injector`, and never really caught on. In my opinion it weakens existing APIs, particularly when it's passed through on high level APIs like `Controller->handleRequest()`, which shouldn't even make any assumptions about a data model being present. I'd rather rely on the global `Injector` getters like most other places in the framework.

tractorcow:
I could see this as being a non-breaking change in 3.2, even without it being opt-in. I'm of a mind it would be very hard to get this into 3.1 though.

Sounds like DataModel could be dropped in 4.0 then :)

sminnee:
If we accept that the API-breakage-risk on this is low or medium, then I think we should get it into 3.2-beta2, as this is a serious performance regression on 2.4.

Identity Map sounds great for 4.0 but it would be good to have a partial solution before then...

What do the other core committers think?

tractorcow:
@pitchandtone what are your thoughts on using the existing `DataObject::get_by_id` cache instead of adding another?

pitchandtone:
@tractorcow would you want to re-implement a version of this?

$cacheKey = "{$filter}-{$orderby}";
if($extra = $SNG->extend('cacheKeyComponent')) {
	$cacheKey .= '-' . implode("-", $extra);
}
$cacheKey = md5($cacheKey);

It feels like it would use a different cacheKey structure, and I'm not sure how if the extension 'cacheKeyComponent' wold be useful in this instance. 

If you did want to I'd be pushing that the cache is either:
a) Attempting to be exactly the same as if you'd used the get_one method, so that they can share the cache properly
b) As is, a completely different cache. 

Two different styles of cache key within one cache doesn't seem like a good idea to me.

tractorcow:
I'm not saying re-implement that method; I mean actually call that method.

sminnee:
Do we want to refactor this to instead become part of DataList::First(), and cache a digest of the relevant query?

tractorcow:
@sminnee That would probably work. I would recommend that we make as much use of existing caching mechanisms in 3.x, where possible.

Having a datalist-level cache is probably a good idea; I wonder if we should implement this properly in 4.0 with a new api for this. I actually see other problems needing solving in this area, such as caching of multiple-row results.

sminnee:
I had a quick go and it looks too invasive for 3.x.

The simplest approach for this PR would probably be to have getComponent() call get_one().

pitchandtone:
Now calling get_one, which is waaaaay clearer and more simple.

model/DataObject.php
```diff
@@ -1417,9 +1417,9 @@ public function getComponent($componentName) {
 		if($class = $this->has_one($componentName)) {
 			$joinField = $componentName . 'ID';
 			$joinID    = $this->getField($joinField);
-			
+
 			if($joinID) {
-				$component = $this->model->$class->byID($joinID);
+				$component = DataObject::get_one($class, 'ID = ' . $joinID);
```
tractorcow: This is actually producing non-standard SQL. You need to quote your SQL identifiers if possible.

Also, when querying on ID there is the risk of an ambiguous field error (if a multi-table class is joined on ID).

If you use `DataObject::get_by_id($class, $joinID)` this is taken care of and uses the same cache. This is why I recommended it earlier. :)

```php
/**
	 * Return the given element, searching by ID
	 *
	 * @param string $callerClass The class of the object to be returned
	 * @param int $id The id of the element
	 * @param boolean $cache See {@link get_one()}
	 *
	 * @return DataObject The element
	 */
	public static function get_by_id($callerClass, $id, $cache = true) {
		if(is_numeric($id)) {
			if(is_subclass_of($callerClass, 'DataObject')) {
				$baseClass = ClassInfo::baseDataClass($callerClass);
				return DataObject::get_one($callerClass,"\"$baseClass\".\"ID\" = $id", $cache);

				// This simpler code will be used by non-DataObject classes that implement DataObjectInterface
			} else {
				return DataObject::get_one($callerClass,"\"ID\" = $id", $cache);
			}
		} else {
			user_error("DataObject::get_by_id passed a non-numeric ID #$id", E_USER_WARNING);
		}
	}
```

tractorcow:
@pitchandtone backing up the discussion a little bit, since we discussed that this was an api change (with a risk) it would be safer to push this to 3.2. I've taken the liberty of rebasing and modifying your original PR into an alternative at https://github.com/silverstripe/silverstripe-framework/pull/4468.

sminnee:
#4468 has been merged :-)

tractorcow:
Yay thx

PR #4188, opened 2015-05-14
if has_one relationship ID has changed, clear component cahe

pitchandtone:
otherwise if the relationship has already been populated, then the original object will be pulled from the instances cache rather than querying against the new ID.

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -2345,6 +2345,10 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 
 					// Value has changed as well, not just the type
 					$this->changed[$fieldName] = 2;
+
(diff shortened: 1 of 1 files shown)


model/DataObject.php
```diff
@@ -2345,6 +2345,10 @@ public function setField($fieldName, $val) {
 
 					// Value has changed as well, not just the type
 					$this->changed[$fieldName] = 2;
+
```
chillu: Can you please describe what this code does through an inline comment? It's not very obvious :)

chillu:
Thanks John! I don't really like the hardcoded knowledge of how foreign keys are constructed in there, but `flushCache()` isn't granular enough for your purpose here. And we already construct the foreign key through string concatenation in half a dozen other places on `DataObject`, so any overhaul will need to look at the bigger picture anyway. Happy to merge once you add that comment.

dhensby:
Fixed in #4290

PR #4189, opened 2015-05-15
API Replace Page Delete with new Page Archive action

tractorcow:
The major distinction between archiving and deleting is that "deleting" only deletes from stage (which is practically pretty useless), but archiving removes both live and stage, but informs the user that the page is safely recorded in the archive.

"delete" gives the false impression that the page is lost forever, which has never been the case.

This PR must be merged and reviewed alongside a CMS pull request incoming.

Lots of old code has been updated and fixed; Some which hasn't worked since 3.0 (e.g. batch action confirmations).

image

### DIFF
--- a/admin/javascript/LeftAndMain.BatchActions.js
+++ b/admin/javascript/LeftAndMain.BatchActions.js
@@ -38,12 +38,89 @@
 					this.serializeFromTree();
 				}
 			},
+			
+			/**
+			 * Register default bulk confirmation dialogs
+			 */
+			registerDefault: function() {
+				// Publish selected pages action
+				this.register('admin/pages/batchactions/publish', function(ids) {
+					var confirmed = confirm(
+						ss.i18n.inject(
+							ss.i18n._t(
+								"CMSMAIN.BATCH_PUBLISH_PROMPT",
+								"You have {num} page(s) selected.\n\nDo you really want to publish?"
+							),
+							{'num': ids.length}
+						)
+					);
+					return (confirmed) ? ids : false;
+				});
+
+				// Unpublish selected pages action
+				this.register('admin/pages/batchactions/unpublish', function(ids) {
+					var confirmed = confirm(
+						ss.i18n.inject(
+							ss.i18n._t(
+								"CMSMAIN.BATCH_UNPUBLISH_PROMPT",
+								"You have {num} page(s) selected.\n\nDo you really want to unpublish"
+							),
+							{'num': ids.length}
+						)
+					);
+					return (confirmed) ? ids : false;
+				});
+
+				// Delete selected pages action
+				// @deprecated since 4.0 Use archive instead
+				this.register('admin/pages/batchactions/delete', function(ids) {
+					var confirmed = confirm(
+						ss.i18n.inject(
+							ss.i18n._t(
+								"CMSMAIN.BATCH_DELETE_PROMPT",
+								"You have {num} page(s) selected.\n\nDo you really want to delete?"
+							),
+							{'num': ids.length}
+						)
+					);
+					return (confirmed) ? ids : false;
+				});
+
+				// Delete selected pages action
+				this.register('admin/pages/batchactions/archive', function(ids) {
+					var confirmed = confirm(
+						ss.i18n.inject(
+							ss.i18n._t(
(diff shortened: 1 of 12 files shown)


tractorcow:
Ok the CMS pr is at https://github.com/silverstripe/silverstripe-cms/pull/1203

tractorcow:
This is a semi-major change, so please ensure that this PR is left open with plenty of time for users to review and give feedback.

admin/javascript/LeftAndMain.BatchActions.js
```diff
@@ -38,12 +38,89 @@
 					this.serializeFromTree();
 				}
 			},
+			
+			/**
+			 * Register default bulk confirmation dialogs
+			 */
+			registerDefault: function() {
+				// Publish selected pages action
+				this.register('admin/pages/batchactions/publish', function(ids) {
```
tractorcow: I found this code here, and it hasn't been working for years. ;)

assertchris:
/me stops hovering over the merge button...

admin/javascript/LeftAndMain.BatchActions.js
```diff
@@ -319,8 +420,10 @@
 					btn.attr('disabled', 'disabled').button('refresh');
 				} else {
 					btn.removeAttr('disabled').button('refresh');
-					// form.submit();
-				} 
+				}
+				
+				// Refresh selected / enabled nodes
+				$('#Form_BatchActionsForm').refreshSelected();
```
tractorcow: This code was removed years ago; Re-added this in to ensure tree items are properly filtered.

tractorcow:
Update to this; I've re-tested and fixed some of the CSS rules in the sitetree, and fixed some issues with recent work to make it work with the updated flags / styles.

tractorcow:
A few updates:

- Restoring an archived pages now restores its parents
- Some more CSS cleanup / fixes to make sure that disabled / unfiltered styles are properly inherited

chillu:
So excited this is finally happening, great work guys! I've just checked it out locally. Long-term I think we need to make it more obvious on how to find archived pages in the site tree again - at the moment they stay in the tree until you reload, which might come as a surprise to users. You can only get them back through the filter panel.

On that note: You should update the userhelp ;)

tractorcow:
I've updated the other PR (https://github.com/silverstripe/silverstripe-cms/pull/1203) based on feedback:

Instead of automatically restoring the parent page, the user is given the option to restore any orphaned pages to the root. The text / description / alert on the action will change depending on the parent page.

tractorcow:
More bugfixes to https://github.com/silverstripe/silverstripe-cms/pull/1203:

- Side by side preview was not working on archive action, so ContentController has been fixed for checking permissions of the page in the current stage
- Fixed some typos in the new alerts
- Changed "Un-publish" to "Unpublish"
- Changed "Restore" and "Restore at top level" to "Restore draft" and "Restore draft at top level"

tractorcow:
Rebased with new colours / hover behaviours. :D

image

image

tractorcow:
Rebased again

sminnee:
This looks nice, although I think that `applicablePages()` might need a test, and I'd also be keen to see how this interacts with advancedworkflow, as a smoke-test of how complicated modules might be broken by this change.

tractorcow:
There aren't any tests anywhere for batch actions... so... I'll just have to make all new ones I guess.

tractorcow:
@sminnee I've added tests but they are on the https://github.com/silverstripe/silverstripe-cms/pull/1203 PR, since framework doesn't actually implement that API anywhere.

tractorcow:
Rebased again

sminnee:
It seems that this breaks the build on postgres, although I can't understand why. :-(

assertchris:
https://github.com/silverstripe/silverstripe-framework/pull/4189/files#diff-69e8356e095b9501b3b000dbed87a13dR276 maybe?

tractorcow:
Nothing to do with postgres; It's that one test that only sometimes passes. I restarted the build for postgres.

tractorcow:
Fixed :D

tractorcow:
Yay thanks

chillu:
Related: https://github.com/silverstripe/silverstripe-framework/issues/6202

Issue #4073, opened 2015-04-15
TextField is not implementing validate() method


bendubuisson:
You can set a maxLength on a TextField but the `validate() `method is not implemented on the `TextField` class. 
Therefore it never returns an error if I set a maxlength and the field is posted with a value over that maxlength.

Or did I miss something?

chillu:
Hey Ben, yeah that'd be a good addition to core - wanna write a (unit tested) pull request against 3.2? :)

bendubuisson:
Hey I had to write a custom formfield that extends TextField and does that validation, so yeah I'll submit that at some stage.

Thanks

kamrandotpk:
@chillu / @bendubuisson : hey Ingo and Ben, I noticed there was no pull request for this issue yet so I made the required changes + tests and created a pull request.

kinglozzer:
Closed via 20c9d6f02e1c45f8ac9926d167558687e35fb446

PR #4190, opened 2015-05-15
Issue #4073: Implemented the validate() method for the TextField class

kamrandotpk:
Issue #4073: Implemented the validate() method for the TextField class and check if the value is not greater than the set max length of the TextField

### DIFF
--- a/forms/TextField.php
+++ b/forms/TextField.php
@@ -80,4 +80,26 @@ class TextField extends FormField {
 
 		return $this->Field();
 	}
+
+    /**
+     * Validate this field
+     *
+     * @param Validator $validator
+     * @return bool
+     */
+    public function validate(Validator $validator) {
+        if(!is_null($this->maxLength) && mb_strlen($this->value) > $this->maxLength) {
+            $validator->validationError(
+                $this->name,
+                _t(
+                    'TextField.VALIDATEMAXLENGTH',
+                    'The value for {name} must not exceed {maxLength} characters in length',
+                    array('name' => $this->getName(), 'maxLength' => $this->maxLength)
+                ),
+                "validation"
+            );
+            return false;
+        }
+        return true;
+    }
 }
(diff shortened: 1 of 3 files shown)


kinglozzer:
@kamrandotpk Great work! Just a few things that need correcting:

- We use tabs instead of spaces for indentation, could you switch to tabs please? (except .yml files as those use spaces). Could you also add an empty new line to the end of `TextFieldTest.php`?
- The indentation of the entity you added to en.yml doesn’t quite match up to the existing entities: each level should be two spaces “deeper” than its parent
- I’d probably remove the `setup()` method and extra properties from the test class and just pass the field name + length as normal - if someone changes that length to affect one test in future, they might accidentally affect other tests. We can also remove the extra validator class: other form field tests just pass an empty `RequiredFields` validator: `$field->validate(new RequiredFields());`

The only other thing is that this PR should’ve been submitted against the 3 branch instead of 3.1 (as we follow [semantic versioning](http://semver.org/), this will need to go in the next “minor” release). If you can rebase this PR against the 3 branch and re-open it, great! If not, don’t worry about it: we can merge it into 3 for you :)

Thanks!

kamrandotpk:
@kinglozzer Many thanks for the excellent feedback =) I'll make the necessary changes and update here. Thanks.

kamrandotpk:
@kinglozzer hey Loz, I've made the necessary changes. 

Regarding submitting the PR against 3 instead of 3.1, it would be awesome if you can merge it to 3 for me. Thanks! =)

kinglozzer:
@kamrandotpk Looks good, thanks! As soon as #4175 is merged I’ll merge this into the 3 branch :)

tractorcow:
#4175 has been merged @kinglozzer

kinglozzer:
Squashed and merged: 20c9d6f02e1c45f8ac9926d167558687e35fb446. Thanks @kamrandotpk!

PR #4191, opened 2015-05-16
BUG: FieldList::changeFieldOrder() leftovers discarded

jonom:
Logical error. Use of + operator means items from second array are only merged if the key does not already appear in the first array. The first array has numeric keys 0,1,2 etc. The second array is keyed by field name, but array_values() resets the keys to be numberic starting at 0. This means that some or all leftovers are discarded instead of appended.

### DIFF
--- a/forms/FieldList.php
+++ b/forms/FieldList.php
@@ -561,7 +561,7 @@ class FieldList extends ArrayList {
 		}
 		
 		// Add the leftover fields to the end of the list.
-		$fields = $fields + array_values($fieldMap);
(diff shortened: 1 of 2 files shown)


kinglozzer:
Nice find! Any change you could write up test that covers this? :)

jonom:
My first test! Look okay?

tractorcow:
Thanks @jonom :D

jonom:
Hey look at that, merged already! Thanks @tractorcow

PR #4194, opened 2015-05-18
add async and defer attributes to js requirements

xini:


xini:
any feedback re this? It would be great if that could be included in one of the next releases... Cheers Flo

dhensby:
Sorry to be a pain, but this needs to go into the 3 or 3.2 branch as its a new feature 

I'm slightly concerned we are adding complexity and maintenance burden for little gain here. Though perhaps this would be useful to some. 

Please also add tests to prove it does what it's meant to

xini:
Thanks. rebased and added tests in #4427

PR #4195, opened 2015-05-19
DOC: Correct composer invocation to install phpunit

oddnoc:
The example as given pulled down a much more recent version of phpunit
than what is compatible with SilverStripe. This change corrects the
example to use 3.7.*@stable, and removes a redundant reference to this
version number.

### DIFF
--- a/docs/en/02_Developer_Guides/06_Testing/index.md
+++ b/docs/en/02_Developer_Guides/06_Testing/index.md
@@ -19,22 +19,16 @@ If you are familiar with PHP coding but new to unit testing then check out Mark'
 You should also read over [the PHPUnit manual](http://www.phpunit.de/manual/current/en/). It provides a lot of
 fundamental concepts that we build on in this documentation.
 
-Unit tests are not included in the normal SilverStripe downloads so you need to install them through git repositories
-([installation instructions](/getting_started/composer)).
+Unit tests are not included in the zip/tar.gz SilverStripe [downloads](http://www.silverstripe.org/software/download/) so to get them, install SilverStripe [with composer](/getting_started/composer).
 
-## Install with Composer
+## Invoking phpunit
 
-Once you've got your project up and running, open a terminal and cd to your project root.
+Once you have used composer to create your project, `cd` to your project root. Composer will have installed PHPUnit alongside the required PHP classes into the `vendor/bin/` directory.
 
-	composer require --dev "phpunit/phpunit:*@stable"
+If you don't want to invoke PHPUnit through its full path (`vendor/bin/phpunit`), add `./vendor/bin` to your $PATH, or symlink phpunit into the root directory of your website:
 
-This will install [PHPUnit](http://www.phpunit.de/) dependency, which is the framework we're
-building our unit tests on.  Composer installs it alongside the required PHP classes into the `vendor/bin/` directory.
-
-### Symlinking the PHPUnit Binary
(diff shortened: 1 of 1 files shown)


tractorcow:
Actually even more of this document is outdated now, as we now include unit test functionality in the core downloads. All that you need to do is ensure that --no-dev is not used and the default installer will install the necessary version of phpunit.

See https://github.com/silverstripe/silverstripe-installer/blob/3.1/composer.json

oddnoc:
Thanks for pointing that out! Here's a new rewrite.

docs/en/02_Developer_Guides/06_Testing/index.md
```diff
@@ -19,22 +19,16 @@ If you are familiar with PHP coding but new to unit testing then check out Mark'
 You should also read over [the PHPUnit manual](http://www.phpunit.de/manual/current/en/). It provides a lot of
 fundamental concepts that we build on in this documentation.
 
-Unit tests are not included in the normal SilverStripe downloads so you need to install them through git repositories
-([installation instructions](/getting_started/composer)).
+Unit tests are not included in the zip/tar.gz SilverStripe [downloads](http://www.silverstripe.org/software/download/) so to get them, install SilverStripe [with composer](/getting_started/composer).
```
tractorcow: `are included`

docs/en/02_Developer_Guides/06_Testing/index.md
```diff
@@ -19,22 +19,16 @@ If you are familiar with PHP coding but new to unit testing then check out Mark'
...
 
-## Install with Composer
+## Invoking phpunit
 
-Once you've got your project up and running, open a terminal and cd to your project root.
+Once you have used composer to create your project, `cd` to your project root. Composer will have installed PHPUnit alongside the required PHP classes into the `vendor/bin/` directory.
 
-	composer require --dev "phpunit/phpunit:*@stable"
+If you don't want to invoke PHPUnit through its full path (`vendor/bin/phpunit`), add `./vendor/bin` to your $PATH, or symlink phpunit into the root directory of your website:
```
tractorcow: I do `composer global require phpunit/phpunit 3.7@stable` and it installs the phpunit bin into my ~/.composer/vendor/bin/ folder. I add this to my path and it all works. :)

That means my phpunit isn't tied to a single project; If it's moved or deleted I don't have to re-link it.

tractorcow:
I just had a few more ideas for your pr. I hope it's not too much trouble! You're doing a good job and thanks for the contribution. :)

docs/en/02_Developer_Guides/06_Testing/index.md
oddnoc: What about the case where some non-SilverStripe project needs a different version of phpunit? Doesn't it make more sense to tie each project to its preferred phpunit? I can see your way making sense in a homogeneous environment. What do you think about presenting both alternatives?

chillu: Agree with Fred, global dependencies lead to confusion, particularly if we then *also* install phpunit locally through the default `composer.json`.

docs/en/02_Developer_Guides/06_Testing/index.md
chillu: Just downloaded from the mentioned URL, and confirmed that `framework/tests` isn't included.

PR #4196, opened 2015-05-19
API Enable filters to perform 'IS NULL' or 'IS NOT NULL' checks

tractorcow:
Solves #3621

### DIFF
--- a/docs/en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
+++ b/docs/en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
@@ -355,6 +355,40 @@ You can use [SearchFilters](searchfilters) to add additional behavior to your `f
 		'PlayerNumber:GreaterThan' => '10'
 	));
 
+### Filtering by null values
+
+Since null values in SQL are special, they are non-comparable with other values, certain filters will add
+`IS NULL` or `IS NOT NULL` predicates automatically to your query. As per ANSI SQL-92, any comparison
+condition against a field will filter out nulls by default. Therefore, it's necessary to include certain null
+checks to ensure that exclusion filters behave predictably.
+
+For instance, the below code will select only values that do not match the given value, including nulls.
+
+
+	:::php
+	$players = Player::get()->filter('FirstName:not', 'Sam');
+	// ... WHERE "FirstName" != 'Sam' OR "FirstName" IS NULL
+	// Returns rows with any value (even null) other than Sam
+
+
+If null values should be excluded, include the null in your check.
+
+
+	:::php
+	$players = Player::get()->filter('FirstName:not', array('Sam', null));
+	// ... WHERE "FirstName" != 'Sam' AND "FirstName" IS NOT NULL
+	// Only returns non-null values for "FirstName" that aren't Sam.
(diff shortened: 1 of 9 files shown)


sminnee:
OK so I've raised a query about whether `FirstName:not => Sam` should include rows with FirstName = null. In my view, `:not` should perform an strict inverse of the query, and therefore rows with FirstName = null should be included in `FirstName:not => Sam`.

The counter-argument would be that it's better to stick closer to the SQL semantics, but I disagree with that, because I find that particular piece of SQL semantics to be horrible & confusing to newcomers.

/cc @tractorcow @dhensby @hafriedlander @chillu thoughts?

sminnee:
(As an aside, for changes such as these it's great to have clear docs & tests to use to guide the discussion, so thanks for that!)

tractorcow:
Yeah this is really hard.... NULL in sql doesn't mean empty, it means we have no known value for this field. That being said, we can't include that in any results where we define the value must have a known (or known excluded) value. This is a little different to the value of null in PHP which is a literal "empty" value.

Essentially we make a call on whether we feel we should attempt to "fix" the sql behaviour for null filtering with more layers, or if we should document it and leave the code as a more shallow wrapper around the low level behaviour.

As a senior developer my initial feeling is to leave the low level behaviour as is, since I'm used to thinking about a "this is how the sql will work" when I'm writing my db filters. Since I understand the low level logic, I don't have to worry about how decisions will be made "for" me by the abstraction layer. It's like wearing thin soled shoes... I know what's directly under my feet without having to check.

A more intermediate developer might be more inclined to care only about the high level logic, and thus would expect these decisions to be made for him in the way you suggested, @sminnee, making judgments internally when a :not filter should or shouldn't include null checks automatically.

I guess we should look at our values as direction here. Shallow wrapper and require developer understanding, or thicker wrapping and not expect developers not to think much about what their SQL looks like?

chillu:
I don't have strong feelings either way, as long as its well documented :) Can't say I've ever hit this case when writing SQL (expecting NULL rows in `!=` comparisons), so I don't think its worth losing much sleep over.

tractorcow:
I lost lots of sleep over this and came to the conclusion that @sminnee is right. :D SS_Filterable is an abstract API and we shouldn't expect users to make assumptions around the data backend.

tractorcow:
Ok, this took a bit more time than I expected (ExactMatchFilter::manyFilter ended up a little complex), but I've implemented the necessary fix.

As before, the documentation explains exactly what the assumptions underlying the code makes, and there are plenty of tests to cover all the various use cases for each filter (not null, null, case-noncase checking, filter many, filter one, exclusions).

tractorcow:
@sminnee pull request updated with magic nulls :)

kinglozzer:
This looks great, I agree that the updated behaviour makes more sense at an API level:

```php
// NULL results obviously shouldn't be included here
->filter('FirstName', 'Sam');

// Inverse of the above, so NULL results should be included.
// Another way of looking at it: we're getting all rows where FirstName isn't Sam.
// Does null == 'Sam'? Obviously not, so null results should be included
->filter('FirstName:not', 'Sam')
```

tractorcow:
Ok then do you want to merge this @kinglozzer ? :)

kinglozzer:
I do indeed!

PR #4198, opened 2015-05-19
Change drag/drop icon and make them only appear on hover

scott1702:
image


tractorcow:
It looks a bit broken on my checkout

image

scott1702:
@tractorcow You might have to flush your cache now that the image is changed.

tractorcow:
Nevermind, cleared my cache and it ... ah too late you saw it.

tractorcow:
An issue I notice is during drag; If you drag a page over another one, the icon will appear, but instead of the icon appearing on the left it shifts the whole text to the right.

image

scott1702:
Good spotting, updated it. See what you can break now :)

tractorcow:
I've re-tested it and it seems fine to me now.

Does anyone else want to put their opinion on this change? @silverstripe/core-team ?

kinglozzer:
I think this looks great! I’ve tested it in IE8 and IE9 too and it works well in both of those :) my only bits of feedback are:

- The area that displays the “move” cursor is quite small, it tends to flicker between cursor types when you try to hover over the drag handle. That doesn’t stop you dragging of course, but the move cursor implies that that’s the only way to move items, so it’d be good to make it easier to “grab”. Perhaps a little extra width on the icon would help?
- The drag handle doesn’t show until you hover over the text to the right of it. Hard to explain, but start with your mouse to the left of the menu and slowly move it to the right and you should see what I mean. Adjusting the padding on the anchor instead of the margin seems to fix this

Loving all these little improvements to the UI!

scott1702:
Very valid points @kinglozzer, I'll have a look into polishing it up a little first thing in the morning.

willmorgan:
As long as it's also visible on focus as well as hover, decluttering the menu design is a :+1: from me.

scott1702:
Thanks for the feedback guys, updated :)

jonom:
Hey @scott1702 I noticed that you modified some css styles that reside in the `thirdparty` directory - I would suggest leaving those files unchanged and instead introducing a style rule to override them in ` admin/scss/_tree.scss`. I assume (or hope) that all those thirdparty files get updated from time to time and we wouldn't want your work to be lost in the process :)

chillu:
:+1:, as long as the thirdparty code modification issue is resolved.

scott1702:
Yep, just found another bug too that I need to fix. I'll make those changes now.
image

scott1702:
Updated, Thanks @jonom

kinglozzer:
I’ve re-tested in IE8, IE9, Firefox and Chrome and everything’s working perfectly. The thirdparty issue is resolved, and I’ve confirmed that the drag handle does show on focus, so I think we’re good to go!

Thanks @scott1702!

willmorgan:
:trophy:

Issue #4187, opened 2015-05-14
File->setName($name) logical error?


cwchong:
In filesystem/File.php under the function setName:
given a file name "test.jpg":

```php
$base = pathinfo($name, PATHINFO_BASENAME); // returns "test.jpg"
$ext = self::get_file_extension($name); // returns "jpg", get_file_extension uses "PATHINFO_EXTENSION"
...
$name = "$base-$suffix$ext"; // returns "test.jpg-2jpg" instead of an expected "test-2.jpg"
```
```php
// fix
$base = pathinfo($name, PATHINFO_FILENAME); 
...
$name = "$base-$suffix.$ext";
```
```php
// test case: assuming you already have a file named "test.jpg" in an "assets/Uploads" folder 
$folder = Folder::find_or_make('Uploads');
$folder->write();

$file1 = new File();
$file1->setParentID($folder->ID);
$file1->write();
$file1->setName('test.jpg');
$file1->write(); // success

$file2= new File();
$file2->setParentID($folder->ID);
$file2->write();
$file2->setName('test.jpg');
$file2->write(); // failed with "Extension is not allowed" error, presumably due to the "test.jpg-2jpg" file name
```

kinglozzer:
Closed via 21e891e28cb7d861d5bbb44f64ad81170cc778ee

PR #4200, opened 2015-05-19
Bug fix: File->setName()

cwchong:
Logical error. $base should use PATHINFO_FILENAME instead of
PATHINFO_BASENAME in order to exclude extension; final $name should
include a period before $ext as $ext uses PATHINFO_EXTENSION which
excludes the period.

Reference issue: https://github.com/silverstripe/silverstripe-framework/issues/4187

### DIFF
--- a/filesystem/File.php
+++ b/filesystem/File.php
@@ -609,7 +609,7 @@ class File extends DataObject {
 
 		// If it's changed, check for duplicates
 		if($oldName && $oldName != $name) {
-			$base = pathinfo($name, PATHINFO_BASENAME);
+			$base = pathinfo($name, PATHINFO_FILENAME);
 			$ext = self::get_file_extension($name);
 			$suffix = 1;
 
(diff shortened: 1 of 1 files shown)


tractorcow:
PATHINFO_BASENAME should return the name WITHOUT extension, and vice versa for PATHINFO_FILENAME which should include the extension. The example you gave is contrary to the documentation.

Can you please include a unit test to prove this fix?

cwchong:
The example I gave behaves exactly as described by the php documentation at http://php.net/manual/en/function.pathinfo.php

$myfile = "/path/to/myfile.txt";
echo pathinfo($myfile, PATHINFO_BASENAME) ."\n"; // returns WITH extension "myfile.txt"
echo pathinfo($myfile, PATHINFO_FILENAME ) ."\n"; // returns WITHOUT extension "myfile"

This is contrary to what you have just listed.
May I ask which documentation are you referring to?

tractorcow:
No, I think you're right... my thinking is backwards. I apologise for not reading the documentation properly and paying attention to your code.

PR #4202, opened 2015-05-19
FIX: TreeDropdownField Folder expansion

jonom:
When viewing a Folder tree, an expansion icon was shown if the Folder had *any* children, but it should be restricted to children that are Folders. This has annoyed me for a long time :)

The `Hierarchy::getChildrenAsUL()` function requires a separate method for counting children, but this hasn't been accommodated in TreeDropdownField so numChildren calculations were wrong. I'm guessing the requirement for the separate counting function predates lazy-loading as now it should be quite efficient to just do `$SomeListMethod()->count()`.

Propose that for SS4.0 the `$numChildrenMethod` argument in `Hierarchy::getChildrenAsUL()` is deprecated, and the total can then be simply calculated using `$childrenMethod()->count()`. Or maybe we could even sneak this in to 3.2?

Companion fix for CMS at https://github.com/silverstripe/silverstripe-cms/pull/1205

### DIFF
--- a/filesystem/File.php
+++ b/filesystem/File.php
@@ -390,7 +390,6 @@ class File extends DataObject {
 
 		//get a tree listing with only folder, no files
 		$folderTree = new TreeDropdownField("ParentID", _t('AssetTableField.FOLDER','Folder'), 'Folder');
-		$folderTree->setChildrenMethod('ChildFolders');
 
 		$fields = new FieldList(
 			new TabSet('Root',
--- a/admin/javascript/LeftAndMain.Content.js
+++ b/admin/javascript/LeftAndMain.Content.js
@@ -46,9 +46,6 @@
 					// when JSTree auto-selects elements on first load.
 					if(!origEvent) {
 						return false;
-					}else if($(origEvent.target).hasClass('jstree-icon') || $(origEvent.target).hasClass('jstree-pageicon')){
-						// in case the click is not on the node title, ie on pageicon or dragicon, 
-						return false;
 					}
 					
 					// Don't allow checking disabled nodes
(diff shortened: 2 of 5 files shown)


jonom:
Hey core devs here's a demonstration of the bug:

**Before:** (nasty!)
before2

**After:**
after2

Github seems to be cutting out the loops so you may need to click through and reload to take it in.

jonom:
@tractorcow any chance of a review on this one? I think it's a reasonably important UX bug and makes the CMS seem 'dodgy'.

filesystem/File.php
```diff
@@ -390,7 +390,6 @@ public function getCMSFields() {
 
 		//get a tree listing with only folder, no files
 		$folderTree = new TreeDropdownField("ParentID", _t('AssetTableField.FOLDER','Folder'), 'Folder');
-		$folderTree->setChildrenMethod('ChildFolders');
```
tractorcow: Can you explain why we don't use this anymore? I thought the fix was to only count folders not all children?

jonom: Still doing it, we're just doing it automatically in TreeDropDownField when the source class is set to Folder. Leaving this line would be harmless but it's redundant.

tractorcow: Oh, I see. I really prefer that treedropdownfield didn't have sepcial exemptions built in, but I guess we can't change that in 3.1.

tractorcow:
thank @jonom for your fix. Really good work. :)

filesystem/File.php
jonom: Yeah it's a bit of a tricky one... the whole approach to generating and marking a tree could probably use a good look over. Thanks for the merge!

jonom:
Thanks mate!

jonom:
@tractorcow there is a related CMS PR for this at https://github.com/silverstripe/silverstripe-cms/pull/1205

PR #4203, opened 2015-05-20
Fix broken inline buttons

scott1702:
To fix Share Draft button and any future buttons not being inline.

https://github.com/silverstripe-labs/silverstripe-sharedraftcontent/issues/31


flashbackzoo:
Fixes https://github.com/silverstripe-labs/silverstripe-sharedraftcontent/issues/31

@uniun Can you remember when you made this `display: block`?

uniun:
Because `display: inline-block;` adds empty space below dropdown field.
I've played with CSS a bit more and found out that it is possible to replace `display: inline-block;` with `vertical-align: bottom;`. It shoud work with *sharedraftcontent* module too.

scott1702:
@uniun Could you please provide a picture of what you mean by `display: inline-block;` adding extra space below the dropdown field? It seems to work better for me using `display: inline-block` (On Chrome).

`display: block;`
image

`display: inline-block`
image

uniun:
Opera 29 / Chrome 43
dropdowns

scott1702:
Ah thanks @uniun, I didn't realise we were talking about different things. Updated and rebased.

PR #4204, opened 2015-05-20
Updated SiteTree status overlay colours

scott1702:
as per conversation on https://github.com/silverstripe/silverstripe-framework/pull/3936
image


tractorcow:
This needs to be rebased somehow. Probably due to the SCSS changes and files being regenerated.

PR #4208, opened 2015-05-20
BUG Fixed setting LastEdited for DataObject with class ancestry

### LINKED ISSUE #4207: [Bug] Testing: failed to set LastEdited field from fixture
Test case 1:

Setting LastEdited field on simple DataObject from fixture results in DB error.
(object ID is missing in where clause)

Test case 2:

Setting LastEdited field on DataObject with class ancestry results in DB error.

For example for Page <|-- SiteTree, the code tries to set Page.LastEdited field.

gregsmirnov:
fixes #4207

### DIFF
--- a/dev/FixtureBlueprint.php
+++ b/dev/FixtureBlueprint.php
@@ -183,14 +183,8 @@ class FixtureBlueprint {
 
 			// If LastEdited was set in the fixture, set it here
 			if($data && array_key_exists('LastEdited', $data)) {
-				$edited = $this->parseValue($data['LastEdited'], $fixtures);
-				DB::manipulate(array(
-					$class => array(
-						"command" => "update", "id" => $obj->id,
-						"fields" => array("LastEdited" => "'".$edited."'")
-					)
-				));
-			}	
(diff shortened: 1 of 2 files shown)


dev/FixtureBlueprint.php
```diff
@@ -183,14 +183,8 @@ public function createObject($identifier, $data = null, $fixtures = null) {
 
 			// If LastEdited was set in the fixture, set it here
 			if($data && array_key_exists('LastEdited', $data)) {
-				$edited = $this->parseValue($data['LastEdited'], $fixtures);
```
tractorcow: Would it be simpler to simply do `$class = ClassInfo::baseDataClass($class);`?

tractorcow: Or if you want to do it generically `$class = ClassInfo::table_for_object_field($class, 'LastEdited');`

tractorcow:
Thanks for the fix @gregsmirnov. I just gave a few suggestions that might simplify the code.

gregsmirnov:
@tractorcow Thank you for comments. Replacing this pull request with pull #4220

PR #4212, opened 2015-05-21
Add 'sticky' behaviour to left menu

flashbackzoo:
This adds dynamic expand / collapse behaviour to the left menu.

The idea is to increase the available space for content editing, allowing the user to focus on their task, and hide away the menu clutter.

The default view upon logging into the CMS remains the same, except for the additional icon in the bottom left.
sticky-off-expanded

Navigating to a page collapses the left menu, allowing more space for content editing.
sticky-off-collapsed-sitetree

If the user navigates back to the top level (anywhere the SiteTree is not rendered in the middle panel) the menu is expanded, back to the state in the first screenshot.

If a user prefers to have the menu expanded or collapsed at all times (disabling the automatic expand / collapse behaviour) they can toggle the 'sticky' icon.

sticky-on-expanded

The menu will now stay in it's expanded state, no matter where the user navigates. Toggling the 'sticky' icon again will revert back to the default automatic expand / collapse behaviour.


tractorcow:
How about something like this to indicate what the icon means

image

image

jonom:
I agree, definitely need something there to explain what it does. Think showing on hover only is the way to go to keep the clutter down. Few thoughts:

1. I wonder if the sticky button should be moved to just to the left of the `<<` icon, that way the button will be associated by proximity to the behaviour it's influencing.

2. I might have misunderstood but it sounds like the auto-collapse behaviour will only apply to the SiteTree. I think for consistency it needs to apply to all uses of that middle panel, which includes the Filter pane in other contexts.

3. When the left panel mode is set to 'auto' and the middle panel (Site Tree or Filter) is opened from the collapsed state, does this cause the left panel to shrink? I think it should.

clarkepaul:
I think he has the hover text built in already but didn't explain it to that detail. 

2&3. We will be moving the filter panel in our next piece of work, but the idea is that this would eventually apply to all areas of the CMS which go three levels deep (like the edit content area).
The collapse of the menu is not traggered by the sitetree being open but by a page in the sitetree being selected

tractorcow:
Can you please rebase this so I can check it out and test it locally?

flashbackzoo:
Just making the changes to the icon, adding an indicator. I'll rebase and push the latest changes up soom :)

flashbackzoo:
Here's the hover state...

collapsed-hover-fixed

expanded-hover-auto

tractorcow:
It looks and feel really good, but there are a few tweaks I would still make. It still feels to do things in an unexpectant way.

> Navigating to a page collapses the left menu, allowing more space for content editing.

What if the screen has lots of space? Can you make it only collapse if the editing width is less than a given threshold? There is a minimum size for this area, so could it be some value between the minimum and a slightly larger (but not too large) boundary?

I don't mind if the toggle only automatically triggers on click, although I'd feel better if it also responded to browser resize (although if that's a problem it's not that big a requirement).

The only thing that annoys me if it disappears if there is a lot of space already.

There is also a bug where if you "fix" it and reload the browser then it contracts, but it looks like some of the inner content isn't contracted properly. Should the "fix" persist via a cookie?

image

flashbackzoo:
@tractorcow I've fixed that bug and the sticky state is now persisted.

@clarkepaul What are your thoughts on the screen width behaviour?

clarkepaul:
I think that I need to see a few users using it to make a call based on their responses, you do make a good point though but for now lets keep it the same for everyone. I was trying to avoid having to design for too many edge cases and create one pattern. There are other things that should happen like drag'drop panels that would improve the usability.

chillu:
I tried to work out the behaviour of the new button explicitly without reading through this thread (just seeing the headling of "sticky nav" in email threads). And as a CMS user, I didn't really get it. "Sticky" and "fixed" could also refer to the scrolling position, or some undiscovered feature of moving the nav to a different edge of the screen? Just playing devil's advocate here.

I think to avoid confusion, the tooltip of "sticky nav" needs to be more explicit about its operation. How about `Collapse the navigation while editing ("auto") or keep the current state ("fixed")`

clarkepaul:
Good spot, its been updated

PR #4215, opened 2015-05-21
DataObject accept arrays or stdClass

russellshome:
The constructor of DataObject can take an array or stdClass for $record.
However, it is access as an array [here](https://github.com/silverstripe/silverstripe-framework/blob/3.1/model/DataObject.php#L416) and [here](https://github.com/silverstripe/silverstripe-framework/blob/3.1/model/DataObject.php#L431)

This pull request ensures $record is an array after validation

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -399,6 +399,9 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 				E_USER_WARNING);
 			$record = null;
 		}
+		if (is_a($record, "stdClass")){
+			$record = (array)$record;
+		}
 
 		// Set $this->record to $record, but ignore NULLs
 		$this->record = array();


willmorgan:
Hmm, good point.

I think we should just be accepting arrays, to be honest. I can't see the benefit in accepting a `stdClass`.

kinglozzer:
`stdClass` support was added intentionally in #917, I’m not sure how this wasn’t picked up then!

kinglozzer:
As we intentionally added support for this we should fix it, though I agree we should probably remove this support and only accept arrays. Added tests and merged in https://github.com/silverstripe/silverstripe-framework/commit/599c054d4fd7c2ef9a97abd5491585511c3edb1b. Thanks @russellshome!

PR #4217, opened 2015-05-22
BUG Fix DirectorTest failing when run with sake

tractorcow:
The tests run fine with phpunit cli, but when run with sake dev/tests/all the `$_SERVER['REQUEST_URI']` ends up with the dev/tests/all in the url, which actually breaks some assumptions made in testAlternativeBaseURL about how absoluteURL works.

I've also added yet another note on absoluteURL behaving weird in 3.1, where url-segments such as parent/child are treated as absolute urls.

### DIFF
--- a/tests/control/DirectorTest.php
+++ b/tests/control/DirectorTest.php
@@ -25,9 +25,11 @@ class DirectorTest extends SapphireTest {
 		if(!self::$originalRequestURI) {
 			self::$originalRequestURI = $_SERVER['REQUEST_URI'];
 		}
+		$_SERVER['REQUEST_URI'] = 'http://www.mysite.com';
 		
 		$this->originalGet = $_GET;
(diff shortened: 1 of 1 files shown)


tractorcow:
@silverstripe/core-team can someone please look at this? Also see https://github.com/silverstripe/silverstripe-framework/pull/4218 for a better url resolution scheme for 4.0 where I better address these issues in a non backwards compatible way.

chillu:
I'm recommending that we remove `sake dev/tests/` from 4.0, since there's a better functional equivalent (the `phpunit` binary). Or at least deprecate it in 4.0. https://groups.google.com/forum/#!topic/silverstripe-dev/I5dHy9dMVng

Doesn't affect this PR though. Merging into 3.1, since it only improves test behaviour, doesn't change APIs.

PR #4218, opened 2015-05-22
API New default behaviour for Director::getAbsoluteUrl

tractorcow:
Relative URL resolution now defaults to base_url, which is more consistent with templates.
You can also declare relative url resolution against site root or the current request (which is the old default)

Essentially a version of https://github.com/silverstripe/silverstripe-framework/pull/4217 which solves some extremely poor behaviour in the old method, but which could not be fixed without breaking backwards compatibility. (e.g. `Director::absoluteURL('my/page')` being treated as a site_root url when it's a relative url).

Also added `Director:: is_root_relative_url` to better handle relative urls which should not sit beneath the base_url by default.

@dhensby this is work that I've had in mind ever since the old fix you added a while back for empty and `.` or `./` urls; Do you feel this API is an improvement over the old code and behaviour? Do you agree that the new default (base_url) is better than the old default (request url)?

### DIFF
--- a/control/Director.php
+++ b/control/Director.php
@@ -16,6 +16,21 @@
  */
 class Director implements TemplateGlobalProvider {
 
+	/**
+	 * Specifies this url is relative to the base
+	 */
+	const BASE = 'BASE';
+
+	/**
+	 * Specifies this url is relative to the site root
+	 */
+	const ROOT = 'ROOT';
+
+	/**
+	 * specifies this url is relative to the current request
+	 */
+	const REQUEST = 'REQUEST';
+
 	static private $urlParams;
 
 	static private $rules = array();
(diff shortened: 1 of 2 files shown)


willmorgan:
This is way more elegant than before. Nice one.

control/Director.php
```diff
@@ -428,36 +443,56 @@ public static function set_current_page($page) {
...
+	 * Defaults to BASE, which is the same behaviour as template url resolution.
+	 * Igored if the url is absolute or site root.
 	 * @return string The fully qualified URL
 	 */
-	public static function absoluteURL($url, $relativeToSiteBase = false) {
-		if(!isset($_SERVER['REQUEST_URI'])) return false;
+	public static function absoluteURL($url, $relativeParent = self::BASE) {
+		if(is_bool($relativeParent)) {
+			// Deprecate old boolean second parameter
+			Deprecation::notice('5.0', 'Director::absoluteURL takes an explicit parent for relative url');
```
tractorcow: Note, we are doing deprecation at the next major version, as per our last public core committers hangout discussion, in case anyone was curious.

control/Director.php
```diff
@@ -428,36 +443,56 @@ public static function set_current_page($page) {
...
+			$relativeParent = $relativeParent ? self::BASE : self::REQUEST;
+		}
 
-		//a url of . or ./ is the same as an empty url
-		if ($url == '.' || $url == './') {
-			$url = '';
+		// Check if there is already a protocol given
+		if(preg_match('/^http(s?):\/\//', $url)) {
```
tractorcow: Because is_absolute_url isn't reliable for validating fully absolute urls.

tests/control/DirectorTest.php
```diff
@@ -97,36 +99,41 @@ public function testAbsoluteURL() {
...
 
-		//test absolute - /
+		// Test site root url
 		$this->assertEquals("$rootURL/", Director::absoluteURL('/'));
 
-		//test relative
-		$this->assertEquals("$rootURL/mysite/sub-page/", Director::absoluteURL('./'));
-		$this->assertEquals("$rootURL/mysite/sub-page/", Director::absoluteURL('.'));
+		// Test non-relative urls
+		foreach(array(Director::BASE, Director::ROOT, Director::REQUEST) as $method) {
```
chillu: Small stylistic niggle: Avoid loops in assertions, or ensure all input params (`$method`) are mentioned in the assertion message. Hard to debug otherwise

control/Director.php
chillu: I assume we'll include this in the 5.0 changelog once that release is prepped, as part of going through all deprecation notices? Might be a bit premature to create a 5.0 changelog already, but it'd help to describe the change while somebody can still remember making it ;)

chillu:
Looks good to me. Can somebody confirm that there's no directly security relevant aspects here? `Director::is_site_url()` is a sensitive method, but doesn't use any of the changed logic here. I guess custom implementations could have `$this->redirect(Director::absoluteURL($request->getVar('redirectPath')))`, e.g. if they want to ensure the correct subsite is used in all cases. I don't see any immediate way to exploit this though. Plus, this custom implementation should really use `is_site_url()` to begin with.

tests/control/DirectorTest.php
tractorcow: True, I'll unroll this. ;D

tractorcow:
@chillu I've updated this and unrolled the test.

Most of the security affecting components of this feature are already covered by tests, so, unaffected AFAIK. :)

tractorcow:
ta :D

PR #4220, opened 2015-05-22
BUG Fixed setting LastEdited for DataObject with class ancestry

### LINKED ISSUE #4207: [Bug] Testing: failed to set LastEdited field from fixture

gregsmirnov:
fixes #4207
replaces pull #4208


tractorcow:
Thanks @gregsmirnov :) Good job!

PR #4224, opened 2015-05-23
DOCS Fix link to common subclasses of form fields

camfindlay:


willmorgan:
`404` http://docs.silverstripe.org/en/developer_guides/forms/fields/common_subclasses
`200` http://docs.silverstripe.org/en/developer_guides/forms/field_types/common_subclasses

PR #3265, opened 2014-07-05
FEATURE allow force resampling on images

stevie-mayhew:
This change allows users to set a flag via configuration to force resampling of images even if they are the correct dimensions.

PR #4229, opened 2015-05-25
Max image methods & Force resampling changes

jonom:
Added ability to scale and crop images while ensuring up-sampling does not occur.

I also tried to simplify the `$force_resample` functionality but will see how the tests pan out. I couldn't run them locally, think I'm lacking the image files the tests need.

This is a work in progress - need to add some tests and docs but wanted to get the basics up for testing and feedback purposes.


jonom:
@tractorcow you wanted me to CC you - over to you ;)

model/Image.php
```diff
@@ -165,6 +165,20 @@ public function forTemplate() {
...
 
 	/**
+	 * Gets the relative URL accessible through the web. If forced resampling is 
+	 * enabled the URL will point to an optimised file.
+	 *
+	 * @uses Director::baseURL()
+	 * @return string
+	 */
+	public function getURL() {
+		if ($this->ClassName != 'Image_Cached' && Config::inst()->get('Image', 'force_resample')) {
```
tractorcow: You could probably use inheritance to solve the specific subclass behaviour here, rather than trying to hack in exemptions based on specific class names.

How about

```
public function getRealURL() { return parent::getURL(); }
public function getURL() {
  if($this->config()->force_resample) {
    return $this->getFormattedImage();
  } else {
    return $this->getRealURL();
  }
}
```

and then in Image_Cached

```
public function getURL() { return $this->getRealURL(); }
```

model/Image.php
```diff
@@ -233,10 +247,10 @@ public function SetRatioSize($width, $height) {
 		$heightRatio = $height / $this->height;
 		if( $widthRatio < $heightRatio ) {
 			// Target is higher aspect ratio than image, so check width
-			if($this->isWidth($width) && !Config::inst()->get('Image', 'force_resample')) return $this;
+			if($this->isWidth($width)) return $this;
```
tractorcow: I definitely prefer your logic in getURL rather than in each resize method; Much cleaner. :)

model/Image.php
```diff
@@ -328,6 +342,97 @@ public function generateSetSize(Image_Backend $backend, $width, $height) {
...
+	 * @param integer $width The relative (used to determine aspect ratio) and maximum width of the output image
+	 * @param integer $height The relative (used to determine aspect ratio) and maximum height of the output image
+	 * @return Image
+	 */
+	public function CropMax($width, $height) {
+		// Is the image already the correct size?
+		if ($this->isSize($width, $height)) return $this;
+		
+		// If not, make sure the image isn't upsampled
+		$imageRatio = $this->width / $this->height;
```
tractorcow: In case the file isn't found, $this->height will be zero, so it could divide by zero here if you're not careful. :D

model/Image.php
```diff
@@ -328,6 +342,97 @@ public function generateSetSize(Image_Backend $backend, $width, $height) {
...
+	 * @return Image
+	 */
+	public function ScaleMaxHeight($height) {
+		if ($this->height > $height) return $this->SetHeight($height);
+		return $this;
+	}
+
+	/**
+	 * Crop this image to the aspect ratio defined by the specified width and height, 
+	 * then scale down the image to those dimensions if it exceeds them.
```
tractorcow: Scale? It crops not scales right?

tractorcow: Ah, nevermind, it seems CroppedImage does that as well.

tractorcow: How about you put an example input -> output file in the phpdocs?

`e.g. an image at 100x200 with CropMax(150, 150) will end up as 100x150.` Not sure if that is right?

model/Image.php
```diff
@@ -328,6 +342,97 @@ public function generateSetSize(Image_Backend $backend, $width, $height) {
...
+		// Is the image already the correct size?
+		if ($this->isSize($width, $height)) return $this;
+		
+		// If not, make sure the image isn't upsampled
+		$imageRatio = $this->width / $this->height;
+		$cropRatio = $width / $height;
+		// If cropping on the x axis compare heights
+		if ($cropRatio < $imageRatio && $this->height < $height) return $this->CroppedImage($this->height*$cropRatio, $this->height);
+		// Otherwise we're cropping on the y axis (or not cropping at all) so compare widths
+		if ($this->width < $width) return $this->CroppedImage($this->width, $this->width/$cropRatio);
```
tractorcow: Do you need $imageRatio < $cropRatio condition here?

tractorcow: Caveat, I'm still not 100% sure of the difference between this method and CroppedImage. I need to be able to read the PHPDoc to understand immediately what the difference is. :)

model/Image.php
jonom: Yeah I was wondering why that check was only on SetRatioSize - but of course the other ones don't do division. I'll add the same check.

model/Image.php
stevie-mayhew: However it will now call the `ResizedImage` function in the case where some dimension ins the same on the image, rather than the original format name. The file name generated will not be the same as if it was called through the original function.

I'm pretty sure thats a backwards breaking API change and wouldn't be appropriate for 3.2 in that case?

tractorcow: The API hasn't really been broken; There is no promise that the file name returned from any specific function follows a specific format; Only that the name reflects the method used to generate that file, and I think that behaviour is still enforced. In other words, we are still naming the file after the method that ends up generating that file; Not the one that was originally requested.

model/Image.php
jonom: Nice! That's a better way.

tractorcow:
Could you please also add unit tests and update the docs? http://docs.silverstripe.org/en/developer_guides/files/image

model/Image.php
stevie-mayhew: I have to disagree - calling a function which has an expected output format has now changed that output format. The promise for a certain file naming convention was implicit and enforced by the way the code has been written and run prior to this change. 

All of these functions are public API, we can't change an output format for them without it being a breaking change.

jonom: As long as `$Image->Link()` etc. still works it shouldn't matter should it? Surely no one is hard coding links to `/assets/Uploads/_resampled/ResizedImageWzEyMDAsODAwXQ==-Sample-IMG-Apples.jpg"` or the like? And if they were it looks like 3.2 includes a breaking change there anyway - the arguments in the generated file name are now... weird. Encoded some how?

Plus I thought 3.2 was not a proper SemVer release, but the point from which we start following SemVer going forward? It already contains lots of breaking changes?

model/Image.php
jonom: That's not quite how that one works but I will add something to that effect :) in that example you're asking for a square (1:1) crop so that's what you will get, but to prevent up-sampling the output image will be only 100x100.

model/Image.php
jonom: The only thing it does differently is prevent upscaling. The output file will look the same as CroppedImage, but the width and height will be less than requested if it would cause up-sampling.

jonom:
Yes definitely will do that, just wanted to get the code up for feedback first.

I can't run the Image tests locally right now - can you tell me where I can get the `assets/ImageTest/` files I need to run them?

model/Image.php
stevie-mayhew: If its already changed for some reason then I withdraw my complaints.

This SemVer / 3.2 / 4.0 thing is getting really confusing - I feel like I have no idea sometimes and other times feel completely on top of it.

Anyway, changing the output of a public API method **is** a breaking change, regardless of what it changes and we should be careful not to make these decisions about the way things work without regarding even *potential* ways that they could have been used. Sure people probably aren't linking directly to that output, *but they might be* and should be able to until a non-BC version is made.

tractorcow: SemVer means that the expectations of the API are not broken; In this case, the expectation is that this method returns an object with the correct image sized. The expectation has nothing to do with HOW that is achieved internally. The promise to be kept is the API, not the specific implementation. 

For instance, [this](https://github.com/gitlabhq/gitlabhq/issues/7072) is not a semver violation.

model/Image.php
tractorcow: Ah I see.

tractorcow:
It looks as though those images are manually generated on the fly; Check GDImageTest.php

You can use the files in /tests/model/testimages if you like. :D

jonom:
Have updated the docs, removed the references to GD only functions and cleaned up a few things while I was there

jonom:
Thanks @tractorcow, I'm trying to run model/ImageTest.php but the tests are skipping for some reason. Maybe due to this?

```php
class ImageTest extends SapphireTest {
	public function setUp() {
		if(get_class($this) == "ImageTest") $this->skipTest = true;
```

jonom:
Oh I see what I'm doing wrong. Never mind :)

tractorcow:
ImageTest isn't ment to be run except by subclasses. :) It's a dumb way of doing abstract test base classes.

jonom:
I've updated the tests to account for changed output. Still need to add new ones for the new methods but first I'm stumped by the last assertion in `ImageTest.php::testRegenerateImagesWithRenaming()` which is failing
```php
$image_generated_2 = new Image_Cached($newRelative);
$this->assertEquals(300, $image_generated_2->getWidth(), 'Cached image was regenerated with correct width');
```
I can't wrap my head around what this test is supposed to do or why it's failing? The output file is a black (empty) png file 200x200px.

jonom:
Is there something fishy going on here? _resampled folder structure is wrong, maybe that's throwing something off.

screen shot 2015-05-25 at 9 35 10 pm

Maybe it will make sense in the morning.

model/Image.php
```diff
@@ -365,6 +491,26 @@ public function generateStripThumbnail(Image_Backend $backend) {
 	}
 
 	/**
+	 * Resample this Image to ensure quality preference is applied.
+	 * Warning: it's possible this will produce a larger file of lower quality
+	 *
+	 * @return Image
+	 */
+	public function Resampled() {
+		return $this->getFormattedImage('Resampled');
```
jonom: Added a Resampled method for better transparency over using the Resized method

tests/model/ImageTest.php
```diff
@@ -157,55 +172,19 @@ public function testReluctanceToResampling() {
...
-		$imageSetRatioSize = $image->SetRatioSize(300, 600);
-		$this->assertTrue($imageSetRatioSize->isSize(300, 300));
-		$this->assertNotEquals($image->Filename, $imageSetRatioSize->Filename);
-		$imageSetRatioSize = $image->SetRatioSize(600, 300);
-		$this->assertTrue($imageSetRatioSize->isSize(300, 300));
-		$this->assertNotEquals($image->Filename, $imageSetRatioSize->Filename);
-		$imageSetRatioSize = $image->SetRatioSize(300, 300);
-		$this->assertTrue($imageSetRatioSize->isSize(300, 300));
-		$this->assertNotEquals($image->Filename, $imageSetRatioSize->Filename);
+		$this->assertTrue($image->Resampled()->getAbsoluteSize() < $image->getAbsoluteSize());
```
jonom: This is failing right now, something is going wrong with the test images I think at least on my environment

model/Image.php
```diff
@@ -163,6 +163,33 @@ public function getTag() {
...
+	 * the URL will point to an optimised file, if it is smaller than the original
+	 *
+	 * @uses Director::baseURL()
+	 * @return string
+	 */
+	public function getURL() {
+		if ($this->config()->force_resample) {
+			//return $resampled->getURL();
+			$resampled = $this->getFormattedImage('Resampled');
+			if ($resampled->getAbsoluteSize() < $this->getAbsoluteSize()) {
```
jonom: Hazah! Now only serving up the resampled image if it is smaller than the original.

docs/en/02_Developer_Guides/14_Files/01_Image.md
```diff
@@ -18,60 +18,81 @@ Allows upload of images through limiting file extensions with `setAllowedExtensi
...
-		
-		public function generateRotateCounterClockwise(GD $gd)	{
-			return $gd->rotate(270);
-		}
-		
-		public function clearResampledImages()	{
-			$files = glob(Director::baseFolder().'/'.$this->Parent()->Filename."_resampled/*-$this->Name");
-		 	foreach($files as $file) {unlink($file);}
-		}
+	class MyImage extends DataExtension {
```
jonom: Demonstrating an extension instead of sub-class

model/Image.php
```diff
@@ -70,7 +70,7 @@ class Image extends File implements Flushable {
 	 * @config
 	 * @var bool Force all images to resample in all cases
 	 */
-	private static $force_resample = false;
+	private static $force_resample = true;
```
jonom: Can we make this default to true now, particularly now that we're only serving up the resampled version if it is smaller than the original? @hafriedlander suggested it go in to 3.2 at https://github.com/silverstripe/silverstripe-framework/pull/3265#issuecomment-48076794

jonom:
I think this is fairly complete now. I do need some help with working out why that one test is failing, and need some feedback about whether it's appropriate to make `$force_resample` default to `true` for 3.2. Also one missing piece of the `force_resample` puzzle is that it doesn't carry through to images inside HTMLEditorField instances. If someone wants to point me in the right direction to add support for that I can investigate.

kinglozzer:
@jonom I’d guess that the `DataDifferencer` failure is because we’re now forcing resamples - and the original test expected it to *not* be resampled, so it expects the original file name/path to be preserved. Quickest fix is to add before/after:

```php
$originalForceResample = Config::inst()->get('Image', 'force_resample');
Config::inst()->update('Image', 'force_resample', false);
// test assertions
Config::inst()->update('Image', 'force_resample', $originalForceResample);
```

It does bring up an interesting point: we might be breaking people’s tests by forcing resampling during testing. Not sure whether we should set `force_resample: false` for tests, or whether it’s acceptable to potentially break tests in semver. Thoughts @tractorcow?

stevie-mayhew:
@kinglozzer can you please read my comments above regarding SemVer and this change - your question about breaking tests is the exact representation of why I believe this is a BC change and shouldn't be part of 3.2.

kinglozzer:
We don’t (and can’t) make any concrete promises about *exactly* what these methods will return: chaining methods etc will cause different filenames. The only thing we can guarantee in our public API is the type of object we return: an `Image` instance. We don’t promise anything beyond that, and that promise isn’t broken by this change, so in that sense this is okay.

I guess it comes down to weighing up risk for benefit. My thoughts:
- Slim possibility of “public facing” breakages. Not really a major concern IMO: it’s unlikely that any “user” code is checking the exact format of the returned URL, I can’t think of a scenario that this would break something (that’s not to say such a scenario doesn’t exist!). Any code that this change breaks is arguably pretty poorly written, every change/fix we introduce has the possibility of breaking poorly designed code.
- Slightly larger chance of test breakages, as demonstrated here. Not sure what semver says about this, definition of “public API” probably comes into question for this. Again, fairly unlikely to happen but still a possibility. Are test breakages for minor versions acceptable? I don’t see why we should allow possible test breakages though, it’s just an extra upgrade pain.
- If there’s a possibility of BC-breaks, why not put this in master? We’re not fixing a bug here, so we probably should be even more cautions with BC-breaks.

I’ve not read through the changes in detail, but would it be possible to compromise and add the base functionality to 3, even if it’s not as cleanly written, and save the API cleanup/potential BC-breaks for master?

jonom:
Hey guys, was going to suggest the same thing. There's already two separate commits here, and the new Max methods are the substance of this PR - the `force_resample` was a bonus - just seemed like a good time to introduce it.

So if all agreed, I'll remove the force_resample commit from this PR, tweak the Max methods commit to emulate the existing force_resample logic on those new methods, then submit a separate PR against master with the rest of these changes.

Just to draw out this debate *slightly* longer though, I had the impression that you @stevie-mayhew and @hafriedlander both were happy for `force_resample` to default to `true` in 3.2 - which also would have changed output in some cases, returning e.g a `_resampled/SetWidth` image instead of the original image when the argument passed to SetWidth is the same as the original image's width. So in that respect I don't think what I did here is that different from the original plan? Doesn't mean your arguments aren't valid Stevie, maybe just casts a different light on them.

@kinglozzer thanks for your comments on the test breaks - yes I understand why those tests are failing that expect a specific image URL and thought I updated those but I didn't think to look outside of ImageTest.php so I have more work to do.

Actually I've just realised that the test that I was really stumped by is only failing on my own environment :confused: I can't seem to get this testing thing set up out properly. I might just have to push some changes to github and let Travis do the testing.

jonom:
@kinglozzer I think something might be critically broken with the images tests - or my assumptions about them are. On my local machine when I run `phpunit framework/tests/model/GDImageTest.php` The images in the `framework/tests/model/testimages/` directory seem to be ignored. New files are created in `assets/ImageTest/` for each of the files listed in `ImageTest.yml` but they are all solid black and measure 300x300px unlike the images in `framework/tests/model/testimages/`. Correcting the path to the images in ImageTest.yml or putting the correct files in the assets directory before the test is run has no effect - always end up with black images.

Some of the tests I need to introduce need to compare the file size of a real image compressed at different levels (I added a couple for this purpose), but it doesn't look to be possible with the current setup. Sorry for my lack of understanding in this area but can you give me any tips on getting these tests to actually use the test image files instead of creating new ones?

Note: I had to get rid of the teardown() function in Image.php to see what the image file output actually was.

jonom:
I've been pouring over the Image tests and I eventually found:

* The images in the `framework/tests/model/testimages/` are never used. References are only namesakes.
* `Image::regenerateFormattedImages()` needs a tweak to be compatible with the Flushable updates, which is why some tests were failing on my computer.

I've updated the tests on this PR to actually make use of the test images and fix the flush issue.

jonom:
Tests are finally passing and while I think both @stevie-mayhew and @tractorcow make valid points about what makes a breaking change I think this is actually safe to merge in to 3.

If this is a breaking change it's beside the point, as [this commit](https://github.com/silverstripe/silverstripe-framework/commit/5fbb3da661248d277af78073c12adae1ddf93808) introduced 2 years ago :open_mouth: already changed (broke) the filename formats for *every* resampled image because it encodes the arguments in the filename.

Let me know your thoughts and if any more changes are required.

stevie-mayhew:
To jump back in - I'm not really worried about the changes in this specific example and thing that it is definitely a really awesome change to behaviour that needed change.

However if something changes anywhere in a public API and we are following SemVer then it shouldn't go into a minor version. Its a really hard rope to walk when a change may seem inconsequential - but the point of following something like SemVer is that it could *potentially* not be.

We can't rely on saying "this previous thing did this" as we move forward actually using semantic versioning properls. It means that as a developer I should be able to rely on code and results being the same AND correct for all minor versions of code. 

Its also a hairy issue at the moment with the changes from 3.1 -> 3.2 not being strictly SemVer. Once 4.0 is out I'd like to see anything which is even debatable as whether it fits into a minor version API as a change be forced up to master.

Awesome work on this @jonom, its good to see some areas which are often overlooked getting some great loving, especially your work to clean up the tests! 

Looking forward to perhaps some focus point getting into master at some point too :wink:

jonom:
@stevie-mayhew yeah I agree with your thinking - if there's any reason to think a change could be breaking then it should go in to the next major version.

I'm really hoping that with the move to SemVer we  completely change our mindset and expectations about what a major version number means, and commit to major versions coming out every 6 months or so. If there are changes to release, let's do it, they don't need to be revolutionary and there doesn't have to be a whole bunch of them. Minor releases shouldn't need to follow any particular schedule, if there's a cool new feature to release that doesn't break anything let's get it out as quickly as possible! We're used to apps on our phones updating themselves constantly, as soon as features are ready - and I would love to see the same approach to minor releases.

I also think that if SemVer is followed  we need to change the way we think about patches. If SilverStripe Framework version 5.1.0 is released but contains a bug, then 5.2.0 comes out with some new features plus a patch for that bug, **there is no need for a 5.1.1 release that also fixes that bug.** If SemVer is followed then 5.2.0 provides a clean upgrade path from 5.1.0, and that is how everyone should patch the bug. If we try to maintain the current patching approach it will discourage frequent minor releases because each new minor release would mean more branches that need to be actively maintained when bugs are fixed. This should mean going forward there is only a single branch for each major version of SilverStripe and just tags for minor release and patches e.g. 3.2.0, 3.2.1, 3.3.0 etc. . I think this is actually how SemVer is supposed to be followed and is probably already be the plan so sorry if I'm just stating the obvious here!

I don't know what the realities of a possible SemVer release schedule are for a complex system like SilverStripe but that's what I'm hoping for. And as long as we're seeing far more frequent major releases, these kinds of debates shouldn't be necessary as 'merging in to master' won't mean 'you will be lucky to see this feature released in your natural lifetime'. It will just mean you might have to wait a few extra months.

Anyway all that aside, I think it's clear that 3.2 doesn't strictly follow SemVer protocols, and this change isn't a breaking one when considered alongside what's already been merged in. So I remain hopeful that this change makes it in to 3.2 :)

jonom:
I'm thinking the existing image function names as well as the ones I've added could be more intuitive. What do you guys think about revising all the names for more clarity and consistency overall? We could alias the old names of existing functions so they continue to work.

New name | Current name | Action
----|----|---
**Fit** | SetRatioSize | Scale image proportionally to fit within bounds
**FitMax** | ScaleMax | As above but without up-sampling
**ScaleWidth** | SetWidth | Scale image proportionally by width
**ScaleMaxWidth** | ScaleMaxWidth | As above but without up-sampling
**ScaleHeight** | SetHeight | Scale image proportionally by height
**ScaleMaxHeight** | ScaleMaxHeight | As above but without up-sampling
**Crop** | CroppedImage | Crop image to specified dimensions
**CropMax** | CropMax | As above but without up-sampling. Resulting image has the same aspect ratio as CroppedImage but may have a lower resolution.
**ClipWidth** | CropMaxWidth | Crop away left and right sides if image exceeds width
**ClipHeight** | CropMaxHeight | Crop away top and bottom if image exceeds height
**Pad** | PaddedImage | Fit image to dimensions and fill leftover space with a solid colour (default white)

Also if this PR needs sweetening to make that merge button more attractive I could probably introduce a couple of other useful functions to work with both Image backends:

* Greyscale
* Trim (Trim away a padded background of specified colour, default white. Good for logos.)

stevie-mayhew:
@jonom - Greyscale in core would be good IMHO, we write that function constantly. (well, copy/paste)

model/Image.php
tractorcow: This is my highly opinionated opinion, but this essentially makes all size checks to PREVENT unnecessary respampling essentially voided. I understand Hamish's point that clients could upload unoptimised graphics, but they could also upload perfectly optimised graphics that deteriorate through respampling. It essentially undoes a lot of the work that was done in the past to prevent this happening.

See https://github.com/silverstripe/silverstripe-framework/pull/1318

I would prefer this was off by default (and stayed off).

model/Image.php
```diff
@@ -328,6 +355,105 @@ public function generateSetSize(Image_Backend $backend, $width, $height) {
...
+	 * Proportionally scale down this image if it is wider or taller than the specified dimensions.
+	 * Similar to SetRatioSize but without up-sampling. Use in templates with $ScaleMax.
+	 * 
+	 * @uses Image::SetRatioSize()
+	 * @param integer $width The maximum width of the output image
+	 * @param integer $height The maximum height of the output image
+	 * @return Image
+	 */
+	public function ScaleMax($width, $height) {
+		return $this->width > $width || $this->height > $height
```
tractorcow: Can you please use the `->getWidth()` and `->getHeight()` methods for the sake of code completion?

Also another discussion we had earlier in another issue, but we feel that getter methods are best invoked explicitly where available. :D

tractorcow:
Can I suggest:



New name | Current name | Action
----|----|---
**Fit** :+1:  | SetRatioSize | Scale image proportionally to fit within bounds
**FitMax** :+1:  | ScaleMax | As above but without up-sampling
**ScaleWidth** :+1:  | SetWidth | Scale image proportionally by width
**ScaleMaxWidth** :+1:  | ScaleMaxWidth | As above but without up-sampling
**ScaleHeight** :+1:  | SetHeight | Scale image proportionally by height
**ScaleMaxHeight** :+1:  | ScaleMaxHeight | As above but without up-sampling
**Crop** :+1:  | CroppedImage | Crop image to specified dimensions
**CropMax** :+1:  | CropMax | As above but without up-sampling. Resulting image has the same aspect ratio as CroppedImage but may have a lower resolution.
**CropWidth** :question:  (should be keep the terminology consistent?)  | CropMaxWidth | Crop away left and right sides if image exceeds width
**CropHeight** :question:  (same) | CropMaxHeight | Crop away top and bottom if image exceeds height
**Pad** :+1:  | PaddedImage | Fit image to dimensions and fill leftover space with a solid colour (default white)


I just about feel we need an ImageTransformation class so that we don't need to fill Image.php with all these use cases.... but that can be a 4.0 update. :D

model/Image.php
tractorcow: I don't get why we don't need the $imageRatio < $cropRatio here.... maybe you can look at how I'd implement this to tell me what I've got wrong?

```php
public function CropMax($width, $height) {
		// Prevent divide by zero on missing/blank file
		if(empty($this->getWidth()) || empty($this->getHeight())) {
			return null;
		}
		
		// Is the image already the correct size?
		if ($this->isSize($width, $height)) {
			return $this;
		}
		
		// If not, make sure the image isn't upsampled
		$imageRatio = $this->width / $this->height;
		$cropRatio = $width / $height;
		
		// Detect if we should crop on the X axis
		if ($cropRatio < $imageRatio && $this->height < $height) {
			return $this->CroppedImage($this->height * $cropRatio, $this->height);
		}
		
		// Detect if we should crop on the Y axes
		if ($cropRatio > $imageRatio && $this->width < $width) {
			return $this->CroppedImage($this->width, $this->width/$cropRatio);
		}
		
		// Crop on both X and Y axes
		return $this->CroppedImage($width, $height);
	}
```

model/Image.php
jonom: @tractorcow I included a [filesize check](https://github.com/silverstripe/silverstripe-framework/pull/4229/files#r31054055) so that the resampled one would only be served up if it's smaller than the original, do you think that won't be enough to prevent deterioration?

tractorcow:
Ok @jonom I gave it a nice thorough review... do you want to have a look through my feedback and let me know what you think? (especially https://github.com/silverstripe/silverstripe-framework/pull/4229#discussion_r31980301 which makes the math part of my brain hurt).

model/Image.php
jonom: Ergh... Aargh... My brain is hurting too. Line 422 catches images that are going to be cropped on the x axis but are undersize and changes the resize arguments to suit. Line 424 catches images that are going to be cropped on the y axis AND images that aren't going to be cropped at all but are undersize. Checking the width only is sufficient at this point to cover both cases because we already know that `$cropRatio>= $imageRatio`. If the image isn't going to be cropped (the `$imageRatio==$cropRatio` scenario) it doesn't matter if we look at width or height, either one can tell us if the image is undersize so that is covered. So... I'm 96% confident the code doesn't need changing :) but, *sigh*, maybe we need more unit tests.

model/Image.php
sminnee: There's an analogy in RESTful APIs that pass around URLs in href attributes from responses. Is the API allowed to change the URL form, and update the href attribute, and consider it a non-breaking change? That would probably come down to:
 
 * If the specific URL patterns were documented, then a change to the URL scheme would be breaking a public API, and wouldn't be appropriate without a major release bump.
 * If the URL patterns were not documented, and instead the API docs "make this request and then fetch an href from it", then a change to the URL scheme would be only breaking an internal API, and it probably can be done in a minor release bump.

So, it comes down to what's documented. Easy! 

Well, not so fast. In SilverStripe unfortunately we have a lot of things that are important not to break, and aren't necessarily documented. So saying "we might change anything not specifically spelled out in the documentation" is likely to ignore more people that is helpful.

Note that another way to deal with this would be negative documentation: something like "please call Link() methods and don't try and construct the URL yourself".

So, right now I don't think we're going to get unambiguous answers for this stuff, and we're going to need to make some assumptions over which undocumented APIs are still commonly-used enough to need to be treated as "public".

On the flipside, it would be great to push for more complete API documentation as we inch towards 4.0 (and, even better, tests that cover the same set of cases).

In the short term I would probably err on the side of conservatism for 3.x, since we don't know which features people are relying on.

sminnee: So, to practise what I preach on this one specific example...

https://github.com/silverstripe/silverstripe-framework/pull/4263

sminnee:
Hey, given the big table of method name changes, maybe we should just shift this to 4.0 and cease the guess-work about whether this breaks APIs?

jonom:
Hey @tractorcow thanks for the review and sorry for the brain pain. Have updated all the functions (including existing) to use getter functions for width and height and will push later after we've settled on function names.

Scale would be a misleading name for SetSize because SetSize is actually just PaddedImage but without the ability to choose the colour - it's always white. Since PaddedImage uses white as a default I don't think there's any need to rename SetSize as I think it's a redundant function - just use Pad without the third argument ;)

If anything should be renamed Scale it should be ResizedImage - but I don't think this function should ever be used as it would tend to create a skewed image and I can't think of any situation where it should be used instead of Fit/ScaleWidth/ScaleHeight which all provide proportionately scaled images. So while it's harmless to leave ResizedImage where it is I don't think we need to encourage it's use by providing a nicer name for it and acknowledging it in the docs.

As for CropMaxWidth and CropMaxHeight I did uhm and ah about that one for a while... but I felt that the behaviour was different enough from Crop that it warranted a different name. Crop and CorpMax provide an image of a known aspect ratio and performs a resize to retain as much of the image as possible. ClipWidth and ClipHeight don't do any resizing so the output will change dramatically depending on what functions you apply to an image before you pass it to these two functions - and they're the only functions that will normally *require* that you chain in another image function first so that you're not working on an original image which could give you an image of ridiculous proportions. If the functions are consistent in name but don't behave consistently it could cause confusion. Also I've used Max in other function names to mean 'like the other one but not upsampled' so I want to be consistent with that too, and it would have a different meaning if used here.

Have I swayed you?

jonom:
@sminnee this did get a bit out of hand didn't it? Scope creep! 

Is there any vague release window on 4.0? I see where you're coming from of course but I would rather get the safe parts of this in to 3.2 and any potentially breaking changes in to 4.0 so we have the option of using these functions sooner rather than later - especially if 4 is a couple of years away. I think renaming functions should be safe for a minor release though as long as it's backwards compatible? Actually I checked:

> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. 
http://semver.org/ (point 7)

 In fact I think the only thing here that's debatable BC-wise is whether or not `$force_resample` should default to true in 3.2, since I've pointed out that all resampled images are already undergoing a major filename change in 3.2 due to encoded arguments.

model/Image.php
jonom: Ha, nice update! :) think I was typing a response to your first comment while you wrote this one. Everything you said here makes perfect sense - but I still think that it doesn't apply to this specific case because we're wondering whether it matters if the resampled image filenames change in a few cases, when already *all* the resampled image filenames are being changed in 3.2 before this PR came along. Now I must go to bed :sleeping:

tractorcow:
We could probably just drop the name changes and focus on getting the core features into 3.2, if you like?

We can always re-name them in 4.0, which is going to be a while away yet.

jonom:
I don't think I'm convincing enough to get this all in to 3.2 :) so will split it up. Can you guys just help me work out what should go where and address any remaining discussion points. I have cheekily left the renaming part in because I think it's a harmless but beneficial change, especially for developers who are new to SilverStripe.

  | Change | Target | Discussion points
---|---|---|---
1 | Add new Image functions | 3.2 |
2 | Rename/alias existing image functions | 3.2 | SemVer states that this is appropriate for a minor release and I think it makes sense to do this sooner rather than later. Pending more feedback from @tractorcow (or other interested devs) on the suitability of the names I chose. If more time is needed to reach a consensus on names then I guess this should wait until 4.0.
3 | Improve image tests | 3.2 |
4 | Update Image docs | 3.2 | 
5 | Separate force resampling logic in to getURL() function | 4.0 | Since this can change the expected filename output @stevie-mayhew and @sminnee feel this should normally wait for a major release and I agree - however I think in this case it wouldn't be harmful for this to go in to 3.2 because filename output is already changing a lot.
6 | `$force_resample` default to true | 4.0 | @tractorcow against due to it making anti-resampling checks redundant but I think the [filesize check](https://github.com/silverstripe/silverstripe-framework/pull/4229/files#r31054055) should provide a nice solution here?

Seek your feedback on these targets and any remaining discussion points and then I'll reduce the scope of this PR and open a separate one for master.

tractorcow:
I think that is more or less fine with me, although the filesize check only protects against large image sizes, not degraded image quality.

jonom:
You're right, but I think as a best practice we should be encouraging people to upload images with close to zero compression to provide a clean and flexible base for resampling. I updated the docs a little as part of this but maybe could go further? It should probably be in the user docs as well as the developer docs. I also would like HtmlEditorField to respect $force_resample in the same way so maybe I should try to add that at the same time now that this part is slated for 4.0.

tractorcow:
3.2.0 beta 1 is coming out soon, so if you'd like this in 3.2 I hope we can finish this soon. ;D

tractorcow:
I like the idea of leaving up the image-value decisions up to the devs; Whether they want to encourage already-optimised images to be uploaded or only rely on the system to upload is a decision they should make on their project.

jonom:
True, so I guess developer docs for `$force_resample` should say something like 'If you set this to true make sure to instruct CMS users to upload hiqh quality images (minimal or zero compression)'.

I will strip this PR back for 3.2 tomorrow or could probably do it tonight if you need it sooner.

jonom:
Phew, okay this has been revised and rebased to remove the force_resample changes. I've tried to support the current way of handling force_resampling without modifying any existing lines so we can have a clean merge between 3.2 and 4. @tractorcow do you want me to open a separate PR now for 4 that includes these commits or do you want to merge 3.2 up and I can follow up with the force_resample additions later?

Also please note that I decided at the last minute that `Fill` is a more descriptive name than `Crop` because the function is limited in that all you can do is 'fill' the given dimensions with as much image as possible, you can't zoom in on a smaller area. This leaves the function name available for a more robust Crop function to be added later.

jonom:
Oh.. I updated the code to use getters but apparently empty() [has to be passed a variable](http://stackoverflow.com/questions/19175675/php-class-fatal-error-cant-use-method-return-value-in-write-context) prior to PHP  5.5 so tests are failing here. Will revise.

jonom:
This should be good to merge now I hope. I've separated the commits for each feature to make it easier to review and if there's any doubt about the new names I've chosen I can remove that commit so this can make it in to 3.2 and we can have more time to decide on the best names for these functions going forward.

docs/en/02_Developer_Guides/14_Files/01_Image.md
```diff
@@ -18,60 +18,81 @@ Allows upload of images through limiting file extensions with `setAllowedExtensi
...
 	:::php
-	// manipulation functions
-	$image->resize(width,height); // Basic resize, just skews the image
-	$image->resizeRatio(width,height) // Resizes an image with max width and height
-	$image->paddedResize(width,height) // Adds padding after resizing to width or height.
-	$image->croppedImage(width,height) // Crops the image from the centre, to given values.
-	$image->resizeByHeight(height) // Maximum height the image resizes to, keeps proportion
-	$image->resizeByWidth(width) // Maximum width the image resizes to, keeps proportion 
-	$image->greyscale(r,g,b) // alters image channels ===
+	// Scaling functions
```
tractorcow: This isn't actually `:::php` format code, so do you want to change the doccode to `:::ss`?

jonom: Good point! will do

docs/en/02_Developer_Guides/14_Files/01_Image.md
```diff
@@ -18,60 +18,81 @@ Allows upload of images through limiting file extensions with `setAllowedExtensi
...
+	$Image.ScaleWidth(150) // Returns a 150x75px image
+	$Image.ScaleMaxWidth(100) // Returns a 100x50px image (like ScaleWidth but prevents up-sampling)
+	$Image.ScaleHeight(150) // Returns a 300x150px image (up-sampled. Try to avoid doing this)
+	$Image.ScaleMaxHeight(150) // Returns a 200x100px image (like ScaleHeight but prevents up-sampling)
+	$Image.Fit(300,300) // Returns an image that fits within a 300x300px boundary, resulting in a 300x150px image (up-sampled)
+	$Image.FitMax(300,300) // Returns a 200x100px image (like Fit but prevents up-sampling)
+	
+	// Cropping functions
+	$Image.Fill(150,150) // Returns a 150x150px image resized and cropped to fill specified dimensions (up-sampled)
+	$Image.FillMax(150,150) // Returns a 100x100px image (like Fill but prevents up-sampling)
```
tractorcow: I'm actually not so sure about this method; Why not fill up to 150 x 150? It seems like this is a case of adding a 'Max' method simply for the sake of it. I'd opt to just drop it.

tractorcow:
> As for CropMaxWidth and CropMaxHeight I did uhm and ah about that one for a while... but I felt that the behaviour was different enough from Crop that it warranted a different name. Crop and CorpMax provide an image of a known aspect ratio and performs a resize to retain as much of the image as possible. ClipWidth and ClipHeight don't do any resizing so the output will change dramatically depending on what functions you apply to an image before you pass it to these two functions - and they're the only functions that will normally require that you chain in another image function first so that you're not working on an original image which could give you an image of ridiculous proportions. If the functions are consistent in name but don't behave consistently it could cause confusion. Also I've used Max in other function names to mean 'like the other one but not upsampled' so I want to be consistent with that too, and it would have a different meaning if used here.


I'm not sure that, as a user, I could get all of that information by looking at the names 'Clip' and 'Crop', so I'm still not sure the value we're expecting from this distinction will actually make more sense.

docs/en/02_Developer_Guides/14_Files/01_Image.md
jonom: I definitely think this is a keeper, because what if you request a Fill(2000,1000) on an image that is only 500x300px? You get a 2000x1000px image that has been greatly upsampled, where as FillMax will give you a 500x350px image.

tractorcow: I thought we were going to keep using `Pad` to be consistent with 3.1.

jonom:
Well since I ended up getting rid of Crop and renaming it to Fill, there are now no functions presently that include Crop in the name. Assuming you're happy to rename CroppedImage to Fill, how about ClipWidth becomes CropWidth and ClipHeight becomes CropHeight?

docs/en/02_Developer_Guides/14_Files/01_Image.md
```diff
@@ -18,60 +18,81 @@ Allows upload of images through limiting file extensions with `setAllowedExtensi
...
+	$Image.Width // Returns width of image
+	$Image.Height // Returns height of image
+	$Image.Orientation // Returns Orientation
+	$Image.Title // Returns the friendly file name
+	$Image.Name // Returns the actual file name
+	$Image.FileName // Returns the actual file name including directory path from web root
+	$Image.Link // Returns relative URL path to image
+	$Image.AbsoluteLink // Returns absolute URL path to image
+
+Image methods are chainable. Example: `<body style="background-image:url($Image.ScaleWidth(800).ClipHeight(800).Link)">`
```
tractorcow: Just a question; Have you checked that the generated URL for these chained methods contains the operations and doesn't just cache on whatever the last operation was? Eg.. `_resampled/setwidth400_setheight200_FileName.jpg`

model/Image.php
```diff
@@ -215,22 +215,25 @@ public function loadUploadedImage($tmpFile) {
 		}
 	}
 
+	public function SetRatioSize($width, $height) {
```
tractorcow: Please also add `@deprecated` docblock for these.

model/Image.php
```diff
@@ -240,77 +243,87 @@ public function SetRatioSize($width, $height) {
...
-	 * Resize the image by preserving aspect ratio, keeping the image inside the
-	 * $width and $height
+	 * Scale image proportionally to fit within the specified bounds
 	 *
 	 * @param Image_Backend $backend
 	 * @param integer $width The width to size within
 	 * @param integer $height The height to size within
 	 * @return Image_Backend
 	 */
-	public function generateSetRatioSize(Image_Backend $backend, $width, $height) {
```
tractorcow: You can't remove this method without deprecated it; You need to deprecate BOTH `Method` and `generateMethod`.

jonom:
In the future I would see us adding a function like this `Crop($x1,$y1,$x2,$y2)` which lets you crop an image any way you like. You could zoom in and crop a headshot out of a larger image with that function, then chain in a ScaleWidth or other function to make it the size you want.

In fact I guess I could add that one to this PR :P but with all the bug squishing going on I'm guessing a 3.2 release is imminent so I wouldn't get it in in time. If this PR is becoming a pain I can target 4.0 instead, don't want to rush you in to any decisions that will stick around for a long time.

model/Image.php
jonom: Oh sorry. I was thinking it isn't relevant anymore but it's still part of the API isn't it. Noted.

model/Image.php
```diff
@@ -328,6 +341,123 @@ public function generateSetSize(Image_Backend $backend, $width, $height) {
...
+		// Is the image already the correct size?
+		if ($this->isSize($width, $height)) return $this;
+		
+		// If not, make sure the image isn't upsampled
+		$imageRatio = $this->getWidth() / $this->getHeight();
+		$cropRatio = $width / $height;
+		// If cropping on the x axis compare heights
+		if ($cropRatio < $imageRatio && $this->getHeight() < $height) return $this->Fill($this->getHeight()*$cropRatio, $this->getHeight());
+		// Otherwise we're cropping on the y axis (or not cropping at all) so compare widths
+		if ($this->getWidth() < $width) return $this->Fill($this->getWidth(), $this->getWidth()/$cropRatio);
```
tractorcow: Resurrecting our discussion from https://github.com/silverstripe/silverstripe-framework/pull/4229#discussion_r31980301

You said

> Ergh... Aargh... My brain is hurting too. Line 422 catches images that are going to be cropped on the x axis but are undersize and changes the resize arguments to suit. Line 424 catches images that are going to be cropped on the y axis AND images that aren't going to be cropped at all but are undersize. Checking the width only is sufficient at this point to cover both remaining cases because if the image isn't going to be cropped (the $imageRatio==$cropRatio scenario) it doesn't matter if we look at width or height, either one can tell us if the image is undersize. So... I'm 96% confident the code doesn't need changing :) but, sigh, maybe we need more unit tests.

Can you show me where `line 422` is in the rebased code so I can review?

docs/en/02_Developer_Guides/14_Files/01_Image.md
jonom: It sure does! Which is pretty cool. Does lead to long filenames though.

model/Image.php
jonom: Will do

model/Image.php
tractorcow: Ok I guess your logic is fine ;d Don't worry about it.

jonom: Yes that is now at 420.

jonom: Comment timing fail :)

tractorcow:
> Well since I ended up getting rid of Crop 

I don't see much of a benefit in Crop > Fill or Crop > Clip in general. We should probably not change a name unless we have a good reason. I'm happy to hear your argument though. :)

tractorcow:
Sorry for always critiquing this PR. That's my way of saying I really like it and want it merged. :)

jonom:
Haha thanks :) I was worried it was becoming a thorn in your side

In hindsight it's painfully obvious that it's easier to show an image than describe it - this should better explain the different methods and hopefully why the names are appropriate:

image-method-diagram

I can put something like this in the docs too, should explain things a lot better. Without Crop though because it's not implemented yet.

Think you can see the difference between what I think a 'true' crop function should do compared to what CroppedImage currently does. Here are a couple of good arguments for all of these new names:

- Shorter: easier to remember, easier to type, generates shorter filenames
- More descriptive of actual function and consistent with common image-editing terminology
- Combine these two points and result is it's easier to grasp for newbies

jonom:
p.s. for this theoretical Crop method I think it might be best to use percentages actually like `Crop(.2,.2,.8,.8)` since then it's easier to work with an image that's been scaled. But could maybe give the option of passing pixel values instead `Crop(50,50,100,100,$pixelArgs = true)`

tractorcow:
I actually really find a lot of value in that kind of screenshot for the docs; It makes review a lot easier, although I understand if it's a lot of work to maintain (so you can save it until we lockdown the API).

I think we should focus on implementing what we have, instead of trying to let this PR grow too far; The original intention is to support MaxWidth / MaxHeight right? Maybe we should do Crop in 3.3?

Discussion with smarter colleagues than I agree that 'Fill' is fine. I'll adjust my thinking appropriately.

tractorcow:
I think I'm happy with the API... let's just get this out there right?

Last few things to tidy up:
- Would like a screenshot for the 01_Image.md, but I won't block merge if it's a pain to do.
- Fix the `:::php` on the markdown.
- squash / rebase (lots of stuff going into 3 branch at the moment so I'll try to merge quick)
- fix deprecation notices in phpdoc
- Add upgrading notes to 3.2.0.md so that users upgrading their templates know which image methods to use.

I think that's about it eh?

Thanks for bearing with me and sorry for making it hard. I'm far too stubborn for my own good. ;D

jonom:
'kay have done that and also moved the content of the Files index page in the docs in to a sub page because it was causing the Image docs to be kind of hidden.

tractorcow:
Goal!

stevie-mayhew:
Awesome work guys, really great :dancer:

jonom:
Boom! Thanks for working through it with me guys, good learning experience for me

jonom:
@tractorcow I'd be happy to wait until this is merged up to master before submitting those force resampling changes in a separate PR if that suits you?

tractorcow:
Sure, we'll do that some time around 3.2 beta is released

shot131:
Is where any suggestion for PadMax method (Pad without upsampling)?

jonom:
It could be added - just needs a PR :)

dhensby:
We are now using `InterventionImage` as our image backend so it's much easier to add new methods you want

PR #4232, opened 2015-05-27
Semantic existing upload renaming

jonom:
Filename          | Old result     | New result
------------------|----------------|-----------
image.jpg         | image2.jpg     | image-v2.jpg
55.jpg            | 56.jpg         | 55-v2.jpg
Image-2402.jpg    | Image-2403.jpg | Image-2402-v2.jpg
Image-2402-v2.jpg | -NA-           | Image-2402-v3.jpg

Numbers are often used at the end of file names in image sequences (e.g. digital photos) and the current renaming scheme can disrupt this, creating a confusing situation for the user. Presence of '-v' flag should help reduce this.

### DIFF
--- a/filesystem/Upload.php
+++ b/filesystem/Upload.php
@@ -157,7 +157,7 @@ class Upload extends Controller {
 			}
 		}
 
-		// if filename already exists, version the filename (e.g. test.gif to test2.gif, test2.gif to test3.gif)
+		// if filename already exists, version the filename (e.g. test.gif to test-2.gif, test-2.gif to test-3.gif)
 		if(!$this->replaceFile) {
 			$fileSuffixArray = explode('.', $fileName);
 			$fileTitle = array_shift($fileSuffixArray);
@@ -175,11 +175,11 @@ class Upload extends Controller {
 				$i = isset($i) ? ($i+1) : 2;
 				$oldFilePath = $relativeFilePath;
 
-				$pattern = '/([0-9]+$)/';
-				if(preg_match($pattern, $fileTitle)) {
-					$fileTitle = preg_replace($pattern, $i, $fileTitle);
+				$pattern = '/-v([0-9]+$)/';
+				if(preg_match($pattern, $fileTitle, $matches)) {
+					$fileTitle = preg_replace($pattern, '-v' . ($matches[1] + 1), $fileTitle);
 				} else {
-					$fileTitle .= $i;
+					$fileTitle .= '-v' . $i;
 				}
 				$relativeFilePath = $relativeFolderPath . $fileTitle . $fileSuffix;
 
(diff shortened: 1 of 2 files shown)


jonom:
Would this be a `FIX` or a `NEW` or no label needed?

tractorcow:
Hm, this is a good fix, but I wonder if we should put this into 4.0 since it could end up breaking other file cleanup code that already exists.

I would say it's a change that fixes an issue, so FIX?

I wonder if it would be awesome to put the capture into a configurable regexp, rather than enforcing {name}-v{number}.{extension} format?

jonom:
Uhhh Yes. Whoops. I meant to send this to 4 after our discussion about SemVer but I must have got confused along the way. Will open a new PR.

Sorry you lost me on that last comment :) what does that mean / how would that look? Would it be like a config variable for the `v-` part or something?

jonom:
Looks like I've got a few more tests to update too.

tractorcow:
What I mean is, maybe make the `$pattern` configurable, also the replacement. My reasoning is that `-v(num)` seems an awfully specific format.

jonom:
So like this?
```php
private static $file_version_pattern = '/-v([0-9]+$)/';
```
I'm not sure how to make the replacement configurable in the same way since I don't see how preg_replace could increase the counter with just a pattern... do you know how? Plus it seems a bit like overkill to me, wouldn't this be enough?
```php
private static $version_prefix = '-v';
```
Then you could make it whatever you like i.e. `_version-` to make `image.jpg` --> `image_version-2.jpg`.

... Convention over configuration? :grinning:

tractorcow:
Sounds good :D Maybe just make the prefix configurable for simplicity?

jonom:
Will do ;)

jonom:
You know, considering this is configurable, it *could* go in to 3.2 as a non-breaking change with `$version_prefix = ''`. Then there will be no change to behaviour by default, but people could enable the functionality if they want. Then for master we could set the default prefix to `-v`. What do you think?

tractorcow:
Yes that's absolutely right. :)

jonom:
Okay if I reopen this with `$version_prefix = ''`should I open a separate pull request against master with the same commit + another commit changing `$version_prefix = '-v'`? Or send as a single commit? Or would I need to wait for the commit against 3 to be merged into master before submitting a PR for that second commit? Not sure how the workflow goes for merging between branches...

jonom:
Apparently if you force push you can no longer reopen a PR - so I've made a new one at https://github.com/silverstripe/silverstripe-framework/pull/4241

@tractorcow appreciate your input on how I should approach changing the default for 4. Is it safe for me to open a separate PR for master that includes all these changes + more or will that cause merge conflicts later? Thanks

tractorcow:
Probably best to open separate PR for each branch. Just make sure the code is exactly the same, so that merging is simple once we come to do it (it should automatically resolve conflicts).

Add a note (comment) in the code to point out the part that differs between the two PR to help when merging up.

jonom:
Here is the master version: https://github.com/silverstripe/silverstripe-framework/pull/4251

Rather than introduce a code comment to linger in master I put one in the 3.1 version.

PR #4233, opened 2015-05-28
Fix merge / test regressions

tractorcow:


tests/control/DirectorTest.php
```diff
@@ -402,6 +402,9 @@ public function testUnmatchedRequestReturns404() {
 	}
 
 	public function testIsHttps() {
+		if(!TRUSTED_PROXY) {
```
tractorcow: See https://github.com/silverstripe-labs/silverstripe-travis-support/commit/09db98db8d141eb5ee23da2a6c4d124461e675b5

thirdparty/simpletest/url.php
```diff
@@ -419,13 +419,16 @@ function asString() {
             $scheme .= "://";
             $host = $this->getHost();
         }
+		if ($this->getPort() && $this->getPort() != 80 ) {
```
tractorcow: This was backported from the newer version of SimpleTest module.

tests/control/DirectorTest.php
hafriedlander: Seems like the TestSkipped string is backwards? !TRUSTED_PROXY means untrusted proxy, but test says test is skipped if !untrusted proxy.

tractorcow: Fixed

PR #4234, opened 2015-05-28
DOCS Fixed code example for Requirements::set_write_js_to_body call

michalkleiner:
Fixed code example for setting write_js_to_body value to false


tractorcow:
Thanks @michalkleiner good job spotting that. :)

PR #4236, opened 2015-05-28
FIX Allow users to specify allowed hosts

nyeholt:
To test, add `define('SS_ALLOWED_HOSTS', 'devhost.com,localhost');` to your _ss_environment.php

Make requests from curl like

`$ curl -I -H "Host: haxed.com" -vv http://localhost/site`

You should have a `HTTP/1.1 400 Invalid Host` error

### DIFF
--- a/core/Constants.php
+++ b/core/Constants.php
@@ -181,6 +181,14 @@ if(!isset($_SERVER['HTTP_HOST'])) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


tractorcow:
I agree with @dhensby ; Let's skip this check if CLI, but otherwise if HTTP_HOST is absent we should also fail with a 400 error.

PR #4237, opened 2015-05-28
updated field name for caption text field.

willmorgan:


willmorgan:
Build is likely failing due to the `dev-` prefix of the branch. Yay composer!

PR #4238, opened 2015-05-28
FIX Incorrect env setting in 3.1.13

tractorcow:
See https://github.com/silverstripe/silverstripe-installer/pull/91

This is why we shouldn't do same-day releases for security fixes or changes.


tractorcow:
We'll probably resolve this in a 3.1.14 release next week.

tractorcow:
In retrospect a rushed 3.1.14 would probably do more harm that good at this stage; I'll re-open against 3.1 branch and remove the temp branches.

PR #4239, opened 2015-05-29
Add defense against SS-2015-013 to nginx example

oddnoc:
The advisory provided a patch for .htaccess, but no information for nginx. This adds an nginx defense.


tractorcow:
Thanks @oddnoc :)

PR #4241, opened 2015-05-29
NEW: Configurable file version prefix

jonom:
Numbers are often used at the end of file names in image sequences (e.g. digital photos) and the current renaming scheme can disrupt this, creating a confusing situation for the user. A configurable file version prefix should help prevent this, with a default of '-v' proposed for SS4.0. Not making that change in this PR as it may be a breaking one.

Filename   | Old result | New result (with '-v' prefix)
-----------|------------|--------------
IMG001.jpg | IMG2.jpg   | IMG001-v2.jpg

### DIFF
--- a/filesystem/Upload.php
+++ b/filesystem/Upload.php
@@ -72,6 +72,17 @@ class Upload extends Controller {
 	 */
 	private static $uploads_folder = "Uploads";
 
+	/**
+	 * A prefix for the version number added to an uploaded file
+	 * when a file with the same name already exists.
+	 * Example using no prefix: IMG001.jpg becomes IMG2.jpg
+	 * Example using '-v' prefix: IMG001.jpg becomes IMG001-v2.jpg
+	 *
+	 * @config
+	 * @var string
+	 */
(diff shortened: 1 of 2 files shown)


filesystem/Upload.php
```diff
@@ -175,11 +186,12 @@ public function load($tmpFile, $folderPath = false) {
 				$i = isset($i) ? ($i+1) : 2;
 				$oldFilePath = $relativeFilePath;
 
-				$pattern = '/([0-9]+$)/';
-				if(preg_match($pattern, $fileTitle)) {
-					$fileTitle = preg_replace($pattern, $i, $fileTitle);
+				$prefix = $this->config()->version_prefix;
```
tractorcow: Be careful if the prefix contains regex characters. E.g. `'.ver'`. Maybe preg_quote this first?

jonom: Good point, will do

tractorcow:
I think it would be good to have a unit test to prove this function works. ;)

jonom:
That's true. I'll write one tomorrow.

PR #4245, opened 2015-06-01
FIX Incorrect env setting in 3.1.13

tractorcow:
Re-opened PR for https://github.com/silverstripe/silverstripe-framework/pull/4238 against 3.1


tractorcow:
See https://github.com/silverstripe/silverstripe-installer/pull/92 for installer PR

sminnee:
Cool, so this leaves the setting open-ended in the implementation but updates the docs to match 3.1.13. I think that's the right move.

tractorcow:
Thanks, should we merge this one now then?

tractorcow:
Thanks :D

PR #4250, opened 2015-06-02
Scaffold has_many relation with RecordEditor/Viewer

gregsmirnov:


kinglozzer:
I don’t think we can rely on running `canEdit()` on a singleton - it’s not uncommon for can* checks to be object-specific (e.g. a user may be able to edit one record of a given class, but not another).

I also think this could be a breaking change, so couldn’t go in 3.1: if someone has written code the expects a relation editor to be returned from the scaffolder, this could break that.

PR #4251, opened 2015-06-03
NEW Configurable file version prefix

jonom:
Numbers are often used at the end of file names in image sequences (e.g. digital photos) and the current renaming scheme can disrupt this, creating a confusing situation for the user. A configurable file version prefix should help prevent this.

Filename   | Old result | New result
-----------|------------|--------------
IMG001.jpg | IMG2.jpg   | IMG001-v2.jpg

Note: this is a version of #4241 for master which introduces a default value of '-v'.

### DIFF
--- a/filesystem/Upload.php
+++ b/filesystem/Upload.php
@@ -72,6 +72,17 @@ class Upload extends Controller {
 	 */
 	private static $uploads_folder = "Uploads";
 
+	/**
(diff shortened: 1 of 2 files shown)


tractorcow:
Thanks @jonom :D I've been noticing you've been doing some good work lately; It's well appreciated!

jonom:
My pleasure

Issue #1574, opened 2013-04-03
[2012-08-23] Allow for multiple remember me tokens


silverstripe-issues:
created by: simon_w
assigned to: simon_w
created at: 2012-08-23
original ticket: http://open.silverstripe.org/ticket/7806

---

The current remember me system stores one token per user, that is either wiped or set on login, and then updated whenever used. While this works, it does mean that as soon as you log in to your account from another browser, or switch out of private browsing, or do anything that changes the cookie store, the token stored in the user's cookies no longer matches the one in the database.

My proposed solution to this is to extract this single field out of Member and into its own DataObject (say MemberRememberToken) that has a has_many relationship with Member. When logging in with remember me enabled, a new MemberRememberToken is created, and its value is used in the cookie.

When falling back to a remember token, Member::autoLogin() will look for a matching MemberRememberToken instead of just a single field. If a matching one is found, the user is logged in and the value of the MemberRememberToken is changed, which is then stored in the cookie again.

On Member::logout(), only the current MemberRememberToken is deleted.

Facebook/Gmail-esque lists of other sessions, browser types and locations can be added on a per site basis, with an extension hooking into populateDefaults()/onBeforeWrite() (depending on if you want the information from when it's created, or every time it changes) storing the UA and IP. I don't see a need for this information to be stored in the core.

dhensby:
Yes, I agree this feature is required. More and more people use multiple devices and only being able to be remembered on one device is a bit silly.

An advantage (if anything) is that multiple tokens would allow more active session management by users. A module could be created to let a user know what devices they are logged in from and to cancel them as needed.

We either have, or have to create, a token scheme that is secure enough that it's "attack vectors" are so low that even having many of them is not going to be feasible to crack.

chillu:
@sminnee has written the following suggestion in response to a recent client request:

> - Add Member has-many MemberAutoLoginKey
> - Each MemberAutoLoginKey has an Identifier, Secret, and Expiry
> - The “remember me?” function will set both an Identifier and a Secret cookie on the user’s machine
>     - The Identifier cookie identifies the device and is not rotated with each visit. This cookie might have a longer life than Secret
>     - The Secret cookie is one-time password, and is rotated with each visit
>  - When secrets are rotated, the existing MemberAutoLoginKey record for the given Identifier is updated.

I've responded: 

[...]

PR #4252, opened 2015-06-03
NEW Cross device "Remember Me" feature

patbolo:
At the moment, using the "Remember me" function on more than one device will only work with the last device used. Previous devices will not auto login.
This PR introduces a new DataObject for storing hashed tokens against multiple devices. Developers can configure if logging out should discard all tokens, or only the one used on the device logging out; token expiry date is 90 days by default but configurable. For added security, the old behaviour can still be enforced if multiple tokens are not desired.
See silverstripe#1574 for additional background

### DIFF
--- a/docs/en/02_Developer_Guides/09_Security/00_Member.md
+++ b/docs/en/02_Developer_Guides/09_Security/00_Member.md
@@ -127,6 +127,13 @@ things, you should add appropriate `[api:Permission::checkMember()]` calls to th
 	  }
 	}
 
+## Saved User Logins ##
+
+Logins can be "remembered" across multiple devices when user checks the "Remember Me" box. By default, a new login token
+will be created and associated with the device used during authentication. When user logs out, all previously saved tokens
+for all devices will be revoked, unless `[api:RememberLoginHash::$logout_across_devices] is set to false. For extra security,
+single tokens can be enforced by setting `[api:RememberLoginHash::$force_single_token] to true.
+
 
 ## API Documentation
 
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -383,13 +383,14 @@ en:
     FIRSTNAME: 'First Name'
     INTERFACELANG: 'Interface Language'
     INVALIDNEWPASSWORD: 'We couldn''t accept that password: {password}'
+    KEEPMESIGNEDIN: 'Keep me signed in'
     LOGGEDINAS: 'You''re logged in as {name}.'
     NEWPASSWORD: 'New Password'
     NoPassword: 'There is no password on this member.'
     PASSWORD: Password
     PASSWORDEXPIRED: 'Your password has expired. Please choose a new one.'
     PLURALNAME: Members
-    REMEMBERME: 'Remember me next time?'
+    REMEMBERME: 'Remember me next time? (for %d days on this device)'
     SINGULARNAME: Member
     SUBJECTPASSWORDCHANGED: 'Your password has been changed'
     SUBJECTPASSWORDRESET: 'Your password reset link'
--- a/security/Member.php
+++ b/security/Member.php
@@ -9,7 +9,6 @@
  * @property string $Surname
  * @property string $Email
  * @property string $Password
- * @property string $RememberLoginToken
  * @property string $TempIDHash
  * @property string $TempIDExpired
  * @property string $AutoLoginHash
@@ -32,7 +31,6 @@ class Member extends DataObject implements TemplateGlobalProvider {
 		'TempIDHash' => 'Varchar(160)', // Temporary id used for cms re-authentication
 		'TempIDExpired' => 'SS_Datetime', // Expiry of temp login
 		'Password' => 'Varchar(160)',
-		'RememberLoginToken' => 'Varchar(160)', // Note: this currently holds a hash, not a token.
 		'AutoLoginHash' => 'Varchar(160)', // Used to auto-login the user on password reset
 		'AutoLoginExpired' => 'SS_Datetime',
 		// This is an arbitrary code pointing to a PasswordEncryptor instance,
@@ -59,6 +57,7 @@ class Member extends DataObject implements TemplateGlobalProvider {
 
 	private static $has_many = array(
 		'LoggedPasswords' => 'MemberPassword',
+		'RememberLoginHashes' => 'RememberLoginHash'
 	);
 
 	private static $many_many = array();
@@ -109,7 +108,6 @@ class Member extends DataObject implements TemplateGlobalProvider {
 	 * @var array
 	 */
 	private static $hidden_fields = array(
-		'RememberLoginToken',
 		'AutoLoginHash',
 		'AutoLoginExpired',
 		'PasswordEncryption',
@@ -447,16 +445,22 @@ class Member extends DataObject implements TemplateGlobalProvider {
 		// This lets apache rules detect whether the user has logged in
 		if(Member::config()->login_marker_cookie) Cookie::set(Member::config()->login_marker_cookie, 1, 0);
 
+		// Cleans up any potential previous hash for this member on this device
+		if ($alcDevice = Cookie::get('alc_device')) {
+			RememberLoginHash::get()->filter('DeviceID', $alcDevice)->removeAll();
+		}
 		if($remember) {
-			// Store the hash and give the client the cookie with the token.
-			$generator = new RandomGenerator();
-			$token = $generator->randomToken('sha1');
-			$hash = $this->encryptWithUserSettings($token);
-			$this->RememberLoginToken = $hash;
-			Cookie::set('alc_enc', $this->ID . ':' . $token, 90, null, null, null, true);
-		} else {
-			$this->RememberLoginToken = null;
+			$rememberLoginHash = RememberLoginHash::generate($this);
+			$tokenExpiryDays = Config::inst()->get('RememberLoginHash', 'token_expiry_days');
+			$deviceExpiryDays = Config::inst()->get('RememberLoginHash', 'device_expiry_days');
+			Cookie::set('alc_enc', $this->ID . ':' . $rememberLoginHash->getToken(), 
+				$tokenExpiryDays, null, null, null, true);
+			Cookie::set('alc_device', $rememberLoginHash->DeviceID, $deviceExpiryDays, null, null, null, true);
+		} else {			
+			Cookie::set('alc_enc', null);
+			Cookie::set('alc_device', null);
 			Cookie::force_expiry('alc_enc');
+			Cookie::force_expiry('alc_device');
 		}
 
 		// Clear the incorrect log-in count
@@ -515,7 +519,9 @@ class Member extends DataObject implements TemplateGlobalProvider {
 	 */
 	public static function autoLogin() {
 		// Don't bother trying this multiple times
-		self::$_already_tried_to_auto_log_in = true;
+		if (!class_exists('SapphireTest') || !SapphireTest::is_running_test()) {
+			self::$_already_tried_to_auto_log_in = true;
+		}
 
 		if(strpos(Cookie::get('alc_enc'), ':') === false
 			|| Session::get("loggedInAs")
@@ -524,36 +530,56 @@ class Member extends DataObject implements TemplateGlobalProvider {
 			return;
 		}
 
-		list($uid, $token) = explode(':', Cookie::get('alc_enc'), 2);
+		if(strpos(Cookie::get('alc_enc'), ':') && Cookie::get('alc_device') && !Session::get("loggedInAs")) {
+			list($uid, $token) = explode(':', Cookie::get('alc_enc'), 2);
+			$deviceID = Cookie::get('alc_device');
 
-		$member = DataObject::get_by_id("Member", $uid);
+			$member = Member::get()->byId($uid);
 
-		// check if autologin token matches
-		if($member) {
-			$hash = $member->encryptWithUserSettings($token);
-			if(!$member->RememberLoginToken || $member->RememberLoginToken !== $hash) {
-				$member = null;
+			$rememberLoginHash = null;
+
+			// check if autologin token matches
+			if($member) {
+				$hash = $member->encryptWithUserSettings($token);
+				$rememberLoginHash = RememberLoginHash::get()
+					->filter(array(
+						'MemberID' => $member->ID,
+						'DeviceID' => $deviceID,
+						'Hash' => $hash
+					))->First();
+				if(!$rememberLoginHash) {
+					$member = null;
+				} else {
+					// Check for expired token
+					$expiryDate = new DateTime($rememberLoginHash->ExpiryDate);
+					$now = SS_Datetime::now();
+					$now = new DateTime($now->Rfc2822());
+					if ($now > $expiryDate) {
+						$member = null;
+					}
+				}
 			}
-		}
 
-		if($member) {
-			self::session_regenerate_id();
-			Session::set("loggedInAs", $member->ID);
-			// This lets apache rules detect whether the user has logged in
-			if(Member::config()->login_marker_cookie) {
-				Cookie::set(Member::config()->login_marker_cookie, 1, 0, null, null, false, true);
+			if($member) {
+				self::session_regenerate_id();
+				Session::set("loggedInAs", $member->ID);
+				// This lets apache rules detect whether the user has logged in
+				if(Member::config()->login_marker_cookie) {
+					Cookie::set(Member::config()->login_marker_cookie, 1, 0, null, null, false, true);
+				}
+
+				if ($rememberLoginHash) {
+					$rememberLoginHash->renew();
+					$tokenExpiryDays = Config::inst()->get('RememberLoginHash', 'token_expiry_days');
+					Cookie::set('alc_enc', $member->ID . ':' . $rememberLoginHash->getToken(), 
+						$tokenExpiryDays, null, null, false, true);
+				}
+
+				$member->write();
+
+				// Audit logging hook
+				$member->extend('memberAutoLoggedIn');
 			}
-
-			$generator = new RandomGenerator();
-			$token = $generator->randomToken('sha1');
-			$hash = $member->encryptWithUserSettings($token);
-			$member->RememberLoginToken = $hash;
-			Cookie::set('alc_enc', $member->ID . ':' . $token, 90, null, null, false, true);
-
-			$member->write();
-
-			// Audit logging hook
-			$member->extend('memberAutoLoggedIn');
 		}
 	}
 
@@ -570,8 +596,13 @@ class Member extends DataObject implements TemplateGlobalProvider {
 
 		$this->extend('memberLoggedOut');
 
-		$this->RememberLoginToken = null;
+		// Clears any potential previous hashes for this member
+		RememberLoginHash::clear($this, Cookie::get('alc_device'));
+
+		Cookie::set('alc_enc', null); // // Clear the Remember Me cookie
 		Cookie::force_expiry('alc_enc');
+		Cookie::set('alc_device', null);
+		Cookie::force_expiry('alc_device');
 
 		// Switch back to live in order to avoid infinite loops when
 		// redirecting to the login screen (if this login screen is versioned)
@@ -1331,6 +1362,8 @@ class Member extends DataObject implements TemplateGlobalProvider {
 
 			// Members shouldn't be able to directly view/edit logged passwords
 			$fields->removeByName('LoggedPasswords');
+
+			$fields->removeByName('RememberLoginHashes');
 
 			if(Permission::check('EDIT_PERMISSIONS')) {
 				$groupsMap = array();
--- a/security/RememberLoginHash.php
+++ b/security/RememberLoginHash.php
@@ -1,0 +1,164 @@
+<?php
+/**
+ * Persists a token associated with a device for users who opted for the "Remember Me"
+ * feature when logging in.
+ * By default, logging out will discard all existing tokens for this user
+ * The device ID is a temporary ID associated with the device when the user logged in
+ * and chose to get the login state remembered on this device. When logging out, the ID 
+ * is discarded as well.
+ *
+ * @package framework
+ * @subpackage security
+ *
+ * @property string $DeviceID
+ * @property string $RememberLoginHash
+ */
+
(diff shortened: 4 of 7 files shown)


tractorcow:
@silverstripe/core-team want to vote on this feature?

willmorgan:
:+1:

JayDevlin:
Maybe there should be a new section in the 'admin/myprofile' and in 'admin/security/EditForm/field/Members', so that users can disable their individual devices themself? ATM there is no way I can logout my phone if it got stolen.
Alternatively -- or perhaps additionally  -- $logout_across_devices should be true by default.

patbolo:
@JayDevlin Agreed with the need to provide a mechanism for administering login tokens. The example you gave is a good one. Some websites might want to give that control to the user, in their user profile, while some others would like to have that done by a website admin through the backend (thru a hot line for example). Also this core feature only provides a cryptic ID for each device, while I imagine a more user friendly description of the device would be needed as well. So all of the above would fit better in a module wouldn't it? 
Agreed as well on the logout_across_devices true by default. Let see what others think
Thanks @JayDevlin !

security/Member.php
```diff
@@ -444,16 +443,22 @@ public function logIn($remember = false) {
...
-			// Store the hash and give the client the cookie with the token.
-			$generator = new RandomGenerator();
-			$token = $generator->randomToken('sha1');
-			$hash = $this->encryptWithUserSettings($token);
-			$this->RememberLoginToken = $hash;
-			Cookie::set('alc_enc', $this->ID . ':' . $token, 90, null, null, null, true);
-		} else {
-			$this->RememberLoginToken = null;
+			$rememberLoginHash = RememberLoginHash::generate($this);
+			$token_expiry_days = Config::inst()->get('RememberLoginHash', 'token_expiry_days');
```
chillu: Should follow coding conventions (camel case)

security/Member.php
```diff
@@ -521,37 +528,49 @@ public static function autoLogin() {
...
+			$rememberLoginHashes = null;
+
+			// check if autologin token matches
+			if($member) {
+				$hash = $member->encryptWithUserSettings($token);
+				$rememberLoginHashes = RememberLoginHash::get()
+					->filter(array(
+						'MemberID' => $member->ID,
+						'DeviceID' => $deviceID,
+						'RememberLoginHash' => $hash
```
chillu: Given this, we should probably add a `$index` for `RememberLoginHash` as well

security/RememberLoginHash.php
```diff
@@ -0,0 +1,140 @@
+<?php
+
+class RememberLoginHash extends DataObject {
+
+	private static $db = array (
+		'DeviceID' => 'VarChar(40)',
```
chillu: Wrong use of camel case

security/RememberLoginHash.php
```diff
@@ -0,0 +1,140 @@
+<?php
+
+class RememberLoginHash extends DataObject {
```
chillu: Needs class level docs

security/RememberLoginHash.php
```diff
@@ -0,0 +1,140 @@
...
+	);
+
+	/**
+	 * Determines if logging out on one device also clears existing login tokens
+	 * on all other devices owned by the member.
+	 *
+	 * @config
+	 * @var bool
+	 */
+	private static $logout_across_devices = false;
```
chillu: Needs to be `true` by default, otherwise I'd consider it "insecure by default" because you're relying on devs building a custom login management profile UI into their implementation. And the docs should make this crystal clear: If you set this to `false`, there's no way for users to revoke a login e.g. from a stolen device, unless you build one.

security/RememberLoginHash.php
```diff
@@ -0,0 +1,140 @@
...
+	/**
+	 * Number of days the device ID will be valid for
+	 *
+	 * @config
+	 * @var int
+	 */
+	private static $device_expiry_days = 365;
+
+	/**
+	 * If true, user can only use auto login on one device
```
chillu: Given this is a security related setting, I'd be a bit clearer here. A user can login from multiple devices regardless of this setting, but if set to `true`, previous tokens from other devices will become invalid.

security/RememberLoginHash.php
chillu: Should note that `DeviceID` is transitory - you won't get the same ID after logging out explicitly, and then using auto-login on the same device again.

security/RememberLoginHash.php
```diff
@@ -0,0 +1,140 @@
...
+			return $rememberLoginHash;
+		}
+	}
+
+	/**
+	 * Generates a new hash for this member but keeps the device ID intact
+	 * @param Member the logged in user
+	 * @return RememberLoginHash
+	 */
+	public function renew(Member $member) {
```
chillu: It's a bit weird to pass in a member on an instance which has a `Member()` association already. Shouldn't it just use that instead?

chillu:
- Have you checked that `Security.autologin_enabled` is still enforced correctly?
- Needs a mention in Markdown docs (`security.md` on `Saved user logins`).
- Needs upgrading docs (we'll need to create a `docs/en/changelogs/4.0.0.md` for that). Particularly point out the new `alc_device` cookie, since some environments will need to whitelist that.
- I think we need to make the wording of the standard "Remember me" checkbox clearer, based on which setting is used. Matt suggested the following at https://github.com/silverstripe/silverstripe-framework/issues/1574#issuecomment-77030979: "Only use this option on a trusted device. We won't ask you for your password on this device again for the next X days"

@patbolo Despite all my niggles above, I think this is an awesome feature addition for 4.0. Thanks so much for doing it!

patbolo:
@chillu I went through all your comments and updated the commit accordingly. I also tested against autologin_enabled and I could not see any regressions. I have also updated the wording for the remember me checkbox, but decided to make it shorter than what matt suggested as it was quite lengthy on the login form. It now reads Remember Me (for x days on this device). This text is still easily editable if  needed
@JayDevlin login_across_devices is now true by default

docs/en/02_Developer_Guides/09_Security/00_Member.md
```diff
@@ -127,6 +127,13 @@ things, you should add appropriate `[api:Permission::checkMember()]` calls to th
 	  }
 	}
 
+## Saved user logins ##
```
chillu: Capitalisation is a bit inconsistent, headline should be capitalised, and "Login", "Logs out" etc in content should be lower case.

dhensby: The contributing guidelines actually say titles should be sentence case, so this is fine here

JayDevlin:
Looks good. Though, I think the test methods are maybe a little bit too long. I suggest to move the 'Re-logging (ie 'alc_enc' has expired)' and 'Logging out ('logout_across_devices' = false/true)' tests into their own test methods. 
Also, $force_single_token is not covered by tests?

docs/en/02_Developer_Guides/09_Security/00_Member.md
chillu: Hah, I wrote that guideline as well ... serves me right. But this specific case is inconsistent either way - sentence case would mean that "logins" is lowercase.

chillu:
Thanks for changing all of this, Jean! We'll need a data migration strategy in the upgrading docs for existing tokens. I suppose a simple `INSERT ... SELECT FROM ...` query would be cross-database enough for this purpose? I'd prefer it over `requireDefaultRecords` hackery or a special `BuildTask`, since both would be considered dead code after migration. 

@silverstripe/core-team Given this is a security related feature which to a certain degree also impacts the user experience, I would want more than two core votes (Will and me) before this can be merged. Any comments or concerns?

security/RememberLoginHash.php
```diff
@@ -0,0 +1,158 @@
...
+	/**
+	 * Generates a new login hash associated with a device
+	 * The device is assigned a globally unique device ID
+	 * The returned login hash stores the hashed token in the 
+	 * database, for this device and this member
+	 * @param Member The logged in user 
+	 * @return RememberLoginHash The generated login hash
+	 */
+	public static function generate(Member $member) {
+		if ($member) {
```
kinglozzer: We're type-hinting ´$member´ and forcing the parameter, so this ´if´ statement shouldn't be necessary, right?

security/RememberLoginHash.php
```diff
@@ -0,0 +1,158 @@
...
+		return $this;
+	}
+
+	/**
+	 * Deletes existing tokens for this member
+	 * if logout_across_devices is true, all tokens are deleted, otherwise
+	 * only the token for the provided device ID will be removed
+	 */
+	public static function clear(Member $member, $alcDevice = null) {
+		if ($member) {
```
kinglozzer: As above

sminnee:
I'm very keen to see this feature included, as SilverStripe's remember me function has been pretty much useless since smartphones came into common use.

Are there API breakages that would stop this from being suitable for the 3 branch?

chillu:
@patbolo Would you be able to rebase and implement Sam's `getRememberLoginToken()` suggestion? SO CLOSE :)

sminnee:
Does anyone object to this going in 3?

security/RememberLoginHash.php
```diff
@@ -0,0 +1,158 @@
...
+
+	private static $indexes = array(
+		'DeviceID' => true,
+		'RememberLoginHash' => true
+	);
+
+	/**
+	 * Determines if logging out on one device also clears existing login tokens
+	 * on all other devices owned by the member.
+	 * If set to false, there is no way for users to revoke a login, unless additional
```
dhensby: or they change their password

dhensby:
I think there's a problem with this. `token_expiry_days` only applies to the Cookie that is set in the browser but not to the token itself.

A user can therefore keep the login token *indefinitely* and login with it from any device even after the 90 days have elapsed.

There doesn't seem to be any pruning of old tokens from the DB, let alone the possibility of this as the token objects themselves don't record an expiry date.


Personally, I don't think this goes far enough; we should have a device object and a hash object, the device object can store things like user-agent, last IP used to access, last active, etc.

You may argue that this could be an extension or module in its own right, but there are no extension points for this, there's no "backend" for implementing or swapping out your auth tokens, so it needs to be in core. This way, you could remove a device from your account.

dhensby:
> Does anyone object to this going in 3?

no

chillu:
@dhensby You can keep the current `Member.RememberLoginToken` indefinitely with the same reasoning as well (expiry only stored in cookie = user modifiable). Do you think this is a security issue? I guess a stolen token (MITM attack, physical device access) could be reused for longer if you don't force expiry - but that's the least of your worries then. Correlating a token with a "device" (as identified by user agent) makes it harder to reuse, but AFAICT it's still vulnerable to MITM attack since the same HTTP communication contains both user agent and cookie value and hence can be emulated from an attacker's system as well. Limiting the token to the IP is too much of a usability impact, since the majority of users will be on dynamic IPs which change every day or so. 

Even tracking the IP by default could be seen as a privacy issue by some implementors = core functionality shouldn't rely on it. Would be a nice case for a module though (through standard `DataExtension` and `Member` login hooks). Ideally we have the extension hooks to influence token validity checking in this commit, but I wouldn't hold it up for it.

dhensby:
@chillu my thoughts covered a couple of points

1. I think its a security flaw that the token never actually expires (despite the implication that it should). Yes, it is a security vulnerability because it increases the attack surface if you know any token that ever existed could work. Theoretical, but one none-the-less.
2. Not storing the expiry date against a token also makes it impossible to prune or clean the DB table, which can be a problem.
3. Seperate to security, I think a device object and device token should both have their own corresponding objects to more easily allow the expiring of tokens in the CMS or what have you. At the moment it'd be impossible to advise a user what token they are expiring.

tractorcow:
I don't think this should go into 3. Security is a tricky enough feature to get right, and there are likely use cases out there that rely on the existing security token feature.

You can add a getter for it, but as far as we have acted up until now, database fields are NOT a part of non-public API.

In order to get this into 3 it would be necessary to support both single and multiple device tokens, and require a config change to flip to multi-device. We could either change the default in 4, or just remove the single-token one.

tractorcow:
@sminnee @chillu @dhensby do you feel my concern is warranted? If not, I guess I'm happy to be out-voted and have this merged into 3.

patbolo:
As I agree with some of @dhensby comments, I have implemented some changes, namely about the storing of the expiry date when a token is initially generated. This is also covered in tests.
I have also added a couple of hooks, for people who would like to store IP address or device name against the token.
PR is still against 4. I can rebase but will only do once a firm decision is made (I have already rebased from 3 to 4 before)

security/Member.php
```diff
@@ -520,36 +526,56 @@ public static function autoLogin() {
...
+					))->First();
+				if(!$rememberLoginHash) {
+					$member = null;
+				} else {
+					// Check for expired token
+					$expiryDate = new DateTime($rememberLoginHash->ExpiryDate);
```
chillu: Shouldn't it also remove the token from the DB in this case?

security/RememberLoginHash.php
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * @property string $DeviceID
+ * @property string $RememberLoginHash
+ */
+
+class RememberLoginHash extends DataObject {
+
+	private static $db = array (
+		'DeviceID' => 'Varchar(40)',
+		'RememberLoginHash' => 'Varchar(160)',
```
chillu: The naming is a bit verbose, same as class name - can we make it `Hash` or `Value`?

security/RememberLoginHash.php
```diff
@@ -0,0 +1,166 @@
...
+			return $rememberLoginHash;
+		}
+	}
+
+	/**
+	 * Generates a new hash for this member but keeps the device ID intact
+	 * @param Member the logged in user
+	 * @return RememberLoginHash
+	 */
+	public function renew() {
```
chillu: Shouldn't this also update the expiry date?

chillu:
I've assumed this goes into 4.0 all along. Agree with Damian that it's a bit risky for a 3.3 release.

sminnee:
Fine, I know when I'm outvoted. A bit annoying that remember me remains basically useless until then but you're probably right re: the risk.

security/Member.php
dhensby: I think keeping expired tokens can be desirable as it allows retrospective inspection of logged in devices, assurances that tokens WERE expired for a device and so on.

dhensby:
It seems to me that this feature has a lot of ways that people may want to use it. At the moment, is it possible for a developer to hook into this and say "I want to store all device tokens indefinitely, as well as the device it was linked to and provide a means for users to log-out individual sessions" - it seems not and not is it architected in a way to allow that use case to be elegantly added.

If this is going into 4 (which I think it probably should) then it needs to be written in a way that allows this fairly predictable set of requirements to be sanely implemented without throwing all this work away.

chillu:
> I want to store all device tokens indefinitely, as well as the device it was linked to and provide a means for users to log-out individual sessions

I have trouble imagining a valid use case for this - particularly if you consider that token expiry is extended on every `Member->autoLogin()` call, so whenever a user accesses the site through a new session. It seems prudent to expire these tokens when the service isn't accessed for a longer time frame (90 days by default). You can extend set this expiry to multiple years if the specific context calls for it. Providing means for users to log out of individual sessions is already in place, and could be extended by a custom UI to view currently active tokens (and "log out" of each of them).

sminnee:
>  it seems not and not is it architected in a way to allow that use case to be elegantly added

I don't see the blocker here. As Ingo says, as UI could be build around the RememberLoginHash DataObject. Could you elaborate on the limitations you'd like to overcome, Dan?

dhensby:
I'm thinking around wanting to see a login history, previous sessions, etc. Seeing ones that have been expired by hand (or by time).

Is there not a use-case where you'd want to allow users to see their current logged in sessions and to log those out on an individual basis? I'm sure I've seen this feature around on services...

Probably not a blocker seeing as this has been an issue for so long and would be good to see it sorted in some way.

security/Member.php
tractorcow: It's good to keep an audit log, as long as they are safely expired. :D

chillu:
> Is there not a use-case where you'd want to allow users to see their current logged in sessions and to log those out on an individual basis?

Sam and myself are acknowledging this use case, but see it as a custom feature rather than core functionality. And the current architecture allows for building that, so we're good.

jonom:
I agree that's a valid use case but doesn't need to be core functionality. Something that might be worth considering for the CMS project to go with this change though is that when you click logout, if the user is logged in to more than one device and `$logout_across_devices` has been set to `false` then the user is presented with a dialogue like:

**You are logged in to more than one device**
`[Log out of this device only]` `[Log out of all (3) devices]`

I don't think it's intuitive for a logout on one device to log me out of all devices but understand it's safer so makes a good default. Think it will be tempting for developers to change this to false to get the most benefit out of the Remember Me functionality, but in that case a little warning such as above whenever people log out would be a good idea. Doesn't go as far as what Dan outlined but could be implemented without expanding the scope of `RememberLoginHash`.

lang/en.yml
```diff
@@ -379,7 +379,7 @@ en:
     PASSWORD: Password
     PASSWORDEXPIRED: 'Your password has expired. Please choose a new one.'
     PLURALNAME: Members
-    REMEMBERME: 'Remember me next time?'
+    REMEMBERME: 'Remember me next time? (for %d days on this device)'
```
jonom: I would prefer if this just said *"Keep me signed in"*. I don't think we need to reveal how long a token will last for and a specific number is misleading because if a user regularly visits the site, they might not have to log in again for much longer than we indicate here. Conversely if they log out on a different device tomorrow, they'll be logged out on this device as well (by default) much sooner than `%d` days. You could argue that this is exposing configuration too, particularly if `%d` represents a long time it could highlight the site as being easier to attack.

If we do want to give users more detailed information about how the login system works maybe we can use a tooltip?

jonom:
With above as a use case, the logout function could be expanded to allow developers flexibility to implicitly log out users on one or all devices:

```php
function logout($allDevices=null) {
	// if logout context isn't specified use default behaviour
	$allDevices = is_null($allDevices) ? self::config()->logout_across_devices : $allDevices;
	..
}
```

security/RememberLoginHash.php
patbolo: Hi @chillu I am reviewing that PR- That's a good question you have here. So if we renew the expiry date each time the visitor logs in, it means that the user could stay logged in well over the time set in `token_expiry_days`. I am happy to make that change, but also thought that it would be beneficial to ensure that the token would expire regularly (so after `token_expiry_days` days after the token was first generated) for security purposes?

lang/en.yml
chillu: Agree with "Keep me signed in" label, it's more accurate. As the tooltip, I would suggest "Remember this device for %d days". Not too worried about exposing configuration. Showing the number of days helps users understand a) why they might be asked to log in again and b) increase confidence that logins won't work indefinitely. Then again, this might be such an established pattern that we don't need to be as explicit about it.

Facebook doesn't have a tooltip, but then again they have a lot more additional controls at work to ensure security. Basecamp says "Remember me on this computer" without a tooltip. Twitter just says "Remember me". Google has the following:
image

Here's a few more resources on the topic:
http://ux.stackexchange.com/questions/22216/do-we-really-need-a-remember-me-option-on-a-login-page
http://ux.stackexchange.com/questions/67539/remember-me-login-functionality?lq=1
http://www.siteflex.com.au/blog/n/remember-me-remember-me-not-140407
http://www.sitepoint.com/3-rules-painless-account-ux-login-screens/

security/RememberLoginHash.php
chillu: @patbolo Can you please fix?

security/RememberLoginHash.php
chillu: Agreed. It limits the exploitability of a compromised cookie value. If an attacker has this cookie value, they can just access the account for a limited amount of time, even for frequent users. Assuming the app secures other privileged actions like changing passwords by requiring to enter the current password first, of course.

chillu:
@patbolo I think we're getting close with this one, thanks for staying on top of all the comments. Reading through the discussion, I think there's only a few formatting issues left. Implementing a checkbox tooltip is a nice-to-have for me, happy to merge without it.

I want to get this feature in 4.0 ASAP, in order to maximise its exposure and real-world testing - it's easy to get the balance between security and usability wrong, or introduce a security flaw somewhere.

@silverstripe/core-team Any merge blockers or strong objections?

security/RememberLoginHash.php
tractorcow: I suggest changing to `if(!$member->exists()) { return; }`, in case an unsaved member is passed in. Return instead of nesting the code enhances readability. :)

tractorcow:
Other than the already given feedback, I have no concerns. :)

lang/en.yml
patbolo: "Keep me signed in" as been added and "Remember me next time (after x days)" has been assigned as the title attribute to the checkbox (so it shows when the mouse is over the box)

patbolo:
I have reviewed and implemented all suggested changes
Not sure why one of the checks fails as it is not giving any details about why

Thanks everyone for your feedback on that PR!

dhensby:
Weird that it's checking against appveyor and not travis ?!

chillu:
No idea what's up with Travis here, can't see it in https://travis-ci.org/silverstripe/silverstripe-framework/pull_requests either - and no obvious way to force a build on a specific branch in the Travis UI. I've run `vendor/bin/phpunit framework/tests/security/` locally and it looks good. Merging!

tractorcow:
Hurray!

jonom:
Sorry I missed the chance for final feedback, have been away :) great to see this feature merged.

My only concern that we could maybe address in a future commit is that it's a fairly inflexible model at the moment regarding whether you are logged out of just one or all devices. This behaviour is controlled by a config variable so we're not empowering developers to use either behaviour depending on context - they have to use one or the other only (or change a config variable back and forth). If we're providing a solid base for developers to build upon I think we should implement something like what I outlined [above](https://github.com/silverstripe/silverstripe-framework/pull/4252#issuecomment-141838451), and/or provide distinct methods for each variant like `logoutThisDevice()` and `logoutAllDevices()`.

PR #4253, opened 2015-06-05
Issue with empty CheckboxSetField value

swilsonalfa:
If the source of CheckboxSetField contains an empty string (i.e, nothing selected) and you have an option with the key value of 0, it will be selected. Explode returns an array of key 0/value empty with an empty input. Therefore, check that the value is not null or empty before exploding, otherwise, return an empty array.

### DIFF
--- a/forms/CheckboxSetField.php
+++ b/forms/CheckboxSetField.php
@@ -96,8 +96,12 @@ class CheckboxSetField extends OptionsetField {
 						}
 					}
 				} elseif($values && is_string($values)) {
-					$items = explode(',', $values);
-					$items = str_replace('{comma}', ',', $items);
+					if(!empty($values)) {
+						$items = explode(',', $values);
+						$items = str_replace('{comma}', ',', $items);
+					} else {
+						$items = array();
+					}
 				}
 			}
 		} else {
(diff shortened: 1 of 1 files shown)


willmorgan:
Can I be cheeky and ask for some unit tests, too, please? :wink:

swilsonalfa:
Changed to use empty() instead and squashed for convenience. Sorry but I'm not familiar with unit testing (but plan to look into it soon) - anything to point me in the right direction>

willmorgan:
Cheers Stewart.

In terms of unit testing, here are a few guides I hope you'll find useful:
* http://docs.silverstripe.org/en/3.1/developer_guides/testing/
* http://docs.silverstripe.org/en/3.1/developer_guides/testing/unit_testing/
* http://docs.silverstripe.org/en/3.1/developer_guides/testing/how_tos/write_a_sapphiretest/

tractorcow:
Is there a chance that some of the duplicate code could be consolidated here?

dhensby:
@tractorcow :(

tractorcow:
It wasn't worth blocking the merge, honestly. :D

PR #4256, opened 2015-06-08
BUG Fix false values for many_many_ExtraFields not being saved

### LINKED ISSUE #4067: [Bug] Can't save an unchecked checkbox for a boolean in a many_many_extraField
I've got an object called `Account` which has a `many_many` to `Users`. `Account` has the following `many_many_extraFields`:

```
private static $many_many_extraFields = array(
	'Users' => array(
		'CanRead' => 'Boolean',
		'CanModify' => 'Boolean',
		'CanManage' => 'Boolean',
		'Note' => 'Varchar'
	)
);
```

I am using the following code to show these options when navigating that relationship via GridField:

```
$usersGridFieldConfig = $fields->fieldByName('Root.Users.Users')->getConfig();

$usersExtraFields = new FieldList();
$usersExtraFields->push(CheckboxField::create(
	'ManyMany[CanRead]',
	'Can Read'
));

$usersGridFieldConfig->getComponentByType('GridFieldDetailForm')->setFields($usersExtraFields);
```

So far so good, the field will appear when navigating the relationship just how I wanted. When checking the checkbox and pressing save, that correctly saves to the DB (confirmed when manually checking the DB and the checkbox appears checked when loaded).

When I now uncheck the checkbox and press save, it says the save is successful however neither the DB or field itself is updated.

I have checked with the "Note" extra field by setting a value, saving and then blanking it and that doesn't cause any issues. Also added another extra field with the Int data type, no issues setting various values for that.

I have also double checked my code against [the example of `many_many_extraFields` in the documentation](http://doc.silverstripe.org/en/developer_guides/forms/field_types/gridfield#many-many-extrafields) and I don't see anything I am doing that would be particularly wrong.

This is with Framework and CMS 3.1.x-dev (updated yesterday). No errors are shown, just doesn't save that field's value. Not sure if this is possibly related to #1664 as well.

tractorcow:
Fixes #4067 

This is because the $extraData was extracted directly from the posted data, when the correct way to retrieve data is to rely on the result of Form::saveInto, as this is how fields such as CheckboxField are able to overwrite values with false if no data is posted.

### DIFF
--- a/forms/gridfield/GridFieldDetailForm.php
+++ b/forms/gridfield/GridFieldDetailForm.php
@@ -495,20 +495,38 @@ class GridFieldDetailForm_ItemRequest extends RequestHandler {
 
 		return $backlink;
 	}
-
-
+	
+	/**
+	 * Get the list of extra data from the $record as saved into it by
+	 * {@see Form::saveInto()}
+	 * 
+	 * Handles detection of falsey values explicitly saved into the
+	 * DataObject by formfields
+	 * 
+	 * @param DataObject $record
+	 * @param SS_List $list
+	 * @return array List of data to write to the relation
+	 */
+	protected function getExtraSavedData($record, $list) {
+		// Skip extra data if not ManyManyList
+		if(!($list instanceof ManyManyList)) {
+			return null;
+		}
+		
+		$data = array();
+		foreach($list->getExtraFields() as $field => $dbSpec) {
+			$savedField = "ManyMany[{$field}]";
+			if($record->hasField($savedField)) {
+				$data[$field] = $record->getField($savedField);
+			}
+		}
+		return $data;
+	}
 
 	public function doSave($data, $form) {
 		$new_record = $this->record->ID == 0;
 		$controller = $this->getToplevelController();
 		$list = $this->gridField->getList();
-
-		if($list instanceof ManyManyList) {
-			// Data is escaped in ManyManyList->add()
-			$extraData = (isset($data['ManyMany'])) ? $data['ManyMany'] : null;
-		} else {
-			$extraData = null;
-		}
 
 		if(!$this->record->canEdit()) {
 			return $controller->httpError(403);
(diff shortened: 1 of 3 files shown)


model/ManyManyList.php
```diff
@@ -230,37 +230,29 @@ public function add($item, $extraFields = array()) {
...
-				$manipulation[$this->joinTable]['command'] = 'update';
 				$manipulation[$this->joinTable]['where'] = array(
 					"\"{$this->joinTable}\".\"{$this->foreignKey}\"" => $foreignID,
 					"\"{$this->joinTable}\".\"{$this->localKey}\"" => $itemID
 				);
-			} else {
-				$manipulation[$this->joinTable]['command'] = 'insert';
 			}
 
-			if($extraFields) {
```
tractorcow: The existing code was trying to hack in custom manipulations, and it was breaking on sqlite3, so I simply took the code from DataObject and ensured it used the same method.

forms/gridfield/GridFieldDetailForm.php
```diff
@@ -495,21 +495,38 @@ protected function getBackLink(){
 
 		return $backlink;
 	}
-
-
+	
+	/**
+	 * Get the list of extra data from the $record as saved into it by
+	 * Form::savInto
```
assertchris: Typo

forms/gridfield/GridFieldDetailForm.php
```diff
@@ -495,21 +495,38 @@ protected function getBackLink(){
...
+	 * 
+	 * Handles detection of falsey values explicitly saved into the
+	 * DataObject by formfields
+	 * 
+	 * @param DataObject $record
+	 * @param SS_List $list
```
assertchris: No @return

forms/gridfield/GridFieldDetailForm.php
tractorcow: You mean, so I've got nothing to fix later? Ok. =(

tractorcow:
Fixed based on your feedback @assertchris

tractorcow:
@dhensby want to look at this while you're reviewing stuff for 3.2?

tractorcow:
Ta :D

gordonbanderson:
This bug still appears to exist in 3.1.15.  I've just into it, Checkboxes are only being saved if their value is true.

tractorcow:
Yep. :D Users will need to upgrade to 3.2.0 to get this fix.

gordonbanderson:
I tried to work around this using a PickerField module, which turned out to have the same issue.  However it was possible to fix it there, and this offers a workaround for 3.1 users https://github.com/briceburg/silverstripe-pickerfield/pull/14#issuecomment-152885148

PR #4258, opened 2015-06-08
API Formalise new additional arguments

tractorcow:
One long long overdue, the other is quite new.

Also would like to fix FormField::setValue, but I think that should be a separate PR (since I actually want to split the second argument out into its own method).


tractorcow:
See https://github.com/silverstripe/silverstripe-cms/pull/1212 for CMS pr

PR #2960, opened 2014-03-16
API: Support for multiple HTMLEditorConfig per page

colymba:
This allows to specify a `HTMLEditorConfig` on a per field basis.

`HtmlEditorField` now takes an optional 4th argument with the identifier of the EditorConfig. This falls back to the active EditorConfig identifier, so this should be backward compatible.

`HTMLEditorConfig` now output `ssTinyMceConfig` as an object with key:value pairs of Identifier and Config. Also takes care of loading all external plugins (maybe this could be done on a per field init basis rather than load them all at the bottom?).

This is similar to https://github.com/silverstripe/silverstripe-framework/issues/2940 but with a different approach. Using *identifiers* rather than *inline* as a parameter might be more intuitive?
This PR requires https://github.com/silverstripe/silverstripe-framework/pull/2297 to solve multi TinyMCE inits.

Pulled against master for now, but may be better against 3.1? This is maybe more a NEW features rather than API change?
This is up for review and comments.

PR #4259, opened 2015-06-09
API Support for multiple HTMLEditorConfig per page

tractorcow:
Rebased version of https://github.com/silverstripe/silverstripe-framework/pull/2960

These alterations were made:
* Changelog notice in 3.2.0.md
* Instead of removing `HtmlEditorField::include_js` I made it raise a deprecation notice and call `HtmlEditorConfig::require_js`.

Tested locally with multiple configs and works flawlessly. nice job. :D

### DIFF
--- a/admin/code/LeftAndMain.php
+++ b/admin/code/LeftAndMain.php
@@ -339,8 +339,6 @@ class LeftAndMain extends Controller implements PermissionProvider {
 
 		if (Director::isDev()) Requirements::javascript(FRAMEWORK_ADMIN_DIR . '/javascript/leaktools.js');
 
-		HTMLEditorField::include_js();
-
 		$leftAndMainIncludes = array_unique(array_merge(
 			array(
 				FRAMEWORK_ADMIN_DIR . '/javascript/LeftAndMain.Layout.js',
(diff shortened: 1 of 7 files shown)


dhensby:
Well, if it works flawlessly...

admin/code/LeftAndMain.php
```diff
@@ -339,8 +339,6 @@ public function init() {
 
 		if (Director::isDev()) Requirements::javascript(FRAMEWORK_ADMIN_DIR . '/javascript/leaktools.js');
 
-		HTMLEditorField::include_js();
```
ntd: This removal gave me some headaches... I am using a TinyMCE editor inside the `fileEditFields` of an `UploadField` and after this commit I need to load that JS by myself (see ntd/silverstripe-carousel@a729b710cb3d8cad7f0843cb84e07f1c27728245).

Not sure it is a regression or something expected though.

dhensby: Hmm :/ sounds like a regression for a 3.x release, tbh.. @tractorcow ?

tractorcow: Yeah, I think it needs to be here. Easy fix?

tractorcow: Fix at https://github.com/silverstripe/silverstripe-framework/pull/5839

PR #4261, opened 2015-06-09
BUG Fix deletion of orphaned versioned records when a parent _versions has been deleted

tractorcow:
Before this we would check that `"{$table}_versions"` exists but we never check if `"{$child}_versions"` exists, and it crashes.


dhensby:
tests?

tractorcow:
I don't really want to (Please appreciate the honesty ;D), but, I guess I could. It won't happen any time soon though, since it'll end up being an obscure edge-case where the table schema is corrupted (missing tables etc).

dhensby:
Hehe, ok.

PR #4263, opened 2015-06-09
DOCS: Updated docblocks of file link methods

sminnee:
... to clarify expectations about return values

This is a response to this discussion about public APIs https://github.com/silverstripe/silverstripe-framework/pull/4229#discussion_r31983112 

We should be clearer about our edge-case expectations of APIs. I'd like to see this being a lot more complete for 4.0 if possible (a man can dream!) :-)

### DIFF
--- a/filesystem/File.php
+++ b/filesystem/File.php
@@ -249,7 +249,9 @@ class File extends DataObject {
 	}
 
 	/**
-	 * Just an alias function to keep a consistent API with SiteTree
+	 * Returns a web link download/view to the file. May be absolute or relateive, depending on configuration.
+	 *
+	 * When linking to files, use this method rather than constructing the link yourself.
 	 *
 	 * @return string The link to the file
 	 */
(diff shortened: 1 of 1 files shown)


dhensby:
@sminnee Seems fine, just curious as to why you've effectively swapped the docs and marked the original functions as aliases?

tractorcow:
I think @dhensby might be right.

sminnee:
It was to set expectations about which one was preferred. Should I rebase this?

filesystem/File.php
```diff
@@ -249,7 +249,9 @@ public static function find($filename) {
 	}
 
 	/**
-	 * Just an alias function to keep a consistent API with SiteTree
+	 * Returns a web link download/view to the file. May be absolute or relateive, depending on configuration.
```
tractorcow: `relative`

tractorcow:
I think it's probably fine, but I think it's more usual for the alias to invoke the method it's aliasing, rather than the other way around. :P Not a merge blocker really.

Please rebase and I'll merge. :) You may as well fix that spelling mistake while you're at it.

tractorcow:
I think this PR is well out of date; In any case all 4.0 File methods have been rewritten.

PR #4264, opened 2015-06-09
API enable deprecation notices to be disabled

tractorcow:
This should reduce the impact of deprecating non-breaking api early.

### DIFF
--- a/dev/Deprecation.php
+++ b/dev/Deprecation.php
@@ -43,6 +43,19 @@ class Deprecation {
 	protected static $version;
 
 	/**
+	 * Override whether deprecation is enabled. If null, then fallback to 
+	 * SS_DEPRECATION_ENABLED, and then true if not defined.
+	 * 
+	 * Deprecation is only available on dev.
+	 * 
+	 * Must be configured outside of the config API, as deprecation API 
+	 * must be available before this to avoid infinite loops.
+	 *
+	 * @var boolean|null
+	 */
+	protected static $enabled = null;
+
+	/**
 	 *
 	 * @var array
 	 */
@@ -116,20 +129,47 @@ class Deprecation {
 			return $called['function'];
 		}
 	}
+	
+	/**
+	 * Determine if deprecation notices should be displayed
+	 * 
+	 * @return bool
+	 */
+	public static function get_enabled() {
+		// Deprecation is only available on dev
+		if(!Director::isDev()) {
+			return false;
+		}
+		if(isset(self::$enabled)) {
+			return self::$enabled;
+		}
+		if(defined('SS_DEPRECATION_ENABLED')) {
+			return SS_DEPRECATION_ENABLED;
+		}
+		return true;
+	}
+
+	/**
+	 * Toggle on or off deprecation notices. Will be ignored in live.
+	 * 
+	 * @param bool $enabled
+	 */
+	public static function set_enabled($enabled) {
+		self::$enabled = $enabled;
+	}
 
 	/**
 	 * Raise a notice indicating the method is deprecated if the version passed as the second argument is greater
 	 * than or equal to the check version set via ::notification_version
 	 *
-	 * @static
-	 * @param $string - The notice to raise
-	 * @param $atVersion - The version at which this notice should start being raised
-	 * @param Boolean $scope - Notice relates to the method or class context its called in.
-	 * @return void
+	 * @param string $atVersion The version at which this notice should start being raised
+	 * @param string $string The notice to raise
+	 * @param bool $scope Notice relates to the method or class context its called in.
 	 */
 	public static function notice($atVersion, $string = '', $scope = Deprecation::SCOPE_METHOD) {
-		// Never raise deprecation notices in a live environment
-		if(Director::isLive(true)) return;
+		if(!static::get_enabled()) {
+			return;
+		}
 
 		$checkVersion = self::$version;
 		// Getting a backtrace is slow, so we only do it if we need it
@@ -179,25 +219,27 @@ class Deprecation {
 
 	/**
 	 * Method for when testing. Dump all the current version settings to a variable for later passing to restore
-	 * @return array - opaque array that should only be used to pass to ::restore_version_settings
+	 * 
+	 * @return array Opaque array that should only be used to pass to {@see Deprecation::restore_settings()}
 	 */
 	public static function dump_settings() {
 		return array(
 			'level' => self::$notice_level,
 			'version' => self::$version,
-			'moduleVersions' => self::$module_version_overrides
+			'moduleVersions' => self::$module_version_overrides,
+			'enabled' => self::$enabled,
 		);
 	}
 
 	/**
 	 * Method for when testing. Restore all the current version settings from a variable
-	 * @static
-	 * @param $settings array - An array as returned by ::dump_version_settings
-	 * @return void
+	 * 
+	 * @param $settings array An array as returned by {@see Deprecation::dump_settings()}
 	 */
 	public static function restore_settings($settings) {
 		self::$notice_level = $settings['level'];
 		self::$version = $settings['version'];
 		self::$module_version_overrides = $settings['moduleVersions'];
+		self::$enabled = $settings['enabled'];
 	}
 }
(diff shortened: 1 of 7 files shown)


tractorcow:
Probably should collect feedback before we merge this. @silverstripe/core-team , @hafriedlander any thoughts?

sminnee:
Thanks, Damian.

 * Have we posted this to silverstripe-dev?
 * Do we want to add support for an _ss_environment define?
 * There's a minor point that should be acknowledged: in removing deprecation notices, we changing behaviour in a minor release. However, because this change actively reduces upgrade friction, which is the goal of semver, I think this is okay.

tractorcow:
@sminnee done :D

tractorcow:
Problem I've found in this... Config class uses Deprecation, so we can't use any `Config` api in here. I'll have to strip it out. =(

tractorcow:
Here is one benefit of this:

image

tractorcow:
Finally fixing issues with deprecation in some tests. Hope I have this finally....

gregsmirnov:
@tractorcow small improvement. You missed method reference in dockblock for dumping settings

::dump_version_settings => ::dump_settings
::restore_version_settings => ::restore_settings

willmorgan:
This is a good PR, although I think deprecation notices should be shown by default in dev mode. Agree that they should never be shown in live mode.

dev/Deprecation.php
```diff
@@ -116,6 +123,34 @@ protected static function get_called_method_from_trace($backtrace, $level = 1) {
...
 	}
+	
+	/**
+	 * Determine if deprecation notices should be displayed
+	 * 
+	 * @return boolean
+	 */
+	public static function get_enabled() {
+		// Never raise deprecation notices in a live environment
+		if(Director::isLive(true)) {
```
dhensby: Personally my preference here would be `!Director::isDev()`

I don't see that deprecation notices should come up under any other circumstance

docs/en/04_Changelogs/3.2.0.md
```diff
@@ -19,6 +19,7 @@
  * `FormField::validate` now requires an instance of `Validator`
  * Implementation of new "Archive" concept for page removal, which supercedes "delete". Where deletion removed
    pages only from draft, archiving removes from both draft and live simultaneously.
+ * `Deprecation` class no longer raises issues automatically. Enable `Deprecation.enabled` config to turn on.
```
dhensby: This patch doesn't appear to use the `Config` system

dhensby:
@willmorgan As in, if dev, show unless explicitly off. If not dev hide unless explicitly shown?

Except really we are saying they should never be shown outside dev environments.

While I agree that deprecations should be on by default, consensus does not lie with that point of view. Most frameworks will only show these kinds of notice is "debug" mode, so that's essentially what we have here.

willmorgan:
> As in, if dev, show unless explicitly off. If not dev hide unless explicitly shown?

No - if dev, show unless explicitly off. If not dev, never show.

dev/Deprecation.php
oddnoc: Director::isLive() takes no arguments.

tractorcow: Ah, it looks like it used to; I'll update. ;D

tractorcow: @dhensby agreed, let's make it !isDev(). Test sites shouldn't show deprecation notices.

tractorcow:
I think that popular opinion is to favour on by default, but only in dev. I think that we should leave it on by default (as it is) and perhaps look at changing the default setting in a later framework version?

tractorcow:
I've updated this based on feedback:

- Fixed docs to not refer to config anymore.
- Fixed phpdoc
- Deprecation only available at all on dev
- Deprecation still enabled by default

dhensby:
I think the two big issues here are: 

1. Depreciation errors break tests (believe this is now fixed in this PR?) 
2. The SS error handler shouts too loudly about them  

I think the ss error handler should be much more subtle about them, perhaps suggesting how they are disabled and what it means and staying as the default php line of text. 

Further, a depreciation notice should probably only be fired once per function, not repeatedly during execution. (this is WP behaviour)

dhensby:
A +1 for this too is that it is BC! If we want them off by default, then it should happen in v4

dhensby:
So, what does this mean for `Deprecation::notice($version)`, should $version say when we expect to take it out or when it was deprecated? The former means devs will have to actively change their notice settings to ever be warned, the latter doesn't

jonom:
Shouldn't `$version` indicate when the function is expected to be removed? The notice should give you an idea of how urgent it is to change I think. Agree a much more subtle notice would be good too, particularly in the case of a module causing a deprecation notice you can't do anything about, especially if the module owner doesn't want to create multiple branches just to get rid of deprecation notices.

On that subject could something like this be an idea to give us a bit more flexibility about which notices to let through?
```php
/**
 * Specify classes which should have deprecation notices muted.
 * Deprecation is only available on dev.
 *
 * @var array
 */
protected static $suppressed = array();
```
```yml
Deprecation:
  suppressed:
    - SomeOldClass
```

dev/Deprecation.php
```diff
@@ -116,20 +126,47 @@ protected static function get_called_method_from_trace($backtrace, $level = 1) {
...
+	 */
+	public static function get_enabled() {
+		// Deprecation is only available on dev
+		if(!Director::isDev()) {
+			return false;
+		}
+		if(isset(self::$enabled)) {
+			return self::$enabled;
+		}
+		if(defined('SS_DEPRECATION_ENABLED')) {
```
dhensby: Should the static var override the constant? I generally the constants take absolute precedence usually.

However, I'd suggest they have equal weight. if either is disabled, then don't show? Maybe that makes it a bit less flexible, though...

hafriedlander:
`$version` is / was supposed to be when it's expected to be taken out. The commit the `Deprecation::notice` call was added is when the deprecation itself occurs. Was more important before semver (now the rule is presumably "next major version").

tractorcow:
> Depreciation errors break tests (believe this is now fixed in this PR?)

If we merge this, we can disable them in travis-support, but developers will still get them on by default.

dev/Deprecation.php
tractorcow: Constants are "environment" and static vars "projects", which is the correct order of inheritance.

dev/Deprecation.php
```diff
@@ -43,6 +43,16 @@ class Deprecation {
 	protected static $version;
 
 	/**
+	 * Override whether deprecation is enabled. If null, then fallback to 
+	 * SS_DEPRECATION_ENABLED, and then true if not defined.
+	 * 
+	 * Deprecation is only available on dev.
```
hafriedlander: Should probably add a note here that we can't use Config system because of infinite loop risk.

tractorcow: Added in updated push.

tractorcow:
Updated @hafriedlander

kinglozzer:
Sorry I’m a few days late to the deprecation party...

+1 for on by default in dev, off in everything else (I think there’s already a consensus on that anyway...)

>So, what does this mean for Deprecation::notice($version), should $version say when we expect to take it out or when it was deprecated? The former means devs will have to actively change their notice settings to ever be warned, the latter doesn't

Do we definitely have an agreement on this? I’m confused about who is voting for which approach.

I’ve changed my mind and am now advocating the *“`$version` = when it was first deprecated”* approach, so that devs *don’t* have to manually change their notice version to see the warnings. The only downside I can see is that if I drop in a 3.2 update to a 3.1 site, I may start seeing deprecation notices popping up. However, that would really be a non-issue if this PR is merged as they’d be easy to turn off.

tractorcow:
I think this is a bit late for 3.2; It's ok to look at this in 3.3, or otherwise just 4.

dhensby:
We should be setting the version to when it was deprecated, not when it will be taken out. There's no obligation to take deprecated functions out at any specific release (though it can only be major ones)

dhensby:
I don't think this is too late for 3.2. 

Its an issue that's been bugging the community for a while and if it waits to 3.3, people could be waiting until next year for this to make it into a stable release

tractorcow:
Closing and re-opening on 3.2 branch.

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/pull/4403

PR #4265, opened 2015-06-10
Fix the order of params

nfauchelle:
The parameters for the strpos and substr functions were wrong. strpos always failed so never triggered the substr.

This would mean $fieldClass would be left with 'Varchar(255)'

Fixing this had no noticeable speed or memory effect.


dhensby:
Wow, classic.

Seems like it was an edge-case waiting to happen. I think it's not a problem as any of the classes that do implement that interface don't take arguments?!

dhensby:
Thanks, though.

PR #4268, opened 2015-06-10
Fix nesting of config in tests to prevent iterative nesting

tractorcow:
Director::test was sometimes not unnesting config, meaning that config wasn't being cleaned up / collected properly.

Also fixed the custom nesting in DataObjectTest


dhensby:
https://github.com/silverstripe/silverstripe-framework/pull/3979 will solve some of these problems...

tractorcow:
See my comment https://github.com/silverstripe/silverstripe-framework/pull/3979#issuecomment-111324369

This fix simply fixes a few known issues in the least risky way possible.

tractorcow:
I'm going to close this and fix this in 3.1 on top of #3979

PR #4272, opened 2015-06-11
FIX TreeDropdownField doesnt change label on unselect

dhensby:
This fixes a front-end regression in the CMS where changing a drop down to an unselected value appeared to make no difference:

Before:

After:

### DIFF
--- a/javascript/TreeDropdownField.js
+++ b/javascript/TreeDropdownField.js
@@ -154,16 +154,20 @@
 				var updateFn = function() {
 					var val = self.getValue();
 					if(val) {
-						
+
 						var node = tree.find('*[data-id="' + val + '"]'),
 							title = node.children('a').find("span.jstree_pageicon")?node.children('a').find("span.item").html():null;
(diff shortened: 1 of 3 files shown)


dhensby:
This needs merging into 3 branch too

javascript/TreeDropdownField.js
```diff
@@ -154,17 +154,21 @@
...
-						
+
 						if(title) {
 							self.setTitle(title);
 							self.data('title', title);
 						}
 						if(node) tree.jstree('select_node', node);
 					}
+					else {
+						self.setTitle(self.data('empty-title'));
```
tractorcow: `self.data('emptyTitle')`

https://api.jquery.com/data/

dhensby: Either will work, if you do camelCase then jQuery will look for the camelCase then the dash-separated names. So it's actually more performant this way

The native JS data API requires camelCase, though.

I can change it, but it's ultimately not going to change the behaviour and be fractionally slower

dhensby:
thanks

PR #4273, opened 2015-06-11
FIX: increase_time_limit_to work when increase_time_limit_max is not yet set

stevie-mayhew:
Ran into this problem when running a task through a CMS action. Not sure if this is a necessarily a good fix, but if we need to throw a user error or something instead let me know :smile: 

Perhaps we can just set the global $_increase_time_limit_max to -1 before this function runs?

### DIFF
--- a/core/Core.php
+++ b/core/Core.php
@@ -27,6 +27,9 @@ error_reporting(E_ALL | E_STRICT);
  */
 require_once dirname(__FILE__).'/Constants.php';
 
+global $_increase_time_limit_max;
+$_increase_time_limit_max = -1;
+
 /**
  * Priorities definition. These constants are used in calls to _t() as an optional argument
  */
@@ -255,7 +258,7 @@ function translate_memstring($memString) {
  */
 function increase_time_limit_to($timeLimit = null) {
 	$max = get_increase_time_limit_max();
-	if($max != -1 && $timeLimit > $max) return false;
+	if($max != -1 && $max != null && $timeLimit > $max) return false;
 
 	if(!ini_get('safe_mode')) {
 		if(!$timeLimit) {
@@ -273,8 +276,6 @@ function increase_time_limit_to($timeLimit = null) {
 		return false;
 	}
 }
-
-$_increase_time_limit_max = -1;
 
 /**
  * Set the maximum allowed value for {@link increase_timeLimit_to()};


dhensby:
I'm confused, `$_increase_time_limit_max = -1;` is the default.

if your `$_increase_time_limit_max` is falsey, then it's been set that way...

stevie-mayhew:
No, its not set at this point in the execution, its set afterwards. If you check the tests, they explicitly set it to -1 before running.

To replicate, you can add the following to the `mysite/_config.php`

```
increase_time_limit_to(300);
```

Then check in Core.php with the following:

```
	error_log(__FUNCTION__);
	if($max != -1 && $timeLimit > $max) return false;
	error_log(__FUNCTION__ . ' passed call');
```

The second log is never called.

dhensby:
OK, I'll try to find some time to take a look - is it possible to unit test this?!

stevie-mayhew:
@dhensby have you had a chance to look at this?

I'm not sure how to unit test this, if you have a suggestion I can run with I can give it a shot.

sminnee:
A simpler test is putting this into mysite/_config.php

    if(!increase_time_limit_to(300)) die("Can't increase to 300");

I can't think of a good unit test for this because what we're testing is global state, which is usually clobbered by unit testing.

Regarding the bug, the issue is that the default value is set too far down Core.php, after the _config.php files are included. It also wasn't global. This fixes that: https://gist.github.com/40b4e2503b130ae4e9c2

Frankly, I think both fixes would be worth doing, as a falsy value of max can only be interpreted as "no max".

Stevie, would you mind pulling that gist into your PR?

sminnee:
I would also note that the code to set max memory limit is similarly broken. However, that code checks for falsy, and the default maximum memory limit that someone tried to put in is stupid and therefore rightly broken—it sets the maximum limit to the current limit.

sminnee:
As an aside, if we wanted to test this kind of setting, I would think that a 'full server stack test' approach would make the most sense, where we have a testing framework for:

 * Instantiating new projects with composer
 * Copying new files / patches into the project
 * Making assertions

So we'd be able to have a test that says:

 * Create a clean project
 * Put this line into mysite/_config.php `if(!increase_time_limit_to(300)) user_error("Can't increase time");`
 * Visit `http://<site>/dev/`
 * Confirm 200 response and no "Can't increase time" in resulting text

There will be other surface thing that we can test in this way I'd expect, for example testing our .htaccess and nginx config instructions. We would probably want to allow the testing system to make modifications to nginx configuration files and things, so running as root/sudo in some vm/container would be helpful.

stevie-mayhew:
Cool, I've made that change for you @sminnee 

I'm not going to open the testing can of worms with this PR, but if I get time (hah!) will definitely have a look at an environment based testing suite as it does make sense like you've said :smile:

dhensby:
@stevie-mayhew can you confirm you tested locally and this acts as expected?

stevie-mayhew:
@dhensby Yeah have tested it locally and it worked :smile:

dhensby:
thanks

stevie-mayhew:
Thanks!

PR #4274, opened 2015-06-11
Update and fix PHP docblocks, document spelling mistakes and strip trailing whitespace on Form.php

stojg:


stojg:
My editor (PHPStorm) went crazy with the inconsistent method docblocks on the Form class. This cleanup should fix most of the inconsistencies.

forms/Form.php
```diff
@@ -152,17 +184,25 @@ class Form extends RequestHandler {
...
-		'handleField', 
+		'handleField',
 		'httpSubmission',
 		'forTemplate',
 	);
 
 	/**
+	 * @var bool
+	 */
+	protected $securityTokenAdded;
```
stojg: This is a new property, set in https://github.com/silverstripe/silverstripe-framework/pull/4274/files#diff-e7adb20524333364f626b479ceeded9cR663 It really looks like it should be on this object, not a parent object.

dhensby: It seems this var just isn't declared anywhere, is that right? so this is effectively declaring it's visibility.

If it's `@var bool` then you either need to assign it an initial value, or you need to do `@var bool|null`

forms/Form.php
```diff
@@ -486,19 +540,21 @@ public function handleField($request) {
...
-	 * Set whether the user should be redirected back down to the 
-	 * form on the page upon validation errors in the form or if 
+	 * Set whether the user should be redirected back down to the
+	 * form on the page upon validation errors in the form or if
 	 * they just need to redirect back to the page
 	 *
-	 * @param bool Redirect to the form
+	 * @param bool $bool
+	 * @return $this
+	 * @internal param Redirect $bool to the form
```
dhensby: What is this for? I don't know this PHPDoc syntax

The description should go at the end of line 549

dhensby:
looking nice to me. cc @assertchris

forms/Form.php
tractorcow: It looks like a typo?

assertchris: So too late, but PHPStorm adds `@internal` when parameter names change. It could have been removed...

forms/Form.php
assertchris: agree.

forms/Form.php
stojg: Yeah, I missed that one. Oh well.

PR #4275, opened 2015-06-12
Float labels with ss-ui-button class in accordion

hailwood:
Currently things like the upload field etc break when you use them inside a ToggleCompositeField:

Broken

This fixes it by adding the float back

Fixed


adrexia:
(y)

hailwood:
@adrexia Compiled css is good to go.

@tractorcow I've been told you're the best person to notify here.

hailwood:
Quick note, the comment change in the compiled css isn't from my change, either it wasn't compiled previously or it was manually removed from the css before being commited in this commit https://github.com/silverstripe/silverstripe-framework/commit/cb7aed46cdba4763d32a244ba6c76cceb8aec659 by @dhensby

tractorcow:
Squashed and merged with https://github.com/silverstripe/silverstripe-framework/commit/9a278b7613c9c91a8d3f2bb1ccc9f4b6affe24ec

PR #4276, opened 2015-06-12
Add 0 Based Position tip.

hailwood:
Sometimes you need a zero based index, unless you read the api there is no information that $Pos is actually a method that takes a starting position rather than just a property.


tractorcow:
How about putting it inline with the `$Pos` documentation, and simply add it as an addendum that this optional parameter can be added, rather than the method version being a separate feature in itself. :D

hailwood:
@tractorcow are you suggesting

* `$Pos`: The current position in the list (integer). Will start at 1. `$Pos(0)` will start at 0.

tractorcow:
I think making it clear that you can put any integer there that you want; You can even put -4 in if you like. ;D

PR #4277, opened 2015-06-12
Make travis more resilient to composer self-update

dhensby:
`composer self-update` has been failing regularly on travis recently.

As `composer` is already installed and it's not strictly essential to have the very latest version, this change allows the build to continue, even if composer can't self-update

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ matrix:
         - sudo apt-get install -y tidy
 
 before_script:
- - composer self-update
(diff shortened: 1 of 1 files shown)


dhensby:
proof it works:

chillu:
So ` || true` a way to ignore the return code. Fair enough, thanks Dan!

dhensby:
I'll go and add this to some other repos (without PRs to save time) when I can

PR #4279, opened 2015-06-13
API: Add explict permission code for CMSProfileController (Fixes #1613)

### LINKED ISSUE #1613: [2012-10-08] profile link broken for members with restricted access
created by: camfindlay
created at: 2012-10-08
original ticket: http://open.silverstripe.org/ticket/7930

---

This was reported in 2.4.5 in ticket  silverstripe/silverstripe-cms#334 

This is also an issue in SS3.0.2 when a member is restricted to a group that only has access to a modeladmin interface, clicking the link to the members's profile in the top left fails and reloads the current page into the overlay window.

wilr:


dhensby:
Test?

admin/code/CMSProfileController.php
```diff
@@ -59,6 +57,7 @@ public function canView($member = null) {
 		if(
 			!Permission::checkMember($member, "CMS_ACCESS_LeftAndMain")
```
tractorcow: checkMember can take an array of codes to check, as long as the third parameter is left default 'any'

tractorcow: Technically, if the user has ANY CMS_ACCESS_* permission, they should be able to use this feature. It's still good to have this explicit permission, but unfortunately the permission model doesn't support nested permissions yet (e.g. CMS_ACCESS_CMSMain is a child of CMS_ACCESS).

tractorcow:
All in all I think this is a slight improvement and is probably good enough for 3.x, but we definitely need to address the limitations in cms permissions in 4.x in a more significant way.

tractorcow:
I'll merge if you add tests. :)

tractorcow:
Ping @wilr :D

dhensby:
ping @wilr

tractorcow:
This was fixed ages ago by @dhensby I believe. https://github.com/silverstripe/silverstripe-framework/pull/4543

Now anyone with any CMS permission can access the profile page, so I think this fix is outdated.

dhensby:
Its true that I fixed an issue so users can access this area, but there is no specific permission for doing so (which this was adding). 

If @willr is interested in finishing this, I think it still makes sense to have the explicit permission?! Maybe?

tractorcow:
Ok, if @wilr wants to update this PR I'll re-open it. :D

wilr:
I'll investigate on my return into the country. Currently minus laptop. Just the test we're worried about? Happy with the permission code. I also worry potentially it's an upgrade issue, I.e special groups won't have this permission code by default. Think anything needs to be included for that case?

tractorcow:
You'll need to test that the permission code grants the user CMS access, but only to the profile section and not any other area I guess.

It would be a risk if that permission suddenly granted them other cms area rights.

PR #4281, opened 2015-06-14
Fixing doc block issues

dhensby:
Continuation of #4274 because "oh well" isn't good enough :P

### DIFF
--- a/forms/Form.php
+++ b/forms/Form.php
@@ -55,27 +55,27 @@ class Form extends RequestHandler {
 	public $IncludeFormTag = true;
 
 	/**
-	 * @var FieldList
+	 * @var FieldList|null
 	 */
 	protected $fields;
 
 	/**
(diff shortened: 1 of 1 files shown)


forms/Form.php
```diff
@@ -546,9 +546,8 @@ public function makeReadonly() {
 	 * form on the page upon validation errors in the form or if
 	 * they just need to redirect back to the page
 	 *
-	 * @param bool $bool
+	 * @param bool $bool Redirect to form on error?
```
chillu: Why is this a question?

dhensby: The bool answers the Q...

dhensby: But, yes, is a bit odd... I could remove the `?` ?

chillu:
All good, I think it was just a case of still being half asleep when looking at it ;)

PR #4270, opened 2015-06-11
Cleanup 3.2 changelog for release

tractorcow:
- Added notes on upgrading usage of SQLQuery::getWhere
- Removed tabs from second level lists
- Consolidated docs from rc1/3.2.0.md
- Added contents section
- Grouped New and changed API together

Fixes #3533

PR #4283, opened 2015-06-15
API make SQLQuery strict semver for 3.2

tractorcow:
Since SQLQuery was changed in a way that WASN'T properly semver compatible I've attempted to rewrite this class a bit so that it behaves as it once did.

This includes the following changes to keep it in line with 3.1:

* `SQLQuery::getWhere` returns an array of strings. This means that (sadly) prepared parameters will be returned merged into the predicates. This does NOT internally merge these parameters; It only ensures the returned values for this method are in the old format.
* `SQLQuery::getDelete` / `setDelete` are restored.

In order to support users of this old class with the new ORM, you can use `SQLQuery::toAppropriateExpression()` in order to automatically convert `SQLQuery` into `SQLSelect` or SQLDelete automatically, where the behaviour of these objects is in line with the new API.

Since we already have an open issue for the documentation of 3.2 changes and upgrading notes, I'll update the changes at https://github.com/silverstripe/silverstripe-framework/pull/4270 instead of in this PR.

CC @hafriedlander

### DIFF
--- a/model/queries/SQLQuery.php
+++ b/model/queries/SQLQuery.php
@@ -11,6 +11,14 @@
 class SQLQuery extends SQLSelect {
 
 	/**
+	 * If this is true, this statement will delete rather than select.
+	 * 
+	 * @deprecated since version 4.0
+	 * @var boolean
+	 */
+	protected $isDelete = false;
+
+	/**
 	 * @deprecated since version 4.0
 	 */
 	public function __construct($select = "*", $from = array(), $where = array(), $orderby = array(),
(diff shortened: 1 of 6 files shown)


model/queries/SQLQuery.php
```diff
@@ -21,19 +29,155 @@ public function __construct($select = "*", $from = array(), $where = array(), $o
...
+	}
+
+	/**
+	 * Given a list of parameterised conditions, return a flattened
+	 * list of condition strings
+	 * 
+	 * @param array $conditions
+	 * @return array
+	 */
+	protected function injectConditions($conditions) {
```
tractorcow: Here be dragons!

tractorcow:
I've rebased this against the changelog PR at #4270 and ensured the upgrading notes were updated for the fixed API.

tractorcow:
Refactored parameter injection into a class at @hafriedlander 's request.

PR #4284, opened 2015-06-15
Upgrading Chosen JS plugin

dhensby:
The version of Chosen we currently use in admin is very old and has
performance issues with large data-sets.


kinglozzer:
@dhensby Have you replaced the custom stuff that’s in there? e.g. https://github.com/silverstripe/silverstripe-framework/blob/3.1/admin/thirdparty/chosen/chosen/chosen.jquery.js#L1078-L1106

dhensby:
You're kidding me?! sheez - no...

I take it that this makes it show above the input if there isn't enough space below? I'll have to test this by hand - surely they've got a solution for this in house?

FYI in the future SS, just implement these functions in a custom script, rather than hacking external libraries!!!

dhensby:
OK, it doesn't flip upwards, unfortunately...

But it does push the scroll down...

dhensby:
To be honest, I don't think it flipping to the top is a good behaviour.

It's easy to miss the search box, it's badly aligned and it's just a bit weird.

stevie-mayhew:
A reasonably fast replacement field can be found at https://github.com/sheadawson/silverstripe-select2

dhensby:
> A reasonably fast replacement field can be found at https://github.com/sheadawson/silverstripe-select2

Nice, though I don't have time to integrate something like that - I was hoping for a drop-in upgrade.

I'm working with dropdown with datasets for ~5,000 and it's crashing.

chillu:
List of customisations: https://github.com/silverstripe/silverstripe-framework/commits/3.1/admin/thirdparty/chosen/chosen/chosen.jquery.js

Agree that it's not ideal. Some of the commits were backports from pull requests, some were just lazy and changed the source code without submitting a fix to the upstream library at the same time.

Regarding a select2 replacement: Looking at the github repos for both projects (https://github.com/select2/select2 and https://github.com/harvesthq/chosen), they both have an incredible amount of open issues, pull requests and general activity. Chosen was the cool thing to use three years ago, now it seems like Select2 has a slight edge. We've spent a lot of time on details like correct sizing in hidden tabs, so I'd want there to be a good reason for a switch. 

Also, if you're using a dropdown with 5000 items, you should use ajax ;)

dhensby:
> Also, if you're using a dropdown with 5000 items, you should use ajax ;)

I would if the response times weren't > 500ms :P

dhensby:
I am not going near this

adrexia:
I thought chosen had fixed all our issues anyway, but it looks like the rise up thing still has an open pull request: https://github.com/harvesthq/chosen/pull/2312

dhensby:
@adrexia It may be the case that the version I was trying to use fixed all the "issues" we had (except the "drop up") but the issues we had aren't documented anywhere, the History on that file in the current code is a nightmare to follow and it's way too much effort for me to look into for something that really should have been trivial.

Personally, I'd drop the support for "drop up" in favour of having a new version of the plugin.

Also, that PR you linked is against master and we're running v0.9 and the plugin is currently on v1 and so upgrading to that would be even more hell than this "simple" upgrade to v0.14

adrexia:
There are two changes in that file that silverstripe made that didn't come from upstream. One was merging a community patch that had sat there for a while ignored by the product owners, and the other was the drop-up functionality (which the chosen community had also been asking for/ supplying code for quite some time, ignored). Both have surrounding code blocks. However, we also manually ported upstream back to that file without the git history. I manually compared against chosen the last time. Then about 6 months later chosen changed their entire API, which made updating less than trivial.

The version of chosen in silverstripe doesn't deal at all with hitting the bottom of the window (no scroll, no dropup, just bugs). I'm also in favour of ditching anything that chosen now has a better way of dealing with. They have also added a lot more event hooks now which make adding custom functionality elsewhere possible. 

The only place I think it might cause trouble is the screen size chooser/mode picker in preview. Chosen have provided a better way to include images now, but we still need the drop down to go up in that case. If there is an event fired in the right place, it won't be a problem. I might take what you've done and look into it.

dhensby:
@adrexia I looked at this: https://github.com/silverstripe/silverstripe-framework/commits/3.1/admin/thirdparty/chosen/chosen/chosen.jquery.js

There have been lots of changes that either were retained or overridden over time, I'm not sure which and didn't have the inclination to look into.

I just felt it was too much risk to push the upgrade and with the missing feature of "drop up" it's probably not worth the fight and I'm not keen to hack the new version to get it in.

When I tried the upgrade myself, it seemed fine, the dropdown just pushed the height of the page if there wasn't enough room and was functional. I hadn't noticed any regressions in other parts... but then again, I didn't know it was used for the "screen size chooser/mode picker in preview"

adrexia:
Sweet. It would be good to get it in. I'll do some more testing.

dhensby:
@adrexia I've restored my branch (in case that makes it easier for you to checkout or cherry-pick).

Let me know if you think it's worth re-opening and if I have a supporter in you for abandoning our hacked changes.

I did look at the JS and couldn't find any hooks. Unfortunately for us, they don't leak a Chose object into the global space, so we can't "overload" or hook into functions (as I'd hoped to) in a modular way, I didn't notice any hooks, eiter.

adrexia:
I realised when it came to looking at this that 0.14.0 is still an old chosen version (pre-api changes). Which is what we need in 3.1, but it's the later changes that I was hopeful would fix our bugs. 

I can confirm that it's working with a couple of exceptions: 
1. the preview selectors
2. iframe based composite style fields (c.f. uploadfield). 

iframes, unfortunately, do not produce scrollbars when chosen goes outside their parameters. The drop up functionality as a style choice is needed for preview without doing something drastic like moving controls to the top of the preview. I do think we can fix this more elegantly using the post 1.0 versions of chosen, but this version has all of our old customisation limitations. 

At this point it might be best to backport our fixes onto 0.14.0* and upgrade properly in 3.2. Do you want me to give it a go? 

--
*  I know it just compounds the history state issue. It already looks deceptively more complex than it is**, but aside from creating a new file and reapplying our patches, I'm unsure what more we can do?

** It looks a bit like: upgrade chosen -> apply patches -> upgrade chosen -> patch our patches -> upgrade chosen. But our actual unique change set is pretty minimal.

adrexia:
Looks like the width fix is now in core, but I hit an issue when I tried to port the rise up functionality. I suspect that the very same reason this is being upgraded is the reason the rise-up functionality won't work anymore: the list is only being  filled _after_ interaction. The places rise up are currently being called are before the lists have anything in them. I have it working, but only after the second interaction with the same list.

dhensby:
Can't we make an assumption that it will always be the max-height and then make the calculation based on that? 

I really want the upgrade because they no longer itterate over item and instatiate jquery objects for each one on CLOSE of the dropdown

adrexia:
That would mean there would be a gap between the dropdown the the dropdown toggle (it adds a rise-up class, but it also sets a negative top property). But I am looking at places the fix could be moved to where it may work. We might just need to handle the flicker of not-quite-right and alter the list position after chosen is down with it. I think we do this for some stuff on the preview toggles currently anyway.

I pushed the current state here: https://github.com/adrexia/framework/tree/pulls/chzn-upgrade. I've been editing the jquery coffee script file, and everything that is ours is surrounded by ###.

Maybe, if we don't care about a slight flicker that we may have to cope with anyway, our adjustments can move somewhere else now.

dhensby:
I don't want to be a party pooper and I really love the effort going into this, but wouldn't it be better to put that effort in upgrading to the latest version (v1.x) of chosen, rather than hacking again at a legacy version?

adrexia:
That up there is just ported the fixes back to the version you upgraded to. Hopefully with the newer version we wouldn't need the hacks at all. 

Though, given the amount of problems chosen has caused over the years, and the way the project is run, perhaps we should look at select2 if we were to go that way? It's more flexible, has more functionality, and is easier to work with.

dhensby:
Sure, I think at this point it's whoever has the inclination / wants to put the effort in can do that.

I reckon it's the same amount of effort as upgrading to latest chosen?

PS: What's the problem with how the project is run? Harvest have always appeared pretty good to me...

adrexia:
When we were looking at this stuff a few years ago there were a lot of community fixes that were just ignored. Many were done incorrectly (editing jquery and prototype files rather than coffee script), but there were others that fixed real community problems that just sat ignored. Some of that may have been that they were changing the API at the time and didn't want to introduce more complexity. I do feel we may use chosen a bit differently to how the core team use it, though they have certainly made things a lot more hook-able in later releases.

_I reckon it's the same amount of effort as upgrading to latest chosen?_

I would say so, yes. It's also something that can't go into a non major release, as there are modules that rely on chosen specifically. And the old API for chosen, at that.

dhensby:
Sure, it'll need to be a 4.0 release

chillu:
> I reckon it's the same amount of effort as upgrading to latest chosen?

Looking at the two github repos, it does look like select2 is more actively maintained. ChosenJS seems to basically ignore all non-trivial pull requests (https://github.com/harvesthq/chosen/pulls?q=is%3Apr+is%3Aclosed) which isn't a good sign. 

I agree with Dan that it any effort might be better spent migrating to select2. Both the ChosenJS upgrade and a select2 switch have to be done properly, but select2 seems to be more future proof. Don't forget dev effort required to adjust Behat tests for select2. I can help with that.

adrexia:
I guess it's about urgency. If there is no demand to upgrade anything in 3.1, then I'm happy to ignore it and move to looking at 4.0 instead.

dhensby:
The current version is completely unsuitable for large lists due to the teardown function and I would love to see this upgraded for my client's sake, but these core hacks make it very hard to do and my knowledge of coffee script is lacking. 

An intirim upgrade of chosen would be my preferred option, but I'm just being pragmatic about the fact it's probably too much effort.

PR #4286, opened 2015-06-16
API Revert DataObject::validate to 3.1 method signature (protected)

tractorcow:
The changes made in https://github.com/silverstripe/silverstripe-framework/pull/2729 were not fully semver compliant, so make this accessible with a `doValidate` method for now (until 4.0 when we can make validate properly public).

### DIFF
--- a/docs/en/04_Changelogs/3.2.0.md
+++ b/docs/en/04_Changelogs/3.2.0.md
@@ -57,7 +57,7 @@
 * Refactor of database connectivity classes into separate components linked together through dependency injection
 * Refactor of `SQLQuery` into separate objects for each query type: `SQLQuery`, `SQLDelete`, `SQLUpdate` and `SQLInsert`
 * PDO is now a standard connector, and is available for all database interfaces
-* `DataObject::validate()` method visibility changed to public
+* `DataObject::doValidate()` method visibility added to access `DataObject::validate` externally
 * `NumericField` now uses HTML5 "number" type instead of "text"
 * `UploadField` "Select from files" shows files in all folders by default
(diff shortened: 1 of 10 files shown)


tractorcow:
Also merge https://github.com/silverstripe/silverstripe-cms/pull/1215

chillu:
We're watering down an API (both `validate()` and `doValidate()` public methods in 4.0) in order to comply to semver? That seems like the wrong approach.

stevie-mayhew:
I'd really like for these to be public, we have had lots of issues this week with the fact that they are not :\

tractorcow:
The validate() method was already protected; We can't water it down, it's already watered. It's wet. I know we don't like this, but we can't break our commitment to semver on that basis alone.

doValidate can be deprecated in 4.0 (`Deprecation::notice('5.0', 'Use validate() instead');`).

tractorcow:
The other alternative is to override __call to expose the method, but that has other issues (discoverability, code completion, documentation, etc).

chillu:
> The validate() method was already protected

So this change to public hasn't been in a stable release yet, right? I thought it was already in place for 3.1, hence my strong reaction. I'm still not very happy about introducing temporary APIs just to deprecate them again in the next major release, but that seems to be preferable to layering on more `__call()` magic to the already have `DataObject` class. Can you raise a PR for the deprecation notice in 4.0?

tractorcow:
Sure, but I'll need to merge this first.

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/pull/4294

PR #4288, opened 2015-06-16
BUG Fix major segfault on PDOConnector after any DDL / Issue in PDOQuery::first()

tractorcow:
A huge amount of testing indicated that the caching of statements will cause a segfault if not cleared after every DDL, since any prepared statement is automatically invalidated.

This could open the door to restoring statement caching for MySQLiConnector, but we can look at that in 3.2-beta2 :D

Also fixed a bug in PDOQuery::first()


tractorcow:
@dhensby this should make everyone happy. :) Is this ok?

dhensby:
Related test failures?

sminnee:
Yeah those test failures look pretty directly related to this change, so we'll need to fix those. Are they happening locally for you, Damian?

Although I can see that bundling the code into previewWrite() is functionally correct, it feels a bit ganky, since we're introducing state manipulation into a function that claims to just be providing information. It's not clear to me why we can't put this code into `PDOConnector::exec()` after its previewWrite() check, as that would seem to be much better. A specific weakness of the current code is that if the previewWrite() functionality is removed, it would break an unrelated feature, the cache cleanup. I think this blocks merge.

While we're on the subject, the structure of the `previewWrite()` function is a little concerning, but I don't think this blocks merge. I've noted that issue in #4289.

tractorcow:
> It's not clear to me why we can't put this code into PDOConnector::exec()

Unfortunately we can't force users to use exec for DDL queries; In fact this method is specific to PDO so even less so.

I'll break it out into a preQuery and call previewWrite from there if that's cleaner?

tractorcow:
Ok discussion with @sminnee and @assertchris yields a desire to move previewWrite and benchmarkQuery into the controller class.

tractorcow:
ok, done. I'll update the other connectors to match this change.

tractorcow:
PostgreSQL update at https://github.com/silverstripe/silverstripe-postgresql/pull/42

Suggestion: Merge postgresql first then restart these tests :D

tractorcow:
SQL Server update at https://github.com/silverstripe/silverstripe-mssql/pull/23

tractorcow:
SQLite3 update at https://github.com/silverstripe-labs/silverstripe-sqlite3/pull/11

tractorcow:
Hopefully this is the last ORMey component to change before 3.2.

@sminnee does this seem to work with you? Happy for you to take over merging all of these if you're happy. ;D

sminnee:
Thanks for that, all merged now :-)

PR #4290, opened 2015-06-16
FIX No longer caching has_one after ID change

dhensby:
Bug where the cache for has_one relations doesn't expire when an relation ID is changed.

eg:

```php
$do = MyDataObject::get()->byID(1);
$do->HasOneRelation()->ID; // 1
$do->HasOneRelationID = 2;
$do->write();
$do->HasOneRelation()->ID; //still 1
```


dhensby:
related: #4188

PR #4293, opened 2015-06-17
Update styles for multi-selection button

scott1702:
To be merged with https://github.com/silverstripe/silverstripe-cms/pull/1218

### DIFF
--- a/admin/scss/_style.scss
+++ b/admin/scss/_style.scss
@@ -906,7 +906,7 @@ body.cms {
 			vertical-align: middle;
 		}
 
-		label {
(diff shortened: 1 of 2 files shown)


admin/scss/_style.scss
```diff
@@ -906,7 +906,7 @@ body.cms {
 			vertical-align: middle;
 		}
 
-		label {
+		span {
```
dhensby: Much better to use a class rather than element selector, then when markup changes we won't have to change our CSS again

scott1702:
Good thinking @dhensby, Updated :)

PR #4294, opened 2015-06-17
API make DataObject::validate public

tractorcow:
Move of https://github.com/silverstripe/silverstripe-framework/pull/2729 to 4.0 to conform to semver

Please also merge https://github.com/silverstripe/silverstripe-cms/pull/1219 at the same time.

### DIFF
--- a/filesystem/File.php
+++ b/filesystem/File.php
@@ -917,7 +917,7 @@ class File extends DataObject {
 		return $labels;
 	}
 
-	protected function validate() {
+	public function validate() {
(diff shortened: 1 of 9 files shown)


willmorgan:
Can't this go in to 3.2 safely? It's adding new functionality in a BC way.

chillu:
@willmorgan See #4286 for context.

tractorcow:
It's a breaking change, because if anyone had overridden validate in their subclass, they would now throw errors in PHP due to mismatch of method visibility.

PR #4300, opened 2015-06-18
Bug: Page in SiteTree is incorrectly moved on reload in some cases

swilsonalfa:
In some cases, the selected page in the site tree will not return to it's correct position when the page is reloaded/refreshed by the browser. If/when the subtree is loaded via ajax (I've identified about 5 subtree's deep), the javascript tells the selected page to be created in the "last" position (see LeftAndMain.Tree.js:267). This will only happen if the page is reloaded via the browser and the subtree has to load via ajax.

The fix is passing through the Sort field from the database via updatetreenodes and use that value to tell jstree create_node where to place the newly created "node"/page.

### DIFF
--- a/admin/code/LeftAndMain.php
+++ b/admin/code/LeftAndMain.php
@@ -976,6 +976,7 @@ class LeftAndMain extends Controller implements PermissionProvider {
 			$data[$id] = array(
 				'html' => $html, 
 				'ParentID' => $record->ParentID,
+				'Sort' => $record->Sort,
 				'NextID' => $next ? $next->ID : null,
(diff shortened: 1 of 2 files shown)


dhensby:
This sounds like a big fix, but it does introduce a new API. 

As 3.2.0-beta.1 is imminant, please can you PR this against 3.2 branch, that way it'll be in stable code soon and there won't be any arguments around semver. 

Thanks.

swilsonalfa:
New PR against 3.2: https://github.com/silverstripe/silverstripe-framework/pull/4301

PR #4310, opened 2015-06-19
Set deprecation level for all changes in 3.x to 4.0

tractorcow:
cc @silverstripe/core-team @hafriedlander

### DIFF
--- a/dev/SapphireTest.php
+++ b/dev/SapphireTest.php
@@ -43,7 +43,11 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 	 */
 	protected $usesDatabase = null;
 
+	/**
+	 * @deprecated since version 4.0
+	 */
 	protected $originalMailer;
+	
 	protected $originalMemberPasswordValidator;
 	protected $originalRequirements;
 	protected $originalIsRunningTest;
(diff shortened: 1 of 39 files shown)


tractorcow:
Also includes fix for:

* FormAction deprecation flagging on null !== '' (which it shouldn't)
* Use of deprecated api
* Rely on Injector::unnest to restore mailer rather than explicitly restoring it

dev/SapphireTest.php
```diff
@@ -221,7 +221,7 @@ public function setUp() {
 		// Set up email
 		$this->originalMailer = Email::mailer();
```
hafriedlander: Seems like we don't need this now?

email/Email.php
```diff
@@ -62,10 +62,10 @@ class Email extends ViewableData {
 	protected $bcc;
 
 	/**
-	 * @deprecated since version 3.3
+	 * @deprecated since version 4.0.0
```
hafriedlander: Why the extra .0?

forms/FormAction.php
```diff
@@ -100,7 +100,7 @@ public function Title() {
 
 		// Remove this method override in 4.0
 		$decoded = Convert::xml2raw($title);
-		if($decoded !== $title) {
+		if($decoded != $title) {
```
hafriedlander: Why the loose equality? Could we cast title to string if it might not be a string instead?

dev/SapphireTest.php
tractorcow: If unit tests relied on $this->originalMailer then they would fail if this was removed.

email/Email.php
tractorcow: Probably doesn't need to be there; Doesn't hurt if it is. :)

forms/FormAction.php
tractorcow: Because we only want to deprecate if they are setting the title to something like `Left &amp; Main`, because `'Left & Main' !== 'Left &amp; Main'`. Not if they set it to false, and `false !== Convert::xml2raw(false)` or `null !== Convert::xml2raw(null)`.

PR #4311, opened 2015-06-19
Install - Check if mysite/_config/config.yml is writeable

JayDevlin:
In 3.2b1, the installer likes to write changes to mysite/_config/config.yml. We should check if this file is writeable first.

Bug:
- `Couldn't write to file /silverstripe-dev-3.2/mysite/_config/config.yml`
- breaks install process


dhensby:
looks legit, thanks

PR #4312, opened 2015-06-19
FIX: docs not included in composer package installs

sminnee:
This change will improve the performance of package-download-based installs of composer packages, by excluding docs. It does this by excluding them from the output of the "git archive" command.

It uses the export-ignore git attribute to do this.

The suggestion came from https://github.com/composer/composer/issues/1750

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+docs/ export-ignore


sminnee:
I'm keen to hear whether other people see any downsides to this. Note also that we could consider wether we want to exclude tests/ as well. Although some users might want to run package tests, they always have the option of installing with `--prefer-source` to do this.

Note that I'm expecting that these changes will flow through to the .tar.gz and .zip downloads on silverstripe.org, once https://github.com/silverstripe/silverstripe-buildtools/issues/8 is done.

chillu:
I'm in favour of removing both `docs/` and `tests/`. If you run tests, it's likely because you're contributing, in which case you can't work off the archive anyway.

PR #4315, opened 2015-06-20
Update 06_Core_committers.md

SpiritLevel:
Rewording, grammar.


dhensby:
Sad face?

SpiritLevel:
Huh ? I closed it because I missed a mistake and I also noticed that it was duplicated....so, no sad face ;)

PR #4320, opened 2015-06-20
DOCS Fixing docs (and bad API usage)

dhensby:
Docs are wrong and some bad code that isn't using HTMLText where it should be

@assertchris @sminnee


sminnee:
Do it in 3.2.

PR #4321, opened 2015-06-20
DOCS Fixing docs (and bad API usage)

dhensby:
see #4320

### DIFF
--- a/forms/ConfirmedPasswordField.php
+++ b/forms/ConfirmedPasswordField.php
@@ -120,7 +120,7 @@ class ConfirmedPasswordField extends FormField {
 	/**
 	 * @param array $properties
 	 *
-	 * @return string
+	 * @return HTMLText
 	 */
 	public function Field($properties = array()) {
 		Requirements::javascript(FRAMEWORK_DIR . '/thirdparty/jquery/jquery.js');
--- a/forms/DatalessField.php
+++ b/forms/DatalessField.php
@@ -31,6 +31,8 @@ class DatalessField extends FormField {
 	/**
 	 * Returns the field's representation in the form.
 	 * For dataless fields, this defaults to $Field.
+	 *
+	 * @return HTMLText
 	 */
 	public function FieldHolder($properties = array()) {
 		return $this->Field($properties);
--- a/forms/DatetimeField.php
+++ b/forms/DatetimeField.php
@@ -88,7 +88,10 @@ class DatetimeField extends FormField {
 		return $this;
 	}
 
-
+	/**
+	 * @param array $properties
+	 * @return HTMLText
+	 */
 	public function FieldHolder($properties = array()) {
 		$config = array(
 			'datetimeorder' => $this->getConfig('datetimeorder'),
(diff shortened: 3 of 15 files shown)


tractorcow:
My feeling is that we should keep the documented API for these methods `@return string` and use `$casting` to cast this method to HTMLText instead. string is the documented (external) api and it would be risky to change it.

User code should not be relying on the fact that we have in some cases been returning objects that only cast to the correct value (via `DBField::__toString`).

I know this looks like we are fixing the docs to match the api, but it would still effectively be an API change.

willmorgan:
Sidenote: is there some CI task that can run some static code analysis to stop us from writing bad PHPDoc in the future?

dhensby:
Isn't it called scrutinizr

dhensby:
Well, we have a problem where the return type varies depending on the class of form field you are using. That's pretty gross.

This is an oversight from when the casting changes were made a while back and so, in my opinion, is a bug.

I also don't think it's hugely "high risk" as these methods are used mainly in the template AND the `__toString` means anyone treating the return value as a string shouldn't have any problems.

dhensby:
@sminnee your view?

tractorcow:
I think the problem is that we are increasing the promise of the API. Before anyone could write a method implementation that returns either a string, or something that could be cast to a string.

Now if they do `return "<div>My field</dive>"` they are breaking the API where they weren't before.

dhensby:
My point is that we are breaking the API with the current docs. 

I'm not that bothered, its a matter of interpretation and how likely someone overloading and returning a strong anyway is going to break anything (or us returning an object when they want a string) whilst taking into account that any operations that would be string manipulations would still work. 

I think we agree this needs to be fixed properly in 4 and that's on my list (still)

tractorcow:
I think it's really a matter of defining what our conventions are. My feeling is that wrapping everything in casted types at the property-getter level makes things like string manipulation harder, and the general complexity of code is increased. 

`$casting` should be enough in 99% of all situations. Especially in 4.0, when we make casting mandatory for all properties accessed by templates, it'll be even more necessary to ensure these are defined.

dhensby:
I think I agree.

But that doesn't mean we aren't currently supplying an arbitrarily varying API and that we shouldn't fix a previous oversight (even if we disagree with it).

Feel free to close or merge this :)

tractorcow:
Well, if those are my choices, I'd close this, and say we should fix it `properly` (tm) in 4.0

@sminnee do you want to give a tiebreaker decision, or are you happy for me to close?

tractorcow:
Discussing with @sminnee we agreed to merge this, but to move the argument to 4.0. There's still a difference in opinion though. :)

tractorcow:
DIscussion on this top moved to https://github.com/silverstripe/silverstripe-framework/pull/3924#issuecomment-126177522

tractorcow:
Doh, I tracked down the "broken InlineFormAction" in 3.2 to this pull request. :D

Should have used FormField::create_tag not FormField::create, and forgot 'type = submit'

PR #4324, opened 2015-06-20
FIX Imported namespaces now correctly used to determine class inhertance

### LINKED ISSUE #3707: Namespaced PermissionProvider implementors are not detected

dhensby:
Fixes #3707 

see #3899

### DIFF
--- a/core/manifest/ClassManifest.php
+++ b/core/manifest/ClassManifest.php
@@ -107,6 +107,33 @@ class SS_ClassManifest {
 	}
 
 	/**
+	 * Create a {@link TokenisedRegularExpression} that extracts the namespaces imported with the 'use' keyword
+	 *
+	 * This searches symbols for a `use` followed by 1 or more namespaces which are optionally aliased using the `as`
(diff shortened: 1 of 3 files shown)


dhensby:
@assertchris  :) namespacing, whoo

tractorcow:
Much more lovely code @dhensby :D Sorry for giving you pain over your other PR.

tractorcow:
Should this fix be back-ported to 3.1?

dhensby:
@sminee was of the opinion that only the most low risk bug fixes should go into 3.1, which makes sense. The upgrade to 3.2 should be quite frictionless. 

Though, if when you release the next patch to 3.1, you want it in there.. Then why not?

PR #3502, opened 2014-09-22
Allow customising the HTTP Cache-Control defaults.

jedateach:
This is my proposed fix for #3501. It allows the default `Cache-Control` header to be customised using a yaml config array. You can choose true/false to enable or disable control settings, or you can set values.
For example:
```yaml
HTTP:
  CacheControl:
    max-age: 0
    must-revalidate: true
    no-transform: false
```
Would produce `Cache-Control: max-age=0, must-revalidate`

**EDIT: I've noticed that this would actually produce:**  `Cache-Control: must-revalidate`, because `0` is treated as false. Keen to hear thoughts before attempting to fix.

PR #4325, opened 2015-06-20
Fixing issues with HTTP cache control

dhensby:
Fixing issue introduced in #3502 

Please let this be a lesson in why we need test coverage.

The tests aren't complete, but they at least check some fundamentals.

This class needs to be go in 4.0 ;)

### DIFF
--- a/control/HTTP.php
+++ b/control/HTTP.php
@@ -308,7 +308,7 @@ class HTTP {
 	/**
 	 * Add the appropriate caching headers to the response, including If-Modified-Since / 304 handling.
 	 *
-	 * @param SS_HTTPResponse The SS_HTTPResponse object to augment.  Omitted the argument or passing a string is
+	 * @param SS_HTTPResponse $body The SS_HTTPResponse object to augment.  Omitted the argument or passing a string is
 	 *                            deprecated; in these cases, the headers are output directly.
 	 */
 	public static function add_cache_headers($body = null) {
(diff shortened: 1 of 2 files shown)


tests/control/HTTPTest.php
```diff
@@ -7,6 +7,26 @@
...
+		HTTP::set_cache_age(30);
+		HTTP::add_cache_headers($response);
+
+		$this->assertNotEmpty($response->getHeader('Cache-Control'));
+
+		Config::inst()->update('Director', 'environment_type', 'dev');
+		HTTP::add_cache_headers($response);
+		$this->assertContains('max-age=0', $response->getHeader('Cache-Control'));
+
+		Config::inst()->update('Director', 'environment_type', 'live');
```
kinglozzer: Do we need to change this back to `dev` after this test has run?

control/HTTP.php
```diff
@@ -383,13 +379,16 @@ public static function add_cache_headers($body = null) {
...
-			} elseif(is_bool($value) || $value === "true") {
+			} elseif((is_bool($value) && $value) || $value === "true") {
 				$cacheControlHeaders[$header] = $header;
 			} else {
 				$cacheControlHeaders[$header] = $header."=".$value;
 			}
 		}
 
+		$responseHeaders['Cache-Control'] = implode(', ', $cacheControlHeaders);
+		unset($cacheControlHeaders, $header, $value);
```
kinglozzer: What’s this for?

chillu:
Note that the controllerpolicy module is overwriting `Cache-Control`, but I don't think the `HTTP` changes alter the module behaviour (https://github.com/silverstripe-labs/silverstripe-controllerpolicy/blob/master/code/policies/CachingPolicy.php#L46)

tests/control/HTTPTest.php
dhensby: No, tests now nest and unnest config and injector automatically

control/HTTP.php
dhensby: freeing variables. $header and $value are generic and can be used again later

tests/control/HTTPTest.php
kinglozzer: Ooh yes, good point! A worry of the past :)

control/HTTP.php
kinglozzer: At the risk of sounding like an idiot: used where? The method doesn’t use any of those variables any more. Just trying to understand why it’s necessary

kinglozzer:
Yeah this shouldn’t alter that behaviour. `RequestProcess::postRequest()` is called after this, and the misleadingly named method `SS_HTTPResponse::addHeader()` will override anything already set here.

control/HTTP.php
dhensby: It's not really, it's just being defensive - I find it can be quite common for people to accidentally reuse vars after loops like this.

dhensby:
Let's get this in then ;)

dhensby:
Thanks :)

Issue #4295, opened 2015-06-18
Resampled prefixes (e.g. SetWidth400) in folder instead of filename


JorisDebonnet:
When people download a picture from a SilverStripe website, they've got a high chance of it having a prefix such as SetWidth400- or SetRatioSize800800- etc, before the actual filename. Those prefixes don't mean anything to them, it all looks very code-ish, people don't like it. 

There's actually a pretty simple solution for it: instead of the resampled files going into `_resampled/SetWidth400-[filename]` they could go into `_resampled/SetWidth400/[filename]`. The URL is still just as unique, but when people download it, they get the original sense-making filename. 

Is this already possible? If not, can we make it possible, perhaps even the default? It's prettier and no one gets hurt! :)

jonom:
I really like this idea, especially since image filenames are going to look even worse in 3.2 onwards due to the arguments being encoded.

An example filename from 3 branch is `ResizedImageWzEyMDAsODAwXQ==-Sample-IMG-Apples` so would just have to make sure that those encoded arguments will always be safe for directory names. On my machine it looks like the arguments are always alphanumeric except for the `=` signs. I tested the `/Uploads/_resampled/ResizedImage/WzEyMDAsODAwXQ==/Sample-IMG-Apples.jpg` structure and it worked fine so I'm definitely keen to see this idea explored more!

JorisDebonnet:
Wow, I didn't even know about that 3.2 change yet! Let's hope we can make this change land before a release candidate is released then. I'll see if I can do what it takes...

jonom:
I think @tractorcow is aiming to get 3.2 beta 1 out within the next week... Good luck! :+1:

jonom:
I think @tractorcow was aiming to get 3.2 beta 1 out within the week - good luck! :+1: (no sarcasm)

By the way there was [some debate](https://github.com/silverstripe/silverstripe-framework/pull/4229#issuecomment-105918366) about whether changing filename output constitutes a breaking change, don't think we reached a consensus. I take the side that this should be okay, because the API doesn't make any promises about the file path for the images that will be generated. Plus in this particular case going from 3.1 to 3.2 the argument encoding is already changing the filenames of all resampled images, and I renamed a bunch of image methods, so it's already pretty jumbled up :)

JorisDebonnet:
[...]

PR #4328, opened 2015-06-21
Save resampled images into a folder structure indicating transformations

JorisDebonnet:
This is a solution for issue #4295. 

Resampled images are now saved into a folder structure that indicates the image transformations that have happened, rather than creating a long filename prefix. 

This change should be transparent to users, so I'm hoping this may make it into 3.2. I have tracked down all implications I could find, but will be happy to create additional patches if any issues come up during beta testing. 

(Note that `Image::getGeneratedImages` uses the SPL function `RecursiveIteratorIterator`. SPL is included and compiled by default since PHP 5.0, and can no longer be disabled since PHP 5.3.)

### DIFF
--- a/filesystem/Filesystem.php
+++ b/filesystem/Filesystem.php
@@ -79,6 +79,29 @@ class Filesystem extends Object {
 	}
 
 	/**
+	 * Remove a directory, but only if it is empty.
+	 *
+	 * @param String $folder Absolute folder path
+	 * @param Boolean $recursive Remove contained empty folders before attempting to remove this one
+	 * @return Boolean True on success, false on failure. 
+	 */
+	public static function removeFolderIfEmpty($folder, $recursive = true) {
+		if (!is_readable($folder)) return false;
+		$handle = opendir($folder);
+		while (false !== ($entry = readdir($handle))) {
+			if ($entry != "." && $entry != "..") {
+				// if an empty folder is detected, remove that one first and move on
+				if($recursive && is_dir($entry) && self::removeFolderIfEmpty($entry)) continue;
+				// if a file was encountered, or a subdirectory was not empty, return false.
+				return false;
+			}
+		}
+		// if we are still here, the folder is empty. 
+		rmdir($folder);
+		return true;
+	}
+
+	/**
 	 * Cleanup function to reset all the Filename fields.  Visit File/fixfiles to call.
 	 */
 	public function fixfiles() {
--- a/filesystem/File.php
+++ b/filesystem/File.php
@@ -229,7 +229,7 @@ class File extends DataObject {
 	 */
 	public static function find($filename) {
 		// Get the base file if $filename points to a resampled file
-		$filename = preg_replace('/_resampled\/[^-]+-/', '', $filename);
+		$filename = preg_replace('/_resampled\/(.+\/|[^-]+-)/', '', $filename);
 
 		// Split to folders and the actual filename, and traverse the structure.
 		$parts = explode("/", $filename);
--- a/forms/HtmlEditorField.php
+++ b/forms/HtmlEditorField.php
@@ -461,7 +461,7 @@ class HtmlEditorField_Toolbar extends RequestHandler {
 				));
 			} else {
 				$url = Director::makeRelative($request->getVar('FileURL'));
-				$url = preg_replace('/_resampled\/[^-]+-/', '', $url);
+				$url = preg_replace('/_resampled\/(.+\/|[^-]+-)/', '', $url);
 				$file = File::get()->filter('Filename', $url)->first();
 				if(!$file) $file = new File(array(
 					'Title' => basename($url),
--- a/model/Image.php
+++ b/model/Image.php
@@ -701,7 +701,7 @@ class Image extends File implements Flushable {
 	}
 
 	/**
-	 * Return the filename for the cached image, given it's format name and arguments.
+	 * Return the filename for the cached image, given its format name and arguments.
 	 * @param string $format The format name.
 	 * @return string
 	 * @throws InvalidArgumentException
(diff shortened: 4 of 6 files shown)


JorisDebonnet:
(closed and reopened to retrigger a Travis CI build, since one of them timed out... it should be all fine since the problematic one built succesfully before)

jonom:
Heck yeah! Awesome work. I tried it out locally and worked great for me. The old resampled images will hang around until manually deleted, not sure if that's an issue or not and the same thing probably happens for images that use deprecated function names too. Could be nice to have a "Delete all resampled" task in the dev task menu that destroys all `_resampled` folders and their contents but that's probably outside the scope of this PR and think that would cause issues for images referenced in HTMLEditorField as I think they're only generated when the field is saved, not when the content is viewed. But I'm rambling... nice work on this @JorisDebonnet, hope this gets accepted.

jonom:
Hey @JorisDebonnet actually found one problem here but probably not due to this PR - if you try to resize any images that are already in HTMLEditorField instances they keep referring to the previously generated image using the old filename format and a new one is not generated to suit the new size.

This will probably affect all images going from 3.1 to 3.2 even without this PR because of the encoded filenames I think? Looking at the way HTMLEditorField stores local image URLs I'm surprised to see it's not stored as a short code that is translated later (like internal page links), but instead is a fully evaluated URL to the resampled image. This is kind of a hack and means we're relying on a very specific output format from an API (abusing it). Probably would be a lot cleaner and less prone to breakage if images were stored as short codes and generated when the content is parsed for a template? I'd say this is the root cause of the old bug where flushing generated images for an `Image` object causes that image to become a broken link on all pages where the image is referenced in HTMLText content, requiring you to save the page again to regenerate the images.

@silverstripe/core-team any thoughts on this? I think there may be a need here in 3.2 for images in HTMLEditorField to be made to use short codes and a dev script provided to convert all existing image tags to the short code format which should safeguard against future filename output changes. The alternative would be to revert all the changes to `Image` that have gone in I guess, but I don't think that should be required because the current way of storing image tags isn't really API-safe and could be considered broken...

JorisDebonnet:
Thanks Jono! :)

And I agree about the problematic way Images are saved in the `HTMLEditorField`s. I encountered a similar issue months ago already in 3.1, due to the resized images on your pages that stop working when you clear out the `_resampled` directory. Intuitively, you'd expect those images to be automatically generated again, so indeed, a shortcode would be the best way to go. It only needs the resize and scale methods (right?), so shouldn't be *that* complicated. 

Even if we do add a shortcode system in the future though, people will still have issues with existing pages because those still contain the old resized images. To facilitate an easier transition period, we'd need to add some extra code to convert "old resized images" into shortcodes. We could still be able to finish that before the release candidate. But let's create a separate issue for that!

JorisDebonnet:
Actually, I'm looking into "recognizing old format images" now. Let's see if I can figure that out and add it to this PR.

JorisDebonnet:
Well -- I fixed it. It was in fact something introduced by my PR, and it's now fixed! Meaning that "old format images" are now recognized, and will be saved into the new format when the page is saved again.

jonom:
Nice! Well that takes care of that part then I guess :) I've opened a separate issue for the shortcode idea: https://github.com/silverstripe/silverstripe-framework/issues/4337

jonom:
Hey @JorisDebonnet just wondering if you've accounted for empty directory cleanup here? Might be a good idea to check for empty folders during Image::deleteFormattedImages(). For each image being deleted I guess you'd check if the containing folder is empty and delete it if so, then keep traversing up doing the same thing until you find a non-empty folder or get to the _resampled folder.

JorisDebonnet:
I actually gave some thought to that, but then considered that you usually only need to delete a format-folder when you change a template. Because even if a format-folder becomes empty, a new image that ends up in the same template will undergo the same transformation, and will use that same folder again. 

Of course images may also be formatted without templates (or other hard-coded sections). I guess that only happens when an image in the CMS is resized. So for that particular situation, it is indeed useful to add a check to `Image::deleteFormattedImages`. I personally believe it will only very very rarely trigger a folder removal, because it only happens if you actually *delete* the image (not if you just change the image's size on the page) -- but I suppose that's beside the point. 

So I've added that, including a new `Filesystem::removeFolderIfEmpty` method (it felt wrong to put all that code in `Image.php` and not share it there). One remaining point perhaps is that `Image::deleteFormattedImages()` does not delete formatted images in the "old" format (SetWidth400)... but of course that's the case before this PR, too. I could add code to search for those as well but that's probably going too far.

jonom:
:+1: Thanks for that, I think that's a nice addition

JorisDebonnet:
Forgive me for 'bumping' -- I'm just curious whether this might get a milestone tag :)

Any remaining concerns about these changes? I realise it may feel like an uncomfortably big change, but @jonom and I tested it extensively (aside of the unit tests, of course), and I'll be happy to double-check any remaining compatibility concerns!

filesystem/Filesystem.php
```diff
@@ -79,6 +79,29 @@ public static function removeFolder($folder, $contentsOnly = false) {
 	}
 
 	/**
+	 * Remove a directory, but only if it is empty.
+	 *
+	 * @param String $folder Absolute folder path
+	 * @param Boolean $recursive Remove contained empty folders before attempting to remove this one
+	 * @return Boolean True on success, false on failure. 
+	 */
+	public static function removeFolderIfEmpty($folder, $recursive = true) {
```
dhensby: static functions need to be seperated by `_`: `remove_folder_if_empty`

dhensby:
Hi guys, thanks so much for this work - unfortunately this is going to need to go into 3-branch.

We need to lock down the 3.2 API. Now that 3.2.0-beta.1 has been released, we shouldn't be adding any new APIs (releasing a beta is us making an implied promise the API is stable enough that devs can rely on it and we're asking for bug reports from now on). Adding this kind of change (new API) to beta.2 would just mandate a beta.3 so that we could be sure this feature got enough time in the wild.

Ultimately, we can point at semver and say it's telling us we can't add this to 3.2.

Please can you open a new PR against 3?

filesystem/Filesystem.php
JorisDebonnet: That confuses me a little: above and below this new method, there are existing static functions called `removeFolder`, `folderModTime`, etc?

JorisDebonnet:
Too bad this was *just* too late for beta 1, then...

So yes, I can make it against 3 instead. 

However, what about the included `base64url_encode` change in this PR, which pretty much is a bug fix? It may be rare, but `base64_encode` (whose result can now appear in resampled image filenames) can include the characters / and +. Especially the former tends to be problematic in filenames. :) 

I could fix this in a separate PR that does just that. Is the end of `Image.php` an okay place to put those helper functions?

dhensby:
@JorisDebonnet yes, sorry :(

Please can you patch this base64 issue against whichever is the lowest version it is in (3.1 or 3.2) then the rest against 3.

thanks a bunch, this is some great work and would have loved it in 3.2, but we should try to stick to our versioning system now

filesystem/Filesystem.php
dhensby: Yep... That's SS's inconsistent naming right there. as part of 3 you could rename and deprecate the old functions so they are inline with our conventions?

sminnee:
If it's any consolation, we won't need to wait nearly as long for 3.3, and if we keep the changes in the 3 branch safe, the beta/rc cycle can me much shorter.

JorisDebonnet:
I'll redo this against the 3 branch as soon as that branch has been brought up to date with 3.2 (since I need to build on top of the `base64url` changes from there).

dhensby:
@JorisDebonnet done

JorisDebonnet:
Redone against 3 in #4507.

PR #4331, opened 2015-06-22
Add extension hook for getManyManyComponents()

jason-zz:


dhensby:
What's the usecase for this?

jason-zz:
@dhensby We're working on a module that allow many_many relationships support both draft and live modes. The module will just simply duplicate relationship table and add suffix ' _Live '. Extension hook is required for switching relationship table. If some many_many relationship requires this extension and it's in Live mode, the module calls getManyManyComponents() and return a new ManyManyList which $table has ' _Live ' suffix.

e.g.
Draft : Page_Slides
Live : Page_Slides_Live

model/DataObject.php
```diff
@@ -1771,6 +1771,9 @@ public function getManyManyComponents($componentName, $filter = null, $sort = nu
 			$componentClass, $table, $componentField, $parentField,
 			$this->manyManyExtraFieldsForComponent($componentName)
 		);
+		
+		$this->extend('updateManyManyComponents', $result);
```
willmorgan: This is a good idea, but I would like to see it after the `setDataModel` call, ideally.

jason-zz:
@willmorgan Thanks for your feedback. The commit has been updated.

sb-relaxt-at:
We really appreciate this new hook, as we are working on a very similar module (support for stages and versions in many-many relations). Unfortunately there is too few information passed with this hook (at least for us ;)). I would argue to include $componentName in order to allow lookup of additional information (e.g. configurations for the relation). 

P.S.: I am not sure if is appropriate to add this comment to an already closed pull request - sorry if not :)

kinglozzer:
@sb-relaxt-at You could open a new pull request (against the 3 branch) to add that extra information that we could then review.

I’ve found myself needing the relationship name in an extension before, this is how I did it (I’m sure this can be improved, it was kinda rushed). [This method](https://gist.github.com/kinglozzer/56736aa1de81171dc79a#file-memberextension-php-L9-L32) builds an array of `join table => relation name`, and you can fetch the join table with `$this->getJoinTable()`, so you should be able to use the combination of the two to find the name of the relationship.

dhensby:
Does the `ManyMantList` not contain this information?

kinglozzer:
Nope, just join tables, foreign keys and extrafields

dhensby:
I guess we need to add another argument, then :P

PR #4333, opened 2015-06-22
Allow shortcodes inside script tags. Fixes #4332.

### LINKED ISSUE #4332: Shortcode parser barfs on simple input
I haven't bisected (running latest stable 3.1), but consider:
```
<script>[2]</script>
```

This gets parsed into
```
<script><img class="--ss-shortcode-marker" data-tagid="0" /></script>
```

Doesn't happen with a single `[2]` without the script tags.

Unfortunately I don't know enough about how it parses yet to debug it.

sminnee:
This change will ensure that short tagnames have to start with a letter or _, rather than starting with a number. This stops '[2]' from being a legal shortcode.

Although this is one way of fixing the bug identified in 4332, it would be good to confirm whether it is the *best* fix.

### DIFF
--- a/parsers/ShortcodeParser.php
+++ b/parsers/ShortcodeParser.php
@@ -320,7 +320,18 @@ class ShortcodeParser extends Object {
 				}
 			}
 		}
-		
+
+		// Step 3: remove any tags that don't have handlers registered
+		// Only do this if self::$error_behavior == self::LEAVE
+		// This is optional but speeds things up.
+		if(self::$error_behavior == self::LEAVE) {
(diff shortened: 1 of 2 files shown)


sminnee:
Upon reflection, it might be better to take a different approach with this, and instead only transform shortcodes if thet have a shortcode parser actually registered.

Has this been discussed before?

sminnee:
OK, I take back my previous comments. It *has* been discussed before, and implemented. The actual bug here was to do with <script> tags. Fix pushed and, since it's a low-risk bugfix, should be okay for 3.1.

PR #4335, opened 2015-06-22
Different status dots for draft and modified

jonom:
Added a colour difference for status dots so status of draft vs modified can be distinguished at a glance. Also made the dots a bit stronger.

screen shot 2015-06-22 at 11 47 50 am

### DIFF
--- a/admin/css/screen.css
+++ b/admin/css/screen.css
@@ -858,6 +858,8 @@ form.import-form label.left { width: 250px; }
 
 .jstree-default a .jstree-icon, .jstree-default-rtl a .jstree-icon, .jstree-classic a .jstree-icon, .jstree-apple a .jstree-icon { background-position: -60px -19px; }
 
+.jstree-apple a { border-radius: 3px; }
+
 /* ensure status is visible in sidebar */
 .cms-content-tools .cms-tree.jstree li { min-width: 159px; }
 .cms-content-tools .cms-tree.jstree a { overflow: hidden; display: block; position: relative; }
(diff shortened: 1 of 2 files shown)


chillu:
The red dot looks like something is wrong with the page: To a new user it might indicate that there's errors on it (e.g. broken links), or maybe that its been deleted despite the "draft" label. Maybe a more orange dot (filled and not filled) makes sense? @stevie-mayhew @flashbackzoo @clarkepaul

jonom:
It's actually the same colour as the 'orange' text :) but I agree it looks red when it's solid so could use some toning back

jonom:
Have updated this PR to be more orange and use a grey ring for 'removed from draft' as per @clarkepaul's suggestion. I also made the blue link for 'normal' links much lighter to be more consistent with the other background labels - the existing blue was actually unmodified from the default 'Apple' theme of the plugin. Also dialled back the corner radius to be consistent with other CMS UI elements (this also was using settings from the Apple theme).

This seems like a good a time as any to mention it again - can we please change the output for all SCSS files in the framework to a format that will be consistent for all contributors? At the moment it seems like everyone who submits a CSS change ends up changing 4000 lines of code because their version of Compass/SASS produces different output to the previous developer's. In this PR I'm getting different (I'd say correct) white space output compared to whoever compiled it previously.

I think `output_style: compact` should make things pretty consistent ([SASS reference](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#output_style)), and with the addition of `sourcemap: true` it will be a lot easier for contributors to trace styles in the CMS to the actual SCSS files they need to edit. Otherwise `output_style: compressed` would definitely fix this issue as the compiled css would all be on a single line.

dhensby:
This needs debasing

In terms of scss, I think there needs to be a contribution guide that outlines this. I don't think we should default out put to single line as it then becomes difficult to review if changes had intended effect

jonom:
Will rebase shortly. For CSS, @flashbackzoo has set aside some time to take a look at consistency. I don't think putting the compiled output on a single line would be a bad thing as long as source maps are generated, because knowing which line of a CSS file a style rule appears on isn't very helpful when you actually need to edit the source SCSS files. This is exactly what source maps do, so we should really generate them even if we keep the expanded output we currently have.

jonom:
Rebased

dhensby:
Oh, I agree sourcemaps are invaluable, we must use them. But that doesn't mean we should have compressed CSS.

As a reviewer you need to be fairly certain that the person putting the PR in actually did re-compile the CSS and that's a lot easier with default output from the compiler

jonom:
Good point. I'd prefer not to have it all on one line anyway, as long as we can get the compiled output consistent across different environments. Using this PR as an example you can't actually review the changes in the CSS file effectively right now (diff suppressed: 4,493 additions, 1,195 deletions not shown). So if we're unable to make the output consistent the next best thing would be putting it all on one line to simplify the diff.

I know it's a silly thing to think about but I also don't want to be recognised in stats for contributing thousands of lines to this project if I actually just change one line in a SCSS file. I like stats, I want them to be accurate :)

dhensby:
Yes, it looks like you have the default output style for SASS, but it should be with all rules on one line (to not break the diffs).

jonom:
Oh man... that's embarrassing! Hang on, I'm just going to go and strangle CodeKit and then I'll force push again

jonom:
Thanks for the catch @dhensby, I didn't realise CodeKit only uses some of the settings from `config.rb`. Whoops! Guess there is no cross-environment white-space problem, just a @jonom problem.

dhensby:
Hehe, no worries.. 

Though, a simple way round this is to just not commit compiled css and let the release system do it

jonom:
Ah, I didn't realise that was an option. Good to know.

dhensby:
Well, that's not possible at the moment. I'm thinking theoretical here :P

chillu:
We commit compiled files because we don't want every dev checking out SS via composer to also require `compass`, or whatever more complex NodeJS tooling we choose to depend on next. But I agree that in the age of gzip compression, its unnecessary to keep CSS on a single line. JS is a bit more complicated (minified diffs would be hard to review even on multiple lines).

dhensby:
THANKS

PR #4339, opened 2015-06-23
DOCS: typos, rewording, updated links.

SpiritLevel:


willmorgan:
How very eloquent!

SpiritLevel:
Thanks @willmorgan. I aim to please ;) Though I suppose what is more important is whether or not the end result is a bit more readable!

dhensby:
On line 22 you updated the version to 3.2 but reference master as being 3.2 also, when it's actually 4.0

SpiritLevel:
OK...will fix :)

willmorgan:
:blush:

PR #4341, opened 2015-06-23
Fix new page SiteTree nesting

flashbackzoo:
Check for undefined as 0 is a valid ParentID


tractorcow:
Seems to fix the issue.

Prior to this, creating any new page would (for some reason) create the tree node outside of the root `<li>` in the tree view. I'm not sure why 3.1 doesn't have this issue.

PR #4343, opened 2015-06-24
Fix padding bug pushing out the 'post options' panel

scott1702:
Before 
image
After 
image


tractorcow:
Can you give a bit more of an explanation? I just want to make sure it's clear that this is a CMS css issue and not one that's in blog.

jonom:
There shouldn't be any need for all those image changes either eh?

flashbackzoo:
@tractorcow It is a CMS issue. I updated the padding when I moved the filters up into the header, which caused this bug. It's only surfaced when a user (or module) has a panel.

@jonom The images haven't changed, it's just the hash that compass adds to the sprite's filename. This happens from time to time. I'm not sure the exact reason, but it's probably to something like differences in compass versions, or the way different systems handle newlines. Something's touching the files on compile, compass thinks the files have changed, so updates the sprite hash.

tractorcow:
Travis timed out again so restarting the build that failed.

jonom:
Thanks @flashbackzoo yeah I realise the images haven't actually changed, which is why it would be nice if there weren't any image changes in this PR - but I know it's not @scott1702's fault. Still it would be cool if we could work out why this happens and try to prevent it. Sort of related is some [comments I made yesterday](https://github.com/silverstripe/silverstripe-framework/pull/4335#issuecomment-114340110) about white space differences and source maps - would be good if we can smooth out SCSS compiling across different environments so contributions are a bit more transparent.

flashbackzoo:
^ I agree - it's a bit scary when you compile and have a bunch of images change like that. I'll do some digging...

flashbackzoo:
@jonom I had a quick look into the SCSS issue yesterday and think we can fix it. I've blocked out some time for Wednesday, the week after next, to try out a fix. I'll tag you in the PR - you might be able to help out with testing it :)

jonom:
Sweet! Yeah tag me, will be happy to test.

jonom:
@flashbackzoo just a heads up - I was being dumb and letting CodeKit compile SCSS with the wrong output style, so I was misinformed about inconsistencies in white-space. But if you can do something about the images that will be cool!

flashbackzoo:
Cool thanks

PR #4345, opened 2015-06-24
DOCS: Link to member custom code change is broken

phillprice:
I've found it as http://docs.silverstripe.org/en/3.2/developer_guides/extending/how_tos/track_member_logins/ but I guess the absolute URL is wrong, and I'm not sure if there are language / version url rewrites in place to use them.


dhensby:
@phillprice you just put this bit in: `/developer_guides/extending/how_tos/track_member_logins/`

phillprice:
Fixed up the commits

willmorgan:
The URL works for me; taking this in.

PR #4347, opened 2015-06-24
Fix background size for spinner @2x

kinglozzer:


dhensby:
What's the issue here? can we see before/after screens?

kinglozzer:
The background size is incorrect - the spinner in this context is squashed vertically. It’s hard to get a screenshot as it only displays briefly. This was just overlooked when [other](https://github.com/kinglozzer/sapphire/blob/pulls/spinner-fix/admin/scss/_retina.scss#L61) [values](https://github.com/kinglozzer/sapphire/blob/pulls/spinner-fix/admin/scss/_retina.scss#L68) were updated.

dhensby:
This is what current beta looks like for me on my "retina" screen:

retina

dhensby:
Ah, it's in a dialog:

kinglozzer:
Yeah sorry I should’ve mentioned that!

Issue #4342, opened 2015-06-23
Bulk action checkboxes are easy to mis-click


jonom:
It's easy to click just outside the bulk action checkboxes in the site tree and that will click the underlying link, loading the page in the CMS and clearing your bulk selections. Pretty annoying if you've just gone through and checked a bunch of pages.

This is an easy thing to fix, I just don't have the time right at this moment. The checkbox is handled by a background image so we can just tweak that image and make the clickable area larger so it goes to the edge of the parent container.

Note that the checkbox graphics are the default ones from an ageing copy of JSTree so if anyone takes on fixing this it could be a good time to introduce some sexier checkboxes. Also the background colour for selected nodes could be a brighter+lighter yellow.

![screen shot 2015-06-23 at 9 30 20 am](https://cloud.githubusercontent.com/assets/1079425/8311371/88915e5c-198a-11e5-9fa6-43cbef94de20.png)

![screen shot 2015-06-23 at 9 31 39 am](https://cloud.githubusercontent.com/assets/1079425/8311389/afd14a22-198a-11e5-9a98-cba93b1af3e4.png)

jonom:
Just realised this is a duplicate of https://github.com/silverstripe/silverstripe-cms/issues/400 but since that issue is 3 years old maybe I'll leave this one open for now?

PR #4352, opened 2015-06-29
Fix: #4342 - larger checkbox regions

torleif:


jonom:
Hey @torleif glad to see someone taking looking at this problem! You would need to make a change to the underlying SCSS file though, then compile that to generate the CSS file. Think you might have to change the border colour for hover and selected state too? Also in 3.2 there are a few extra hover colours so would need to account for that if you submit a separate patch for that version.

An easier option instead of adding a border might be to just increase the size of the checkbox element and add some empty space around the checkbox graphic(s) so that the other sprites don't bleed in to the space. Then you won't have to account for any different hover/selected background colours.

torleif:
@jonom ah, didn't even think that this file came from a scss file. I'll change that now.

I wanted to simply increase the padding on the element, but the graphic looks like it was provided by a third party so I avoided editing it. There's also some CSS3 features that would fix this issue. Another option is having an extra div/span inside the ins element.

dhensby:
Yes, I think there's a better solution than a border.

Can't we wrap the element containing the checkbox and add padding to the wrapping el (but make it clickable)? (hope that makes sense...)

jonom:
Oh I just thought of a really easy solution for this. Instead of putting the background image on the `ins` directly, we can apply it to a `:before` pseudo element. Style the `:before` element to be `display:block` abd the same size as the ins currently, add some padding and a negative margin on the `ins`. Boom.

torleif:
@jonom sadly in that case the hit area is still applied to the parent a area. I also tried adding a parent/containing element using jquery, though there's too many hard coded css & js references in the jquery ui (and it doesn't allow for easy extending)

There's also the CSS 3.0 feature for using an image mask, though that does not have MSIE support.

So I did the next best thing: I moved the checkbox out of the

jonom:
Hey @torleif I gave my idea a try and it seems to work well, think it's a nice simple solution: https://github.com/silverstripe/silverstripe-framework/pull/4366

torleif:
ah brilliant, your solution is much better. I'll close off my pull request

PR #4353, opened 2015-06-29
Update styles for action buttons and tabs

scott1702:
Before: image

After: image

also needed - https://github.com/silverstripe/silverstripe-cms/pull/1233


dhensby:
looking gewd

scott1702:
Don't merge this just yet, still a couple of changes to come in :)

jonom:
Very nice

scott1702:
Updated with new, more consistent padding layout + slightly enhanced colours.
@clarkepaul

chillu:
Looking good, Scott! In order to make review and feedback easier, can you please always provide before/after screenshots on style changes?

scott1702:
@chillu done, will keep that in mind  :)

chillu:
Same here, no keyboard focus on buttons breaks accessibility

feejin:
I much prefer the Blog Posts etc links aligned left.

Not keen on losing buttons and tabs though – it looks good visually, not sold on it from a UX angle.

I do a lot of CMS support with not-that-computer-literate people and without the familiarity of a conventional button / tab it's going to make it harder for them. Especially the "Add New" which is pretty lost without any kind of colour to it.

That said, I'm not a fan of the existing bulky gradients so this feels like a step in the right direction IMHO.

adrexia:
I like the tabs a lot. Its a big improvement, space wise. I'm less keen on buttons looking like regular text. Generally those things are supposed to be calls to action, and I'm afraid they will just be overlooked without definition.

I am a fan of dropping the gradients though.

kinglozzer:
One minor bug - the border here:
screen shot 2015-06-30 at 11 29 16

I’m a tiny bit on the fence about the tab changes. The new style looks great, I just worry that the tabs might be a bit lost: if I tell a client to click the "Categorisation" tab in the example screenshots, I feel like they won’t spot it. Is there any way we could make them stand out a little more?

I don’t like the button changes though: they’re harder to spot/identify, inconsistent with all the other buttons in the CMS and I don’t really think they look much better anyway.

adrexia:
"I don’t like the button changes though: they’re harder to spot/identify, inconsistent with all the other buttons in the CMS and I don’t really think they look much better anyway."

That's a good point. Would changing the colours up to look more like links help, do you think? You could also make the selected 'tab' look more link a tab, and have the others more minimal, but I'm not sure you'd save on space quite as much.

SpiritLevel:
My 2 cents on this very subjective topic! :)

I agree with @feejin about looking cleaner visually but being less obvious for users to know what to do. My immediate reaction to the loss of the tabs was "how do I get to that area now?". Tabs are so familiar to users that they require no thinking whereas the new style would be less hard-wired so feels less natural and requires a millisecond of thought. 

I kind of like the gradient buttons! After all, Github uses them so they must be good!!! Again, I think gradients buttons are so familiar as a place to click to do something that anything else is less natural. As an alternative to gradients, one could have a very subtly raised flat button; anything would be fine as long as it looks like a physical button with a slight 3D effect. If not buttons, then the labels should have verbs...for example, "Show Settings" and "Show Content" and "Edit Content" or similar.

I am neutral about left or right justification of the tabs since either way it is clear how to access those areas.

clarkepaul:
Thanks for the feedback, I think I need to justify some of the changes. We will be increasing the size of "North" bar to accommodate one line of breadcrumbs and another for the page name, if we keep the old style tabs which take the full space they would be huge and stand out far too much. It is the second set of tabs (if they exist) that the user interacts with mostly.

I believe that within the CMS we have too many things competing for the users attention. There are too many places where bold text is being used and too many boxes used to make things standout. The items which matter most like the editing the page, save and publish become prominent.

With the tabs we could look at some other options with color or highlighting the selected item, I've tried changing the tabs before with multiple options and there was no consensus on a style. The solution I've provided will help with a space issue (when there are a lot of tabs) because we don't have the bulk of the old tabs. FYI we still need to solve the wrapping tab issue.

One of the main reasons for removing the border style from the buttons is to allow for all the actions to be consistent between the larger and smaller site tree view and without the bulk of buttons there is a lot more space to allow for additional actions.

IMHO The age of making everything look like buttons has passed – we need the main actions to be more obvious and the less prominent actions to be available. All of your comments are valuable and the team will consider it all before it is released – that's what the Beta is for :).

Anyway this is what we are aiming for at the moment, hopefully it is progressing on what was there previously. 
silverstripe ui_main

jonom:
I really like the new tabs, but I think the secondary tabs need a bit of extra contrast somehow - the North tabs stand out thanks to a different background colour but think the second row are getting a little lost.

While we're backing off of gradients a bit, now seems like a great time to remove all the gradients from the left sidebar. They're starting to look a bit out of place and that whole section already stands out heaps thanks to the darker background so don't think we need to draw more attention to the options with gradients.

SpiritLevel:
One last idea from the peanut gallery: what about using the style of tree view/list view buttons on top right of middle panel for all main actions ? This would bring a really satisfying consistency. In the trio of "Add Page  Actions Options" in the middle panel you could leave the selected one, Actions, as it is but add thin black outlines to Add Page and Options, to match the black outline around the tree view/list view buttons and to indicate they are clickable. You could also use this style for the "Go" the "Edit" buttons and all the tabs selectors too. Heck, event those big old gradient blobs in the left panel could be done this way.  Given the choice between obvious buttons and something obscure that takes time to work out if it is even clickable, I'll take the buttons. Long live the buttons!!! Same with tabs. Long live the tabs !!!

adrexia:
That design looks really nice! I like the selected/hover(?) state on the button. I think that might help convey understanding. The iconography is also a bit larger too, yes? 

I understand that this is an incremental thing, but there's a lot in that 'direction design' that makes the interface make sense in a way that the current implementation doesn't. For one, the breadcrumb positioning helps draw attention to the second level tabs. How long before we can get all of it? ;)

flashbackzoo:
I like the minimal look of the tabs. Similar to what Google is currently doing - which is nice also.

tabs

The buttons could have an outline maybe - to make them more button like. Overall it looks really good!

scott1702:
Rebased against @flashbackzoo's 3 column layout
image

flashbackzoo:
Couple of things:

cms-tabs-01

- Vertical alignment is a bit off
- The height of the top bar varies depending on what section you're in. Navigate between 'Reports' and 'Security' and you'll see what I mean.
- Can we apply the new button styles to the SiteTree only? Or apply them to all top level sections? Have the new styles only in 'Pages' seems weird.

flashbackzoo:
Just noticed the width is incorrect then the panel is collapsed

cms-updates

scott1702:
Updated with numerous bug-fixes + spacing changes. This should now look a lot like @clarkepaul's screenshot. I haven't changed much with the designs of the buttons/tabs but if you haven't already checkout the branch, have a play around and post your feedback on what you think.

admin/composer.json
```diff
@@ -0,0 +1,5 @@
+{
+    "require": {
+        "silverstripe/blog": "dev-master"
```
chillu: Needs to be removed, we're not requiring blog in core. I assume that's a dev artefact.

chillu:
I think the buttons styling still needs work: It's too subdued at the moment. There's a few UI elements like breadcrumbs, the URL segment and the page tree which are obvious links (blue highlight). There's others which are obvious buttons ("save"). Either of them communicates clearly that the UI element is clickable and will invoke an action. Then there's some grey UI elements which don't do anything (current page in breadcrumbs, form field labels, "your site name"), and others which perform an important action (tabs and "add page"). @clarkepaul That seems quite inconsistent?

adrexia:
Perhaps you could copy what google does and use colours to at least highlight the active tab? Maybe using the link colours or some other shade of blue?

chillu:
The UI elements become even more imbalanced when you have unsaved edits and preview mode (big fat green buttons) - we need to look at this holistically. Also, visible hover/focus feedback is great, but ideally we come up with a UI paradigm that also works well on touch devices where you don't have that level of feedback.

image

clarkepaul:
I think we'll revisit the site tree / gridfield actions and tabs based on peoples feedback and see what other options might be an improvement - personally I think its more about what people are used to and not being confident that the change will be meaningful enough. From the limited testing I did with real authors they had no issues whether or not they where actions or tabs - I didn't test on multiple designs though so not the best tests so far. 

@adrexia I've already tried using a blue instead of Googles red for tabs and it just didn't look right. 
@chillu Plenty of interfaces these days use iconography instead of the button look, and while we are trying to find solutions which can be rolled out to the whole interface we cannot do that all at once unfortunately and the way the main actions work is another complexity which makes things harder for sure.
Example:
pasted_image_7_07_15_6_01_pm

chillu:
> Plenty of interfaces these days use iconography instead of the button look

Yep, if done consistently. My point is that the "half redesigned" state at the moment creates weird visual focus points, some action stand out much more than others due to blue text or button styling. And those aren't necessarily the most important actions. But I do acknowledge that we have to start somewhere. Given this is on the 3 branch, and 3.3 won't be released soon, I'm happy to merge this as "work in progress", and look forward to further UI concepts to make this more consistent across the board.

clarkepaul:
@chillu can we get this merged as work in progress? - we realise there is more to do before 3.3 is released.
Outstanding work which we aim to get prioritized:
- Button consistency across entire UI (provide a few alternatives to community for review)
- Interface iconography to be more consistent across UI
- Tab refinement (provide a few alternatives to community for review)

chillu:
Sweet, merged! Great to hear those three areas are on your todo list for 3.3.

jonom:
@clarkepaul I'm still a bit worried about the secondary tabs - agree that blue wouldn't work well because it's the standard link colour in SilverStripe, and I think that's why Google uses red - it looks like an active state rather than a link. Red wouldn't work for us but how about that green/teal colour used on silverstripe.org? I think that could work quite well (forgive the dodgy photoshop colouring):

tabs

In fact it's already used in a similar way on silverstripe.org:

screen shot 2015-07-16 at 9 17 09 am
screen shot 2015-07-16 at 9 17 16 am

clarkepaul:
Thanks for that @jonom, yeah I've added something to our backlog of work to cover those three items because of the feedback we've received to see if we can make everyone more comfortable with any changes. 

Feel free for anyone to provide some options for consideration - when the work comes up I can fine tune the ideas for a better comparison.

stevie-mayhew:
I actually really like the way which @Rhym has done this in his re-theme module https://github.com/Rhym/silverstripe-cms-theme - a subtle indication without the massive change in colour of everything works well.
Subtle

jonom:
That does look good but I don't know if it would work well for a secondary/nested set of tabs

PR #4356, opened 2015-06-29
Define function as static (line 59)

jakxnz:
Not defining function on line 59 as static triggers php error: [Strict Notice] call_user_func() expects parameter 1 to be a valid callback, non-static method Page::MyShortCodeMethod() should not be called statically

Note: PHP 5.5.12

### DIFF
--- a/docs/en/02_Developer_Guides/05_Extending/04_Shortcodes.md
+++ b/docs/en/02_Developer_Guides/05_Extending/04_Shortcodes.md
@@ -56,7 +56,7 @@ First we need to define a callback for the shortcode.
 			'MyShortCodeMethod' => 'HTMLText'
 		);
 
-		public function MyShortCodeMethod($arguments, $content = null, $parser = null, $tagName) {
+		public static function MyShortCodeMethod($arguments, $content = null, $parser = null, $tagName) {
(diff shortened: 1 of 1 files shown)


docs/en/02_Developer_Guides/05_Extending/04_Shortcodes.md
```diff
@@ -56,7 +56,7 @@ First we need to define a callback for the shortcode.
 			'MyShortCodeMethod' => 'HTMLText'
 		);
 
-		public function MyShortCodeMethod($arguments, $content = null, $parser = null, $tagName) {
+		public static function MyShortCodeMethod($arguments, $content = null, $parser = null, $tagName) {
```
dhensby: Good spot, if it's a static method it should follow our naming conventions and be `my _shortcde_method`. Would you mind updating the docs to use that name?

jakxnz:
@chillu would you like me to implement @dhensby's suggestion?

dhensby:
@jakxnz it comes down to whether you want to or not :P

jakxnz:
Thx @dhensby! I asked irc/silverstripe and the consensus was that camel cased shortcode methods was somewhat of a convention unto it's own.

dhensby:
Hehe, I'm not sure we should have conventions that are anti convention to the current convention, but ok. Not a big deal for the moment

Issue #4359, opened 2015-07-01
3.2 requires a PHP version upgrade


kinglozzer:
From 5.3.2 > 5.3.3. I wouldn’t call this a blocker (if you’re running 5.3.2, or in fact any version of 5.3, please, please upgrade!), but we should practice what we preach with semver. Especially as it [looks like](https://github.com/silverstripe/silverstripe-framework/pull/2577) it was only bumped up so we could include the `JSON_NUMERIC_CHECK` flag.

I’m sure we can develop a workaround for using the flag, but it’d be good to check that everything still works on 5.3.2. Travis will run builds against 5.3.3 (composer does, for example), but I’m not sure about 5.3.2.

PR #4360, opened 2015-07-01
Remove PHP version upgrade requirement for 3.2

kinglozzer:
#4359

I’ve added an assertion to `ImageTest` that covers the bug that #2577 was trying to fix... let’s see what Travis makes of adding 5.3.2 to the build matrix

### DIFF
--- a/model/Image.php
+++ b/model/Image.php
@@ -711,7 +711,7 @@ class Image extends File implements Flushable {
 		array_shift($args);
 		$folder = $this->ParentID ? $this->Parent()->Filename : ASSETS_DIR . "/";
 
-		$format = $format . base64_encode(json_encode($args, JSON_NUMERIC_CHECK));
+		$format = $format . base64_encode(json_encode(array_map('strval', $args)));
 		$filename = $format . "-" . $this->Name;
 		$patterns = $this->getFilenamePatterns($this->Name);
 		if (!preg_match($patterns['FullPattern'], $filename)) {
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
 		}
 	],
 	"require": {
-		"php": ">=5.3.3",
+		"php": ">=5.3.2",
 		"composer/installers": "~1.0"
 	},
 	"require-dev": {
--- a/dev/install/install.php
+++ b/dev/install/install.php
@@ -4,17 +4,17 @@
  ************************************************************************************
  **                                                                                **
  **  If you can read this text in your browser then you don't have PHP installed.  **
- **  Please install PHP 5.3.3 or higher, preferably PHP 5.3.4+.                    **
+ **  Please install PHP 5.3.2 or higher, preferably PHP 5.3.4+.                    **
  **                                                                                **
  ************************************************************************************
  ************************************************************************************/
 
 /**
- * PHP version check. Make sure we've got at least PHP 5.3.3 in the most friendly way possible
+ * PHP version check. Make sure we've got at least PHP 5.3.2 in the most friendly way possible
  */
 define('FRAMEWORK_NAME', 'framework');
 
-if (version_compare(phpversion(), '5.3.3', '<')) {
+if (version_compare(phpversion(), '5.3.2', '<')) {
 	header("HTTP/1.1 500 Server Error");
 	echo str_replace(
 		array('$PHPVersion', 'sapphire'),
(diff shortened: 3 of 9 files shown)


kinglozzer:
Unsurprisingly, Travis doesn’t like 5.3.2. Any volunteers for running tests on that PHP version? If not I’ll try to find time to build it myself and test

kinglozzer:
I found a vagrant box with 5.3.2, everything in 3.2 I tested works as expected.

Tests are passing with the exceptions of some `DirectorTest` assertions, as parse_url() emits `E_WARNING` in 5.3.2 instead of just returning false, and some whitespace issues that are probably caused by something environmental (tidy not installed or something).

Acceptable? As there is no version of PHPUnit out there anymore (now that pear.phpunit.de is dead) that officially supports PHP 5.3.2, test failures are probably acceptable IMO.

We could also add 5.3.3 to travis.yml for the 3.x line just in case, that’s as close as we can get with Travis.

model/Image.php
```diff
@@ -711,7 +711,7 @@ public function cacheFilename($format) {
 		array_shift($args);
 		$folder = $this->ParentID ? $this->Parent()->Filename : ASSETS_DIR . "/";
 
-		$format = $format . base64_encode(json_encode($args, JSON_NUMERIC_CHECK));
+		$format = $format . base64_encode(json_encode(array_map('strval', $args)));
```
chillu: http://php.net/manual/en/json.constants.php says: "Encodes numeric strings as numbers.". So I guess this should be `intval` rather than `strval`?

chillu:
We test PHP 5.3 on Travis for 3.2 (https://github.com/silverstripe/silverstripe-framework/blob/3.2/.travis.yml#L33). I think Travis might actually support 5.3.2, see http://docs.travis-ci.com/user/ci-environment/#PHP-VM-images and https://github.com/php-build/php-build/tree/master/share/php-build/definitions. Wanna give it a go?

model/Image.php
kinglozzer: Yep, technically it should be but the args are sometimes strings (e.g. FFFFFF for PaddedImage) which would presumably always come out as '1' resulting in collisions

kinglozzer:
@chillu I’ve been playing around with it, but haven’t had any luck so far :(. I tried simply adding it to the build matrix last week, but that didn’t work. I just tried using `before_script` and `phpenv` to switch versions, each attempt failed for different reasons:

- `phpenv global 5.3.2` - `rbenv: version '5.3.2' not installed` - https://travis-ci.org/silverstripe/silverstripe-framework/jobs/69857370
- `phpenv install 5.3.2 && phpenv global 5.3.2` - `rbenv: no such command 'install'` - https://travis-ci.org/silverstripe/silverstripe-framework/jobs/69859324

I can only assume the Travis docs mean that their VMs are _pre-built_ with php-build, as it [doesn’t seem to be installed](https://travis-ci.org/silverstripe/silverstripe-framework/jobs/69861701). I tried installing `php-build` so that `phpenv install x.x.x` would work, then installing 5.3.2... [no joy with that either](https://travis-ci.org/silverstripe/silverstripe-framework/jobs/69864975).

kinglozzer:
Well, I give up! I even tried manually building PHP before running the tests (not that something that impossibly slow would ever be merged) just to see if things ran okay, but couldn’t even get that to work :disappointed:.

model/Image.php
jonom: Don't know if this is related but @JorisDebonnet noted a problem with base64_encode vs base64_urlencode and fixed it up in [a PR](https://github.com/silverstripe/silverstripe-framework/pull/4328) that's awaiting review.

kinglozzer: Good spot, from reading the PR/issue it looks like that was added because `base64_encode()` isn’t url-safe. Looks like whoever gets merged last will need to rebase, but that shouldn’t be an issue :)

chillu:
So, the upgrade hardliner in me says: If we can't even get it to build ourselves, can't run tests on it, and even the most ancient Linux distros have at least 5.3.3 ... why go through the trouble?

kinglozzer:
Agreed. I wanted to avoid our first “semver-like” release requiring the most obvious type of BC break, a PHP version upgrade, but this is a bit ridiculous. I doubt there’s anyone out there diligent enough to upgrade their sites to 3.2, but not diligent enough to be running a PHP version that isn’t EOL.

PR #4361, opened 2015-07-01
Amending composer.json constraints

dhensby:
see https://github.com/silverstripe/silverstripe-installer/issues/100

- Moves composer installers to be locked onto v1
- Removed minimum-stability as it is root-only setting (https://getcomposer.org/doc/04-schema.md#minimum-stability)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,13 +17,12 @@
 	],
 	"require": {
 		"php": ">=5.3.3",
-		"composer/installers": "*"
+		"composer/installers": "~1.0"
 	},
 	"require-dev": {
-		"phpunit/PHPUnit": "~3.7@stable"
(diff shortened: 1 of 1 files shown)


sminnee:
I think that we should leave the `@stable` marker in there for those people who have already created their project and haven't yet added `prefer-stable: true` to their `composer.json`.

dhensby:
It's redundant at the moment as dev deps for nested dependencies aren't installed... Probably could remove it completely

tractorcow:
Restarted the build... although still waiting for it to restart. :D

dhensby:
all pass - ping @tractorcow

tractorcow:
Yep, as @dhensby said require-dev is root only. https://getcomposer.org/doc/04-schema.md#require-dev

PR #4363, opened 2015-07-01
Updated RestfulService to include PATCH

robpetreski:
RestfulService now allows PATCH and also removed CURLOPT_POST as it seems redundant if implying POST via the $method

### DIFF
--- a/api/RestfulService.php
+++ b/api/RestfulService.php
@@ -170,7 +170,7 @@ class RestfulService extends ViewableData implements Flushable {
 		$url = $this->getAbsoluteRequestURL($subURL); 
 		$method = strtoupper($method);
 		
-		assert(in_array($method, array('GET','POST','PUT','DELETE','HEAD','OPTIONS')));
+		assert(in_array($method, array('GET','POST','PUT','DELETE','HEAD','OPTIONS','PATCH')));
 		
 		$cache_path = $this->getCachePath(array(
 			$url,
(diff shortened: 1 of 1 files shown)


kinglozzer:
Thanks @daOutlawz. As this is a new feature, it will need to go against the `3` branch instead of `3.1`.

Is `CURLOPT_POST` definitely redundant? What happens if you’re sending a POST request, but `$data` is null? I don’t see any harm in explicitly setting it like we currently are

daOutlawz:
@kinglozzer, thanks for that.

Sending null data without explicitly setting <code>CURLOPT_POST</code> to <code>true</code> shouldn't be an issue.  I actually thought there would be more issues setting that explicitly with a PATCH as <code>CURLOPT_POST</code> assumes your <code>Content-Type</code> is <code>application/x-www-form-urlencoded</code>.

daOutlawz:
@kinglozzer - is there any movement on this?  Or did you imply that you wanted it changed to re-include <code>CURLOPT_POST</code>

robpetreski:
Hi,

If I do this against 3.0 will it be merged into 3.1 also?  As we need support for 3.1 at this stage.

jonom:
@robpetreski new features can only go in to the 3 or master branch sorry. Good news is it shouldn't be hard to upgrade a site from 3.1 to 3.2 or 3.3 when it comes out... there shouldn't be a compelling reason to leave a site on 3.1 now that 3.2 is out.

tractorcow:
If you want to use a custom implementation in 3.1 you'll need to use injector to set your custom class.

E.g.

```yaml
Injector:
  RestfulService: CustomRestfulService
```

And implement your overridden method in your class. :)

Caveat: I'm not 100% sure that it will let you override it in this way.

In any case, this can't be merged into 3.1 so I'm closing this PR for now. If you'd like to merge this into the 3 branch, I suggest you create it soon because we're soon to release 3.3.0 beta, after which new features won't be available until 3.4.

PR #4365, opened 2015-07-03
Use container-based travis

tractorcow:
See https://github.com/silverstripe-labs/silverstripe-travis-support/issues/15#issuecomment-117972388

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
 language: php
 
+sudo: false
+
+cache:
+  directories:
+    - $HOME/.composer/cache
+
 php:
   - 5.4
 


tractorcow:
Wow it is a lot faster; I created this 4 minutes ago and one of the builds has already finished.

tractorcow:
I think we should look at adding `~/.composer/cache` to the cache. See http://docs.travis-ci.com/user/caching/

that should speed up repeated builds more, right?

tractorcow:
Ok, so apparently caches aren't available except on travis premium. =(

tractorcow:
Removed unsupported features; This is ready to merge.

Requesting feedback from @silverstripe/core-team

tractorcow:
Public note to self: Investigate circleci.

It has a much cheaper "enterprise" package, and also supports parallel execution of tests. https://circleci.com/

dhensby:
@tractorcow what I can tell of caches is that they are either premium only feature OR they are available in container based runs http://docs.travis-ci.com/user/workers/container-based-infrastructure/

> allow the use of caches for public repositories

tractorcow:
Hm, maybe the tests I ran needed a few runs to pick up caching? I'll re-enable tomorrow and see if we can get it to work.

dhensby:
Possibly, though what are you expecting to cache? It seems that composer deps don't really need to be cached as Travis employs their own caching on these kinds of external items. I wouldn't say the composer install is particularly slow aspect of the builds. It takes ~17 seconds to do the prep

It seems that only 30 seconds of the build is spent on things that aren't running the test suite.

Travis docs on caching do say that if all you're doing is downloading code, it probably won't be any faster using the cache (as this still needs to be downloaded onto the box).

tractorcow:
Actually the docs say

>  so everything related to tools like Bundler, pip, Composer, npm, Gradle, Maven, is what should go into the cache.

So I'll restore it. :D

tractorcow:
Ok, I think this is ready to merge if the build passes. :D

dhensby:
Where does that `$COMPOSER_CACHE_DIR` come from?

Also, it does say that, but full quote:

> The cache’s purpose is to make installing language-specific dependencies easy and fast, so everything related to tools like Bundler, pip, Composer, npm, Gradle, Maven, is what should go into the cache.
> 
> For other things, the cache won’t be an improvement. **Installing them usually takes only short amounts of time, but downloading them will be the same speed when pulled from the cache as it will be from their original source**. You possibly won’t see a speedup putting them into the cache.
> 
> Things like:
> 
> - Android SDKs
> - Debian packages
> - JDK packages
> - Compiled binaries
> 
> Anything that’s commonly not changing is better suited for something like our APT caching proxy. Please shoot us an email and we’ll see about adding your custom source to our cache.

As composer downloads take such a minuscule amount of time, it won't be significantly faster to download them from travis' cache server either...

If we see a speed improvement, then sure, I guess, but it's not going to change our build times :P

dhensby:
Appears the cache isn't working, but no indication as to why:

tractorcow:
I got it from http://frankneff.ch/quality-assurance/2014/04/19/use-travis-build-matrix-to-test-against-multiple-symfony-versions/ actually :D

I thought that full quote applied to

> For other things

Reading comprehension is really not my thing, so, I'm really just hitting two sticks together until I see fire. :D

tractorcow:
Maybe it'll work if we explicitly set the composer_cache_dir....

tractorcow:
It looks like in https://github.com/yiisoft/yii2/blob/f9bb1e21cb8ef351fa5f9af55d7ce380457c3b40/.travis.yml#L34 that they explicitly set the cache dir to `~/.composer/cache`.

tractorcow:
Let's try this. :D

tractorcow:
```
could not download cache
travis_time:end:1d02f37e:start=1436245836843196773,finish=1436245838975822335,duration=2132625562
[0Ktravis_time:start:0a4f99ee
[0Kadding /home/travis/.composer/cache to cache
```

I'll try to restart and see if the second run picks up the cache

tractorcow:
Second run is great success

```
Setting up build cache
$ export CASHER_DIR=$HOME/.casher
$ Installing caching utilities
attempting to download cache archive
fetching PR.4365/cache--php-5.4.tgz
found cache
adding /home/travis/.composer/cache to cache
```

then


```
Installing dependencies (including require-dev)
  - Installing composer/installers (dev-master e420b53)
    Loading from cache

  - Installing silverstripe/framework (3.1.x-dev)
    Loading from cache

  - Installing silverstripe/sqlite3 (1.3.x-dev 6bb7d5a)
    Loading from cache

  - Installing silverstripe-themes/simple (dev-master 5bf8166)
    Loading from cache

  - Installing silverstripe/postgresql (1.1.x-dev 6068792)
    Loading from cache

  - Installing symfony/yaml (2.8.x-dev 000e7fc)
    Loading from cache

  - Installing phpunit/php-text-template (1.2.1)
    Loading from cache

```


#mergeme

dhensby:
nice, and the time saving is....

tractorcow:
Just to clarify, the main reason for this approach isn't so much the speed boost (which is tiny, but present), but more the fact that we are finding several builds timing out due to connectivity with github. A cache should help mitigate these issues.

markguinn:
Brilliant! Great work guys!

dhensby:
Ok ok... let's do this thing

dhensby:
Now let's go and apply this to all our travis modules :)

PR #4366, opened 2015-07-03
Expanded bulk action checkbox hit zone

### LINKED ISSUE #4342: Bulk action checkboxes are easy to mis-click

jonom:
Increased checkbox click area so it's harder to accidentally click underlying link (fixes #4342, fixes https://github.com/silverstripe/silverstripe-cms/issues/400, closes #4352)


jonom:
Have tested this in latest Chrome + IE8, no issues in either.

chillu:
Good stuff! But these kind of improvements should go into 3.2 now, 3.1 is bugfix only

jonom:
It's sort of bug-like :) but okay will re-target for 3.2

jonom:
3.2 version at https://github.com/silverstripe/silverstripe-framework/pull/4367

PR #4369, opened 2015-07-05
FIX memory exhaustion in MySQLStatement->bind()

### LINKED ISSUE #4322: Install of 3.2.0-beta1 requires memory increase in php.ini
When installing 3.2.0-beta1, composer brings the files down fine but going to localhost/ss32 hits the default memory limit of 128M in php.ini. I raised it to 256MB and the installation completes properly now.

@bueckl pointed this out to me in issue #4319 and @dhensby suggested I open this as a new issue and cc @tractorcow

I did not encounter this problem installing 3.1.13.

Cheers,
D.

PS: if this is new to 3.2 perhaps it should be mentioned in the install docs ?

micmania1:
This is an attempt to fix #4322. It definitely fixed my problem but I'm not sure what side-effects there may be.

I basically done what was advised here: http://php.net/manual/en/mysqli-stmt.bind-result.php#57564

cc: @tractorcow

### DIFF
--- a/model/connect/MySQLStatement.php
+++ b/model/connect/MySQLStatement.php
@@ -58,12 +58,13 @@ class MySQLStatement extends SS_Query {
 			$variables[] = &$this->boundValues[$field->name];
 		}
 
-		call_user_func_array(array($this->statement, 'bind_result'), $variables);
 		$this->bound = true;
 		$this->metadata->free();
 
 		// Buffer all results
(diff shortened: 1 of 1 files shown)


dhensby:
I tested this on my machine that was running out of memory and it's fixed the memory issue :)

tractorcow:
Wow nice find, @micmania1.

Issue #4336, opened 2015-06-22
dev/build in CLI doesn't always exit with proper exitcode


stojg:
When running dev/build from the command line and there is an error, the PHP Process doesn't end with a proper non zero exit code. (this might be true for all CLI) and errors are printed to STDOUT instead of STDERR.

I've noticed that non reachable database exits with a non zero exitcode.

This makes it harder to automate deployments and rollbacks.

hafriedlander:
So non-reachable database does correctly exit with a non-zero exit code? Do you have an example where it incorrectly exits with a zero exit code?

stojg:
How to reproduce:

```php
class Page extends SiteTree {

	public static $db = [
		'WillFail' => 'NotAtype'
	];
}
```

`./framework/sake dev/build; echo $? `

or

` php ./framework/cli-script.php dev/build; echo $?`

Expected exit code is non-zero. Even if for example [Exit Codes With Special Meanings](http://www.tldp.org/LDP/abs/html/exitcodes.html#EXITCODESREF) tells us not to use `1` or `2`, I'm not shy of breaking the rules and using `1` for warning and `2` as error as per the [Nagios plugin return codes](http://nagios.sourceforge.net/docs/3_0/pluginapi.html) rules just because a lot of monitoring tools are using this as a standard.

But anything else than `0` is better.

The same goes for STDERR

```
$ php ./framework/cli-script.php dev/build  2>&1 >/dev/null
```

Unless I have the logger setup as per:
```
$sysLogWriter = new SS_SysLogWriter();
SS_Log::add_writer($sysLogWriter, Zend_Log::DEBUG, '<=');
```

Then it will clearly print a syslog message to my terminal session:

```
Jun 23 21:39:53 stone SilverStripe_log[56823] <Error>: [23-Jun-2015 21:39:53] Error at framework/control/injector/InjectionCreator.php line 14: Uncaught ReflectionException: Class NotAtype does not exist (http://sandbox.dev/dev/build)
```

sminnee:
So, this patch will fix it: https://gist.github.com/6ae8007a12b05a86b17d

I've manually tested the following scenarios:

 * parse error (exit code 255)
 * trigger_error `E_USER_ERROR` (exit code 1)
 * unhandled exception (exit code 1)
 * trigger_error `E_USER_WARNING` (continue execution, exit code 0)

I've checked dev & live environment types, and also setting display_errors = off.

[...]

PR #4371, opened 2015-07-06
FIX undefined method should be fatal error

micmania1:
There's currently no solid way to detect that a build has failed on CLI. At the moment if you call $class->undefinedMethod() it will throw an exception and end execution but it returns with exit code 0 meaning the script has ended successfully.

This fix passes handling through SilverStripe and back to PHP to throw the default exit code (255).

Ideally, this would all be handled within SilverStripe but that requires an API change so something for the future.

To replicate the issue:

On any DataObject subclass add:

    public function requireDefaultRecords() {
        parent::requireDefaultRecords();
        $this->undefinedMethod();
    }

Then on the CLI run:
    `php framework/cli-script.php dev/build > /dev/null && echo $?`

Expected: Fatal Error
Actual: 0

### DIFF
--- a/core/Object.php
+++ b/core/Object.php
@@ -758,7 +758,8 @@ abstract class Object {
 					);
 			}
 		} else {
-			// Please do not change the exception code number below.
+			// Please do not change the exception code number below as it effects VirtualPage in the cms.
+			// If anybody knows what 2175 means please update this comment. Thanks.
 			$class = get_class($this);
 			throw new Exception("Object->__call(): the method '$method' does not exist on '$class'", 2175);
 		}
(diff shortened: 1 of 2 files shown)


micmania1:
I've updated the PR based on this discussion: https://github.com/silverstripe/silverstripe-framework/issues/4336

I've left the user_error() instead of Exception because it doesn't really make sense to throw an Exception for undefined methods.

tractorcow:
There needs to be substitute fix for `VirtualPage`; See https://github.com/silverstripe/silverstripe-framework/commit/1eb80b7498e91362f34212eb8ca2b0c51c14f628

Issue #3798, opened 2015-01-17
Should 'i18nTextCollector->collectFromTemplate()' really collect from included files?


Zauberfisch:
[i18nTextCollector->collectFromTemplate()](https://github.com/silverstripe/silverstripe-framework/blob/eb069e605d8cbb029d9c0270a9081060f6ded229/i18n/i18nTextCollector.php#L334) currently parses a template file, and also parses all other template files that are included with `<% include ___ %>`.

From what I can tell this is a bad idea for the following reasons:
- if the file is part of the module that is currently parsed, it will be parsed anyway, because `i18nTextCollector->collectFromTemplate()` is called for every .ss file in this module, so this is redundant.
- if the file is part of another module, it will still include the translations in this module. meaning if your project includes a template file from a module, now you will have translation strings from that module in your mysite folder, which overwrites the already existing translations with the default strings
- it is a bad reimplementation (just basic regex) of the template engine, and is bound to be inconsistent with SSViewer, in fact, I believe it already is, I think this regex: `'/<' . '% include +([A-Za-z0-9_]+) +%' . '>/'` will only find basic includes, but not the ones that pass arguments to the include
- and there is this comment in the code: `// @todo Will get massively confused if you include the includer -> infinite loop` which sounds pretty nasty

can we safely remove this part of `collectFromTemplate()`? or is there a good reason why its being kept around?

chillu:
I think we can. This was necessary when we auto-added namespaces to entities based on the template file name. For example, `<% _t('MYENTITY', 'foo') %>` placed in `Includes/SomeInclude.ss` would generate the entity `SomeInclude.ss.MYENTITY`. That's no longer the case, you need to explicitly add namespaces as required, both in *.php and *.ss

Wanna remove it and check that tests still pass?

tractorcow:
I'm addressing this in a fix for textcollector.

tractorcow:
Fixed with #4373

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-framework/pull/4373

Issue #3797, opened 2015-01-17
i18nTextCollector->getFilesRecursive() is unreliable


Zauberfisch:
`i18nTextCollector->getFilesRecursive()` recursively loops a folder to find all files and returns them in an array.

However, the array may not contain all files, because it uses the file name without path or extension as key.

thus, if you have:

    mysite/code/Contact/Form.php
    mysite/code/Login/Form.php

it will only contain the later one because the first occurrence of Form is [overwritten](https://github.com/silverstripe/silverstripe-framework/blob/eb069e605d8cbb029d9c0270a9081060f6ded229/i18n/i18nTextCollector.php#L461)

Fixing this could be as simple as creating the array without string keys. But we should consider if there are any cases where this would break existing classes.

Zauberfisch:
also, I just realised it merges the entries double, so this bug actually hides another bug:
https://github.com/silverstripe/silverstripe-framework/blob/eb069e605d8cbb029d9c0270a9081060f6ded229/i18n/i18nTextCollector.php#L465

it passes the existing list to the recursion, so the recursive call to the method already adds new items to the existing list. After that it returns the full list, which is then array_merged with the existing list again, producing double records every recursion.

dhensby:
@Zauberfisch do you mind writing a test to prove this issue (Look at `NamespacedManifestClassTest` as a starting point) then we can get a patch in for these issues. sounds like a fairly hilarious bug....

hilariously frustrating

tractorcow:
I'm working on a fix for this.

tractorcow:
Fixed with #4373

Issue #3417, opened 2014-08-21
i18nTextCollector should respect module


tractorcow:
When running it on third party modules which share translations which belong to other modules (e.g. SiteTree.PLURALNAME) it should not duplicate those keys in the dependant modules. It results in a lot of double up of translation strings across modules.

Also `_t(__CLASS__.'.MyString')` isn't being extracted properly. See if there's a possible solution here.

tractorcow:
I've begun a cleanup of `i18nTextCollector` which should address this.

tractorcow:
Fixed with #4373

tractorcow:
Closed with https://github.com/silverstripe/silverstripe-framework/pull/4373

PR #4373, opened 2015-07-06
BUG Fix of multiple i18nTextCollector issues: #3797, #3798, #3417

tractorcow:
This PR is necessary to ensure better automated updates of language files. This includes the following:

- Deprecation notice (with no change in API) if running i18nTextCollector without 'merge=true'. I propose to make this true by default, with the ability to opt-out in 4.0 (merge=false or merge=0).
- Fix for modules "pollenating" other modules with foreign localisation. This is fixed by firstly fixing #3798, which could introduce translations into the wrong module, as well as #3417 by doing a class-lookup for modules against specific classes, so that the module which owns that class gets ownership of the string.
- All output keys are sorted prior to write, so that incremental changes are more robust, and should not result in random re-orderings.
- Empty locales no longer write empty `{}` to yml files.
- Hella phpdoc, refactoring, and code cleanup for readability and maintenance.
- Use DI to construct dependencies (standard practice).

This also includes an annoying fix to `ERROR [User Warning]: Cookie 'alternativeDatabaseName' can't be set.` when running from the CLI. This caused issues testing locally and was well worth simply turning off.

Normally I wouldn't try to force such massive refactors into 3.1, but we are looking to roll out a new release tool soon, and it's necessary to either bring i18nTextCollector up to standard, or else not do translations for 3.1.x anymore. Since this is quite a controversial change for this branch, I'd like to get @silverstripe/core-team 's feelings on how this could affect semver. Are we happy that the public methods of this class still offer the same interface, despite the reworked internals? I had some methods that were made public so that unit tests could access them, although I'd really rather they weren't (since it's new visible API).


tractorcow:
I wrote this pretty late at night; I think we should sit on this for a bit, and maybe tomorrow I'll do another review to fix any semver issues I notice.

tractorcow:
I've updated this PR to change all new public methods to protected, so as to not introduce any new public api, and updated the tests accordingly.

i18n/i18nTextCollector.php
```diff
@@ -27,12 +27,20 @@
  */
 class i18nTextCollector extends Object {
 	
+	/**
+	 * Default (master) locale
+	 *
+	 * @var string
```
dhensby: `string|null` it seems

i18n/i18nTextCollector.php
```diff
@@ -27,12 +27,20 @@
...
 	protected $defaultLocale;
 	
 	/**
-	 * @var string $basePath The directory base on which the collector should act.
+	 * The directory base on which the collector should act.
 	 * Usually the webroot set through {@link Director::baseFolder()}.
+	 * 
 	 * @todo Fully support changing of basePath through {@link SSViewer} and {@link ManifestBuilder}
+	 * 
+	 * @var string
```
dhensby: `string|null`

i18n/i18nTextCollector.php
```diff
@@ -44,98 +52,273 @@ class i18nTextCollector extends Object {
...
 
+	/**
+	 * Gets the currently assigned writer, or the default if none is specified.
+	 * 
+	 * @return i18nTextCollector_Writer
+	 */
 	public function getWriter() {
-		if(!$this->writer) $this->writer = new i18nTextCollector_Writer_RailsYaml();
+		if(!$this->writer) {
+			$this->writer = Injector::inst()->get('i18nTextCollector_Writer');
```
dhensby: Should this not use `setWriter` - if ever there is any logic that needs to be attached whenever a writer is set, it should go in the setWriter, so that should be the only function to ever do this assignment

i18n/i18nTextCollector.php
```diff
@@ -44,98 +52,273 @@ class i18nTextCollector extends Object {
...
-		$modules = scandir($this->basePath);
-		$themeFolders = array();
+	/**
+	 * Gets the list of modules in this installer
+	 * 
+	 * @param string $directory Path to look in
+	 * @return array List of modules as paths relative to base
+	 */
+	protected function getModules($directory) {
+		$ignore = array('assets', 'vendor');
```
dhensby: these ignored dirs aren't configurable and considering they don't really do anything that wouldn't be caught by the check for configs in the module, I don't see much value to it.

In theory someone COULD change the assets dir to be something else and then have a module called assets...

tasks/i18nTextCollectorTask.php
```diff
@@ -31,13 +33,50 @@ public function init() {
...
+	 * @param SS_HTTPRequest $request
+	 */
+	protected function getIsMerge($request) {
+		$merge = $request->getVar('merge');
+		
+		// Default to false if not given
+		if(!isset($merge)) {
+			Deprecation::notice(
+				"4.0",
+				"merge will be enabled by default in 4.0. Please use merge=true"
```
dhensby: please use merge=false ?

tasks/i18nTextCollectorTask.php
```diff
@@ -31,13 +33,50 @@ public function init() {
...
+			);
+			return false;
+		}
+
```
dhensby: This is kind of picky, but can you clean whitespace from EOL ?

tasks/i18nTextCollectorTask.php
dhensby: If the current behaviour is that no flag means `?merge=false` then the deprecation should tell people what they need to change to retain the current behaviour, not what they need to do to get the new behaviour

i18n/i18nTextCollector.php
kinglozzer: I always thought you could leave the `null` out?

i18n/i18nTextCollector.php
kinglozzer: `ASSETS_DIR` is probably more suitable

dhensby:
### Thoughts on going into 3.1

Everything I've discussed with @sminnee has concluded that only minor patches should be going into 3.1, @chillu has also echoed that in issue discussion on the issue tracker. This is a big refactor and bug fix which, by the previous assertion, shouldn't go into 3.1.

If it's going to vastly simplify/improve releases, maybe it's worthy?

### SemVer

If it's a patch fix, then no new APIs; if the old APIs still do what they used to (minus bugs) then that's fine.

### Testing

Those methods you made public for tests shouldn't be public. To test non-public APIs you should use `ReflectionMethod`

```php
public function testPrivateMethod() {
    $object = new MyObject();
    $method = new ReflectionMethod('PrivateMethod', $object);
    $method->setAccessible(true);
    $value = $method->invoke($object);
    $this->assertEquals('ExpectedValue', $value);
}
```

i18n/i18nTextCollector.php
dhensby: I just think there's no benefit to the ignores. The code goes over all of the folders anyway and then checks for a _config.php or _config folder, so it'd never be a candidate to be ignored. Then the ignores (which can't be configured) are removed from all the candidates...

i18n/i18nTextCollector.php
dhensby: @assertchris ?

tasks/i18nTextCollectorTask.php
tractorcow: Ok, fair enough.

tractorcow:
That's what I used `i18nTextCollectorTest_Collector` for, in order to make public accessors for protected methods. :D I went over the collector and made sure that there were no new public methods; If there are then point them out and I'll fix them.

It is a refactor, but it's a refactor + fix, not a refactor + new api. The behaviour of the old textcollector was incorrect in many places (e.g. assuming that included templates are always a part of the same module as the parent, which is not true).

If this can't be put into 3.1, then the alternative is to make this a new module, and instead of running the core textcollector, then we ignore it and run this module. However this hugely complicates the release process more than it needs to be, and we miss out on the benefit of this fix.

I think we are missing the bigger picture of semver; Bug fixes SHOULD go into patch releases.

I guess we need another person to review? It's not an easy fix I understand, so it's good to pass it through rigorous checking.

i18n/i18nTextCollector.php
tractorcow: I think it's always ment to be set as string.

i18n/i18nTextCollector.php
tractorcow: Same as above; Is always a string.

i18n/i18nTextCollector.php
tractorcow: I've updated this.

i18n/i18nTextCollector.php
tractorcow: I've removed these exemptions; We now rely on the typical module check to filter these.

tasks/i18nTextCollectorTask.php
tractorcow: I've updated this.

tasks/i18nTextCollectorTask.php
tractorcow: Done.

tractorcow:
@dhensby I've fixed the issues you've noted, and ensured we don't have any new public api.

I've also removed the CLI / cookie issue and put this into its own PR. See https://github.com/silverstripe/silverstripe-framework/pull/4386

dhensby:
Ok, if there are no new public APIs then I'm happy with it to go into 3.1, I'm just repeating what the other general view on the subject is (regarding non-trivial patches)

i18n/i18nTextCollector.php
dhensby: The point is that it's initially set to null so it has the potential to be null or a string.

sminnee:
As it stands, the most recent iteration of this change doesn't impact public APIs, so I don't see it as breaking semver.

However, I have been recommending another rule which is "don't do refactorings or changes that have a moderate risk of side-effects/regressions in point releases unless they're needed to fix critical bugs". So, if this patch was an feature for website developers or end-users, I would say that it's too invasive for a point-release changes.

However, the refactoring is limited to i18ntextcollector, which for the most part isn't so much a developer tool as part of our packaging process. Because of that, and since the change will make it easier for us to reliably produce subsequent 3.1.x releases, I think that it's good for the 3.1 branch. **But I'd like to get 2 more thumbs-up from the core team before merging.**

In 4.0, I think that we should refactor this code into a separate module that is a require-dev dependency on each module that makes use of translations: it can be checked out by people who maintain the module, but not those who use it. Note that this will only make sense if the folder-structure-flexibility work is also completed, which also is on the cards for 4.0.

dhensby:
+1 from me

tractorcow:
Yeah its definitely more suited to being a Dev dependency. Thanks for the
review Sam. :)

willmorgan:
This is a bit of a semver edge case. 

Looking at what this is bringing to the project's release process, and considering that this is more tooling than frequently relied-upon userland code, I think the potential headache of opening up API methods as public in 3.1 are outweighed by the bugfixes.

I would rather see this go in to 3.2, but realistically the new methods probably won't do any harm, and a quick refactor of userland code to use new method names should they clash won't be a problem. It'd be crazy to develop on 3.1.14 and then deploy on 3.1.13, for instance, so... tentatively happy for this to go in to 3.1.

:+1:

sminnee:
That's good enough for me; I'm merging.

PR #4374, opened 2015-07-06
Enable *_Readonly subclass on FormField

ntd:
When rendering read-only fields only the *_Disabled subclass was checked
for in FormField::performReadOnlyTransformation().

Added a *_Readonly check and kept the *_Disabled check for backward
compatibility.

Cleaned up all subclasses by using removing bogus overrides and using
parent::performReadOnlyTransformation() instead of
castedCopy('*_Readonly').

### DIFF
--- a/forms/FormField.php
+++ b/forms/FormField.php
@@ -1039,10 +1039,16 @@ class FormField extends RequestHandler {
 	 * @return FormField
 	 */
 	public function performReadonlyTransformation() {
-		$readonlyClassName = $this->class . '_Readonly';
+		// Strip suffix so the transformation works on already transformed classes
+		$class = preg_replace('/(_readonly|_disabled)$/i', '', $this->class, 1);
 
+		$readonlyClassName = $class . '_Readonly';
+		$disabledClassName = $class . '_Disabled';
 		if(ClassInfo::exists($readonlyClassName)) {
 			$clone = $this->castedCopy($readonlyClassName);
+		} elseif(ClassInfo::exists($disabledClassName)) {
+			Deprecation::notice('3.5', "Using $disabledClassName as a fallback for $readOnlyClassName will be removed");
+			$clone = $this->castedCopy($disabledClassName);
 		} else {
 			$clone = $this->castedCopy('ReadonlyField');
 		}
--- a/tests/forms/FormFieldTest.php
+++ b/tests/forms/FormFieldTest.php
@@ -159,6 +159,16 @@ class FormFieldTest extends SapphireTest {
 		$this->assertContains('readonly="readonly"', $field->getAttributesHTML());
 		$field->setReadonly(false);
 		$this->assertNotContains('readonly="readonly"', $field->getAttributesHTML());
+		$this->assertInstanceOf('ReadonlyField', $field->performReadonlyTransformation());
+
+		$field = new FormFieldTest_TestField('MyField');
+		$this->assertInstanceOf('FormFieldTest_TestField_Readonly', $field->performReadonlyTransformation());
+
+		$field = new FormFieldTest_TestField_Readonly('MyField');
+		$this->assertInstanceOf('FormFieldTest_TestField_Readonly', $field->performReadonlyTransformation());
+
+		$field = new FormFieldTest_TestField_Disabled('MyField');
+		$this->assertInstanceOf('FormFieldTest_TestField_Readonly', $field->performReadonlyTransformation());
 	}
 
 	public function testDisabled() {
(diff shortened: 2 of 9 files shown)


tests/forms/FormFieldTest.php
```diff
@@ -159,6 +159,9 @@ public function testReadonly() {
 		$this->assertContains('readonly="readonly"', $field->getAttributesHTML());
 		$field->setReadonly(false);
 		$this->assertNotContains('readonly="readonly"', $field->getAttributesHTML());
+		$this->assertInstanceOf('FormField', $field->performReadonlyTransformation());
+		eval("class FormField_Readonly extends FormField {}");
```
tractorcow: Is there a way we can do this without the eval?

ntd: `eval()` is the only way I found to dynamically define a class. If this is a problem I can just drop the test with FormField_Readonly defined.

kinglozzer: @ntd Could you create two formfield subclasses at the bottom of that test file and use those? `FormFieldTest_TestField` and `FormFieldTest_TestField_Readonly`, for example?

ntd:
@kinglozzer Yes, this can be done quite easily. I just pushed a commit that uses a dedicated field class.

ntd:
Rebased and squashed.

@sminnee I've rebased against 3.5, only after the push I've seen you set the base branch to 3.2. Is this ok or do you prefer I rebase this on top of 3.2?

sminnee:
3.5 is good thanks, we'll merge into that

sminnee:
However there is a test failure that needs to be sorted:

1) CMSProfileControllerTest::testExtendedPermissionsStopEditingOwnProfile
Exception: Object->__call(): the method 'getdatefield' does not exist on 'ReadonlyField', or the method is not public.
/home/travis/builds/ss/framework/core/Object.php:780
/home/travis/builds/ss/framework/forms/DatetimeField.php:342
/home/travis/builds/ss/framework/forms/DatetimeField.php:342

ntd:
For some reason, the failing test was calling `performReadonlyTransformation` on an already transformed class (`DatetimeField_Readonly`), so the code was looking for the `DatetimeField_Readonly_Readonly` class, hence failing. Updated the code to strip eventual suffixes before (i.e. `_Readonly` and `_Disabled`).

If this is accepted I'll made a PR to add the same logic to `performDisabledTransformation`, just to keep the two methods consistent in the look and in the behavior.

forms/FormField.php
```diff
@@ -1039,12 +1039,18 @@ public function setDisabled($disabled) {
 	 * @return FormField
 	 */
 	public function performReadonlyTransformation() {
-		$readonlyClassName = $this->class . '_Readonly';
+		// Strip suffix so the transformation works on already transformed classes
+		$class = preg_replace('/(_readonly|_disabled)$/i', '', $this->class, 1);
```
dhensby: I feel like it would be better that `ReadonlyField::performReadonlyTransformation()` should just do nothing instead.

forms/FormField.php
```diff
@@ -1039,12 +1039,18 @@ public function setDisabled($disabled) {
...
+		$readonlyClassName = $class . '_Readonly';
+		$disabledClassName = $class . '_Disabled';
 		if(ClassInfo::exists($readonlyClassName)) {
 			$clone = $this->castedCopy($readonlyClassName);
+		} elseif(ClassInfo::exists($disabledClassName)) {
+			Deprecation::notice('3.5', "Using $disabledClassName as a fallback for $readOnlyClassName will be removed");
+			$clone = $this->castedCopy($disabledClassName);
 		} else {
-			$clone = $this->castedCopy('ReadonlyField');
+			$clone = clone $this;
```
dhensby: We should keep the current behaviour of returning a `ReadonlyField`

forms/FormField.php
dhensby: I feel like it would be better that `ReadonlyField::performReadonlyTransformation()` should just do nothing instead.

forms/FormField.php
ntd: Yes, my bad: I screwed up the old behavior.

ntd:
@dhensby regardling the suffix stripping there's not much I can do about it, that is if I just append `_Readonly` to the class name (as I was originally doing) the `CMSProfileControllerTest::testExtendedPermissionsStopEditingOwnProfile` test fails (see the [relevant comment](https://github.com/silverstripe/silverstripe-framework/pull/4374#issuecomment-299882791)). Previously it was working because the class name was hardcoded.

I find the behavior of switching between `_Disabled` and `_Readonly` quite logic but in reality its primary purpose is to allow calling `performReadonlyTransformation` on a readonly field.

maxime-rainville:

maxime-rainville:
I've double checked to see if this could be worthwhile rebasing in SilverStripe 4. FormField now has a `performDisabledTransformation` which would probably get in the way of this change.

PR #4375, opened 2015-07-07
Improved label clipping in site tree

jonom:
Deeply nested site tree nodes show a few more characters and any clipped text in labels is marked as such (thanks @clarkepaul) (fixes https://github.com/silverstripe/silverstripe-cms/issues/1235)

labels

### DIFF
--- a/admin/css/screen.css
+++ b/admin/css/screen.css
@@ -862,8 +862,8 @@ form.import-form label.left { width: 250px; }
 .jstree-apple a { border-radius: 3px; }
 
 /* ensure status is visible in sidebar */
-.cms-content-tools .cms-tree.jstree li { min-width: 159px; }
-.cms-content-tools .cms-tree.jstree a { overflow: hidden; display: block; position: relative; }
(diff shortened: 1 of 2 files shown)


dhensby:
That is sexy

dhensby:
Hmmm, what's the value in clipping the long names, it seems they can't be uncovered. Shouldn't there be a way to see the full title? 

If the side bar is scrollable why not have wide tree elements?

jonom:
It's a balancing act - the reason the tree nodes are clipped now is so that the status can be visible when you roll over it (at least for the first 5 levels). If we don't clip the node length then the status text would require scrolling to get to it, which is one of the problems the initial status dots PR was trying to address. Trade-off is long labels get clipped but I think it's worth the cost.

jonom:
p.s. see [idea for revealing full title here](https://github.com/silverstripe/silverstripe-cms/issues/1235#issuecomment-119255060). I don't think it's super important to be able to see the entire label - Menu Titles should be short and if you click the link you'll see the full title in the editing interface, but a tooltip could be a solution for people that need it.

PR #4377, opened 2015-07-07
Improved site tree tooltips

jonom:
Included full title and friendly singular name for page type. Also standardised white space and use of colon character in translations.

screen shot 2015-07-07 at 2 21 24 pm

### DIFF
--- a/admin/code/LeftAndMain.php
+++ b/admin/code/LeftAndMain.php
@@ -1958,8 +1958,9 @@ class LeftAndMain_TreeNode extends ViewableData {
 		$obj = $this->obj;
 		return "<li id=\"record-$obj->ID\" data-id=\"$obj->ID\" data-pagetype=\"$obj->ClassName\" class=\""
 			. $this->getClasses() . "\">" . "<ins class=\"jstree-icon\">&nbsp;</ins>"
-			. "<a href=\"" . $this->getLink() . "\" title=\"" . _t('LeftAndMain.PAGETYPE','Page type: ')
-			. "$obj->class\" ><ins class=\"jstree-icon\">&nbsp;</ins><span class=\"text\">" . ($obj->TreeTitle)
+			. "<a href=\"" . $this->getLink() . "\" title=\"("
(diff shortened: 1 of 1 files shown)


chillu:
Man, you're on fire! :D Unfortunately, the *.yml translations need to be applied directly on transifex, otherwise they'll be overwritte again later (see http://docs.silverstripe.org/en/3.1/contributing/translations/). Which then needs to be synced back into the master codebase, and merged "down" into 3.1. All in all, a PITA. They'll also rely on the double colon in `PAGETYPE` in other places. I think the path of least resistance is a `str_replace` for your use case.

jonom:
Problem is that only some translations have the colon included, others don't, plus some include a space and others don't. Agree on the path of least resistance but seems a shame. Is this enabled for master? http://docs.transifex.com/faq/#can-i-update-source-files-automatically if so I could submit the translations as a separate PR to master...

jonom:
@chillu @tractorcow happy to remove the translation changes and provide a workaround for the colon/space inconsistencies but just wondering if you guys are aware of the auto-update option on Transifex? Assume you probably are and there is a reason we can't use it but if not it would be great if we could get that set up to make it possible to edit translations directly through PRs http://docs.transifex.com/faq/#can-i-update-source-files-automatically let me know your thoughts and I'll adjust this PR as necessary.

chillu:
Hey Jono, I didn't know about that option, very handy! Unfortunately it only seems to be syncing "source files" (`en.yml`), not translation files. So we'd still need to merge all those translation files into master with your edits, merge with any new translations from Transifex, and then push the merged result back up to Transifex. Doable, but a bit more work than changing the PR to use `str_replace()` instead ;)

jonom:
Mmm true dat! Thanks for looking in to it. Will update this PR shortly.

jonom:
Okay that's done now. Also changed the brackets to round ones as they look a bit friendlier.

screen shot 2015-07-16 at 8 50 25 am

chillu:
Nice!

PR #4379, opened 2015-07-07
BUG Fix issue when SS_ALLOWED_HOSTS is used in CLI

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/commit/9c8fa5132137ef050dcbd12a5b5e7634cfab51fa for discussion.

### DIFF
--- a/core/Constants.php
+++ b/core/Constants.php
@@ -184,9 +184,10 @@ if(!isset($_SERVER['HTTP_HOST'])) {
 	}
 }
 
-if (defined('SS_ALLOWED_HOSTS')) {
+// Filter by configured allowed hosts
+if (defined('SS_ALLOWED_HOSTS') && php_sapi_name() !== "cli") {
 	$all_allowed_hosts = explode(',', SS_ALLOWED_HOSTS);
-	if (!in_array($_SERVER['HTTP_HOST'], $all_allowed_hosts)) {
+	if (!isset($_SERVER['HTTP_HOST']) || !in_array($_SERVER['HTTP_HOST'], $all_allowed_hosts)) {
 		header('HTTP/1.1 400 Invalid Host', true, 400);
 		die();
 	}


nyeholt:
Given this is probably the best spot to continue the discussion - yes, this should have an isset() check, and probably a cli check as described, for cases where it may end up without one defined. While the problem of an invalid hostnames does still affect CLI scripts (eg generation of static cache), if someone's able to inject it into a cli executed process prior to this point, there's potentially bigger issues going on :). I'd had no issue with running this over CLI prior to committing, but our project base has a default FILE\_TO mapping defined which is why I'd overlooked it. Another (probably incorrect!) assumption being that if you were of a mind to update your \_ss\_env to define a security parameter, you'd probably have defined the FILE\_TO\_URL also. 

As to whether it should be .htaccess or not, my thinking against that would be that it doesn't cover non-apache servers, and in some cases you may not have access to the apache config to make these changes. I've also got in the back of my mind the possibility of automatically generating this list based on subsites/multisites configuration, but haven't really gone down that path as yet. 

Pinging @dhensby and @kinglozzer to let them know my thoughts are here instead of elsewhere :)

dhensby:
I wouldn't put a 400 response in as part of a patch release. 

The cli normally spits notices for underlined index where the mapping isn't defined and I'd expect the app fails to get this far without it over http

I see this big fix more as simply not performing an irrelevant check on cli executions

kinglozzer:
Thanks @nyeholt @tractorcow. My concern was partly because the original PR looked a little rushed through (merged within a few minutes, released under a new tag an hour later).

> I wouldn't put a 400 response in as part of a patch release.

Given that this is *opt-in* and a security issue I think that should be okay.

I was still a bit unsure whether this is really an API we should offer, instead of leaving it at a server level. However, as @nyeholt pointed out it’s necessary for those without access to apache/nginx configuration options and it looks like [Symfony offers the same thing](https://github.com/symfony/HttpFoundation/blob/ecb2d8260ab57b4445f901ee1e0e0067fec26c55/Request.php#L1295-L1311).

Any objections to merging this @dhensby?

dhensby:
nope :)

Issue #4380, opened 2015-07-08
CountryDropdownField list should be configurable


stadtwald:
**WAS:** CountryDropdownField includes several non-existent countries

The CountryDropdownField (source: forms/CountryDropdownField.php) includes several countries that no longer exists (e.g. Soviet Union).  I have prepared a patch which filters these countries out.  I have used ISO 3166-3 when creating this filter, but have not included some countries in the filter because they were not present in the dropdown.  I also note that some non-existent country's ISO codes seem to have been reused, so simply putting all of ISO 3166-3 in the filter would be dangerous.

My patch is:

``` diff
diff --git a/forms/CountryDropdownField.php b/forms/CountryDropdownField.php
index 61a5f0e..cd989a6 100644
--- a/forms/CountryDropdownField.php
+++ b/forms/CountryDropdownField.php
@@ -50,6 +50,18 @@ class CountryDropdownField extends DropdownField {

            // We don't want "unknown country" as an option
            unset($source['ZZ']);
+
+           // We don't want a number of countries which have ceased to exist
+           unset($source['SU']); // Soviet Union
+           unset($source['BQ']); // British Antarctic Territory
+           unset($source['BY']); // Byelorussian SSR
+           unset($source['CT']); // Canton and Enderbury Islands
+           unset($source['NQ']); // Dronning Maud Land
+           unset($source['FX']); // France, Metropolitan
+           unset($source['FQ']); // French Southern and Antarctic Territories
+           unset($source['NT']); // Iraq-Saudi-Arabia Neutral Zone
+           unset($source['PZ']); // Panama Canal Zone
+           unset($source['CS']); // Serbia and Montenegro
        }

        parent::__construct($name, ($title===null) ? $name : $title, $source, $value, $form);
```

stadtwald:
Just realised that BY should not be filtered out.  Adjusted patch:

```diff
diff --git a/forms/CountryDropdownField.php b/forms/CountryDropdownField.php
index 61a5f0e..cd989a6 100644
--- a/forms/CountryDropdownField.php
+++ b/forms/CountryDropdownField.php
@@ -50,6 +50,18 @@ class CountryDropdownField extends DropdownField {

            // We don't want "unknown country" as an option
            unset($source['ZZ']);
+
+           // We don't want a number of countries which have ceased to exist
+           unset($source['SU']); // Soviet Union
+           unset($source['BQ']); // British Antarctic Territory
[...]

PR #4382, opened 2015-07-08
In CountryDropdownField, filter out countries which no longer exist as per ISO 3166-3

stadtwald:
Details on issue #4380


dhensby:
Ok, so my biggest concern is the fact this is not configurable. If we can move the keys that are removed to a config option, then I think this will be acceptable.

I'd also say this can go into 3 if it is configurable. I don't see a huge advantage in putting it into master if the long term goal is either to move away from zend or to upgrade it.

stadtwald:
Thanks @dhensby. Will work on a new patch to implement that.  Closing this PR.

PR #4385, opened 2015-07-08
Update compass config

flashbackzoo:
Adds sourcemaps and nested CSS format for debugging. Also sets config options for more consistent builds across environments.

Because there was a change to `config.rb` compass does a 'clean' before compiling which is the reason for all the CSS and sprite changes in this commit.

Compass' default environment is `:development` which we've been treating as `:production` until now. So `compass compile` now generates CSS in an expanded format and also generates sourcemaps. For pushing back upstream we should use `compass compile -e production`.

/cc @jonom


jonom:
So excited for source maps! I'm not a huge fan of the development/production approach here though - is it harmful to leave the source maps on all the time? People can access the scss files in a browser either way so I don't think it poses a security risk. I'm not sure there's much advantage doing nested output for dev versions either, most people would view the CSS in dev tools where it's reformatted anyway wouldn't they? And if they have to ultimately edit an SCSS file to change things I don't think it really matters if the CSS output file is more readable. With this approach potential contributors would have to set up the project and re-compile the SCSS just to trace a style to the source, think that barrier will deter a lot of community members from investigating and fixing a CSS problem.

In short I'd much rather leave source maps turned on all the time and always produce compact output so we don't create a lot of extra compiling work.

flashbackzoo:
I've added sourcemaps to `.gitignore` so they wont be in the repo.

Yeah that makes sense - I'll update it so we're always generating compact CSS.

flashbackzoo:
Update so we're always generating compact CSS and always generating sourcemaps

jonom:
Why not include the sourcemaps in the repo?

flashbackzoo:
There's already a comment in the config which says 'Disable comments in the output. We want admin comments to be verbose'. Sourcemaps display comments, so I've excluded them. I don't think it's security issue, so they might be OK to include, anyone else want to weigh in here? They're not a performance issue as they're only requested if devtools is open.

flashbackzoo:
It's an opensource project though - so you can see the comments if you really want to...

jonom:
I think that line means that we want lots of comments in the source file, but that bloats the file, so we won't include them in the output that gets sent to people's browsers. Source maps and the target SCSS files are only loaded by the browser if you have dev tools open so should be all good I think :)

jonom:
> They're not a performance issue as they're only requested if devtools is open.

Oh you already said that heh

flashbackzoo:
OK - sorry for the mucking around. I've included them :smile:

jonom:
Nice! Thanks. Looks great to me.

adrexia:
I approve of sourcemaps, but don't tend to need them myself. But whatever helps other devs track things down is a gain. :+1:

tractorcow:
Does anyone have a feeling on whether unix_newlines as standard is appropriate for css output? Shouldn't text transmitted by HTTP use `<some standard>` newline?

adrexia:
I'm not sure it matters?

From: http://sass-lang.com/documentation/file.SASS_REFERENCE.html#unix_newlines-option

:unix_newlines
If true, use Unix-style newlines when writing files. Only has meaning on Windows, and only when Sass is writing the files (in Rack, Rails, or Merb, when using Sass::Plugin directly, or when using the command-line executable).

flashbackzoo:
Yeah I added it to increase the consistency of output across environments. I read the docs @adrexia posted when I added the option, and read it to mean unix newlines are used by default, as it's only used on Windows.

PR #4386, opened 2015-07-08
BUG Fix cookie errors when running in CLI

tractorcow:
This resolves an issue where errors during unit tests on CLI could result in a cookie write being attempted, which would often mask other errors in the output. 

`ERROR [User Warning]: Cookie 'alternativeDatabaseName' can't be set.`


dhensby:
Why is this even a thing? Setting db name cookies, I mean...

tractorcow:
You have no idea how many years I've put up with this issue without fixing it lol.

PR #4388, opened 2015-07-09
Remove deprecated methods from ViewableData

kinglozzer:


dhensby:
Totally not semver to remove this from anything but 4.0

kinglozzer:
Doh, quite right. Deprecation notice !== fatal error.

PR #4390, opened 2015-07-10
API respect custom attributes on OptionsetField and CheckboxSetField

tractorcow:
Partially reverts a change made in https://github.com/silverstripe/silverstripe-framework/commit/0c4ec4708bcbb8fb09a1bb9acf75639077cbaaa9; These templates did not have duplicate IDs (child ID attributes related to objects other than the parent) so that change was unnecessary.

Required to make module-specific hacks such as https://github.com/unclecheese/silverstripe-display-logic/commit/0ae22159c198db32f0ba10079947ebbbc80ec5e7 unnecessary.

See https://github.com/unclecheese/silverstripe-display-logic/issues/32

Must be merged with https://github.com/silverstripe/silverstripe-cms/pull/1241

cc @unclecheese


admin/scss/_actionTabs.scss
```diff
@@ -125,7 +125,7 @@ $border: 1px solid darken(#D9D9D9, 15%);
 					font-size: 12px;
 				}	
 
-				#Form_AddForm_PageType_Holder ul {
+				#Form_AddForm_PageType ul {
```
dhensby: Why are we even using IDs a selectors?

adrexia: I'm unsure this even applies to anything? The feature this applies to was never implemented, but got merged with the bottom actiontabs in 2012. :-/

tractorcow: We should not be, but I couldn't figure out why this selector was here. My suspicion is that if we remove it we'd find out soon enough why it was necessary, so I left it in place. :D

chillu:
A bit hard to follow (the commit doesn't actually revert lines of code introduced, but rather changes it to use `$AttributesHTML` which in turn uses `FormField->ID()`). But you're right, the `CheckboxSetField` and `OptionsetField` don't *need* to use `$HolderID`. But I'd say its consistent with other fields (same holder DOM structure). The problem with https://github.com/unclecheese/silverstripe-display-logic/issues/32 is that the form doesn't use `$AttributesHTML` in the first place, right? So you can't add `data-*` attributes to use for stuff like display-logic code?

chillu:
ping @tractorcow

tractorcow:
The `revert` I referenced wasn't a direct revert of the code; It was a reversion of the behaviour. The change in https://github.com/silverstripe/silverstripe-framework/commit/0c4ec4708bcbb8fb09a1bb9acf75639077cbaaa9 swapped out the `ID` with `HolderID`, which is arguable a change in behaviour.

Shouldn't the holder ID be on the Field_holder.ss template, not on the Field.ss template, by convention anyway? :D

tractorcow:
Missed https://github.com/silverstripe/silverstripe-cms/pull/1241 :D

PR #3882, opened 2015-02-12
Include php version in default cache folder name

JorisDebonnet:
Updated fix (from pull request #3881) to fix issue #2986: tying the automatically generated cache folder to the php version.

Generates folder names such as: silverstripe-cache-php5.6.5-home-user-...

PR #4391, opened 2015-07-10
extends temp folder versioning for local cache

pine3ree:
Allow php versioning for local cache folder
/path/to/public/silverstripe-cache/user-php-PHP_VERISION

### DIFF
--- a/core/TempPath.php
+++ b/core/TempPath.php
@@ -37,6 +37,7 @@ function getTempFolderUsername() {
 	}
 	if(!$user) $user = 'unknown';
 	$user = preg_replace('/[^A-Za-z0-9_\-]/', '', $user);
+	$user .= '-php-' . preg_replace('/[^\w-\.+]+/', '-', PHP_VERSION);
 	return $user;
 }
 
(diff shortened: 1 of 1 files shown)


dhensby:
This is already in? see #3882

kinglozzer:
Hmm, so this change would result in putting the PHP version in the _user_ directory name, rather than the _site_ directory name? What’s the benefit of this over the current behaviour?

Test failures look related.

pine3ree:
Hi@kinglozzer 

this let you have local cache directories like the following:

/path/to/project/silverstripe-cache/user-php-5-4.42
/path/to/project/silverstripe-cache/user-php-5-5.26

under the same standard

/path/to/project/silverstripe-cache

when this directory is found, whil the original PR is adding versioning to the cache directory under system tmp only

It also keeps all the versioned cache directories under a unique common cache folder for the project

I haven't updated the tests, wanted to know if this could be useful to other ppl first. I believe they are just checking the original versioning PR name generations.

kinglozzer:
Are there any obvious benefits besides the slightly tidier directory structure? The only other benefit I can think of is if the user has created a `docroot/silverstripe-cache` directory and then upgraded their PHP version, they won’t get a unique directory name so will experience the crash that this was intended to resolve. Moving the PHP version to the user (i.e. `docroot/silverstripe-cache/www-data-php-5-5-26`) would solve this

pine3ree:
@kinglozzer 
The benefit you recognized is the main purpose of this PR. :-)

dhensby:
Ah-ha, I see.

Ok, then I think the current version just needs to be moved onto the username rather than having it twice.

tractorcow:
I think this needs moving to 3 branch, squashing, and fix tests please, then it can be merged.

It might be a bit late for 3.2 sorry.

sminnee:
Anyone want to reapply this to the `3` branch?

dhensby:
> Anyone want to reapply this to the 3 branch?

and fix the tests

tractorcow:
Closing, since we can't merge it into 3.2.

PR #4393, opened 2015-07-12
FIX #4392: Ensure headers are checked first before being clobbered

### LINKED ISSUE #4392: HTTP::add_cache_headers() always overrides headers that are already set.
The method `HTTP::add_cache_headers()` maintains global state (which may be fine for initializing defaults) will always override headers which may already be setup in a passed `SS_HTTPResponse` object. This is pretty bad when you need to setup cache control headers but don't want to maintain global state and, for consistency, may have other headers which need to be setup but don't happen to have setter methods on the `HTTP` object.

Right now I need to setup cache headers on dynamic pages on an as-needed basis, but since `Controller`'s `handleRequest()` method calls this on every page load, my only way around this is to modify the framework to get this to work as I need.

patricknelson:
Proposed fix for  #4392


kinglozzer:
Makes sense to me. Given the very minimal regression risk, and that there’s no workaround if you’re hitting `Controller::handleRequest()`, I think this is a valid candidate for 3.1. Tests please! :)

@dhensby you’ve been doing some cache stuff for 3.2, any feedback on this?

patricknelson:
Also let me know what sort of test coverage you'd prefer (not 100% on the process here, pretty solid on the general PR/issue process at least). In my case I'd plan on writing coverage mainly just for the change I made, throwing in a bunch of candidate headers which I know are common HTTP headers and seeing if they get changed after passing through `::add_cache_headers()` and that's basically it. 

Just let me know if it'd need to be any more than that!

dhensby:
Hmm, there's quite a distinct behavioural change here.

The header that takes precedent is changing (from the one that was going to be set by the function to the one that is already set on the body object).

Whether that's a bug or a feature, I'm happy to discuss, but at the moment I think that shouldn't change in a patch release unless there's a compelling reason to do so.

Where does this function get called in the execution chain? Is at at the end or beginning of a request? Presumably if someone is calling this method by hand it's because they want to reset the headers with those determined by the internal logic of the function? Otherwise they can set their own and avoid this function.

However, if it's called at the end of the execution (by the framework) just before being sent, then we obviously leave no opportunity for the dev to have set these headers themselves (and is thus a bug).

kinglozzer:
@dhensby I’d argue that it’s a bug, the way I see it there are these scenarios:

1. You’ve not set any headers, `HTTP::add_cache_headers()` is doing it for you - no change.
2. You’ve set a cache header in your code, `HTTP::add_cache_headers()` then always overrides it with its own. This PR would result in *your* headers being output, instead of the global ones from `HTTP` (which is what you intended when you set your own cache headers)
3. You’ve set a header, then intentionally overridden it (either by just letting framework override it, or by calling the method yourself) using `HTTP::add_cache_headers()`. I think it’s pretty unlikely anyone has done this? How likely is it that someone would go to the effort of adding their own caching headers to then reset them?

The method is called fairly late in the request: https://github.com/silverstripe/silverstripe-framework/blob/8331171f2c6cd6d284a357fa5dee6e44219c22bd/control/Controller.php#L165. You don’t have a chance to skip it/override it without overriding `handleRequest()` in your controller.

Looking at it again, it might be fairly simple to work around, so I guess we could delay this change, a workaround for now could be:

```php
public function handleRequest(SS_HTTPRequest $request, DataModel $model) {
  $response = parent::handleRequest($request, $model);
  $response->addHeader('foo', 'bar');
}
```

@patricknelson could that work for you?

dhensby:
I think that if you're setting your own cache headers then you're doing something quite advanced and you should be implementing other work arounds like the one you've suggested.

Having said that, the HTTP utility class is rather... crap. it's inextensible, it's got a lot of hard coded logic and it's just not flexible. I'd welcome patches to do this properly - really I think it should be part of the `HTTPResponse` object, maybe with a class in charge of determining cache headers...

But @patricknelson it sounds like you're doing something quite advanced and that, really, you want to be computing your cache headers at the last possible moment, as @kinglozzer suggests, it's probably better that you overload `handleRequest` to add your headers in AFTER the framework has had enough of trying to be helpful.

patricknelson:
You're right @dhensby right now I am doing something pretty advanced, but it's not extraordinary (or unusual these days) for people to be using upstream CDN's (e.g. cloudflare or in my case, Akamai). I need to be able to set these headers without too much "magic" on the SilverStripe side overriding something I had already explicitly told it, causing me to further hunt down, trace the execution path (as @kinglozzer just had to do) in order to determine a method of working around this. To me, it's very hacky. I understand the intent SS has which is to be helpful and setting headers, which is is currently doing and that's good - I don't want it to A.) Pretend it knows better than me (hyperbole here, just for the sake of making a point) and B.) Require more contortions for something that should be fairly simple.

The current workaround recommended by @kinglozzer may be what I'll have to do if I want to maintain compatibility with future bugfixes/patches if this merge doesn't make it into 3.1 (which I agree, some people may coincidentally be relying on this, which I believe is a bit of an unintuitive quirk, if they *set* some headers ahead of time and *don't realize* they aren't all actually being *sent*).

Think of it from the perspective of a developer picking up the system (or like me working on it for 7+ mo) who may not know a huge amount about SS. Its much more well known to perform HTTP header related actions on `init()` (and the framework/CMS already implicitly condones this in some areas as it is). 

On the flip side, it may make more sense to setup the cache headers earlier, rather than later. Especially since so much global state is being maintained; it's best for that to work more as an initialization and/or set of "sane" defaults, rather than dumping it in so late in the game.

dhensby:
@patricknelson I agree, really.

But, this solution is a band-aid to a bigger architectural problem. As I said in my last comment HTTP is far too niaeve and doesn't provide any mechanism for customisation.

What would be much better is a cache header computator as part of the HTTPResponse object that could be replaced or customised to work with your requirements (or removed entirely). Then the headers can be calculated at the last possible moment (so there are no edge-cases around changed content after headers are set).

In fact, I think you can actually set a `RequestProcessor` to listen to the `postRequest` event (which happens JUST before the response is echoed to the browser) as the best time to do advanced work calculating cache headers.

Anyway, I just think we need a more complete, flexible and versatile solution than this crappy `HTTP` utility class.

dhensby:
The other risk is, of course, that a Response object could have default headers that SHOULD or at least COULD be overridden here; perhaps that's realistic intended behaviour.

@kinglozzer There's no point us blocking this because it's not the perfect solution. It's the minimum viable (and risk) way of not overriding a dev's explicitly set headers. If you're happy with it, take it in.

patricknelson:
I agree @dhensby. I think I remember having another issue somewhere else that I had to workaround (can't recall for the life of me now) and was frustrated that I couldn't use the injector library to override it somehow.

Also, it might also be feasible to set this up with the Injector and then abstract this method into two methods: `add_cache_headers()` and `get_cache_headers()` where the former method would rely on the response from the latter. That way I can override `add_cache_headers()` with my own method (via the Injector class I'd setup) and then pull in the values from `get_cache_headers()` myself so I don't have to reproduce all of that code.

dhensby:
yep

dhensby:
or you'd inject the dependency into SS_HTTPResponse

dhensby:
heh, looks like this had been done to some extent: https://github.com/silverstripe-labs/silverstripe-controllerpolicy

This PR may break the expected behaviour @chillu ?

dhensby:
Hmm, doesn't look like it because it does hook into precisely the `postRequest` function I had spoken about earlier.

patricknelson:
@dhensby, @kinglozzer and @tractorcow  I'm assuming this will also need to be rebased on `3.2` (bug fix). It can be merged post release and rolled into `3.2.1` but just to save myself work in the future since I'm trying to organize all of this now (I've got a spreadsheet going now of all this stuff).

Same question actually applies to https://github.com/silverstripe/silverstripe-cms/pull/1287

dhensby:
@patricknelson I think this can go into 3.2, yes.

I'm still cautious about this approach, though... Can we safely say this doesn't break expected behaviour and is indeed a bug and not intended?

patricknelson:
Damn @dhensby -- I pushed to the same branch name (rewrote history). I forgot I should probably close this first. To fix the messed up log above I think I'll rebase again against `3.1` and just setup a different branch :neutral_face:

But yes -- I'll be reopening this PR against `3.2` and I think I'll need to refactor it a bit to abstract the determination of which headers to send away from the part that actually sends them (the part with the "bug"). I'll post a link here when I create the PR.

patricknelson:
@dhensby @kinglozzer 

New PR is here: https://github.com/silverstripe/silverstripe-framework/pull/4673

PR #4397, opened 2015-07-15
Update tree status after bulk action is applied

flashbackzoo:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1226

### DIFF
--- a/admin/javascript/LeftAndMain.BatchActions.js
+++ b/admin/javascript/LeftAndMain.BatchActions.js
@@ -350,8 +350,9 @@
 					complete: function(xmlhttp, status) {
 						button.removeClass('loading');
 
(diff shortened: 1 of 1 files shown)


dhensby:
The comment immediately above now no longer relates directly to the code that is there. Please could you explain a little why the refresh is the correct method?

flashbackzoo:
Hey, sorry for the lack of description, I've updated the comment too.

`uncheck_all` does what it says on the tin.

`refresh` saves the open 'branches' (expanded items stay expanded) and refreshes the tree, so the correct CSS class are applied.

The issue was that old CSS classes were present after a bulk action was applied, so pages appeared to still be in their old state.

dhensby:
Thanks

PR #4399, opened 2015-07-15
Correct Memcached config in example

inflammable:
Unlike Libmemcached, the Memcached backend server config must be nested inside an array() with a single key 'servers' - otherwise the default host of 127.0.0.1 is used.

array(
	'servers' => array(
		'host' => 'localhost', 
		'port' => 11211, 
		'persistent' => true, 
		'weight' => 1, 
		'timeout' => 1,
		'retry_interval' => 15, 
		'status' => true, 
		'failure_callback' => ''
	)
)


dhensby:
thanks

PR #4401, opened 2015-07-15
Fixed type in code example

marijnkampf:


dhensby:
Thanks.

FYI: When you make docs changes please prefix the commit message with "DOCS" (it helps us with change logging)

marijnkampf:
I'll try to remember, but I only submit changes to the docs occasionally, so might forget again. Would there be a way to add this in a note when editing the docs?

PR #4402, opened 2015-07-15
Add Convert::fqcn2htmlname()

nedmas:
Create helper function for converting namespaced class names to html safe strings for use in `LeftAndMain.php` and `CMSMenu.php` see PR #4167

### DIFF
--- a/core/Convert.php
+++ b/core/Convert.php
@@ -14,6 +14,8 @@
  *  html: HTML source suitable for use in a page or email
  *  text: Plain-text content, suitable for display to a user as-is, or insertion in a plaintext email.
  *
+ *  fqcn: A PHP fully qualified class name
+ *
  * Objects of type {@link ViewableData} can have an "escaping type",
  * which determines if they are automatically escaped before output by {@link SSViewer}.
  *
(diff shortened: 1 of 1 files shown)


chillu:
I think this is scope creep for the framework, there's little benefit in increasing the core API (rarely used feature), and its easy to replicate in custom code. Also, no tests. Unless another core member objects within the next day or so, I'm closing this pull request.

chillu:
I should clarify "custom code" here: For the two places where this is used in the CMS (e.g. menu), I'd inline it there and document its intent.

tractorcow:
It's a very narrow use-case scope; I would recommend that this problem is `solved` rather by a comment on the `Convert::raw2htmlname` that if characters such as slashes have meaning, they should be escaped to one of the safe characters prior to being passed into this method.

nedmas:
I agree that scope creep should be minimised and am happy for the core members to decide on that. However with 4.0 and the potential for namespacing framework I would suggest that the need for such a conversion will increase beyond the initial two instances (potentially FormFields and Db tables).

chillu:
Agree with Damian on that one, so closing. In general, we want to decrease the amount of global and static dependencies in our code - and that's one way to do it :)

robbieaverill:
For reference, [`TaskRunner` does this](https://github.com/silverstripe/silverstripe-framework/blob/master/src/Dev/TaskRunner.php#L137) - as an example of a one off use case

Issue #4254, opened 2015-06-05
RFC-3: Retiring and replacing the web test runner


willmorgan:
RFC-3: Retiring and replacing the web test runner
-----
Status: Draft

---

The core team recently voted in favour of retiring the current web test runner (`TestRunner`) (accessible via `dev/tests`).

The current web test runner functions differently to the sake/phpunit test runners, is susceptible to superglobal leakage, and is basically a bit of a pain to maintain because of the difference in running environments. Also, it's extra code weight and attack surface for live environments when it should really be a dev dependency.

This RFC suggests it is replaced with a module that:
* functions as a web wrapper to call the test runner via the `phpunit` binary
* with a [listener](https://phpunit.de/manual/4.7/en/extending-phpunit.html#extending-phpunit.PHPUnit_Framework_TestListener) feeding back to the UI via WebSocket.
* would run only in dev mode
* strongly discourage being loaded as anything but a dev dependency

First steps required:
* community scrutiny
* a proof of concept :wink:

Ingo's initial post on this: https://groups.google.com/forum/#!topic/silverstripe-dev/I5dHy9dMVng

micmania1:
I would argue that it doesn't need to be put into a module - does anybody really want to maintain something we don't need?

chillu:
Hey @willmorgan, do you want to incorporate my post into this RFC? https://groups.google.com/forum/#!topic/silverstripe-dev/I5dHy9dMVng
For example, I had already outlined what the equivalent replacement for each `sake` command would be, detail which is important for an RFC. 

I think while a `phpunit` web wrapper and websocket implementation would be neat, its not required to get the functionality out of core (my main goal). I suggest we put `TestRunner` into a module as-is, mark it as unmaintained, and ask for maintainers on the mailing list. That allows users who rely on this feature to continue using it, but doesn't take up core maintainers time with maintenance (my secondary goal). My underlying assumption here is that its less work to just separate out the functionality than to create this new wrapper. Will, what do you think?

[...]

PR #4404, opened 2015-07-16
Remove TestRunner and JSTestRunner

chillu:
See https://github.com/silverstripe/silverstripe-framework/issues/4254 for background. As per latest state of discussion, not re-adding this functionality via a module. The docs still have too much duplication, reworked them slightly in addition to removing the oudated references. More consolidation work is required in there, but not a dependency on this pull request.


dev/SapphireTest.php
```diff
@@ -30,12 +28,7 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 	 * @var FixtureFactory
 	 */
 	protected $fixtureFactory;
-	
-	/**
-	 * @var bool Set whether to include this test in the TestRunner or to skip this.
-	 */
-	protected $skipTest = false;
```
tractorcow: I know this references TestRunner, but are you 100% sure this no longer has any use? Does markTestSkipped skip the entire class or just the one test method?

chillu: When `markTestSkipped()` is placed in `setUp()`, it'll skip every test (https://phpunit.de/manual/current/en/incomplete-and-skipped-tests.html#incomplete-and-skipped-tests.skipping-tests). Which I've made sure is the case for the few core uses. You're correct in saying that if somebody discovered this flag, and used without also using `markTestSkipped()`, this test would now be executed. I think that's an acceptable change for a major version, particularly since its duplicating existing PHPUnit functionality. I've made this change more explicit in the upgrading docs though.

chillu:
WTF I really don't know why the tests are failing here. framework `master` has a `branch-alias` of `4.0.x-dev` which should in theory kick in for this pull request branch. Or did something change in composer which means we need to mention that alias specifically in the root `composer.json`?

tractorcow:
Maybe there's something wrong with aliases on master branches. I'venoticed some issues on other modules (but it's just a hunch). Will look at it again later.

tractorcow:
Sorry I just remembered... @dhensby is merging up fixes for travis up to 3.2 branch.

Will need to merge this in manually if you want composer to work. https://github.com/tractorcow/silverstripe-framework/commit/2036f7e6c21a024d51ad0e0b2ae19aba25ea1e61

dhensby:
Yer, travis is really having a bad time :/

chillu:
What does container based travis have to do with branch aliases resolving in composer? Any objections against merging 3.2 into 3, then into master in order to get the travis fix ported from 3.1?

dhensby:
@chillu it's not a branch alias resolving problem that I can tell?

I've merged up from 3.1 -> 3.2 and 3.2 -> 3. I'm not sure that merging into master will fix this problem (a rebase might).

chillu:
Ah right, now Travis is failing for a different reason. On the last run it was complaining about an incompatible version of sqlite. I've just rebased on master again, let's see what this does. On a related note: Does anybody know why framework `master` has the `3` branch merged in, but cms `master` doesn't? I've just tried to merge, but it stuffs up the CMS UI (tabs are missing).

dhensby:
I didn't merge into master in framework or cms today :/ so I don't know.

dhensby:
rebase?

chillu:
OK, tests are passing - apart from Behat which relies on this PR being merged: https://github.com/silverstripe-labs/silverstripe-behat-extension/pull/95

dhensby:
I can't merge in that repo :/

tractorcow:
I've merged the other PR and restarted travis.

dhensby:
@chillu needs rebasing, please

chillu:
@dhensby Done!

dhensby:
Wow.. Still no green button :S

chillu:
@dhensby Rebased, and confirmed tests pass. Wanna merge?

tests/model/GDImageTest.php
```diff
@@ -3,17 +3,14 @@
 class GDImageTest extends ImageTest {
 
 	public function setUp() {
-		$skip = !extension_loaded("gd");
-		if($skip) {
-			$this->skipTest = true;
+		if(!extension_loaded("gd")) {
+			$this->markTestSkipped("The GD extension is required");
```
tractorcow: markTestSkipped throws an exception; The below code doesn't even run.

tractorcow: just put the skip after parent::setup?

tests/model/ImagickImageTest.php
```diff
@@ -1,9 +1,10 @@
 <?php
 class ImagickImageTest extends ImageTest {
 	public function setUp() {
-		$skip = !extension_loaded("imagick");
-		if($skip) {
-			$this->skipTest = true;
+		if(!extension_loaded("imagick")) {
+			$this->markTestSkipped("The Imagick extension is not available.");
+			parent::setUp();
```
tractorcow: Same as above.

tractorcow:
I noticed an issue which would lead to incorrect nesting... (parent::tearDown would get called more often than parent::setup in case of skipped tests).

dhensby:
@tractorcow does that mean we can't merge this

tractorcow:
We can merge it now.

dhensby:
Any other large deletions we can make?

tractorcow:
Mailer?

tractorcow:
RestfulService

PR #4405, opened 2015-07-16
FIX: DataQuery uses an undefined variable $callerClass in user_error() message

christopherdarling:
replaced with $this->dataClass


dhensby:
thanks

PR #4358, opened 2015-07-01
Pr/2978

tractorcow:
Fixup for https://github.com/silverstripe/silverstripe-framework/pull/2978

PR #4407, opened 2015-07-16
HTML error message in ChangePasswordForm

lekoala:
Use third argument in sessionMessage to avoid escaping br in the error message


kinglozzer:
Fixed in #4358

PR #4410, opened 2015-07-16
API Force resampling by default

jonom:
Simplified force resampling by serving up a resampled image throught the getURL() function, but only if the resampled image has a smaller file size than the original.

@tractorcow @stevie-mayhew following up from https://github.com/silverstripe/silverstripe-framework/pull/4229

### DIFF
--- a/model/Image.php
+++ b/model/Image.php
@@ -70,7 +70,7 @@ class Image extends File implements Flushable {
 	 * @config
 	 * @var bool Force all images to resample in all cases
 	 */
-	private static $force_resample = false;
+	private static $force_resample = true;
 
 	/**
 	 * @config
@@ -163,7 +163,34 @@ class Image extends File implements Flushable {
 	public function forTemplate() {
 		return $this->getTag();
 	}
+	
+	/**
+	 * Gets the source image URL for this resource
+	 * 
+	 * @return string
+	 */
+	public function getSourceURL() {
+		return parent::getURL();
+	}
 
+	/**
+	 * Gets the relative URL accessible through the web. If forced resampling is enabled 
+	 * the URL will point to an optimised file, if it is smaller than the original
+	 *
+	 * @uses Director::baseURL()
+	 * @return string
+	 */
+	public function getURL() {
+		if ($this->config()->force_resample) {
+			//return $resampled->getURL();
+			$resampled = $this->getFormattedImage('Resampled');
+			if ($resampled->getAbsoluteSize() < $this->getAbsoluteSize()) {
+				return $resampled->getURL();
+			}
+		}
+		return $this->getSourceURL();
+	}
+	
 	/**
 	 * File names are filtered through {@link FileNameFilter}, see class documentation
 	 * on how to influence this behaviour.
@@ -232,10 +259,10 @@ class Image extends File implements Flushable {
 		
 		if( $widthRatio < $heightRatio ) {
 			// Target is higher aspect ratio than image, so check width
-			if($this->isWidth($width) && !Config::inst()->get('Image', 'force_resample')) return $this;
+			if($this->isWidth($width)) return $this;
 		} else {
 			// Target is wider or same aspect ratio as image, so check height
-			if($this->isHeight($height) && !Config::inst()->get('Image', 'force_resample')) return $this;
+			if($this->isHeight($height)) return $this;
 		}
 
 		// Item must be regenerated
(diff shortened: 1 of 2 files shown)


model/Image.php
```diff
@@ -1028,7 +1075,11 @@ public function __construct($filename = null, $isSingleton = false) {
 		$this->ID = -1;
 		$this->Filename = $filename;
 	}
-
+	
+	public function getURL() {
```
stevie-mayhew: Duplicate function?

model/Image.php
```diff
@@ -163,8 +163,35 @@ public function getTag() {
...
+	 * the URL will point to an optimised file, if it is smaller than the original
+	 *
+	 * @uses Director::baseURL()
+	 * @return string
+	 */
+	public function getURL() {
+		if ($this->config()->force_resample) {
+			//return $resampled->getURL();
+			$resampled = $this->getFormattedImage('Resampled');
+			if ($resampled->getAbsoluteSize() < $this->getAbsoluteSize()) {
```
stevie-mayhew: I know this is a bit stupid, but what if you wanted to resample something from `100px by 100px` up to `700px by 88px` - the resultant file would be larger and also the one you want.

Yes, you should be setting widths in CSS or somewhere to target that but you could also not. If somethings being resampled then it should always be returned as the resampled one I think?

jonom: The size check here is filesystem size (bits) rather than resolution/pixels - force_resampling only kicks in if you're requesting an image at its original dimensions - so should be all good?

model/Image.php
jonom: Nope cos it's in Image_Cached ;) hard to tell here

stevie-mayhew: We should split these sorts of things out into their own files! :smile:

model/Image.php
stevie-mayhew: Yeah I guess that I still expect a resampled image though, not the original? Not saying that its a bad thing, ideally what ever is served should be the smallest option - I just think that the expectation here is for a resampled image to be returned if its resampling ;)

jonom: Yeah I see what you mean... I think this is a good solution but the terminology might be misleading. I guess the setting could be called 'optimize_images' or something instead of 'force_resample'? Would mean a less direct upgrade path though.

tests/model/ImageTest.php
```diff
@@ -69,29 +69,44 @@ public function tearDown() {
 	}
 
 	public function testGetTagWithTitle() {
+		$origForceResample = Config::inst()->get('Image', 'force_resample');
```
dhensby: You don't need to do this in tests anymore. `Config` and `Injector` are sand-boxed for each test

tests/model/ImageTest.php
```diff
@@ -168,55 +183,21 @@ public function testReluctanceToResampling() {
...
-		$imageFit = $image->Fit(300, 600);
-		$this->assertTrue($imageFit->isSize(300, 300));
-		$this->assertNotEquals($image->Filename, $imageFit->Filename);
-		$imageFit = $image->Fit(600, 300);
-		$this->assertTrue($imageFit->isSize(300, 300));
-		$this->assertNotEquals($image->Filename, $imageFit->Filename);
-		$imageFit = $image->Fit(300, 300);
-		$this->assertTrue($imageFit->isSize(300, 300));
-		$this->assertNotEquals($image->Filename, $imageFit->Filename);
+		$this->assertTrue($image->Resampled()->getAbsoluteSize() < $image->getAbsoluteSize());
```
dhensby: `assertLessThan`

tests/model/ImageTest.php
```diff
@@ -168,55 +183,21 @@ public function testReluctanceToResampling() {
...
+		$this->assertTrue(strpos($image->getURL(), '_resampled') !== false);
```
dhensby: `assertContains`

tests/model/ImageTest.php
```diff
@@ -168,55 +183,21 @@ public function testReluctanceToResampling() {
...
+		
+		// Test original file is served when force_resample = false
+		Config::inst()->update('Image', 'force_resample', false);
+		$this->assertTrue(strpos($image->getURL(), '_resampled') === false);
```
dhensby: `assertNotContains`

model/Image.php
tractorcow: When we do namespacing in 4.0 we'll be doing this anyway. :D

tractorcow:
@dhensby will you merge up to master soon? That will fix these tests.

tests/model/ImageTest.php
jonom: Sweet.

jonom:
@dhensby I changed those tests quite a bit if you want to take another look

tractorcow:
One of the builds broke so I restarted it.

PR #4411, opened 2015-07-17
BUGFIX: Ampersands in Form data cut off

lingo:
When ampersands are entered in the form data, all data after the ampersand is cut off.
Here's the scenario:

With a site I'm working on, a form is created which later saves its data to several different objects.

The fields are created along these lines:

``` php
        $smokerField = OptionsetField::create('Customer[Smoke]', _t('Quote.SmokerLabel', 'Êtes-vous fumeur(ese)'), array(
            'no'  => _t('Quote.db_Smoke_No', 'Non'),
            'yes' => _t('Quote.db_Smoke_Yes', 'Oui')
        ));
```

Now, when the data is submitted, various things are processed and the data is saved roughly like this:

``` php
        $form->saveInto($customer, array_keys($data['Customer']));
```

In this situation the bug occurs due to the handling of square brackets by `saveInto`.

``` php
// framework/forms/Form.php
 . . .
                // If field is in array-notation we need to access nested data
                else if(strpos($name,'[')) {
                    // First encode data using PHP's method of converting nested arrays to form data
                    $flatData = urldecode(http_build_query($data));
                    // Then pull the value out from that flattened string
                    preg_match('/' . addcslashes($name,'[]') . '=([^&]*)/', $flatData, $matches);

                    if (isset($matches[1])) {
                        $exists = true;
                        $val = $matches[1];
                    }
. . .
```

This is because `urldecode` is being called, which is decoding not only `&s` between fields, but also within them.

### DIFF
--- a/forms/Form.php
+++ b/forms/Form.php
@@ -1415,9 +1415,10 @@ class Form extends RequestHandler {
 				// If field is in array-notation we need to access nested data
 				else if(strpos($name,'[')) {
 					// First encode data using PHP's method of converting nested arrays to form data
-					$flatData = urldecode(http_build_query($data));
+					$flatData    = http_build_query($data);
+					$encodedName = urlencode($name);
 					// Then pull the value out from that flattened string
-					preg_match('/' . addcslashes($name,'[]') . '=([^&]*)/', $flatData, $matches);
+					preg_match('/' . $encodedName . '=([^&]*)/', $flatData, $matches);
 
 					if (isset($matches[1])) {
 						$exists = true;


lingo:
I'm looking into the failed builds now...

lingo:
Hmm, looks like builds failed for technical reasons that aren't related to tests?

forms/Form.php
```diff
@@ -1415,9 +1415,10 @@ public function loadDataFrom($data, $mergeStrategy = 0, $fieldList = null) {
 				// If field is in array-notation we need to access nested data
 				else if(strpos($name,'[')) {
 					// First encode data using PHP's method of converting nested arrays to form data
-					$flatData = urldecode(http_build_query($data));
+					$flatData    = http_build_query($data);
+					$encodedName = urlencode($name);
```
chillu: Hey Luke, great to hear from you! Stylistically, I think declaring a one-off variable here is adding noise rather than clarity.

chillu:
I'm a bit confused. The pasted core core and your change is around `loadDataFrom()`, not `saveInto()`. Can you write a unit test for this case please? And maybe an example of form data before and after being processed, both with and without your fix?

lingo:
Hi Ingo I hope you're well.  I meant to say `loadDataFrom`, yes!
I've not been able to create a test for this yet, but will do so when I get a moment (silverstripe unit tests run terrrrrrrribly slowly on my machine, sadly, and the codebase I'm using doesn't work with sqlite).

dhensby:
@lingo you can run just one test class or set of tests by doing: `phpunit framework/tests/forms/FormTest`

lingo:
@dhensby Thanks, I usually run single tests as needed, but even running a single unit test seems to take upwards of 5 minutes.  Not exactly encouraging TDD!  I think the speed problems come from the fixture loading, but I've not been able to speed it up.

chillu:
@lingo Any progress on those unit tests? You can also run a single method with the `--filter` argument.

sminnee:
*ping* - a unit test would be handy, @lingo!

lingo:
*Ping* was needed, thanks Sam!  I'll try to find some time this week to add the unit tests.

sminnee:
Now worries @lingo, I've been guilty of it myself. ;-)

dhensby:
It would be good to rebase this just to get the latest test suite / CI running

forms/Form.php
```diff
@@ -1415,9 +1415,10 @@ public function loadDataFrom($data, $mergeStrategy = 0, $fieldList = null) {
...
 					// Then pull the value out from that flattened string
-					preg_match('/' . addcslashes($name,'[]') . '=([^&]*)/', $flatData, $matches);
+					preg_match('/' . $encodedName . '=([^&]*)/', $flatData, $matches);
```
tractorcow: Can we use preg_quote here? I believe that was the intention of addcslashes.

tractorcow: `preg_match('/\b' . preg_quote($encodedName, '/') . '=([^&]*)/', $flatData, $matches);`

dhensby:
bump

quamsta:
Related to: https://github.com/silverstripe/silverstripe-framework/issues/5277

Also this still seems to be happening with ManyMany['xx'] titled fields.

PR #4412, opened 2015-07-17
Ensure proper exit code for exceptions & errors (fixes #4336)

### LINKED ISSUE #4336: dev/build in CLI doesn't always exit with proper exitcode

sminnee:


sminnee:
Please ignore the appveyor failure; it's not yet relevant.

PR #4414, opened 2015-07-20
API: Lazy load template parser

kinglozzer:
Rationale: loading `SSTemplateParser.php` is one of the most CPU and memory intensive operations that we do, yet we cache the template files that it generates - so for most requests, it’s not required _at all_.

Benchmarked on PHP 5.5.24:
screen shot 2015-07-20 at 21 02 16
screen shot 2015-07-20 at 21 06 57

Yes, that memory saving is just from not loading one file!

Marked as API and PR-ed against master as a precaution really, but I guess this could possibly be a candidate for 3 if accepted?

### DIFF
--- a/view/SSViewer.php
+++ b/view/SSViewer.php
@@ -785,7 +785,9 @@ class SSViewer implements Flushable {
 	 *  </code>
 	 */
 	public function __construct($templateList, TemplateParser $parser = null) {
-		$this->setParser($parser ?: Injector::inst()->get('SSTemplateParser'));
+		if ($parser) {
+			$this->setParser($parser);
+		}
 
 		if(!is_array($templateList) && substr((string) $templateList,-3) == '.ss') {
 			$this->chosenTemplates['main'] = $templateList;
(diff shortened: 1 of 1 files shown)


dhensby:
I'd definitely accept this into 3.

Looking at the `SSTemplateParser` file, it'll need some element of re-factoring before 4.0 is released so this may never get out in the wild otherwise

chillu:
Wow, fantastic result Loz!

view/SSViewer.php
```diff
@@ -829,6 +831,9 @@ public function setParser(TemplateParser $parser)
 	 */
 	public function getParser()
 	{
+		if (!$this->parser) {
+			$this->parser = Injector::inst()->get('SSTemplateParser');
```
willmorgan: `setParser`, perhaps

willmorgan:
3 would be nice.

view/SSViewer.php
dhensby: good spot

kinglozzer: Oo yes, good point

kinglozzer:
Great, thanks for the feedback, will reopen against 3 :)

PR #4415, opened 2015-07-21
Lazy load template parser to save memory

kinglozzer:
See #4414 for context

### DIFF
--- a/view/SSViewer.php
+++ b/view/SSViewer.php
@@ -786,7 +786,9 @@ class SSViewer implements Flushable {
 	 *  </code>
 	 */
 	public function __construct($templateList, TemplateParser $parser = null) {
-		$this->setParser($parser ?: Injector::inst()->get('SSTemplateParser'));
+		if ($parser) {
+			$this->setParser($parser);
+		}
 
 		if(!is_array($templateList) && substr((string) $templateList,-3) == '.ss') {
 			$this->chosenTemplates['main'] = $templateList;
@@ -830,6 +832,9 @@ class SSViewer implements Flushable {
 	 */
 	public function getParser()
 	{
+		if (!$this->parser) {
+			$this->setParser(Injector::inst()->get('SSTemplateParser'));
+		}
 		return $this->parser;
 	}
 
@@ -1106,9 +1111,11 @@ class SSViewer implements Flushable {
 		// through $Content and $Layout placeholders.
 		foreach(array('Content', 'Layout') as $subtemplate) {
 			if(isset($this->chosenTemplates[$subtemplate])) {
-				$subtemplateViewer = new SSViewer($this->chosenTemplates[$subtemplate], $this->parser);
+				$subtemplateViewer = clone $this;
+				// Disable requirements - this will be handled by the parent template
 				$subtemplateViewer->includeRequirements(false);
-				$subtemplateViewer->setPartialCacheStore($this->getPartialCacheStore());
+				// The subtemplate is the only file we want to process, so set it as the "main" template file
+				$subtemplateViewer->chosenTemplates = array('main' => $this->chosenTemplates[$subtemplate]);
 
 				$underlay[$subtemplate] = $subtemplateViewer->process($item, $arguments);
 			}
@@ -1174,7 +1181,7 @@ class SSViewer implements Flushable {
 	}
 
 	public function parseTemplateContent($content, $template="") {
-		return $this->parser->compileString(
+		return $this->getParser()->compileString(
 			$content,
 			$template,
 			Director::isDev() && Config::inst()->get('SSViewer', 'source_file_comments')
@@ -1243,7 +1250,10 @@ class SSViewer_FromString extends SSViewer {
 	protected $cacheTemplate;
 	
 	public function __construct($content, TemplateParser $parser = null) {
-		$this->setParser($parser ?: Injector::inst()->get('SSTemplateParser'));
+		if ($parser) {
+			$this->setParser($parser);
+		}
+
 		$this->content = $content;
 	}
 


dhensby:
I like this. So is the slowdown and memory just from parsing the file or is it an instantiation that is expensive? 

If it's the constructor, can't this be optimised to be lazier?

kinglozzer:
It’s mainly from the file being loaded/compiled, the actual instantiation of `SSTemplateParser` isn’t very expensive at all. Having it loaded in `SSViewer::__construct()` triggers the autoloader, which then triggers the compilation of the file.

What’s bugging me now I look at this again is that this should be mitigated by opcode caching... let me do some more digging and make sure this isn’t just an environment thing :)

kinglozzer:
Guess who didn’t have opcache enabled?

shm2

It does still save some memory, just not anywhere near as much as I first thought (with an opcode cache enabled, it saves ~100kb). That sucks :disappointed:.

The original saving is still there for people on 5.3/5.4 without an opcode cache installed (or idiots like me who have forget they’ve disabled it :wink:), so I’ll leave this open.

dhensby:
I think it's worth having in anyway, it's still a performance improvement.

But i don't like the one line where we directly access the property instead of the getter.

I'd rather pass through the current object and have the other class extract the parser if needed... 

There's no point having a getter if it's not used consistently.

willmorgan:
Shame

100kb memory saving and properly encapsulating the parser property is still worth a merge though.

> But i don't like the one line where we directly access the property instead of the getter.

Where's that?

view/SSViewer.php
```diff
@@ -1106,6 +1111,8 @@ public function process($item, $arguments = null, $inheritedScope = null) {
 		// through $Content and $Layout placeholders.
 		foreach(array('Content', 'Layout') as $subtemplate) {
 			if(isset($this->chosenTemplates[$subtemplate])) {
+				// NB: We intentionally avoid using ->getParser() here as it may trigger expensive 
+				// template parser instantiation that may not be necessary
 				$subtemplateViewer = new SSViewer($this->chosenTemplates[$subtemplate], $this->parser);
```
willmorgan: `$this->getParser()`

kinglozzer:
> But i don't like the one line where we directly access the property instead of the getter.

Thirded. I considered refactoring it to use `clone $this` instead, but then it’s becoming more of an API change (`$this` !== `SSViewer` necessarily, it could be a subclass) and we need to add a method to reset `$this->chosenTemplates`.

> I'd rather pass through the current object and have the other class extract the parser if needed...

Hm? I don’t follow

dhensby:
> > I'd rather pass through the current object and have the other class extract the parser if needed...
> Hm? I don’t follow

You know, do `$subtemplateViewer = new SSViewer($this->chosenTemplates[$subtemplate], *$this*);`

Then the `SSViewer` can chose to grab the "parent" parser if needed.

kinglozzer:
I’m not sure that will work. We’d have to alter the method signature (either to alter the type-hinting or add an extra argument). Even then we’d still be loading the parser prematurely:

```php
public function __construct($templateList, TemplateParser $parser = null, SSViewer $parentViewer = null) {
		if ($parser) {
			$this->setParser($parser);
		} else if ($parentViewer) {
			// This would trigger the autoloader
			$this->setParser($parentViewer->getParser());
		}
}
```

We could do this, but I don’t like it at all. We’re still changing the method signature of `SSViewer::__construct()`:
```php
public function __construct($templateList, TemplateParser $parser = null, SSViewer $parentViewer = null) {
	if ($parser) $this->setParser($parser);
	if ($parentViewer) $this->setParent($parentViewer);
}

public function getParser() {
	if (!$this->parser) {
		if ($parent = $this->getParent()) {
			$this->setParser($parent->getParser());
		} else {
			$this->setParser(Injector::inst()->get('SSTemplateParser'));
		}
	}

	return $this->parser;
}
```

I think we could actually use `clone $this` after all. The only case in which it could break backwards compatibility is if someone has subclassed `SSViewer`, altered some of its behaviour and expects it to use the _original_ behaviour for the “subtemplates”.

I’d argue that we should really be using `clone` there anyway - if you’ve subclassed `SSViewer`, you don’t want your main template to go through your subclass and the child templates to go through the parent class - you want everything to go through your subclass.

kinglozzer:
I’ve pushed up a second commit which uses `clone $this` for creating “subtemplate viewers”, thoughts?

view/SSViewer.php
```diff
@@ -1106,9 +1111,11 @@ public function process($item, $arguments = null, $inheritedScope = null) {
 		// through $Content and $Layout placeholders.
 		foreach(array('Content', 'Layout') as $subtemplate) {
 			if(isset($this->chosenTemplates[$subtemplate])) {
-				$subtemplateViewer = new SSViewer($this->chosenTemplates[$subtemplate], $this->parser);
+				$subtemplateViewer = clone $this;
+				// Disable requirements - this will be handled by the parent template
 				$subtemplateViewer->includeRequirements(false);
-				$subtemplateViewer->setPartialCacheStore($this->getPartialCacheStore());
+				// The subtemplate is the only file we want to process, so set it as the "main" template file
+				$subtemplateViewer->chosenTemplates = array('main' => $this->chosenTemplates[$subtemplate]);
```
kinglozzer: An alternative here would be to add a method to clear the chosen templates (`$this->resetTemplates()` or something) and then use `setTemplateFile('main', $this->chosenTemplates[$subtemplate])`. I can do that if that would be preferred.

dhensby:
> > I'd rather pass through the current object and have the other class extract the parser if needed...
> Hm? I don’t follow

Ok, don't worry, I think it'd be horrible the way I'm thinking about it.

Hmm, interesting, I'm not clear on why `clone` is more appropriate than instantiating a new one; is it just so an extended instance of `SSViewer` would cascade to the subtemplates?

kinglozzer:
> I'm not clear on why clone is more appropriate than instantiating a new one; is it just so an extended instance of SSViewer would cascade to the subtemplates?

It’s so that the parser is passed down to the clone without calling `getParser()`

dhensby:
> > I'm not clear on why clone is more appropriate than instantiating a new one; is it just so an extended instance of SSViewer would cascade to the subtemplates?
> It’s so that the parser is passed down to the clone without calling getParser()

OK!

dhensby:
I'm happy, lets squash this and get it in!!!

dhensby:
Good stuff

Issue #4417, opened 2015-07-21
[FORMS] Form calls ->removeValidation() on validator, but may not always be defined.


patricknelson:
On this line: https://github.com/silverstripe/silverstripe-framework/blame/3.2/forms/Form.php#L683

``` php
        // We have to remove validation, if the fields are not editable ;-)
        if($this->validator)
            $this->validator->removeValidation();
```

However, if you create your own implementation of `Validator` there's no contract (abstract definition) to inform you that you must also implement a `->removeValidation()` method on your validator. I'd suggest either defining this as a default method on the `Validator` abstract class or, better yet, set this up as an abstract method so that it will fail earlier so that results are more consistent and predictable (since you may not realize this will fail until it's too late and already in production).

Thankfully I haven't had taht issue just yet, but I didn't realize this method was needed until months after I ended up disabling all the fields on my `DataObject` :)

patricknelson:
FYI, this affects `3.1.x` and `3.2` from what I can tell.

sminnee:
It looks like this was fixed in #4542 which has been available since 4.0; closing

PR #4418, opened 2015-07-21
FIX for #4417: Ensuring ->removeValidation() is defined on instances of Validator.

patricknelson:
Quick fix for #4417. Went ahead and based this on `3.1`.

Might cause some issues for people who haven't defined this method and happen to also not actually need it, even though they're still technically at risk of a fatal error if that method is called in the future, should they ever have a validated form which is also not editable.

### DIFF
--- a/forms/Validator.php
+++ b/forms/Validator.php
@@ -4,7 +4,7 @@
  * This validation class handles all form and custom form validation through
  * the use of Required fields.
  *
- * Relies on javascript for client-side validation, and marking fields after serverside validation.
+ * Relies on javascript for client-side validation, and marking fields after server-side validation.
  *
  * Acts as a visitor to individual form fields.
  *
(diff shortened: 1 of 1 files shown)


dhensby:
This can't go in 3.1 as it is going to cause errors for anyone who has extended the base Validator but hasn't implemented that function. 
 
I think the fix for 3.1 should be to check the method exists on validator before calling it. Probably the same for 3. 

Then master can have this change

tractorcow:
Please re-open a new PR for this branch on master, ta!

dhensby:
Can a core member her this in master if we don't see this opened by the original author?

patricknelson:
Hello visitors from the future! New PR is here: https://github.com/silverstripe/silverstripe-framework/pull/4542

patricknelson:
I forgot to mention two things:

# More important: I typecast `->requireField()` since it appears it would error anyway if the `$data` variable weren't an array (at least with a warning). I think it makes sense to do this since you should never be passing anything by an array here.
# I added a bunch of PHPDoc too for clarity and type hinting.

PR #4420, opened 2015-07-22
harden addFilter method

spekulatius:
currently anything could be given to the method - causing the foreach to fatal fail if the given param is not iterable.

### DIFF
--- a/model/DataList.php
+++ b/model/DataList.php
@@ -372,8 +372,10 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 
 	/**
 	 * Return a new instance of the list with an added filter
+	 *
+	 * @param  array $filterArray
 	 */
-	public function addFilter($filterArray) {
+	public function addFilter(array $filterArray) {
 		$list = $this;
 
 		foreach($filterArray as $field => $value) {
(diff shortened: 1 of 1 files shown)


dhensby:
Hmm, this is an API change so would have to go in master.

I'm not sure what the benefit of restricting it solely to `array`s is. It really comes down to a coding convention and argument of duck-typing vs strict typing...

I, personally, don't think this is a worthwhile addition. If a dev sends in a non-iterable data type, then they are going to get an error under either circumstance and I'd prefer to give people the freedom to pass in something other than an array if they wanted.

sminnee:
Unfortunately there is no way to type hint against array or Iterator. You could put an assertion as the first line of the function instead?

Passing an Iterator is a little unusual but I'd see no reason to block it.

tractorcow:
I concur with @dhensby ; We don't really have any reason not to support iterators here, as well as this being a breaking API change.

If we wanted to force this type, we cannot use php typehints, but it can still document it properly in PHPDoc.

If you want to amend your pull request to only be phpdoc change then I'll merge it. Any actual api change will need to be in 4.0. We could do something like in http://stackoverflow.com/questions/3584700/iterable-objects-and-array-type-hinting

```php
function is_iterable($var) {
    return (is_array($var) || $var instanceof Traversable);
}
```

spekulatius:
Fair enough @dhensby and @tractorcow. Makes sense to not change the API between minor versions. I adjusted it to only the phpdoc block

PR #4421, opened 2015-07-22
Upgrade template translation syntax

kinglozzer:
Given that this “new” syntax has been in place [since 2012](https://github.com/silverstripe/silverstripe-framework/pull/342), I think it’s probably time we upgraded core templates to use it :)

### DIFF
--- a/admin/templates/Includes/ModelAdmin_ImportSpec.ss
+++ b/admin/templates/Includes/ModelAdmin_ImportSpec.ss
@@ -1,8 +1,8 @@
 <div class="importSpec" id="SpecFor{$ModelName}">
-	<a href="#SpecDetailsFor{$ModelName}" class="detailsLink"><% sprintf(_t('ModelAdmin_ImportSpec_ss.IMPORTSPECLINK', 'Show Specification for %s'),$ModelName) %></a>
+	<a href="#SpecDetailsFor{$ModelName}" class="detailsLink"><%t ModelAdmin_ImportSpec_ss.IMPORTSPECLINK 'Show Specification for %s' s=$ModelName %></a>
 	<div class="details" id="SpecDetailsFor{$ModelName}">
-	<h4><% sprintf(_t('ModelAdmin_ImportSpec_ss.IMPORTSPECTITLE', 'Specification for %s'),$ModelName) %></h4>
-		<h5><% _t('ModelAdmin_ImportSpec_ss.IMPORTSPECFIELDS', 'Database columns') %></h5>
+	<h4><%t ModelAdmin_ImportSpec_ss.IMPORTSPECTITLE 'Specification for %s' s=$ModelName %></h4>
+		<h5><%t ModelAdmin_ImportSpec_ss.IMPORTSPECFIELDS 'Database columns' %></h5>
 		<% loop $Fields %>
 		<dl>
 			<dt><em>$Name</em></dt>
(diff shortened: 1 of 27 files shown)


admin/templates/Includes/ModelAdmin_ImportSpec.ss
```diff
@@ -1,8 +1,8 @@
 <div class="importSpec" id="SpecFor{$ModelName}">
-	<a href="#SpecDetailsFor{$ModelName}" class="detailsLink"><% sprintf(_t('ModelAdmin_ImportSpec_ss.IMPORTSPECLINK', 'Show Specification for %s'),$ModelName) %></a>
+	<a href="#SpecDetailsFor{$ModelName}" class="detailsLink"><%t ModelAdmin_ImportSpec_ss.IMPORTSPECLINK 'Show Specification for %s' s=$ModelName %></a>
 	<div class="details" id="SpecDetailsFor{$ModelName}">
-	<h4><% sprintf(_t('ModelAdmin_ImportSpec_ss.IMPORTSPECTITLE', 'Specification for %s'),$ModelName) %></h4>
-		<h5><% _t('ModelAdmin_ImportSpec_ss.IMPORTSPECFIELDS', 'Database columns') %></h5>
+	<h4><%t ModelAdmin_ImportSpec_ss.IMPORTSPECTITLE 'Specification for %s' s=$ModelName %></h4>
```
kinglozzer: I’ve named the injection argument `s`, because it sort of matches `%s`. It doesn’t actually matter what the name is, `i18n::_t()` will replace them in order, just as long as it’s `name=$Value` (it won’t just accept `$Value` on its own).

Really, we should update the use of `%s` in _all_ entities to actual placeholders like `{modelname}`, but that’s a bigger change than this PR.

templates/email/ChangePasswordEmail.ss
```diff
@@ -1,11 +1,11 @@
-<p><% _t('ChangePasswordEmail_ss.HELLO', 'Hi') %> $FirstName,</p>
+<p><%t ChangePasswordEmail_ss.HELLO 'Hi' %> $FirstName,</p>
 
 <p>
-	<% _t('ChangePasswordEmail_ss.CHANGEPASSWORDTEXT1', 'You changed your password for', 'for a url') %> $AbsoluteBaseURL.<br />
-	<% _t('ChangePasswordEmail_ss.CHANGEPASSWORDTEXT2', 'You can now use the following credentials to log in:') %>
+	<%t ChangePasswordEmail_ss.CHANGEPASSWORDTEXT1 'You changed your password for' is 'for a url' %> $AbsoluteBaseURL.<br />
```
kinglozzer: `is 'context goes here'` is the way to pass context information through the new translation syntax. As far as I can tell, the context is never actually used by anything at all, but I figured it best to leave it in as a precaution

sminnee: The intended use of the context info is to pull it into Transifex to help translators.

tractorcow:
Hm, testing this locally (3 branch) reveals quite a few template parse errors.

```
Error at line 4019 of /framework/view/SSTemplateParser.php
UploadFieldTest::testConfigOverwriteWarningCannotRelaceFiles
Parse error in template on line 15. Error was: Malformed opening block tag t. Perhaps you have tried to use operators?
```

PR #4427, opened 2015-07-23
add async and defer attributes to js requirements

xini:
as discussed in #4194 : rebased to 3.2 and added some tests to make sure it all works.


sminnee:
I think we want to move away from a lot of arguments to use a $options map instead.

$options could be "media" for CSS and "async" or "defer" for Javascript.

Where there is already a string media argument, replace that with $options, and do a deprecated BC check:

    If(is_string($options))  {
        $options = array('media' => $options);
    }

tractorcow:
Sorry to be yet another pain, but if we are adding parameters to existing methods (in this case, Requirements_Backend) then we need to consider it an api breakage, since someone could have overridden it in their project.

can you please rebase against master?

chillu:
@xini Can you please rebase against master and change to `$options` as per Sam's suggestion?

xini:
no problem. see #4555.

Issue #4426, opened 2015-07-23
Use of $_FILE_TO_URL_MAPPING without the 'global' keyword in Constants.php


nicocin:
In framework/core/Constants.php the $_FILE_TO_URL_MAPPING global is used to update the $_SERVER variable for cli use.

It seems to me there should be a line like `global $_FILE_TO_URL_MAPPING;` before the first use of the variable (that is, before the isset check), as `isset($_FILE_TO_URL_MAPPING)` will always return false without it.

dhensby:
Have you checked if it's actually a problem? I'd be surprised if it is as it'd break CLI for anyone using it.

It's probably being pulled into scope by some file that includes Constants.php.

However, I agree it's not clear and should be declared global for clarity, PRs welcome :)

Firesphere:
We checked. (nicocin is my colleague). Adding the global declaration fixes errors on the file to url mapping bugging out. I'll create a PR in a second. I need to update my fork of the framework.

nicocin:
Daniel, thanks for your reply. 
I think you are right, it does not seem to create a problem with normal CLI use, because /core/Core.php is the first include from cli-script.php, and Core.php immediately includes the Constants.php.

Some background info:
It fails when the _ss_environment.php is required *before* the core/Core.php is required.

For me this happend in a specific situation:
I'm using https://github.com/silverstripe-labs/silverstripe-behat-extension for behat testing, and when I add the `@database-defaults` tag to a feature, an error occurs because of this issue.

This extension seems to 'initialize' the core using  `require_once $frameworkPath . '/core/Core.php';` after  first requiring the _ss_environment.php file. 
because Constants.php, uses require_once, the file is not required again and $_FILE_TO_URL_MAPPING is used while it is not in scope (any more).
Later on, `Director::protocolAndHost()` is called while generating some absolute URL, as the result of a `Director::test()` call in `ErrorPage::requireDefaultRecords()` and a user_error is thrown because `$_SERVER['HTTP_HOST']` is not set.

@Firesphere, thanks for creating the PR

sminnee:
Aha thanks for the explanation and PR, guys! All merged now.

Firesphere:
@nicocin Thanks for the explenation, since I didn't know all of it :) .

kinglozzer:
@sminnee did you mean to merge #4428?

sminnee:
@kinglozzer Yeah, I did—was there a problem? 

[...]

PR #4428, opened 2015-07-23
Declare the global mapping

Firesphere:
Fix for  #4426


Firesphere:
In addition, this doesn't break working CLI commands, but un-breaks a few non-working CLI calls, like Behat testing.

dhensby:
If this is a bug fix, it's safe to go into 3.1 IMO

sminnee:
If it were raised in 3.1 then I'd merge it but it in 3.2

tractorcow:
By the way this caused a regression. :) 

https://github.com/silverstripe-labs/silverstripe-travis-support/pull/19

dhensby:
Heh, I knew I should have been more cautious :P

PR #4430, opened 2015-07-26
Replace Zend log with Monolog; rely more on PSR-3

sminnee:
This change rips the guts out of our custom logging and instead relies much more on Monolog.

A few notes:

 * The coupling to Monolog is small, but non-zero. Most of the work is done by the Injector system, but the `MonologLoader` class has some plumbing in it to make it a bit less verbose to set up handlers. Arguably, this could be done by adding features to the Injector, but I chose not to dig too far into this.
 * It's still fairly verbose to configure common log options like log-to-file. If we wanted, we could add more support to MonologLoader to read more concise config values and configure the appropriate handlers.
 * Disabling display_errors in main.php is a bit weird but necessary to make this work. Of note: ErrorControlChain makes it impossible to do this within the regular set-up code.
 * The two handlers I have, for displaying detailed and friendly errors, might instead be better as formatters. We may also wish to consider whether it's best not to maintain our own implementations of those, although the default ones provided by Monolog are a bit dry.
 * If we replaced those with formatters, we'd need to think about what kinds of handlers would be best. We could just use:
  * A StreamHandler, pointing to STDOUT
  * A new handler that calls `echo` and `header()` as appropriate
  * A new handler that makes use of `SS_HTTPResponse` (to be useful, this would need to tied into Director somehow
 * In future, it would be good to pass the logger in as an instance variable, rather than relying on calls to `Injector::inst()->get('Logger')`.

It's satisfying to pull out so much unnecessary code!


tractorcow:
Wow, look at all that Zend that isn't there anymore.

tractorcow:
Have you considered how we could potentially use a DI error logger / handler as a replacement for the existing errorPage handling? I would suggest looking at both of these features in the context of one another to see if they could be a bit more cooperative; At the moment (on live) it's necessary for ContentController::httpError to be invoked explicitly for error page generation.

It fits well into the monolog api, as it could be a separate handler that is injected independently of others in the list without needing any cms-specific code baked into framework. With the error context available, it could be triggered to only respond to certain errors.

sminnee:
I agree decoupling would be good, something like this would probably work:

 * DisplayFriendlyErrorHandler should be turned into a formatter, and broken into two classes: `FriendlyDebugViewFormatter` and `FriendlyErrorPageFormatter`. The former would use DebugView, the latter would be bundled with the CMS and would make use of ErrorPage. 
 * The CMS would come with some yaml config to swap out the formatter of the relevant handler.

I don't know if I'd dig deeper into refactoring error page generation as part of this; ErrorPage wouldn't benefit from being couple directly to monolog handlers itself IMO.

dev/logging/MonologLoader.php
```diff
@@ -0,0 +1,56 @@
...
+ * Adaptor connecting Monolog to SilverStripe for error handling
+ */
+class MonologLoader
+{
+	function __invoke() {
+		// Register the default error handler
+		$logger = \Injector::inst()->get('Logger');
+		ErrorHandler::register($logger);
+
+		// Load the error handlers into the monolog loader
```
tractorcow: I would push all of the below into a factory, and then the rest of this class could be simplified to error handler registration.

assertchris:
Looks good! I would love to see even less coupling to Monolog. Ideally it should only be referenced at the `Injector` level but a good middle-ground is to put all `Monolog` references behind a factory (similarly to how the recent Solr stuff has gone). 

Also, though I love the use of namespaces, I'm not sure about the `Dev\` part should exist anymore. `SilverStripe\Framework\Logging` seems good, and I'm sure we could find reasonable prefixes for all the other things in there. `Dev` is a lazy shortcut, historical though it may be... :)

sminnee:
The reason I haven't gone with referencing Monolog only in the Injector is that the Injector config gets really verbose and I'm worried that we're making things hard for users. The factory seemed like a reasonable trade-off. I'm keen to get feedback on that, as it's a judgement call.

The namespace change makes sense. Will do, and I'll shuffle the folder structure to match.

tractorcow:
Put all the bits in config, and just have a (very simple) factory put the bits together. At the least I'd see it as pulling in a DI logger, and adding all handlers defined in another yml config. Perhaps you could also apply some kind of sorting (since the order of handlers is quite important) so that third parties can control the order of these.

sminnee:
@tractorcow I think you're describing the current implementation?

dev/logging/MonologLoader.php
sminnee: MonologLoader was intended to be a factory. Perhaps making it callable is a problem?

sminnee: I don't think it makes sense to have a separate class whose sole responsibility is to call ErrorHandler::register($logger).

tractorcow:
The problem with yml is that it's hard to merge configs exactly the way you want :D you can use before / after but in some cases it's better to just have a priority.

E.g.

```yml
Injector:
  NoticeHandler:
    handlers:
      class: SilverStripe\Framework\Dev\DisplayDetailedErrorHandler
      constructor:
        - "notice"
  ErrorHandler:
    handlers:
      class: SilverStripe\Framework\Dev\DisplayFriendlyErrorHandler
      constructor:
        - "error"
Loggers:
  handlers:
    100: NoticeHandler
    200: ErrorHandler
```

and

```yml
Loggers:
  handlers:
    150: MyCustomHandler
```

assertchris:
@tractorcow Kinda don't like that approach. Like a lot.

sminnee:
We need to balance architectural cleanliness with developer experience.

I've had a bit of a play and refactoring the MonologLoader into an Injector-compatible factory works nicely; the yml config remains pretty much the same (just add a factory property to the definition of Injector.Logger) but the code has more of a clear single-responsibility. I'm not yet sure where to put the call to `ErrorHander::register()` but I'll figure something out.

I dislike having to inject individual objects and then assemble them in a separate list. It's verbose & fiddly. I've done it a few times when working with existing code—it's quite flexible for "making something work" and so I guess is tolerable for unplanned edge-cases—but I don't think it's acceptable for something as common as logging to a file.

The underlying issue is that Injector lacks a few expressive capabilities that would be helpful. I think that should be addressed, but it's a separate issue and shouldn't block merge of this change.

sminnee:
On priority numbers: we already have "before" and "after". I don't see any reason to introduce a 2nd completely different mechanism for specifying priority in this one case. We'd need a pretty solid argument to justify it.

The ordering of handlers also doesn't seem particularly critical, as the default behaviour is that they are all called: we don't block any bubbling in our default config, and for any given message our default config only triggers a single handler, so it's unlikely you'll need to make a config that inserts a handler between those two.

There is a notable limitation: I can't currently see a good way of *removing* the default handlers. I've got to figure out how I can make the config system allow for this... ;-) @hafriedlander?

tractorcow:
Well, you could look at the way that RequestFilter does it 

```yaml
---
Name: requestprocessors
---
Injector:
  RequestProcessor:
    properties:
      filters:
        - '%$FlushRequestFilter'
        - '%$VersionedRequestFilter'
```

sminnee:
OK, I've pushed up an update, keen to get feedback.

Note the mechanism for disabling the default error handler. It's a bit ganky, but config doesn't currently provide a better mechanism. Hamish reckons that this is only due to 2.4-bc-support and an alternative handling of null values in config could do away with the need for a special check on class == none/disable.

I've preserved the special treatment of the handlers setting by the MonologFactory; this seems sufficiently orthodox and better for users than the 2-part approach taken in request filter, which is well into yaml-duck-programming land :-/

logging/MonologFactory.php
```diff
@@ -0,0 +1,55 @@
...
+		$handlerID = null;
+		$handlers = \Config::inst()->get('Logger', 'handlers');
+		foreach($handlers as $handler) {
+			// A hacky mechanism for disabling handlers in yaml config
+			if(in_array(strtolower($handler['class']), array('none', 'disable'))) {
+				continue;
+			}
+
+			// Special treatment for Formatter
+			if(isset($handler['properties']['Formatter']['class'])) {
```
tractorcow: Why not just make the handlers a dependency of MonologFactory itself, like how filters is a dependency of RequestProcessor?

```

/** DI this */
protected $handlers = array();

public function create($class, array $params = array()) {
		// Register the default error handler
		$logger = parent::create($class, $params);
		foreach($this->getHandlers() as $handler) {
			$logger->pushHandler($handler);
		}
		return $logger;
	}
```

sminnee: >Why not just make the handlers a dependency of MonologFactory itself.

My main concern with that is that it makes it quite arbitrary where you are supposed to insert the configuration. You need to be aware of the MonologFactory and how it works. It starts getting quite verbose, as well.

    Injector:
      SilverStripe\Framework\Logging\MonologFactory:
        properties:
          handlers:
            Mail:
              class: Monolog\Handler\NativeMailerHandler
              ...

Maybe it would be better if it were injected into the Logger instead?

    Injector:
      Logger:
        properties:
          handlers:
            Mail:
              class: Monolog\Handler\NativeMailerHandler
              ...

It would be good to get a broader set of views as to the direction we should go on this. It's pretty straightforward to implement whichever approach we want; it's a trade-off between configuration brevity and minimising the amount of custom instantiation code we have.

My preference would be to go with an approach that we can consistently use and is reasonably enjoyable and clear for developers to use—I don't want people to have to learn "programming in Injector".

> I don't see why you can't just use Config api if you need a finer degree of control in your app

I don't really know what you mean by this; it's more helpful to give specific examples. The use-case that I had in mind (disabling the default handler) wasn't well catered for by the config API. Moving from an array to a map helps (because you can name the items that you want to disable) but there's a change to the config API necessary to make it work without my `if class == "none" or "disabled"` check. It sounds like this would be a reasonable ss4 change.

stevie-mayhew:
Have you had a look at https://github.com/camspiers/silverstripe-loggerbridge and https://github.com/heyday/silverstripe-backstop @sminnee ?

We've been using loggerbridge for a long time now and end up writing a configuration like the below for each project - I don't think that its too verbose as the base use is actually really simple.


```yml
---
Only:
  environment: dev
---
Injector:
  Monolog:
    class: Monolog\Logger
    constructor:
      0: App

---
Only:
  environment: test
---
Injector:
  Monolog:
    class: Monolog\Logger
    constructor:
      0: App

---
Only:
  environment: live
---
Injector:
  Raven:
    class: CustomRavenClient
    constructor:
      0: #SENTRYDSN#
  Monolog:
    class: Monolog\Logger
    constructor:
      0: App
      1:
        - '%$RavenHandler'
        - '%$NativeMailerHandler'
      2:
        - '%$MemoryPeakUsageProcessor'

---
Name: logging
---
Injector:
  Backstop:
    class: Heyday\Backstop\Backstop
    constructor:
      0: '%$Monolog'
      1:
        404: info
        4??: error
        5??: critical

  LoggerBridge:
    class: 'Camspiers\LoggerBridge\LoggerBridge'
    constructor:
      0: '%$Monolog'
    properties:
      ReportBacktrace: true

  RavenHandler:
    class: Monolog\Handler\RavenHandler
    constructor:
      0: '%$Raven'

  ChromePHPHandler:
    class: Monolog\Handler\ChromePHPHandler

  FirePHPHandler:
    class: Monolog\Handler\FirePHPHandler

  NativeMailerHandler:
    class: Monolog\Handler\NativeMailerHandler
    constructor:
      0:
        - 'webmaster@littlegiant.co.nz'
      1: 'Critical error on #PLATFORM#'
      2: webmaster@littlegiant.co.nz
      3: 500

  MemoryPeakUsageProcessor:
    class: Monolog\Processor\MemoryPeakUsageProcessor
```

sminnee:
@stevie-mayhew thanks for that! It's good to see how Monolog is already being used, but 80 lines of configuration doesn't really scream "simple" to me... ;-)

I see that it uses the 2 step approach for configuring loggers: define each handler as a service, and then reference the services in the Monolog constructor. This seems quite verbose, and would get particularly unintuitive if you were tweaking the only handlers passed to Monolog constructor, e.g. adding one more handler.

Do you think it's a better approach than inlining the handler service definitions into a handlers array, as I have done?

The big difference between those module and this pull request is that the framework will come with something out of the box, so it's going to be more important to give people the ability to modify the default implementation, whereas these modules assume you are building a logging configuration yourself from scratch.

Obviously, since this PR makes everything pluggable, you could keep using the configuration you've provided, with a few small tweaks:

 * Rename `Injector.Monolog` to `Injector.Logger`.
 * Rename `%$Monolog` to `%$Logger` in the backstop service definition.
 * Remove `Injector.LoggerBridge` as core already does this.

tractorcow:
Actually I think that @stevie-mayhew has hit the nail on the head. See this part of the code

```yaml
constructor:
      0: App
      1:
        - '%$RavenHandler'
        - '%$NativeMailerHandler'
```

Notice how he's not defining the actual dependencies there; All that he's doing is declaring the link between this particular logger and the handlers, which are defined separately. In fact the way he's implemented it negates the need to even have a factory.

The value here is that the injector is creating all the objects without having to manually loop through a list and inject them one at a time.

I think its better than inlining the definition, because as soon as you have multiple loggers, especially where they might use the same handlers, you have to duplicate that definition in each.

@sminnee I agree with your tweaks below; We should name them as generic as possible, since monolog is only one implementation.

Ok, here's another solution to the issue of factories obscuring where the config goes; Let's just subclass MonoLog with a wrapper that works better with our DI pattern. As @sminnee suggested, lets just make it that you inject these onto the logger itself.

```php
/** Allows handlers to be DI into a logger */
class SilverStripeLogger extends \Monolog\Logger {
    public setHandlers($handlers) {
      $this->handlers = array_values($handlers);
    }
}
```

and then our config becomes

```
Injector:
  ErrorLogger:
    class: SilverStripeLogger
    properties:
      handlers:
        RavenHandler: '%$RavenHandler'
        MailHandler: '%$NativeMailerHandler'
```

Where each handler definition is as per @stevie-mayhew's example. I do seriously suggest keeping the separation of handler and logger definitions separate to keep each definition simple (more and smaller instead of fewer and larger).

I've tested this locally, so you can define your array of handlers as an associative array (which is the format you preferred, right @sminnee? ) but we simply flatten this to a normal array when saved into the `Logger` instance.

sminnee:
OK, adding `Monolog\Logger::setHandlers` certainly makes things simpler, although I'm loathe to subclass it. A PR to the Monolog project would be nicer.

tractorcow:
Hm, then let's do that? :P I can't see the harm in trying!

sminnee:
OK, adding `Monolog\Logger::setHandlers` certainly makes things simpler, although I'm loathe to subclass it. A PR to the Monolog project would be nicer; I'll try that.

tractorcow:
They might not let us do `array_filter` or `array_values` though, if we were to add this into their core.

Is that a problem we can solve with injector / config?

sminnee:
I'm still a little lukewarm on relying on the service definition code as-is, as it gets kind of verbose once you incorporate formatters. Here's a sketch of the default configuration we'd be looking at:

    ---
    Name: logging
    ---
    Injector:
      ErrorHandler:
        class: SilverStripe\Framework\Logging\MonologErrorHandler
        properties:
          Logger: %$Logger
      Logger: 
        type: singleton
        factory: SilverStripe\Framework\Logging\MonologFactory
        class: Monolog\Logger
        constructor:
          - "error-log"

    ---
    Name: dev-logging
    Only:
      environment: dev
    ---
    Injector:
      Logger:
        properties:
          Handlers:
            DisplayErrorHandler: %$DisplayErrorHandler
      DisplayErrorHandler:
        class: SilverStripe\Framework\Logging\HTTPOutputHandler
        constructor:
          - "notice"
        properties:
          Formatter: %$DisplayErrorFormatter

      DisplayErrorFormatter:
        class: SilverStripe\Framework\Logging\DetailedErrorFormatter
    ---
    Name: live-logging
    Except:
      environment: dev
    ---
    Injector:
      Logger:
        properties:
          Handlers:
            DisplayErrorHandler: %$DisplayErrorHandler
      DisplayErrorHandler:
        class: SilverStripe\Framework\Logging\HTTPOutputHandler
        constructor:
          - "error"
        properties:
          Formatter: %$DisplayErrorFormatter

      DisplayErrorFormatter:
        class: SilverStripe\Framework\Logging\DebugViewFriendlyErrorFormatter

And here's an example of adding email errors:

    Injector:
      Logger:
        properties:
          Handlers:
            Mail:
              class: Monolog\Handler\NativeMailerHandler
              constructor:
                - sam@example.com
                - There was an error on your test site
                - sam@example.com
                - warning
              properties:
                ContentType: text/html
                Formatter: %$MailHandlerFormatter

      MailHandlerFormatter:
        class: SilverStripe\Framework\Logging\DetailedErrorFormatter

To be honest, if we can get the monolog patch acceptance, then my reticence about this configuration format probably doesn't justify extra custom code where we otherwise could just run straight with the Injector.

It does suggest a number of improvements to the config & injector system that could be made, however. For example, it would be nice to do this:

    Injector.Logger.properties.Handlers.Mail:
      class: Monolog\Handler\NativeMailerHandler
      constructor:
        - sam@silverstripe.com
        - There was an error on your test site
        - sam@silverstripe.com
        - warning
      properties:
        ContentType: text/html
        Formatter: %$MailHandlerFormatter    
    Injector.MailHandlerFormatter:
      class: SilverStripe\Framework\Logging\DetailedErrorFormatter

sminnee:
FYI I've posted a PR to monolog here:
https://github.com/Seldaek/monolog/pull/617

sminnee:
In the meantime, while that Monolog change is being reviewed, I think I'm going to try and refactor the factory so that it does exactly what the Monolog PR would do. I think this has less long-term impact than subclassing Monolog.

sminnee:
OK I'll be able to deal with this much more elegantly if we can implement #4442—what do people think of that?

stevie-mayhew:
Looks like you've come around to my way (well, @camspiers way) of thinking without me having to make much argument so thats good.

I really like the idea of having the DI containers be more flexible - we've previously worked around the SS DI by using (Pimple)[http://pimple.sensiolabs.org/] as well as the SS DI to be able to get the flexibility we needed.

sminnee:
I still don't particularly like the verbosity of the config, but I agree that it should probably be fixed in Injector, not Logger. ;-) It sounds like you agree with that, @stevie-mayhew?

stevie-mayhew:
Yes definitely agree with that!

We can agree to disagree on the verbosity issue, I'm always going to be a fan of exposing as much as possible :smiley: We can have a big discussion next time we cross paths @sminnee :wink: 

For the issue this PR addresses however, I think these steps are fantastic!

nyeholt:
Just a quick comment that I'm in favour of @stevie-mayhew's more verbose layout - but then I used to be a java developer so.... :). 

I _would_ like to improve some of the way properties are resolved so that, for example, a property could be another service definition that the injector would subsequently bootstrap, or retrieved from the environment, or x or y or z rather than the hardcoded scalar value or %$ lookup as current. I think that's partly what you're getting at with some of the config examples you've put up @sminnee  ?

sminnee:
@nyeholt yes, that's what I disliked. In the case of logger, I disliked having to create a separate Formatter service definition just to map it as a property of the handler service definition. It pollutes the global namespace, and increases the number of entities that a developer has to remember to pull together.

Let's shift the discussion of Injector improvements over to #4442, maybe?

sminnee:
OK I had a missing comma in my composer.json but that's fixed now.

From my perspective, this is ready to merge. I'd appreciate a final review (maybe from someone who hasn't been involved in the discussion to date?) and then I think we can get it merged! :-)

Thanks to everyone who has given input on this, especially @stevie-mayhew and @tractorcow.

camspiers:
I'm happy to have a look.

logging/HTTPOutputHandler.php
```diff
@@ -0,0 +1,66 @@
...
+	 * Set the HTTP status code to use when displaying this error.
+	 * Default 500
+	 */
+	public function setStatusCode($statusCode) {
+		$this->statusCode = $statusCode;
+	}
+
+	protected function write(array $record) {
+		// TODO: This coupling isn't ideal	
+		if(\Controller::has_curr()) {
```
camspiers: I've run into the same thing a number of times. We should really give this some thought. It would be good to have a github issue number here if we are going to leave a TODO.

composer.json
```diff
@@ -17,7 +17,8 @@
 	],
 	"require": {
 		"php": ">=5.3.3",
-		"composer/installers": "~1.0"
+		"composer/installers": "~1.0",
+		"monolog/monolog": "~1.15"
```
camspiers: Are we sure this is the right version requirement? Are you relying on features that were added between 1.0 and 1.15?

kinglozzer:
Just testing this locally, two minor issues I’ve spotted:

- Not catching an exception results in: `[Emergency] Exception message goes here`. You can obviously work it out with the message and available stack trace below, but I’d still prefer to see “Uncaught exception” somewhere in there too
- Calling `user_error('Oh no!', E_USER_ERROR)` results in two errors + stack traces. Not sure what’s going on there, only happens for `E_USER_ERROR`, seems like it’s triggering a non-fatal error followed by a fatal one.

I don’t think the config verbosity is an issue. Like you said above, we might be able to improve on that in Injector. Plus you’ve written excellent documentation to go with it anyway :wink:.

logging/DebugViewFriendlyErrorFormatter.php
```diff
@@ -0,0 +1,90 @@
...
+
+		} else {
+			// TODO: Refactor this into CMS
+			if(class_exists('ErrorPage')){
+				$errorFilePath = \ErrorPage::get_filepath_for_errorcode(
+					$this->statusCode,
+					class_exists('Translatable') ? \Translatable::get_current_locale() : null
+				);
+
+				if(false && file_exists($errorFilePath)) {
```
camspiers: This code will never run

main.php
```diff
@@ -117,6 +117,8 @@
 require_once('core/startup/ErrorControlChain.php');
 require_once('core/startup/ParameterConfirmationToken.php');
 
+ini_set('display_errors', 0);
```
camspiers: Is this meant to be here? Obviously this should be off in prod, but do we always want it off?

camspiers:
Looks really good. Just a couple of comments.

I'm not sure if it matters, but we might want to honor error suppression (when @ is used):

https://github.com/camspiers/silverstripe-loggerbridge/blob/master/src/Camspiers/LoggerBridge/LoggerBridge.php#L418-L421

Generally I would have a quick look over LoggerBridge to see if there is anything useful there:

https://github.com/camspiers/silverstripe-loggerbridge/blob/master/src/Camspiers/LoggerBridge/LoggerBridge.php

For example, currently if a request hits the memory limit, when using Monolog you likely won't have enough memory to complete the logging of the error.

composer.json
tractorcow: that's at least 1.15, e.g. 1.20 or 1.9999 (theoretically).

~ means "at least, but only allowed to increment the last digit"

main.php
tractorcow: Could this be done in such a way that it's only suppressed AFTER logging is setup, and some kind of check to prove that it's able to display errors? I'm scared of white screen of death appearing. Probably some time just after set_error_handler?

tractorcow: For instance, a YAML error might make the logging configuration broken, in which case I'd expect an error to be displayed.

composer.json
sminnee: @camspiers's point is valid: if my code works with 1.`<15` it would be nice to allow for that package. The main thing is that I haven't tested.

I'll check out an older version and re-test.

logging/DebugViewFriendlyErrorFormatter.php
sminnee: Doh - stupid debugging hack.

main.php
sminnee: The old code disabled the error within SilverStripe's error trapper, so that errors were not displayed twice. I believe it also incorporated the  `error_reporting()` check that @camspiers mentioned below.

Putting `display_errors` at the top here is awkward—it was placed here (rather than, for example, in the code where the handlers are registered) because of ErrorControlChain; what we probably need to do is introduce new functionality into the error handler. This may mean subclassing the monolog error handler, or contributing to it.

I had hoped we could do this in the `HTTPOutputHandler` class, but this is called *after* the built-in handler, as best I can see. I'll re-test, though.

sminnee:
@camspiers it looks like error suppression should already be honoured? https://github.com/Seldaek/monolog/blob/master/src/Monolog/ErrorHandler.php#L142

I'll check, though.

logging/HTTPOutputHandler.php
sminnee: I'll post an issue and leave a link in the TODO: I've refactored this from existing code and so I think actually fixing it is beyond the scope of this PR.

sminnee:
It looks like the monolog error handler already has memory reservation code too. :-)

main.php
sminnee: OK I stand corrected: moving the display_errors supression to HTTPOutputHandler works a treat.

sminnee:
OK, I've done the above update. A couple of things I haven't changed:

 * Error suppression and out-of-memory handling: this is already covered by Monolog's error handler.
 * The double-display of fatal errors is a Monolog bug: https://github.com/Seldaek/monolog/issues/622

sminnee:
The backtrace includes the 4 monolog calls. I could filter this, I guess—in the past we've done this by having a list of methods to ignore from the backtrace. Alternatively I could look through the backtrace for the file/line of the error message and hide everything before that.

I'm definitely not keen to monkey with the register_error_handler side of things, I really like how much of this we can delegate to monolog.

sminnee:
There we go...

camspiers:
You are right, It looks like Monolog does the error suppression correctly, but I can't see how the memory limit stuff works. A variable is set called reservedMemory but I don't see how it is used.

camspiers:
Regarding the double logging I am pretty sure that is why I added this:

https://github.com/camspiers/silverstripe-loggerbridge/blob/master/src/Camspiers/LoggerBridge/LoggerBridge.php#L465

sminnee:
> I can't see how the memory limit stuff works. A variable is set called reservedMemory but I don't see how it is used.

It took me a while to figure out. It's a 20,000 character string that is set to null before error processing happens. When it is set to null, 20,000 bytes of memory is freed up. :-)

The benefit is that is doesn't require the system to find more memory, it reserves 20kb pre-emptively. Obviously you wouldn't want to keep 5mb free like your code ;-)

sminnee:
In general, however, I'm much more in favour of deferring any such concerns the monolog error handler, rather than patching it within SilverStripe.

camspiers:
Oh I see that is a cool approach. Regarding the 5M, it wasn't arbitraily chosen. I experimented with out of memory errors and found that log handling wasn't always achieved with anything less. This was probably do with backtrace memory allocation, combined with the loggers and their handlers. Experiment, you'll see ;) the memory limit is configurable though :)

sminnee:
OK I'm going to squash cb4e990bbcabdc98f2789b2506d80b3a154742c1 and 449d89c9717a2e82d5c8c889e93f153aaeffe7fa into the main commit now, to make things tidier after merge.

sminnee:
Anyone want to merge this? :)

tractorcow:
It's un-mergeable sorry. Rebase?

tractorcow:
I think I need to spend a lot of time reviewing this properly. :)

sminnee:
Rebased.

logging/Log.php
```diff
@@ -0,0 +1,87 @@
...
+	 * The message parameter can be either a string (a simple error
+	 * message), or an array of variables. The latter is useful for passing
+	 * along a list of debug information for the writer to handle, such as
+	 * error code, error line, error context (backtrace).
+	 *
+	 * @param mixed $message Exception object or array of error context variables
+	 * @param const $priority Priority. Possible values: SS_Log::ERR, SS_Log::WARN or SS_Log::NOTICE
+	 * @param  mixed    $extras    Extra information to log in event
+	 */
+	public static function log($message, $priority, $extras = null) {
```
tractorcow: i18n::_t is still calling this method.

tractorcow: Shall we add a `@deprecation` phpdoc for both deprecated methods and classes? This will show up on api.silverstripe.org as deprecated (striked out)

logging/DetailedErrorFormatter.php
```diff
@@ -0,0 +1,116 @@
...
+			);
+		} else {
+			$context = $record['context'];
+			foreach(array('code','message','file','line') as $key) {
+				if(!isset($context[$key])) {
+					$context[$key] = isset($record[$key]) ? $record[$key] : null;
+				}
+			}
+
+			$trace = debug_backtrace();
```
tractorcow: For fatal errors this is pointless, since it's registered as a `register_shutdown_function`. This gives the following stacktrace.

```
SilverStripe\Framework\Logging\DetailedErrorFormatter->format(Array) 
AbstractProcessingHandler.php:35
Monolog\Handler\AbstractProcessingHandler->handle(Array) 
Logger.php:289
Monolog\Logger->addRecord(550,Fatal Error (E_USER_ERROR): OH NO,Array) 
Logger.php:467
Monolog\Logger->log(alert,Fatal Error (E_USER_ERROR): OH NO,Array) 
ErrorHandler.php:176
Monolog\ErrorHandler->handleFatalError() 
```

tractorcow: Non fatal errors (warnings etc) are fine.

logging/DetailedErrorFormatter.php
```diff
@@ -0,0 +1,116 @@
...
+			}
+		}
+		return null;
+	}
+
+	/**
+	 * Render a developer facing error page, showing the stack trace and details
+	 * of the code where the error occured.
+	 *
+	 * @param unknown_type $errno
```
tractorcow: I think @assertchris would be angry at me if I didn't ask you to type these. :) `mixed` is probably ok, but I'd guess int, string, string, int, array

core/Core.php
```diff
@@ -136,8 +135,9 @@
 /**
  * Load error handlers
  */
-Debug::loadErrorHandlers();
 
+$errorHandler = Injector::inst()->get('ErrorHandler');
+$errorHandler->start();
```
tractorcow: If we are defining a service with a promised method `start` should we define this as an interface somewhere? SilverStripe\Framework\Logging\ErrorHandler?

logging/DetailedErrorFormatter.php
sminnee: Does this block merge? I'm not forking Monolog just to get a slightly nicer stacktrace in edge-cases.

logging/Log.php
sminnee: Good call, will do.

conf/ConfigureFromEnv.php
```diff
@@ -139,7 +142,12 @@
 }
 
 if(defined('SS_ERROR_LOG')) {
-	SS_Log::add_writer(new SS_LogFileWriter(BASE_PATH . '/' . SS_ERROR_LOG), SS_Log::WARN, '<=');
+	$logger = Injector::inst()->get('Logger');
+	if($logger instanceof Logger) {
+		$logger->pushHandler(new StreamHandler(BASE_PATH . '/' . SS_ERROR_LOG, Logger::WARN));
```
tractorcow: Rather than adding this handler here, why not create a new custom SS handler (which extends StreamHandler) and then add this to the logger via yml.

You can add the logic for `if(defined('SS_ERROR_LOG')` to the class itself, which does nothing if it's not defined.

Less coupling between monolog that way. :)

logging/DetailedErrorFormatter.php
sminnee: Good call, will fix.

core/Core.php
sminnee: Good call, will fix.

control/HTTPResponse.php
```diff
@@ -271,7 +271,10 @@ public function output() {
 			// an error, and the response doesn't have any body yet that might contain
 			// a more specific error description.
 			if(Director::isLive() && $this->isError() && !$this->body) {
-				Debug::friendlyError($this->statusCode, $this->getStatusDescription());
+				$formatter = new SilverStripe\Framework\Logging\DebugViewFriendlyErrorFormatter();
```
tractorcow: `$formatter = Injector::get('FriendlyErrorFormatter');`

dev/DebugView.php
```diff
@@ -99,10 +132,45 @@ public function Breadcrumbs() {
...
+
+	public function writeVariable($val, $caller) {
+		Deprecation::notice('4.0', 'Use renderVariable() instead');
+		echo $this->renderVariable($val, $caller);
+	}
+
 	/**
 	 * Render HTML header for development views
 	 */
-	public function writeHeader() {
```
tractorcow: PHPDoc please. :D

logging/DetailedErrorFormatter.php
sminnee: Spoke to @tractorcow, we decided that in these cases it's probably better to avoid the trace entirely: https://gist.github.com/124f8f3d3bc8db8fbbf7

conf/ConfigureFromEnv.php
sminnee: I disagree.

 * Less Code > More Code.
 * ConfigureFromEnv is currently the thing responsible for mapping env vars to internal config
 * The old implementation did this way; if we want to refactor ConfigureFromEnv, that's beyond the scope of this PR.

control/HTTPResponse.php
sminnee: Good call.

dev/DebugView.php
sminnee: :+1:

conf/ConfigureFromEnv.php
tractorcow: Ok, I don't think it blocks merge, but this isn't something I don't think I can't suggest a change to later. Maybe I'll raise a PR then.

In my idea, the subclass would be injected, so it would only rely on a yml config to refactor it out, whereas configurefromenv isn't. The use case I am thinking of "I want to inject a new handler for sites which define SS_ERROR_LOG". which currently isn't possible.

sminnee: Let's pick this up as part of a more complete refactor of `ConfigureFromEnv.php` into yaml config.

sminnee:
OK—I've addressed the above points and rebased to latest master. Think we're ready to merge?

tractorcow:
I think so!

stevie-mayhew:
:+1: 

Nice work Sam.

camspiers:
:+1:

sminnee:
Woo!

tractorcow:
I just noticed that there's still a `Debug::loadErrorHandlers();` reference in TestRunner.

PR #3411, opened 2014-08-20
Update GridFieldExportButton.php

JorisDebonnet:
Quotes need to be escaped as "", not \".
http://tools.ietf.org/html/rfc4180#page-3 (point 7)

By the way, the `generateExportFileData` function can probably be simplified a lot by using [fputcsv](http://php.net/manual/en/function.fputcsv.php) (like [this](http://www.zyxware.com/articles/3527/how-to-create-csv-files-using-php)), but I didn't want to risk breaking any custom functionality for this simple change.

PR #4432, opened 2015-07-27
Use fputcsv in GridFieldExportButton

JorisDebonnet:
This is a (slightly improved) rebase of #3411. 

Uses php's built-in `fputcsv` in order to generate csv in the standard format. This includes the ability to specify the enclosure, which defaults to ". 

CSV output will change very slightly with this, because values that do not need enclosing, no longer get enclosing. In other words, the method previously wrapped every single value in quotes, while `fputcsv` only does that when that is required (due to a linebreak or a character that needs escaping). 

(I can make these changes against 3.2 instead, if that is desired)

### DIFF
--- a/forms/gridfield/GridFieldExportButton.php
+++ b/forms/gridfield/GridFieldExportButton.php
@@ -21,6 +21,11 @@ class GridFieldExportButton implements GridField_HTMLProvider, GridField_ActionP
 	protected $csvSeparator = ",";
 
 	/**
+	 * @var string
+	 */
+	protected $csvEnclosure = '"';
+
+	/**
 	 * @var boolean
 	 */
 	protected $csvHasHeader = true;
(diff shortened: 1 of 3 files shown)


forms/gridfield/GridFieldExportButton.php
```diff
@@ -165,7 +164,14 @@ public function generateExportFileData($gridField) {
...
 
-		return $fileData;
+		// Convert the $fileData array into csv by capturing fputcsv's output
+		ob_start();
+		$fp = fopen('php://output', 'w');
+		foreach($fileData as $line) {
+			fputcsv($fp, $line, $this->csvSeparator, $this->csvEnclosure);
+		}
+		fclose($fp);
+		return ob_get_clean();
```
tractorcow: I would suggest using a temp buffer instead of piggy backing off the output buffer (which could already contain buffered content).

http://www.php.net/manual/en/function.fputcsv.php#74118

JorisDebonnet: `ob_start` calls can be nested though, so it should work even if people are already buffering content. But I'm happy to use that other method, it does look more elegant!

sminnee:
Merging now before it has time to get stale again. Looks great, and thanks for the rebase, Joris!

tractorcow:
You've done well @JorisDebonnet!

JorisDebonnet:
Thank you! :)

Issue #4063, opened 2015-04-05
ViewableData::defineMethods() wasting memory?


kinglozzer:
https://github.com/silverstripe/silverstripe-framework/blob/1db08bac88f9330dc4e6dda1ae08628f245a5212/view/ViewableData.php#L177-L183

This is run for every subclass of `ViewableData` that’s called during a request, fetching and then iterating over every single method of every class! I can’t find this caching behaviour documented anywhere, looks like it’s probably a pre-open-source relic? Either way, given that it’s using around 1mb of memory, I think it should probably be removed...

[Before](http://i.imgur.com/FrPEpoY.png) and [after](http://i.imgur.com/F9B8bp7.png) XHProf

tractorcow:
That's a nice performance increase. :)

This is one reason I'd love to have dynamically assigned traits instead of extension methods. _call no longer has to do all the hard work. But, 4.0 land. :)

Does viewable data even have the chance to call defineMethods from the failover? During construction of the object that variable isn't even assigned is it?

How many methods match the second part of the function, which are prefixed with a single underscore? Is this a template pattern that SSViewer relies on?

kinglozzer:
> Does viewable data even have the chance to call defineMethods from the failover? During construction of the object that variable isn't even assigned is it?

`defineMethods()` is called right at the end of `Object::__construct()`, and subclasses [seem to](https://github.com/silverstripe/silverstripe-cms/blob/7f1f6156347d6e5c6135d12b35090368bad94188/code/controllers/ContentController.php#L46) set the failover before calling `parent::__construct()`, so I think the failover object is available at this point. Using `?debugfailover=1` seems to indicate that too

> How many methods match the second part of the function, which are prefixed with a single underscore? Is this a template pattern that SSViewer relies on?

I can’t find any in core that it’s even possible to use from a template. `_t()` matches the convention, but that’s a special case that wouldn’t hit this method.

If there’s a BC risk, we could always change the method to something other than `cachedCall` that triggers a deprecation notice before calling `cachedCall` and remove this in 4.0?

kinglozzer:
Closed in https://github.com/silverstripe/silverstripe-framework/pull/5232

PR #4436, opened 2015-07-27
Deprecate ViewableData magic caching logic

kinglozzer:
Proposed method of deprecating the unnecessary caching logic that `ViewableData::defineMethods()` adds. See #4063 for more info, but basically:

- It’s not documented anywhere.
- It wastes memory - it will loop over *every* method in `ViewableData` (and *every* method of *every* subclass of `ViewableData`) looking for methods that begin with an underscore and magically caching them.

It’s a bit confusing at a glance, this is roughly how it works:

```php
<?php

class MyController extends Controller {
	public function index() {
		// Will only echo 'Magic!' once
		$this->magic();
		$this->magic();
	}

	public function _magic() {
		echo 'Magic!';
	}
}
```

The notice text is a bit verbose, but it’s confusing behaviour and I didn’t know how else to describe it. Plus the notice tells you it was called from `__lambda_func`, which isn’t very helpful.

Given that this isn’t documented at all, it doesn’t technically form part of our public API so it might be worth considering introducing this change earlier? I’d say it’s _very_ unlikely that anyone is actually using this behaviour, but I guess it’s possible which is why I was cautious and decided to open this against master.

### DIFF
--- a/view/ViewableData.php
+++ b/view/ViewableData.php
@@ -177,12 +177,31 @@ class ViewableData extends Object implements IteratorAggregate {
 		foreach($this->allMethodNames() as $method) {
 			if($method[0] == '_' && $method[1] != '_') {
 				$this->createMethod(
-					substr($method, 1), "return \$obj->cachedCall('$method', \$args, '" . substr($method, 1) . "');"
+					substr($method, 1),
+					"return \$obj->deprecatedCachedCall('$method', \$args, '" . substr($method, 1) . "');"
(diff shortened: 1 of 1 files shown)


dhensby:
If it's in master, just remove the behaviour, don't add a deprecation

Deprecate it in a 3.x release - otherwise this will hang around in 4.x too

kinglozzer:
D’oh. #4437

PR #4441, opened 2015-07-27
FIX Missing thumbnails and inconsistencies

jonom:
Some quick fixes for consistency when selecting files. Companion fix for CMS at https://github.com/silverstripe/silverstripe-cms/pull/1250

* Standardised thumbnail presentation
* Allow sort by created date for accessing recently uploaded files
* Appropriate pagination limits to make use of space without hiding form action buttons
* Display Title instead of Filename

**Before:**
silverstripe_-_edit_page

**After:**
silverstripe_-_edit_page

**Before:**
silverstripe_-_edit_page
**After:**
silverstripe_-_edit_page

**Before:**
silverstripe_-_files

**After:**
silverstripe_-_files

### DIFF
--- a/forms/HtmlEditorField.php
+++ b/forms/HtmlEditorField.php
@@ -311,7 +311,7 @@ class HtmlEditorField_Toolbar extends RequestHandler {
 			new GridFieldFilterHeader(),
 			new GridFieldSortableHeader(),
 			new GridFieldDataColumns(),
-			new GridFieldPaginator(5),
+			new GridFieldPaginator(7),
 			// TODO Shouldn't allow delete here, its too confusing with a "remove from editor view" action.
 			// Remove once we can fit the search button in the last actual title column
 			new GridFieldDeleteAction(),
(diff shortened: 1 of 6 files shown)


forms/HtmlEditorField.php
```diff
@@ -323,8 +323,12 @@ public function MediaForm() {
...
 		$columns = $fileField->getConfig()->getComponentByType('GridFieldDataColumns');
 		$columns->setDisplayFields(array(
-			'CMSThumbnail' => false,
-			'Name' => _t('File.Name'),
+			'StripThumbnail' => false,
+			'Title' => _t('File.Title'),
+			'Created' => singleton('File')->fieldLabel('Created'),
+		));
+		$columns->setFieldCasting(array(
+			'Created' => 'Date->Nice'
```
dhensby: I think it'd be better for this to be datetime rather than date

jonom: For me the time it was created/uploaded is not very useful and just adds clutter - but appreciate it could be different for other users. The sort will still take the time in to account I believe because the sort is done on the original rather than presented value, it's just not shown here. This is consistent with presentation of created date in the Files area currently. @clarkepaul do you have any thoughts on this one?

clarkepaul: I'm not exactly sure where the time is being displayed but I'm guessing we are talking about it being displayed in the grid view. As long as the user can drill into the details to access the time I would say that it isn't that necessary (of course there will be edge cases where it would be very useful). A middle ground could be to add the time next to the date but make it lighter in color so it doesn't stand out as much making it secondary information. I think I'd be more inclined to go without it if lightening the time doesn't look good or adds too much clutter.

jonom: Here's how it looks with date + time (`'Created' => 'SS_Datetime->Nice'`). I can't see an easy way to style the time differently than the date as they're not in separate html elements, but I don't think it looks too bad this way. Now that I've tried it out I don't think it looks too cluttered and I can see how it would be useful to some people so I'm in favour of including the time after all if you guys agree?

By the way I thought 'Uploaded' might be a better label than 'Created' for that field but I wasn't sure what implications changing that would have for i18n so I didn't go there.

screen shot 2015-07-30 at 9 17 48 am

clarkepaul: Cool just go with it - we can just keep the Uploaded/Created/Added label thing in mind, looking good

jonom:
Updated so all 3 GridFields now show date+time

tractorcow:
@jonom for president of UX!

jonom:
lol thanks for merge!

PR #4443, opened 2015-07-28
Url-safe alternative for base64_encode in resampled Image filenames

JorisDebonnet:
This fixes a potential issue when `base64_encode` generates e.g. a `/` in a resampled Image filename.

### DIFF
--- a/core/Convert.php
+++ b/core/Convert.php
@@ -290,8 +290,8 @@ class Convert {
 	/**
 	 * Create a link if the string is a valid URL
 	 *
-	 * @param string The string to linkify
-	 * @return A link to the URL if string is a URL
+	 * @param string $string The string to linkify
+	 * @return string A link to the URL if string is a URL
 	 */
 	public static function linkIfMatch($string) {
 		if( preg_match( '/^[a-z+]+\:\/\/[a-zA-Z0-9$-_.+?&=!*\'()%]+$/', $string ) )
@@ -305,7 +305,9 @@ class Convert {
 	 *
 	 * @param string $data Input data
 	 * @param bool $preserveLinks
-	 * @param int $wordwrap
+	 * @param int $wordWrap
+	 * @param array $config
+	 * @return string
 	 */
 	public static function html2raw($data, $preserveLinks = false, $wordWrap = 0, $config = null) {
 		$defaultConfig = array(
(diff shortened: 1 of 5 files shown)


JorisDebonnet:
Note that I ended up using `~_` instead of `-_` as the safe character replacements, since 3.2 is still using the `-` character in resampled Image filenames to make the distinction between the transformation part and the original-filename part (e.g. `ScaleHeightWzIwMF0-filename.jpg`).

model/Image.php
```diff
@@ -1051,3 +1051,14 @@ public function write($showDebug = false, $forceInsert = false, $forceWrite = fa
 		throw new Exception("{$this->ClassName} can not be written back to the database.");
 	}
 }
+
```
tractorcow: I don't think this is the right place to add new global methods. These should be in `Convert` class as static methods instead. Perhaps consider including the json_en/decoding in the method as well as a generic 'encode array as url safe text'

JorisDebonnet: Isn't there any kind of helper file where global methods can be defined? It would make some sense because `base64url_decode` and `base64url_decode` are often used as names for these purposes in [other languages](http://scriptogr.am/hdknr/post/base64url-encoding) as well. Granted, I had to change it a little from the usual way, to include `~` instead of `-`... 

If they go into Convert.php, should they be called `raw2base64url` (to be consistent with several other methods there) or just the same names as below?

tractorcow: > Isn't there any kind of helper file where global methods can be defined?

Global text conversion methods are in `Convert` :D

We tend to avoid creating global methods where possible. In this case, it's fine to scope it to a class, since these aren't really standard php methods (even if they exist on other languages).

you can call it whatever you want as long as it's documented and makes sense :P

sminnee:
I'm a little concerned about whether encoding and then decoding data that contains special characters is going to preserve them. Can we have some ConvertTest.php cases for that? Something like:

    $data = 'scarystringorarray';
    $this->assertEquals($data, Convert:: base64url_decode(Convert:: base64url_encode($data));

Might be a good place to use `@dataProvider` :)

We're about to release beta2, but it would be nice to get this into rc1, as it is essentially a bugfix.

core/Convert.php
```diff
@@ -414,8 +416,29 @@ public static function raw2url($title) {
...
+	}
+
+	/**
+	 * Decode a value that was encoded with Convert::base64url_encode.
+	 *
+	 * @param string $val Value to be decoded
+	 * @return mixed Original value
+	 */
+	public static function base64url_decode($val) {
+		return json_decode(base64_decode(str_pad(strtr($val, '~_', '+/'), strlen($val) % 4, '=', STR_PAD_RIGHT)));
```
tractorcow: You're going to run into issues here where you base64url_encode an associative array, and it is returned as a stdObj. Maybe json_decode should have second parameter true?

JorisDebonnet: Thanks for the suggestion, I'll add that.

JorisDebonnet:
I added a Test! 
I also added in the PHPDoc for `Convert::base64url_encode` that any string data must be UTF-8 encoded (just like the `json_encode` PHPDoc says). That will barely ever be an issue however, since it's currently only used for encoding Image transformation parameters.

dhensby:
Looks good

PR #4444, opened 2015-07-28
Making extractTags public

nickspiel:
It would be super handy if the `extractTags` function was public.
This class is a real hassle to extend.

### DIFF
--- a/parsers/ShortcodeParser.php
+++ b/parsers/ShortcodeParser.php
@@ -143,93 +143,7 @@ class ShortcodeParser extends Object {
 
 		return $content;
 	}
-
-	// --------------------------------------------------------------------------------------------------------------
 	
-	protected function removeNode($node) {
-		$node->parentNode->removeChild($node);
-	}
-	
-	protected function insertAfter($new, $after) {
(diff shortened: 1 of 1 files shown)


tractorcow:
Can you merge this into master? Changing method visibility can cause regressions where methods are overridden in subclasses.

I would also suggest not re-arranging the method order if possible, since it makes the diffs huge and hard to merge up.

nickspiel:
I thought I would rearrange the file so it was with the rest of the public functions, this seemed logical based on how functions were grouped in the file. Happy to revert this to it's original position if this is not correct.

Will merge this over to master now.

dhensby:
@nickspiel the problem with large re-arrangements is that when patches are merged up the resulting conflicts are very hard to resolve. Ideally we would order the functions logically, but it does pollute the merge process so we have to balance that out.

tractorcow:
If it's merged into master, the re-arrangement is ok. :D

dhensby:
I think we'll need to merge this into master by hand.

nickspiel:
Sorry this has taken awhile guys. I have opened a new PR (#4607) on master without reordering so it can more easily merge.

PR #4449, opened 2015-07-29
ENH: moving hardcoded nice date/time formats to config

IgorNadj:
I was using Date.Nice() in my template today, and wanted it to have an even nicer format, but soon found that it was hardcoded, and I could only use Date.Format('...'), copy-pasting it across my template files.

It's generally good practice to move the default date format out to a config somewhere, so I have done so in this commit. The default stays the same, but if you want to change it, you just pop into your site's config.yml and do:

Date:
  niceFormat: '...'
Datetime:
  niceFormat: '...'
Time:
  niceFormat: '...'

Below is some code that might be handy for the people reading this if they want to play around with this. Thanks,
Igor

--------------------------------------------------------------
// mysite/code/TestDateFormat.php
<?php 

class TestDateFormat extends Controller {
	
	public function index(){
		echo '<pre>';
		
		$objs = array(
			'Date'        => new Date(),
			'SS_Datetime' => new SS_Datetime(),
			'Time'        => new Time(),
		);
		$objs['Date']->setValue('2015-07-29');
		$objs['SS_Datetime']->setValue('2015-07-29 17:15:55');
		$objs['Time']->setValue('17:15:55');
		
		$newNiceFormats = array(
			'Date'        => 'd F Y',
			'SS_Datetime' => 'd F Y, H:i:s',
			'Time'        => 'H:i:s',	
		);
		
		foreach($objs as $class => $obj){
			echo PHP_EOL.PHP_EOL.'CLASS: '.$class.PHP_EOL;
			
			$defaultNiceFormat = Config::inst()->get($class, 'niceFormat');
			echo 'Default niceFormat is: "'.$defaultNiceFormat.'"'.PHP_EOL;
			echo 'getValue() = "'.$obj->getValue().'"'.PHP_EOL;
			echo 'Nice()     = "'.$obj->Nice().'"'.PHP_EOL.PHP_EOL;
			
			$newNiceFormat = $newNiceFormats[$class];
			Config::inst()->update($class, 'niceFormat', $newNiceFormat);
			echo 'New niceFormat is: "'.$newNiceFormat.'"'.PHP_EOL;
			echo 'Nice()     = "'.$obj->Nice().'"'.PHP_EOL.PHP_EOL;
		}
		
		
		die;
		
	}
	
}

// mysite/_config.php
Config::inst()->update('Director', 'rules', array('testdateformat' => 'TestDateFormat')); // now go to http://localhost/testdateformat

### DIFF
--- a/model/fieldtypes/Date.php
+++ b/model/fieldtypes/Date.php
@@ -19,7 +19,15 @@
  * @subpackage model
  */
 class Date extends DBField {
-
+	
+	/**
+	 * @config
+	 * @see also SS_DateTime::niceDatetimeFormat
+	 * @see also Time::niceFormat
+	 */
+	private static $niceFormat = 'd/m/Y';
+	
+	
 	public function setValue($value, $record = null) {
 		if($value === false || $value === null || (is_string($value) && !strlen($value))) {
 			// don't try to evaluate empty values with strtotime() below, as it returns "1970-01-01" when it should be
@@ -59,16 +67,18 @@ class Date extends DBField {
 	}
 
 	/**
-	 * Returns the date in the format dd/mm/yy
-	 */
+	 * Returns the date in the format specified by the config value niceFormat, or dd/mm/yy by default
+	 */	 
 	public function Nice() {
-		if($this->value) return $this->Format('d/m/Y');
+		if($this->value) return $this->Format(Config::inst()->get(__CLASS__, 'niceFormat'));
 	}
-
+	
 	/**
 	 * Returns the date in US format: “01/18/2006”
+	 * @deprecated 3.2 Use Nice() with config setting Date::niceFormat instead
 	 */
 	public function NiceUS() {
+		Deprecation::notice('3.2', 'Use Nice() with config setting '.__CLASS__.'::niceFormat instead');
 		if($this->value) return $this->Format('m/d/Y');
 	}
 
--- a/model/fieldtypes/Time.php
+++ b/model/fieldtypes/Time.php
@@ -15,7 +15,15 @@
  * @subpackage model
  */
 class Time extends DBField {
-
+	
+	/**
+	 * @config
+	 * @see also Date::niceFormat
+	 * @see also SS_DateTime::niceFormat
+	 */
+	private static $niceFormat = 'g:ia';
+	
+	
 	public function setValue($value, $record = null) {
 		if($value) {
 			if(preg_match( '/(\d{1,2})[:.](\d{2})([a|A|p|P|][m|M])/', $value, $match )) $this->TwelveHour( $match );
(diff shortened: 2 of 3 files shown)


model/fieldtypes/Date.php
```diff
@@ -19,7 +19,15 @@
...
  */
 class Date extends DBField {
-
+	
+	/**
+	 * @config
+	 * @see also SS_DateTime::niceDatetimeFormat
+	 * @see also Time::niceFormat
+	 */
+	private static $niceFormat = 'd/m/Y';
```
micmania1: statics should be underscores :p $nice_format

dhensby:
You can also use `$this->config()->nice_format` rather than the more verbose one you are using.

Whilst I agree the formats should not be hard coded (unless they are adhering to a standard) it's possible to add your own format instead. see: https://gist.github.com/dhensby/3b542c8c634e18da22fa

An extension like that can be added to Date and then provide a template helper `$Now.DefaultFormat`

This also gives you some flexibility in terms of adding logic, which the config does not (as it is static).

Lastly, this would have to go against 3 branch as it is a new feature/API.

model/fieldtypes/Date.php
```diff
@@ -59,16 +67,18 @@ public function setValue($value, $record = null) {
 	}
 
 	/**
-	 * Returns the date in the format dd/mm/yy
-	 */
+	 * Returns the date in the format specified by the config value niceFormat, or dd/mm/yy by default
+	 */	 
 	public function Nice() {
-		if($this->value) return $this->Format('d/m/Y');
+		if($this->value) return $this->Format(Config::inst()->get(__CLASS__, 'niceFormat'));
```
dhensby: `$this->config()->niceFormat` or `$this->config()->get('niceFormat')`

IgorNadj:
>statics should be underscores :p $nice_format

@micmania1 Doh, late afternoon commits ftl

>Whilst I agree the formats should not be hard coded (unless they are adhering to a standard) it's possible to add your own format instead. see: https://gist.github.com/dhensby/3b542c8c634e18da22fa

@dhensby Yep, I get the extension side of it, refactoring the hard-coded thing is my main motivator here

>$this->config()->niceFormat

Ah, much better thanks!

>Lastly, this would have to go against 3 branch as it is a new feature/API.

Why do you think this is a new feature/API? To me it seems like simple cleanup, no behaviour is changed.

model/fieldtypes/Time.php
```diff
@@ -26,22 +34,25 @@ public function setValue($value, $record = null) {
...
 	/**
 	 * Return a user friendly format for time
 	 * in a 24 hour format.
-	 *
+	 * 
 	 * @return string Time in 24 hour format
+	 * @deprecated 3.2 Use Nice() with config setting niceFormat instead
 	 */
 	public function Nice24() {
+		Deprecation::notice('3.2', 'Use Nice() with config setting '.__CLASS__.'::niceFormat instead');
```
tractorcow: I'd make this a separate config rather than deprecating it. Saying to use another method and changing the config defeats the purpose of having two methods; It's useless having it if you can only have one global format for your project. What if you want $Nice in one place and $Nice24 in another?

IgorNadj: Yeah I was thinking about that. "Nice()" to me means give me something human readable. Whereas Nice24() means I want a specific format. I guess maybe Nice24 should be renamed to something else then, but I can't think of what.

dhensby:
I think we just need a nicer way to name and fetch date formats...

sminnee:
@ig

sminnee:
@IgorNadj did you want to finish this one off?

IgorNadj:
@sminnee sure, I will recreate against the 3 branch. It looks like the only remaining concern was what to do with Nice24(). I will leave it as is for now, simplest solution.

tractorcow:
@IgorNadj please link to the PR when you create it.

Issue #4442, opened 2015-07-27
Injector factories should be passed the full service definition


sminnee:
Injector factories could be more flexible if they were passed the full service definition as a 3rd argument.

For example, we could use a custom factory to do some special processing on selected `properties`. This relates to my work on #4430

 * Advantage: Factories can be used to tidy up confusing aspects of Injector service definition
 * Disadvantage: Factories could do anything, risking that we get a more divergent & confusing set of configuration approaches. We may end up with service definitions that have options other than `class`, `constructor`, `properties`, and `dependencies`.

One way to mitigate that disadvantage would be to set some guidelines for writing good factories.

The change would be trivial and I'll post a PR if I get buy-in, but I want to run it past people first.

sminnee:
Note: the specific use-case that I want is that I want to grab the `properties.Handler` argument, do some processing, and pass its contents as repeated calls to `Logger::pushHandler()`. In doing this, my hope is that same configuration approach can be maintained even if Monolog improves things in the future to mitigate the need for a custom factory.

An alternative approach to this would be to shift the handling of the `properties` and `dependencies` argument to become part of the factory. Since it all relates to instantiation & set-up of the object, I don't see why that would be *wrong*, but others may disagree.

@nyeholt since you wrote the original class, it would be good to get your feedback.

sminnee:
@stevie-mayhew you may have views on this.

sminnee:
A related issue that came up discussing logger was the ability to inline service definitions within properties.

Right now, to create a handler with a custom formatter, two service definitions are necessary:

    Injector:
      DisplayErrorHandler:
        class: SilverStripe\Framework\Logging\HTTPOutputHandler
        constructor:
          - "notice"
        properties:
          Formatter: %$DisplayErrorFormatter
      DisplayErrorFormatter:
        class: SilverStripe\Framework\Logging\DetailedErrorFormatter

It would be nicer if a service definition could be inlined:

    Injector:
      DisplayErrorHandler:
        class: SilverStripe\Framework\Logging\HTTPOutputHandler
        constructor:
          - "notice"
        properties:
          Formatter:
            class: SilverStripe\Framework\Logging\DetailedErrorFormatter

However, this gets a little bit "magic": you have to know that an map containing a class key should be interpreted as a service definition. It would be nicer if there was a mechanism for indicating an inlined service definition. I'm not sure what that should be, though.

[...]

PR #4450, opened 2015-07-29
NEW: Add ‘calls’ section to Injector configs.

sminnee:
As well as properties, you can now configure a series of method calls in your service definitions.

I started tinkering with Injector in response to #4442, and I ended up deciding that this was a less invasive change to start with for now.

### DIFF
--- a/control/injector/Injector.php
+++ b/control/injector/Injector.php
@@ -585,12 +585,49 @@ class Injector {
 		$objtype = $asType ? $asType : get_class($object);
 		$mapping = isset($this->injectMap[$objtype]) ? $this->injectMap[$objtype] : null;
 
+		$spec = empty($this->specs[$objtype]) ? array() : $this->specs[$objtype];
+
 		// first off, set any properties defined in the service specification for this
 		// object type
-		if (isset($this->specs[$objtype]) && isset($this->specs[$objtype]['properties'])) {
+		if(!empty($spec['properties']) && is_array($spec['properties'])) {
 			foreach ($this->specs[$objtype]['properties'] as $key => $value) {
 				$val = $this->convertServiceProperty($value);
 				$this->setObjectProperty($object, $key, $val);
+			}
+		}
+
+		// Populate named methods
+		if (!empty($spec['calls']) && is_array($spec['calls'])) {
+			foreach ($spec['calls'] as $method) {
+				// Ignore any blank entries from the array; these may be left in due to config system limitations
+				if(!$method) continue;
+
+				// Format validation
+				if(!is_array($method) || !isset($method[0]) || isset($method[2])) {
+					throw new \InvalidArgumentException(
+						"'calls' entries in service definition should be 1 or 2 element arrays."
+					);
+				}
+				if(!is_string($method[0])) {
+					throw new \InvalidArgumentException("1st element of a 'calls' entry should be a string");
+				}
+				if(isset($method[1]) && !is_array($method[1])) {
+					throw new \InvalidArgumentException("2nd element of a 'calls' entry should an arguments array");
+				}
+
+				// Check that the method exists and is callable
+				$objectMethod = array($object, $method[0]);
+				if (!is_callable($objectMethod)) {
+					throw new \InvalidArgumentException("'$method[0]' in 'calls' entry is not a public method");
+				}
+
+				// Call it
+				call_user_func_array(
+					$objectMethod,
+					$this->convertServiceProperty(
+						isset($method[1]) ? $method[1] : array()
+					)
+				);
 			}
 		}
 
--- a/docs/en/02_Developer_Guides/05_Extending/05_Injector.md
+++ b/docs/en/02_Developer_Guides/05_Extending/05_Injector.md
@@ -111,6 +111,16 @@ Now the dependencies will be replaced with our configuration.
 	echo ($object->textProperty == 'My Text Value');
 	// returns true;
 
+As well as properties, method calls can also be specified:
+
+	:::yml
+	Injector:
+	  Logger:
+	    class: Monolog\Logger
+	    calls:
+	      - [ pushHandler, [ %$DefaultHandler ] ]
+
+
 ## Factories
 
 Some services require non-trivial construction which means they must be created by a factory class. To do this, create
--- a/tests/injector/InjectorTest.php
+++ b/tests/injector/InjectorTest.php
@@ -619,6 +619,91 @@ class InjectorTest extends SapphireTest {
 		$this->assertInstanceOf('TestObject', $injector->get('service'));
 	}
 
+	public function testMethods() {
+		// do it again but have test object configured as a constructor dependency
+		$injector = new Injector();
+		$config = array(
+			'A' => array(
+				'class' => 'TestObject',
+			),
+			'B' => array(
+				'class' => 'TestObject',
+			),
+			'TestService' => array(
+				'class' => 'TestObject',
+				'calls' => array(
+					array('myMethod', array('%$A')),
+					array('myMethod', array('%$B')),
+					array('noArgMethod')
+				)
+			)
+		);
+
+		$injector->load($config);
+		$item = $injector->get('TestService');
+		$this->assertTrue($item instanceof TestObject);
+		$this->assertEquals(
+			array($injector->get('A'), $injector->get('B'), 'noArgMethod called'),
+			$item->methodCalls
+		);
+	}
+
+	/**
+	 * @expectedException InvalidArgumentException
+	 */
+	public function testNonExistentMethods() {
+		$injector = new Injector();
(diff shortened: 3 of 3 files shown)


docs/en/02_Developer_Guides/05_Extending/05_Injector.md
```diff
@@ -111,6 +111,16 @@ Now the dependencies will be replaced with our configuration.
...
 	// returns true;
 
+As well as properties, method calls can also be specified:
+
+	:::yml
+	Injector:
+	  Logger:
+	    class: Monolog\Logger
+	    methods:
+	      - [ pushHandler: [ %$DefaultHandler ] ]
```
tractorcow: Excuse my yaml, but isn't this the same as

```yaml
methods: [[ pushHandler: [ %$DefaultHandler ] ]]
```

sminnee: Yeah but I like that less ;-) in particular I like 1 line per method call

tractorcow: I think it's meant to be a two-length array, I mean, not a one-length associative array?

sminnee: oh, yes, I see now. i'll fix.

note that i've opted to copy Symfony's syntax for these.

sminnee:
@nyeholt @camspiers @stevie-mayhew after much talking, a tiny piece of code. Any objections to merging this?

nyeholt:
Looks fine to me. The length of that method now stands out as a prime target for refactoring - but that's for another day.

tractorcow:
@sminnee could the list of methods be an associative array? Maybe if you have named method invocations, you could remove them via this key (or assign as null).

I think the current code supports it, but maybe add a test and docs just to verify that?

```yaml
methods:
  handlers: [ pushHandler, [ %$DefaultHandler ] ]
```

could be overridden by

```yaml
methods:
  handlers: null
``

**Comment by stevie-mayhew** [comment]
Works well from my end - although `setHandlers` is still going to be nicer once merged in monolog :)

Nice work Sam.

File: control/injector/Injector.php [hunk 2]
```diff
@@ -594,6 +594,20 @@ public function inject($object, $asType=null) {
 			}
 		}
 
+		// Populate named methods
+		if (isset($this->specs[$objtype]['methods'])) {
+			foreach ($this->specs[$objtype]['methods'] as $method) {
+				// Ignore any blank entries from the array; these may be left in due to config system limitations
+				if(!$method) continue;
+
+				list($methodName, $methodArgs) = $method;
```
camspiers: I think you should assert that the length is two before doing this.

control/injector/Injector.php
```diff
@@ -594,6 +594,20 @@ public function inject($object, $asType=null) {
...
+				call_user_func_array(
+					array($object, $methodName),
```
camspiers: This assumes the method is public. Are we happy with the default error that will be generated?

control/injector/Injector.php
```diff
@@ -594,6 +594,20 @@ public function inject($object, $asType=null) {
...
+					$this->convertServiceProperty($methodArgs)
```
camspiers: We should assert that methodArgs is an array, otherwise we will get a confusing error from call_user_func_array

camspiers:
Looks good to me. I would recommend calling it 'calls' over 'methods', or maybe even 'methodCalls'. 'calls' would match Symfony too.

control/injector/Injector.php
```diff
@@ -594,6 +594,20 @@ public function inject($object, $asType=null) {
 			}
 		}
 
+		// Populate named methods
+		if (isset($this->specs[$objtype]['methods'])) {
```
camspiers: This will error in the case that `$this->specs[$objtype]` isn't set. Use:

`if (isset($this->specs[$objtype]) && isset($this->specs[$objtype]['methods'])) {`

Or alternatively combine with the block above this block.

control/injector/Injector.php
```diff
@@ -594,6 +594,20 @@ public function inject($object, $asType=null) {
...
+			foreach ($this->specs[$objtype]['methods'] as $method) {
```
camspiers: This makes the assumption that `$this->specs[$objtype]['methods']` is an array. It might not be. Someone might have goofed something up.

control/injector/Injector.php
sminnee: I don't think that's true; I just tested this code and it doesn't error:

    <?php
    $from = array();
    echo isset($from['bad']['to']['worse']) ? "yes\n" : "no\n";

...at least not on PHP 5.4. Has it started erroring on a more recent version?

control/injector/Injector.php
sminnee: I'll see what the error is but that should be fine.

sminnee: Error is warning-level:

`PHP Warning:  call_user_func_array() expects parameter 1 to be a valid callback, cannot access protected method Bla::func()`

It's a bit ganky; I'll reflect & test.

control/injector/Injector.php
sminnee: In any case, I've refactored so it's not an issue.

camspiers: You are right. Don't know what I was thinking :)

sminnee:
Alright, thanks @camspiers for keeping me honest about boundary conditions. ;-) The new code is longer but safer.

I opted to change 'methods' to 'calls' to match Symfony: if in doubt, following an existing implementation.

control/injector/Injector.php
```diff
@@ -585,15 +585,56 @@ public function inject($object, $asType=null) {
...
 		}
 
+		// Populate named methods
+		if (!empty($spec['calls']) && is_array($spec['calls'])) {
+			foreach ($spec['calls'] as $method) {
+				// Ignore any blank entries from the array; these may be left in due to config system limitations
+				if(!$method) continue;
+
+				// Format validation
+				if(!is_array($method) || !isset($method[0]) || isset($method[2])) {
```
camspiers: I think this won't quite work for length 1 arrays. Because it will always error due to the last isset. But besides that, make sure to change the index from 2 to 1

control/injector/Injector.php
```diff
@@ -585,15 +585,56 @@ public function inject($object, $asType=null) {
...
+				}
+				if(!is_string($method[0])) {
+					throw new \InvalidArgumentException("1st element of a 'calls' entry should be a string");
+				}
+				if(isset($method[1]) && !is_array($method[1])) {
+					throw new \InvalidArgumentException("2nd element of a 'calls' entry should an arguments array");
+				}
+
+				// Check that the method exists
+				try {
```
camspiers: It will probably be more performant to do this (reflection is fairly expensive):

```
$objectMethod = array($object, $method[0]);
if (!is_callable($objectMethod)) {
//
}
```

control/injector/Injector.php
camspiers: Would be good to have a unit test to catch this too

camspiers:
This is awesome work @sminnee. I really wanted this back when I was working in SilverStripe (I used workarounds), so I imagine it will be liked by SilverStripe devs.

control/injector/Injector.php
camspiers: Oh! I totally read the code wrong!, I didn't see that they wasn't a `!` on the last isset :) Ignore my previous comments!

sminnee:
Cool - final changes pushed.

Thanks for the feedback, @camspiers — I hope things are going well at Facebook! :-)

sminnee:
OK, squashed and force pushed. Once Travis stops failing randomly, I think this is ready for merge! :-)

Then I can get back to that logger change.

camspiers:
Looks awesome @sminnee :+1: 

Facebook is going well, hit me up next time you travel these ways :)

docs/en/02_Developer_Guides/05_Extending/05_Injector.md
```diff
@@ -111,6 +111,16 @@ Now the dependencies will be replaced with our configuration.
 	echo ($object->textProperty == 'My Text Value');
 	// returns true;
 
+As well as properties, method calls can also be specified:
+
+	:::yml
+	Injector:
+	  Logger:
+	    class: Monolog\Logger
+	    calls:
```
tractorcow: Yeah I think I like 'calls' more too :D

tractorcow:
I've restarted the one build that timed out. Will merge on success.

tractorcow:
If you update this PR, feel free to add `sudo: false` to `.travs.yml` to resolve CI timeout.

tractorcow:
Good job @sminnee. :)

PR #4457, opened 2015-07-31
API Remove SQLQuery

tractorcow:
As SQLQuery was deprecated since 3.2 we can remove it in 4.0


tractorcow:
Rebased.

kinglozzer:
Shouldn’t we also rename `SQLQueryTest` to `SQLSelectTest` for everything in the tests/fixtures etc?

tractorcow:
It's probably not necessary, and would make merging harder. The name doesn't need to match the class exactly. ;) Especially because it makes merging harder...

dhensby:
@tractorcow it should really be renamed; it's a merge pain BUT people need to be able to find tests by the <ClassName>Test(.php) convention

tractorcow:
Rename inc. :D

tractorcow:
Renamed

tractorcow:
test pass, @dhensby :D

PR #4458, opened 2015-07-31
BUG Fix change detection on browser back button

### LINKED ISSUE #4419: Back button doesn't trigger, "are you sure you wish to leave this page?" confirm
* Load admin/pages
 * Select a page
 * Make some changes, but don't press save
 * Press back button

You would expect the message "Are you sure you wish to navigate away?" to appear, but it doesn't. The reason is that this feature is called from `loadPanel()` in `LeftAndMain.js`, and that's not used by the back button.

Marking as critical as this can cause users to lose their work quite easily. The most frequent time this happens is when pressing backspace without a form field highlighted.

tractorcow:
FIxes #4419

Tested in ie8, ie11, safari, chrome.

Only noticeable oddities are where, if you press the back button, the URL of the "back" page is displayed while the confirm dialog is open. If you press "cancel" it restores the current url.

In ie8 I wasn't able to restore the current url properly, but the change detection is safe.

Another limitation is that the "next" browser history reset is overwritten by the restored url, so you can't press back and then forwards. :P Only if you click "no" to losing changes though.

### DIFF
--- a/admin/javascript/LeftAndMain.js
+++ b/admin/javascript/LeftAndMain.js
@@ -253,12 +253,13 @@ jQuery.noConflict();
 				$('body').removeClass('loading');
 				$(window).unbind('resize', positionLoadingSpinner);
 				this.restoreTabState();
-
 				this._super();
 			},
 
 			fromWindow: {
-				onstatechange: function(){ this.handleStateChange(); }
+				onstatechange: function(e){
+					this.handleStateChange(e); 
+				}
 			},
 
 			'onwindowresize': function() {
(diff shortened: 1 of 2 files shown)


admin/javascript/LeftAndMain.js
```diff
@@ -377,18 +405,9 @@ jQuery.noConflict();
 				if(!title) title = "";
 				if (!forceReferer) forceReferer = History.getState().url;
 
-				// Check change tracking (can't use events as we need a way to cancel the current state change)
-				var contentEls = this._findFragments(data.pjax ? data.pjax.split(',') : ['Content']);
```
sminnee: Why have we dropped the data.pjax reference?

tractorcow: It took me a while to figure out why this was necessary, will restore this. ;)

PR #4462, opened 2015-08-01
NEW: Add support for ‘%%remove%%’ market in Yaml Config.

sminnee:
Sometimes, it’s not enough to set a config option to null. For example,
you may configure a list of items as a constructor property setter and
it won’t expect empty items in the string.

This change introduces a special string that you can use to mark config
options that you would like to have removed with unset().

Note that this means we can’t set a configuration option to the string
“%%remove%%”; hence the reason for using a slightly convoluted string.

@camspiers @nyeholt @stevie-mayhew @tractorcow, this flows on pretty
directly from my work on Injector, so I'd be keen to get your views.

@hafriedlander I'm nervous about making any changes to Config without
a thumbs-up from you. ;-) I don't think this will have too much performance
impact.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,8 +37,6 @@ matrix:
       env: DB=MYSQL
     - php: 5.4
       env: DB=MYSQL BEHAT_TEST=1
-    - php: 5.3
-      env: DB=MYSQL
     - php: 7.0
       env: DB=MYSQL
     - php: nightly
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
 		}
 	],
 	"require": {
-		"php": ">=5.3.3",
+		"php": ">=5.4.0",
 		"composer/installers": "~1.0"
 	},
 	"require-dev": {
--- a/core/Config.php
+++ b/core/Config.php
@@ -144,6 +144,17 @@ class Config {
 	const IS_ARRAY = 2;
 
 	/**
+	 * Return flag for get_value_type indicating value is a remove marker
+	 * @const
+	 */
+	const IS_REMOVE_MARKER = 3;
+
+	/**
+	 * The string used as a marker for remove actions
+	 */
+	const REMOVE_MARKER_STRING = '%%remove%%';
+
+	/**
 	 * Get whether the value is an array or not. Used to be more complicated,
 	 * but still nice sugar to have an enum to compare and not just a true /
 	 * false value.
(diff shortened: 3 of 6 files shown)


sminnee:
...ok, so I've used PHP 5.4's array style on this, because we've talked about dropping PHP 5.3 support for SilverStripe 4. As such, I've removed the travis build config for 5.3.

nyeholt:
Am away from dev system at the moment so I can't checkout the code itself, but... a long long while ago, Hamish and I put some code together to allow the complete overriding of configuration blocks rather than merging which is the current behaviour. It never ended up being merged back to core (that I'm aware of anyway!), but I'm wondering how this would work to possibly provide something like that anyway. What happens if you have a `property: %%remove%%` followed by another definition later on based on the Before/After rules? 

Anyway - this kind of feels like a hatchet approach to the problem, but having dived into the depths of the Config system before, sometimes a hatchet can be the appropriate tool for the job :)

sminnee:
I'll add some tests, but it should be fine to add, remove, add again, as long as they're separate blocks.

The complete overriding of blocks might be useful in some cases, but the particular issue I was grappling with was manipulating maps and arrays (eg disabling monolog handlers). Although you could start from scratch with a block override, you couldn't remove one handler. Not sure how problematic this is in real life, though.

It would be good to dig up your config-replacement work in progress; seems like now would be a good time to get it into master.

tractorcow:
Hi @sminnee, sorry for not responding to your pull request in a timely fashion.

`%%remove%%` solves the specific problem of not being able to remove config, but I think that the majority of cases where the system fails short is being able to replace the block.

I would rather see a good `this config is a replacement` syntax with the ability to specify a blank block than one that only supports removal.

For instance, say we are using a `!` suffix for config which should override the existing block. On an existing set of options, each of which takes an arbitrary non-associative list.

E.g.

```yml
MyClass:
  adds:
    - newoption
  replaces!:
    - replacedoptions
  clears!: []
```

sminnee:
Thanks, Damian. I tend to agree and I think that Hamish had been working on something like that.

The only thing about your clear syntax is that it doesn't actually delete the setting: it changes it to an empty array. Rather than a `!` suffix I was wondering about having different suffixes for different operations, e.g. `(replace)` and `(remove)`.

The blocker for that was apparently that we can't currently cache the parsing of those operations and Hamish wanted to refactor things.

tractorcow:
```yaml
MyClass:
  adds:
    - newoption
  replace!replace:
    - replacedoptions
  clears!remove
```

Something like that. Is that last line valid yml?

dhensby:
Will this be resurrected? how about just creating a yml doc solely for removals?

```yml
---
name: removals
type: unset
--
MyClass:
  extensions:
    - ExtensionName
```

sminnee:
Damian's 2nd suggestion would probably be my preference there.

sminnee:
As part of incorporation of @micmania1's new config system into SS4, @tractorcow has been investigating another approach: interpreting a config value of `null` as removing the config value.

dhensby:
I like that idea

tractorcow:
You can now use `null` to reset a value.

You can also remove keys from array with:

```
---
Name: original
---
Class:
  data:
    - one
    - two
---
Name: override1
---
Class:
  data: null
---
Name: override2
---
Class:
  data:
    - one
    - three
```

Granted it's a bit annoying to use a separate block to first reset, and then another to re-assign array values, but at least it's possible. @flamerohr has tested this out and confirmed it works. :)

Does this meet the need for the mean time?

PR #4466, opened 2015-08-02
Stray backslash in get_cms_login_form argument ?

SpiritLevel:


sminnee:
It's not a stray backslash, it ensure that the top-level namespace is referenced. It's not much of an issue right now but it may become one in the future...

SpiritLevel:
Thanks @sminnee. It was inconsistent with the method get_login_form above it so thought it might be a problem.

PR #4471, opened 2015-08-03
API batch restore action

tractorcow:
Localisation for https://github.com/silverstripe/silverstripe-cms/pull/1255


tractorcow:
Thanks sam

PR #4473, opened 2015-08-03
Check that method 'hasMethod' exists in GridFieldSortableHeader

kinglozzer:
In some specific cases, this can cause a fatal error (`Call to a member function hasMethod() on a non-object`) which is hard to diagnose (this came up on IRC).

Example code is:
```php
private static $has_one = array
    'Image' => 'Image'
);
private static $summary_fields = array(
		'Image.CMSThumbnail.Tag' => 'Image'
);
```

This code won’t present the tag anyway (just escaped HTML for it), but before this change it gives the fatal error that a beginner wouldn’t be able to understand.


dhensby:
Yes, this is a component that needs some love and this is a good place to start.

PR #4477, opened 2015-08-04
Fixes #4424

### LINKED ISSUE #4424: Model Admin : 'Show Specification for' toggle link breaks if Title contains spaces
I noticed that when I set the Title for an object to contain spaces the  'Show Specification for' toggle link no longer works as it uses the ModelName as part of the anchor URL. 

I can't see where this toggle link would appear twice so you can either remove the ModelName from the template (\framework\admin\templates\Includes\ModelAdmin_ImportSpec.ss) or add something in to remove spaces.

nickbooties:
Model Admin : 'Show Specification for' toggle link breaks if Title contains spaces #4424

### DIFF
--- a/admin/code/ModelAdmin.php
+++ b/admin/code/ModelAdmin.php
@@ -366,7 +366,7 @@ abstract class ModelAdmin extends LeftAndMain {
 			$specRelations->push(new ArrayData(array('Name' => $name, 'Description' => $desc)));
 		}
 		$specHTML = $this->customise(array(
-			'attrModelName' => str_replace(' ', '',$modelName),
+			'ClassName' => str_replace('\\', '_', $className),
 			'ModelName' => Convert::raw2att($modelName),
 			'Fields' => $specFields,
 			'Relations' => $specRelations, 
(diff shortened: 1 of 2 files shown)


sminnee:
Thanks for providing a patch, @nickbooties! I see that this is your first pull request to our project; it's great that you can become a contributor to SilverStripe.

I can see that this follows the approach recommended by @kinglozzer, but I'm a little concerned that this is not a broad enough; what if another illegal character was introduced, such a quote? When fixing a bug, it's good to fix not only the specific issue raised, but any related issues that might come up. In this case, I think that basing an HTML ID on text intended to be human-readable isn't the best approach.

I think that it might be better to use `$modelClass` to build the ID, instead. Since this will follow the rules of PHP class names, we can be more confident that it will only contain safe characters. However, in case people have used namespaces, it's probably worth transforming it like this first:

    str_replace('\\', '_', $modelClass);

Also, once you've fixed, can you please "squash" your changes together to make a single commit, and provide a more descriptive commit message, that describes the issue in a few words, as well as linking to #4424.

There's a tutorial on squashing commits here: http://davidwalsh.name/squash-commits-git

nickbooties:
Ahh I was wondering how to bundle all my edits into a single commit, thanks!

sminnee:
Congratulations - you're now officially a contributor to SilverStripe :-)

PR #4478, opened 2015-08-04
Remove PHP 5.3 from Travis suite, updating minimum version requirement.

sminnee:
This PR is intended to be the official removal of PHP 5.3 support from master (which will become SilverStripe 4).

Assuming that we don't have any objections on https://groups.google.com/forum/#!topic/silverstripe-dev/FR_d5QVjXe4 by 7 August, it should be good to merge.

If we get objections, we should probably discuss whether we listen to them. We'd need a pretty darn good reason to retain 5.3 support in ss4.

Thanks,
Sam

PS: There's a lot more discussion about this in #4463

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,8 +37,6 @@ matrix:
       env: DB=MYSQL
     - php: 5.4
       env: DB=MYSQL BEHAT_TEST=1
-    - php: 5.3
-      env: DB=MYSQL
     - php: 7.0
       env: DB=MYSQL
     - php: nightly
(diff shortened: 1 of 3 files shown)


nfauchelle:
@sminnee 
I am all in favour of this, but should it be changed to 5.5 given that 5.4 is already out of Active Support and Security Support ends middle of September....

tractorcow:
Seconded!

sminnee:
Ok the date has passed - does someone want to merge this? We can push out 5.4 if needs be later but I want to start being able to accept 5.3-breaking changes.

kinglozzer:
atlast

PR #4479, opened 2015-08-04
Fix datefield to work with other form markup

hex0id:
Changed onClick selector to catch any date field, so it will work with "nonstandard" form markups like adding new lines from https://github.com/silverstripe-australia/silverstripe-gridfieldextensions

### DIFF
--- a/javascript/DateField.js
+++ b/javascript/DateField.js
@@ -26,7 +26,7 @@
 		}
 	});
 
-	$(document).on("click", ".field.date input.text,.fieldholder-small input.text.date", function() {
+	$(document).on("click", "input.text.date", function() {
 		$(this).ssDatepicker();
 
 		if($(this).data('datepicker')) {
--- a/javascript/DateField.js
+++ b/javascript/DateField.js
@@ -26,7 +26,7 @@
 		}
 	});
 
-	$(document).on("click", "input.text.date", function() {
+	$(document).on("click", ".field.date input.text,input.text.date", function() {
 		$(this).ssDatepicker();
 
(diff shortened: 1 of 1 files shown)


chillu:
Given `DateField.js` is automatically included if used in forms for your own website, this change has a wider impact than just fixing CMS behaviour. So we need to be careful with behaviour changes here. The proposed selector change also removes `.field.date input.text`, replacing it with `input.text.date` - different semantics. I would propose the selector `.field.date input.text,input.text.date` instead to maintain backwards compatibility, so effectively just removing `.fieldholder-small`. What do you think?

hex0id:
actually initial change was just as you propose, so I absolutely agree.
thank you Ingo.

On Thursday, August 27, 2015, Ingo Schommer <notifications@github.com>
wrote:

> Given DateField.js is automatically included if used in forms for your
> own website, this change has a wider impact than just fixing CMS behaviour.
> So we need to be careful with behaviour changes here. The proposed selector
> change also removes .field.date input.text, replacing it with
> input.text.date - different semantics. I would propose the selector .field.date
> input.text,input.text.date instead to maintain backwards compatibility,
> so effectively just removing .fieldholder-small. What do you think?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/silverstripe/silverstripe-framework/pull/4479#issuecomment-135287519>
> .
>


-- 
Regards, Alex.

chillu:
@hex0id Do you want to update this pull request?

hex0id:
@chillu done

PR #4480, opened 2015-08-04
FIX extraClass() method to match parent method

xini:
the method name is wrong and doesn't work.

### DIFF
--- a/forms/CompositeField.php
+++ b/forms/CompositeField.php
@@ -123,8 +123,8 @@ class CompositeField extends FormField {
 		return $this->legend;
 	}
 
-	public function extraClasses() {
-		$classes = array('field', 'CompositeField', parent::extraClasses());
+	public function extraClass() {
+		$classes = array('field', 'CompositeField', parent::extraClass());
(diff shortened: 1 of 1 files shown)


xini:
this will also have to be applied to 3.2 and master.

willmorgan:
This is an API change and therefore will have to be submitted against 3, I'm afraid.

xini:
really? that fixes a bug that's been there since 2011... which branch do you want it to be rebased to? master?

dhensby:
That function definitely can't just be removed in 3.1 - people can easily be using it (isn't it being used in templates already?!)

I'd agree it's a bug and should be added in, you'd need to deprecate the old method though (not just remove it).

tractorcow:
@xini if you want to rebase against 3 branch we can merge it there, but you'll need to add a stub to prevent regressions.


```php
/**
 * @deprecated
 */
public function extraClasses() {
	Deprecation::notice('4.0', 'Use extraClass() instead');
	return $this->extraClass();
}
```

xini:
rebased against 3, see #4873

Issue #4474, opened 2015-08-03
DataQuery passes classes that don’t exist to ClassInfo::subclassesFor()


kinglozzer:
[See here](https://github.com/silverstripe/silverstripe-framework/blob/ffbeac6b7d3ed1a6e02a150573ee4b2f9251cf9c/model/DataQuery.php#L208). If there is a collision in a join table (e.g. `SiteTree.Sort` colliding with `Page_ManyManyRelation.Sort`), then `Page_ManyManyRelation` is passed to `ClassInfo::subclassesFor()`, even though no PHP class exists for it.

Simplest fix is to amend the line linked above to:
```php
if (class_exists($collisionBase)) {
    $collisionClasses = ClassInfo::subclassesFor($collisionBase);
} else {
    $collisionClasses = array($collisionBase => $collisionBase);
}
```

kinglozzer:
Now I look at this again, I don’t see how this has ever worked. Perhaps there’s another regression I’m missing... will dig further tomorrow.

kinglozzer:
Okay this is seems to be a valid bug. The SQL that _was_ produced is valid, but doesn’t really make any sense. Pseudo-code until I get time to write up a test case:

```php
class Page extends SiteTree {
    private static $many_many = array(
        'Slides' => 'Slide'
    );

    private static $many_many_extraFields = array(
        'Slides' => array(
            'Sort' => 'Int'
        )
   );

    public function getCMSFields() {
        $fields = parent::getCMSFields();
        $fields->addFieldToTab('Root.Slides', GridField::create('Slides', 'Slides', $this->Slides(), GridFieldConfig_RecordEditor::create()));
        return $fields;
    }
}

class Slide extends DataObject {
    private static $db = array(
        'Title' => 'Varchar'
    );

    private static $belongs_many_many = array(
        'Pages' => 'Page'
    );
}
```

The error happens when you then try to edit a slide. Seems to be triggered by a `GridField` component, as adding `$fields->removeByName('Pages');` to the slide stops it happening (GF is not the cause of the issue, just what’s triggering it in this case).

Tacking on showqueries shows that the _old_ SQL produced included:

```sql
SELECT DISTINCT CASE
  WHEN "SiteTree"."ClassName" IN ('Page_Slides')
    THEN "Page_Slides"."Sort"
[...]

PR #4482, opened 2015-08-04
Ensure ClassInfo is backwards compatible with non-existant classes

kinglozzer:
See [this comment](https://github.com/silverstripe/silverstripe-framework/pull/3949#issuecomment-127300189) and #4474 for context.

Of course it doesn’t make any sense to pass non-existant classes to `ClassInfo`, but we have supported it so unfortunately we probably need to until 4.0. The example from #4474 is that `ClassInfo::subclassesFor('IDontExist')` used to return `array('IDontExist'=>'IDontExist')` but now triggers a fatal error.

Not marked as a fix for #4474 because it will need to be fixed another way.

### DIFF
--- a/core/ClassInfo.php
+++ b/core/ClassInfo.php
@@ -159,6 +159,13 @@ class ClassInfo {
 	public static function class_name($nameOrObject) {
 		if (is_object($nameOrObject)) {
 			return get_class($nameOrObject);
+		} elseif (!self::exists($nameOrObject)) {
+			Deprecation::notice(
+				'4.0',
+				"ClassInfo::class_name() passed a class that doesn't exist. Support for this will be removed in 4.0",
+				Deprecation::SCOPE_GLOBAL
+			);
+			return $nameOrObject;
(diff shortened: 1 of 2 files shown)


core/ClassInfo.php
```diff
@@ -159,6 +159,13 @@ public static function subclassesFor($class) {
 	public static function class_name($nameOrObject) {
 		if (is_object($nameOrObject)) {
 			return get_class($nameOrObject);
+		} elseif (!class_exists($nameOrObject)) {
```
dhensby: should we used `self::exists()` here?

kinglozzer: Makes sense, wasn’t even aware of that method! Have changed that now

dhensby:
@kinglozzer do you want to get a test into this?

kinglozzer:
@dhensby Sure, added one :)

sminnee:
Thanks @kinglozzer! :-)

Issue #4271, opened 2015-06-11
Int can't be a class name in php 7


g4b0:
Trying to run SS 3.1.13 on a php7 box I get this error:

Fatal error: "int" cannot be used as a class name in /vagrant/tribedistribution.zk/framework/model/fieldtypes/Int.php on line 8

tractorcow:
Possibly we'll need to resolve this in 4.x when we move to namespaces.

Can SilverStripe/Framework/Int be a namespaced class?

simonwelsh:
No. Classes can't be named after keywords.

kinglozzer:
Same issue applies to `float` keyword. How about `SilverStripe/Framework/Integer` and `SilverStripe/Framework/FloatingPoint`?

It’s either that or `SilverStripe/Framework/Floater` :hankey:

jonom:
> It’s either that or SilverStripe/Framework/Floater :hankey:

First time I've ever LOL'd at a github notification

dhensby:
`SilverStripe/DBField/Int` is more likely to be the convention. Amirite @sminee @assertchris? 

If we can't even use int in a namespaced class, then we are going to have a bad time.

kinglozzer:
@dhensby Yeah my examples were just rough, unfortunately we can't use int or float (or string or bool, but we don't use those anyway), even when namespaced. :(

tractorcow:
`SS_Int`? Simon already pointed out you cant have a classname that's a keyword, even if it's in a namespace.

kinglozzer:
It’ll probably come down to personal preference, my preference being to ditch the `SS_` prefix and name them something else - IIRC that prefix was only really added to work around namespacing. We’d end up with even less consistency:

- `SS_Int`
- `SS_Float`
- `Varchar`
- `Boolean`

`Int` -> `Integer` could work, but I don’t know what we’d do with `Float`. Unless we map `FieldType` to `SS_FieldType` internally?

stevie-mayhew:
I think that re-classing them all as `SS_*` makes sense - we already have `SS_Datetime` which follows this convention? As you say though, personal preference :wink:

jonom:
If they're all subclasses of DBField, I would prefer a prefix of DB than SS and use it for all of them for consistency.
DBInt
DBDate
DBVarchar etc.
Looks like there is already a `DBLocale` so there is a precedent :grinning:

simonwelsh:
[...]

PR #4485, opened 2015-08-05
Explicitly don't support PHP7

yabbagail:
SilverStripe doesn't work in PHP7 (see #4271 at least), so have composer reflect that.

Should probably be merged up to master as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ env:
 matrix:
   allow_failures:
     - php: hhvm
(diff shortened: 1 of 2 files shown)


sminnee:
I'm happy with this for a 3.x change. For 4.0 I'd prefer our efforts to go into fixing it. :-)

yabbagail:
That is a better option. Will leave it up to the person merging upwards :)

PR #4486, opened 2015-08-05
BUGFIX. FulltextFilter requires table identifiers in match query

### LINKED ISSUE #4483: BUG FulltextFilter requires table identifiers in match query
See https://github.com/silverstripe/silverstripe-framework/pull/3554#issuecomment-127612834

We should use `MATCHES("Table"."Column")` syntax rather than matching against only column, as we could run into ambiguous column error.

uniun:
Fixes #4483

### DIFF
--- a/search/filters/FulltextFilter.php
+++ b/search/filters/FulltextFilter.php
@@ -58,12 +58,12 @@ class FulltextFilter extends SearchFilter {
 		if(is_array($indexes) && array_key_exists($this->getName(), $indexes)) {
 			$index = $indexes[$this->getName()];
 			if(is_array($index) && array_key_exists("value", $index)) {
-				return $index['value'];
+				return $this->prepareColumns($index['value']);
 			} else {
 				// Parse a fulltext string (eg. fulltext ("ColumnA", "ColumnB")) to figure out which columns
 				// we need to search.
 				if(preg_match('/^fulltext\s+\((.+)\)$/i', $index, $matches)) {
-					return $matches[1];
+					return $this->prepareColumns($matches[1]);
 				} else {
 					throw new Exception("Invalid fulltext index format for '" . $this->getName()
 						. "' on '" . $this->model . "'");
@@ -72,6 +72,19 @@ class FulltextFilter extends SearchFilter {
 		}
 
 		return parent::getDbName();
+	}
+	
+	private function prepareColumns($columns) {
+		$class = '';
+		$parts = explode(',', $columns);
+		array_walk($parts, function(&$col, $key) use (&$class) {
+			$name = trim($col, " \t\n\r\0\x0B\"");
+			if (empty($class)) {
+				$class = ClassInfo::table_for_object_field($this->model, $name);
+			}
+			$col = sprintf('"%s"."%s"', $class, $name);
+		});
+		return implode(',', $parts);
 	}
 
 }
--- a/search/filters/FulltextFilter.php
+++ b/search/filters/FulltextFilter.php
@@ -73,18 +73,20 @@ class FulltextFilter extends SearchFilter {
 
 		return parent::getDbName();
 	}
-	
-	private function prepareColumns($columns) {
-		$class = '';
-		$parts = explode(',', $columns);
-		array_walk($parts, function(&$col, $key) use (&$class) {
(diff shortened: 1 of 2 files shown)


search/filters/FulltextFilter.php
```diff
@@ -73,5 +73,18 @@ public function getDbName() {
 
 		return parent::getDbName();
 	}
+	
+	private function prepareColumns($columns) {
```
willmorgan: PHPDoc here, please.

search/filters/FulltextFilter.php
```diff
@@ -73,5 +73,18 @@ public function getDbName() {
...
+		$class = '';
+		$parts = explode(',', $columns);
+		array_walk($parts, function(&$col, $key) use (&$class) {
```
willmorgan: It might be clearer to just use `array_map` here instead of getting messy with references, which are rather error prone.

search/filters/FulltextFilter.php
```diff
@@ -73,5 +73,18 @@ public function getDbName() {
...
+			$name = trim($col, " \t\n\r\0\x0B\"");
```
willmorgan: I wonder if there is a consistently used constant or some reusable resource that contains all possible characters to trim here, rather than having them specified as a string.

search/filters/FulltextFilter.php
```diff
@@ -73,5 +73,18 @@ public function getDbName() {
...
+			if (empty($class)) {
```
willmorgan: Nitpicking, but would it be safer for us to ask `if($class == '')` instead?

willmorgan:
Thanks for having a go at this. I've restarted the Travis builds to see whether tests pass, as I noticed an error with the first run.

willmorgan:
It appears the tests need to be updated slightly, but other than that, the code is covered! :+1: 

```
There was 1 failure:
1) FulltextFilterTest::testGenerateQuery
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'"ColumnA", "ColumnB"'
+'"FulltextFilterTest_DataObject"."ColumnA","FulltextFilterTest_DataObject"."ColumnB"'
/home/travis/builds/ss/framework/tests/search/FulltextFilterTest.php:48
```

search/filters/FulltextFilter.php
dhensby: If this is going to get into a patch release, it really shouldn't be anything other than private??

dhensby:
Some comments in the code would be really useful to understand what's going on.

search/filters/FulltextFilter.php
dhensby: `empty` is fine surely? I think the only other option would be `$class === ''` but I don't think that adds any value?

search/filters/FulltextFilter.php
tractorcow: Protected is fine; It just means no new public api.

tractorcow:
Please fix the tests, and I concur with @willmorgan on `protected` method with proper phpdoc.

At the very least, even private methods need phpdoc. :)

search/filters/FulltextFilter.php
tractorcow: Also check out DBSchemaManager::explodeColumnString for an existing implementation of this index spec parsing.

I wish I made that public... oh well. :)

dhensby:
bump @uniun

uniun:
Sorry for the delay :)
I've rewritten the function according to your feedback. I hope it'll be OK now.

willmorgan:
It looks like tests are still failing, @uniun :cry:

tractorcow:
Code is excellent. You just need to fix the additional space in the unit tests.

uniun:
Damn whitespaces! :) FIxed!

dhensby:
aaaaaaaaaand it's gone

dhensby:
thanks!

willmorgan:
We not waiting for squashes anymore?

tractorcow:
@dhensby is ment to squash it himself if he merges it. :D

dhensby:
annnnnnnnnd squashed commits are so yesterday

uniun:
@tractorcow don't forget to close #4483 ;)

PR #4491, opened 2015-08-05
remove tailing spaces in the codebase

spekulatius:
just removing some tailing spaces...

### DIFF
--- a/control/Cookie_Backend.php
+++ b/control/Cookie_Backend.php
@@ -15,7 +15,7 @@ interface Cookie_Backend {
 	 * @param array $cookies The existing cookies to load into the cookie jar
 	 */
 	public function __construct($cookies = array());
-	
+
 	/**
 	 * Set a cookie
 	 *
@@ -28,7 +28,7 @@ interface Cookie_Backend {
 	 * @param boolean $httpOnly Prevent the cookie being accessible by JS
 	 */
 	public function set($name, $value, $expiry = 90, $path = null, $domain = null, $secure = false, $httpOnly = true);
-	
+
 	/**
 	 * Get the cookie value by name
 	 *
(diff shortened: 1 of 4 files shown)


sminnee:
Hey @spekulatius thanks for tidying these up! I've noticed a few of these PRs coming through, a few files at a time. Perhaps it would be worth having a more complete go at removing these? 

Also, I don't know if we want to make use of Scrutinizer to keep these out of the codebase in the future?

spekulatius:
Hey Sam,

I could do all at once. But this would probably cause a lot of grieve later on with merging pull requests - either while rebasing or merging in there will be plenty conflicts... This doesn't seem ideal.

Setting up Scruntinizer -or any other capable- solution would be good. But Scrutinizer isn't free afaik. Jenkins could do... I'll see if I can fit it on my servers somewhere... I'll come back to you for this.

Peter

dhensby:
Scrutinizr is free for OSS I think Sam's question is more about whether we can configure it to dislike training whitespace. 

When I merge up I tend to ignore space changes otherwise it becomes a nightmare, so it's not a big deal but means some whitespace amends don't get through.

spekulatius:
Ah, okay, I haven't read the pricing further after I saw there isn't a free option...  Using PHPCS we should be able to get notifications about tailing spaces. 

Here you find information about the configuration:
https://scrutinizer-ci.com/docs/tools/php/code-sniffer/

This ruleset should contain a rule for disallowing tailing spaces:
https://github.com/spekulatius/silverstripe-codesniffer

dhensby:
Apparently our scrutinizr already has this setting:

spekulatius:
mmm, interesting. Seems not really to apply - or all of those trailing whitespaces are from the time before Scrutinizr was used?!

PR #4496, opened 2015-08-07
MINOR: be more specific with background class for menu

stevie-mayhew:
See https://github.com/Rhym/silverstripe-cms-theme/pull/6


sminnee:
This seems reasonable but I'm curious about what problems this is fixing, exactly?

adrexia:
It means that if you are trying to override other background settings later with css (size, position, repeat etc), and not trying to override the entire background image, you can do so without using an !important. Looks like it's a direct response to the cms theming modules some folk are building.

sminnee:
Cool - thanks for the context, as well as the change! :-)

stevie-mayhew:
Thanks for putting that more concisely than I could have @adrexia - and thanks for the merge @sminnee :smile:

PR #4497, opened 2015-08-07
FIX: Fixed regression from ClassInfo case-sensitivity fix.

sminnee:
This fixes a bug introduced by ffbeac6b7d3ed1a6e02a150573ee4b2f9251cf9c.
ClassInfo::subclassesFor() didn't previously throw an Exception if passed
an invalid class; it just returned no values. This will annoy minor-release
upgrades, and so I've made it return null instead in these situation.

### DIFF
--- a/core/ClassInfo.php
+++ b/core/ClassInfo.php
@@ -61,6 +61,8 @@ class ClassInfo {
 	 * @return array List of subclasses
 	 */
 	public static function getValidSubClasses($class = 'SiteTree', $includeUnbacked = false) {
+		if(is_string($class) && !class_exists($class)) return null;
+
 		$class = self::class_name($class);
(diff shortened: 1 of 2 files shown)


kinglozzer:
This needs to be merged into 3.1 too, and is a blocker for 3.1.14 IMO - #4482

sminnee:
Good reason for be more conservative with what we back port - the original change was not a low risk change.

Also - we must lack test coverage of member querying... :-/ I'll see about making a suitable test.

kinglozzer:
Yeah, I tried to address these two issues separately (the BC-break in `ClassInfo` behaviour, and the `DataQuery` issue) but I couldn’t manage to capture the `DataQuery` stuff in a test

PR #4498, opened 2015-08-09
BUG Fix duplicate primary key crash on duplicate

tractorcow:
Duplicating objects should clear the version ID


tractorcow:
ta :D

dhensby:
Hmm, when an object is duplicated, do you want to reset it's history or copy it? 

Could be useful to copy, but would add some DB bloat...

PR #4499, opened 2015-08-11
Adding nl_BE and fr_BE as common_locales

jpvdpoel:
In order to make them selectable as language in admin.

### DIFF
--- a/i18n/i18n.php
+++ b/i18n/i18n.php
@@ -1100,6 +1100,10 @@ class i18n extends Object implements TemplateGlobalProvider, Flushable {
 			'name' => 'Dutch',
 			'native' => 'Nederlands'
 		),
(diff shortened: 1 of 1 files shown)


dhensby:
Please can you add them in the "correct" place in the array (ie: in order next to "french" and "dutch") rather than right at the end?

jpvdpoel:
Ahh, my bad, changed it.

willmorgan:
These commits now need to be squashed

dhensby:
@jpvdpoel are you able to squash these commits?

tractorcow:
Squashed and merged with https://github.com/silverstripe/silverstripe-framework/commit/5dbb9c1da087bb603481366b6f70ce02977013b6

PR #4500, opened 2015-08-14
FEATURE: implement getter and setter usage for response

stevie-mayhew:
I had a problem on a project where something weird was happening, so I've done this.

### DIFF
--- a/admin/code/LeftAndMain.php
+++ b/admin/code/LeftAndMain.php
@@ -227,7 +227,7 @@ class LeftAndMain extends Controller implements PermissionProvider {
 
 		// Allow customisation of the access check by a extension
 		// Also all the canView() check to execute Controller::redirect()
-		if(!$this->canView() && !$this->response->isFinished()) {
+		if(!$this->canView() && !$this->getResponse()->isFinished()) {
 			// When access /admin/, we should try a redirect to another part of the admin rather than be locked out
 			$menu = $this->MainMenu();
 			foreach($menu as $candidate) {
@@ -445,8 +445,10 @@ class LeftAndMain extends Controller implements PermissionProvider {
 			$msgs = _t('LeftAndMain.ValidationError', 'Validation error') . ': '
 				. $e->getMessage();
 			$e = new SS_HTTPResponse_Exception($msgs, 403);
-			$e->getResponse()->addHeader('Content-Type', 'text/plain');
-			$e->getResponse()->addHeader('X-Status', rawurlencode($msgs));
+			$errorResponse = $e->getResponse();
+			$errorResponse->addHeader('Content-Type', 'text/plain');
+			$errorResponse->addHeader('X-Status', rawurlencode($msgs));
+			$e->setResponse($errorResponse);
 			throw $e;
 		}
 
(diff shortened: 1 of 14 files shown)


willmorgan:
`getResponse` is a good idea, but its implementation in a simple find-and-replace isn't the best here - especially if an implementor of that method performs some work every time it is called. It would make more sense to pull that out in to a variable for re-use throughout each function that uses it.

control/Controller.php
```diff
@@ -212,9 +211,21 @@ public function getURLParams() {
...
 		return $this->response;
 	}
 
+	/**
+	 * Sets the SS_HTTPResponse object that this controller is building up.
+	 *
+	 * @param SS_HTTPResponse $response
+	 */
+	public function setResponse(SS_HTTPResponse $response) {
+		$this->response = $response;
```
dhensby: return $this

chillu:
@stevie-mayhew On your question if we should use `$response = getResponse()` for `$this->getResponse()->addHeader(...)`, I'd say just in case there's more than one use in the same function scope - otherwise it's fine to chain.

dhensby:
@stevie-mayhew can we get the feedback done; then we can merge :)

stevie-mayhew:
@dhensby cool - have done.

The way which the response is handled in `handleRequest` in `LeftAndMain.php` is mighty confusing looking now this has been done.

stevie-mayhew:
Whoops, broke the build a few times. Sorry :+1:

admin/code/LeftAndMain.php
```diff
@@ -455,9 +457,11 @@ public function handleRequest(SS_HTTPRequest $request, DataModel $model = null)
 		if(!$response->getHeader('X-Title')) $response->addHeader('X-Title', urlencode($title));
 
 		// Prevent clickjacking, see https://developer.mozilla.org/en-US/docs/HTTP/X-Frame-Options
-		$this->response->addHeader('X-Frame-Options', 'SAMEORIGIN');
-		$this->response->addHeader('Vary', 'X-Requested-With');
-		
+		$originalResponse = $this->getResponse();
+		$originalResponse->addHeader('X-Frame-Options', 'SAMEORIGIN');
+		$originalResponse->addHeader('Vary', 'X-Requested-With');
+		$this->setResponse($originalResponse);
```
dhensby: You don't need to reassign the response back as its an object so it'll be a reference, not a copy

stevie-mayhew: Good point.

sminnee:
OK it looks like @dhensby's points have been covered, I'm happy with the change, and the tests pass—merging.

control/Controller.php
```diff
@@ -119,7 +119,6 @@ public function handleRequest(SS_HTTPRequest $request, DataModel $model) {
 		$this->pushCurrent();
 		$this->urlParams = $request->allParams();
 		$this->setRequest($request);
-		$this->response = new SS_HTTPResponse();
```
dhensby: OK - so this is a potential regression here as people are likely to require this property to be populated as part of `onBeforeInit`, `init` and `onAfterInit`

Issue #4502, opened 2015-08-14
Ability to sort search results from GridFieldAddExistingAutocompleter


patricknelson:
This `->doSearch()` method is pretty configurable, but currently it's not possible to sort results coming from it, see this line: 

https://github.com/silverstripe/silverstripe-framework/blob/3.2/forms/gridfield/GridFieldAddExistingAutocompleter.php#L214

Would it make sense to set this up via setter/getter so that it's configurable, similar to the `->resultsFormat` property? I currently have the need to set this up so that it will return a list of articles in descending order by date (even though it's searching the title). In this case, it would make sense to add a single new method either `->sort()` or `->setSort()` which will be the argument[s] passed through to the `->sort()` method on the `DataList` instance in `->doSearch()`.

patricknelson:
Hah, amazingly, in all my work in trying to fix this server-side using my own non-interfering code (i.e. via `Injector`), I end up realizing it's all for nothing. This is because it will return a JSON object with properties being the `ID` of the `DataObject` and `Title` as the value. Since you have a JSON object which is supposed to be "unordered" (and not an array which is ordered), JSON.parse() will order it by the object keys (the `ID` field on the DataObject).

So, 

```json
{
  "321": "Item 1",
  "123": "Item 2"
}
```

Ends up getting brutalized by `JSON.parse()` into:

```json
{
  "123": "Item 2",
  "321": "Item 1"
}
```

:(

patricknelson:
Unfortunately a framework level fix for something like this may be a bit too much work for me to handle. I've easily fixed it by simply doing this in `GridField.js`

```js
	success: function(data) {
		response(JSON.parse(data));
	},
```

And then modifying the PHP JSON return to already be formatted properly so that `JSON.parse()` doesn't modify the ordering, but this will break the unit tests:

```php
			$title = html_entity_decode(SSViewer::fromString($this->resultsFormat)->process($result));
			$json[] = array(
				"label" => $title,
				"value" => $title,
				"id" => $result->ID,
			);
```

So I'm afraid fulfilling client request here means making an incompatible framework change.

maxime-rainville:
Unfortunately, [Silverstripe CMS 3 has entered limited support in June 2018](https://www.silverstripe.org/blog/update-on-silverstripe-5-x/). This means we'll only be fixing critical bugs and security issues for Silverstripe CMS 3 going forward.

[...]

PR #4503, opened 2015-08-14
FIX for #4502 Prevents JSON.parse() from scrambling sorted results from server-side.

patricknelson:
This at least makes it possible to take results which were already sorted on the server and ensures they are displayed in the same order. See #4502 for a breakdown of what happens when JSON.parse() encounters an object which is 'unsorted' vs. an array, which is sorted.

### DIFF
--- a/forms/gridfield/GridFieldAddExistingAutocompleter.php
+++ b/forms/gridfield/GridFieldAddExistingAutocompleter.php
@@ -232,7 +232,12 @@ class GridFieldAddExistingAutocompleter
 		$originalSourceFileComments = Config::inst()->get('SSViewer', 'source_file_comments');
 		Config::inst()->update('SSViewer', 'source_file_comments', false);
 		foreach($results as $result) {
-			$json[$result->ID] = html_entity_decode(SSViewer::fromString($this->resultsFormat)->process($result));
+			$title = html_entity_decode(SSViewer::fromString($this->resultsFormat)->process($result));
+			$json[] = array(
+				"label" => $title,
+				"value" => $title,
+				"id" => $result->ID,
+			);
 		}
 		Config::inst()->update('SSViewer', 'source_file_comments', $originalSourceFileComments);
 		return Convert::array2json($json);
--- a/javascript/GridField.js
+++ b/javascript/GridField.js
@@ -371,9 +371,7 @@
 							url: $(searchField).data('searchUrl'),
 							data: encodeURIComponent(searchField.attr('name'))+'='+encodeURIComponent(searchField.val()), 
 							success: function(data) {
-								response( $.map(JSON.parse(data), function( name, id ) {
-									return { label: name, value: name, id: id };
-								}));
+								response(JSON.parse(data));
 							},
 							error: function(e) {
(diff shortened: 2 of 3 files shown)


patricknelson:
Side note: I think this affects all of `3` as well as `4.x` since it appears to be returning the same object format, which means it'll never ever actually be sorted like you think it is coming from the server-side. The field will always end up getting re-ordered by the ID.

willmorgan:
I read #4502 and understand what you're trying to do here - you've essentially fixed the strange data structure and format. :+1: 

As good a change this is, I think this belongs in the `3` branch, not `3.1`.

dhensby:
Hmm, this looks like it could be a serious change to method signature that could cause some headaches.

Specifically your tests now assert that the arrays are nested. This is a sensible change, though.

@tractorcow what do you think ;) ?

patricknelson:
@willmorgan @dhensby @tractorcow 

Should I rebase against `3` instead of `3.1`? Also, since my current production code is based on `3.1` (should maybe be `3.2` but just need time to test it) will I be able to ensure that this will make it into my `3.x` code eventually if/when the PR gets merged? Not 100% familiar with the workflow.

I just need this for a client-requested feature (manually ordering results via date, for example) so I can't go without it even with security updates, which might just mean more maintenance work on my end :)

patricknelson:
Also: @dhensby Yep, at the very least the outermost layer must be an array, not an object (which `json_encode()` will use if the keys aren't integers starting with zero), because once it's an object in JavaScript, that order of properties will not *necessarily* be retained. You could potentially reduce the redundancy in the label/value but that might be micro-optimization at the cost of legibility (so I didn't worry about it).

dhensby:
If this goes into 3 branch then it'll be released with 3.3 - so no solid timeframe for that.

If you *need* it for a project, your best best is to extend the `GridFieldAddExistingAutocompleter` and swap it in your grid config

patricknelson:
@dhensby

I've already done that. The main issue is the right binding it has (or the required mods) on the `GridField.js` file which would be lost with updates.

tractorcow:
I think given the return type is now substantially different in format, it probably can't go into 3.1. I would probably request master, which should be ok, given that you've overridden it in your own project ok?

forms/gridfield/GridFieldAddExistingAutocompleter.php
```diff
@@ -232,7 +232,12 @@ public function doSearch($gridField, $request) {
 		$originalSourceFileComments = Config::inst()->get('SSViewer', 'source_file_comments');
 		Config::inst()->update('SSViewer', 'source_file_comments', false);
 		foreach($results as $result) {
-			$json[$result->ID] = html_entity_decode(SSViewer::fromString($this->resultsFormat)->process($result));
+			$title = html_entity_decode(SSViewer::fromString($this->resultsFormat)->process($result));
+			$json[] = array(
+				"label" => $title,
```
willmorgan: Sorry to nitpick, but these should really be single quoted to stay consistent with the rest of the file (and SS in general)

dhensby:
Merged into master with 0768434997ea2d44546f61ff9f01d54f56383ca4 - taking into account @willmorgan's comments

willmorgan:
Cheers Dan!

PR #4504, opened 2015-08-16
When loading data into a form, make sure its using ALL fields

dhensby:
Fixes issue where loading data into a form doesn't take into account fields that could have been added by `getExtraFields`.

### DIFF
--- a/tests/forms/FormTest.php
+++ b/tests/forms/FormTest.php
@@ -544,6 +544,26 @@ class FormTest extends FunctionalTest {
 			$messageEls[0]->asXML()
 		);
 	}
+
+    public function testGetExtraFields()
+    {
+        $form = new FormTest_ExtraFieldsForm(
+            new FormTest_Controller(),
+            'Form',
+            new FieldList(new TextField('key1')),
+            new FieldList()
+        );
+
+        $data = array(
+            'key1' => 'test',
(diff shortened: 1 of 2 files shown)


tests/forms/FormTest.php
```diff
@@ -673,41 +693,57 @@ public function doSubmit($data, $form, $request) {
...
-				new FormAction('doSubmit')
-			)
-		);
-		$form->setFormMethod('POST');
-		$form->setStrictFormMethodCheck(true);
-		$form->disableSecurityToken(); // Disable CSRF protection for easier form submission handling
+class FormTest_ExtraFieldsForm extends Form implements TestOnly {
+
+    public function getExtraFields() {
+        $fields = $this->getExtraFields();
```
chillu: @dhensby This causes recursion

dhensby: lol, yes, should be `parent::getExtraFields`

chillu:
Technically, the `TEST` commit should be merged into the `FIX` one, they're dealing with the same thing. Also, there's no `TEST` "commit tag" documented in SilverStripe, and I don't see any value in declaring it like this. But these are both minor niggles, so merging as-is :)

dhensby:
@chillu - I like to do a test commit that fails first to prove that the fix works, I then usually fixup the commits into one

PR #4507, opened 2015-08-17
Save resampled images into a folder structure indicating transformations

JorisDebonnet:
Redone from #4328.

If I should rename static functions in `Filesystem.php` to be lowercase, please let me know which version number to use for the Deprecation notice.


dhensby:
@tractorcow are you able to weigh in on this? I think I'm happy with it in principle

filesystem/File.php
```diff
@@ -238,7 +238,7 @@ public function exists() {
 	 */
 	public static function find($filename) {
 		// Get the base file if $filename points to a resampled file
-		$filename = preg_replace('/_resampled\/[^-]+-/', '', $filename);
+		$filename = preg_replace('/_resampled\/(.+\/|[^-]+-)/', '', $filename);
```
tractorcow: This isn't going to be backwards compatible with the old filename pattern. Can you make the additional capture optional?

Also be careful of the `.` regexp pattern. It'll match slashes as well (which you probably want to avoid). Try `[^\\/]` instead?

You've also got the same logic as in HtmlEditorField. You should put this into a new method in file to get original filename from resampled filename. Then you can add a unit test just for that method. :) (hint hint).

model/Image.php
```diff
@@ -697,12 +697,25 @@ public function getFormattedImage($format) {
...
 		$format = $format . Convert::base64url_encode($args);
-		$filename = $format . "-" . $this->Name;
-		$patterns = $this->getFilenamePatterns($this->Name);
-		if (!preg_match($patterns['FullPattern'], $filename)) {
+		$filename = $format . "/" . $this->Name;
+		
+		$pattern = $this->getFilenamePatterns($this->Name);
+		
+		// Any previous formats need to be derived from this Image's directory, and prepended to the new filename
+		$prepend = array();
```
tractorcow: So each manipulation on a file is a new subdirectory?

E.g. `$Image.SetWidth(100).SetHeight(200)`?

filesystem/File.php
JorisDebonnet: It is backwards compatible though: old filenames (both before and after the base64 change) will match the second part of that "or" relation. New filenames will match the first part - which may in fact have slashes in them because the image may have been transformed multiple times (e.g. a rotate and then a translate), resulting in `_resampled/[rotate]/[translate]/orig_filename.jpg`. Since we want everything before `orig_filename.jpg` to be gone, it does need to match that 'middle' slash as well :)

As for putting it in a new method, sure, I could do that. A static method in the Image class perhaps?

model/Image.php
JorisDebonnet: That is correct.

tractorcow: Is it ok to ask for a unit test for this? To show that, after each additional manipulation, there is a new directory with the right name?

filesystem/File.php
tractorcow: Doof... I'm wrong again. Ok, on second look that will probably match the old pattern too. :) This is why I hate reviewing regexps! :D

Yeah, a static method with a test would help.

tractorcow:
Your code is good. I think just a few unit tests so that we are confident on merging would help get this over the line. :)

For users upgrading to 3.3 you might want to add a note under "upgrading" as in 3.2 (which has a huge list of upgrading notes). https://docs.silverstripe.org/en/3.2/changelogs/3.2.0

Just something small like "by the way, your resampled files were `<here>` but are now `<there>`"

model/Image.php
JorisDebonnet: Well, I had to change `ImageTest::testMultipleGenerateManipulationCalls_Regeneration` a bit already: this one already does several tests concerning multiple transformations - perhaps it already suffices? It does not explicitly check for particular subdirectories, but it does check that everything sits in the right place (even after regeneration).

tractorcow: That test is sufficient.

JorisDebonnet:
I actually got a bit confused while trying to check the 3.2 upgrade notes -- when I click on that link, I get redirected to `../#upgrading_notes` !! Must be an issue with the base path on that docs domain or something. Thought I'd signal that :P 

As for writing my own upgrade notes about this: there is no `04_Changelogs/3.3.0.md` yet. Can I create it in this pull request?

filesystem/File.php
JorisDebonnet: Before I push my changes, thought I'd ask you opinion on the name of that static method. I started out called it `Image::get_original_filename`, but that ends up not making sense because it makes it look like you can only provide an Image filename as a parameter, while its usages accept any kind of of Filename (pdf, xls,..), and they just want to make sure it doesn't have some _resampled stuff at the from. Feels wrong to put an xls filename into an `Image::get_original_filename` method, so I ended up calling it `Image::strip_resampled_prefix` instead. Make sense?..

tractorcow:
Waiting for travis to finish.

tests/model/ImageTest.php
```diff
@@ -411,6 +411,38 @@ public function testPathPropertiesCachedImage() {
...
+		$paths[] = 'assets/ImageTest/_resampled/ScaleHeightWzIwMF0=-test.image.with.dots.png';
+		//$paths[] = 'assets/ImageTest/_resampled/ScaleHeightWzIwMF0=-ScaleWidthWzQwMF0=-test.image.with.dots.png';
+		// Does not work, because the regex did not support this in 3.2!
+
+		// 3.1 (and earlier) format 
+		$paths[] = 'assets/ImageTest/_resampled/ScaleHeight200-';
+		//$paths[] = 'assets/ImageTest/_resampled/ScaleHeight200-ScaleWidth200-test.image.with.dots.png';
+		// Does not work, because the regex did not support this in 3.1!
+
+		foreach($paths as $path) {
```
tractorcow: Generally looped assertions are hard to review, since some test loggers just say "failure on line 441". In future it's preferable to unroll these. :)

tractorcow: Don't change anything though... I'll merge once the tests pass.

JorisDebonnet:
Took me a while to get the new ImageTest working... because apparently the *old* regex (in 3.2 and lower) did not support multiple transformations! Probably because that was never required: I think it was only ever used for images that were resized in the CMS (which takes only 1 transformatin). So I left the testcase for multiple transformations out for those old formats, and left it in for the new format (because there, it does work). 

Anyway, looks like there is still an issue... so still working on it. (for some reason I don't manage to get these errors on my own installation with dev/test/ImageTest)

tractorcow:
Ah, well you're close. ;) Let me know when tests are ok.

PR #4508, opened 2015-08-18
FIX: Using undefined var in ModelAdmin

kinglozzer:
See #4477. `$className` is undefined. Figured we may as well use `sanitiseClassName()` too as that’s what the method was added for.

ping @dhensby as you’ve been merging upstream lately. Looks like this is an issue across 3.1, 3.2, 3 and master. Good luck! :stuck_out_tongue_winking_eye:


dhensby:
Classic!

sminnee:
This was my fault, sorry. I merged a PR with a bug; the relevant code in ModelAdmin wasn't covered by a test.

dhensby:
No worries @sminnee. We need to get better at checking scrutinizr (and have it fail on these kinds of things). I was more commenting that I merged a big up to all branches, than the fact we pulled in a bug.

PR #4509, opened 2015-08-18
Make gridfield button sizes consistent with the rest of the cms

scott1702:
Before 
image
After 
image


sminnee:
This seems like a 3.2 change to me.

Issue #4515, opened 2015-08-20
RSSFeedTest::testRSSFeed fails in CLI with an error


AdenFraser:
When running the framework tests in CLI with PHPUNIT the follwoing error occurs:

There was 1 error:

1) RSSFeedTest::testRSSFeed
Undefined index: HTTP_HOST

/framework/tests/api/RSSFeedTest.php:78
/vendor/phpunit/phpunit/composer/bin/phpunit:63

Because there is no request header in the CLI, no HTTP_HOST is defined.

kinglozzer:
Have you defined an entry for `$_FILE_TO_URL_MAPPING`? https://docs.silverstripe.org/en/3.1/developer_guides/cli/#configuration. I think there are other tests that need it too

AdenFraser:
Have a slow day I think... Realized I had $_FILE_TO_URL_MAPPING for my Vagrant machine, but not my host machine... And I was running the tests on my host by mistake.

PR #4516, opened 2015-08-20
Fix Undefined $_SERVER index of `HTTP_HOST`

AdenFraser:
Fix for Issue #4515


AdenFraser:
Actually it's possible that this will break subsequent tests, since on tearDown() an originally unset $_SERVER['HTTP_HOST'] will now be set as an empty string from the $original_host var.

Thoughts?

AdenFraser:
Have a slow day I think... Realized I had $_FILE_TO_URL_MAPPING for my Vagrant machine, but not my host machine... And I was running the tests on my host by mistake.

PR #4517, opened 2015-08-20
Fix issue where Access All CMS Sections doesnt work

dhensby:
At the moment, if you select "Access to all CMS sections" in the Permissions for a group, a user belonging to the group can't access the Members.

Member canView requires the CMS_ACCESS_Security (which is a sub section of CMS_ACCESS_Security).

This patch automatically adds CMS_ACCESS_LeftAndMain to the permissions codes we check against if any CMS_ACCESS_* code is supplied.

### DIFF
--- a/security/Permission.php
+++ b/security/Permission.php
@@ -162,16 +162,23 @@ class Permission extends DataObject implements TemplateGlobalProvider {
 		} else {
 			$memberID = (is_object($member)) ? $member->ID : $member; 
 		}
-		
+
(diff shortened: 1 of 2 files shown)


tractorcow:
I think this fix is probably ok, but it'd be good to have a unit test :)

Maybe canEdit / canView on Member and Group?

dhensby:
Added a test on permission itself as this is intended to be about inherited permissions and not about explicitly being able to view or edit members

The fact this permission should allow view/edit/etc of members isnt knowledge we should be tightly coupling to IMO.

PR #3396, opened 2014-08-18
Removing @deprecated 3.1 functionality and classes.

halkyon:
Changelog has been updated to include what was removed in terms of
major functionality, and what to use as a replacement.

PR #4521, opened 2015-08-21
Revert #3425 #3396 to restore deprecated functionality

### LINKED ISSUE #4514: Don't remove stuff from 3.x
We need to revert 404478b07feb0f8efc8b19b1371d6f3b0398ff02 from the 3.2 and 3 branches, but not master.

These things shouldn't have been removed between 3.1 and 3.2.

The original PR #3396 was merged to master before be agreed semver and branched 3. :-(

I believe this blocks 3.2.0-rc1

tractorcow:
Fixes #4514

This should ensure 3.2 is properly semver.

There's one exception in Member where LastVisited and NumVisit has been removed, but documentation is included on how to restore it, and having it enabled by default has performance concerns.

### DIFF
--- a/control/Controller.php
+++ b/control/Controller.php
@@ -71,6 +71,20 @@ class Controller extends RequestHandler implements TemplateGlobalProvider {
 	public function init() {
 		if($this->basicAuthEnabled) BasicAuth::protect_site_if_necessary();
 
+		// Directly access the session variable just in case the Group or Member tables don't yet exist
+		if(Member::config()->log_last_visited) {
+			Deprecation::notice(
+				'4.0',
+				'Member::$LastVisited is deprecated. From 4.0 onwards you should implement this as a custom extension'
+			);
+			if(Session::get('loggedInAs') && Security::database_is_ready() && ($member = Member::currentUser())) {
+				DB::prepared_query(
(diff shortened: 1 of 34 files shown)


tractorcow:
@silverstripe/core-team can you please review to make sure this PR is appropriate?

dhensby:
It's my view that `LastVisted` and `NumVisits` behaviour needs to be restored.

It's conceivable that this behaviour is required by devs doing a 3.1 -> 3.2 upgrade and I don't see how it could be deemed as low risk or compliant with SemVer to remove this behaviour.

I don't expect that having it in the docs to be particularly safe way to avoid regressions. If the view is "putting it in the docs is good enough" then I'd put in the docs how to disable it.

tractorcow:
Ok, let's do this. Put the behaviour in, deprecate it, have it disabled by default with a single config change to turn on.

tractorcow:
See original discussion at https://github.com/silverstripe/silverstripe-framework/pull/2509

tractorcow:
Ok, LastVisited and NumVisit are back.

sminnee:
@tractorcow could you handle master now?

 * Do any planned merge forwards
 * Revert or exclude d0f94b5 from master, as appropriate

tractorcow:
Sure thanks :D

tractorcow:
I've merged up, excluding this commit from the merge to master.

PR #4523, opened 2015-08-21
BUG Fix bulk actions making sitetree unclickable

tractorcow:
If you select a bulk action with disabled items in it, disabling the action dropdown should make the whole sitetree clickable again. Before it would leave disabled items disabled.

3.3 version of https://github.com/silverstripe/silverstripe-framework/pull/4522


tractorcow:
I pushed this to origin by accident sorry. Re-opening at https://github.com/silverstripe/silverstripe-framework/pull/4524

PR #4524, opened 2015-08-21
BUG Fix bulk actions making sitetree unclickable

tractorcow:
If you select a bulk action with disabled items in it, disabling the action dropdown should make the whole sitetree clickable again. Before it would leave disabled items disabled.

3.3 version of #4522


dhensby:
Seems to be linking to the wrong travis build - it definitely passed.

tractorcow:
Yeah I accidentally pushed this commit (same sha) to the origin repo instead of my work. It's probably linking to that.

PR #4525, opened 2015-08-21
MINOR Adjustment to teamcity output

tractorcow:
This ensures that on error, the output will contain the full diff of any expected vs received values.


tractorcow:
Before:

image

After:

image

dhensby:
Very nice :)

chillu:
:star: :star: :star:

PR #4526, opened 2015-08-21
Add support for width and height to embedded photos

### LINKED ISSUE #4527: Images "from the web" do not respect height and width settings
Images "from the web" do not respect height and width settings.  Pull request resolving the issue:
https://github.com/silverstripe/silverstripe-framework/pull/4526

danrye:
Noticed that "from the web" images don't respect the width and height settings, this fixes that.

Fixes #4527

### DIFF
--- a/oembed/Oembed.php
+++ b/oembed/Oembed.php
@@ -238,7 +238,14 @@ class Oembed_Result extends ViewableData {
 		if(isset($options['class'])) {
 			$this->extraClass = $options['class'];
 		}
-
+		if($options) {
+			if(isset($options['width'])) {
+				$this->Width = $options['width'];
+			}
+			if(isset($options['height'])) {
+				$this->Height = $options['height'];
+			}
+		}
 		parent::__construct();
 	}
 


dhensby:
How are these properties then used?

danrye:
They were already being used: https://github.com/danrye/silverstripe-framework/blob/patch-1/oembed/Oembed.php#L341

dhensby:
derp

dhensby:
THANKS! :)

danrye:
My understanding was that PRs for 3 where for 3.* and would make the next release in each 3.* branch.  I've since been informed that this branch is for 3.3  How would I get this into 3.1.*? Should I submit a new PR against 3.1?

dhensby:
branches will be for the next version BELOW the name of the branch... eg:

branch 3 is for next 3.x release

branch 3.1 is for next 3.1.x release

If you want this in 3.1.x you'll need to open a PR against 3.1 branch. but note that it may get some resistance to 3.1.x because we're slowly locking that down to very critical bug fixes rather than ones that have moderate risk of regression.

I'm not saying this is a risky change, but the core team will have to take a view of it.

PS: if it gets into 3.1.x then it will be merged up to 3.2 3 and master.

PR #4529, opened 2015-08-24
update the filtered tree styling

scott1702:
To be merged with https://github.com/silverstripe/silverstripe-cms/pull/1264
Before
image
After
image
image


tractorcow:
Is there a before screen that helps show the difference?

dhensby:
@tractorcow looks like the original comment has been updated

tractorcow:
Did you test updates to the listview template as well? `CMSMain_ListView.ss`?

scott1702:
Yep, still needs work in list view, changing it now

PR #4530, opened 2015-08-24
BUG Ensure tests have relative urls not site-root urls

tractorcow:
Symptoms come up in testing environments where the baseurl has a subdirectory.

image

The fault is on the code assuming that `/some-url` is the root relative url when it's not always the case.

### DIFF
--- a/tests/control/ControllerTest.php
+++ b/tests/control/ControllerTest.php
@@ -328,7 +328,7 @@ class ControllerTest extends FunctionalTest {
 	*/
 
 	public function testRedirectBackByReferer() {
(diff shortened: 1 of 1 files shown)


kinglozzer:
@tractorcow easier to cherry-pick #4459?

tractorcow:
Oh, you're right it's the same issue! I'll cherry pick.

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-framework/commit/99a8a81e9a92ab3f2c7a8e20e6d8d9346e34d9f3

PR #4535, opened 2015-08-24
Disable IRC notifications for Travis builds

kinglozzer:
Any objections to this? These notifications are only really “useful” to core committers (almost all of which never use IRC anyway), and even then their use is limited. They’re basically white noise and anyone who enables Travis on their own fork, which is pretty handy, also adds to that.

Figured 3.1 still gets a few builds, so it’d make more sense to change this here and then merge up or cherry-pick.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,11 +68,6 @@ branches:
     - 2.3
     - translation-staging
 
-notifications:
-  irc:
-    channels:
-     - "irc.freenode.org#silverstripe"
-
 #  global:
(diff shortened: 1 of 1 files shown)


dhensby:
I'd have no problem with removing this.. What's the sentiment on IRC?

I believe @tractorcow looked into it previously and the mood was "keep them"

stevie-mayhew:
I don't think its an issue - no one on IRC has made any huge fuss about them. Although they probably wouldn't make a huge fuss if they weren't there.

Mostly IRC uses them to make comments about who is breaking the build rather than anything useful though.

sminnee:
The inclusion of the IRC config in personal forks is for me the main motivation for merging this PR.

I'll post a link to this PR to IRC and suggest that people ask us not to merge it if they like the notifications.

sminnee:
One of the benefits of IRC notification is that it tracked new tags. Instead of this, ss-helper now follows https://packagist.org/feeds/package.silverstripe/installer.rss

sminnee:
Everyone seems fine with this, I'm going to merge.

PR #4537, opened 2015-08-25
Remove forward slash in URLs in multibyte mode

chillu:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1262.

This change only affects projects which have `URLSegmentFilter.default_allow_multibyte` set to `true` (defaults to `false`). In that case, the removal of non-ASCII chars is disabled, but certain characters with semantic meaning in the URI context still need to be removed. I don't think any projects would've relied on this faulty behaviour, since it causes the page not to be found. Hence 3 as the target branch.

### DIFF
--- a/model/URLSegmentFilter.php
+++ b/model/URLSegmentFilter.php
@@ -32,6 +32,7 @@ class URLSegmentFilter extends Object {
 		'/\s|\+/u' => '-', // remove whitespace/plus
 		'/[_.]+/u' => '-', // underscores and dots to dashes
 		'/[^A-Za-z0-9\-]+/u' => '', // remove non-ASCII chars, only allow alphanumeric and dashes
+		'/\/+/u' => '-', // remove forward slashes in case multibyte is allowed (and ASCII chars aren't removed)
 		'/[\-]{2,}/u' => '-', // remove duplicate dashes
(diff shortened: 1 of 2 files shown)


model/URLSegmentFilter.php
```diff
@@ -32,6 +32,7 @@ class URLSegmentFilter extends Object {
 		'/\s|\+/u' => '-', // remove whitespace/plus
 		'/[_.]+/u' => '-', // underscores and dots to dashes
 		'/[^A-Za-z0-9\-]+/u' => '', // remove non-ASCII chars, only allow alphanumeric and dashes
+		'/\/+/u' => '', // remove forward slashes in case multibyte is allowed an the "remove non-ASCII chars" rule doesn't apply
```
kinglozzer: dat line length - can I be a pain and ask you to shorten it? Perhaps `// remove slashes if multibyte is allowed and the "remove non-ASCII chars" rule doesn't apply`? That gets it under 120 chars :)

dhensby: shouldn't it be replaced with a dash (`-`), too?

chillu: Both doneburgers.

kinglozzer:
Thanks, merged:hamburger:s!

PR #4539, opened 2015-08-25
Allow abstract BuildTask subclasses (closes #4538)

### LINKED ISSUE #4538: Unable to declare an abstract BuildTask
```php
abstract class AbstractUpdateTask extends BuildTask {

}

class MyUpdateTask extends AbstractUpdateTask {

}
```

results in `Unable to instantiate abstract class` errors when opening /dev/tasks or trying to run any other task.

kinglozzer:


dhensby:
Hmm, do those test classes need to implement `TestOnly`?

Couldn't `taskEnabled` be static?

I think, tbh, this should almost be a function of the class manifest (to tell us if classes are instantiate-able) and then we could also use this to create abstract DO classes

kinglozzer:
> Hmm, do those test classes need to implement TestOnly?

That was my initial thought, but we have a [separate class manifest](https://github.com/silverstripe/silverstripe-framework/blob/3/tests/bootstrap.php#L62) for unit tests so the ones I added don’t appear unless you’re running tests. Given that, I’m not entirely sure why we have `TestOnly`...

> Couldn't taskEnabled be static?

You’re right, it could’ve (probably should’ve) been static.

> I think, tbh, this should almost be a function of the class manifest (to tell us if classes are instantiate-able) and then we could also use this to create abstract DO classes

That’s a good idea, `ClassInfo::is_instantiable($class)` or something? Not sure whether storing that info in the class manifest would be worth the extra effort, it’s not something we’re likely to check hundreds of times per request. If it’s simple to add to the manifest though, then why not?

dhensby:
yes, adding it to `ClassInfo` as a minimum seems the most sensible (As it's a generic function that is about classes and can be helpful in other places)

`TestOnly` I think is more for `DataObject`s and you have to then explicitly state you need them in `extraDataObjects` on the test. so I guess you don't need it here.

PR #4543, opened 2015-08-25
FIX Members can access their own profiles in CMS

dhensby:
This fixes an issue where CMS users with limited permissions cannot access their own profile in the CMS.

### DIFF
--- a/admin/code/CMSProfileController.php
+++ b/admin/code/CMSProfileController.php
@@ -50,20 +50,20 @@ class CMSProfileController extends LeftAndMain {
 	}
 
 	public function canView($member = null) {
-		if(!$member && $member !== FALSE) $member = Member::currentUser();
-		
+		if(!$member && $member !== false) $member = Member::currentUser();
+
 		// cms menus only for logged-in members
 		if(!$member) return false;
-		
-		// Only check for generic CMS permissions
+
+		// Check they can access the CMS and that they are trying to edit themselves
 		if(
-			!Permission::checkMember($member, "CMS_ACCESS_LeftAndMain")
-			&& !Permission::checkMember($member, "CMS_ACCESS_CMSMain")
+			Permission::checkMember($member, "CMS_ACCESS")
+			&& $member->ID === Member::currentUserID()
 		) {
-			return false;
+			return true;
 		}
-		
-		return true;
+
+		return false;
 	}
 
 	public function save($data, $form) {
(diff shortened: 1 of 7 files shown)


admin/code/CMSProfileController.php
```diff
@@ -57,7 +57,7 @@ public function canView($member = null) {
 		
 		// Only check for generic CMS permissions
 		if(
-			!Permission::checkMember($member, "CMS_ACCESS_LeftAndMain")
```
dhensby: Note, this has been removed as a recent 3.1 fix has made `CMS_ACCESS_*` implicitly `CMS_ACCESS_LeftAndMain`

tractorcow:
Need to ensure that for users without CMS access that canView() is false. :)

tractorcow:
I suggested the following code in Permission to provide a generic wildcard match for CMS permissions.

```php
if($permCode === ‘CMS_Access_*’) {
	foreach(self::$cache_permissions[$memberID] as $hasCode) {
		if(stripos($hasCode, ‘CMS_Access_’) === 0) {
			return true;
		}
	}
}
```

In 4.0 it would be great to support generic wildcards.

kinglozzer:
The fix itself looks good to me, will await feedback from others: should this be 3.1 or 3? It’s arguably both a bug fix and a new feature. BC risk looks pretty tiny here, and CMS users without full `LeftAndMain` or `CMSMain` permissions are unable to edit basic stuff like their email address out of the box, so I’m 75/25 in favour of 3.1 :stuck_out_tongue_closed_eyes:.

dhensby:
@tractorcow I've kind of followed your advice here.

I'm not using `CMS_ACCESS_*` as the param, I'm using `CMS_ACCESS` as I feel that lends itself to the idea that the `CMS_ACCESS_*` are children and that this isn't explicitly accepting some kind of regex or wildcard syntax (it's not).

I'm not using `strpos` because we need to know if the string starts with our required string and strpos will search the whole string if a match isn't found at the start.

tractorcow:
stripos() === 0 is used to check if the string starts with a value. :) hence the `===0`

dhensby:
<3

chillu:
Hmm, more magic in permissions API :( Can you please document that on https://docs.silverstripe.org/en/3.1/developer_guides/security/permissions? Also highlight the fact that this  grouping approach only works with that specific string (`CMS_ACCESS`), not with any other groupings devs might come up with.

dhensby:
ee #4546

tractorcow:
In 4.x I want to implement a proper api for inheritable (and inheriting) permissions.

PR #4544, opened 2015-08-26
FIX: Recursion errors when sorting customised ViewableData objects (fixes #4464)

### LINKED ISSUE #4464: "Nesting level too deep" while sorting
I recently came across this bug on one of my modules: the script crashed with the message

    Fatal error: Nesting level too deep - recursive dependency? in .../framework/model/ArrayList.php on line 449

After a lot of testing I have been able to replicate the bug with the following snippet:

```
    public function testSortWithDuplicates() {
        $items = ArrayList::create();

        $items->add(DataObject::create(array('Order' => 1)));
        $items->add(DataObject::create(array('Order' => 1)));

        // Without this cycle sort does not fail
        foreach ($items as $item) {
            $item->customise(array());
        }

        // This call fails
        $items->sort('Order');
    }
```

I did not see anything inherently wrong above.

Tested on 3.1.13 and 3.2.0-beta1 on a server with PHP 5.6.11. Reported from a server with PHP 5.4.20.

kinglozzer:
Basically, if the sort values are the same then PHP will try to inspect the object’s properties, which will completely break if there are recursive dependencies: https://3v4l.org/s9UZM. This issue was highlighted by https://github.com/bummzack/sortablefile/pull/41. I don’t think we need to worry about people passing in their own object types with recursive dependencies (we can’t fix PHP after all), but I think we should fix it for `ViewableData`.

`spl_object_hash()` seems to be good enough. If the objects being compared are identical (i.e. the same instance of the object) then the hashes are the same, but `array_multisort()` is already able to handle that anyway.

This is an issue in 3.1 too, but we don’t have an easy way of telling whether an object is customised (`ViewableData::getCustomisedObj()` doesn’t exist in 3.1). I’m happy to resubmit this to 3.1 if anyone can think of a workaround for that.

### DIFF
--- a/tests/model/ArrayListTest.php
+++ b/tests/model/ArrayListTest.php
@@ -426,6 +426,25 @@ class ArrayListTest extends SapphireTest {
 	}
 
 	/**
+	 * Check that we don't cause recursion errors with array_multisort() and customised ViewableData objects
+	 */
+	public function testSortWithRecursiveReferences() {
+		$items = ArrayList::create();
+
+		// Objects need the same sort value so that their properties are compared by array_multisort
+		$items->add(DataObject::create(array('Order' => 1)));
+		$items->add(DataObject::create(array('Order' => 1)));
+
+		// Customising the object adds recursive dependencies (between the object and its customised instance)
+		foreach ($items as $item) {
+			$item->customise(array());
+		}
+
+		// This call will trigger a fatal error if there are recursion issues
+		$items->sort('Order');
+	}
+
+	/**
 	 * $list->filter('Name', 'bob'); // only bob in the list
 	 */
 	public function testSimpleFilter() {
--- a/model/ArrayList.php
+++ b/model/ArrayList.php
@@ -425,6 +425,7 @@ class ArrayList extends ViewableData implements SS_List, SS_Filterable, SS_Sorta
 		// This the main sorting algorithm that supports infinite sorting params
 		$multisortArgs = array();
 		$values = array();
+		$objectHashes = array();
 		foreach($columnsToSort as $column => $direction ) {
 			// The reason these are added to columns is of the references, otherwise when the foreach
 			// is done, all $values and $direction look the same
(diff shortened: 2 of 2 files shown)


tests/model/ArrayListTest.php
```diff
@@ -426,6 +426,25 @@ public function testMultiSort() {
...
+	public function testSortWithRecursiveReferences() {
+		$items = ArrayList::create();
+
+		// Objects need the same sort value so that their properties are compared by array_multisort
+		$items->add(DataObject::create(array('Order' => 1)));
+		$items->add(DataObject::create(array('Order' => 1)));
+
+		// Customising the object adds recursive dependencies (between the object and its customised instance)
+		foreach ($items as $item) {
+			$item->customise(array());
```
willmorgan: This is elegant, but prone to breaking should `customise`'s behaviour change.

Is it possible to manually create these recursive references?

willmorgan:
Nice PR! Only one comment from me :smile:

tests/model/ArrayListTest.php
kinglozzer: I’m not sure if it’s possible to set them up manually, but I didn’t want to anyway as we’re testing a specific scenario here - where customised `ViewableData` objects cause recursion errors. Setting them up manually is more leaning towards testing PHP’s behaviour with recursive/cyclic dependencies.

I don’t think this is likely to be broken unless we completely change how `ViewableData::customise()` works, in which case we’d probably _want_ this to break so that it highlights that we can (hopefully) remove the extra logic I added to `ArrayList::sort()`. If I set the dependencies up manually and `ViewableData::customise()` changes, then this test would continue passing (and be forgotten about) even if it’s completely redundant.

kinglozzer:
Just spoken to @dhensby about this on Gitter and he has pointed out why @willmorgan is correct, so I’ll amend this PR. New plan is just to get the hash of every object that passes through the sort method, which should fix it for _all_ circular dependencies instead of just `ViewableData` :)

tests/model/ArrayListTest.php
dhensby: We are tightly coupling to customise's behaviour here - we probably shouldn't do that.

Isn't this patch solving a problem with recursive references, not customise?

I see no reason not to solve this for all objects rather than just `ViewableData`

dhensby:
Is this a bug that should be fixed in 3.1?

kinglozzer:
> Is this a bug that should be fixed in 3.1?

Quite, and now that I no longer need that `ViewableData` getter it can be fixed there too :)

PR #4545, opened 2015-08-26
Correct link to previous changelog

JorisDebonnet:


tractorcow:
That changelog doesn't have the full list of upgrading instructions though, which is the purpose of this backlink. Otherwise it needs to be duplicated across each file. :)

In the final release we'll ensure the full changelog file has this information. Perhaps we should point to there instead?

tractorcow:
Shall we link to https://github.com/silverstripe/silverstripe-framework/blob/3.2/docs/en/04_Changelogs/3.2.0.md instead?

JorisDebonnet:
I was assuming you were expecting people to click click through; first see the newest changes in rc1, then the changes in beta2, then the changes in beta1, ...

But if there is a general changelog for 3.2.0, then I'd say that would indeed be the most sensible one to link to!

PR #4547, opened 2015-08-27
Update messages_sk.js

rasstislav:
added spaces


chillu:
Please submit this patch upstream at the third party library. Once its accepted, you can submit a pull request updating the core library in SilverStripe. This sounds like a lot of work, but we don't want to fork these third party libraries - otherwise we run the danger of removing your patch again once we update. We're running a five year old version of the library (https://github.com/jzaefferer/jquery-validation/commit/858ca34cf44ad951e65027cb5f3699cf6b580795), so you'll need to go through the changelogs since 1.8.0 and check if we need to change any API usage (https://github.com/jzaefferer/jquery-validation/releases). Closing this issue for now since its not actionable.

PR #4551, opened 2015-08-27
PHP7 support

sminnee:
Here's a first cut of the PHP7 support. The most substantive change is the renaming of DB classes. I haven't done all the classes yet—just the minimum to make PHP7 build—but the approach that I have taken is:

 * Use namespace `SilverStripe\Model\FieldType`. This namespace leaves open the room for branching the ORM (and other sub components) into their own, smaller packages, and keeps the namespace concise.
 * Prefix all classnames with `DB`. This is in keeping with `DBField` and had been suggested before. It also suggests a good name for the relevant class in the asset refactoring: `DBFile`. :-)
 * Use the Injector to map `Int` to `SilverStripe\Model\FieldType\DBInt`. I think that this is a good first step upon which we can decide to build if we find that a more sophisticated mapping system is needed down the line.

If people like this approach I'll apply it to other the other DBField subclasses.


kinglozzer:
Looks good so far :+1:

> Use namespace `SilverStripe\Model\FieldType`.

Naming looks fine to me, always seems to be subjective though!

> Prefix all classnames with `DB`.

Looking at the diff, only `Int` and `Float` have been namespaced & renamed so far? Are you intending to rename all of them as part of this PR before it’s merged, or do it later? IMO we may as well rename all of them in this PR and just get it done, rather than end up with inconsistent naming/namespacing.

> Use the Injector to map `Int` to `SilverStripe\Model\FieldType\DBInt`.

I think this is acceptable. Like you said, we can wait and see if we really need a mapping system later. As far as I’m aware this is the only scenario we’ve encountered that we thought had a requirement of one - people can already use (and are already using) Injector in this way for code outside of this “db” context, so it makes sense to use what we have if we can :)

sminnee:
> Looking at the diff, only Int and Float have been namespaced & renamed so far? Are you intending to rename all of them as part of this PR before it’s merged, or do it later? 

Yes, I was keen to break out the minimum necessary as a separate commit (in case someone needed to backport) but I don't think the PR should be merged until this is done. First, though, I wanted to get the PHP7 build actually functioning. ConfigTest seems to be segfaulting. :-(

sminnee:
OK, I've gotten PHP7 support working now. Now I am updating all the DBField classes to follow the same pattern, which is a bit more involved.

One thing that it highlights is that it will be handy to have a script that:

 * parses PHP files
 * looks for simple references to altered classnames: `<literal>::` or `new <literal>`.
 * replaces the class name with the new one (e.g. `StringField` => `DBString`) and adds the relevant `use` statement to the top of the file.

I'll raise this as a separate ticket.

sminnee:
OK—I've applied this work to all the DBField subclasses now, which is a bit more of an involved change.

I've also put together a namespacing RFC at #4562

sminnee:
It's probably best to hold off merging this until we've seen some feedback on the namespacing RFC, and taken a core-committer vote on it.

tractorcow:
Would it not make autoloading, should we decide to go the psr-4 route, more concise to use the `<Vendor>/<Package>` convention?

```json
{
    "autoload": {
        "psr-4": {
            "SilverStripe\\Framework\\": "code/"
        }
    }
}
```

I would expect other modules would have their own root directory. Strictly, it's not necessary to have unique namespaces for each module (each could just be SilverStripe\\) but it's less efficient.

I'm not worried about having to type longer names because we have tools to manage namespaced classes for us, and only means a slightly longer `use` statement. :)

core/Config.php
```diff
@@ -703,8 +703,8 @@ class Config_LRU {
 
 	public function __construct() {
 		Deprecation::notice('4.0', 'Please use Config_MemCache instead', Deprecation::SCOPE_CLASS);
-		if (version_compare(PHP_VERSION, '5.3.7', '<')) {
-			// SplFixedArray causes seg faults before PHP 5.3.7
+		if (version_compare(PHP_VERSION, '5.3.7', '<') || version_compare(PHP_VERSION, '6.99.99', '>')) {
+			// SplFixedArray causes seg faults before PHP 5.3.7, and also in 7.0.0-RC1
```
camspiers: Did you file a bug for this? Or is it already tracked?

sminnee:
@tractorcow 

> Would it not make autoloading, should we decide to go the psr-4 route, more concise to use the <Vendor>/<Package> convention?

In short: if/when we move the ORM to a silverstripe/model package, I wouldn't want to have to change the namespaces. Best place to discuss that is probably #4562.

core/Config.php
sminnee: No, I'll see what I can do. It's hard for me to reproduce as it only happens on travis.

tractorcow:
Oh I see! Ok, well if that's the case, then you've already thought 1 step ahead of me.

tractorcow:
I was wondering, should we not encourage users to use the correct name of the class, even if we are aliasing them?

If I had `private static $db = array('Name' => 'DBVarchar');` should that work as well? I would be happy with Varchar as an alias for DBVarchar (which itself is an alias for the fully namespaced classname).

sminnee:
OK it sounds like we've got the go-ahead with namespacing RFC (RFC-6) and so the namespaces used in this PR are okay.

What further changes block merge?

tractorcow:
A rebase it seems. :)

sminnee:
I might hold off until your composite field changes have been merged. Anything else?

core/Config.php
kinglozzer: Assuming it still applies to newer PHP 7 RCs, is it worth patching this? `Config_LRU` needs to be removed in 4.0 anyway, it could even be removed as part of this PR (scope creep, yay!)

kinglozzer:
Err, looks like I commented on the commits instead of the diff, sorry :sweat_smile:

sminnee:
Yeah, rebasing this will take some fixing, I'll try and find a few hours to get onto it soon!

core/Config.php
dhensby: We are on RC7 now... perhaps it's sorted?

dhensby:
@sminnee ping

core/Config.php
sminnee: Yeah this needs a rebase, which isn't trivial (asset refactor work
overlaps). Will try and find some time to do it.

nathanbrauer:
Following
(Sorry for the notification.)

sminnee:
OK, there's 1 test failing due to global state issues, but this is almost done!

tractorcow:
I can't make heads or tails how that error even came about...

sminnee:
Yeah, it doesn't happen if you run the test on its own. I'm guessing it's something related to junk left in the filesystem.

sminnee:
Sweet, that fixed it. Just rebasing now, but this is ready for final code review and, ideally, merging.

model/fieldtypes/CompositeDBField.php
```diff
@@ -7,7 +14,7 @@
  *
  * Example with a combined street name and number:
  * <code>
-* class Street extends CompositeDBField {
+* class Street extends DBComposite {
```
tractorcow: File needs to be renamed to match the class.

filesystem/storage/DBFile.php
```diff
@@ -1,9 +1,20 @@
 <?php
 
+namespace SilverStripe\Model\FieldType;
+
```
tractorcow: use SilverStripe\Model\FieldType\DBComposite; ?

filesystem/storage/DBFile.php
```diff
@@ -1,9 +1,20 @@
...
+use AssetField;
+use File;
+use Director;
+use Permission;
+use ShortcodeHandler;
+use ValidationResult;
+use ValidationException;
+
 // Un-comment once https://github.com/silverstripe/silverstripe-framework/pull/4551/ is merged
 // namespace SilverStripe\Filesystem\Storage;
```
tractorcow: By the way, uncomment this :D

tractorcow: This begs the question; We ARE planning on psr-4 standardising the directories to these namespaces right?

model/fieldtypes/Enum.php
```diff
@@ -7,7 +14,7 @@
  * @package framework
  * @subpackage model
  */
-class Enum extends StringField {
+class DBEnum extends DBString {
```
tractorcow: Shouldn't we be renaming the files to match the real classname?

tractorcow:
I had a look over, and the only real concern I had was with filenames / directories. It's probably ok if we don't worry TOO much about exact directories until / if we move to psr-4 (as we've already discussed this), but shouldn't we follow the file naming convention for the renamed classes?

tractorcow:
I suppose we want to make it easy to merge up from 3 -> master for the mean time... in which case... as long as it's temporary. :)

tractorcow:
+1 for merge.

sminnee:
Regarding file names, git should be able to follow them through merges if we do them as a separate commit. I'll do that now.

sminnee:
I'll leave the directory name until an actual PSR-4 change though, I think. I'll just change the filenames

tractorcow:
Yeah I think that is a good compromise, considering our filenames already match while the directories haven't so much. :) That can be a larger later task.

sminnee:
OK, DBFile is back as part of the filesystem namespace, and SilverStripe\Model\FieldType and its tests have more appropriate classnames.

From my perspective, this is ready to merge. I've squashed things into logical groupings, the remaining stuff that's in separate commits was done so deliberately, but if you think something should be squashed further let me know. Note that the file renaming in its own commit was to give git's file-rename detection a helping hand to for cleaner merge-throughs.

kinglozzer:
I’ve had another look over this and (apart from the nitpicking above) I think we’re good!

model/FieldType/SS_Datetime.php
```diff
@@ -0,0 +1,49 @@
...
+		self::deprecation_notice();
+		return parent::clear_mock_now();
+	}
+
+	public static function get_template_global_variables() {
+		self::deprecation_notice();
+		return parent::get_template_global_variables();
+	}
+
+	protected function deprecation_notice() {
```
tractorcow: Did you mean to make this method static?

tests/dev/FixtureBlueprintTest.php
```diff
@@ -287,7 +287,7 @@ class FixtureBlueprintTest_SiteTree extends DataObject implements TestOnly {
 class FixtureBlueprintTest_Page extends FixtureBlueprintTest_SiteTree {
 
 	private static $db = array(
-		'PublishDate' => 'SS_DateTime'
+		'PublishDate' => 'SS_Datetime'
```
tractorcow: Should we use the non-deprecated class alias `DateTime` here?

tests/model/DbDatetimeTest.php
```diff
@@ -1,145 +1,224 @@
...
+ *  the current date (not hour, minute, second) and assume that the date
+ *  doesn't switch throughout the test execution. This means tests might
+ *  fail when run at 23:59:59.
+ *
+ * @package framework
+ * @subpackage tests
+ */
+class DBDatetimeTest extends SapphireTest {
+	public function testNowWithSystemDate() {
+		$systemDatetime = DBField::create_field('SS_Datetime', date('Y-m-d H:i:s'));
```
tractorcow: I suggest to swap out all `SS_Datetime` strings in this class with the non-deprecated `DateTime`

tractorcow:
I left a few comments; I think that it might be fine to continue to support SS_Datetime strings being mapped to the new class, but do we want to encourage users to use `DateTime` instead when defining their DB fields?

I think we probably need a note on upgrading in https://github.com/silverstripe/silverstripe-framework/blob/master/docs/en/04_Changelogs/4.0.0.md.

Something like the below:


```markdown
### Upgrading references to DBField and subclasses

A major change in 4.0 is the introduction of namespaced DBField subclasses. Now as a standard, all DBField subclasses have a `DB` prefix, are namespaced, and have an associative alias which omits the DB prefix.

This means that for the most part, code that worked in 3.0 won't need to be changed, although if you have any hard class literals which reference the old classes, they will need to be updated to point to the new namespaced classes.

An exception to this is any classes which once had the `SS_` prefix, which will now be instead prefixed with `DB`, such as `SS_Datetime`,  which is now `DBDateTime`.

For example:

class MyObject extends DataObject {
	private static $db = array(
		'Number' => 'Int',
		'Time' => 'SS_Datetime'
	);
	
	/**
	 * @param Int $val
	 * @return SS_Datetime
	 */
	public function numberToDate(Int $val) {
		return new SS_Datetime(date('U', $val->Value());
	}
}

Will become:


use SilverStripe\Model\FieldType\DBInt;
use SilverStripe\Model\FieldType\DBDateTime;

class MyObject extends DataObject {
	private static $db = array(
		'Number' => 'Int',
		'Time' => 'DateTime'
	);
	
	/**
	 * @param DBInt $val
	 * @return DBDatetime
	 */
	public function numberToDate(DBInt $val) {
		return new DBDatetime(date('U', $val->Value());
	}
}

Note that string references to SS_Datetime passed to injector, or used in config values, will still work, and will refer to the updated class names.

```

sminnee:
Cheers, I'll incorporate those changes.

chillu:
Happy with the change, plus @tractorcow's comments. Damian, in your upgrade comment suggestion, should we encourage usage of injector rather than `new SS_Datetime`?

tractorcow:
Yes we should, but if I put that in the example it wouldn't show how to update class references anymore.

kinglozzer:
Ping @sminnee

JorisDebonnet:
I notice 4.0-alpha1 is set as milestone. Is this definite, or may php7 support still be added to 3.x?

tractorcow:
I'm pretty sure 3.x will never be 7.x compatible, since we'd have to remove classes to make it work, which would violate semver.

sminnee:
Hey everyone, this is rebased and should be ready to merge now. I added some docs as requested above.

camspiers:
Yay!

dhensby:
This is the future!

sminnee:
> This is the future!

...or at least the present. Sorry it took so long to rebase, guys, but I'm glad to have this in.

dhensby:
Erm, so who wants to do PHP 6 support?

assertchris:
wat

sullit:
assert(what's taking so long with the php7.x refactoring?)

;P

mycrazydream:
You guys have hands down the most professional and amazing PHP framework/CMS and people know Wordpress much more than they know you. I mean, have you looked at Wordpress code. It's absolute shite. A ton of if_function_exists() and playing jenga with the code. I build a moderately sized site for a Structural Engineering company on contract (I did it half as favor and half as I could use the money; charge him about half my normal rate) and the developer they brought in after me said when they asked him that annoying question about why he wanted to work at B.R.Assocates was, to learn how to create sites like this. 

And that wasn't me. That was you guys that had his jaw dropping no doubt. And you are getting php7 support into a release soon? simply amazing. It keeps me interested in code because I am stuck in a M$ shop ATM and while C# is a reasonable language, working in .NET is laborious as hell.

Keep up the good work, SilverStripe.

- A devoted fan

sminnee:
Thanks for the kind words, @mycrazydream!

WindSaber:
First of all, I want to say Hello and thank you for this amazing platform

I've read about the DBField changes, but I'm afraid there's a place where it's not working well

In **php 7** with the **CMS & Framework v4.0.0-alpha1** version, I've tried to configurate the Search control `FulltextSearchable::enable();`

But I'm getting the DBField Error

```
[Emergency] Uncaught Error: Class 'DBField' not found
GET /ss/sobre-nosotros/pruebanuevo-pagina/SearchForm?Search=nuevo&action_results=L
Line 45 in C:\xampp\htdocs\ss\cms\code\search\ContentControllerSearchExtension.php
```

Am I missing something? Or Is this part of the php7 working support?

WindSaber:
My fault! I forgot to check all the silvertripe projects...

I've add a pull request [`silverstripe-cms  pull#1498`](https://github.com/silverstripe/silverstripe-cms/pull/1498)

Again thanks for this amazing CMS

phptek:
Hi all - what's the status of this PR? It's been a few months since the last comment and the rebase work.

sminnee:
It's merged, @phptek. SilverStripe 4 works with PHP7.

phptek:
Duh. Completely missed the big blue "merged" lozenge above - bloody good work all round @sminnee thanks!

PR #4552, opened 2015-08-28
Fix alignment of logout area

scott1702:


dhensby:
pic please?

dhensby:
oh, and you need to resolve the conflict in your map file :P

scott1702:
Doh! lol
image

dhensby:
ok, *personally* I don't like that it has so much space, but if that's what we're going for, then let's do it.

PR #4556, opened 2015-08-28
MINOR: Drop wasteful 'allow-failures' builds; simplify build matrix

sminnee:
Our build time is an impediment: frequently we're slowed down waiting for
test runs to complete. In this context, running allowed-failure builds
that everyone ignores is just a waste.

I'd love to see HHVM work, but that's going to require that someone
actually gets the build to pass. At that point, we can add it back into
the build matrix.

I'm working on a PHP7 branch that will hopefully also fix our nightly
builds. Once I get that going, I'll add those two back, but not have
'allow failures'

Even if we decide, at some point in the future, that supporting nightly
doesn't make sense, I see the *only* value of Travis is maintaining our
green tick. We either care about builds or we don't: I've never seen a
'grey area' build result in anything other than embarrassment and
technical debt.

Let's decide what matters to us and treat it like it matters to us:
it's part of the build or not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,11 +23,6 @@ env:
     - DB=PGSQL
 
 matrix:
-  allow_failures:
-    - php: hhvm
-    - php: 7.0
-    - php: nightly
-
   include:
     - php: 5.4
       env: DB=MYSQL PDO=1
@@ -37,12 +32,6 @@ matrix:
       env: DB=MYSQL
     - php: 5.4
       env: DB=MYSQL BEHAT_TEST=1
-    - php: 7.0
-      env: DB=MYSQL
-    - php: nightly
-      env: DB=MYSQL
-    - php: hhvm
-      env: DB=MYSQL
 
 before_script:
  - composer self-update || true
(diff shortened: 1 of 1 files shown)


chillu:
Yeah, unless we get clear metrics about number of failed tests in allowed_failures, I'm inclined to agree. Number of failed tests over time could prevent us from breaking builds on HHVM even *more* than already, but that's not an easy metric to track from Travis. I don't think its particularly wasteful in terms of our overall build time since Travis runs all builds concurrently. But it's wasting Travis' server resources. It's a free resource for us, but we should care about the sustainability of Travis anyway :)

sminnee:
> Number of failed tests over time could prevent us from breaking builds on HHVM even more than already, but that's not an easy metric to track from Travis.

I've heard about this in theory. In practise I've only seen it work when there's a dedicated test manager (Phil B), and even then it was a soul-destroying uphill battle.

Note: I'm inclined to take the same approach with stuff like trailing whitespace: fix it all and then treat any violation as a build failure.

PR #4557, opened 2015-08-28
Defining more consistent API across ArrayList and DataList

dhensby:
see #3828

### DIFF
--- a/model/DataList.php
+++ b/model/DataList.php
@@ -884,7 +884,7 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 	 * @param array $ids Array of integers
 	 * @return DataList
 	 */
-	public function byIDs(array $ids) {
+	public function byIDs($ids) {
 		return $this->filter('ID', $ids);
 	}
 
--- a/model/Filterable.php
+++ b/model/Filterable.php
@@ -34,6 +34,30 @@ interface SS_Filterable extends SS_List {
 	public function filter();
 
 	/**
+	 * Return a copy of this list which contains items matching any of these charactaristics.
+	 *
+	 * @example // only bob in the list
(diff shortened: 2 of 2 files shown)


model/ArrayList.php
```diff
@@ -477,6 +477,72 @@ public function canFilterBy($by) {
...
+			}
+		}
+
+		$list = clone $this;
+		$list->items = $itemsToKeep;
+		return $list;
+
+	}
+
+	protected function normaliseFilterArgs() {
```
tractorcow: PHPDoc please. :D

model/ArrayList.php
```diff
@@ -477,6 +477,72 @@ public function canFilterBy($by) {
...
+	 * @return DataList
+	 */
+	public function filterAny() {
+		$keepUs = call_user_func_array(array($this, 'normaliseFilterArgs'), func_get_args());
+
+		$itemsToKeep = array();
+
+		foreach ($this->items as $column => $value) {
+			$keepItem = false;
+			if ((is_array($value) && in_array($this->extractValue($item, $column), $value))
```
camspiers: It seems like there is some code missing here. Don't you want to iterate over $keepUs and for any passing conditions set keepItem true then break?

Probably a good idea to add tests for this too

tractorcow: Yep, just look at the deleted code in `filter` for this logic.

tractorcow:
I've also added some comments to https://github.com/silverstripe/silverstripe-framework/pull/3828.

dhensby:
feedback from other PR incorporated here

dhensby:
bump @tractorcow

tractorcow:
Rebased and merged with https://github.com/silverstripe/silverstripe-framework/commit/53ffc1a0dbac52ad682a150d387b104c1dc0616f

PR #4558, opened 2015-08-29
MINOR: Simplify build matrix for 3.x builds.

sminnee:
This is a companion to https://github.com/silverstripe/silverstripe-framework/pull/4556
but targeted at the build needs of SilverStripe 3.x. Commit into 3.1 and
then the merge forward into 3.2 and 3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,6 @@ addons:
     packages:
       - tidy
 
-php:
-  - 5.4
-
 env:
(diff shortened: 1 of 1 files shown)


chillu:
Hmm, yeah I guess it's fairly easy to determine if a fault comes from PHP 5.4 or Postgres, so not much value in running the full matrix.

PR #4563, opened 2015-08-30
Splitting BulkLoader imports to reduce memory consumption

assertchris:
This allows BulkLoaders to split large files into smaller files, so that a limited amount of CSV data is read into memory during imports.

Internal issue tracker reference: https://silverstripe.atlassian.net/browse/CWP-574


dev/BulkLoader.php
```diff
@@ -143,7 +152,83 @@ public function load($filepath) {
...
+	 * @return array List of file paths
+	 */
+	protected function splitFile($path, $lines = null) {
+		ini_set('auto_detect_line_endings', true);
+
+		if (!is_int($lines)) {
+			$lines = $this->config()->get("lines");
+		}
+
+		$new = $this->getNewSplitFileName();
```
chillu: I'd feel better about this if the temporary files were cleaned up after processing. It's a fairly generic utility, and hence could be used to upload quite sensitive files. If you look at `getTempParentFolder()`, you'll see that those temporary files can end up in the webroot (`silverstripe-cache`). So while the uniqid makes them fairly hard to guess, it adds unnecessary security surface IMHO.

assertchris: https://github.com/silverstripe/silverstripe-framework/pull/4563/files#diff-33224465f3edd7ea4b14e3a3a37f6423R168

chillu:
The bulk loader is designed as a generic importer (JSON, XML, etc) so shouldn't assume you're only dealing with CSV files. Can you please move the splitting into the `CSVBulkLoader` subclass?

assertchris:
Good point - will do!

assertchris:
@chillu better?

dev/CsvBulkLoader.php
```diff
@@ -47,8 +56,99 @@ public function preview($filepath) {
...
+		foreach ($files as $file) {
+			$next = $this->processChunk($file);
+
+			if ($result instanceof BulkLoader_Result) {
+				$result->merge($next);
+			} else {
+				$result = $next;
+			}
+
+			@unlink($file);
```
chillu: Maybe wrap this whole loop in a try/catch as well with the unlink in the catch? `tinfoilhat`

dev/CsvBulkLoader.php
```diff
@@ -47,8 +56,99 @@ public function preview($filepath) {
...
+	/**
+	 * Splits a large file up into many smaller files.
+	 *
+	 * @param string $path Path to large file to split
+	 * @param int $lines Number of lines per file
+	 *
+	 * @return array List of file paths
+	 */
+	protected function splitFile($path, $lines = null) {
+		ini_set('auto_detect_line_endings', true);
```
chillu: You should reset to the original value after you're done, right?

dev/CsvBulkLoader.php
```diff
@@ -47,8 +56,99 @@ public function preview($filepath) {
...
+			$lines = $this->config()->get("lines");
+		}
+
+		$new = $this->getNewSplitFileName();
+
+		$to = fopen($new, 'w+');
+		$from = fopen($path, 'r');
+
+		// pop the headers and write them to the new file
+		$header = fgets($from, 4096);
```
chillu: CSVBulkloader doesn't mandate header rows (see `CSVBulkLoader-> hasHeaderRow`), so this could lead to duplicate data imports.

dev/CsvBulkLoader.php
```diff
@@ -47,8 +56,99 @@ public function preview($filepath) {
...
+		fwrite($to, $header);
+
+		$files = array();
+		$files[] = $new;
+
+		$count = 0;
+
+		while (!feof($from)) {
+			fwrite($to, fgets($from, 4096));
```
chillu: How about lines longer than 4k? CSV column data can get pretty large...

tests/dev/CsvBulkLoaderTest.php
```diff
@@ -229,6 +229,17 @@ protected function getLineCount(&$file) {
...
 
+	public function testLargeFileSplitIntoSmallerFiles() {
+		Config::inst()->update('CsvBulkLoader', 'lines', 10);
+
+		$loader = new CsvBulkLoader('CsvBulkLoaderTest_Player');
+		$path = $this->getCurrentAbsolutePath() . '/CsvBulkLoaderTest_LargeListOfPlayers.csv';
+
+		$results = $loader->load($path);
+
+		$this->assertEquals(100, $results->Count());
```
chillu: This will cause a hundred row writes for the test, which is a bit excessive in terms of keeping our tests fast. How about using 10 lines total, and chunking each 3 lines?

dev/CsvBulkLoader.php
assertchris: What would you suggest?

tests/dev/CsvBulkLoaderTest.php
assertchris: I laughed at "keeping our tests fast". Sure, I'll reduce the number of rows ;)

dev/CsvBulkLoader.php
chillu: Don't specify a length. That's what `fgetcsv()` does by default as well (http://php.net/manual/en/function.fgetcsv.php)

assertchris: ah, good idea.

dev/CsvBulkLoader.php
assertchris: If I put the unlink in catch, these files will only be removed if there is some error...

dev/CsvBulkLoader.php
```diff
@@ -47,8 +56,115 @@ public function preview($filepath) {
...
+
+				$next = $this->processChunk($file, false);
+
+				if ($result instanceof BulkLoader_Result) {
+					$result->merge($next);
+				} else {
+					$result = $next;
+				}
+
+				@unlink($file);
```
chillu: Hmm, `@` error suppression? As a dev, I'd want to know if deleting this file failed - it might cause potentially sensitive uploads to stick around longer than desired on the temporary path.

chillu:
Sorry for the late feedback Chris, other than that `@` thing above it's looking great! :)

Issue #4550, opened 2015-08-27
NumericField number type not accept comma value


hutlim:
After change input type to number, it's unable to display the value which contain the comma like 1,000.00

```
public function setValue($value, $data = array()) {
	require_once "Zend/Locale/Format.php";

	// If passing in a non-string number, or a value
	// directly from a DataObject then localise this number

	if(is_int($value) || is_float($value) || $data instanceof DataObject) {
		$locale = new Zend_Locale($this->getLocale());

		$this->value = Zend_Locale_Format::toNumber(
			$value,
			array('locale' => $locale)
		);
	} else {
		$this->value = $this->clean($value);
	}

	return $this;
}
```

```
public function getAttributes() {
	return array_merge(parent::getAttributes(), array(
		'type' => 'number',
		'step' => 'any' // allows entering float/decimal numbers like "1.2" instead of just integers
	));
}
```

dhensby:
Which version of SS; @tractorcow did some work on supporting this a while ago now...

dhensby:
https://github.com/silverstripe/silverstripe-framework/pull/3008

released in 3.1.10

hutlim:
it's version 3.2
not the locale issue. is the html5 number input type issue.
number input type only display numeric value, value with comma like 1,000 is not display in html5 number input

tractorcow:
@hutlim which locales are you testing it in? This field is very locale sensitive.

The backend for this field should be a numeric type, not a string type. If you are attempting to store string values such as `1,000,000` as varchar you'll run into issues. ;) Can you please clarify your model?

tractorcow:
Ok, I've read everything again and I think you are right about there being an issue with html5 support for this field. I'll look into it. :)

tractorcow:
Wow, it appears drupal encountered this issue as well.

https://www.drupal.org/node/2290029

It looks like the resulting decision was to drop support for `number` html5 type and just use `text`.

tractorcow:
Fix at https://github.com/silverstripe/silverstripe-framework/pull/4565

PR #4565, opened 2015-08-31
BUG Remove html5 number field due to insufficient localisation support

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/4550

It looks as though drupal encountered this same issue. See https://www.drupal.org/node/2290029 for the discussion. We may as well save the discussion and concur with them on the current solution.

I feel that accurate localisation is more important to silverstripe than conforming to standards for the sake of standardisation (although we prefer it where these align with one another).

One day when html5 number supports thousands separators, as per proper numeric localisation, we can re-enable this.

An alternative solution (which I'm going to reject) is to strip thousands separators. If there's a nice locale-safe way to do so, then we could do this, but since there are at least three different thousands separator formats (space, comma, period, and maybe more) I don't feel like it's wise to expose to so many edge cases. It would also reduce the localisation quality of the resulting string.

### DIFF
--- a/forms/NumericField.php
+++ b/forms/NumericField.php
@@ -82,13 +82,6 @@ class NumericField extends TextField {
 		return 'numeric text';
 	}
 
-	public function getAttributes() {
-		return array_merge(parent::getAttributes(), array(
-			'type' => 'number',
-			'step' => 'any' // allows entering float/decimal numbers like "1.2" instead of just integers
-		));
-	}
-
 	/**
 	 * Validate this field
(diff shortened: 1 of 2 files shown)


chillu:
I concur. But you'll need to fix `NumericFieldTest::testNumberTypeOnInputHtml`

tractorcow:
oh whoops. ;)

tractorcow:
Ok, fixed. :)

PR #4568, opened 2015-09-01
API Cleanup and refactor of select fields and Standardise Relation interface

tractorcow:
This change is focused on consolidating and improving the general quality of the single and multi-select formfields.

In order to improve the substitutability of these fields a new class hierarchy has been arranged:

* SelectField (abstract)
  * MultiSelectField (abstract)
    * ListboxField
    * LookupField
    * CheckboxSetField
  * SingleSelectField (abstract)
    * DropdownField
      * CountryDropdownField
      * GroupedDropdownField
    * OptionsetField
      * MemberDatetimeOptionsetField
      
      
This marks a change from "everything extends DropdownField" which is frankly dumb, because you can't substitute a checkboxsetfield where-ever a dropdown exists.

Validation has been moved to the two base classes for each branch, along with most data-processing functions. All fields now support SS_List, including Relations and SS_Map as data sources, as well as arrays.

In order to make `Relation` more useful, I've created an interface to represent these relations, and applied them to both `UnsavedRelationList` and `RelationList` (where once they had no shared parent other than `SS_List`).

I hacked around with chosen to get empty options to work properly with allow_single_deselect in a html5 standard way.

@sminnee the bugfix in TestRunner seems to be an oversight to your monolog pull request. I hacked in something that _might_ be the correct fix?

Dumb stuff has been deleted. You can still use a Varchar for your multi-select fields, but it only works with json_encoded values now (where before some fields kind of supported it, others supported some other odd format).

All the deprecated methods in `UnsavedRelationList` have been un-deprecated. They were a hangover from when this class USED to extend DataList. All of those methods work fine on ArrayList, however.

`ListboxField::setMultiple` is gone. It's always a multiple select field. Use DropdownField if you want single-select.

More formfields now are rendered via templates, where before they were manually generated via PHP (ew).

Other pull requests to merge with this:

 - https://github.com/silverstripe/silverstripe-cms/pull/1271
 - https://github.com/silverstripe/silverstripe-siteconfig/pull/8

### DIFF
--- a/admin/thirdparty/chosen/chosen/chosen.jquery.js
+++ b/admin/thirdparty/chosen/chosen/chosen.jquery.js
@@ -49,7 +49,8 @@
 
     SelectParser.prototype.add_option = function(option, group_position, group_disabled) {
       if (option.nodeName.toUpperCase() === "OPTION") {
-        if (option.text !== "") {
+        // workaround for https://github.com/harvesthq/chosen/issues/2125
+        if (!option.text.match(/^\s*$/g)) {
           if (group_position != null) {
             this.parsed[group_position].children += 1;
           }
@@ -135,7 +136,12 @@ Copyright (c) 2011 by Harvest
       this.results_showing = false;
       this.result_highlighted = null;
       this.result_single_selected = null;
-      this.allow_single_deselect = (this.options.allow_single_deselect != null) && (this.form_field.options[0] != null) && this.form_field.options[0].text === "" ? this.options.allow_single_deselect : false;
+      this.allow_single_deselect = (this.options.allow_single_deselect != null)
+			  && (this.form_field.options[0] != null)
+			  // workaround for https://github.com/harvesthq/chosen/issues/2125
+			  && this.form_field.options[0].text.match(/^\s*$/g)
+				? this.options.allow_single_deselect
+				: false;
       this.disable_search_threshold = this.options.disable_search_threshold || 0;
       this.disable_search = this.options.disable_search || false;
       this.search_contains = this.options.search_contains || false;
--- a/forms/CountryDropdownField.php
+++ b/forms/CountryDropdownField.php
@@ -11,13 +11,18 @@ class CountryDropdownField extends DropdownField {
 
 	/**
 	 * Should we default the dropdown to the region determined from the user's locale?
+	 *
+	 * @config
 	 * @var bool
 	 */
 	private static $default_to_locale = true;
 
 	/**
-	 * The region code to default to if default_to_locale is set to false, or we can't determine a region from a locale
-	 *  @var string
+	 * The region code to default to if default_to_locale is set to false, or we can't
+	 * determine a region from a locale.
+	 *
+	 * @config
+	 * @var string
 	 */
 	private static $default_country = 'NZ';
 
@@ -28,48 +33,57 @@ class CountryDropdownField extends DropdownField {
 	 * @return string
 	 */
 	protected function locale() {
-		if (($member = Member::currentUser()) && $member->Locale) return $member->Locale;
+		if (($member = Member::currentUser()) && $member->Locale) {
+			return $member->Locale;
+		}
 		return i18n::get_locale();
 	}
 
-	public function __construct($name, $title = null, $source = null, $value = "", $form=null) {
-		if(!is_array($source)) {
-			// Get a list of countries from Zend
-			$source = Zend_Locale::getTranslationList('territory', $this->locale(), 2);
-
-			// We want them ordered by display name, not country code
-
-			// PHP 5.3 has an extension that sorts UTF-8 strings correctly
-			if (class_exists('Collator') && ($collator = Collator::create($this->locale()))) {
-				$collator->asort($source);
-			}
-			// Otherwise just put up with them being weirdly ordered for now
-			else {
-				asort($source);
-			}
-
-			// We don't want "unknown country" as an option
-			unset($source['ZZ']);
+	public function setSource($source) {
+		if($source) {
+			return parent::setSource($source);
 		}
 
-		parent::__construct($name, ($title===null) ? $name : $title, $source, $value, $form);
+		// map empty source to country list
+		// Get a list of countries from Zend
+		$source = Zend_Locale::getTranslationList('territory', $this->locale(), 2);
+
+		// We want them ordered by display name, not country code
+
+		// PHP 5.3 has an extension that sorts UTF-8 strings correctly
+		if (class_exists('Collator') && ($collator = Collator::create($this->locale()))) {
+			$collator->asort($source);
+		} else {
+			// Otherwise just put up with them being weirdly ordered for now
+			asort($source);
+		}
+
+		// We don't want "unknown country" as an option
+		unset($source['ZZ']);
+
+		return parent::setSource($source);
 	}
 
 	public function Field($properties = array()) {
 		$source = $this->getSource();
 
-		if (!$this->value || !isset($source[$this->value])) {
-			if ($this->config()->default_to_locale && $this->locale()) {
-				$locale = new Zend_Locale();
-				$locale->setLocale($this->locale());
-				$this->value = $locale->getRegion();
-			}
+		// Default value to best availabel locale
+		$value = $this->Value();
+		if ($this->config()->default_to_locale
+			&& (!$value || !isset($source[$value]))
+			&& $this->locale()
+		) {
+			$locale = new Zend_Locale();
+			$locale->setLocale($this->locale());
+			$value = $locale->getRegion();
+			$this->setValue($value);
 		}
 
-		if (!$this->value || !isset($source[$this->value])) {
-			$this->value = $this->config()->default_country;
+		// Default to default country otherwise
+		if (!$value || !isset($source[$value])) {
+			$this->setValue($this->config()->default_country);
 		}
 
-		return parent::Field();
+		return parent::Field($properties);
 	}
 }
--- a/forms/MemberDatetimeOptionsetField.php
+++ b/forms/MemberDatetimeOptionsetField.php
@@ -5,65 +5,69 @@
  */
 class MemberDatetimeOptionsetField extends OptionsetField {
 
+	const CUSTOM_OPTION = '__custom__';
+
 	public function Field($properties = array()) {
 		Requirements::javascript(FRAMEWORK_ADMIN_DIR . '/javascript/MemberDatetimeOptionsetField.js');
+		$options = array();
+		$odd = false;
 
-		$options = '';
-		$odd = 0;
-		$source = $this->getSource();
-
-		foreach($source as $key => $value) {
-			// convert the ID to an HTML safe value (dots are not replaced, as they are valid in an ID attribute)
-			$itemID = $this->id() . '_' . preg_replace('/[^\.a-zA-Z0-9\-\_]/', '_', $key);
-			if($key == $this->value) {
-				$useValue = false;
-				$checked = " checked=\"checked\"";
-			} else {
-				$checked = "";
+		// Add all options striped
+		$anySelected = false;
+		foreach($this->getSourceEmpty() as $value => $title) {
+			$odd = !$odd;
+			if(!$anySelected) {
+				$anySelected = $this->isSelectedValue($value, $this->Value());
 			}
-
-			$odd = ($odd + 1) % 2;
-			$extraClass = $odd ? "odd" : "even";
-			$extraClass .= " val" . preg_replace('/[^a-zA-Z0-9\-\_]/', '_', $key);
-			$disabled = ($this->disabled || in_array($key, $this->disabledItems)) ? "disabled=\"disabled\"" : "";
-			$ATT_key = Convert::raw2att($key);
-
-			$options .= "<li class=\"".$extraClass."\">"
-				. "<input id=\"$itemID\" name=\"$this->name\" type=\"radio\" value=\"$key\"$checked $disabled"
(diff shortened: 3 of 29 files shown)


tractorcow:
If you need a good reason why this should be merged.

image

tractorcow:
image

tractorcow:
Pushed an update to improve quality based on scrutinizer feedback.

admin/thirdparty/chosen/chosen/chosen.jquery.js
```diff
@@ -49,7 +49,8 @@
 
     SelectParser.prototype.add_option = function(option, group_position, group_disabled) {
       if (option.nodeName.toUpperCase() === "OPTION") {
-        if (option.text !== "") {
+        // workaround for https://github.com/harvesthq/chosen/issues/2125
+        if (!option.text.match(/^\s*$/g)) {
```
stojg: are the changes in this plugin an update from upstream or is it hotfixed?

stojg:
A valid effort, but I can see some non merge blocking issues with the OOP approach.

CountryDropdownField inherits from SingleSelectField, so therefor it would be hard to make a MultipleCountryDropdownField?

Some of these should probably be using more of composition than inheritance. But this PR will make it easier to refactor this is in the future?

admin/thirdparty/chosen/chosen/chosen.jquery.js
tractorcow: It's hotfixed, not fixed upstream, because I don't know coffeescript. =(

tractorcow:
A MultiCountryDropdownField would extend ListboxField. Class inheritance is determined by how substitutable an instance is, not simply a common theme. :) I could put some of the locale code into a trait if that helps.

tractorcow:
@sminnee namespaced `SilverStripe\Model\Relation` :)

admin/thirdparty/chosen/chosen/chosen.jquery.js
assertchris: That's unfortunate. What happens when someone updates Chosen in future?

forms/CountryDropdownField.php
```diff
@@ -28,48 +33,57 @@ class CountryDropdownField extends DropdownField {
...
 
-			// We don't want "unknown country" as an option
-			unset($source['ZZ']);
+		// PHP 5.3 has an extension that sorts UTF-8 strings correctly
+		if (class_exists('Collator') && ($collator = Collator::create($this->locale()))) {
+			$collator->asort($source);
+		} else {
+			// Otherwise just put up with them being weirdly ordered for now
+			asort($source);
 		}
```
assertchris: Many things like this should be extracted to smaller methods. Why does the constructor need to know about a global collator? How can that be tested? I mean: how would you test the field with a collator and then without?

assertchris:
I have two concerns:

1. Has this been tested with a set of commonly-used modules to make sure it doesn't break things too terribly? [Something like this...](https://gitlab.cwp.govt.nz/cwp/cwp-recipe-basic). It's one thing to pass the tests of the framework, at the same version. To be internally compatible, and allow some breaking changes. It's another to tell many developers their modules are useless. Think Python 2 vs. Python 3.

2. There are many classes with similar names and difficult-to-discern differences. Is it more complicated than it needs to be, just to avoid code duplication? Could you explain all these classes to someone who doesn't already know them from 3.2?

chillu:
Hard to peer review in its entirety, but the inheritance structure looks good. Regarding a chosen.js pull request for the upstream repo, I think we've lost that battle already anyway, and a more viable option forward is replacing the implementation with select2 which looks more promising anyway (see https://github.com/silverstripe/silverstripe-framework/pull/4284).

tractorcow:
Potentially renaming classes to the below:


* RelationField (abstract)
  * MultiField (abstract)
    * MultiOptionField (was ListboxField)
    * LookupField (unchanged, some read-only thing)
    * MultiCheckboxField (was CheckboxSetField)
  * SingleField (abstract)
    * OptionField (was DropdownField)
      * CountryOptionField (was CountryDropdownField)
      * GroupedOptionField (was GroupedDropdownField)
    * RadioField (was OptionsetField)
      * DateFormatField (was MemberDatetimeOptionsetField)

tractorcow:
Ok, I've taken on feedback about class naming. I had a catch up with a few Auckland staff a while back and we decided on the new suggested names for these classes (as above).

This is ready for review again. :)

tractorcow:
Other pull requests to merge with this:

 - https://github.com/silverstripe/silverstripe-cms/pull/1271
 - https://github.com/silverstripe/silverstripe-siteconfig/pull/8

tractorcow:
After more discussion, I've adjusted the class hierarchy to the below:

* ListField (abstract)
  * MultiField (abstract)
    * MultiSelectField (was ListboxField)
    * LookupField (unchanged, some read-only thing)
    * MultiCheckboxField (was CheckboxSetField)
  * SingleField (abstract)
    * SelectField (was DropdownField)
      * CountrySelectField (was CountryDropdownField)
      * GroupedSelectField (was GroupedDropdownField)
    * RadioField (was OptionsetField)
      * DateFormatField (was MemberDatetimeOptionsetField)

tractorcow:
I've also been suggested to rename `MultiCheckboxField` to `CheckboxGroupField` (which could possibly just stay as `CheckboxSetField`).

What do you think @chillu ?

assertchris:
On request of @tractorcow: I prefer an approach closer to the actual form input types...

* `BaseField` (maybe `abstract`)
  * `SelectField` (with `setAllowMultipleSelection` / `getAllowMultipleSelection`)
    * `GroupedSelectField`
    * `LookupField`
      * `CountryLookupField`
  * `CheckboxField`
  * `CheckboxGroupField`
  * `RadioField` (maybe `abstract`)
  * `RadioGroupField`
    * `DateFormatField`

tractorcow:
Ok, I think I understand better your concern. :)

Something I didn't clarify earlier is that part of this refactor was intended to untangle the "this field sometimes acts as multi-select and sometimes as single select", since the code for one or the other implementation often had to be duplicated in each class to handle all of the special cases.

The goal of the two base classes for single / multi select is to standardise and consolidate all of these duplicated operations.

We _could_ probably pull back in some functions into a common base class, however, and just use conditions to switch between each on demand. That could allow us to, for instance, implement SelectField as a replacement for both DropdownField and ListboxField, which essentially acted as the single and multi-select versions of one another.

I feel it's cleaner though to simply require users to choose the class that meets their use case, meaning there is no conditional logic in the inner workings.

assertchris:
@tractorcow that duplicated functionality can be moved to traits. And when you think about it, there are a small set of actions that can be done through these fields:

1. Nothing: the value is merely stored and retrieved by something else acting on the field.
2. Save a value to somewhere, like a `DataObject` field
3. Save a 1|many-1|many relationship

Those actions can be in traits, and the rest is how each field renders markup, and configures potential JavaScript. We do not need an insane hierarchy for this.

tractorcow:
Sure, I'll look into maybe dumping a level of hierarchy and using traits instead.

tractorcow:
OK, I've had a feeling that this PR has got too extensive, so I've dialed it back a bit.

The functionality has all been refactored into substitutable base classes (with no overlapping duplicate functionality), but I've abandoned the form field rename, which was really unnecessary upgrading overhead.

Maybe it can be merged as is then? It needs more review from a functionality perspective, though, before then. @dhensby maybe want to spot some obvious issues?

dhensby:
I would LOVE to spot some obvious issues ;)

admin/thirdparty/chosen/chosen/chosen.jquery.js
dhensby: We can't do this, we've had this problem too much with dropdown.

Some people do know coffeescript and fixed our coffee script and compiled, some people haven't and did this kind of fix and then coffee script got fixed again and compiled over the changes.

NO "FIXING" EXTERNAL LIBS PLEASE

May we don't fix this issue in this PR and instead we replace chosen with select2 in another PR for 4.0?

forms/MemberDatetimeOptionsetField.php
```diff
@@ -5,65 +5,69 @@
...
+		$option = $this->getFieldOption(
+			self::CUSTOM_OPTION,
+			_t('MemberDatetimeOptionsetField.Custom', 'Custom'),
+			$odd
+		);
+		$option->setField('isChecked', $isChecked);
+		$option->setField('CustomName', $this->getName().'[Custom]');
+		$option->setField('CustomValue', $this->Value());
+		if($this->Value()) {
+			$preview = Convert::raw2xml(Zend_Date::now()->toString($this->Value()));
```
dhensby: We shouldn't use "now" as the date string because it can be ambiguous. We should show a non-ambiguous date string for users to pick from. eg: 25/12/2015 (maybe the year can be current!?)

forms/MultiSelectField.php
```diff
@@ -0,0 +1,221 @@
...
+		}
+		
+		// If json deserialisation fails, then fallover to legacy format
+		$result = json_decode($value, true);
+		if($result !== false) {
+			return $result;
+		}
+		
+		// Parse data in legacy {comma} format
+		Deprecation::notice('4.0', 'Storing multi-selection values in comma separated format is deprecated');
```
dhensby: please we deprecate this in 3.x and just remove it from 4

forms/OptionsetField.php
```diff
@@ -50,35 +50,46 @@
...
+	 * @param mixed $value Value of the option
+	 * @param string $title Title of the option
+	 * @param boolean $odd True if this should be striped odd. Otherwise it should be striped even
+	 * @return ArrayData Field option
 	 */
-	protected $disabledItems = array();
+	protected function getFieldOption($value, $title, $odd) {
+		// Check selection
+		$selected = $this->isSelectedValue($value, $this->Value());
+		$itemID = $this->ID() . '_' . Convert::raw2htmlid($value);
```
dhensby: Would this be better as a function?

forms/OptionsetField.php
```diff
@@ -50,35 +50,46 @@
...
+		));
+	}
 
-		if($source) {
-			foreach($source as $value => $title) {
-				$itemID = $this->ID() . '_' . preg_replace('/[^a-zA-Z0-9]/', '', $value);
-				$odd = ($odd + 1) % 2;
-				$extraClass = $odd ? 'odd' : 'even';
-				$extraClass .= ' val' . preg_replace('/[^a-zA-Z0-9\-\_]/', '_', $value);
+	protected function getItemName() {
```
dhensby: What's this function for?

forms/OptionsetField.php
```diff
@@ -50,35 +50,46 @@
...
-		$odd = 0;
-		$options = array();
+		return new ArrayData(array(
+			'ID' => $itemID,
+			'Class' => $extraClass,
+			'Name' => $this->getItemName(),
+			'Value' => $value,
+			'Title' => $title,
+			'isChecked' => $selected,
+			'isDisabled' => $this->isDisabled() || in_array($value, $this->getDisabledItems()),
```
dhensby: perhaps `in_array($value, $this->getDisabledItems()` should be a function?

templates/forms/DropdownField.ss
```diff
@@ -1,5 +1,8 @@
 <select $AttributesHTML>
 <% loop $Options %>
-	<option value="$Value.XML"<% if $Selected %> selected="selected"<% end_if %><% if $Disabled %> disabled="disabled"<% end_if %>><% if Title %>$Title.XML<% else %>&nbsp;<% end_if %></option>
+	<option value="$Value.XML"
+		<% if $Selected %> selected="selected"<% end_if %>
+		<% if $Disabled %> disabled="disabled"<% end_if %>
+		><% if $Title %>$Title.XML<% else %>&nbsp;<% end_if %></option>
```
dhensby: I'd close the `<option>` on the next line

dhensby:
1. It would be nice to fix the issue where you can't have the same value in a dropdown twice (for example, having a country at the top of a dropdown as well as in place alphabetically).
2. `Zend_Locale` has an outdated list of countries, can we update it or pull it in with composer?
3. forms/MemberDatetimeOptionsetField.php should be part of /admin


Whilst 1 & 2 can probably be in a separate PR, they're worth mentioning.

dhensby:
ping @tractorcow :tractor: :cow2:

tractorcow:
Thanks Dan I'll fix up when I get some time this week

tractorcow:
@dhensby I've responded to all the line-comments.

> It would be nice to fix the issue where you can't have the same value in a dropdown twice (for example, having a country at the top of a dropdown as well as in place alphabetically).

Not sure how to solve that, at the moment. Maybe that feature should be a new pull request (to keep this one from getting overly huge).

> Zend_Locale has an outdated list of countries, can we update it or pull it in with composer?

I'm terrified of updating Zend library at the moment... it will break so many things. Best to make this a separate pull request.

> forms/MemberDatetimeOptionsetField.php should be part of /admin

@sminnee can you make a call on this? I'm not sure where it fits in with your namespacing plans.

dhensby:
Agreed with separate PRs for my first 2 suggestions

tractorcow:
Bugfix and rebase for 4.0 :)

dhensby:
sorry - I haven't seen this.. needs a rebase now :/

tractorcow:
Well, it's no longer ready for merge since something else on master changed... crikey

tractorcow:
Rebased; todo - come back and check tests pass later.

dhensby:
let's get this show on the road

tractorcow:
Woot, that's about 2 years of work finally merged. :)

SpiritLevel:
Well done on delivering the gudes.

PR #4571, opened 2015-09-02
Upload field: if upload button is not enabled, don't delete …

mikenz:
In the Upload field with 'canUpload' set to false, this forces the user to upload a file using /admin/assets/ then select in using the Upload field.
Once selected they can directly delete the file from the Upload field

This change also hides the Delete from Files button when Upload is hidden.


dhensby:
You made this PR against your master which looks like it has subsequently been updated :(

PR #4578, opened 2015-09-02
FIX Resampled images inherit source properties

### LINKED ISSUE #4569: BUG Image_Cached objects (resampled images) don't inherit all properties
When `Image_Cached` objects are created in `Image::getFormattedImage()` only the Title and ParentID properties are copied from the source `Image` object. All other properties, including any assigned by extensions, are ignored. This means you can't access any custom Image properties (such as a Caption or a set of [focus point](https://github.com/jonom/silverstripe-focuspoint) coordinates :wink:) on resampled image instances in templates. This is bad for chainabiilty and template flexibility.

I figure this is a bug, and hopefully not very hard to fix but I'm not sure the best way to go about it. I would think though that rather than specify the properties you *do* want to copy to the new Image_cached object, it would be better to specify the exclusions (ID and Filename?) and copy everything else.

jonom:
Ensure Image_Cached objects can access all the properties of the source image (fixes #4569)

### DIFF
--- a/model/Image.php
+++ b/model/Image.php
@@ -691,11 +691,7 @@ class Image extends File implements Flushable {
 				call_user_func_array(array($this, "generateFormattedImage"), $args);
 			}
 
-			$cached = new Image_Cached($cacheFile);
-			// Pass through the title so the templates can use it
-			$cached->Title = $this->Title;
-			// Pass through the parent, to store cached images in correct folder.
-			$cached->ParentID = $this->ParentID;
+			$cached = new Image_Cached($cacheFile, false, $this);
 			return $cached;
 		}
 	}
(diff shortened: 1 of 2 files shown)


model/Image.php
```diff
@@ -1037,8 +1033,9 @@ class Image_Cached extends Image {
 	 * @param boolean $isSingleton This this to true if this is a singleton() object, a stub for calling methods.
 	 *                             Singletons don't have their defaults set.
 	 */
-	public function __construct($filename = null, $isSingleton = false) {
+	public function __construct($filename = null, $isSingleton = false, Image $sourceImage = null) {
 		parent::__construct(array(), $isSingleton);
+		if ($sourceImage) $this->update($sourceImage->toMap());
```
tractorcow: Nice job.

jonom:
I regret not sending this to 3.2 branch... @tractorcow is it appropriate if I open a new PR to do that?

model/Image.php
jonom: In hindsight should I have done this?
```php
$properties = $sourceImage->toMap();
unset($properties['RecordClassName'], $properties['ClassName']);
if ($sourceImage) $this->update($properties);
```
is it hazardous to pass the RecordClassName and ClassName properties?

tractorcow: I'm not sure it makes a big difference.

tractorcow:
This would probably be a feature enhancement, and is too late for 3.2 sorry. =(

jonom:
No worries!

model/Image.php
jonom: It seems to cause a problem if you swap out the Image class using the injector. Couldn't say why exactly - but code above fixes the problem. Seems like a safe change so will open a PR on 3 if you agree?

The whole Image_Cached model is a bit awkward right now but think you've completely overhauled that for 4.0? :D

PR #4580, opened 2015-09-04
API Enable chained injector dependencies

tractorcow:
This feature will allow more advance configuration of changed dependencies (especially if chained across modules).

### DIFF
--- a/control/injector/Injector.php
+++ b/control/injector/Injector.php
@@ -543,14 +543,25 @@ class Injector {
 		}
 		$class = $spec['class'];
 
+		// Check type
+		if (!$type) {
+			$type = isset($spec['type']) ? $spec['type'] : null;
+		}
+
 		// create the object, using any constructor bindings
 		$constructorParams = array();
 		if (isset($spec['constructor']) && is_array($spec['constructor'])) {
 			$constructorParams = $spec['constructor'];
 		}
 
-		$factory = isset($spec['factory']) ? $this->get($spec['factory']) : $this->getObjectCreator();
-		$object = $factory->create($class, $constructorParams);
+		// Catch service references
+		if (strpos($class, '%$') === 0) {
+			$otherID = substr($class, 2);
+			$object = $this->get($otherID, $type != 'prototype', $constructorParams);
+		} else {
+			$factory = isset($spec['factory']) ? $this->get($spec['factory']) : $this->getObjectCreator();
+			$object = $factory->create($class, $constructorParams);
+		}
 
 		// figure out if we have a specific id set or not. In some cases, we might be instantiating objects
 		// that we don't manage directly; we don't want to store these in the service cache below
@@ -560,10 +571,6 @@ class Injector {
 
 		// now set the service in place if needbe. This is NOT done for prototype beans, as they're
 		// created anew each time
-		if (!$type) {
-			$type = isset($spec['type']) ? $spec['type'] : null;
-		}
-
 		if ($id && (!$type || $type != 'prototype')) {
 			// this ABSOLUTELY must be set before the object is injected.
 			// This prevents circular reference errors down the line
--- a/docs/en/02_Developer_Guides/05_Extending/05_Injector.md
+++ b/docs/en/02_Developer_Guides/05_Extending/05_Injector.md
@@ -225,6 +225,37 @@ Would setup the following
 * Look at the properties to be injected and look for the config for `MySQLDatabase`
 * Create a MySQLDatabase class, passing dbusername and dbpassword as the parameters to the constructor.
 
+## Chaining services
+
+Furthermore, you could use aliases for multiple or optional backend service via the used of chained references.
+
+
+	:::yaml
+	Injector:
+	  JSONProvider:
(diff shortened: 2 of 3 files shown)


control/injector/Injector.php
```diff
@@ -543,14 +543,25 @@ protected function instantiate($spec, $id=null, $type = null) {
...
 		$constructorParams = array();
 		if (isset($spec['constructor']) && is_array($spec['constructor'])) {
 			$constructorParams = $spec['constructor'];
 		}
 
-		$factory = isset($spec['factory']) ? $this->get($spec['factory']) : $this->getObjectCreator();
-		$object = $factory->create($class, $constructorParams);
+		// Catch service references
+		if (strpos($class, '%$') === 0) {
+			$otherID = substr($class, 2);
```
tractorcow: At a higher level the "raw" dependency config is stuck in an `array('class' => $spec)` so I've opted to keep it simple here and just check the classname.

sminnee:
Can we have some docs please?

sminnee:
I don't think that random tweaks to Injector are a good idea to merge. Can we get a bit more detail about why this is needed, an amendment to the docs that describes that use-case, and a test that covers that use-case?

tractorcow:
This is the typical use case where this is useful:

```yaml
---
Name: assetstore
---
Injector:
  FlysystemAssetStore:
    class: 'SilverStripe\Filesystem\Flysystem\FlysystemAssetStore'
    properties:
      Filesystem: '%$FlysystemBackend'
  # Assign as default story
  AssetStore: '%$FlysystemAssetStore'
```

Given this config, without this code trying to create an instance of `AssetStore` via injector gives the following error.

```
Class %$FlysystemAssetStore does not exist
ReflectionClass->__construct(%$FlysystemAssetStore)
InjectionCreator.php:14

InjectionCreator->create(%$FlysystemAssetStore,Array)
Injector.php:563

Injector->instantiate(Array,AssetStore)
Injector.php:892

Injector->get(AssetStore)
DBFileTest.php:5

--------------------
```

I could construct my yml this way


```yaml
Injector:
  AssetStore:
    class: 'SilverStripe\Filesystem\Flysystem\FlysystemAssetStore'
    properties:
      Filesystem: '%$FlysystemBackend'
  ```
  
  But now to override it with a store that doesn't use flysystem, I'd need to do something like the below:
  
  
```yaml
Injector:
  AssetStore:
    class: 'SomeOtherClass'
    properties:
      Filesystem: null
```

Whereas with this code, I could simply do

```yaml
Injector:
  AssetStore: '%$SomeOtherService'
```

tractorcow:
Documentation with another example added, and I've updated and renamed the test.

sminnee:
Hrm, I think that the underlying issue here is probably this part:

> But now to override it with a store that doesn't use flysystem, I'd need to do something like the below

It's too hard to override a module-default configuration. I'm a bit worried if we're expecting that all module-default injector configurations need to use the chaining syntax to avoid this issue.

Injector probably needs to provide an easier way of replacing an entire service definition, something like this:

    Injector:
      AssetStore[replace]:
        class: 'SomeOtherClass'

I'm not convinced that it's a good idea to merge this change. I'm not expecting that its absence should block the filesystem work, should it, @tractorcow?

I think it would be better to think through a few limitations of Injector and propose a more broadly set of changes to it for SS4. Otherwise we risk chasing our tail.

tractorcow:
We already discussed merge strategies for config a while back; I think our conclusion is that it was possible but it would have more performance impact.

I don't think chaining is necessarily a poor pattern, since it only adds flexibility, and doesn't actually oblige users to utilise it. It solves the same problem that php variables do; An assigned value can be delegated to another variable (or passed to a method). This feature just makes the configuration we define a lot more transportable. Is that the right word?

Without it, users need to be careful about how they assign configs for services, since it could easily be done in such a way as to make it difficult to override. Up until now we haven't had this problem, but now that we are defining complex configuration for things such as logging and asset abstraction, it's going to be necessary to change something.

sminnee:
It would be good to get some other views here—I'm willing to be swayed on this one. ;-)

stevie-mayhew:
I would like to see a way to over ride configuration values rather than merge - we run into the problem quite a lot when working with other developers modules (or our own :smile:)

A common use case is where we have to use the PHP configuration layer to do something like the following to remove a tab from a model admin.

```php
Config::inst()->remove('ShopAdmin', 'managed_models');
Config::inst()->update('ShopAdmin', 'managed_models', array(
    'Order',
    'ShopConfig'
));
```

This would be better as something like 
```
Config::inst()->replace('ShopAdmin', 'managed_models', array(
    'Order',
    'ShopConfig'
));
```

And a YML layer configuration such as the one described by @sminnee would be a good move as well, IMHO. What happens if you are using a module which defines a default asset store, but you want to have another one running in parallel? Or replace the default store with another, then add a third?

tractorcow:
I want both those features AND this one. :)

tractorcow:
Closing this PR since it doesn't seem to be going anywhere.

PR #4582, opened 2015-09-04
API showqueries=1 now shows parameters

tractorcow:
It makes debugging a bit nicer. Rather than seeing `WHERE "Column" = ?` you can see the parameters as a separate following list.

### DIFF
--- a/model/connect/Database.php
+++ b/model/connect/Database.php
@@ -132,7 +132,8 @@ abstract class SS_Database {
 			$sql,
 			function($sql) use($connector, $parameters, $errorLevel) {
 				return $connector->preparedQuery($sql, $parameters, $errorLevel);
-			}
+			},
+			$parameters
 		);
 	}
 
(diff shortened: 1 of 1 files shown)


kinglozzer:
Makes sense, are the parameters quoted?

`params: "foo","bar, baz","bam"` or `params: foo,bar,baz,bam`?

tractorcow:
nope. ;D

tractorcow:
Ok now it does.

dhensby:
Would it not be possible just to replace the `?` in place so it looks more like a real query... which would be much more helpful for debugging...

kinglozzer:
@dhensby I considered that but I prefer having them separate - it’s more verbose, but it’s more obvious that the query is parameterised. Doctrine’s DBAL does the same (in fact, you have to manually output the SQL and the parameters separately)

dhensby:
I understand that it's clearer that the query is parametrised, but is that really important?? We can just say (parametrised) at the beginning or end of the log.

I'm not sure than having to manually associate Nth param with Nth `?` is easy or helpful and also the reasoning "but others do it that way" doesn't convince me.

One of the things I used the query logging for is to be able to copy and past into a MySQL terminal to see outputs, errors, etc. This is made significantly more difficult by parametrising like this.

dhensby:
As another thought, you could just have ?showqueries echo  out "Using MySQLPDODatabase with parametrised queries" or something

kinglozzer:
A string replacement to produce what looks like a raw SQL query implies that “this is the SQL string that’s being run” when it technically isn’t, however I’m not sure in which scenario a parameterised query _wouldn’t_ actually match up to its find+replace equivalent? I know that they do escaping etc, but do DBMSs ever transform the values of parameters before execution in a way that would render this sort of debugging useless?

You’re probably right that this is much less useful to those using `?showqueries` to copy + paste SQL for debugging though. If the find+replace version of the query is essentially equivalent to the parameterised one (at least for the purposes of debugging in development) then I don’t see a problem with adding that functionality.

dhensby:
The DB layer should know to expect an int or a string or a date, so it'll quote the value accordingly, which we wouldn't be doing for this, I guess...

PR #4583, opened 2015-09-04
API Refactor CompositeDBField into an abstract class

tractorcow:
CompositeDBField is now an interface to nested fields on an underlying dataobject, allowing field manipulation to be performed at the field and dataobject level without having to synchronise them manually.

Cleanup and rationalisation of changes:

* ViewableData::castingHelper is overridden in DataObject rather than it trying to do magic reflection on its subclasses. Better to override than try to build in too much logic into a base class.
* Money::getName Renamed to getCurrencyName, since it conflicted with DBField::getName which had a different purpose.
* markChanged parameter added to DBField::setValue in order to make the API consistent between CompositeDBField and DBField.
* DBField::setValue has been updated in a few instances in order to better interact with any parent dataobject. PrimaryKey will use the $record parameter for generating a default formfield, and DBField objects will record which tables they should use for augmenting dataqueries (so better SQL is generated).
* All existing compositeDBFields were re-written to support this being an abstract base class, instead of an interface.
* CompositeDBField now works closer as an interface on top of an underlying DataObject, with get/setField and dbObject to support casting and modification of its sub-field data. As with DBField, setValue is used to bind it to a parent object with which it uses as a data source. Check CompositeDBFieldTest::testFieldBinding for an example of how this works in practice.
* As a result of some of this API change, special cases in DataObject::dbObject have been consolidated together. $fixed_fields['ID'] has been changed from type 'Int' to 'PrimaryKey' (because that's what it is).
* DataObject::get/setField changed to resolve regressions in assigning values as DBField instances.
* DataObject::getChangedFields had to be changed to fix an infinite loop in CompositeDBField::isChanged.
* DataObject::composite_fields now includes polymorphic foreign key fields.
* PolymorphicForeignKey has been changed from Int/Enum to Int/Varchar due to it performing excessive schema queries. Possibly worth looking into again in the future (if we can force it to only query during dev/build).

### DIFF
--- a/forms/MoneyField.php
+++ b/forms/MoneyField.php
@@ -106,15 +106,18 @@ class MoneyField extends FormField {
 	 * (see @link MoneyFieldTest_CustomSetter_Object for more information)
 	 */
 	public function saveInto(DataObjectInterface $dataObject) {
-		$fieldName = $this->name;
+		$fieldName = $this->getName();
 		if($dataObject->hasMethod("set$fieldName")) {
 			$dataObject->$fieldName = DBField::create_field('Money', array(
 				"Currency" => $this->fieldCurrency->dataValue(),
 				"Amount" => $this->fieldAmount->dataValue()
 			));
 		} else {
-			$dataObject->$fieldName->setCurrency($this->fieldCurrency->dataValue());
-			$dataObject->$fieldName->setAmount($this->fieldAmount->dataValue());
+			$currencyField = "{$fieldName}Currency";
+			$amountField = "{$fieldName}Amount";
+
+			$dataObject->$currencyField = $this->fieldCurrency->dataValue();
+			$dataObject->$amountField = $this->fieldAmount->dataValue();
 		}
 	}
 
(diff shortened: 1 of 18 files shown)


tractorcow:
I'm going to keep making these pull requests until we have no more code left.

image

tractorcow:
Don't merge this I have to review this more I think before it's ready

tractorcow:
I want to see if I can refactor the old polymorphic classname detection into `PolymorphicForeignKey::requireField()`

tractorcow:
Thing to fix:
- move ClassName into its own dbfield (extends Enum) that knows to only do DB queries for existing classes when doing a schema update
- ensure that dbfield::table is set to the correct table (rather than just assuming get_class($record)).

tractorcow:
setTable is fixed in the forked branch.

I'll re-open this once I resolve the former issue.

dhensby:
@tractorcow will you ever re-open?

tractorcow:
This was merged with #4590

Issue #1913, opened 2013-05-13
[2010-03-07] Allow setting utf8_general_ci to one of the other utf8 variants


silverstripe-issues:
created by: @MarcusDalgren (smurkas)
created at: 2010-03-07
original ticket: http://open.silverstripe.org/ticket/5145

---

In general utf8_general_ci will work ok on a non English site but when running on utf8_general_ci MySQL doesn't understand the sorting order of special characters from other languages.

So for instance in Swedish ,  and  get treated as the characters a and o when sorting a result set. As far as I know the only way to fix this is to set the relevant columns to utf8_swedish_ci so that MySQL knows how to treat the special characters.

chillu:
Configurable via `Config::inst()->get('SilverStripe\ORM\Connect\MySQLDatabase', 'connection_collation')` and `$collation = Config::inst()->get('SilverStripe\ORM\Connect\MySQLDatabase', 'collation')` now

PR #4584, opened 2015-09-05
UPDATE: Add support for user defined charset and collation

Olliepop:
Support for user defined charset and collation. In my case I need to support emoji's as I'm writing the backend of a mobile app with SilverStripe. Therefore I need my columns set to `charset utf8mb4` and `collate utf8mb4_unicode_ci`. 

I've set the default collation `MySQLDatabase connection_collation` to utf8_unicode_ci. 

I didn't understand why there was a config option `connection_charset` if it had no effect on the database columns. Took a look and found that utf8 is defined statically everywhere. 

Fixes https://github.com/silverstripe/silverstripe-framework/issues/1913

### DIFF
--- a/model/fieldtypes/Varchar.php
+++ b/model/fieldtypes/Varchar.php
@@ -50,11 +50,14 @@ class Varchar extends StringField {
  	 * @see DBField::requireField()
  	 */
 	public function requireField() {
+		$charset = Config::inst()->get('MySQLDatabase', 'connection_charset');
+		$collation = Config::inst()->get('MySQLDatabase', 'connection_collation');
+
 		$parts = array(
 			'datatype'=>'varchar',
 			'precision'=>$this->size,
-			'character set'=>'utf8',
-			'collate'=>'utf8_general_ci',
+			'character set'=> $charset,
+			'collate'=> $collation,
 			'arrayValue'=>$this->arrayValue
 		);
 
--- a/model/fieldtypes/Varchar.php
+++ b/model/fieldtypes/Varchar.php
@@ -50,8 +50,8 @@ class Varchar extends StringField {
  	 * @see DBField::requireField()
  	 */
 	public function requireField() {
-		$charset = Config::inst()->get('MySQLDatabase', 'connection_charset');
-		$collation = Config::inst()->get('MySQLDatabase', 'connection_collation');
+		$charset = Config::inst()->get('MySQLDatabase', 'charset');
+		$collation = Config::inst()->get('MySQLDatabase', 'collation');
 
 		$parts = array(
 			'datatype'=>'varchar',
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -6,7 +6,8 @@ Upload:
   replaceFile: false
 MySQLDatabase:
   connection_charset: utf8
-  connection_collation: utf8_general_ci
+  charset: utf8
+  collation: utf8_general_ci
 HTTP:
   cache_control:
     max-age: 0
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -5,6 +5,8 @@ Upload:
   # Replace an existing file rather than renaming the new one.
   replaceFile: false
 MySQLDatabase:
+  # You are advised to backup your tables if changing settings on an existing database
+  # `connection_charset` and `charset` should be equal, similarly so should `connection_collation` and `collation`
   connection_charset: utf8
   connection_collation: utf8_general_ci
   charset: utf8
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -6,6 +6,7 @@ Upload:
   replaceFile: false
 MySQLDatabase:
   connection_charset: utf8
+  connection_collation: utf8_unicode_ci
 HTTP:
   cache_control:
     max-age: 0
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -6,6 +6,7 @@ Upload:
   replaceFile: false
 MySQLDatabase:
   connection_charset: utf8
+  connection_collation: utf8_general_ci
   charset: utf8
   collation: utf8_general_ci
 HTTP:
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -6,7 +6,7 @@ Upload:
   replaceFile: false
 MySQLDatabase:
   connection_charset: utf8
-  connection_collation: utf8_unicode_ci
+  connection_collation: utf8_general_ci
 HTTP:
   cache_control:
     max-age: 0
--- a/model/connect/MySQLDatabase.php
+++ b/model/connect/MySQLDatabase.php
@@ -159,7 +159,7 @@ class MySQLDatabase extends SS_Database {
 			$baseClasses[$class] = '"' . $class . '"';
 		}
 
-		$charset = Config::inst()->get('MySQLDatabase', 'connection_charset');
+		$charset = Config::inst()->get('MySQLDatabase', 'charset');
 
 		// Make column selection lists
 		$select = array(
--- a/model/connect/MySQLDatabase.php
+++ b/model/connect/MySQLDatabase.php
@@ -31,6 +31,13 @@ class MySQLDatabase extends SS_Database {
 			$parameters['charset'] = $charset;
 		}
 
+		// Set collation
+		if( empty($parameters['collation'])
+			&& ($collation = Config::inst()->get('MySQLDatabase', 'connection_collation'))
+		) {
+			$parameters['collation'] = $collation;
+		}
+
 		// Notify connector of parameters
 		$this->connector->connect($parameters);
 
--- a/model/connect/MySQLDatabase.php
+++ b/model/connect/MySQLDatabase.php
@@ -159,18 +159,20 @@ class MySQLDatabase extends SS_Database {
 			$baseClasses[$class] = '"' . $class . '"';
 		}
 
+		$charset = Config::inst()->get('MySQLDatabase', 'connection_charset');
+
 		// Make column selection lists
 		$select = array(
 			'SiteTree' => array(
 				"ClassName", "$baseClasses[SiteTree].\"ID\"", "ParentID",
 				"Title", "MenuTitle", "URLSegment", "Content",
 				"LastEdited", "Created",
-				"Filename" => "_utf8''", "Name" => "_utf8''",
+				"Filename" => "_{$charset}''", "Name" => "_{$charset}''",
 				"Relevance" => $relevance['SiteTree'], "CanViewType"
 			),
 			'File' => array(
-				"ClassName", "$baseClasses[File].\"ID\"", "ParentID" => "_utf8''",
-				"Title", "MenuTitle" => "_utf8''", "URLSegment" => "_utf8''", "Content",
+				"ClassName", "$baseClasses[File].\"ID\"", "ParentID" => "_{$charset}''",
+				"Title", "MenuTitle" => "_{$charset}''", "URLSegment" => "_{$charset}''", "Content",
 				"LastEdited", "Created",
 				"Filename", "Name",
 				"Relevance" => $relevance['File'], "CanViewType" => "NULL"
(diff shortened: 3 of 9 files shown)


kinglozzer:
I don’t know much about character sets & collations, but one thing that springs to mind here: what if you’re not using MySQL? You’re injecting the value of `MySQLDatabase.connection_charset` - is this something that only applies to MySQL? If not, should it be renamed so it’s generic for other database types?

Olliepop:
It would be ideal to define different character sets for different database connectors. For example in my case I require the utf8mb4 character set because the MySQL implementation of UTF8 is virtually broken but that's another story (see http://www.codeka.com.au/blog/2014/02/unicode-support-in-mysql-is--). However if I was to migrate to Postgres then utf8 would be fine.

Olliepop:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/1913

chillu:
Interesting issue with MySQL and Emojis - I can see this becoming more common over time.

I think reusing the existing `MySQLDatabase.connection_charset` config setting for the charset on database content (rather than hardcoding `utf8`) is a bit misleading for anybody who's familiar with the MySQL character handling. There's a difference between a connection charset and a database content charset (database, table, column). Compare https://dev.mysql.com/doc/refman/5.5/en/charset-connection.html with https://dev.mysql.com/doc/refman/5.0/en/charset-column.html. So while very uncommon, MySQL allows you to send statements in one charset (e.g. ISO8859-1), and then convert them into another charset on write (e.g. UTF8).

So I'm all for removing the hardcoding, but it should be through a new set of config variables (`MySQLDatabase.charset` and `MySQLDatabase.collation`?). These should be well documented to ensure they're changed in sync with the connection equivalents. I'm fine with having this MySQL specific for now, and patching the other DB connectors on demand.

Olliepop:
Good point @chillu - just implemented your suggestion

chillu:
I think for completeness sake, we should also add `connection_collation` (https://dev.mysql.com/doc/refman/5.0/en/charset-connection.html). And then add a comment in `config.yml` to point out two important things: 1. It's not recommended to have different connection and data settings, change them at the same time. 2. Avoid changing data settings on existing databases, could be lossy. I haven't really read up much about gotchas regarding charset handling in MySQL, but we should make it less likely that somebody's going to shoot themselves in the foot with it ;)

Olliepop:
Let me know if there's anything else you guys would like to see to get this merged.

patricknelson:
+1 on this. My only recommendation here is, if indeed someone is using a completely different database abstraction, it would make sense to *not* be setting (and then pulling) defaults for the `StringField` descendant instances (like `Enum`, `Varchar`, etc) or some other class for values like `charset` and `collation` from the MySQLDatabase config, strictly due to separation of concerns. As long as it's not associated with a particular kind of database. 

I think that the nomenclature of `utf8` (for internal reference) should be an internal nomenclature (coincidentally matching MySQL's) but detached from the actual *implementation* of which charset you actually call it on the database-end (e.g. it could be called `UTF8` in other databases and could be case-sensitive). At that level, you'd then use the `MySQLSchemaManager` (in  `4.x` or `MySQLDatabase` in `3.x`) to perform the database-level translation.

chillu:
Sorry, this dropped off my radar. 

Looking through the diff again, I agree with @kinglozzer and @patricknelson regarding the use of `MySQLDatabase` configuration values in `DBField` classes - we should leave this to the database driver. This would remove the ability to have per-field character sets and collations, which I think would be acceptable.

But in the end, we're making the current situation slightly better: At the moment the `character set` and `collate` keys are added by `Varchar->requireField()`, and subsequently ignored by implementations other than MySQL (see Postgres: https://github.com/silverstripe/silverstripe-postgresql/blob/master/code/PostgreSQLSchemaManager.php#L1206). So we're just making the current hardcoding more obvious by explicitly requesting the `MySQLDatbase` values. 

MSSQL has its own configuration value for collation: https://github.com/silverstripe/silverstripe-mssql/blob/master/code/MSSQLSchemaManager.php#L785 and https://github.com/silverstripe/silverstripe-mssql/blob/master/code/MSSQLDatabase.php#L84. It also uses `UTF-8` for the connection, so slightly different notation from MySQL.

So in this context, I'm happy to merge as-is - but keen to get agreement with more core committers first. /cc @tractorcow

tractorcow:
It looks fine to me. ;D

I wonder if we need to create test fixtures for schema-specific tests across different collations? I'm not sure how well the framework supports case sensitive collation, for instance. I know that SearchFilters have explicit support for case sensitive or case insensitive filtering.

FinBoWa:
Just have to comment on this: A pull request on this has been on my list for a long time and really happy to see that some on has finally done this. 

This has a big impact (a big shout out to you @Olliepop on creating a pull request!!) on easing issues with non English languages, for example my Finnish, on getting correctly sorted data out of the database directly. We ended to use our forked version where we use the Swedish collation with some nasty hard coding. 

So I'm really eager to see this as a patch on 3.1.x soon as possible and also to the 3.2 branch :)

I was also thinking on suggesting a change on default from collation to  utf8_unicode_ci .

http://stackoverflow.com/questions/766809/whats-the-difference-between-utf8-general-ci-and-utf8-unicode-ci

As latter would supper better sorting out of the box but that would be a bc change then. Also that causes a slighly performance hit apparently.

patricknelson:
Bump. Why is this languishing?

I'm still relying on a custom implementation of the `MySQLDatabase` class just to pull off supporting multibyte character sets in MySQL.

tractorcow:
Ah, this is a shame to have been left out. I apologise for letting this PR get left out.

We won't be able to push this in this side of christmas, but I'll see if I can give this issue some time in the new year.

tractorcow:
Maybe instead of encoding charset and collation directly into the DBFields, we don't declare them there at all, and allow the various DBSchemaManager classes to set these values as necessary on build? (falling over to the old hard coded values for managers which don't support those values, for backwards compatibility).

patricknelson:
@tractorcow Are you responding to my comment here? https://github.com/silverstripe/silverstripe-framework/pull/4584#issuecomment-143620728 

I think the reasoning behind defining this in the `DBField` is because this allows a lot more flexibility in defining your own constraints on custom column types. And also because this can be defined at that resolution (and I may not **want** or *need* it at a global level for all instance of that column type). For example, here's what I ended up implementing on my end in order to simply support emoji's :smile: but got caught up with the lower level `MySQLDatabase` abstraction (circa `3.1`) didn't pass through my charset definitions :hankey: 

```php
class StorySocialPost extends StoryChild {
	// ... redacted excess code...

	// Database fields.
	private static $db = [
		// ... redacted excess code...
		"PostText" => "TextUTF8MB4",
		"PostHTML" => "TextUTF8MB4",
	];

	// ... redacted excess code...
}
```

Then simply defined my own custom column type. Ideally maybe in the future I could setup (or utilize new) classes which accept that as a value in the constructor, so I could pass `Text(UTF8MB4)` which might allow more flexibility since then I wouldn't need to create a new class of my own (but I digress):

```php
<?php
/**
 * Provides compatibility for multi-byte UTF-8 which allows for storage of up to 4 bytes instead of the default 3
 * available in the default "utf8" MySQL storage capability. Unfortunately, it's not as easy as it looks below... this
 * still requires several things:
 *
 * - utf8mb4 database connection encoding
 * - a CustomMySQLDatabase implementation so that the below requirements are actually USED
 *
 * @author	Patrick Nelson, patnelson@ebay.com
 * @since	2015-09-24
 */

class TextUTF8MB4 extends Text {
	public function requireField() {
		$parts = array(
			'datatype' => 'mediumtext',
			'character set' => 'utf8mb4', // Unfortunately MySQLDatabase->text() ignores this, so CustomMySQLDatabase was invented.
			'collate' => 'utf8mb4_general_ci', // Same here.
			'arrayValue' => $this->arrayValue
		);

		$values= array(
			'type' => 'text',
			'parts' => $parts
		);

		DB::requireField($this->tableName, $this->name, $values, $this->default);
	}
}
```

Finally, I had to create my own injected/modified version of the schema manager (as it's now known) for 3.1:
```php
class CustomMySQLDatabase extends MySQLDatabase {

	public function text($values) {
		// Only deal with charset/collate settings right now because that's all I need and currently have patience for . . . . .
		$charset = "utf8_general_ci";
		if (isset($values["character set"])) $charset = $values["character set"];

		$collate = "utf8_general_ci";
		if (isset($values["collate"])) $collate = $values["collate"];

		return "mediumtext character set $charset collate $collate";
	}

}
```

So -- defining this at the column type level is, in my opinion, just fine, since there's already a logical/reasonable structure there to set an internal (i.e. SS-specific) preference for charset and collation. That's why in my above-linked comment I said that the internal nomenclature for UTF8 can be `utf8` which we would then *convert* to whatever the DB-specific nomenclature would be (in this case for MySQL it's actually still `utf8`). In PostgreSQL it might instead be `UTF8` (not sure if case-sensitive, so maybe bad example).

Also to me this may also be my way of saying "as necessary" (like you mentioned in your last comment). Abstracting away MySQL specific terminology away from `DBField` (which is a good thing) for better separation of concern (i.e. regardless of the schema manager).

dhensby:
this needs rebasing now - let's not let this fall off the radar!!!

stevie-mayhew:
Pinging @Olliepop on behalf of @dhensby :)

Olliepop:
@dhensby should I submit a new pr with support for this on the 3.3.0 or master branch?

tractorcow:
Sorry, we are in the middle of an RC at the moment.

3.3.0 = only critical bugfixes, and any pulls here will necessitate a new RC
3.3 = only bugfixes
3 = features

This is on the right branch; 3 should still be used for non-breaking features.

dhensby:
@Olliepop no, just [rebasing](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request#how-do-i-rebase)

It's just because your branch is just out of date (rather than targeting the wrong branch)

Olliepop:
@dhensby help

dhensby:
OK - It looks like your old commit was c3a874df6540dda96d2e516a6ac90a63d8e461b1 and you've merged your work with some other branch... not sure exactly.

Basically - you need to go back to the state the branch was in before it was all funky and do a rebase.

I've done this for you (https://github.com/dhensby/silverstripe-framework/tree/Olliepop-feature-config-charset-collation).

To get this PR right, you'll need to do this:

```sh
$ git remote add dhensby https://github.com/dhensby/silverstripe-framework
$ git fetch dhensby
$ git checkout feature-config-charset-collation
$ git reset --hard f513f4873f58bf82fba1e32fe546b450898abdcc
$ git push -f origin feature-config-charset-collation
```

Olliepop:
Thanks @dhensby - owe you a beer

model/fieldtypes/Varchar.php
```diff
@@ -50,11 +50,14 @@ public function getSize() {
...
+		$charset = Config::inst()->get('MySQLDatabase', 'charset');
+		$collation = Config::inst()->get('MySQLDatabase', 'collation');
+
 		$parts = array(
 			'datatype'=>'varchar',
 			'precision'=>$this->size,
-			'character set'=>'utf8',
-			'collate'=>'utf8_general_ci',
+			'character set'=> $charset,
+			'collate'=> $collation,
```
tractorcow: I'm not sure these array keys are even used anywhere... are they?

tractorcow: If that's the case, maybe it's best to drop all of the mysql-specific code from the DBField subclasses. Leave them as is for now, and maybe add a comment to remove these in 4.0. (to prevent semver regressions).

tractorcow:
Hi @Olliepop I gave this a good review and I think it's excellent. I just added a note about the mysql logic in dbfield subclasses being potentially unnecessary.

If you rebase and squash we can get this ready for merge.

dhensby:
merged with 758c0273fa988d942e22de5cfdd43acbec69cbf7

Thanks @Olliepop

PR #4587, opened 2015-09-07
Disable split view mode when not enough room to show preview

scott1702:
Disable the ability to change into 'split mode' if there isn't enough screen room to show the site preview, shows a tooltip on hover as to why split mode is disabled.
image

If the user is already in split mode and the site preview isn't shown, the user is notified as to why.


admin/javascript/LeftAndMain.Preview.js
```diff
@@ -1,4 +1,20 @@
 (function($) {
+
+	var disableSplitView = function (splitViewListItem) {
+
+		if ($('#pages-controller-cms-content').width() < 1340 && $('.cms-preview').width() < 1340 && !splitViewListItem.hasClass('result-selected')) {
```
tractorcow: Can you make this a bit more soft-coded? check out LeftAndMain.js / LayoutOptions for an example of this being done right. :) There is a minPreviewWidth already.

LeftAndMain.Layout.js is where this logic seems to be invoked (`obj.layout`). We should keep all layout logic logic like this in the same place shouldn't we? :P

admin/javascript/LeftAndMain.Preview.js
```diff
@@ -607,8 +622,13 @@
 		$('.cms-preview.column-hidden').entwine({
 			onmatch: function() {
 				$('#preview-mode-dropdown-in-content').show();
+				// Alert the user as to why the preview is hidden
+				if ($('.cms-preview .result-selected').hasClass('font-icon-columns')) {
+					statusMessage("Screen too small to show site preview in split mode", "error");
```
willmorgan: This should really be internationalised, no?

tractorcow: Yes, that's a good spot. 

This will need to go into https://github.com/silverstripe/silverstripe-framework/blob/3/admin/javascript/lang/src/en.js

scott1702: Ah thanks for that, updated

tractorcow:
I've tested this and it works perfectly.

PR #4588, opened 2015-09-07
BUG Fix missing framework/admin/tests

tractorcow:
These tests have been left out for who knows how long

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,7 @@ before_script:
 
 script:
  - "if [ \"$BEHAT_TEST\" = \"\" ]; then vendor/bin/phpunit framework/tests; fi"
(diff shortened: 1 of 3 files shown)


tractorcow:
It looks like postgres is having a little trouble with LeftAndMainTest::testUpdateTreeNodes

tractorcow:
Marking as impact high, as it resolves currently missing broken tests.

dhensby:
It seems a bit worrying that we have to ANSI quote the sort columns... (why) isn't this done automatically?

tractorcow:
Because it's not a simple column name; It's a sql fragment, and there's no automatic quoting of columns within fragments in framework.

It fails in postgres because it's less resilient to non-standard sql.

PR #4590, opened 2015-09-08
API Refactor CompositeDBField into an abstract class

tractorcow:
Adapted from https://github.com/silverstripe/silverstripe-framework/pull/4583

CompositeDBField is now an interface to nested fields on an underlying dataobject, allowing field manipulation to be performed at the field and dataobject level without having to synchronise them manually.

Cleanup and rationalisation of changes:

* ViewableData::castingHelper is overridden in DataObject rather than it trying to do magic reflection on its subclasses. Better to override than try to build in too much logic into a base class.
* Money::getName Renamed to getCurrencyName, since it conflicted with DBField::getName which had a different purpose.
* markChanged parameter added to DBField::setValue in order to make the API consistent between CompositeDBField and DBField.
* DBField::setValue has been updated in a few instances in order to better interact with any parent dataobject. PrimaryKey will use the $record parameter for generating a default formfield, and DBField objects will record which tables they should use for augmenting dataqueries (so better SQL is generated).
* All existing compositeDBFields were re-written to support this being an abstract base class, instead of an interface.
* CompositeDBField now works closer as an interface on top of an underlying DataObject, with get/setField and dbObject to support casting and modification of its sub-field data. As with DBField, setValue is used to bind it to a parent object with which it uses as a data source. Check CompositeDBFieldTest::testFieldBinding for an example of how this works in practice.
* As a result of some of this API change, special cases in DataObject::dbObject have been consolidated together. $fixed_fields['ID'] has been changed from type 'Int' to 'PrimaryKey' (because that's what it is).
* DataObject::get/setField changed to resolve regressions in assigning values as DBField instances.
* DataObject::getChangedFields had to be changed to fix an infinite loop in CompositeDBField::isChanged.
* DataObject::composite_fields now includes polymorphic foreign key fields.
* PolymorphicForeignKey has been changed from Int/Enum to Int/Varchar due to it performing excessive schema queries. Possibly worth looking into again in the future (if we can force it to only query during dev/build).
* ClassName field was refactored into a separate DBField subclass in order to better encapsulate the shared logic in DataObject and PolymorphicForeignKey. A better mechanism is used to ensure that obsolete-class queries are only performed during dev/build.


tractorcow:
also cc @sminnee who had started looking at the prior PR . :)

sminnee:
OK, I've reviewed the change, rather than create a new dbHelper() method, I believe that the following is in order:

 * Move the functionality of dbHelper() into db().
 * Mark it as an API change, explaining which new columns are now included. Add to upgrading guide.
 * Amend uses of db() as appropriate in response to that.

Alternatively, mark dbHelper() as experimental, and we can mop up all of these methods subsequently. I think that the following methods could use a re-think - there' just *way* too many of them:

 * database_fields()
 * custom_database_fields()
 * db()
 * dbHelper()
 * dbObject()
 * is_composite_field()
 * composite_fields()
 * getComponent()
 * getRelationClass()
 * getManyManyComponents()
 * hasOne()
 * hasOneComponent()
 * hasMany()
 * hasManyComponent()
 * belongsTo()
 * belongsToComponent()
 * manyMany()
 * manyManyComponent()
 * hasDatabaseField()
 * hasOwnTableDatabaseField()
 * has_own_table_database_field()
 * has_own_table() 
 * inheritedDatabaseFields()

tractorcow:
Hi @sminnee I've done a lot of work to fix this up the best I could. This is the gist of the change:

DataObject::database_fields now does the bulk of the work of incorporating all low level db fields together. This is used by higher level functions (such as dbobject() or db()) as well as for generating the db schema.

I found that "database_fields only returns some fields" to be unintuitive and misleading, so it also returns ID fields (for every table) and fixed_fields (for the base table). If user-code should ignore ID fields, then it's up to that code to either use custom_fields or to remove it explicitly.

I almost removed custom_database_fields but it's probably going to break more things that I'd like. We could still possibly deprecate it. :)

DataObject::db now acts as dbHelper once did, but retains its ability to either loop over all fields, or to extract specifications for a single one. Unlike database_fields, it treats logical fields (such as composite fields) the same way as physical fields. This means that it can be reliably used to provide a casting helper for any field the user requests, but it should not be used to determine what fields to create on a table.

dbHelper is gone.

A bug in HTTPOutputHandler.php is fixed. :)

You can now use $obj->CompositeField->SubField to get or assign values.

A bunch of caches have been consolidated; the bulk of caching is now done by database_fields.

A lot of code (especially to do with custom fields, or composite fields) have been slimmed down to work on top of db() or database_fields().

The rest of the changes are mostly just fixing regressions from the new behaviour.

inheritedDatabaseFields is deprecated.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -6,9 +6,51 @@
...
+These methods have been updated to include base fields (such as ID, ClassName, Created, and LastEdited), as
+well as composite DB fields.
+
+`DataObject::database_fields` will return the list of all fields on the given table, but 
+will also return the list of CompositeDBField specs by reference via the second parameter.
+
+If user code requires the list of fields excluding base fields, then use custom_database_fields instead, or
+make sure to call `unset($fields['ID']);` if this field should be excluded.
+
+`DataObject:db()` will return all logical fields, including composite DB Fields, alongside
```
sminnee: Does `db()` return foreign keys? If so, we should mention that.

tractorcow: Ah, yep, i'll mention it. ;D

tractorcow: I've fixed this.

tractorcow:
:mergeit: ?

tractorcow:
Rebased since changes have been merged up from 3 branch

tractorcow:
@chillu ignore everything I just said. I've removed the composite fields being passed back by reference.

Seem ok now?

model/DataObject.php
```diff
@@ -1829,13 +1820,18 @@ public function hasOne($component = null) {
 	/**
 	 * Return data for a specific has_one component.
 	 * @param string $component
+	 * @param string $table Out parameter of the table this has_one field belongs to
 	 * @return string|null
 	 */
-	public function hasOneComponent($component) {
-		$hasOnes = (array)Config::inst()->get($this->class, 'has_one', Config::INHERITED);
+	public function hasOneComponent($component, &$table = null) {
```
chillu: `$table` isn't used in here?

model/DataObject.php
```diff
@@ -1916,25 +1912,31 @@ public function db($fieldName = null) {
...
-					return $dbItems[$fieldName];
+			// Merge fields with new fields and composite fields
+			$fields = self::database_fields($class);
+			$compositeFields = self::composite_fields($class, false);
+			$db = array_merge($db, $fields, $compositeFields);
+
+			// Check for search field
+			if($fieldName && isset($db[$fieldName])) {
+				// Set table for the given field
+				if($fieldName === 'ID') {
```
chillu: Shouldn't this be checking for `fixed_fields` rather than just `ID`?

model/DataObject.php
```diff
@@ -1902,13 +1898,13 @@ public function belongsToComponent($component, $classOnly = true) {
...
-	 * Return all of the database fields defined in self::$db and all the parent classes.
-	 * Doesn't include any fields specified by self::$has_one.  Use $this->hasOne() to get these fields
+	 * Return all of the database fields in this object
 	 *
 	 * @param string $fieldName Limit the output to a specific field name
+	 * @param string $table Out parameter of the table this db field is set to
 	 * @return array The database fields
 	 */
-	public function db($fieldName = null) {
+	public function db($fieldName = null, &$table = null) {
```
chillu: I have the same concerns than the `&$compositeFields` comments earlier on - using a pass-by-reference parameter as a second return type waters down the API and makes it more prone to side effects. The one use of this parameter (`dbObject()`) could use an internal cache which is also created through `cache_database_fields()`? Admittedly hacky in its own way. And we really don't want to repeat this loop, given how often `dbObject()` is used throughout the class. Any smart ideas on how to avoid this reference?

model/DataObject.php
tractorcow: Look a few lines down. ;D

tractorcow: I think this change can be reverted though. It doesn't seem to add much value.

model/DataObject.php
tractorcow: It doesn't matter; Fixed fields will already be assigned to the base table without this code. The reason for ID being an exception here is because it is the only field that is on EVERY table, so it should default to the best (base) table.

However, I'll change this code to make it more explicitly map fixed_fields to the base. It won't hurt and hopefully will improve readability. :D

model/DataObject.php
tractorcow: I could change the return type to be `Table.Column(spec)` format instead of `Column(spec)`. :)

chillu: That'll mess up a lot of existing uses of `db()` - it's a fairly common method in deeper model customisations.

tractorcow: Not too hard to use. :dancer: 
```php
$helper = $this->db($fieldName, true);
if($helper) {
	list($table, $spec) = explode('.', $helper);
	
```

tractorcow: Only if you pass in true as a second parameter. It's a new parameter, so it's safe to extend.

Without it the method behaves as it did before.

chillu:
Nice, I'm happy with this PR now. A bit hard to tell if there's any regressions, given the depth of changes here - but we have good unit test coverage on that layer :) Anybody else want to chime in here before we merge? @dhensby @sminnee?

sminnee:
@tractorcow are there tests for the includeTable usage?

tractorcow:
There are now.

tests/model/DataObjectTest.php
```diff
@@ -35,19 +35,29 @@ public function testDb() {
...
+		$this->assertArrayHasKey('ClassName', $dbFields);
+		$this->assertArrayHasKey('ID', $dbFields);
 
 		// Assert that the correct field type is returned when passing a field
 		$this->assertEquals('Varchar', $obj->db('Name'));
 		$this->assertEquals('Text', $obj->db('Comment'));
 
+		// Test with table required
+		$this->assertEquals('DataObjectTest_TeamComment.Varchar', $obj->db('Name', true));
+		$this->assertEquals('DataObjectTest_TeamComment.Text', $obj->db('Comment', true));
```
tractorcow: Example of new usage.

sminnee:
I'm not overly excited about using a string return type that is later extracted with an `explode()` call, as as a 'way of avoiding a mixed return type' it seems like we're kidding ourselves: we *have* a mixed return type we've just obfuscated that reality.

However, I think that we've gone around in circles on this one enough and so I'd be inclined to merge the PR regardless of that, and to pick up the mixed-return-type issue as part of a broader review of the DataObject meta-data methods, as I've mentioned elsewhere.

I guess I should do an RFC for that? ;-)

sminnee:
@chillu i'll let you take one last look and then can you press the button?

chillu:
Mixed return types are the lesser of two evils for me here. Its more clear on what happens than with a random pass-by-reference parameter at least (which could be read, overwritten, or amended, based on the implementation). Agree that we need more expressive ways to pass around this metadata in the ORM.

tractorcow:
Yay!

PR #3993, opened 2015-03-10
BUG Fix yaml generation to conform to version 1.1, accepted by transifex

tractorcow:
Fixes #3992

PR #4591, opened 2015-09-08
Load YAML component via composer, remove Spyc

kinglozzer:
NB: doesn’t actually remove the old `sfYaml` component. We were using the one bundled in with `zend_translate_railsyaml`: I didn’t want to modify a thirdparty lib (even if it is built by one of our own :wink:). I’m happy to update `zend_translate_railsyaml` and remove the `sfYaml` library if it’s preferred.

In `i18nTextCollector` we were temporarily switching spec versions back from 1.2 to 1.1 to fix issues with truthy/falsy strings (see #3993), but the tests added in that PR now pass so it looks like that’s no longer an issue. @tractorcow, what do you think? :)

Had to update some of the YAML files:
- The Oembed stuff had trailing commas, which according to [the spec](http://www.yaml.org/spec/1.2/spec.html) are only valid if used inside square brackets, so I got rid of them.
- Some of the fixture files were tab-indented. All of the YAML specs (1.0 through 1.2) explicitly state that tabs are not allowed, so they’ve been converted to spaces.

Added a `4.0.0.md` changelog too, because why the heck not?

:pray: screen s :pray:


kinglozzer:
Also, merging this will break CMS builds until [this line](https://github.com/silverstripe/silverstripe-cms/blob/7d7c75414c338bfd906f7e63eccaccff913951ef/tests/model/SiteTreeBrokenLinksTest.yml#L23) has the tab removed. Doing a PR for only that seems a bit OTT, so if this is going to be merged can whoever clicks the green button just edit that file in the web view? The same issue may potentially occur for other modules, so I’ll run through our core modules soon and check for invalid YAML.

tractorcow:
Symfony doesn't support multiple yml documents in a file, which is necessary for some modules to work.

https://github.com/symfony/Yaml/blob/master/Parser.php#L239

tractorcow:
Also, we were using yml spec 1.1 because that's the format required by transifex for i18nTextCollector. I guess if the tests are ok, then the new lib is fine.

I'd still like to test running this code through the translation process just to make sure that it doesn't break transifex, before merging.

micmania1:
Nice, I was looking at doing this last week.

@tractorcow the current yaml parser doesn't support multiple docs either. There's a workaround here: https://github.com/kinglozzer/sapphire/blob/symfony-yaml/core/manifest/ConfigManifest.php#L267

tractorcow:
Oh, that makes a bit more sense. I guess it's not so much an issue to swap in symfony in that case, is there?

kinglozzer:
Yeah that was the first thing I looked for, but still no multi doc support in any lib I could find :(. Spyc is now on Github + Packagist, but is less popular and less frequently updated, so I thought it best to opt for Symfony.

>I'd still like to test running this code through the translation process just to make sure that it doesn't break transifex, before merging.

Yep, good idea. Is that something I can do/help with?

> I guess it's not so much an issue to swap in symfony in that case, is there?

It’s low effort, low impact really. sfYaml _is_ Symfony, just a super-old version of it - they have a near-identical API though. Swapping out Spyc for Symfony is a tiny bit more work because Symfony won’t open a file for you (hence the changes [here](https://github.com/silverstripe/silverstripe-framework/pull/4591/files#diff-cb185d5bd15c06ece01ce250a6e4582cL127)).

---

Regarding `zend_translate_railsyaml`, I’m getting a bit off-topic here but I think we could tidy it up:

- Composer it
- Put it on Packagist with new Symfony YAML as a dependency instead of bundling it in there
- Pull it in by adding it to our composer.json.

Thoughts @chillu? I’m happy to do the legwork on it where I can. Even further off-topic, I had a [very brief go](https://github.com/kinglozzer/sapphire/blob/c83d1347ffa7e9d7db4f75e52293d9cec97eb264/i18n/RailsYamlLoader.php) at a ZF2-compatible version of it when tinkering with upgrading to ZF2. Anyway, this can all wait until later, I’m just thinking out loud!

kinglozzer:
Looks like I need to rebase this.. any further comments before I do that? :)

tractorcow:
As long as transifex doesn't break on unquoted boolean values (see https://github.com/silverstripe/silverstripe-framework/blob/3/lang/en.yml#L64) I'm in favour of this.

Can you please run the task and let me know if there are any errors? Otherwise, good to merge.

kinglozzer:
@tractorcow When I run `i18nTextCollector` I see a lot of entities removed (though I see the same ones removed regardless of whether I run on master or on my branch), but the boolean values still come out as quoted `'Yes'`/`'No'` strings:

```yml
Boolean:
  ANY: Any
  NOANSWER: 'No'
  YESANSWER: 'Yes'
```

tractorcow:
@kinglozzer yeah that's normal behaviour. We'll change `merge` to true by default to fix that.

tractorcow:
Thanks @kinglozzer and great work. :)

PR #4593, opened 2015-09-09
Attempting to call Fields() on instance of SS_HTTPResponse

### LINKED ISSUE #4592: Attempting to call Fields() on instance of SS_HTTPResponse
On a subclass of `AssetAdmin` in its `getEditForm()` method, I can call `parent::getEditForm($id, $fields);` and find it's possible for the framework to return an instance of `SS_HTTPResponse` rather than the expected `Form` if a `canView()` check up the chain in `LeftAndMain` fails.

Looking in `LeftAndMain::getEditForm()` checks are failing on line 1174 and I'm being returned an instance of `SS_HTTPResponse` rather than `Form`. This means I get the error:

`Fatal error: Call to undefined method SS_HTTPResponse::Fields() in /var/www/htdocs/my-project/cms/code/controllers/AssetAdmin.php on line 158`

phptek:
- "Moves" the error down to the view, from a header

https://github.com/silverstripe/silverstripe-framework/issues/4592

### DIFF
--- a/admin/code/LeftAndMain.php
+++ b/admin/code/LeftAndMain.php
@@ -1171,10 +1171,9 @@ class LeftAndMain extends Controller implements PermissionProvider {
 			$record = $id;
 		} else {
 			$record = $this->getRecord($id);
-			if($record && !$record->canView()) return Security::permissionFailure($this);
 		}
 
-		if($record) {
+		if($record && $record->canView()) {
 			$fields = ($fields) ? $fields : $record->getCMSFields();
 			if ($fields == null) {
 				user_error(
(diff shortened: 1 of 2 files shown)


sminnee:
This seems good, but will need a rebase and test. Can we put together a test-case for this feature?

dhensby:
Bump?

maxime-rainville:

maxime-rainville:
I've double checked to see if it could be rebased into SilverStripe 4. LeftAndMain and AssetAdmin have changed too much since the issue was raised.

PR #4595, opened 2015-09-10
FIX: Editing existing file links in HtmlEditorField was broken

kinglozzer:
The old `el = $('#' + fieldName);` doesn’t work as the field IDs have changed. This uses the [input name](https://github.com/silverstripe/silverstripe-framework/blob/06d7bb77c08a54032f1d49f4c99d50f271ff9609/templates/UploadField.ss#L53) instead, which is `file[Uploads][]`, to get to the field.

### DIFF
--- a/javascript/HtmlEditorField.js
+++ b/javascript/HtmlEditorField.js
@@ -823,9 +823,11 @@ ss.editorWrappers['default'] = ss.editorWrappers.tinyMCE;
 						} else if(el.is(':radio')) {
 							el.val([selected]).change();
 						} else if(fieldName == 'file') {
-							// Can't rely on fieldName, ad UploadFields have different naming convention
-							el = $('#' + fieldName);
-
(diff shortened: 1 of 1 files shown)


willmorgan:
Hmm, the `fieldName` also includes the SS form's name too, right? Otherwise we might have a problem with name clashes.

kinglozzer:
@willmorgan `fieldName` is just `file`, but using `this.find()` means it is [scoped to the current form](https://github.com/kinglozzer/sapphire/blob/pulls/uploadfield-file-link-fix/javascript/HtmlEditorField.js#L555), so unless someone adds an input with an identical name to the link form specifically then that shouldn’t be an issue. It’s [what we already use](https://github.com/kinglozzer/sapphire/blob/pulls/uploadfield-file-link-fix/javascript/HtmlEditorField.js#L816) for the other fields too :)

kinglozzer:
<3

PR #4599, opened 2015-09-11
API Implementation of RFC-1 Asset Abstraction

tractorcow:
Partial implementation of https://github.com/silverstripe/silverstripe-framework/issues/3792

This PR includes creation of the APL (Asset Persistence Layer) using flysystem as a default backend, and creation of the DBFile field.

This is built on top of https://github.com/silverstripe/silverstripe-framework/pull/4590 which should be merged first.

### DIFF
--- a/docs/en/02_Developer_Guides/14_Files/01_File_Management.md
+++ b/docs/en/02_Developer_Guides/14_Files/01_File_Management.md
@@ -2,9 +2,125 @@ summary: Learn how to work with File and Image records
 
 # File Management
 
-## Files, Images and Folders as database records
+## Storage via database columns
 
-All files, images and folders in the 'assets' directory are stored in the database. Each record has the following database fields:
+Asset storage is provided out of the box via a [Flysystem](http://flysystem.thephpleague.com/) backed store.
+However, any class that implements the `AssetStore` interface could be substituted to provide storage backends
+via other mechanisms.
+
+Internally, files are stored as `[api:DBFile]` records on the rows of parent objects. These records are composite fields
+which contain sufficient information useful to the configured asset backend in order to store, manage, and 
+publish files. By default this composite field behind this field stores the following details:
+
+
+| Field name     | Description |
+| ----------     | -----------   
+| `Hash`         | The sha1 of the file content, useful for versioning (if supported by the backend) |
+| `Filename`     | The internal identifier for this file, which may contain a directory path (not including assets). Multiple versions of the same file will have the same filename. |
+| `Variant`      | The variant for this file. If a file has multiple derived versions (such as resized files or reformatted documents) then you can point to one of the variants here. |
+
+
+Note that the `Hash` and `Filename` always point to the original file, if a `Variant` is specified. It is up to the
+storage backend to determine how variants are managed.
+
+Note that the storage backend used will not be automatically synchronised with the database. Only files which
+are loaded into the backend through the asset API will be available for use within a site.
+
+## Compatibility with 3.x filename paths
+
+If upgrading to 4.0 from earlier versions when using the default asset store, it's important to consider
+how existing files will be migrated.
+
+Because the filesystem now uses the sha1 of file contents in order to version multiple versions under the same
+filename, the default storage paths in 4.0 will not be the same as in 3.
+
+In order to retain existing file paths in line with framework version 3 you should set the
+`\SilverStripe\Filesystem\Flysystem\FlysystemAssetStore.legacy_paths` config to true.
+Note that this will not allow you to utilise certain file versioning features in 4.0.
+
+	:::yaml
+	\SilverStripe\Filesystem\Flysystem\FlysystemAssetStore:
+		legacy_paths: true
+
+## Loading content into `DBFile`
+
+A file can be written to the backend from a file which exists on the local filesystem (but not necessarily
+within the assets folder).
+
+For example, to load a temporary file into a DataObject you could use the below:
+
+
+	:::php
+	<?
+	class Banner extends DataObject {
+		private static $db = array(
+			'Image' => 'DBFile'
+		);
+	}
+
+	// Image could be assigned in other parts of the code using the below
+	$banner = new Banner();
+	$banner->Image->setFromLocalFile($tempfile['path'], 'uploads/banner-file.jpg');
+
+
+When uploading a file it's normally necessary to give the file a useful name and directory, otherwise the
+asset storage backend will choose one for you.
+
+Alternatively, content can be loaded using one of the below methods:
+
+
+| Method                     | Description                             |
+| -------------------------- | --------------------------------------- |
+| `DBFile::setFromLocalFile` | Load a local file into the asset store  |
+| `DBFile::setFromStream`    | Will store content from a stream        |
+| `DBFile::setFromString`    | Will store content from a binary string |
+
+
+The result of these methods will be an array of data (tuple) which contains the values of the above fields
+(Filename, Hash, and Variant).
+
+## Conflict resolution
+
+When storing files, it's possible to determine the mechanism the backend should use when it encounters
+an existing file pattern. The conflict resolution to use can be passed into the third parameter of the
+above methods (after content and filename). The available constants are:
+
+| Constant                            | If an existing file is found then:  |
+| ----------------------------------- | ----------------------------------- |
+| `AssetStore::CONFLICT_EXCEPTION`    | An exception will be thrown         |
+| `AssetStore::CONFLICT_OVERWRITE`    | The existing file will be replaced  |
+| `AssetStore::CONFLICT_RENAME`       | The backend will choose a new name. |
+| `AssetStore::CONFLICT_USE_EXISTING` | The existing file will be used      |
+
+
+If no conflict resolution scheme is chosen, or an unsupported one is requested, then the backend will choose one.
+The default asset store supports each of these.
+
+## Getting content from a `DBFile`
+
+When placed into a template (e.g. `$MyFileField`) then `[api:DBFile]` will automatically generate the appropriate
+template for the file mime type. For images, this will embed an image tag. For documents a download link will be presented.
+
+As with storage, there are also different ways of loading the content (or properties) of the file:
+
+| Method                   | Description                                                |
+| ------------------------ | ---------------------------------------------------------- |
+| `DBFile::getStream`      | Will get an output stream of the file content              |
+| `DBFile::getString`      | Gets the binary content                                    |
+| `DBFile::getURL`         | Gets the url for this resource. May or may not be absolute |
+| `DBFile::getAbsoluteURL` | Gets the absolute URL to this resource                     |
+| `DBFile::getMimeType`    | Get the mime type of this file                             |
+| `DBFile::getMetaData`    | Gets other metadata from the file as an array              |
+
+
+As with other db field types, `DBField` can also be subclassed or extended to add additional properties
+(such as SEO related fields).
+
+## Storage via `File` DataObject
+
+Other than files stored exclusively via DBFile, files can also exist as subclasses of the `File` DataObject.
+
+Each record has the following database fields:
 
 | Field name     | Description                                                                                                                                          |
 | ----------     | -----------                                                                                                                                          |
@@ -16,6 +132,7 @@ All files, images and folders in the 'assets' directory are stored in the databa
 | `ShowInSearch` | Whether the file should be shown in search results, defaults to '1'. See ["Tutorial 4 - Site Search"](/tutorials/site_search) for enabling search. |
 | `ParentID`     | The ID of the parent Folder that this File/Folder is in. A ParentID of '0' indicates that the File/Folder is in the 'assets' directory.              |
 | `OwnerID`      | The ID of the Member that 'owns' the File/Folder (not related to filesystem permissions).                                                            |
+
 
 ## Management through the "Files" section of the CMS
 
--- a/docs/en/04_Changelogs/4.0.0.md
+++ b/docs/en/04_Changelogs/4.0.0.md
@@ -12,6 +12,27 @@
 
 ## Upgrading
 
+### New asset storage mechanism
+
+File system has been abstracted into an abstract interface. By default, the out of the box filesystem
+uses [Flysystem](http://flysystem.thephpleague.com/) with a local storage mechanism (under the assets directory).
+
+Because the filesystem now uses the sha1 of file contents in order to version multiple versions under the same
+filename, the default storage paths in 4.0 will not be the same as in 3.
+
+In order to retain existing file paths in line with framework version 3 you should set the
+`\SilverStripe\Filesystem\Flysystem\FlysystemAssetStore.legacy_paths` config to true.
+Note that this will not allow you to utilise certain file versioning features in 4.0.
+
+
+	:::yaml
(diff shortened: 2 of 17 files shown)


sminnee:
OK I've posted a bunch of comments, but this is looking pretty solid! I reckon we post a link to this PR to ss-dev so that we can broaden its visibility.

tractorcow:
I had reader and writer separate initially, but @hafriedlander didn't like it so I removed it. :D

filesystem/flysystem/FlysystemAssetStore.php
```diff
@@ -0,0 +1,269 @@
...
+	 */
+	private static $legacy_filenames = false;
+
+	/**
+	 * Assign new flysystem backend
+	 *
+	 * @param Filesystem $filesystem
+	 */
+	public function setFilesystem(Filesystem $filesystem) {
+		$this->filesystem = $filesystem;
```
dhensby: `return $this`?

filesystem/flysystem/FlysystemAssetStore.php
```diff
@@ -0,0 +1,269 @@
...
+
+/**
+ * Asset store based on flysystem Filesystem as a backend
+ */
+class FlysystemAssetStore implements AssetStore {
+
+	/**
+	 * @var Filesystem
+	 */
+	protected $filesystem;
```
dhensby: If we have rigid getters and setters, this should be `private`

tractorcow: I thought you were against private. :) https://github.com/silverstripe/silverstripe-framework/pull/3874

filesystem/flysystem/AssetAdaptor.php
```diff
@@ -0,0 +1,36 @@
...
+namespace SilverStripe\Filesystem\Flysystem;
+
+use Controller;
+use Director;
+use League\Flysystem\Adapter\Local;
+
+/**
+ * Adaptor for local filesystem based on assets directory
+ */
+class AssetAdaptor extends Local {
```
kinglozzer: Adapt*o*r or adapt*e*r? I don’t even know which is correct, but I’d prefer if we were consistent with Flysystem’s naming (`use League\Flysystem\Adapter\Local`)

filesystem/flysystem/FlysystemUrlPlugin.php
```diff
@@ -0,0 +1,70 @@
...
+     * @return string The full url to the file
+     */
+    public function handle($path) {
+		// Default adaptor
+		if($this->adapter instanceof AssetAdaptor) {
+			return $this->adapter->getPublicUrl($path);
+		}
+
+		// Check S3 adaptor
+        if (class_exists('League\Flysystem\AwsS3v2\AwsS3Adapter')
```
kinglozzer: Is there a good reason for hard-coding a check for an S3 adapter here? Shouldn’t it be registered as a separate “plugin” or something? Seems weird to hard-code the `https://s3.amazonaws.com/%s/%s` like this

docs/en/02_Developer_Guides/14_Files/01_File_Management.md
```diff
@@ -2,9 +2,122 @@ summary: Learn how to work with File and Image records
...
+	<?
+	class Banner extends DataObject {
+		private static $db = array(
+			'Image' => 'DBFile'
+		);
+	}
+
+	// Image could be assigned in other parts of the code using the below
+	$banner = new Banner();
+	$banner->Image->setFromLocalFile($tempfile['path'], 'uploads/banner-file.jpg');
```
kinglozzer: Does this need to be `$banner->dbObject('Image')`?

filesystem/flysystem/AssetAdaptor.php
tractorcow: Ah, so there are two spellings. :) I'll make sure to use the consistent one.

filesystem/flysystem/FlysystemUrlPlugin.php
tractorcow: This is simply imported from a third party library which I extended for our own purposes.
https://github.com/SmartestEdu/FlysystemPublicUrlPlugin

Unfortunately, flysystem limits extensions (plugins) to be added only to the filesystem object, so there's no way of actually adding the getPublicURL Extension itself to the backend adapter.

I'm really waiting for the flysystem devs to give us an official API for getting public urls; At the moment it's just not possible without lots of hard coding (as in this case).

docs/en/02_Developer_Guides/14_Files/01_File_Management.md
tractorcow: It doesn't need to be anymore. :) Composite DBFields can be accesed via normal getters now.

kinglozzer: Awesome!

filesystem/flysystem/FlysystemUrlPlugin.php
kinglozzer: I see :(

filesystem/flysystem/FlysystemAssetStore.php
dhensby: I'm for consistency! (Ironic, I know).

But if there are no consistent getters/setters then we can't use private because ancestors are unable to access key properties. however, if the properties are accessible via methods, this is fine.

filesystem/flysystem/FlysystemAssetStore.php
```diff
@@ -0,0 +1,271 @@
...
+		if($content === false) {
+			throw new InvalidArgumentException("$path does not exist");
+		}
+
+		// Write as string
+		return $this->setFromString($content, $filename, $conflictResolution);
+	}
+
+	public function setFromStream($stream, $filename, $conflictResolution = null) {
+		$content = stream_get_contents($stream);
```
chillu: Why isn't this using `writeStream()` in Flysystem? I assume that'd be a lot more memory efficient, right? See http://flysystem.thephpleague.com/api/

filesystem/flysystem/FlysystemAssetStore.php
```diff
@@ -0,0 +1,271 @@
...
+		}
+
+		return array(
+			'Hash' => $hash,
+			'Filename' => $filename,
+			'Variant' => ''
+		);
+	}
+
+	public function getMetadata($hash, $filename, $variant = null) {
```
chillu: Should document the return format (I assume it's an associative array?)

filesystem/flysystem/FlysystemAssetStore.php
sminnee: Writing to a temporary file and using http://php.net/manual/en/function.sha1-file.php could be helpful for writing large streams. I don't think it blocks merge, though.

sminnee: FYI: I just did a quick test: running `sha1_file()` on a 2.1GB file executed without incident in 5.8s and `memory_get_peak_usage()` return 262kb.

filesystem/flysystem/FlysystemAssetStore.php
tractorcow: It's documented in the interface it implements.

filesystem/flysystem/FlysystemAssetStore.php
tractorcow: A lot of testing later, I also found I can use the following to sha1 an arbitrary stream.

```php
protected function getStreamSHA1($stream) {
	$context = hash_init('sha1');
	hash_update_stream($context, $stream);
	return hash_final($context);
}
```

tractorcow:
I've updated FlysystemAssetStore to never unwrap files or streams into raw strings in memory, and using stream / file based sha1 generation if possible.

Where streams are not seekable, they are written directly into temporary files and new streams generated from these.

tractorcow:
Rebased from master, as the composite DBField work has been merged.

filesystem/flysystem/FlysystemAssetStore.php
```diff
@@ -0,0 +1,368 @@
...
+		if(($pos = strpos($name, '.')) !== false) {
+			$extension = substr($name, $pos);
+			$name = substr($name, 0, $pos);
+		}
+
+		// Unless in legacy mode, inject hash just prior to the filename
+		if(Config::inst()->get(__CLASS__, 'legacy_filenames')) {
+			$fileID = $name;
+		} else {
+			$fileID = substr($hash, 0, 10) . '/' . $name;
```
chillu: This could theoretically lead to edge cases where a version of a file overwrites another one, because the shortened content SHA1 isn't sufficiently unique, right? `123decafbad9999999999999999` and `123decafbad111111111111111` look like the same hash when shortened to the first 10 chars. It's veeery unlikely of course. Just wondering if we should avoid it architecturally by leaving the original hash intact.

model/fieldtypes/DBFile.php
```diff
@@ -0,0 +1,203 @@
...
+	 * Return an XHTML img tag for this file
+	 *
+	 * @return string
+	 */
+	public function forTemplate() {
+		return $this->getTag() ?: '';
+	}
+
+	/**
+	 * Return an XHTML img tag for this file
```
chillu: It isn't always an "img tag", right? `DBFile_download.ss` returns `<a>`

docs/en/02_Developer_Guides/14_Files/01_File_Management.md
```diff
@@ -2,9 +2,122 @@ summary: Learn how to work with File and Image records
...
+
+
+| Field name     | Description |
+| ----------     | -----------   
+| `Hash`         | The sha1 of the file content, useful for versioning (if supported by the backend) |
+| `Filename`     | The internal identifier for this file, which may contain a directory path (not including assets). Multiple versions of the same file will have the same filename. |
+| `Variant`      | The variant for this file. If a file has multiple derived versions (such as resized files or reformatted documents) then you can point to one of the variants here. |
+
+
+Note that the `Hash` and `Filename` always point to the original file, if a `Variant` is specified. It is up to the
```
chillu: This should probably mention that files need to be managed exclusively via `AssetStore` - new files uploaded directly to e.g. S3 won't be picked up by it, and renamed or changed files risk losing their association with metadata stored through `DBFile`.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -12,6 +12,27 @@
 
 ## Upgrading
 
+### New asset storage mechanism
+
+File system has been abstracted into an abstract interface. By default, the out of the box filesystem
+uses [Flysystem](http://flysystem.thephpleague.com/) with a locale storage mechanism (under the assets directory).
```
chillu: You mean "local" rather than "locale" here, right?

filesystem/flysystem/FlysystemAssetStore.php
tractorcow: It's a 1 in 1,000,000,000,000 chance for these to conflict.

If they do, that's what the conflictResolution is for. :)

model/fieldtypes/DBFile.php
tractorcow: yep I should update this.

docs/en/02_Developer_Guides/14_Files/01_File_Management.md
tractorcow: Ok. :D

docs/en/04_Changelogs/4.0.0.md
tractorcow: Ah yes. Too much fluent lol.

filesystem/flysystem/FlysystemAssetStore.php
tractorcow: I think the improvement in the url length is probably justified in this case. Long urls with arbitrary characters obscure the value of the filename.

chillu: Fair enough :)

tractorcow:
Fixed and squashed.

sminnee:
Given the scope of this change, it would be good to get 3 :+1:s from core committers before merging.

:+1: from me. Can we get 2 more?

chillu:
:+1:

tractorcow:
:+1:

tractorcow:
@dhensby do you want to have a look at this work? @jonom?

docs/en/02_Developer_Guides/14_Files/01_File_Management.md
```diff
@@ -2,9 +2,125 @@ summary: Learn how to work with File and Image records
...
+`\SilverStripe\Filesystem\Flysystem\FlysystemAssetStore.legacy_paths` config to true.
+Note that this will not allow you to utilise certain file versioning features in 4.0.
+
+	:::yaml
+	\SilverStripe\Filesystem\Flysystem\FlysystemAssetStore:
+		legacy_paths: true
+
+## Loading content into `DBFile`
+
+A file can be written to the backend from a while which exists on the local filesystem (but not necessarily
```
wilr: "A file can be written to the backend from a while which". Not correct?

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -12,6 +12,27 @@
...
+### New asset storage mechanism
+
+File system has been abstracted into an abstract interface. By default, the out of the box filesystem
+uses [Flysystem](http://flysystem.thephpleague.com/) with a local storage mechanism (under the assets directory).
+
+Because the filesystem now uses the sha1 of file contents in order to version multiple versions under the same
+filename, the default storage paths in 4.0 will not be the same as in 3.
+
+In order to retain existing file paths in line with framework version 3 you should set the
+`\SilverStripe\Filesystem\Flysystem\FlysystemAssetStore.legacy_paths` config to true.
```
wilr: Is this going to cause an issue with images embedded into TinyMCE? If so, migration tool?

filesystem/flysystem/AssetAdapter.php
```diff
@@ -0,0 +1,36 @@
...
+
+namespace SilverStripe\Filesystem\Flysystem;
+
+use Controller;
+use Director;
+use League\Flysystem\Adapter\Local;
+
+/**
+ * Adaptor for local filesystem based on assets directory
+ */
```
wilr: Missing apidoc headers

docs/en/02_Developer_Guides/14_Files/01_File_Management.md
tractorcow: Should have said "file"

docs/en/04_Changelogs/4.0.0.md
tractorcow: You can't embed these into tinymce for now.

filesystem/flysystem/AssetAdapter.php
tractorcow: I can probably add some in.

docs/en/04_Changelogs/4.0.0.md
tractorcow: When 4.0 is released embedded files should work as they did before, although the new DBFile will get its own shortcode handler.

tractorcow:
@wilr your feedback has been resolved. ;)

wilr:
Gave it a spin locally. Nice, then a :+1: from me then.

sminnee:
And we're in!

PR #4602, opened 2015-09-14
FIX: #103 in silverstripe-installer

SpiritLevel:
See also #58 in silverstripe-installer.


SpiritLevel:
@tractorcow  :(

Just tried another install without _ss_environment.php and "DirectoryIndex disabled" is still being deleted.

PR #4607, opened 2015-09-16
Changed extractTags from protected to public

nickspiel:
It would be super handy if the extractTags function was public.
This class is a real hassle to extend.


tractorcow:
Ta :D

PR #4608, opened 2015-09-17
BUG Fix issue with GridFieldSortableHeader incorrectly walking database schema

tractorcow:
This seems to be an issue exposed by, but not caused by, the recent ClassInfo improvements.

The incorrect behaviour is `ClassInfo::table_for_object_field(` being passed in aliases rather than actual classnames, meaning that the fallback would occur more often than it should, and would not correctly include the table.

The fix to this issue is to distinguish between tables joined via aliases (as per the SortColumn parameter) and those tables joined implicitly (through db schema reflection).

Variables named `$tmpItem` did not help solve this issue..........

### DIFF
--- a/forms/gridfield/GridFieldSortableHeader.php
+++ b/forms/gridfield/GridFieldSortableHeader.php
@@ -213,29 +213,43 @@ class GridFieldSortableHeader implements GridField_HTMLProvider, GridField_DataM
 		// if we have a relation column with dot notation
 		if(strpos($column, '.') !== false) {
 			$lastAlias = $dataList->dataClass();
-			$tmpItem = singleton($lastAlias);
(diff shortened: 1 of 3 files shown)


sminnee:
Also, can we stop messing around with 3.1? @dhensby's original change to ClassInfo never should have gone in there, and we've had this spiral of subsequent changes being sucked into it.

sminnee:
OK—I think that this change should be shifted to the `3` branch.

tractorcow:
Agreed on 3.1. Let's just patch ClassInfo there.

For 3 I'll see about refactoring this into DataList.

tractorcow:
I've moved my fix for this to https://github.com/silverstripe/silverstripe-framework/pull/4611 in the 3 branch.

composite field sorting has been implemented in DataList

PR #4609, opened 2015-09-17
BUG Fix ClassInfo::table_for_object_field

tractorcow:
Substitute fix for https://github.com/silverstripe/silverstripe-framework/pull/4608

Apparently on < 5.3.7 this silent issue is a non-silent one, so class_exists must be called before checking is_subclass_of.

The actual issue in gridfieldsortableheader still exists, but we've decided to focus efforts on fixing this in the 3.x branch.


tractorcow:
@sminnee is this fix ok with you?

sminnee:
it is, but more importantly, it's okay with Hamish :-)

PR #4611, opened 2015-09-17
API Enable DataList::sort to support composite field names

tractorcow:
So you can do syntax like `Company::get()->sort('Owner.Name');`

The re-implementation of this code in GridFieldSortableHeader has been dumped out.

### DIFF
--- a/tests/model/DataListTest.php
+++ b/tests/model/DataListTest.php
@@ -530,6 +530,17 @@ class DataListTest extends SapphireTest {
 		$this->assertEquals('Phil', $list->last()->Name);
 	}
 
+	public function testSortInvalidParameters() {
+		$this->setExpectedException(
+			'InvalidArgumentException',
+			'Fans is not a linear relation on model DataObjectTest_Player'
+		);
+		$list = DataObjectTest_Team::get();
+		$list = $list->sort('Founder.Fans.Surname'); // Can't sort on has_many
+	}
+
+
+
 	/**
 	 * $list->filter('Name', 'bob'); // only bob in the list
 	 */
--- a/tests/model/DataListTest.php
+++ b/tests/model/DataListTest.php
@@ -521,6 +521,15 @@ class DataListTest extends SapphireTest {
 		$this->assertEquals('Phil', $list->last()->Name, 'Last comment should be from Phil');
 	}
 
+	public function testSortWithCompositeSyntax() {
+		// Phil commented on team with founder surname "Aaron"
+		$list = DataObjectTest_TeamComment::get();
+		$list = $list->sort('Team.Founder.Surname', 'asc');
+		$this->assertEquals('Phil', $list->first()->Name);
+		$list = $list->sort('Team.Founder.Surname', 'desc');
+		$this->assertEquals('Phil', $list->last()->Name);
+	}
+
 	/**
 	 * $list->filter('Name', 'bob'); // only bob in the list
 	 */
--- a/forms/gridfield/GridFieldSortableHeader.php
+++ b/forms/gridfield/GridFieldSortableHeader.php
@@ -208,53 +208,6 @@ class GridFieldSortableHeader implements GridField_HTMLProvider, GridField_DataM
 			return $dataList;
 		}
 
-		$column = $state->SortColumn;
-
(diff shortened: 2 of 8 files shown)


dev/CsvBulkLoader.php
```diff
@@ -60,6 +60,7 @@ public function preview($filepath) {
 	 * @return null|BulkLoader_Result
 	 */
 	protected function processAll($filepath, $preview = false) {
+		$filepath = Director::getAbsFile($filepath);
```
tractorcow: This failed when testing locally.

kinglozzer:
:heart: this. I should’ve taken this approach originally instead of writing all that hideous code in `GridFieldSortableHeader` (sorry!).

What happens if you attempt to sort on a has_many/many_many? e.g. `Category::get()->sort('Products.Title')`

dhensby:
If it works anything like `filter` it'll "just work"

kinglozzer:
Checked it out locally and it does seem to “just work”. You do end up with duplicate results, though. E.g. if Default Admin is in two groups:

```php
$members = Member::get()->sort('Groups.Title');
foreach($members as $member) {
	echo $member->getName() . '<br />';
}

// Default Admin
// Default Admin
```

The SQL seems to make sense (see below). You can use `->alterDataQuery()` / `$dq->groupby('ID')` to remove the duplicates. I’m not sure if this is something we should handle for the user, though? Should duplicate rows _ever_ be returned by `DataList`?

```sql
SELECT DISTINCT
	"Member"."ClassName",
	...,
	"Group"."Title" AS "_SortColumn0"
FROM "Member"
INNER JOIN "Group_Members"
	ON "Group_Members"."MemberID" = "Member"."ID"
LEFT JOIN "Group"
	ON "Group_Members"."GroupID" = "Group"."ID"
ORDER BY "_SortColumn0" ASC
```

sminnee:
If DataList starts returning duplicates I think we're going to get in trouble unless we give it a lot more thought. The current query construction rules have been designed to only return distinct records, so I think this `->sort('Group.Title')` behaviour is a bug.

tractorcow:
Hm, we probably should ONLY allow relations to sort on has_one, in that case. If a team has several members, sorting on Members.Name isn't deterministic anyway is it?

sminnee:
It's an easier place to start, I guess. Otherwise I guess we could sort on `MIN(Members.Name)` or something, but our choice of algorithm gets a bit arbitrary. It's difficult to see how it would be of use to developers.

tractorcow:
Ok, I have rebased and added a brand new API:

DataList and DataQuery now both have applyLinearRelation, which will work the same way as applyRelation, but will error on relations which would break the behaviour of sort in non-deterministic ways.

tractorcow:
I just realised that I'd added DataList applyRelation in this pull request, so I have folded the two similar methods into one for simplicity :D

tests/model/DataListTest.php
```diff
@@ -521,6 +521,26 @@ public function testSortWithMultipleArraySyntaxSort() {
...
+		$list = $list->sort('Team.Founder.Surname', 'asc');
+		$this->assertEquals('Phil', $list->first()->Name);
+		$list = $list->sort('Team.Founder.Surname', 'desc');
+		$this->assertEquals('Phil', $list->last()->Name);
+	}
+
+	public function testSortInvalidParameters() {
+		$this->setExpectedException(
+			'InvalidArgumentException',
+			'Fans is not a linear relation on model DataObjectTest_Player'
```
tractorcow: @sminnee this is the critical part from your feedback; We simply won't allow sorting to affect the possible number of results returned, in case we are joining on many_many / has_many.

tractorcow:
I had a chat with @sminnee and we decided to ship this to the master branch instead of 3.x, that way we can make breaking changes to the API and hopefully consolidate a few of these methods.

dhensby:
Sounds good - when will we get the new PR?

tractorcow:
@dhensby here it is https://github.com/silverstripe/silverstripe-framework/pull/4812

tractorcow:
Merged with #4812 into master. ;) thanks @dhensby

PR #4615, opened 2015-09-18
Minor: Prevent Upload from showing default controller template (fixes #4613)

### LINKED ISSUE #4613: Upload class/url always available and always shows an error
The /Upload url is available for everyone while it extends Controller.php and index is in allowed actions.
Also Upload.php doesn't have an index function, this means it always shows the default Silverstripe Controller.ss.

JayDevlin:
Copied from Security->index()

#4613


sminnee:
Seems fine.

PR #4618, opened 2015-09-20
Members with no ID have no permissions

dhensby:
This fixes an issue where empty objects (or indeed IDs of 0) being passed in to a permission::check fall back to the permission set of the currently logged in user.

### DIFF
--- a/security/Permission.php
+++ b/security/Permission.php
@@ -160,7 +160,11 @@ class Permission extends DataObject implements TemplateGlobalProvider {
 		if(!$member) {
 			$memberID = $member = Member::currentUserID();
 		} else {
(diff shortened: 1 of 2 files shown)


tractorcow:
I think we should put this into 3.x-dev, given that it could have regressions in 3.1.

Once 3.2 is out we don't want to resolve any risky fixes in 3.1, and even 3.2 should not be receiving any behaviour-altering changes.

sminnee:
Do we have instances of this occurring? If this is a security issue we should patch 3.1.

tractorcow:
Ok, I can merge this in if you add a test please. :)

dhensby:
ping @tractorcow

tractorcow:
Ok, fair enough. :)

PR #4619, opened 2015-09-21
Allow to handle nested params in search context

ntd:
I implemented a date range field as a single FieldGroup in this way:

```
class DateRangeField extends FieldGroup {

    public function __construct($name) {
        parent::__construct(
            DateField::create($name . '[min]', 'Min'),
            DateField::create($name . '[max]', 'Max')
        );
        parent::setName($name);
    }

    public function setName($name) {
        $this->children[0]->setName($name . '[min]');
        $this->children[1]->setName($name . '[max]');
        return parent::setName($name);
    }
}
```

Then when using it inside a search context I got a fatal error in the trim() call because trim cannot be applied to an array.

The provided patch solves the issue by making the trim call recursive and it should not have any drawback.

### DIFF
--- a/core/ArrayLib.php
+++ b/core/ArrayLib.php
@@ -164,6 +164,23 @@ class ArrayLib {
 	}
 
 	/**
+	 * Similar to array_map, but recurses when arrays are encountered.
+	 *
+	 * Actually only one array argument is supported.
+	 *
+	 * @param $f callback to apply
+	 * @param $array array
+	 * @return array
+	 */
+	public static function array_map_recursive($f, $array) {
+		$applyOrRecurse = function($v) use($f) {
+			return is_array($v) ? ArrayLib::array_map_recursive($f, $v) : call_user_func($f, $v);
+		};
+
+		return array_map($applyOrRecurse, $array);
+	}
+
+	/**
 	 * Recursively merges two or more arrays.
 	 *
 	 * Behaves similar to array_merge_recursive(), however it only merges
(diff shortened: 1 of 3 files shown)


core/ArrayLib.php
```diff
@@ -164,6 +164,23 @@ public static function in_array_recursive($needle, $haystack, $strict = false) {
...
+	 *
+	 * Actually only one array argument is supported.
+	 *
+	 * @param $f callback to apply
+	 * @param $array array
+	 * @return array
+	 */
+	public static function array_map_recursive($f, $array) {
+		$applyOrRecurse = function($v) use($f) {
+			return is_array($v) ? ArrayLib::array_map_recursive($f, $v) : $f($v);
```
tractorcow: I would suggest `call_user_func($f)` over `$f`, since this extends the ability of this method.

For instance, I might want to pass in `array($this, 'mymapmethod)` to the first parameter.

ntd: Yes, you are right. And AFAIK it does not have any downside.

tractorcow:
Something's up with behat, so I've restarted the build.

If you would like this to get merged do you mind squashing your commits?

ntd:
> If you would like this to get merged do you mind squashing your commits?

Done.

gregsmirnov:
@tractorcow Please merge this PR into 3.2 branch as well, since trimming was introduced in 3.2.0-beta1 https://github.com/silverstripe/silverstripe-framework/commit/7484b4f

tractorcow:
Done https://github.com/silverstripe/silverstripe-framework/pull/5056

PR #4623, opened 2015-09-23
Moved upload files page to main files area

scott1702:
Required CMS PR - https://github.com/silverstripe/silverstripe-cms/pull/1279

Before:
image
image
image

After:
image
image
Hover states when dragging a file:
image
image


chillu:
I like the "drop anywhere". But if you're showing folders on that screen, I'd assume that dragging files onto that folders would upload them into this child folder, not the currently shown one. Assuming that's not implemented, we could grey out the screen with some "drop files here" label when a user drags files over the CMS window, before letting go? 

The centred button and text looks out of place in the AssetAdmin UI, with the left aligned "add folder" button and the right aligned "information" icon. I actually like it better in its original position, in line with the other buttons. We've deemphasised the upload button quite a bit compared to 3.1 (it used to be green and fairly obvious). Looks like this UI change is now trying to emphasise it again. Are we making "white background with blue text" the new "emphasised button" style?

/cc @clarkepaul

scott1702:
Ah yep, I should've included the drag state in the screenshots, updated now

jonom:
Nice stuff! That's a huge improvement to the insert media screen. Curious to see what happens when you select an asset now too - does it take you to an editing form with a back button or something?

I like the new button style and centred look on its own, but agree with Ingo that in this context it looks a bit out of place. Is this the first step in a broader update that this change will be at home in?

In the modal window vertical space is expensive since the height of the modal is limited to a fixed height (I'm not sure why...) so at least there I think the button and description would be better side by side.

Maybe a useful addition would also be to display the upload destination on the drag and drop screen to clear up any potential confusion about where a file is going? Could encourage users to be thoughtful about where they dump files too.

upload

clarkepaul:
I like the addition of the folder location on drag, at the moment it would require extra dev time - I think we can create a list of things to add once we have the base functionality down and that can be one of them.

With the upload button I agree it didn't quite fit with the pattern which is developing. However I still think it is best to keep that cluster of actions centered (upload, drag/drop, add URL) as it is the primary action and leaves space left and right for other actions which are still to come in the two views (gridfield and gallery) as per design attached. In the gallery view there will be a sort dropdown and and access to info about upload requirements.

@jonom We are still to come up with the design for the modal for selecting a file and editing it - coming very soon though - you will be able to select multiple files to edit.

FYI I think we can also look at increasing the modal size.

pasted_image_24_09_15_12_48_pm

jonom:
Cool looking forward to seeing how it develops :)

scott1702:
Just updated with the following:
• Increased modal box both width/height by 100px
• Added some more font icons for the upload button and for future work
• Increased the insert url field to full width of the insert media dialog box

tractorcow:
If you select a folder, then upload a folder, it should go to that folder. It seems to always upload to Uploads.

This is a different issue to displaying the target folder.

tractorcow:
I would also change "choose or select" to "select folder"

tractorcow:
ok, @scott1702 has decided to fix these features in a future release. We can merge this for now.

PR #4625, opened 2015-09-23
FIX: Issues with field focus in edit forms (fixes #4621)

### LINKED ISSUE #4621: First field isn’t focused when editing a record in ModelAdmin/GridFieldDetailForm
In 3.1, when adding/editing a record the first field automatically gets focus. This doesn’t happen in 3.2.0-rc1.

It looks like this is caused by the changes in #2397. What’s super confusing is that I can’t find out what happened to that commit - it was apparently merged into 3.1, yet [blame view](https://github.com/silverstripe/silverstripe-framework/commits/3.1/admin/javascript/LeftAndMain.EditForm.js) can’t find it and Github is derping out saying the commit was removed. The code changes are present in 3.2, but don’t seem to work in `ModelAdmin` or `GridFieldDetailForm` - was the commit reverted from 3.1 or something? :confused:

kinglozzer:


dhensby:
Good spot

PR #4627, opened 2015-09-23
API Support trait loading

tractorcow:


sminnee:
Can we have a test for `getItemPath()`, since that's the part that will actually affect autoloading?

sminnee:
Oh, you had.

dhensby:
Wouldn't we also want a way of saying "get classes with trait"? like we do with interfaces?

PR #4628, opened 2015-09-24
FIX: Repeated template lookups were not cached (fixes #4626)

### LINKED ISSUE #4626: ViewableData/SSViewer caching issues, behavioural change (3.2)
Firstly, in the example below, I believe “Running getter” should only be logged once when it is in fact logged 3 times:

```html
<!-- Page.ss -->
<% if $Test %>
	<% with $Test %>
		$Message
	<% end_with %>
	
	$Test.Message
<% end_if %>
```
```php
// Page_Controller
public function Test() {
	Debug::log('Running getter');
	return ArrayData::create(array('Message' => 'Hi'));
}
```

Running an `<% if $Test %>` **does** cache the value. However, using either `<% with $Test %>` or `$Test.Message` **does not** cache the value - both are translated internally to `obj('Test', null, true)`. This can be fixed by adding a fourth parameter in the template parser to enable caching (`obj('Test', null, true, true)`), but I wanted to log an issue to get some extra eyes on this before I create a PR.

---

The actual behavioural change between 3.1 and 3.2 is caused by this issue and #3985. Before that PR was merged, `ViewableData` would read data from the cache _even if the `$cache` parameter was false_. So things went in this order:

- `<% if $Foo %>` resulted in a cached call, the value of `$Foo` was added to the cache;
- `<% with $Foo %>` resulted in a non-cached call, but `SSViewer` saw the data in the cache and returned it anyway.

After #3895, `SSViewer` will avoid the cache entirely if `$cache` is false - even if the value already exists in the cache. [This gist](https://gist.github.com/kinglozzer/7e21295459bc9fd7dc47) contains my sample code that used to work in 3.1, but no longer works in 3.2.

kinglozzer:
Hopefully the test makes a bit more sense than my wall of text in #4626. Basically, `<% with $Foo %>`, `<% loop $Foo %>` and `{$Foo.Bar}` are not cached.

As far as I can see in git history, they never have been explicitly cached. However, they were somewhat _accidentally_ cached in 3.1. This is what happened in 3.1:

- `<% if $Foo %>` looks up `Foo` and then caches the result
- `<% with $Foo %>` would _not_ be cached, but as the result was already in the cache from the `<% if %>` it would be pulled out anyway

#3985 changed this behaviour, meaning that now if `$cache` is set to false, it avoids the cache - even if `Foo` is already present in the cache.

I don’t like using regular expressions to add a fourth parameter to `obj()` calls, but I can’t see any way to avoid it. Its method signature is different to `XML_val()` and `hasValue()`, yet it looks like the template parser expects it to be identical.

cc @hafriedlander

### DIFF
--- a/view/SSTemplateParser.php
+++ b/view/SSTemplateParser.php
@@ -742,12 +742,16 @@ class SSTemplateParser extends Parser implements TemplateParser {
 		$res['LookupSteps'][] = $sub;
 		
 		$property = $sub['Call']['Method']['text'];
-		
+		$callArgs = "null";
 		if (isset($sub['Call']['CallArguments']) && $arguments = $sub['Call']['CallArguments']['php']) {
-			$res['php'] .= "->$method('$property', array($arguments), true)";
+			$callArgs = "array($arguments)";
 		}
-		else {
-			$res['php'] .= "->$method('$property', null, true)";
+
+		if ($method === 'obj') {
+			// obj() has an extra parameter for caching
+			$res['php'] .= "->$method('$property', $callArgs, true, true)";
+		} else {
+			$res['php'] .= "->$method('$property', $callArgs, true)";
 		}
 	}
 
@@ -3741,8 +3745,9 @@ class SSTemplateParser extends Parser implements TemplateParser {
 			if ($arg['ArgumentMode'] == 'string') {
 				throw new SSTemplateParseException('Control block cant take string as argument.', $this);
 			}
-			$on = str_replace('$$FINAL', 'obj', 
-				($arg['ArgumentMode'] == 'default') ? $arg['lookup_php'] : $arg['php']);
+			$on = str_replace('$$FINAL', 'obj', ($arg['ArgumentMode'] == 'default') ? $arg['lookup_php'] : $arg['php']);
+			// obj() has a fourth argument to enable caching, so add that to the end of the call
+			$on = preg_replace('/(\))$/', ', true)', $on);
 		}
 
 		return
@@ -3775,6 +3780,8 @@ class SSTemplateParser extends Parser implements TemplateParser {
 		}
 		
 		$on = str_replace('$$FINAL', 'obj', ($arg['ArgumentMode'] == 'default') ? $arg['lookup_php'] : $arg['php']);
+		// obj() has a fourth argument to enable caching, so add that to the end of the call
+		$on = preg_replace('/(\))$/', ', true)', $on);
 		return 
 			$on . '; $scope->pushScope();' . PHP_EOL .
 				$res['Template']['php'] . PHP_EOL .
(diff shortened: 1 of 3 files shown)


view/SSTemplateParser.php
```diff
@@ -3775,6 +3780,8 @@ function ClosedBlock_Handle_With(&$res) {
 		}
 		
 		$on = str_replace('$$FINAL', 'obj', ($arg['ArgumentMode'] == 'default') ? $arg['lookup_php'] : $arg['php']);
+		// obj() has a fourth argument to enable caching, so add that to the end of the call
+		$on = preg_replace('/(\))$/', ', true)', $on);
```
tractorcow: This looks like risky code. You're using a regexp to hack in an additional parameter to a method call?

tractorcow: Ok, I see you've addressed this in your PR description. I guess we should get @hafriedlander to look at it.

tractorcow:
Tentatively setting milestone for this issue to 4.0 (even though it's a pr against the 3.2 branch).

kinglozzer:
> Essentially we should REMOVE $forceReturnedObject, always return an object, and make both ViewableData::obj and ViewableData::*_val have the same method signature. Does that remove the need to hack in the extra param and special case for $method === 'obj'?

Yes, making both method signatures identical would solve the issue (and remove the need for either hack). We can definitely fix this (for master) a lot more cleanly if we do it that way.

> I would prefer not changing it again at this stage, though, since it's actually working predictably.

The API in `ViewableData` is more predictable but the way it works in “template-land” is, if anything, _less_ predictable than 3.1 as a result. I agree that we probably shouldn’t revert that commit, though.

My concern is that this is potentially a BC breaking issue when upgrading to 3.2 as this might cause unexpected repeated queries etc with no notice of what’s happening or why: it took me some time to track down why [my example](https://gist.github.com/kinglozzer/7e21295459bc9fd7dc47) wasn’t working in 3.2 but was working in 3.1.

It’s not a critical issue - it’s possible to work around it, at least for my example - but it’s a pain and is potentially a (minor) performance issue depending on what data is being fetched, so I’d _like_ to see this fixed for 3.2 in some way if at all possible.

view/SSTemplateParser.php
kinglozzer: Yeah, not ideal :(. I’m sure it’s possible to achieve the same output some other way, I’m just not anywhere near familiar enough with the parser to do that sort of refactoring!

tractorcow:
Hm, it might be a bit late for 3.2.0, since it's in rc2 at the moment. =(

sminnee:
Shall we rebase this against master and merge it in, @tractorcow @kinglozzer?

kinglozzer:
I think it’s best to wait for #3924 to be merged before looking at this again. Either way this won’t be merged into 3.2, so closing

tractorcow:
I'm in favour of removing $forceReturnedObject in master. :)

PR #4630, opened 2015-09-25
Total Items of a Paginated List

marcokernler:
Just added support for displaying the number of total items contained in a Paginated List across templates

### DIFF
--- a/core/PaginatedList.php
+++ b/core/PaginatedList.php
@@ -435,6 +435,13 @@ class PaginatedList extends SS_ListDecorator {
 	}
 
 	/**
+	 * Returns the total number of items in the list
+	 */
+	public function TotalItems() {
+		return $this->getTotalItems();
+	}
+
+	/**
 	 * Set the request object for this list
 	 *
 	 * @param SS_HTTPRequest


kinglozzer:
@marcokernler Doesn’t `$TotalItems` already work? Using `$TotalItems` should automatically call `getTotalItems()`

kinglozzer:
Oh, I see, you do have to call `$getTotalItems` in templates. Looks like `SS_ListDecorator::TotalItems()` is wrong - it should call `$this->list->count()` instead. I think it would be better to fix this there.

Do you want to have a go at that instead @marcokernler?

marcokernler:
Hi @kinglozzer, just tried your suggestion with calling $this->list->count() within SS_ListDecorator::TotalItems(), but using $Results.TotalItems will only return the number of items for the current page...

Any other suggestions?

kinglozzer:
@marcokernler Hmm, that’s not the behaviour I see if I make that change and the use `$PaginatedList.TotalItems`... how about `$PaginatedList.Count`? That does the same thing

marcokernler:
@kinglozzer Just installed a new SilverStripe instance (3.1.15) and enabled the Site Search. To display the total number of items, i added  `$Results.TotalItems` to the Page_results.ss template right beneath `<% if $Results %>`

Here are the results of my tests:

* Calling `$Results.TotalItems` results in the error: Object->__call(): the method 'totalitems' does not exist on 'ArrayList'.
* Changing `ListDecorator::TotalItems()` from `return $this->list->TotalItems()` to `return $this->list->count()`removes the error, but just will return the number of items per page.
* Calling `$Results.Count`also just return the number of items per page.

Maybe it's because the search does return an ArrayList instead of a PaginatedList?

kinglozzer:
Ah, the page search functionality does use `ArrayList`, which is why `$Results.Count` doesn’t work (and why the fix I suggested won’t work too). However, the search form _does_ store the total item count using `setTotalItems()` (looks like this exact scenario is what that method is for!) - so I’ve come full circle and decided that this is the correct fix after all!

Thanks for bearing with me @marcokernler :stuck_out_tongue_closed_eyes:

PR #4631, opened 2015-09-27
Optimization: Manifest regen. optimizations

SilbinaryWolf:
Updated a few is_array() checks to use the ($var === (array)$var) method as it's much faster. On my 3.2ghz i5, I saw a 4 second speedup when regenerating the manifest on a clean SS build.

### DIFF
--- a/core/manifest/ConfigStaticManifest.php
+++ b/core/manifest/ConfigStaticManifest.php
@@ -183,17 +183,15 @@ class SS_ConfigStaticManifest_Parser {
 	}
 
 	/**
-	 * Get the next token to process, incrementing the pointer
+	 * Get the next token to process, incrementing the pointer, skips any whitespace tokens
 	 *
-	 * @param bool $ignoreWhitespace - if true will skip any whitespace tokens & only return non-whitespace ones
 	 * @return null | mixed - Either the next token or null if there isn't one
 	 */
-	protected function next($ignoreWhitespace = true) {
+	protected function next() {
 		do {
-			if($this->pos >= $this->length) return null;
-			$next = $this->tokens[$this->pos++];
+			$next = ($this->pos >= $this->length) ? null : $this->tokens[$this->pos++];
 		}
-		while($ignoreWhitespace && ($next === (array)$next) && $next[0] == T_WHITESPACE);
+		while(($next === (array)$next) && $next[0] == T_WHITESPACE);
 
 		return $next;
 	}
@@ -328,7 +326,7 @@ class SS_ConfigStaticManifest_Parser {
 		if($token == '=') {
 			$depth = 0;
 
-			while($token = $this->next(false)){
+			while($token = ($this->pos >= $this->length) ? null : $this->tokens[$this->pos++]) {
 				$type = ($token === (array)$token) ? $token[0] : $token;
 
 				// Track array nesting depth
--- a/core/manifest/ConfigStaticManifest.php
+++ b/core/manifest/ConfigStaticManifest.php
@@ -193,7 +193,7 @@ class SS_ConfigStaticManifest_Parser {
 			if($this->pos >= $this->length) return null;
 			$next = $this->tokens[$this->pos++];
 		}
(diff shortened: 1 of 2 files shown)


kinglozzer:
Nice work, looks promising. Which PHP version did you test on? I haven’t had a change to review the changes in any detail, but a quick before/after benchmark looks like this (PHP 5.5):

screen shot 2015-10-12 at 17 10 04

`SS_ConfigStaticManifest` has been massively simplified in master to just use reflection, but I don’t see any reason why we can’t improve the code we’re stuck with until 4.0.

How much of a difference do the changes in `FileFinder` make? It looks like a micro-optimisation to me, but I guess depending on how many files we’re iterating over it could add up to a noticeable difference

SilbinaryWolf:
I'll look at my laptop later to see what version of PHP I'm running on it. I'm assuming it's 5.5 since I installed WAMP on it earlier this year.

The FileFinder changes aren't probably going to make a noticeable difference.

Interestingly enough, inlining the SS_ConfigStaticManifest_Parser->next() function at the very top level yielded another 2 second speedup, but I don't think it's worth it if it's going to make maintaining the code more of a hassle.

core/manifest/ConfigStaticManifest.php
```diff
@@ -183,17 +183,15 @@ function getStatics() {
 	}
 
 	/**
-	 * Get the next token to process, incrementing the pointer
+	 * Get the next token to process, incrementing the pointer, skips any whitespace tokens
 	 *
-	 * @param bool $ignoreWhitespace - if true will skip any whitespace tokens & only return non-whitespace ones
 	 * @return null | mixed - Either the next token or null if there isn't one
 	 */
-	protected function next($ignoreWhitespace = true) {
```
dhensby: Unfortunately we shouldn't be removing these method params on a minor or patch release, so this needs to stay :/

dhensby:
I'm not sure about the filefinder optimisations. Idealistically, we shouldn't be accessing that array directly and making assumptions about the keys; however, as there are default options defined we *can*. If this has performance improvements, I'll accept it.

The whitespace fix is nice if it's saving ~2seconds then it should be considered, but I don't think we can remove that parameter from the function :/ - we can, however, keep the param but inline the logic anyway and add a deprecation warning.

core/manifest/ConfigStaticManifest.php
SilbinaryWolf: Hm. But if we've removed the only case where that function is called with the ignoreWhitespace param then I don't see anything breaking TBH. I also can't imagine anyone using/extending the SS_ConfigStaticManifest_Parser personally.

kinglozzer: I don’t think we can merge this particular change in to 3. It’s _unlikely_ that anyone is extending the parser, but it’s still possible - and this change could be potentially breaking even if we leave the parameter there (what if someone has subclassed the parser, and is using `->next(false)` in their own code?).

tractorcow: If someone is extending it, or if they pass in parameters to the method call (where none is expected) nothing is broken.

It's the inverse that is true; Extending methods WITHOUT the necessary parameters, or calling methods without required parameters included.

kinglozzer: @tractorcow I was referring to the behavioural change that removing the argument implies rather than the method signature. By removing that argument, we’re now making the function _always_ ignore whitespace with no way to opt out

SilbinaryWolf: The only way to overload the SS_ConfigStaticManifest_Parser is to straight up replace it in the framework and I really really doubt that anyone is extending this and using it in personal code.

However, if you guys are blocking this pull due to that, we can just close this and I can just make a pull where I only make the "$var == (array)$var" optimization, as that's where most of the speed up comes from.

dhensby: You don't need to close this PR, you just need to undo this change and push it up (preferably squashed).

Unfortunately SemVer does not make a distinction between APIs we think aren't used and ones we think are.

PR #4640, opened 2015-09-29
BUG: Permission::checkMember() use of undefined variable $codes

memdev:
An undefined variable in the method `checkMember()` was used to generate an SQL `codeClause`.
This resulted in a logical incomplete permission query, which may cause security issues.


tractorcow:
Doh, this needs to be fixed in 3.2 as well. =(

tractorcow:
I've cherry picked this for the 3.2 branch.

PR #4644, opened 2015-09-30
FIX: GridFieldSortableHeader incorrectly reporting fields as sortable

kinglozzer:
`DataObject::hasField()` will return true if a `get<fieldname>()` method exists, so `GridFieldSortableHeader` thinks it can sort on it when it can’t.

Example:
```php
<?php

class MemberExtends extends Extension {
    public function getDisplayName() {
        return $this->owner->Salutation . ' ' . $this->owner->Surname;
    }
}
```

Adding `DisplayName` to `summary_fields`, or otherwise passing it to `GridFieldSortableHeader`, will add the sort arrow/click option to the GridField, but attempting sort will result in a database error:

> Unknown column 'Member.DisplayName' in 'field list'


kinglozzer:
Thanks!

PR #4650, opened 2015-10-05
Fix mysql decimal and enum default sql

mikenz:
decimal has and extra space before 'default' and wasn't quoted.
enum was quoting digit only defaults unnecessarily . Both were causing 'changes' in /dev/build that weren't doing anything.


tractorcow:
Nice fix, but I'm surprised by the way that enum is treating numeric values. Is this because `MySQLSchemaManager::fieldList` is returning unquoted defaults for numeric values?

Please include the php you are using to generate this enum so I can test it locally.

sminnee:
@mikenz @tractorcow are we able to nudge this one along?

mikenz:
Fixed for me, likely in https://github.com/silverstripe/silverstripe-framework/pull/4731

tractorcow:
Sorry I didn't get to this in time. :)

PR #4652, opened 2015-10-05
API Implementation of RFC-1 Asset Abstraction (DataObject Refactoring)

tractorcow:
This pull request in the continuation of the earlier https://github.com/silverstripe/silverstripe-framework/pull/4599, and continues implementation of the RFC specified at https://github.com/silverstripe/silverstripe-framework/issues/3792.

In this pull request I have made the following changes:
* File dataobject now has a single $db field of type DBFile to store the actual filereference.
* File dataobject now implements the same `AssetContainer` interface as DBFile does, in order to provide a standard api for code which must read from and write to files. In a future update, this will be used by UploadField to allow saving into DBFile fields.
* Filesystem synchronisation no longer exists; The reasoning is that there is currently no API for regenerating a local filesystem from the remote asset store. This may be re-implemented in a future update.
* Image dataobject is now a simple deprecated subclass of File. All image manipulations have been moved into a new trait `ImageManipulations`, and are implemented against both `File` and `DBFile`. A new manipulations API has been build upon the variant feature in the asset persistence layer.
* A new API for shortcodes handlers has been added. DBFile (in a future update) will be added.
* Default tuple order changed from hash, filename, variant to filename, hash, variant to better reflect the optionality of hash and variant in certain situations, where filename is the only field necessary in every api call.

Companion CMS pull request at https://github.com/silverstripe/silverstripe-cms/pull/1291


filesystem/File.php
```diff
@@ -686,114 +593,117 @@ protected function updateLinks($old, $new) {
...
+	 * file is moved.
+	 *
+	 * It is necessary for this to be distinct from getFilename in order to ensure that
+	 * we can predict its value prior to save.
+	 *
+	 * @return string
 	 */
-	public function setFilename($val) {
-		$this->setField('Filename', $val);
+	public function getExpectedFilename() {
```
tractorcow: I had trouble coming up with a good name for this method.

filesystem/GD.php
```diff
@@ -1,578 +0,0 @@
-<?php
```
tractorcow: This file was renamed to match the classname.

filesystem/storage/DBFile.php
```diff
@@ -47,8 +53,9 @@ protected function getStore() {
 	);
 
 	public function scaffoldFormField($title = null, $params = null) {
-		// @todo - This doesn't actually work with DBFile yet
-		return new UploadField($this->getName(), $title);
+		return null;
+		// @todo
+		//return new AssetUploadField($this->getName(), $title);
```
tractorcow: This will come in a future pull request.

filesystem/storage/DefaultAssetNameGenerator.php
```diff
@@ -76,6 +83,7 @@ public function __construct($filename) {
 		$this->filename = $filename;
 		$this->directory = ltrim(dirname($filename), '.');
 		$name = basename($this->filename);
+		// Note: Unlike normal extensions, we want to split at the first period, not the last.
```
tractorcow: so that renames end up as `file-v2.tar.gz` not `file.tar-v2.tg`.

forms/FormField.php
```diff
@@ -1260,4 +1260,24 @@ public function castedCopy($classOrCopy) {
 		return $field;
 	}
 
+	/**
+	 * Determine if escaping of this field should be disabled
+	 *
+	 * @param bool $dontEscape
+	 * @return $this
+	 */
+	public function setDontEscape($dontEscape) {
```
tractorcow: I dislike having negated method names, but this was added to provide a setter for an existing property.

core/Convert.php
```diff
@@ -445,4 +445,42 @@ public static function base64url_decode($val) {
...
+
+	/**
+	 * Convert a php.ini value (eg: 512M) to bytes
+	 *
+	 * @todo unit tests
+	 *
+	 * @param string $iniValue
+	 * @return int
+	 */
+	public static function ini2bytes($iniValue) {
```
hafriedlander: Feels a bit icky having these in Convert. I'm not sure where they should go. We don't want Convert to be just a grab-bag of presenters though - ideally it would be "safely convert between types", not "parse and/or present a type".

dev/FixtureBlueprint.php
```diff
@@ -193,12 +195,11 @@ public function createObject($identifier, $data = null, $fixtures = null) {
 				$this->overrideField($obj, 'LastEdited', $data['LastEdited'], $fixtures);
 			}
 		} catch(Exception $e) {
-			Config::inst()->update('DataObject', 'validation_enabled', $validationenabled);
```
hafriedlander: Can maybe use `finally` here if we can get 5.5 support confirmed.

core/Convert.php
tractorcow: I might just move them back to File dataobject then; They didn't hurt being there really.

dev/FixtureBlueprint.php
tractorcow: I have a feeling we'll be limited to 5.4 or below.

dev/DebugView.php
```diff
@@ -300,6 +300,9 @@ public function renderSourceFragment($lines, $errline) {
 	 * @return string
 	 */
 	public function renderTrace($trace) {
+		if(empty($trace)) {
+			return '';
+		}
```
hafriedlander: The code below returns a `</div>`, so does the caller except the closing div? Is there a reason this is just (basically) a silent error swallow rather than returning "Trace: no trace object passed" or something?

docs/en/01_Tutorials/02_Extending_A_Basic_Site.md
```diff
@@ -462,16 +462,19 @@ Nothing here should be new. The *StaffPage* page type is more interesting though
 	:::php
 	<?php
 	class StaffPage extends Page {
-		private static $db = array(
-		);
+
 		private static $has_one = array(
-			'Photo' => 'Image'
+			'Photo' => 'File'
```
hafriedlander: Mentioned in person, but I don't really like that we're loosing the typing here. Maybe `File("Image")` or `File(FILETYPE_IMAGE)` or something? Will review whole diff, then comment more globally.

docs/en/01_Tutorials/02_Extending_A_Basic_Site.md
```diff
@@ -462,16 +462,19 @@ Nothing here should be new. The *StaffPage* page type is more interesting though
...
 		);
 		
 		public function getCMSFields() {
 			$fields = parent::getCMSFields();
 			
-			$fields->addFieldToTab("Root.Images", new UploadField('Photo'));
+			$fields->addFieldToTab(
+				"Root.Images",
+				UploadField::create('Photo')
+					->setAllowedFileCategories('image')
```
hafriedlander: Irritating I'm hitting the docs before the code :). Why a string, and not a const?

docs/en/02_Developer_Guides/03_Forms/Field_types/05_UploadField.md
```diff
@@ -60,52 +78,60 @@ Enable multiple fileuploads by using a many_many (or has_many) relation. Again,
...
-```
 
-```yml
-  Image:
-    extensions:
-        - GalleryImageExtension
-```
+
+	:::yaml
+	Image:
```
hafriedlander: `Image` (the class) doesn't exist now right (or is deprecated), so this would have to be on `File`.

docs/en/01_Tutorials/02_Extending_A_Basic_Site.md
tractorcow: I think a const would be fine; Although the types are user-definable.

docs/en/01_Tutorials/02_Extending_A_Basic_Site.md
tractorcow: This is the `hard` problem to solve for this pull request; The problem is, if the type is NOT image, and an image is uploaded, do we provide the same support for that file as though it was saved in a DBFile('image')? In which case, what's the benefit of the distinction? Is it simply a type hint, or is it only a validation rule to limit allowed mime types?

docs/en/02_Developer_Guides/03_Forms/Field_types/05_UploadField.md
```diff
@@ -213,142 +272,152 @@ First create a [DataExtension](/developer_guides/extending/extensions) like this
...
 
- - `setUploadTemplateName`: (string) javascript template used to display uploading files, see javascript/UploadField_uploadtemplate.js.
+### Configuration Reference
 
- - `setCanPreviewFolder`: (boolean|string) Is the upload folder visible to uploading users? String values are interpreted as permission codes.
+ * `setAllowedMaxFileNumber`: (int) php validation of allowedMaxFileNumber only works when a db 
+   relation is available, set to null to allow unlimited if record has a has_one and
+   allowedMaxFileNumber is null, it will be set to 1.
+ * `setAllowedFileExtensions`: (array) List of file extensions allowed.
+ * `setAllowedFileCategories`: (array|string) List of types of files allowed. May be any of
```
hafriedlander: How does setAllowedFileCategories and setAllowedFileExtensions interact?

filesystem/File.php
```diff
@@ -85,7 +75,7 @@ class File extends DataObject {
 	private static $db = array(
 		"Name" => "Varchar(255)",
 		"Title" => "Varchar(255)",
-		"Filename" => "Text",
+		"File" => "DBFile",
 		"Content" => "Text",
```
hafriedlander: I'll have to have a search, but this twingles my smell sensor - is this the Content of the file somehow stored in the DB? (I'm aware this isn't knew, just worried about breakage).

docs/en/01_Tutorials/02_Extending_A_Basic_Site.md
tractorcow: This is the hard part of this change. I don't have a better solution at the moment.

I think this needs to be addressed at the DBFile level before on the actual DataObject. Read on and get back to me. :D

filesystem/File.php
```diff
@@ -466,35 +432,12 @@ public static function get_app_category($ext) {
...
-	}
-
-	/**
-	 * Return the relative URL of an icon for the file type,
-	 * based on the {@link appCategory()} value.
-	 * Images are searched for in "framework/images/app_icons/".
-	 *
-	 * @return String
-	 */
-	public function Icon() {
```
hafriedlander: Is this restored somewhere else? What's the thinking behind just removing this, but deprecated Image? Just - it's hard to deprecate?

docs/en/02_Developer_Guides/03_Forms/Field_types/05_UploadField.md
tractorcow: `setAllowedFileCategories` internally maps the given categories to extensions, then invokes `setAllowedFileExtensions`

filesystem/File.php
tractorcow: No, the tuple ;D

filesystem/File.php
hafriedlander: getPermaname? getFutureFilename? Might need to catch up IRL, seems like this should be getFilename, and getFilename should be something else (getCurrentPath or something).

core/Convert.php
hafriedlander: Maybe we need a `Humanize` class that converts types to human readable values?

filesystem/Filesystem.php
```diff
@@ -164,80 +164,9 @@ public static function isAbsolute($filename) {
...
-				}
-			}
-		}
-
-		return _t(
-			'Filesystem.SYNCRESULTS',
-			'Sync complete: {createdcount} items created, {deletedcount} items deleted',
-			array('createdcount' => (int)$results['added'], 'deletedcount' => (int)$results['deleted'])
-		);
+		Deprecation::notice("4.0", "Filesystem can no longer be automatically synchronised");
```
hafriedlander: What's our plan for this? I'm happy to have this change, but "you can't sync from the filesystem" seems problematic. I believe this was covered in the RFC somewhat.

filesystem/File.php
tractorcow: It's now in `ImageManipulations`

filesystem/Folder.php
```diff
@@ -501,9 +230,71 @@ public function CMSTreeClasses() {
 	 * @return string
 	 */
 	public function getTreeTitle() {
-		return $treeTitle = sprintf(
+		return sprintf(
 			"<span class=\"jstree-foldericon\"></span><span class=\"item\">%s</span>",
-			Convert::raw2xml(str_replace(array("\n","\r"),"",$this->Title))
+			Convert::raw2att(Convert::nl2os($this->Title, ' '))
```
hafriedlander: nl2os doesn't _remove_ "\n" and "\r", it just converts them to the operating system default. Is that the change you intended?

filesystem/flysystem/FlysystemAssetStore.php
```diff
@@ -325,12 +393,12 @@ protected function getOriginalFilename($fileID, &$variant = '') {
...
 	 * The resulting file will look something like my/directory/EA775CB4D4/filename__variant.jpg
 	 *
-	 * @param string $hash
 	 * @param string $filename Name of file
+	 * @param string $hash Hash of original file
 	 * @param string $variant (if given)
 	 * @return string Adaptor specific identifier for this file/version
 	 */
-	protected function getFileID($hash, $filename, $variant = null) {
+	protected function getFileID($filename, $hash, $variant = null) {
```
hafriedlander: This is the dangerous bit of unsubclassable scalar types - we can't typehint in a way that would catch an error in this refactor

filesystem/Filesystem.php
tractorcow: It was mentioned that it could be done via the default APL, but we would have to implement this in a separate story. The existing system would have to be removed in any case. Talk to @sminnee about this.

The reference to this in the rfc is at https://github.com/silverstripe/silverstripe-framework/issues/3792#issuecomment-69996682

I imagine the apl would need to have some kind of `canSynchronise` and `doSynchronise` api to reverse generate a set of dataobjects from the asset store.

filesystem/storage/DBFile.php
```diff
@@ -66,36 +73,34 @@ public function forTemplate() {
...
 
 	/**
-	 * Given a mime type, determine the template to render as on the frontend
+	 * Determine the template to render as on the frontend
 	 *
-	 * @param string $mimetype
 	 * @return string Name of template
 	 */
-	protected function getTemplateForMime($mimetype) {
+	public function getFrontendTemplate() {
```
hafriedlander: Why the rename?

filesystem/Folder.php
tractorcow: to the operating system default, OR a value passed in. In this case, a space.

I probably should have called that method something else.

filesystem/flysystem/FlysystemAssetStore.php
tractorcow: Hopefully this is the only time I need to do such a dangerous refactor.

filesystem/storage/DBFile.php
tractorcow: Because since it acts on the current object its a "get template for this" not "get template for the given mime".

filesystem/Filesystem.php
tractorcow: As discussed with @sminnee and yourself in other places, sync is dead for the time being, but we can resurrect it another day if we decide to bring it back.

filesystem/File.php
tractorcow: I'll make this getFilename and remove the other method. Internal code will have to invoke $file->File->getFilename() to get the internal value.

core/Convert.php
tractorcow: `Describe::<input>_as_<output>`

docs/en/02_Developer_Guides/03_Forms/Field_types/05_UploadField.md
tractorcow: Un-deprecated again.

docs/en/01_Tutorials/02_Extending_A_Basic_Site.md
tractorcow: Changed as suggested. `DBFile('image')` is new supported

tractorcow:
I've un-deprecated `Image` and implemented support for `DBFile('image')` to allow for image-only dbfile references.

At the last minute I've also added File::join_paths to deal with flysystem being sensitive to the way that paths are joined together.

tractorcow:
I didn't much like the way I had implemented 'unsupported_images', so rather than manipulate the 'image' category (so that it ment different things at different times), I've opted to leave 'image' intact as a category (all image files, including unresizeable) and have created 'image/supported' sub-category, which is used by 'Image' dataobject to specify the list of allowed file types for manipulations.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -28,12 +85,187 @@ Note that this will not allow you to utilise certain file versioning features in
...
+is applied to both `File` and `DBFile`. These both implement a common interface `AssetContainer`, which
+has the `getIsImage()` method. In some cases, it may be preferable to invoke this method to detect
+if the asset is an image or not, rather than checking the subclass, as the asset may also be a `DBFile` with
+an image filter applied, rather than an instance of the `Image` dataobject.
+
+In addition, a new file category `image/supported` has been added, which is a subset of the `image` category.
+This is the subset of all image types which may be assigned to the `[api:Image]` dataobject, and may have
+manipulations applied to it. This should be used as the file type restriction on any `[api:UploadField]` which
+is intended to upload images for manipulation. All image categories have also been specified via constants
+on the `[api:File]` DataObject.
```
chillu: What's the rationale on using constants here? It makes configuration inconsistent, since you still need to use the string keys ("image" rather than `File::CATEGORY_IMAGE`) for any YAML config via `File.app_categories`. We could just check that a string key exists in `File.app_categories`? Moves it from a static to a runtime check, but more consistent between PHP and YAML use. It also means devs can introduce new categories easily - they couldn't create a new constant in `File`.

tractorcow: It was actually a suggestion from @sminnee ; I'm also in favour as it adds better code reflection (find usages of particular file categories, etc).

In yml you need to use the literal strings, so this is seen as a helper for this functionality rather than a replacement. The use of constants over strings also helps improve error detection, since typos become parse errors rather than runtime errors.

There's no requirement that a category HAS to be a constant. I disagree that it is harder to add categories. User code could even create their own constants, it just can't be inside the `File` dataobject.

tractorcow: @sminnee what is your feeling on this?

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -28,12 +85,187 @@ Note that this will not allow you to utilise certain file versioning features in
...
+		$file->setFromLocalFile($tmp, 'imported/'.basename($tmp));
+		$file->write();
+	}
+
+
+Note that 'assets' is no longer present in the new code, and the path beneath what was once assets is now
+used to generate the 'filename' value. This is because there is no longer an assumption that files are
+stored in the assets folder.
+
+Since the structure of `File` dataobjects has changed, a new task `MigrateFileTask` has been added to assist
```
chillu: This should probably get a more prominent position in the changelog overview, to avoid devs just running "dev/build" (which is usually assumed to be fairly non-destructive). Also, `FileMigrationHelper` looks like it might run out of memory or exeuction time on large filesystems - we should add a recommendation to run this from the command-line, and use `increase_time_limit_to()`/`increase_memory_limit_to()` the in `MigrateFileTask`

tractorcow: Good idea, I'll add this in. ;D

docs/en/04_Changelogs/4.0.0.md
chillu: A category constant could be created on a class other than `File`, but then you still need to duplicate its string value in `File.app_categories` YAML - unless you use `Config::inst()->update()` where you can refer to constants. The constant use just feels like a half-hearted attempt at making a config schema, providing a false sense of security and making it more likely that config keys get out of sync in projects (hard to "global find and replace" due to different notations). We can throw exceptions whenever an invalid key is used, just not through static analysis.

filesystem/File.php
```diff
@@ -9,97 +14,120 @@
...
 
 	private static $db = array(
-		"Name" => "Varchar(255)",
-		"Title" => "Varchar(255)",
-		"Filename" => "Text",
-		"Content" => "Text",
+		"Name" =>"Varchar(255)",
+		"Title" =>"Varchar(255)",
+		"File" =>"DBFile",
+		"Content" =>"Text",
```
chillu: Probably a good time to remove the `File.Content` field? Its not exposed via the CMS form, and stays unpopulated unless you write custom functionality - so quite confusing to newcomers. The [textextraction](https://github.com/silverstripe-labs/silverstripe-textextraction) module uses a separate `File.FileContentCache` field as well.

filesystem/File.php
```diff
@@ -224,28 +282,27 @@ public static function link_shortcode_handler($arguments, $content = null, $pars
...
 	public static function find($filename) {
-		// Get the base file if $filename points to a resampled file
-		$filename = Image::strip_resampled_prefix($filename);
-
 		// Split to folders and the actual filename, and traverse the structure.
 		$parts = explode("/", $filename);
 		$parentID = 0;
 		$item = null;
 		foreach($parts as $part) {
-			if($part == ASSETS_DIR && !$parentID) continue;
```
chillu: Skipping `assets/` on the root level might still be a useful feature? It can lead to subtle bugs in existing code to drop this (only occurs for some files). You couldn't find a file with `<webroot>/assets/assets/somefile.png` with the check in place, but that's already a bug in the implementation. If it's removed, we have to be clearer about the expected file path in `@param`

filesystem/File.php
```diff
@@ -504,103 +545,67 @@ public function onAfterUpload() {
...
-			if(!$success) throw new Exception("Cannot move $pathBeforeAbs to $pathAfterAbs");
+		// Check path updated record will point to
+		// If no changes necessary, skip
+		$pathBefore = $this->File->getFilename();
+		$pathAfter = $this->getFilename();
+		if($pathAfter === $pathBefore) {
+			return false;
 		}
 
+		// Copy record to new location via stream
```
chillu: This changes the semantics of `updateFilesystem()` to *copy* the file, rather than *move* it. What's the reason for keeping it in its old place? Also, reading from a stream while writing to it sounds like a recipe for edge cases in some adapters. Flysystem has `rename()`?

filesystem/storage/DBFile.php
```diff
@@ -0,0 +1,457 @@
...
+		$this->setAllowedCategories($allowed);
+	}
+
+	/**
+	 * Determine if a valid non-empty image exists behind this asset, which is a format
+	 * compatible with image manipulations
+	 *
+	 * @return boolean
+	 */
+	public function getIsImage() {
```
chillu: "supported image" is a bit of an ambiguous concept here: It might be displayable via an `` tag, we just don't support processing it with the available `Image_Backend` implementations. So `getFrontendTemplate()` might return empty on images which otherwise would be fine in ``. It's such an edge case that we can probably ignore it though.

tractorcow: At the moment, the "can be displayed in an img tag" and "can be resized" are equivalent tests. We could make a new API for "get the tag of the arbitrary file", but we'd end up with a diverged purpose of the `Image` dataobject. As it stands, the expectations of it are pretty clear, and it would be awkward to muddy the definition. The `image` / `image/supported` is probably as awkward as I'd like the distinction to get, since even within non-supported images, not all are embeddable in HTML.

chillu:
* The local `AssetAdapter` should respect `Filesystem.folder_create_mask` and `Filesystem.file_create_mask`
 * `Filesystem::makeFolder()` still uses `mkdir()`, which is a bit weird if you opt to only store files in S3. I guess there's usually a local cache of S3 files which means a local filesystem folder is still needed, but from my understanding of the API local file storage isn't *required*. Why aren't we using `League\Flysystem\AdapterInterface->createDir()`?

filesystem/File.php
tractorcow: Unfortunately our RFC never covered deletion or removal of files, so, it's out of scope of this unless we can decide how to implement it at the interface level.

filesystem/flysystem/FlysystemAssetStore.php
```diff
@@ -180,58 +192,110 @@ protected function isSeekableStream($stream) {
...
+		return AssetStore::CONFLICT_RENAME;
+	}
+
+	/**
+	 * Determine if legacy filenames should be used. These do not have hash path parts.
+	 *
+	 * @return bool
+	 */
+	protected function useLegacyFilenames() {
+		return Config::inst()->get(__CLASS__, 'legacy_filenames');
```
chillu: Using `__CLASS__` for config settings is a bit of an anti-pattern IMHO, since subclasses would break functionality by default (no "inheritance tree" in YAML config)

filesystem/File.php
tractorcow: the assets folder shouldn't really be referenced anywhere in the user code anymore. Leaving it in one place would mean having to implement this exception across the board, and would be a risky invitation to users to continue referencing it in their user code.

For example, if we omit the asset when finding an existing filename, we'd have to ensure that creation of new files omitted the asset prefix, otherwise findOrCreate methods could behave unpredictably.

If people try to create a file in the new API with the Filename `assets/parent/file.jpg` then it does less harm to just create the assets parent directory than to try to work around it. As long as the approach is consistent, it shouldn't break anything.

filesystem/File.php
tractorcow: I think so. :D

tractorcow:
In general, Filesystem class has anything to do with assets anymore, so it should probably be relegated to simply a folder helper for testing / temp directory manipulation.

I'll have a look at customising the config of the create_mask.

Our new APL doesn't care about folders, and leaves it up to the backend to manage those. Thus, Filesystem::makeFolder shouldn't touch anything in assets (other than testing with the hard-coded asset store). If a local cache of s3 files is needed, then it's up to the hybrid s3 backend (yet undeveloped) to create these folders as necessary.

The primary objects in the new APL are physical files, and the `Folder` dataobject only exists now as a logical organisation mechanism, and doesn't necessarily rely on a physical folder anymore. In fact DBFile doesn't even have any relation to Folder. :)

filesystem/flysystem/FlysystemAssetStore.php
tractorcow: I should use `get_class($this)`

docs/en/02_Developer_Guides/14_Files/01_File_Management.md
```diff
@@ -122,16 +122,16 @@ Other than files stored exclusively via DBFile, files can also exist as subclass
...
+| Field name     | Description																																			|
+| ----------     | -----------																																			|
+| `ClassName`    | The class name of the file (e.g. File, Image or Folder).																								|
+| `Name`         | The 'basename' of the file, or the folder name. For example 'my-image.jpg', or 'images' for a folder.												|
+| `Title`        | The optional, human-readable title of the file for display only (doesn't apply to folders).															|
+| `File`         | The `[api:DBFile]` field (see above) which stores the underlying asset content.																		|
+| `Content`      | Typically unused, but handy for a textual representation of files. For example for fulltext indexing of PDF documents.								|
+| `ShowInSearch` | Whether the file should be shown in search results, defaults to '1'. See ["Tutorial 4 - Site Search"](/tutorials/site_search) for enabling search.	|
+| `ParentID`     | The ID of the parent Folder that this File/Folder is in. A ParentID of '0' indicates that the File/Folder is in the 'assets' directory.				|
```
chillu: That's not always correct, we should rephrase to "asset root, usually a folder named `assets` for local file storage"

filesystem/File.php
chillu: OK, I'm convinced :) Guess its consistent not to mention anything about "assets" in the PHPDoc either then - and I've confirmed its well documented in the upgrading guide as well as `01_File_Management.md`

filesystem/File.php
chillu: The inability to delete files from the asset store seems like a fairly big omission in terms of getting this feature production ready. The [RFC](https://github.com/silverstripe/silverstripe-framework/issues/3792) mentions "...we also need to decide how to provide for the delete, copy, move and "check if exists" operations.". Do we need a backlog item for this? It's going to be a tricky one, given the multiple DBField references you mentioned. But simply not allowing unless you drop down two levels of abstraction (`$dbField->getStore()->getFilesystem()`) will confuse a lot of devs.

filesystem/flysystem/FlysystemAssetStore.php
chillu: `get_class()` has the same issue though, right? `class MyFlysystemAssetStore extends FlysystemAssetStore` will return `MyFlysystemAssetStore`. And since the class is not `final`, and probably injectable, we can't use the instance class name.

filesystem/Filesystem.php
```diff
@@ -162,82 +162,4 @@ public static function isAbsolute($filename) {
...
-	 * back to the equivalent filesystem record.
-	 *
-	 * @param int $folderID Folder ID to sync along with all it's children
-	 * @param Boolean $syncLinkTracking Determines if the link tracking data should also
-	 *        be updated via {@link SiteTree->syncLinkTracking()}. Setting this to FALSE
-	 *        means that broken links inside page content are not noticed, at least until the next
-	 *        call to {@link SiteTree->write()} on this page.
-	 * @return string Localized status message
-	 */
-	public static function sync($folderID = null, $syncLinkTracking = true) {
```
chillu: This should also remove `Filesystem.sync_blacklisted_patterns`

chillu:
> In general, Filesystem class has nothing to do with site assets anymore, so it should probably be relegated to simply a folder helper for testing / temp directory manipulation.

OK, then we need to remove the `makeFolder()` side effect of creating a `Folder` record - otherwise we're implying the method is working alongside the asset store, right? And update the API docs + changelog to point out the changed semantics of the `Filesystem` API. I get that the asset store abstraction doesn't really deal with folders, but many backends (such as local and S3) support them. There'll still be *some* need for creating folders records, for example default locations for `UploadField` before the first file is uploaded. Maybe that's contained in the backlog task for creating the S3 adapter though?

> Our new APL doesn't care about folders, and leaves it up to the backend to manage those.

It's a bit of a leaky abstraction though. At the moment, calling `$myFolder->delete()` will delete the record and all children `File` and `Folder` records, but leave all filesystem entries in place. That's very different from what 3.x does, and IMHO too big of an API change for such a commonly used feature.

docs/en/02_Developer_Guides/14_Files/01_File_Management.md
tractorcow: Yep that should be renamed.

docs/en/04_Changelogs/4.0.0.md
tractorcow: I'll un-const it then ;D

filesystem/File.php
tractorcow: It'll have to be another story then. Deletion isn't impossible, but it isn't currently safe (without the risk of creating broken links). When there was a 1-to-1 relationship between File dataobjects and actual files, this wasn't a problem, but now it's but one mechanism for referencing files, and the absense of a File dataobject (i.e. deletion) doesn't necessarily mean the absence of references to the underlying physical file.

filesystem/flysystem/FlysystemAssetStore.php
tractorcow: get_class($this) is fine because config inherits.

tractorcow:
> OK, then we need to remove the makeFolder() side effect of creating a Folder record 

I've done that in Folder.php. ;)

> There'll still be some need for creating folders records, for example default locations for UploadField before the first file is uploaded.

These are just strings, and are only used to generate the `Filename` property of new files. There should be no need at all for explicit "create folder" or "change folder" API.

> It's a bit of a leaky abstraction though. At the moment, calling $myFolder->delete() will delete the record and all children File and Folder records, but leave all filesystem entries in place. That's very different from what 3.x does, and IMHO too big of an API change for such a commonly used feature.

tractorcow:
I need another word for "physical folder/file" and "dataobject folder/file".

chillu:
Had an offline chat with Damian, and he pointed out that `Folder->write()` with a different name in fact does propagate to its children (via `File->updateFilesystem()`). The problem is that it's *copying* files rather than moving them, leaving the old files in place. This is an issue because anybody using this feature will expect URLs to the old files to become invalid, and subsequent deletes to actually delete the file (my example use case for this was an accidental upload of a sensitive document which needs to be removed again).

I've re-read RFC discussions (#3792) on this again. Mateusz said:

> The abstract APL doesn't care about the Folders. The Folder is a construct used by File & Images to structure the files in the CMS via the database. A concrete APL might store files in a similar tree on the filesystem (read up on Simple APL), but it does not guarantee to do so.

Marcus said: 

> The Model level representation (Folder / File objects) shouldn't dictate the file system representation, so that changing the title of a Folder data object shouldn't trigger a cascading change to any of the blobs on disk (or remote storage). Some file backends may not support a 'simple' way to rename folders in a file hierarchy, while it may be way too expensive in terms of number of operations for other backends - and a failure halfway through may not be easily recovered.

jonom:
@tractorcow What about 'file asset' and 'directory' when talking about file
system concepts and 'file object' and 'folder object' when talking about
DOs? Or 'wrapper'? 'Physical' gets tossed around a lot but I don't think
that label is any more appropriate for a file than a database object -
they're both only physical in the sense that they're stored (traditionally)
in grooves on a disk.

tractorcow:
> The Model level representation (Folder / File objects) shouldn't dictate the file system representation, so that changing the title of a Folder data object shouldn't trigger a cascading change to any of the blobs on disk (or remote storage). Some file backends may not support a 'simple' way to rename folders in a file hierarchy, while it may be way too expensive in terms of number of operations for other backends - and a failure halfway through may not be easily recovered.

Probably a bit past this point, but the folder / file structure is synchronised with the file backend. It could be turned off, I suppose, but it could potentially be problematic. E.g. when doing File::find. If the filename of the dataobjects and apl no longer match, which one should be searched?.

tractorcow:
> The local AssetAdapter should respect Filesystem.folder_create_mask and Filesystem.file_create_mask

Instead of having AssetAdaptor inherit the Filesystem permissions, I've created a config api compatible `file_permissions` config directly on that class.

tractorcow:
@chillu updated based on your feedback:

- Migration task moved to top of changelog, and is set to no max memory / time limit. The migrate_legacy_file config is still available, but now turned off by default.
- constants for categories are reverted back to strings.
- file / folder creation masks for the default asset store are configurable, although it doesn't use the same mask as per the old `Filesystem` object.
- `Filesystem.sync_blacklisted_patterns` is removed.
- `File::Content` db field is gone.
- Documentation has been updated.

chillu:
Awesome, thanks Damian! 

The upgrading docs should document the switch from `Filesystem.folder_create_mask` to `AssetAdapter.folder_permissions`. And given that it's a bit harder to identify this setting now, I think it also needs to get a mention in `01_File_Management.md`.

> `File::Content` database field is removed

To avoid freaking out people reading the upgrade guide about potentially destructive actions, I'd add " since it is not populated by default". 

I'm waiting for another +1 on this before merging. @hafriedlander @sminnee @mateusz?

tractorcow:
Well, I may as well wait for more feedback before pushing up more minor changes. I'll update once a few more people have had a chance to review.

I think @sminnee definitely needs to see this conversation before we make any new API decisions.

docs/en/04_Changelogs/4.0.0.md
sminnee: I just copied the idea from Monolog... ;-) I'm fine with us just using strings.

sminnee:
Hey everyone,

I think that asset deletion/moving needs to be dealt with before 4.0-alpha1 but I don't think that it blocks the merge of this PR. I've raised #4677 to track this.

On the merging of PRs that still have things to finish: I'm much more keen to get several incremental changes to master rather than having long-lived PRs (see http://martinfowler.com/bliki/FeatureBranch.html for a bit of the rationle of why long-lived PRs are bad).

PR #4654, opened 2015-10-05
BUG Fix duplicate HolderID on TreeDropdownField

### LINKED ISSUE #4624: TreeDropdownField has duplicate HTML IDs (3.2)
`TreeDropdownField.ss` includes `$HolderID`, but it already gets that ID from `FormField_holder.ss`. I haven’t checked if this applies to any other fields.

tractorcow:
Fixes #4624


tractorcow:
I've manually tested that this field continues to work in the CMS areas:
- Adding a new page
- Editing location of a file
- Editing parent of a page under settings

PR #4659, opened 2015-10-06
possibility to override default cache lifetime

gherkins:
this would be really useful imho, as you could define the default cache lifetime in `_ss_environment.php`
`define("SS_CACHE_DEFAULT_LIFETIME", 60*60*8 );`

So that the configuration (`SS_Configuration`) itself would be cached longer than the default 10 minutes.

### DIFF
--- a/cache/Cache.php
+++ b/cache/Cache.php
@@ -27,11 +27,21 @@ class SS_Cache {
 	protected static $cache_lifetime = array();
 
 	/**
+	* @var int $default_cache_lifetime
+ 	*/
+	protected static $default_cache_lifetime = 600;
+	
+	/**
 	 * Initialize the 'default' named cache backend.
 	 */
 	protected static function init(){
 		if (!isset(self::$backends['default'])) {
 			$cachedir = TEMP_FOLDER . DIRECTORY_SEPARATOR . 'cache';
+
+
+			if( defined("SS_CACHE_DEFAULT_LIFETIME") ){
+				self::$default_cache_lifetime = constant("SS_CACHE_DEFAULT_LIFETIME");
+			}
 
 			if (!is_dir($cachedir)) {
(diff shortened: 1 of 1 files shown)


gherkins:
Just to clarify:

Of course you can set almost any cache lifetimes via `SS_Cache::set_cache_lifetime()` but not the one for the configuration, since its cache is set up before actually reading it.
So that the configuration would always be cached with the default lifetime, which is at the moment hardcoded to 600secs.

In a production environment, I would rather cache the configuration for a way longer time.

Overriding the default lifetime seemed like a quick solution, that would not break anything...

gherkins:
@tractorcow May I ask why this is rejected without a comment?

I can see this being more of a workaround, 
but wouldn't you consider this an issue worth investigating?

kinglozzer:
@gherkins I believe it was automatically closed by Github because the “develop” branch that you opened this against was deleted (that branch shouldn’t have ever existed anyway!). The change looks like it makes sense to me. As it’s an enhancement, it will need to be submitted against the 3 branch

kinglozzer:
@gherkins Looking at `SS_ConfigManifest`, the [lifetime is set to `null`](https://github.com/silverstripe/silverstripe-framework/blob/3/core/manifest/ConfigManifest.php#L118) - does that not make it infinite? I believe that’s the intention, even if it doesn’t work

gherkins:
@kinglozzer thanks for your reply !

you're right - my bad.

The default cache lifetime does end up in the `$basicOptions` for `SS_Configuration`
https://github.com/silverstripe/silverstripe-framework/blob/3/cache/Cache.php#L180
...but is then overwritten with `null` when merged with the `$frontendOptions`.
This results in infinite expiry as expected.

Thanks again!

Issue #4663, opened 2015-10-06
TabSet does not follow parent return values. Affects all versions.


patricknelson:
Hi, I noticed that `TabSet` doesn't appear to be consistently following the return values of it's parent class `CompositeField` for a few methods of interest to me, particularly `->insertBefore()` and `->insertAfter()`. The framework's API is already very limited with respect to the ability to recursively find fields, so I ended up needing to resort to these methods to see if a composite field existed (for my unit testing) since only DataFields can currently be found recursively.

Anyway -- I noticed that even though it appears that `FieldList` and `CompositeField` were returning a non-null value if it was able to insert before/after, I should be good -- but no matter what, I could never get the result I needed. Here is the reason: https://github.com/silverstripe/silverstripe-framework/blob/3/forms/TabSet.php#L154 

The end of this method isn't returning anything, so that completely mitigates most situations where at least I could find this useful (e.g. `->getCMSFields()` which always has a `TabSet` if I recall). 

**Question:** I can fix this easily. What branch should I put this in? I'm currently using `3.1`.

jonom:
I'll let someone else decide where a change like this should go, but I feel your pain on form field traversal. See https://groups.google.com/forum/#!topic/silverstripe-dev/c3RajiT7gMI for some ideas about this and feel free to reply there if you have anything to add. If we can get some discussion going on it maybe we can get to RFC stage and improve the situation for 4.0.

patricknelson:
@jonom Ok thanks. I'm going ahead with just `3` since nothing else is going really into `3.2` (since it's almost "stable" I guess, even though I think this would make it *more stable*) nor are items going into `3.1` unless they're serious, so `3` is the safest bet for now.

Yeah this is definitely in need of an API overhaul due to the legacy nomenclature clashing with other methods (unrelated, but using different prefixes like "find" or suffixes like "ByName"). I was thinking a recursive function could be like `->findByName()` but you end up with with another method that is confusingly similar but implements dot notation (`->fieldByName()`). I don't even want to get started on that . . .

[...]

PR #4664, opened 2015-10-06
FIX for #4663 ensuring TabSet return values are retained from parent.

patricknelson:
Proposed solution to address #4663. 

Please also see note about the unrelated (but redundant) override of `removeByName` method. Why is that even there?

### DIFF
--- a/forms/TabSet.php
+++ b/forms/TabSet.php
@@ -114,7 +114,10 @@ class TabSet extends CompositeField {
 	}
 
 	/**
-	 * Returns the named tab
+	 * Returns a named field.
+	 *
+	 * @param string $name Name of the field you want to find. Allows for dot notation.
+	 * @return FormField|null
 	 */
 	public function fieldByName($name) {
 		if(strpos($name,'.') !== false)	list($name, $remainder) = explode('.',$name,2);
@@ -135,6 +138,8 @@ class TabSet extends CompositeField {
 				}
 			}
 		}
+
+		return null;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


forms/TabSet.php
```diff
@@ -152,17 +152,18 @@ public function push(FormField $field) {
...
 	public function insertAfter($insertAfter, $field) {
-		parent::insertAfter($insertAfter, $field);
 		if($field instanceof Tab) $field->setTabSet($this);
 		$this->sequentialSet = null;
+		return parent::insertAfter($insertAfter, $field);
 	}
-
+	
+	// TODO: What's the point in overriding this if it doesn't do anything?
 	public function removeByName( $tabName, $dataFieldOnly = false ) {
```
tractorcow: It's been this way since the very first time it was added to source control.

You can remove this method.

forms/TabSet.php
```diff
@@ -152,17 +152,18 @@ public function push(FormField $field) {
...
-		parent::insertBefore($insertBefore, $field);
 		if($field instanceof Tab) $field->setTabSet($this);
 		$this->sequentialSet = null;
+		return parent::insertBefore($insertBefore, $field);
 	}
 
 	public function insertAfter($insertAfter, $field) {
-		parent::insertAfter($insertAfter, $field);
 		if($field instanceof Tab) $field->setTabSet($this);
 		$this->sequentialSet = null;
```
tractorcow: This variable (sequentialSet) is a hangover from when CompositeField used to extend FieldList. It doesn't anymore, and is no longer useful.

patricknelson:
@tractorcow I've gone in and updated some of the items you've pointed out.

Do you think maybe it would make sense to NOT have methods return different datatypes when it's not necessary? That is, like with `->insertAfter()` and `->insertBefore()` they return `false` (and only `false`) if they don't return `FormField`. Why change the data type when you can just return nothing? Typically you'd return an object (or child of that object or typehint an interface) 100% of the time or, if not that, then you'd just return null (e.g. `->first()` operation on a Laravel empty collection). I understand false/null are both falsey, but one is a type with multiple possible values and the other is datatype that is more commonly used when objects are involved. Actually, `null` is literally the return value of a function that doesn't return anything (in PHP at least, it's version of void). I know -- it's called a function :) 

Is that the sort of API change you'd avoid doing even in `3` or no? I'll throw it out here so you can see what I mean in a moment but I can change back easily.

patricknelson:
@tractorcow 

Ok done. As you can see now it's a bit easier to read/understand with not only explicit return types (not just the implicit `null` if no return is provided) but also documenting the `@return` as well.

forms/TabSet.php
```diff
@@ -114,7 +114,10 @@ public function getAttributes() {
 	}
 
 	/**
-	 * Returns the named tab
+	 * Returns a named field.
+	 *
+	 * @string string $name Name of the field you want to find. Allows for dot notation.
```
tractorcow: `@string string` ?

tractorcow:
I think that changing the return type will be a semver violation; but you can do this in 4 if you like.

By the way, we are aiming to get 4.0 out early next year. 3.2 is the only disaster release that should ever take as long as it has. :D

forms/TabSet.php
patricknelson: Oops, I meant `@param` there.

patricknelson:
Ok -- that's back to just `false` for now, then. 

Later on, I think there should just be a larger effort at some point (possibly new issue?) to ensure all methods that optionally return an object or a falsey value get normalized to returning `null` instead of `false` and, if they do return `null`, it's via explicit return and not implicit. If you look right above to `->dataFieldByName()` you'll see it returns null instead of falls (missing a `return` statement, *gasp*). 

Too busy on other stuff to weed through all that, just "as long as it works" ...

PR #4667, opened 2015-10-07
FIX: castingHelper failed to find many_many_extraFields data (fixes #4661)

### LINKED ISSUE #4661: many_many_extraFields are not cast correctly in templates
They are always cast to whatever `default_cast` is set to. Easiest fix is probably to add a `DataObject::castingHelper()` method that looks for casting in extra fields.

kinglozzer:


view/ViewableData.php
```diff
@@ -249,6 +249,12 @@ public function castingHelperPair($field) {
 	 */
 	public function castingHelper($field) {
 		if($this->hasMethod('db') && $fieldSpec = $this->db($field)) {
+			Deprecation::notice(
+				'4.0',
+				'ViewableData::castingHelper() will no longer extract casting information "db". Please override
+				castingHelper in your ViewableData subclass.',
+				Deprecation::SCOPE_GLOBAL
+			);
 			return $fieldSpec;
```
kinglozzer: I figured this “db” functionality is better placed in `DataObject::castingHelper()` now that I’ve added that method, so it can be removed from here. In theory someone might be relying on it for a `ViewableData` subclass, so I’ve taken the safer route of deprecating it instead of removing it now.

model/DataObject.php
```diff
@@ -1782,6 +1804,11 @@ public function getManyManyComponents($componentName, $filter = null, $sort = nu
 
 		$extraFields = $this->manyManyExtraFieldsForComponent($componentName) ?: array();
 		$result = ManyManyList::create($componentClass, $table, $componentField, $parentField, $extraFields);
+
+		// Store component data in query meta-data
+		$result = $result->alterDataQuery(function($query) use ($extraFields) {
+			$query->setQueryParam('Component.ExtraFields', $extraFields);
+		});
```
kinglozzer: This is, as far as I can see, the most efficient way of ensuring we have access to this info in `castingHelper()`. No extra lookups/singletons are needed. Wasn’t really sure what to name the parameter though...

view/ViewableData.php
tractorcow: This has already been refactored in master.

tractorcow:
Non-critical changes are too late to go into 3.2.0, by the way, I'm afraid. =(

Would this be able to go into 3.3? It is a fix to an issue, but it introduces new API as well. At best this could go into 3.2.1.

kinglozzer:
That’s fine, I was assuming it wouldn’t be merged until 3.2.0 is out. I did think 3.2.1 as I thought it’s primarily a bug fix, but I don’t mind 3.3 if that’s preferred - there is a workaround after all :)

tractorcow:
@kinglozzer make sure you check out the DataObject dbfield changes in master. Your changes are still a great enhancement to these though, but worth looking at how these would be eventually merged. :)

dhensby:
@tractorcow does our default casting fix deal with `many_many_extraFields` approiately?

tractorcow:
For extraFields it just works the same way as this change, except it'll fall back to whatever is set as default_cast for the parent object. So yeah it should work fine. :)

PR #4671, opened 2015-10-08
FIX: Missing dotfiles matched as root URL

threesquared:
Missing dotfiles for example `.passwd` are matched as files and split by the dot to retrieve the extension. However this means `$this->url` is set to empty which in turn means `$this->dirParts` is empty which is then matched as the root URL instead of returning a 404.

This regex modification only matches files with names. I thought this was the least intrusive way to fix this issue.


tractorcow:
I think this pull request will need a unit test, since it covers a potentially breaking area of the code.

dhensby:
Bump

PR #4673, opened 2015-10-09
FIX #4392: Ensure headers are checked first before being clobbered by globally maintained state.

### LINKED ISSUE #4392: HTTP::add_cache_headers() always overrides headers that are already set.

patricknelson:
Proposed fix for #4392. Unit tests are coming soon.

**Note:** This is a duplicate of the old PR #4393 to bring it up to date with the `3.2` branch.

### DIFF
--- a/control/HTTP.php
+++ b/control/HTTP.php
@@ -448,8 +448,14 @@ class HTTP {
 
 		// Now that we've generated them, either output them or attach them to the SS_HTTPResponse as appropriate
 		foreach($responseHeaders as $k => $v) {
-			if($body) $body->addHeader($k, $v);
-			else if(!headers_sent()) header("$k: $v");
+			if($body) {
+				// Set the header now if it's not already set.
+				if ($body->getHeader($k) === null) {
+					$body->addHeader($k, $v);
+				}
+			} elseif(!headers_sent()) {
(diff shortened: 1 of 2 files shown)


patricknelson:
Alright @kinglozzer and @dhensby, modernized for `3.2` and dropped in a test case. I updated the test there to ensure it stays isolated since, as it seems (and as I feared) there *were* scenarios where it depended upon the clobbering of existing values which is core to the change I made. That's why I separated the tests out to use different instances of `SS_HTTPResponse` so that they don't contaminate each other.

dhensby:
Ok, let's do this. thanks :)

patricknelson:
Great success. Thanks @dhensby!

PR #4674, opened 2015-10-12
Fix page reordering bug with ParentID

ss23:
If you are viewing PageA in the CMS, but move PageB into PageC, the edit form will recieve an edit form ParentID of PageC.
This is incorrect, as only PageB had it's ParentID change.


ss23:
cc @chillu

PR #4675, opened 2015-10-12
Only notify users about PDO native mode once

kinglozzer:
Users don’t need to be notified about this for every table! I’ll await others’ input on whether this is a bug and should be fixed against 3.2.

Before:
screen shot 2015-10-12 at 16 45 49

After:
screen shot 2015-10-12 at 16 48 10

### DIFF
--- a/model/connect/MySQLSchemaManager.php
+++ b/model/connect/MySQLSchemaManager.php
@@ -125,9 +125,16 @@ class MySQLSchemaManager extends DBSchemaManager {
 	}
 
 	public function checkAndRepairTable($tableName) {
+		// Flag to ensure we only send the warning about PDO + native mode once
+		static $pdo_warning_sent = false;
+
 		// If running PDO and not in emulated mode, check table will fail
 		if($this->database->getConnector() instanceof PDOConnector && !PDOConnector::is_emulate_prepare()) {
-			$this->alterationMessage('CHECK TABLE command disabled for PDO in native mode', 'notice');
+			if (!$pdo_warning_sent) {
+				$this->alterationMessage('CHECK TABLE command disabled for PDO in native mode', 'notice');
+				$pdo_warning_sent = true;
+			}
+
 			return true;
 		}
 


tractorcow:
If you wanted to avoid adding a new api method, you could just have a flag (e.g. `static $sent = false;`) on the schema manager that remembers if the message was sent.

dhensby:
It may be better to store the hashes as the key of the variable and do an isset check. 

Also, please make the new var private and add functions to add to and remove from the array. 

The new API, either as implemented or as suggested by @tractorcow, mandates a 3.3 release, I'd have thought.

kinglozzer:
> If you wanted to avoid adding a new api method, you could just have a flag (e.g. `static $sent = false;`) on the schema manager that remembers if the message was sent.

That sounds like a better approach. I’ve made that change now, so no new API has been introduced. I didn’t make the new property static - nothing else in the schema managers is static (except config vars) so it kinda stuck out to me. Can change if preferred.

dhensby:
If you're going to add a private var you really should have accessors to allow extensibility

kinglozzer:
> If you're going to add a private var you really should have accessors to allow extensibility

But that would be an API addition which would mean this has to go against 3.3 instead of 3.2, right? :(

dhensby:
That's right - this is just a minor annoyance and not a critical bug, so I'd say it shouldn't go into 3.2 anyway

tractorcow:
I ment:

```php
public function checkAndRepairTable($tableName) {
	static $sent = false;
}
```

That's not new API.

kinglozzer:
I’d completely forgotten that you can do that in PHP, awesome. I guess the jury’s still out on whether this is 3.2 or 3.3... IMO if we can fix this annoyance with zero BC risk and no new APIs, then why not put it in 3.2? :)

dhensby:
> IMO if we can fix this annoyance with zero BC risk and no new APIs, then why not put it in 3.2?

agreed

PR #4676, opened 2015-10-13
Bugfix for DropdownField->validate (can't validate null)

JorisDebonnet:
Blog 2.0 errored when I tried to add a blog item without tag or category (on SilverStripe 3.2). Perhaps that's a bug in the blog module, but why not make this validator be able to handle null values.


JorisDebonnet:
hold on ... working on a better fix

JorisDebonnet:
Well, never mind... the issue was with installing a version of TagField (through Blog) that did not yet overwrite DropdownField's validate method... I guess the bug then is with Blog still requiring TagField 1.0.x instead of 1.1.x.

Issue #4678, opened 2015-10-14
HTML Editor Shortcodes don't work for images


tractorcow:
Because if we embed `<img src="[file_link id=4]" />` the preview area doesn't know how to display the image to the content editor.

jonom:
I think shortcodes have long been a powerful but under-utilised feature because they're not user-friendly - too much like 'coding' for most CMS users. Maybe this is a good opportunity to establish a reusable pattern for rendering shortcode content in the editor and making it editable through a form? I've outlined a possible approach for this here, would love if you could take a look: https://github.com/silverstripe/silverstripe-framework/issues/4337#issuecomment-126009280

sminnee:
That's a fair point, @jonom, however, this issue was more narrowly focused. In particular, I that a form for editing the shortcode would be a bad idea for media insertion—the change requested by this issue should be invisible to the user; it's just a leftover issue from some of the about-to-be-merged filesystem work.

jonom:
Yeah fair enough - wasn't sure how much work would be required to fix this narrow issue, so if substantial work is required just thought it would be worth keeping this idea in mind. I didn't mean to imply that it would have to return a simplistic/generic form either - I was imagining in the case of media that it would behave the same way it does currently, but if we could define this behaviour in a reusable way it might lead to people embracing shortcodes more. Something to consider for the future maybe.

sminnee:
A piece that relates to this ticket: Having a facility for shortcodes to be transformed to and from placeholder content would be nice. We already have some kind of API for this (the oembeds use it), but I'm not sure how easy to use / well documented the APIs for this are.

tractorcow:
Yikes, this is a duplicate issue then. I'll close this in favour of #4337 . Sorry @jonom :D I'll add this as a continuation of that discussion rather than a separate one.

Issue #4337, opened 2015-06-22
Shortcodes for Image references in HTMLText fields


jonom:
Carrying on from a discussion at https://github.com/silverstripe/silverstripe-framework/pull/4328#issuecomment-114239415, there seems to be a long term issue with the CMS where images inserted in to a `HTMLText` field via `HTMLEditorField` are not properly linked back to the image object.

When a HTMText field is saved, any resampled images are generated as necessary and a fully translated URL is used to reference the resampled or original image. This causes a couple of problems:

* If the referenced resampled image is deleted from the filesystem (e.g. using `Image::deleteFormattedImages()`) any links to that image from SiteTree or other objects will be broken until the object containing the link is re-saved, which triggers re-generation of the image.
* If the original Image is renamed in the CMS, any references to that image from a HTMLText field will no longer be able to find the original image so will not be able to be resized or regenerated anymore.

Think a good solution would be to make `img` tags in HTMLText use ShortCodes, just like internal hyperlinks do. This way they will be evaluated whenever the content is displayed and the images can be regenerated if necessary, and since the image will be referenced by ID it won't matter if the image is renamed. This should also protect against any future changes to filename output for generated images.

JorisDebonnet:
I think that's a great (and necessary) idea! And if we can make this happen, I would personally say it would make sense for `flush=all` to remove all resampled images, since they'll be regenerated through the shortcodes anyway.

dhensby:
+1

kinglozzer:
I like this idea (I’m sure it was suggested in a ticket before too, perhaps the asset abstraction one?). The only barrier to overcome is how to load `<img src="[image_src,id=5]" />` when we’re editing in the CMS

JorisDebonnet:
Perhaps it should be converted just before display and just before saving? 

The src attribute needs to hold the shortcode in the database, but when displaying it in the CMS, it could be converted into the evaluated version, while the shortcode could be kept in something like a data-src attribute. Which is then converted back again just before saving.

jonom:
Yeah I think something like what @JorisDebonnet suggested is probably the go and there must already be some manipulations like that taking place to add the interactivity to image and video elements so hopefully we can just tweak that.

It would be really nice to do this in a re-usable way so that *any* shortcode could be rendered properly with in the editor view, and clicking the rendered content could open a modal editing form, like what happens in @sheadawson's [shortcodable module](https://github.com/sheadawson/silverstripe-shortcodable)

[...]

PR #4679, opened 2015-10-14
BUG Fix file link tracking for new asset abstraction

tractorcow:
In the prior pull request at https://github.com/silverstripe/silverstripe-framework/pull/4652 CMS file link tracking was broken.

Merging this should fix the tests for https://github.com/silverstripe/silverstripe-cms/pull/1291.

Link tracking is now managed entirely via data-fileid / shortcodes, as the file url is no longer useful for tracking.

### DIFF
--- a/forms/HtmlEditorField.php
+++ b/forms/HtmlEditorField.php
@@ -74,41 +74,14 @@ class HtmlEditorField extends TextareaField {
 			);
 		}
 
-		$htmlValue = Injector::inst()->create('HTMLValue', $this->value);
+		// Resample images
+		$value = Image::regenerate_html_links($this->value);
+		$htmlValue = Injector::inst()->create('HTMLValue', $value);
 
 		// Sanitise if requested
 		if($this->config()->sanitise_server_side) {
 			$santiser = Injector::inst()->create('HtmlEditorSanitiser', HtmlEditorConfig::get_active());
 			$santiser->sanitise($htmlValue);
-		}
-
-		// Resample images and add default attributes
-		if($images = $htmlValue->getElementsByTagName('img')) foreach($images as $img) {
-			// strip any ?r=n data from the src attribute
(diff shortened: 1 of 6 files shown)


forms/HtmlEditorField.php
```diff
@@ -82,35 +84,6 @@ public function saveInto(DataObjectInterface $record) {
...
-				}
-			}
-
-			// Add default empty title & alt attributes.
-			if(!$img->getAttribute('alt')) $img->setAttribute('alt', '');
-			if(!$img->getAttribute('title')) $img->setAttribute('title', '');
-
-			// Use this extension point to manipulate images inserted using TinyMCE, e.g. add a CSS class, change default title
-			// $image is the image, $img is the DOM model
-			$this->extend('processImage', $image, $img);
```
chillu: Why was this extension point removed?

tractorcow: It's moved to `Image:: regenerate_html_links`

```
// Use this extension point to manipulate images inserted using TinyMCE,
+			// e.g. add a CSS class, change default title
+			self::singleton()
+				->extend('regenerateImageHTML', $imageDO, $imageElement);
```

tractorcow: Added to changelog.

chillu:
This will insert the `date-fileid` property in public HTML at the moment. Assuming that's not desireable, should we cover this by a Github issue to be resolved before 4.0 alpha? It doesn't seem to be covered in https://github.com/silverstripe/silverstripe-framework/issues/4678. Do you think we need to reach a conclusion on https://github.com/silverstripe/silverstripe-framework/issues/4337 to implement this? I want to ensure this doesn't get forgotten about and we end up releasing 4.0 with cluttered HTML output.

tractorcow:
We have a backlog story to cover this. We can make removing it a 4.0 alpha blocker.

tractorcow:
I've raised a separate github issue just for that item. https://github.com/silverstripe/silverstripe-framework/issues/4681

tractorcow:
Thanks @chillu

PR #4680, opened 2015-10-14
API Generated files API

tractorcow:
This feature is rebased against the open asset refactor pull requests:
 - https://github.com/silverstripe/silverstripe-framework/pull/4679
 - https://github.com/silverstripe/silverstripe-cms/pull/1291
 
 This feature replaces the old requirements combination API and introduces a new generated files handler. The Requirements API has been updated quite a bit (better coding conventions, error checking). JS minification has been removed from combined files, since there are already thousands of existing front-end minification tools available that don't rely on server side minification.
 
 This is also used by the ErrorPage handler, which is refactored into a separate error handler in an upcoming pull request.

### DIFF
--- a/docs/en/04_Changelogs/4.0.0.md
+++ b/docs/en/04_Changelogs/4.0.0.md
@@ -19,6 +19,8 @@
    the `DataObject::ID` in a `data-fileid` property, or via shortcodes. This is necessary because file
    urls are no longer able to identify assets.
  * Extension point `HtmlEditorField::processImage` has been removed, and moved to `Image::regenerateImageHTML`
+ * `Requirements::combined_files` no longer does JS minification. Please ensure that JS files are optimised
+   before combining.
 
 ## New API
 
@@ -26,6 +28,8 @@
  * `ShortcodeHandler` interface to help generate standard handlers for HTML shortcodes in the editor.
  * `AssetNameGenerator` interface, including a `DefaultAssetNameGenerator` implementation, which is used to generate
    renaming suggestions based on an original given filename in order to resolve file duplication issues.
+ * `GeneratedAssetHandler` API now used to store and manage generated files (such as those used for error page
+   cache or combined files).
 
 ## Deprecated classes/methods
 
--- a/filesystem/storage/CacheGeneratedAssetHandler.php
+++ b/filesystem/storage/CacheGeneratedAssetHandler.php
@@ -1,0 +1,167 @@
+<?php
+
+namespace SilverStripe\Filesystem\Storage;
+
+use Config;
+use Exception;
+use Flushable;
+use SS_Cache;
+use Zend_Cache_Core;
+
+/**
+ * Handle references to generated files via cached tuples
+ *
+ * Important: If you are using the default FlysystemStore with legacy_filenames, you will need to ?flush
+ * in order to refresh combined files.
+ * 
+ * @package framework
+ * @subpackage filesystem
+ */
+class CacheGeneratedAssetHandler implements GeneratedAssetHandler, Flushable {
+
+	/**
+	 * Lifetime of cache
+	 *
+	 * @config
+	 * @var int
+	 */
+	private static $lifetime = 0;
+
+	/**
+	 * Backend for generated files
+	 *
+	 * @var AssetStore
+	 */
+	protected $assetStore = null;
+
+	/**
+	 * Assign the asset backend
+	 *
+	 * @param AssetStore $store
+	 * @return $this
+	 */
+	public function setAssetStore(AssetStore $store) {
+		$this->assetStore = $store;
+		return $this;
+	}
+
+	/**
+	 * Get the asset backend
+	 *
+	 * @return AssetStore
+	 */
+	public function getAssetStore() {
+		return $this->assetStore;
+	}
+
+	/**
+	 * @return Zend_Cache_Core
+	 */
+	protected static function get_cache() {
+		$cache = SS_Cache::factory('CacheGeneratedAssetHandler');
+		$cache->setLifetime(Config::inst()->get(__CLASS__, 'lifetime'));
+		return $cache;
+	}
+
+	/**
+	 * Flush the cache
+	 */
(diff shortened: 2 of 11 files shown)


tractorcow:
Rebased since the parent bugfix was merged. :)

chillu:
> JS minification has been removed from combined files, since there are already thousands of existing front-end minification tools available that don't rely on server side minification.

This will increase the size of CMS combined files quite a bit. Even if you account for GZIP transfer, it's 367KB rather than 195KB for the three CMS combined files. Without GZIP, its 1500KB rather than 850KB. Better frontend tooling is currently in the "maybe" pile for 4.0, so for now I'd consider removing JSMin a performance regression.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -26,6 +28,8 @@
  * `ShortcodeHandler` interface to help generate standard handlers for HTML shortcodes in the editor.
  * `AssetNameGenerator` interface, including a `DefaultAssetNameGenerator` implementation, which is used to generate
    renaming suggestions based on an original given filename in order to resolve file duplication issues.
+ * `GeneratedAssetHandler` API now used to store and manage generated files (such as those used for error page
```
chillu: Needs to mention all the changed method names for `Requirements`, e.g. `clear_combined_files()` -> `clearCombinedFiles()`

view/Requirements.php
```diff
@@ -1034,269 +1165,275 @@ protected function path_for_file($fileOrUrl) {
...
+	}
 
-				$combinedData .= $fileContent . "\n";
-			}
+	/**
+	 * Check if combined files are enabled
+	 *
+	 * @return bool
+	 */
+	protected function enabledCombinedFiles() {
```
chillu: Why is this worded the other way around? `enabledCombinedFiles` vs `combinedFilesEnabled`

filesystem/storage/CacheGeneratedAssetHandler.php
```diff
@@ -0,0 +1,167 @@
...
+		$cache = SS_Cache::factory('CacheGeneratedAssetHandler');
+		$cache->setLifetime(Config::inst()->get(__CLASS__, 'lifetime'));
+		return $cache;
+	}
+
+	/**
+	 * Flush the cache
+	 */
+	public static function flush() {
+		self::get_cache()->clean();
```
chillu: I'd expect this to delete the underlying files as well, not only the cached references to them

tractorcow: Ah, well, there is no "delete" method in the APL yet. I'll add a "todo" :D

chillu:
How are we handling file deletion here? At the moment, any change to underlying files will create a new file record. Depending on how often you deploy, those can pile up. Example structure after two edits to `LeftAndMain.js`

```
./_combinedfiles/07a900d56b/leftandmain.js
./_combinedfiles/43d7f16f64/cmsmain.js
./_combinedfiles/777d368a44/leftandmain.js
./_combinedfiles/bbb1272b58/lib.js
./_combinedfiles/fcf76a715a/leftandmain.js
``

Do you see this being solved by the [file deletion discussion](https://groups.google.com/forum/#!topic/silverstripe-dev/PihYhP3ZTN8) on the mailinglist as well? Still trying to get my head around the different options there. We could treat the `SS_Cache` entries as "references" to the file?

**Comment by tractorcow** [comment]
As soon as we can agree on a deleted files API then yes we can solve this problem. :)

File: view/Requirements.php (hunk 2 above)
Comment (tractorcow) [inline view/Requirements.php]: Ah, my phpdoc is wrong, that's why. :) There's already `getCombinedFilesEnabled`

Comment (tractorcow) [inline view/Requirements.php]: I had two methods that did the same thing. The other was getCombinedFilesEnabled, but I'll merge them. :D

**Comment by tractorcow** [comment]
@chillu I have pushed up an update:

- Minification is back. I've pushed this into an injectable `Requirements_Minifier`, so `JSMin` isn't hard coded anymore.
- Still no delete, but we'll look at this again in the future.
- Lots and lots of additional upgrading notes in the changelog, especially around the changes to ErrorPage in the CMS pull request.

File: .travis.yml [hunk 4]
```diff
@@ -17,9 +17,9 @@ env:
 
 matrix:
   include:
-    - php: 5.4
+    - php: 5.5
```
chillu: We've made the decision to only support 5.5+ in 4.0 this week. But you probably shouldn't sneak this in this unrelated commit, makes it hard to trace back afterwards.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -26,9 +26,22 @@
...
+   cache or combined files).
+ * `Requirements_Midifier` API can be used to declare any new mechanism for minifying combined required files.
```
chillu: Misspelled `Minifier`

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -74,6 +87,44 @@ The following filesystem synchronisation methods are also removed
 
 The Spyc YAML library has been removed from /thirdparty. Please load it yourself, or use the Symfony YAML component that’s automatically installed by composer.
 
+### Requirements
+
+The following methods and properties on `Requirements_Backend` have been renamed:
+
+ * `Requirements_Backund::$combine_files` made protected and renamed `$combinedFiles`
+ * `Requirements_Backend::$combine_js_with_min` made protected and renamed `$minifyCombinedFiles`
```
chillu: Spelling - should read `combine_js_with_jsmin`

chillu:
I've pushed these changes now - waiting for build to pass before merging

chillu:
Heh OK, so the PR implicitly relies on the PHP 5.5 increment. I still think it's not the right place to commit this alongside other work, so created separate PRs for it: https://github.com/silverstripe/silverstripe-framework/pull/4694 and https://github.com/silverstripe/silverstripe-cms/pull/1303

PR #4682, opened 2015-10-15
BUG Fix regressions in fulltextsearch

tractorcow:
Yet another fix to framework to resolve https://github.com/silverstripe/silverstripe-cms/pull/1291


sminnee:
This means that the we're no longer able to search file content with the internal search code. It's a bit moot, because doing so required that there was another system that populated the Content fields, which usually wasn't implemented, but it should probably be noted in the upgrade guide.

sminnee:
@tractorcow I'm going to merge, but can you amend the upgrade guide to note that the `File.Content` field is no longer available when you get a chance?

PR #4690, opened 2015-10-19
FIX only use sethasemptydefault if exists.

camfindlay:
Backport of #4672 with a slight rewrite incase of subclasses of the field.


camfindlay:
@tractorcow as discussed, PR to fix this bug.

PR #4691, opened 2015-10-20
BUG Fix duplicate files being included in case of flush

### LINKED ISSUE #4553: Flushing causes CMS to not function correctly until reloaded
**Steps to reproduce**

1. Open the CMS
2. In another tab (or browser) do ?flush=1 on any page
3. Try to navigate to another page in the CMS

**Notes**

* The CMS appears to still be XHRing the new page correctly, but is also attempting to load the new version of lib.js from _combinedfiles that was created by the flush.
* Occurs cross-session, a different user could perform a ?flush and it would affect all users of the CMS.

tractorcow:
Fixes #4553 

The premise is if HTTP headers are used to load

/framework/javascript/myfile.js?m=54897
and
/framework/javascript/myfile.js?m=21323

Then only the first file is loaded.

@flashbackzoo could you please review?

### DIFF
--- a/javascript/jquery-ondemand/jquery.ondemand.js
+++ b/javascript/jquery-ondemand/jquery.ondemand.js
@@ -15,34 +15,74 @@
  */
 (function($){
 
+	/**
+	 * Decode URL from header string
+	 * 
+	 * @param {String} str
+	 * @returns {String}
+	 */
 	var decodePath = function(str) {
 		return str.replace(/%2C/g,',').replace(/\&amp;/g, '&').replace(/^\s+|\s+$/g, '');
+	};
+	
+	/**
+	 * Remove querystring from the url. Necessary to prevent a file being included twice in the same page if
+	 * requested at different times (e.g. during flush)
+	 * 
+	 * @param {String} str
+	 * @returns {String}
+	 */
+	var removeQuerystring = function(str) {
+		return str.split("?")[0];
 	};
 
 	$.extend({
 
 		// loaded files list - to protect against loading existed file again  (by PGA)
 		_ondemand_loaded_list : null,
+		
+		/**
+		 * Add a new url to the loaded item
+		 * 
+		 * @param {String} url
+		 */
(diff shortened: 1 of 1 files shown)


flashbackzoo:
Looks good :smile:

One really minor thing. We should use better `undefined` checking (line 76). Can you please change that to...

```javascript
return typeof this._ondemand_loaded_list[removeQuerystring(scriptUrl)] !== 'undefined';
```

tractorcow:
Ok, I've fixed it now. :D

flashbackzoo:
Nice. I'm happy.

ss23:
@tractorcow This is a bug for 3.0 and 3.1 too.
Are we able to get it fixed in 3.1 too?

tractorcow:
3.0 and 3.1 shouldn't really be receiving non-critical bugfixes, as we've now released a stable of the next minor version (3.2.0). This is a relatively minor annoyance.

ss23:
@tractorcow some people have sites that are overzealous in their flushing, which, combined with this bug can make the cms unusable. It ends up being a critical bug.

Also, the fix looks very low risk, which is a mitigating factor.

tractorcow:
If we merged it back, it would likely never end up in a stable release unless we had a security issue we deemed worthy enough to backport and re-issue a new stable fix.

sminnee:
If it's rendering the CMS unusable, then it's a critical bug, and if it's a low-risk change we probably want to fix it 3.1. Worst-case scenario, people can require `3.1.x-dev` if we haven't pushed another stable release.

sminnee:
@ss23 can you describe what you mean by "overzealous in their flushing"? Flush is a user-triggered activity that people can choose not to do—can you just ask them to stop being so overzealous?

tractorcow:
Backported to 3.1 at https://github.com/silverstripe/silverstripe-framework/pull/4720

hafriedlander:
Closing as this should be fixed when 3.1 is up-merged by #4720

PR #4692, opened 2015-10-20
BUGFIX: Use Object->hasMethod() instead of method_exists()

madmatt:
This allows you to for example return a `PaginatedList` object from `ModelAdmin::getList()` without `GridField` failing when trying to render that list in certain circumstances.

I can't see a downside to this, a slight performance hit as it now checks for extension methods as well as the standard `method_exists()`, but it's only executed once per rendered `GridField`.

Wasn't sure if I should raise this against 3.1 or 3.2?

### DIFF
--- a/forms/gridfield/GridField.php
+++ b/forms/gridfield/GridField.php
@@ -148,7 +148,7 @@ class GridField extends FormField {
 			return $this->modelClassName;
 		}
 
-		if($this->list && method_exists($this->list, 'dataClass')) {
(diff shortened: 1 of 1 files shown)


tractorcow:
3.2 please. :D

dhensby:
This never made it in to core, which is a shame - @tractorcow you strict man!

@madmatt do you want to get it in to 3.2? :)

tractorcow:
This was merged with https://github.com/silverstripe/silverstripe-framework/pull/4728

PR #4693, opened 2015-10-21
DOCS Upgrading information about 3.2.0 and the additional modules req…

camfindlay:
…uired (reports & siteconfig)

Based on a community question about upgrading their site manually, they followed the instructions but were not aware of the additional modules now required.

### DIFF
--- a/docs/en/03_Upgrading/index.md
+++ b/docs/en/03_Upgrading/index.md
@@ -1,22 +1,25 @@
 title: Upgrading
 introduction: Keep your SilverStripe installations up to date with the latest fixes, security patches and new features.
 
-# Upgrading
+# Upgrading to SilverStripe 3.2
 
-SilverStripe Applications should be kept up to date with the latest security releases. Usually an update or upgrade to 
-your SilverStripe installation means overwriting files, flushing the cache and updating your database-schema. 
+SilverStripe applications should be kept up to date with the latest security releases. Usually an update or upgrade to your SilverStripe installation means overwriting files, flushing the cache and updating your database-schema. 
 
 <div class="info" markdown="1">
-See our [upgrade notes and changelogs](/changelogs) for release-specific information.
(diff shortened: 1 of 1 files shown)


docs/en/03_Upgrading/index.md
```diff
@@ -1,22 +1,25 @@
 title: Upgrading
 introduction: Keep your SilverStripe installations up to date with the latest fixes, security patches and new features.
 
-# Upgrading
+# Upgrading to SilverStripe 3.2
```
tractorcow: This section is going to get stale very fast. It needs to be generic, otherwise we'll have to constantly update this section each release.

camfindlay:
Stale vs useful - I'd opt for useful... it's already tripped someone up. This should only apply in the 3.2 docs.

docs/en/03_Upgrading/index.md
camfindlay: what if we keep it to 3.2.x only (no change for patch versions?)

docs/en/03_Upgrading/index.md
```diff
@@ -1,22 +1,25 @@
...
 
-For projects managed through Composer, check the version defined in your `composer.json` file. Update the version 
-constraints if required and update composer.
+For projects managed through Composer, update the version number of `framework` and `cms` to `^3.2` in your `composer.json` file and run `composer update`. 
 
-	:::bash
-	composer update
+```json
+	"require": {
+		"silverstripe/framework": "^3.2",
```
tractorcow: This is risky in that it would automatically update your project to 3.3 on composer update. Did you mean ~3.2.0?

tractorcow:
Then I'll just have to critique something else. :)

If you fix the above I can merge. Otherwise, tell me it's better the way it is, and I can merge it too.

docs/en/03_Upgrading/index.md
camfindlay: Isn't that the point when looking at an upgrading doc?

tractorcow:
Fair enough!

PR #4694, opened 2015-10-22
Increase minimum required PHP version to 5.5

chillu:
See mailinglist discussion at https://groups.google.com/forum/#!topic/silverstripe-dev/7mVvhNNeNps.
Please merge together with https://github.com/silverstripe/silverstripe-cms/pull/1303.
I haven't tested that the IIS setup still works with PHP 5.5, but I'm sure it doesn't with the PHP 5.3 and PHP 5.4 versions mentioned before (at least once https://github.com/open-sausages/silverstripe-framework/commits/features/dbfile-uploadfield is merged, which uses PHP 5.5 only syntax).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,9 @@ env:
 
 matrix:
   include:
-    - php: 5.4
+    - php: 5.5
       env: DB=MYSQL
-    - php: 5.4
+    - php: 5.5
       env: DB=PGSQL
     - php: 5.5
       env: DB=SQLITE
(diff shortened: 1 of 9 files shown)


kinglozzer:
Do we also need to update [install.php](https://github.com/silverstripe/silverstripe-framework/blob/master/dev/install/install.php) and i[nstall.php5](https://github.com/silverstripe/silverstripe-framework/blob/master/dev/install/install.php5)?

chillu:
@kinglozzer Good point, done!

kinglozzer:
It gives me great pleasure to click the green button on this :wink:

PR #4697, opened 2015-10-22
New AssetField for uploading to a DBFile database field

chillu:


forms/FormField.php
```diff
@@ -463,6 +463,14 @@ public function extraClass() {
 			$classes[] .= 'holder-' . $this->MessageType();
 		}
 
+		if($this->isDisabled()) {
```
chillu: Given the generic naming here, this could throw off a lot of custom CSS/JS implementations. I'd prefer to keep it on `UploadField` and `AssetField` for now, and discuss the implications of moving this to `FormField` through a separate PR.

chillu:
Peer reviewed this request and looks great (minus the one niggle mentioned above). Still requires tests which Damian is writing at the moment.

tractorcow:
Tests done!

This has been rebased against a current outstanding PR. Please merge https://github.com/silverstripe/silverstripe-framework/pull/4699 first if possible.

tractorcow:
@chillu raised this PR but it's actually my code.

I give him permission to merge it. :D

view/Requirements.php
```diff
@@ -1392,9 +1392,6 @@ public function processCombinedFiles() {
 					break;
 				}
 			}
-			if($combinedURL && !$included) {
```
tractorcow: I've removed this, because there is a valid case where a combinedFile should be blocked, which is if all of it's source files have been cleared with `->clear`

chillu:
> Given the generic naming here, this could throw off a lot of custom CSS/JS implementations. I'd prefer to keep it on UploadField and AssetField for now, and discuss the implications of moving this to FormField through a separate PR.

This comment for some reason only shows as on an outdated diff, even if the code there hasn't changed. Which is why I missed that comment and thought Damian had fixed already - my mistake. I'm proposing to revert this aspect, and move `extraClasses()` back to `UploadField` and `AssetField` to minimise side effects.

PR #4711, opened 2015-10-27
Addition of extend for newRowClasses

marijnkampf:
Allows for creation of extension of row classes in gridfield.

### DIFF
--- a/forms/gridfield/GridField.php
+++ b/forms/gridfield/GridField.php
@@ -583,7 +583,7 @@ class GridField extends FormField {
 			$classes[] = 'odd';
 		}
 		
-		$this->owner->extend('getRowClasses', $classes, $total, $index, $record);
+		$this->extend('updateNewRowClasses', $classes, $total, $index, $record);
 
 		return $classes;
 	}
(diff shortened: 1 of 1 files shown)


JZubero:
+1 I was thinking about exactly the same!

forms/gridfield/GridField.php
```diff
@@ -582,6 +582,8 @@ protected function newRowClasses($total, $index, $record) {
 		} else {
 			$classes[] = 'odd';
 		}
+		
+		$this->owner->extend('getRowClasses', $classes, $total, $index, $record);
```
tractorcow: This isn't an extension, and doesn't have a `$this->owner`. Try `$this->extend`

dhensby:
Ping

tractorcow:
I guess someone needs to re-create this themselves...

forms/gridfield/GridField.php
dhensby: Also, `getRowClasses` seems wrong - `updateNewRowClasses` seems more inline with convention...

tractorcow:
Can you squash please?

dhensby:
we can do it our selves now! yay

marijnkampf:
Yeah from me too! I hadn't made the time yet to re-look up how to squash.

Issue #4712, opened 2015-10-27
Minor: Image->generateFormattedImage() and ->generateResizedImage() should document their possible null return value.


patricknelson:
Currently the methods `->generateFormattedImage()` and `->generateResizedImage()` on the `Image` object and it's derivatives (e.g. `->PaddedImage()`) don't indicate that they could potentially return `null` which has resulted in some unexpected results as I've built wrappers around this API. It's somewhat dangerous since obviously calling another method on a null results in a fatal error.

Since this is a documentation only change (like adding `@return Image|null`), I think a PR against `3.2` should be sufficient.

jonom:
I guess the other option would be to return an empty Image object instead of null, so method chaining is safe. When do you get a null result?

patricknelson:
If the database goes out of sync with the file system. That is, if for some reason one of my servers ends up with an issue with synchronization and an image file goes missing, I end up with a catastrophic page crash which would potentially be cached by upstream servers (thankfully it doesn't cache 500 errors).

See the `$this->exists()` check here: https://github.com/silverstripe/silverstripe-framework/blob/3.2/model/Image.php#L675

patricknelson:
By the way, this is because I've setup an extension to the base `Image` class (for CDN cache busting) with a method like `->setURLSuffix()` so that the native `->URL()` method can be used. Anyway, needless to say I can't call the extension method on a null object, which is fine since I should be doing the check anyway (which I'm usually pretty good about). I typically infer this by quickly clicking/referencing the original method documentation to see what the return type to see if there's a chance it'd not return an object or not return the expected type.

In this case it's just an extra step in helping ensure the developer does the proper checks by documenting the possible return types.

patricknelson:
See: https://github.com/silverstripe/silverstripe-framework/pull/4713

jonom:
Documenting the current functionality is fair, but I worry about changing the expectations of the API. If it's returning null something has gone wrong right? That feels like `user_error` territory to me. I think that according to SemVer, we should instead patch the functionality to make good on what the API has promised, rather than change the API promise to line up with the functionality.

I'm probably in over my head here though, @silverstripe/core-team thoughts?

patricknelson:
[...]

PR #4713, opened 2015-10-27
FIX for #4712: Dropping in some PHP documentation on return types for dynamically generated image methods.

patricknelson:
Quick fix to address #4712. Documentation update only.

### DIFF
--- a/model/Image.php
+++ b/model/Image.php
@@ -208,7 +208,7 @@ class Image extends File implements Flushable {
 	 *
 	 * @param integer $width The width to size within
 	 * @param integer $height The height to size within
-	 * @return Image
+	 * @return Image|null
 	 */
 	public function Fit($width, $height) {
 		// Prevent divide by zero on missing/blank file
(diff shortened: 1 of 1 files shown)


model/Image.php
```diff
@@ -564,6 +564,11 @@ public function generateSetSize(Image_Backend $backend, $width, $height) {
 		return $backend->paddedResize($width, $height);
 	}
 
+	/**
+	 * TODO: Documentation.
```
tractorcow: Seems like this is the opportunity to write that documentation? :D

You could just copy the docs from the method below `Resize this image for the CMS. Use in templates with $CMSThumbnail.`

patricknelson: I'm sorry -- I didn't have the time to research it but I did want to point it out. I'll paste that in; otherwise a bit too busy, I apologize :(

patricknelson: See https://github.com/silverstripe/silverstripe-framework/pull/4713/files#diff-bf03e641675b755b14906a961cf8485fR568

PR #4715, opened 2015-10-28
BUG Fix broken InlineFormAction

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-translatable/issues/198

This was actually a semi-major regression in 3.2.0 since all inline actions broke.


dhensby:
I think your test is a bit too tightly coupled to the output. A valid tag could be returned with the order of params changed and the test would fail. Can't we assert the things we need a bit more loosely?

tractorcow:
Yeah that's fair enough. I'll loosen it up. :D

chillu:
Replaced by https://github.com/silverstripe/silverstripe-framework/pull/4717

PR #4717, opened 2015-10-28
BUG Fix broken InlineFormAction

chillu:
Replaces https://github.com/silverstripe/silverstripe-framework/pull/4715, with more granular tests


chillu:
@tractorcow Happy with that? I started with `SS_HTML4Value` but that seemed a bit overkill - plus, the `DOMDocument` API is a pain to work with

PR #4396, opened 2015-07-14
FIX Files should only report as "existing" if the file actually exists

dhensby:
see https://github.com/silverstripe/silverstripe-framework/pull/4060

PR #4719, opened 2015-10-29
BUG Fix regression from #4396 in test fixtures

tractorcow:
During test initialisation, if a Folder is created it doesn't "exist" until the physical filesystem folder is created.

This means that if a File fixture has a folder as a parent, it will fail to attach during creation, and will end up being set to the root directory.

E.g. this fixture:

```yaml
Folder:
  folder1:
    Name: UploadFieldTest
File:
  file1:
    Title: File1
    Filename: assets/UploadFieldTest/file1.txt
    Parent: =>Folder.folder1
```

Will end up with a file in assets/file1.txt

The only way to ensure a created directory is valid is to create the underlying filesystem directory on the fly, immediately after saving it to the database (but before being assigned children).

### DIFF
--- a/dev/FixtureBlueprint.php
+++ b/dev/FixtureBlueprint.php
@@ -192,6 +192,11 @@ class FixtureBlueprint {
 			if($data && array_key_exists('LastEdited', $data)) {
 				$this->overrideField($obj, 'LastEdited', $data['LastEdited'], $fixtures);
 			}
+
+			// Ensure Folder objects exist physically, as otherwise future File fixtures can't detect them
+			if($obj instanceof Folder) {
+				Filesystem::makeFolder($obj->getFullPath());
+			}
 		} catch(Exception $e) {
 			Config::inst()->update('DataObject', 'validation_enabled', $validationenabled);
 			throw $e;


gregsmirnov:
@tractorcow you need to respect `File::update_filesystem = false` since some tests might be interested in the database data only.

tractorcow:
Possibly, but that would still cause regressions when a folder doesn't exist. Folder::exists would have to return true if the update_filesystem is false, otherwise the ORM would still fail.

Perhaps folder::exists could negate the "folder must exist" check that was added to File::exists?

@dhensby do you have a view on this?

dhensby:
What's the purpose of `update_filesystem`? is it for testing purposes only or is it for some other case when you'd only want DB updates?

I don't see a realistic use case where files should be in the DB but not on the file system. What good is a File record that points to a non-existing file path?

tractorcow:
I think it's not so much to avoid creation of the file, but rather to manually create it later (in case custom file content needs to be written after fixture setup).

It might also be invoked outside of testing (maybe secureassets does this?)

gregsmirnov:
1. We use this feature in testing, for example to see how Filename is used when CDN is enabled or watermark generator is applied. We don't really need real files for this.

2. We use this feature in data migration from other platforms (CSMs). The database is updated first, then files are copied either with rsync or dynamically with the script.

> I don't see a realistic use case where files should be in the DB but not on the file system.

This is completely valid statement for the production system. :-)

dhensby:
Ok - so the sensible thing here (as suggested) seems to be that if `update_filesystem` is false, then we don't check the file exists on the filesystem and instead, just that it exists in the DB.

Doing a `isSystemUnderTest` check seems too tightly coupled to testing and ignores any edge cases that there might be under production

tractorcow:
In 4.0 folders won't be hard-linked to the local filesystem anymore. :)

dhensby:
indeed, we can have a test backend for files that just performs noops

tractorcow:
Actually, what I've done instead is setup temporary folders which are assigned fixture-specific asset backends. This also helps test that the backend is actually being relied on properly without simply assuming that files are stored in /assets. :)

PR #4720, opened 2015-10-29
BUG Fix duplicate files being included in case of flush

### LINKED ISSUE #4553: Flushing causes CMS to not function correctly until reloaded

tractorcow:
Fixes #4553 for 3.1

See https://github.com/silverstripe/silverstripe-framework/pull/4691#issuecomment-152001295 for 3.2 PR.

### DIFF
--- a/javascript/jquery-ondemand/jquery.ondemand.js
+++ b/javascript/jquery-ondemand/jquery.ondemand.js
@@ -15,34 +15,74 @@
  */
 (function($){
 
+	/**
+	 * Decode URL from header string
+	 * 
+	 * @param {String} str
+	 * @returns {String}
+	 */
 	var decodePath = function(str) {
 		return str.replace(/%2C/g,',').replace(/\&amp;/g, '&').replace(/^\s+|\s+$/g, '');
+	};
+	
+	/**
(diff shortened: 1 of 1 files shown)


hafriedlander:
I un-merged this. Problems with it:

- Breaks require calls where there are multiple requires called and the URI differs only by GET parameter (?language=en, ?language=jp)

- Doesn't actually fix the problem, just replaces bad bug with slightly less bad bug. Simply saying "I've already included _a_ version of whatever.js, so I'm not going to include _this_ version of whatever.js" won't fix dependancies on new features in this newly requested version by other required js.

hafriedlander:
The proper fix is that the `m` parameter shouldn't change unless the underlying javascript file has changed (so flush shouldn't do anything unless there's an actual code change) - it could be SHA rather than timestamp if that helped. And, if two different versions of the same file are requested, force a full-page refresh.

Issue #4710, opened 2015-10-27
"Please enable pop-ups for this site"


jacobbuck:
![screen-shot-2015-10-27-at-5 06 30-pm](https://cloud.githubusercontent.com/assets/830470/10749354/3852d75c-7cce-11e5-84fc-638893bd6f32.png)

Is this still necessary in 2015?

Also I get this whenever I have the cms open in a tab while the web server goes down.

sminnee:
The underlying issue is that we use a pop-up to let the user re-login without losing their data. I guess it could be re-coded it as something ajax-based, if you want to do that.

patricknelson:
I will say that the alert box annoys the hell out of me, even though I've enabled pop-ups. It would be best if this were implemented some other way, since the alert steals focus from your current tab/work and forces you to close it first before moving on again. In my case this happens essentially every day when I close the lid on my laptop and open back up again (since it goes to sleep). To avoid this it would make more sense to use some sort of modal, toaster pop-up or something inline within the page instead of an old school 90's style alert box.

patricknelson:
Also I just realized we all have sunglasses in our profile photos.

jonom:
@patricknelson Haha!

jonom:
Aw, I spoiled the effect by joining the party. :sunglasses: @jacobbuck @patricknelson do one of you guys want to have a stab at changing this behaviour? Otherwise this issue will probably need to be closed as it's a feature request rather than a bug.

patricknelson:
Not saying I'd do it, but for reference: Would this go in `master` or `3`? My money is on `3` even though I don't think SS will ever release a `3.3`.

jonom:
I would put it in 3 too - I'm pretty hopeful we will see a 3.3 release. Releasing 3.2 took so long because so much had to be reverted to make it semver compliant - 3.3 should be far easier. The goals for 4.0 are pretty ambitious so I wouldn't be surprised if it is a ways off still and we have a 3.3 release in the mean time (who knows - maybe even a 3.4!).

sminnee:
3 branch is fine for this change. We'll do a 3.3 release but it will be "low fuss"—we want have a bunch of dedicated product dev time going into 3.3, instead it will just be "here's all the small things that have been added". Beginning of next year, I'd expect.

patricknelson:
Ok, see here: https://github.com/silverstripe/silverstripe-framework/pull/4722

[...]

PR #4722, opened 2015-10-29
FIX for #4710: Use INTERNAL dialogs instead of old alert() popups on server connection errors during ping.

patricknelson:
Fix to address #4710. Also adding login prompt, in case server connectivity issues last long enough to timeout session due to failed pings.

**Goals:**
- Use internal dialog instead of `alert()`, mainly to prevent window from stealing focus and for a less interrupted experience.
- Provide the opportunity for the user to log back in via more fail-safe method on server access errors, i.e. new window instead in case server is down and `<iframe>` is not able to load inline within page.
- Close the internal message dialog if everything is back to normal (at least in terms of server response).
- In case user ends up locked out of server long enough (due to downtime) and then their **session** times out, provide opportunity once server is back online to log back into session.

**What I had to change**
- Ping code, added more in-depth and specific checking on server response with more intelligent handling. Preventing display of multiple message over and over by checking for existing messages first.
- Minor tweak to `ssui.ssdialog` in `ssui.core.js` to not require an `<iframe>` by simply checking to see if a URL was passed. That way you can still toggle/style modal dialogs without having to load a URL. This is important here, given the server could be down!
- Performing an actual security check on `ping` (in the `Security` controller, no less) to facilitate more specific response checking client-side, which I think makes sense here.

**What is still needed**
- Fix the entwine namespacing issues (see below).
- Abstract that simple message code that I wrote to something reusable, maybe. I can see this being useful elsewhere; after modifying `ssui.ssdialog` in `ssui.core.js`, you no longer MUST have an `<iframe>` in order to display a nicely formatted SilverStripe dialog that just contains arbitrary HTML. I don't believe that you should have to create a tag; I think you should be able to call some sort of function (namespaced itself) and just give it the HTML or message you want to render. e.g. `$.ss.simpleModal("My message here.")`.

---

**IMPORTANT (unrelated to original PR):** Originally I was able to work with calling the following in older `3.1` code, but ended up having to perform a hack/workaround to deal with a potential bug (or change?) in entwine namespacing. That is, calling this would work from within the `ss.ping` entwine namespace:

``` js
$('.cms-container').entwine('ss').showLoginDialog();
```

Now, I have to resort to something more like this in order to steal from the _working_ scope of `ss` on its own, since this no longer works inside of the `ss.ping` scope:

``` js
var containerScope;
$.entwine('ss', function($){
    containerScope = $('.cms-container');
    window.containerScope = containerScope;
});

$.entwine('ss.ping', function($){
    $('.cms-container').entwine({
        _setupPinging: function () {
            containerScope.showLoginDialog();
        }
    });
});
```

### DIFF
--- a/admin/javascript/LeftAndMain.Ping.js
+++ b/admin/javascript/LeftAndMain.Ping.js
@@ -2,6 +2,13 @@
  * File: LeftAndMain.Ping.js
  */
 (function($) {
+	// TODO: Coping with an apparent bug in entwine namespacing! Normally $('.cms-container').entwine('ss') would still work even from within the 'ss.ping' namespace.
+	// TODO: Best solution for now at least may be to tweak the 'ss' namespace to just 'ss'.
+	var containerScope;
+	$.entwine('ss', function($){
+		containerScope = $('.cms-container');
+	});
+
 	$.entwine('ss.ping', function($){
 
 		$('.cms-container').entwine(/** @lends ss.Form_EditForm */{
@@ -21,16 +28,67 @@
 			 *
 			 * This function is called by prototype when it receives notification that the user was logged out.
 			 * It uses /Security/ping for this purpose, which should return '1' if a valid user session exists.
-			 * It redirects back to the login form if the URL is either unreachable, or returns '0'.
+			 * It opens the login form in a new window if the URL is either unreachable (doesn't return HTTP 200) or doesn't return the expected '1'.
 			 */
 			_setupPinging: function() {
+				var loginWindow;	// Reference to pop-up window generated if/when ping fails.
+				var dialog;			// Reference to message dialog, if one is ever opened.
+
 				var onSessionLost = function(xmlhttp, status) {
-					if(xmlhttp.status > 400 || xmlhttp.responseText == 0) {
-						// TODO will pile up additional alerts when left unattended
-						if(window.open('Security/login')) {
-							alert('Please log in and then try again');
-						} else {
-							alert('Please enable pop-ups for this site');
+					// See if we're getting a response that contains a request to re-authenticate. If so, do that now.
+					if (xmlhttp.getResponseHeader('X-Reauthenticate')) {
+						//TODO: Change to the following line when entwine bug is found/fixed.
+						// $('.cms-container').entwine('ss').showLoginDialog();
+						containerScope.showLoginDialog();
+
+					} else if(xmlhttp.responseText !== "1") {
+						// In this case, the server may not have even responded at all. To be safe, open a new window and
+						// show a default message requesting user to login, in case the login window is already open or
+						// opens successfully. This is necessary so we don't attempt to load a login iframe to a down
+						// server (better to do that in a new window instead).
+						var message = 'Please log in and then try again.';
+
+						if (!loginWindow || (loginWindow && loginWindow.closed)) {
+							// Window hasn't yet been opened (or has been closed but must be reopened again).
+							loginWindow = window.open('Security/login');
+							if(!loginWindow) message = 'Please enable pop-ups for this site.';
+						}
+
+						// Render message dialog now.
+						// TODO: This generic dialog code should be abstracted so it can be reused throughout the system in place of 90's style alert() dialogs.
+						dialog = $("#message-dialog");
+						window.testDialog = dialog;
+						if (dialog.length == 0) {
+							dialog = $('<div id="message-dialog" align="center"></div>');
+							$('body').append(dialog);
+						}
+						dialog.html('<h4>' + message + '</h4>');
+
+						dialog.ssdialog({
+							autoOpen: true,
+							minWidth: 450,
+							maxWidth: 450,
+							minHeight: 0,
+							maxHeight: 0,
+							closeOnEscape: true
+						});
+
+					} else {
+						if (loginWindow || dialog) {
+							// Close login window and warning dialog now that they are no longer necessary.
+							if (loginWindow) loginWindow.close();
+							if (dialog) dialog.remove();
+							loginWindow = dialog = null;
+
+							// However, we still want to confirm that the user still has a valid session (in case it timed
+							// out server-side). If this fails, we'll detect the presence of X-Reauthenticate and trigger
+							// the display of the standard login prompt within the page.
+							$.ajax({
+								url: 'Security/ping',
+								global: false,
+								type: 'POST',
+								complete: onSessionLost
+							});
 						}
 					}
 				};
--- a/security/Security.php
+++ b/security/Security.php
@@ -376,7 +376,8 @@ class Security extends Controller implements TemplateGlobalProvider {
 	 * sessions don't timeout. A common use is in the admin.
 	 */
 	public function ping() {
-		return 1;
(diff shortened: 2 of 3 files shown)


sminnee:
We'll want to make sure we test this with a server down, pop-ups enabled, pop-ups disabled, etc, to confirm that it's solid, but the approach looks good.

Is there any way that we can test this with Behat? E.g. could we load the CMS, visit Security/logout in another window, pause, and then confirm the presence of the log-in-again screen? @chillu do you have any idea of the feasibility of that?

patricknelson:
Oops, left the `window` debug variable in there, I'll have to take that out.

admin/javascript/LeftAndMain.Ping.js
```diff
@@ -2,6 +2,14 @@
  * File: LeftAndMain.Ping.js
  */
 (function($) {
+	// TODO: Coping with an apparent bug in entwine namespacing! Normally $('.cms-container').entwine('ss') would still work even from within the 'ss.ping' namespace.
+	// TODO: Best solution for now at least may be to tweak the 'ss' namespace to just 'ss'.
+	var containerScope;
+	$.entwine('ss', function($){
+		containerScope = $('.cms-container');
+		window.containerScope = containerScope;
```
patricknelson: Remove.

admin/javascript/LeftAndMain.Ping.js
```diff
@@ -21,16 +29,68 @@
 			 *
 			 * This function is called by prototype when it receives notification that the user was logged out.
 			 * It uses /Security/ping for this purpose, which should return '1' if a valid user session exists.
-			 * It redirects back to the login form if the URL is either unreachable, or returns '0'.
+			 * It opens the login form in a new window if the URL is either unreachable (doesn't return HTTP 200) or doesn't return the expected '1'.
 			 */
 			_setupPinging: function() {
+				var loginWindow;	// Reference to pop-up window generated if/when ping fails.
+				var dialog;			// Reference to message dialog, if one is ever opened.
+				this._super();
```
patricknelson: Also not need.

dhensby:
Will this get any love?

dhensby:
Rebase, please.

admin/javascript/LeftAndMain.Ping.js
```diff
@@ -21,16 +28,67 @@
...
+					} else if(xmlhttp.responseText !== "1") {
+						// In this case, the server may not have even responded at all. To be safe, open a new window and
+						// show a default message requesting user to login, in case the login window is already open or
+						// opens successfully. This is necessary so we don't attempt to load a login iframe to a down
+						// server (better to do that in a new window instead).
+						var message = 'Please log in and then try again.';
+
+						if (!loginWindow || (loginWindow && loginWindow.closed)) {
+							// Window hasn't yet been opened (or has been closed but must be reopened again).
+							loginWindow = window.open('Security/login');
```
tractorcow: We should be using the in-cms reauthentication here, not the default Security/login in another window.

patricknelson: Good point, this was just an iterative improvement upon what was already setup. In fact, it took me a while to realize this (haven't touched this code in *months*), however I believe that's what this is doing right now. See `containerScope.showLoginDialog();` above. If you read the comments just below that, it only attempts to open in a new tab/window as a last resort in case the server is not responding properly. That should hopefully make it less likely for someone to be locked out of any work that they may currently have in the window which is being obscured by the login dialog and, of course, refreshing could cause loss of work.

dhensby:
rebase please

patricknelson:
Ok @dhensby, rebased and squeaky clean. Let me know what you think @tractorcow about this:
https://github.com/silverstripe/silverstripe-framework/pull/4722#discussion_r57829168

Discussion still continues however on #4710, so I don't really feel like this is ready for merge per se. Or if it's even usable (would need more time to look into it later when I can focus more on this).

dhensby:
@tractorcow :)

security/Security.php
```diff
@@ -376,7 +376,8 @@ public function Link($action = null) {
 	 * sessions don't timeout. A common use is in the admin.
 	 */
 	public function ping() {
-		return 1;
+		if (Member::currentUserID()) return 1;
+		return Security::permissionFailure($this);
```
tractorcow: Ok, there's a small problem with this... in that ping COULD be used outside of the CMS, such as in frontend forms, and this method throwing errors could cause breakages.

I suggest this change; You can still check if the user isn't logged in, but don't make it an error condition.

```php
public function ping() {
	$response = new SS_HTTPResponse("1");
		if(!Member::currentUserID() && Director::is_ajax() && CMSSecurity::enabled()) {
			$this->getResponse()->addHeader('X-Reauthenticate', '1');
		}
		return $response;
	}
```

admin/javascript/LeftAndMain.Ping.js
```diff
@@ -21,16 +28,67 @@
...
+							if(!loginWindow) message = 'Please enable pop-ups for this site.';
+						}
+
+						// Render message dialog now.
+						// TODO: This generic dialog code should be abstracted so it can be reused throughout the system in place of 90's style alert() dialogs.
+						dialog = $("#message-dialog");
+						window.testDialog = dialog;
```
tractorcow: What is `testDialog`? Is this testing code you left in? :P

admin/javascript/LeftAndMain.Ping.js
```diff
@@ -21,16 +28,67 @@
...
+						//TODO: Change to the following line when entwine bug is found/fixed.
+						// $('.cms-container').entwine('ss').showLoginDialog();
+						containerScope.showLoginDialog();
+
+					} else if(xmlhttp.responseText !== "1") {
+						// In this case, the server may not have even responded at all. To be safe, open a new window and
+						// show a default message requesting user to login, in case the login window is already open or
+						// opens successfully. This is necessary so we don't attempt to load a login iframe to a down
+						// server (better to do that in a new window instead).
+						var message = 'Please log in and then try again.';
```
tractorcow: We should i18n these.

admin/javascript/LeftAndMain.Ping.js
```diff
@@ -21,16 +28,67 @@
...
+						dialog.ssdialog({
+							autoOpen: true,
+							minWidth: 450,
+							maxWidth: 450,
+							minHeight: 0,
+							maxHeight: 0,
+							closeOnEscape: true
+						});
+
+					} else {
```
tractorcow: Can you chain these instead of the extra nesting please?

admin/javascript/LeftAndMain.Ping.js
patricknelson: Hah, yes it is, nice catch.

dhensby:
bump

dhensby:
As @patricknelson has made somewhat clear before, he won't have time to work on these PRs if changes are needed - can any core members take it on to get it over the line?

maxime-rainville:

maxime-rainville:
We now have a dedicated modal for this. So there's no point rebasing this on SilverStripe 4.

image

PR #4725, opened 2015-10-30
BUG Ensure that filters on any fixed field are scoped to the base data table

### LINKED ISSUE #4700: Default values of custom fields added to existing records not populated.
When introducing new $db properties to Image, default values will not be applied to existing records if the record only exists in the File table and not the Image table. Example: 'TestImageProperty' => 'Boolean(true)'

----

WAS:

I'm just upgrading my boilerplate project and noticed I was having trouble with Images. A bunch of them seemed to be behaving only as Files all of a sudden. I checked the database and while there were lots of records in the File table with a ClassName of Image, there were only a handful of records in the Image table.

There should be a record in the Image table for each ClassName=Image record in the File table right?

To test I deleted the Image table by removing the extension that's adding properties to Image, and rebuilt the database. Then I created a simple extension to add a $db property to Image, and rebuilt again. An Image table is built but it is empty. Image::get() comes up empty.

Anyone know what's going on here? I'm going to be away for a few days but here are some test files if anyone wants to investigate: http://cl.ly/402J242f3D10/ImageTableTest.zip

tractorcow:
Fixes #4700

### DIFF
--- a/core/ClassInfo.php
+++ b/core/ClassInfo.php
@@ -307,16 +307,20 @@ class ClassInfo {
 
 		//normalise class name
 		$candidateClass = self::class_name($candidateClass);
-
 		$exists = self::exists($candidateClass);
 
-		while($candidateClass && $candidateClass != 'DataObject' && $exists) {
-			if(DataObject::has_own_table($candidateClass)) {
-				$inst = singleton($candidateClass);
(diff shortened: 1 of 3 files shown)


tractorcow:
ta! @sminnee

PR #4727, opened 2015-10-31
Don't try to unbind events from inside top-level iframes

edlinklater:
Some JavaScript APIs (notably all Google APIs) rely on inserting a hidden iframe within the body.

`LeftAndMain.ActionTabSet.js` has functionality to find all iframes and unbind all JS events within the interior pages. This ends up trying to access the contents of an iframe on Google's domain resulting in CORS errors and halting further JavaScript execution in the Admin.

There should perhaps be a wider discussion here around trying to check whether an iframe is internal or external before attempting to access the contents, but in the meantime this will unblock iframes that are direct children of `body`/`.cms` which will cover most iframes that are inserted by third-party JavaScript.

As far as I can tell the iframes we are trying to target for this unbinding operation will always be within `.cms-container`.

### DIFF
--- a/admin/javascript/LeftAndMain.ActionTabSet.js
+++ b/admin/javascript/LeftAndMain.ActionTabSet.js
@@ -28,7 +28,7 @@
 				// Remove all bound events.
 				// This guards against an edge case where the click handlers are not unbound
 				// because the panel is still open when the ajax edit form reloads.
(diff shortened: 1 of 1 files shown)


chillu:
Fair enough. `.cms-container` is more specific as the inner container, and as a selector also part of the "CMS Entwin API" (has `loadPanel()` attached to it).

PR #4728, opened 2015-10-31
BUGFIX: Use Object->hasMethod() instead of method_exists()

madmatt:
BUGFIX: Use `Object->hasMethod()` instead of `method_exists()`

This fixes an issue when you pass a `PaginatedList` back from `ModelAdmin::getList()`. The original issue can be replicated with the following:
```php
<?php
class AdminTest extends ModelAdmin {
    private static $managed_models = array('SiteTree');
    private static $url_segment = 'test';
    private static $menu_title = 'Test';

    public function getList() {
        return new PaginatedList(parent::getList(), $this->request);
}
```

### DIFF
--- a/forms/gridfield/GridField.php
+++ b/forms/gridfield/GridField.php
@@ -149,7 +149,7 @@ class GridField extends FormField {
 			return $this->modelClassName;
 		}
 
-		if($this->list && method_exists($this->list, 'dataClass')) {
+		if($this->list && $this->list->hasMethod('dataClass')) {
 			$class = $this->list->dataClass();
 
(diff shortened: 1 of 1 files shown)


madmatt:
Note: The reason I'm passing a `PaginatedList` back from `ModelAdmin::getList()` is for performance reasons.

With 1.4m records, it takes ~14 seconds to render the standard `ModelAdmin` interface without `getList()` being overridden (even with GridFieldPaginator), and takes ~4 seconds after this patch. Not really sure why as they both limit the list, just anecdotal evidence :)

PR #4729, opened 2015-11-01
BUGFIX: If ->loadDataFrom is given an array, allow multiples for namespaced keys

madmatt:
When `Form->loadDataFrom()` is passed an array for `$data` (e.g. what selecting
multiple values for a `CheckboxsetField` would do), this allows those values to
be correctly set by `loadDataFrom`.

I can probably provide a reproducible test case if required, for me it
manifested in a `GridField` inside a `ModelAdmin` `GridField` on a
`CheckboxsetField` which has a field name something like `Schedule[Daily][Days]`,
which means e.g. the first value would be `Schedule[Daily][Days][1] = 1`.

### DIFF
--- a/forms/Form.php
+++ b/forms/Form.php
@@ -1416,10 +1416,13 @@ class Form extends RequestHandler {
 					// First encode data using PHP's method of converting nested arrays to form data
 					$flatData = urldecode(http_build_query($data));
 					// Then pull the value out from that flattened string
-					preg_match('/' . addcslashes($name,'[]') . '=([^&]*)/', $flatData, $matches);
+					preg_match_all('/' . addcslashes($name,'[]') . '[^=]*=([^&]*)/', $flatData, $matches);
 
-					if (isset($matches[1])) {
+					if (isset($matches[1]) && sizeof($matches[1]) == 1) {
 						$exists = true;
+						$val = $matches[1][0];
+					} elseif(isset($matches[1])) {
+						$exists	= true;
 						$val = $matches[1];
 					}
 				}
--- a/tests/forms/FormTest.php
+++ b/tests/forms/FormTest.php
@@ -32,12 +32,21 @@ class FormTest extends FunctionalTest {
 				new TextField('key1'),
 				new TextField('namespace[key2]'),
 				new TextField('namespace[key3][key4]'),
-				new TextField('othernamespace[key5][key6][key7]')
+				new TextField('othernamespace[key5][key6][key7]'),
+				new CheckboxsetField('namespace2[Key1][Key2]', 'CheckboxsetField', array(
+					1 => "Monday",
+					2 => "Tuesday",
+					3 => "Wednesday",
+					4 => "Thursday",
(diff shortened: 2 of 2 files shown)


tractorcow:
Let me test this out locally. :D

tractorcow:
image

tractorcow:
Note that this only fails if you select a SINGLE option.

Selecting multiple options is fine.

I think your fix doesn't fix it as you thought it would. :)

tractorcow:
Have I fixed this in my select list rewrite at https://github.com/silverstripe/silverstripe-framework/pull/4568?

forms/Form.php
```diff
@@ -1416,10 +1416,13 @@ public function loadDataFrom($data, $mergeStrategy = 0, $fieldList = null) {
 					// First encode data using PHP's method of converting nested arrays to form data
 					$flatData = urldecode(http_build_query($data));
 					// Then pull the value out from that flattened string
-					preg_match('/' . addcslashes($name,'[]') . '=([^&]*)/', $flatData, $matches);
+					preg_match_all('/' . addcslashes($name,'[]') . '[^=]*=([^&]*)/', $flatData, $matches);
 
-					if (isset($matches[1])) {
+					if (isset($matches[1]) && sizeof($matches[1]) == 1) {
 						$exists = true;
+						$val = $matches[1][0];
```
tractorcow: This isn't strictly correct is it? What if it's an associative array, and the key of the first element isn't a strict '0'?

tractorcow:
Here's the code I'm using to test this.

```php
<?php

class TestPage extends Page {
	
}

class TestPage_Controller extends Page_Controller {

	private static $allowed_actions = array(
		'Form'
	);

	public function Form() {
		$fields = new FieldList(
			new CheckboxsetField('namespace2[Key1][Key2]', 'CheckboxsetField', array(
				1 => "Monday",
				2 => "Tuesday",
				3 => "Wednesday",
				4 => "Thursday",
				5 => "Friday",
				6 => "Saturday",
				7 => "Sunday"
			))
		);
		return new Form($this, 'Form', $fields, new FieldList(new FormAction('save')));
	}

	public function save($data, Form $form) {
		Debug::dump($data);
		Debug::dump($form->Fields()->dataFieldByName('namespace2[Key1][Key2]')->Value());
	}
}
```

madmatt:
Thanks @tractorcow, I'll take a look :)

madmatt:
Yeah this ends up breaking a bunch of stuff that it shouldn't, I'll take another look...

tractorcow:
This pull request has made my morning thanks @madmatt :D

tractorcow:
Sorry @madmatt :P

madmatt:
No worries - I think I ended up determining it was the way the values were being passed from the frontend that wasn't correct (I thought they were being generated by an SS form object, but it appears they were appended manually following the SS2.4 format, and they didn't get updated to 3.x format), so it's a bad PR anyway :)

PR #4731, opened 2015-11-02
BUG Prevent dev/build continually regenerating Number field type

tractorcow:
This is because the default value thinks `0.00` and `'0.00'` are different values, and continually attempts to rebuild the database.

### DIFF
--- a/model/connect/MySQLSchemaManager.php
+++ b/model/connect/MySQLSchemaManager.php
@@ -407,15 +407,17 @@ class MySQLSchemaManager extends DBSchemaManager {
 			$precision = $values['precision'];
 		}
 
-		$defaultValue = '';
+		// Fix format of default value to match precision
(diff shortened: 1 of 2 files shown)


tractorcow:
Hm, for some reason tests are failing because http://pecl.php.net/package/imagick is out of action. I've disabled imagick install for now.

tractorcow:
It looks like the postgres module needs fixing too.

tractorcow:
PGSQL fix at https://github.com/silverstripe/silverstripe-postgresql/pull/43

dhensby:
Can we try putting imagick back in please?

tractorcow:
Ok I'll try again.

tractorcow:
Re-pushed.

tractorcow:
All good @dhensby

PR #4732, opened 2015-11-02
Revert "BUG Fix duplicate files being included in case of flush"

tractorcow:
Reverts silverstripe/silverstripe-framework#4720

### DIFF
--- a/javascript/jquery-ondemand/jquery.ondemand.js
+++ b/javascript/jquery-ondemand/jquery.ondemand.js
@@ -15,74 +15,34 @@
  */
 (function($){
 
-	/**
-	 * Decode URL from header string
-	 * 
-	 * @param {String} str
-	 * @returns {String}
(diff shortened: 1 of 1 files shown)


tractorcow:
@hafriedlander your revert didn't revert the 3.1 branch properly. :D

dhensby:
@tractorcow how come this got reverted? Is the original bug still present?

tractorcow:
Because the solution was wrong; This fix works fine for static files, but would break on any php minified scripts. E.g. `/minifier.php?script=myfile.js`

dhensby:
Fun stuff! so we've identified this as a bug and know the cause, we just need a fix, yes?

Do we have an open issue?

tractorcow:
I've re-opened https://github.com/silverstripe/silverstripe-framework/issues/4553

PR #4733, opened 2015-11-02
Check for id attributes for anchor targets and prevent partial matches

feejin:
Detects id attributes as well as names when looking for anchor targets.
Prevents partial matches e.g. `<a data-myname="foo">` will no longer match.

### DIFF
--- a/forms/HtmlEditorField.php
+++ b/forms/HtmlEditorField.php
@@ -519,7 +519,7 @@ class HtmlEditorField_Toolbar extends RequestHandler {
 			}
 
 			// Similar to the regex found in HtmlEditorField.js / getAnchors method.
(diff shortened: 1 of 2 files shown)


feejin:
Forgot to add the reason for including IDs. If, for any reason, the ```<a>``` element has no validation rules in TinyMCE, or name is not explicitly included in the validation rules, it will output anchors as id. E.g. ```<a id="myAnchor"></a>```. This prevents HTMLEditorField.php and HTMLEditorField.js from detecting them as the regex only looks for ```name```.

Discovered when I allowed all elements with ```HtmlEditorConfig::get('cms')->setOptions('valid_elements' => '*[*]')```. Workaround is to change to ```*[*|name]```but a proper fallback within SilverStripe would be more user friendly.

PR #4736, opened 2015-11-03
Corrected TotalItems() method to use Count(). (fixes #4646)

### LINKED ISSUE #4646: SS_ListDecorator::TotalItems() calling wrong method
[`SS_ListDecorator::TotalItems()`](https://github.com/silverstripe/silverstripe-framework/blob/eb069e605d8cbb029d9c0270a9081060f6ded229/model/ListDecorator.php#L84) calls `$this->list->TotalItems()`, but `TotalItems()` is not defined in the `SS_List` interface. It should call either `$this->list->count()`, or just `$this->Count()` instead.

Cheddam:
I have also added a test case to GroupedListTest to cover this particular method (the entire ListDecorator class should probably be covered by unit tests to avoid similar bugs in the future).

### DIFF
--- a/model/ListDecorator.php
+++ b/model/ListDecorator.php
@@ -81,7 +81,7 @@ abstract class SS_ListDecorator extends ViewableData implements SS_List, SS_Sort
 	}
 
 	public function TotalItems() {
-		return $this->list->TotalItems();
+		return $this->list->Count();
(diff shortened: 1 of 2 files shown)


kinglozzer:
Thanks @Cheddam

Issue #4738, opened 2015-11-03
Feature: Ability to nest layout templates into sub-folders.


patricknelson:
**For example**, to allow:

```
/themes/mysite/templates/Layout/SubDirectory/MyPage.ss
```

... instead of this:

```
/themes/mysite/templates/Layout/MyPage.ss
```

That way you can naturally organize your layout templates long side potentially 20 other unrelated layouts that belong to the same theme. I believe this would be addressed via the `->handleFile()` callback in the `TemplateManifest` class. PR to demonstrate this concept coming soon (on `3` branch).

jonom:
:+1: Sounds like a good idea to me! I think there might be a much simpler solution for this though. Have a look at https://github.com/jonom/silverstripe-framework/commit/0a98d228a09cbc7f7ed90bb6ddbb8ec2e6bc84ae

SpiritLevel:
Yes, this would be REALLY handy :) I'd like to see the ability to have an arbitrary folder structure under Layout be recursively searched for templates. Similarly for Includes.

patricknelson:
@jonom I agree that's simpler, in terms of lines but I wanted to stick to concept (i.e. parent folder and grouping with the following array structure:

```php
    [mycustompage] => Array
        (
            [themes] => Array
                (
                    [mysite] => Array
                        (
                            [Layout] => /path-to-root/themes/mysite/templates/Layout/Subfolder/MyCustomPage.ss
                        )

                )

        )
```

That is, this would be more future proof and isn't as tightly bound to the `Layout` string. It's more extensible and reusable in the future and again is tied more to concept (i.e. the 4th dimensional array, just under `mysite` in this example) rather than being tightly bound to the specific type of `Layout` and otherwise always using the *parent* folder as the type. That's more magical and esoteric, in my opinion.

For example, with at https://github.com/jonom/silverstripe-framework/commit/0a98d228a09cbc7f7ed90bb6ddbb8ec2e6bc84ae, say you have nested sub-folders for your includes (this works right now coincidentally anyway, but of course it doesn't with Layouts thus necessitating the change, but I digress!) -- the result of 

```php
$templateLoader = SS_TemplateLoader::instance();
$templateManifest = $templateLoader->getManifest();
$templateManifest->regenerate();
$templates = $templateManifest->getTemplates();
print_r($templates);
```

... would be:

```php
// Truncated results above/below, focusing just on one file...
    [myinclude] => Array
        (
            [themes] => Array
[...]

PR #4739, opened 2015-11-03
NEW #4738: Ability to nest layout templates into sub-folders.

patricknelson:
New feature PR to demonstrate an implementation for issue #4738

### DIFF
--- a/core/manifest/TemplateManifest.php
+++ b/core/manifest/TemplateManifest.php
@@ -178,29 +178,62 @@ class SS_TemplateManifest {
 		$this->inited = true;
 	}
 
+
+	/**
+	 * Callback triggered by ->regenerate() above, configured via $finder->setOptions() and then ultimately triggered
+	 * by the call to $finder->find().
+	 *
+	 * @param	string	$basename	Filename of the template (excluding path, but still including file extension).
+	 * @param	string	$pathname	Full path to template file, including filename itself.
+	 * @param	string	$depth		This is not used here, but passed by the ManifestFileFinder.
+	 */
 	public function handleFile($basename, $pathname, $depth) {
 		$projectFile = false;
 		$theme = null;
 
-		if (strpos($pathname, $this->base . '/' . THEMES_DIR) === 0) {
-			$start = strlen($this->base . '/' . THEMES_DIR) + 1;
+		// Get base directory of current template.
+		$basedir = basename(dirname($pathname));
+
+		// Default to deriving the template type directly from parent folder.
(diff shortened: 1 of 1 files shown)


dhensby:
Would it be possible to write some tests to prove this code works as intended?

tractorcow:
Can you please make sure that your tests demonstrate how well the code works with `module/templates` files, as well as `themes/mytheme/templates` files? :P

tractorcow:
This feature has been superseded by the new strict template lookup in 4.x. Folders can be used to organise nested templates, but names are strictly applied, and must be used to refer to the template, rather than being a loose lookup. "folders used to organise templates but ignored during lookup" is a concept we've essentially outlawed, sorry.

tractorcow:
Nevermind, this is a PR for 3.x. I'll leave it here then. :)

dhensby:
This hasn't been touched for 1.5 years - I think it's a candidate for either closing or having someone take on the responsibility of writing the tests

tractorcow:
I'll give @patricknelson  a chance :)

patricknelson:
@tractorcow Even though this applies to `3.x`, you make a very good point here:

> Folders can be used to organise nested templates, but names are strictly applied, and must be used to refer to the template, rather than being a loose lookup. "folders used to organise templates but ignored during lookup" is a concept we've essentially outlawed, sorry.

That said, due to the impact of this change, I might pivot my approach to instead allowing the override the functionality with my own using SilverStripe's built in flexibility: Injector. That is, submit a PR to instead make this relatively simple change:

```php
// Load template manifest the main "Layouts" directory.
SS_TemplateLoader::instance()->pushManifest(Injector::inst()->create("SS_TemplateManifest",
	BASE_PATH, project(), false, $flush
));
```

... instead of `new SS_TemplateManifest` which is what is currently there right now both [in v3](https://github.com/silverstripe/silverstripe-framework/blob/3/core/Core.php#L132) and  [still apparently in v4](https://github.com/silverstripe/silverstripe-framework/blob/master/src/Core/Core.php#L101) in some form (not sure).

dhensby:
@patricknelson I'd be happy for a PR like that - I'll close this as it seems like we won't be getting it merged unless you add some tests - happy to reopen if you have the chance to do so.

patricknelson:
@dhensby See issue https://github.com/silverstripe/silverstripe-framework/issues/7305 / PR https://github.com/silverstripe/silverstripe-framework/pull/7306 (I always create an issue so I can version/number my branches anyway).

PR #4741, opened 2015-11-04
FIX xml links in SVG not being re-written

dhensby:
With SVG sprites you can do the following:

```xml
<svg class="icon">
    <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#icon-blah"></use>
</svg>
```

However, the xlink:href is not picked up because it is not an `<a>`.

This patch allows `href`s on any tag being re-written.

### DIFF
--- a/tests/view/SSViewerTest.php
+++ b/tests/view/SSViewerTest.php
@@ -1176,7 +1176,6 @@ after')
 	}
 
 	public function testRewriteHashlinks() {
-		$orig = Config::inst()->get('SSViewer', 'rewrite_hash_links'); 
 		Config::inst()->update('SSViewer', 'rewrite_hash_links', true);
 
 		$_SERVER['HTTP_HOST'] = 'www.mysite.com';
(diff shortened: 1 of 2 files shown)


dhensby:
OK, so it appears there is a deliberate test to make sure this behaviour *doesn't* happen. Which is a bit strange.

https://github.com/silverstripe/silverstripe-framework/blob/c4dc10b255a3030d54c7ee7f06dc0c4d790c0809/tests/view/SSViewerTest.php#L1199-L1203

This behaviour breaks SVG spites and also seems counter intuitive.

I don't understand why we'd only want to target `<a>` elements; plus this regex is too simplistic as it would rewrite the wrong href in this circumstance:

`<a href="#test" data-href="#test2">`

dhensby:
@tractorcow any insight on this?

tractorcow:
I think the test can be changed, but not in 3.1. Can you change the behaviour and put it on 3?

sminnee:
> My reasoning is that anyone relying on 3.1 already probably shouldn't have unexpected change in behaviour, where changes to 3 branch will only affect new projects (or adventurous people).

That sounds like a reason for it to go into `master`, guys.

kinglozzer:
@dhensby does https://github.com/silverstripe/silverstripe-framework/pull/3464 shed any light?

dhensby:
TBH I'm going to close this - I haven't used svg sprites in the wild and so I don't know how the base tag effects them. I wrote this PR because the error was reported to me but it's fallen flat and I'm not (personally) bothered enough to resolve the issue.

PR #4747, opened 2015-11-05
ENHANCEMENT CurrencyField to use Currency.currency_symbol

### LINKED ISSUE #4035: CurrencyField '$' is hardcoded
`CurrencyField` hardcodes the $ symbol in a [couple](https://github.com/silverstripe/silverstripe-framework/blob/559abecd56eec1c9a153a515c0645f1c84088e39/forms/CurrencyField.php#L20) of [places](https://github.com/silverstripe/silverstripe-framework/blob/559abecd56eec1c9a153a515c0645f1c84088e39/forms/CurrencyField.php#L48). We could use the `Currency.currency_symbol` config setting here, though it’s kinda making the form field dependant on the db field, so perhaps it’d be better to add another config option for the form field. Probably a target for 3, it was never documented that it should be configurable so it’s not really a bug.

Similar issue to #1388, though that would be a big (probably backward-incompatible) refactor.

muskie9:
fixes #4035

I have limited experience with regex, so I hope I did it correctly. I was able to save/save & publish with the curent regex and the values look good.

### DIFF
--- a/forms/CurrencyField.php
+++ b/forms/CurrencyField.php
@@ -17,7 +17,7 @@ class CurrencyField extends TextField {
 	 */
 	public function setValue($val) {
 		if(!$val) $val = 0.00;
-		$this->value = '$' . number_format((double)preg_replace('/[^0-9.\-]/', '', $val), 2);
+		$this->value = Config::inst()->get('Currency','currency_symbol') . number_format((double)preg_replace('/[^0-9.\-]/', '', $val), 2);
 		return $this;
 	}
 	/**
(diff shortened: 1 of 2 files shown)


dhensby:
Looks legit, can we get a test, please?

muskie9:
I'd be happy to take a crack at tests, might need a bit of help though.

dhensby:
Sure, what help do you need?

muskie9:
I've never really done tests for form fields before. I'll take a look at other form fields' tests in the framework and see what I can come up with. If I run into an issue I'll comment on it.

muskie9:
I've updated the existing `CurrencyFieldTest` tests. Essentially I duped what was there and tested a second time in each function with the config option updated using `Config::inst()->update()`. Let me know if I should adjust this at all.

dhensby:
Fantastic, thanks.

PR #4748, opened 2015-11-06
Cache the merged config to a persistent cache

micmania1:
This caches merged config to a persistent cache so that it doesn't have to merge on each request.

Initial tests show page load times on a base installation drop from ~~2 seconds to 0.35 seconds~~ (edit: not properly performance tested).

I'll provide more data shortly.

**Key**
localhost:8181 - **master**
vagrant:8181 - **pre-merge-config**

### DIFF
--- a/core/Config.php
+++ b/core/Config.php
@@ -273,7 +273,25 @@ class Config {
 	 */
 	public function __construct() {
 		$this->cache = new Config_MemCache();
-		$this->persistentCache = new Config_PersistentCache();
+
+		// If SS_PERSISTENTCONFIGCACHE is defined, then this will overrite the default.
+		$persistentCache = defined('SS_PERSISTENTCONFIGCACHE') ? SS_PERSISTENTCONFIGCACHE : false;
+		if(class_exists($persistentCache) && is_a($persistentCache, 'ConfigCacheInterface', true)) {
+			$this->persistentCache = new $persistentCache();
+		} else {
+			// We're too early in the startup to use config to determined 
+			// dev/live mode. We check for environment type constant.
+			$mode = defined('SS_ENVIRONMENT_TYPE') ? SS_ENVIRONMENT_TYPE : false;
+			if(!$mode || strtolower($mode) != 'dev') {
+				// If in live or test mode, use peristent cache which has performance improvements for the front-end
+				// but slows down the development process by required flush for private static changes.
+				$this->persistentCache = new Config_PersistentCache();
+			} else {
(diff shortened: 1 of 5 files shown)


kinglozzer:
Interesting approach of disabling the persistent cache for certain items when `Config::inst()->update()` is called. First question (always seems to be the case with these changes!) - how is memory use impacted?

If a request to a home page adds 50 items to the persistent cache, then a subsequent request to a different page adds another 50, you end up with 100 items in the cache: requests after that will then load all 100 items into memory. Eventually we’ll end up with nearly every config setting being effectively pre-loaded into memory, right?

micmania1:
I'm continuing work on this today. I expect some memory increase but not huge. The private statics change I made saved ~1mb so I'm cashing that in :)

micmania1:
Okay, so I've had a chance to performance test it today. There's a decent speed increase, but not 2 to 0.3 seconds. It drops from ~650ms to ~580ms (10%).

xhprof-1

The calls to SS_ConfigStaticManifest and SS_ConfigManifest drop from ~1100 to a total of 8 which is where the savings come from.

Another point to note is that memory actually drops with this PR. I'm assuming the reason for this is that its loading the config from a single file, rather than loading from 100s of files. I'm assuming there's overhead in doing this.

These tests were on a base install, MariaDB 10.0, PHP 5.6 and Apache 2.4.

I'll test on a real project and see whether there's any gains or not. However, they'll be on 3.2 and a cherry-picked commit as I don't know any projects running on 4.0.

micmania1:
For this test I used the silverstripe demo project (one ring rentals site) with demo content. My version is running 3.2.0 with this commit cherry-picked.

xhprof-2

micmania1:
https://github.com/silverstripe/silverstripe-framework/issues/4752 - I've raised this as a bug.

Its to do with the registerChangeCallback method which is removed in this PR. I don't think removing it is sensible as this means changing dev mode for example, won't clear the cache. Its important that it does clear the cache because config is different depending on except/only blocks in yaml config.

micmania1:
I've refactored this a bit from my first PR and included @sminnee fix from #4762

I need to do some more performance testing to make sure I haven't introduced anything strange but then I'll remove the DNM tag and squash when/if everyone is happy.

micmania1:
This is ready to be reviewed :)

micmania1:
I've closed this for now but will re-open. I'm getting really up and down results on performance tests so will figure out why.

PR #3748, opened 2014-12-31
BUG Prevent url= querystring argument override

tractorcow:
fixes #1368

The issue can be detected whenever `$_SERVER['QUERY_STRING']` contains two `url=.*` fragments, where the leading fragment is the one added via .htaccess, and the other the one appended to the url.

Existing url parsing functionality (built on top of parse_str) has been refactored to help and solving this issue, plus a bit of duplicate code cleanup.

I originally attempted to resolve this issue via htaccess, but decided against this, since this issue also affected IIS, and many installations will use custom htaccess rules which would not be immediately upgradable.

Tested on apache (mod_php), iis7, php built in server.

PR #4749, opened 2015-11-06
BUG: Implode url GET var to deal with array input

threesquared:
Silverstripe expects the url var to be a string and functions like trim() will fail if it is an array. Also possible security concern as errors are shown to the end user.

http://www.silverstripe.org/?url[]=test

### DIFF
--- a/main.php
+++ b/main.php
@@ -84,7 +84,7 @@ if (isset($_GET['url']) && php_sapi_name() !== 'cli-server') {
 		$parseQuery($queryString);
 	}
 
-	$url = $_GET['url'];
+	$url = is_array($_REQUEST['url']) ? implode($_GET['url']) : $_GET['url'];
 
 	// IIS includes get variables in url
 	$i = strpos($url, '?');
--- a/main.php
+++ b/main.php
@@ -84,7 +84,7 @@ if (isset($_GET['url']) && php_sapi_name() !== 'cli-server') {
 		$parseQuery($queryString);
 	}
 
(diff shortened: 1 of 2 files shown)


dhensby:
if url is an array, we should just 403 IMO

main.php
```diff
@@ -84,7 +84,7 @@
 		$parseQuery($queryString);
 	}
 
-	$url = $_GET['url'];
+	$url = is_array($_REQUEST['url']) ? implode($_GET['url']) : $_GET['url'];
```
tractorcow: You're mixing $_REQUEST with $_GET here in a dangerous fashion.

tractorcow:
I concur with @dhensby ; If url[] is an array it should error out immediately. I think it should be 400 (bad request) rather than 403 (permission denied), since the request itself is malformed.

dhensby:
> I think it should be 400 (bad request) rather than 403 (permission denied)

derp

tractorcow:
Actually, I thought this would have been prevented by my fix the other year...

https://github.com/silverstripe/silverstripe-framework/pull/3748

I guess not?

tractorcow:
implode() seems risky. I understand that it's used to get the string value, but you could end up with potentially multiple URLS concatenated together in a senseless way.

dhensby:
if the URL is an array, it is a bad request, I don't see any advantage in even trying to handle that case - it can only be a bad request because no one should be using `?url[]=x&url[]=y` on their URLs

tractorcow:
`url` querystring arg is now gone in silverstripe 4; Closing.

PR #4750, opened 2015-11-06
FIX Words ending ay/ey/iy/oy/uy are not pluralised correctly

dhensby:
Words that end in a vowel and then a "y" should have an "s" added to the end, not  have the "y" replaced with "ies".

play -> plays

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -781,7 +781,10 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 			return $name;
 		} else {
 			$name = $this->singular_name();
-			if(substr($name,-1) == 'y') $name = substr($name,0,-1) . 'ie';
+			//if the penultimate character is not a vowel, replace "y" with "ies"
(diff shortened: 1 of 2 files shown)


model/DataObject.php
```diff
@@ -781,7 +781,15 @@ public function plural_name() {
 			return $name;
 		} else {
 			$name = $this->singular_name();
-			if(substr($name,-1) == 'y') $name = substr($name,0,-1) . 'ie';
+			//if it ends in a y, then it's a special case
+			if(substr($name, -1) == 'y') {
+				$vowels = array('a', 'e', 'i', 'o', 'u');
```
willmorgan: Not a fan of regex then? :wink:

dhensby: I'd be surprised if regexing a single char was faster than an `in_array` though I'd rather do the fastest option.

somthing like `[^aeiou]y$` ?

tractorcow:
I'd prefer to fix it up stream (in the English language) but I guess I'm a few years too late for that release. :)

PR #4751, opened 2015-11-06
Fixing issue where words ending ay/ey/iy/oy/uy are not pluralised correctly

dhensby:
Words that end in a vowel and then a "y" should have an "s" added to the end, not have the "y" replaced with "ies".

play -> plays

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -781,7 +781,10 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 			return $name;
 		} else {
 			$name = $this->singular_name();
-			if(substr($name,-1) == 'y') $name = substr($name,0,-1) . 'ie';
+			//if the penultimate character is not a vowel, replace "y" with "ies"
+			if (preg_match('/[^aeiou]y$/i', $name)) {
+				$name = substr($name,0,-1) . 'ie';
+			}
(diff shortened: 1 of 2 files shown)


dhensby:
replaces #4750

jonom:
Nice. Just wondering - does this get used for all locales by default? Wonder if it should be restricted to a subset, for example:

> the almost regular forming of plural noun forms (appending an ‘s’) is hardly found in German
> http://www.gnu.org/software/gettext/manual/gettext.html#Plural-forms

Maybe we could bake in [gettext support for plural forms](http://php.net/manual/en/function.ngettext.php)? Does `i18n_plural_name()` already do something like this or do plural forms have to be manually entered for non-English languages?

dhensby:
@jonom we need a lexicon class that works for different languages for something like you're suggesting.

The way the system works is "assume English and use language files for others" and that will continue to work with this implementation.

dhensby:
Basically, write your code and classes in english and translate them using language files is the process for SS (and, I think, most programming)

jonom:
Fair enough :+1:

sminnee:
In general, this kind of automatic language processing is necessarily going to be pretty limited / language specific and so if people complain about their edge-case being missed, "using a yaml lang file" is a better solution.

Issue #4758, opened 2015-11-09
excludeMany doesn't work if there are no values on MariaDB


kiwinewt:
Example error:
```
SELECT DISTINCT count(DISTINCT "SiteTree"."ID") AS "Count"

FROM "SiteTree"

WHERE ("SiteTree"."ParentID" = ?)
 AND (("SiteTree"."ID" != ?))
 AND (("SiteTree"."ClassName" NOT IN ()))

You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ')))' at line 5 (http://mysite/admin/pages/listview)
```

This is easily fixable in framework/search/filters/ExactMatchFilter.php, the excludeMany function (line 104?) by going from:
```
if ($caseSensitive === null) {
        //more code here
} else {
        //more code here
}
```
to
```
if(!empty($values)) {
    if ($caseSensitive === null) {
        //more code here
    } else {
        //more code here
    }
}
```

tractorcow:
In 3.1 it looks like this was the code that handled empty arrays

```php
$valueStr = "'" . implode("', '", $values) . "'";
return $query->where(sprintf(
	'%s IN (%s)',
	$this->getDbName(),
	$valueStr
));
```

Which essentially is `%S in ('')`

kiwinewt:
I was running this on 3.2, and having the wrap there is working across MySQL and MariaDB. I don't know if the code has changed since 3.2?

dhensby:
This is a problem for `filter` and `exclude`

I think just wrapping the sql generation in an `if (!empty())` will have unintended consequences

if the values are empty then we should do `!= ''` or `= ''` depending on filter/exclude

tractorcow:
@dhensby we need to be careful because some ORM don't like string comparison vs integer field types. It's not safe to assume empty string comparison either.

tractorcow:
Ok, for 3.2 let's just keep the 3.1 behaviour. It's not strictly "correct" but we can't change it too much in a minor release.

dhensby:
agreed

kiwinewt:
According to https://github.com/silverstripe/silverstripe-framework/blame/3.2/search/filters/ExactMatchFilter.php this has been the case since 2013? If so wouldn't changing it cause more issue now?

kiwinewt:
It also looks like it's completely reworked in the master branch, which looks really good

tractorcow:
[...]

PR #4760, opened 2015-11-10
BUG Correct behaviour for empty filter array (as per 3.1)

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/4758

This isn't strictly correct, but it does bring the 3.2 behaviour in line with 3.1, which was to map filters against empty arrays to empty string comparison.

### DIFF
--- a/search/filters/ExactMatchFilter.php
+++ b/search/filters/ExactMatchFilter.php
@@ -56,6 +56,10 @@ class ExactMatchFilter extends SearchFilter {
 			// For queries using the default collation (no explicit case) we can use the WHERE .. IN .. syntax,
 			// providing simpler SQL than many WHERE .. OR .. fragments.
 			$column = $this->getDbName();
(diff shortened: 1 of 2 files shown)


tractorcow:
In 4.x I think we need to address a better solution for these kinds of situations.

`->filter('Field', array())` is effectively saying there are no valid values for Field, so it should be an empty set.

However, in 3.1 this does empty string comparison, so we can't change it in 3.2 without regression.

I'd prefer InvalidArgumentException, personally. :P

tractorcow:
I'm merging this change up to master now, and there are a few issues... especially with the 'null' check, so I'm going to substitute the "3.1 hack" with an InvalidArgumentException for the merge up. :)

dhensby:
:(

tractorcow:
I think a problem is that in master we have a new "null" check, and it was mistakenly treating filter("Field", array()" as filter("Field", null)

PR #4762, opened 2015-11-11
FIX: Allow caching of false config values (Fixes #4755)

### LINKED ISSUE #4755: False config values are repeatedly calculated
`Config::get()` will treat [`false` values returned from the cache](https://github.com/silverstripe/silverstripe-framework/blob/320c717358c8237100c113b6e4f21f77ff37a438/core/Config.php#L588) as if they haven’t been calculated yet, meaning that false config values may be looked up repeatedly with `Config::getUncached()`. As an example, opening a page in the CMS results in `SSViewer.theme_enabled` being re-caculated 30 times. It’s still not a major performance issue, but every little helps :wink:.

I’d suggest swapping to checking against `null`, but then we’d have the same problem for null values. The only way I can think to resolve this is to return a constant (something like `Config::FALSE_VALUE`) to indicate that the value returned from the cache needs to be transformed to `false`:

```php
// Config
const FALSE_VALUE = '%%CONFIG_FALSE%%';

public function get() {
    if ($result === false) {
        // get uncached
    } elseif ($result === self::FALSE_VALUE) {
        $result = false;
    }
    return $result;
}
```

sminnee:
False values were indistinguishable from cache misses. This patch solves
this by introducing a new multiple-return-value varation on cache->get().
The new method is named differently so as not to break a public API.

### DIFF
--- a/core/Config.php
+++ b/core/Config.php
@@ -585,7 +585,8 @@ class Config {
 		// Have we got a cached value? Use it if so
 		$key = $class.$name.$sourceOptions;
 
-		if (($result = $this->cache->get($key)) === false) {
+		list($cacheHit, $result) = $this->cache->checkAndGet($key);
+		if (!$cacheHit) {
 			$tags = array();
 			$result = null;
 			$this->getUncached($class, $name, $sourceOptions, $result, $suppress, $tags);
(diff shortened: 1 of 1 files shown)


core/Config.php
```diff
@@ -832,13 +853,23 @@ public function stats() {
 	}
 
 	public function get($key) {
+		list($hit, $result) = $this->checkAndGet($key);
+		return $hit ? $result : false;
+		return false;
```
micmania1: You have a double return.

micmania1:
I came across this issue last week. The fix looks good, test failure aside.

kinglozzer:
Nice work @sminnee! I like this approach much more than my own

core/Config.php
```diff
@@ -832,13 +853,23 @@ public function stats() {
...
+	}
+
+	/**
+	 * Checks for a cache hit and returns the value as a multi-value return
+	 * @return array First element boolean, isHit. Second element the actual result.
+	 */
+	public function checkAndGet($key) {
 		if(isset($this->cache[$key])) {
```
tractorcow: Should this be array_key_exists, since `null` could be cached?

core/Config.php
```diff
@@ -769,17 +770,37 @@ public function stats() {
...
+
+	/**
+	 * Checks for a cache hit and looks up the value by returning multiple values.
+	 * Distinguishes a cached 'false' value from a cache miss.
+	 *
+	 * @param  string $key The cache key
+	 * @return array  First element boolean, isHit. Second element the actual result.
+	 */
+	public function checkAndGet($key) {
 		if (isset($this->indexing[$key])) {
```
tractorcow: As below, should we use `array_key_exists` to support null values in this array?

sminnee: I didn't change this code, but it's probably a good idea.

core/Config.php
sminnee: I want to make really sure that it returns false.

kinglozzer:
Bump @sminnee

dhensby:
Bump!

core/Config.php
tractorcow: Trying to read this code is hard; $i, $indexing etc aren't documented so it makes no sense to me. Worth documenting these if you ever get around to resurrecting this PR @sminnee :D

tractorcow:
Are we still keeping this?

sminnee:
Alright I'll have a look when I get a chance.. .;)

dhensby:
The time is now!

sminnee:
*sigh* ...okay!

sminnee:
@dhensby fixed the duplicate return and array_key_exists.

kinglozzer:
Nice work @sminnee 😄

PR #4763, opened 2015-11-11
FIX: Don’t reuse DBConnector (fixes #4735)

### LINKED ISSUE #4735: [3.2] There are no way to connect to different databases with DB::connect()
The reason is that Injector builds all databases with shared connector.

So, for MySQLDatabase there is just one shared service MySQLiConnector.
```
$dbA = DB::connect($dbAconfig, 'default');
$dbB = DB::connect($dbBconfig, 'alt');
```
$dbB is actually connected to the server A.

sminnee:
Without this fix, multiple Database objects will use the same
DBConnector object, meaning that SilverStripe can’t connect to multiple
databases.

This fix ensures that the service cache generates new DBConnector
subclasses each time it is called.

### DIFF
--- a/_config/database.yml
+++ b/_config/database.yml
@@ -14,3 +14,9 @@ Injector:
       connector: %$MySQLiConnector
       schemaManager: %$MySQLSchemaManager
       queryBuilder: %$MySQLQueryBuilder
+  MySQLiConnector:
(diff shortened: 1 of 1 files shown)


tractorcow:
Yep good fix @sminnee !

PR #4768, opened 2015-11-12
BUG: magic getter in ViewableData behaves differently than direct calls

stojg:
There is my opinion a bug when calling a ViewableData method with an argument and using either the `getMethod('argument')` (true method name) and `Method('argument')` (magic function lookup).

This PR will do the method lookup in ViewableData::obj() instead of in ViewableData::__get() where the arguments are lost.

This often gets confusing in SilverStripe templates.

### DIFF
--- a/tests/view/ViewableDataTest.php
+++ b/tests/view/ViewableDataTest.php
@@ -8,6 +8,22 @@
  */
 class ViewableDataTest extends SapphireTest {
 
+	/**
+	 * This test checks that getting a value by `PropertyName(property)` and `getPropertyName(property)`
+	 * returns the same value when there is a PropertyName() method in this object
+	 */
+	public function testMagicGetterMethod() {
+		$obj = new ViewableDataTest_MethodLookup();
+		$funcCall = $obj->obj('getMyMethod', array('fibbonacci'));
+		$getValue = $obj->obj('MyMethod', array('fibbonacci'));
+		$this->assertEquals($funcCall->value, $getValue->value);
+		// double check that the default value still works
+		$defaultFuncCall = $obj->obj('getMyMethod');
+		$this->assertEquals('default', $defaultFuncCall->value);
+		$defaultValueGet = $obj->obj('MyMethod');
+		$this->assertEquals('default', $defaultValueGet->value);
+	}
+
 	public function testRequiresCasting() {
 		$caster = new ViewableDataTest_Castable();
 
@@ -282,6 +298,13 @@ class ViewableDataTest_NotCached extends ViewableData {
 	protected function objCacheGet($key) {
 		// Disable caching
 		return null;
+	}
+}
+
+class ViewableDataTest_MethodLookup extends ViewableData {
+
+	public function getMyMethod($argument = 'default') {
+		return $argument;
 	}
 }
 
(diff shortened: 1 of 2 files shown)


dhensby:
Hmm, this has been an "issue" for a long time. 

My approach has always been that a function prefixed with "get" was a way to either override the default property successor logic or polyfill your own pretend properties into an object. 

As properties can't take arguments, these functions shouldn't take arguments either. 

By patching this "bug" like you have, we also open up ambiguity around order of importance of functions and properties on objects. 

I can see $Field being particularly problematic in templates given its a function on viewable data, but also in form loops. I'm sure there will be other issues too. 

The order of resolution of these in the template and other areas is already inconsistent. I fear this would make it worse and loosens any conventions we have around method naming

tests/view/ViewableDataTest.php
```diff
@@ -8,6 +8,22 @@
  */
 class ViewableDataTest extends SapphireTest {
 
+	/**
+	 * This test checks that getting a value by `PropertyName(property)` and `getPropertyName(property)`
+	 * returns the same value when there is a PropertyName() method in this object
+	 */
+	public function testMagicGetterMethod() {
+		$obj = new ViewableDataTest_MethodLookup();
+		$funcCall = $obj->obj('getMyMethod', ['fibbonacci']);
```
dhensby: You can't use those silly square brackets in 5.3

stojg: Oh. This isn't JavaScript? Will fix.

stojg:
Personally I've not a huge fan of using the magic function calls in PHP (or any other languages) due to it makes, among other things, debugging harder.

It really is a bug dressed up in a feature clothing and as you said, for experienced silverstripe developers it's handy to have around. I can user `MyField()`, `getMyField()` and `obj()` and get slightly different results, which can be nice.

So how do we follow the 'principle of least confusion` in this case? Documentation is probably last resort unless it's very very good.

sminnee:
This is not a bug.

sminnee:
OK, a few issues.

1. You're duct-taping together two unrelated parts of the system: magic properties and the template engine. Now, I can understand that there is a confusion here but I think it would need more consideration than an extra if statement.

2. It's confusing enough that there are 2 different options "methods or properties", but now we have 3, and "methods, properties, or methods that would exist if you put 'get' in front of what I wrote" is difficult to advocate as a clear rule.

3. It's generally a bad idea to address these kind of things for a 2-line patch. If we're going to adjust the semantics of the template language I think a complete review (e.g. an RFC) would be more appropriate and likely the resulting change would be a 4.x change.

5. Decoupling the template engine from the rest of the framework (so you can put non-ViewableData objects into it), and changes that increase our reliance/expectation of magic from the data being put into the template engine (such as this one) pull away from that.

If you're trying to pass arguments to a magic getter, we should probably throw a warning that those arguments are ignored, rather than trying to guess what maybe the programmer meant. However, I'm not sure if throwing a warning where previously we silently ignored the issue is a 3.x fix for a 4.x fix?

In any case, telling the developer what they're doing wrong is probably better than trying to guess what they might have meant.

tractorcow:
Ah sorry but I have to concur with @sminnee and @dhensby.

The standard should be "don't write magic getters with methods". E.g. `getProperty()` should not have parameters, since magic-properties cannot be called with parameters. `$obj->Property('var')` should fail in PHP, so it isn't correct to refer to it that way in templates either.

PR #4771, opened 2015-11-13
DOCS Update conflict resolution help in code of conduct

camfindlay:
Original link was 404.


tractorcow:
I also think we should be careful not to be so conflict-adverse that we don't allow a healthy level of disagreement on open source. Having others disagree with us (and focus on both negative and positive elements of suggestions) is important to our success.

Just my 2c though. :)

camfindlay:
Conflict resolution is not about being adverse to conflict. It's about acknowledging it and working through the process.

tractorcow:
In any case, the link is fixed and I think the document adds value. Thanks @camfindlay for fixing this.

PR #4772, opened 2015-11-13
Improved PHPDoc comments

assertchris:
This covers roughly half the class files in `framework/control`. Breaking the main body of work into two pull requests to help with review.


tractorcow:
Nice improvement, thanks for all your hard work!

PR #4777, opened 2015-11-16
Don't run forceSSL() on CLI

chillu:
Was performing an inefficient check for this previously.
REQUEST_URI is set through cli-script.php, hence it's not
an accurate way to determine CLI operation.

The effect of this bug is that command line tasks will fail to
run since they'll return a HTTP redirect which a CLI client can't interpret.

Note: I don't know how to unit test this, since `Director::is_cli()` is called statically and uses `php_sapi_name()` which I can't mock.


tractorcow:
It's an exact duplicate. :)

micmania1:
Oops. Updated to #4025 :p

chillu:
Since the discussion has advanced more on #4025, I'm going to close this one.

PR #4781, opened 2015-11-18
Fixing regression from #4733

dhensby:
Fixes regression casing behat failures on CMS.

### DIFF
--- a/forms/HtmlEditorField.php
+++ b/forms/HtmlEditorField.php
@@ -578,7 +578,7 @@ class HtmlEditorField_Toolbar extends RequestHandler {
 
 			// Similar to the regex found in HtmlEditorField.js / getAnchors method.
 			if (preg_match_all("/\s(name|id)=\"([^\"]+?)\"|\s(name|id)='([^']+?)'/im", $page->Content, $matches)) {
-				$anchors = $matches[1];
+				$anchors = array_filter(array_merge($matches[2], $matches[4]));
 			}
 
 		} else {


kinglozzer:
I think this might be too simplistic - the regular expression tries to support both single and double quotes, which come out separately: https://3v4l.org/0MhfT :(

(neither this change or #4733 introduced that, the original regex suffers from the same issue)

forms/HtmlEditorField.php
```diff
@@ -578,7 +578,7 @@ public function getanchors() {
 
 			// Similar to the regex found in HtmlEditorField.js / getAnchors method.
 			if (preg_match_all("/\s(name|id)=\"([^\"]+?)\"|\s(name|id)='([^']+?)'/im", $page->Content, $matches)) {
```
willmorgan: This regex should really get cleaned up - it's not massively clear what the 2nd position even is in this case.

kinglozzer:
We should be able to safely merge the two sets of matches, though: https://3v4l.org/21qve. Does this impact JS too?

dhensby:
The JS is not effected as it works quite differently.

Good spot, though.

forms/HtmlEditorField.php
dhensby: I thought the same thing; but cleaning up the regex to something like:

`/\s(name|id)\s*=\s*["']?([^'"]+?)['"]?\s*/` (ie: one unified regex) means that an attribute like so wouldn't be matched properly: `<a name="test's">` as the `'` would cause the regex to return `test` instead of `test's`

dhensby: PS: I'm trying to patch a bug, not improve the code, that can go in 3.x

kinglozzer:
Okay, tests passing, gonna merge this as it’s the simplest & safest fix for a patch release.

kinglozzer:
The regex really does need some love, trouble is there are so many variants to account for!

```
<div name="foo"></div>
<div name='bar'></div>
<div id="baz"></div>
<div id='bam'></div>
<div id = "baz"></div>
<div id="some'id"></div>
<div id=bar></div>
... and probably more
```

Perhaps our old friend `DOMDocument` might be safer (albeit probably slower).

dhensby:
see #4782 for improved regex on 3.x-dev

PR #4785, opened 2015-11-19
DOCS update LinkingModes description

muskie9:
see silverstripe/silverstripe-cms#1325

### DIFF
--- a/docs/en/02_Developer_Guides/01_Templates/02_Common_Variables.md
+++ b/docs/en/02_Developer_Guides/01_Templates/02_Common_Variables.md
@@ -153,30 +153,30 @@ link.
 ### Linking Modes
 
 	:::ss
-	$LinkingMode
+	$isSection
+	$isCurrent
 
-When looping over a list of `SiteTree` instances through a `<% loop $Menu %>` or `<% loop $Children %>`, `$LinkingMode`
-will return context about the page relative to the currently viewed page. It can have the following values:
-
- * `link`: You are neither on this page nor in this section.
- * `current`: You are currently on this page.
- * `section`: The current page is a child of this menu item, so the current "section"
+When looping over a list of `SiteTree` instances through a `<% loop $Menu %>` or `<% loop $Children %>`, `$isSection` and `$isCurrent`
+will return true or false based on page being looped over relative to the currently viewed page. 
 
 For instance, to only show the menu item linked if it's the current one:
 
 	:::ss
-	<% if $LinkingMode = current %>
+	<% if $isCurrent %>
 		$Title
 	<% else %>
 		<a href="$Link">$Title</a>
 	<% end_if %>
+	
+An example for checking for `current` or `section` is as follows:
 
-`$LinkingMode` is reused for several other variables and utility functions.
+    :::ss
+    <a class="<% if $isCurrent %>current<% else_if $isSection %>section<% end_if %>" href="$Link">$MenuTitle</a>
 
- * `$LinkOrCurrent`: Determines if the item is the current page. Returns "link" or "current" strings.
- * `$LinkOrSection`: Determines if the item is in the current section, so in the path towards the current page. Useful 
- for menus which you only want to show a second level menu when you are on that page or a child of it. Returns "link" 
- or "section" strings.
+
+
+**Additional Utility Method**
+
  * `$InSection(page-url)`: This if block will pass if we're currently on the page-url page or one of its children.
 
 	:::ss
--- a/docs/en/02_Developer_Guides/01_Templates/How_Tos/01_Navigation_Menu.md
+++ b/docs/en/02_Developer_Guides/01_Templates/How_Tos/01_Navigation_Menu.md
@@ -11,15 +11,15 @@ top level menu with a nested second level using the `Menu` loop and a `Children`
 	<ul>
 		<% loop $Menu(1) %>
 			<li>
-				<a href="$Link" title="Go to the $Title page" class="$LinkingMode">
+				<a href="$Link" title="Go to the $Title page" class="<% if $isCurrent %>current<% else_if $isSection %>section<% end_if %>">
 					$MenuTitle
 				</a>
 
-				<% if $LinkOrSection == section %>
+				<% if $isSection %>
 					<% if $Children %>
 						<ul class="secondary">
 							<% loop $Children %>
-								<li class="$LinkingMode"><a href="$Link">$MenuTitle</a></li>
+								<li class="<% if $isCurrent %>current<% else_if $isSection %>section<% end_if %>"><a href="$Link">$MenuTitle</a></li>
 							<% end_loop %>
 						</ul>
 					<% end_if %>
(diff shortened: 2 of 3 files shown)


dhensby:
Can you please recommend to use the `is` functions in the template than relying on the returned string from the functions?

muskie9:
Sure can, I saw your comment on the other PR. I probably would've gone with just a docs update initially rather than the cms PR if I had realized those `is` functions existed.

muskie9:
I've updated this to reference the `is` methods. Let me know if it should be adjusted at all.

docs/en/02_Developer_Guides/01_Templates/02_Common_Variables.md
```diff
@@ -184,6 +184,14 @@ For instance, to only show the menu item linked if it's the current one:
 		<p>You are viewing the about us section</p>
 	<% end_if %>
 
+If you need alternative values for for "current" or "section" it's suggested you check against the `is` methods (`$isSection` and `$isCurrent`) and define you'r own value in the template:
```
dhensby: `you'r` should be your

jonom: * `for for "current"` should be `for "current"`
* `you'r` should be `your`

docs/en/02_Developer_Guides/01_Templates/02_Common_Variables.md
```diff
@@ -184,6 +184,14 @@ For instance, to only show the menu item linked if it's the current one:
...
+
+    :::ss
+    <a class="
+        <% if $isSection %>my-section-class<% end_if %>
+        <% if $isCurrent %>my-current-class<% end_if %>
+    " href="foo">Foo</a>
```
jonom: Can you make this `" href="$Link">$MenuTitle</a>`

(sorry I should have included this in the original example)

jonom:
This is a good start and would be a good addition on its own. If you're up for it though @muskie9 think it would be nice to also deprecate the `LinkOrCurrent` and `LinkOrSection` methods in this PR (and possibly the LinkingMode method too) and update all the examples that use those methods to use the `is` methods instead. Then it can be all wrapped up neatly in one commit.

Relevant doc files are:
* 01_Tutorials/01_Building_A_Basic_Site.md
* 02_Developer_Guides/01_Templates/02_Common_Variables.md
* 02_Developer_Guides/01_Templates/How_Tos/01_Navigation_Menu.md

I'd be fine with deprecating `LinkingMode` and I think @dhensby is too. @clarkepaul do you have an opinion on this since it's more of a front-end issue?

Background discussion here: https://github.com/silverstripe/silverstripe-cms/pull/1325

muskie9:
I should've proof read my commit a bit more before force pushing I guess ;) I can take a shot at deprecating the methods if that's the consensus. I'm assuming the PR would go against cms:master for the deprecation.

dhensby:
I think avoid the depreciation here. It will block merge and we'll need concensus to take it in.

jonom:
> Minor version Y (x.Y.z | x > 0) ... MUST be incremented if any public API functionality is marked as deprecated

So FYI marking methods deprecated for 4.0 would go in to 3 branch, but @dhensby's got a point so maybe don't do that yet ;)

Still would be cool to update the docs to not mention those string methods if you have time, would highlight their redundancy and pave the way for deprecating them in the future.

dhensby:
Ahreed

muskie9:
:+1:

muskie9:
I've done some preliminary updates to the docs noted above. Let me know if I missed anything or wording needs to be changed.

docs/en/02_Developer_Guides/01_Templates/How_Tos/01_Navigation_Menu.md
```diff
@@ -11,15 +11,15 @@ top level menu with a nested second level using the `Menu` loop and a `Children`
 	<ul>
 		<% loop $Menu(1) %>
 			<li>
-				<a href="$Link" title="Go to the $Title page" class="$LinkingMode">
+				<a href="$Link" title="Go to the $Title page" class="<% if $isSection %>section<% end_if %> <% if $isCurrent %>$current<% end_if %>">
```
jonom: Extra $ sign in there, should be `<% if $isCurrent %>current<% end_if %>`

docs/en/02_Developer_Guides/01_Templates/How_Tos/01_Navigation_Menu.md
```diff
@@ -11,15 +11,15 @@ top level menu with a nested second level using the `Menu` loop and a `Children`
...
 					$MenuTitle
 				</a>
 
-				<% if $LinkOrSection == section %>
+				<% if $isSection %>
 					<% if $Children %>
 						<ul class="secondary">
 							<% loop $Children %>
-								<li class="$LinkingMode"><a href="$Link">$MenuTitle</a></li>
+								<li class="<% if $isSection %>section<% end_if %> <% if $isCurrent %>$current<% end_if %>"><a href="$Link">$MenuTitle</a></li>
```
jonom: same here :)

muskie9: Must have been muscle memory lol

jonom:
I think it looks good apart from a couple of stray `$` signs.

Sorry for being so pedantic and changing my mind on my own feedback but I wonder if we should change all these:
```
class="<% if $isSection %>section<% end_if %> <% if $isCurrent %>current<% end_if %>"
```
to 
```
class="<% if $isCurrent %>current<% else_if $isSection %>section<% end_if %>"
```
This then mirrors the $LinkingMode output but without adding the redundant 'link' class, and it looks a bit cleaner. More backwards compatible and a good example for accomplishing your original goal.

muskie9:
@jonom I was actually going to ask if we should go as far as `<% if $isCurrent %>current<% else_if $isSection %>section<% else %>link<% end_if %>`

jonom:
@muskie9 :+1:

p.s. my bad I should have remembered that tutorials are deprecated in favour of lessons. Still worth updating I think though.

@unclecheese what would be involved in updating http://www.silverstripe.org/learn/lessons/adding-dynamic-content to not use $LinkingMode? We're thinking it's easy to accomplish same with $isSection and $isCurrent, so may be better to show developers how to provide their own class names in example. This should be especially handy for people working with frameworks like Bootstrap.

jonom:
> I was actually going to ask if we should go as far as <% if $isCurrent %>current<% else_if $isSection %>section<% else %>link<% end_if %>

Personally I would prefer to leave out the link part... I can't think of anytime the 'link' class would be useful when you can just target the `a` element, so it seems redundant to me.

muskie9:
I've updated the docs per your suggestion @jonom. I don't do a lot of front-end dev, so I presume most links are set via `a` rather than `a.link` in most cases (outside of current and/or section).

jonom:
Thanks for seeing it through @muskie9!

muskie9:
Happy to! Glad we got it figured out!

clarkepaul:
@jonom @muskie9  I haven't really been involved in the frontend side of things for a while now but from what I can tell this sounds like a good idea and I can see it being useful. I've sent a link to few others here just in case there is any concerns.

clarkepaul:
Ah, I didn't see this was merged - no worries, happy days

jonom:
@clarkepaul we haven't deprecated anything yet though so if you have an opinion on that let us know ;)

dhensby:
@clarkepaul this is just a DOCs change now; we haven't made it customisable... yet

PR #4786, opened 2015-11-20
Adding a background to the UploadField preview

nickspiel:
I could not see a preview of a white image with background transparency in the preview area. Adding a background to this area helps make it more visible, at the very least you can tell them image has transparency. 

**Before**
before

**After:**
after


jonom:
This seems like a good idea to me - can you remove any layout impact in this change though by applying the background image to the img element instead of the container?

In that case the change is so minor I think it might be okay to sneak this in to 3.2 as a bug fix :D but otherwise I'd say it should probably go in to 3 as a new feature according to SemVer.

nickspiel:
Good call, have updated this so img tag will get the background - avoiding any layout change.

dhensby:
Please squash these commits

nickspiel:
Squashed.

dhensby:
thanks

jonom:
Thanks Nick!

PR #4793, opened 2015-11-24
FIX Image_Cached record class name

jonom:
This prevents Image_Cached objects having their `ClassName` and `RecordClassName` properties set to 'Image' instead of 'Image_Cached'.

I don't know what those properties are used for but if I try to replace the Image class with the injector I run in to problems without this fix.


jonom:
ping @tractorcow this is a follow-up from https://github.com/silverstripe/silverstripe-framework/pull/4578 that would be good to have in for 3.3.0.

tractorcow:
thanks @jonom :D

PR #4795, opened 2015-11-25
API refactor LeftAndMain_Menu.ss into individually overridable components

tractorcow:
Since many modules override the main menu, it makes sense to cut down on code duplication (and maintenance) by splitting out re-usable components.

Fluent, translatable, and subsites, for instance, each add custom left hand menus.

### DIFF
--- a/admin/templates/Includes/LeftAndMain_Menu.ss
+++ b/admin/templates/Includes/LeftAndMain_Menu.ss
@@ -1,42 +1,13 @@
 <div class="cms-menu cms-panel cms-panel-layout west" id="cms-menu" data-layout-type="border">
 	<div class="cms-logo-header north">
-		<div class="cms-logo">
-			<a href="$ApplicationLink" target="_blank" title="$ApplicationName (Version - $CMSVersion)">
-				$ApplicationName <% if $CMSVersion %><abbr class="version">$CMSVersion</abbr><% end_if %>
-			</a>
-			<span><% if $SiteConfig %>$SiteConfig.Title<% else %>$ApplicationName<% end_if %></span>
-		</div>
-
-		<div class="cms-login-status">
-			<a href="$LogoutURL" class="logout-link font-icon-logout" title="<%t LeftAndMain_Menu_ss.LOGOUT 'Log out' %>"></a>
-			<% with $CurrentMember %>
-				<span>
-					<%t LeftAndMain_Menu_ss.Hello 'Hi' %>
-					<a href="{$AbsoluteBaseURL}admin/myprofile" class="profile-link">
-						<% if $FirstName && $Surname %>$FirstName $Surname<% else_if $FirstName %>$FirstName<% else %>$Email<% end_if %>
-					</a>
-				</span>
-			<% end_with %>
-		</div>
+		<% include LeftAndMain_MenuStatus %>
 	</div>
 
 	<div class="cms-panel-content center">
-		<ul class="cms-menu-list">
-		<% loop $MainMenu %>
-			<li class="$LinkingMode $FirstLast <% if $LinkingMode == 'link' %><% else %>opened<% end_if %>" id="Menu-$Code" title="$Title.ATT">
-				<a href="$Link" $AttributesHTML>
-					<span class="icon icon-16 icon-{$Code.LowerCase}">&nbsp;</span>
-					<span class="text">$Title</span>
-				</a>
-			</li>
-		<% end_loop %>
-		</ul>
+		<% include LeftAndMain_MenuList %>
 	</div>
 
 	<div class="cms-panel-toggle south">
-		<button class="sticky-toggle" type="button" title="Sticky nav">Sticky nav</button>
-		<span class="sticky-status-indicator">auto</span>
-		<a class="toggle-expand" href="#"><span>&raquo;</span></a>
-		<a class="toggle-collapse" href="#"><span>&laquo;</span></a>
+		<% include LeftAndMain_MenuToggle %>
 	</div>
 </div>
--- a/admin/templates/Includes/LeftAndMain_MenuList.ss
+++ b/admin/templates/Includes/LeftAndMain_MenuList.ss
@@ -1,0 +1,10 @@
+<ul class="cms-menu-list">
+	<% loop $MainMenu %>
+		<li class="$LinkingMode $FirstLast <% if $LinkingMode == 'link' %><% else %>opened<% end_if %>" id="Menu-$Code" title="$Title.ATT">
+			<a href="$Link" $AttributesHTML>
+				<span class="icon icon-16 icon-{$Code.LowerCase}">&nbsp;</span>
+				<span class="text">$Title</span>
+			</a>
+		</li>
+	<% end_loop %>
+</ul>
--- a/admin/templates/Includes/LeftAndMain_MenuStatus.ss
+++ b/admin/templates/Includes/LeftAndMain_MenuStatus.ss
@@ -1,0 +1,17 @@
+<div class="cms-logo">
+    <a href="$ApplicationLink" target="_blank" title="$ApplicationName (Version - $CMSVersion)">
+		$ApplicationName <% if $CMSVersion %><abbr class="version">$CMSVersion</abbr><% end_if %>
+    </a>
+    <span><% if $SiteConfig %>$SiteConfig.Title<% else %>$ApplicationName<% end_if %></span>
+</div>
+<div class="cms-login-status">
+	<a href="$LogoutURL" class="logout-link font-icon-logout" title="<%t LeftAndMain_Menu_ss.LOGOUT 'Log out' %>"></a>
+	<% with $CurrentMember %>
+		<span>
+			<%t LeftAndMain_Menu_ss.Hello 'Hi' %>
+			<a href="{$AbsoluteBaseURL}admin/myprofile" class="profile-link">
+				<% if $FirstName && $Surname %>$FirstName $Surname<% else_if $FirstName %>$FirstName<% else %>$Email<% end_if %>
+			</a>
+		</span>
+	<% end_with %>
+</div>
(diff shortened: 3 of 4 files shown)


admin/templates/Includes/LeftAndMain_Menu.ss
```diff
@@ -1,36 +1,10 @@
 <div class="cms-menu cms-panel cms-panel-layout west" id="cms-menu" data-layout-type="border">
 	<div class="cms-logo-header north">
```
stevie-mayhew: It make more sense to pull this line into the include as sometimes we don't use the north section and it would allow further customisation. Same for the MenuStatus include.

admin/templates/Includes/LeftAndMain_Menu.ss
```diff
@@ -1,36 +1,10 @@
...
-				</a>
-			</li>
-		<% end_loop %>
-		</ul>
+
+    <div class="cms-panel-content center">
+		<% include LeftAndMain_MenuList %>
 	</div>
 		
 	<div class="cms-panel-toggle south">
```
stevie-mayhew: Could this also be included as a template - might as well if the rest are so? (We've also restyled this in custom left and main includes)

tractorcow: It's a pretty minor component. Does it deserve it's own template? (Maybe?)

stevie-mayhew: In the case where you wanted to remove/customise it then you would have to customise this file rather than just the include - I think it makes sense as it is a "component" not a required piece of code?

admin/templates/Includes/LeftAndMain_Menu.ss
tractorcow: It would make overriding the top component easier, since you could then implement only that template.

However my intention is to cut down on duplication of the low level html in overriding templates, to improve maintainability.

E.g.

```
<div class="cms-logo-header north">
	<% include LeftAndMain_MenuStatus %>
	<div>My custom north component</div>
</div>
```

Versus

```
<div class="cms-logo">
    <a href="$ApplicationLink" target="_blank" title="$ApplicationName (Version - $CMSVersion)">
		$ApplicationName <% if $CMSVersion %><abbr class="version">$CMSVersion</abbr><% end_if %>
    </a>
    <span><% if $SiteConfig %>$SiteConfig.Title<% else %>$ApplicationName<% end_if %></span>
</div>
<div class="cms-login-status">
	<a href="$LogoutURL" class="logout-link font-icon-logout" title="<%t LeftAndMain_Menu_ss.LOGOUT 'Log out' %>"></a>
	<% with $CurrentMember %>
		<span>
			<%t LeftAndMain_Menu_ss.Hello 'Hi' %>
			<a href="{$AbsoluteBaseURL}admin/myprofile" class="profile-link">
				<% if $FirstName && $Surname %>$FirstName $Surname<% else_if $FirstName %>$FirstName<% else %>$Email<% end_if %>
			</a>
		</span>
	<% end_with %>
</div>
<div>My custom north component</div>
```

admin/templates/Includes/LeftAndMain_Menu.ss
tractorcow: RE your above comment, maybe I should split it out. What do you think?

admin/templates/Includes/LeftAndMain_Menu.ss
stevie-mayhew: Ah okay I see so its about overriding the LeftAndMain_Menu.ss rather than the custom components which makes sense. Sorry I didn't think about flipping it that way around :smile:

tractorcow: https://github.com/silverstripe/silverstripe-subsites/blob/master/templates/LeftAndMain_Menu.ss is an example of one override.

stevie-mayhew: Yeah cool - ignore the above.

An example in the docs of the intended usage would be a good idea.

admin/templates/Includes/LeftAndMain_Menu.ss
stevie-mayhew: Nope, now I understand the way its supposed to be used which is the opposite of my original thoughts it can stay as is!

stevie-mayhew: Although if this changes in Framework every module would have to update it... maybe it is better as an include..

tractorcow: I've broken it out as you suggested and pushed it up. ;)

stevie-mayhew:
:+1:

dhensby:
rebase?

dhensby:
:+1:

tractorcow:
Yep I'll do it soonish.

jonom:
Hey @tractorcow, @stevie-mayhew pointed me here from https://github.com/silverstripe-australia/silverstripe-grouped-cms-menu/issues/8#issuecomment-194457578 I think this is a great idea but does it have to go in to master? If all you're doing here is separating some sections out in to includes I don't think that's an API change, and people will be using 3 for a long time yet so think we should do the same thing either as an 'interoperability' bugfix for 3.3 (I know, bit of a stretch SemVer-wise), or on 3 branch as a new feature.

Also think it might be beneficial to include the wrapper elements in the includes in case people want to add extra classes to them.

Suggestion from other thread:
```html
<div class="cms-menu cms-panel cms-panel-layout west" id="cms-menu" data-layout-type="border">
		<% include CMSLogo %>
		<% include CMSLoginStatus %>
		<% include CMSMenuList %> // within this have a <% include $CMSMenuItem %> for looping
		<% include CMSMenuToggle %>
</div>
```

tractorcow:
```html
<div class="cms-menu cms-panel cms-panel-layout west" id="cms-menu" data-layout-type="border">
        <% include CMSLogo %>
        <% include CMSLoginStatus %>
        <% include CMSMenuList %> // within this have a <% include $CMSMenuItem %> for looping
        <% include CMSMenuToggle %>
</div>
```

The problem with this is that it requires users overridding the core templates to have to duplicate too much.

The intention is that most users would override the TOP level template, not the sub-templates.

For instance, to add a new element into the top panel, you only have to add a new <%include %> into the `LeftAndMain_Menu.ss`. If you include that outer div in the inner include, you have to copy and duplicate all the html in `LeftAndMain_MenuStatus.ss`, which is a bit larger and is more prone to change over time.

tractorcow:
in any case, I've now rebased this.

tractorcow:
In your example, LeftAndMain_Menu.ss could be overridden by subsites, but LeftAndMain_MenuList.ss could still be overridden by grouped CMS menu module.

jonom:
> The problem with this is that it requires users overriding the core templates to have to duplicate too much.

I don't see how - I'm only talking about including one extra wrapper div per include so that's a minor difference and gives the overrider access to that container element.

> The intention is that most users would override the TOP level template, not the sub-templates.

I think we should encourage the overriding of sub-templates instead, especially with modules, because this limits the impact of a single override. The more include templates you have, the less likely two modules will override the same one so the less likely you're going to get a conflict.

### Idea: Include extension hooks

Here's an idea to make the common need to just 'add extra stuff' easier to do with no risk of conflicts. What if we had extension hooks for Includes. Filename format: `[IncludeName]_[InsertionPoint]_[Identifier]` (InsertionPoint is either 'before' or 'after')

Say you had `<% include MyInclude %>` in a template. The template process would locate `MyInclude.ss` but it would also look for any templates that begin with `MyInclude_before_` and `MyInclude_after_` and render any that are found at either the top or bottom of the main Include. The identifier on the end provides namespacing which would allow you to have multiple before and after includes but also let you override any single one, through the same specificity matching we use for other templates.

So you could have a `CMSMenuList.ss` in CMS module, a `CMSMenuList_before_cms-grouped-menu.ss` template in the cms-grouped-menu module, and a `CMSMenuList_before_subsites.ss` in the subsites module. They won't conflict with one another and the user is free to override either the main include or just the before/after include.

Thoughts?

jonom:
> In your example, LeftAndMain_Menu.ss could be overridden by subsites, but LeftAndMain_MenuList.ss could still be overridden by grouped CMS menu module.

Yes that's idea - we want modules to be able to override only the bits they *need* to override, right?

tractorcow:
> I don't see how - I'm only talking about including one extra wrapper div per include so that's a minor difference and gives the overrider access to that container element.

The problem is when you want to include content BETWEEN the wrapper and the rest of the content, which is 90% of use cases for overriding this template. Now it forces you to re-duplicate the entire inner block.

It would be a lot easier if you could do `parent::` inside templates to nest the template being overridden; I can't see how easy that would be to do though. ;)

tractorcow:
> Idea: Include extension hooks

I thought about that when writing this too, but, so complex. :D

jonom:
> The problem is when you want to include content BETWEEN the wrapper and the rest of the content, which is 90% of use cases for overriding this template. Now it forces you to re-duplicate the entire inner block.

Okay I'm happy - not sure the 90% part is true but I concede that your approach is more pragmatic :D can I just ask for one change then which is to split the `cms-logo` and `cms-login-status` divs out in to separate includes so people can override one or the other instead of having to override both.

Okay 2 changes - can we get this on the 3 branch as well? I can do a separate PR if you want.

Think the extension hooks idea might be worth looking in to to compliment this? If we had that functionality we wouldn't need to have this argument as we could get the best of both worlds :D I don't think it's *that* complex of an idea is it and it would be cool to offer a similar extension pattern to what we do in PHP classes. Maybe worth a mailing list discussion?

tractorcow:
> Okay 2 changes - can we get this on the 3 branch as well? I can do a separate PR if you want.

I think this can be on the 3 branch.

tractorcow:
Replaced with https://github.com/silverstripe/silverstripe-framework/pull/5176

PR #4796, opened 2015-11-26
API Replace CacheGeneratedAssetHandler

tractorcow:
API Replace CacheGeneratedAssetHandler with FlysystemGeneratedAssetHandler
API Reduce GeneratedAssetHandler API
API Re-introduce Requirements::delete_all_combined_files();
API Re-introduce Requirements::flush()
API Combined files now uses new filenames distinguished by sha1 of sources

Replaces https://github.com/silverstripe/silverstripe-framework/pull/4709

This fix resolves a core issue with the original implementation of this feature, in that it was mis-using SS_Cache as an intermediary storage mechanism.

The new solution (after discussion with @sminnee ) is to use a Flysystem instance as a direct dependency of the `GeneratedAssetStore` implementor, allowing the generated asset store to better control the filename, without having to maintain a SHA1 of the resulting file.

This means that we can use `Requirements::delete_all_combined_files` again. :)

Instead of using mtime to detect changes to the file, the combined hash of all source files are used to generate the resulting filename. This should prevent issues with clustered environments where mtimes might not be synchronised perfectly, or touched over time.

Note that this is UNLIKE the `AssetStore` api, which uses the hash of the RESULTING file instead.

CC @caffeineinc on this in case you want to confirm my method here is correct.


tractorcow:
CMS Pull request https://github.com/silverstripe/silverstripe-cms/pull/1337

1337 get :D

view/Requirements.php
```diff
@@ -1482,20 +1515,20 @@ public function getCombinedFilesEnabled() {
...
+		// Get hash based on hash of each file
 		$base = Director::baseFolder() . '/';
-		$srcLastMod = 0;
+		$hash = '';
 		foreach($fileList as $file) {
 			if(file_exists($base . $file)) {
-				$srcLastMod = max(filemtime($base . $file), $srcLastMod);
+				$hash .= sha1_file($base . $file);
 			} else {
 				throw new InvalidArgumentException("Combined file {$file} does not exist");
```
caffeineinc: This doesn't account for dynamically included files, but that may no longer be the case, if so we should probably throw an exception on the include also.

filesystem/flysystem/FlysystemGeneratedAssetHandler.php
```diff
@@ -0,0 +1,107 @@
...
+	/**
+	 * Check if the file exists or that the $callback provided was able to regenerate it.
+	 *
+	 * @param string $filename
+	 * @param callable $callback
+	 * @return bool Whether or not the file exists
+	 * @throws Exception If an error has occurred during save
+	 */
+	protected function checkOrCreate($filename, $callback = null)
+	{
```
chillu: To PSR2 or not to PSR2? ;)

view/Requirements.php
```diff
@@ -1439,16 +1464,24 @@ function() use ($fileList, $minify, $type) {
...
+	 *
+	 * @param string $name
+	 * @param array $files
+	 * @return string
+	 */
+	protected function hashedCombinedFilename($combinedFile, $fileList) {
+		$name = pathinfo($combinedFile, PATHINFO_FILENAME);
+		$hash = $this->hashOfFiles($fileList);
+		$extension = File::get_file_extension($combinedFile);
+		return $name . '-' . substr($hash, 0, 7) . '.' . $extension;
```
chillu: Changing from `myfile.js?m=<hash>` to `myfile-<hash>.js` introduces synchronisation issues on distributed environments which are using `csync` on local filesystems rather than a centralised store like S3. A request to `admin/pages` might be served by instance1, including `<script src="myfile-<hash>.js">`. The browser then requests `myfile-<hash>.js` from instance2 shortly afterwards (<1s). Depending on `csync` latency, this file might not exist there yet. This problem was less likely with `myfile.js?m=<hash>` since the "old" file was served. Of course this could lead to its own issues if the generated HTML doesn't match the served JS or CSS, but at least it didn't crash the CMS completely.

`csync` will become a less common approach in 4.0, but this changeset makes its use more fragile, right?

/cc @caffeineinc

view/Requirements.php
tractorcow: What's a dynamically included file? How can we minify a file that doesn't exist? :P

view/Requirements.php
tractorcow: in this case it should only really be an issue on the cusp of a deployment, correct?

tractorcow: Using a hash also solves the problem of a file being duplicate included if the mtime differs, but the content is the same (since cms js loading treats the different urls as different files, even though only the mtime differs).

It also provides a solution to determining when a file MUST be regenerated, since it's not immediately obvious by looking at the filename whether it is out of date (if the mtime is unreliable).

tractorcow:
Alternative idea; Instead of putting the SHA in the filename, we could add it as a header to the generated file, and ensure that Requirements read that initial line when validating that it was current.


```js
/** SHA1:c5be976b2b0b870198a7ca1ea0f47e12354fe0bc **/
```

However this means that the whole API for "generated" files gets watered down, since it becomes nothing more than a glorified file get / put store, and all the actual business logic ends up in Requirements again as it was before.

view/Requirements.php
chillu: His use case was `tiny_mce_gzip.php?m=1434761806&js=1&plugins=table,emotions,paste,media,fullscreen...`, but we don't run that through `combine_files()` anyway. Cleared up that misunderstanding with @caffeineinc, we think there's no need to support these "dynamically included files" in the new API

chillu:
> in this case it should only really be an issue on the cusp of a deployment, correct?

Assuming that the input files for a `combine_files()` call are usually altered through a deployment, that's true. It might also be caused by an asset snapshot restore or other filesystem-related operations outside of a deployment though.

> Using a hash also solves the problem of a file being duplicate included if the mtime differs

It solves it by breaking the CMS load completely, rather than loading an outdated file - hardly ideal :)

For the record, here's my previous rationale on how to solve the filename hash issue: 
- Assuming that SilverStripe Platform will be an early adopter of S3-backed storage, all SilverStripe Ltd. 4.0 clients wouldn't have this issue
- We can't make that assumption for other larger projects with existing distributed infrastructure based on csync. They might want to switch from 3.x to 4.x faster than the infrastructure can change to a centralised asset store.
- We'll likely precompile core JS files rather than using `combine_files()`, so the issue is much less likely to occur on 4.x.
Conclusion: I think we should explicitly warn upgraders not to use csync in a distributed setup on 4.x, but that doesn't invalidate the filename+hash approach you've proposed

Your new proposal (including the SHA in the file itself) solves the issue in a more robust way. I think this is a better way forward.

> However this means that the whole API for "generated" files gets watered down, since it becomes nothing more than a glorified file get / put store, and all the actual business logic ends up in Requirements again as it was before.

It would still keep the original API for error page generation, right? My assumption is that the Requirements SHA check would be layered on top `FlysystemGeneratedAssetHandler`, which wouldn't need to change.

tractorcow:
> It solves it by breaking the CMS load completely, rather than loading an outdated file - hardly ideal :)

No it doesn't, except during the edge cases which is using the CMS in the middle of a deployment, which is tiny in any case.

In normal use, having slightly different mtimes across instances is disastrous in the long term. What I'm saying is that it prevents that. (unless we want to go back to removing querystring from the ondemand script).

tractorcow:
Ok, I had a chat with @chillu and our solution is four fold:

- Option to configure new combined files directory, in a place where it can be committed to source control (e.g. mysite/javascript). This is already available, but needs documentation.
- Option to disable flushing of combined files.
- Option to revert filename to old system, including the mtime querystring.
- Lots of documentation on how to configure this for your platform configuration.

caffeineinc:
it be cool to include a configurable callback too, but default to sha..

tractorcow:
I've pushed up a new commit that focuses on providing additional advanced configuration options for how requirements behaves, intended for use by users with weird server configurations (csynced distributed filesystems, for instance) and included documentation to guide users.

tractorcow:
I've updated some of the PHPDoc on the new configs. :)

chillu:
Discussed the different `?m=` values with Damian. Just for the record, combined files with `combine_hash_querystring=true` use the combined source hash (64 chars hexadecimal), or the shortened hash as part of the file name. Normally included (uncombined) files on the other hand use the shorter mtime value of the file. We've decided that's an acceptable inconsistency.

PR #4797, opened 2015-11-26
Varchar(50) for "Code" is much too short.

Novusvetus:
I had problems because I came to the 50 character limit with automatically generated permission codes.
I think a longer Varchar should be in the framework, instead of overwriting it in .yml files all the time.


tractorcow:
Good improvement, although please make sure you write your commit messages after what you have actually changed, as these will be used to generate our release changelogs. Since this is an API change it should be tagged as such. E.g. `API Increase Permission.Code db field to 255 characters`

tractorcow:
Merged with https://github.com/silverstripe/silverstripe-framework/commit/1bc7297c8fa7a6cfbb90299793b046d5cdb41e4e

Issue #4798, opened 2015-11-26
Member::validate has changed to protected in 3.2 from public in 3.1


dhensby:
Seems the visibility of the `validate` function has changed to protected in 3.2 which is breach of semver

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-framework/pull/4799

kinglozzer:
Closed via c3544fdbea696833c31e27c9830c77e42e5d823d

PR #4799, opened 2015-11-27
BUG Revert method visibility regression

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/4798


dhensby:
Thanks for sorting!

willmorgan:
Cheers @tractorcow!

tractorcow:
Ironically, this api regression came about because we were trying to revert an API regression. ;)

PR #4806, opened 2015-11-27
DOCS 2.4 : Eliminate hard coded numbers in links and labels

SpiritLevel:
Eliminates hard coding as well as fixes internal links with the aim of reducing the number of .htaccess rewrite rules for doc.silverstripe.org - I am using gurlchecker to eliminate all such errors with internal links.


SpiritLevel:
arrgghh...

PR #4807, opened 2015-11-28
DOCS 2.4 : eliminate hard coded numbers in tutorial labels & links

SpiritLevel:
This removes hard coded numbers as well as fixes internal links, with the aim of eliminating .htaccess rewrite rules for doc.silverstripe.org (as part of a tidy up of the rewrite rules there). Note that gurlchecker was used to identify broken internal links. Also note that multiple PRs will likely result since different doc versions have different internal link problems.


SpiritLevel:
@dhensby @tractorcow Am I doing something wrong here to cause a build fail ? It appears composer needs updating and there exists a database connection problem. I did exactly the same thing with

https://github.com/silverstripe/silverstripe-framework/pull/4808

and it builds properly.

Should I submit PRs to branch 2.4 or branch post-2.4 ?

dhensby:
hmm, I don't know if this is the right branch to be using... 

As 2.4 barely gets any PRs, the test suite is going to be fairly fragile and un-used leading to these kinds of errors. I don't think you've done anything wrong, though

SpiritLevel:
Perhaps we can turn the tests off for this work since I am only changing documentation links and documentation content, and not touching any code ?

tractorcow:
The `$_FILE_TO_URL_MAPPING` config is probably what's causing this issue; A bug fix was merged ages ago that set the visibility of that variable properly.

Maybe we should turn off travis for 2.4 now.

SpiritLevel:
@tractorcow Thanks :) I just checked and I have a $_FILE_TO_URL_MAPPING set in my _ss_environment.php for userhelp.silverstripe.org and doc.silverstripe.org but not framework. Should I add this to my _ss_environment.php ?

tractorcow:
I think the fix was to add a `global` before it, where it wasn't before.

I have this in my web root, but you can use localhost if you don't have a dns pattern as I do. ;P

```php
global $_FILE_TO_URL_MAPPING;
$basepath = dirname(__FILE__);
foreach(scandir($basepath) as $dir) {
	if(preg_match('/^\.(.*)/', $dir)) continue;
	if(!is_dir("$basepath/$dir")) continue;
	$_FILE_TO_URL_MAPPING["$basepath/$dir"] = "http://{$dir}.loc";
}
```

SpiritLevel:
Cool :) Will follow suit.

SpiritLevel:
@tractorcow When the docs are built it looks like the 2.4 branch is used rather than the post-2.4 branch:

https://github.com/silverstripe/doc.silverstripe.org/blob/master/bin/update.sh

So, in fixing all internal links in the 2.4 docs, should I PR against the 2.4 branch instead of the post-2.4 branch ? And, will the above merge into post-2.4 automatically get into 2.4 or do we need to put it there ?

PR #4808, opened 2015-11-28
DOCS 3.0 : eliminate hard coded numbers in tutorial labels & links

SpiritLevel:
This removes hard coded numbers as well as fixes internal links, with the aim of eliminating .htaccess rewrite rules for doc.silverstripe.org (as part of a tidy up of the rewrite rules there). Note that gurlchecker was used to identify broken internal links. Also note that multiple PRs will likely result since different doc versions have different internal link problems.


dhensby:
ping @camfindlay

camfindlay:
@dhensby this looks good to merge.

tractorcow:
@camfindlay the files have been renamed so they don't have the 01_, 02_, etc prefix. Are we not doing this anymore?

camfindlay:
@tractorcow this is an old version _before_ we moved to that convention. The prefixes here were to pseudo do in an old version of docsviewer what the new version does now for sorting.

tractorcow:
Oh, I see.

SpiritLevel:
Thanks...@tractorcow  more coming for 3.0.

PR #4812, opened 2015-11-30
API Enable DataList::sort to support composite field names

tractorcow:
4.0 replacement pull request for https://github.com/silverstripe/silverstripe-framework/pull/4611

Pulling to 4.0 means we can remove DataList::getRelation, and modify the method signature on DataQuery::applyRelation. :)


tractorcow:
ta!

PR #4813, opened 2015-12-01
FIX Use config for determining Vary header

nyeholt:
Existing implementation hardcodes the Vary header; swap to using Config layer
instead

(Now, the philosophical argument - is this a fix to eliminate a hardcoded, untested string, or a new, backwards compatible public API? :D)

### DIFF
--- a/docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
+++ b/docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
@@ -31,3 +31,23 @@ clause in `Cache-Control` and `Pragma` will be included.
 
 Used to set the modification date to something more recent than the default. [api:DataObject::__construct] calls 
 [api:HTTP::register_modification_date(] whenever a record comes from the database ensuring the newest date is present.
+
+### Vary: cache header
+
+By default, SilverStripe will output a `Vary` header (used by upstream caches for determining uniqueness) 
+that looks like
+
+```
+Cookie, X-Forwarded-Protocol, User-Agent, Accept
+```
+
+To change the value of the `Vary` header, you can change this value by specifying the header in configuration
+
+```yml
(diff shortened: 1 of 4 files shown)


tractorcow:
It's a new feature, and I think it'll need to go onto the 3 branch, please. :)

I think it might be better to put the config into a `private static $var`, since it'll then be closer at hand to users reading the code. What do you think?

docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
```diff
@@ -31,3 +31,23 @@ clause in `Cache-Control` and `Pragma` will be included.
...
+
+```
+Cookie, X-Forwarded-Protocol, User-Agent, Accept
+```
+
+To change the value of the `Vary` header, you can change this value by specifying the header in configuration
+
+```yml
+HTTP:
+  vary: ""
```
sminnee: Does this actually work? I know that we've had issues with removing values in this way in the past.

nyeholt: Because it's defined as a string, and the alternate is provided as an explicit empty string, it does - if it was an array of options, it'd just merge the existing array together with the new array and not override anything.

At least, it works that way if calling Config::update() and passing in an empty string in the testcase, and by manual testing with a config.yml in mysite/_config/. If it's not _meant_ to, well - config voodoo :eyes: :)

nyeholt:
No worries Damien, will close this and re-pull against the 3 branch.

sminnee:
> At least, it works that way if calling Config::update() and passing in an empty string in the testcase, and by manual testing with a config.yml in mysite/config/. If it's not _meant to, well - config voodoo 

No, it's meant to, I just know there are some known issues in that area. Glad to hear it works!

PR #4815, opened 2015-12-01
DOCS 2.4 : Fixed all internally generated 404s

SpiritLevel:
@tractorcow....it seems to be ok submitting PR for 2.4 rather than post-2.4 :)


SpiritLevel:
arrgghhh....can't get everything squashed....

PR #4823, opened 2015-12-03
Add getter/setter functions for FormField's dontEscape

mcccclean:
I've got a couple areas in the CMS on a project I'm working on where it makes sense to include a link to another record in amongst the form fields, and it's a pain to have to create an extra holder variable when everything else can be accessed by setters through `ReadOnlyField::create`.

Maybe it'd make more sense in the long term to make some kind of ClickableLinkReadOnlyField? But this at least brings $dontEscape inline with the other properties.

### DIFF
--- a/filesystem/File.php
+++ b/filesystem/File.php
@@ -406,17 +406,16 @@ class File extends DataObject {
 			)->setName("FilePreviewImage")->addExtraClass('cms-file-info-preview'),
 			CompositeField::create(
 				CompositeField::create(
-					new ReadonlyField("FileType", _t('AssetTableField.TYPE','File type') . ':'),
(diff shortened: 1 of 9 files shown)


dhensby:
@tractorcow I'm sure there's already a way to do this?

dhensby:
Why not use a `LiteralField`?

tractorcow:
Yes I think it needs getter/setter.

However, I really dislike the existing naming of this variable.

How about flipping the name though to make it more understandable.

```php
public function setEscapeEnabled($enabled) {
	$this->dontEscape = !$enabled;
}
```

dhensby:
I'm going toclose this as stale - it's been open a long time with no progress from orignal author (sorry @mcccclean)

If you find you're able to implement the feedback, that'd be great and we can reopen this.

PR #4825, opened 2015-12-03
PHP Fatal // member function called on boolean // added is_object()

steiha:
PHP Fatal error: Call to a member function dbObject() on boolean in /var/www/ [...] /public/framework/model/DataObject.php on line 3016

Not quite sure if I just messed up the relation definitions on my DataObject or this is a real bug.
Anyway it might be an improvement to check if $component is an object.

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -3010,7 +3010,11 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 				}
 			}
 
-			$object = $component->dbObject($fieldName);
+			if (is_object($component)){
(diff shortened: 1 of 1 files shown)


tractorcow:
Can you explain how $component ended up as a boolean? I see that this changes the outcome of the error, but it doesn't really explain what's causing it (or more importantly, how to avoid it in the future).

Do you have sample code, or a stacktrace?

steiha:
I'll look into it but probably after the holidays.

dhensby:
Any update?

tractorcow:
No response in a few months; Will close for the time being. :)

PR #4827, opened 2015-12-05
Fixed bug where inheriting an abstract class broke scaffoldFormFields / fieldLabels(true)

SilbinaryWolf:
Fixed bug where inheriting an abstract class broke scaffoldFormFields / fieldLabels(true)


tractorcow:
Good fix; Can we have a unit test to prove this? :P

Maybe we should fix up https://github.com/silverstripe/silverstripe-framework/pull/4828 first, and then from that point on we can write unit tests for both cases for abstract dataobjects?

tractorcow:
I'm going to merge it anyway since it improves the quality of the code.

PR #4828, opened 2015-12-05
Fixed bug where inheriting an abstract class with a DataObject breaks /dev/build

SilbinaryWolf:
Fixed bug where inheriting an abstract class with a DataObject breaks /dev/build

### DIFF
--- a/model/DatabaseAdmin.php
+++ b/model/DatabaseAdmin.php
@@ -228,7 +228,11 @@ class DatabaseAdmin extends Controller {
 		$dbSchema->schemaUpdate(function() use($dataClasses, $testMode, $quiet){
 			foreach($dataClasses as $dataClass) {
 				// Check if class exists before trying to instantiate - this sidesteps any manifest weirdness
-				if(!class_exists($dataClass)) continue;
+				if(!class_exists($dataClass) 
+					|| (($reflection = new ReflectionClass($dataClass)) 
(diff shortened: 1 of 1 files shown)


model/DatabaseAdmin.php
```diff
@@ -228,7 +228,11 @@ public function doBuild($quiet = false, $populate = true, $testMode = false) {
 		$dbSchema->schemaUpdate(function() use($dataClasses, $testMode, $quiet){
 			foreach($dataClasses as $dataClass) {
 				// Check if class exists before trying to instantiate - this sidesteps any manifest weirdness
-				if(!class_exists($dataClass)) continue;
+				if(!class_exists($dataClass) 
+					|| (($reflection = new ReflectionClass($dataClass)) 
+					&& $reflection->isAbstract())) {
```
tractorcow: This actually breaks dev/build in another way; Any `$db` on this abstract class will not be built, although it will be silently ignored.

tractorcow:
I remember trying to solve this issue before in the same way. ;) Unfortunately it's not quite as simple as this.

tractorcow:
I think we need to fix this in master, since we can't properly support abstract data objects without a more significant rewrite. For the time being, mandating non-abstract dataobjects in 3.x will probably stick around. :)

If you want to take this on, feel free, and I'll do my best to review. :D

PR #4830, opened 2015-12-07
API Disable unauthenticated get parameter access to site stage mode

tractorcow:
Also fixes a missing returned SS_HTTPResponse from Security::permissionFailure that this fix relies on.

This code involves an ugly work-around for session state that requires master changes.

Removing `$session` parameter from the `Versioned::choose_site_stage` method is not an api breakage.

### DIFF
--- a/control/VersionedRequestFilter.php
+++ b/control/VersionedRequestFilter.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Initialises the versioned stage when a request is made.
  *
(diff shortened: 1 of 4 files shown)


tractorcow:
Ok, it turns out this PR requires https://github.com/silverstripe/silverstripe-framework/pull/4831 to be merged first since it adds the non_live_permissions config.

chillu:
Peer review passed, merge is dependant on https://github.com/silverstripe/silverstripe-framework/pull/4831/
Confirmed with Damian that the `permissionFailure()` changes look big, but are in fact minimal (removing the `else` branch and returning the `$controller->redirect()` invocation.

PR #4831, opened 2015-12-07
API Create default security permission model for versioned data objects

tractorcow:
This fix moves a lot of stage security into the core of Versioned extension. This will provide a default permission model for any Versioned dataobjects that is a lot more secure than the prior default.

Extending classes can customise this by either overriding canViewVersioned(), or via extensions implementing canViewDraft or canViewArchive

See https://github.com/silverstripe/silverstripe-cms/pull/1341 for replaced functionality


docs/en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -137,8 +137,74 @@ Example: Get the first 10 live records, filtered by creation date:
...
+ 
+If you want to control permissions of an object in an extension, you can also use
+one of the below extension points in your `DataExtension` subclass:
+
+ * `canView` to update the visibility of the object's `canView`
+ * `canViewDraft` to update the visibility of this object only in draft mode.
+ * `canViewArchive` to update the visibility of this object only in archive mode.
+ 
+Note that unlike canViewVersioned, the canViewDraft or canViewArchived will 
+only be invoked if the object is confirmed to be i the appropriate stage, so
```
chillu: Typo, should be "in"

tractorcow: Fixed, thank you. ;D

model/Versioned.php
```diff
@@ -729,6 +737,101 @@ public function onAfterSkippedWrite() {
...
+		$latestVersion = Versioned::get_versionnumber_by_stage($this->owner->class, 'Live', $this->owner->ID);
+		if ($latestVersion == $this->owner->Version) {
+			// Even if this is loaded from a non-live stage, this is the live version
+			return true;
+		}
+
+		// Extend versioned behaviour
+		$extended = null;
+		if($mode === 'stage') {
+			$extended = $this->owner->extendedCan('canViewDraft', $member);
```
hafriedlander: So we're assuming that if mode == stage and the $latestVersion check above failed, we're definitely viewing draft? I wonder if this is reliably true.

model/Versioned.php
```diff
@@ -131,6 +131,14 @@ class Versioned extends DataExtension implements TemplateGlobalProvider {
 	protected static $versionableExtensions = array('Translatable' => 'lang');
 
 	/**
+	 * Permissions necessary to view records outside of the live stage (e.g. archive / draft stage).
+	 *
+	 * @config
+	 * @var array
+	 */
+	private static $non_live_permissions = array('CMS_ACCESS_LeftAndMain', 'CMS_ACCESS_CMSMain', 'VIEW_DRAFT_CONTENT');
```
hafriedlander: The way the check is done, this is an `or`, so any of these permissions would do?

model/Versioned.php
```diff
@@ -729,6 +737,101 @@ public function onAfterSkippedWrite() {
...
+	/**
+	 * Determines canView permissions for the latest version of this object on a specific stage.
+	 * Usually the stage is read from {@link Versioned::current_stage()}.
+	 *
+	 * This method should be invoked by user code to check if a record is visible in the given stage.
+	 *
+	 * Special considerations for this method:
+	 * - This method should not be called via ->extend('canViewStage'), but rather
+	 *   should be, either overridden in the extended class, or extended with
+	 *   updateCanViewStage().
```
hafriedlander: `updateCanViewStage` doesn't appear to be called anywhere.

model/Versioned.php
hafriedlander: There's also lots of situations where are _are_ viewing draft, but this check won't be called (because mode != `stage`). For instance, mode can also be `stage_unique`, `latest_versions`, `all_versions`, etc.

model/Versioned.php
tractorcow: Yep, or is fine.

model/Versioned.php
tractorcow: This documentation was wrong; It should have been a straight re-factor from SiteTree to here.

docs/en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -137,8 +137,69 @@ Example: Get the first 10 live records, filtered by creation date:
 
 ### Permissions
 
-The `Versioned` extension doesn't provide any permissions on its own, but you can have a look at the `SiteTree` class 
-for implementation samples, specifically `canPublish()` and `canDeleteFromStage()`.
+By default, `Versioned` will come out of the box with security extensions which restrict
+the visibility of objects in Draft (stage) or Archive viewing mode.
```
hafriedlander: maybe add "(as long as you call `canView` correctly)"?

tractorcow: I've pushed up an update with a better summary of the issue and resolution. :D

model/Versioned.php
```diff
@@ -729,6 +737,91 @@ public function onAfterSkippedWrite() {
...
+	 * This method should be invoked by user code to check if a record is visible in the given stage.
+	 *
+	 * This method should not be called via ->extend('canViewStage'), but rather should be
+	 * overridden in the extended class.
+	 *
+	 * @param string $stage
+	 * @param Member $member
+	 * @return bool
+	 */
+	public function canViewStage($stage = 'Live', $member = null) {
```
hafriedlander: Whoops :)

tractorcow: No whoops there. :) It's meant to be here!

PR #4837, opened 2015-12-10
Documentation on the release process

tractorcow:
This is the release process we have been developing for releasing CMS and Framework versions. It's not perfect, but it's now public. :) Hopefully this will allow us to be more transparent in our releases.


chillu:
Read through this as part of 3.3.0-beta1 release prep, and it's looking great! Merging so that I can propose some amendments in further PRs :)

PR #4842, opened 2015-12-11
Update icon-button-group css

scott1702:
Icons were not consistently sized/spaced/aligned

Before
image

After 
image


tractorcow:
Before:
<image>

After:
<exactly the same image>

No problems here! Merging.

PR #4845, opened 2015-12-13
API before/afterExtend now support parameters passed by reference

### LINKED ISSUE #4810: Use call_user_func_array in beforeExtendCallbacks (Object Class)
Problem:
Note that the parameters for call_user_func() are not passed by reference.
(http://php.net/call_user_func#example-5664)

example (does not work)
snap_2015 11 30 13 47 49_007
fix
snap_2015 11 30 13 38 23_004

tractorcow:
Closes #4810

### DIFF
--- a/core/Object.php
+++ b/core/Object.php
@@ -970,7 +970,7 @@ abstract class Object {
 
 		if(!empty($this->beforeExtendCallbacks[$method])) {
 			foreach(array_reverse($this->beforeExtendCallbacks[$method]) as $callback) {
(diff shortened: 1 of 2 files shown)


tractorcow:
3.3 is in beta, so this has to go into 3.4

dhensby:
PHP 5.6:

```
function foo(&...$args) {
    $args[0] = 'bar';
}
```

tractorcow:
That's cool @dhensby :D

PR #4847, opened 2015-12-14
NEW: Implement unshift() in field list classes (closes #4834)

### LINKED ISSUE #4834: FieldList, CompositeField, TabSet etc should implement unshift()
We have `ArrayList::unshift()` that could be useful in the context of forms/fields.

`FieldList` will currently *mostly* work, because it inherits it from `ArrayList::unshift()`, but it should call the same extra methods that `FieldList::push()` does.

kinglozzer:


forms/FieldList.php
```diff
@@ -412,16 +412,27 @@ public function insertAfter($name, $item) {
...
 
 	/**
-	 * Push a single field into this FieldList instance.
+	 * Push a single field onto the end of this FieldList instance.
 	 *
 	 * @param FormField $item The FormField to add
-	 * @param string $key An option array key (field name)
 	 */
-	public function push($item, $key = null) {
+	public function push($item) {
```
kinglozzer: No BC issues here - this param isn’t used (in this class, or in the parent `ArrayList` class), and won’t error if passed anyway or added in a subclass

tractorcow: Hm, it looks ok to me.... parent::push() doesn't have a second parameter anyway.

forms/FieldList.php
```diff
@@ -412,16 +412,27 @@ public function insertAfter($name, $item) {
...
-		return parent::push($item, $key);
+		return parent::push($item);
+	}
+
+	/**
+	 * Push a single field onto the beginning of this FieldList instance.
+	 *
+	 * @param FormField $item The FormField to add
+	 */
+	public function unshift($item) {
```
tractorcow: Why not use a strict typehint as you did in FormField::unshift?

dhensby:
Can we get this as part of a list interface for master, please?

tractorcow:
We'll have to change the typehint from `FormField` to `mixed` but I agree it's good addition.

PR #4848, opened 2015-12-14
FIX: Allow omitting a value for OptionsetField submissions (fixes #4284)

kinglozzer:


forms/OptionsetField.php
```diff
@@ -90,31 +88,15 @@ public function Field($properties = array()) {
 		);
 	}
 
-	public function performReadonlyTransformation() {
```
tractorcow: Why did this API get dumped here?

kinglozzer: Second commit - it’s an exact duplicate of `DropdownField:: performReadonlyTransformation()`

tractorcow:
It looks like this needs to go into DropdownField as well, maybe.

PR #4851, opened 2015-12-15
Removed validation from LookupField as it's readonly

SilbinaryWolf:
Removed validation from LookupField as it's readonly anyway and is preventing me from cleanly just showing the user a ManyManyList due to the data not matching the $source.

### DIFF
--- a/forms/LookupField.php
+++ b/forms/LookupField.php
@@ -80,13 +80,21 @@ class LookupField extends DropdownField {
 	}
 
 	/**
-	 * Validate this field
+	 * Ignore validation as the field is readonly 
 	 *
 	 * @param Validator $validator
 	 * @return bool
 	 */
 	public function validate($validator) {
 		return true;
+	}
+
+	/**
+	 * Stubbed so invalid data doesn't save into the DB
+	 *
+	 * @param DataObjectInterface $record DataObject to save data into
+	 */
+	public function saveInto(DataObjectInterface $record) {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


tractorcow:
Is there a chance that users could exploit this to save invalid data into an otherwise readonly field? What happens if we put an `<input type="hidden" name="thefield" value="someinvalidcontent" />` and try to post it; Does it get blocked? :P

Maybe we should override saveInto() to block any value (valid or not) getting saved back into the dataobject.

SilbinaryWolf:
Done and done. 
Though now I wonder, if it's worth just changing FormField to not save if $readonly is set to true. Are there cases where someone may want a readonly FormField to save into their DB?

tractorcow:
Do you feel confident in writing a unit test? Just to prove that forms with lookup fields don't overwrite (or clear) dataobjects, and that forms with invalid lookup values pass validation. ;)


```
$form = new TestForm();
$form->addLookupField();
$form->loadDataFrom($somePostedContent);
$form->saveInto($object);
$this->assertAllGood($object);
```

and 

```php
$form = new TestForm();
$form->addLookupField();
$form->loadDataFrom($somePostedContent);
$this->assertTrue($form->validate());
```

tractorcow:
I think this can be merged in.

SilbinaryWolf:
Sorry for the lack of reply, getting slammed at work and I haven't needed to write Unit Tests before, so it's something I can look at over the Christmas break potentially.

tractorcow:
I want to get this into the 3.3.0 beta, so you can still write that test if you get time. :)

Thanks for this and your other contributions... hope you get done in time for christmas. :)

PR #4852, opened 2015-12-15
Fixing deselect failing to change field title

### LINKED ISSUE #4240: Cannot Deselect Last Item in TreeMultiselectField
When deselecting the last item from a TreeMultiselectField, despite the checkbox becoming unticked, the item remains as the field's value.

**To reproduce:**
1. Create a TreeMultiselectField
2. In the CMS open the drop-down and tick more than one item. Each ticked item appears in the field's value.
3. Untick an item and it is correctly removed from the field's value.
4. Untick the last item.

**What happens:**
Despite being unticked, it's value remains listed in the field.

**What I expect to happen:**
The name of the last item to be removed from the field's value, and the default "(Choose)" text to appear, indicating that no items have been selected.

dhensby:
fixes #4240

- Fixes missing empty title value for `TreeMultiselectField`
- Uses empty value when no value is passed to the field

### DIFF
--- a/forms/TreeMultiselectField.php
+++ b/forms/TreeMultiselectField.php
@@ -103,6 +103,7 @@ class TreeMultiselectField extends TreeDropdownField {
 		$titleArray = array();
 		$idArray = array();
 		$items = $this->getItems();
+		$emptyTitle = _t('DropdownField.CHOOSE', '(Choose)', 'start value of a dropdown');
 
 		if($items && count($items)) {
 			foreach($items as $item) {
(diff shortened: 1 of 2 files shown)


tractorcow:
Can you use the FIX or BUG tags in future for bugfix commits. :)

Apparently this is the regexp we've been using to detect bugs for changelog purposes.

`'/^(BUGFIX|BUGFUX|BUG|FIX)\s?:?/'`

stevie-mayhew:
BUGFUX :laughing:

tractorcow:
That is actually there yes.....

kinglozzer:
Perhaps that’s just how Kiwis pronounce it? :wink:

dhensby:
@tractorcow when you press the merge button you can name the merge commit what you like ;)

that's what I do for bugs - will try to remember to do this from now on :)

tractorcow:
Oh, I never thought about that! Good idea.

PR #3958, opened 2015-03-02
Update SapphireTestReporter.php

azt3k:
- Updates code to be compatible with PHP 5.0.2
- Allows the autoloader to be called so as composer can be used (PHPUnit PEAR install is now deprecated)

http://php.net/manual/en/function.interface-exists.php
http://php.net/manual/en/function.class-exists.php
https://github.com/sebastianbergmann/phpunit/wiki/End-of-Life-for-PEAR-Installation-Method

PR #4856, opened 2015-12-17
BUG Fix SapphireTestReporter erroring

tractorcow:
This class hasn't been updated much, so I've given it a general code quality bump.

There are some odd edge cases that occur if an error, failure, or notice is raised when there is no current suite or test being run, so rather than creating an invalid state, ensure that these are logged to a fallback store ($currentSession).

I suspect that these occur during setup / teardown methods when a test or suite may not be available, or during certain bootstrapping or exit conditions.

### DIFF
--- a/dev/SapphireTestReporter.php
+++ b/dev/SapphireTestReporter.php
@@ -1,9 +1,7 @@
 <?php
-if(!class_exists('PHPUnit_Framework_TestResult', false)) {
-	require_once 'PHPUnit/Framework/TestResult.php';
-}
-if(!class_exists('PHPUnit_Framework_TestListener', false)) {
-	require_once 'PHPUnit/Framework/TestListener.php';
+
+if(!interface_exists('PHPUnit_Framework_TestListener')) {
+	return;
 }
 
 /**#@+
--- a/dev/CliTestReporter.php
+++ b/dev/CliTestReporter.php
@@ -1,4 +1,9 @@
 <?php
+
+if(!class_exists('SapphireTestReporter')) {
+	return;
+}
+
 /**
  * Test reporter optimised for CLI (ie, plain-text) output
  *


dev/SapphireTestReporter.php
```diff
@@ -1,6 +1,10 @@
 <?php
-if(!class_exists('PHPUnit_Framework_TestResult', false)) require_once 'PHPUnit/Framework/TestResult.php';
-if(!class_exists('PHPUnit_Framework_TestListener', false)) require_once 'PHPUnit/Framework/TestListener.php';
+if(!class_exists('PHPUnit_Framework_TestResult', false)) {
```
dhensby: Can't we get rid of this now? Composer and so on...

tractorcow: phpunit could have been installed globally, instead of via composer. In which case, it could have been declared in the global include paths, couldn't it?

dhensby: It could have been except:

1. PHPUnit doesn't support PEAR installs any more; if it's installed globally as a phar then I doubt we can extend it like this anyway, which means...
1. It'll be installed by composer to get this to work anyway
1. This is master, we can break what we want

dhensby: see https://github.com/silverstripe/silverstripe-framework/pull/3958 as a half-way-house (allowing the autoloader to find the class)

Personally, I'd rather just see it gone

dhensby:
FYI - I deleted those `class_exists` checks and associated `require_once` calls and upgraded `PHPUnit` to ~4.8@stable and all the tests passed no problem:

```
[root@localhost html]# ./vendor/bin/phpunit framework/tests
PHPUnit 4.8.21 by Sebastian Bergmann and contributors.

.............................................................   61 / 1673 (  3%)
......................................I......................  122 / 1673 (  7%)
..............................I..............................  183 / 1673 ( 10%)
........I....................................................  244 / 1673 ( 14%)
.............................................................  305 / 1673 ( 18%)
.............................................................  366 / 1673 ( 21%)
.............................................................  427 / 1673 ( 25%)
.............................................................  488 / 1673 ( 29%)
.............................................................  549 / 1673 ( 32%)
.....................................................I.......  610 / 1673 ( 36%)
.............................................................  671 / 1673 ( 40%)
.............................................................  732 / 1673 ( 43%)
.............................................................  793 / 1673 ( 47%)
.............................................................  854 / 1673 ( 51%)
....I........................................................  915 / 1673 ( 54%)
.............................................................  976 / 1673 ( 58%)
............................................................. 1037 / 1673 ( 61%)
............................................................. 1098 / 1673 ( 65%)
............................................................. 1159 / 1673 ( 69%)
.....................................................SSSSSSSS 1220 / 1673 ( 72%)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSS............................... 1281 / 1673 ( 76%)
................SS........................................... 1342 / 1673 ( 80%)
............................................................. 1403 / 1673 ( 83%)
............................................................. 1464 / 1673 ( 87%)
............................................................. 1525 / 1673 ( 91%)
................I............................................ 1586 / 1673 ( 94%)
............................................................. 1647 / 1673 ( 98%)
..........................

Time: 11.05 minutes, Memory: 167.50Mb

OK, but incomplete, skipped, or risky tests!
Tests: 1673, Assertions: 6853, Skipped: 40, Incomplete: 6.
[root@localhost html]# ./vendor/bin/phpunit cms/tests/
PHPUnit 4.8.21 by Sebastian Bergmann and contributors.

.........I.....................................................  63 / 202 ( 31%)
............................................................... 126 / 202 ( 62%)
............................................................... 189 / 202 ( 93%)
.............

Time: 2.61 minutes, Memory: 100.25Mb

OK, but incomplete, skipped, or risky tests!
Tests: 202, Assertions: 873, Incomplete: 1.
[root@localhost html]# ./vendor/bin/phpunit reports/tests/
PHPUnit 4.8.21 by Sebastian Bergmann and contributors.

....

Time: 7.96 seconds, Memory: 74.75Mb

OK (4 tests, 26 assertions)
[root@localhost html]# ./vendor/bin/phpunit siteconfig/tests/
PHPUnit 4.8.21 by Sebastian Bergmann and contributors.

.....

Time: 7.63 seconds, Memory: 74.25Mb

OK (5 tests, 10 assertions)
```

dev/SapphireTestReporter.php
tractorcow: Ok, I'll gone it.

tractorcow:
Since we are using vendor/bin/phpunit, I guess it's ok to rely on composer to autoload.

tractorcow:
Why when I rebase against master does my PR disappear...

tractorcow:
Oh lol,

It looks like I accidentally merged this commit into my Asset rewrite PR. 

https://github.com/silverstripe/silverstripe-framework/commit/037467beae90b92598500c27ecfe5d1fcbfea883#diff-e4650b325b1c715b1515476c9b8654d1

tractorcow:
@dhensby I've removed the includes, and now we rely purely on composer autoloading.

If phpunit isn't installed (e.g. if require-dev is excluded) then the classes return safely.

dhensby:
What benefit is there to returning in stead of a PHP error... This is a dev class. Isn't the error that the extended class doesn't exist better than a false assertion that the testlistener can't be found?

tractorcow:
It's not an error though, it just means the user isn't running dev tools. ClassManifest will still hit it, won't it?

dhensby:
Wouldn't PHP throw the wrong error. It'll say the testrunner class can't be found in stead of that the PHPUnit dependency is missing (which is more helpful)

tractorcow:
Yeah but it wouldn't do that unless you were trying to run tests without require-dev installed. In which case, you should get an error. I'd rather the error there then than in SS_ClassManifest though. ;P

PR #4857, opened 2015-12-17
DOCS : Update 02_Composer.md gdmedia/ss-auto-git-ignore as dev requirement

ivoba:
gdmedia/ss-auto-git-ignore is rather a dev requirement, isnt it?

### DIFF
--- a/docs/en/00_Getting_Started/02_Composer.md
+++ b/docs/en/00_Getting_Started/02_Composer.md
@@ -140,7 +140,7 @@ This package will maintain your [.gitignore](http://git-scm.com/docs/gitignore) 
 
 Include the package in your project by running this command
 
-    composer require gdmedia/ss-auto-git-ignore
+    composer require gdmedia/ss-auto-git-ignore --dev
 
 Edit your composer.json and insert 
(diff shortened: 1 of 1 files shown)


docs/en/00_Getting_Started/02_Composer.md
```diff
@@ -140,7 +140,7 @@ This package will maintain your [.gitignore](http://git-scm.com/docs/gitignore)
 
 Include the package in your project by running this command
 
-    composer require gdmedia/ss-auto-git-ignore
+    composer require gdmedia/ss-auto-git-ignore --dev
```
tractorcow: I think --dev is implied by default now.

ivoba: I think this is not for require. how would you add a prod dependency then?
for update and install its the other way around, there dev ist the default and you need to use --no-dev to exclude dev deps.

tractorcow: Hah, you're right. I've conflated the `update --dev` with `require --dev`. :)

PR #4863, opened 2015-12-21
API Asset Access Control implementation

tractorcow:
For the background on this feature, please see https://groups.google.com/forum/#!topic/silverstripe-dev/PihYhP3ZTN8

This implementation provides a core security API for assets stored via the default backend, and a low level mechanism for securing, publishing, granting, and denying access to individual files.

The key mechanism is implemented via a pair of Flysystem filesystem instances, one of which allows direct web traffic, and the other which is only accessible via a silverstripe controller, which moderates secure access.

A new default extension `AssetControlExtension` is now applied to DataObject, which provides a default interface between dataobject operations and the security of linked assets. This class will be significantly updated in a subsequent pull request with the following features:

- Publication of linked assets when versioned dataobjects are published
- Deletion of old files when replaced with new
- Ensure new draft versioned dataobjects store files in the protected store by default
- Ensure that unversioned or live dataobjects with non-default security permissions also store files in the protected store


chillu:
So excited we got this feature into pull request stage, thanks for your massive efforts Damian! :)
I've peer reviewed most of this already in detail while it was built, keen on other opinions and reviews of course. 

Have you considered increasing the PHP execution time based on file download size? secureassets module does that: https://github.com/silverstripe-labs/silverstripe-secureassets/blob/6120a21ab735908156bd0d62d1e423bb2db9569f/code/SecureFileController.php#L87

Also, do we need to worry about PHP request blocking? https://github.com/silverstripe-labs/silverstripe-secureassets/blob/6120a21ab735908156bd0d62d1e423bb2db9569f/code/SecureFileController.php#L101

Have we considered (optional) `X-Sendfile` support to get around some of the performance limitations? http://blog.jasny.net/articles/how-i-php-x-sendfile/

@micmania1 You've raised performance concerns on this approach previously, want to check if you're happy with it?

tractorcow:
Regarding all of the above, I think that we will need to implement these functions at some point before stable.

I've been looking at creating a streamable subclass of SS_HTTPResponse, which is able to wrap around a Flysystem stream as an output. This will allow us to retain streaming response support, even for cases where we are not using a local filesystem (e.g. streaming from a secure S3 bucket).

It could possible make use of http://php.net/manual/en/httpresponse.setstream.php

Unfortunately sendfile won't work for all backends, so we'll need to come up with a solution that is flexable. I'm hesitant to de-abstract the filesystem too much in order to resolve all assets to local file paths.

So, in short, yes, but maybe later? :P

filesystem/flysystem/FlysystemAssetStore.php
```diff
@@ -436,4 +735,91 @@ protected function getFileID($filename, $hash, $variant = null) {
...
+	 * @param Filesystem $flysystem
+	 * @param string $fileID
+	 * @return SS_HTTPResponse
+	 */
+	protected function createResponseFor(Filesystem $flysystem, $fileID) {
+		// Build response body
+		// @todo: gzip / buffer response?
+		$body = $flysystem->read($fileID);
+		$mime = $flysystem->getMimetype($fileID);
+		$response = new SS_HTTPResponse($body, 200);
```
micmania1: If we pass the file back through the framework (ie. SS_HTTPResponse) it'll store the entire file in memory before streaming it back. 

eg. 100MB file
$body stores 100MB in memory
$response->body then copies that 100MB into memory (now at 200MB for a single request).
After this its passed off to postRequestFilters which *could* create even more copies of $response->body

Does Flysystem support streaming directly back to the client (something like stream_copy_to_stream())?

tractorcow:
I've pushed up a fix to documentation, as well as a few bugfixes I noted during testing.

filesystem/flysystem/FlysystemAssetStore.php
tractorcow: Yep, but SS_HTTPResponse doesn't support streams as a source, hence my suggestion that we'll need to subclass it.

tests/filesystem/ProtectedFileControllerTest.php
```diff
@@ -0,0 +1,201 @@
...
+			array('name\.jpg', false)
+		);
+	}
+
+	/**
+	 * Test that certain requests are denied
+	 */
+	public function testRequestDenied() {
+		$result = $this->get('assets/.protected/file.jpg');
+		$this->assertResponseEquals(400, null, $result);
```
micmania1: Unless this is 404 then it could be considered as information disclosure. Maybe just get a few opinions on what response should be given? @chillu @hafriedlander

filesystem/flysystem/FlysystemAssetStore.php
```diff
@@ -137,7 +265,144 @@ public function setFromStream($stream, $filename, $hash = null, $variant = null,
...
+	 * Determine if a grant exists for the given FileID
+	 *
+	 * @param string $fileID
+	 * @return bool
+	 */
+	protected function isGranted($fileID) {
+		// Since permissions are applied to the non-variant only,
+		// map back to the original file before checking
+		$originalID = $this->removeVariant($fileID);
+		$granted = Session::get('AssetStore_Grants') ?: array();
```
micmania1: self::GRANTS_SESSION :)

filesystem/flysystem/FlysystemAssetStore.php
micmania1: That sounds good.

filesystem/flysystem/FlysystemAssetStore.php
tractorcow: Ah good spotting. :D

filesystem/flysystem/FlysystemAssetStore.php
```diff
@@ -436,4 +735,91 @@ protected function getFileID($filename, $hash, $variant = null) {
 		return $fileID;
 	}
 
+	/**
+	 * Ensure each adapter re-generates its own server configuration files
+	 */
+	public static function flush() {
+		// Ensure that this instance is constructed on flush, thus forcing
+		// bootstrapping of necessary .htaccess / web.config files
+		$instance = singleton('AssetStore');
```
micmania1: I'm not clear when this flush method is called. Is it definitely called **after** the config has been re-generated? Just making sure that if you change AssetStore then flush, you don't get the old value here (which could be null and allow public access to everything).

docs/en/00_Getting_Started/01_Installation/How_To/Configure_Nginx.md
```diff
@@ -38,7 +38,7 @@ But enough of the disclaimer, on to the actual configuration — typically in `n
 	
 		location ^~ /assets/ {
 			sendfile on;
-			try_files $uri =404;
+			try_files $uri /framework/main.php?url=$uri&$query_string;
```
micmania1: I don't think this would block direct access to the .protected folder.

docs/en/02_Developer_Guides/14_Files/03_File_Security.md
```diff
@@ -0,0 +1,375 @@
...
+
+For instance, this will allow your nginx site to serve files directly, while ensuring
+dynamic requests are processed via the Framework:
+
+
+    :::
+    location ^~ /assets/ {
+        sendfile on;
+        try_files $uri /framework/main.php?url=$uri&$query_string;
+    }
```
micmania1: As above with nginx rule - this won't block direct access to .protected.

templates/filesystem/Assets_HTAccess.ss
```diff
@@ -0,0 +1,27 @@
...
+    # Disable PHP handler
+    RewriteCond %{REQUEST_URI} .(?i:php|phtml|php3|php4|php5|inc)$
+    RewriteRule .* - [F]
+
+    # Allow error pages
+    RewriteCond %{REQUEST_FILENAME} -f
+    RewriteRule error[^\\/]*\.html$ - [L]
+
+    # Block invalid file extensions
+    RewriteCond %{REQUEST_URI} !\.(?i:<% loop $AllowedExtensions %>$Extension<% if not $Last %>|<% end_if %><% end_loop %>)$
```
micmania1: This will break for extensions like .tar.gz. It needs some kind of escaping.

filesystem/flysystem/FlysystemAssetStore.php
tractorcow: It's called well after that. :) It's a part of the RequestFilter API, so it's safe now.

docs/en/00_Getting_Started/01_Installation/How_To/Configure_Nginx.md
tractorcow: I'll see if I can come up with a fix.

templates/filesystem/Assets_HTAccess.ss
tractorcow: tar.gz isn't a valid allowed extension. In this case you could configure 'gz' but not 'tar.gz'.

tractorcow: Oh, actually, I am escaping it anyway. :) If you check AssetAdapter.php you'll see I'm doing preg_quote.

image

micmania1:
Okay, so here's what I think I understand for clarity and a few questions. Hopefully to help anybody else get to grips with this a little faster...

### Things I think I understand...
* All public files can be accessed directly without needing to hit SilverStripe. This is done by a dynamically generated .htaccess in the assets folder which allows direct access to files with accepted file types.
* Direct access to protected files are blocked by a dynamically generated .htaccess in the .protected folder.
* Requests to private files are passed to SilverStripe where canView

### Questions...
* Why are permissions granted in sessions? (I assume this is something to do with access through DataObject?)
* Is the session data ever cleared for permissions (is this expected to grow large)?
* Can the protected folder be outside of the web root by default (if permissions allow)? I would imagine this to work like the silverstripe-cache folder when the /tmp folder isn't writable.

tractorcow:
> Why are permissions granted in sessions? (I assume this is something to do with access through DataObject?)

Because otherwise we'd have to start storing temporary data in the database, and it would have to link back to those users in any case.

There are also situations where you MAY allow some protected assets to users who are not logged in with a Member dataobject (e.g. unsecured draft site), so you can't rely on a member dataobject to identify those users.

> Is the session data ever cleared for permissions (is this expected to grow large)?

Sessions should expire just per normal PHP sessions. We only store one record for each file, regardless of the number of variants that file has, so it's kept as small as possible.

> Can the protected folder be outside of the web root by default (if permissions allow)? I would imagine this to work like the silverstripe-cache folder when the /tmp folder isn't writable.

That would be nice, but it would interfere with SilverStripe's standard "works out of the box". I've been intentionally careful to avoid breaking the experience of any first time silverstripe devs.

tests/filesystem/ProtectedFileControllerTest.php
chillu: I struggle to come up with a realistic attack scenario here, and the distinction between a 400 and 404 will make it much easier for developers to debug this quite complex piece of (session dependant) logic. I'd vote to keep it a 400.

docs/en/02_Developer_Guides/14_Files/03_File_Security.md
chillu: In order to get this PR over the line, I'd probably avoid mentioning any nginx specific config in the first place. Could be a community contributed addition? cc @oddnoc

docs/en/02_Developer_Guides/14_Files/03_File_Security.md
```diff
@@ -0,0 +1,375 @@
...
+    $store = singleton('AssetStore');
+    $store->setFromString('My protected content', 'Documents/Mydocument.txt', null, null, array(
+        'visibility' => AssetStore::VISIBILITY_PROTECTED
+    ));
+
+
+## User access control
+
+Access for files is granted on a per-session basis, rather than on a per-member basis, via a
+whitelist action. This means that access to any protected asset must be made prior to the user
```
chillu: Suggest reword to "via whitelisting accessed assets" (since Michael didn't pick up on this as per his comments)

tests/filesystem/ProtectedFileControllerTest.php
chillu: Talked to @micmania1 a bit, here's a scenario (a bit far fetched, but not impossible): `assets/2014/staff_layoffs.pdf` could lead to somebody guessing there's upcoming layoffs by trying `assets/2016/staff_layoffs.pdf` (even though that document isn't published yet). So 404 seems safer

tractorcow: I think you are looking at it the wrong way; It returns a 400 if the REQUEST is invalid. That is, it contains a `.name` directory.

It doesn't check if the file exists, since it is refusing to process an invalid request. :)

tractorcow:
Regarding the disclosure of secure assets, I will make a config to set the reporting mechanism for secure files, which would otherwise be a 403. As per "secure by default" standard, we can make it 404, but users should be able to set it to 403 via config, which is (ideally) more "correct".

This code is a different method and test to the one commented on above, however, hence my earlier comment. :)

filesystem/flysystem/FlysystemAssetStore.php
```diff
@@ -436,4 +735,91 @@ protected function getFileID($filename, $hash, $variant = null) {
...
+		return $this->createResponseFor($filesystem, $asset);
+	}
+
+	/**
+	 * Generate an {@see SS_HTTPResponse} for the given file from the source filesystem
+	 * @param Filesystem $flysystem
+	 * @param string $fileID
+	 * @return SS_HTTPResponse
+	 */
+	protected function createResponseFor(Filesystem $flysystem, $fileID) {
```
lcobucci: You should depend on the ```FilesystemInterface``` instead of ```Filesystem```. Easier to mock, easier to change.

tests/filesystem/AssetStoreTest.php
```diff
@@ -1,10 +1,12 @@
 <?php
```
lcobucci: This whole test could use something like [vfsStream](Filesystem) to don't write/read directly from the filesystem (which is not a good practice in software testing)

tests/filesystem/ProtectedFileControllerTest.php
lcobucci: **400** should be used when the request couldn't be understood by the server, which is clearly not the case (at least according with the test case name). Unless I got everything wrong this case is where the file exists however the user don't have access to it (even if it's logged), so it should be a **403**.

http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4

lcobucci: If it's not that, the test method should probably be renamed :smile:

tractorcow: > 400 should be used when the request couldn't be understood by the server

The server cannot understand requests for assets that have a dot-prefixed directory in it. :)

filesystem/flysystem/FlysystemAssetStore.php
tractorcow: Oh, excellent idea. Thanks for the feedback!

filesystem/flysystem/FlysystemAssetStore.php
tractorcow: @sminnee suggested looking at implementing http://www.php-fig.org/psr/psr-7/ at the same time.

docs/en/02_Developer_Guides/14_Files/03_File_Security.md
tractorcow: I tested the below and it seems to do the trick.

```
location ^~ /assets/ {
	location ~ /\. {
		deny all;
	}
	sendfile on;
	try_files $uri /framework/main.php?url=$uri&$query_string;
}
```

tractorcow:
I've rebased against 4.0 and addressed a few things:

- Proper nginx rules, which I've tested to ensure they block .protected. :)
- Documentation tweaks as suggested
- Renamed ProtectedFileController::testRequestDenied to testInvalidRequest, since the method was named incorrectly.
- Prevent disclosure of secure assets by reporting denied files as 404. This response can be configured by altering the FlysystemAssetStore.denied_request_code to another value.
- Replaced a few Filesystem references with FilesystemInterface, at least where we were able to.

To address a few other things:
- I'd like to implement vfsStream, but I need to spend a bit more time to understand and test it.
- We are planning to implement streamed files, but will be doing so at a later date. We'd like this work to be based around, potentially, implementing http://www.php-fig.org/psr/psr-7/

Is anything else blocking merge?

chillu:
There's been lots of opportunities for feedback (from the RFC to mailinglist updates as the feature developed), plus nearly a month on this pull request. Damian has been very diligent in developing this feature, and responding to feedback. Given all of this, I don't think anything else is blocking merge. Congratulations Damian, great work!

tractorcow:
Thanks @chillu :D

PR #4864, opened 2015-12-21
Updated RestfulService to include PATCH

robpetreski:
RestfulService now allows PATCH and also removed CURLOPT_POST


tractorcow:
Will merge once tests pass.

PR #4865, opened 2015-12-22
BUG Change Requirements::include_in_response() to not add empty 'X-Include-JS' or 'X-Include-CSS' headers

jacobbuck:


tractorcow:
Thanks for the fix!

PR #4867, opened 2015-12-22
BUG Workaround for issues in testing versioned

tractorcow:
Using `return false` meant that cms tests couldn't test that the returned response couldn't be examined, so adjusted the behaviour slightly for tests.

### DIFF
--- a/control/VersionedRequestFilter.php
+++ b/control/VersionedRequestFilter.php
@@ -32,7 +32,7 @@ class VersionedRequestFilter implements RequestFilter {
 			$dummyController->popCurrent();
 			// Prevent output in testing
 			if(class_exists('SapphireTest', false) && SapphireTest::is_running_test()) {
-				return false;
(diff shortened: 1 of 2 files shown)


control/VersionedRequestFilter.php
```diff
@@ -32,7 +32,7 @@ public function preRequest(SS_HTTPRequest $request, Session $session, DataModel
 			$dummyController->popCurrent();
 			// Prevent output in testing
 			if(class_exists('SapphireTest', false) && SapphireTest::is_running_test()) {
-				return false;
```
sminnee: Happy to merge as-is, but my only comment here would be "can we make this the code execution path for regular requests as well as tests?"

tractorcow: I think in 4.x we could probably push up the "testing case exemption" up to director. If we allow request filters to throw SS_HTTPResponses normally, it wouldn't be a problem, but that would be an API breakage in 3.x.

PR #4869, opened 2015-12-22
FIX: Parameters passed to includes overwrite all scopes (fixes #2617)

### LINKED ISSUE #2617: <% include %> parameters ignore scope / overwrite variables inside loop
`<% include Foo %>` allows to pass parameters, but those parameters seem to ignore scopes.
in the following example you can that event though `$Title` is inside a `<% loop %>` the result is the `$Title` of the parent scope. `$getTitle` however actually returns the value of the DataObject.

    // code/Page.php
    class Page extends SiteTree {
        private static $db = array(
            'ListTitle' => 'Varchar',
        );
        private static $has_many = array(
            'ListItems' => 'ListItem',
        );
    }

    // code/ListItem.php
    class ListItem extends DataObject {
        private static $db = array(
            'Title' => 'Varchar',
        );
            'Page' => 'Page',
        );
    }

    // templates/Page.ss
    ...
    <% include ListBox Title=$ListTitle,Items=$ListItems %>
    ....

    // templates/Includes/ListBox.ss
    <h3>$Title</h3>
    <ul>
        <% loop $Items %>
            <li>
                $Title // returns Page->ListTitle
                $getTitle // returns in ListItem->Title
            </li>
        <% end_loop %>
    </ul>

kinglozzer:


dhensby:
Interesting. I want to review on desktop

dhensby:
Needs a rebase too

view/SSViewer.php
```diff
@@ -23,10 +23,18 @@
  * @subpackage view
  */
 class SSViewer_Scope {
+
+	const ITEM = 0;
```
tractorcow: Very good. :)

kinglozzer:
Rebased :) probably easier to review the two commits separately than look at the main diff

view/SSViewer.php
```diff
@@ -24,9 +24,17 @@
...
+	const POP_INDEX = 3;
+	const UP_INDEX = 4;
+	const CURRENT_INDEX = 5;
+	const ITEM_OVERLAY = 6;
+	
 	// The stack of previous "global" items
-	// And array of item, itemIterator, itemIteratorTotal, pop_index, up_index, current_index
-	private $itemStack = array();
+	// An indexed array of item, item iterator, item iterator total, pop index, up index, current index & parent overlay
+	protected $itemStack = array();
```
dhensby: Don't change the visibility, create a setter and getter func :)

view/SSViewer.php
```diff
@@ -519,6 +527,64 @@ public function getInjectedValue($property, $params, $cast = true) {
...
+	 * $Up and $Top need to restore the overlay from the parent and top-level
+	 * scope respectively.
+	 */
+	public function obj($name, $arguments = null, $forceReturnedObject = true, $cache = false, $cacheName = null) {
+		$overlayIndex = false;
+
+		switch($name) {
+			case 'Up':
+				if ($this->upIndex === null) {
+					user_error('Up called when we\'re already at the top of the scope', E_USER_ERROR);
```
dhensby: Is this a new error? Do we really want to throw fatal errors due to template issue?

Would this prevent `<% if $Up %>`?

kinglozzer: It’s the same error as [`SSViewer_Scope::obj()`](https://github.com/silverstripe/silverstripe-framework/blob/3ee8f505b75af39817d394bcf48623e092f43ed2/view/SSViewer.php#L98). In fact, I think might make the original `user_error()` redundant - I was just reluctant to remove it in case some code (core or otherwise) might bypass `SSViewer_DataPresenter`...

It stops you from trying to do `$Up.Title` where there is no parent scope.

view/SSViewer.php
kinglozzer: Good idea

view/SSViewer.php
dhensby: OK - if it's an existing error, then fine... but I think that's a bit harsh when includes could be nested at some unknown level, so stopping access to $Up by way of error instead of just a warning and continuing seems a bit harsh/fragile

dhensby:
cool - I'm a fan of this :)

view/SSViewer.php
kinglozzer: You can’t access the includer/parent’s scope from inside an include anyway (unless you pass it through manually), so `$Up` would never point to anything even if the include was inside a with/loop

dhensby: The strange thing is I think this actually depends - I'm sure I've tried that and sometimes it works sometimes it doesn't... but yer... not worth getting into here :P

view/SSViewer.php
```diff
@@ -181,6 +189,27 @@ public function __call($name, $arguments) {
...
+	protected function getItemStack() {
+		return $this->itemStack;
+	}
+
+	/**
+	 * @param array
+	 */
+	protected function setItemStack(array $stack) {
+		$this->itemStack = $stack;
+	}
```
kinglozzer: The alternative to adding a setter was to use references to modify the property directly:

```php
protected function &getItemStack() {
    return $this->itemStack;
}

// ...

$itemStack =& $this->getItemStack();
$itemStack[$this->getUpIndex()][SSViewer_Scope::ITEM_OVERLAY] = $this->overlay;
```

kinglozzer:
@dhensby I’ve added getters/setters to undo the visibility change

view/SSViewer.php
```diff
@@ -519,6 +548,70 @@ public function getInjectedValue($property, $params, $cast = true) {
...
+	 * scope respectively.
+	 */
+	public function obj($name, $arguments = null, $forceReturnedObject = true, $cache = false, $cacheName = null) {
+		$overlayIndex = false;
+
+		switch($name) {
+			case 'Up':
+				$upIndex = $this->getUpIndex();
+				if ($upIndex === null) {
+					user_error('Up called when we\'re already at the top of the scope', E_USER_ERROR);
```
dhensby: Should a bad `$Up` call break the entire execution thread?

I think this is a bit mean

kinglozzer: I think there’s an echo in here... see https://github.com/silverstripe/silverstripe-framework/pull/4869#discussion_r50678916 :D

dhensby: classic

PR #4870, opened 2015-12-22
NEW: Add ViewableData::setFailover() to refresh detected methods when changing failover

### LINKED ISSUE #4821: Object::defineMethods() doesn’t play nicely with ViewableData::$failover
Symptom: a non-top-level page without a controller class defined won’t call extra methods on its underlying datarecord. This has been fun to debug, sample code below. I’ll try to write up a simplified test soon.

```php
<?php

class PageWithImage extends Page {
        'Image' => 'Image'
    );
}

// IMPORTANT - add this page as a child of a regular 'Page'
// No controller class here, so it just falls back to Page_Controller
```
```html
<!-- PageWithImage.ss -->
{$Image}
```

No image is output. The reason seems to be something to do with `ViewableData::$failover`, the controller and `Object::defineMethods()`.

---

To make it easier to digest, here it is step-by-step with the example above:

1. A request is made to `http://mysite.com/some-page/page-with-image`.
2. `Page_Controller` is initalised for the first time for `/some-page/`. This results in a call to `ViewableData::defineMethods()`, which scans the underlying `Page` object (`$failover`) for extra methods, and finds none.
3. SilverStripe then tries to build the controller for `/page-with-image/`. Because you don’t need to define a controller for each page type anymore, it will fall back to `Page_Controller` if no controller class exists for the page type `PageWithImage`.
4. `Page_Controller` is constructed for a second time, this time for `/page-with-image/` as the underlying datarecord. `Object::__construct()` sees that `Page_Controller` has been initialised before, and [avoids calling `defineMethods()` again](https://github.com/silverstripe/silverstripe-framework/blob/4c19c183fe7b76329d59fdc944a1bebd544aabbb/core/Object.php#L675).
5. When you try to output the image, `Object::hasMethod()` won’t be able to find the `Image` method. This is because the “defined” methods were from the first page (the first time that `Page_Controller` was constructed).

I’m not really what the solution would be. Arguably `ViewableData::defineMethods()` needs to be called every time the `$failover` changes.

kinglozzer:
Hopefully the tests explain this better than my bug ticket did...

This is the only workaround I could think up for this issue. We don’t have a setter like `ViewableData::setFailover()` to hook into, and even if we added one we have no guarantee that people would use it (especially because setting `$this->failover = $foo;` has been an undocumented convention for a while now).

### DIFF
--- a/tests/view/ViewableDataTest.php
+++ b/tests/view/ViewableDataTest.php
@@ -20,7 +20,8 @@ class ViewableDataTest extends SapphireTest {
 
 	public function testFailoverRequiresCasting() {
 		$caster = new ViewableDataTest_Castable();
-		$container = new ViewableDataTest_Container($caster);
+		$container = new ViewableDataTest_Container();
+		$container->setFailover($caster);
 
 		$this->assertTrue($container->obj('alwaysCasted') instanceof ViewableDataTest_RequiresCasting);
 		$this->assertTrue($caster->obj('alwaysCasted', null, false) instanceof ViewableDataTest_RequiresCasting);
@@ -175,6 +176,19 @@ class ViewableDataTest extends SapphireTest {
 		$this->assertEquals('BBB', $objNotCached->obj('Test', null, true, true));
 	}
 
+	public function testSetFailover() {
+		$failover = new ViewableData();
+		$container = new ViewableDataTest_Container();
+		$container->setFailover($failover);
+
+		$this->assertSame($failover, $container->getFailover(), 'getFailover() returned a different object');
+		$this->assertFalse($container->hasMethod('testMethod'), 'testMethod() is already defined when it shouldn’t be');
+
+		// Ensure that defined methods detected from the failover aren't cached when setting a new failover
+		$container->setFailover(new ViewableDataTest_Failover);
+		$this->assertTrue($container->hasMethod('testMethod'));
+	}
+
 }
 
 /**#@+
@@ -252,10 +266,6 @@ class ViewableData_Caster extends ViewableData {
 
 class ViewableDataTest_Container extends ViewableData {
 
-	public function __construct($failover) {
-		$this->failover = $failover;
-		parent::__construct();
-	}
 }
 
 class ViewableDataTest_CastingClass extends ViewableData {
@@ -285,4 +295,6 @@ class ViewableDataTest_NotCached extends ViewableData {
 	}
 }
 
-/**#@-*/
+class ViewableDataTest_Failover extends ViewableData {
+	public function testMethod() {}
+}
--- a/view/ViewableData.php
+++ b/view/ViewableData.php
@@ -127,6 +127,25 @@ class ViewableData extends Object implements IteratorAggregate {
 			$this->setField($property, $value);
 		}
 	}
+
+	/**
+	 * Set a failover object to attempt to get data from if it is not present on this object.
+	 * 
+	 * @param ViewableData $failover
+	 */
+	public function setFailover(ViewableData $failover) {
+		$this->failover = $failover;
+		$this->defineMethods();
+	}
+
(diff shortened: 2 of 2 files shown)


view/ViewableData.php
```diff
@@ -127,6 +134,41 @@ public function __set($property, $value) {
...
+		return parent::allMethodNames($custom);
+	}
+
+	/**
+	 * Check whether the underlying failover object has changed or been removed, and call
+	 * defineMethods() if it has
+	 */
+	protected function checkAndRedefineMethods() {
+		if ($this->failover !== $this->oldFailover) {
+			$this->oldFailover = $this->failover;
```
kinglozzer: I did consider using `spl_object_hash()` for comparing the failover objects (instead of keeping two copies of them in memory), but I don’t think it’s reliable enough - the same hash can be re-used once an object is destroyed.

tractorcow:
Here's a small critique; Will we ever need to re-generate the extra methods when assigning a different instance of a failover of the same class? It looks like the instance is extracted from the failover in `Object::__call`.

You might also need a mechanism and test to ensure you remove methods from the removed failover, since it looks like the new ones are being merged with the old.

view/ViewableData.php
```diff
@@ -127,6 +134,41 @@ public function __set($property, $value) {
...
+	public function __call($method, $arguments) {
+		$this->checkAndRedefineMethods();
+		return parent::__call($method, $arguments);
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function hasMethod($method) {
+		$this->checkAndRedefineMethods();
```
tractorcow: Maybe this should be on setFailover instead, rather than duplicating it over all methods that rely on the failover?

kinglozzer: There is no `setFailover()` currently (see original comment) :(

kinglozzer:
> Will we ever need to re-generate the extra methods when assigning a different instance of a failover of the same class

It’s unlikely, but given that extra methods are defined at an instance-level instead of statically it’s _possible_, which is why I used that approach.

Upon reflection, I’m wondering if it might be better to add a `setFailover()` and call `defineMethods()` in there. Any user-code which doesn’t use that setter may encounter the bug that this is trying to fix, but I don’t think I’ve seen this raised before so I suppose it’s an edge-case?

view/ViewableData.php
micmania1: +1. If you have it all over the place (especially in __call()) then its going to redefine all methods every time something on the failover object changes.

view/ViewableData.php
micmania1: If force defineMethods() every time setFailover() or __construct() is called, would this remove the need for oldFailover check?

tests/view/ViewableDataTest.php
```diff
@@ -253,9 +269,14 @@ public function setValue() {}
 class ViewableDataTest_Container extends ViewableData {
 
 	public function __construct($failover) {
-		$this->failover = $failover;
+		$this->setFailover($failover);
 		parent::__construct();
 	}
+
+	public function setFailover($failover) {
```
tractorcow: This is setFailover :)

kinglozzer: Ah yes, I only added that to the test class. I think I’ll redo this PR and add it to `ViewableData`, we should probably have a setter anyway.

view/ViewableData.php
kinglozzer: Indeed it would, I think that’ll be the best approach :)

tractorcow:
Ok, thanks for your response. :) I'll have a look once you have an update.

kinglozzer:
Okay, updated. This PR is looking a bit simpler now :). I’ll do a PR for cms to go with this one that actually fixes #4821

PR #4873, opened 2015-12-23
FIX extraClass() method to match parent method

xini:
rebased against 3, see #4480

### DIFF
--- a/forms/CompositeField.php
+++ b/forms/CompositeField.php
@@ -122,9 +122,17 @@ class CompositeField extends FormField {
 	public function getLegend() {
 		return $this->legend;
 	}
-
+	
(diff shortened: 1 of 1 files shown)


dhensby:
thanks

tractorcow:
There's a small regression in this; See https://github.com/silverstripe/silverstripe-framework/commit/5470f23e2f46e95c97ab00dc891ffbda6283f035#commitcomment-15274570

dhensby:
You fix a bug and you find instances that relied on that bug.... Sigh.

tractorcow:
Logged at https://github.com/silverstripe/silverstripe-framework/issues/4895

PR #4877, opened 2015-12-26
Use newer ConfigStaticManifest from the master in 3.3

SilbinaryWolf:
Adds support for using the newer ConfigStaticManifest from the 4.0/master branch. 

If the developer is doing something weird that involves accessing the 'getStatics()', they can simply delete the ConfigStaticManifest40.php file to bring back full compatibility. 

My reasoning for this addition is that it reduces the /dev/build process greatly and increases the programmers productivity and improves their quality of life (in my opinion anyway, less waiting for 7-12s for private static changes).

I'm thinking this should go into 3.3.


micmania1:
This shouldn't need the extra class. It is backwards compatible but I put it in 4.0 because it's such a core change that it has a high chance of breaking something.

We really need a way for developers to define their own start-up, and at the moment its forking, or using the injector if available.

At this point its too early to "inject" so I'd say use constants to define the config class (instead of this change).

Hopefully by 4.0 these core features can be defined by the developers (index.php ftw).

SilbinaryWolf:
How about this change? If they're using deprecated features it tells the developer how to re-enable the backwards compatible SS_ConfigStaticManifest in the case that does actually break something.

I've been using the 4.0 SS_ConfigStaticManifest with 3.2.1 on two sites and I didn't encounter any issues over the 3 week development cycle.

All they need to do is add define('SS_USE_OLD_CONFIGSTATICMANIFEST', true); in their _ss_environment.php file.

micmania1:
I'd still be a bit cautious about making this the default, but it would be good to have it as an option that could be overriden (others are using forks).

Up to core devs to decide how/if this goes in :)

##### My Comments

It would be nice if the constant was consistent with existing constants (eg. `SS_CONFIGSTATICMANIFEST`) This would match `SS_MANIFESTCACHE` here https://github.com/silverstripe/silverstripe-framework/blob/3/core/manifest/ConfigStaticManifest.php#L43.

Another thing i'd comment on is that the `_40` suffix might confuse people as you need context of the change. Using `_Reflection` is a little more descriptive of what the class is doing.

tractorcow:
By the way, given that 3.3 is already in beta, it's probably too late for new major features to go in. New 3.x features will probably not be available until 3.4.

SilbinaryWolf:
Adjusted the code based on your feedback micmania1. I agree *_Reflection is a better name. Also changed the define to be like SS_MANIFESTCACHE for consistency.

core/manifest/ConfigStaticManifest.php
```diff
@@ -1,4 +1,71 @@
...
+	 */
+	public function __construct($base, $includeTests = false, $forceRegen = false, $cache = true) {
+		// Stubbed as these parameters are not needed for the newer SS_ConficStaticManifest version.
+	}
+
+	/**
+	 * Completely regenerates the manifest file.
+	 */
+	public function regenerate($cache = true) {
+		Deprecation::notice('3.4', 'This is no longer available as SS_ConfigStaticManifest now uses Reflection. For backwards compatibility define SS_USE_OLD_CONFIGSTATICMANIFEST in your _ss_environment.php file.');
```
kinglozzer: You can shorten this to just `'This method is no longer available as ' . __CLASS__ . ' uses Reflection.'` as this is now opt-in. Should solve the line-length issue too :)

core/manifest/ConfigStaticManifest.php
```diff
@@ -1,4 +1,71 @@
...
+					}
+				}
+
+			}
+		}
+		return null;
+	}
+
+	public function getStatics() {
+		Deprecation::notice('3.4', 'This is no longer available as SS_ConfigStaticManifest now uses Reflection. For backwards compatibility define SS_USE_OLD_CONFIGSTATICMANIFEST in your _ss_environment.php file.');
```
kinglozzer: Same as above

core/manifest/ConfigStaticManifest.php
```diff
@@ -1,4 +1,71 @@
...
+
+			// The config system is case-sensitive so we need to check the exact value
+			$reflection = new ReflectionClass($class);
+			if(strcmp($reflection->name, $class) === 0) {
+
+				if($reflection->hasProperty($name)) {
+					$property = $reflection->getProperty($name);
+					if($property->isStatic()) {
+						if(!$property->isPrivate()) {
+							Deprecation::notice('3.4', "Config static $class::\$$name must be marked as private",
```
kinglozzer: Can you set this to `4.0`? It should really be the same as the old `SS_ConfigStaticManifest` class: https://github.com/silverstripe/silverstripe-framework/blob/e14f743bf031bda2022cd12e37d26f420ce7c403/core/manifest/ConfigStaticManifest.php#L82

kinglozzer:
A couple of minor points but I think this looks okay otherwise, good work! We probably don’t want to be doing this sort of thing too often (i.e. back-porting things from master), but given how simple the extra code is and the positive impact it will have for developers I’m not against this.

Is there any value in adding docs to let people know about this somewhere, even if it’s just for the lifetime of 3.x?

core/manifest/ConfigStaticManifest.php
tractorcow: Deprecation notices should point at the version that the code will be removed; In this case, `4.0`.

core/manifest/ConfigStaticManifest.php
tractorcow: +1

core/Core.php
```diff
@@ -113,7 +113,12 @@
 }
 
 // Now that the class manifest is up, load the static configuration
-$configManifest = new SS_ConfigStaticManifest(BASE_PATH, false, $flush);
+if (defined('SS_CONFIGSTATICMANIFEST')) {
```
tractorcow: This needs documentation, both in the ss_enviroment setting list, and in the 3.4.0 changelog please. :)

SilbinaryWolf:
Before i go ahead and make the deprecation, documentation changes and whatnot, I feel it makes more sense to make it opt-out as newcomers to Silverstripe won't need to /dev/build each time a 'private static' is changed and they'll just have a better quality of life. 

Experienced users upgrading between 3.x to 3.4 should be told the ConfigStaticManifest has changed and that if any issues arise with the upgrade, to fallback to the 3.2 and lower ConfigStaticManifest.

However, if you guys disagree, I'll make the finish touches.

micmania1:
I'd prefer to err on the side of caution and carry on using the old static manifest until this has some real-world usage. I don't really have a strong opinion either way though as i'm aware of people already using it in production.

ps. You don't need to dev/build on each static change (currently) unless its database related. Even with this change you need to dev/build on static changes that affect the db - you just don't need to flush.

tractorcow:
> Experienced users upgrading between 3.x to 3.4 should be told the ConfigStaticManifest has changed 

I thought that by default, it already uses the existing manifest. A site upgrading from 3.3 to 3.4 should still be using SS_ConfigStaticManifest by default, correct?

SilbinaryWolf:
So to give a bit of context of where I was at this time last year. Had mostly down agency work which is "use latest and greatest" versions and just ship ASAP. I'll adjust and rebase this now as it's so close.

core/Core.php
SilbinaryWolf: @tractorcow Where are the files located that I need to modify for adding the documentation?

SilbinaryWolf:
Adjusted based on feedback. (ie. deprecation notices)
Just need to know what changelog / ss_environment documentation files I need to modify and then it'll be ready.

SilbinaryWolf:
Adjusted the environment variables document.

dhensby:
Seems good to me

PR #4879, opened 2015-12-30
Documentation : Using SQLite for Testing

gordonbanderson:
Additional section with instructions on running tests using in memory SQLite,   This makes running tests around 20 times faster.

### DIFF
--- a/docs/en/02_Developer_Guides/06_Testing/00_Unit_Testing.md
+++ b/docs/en/02_Developer_Guides/06_Testing/00_Unit_Testing.md
@@ -116,6 +116,46 @@ limited than using `phpunit` directly, particularly around formatting test outpu
 	sake dev/tests/all SkipTests=MySkippedTest
 	# Skip some tests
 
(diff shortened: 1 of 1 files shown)


tractorcow:
It might be better to leave "dev only" code out of the core project.

In my local environment I have the following in my `_ss_environment.php` file, and doesn't touch local code.

```php
// Select DB connector
if(!defined('SS_DATABASE_CLASS')) {
	if (php_sapi_name() == "cli") {
		// Testing dev db config
		define('SS_DATABASE_CLASS', 'SQLite3Database');
		define('SS_SQLITE_DATABASE_PATH', ':memory:');
		define('SS_DATABASE_USERNAME', 'root');
		define('SS_DATABASE_PASSWORD', '');
	} else {
		// Standard dev db config
	}
}
```

gordonbanderson:
That does make perfect sense :)

tractorcow:
I think your code is probably ok for testing. ;) It adds value so lets merge it in!

PR #4880, opened 2015-12-31
NOT FOR MERGE: W.I.P. Add <% withif %> template block

jonom:
***Don't merge yet please*** - this is a proof of concept to start a discussion.

Before
```html
<% if $Thing %>
	<% with $Thing %>
		$SomeProperty
	<% end_with %>
<% end_if %>
```

After
```html
<% withif $Thing %>
	$SomeProperty
<% end_withif %>
```

Just a little time saver for a common pattern. If people like the idea I'll add tests and docs etc. to this PR, and maybe a `loopif` equivalent - though that's a bit less useful as you often want to add a wrapper element around a loop block.

Assuming people like this idea what do we think of the block name `withif`? I don't love it but couldn't think of anything else that's fairly expressive but also compact.

Note: white space changes are due to .editorconfig... thought about removing them but if the config says that's the way to do it do we leave them in? FYI you can view the changes and [ignore whitespace changes by adding '?w=1' to the URL](https://github.com/silverstripe/silverstripe-framework/pull/4880/files?w=1).

### DIFF
--- a/view/SSTemplateParser.php
+++ b/view/SSTemplateParser.php
@@ -25,46 +25,46 @@ else {
  * @subpackage view
  */
 class SSTemplateParseException extends Exception {
-	
+
 	function __construct($message, $parser) {
 		$prior = substr($parser->string, 0, $parser->pos);
-		
+
 		preg_match_all('/\r\n|\r|\n/', $prior, $matches);
 		$line = count($matches[0])+1;
-		
+
 		parent::__construct("Parse error in template on line $line. Error was: $message");
 	}
-	
+
 }
 
 /**
   * This is the parser for the SilverStripe template language. It gets called on a string and uses a php-peg parser
   * to match that string against the language structure, building up the PHP code to execute that structure as it
   * parses
-  * 
+  *
   * The $result array that is built up as part of the parsing (see thirdparty/php-peg/README.md for more on how
   * parsers build results) has one special member, 'php', which contains the php equivalent of that part of the
   * template tree.
-  * 
+  *
   * Some match rules generate alternate php, or other variations, so check the per-match documentation too.
-  * 
+  *
   * Terms used:
-  * 
+  *
   * Marked: A string or lookup in the template that has been explictly marked as such - lookups by prepending with
   * "$" (like $Foo.Bar), strings by wrapping with single or double quotes ('Foo' or "Foo")
-  * 
+  *
   * Bare: The opposite of marked. An argument that has to has it's type inferred by usage and 2.4 defaults.
-  * 
+  *
   * Example of using a bare argument for a loop block: <% loop Foo %>
-  * 
+  *
   * Block: One of two SS template structures. The special characters "<%" and "%>" are used to wrap the opening and
   * (required or forbidden depending on which block exactly) closing block marks.
-  * 
+  *
   * Open Block: An SS template block that doesn't wrap any content or have a closing end tag (in fact, a closing end
   * tag is forbidden)
-  * 
+  *
   * Closed Block: An SS template block that wraps content, and requires a counterpart <% end_blockname %> tag
-  * 
+  *
   * Angle Bracket: angle brackets "<" and ">" are used to eat whitespace between template elements
   * N: eats white space including newlines (using in legacy _t support)
   *
@@ -120,7 +120,7 @@ class SSTemplateParser extends Parser implements TemplateParser {
 
 	/**
 	 * Set the closed blocks that the template parser should use
-	 * 
+	 *
 	 * This method will delete any existing closed blocks, please use addClosedBlock if you don't
 	 * want to overwrite
 	 * @param array $closedBlocks
@@ -195,7 +195,7 @@ class SSTemplateParser extends Parser implements TemplateParser {
 			);
 		}
 	}
-	
+
 	/* Template: (Comment | Translate | If | Require | CacheBlock | UncachedBlock | OldI18NTag | Include | ClosedBlock |
 	OpenBlock | MalformedBlock | Injection | Text)+ */
 	protected $match_Template_typestack = array('Template');
@@ -450,7 +450,7 @@ class SSTemplateParser extends Parser implements TemplateParser {
 	function Template_STR(&$res, $sub) {
 		$res['php'] .= $sub['php'] . PHP_EOL ;
 	}
-	
+
 	/* Word: / [A-Za-z_] [A-Za-z0-9_]* / */
 	protected $match_Word_typestack = array('Word');
 	function match_Word ($stack = array()) {
@@ -538,14 +538,14 @@ class SSTemplateParser extends Parser implements TemplateParser {
 
 
 
-	/** 
+	/**
 	 * Values are bare words in templates, but strings in PHP. We rely on PHP's type conversion to back-convert
 	 * strings to numbers when needed.
 	 */
 	function CallArguments_Argument(&$res, $sub) {
 		if (!empty($res['php'])) $res['php'] .= ', ';
-		
-		$res['php'] .= ($sub['ArgumentMode'] == 'default') ? $sub['string_php'] : 
+
+		$res['php'] .= ($sub['ArgumentMode'] == 'default') ? $sub['string_php'] :
 			str_replace('$$FINAL', 'XML_val', $sub['php']);
 	}
 
@@ -727,22 +727,22 @@ class SSTemplateParser extends Parser implements TemplateParser {
 
 
 
-	
+
 	function Lookup__construct(&$res) {
 		$res['php'] = '$scope->locally()';
 		$res['LookupSteps'] = array();
 	}
-	
-	/** 
-	 * The basic generated PHP of LookupStep and LastLookupStep is the same, except that LookupStep calls 'obj' to 
+
+	/**
+	 * The basic generated PHP of LookupStep and LastLookupStep is the same, except that LookupStep calls 'obj' to
 	 * get the next ViewableData in the sequence, and LastLookupStep calls different methods (XML_val, hasValue, obj)
 	 * depending on the context the lookup is used in.
 	 */
 	function Lookup_AddLookupStep(&$res, $sub, $method) {
 		$res['LookupSteps'][] = $sub;
-		
+
 		$property = $sub['Call']['Method']['text'];
-		
+
 		if (isset($sub['Call']['CallArguments']) && $arguments = $sub['Call']['CallArguments']['php']) {
 			$res['php'] .= "->$method('$property', array($arguments), true)";
 		}
@@ -1083,7 +1083,7 @@ class SSTemplateParser extends Parser implements TemplateParser {
 		$matchrule = "QuotedString"; $result = $this->construct($matchrule, $matchrule, null);
 		$_142 = NULL;
 		do {
-			$stack[] = $result; $result = $this->construct( $matchrule, "q" ); 
+			$stack[] = $result; $result = $this->construct( $matchrule, "q" );
 			if (( $subres = $this->rx( '/[\'"]/' ) ) !== FALSE) {
 				$result["text"] .= $subres;
 				$subres = $result; $result = array_pop($stack);
@@ -1093,7 +1093,7 @@ class SSTemplateParser extends Parser implements TemplateParser {
 				$result = array_pop($stack);
 				$_142 = FALSE; break;
 			}
-			$stack[] = $result; $result = $this->construct( $matchrule, "String" ); 
+			$stack[] = $result; $result = $this->construct( $matchrule, "String" );
 			if (( $subres = $this->rx( '/ (\\\\\\\\ | \\\\. | [^'.$this->expression($result, $stack, 'q').'\\\\])* /' ) ) !== FALSE) {
 				$result["text"] .= $subres;
 				$subres = $result; $result = array_pop($stack);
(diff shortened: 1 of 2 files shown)


jonom:
The other option would be to change With and Loop to behave like this by default. This would be fine by me as I don't want to render content if I'm working on an empty result, but I don't know how other developers would feel. Plus it would be a breaking change of course.

sminnee:
Since this is a PoC for discussion, I've tagged it as a draft RFC. We haven't discussed that previously but it seems like a good starting point?

I like the idea of having some syntatic sugar for this, but am also not particularly happy with withif.

Other options:

 * `<% ifwith $Var %> .. <% end_ifwith %>` (rationale: it's short for if, then with, so keep the order consistent)
 * `<% if/with $Var %> .. %<% end_if %>` (rational: it's a par of tags, keep them separate. also leaves room for `< if/loop $Var >`

@jonom, do you want to cross-post to ss-dev to broaden input on this?

jonom:
@sminnee sure I'll post something on the list about it.

> <% ifwith $Var %>

I went back and forth between `ifwith` and `withif` following the same logic as you, but ultimately I thought `withif` reads a little better. `ifwith` reads a bit like `if ($with)` to me whereas `withif` sounds more like `with (if it exists) $something`. Difference between `loopif` and `ifloop` probably illustrates the idea better.

> <% if/with $Var %> .. %<% end_if %>

I don't quite follow this one, are you saying there would be one opening but two closing tags? I assume you were typing in a hurry and meant <% if/with $Var %> .. <% end_if/with %> - I quite like that idea, the only thing I wonder about is if including a slash will cause syntax highlighting issues for editors. Quick test in Coda didn't cause any problems though.

tractorcow:
You could even go the route of creating a new keyword altogether.

E.g. `<% use $Var %> ... <% end_use %>`

Another alternative is to simply make `with` blocks disabled for non-existant values. It feels like we are trying to bake in better logic into a more awkward syntax, when it should probably be the default behaviour in the first place.

hafriedlander:
Another option would be to move the if to be an extension of the current loop and with blocks, ruby style like `<% loop $Var if $Var %>Bong<% end_loop %>`. This would also allow `<% loop $Var if $OtherVar = = 2 %>Bong<% end_loop %>` and so on, but is slightly more verbose in the simplest case. Avoids the two elements though.

jonom:
> Another alternative is to simply make with blocks disabled for non-existant values

Yeah I agree - echoed the same idea in an earlier comment. This makes sense to me as a default behaviour, I just wasn't sure if it would suit all developers.

hafriedlander:
> Another alternative is to simply make with blocks disabled for non-existant values
I can't think of any useful value in keeping the current behaviour, so I'd support this in 4.0

jonom:
@hafriedlander that is a neat idea too. The main thing I want to achieve though is less typing for those really simple 'if value' checks that I seem to do a lot in templates. I think if you need a more complex if statement the current syntax is probably okay, and is a good flexible default for this kind of markup:
```
<$ if $List %>
	<ul>
	<% loop $List %>
		<li>$property</li>
	<% end_loop %>
	</ul>
<% end_if %>
```

jonom:
Me, @hafriedlander and @tractorcow all in favour of disabling loop and with blocks with no value I think, then there's no need for any new syntax. @sminnee what do you think?

hafriedlander:
If we do change to "blocks are skipped on empty value", it'd be nice to introduce `else` to `loop` and `with` blocks, so you can do:

```
<ul>
  <% loop $List %>
    <li>$Title</li>
  <% else %>
    <li>No entries</li>
  <% end_loop %>
</ul>
```

(and the same idea for with)

patricknelson:
Chiming in here too. I'm thinking maybe it'd be more intuitive to simply support additional statements after the `if` conditional. The trouble would be in terminating those extra statements. While in my email to the dev-list may have only assumed a terminating `<% end_with %>` to terminate both tags, what if you simply incorporated your terminating tags in the inverse order?

Let's call this approach "sub-tagging"

**Example 1**

```asp
<% if $Thing with $Thing %>
    $SomeProperty
<% end_if %>

<% if $Thing loop $Thing %>
    $SomeProperty
<% end_if %>
```

jonom:
>  it'd be nice to introduce else to loop and with blocks

:+1:

patricknelson:
Also @hafriedlander and @jonom I think the potential for incorporating this logic with `else` would be more expressive using the sub-tag syntax I proposed above :smile: 

**Another example:**

```asp
<% if $Foo loop %>
    $SomeProperty
<% else_if with $Bar %>
   $AnotherProperty
<% else %>
  $Baz
<% end_if %>
```

To me you could intuitively chain various `loop`'s, `with`'s and a final else statement if no values are truthy.

tractorcow:
I thought that loops were already skipped for empty values, but I guess there are a few places they aren't doing so robustly?

patricknelson:
Ugh @tractorcow now I feel duped :sob: 

But, this bolsters my sub-tag syntax argument :+1: Why? Because while a `<% loop %>` will skip contents if empty, this new syntax would more eloquently allow you to skip on to your extra `<% else_if %>`'s or `<% else %>`'s.

patricknelson:
Or, if you wanted to, maybe pass an extra param to your new `loop` sub-tag to encapsulate the entirety of your `else_if` block without having to write an additional terminating `<% end_loop %>` statement.

tractorcow:
I think this fix should be less about merging markup tags together, and more about making markup tags behave predictably. If I do a <loop> or <with> over an empty value, I would predict it should know not to render the body.

patricknelson:
Maybe, but I largely agree. I only suspect issues when the context isn't very narrow. That is, I'm sure there are situations where a developer is encapsulating a possibly large region with a `<% with $Object %>` with the expectation that contents are still rendered even if `$Object` is empty, simply because they wanted to have a shortcut to prevent them from having to type the `$Object.` over and over again. That may be an edge case, however, and also bad form.

Are there any other tags that could benefit from this sort of logic? Otherwise, it would make sense to me to simply alter how `<% with ... %>` works.

jonom:
> I think this fix should be less about merging markup tags together, and more about making markup tags behave predictably. If I do a or over an empty value, I would predict it should know not to render the body.

I agree. I only introduced a new block type because of the possibility that Patrick highlighted:

> I'm sure there are situations where a developer is encapsulating a possibly large region with a <% with $Object %> with the expectation that contents are still rendered even if $Object is empty

But if everyone agrees this is an edge-case and not a practice we want to support then let's change the default behaviour. That's why we have a master branch :D

I'll let this simmer for a few days in case there are objections and update the PR accordingly if we have consensus. Or I can close this and write an RFC if necessary but it seems like overkill?

jonom:
> I thought that loops were already skipped for empty values, but I guess there are a few places they aren't doing so robustly?

I didn't realise this - tested now and apparently an empty list does skip the loop but a falsey value doesn't.

```html
<% with $Nothing %>
	<p>Value 1: $MoreNothing</p>
<% end_with %>
<!-- Result: <p>Value 1: </p> -->

<% loop $Nothing %>
	<p>Value 2: $MoreNothing</p>
<% end_loop %>
<!-- Result: <p>Value 2: </p> -->

<% loop $List('Page').filter('URLSegment', 'nothing') %>
	<p>Value 3: $MoreNothing</p>
<% end_loop %>
<!-- Result: (empty)-->
```

tractorcow:
Then it's dumb and needs fixing and testing. :)

sminnee:
> If we do change to "blocks are skipped on empty value", it'd be nice to introduce else to loop and with blocks.

This approach gets my vote for 4.0, because it introduces the necessary behaviour without the need for additional syntax, but instead expands on the usefulness/flexibility of the current syntax.

> That is, I'm sure there are situations where a developer is encapsulating a possibly large region with a <% with $Object %> with the expectation that contents are still rendered even if $Object is empty, simply because they wanted to have a shortcut to prevent them from having to type the $Object. over and over again. That may be an edge case, however, and also bad form.

I agree, but I think that only means that we can't make the change in 3.x.

jonom:
Oh yeah I forgot I pointed this at 3. I'll close this and re-open against master assuming we decide to change the default behaviour. I'm not sure I have the chops to introduce else block option within `with` and `loop` but I'll see how I go. It could always follow in a later commit otherwise.

patricknelson:
So to be clear, this means only the introduction of `else` and **not** `else` into the `loop` and `with` template blocks, correct?

assertchris:
I feel the need to point out (and feel dreadful doing so) that this pull request should not have been made against the official repo, nor should it contain formatting changes. It has often been expressed that the core committer team does not like WIP pull requests, and while I disagree with that wholeheartedly, I feel the need for us to be consistent. No less so because this fine change is being suggested by a core committer to begin with.

jonom:
@assertchris no problem, I appreciate the feedback. Still learning the ropes. Didn't seem like a big enough change to warrant an RFC and I was under the impression that most PRs are works in progress as they go through several changes before merge... but I wasn't really comfortable  opening a PR that wasn't ready for merge so should have known it wasn't a good move. Just FYI the reason I marked it as a WIP wasn't that I felt the functional code was incomplete, I just wanted to bundle tests and docs with it if merged - but didn't want to go to that effort if that solution wasn't sound. As it turns out it would have been a wasted effort :)

I'll close this for now as the consensus is leaning towards a different solution anyway, but let's let the discussion continue here and on the mailing list. In hindsight I should have just posted to the core dev list instead of opening a PR I suppose.

For next steps I think we pretty much have a consensus that we should keep the current syntax, but
1. skip rendering `with` and `loop` blocks for falsey and empty values.
2. introduce the ability to add an `else` section within `with` and `loop` blocks, for fallback content.

Should I open a PR against master for that solution or do we think this needs to go through an RFC process?

assertchris:
@hafriedlander has been talking about a fast feedback process, which would render the RFC process optional in the case of small changes. If that's up anywhere for us to read, that would be awesome.

I don't think this is a big enough change to warrant an RFC (not that I make those kinds of decisions), and as I said; I wholeheartedly disagree with not allowing WIP pull requests.

jonom:
p.s. @assertchris just on the white space - I did mention this in my opening comment: It seems the framework project has a .editorconfig bundled with it now which is the source of the whitespace changes. It seems counter-productive to have a .editorconfig file included at this point if we don't want files formatted that way? Ideally when that file was added all code would have been converted to conform to those rules.

assertchris:
@jonom I agree, those rules should govern the addition or modification of code. I don't personally mid formatting changes through modified files, but it does lead to increased review periods. That same checklist says "It's better to submit multiple patches with separate bits of functionality than a big patch containing lots of changes", but whether it applies to formatting + functionality changes or just lots of functionality changes is open for debate.

I'm sorry for not seeing your formatting comment sooner. I assumed the opening comment was just a repeat of the mailing list post. I make silly assumptions far too often...

jonom:
lol all good. You know who might be able to help with whitespace though is our clever @helpfulrobot pal ;) maybe he could open PRs to convert process all code to conform to the .editorconfig files? We're going to come up against this issue fairly regularly otherwise I guess.

assertchris:
I think that has been happening, but Framework and CMS are edge cases. I think the idea is for PSR-2 closer to an alpha release? cc @sminnee @tractorcow

sminnee:
In the meantime...
https://github.com/silverstripe/silverstripe-framework/pull/4888

tractorcow:
Yes once we are closer to releasing 4.0 we'll do a code reformat of all core modules. At the moment, doing so would simply make merging up more difficult.

:)

Also @assertchris is right; I should probably remind that we aren't keen on "WIP" pull requests. It's fine to create an issue pointing to user code / commits if you aren't ready for merge.

jonom:
> once we are closer to releasing 4.0 we'll do a code reformat of all core modules

@tractorcow does that apply for whitespace too? Does this mean we should be disabling editor config support in the mean time so we're not introducing whitespace changes?

> It's fine to create an issue pointing to user code / commits if you aren't ready for merge.

Noted, thanks.

tractorcow:
Yep it'll apply to whitespace.

Editorconfig says indentation = tab (which is the current status quo), so I'm not sure why it's a problem.

jonom:
It's trailing whitespace that's causing a problem in this case, not tabs.

jonom:
Particularly because of the effect `trim_trailing_whitespace = true` has on blank/empty lines. These lines commonly contain some white space to set indentation to the same level as the preceding line but this setting causes it to be removed.

jonom:
I've had a try at modifying `with` and `loop` to render nothing if empty but I'm having some trouble as I don't know enough about SSTemplateParser or SSViewer. Does someone with more experience in those areas want to take over? What I've done so far is here: https://github.com/jonom/silverstripe-framework/commit/8939f2f5649ed1d4abee4d26767e8bc4dbaac9ff

Got tripped up by a couple of things:

1. I wasn't aware that you can do `<% loop %>` with no argument to loop over the current scope. Not sure how to test for a value there without unintentionally changing the scope.
2. Found an edge case which is that if you do `<% with $EmptyList %>` it should probably render the block because even though the list is empty it could have properties we still want to use. Currently `<% if $EmptyList %>` will fail but I think it should pass for this reason, and `<% if $EmptyList.Count %>` should be used to test that lists aren't empty instead. This is more consistent with how SS_List objects (or any object) work in PHP code.

Does anyone want to take this over? To recap, the goal is to get here:

```
<% with $Thing %>
    ... // Change scope to $Thing if <% if $Thing %> test would pass
<% else %> // Optional
    ... // Otherwise render a fallback
<% end_with %>
```
and

```
<% loop $List %>
    ... // if $List *contains items* loop through them changing scope to each
<% else %> // Optional
    ... // Otherwise render a fallback
<% end_loop %>
```

Issue #4881, opened 2016-01-02
SiteTree Drag/Drop sorting bugfix


sb-relaxt-at:
This Issue is based on https://github.com/silverstripe/silverstripe-cms/issues/744 as jquery.jstree.js is part of the framework: 
> Dragging and dropping sometimes doesn't respond - Not sure why, but about half the time nothing happens (no ajax response / request). This is a seriously annoying bug.

The problem occurs if the dragged page is dropped exactly on the drop indicator (the line).

scott1702:
Can be closed now :)

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-framework/pull/4882

Issue #3749, opened 2015-01-01
BUG Edge case in SS_DatetimeTest::testFormatFromSettings()


tractorcow:
There is an edge case in the timezone calculation used by this method which seems to surface during new years day. About 13 hours every new year (while NZ is one year ahead of GMT time) this test case will fail.

```
$ ./framework/sake dev/tests/SS_DatetimeTest flush=all
SS_DATETIMETEST

---------------


............F
SS_DatetimeTest::testFormatFromSettings
Failed asserting that two strings are equal.
PHPUnit_Framework_Assert::assertEquals(31/12/2014 10:11:01,31/12/2015 10:11:01)
DatetimeTest.php:170
```

It seems that at 1pm on new years day (since nz is +13GMT during daylight saving) the issue magically resolves itself. Perhaps some issue with zend_date?

sb-relaxt-at:
I guess this issue is the same @jonom discovered with my pull request (https://github.com/silverstripe/silverstripe-framework/pull/4882#issuecomment-168538805). I have investigated this further and I am afraid there is a pattern mistake in using Zend_Date. In Zend_Date 'Y' (http://framework.zend.com/manual/1.12/en/zend.date.constants.html#zend.date.constants.selfdefinedformats) refers to the year as specified in ISO 8601, which in turn is not always equal to the calendar year (namely around the beginning of a year: https://en.wikipedia.org/wiki/ISO_week_date#Disadvantages). See also this related discussion: http://stackoverflow.com/questions/1978051/zend-datetostring-outputs-the-wrong-year-bug-in-my-code-or-zend-date

From my understanding we have to change all 'Y' (uppercase) occurencies when using Zend_Date to 'y' (lowercase). Fortunately the options in Member seems to use the lowercase 'y'. Nevertheless the format description on the profile page advertise the use of uppercase 'Y'. I am not sure whether or not there are other parts to change. I will start with updating the testcase accordingly.

sb-relaxt-at:
The update for the testcase: https://github.com/sb-relaxt-at/silverstripe-framework/commit/642751f7e410069d9c2d56a72f3bad6ae5913e70

jonom:
Thanks for your work on this @sb-relaxt-at!

tractorcow:
We can probably fix this for 2.2.2 :)

chillu:
[...]

PR #4882, opened 2016-01-02
jquery.jstree patched to improve drag-and-drop handling

sb-relaxt-at:
This proposes a fix of #4881. The two small patches are backported from https://github.com/vakata/jstree/issues/174#issuecomment-8060768 and https://github.com/vakata/jstree/issues/174#issuecomment-8135128.

There are some other useability issues remaining but it is the fastest way to fix this especially annoying issue. Updating to a new jstree version seems to be more work...

P.S.: I am not sure about how to handle changes to a thirdparty library. I have only added a notice at the beginning of the js file.

### DIFF
--- a/thirdparty/jstree/jquery.jstree.js
+++ b/thirdparty/jstree/jquery.jstree.js
@@ -1,5 +1,5 @@
 /*
- * jsTree 1.0-rc3
+ * jsTree (custom version for SilverStripe: 1.0-rc3 + minor bugfixes)
  * http://jstree.com/
  *
  * Copyright (c) 2010 Ivan Bozhanov (vakata.com)
@@ -2442,7 +2442,7 @@
 							0);
 						}
 					}, this))
-				.delegate("a", "mouseup.jstree", $.proxy(function (e) { 
+				.delegate("a, #jstree-marker-line", "mouseup.jstree", $.proxy(function (e) {
 						if($.vakata.dnd.is_drag && $.vakata.dnd.user_data.jstree) {
 							this.dnd_finish(e);
 						}
@@ -2524,7 +2524,7 @@
 								$.vakata.dnd.helper.children("ins").attr("class","jstree-invalid");
 							}
 						}, this))
-					.delegate(s.drop_target, "mouseup.jstree-" + this.get_index(), $.proxy(function (e) {
+					.delegate(s.drop_target + ", #jstree-marker-line", "mouseup.jstree-" + this.get_index(), $.proxy(function (e) {
 							if(this.data.dnd.active && $.vakata.dnd.helper.children("ins").hasClass("jstree-ok")) {
 								this._get_settings().dnd.drop_finish.call(this, { "o" : o, "r" : $(e.target), "e" : e });
 							}


jonom:
This is a super annoying bug so it would be great to have it fixed. At some point I hope we'll upgrade to the latest version of jsTree or something different, but that's pretty sure no to happen in the 3.x line so I'm definitely in favour if patching this file in the mean time. I don't see any harm in adding a patch because the 1.x line of jsTree will never be updated again so we don't have to worry about re-patching future releases.

Behat tests passed, just an unrelated date test failing - some kind of New Year issue (expected 2015 got 2016). @tractorcow maybe you know what happened there?

thirdparty/jstree/jquery.jstree.js
```diff
@@ -1,5 +1,5 @@
 /*
- * jsTree 1.0-rc3
+ * jsTree 1.0-rc3 (patched)
```
jonom: Maybe make this a bit more specific like `jsTree custom for SilverStripe (1.0-rc3 + minor bugfixes)`

sb-relaxt-at:
Thanks for the feedback, I have updated my pull request. Regarding the not passing tests, see #3749.

Updating jstree seems to be difficult, I have tried to update to the final 1.0 version (not a release candidate), but there have been some api changes between these versions.

jonom:
Thanks @sb-relaxt-at, closed and re-opened to force the tests to re-run

tractorcow:
Tests seem ok now. :)

I'll ask the UX team here to have a go at testing your fix.

sb-relaxt-at:
Great, looking forward for this :-). We have done some limited testing in Firefox/Chrome/IE11 and deployed it to some of our SilverStripe instances. Seems to be fine so far. Although this bugfix doesn't resolve all drag and drop problems, it should be a noticeable improvement.

dhensby:
Can't this be fixed without editing the js plugin or fixed in the upstream?

sb-relaxt-at:
The related upstream issue has been closed by @vakata (owner) in 2013 (most likely because it has been fixed in 1.0, 2.0 or 3.0 ;-)). We can be quite sure, that there will be no compatible upstream updates for the used version 1.0-rc3. Even updating to 1.0 seems to be difficult due to some changes. Therefore I believe direct patching (which I usually avoid as hell) is acceptable. 

It might be possible to fix it without touching the original library, by adding some event handlers and delegations. Nevertheless this seems cumbersome, would require a lot more effort.

vakata:
I thought I'd chime in with some info on jstree. As far as I can remember all tickets regarding v.1 were closed when v.3 proved to be at least as stable as v.1. 
The reason for closing all the tickets was that unfortunately I did not have enough resources to support both versions, so I decided to drop support for v.1. 
As for the current situation - v.1 has not been supported for years. v.3 is the only supported version, with v.4 planned for the end of 2016. If you are still using v.1 code I strongly suggest to upgrade to v.3.

jonom:
@dhensby as @vakata points out we really shouldn't be using the v1 line, but there is apparently a lot of custom code that relies on it in the CMS so updating to a new version looks pretty hard. I don't think it's worth the effort for the SS 3.x line especially if it doesn't carry over to 4.x, which may be the case if we're redeveloping a lot of components. Patching the file seems like the best short-term option.

jonom:
By the way thanks for the input @vakata and for your plugin!

tractorcow:
Time to update to v3 I guess!

scott1702:
I'm happy with this fix - nice work!

sb-relaxt-at:
@scott1702 Thanks, I am looking forward to see it in the release :-)

PR #4884, opened 2016-01-04
Small optimizations for 'Hierachy::Children' and 'Config' class.

SilbinaryWolf:
Replaced filterByCallback for 'Children' to just create a new array as function calls are exponentially expensive in PHP (the more functions that exist, the slower a function call becomes) and replaced 'array_key_exists' with op-code equivalent for speed. The best increase isn't really noticeable but we should work towards optimizing the core as much as possible.

### DIFF
--- a/core/Config.php
+++ b/core/Config.php
@@ -514,7 +514,7 @@ class Config {
 		// Include extensions only if not flagged not to, and some have been set
 		if (($sourceOptions & self::EXCLUDE_EXTRA_SOURCES) != self::EXCLUDE_EXTRA_SOURCES) {
 			// If we don't have a fresh list of extra sources, get it from the class itself
-			if (!isset($class[$this->extraConfigSources]) || $class[$this->extraConfigSources] === null) {
+			if (!array_key_exists($class, $this->extraConfigSources)) {
 				$this->extraConfigSources[$class] = Object::get_extra_config_sources($class);
 			}
 
@@ -617,7 +617,7 @@ class Config {
 		} else {
 			if (!isset($this->overrides[0][$class])) $this->overrides[0][$class] = array();
 
-			if (!isset($name[$this->overrides[0][$class]]) || $name[$this->overrides[0][$class]] === null) {
+			if (!array_key_exists($name, $this->overrides[0][$class])) {
 				$this->overrides[0][$class][$name] = $val;
 			} else {
 				self::merge_high_into_low($this->overrides[0][$class][$name], $val);
--- a/core/Config.php
+++ b/core/Config.php
@@ -514,7 +514,7 @@ class Config {
 		// Include extensions only if not flagged not to, and some have been set
 		if (($sourceOptions & self::EXCLUDE_EXTRA_SOURCES) != self::EXCLUDE_EXTRA_SOURCES) {
 			// If we don't have a fresh list of extra sources, get it from the class itself
-			if (!array_key_exists($class, $this->extraConfigSources)) {
+			if (!isset($class[$this->extraConfigSources]) || $class[$this->extraConfigSources] === null) {
 				$this->extraConfigSources[$class] = Object::get_extra_config_sources($class);
 			}
 
@@ -617,7 +617,7 @@ class Config {
 		} else {
 			if (!isset($this->overrides[0][$class])) $this->overrides[0][$class] = array();
 
-			if (!array_key_exists($name, $this->overrides[0][$class])) {
+			if (!isset($name[$this->overrides[0][$class]]) || $name[$this->overrides[0][$class]] === null) {
 				$this->overrides[0][$class][$name] = $val;
 			} else {
 				self::merge_high_into_low($this->overrides[0][$class][$name], $val);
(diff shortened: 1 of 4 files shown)


core/Config.php
```diff
@@ -514,7 +514,7 @@ protected function getUncached($class, $name, $sourceOptions, &$result, $suppres
 		// Include extensions only if not flagged not to, and some have been set
 		if (($sourceOptions & self::EXCLUDE_EXTRA_SOURCES) != self::EXCLUDE_EXTRA_SOURCES) {
 			// If we don't have a fresh list of extra sources, get it from the class itself
-			if (!array_key_exists($class, $this->extraConfigSources)) {
+			if (!isset($class[$this->extraConfigSources]) || $class[$this->extraConfigSources] === null) {
```
tractorcow: The null check is redundant, since `isset` returns false for null values in an array.

If you want to treat "null" values as set, you need to use array_key_exists, as the code used to be, otherwise you are changing the behaviour of the code anyway, and it's not an optimisation anymore.

Before: Null values would not invoke the next line (array_key_exists is true), but after your change they would (isset is false and `!` would flag it true).

This would potentially be LESS optimal, since null values would no longer be cached, and Object:: get_extra_config_sources() would have to be run more often just to return null again.

tractorcow:
Thanks for making me think!

core/Config.php
SilbinaryWolf: I actually realized my mistake late yesterday with this optimization when I was playing with 'hasField' optimizations. Planning to use instead an isset() followed by array_key_exists(). 

ie. !isset($class[$this->extraConfigSources]) || !array_key_exists($class, $this->extraConfigSources)

As that should give a speed boost in all not variable !== null cases. I'm probably going to experiment with writing an optimizer in PHP that just converts 'is_array' into the cast optimized version as well as this, as that should improve the floor speed of Silverstripe by a few noticeable ms.

SilbinaryWolf: Also planning to add the reverse, a Deoptimizer if you will, as is_array()/is_string() etc are op-code based in PHP7 if what I've read holds true.

tractorcow: Ok, I'll wait to see what result comes out of your next update then. :)

SilbinaryWolf:
Removed the Config class changes.

Also pushed my other change (that previously forgot to push).
"Simplified ClassLoader to just run the logic that existed in ClassInfo itself. Since the function 'ClassInfo::exists' is called ~200 times in vanilla Silverstripe 3.1, I figured it was worth the optimization."

I figure deprecate the ClassLoader function in 4.0 now instead of 3.3 but just waiting for feedback on the message.

core/Config.php
SilbinaryWolf: The Optimizer plan is going to come as a seperate PR. Just wondering, since this PHP file is meant to be run from CLI on the entire framework codebase, should I have two commits with just the Optimizer::optimize() / Optimizer::deoptimize() and then the next one that touches whatever files use is_array(), array_key_exists(), etc? or just one for it all?

SilbinaryWolf: Tried writing an optimizer. Due to the is_array call count being trivial (ie. 1000 calls on vanilla Silverstripe), it didn't really do much and it's probably not worth. Not even going to bother with the rest as they're all also under 1000 calls.

core/manifest/ClassLoader.php
```diff
@@ -101,6 +101,7 @@ public function getItemPath($class) {
 	 * @return bool
 	 */
 	public function classExists($class) {
+		Deprecation::notice('3.3', 'Not necessary anymore as this function is inlined in \'ClassInfo::exists\' for speed and that\'s the only place where it is used in core.');
```
tractorcow: Please set deprecation level to 4.0 (the version this will be removed in).

The description should just be "Use ClassInfo::exists"; You don't need to include the justification in the notice. Maybe that's better as a comment?

I've also noticed here that this mechanism does NOT work well with psr-4 autoloading. Is that going to be a problem? If a class isn't already loaded, and requires psr-4 to load, then it will incorrectly report false.

tractorcow:
@sminnee how do you feel about us supporting classes loaded in other non-module libraries? Does ClassInfo::exists need to support those? Maybe we should add a disclaimer to ClassInfo that it only works with silverstripe module code, and code in non-module libraries should use class_exists.

core/manifest/ClassLoader.php
SilbinaryWolf: I don't really see the appeal of PSR-4 so I don't completely understand it's purpose. How do my changes not work well with PSR-4?

SilbinaryWolf:
Made the changes as instructed.

tractorcow:
Looks good made. I'll check this out again soon before I merge it.

You might want to squash those commits, since "Updated deprecation message as instructed" won't be very useful to people reviewing the git history. :)

dhensby:
This seems legit to me... Just need to squash it down and have Damian happpy

tractorcow:
I'm never / always happy.

Issue #4606, opened 2015-09-15
RFC-7: JavaScript Module Management


flashbackzoo:
# RFC-7: JavaScript Module Management

## Purpose and outcome

Expanding on [RFC-4: Client-side dependency management](https://github.com/silverstripe/silverstripe-framework/issues/4372), RFC-7 proposes an approach for managing JavaScript modules (vendor and application code), within the Framework.

Implementing this RFC would give SilverStripe Framework a build process for JavaScript modules, access to ES6 features, and allow for additional tooling such as unit testing to be implemented in the furure. If accepted - this approach should also be adopted by SilverStripe CMS and Supported Modules.

## Motivation

Client and server code are tightly coupled in Framework. Greater decoupling will allow us to build more sophisticated client-side applications.

The latest version of JavaScript (ES6) is now a standard. We should start to take advantage of some of the new features.

## JavaScript Modules

The purpose of a JavaScript module is to encapsulate a useful piece of functionality, have the ability to register that functionality, and have other modules reference and use it.

JavaScript modules come in a few flavours, here's a brief overview of each, to provide some context for the rest of the proposal.

### UMD

Universal Module Definition is used for code that runs on the client and server, and where the target browser environment is unknown. UMD modules typically look like this:

```javascript
(function (root, factory) {
    if (typeof define === 'function' && define.amd) {
        // AMD
        define(['dep1', 'dep2'], factory);
    } else if (typeof exports === 'object') {
        // CommonJS
        module.exports = factory(require('dep1'), require('dep2'));
    } else {
        // Default browser with no bundling (root is window)
        root.MyModule = factory(root.dep1, root.dep2);
    }
}(this, function (dep1, dep2) {
    // Module code here
}));
```

The advantage of UMD is your module can be consumed in multiple environments without knowing anything about those environments in advance.

### AMD

Asynchronous Module Definition allows a module's dependencies, spread across multiple locations, to be loaded asynchronously. AMD modules look something like this:

```javascript
define('MyModule', ['dep1', 'dep1'], function (dep1, dep1) {
    // Define the module by returning something
    return {};
});
```

AMD modules need a module loader like Require.js to be present in the host environment.

### CommonJS

The CommonJS pattern was popularised by Node.js, has been adapted to work in browsers, and looks like this:

```javascript
var dep1 = require('dep1');
var dep2 = require('dep2');

var myModule = {};

module.exports = myModule;
```

Use in browsers requires a package bundler like Browserify.

[...]

PR #4888, opened 2016-01-05
MINOR: Remove trailing whitespace

sminnee:
Ran this command on all files:

    find . -not \( -name .svn -prune -o -name .git -prune \) -type f | grep -vE .\(gif\|png\|jpg\) | xargs perl -pi -e 's/ +$//'

The main benefit of this is so that authors who make use of .editorconfig don't end up with whitespace changes in their PRs.

Done on `3` with the expectation that it can be merged forward to `master`. If everyone like it, I can do the same on CMS.


dhensby:
Did this effect binary files? I see pdfs, swf and zip showing up as changed...

spekulatius:
I have run the following command to achieve this:

    find . -type f -name '*.ss' -exec sed --in-place 's/[[:space:]]\+$//' {} \+;
    find . -type f -name '*.php' -exec sed --in-place 's/[[:space:]]\+$//' {} \+;

Its a bit more simple and limits it to .php and .ss files straight away.

jonom:
Besides binary files we might want to skip over anything in a `thirdparty` folder too

sminnee:
Good catch on the binary files. I hadn't excluded them explicitly.

I've had a look at the extensions referenced in the current commit.

These are the main ones, with >4 files each in the commit I just did:

ss
css
scss
html
yml
php
js

But there are also these, which I think it would be nicer to include for completeness.

Cakefile
csv
htm
inc
php4
php5
README
todo
LICENSE
xml
textile
txt
coffee

These should be excluded:

wav
zip
psd
swf
pdf

This one is excluded by our editorconfig and so can be excluded;

md

This one should also be excluded by the editorconfig:

markdown

I don't think that we should exclude the thirdparty folder, unless we also want to update our `.editorconfig` with `trim_trailing_whitespace = false`. Otherwise if we ever get a patch to a thirdparty file (it happens occasionally), then we'll have the same issue.

What is better? Excluding the .editorconfig rule from thirdparty or applying this process to thirdparty?

jonom:
> What is better? Excluding the .editorconfig rule from thirdparty or applying this process to thirdparty?

Personally I'd say it's better to exclude the .editorconfig rule from thirdparty so we can try to keep those files as close to their distributed state as possible. Or maybe we don't exclude that folder from the .editorconfig rules but we still skip it for this conversion process, as people editing third party files is an edge case and shouldn't happen too often so it's probably not a big concern if they trigger white space changes during a patch?

It's also likely (going forward anyway) that modules in a thirdparty directory will have their own .editorconfig files, so rather than apply our own rules to them it would be better to respect theirs.

assertchris:
> It's also likely (going forward anyway) that modules in a thirdparty directory will have their own .editorconfig files, so rather than apply our own rules to them it would be better to respect theirs.

Furthermore, third-party PHP libraries should be controlled by Composer and third-party JavaScript libraries may be controlled by https://github.com/silverstripe/silverstripe-framework/issues/4606. I imagine the third-party folder fading into obscurity.

sminnee:
OK, I've re-done this with the following command, and patched .editorconfig

    for match in '*.ss' '*.css' '*.scss' '*.html' '*.yml' '*.php' '*.js' '*.csv' '*.inc' '*.php5'; do
    	find . -path ./thirdparty -prune -o -type f -name "$match" -exec sed -i '' 's/[[:space:]]\+$//' {} \+
    	find . -path ./thirdparty -prune -o -type f -name "$match" | xargs perl -pi -e 's/ +$//'
    done

Note that I also tried out https://www.npmjs.com/package/editorconfig-tools but it also fiddled with tabs vs spaces which I thought was counter-productive right now.

tractorcow:
Parse error in map.php. :)

Somehow you trimmed trailing `+` as well?

tractorcow:
Also a comment on https://github.com/silverstripe/silverstripe-cms/pull/1354/

sminnee:
OK, one of the issues here is that in adding newlines to the templates used to render form fields, I've actually changed the HTML output of the form fields. 

My thinking is that I should fix this by making this change. This will remove the trailing newline from custom templates that developers have built, but this seems like a less invasive change than changing the HTML output of default form fields.

    --- a/forms/FormField.php
    +++ b/forms/FormField.php
    @@ -839,7 +839,7 @@ class FormField extends RequestHandler {
     
                    $this->extend('onBeforeRender', $this);
     
    -               return $context->renderWith($this->getTemplates());
    +               return trim($context->renderWith($this->getTemplates()));
            }
 

Make sense?

jonom:
Forcing a newline and trimming trailing white space could potentially be problematic for any .ss file I suppose since they can be partials/includes and whitespace has meaning in html at least in an inline context. For instance:
```html
<p>Some
text</p>
```
renders as 'Some text' because the newline is resolved to a space character.

sminnee:
I think building a system where the presence or absence of a newline at the end is a bad idea. Also, we've been working with an editorconfig that adds trailing newlines for a long time, we just haven't applied that rule consistently. Having a complex set of rules about where newlines are or aren't added will get confusing.

Arguably, we could trim whitespace from the template rendered into ` <% include %>` calls too, since typically either the include tag will have a newline after it, and if not, the developer probably prefers that the text is displayed inline.

tractorcow:
I think trimming on render is fine.

sminnee:
@tractorcow do you think we should do it for includes as well as form fields?

tractorcow:
It might be less ugly to push up the behaviour to SSViewer and then make it a configurable option. It's better than a trim() put everywhere templates are rendered.

tractorcow:
```php
if($this->config()->trim_whitespace) {
	$output = trim($output);
}
return DBField::create_field('HTMLText', $output, null, array('shortcodes' => false));
```

sminnee:
Hrm. Upon reflection I don't think that we should change the rendering of include whitespace in 3, as there's nothing we've actually broken that we need to fix. The form fields was the only one of those.

I'd prefer not to have a config option, and instead just make a more considered change in 4.

tractorcow:
I think all you need to do is make sure your tests don't care about trailing whitespace or not. :)

jonom:
Yeah I think Damien's probably right - the modified html output your talking about Sam probably doesn't actually render differently in the browser does it? It just looks different in source view?

sminnee:
The tests were failing because they picked up on the fact that the HTML of all default formfields had additional newlines. This is a change, and it is a change that I didn't want.

To prevent the change from happening, I've made a change to FormField.php. I've broken my non-whitespace changes as a separate commit out now.

sminnee:
Some of the tests I have made ignore whitespace: these are tests where the templates modified were just the test templates, so I haven't changed a public API and it's fine to ignore the whitespace.

sminnee:
See #4896 for my SS4 suggestions.

tractorcow:
Comments on https://github.com/sminnee/silverstripe-framework/commit/f3c730b700e1324de45711a97ac3c24774327065

sminnee:
Changes made.

tractorcow:
I'm now merging 3 up into master and I must say, I'm now regretting lots of life choices. :)

PR #4889, opened 2016-01-05
DOCS Explicitly noting lack of PHP 7 support

dhensby:
Coz you know... apparently 5.3.2 is ambiguous about not supporting PHP 7

### DIFF
--- a/docs/en/00_Getting_Started/00_Server_Requirements.md
+++ b/docs/en/00_Getting_Started/00_Server_Requirements.md
@@ -8,7 +8,7 @@ Our web-based [PHP installer](installation/) can check if you meet the requireme
 
 ## Web server software requirements
 
- * PHP 5.3.2+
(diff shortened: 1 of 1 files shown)


sminnee:
I think it would be more helpful to include a section such as this, to answer the inevitable question.

    ## Why doesn't SilverStripe 3 work with PHP 7?
    Unfortunately, SilverStripe has classes named the same as PHP reserved words, such as "Int" and "Float". This means that we are unable to make SilverStripe 3 support PHP 7 without breaking backward compatibility. SilverStripe 4 will work with PHP 7 and will be released in 2016. Until then, we recommend that you use PHP 5.6.

dhensby:
Sure. I'll add it

chillu:
@dhensby ping

dhensby:
@chillu done.

PR #4894, opened 2016-01-05
fixing examples

spekulatius:
PSR-2 style for examples

### DIFF
--- a/docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
+++ b/docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
@@ -30,37 +30,37 @@ as close to rendering as possible (e.g. in `[api:FormField]`.
 
 	class MyCustomController extends Controller {
 
-		public function init() {
-			parent::init();
+        public function init() {
+            parent::init();
 		
-			Requirements::javascript("cms/javascript/LeftAndMain.js");
-			Requirements::css("cms/css/TreeSelector.css");
-		}
-	}
+            Requirements::javascript("cms/javascript/LeftAndMain.js");
+            Requirements::css("cms/css/TreeSelector.css");
+        }
+    }
 
 ### CSS Files
 
 	:::php
-	Requirements::css($path, $media);
+    Requirements::css($path, $media);
 
 If you're using the CSS method a second argument can be used. This argument defines the 'media' attribute of the 
 `<link>` element, so you can define 'screen' or 'print' for example.
 
 	:::php
-	Requirements::css("cms/css/TreeSelector.css", "screen,projection");
+    Requirements::css("cms/css/TreeSelector.css", "screen,projection");
 
 ### Javascript Files
 
 	:::php
-	Requirements::javascript($path);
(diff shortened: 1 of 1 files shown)


spekulatius:
Hello @dhensby 

do you want to merge this in quickly?

Cheers,
Spek

tractorcow:
That's not actually psr-2 though :) Or did you just mean tab to space conversion?

tractorcow:
I'm not sure why this is going into 3.1 branch; Is there a context that I'm missing?

dhensby:
Hmm. How does out docs viewer like the space changes?

tractorcow:
I think it works with 4xspaces the same as it does tabs.

SpiritLevel:
Happy New Year SilverStripers! I hope everyone is relaxed and refreshed, especially after the stunning holiday weather we have had here in NZ. 

@tractorcow, @dhensby...quick question: Is the aim to eventually have all SS code strictly following PSR-2 ? Is there a summary or discussion of which parts of PSR-2 you want SS to strictly adhere to ? 
I have noticed that a lot of SS code has 8 spaces but PSR-2 only requries 4, and that PSR-2 requires opening "{" to be on a new line whereas SS typically leaves this on the same line as the class or function declaration. There are other things as well. I am reading a lot of the SS internal code lately and would be happy to convert to PSR-2 as I go. I don't want to irritate you with a flood of PRs so if you let me know what you'd like done, I'd be happy to chip away at it :) Or, are you going to get helpfulrobot onto it ?

tractorcow:
> Is the aim to eventually have all SS code strictly following PSR-2 ?

Yes, but it'll be a gradual adoption rather than immediate. I expect some non-psr-2 names will stick around for the time being.

Framework is currently not psr-2 at all, but we'll do a reformat before releasing 4.0.

SpiritLevel:
OK...Master only :) 

Is it just me, or do other people cringe at the sight of camel case as well ? Same with putting "{" on new lines. I much prefer K&R style bracing. Oh well, these are minor and there are benefits to conforming...

tractorcow:
I think the benefit of adopting psr-2 is having to avoid all those conversations. :) Someone somewhere has already argued about all these boring things, and psr-2 is the result.

SpiritLevel:
I know, I know...I just had to get my two cent's worth in!   :P  From this point forward I'll be a religious zealot, and follow the most important commandment: Thou shalt not put any other coding styles before PSR-1,2,....

tractorcow:
The only annoying part is that we'd recently re-named some methods to FOLLOW the silverstripe convention. If we'd left them, they'd be closer to psr-2 than they are now. :)

SpiritLevel:
That reminds me of Dan's recent comment about fixing a bug and then coming across implementations that require the bug! Double handling is painful not only in coding but also in moving furniture, a currently felt pain since I moved house over the holidays and am still unpacking boxes and shifting furniture around.

tractorcow:
Yeah, @dhensby, stop putting bugs in other people's changes. :P

PR #4897, opened 2016-01-07
Removed validation from LookupField as it's readonly

benmanu:
This pull request covers an issue fixed in pull request #4851 (cherry picked the changes). The affect of not including this is that when a CMS editor reverts a sitetree page a validation error is thrown on null dropdown fields, core cms feature issue. Have discussed this with @chillu and got his ok.

### DIFF
--- a/forms/LookupField.php
+++ b/forms/LookupField.php
@@ -80,6 +80,24 @@ class LookupField extends DropdownField {
 	}
 
 	/**
+	 * Ignore validation as the field is readonly 
+	 *
+	 * @param Validator $validator
+	 * @return bool
+	 */
+	public function validate($validator) {
+		return true;
+	}
+
+	/**
+	 * Stubbed so invalid data doesn't save into the DB
+	 *
+	 * @param DataObjectInterface $record DataObject to save data into
+	 */
+	public function saveInto(DataObjectInterface $record) {
+	}
+
+	/**
 	 * @return LookupField
 	 */
 	public function performReadonlyTransformation() {


tractorcow:
Hi @benmanu I'm not sure it's safe to merge this into 3.2, as it's not a strict bugfix, but could be considered an API change.

@chillu want to double check my thinking here?

dhensby:
Seems like a bug fix over an API change. However, why is saveInto a NOOP? Why shouldn't read onlys be saved into the object?

tractorcow:
I think it's because, as a formfield, it contains content which is posted backfrom the frontend, which should not be allowed to be edited. Unfortunately, there isn't another way of maintaining the state of the formfield easily between postbacks (outside of maybe, a session state?) we need to handle re-posted data to maintain the visible state of this field.

On the other hand, we should never trust this data, and thus should not be saving it back into any dataobject, otherwise malicious users could use HTML injection to modify fields that are readonly.

dhensby:
hmmm is this not a bit of a catch-22 then?

tractorcow:
Kind of, yeah.

dhensby:
Leave on 3.3, I think. It's a new API. If someone needs the bug fix they can upgrade to 3.3

dhensby:
ok! this needs to go in. what I didn't appreciate is that this breaks roll backs in the CMS to old versions of SiteTree objects

dhensby:
Actually, I can't replicate a problem in 3.2 :S

tractorcow:
Let's not change API in 3.2 unless we absolutely have to. :D

Issue #4372, opened 2015-07-06
RFC-4: Client-side dependency management


flashbackzoo:
# RFC-4: Client-side dependency management

Author: @flashbackzoo
Status: Pending review
Version: 0.3

## Purpose and outcome

This RFC proposes the addition of a client-side package manager to the Framework. This would provide a level of automation around managing client-side dependencies such as jQuery, its associated plugins, and other third-party JavaScript libraries.

The implementation of a build tool-chain is outside the scope of this RFC but is mentioned due to the package manager’s ability to include dev dependencies and run build scripts. A secondary outcome of client-side dependency management is the ability to implement a client-side build tool-chain in the future.

### Client-side dependencies are difficult to maintain.
Currently we have to manually check upstream repositories for updates, which is inefficient, given the number of libraries we use. This manual process causes a number of issues including: libraries becoming out-of-date, the inability to identify compatibility issues between libraries, and the introduction of technical debt when libraries are customised in the Framework codebase.

An example of an out-of-date library is jsTree, which is used for the SiteTree expand / collapse and drag-and-drop functionality. There are currently a number of bugs raised on GitHub relating to this functionality, which could be resolved by upgrading the library from our current version v1.0-rc3, to the latest version 3.1.1. The ability to automatically check for updates would enable us update libraries more regularly, perhaps every point release, which would improve the overall quality of the product for end users and make developing the Framework much easier.

Given our current manual approach to updating client-side dependencies, there is no easy way to determine what libraries or what versions of libraries, are incompatible with each other. For example if we manually upgrade jQuery, we have to manually check each library that depends on jQuery, to ensure they’re compatible with the new version. A package manager solves this by distilling a single set of compatible libraries.

Some libraries we use have been customised within Framework which makes upgrading them almost impossible. An example of this is Chosen.js which we use for dropdowns. Using a package manager eliminates this technical debt by forcing developers to fix library bugs upstream. And implement framework specific fixes outside of the library.

### Inconsistent build artifacts

Tools such as Compass are managed by developers in their local environments. We often get conflicting output of sprites and stylesheets because people have different versions installed. Managing dev dependencies with a package manager would mean everybody is on the same version and yield more consistent build artifacts.

### Manage client-side dependencies with npm

As of SilverStripe 4.0 we use [npm 3.x](https://github.com/npm/npm/releases/tag/v3.0.0) to manage JavaScript dependencies.

[...]

PR #4898, opened 2016-01-07
Adds npm for managing client-side dependencies

flashbackzoo:
This is an implementation of https://github.com/silverstripe/silverstripe-framework/issues/4372

Not all libraries are handled by npm yet. We're still on really old versions of libraries aren't available on npm and/or there is no `package.json` file on GitHub. More recent versions _are_ available on npm / GitHub and we've got stories in our internal backlog to upgrade these old libraries.

So this PR includes...

- A package.json file with build and CI scripts
- Shrinkwrap file
- Gulp for running build tasks
- A 'build' task for copying library files from node_modules to thirdparty
- A 'sanity' task for makes sure library files in thirdparty match what's in node_modules
- Updates to .travis.yml for running the sanity task as part of CI


sminnee:
If node_modules is ignored, how can users of SilverStripe use it without installing NPM? Only people developing on SilverStripe should need to install NPM.

I believe that the solution to this is to check node_modules into the repo, and have some kind of post-commit hooks for this.

gulpfile.js
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// Make sure the version of Node being used is valid.
+if (!semver.satisfies(process.versions.node, packageJson.engines.node)) {
+    console.error('Invalid Node.js version. You need to be using ' + packageJson.engines.node);
+    process.exit(1);
+}
+
+gulp.task('build', function () {
+    copyFiles(blueimpFileUploadConfig);
```
sminnee: Will we need extra lines in here for every library? That seems suboptimal.

flashbackzoo: Yeah we'll need one line per library. I did consider this and think it's the best solution, given copying files to thirdparty will become obsolete once we've upgraded our library versions, and https://github.com/silverstripe/silverstripe-framework/issues/4606 is implemented.

The other option was to have an object (or even better a separate json file) which has file / path mappings for all libraries, and everything in there gets copied to thirdparty automagically. I felt like having an extra config file was over engineering seeing it's a stopgap solution that will be super-seeded fairly soon. But I'm happy to reconsider if others feel the same way...

sminnee: Ok, if it's just a short-term fix, that's okay. Bring on #4606!

sminnee:
Given the implications of this change, I'd like to see 3 core committers give this a :+1: before we merge.

:+1: from me.

tractorcow:
Why can't I see travis results for this PR? Is the .travis.yml broken?

tractorcow:
Yep, it's borked. :dancer: 
image

.travis.yml
```diff
@@ -8,6 +8,7 @@ addons:
       - tidy
 
 env:
+  - TRAVIS_NODE_VERSION="4.2.0"
```
tractorcow: Invalid yml here; That needs to go into global:

.travis.yml
```diff
@@ -41,6 +44,7 @@ before_script:
  - php ~/travis-support/travis_setup_php54_webserver.php --if-env BEHAT_TEST
 
 script:
+ - npm run sanity
```
tractorcow: Do we need this every run? Why not just on, for example, BEHAT_TEST = 1 ?

.travis.yml
flashbackzoo: Ahh thanks. I'll fix that up.

.travis.yml
flashbackzoo: Hmmm it's not really related to Behat but I see what you mean. Maybe a new variable `JAVASCRIPT_TESTS` (or something similar) which runs the sanity check, and when we have them, JS unit tests?

flashbackzoo:
@tractorcow I've fixed `.travis.yml` and added a `JAVASCRIPT_TEST` flag to run the sanity check. This can also be used for JS unit tests when we have them :smile:

package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "silverstripe-framework",
+  "version": "4.0.0",
+  "description": "The SilverStripe framework",
+  "directories": {
+    "doc": "docs",
+    "test": "tests"
+  },
+  "engines": {
+    "node": "~4.2.0"
```
chillu: Assuming node follows semver, isn't this too strict? We should require `~4`, so anything lower than 5.x.x.

docs/en/05_Contributing/01_Code.md
```diff
@@ -262,6 +262,38 @@ Helpful hint: You can always edit your last commit message by using:
 
 	$ git commit --amend
 
+## Working with client-side dependencies
+
+From time to time client-side dependencies like jQuery need to be upgraded, added, or removed. We have some tools for dealing with that.
```
chillu: Make it clearer that this relates to core dependencies, and that users are free to use their own separate library versions in their own build systems.

package.json
flashbackzoo: Node ~4.2.0 is LTS. That seemed like a good baseline...

docs/en/05_Contributing/01_Code.md
flashbackzoo: Cool - I'll update this.

chillu:
Wow, that leaves a lot of JS deps to be upgraded, either because our release is too old and not tracked via NPM, or because we've "forked" the dep by making local modifications in `thirdparty/*`. Assuming that we won't completely rewrite the UI for 4.0, we'll still need important libraries like jQuery UI or Chosen.js though, so that work will have to be done at some point. But that doesn't block merging this PR, just flagging that the majority of work is still ahead of us. Here's a complete list from our internal backlog:

image

So a +1 from me.

flashbackzoo:
@chillu I've updated the docs to make it clear the tooling only applies to core SilverStripe dependencies.

I also updated the error message you get when using an invalid Node.js version. It now includes 'If you want to manage multiple Node.js versions try https://github.com/creationix/nvm'

chillu:
Sweet, three core committer votes (Sam, Damian, Ingo), merging!

PR #4899, opened 2016-01-07
Fix: Allow TinyMCE from-web image inserts to use https scheme.

wjagodfrey:
A [bug fix](https://github.com/silverstripe/silverstripe-framework/pull/556/files) for a [url validation issue](http://open.silverstripe.org/ticket/7494#no1) made four years ago rendered CMS authors unable to add images from-web using the https scheme.

This is troublesome for some government agencies that have https-only sites - this fix was requested by Pharmac for an https-only site they maintain.

### DIFF
--- a/javascript/HtmlEditorField.js
+++ b/javascript/HtmlEditorField.js
@@ -1136,9 +1136,9 @@ ss.editorWrappers['default'] = ss.editorWrappers.tinyMCE;
 
 			validate: function() {
 				var val = this.val(), orig = val;
-				
+
 				val = $.trim(val);
-				val = val.replace(/^https?:\/\//i, '');
+
 				if (orig !== val) this.val(val);
 
 				this.getAddButton().button(!!val ? 'enable' : 'disable');
@@ -1158,8 +1158,17 @@ ss.editorWrappers['default'] = ss.editorWrappers.tinyMCE;
 				var urlField = this.getURLField(), container = this.closest('.CompositeField'), form = this.closest('form');
 
 				if (urlField.validate()) {
+
+					var val = urlField.val(), orig = val;
+
 					container.addClass('loading');
-					form.showFileView('http://' + urlField.val()).done(function() {
+
+					// add "http://" if a protocol is missing from the url
+					if (val.match(/^https?:\/\//i) == void 0) {
+						val = 'http://' + val;
+					}
+
+					form.showFileView(val).done(function() {
 						container.removeClass('loading');
 					});
 					form.redraw();
--- a/forms/HtmlEditorField.php
+++ b/forms/HtmlEditorField.php
@@ -350,7 +350,7 @@ class HtmlEditorField_Toolbar extends RequestHandler {
 		$fromWeb = new CompositeField(
 			new LiteralField('headerURL',
 				'<h4>' . sprintf($numericLabelTmpl, '1', _t('HtmlEditorField.ADDURL', 'Add URL')) . '</h4>'),
-			$remoteURL = new TextField('RemoteURL', 'http://'),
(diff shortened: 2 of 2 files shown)


javascript/HtmlEditorField.js
```diff
@@ -1158,8 +1158,17 @@ ss.editorWrappers['default'] = ss.editorWrappers.tinyMCE;
...
 
 				if (urlField.validate()) {
+
+					var val = urlField.val(), orig = val;
+
 					container.addClass('loading');
-					form.showFileView('http://' + urlField.val()).done(function() {
+
+					// add "http://" if a protocol is missing from the url
+					if (val.match(/^https?:\/\//i) == void 0) {
```
jonom: Maybe this pattern should be `/^(https?:)?\/\//i` so that we can support scheme-less protocols as well? I know it's considered a bit of an anti-pattern now but some people may still want to use it?

javascript/HtmlEditorField.js
```diff
@@ -1158,8 +1158,17 @@ ss.editorWrappers['default'] = ss.editorWrappers.tinyMCE;
...
+						val = 'http://' + val;
```
jonom: This could potentially be scheme-less `//` too but I can't decide if it's a better default than `http://` or not...

jonom:
:+1: I've noticed this issue before too, definitely a good idea to fix it up. Would you mind posting a screenshot of the interface after this change? Does it look broken at all with the 'http://' label removed?

javascript/HtmlEditorField.js
tractorcow: Allowing scheme-less protocols for this feature could be a security risk; I'd rather mandate that it was included.

For instance, there are cases wher $_SERVER['REQUEST_URI'] could be populated with a relative url that looks like a scheme-less absolute url.

http://www.mysite.com//page would have a REQUEST_URI val of //page, which when placed directly in a link would act as an absolute url (not a relative url).

In any case, although I'm not sure where this could surface exactly, it's another risk you need to investigate if you allow users to omit protocol. :P

It could also mess with fixes to the above security issue, which might in some cases mistakenly evaluate those urls to relative ones.

tractorcow: On the other hand, then you have risks with mixed content warnings. :) Oh well...

tractorcow:
I remember having to address this problem with subsites

https://github.com/silverstripe/silverstripe-subsites/pull/221

Maybe a better alternative is to have a "default" protocol (and require it), but allow the protocol to be re-written when rendered? (maybe a config value).

javascript/HtmlEditorField.js
jonom: Well the thinking is now that if you can link to a https resource you should - so if you're going to use `//` you would be better off enforcing `https://`.

I'm guessing in the current state of this PR scheme-less URLs would have `http://` added to give you `http:////`? So maybe what we should do instead is at least detect scheme-less URL input here and upgrade it to the `https://`  if found?

jonom:
> Maybe a better alternative is to have a "default" protocol (and require it), but allow the protocol to be re-written when rendered? (maybe a config value) to prevent mixed content warnings.

But if we're talking about external images you may have hotlinked images residing on several different domains - so you can't assume they all support SSL connections.

Having schemeless protocols or forcing SSL ensures no mixed content warnings but you could also get broken images if the target server doesn't support SSL - that's probably worse than a content warning I guess?

jonom:
The best solution: don't hotlink :grinning:

maxime-rainville:
I just tried in on SilverStripe 4.3. The "insert media from URL" WYSIWYG feature works for images with HTTP and HTTPS. So this bug doesn't affect SilverStripe 4.

Issue #4875, opened 2015-12-23
Media tabs retaining browser default visual focus


feejin:
Not really a problem, just not very pretty. The tabs can't be navigated with TAB (at least not in Safari with TAB navigation enabled) so having a :focus outline seems redundant anyway.

Occurs to Chrome 47.0.2526.106 and Safari 9.0.2, does not occur in Firefox 42.

![safari](https://cloud.githubusercontent.com/assets/1774129/11978314/74a4c7f2-a982-11e5-9fbb-765591ce69e1.png)

jonom:
I happened to be looking at this same issue yesterday. It seems to be a jQuery UI 'feature' but I find it bizarre. That styling is caused by a css class which is added by javascript, rather than a `:focus` pseudo-class. It occurs in a few places throughout the CMS.

Ideally we'd remove these styles but I'm not sure it will be trivial. Then we could apply a similar effect with a `:focus` style but nullify it with an `:active` style so you get a blue outline when tabbing around, but not when you click something.

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-framework/pull/4900

PR #4900, opened 2016-01-08
Remove the focus outline when clicking on tabs

scott1702:
When using tab navigation the outline appears on the `li` element, but after clicking on a tab the outline appears on the `a` element. see https://github.com/silverstripe/silverstripe-framework/issues/4875 for more information/pics


jonom:
I'm not sure about this... I agree this would make the appearance of clicking around the navigation better, but I feel like this might be treating the symptoms rather than the root cause (jQuery UI)? Tabbed navigation doesn't work (try tabbing to Settings or History) and it seems weird in the beginning that we're attaching behaviour to the li element of the tabs instead of the a element.

Rather than treat this as just a visual bug and add more styles to sweep it under the rug I'd much rather we address the bigger issues if possible. Do we need to use jQuery UI's implementation of tabs? Couldn't we roll our own and use native pseudo-classes for styling their hover/focus/active states? Then we could address some other issues - like at the moment when you click a tab it will become active regardless of whether or not that content actually loads. (Example: try editing some content on a page then switching to the Settings tab without saving. Click cancel on the prompt and the Settings tab is highlighted even though you're actually viewing the Content tab)

Does anyone know if we'll still be using jQuery UI in SS 4.0? If we're dropping it or re-doing the tabs from scratch there then I'd say this PR makes sense because it might be a wasted effort to do a more complete fix. But if we're sticking with the current approach then adding more styles seems like it might make it harder to work with later if we want to address the bigger issues.

scott1702:
So I had a look into this and the tab navigation isn't actually broken, its just made so you press the arrow keys to change tabs once you have focussed on a tab. Eg tab to 'Content' and press the right arrow key to switch to 'Settings' (This seems to be a jQuery UI thing).

For 4.0 I believe we will still be using jQuery UI, however in future versions I can imagine it being slowly replaced with React components.

jonom:
Oh right, that's interesting - thanks for checking it out. I've just tried this and left/right indeed change the focused link, but that's not a standard behaviour for keyboard navigation and I don't think it's intuitive/obvious is it? I don't know how accessible the CMS aims to be but as far as I know assistive technologies rely on the ability to navigate through all interactive items on a page through just the tab key so that might be an issue. Also jQuery UI tabs are automatically loaded when you focus on them which is certainly not a standard keyboard navigation behaviour.

Even if we continue to use jQuery UI I think it would be best to stop using their version of tabs and roll our own, so we could have native tab-based keyboard navigation and state styling and fix some quirks (like active tab being updated even if not successfully loaded). It doesn't seem like it would be a particularly big/hard job, but that said I don't have time to do it right now and if you don't either we're probably best just merging this in as a cosmetic fix in the mean time? :P

scott1702:
I don't think I'll have time to fix it up either, but I agree it would be better to create our own. I'm happy for this to go in now and I'll make a note to have a go at some new tabs in the future when I get the time :smile:

jonom:
Sounds good :) thanks

PR #4907, opened 2016-01-11
Display all roles when an admin, filter by OnlyAdminCanApply when not

d-rivera-c:
There's a logic problem when trying to apply Roles to Groups. When the `OnlyAdminCanApply` checkbox for Roles is check, the Roles don't show up in the dropdown for Groups when the logged in user is an admin. It does appear when the logged in user is not an admin.
screen shot 2016-01-11 at 14 37 16

screen shot 2016-01-11 at 14 37 59

This doesn't make any sense, so I'm just reverting the logic there. When the user is an admin it will display all Roles, when not, it will filter by the `OnlyAdminCanApply` attribute.

### DIFF
--- a/security/Group.php
+++ b/security/Group.php
@@ -183,7 +183,7 @@ class Group extends DataObject {
 
 			// Add roles (and disable all checkboxes for inherited roles)
 			$allRoles = PermissionRole::get();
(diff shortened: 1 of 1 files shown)


tractorcow:
I think you're right. ;D

PR #4908, opened 2016-01-11
FIX: Update field IDs for file link (fixes silverstripe/silverstripe-cms#1307)

kinglozzer:
Also added a small snippet to remove an extra border that’s left hanging around


kinglozzer:
Will take a look at how easy it is to get behat coverage for this

kinglozzer:
https://github.com/silverstripe/silverstripe-cms/pull/1361

Issue #4901, opened 2016-01-08
[FORMS] OptionsetField returns invalid HTML


torleif:
OptionsetField renders a name="ElementName" in the <ul> element, this tag is invalid HTML. One fix would be to exclude the name element when rendering the html:

**framework/forms/OptionsetField.php**

```
    public function getAttributesHTML($attributes = null) {
        $exclude = null;

        if(is_string($attributes)) {
            $exclude = func_get_args();
        }

        if(!$attributes || is_string($attributes)) {
            $attributes = $this->getAttributes();
        }

        $attributes = (array) $attributes;

        $attributes = array_filter($attributes, function ($v) {
            return ($v || $v === 0 || $v === '0');
        });

        if($exclude) {
            $attributes = array_diff_key(
                $attributes,
                array_flip($exclude)
            );
        }

        // Create markup
        $parts = array();

        foreach($attributes as $name => $value) {
            if($name == "name") {
                continue;
            }
            if($value === true) {
                $parts[] = sprintf('%s="%s"', $name, $name);
            } else {
                $parts[] = sprintf('%s="%s"', $name, Convert::raw2att($value));
            }
        }

        return implode(' ', $parts);
    }
```

tractorcow:
Could we just do a smaller filter by overriding getAttributes(), remove that key, and return the result? It would save having to duplicate that whole method. :D

tractorcow:
```php
public function getAttributes() {
	$attributes = parent::getAttributes(),
	unset($attributes['name']);
	return $attributes;
}
```

torleif:
@tractorcow exactly what I did when I wrote an extension on OptionsetField to avoid modifying the framework. I'll write up a PR tomorrow

torleif:
I'm also checking for required as it will also insert to the \<ul\>. The required attribute is obviously broken, one fix would be to check it in the OptionSet.ss file and place it in each input tag. I'm wondering however if optionset could do with an extra method that will be called on each input tag in the loop?

robbieaverill:
This was fixed in https://github.com/silverstripe/silverstripe-framework/pull/4914 for SilverStripe 3 and was merged up into 4

PR #4910, opened 2016-01-11
FIX  OptionsetField returns valid HTML #4901

torleif:


tractorcow:
Replaced by https://github.com/silverstripe/silverstripe-framework/pull/4912

PR #4912, opened 2016-01-11
FIX OptionsetField returns valid HTML #4901

torleif:


forms/OptionsetField.php
```diff
@@ -102,4 +102,19 @@ public function validate($validator) {
...
+	public function getAttributes() {
+		$attributes = parent::getAttributes();
+		unset($attributes['name']);
+		if(isset($attributes['required'])) {
+			unset($attributes['required']);
+		}
+		if(isset($attributes['role'])) {
+			unset($attributes['role']);
+		}
+		if(isset($attributes['aria-required'])) {
```
tractorcow: How about doing something like: http://stackoverflow.com/questions/34300154/is-aria-required-allowed-on-radio-buttons-or-checkboxes

We'd need to put the options into a fieldset, though.

forms/OptionsetField.php
```diff
@@ -102,4 +102,19 @@ public function validate($validator) {
 	public function ExtraOptions() {
 		return new ArrayList();
 	}
+	
+	public function getAttributes() {
+		$attributes = parent::getAttributes();
+		unset($attributes['name']);
+		if(isset($attributes['required'])) {
+			unset($attributes['required']);
```
tractorcow: You don't need to check isset() before doing unset(). It's a safe operation. :)

forms/OptionsetField.php
torleif: could we could do something smart, like checking if there is a fieldset as a parent and then set the aria-required on it?

tractorcow: I think if we build it in, it'll probably not need the check.


```
<fieldset $AttributesHTML>
    <ul>
	<% loop $Options %>
		<li class="$Class">
			<input id="$ID" class="radio" name="$Name" type="radio" value="$Value"<% if $isChecked %> checked<% end_if %><% if $isDisabled %> disabled<% end_if %> />
			<label for="$ID">$Title</label>
		</li>
	<% end_loop %>
</ul>
</fieldset>
```

tractorcow: Is it semver-safe to wrap a HTML element in a fieldset? Maybe a front-ender could help answer that question. :) @jonom ?

torleif: On second thought checking the parent would be a dumb idea, as the parent fieldset container could contain non required elements.

torleif: looking into this a little deeper, according to the w3 spec: *Attribute aria-required is unnecessary for elements that have attribute required.*

So aria-required seems to be now superfluous. thus the simple fix here: https://github.com/torleif/silverstripe-framework/commit/31fe519ec3b3d7d05cc422e7f574552127f7e007

should be able to resolve the required issue 

Note the required element only has to be on the first element, and I've done it for all. I could add a <% if First %> ... on there too.

tractorcow: I've done a bit of digging, and it seems (at first glance) that the correct thing to do is apply aria-required / required properties to the actual `<input />` elements.

http://jqueryvalidation.org/files/demo/radio-checkbox-select-demo.html

Maybe that's outside the scope of this fix though; Adding it to the fieldset would at least assist with screen-reader devices, which was addressed in my original link. :) It must be _somewhere_ on the formfield.

tractorcow: Ah, you responded too soon.

Yes, I think you might be right there. :)

PR #4914, opened 2016-01-11
FIX OptionsetField returns valid HTML #4901

torleif:


tractorcow:
Dupe of https://github.com/silverstripe/silverstripe-framework/pull/4912?

torleif:
appologies, it's a new pull request to reflect the comment you made here: 4912#discussion_r49387032

tractorcow:
Shall we leave out the unset on aria-required for now? It's not strictly invalid html to include it on a UL is it?

We will definitely need to allow users to apply this property to formfields where it's necessary, especially sites that have strict accessibility requirements.

torleif:
@tractorcow sweet, I've popped back in the aria-required

tractorcow:
Thanks @torleif

dhensby:
Without the correct markup an aria-required in the wrong place is invalid html

tractorcow:
Yeah, that's right... I probably need to find out exactly what is valid and do this across fields.

dhensby:
Yes, I had to make a lot of changes to mark-up for an accessible site I built - I should try to go back and look at it all.

the short of it is: our forms are all wrong.

tractorcow:
The problem is that setAttribute() doesn't let you say WHERE the attribute goes. For each field, the target element is different.

dhensby:
Erm, set attribute always works on the form input, I think... Its the extra classes that goes all over the place.

tractorcow:
Except in optionsetfield, which doesn't have one `<input>` element. Or grouped fields, etc.

PR #4919, opened 2016-01-12
Adding control to PDOConnector over stringifying returned values

dhensby:
At the moment we don't provide control over the `ATTR_STRINGIFY_FETCHES` attribute for PDO connections.

This plays a central role in correctly casting returned values from the DB,

### DIFF
--- a/src/ORM/Connect/PDOConnector.php
+++ b/src/ORM/Connect/PDOConnector.php
@@ -187,14 +187,10 @@ class PDOConnector extends DBConnector
             $charset = $connCharset;
         }
         $options = array(
-            PDO::MYSQL_ATTR_INIT_COMMAND => 'SET NAMES ' . $charset . ' COLLATE ' . $connCollation
+            PDO::MYSQL_ATTR_INIT_COMMAND => 'SET NAMES ' . $charset . ' COLLATE ' . $connCollation,
+            PDO::ATTR_EMULATE_PREPARES => self::is_emulate_prepare(),
+            PDO::ATTR_STRINGIFY_FETCHES => self::is_stringify_fetches(),
         );
-        if (self::is_emulate_prepare()) {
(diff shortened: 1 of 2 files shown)


kinglozzer:
Test failures :( looks like the test is using `preparedQuery()` instead of just `query()`. Apart from that this looks good, false is a sensible default IMO. I’m assuming you submitted this to master in case anyone is relying on the returned values being strings?

tractorcow:
Yeah, that one fix will correct the issue. :)

On the other hand, you will need to ensure your tests work if stringify is set to true. For that matter, what OTHER tests fail if stringify is true? :P

tractorcow:
http://stackoverflow.com/questions/10103846/php-pdo-with-native-mysql-driver-mysqlnd-fetch-data-retaining-native-colum

We might not be able to use this with mysql I'm afraid. =(

tractorcow:
On the other hand, http://stackoverflow.com/questions/20079320/php-pdo-mysql-how-do-i-return-integer-and-numeric-columns-from-mysql-as-int seems to have some success.

dhensby:
OK - looks like my test passes now but another is now failing.

there was a line that would only set `ATTR_EMULATE_PREPARES` to true if the config was true (but not false if config was false), this meant that the value was never enforced to be false and instead remained true (I assume the default).

So I found a bug... YAY

kinglozzer:
@tractorcow Yeah that second link is the one I read - whether it will work seems to depend on whether you’re using `libmysql` or `mysqlnd`

dhensby:
I don't think this is going to work out. Too many DB drivers don't even support the native typing as it is and this isn't passing tests so there's no value in solving the problem this way for a minority of the number of users...

It'd be better to get the ORM to force the type for data that is passed from the DB to the model...

PR #4902, opened 2016-01-08
DOCS: Added tutorial for installing silverstripe with Vagrant and Virtualbox

flamerohr:
The original document is here, which I had converted to markdown from:
https://docs.google.com/document/d/14wnHHSaGjul5y3YYkbNBdhkLl2UauPMoWUSukkVZm5o/edit#

PR #4922, opened 2016-01-12
DOCS: Added tutorial for installing silverstripe with Vagrant and Virtualbox

flamerohr:
[This pull request](https://github.com/silverstripe/silverstripe-framework/pull/4902) for branch 3.2


dhensby:
It's great that we have some good docs to encourage people to use vagrant for dev. 

However, I do think this is a bit to manual and technical and something like [my vagrant skeleton](https://github.com/dhensby/vagrant-skeleton) is a better approach to encouraging this kind of thing. 

There are a few bits where these docs can be made more efficient (`vagrant unit boxcutter/centos7` for example) and simplify the docs

flamerohr:
Fair point, I was thinking about setting up a git repo where people could
just install from and away they go.
Similar to how laravel had the homestead vagrant environment.

dhensby:
Yep, that's kind of what I was going for bit less fancy ;)

PR #4924, opened 2016-01-13
i18n::default_locale should return the value set via config.

### LINKED ISSUE #4920: Setting i18n::default_locale via config file has no effect.
Setting the default locale in a YAML config file has no effect on `i18n::get_locale()` or `i18n::default_locale()`.
Example YAML:
```yml
i18n:
  default_locale: 'fr_FR'
```

The reason seems to be that `i18n::default_locale()` never accesses `Config` but directly returns the private static.
Proper behavior can be achieved by adding the following to `_config.php`:
```php
i18n::set_default_locale(Config::inst()->get('i18n', 'default_locale'));
```

Is this intended behavior? If not, I'll gladly create a PR that fixes getter and setter for `default_locale`.

bummzack:
Deprecate `default_locale` setter and getter methods, analoguous to other config getters and setters.
Fixes #4920.

### DIFF
--- a/i18n/i18n.php
+++ b/i18n/i18n.php
@@ -2150,7 +2150,7 @@ class i18n extends Object implements TemplateGlobalProvider, Flushable {
 			self::$translators[$defaultPriority] = array(
 				'core' => new Zend_Translate(array(
 					'adapter' => 'i18nRailsYamlAdapter',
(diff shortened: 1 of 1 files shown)


tractorcow:
Perfect.

PR #4925, opened 2016-01-13
FIX Emulate Prepares now will set to false explicitly

dhensby:
Emulate prepares config is ignored if it's false (which it looks like we want it to be).

I assume the expectation was the default is false and we should only set it if it's true, but this appears to be incorrect and the default is true (in PHP 5.4 at least).


tractorcow:
You've un-ravelled the ball of yarn. The worms can is open. Do you want to fix all these issues too? :P

It might be better to just leave it as is. :P

dhensby:
Leave it broken or fix it? I choose FIX IT!!! but I'm just not sure what on earth is causing this error :/

tractorcow:
Can you fix it so that PDO::MYSQL_ATTR_USE_BUFFERED_QUERY is automatically set depending on the value of is_emulate_prepare() ?

dhensby:
@tractorcow I guess - is this connector solely for the use of MySQL

dhensby:
Looks like no improvement :/

tractorcow:
No it's used for lots of connectors; Postgres / sqlite etc should all have pdo support.

PR #4928, opened 2016-01-14
DOCS 3.0 : fixes broken internal links

SpiritLevel:
Following up on rewrite rules (see silverstripe/doc.silverstripe.org#91)  :)

Note:

(1) links to missing changelogs have been deleted (4 in all).
(2) links to .md files that do not exist and have no obvious alternative are also deleted (2 in all).

### DIFF
--- a/docs/en/changelogs/index.md
+++ b/docs/en/changelogs/index.md
@@ -23,7 +23,7 @@ For information on how to upgrade to newer versions consult the [upgrading](/ins
  * [2.4.7](2.4.7) - 1 February 2012
  * [2.4.6](2.4.6) - 18 October 2011
  * [2.4.10](2.4.10) - 2013-02-19
- * [2.4.9](2.4.9) - 2012-12-04
+ * 2.4.9 - 2012-12-04
  * [2.4.8](2.4.8) - 2012-10-30
  * [2.4.7](2.4.7) - 2012-02-01
  * [2.4.6](2.4.6) - 2011-10-17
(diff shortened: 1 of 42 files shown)


docs/en/changelogs/index.md
```diff
@@ -23,7 +23,6 @@ For information on how to upgrade to newer versions consult the [upgrading](/ins
  * [2.4.7](2.4.7) - 1 February 2012
  * [2.4.6](2.4.6) - 18 October 2011
  * [2.4.10](2.4.10) - 2013-02-19
- * [2.4.9](2.4.9) - 2012-12-04
```
tractorcow: How about we make these non-links, but leave the info? I think it's good to have the release history there.

docs/en/reference/grid-field.md
```diff
@@ -7,7 +7,7 @@ It's built in a way that provides developers with an extensible way to display t
 table and minimise the amount of code that needs to be written.
 
 In order to quickly get data-focused UIs up and running,
-you might also be interested in the [/reference/modeladmin](ModelAdmin) class
```
tractorcow: We should keep the case the same as the correct class shouldn't we?

SpiritLevel: Absolutely...also, it should be [ModelAdmin](/reference/modeladmin) :) Fixed.

docs/en/changelogs/index.md
SpiritLevel: I agree. :) Added back in...

PR #4929, opened 2016-01-14
API Enable single javascript files to declare that they include other files

tractorcow:
I wrote this to assist @flashbackzoo initially since he's using browserify to do front-end combination.

There are a few limitations as documented in the code; The major one is that there's NO way to block "provided" scripts, since you can't un-include a file that is pre-combined.

You also cannot retro-actively provide a file; You need to include your combined files prior, in order to ensure subsequent single-includes are skipped.

It was a pain to get this working with combined files; Refactor at your own risk!

### DIFF
--- a/view/Requirements.php
+++ b/view/Requirements.php
@@ -110,9 +110,11 @@ class Requirements implements Flushable {
 	 * Register the given JavaScript file as required.
 	 *
 	 * @param string $file Relative to docroot
+	 * @param array $options List of options. Available options include:
+	 * - 'provides' : List of scripts files included in this file
 	 */
-	public static function javascript($file) {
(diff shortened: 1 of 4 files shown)


view/Requirements.php
```diff
@@ -1044,90 +1106,90 @@ public function unblockAll() {
 	 * @return string HTML content augmented with the requirements tags
 	 */
 	public function includeInHTML($templateFile, $content) {
```
tractorcow: Most of this code is just un-nesting the giant if() block.

sminnee:
Looks reasonable.

tractorcow:
Fixed up your feedback on the docs.

chillu:
Looks good!

PR #4933, opened 2016-01-15
DOCS 2.4 : added back inactive links to missing changelogs

SpiritLevel:
Following up on silverstripe/silverstripe-framework/pull/4928 for 2.4 docs.


SpiritLevel:
Regarding the unsquashed 613836d above: I left this unsquashed since it is really a separate task (it deals with external links, not internal ones) but similar enough to group with the prior commit. If you would prefer this commit as a separate PR, let me know :)

Many of the broken links to external sites (ie: links to sites other than docs.silverstripe.org) are to api.silverstripe.org and suffer from silverstripe/silverstripe-docsviewer/issues/89 or they are to open.silverstripe.org and need to be sorted according to silverstripe/doc.silverstripe.org/issues/101.
I opted to leave these links in as text but disable them, as requested in silverstripe/silverstripe-framework/pull/4928.

A minority of the broken external links could be kept active but would require a bit of hunting down and I felt that this might be wasting time for an older version such as 2.4. I am thinking of approaching 3.0 external links the same way. But, for 3.1 and up, I would take the time to keep as many external links as possible active by finding suitable updates. 

If people feel there is a need, I can go back later and hunt down updated links for 2.4 and 3.0, but my main goals at the moment are to (i) get the .htaccess for docs.silverstripe.org sorted and (ii) get the links for docs 3.1+ working without 404s. 

Let me know if you are happy with this approach :)

Thanks, D.

tractorcow:
This seems like you did a lot of work, but it's quite a nice approach! Thanks for all the effort you've made (and been making).

SpiritLevel:
It would be great to fix this since it affects all the docs: silverstripe/silverstripe-docsviewer#89 

I haven't had a chance to dig into this yet. Looks like it would be quick to fix. Can someone point me to the right spot and I'll give it a go...

SpiritLevel:
@tractorcow If we can get https://github.com/silverstripe/silverstripe-docsviewer/pull/90 merged in then I can put back all those api links back in :)

PR #4937, opened 2016-01-18
Implement RFC-7 JavaScript module loader

flashbackzoo:
This is an implementation of https://github.com/silverstripe/silverstripe-framework/issues/4606

Changes:
- Updated JavaScript folder structure
- Adds ES6 support via Babel
- Transforms existing JavaScript to UMD modules
- Adds module bundling via Browserify
- lib.js and leftandmain.js are bundled using browserify
- JavaScript minifying of bundles handed by gulp
- Throw an exception when Requirements are not found
- Updated source files to work in 'strict' mode

## Updated JavaScript folder structure
Source files are now located at `admin/javascript/src` and `javascript/src`. These are transformed (see below) and output to the `admin/javascript/dist` and `javascript/dist` directories respectively. Both `src` directories have been added to `.gitattributes` with `export-ignore`.

## Build tooling
There are three npm scripts available, each invokes a Gulp task.

### thirdparty

`$ npm run thirdparty`

This was previously called `build` and has been renamed so we can use the `build` name for something else (see below). Thirdparty functions exactly the same as previously, it copies libraries from `node_modules` to `thirdparty`.

### sanity

`$ npm run sanity`

This is unchanged.

### build

`$ npm run build`

The `build` task is responsible for two things, generating UMD modules, and generating bundles. The build task automatically watches source files and regenerates dist files as needed. You also get a handy system notification if you break the build.

When developing locally you can include the development flag `$ npm run build --development` which will include source maps for easy debugging.

#### Generating UMD modules

All JavaScript source files are transformed from ES6 to ES5 by Babel and output to the `dist` directory as UMD modules.

#### Generating bundles

The `leftandmain.js` and `lib.js` combined files are no longer managed by the `Requirements` class. They are now managed by Browserify. Bundles are minified as part of build process.

## Throw an exception when Requirements are not found

Because the folder structure has changed, and because it's probably a good idea in general, an exception is thrown when a Requirement is missing. I found a few existing bugs when I added this, files that had been removed but the `Requirement` was still there, and typos in path names.

## Updated source files to work in strict mode

Browserify and Babel add 'use strict' to output files. This caused a few issues due to things like using undeclared variables. I fixed these as needed.


flashbackzoo:
Related CMS change https://github.com/silverstripe/silverstripe-cms/pull/1367

tractorcow:
I'm not sure if renaming admin/javascript/lib.js to admin/javascript/src/custom.js improves readability. "custom.js" is a bit generic you must admit? Why not something like "ss-url-lib.js" or something?

tractorcow:
Also, please ensure you upgrade any upgrading notes in the changelog. For instance, if users are relying on fixed locations for certain files, then they might need to update their Requirements calls.

flashbackzoo:
Thanks @tractorcow I'll make those changes.

flashbackzoo:
@tractorcow I've made those changes :smile:

tractorcow:
Rebase please. :) I've merged up from 3 branch.

tractorcow:
Fixed the bugs!

tractorcow:
@dhensby I've reviewed this and want to merge it, but now that I've touched it I can't merge it as well. Want to do the honours?

dhensby:
NEEDS A REBASE!

tractorcow:
Sheesh, again!

tractorcow:
Ok, rebased. :D

chillu:
- We should apply the same principles (gulp build tooling, browserify, ES6 import statements) in the CMS module as well, otherwise it's a bit inconsistent
- The *.js files are removed and re-added rather than using `git mv` - which means all their history will be lost. Not great given the relative complexity of these files. It also masks the actual changes in the files. It's going to be a pain in the ass, but I think important to get fixed.
- Should exclude minified files from diff by default (http://blog.andrewray.me/dealing-with-compiled-files-in-git/)
- The 4.0.0 changelog didn't emphasise enough that in a normal project, nothing has to change - people will get freaked if they now require NPM and Browserify for their stuff (which they don't). I've amended the changelog accordingly.

Great to see that the overall JS on-the-wire size (after gzip) went down from 1.2MB to 1.1MB just by using a better compression mechanism.

tractorcow:
> We should apply the same principles (gulp build tooling, browserify, ES6 import statements) in the CMS module as well, otherwise it's a bit inconsistent

Agreed: @flashbackzoo can you make this change please? I don't think it's a blocker to this PR though, but it will depend on this getting merged first.

> The *.js files are removed and re-added rather than using git mv - which means all their history will be lost. Not great given the relative complexity of these files. It also masks the actual changes in the files. It's going to be a pain in the ass, but I think important to get fixed.

http://stackoverflow.com/questions/1094269/whats-the-purpose-of-git-mv

It won't make a difference; if git doesn't detect the move now, it won't then.

> Should exclude minified files from diff by default (http://blog.andrewray.me/dealing-with-compiled-files-in-git/)

I'll do this. :)

> The 4.0.0 changelog didn't emphasise enough that in a normal project, nothing has to change - people will get freaked if they now require NPM and Browserify for their stuff (which they don't). I've amended the changelog accordingly.

I'll also make this update.

flashbackzoo:
> We should apply the same principles (gulp build tooling, browserify, ES6 import statements) in the CMS module as well, otherwise it's a bit inconsistent

Yeah good idea, I can make that change. I'm on holiday through until February 1st, so won't be able to do it until then.

PR #4939, opened 2016-01-18
Minor: Unbolded labels in CMS

lry6003:
- updated _forms.scss removed font-weight
- also removed drop-shadow include
Luke Ryley - open source academy


tractorcow:
It looks like you are pulling to a different branch you created this from... do you need a rebase?

scott1702:
Nice! Happy with this change @tractorcow

tractorcow:
What's the rationale for un-bolding labels in the CMS? This feels like it needs some discussion before I can justify merging it. :)

jonom:
@lry6003 would you mind posting a before and after screenshot please?

clarkepaul:
This was under my direction to minimize elements within the interface. There are too many competing elements in the current interface and labels are only a small part of that. The bold labels stand out a lot and the focus on the page should be more on the fields themselves.

clarkepaul:
This is what I'm expecting (below). I would hope that it has had a bit of testing to make sure it makes sense with a range of scenarios. 

label-weight

jonom:
That's a nice improvement

tractorcow:
Sounds good.... merging. ;0

PR #4950, opened 2016-01-19
Minor: changed px value on tabs (border-bottom)

divya-patel:
Changed px value of 4 to 3px on border-bottom for tabs.


scott1702:
Looks like this needs to be compiled @divya-patel :)

tractorcow:
Merged with https://github.com/silverstripe/silverstripe-framework/commit/7ead6f0ca9bba27676fbaf34e4ccecc26ba17baa

PR #4951, opened 2016-01-20
Minor:updated the menu on the top left; changed styles

divya-patel:
screenshot from 2016-01-20 14 05 11 after styles added

screenshot from 2016-01-20 14 06 17 before 

@clarkepaul


clarkepaul:
Looks good, added a minor suggestion to making the blue a variable so it can be re-used in other places. It looks a little out of place without the rest of the menu changed yet but now we have an incentive to continue the look down. Looking much brighter and friendlier already.

tractorcow:
Restarting build: A fix was merged into the postgres module.

kinglozzer:
Green ticks, looks great! :+1:

PR #4952, opened 2016-01-20
Install sake into /usr/local/bin. /usr is now protected on OSX

wilr:
https://support.apple.com/en-us/HT204899


wilr:
Bar missed the label. Shouldn't have had a strict regex

tractorcow:
You need to fix the message ` 	echo "Installing sake to /usr/local/bin..."`

PR #4953, opened 2016-01-20
Search Icon Hover/Active/Focus changes

lry6003:
Changed the color of the search icon when it is in Hover/Active/Focus state and removed background color changes in these states. @clarkepaul


clarkepaul:
Looks good, cheers

PR #4954, opened 2016-01-20
BUG Fix VersionedTest sometimes failing given certain querystring arguments

tractorcow:
It looks like this test was susceptible to state set by prior tests.


dhensby:
gosh, why are we messing with global state like that :(

tractorcow:
Some tests don't clean up after themselves, it seems. =/

A new global state mechanism is necessary; It'll have to be 4.x. Hopefully one that overrides the necessity for any global references to $_GET / $_POST etc.

tractorcow:
To fix with https://silverstripe.uservoice.com/forums/251266-new-features/suggestions/6425996-core-state-handling-to-fix-subsites-translatable

PR #4957, opened 2016-01-20
NEW $belongs_to relations to be used in $searchable_fields of gridfie…

RVXD:
GridFieldAddExistingAutocompleter uses the $searchable_fields from DataObject.
If you use fields of related objects to search in, you can use it like this:
$searchable_fields = array(
'Relation.Field'=>'Field'
);
This works if Relation is a $has_one field, but not if it is a $belongs_to field.
The $belongs_to fields are supported if used in $summary_fields, but strangely not if used in $searchable_fields.
This change adds the needed extra leftJoin to enable searching in $belongs_to.

### DIFF
--- a/forms/gridfield/GridFieldAddExistingAutocompleter.php
+++ b/forms/gridfield/GridFieldAddExistingAutocompleter.php
@@ -214,6 +214,17 @@ class GridFieldAddExistingAutocompleter
 			->sort(strtok($searchFields[0], ':'), 'ASC')
 			->limit($this->getResultsLimit());
 
+		$belongs_to = Config::inst()->get($dataClass, 'belongs_to');
+		if( !empty($belongs_to) ) {
+			foreach ($searchFields as $searchField) {
+				if( $relation = strtok($searchField,'.') ) {
+					if (in_array($relation, $belongs_to)) {
+						$results = $results->leftJoin($relation, "{$relation}.{$dataClass}ID={$dataClass}.ID");
+					}
+				}
+			}
+		}
+
 		$json = array();
 		$originalSourceFileComments = Config::inst()->get('SSViewer', 'source_file_comments');
 		Config::inst()->update('SSViewer', 'source_file_comments', false);


forms/gridfield/GridFieldAddExistingAutocompleter.php
```diff
@@ -214,6 +214,17 @@ public function doSearch($gridField, $request) {
 			->sort(strtok($searchFields[0], ':'), 'ASC')
 			->limit($this->getResultsLimit());
 
+		$belongs_to = Config::inst()->get($dataClass, 'belongs_to');
+		if( !empty($belongs_to) ) {
+			foreach ($searchFields as $searchField) {
+				if( $relation = strtok($searchField,'.') ) {
+					if (in_array($relation, $belongs_to)) {
+						$results = $results->leftJoin($relation, "{$relation}.{$dataClass}ID={$dataClass}.ID");
```
tractorcow: Please ANSI quote your join condition. :D

`"\"{$relation}\".\"{$dataClass}ID\" = \"{$dataClass}\".\"ID\""`

Also, $dataClass isn't the correct key; You need to look up the relation name on the other end.

E.g. a belongs_to will have a has_one on the other side, and the name of the has_one will determine the database field to join on.

You can use DataObject::getRemoteJoinField to get this.

tractorcow:
Thanks for the pull request; I've left a note on two things that need fixing.

Are you also able to please add a unit test to show that it works on belongs_to? :P Ta!

RVXD:
Thanks for the suggestions. Still working on the tests, is appears there's more work to be done to get this right. When running the tests it seems the belong_to fields are not included in the where condition.

Currently it's like this:
```php
		$belongs_to = Config::inst()->get($dataClass, 'belongs_to');
		if( !empty($belongs_to) ) {
			foreach ($searchFields as $searchField) {
				if( strpos($searchField, '.') !== false && $relation = strtok($searchField,'.') ) {
					if (!empty($belongs_to[$relation])) {
						$relationClass = $belongs_to[$relation];
						$remoteJoinField = $dataClass::create()->getRemoteJoinField($relation,'has_one');
						$results = $results->leftJoin($relationClass, "\"{$relation}\".\"{$remoteJoinField}\" = \"{$dataClass}\".\"ID\"",$relation);
					}
				}
			}
		}
```

tractorcow:
>                         $remoteJoinField = $dataClass::create()->getRemoteJoinField($relation,'has_one');


you need to use the type of field you are joining FROM. In this case, `belongs_to` not `has_one`

tractorcow:
Any response to my feedback? I haven't seen any updates in a few months.

RVXD:
I haven't found the time to figure this one out yet. Ran into some issues with the tests, after that I ran out of time...

dhensby:
@RVXD can you push what you've got so far in the way of tests/fixes?

maxime-rainville:
I think the root of the problem is that the search doesn't honour the dot donation for `belongs_to` relationship. I tried to check the behaviour is still present in SS4 and it's still a problem. I raise an issue for SS4: https://github.com/silverstripe/silverstripe-framework/issues/8719

PR #4958, opened 2016-01-20
Remove question marks, equal signs and hashes

Taitava:
Those were not removed if multibyte was allowed.


tractorcow:
Could you please include a test in URLSegmentFilterTest with setAllowMultibyte(true) and some of those characters?

Taitava:
Ok, I did that, but apparently it came up as a new pull request #4963. I thought that GitHub would combine the pull requests when creating a new one while having another still open. Anyway, is this good now?

tractorcow:
Are you able to please, maybe, try putting them into the same PR? Otherwise the tests won't pass with the fix. You'll need to switch to your patch-1 branch in the github editor and put the changes there.

dhensby:
see #4973

PR #4960, opened 2016-01-20
Add special flag to CMS preview requests

IgorNadj:
This affects the iframe used in the CMS for page previews. With this change, we can have the page behave differently based on whether it's a preview or not. The page itself would have to check for the parameter, nothing is done automatically.

Our specific use case is to prevent redirecting to another page, as it would normally. However this may also be useful in e.g. plastering a PREVIEW message, or having other preview-specific content.

### DIFF
--- a/admin/javascript/LeftAndMain.Preview.js
+++ b/admin/javascript/LeftAndMain.Preview.js
@@ -434,6 +434,9 @@
 					// No state available at all.
 					this.setCurrentStateName(null);
 				}
+				
+				// Mark url as a preview url so it can get special treatment
+				url += ((url.indexOf('?') === -1) ? '?' : '&') + 'CMSPreview=1';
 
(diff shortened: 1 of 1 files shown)


tractorcow:
It's not going to work, however, if you navigate to another page from the preview, is it? E.g. viewing page A, click page B in the menu, and the CMS interface + preview updates to page B.

Also, since this is a new feature, you'll need to put this in the 3.4 branch, since 3.1 is only accepting critical bugfixes (and 3.2 and 3.3 is only bugfixes too).

IgorNadj:
I've just tested it quickly on my machine and it sends CMSPreview=1 in the scenario you describe, where you navigate to another page in the site tree.

Have re-created PR here: https://github.com/silverstripe/silverstripe-framework/pull/4967

Thanks

tractorcow:
Closed with https://github.com/silverstripe/silverstripe-framework/pull/4967

PR #4963, opened 2016-01-21
Test removing some bad characters when multibyte is allowed

Taitava:


tractorcow:
Thanks @dhensby for the re-make with tests. Also thanks to @Taitava for the fix. :D

Closed with https://github.com/silverstripe/silverstripe-framework/pull/4973

Taitava:
Thank you @tractorcow and @dhensby! :) And sorry for the inconvenience with the two pull requests, I will try to do it better next time! :)

Taitava:
Hi, I can't find the original issue related to this pull request, so I'm writing my suggestion here.

I think we should add `:` to the list of removable characters when multibyte characters are allowed in URL segments. What do you think? If this sounds like a good idea, I can create a new issue for this and a pull request also. I'm using SS 4.3.3.

So in practise:
```diff
class URLSegmentFilter
{
    /**
     * @config
     * @var array See {@link setReplacements()}.
     */
    private static $default_replacements = array(
        '/&amp;/u' => '-and-',
        '/&/u' => '-and-',
        '/\s|\+/u' => '-', // remove whitespace/plus
        '/[_.]+/u' => '-', // underscores and dots to dashes
        '/[^A-Za-z0-9\-]+/u' => '', // remove non-ASCII chars, only allow alphanumeric and dashes
-       '/[\/\?=#]+/u' => '-', // remove forward slashes, question marks, equal signs and hashes in case multibyte is allowed (and non-ASCII chars aren't removed)
+       '/[\/\?=#:]+/u' => '-', // remove forward slashes, question marks, equal signs, hashes and colons in case multibyte is allowed (and non-ASCII chars aren't removed)
        '/[\-]{2,}/u' => '-', // remove duplicate dashes
        '/^[\-]+/u' => '', // Remove all leading dashes
        '/[\-]+$/u' => '' // Remove all trailing dashes
    );
```

dhensby:
@Taitava I'm sure if you open a new PR with the change and your reasoning, it'll be considered

Taitava:
Thanks @dhensby . Does the pull request need to reference any issue? Or is it enough to reference just the old pull request #4958? I guess I didn't create an issue when I made the old pull request in 2016, but I red from the contribution guide (or somewhere) that it would be good for pull requests to refer an issue.

dhensby:
I think if your PR is low effort (changing 1 line and a test) then you can roll the issue report and PR into one. You can reference this issue for completeness, though I'm not sure it's entirely useful as there's no significant description of the problem in this issue.

PR #4964, opened 2016-01-21
API Standardise SS_List::map() implementation

### LINKED ISSUE #1593: [2012-09-17] DataList::map() and ArrayList::map() return different types
created by: @frankmullenger (frankmullenger)
created at: 2012-09-17
original ticket: http://open.silverstripe.org/ticket/7869

---

DataList::map() returns an SS_Map object:
https://github.com/silverstripe/sapphire/blob/3.0/model/DataList.php#L605

ArrayList::map() returns an array:
https://github.com/silverstripe/sapphire/blob/3.0/model/ArrayList.php#L245

Ideally both methods would have the same return type for consistency.

tractorcow:
Fixes #1593

### DIFF
--- a/model/ArrayList.php
+++ b/model/ArrayList.php
@@ -275,21 +275,13 @@ class ArrayList extends ViewableData implements SS_List, SS_Filterable, SS_Sorta
 	/**
 	 * Returns a map of this list
 	 *
-	 * @param string $keyfield - the 'key' field of the result array
-	 * @param string $titlefield - the value field of the result array
-	 * @return array
+	 * @param string $keyfield The 'key' field of the result array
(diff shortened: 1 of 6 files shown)


dhensby:
@tractorcow you need to fix the test in `MemberTest::testMap_in_groupsReturnsAdmins` that relies on old behaviour

tractorcow:
That's why I should run all tests before pushing... :)

tractorcow:
Pushed up again.

dhensby:
thank goodness, at last we have this crappy inconsistent API fixed!!! :)

tractorcow:
`$fristItems`

PR #4968, opened 2016-01-22
BUG Fix Versioned::doRollbackTo

### LINKED ISSUE #4936: Wrong Version creation on DataObjects with inherited Versioned extension
From my understanding Versioning is inherited. If a DataObject, that has the Versioned extension applied, is extended, than the extended DataObject should be versioned as well. Unfortunately the VersionedTest has insufficient test cases for this. There are some border cases which result in errors, if the child does not declare the Versioned extension itself.

There are two subclasses used in the VersionedTest:
* VersionedTest_Subclass (Versioned extension reapplied)
* VersionedTest_AnotherSubclass (Versioned extension not reapplied)

Adapting VersionedTest#testHasOwnTableDatabaseFieldWithVersioned to use VersionedTest_AnotherSubclass instead of VersionedTest_Subclass results in an error. If the VersionedTest#testRollbackTo method is switched to use VersionedTest_Subclass, it results in incosistent Versioned table entries, where the latest version number in the tables VersionedTest_AnotherSubclass and VersionedTest_DataObject does not match.

tractorcow:
Document correct configuration of Versioned DataObjects
Fixes #4936

Super special thanks to @sb-relaxt-at  for helping with this. I've tried my best to keep your original test cases as demonstration of this issue. :)

The main problem is that _nextWriteWithoutVersion was getting ignored at certain times, and a _versions manipulation was getting created anyway. I've moved that action into an else so it always blocks the manipulation creation.

### DIFF
--- a/docs/en/02_Developer_Guides/00_Model/10_Versioning.md
+++ b/docs/en/02_Developer_Guides/00_Model/10_Versioning.md
@@ -27,6 +27,11 @@ The extension is automatically applied to `SiteTree` class. For more information
 [Extending](../extending) and the [Configuration](../configuration) documentation.
 </div>
 
+<div class="warning" markdown="1">
+Versioning only works if you are adding the extension to the base class. That is, the first subclass
+of `DataObject`. Adding this extension to children of the base class will have unpredictable behaviour.
+</div>
+
 ## Database Structure
 
 Depending on how many stages you configured, two or more new tables will be created for your records. In the above, this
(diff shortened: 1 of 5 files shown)


tractorcow:
Again, fixing in 3.2 branch, but need to get this merged in and then up to 3.3 branch for 3.3.0-beta1 release.

tractorcow:
Missing extradataobjects in some test fixtures... added and re-pushed.

tractorcow:
This might be easiest to review in split mode... the diffs came out a bit weird.

PR #4940, opened 2016-01-18
Minor: changed background color

divya-patel:
> - I added another header color, sidebar color and center color
> first open source fix made by divya for silverstripe
image
@clarkepaul

PR #4970, opened 2016-01-22
Update background colors

scott1702:
The colours in the designs have changed once more since https://github.com/silverstripe/silverstripe-framework/pull/4940. 
Now looks like 
image


tractorcow:
Thanks. :D

Issue #4971, opened 2016-01-22
No obvious “back” action for GridFieldDetailForm (3.3)


kinglozzer:
Tentatively adding this as an issue...

When editing an item in a `GridField`, there is no obvious way of getting back to the parent record/page. The only way back is via the link in the breadcrumb trail, which is pretty small and hard to spot:

![screen shot 2016-01-22 at 10 47 16](https://cloud.githubusercontent.com/assets/1655548/12508732/8fe1ed12-c0f5-11e5-9c14-d0a14108c795.png)

The breadcrumb looks nicer (and admittedly, re-adding the back button does kinda spoil the look - see below), but I’m concerned it’s not as easy to use. It took me far too long to spot!

![screen shot 2016-01-22 at 10 56 17](https://cloud.githubusercontent.com/assets/1655548/12508924/c831dfbe-c0f6-11e5-97b0-830ccbe34e11.png)

I’ll look at re-adding the button but with the new button styles and see how that looks

kinglozzer:
Closed via d339642cac303e45fdb5e912a110cb233f1c7b59

PR #4972, opened 2016-01-22
Re-instate back button for item edit forms

kinglozzer:
See #4971.

Before:
screen shot 2016-01-22 at 10 47 16

After:
screen shot 2016-01-22 at 11 18 37

cc @clarkepaul @scott1702


kinglozzer:
I did try this with the new button styles too, but it just looked a bit odd to me...

screen shot 2016-01-22 at 11 13 48
screen shot 2016-01-22 at 11 14 02

It’s easy enough to switch if that style is preferred as the CSS doesn’t need to change, just the HTML for the anchor

tractorcow:
I'll let the aforementioned people comment on this... although they are away on holiday today. ;D

tractorcow:
Where else does this change add the back button? Does it add it in the page edit area, or on model admins?

tractorcow:
Actually, if I remember correctly, the back button was considered superseded by the ability to use the breadcrumbs and/or your browser's back button. "Back" really just takes you to the parent breadcrumb, which you should be able to still do.

kinglozzer:
@tractorcow Yeah, that’s right - the breadcrumbs “replace” it in a sense, my concern is just that with the reduced font size there’s no clear way of getting back to the parent item like there used to be (besides the browser back button of course)

This is just my concern and proposal for how I’d solve it, so if anyone disagrees then feel free to close this and the related issue :)

kinglozzer:
> Where else does this change add the back button? Does it add it in the page edit area, or on model admins?

It adds it to the `GridFieldDetailForm` - so it’ll appear in a typical ModelAdmin, and any GridFields attached to pages. `LeftAndMain_EditForm.ss` sounds quite generic, but the `$BackLink` template var comes from `GridFieldDetailForm::getBackLink()`

clarkepaul:
On your example @kinglozzer , when you click back are you on the "Product" view? If in this example the breadcrumb doesn't change then I would recommend the following back arrow but larger.

pasted_image_26_01_16__9_51_am

kinglozzer:
@clarkepaul On the example, both the “Product” link and the back button will take you back to the `ModelAdmin` interface, but it could equally take you back to another edit form.

That arrow isn’t currently part of the icon font or anything, so I tried a similar effect with the existing arrow:

cc4xhzzxlm

tractorcow:
@clarkepaul are you happy with this?

clarkepaul:
Lets keep the back the button but can we change the appearance slightly. @scott1702 is going to add the proper icon to the set which is bigger, we don't need the background colour change on hover but the icon itself will get darker on hover (it should be the same pattern as the search icon).

scott1702:
@kinglozzer icons have been added, along with a .no-text class which can be added to font-icons to changes their hover/active/focus state https://github.com/silverstripe/silverstripe-framework/pull/5009

kinglozzer:
@scott1702 @clarkepaul :heart:

Rebased and switched to the new arrow, looks like this:

screen shot 2016-02-04 at 09 13 42

Let me know if you think the font-size for the icon needs adjusting (above is the default, 16px):

18px
screen shot 2016-02-04 at 09 17 30

20px
screen shot 2016-02-04 at 09 17 37

22px
screen shot 2016-02-04 at 09 17 47

tractorcow:
@scott1702 can you check this please?

clarkepaul:
Thanks @kinglozzer 22px looks good, it doesn't need so much visual space left and right of the icon. To keep the hit point large you might be able to apply a negative margin on the left side. Nice one

kinglozzer:
Done and done! :)

screen shot 2016-02-05 at 09 38 27

scott1702:
Nice! Looking good :)

dhensby:
looking good == merge

kinglozzer:
Thanks @scott1702 and @clarkepaul for your help on this one :heart:

SpiritLevel:
Just reflecting on how much effort goes into perfecting each and every little detail of SS. OSS is such a fascinating process to observe. :)

clarkepaul:
Smoooooth

Issue #4909, opened 2016-01-11
RSSFeed_Entry should be instantiated via ::create()


patricknelson:
To facilitate better abstraction/customization of RSS feed contents and mark-up, the following method in `RSSFeed` should create an instance of it's related `RSSFeed_Entry` object via `::create()`:

https://github.com/silverstripe/silverstripe-framework/blob/3/api/RSSFeed.php#L151
```php
$output->push(
	new RSSFeed_Entry($entry, $this->titleField, $this->descriptionField, $this->authorField));
```

The only way around that is to override `Entries()` method and copy/paste the full contents and modify the affected line. I can submit a PR for this later this week when I really dive into it, but wanted to drop the issue in now so I don't forget it for later.

tractorcow:
Yep I agree. :)

patricknelson:
@tractorcow Ok finally getting into this. This is minor, is it fine to apply this to `3.2` branch?

patricknelson:
Hmm I see we're on the cusp of `3.3`, so -- there instead or maybe `3`?

tractorcow:
Ok, but PR it soon because RC1 is going out soon.

patricknelson:
Ok, I'll take that to mean it should go against `3.3`. I'll see what I can do.

tractorcow:
Yes. :D and thank you!

patricknelson:
Ok, please @tractorcow  sorry for the huge delay. Please see here: https://github.com/silverstripe/silverstripe-framework/pull/4974

Had to push this out now since I won't be available for any changes for the next week so don't wait on me to merge this (obviously) since I cannot do anything else for quite some time :smile: I may have overstepped a bit (see `->outputToBrowser()`) but hopefully that's not a huge deal. Let me know, please.

PR #4974, opened 2016-01-22
FIX for #4909: Ensure RSSFeed_Entry is instantiated using the injector.

patricknelson:
For #4909

Ensure RSSFeed_Entry is instantiated using the injector. Also extending `->outputToBrowser()` to loosen dependence on global state for easier testing/prototyping so dev can inject custom SS_HTTPResponse instance. There's already a lot of tight binding to global state, but this at least prevents it from crashing in case there is no "current controller" (yes I've setup a small framework to breakway from SS's standard paradigm to make it easier for me to prototype/test code in a totally isolated environment). Almost like unit tests, but not quite as formal :+1:

### DIFF
--- a/api/RSSFeed.php
+++ b/api/RSSFeed.php
@@ -149,7 +149,7 @@ class RSSFeed extends ViewableData {
 		if(isset($this->entries)) {
 			foreach($this->entries as $entry) {
 				$output->push(
-					new RSSFeed_Entry($entry, $this->titleField, $this->descriptionField, $this->authorField));
+					RSSFeed_Entry::create($entry, $this->titleField, $this->descriptionField, $this->authorField));
 			}
 		}
 		return $output;
@@ -184,7 +184,11 @@ class RSSFeed extends ViewableData {
 	}
 
 	/**
-	 * Output the feed to the browser
+	 * Output the feed to the browser.
+	 *
+	 * TODO: Pass $response object to ->outputToBrowser() to loosen dependence on global state for easier testing/prototyping so dev can inject custom SS_HTTPResponse instance.
+	 *
+	 * @return	HTMLText
 	 */
 	public function outputToBrowser() {
 		$prevState = Config::inst()->get('SSViewer', 'source_file_comments');


dhensby:
@tractorcow are we ok to put this API change into 3.3 or will it need to be 3?

api/RSSFeed.php
```diff
@@ -184,13 +184,17 @@ public function Description() {
...
 
 	/**
-	 * Output the feed to the browser
+	 * Output the feed to the browser.
+	 *
+	 * @param	SS_HTTPResponse|null	$response
+	 * @return	HTMLText
 	 */
-	public function outputToBrowser() {
+	public function outputToBrowser(SS_HTTPResponse $response = null) {
```
tractorcow: Ah, this is not a compatible change I'm afraid... this can't even go into 3.x if this is the change that's necessary. It'll have to go into master.

tractorcow:
@dhensby we aren't treating new arguments in public methods as non-breaking... subclasses that override those methods would suddenly develop strict method signature declaration errors otherwise.

@patricknelson sorry please can you re-open on master?

patricknelson:
I'll just update it to remove the irrelevant change, no big deal. I can reintroduce that into master instead, if need be.

tractorcow:
3.3 is in RC now, so, you won't be able to get this into the release I'm afraid! =(

You'll still need to PR to the 3 branch from now on, I think. 3.3 branch is only for bugfixes, which this is technically not.

dhensby:
@tractorcow I think replacing `new X` with `X::create()` should be allowed into patch releases. The risk is incredibly low and really all object instantiation should be like that (so it kind of maybe sort of is a bug)

in reality the risk is so low and the reward to a dev is so high I, personally, see no need to block that single change from 3.3

tractorcow:
> I think replacing new X with X::create()  should be allowed into patch releases. 

@chillu what's your feeling on this?

dhensby:
Does @chillu have feelings?

dhensby:
How about @kinglozzer or @willmorgan?

kinglozzer:
Should be 3 IMO. The chance of any regressions is negligible, but it’s definitely not a bug - it’s a new feature. I think we have to stick to being strict with semver, plus it’s more of an incentive to speed up releases than if we sneak low-risk things into patch releases.

api/RSSFeed.php
willmorgan: Agreed, sadly. Also, IMHO, the `$response` parameter should be required so that we get explicit and consistent results rather than falling back on a singleton.

willmorgan:
I consider swapping instantiation to using `::create` as a patch. Notwithstanding my comments the optional `$response` parameter in `outputToBrowser` code should be in a separate PR, and ideally, it should be made required.

patricknelson:
@willmorgan's last point is my current goal. I would have already done all of this if I weren't so darned busy this week.

tractorcow:
so we are exactly 50-50 back where we last were. :)

Since there's no consensus, let's favour merging.

tractorcow:
Please remove `+	public function outputToBrowser(SS_HTTPResponse $response = null) {` so I can merge please.

patricknelson:
I'm not going to be able to do anything until next week I'm sorry (in the middle of some stuff personally and I'm too busy to get to my computer to reliably do it anytime soon). Maybe by Sunday. If not then it will be Tuesday. 


- Patrick Nelson

tractorcow:
Tuesday is fine. :)

patricknelson:
This can still be against 3.3 correct? Been a while. Just making sure since I know 4 is on the horizon

tractorcow:
No, it needs to go into 3 (if you remove the change we noted), or master (if you leave it as is).

dhensby:
> Since there's no consensus, let's favour merging.

patricknelson:
In case the only thing remaining is just removing the parameter, then we should now already be set @dhensby (just rebased and updated that line). Otherwise I can still reopen this against `3` now instead of `3.3` per @tractorcow - still confused on that part now.

tractorcow:
It's merged :D

patricknelson:
Wowza, thanks @tractorcow :+1:

PR #4975, opened 2016-01-23
No longer used variable

spekulatius:


tractorcow:
I actually like this fix... are you re-opening this?

spekulatius:
Sure, I can do. I just closed it because of the failing tests - and that I haven't had time to look into this.

dhensby:
The problem is ALL the calls to this function need to be updated - that's a lot of work over ALL the supported modules (and tests). It's a painful API change for little benefit.

I'm all for doing it! (if this var really isn't needed) but just worth noting the effort required

tractorcow:
We could possibly do it with a deprecation message, in case func_num_args() > 1? At least old modules will continue to work.

sminnee:
The failing tests look pretty straightforward: RequirementsTest tests this function and you're not 

Furthermore SSViewer calls this function and you haven't updated that call.

@dhensby do many modules actually call this function, or do they rely on SSViewer to do it?

dhensby:
@sminnee good question - but who cares - I've changed my view, we should get rid of it :P

Though I'd like a cleaner commit history ;)

tractorcow:
I've merged this in with https://github.com/silverstripe/silverstripe-framework/pull/5337

spekulatius:
Cool, that this found its way. Thanks everyone :)

Issue #4179, opened 2015-05-12
QUESTION: getIDList produces unsuspecting outcome, ignoring sort


sunnysideup:
Hi  

I recently an issue on a website due to misunderstanding getIDList. I am not sure if this is a bug or just my lack of understanding - however, in case the latter applies (I am sure it does), it would be good to have clearer comments above the getIDList method so that you know what to expect from it. 

Basically, getIDList ignores the sort that applies to the DataList. 

Here is how to see the bug in action.... take any random large website with a ton of pages.

method:

``` php

    function test(){
        echo "<h1><pre>\$pages = Page::get()->limit(10)</pre></h1>";
        $pages = Page::get()->limit(10);
        $array1 = array();
        echo "<h2>For each DataList Object: </h2>";
        foreach($pages as $key => $page) {
            DB::alteration_message($key." === ".$page->ID);
            $array1[$key] = $page->ID;
        }
        echo "<hr />";
        $pagesList = $pages->getIDList();
        $i = 0;
        $array2 = array();
        echo "<h2>For each <pre>\$pages->getIDList</pre></h2>";
        foreach($pagesList as $key => $pageID) {
            DB::alteration_message($i." === ".$pageID);
            $array2[$i] = $pageID;
            $i++;
        }
        echo "<hr />";
        $map = $pages->map("ID", "ID")->toArray();
        $i = 0;
        $array3 = array();
        echo "<h2>For each <pre>\$pages->map(\"ID\", \"ID\")->toArray()</pre></h2>";
        foreach($map as $key => $pageID) {
            DB::alteration_message($i." === ".$pageID);
            $array3[$i] = $pageID;
            $i++;
        }
        echo "<h2>Comparison between DataList and getIDList</h2>";
        foreach($array1 as $key => $value) {
            if($array2[$key] == $value) {
                DB::alteration_message("MATCH: $key ...  $value = $value", "created");
            }
            else {
                DB::alteration_message("ERROR: $key ... $value != ".$array2[$key]."", "deleted");
            }
        }
        echo "<h2>Comparison between DataList and Map</h2>";
        foreach($array1 as $key => $value) {
            if($array3[$key] == $value) {
                DB::alteration_message("MATCH: $key ...  $value = $value", "created");
            }
            else {
                DB::alteration_message("ERROR: $key ... $value != ".$array3[$key]."", "deleted");
            }
        }
    }
```
# outcome: 
# $pages = Page::get()->limit(10);
[...]

PR #4977, opened 2016-01-24
FIX PHPDocs on DataList::getIDList() and UnsavedRelationList::getIDList

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/4179#issuecomment-174289595

getIDList no longer produces unsuspecting outcome.


dhensby:
Hmm reading up online, it seems that `array_combine` *does* maintain order, the problem is that the keys/values will be cast to int and so they will explicitly set the order of the array.

You'd need to manually loop the array to maintain the order explicitly casting the keys as strings... I *think*

probably not worth the computational overhead, though.

PR #4978, opened 2016-01-25
Removed PlayerCsvBulkLoader as importer

mediabeastnz:
Removed PlayerCsvBulkLoader as model importer class and replaced it with CsvBulkLoader instead.


tractorcow:
Why the change? I thought the intention of this code is to demonstrate how to override the importer with a custom class.

mediabeastnz:
The page starts off explaining how to use CsvBulkLoader without any customizations and then PlayerCsvBulkLoader is used as an example. I just seemed a little bit confusing as the PlayerCsvBulkLoader class doesn't actually get mentioned until right at the bottom of this page.

tractorcow:
Fair enough... it does read that way. :)

PR #4979, opened 2016-01-25
Replaced filterByCallback for 'Children' to just create a new array a…

dhensby:
…s function calls are exponentially expensive in PHP (the more functions that exist, the slower a function call becomes) and replaced 'array_key_exists' with op-code equivalent for speed. The best increase isn't really noticeable but we should work towards optimizing the core as much as possible.

Squashed version of #4884


dhensby:
ping @tractorcow for mergin

PR #4980, opened 2016-01-25
Bumping PHPUnit version

dhensby:
Let's get with it!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ before_script:
 # - printf "\n" | pecl install imagick
  - composer self-update || true
  - phpenv rehash
- - git clone git://github.com/silverstripe-labs/silverstripe-travis-support.git ~/travis-support
+ - git clone git://github.com/dhensby/silverstripe-travis-support.git ~/travis-support
  - "if [ \"$BEHAT_TEST\" = \"\" ]; then php ~/travis-support/travis_setup.php --source `pwd` --target ~/builds/ss; fi"
  - "if [ \"$BEHAT_TEST\" = \"1\" ]; then php ~/travis-support/travis_setup.php --source `pwd` --target ~/builds/ss --require silverstripe/behat-extension; fi"
  - "if [ \"JS_SANITY_CHECK\" = \"1\" ]; then nvm install $TRAVIS_NODE_VERSION && npm install; fi"
--- a/tests/behat/features/bootstrap/SilverStripe/Framework/Test/Behaviour/CmsUiContext.php
+++ b/tests/behat/features/bootstrap/SilverStripe/Framework/Test/Behaviour/CmsUiContext.php
@@ -2,20 +2,15 @@
 
 namespace SilverStripe\Framework\Test\Behaviour;
 
-use Behat\Behat\Context\ClosuredContextInterface,
-	Behat\Behat\Context\TranslatedContextInterface,
-	Behat\Behat\Context\BehatContext,
-	Behat\Behat\Context\Step,
-	Behat\Behat\Exception\PendingException,
-	Behat\Mink\Exception\ElementNotFoundException,
-	Behat\Gherkin\Node\PyStringNode,
-	Behat\Gherkin\Node\TableNode,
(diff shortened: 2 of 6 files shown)


tractorcow:
Get those behat tests passing and we can probably merge.

It's probably worth calling for a vote from @silverstripe/core-team to double check this upgrade is ok.

tractorcow:
See:
- https://github.com/silverstripe-labs/silverstripe-behat-extension/pull/104
- https://github.com/silverstripe/silverstripe-installer/pull/112

dhensby:
The impact of this is surprisingly low - I haven't had to change a single line of a test to get the test suite passing - the only problem now is that behat depends on 3.7 and so proving it works on this PR (getting green ticks) is just really hard

tractorcow:
If you can get it working with behat then let's merge it, then. :D

dhensby:
I need to get these behat tests to run/pass - I've forked and messed around with the travis-test and behat-extensions module but the travis builds aren't checking out the right version of framework to test against, which is nuts.

I'll look into it when I have time

madmatt:
Note that this will likely necessitate some updates to the `hafriedlander/phockito` module, because PHPUnit 4 starts including it's own Hamcrest-style matchers like `any()`, which conflict with Hamcrest, loaded globally by Phockito.

I don't think this is used in core, but for those playing at home, one way to resolve this is to have Phockito not load these globally by calling `Phockito::include_hamcrest(false);`, and then name-spacing the calls to Hamcrest matchers, e.g.
```php
$service = Phockito::mock('MyService');
Phockito::when($service)
    ->addUser(\Hamcrest_Matchers::anything(), \Hamcrest_Matchers::anything())
    ->return(array('status' => 'success'));
```

The key change here is to use `\Hamcrest_Matchers::anything()` instead of just using `anything()`.

dhensby:
thanks @madmatt - noted

tests/behat/features/bootstrap/SilverStripe/Framework/Test/Behaviour/CmsUiContext.php
```diff
@@ -3,19 +3,14 @@
...
-	Behat\Mink\Exception\ElementNotFoundException,
-	Behat\Gherkin\Node\PyStringNode,
-	Behat\Gherkin\Node\TableNode,
-	Behat\Mink\Element\NodeElement;
-
-
-// PHPUnit
-require_once 'PHPUnit/Autoload.php';
-require_once 'PHPUnit/Framework/Assert/Functions.php';
+use Behat\Behat\Context\TranslatedContextInterface,
```
chillu: PHP syntax error (comma instead of semicolon)

chillu:
The `Functions.php` require call is still causing trouble for me. I've created a CMS PR to fix this in the module: https://github.com/silverstripe/silverstripe-cms/pull/1399. Also asked @dhensby to patch his PR for behat-extension at https://github.com/silverstripe-labs/silverstripe-behat-extension/pull/104

.travis.yml
```diff
@@ -34,7 +34,7 @@ before_script:
 # - printf "\n" | pecl install imagick
  - composer self-update || true
  - phpenv rehash
- - git clone git://github.com/silverstripe-labs/silverstripe-travis-support.git ~/travis-support
+ - git clone git://github.com/dhensby/silverstripe-travis-support.git ~/travis-support
```
chillu: Don't forget to revert this before merge

chillu:
Rebased, squashed, and merged through https://github.com/silverstripe/silverstripe-framework/commit/2fdc96a0de33c34b4dbcdeb971cd4ccf319d64a0. Behat and PHPUnit tests were passing locally. It's becoming a bit of a pain to verify passing tests in the pull requests on this one (setting temporary repos etc), so I went ahead and merged. The code had enough core team scrutiny already, and if tests fail on one of the master branches now I'll fix them right away.

dhensby:
thanks @chillu much appreciated - I was going to pull out my hair trying to get this working in travis :P

PR #4981, opened 2016-01-26
Fix incomplete functionality from #3734

JorisDebonnet:
#3734 introduced the ability to *create* `mailto:` links with a subject in them, without the ability to *edit* them. This fixes that.

### DIFF
--- a/javascript/HtmlEditorField.js
+++ b/javascript/HtmlEditorField.js
@@ -888,10 +888,11 @@ ss.editorWrappers['default'] = ss.editorWrappers.tinyMCE;
 					action = "update";
 				}
 
-				if(href.match(/^mailto:(.*)$/)) {
+				if(href.match(/^mailto:([^?]*)(\?subject=(.*))?$/)) {
 					return {
 						LinkType: 'email',
 						email: RegExp.$1,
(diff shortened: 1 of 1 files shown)


javascript/HtmlEditorField.js
```diff
@@ -888,10 +888,11 @@ ss.editorWrappers['default'] = ss.editorWrappers.tinyMCE;
 					action = "update";
 				}
 
-				if(href.match(/^mailto:(.*)$/)) {
+				if(href.match(/^mailto:([^?]*)(\?subject=(.*))?$/)) {
 					return {
 						LinkType: 'email',
 						email: RegExp.$1,
+						Subject: decodeURIComponent(RegExp.$3),
```
tractorcow: Will RegExp.$3 silently failover to empty if there's no subject, or will it cause an error?

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1374

javascript/HtmlEditorField.js
JorisDebonnet: RegExp.$3 will indeed be empty in that case, I have succesfully tested this.

PR #4982, opened 2016-01-26
BUG Fix regressions in #4873

tractorcow:
See  https://github.com/silverstripe/silverstripe-framework/pull/4873

Previously, it did nothing. However, with the prior fix, suddenly extra classes popped up in places which broke certain layouts.

This fix reverts the method so that the old undesirable behaviour remains ONLY in the deprecated method.

### DIFF
--- a/forms/CompositeField.php
+++ b/forms/CompositeField.php
@@ -128,10 +128,7 @@ class CompositeField extends FormField {
 	 */
 	public function extraClasses() {
 		Deprecation::notice('4.0', 'Use extraClass() instead');
-		return $this->extraClass();
-	}
 
-	public function extraClass() {
 		$classes = array('field', 'CompositeField', parent::extraClass());
 		if($this->columnCount) $classes[] = 'multicolumn';
 


dhensby:
So this is basically a revert of the last fix? 

Isn't it better to keep the correct API and fix workarounds for when the classes weren't present?

dhensby:
OK, so this "undesirable" behaviour is literally core standard behaviour that people simply didn't cater for (because they didn't need to).

The fix is to fix the layouts not to declare the standard API undesirable.

If the current behaviour is undesirable, it should be undesirable *everywhere* not just this field.

proper fixes: [framework](https://github.com/silverstripe/silverstripe-framework/pull/4984) and [cms](https://github.com/silverstripe/silverstripe-cms/pull/1376)

tractorcow:
> Isn't it better to keep the correct API and fix workarounds for when the classes weren't present?

This is the correct API; It just leaves the "relic" css classes unused. They were NEVER used, and should not suddenly be used now.

E.g. 'field' and 'CompositeField' classes should not suddenly be appearing in the HTML. In 4.0 I'll drop that method, and eradicate them altogether.

This method doesn't need an extraClass() method, because the parent implementation doesn't need to be overridden. That's why it looks like we are just "going back to using the old method", which is not the case.

tractorcow:
Well, your fix makes the new CSS not break the layout, so it's probably ok to use them from now on.

dhensby:
> E.g. 'field' and 'CompositeField' classes should not suddenly be appearing in the HTML. In 4.0 I'll drop that method, and eradicate them altogether.

I agree - I dislike this incredibly - but it should be consistently terrible - let's get rid of it ASAP

tractorcow:
> I agree - I dislike this incredibly - but it should be consistently terrible - let's get rid of it ASAP

Another pull request? It's not a drama anymore though, so I guess it's not essential.

dhensby:
It's a frustrating problem to solve, we shouldn't really be mixing HTML and PHP (which is what this "extraClasses" stuff does) but it's undeniably useful and less frustrating than a million templates for each instance where you need to add a class.

Finding a nice way to add and remove default classes on a field by field basis would be best. We can already *add* default classes to whole Field types but we can't block them. A lot of the time you only want to add or block them in some instances (like just in the CMS).

PR #4992, opened 2016-01-28
Remove Object::useCustomClass and break Object into traits

tractorcow:
This means that we can now do `use Extendable` in classes that don't extend Object (e.g. subclasses of third party libraries) and we can add extensions to them.

All that's left in Object is a bit of reflection code and a few helper methods; These could possibly go into ClassInfo / ViewableData, then we won't have Object at all. :)

This was pulled out of another piece of work I'm working on, which required extensions on things that aren't subclasses of Object.

Object.php is down from > 1000 lines to < 300 :D

CMS pull request to fix classnames is at https://github.com/silverstripe/silverstripe-cms/pull/1395

### DIFF
--- a/core/Configurable.php
+++ b/core/Configurable.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace SilverStripe\Framework\Core;
+
+use ClassInfo;
+use Config;
+use Extension;
+use Injector;
+
+/**
+ * Provides extensions to this object to integrate it with standard config API methods.
+ *
+ * Note that all classes can have configuration applied to it, regardless of whether it
+ * uses this trait.
+ */
+trait Configurable {
+
+	/**
+	 * Get a configuration accessor for this class. Short hand for Config::inst()->get($this->class, .....).
+	 * @return \Config_ForClass
+	 */
+	public static function config() {
+		return Config::inst()->forClass(get_called_class());
+	}
+
+	/**
+	 * Gets the first set value for the given config option
+	 *
+	 * @param string $name
+	 * @return mixed
+	 */
+	public function stat($name) {
+		return Config::inst()->get(get_class($this), $name, Config::FIRST_SET);
+	}
+
+	/**
+	 * Update the config value for a given property
+	 *
+	 * @param string $name
+	 * @param mixed $value
+	 */
+	public function set_stat($name, $value) {
+		Config::inst()->update(get_class($this), $name, $value);
+	}
+
+	/**
+	 * Gets the uninherited value for the given config option
+	 *
+	 * @param string $name
+	 * @return mixed
+	 */
+	public function uninherited($name) {
+		return Config::inst()->get(get_class($this), $name, Config::UNINHERITED);
+	}
+}
--- a/core/CustomMethods.php
+++ b/core/CustomMethods.php
@@ -1,0 +1,241 @@
+<?php
+
+namespace SilverStripe\Framework\Core;
+
+use BadMethodCallException;
+use InvalidArgumentException;
+
+/**
+ * Allows an object to declare a set of custom methods
+ */
+trait CustomMethods {
+
+    /**
+     * Custom method sources
+     *
+     * @var array
+     */
+	protected static $extra_methods = array();
+
+	/**
+	 * Name of methods to invoke by defineMethods for this instance
+	 *
+	 * @var array
+	 */
+	protected $extra_method_registers = array();
+
+    /**
+     * Non-custom methods
+     *
+     * @var array
+     */
+    protected static $built_in_methods = array();
+
+    /**
+     * Attempts to locate and call a method dynamically added to a class at runtime if a default cannot be located
+     *
+     * You can add extra methods to a class using {@link Extensions}, {@link Object::createMethod()} or
+     * {@link Object::addWrapperMethod()}
+     *
+     * @param string $method
+     * @param array $arguments
+     * @return mixed
+     * @throws BadMethodCallException
+     */
+	public function __call($method, $arguments) {
+		// If the method cache was cleared by an an Object::add_extension() / Object::remove_extension()
+		// call, then we should rebuild it.
+        $class = get_class($this);
+		if(!array_key_exists($class, self::$extra_methods)) {
+			$this->defineMethods();
+		}
+
+		// Validate method being invked
+		$method = strtolower($method);
+		if(!isset(self::$extra_methods[$class][$method])) {
+			// Please do not change the exception code number below.
+			$class = get_class($this);
+			throw new BadMethodCallException("Object->__call(): the method '$method' does not exist on '$class'", 2175);
+		}
+
+		$config = self::$extra_methods[$class][$method];
+
+		switch(true) {
+			case isset($config['property']) :
+				$obj = $config['index'] !== null ?
+					$this->{$config['property']}[$config['index']] :
+					$this->{$config['property']};
+
+				if($obj) {
+					if(!empty($config['callSetOwnerFirst'])) $obj->setOwner($this);
+					$retVal = call_user_func_array(array($obj, $method), $arguments);
+					if(!empty($config['callSetOwnerFirst'])) $obj->clearOwner();
+					return $retVal;
+				}
+
+				if(!empty($this->destroyed)) {
+					throw new BadMethodCallException(
+						"Object->__call(): attempt to call $method on a destroyed $class object"
+					);
+				} else {
+					throw new BadMethodCallException(
+						"Object->__call(): $class cannot pass control to $config[property]($config[index])."
+							. ' Perhaps this object was mistakenly destroyed?'
+					);
+				}
+
+			case isset($config['wrap']) :
+				array_unshift($arguments, $config['method']);
+				return call_user_func_array(array($this, $config['wrap']), $arguments);
+
+			case isset($config['function']) :
+				return $config['function']($this, $arguments);
+
+			default :
+				throw new BadMethodCallException(
+					"Object->__call(): extra method $method is invalid on $class:"
+						. var_export($config, true)
+				);
+		}
+	}
+
+    /**
+	 * Adds any methods from {@link Extension} instances attached to this object.
+	 * All these methods can then be called directly on the instance (transparently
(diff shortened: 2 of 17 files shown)


core/Extendable.php
```diff
@@ -0,0 +1,465 @@
...
+			$this->afterExtendCallbacks[$method] = array();
+		}
+		$this->afterExtendCallbacks[$method][] = $callback;
+	}
+
+    protected function constructExtensions() {
+        $class = get_class($this);
+
+        // Register this trait as a method source
+        $this->registerExtraMethodCallback('defineExtensionMethods', function() {
```
tractorcow: I added this part to de-couple Extendable from CustomMethods a little. It's just a callback so that CustomMethods::defineMethods will invoke defineExtensionMethods.

core/Configurable.php
```diff
@@ -0,0 +1,55 @@
...
+	/**
+     * Gets the uninherited value for the given config option
+     *
+	 * @param string $name
+     * @return mixed
+	 */
+	public function uninherited($name) {
+		return Config::inst()->get(get_class($this), $name, Config::UNINHERITED);
+	}
+}
```
kinglozzer: Indentation is a mix of tabs + spaces in the docblocks in this file

core/Configurable.php
```diff
@@ -0,0 +1,55 @@
...
+ * Note that all classes can have configuration applied to it, regardless of whether it
+ * uses this trait.
+ */
+trait Configurable {
+
+	/**
+	 * Get a configuration accessor for this class. Short hand for Config::inst()->get($this->class, .....).
+	 * @return \Config_ForClass
+	 */
+	static public function config() {
```
kinglozzer: \<pedant\>`public static`?\</pedant\> :D

core/Extendable.php
kinglozzer: So just to confirm:

1. `Object::__construct()` calls `Extendable::constructExtensions()`
2. `Extendable::constructExtensions()` then registers `defineExtensionMethods()` as a callback in `CustomMethods`
3. `Extendable::constructExtensions()` calls `CustomMethods::defineMethods()` which runs through the registered callbacks
4. `defineExtensionMethods()` is called and adds the methods from extensions

Phew! I think this is as good an abstraction as we can hope for, nice job

kinglozzer:
Awesome work!

Test failure looks like it comes from [here](https://github.com/tractorcow/silverstripe-framework/blob/pulls/4.0/cleanup-object/tests/model/DataObjectTest.php#L1795) - would be resolved by #4551 anyway :)

core/Extendable.php
```diff
@@ -0,0 +1,465 @@
...
+use Injector;
+
+/**
+ * Allows an object to have extensions applied to it.
+ *
+ * Bootstrap by calling $this->constructExtensions() in your class constructor.
+ *
+ * Requires CustomMethods trait
+ */
+trait Extendable {
```
kinglozzer: Should this be `Extensible`?

tractorcow: Ah, that's correct!

core/Extendable.php
tractorcow: Yep, and if you just wanted to use ExtraMethods without extensions, you'd need to call defineMethods() instead of constructExtensions(). But yeah, it was a quick de-couple that's not perfect.

core/Configurable.php
tractorcow: Ew my bad. Too much copy-pasting. :)

tractorcow:
I think that test failure is due to my removal of Object::useCustomClass. Easy to fix. :)

tractorcow:
I've fixed the above and replaced the `Datetime` db field with `SS_Datetime` for the mean time in that one test fixture.

core/CustomMethods.php
```diff
@@ -0,0 +1,252 @@
...
+     * @param string $method
+     * @param array $arguments
+     * @return mixed
+     * @throws \Exception
+     */
+	public function __call($method, $arguments) {
+		// If the method cache was cleared by an an Object::add_extension() / Object::remove_extension()
+		// call, then we should rebuild it.
+        $class = get_class($this);
+		if(empty(self::$extra_methods[$class])) {
```
dhensby: `!array_key_exists`?

core/CustomMethods.php
```diff
@@ -0,0 +1,252 @@
...
+
+					if($obj) {
+						if(!empty($config['callSetOwnerFirst'])) $obj->setOwner($this);
+						$retVal = call_user_func_array(array($obj, $method), $arguments);
+						if(!empty($config['callSetOwnerFirst'])) $obj->clearOwner();
+						return $retVal;
+					}
+
+					if(!empty($this->destroyed)) {
+						throw new \Exception (
```
dhensby: What kind of exception? logicexception?

core/CustomMethods.php
```diff
@@ -0,0 +1,252 @@
...
+				default :
+					throw new \Exception (
+						"Object->__call(): extra method $method is invalid on $class:"
+							. var_export($config, true)
+					);
+			}
+		} else {
+			// Please do not change the exception code number below.
+			$class = get_class($this);
+			throw new \Exception("Object->__call(): the method '$method' does not exist on '$class'", 2175);
```
dhensby: How about we create a new exception type rather than relying on that stoopid code?

core/CustomMethods.php
```diff
@@ -0,0 +1,252 @@
...
+	 */
+	protected function defineMethods() {
+		$class = get_class($this);
+
+		// Define from all registered callbacks
+		foreach($this->extra_method_registers as $callback) {
+			call_user_func($callback);
+		}
+
+		if(isset($_REQUEST['debugmethods']) && isset(self::$built_in_methods[$class])) {
```
dhensby: let's remove accessing the super globals

dhensby: let's remove this whole block

dhensby:
I think you need more inline comments. I'm finding parts of it hard to really know what's going on. For example: I can't see any documented spec for the array keys you want in `CustomMethods`

core/Extensible.php
```diff
@@ -0,0 +1,465 @@
...
+	 * Keep in mind that the extension will only be applied to new
+	 * instances, not existing ones (including all instances created through {@link singleton()}).
+	 *
+	 * @see http://doc.silverstripe.org/framework/en/trunk/reference/dataextension
+	 * @param string $classOrExtension Class that should be extended - has to be a subclass of {@link Object}
+	 * @param string $extension Subclass of {@link Extension} with optional parameters
+	 *  as a string, e.g. "Versioned" or "Translatable('Param')"
+	 * @return bool Flag if the extension was added
+	 */
+	public static function add_extension($classOrExtension, $extension = null) {
```
dhensby: why not ditch this in preference of `Config`?

core/Extensible.php
```diff
@@ -0,0 +1,465 @@
...
+	 * permission-checks through extend, as they use min() to determine if any of the returns is FALSE. As min() doesn't
+	 * do type checking, an included NULL return would fail the permission checks.
+	 *
+	 * The extension methods are defined during {@link __construct()} in {@link defineMethods()}.
+	 *
+	 * @param string $method the name of the method to call on each extension
+	 * @param mixed $a1,... up to 7 arguments to be passed to the method
+	 * @return array
+	 */
+	public function extend($method, &$a1=null, &$a2=null, &$a3=null, &$a4=null, &$a5=null, &$a6=null, &$a7=null) {
```
dhensby: Yucky :(

core/CustomMethods.php
dhensby: Also, you've imported `Exception` (with a `use) so you don't need to explicitly denote it's in the global namespace with the leading `\`

core/Extensible.php
```diff
@@ -0,0 +1,465 @@
...
+			foreach(array_reverse($this->beforeExtendCallbacks[$method]) as $callback) {
+				$value = call_user_func_array($callback, array(&$a1, &$a2, &$a3, &$a4, &$a5, &$a6, &$a7));
+				if($value !== null) $values[] = $value;
+			}
+			$this->beforeExtendCallbacks[$method] = array();
+		}
+
+		if($this->extension_instances) foreach($this->extension_instances as $instance) {
+			if(method_exists($instance, $method)) {
+				$instance->setOwner($this);
```
dhensby: Doesn't `CustomMethods` take care of this?

core/Extensible.php
tractorcow: Config updates can't live-update extensions; It requires an event to re-define methods, etc, which tests rely on.

core/CustomMethods.php
tractorcow: ok

dhensby:
What a legend +1 from me

tractorcow:
I'll clean up the formatting + other easy things you spotted @dhensby

core/CustomMethods.php
tractorcow: How about BadMethodCallException

core/CustomMethods.php
tractorcow: It'll get fixed later when I fix virtualpage. See https://github.com/silverstripe/silverstripe-cms/issues/1386

core/Extensible.php
tractorcow: This is all old code; I'm not the one who made it yucky. ;P

core/Extensible.php
tractorcow: Extension doesn't have CustomMethods trait, only Object subclasses.

tractorcow:
@dhensby rebased and your fixes implemented.

CMS fix at https://github.com/silverstripe/silverstripe-cms/pull/1395

core/Configurable.php
tractorcow: Fixed.

tractorcow:
> I think you need more inline comments. I'm finding parts of it hard to really know what's going on. For example: I can't see any documented spec for the array keys you want in CustomMethods

I wish I knew what they all were myself. Maybe we can simplify the code at a later date. For now this is just a refactor.

dhensby:
nice work @tractorcow

PR #4996, opened 2016-01-28
4991 manymany extrafields

mrmorphic:
Fix for 4991.


tractorcow:
It looks like the branch you branched from isn't the same one that you're pulling to. Do you need to re-create this PR?

mrmorphic:
Hmm Ok. I branched from 3.2, and thought that's what I was generating the PR to. I'll close this and try it again.

PR #4997, opened 2016-01-28
BUG edge case on many many extra fields (fixes 4991)

mrmorphic:
Fixes an edge case where extraFields are not returned if
one side of a many many is added via extension (although this
may not be the only failure case). Fixes a
downstream issue with dms breaking the CMS on framework 3.2.

The bug is where a many many relationship exists on a class,
and a sub-class attempts to get the extra fields of the
relationship. The change fixes the test for exact matching of
the relationship class to the instance class, to checking if
the instance is the class or a subclass of the relationship.

The unit tests check the dms failure case, which is a more
complex failure case.

### DIFF
--- a/javascript/HtmlEditorField.js
+++ b/javascript/HtmlEditorField.js
@@ -888,10 +888,11 @@ ss.editorWrappers['default'] = ss.editorWrappers.tinyMCE;
 					action = "update";
 				}
 
-				if(href.match(/^mailto:(.*)$/)) {
+				if(href.match(/^mailto:([^?]*)(\?subject=(.*))?$/)) {
 					return {
 						LinkType: 'email',
 						email: RegExp.$1,
+						Subject: decodeURIComponent(RegExp.$3),
 						Description: title
 					};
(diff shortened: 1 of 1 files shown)


mrmorphic:
I'm unsure how to progress the test failures. I have run unit tests (all, framework and specific tests) on php 5.6, and I don't get the error about Page not existing. Can someone provide some insight about what I need to do to make it work?

kinglozzer:
@mrmorphic When the framework tests are run on Travis the CMS module isn’t installed, so `Page` doesn’t exist. I’d just replace it with a couple of parent/child `DataObject` classes instead

mrmorphic:
Ah, thanks for that. It's a shame in this case, because the actual failure case is between SiteTree and Page so it would be the best test, but I can replicate the structure.

dhensby:
great work on this :)

tractorcow:
I've reviewed your fix and confirm that it's a good fix. Please adjust the tests to not use SiteTree and I can merge.

tractorcow:
@mrmorphic are you going to open a new PR?

mrmorphic:
yep, sorry forgot to add the ref: https://github.com/silverstripe/silverstripe-framework/pull/5010

PR #5000, opened 2016-01-29
SapphireTest fix for issue that caused assertEmailSent to always fail

guttmann:
Ran into this issue when upgrading a couple of sites from 3.1 to 3.2.
# Reproduce

Install SilverStripe with composer `composer create-project silverstripe/installer . 3.2.1`

Add a method to Page.php that can be tested for sending email

``` php
class Page extends SiteTree {

    private static $db = array();
    private static $has_one = array();

    public static function method_to_test() {
        $email = new Email('test@test.com', 'test@test.com', 'test', 'test');
        $email->send();
    }

}
```

Write a test for that method

``` php
class PageTest extends SapphireTest {

    public function testMethodToTest() {
        Page::method_to_test();

        $this->assertEmailSent('test@test.com');
    }

}
```

Running this test will fail with an error similar to:

```
F
PageTest::testMethodToTest
Failed asserting that an email was sent to 'test@test.com'.
```
# Fix

I'm not smart enough to explain why it was broken but the fix was to use a mailer method (rather than a field) similar to what the `Email` class does.

In the process of making this change I found that the `$originalMailer` and `$mailer` fields were no longer required so have removed any references to these I could find.

### DIFF
--- a/dev/SapphireTest.php
+++ b/dev/SapphireTest.php
@@ -55,7 +55,9 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 	protected $originalNestedURLsState;
 	protected $originalMemoryLimit;
 
-	protected $mailer;
+	public static function mailer() {
+		return Injector::inst()->get('Mailer');
+	}
 
 	/**
 	 * Pointer to the manifest that isn't a test manifest
@@ -223,9 +225,7 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 		$prefix = defined('SS_DATABASE_PREFIX') ? SS_DATABASE_PREFIX : 'ss_';
 
 		// Set up email
-		$this->originalMailer = Email::mailer();
-		$this->mailer = new TestMailer();
-		Injector::inst()->registerService($this->mailer, 'Mailer');
+		Injector::inst()->registerService(new TestMailer(), 'Mailer');
 		Config::inst()->remove('Email', 'send_all_emails_to');
 
 		// Todo: this could be a special test model
@@ -496,10 +496,6 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 		// Preserve memory settings
 		ini_set('memory_limit', ($this->originalMemoryLimit) ? $this->originalMemoryLimit : -1);
 
-		// Restore email configuration
-		$this->originalMailer = null;
-		$this->mailer = null;
-
 		// Restore password validation
 		if($this->originalMemberPasswordValidator) {
 			Member::set_password_validator($this->originalMemberPasswordValidator);
@@ -558,7 +554,7 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 	 * Clear the log of emails sent
 	 */
 	public function clearEmails() {
-		return $this->mailer->clearEmails();
+		return static::mailer()->clearEmails();
 	}
 
 	/**
@@ -572,7 +568,7 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 	 *               'customHeaders', 'htmlContent', 'inlineImages'
 	 */
 	public function findEmail($to, $from = null, $subject = null, $content = null) {
-		return $this->mailer->findEmail($to, $from, $subject, $content);
+		return static::mailer()->findEmail($to, $from, $subject, $content);
 	}
 
 	/**


tractorcow:
5000 get!

dev/SapphireTest.php
```diff
@@ -43,11 +43,6 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 	 */
 	protected $usesDatabase = null;
 
-	/**
-	 * @deprecated since version 4.0
-	 */
-	protected $originalMailer;
```
tractorcow: Unfortunately we cannot remove $originalMailer until 4.0, as semver disallows removal of API in patch releases.

guttmann: Have put this back in - do you want me to reinstate the references to it elsewhere? (they literally do nothing now)

tractorcow:
> I'm not smart enough to explain why it was broken but the fix was to use a mailer method (rather than a field) similar to what the Email class does.

I'm sure you are very smart, you are at least as smart as I, as I can't see what was wrong either. Very peculiar.

Is it worth adding your unit test in to show that it's fixed? The one that was breaking?

We also probably can't remove $mailer either. It's fine if you ignore it, though, but you probably need to keep both assignments in the setUp() method, as existing unit tests could rely on both or one of those being assigned.

dev/SapphireTest.php
```diff
@@ -496,10 +496,6 @@ public function tearDown() {
 		// Preserve memory settings
 		ini_set('memory_limit', ($this->originalMemoryLimit) ? $this->originalMemoryLimit : -1);
 
-		// Restore email configuration
```
tractorcow: Do you think this is the bit that was breaking your tests? Having a null mailer / originalMailer?

dev/SapphireTest.php
tractorcow: It's ok if your fixed code doesn't rely on it, as long as third party tests (subclasses of SapphireTest) which once relied on these don't break.

guttmann:
Cool, I can add that stuff back in and I'll come up with a test for this ... SapphireTestTest? Or is that too meta?

dev/SapphireTest.php
guttmann: Nope - when I was debugging this $mailer was set to an instance of TestMailer it was just somehow a different instance to what was in Injector. I had assumed they would reference the same object but they weren't and I couldn't figure out why.

tractorcow:
There's an existing EmailTest so that would be the place :P

guttmann:
Sweet will sort that out over the weekend. Thanks for the help.

guttmann:
OK so I did some more digging into this after getting odd results when testing at home... it appears as though this is only an issue with the first test in a class and it's because the setUp method isn't being run before the first test...

Will keep investigating...

tractorcow:
>  it's because the setUp method isn't being run before the first test...

Yeah that's probably what I suspected too. Are you running tests with sake or with phpunit cli?

guttmann:
All my tests so far have been with sake - I will see what happens with the phpunit cli

guttmann:
It has been a while but I've just had a chance to test this again.

It's still happening in SilverStripe 3.3.1 and occurs when using sake or phpunit to run tests.

I'm at a loss for where to look to fix this properly (don't think my original fix above is actually fixing the root cause - setUp not being run for the first test)

patricknelson:
I've narrowed this down to the fact that there are apparently two different instances of `TestMailer` and it changes over when `::create_temp_db()` is called. Then, magically just in time when you want to perform a `->findEmail()` it will revert to the original instance that was initially setup slightly higher up in the `->setUp()` method. Why this happens, I have no clue.

You can debug this yourself by repeatedly dumping the output of `spl_object_hash(Injector::inst()->get('Mailer'))` and watching as it vacillates between two different hashes. I'm not sure if I can spend much more time on this but that's as far as I got :joy:

patricknelson:
I too became interested in this because of an `Injector` related issue I was having here https://github.com/silverstripe/silverstripe-framework/issues/5286 and that had some overlap with another unit test I was writing which I've discussed with @kinglozzer here https://github.com/silverstripe/silverstripe-framework/commit/e14f743bf031bda2022cd12e37d26f420ce7c403#commitcomment-17034221

patricknelson:
I'm thinking maybe this has to do with the fact that both `->setUp()` and `->setUpOnce()` get called and  both contain the `::nest()` calls to `Config` and, more importantly, `Injector`.  I'd propose maybe moving these `::nest()` statements to one or the other. Possibly just to `->setUp()`?

patricknelson:
Found it! :+1: Here's what's happening:

1. `SapphireTest->setUp()` effectively `calls Injector::inst()->registerService(new TestMailer(), 'Mailer')`
2. Calls `self::create_temp_db()` https://github.com/silverstripe/silverstripe-framework/blob/3/dev/SapphireTest.php#L240
3. Calls `$st->resetDBSchema()` https://github.com/silverstripe/silverstripe-framework/blob/3/dev/SapphireTest.php#L878
4. Undoes our service registration above via `Injector::inst()->unregisterAllObjects()` https://github.com/silverstripe/silverstripe-framework/blob/3/dev/SapphireTest.php#L910

So maybe the "real" solution here is to address this stomping on each others global state, or (due to global state being needed and cross contamination of concerns here, maybe) just change the order of how this is executing. Another method may be to, instead of completely wiping everything out in the `Injector` here on step 4 above to instead just be more specific, instead of wiping everything out (but I understand that may not be possible).

So I think maybe the cleanest route may just be to setup the email service registration after the rebuild of the database schema so that the service can persist like it should

kinglozzer:
@patricknelson That does indeed look like a bug. Thankfully calling `Injector::inst()->get('Mailer')` in tests will still return a test mailer (otherwise we’d have seen a lot of angry bug reports by now I’m sure!), because the [class name is cached](https://github.com/silverstripe/silverstripe-framework/blob/3ee8f505b75af39817d394bcf48623e092f43ed2/control/injector/Injector.php#L754) and that isn’t wiped by `unregisterAllObjects()`. Do you want to create a new issue for that?

tractorcow:
Moving the mailer setup to after the DB initialisation would be fine.

tractorcow:
> because the class name is cached and that isn’t wiped by unregisterAllObjects()

It's wiped by Injector::unnest() though in tearDown. :)

sminnee:
>  Thankfully calling Injector::inst()->get('Mailer') in tests will still return a test mailer (otherwise we’d have seen a lot of angry bug reports by now I’m sure!), because the class name is cached and that isn’t wiped by unregisterAllObjects().

Someones 2 wrongs _do_ make a right!

patricknelson:
Re: @kinglozzer 
> tests will still return a test mailer (otherwise we’d have seen a lot of angry bug reports by now I’m sure!), because the class name is cached and that isn’t wiped by unregisterAllObjects(). Do you want to create a new issue for that?

Not really since I'd need to hunt down why it's happening to determine for certain if it's a legit issue and not some emergent complexity of what's happening here with the nesting. In fact, I think `->setUp()` is called multiple times, but the temporary database is only performed once since it maintains global state (see my comment above about that). I think we all know global state is evil and unit tests help reveal this more than ever, don't they? :sunglasses:

patricknelson:
So actually there isn't some additional issue with `->unregisterAllObjects()` per se. I'm curious: Is it necessary to perform these nesting calls in both `->setUp()` and `->setUpOnce()`? I'm not sure it really matters all that much, but figured I would ask.

```php
Config::nest();
Injector::nest();
```

tractorcow:
Try not invoking it in the DB setup; It doesn't really feel necessary, as though it was trying to do a cache invalidation.

dhensby:
Status update? I think this may have been fixed now anyway...

tractorcow:
> Is it necessary to perform these nesting calls in both ->setUp() and ->setUpOnce()?

Just in the setUp(), since setUpOnce() would be redundant I guess?

The way framework unrolls these is illogical anyway, so I wouldn't rely on setUpOnce to be honest.

tractorcow:
> So actually there isn't some additional issue with ->unregisterAllObjects() per se.

I suggest trying `DataObject::reset()` in place of this call. That way your test mailer persists properly.
If It still fails, move the `->unregisterAllObjects()` to after nesting, but before your mailer is injected.

dhensby:
So, I believe this was fixed in master with https://github.com/silverstripe/silverstripe-framework/pull/5271 - can we take a similar approach?

tractorcow:
Yeah that might work... and it'll sure make merging up easier.

Issue #4932, opened 2016-01-14
RFC Neo-Versioning / Change Sets API


tractorcow:
# RFC Neo-Versioning / Change Sets API

https://groups.google.com/forum/#!topic/silverstripe-dev/p2sHTRqNmsc

# 1. Problem

DataObject versioning suffers from a set of problems, including but not limited to the below:

* Not every object is versioned, but that is not always easily distinguished via the CMS 

* Not every action taken by a content editor will (or will not) obviously result in an immediate change to the live site.

* It is often difficult to preview changes before publishing which affect on multiple items, especially those which rely on changes to unversioned objects.

* Once changes are made within a CMS, the process of making these changes live can be a tedious and error-prone process.

* Versioning is unfairly prioritised for SiteTree objects, but in more advanced sites, it is equally as important to version and control non-sitetree objects. E.g. Files or content sections.

* Many_Many relationships can not be versioned.

# 2. Proposal

In order to address these issues a new neo-versioning mechanism based on change-sets is proposed. Rather than focusing on operations made to individual objects, the major concept of this feature is that the content editor is always working with groups of changes pending publishing. Tracking changes to objects which rely on changes to other objects, such as those via dataobject relationships will ensure that all necessary objects are grouped together in the same group. To support this feature, DataObjects will now be versioned by default (with some exceptions), as will many_many relationships between objects.

On top of this a new user interface will be developed to assist content editors with control over changes pending publication. The requirements of this interface would be to:

* Address visibility of versioned objects

* Provide flexibility and control over the publishing process

* Communicate requirements to the user necessary to maintain backend consistency

* Support "quick publish" of any currently viewed object

* Preview mechanism

The focus of this RFC is solely on the low level API. UX design and development will be covered separately, and thus not addressed in great detail.

# 3. Concepts

The following new concepts are introduced in this RFC:

* **Change Set**: A set of links to objects with unpublished changes, grouped together.

* **Owned objects**: If an object owns another object, that means that the owner object requires those owned object in order for it to be properly rendered. E.g. a page owns content blocks, a content block owns images.

    * Ownership can be shared, and is not exclusive.

    * Ownership is recursive, as long as the ownership chain is unbroken. E.g. if a page owns a content block, and a content block owns an image, then the page also owns that image.

    * All shortcodes in HTML areas are automatically considered owned by the object containing that field.

    * There is no intrinsic ownership direction for has_ones or many_many relationships.

[...]

PR #5001, opened 2016-01-29
API File is now Versioned

tractorcow:
This is the last major refactor necessary to implement [RFC-1 Asset Versioning](https://github.com/silverstripe/silverstripe-framework/issues/3792). In this update File finally is given the `Versioned` extension.

AssetAdmin has been upgraded with versioned file management.

The most important parts of this pull request are:
- AssetControlExtension extension for performing visibility control of underling DBFile assets on the dataobject.
- Update of Versioned to make it useful for non-SiteTree objects
- VersionedGridFieldItemRequest Which supports versioned objects in the GridField. Credit to @icecaster and @unclecheese for inspiration and ideas. :)

Not included in this pull request, and still remaining:
- New UX for managing assets
- Shortcodes for images in HTML area
- Secure file streaming
- Bulk versioned change management

Rebased against dependant PR https://github.com/silverstripe/silverstripe-framework/pull/4992.

CMS pull request https://github.com/silverstripe/silverstripe-cms/pull/1381.

### DIFF
--- a/filesystem/AssetControlExtension.php
+++ b/filesystem/AssetControlExtension.php
@@ -4,110 +4,294 @@ namespace SilverStripe\Filesystem;
 
 use DataObject;
 use Injector;
+use Member;
+use Versioned;
 use SilverStripe\Filesystem\Storage\AssetStore;
 
 /**
- * Provides an interaction mechanism between objects and linked asset references.
+ * This class provides the necessary business logic to ensure that any assets attached
+ * to a record are safely deleted, published, or protected during certain operations.
+ *
+ * This class will respect the canView() of each object, and will use it to determine
+ * whether or not public users can access attached assets. Public and live records
+ * will have their assets promoted to the public store.
+ *
+ * Assets which exist only on non-live stages will be protected.
+ *
+ * Assets which are no longer referenced will be flushed via explicit delete calls
+ * to the underlying filesystem.
+ *
+ * @property DataObject|Versioned $owner A {@see DataObject}, potentially decorated with {@see Versioned} extension.
  */
-class AssetControlExtension extends \DataExtension {
+class AssetControlExtension extends \DataExtension
+{
 
-    /**
-     * When archiving versioned dataobjects, should assets be archived with them?
-     * If false, assets will be deleted when the object is removed from draft.
-     * If true, assets will be instead moved to the protected store.
-     *
-     * @var bool
-     */
-    private static $archive_assets = false;
+	/**
+	 * When archiving versioned dataobjects, should assets be archived with them?
+	 * If false, assets will be deleted when the dataobject is archived.
+	 * If true, assets will be instead moved to the protected store, and can be
+	 * restored when the dataobject is restored from archive.
+	 *
+	 * Note that this does not affect the archiving of the actual database record in any way,
+	 * only the physical file.
+	 *
+	 * Unversioned dataobjects will ignore this option and always delete attached
+	 * assets on deletion.
+	 *
+	 * @config
+	 * @var bool
+	 */
+	private static $keep_archived_assets = false;
 
-    public function onAfterDelete() {
-        $assets = $this->findAssets($this->owner);
+	/**
+	 * Ensure that deletes records remove their underlying file assets, without affecting
+	 * other staged records.
+	 */
+	public function onAfterDelete()
+	{
+		// Prepare blank manipulation
+		$manipulations = new AssetManipulationList();
 
-        // When deleting from live, just secure assets
-        // Note that DataObject::delete() ignores sourceQueryParams
-        if($this->isVersioned() && \Versioned::current_stage() === \Versioned::get_live_stage()) {
-            $this->protectAll($assets);
-            return;
-        }
+		// Add all assets for deletion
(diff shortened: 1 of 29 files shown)


filesystem/AssetControlExtension.php
```diff
@@ -4,110 +4,289 @@
...
+		}
+
+		// Check if canView permits anonymous viewers
+		return $record->canView(Member::create())
+			? AssetManipulationList::STATE_PUBLIC
+			: AssetManipulationList::STATE_PROTECTED;
+	}
+
+	/**
+	 * Given a
```
stevie-mayhew: Missing the rest of the docblock?

tractorcow: Whoops I did a tractorcow

tractorcow:
I find it amazing the tests all passed the first time.

I'll let this sit over the weekend and respond to all comments after that. :)

filesystem/AssetControlExtension.php
stevie-mayhew: Its catching :stuck_out_tongue:

model/versioning/VersionedGridFieldDetailForm.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+/**
+ * Extends {@see GridFieldDetailForm}
+ */
+class VersionedGridFieldDetailForm extends Extension {
```
chillu: What's the rationale for using an extension here if we're always applying it in core anyway?

chillu:
A `dev/build` on this branch leaves the DB in an inconsistent state: Files are accessible by URL, but appear unpublished in the database. I'd expect existing records to be copied from `File` to `File_Live` tables, emulating 3.x behaviour. AFAICT, we can't infer visibility on existing files by record relationships - we have to assume all files should stay public. Should this be another feature of `MigrateFileTask`?

Do you think `AssetControlExtension` will still be required once we implement the Neo Versioning as described in https://github.com/silverstripe/silverstripe-framework/issues/4932? It seems to tackle a subset of this problem space, with a few specialisations like ignoring variants.

model/versioning/VersionedGridFieldDetailForm.php
tractorcow: Because if it's core, it cannot be substituted or removed.

Versioned should be left as loosely coupled as possible, so that it can be removed if need be.

tractorcow:
> Do you think AssetControlExtension will still be required once we implement the Neo Versioning as described in #4932? It seems to tackle a subset of this problem space, with a few specialisations like ignoring variants.

Absolutely; All this does is synchronise the filesystem with the versioned state of specific records. We don't want to have to make changesets have to micromanage assets. Instead, it'll be left in change of performing dataobject-level operations, and leave AssetControlExtension to translate those actions into filesystem operations.

tractorcow:
> A dev/build on this branch leaves the DB in an inconsistent state: Files are accessible by URL, but appear unpublished in the database. I'd expect existing records to be copied from File to File_Live tables, emulating 3.x behaviour. AFAICT, we can't infer visibility on existing files by record relationships - we have to assume all files should stay public. Should this be another feature of MigrateFileTask?

This seems like a harder problem, but I wonder if it could be done. Maybe instruct the creation of _Live tables to duplicate the stage tables?

tractorcow:
It looks like file and link tracking in https://github.com/silverstripe/silverstripe-cms/pull/1381 is very broken. I'd hold off merging anything until I can get it fixed.

tractorcow:
https://github.com/silverstripe/silverstripe-cms/pull/1381 has been fixed; Back links have been disabled on live for the time being.

model/versioning/VersionedGridFieldDetailForm.php
chillu: Fair enough, it allows us to decouple Versioned from form fields later

tractorcow: Yes, I still believe versioning should be a separate module... but I guess it remains to be seen if that's possible yet.

chillu:
Happy with this change. Can we get at least one other +1 from @silverstripe/core-team please?

kinglozzer:
@tractorcow The commit from #4992 is breaking other things with those pesky `DateTime`/`SS_DateTime` changes :(. When I check out this PR I get errors coming from `GridField` casting - looks like `CMSMain` (~ line 800) needs to be updated in the CMS PR to replace `Datetime->Ago` with `SS_DateTime->Ago`. We need #4551 merged!

That aside, I’ve looked through and I’m happy. UX needs some work but I appreciate that’s out of the remit of this PR :)

tractorcow:
Ah, thanks for spotting those. I'll have a look at fixing and re-uploading.

tractorcow:
I keep running out of time to look at this... I'll try to get it over the line tomorrow guys. :)

tractorcow:
https://github.com/silverstripe/silverstripe-framework/pull/4992 has been updated, and rebased this on top of it. :dancer: 

Does this meet all concerns? :P

tractorcow:
Also, this PR has been checked that it doesn't rely on #4551 for now.

tractorcow:
Since #4992 is merged I've rebased this one.

tractorcow:
Rebased again....

PR #5005, opened 2016-02-02
Set the stage reading mode in LeftAndMain

dhensby:
`LeftAndMain` is used as the foundation of the CMS.

All edits in the CMS should really be against the stage version of a `DataObject` otherwise there becomes inconsistencies (see https://github.com/silverstripe/silverstripe-cms/issues/1094)

### DIFF
--- a/admin/code/LeftAndMain.php
+++ b/admin/code/LeftAndMain.php
@@ -444,6 +444,9 @@ class LeftAndMain extends Controller implements PermissionProvider {
 		// The user's theme shouldn't affect the CMS, if, for example, they have
 		// replaced TableListField.ss or Form.ss.
 		Config::inst()->update('SSViewer', 'theme_enabled', false);
+
+		//set the reading mode for the admin to stage
+		Versioned::set_reading_mode('Stage');
 	}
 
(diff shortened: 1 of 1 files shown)


tractorcow:
I everything agree with this.

admin/code/LeftAndMain.php
```diff
@@ -444,6 +444,9 @@ public function init() {
 		// The user's theme shouldn't affect the CMS, if, for example, they have
 		// replaced TableListField.ss or Form.ss.
 		Config::inst()->update('SSViewer', 'theme_enabled', false);
+
+		//set the reading mode for the admin to stage
+		Versioned::set_reading_mode('Stage');
```
tractorcow: `Versioned::reading_stage('Stage');` not `Versioned::set_reading_mode('Stage');`

You could do `Versioned::reading_stage('Stage.Stage');` too

dhensby: Derp

tractorcow: It's an annoying API. :D I'll fix in master soon, when we do changesets.

PR #5008, opened 2016-02-03
Set the stage reading mode in LeftAndMain

dhensby:
Attempt number 2 see #5005


tractorcow:
Thanks @dhensby :D

dhensby:
@tractorcow green ticks

tractorcow:
That's the only thing we look at!

PR #5010, opened 2016-02-03
BUG edge case on many many extra fields (fixes 4991)

mrmorphic:
Fixes an edge case where extraFields are not returned if
one side of a many many is added via extension (although this
may not be the only failure case). Fixes a
downstream issue with dms breaking the CMS on framework 3.2.

The bug is where a many many relationship exists on a class,
and a sub-class attempts to get the extra fields of the
relationship. The change fixes the test for exact matching of
the relationship class to the instance class, to checking if
the instance is the class or a subclass of the relationship.

The unit tests check the dms failure case, which is a more
complex failure case.

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -2077,7 +2077,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 				$candidateManyManys = (array)Config::inst()->get($candidate, 'many_many', Config::UNINHERITED);
 
 				foreach($candidateManyManys as $relation => $relatedClass) {
-					if($relatedClass === $this->class) {
+					if (is_a($this, $relatedClass)) {
 						$relationName = $relation;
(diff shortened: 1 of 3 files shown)


mrmorphic:
This is a re-issue with tests isolated from cms.

dhensby:
THANKS!

model/DataObject.php
```diff
@@ -2077,7 +2077,7 @@ public function manyManyExtraFieldsForComponent($component) {
 				$candidateManyManys = (array)Config::inst()->get($candidate, 'many_many', Config::UNINHERITED);
 
 				foreach($candidateManyManys as $relation => $relatedClass) {
-					if($relatedClass === $this->class) {
+					if (is_a($this, $relatedClass)) {
```
tractorcow: good job!

PR #5024, opened 2016-02-10
Ensure filterByCallback returns object of called class

froog:
This change ensures that filterByCallback returns a filtered list of the same called class, if ArrayList is extended.

The current filterByCallback always instantiates an ArrayList with the filtered results, which is probably not what you want if you extend ArrayList.

eg; currently this occurs
```
class MyArrayList extends ArrayList {
   public function sayHello() {
      return 'hi'; 
   }
}

$list = MyArrayList::create();
//populate $list here
$filtered = $list::filterByCallback($someCallback);

$filtered->sayHello();  
// -> The method 'sayHello' does not exist on 'ArrayList' 
```

### DIFF
--- a/model/ArrayList.php
+++ b/model/ArrayList.php
@@ -548,7 +548,7 @@ class ArrayList extends ViewableData implements SS_List, SS_Filterable, SS_Sorta
 			));
 		}
 
-		$output = self::create();
+		$output = static::create();
 
 		foreach($this as $item) {
 			if(call_user_func($callback, $item, $this)) $output->push($item);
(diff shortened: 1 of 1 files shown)


dhensby:
Don't we want to use `static::create()`?

froog:
I was wondering about this exact question actually - I agree with you, but it points out a logical quirk in Object::create(), eg;

$x = self::create();
$y = new self;

$x will be of class MyArrayList while $y of class ArrayList

Where as static will both be MyArrayList

It just seems a bit odd that the instantiation method that create() replaces behaves differently (although maybe that's the point!)

I'll change it to static.

dhensby:
Thanks

PR #5027, opened 2016-02-10
Integrate Bootstrap4 Alpha.2 and replace Compass with gulp-sass

scott1702:
Integration of Bootstrap4 alpha-2 into framework/admin. Bootstrap is pulled in with npm, then we have custom scss files in framework/admin/scss so we can fill bootstrap's _variables.scss with our own values, gulp then compiles these out to framework/admin/thirdparty.

While doing this I figured it would be better if everything in the scss folder was handled through gulp. I'm happy to split this out in a separate pull request if anyone feels it is necessary.

Motivation: https://groups.google.com/forum/#!topic/silverstripe-dev/zY-yzVMA0ds

To compile css simply `npm run compile`. This will compile the sprites from `framework/admin/images/sprites/src` into `framework/admin/images/sprites/dist` and the scss from `framework/scss`, `framework/admin/scss` and `framework/dev/install/scss` into minified css.
Running `npm run compile --development` will compile the expanded css and 'watch' each scss directory for changes.

ping @ianherbert


admin/scss/_actionTabs.scss
```diff
@@ -170,7 +170,7 @@ $border: 1px solid darken(#D9D9D9, 15%);
 					>li{
 						width: auto;
 						a.tab-nav-link{
-							@include box-sizing(border-box);
+							box-sizin: border-box;
 							@include duration(0.5s);
```
jonom: Could this be handled by autoprefixer too with just `transition-duration`? It looks to me like the `transition` and `duration` mixins in admin/scss/_mixins.scss could be removed now - `transition()` doesn't seem to be used at all and `duration()` is only used here a couple of times, without the second argument. Not a must-have, just seems like a good a time as any to include :D

jonom:
So this adds bootstrap files but they're not actually in use yet is that right? The meat of the change is moving away from Compass which seems like a good idea. I've had a quick click around locally and can't spot any new layout/visual issues so looks okay to me.

scott1702:
Correct, nothing actually uses the bootstrap files yet. I also didn't change the _variables.scss in any way from bootstrap's default values (brand-color, bg-color etc) but I would assume those would get changed when it comes to implementation.

admin/scss/_actionTabs.scss
scott1702: Yes! [autoprefixer is current used with bootstrap](https://github.com/silverstripe/silverstripe-framework/pull/5027/files#diff-b9e12334e9eafd8341a6107dd98510c9R236) in the gulpfile and I see no reason why we can't use it to handle our own css as well. I'll get rid of the transition and duration mixins now which [don't actually need to be prefixed anymore](http://caniuse.com/#feat=css-transitions).

scott1702:
@jonom updated with transition and duration mixins removed.

jonom:
Nice, thanks :+1:

jonom:
This looks good to me but should probably be reviewed/merged by another @silverstripe/core-team member that is experienced with gulp/npm as I haven't caught up on those technologies yet.

admin/scss/bootstrap/_variables.scss
```diff
@@ -0,0 +1,661 @@
...
+// General variable structure
+//
+// Variable format should follow the `$component-modifier-state-property` order.
+
+// Colors
+//
+// Grayscale and brand colors for use across Bootstrap.
+
+$gray-dark:                 #373a3c !default;
+$gray:                      #55595c !default;
```
dhensby: Do we really want to name a variable `$gray`.

Also, can't we use lighten / darken to adjust the colours?

I'm not a fan of naming variables after colours because if we decide that grey should now be very dark blue, then we either have a variable called `$gray` that is now returning blue colours OR we have to rename the variable (in which case we might as well do a search and replace on the colour hex anyway.

I've seen this kind of variable called `$scaffolding` (which makes sense if it's used for backgrounds, borders, etc.) or maybe `$base` or `$neutral` to better describe it

dhensby:
> gulp then compiles these out to framework/admin/thirdparty

Why do we compile to 3rd party instead of just to CSS files?

---

I think we also need better dev documentation on how to add sprites and other such customisations.

---

@jonom I think this is quite a bit change that needs more than a few hours for feedback - this should stay open 'til we have a decent amount of feedback

dhensby:
Also, have we considered bringing [bourbon](http://bourbon.io/) into the mix as a mixin lib?

jonom:
> I think this is quite a bit change that needs more than a few hours for feedback - this should stay open 'til we have a decent amount of feedback

I agree but I think there's less going on in this PR than it appears on the surface - basically just dropping in a vanilla version of bootstrap 4 alpha to be customised later, it's not integrated at all. Agree though that any files we plan on customising shouldn't go in to thirdparty.

@scott1702 I think it would help to split this up in to two PRs actually - it's a bit hard to follow along, particularly because some of the bootstrap changes are in the first commit and some are in the second so the commits don't seem to quite be cleanly separated (i.e. admin/scss/bootstrap/_variables.scss file is introduced in the gulp-compass commit).

Most of the changes are from dropping compass so if we subtract that noise it should be easier to discuss the bootstrap introduction. Maybe start with the compass-gulp PR and introduce bootstrap after?

scott1702:
> Why do we compile to 3rd party instead of just to CSS files?

I compiled it out to thirdparty because all it is is bootstrap with custom variables, although since we are technically altering it by having custom variables we probably should just compile it to the CSS folder

> I think we also need better dev documentation on how to add sprites and other such customisations.

Agreed, I'll write some more

> Also, have we considered bringing bourbon into the mix as a mixin lib?

That has been considered, @clarkepaul did a large amount of research with a lot of different frameworks (including bourbon, summary here https://groups.google.com/d/msg/silverstripe-dev/zY-yzVMA0ds/GBzlt9AnAwAJ)

> I think it would help to split this up in to two PRs actually - it's a bit hard to follow along
Yeah I'll split this out today and we can continue the conversations on the respective pull requests once they are made

scott1702:
gulp-compass PR https://github.com/silverstripe/silverstripe-framework/pull/5034, will PR bootstrap after the outcome of https://github.com/silverstripe/silverstripe-framework/pull/5034 is decided

Issue #5028, opened 2016-02-11
YAML config errors if you have no configuration.


patricknelson:
Typically a YAML configuration is expected to have contents, however, due to the framework not first checking to ensure that a `$fragment` is an array, an error will be thrown if you happen to have an empty configuration file (see https://github.com/silverstripe/silverstripe-framework/blob/3/core/manifest/ConfigManifest.php#L655). This should be a quick fix. I think it should be done since sometimes you may need to setup boilerplate code (like in my case) on a per-environment basis separated out into different files but may not have any configuration needed for that particular environment. 

For example, this will cause an error since there's no real data after the second `---`.

```yaml
---
Name: mysite-dev
After:
  - 'framework/*'
  - 'cms/*'
Only:
  environment: dev
---
# Environment: 'dev'
```

tractorcow:
I agree, sounds fair!

The block will still be needed for before/after rules, even if the body is empty.

patricknelson:
@tractorcow  Since I've not gotten to the PR yet -- what branch do you want the PR to be against?

tractorcow:
3 branch is probably safest.

patricknelson:
Ok @tractorcow  see https://github.com/silverstripe/silverstripe-framework/pull/5032

tractorcow:
Thanks for that. ;D Good job.

dhensby:
Fixed by #5032 (i think?)

PR #5032, opened 2016-02-12
FIX for #5028: Ensure empty YML configs don't break when merging them in.

patricknelson:
This is simply making sure that the $fragment variable is traversable before foreach'ing over it. Should work as a solid fix against #5028.


tractorcow:
Are you able to please provide a test?

tractorcow:
It's fine. ;D

patricknelson:
Sorry, I'm actually in the middle of a 3k mile move at the moment 😖

tractorcow:
So you're 1.5ks there?

patricknelson:
Not sure what that means. Just forgot this existed, thanks for the ping.

PR #5036, opened 2016-02-15
DOCS: including join table data - an example

sunnysideup:
Hopefully this works, please adjust as required or create a new pull request with a better example.

### DIFF
--- a/docs/en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
+++ b/docs/en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
@@ -492,8 +492,18 @@ You can specify a join with the `innerJoin` and `leftJoin` methods.  Both of the
 	
 <div class="alert" markdown="1">
 Passing a *$join* statement to will filter results further by the JOINs performed against the foreign table. It will 
-**not** return the additionally joined data.
-</div>
+**not** return the additionally joined data. Below is an example on how to include fields from a join table.
+ </div>	
+ 		 
+To include data from joined tables, you can use the following code:
+
+	:::php
+	$members = Member::get()
+		->innerJoin("JoinTable", "\"Group_Members\".\"MemberID\" = \"Member\".\"ID\"");
+
+	$newQuery = members->dataQuery()->selectFromTable("JoinTable", array("A_Field_From_JoinTable"));
+	$members = $members->setDataQuery($newQuery);
+
 
 ### Default Values
 


docs/en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
```diff
@@ -492,8 +492,18 @@ You can specify a join with the `innerJoin` and `leftJoin` methods.  Both of the
...
-**not** return the additionally joined data.
-</div>
+**not** return the additionally joined data. Below is an example on how to include fields from a join table.
+ </div>	
+ 		 
+To include data from joined tables, you can use the following code:
+
+	:::php
+	$members = Member::get()
+		->innerJoin("JoinTable", "\"Group_Members\".\"MemberID\" = \"Member\".\"ID\"");
```
tractorcow: You could also do;


```php
$members = $members->alterDataQuery(function($query) {
	$query->selectFromTable("JoinTable", array("A_Field_From_JoinTable"));
});
```

To skip having to re-assign the dataquery afterwards. :)

sunnysideup: sure, I am a bit scared of adding functions like that, but that is just me ;-)  I was sure there was a better way, but the key thing is having some help here as I am sure many people would find this useful (it is not so easy to find).

tractorcow: Yeah that's what I'm here for... to help you out. :)

dhensby:
Satus of this?

tractorcow:
@sunnysideup it's probably safer to use alterDataQuery than have users manually re-assign a DataQuery to a DataList. It's already risky for user-code to be manipulating queries in this way at all. My intention is to remove as much risky API surface from the user docs as possible. :)

axyr:
And how would you do custom '$onPredicate' that is possible with SQLQuery>addInnerJoin ...

GridFieldComponents relies on DataLists, but a DataList is not always a simple 'SELECT * FROM Table'.

Always if the bussines logic of a client is a little more complex than 'average', I have to use SQLQuery and create ArrayLists to get the data, just because DataList is very limited. And that brings trouble with gridField manipulations.

I understand that you want a 'save' framework, but if a developer managed it to the point it needs custom queries, why not give him the power to do so in  a straight forward way?

tractorcow:
> And how would you do custom '$onPredicate' that is possible with SQLQuery>addInnerJoin ...

You can do all the same queries inside the alterDataQuery as you can inside, just without having to manually shuffle the dataquery into a new datalist. You would have to do more work though.

> I understand that you want a 'save' framework, but if a developer managed it to the point it needs custom queries, why not give him the power to do so in a straight forward way?

They have that way, but we also want to avoid error, so we encourage the use of error prone API. in the case of alterDataQuery, this enforces the immutability of the datalist, which in turns ensures the predictable behaviour of chained queries.

If any developer can solve a problem using a method, that's awesome; But our core docs need to demonstrate the best and most robust method to solve each problem. By using the correct API you get the correct behaviour, hence the rigour of this review.

PR #5037, opened 2016-02-15
API Update TinyMCE to 4.x

tractorcow:
Plus a bunch of other things:
- API Allow HtmlEditorField to be individually configured
- BUG Fix incorrect change detection
- BUG Fix missing i18n files

Due to tinymce being overly zealous with it's change detection, I've modified the change warning on the page to suppress duplicate messages popping up.

CMS pull request with updated behat tests at https://github.com/silverstripe/silverstripe-cms/pull/1394


tractorcow:
As a note, it seems that IE8 support for the CMS broke a few commits ago, so I'm unable to test support for it at this stage. IE9 and above are perfectly fine though.

JorisDebonnet:
Makes me happy to see this! 

Are you going for IE8 support? I'm just curious whether SilverStripe has made a pledge about this, concerning future versions. 
IE8 and IE9 (which are both not supported by Microsoft anymore) do not support some of TinyMCE's [image plugins](https://www.tinymce.com/docs/get-started/system-requirements/#browsercompatibility). Extra support is nice to have, of course, but perhaps it's a shame if useful javascript plugins need to be avoided because of them. Companies that are stuck with web administrators on WinXP could stay with SS3, no? Just my thoughts :)

tractorcow:
I really don't want to support ie8, but maybe @sminnee  will make a call on it and force us to. ;P

jonom:
Yikes, I'm with you on IE8 support @tractorcow.

tractorcow:
I've pushed up a fix to TextareaField test regression. It turns out the htmlentities in Value() was necessary after all. :)

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -38,6 +39,10 @@
  * `AssetControlExtension` is applied by default to all DataObjects, in order to support the management
    of linked assets and file protection.
  * `ProtectedFileController` class is used to serve up protected assets.
+ * `HtmlEditorConfig` is now an abstract class, with a default implementation `TinyMCEConfig` for the built in
+   TinyMCE editor.
+ * `HtmlEditorField::setEditorConfig` my now take an instance of a `HtmlEditorConfig` class, as well as a
```
chillu: Do you mean "can now take"?

admin/javascript/src/LeftAndMain.EditForm.js
```diff
@@ -121,37 +123,37 @@ $.entwine('ss', function($){
...
 			this.trigger('beforesubmitform');
-			if(!this.is('.changed')) {
+			if(!this.is('.changed') || this.is('.discardchanges')) {
 				return true;
 			}
 			var confirmed = confirm(i18n._t('LeftAndMain.CONFIRMUNSAVED'));
 			if(confirmed) {
 				// confirm discard changes
-				this.removeClass('changed');
+				this.addClass('discardchanges');
```
chillu: Can you please document why this class is necessary? I guess it's semantically more correct this way anyway, since form field content is still changed but the user has made the explicit decision to ignore it.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -23,6 +23,7 @@
  * `Upload::load` now stores assets directly without saving into a `File` dataobject.
  * Protected file storage is now a core Framework API. See [/developer_guides/files/file_security] for
    more information.
```
chillu: Link to the TinyMCE upgrading guide here: http://archive.tinymce.com/wiki.php/Tutorial:Migration_guide_from_3.x. There's some removed and added plugins which devs might've blocked. For example, devs might have disallowed image insertion by removing `advimage`, which has now been recast as `image`.

chillu:
I think we need to document which options/plugins have been removed from the UI, in order to give devs a chance to restore the 3.x functionality - and prevent author frustration. For example, some authors might heavily rely on inserting special characters.

Looking at a side-by-side view, I can identify the following missing: 

- strikethrough
- styles
- block quotes
- hr
- undo/redo/cut (built into browser now, with more people knowing the shortcuts)
- select all
- fullscreen

image

Of these, I think `styles` is probably the most important one, I've customised that dropdown on a few installations.

chillu:
It's a bit hard to review the PHP classes since they've changed completely. @tractorcow anything you'd like to point out there in terms of changes?

model/HTMLValue.php
```diff
@@ -55,6 +54,11 @@ public function getContent() {
 			return Convert::raw2att($attrs[$matches[0]]);
 		}, $res);
 
+		// Prevent &nbsp; being encoded as literal utf-8 characters
```
chillu: Did the TinyMCE behaviour change to cause this issue?

tractorcow: It was an issue we never bothered to fix before. It was affecting change detection, because tinymce was quite rightly encoding the UTF-8 equivalent back to `&nbsp;`, and causing change detection to trigger the "you have unsaved changes" warning even when nothing was touched by the user.

It was easier to make the core follow the standard than to hack change detection.

tractorcow: If we went the route of converting ALL utf-8 characters to the necessary entities, it would have had adverse affects on non-english locales. E.g. accented characters, asian, etc. so I've opted for the simplest solution.

docs/en/04_Changelogs/4.0.0.md
tractorcow: Ment to say "may now"

admin/javascript/src/LeftAndMain.EditForm.js
tractorcow: Sure;
The original behaviour was intended to remove the changed flag, but this relied on the class not yet being re-applied during navigation.

The issue is that, perhaps due to some onNavigation hook somewhere, tinymce is doing some kind of "cleanup" that yet again triggers change detection, and re-adds the changed class.

As a result, you ended up with the notification being shown TWICE.

If you say "no I don't want to save my changes" then it doesn't matter if there are changes made immediately after, you have already specified your intention to abandon these changes, so it shouldn't matter if "changed" class is added again.

The solution to this is to "fix" the users decision with a custom add-only class (which can only be removed by re-loading the form).

tractorcow:
> It's a bit hard to review the PHP classes since they've changed completely. @tractorcow anything you'd like to point out there in terms of changes?

The major behaviour is that all tinymce specific configuration has been moved into a new subclass, TinyMCEConfig. The actual configuration is now injected via a data-attribute, which means that the old global `var` is removed.

Loading of plugins has changed slightly, since tinymce breaks unless you use absolute URLS for plugin files. If you search the web, there has been a huge amount of issues with this in other projects. In order to bypass this completely (I got frustrated) all plugin urls are passed as absolute urls.

tractorcow:
> Of these, I think styles is probably the most important one, I've customised that dropdown on a few installations.

Unfortunately this isn't available as a button anymore. You need to enable the menu to use this feature, which is a bit annoying since that adds a whole new row to the UX interface.

admin/javascript/src/LeftAndMain.EditForm.js
chillu: Yeah I understand the technical reasons for it, was just asking about adding a comment line in the code :)

chillu:
> Unfortunately this isn't available as a button anymore. You need to enable the menu to use this feature, which is a bit annoying since that adds a whole new row to the UX interface.

That's fine, we should just point that out a bit clearer. Ideally TinyMCE would have those per-config upgrading guides so we can just point devs to it, but it doesn't look like they're going into too much detail - so we need to add a bit more detail on our side in the release notes. Don't think we need to re-create this as a button

admin/javascript/src/LeftAndMain.EditForm.js
tractorcow: Added

```php
// Ensures that once a form is confirmed, subsequent
// changes to the underlying form don't trigger
// additional chance confirmation requests
```

tractorcow:
I've added a note on "some features are not enabled by default, read the tinymce docs to figure out how to re-enable them" to the changelog.

tractorcow:
Merge please! @silverstripe/core-team

kinglozzer:
Rebase please! @tractorcow :D

tractorcow:
ok, rebased. ;D

chillu:
This broke the `ss_macron` plugin in core. Just looking to replacing that with a custom configuration for the built-in `charmap` plugin.

tractorcow:
@chillu your changes are fine... feel free to merge / squash.

camfindlay:
:tada:

bnolan:
Is the new tinymce the one on the left? It looks nice.

chillu:
@bnolan Yeah, the one on the left - nice and cleaned up. We've written a few quite custom plugins over the years around inserting images and links, so was a mission to get this upgraded.

Issue #5040, opened 2016-02-15
Reverting a page to a specific version is not working correctly : SS 3.2


priyashantha:
When we use 'Revert to this version' in History tab of a cms page, it creates a new version record for child Page_versions table but not in SiteTree_versions (Page extends SiteTree and when Page has db array). 

Then we can't do anymore reverts as we have orphaned record in Page_versions table. This happens only in SS 3.2 version but not in SS 3.1

tractorcow:
Have you tested this in 3.2.x-dev branch, or only 3.2.1 stable?

tractorcow:
Ok, it looks like this could be an issue.


```
$from->forceChange();
if($createNewVersion) {
	$latest = self::get_latest_version($baseClass, $this->owner->ID);
	$this->owner->Version = $latest->Version + 1;
} else {
	$from->migrateVersion($from->Version);
}
```

The `$this->owner->Version` should be `$from->Version`.

priyashantha:
I tested this with 3.2.1. Please see my reply in PR #5041.

priyashantha:
I think we have the issue in $manipulation array according to my scenario.

in augmentWite() function line number 663: if(empty($manipulation[$table]['fields']['Version']))
Here we have 2 tables (Page and SiteTree). Above condition is false only for SiteTree table as it contains Version field but not for Page table. I tested this against to SS3.1 installation and there we have Version field for both of Page and SiteTree tables. That's why there is no new version record in SS3.1 when rollbacking.

tractorcow:
Putting a duplicate Version field on each table will cause side effects... you are essentially de-normalising the DB schema and introducing the risk of breakage in case that the duplicated Version field gets de-synchronised.

`$version = isset($manipulation[$table]['fields']['Version'])`

Let's just change this so the version is obtained from the base table, not the same table being written.

We'll also need to adjust the logic in augmentWriteVersioned.

priyashantha:
Here I don't understand what should be the final solution. In 3.2 release, are we creating new version records in all `_versions` tables for the class ancestry? Currently with the check `empty($manipulation[$table]['fields']['Version'])` in `augmentWrite()`, we have allowed to create a new version records only in child class tables(i.e. Page) but not the `$baseDataClass` table(i.e. SiteTree) given that field `Version` is not there in child tables and the referenced version already in the `$baseDataClass_versions`table (eg: when roll back)

[...]

PR #5041, opened 2016-02-15
Reverting a page to a specific version is not working correctly : SS 3.2

priyashantha:
disabled creating a new version record for child Page_versions table but not in SiteTree_versions (Page extends SiteTree and when Page has db array) . Issue #5040

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -2733,6 +2733,11 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 		// Since database_fields omits 'ID'
 		if($field == "ID") return "Int";
 
+        // Add fields from Versioned extension
+        if($field == 'Version' && singleton($class)->hasExtension('Versioned')) {
+            return 'Int';
+        }
+
 		$fieldMap = self::database_fields($class, false);
 
 		// Remove string-based "constructor-arguments" from the DBField definition
--- a/model/Versioned.php
+++ b/model/Versioned.php
@@ -1283,7 +1283,7 @@ class Versioned extends DataExtension implements TemplateGlobalProvider {
 	 */
 	public function doRollbackTo($version) {
 		$this->owner->extend('onBeforeRollback', $version);
-		$this->publish($version, "Stage");
+		$this->publish($version, "Stage", true);
 
 		$this->owner->writeWithoutVersion();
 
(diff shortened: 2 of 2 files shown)


tractorcow:
I have no idea why it said I changed the title... I didn't =/

tractorcow:
Why does 'false' for the `$createNewVersion` parameter force creation of a new version in each table? IT seems to imply the opposite of the desired affect.

Maybe a unit test would help to prove this fix works? :D

tractorcow:
See my comment at https://github.com/silverstripe/silverstripe-framework/issues/5040#issuecomment-184403197 with what I suspect is the actual problem.

priyashantha:
As you said it makes sense when we revert to a older version, it creates a  new version. But the problem is the newly created version is only in for near child version table but not in parent. For instance : 

Page extends SiteTree {
  private static $db = array('test' => 'Varchar')
}

Now we have both Page and SiteTree table. When I revert page to a older version, the new version is only created in Page_versions table but not in SiteTree_versions table. Then we have an orphaned record in Page_versions table and we getting duplicate entry error when we re-try to revert to another version.

tractorcow:
> As you said it makes sense when we revert to a older version, it creates a new version. But the problem is the newly created version is only in for near child version table but not in parent. For instance :


I agree this needs to be fixed. I'm just not sure that this specific fix will resolve the issue in a semver-compatible way; I'd rather one that we could merge without breaking semver. Or for that matter, without changes in behaviour we could avoid. :D

priyashantha:
I came up with this solution because in previous SS version (3.1) there was no new version creation when we revert to a new version :) and I felt good to make $createNewVersion false in doRollbackTo() function as we'd not need a new version when we revert it back to an existing one. But that guess is not valid anymore because creating a new version record when reverting is a feature in the new version. 

I'm looking into fix for this and will let you know my findings.

priyashantha:
I think we have the issue in $manipulation array according to my scenario. 

in augmentWite() function line number 663: if(empty($manipulation[$table]['fields']['Version']))
Here we have 2 tables (Page and SiteTree). Above condition is false only for SiteTree table as it contains Version field but not for Page table. I tested this against to SS3.1 installation and there we have Version field for both of Page and SiteTree tables. That's why there is no new version record in SS3.1 when rollbacking.

tractorcow:
Maybe it'd be best to start with a unit test that reproduces the error?

tractorcow:
Anyone looking here; I wrote a PR that might fix it at https://github.com/silverstripe/silverstripe-framework/pull/5074

dhensby:
@tractorcow did it fix it?

tractorcow:
Yep. :D

PR #5042, opened 2016-02-15
Image deprecations and optimisation

jonom:
I noticed a couple of thumbnail methods were inconsistent with everything else here and inefficiently creating images in their own name space instead of using the common API. I think that has been cleaned up for 4.0 but may as well clean it up here too.

@tractorcow let me know if this is too late for 3.3 but I thought it would be good to tag those generate methods for deprecation sooner rather than later? Let me know if you want the message to be changed to something more informative - I'm not sure what the equivalent is in 4.0.

### DIFF
--- a/model/Image.php
+++ b/model/Image.php
@@ -95,8 +95,8 @@ class Image extends File implements Flushable {
 
 	/**
 	 * Retrieve the original filename from the path of a transformed image.
-	 * Any other filenames pass through unchanged. 
-	 * 
+	 * Any other filenames pass through unchanged.
+	 *
 	 * @param string $path
 	 * @return string
 	 */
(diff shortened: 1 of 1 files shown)


tractorcow:
It's not too late for 3.3, but it IS too late for 3.3.0.

The AssetLibraryPreview addition, however, can't go into 3.3. It'll need to be 3.4, so thus 3 branch. Up to you if you put it all into 3, or split 3.3 and 3 pull requests.

tractorcow:
I make semver sound so complicated...

jonom:
I'll just put it all in 3. No one should be using those generate methods on their own anyway. https://github.com/silverstripe/silverstripe-framework/pull/5043

PR #5043, opened 2016-02-15
Image deprecations and optimisation

jonom:
Consolidates thumbnail methods to use the common image manipulation API, to prevent unnecessary creation of identical images with different filenames. Generate methods also tagged for removal in 4.0.

### DIFF
--- a/model/Image.php
+++ b/model/Image.php
@@ -249,8 +249,10 @@ class Image extends File implements Flushable {
 	 * @param integer $width The width to size within
 	 * @param integer $height The height to size within
 	 * @return Image_Backend
+	 * @deprecated 4.0 Generate methods are no longer applicable
 	 */
 	public function generateFit(Image_Backend $backend, $width, $height) {
+		Deprecation::notice('4.0', 'Generate methods are no longer applicable');
 		return $backend->resizeRatio($width, $height);
 	}
 
@@ -294,8 +296,10 @@ class Image extends File implements Flushable {
 	 * @param integer $width Width to crop to
 	 * @param integer $height Height to crop to
 	 * @return Image_Backend
+	 * @deprecated 4.0 Generate methods are no longer applicable
 	 */
 	public function generateFill(Image_Backend $backend, $width, $height) {
+		Deprecation::notice('4.0', 'Generate methods are no longer applicable');
 		return $backend->croppedResize($width, $height);
 	}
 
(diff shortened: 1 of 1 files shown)


model/Image.php
```diff
@@ -582,38 +592,77 @@ public function generateSetSize(Image_Backend $backend, $width, $height) {
...
 		return $backend->paddedResize($this->stat('cms_thumbnail_width'),$this->stat('cms_thumbnail_height'));
 	}
 
 	/**
 	 * Resize this image for preview in the Asset section. Use in templates with $AssetLibraryPreview.
+	 *
+	 * @return Image_Cached|null
+	 */
+	public function AssetLibraryPreview() {
+		return $backend->paddedResize($this->stat('asset_preview_width'),$this->stat('asset_preview_height'));
```
tractorcow: Undefined variable $backend here. :)

jonom: Oops!

jonom:
Have reduced this a bit - AssetLibraryPreview and AssetLibraryThumbnail don't seem to be used anywhere in Framework or CMS and are not present in 4.0 so no sense introducing new API. I notice the thumbnail config variables are a bit of a mess and used in unintended ways but I see you've addressed that in 4.0 (Thanks!). So it's really just StripThumbnail getting it's own method now, and a few deprecation notices.

model/Image.php
```diff
@@ -582,38 +592,59 @@ public function generateSetSize(Image_Backend $backend, $width, $height) {
...
+		Deprecation::notice('4.0', 'Generate methods are no longer applicable');
 		return $backend->paddedResize($this->stat('asset_thumbnail_width'),$this->stat('asset_thumbnail_height'));
 	}
 
 	/**
 	 * Resize this image for use as a thumbnail in a strip. Use in templates with $StripThumbnail.
+	 *
+	 * @return Image_Cached|null
+	 */
+	public function StripThumbnail() {
```
tractorcow: Can you also PR to master to make sure this method is in ImageManipulations trait?

jonom: It's [already there](http://api.silverstripe.org/master/class-SilverStripe.Filesystem.ImageManipulation.html#_StripThumbnail) ;)

tractorcow: You're so efficient you'd done it before I'd even asked!

PR #5034, opened 2016-02-14
Replace Compass with gulp-sass

scott1702:
Followed on from https://github.com/silverstripe/silverstripe-framework/pull/5027

CMS PR - https://github.com/silverstripe/silverstripe-cms/pull/1393

@dhensby @jonom

PR #5046, opened 2016-02-16
Added Bootstrap Alpha.2

scott1702:
- Managed by npm
- Created an `npm run css` task to compile
- Updated shrinkwrap
- Updated changelog
- autoprefix css in framework

The other half of https://github.com/silverstripe/silverstripe-framework/pull/5034 which was split from https://github.com/silverstripe/silverstripe-framework/pull/5027

Based on which gets merged first (https://github.com/silverstripe/silverstripe-framework/pull/5034 or this), the unmerged PR will need some changing. ie both of those PRs have their own `npm run css` task which compiles different things

@dhensby @jonom @ianherbert


scott1702:
This has been updated now that #5034 has been merged.

admin/scss/_preview.scss
```diff
@@ -277,7 +277,6 @@
 			@extend .icon-sprites-64x64;
 			@include sprite($sprites-64x64-preview);
 
-			display: block;
```
chillu: Why did you remove this?

scott1702: The second param of the sprite mixin is the display value, if one is not given then it defaults to `display: block` making that line redundant (see https://github.com/silverstripe/silverstripe-framework/blob/master/admin/scss/_spritey.scss#L184).

PR #5051, opened 2016-02-17
BUG Setting $tmpFile property in Upload class. Attempt 2.

flamytwista:
Attempt 2.

There is a protected $tmpFile property.
If it is defined in this place (see my commit)
I would be able to create an extension for upload class like this (example is simplified):
```
class CustomUploadExtension extends Extension {
    public function onAfterLoad($file){
        $originalName = $this->owner->getField('tmpFile')['name'];
        $file->setField('Title', $originalName);
        $file->write();
        return;
    }
}
```
------------------------------------
```
Upload::add_extension('CustomUploadExtension');
Config::inst()->update('FileNameFilter', 'default_replacements', array(
    '/\s/' => '-', // remove whitespace
    '/_/' => '-', // underscores to dashes

    '/а/u'=>"a",'/б/u'=>"b",'/в/u'=>"v",
        /*......*/
    '/Я/u'=>"YA",

    '/[^A-Za-z0-9+.\-]+/' => '', // remove non-ASCII chars, only allow alphanumeric plus dash and dot
    '/[\-]{2,}/' => '-', // remove duplicate dashes
    '/^[\.\-_]+/' => '', // Remove all leading dots, dashes or underscores
));
```
It is useful for uploading cyrrilic-named file and leaving title as original filename.

### DIFF
--- a/filesystem/Upload.php
+++ b/filesystem/Upload.php
@@ -134,6 +134,8 @@ class Upload extends Controller {
 
 		$valid = $this->validate($tmpFile);
 		if(!$valid) return false;
+		
+		$this->tmpFile = $tmpFile;
 
 		// @TODO This puts a HUGE limitation on files especially when lots
 		// have been uploaded.


tractorcow:
I would prefer to minimise the size of the state of the current upload. Otherwise you need to maintain this value to prevent it getting stage (e.g. over multiple file uploads).

Why not add `$tmpFile` to the extension?

```php
$this->extend('onAfterLoad', $this->file, $tmpFile);
```

tractorcow:
Upload::$tmpFile is protected and neither set or referenced anywhere. Maybe it should just be removed. :D

flamytwista:
@tractorcow
Your offer 
`$this->extend('onAfterLoad', $this->file, $tmpFile);`
is appropriate too:
before I've created pull-request I was selecting between this two decisions.

I guess I need to create new pull-request with this code, isn't it?
`$this->extend('onAfterLoad', $this->file, $tmpFile);`

tractorcow:
You can make a new PR, or you could just overwrite this one with your change. :)

flamytwista:
@tractorcow 
I've created new PR.
https://github.com/silverstripe/silverstripe-framework/pull/5076
Look please.

tractorcow:
Closed with #5076

PR #5055, opened 2016-02-18
Removing unused thirdparty dependency - greybox.

benmanu:


benmanu:
@tractorcow - part of the Open Sorcerers internal backlog

PR #5056, opened 2016-02-19
FIX Do not hang on nested parameters in search context

tractorcow:
Backport of https://github.com/silverstripe/silverstripe-framework/pull/4619 to 3.2

### DIFF
--- a/core/ArrayLib.php
+++ b/core/ArrayLib.php
@@ -164,6 +164,23 @@ class ArrayLib {
 	}
 
 	/**
+	 * Similar to array_map, but recurses when arrays are encountered.
+	 *
+	 * Actually only one array argument is supported.
+	 *
+	 * @param $f callback to apply
+	 * @param $array array
+	 * @return array
+	 */
+	public static function array_map_recursive($f, $array) {
+		$applyOrRecurse = function($v) use($f) {
+			return is_array($v) ? ArrayLib::array_map_recursive($f, $v) : call_user_func($f, $v);
+		};
+
+		return array_map($applyOrRecurse, $array);
+	}
+
+	/**
 	 * Recursively merges two or more arrays.
 	 *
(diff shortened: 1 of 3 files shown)


ntd:
This is still missing from 3.2.2. Do you plan to include it in 3.2.3?

core/ArrayLib.php
```diff
@@ -164,6 +164,23 @@ public static function in_array_recursive($needle, $haystack, $strict = false) {
...
 	/**
+	 * Similar to array_map, but recurses when arrays are encountered.
+	 *
+	 * Actually only one array argument is supported.
+	 *
+	 * @param $f callback to apply
+	 * @param $array array
+	 * @return array
+	 */
+	public static function array_map_recursive($f, $array) {
```
dhensby: Do funcs on the array lib need to start with `array_`?

tractorcow: They don't have to I guess?

tractorcow:
@dhensby do you think it's too late to fix this? It's a new API so probably can't be included in 3.2 anymore.

The problem is that we need a recursive function, so it would need to invoke itself (thus new API).

@ntd do you think you could make a version of this fix that's semver compatible? I.e. adds no new methods (inline only). You might need to put this into a self-referential closure. See http://stackoverflow.com/questions/7877970/is-it-possible-to-reference-an-anonymous-function-from-within-itself-in-php for how to do this.

tractorcow:
Thanks for that!

ntd:
:thumbsup:

dhensby:
@tractorcow do we not want to fix this "properly" in 3.x?

tractorcow:
That's what https://github.com/silverstripe/silverstripe-framework/pull/4619 is

dhensby:
Great! Thanks everyone!

PR #5057, opened 2016-02-19
Removing unused thirdparty dependency - query-selector.

benmanu:
@tractorcow  - part of the Open Sorcerers internal backlog. This is already included as part of the thirdparty/jquery-entwine dist files.


tractorcow:
Restarting tests.

PR #5058, opened 2016-02-19
Removing unused thirdparty dependency - jquery-metadata.

benmanu:
@tractorcow - part of the Open Sorcerers internal backlog. Already included on it's dependent module jquery-validate in `thirdparty/jquery-validate/lib`.


tractorcow:
You know you can put all these remove PRS into one request... breaking it up just means we have to wait for all the tests to pass. :)

PR #5061, opened 2016-02-19
Removing unused thirdparty dependency - jquery-effen.

benmanu:
@tractorcow - part of the Open Sorcerers internal backlog. Last one of the removals!


tractorcow:
One less effen thirdparty.

PR #5063, opened 2016-02-19
Moving thirdparty module to npm dependency - json-js

benmanu:
* Adding [json-js](https://www.npmjs.com/package/json-js) npm dependency.
* Replacing `admin/javascript/src/bundles/lib.js` requirement with npm dependency.
* Updating shrinkwrap.
* Running `$ npm run build` to rebuild bundle-lib.js.


flashbackzoo:
The build file has additional comments because shrinkwrap doesn't currently lock dev-dependencies and newer versions of Browserify generate bundles slightly differently.

This has been fixed in an upcoming PR (https://github.com/open-sausages/silverstripe-framework/tree/integrate-react) where dev-dependencies are also locked down and prod versions of JavaScript bundles are minified.

tractorcow:
Indentation in the dist files are weird... is there a setting we need to change?

I've restarted the tests since they stalled.

flashbackzoo:
It's because Ben's on a newer version of Browserify and or Babel (dev-dependencies) which aren't currently locked down by shrinkwrap.

There are updates to the build tooling in the above mentioned branch which will fix this by locking down dev-dependencies and minifying JavaScript.

tractorcow:
Is browserify version something we can lock down with packages.json?

flashbackzoo:
It's something we can lock down with a shrinkwrap file, the same way we lock deps down with a composer lock file. Our current shrinkwrap implementation doesn't lock down dev-dependencies because the native Node shrinkwrap is a bit rubbish at handing dev-dependencies. We have a workaround in in the above branch which uses https://github.com/uber/npm-shrinkwrap via the `npm run lock` script.

flashbackzoo:
I think this is fine for now, the version Ben has committed it functionally fine, and won't make it into a release in it's current state because the updated build tooling (when merged) will tidy it up.

tractorcow:
Rebase please.

tractorcow:
@benmanu

benmanu:
@tractorcow rebased on top of latest master, still some odd build changes, but not as bad as before. If it's not looking as neat as it should I could throw it into a new pull request and cherry pick only the module dep changes, see if that builds out neater

tractorcow:
Some build problems...

tractorcow:
```
1) GridFieldAddExistingAutocompleterTest::testSearch
InvalidArgumentException: File framework/thirdparty/json-js/json2.js does not exist
```

Repeated lots of times. :D

tractorcow:
Hm, it looks like I need to make filesystem less sensitive to test errors.

See https://github.com/silverstripe/silverstripe-framework/pull/5117 for the code I used to help diagnose the errors in this PR.

tractorcow:
You need to remove all instances of `THIRDPARTY_DIR . '/json-js/json2.js',`

tractorcow:
@benmanu This needs rebasing yet again.

benmanu:
@tractorcow rebased with latest master.

tractorcow:
Thanks @benmanu

PR #5074, opened 2016-02-22
BUG Correct behaviour of publish with $createNewVersion = true

### LINKED ISSUE #5040: Reverting a page to a specific version is not working correctly : SS 3.2

tractorcow:
Fixes #5040
Cleanup code to make behaviour more apparent.

Main points in this fix:
- Instead of guessing next version, set Version = null prior to write to force new version on write.
- During write, get version from base table instead of the current table (as only the base table has a Version field).

Optional optimisation:
- In 4.0 we should probably pass back nextVersion from augmentWriteVersioned to prevent the DB::query having to be executed for multiple tables.

### DIFF
--- a/model/Versioned.php
+++ b/model/Versioned.php
@@ -5,6 +5,8 @@
  * allowing you to rollback changes and view history. An example of this is
  * the pages used in the CMS.
  *
+ * @property int $Version
+ *
  * @package framework
  * @subpackage model
  */
@@ -487,12 +489,12 @@ class Versioned extends DataExtension implements TemplateGlobalProvider {
 
 						foreach($versionedTables as $child) {
 							if($table === $child) break; // only need subclasses
-							
+
 							// Select all orphaned version records
 							$orphanedQuery = SQLSelect::create()
 								->selectField("\"{$table}_versions\".\"ID\"")
 								->setFrom("\"{$table}_versions\"");
-								
+
 							// If we have a parent table limit orphaned records
 							// to only those that exist in this
 							if(DB::get_schema()->hasTable("{$child}_versions")) {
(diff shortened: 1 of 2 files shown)


tractorcow:
@priyashantha  please test and confirm this fixes your issue.

model/Versioned.php
```diff
@@ -805,56 +817,64 @@ public function latestPublished() {
...
 		if(is_numeric($fromStage)) {
 			$from = Versioned::get_version($baseClass, $this->owner->ID, $fromStage);
 		} else {
 			$this->owner->flushCache();
-			$from = Versioned::get_one_by_stage($baseClass, $fromStage, "\"{$baseClass}\".\"ID\"={$this->owner->ID}");
+			$from = Versioned::get_one_by_stage($baseClass, $fromStage, array(
+				"\"{$baseClass}\".\"ID\" = ?" => $this->owner->ID
+			));
+		}
+		if(!$from) {
```
tractorcow: This early return (instead of a huge if() body) makes this change look larger than it really is. Maybe view side-by-side?

priyashantha:
@tractorcow Thank you very much for the update and I'll have test with the issue and let you know :)

tractorcow:
Yikes, it looks like there are some DB schema specific failures here.

tractorcow:
I forgot that some connectors like returning string values for numeric values... changed `===` to `==` for test.

model/Versioned.php
dhensby: appending `?w=1` to the diff URL will hide indentation changes

dhensby:
thanks :tractor: :cow:

tractorcow:
Yay.

priyashantha:
@tractorcow This is great and working fine. Is there any new tag including this fix to refer us in composer.json?

tractorcow:
Not yet, I'm investigating a possible regression in this fix. We will need to double check it's compatibility with other modules before we tag it.

See https://github.com/silverstripe-labs/silverstripe-fulltextsearch/pull/109 which was broken at this fix.

PR #5081, opened 2016-02-23
BUG Fix deprecated API usage in DataFormatter API and DataObjectTest

tractorcow:
PHPDoc Cleanup
API DataFormatter::convertStringToArray is now abstract

These classes are so messy.


tractorcow:
THANKS!

dhensby:
The data formatters are broken because they don't use the same priority for using `get` functions, methods and properties as one would expect 

This causes problems in the restfuleserver module

tractorcow:
Do you mean broken now, or broken with this change? More info please! :D

dhensby:
Sorry, were broken - not by this change.

Issue #3186, opened 2014-06-04
Versioned::$versionableExtensions - need a way to change this other than changing Versioned.php


tylerkidd:
Currently Versioned.php has this protected static pretty much hard coded:

```
protected static $versionableExtensions = array('Translatable' => 'lang');
```

It's used in two places. What if we changed the references to it from:

```
Versioned::$versionableExtensions
```

to

```
Config::inst()->get('Versioned', 'versionableExtensions')
```

Allowing it to be manipulated from other modules via:

```
Config::inst()->update('Versioned', 'versionableExtensions', array('Extension' => array('suffix1','suffix2')));
```

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-framework/pull/5085

PR #5085, opened 2016-02-23
#3186 - Enhancement and fix for issue - cannot change Versioned $versionableExtensions

tylerkidd:
Allow Versioned $versionableExtensions to be manipulated via config.


tractorcow:
This is an excellent PR, thanks!

tylerkidd:
Thanks, @tractorcow !

dhensby:
Late to the party, but awesome work!

PR #5087, opened 2016-02-23
Fix incorrect height on main nav after toggle

### LINKED ISSUE #5012: BUG: CMS Menu shorter after restoring with long menus lists (3.3 rc1)
Looks like there's some odd behavior going on with the CMS menu in 3.3 on a fresh load of the cms with a long list of items and wrapped items, the cms menu on the left is full height and does not have a scroll bar at some resolutions. However upon hiding and restoring the menu the menu is shorter than it was and a scroll bar appears. If you resize the window the height goes back to full height however the scroll bar remains because of the wrapping it causes. I'm seeing this in Firefox 44, and Chrome 48. Appears this was also a thing in 3.2.

cms-menu-bug

scott1702:
@UndefinedOffset, Fixes #5012

### DIFF
--- a/admin/javascript/LeftAndMain.Menu.js
+++ b/admin/javascript/LeftAndMain.Menu.js
@@ -204,6 +204,7 @@
 			fromContainingPanel: {
 				ontoggle: function(e){
 					this.toggleClass('collapsed', $(e.target).hasClass('collapsed'));
+					$(window).resize(); //Trigger jLayout
 				}
 			},
 


chillu:
OK, this causes Behat builds to fail in 3.2, 3.3 and master at the moment. They're breaking on `cms`, not `framework`, which is why it wasn't noticed in this PR. The symptom is that Behat attempts to use the "Insert Link" feature, opens a page dropdown, but it immediately closes. Which means its target isn't clickable, failing the step.

Here's the crazy chain of events:
- Opening the treedropdown for the first time causes an ajax load
- Since https://github.com/silverstripe/silverstripe-framework/commit/3bac6fadfe118bce2e2f45992bd318c2e3c47926, all ajax responses call an `ajaxComplete` handler in `LeftAndMain.Menu.js`, which in turn calls `togglePanel()`.
- Since this PR, `togglePanel()` will also call `window.resize()`
- `window.resize()` triggers a close of the treedropdown, presumably to avoid it going off screen (https://github.com/silverstripe/silverstripe-framework/blob/3/javascript/TreeDropdownField.js#L10)

David, can you remember why `togglePanel()` needs to be called on *any* ajax response? It sounds way too global, what's the benefit over calling it on page load only?

That took about an hour of detective work tonight ;)

/cc @flashbackzoo @tractorcow @scott1702 @dhensby

tractorcow:
Let's just revert then. :)

chillu:
Fixed by triggering a more lowlevel resize event. Ideally it would be called something more accurate like `componentresize` rather than `windowresize`, but it'll have to do for now - will all change in 4.x anyway :) https://github.com/silverstripe/silverstripe-framework/pull/5135

Issue #5086, opened 2016-02-23
Member_Validator doesn't validate existing Members (identifier field)


bummzack:
There seem to be some issues with [`Member_Validator`](https://github.com/silverstripe/silverstripe-framework/blob/8e4db95f7297f4621413609fd0ff5bbe0385bf37/security/Member.php#L1812):

1.  It doesn't validate when there's no ID present (eg. doesn't work when creating new members)
2.  It seems to use some SilverStripe 2.x request lookup to determine the member-id (`$_REQUEST['ctf']['childID']`)
3.  It has a completely broken way to run validation on extensions (extending the validator basically doesn't work at all)

This isn't too much of an issue, since once validation fails, there is always `Member::onBeforeWrite` that will act as a safety net. The nice thing about a properly working validator is, that custom built Forms won't be completely wiped when trying to create a member.

Here's a [gist](https://gist.github.com/bummzack/95840febd7515198b36a) of what I'm currently using as a replacement. It could be a *complete* replacement, if [this piece of code](https://github.com/silverstripe/silverstripe-framework/blob/8e4db95f7297f4621413609fd0ff5bbe0385bf37/security/Group.php#L118) would be aware of injections :)

I'd gladly provide my implementation of the validator as a replacement for `Member_Validator`, if wanted.

tractorcow:
I think your validator is a perfect drop-in. Please PR!

Also at the same time, I also agree with making the new Member_Validator() usage change to use injector API.

tractorcow:
I suggest this alteration.

```php
$members = Member::get()->filter($identifierField, $data[$identifierField]);
if($id) {
	$members = $members->exclude('ID', $id);
}
$existingMember = $members->First();
```

bummzack:
Ok, will do and create a PR.

tractorcow:
It's super preferred, since it handles all DB field manipulations behind the scenes, and allows developers to just work with model-level names. It's also much more secure, and handles null values nicely. :)

tractorcow:
Please PR to the 3.3 branch as a bugfix.

bummzack:
Ah, good to know. I thought the new parameterized syntax was preferred with the introduction of 3.2. Thanks for clarifying.

tractorcow:
Yeah, it's still useful when you need to work with very low level SQL queries. :)

bummzack:
[...]

PR #5088, opened 2016-02-23
Fix for issue #5086

bummzack:
- allow extensions.
- remove old code and replace with new syntax and config api.

### DIFF
--- a/security/Member.php
+++ b/security/Member.php
@@ -1801,21 +1801,30 @@ class Member_ForgotPasswordEmail extends Email {
  * Member Validator
  *
  * Custom validation for the Member object can be achieved either through an
- * {@link DataExtension} on the Member object or, by specifying a subclass of
+ * {@link DataExtension} on the Member_Validator object or, by specifying a subclass of
  * {@link Member_Validator} through the {@link Injector} API.
+ *
+ * Custom required fields can be set via config API, eg.
+ * Member_Validator:
+ *   customRequired:
+ *     - Surname
  *
  * {@see Member::getValidator()}
  *
  * @package framework
  * @subpackage security
  */
-class Member_Validator extends RequiredFields {
-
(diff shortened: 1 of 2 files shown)


security/Member.php
```diff
@@ -1801,22 +1801,31 @@ public function __construct() {
...
-
-	protected $customRequired = array(
+class Member_Validator extends RequiredFields
+{
+	/**
+	 * Fields that are required by this validator
+	 * @config
+	 * @var array
+	 */
+	private static $customRequired = array(
```
tractorcow: Oh, I just realised that changing the customRequired to config would be an API change. this COULD go into 3 branch, but you'd need to keep it as protected var to merge it into 3.3.

tractorcow:
Is there a chance we could have a test for this? Maybe one that checks existing records produce an error.

security/Member.php
bummzack: I see. I guess that would make the config API unusable for that case? 

Then I'll just revert to `protected $customRequired`. Currently, modifying the required fields of the Member_Validator can be achieved by implementing `updateValidator` in an extension to Member, so the config system isn't really needed. I just thought it would be more consistent.

bummzack:
Will write tests for this and update the PR.

security/Member.php
dhensby: You can still use the config system, just don't change the visibility of this var. 

The config system does parse the PHP code, but that doesn't mean it has to be defined in PHP to be in the config

bummzack: I'm aware of that, but how should I apply the variables? Merge the ones from config and the values in the protected variable? Does that make sense?

bummzack:
Sorry, I had trouble updating my branch without adding multiple commits. Closing this in favor of #5090 where all commits are squashed into one.

PR #5089, opened 2016-02-24
added if statement to catch NULL validators

jakedaleweb:
Problem found on CWP:

"Since upgrading to 1.2.0, we've noticed a problem with workflows that have a second transition on them ie accept or reject. When an author goes to publish, they receive a message in the right hand corner saying "Not Allowed".

When trying to preview pages, we get a message saying "Service Temporarily Unavailable". 

We can go through and approve items in the workflow, but cannot preview them first. It is as if the pages are locked out for editing for everyone but when this is disabled for the workflow it makes no difference. 

Workflows with single transitions aren't affected. "

Error was:

"auth.log:Feb 24 14:42:08 EXAMPLE SilverStripe[27392]: PHP Catchable fatal error:  Argument 1 passed to Form::setValidator() must be an instance of Validator, null given, called in /sites/EXAMPLE/releases/20160223201643/framework/admin/code/LeftAndMain.php on line 1298 and defined in /sites/EXMAPLE/releases/20160223201643/framework/forms/Form.php on line 654"

### DIFF
--- a/admin/code/LeftAndMain.php
+++ b/admin/code/LeftAndMain.php
@@ -1301,7 +1301,9 @@ class LeftAndMain extends Controller implements PermissionProvider {
 				// The clientside (mainly LeftAndMain*.js) rely on ajax responses
 				// which can be evaluated as javascript, hence we need
 				// to override any global changes to the validation handler.
-				$form->setValidator($validator);
+				if($validator != NULL){
+					$form->setValidator($validator);
+				}
 			} else {
 				$form->unsetValidator();
 			}


tractorcow:
Travis is overwhelmed at the moment; I'll restart these later.

kinglozzer:
Is this the right place to fix this? Considering that code block as a whole:

```php
if($record->hasMethod('getCMSValidator')) {
	$validator = $record->getCMSValidator();
	$form->setValidator($validator);
}
```

So the model in question has defined `getCMSValidator()`, but it just returns null? Seems to me that it should be fixed there or the method should just be removed

tractorcow:
I think @jakedaleweb is in the right though; Having a method isn't a promise to return a non-null value. The issue is to do with strict typehints not allowing nulls as a parameter, though. Can we merge this?

kinglozzer:
That’s a fair point - given that the method doesn’t exist in an interface or anything we don’t have any guarantees about what it will return. Consider me convinced :)

Other minor, non-merge-blocking comments:

- Should we use `if ($validator instanceof Validator)` instead?
- Is 3 the right branch for this?

tractorcow:
It could go into 3.2 if we wanted. If a non-validator type is returned (non-empty) that'd be another error condition. I don't think we need to be too strict.

tractorcow:
@kinglozzer please merge if you're satisfied with this PR.

PR #5090, opened 2016-02-24
Fixes #5086

### LINKED ISSUE #5086: Member_Validator doesn't validate existing Members (identifier field)

bummzack:
- allow extensions.
- remove old code and replace with new syntax and add config api.

Added unit-tests.

### DIFF
--- a/security/Group.php
+++ b/security/Group.php
@@ -115,7 +115,7 @@ class Group extends DataObject {
 			$config->getComponentByType('GridFieldAddExistingAutocompleter')
 				->setResultsFormat('$Title ($Email)')->setSearchFields(array('FirstName', 'Surname', 'Email'));
 			$config->getComponentByType('GridFieldDetailForm')
-				->setValidator(new Member_Validator())
+				->setValidator(Member_Validator::create())
 				->setItemEditFormCallback(function($form, $component) use($group) {
 					$record = $form->getRecord();
 					$groupsField = $form->Fields()->dataFieldByName('DirectGroups');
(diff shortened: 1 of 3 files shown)


bummzack:
I seem to have butchered some doc-comments. I'm really sorry. Will PR again.

dhensby:
@bummzack ping me with a mention when it's open

tractorcow:
me also :D

bummzack:
@tractorcow It's over here: #5091, sorry I created two PRs for this.

PR #5091, opened 2016-02-24
Fix for issue #5086

bummzack:
- allow extensions.
- remove old code and replace with new syntax and add config API.

Added unit-tests.

### DIFF
--- a/security/Member.php
+++ b/security/Member.php
@@ -766,6 +766,7 @@ class Member extends DataObject implements TemplateGlobalProvider {
 	 */
 	public function getValidator() {
 		$validator = Injector::inst()->create('Member_Validator');
+		$validator->setForMember($this);
 		$this->extend('updateValidator', $validator);
 
 		return $validator;
@@ -1801,21 +1802,39 @@ class Member_ForgotPasswordEmail extends Email {
  * Member Validator
  *
  * Custom validation for the Member object can be achieved either through an
- * {@link DataExtension} on the Member object or, by specifying a subclass of
+ * {@link DataExtension} on the Member_Validator object or, by specifying a subclass of
  * {@link Member_Validator} through the {@link Injector} API.
- *
+ * The Validator can also be modified by adding an Extension to Member and implement the
+ * <code>updateValidator</code> hook.
  * {@see Member::getValidator()}
+ *
+ * Additional required fields can also be set via config API, eg.
+ * <code>
+ * Member_Validator:
+ *   customRequired:
+ *     - Surname
+ * </code>
  *
  * @package framework
  * @subpackage security
  */
-class Member_Validator extends RequiredFields {
-
+class Member_Validator extends RequiredFields
+{
+	/**
+	 * Fields that are required by this validator
+	 * @config
+	 * @var array
+	 */
 	protected $customRequired = array(
 		'FirstName',
 		'Email'
 	);
 
+	/**
+	 * Determine what member this validator is meant for
+	 * @var Member
+	 */
+	protected $forMember = null;
 
 	/**
 	 * Constructor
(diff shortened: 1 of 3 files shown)


bummzack:
@dhensby Ok, this one should be good if CI passes.

security/Member.php
```diff
@@ -1842,31 +1860,35 @@ public function __construct() {
...
+		if(!$id && ($ctrl = $this->form->getController())){
+			// get the record when within GridField (Member editing page in CMS)
+			if($ctrl instanceof GridFieldDetailForm_ItemRequest && $record = $ctrl->getRecord()){
+				$id = $record->ID;
+				// set the ID to the data array, so that extensions can also use it
+				$data['ID'] = $id;
+			}
+		}
+
+		$members = Member::get()->filter($identifierField, $data[$identifierField]);
```
dhensby: What if `$data[$identifierField]` is not set?

dhensby:
see https://github.com/silverstripe/silverstripe-framework/pull/5088

security/Member.php
bummzack: You mean, if somebody removes the identifier field from the list of required fields? Seems unlikely, but better be safe than sorry… should I create a new PR?

bummzack:
Is there something I missed from the discussion in #5088?

dhensby:
No, in just cross linking so that people can see the discuss ons

security/Member.php
dhensby: No. You can just amend this PR and force push. The updates will come into the PR automatically

bummzack:
@dhensby Thanks. I forgot about amending commits. I updated the commit with a safeguard check for `$data[$identifierField]`. I also went ahead and implemented the behavior I proposed in #5086 to solve cases where the validator might not have an `ID` passed with `$data`, yet still "belongs" to a Member.

bummzack:
Any idea what this test failure means: 'Failed asserting that 'JoeEdited' is not equal to <string:JoeEdited>.' ?

dhensby:
I've restarted those tests, bizarre that a couple have it and a couple don't not really sure.

The behat test is more worrying. If you check this out locally and test, does it work in the CMS when you add/save/edit/delete members?

bummzack:
Yeah, everything works fine in the CMS. The warning dialogue now shows up underneath the E-Mail field though… maybe that's causing a problem there?

dhensby:
hmm - I don't think that'd be a problem

bummzack:
Strangely enough all tests have passed with my previous PR… not sure why it failed now. When testing in the CMS, everything works/validates as expected.

dhensby:
someone with more knowledge of PHPUnit will have to explain this one :/

bummzack:
@dhensby I figured out the problem. It is important that the validator checks the `ID` passed via `$data` first. Only if there's no ID passed via `$data` or via controller->record, the member assigned to the Validator should come into play.

dhensby:
Thanks - I'm pretty happy with this and it could lay the foundation for skipping validation on default admins too

dhensby:
@tractorcow thoughts?

dhensby:
oh - as it's a new feature, maybe it'll have to go against 3 :/

bummzack:
You mean, it's a feature that the `Member_Validator` is suddenly working? For me this is more of a Bugfix :)

tractorcow:
Yeah sorry, I meant to say this on the other thread: Can you raise on the 3 branch since it introduces new method API?

tractorcow:
Code is pristine by the way.

bummzack:
Ok, PR against 3 branch is here: #5098. Sorry for the many PRs

Issue #3745, opened 2014-12-29
Object: 'invokeWithExtensions' $argument should be call-by-reference to be consistent with 'extend'


peda:
In the Object class currently the `$argument` of `invoke withExtensions` is a call-by-value:

``` php
public function invokeWithExtensions($method, $argument = null)
```

whereas the arguments are call-by-reference for the method `extend`:

``` php
public function extend($method, &$a1=null, &$a2=null, &$a3=null, &$a4=null, &$a5=null, &$a6=null, &$a7=null) 
```

In order to be consistent it would be great if the `$argument` argument of `invokeWithExtensions` could be changed to call by-reference as well. This inconsistency already lead to mistakes even in silvertripe-labs modules (https://github.com/silverstripe-labs/silverstripe-googlesitemaps/issues/78)

patricknelson:
:+1: I think I'll do a quick PR against this since it's pretty annoying.

tractorcow:
Too late I've already done it. It's getting PR up soon. :D

patricknelson:
Wait... hey! So -- you put that up against `3` right? And you were fine with the mild API change, that is: Forcing `$arg1` to effectively be a pass by reference making calls like `$obj->invokeWithExtensions('method', 'string constant')` illegal)?

tractorcow:
It's against master at the moment, and is a part of a larger piece of work.

https://github.com/open-sausages/silverstripe-framework/commit/697d57bdba55db87a40b6c64732c6e8f0df4d02a#diff-82f6624f9301c842e9bd32d95d0f4be4R380

No I'm not fine with the API change... method signature changes are (nearly) always semver breaking.

patricknelson:
Yeah that's what I thought. Only after I dove into this did I realize that switching to pass-by-reference would likely break calls which were using literals and not variables :disappointed:  This makes my work **much** more frustrating since I refuse to modify core unless I know it'll make it into `3` so, instead, I'll try to approach a fix via PR against https://github.com/silverstripe-labs/silverstripe-googlesitemaps/issues/78 instead.

patricknelson:
Thanks for the head's up, by the way! Saved me a little bit of time at least!

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/pull/5094

PR #5094, opened 2016-02-25
API Extensible::invokeWithExtension has same method signature as Extensible::extend

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/3745

### DIFF
--- a/core/Extensible.php
+++ b/core/Extensible.php
@@ -48,7 +48,7 @@ trait Extensible {
 	private static $unextendable_classes = array('Object', 'ViewableData', 'RequestHandler');
 
 	/**
-	 * @var array all current extension instances.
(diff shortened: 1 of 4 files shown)


patricknelson:
Awesome :sunglasses::+1: 

Maybe someday when we only support PHP 5.6 and up, we'll be able to transition over to the *much better* variadic syntax where basically all you'd need to do is:

```php
public function invokeWithExtensions($method, &...$arguments) {
    // Do stuff here.
}
```

... so that then you can have unlimited arguments *and* by reference at that. For now though we'll have to settle with this ugliness since unfortunately `func_get_args()` doesn't support references :disappointed:

tractorcow:
Right, I saw that. :D

PR #5098, opened 2016-02-25
Fix for issue #5086

bummzack:
- allow extensions.
- remove old code and replace with new syntax and add config API.

Added unit-tests.

Establish a link between the member and the validator for said member.


dhensby:
thanks so much for seeing this through!

PR #5099, opened 2016-02-25
BUG Fix regression in gridfield get actions

tractorcow:
Fixes regression discovered in https://github.com/silverstripe/silverstripe-blog/issues/380#issuecomment-188955153


tractorcow:
I'm going to back-port this to 3.1 and 3.2, don't worry guys help is on the way!

PR #5103, opened 2016-02-28
Populate foreign key before getting CMS fields

hailwood:
In it's current state you need to revert to something like `Session::get('CMSMain.currentPage')` to get the foreign key of what the item you are creating relates to (e.g. a Book => has_many Author) - if you create a new Author you may need to reference the owning Book in the `getCMSFields` function.

This is just a small quality of life buff that populates that foreign key before calling `getCMSFields()` rather than after.

This should not break backwards compatibility in any way and isn't exactly a new feature so could be considered a bug fix.

### DIFF
--- a/forms/gridfield/GridFieldDetailForm.php
+++ b/forms/gridfield/GridFieldDetailForm.php
@@ -385,20 +385,22 @@ class GridFieldDetailForm_ItemRequest extends RequestHandler {
 				$actions->push(new LiteralField('cancelbutton', $text));
 			}
 		}
+		
+		// If we are creating a new record in a has-many list, then
+		// pre-populate the record's foreign key.
+		if($list instanceof HasManyList && !$this->record->isInDB()) {
+			$key = $list->getForeignKey();
+			$id = $list->getForeignID();
+			$this->record->$key = $id;
+		}
 
 		$fields = $this->component->getFields();
 		if(!$fields) $fields = $this->record->getCMSFields();
 
 		// If we are creating a new record in a has-many list, then
-		// pre-populate the record's foreign key. Also disable the form field as
-		// it has no effect.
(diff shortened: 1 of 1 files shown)


forms/gridfield/GridFieldDetailForm.php
```diff
@@ -385,20 +385,22 @@ public function ItemEditForm() {
 				$actions->push(new LiteralField('cancelbutton', $text));
 			}
 		}
+		
+		// If we are creating a new record in a has-many list, then
+		// pre-populate the record's foreign key.
+		if($list instanceof HasManyList && !$this->record->isInDB()) {
+			$key = $list->getForeignKey();
+			$id = $list->getForeignID();
+			$this->record->$key = $id;
```
tractorcow: Does it work if you retain the `if(!$this->record->isInDB()) {` check? I can forsee change detection being an issue here.

Besides, you only need this for new records (not in db) right?

tractorcow:
Really nice idea. :) Could you respond to the one concern I had... I'm not sure if we should re-assign the foreign key if it's already assigned.

hailwood:
@tractorcow I have retained the `isInDB()` check, I've simply merged it in with the main if since we only need that branch to run at all if it's not in the database.

tractorcow:
Thanks @hailwood !

SpiritLevel:
Yay! Thanks @dhensby for pointing this PR out to me...it was exactly what I was looking for! And thanks @hailwood for the fix...and, since gratitude is pouring out all over the place at the moment, thanks also go to @tractorcow for guiding this fix into the code base....what a team!

dhensby:
Morning @SpiritLevel  ;)

SpiritLevel:
@dhensby Gidday!

PR #5104, opened 2016-02-28
FIX Do not hang on nested parameters in search context

ntd:
Backport of 0b5a57389b09 for 3.2 that does not add a new API, as requested by @tractorcow on #5056  to be semver compatible.


tractorcow:
Excellent. :) Will merge once tests pass.

PR #5106, opened 2016-02-29
"npm run lock", exclude node_modules from manifest

chillu:
Requires https://github.com/silverstripe/silverstripe-cms/pull/1404

/cc @flashbackzoo


chillu:
@tractorcow @dhensby Can one of you merge please? It's a best practice that we've already established in our https://github.com/open-sausages/silverstripe-framework/tree/integrate-react/ branch, after battling with the native `npm shrinkwrap`. It's beneficial to get this back into core earlier than the React feature will be ready to merge in its entirety.

PR #5107, opened 2016-02-29
invokeWithExtensions should accept multiple args

hailwood:
Currently it's setup so it only accepts a single arg, and it isn't passed by reference even though it's effective purpose is to first call the method on the Object itself and finally on any extensions as opposed to extend which can accept 7 args and passes them by reference.

An example of where this is an issue is on the duplicate method of DataObject - the doWrite value is never received.
https://github.com/silverstripe/silverstripe-framework/blob/3/model/DataObject.php#L537

Since this function is the same +1 I've mirrored the constructor.
I'm not sure if this is likely to break anything now that we're passing by reference so I'd be happy to release in 3.4 or 3.3.* with just multiple parameters but not passing by reference, and in 4.0 update it to pass by reference if we believe this counts as a backwards compatibility breaker.

### DIFF
--- a/core/Object.php
+++ b/core/Object.php
@@ -939,13 +939,13 @@ abstract class Object {
 	 * all results into an array
 	 *
 	 * @param string $method the method name to call
-	 * @param mixed $argument a single argument to pass
+	 * @param mixed $a1,... up to 7 arguments to be passed to the method
 	 * @return mixed
 	 * @todo integrate inheritance rules
 	 */
-	public function invokeWithExtensions($method, $argument = null) {
-		$result = method_exists($this, $method) ? array($this->$method($argument)) : array();
-		$extras = $this->extend($method, $argument);
+	public function invokeWithExtensions($method, &$a1=null, &$a2=null, &$a3=null, &$a4=null, &$a5=null, &$a6=null, &$a7=null) {
+		$result = method_exists($this, $method) ? array($this->$method($a1, $a2, $a3, $a4, $a5, $a6, $a7)) : array();
+		$extras = $this->extend($method, $a1, $a2, $a3, $a4, $a5, $a6, $a7);
 
 		return $extras ? array_merge($result, $extras) : $result;
 	}
--- a/core/Object.php
+++ b/core/Object.php
@@ -943,7 +943,7 @@ abstract class Object {
 	 * @return mixed
 	 * @todo integrate inheritance rules
 	 */
-	public function invokeWithExtensions($method, &$a1=null, &$a2=null, &$a3=null, &$a4=null, &$a5=null, &$a6=null, &$a7=null) {
+	public function invokeWithExtensions($method, $a1=null, $a2=null, $a3=null, $a4=null, $a5=null, $a6=null, $a7=null) {
 		$result = method_exists($this, $method) ? array($this->$method($a1, $a2, $a3, $a4, $a5, $a6, $a7)) : array();
 		$extras = $this->extend($method, $a1, $a2, $a3, $a4, $a5, $a6, $a7);
 


tractorcow:
If you want this into 3, it's fine, but you cannot change the method constructior.

Please use func_get_args() as a semver-compatible alternative. Of course, you won't be able to do pass by reference, but we can't do that in 3.x anyway.

Also see https://github.com/silverstripe/silverstripe-framework/pull/5094 where this was implemented in 4.x.

dhensby:
@tractorcow adding params where they are $x=null is backwards compatible new feature, isn't it? So should be allowed in 3.x

hailwood:
@tractorcow I am wondering why we cannot change the signature if it's backwards compatible as @dhensby says?

I'd rather avoid doing `func_get_args()` since it doesn't make it obvious that you can pass through multiple parameters unless you pull up the extended documentation in your IDE.

tractorcow:
No it's not, because existing overrides for those methods will suddenly have too few parameters to meet the method signature, and create a strict error.

tractorcow:
```php
class MyUserCode extends CoreCode {
	public function method($arg) {
		return parent::method($args) . "?";
	}
}

class CoreCode {
	public function method($arg) {
		return $arg . "!";
	}	
}
```

Then after adding args

```php
class MyUserCode extends CoreCode {
	public function method($arg) { // Strict method signature declaration error
		return parent::method($args) . "?";
	}
}

class CoreCode {
	public function method($arg, $arg2 = null) {
		return $arg . .$arg2 . "!";
	}	
}
```

hailwood:
Hmm wasn't aware of that one.

Ok I'll organize doing it with `func_get_args`. I'll drop the argument parameter as well since it's implied.
Any thoughts on what the docblock for the function should read as to make it obvious that you can pass in the 7+1 args?

dhensby:
Strict errors aren't breaking :/

This would be a problem for someone who overrode the method and added their own custom params; but it's low risk.

I'm not too fussed about it, so concede to @tractorcow - but I think this will raise its head again with something more important and this probably shouldn't be our hard-and-fast position

tractorcow:
@dhensby we need a tie breaker then. We need to come to some consensus on this, because we are definitely going to run into this going forward. :)

@chillu what about your view? Are strict errors semver violations or not?

sminnee:
We show strict errors by default, so if we include this and someone upgrades, they're going to start seeing errors. That's breaking in my book.

hailwood:
I may be thinking a bit narrow, but I cannot see any reason anyone would override `invokeWithExtensions` other than to apply this same patch (potentially with a more limited param set which would raise the strict error). In which case it's probably an ideal situation if the dev in charge of doing the updates gets some indication that this has changed so they don't keep applying the same patch anyway?

dhensby:
> I may be thinking a bit narrow, but I cannot see any reason anyone would override invokeWithExtensions other than to apply this same patch (potentially with a more limited param set which would raise the strict error). In which case it's probably an ideal situation if the dev in charge of doing the updates gets some indication that this has changed so they don't keep applying the same patch anyway?

I think that's probably true, but it's not safe to make these kinds of assumptions, who knows what crazy shizzle people are up to :P

hailwood:
Great, so the basic issue here is we need to decide if overriding `invokeWithExtensions` is part of the public API - given @dhensby's comment I'll mark that as one vote towards "public api".
What's everyone elses thoughts @tractorcow  @sminnee?

And again I'll ask, if we do go with the `func_get_args()` solution any thoughts on an appropriate docblock comment?

micmania1:
http://stackoverflow.com/questions/5549556/how-to-write-a-doxygen-comment-for-variadic-function-i-e-a-function-with-unde

Based on the above I think the following should be okay:

    * @param mixed ... Description

dhensby:
The decision has been made that this won't go into 3.x, I'm afraid - regardless of whether we think it's breaking or not. Current policy is that a strict notice is a breaking change.

We're relatively new to semver and so we're starting of being as paranoid as possible but this may become more lenient over time. Let's see :)

hailwood:
@dhensby  Is it that it won't go into 3.x or won't go into 3.x with a strict error? Since using `func_get_args`, `shift`,  and `call_user_func_array` wouldn't change the method signature but would give us the required end result until we go to 4.x with this properly patched.

dhensby:
I think if you can avoid the strict notice, then it'll go into 3.x (which doesn't make much sense ;) ) but yep - that's the decision being made here

tractorcow:
You can get it into 3.x, but you won't be able to benefit from pass-by-ref with using func_get_args(). If you can live with that though (and we have been) then that's fine.

tractorcow:
I'm going to close this PR in the mean time, since there's been a few months with no response. :)

tractorcow:
Please update and re-open if you wish to continue working on this. I'll still keep an eye on any updates to this. :D

hailwood:
Ah I didn't actually see your response @tractorcow 
I'll get onto that some time next week.

--
Matt

PR #5112, opened 2016-02-29
Improve documentation on documentation

tractorcow:
Improves the release notes on updating documentation, so that it's actually obvious what needs to be done.


tractorcow:
Updated based on feedback from @camfindlay

tractorcow:
Updated again to improve understanding of markdown update processes.

PR #5113, opened 2016-03-01
simplifying member param check in Member::canDelete

sunnysideup:
the member param check in Member::canDelete seemed overly complicated.

### DIFF
--- a/security/Member.php
+++ b/security/Member.php
@@ -1538,7 +1538,9 @@ class Member extends DataObject implements TemplateGlobalProvider {
 	 * Otherwise they'll need ADMIN or CMS_ACCESS_SecurityAdmin permissions
 	 */
 	public function canDelete($member = null) {
-		if(!$member || !(is_a($member, 'Member')) || is_numeric($member)) $member = Member::currentUser();
+		if(!is_a($member, 'Member')){
+			$member = Member::currentUser();	
+		} 
 
 		// extended access checks
 		$results = $this->extend('canDelete', $member);


tractorcow:
In some older parts of the code, can* methods sometimes allow the member ID to be passed in. We can't remove this behaviour except on the master branch.

Want to open it there instead?

tractorcow:
You could proabably even use `instanceof` if you wanted to avoid the literal classname. :D

tractorcow:
Sorry @sunnysideup :D I'm always finding a way to make things hard eh? :P

sunnysideup:
No, sorry, I forgot to choose the right branch.

In any case...

I am having a look at the canDelete (and canEdit / canView, etc...) for Member and DataObject.

There are a couple of things that I reckon need attention:

1. there seems to be a lot of repeat of code

Member:

image

DataObject:
image


as you can see, this is basically the same thing.  There are two issues:

a. if it is exactly the same thing, we should just rely on DataObject::extendedCan for both classes.  

b. if it is not exactly the same thing, then it is confusing and messy. 

I notice that The extendedCan method in the DataObject only use a member if a one if provided, while the canMethod extensions in Member use the Member::currentMember() and pass that to the extended canMethod methods (ha ha ha ha... ok, I hope that sentence makes sense). 

2. in Member, canDelete relies on canEdit, and canEdit relies on canView - is this standard? It seems dangerous, because, for example, if you extend the canView method in a DataExtension then this has a direct influence on the canDelete method for the member. 

image

image

All in all, it seems to me that these canMethods need to be simplified and/or standardised to improve Security of any code that interacts with or extends these methods.

sunnysideup:
@tractorcow what do you think about comments above?

tractorcow:
> a. if it is exactly the same thing, we should just rely on DataObject::extendedCan for both classes.

Right, let's do this then. :dancer: 

> in Member, canDelete relies on canEdit, and canEdit relies on canView - is this standard? It seems dangerous, because, for example, if you extend the canView method in a DataExtension then this has a direct influence on the canDelete method for the member.

I think it's a risk, yes. We probably shouldn't fall back to canView(). Maybe change it in 4.x?

PR #5114, opened 2016-03-01
ENHANCEMENT: more consistent Member can* methods.

sunnysideup:
now uses extendedCan rather than custom code
removed superfluous stuff
stopped relying on each other (canDelete was relying on canView to make the final decision)


sunnysideup:
this is just my draft.... great if it is accepted, but feel free to challenge it of course!

sunnysideup:
creating a new one

PR #5117, opened 2016-03-01
BUG Prevent fatal errors during test failure halt tests

tractorcow:
During test errors, sometimes dependencies might not get setup correctly.

This fix will prevent fatal errors from halting all test execution, and will prevent the original error from being masked.

See https://github.com/silverstripe/silverstripe-framework/pull/5063 for example of a valid non-fatal error getting masked fatal error which it triggered.

### DIFF
--- a/filesystem/flysystem/FlysystemAssetStore.php
+++ b/filesystem/flysystem/FlysystemAssetStore.php
@@ -5,6 +5,7 @@ namespace SilverStripe\Filesystem\Flysystem;
 use Config;
 use Generator;
 use Injector;
+use LogicException;
 use Session;
 use Flushable;
 use InvalidArgumentException;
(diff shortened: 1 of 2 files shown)


filesystem/flysystem/FlysystemAssetStore.php
```diff
@@ -110,6 +110,9 @@ public function setPublicFilesystem(Filesystem $filesystem) {
 	 * @return Filesystem
 	 */
 	public function getPublicFilesystem() {
+		if(!$this->publicFilesystem) {
+			throw new Exception("Filesystem misconfiguration error");
```
chillu: Shouldn't this be a more specific exception, e.g. `LogicException` at least?

dhensby: We can even have our own exception classes! Imagine. 

But yes, let's not just throw around base `Exception`s

chillu: "Our own exception classes" gets messy quickly - do you have one or more per class? Or per subsystem, e.g. all `SilverStripe/Forms` namespace? I find you get most of the value out of determining an "app-level" vs. "PHP error" exception. Although for specific utilities like Guzzle, it's very handy to have library specific exceptions.

tractorcow:
Re-pushed, but now tests are dying. :)

dhensby:
lul - add your API key ;)

chillu:
I've set up a separate `silverstripe-ci` user in Github for its API token allowance now. The token is just added to core builds (`framework` and `cms`) via an environment variable directly in Travis settings, not in `travis.yml`. Hopefully this will isolate core from these rate limiting failures at least.

PR #5119, opened 2016-03-01
BUG: cleaner Member::can* Methods

sunnysideup:
see: https://github.com/silverstripe/silverstripe-framework/pull/5113


sunnysideup:
this is interesting.  I inspected the error and I actually think the test is not written correctly ;-)

the test assumes that an extension always overwrites the canView return value.  However, in the test case, the canView is always false because there is no currentMember. Member::canView returns false because the user viewing the member does not exist and no one is logged in... as shown in this code:

image

Should I rewrite the code or should I rewrite the test?

sunnysideup:
here is where the test fails: https://github.com/silverstripe/silverstripe-framework/blob/master/tests/security/MemberTest.php#L527

dhensby:
Isn't the line 2 lines up where an extension is added meant to make canView always return true? 

I think the core code is wrong, not the test. If an extension wants to return true when there's no member, then it should be able to

sunnysideup:
agreed, will fix.

sunnysideup:
I have now changed the code.

security/Member.php
```diff
@@ -1470,25 +1470,30 @@ public function fieldLabels($includerelations = true) {
...
+		//most checks below rely on $member being a Member
+		if(!$member) return false;		
 
-		// members can usually edit their own record
-		if($member && $this->ID == $member->ID) return true;
+		// members can usually view their own record
+		if($this->ID == $member->ID) return true;
 
+		//standard check
 		if(
```
tractorcow: This whole block can be shortened to

```php
return Permission::checkMember($member, 'CMS_ACCESS_SecurityAdmin');
```

ADMIN permissions will return true for any security check, so don't need to check it explicitly.

security/Member.php
```diff
@@ -1470,25 +1470,30 @@ public function fieldLabels($includerelations = true) {
...
 		}
+		
+		//need to be logged in and/or
+		//most checks below rely on $member being a Member
+		if(!$member) return false;		
 
-		// members can usually edit their own record
-		if($member && $this->ID == $member->ID) return true;
+		// members can usually view their own record
+		if($this->ID == $member->ID) return true;
```
tractorcow: Please use braces on if conditions.

security/Member.php
```diff
@@ -1470,25 +1470,30 @@ public function fieldLabels($includerelations = true) {
...
+		}
+		//check for extensions, we do this first as they can overrule everything
+		$extended = $this->extendedCan(__FUNCTION__, $member);
+		if($extended !== null) {
+			return $extended;
 		}
+		
+		//need to be logged in and/or
+		//most checks below rely on $member being a Member
+		if(!$member) return false;
```
tractorcow: Please use braces on if conditions.

security/Member.php
```diff
@@ -1497,26 +1502,34 @@ public function canView($member = null) {
...
-		// with "higher" permissions than himself, but thats hard to determine.
+		//check for extensions, we do this first as they can overrule everything
+		$extended = $this->extendedCan(__FUNCTION__, $member);
+		if($extended !== null) {
+			return $extended;
+		}
+		
+		//need to be logged in and/or
+		//most checks below rely on $member being a Member
+		if(!$member) return false;
```
tractorcow: Please use braces on if conditions.
Comment can fit on one line. :)

security/Member.php
```diff
@@ -1497,26 +1502,34 @@ public function canView($member = null) {
...
+		//need to be logged in and/or
+		//most checks below rely on $member being a Member
+		if(!$member) return false;		
+		
+		// HACK: we should not allow for an non-Admin to edit an Admin
 		if(!Permission::checkMember($member, 'ADMIN') && Permission::checkMember($this, 'ADMIN')) return false;
 
-		return $this->canView($member);
+		// members can usually edit their own record
+		if($member && $this->ID == $member->ID) return true;
```
tractorcow: Please use braces on if conditions.
You don't need to check for `if($member)` since it's been done above.

security/Member.php
```diff
@@ -1497,26 +1502,34 @@ public function canView($member = null) {
...
+		if($extended !== null) {
+			return $extended;
+		}
+		
+		//need to be logged in and/or
+		//most checks below rely on $member being a Member
+		if(!$member) return false;		
+		
+		// HACK: we should not allow for an non-Admin to edit an Admin
 		if(!Permission::checkMember($member, 'ADMIN') && Permission::checkMember($this, 'ADMIN')) return false;
```
tractorcow: `HACK` is only code we are ashamed of... you should be proud of this code. :)

tractorcow: But yeah, `Please use braces on if conditions.` (I bet you're sick of hearing this from me) :P

security/Member.php
```diff
@@ -1497,26 +1502,34 @@ public function canView($member = null) {
...
 
-		return $this->canView($member);
+		// members can usually edit their own record
+		if($member && $this->ID == $member->ID) return true;
+
+		//standard check
+		if(
```
tractorcow: As above, can shorten to

```php
return Permission::checkMember($member, 'CMS_ACCESS_SecurityAdmin');
```

security/Member.php
```diff
@@ -1524,23 +1537,35 @@ public function canEdit($member = null) {
...
 
-		return $this->canEdit($member);
+		//standard check
+		if(
+			Permission::checkMember($member, 'ADMIN')
+			|| Permission::checkMember($member, 'CMS_ACCESS_SecurityAdmin')			
+		) {
+			return true;
+		}
+		return false;
```
tractorcow:
```php
return Permission::checkMember($member, 'CMS_ACCESS_SecurityAdmin');
```

tractorcow:
Since we are focusing on improving coding style for this PR, I thought of a couple of other ways we could tighten up the implementation. :) They are just ideas; I think your code is an awesome improvement, so don't feel like I'm putting it down. :D

sunnysideup:
thank you for all the feedback.  Much appreciated.  I really value it.  

Happy to make more fixes. Once this is done, we can also review the can* methods for some other classes (e.g. Group) as they could probably do with some clean up as well. 

MINOR DETAIL: I shout at people in the office when they write if statements without braces and my lazy mind copied a brace-less piece of code that was already there.  I was worried that I also had to apply the awful PSR-2 bracket style 
```php
    if ($foo == $bar) 
    {

    }
so I am glad you did not mention that ...

**Comment by sunnysideup** [comment]
I dont understand why travis stopped working now :-(

**Comment by tractorcow** [comment]
> MINOR DETAIL: I shout at people in the office when they write if statements without braces and my lazy mind copied a brace-less piece of code that was already there. I was worried that I also had to apply the awful PSR-2 bracket style

That's ok, I'm the same way. :) My policy is "if you touch it, you clean it", so I like to fit in fixing up old code whenever I'm working nearby.

I've restarted the tests; I think you might need to squash your commits though, since you have a few with the same commit message.

**Comment by tractorcow** [comment]
Code is absolutely pristine. Good job. :D

**Comment by sunnysideup** [comment]
I tried to merge the commits, but it did not seem work ;-(

**Comment by dhensby** [comment]
@sunnysideup the coding style PSRs don't require if statements have the curly brace on a new line (https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#11-example)

**Comment by dhensby** [comment]
@sunnysideup you don't want to merge your commits, you want to "squash" them.

I think you want to do something like this:

```sh
$ git reset --hard 209e2c3
$ git reset HEAD~6
$ git add -A
$ git commit --amend
$ git push -f
```

This will reset your history back to how it was before your merge, then roll back 6 commits (but keeping the changes in your working dir), then you add all the changes, amend the last commit (which is your first in this PR) and then you'll push it up into the branch and thus this PR.

dhensby:
@sunnysideup we do need tests to pass too :(

sunnysideup:
hi @dhensby , thank you for your feedback. 

I did: 
```sh
git rebase -i HEAD~6
```
which seemed to work fine.  I picked the first one and squashed the others, but does not seemed to have worked. One thing I do not know so well is if after a rebase you have to commit / push.

sunnysideup:
can we reschulet this one: https://scrutinizer-ci.com/g/silverstripe/silverstripe-framework/inspections/4c1c4768-6a8b-4623-af22-b3ae35fbc6cd?

dhensby:
@sunnysideup my instructions will still work and get the PR how it should be :)

sunnysideup:
I can the code as you suggested @dhensby , but that seemed to make more of a mess than anything. I may try this again next week from a clean start because it is getting a little frustrating.

sunnysideup:
I just post the code here for future reference:

```

	/**
	 * Users can view their own record.
	 * Otherwise they'll need ADMIN or CMS_ACCESS_SecurityAdmin permissions.
	 * This is likely to be customized for social sites etc. with a looser permission model.
	 */
	public function canView($member = null) {
		//get member
		if(!($member instanceof Member)) {
			$member = Member::currentUser();
		}
		//check for extensions, we do this first as they can overrule everything
		$extended = $this->extendedCan(__FUNCTION__, $member);
		if($extended !== null) {
			return $extended;
		}
		
		//need to be logged in and/or most checks below rely on $member being a Member
		if(!$member) {
			return false;		
		}
		// members can usually view their own record
		if($this->ID == $member->ID) {
			return true;
		}
		//standard check
		return Permission::checkMember($member, 'CMS_ACCESS_SecurityAdmin');
	}
	/**
	 * Users can edit their own record.
	 * Otherwise they'll need ADMIN or CMS_ACCESS_SecurityAdmin permissions
	 */
	public function canEdit($member = null) {
		//get member
		if(!($member instanceof Member)) {
			$member = Member::currentUser();
		}
		//check for extensions, we do this first as they can overrule everything
		$extended = $this->extendedCan(__FUNCTION__, $member);
		if($extended !== null) {
			return $extended;
		}
		
		//need to be logged in and/or most checks below rely on $member being a Member
		if(!$member) {
			return false;		
		}
		
		// HACK: we should not allow for an non-Admin to edit an Admin
		if(!Permission::checkMember($member, 'ADMIN') && Permission::checkMember($this, 'ADMIN')) {
			return false;
		}
		// members can usually edit their own record
		if($this->ID == $member->ID) {
			return true;
		}
		//standard check
		return Permission::checkMember($member, 'CMS_ACCESS_SecurityAdmin');
	}
	/**
	 * Users can edit their own record.
	 * Otherwise they'll need ADMIN or CMS_ACCESS_SecurityAdmin permissions
	 */
	public function canDelete($member = null) {
		if(!($member instanceof Member)) {
			$member = Member::currentUser();
		}
		//check for extensions, we do this first as they can overrule everything
		$extended = $this->extendedCan(__FUNCTION__, $member);
		if($extended !== null) {
			return $extended;
		}
		
		//need to be logged in and/or most checks below rely on $member being a Member
		if(!$member) {
			return false;		
		}
		// Members are not allowed to remove themselves,
		// since it would create inconsistencies in the admin UIs.
		if($this->ID && $member->ID == $this->ID) {
			return false;			
		}
		
		// HACK: if you want to delete a member, you have to be a member yourself.
		// this is a hack because what this should do is to stop a user
		// deleting a member who has more privileges (e.g. a non-Admin deleting an Admin)
		if(Permission::checkMember($this, 'ADMIN')) {
			if( ! Permission::checkMember($member, 'ADMIN')) {
				return false;				
			}
		}
		//standard check
		return Permission::checkMember($member, 'CMS_ACCESS_SecurityAdmin');
	}

```

tractorcow:
@sunnysideup at this point, you probably should be git rebasing against hard SHA values. :)

Open source is like this a lot isn't it....

Maybe you could make a clean branch and just cherry pick back if you get stuck.

dhensby:
@sunnysideup you're getting more than 8 commits when you do that because you're rebasing over a merge commit so it'll pull in the merged commits too.

You need to follow my instructions, but it looks like I got the commit hash wrong for the initial reset, it should be ` b2574c5` instead.

dhensby:
see #5136

PR #5121, opened 2016-03-01
FIX getIP from behind a load-balancer that adds many IPs to the header

dhensby:
Found when using behind a load balancer that the header can return more than one IP.

This will loop through the IPs that are extracted from the trusted header and pull out the first non-private IP that is returned.

### DIFF
--- a/control/HTTPRequest.php
+++ b/control/HTTPRequest.php
@@ -669,14 +669,36 @@ class SS_HTTPRequest implements ArrayAccess {
 		}
 
 		if ($headerOverrideIP) {
-			return $headerOverrideIP;
+			return $this->getIPFromHeaderValue($headerOverrideIP);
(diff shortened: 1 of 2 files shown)


control/HTTPRequest.php
```diff
@@ -669,14 +669,36 @@ public function getIP() {
...
 			return null;
 		}
 	}
-	
+
+	/**
+	 * Extract an IP address from a header value that has been obtained. Accepts single IP or comma separated string of
+	 * IPs
+	 *
+	 * @param $headerValue string The value from a trusted header
```
tractorcow: phpdoc is backwards. `@param string $headerValue`

dhensby: fixed - I was just checking you were paying attention

Issue #4938, opened 2016-01-18
RFC: FormField React Integration API


tractorcow:
https://groups.google.com/forum/#!topic/silverstripe-dev/bHE7sIK-fV8

# 1. Introduction

## 1.1. Problem

When implementing React based front-end components, there is limited ability for forms declared via PHP to influence the front end. Likewise, the development of custom components used in front-end React based forms would require re-implementation of that behaviour via the traditional SilverStripe Forms API.

Since a key benefit of SilverStripe is that it is easy to tailor the UI to fit your site’s custom data model, this isn’t ideal.

It results in a bottleneck in development, where React forms cannot be developed effectively by developers who do not have both a strong front and back-end expertise, as well as the patience and time to solve several integration issues. It also means that client and server code is tightly coupled. Creating a new form field UI shouldn't necessitate creating a PHP subclass. And creating a new form field in PHP shouldn't require developers to create a new React component. The outcomes of this are slower development time, less flexibility, and buggy applications.

## 1.2. Goals

The goal of this RFC is to present a solution that addresses:

* Ability for front-end developers to build custom React form components with minimal back-end development work.

* Ability for back-end developers to build custom forms with no front-end React development work.

* Ability for developers to implement custom back end form components which may rely on custom front-end components.

* Define a standard schema for both the declaration, and error processing of form validation rules.

* Ensure that any FormField components are available to custom extensions on DataObjects, or subclasses of Forms.

* The solution must be usable within both the CMS and public facing site.

* The standard form API must work without react enabled, even if components are available

* Standard form security rules (CSRF / XSS protection, redirection validation, input validation, etc) must be enforced.

* Maintain separation of schema (structure) from content (current state).

* Existing mechanisms (e.g. form postback) should be used where possible.

## 1.3. Proposed Solution

We propose that we create an intermediary data schema, in order to map an abstract representation of the form back-end, to the front-end rendering mechanism. This schema will be JSON structured data which can be extracted both from Form objects, as well as individual FormField instances.

[...]

PR #5132, opened 2016-03-02
Form schema

chillu:
Implements Form Schema RFC: https://github.com/silverstripe/silverstripe-framework/issues/4938.
Relies on https://github.com/silverstripe/silverstripe-cms/pull/1410

The CMS integration isn't actually used yet, since no client-side code is sending the required `X-Formschema-Request` header. This will soon be the case once we make use of the schema in `AssetAdmin`, still in feature branches:
https://github.com/open-sausages/silverstripe-framework/tree/integrate-react
https://github.com/open-sausages/silverstripe-cms/tree/integrate-react
https://github.com/open-sausages/silverstripe-asset-admin

The easiest way to test its operation for now is using `curl`:
- Perform a CMS action
- Right click on the ajax request in the "network" dev tools tab, "copy as curl"
- Paste in terminal, add the following: `-H 'X-Formschema-Request: schema,state'`

Tests to perform:
- Save a folder in admin/assets
- `GET admin/assets/schema/EditForm`
- Save a folder with a validation error in admin/assets. This requires the following addition to `Folder.php`:

```
public function getCMSValidator() {
	return new RequiredFields('Name');
}
```

I'm not super happy with the solution:
- Too much copypaste with `setValidationResponseCallback()`
- Adds more `FormField` public API surface. Discussed with @sminnee and @tractorcow and decided to at least mark the API as experimental. 
- Messy CMS response negotiation. We have to make this work in parallel to PJAX responses for now, since only admin/assets will be React powered in 4.x. 

Long term, I want to refactor `Form` and `FormField` to become pure value objects, with form handling and validation logic being moved to the controller. Presentation into different formats (HTML, JSON, etc) could then also be handled in the controller layer, or in a component which is decoupled from `Form`. This would make `FormSchema` use a lot cleaner in regards to CMS controllers. But it requires changes as deep as `DataObject->getCMSFields()`, so wasn't really feasible in the context of this PR.


admin/code/LeftAndMain.php
```diff
@@ -1260,14 +1353,27 @@ public function getEditForm($id = null, $fields = null) {
...
+			$negotiator = $this->getResponseNegotiator();
+			$form = Form::create(
 				$this, "EditForm", $fields, $actions
 			)->setHTMLID('Form_EditForm');
-			$form->setResponseNegotiator($this->getResponseNegotiator());
 			$form->addExtraClass('cms-edit-form');
 			$form->loadDataFrom($record);
 			$form->setTemplate($this->getTemplatesWithSuffix('_EditForm'));
 			$form->setAttribute('data-pjax-fragment', 'CurrentForm');
+			$form->setValidationResponseCallback(function() use ($negotiator, $form) {
```
tractorcow: Won't you need to set this callback on the other leftandmain subclasses too? E.g. SecurityAdmin swaps out CMSForm for Form, but doesn't have this code. Also ModelAdmin.

forms/FormField.php
```diff
@@ -162,7 +164,30 @@ class FormField extends RequestHandler {
...
 	 */
-	protected $attributes = array();
+	protected $attributes = [];
+
+	/**
+	 * The type of front-end component to render the FormField as.
+	 *
+	 * @var string
+	 */
+	protected $schemaComponent;
```
tractorcow: Why aren't all these $schema* properties in the trait?

tractorcow:
Tests stalled; I've restarted them.

admin/code/LeftAndMain.php
chillu: `SecurityAdmin` uses `parent::getEditForm()` which gets the callback from `LeftAndMain`.
`ModelAdmin` uses `GridFieldDetailForm`, which in turn still uses its own PJAX logic (in `generateValidationResponse()`), I've confirmed a ModelAdmin can still display validation errors.
Once these interfaces are switched over to React, requiring JSON responses, we can look into migrating the new logic there.

forms/FormField.php
chillu: Fixed now

tractorcow:
@chillu still need that branch?

chillu:
Nope, I've removed the branch. David will rebase on master for his JS-based form schema work

PR #5135, opened 2016-03-03
Fix TreeDropdownField auto-close regression

chillu:
Causes build failures due to Behat not being able to select
a link from a dropdown in "Insert Media".

Follow on effect from triggering a window.resize event
when the window wasn't actually resized (in order to force a layout redraw).

See https://github.com/silverstripe/silverstripe-framework/pull/5087 for context

Needs to be merged up to 3.3, 3 and master - since it currently breaks all those builds.


scott1702:
Just tested this out and it fixes the original issue :+1:

PR #5141, opened 2016-03-04
Info on enabling i18n in SS templates

sebastiand:
Added description on how to activate i18n for SS templates, so that <%t %> can be used in templates.
Provided code example for Page_Controller::init() function.
Without this information no one will be able to use the <%t %> function in templates


dhensby:
Shouldn't this be in the translatable docs, not framework?

tractorcow:
I think it's fine to have i18n bootstrapping code here, but I'm hesitant to have it pull from translatable, as it's not a core module.

How about changing it to some other data source. Intl has some good examples.

Try http://php.net/manual/en/locale.acceptfromhttp.php

sebastiand:
You are right. As it concerns a third party module this obviously isn´t the right place.
This already is in the translatable docs: https://github.com/silverstripe/silverstripe-translatable/blob/master/docs/en/index.md#setting-the-i18n-locale

PR #5142, opened 2016-03-06
Include CMS smoke tests in Travis builds

chillu:
We've had a few failures where framework caused regressions in CMS,
so these builds are helpful. They'll increase the overall build
times on the "silverstripe" user because of Travis' build limitations.

The parallel per-build run times shouldn't increase, since
framework builds take longer than cms builds anyway:
CMS Behat build took 13:53 on last 3.3 run,
framework MySQL PDO build took 16:12.

Talked to @tractorcow and he recommended to trial this on the 3.3 branch for a bit, since its the most active one. If it helps more than it gets in the way, we can backport to 3.2 and upwards to 3 and master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,10 @@ matrix:
       env: DB=MYSQL PDO=1
     - php: 5.6
       env: DB=MYSQL BEHAT_TEST=1
+    - php: 5.6
+      env: DB=MYSQL CMS_TEST=1
+    - php: 5.6
+      env: DB=MYSQL BEHAT_TEST=1 CMS_TEST=1
 
 before_script:
 # Until http://pecl.php.net/package/imagick is working again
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -37,13 +41,16 @@ before_script:
  - "if [ \"$BEHAT_TEST\" = \"\" ]; then php ~/travis-support/travis_setup.php --source `pwd` --target ~/builds/ss; fi"
  - "if [ \"$BEHAT_TEST\" = \"1\" ]; then php ~/travis-support/travis_setup.php --source `pwd` --target ~/builds/ss --require silverstripe/behat-extension; fi"
  - cd ~/builds/ss
+ - "if [ \"$CMS_TEST\" = \"1\" ]; then composer require --prefer-dist silverstripe/cms:$CORE_RELEASE.x-dev"
```
tractorcow: `master-x.dev` :D just a note we'll need to change this on master, heh.

Does self.version work here?

tractorcow: Turns out this is invalid bash syntax anyway. I think you need to put that arg into some `\" \"` to escape properly?

chillu: Yeah I was aware of the `master-x.dev` issue, came to the same conclusion as you - we'd need to manually fix that. Didn't want to add it to `--require` in the bash lines above it, because it would always check out CMS with all its side effects.

.travis.yml
```diff
@@ -37,13 +41,16 @@ before_script:
  - "if [ \"$BEHAT_TEST\" = \"\" ]; then php ~/travis-support/travis_setup.php --source `pwd` --target ~/builds/ss; fi"
  - "if [ \"$BEHAT_TEST\" = \"1\" ]; then php ~/travis-support/travis_setup.php --source `pwd` --target ~/builds/ss --require silverstripe/behat-extension; fi"
  - cd ~/builds/ss
+ - "if [ \"$CMS_TEST\" = \"1\" ]; then composer require --prefer-dist \"silverstripe/cms:$CORE_RELEASE.x-dev\""
```
tractorcow: Still need the trailing `; fi`

dhensby:
@chillu it's broken

.travis.yml
```diff
@@ -37,13 +41,16 @@ before_script:
  - "if [ \"$BEHAT_TEST\" = \"\" ]; then php ~/travis-support/travis_setup.php --source `pwd` --target ~/builds/ss; fi"
  - "if [ \"$BEHAT_TEST\" = \"1\" ]; then php ~/travis-support/travis_setup.php --source `pwd` --target ~/builds/ss --require silverstripe/behat-extension; fi"
  - cd ~/builds/ss
+ - "if [ \"$CMS_TEST\" = \"1\" ]; then composer require --prefer-dist \"silverstripe/cms:$CORE_RELEASE.x-dev\"; fi"
```
tractorcow: If CMS_TEST = 1, it's too late to install here. You need to include silverstripe/cms at the `~/travis-support/travis_setup.php` step, since that script removes the Page.php if it detects CMS isn't installed.

tractorcow: Maybe the `--require` argument should be its own environment value.

chillu:
@dhensby @tractorcow Implemented the suggestions, and build is passing now. Haven't switched `--require` to an environment var, I don't think that would decrease the complexity much - you'd still have four different statements for the actual test execution anyway, at least there's a matching amount now ;)

dhensby:
happy @tractorcow ?

tractorcow:
CORE_RELEASE is fixed for each branch, so as long as we remember to change it to `master` when merging up, we should be ok. Still prefer a better solution to the branch naming, but that can be solved another day.

PR #5143, opened 2016-03-06
BUG Better filtering of versionable tables during SQL augmentation

tractorcow:
Fixes test failures in https://github.com/silverstripe/silverstripe-cms/pull/1401


model/versioning/Versioned.php
```diff
@@ -264,7 +264,7 @@ public function augmentSQL(SQLSelect $query, DataQuery $dataQuery = null) {
 		case 'archive':
 			$date = $dataQuery->getQueryParam('Versioned.date');
 			foreach($query->getFrom() as $table => $dummy) {
-				if(!DB::get_schema()->hasTable($table . '_versions')) {
```
tractorcow: Yay fewer queries.

PR #5145, opened 2016-03-07
BUG Fix FulltextsearchEnable

### LINKED ISSUE #5126: FulltextSearchable::enable() causes build error (3.3.1)
It seems the DB engine doesn't actually get updated to MyISAM, resulting in a "The used table type doesn't support FULLTEXT indexes" build-error when applying the FulltextSearchable extension in FulltextSearchable::enable().

Both MySQLDatabase and MySQLPDODatabase have this issue. 
Manually setting the engine to MyISAM seems to resolve it but this shouldn't be required as the enable method should update the tables to use the MyISAM table.

tractorcow:
Fixes #5126

It looks like there were a few `get_class($this)` left over in the old ORM connectors that needed refactoring out.

### DIFF
--- a/model/connect/DBSchemaManager.php
+++ b/model/connect/DBSchemaManager.php
@@ -310,15 +310,14 @@ abstract class DBSchemaManager {
 
 			// Check if options changed
 			$tableOptionsChanged = false;
-			if (isset($options[get_class($this)]) || true) {
-				if (isset($options[get_class($this)])) {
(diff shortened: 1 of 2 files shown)


tractorcow:
Note to self: Close https://silverstripe.atlassian.net/browse/CWP-536 once merged

tractorcow:
Ah, thanks!

Issue #5147, opened 2016-03-07
3.3.1: filesystem/FolderTest fails in tearDown trying to delete file


robbieaverill:
I've installed SS framework 3.3.1 with the CMS addon, have configured my environment and have run the PHPUnit test suite:

**PHPUnit:** 3.7.38 (via composer)

```
Time: 10.19 minutes, Memory: 173.75Mb

There was 1 error:

1) FolderTest::testDeleteAlsoRemovesContainedFilesInDatabaseAndFilesystem
unlink(/var/www/assets/FileTest.GIF): No such file or directory

/var/www/framework/tests/filesystem/FolderTest.php:285
/var/www/vendor/phpunit/phpunit/phpunit.php:46

FAILURES!
Tests: 1853, Assertions: 7719, Errors: 1, Incomplete: 7, Skipped: 21.
```

Seeing as the code that triggers this error is part of a tear down method:

```php
	public function tearDown() {
			// ...
			if($file && file_exists(BASE_PATH."/$file->Filename")) unlink(BASE_PATH."/$file->Filename");
```

...I see no reason to silence any errors coming from the `unlink` method.

I'm not sure why it would determine a file to have existed, then change its mind when executing the logic in the if statement...

Anyway, with this (`@unlink`) in place the suite completes successfully (albeit with some skipped and incomplete):

```
Time: 10.3 minutes, Memory: 173.75Mb

OK, but incomplete or skipped tests!
Tests: 1853, Assertions: 7719, Incomplete: 7, Skipped: 21.
```

robbieaverill:
Is vfsStream not used for filesystem tests?

tractorcow:
No, but we have considered it. I'm yet to be convinced another layer of abstraction is necessary, as we are already using flysystem on master.

robbieaverill:
Closed for now.

PR #5148, opened 2016-03-07
[#5147] Silence tearDown unlink errors

robbieaverill:
Addresses #5147

### DIFF
--- a/tests/filesystem/FolderTest.php
+++ b/tests/filesystem/FolderTest.php
@@ -272,7 +272,7 @@ class FolderTest extends SapphireTest {
 		$fileIDs = $this->allFixtureIDs('File');
 		foreach($fileIDs as $fileID) {
 			$file = DataObject::get_by_id('File', $fileID);
-			if($file && file_exists(BASE_PATH."/$file->Filename")) unlink(BASE_PATH."/$file->Filename");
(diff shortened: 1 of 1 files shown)


tractorcow:
This seems to be hiding a potentially real error; If you have a permission problem on `/assets` and your tests aren't able to delete files between tests, it's going to affect test results.

Are you by any chance running tests on windows?

robbieaverill:
Hey @tractorcow 

I'm running this in a docker container on OSX El Capitan via dinghy.

It seems odd to me that it can detect that the file exists then not be able to remove it.

Incidentally, I've since been running this through CircleCI and the error has stopped being reported... Perhaps close as cannot reproduce.

Cheers,
Robbie

tractorcow:
I recall often needing to chmod my assets folder on composer install... what were your permissions?

robbieaverill:
Hi @tractorcow 

The original permission was 755. I'm having a little trouble running the suite now after adding the blog module, but will test, chmod to 777 then re-test again and will close this if there's no difference.

Cheers,
Robbie

tractorcow:
Hm, what else could it be... umask maybe?

robbieaverill:
Not sure. Happy to close for now

PR #5151, opened 2016-03-07
FIX Singleton creation

### LINKED ISSUE #4878: populateDefaults not working on extensions
It seems, extending SiteConfig with a file and having a populate defaults breaks the build because SiteConfig is setup before File?

jonom:
Ensure DataObject instances are aware they are singletons so functions like populateDefaults() can be skipped. (fixes #4878)

### DIFF
--- a/control/injector/Injector.php
+++ b/control/injector/Injector.php
@@ -549,6 +549,14 @@ class Injector {
 			$constructorParams = $spec['constructor'];
 		}
 
+		// if this is a DataObject sub-class and we want a singleton set second arg $isSingleton to true
+		if ($type != 'prototype' && is_subclass_of($class, 'DataObject')) {
+			if (count($constructorParams) < 1) {
+				$constructorParams[0] = null;
+			}
+			$constructorParams[1] = true;
+		}
+
 		$factory = isset($spec['factory']) ? $this->get($spec['factory']) : $this->getObjectCreator();
 		$object = $factory->create($class, $constructorParams);
 


jonom:
@tractorcow one of the builds failed during setup, I'm guessing re-running test won't help here?

tractorcow:
Hm, the logic looks sound... but maybe @chillu should look over this too.

Perhaps in 4.x we should override Object::singleton() in DataObject, in order to inject the new parameter, using the 3rd argument to `Injector::get()`. That way all of the DataObject specific stuff stays in code.

@chillu is this specific fix ok for 3.x?

tractorcow:
Don't worry about that; It's the fault of imagick no longer supporting php 5.3

jonom:
> Perhaps in 4.x we should override Object::singleton() in DataObject, in order to inject the new parameter, using the 3rd argument to Injector::get()

That might work but some refactoring would be required first. `Object::singleton()` isn't called at all during build currently, instead the global `singleton($className)` function as defined in Core.php is used. I'm guessing there's a good reason for this but outside my expertise :)

BTW the manipulation I'm doing of constructor args could probably be done in `get()` instead of `instantiate()`, not sure if either one is better?

tractorcow:
Yeah, that's why I hate global functions. Itcan use $class::singleton() fine though.;)

jonom:
So the question is then, is there a reason we need a global singleton method?

Unless we can remove it (or effectively remove it by passing through calls to `$className::singleton()`) we're going to get inconsistent results between `singleton($className)` and `$className::singleton()` if we take the `DataObject::singleton()` path, which I don't think would be acceptable would it?

Let's give it a try... https://github.com/silverstripe/silverstripe-framework/pull/5154 (I've really got to get tests working locally)

tractorcow:
> So the question is then, is there a reason we need a global singleton method?

Not anymore, really. Injector::inst()->get() should be used for non-object subclasses, and anything that extends Object (or trait Injectable in master) should just use ::singleton().

jonom:
Closing this for now pending outcome of https://github.com/silverstripe/silverstripe-framework/pull/5154

jonom:
New version at https://github.com/silverstripe/silverstripe-framework/pull/7850

PR #5152, opened 2016-03-07
FIX file sync removes folders with dot in name

jonom:
/cc @tractorcow

### DIFF
--- a/filesystem/Folder.php
+++ b/filesystem/Folder.php
@@ -165,6 +165,7 @@ class Folder extends File {
 
 				// Check allowed extensions, unless admin users are allowed to bypass these exclusions
 				if($checkExtensions
+					&& !is_dir($baseDir . $actualChild)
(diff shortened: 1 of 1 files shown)


filesystem/Folder.php
```diff
@@ -165,6 +165,7 @@ public function syncChildren() {
 
 				// Check allowed extensions, unless admin users are allowed to bypass these exclusions
 				if($checkExtensions
+					&& !is_dir("$baseDir/$actualChild")
```
tractorcow: On line 199 it does `$baseDir . $actualChild` to concatenate these. Does does $baseDir have a trailing slash?

jonom: Yes looks like it always does after some quick testing. Have updated.

tractorcow:
I'm going to forgive the test failure; It's due to imagick no longer being supported on 5.3.

jonom:
Cheers

PR #5153, opened 2016-03-07
BUG Don't install imagick on php 5.3

tractorcow:


tractorcow:
If travis doesn't show up, that means the .travis.yml is invalid;; Don't trust green ticks!

chillu:
So that means image magick tests won’t on 5.3? Seems acceptable.

tractorcow:
Right, it'll just use GD, because imagick won't install on php 5.3 anymore.

PR #5154, opened 2016-03-07
FIX Singleton creation

jonom:
@tractorcow if tests pass this should be a better fix than https://github.com/silverstripe/silverstripe-framework/pull/5151

...Friggin' white space...

### DIFF
--- a/core/Core.php
+++ b/core/Core.php
@@ -154,7 +154,9 @@ function singleton($className) {
 	if(!isset($className)) user_error("singleton() Called without a class", E_USER_ERROR);
 	if(!is_string($className)) user_error("singleton() passed bad class_name: " . var_export($className,true),
 		E_USER_ERROR);
-	return Injector::inst()->get($className);
+	return ($className == 'Object' || is_subclass_of($className, 'Object'))
+		? $className::singleton()
+		: Injector::inst()->get($className);
 }
 
 function project() {
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -423,6 +423,20 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	}
 
 	/**
+	 * Creates a class instance by the "singleton" design pattern.
+	 * It will always return the same instance for this class,
+	 * which can be used for performance reasons and as a simple
+	 * way to access instance methods which don't rely on instance
+	 * data (e.g. the custom SilverStripe static handling).
+	 *
+	 * @return static The singleton instance
+	 */
+	public static function singleton() {
+		$class = get_called_class();
+		return Injector::inst()->get($class, true, array(null, true));
+	}
+
+	/**
 	 * Construct a new DataObject.
 	 *
 	 * @param array|null $record This will be null for a new database record.  Alternatively, you can pass an array of
@@ -1078,7 +1092,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 
 	/**
 	 * Public accessor for {@see DataObject::validate()}
-	 * 
+	 *
 	 * @return ValidationResult
 	 */
 	public function doValidate() {
@@ -1606,7 +1620,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 		}
 
 		if($filter !== null || $sort !== null || $limit !== null) {
-			Deprecation::notice('4.0', 'The $filter, $sort and $limit parameters for DataObject::getComponents() 
+			Deprecation::notice('4.0', 'The $filter, $sort and $limit parameters for DataObject::getComponents()
 				have been deprecated. Please manipluate the returned list directly.', Deprecation::SCOPE_GLOBAL);
 		}
 
@@ -1694,7 +1708,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 			$remoteClass = $this->hasManyComponent($component, false);
 		} else {
 			$remoteClass = $this->belongsToComponent($component, false);
-		}		
+		}
 
 		if(empty($remoteClass)) {
 			throw new Exception("Unknown $type component '$component' on class '$this->class'");
@@ -1769,8 +1783,8 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 			= $this->manyManyComponent($componentName);
 
 		if($filter !== null || $sort !== null || $join !== null || $limit !== null) {
-			Deprecation::notice('4.0', 'The $filter, $sort, $join and $limit parameters for 
-				DataObject::getManyManyComponents() have been deprecated. 
+			Deprecation::notice('4.0', 'The $filter, $sort, $join and $limit parameters for
+				DataObject::getManyManyComponents() have been deprecated.
 				Please manipluate the returned list directly.', Deprecation::SCOPE_GLOBAL);
 		}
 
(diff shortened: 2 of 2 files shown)


model/DataObject.php
```diff
@@ -423,6 +423,20 @@ private static function cache_composite_fields($class) {
...
 	/**
+	 * Creates a class instance by the "singleton" design pattern.
+	 * It will always return the same instance for this class,
+	 * which can be used for performance reasons and as a simple
+	 * way to access instance methods which don't rely on instance
+	 * data (e.g. the custom SilverStripe static handling).
+	 *
+	 * @return static The singleton instance
+	 */
+	public static function singleton() {
```
tractorcow: I would like this in 3.x, but I'm afraid it wouldn't be semver. XD

The current `singleton()` method on Object is only on the master branch, I'm afraid.

tractorcow:
Excellent, but a couple of things:

- We should probably replace usages of singleton($class) with $class::singleton(), rather than modifying the internals of singleton(), but only IF we know the $class in an Object subclass (some aren't).
- Master branch only, since this is new API. Maybe we could back-port the method to 3 branch, but you'd need to have it on Object base class as well as DataObject.

Your other fix is probably fine for 3.x

model/DataObject.php
jonom: But https://github.com/silverstripe/silverstripe-framework/blob/3.2/core/Object.php#L155-L164

jonom:
I think this is SemVer compliant personally :grinning: Object::singleton() exists on Object in 3.2 thus it exists on DataObject as well.

> We should probably replace usages of singleton($class) with $class::singleton()

I agree, but then it wouldn't be SemVer.

jonom:
Well actually, replacing all those instances would be fine, but we couldn't remove the global function, just deprecate it

Excellent:
I think I’ve been added to this email chain by accident. 

Good luck on - this!

jonom:
@tractorcow if you agree this can go in to 3.2 as a bugfix I'd suggest the only additional change I make here is marking the global singleton function as deprecated @ 4. Then once this has been merged up to master we can remove the global singleton function and replace all the calls to it as per your suggestion in a subsequent PR. Reason for that is there are around 68 files in framework that use the global `singleton()` function and 18 in CMS so might cause an unnecessary merge headache if we try to change this in an older version, with no real benefit?

jonom:
Heh, Excellent.

Excellent:
I think I’ve been added to this email chain by accident. Could you remove me please? 
Where did you get this email from - it’s my spam email?

tractorcow:
Ugh, I'm sorry... 

It's because I added your github name after the `@` when I meant to write the literal `Excellent`.

https://github.com/silverstripe/silverstripe-framework/pull/5154#issuecomment-193487222

I've removed it... but you might stay a part of this conversation. You may need to click the "Unsubscribe" button and it will remove you. :D

tractorcow:
Just smash this thing (or the link in your email)

image

Excellent:
thanks - sorry it was just I was getting 100 emails a minute - no problem, christ I must have had this account for years if it's using this email! Good luck.

tractorcow:
At least you didn't call your account name `todo`.

tractorcow:
> I think this is SemVer compliant personally :grinning: Object::singleton() exists on Object in 3.2 thus it exists on DataObject as well.

For some reason I thought this was on master only... right you are!

@jonom I'm happy to deprecate the global method, but we'll need to solve the issue with this change breaking `singleton()` calls on non-object subclasses.

model/DataObject.php
tractorcow: My mistake, that's what I get for trusting my memory.

I did commit this to master, but this was back when master WAS 3.2 :D

core/Core.php
```diff
@@ -154,7 +154,7 @@ function singleton($className) {
 	if(!isset($className)) user_error("singleton() Called without a class", E_USER_ERROR);
 	if(!is_string($className)) user_error("singleton() passed bad class_name: " . var_export($className,true),
 		E_USER_ERROR);
-	return Injector::inst()->get($className);
+	return $className::singleton();
```
tractorcow: Will break if $className isn't Object subclass. :)

jonom: Sure but can you give an example of when this would happen?

jonom:
> I'm happy to deprecate the global method, but we'll need to solve the issue with this change breaking singleton() calls on non-object subclasses.

I've updated the global method to restore the original behaviour for non-object subclasses but if this is a valid concern, I guess we can't really remove the global function? Unless in those cases we want to replace `singleton($class)` with `Injector::inst()->get($class)` which I suppose is okay. Can you give an example of when you would need a singleton for something that isn't a subclass of object?

jonom:
Tests exploded

tractorcow:
Restarted...

tractorcow:
I know you'll hate me for this, but I just noticed you didn't actually add a test. :)

can you test that `singleton("SomeDataObjectClass")` correctly omits the populate defaults? That's what you're testing right?

tractorcow:
@jonom also I'm impressed with your commitment to this. :) It must be a pain listening to my critique on end lol...

tractorcow:
```
public function testSingleton() {
	$singleton = singleton('MyObject');
	$inst = MyObject::create();
	$this->assertEquals('Default Title', $inst->Title);
	$this->assertEquals('Default Name', $inst->Name);
	$this->assertEmpty($singleton->Title);
	$this->assertEmpty($singleton->Name);
}

class MyObject extends DataObject implements TestOnly {
	private static $db = array(
		"Title" => "Text",
		"Name" => "Text",
	);
	private static $defaults = array(
		"Title" => "Default Title",
		"Name" => "Default Name"
	);
}
```

jonom:
Thanks for the free test! Have worked in something similar, just running locally but having some issues... if I do DataObjectTest_Fixture::create() the defaults are populated but if I do DataObjectTest_Fixture::create() they are not... is that because of something I did?

tractorcow:
> `DataObjectTest_Fixture::create()` 

You mean doing the same thing exactly twice?

jonom:
Updated the comment, made a mistake :)

tractorcow:
::create() should trigger new defaults.... try testing this without your change (just the current 3.2 branch head) to test your assumption. :D

tractorcow:
How about `DataObjectTest_Fixture::singleton()` ?

jonom:
That's what I thought to do but after changing branches PHPUnit won't run test... eerrgggh

jonom:
Yep looks like I've accidentally made sure `populateDefaults()` never works :) I'll close this for now and work on it more later.

tractorcow:
Ah, so close!

jonom:
Have pushed up the latest if you're interested... https://github.com/jonom/silverstripe-framework/commit/4332f3502fff94edf467c907271e6b09d3c26282#diff-79a102910bcc173b82e657e1926cfcc4R39 

I'm momentarily stumped as to why this would have spread beyond singleton instances but throughout the Injector singleton support is pretty choppy - in many cases it's just assumed you want a singleton if you haven't specified a type so it's hard to know exactly what you're going to get.

I assumed everything was okay with this PR because tests passed but all the fixtures in DataModelTest use `new` instead of `create()` so I guess that's why this didn't get caught.

@chillu did you write the Injector? Don't suppose you have an idea what's going on here?

tractorcow:
https://github.com/jonom/silverstripe-framework/commit/4332f3502fff94edf467c907271e6b09d3c26282#commitcomment-16575894

JamT04D:
Hi there,

Any progress with this topic?

Thanks

tractorcow:
Given that Object class no longer exists, I'm afraid there isn't going to be any more progress on this unless someone else picks it up for the master branch.

JamT04D:
OK thank you for the feedback.

jonom:
I ran in to this problem yet again, so opened a new PR to try to finally get this fixed in 3.x. https://github.com/silverstripe/silverstripe-framework/pull/7850

PR #5155, opened 2016-03-08
Don't delete all assets when uploading a file in the root

mikenz:
Catch "." dirnames in more places. truncateDirectory('.') will delete every asset. That's an undesirable thing.
findVariants() wouldn't return any variants if the dirname in "."


tractorcow:
Thanks for this!

dhensby:
should this not go into 3?

tractorcow:
It'll need to be a different PR... since the asset system is completely different in 3. :)

mikenz:
It was only recently introduced to master so I don't think it'll be affecting 3. But we don't use 3 here so I haven't tested it.

dhensby:
ah, ok - thanks for clearing that up!

PR #5157, opened 2016-03-08
API Owned objects are now automatically published

tractorcow:
Versioned extension has had a cleanup and refactor in order to make it bearable to look at for more than a few seconds.

This is the first real-world use case for owned versioned dataobjects; Now objects will trigger publishing on owned objects when they are published to live.

Some parts of File/Folder needed to be changed to cope with partial publishing of the folder tree.

CMS PR at https://github.com/silverstripe/silverstripe-cms/pull/1415

### DIFF
--- a/filesystem/File.php
+++ b/filesystem/File.php
@@ -551,7 +551,7 @@ class File extends DataObject implements ShortcodeHandler, AssetContainer {
 	 * This method will update the File {@see DBFile} field value on success, so it must be called
 	 * before writing to the database
 	 *
-	 * @param bool True if changed
+	 * @return bool True if changed
 	 */
 	public function updateFilesystem() {
 		if(!$this->config()->update_filesystem) {
@@ -563,24 +563,22 @@ class File extends DataObject implements ShortcodeHandler, AssetContainer {
 			return false;
 		}
 
+		// Avoid moving files on live; Rely on this being done on stage prior to publish.
+		if(Versioned::get_stage() !== Versioned::DRAFT) {
+			return false;
+		}
+
 		// Check path updated record will point to
 		// If no changes necessary, skip
 		$pathBefore = $this->File->getFilename();
-		$pathAfter = $this->getFilename();
+		$pathAfter = $this->generateFilename();
 		if($pathAfter === $pathBefore) {
 			return false;
 		}
 
 		// Copy record to new location via stream
 		$stream = $this->File->getStream();
-		$result = $this->File->setFromStream($stream, $pathAfter);
-
-		// If the backend chose a new name, update the local record
-		if($result['Filename'] !== $pathAfter) {
-			// Correct saved folder to selected filename
-			$pathAfter = $result['Filename'];
-			$this->setFilename($pathAfter);
-		}
+		$this->File->setFromStream($stream, $pathAfter);
 		return true;
 	}
 
@@ -588,8 +586,10 @@ class File extends DataObject implements ShortcodeHandler, AssetContainer {
 	 * Collate selected descendants of this page.
 	 * $condition will be evaluated on each descendant, and if it is succeeds, that item will be added
 	 * to the $collator array.
-	 * @param condition The PHP condition to be evaluated.  The page will be called $item
-	 * @param collator An array, passed by reference, to collect all of the matching descendants.
+	 *
+	 * @param string $condition The PHP condition to be evaluated.  The page will be called $item
+	 * @param array $collator An array, passed by reference, to collect all of the matching descendants.
+	 * @return true|null
 	 */
 	public function collateDescendants($condition, &$collator) {
 		if($children = $this->Children()) {
@@ -610,7 +610,8 @@ class File extends DataObject implements ShortcodeHandler, AssetContainer {
 	 *
 	 * Does not change the filesystem itself, please use {@link write()} for this.
 	 *
-	 * @param String $name
+	 * @param string $name
+	 * @return $this
 	 */
 	public function setName($name) {
 		$oldName = $this->Name;
@@ -655,7 +656,7 @@ class File extends DataObject implements ShortcodeHandler, AssetContainer {
 			$this->Title = str_replace(array('-','_'),' ', preg_replace('/\.[^.]+$/', '', $name));
 		}
 
-		return $name;
+		return $this;
 	}
 
 	/**
@@ -704,13 +705,18 @@ class File extends DataObject implements ShortcodeHandler, AssetContainer {
 		return Director::absoluteBaseURL()."admin/assets/removefile/".$this->ID;
 	}
 
-	public function getFilename() {
+	/**
+	 * Get expected value of Filename tuple value. Will be used to trigger
+	 * a file move on draft stage.
+	 *
+	 * @return string
+	 */
+	public function generateFilename() {
 		// Check if this file is nested within a folder
 		$parent = $this->Parent();
 		if($parent && $parent->exists()) {
 			return $this->join_paths($parent->getFilename(), $this->Name);
 		}
-
 		return $this->Name;
 	}
 
@@ -1025,6 +1031,10 @@ class File extends DataObject implements ShortcodeHandler, AssetContainer {
 
 	public function getIsImage() {
 		return false;
+	}
+
+	public function getFilename() {
+		return $this->File->Filename;
 	}
 
 	public function getHash() {
--- a/filesystem/FileMigrationHelper.php
+++ b/filesystem/FileMigrationHelper.php
@@ -34,8 +34,8 @@ class FileMigrationHelper extends Object {
 
 		// Loop over all files
 		$count = 0;
-		$originalState = \Versioned::get_reading_mode();
-		\Versioned::reading_stage('Stage');
+		$originalState = Versioned::get_reading_mode();
+		Versioned::set_stage(Versioned::DRAFT);
 		$filenameMap = $this->getFilenameArray();
 		foreach($this->getFileQuery() as $file) {
 			// Get the name of the file to import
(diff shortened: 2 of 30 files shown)


model/versioning/Versioned.php
```diff
@@ -1556,16 +1647,7 @@ public static function set_reading_mode($mode) {
...
 	public static function get_reading_mode() {
-		return Versioned::$reading_mode;
-	}
-
-	/**
-	 * Get the name of the 'live' stage.
-	 *
-	 * @return string
-	 */
-	public static function get_live_stage() {
```
kinglozzer: Shouldn’t we deprecate this? I know we _can_ just remove it in a new major version, but it’s probably kinder to raise a dep notice for now. Saves devs having to refer back to the upgrade docs when they get a fatal error for a method not existing. Same for `current_stage()` and `reading_stage()`

kinglozzer:
I’ve been through this and apart from the minor deprecation comments I don’t have any concerns, so +1 from me. Nice work!

model/versioning/Versioned.php
tractorcow: OK :D

tractorcow: On second thoughts, we can delete it, so let's.

If we deprecate it, we should port it back to 3 branch and maybe raise the deprecation notice there.

filesystem/Folder.php
```diff
@@ -258,11 +258,24 @@ public function getAbsoluteURL() {
...
+		// Update draft version of all child records
 		$this->updateChildFilesystem();
 	}
 
+	public function onAfterDelete() {
+		parent::onAfterDelete();
+
+		// Cascade deletions to live
+		if(Versioned::get_stage() === Versioned::DRAFT_STAGE) {
+			$this->deleteFromStage(Versioned::LIVE_STAGE);
```
dhensby: why delete from stage if the current stage is draft?

model/versioning/Versioned.php
dhensby: Add it to the upgrade notes - I think we'll probably need to make a 3.x version that does have all the deprecation notices nearer the time

model/versioning/Versioned.php
```diff
@@ -301,12 +308,13 @@ public function augmentSQL(SQLSelect $query, DataQuery $dataQuery = null) {
 		// Reading a specific stage (Stage or Live)
 		case 'stage':
 			$stage = $dataQuery->getQueryParam('Versioned.stage');
-			if($stage && ($stage != $this->defaultStage)) {
+			if($this->hasVersionedStage($stage) && $stage !== static::DRAFT_STAGE) {
```
dhensby: if it doesn't have stage `$stage` then we silently fail? shouldn't we throw something or at least warn?

tractorcow: The current stage (set via session) might not match the given stage for a particular object; In which case, it should revert to its default stage.

The particular case is where an object only has ONE stage (e.g. `Stage`), but should still be visible if the site is set to set_stage('Live')`. Single-stage objects are a valid use case where we want only the versioning component, but not the staging one.

filesystem/Folder.php
tractorcow: Yeah, the problem is Folder extends File, but we want versioning on Files, and not necessarily Folders...

model/versioning/Versioned.php
tractorcow: It should be there, but I can elaborate.

dhensby:
Well, I've reviewed the one part that I've been asked to. I think someone with more time/focus should review the whole thing and merge (if appropriate)

model/versioning/Versioned.php
```diff
@@ -202,10 +204,10 @@ public static function reset() {
 	 * Construct a new Versioned object.
 	 *
 	 * @var array $stages The different stages the versioned object can be.
-	 * The first stage is considered the 'default' stage, the last stage is
-	 * considered the 'live' stage.
+	 * The first stage must be the value of self::DRAFT_STAGE. If more than one stage is given,
+	 * the final stage must be the value of self::LIVE_STAGE.
 	 */
-	public function __construct($stages = array('Stage','Live')) {
+	public function __construct($stages = array(self::DRAFT_STAGE, self::LIVE_STAGE)) {
```
hafriedlander: Is there any point keeping this as an argument? Given we basically just force the default through the checks below?

model/versioning/Versioned.php
hafriedlander: Do we have any examples of that @tractorcow? And wouldn't it need to be the `Live` stage, because the default read stage is Live?

model/versioning/Versioned.php
```diff
@@ -1196,12 +1196,96 @@ public function doPublish() {
...
 
+	/**
+	 * Trigger publishing of owned objects
+	 */
+	public function onAfterPublish() {
+		$owner = $this->owner;
+
+		// Publish owned objects
+		$published = new ArrayList();
+		foreach ($owner->findOwned(false, $published) as $object) {
```
hafriedlander: How does the extended class know it needs to implement this?

Can we either check for this method & no-op if missing, or, better,  have an interface, like `Owner`, which specifies this method and we check for `is_a('Owner')` and no-op if not.

hafriedlander: Oh, this is going to be added to DataObject. Is that in this PR? I don't see it declared anywhere.

hafriedlander: NM, found.

model/versioning/Versioned.php
dhensby: I had that thought too, but you can pass 1 stage or 3, technically, just the first and last are important

model/versioning/Versioned.php
dhensby: Some of the shop modules use 1 stage for versioning so they can always refer to the price of an item at the time of order.

tractorcow:
I've completed the rest of the triggered actions, and merged it into this PR.

I've also done a bit of refactor based on @hafriedlander's feedback.

model/versioning/Versioned.php
tractorcow: Ok, we've kept 'unstaged' versioning, but have dropped support for >2 stages.

filesystem/File.php
```diff
@@ -563,24 +563,22 @@ public function updateFilesystem() {
 			return false;
 		}
 
+		// Avoid moving files on live; Rely on this being done on stage prior to publish.
+		if(Versioned::get_stage() === Versioned::LIVE) {
```
hafriedlander: Seems safer to reverse this (`Versioned::get_stage() !== Versioned::DRAFT`)

filesystem/FileMigrationHelper.php
```diff
@@ -35,7 +35,7 @@ public function run($base = null) {
 		// Loop over all files
 		$count = 0;
 		$originalState = \Versioned::get_reading_mode();
```
hafriedlander: We're not in a namespace, wonder why `\Versioned` used here. Would prefer `use \Versioned` anyway, makes it easier to refactor later. (Aware this patch doesn't introduce this oddity).

model/DataObject.php
```diff
@@ -398,10 +398,15 @@ public static function composite_fields($class = null, $aggregated = true) {
 	 * field values.  Normally this contructor is only used by the internal systems that get objects from the database.
 	 * @param boolean $isSingleton This this to true if this is a singleton() object, a stub for calling methods.
 	 *                             Singletons don't have their defaults set.
+	 * @param DataModel $model
+	 * @param array $queryParams List of DataQuery params necessary to lazy load, or load related objects.
 	 */
-	public function __construct($record = null, $isSingleton = false, $model = null) {
+	public function __construct($record = null, $isSingleton = false, $model = null, $queryParams = array()) {
```
hafriedlander: Odd to introduce a new constructor param in this PR. What's the motivation?

model/Image.php
```diff
@@ -7,8 +7,8 @@
  * @subpackage filesystem
  */
 class Image extends File {
-	public function __construct($record = null, $isSingleton = false, $model = null) {
-		parent::__construct($record, $isSingleton, $model);
+	public function __construct($record = null, $isSingleton = false, $model = null, $queryParams = array()) {
+		parent::__construct($record, $isSingleton, $model, $queryParams);
```
hafriedlander: Right, so this shows the problem with mucking with the constructor. Every class that overrides it now needs to do this or sad times in sad town for sad people.

tractorcow:
Since versioned.php is too large to preview I've created https://github.com/open-sausages/silverstripe-framework/commits/features/triggered-publishing-split

filesystem/File.php
tractorcow: Will change.

filesystem/FileMigrationHelper.php
tractorcow: Yeah that's ok. It's more because at the time of writing I was namespacing everything, but not in this specific change. I'll revert that.

model/DataObject.php
tractorcow: Yes, but it's necessary because otherwise lazy-loading blows up during dependency injection. VirtualPage cost me at least 2 hours debugging due to this issue; DI will request various properties from an object, and VirtualPage nearly always loads the virtualised page during these.

The only way to fix this is to assign query parameters at object construction, so they are always available.

model/Image.php
tractorcow: Yes but it's necessary.

tractorcow: I'll elaborate in the 4.0.0.md changelog.

tractorcow:
Another note; `$owner = $this->owner` everywhere in Versioned is a work-around for another bug I haven't quite solved yet; if you call a method on an object from an extension, sometimes it'll reset $this->owner back to null after the nested method returns.

tractorcow:
Thank you! I notice the behat test issues on CMS; I'll fix those with a PR to the behat-extension module tomorrow.

tractorcow:
Already fixed with https://github.com/silverstripe/silverstripe-framework/pull/5074

PR #5161, opened 2016-03-08
New styles for main nav

clarkepaul:
Screenshots for the following updates:
Updated nested navigation styles for the navigation incl. when it is collapsed. It has been designed so that the group should have an icon (rather than the nested items). Tested and works on touch device as well.

Known bugs:
When the menu is collapsed the group title should no longer be clickable.
Groups need the ability to add an icon (might already be possible?).

pasted_image_9_03_16__12_29_pm
pasted_image_9_03_16__12_40_pm


tractorcow:
I've tested this and it looks good. Please rebase and I can merge @clarkepaul .

scott1702:
Rebased @tractorcow

tractorcow:
Merged, thanks.

PR #5162, opened 2016-03-09
IMPROVEMENT round trip email subject in html editor

mikenz:
In the html editor, when editing an email email link that has a subject, the subject should be easily editable. Currently the subject is left appending to the end of the email address and still url encoded.

I'm submitting this via the github editor so can't run a gulp build to update the dist and bundled files.


javascript/src/HtmlEditorField.js
```diff
@@ -844,7 +844,14 @@ $.entwine('ss', function($) {
 				action = "update";
 			}
 
-			if(href.match(/^mailto:(.*)$/)) {
+            if(href.match(/^mailto:(.*)\?subject=(.*)$/)) {
```
dhensby: I think we can make this more generic to parse the url better and pull out all the query params.

dhensby:
This is definitely a `FIX`. Is it present in 3 or just master?

mikenz:
Sorry I don't run 3 only master.

javascript/src/HtmlEditorField.js
dhensby: This would also fail for any links that were like `mailto:me@example.com?body=Body+here&subject=Subject+here`

dhensby:
OK - I think we need the query string parser to be a bit more resilient before this can go in

javascript/src/HtmlEditorField.js
mikenz: The link editor UI doesn't have a way to create a link like that. If the user manually creates a link with a body then they're going to have manually edit it in the future too. Or a you suggesting we add a field for editing the body in the link editor too?

mikenz:
My change was to parse the links created by the link editor in the html editor which can only create links in one format:
`					case 'email':
						href = 'mailto:' + this.find(':input[name=email]').val();
						if (subject) {
							href += '?subject=' + encodeURIComponent(subject);
						}
						target = null;
						break;`

extending it to support other formats that can't be round tripped without data loss (eg the body being removed) is outside of the scope of this PR. Happy to close the PR and raise a issue if you want.

javascript/src/HtmlEditorField.js
dhensby: I'm suggesting we make this a bit more versatile so we don't introduce regressions with future features and we allow this to be extensible by other devs that decide to add other parameters in their own code-base

tractorcow: `/^mailto:(.*)(\?|&)subject=([^&]*)$/` maybe?

kinglozzer:
Confirmed as fixed in SS4, sorry we missed the SS3 boat on this one @mikenz!

PR #5164, opened 2016-03-09
FIX Inserting a 'Download a file' without selecting an image

mikenz:
In the html editor, when inserting a 'Download a file' link without selecting a file to link to, no link should be inserted. Currently is inserting a link to an undefined file id: [file_link,id=undefined]


dhensby:
great catch - is this only happening in master or is it a problem in 3?

mikenz:
I don't run 3 so don't know, sorry.

dhensby:
Wow, nice. Thanks for running v4, it's great to see it getting an outing in the real world.

tractorcow:
Thanks, but since this is in a src file, can you please run the node compiler over this to build the dist file?

`npm install && npm run build` should do it. ;)

mikenz:
Updated dist files, of course this will cause merge conflicts with some of my other pull requests for whoever is merging.

dhensby:
merge on green ticks

dhensby:
closed with 37f11f3

dhensby:
thanks!

javascript/dist/i18n.js
```diff
@@ -132,7 +132,7 @@
 						return match;
 					}
 
-					return subMatch1 + params[i += 1];
+					return subMatch1 + params[i++];
```
tractorcow: I think I forgot to ask someone to build dist files in a prior PR... oh well this fixes it!

PR #5172, opened 2016-03-13
Remove pagination get param from links pointing to page 1 within pagination

Andrew-Mc-Cormack:
On page 2 of a paginated list the "previous" and "1" pagination link point to /?start=0 as the first page in the pagination series. Both / and /?start=0 will return the same page and therefore its preferable to point to the non param version for duplicate content reasons. A page when first visited will use the non param version / of the URL. Its only when we start navigation through pagination and are on page 2 that there is a link back to /?start=0. Search engines will navigate the site the same as users can and will crawl this this URL (/?start=0) and index it alongside the main URL (/) causing a duplicate page.

### DIFF
--- a/core/PaginatedList.php
+++ b/core/PaginatedList.php
@@ -229,9 +229,19 @@ class PaginatedList extends SS_ListDecorator {
 		}
 
 		for ($i = $start; $i < $end; $i++) {
+
+			// Remove the query sting for the first page link
+			// /?start=0 returns the same page as /
+			$link = HTTP::setGetVar($this->getPaginationGetVar(), $i * $this->getPageLength());
+
+			if($i === 0) {
+				$parts = explode("?", $link);
+				$link = $parts[0];
+			}
+
 			$result->push(new ArrayData(array(
 				'PageNum'     => $i + 1,
-				'Link'        => HTTP::setGetVar($this->getPaginationGetVar(), $i * $this->getPageLength()),
+				'Link'        => $link,
 				'CurrentBool' => $this->CurrentPage() == ($i + 1)
 			)));
 		}
@@ -430,7 +440,16 @@ class PaginatedList extends SS_ListDecorator {
 	 */
 	public function PrevLink() {
 		if ($this->NotFirstPage()) {
-			return HTTP::setGetVar($this->getPaginationGetVar(), $this->getPageStart() - $this->getPageLength());
+
+			$start = $this->getPageStart() - $this->getPageLength();
+
+			$link = HTTP::setGetVar($this->getPaginationGetVar(), $start);
+
+			if($start === 0) {
+				$parts = explode("?", $link);
+				$link = $parts[0];
+			}
+			return $link;
 		}
 	}
 
--- a/core/PaginatedList.php
+++ b/core/PaginatedList.php
@@ -235,8 +235,7 @@ class PaginatedList extends SS_ListDecorator {
 			$link = HTTP::setGetVar($this->getPaginationGetVar(), $i * $this->getPageLength());
 
 			if($i === 0) {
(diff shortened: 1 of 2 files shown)


core/PaginatedList.php
```diff
@@ -229,9 +229,19 @@ public function Pages($max = null) {
 		}
 
 		for ($i = $start; $i < $end; $i++) {
+
+			// Remove the query sting for the first page link
+			// /?start=0 returns the same page as /
+			$link = HTTP::setGetVar($this->getPaginationGetVar(), $i * $this->getPageLength());
+
+			if($i === 0) {
+				$parts = explode("?", $link);
```
tractorcow: Won't this remove ALL querystring args? What if there's another arg that is needed to render.

E.g.

`http://www.mysite.com/page?show=category1&start=0` should be changed into `http://www.mysite.com/page?show=category1` not `http://www.mysite.com/page`

tractorcow: I suggest parsing out querystring args, removing the `start` argument, then re-encoding back into a string.

core/PaginatedList.php
```diff
@@ -457,4 +474,30 @@ public function getRequest() {
...
+	 * Remove the first page pagination get var to stop duplicate page issues
+	 * / and /?start=0 return the same page
+	 *
+	 * @param  string $link
+	 * @return string
+	 */
+	private function removeFirstPagePaginationParam($link)
+	{
+		// get the path and query string
+		$parts = explode("?", $link);
```
dhensby: This needs the 3rd parameter to be `2`

tests/model/PaginatedListTest.php
```diff
@@ -325,7 +325,8 @@ public function testPrevLink() {
 
 		$this->assertNull($list->PrevLink());
 		$list->setCurrentPage(2);
-		$this->assertContains('start=0', $list->PrevLink());
+		$this->assertNotNull($list->PrevLink());
+		$this->assertNotContains('start=0', $list->PrevLink());
```
dhensby: Please add a test where other query params are present so we can ensure there are no regressions

core/PaginatedList.php
```diff
@@ -457,4 +474,30 @@ public function getRequest() {
...
 	}
 
+	/**
+	 * Remove the first page pagination get var to stop duplicate page issues
+	 * / and /?start=0 return the same page
+	 *
+	 * @param  string $link
+	 * @return string
+	 */
+	private function removeFirstPagePaginationParam($link)
```
dhensby: I think this function can be implemented in a more simple manner. 

Why not use the inbuilt query string parser and then just under the pagination get var if it's 0

This function could be more generic and called "clean url params" and always run without checking that start is =0 in other functions first

tractorcow: This whole method could be one preg_replace.

```php
return preg_replace('/([?&])'. preg_quote($this-> getPaginationGetVar()) . '=0(&|$)/', '$1', $link);
```

dhensby: I think the last set of parenthesis can be `(?|&|$)` because you can join query strings with `?` I believe

Andrew-Mc-Cormack: All the above noted, will re-visit this later and put together something more solid.

tractorcow: Thanks. :D

tractorcow: And if anyone is using multiple `?` to join all their querystrings, I hope their site breaks....

dhensby:
closing due to inactivity for 1.5 years. I'd put the time in to get it over the line but I see this as non-essential improvement.

dhensby:
@Andrew-Mc-Cormack feel free to reopen if you would like to get this over the line :)

PR #3355, opened 2014-08-02
Update Controller to allow extension in handleAction()

jaredkipe:
Controller's parent class (RequestHandler) has two extensions in its handleAction() method that are obscured by Controller's implementation.

There are many reasons to hook into action, custom logging being the most obvious.

PR #5173, opened 2016-03-13
BUG Remove duplicate extension hook

### LINKED ISSUE #5170: [CONTROL] [3.2] beforeCallActionHandler is called two times
The problem comes from bad fix in de09bffd. The real problem is that default action is implicit and handled differently from other actions. It should be implemented as `Controller::index`

```
public function index(SS_HTTPRequest $request) {
    return $this->getViewer('index')->process($this);
}
```

see http://stackoverflow.com/questions/35865868/silverstripe-publish-page-in-the-future for example of extension

tractorcow:
Fixes #5170
Reverts #3355

This isn't needed, since the hook will be invoked in the `parent::handleAction` call.

### DIFF
--- a/control/Controller.php
+++ b/control/Controller.php
@@ -197,8 +197,6 @@ class Controller extends RequestHandler implements TemplateGlobalProvider {
 	 * @return HTMLText|SS_HTTPResponse
 	 */
 	protected function handleAction($request, $action) {
-		$this->extend('beforeCallActionHandler', $request, $action);
-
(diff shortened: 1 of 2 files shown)


tractorcow:
I've fixed this so the original issue addressed in #3355 remains fixed.

dhensby:
Yes, it makes sense to send the response to the after handler.

tractorcow:
Done. :D

PR #5174, opened 2016-03-13
Tinymce image shortcodes

chillu:
Required to work with new filesystem abstraction - use `[image src="/assets/myfile.jpg" id="99"]` instead of ``. The `src` attribute is added in order to make TinyMCE previews possible, and auto-updated behind the scenes.

We'll also need to write a migration task for existing content, since image relationship tracking won't work with the "legacy" format. It's on our team backlog, before 4.0 stable. For now, it'll only convert images when a record is saved through the CMS.

Raising this PR a bit early mainly for @tractorcow's peer review, which will probably squash some of these commits.

Tested on Chrome latest and IE9

Merge with https://github.com/silverstripe/silverstripe-cms/pull/1423


model/Image.php
```diff
@@ -26,60 +27,106 @@ public function getIsImage() {
...
-		$imageElements = $htmlValue->getElementsByTagName('img');
-		if($imageElements) foreach($imageElements as $imageElement) {
-			$imageDO = null;
-			$src = $imageElement->getAttribute('src');
+	public static function handle_shortcode($args, $content, $parser, $shortcode, $extra = array()) {
+		// Check if there is a suitable record
+		$record = static::find_shortcode_record($args);
+		if (!$record) {
+			return null;
+		}
```
tractorcow: You're missing an important canView() check before the `$record->Link()`... this will potentially expose non-allowed assets to users or site visitors.

tractorcow: I should add to this; Calling getURL() on an image causes it to be whitelisted for that user. The whitelisting relies on user-code doing the appropriate permission checking before generating a url.

tractorcow:
All other code looks good. I've tested locally and am happy, other than the missing permission check. ;D

tractorcow:
Squashed and rebased.

tractorcow:
Squashed and rebased again.

tractorcow:
Thank you!

PR #5175, opened 2016-03-14
Improvement: Set the owner on an extension

Firesphere:
Before getting all the available method names.

This enhances the extensions to enable certain features like adding wrapper methods on construct instead of having to have them on the object itself.

### DIFF
--- a/core/Object.php
+++ b/core/Object.php
@@ -816,7 +816,9 @@ abstract class Object {
 		}
 
 		if(method_exists($extension, 'allMethodNames')) {
+			if ($extension instanceof Extension) $extension->setOwner($this);
 			$methods = $extension->allMethodNames(true);
+			if ($extension instanceof Extension) $extension->clearOwner();
 
 		} else {
 			if(!isset(self::$built_in_methods[$extension->class])) {
--- a/tests/model/DataExtensionTest.php
+++ b/tests/model/DataExtensionTest.php
@@ -166,6 +166,13 @@ class DataExtensionTest extends SapphireTest {
 		$this->assertEquals("hello world", $do->testMethodApplied());
 	}
 
+	public function testExtensionAllMethodNamesHasOwner() {
+		/** @var DataExtensionTest_MyObject $do */
+		$do = DataExtensionTest_MyObject::create();
+
+		$this->assertTrue($do->hasMethod('getTestValueWithDataExtensionTest_MyObject'));
+	}
+
 	public function testPageFieldGeneration() {
 		$page = new DataExtensionTest_CMSFieldsBase();
 		$fields = $page->getCMSFields();
@@ -365,10 +372,23 @@ class DataExtensionTest_Faves extends DataExtension implements TestOnly {
 
 }
 
-class DataExtensionTest_AppliedToDO extends DataExtension implements TestOnly {
+class DataExtensionTest_AppliedToDO extends DataExtension implements TestOnly
+{
 
-	public function testMethodApplied() {
+	public function testMethodApplied()
+	{
 		return "hello world";
+	}
+
+}
+
+class DataExtensionTest_AllMethodNames extends DataExtension implements TestOnly
+{
+	public function allMethodNames()
+	{
+		return array(
+			strtolower('getTestValueWith'.$this->owner->ClassName)
+		);
 	}
 
 }
@@ -376,6 +396,7 @@ class DataExtensionTest_AppliedToDO extends DataExtension implements TestOnly {
 DataExtensionTest_MyObject::add_extension('DataExtensionTest_Ext1');
 DataExtensionTest_MyObject::add_extension('DataExtensionTest_Ext2');
 DataExtensionTest_MyObject::add_extension('DataExtensionTest_Faves');
+DataExtensionTest_MyObject::add_extension('DataExtensionTest_AllMethodNames');
 
 /**
  * Base class for CMS fields


tractorcow:
Can you show me an example of user code that would benefit from this change? I think I need to see it in context before I can make a judgement call on this.

tractorcow:
Tests stalled, restarted. :D

Firesphere:
It's used in the encryption module, to set custom handling of getting the database values. To make a getter return the unencrypted value here:
https://github.com/madmatt/silverstripe-encrypt-at-rest/blob/master/code/extensions/EncryptDataObjectFieldsExtension.php
(Do note, this module is a very crude/cruel v0.0.1, so don't judge it too hard)

The specific use case here, is to be able to hook into the owner at construct. Otherwise, we are unable to check which fields are encrypted.

Other usecases would be, to setup custom functionality on core classes like member and file etc.

tractorcow:
No I see, that's very useful.

I think you should squash this an try again to get the tests passing... travis is having a sulk today.

Firesphere:
Yeah, I also need to add a slight fix.

Now, I'm checking if the method exists. But it should actually check, if the extension is of type Extension, so it won't accidentally call setOwner on a wrong type of class.

A case here, is for example using setOwner/getOwner on a user having items.

I'll fix and squash later this afternoon. It's lunchtime first ;)

tractorcow:
Yeah that's a good idea; That would be a lot more robust. ;D

Firesphere:
There we go. Green tests.

Does this feature require a test? I might be looking wrong, but I can't find an ExtensionTest.

tractorcow:
I think so. :D

There is `DataExtensionTest`

Firesphere:
Yes, I'm blind.

Now to think of a way on how to test it.

tractorcow:
Maybe your extension adds a method that returns the name of the class that was set as the original object, and then you call that method and test that object string is returned?

Firesphere:
Since it's specifically the call for allMethodsName, I've added the test to see if at that point of calling, the owner is known, by checking if the wrapperfunction name contains the name of the DataObject it's extending.

Created as a separate commit, seemed like the cleaner thing to do.

Firesphere:
Ow, by the way, would it be useful to add a comment there?

And what's with the behat tests failing? :-1:

madmatt:
Hrm yeah, the travis-setup.php script seems to be failing, some repository can't be cloned?

@tractorcow This is merging into the 3 branch, has something changed/broken there maybe?

core/Object.php
```diff
@@ -816,7 +816,9 @@ protected function addMethodsFrom($property, $index = null) {
 		}
 
 		if(method_exists($extension, 'allMethodNames')) {
+			if (is_a($extension, 'Extension')) $extension->setOwner($this);
```
tractorcow: I think we prefer `instanceof Extension`, since is_a has some php-version sensitive behaviour.

tests/model/DataExtensionTest.php
```diff
@@ -159,13 +159,20 @@ public function testExtensionCanBeAppliedToDataObject() {
...
 
 		$this->assertEquals("hello world", $mo->testMethodApplied());
 		$this->assertEquals("hello world", $do->testMethodApplied());
 	}
 
+	public function testExtensionAllMethodNamesHasOwner() {
+		/** @var DataExtensionTest_MyObject $do */
+		$do = DataExtensionTest_MyObject::create();
+
+		$this->assertTrue($do->hasMethod('getTestValueWithDataExtensionTest_MyObject'));
```
tractorcow: Good test. :)

tractorcow:
Problem is nothing to do with 3 branch; It's github rate limiting us.

tests/model/DataExtensionTest.php
```diff
@@ -159,13 +159,20 @@ public function testExtensionCanBeAppliedToDataObject() {
 		$do = new DataObject();
 		$mo = new DataExtensionTest_MyObject();
 
-		$this->assertTrue($do->hasMethod('testMethodApplied'));
 		$this->assertTrue($mo->hasMethod('testMethodApplied'));
+		$this->assertTrue($do->hasMethod('testMethodApplied'));
```
dhensby: Any reason for swapping these lines?

tests/model/DataExtensionTest.php
```diff
@@ -159,13 +159,20 @@ public function testExtensionCanBeAppliedToDataObject() {
...
 
 		$this->assertEquals("hello world", $mo->testMethodApplied());
 		$this->assertEquals("hello world", $do->testMethodApplied());
 	}
 
+	public function testExtensionAllMethodNamesHasOwner() {
+		/** @var DataExtensionTest_MyObject $do */
+		$do = DataExtensionTest_MyObject::create();
```
dhensby: Shouldn't tests use `new` instead of the create because this will be fragile to project ovverrides...

tests/model/DataExtensionTest.php
Firesphere: No. None. I think I temporarily removed the DO, to see if that caused an error in my initial test, and pasted it back on the wrong line, without checking if it was the right line.
Sorry

tests/model/DataExtensionTest.php
Firesphere: In the tests, it's a bit a mix between `new` and `::create()`. I just went with create because it's what I use most.

core/Object.php
Firesphere: Understandable. Same goes for the line below the allMethodNames() check.

I'll see if I have time to fix that monday and squash them together.

tests/model/DataExtensionTest.php
Firesphere: Heh, it took me some pondering to make a valid test, but in the end, I figured "All that needs testing, is if the owner is set". Thus, just checking and add some syntactic sugar to it was all that it needed :)

tractorcow:
@Firesphere do you have time this week to do any of the code cleanup suggestions? I don't think there's anything blocking merge, unless you are keen. :D

If you make any changes, it wouldn't hurt to squash at the same time. Ping me if you are ready.

Firesphere:
Updated to use instanceof, squashed all into single commit.

PR #5182, opened 2016-03-15
DOCS: FulltextSearchable code example

christopherdarling:


dhensby:
Thanks. This could probably be fixed in 3.1 and up too

PR #5186, opened 2016-03-15
NEW consistent file icons

jonom:
File icons have been updated for consistency and a StripThumbnail method added to File for icon display in gridfield instances (such as file selection interface)

Related CMS PR: https://github.com/silverstripe/silverstripe-cms/pull/1429

/cc @clarkepaul @scott1702 

#### Before

![screen shot 2016-03-15 at 11 52 06 am] (https://cloud.githubusercontent.com/assets/1079425/13790295/fcd8bca6-eaa4-11e5-8eae-7d9404760336.png) screen shot 2016-03-15 at 11 53 18 am screen shot 2016-03-15 at 11 54 58 am

image_32 mov_32 xls_32 audio_32 doc_32 pdf_32 zip_32 generic_32

#### After

screen shot 2016-03-15 at 11 44 48 am screen shot 2016-03-15 at 11 11 31 am screen shot 2016-03-15 at 11 10 42 am

generic mov doc audio zip

Note: an image icon shouldn't be necessary because a thumbnail will be displayed instead of an icon


clarkepaul:
Good improvement, can't wait to do a complete overhaul of all icons.

jonom:
Yeah I'm amped for that too - the SS4 concepts are looking so good so far!

dhensby:
@jonom where did you get the icons from?

Are there any more icons, like spreadsheet?

I think this looks a little too close to the bottom and should be spaced evenly above and below:

jonom:
The zip and audio icons contain elements from royalty free 'commercial use okay, no credit required' icons, the doc and video ones I did in photoshop. The background comes from an existing SS asset.

I didn't do a separate spreadsheet icon because in SS3 there's no spreadsheet file category, common spreadsheet file extensions are grouped under 'doc'. I didn't think I could move extensions out of one file category and in to another in 3 branch due to SemVer (I assume this would count as modifying existing API) and don't really want to create any icons that are extension specific - I think that SS should only include one icon per category and be brand agnostic (e.g. no Word / Excel / Adobe icons).

I guess I could include a spreadsheet icon and let people introduce their own spreadsheet category in yml config, but that may be impeded by developer's inability to remove array items from a config item - i.e. they will be able to add 'xlsx' to to a new 'spreadsheet' category but won't be able to remove same from the doc category.

Or I could be less of an idealist and make an exception for spreadsheets by creating an icon and duplicating it three times as 'xls', 'xlsx' and 'numbers'. Since 3 is on the way out and file management is being completely redone in 4 maybe that's the most pragmatic thing to do.

> I think this looks a little too close to the bottom and should be spaced evenly above and below

Yeah I definitely agree, I didn't change any CSS for this field but I should. Think the initial design was only intended to accommodate image thumbnails not icons so the image is displayed border to border. I figured it looks better than it currently does even if still not great :P

dhensby:
I agree with what you're saying. If there's not spreadsheet icon (remember csv would be one too) then don't add one.

jonom:
I've updated the commit to style the file icons better after upload. I noticed while I was doing that that we use a range of different thumbnail files for images with different styles (crop/pad/fit) so they're not only inconsistent in style but inefficient - would be better for clients to only have to download one 'icon size' thumbnail resource for each file, and one larger one if they open a detail edit form. I started out on cleaning this up but it's a bit of a rabbit hole - there are all these config settings to do with preview widths and heights which I think are redundant (would probably break the layout if you changed the settings) but don't want to get in to them :) Will all be re-done in 4 anyway so don't want to go too deep.

#### Before

screen shot 2016-03-19 at 10 48 17 am screen shot 2016-03-19 at 10 49 07 am screen shot 2016-03-19 at 10 49 11 am

#### After

screen shot 2016-03-19 at 9 55 16 am screen shot 2016-03-19 at 9 40 47 am screen shot 2016-03-19 at 9 38 10 am

jonom:
> If there's not spreadsheet icon (remember csv would be one too) then don't add one.

Before this PR there is an itty-bitty icon for xls extension but not xlsx, numbers or csv. xls_32 This PR removes that icon so xls uploads would get the larger doc icon instead. It's not a spreadsheet icon but it's still an improvement I think.

dhensby:
I'm happy. Thanks. Great improvement

jonom:
Cheers :+1:

tractorcow:
Hm, we probably should not be deleting assets in a non-major release; They could break hard-coded image links.

dhensby:
> Hm, we probably should not be deleting assets in a non-major release; They could break hard-coded image links.

Good point @jonom can you bring back the assets that were deleted but not replaced?

jonom:
I guess... seems like one of those SemVer grey areas. Would be hard to ever change anything if we worry about people hard-coding things instead of using the API.

The problem is that if I put back those file-extension specific files as they were, any uploads with those extensions will go back to looking crappy. I guess I could duplicate the category icons with the names of the old files though?

jonom:
Oh actually they won't go back to looking crappy because the Icon() method is looking for PNGs now instead of GIFs. So yeah I can do that if you like.

dhensby:
:) please

jonom:
https://github.com/silverstripe/silverstripe-framework/pull/5209

PR #5190, opened 2016-03-16
Converted AdminRootController to PSR-2

assertchris:
This mainly converts `AdminRootController` to PSR-2, also limiting line length to 100 and adding a few type hints.

I recommend reviewing it at [this link](https://github.com/silverstripe/silverstripe-framework/pull/5190/files?w=1).

This is a test-case for [a wider discussion](https://groups.google.com/forum/#!topic/silverstripe-dev/IoMtKlju3Rc).

### DIFF
--- a/admin/code/AdminRootController.php
+++ b/admin/code/AdminRootController.php
@@ -4,96 +4,121 @@
  * @package framework
  * @subpackage admin
  */
-class AdminRootController extends Controller {
+class AdminRootController extends Controller
+{
+    /**
+     * The url base that all LeftAndMain derived panels will live under. Won't automatically update
+     * the base route if you change this - that has to be done separately.
+     *
+     * @var string
+     *
+     * @config
+     */
+    private static $url_base = 'admin';
 
-	/**
-	 * @var string
-	 * @config
-	 * The url base that all LeftAndMain derived panels will live under
-	 * Won't automatically update the base route if you change this - that has to be done seperately
-	 */
-	private static $url_base = 'admin';
+    /**
+     * The LeftAndMain child that will be used as the initial panel to display if none is selected,
+     * like when you visit "/admin".
+     *
+     * @var string
+     *
+     * @config
+     */
+    private static $default_panel = 'SecurityAdmin';
 
-	/**
-	 * @var string
-	 * @config
-	 * The LeftAndMain child that will be used as the initial panel to display if none is selected (i.e. if you
-	 * visit /admin)
-	 */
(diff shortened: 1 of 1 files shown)


admin/code/AdminRootController.php
```diff
@@ -4,96 +4,117 @@
...
+    /**
+     * Holds an array of url_pattern => controller k/v pairs, the same as Director::rules. However
+     * this is built dynamically from introspecting on all the classes that derive from LeftAndMain.
+     *
+     * Don't access this directly - always access via the rules() accessor below, which will build
+     * this array the first time it's accessed.
+     *
+     * @var array
+     */
+    private static $_rules = null;
```
kinglozzer: > Property names SHOULD NOT be prefixed with a single underscore to indicate protected or private visibility.
http://www.php-fig.org/psr/psr-2/#4-2-properties

Should be fine to rename this to `$rules` as it’s private

admin/code/AdminRootController.php
```diff
@@ -4,96 +4,117 @@
...
+     * @param string $controllerClass
+     */
+    protected static function add_rule_for_controller($controllerClass)
+    {
+        $urlSegment = Config::inst()->get($controllerClass, 'url_segment', Config::FIRST_SET);
+        $urlRule = Config::inst()->get($controllerClass, 'url_rule', Config::FIRST_SET);
+
+        if ($urlSegment) {
+            // Make director rule
+            if ($urlRule[0] == '/') $urlRule = substr($urlRule, 1);
```
kinglozzer: R.e. control structures:

> The body of each structure MUST be enclosed by braces.
http://www.php-fig.org/psr/psr-2/#5-control-structures

I’m guessing `php-cs-fixer` can’t do this automatically :disappointed:

admin/code/AdminRootController.php
assertchris: My bad. I did the conversation by hand, and forgot that rule.

admin/code/AdminRootController.php
assertchris: My bad. I did the conversation by hand, and forgot that rule.

admin/code/AdminRootController.php
assertchris: Made all the non-breaking changes I could. I think we need to employ a better migration strategy than "convert everything and fix the breakages". :)

admin/code/AdminRootController.php
dhensby: > I did the conversation by hand

Hardcore

PR #5199, opened 2016-03-18
Allow '0' in DropdownField

### LINKED ISSUE #5196: Enum('0,+,++','0') does not list '0' as an option in a DropDownList
If I use `Enum('0,+,++','0')`, the entry '0' will not get listed in an automatically populated drop-down list. (v3.3.1)

JorisDebonnet:
This fixes #5196. 

It's actually sort of a workaround, because I feel like whatever code prettifies `<select>`s in the CMS should not ignore `<option>`s without any content. But this works too, without any harm (it just adds an invisible space to _all_ values, instead of just to falsy values).

### DIFF
--- a/templates/forms/DropdownField.ss
+++ b/templates/forms/DropdownField.ss
@@ -1,5 +1,5 @@
 <select $AttributesHTML>
 <% loop $Options %>
-	<option value="$Value.XML"<% if $Selected %> selected="selected"<% end_if %><% if $Disabled %> disabled="disabled"<% end_if %>><% if Title %>$Title.XML<% else %>&nbsp;<% end_if %></option>
+	<option value="$Value.XML"<% if $Selected %> selected="selected"<% end_if %><% if $Disabled %> disabled="disabled"<% end_if %>><% if $Title.exists %>$Title.XML<% else %>&nbsp;<% end_if %></option>
 <% end_loop %>
 </select>


JorisDebonnet:
(and this could probably be merged all the way to 3.1)

dhensby:
If you want it applied to 3.1 and up then the PR needs to go into the 3.1 branch as we only merge code up from 3.1 -> master and not down.

I think we can wrap the `&nbsp;` in a <% if not $Title %> just to keep the behaviour the exact same and to prevent any regressions.

Have you tested this in the CMS thoroughly?

JorisDebonnet:
Ah, I thought maybe some commits from the 3-branch were sometimes merged into older 3.x branches. They only go into the most recent (i.e. 3.3) then, or not even that (and only 3.4)? Is there maybe a page somewhere detailing how the branches work? :D 

Anyway, I personally don't really care whether it goes into previous versions. For now, I just improved this fix: the only change is now that the condition is changed from `Title` to `$Title.exists` (which allows the string '0').

dhensby:
the `.exists` is implied (ie: the template checks the `->exists` method to determine if there's a value.

JorisDebonnet:
Hmm. Well, I can only report what I see: when I put `.exists` there, 0 appears in the dropdown, and if it's not, it doesn't!

dhensby:
Wierd :/

on branches see: last point in https://docs.silverstripe.org/en/3.2/contributing/code/#quickfire-do-s-and-don-t-s and https://docs.silverstripe.org/en/3.2/contributing/release_process/#supported-versions

tractorcow:
Mega super related; https://github.com/silverstripe/silverstripe-framework/pull/4053

I suggest looking through the code in that change to discover where your '0' is getting treated as false.

I am hesitant about merging this in, because it's only resolving the symptoms, not cause of, a bug.

tractorcow:
@dhensby I wonder if this would be fixed by your "always cast" pull request, since it would force the value to be cast to a Varchar, which must have .exist() called on it.

sminnee:
I'm going to merge this because I don't think doing so is problematic. If there are other, deeper issues at play, then it's fine to look at them too, but I don't think think stops us form including this fix.

PR #5200, opened 2016-03-18
FIX Parameter tokens now redirect to correct url if mod_rewrite is off

### LINKED ISSUE #5197: Redirect after flush=... ignores index.php in URL
http<nolink>://www.example.com/**index.php/admin**?flush=1 will, after the flushing operation, redirect to http<nolink>://www.example.com/**admin**?flush=1&&flushtoken=...

That is a bit annoying when developing on a local server that needs `index.php` for whatever reason, because flushing throws you off the page you were on every time.

dhensby:
fixes #5197


tractorcow:
Seems ok with me.

PR #5208, opened 2016-03-20
FIX tab wrapping

jonom:
jQuery UI css specificity is different to SS so causes some conflicts, including making the active link slightly taller than others. I've also changed the layout slightly so the first links in all rows line up.

It's really annoying having an .editorconfig file in this repo that isn't consistent with the files in it. What's the current thinking on white space? Should I redo this commit with editor config support disabled? Here are the actual changes: https://github.com/silverstripe/silverstripe-framework/pull/5208/files?w=0

**Before**

screen shot 2016-03-20 at 9 55 50 am

**After**

screen shot 2016-03-20 at 11 12 00 am


dhensby:
I'm happy with the whitespace being fixed PR by PR. I know others would rather that PRs weren't polluted by whitespace cleanup. 

I get the editor config contradicts the current style, but that's to prevent new contributions compounding the bad practice.

dhensby:
Ping me on green ticks. 

Have you tested to ensure there aren't regressions? 

It may be worth fixing in 3.1 and up?

jonom:
> It may be worth fixing in 3.1 and up?

Don't know if I can be bothered - no one ever filed an issue about it before and the tab style is quite different starting from 3.3. Do you think it's important?

> Have you tested to ensure there aren't regressions?

I've clicked around the CMS looking at any tabbed interfaces but can't find any negative effects. Works in IE8 too:

screen shot 2016-03-20 at 11 35 01 am

Oh hey... merged! I'll stop typing then lol

dhensby:
;P

Well, I'm sure this was reported in old versions, patched and reverted. 

I think fixing it in older versions is quite important, yes. It's pretty embarrassing...

jonom:
I like the idea of only porting critical bug fixes to older minor releases... because I also like the idea of more regular minor releases and less merge regressions. 3 is kind of busted right now for instance, might not be if we didn't have to push so much forward from older minor releases? Now that we follow SemVer I don't think we should be supporting previous minor releases very much.

dhensby:
Ok

PR #5209, opened 2016-03-20
Restore deleted file icons

jonom:
Reverts asset removal from commit 292378735251afe221f77c097095c859dc49caf5. ping @dhensby


dhensby:
@jonom I think some of these can be replaced with the ones you've already made (mov, generic, etc)

jonom:
They could but does that make sense? They won't get used at all out of the box and if someone is using them by hard-coded reference they maybe want to keep using these icons.

dhensby:
Good point

PR #5213, opened 2016-03-21
API and BUG fix; Update ssbuttons tinymce plugin for image shortcode parsing

tractorcow:
Fixes regressions in https://github.com/silverstripe/silverstripe-framework/pull/5174
Required for https://github.com/silverstripe/silverstripe-cms/pull/1423

I've also moved this out of thirdparty since it's not a thirdparty module.


chillu:
Tested on IE10, works like a charm!

PR #5218, opened 2016-03-21
Fixed issue where `canViewVersioned` caused a DB error…

### LINKED ISSUE #5189: Versioned::canViewVersioned checks for a `Live` version, even though that may not exist!
In 3.3, a new `canViewVersioned` permission check was implemented for Versioned DataObjects.

This check assumes, that *all* versioned DataObjects have at least two stages, where one is named `Live`. It doesn't consider the following (completely valid) cases:

```yml
MyDataObject:
  extensions:
    - "Versioned('Bleep','Bloop')"
```

This will create tables: `MyDataObject`, `MyDataObject_versions` and `MyDataObject_Bloop`.

Same goes for a case where only versioning is needed, but not different stages:

```yml
MyVersionedObject:
  extensions:
    - "Versioned('Live')"
```
This will create tables: `MyVersionedObject` and `MyVersionedObject_versions`.

Both these cases will cause errors with the current implementation of `canViewVersioned`, since this will [check for a `_Live` table via `Versioned::get_versionnumber_by_stage`](https://github.com/silverstripe/silverstripe-framework/blob/9fed5561f41ae4774d012aab00ea726c59ebec89/model/Versioned.php#L823)

So `canViewVersioned` should be aware of the stages and not just presume there's `Stage` and `Live` present. I think it would be sufficient to only check the current live-version in case of the default setup of `Stage` and `Live` and otherwise skip this check (not needed in a single-stage setup and can't be properly determined with arbitrarily named stages)

bummzack:
… when Versioned was used with stages other than the default "Stage" and "Live".
Updated VersionedTest to also check an Object with a single stage in the canView test.

Fixes #5189

### DIFF
--- a/model/Versioned.php
+++ b/model/Versioned.php
@@ -795,9 +795,14 @@ class Versioned extends DataExtension implements TemplateGlobalProvider {
 			return true;
 		}
 
+		// If there are less than 2 stages, we can exit early since comparing stages is not needed
+		if(count($this->stages) < 2){
+			return true;
+		}
+
 		// If we weren't definitely loaded from live, and we can't view non-live content, we need to
-		// check to make sure this version is the live version and so can be viewed
-		$latestVersion = Versioned::get_versionnumber_by_stage($this->owner->class, 'Live', $this->owner->ID);
+		// check to make sure this version is the live version and so can be viewed.
+		$latestVersion = Versioned::get_versionnumber_by_stage($this->owner->class, $this->liveStage, $this->owner->ID);
 		if ($latestVersion == $this->owner->Version) {
 			// Even if this is loaded from a non-live stage, this is the live version
 			return true;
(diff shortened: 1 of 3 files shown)


model/Versioned.php
```diff
@@ -796,8 +796,11 @@ public function canViewVersioned($member = null) {
 		}
 
 		// If we weren't definitely loaded from live, and we can't view non-live content, we need to
-		// check to make sure this version is the live version and so can be viewed
-		$latestVersion = Versioned::get_versionnumber_by_stage($this->owner->class, 'Live', $this->owner->ID);
+		// check to make sure this version is the live version and so can be viewed.
+
+		// consider multiple- and single-stage versioning
+		$versionStage = count($this->stages) > 1 ? $this->liveStage : 'Stage';
```
tractorcow: I think if count($this->stages) < 2 then we don't even need to check the version. We probably just `return true` at that point. :D

bummzack: I just updated the PR with that change.

tractorcow:
Thank you @bummzack !

PR #5219, opened 2016-03-22
API Improvements to ownership api

tractorcow:
This introduces formal support for custom ownership relations, as well as making owned_by optional for standard DB relations. This should make usage in most real-world use cases (e.g. owning files) a lot simpler, and reduce the need to add custom extensions to core modules.

Part of this work requires the abliity to traverse relations on DataObjects where only one side (e.g. the has_one) is declared. I've implemented getNonReciprocalComponent to mock missing relations, although this can still be used even if both sides of the relation are declared.

Extension setOwner/clearOwner is now nested. This resolves issues with $this->owner getting set to null occasionally in extensions when extensions invoke methods recursively. Until now I was working around this with `$owner = $this->owner` at the start of every extension method, but this should be a more robust fix.

### DIFF
--- a/core/Extension.php
+++ b/core/Extension.php
@@ -12,6 +12,7 @@
  * @subpackage core
  */
 abstract class Extension {
+
 	/**
 	 * This is used by extensions designed to be applied to controllers.
 	 * It works the same way as {@link Controller::$allowed_actions}.
@@ -32,10 +33,12 @@ abstract class Extension {
 	protected $ownerBaseClass;
 
 	/**
-	 * Reference counter to ensure that the owner isn't cleared until clearOwner() has
-	 * been called as many times as setOwner()
+	 * Ownership stack for recursive methods.
+	 * Last item is current owner.
+	 *
+	 * @var array
 	 */
-	private $ownerRefs = 0;
+	private $ownerStack = [];
 
 	public $class;
 
@@ -55,6 +58,7 @@ abstract class Extension {
 
 	/**
 	 * Set the owner of this extension.
+	 *
 	 * @param Object $owner The owner object,
 	 * @param string $ownerBaseClass The base class that the extension is applied to; this may be
 	 * the class of owner, or it may be a parent.  For example, if Versioned was applied to SiteTree,
@@ -62,17 +66,32 @@ abstract class Extension {
 	 * would be 'SiteTree'.
 	 */
 	public function setOwner($owner, $ownerBaseClass = null) {
-		if($owner) $this->ownerRefs++;
+		if($owner) {
+			$this->ownerStack[] = $owner;
+		}
 		$this->owner = $owner;
 
-		if($ownerBaseClass) $this->ownerBaseClass = $ownerBaseClass;
-		else if(!$this->ownerBaseClass && $owner) $this->ownerBaseClass = $owner->class;
+		// Set ownerBaseClass
+		if($ownerBaseClass) {
+			$this->ownerBaseClass = $ownerBaseClass;
+		} elseif(!$this->ownerBaseClass && $owner) {
+			$this->ownerBaseClass = get_class($owner);
+		}
 	}
 
+	/**
+	 * Clear the current owner, and restore extension to the state prior to the last setOwner()
+	 */
 	public function clearOwner() {
-		if($this->ownerRefs <= 0) user_error("clearOwner() called more than setOwner()", E_USER_WARNING);
-		$this->ownerRefs--;
-		if($this->ownerRefs == 0) $this->owner = null;
+		if(empty($this->ownerStack)) {
+			throw new BadMethodCallException("clearOwner() called more than setOwner()");
+		}
+		array_pop($this->ownerStack);
+		if($this->ownerStack) {
+			$this->owner = end($this->ownerStack);
+		} else {
+			$this->owner = null;
+		}
 	}
 
 	/**
@@ -97,7 +116,4 @@ abstract class Extension {
 		return $parts[0];
 	}
 
-
-
 }
-
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -1626,7 +1626,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	/**
 	 * Find the foreign class of a relation on this DataObject, regardless of the relation type.
 	 *
-	 * @param $relationName Relation name.
+	 * @param string $relationName Relation name.
 	 * @return string Class name, or null if not found.
 	 */
 	public function getRelationClass($relationName) {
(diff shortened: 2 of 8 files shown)


model/DataObject.php
```diff
@@ -1647,6 +1648,121 @@ public function getRelationClass($relationName) {
...
+	 *  - Cannot be used on polymorphic relationships
+	 *  - Cannot be used on unsaved objects.
+	 *
+	 * @param string $remoteClass
+	 * @param string $remoteRelation
+	 * @return DataList|DataObject The component, either as a list or single object
+	 * @throws BadMethodCallException
+	 * @throws InvalidArgumentException
+	 */
+	public function getNonReciprocalComponent($remoteClass, $remoteRelation) {
```
stojg: I had to look up the word `Reciprocal`, still don't understand what it means in this context.

model/DataObject.php
```diff
@@ -1647,6 +1648,121 @@ public function getRelationClass($relationName) {
...
+		/** @var DataObject $remote */
+		$remote = $remoteClass::singleton();
+		$class = $remote->getRelationClass($remoteRelation);
+
+		// Validate arguments
+		if(!$this->isInDB()) {
+			throw new BadMethodCallException("getNonReciprocalComponent cannot be called on unsaved objects");
+		}
+		if(empty($class)) {
+			throw new InvalidArgumentException(sprintf("Invalid relation %s on s", $remoteRelation, $remoteClass));
```
stojg: can the exception message tell me what I can do to make it go away? i.e. what did I do wrong?

model/DataObject.php
```diff
@@ -1647,6 +1648,121 @@ public function getRelationClass($relationName) {
...
+		}
+		if($class === 'DataObject') {
+			throw new InvalidArgumentException("getNonReciprocalComponent doesn't work with polymorphic relations");
```
stojg: Given this exception message, I wouldn't understand what I did wrong.

model/DataObject.php
```diff
@@ -1647,6 +1648,121 @@ public function getRelationClass($relationName) {
...
+			$relations = Config::inst()->get($this->class, $type);
+			if($relations && isset($relations[$component])) {
+				return $type;
+			}
+		}
+		return null;
+	}
+
+	/**
+	 * Given a relation declared on a remote class, mock the component.
```
stojg: `Mock` as used in testing, i.g. I will get a component (or a components, it seems to be different from the @return) that are not the real objects, but behave like them?

stojg:
Sorry, you have just been randomly selected for a random peer review by someone who doesn't understand the feature fully :)

model/DataObject.php
tractorcow: If I buy you a coffee, and then the next day you buy me a coffee, that means you reciprocated the gift. It just means returned or repeated on both ends.

A has_many with a has_one on the other end is a reciprocated relationship.

A has_one with no has_many is a non-reciprocated relationship.

tractorcow: I think I should use 'reciprocative' not 'reciprocal' though.... it might be better grammar. :)

tractorcow:
Haha, good job @stojg . Glad to see you in open source. :)

model/DataObject.php
tractorcow: I've changed the message to `getNonReciprocalComponent invoked with invalid relation %s.%s`

model/DataObject.php
tractorcow:
```
throw new InvalidArgumentException(
	"getNonReciprocalComponent cannot generate component of relation %s.%s as it is polymorphic. " .
	"This method does not support polymorphic relationships"
);
```

tractorcow:
Rebased and updated based on feedback.

model/DataObject.php
```diff
@@ -1655,6 +1655,137 @@ public function getRelationClass($relationName) {
...
+	 *  - Cannot be used on polymorphic relationships
+	 *  - Cannot be used on unsaved objects.
+	 *
+	 * @param string $remoteClass
+	 * @param string $remoteRelation
+	 * @return DataList|DataObject The component, either as a list or single object
+	 * @throws BadMethodCallException
+	 * @throws InvalidArgumentException
+	 */
+	public function inferOppositeComponent($remoteClass, $remoteRelation) {
```
halkyon: what about `getOppositeComponent`? keep it consistent ;-)

tractorcow:
We decided to go back to `inferReciprocalComponent` :D

core/Extension.php
```diff
@@ -62,17 +66,32 @@ public static function add_to_class($class, $extensionClass, $args = null) {
...
 		$this->owner = $owner;
 
-		if($ownerBaseClass) $this->ownerBaseClass = $ownerBaseClass;
-		else if(!$this->ownerBaseClass && $owner) $this->ownerBaseClass = $owner->class;
+		// Set ownerBaseClass
+		if($ownerBaseClass) {
+			$this->ownerBaseClass = $ownerBaseClass;
+		} elseif(!$this->ownerBaseClass && $owner) {
+			$this->ownerBaseClass = get_class($owner);
+		}
```
hafriedlander: OwnerBaseClass doesn't get checked or restored properly by clearOwner later - problem?

model/versioning/Versioned.php
```diff
@@ -944,13 +944,113 @@ public function findOwned($recursive = true, $list = null)
...
+	 * Note that objects will only be searched in the same stage as the given record.
+	 *
+	 * @param bool $recursive True if recursive
+	 * @param ArrayList $list List to add items to
+	 * @param array $lookup List of reverse lookup rules for owned objects
+	 * @return ArrayList list of objects
+	 */
+	public function findOwnersRecursive($recursive, $list, $lookup) {
+		// First pass: find objects that are explicitly owned_by (e.g. custom relationships)
+		$owners = $this->findRelatedObjects('owned_by', false);
```
hafriedlander: If a relationship has an explicitly specified owned_by, I'm not sure we should also be using the inferred reciprocal relationship too. Imagine a class with a has_one referenced by an owns in Foo, but in Bar there's a custom relationship to act as the reciprocal referenced by owned_by.

core/Extension.php
tractorcow: I don't think it is; One base class many instances. :D

model/versioning/Versioned.php
tractorcow: The problem is when one core module has an ownership, and then user code wishes to add a new relation, such as a has_one to Image. You are essentially saying anything with a 'owned_by' can no longer be extended with more owners.

tractorcow: In your example, it would probably calculate the ownership twice, and the list would be the same set.

It does rely on `$object->Owned()->Owner() === $object` though, and if user code is doing custom relations that violate this, then they probably should be using custom relationships on both ends.

dhensby:
:hooray:

PR #5221, opened 2016-03-22
remove $CleartextPassword from ChangePasswordEmail.ss

christopherdarling:
4.x version of https://github.com/silverstripe/silverstripe-framework/pull/5194

### DIFF
--- a/templates/email/ChangePasswordEmail.ss
+++ b/templates/email/ChangePasswordEmail.ss
@@ -2,10 +2,6 @@
 
 <p>
 	<%t ChangePasswordEmail_ss.CHANGEPASSWORDTEXT1 'You changed your password for' is 'for a url' %> $AbsoluteBaseURL.<br />
-	<%t ChangePasswordEmail_ss.CHANGEPASSWORDTEXT2 'You can now use the following credentials to log in:' %>
-</p>
-
-<p>
-	<%t ChangePasswordEmail_ss.EMAIL 'Email' %>: $Email<br />
-	<%t ChangePasswordEmail_ss.PASSWORD 'Password' %>: $CleartextPassword
+	<%t ChangePasswordEmail_ss.CHANGEPASSWORDFOREMAIL 'The password for account with email address {email} has been changed. If you didn't change your password please change your password using the link below' email=$Email %><br />
+	<a href="Security/changepassword"><%t ChangePasswordEmail_ss.CHANGEPASSWORDTEXT3 'Change password' %></a>
 </p>


dhensby:
Line 5 really needs changing as it doesn't show the credentials any more, just email

This should probably just say something like "The password for account with email address $Email has been changed. If you didn't change your email please change your password (link to password reset form) urgently"

christopherdarling:
updated CHANGEPASSWORDTEXT2 with your suggestion and removed the output of Email on line 9 as it's now in the text of CHANGEPASSWORDTEXT2

dhensby:
`(link to password reset form)` needs to be an actual link ;)

Maybe easier with translations to say "using the link below" then include a link to `Security/changepassword`

christopherdarling:
Whoops! PR updated with link on new line.  One thought is from @jonom's comment on #5194 should we be displaying different text ('for account with email...') if Member.unique_identifier_field is not Email?

tractorcow:
I think it's probably fine; If you're using a custom field, you can also just as easily add a custom template too.

templates/email/ChangePasswordEmail.ss
```diff
@@ -2,10 +2,6 @@
 
 <p>
 	<%t ChangePasswordEmail_ss.CHANGEPASSWORDTEXT1 'You changed your password for' is 'for a url' %> $AbsoluteBaseURL.<br />
-	<%t ChangePasswordEmail_ss.CHANGEPASSWORDTEXT2 'You can now use the following credentials to log in:' %>
-</p>
-
-<p>
-	<%t ChangePasswordEmail_ss.EMAIL 'Email' %>: $Email<br />
-	<%t ChangePasswordEmail_ss.PASSWORD 'Password' %>: $CleartextPassword
+	<%t ChangePasswordEmail_ss.CHANGEPASSWORDTEXT2 'The password for account with email address {email} has been changed. If you didn't change your password please change your password using the link below' email=$Email %><br />
```
tractorcow: Please don't re-define translation strings re-using old keys; You'll need to create a new translation string instead.

All translations are shared across all versions, so this would cause this string to appear on all 3.x releases, unfortunately.

tractorcow:
One more problem to fix please. :)

templates/email/ChangePasswordEmail.ss
dhensby: Whoops, my bad.

christopherdarling:
Good point. All done :-)

PR #5223, opened 2016-03-23
API Add SS_Database::withTransaction for nice enclosed transactions

tractorcow:
@hafriedlander FYI

### DIFF
--- a/model/connect/Database.php
+++ b/model/connect/Database.php
@@ -489,6 +489,42 @@ abstract class SS_Database {
 	 */
 	abstract public function supportsTransactions();
 
+	/**
+	 * Invoke $callback within a transaction
+	 *
+	 * @param callable $callback Callback to run
+	 * @param callable $errorCallback Optional callback to run after rolling back transaction.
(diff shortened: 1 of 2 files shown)


model/connect/Database.php
```diff
@@ -489,6 +489,28 @@ public function affectedRows() {
...
 
+	/**
+	 * Invoke $callback within a transaction
+	 *
+	 * @param callable $callback Callback to run
+	 * @param callable $errorCallback Optional callback to run after rolling back transaction.
+	 * @param bool|string $transactionMode Optional transaction mode to use
+	 * @throws Exception
+	 */
+	public function withTransaction($callback, $errorCallback = null, $transactionMode = false) {
```
hafriedlander: As is, it's quite awkward for the caller to wrap a callback in a withTransaction block, but if the DB doesn't support transactions just use the callback directly.

We probably want an $errorIfTransactionsUnsupported parameter defaulting to false, and to check supportsTransactions on each step.

tractorcow:
I've added the suggested argument.

PR #5224, opened 2016-03-23
Location redirect issue

Andrew-Mc-Cormack:
When a redirect path is passed to the redirect method without a starting /, it is appended at the end of this method and added to the Location header list. However if the same path (without starting /) is passed again it throws an error as the paths are treated a different - one without starting / and one with starting /.

The user_error can be triggered on every user logout.

observed using v3.1 and subsites v1.0.8

### DIFF
--- a/control/Controller.php
+++ b/control/Controller.php
@@ -463,13 +463,20 @@ class Controller extends RequestHandler implements TemplateGlobalProvider {
 	 */
 	public function redirect($url, $code=302) {
 		if(!$this->response) $this->response = new SS_HTTPResponse();
-		
-		if($this->response->getHeader('Location') && $this->response->getHeader('Location') != $url) {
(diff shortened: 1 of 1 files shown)


dhensby:
Can you write a test for this, please? It seems a bit fragile to do this :/

control/Controller.php
```diff
@@ -463,13 +463,20 @@ public function popCurrent() {
...
-		
-		if($this->response->getHeader('Location') && $this->response->getHeader('Location') != $url) {
-			user_error("Already directed to " . $this->response->getHeader('Location')
-				. "; now trying to direct to $url", E_USER_WARNING);
-			return;
-		}
 
+		$location = $this->response->getHeader('Location');
+		if($location){
+
```
tractorcow: What if redirecting to protocol-less url? E.g. `$this->redirect('//google.com');`

PR #5230, opened 2016-03-26
Deprecate TextField::InternallyLabelledField() (closes #1684)

### LINKED ISSUE #1684: [2013-02-21] Deprecate TextField->InternallyLabelledField()
created by: @drzax (drzax)
created at: 2013-02-21
original ticket: http://open.silverstripe.org/ticket/8274

---

I can't think of a good reason for this function to exist. As far as I can tell it isn't used anywhere in CMS or Framework and promotes the very bad practice of using the value attribute as a method of adding placeholder text to a form field. 

An alternative to deprecation would be to populate the placeholder attribute instead which would make much more sense, but is a change to the API.

kinglozzer:


dhensby:
We need to remove it from master too ;)

PR #5232, opened 2016-03-27
Remove deprecated caching behaviour from ViewableData (closes #4063)

### LINKED ISSUE #4063: ViewableData::defineMethods() wasting memory?

kinglozzer:


tractorcow:
Tag with `API` please?

kinglozzer:
Done :)

Issue #5235, opened 2016-03-28
Debug should not be used for live responses


dhensby:
see: https://github.com/silverstripe/silverstripe-framework/blob/3.3.1/control/HTTPResponse.php#L273-L277

1. `Debug` should not be used for live responses
2. When using the framework without CMS, calling `Controller::httpError(404)` will result in a [`plain/text` ](https://github.com/silverstripe/silverstripe-framework/blob/3.3.1/control/HTTPResponse.php#L327-L328) response with an HTML body.

PR #5236, opened 2016-03-28
FIX Dont allow plain text friendly errors

dhensby:
Related to #5235

### DIFF
--- a/dev/Debug.php
+++ b/dev/Debug.php
@@ -340,6 +340,7 @@ class Debug {
 		if(Director::is_ajax()) {
 			echo $friendlyErrorMessage;
 		} else {
+			if(!headers_sent()) header('Content-Type: text/html');
 			if(class_exists('ErrorPage')){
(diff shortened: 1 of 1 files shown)


jonom:
This makes sense to me because a HTML response is being generated either way after this point. Just wondering though, there is a check to see whether headers have already been set but if they have been, is it possible that the type has already been set to plain text? So in the case of headers already being sent should we check what the content type was set to and dump a plain text message if appropriate?

dhensby:
> What change should be made in master, given that we are now using FriendlyErrorFormatter ?

It looks like `DebugViewFriendlyErrorFormatter@output` should be picking the content type there.

> there is a check to see whether headers have already been set but if they have been, is it possible that the type has already been set to plain text

So you __could__ use `headers_list()` and inspect them to see if the content-type has been set/sent and adjust output based on that - but we don't have means to change the output type of `Debug::create_debug_view()` at the moment

tractorcow:
Note; This is still needs fixing in master, but it doesn't need to block fixing in 3.x.

PR #5237, opened 2016-03-28
Include Bootstrap by default

chillu:
Includes bootstrap in the CMS by default, including the new `bootstrap-reboot.scss` to reset styles. This will eventually replace Eric Meyer's `framework/admin/scss/_reset.scss` reset styles, but for now they have a different scope (e.g. `_reset.scss` also resets `<fieldset>`, `<ul>`), so it'll need more work to remove that. Adjusted some CSS styles to deal with these slightly different resets.

Bootstrap itself had already been added previously (and documented in the upgrade guide), see https://github.com/silverstripe/silverstripe-framework/pull/5046 and https://github.com/silverstripe/silverstripe-framework/pull/5027. This is just getting it to the point where it can be included by all CMS classes without messing up styles. 

This is the first in a series of pull requests which prepare the new 4.0 asset-admin feature.

Merge together with https://github.com/open-sausages/silverstripe-asset-admin/pull/94

/cc @scott1702 @flashbackzoo


admin/code/LeftAndMain.php
```diff
@@ -454,7 +454,9 @@ public function init() {
 			Requirements::javascript(THIRDPARTY_DIR . '/jquery-entwine/src/jquery.entwine.inspector.js');
 			Requirements::javascript(FRAMEWORK_ADMIN_DIR . '/javascript/dist/leaktools.js');
 		}
-
+		
+		Requirements::css(FRAMEWORK_ADMIN_DIR . '/css/bootstrap/bootstrap-reboot.css');
```
scott1702: bootstrap.css includes bootstrap-reboot so this isn't necessary.

chillu: Sweet, fixed now

chillu:

kinglozzer:
Checked out these changes last night and couldn’t spot any regressions, so +1 from me

sminnee:
I've reviewed and I'm happy, with +1 from Loz I'm merging.

PR #5238, opened 2016-03-28
Build tooling cleanup

chillu:
There's more to come here, I'm basically trying to chunk up pull requests from the pending 4.0 asset-admin and campaign admin features as much as possible to make review easier.


chillu:

tractorcow:
I want to merge this, but I think I'll need you to summarise what's effectively changed by this pull request.

sminnee:
I've had a look over it and I'm happy to merge.

PR #5240, opened 2016-03-28
MINOR: Fix PHPUnit @covers assertions for DBStringTest

madmatt:
This resolves issues when PHPUnit is run strictly, where it exits when
it hits an @covers annotation that references a method that doesn't exist


gordonbanderson:
:)

PR #5241, opened 2016-03-29
BUG changing all cases of filesize spelling to file size

timkung:


chillu:
Language files other than `en.js` and `en.yml` will be overwritten the next time we pull in from transifex (see https://docs.silverstripe.org/en/3.3/contributing/translation_process/), but that's fine - translators always have the choice to spell things right or wrong through the transifex UI :)

PR #5242, opened 2016-03-29
Add example of blocking access to /admin by IP to security section

hailwood:
Honestly I'm pretty bad at .htaccess files so there may be a better way, but I thought I'd submit this and see where we get.


chillu:
Hmm, this advice isn't really specific to SilverStripe. I'm a bit cautious giving out security-related advice through the SilverStripe docs. As an example, in this case I'm not sure if it efficiently locks down URLs through `some-path/?url=admin/` and the `url` GET parameter used internally for routing. It also doesn't mention that its product-specific advice (Apache only), requires `AllowOverride`, and doesn't apply to IPv6. Given all of this, I'd rather leave this sort of docs to StackOverflow. 

@tractorcow what do you think?

hailwood:
@chillu I basically agree with you on all your points, I've more submitted it as a starting point for a discussion.

It is something that took me far to long to work out with a combination of googling, the IRC channel etc, and as you have pointed out it's not covering a few different use cases such as ?url=admin etc.

And after me I'm sure there will be bob, joe, jill, and susan who all want to do the same thing and most likely miss the same holes that I have.
Stackoverflow may be a decent solution if we can get a question going on there with a good set of answers covering Apache/Nginx etc and any entry points into the app so when the question comes up we have somewhere we can link users to get the information.

Could we use this discussion to work out some good defaults before we create the question if we go that route?

chillu:
You could ask on SO and answer your own question? ;) It sucks that there's no good place to collect SilverStripe "snippets" any more (what used to be leftandmain.com and http://www.ssbits.com/). Maybe somebody wants to start an unofficial wiki? You could revive this discussion on https://groups.google.com/forum/#!forum/silverstripe-dev.

As the core team, we can't really take on all those snippets into the core documentation since it's a long-term commitment to keep them working. So with that in mind, I'm going to close this pull request

wernerkrauss:
@chillu That's why Barry and me started www.silverstrip.es, to collect those snippets we find helping out in IRC every day. 

@hailwood if you want to write a little post I'm happy to either provide you access to silverstrip.es or you can send it to me and I'll publish it with your name mentioned.

chillu:
@wernerkrauss Oh, wasn't aware of http://www.silverstrip.es/, awesome! I've just posted it around on our company Slack :) Maybe post about the site on https://groups.google.com/forum/#!forum/silverstripe-dev, making more people aware of it?

hailwood:
Hey @wernerkrauss I'll ask a question on SO tonight about improvements on the blocking to see what I can muster up and then send you through something later during the week.

PR #5243, opened 2016-03-29
Implement client-side routing with Page.js

flashbackzoo:
Routing required to implement React based CMS sections.

- Removes thirdparty dependency History.js
- Adds thirdparty dependency Page.js to manage client-side routing
- Adds a wrapper around Page.js for SilverStripe specific behaviour
- Increased minimum browser requirement to IE10. Native HTML History API routing requires IE10 or newer (necessitated by removal of History.js)
- PJAX pannel loading via now uses promises rather than callbacks
- Adds getClientConfig method to LeftAndMain which can be used to pass config to the client application.

@chillu


chillu:
Note that the browser support requirement of IE10+ has been discussed and de-facto agreed on through https://groups.google.com/forum/#!topic/silverstripe-dev/ApaaWnlaHpM)

chillu:

admin/code/LeftAndMain.php
```diff
@@ -180,6 +180,50 @@ class LeftAndMain extends Controller implements PermissionProvider {
...
+		$combinedClientConfig = ['sections' => []];
+		$cmsClassNames = CMSMenu::get_cms_classes();
+
+		// Ordered by url_priority because the routing rules
+		// need to be registered in priority order on the client.
+		usort($cmsClassNames, function ($a, $b) {
+			$priorityA = Config::inst()->get($a, 'url_priority');
+			$priorityB = Config::inst()->get($b, 'url_priority');
+
+			if ($a == $b) {
```
kinglozzer: Shouldn’t this be `if ($priorityA == $priorityB) {`?

flashbackzoo: yeah it should

tractorcow: These are already sorted in AdminRootController::rules() method. I would just move the logic from there to just before the `return` in `CMSMenu::get_cms_classes`.

tractorcow:
Seems to have a bug trying to load the cms.

image

tractorcow:
I've added another commit, fixing the issue with window.history.state being null on initial load, and putting some of the sorting for menu items into CMSMain class.

flashbackzoo:
I think there could be something wrong with the bundles in that commit @tractorcow if you look at the lines added +38,671 LOC in `bundle-lib.js`

You probably just need to `npm run build` to minify it. Note you won't need to do this as soon :smile:

PR #5246, opened 2016-03-29
DOCS Session::save() needs to be called after Session::set()

muskie9:
I'm not sure if this is needed in earlier versions. If it does I can do additional pr's if needed.


kinglozzer:
As far as I’m aware, this is handled automatically at the end of the request lifecycle (unless that’s never reached due to an `exit()` or `die()` for example)... if it’s not then it’s probably a bug

muskie9:
@kinglozzer would a `Debug::show()` affect it as well?

muskie9:
Ignore that... turns out there was a die in my code I missed. The session seems to run the save like you said. Closing.

PR #5247, opened 2016-03-29
API Implement ChangeSets for batch publishing

tractorcow:
Includes fixes to SapphireTest, and update DataObject::can to include $context param.

CMS pr at https://github.com/silverstripe/silverstripe-cms/pull/1435

### DIFF
--- a/dev/SapphireTest.php
+++ b/dev/SapphireTest.php
@@ -982,22 +982,41 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 	/**
 	 * Create a member and group with the given permission code, and log in with it.
 	 * Returns the member ID.
+	 *
+	 * @param string|array $permCode Either a permission, or list of permissions
+	 * @return int Member ID
 	 */
 	public function logInWithPermission($permCode = "ADMIN") {
-		if(!isset($this->cache_generatedMembers[$permCode])) {
-			$group = Injector::inst()->create('Group');
+		if(is_array($permCode)) {
+			$permArray = $permCode;
+			$permCode = implode('.', $permCode);
+		} else {
+			$permArray = array($permCode);
+		}
+
+		// Check cached member
+		if(isset($this->cache_generatedMembers[$permCode])) {
+			$member = $this->cache_generatedMembers[$permCode];
+		} else {
+			// Generate group with these permissions
+			$group = Group::create();
 			$group->Title = "$permCode group";
 			$group->write();
 
-			$permission = Injector::inst()->create('Permission');
-			$permission->Code = $permCode;
-			$permission->write();
-			$group->Permissions()->add($permission);
+			// Create each individual permission
+			foreach($permArray as $permArrayItem) {
+				$permission = Permission::create();
+				$permission->Code = $permArrayItem;
+				$permission->write();
+				$group->Permissions()->add($permission);
+			}
 
 			$member = DataObject::get_one('Member', array(
 				'"Member"."Email"' => "$permCode@example.org"
 			));
-			if(!$member) $member = Injector::inst()->create('Member');
+			if (!$member) {
+				$member = Member::create();
+			}
 
 			$member->FirstName = $permCode;
 			$member->Surname = "User";
@@ -1007,9 +1026,8 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 
 			$this->cache_generatedMembers[$permCode] = $member;
 		}
-
-		$this->cache_generatedMembers[$permCode]->logIn();
-		return $this->cache_generatedMembers[$permCode]->ID;
+		$member->logIn();
+		return $member->ID;
 	}
 
 	/**
--- a/dev/SapphireTest.php
+++ b/dev/SapphireTest.php
@@ -670,29 +670,36 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 	 * @param mixed $dataObjectSet The {@link SS_List} to test.
 	 */
 	public function assertDOSEquals($matches, $dataObjectSet) {
-		if(!$dataObjectSet) return;
-
+		// Extract dataobjects
 		$extracted = array();
-		foreach($dataObjectSet as $item) $extracted[] = $item->toMap();
-
-		foreach($matches as $match) {
-			$matched = false;
-			foreach($extracted as $i => $item) {
-				if($this->dataObjectArrayMatch($item, $match)) {
-					// Remove it from $extracted so that we don't get duplicate mapping.
-					unset($extracted[$i]);
-					$matched = true;
-					break;
-				}
+		if($dataObjectSet) {
+			foreach ($dataObjectSet as $item) {
+				/** @var DataObject $item */
+				$extracted[] = $item->toMap();
 			}
+		}
 
-			// We couldn't find a match - assertion failed
-			$this->assertTrue(
-				$matched,
-				"Failed asserting that the SS_List contains an item matching "
-				. var_export($match, true) . "\n\nIn the following SS_List:\n"
-				. $this->DOSSummaryForMatch($dataObjectSet, $match)
-			);
+		// Check all matches
+		if($matches) {
+			foreach ($matches as $match) {
+				$matched = false;
+				foreach ($extracted as $i => $item) {
+					if ($this->dataObjectArrayMatch($item, $match)) {
+						// Remove it from $extracted so that we don't get duplicate mapping.
+						unset($extracted[$i]);
+						$matched = true;
+						break;
+					}
+				}
+
+				// We couldn't find a match - assertion failed
+				$this->assertTrue(
+					$matched,
+					"Failed asserting that the SS_List contains an item matching "
+					. var_export($match, true) . "\n\nIn the following SS_List:\n"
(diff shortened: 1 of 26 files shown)


model/DataObject.php
```diff
@@ -2796,10 +2796,11 @@ public static function has_own_table($dataClass) {
 	 * @param string $perm The permission to be checked, such as 'View'.
 	 * @param Member $member The member whose permissions need checking.  Defaults to the currently logged
 	 * in user.
+	 * @param array $context Additional $context to pass to extendedCan()
 	 *
 	 * @return boolean True if the the member is allowed to do the given action
 	 */
-	public function can($perm, $member = null) {
```
sminnee: Is adding `$context` a new API or are you merely formalising an existing API?

dev/SapphireTest.php
```diff
@@ -670,29 +670,36 @@ public function assertDOSContains($matches, $dataObjectSet) {
 	 * @param mixed $dataObjectSet The {@link SS_List} to test.
```
sminnee: I'm reviewing this and it seems like a cosmetic change unrelated to ChangeSet publishing. In future, I think it would be better if a change like this was broken out as a separate commit with a message that explains why you've done it. E.g. "PSR-2 compliance", "Restructured code for clarity", or "Fixed bug XYZ".

model/DataObject.php
tractorcow: It's formalising the additional argument on canCreate().

dev/SapphireTest.php
tractorcow: This is a bugfix to assertDOSEquals silently passing if $dataObjectSet is null.

tractorcow: I can put this in a separate commit.

sminnee:
Although I don't think this should be an optional add-on module, as it will be cleaner to make the versioning system depend on it and thereby only have 1 code path for managing versioned files, would it be better placed in the CMS module?

model/versioning/ChangeSet.php
```diff
@@ -0,0 +1,331 @@
...
+		if(!$this->isSynced()) {
+			throw new ValidationException(
+				"ChangeSet does not include all necessary changes and cannot be published."
+			);
+		}
+		if(!$this->canPublish()) {
+			throw new Exception("The current member does not have permission to publish this ChangeSet.");
+		}
+
+		DB::get_conn()->withTransaction(function(){
```
sminnee: Does `withTransaction()` actually create a transaction with our major DB backends?

tests/model/ChangeSetTest.php
```diff
@@ -0,0 +1,439 @@
...
+	];
+
+	/**
+	 * Automatically publish all objects
+	 */
+	protected function publishAllFixtures() {
+		$this->logInWithPermission('ADMIN');
+		foreach($this->fixtureFactory->getFixtures() as $class => $fixtures) {
+			foreach ($fixtures as $handle => $id) {
+				$this->objFromFixture($class, $handle)->doPublish();
```
sminnee: The difference between `SiteTree::doPublish()` and `ChangeSet::publish()` is a bit concerning.

In particular, it looks like a ChangeSet won't do anything that `SiteTree::doPublish()` does.

tractorcow:
Do you mean moving versioning altogether into CMS module?

I personally would rather see Versioned, ChangeSets, and the new admin altogether being put into a self-contained module, but for the time being we have kept everything together in framework. Splitting it across modules would make development and testing harder.

tests/model/ChangeSetTest.php
sminnee: Specifically, this will mean that pages published through a changeset will call neither `onBeforePublish()` nor `onAfterPublish()`. This seems like it will introduce bugs.

tractorcow: `doPublish()` is explicitly for publishing outside of changesets. it's the "automatic publish" feature.

`publish()` is the low level api that is used by both changesets and doPublish() that actually does all the work. The difference is that publish() can be used to publish in either direction between draft and live, and that it doesn't attempt to do recursive publishing.

ChangeSets provide the users an ability to group unrelated changes together in an atomic way, which doPublish() cannot do. It also acts as a "trusted" action, as all changes in a changeset are considered vetted and approved by a human operator, whereas doPublish() is automatic and has to make more decisions itself, thus is less "trusted".

model/versioning/ChangeSet.php
tractorcow: yes.

sminnee:
> I personally would rather see Versioned, ChangeSets, and the new admin altogether being put into a self-contained module, but for the time being we have kept everything together in framework. Splitting it across modules would make development and testing harder.

OK, I can buy that logic. Under that scheme, a module like `silverstripe/model-versioning` would be a dependency of `silverstripe/cms`, `silverstripe/assets`, and anything else that used versioned data.

My only question would be: is it worth shifting it into a new module _before_ adding this feature

tractorcow:
> OK, I can buy that logic. Under that scheme, a module like silverstripe/model-versioning would be a dependency of silverstripe/cms, silverstripe/assets, and anything else that used versioned data.

Yep that would be the plan. :D

> My only question would be: is it worth shifting it into a new module before adding this feature

I want to! Should we do this for alpha 2 though, so as to not distract from the work at hand? :P

tractorcow:
I've broken this PR into separate commits for each change @sminnee

tests/model/ChangeSetTest.php
tractorcow: Developers can use onAfterVersionedPublish, which takes both source and destination stages as arguments as well. This is unchanged from 3.x API.

tractorcow: The alternative I considered was to add a new parameter to doPublish($recursive) to flag whether or not related objects should be published. This would be left true (default) when doing single page publishing, but changesets could set it to false, since it handles related object publishing itself.

Would you prefer this api change to the one we're using?

tractorcow: I think renaming ChangeSet::publish to ChangeSet::doPublish makes sense too.

sminnee: `onAfterVersionedPublish()` is called for rollbacks as well as publishes.

 * If we're expecting people to always use that, we should deprecate `onAfterPublish()`.
 * If `onAfterPublish()` still has its uses, we should make it work with both ChangeSets and direct publishes.

hafriedlander:
I agree that whether the publish happens by calling `doPublish` directly on an `DataObject` or on a `ChangeSet`, `onBeforePublish` and `onAfterPublish` should be called.

After discussing with Damian, the core of the issue seems to be that `publish` _doesn't_ publish, it copies from one stage to another, and it used for both publishing and reverting.

I'd suggest renaming `publish` to `copyBetweenStages` or some such. The next bit depends a bit, but either:
- Introduce a new `publish` method that _does_ do the publish (by calling `copyBetweenStages`) and triggers `onBeforePublish` and `onAfterPublish` but never recurses, and keeping `doPublish` as exclusively recursive (nicer for any future refactoring to remove simple publishing and replace with automatic change-set creation).
- Just have `doPublish`, but introduce an argument to it that says whether to recurse or not.

I prefer the former, but it does make the simple publish call-chain a bit long (`doPublish` -> `publish` -> `copyBetweenStages`)

sminnee:
I'm fine with the former option, but `publish` for the the middle method isn't a helpful name. `nonRecursivePublish` is kind of verbose and ugly, but it at least says more about how it differs from `doPublish`. `publishItem` and `atomicPublish` are other ideas.

tractorcow:
Spoke with hamish, and I suggest:

- doPublish - Deprecate and invoke recursivePublish(). Prefixing "do" to an action doesn't really elaborate on anything.
- rename existing publish() implementation to copyToStage() (since first argument can be a version number, probably better than copyBetweenStage)
- publish() (new method with the old name) now does which does non-recursive publish. We can capture additional arguments and throw deprecation notice on code that used to call the old publish() method. This method will invoke onBeforePublish and onAfterPublish.
- create new method recursivePublish(), which invokes publish()

sminnee:
Cool.

tractorcow:
Hamish says maybe 'publishSingle' instead of 'publish' so we avoid the clash with the old name.
publish can be deprecated and invoke copyVersionToStage()

sminnee:
Yeah I'd agree with that.

tractorcow:
Done!

docs/en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -127,7 +127,7 @@ for this: `publish(<from-stage>, <to-stage>)`.
 	// and write a row to `MyRecord_versions`.
 	$record->write(); 
 	// will copy the saved record information to the `MyRecord_Live` table
-	$record->publish('Stage', 'Live');
+	$record->copyVersionToStage(Versioned::DRAFT, Versioned::LIVE);
```
hafriedlander: This doc change is incomplete (the publish method is still referenced a few lines above).

tractorcow: Fixed with better docs.

PR #5249, opened 2016-03-30
Standardise spelling of "customise"

### LINKED ISSUE #3988: $this->customize returns method not found error on Controller
I've encountered an issue when calling `$this->customize` on the Controller class.

It may be worth noting that calling `$this->customise` produces the desired effect.

There may be an issue with the spelling since the docs use American spelling but it seems that the UK spelling is used by the framework, thus producing the error.

Hope this helps. If needed, I can change the spelling in the docs or change the spelling in the framework.

Best regards,

John Dalangin

tractorcow:
Fixes #3988


tractorcow:
Sorry yanks! :D

SpiritLevel:
....and Canadians.....however, I am an ex-Canadian and have dropped 'z' and embraced 's' in order to properly integrate into NZ :P

tractorcow:
Really, I thought canadians used british spellings? :P

SpiritLevel:
Actually, Canadian spellings are a bit of a mash up between British and American. We spell "customize" like the yanks and "crystalline" like the Brits....we also have the French spellings working their way in....

http://www.lukemastin.com/testing/spelling/cgi-bin/database.cgi?action=view_category&database=spelling&category=C

PR #5252, opened 2016-03-30
Fixed bug where breadcrumb title wouldn't appear if title was blank

SilbinaryWolf:
Fixed bug where breadcrumb title wouldn't appear if the title was blank.
(This bug showed itself to me when I overwrote 'getTitle()' on my data objects)

### DIFF
--- a/forms/gridfield/GridFieldDetailForm.php
+++ b/forms/gridfield/GridFieldDetailForm.php
@@ -680,8 +680,9 @@ class GridFieldDetailForm_ItemRequest extends RequestHandler {
 
 		$items = $this->popupController->Breadcrumbs($unlinked);
 		if($this->record && $this->record->ID) {
+			$title = ($this->record->Title) ? $this->record->Title : "#{$this->record->ID}";
 			$items->push(new ArrayData(array(
-				'Title' => $this->record->Title,
+				'Title' => $title,
 				'Link' => $this->Link()
 			)));
 		} else {


kinglozzer:
We already do this in [`DataObject::getTitle()`](https://github.com/silverstripe/silverstripe-framework/blob/3/model/DataObject.php#L829). I’m not sure how I feel about how much responsibility should fall on the model for this... I think it’s perfectly valid to not want a title to be shown in some situations, and returning a falsy value definitely shouldn’t break the CMS layout or breadcrumbs.

Perhaps we’d be better served by adding a separate `getCMSTitle()` method or something? We’d only really be moving the problem (i.e. we create a `getCMSTitle()`, rather than forcing devs to create a separate `getFrontendTitle()` or something to work around this), but I’d be comfortable with that.

Thoughts?

SilbinaryWolf:
I like the idea of 'getCMSTitle' as I've wanted more control over the breadcrumb titles before for certain data objects.

I feel the name should probably be more accurate as to what it controls though, so maybe 'getCMSBreadcrumbTitle' is more suited because you'll immediately know what this function controls.

sminnee:
I'm a little nervous about polluting the core DataObject API with even more methods—it's already pretty crowded.

The general patterns of scaffolded UIs such as these breadcrumbs has been to:

 * take what information they can from the data model
 * where no information is available, provide some kind of default so that things don't look broken

This PR fits that general pattern, and so I'm pretty comfortable with the change proposed by this PR.

Yes, it makes assumptions about what should appear in the UI, but so does the scaffolding of the new item title (line 689). If people really want to tweak things, they could always go with the brute-force approach of subclassing `GridFieldDetailForm` and overriding `Breadcrumbs()`.

I feel like a better consideration of how to customise data model hints for various UIs would probably benefit from broader consideration and a new API rather than adding one-more-method.

For example, it might be more like a method like getLabelForContext($contexts), where new contexts can be defined by the application, and the caller can provide a cascade for DataObjects to choose from, based on what they have defined:

    `$dataObject->getLabelForContext(['BreadcrumbTitle', 'MenuTitle', 'Title'])

 * A data-object that specifies a `BreadCrumbTitle` will use that.
 * A `SiteTree` will use `MenuTitle`
 * By default data-objects will use title.

PR #5254, opened 2016-03-31
BUG Prevent live manifest loading files from nested test directories

tractorcow:
E.g. framework/admin/tests

I have no idea why, but 3.x doesn't seem to whinge about this so much.

At the moment trying to deploy a 4.0 site with --no-dev and --prefer-dist is impossible because certain classes can't be loaded properly (lack of phpunit means test classes are unloadable).


chillu:
Looks good to me, just letting the build pass before merge

PR #5255, opened 2016-03-31
BUG move test file to correct folder and fix class_exists

tractorcow:
To fix a bug where not including phpunit in a live deployment broke the site.


tractorcow:
Thank you.

PR #5256, opened 2016-03-31
BUG fix ErrorControlChain causing errors to be displayed if display_errors in php.ini is false

### LINKED ISSUE #5250: BUG Errors are displayed when thrown too early
I'm investigating an issue where errors are being displayed in live mode when they're thrown too early (ie. db connection issues). It also returns a 200 HTTP response instead of 500 so health check tools don't detect a failure.

A fix i've come up with is to change the following to suppress errors (true): https://github.com/silverstripe/silverstripe-framework/blob/3.2/main.php#L125

Is there a reason why this isn't the case already? What's the thought behind wanting to display errors by default?

tractorcow:
Fixes #5250


core/startup/ErrorControlChain.php
```diff
@@ -36,9 +36,21 @@ public function setErrored($error) {
...
+	 * - Errors cannot be un-suppressed if original mode dis-allowed visible errors
+	 *
+	 * @param bool $suppression
+	 */
 	public function setSuppression($suppression) {
 		$this->suppression = (bool)$suppression;
-		if ($this->handleFatalErrors) ini_set('display_errors', !$suppression);
+		// Don't modify errors unless handling fatal errors, and if errors were
+		// originally allowed to be displayed.
+		if ($this->handleFatalErrors && $this->originalDisplayErrors) {
```
silverstripe-elliot: Two valid options for display_errors are 'On' and 'Off'. `$this->originalDisplayErrors` will always be true in this case.

halkyon: Hmm, you're right. `(bool)` would mean this is always setting to true regardless of the setting. @tractorcow can you confirm this?

silverstripe-elliot: I can't see where in the code `this->suppression` gets set to the value obtained from `display_errors`? Seems like it's always true or false.  My comment is related to line 51, where the second condition would always evaluated to true

tractorcow: See https://github.com/silverstripe/silverstripe-framework/issues/5323

PR #5257, opened 2016-04-01
Changed CMS close from an 'x' to a silverstripe font icon

steve-silverstripe:
BEFORE:
image

AFTER:
image


tractorcow:
Thanks @steve-silverstripe :D

PR #5258, opened 2016-04-01
TinyMCE 4 optimisations. Use TinyMCE_Compressor.

halkyon:


chillu:
Thanks Sean, merged as b1b85b5, c2b6f61 and c4e5074. I've reinstated the `TEMP_FOLDER` use from the 3.x customisations on `tiny_mce_gzip.php` - using `sys_get_temp_folder()` might result in a different path than other SilverStripe caches, which complicates deployments and debugging. 

I've also set `disk_cache` back to `true`, and specifically marked it as a customisation in the file - otherwise the cached file was never used ;) I've also run `npm run build` over the dist files, rather than `npm run build --development` - the dev version creates unminified files which we don't want upstream.

You've also misunderstood the use of the tinymce references in `gulpfile.js` - those are still required because it copies tinymce files from the `framework/node_modules` folder into the webroot, behaviour we still need.

@tractorcow Sean and myself have discussed using either the `Requirements` class or `gulp` build tooling for this instead, neither approach was feasible. For `gulp`, you can't "unload" plugins by removing them from your `HtmlEditorConfig` (they apply procedurally on inclusion). And `Requirements` would need to be a lot smarter about TinyMCE specifics like plugin/lang/theme locations, minified file naming, etc - which didn't seem to be worth the effort at this point. Alternatively, we could've "forked" the `tiny_mce_gzip.php` minification logic into `HtmlEditorField`, but that seemed to increase the maintenance burden in case TinyMCE changes its file locations or naming conventions, for relatively little gain.

forms/htmleditor/TinyMCEConfig.php
```diff
@@ -409,20 +409,51 @@ protected function getEditorCSS() {
...
+	 * This ends up "pre-loading" TinyMCE bundled with the required plugins
+	 * so that multiple HTTP requests on the client don't need to be made.
+	 */
+	public function requireJS() {
+		require_once THIRDPARTY_PATH . '/tinymce/tiny_mce_gzip.php';
+
+		$useGzip = Config::inst()->get('HtmlEditorField', 'use_gzip');
+		$languages = array();
+
+		foreach(self::$configs as $configID => $config) {
```
tractorcow: $configs is private in a parent class... this works?

forms/htmleditor/TinyMCEConfig.php
```diff
@@ -409,20 +409,51 @@ protected function getEditorCSS() {
...
+				'url' => THIRDPARTY_DIR . '/tinymce/tiny_mce_gzip.php',
+				'plugins' => implode(',', array_keys($this->getPlugins())),
+				'themes' => $this->getTheme(),
+				'languages' => implode(',', array_filter($languages))
+			), true);
+			preg_match('/src="([^"]*)"/', $tag, $matches);
+
+			Requirements::javascript(html_entity_decode($matches[1]));
+		} else {
+			Requirements::javascript(THIRDPARTY_DIR . '/tinymce/tinymce.min.js');
```
tractorcow: Shouldn't this be needed even if useGzip is true, or does the gzip script include it automatically?

forms/htmleditor/TinyMCEConfig.php
```diff
@@ -409,20 +409,51 @@ protected function getEditorCSS() {
...
+
+		foreach(self::$configs as $configID => $config) {
+			$languages[] = $config->getOption('language');
+		}
+
+		// tinyMCE JS requirement
+		if($useGzip) {
+			$tag = TinyMCE_Compressor::renderTag(array(
+				'url' => THIRDPARTY_DIR . '/tinymce/tiny_mce_gzip.php',
+				'plugins' => implode(',', array_keys($this->getPlugins())),
```
tractorcow: You're only including array keys; If I register a plugin with a custom URL how will it be included?

tractorcow: I guess they are still included, just not via the minifier. :D

thirdparty/tinymce/tiny_mce_gzip.php
```diff
@@ -0,0 +1,385 @@
...
+
+// CUSTOM SilverStripe: Copied from Core.php
+if(!function_exists('getSysTempDir')) {
+	function getSysTempDir() {
+		if(function_exists('sys_get_temp_dir')) {
+			$sysTmp = sys_get_temp_dir();
+		} elseif(isset($_ENV['TMP'])) {
+			$sysTmp = $_ENV['TMP'];    	
+		} else {
+			$tmpFile = tempnam('adfadsfdas','');
```
tractorcow: What's this?

tractorcow:
I'll stop reviewing... @chillu has fixed these in the merge commits. ;D

chillu:
@halkyon Can you please explain the differences between `TinyMCEConfig->requireJS()` (https://github.com/silverstripe/silverstripe-framework/commit/b1b85b539e04926e0f02556d4f7c5834d44b1991) and the 3.x version of `HtmlEditorConfig::require_js()` (https://github.com/silverstripe/silverstripe-framework/blob/3/forms/HtmlEditorConfig.php#L306)?

halkyon:
@chillu thanks for fixing that up ;-) the difference is mostly the removal of the plugin configuration which isn't needed now due to changes in the way `TinyMCEConfig` passes the config through to the frontend via data attributes. The compressor now simply "preloads" TinyMCE and it's required plugins and languages.

tractorcow:
I've addressed my concerns at https://github.com/silverstripe/silverstripe-framework/issues/5261

Issue #5251, opened 2016-03-30
Low level URL parsing bug due to inconsistency.


patricknelson:
Hi, I pulled a little hair out trying to figure this one out. I tracked down a bug in `main.php` relating to how it determines the current URL. The problem is basically here when you have encoded entities in the URL and you're using `$_SERVER['REQUEST_URI']` to get the current URL instead of just pulling it out of the query string: https://github.com/silverstripe/silverstripe-framework/blob/3/main.php#L97

As you can imagine, the bug is due to the fact that the contents of the `url` parameter in the query string will be decoded for you automatically if you pull it from the `$_GET` super-global, but naturally variables inside of the `$_SERVER` super-global aren't treated the same way.

Ultimately, it expressed itself when I suited the following conditions:

- GridField with a name that has a space in it
- Setup `.htaccess` to not use `url=` but instead simply RewriteRule directly to `main.php`. I had to do this so that I'd be able to get the real source URI and not the rewritten one (since I have my own `RewriteRule`'s in `.htaccess`). So this may be a common problem.

So, the short term fix for me is to either:
- Update my gridfields to not have spaces for names (but the underlying issue will still be there and may arise from time to time)
- Manually tweak the parsing of `$url` to `urldecode()` where it pulls it from the request URI via `$_SERVER` super-global

Would this second option be a suitable fix at a framework level? It seems to work for me but I'm not sure if it's a good solution at the broader level for more use cases.

tractorcow:
Can you post the two strings you got from:
- `$_SERVER['REQUEST_URI']`
- `$_GET['url']`

In the given situations please, so that I may see how the strings are encoded differently?

patricknelson:
Sure.  Note that the actual raw URI being requested is: `/admin/pages/edit/EditForm/field/Internal%20Articles`.

- The value from `$_GET['url']` is `/admin/pages/edit/EditForm/field/Internal Articles`. This happens when `.htaccess` is configured as `RewriteRule .* framework/main.php?url=%1 [QSA]`
- The value from `$_SERVER['REQUEST_URI']` comes over as `/admin/pages/edit/EditForm/field/Internal%20Articles`. This happens when `.htaccess` is configured as `RewriteRule .* framework/main.php [QSA]`. 
[...]

Issue #3548, opened 2014-10-13
Namespace all classes


tractorcow:
This is on the agenda and in development.

See https://silverstripe.uservoice.com/forums/251266-new-features/suggestions/6425951-have-silverstripe-classes-be-in-namespaces

chillu:
@tractorcow has made a start on this in https://github.com/tractorcow/silverstripe-framework/commit/cd114a4e3015ac32402f2343ff0f144d88c96b44 a while back - just adding link here to make it easier to find, and avoid any double work.

sminnee:
PSR at #4562

sminnee:
Closing this a duplicate of #4562

PR #5259, opened 2016-04-01
FIX for #5251 to address minor URL decoding/parsing bug.

patricknelson:
See issue #5251

Ping @tractorcow

### DIFF
--- a/main.php
+++ b/main.php
@@ -94,6 +94,7 @@ if (isset($_GET['url']) && php_sapi_name() !== 'cli-server') {
 
 	// Lighttpd and PHP 5.4's built-in webserver use this
 } else {
+	// Get raw URL -- still needs to be decoded below (after parsing out query string).
 	$url = $_SERVER['REQUEST_URI'];
 
 	// Querystring args need to be explicitly parsed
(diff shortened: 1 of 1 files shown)


patricknelson:
p.s. Since so much logic is crammed here into `main.php` I'm not sure how I could apply good test coverage. Do you?

sminnee:
This seems like a reasonable patch-release change. It won't affect the behaviour of "ordinary" URLs, and for the URLs that it does effect, current behaviour is clearly broken.

Yeah—there's not a lot we can do about writing testing for this code without adding more infrastructure; solving that feels like a problem for a more significant refactor on `master`.

tractorcow:
Tested locally, and it seems fine.

patricknelson:
Yep. Maybe this can be thrown into the pile for `v4.0` :sunglasses:

tractorcow:
Better merge it up then...

patricknelson:
Haha, yes sir. Well, then again, I'd have to code it up first and that just seems like actual work. I'll keep it in mind, though, since I routinely encounter this `main.php` file and end up thinking to myself "Why?!"

patricknelson:
@tractorcow Re @sminnee comment from before:

> Yeah—there's not a lot we can do about writing testing for this code without adding more infrastructure; solving that feels like a problem for a more significant refactor on master.

I was thinking that a lot of the core framework files, particularly `main.php` and `Core.php` (among others) which make direct references to other files via `require_once()` statements point to a more fundamental issue which would relate tangentially to the test coverage of a bug like this. This is my thought process:

1. Find a way to restructure/abstract this code into classes, which requires
2. Having some basic autoloading so that the `require_once()` statements aren't necessary and files are no longer vulnerable to the special order of loading issues, which requires [possibly]
3. Doing some namespacing to assist with folder structure so that files are located properly, which ties into the mega-issue https://github.com/silverstripe/silverstripe-framework/issues/3548

Or am I thinking too broadly?

PR #5260, opened 2016-04-02
themedScript Added

tony13tv:
A function to call javascript files just by its name as it occurs with themedCSS

### DIFF
--- a/view/Requirements.php
+++ b/view/Requirements.php
@@ -188,8 +188,8 @@ class Requirements implements Flushable {
 	 * @param string $type  Comma-separated list of types to use in the script tag
 	 *                       (e.g. 'text/javascript,text/ecmascript')
 	 */
-	public static function themedScript($name, $module = null, $type = null) {
-		return self::backend()->themedScript($name, $module, $type);
+	public static function themedJavascript($name, $module = null, $type = null) {
+		return self::backend()->themedJavascript($name, $module, $type);
 	}
 
 	/**
@@ -1345,6 +1345,8 @@ class Requirements_Backend {
 			$this->css($theme . $css, $media);
 		} elseif($module) {
 			$this->css($module . $css, $media);
+		} elseif (Director::isDev()) {
+			throw Exception("The css file doesn't exists. Please check if the file $name.css exists in any context or search for themedCSS references calling this file in your templates.");
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


sminnee:
It's a reasonable addition to the api. For consistency my preference would be to call it `themedJavascript()`.

Also, ideally we'd have some tests for this, although I note that we don't yet have tests for `themedCSS()` either. Are you familiar with creating tests, @tony13tv? It would be awesome if you could get a basic test for both `themedCSS()` and `themedJavascript()`.

view/Requirements.php
```diff
@@ -1332,6 +1349,38 @@ public function themedCSS($name, $module = null, $media = null) {
...
+		$absproject = $absbase . $project;
+		$js = "/javascript/$name.js";
+
+		if(file_exists($absproject . $js)) {
+			$this->javascript($project . $js);
+		} elseif($module && file_exists($abstheme . '_' . $module.$js)) {
+			$this->javascript($theme . '_' . $module . $js);
+		} elseif(file_exists($abstheme . $js)) {
+			$this->javascript($theme . $js);
+		} elseif($module) {
```
tractorcow: I would prefer a check that the file exists here, and throw an Exception if the file cannot be loaded in any context.

tractorcow:
This needs unit tests please. :)

tractorcow:
If you need to mock BASE_PATH for testing you can use Director::baseFolder(), and then setup a 'testing site', as other tests do.

view/Requirements.php
tony13tv: That is a great addition! I will make the changes in both themedCSS and themedJavascript and will throw the exception only if the site is in dev mode.

tractorcow:
Thanks for the feedback on the error handling. How do you feel about having tests? :D

view/Requirements.php
tractorcow: If it's an error, then it should be thrown in all contexts, not only in dev mode.

view/Requirements.php
```diff
@@ -1328,6 +1345,42 @@ public function themedCSS($name, $module = null, $media = null) {
 			$this->css($theme . $css, $media);
 		} elseif($module) {
 			$this->css($module . $css, $media);
+		} elseif (Director::isDev()) {
+			throw Exception("The css file doesn't exists. Please check if the file $name.css exists in any context or search for themedCSS references calling this file in your templates.");
```
dhensby: I think an `InvalidArgumentException` makes most sense here.

dhensby:
I'm going to assume that if we want test coverage, we need to write it ourselves

view/Requirements.php
```diff
@@ -1328,6 +1345,42 @@ public function themedCSS($name, $module = null, $media = null) {
 			$this->css($theme . $css, $media);
 		} elseif($module) {
 			$this->css($module . $css, $media);
+		} elseif (Director::isDev()) {
```
dhensby: If we're throwing exceptions, they should always be thrown, not just in some cases

sminnee: As highlighted in #6373, this breaks behaviour that modules/sites were relying on: by silently ignoring missing files, modules used this to optionally include a file if a theme provided. So this is a semver failure.

PR #5261, opened 2016-04-03
BUG fix missing language on non-global TinyMCE Configs

tractorcow:
BUG Prevent external plugins from being passed to the minifier
API Change TinyMCEConfig::requireJS to getScriptURL() for testabilitiy

Fixes issues noted in https://github.com/silverstripe/silverstripe-framework/pull/5258

### DIFF
--- a/forms/htmleditor/TinyMCEConfig.php
+++ b/forms/htmleditor/TinyMCEConfig.php
@@ -34,6 +34,7 @@ class TinyMCEConfig extends HtmlEditorConfig {
 		'remove_script_host' => true,
 		'convert_urls' => false, // Prevent site-root images being rewritten to base relative
 		'menubar' => false,
+		'language' => 'en',
     );
 
 	/**
@@ -196,6 +197,24 @@ class TinyMCEConfig extends HtmlEditorConfig {
 	 */
 	public function getPlugins() {
 		return $this->plugins;
+	}
+
+	/**
+	 * Get list of plugins without custom locations, which is the set of
+	 * plugins which can be loaded via the standard plugin path, and could
+	 * potentially be minified
+	 *
+	 * @return array
+	 */
+	public function getInternalPlugins() {
+		// Return only plugins with no custom url
+		$plugins = [];
+		foreach($this->getPlugins() as $name => $url) {
+			if(empty($url)) {
+				$plugins[] = $name;
+			}
+		}
+		return $plugins;
 	}
 
     /**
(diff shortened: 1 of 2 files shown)


forms/htmleditor/TinyMCEConfig.php
```diff
@@ -416,31 +435,26 @@ protected function getEditorCSS() {
...
 	 */
-	public function requireJS() {
-		require_once THIRDPARTY_PATH . '/tinymce/tiny_mce_gzip.php';
-
+	public function getScriptURL() {
+		// If gzip is disabled just return core script url
 		$useGzip = Config::inst()->get('HtmlEditorField', 'use_gzip');
-		$languages = array();
-
-		foreach(self::$configs as $configID => $config) {
```
tractorcow: The current config might not belong to self::$configs, so this fix only loads its own language.

forms/htmleditor/TinyMCEConfig.php
```diff
@@ -416,31 +435,26 @@ protected function getEditorCSS() {
...
-			preg_match('/src="([^"]*)"/', $tag, $matches);
-
-			Requirements::javascript(html_entity_decode($matches[1]));
-		} else {
-			Requirements::javascript(THIRDPARTY_DIR . '/tinymce/tinymce.min.js');
-		}
+		require_once THIRDPARTY_PATH . '/tinymce/tiny_mce_gzip.php';
+		$tag = TinyMCE_Compressor::renderTag(array(
+			'url' => THIRDPARTY_DIR . '/tinymce/tiny_mce_gzip.php',
+			'plugins' => implode(',', $this->getInternalPlugins()),
```
tractorcow: Only minify plugins specified without a custom path, since you can't pass paths to the minifier.

forms/htmleditor/TinyMCEConfig.php
```diff
@@ -34,6 +34,7 @@ class TinyMCEConfig extends HtmlEditorConfig {
 		'remove_script_host' => true,
 		'convert_urls' => false, // Prevent site-root images being rewritten to base relative
 		'menubar' => false,
+		'language' => 'en',
```
tractorcow: Set a default here... maybe some logic could check if the current i18n locale is also a valid tinymce locale? Potential future improvement.

tractorcow:
FYI @chillu

tractorcow:
Also, my notes above are commentary.... I'm not critiquing my own PR. :)

forms/htmleditor/TinyMCEConfig.php
chillu: Just to clarify, this will be set to `i18n::get_tinymce_lang()` by `LeftAndMain->init()`. Which is a mapping of ISO locales to known tinymce language files, see `i18n::$tinymce_lang`. This functionality should probably be placed in `TinyMCEConfig`, but in general it works as intended

forms/htmleditor/TinyMCEConfig.php
chillu: I guess you could have a translation interface where different TinyMCE instances are shown in their respective native language on the same screen. But that seems pretty far fetched - any given author wants a view to be consistently in one language or another. So I think this simplification makes sense.

forms/htmleditor/TinyMCEConfig.php
tractorcow: Ok, that's a good idea... as much "tinymce specific" content should really be inside this class, and any interface for "set current editor language" should probably be abstracted to the config abstract interface.

Issue #4774, opened 2015-11-16
Injector service definitions should not inherit to subclasses


kinglozzer:
*sminnee says:*

Class Inheritance and Depedency Injection are two different extensibility paradigms. The labels in Injector should be thought of as service names, and should be decoupled from the class inheritance tree entirely.

------

Original:

I don’t really know what to make of this:

```yml
Injector:
  CoreService:
    class: MyCustomCoreService
```
```php
class CoreService extends Object {
}

class MyCustomCoreService extends CoreService {
}

class ServiceSubclass extends CoreService {
	public function foo() {
	}
}

// Object->__call(): the method 'foo' does not exist on 'MyCustomCoreService'
ServiceSubclass::create()->foo();
```

I understand what’s happening - Injector can’t find any configuration for `ServiceSubclass`, so looks to the parent class and uses that config - but I don’t really understand why. Injector was clearly built this way intentionally, am I misunderstanding/misusing it here?

Feel free to close if I’m being a doofus :)

dhensby:
This is really interesting.

Imagine you want to override `DropdownField` to use `CustomDropdownField` and you add some behaviour.

What if you then have another class extend `DropdownField` (`AnotherDropdownField`)? We can't make that inherit from our new custom class, though `Injector` has been told `DropdownField` shouldn't be used...

It's actually a really tricky issue to determine correct behaviour for... As we can't even force `AnotherDropdownField` to inherit the behaviour of `CustomDropdownField` maybe we shouldn't even try, rather than just return a `CustomDropdownField`.

sminnee:
The Injector is used to retrieve the objects corresponding to named services, and by default, the services are named the same as the class. In better/more-bloated architectures, the service is named according to an interface, rather than a class, and you load it with some implementor of that interface.

If you have a service named by a parent-class and you have requested a sub-class, it doesn't seem likely that you're wanting to grab a copy of the parent-class' service. So I'd call this a bug.

Even though it's a bug, since it seems possible that people might rely on the current behaviour, I'd put the fix into `master`.

----

The whole "inject CustomDropdownField into DropdownField" pattern doesn't mean "don't use DropdownField". What is means is "the DropdownField service is now provided by CustomDropdownField". It's easier for me to conceptualise if I imagine that I am injecting into a DropdownFieldInterface instead.

[...]

PR #5271, opened 2016-04-04
Fix 5244: Removed Email Subclasses

bummzack:
Removed Email Subclasses used by the Member class (`Member_ChangePasswordEmail` and `Member_ForgotPasswordEmail`).

Added a test for the forgot password email.
Improved the test for the change-password email.
It's not guaranteed that the `TestMailer` instance will remain the same across a unit-test, made `$emailsSent` a static member to fix this issue.

### DIFF
--- a/tests/security/MemberTest.php
+++ b/tests/security/MemberTest.php
@@ -185,16 +185,44 @@ class MemberTest extends FunctionalTest {
 	 * Test that changed passwords will send an email
 	 */
 	public function testChangedPasswordEmaling() {
+		$oldSetting = Config::inst()->get('Member', 'notify_password_change');
+		Config::inst()->update('Member', 'notify_password_change', true);
+
 		$this->clearEmails();
 
 		$member = $this->objFromFixture('Member', 'test');
 		$this->assertNotNull($member);
 		$valid = $member->changePassword('32asDF##$$%%');
 		$this->assertTrue($valid->valid());
-		/*
-		$this->assertEmailSent("sam@silverstripe.com", null, "/changed password/",
-		'/sam@silverstripe\.com.*32asDF##\$\$%%/');
-		*/
+
+		$this->assertEmailSent("dummy@silverstripe.tld", null, "Your password has been changed",
+			'/dummy@silverstripe\.tld/');
+
+		Config::inst()->update('Member', 'notify_password_change', $oldSetting);
+	}
+
+	/**
+	 * Test that triggering "forgotPassword" sends an Email with a reset link
+	 */
+	public function testForgotPasswordEmaling() {
+		$this->clearEmails();
+		$this->autoFollowRedirection = false;
+
+		$member = $this->objFromFixture('Member', 'test');
+		$this->assertNotNull($member);
+
+		// Initiate a password-reset
+		$response = $this->post('Security/LostPasswordForm', array('Email' => $member->Email));
+
+		$this->assertEquals($response->getStatusCode(), 302);
+
+		// We should get redirected to Security/passwordsent
+		$this->assertContains('Security/passwordsent/dummy@silverstripe.tld',
+			urldecode($response->getHeader('Location')));
+
+		// Check existance of reset link
+		$this->assertEmailSent("dummy@silverstripe.tld", null, 'Your password reset link',
+			'/Security\/changepassword\?m='.$member->ID.'&t=[^"]+/');
 	}
 
 	/**
--- a/dev/SapphireTest.php
+++ b/dev/SapphireTest.php
@@ -192,7 +192,7 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 		self::$is_running_test = true;
 
 		// i18n needs to be set to the defaults or tests fail
-		i18n::set_locale(i18n::default_locale());
+		i18n::set_locale(Config::inst()->get('i18n', 'default_locale'));
 		i18n::config()->date_format = null;
 		i18n::config()->time_format = null;
 
@@ -222,12 +222,6 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 		$fixtureFile = static::get_fixture_file();
 
 		$prefix = defined('SS_DATABASE_PREFIX') ? SS_DATABASE_PREFIX : 'ss_';
-
-		// Set up email
-		$this->originalMailer = Email::mailer();
-		$this->mailer = new TestMailer();
-		Injector::inst()->registerService($this->mailer, 'Mailer');
-		Config::inst()->remove('Email', 'send_all_emails_to');
 
 		// Todo: this could be a special test model
 		$this->model = DataModel::inst();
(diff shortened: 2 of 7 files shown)


dev/TestMailer.php
```diff
@@ -4,14 +4,14 @@
  * @subpackage email
  */
 class TestMailer extends Mailer {
-	protected $emailsSent = array();
+	protected static $emailsSent = array();
```
tractorcow: This is going to be problematic if running across multiple tests... creating and re-setting a new TestMailer won't flush all old emails.

I suggest keeping these as non-static if you can.

tractorcow:
Thanks, very good overall! Just a comment on the member variables.

dev/TestMailer.php
bummzack: I think that's what `clearEmails` is for (flushing emails, that is)? Since TestMailer isn't enforced to be a singleton during a test, you can get multiple instances of it, where one will have Emails stores and another one doesn't. This was the case for me when I triggered an Email via `FunctionalTest::post` (see my `testForgotPasswordEmaling` test). There seem to be two different instances of TestMailer being instantiated in that case and the Mailer used within the tests itself won't contain any mails, unless `$emailsSent` is a static member.

I think overall a static member makes sense, since you can't enforce the mailer instance via `Mailer::set_mailer`. Tests that send Emails should use `$this->clearEmails` at the start… I don't see any issues with this, unless you're starting to run tests in parallel (and even then, you can find your sent Email using the existing API).

bummzack: I investigated a bit further. It seems like using `Email::mailer()` will get you the correct instance, while `SapphireTest::$mailer` will return another (unpopulated) instance. That's really strange, since SapphireTest sets `Injector::inst()->registerService($this->mailer, 'Mailer');` in the startup, which should make them the same instance.

I guess it's time to start the debugger to see what's going on.

bummzack: Ok, the issue wasn't related to `FunctionalTest::post` at all (I thought it might have something to do with Injector nesting that happens within `Director::test`).

The `SapphireTest->mailer` instance gets already overridden/canceled-out within the `setUp` method whenever `self::create_temp_db()` is being called (which in turn calls `SapphireTest::resetDBSchema` that clears the service cache, via `Injector::unregisterAllObjects`).

Solving the problem is relatively simple, by just setting up the Mailer at the end of the `setUp` method. Will amend this PR with these changes and revert the changes made to `TestMailer` (eg. not using static members).

bummzack:
@tractorcow Apparently there was a GitHub outage. Could you re-run the CI to check if tests pass?

tests/security/MemberTest.php
```diff
@@ -185,16 +185,44 @@ public function testPasswordChangeLogging() {
...
+		$response = $this->post('Security/LostPasswordForm', array('Email' => $member->Email));
+
+		$this->assertEquals($response->getStatusCode(), 302);
+
+		// We should get redirected to Security/passwordsent
+		$this->assertContains('Security/passwordsent/sam@silverstripe.com',
+			urldecode($response->getHeader('Location')));
+
+		// Check existance of reset link
+		$this->assertEmailSent("sam@silverstripe.com", null, 'Your password reset link',
```
tractorcow: Ah, just one more thing... can you please change this to not be a real email address? I just want to avoid the case where someday down the line sam ends up with 10000 test emails from travis.

bummzack: Haha, sure thing. Maybe it would be a good idea to generally replace all "real" E-Mail Addresses from fixture-files?

dev/TestMailer.php
dhensby: Fantastic fix - I had this problem the other day and just gave up testing emails, I think there's even an issue open about this

dhensby:
OK - I'm not particularly happy with this because it actually makes these emails much more rigid than they used to be.

I'd assume the reason they are their own subclasses is because it allows better customisation, however there's no way here to change their behaviour (ie: no use of config or extension hooks)?

This is more a problem with the injector that should be fixed IMO rather than this (see #4774 where it's been decided that we shouldn't override all child classes). This means it's on the dev to override all subclasses individually if needed.

dhensby:
related: https://github.com/silverstripe/silverstripe-framework/pull/5000

bummzack:
What kind of customizations did you have in mind? The most common one is obviously overriding the template which is still possible.
For me, Emails are pretty much presentational and shouldn't contain any logic, or at least no logic that is specific to the member tasks being discussed here? As a developer you might want to have a more fine grained control when and under what circumstances these Emails are being sent, but this is out of the scope of an Email subclass, isn't it? Same goes for what data is being fed into the Email template… it's something that happens where the Email is instantiated, not in the Email class itself.

We could add an Injector "placeholder", something like this:

```yml
Injector:
  Member_ChangePasswordEmail:
    class: Email
```

And then use `Injector::inst()->create('Member_ChangePasswordEmail');` in code. This will still create an instance of `Email` by default, but could easily be swapped out for something else. As outlined before, I don't see much benefit in doing this, though.

And: Yeah, it seems like the issue #5000 is the same I encountered when writing tests for this fix.

dhensby:
You're right. But that doesn't mean the devs aren't doing it and without providing other means to allow them to, makes it more rigid. 

Your placeholder solution compounds the problem in #5000, surely? Unless the injector looks for the class that replaces email when insttantiating this?

bummzack:
All I'm saying is that issue #5000 would be fixed by this PR. 
The injector placeholder was just a suggestion to mitigate the drawbacks of having these Email subclasses removed and is not related to issue #5000 at all. Sorry if that was unclear.

bummzack:
And a note regarding Email subclasses: I think these exist, because in earlier versions of SilverStripe (2.2 and prior) didn't have a `setTemplate` method on `Email`, so subclassing was the only way of using a separate Template.

The shop module also deprecated [Email subclasses](https://github.com/silvershop/silvershop-core/blob/1.2.0/code/checkout/OrderEmail.php) in favor of using `setTemplate`.

My point is: These subclasses are here because of legacy reasons and no longer represent best-practice.

I'll leave it up to you ( @dhensby , @tractorcow anyone else? ) to decide if you want to keep these subclasses or not. Please let me know, so that I can amend this PR, because I think it would contain other useful improvements (such as improving Email testing). Cheers.

dhensby:
@bummzack OK - I've spoken to @kinglozzer and I think it's worth merging

I do think this is more rigid that it was before, but allowing custom templates should do away with most of the customisations people will need.

dhensby:
thanks

bummzack:
Great. From my point of view, issue #5000 could also be closed now. Should I create a separate PR for 3.x to fix the Email testing issue there as well?

dhensby:
@bummzack yes, please patch 3.x for that issue (even patching 3.2.x would make sense)

PR #5272, opened 2016-04-05
Set HTTP status code to 400 on AJAX validation error response

samwalshnz:
Would be good for form validation errors to respond with an appropriate 400 HTTP status code. This also means that it would trigger an error with javascript AJAX request.

### DIFF
--- a/forms/Form.php
+++ b/forms/Form.php
@@ -874,7 +874,7 @@ class Form extends RequestHandler {
 	/**
 	 * Set the target of this form to any value - useful for opening the form contents in a new window or refreshing
 	 * another frame
-	 *
+	 * 
 	 * @param string|FormTemplateHelper
 	 */
 	public function setTemplateHelper($helper) {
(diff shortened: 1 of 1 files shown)


tractorcow:
This seems like a change in existing API... although we cannot promise it will (or won't) break any existing features, it probably needs to go into the 3 branch for semver reasons.

Is there any reason we shouldn't be sending the same status code for text/html content type response?

https://www.quora.com/Which-HTTP-code-is-best-suited-for-validation-errors-400-or-422 suggests that maybe we should be using 422 for validation errors?

samwalshnz:
Would you like me to pull request to `silverstripe-framework:3` tag?

I think the same status code should be sent for text/html responses also.

I agree, a 422 HTTP Status code would make sense.

dhensby:
Aren't there issues with returning error codes for front-end validation; doesn't jQuery behave quite differently for error responses?

I'm pretty sure this would result in `jquery.load` breaking (basically anyone relying on `jQuery.success` being called) extracting the HTML response from the jQuery callbacks isn't as straight forward. You then need to extract the response text yourself as well as parse it correctly and so on.

It's going to have a huge impact on front end implementations if we respond with an error.

This also isn't a REST service, so I don't think we need semantically correct response codes to ajax requests like this; think about it, when you submit a form without ajax and there's a validation error, you don't throw a 400/422 at them, it's a 200. AJAX shouldn't get a different response code to normal requests.

This certainly needs to go against master and I would have the same concerns there too.

PR #5273, opened 2016-04-05
BUG Fix error when using search bar

tractorcow:
This fixes two problems:
- router.js was removing querystring args
- sspath.js was not properly decoding querystring args when converting to array format; As these array values are re-encoded when being combined again into the querystring, they were being double encoded.


tractorcow:
After merging please restart https://github.com/silverstripe/silverstripe-cms/pull/1435

chillu:
Merged this now, but needed another fix since this patch caused a follow-on regression: Searches with spaces were no longer working. Encoding is fun. 7337f26d3ae877dbfee288f868da408a3720a8ae

Issue #5276, opened 2016-04-05
BUG: Forward and Back button behaviour


flashbackzoo:
Navigating using the forward and back browser buttons is inconsistent.

__Steps to reproduce:__
1. Load the CMS at `admin/pages`
2. Navigate to the Reports section using the left menu
3. Navigate back using the browser's back button

__Expected:__
URL changes back to `admin/pages` and the pages section loads

__Actual:__
URL changes back to `admin/pages` but the section doesn't load.

tractorcow:
Is this only on the master branch? Can you confirm that the 3.x behaviour (with History.js) works as expected?

flashbackzoo:
3.3 works as expected, master has the broken behaviour.

Issue #5270, opened 2016-04-04
Duplicate active indicators on tabs (underlines)


tractorcow:
![image](https://cloud.githubusercontent.com/assets/936064/14239354/c90426c8-fa8e-11e5-924b-f0fbf486ce26.png)

Wrong indicators of tabs (the underlines): go to Security > Groups > click “Content Authors” > click “Permissions” tab > click “Groups” at the top of the page

Or go to `admin/security/groups/`

Going to `admin/security` and clicking groups doesn't change the URL.

Jira: https://silverstripe.atlassian.net/browse/OSS-1732

clarkepaul:
I don't see this, maybe it has been fixed?

tractorcow:
I've re-tested it with the above instructions. The issue is still there.

tractorcow:
Both tabs have the active class.

![image](https://cloud.githubusercontent.com/assets/936064/15847772/af906816-2cdc-11e6-8502-625f026d10bf.png)

dhensby:
I've experienced this too

Issue #5269, opened 2016-04-04
MINOR: Split lines not aligned horizontally in the CMS


tractorcow:
![image](https://cloud.githubusercontent.com/assets/936064/14239313/5d455d30-fa8e-11e5-8c43-de776ef99328.png)

Affects master branch.

Jira: https://silverstripe.atlassian.net/browse/OSS-1731

chillu:
Has been fixed now

Issue #5268, opened 2016-04-04
BUG Clicking on breadcrumbs just outside of the text links to admin/undefined


tractorcow:
Clicking on the page header redirects to undefined page in the CMS: `admin/undefined`

![image](https://cloud.githubusercontent.com/assets/936064/14239154/063e3706-fa8d-11e5-952f-20a0614a6b56.png)

Jira: https://silverstripe.atlassian.net/browse/OSS-1729

clarkepaul:
Tested this and couldn't find any issues as described, I think this has already been fixed maybe by #5278 ?

tractorcow:
Yep it's fixed, thanks.

PR #5278, opened 2016-04-05
Campaigns Feature

chillu:
@silverstripe/core-team @unclecheese @stevie-mayhew This is the first significant piece of React work we want to merge, and would appreciate all your feedback :)

# Overview

First cut of a UI for the [Neo-Versioning RFC](https://github.com/silverstripe/silverstripe-framework/issues/4932) feature - note that the model is called `ChangeSet`, but the user-facing terminology is `Campaign`. From a UI perspective, it simply lists the available campaigns. You can't publish a campaign, add pages or files to a campaign - all of which will follow in subsequent pull requests. The UI is based on @clarkepaul's designs at https://projects.invisionapp.com/boards/9GX68PTXUQFH/.

# Related Features

It's also the first implementation of the [Form Schema RFC](https://github.com/silverstripe/silverstripe-framework/issues/4938), with a stub implementation of a React-based `GridFieldComponent`. It's not meant to fully work with the PHP `GridField` yet, that's a whole different topic. It has just enough functionality to show the campaigns list through a hardcoded form schema JSON in `CampaignAdmin`. As the feature evolves, we'll have the schema generated by the PHP `GridField`. We'll also likely use a third party React component, so didn't spend too much time polishing the current `GridFieldComponent`.

Similarly, the `CampaignAdmin` PHP class creates the minimum API surface necessary to drive the feature. We'll need to have a discussion on how APIs work with GridField, how we can work towards a consolidated webservice approach in core, how we manage API state in the client (Relay+GraphQL?) - none of which will be solved by implementing this feature.

# Background

This is a large pull request with a lot of React and build tooling groundwork for the feature. It's been an unnaturally long-lived feature branch because we tried to get the feature "presentable". But we're at a point now where keeping the code away from master is doing more harm than good. I've spent many hours squashing and rebasing this, it's getting a bit too time intensive to split it up further. Particularly the ESLint rewrite changes virtually every file, hence it's hard to separate out. I'd recommend going through each commit individually to review, rather than looking at the final diff in Github.

# Remaining work

There's a bit of work left which Open Sourcerers are committing to addressing in the next weeks (before alpha1):
- Better documentation for build tooling and the React base layer
- Automatic ESLint on gulp builds (and via Travis)
- [Inconsistent back button navigation](https://github.com/silverstripe/silverstripe-framework/issues/5276)

And known bugs which we'll get around to before beta:
- https://github.com/silverstripe/silverstripe-framework/issues/5265
- https://github.com/silverstripe/silverstripe-framework/issues/5267
- https://github.com/silverstripe/silverstripe-framework/issues/5268
- https://github.com/silverstripe/silverstripe-framework/issues/5269
- https://github.com/silverstripe/silverstripe-framework/issues/5270

# How to run locally

You don't need to run an `npm install` or `npm run build` unless you want to change frontend code, but it'd be good to try and get it running anyway.

Since there's no way to create a Campaign just yet, here's a snippet from @tractorcow to create one.

```
class Page_Controller extends ContentController {

	private static $allowed_actions = array (
		'setup'
	);

	public function setup() {
		// You can include any CSS or JS required by your project here.
		// See: http://doc.silverstripe.org/framework/en/reference/requirements

		$set = ChangeSet::get()->filter('Name', 'Test')->first();
		if(!$set) {
			$set = new ChangeSet();
		}
		$set->Name = 'Test';
		$set->write();
		foreach(Page::get()->filter('ParentID', 0) as $page) {
			$set->addObject($page);
		}
		$set->write();

	}

}
```


madmatt:
> note that the model is called `ChangeSet`, but the user-facing terminology is `Campaign`.

Is there a reason for this? Seems unnecessarily complicated to have two different terms, and long-term will lead to the term 'ChangeSet' leaking to the frontend, or the term 'Campaign' becoming used as part of some new data model. I guess I see that 'Campaign' is a pretty common class name (and AFAIK full namespacing support hasn't landed yet), but not sure if there's other considerations?

tractorcow:
Because a campaign is the UX term for changeset + set of changeset items.

I don't have a problem with abstracting model + view in this regard; A little bit of independence is probably allowable. We might want to change completely from "campaign" in the future without changing the model, anyway.

tractorcow:
> Since there's no way to create a Campaign just yet, here's a snippet from @tractorcow to create one.

```php
// temp code - do not use
```

... somehow makes its way into official testing process anyway. :P

admin/code/CampaignAdmin.php
```diff
@@ -0,0 +1,408 @@
...
+		return Controller::join_links(
+			$this->Link('item'),
+			$itemID
+		);
+	}
+
+	/**
+	 *
+	 */
+	public function FindReferencedChanges() {
```
tractorcow: Probably need to remove this stub; It serves no purpose.

admin/code/LeftAndMain.php
```diff
@@ -451,7 +451,7 @@ public function init() {
 			]
 		]);
 
-		Requirements::javascript(FRAMEWORK_ADMIN_DIR . '/javascript/dist/bundle-leftandmain.js', [
+		Requirements::javascript(FRAMEWORK_ADMIN_DIR . '/javascript/dist/bundle-legacy.js', [
```
tractorcow: Why is this called legacy now?

tractorcow:
@chillu don't forget the CMS pr to go with this one.

sminnee:
It's a bit confusing to me why this is in `framework`, rather than `cms`?

tractorcow:
I think we discussed this offline, @sminnee ; Versioned + ChangeSet + Campaign Admin are all being kept together, rather than making versioning require CMS.

admin/code/LeftAndMain.php
chillu: Because it is. It contains all the entwine code, which is there for legacy reasons (we don't have time to recreate it all in React just yet), and shouldn't be modified without being aware of that fact. I'm not too fussed about "legacy" terminology, but `leftandmain` is wrong - other sections extend of the `LeftAndMain` PHP class (e.g. `CampaignAdmin`), and aren't powered by code in `bundle-leftandmain.js`

chillu:
> don't forget the CMS pr to go with this one

There isn't one, just a few build tooling fixes in `cms` which this PR isn't dependant on.

admin/code/CampaignAdmin.php
```diff
@@ -0,0 +1,408 @@
...
+				'editForm' => [
+					'schemaUrl' => $this->Link('schema/EditForm')
+				]
+			]
+		]);
+	}
+
+	public function schema($request) {
+		// TODO Hardcoding schema until we can get GridField to generate a schema dynamically
+		$json = <<<JSON
```
hafriedlander: Fairly minor, but slightly smelly to have JSON as a big string here. There doesn't seem to be any interpolation here, so this could be made either by json_encode'ing a PHP array structure, or by having a admin/code/CampaignAdmin.schema.json file. But if going away soon, not a critical fix.

admin/code/CampaignAdmin.php
```diff
@@ -0,0 +1,408 @@
...
+
+	/**
+	 * REST endpoint to create a campaign.
+	 *
+	 * @param SS_HTTPRequest $request
+	 *
+	 * @return SS_HTTPResponse
+	 */
+	public function createCampaign(SS_HTTPRequest $request) {
+		$response = new SS_HTTPResponse();
```
hafriedlander: We do this pattern a lot. Worth abstracting? (`new SS_HTTPResponse("application/json", ['campaign' => 'create']);` or `new SS_JSONResponse(['campaign' => 'create']);`. Again, could wait if we expect to replace.

admin/code/CampaignAdmin.php
```diff
@@ -0,0 +1,408 @@
...
+
+	/**
+	 * Get list contained as a hal wrapper
+	 *
+	 * @return array
+	 */
+	protected function getListResource() {
+		$items = $this->getListItems();
+		$count = $items->count();
+		$hal = [
```
hafriedlander: Similar to HTTPResponse, this is a common repeated pattern looking to be abstracted.

admin/code/CampaignAdmin.php
```diff
@@ -0,0 +1,408 @@
...
+			'_links' => [
+				'self' => [
+					'href' => $this->Link('items')
+				]
+			],
+			'_embedded' => ['ChangeSets' => []]
+		];
+		foreach($items as $item) {
+			/** @var ChangeSet $item */
+			$resource = $this->getChangeSetResource($item);
```
hafriedlander: No canView checking?

admin/code/CampaignAdmin.php
```diff
@@ -0,0 +1,408 @@
...
+	}
+
+
+
+	/**
+	 * Gets viewable list of campaigns
+	 *
+	 * @return SS_List
+	 */
+	protected function getListItems() {
```
hafriedlander: Icky name

admin/code/CampaignAdmin.php
```diff
@@ -0,0 +1,408 @@
...
+	 * REST endpoint to create a campaign.
+	 *
+	 * @param SS_HTTPRequest $request
+	 *
+	 * @return SS_HTTPResponse
+	 */
+	public function createCampaign(SS_HTTPRequest $request) {
+		$response = new SS_HTTPResponse();
+		$response->addHeader('Content-Type', 'application/json');
+		$response->setBody(Convert::raw2json(['campaign' => 'create']));
```
hafriedlander: This currently does nothing right? That needs a comment saying so, or it'll get forgotten.

admin/code/CampaignAdmin.php
```diff
@@ -0,0 +1,408 @@
...
+	 *
+	 * @param SS_HTTPRequest $request
+	 *
+	 * @return SS_HTTPResponse
+	 */
+	public function updateCampaign(SS_HTTPRequest $request) {
+		$response = new SS_HTTPResponse();
+		$response->addHeader('Content-Type', 'application/json');
+		$response->setBody(Convert::raw2json(['campaign' => 'update']));
+
```
hafriedlander: Needs todo comment.

admin/code/CampaignAdmin.php
```diff
@@ -0,0 +1,408 @@
...
+	 *
+	 * @param SS_HTTPRequest $request
+	 *
+	 * @return SS_HTTPResponse
+	 */
+	public function readCampaign(SS_HTTPRequest $request) {
+		$response = new SS_HTTPResponse();
+		$response->addHeader('Content-Type', 'application/json');
+		$response->setBody('');
+
```
hafriedlander: Needs todo comment.

admin/code/CampaignAdmin.php
```diff
@@ -0,0 +1,408 @@
...
+			return (new SS_HTTPResponse(json_encode(['status' => 'error']), 404))
+                ->addHeader('Content-Type', 'application/json');
+		}
+
+		if(!$record->canDelete()) {
+			return (new SS_HTTPResponse(json_encode(['status' => 'error']), 401))
+                ->addHeader('Content-Type', 'application/json');
+		}
+
+		$record->delete();
```
hafriedlander: canDelete

admin/javascript/src/boot/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+  // Set the initial config state.
+  configActions.setConfig(window.ss.config)(window.store.dispatch);
+}
+
+// TODO: This should be using `window.onload` but isn't because
+// Entwine hooks are being used to set up the <Provider>.
+// `window.onload` happens AFTER these Entwine hooks which means
+// the store is undefined when the <Provider> is constructed.
+$('body').entwine({ onadd: () => { appBoot(); } });
```
hafriedlander: No `{}` around single statement arrow expressions in AirBNB

hafriedlander: Also you probably shouldn't use arrow expressions with Entwine, although there's no harm here (`this` won't be want you think it is)

admin/javascript/src/components/form-action/index.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  render() {
+    const props = {
+      type: this.props.type,
+      className: this.getButtonClasses(),
+      disabled: this.props.disabled,
+    };
+
+    if (typeof this.props.id !== 'undefined') {
```
hafriedlander: More curious than complaining - is this required when using the JSX spread operator `{...props}` - that is, it's not clever enough to filter out keys with undefined values?

admin/code/CampaignAdmin.php
```diff
@@ -0,0 +1,408 @@
...
+		$hal = [
+			'_links' => [
+				'self' => [
+					'href' => $this->ItemLink($changeSetItem->ID)
+				]
+			],
+			'ID' => $changeSetItem->ID,
+			'Created' => $changeSetItem->Created,
+			'LastEdited' => $changeSetItem->LastEdited,
+			'Title' => $changeSetItem->getTitle(),
```
hafriedlander: Doesn't exist? Implicit ChangeSetItems are ephemeral.

admin/javascript/src/components/form-action/index.js
```diff
@@ -0,0 +1,119 @@
...
+   * @returns string
+   */
+  getButtonClasses() {
+    const buttonClasses = ['btn'];
+
+    // Add 'type' class
+    buttonClasses.push(`btn-${this.props.style}`);
+
+    // If there is no text
+    if (typeof this.props.label === 'undefined') {
```
hafriedlander: Why are we `typeof X === undefined`ing? Isn't `if (!this.props.label)` enough?

unclecheese: The label could conceivably be `0`, or perhaps empty, if it were just showing an icon applied via CSS.

admin/javascript/src/components/form-action/styles.scss
```diff
@@ -0,0 +1,124 @@
...
+	float: left;
+	margin: 0 4px 0 0;
+	height: 20px;
+	position: absolute;
+	left: 50%;
+	top: $btn-padding-y;
+	transform: translate(-50%);
+
+	svg {
+		width: 24px;
```
hafriedlander: Presume you've checked in IE? http://caniuse.com/#feat=svg says IE up to 11 doesn't scale properly.

admin/javascript/src/components/form-action/index.js
unclecheese: No, the spread operator will never delete keys. It's all driven by `hasOwnProperty`.

admin/javascript/src/components/form-action/index.js
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    // Add icon class
+    if (typeof this.props.icon !== 'undefined') {
+      buttonClasses.push(`font-icon-${this.props.icon}`);
+    }
+
+    // Add loading class
+    if (this.props.loading === true) {
+      buttonClasses.push('btn--loading');
```
unclecheese: Is the double hyphen deliberate?

admin/javascript/src/components/form-builder/index.js
```diff
@@ -0,0 +1,229 @@
...
+  },
+};
+
+export class FormBuilderComponent extends SilverStripeComponent {
+
+  constructor(props) {
+    super(props);
+
+    this.formSchemaPromise = null;
+    this.isFetching = false;
```
unclecheese: Shouldn't this be state?

admin/javascript/src/components/form-builder/index.js
```diff
@@ -0,0 +1,229 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+import * as schemaActions from 'state/schema/actions';
+import SilverStripeComponent from 'silverstripe-component';
+import FormComponent from 'components/form/index';
+import TextField from 'components/text-field/index';
+import HiddenField from 'components/hidden-field/index';
+import GridField from 'components/grid-field/index';
+import fetch from 'isomorphic-fetch';
```
hafriedlander: Why use isomorphic fetch? Are we planning on running this server side ever?

admin/javascript/src/components/form-builder/index.js
```diff
@@ -0,0 +1,229 @@
...
+    if (!schema) {
+      return null;
+    }
+
+    const formProps = {
+      actions: schema.schema.actions,
+      attributes: schema.schema.attributes,
+      data: schema.schema.data,
+      fields: schema.schema.fields,
+      mapFieldsToComponents: this.mapFieldsToComponents,
```
unclecheese: The repetitive naming here is a bit awkward. Can we name `const schema` something else for clarity? It's weird that `schema` is a property of a schema.

admin/javascript/src/components/form-builder/index.js
```diff
@@ -0,0 +1,229 @@
...
+   * @return object|null
+   */
+  getComponentByDataType(dataType) {
+    switch (dataType) {
+      case 'String':
+        return this.components.TextField;
+      case 'Hidden':
+        return this.components.HiddenField;
+      case 'Text':
+        // Textarea field (not implemented)
```
hafriedlander: Since this is just a placeholder anyway, makes sense to add these as we create them? I assume dataType isn't going to eventually be a fixed list? (Will check form RFC now, maybe it makes it clearer).

admin/javascript/src/components/form/index.js
```diff
@@ -0,0 +1,65 @@
...
+    const actions = this.getFormActionComponents();
+
+    return (
+      <form
+        id={attr.id}
+        className={attr.className}
+        encType={attr.enctype}
+        method={attr.method}
+        action={attr.action}
+      >
```
unclecheese: `<form {...attr}>` would allow user defined attributes.

admin/code/CampaignAdmin.php
chillu: Both `json_encode([])` and `CampaignAdmin.schema.json` would just give the false appearance that it's a good pattern to follow, and it's not. This code is temporary until we figure out how to serialise GridField metadata into our JSON schema :)

admin/code/CampaignAdmin.php
chillu: Not a fan of introducing a new API with `SS_JSONResponse`. `SS_HTTPResponse` already has a constructor signature which makes this a bit shorter (`($body = null, $statusCode = null, $statusDescription = null)`), and with PHP's new constructor bracketing we could remove the need for `$response` altogether. This is a lot of boilerplate which we should move into a generic API layer, but that's currently marked as optional for 4.0.0. I don't think it's worth polishing that code a lot, other than making sure its secure.

admin/javascript/src/components/form-builder/index.js
```diff
@@ -0,0 +1,229 @@
...
+   *
+   * @param boolean schema - If form schema data should be returned in the response.
+   * @param boolean state - If form state data should be returned in the response.
+   *
+   * @return object - Promise from the AJAX request.
+   */
+  fetch(schema = true, state = false) {
+    const headerValues = [];
+
+    if (this.isFetching === true) {
```
hafriedlander: Ick. Why don't ES6 promises have isFufilled()? They don't even have a reject. Ick.

admin/javascript/src/components/form-builder/index.js
```diff
@@ -0,0 +1,229 @@
...
+
+    const formProps = {
+      actions: schema.schema.actions,
+      attributes: schema.schema.attributes,
+      data: schema.schema.data,
+      fields: schema.schema.fields,
+      mapFieldsToComponents: this.mapFieldsToComponents,
+    };
+
+    return <FormComponent {...formProps} />;
```
hafriedlander: todo: Need to get someone to explain how this works to me.

admin/javascript/src/components/form-builder/index.js
```diff
@@ -0,0 +1,229 @@
...
+        ? fakeInjector.getComponentByName(field.component)
+        : fakeInjector.getComponentByDataType(field.type);
+
+      if (Component === null) {
+        return null;
+      }
+
+      // Props which every form field receives.
+      const props = {
+        attributes: field.attributes,
```
hafriedlander: Why not just pass all the field props through? Like `const {type, ...props} = field;`?

admin/code/CampaignAdmin.php
chillu: How do you mean? It's enforced two lines above the `delete()` call?

admin/javascript/src/boot/index.js
chillu: ESLint doesn't fail on this line, and I can't see any specific guidance one way or another in https://github.com/airbnb/javascript#arrow-functions. I've removed the `{}` though, looks nicer :). Leaving arrow expression in, since we're not using `this` - but thanks for the reminder.

admin/javascript/src/SilverStripeComponent.js
```diff
@@ -1,55 +0,0 @@
-/**
```
hafriedlander: This rename seems wrong. https://github.com/airbnb/javascript#naming--filename-matches-export

admin/javascript/src/components/form/index.js
```diff
@@ -0,0 +1,65 @@
...
+   */
+  getFormActionComponents() {
+    return this.props.actions.map((action) =>
+      <FormActionComponent {...action} />
+    );
+  }
+
+  render() {
+    const attr = this.props.attributes;
+    const fields = this.props.mapFieldsToComponents(this.props.fields);
```
hafriedlander: Expensive? Re-runs every re-render.

admin/javascript/src/components/grid-field/row.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import SilverStripeComponent from 'silverstripe-component';
+
+class GridFieldRowComponent extends SilverStripeComponent {
+
+  render() {
+    return (
+      <li className="grid-field-row-component [ list-group-item ]">{this.props.children}</li>
```
hafriedlander: Why are we using ul/li and not tables? Makes it impossible (or at least very difficult) to have a whole column be the right width later on.

hafriedlander: In fact we're even using display: table etc in the CSS. Table is clearly more semantic than ul.

admin/javascript/src/components/grid-field/table.js
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import SilverStripeComponent from 'silverstripe-component';
+
+class GridFieldTableComponent extends SilverStripeComponent {
```
hafriedlander: Why is this seperate from GridField itself?

admin/javascript/src/components/hidden-field/index.js
```diff
@@ -0,0 +1,48 @@
...
+  constructor(props) {
+    super(props);
+
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  render() {
+    return (
+      <div className="field hidden">
+        <input {...this.getInputProps()} />
```
hafriedlander: `<input {...this.props} className={['hidden', this.props.extraClass].join(' ')} type="hidden">`
or if https://github.com/JedWatson/classnames is included
`<input {...this.props} className={className(this.props.extraClass, 'hidden')} type="hidden">`

admin/code/CampaignAdmin.php
chillu: Damian fixed that in a later commit already: https://github.com/open-sausages/silverstripe-framework/commit/628cc2f3834f47fce926360530d15696d3058bb3

admin/javascript/src/components/form-builder/index.js
unclecheese: By using the spread operator, the schema becomes authoritative, composing the component any way it wants to. My preference would be for the form component to be composed explicitly as is.

unclecheese: Prefer immutable data structures:

`attributes: {...field.attributes},`

admin/javascript/src/components/form/index.js
```diff
@@ -0,0 +1,65 @@
...
+   * @return array|null
+   */
+  getFormActionComponents() {
+    return this.props.actions.map((action) =>
+      <FormActionComponent {...action} />
+    );
+  }
+
+  render() {
+    const attr = this.props.attributes;
```
unclecheese: `const attr = {...this.props.attributes};`

admin/javascript/src/components/grid-field/row.js
unclecheese: TR's are a real pain to drag-and-drop, IIRC.

admin/javascript/src/components/hidden-field/index.js
```diff
@@ -0,0 +1,48 @@
...
+
+  getInputProps() {
+    return {
+      className: ['hidden', this.props.extraClass].join(' '),
+      id: this.props.name,
+      name: this.props.name,
+      onChange: this.props.onChange,
+      type: 'hidden',
+      value: this.props.value,
+    };
```
unclecheese: This is probably a case where you should spread the props, so the user has a public api to the attributes..

admin/javascript/src/components/form-action/styles.scss
chillu: Good catch! I've checked it in IE10 and IE11, and it shows up without animation. I've created a bug for now, probably say its medium priority (loading animations are important user feedback): https://github.com/silverstripe/silverstripe-framework/issues/5283. Do you think it blocks the merge of this PR?

admin/javascript/src/components/hidden-field/index.js
hafriedlander: Oh, id is set to props.name. That makes it harder. I still feel like a callout to a seperate method is unneeded though.

admin/javascript/src/components/north-header-breadcrumbs/index.js
```diff
@@ -0,0 +1,42 @@
...
+      </div>
+    );
+  }
+
+  getBreadcrumbs() {
+    if (typeof this.props.crumbs === 'undefined') {
+      return null;
+    }
+
+    const breadcrumbs = this.props.crumbs.map((crumb, index, crumbs) => {
```
hafriedlander: This is super ganky.
```
breadcrumbs = [].concat(
 this.props.crumbs.slice(0,-1).map((crumb, index) => [
   <a key={index} className="cms-panel-link crumb" href={crumb.href}>{crumb.text}</a>,
   <span className="sep">/</span>,
 ]),
this.props.crumbs.slice(-1).map((crumb, index) =>  <span key={index} className="crumb last">{crumb.text}</span>
));
``

File: admin/javascript/src/components/form-action/index.js (hunk 17 above)
Comment (chillu) [inline admin/javascript/src/components/form-action/index.js]: I guess its a "modifier" in BEM speak, so I'd say its deliberate (http://getbem.com/naming/)

File: admin/javascript/src/components/text-field/index.js [hunk 34]
```diff
@@ -0,0 +1,55 @@
...
+  render() {
+    return (
+      <div className="field text">
+        {this.props.label &&
+          <label className="left" htmlFor={`gallery_${this.props.name}`}>
+            {this.props.label}
+          </label>
+        }
+        <div className="middleColumn">
+          <input {...this.getInputProps()} />
```
hafriedlander: Same comment as for HiddenField, seems over-engineered to have this split out. `{...props}` seems very idiomatic (I see unclecheese is basically saying the same thing :)).

admin/javascript/src/components/text-field/index.js
```diff
@@ -0,0 +1,55 @@
...
+      id: `gallery_${this.props.name}`,
+      name: this.props.name,
+      onChange: this.props.onChange,
+      type: 'text',
+      value: this.props.value,
+    };
+  }
+
+  handleChange() {
+    if (typeof this.props.onChange === 'undefined') {
```
hafriedlander: Again, why typeof == undefined, and not just falsy?

admin/javascript/src/reducer-register.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * The register of Redux reducers.
+ * @private
+ */
+const register = {};
+
+/**
+ * The central register of Redux reducers for the CMS.
```
hafriedlander: for the Admin Section. Should it be for _all_ admin code, or just framework / cms module?

admin/javascript/src/sections/campaign-admin/tests/campaign-admin-test.js
```diff
@@ -0,0 +1,5 @@
+jest.dontMock('../controller');
+
+describe('CampaignAdminContainer', () => {
+
+});
```
hafriedlander: Empty tests make baby kittens cry :(.

javascript/src/UploadField.js
```diff
@@ -430,13 +430,14 @@ $.entwine('ss', function($) {
 			}
 
 			if (iframe.attr('src') == 'about:blank') {
+				var disabled = this.siblings();
```
hafriedlander: Why? Some ES6 lint thing?

admin/code/CampaignAdmin.php
hafriedlander: Ah yes, over-eager. Carry on.

admin/javascript/src/boot/index.js
hafriedlander: > 8.2 If the function body consists of a single expression, omit the braces and use the implicit return.
I doubt ESLint is able to exhaustively check AirBNB syntax, there's a bit of higher semantic knowledge needed to follow it.

BTW, from below, I like this pattern more (no need to change here)

```
$('body').entwine({ 
  onadd() { appBoot(); } 
});
```

admin/javascript/src/components/form-action/index.js
hafriedlander: Makes less sense for the one just below though (font-icon-0 I can just about see). If this is standard idiom for React, fine, it's just a lot slower, and does odd things if set to unexpected falsey values (font-icon-null).

admin/javascript/src/components/form-action/styles.scss
hafriedlander: No, probably fine, as long as it sort of works.

admin/javascript/src/components/form-builder/index.js
hafriedlander: > By using the spread operator, the schema becomes authoritative, composing the component any way it wants to.
Yeah, that was my point too - why wouldn't the schema with a generic field be able to pass additional props down to a custom component?

admin/javascript/src/components/form-builder/index.js
hafriedlander: Since `schema` is only used in the if block above,
```const schema = this.props.schemas[this.props.schemaUrl], schema = schema && schema.schema;```

Comment (hafriedlander) [inline admin/javascript/src/components/form-builder/index.js]: (This works with var, const might complain :)).

File: admin/javascript/src/components/form/index.js (hunk 31 above)
Comment (hafriedlander) [inline admin/javascript/src/components/form/index.js]: attr is never mutated, and making copies all the time is surely expensive. Is this a pre-emptive protection against possible code change that mutates?

File: admin/javascript/src/components/form-builder/index.js (hunk 19 above)
Comment (chillu) [inline admin/javascript/src/components/form-builder/index.js]: Might've been a bit too much future proofing, agreed. I don't see us server-side rendering SS4 views via NodeJS anytime soon. The important bit was using the new `fetch()` API standard, rather than `jQuery.ajax()`. How about we just use the https://github.com/github/fetch polyfill which `isomorphic-fetch` is based on?

File: admin/javascript/src/components/text-field/index.js (hunk 35 above)
Comment (unclecheese) [inline admin/javascript/src/components/text-field/index.js]: Agreed. This could be more easily expressed as `this.props.onChange && this.props.onChange()`

File: admin/javascript/src/state/config/actions.js [hunk 39]
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * @param object config
+ */
+export function setConfig(config) {
+  return (dispatch) =>
+    dispatch({
+      type: ACTION_TYPES.SET_CONFIG,
+      payload: { config },
+    });
+}
```
unclecheese: Does this need to be a thunk?

admin/javascript/src/components/form-builder/index.js
chillu: I've had another look, and `isomorphic-fetch` doesn't actually add any code on top of `whatwg-fetch`, it simply makes it consumable by browserify in CommonJS format. So we'll stick with that for now.

admin/javascript/src/components/form-builder/index.js
chillu: I've renamed it to a more accurate `formSchema`. This is a follow-on effect of calling something "schema" at the endpoint level, but then allowing both a "schema" and "state" key as part of the response.

admin/javascript/src/components/form-builder/index.js
chillu: Agreed, I've removed them.

admin/javascript/src/components/form-builder/index.js
chillu: Avoiding multiple async calls will likely shift down into `silverstripe-backend`, with the component passing in a promise, but not directly adding promise resolution callbacks. Instead, it'll listen to state updates via `mapStateToProps`. See Sam's asset refactoring: https://github.com/silverstripe/silverstripe-asset-admin/pull/109/files

admin/javascript/src/components/form-builder/index.js
chillu: `attributes: {...field.attributes}` isn't actually part of the ES6 spec yet, I had to enable "Stage 2: Draft" in Babel for it to parse (http://babeljs.io/docs/plugins/#presets). And even then, I can't get ESLint to parse it. So in short, it's probably a bit too early to use the syntax like that. I've resorted to the `deep-freeze` library we already have in the codebase for other purposes.

admin/javascript/src/components/form/index.js
chillu: Since there's only a limited number of props passed in here which can lead to state tree mutations, I think that's acceptable in this case. Also, don't really know how to cache this outside of `render()` - against a nested object identity of `this.props.fields`? A mutation in that prop is the most likely bit to trigger the re-render in the first place, so will already have the same cache logic.

admin/javascript/src/components/form/index.js
chillu: Yep good stuff, implemented that

admin/javascript/src/components/grid-field/row.js
chillu: @scott1702 can you explain please? I had the same concerns as Hamish. From memory, you wanted to ensure that tables can be rearranged responsively (e.g. through http://v4-alpha.getbootstrap.com/content/tables/#responsive-tables). I also think `<table>` is more semantic in this case, rather than `<ul>/<li>` with random `<div>` scattered in them as "cells"

admin/javascript/src/components/grid-field/table.js
chillu: A "grid field" is more than just a table, there's button rows, etc. Also, `GridField` should really be called `GridFieldContainer` - it's the "smart" component which hooks into Redux state management, and instanciates a "dumb" `GridFieldTableComponent`.

I wouldn't overthink the current component structure, my hope is that we'll replace this with a third party component like  https://github.com/GriddleGriddle/Griddle (there's even a PoC for SS already: https://github.com/Tom-Alexander/silverstripe-griddle)

admin/javascript/src/components/hidden-field/index.js
chillu: Sounds like a good improvement, not essential for this PR though.

admin/javascript/src/components/text-field/index.js
chillu: Again, don't want to polish these components too much because they'll likely be replaced by bootstrap ones: https://react-bootstrap.github.io/components.html#forms

admin/javascript/src/reducer-register.js
chillu: Since there can only be one "root reducer" in Redux that's composed of other reducers, it makes sense to only have one central `ReducerRegister` for any admin code that wants to influence state changes, regardless of module. So the documentation is correct stating "for the CMS".

admin/javascript/src/state/config/actions.js
chillu: Good catch, I've fixed that both for the `config` and `schema` actions - got a bit carried away with the async pattern there hehe.

javascript/src/UploadField.js
chillu: @scott1702 I was wondering the same thing actually. It wasn't linting, but rather something around the asset work which snuck into this PR

admin/code/CampaignAdmin.php
chillu: Fixed

admin/code/CampaignAdmin.php
chillu: Added comments.

admin/javascript/src/components/form-builder/index.js
chillu: Good catch, I've converted it to state. As mentioned in another comment around here, the whole idea of a component handling its own ajax loading and load status management is probably going away, in favour of a more centralised data and request/promise management around `silverstripe-backend`

admin/javascript/src/SilverStripeComponent.js
chillu: Yeah, we followed the WP Calypso directory and file naming conventions here, which makes sense for "components with sub-components in the same folder". I slightly prefer the "after" variation below (complying with AirBnB), but the folder/file name duplication still irks me. Can we merge this without a major file rename please, and defer the discussion? The PR is already big enough ;)

```
// before
src/
  components/
    grid-field/
      index.js
      header.js
      styles.css
      README.md

// after
  components/
    GridField/
      GridField.js
      GridFieldHeader.js
      styles.css
      README.md
```

admin/javascript/src/components/north-header-breadcrumbs/index.js
chillu: Yep, using your code now.

admin/javascript/src/sections/campaign-admin/tests/campaign-admin-test.js
chillu: Yeah, the container isn't doing very much, all the meat is either in GridField, or yet to be built (preview, item list, etc). We could have more test coverage, but we're also still in the experimentation stage as you can see from the discussions on this pull request - so trying to strike a responsible balance between unit test coverage and fast refinement of the architecture.

chillu:
@hafriedlander @unclecheese I think I've responded to all of your feedback, please let me know if anything is missing. I've added a few commits to the PR to address the specifics. Figured that's easier than merging it into the respective base commits.

admin/javascript/src/components/form-builder/index.js
unclecheese: @chillu I get that nothing that doesn't affect the UI should be stored as state, but don't you think it's conceivable that the user might want to render UI state based on `isFetching === true`?

admin/javascript/src/components/form/index.js
unclecheese: That's my bad. I missed the next line where this is used as a spread in `<form {...attr}>` ... or maybe that's new, but either way, not needed anymore.

hafriedlander:
OK, seems good to merge to me. There's a few idiomatic things that aren't best practise, but I'll leave that for later PRs to clean up. Specifically, these are added to internal TODO:
- Filenames wrong by AirBNB standard
- Trust the splat
- Don't typeof == undefined unless you need to

stevie-mayhew:
Sorry I didn't have time to review this as asked @chillu 

Really quick merge on a huge feature, I figured I would have time to go through it on the weekend.

sminnee:
@stevie-mayhew it would still be great to get your feedback on the work over the weekend or next week, I don't think anything that has been merged is set in concrete.

kinglozzer:
Firstly, apologies for not raising what follows earlier. I admit I’d neglected this particular PR as much of the JavaScript goes way over my head, but I should’ve commented on the overall concepts at play here sooner.

---

Are there any plans to abstract `CampaignAdmin` out of framework? Thinking back through all of the framework-only projects I’ve done, I can think of only one that would have any use for this admin UI. I know we can probably hide the menu item, but still...

> Versioned + ChangeSet + Campaign Admin are all being kept together, rather than making versioning require CMS.

I seem to remember a while back it was said that splitting out versioning into its own module wouldn’t be feasible, is that still the case? Even if `Versioned` has to remain in framework, how heavily tied to changesets + campaigns is it? Could the changesets model + the campaigns UI be split off and made a requirement of CMS? Should the CMS even _require_ this model + UI anyway?

There are a lot moving parts here, this is how I see the “ideal” component structure in my head:

- `silverstripe-framework`
- `silverstripe-versioned` - requires `silverstripe-framework`
- `silverstripe-changesets` - requires `silverstripe-framework` and `silverstripe-versioned`
- `silverstripe-cms` - requires `silverstripe-versioned`, `silverstripe-changesets` optional

Of course, it’s easy for me to sit here and say “this is how I’d like it to work” with no consideration for the amount of effort it would take, or even if it’s entirely possible (please feel free to call me out on why the above can’t happen). There has been so much awesome work that’s gone into all of this that’s really appreciated, I’m just a little worried that framework is becoming more of a monolith, rather than less of one!

Anyway, my immediate concern is that enabling `CampaignAdmin` by default in framework seems wrong. Versioning, while present in framework, is opt-in and this should really be too. Unless I’m mistaken, it’s opt-in at a model level so it follows the UI should be the same, rather than just appear empty.

dhensby:
24 hours from opening to merge gives us volunteers very little time to respond :/

javascript/src/UploadField.js
scott1702: Ah yes, this was because of 'use strict'; the variables just have to be declared properly.

admin/javascript/src/components/grid-field/row.js
scott1702: @clarkepaul and I created this structure based on some responsive designs that would've relied on quite a bit of JavaScript to implement if we used a html table structure, whereas with ul/li's we can do it all with css + breakpoints to match the designs in mobile, hence the `display: table` in CSS which would be overridden on mobile.

chillu:
@kinglozzer @stevie-mayhew @dhensby Still very keen on your feedback. We're trying to make smaller PRs which are less of a blocker for further work. You can also follow along at https://github.com/open-sausages/silverstripe-framework/branches

> Are there any plans to abstract CampaignAdmin out of framework?

My take on it was that by discussing the "changeset" RFC in context of a core feature, and merging that as such, the logical next step was creating a UI for it in framework as well. So technically speaking, those concerns should've been raised in the context of that RFC already. 

@sminnee also had a preference for both model and UI to be part of core, with the reasoning that it should be straightforward for devs to use changesets in the existing content publication workflows. I think the concept will become more important as more "content blocks" modules leverage the feature in SS4. 

I'd like to further split up the monolith as well, but as you correctly pointed out, particularly splitting out `Versioned` does cause extra work and complexity in supporting multiple variations. I think we should revisit this discussion once devs had a bit more exposure to the changesets feature.

> Anyway, my immediate concern is that enabling CampaignAdmin by default in framework seems wrong. 

Agreed, without `AssetAdmin` and `CMSMain` there might not be any models with `Versioned` attached to it, hence no need for `CampaignAdmin` to be visible. We should enable it by default in the `cms` though.

admin/javascript/src/components/grid-field/row.js
chillu: @scott1702 I assume you're referring to the small width table in https://projects.invisionapp.com/boards/9GX68PTXUQFH#/1547737/73357185? I think that's still a table actually even on the mobile breakpoints - it just exposes another piece of content ("20 changes") underneath the title. What requires a lot of JavaScript here?

clarkepaul: We wanted the view to be flexible for all gridfield type views as well, so they have the ability to become responsive without having to be forced to remove data. The idea was that the most important secondary data could sit below the title inline (dates, user, type...). 

@chillu so do you think its better to display all fields in the one cell for smaller devices, it would make presentation easier but then it doesn't make using tables as semantic for mobile anyway.

kinglozzer:
> My take on it was that by discussing the "changeset" RFC in context of a core feature, and merging that as such, the logical next step was creating a UI for it in framework as well. So technically speaking, those concerns should've been raised in the context of that RFC already.

Yep, in hindsight it is something that I should’ve thought about much sooner. I guess I’d taken it to be a similar approach to `Versioned`: that it would sit in framework but the UI could vary by implementation.

I understand the point about keeping everything together to make things easy for devs to integrate changesets into their existing workflows, but I think the exact opposite is equally important. A lot of framework-only projects and “brochure sites” have no requirement for changesets/campaigns. Hiding the UI by default in framework goes some way to alleviating that, but it smells wrong ;)

I see @sminnee has created RFC tickets for splitting out things into modules, so I’ll post the rest of my musings there :)

Issue #5279, opened 2016-04-06
Minor: Director code quality and invalid URL's in comments.


patricknelson:
`Director::is_https()` references an invalid URL and should point to https://support.microsoft.com/?kbID=307347. Also some chunks of code are redundant (i.e. initializing `$return`) unnecessary (use of `$return` in the first place) or not very PSR-2 compliant. Quick PR coming soon.

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-framework/pull/5282

PR #5280, opened 2016-04-06
FIX for #5279 Addressing a few PSR-2 items in Director.php and invalid URL's.

patricknelson:
Addresses #5279. A full list of my changes summarized here:

- Added space between `if(` and made it `if (`
- Took `if`/`else` statements missing braces and added braces.
- Combined `else if` to `elseif`
- Took out redundancy in `::is_https()`, simplifying it and making it easier to read as well
- Replaced 2 instances of invalid URL's going to Microsoft's knowledgebase (they must have failed to redirect, should have been easy for Microsoft). From https://support.microsoft.com/?kbID=307347 to https://support.microsoft.com/en-us/kb/307347.

Question: What do you guys think about the process of setting a variable inline within an `if` statement? To me it feels icky. For example, in `Director.php` we have:

```php
if ($env = self::session_environment()) return $env === 'dev';
```

### DIFF
--- a/control/Director.php
+++ b/control/Director.php
@@ -102,15 +102,15 @@ class Director implements TemplateGlobalProvider {
 	public static function direct($url, DataModel $model) {
 		// Validate $_FILES array before merging it with $_POST
 		foreach($_FILES as $k => $v) {
-			if(is_array($v['tmp_name'])) {
+			if (is_array($v['tmp_name'])) {
 				$v = ArrayLib::array_values_recursive($v['tmp_name']);
 				foreach($v as $tmpFile) {
-					if($tmpFile && !is_uploaded_file($tmpFile)) {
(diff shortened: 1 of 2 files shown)


tractorcow:
The code quality improvements are good, but they will make merging up problematic.

With regards to things that are only whitespace changes, we really only want to perform this on the current master, otherwise it makes merge regressions harder to resolve.

We aren't going to implement psr-2 in any 3.x branches, I'm afraid, so sorry if it wasted your time. =(

patricknelson:
No, not a huge waste, just something I spotted and shrieked at a bit while I was reviewing code for other reasons. So, does this mean I can redo these mods against master (if indeed they're still necessary there)? I wouldn't have a problem with that at all.

patricknelson:
@tractorcow Please see  #5282 instead

PR #5282, opened 2016-04-07
FIX for #5279 Addressing only a few PSR-2 items in one file, but primarily targeting Director::is_https() and invalid URL's.

patricknelson:
Addresses issue #5279 and supersedes PR #5280 (now it is on master, not 3.3).

Also @tractorcow in this revision, I tackled a ton of little spots where commas were strung together without spacing, so it should be much more readable now as well. The contents of `::makeRelative()` were particularly egregious.

### DIFF
--- a/control/Director.php
+++ b/control/Director.php
@@ -145,15 +145,15 @@ class Director implements TemplateGlobalProvider {
 	public static function direct($url, DataModel $model) {
 		// Validate $_FILES array before merging it with $_POST
 		foreach($_FILES as $k => $v) {
-			if(is_array($v['tmp_name'])) {
+			if (is_array($v['tmp_name'])) {
 				$v = ArrayLib::array_values_recursive($v['tmp_name']);
 				foreach($v as $tmpFile) {
-					if($tmpFile && !is_uploaded_file($tmpFile)) {
+					if ($tmpFile && !is_uploaded_file($tmpFile)) {
 						user_error("File upload '$k' doesn't appear to be a valid upload", E_USER_ERROR);
 					}
 				}
 			} else {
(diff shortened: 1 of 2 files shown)


patricknelson:
Also, wanted to reiterate previous question:

> Question: What do you guys think about the process of setting a variable inline within an `if` statement? To me it feels icky. For example, in `Director.php` we have:

```php
if ($env = self::session_environment()) return $env === 'dev';
```

kinglozzer:
> Question: What do you guys think about the process of setting a variable inline within an if statement? To me it feels icky. For example, in Director.php we have:
`if ($env = self::session_environment()) return $env === 'dev';`

This is perfectly fine IMO. It’s a more concise way of writing it, no less difficult to understand:

```php
if ($env = self::session_environment()) {
    return $env === 'dev';
}

// vs

$env = self::session_environment();
if ($env) {
    return $env === 'dev';
}
```

patricknelson:
@kinglozzer I don't entirely disagree. I suppose my favorite method of explaining my thought process behind this would be to use an example: https://github.com/bummzack/sortablefile/issues/26

On one hand there isn't much way of opposition to the use of setting variables and then checking them before operating on them (pretty important when working with objects) I tend to encourage people not to overuse it (or avoid it) because:

1. Awareness: It's possible for someone to use it and not understand order of execution to the right of the equal sign (see example) and misuse it by incorporating other statements inline with the single `=` without wrapping in parentheses. 
2. It's possible that the person writing the code didn't intend to use a single `=`, or, it may confuse other people reading the code to assume maybe it was or was not intended (will vary depending on context, author intent and their code comment quality).

It's not a strong argument. It's really just sort of a way for me to pile on a recommendation, like PSR-2 is supposed to be :smile:

PR #5284, opened 2016-04-07
NEW: Expose silverstripe-backend for modules to access.

sminnee:
NEW: Expose silverstripe-backend for modules to access.
FIX: Ensure post has correct content-type header.

Exporting silverstripe-backend lets other modules (such as asset-admin)
use it.

In addition, the Content-type header of a post request of URL-encoded
data was being set to text/plain by default, which isn’t correct and
stopped PHP from interpreting it.


admin/javascript/dist/silverstripe-backend.js
```diff
@@ -82,7 +82,14 @@
     }, {
       key: 'post',
       value: function post(url, data) {
-        return this.fetch(url, { method: 'post', credentials: 'same-origin', body: data }).then(checkStatus);
+        return this.fetch(url, {
+          method: 'post',
+          headers: new Headers({
+            'Content-Type': 'application/x-www-form-urlencoded'
```
chillu: We should allow passing in the desired content type, but given that the whole "backend from scratch" approach is likely superseded by something else, I won't be too picky for now ;)

PR #5287, opened 2016-04-10
Fix React DOM warnings about <form> attrs

chillu:
"class" is a reserved word in JS, hence React uses "className".
"encType" has to be camel cased correctly, see https://facebook.github.io/react/docs/tags-and-attributes.html#html-attributes

Also marked less attrs on <form> as required


tractorcow:
I'l merge once tests pass (if reminded. :P)

chillu:
@tractorcow Build passes

PR #5292, opened 2016-04-11
Panel and layout adjustments, swapped cms typography styles for bootstrap typography

clarkepaul:
Fix for Add page steps
Fix for the south actions panel being different (Site tree vs Edit page)
Increased the width of the collapsed sitetree, for improved click-ability for touch devices
Added close panel btn styles
Replaced styles "reset" with bootstrap "reboot" which will have some minor regressions, but from what I can see from some quick testing there are no major concerns.
Couldn't test on IE as the CMS wouldn't load.
There will be some issues to do with the heading styles now coming from Bootstrap which hopefully we can rectify as we progress through to Beta.


tractorcow:
You have some merge issues in https://github.com/open-sausages/silverstripe-framework/blob/features/cms-panel-and-layout/admin/css/ie7.css. Notice the git SHA in the header?

Fix by removing this file and re-running `npm run css`

tractorcow:
Actually, please remove all ie7.css / ie8.css files, since they aren't supported anymore.

clarkepaul:
@tractorcow Deleted IE files.

SpiritLevel:
That must have been a very satisfying deletion :)

clarkepaul:
Indeed :)

admin/scss/_menu.scss
```diff
@@ -106,7 +108,8 @@
 		position: relative !important;
 		top:auto !important;
 		left:auto !important;
-		box-shadow: inset #C1C7CC -1px 0 0
+		box-shadow: inset #C1C7CC -1px 0 0;
+		min-height: calc(100% - 104px);
```
chillu: What's with the magic numbers? Please make this a var.

admin/scss/_style.scss
```diff
@@ -604,6 +600,35 @@ body.cms {
 }
 
 /**
+ * SilverStripe 4 tabs.
+ * Tab styles built on top of Bootstrap 4 tab functionality
+ */
+
+.nav-tabs {
```
chillu: This should move to its own component

admin/scss/_menu.scss
clarkepaul: I've added a variable in another merge request ( https://github.com/silverstripe/silverstripe-framework/pull/5317 ) which has a variable to use 2 * $navbar-total-height

chillu:
- Moved out the IE deletions to a separate commit
- Fixed some earlier merge errors around the `cms-south-actions` in `_layout.scss`
- Ignoring `.nav-tabs` as component for now, will need to be handled iteratively.

PR #5293, opened 2016-04-11
Remove <IE10 specific workarounds

chillu:
We no longer support IE browsers older than IE10.

/cc @clarkepaul


clarkepaul:
Looks good, feels like the world has just got a whole lot nicer. I don't have access to merge though.

kinglozzer:
:+1:

tractorcow:
Forgot to remove ie7.css and ie8.css files. :D

PR #5296, opened 2016-04-11
Fix Email test issue discovered in #5271.

bummzack:
Updated/added tests for changed- and forgot-password Emails.
Updated fixture and tests to no longer use a real Email address.

### DIFF
--- a/tests/security/MemberTest.php
+++ b/tests/security/MemberTest.php
@@ -185,16 +185,44 @@ class MemberTest extends FunctionalTest {
 	 * Test that changed passwords will send an email
 	 */
 	public function testChangedPasswordEmaling() {
+		$oldSetting = Config::inst()->get('Member', 'notify_password_change');
+		Config::inst()->update('Member', 'notify_password_change', true);
+
 		$this->clearEmails();
 
 		$member = $this->objFromFixture('Member', 'test');
 		$this->assertNotNull($member);
 		$valid = $member->changePassword('32asDF##$$%%');
 		$this->assertTrue($valid->valid());
-		/*
-		$this->assertEmailSent("sam@silverstripe.com", null, "/changed password/",
-		'/sam@silverstripe\.com.*32asDF##\$\$%%/');
-		*/
+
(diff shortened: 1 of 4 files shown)


bummzack:
@dhensby As discussed in #5271, this patches the issue with Email testing for 3. Will create a PR for 3.2 as well if you agree to these changes.

dhensby:
@bummzack if you create a PR against 3.2 then it'll be merged into 3 and master - you don't need individual PRs for each branch :)

NB: I know it's already in master

tests/security/MemberTest.php
```diff
@@ -185,16 +185,44 @@ public function testPasswordChangeLogging() {
 	 * Test that changed passwords will send an email
 	 */
 	public function testChangedPasswordEmaling() {
+		$oldSetting = Config::inst()->get('Member', 'notify_password_change');
+		Config::inst()->update('Member', 'notify_password_change', true);
```
dhensby: You don't need to store the old settings, the most you have to do is nest and unnest the config like so:

```php
function testThing() {
    Config::nest();
    //test code
    Config::unnest();
}
```

But I think we actually made each test "sandboxed" now so the nesting and unnesting is done automatically.

tests/security/MemberTest.yml
```diff
@@ -42,7 +42,7 @@ Member:
   test:
     FirstName: Test
     Surname: User
-    Email: sam@silverstripe.com
+    Email: dummy@silverstripe.tld
```
dhensby: I'd prefer using `@example.com` which will never become a real domain rather than one that just isn't real yet.

bummzack: Oh, ok. I thought .tld was "protected" from becoming a real TLD as well, but I guess that's not true and you never know with ICANN. These changes here are in line with the ones I made to the master branch though… is this something that should be changed in master as well?

bummzack:
Closed in favor of #5298 (which is against 3.2)

tests/security/MemberTest.yml
dhensby: We can just merge them in when we do the merge from 3.2 -> 3 -> master

PR #5297, opened 2016-04-11
Gulpfile eslint

chillu:
Adjusted to AirBnB coding styles, and added to `npm run lint` command. Running `eslint .` takes much longer than `npm run lint`, I haven't found a good way to whitelist folders via `eslintignore` (rather than trawl through all of them and *then* decide which ones to use)


chillu:
@sminnee Wanna reciprocate and merge my eslint fix as well? ;)

PR #5298, opened 2016-04-11
Fix Email test issue discovered in #5271.

bummzack:
Updated/added tests for changed- and forgot-password Emails.
Updated fixture and tests to no longer use a real Email address.

### DIFF
--- a/tests/security/MemberTest.yml
+++ b/tests/security/MemberTest.yml
@@ -42,7 +42,7 @@ Member:
   test:
     FirstName: Test
     Surname: User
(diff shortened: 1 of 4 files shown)


dhensby:
Thanks

tests/security/MemberTest.yml
```diff
@@ -42,7 +42,7 @@ Member:
   test:
     FirstName: Test
     Surname: User
-    Email: sam@silverstripe.com
+    Email: testuser@example.com
```
tractorcow: 👍

PR #5307, opened 2016-04-12
urlTemplateParams support in createEndpointFetcher()

chillu:
Required for REST URLs which require data to be populated on call.
For example, `PUT admin/campaigns/set/:id/publish` to publish a campaign.
At the time when createEndpointFetcher() is called, the `:id` parameter can not be resolved,
since the "data" argument is passed in once the returned function is evaluated (in the component using it).
Since the whole idea of createEndpointFetcher() is that components shouldn't know about the backend internals,
its not the right place to interpolate urls.

I'm sending through any payload data in addition to replacing URLs, to keep the API simple for the component.
Again, it shouldn't worry about the presence URL interpolation. The trade off here is that data will be submitted
both in the payload (in case of a POST, PUT or DELETE request), as well as (optionally) in the URL.


admin/javascript/src/tests/silverstripe-backend-test.js
```diff
@@ -163,5 +163,29 @@ describe('SilverStripeBackend', () => {
         expect(result).toEqual({ status: 'ok', message: 'happy' });
       });
     });
+
+    it('should replace url template parameters', () => {
+      const mock = getBackendMock({
+        text: () => Promise.resolve('{"status":"ok"'),
```
sminnee: Missing `}`

admin/javascript/src/tests/silverstripe-backend-test.js
```diff
@@ -163,5 +163,29 @@ describe('SilverStripeBackend', () => {
...
+          ':type': 'type',
+        },
+      });
+      const promise = endpoint({
+        id: 1,
+        type: 'mytype',
+        other: true
+      });
+      expect(mock.post.mock.calls[0][0]).toEqual('http://example.com/mytype/1/?foo=bar');
+      expect(mock.post.mock.calls[0][1]).toEqual('id=1&type=mytype&other=true');
```
sminnee: My prference would be for `expect(mock.post).toBeCalledWith('http://example.com/mytype/1/?foo=bar', 'id=1&type=mytype&other=true');`

admin/javascript/src/tests/silverstripe-backend-test.js
```diff
@@ -163,5 +163,29 @@ describe('SilverStripeBackend', () => {
...
+      const mock = getBackendMock({
+        text: () => Promise.resolve('{"status":"ok"'),
+        headers: new Headers({
+          'Content-Type': 'application/json',
+        }),
+      });
+      const endpoint = mock.createEndpointFetcher({
+        url: 'http://example.com/:type/:id/?foo=bar',
+        method: 'post',
+        urlTemplateParams: {
```
sminnee: Just thinking about future-proofing this a bit... I expect that we'll have other hooks to express what we want to do with the input. I wonder if this kind of approach might be more extensible?

    payloadSchema: {
      id: { urlReplacement: ':id', remove: true }
      type: { urlReplacement: ':id' }
      foo: { querystring: true }
    }

For example, in the example above I've said that `id` should be removed from the main payload but `type` should still be included, and that `foo` should be added to the querystring.

admin/javascript/src/silverstripe-backend.js
```diff
@@ -152,8 +160,17 @@ class SilverStripeBackend {
       };
 
       const encodedData = encode(refinedSpec.payloadFormat, data);
+      const url = populateTemplate(
+        refinedSpec.url,
+        refinedSpec.urlTemplateParams,
+        data
+      );
+
```
sminnee: Urgh, can we rationalise the arguments for each method instead?

admin/javascript/src/tests/silverstripe-backend-test.js
chillu: I find my `expect()` calls more helpful. `expect().toBeCalledWith()` simply says it hasn't been called with that combination of arguments, which isn't entirely helpful - it might mean the method doesn't get called at all, or that you're one character off in an argument somewhere.

sminnee: fair call.

chillu:
@sminnee Updated. Planning to move docs to a `README.md`, so applying markdown formatting in my alterations already. Figured it's worth a separate PR since it'll move the file into a `framework/admin/javascript/src/lib/backend/` folder

admin/javascript/src/tests/silverstripe-backend-test.js
```diff
@@ -144,24 +140,100 @@ describe('SilverStripeBackend', () => {
...
+        text: () => Promise.resolve('{"status":"ok"}'),
+        headers: new Headers({
+          'Content-Type': 'application/json',
+        }),
+      });
+      const endpoint = mock.createEndpointFetcher({
+        url: 'http://example.com/:one/:two/?foo=bar',
+        method: 'get',
+        payloadSchema: {
+          one: { urlReplacement: ':one', replace: true },
```
sminnee: `replace` should be `remove`, I think?

sminnee:
Apart from the naming of the `replace` property (I had understood it as `remove`, as in "remove from the payload data"), this looks good.

chillu:
@sminnee True that, changed `replace` to `remove` now

PR #5308, opened 2016-04-12
MINOR: Fix linting errors in tests.

sminnee:
Tests were previously excluded from eslint calls. This leads to messier
code inside tests, and potentially errors.

Now they all meet the listing rules, and so we don’t need to ignore
tests in .eslintignore.


admin/javascript/src/components/form-builder/tests/form-builder-test.js
```diff
@@ -1,26 +1,27 @@
...
 
-import { FormBuilderComponent } from '../';
+// import { FormBuilderComponent } from '../';
 
 describe('FormBuilderComponent', () => {
-
   describe('getFormSchema()', () => {
-
-    var formBuilder;
+    /*
```
chillu: I assume you've commented this out to make it clear that no tests are being run, correct?

chillu:
Yay!

admin/javascript/src/components/form-builder/tests/form-builder-test.js
sminnee: Yeah, also because the code complained about variables being assigned and not used... because no tests are bring run ;-)

PR #5309, opened 2016-04-12
Upgrade chosen from 0.9.8 to 1.5.1

hafriedlander:
This is a BC breaking change, but in general, the equivalent of `s/chzn|liszt/chosen/g` will upgrade dependant code.


chillu:
Looks like behat-extension is already using the new `chosen` notation for events, in parallel to the old `liszt`: https://github.com/silverstripe-labs/silverstripe-behat-extension/blob/805a0e69ed859f85daf325a687dd7a2532dcab5b/src/SilverStripe/BehatExtension/Context/SilverStripeContext.php#L496

dhensby:
needs rebasing already :/

dhensby:
see #4284

hafriedlander:
Yeah, almost all PRs that touch javascript will end up needing rebasing before merging now, side-effect of committing the compiled code.

Hadn't seen #4284, but having read through it, my feeling is rise-up isn't important enough to fork dependancy. It's only used in preview, and can be implemented in CSS if needed (as this PR does). I also wouldn't change to select2 unless there's a better reason than "it might be better?" - there's not a big BC break here, and select2 probably has it's own unfortunate issues too.

tractorcow:
I'm testing and I'm finding lots of places where dropdowns are no longer being chosen-ified, although some places where they are.

tractorcow:
The problem is that the shrinkwrap file doesn't include the chosen lib, so npm install won't install it, and rebuilding the libraries breaks distribution files.

As far as I can see, installing chosen with --dev is nigh impossible, because of the sheer amount of modules. My local machine ran out of memory about 4 times trying to install all deps.

If I could find out a way to update the shrinkwrap without install --dev then that would be fine, but no go.

tractorcow:
I suggest we put this story off, and upgrade npm to 3, later.

tractorcow:
I've fixed these issues, rebased, and pushed up.

CMS fix at https://github.com/silverstripe/silverstripe-cms/pull/1454.

Will merge once tests pass.

PR #5311, opened 2016-04-12
FIX Stop "success" message showing in CMS

### LINKED ISSUE #5281: 'success' notifications in chrome with spdy (probably h2 too)
With Google's Chrome and SPDY we have the problem, that a lot of "success" notifications are displayed during navigation (see screenshot).

We found out that this happens when JS files are loaded dynamically during the usage of the CMS. The xhr's `statusText` is "success" instead of "OK", which is the default status text if none is given (by jquery I think, according to this old bug discussion https://bugs.jquery.com/ticket/8534).
In `LeftAndMain.js` (https://github.com/silverstripe/silverstripe-framework/blob/3/admin/javascript/LeftAndMain.js#L179) the check whether the message should be ignored (`statusText == 'OK'`) does not work and the actual statusText "success" is displayed.

The reason for this seems to be the fact, that in HTTP/2 (and obviously SPDY) the reason-phrase (eg "OK") was completely removed (see https://evertpot.com/http-2-finalized/).
So instead of `200 OK` only `200` is returned. It looks like Chrome is handling this differently to Firefox for example. Because with Firefox `statusText` still returns "OK".

During the research in the web I only found this bug of chromium where they fixed a bug related to this reason phrase: https://bugs.chromium.org/p/chromium/issues/detail?id=475223

I think we have to ignore the "success" message too. What do you think?

clipboard01

dhensby:
fixes #5281

### DIFF
--- a/admin/javascript/LeftAndMain.js
+++ b/admin/javascript/LeftAndMain.js
@@ -179,7 +179,7 @@ jQuery.noConflict();
 			var msg = (xhr.getResponseHeader('X-Status')) ? xhr.getResponseHeader('X-Status') : xhr.statusText,
 				reathenticate = xhr.getResponseHeader('X-Reauthenticate'),
 				msgType = (xhr.status < 200 || xhr.status > 399) ? 'bad' : 'good',
-				ignoredMessages = ['OK'];
+				ignoredMessages = ['OK', 'success'];
 
 			// Enable reauthenticate dialog if requested
 			if(reathenticate) {


kinglozzer:
Is 3 the correct branch if this is a bugfix?

dhensby:
@kinglozzer bug fixes don't have to go in 3.x branch, 3 is fine.

I just couldn't be bothered with deciding if this should fix 3.2 or 3.3 or even 3.1!

kinglozzer:
> @kinglozzer bug fixes don't have to go in 3.x branch, 3 is fine.

Yeah I know, someone else just reported this on IRC though... Perhaps given that, and how simple this fix is, it might be prudent to target an earlier release. By rights we _should_ target 3.1 and merge up, right?

dhensby:
I agree with that.

PR #5312, opened 2016-04-12
FIX Stop "success" message showing in CMS

### LINKED ISSUE #5281: 'success' notifications in chrome with spdy (probably h2 too)

dhensby:
fixes #5281

see #5311

### DIFF
--- a/admin/javascript/LeftAndMain.js
+++ b/admin/javascript/LeftAndMain.js
@@ -141,7 +141,7 @@ jQuery.noConflict();
 			var msg = (xhr.getResponseHeader('X-Status')) ? xhr.getResponseHeader('X-Status') : xhr.statusText,
 				reathenticate = xhr.getResponseHeader('X-Reauthenticate'),
 				msgType = (xhr.status < 200 || xhr.status > 399) ? 'bad' : 'good',
-				ignoredMessages = ['OK'];
+				ignoredMessages = ['OK', 'success'];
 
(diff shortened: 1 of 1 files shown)


dhensby:
ping @kinglozzer

tractorcow:
Thanks for fixing that.

PR #5314, opened 2016-04-12
MINOR: Add “npm run add-git-hooks” action.

sminnee:
This commit add an "npm run add-git-hooks" command that will configure
an environment's git hooks and config file to support the development
conventions that we set out.

It's based on recommendations from this post:
http://blog.andrewray.me/dealing-with-compiled-files-in-git/

More git hooks can be added to .githooks

To do:
- [ ] Test what happens when the merge failures happen without the git config
- [ ] Test what happens when you a conflicting non-rebase merge happens
- [ ] Add docs to `en/05_Contributing/01_Code.md#npm`
- [ ] Add checks for SCSS files

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,8 @@
     "test": "NODE_PATH=\"./javascript/src:./admin/javascript/src\" jest",
     "coverage": "NODE_PATH=\"./javascript/src:./admin/javascript/src\" jest --coverage",
     "thirdparty": "gulp thirdparty",
-    "lint": "eslint gulpfile.js & eslint javascript/src & eslint admin/javascript/src"
+    "lint": "eslint gulpfile.js & eslint javascript/src & eslint admin/javascript/src",
+    "add-git-hooks": "./node_modules/.bin/git-hooks --install; git config merge.ours.driver true"
   },
   "repository": {
     "type": "git",
@@ -70,6 +71,7 @@
     "eslint-config-airbnb": "^6.2.0",
     "eslint-plugin-react": "^4.2.3",
     "event-stream": "^3.3.2",
+    "git-hooks": "^1.0.2",
     "glob": "^6.0.4",
     "gulp": "^3.9.0",
     "gulp-babel": "^6.1.1",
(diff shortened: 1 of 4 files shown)


hafriedlander:
With the change to .gitattributes being immediate, but the addition of the merge strategy only happening after "npm run add-git-hooks", have you checked what happens if add-git-hooks isn't run? Does it just error out?

(check-built-files.sh is also pretty lossy - won't pick up scss changes for instance - but probably good enough as a start)

hafriedlander:
Something else I wanted to test: what happens if you use merge instead of rewrite and you get a clash, does "post-rewrite" still get called?

package.json
```diff
@@ -18,7 +18,8 @@
     "test": "NODE_PATH=\"./javascript/src:./admin/javascript/src\" jest",
     "coverage": "NODE_PATH=\"./javascript/src:./admin/javascript/src\" jest --coverage",
     "thirdparty": "gulp thirdparty",
-    "lint": "eslint gulpfile.js & eslint javascript/src & eslint admin/javascript/src"
+    "lint": "eslint gulpfile.js & eslint javascript/src & eslint admin/javascript/src",
+    "add-git-hooks": "./node_modules/.bin/git-hooks --install; git config merge.ours.driver true"
```
flashbackzoo: You should be able to just do `git-hooks --install; git config merge.ours.driver true`. See https://docs.npmjs.com/misc/scripts#path

flashbackzoo:
We've been documenting npm scripts https://github.com/silverstripe/silverstripe-framework/blob/master/docs/en/05_Contributing/01_Code.md#npm can you please add the new script?

sminnee:
To do:
 - [ ] Test what happens when the merge failures happen without the git config
 - [ ] Test what happens when you a conflicting non-rebase merge happens
 - [ ] Add docs to `en/05_Contributing/01_Code.md#npm`
 - [ ] Add checks for SCSS files

sminnee:
Abandoning this as stale.

PR #5304, opened 2016-04-11
API Implement campaign list view

tractorcow:
This change introduces a working navigable list of campaign items, which can be viewed by clicking on any item in the changeset admin interface.

This feature is not fully complete, but will be built upon in later pull requests to fully flesh out the campaign admin interface. Known non-working elements include:

- Live breadcrumbs (currently hard-coded)
- Linked campaign items are not properly rendered
- Delete / Edit view for campaigns are not yet implemented
- Preview view for campaign items is currently hard coded to the site home page

I have separately include a new i18n change to support pluralisation, based on the Zend_Translate API used by other parts of i18n. This includes support for `DataObject::i18n_pluralise`, and is used by ChangeSet model to help describe items linked by that changeset.

image

PR #5316, opened 2016-04-13
[WIP] Campaign publish button

chillu:
Implements the frontend UI and backend API logic for publishing campaigns. See https://projects.invisionapp.com/boards/9GX68PTXUQFH/#/1547737/71700347

- Builds on https://github.com/silverstripe/silverstripe-framework/pull/5304, so needs to be rebased once that other pull request is merged.
- Also reliant on https://github.com/silverstripe/silverstripe-framework/pull/5307 being merged.
- No PHPUnit tests for `CampaignAdmin` functionality. It's pretty basic stuff, and all those API endpoints should be considered temporary - we haven't gotten around to discussing a "proper API layer" in SS4 yet.
- Contains a refactor of the `records` state shape, from array based to object+key based. See commit for rationale. Technically it could be a separate PR, but it also changes the way campaign admin interacts with this store, which in turn is bound to https://github.com/silverstripe/silverstripe-framework/pull/5304

Here's an easy way for `Page_Controller` to generate a campaign with a changed page (since there's no way to do this through the UI yet). Careful, destructive action.

```
public function setup() {
	DB::query('TRUNCATE ChangeSet');
	DB::query('TRUNCATE ChangeSetItem');

	$page = Page::get()->byID(1);
	$page->Content = rand();
	$page->Version = $page->Version + 1;
	$page->writeToStage('Stage');

	$set = ChangeSet::get()->filter('Name', 'Test')->first();
	if(!$set) {
		$set = new ChangeSet();
	}
	$set->Name = 'Test';
	$set->write();
	foreach(Page::get()->filter('ParentID', 0) as $page) {
		$set->addObject($page);
	}
	$set->write();

}
```


admin/javascript/src/silverstripe-backend.js
```diff
@@ -153,7 +153,11 @@ class SilverStripeBackend {
 
       const encodedData = encode(refinedSpec.payloadFormat, data);
 
-      return this[refinedSpec.method](endpointSpec.url, encodedData, headers)
+      const args = refinedSpec.method === 'get'
```
flashbackzoo: Might be worth converting the method name to lowercase like we do in `admin/javascript/src/state/records/actions.js`

For example `refinedSpec.method.toLowerCase()`

admin/javascript/src/state/records/actions.js
```diff
@@ -32,7 +32,11 @@ export function fetchRecords(recordType, method, url) {
       type: ACTION_TYPES.FETCH_RECORDS_REQUEST,
       payload,
     });
-    return backend[method.toLowerCase()](populate(url, payload), {}, headers)
+    const args = method.toLowerCase() === 'get'
```
flashbackzoo: Wouldn't fetching records always use the `GET` verb i.e. do we even need this conditional assignment?

admin/code/CampaignAdmin.php
```diff
@@ -416,6 +422,43 @@ public function deleteCampaign(SS_HTTPRequest $request) {
...
+		try {
+			$record->publish();
+		} catch(LogicException $e) {
+			return (new SS_HTTPResponse(json_encode(['status' => 'error', 'message' => $e->getMessage()]), 401))
+				->addHeader('Content-Type', 'application/json');
+		}
+
+		return (new SS_HTTPResponse(
+			Convert::raw2json($this->getChangeSetResource($record)),
+			401
```
flashbackzoo: This should be `200` right?

chillu: Fixed in later commit

admin/javascript/src/sections/campaign-admin/list.js
```diff
@@ -79,12 +81,44 @@ class CampaignListContainer extends SilverStripeComponent {
...
+            {this.renderButtonToolbar()}
+          </div>
         </div>
         { previewUrl && <CampaignPreview previewUrl={previewUrl} /> }
       </div>
     );
   }
 
+  renderButtonToolbar() {
+    const items = this.getItems(this.props.campaignId);
```
flashbackzoo: It looks like the `this.getItems` method doesn't use the param it's being passed here.

admin/javascript/src/sections/campaign-admin/list.js
```diff
@@ -79,12 +81,44 @@ class CampaignListContainer extends SilverStripeComponent {
...
+         i18n._t('Campaigns.ITEM_SUMMARY_SINGULAR')
+         : i18n._t('Campaigns.ITEM_SUMMARY_PLURAL'),
+       items.length
+     );
+
+      return (
+        <div className="btn-toolbar">
+          <FormAction
+            label={i18n._t('Campaigns.PUBLISHCAMPAIGN')}
+            style={'success'}
```
flashbackzoo: This seems a bit weird. The `style` prop should take an object. See https://facebook.github.io/react/tips/inline-styles.html

admin/javascript/src/state/campaign/reducer.js
```diff
@@ -15,6 +15,9 @@ function campaignReducer(state = initialState, action) {
         view: action.payload.view,
       }));
 
+    case ACTION_TYPES.PUBLISH_CAMPAIGN_SUCCESS:
```
flashbackzoo: I think this should unset `campaignId`? Also needs cases for `PUBLISH_CAMPAIGN_REQUEST` and `PUBLISH_CAMPAIGN_FAILURE` which should set / unset `campaignId`.

admin/javascript/src/components/form-action/index.js
```diff
@@ -100,7 +101,7 @@ class FormActionComponent extends SilverStripeComponent {
 
 FormActionComponent.propTypes = {
   id: React.PropTypes.string,
-  handleClick: React.PropTypes.func.isRequired,
+  handleClick: React.PropTypes.func,
```
flashbackzoo: If `handleClick` is no longer required we should put a guard in `this.handleClick` so that `this.props.handleClick` isn't invoked when it's `undefined`

admin/javascript/src/state/records/reducer.js
```diff
@@ -29,7 +28,8 @@ function recordsReducer(state = initialState, action) {
     case ACTION_TYPES.FETCH_RECORDS_SUCCESS:
       recordType = action.payload.recordType;
       // TODO Automatic pluralisation from recordType
-      records = action.payload.data._embedded[`${recordType}s`];
+      records = action.payload.data._embedded[`${recordType}s`] || [];
```
flashbackzoo: I think this should be an object literal...

admin/javascript/src/sections/campaign-admin/list.js
```diff
@@ -212,6 +213,7 @@ class CampaignListContainer extends SilverStripeComponent {
 
 CampaignListContainer.propTypes = {
   publishApi: React.PropTypes.func.isRequired,
+  isPublishing: React.PropTypes.bool,
```
flashbackzoo: This should be:

```
campaign: React.PropTypes.shape({
    isPublishing: React.PropTypes.bool,
}),
```

flashbackzoo: We should also add a key for `record`

admin/javascript/src/state/campaign/reducer.js
```diff
@@ -15,6 +15,17 @@ function campaignReducer(state = initialState, action) {
         view: action.payload.view,
       }));
 
+    case ACTION_TYPES.PUBLISH_CAMPAIGN_REQUEST:
```
flashbackzoo: We need to add `isPublishing: false` to the initial state

admin/javascript/src/state/campaign/reducer.js
chillu: Why? You're still viewing that campaign after publishing, right? On a related note: What's best practice for cleaning up Redux UI state after a view is "closed"? I guess as long as any new instance of this view derives its new state correctly (e.g. from routing), it doesn't matter too much.

admin/javascript/src/state/records/actions.js
chillu: In a REST sense, yes it would always be GET. But from a generic API handling perspective, I wouldn't make that assumption.

admin/javascript/src/sections/campaign-admin/list.js
```diff
@@ -99,56 +99,53 @@ class CampaignListContainer extends SilverStripeComponent {
...
+      actionProps = Object.assign(actionProps, {
+        label: i18n._t('Campaigns.PUBLISHCAMPAIGN'),
+        bootstrapButtonStyle: 'success',
+        loading: this.props.campaign.isPublishing,
+        handleClick: this.handlePublish,
+        icon: 'rocket',
+      });
+    } else if (this.props.record.State === 'published') {
+      // TODO Implement "revert" feature
+      actionProps = Object.asign(actionProps, {
```
chillu: @flashbackzoo ES8 syntax? ;) Fixed it locally

admin/javascript/src/sections/campaign-admin/list.js
```diff
@@ -219,8 +216,13 @@ class CampaignListContainer extends SilverStripeComponent {
 }
 
 CampaignListContainer.propTypes = {
+  campaign: React.PropTypes.shape({
+    isPublishing: React.PropTypes.bool.isRequired,
+  }),
+  campaignActions: React.PropTypes.object.isRequired,
   publishApi: React.PropTypes.func.isRequired,
-  isPublishing: React.PropTypes.bool,
+  record: React.PropTypes.array.isRequired,
```
chillu: This should be `React.PropTypes.object` - fixed locally

admin/code/CampaignAdmin.php
```diff
@@ -416,6 +424,44 @@ public function deleteCampaign(SS_HTTPRequest $request) {
 	}
 
 	/**
+	 * REST endpoint to publish a {@link ChangeSet} and all of its items.
+	 *
+	 * @param SS_HTTPRequest $request
+	 *
+	 * @return SS_HTTPResponse
+	 */
+	public function publishCampaign(SS_HTTPRequest $request) {
```
tractorcow: We'll need CSRF on all destructive requests. :)

chillu:
@flashbackzoo Have squashed a bit more, fixed some react warning about array vs object prop types.

package.json
```diff
@@ -70,6 +70,7 @@
     "eslint": "^2.5.3",
     "eslint-config-airbnb": "^6.2.0",
     "eslint-plugin-react": "^4.2.3",
+    "espree": "https://github.com/eslint/espree.git",
```
tractorcow: We can take this out now.

admin/javascript/src/silverstripe-backend.js
```diff
@@ -58,7 +58,7 @@ class SilverStripeBackend {
    * Both `payloadFormat` and `responseFormat` can use the following shortcuts for their
    * corresponding mime types:
    *
-   *   - urlencoded: application/x-www-form-url-encoded
+   *   - urlencoded: application/x-www-form-urlencoded
```
chillu: @sminnee Just checking that you didn't write `application/x-www-form-url-encoded` on purpose? The last dash is wrong, and it causes data to be interpreted differently by CGI (doesn't end up in `$_POST`)

sminnee: No, that me causing Ingo a lot of pain. Sorry!
On Mon, 18 Apr 2016 at 8:59 AM, Ingo Schommer <notifications@github.com>
wrote:

> In admin/javascript/src/silverstripe-backend.js
> <https://github.com/silverstripe/silverstripe-framework/pull/5316#discussion_r59992621>
> :
>
> > @@ -58,7 +58,7 @@ class SilverStripeBackend {
> >     * Both `payloadFormat` and `responseFormat` can use the following shortcuts for their
> >     * corresponding mime types:
> >     *
> > -   *   - urlencoded: application/x-www-form-url-encoded
> > +   *   - urlencoded: application/x-www-form-urlencoded
>
> @sminnee <https://github.com/sminnee> Just checking that you didn't write
> application/x-www-form-url-encoded on purpose? The last dash is wrong,
> and it causes data to be interpreted differently by CGI (doesn't end up in
> $_POST)
>
> —
> You are receiving this because you were mentioned.
>
>
> Reply to this email directly or view it on GitHub
> <https://github.com/silverstripe/silverstripe-framework/pull/5316/files/ba059b30c96588431b3fd7e407047dab24568e41..57f94e6794f676df98e71caa9561d966ae22e558#r59992621>
>
-- 
Sam Minnée
CEO
SilverStripe Limited

package.json
chillu: Sweet, removed now

admin/code/CampaignAdmin.php
chillu: Done now, for publish. Delete has the same issue (existing feature, not part of this PR) - created an issue for it at https://github.com/silverstripe/silverstripe-framework/issues/5328

admin/javascript/src/silverstripe-backend.js
```diff
@@ -270,18 +273,20 @@ class SilverStripeBackend {
...
     );
 
-    return (data) => {
+    return (data = {}) => {
       const headers = {
         Accept: refinedSpec.responseFormat,
         'Content-Type': refinedSpec.payloadFormat,
       };
 
+      const mergedData = Object.assign({}, refinedSpec.defaultData, data);
```
flashbackzoo: Keep in mind `Object.assign` only does a shallow merge. For example:

```js
var defaults = {
  title: 'First name',
  value: null,
  attributes: {
    class: 'text-field',
    id: 1,
    name: 'first-name',
  },
};

var data = {
  title: 'Full name',
  attributes: {
    name: 'full-name',
  },
};

Object.assign({}, defaults, data); // { title: 'Full name', value: null, attributes: { name: 'full-name' } }
```

admin/javascript/src/sections/campaign-admin/controller.js
```diff
@@ -161,7 +162,8 @@ CampaignAdminContainer.propTypes = {
 
 function mapStateToProps(state, ownProps) {
   return {
-    config: state.config.sections[ownProps.sectionConfigKey],
+    config: state.config,
+    sectionConfig: state.config.sections[ownProps.sectionConfigKey],
```
flashbackzoo: Should have a `propTypes` key.

`sectionConfig: React.PropTypes.object.isRequired,`

admin/javascript/src/silverstripe-backend.js
chillu: Yeah it will, good catch! Using https://www.npmjs.com/package/merge now.

admin/javascript/src/sections/campaign-admin/controller.js
chillu: Done

admin/code/CampaignAdmin.php
```diff
@@ -62,6 +64,10 @@ public function getClientConfig() {
 			],
 			'campaignViewRoute' => $urlSegment . '/:type?/:id?/:view?',
 			'itemListViewEndpoint' => $this->Link('set/:id/show'),
+			'publishEndpoint' => [
+				'url' => $this->Link('set/:id/publish'),
```
tractorcow: Be careful; Link takes an action, and in this case only `set` is a valid action. Additional arguments should probably be appended afterwards.

admin/code/CampaignAdmin.php
```diff
@@ -416,6 +424,50 @@ public function deleteCampaign(SS_HTTPRequest $request) {
...
 	/**
+	 * REST endpoint to publish a {@link ChangeSet} and all of its items.
+	 *
+	 * @param SS_HTTPRequest $request
+	 *
+	 * @return SS_HTTPResponse
+	 */
+	public function publishCampaign(SS_HTTPRequest $request) {
+		// Protect against CSRF on destructive action
+		if(!SecurityToken::inst()->checkRequest($request)) {
```
tractorcow: 👍

admin/code/LeftAndMain.php
```diff
@@ -194,6 +194,10 @@ public function getCombinedClientConfig() {
 			$combinedClientConfig['sections'][$className] =  Injector::inst()->get($className)->getClientConfig();
 		}
 
+		// Get "global" CSRF token for use in JavaScript
+		$token = new SecurityToken();
```
tractorcow: You should respect `SecurityToken::inst()`, not force a specific class.

tractorcow:
Rebased :D

PR #5317, opened 2016-04-14
CSS updates to Campaign Admin

clarkepaul:
Squashed commits contains:
Refactored breadcrumbs to use bootstrap and variables
Grid table - fix border lines on table
Layout variables updated for spacing, added toolbar heights and spacing
Added action toolbar for content, swapped values to variables
Move accordion styles to its own component
Campaign items, styles for linked items and state badges


clarkepaul:
Might be best to merge first https://github.com/silverstripe/silverstripe-framework/pull/5317/files#diff-c6262cc230f275d68e5093c7a37459b0

admin/javascript/src/components/accordion/styles.scss
```diff
@@ -0,0 +1,53 @@
...
+				font-size: $font-size-lg;
+				line-height: $line-height-base;
+				color: lighten($body-color,30);
+				text-align: center;
+				content: "7";
+				font-family: silverstripe;
+			}
+
+			&.collapsed::before {
+				content: "6";
```
chillu: Not very readable way of referencing icons in the icon font - but looks like there's no straightforward way to fix this (found http://varya.me/en/posts/dealing-atomically-with-icon-fonts/). One significant disadvantage here: From what I can tell, you need access to a specific http://fontastic.me/ account (and icon collection) to add icons to that font. How would you see a community member contributing there?

admin/javascript/src/components/accordion/styles.scss
```diff
@@ -0,0 +1,53 @@
+// Accordion styles
+.accordion-group {
+	margin-top: $spacer-y;
+
+	&__title {
+		margin-bottom: 0;
+
+		a {
```
chillu: Still think we shouldn't use tag names as selectors. BEM recommends this as well: http://getbem.com/faq/#custom-tags-for-blocks

admin/javascript/src/components/accordion/styles.scss
```diff
@@ -0,0 +1,53 @@
...
+			text-transform: uppercase; 
+			padding: $spacer-x*.75 $spacer-y;
+			/* end table header */
+			
+			display: block;
+			margin-left: -$spacer-y;
+			margin-right: -$spacer-y;
+			text-decoration: none;
+			position: relative;
+			border-bottom: 1px solid #ddd;
```
chillu: Shouldn't this colour be a variable?

admin/javascript/src/components/north-header/styles.scss
```diff
@@ -1,4 +1,25 @@
-.north-header-component {
-    @extend .cms-content-header;
+.north-header {
+    border-bottom: 1px solid $border-color-dark;
+    background-color: $background-north;
     width: 100%;
+    height: $navbar-total-height;
+    z-index: 60;
+    
+    .north-header__navigation {		// Typically to hold breadcrumbs and back button
```
chillu: BEM says not to nest selectors (they should only be prefixed)

admin/javascript/src/sections/campaign-admin/controller.js
```diff
@@ -52,13 +53,21 @@ class CampaignAdminContainer extends SilverStripeComponent {
...
         <div className="cms-campaigns collapse in" aria-expanded="true">
-          <NorthHeader />
-          <FormAction
-            label={i18n._t('Campaigns.ADDCAMPAIGN')}
-            icon={'plus-circled'}
-            handleClick={this.addCampaign}
-          />
-          <FormBuilder schemaUrl={schemaUrl} createFn={this.createFn} />
+          <NorthHeader>
+            <h2 className="text-truncate north-header__heading">Campaigns</h2>
```
chillu: Needs `i18n()` for the label

admin/javascript/src/sections/campaign-admin/controller.js
```diff
@@ -119,6 +129,19 @@ class CampaignAdminContainer extends SilverStripeComponent {
     return <Component key={props.name} {...props} />;
   }
 
+  getBreadcrumbs() {
```
chillu: Needs TODO tag, since this is hardcoded

admin/javascript/src/sections/campaign-admin/styles.scss
```diff
@@ -1,210 +1,121 @@
...
+			font-size: 10px;
+			font-weight: 400;
+			letter-spacing: .4px;
+		}
 
-	// Accordion styles
-	.accordion-group {
-		margin-top: 16px;
+		// Show linked items
+		&--haslinks,
```
chillu: These should be dash separated: `--has-links` and `--is-linked`

admin/scss/bootstrap/_variables.scss
```diff
@@ -38,6 +38,16 @@ $brand-success:             #3fa142;
 // $brand-warning:             #f0ad4e;
 $brand-danger:              #D40404;
 
+// Backgrounds
+$background-main: 				#f6f7f8;
```
chillu: Either we correctly align all values here in terms of whitespace, or we admit defeat and only use a single whitespace after the double colon. We shouldn't mix both approaches, it looks even messier ;)

admin/javascript/src/sections/campaign-admin/controller.js
chillu: Come to think of it, this should probably be passed down from the server-side config - since that's the way we configure ModelAdmins for example. Let's hold off on this, hardcoding is OK for now.

chillu:
Alright, implemented my feedback myself - we have too many inter-dependant CSS pull requests in the queue here :) This PR breaks the bottom toolbar in campaign admin, so there's no way to use the "publish" button. Presumably this will be added back via a follow up pull request: https://github.com/silverstripe/silverstripe-framework/pull/5292

@clarkepaul I've simplified the breadcrumbs component a bit, and added some more specific CSS selectors which don't follow BEM but unfortunately are necessary until we solve a) Bootstrap namespacing and b) global CSS styles (e.g. `.cms h2`).

PR #5318, opened 2016-04-14
BUG fix routing for react sections

tractorcow:
Merge together with https://github.com/silverstripe/silverstripe-asset-admin/pull/116

There is an interesting and still to be improved issue when the top level CMS section is changing; Registered routes which handle sub-sections within the CMS will need to suppress ajax loading in cases where they are active, otherwise LeftAndMain.js will cause a reload (via ajax).

However if they are not active, suppressing these routes will cause no reload at all, and only the url will appear to change.

In order to detect boundary navigation, I've temporarily put in a hack to detect requests from loadPanel, and use the referer to determine if the current route handler should delegate to the top level handler or not.

This working horrible hack should be replaced with a global top level "current section" state, which can only be set by LeftAndMain, but can be read by any route handler. Ideally this should be implemented when refactoring the top level route handling out of leftandmain and into a separate class.


npm-shrinkwrap.json
```diff
@@ -5,16 +5,16 @@
   "node-version": "v4.4.3",
   "dependencies": {
     "autoprefixer": {
-      "version": "6.3.6",
-      "resolved": "https://registry.npmjs.org/autoprefixer/-/autoprefixer-6.3.6.tgz",
+      "version": "6.3.3",
```
sminnee: These changes are a bit of a code smell.

Maybe this will tidy it up?

    git checkout -f HEAD npm-shrinkwrap.json
    npm install
    npm install url --save-dev
    git run lock

admin/javascript/src/config.js
```diff
@@ -26,8 +26,9 @@ class Config {
     const topLevelRoutes = [];
 
     Object.keys(window.ss.config.sections).forEach((key) => {
+      // Top level routes should have only one url-segment beneath admin/ and have a trailing slash
       const route = window.ss.config.sections[key].route;
-      const isTopLevelRoute = route.indexOf('/') === -1;
+      const isTopLevelRoute = route.match(/^admin\/([^\/]+)\/$/);
```
sminnee: Do these routes always end in `/`? Feels a bit brittle, maybe it needs a `?` before `$`?

javascript/src/router.js
```diff
@@ -3,25 +3,45 @@
...
-    // Normalise `path` so that pattern matching is more robust.
-    // For example if your route is '/pages' it should match when `path` is
-    // 'http://foo.com/admin/pages', '/pages', and 'pages'.
-    const el = document.createElement('a');
-    let pathWithSearch;
-    el.href = path;
-    pathWithSearch = el.pathname;
-    if (el.search) {
-      pathWithSearch += el.search;
+function normalise(path) {
```
sminnee: `normalise` feels a bit too generic a name, maybe something `pathRelativeToBase` or `baseRelativePath` would be better?

javascript/src/router.js
```diff
@@ -36,6 +56,21 @@ function routeAppliesToCurrentLocation(route) {
...
+function getAbsoluteBase() {
+  const baseTags = window.document.getElementsByTagName('base');
+  if (baseTags && baseTags[0]) {
+    return baseTags[0].href;
+  }
+  return null;
+}
+
+page.getAbsoluteBase = getAbsoluteBase.bind(page);
+page.normalise = normalise.bind(page);
```
sminnee: Does this need to be added as a public API? If we add it, we're committing to support it.

javascript/src/router.js
sminnee: If it's only exposed for testing, maybe we want to mention as such in a docblock?

admin/javascript/src/sections/campaign-admin/controller.js
```diff
@@ -19,8 +19,23 @@ class CampaignAdminContainer extends SilverStripeComponent {
...
 
   componentDidMount() {
-    window.ss.router(`/${this.props.config.campaignViewRoute}`, (ctx) => {
+    window.ss.router(`/${this.props.config.campaignViewRoute}`, (ctx, next) => {
       this.props.actions.showCampaignView(ctx.params.id, ctx.params.view);
+
+      // If we are coming from a non-campaigns section, we need to call next()
+      // to trigger an ajax reload
+
+      // @todo - Use state to detect section changed behaviour rather than referer
```
tractorcow: I wonder if we can use https://github.com/visionmedia/page.js/#pageexitpath-callback-callback- instead, when we enter a campaign section we could set a variable to true, and set it to false again on exit.

admin/javascript/src/config.js
tractorcow: Yes, in LeftAndMain::Link() it puts a trailing / on all top level sections. We can use a ? though.

npm-shrinkwrap.json
tractorcow: I'll try a tidy up tomorrow... not sure why the versions regressed.

javascript/src/router.js
tractorcow: It's used in campaign admin too, but I agree with a rename.

javascript/src/router.js
tractorcow: I think we should support it, since it's going to be a standard route mapping between "traditional SS urls" and "paths expected by page.js". It will probably live as long as page.js does...

javascript/src/router.js
tractorcow: I'm going to call it `resolveURLToBase`

sminnee:
:+1: from me

admin/javascript/src/LeftAndMain.js
```diff
@@ -227,18 +227,15 @@ $.entwine('ss', function($) {
 			var self = this,
 				basePath = getUrlPath($('base')[0].href);
 
-			// Avoid adding a double slash if the base path is '/'
+			// Remove trailing / from base
 			if (basePath[basePath.length - 1] === '/') {
-				basePath += 'admin';
-			} else {
-				basePath = '/admin';
+				basePath = basePath.substring(0, basePath.length - 1);
```
chillu: That's encoding assumptions about how SS PHP formats the `<base>` tag. That's overwriteable behaviour, so might not always contain a trailing slash. Can we please use a more targeted `.replace(/\/$, ''/)` there?

admin/javascript/src/config.js
```diff
@@ -26,8 +26,9 @@ class Config {
     const topLevelRoutes = [];
 
     Object.keys(window.ss.config.sections).forEach((key) => {
+      // Top level routes should have only one url-segment beneath admin/ and have a trailing slash
       const route = window.ss.config.sections[key].route;
-      const isTopLevelRoute = route.indexOf('/') === -1;
+      const isTopLevelRoute = route.match(/^admin\/[^\/]+\/?$/);
```
chillu: Why does this distinction matter here in the first place? If a route is placed in `window.ss.config.sections[key].route`, isn't it always assumed to be "top level"? Other routes like GridField detail views won't be registered through that config key path.

admin/code/LeftAndMain.php
```diff
@@ -206,7 +206,7 @@ public function getCombinedClientConfig() {
 	 */
 	public function getClientConfig() {
 		return [
-			'route' => Config::inst()->get($this->class, 'url_segment')
+			'route' => $this->Link()
```
sminnee: Just a note that this affects every admin section. Have we confirmed that the current sections (and a typical model admin) still work? That is, it's not a common practise for LeftAndMain subclasses to override `Link()` in stupid ways?

admin/javascript/src/LeftAndMain.js
tractorcow: Ok, I'll update this code to "autofix" missing trailing slash.

admin/javascript/src/config.js
tractorcow: There are other CMS sections which aren't top levels. E.g. CMSPageAddController which has a url segment `admin/pages/add`.

admin/code/LeftAndMain.php
tractorcow: Yes; getClientConfig() affects the code in LeftAndMain.js which was updated in the same pull request.

tractorcow:
I've now removed routing from silverstripe-component. See https://github.com/silverstripe/silverstripe-asset-admin/pull/116 for the associate fix to asset-admin.

admin/code/CampaignAdmin.php
```diff
@@ -53,19 +53,17 @@ class CampaignAdmin extends LeftAndMain implements PermissionProvider {
...
 					'schemaUrl' => $this->Link('schema/EditForm')
 				]
 			],
-			'campaignViewRoute' => $urlSegment . '/:type?/:id?/:view?',
-			'itemListViewEndpoint' => $this->Link('set/:id/show'),
+			'campaignViewRoute' => $this->Link() . ':type?/:id?/:view?',
+			'itemListViewEndpoint' => $this->Link() . 'set/:id/show',
 			'publishEndpoint' => [
-				'url' => $this->Link('set/:id/publish'),
+				'url' => $this->Link() . 'set/:id/publish',
```
tractorcow: The problem with this is that anything with a / in it isn't actually an action; ? can also get mangled my Controller::join_links, since it looks like a querystring delimiter.

admin/javascript/dist/LeftAndMain.js
```diff
@@ -181,16 +181,13 @@
...
-				} else {
-					basePath = '/admin';
+					basePath = basePath.substring(0, basePath.length - 1);
 				}
-
 				_router2.default.base(basePath);
 
 				_config2.default.getTopLevelRoutes().forEach(function (route) {
-					(0, _router2.default)('/' + route + '/*', function (ctx, next) {
-						if (document.readyState !== 'complete' || typeof ctx.state.__forceReferer === 'undefined') {
```
tractorcow: having the __forceReferer check here broke going "back" to the initial state, since it wasn't loaded via loadPanel().

Instead, sub-section loaders can conditionally avoid calling next() in their route handlers to prevent bubbling back up to this root controller.

admin/javascript/src/LeftAndMain.js
```diff
@@ -227,21 +227,18 @@ $.entwine('ss', function($) {
...
 			}
-
 			router.base(basePath);
 
 			// Register all top level routes.
 			Config.getTopLevelRoutes().forEach((route) => {
-				router(`/${route}/*`, (ctx, next) => {
-					// If the page isn't ready or the request hasn't come from 'loadPanel'
-					// then don't PJAX load the panel. Note: __forceReferer is set by 'loadPanel' only.
-					if (document.readyState !== 'complete' || typeof ctx.state.__forceReferer === 'undefined') {
```
tractorcow: having the __forceReferer check here broke going "back" to the initial state, since it wasn't loaded via loadPanel().

Instead, sub-section loaders can conditionally avoid calling next() in their route handlers to prevent bubbling back up to this root controller.

tractorcow:
Will rebase once https://github.com/silverstripe/silverstripe-framework/pull/5346 is merged.

tractorcow:
Oops, I didn't update the tests. It's fixed though by now. ;D

PR #5321, opened 2016-04-15
Converted grid-field to table markup and SCSS to BEM

clarkepaul:


dev/install/css/install.css
```diff
@@ -1,3 +1,706 @@
...topright:3px;-moz-border-radius-bottomright:3px;-moz-border-radius-bottomleft:3px;border-radius:3px 3px 3px 3px;-webkit-box-shadow:inset 0 0 3px 0 #ddd;box-shadow:inset 0 0 3px 0 #ddd;-webkit-box-shadow:0 1px 0 0 #fff;box-shadow:0 1px 0 0 #fff}body ul#Themes{float:left;width:445px}body .databaseError{width:422px}body #Footer{margin-top:67px;margin-bottom:18px}body #Footer p{font-size:12px;color:#999}body #Header h1{margin-top:45px;margin-bottom:27px}body #Header h3{margin-top:-10px}#adminAcc,#devHelp,#localeHelp{padding-top:20px}#devHelp,#devSection{height:18em}#use_environment_field{margin-bottom:25px;background-color:#FAFAFA;border:1px solid #DDDDDD;border-radius:3px 3px 3px 3px;-webkit-box-shadow:0 1px 0 0 white;box-shadow:0 1px 0 0 white;margin-bottom:9px;padding:10px 10px 8px}#use_environment_field input{float:left;margin-right:5px}.databaseError ul{margin-bottom:0}.databaseError ul,.databaseError li{margin-left:0;padding-left:0}#Layout p.goodInstall{color:#359318}.goodInstall{padding:0.5em;background-color:#e2fee1;border:1px #43cb3e solid;color:#359318;overflow:hidden;line-height:18px;padding:10px}.goodInstall a{float:right;font-size:18px;padding:0 2px 2px 0;font-weight:bold}
```
sminnee: I believe that these should have been generated minified?

admin/scss/bootstrap/_variables.scss
```diff
@@ -57,7 +57,7 @@ $brand-danger:              #D40404;
 // Control the default styling of most Bootstrap elements by modifying these
 // variables. Mostly focused on spacing.
 
-$spacer:                     1rem;
+$spacer:                     1.2308rem;
```
sminnee: This is an awfully specific number. Is it magical? It might benefit from a comment as to how this number was chosen.

admin/scss/bootstrap/_variables.scss
```diff
@@ -135,7 +137,7 @@ $body-color:                 $gray-dark;
 // Set the number of columns and specify the width of the gutters.
 
 // $grid-columns:               12;
-// $grid-gutter-width:          1.875rem; // 30px
+$grid-gutter-width:          2.4615rem; // now 32px, was 1.875rem; ~ 30px
```
sminnee: Can we use formulas. Perhaps it would be better as expressing this as `32/13` assuming that there are 13px per rem in our default set-up.

admin/scss/bootstrap/_variables.scss
sminnee: Might be better to write `16/13`? A comment at the top of the section saying "at default zoom there are 13 pixels per rem, so measurements are expressed as desired pixels / rem.

Better yet, make 1/3 a variable called `$px_to_rem` and then say `spacer: (16 * $px_to_rem)rem` or whatever the syntax is.

admin/scss/bootstrap/_variables.scss
clarkepaul: If we are to do that then we might as well just say its 32px. I guess I was trying to keep the rem thing going as much as possible so it can be changed globally based on the font sizing. I could change both $spacer and $grid-gutter-width to pixels values for now and see if that runs into any complications down the road like if the font-size changes, but it might be safer to leave as is with better comments.

dev/install/css/install.css
clarkepaul: I'm not sure what's the deal with this. @flashbackzoo do you know if I'm supposed to do anything different with it?

clarkepaul: ok done, didn't realise --development doesn't minify the css. I need to check my other pull requests.

admin/scss/bootstrap/_variables.scss
clarkepaul: I've kept $spacer in rems with explanation and gutter width is 2x$spacer, I think that should be good.

clarkepaul:
https://github.com/silverstripe/silverstripe-framework/pull/5292/files should probably be merged before this one.

dev/install/css/install.css
chillu: This has been fixed since then, we're running `outputStyle: 'compressed'` regardless of `--development` now

admin/scss/bootstrap/_variables.scss
chillu: This convention is taken from bootstrap: https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss#L164. Since devs wouldn't be expected to calculate or copy these values around themselves, its a low priority to figure out. Paul said something about independence between font-sizes and spacers, but they're related through line-height. Getting a bit too complicated for me, so I'm leaving it as-is ;)

admin/scss/bootstrap/_variables.scss
```diff
@@ -403,23 +403,23 @@ $navbar-padding-vertical:           ($spacer / 2);
...
-$nav-tabs-link-border-width:                0;
-$nav-tabs-link-hover-border-color:          green; //$gray-lighter;
-
-$nav-tabs-active-link-hover-bg:             $body-bg;
-$nav-tabs-active-link-hover-color:          $body-color;
-$nav-tabs-active-link-hover-border-color:   #ddd;
-
-$nav-tabs-justified-link-border-color:            #ddd;
-$nav-tabs-justified-active-link-border-color:     $body-bg;
+// $nav-link-padding:                          .5em 1em;
```
tractorcow: We should be removing not commenting out this code.

clarkepaul: We are uncommenting stuff as and when we require them and its acting as guide for the default settings. It makes it easier to see what we have customised and what we haven't changed yet. If we delete it all I think that people would create custom code instead of using bootstrap settings.

tractorcow: We'll need some comments at the top of this file in order to describe what these commented out lines mean. I assume you mean that these are the values as defined in _bootstrap.scss, in the node_modules, correct?

The only problem I want to solve is commented out code with no apparent value. If there is a reason, I just need to make sure others who view this file understand the same benefit that you do.

PR #5105, opened 2016-02-29
fix to support facebook autodiscover

mikeyc7m:
the autodiscover function was failing on facebook because it returned the wrong match. also added a thumbnail lookup for facebook.

see issue https://github.com/silverstripe/silverstripe-framework/issues/5066

PR #5322, opened 2016-04-15
fix to support facebook autodiscover

tractorcow:
Replaces https://github.com/silverstripe/silverstripe-framework/pull/5105

@mikeyc7m this is the level of unit testing we need for changes that involve parsing of html. :) I hope that helps you in future contributions!


dhensby:
thanks @tractorcow

PR #5327, opened 2016-04-17
no one writes as "한국어(韓國語)"  when referring to korean language

cloph:
so remove the chinese characters – "한국어(韓國語)" looks like
"English(anglicus)". Korean youth probably don't even know the meaning
of the chinese.


sminnee:
Thanks, @cloph it's great to get this kind of feedback from our community! :-)

PR #5329, opened 2016-04-17
Added placeholder panel for items with no preview

clarkepaul:
Would ideally become a component but have left the styles in Layout.scss for now.


clarkepaul:
Just noticed admin/javascript/src/styles/_styles.scss was a bit rushed and could be refactored better, also maybe they would fit better on a forms.scss better.

admin/javascript/src/components/breadcrumb/index.js
```diff
@@ -24,7 +24,7 @@ class BreadcrumbComponent extends SilverStripeComponent {
      ]),
        this.props.crumbs.slice(-1).map((crumb, index) => [
          <li className="breadcrumb__item breadcrumb__item--last">
-           <h2 className="breadcrumb__item-title breadcrumb__item-title--last" key={index}>
+           <h2 className="breadcrumb__item-title breadcrumb__item-title--last text-truncate" key={index}>
```
chillu: Adding a global `text-truncate` modifier is a bit of a BEM antipattern: http://getbem.com/faq/#can-i-create-global-modifier. Can we use a SCSS include instead, and apply it to `breadcrumb__item--title`?

clarkepaul: Yup sure, .truncate was created before hand but should be an easy fix.
The breadcrumb links when there are too many (top line) don't truncate at the moment and might be a bit fiddly to find a good solution.

PR #5330, opened 2016-04-17
Add babel-polyfill for better ES6 support

chillu:
See http://babeljs.io/docs/usage/polyfill/. Adds 30KB gzipped to the build, which isn't ideal. We can look into including more specific polyfills later. This fixes the campaign item view in IE for now. And is generally important since we don’t have CI notifying us of breakages like lacking `Array.find()` support in IE10, so very easy to introduce regressions there unless we have the polyfills in place.

Incidentally Sam had already added polyfill via https://github.com/silverstripe/silverstripe-framework/commit/b088efc6f9554b4750cb52626c3fcbf8b6a3f6db, but hadn't included it in the build.


package.json
```diff
@@ -35,7 +35,7 @@
   },
   "homepage": "https://github.com/silverstripe/silverstripe-framework#readme",
   "dependencies": {
-    "babel-polyfill": "^6.7.4",
+    "babel-polyfill": "https://registry.npmjs.org/babel-polyfill/-/babel-polyfill-6.7.4.tgz",
     "blueimp-file-upload": "^6.0.3",
```
sminnee: Why this change?

package.json
```diff
@@ -35,7 +35,7 @@
   },
   "homepage": "https://github.com/silverstripe/silverstripe-framework#readme",
   "dependencies": {
-    "babel-polyfill": "^6.7.4",
+    "babel-polyfill": "https://registry.npmjs.org/babel-polyfill/-/babel-polyfill-6.7.4.tgz",
```
hafriedlander: Why?

hafriedlander:
Haha. Sam and I picked up the same thing. We can try using tree-shaking or a library specific optimisation method if needed (like only including this in IE).

package.json
sminnee: snap.

hafriedlander: You sunk my battleship.

chillu: It seems to be related to running `npm install <package>` when it's already present in `npm-shrinkwrap.json`. We haven't tracked it down yet. I've manually fixed it for now.

PR #5334, opened 2016-04-18
FIX: service class name definitions incorrectly inherited by subclasses (fixes #4774)

### LINKED ISSUE #4774: Injector service definitions should not inherit to subclasses

kinglozzer:
Fix for #4774. I’m not sure whether the config locator should actually be aware of/care about what’s *in* the config it finds, but it seemed like the cleanest place to put this. If we put it here, we _know_ that the config that was found was inherited from a parent class - if we move it elsewhere then we have no way of knowing that.

I also figured that `class` is the only key of the config definition that shouldn’t be inherited - everything else makes sense when inherited afaict. 

Ping @tractorcow. Also pinging others who I can spot in the blame view for `Injector` for their input: @hafriedlander @nyeholt @chillu.

### DIFF
--- a/control/injector/SilverStripeServiceConfigurationLocator.php
+++ b/control/injector/SilverStripeServiceConfigurationLocator.php
@@ -33,6 +33,12 @@ class SilverStripeServiceConfigurationLocator extends ServiceConfigurationLocato
 				// have we already got for this?
 				$config = $this->configFor($parent);
 				if($config !== null) {
+					// Class name should not be inherited: attempting to instantiate a child class
+					// should never result in the parent class being created
+					if (isset($config['class'])) {
+						$config['class'] = $name;
+					}
+
 					// Cache this result
 					$this->configs[$name] = $config;
 					return $config;
--- a/tests/injector/InjectorTest.php
+++ b/tests/injector/InjectorTest.php
@@ -521,7 +521,36 @@ class InjectorTest extends SapphireTest {
 		$this->assertEquals($obj->one, 'the three');
 	}
 
-	public function testSameNamedSingeltonPrototype() {
+	public function testInheritedServiceDefinitions() {
+		$injector = new Injector(array('locator' => 'SilverStripeServiceConfigurationLocator'));
+
+		$this->assertInstanceOf('InjectorTest_CoreService', $injector->get('InjectorTest_CoreService'));
+		$this->assertInstanceOf('InjectorTest_CoreServiceSubclass', $injector->get('InjectorTest_CoreServiceSubclass'));
+
+		// Swap out the core service for a custom implementation
+		Config::inst()->update(
+			'Injector',
+			'InjectorTest_CoreService',
+			array('class' => 'InjectorTest_CustomCoreServiceImplementation')
+		);
+
+		// Use new Injector instance to avoid cached copies of configuration data
+		$injector = new Injector(array('locator' => 'SilverStripeServiceConfigurationLocator'));
+		// Attempting to create the core service should now find the custom implementation instead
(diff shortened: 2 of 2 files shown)


tractorcow:
I wonder if, assuming it's NOT safe to inherit the class name, why is it safe to inherit all other configs?

Assuming we configure dependencies with a certain structure, e.g.

```yaml
Injector:
  SomeService:
    class: SomeServiceImpl
    constructor: ['arg']
```

Having a class that extends SomeService would cause it to try to inject those constructor args into a class it was not defined for.... it could lead to unintended consequences where a certain config would not behave the way you would expect.

How about an alternative; Configs for parent classes only inherit if they have an `inherit: true` assigned. Otherwise, nothing is inherited. That way, we either inherit both class and other properties, or none at all, and it is opt in only.

E.g.

```yaml
Injector:
  SomeService:
    class: SomeServiceImpl:
    constructor: [`arg`]
    inherit: false
```

The default for inherit can be true in 3.x (to maintain semver), but we can turn it to false by default in 4.x.

How does this sound?

kinglozzer:
This scenario and fix only applies to subclasses, not implementors or anything. If you’ve got dependencies defined, those will be inherited automatically (which makes sense IMO, as the PHP class will already inherit the setter/getter and property). If you change the method signature of the setter, it’ll error.

If you’ve got constructor args defined, they’d be inherited too. Annoyingly, however, there’s no error raised if you mess around with constructor arguments, because PHP:

> Unlike with other methods, PHP will not generate an E_STRICT level error message when __construct() is overridden with different parameters than the parent __construct() method has.

I still don’t think that’s an issue, though. If you subclass something in PHP, you expect to inherit all the parent classes methods and properties. If you subclass something and mess around with the constructor without checking what the parent constructor does, then you’re asking for trouble!

I’m not sure that an extra config setting makes sense here:

- `inherit: false` - you get the correct class, but `$myServiceImpl->getSomeDependency()` would return null, so you have to duplicate the entire injector config from the parent class
- `inherit: true` - all the dependencies are set up, but the incorrect class is returned

tractorcow:
>  (which makes sense IMO, as the PHP class will already inherit the setter/getter and property). 

Unless you've declared those getters and setters on a more specific subclass, which you declare along with the `class` property in the config. The getters and setters will be inherited and will break.

Let me give an example:

```
class A {}
class B extends A {
	function getVal() {}
	function setVal($val) {}
}
class C extends A {
}
```

And this config

```yaml
Injector:
  A:
    class: B // Not inherited due to #5334
    properties: // Inherited
      Val: 'someval'
```

and then

```php
Injector::inst()->create('C') // Fails with attempt to invoke setVal() on class C.
```

kinglozzer:
Ahh, I see what you mean... that’s a bummer :(

PR #5338, opened 2016-04-19
Fixes a spacing issue for grouped buttons within btn-toolbar

clarkepaul:
Before: 
pasted_image_19_04_16__12_03_pm

After: 
pasted_image_19_04_16__12_02_pm


chillu:
On a more general note: I find the "invisible button" approach not very appealing because it means the icons don't align with the remaining text:

image

It makes sense on hover (button background style). Can we add a button outline here for the default state? Or make it a "proper link" without the button padding, and don't use a button style on hover.

image

clarkepaul:
I am wanting align the icon and have the hovered background slightly set to the left but only visible on hover, but I haven't worked out the exact overhang yet and didn't think it was critical yet, but yes it bugs me too.
pasted_image_20_04_16__10_33_am

PR #5346, opened 2016-04-19
Static file restructure

chillu:
/cc @silverstripe/core-team 

Depends on https://github.com/silverstripe/silverstripe-cms/pull/1456 and https://github.com/silverstripe/silverstripe-asset-admin/pull/117

Overview extracted from the upgrading guides:

All static files (images, javascript, stylesheets, fonts) used for the CMS and forms interfaces
in `framework` and `cms` have moved locations. These assets are now placed in a `client/` subfolder,
to account for a structural change where both javascript and styles are co-located in component-specific folders. This will affect you if you have used `Requirements::block()` on files in the `framework/` or `cms/` folder. Care should also be taken when referencing images in these folders from your own stylesheets.

```
framework/javascript       => framework/client/dist/
framework/images           => framework/client/dist/images/
framework/css              => framework/client/dist/css/
framework/scss             => framework/client/src/styles/
admin/javascript/          => admin/client/src/
admin/javascript/src/      => admin/client/src/legacy/ (mostly)
admin/scss/                => admin/client/styles/legacy/
admin/css/                 => admin/client/dist/css/
admin/css/screen.css       => admin/client/dist/css/bundle.css
admin/images/              => admin/client/dist/images/
admin/images/sprites/src/  => admin/client/src/sprites/
admin/images/sprites/dist/ => admin/client/dist/sprites/
admin/font/                => admin/client/dist/font/
```

This all started with renaming the `javascript/` folder to `client/` - we followed BEM and wanted to write ReactJS components, which mean encapsulating component styles as well as JavaScript. It made sense to put these all in one place, alongside component specific documentation and tests. You can see this system in action over at Wordpress Calypso as well: https://github.com/Automattic/wp-calypso/tree/master/client/components

With a `client/` folder in place, the logical next step was to also move `css/` and other static assets over there. The folder encapsulates the non-PHP aspects of the codebase quite nicely now.

We've consciously decided not to call this folder `public/`, since it contains non-public assets, and isn't considered the actual webroot (a common convention in other projects around the NodeJS and Ruby space).

I've also moved most existing CSS and JS into respective `legacy/` folders, to make it clearer what's old and new code. We're trying to establish standards around writing these formats, and plan to iteratively migrate functionality over. Hence we need a clear indicator on what files are considered best practice and which ones need work.

There's a few rough edges to look at - I don't think any of these block the PR merge though:
- Solution for `lang/*.js` and `lang/src/*.js` files (will require releases tooling change)
- The 'dist/` folder contains a mix of static files (`images/`), as well as generated files (`images/sprites/`, `css/`). Not ideal since you'd usually expect to be able to completely rebuild a `dist/` folder from scratch through build tooling. I think that's a necessary evil until we come up with a wider SilverStripe module standard for declaring a "public" resources and some automation to move files around.
- We could make better use of PHP constants, a lot of repitition in the file paths at the moment

This is a BC breaking change for any module or project relying on static file locations in core. But given the scope of changes we're implementing across the CMS, its probably not a good idea to rely on those anyway. The one use case of `Requirements::block()` will become less common as ES6 modules favour a different workflow. I'm also keen to reduce the number of JS files we expose and load due to the same reason, but that's for another pull request.

I've tested this by installing the [frameworktest](https://github.com/silverstripe-labs/silverstripe-frameworktest) module, which adds heaps of form field variations to test pages in the CMS. I've then tested the UI side-by-side in Chrome to spot any visual differences, clicking through each major section and performing most common actions (add page, batch actions, search, upload, etc). Then I've sanity checked in IE10 with a reduced set of actions (IE10 testing is a pain at the moment because the routing is broken...).

Git seems to track the renames correctly, which should make it easier to merge existing pull requests on top of this. Nevertheless, as you can imagine I want to get this pull request over the line quickly to avoid conflicts.

Note that our [module standard](https://www.silverstripe.org/software/addons/supported-modules-definition/) doesn't prescribe folder naming or placement for frontend tooling.

Todo:

- [x] Fix "emtpy" sprite (`sprite-..png`)
- [x] Check Behat tests


chillu:
I've collated feedback from @hafriedlander @flashbackzoo and @tractorcow. A summary of the discussion:

- Hamish: Directory paths are a bit long. Ingo: Not really an issue with fast-find in most IDEs, and any less folder depth will remove semantic value
- David: Inconsistent naming between `src/styles` and `dist/css`. Ingo: Fixed this by renaming to `dist/styles`.
- Hamish: Ideally we'd change the Requirements API to be more BC break resilient against file name changes (e.g. through aliasing), but not blocking this PR.

chillu:
Merging on behalf of, and with the approval of @tractorcow - I asked him to hold off on merge until I squashed the last commits, and ensure git picks up the renames properly (rather than relying on `git merge --squash` through the Github UI).

PR #5349, opened 2016-04-19
FIX: Hierarchy was incorrectly unexpanding nodes that were previously expanded

madmatt:
Given the following structure with a TreeMultiselectField:
image

If this is saved and then you re-open the tree, you'll see the following bug:
screencast-2016-04-20 11-35-39-f1dv2

Opening the tree marks the field as dirty, and if you save the object it will remove the relations that are no longer ticked (`Grandchild 1` and `Grandchild 2` in this case).

This is due to a problem in `Hierarchy` when it recurses through nodes to find ones that need to be marked for inclusion when generating the tree structure. It correctly marks `Child 1` to be expanded, but then when it marks `Child 2`, it marks `Child 1` as not expanded but having children (this also explains the less serious bug with the arrow pointing down when it shouldn't).

XHProf comparison before and after shows it doesn't appear to recurse any further than it did previously.

Changes to other tests aren't because of this bug, but just because I had to add more children/grandchildren to confirm the bug is fixed.

### DIFF
--- a/model/Hierarchy.php
+++ b/model/Hierarchy.php
@@ -315,7 +315,8 @@ class Hierarchy extends DataExtension {
 			foreach($children as $child) {
 				$markingMatches = $this->markingFilterMatches($child);
 				if($markingMatches) {
-					if($child->$numChildrenMethod()) {
+					// Mark a child node as unexpanded if it has children and has not already been expanded
+					if($child->$numChildrenMethod() && !$child->isExpanded()) {
 						$child->markUnexpanded();
 					} else {
 						$child->markExpanded();
--- a/tests/model/HierarchyTest.php
+++ b/tests/model/HierarchyTest.php
@@ -61,12 +61,12 @@ class HierarchyTest extends SapphireTest {
 		// Obj 3 has been deleted; let's bring it back from the grave
 		$obj3 = Versioned::get_including_deleted("HierarchyTest_Object", "\"Title\" = 'Obj 3'")->First();
 
-		// Check that both obj 3 children are returned
-		$this->assertEquals(array("Obj 3a", "Obj 3b", "Obj 3c"),
+		// Check that all obj 3 children are returned
+		$this->assertEquals(array("Obj 3a", "Obj 3b", "Obj 3c", "Obj 3d"),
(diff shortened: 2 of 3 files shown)


tractorcow:
Please fix the tests. :D

The regression is caused by an assumption about intrinsic ordering of items in the hierarchy; If you don't explicitly order them, it's non-deterministic. Postgres is simply pointing out the error in our assumption. ;)

Actual code seems fine.

tractorcow:
Adding default_sort on Title should be enough.

madmatt:
Oops, missed your comments sorry @tractorcow, have added this now. Will see what postgres does!

tractorcow:
Absolute @madmatt :D

tractorcow:
Notice an error when testing permissions tab; Can you please check locally to see if this error is real?

dhensby:
bump

madmatt:
The failing Behat test seems to work fine locally but I'm not running it the same way Travis is. Have just force-pushed a rebase up so will see if that fixes the issue or not

madmatt:
@tractorcow, @dhensby: Yay, it works now! Now I'm not sure if the 3 branch was the right place to merge into, but it should all be up to date, I've rebased on top of 3 today. Let me know if it's wrong :)

dhensby:
looks like a bug fix, so could go into 3.2 if you're so inclined?

madmatt:
I don't know what the usual practice is TBH, happy to be guided. I'm happy for it to just go out with the next 3.x release, whenever that is - for now I'll just have to maintain a fork but it's not a massive bug, and is easy to keep rebasing as needed.

dhensby:
@madmatt I'd like it in 3.2 and we're hoping to do a release pretty soon, so if you PR against 3.2 it could make the next set of stables.

I'm sure you know how to do this but please can you:

1. Close PR
2. New branch from 3
3. Cherry pick this commit (6168ced)
4. Open a new PR against 3 (mentioning this commit)

madmatt:
@dhensby This is already a PR against the `3` branch actually - did I do it wrong?

dhensby:
@madmatt derp - I meant 3.2

madmatt:
Closed in favour of opening #5823

PR #5358, opened 2016-04-20
API redux-logger respects ss environment

tractorcow:
So that you don't get console spam when deploying or testing.

Rebased on top of https://github.com/silverstripe/silverstripe-framework/pull/5346 because that's getting merged soon.


tractorcow:
Rebased. :D

sminnee:
Seems reasonable to me. Including environment type in the standard config doesn't seem like it would introduce risk on production environments.

chillu:
Tested locally, and rebased. We'll need to apply the same to the `asset-admin` module - although I'm a bit confused why this module has a `rootReducer` as well, it should just use the `lib/ReducerRegister` functionality.

sminnee:
> Although I'm a bit confused why this module has a rootReducer as well, it should just use the lib/ReducerRegister functionality.

I think its implementation predates the existence of lib/ReducerRegister. Agree that refactor would make sense.

PR #5362, opened 2016-04-21
Updated icon set with new icons and license, removed two unused stylesheets

clarkepaul:
Added back button, and happy and sad faces for notifications
Moved button styles to action component.


admin/client/src/components/form-action/styles.scss
```diff
@@ -95,6 +95,10 @@
 	&:focus {
 		background-color: $gray-lighter;
 	}
+
+	.toolbar--content & {
```
chillu: That's technically not BEM, but i'll let it slide for now because we need some class refactor in there anyway

PR #5366, opened 2016-04-21
Fix AssetAdmin icon reference

chillu:
Needs to use namespaced class name (see LeftAndMain->Menu()).
Not replacing the old .icon-assetadmin rule since we want the old and new UI to co-exist for a short while.
Ideally we'd move these styles to the asset-admin module, but the whole way of referencing image sprites is outdated anyway - so I'd prefer to merge this quick fix and think about SVG icons for menu entries instead.

Merge with https://github.com/silverstripe/silverstripe-asset-admin/pull/119


tractorcow:
Thanks me!

PR #5368, opened 2016-04-21
Add AddToCampaign dialog box functionality

hafriedlander:


tractorcow:
Also see https://github.com/silverstripe/silverstripe-cms/pull/1459

admin/client/src/styles/legacy/_style.scss
```diff
@@ -2043,6 +2043,110 @@ body.cms-dialog {
 }
 
 /** --------------------------------------------
+ * Add to campaign dialog
+ * -------------------------------------------- */
+
+#add-to-campaign__dialog {
```
tractorcow: We shouldn't be using ID selectors in CSS.

admin/code/AddToCampaignHandler.php
```diff
@@ -0,0 +1,232 @@
...
+ *
+ *     function addtocampaign($data, $form) {
+ *         $handler = AddToCampaignHandler::create($form, $data);
+ *         return $handler->handle();
+ *     }
+ *
+ *  and add an AddToCampaignHandler_FormAction to the EditForm, possibly through getCMSActions
+ */
+class AddToCampaignHandler {
+
```
tractorcow: If you `use Injectable` then you can use `::create` in CMS to construct this.

admin/code/AddToCampaignHandler.php
```diff
@@ -0,0 +1,232 @@
...
+	/**
+	 * @var Form - the EditForm that contains the action we're being delegated to from
+	 */
+	protected $editForm;
+	/**
+	 * @var [string] - the submitted form data
```
tractorcow: `@var string` no square brackets in PHPdoc.

admin/code/AddToCampaignHandler.php
```diff
@@ -0,0 +1,232 @@
...
+ *         $handler = AddToCampaignHandler::create($form, $data);
+ *         return $handler->handle();
+ *     }
+ *
+ *  and add an AddToCampaignHandler_FormAction to the EditForm, possibly through getCMSActions
+ */
+class AddToCampaignHandler {
+
+	/**
+	 * @var Form - the EditForm that contains the action we're being delegated to from
```
tractorcow: You don't use hyphens to delimit phpdoc.

admin/code/AddToCampaignHandler.php
```diff
@@ -0,0 +1,232 @@
...
+	 */
+	public function __construct($editForm, $data) {
+		$this->editForm = $editForm;
+		$this->data = $data;
+	}
+
+	/**
+	 * Perform the action. Either returns a Form or performs the action, as per the class doc
+	 *
+	 * @return HTML|SS_HTTPResponse
```
tractorcow: Sorry, HTML isn't a class. Did you mean `DBHTMLText`?

admin/code/AddToCampaignHandler.php
tractorcow: Sorry, array of strings is `@var string[]` :P

Its also not an array of explicitly strings; It can be nested arrays. You probably just want to cast this as `@var array`

admin/code/AddToCampaignHandler.php
```diff
@@ -0,0 +1,232 @@
...
+	/**
+	 * @var [string] - the submitted form data
+	 */
+	protected $data;
+
+	/**
+	 * AddToCampaignHandler constructor.
+	 *
+	 * @param $editForm - The parent form that triggered this action
+	 * @param $data - The data submitted as part of that form
```
tractorcow: No hypens please. :D

admin/code/AddToCampaignHandler.php
```diff
@@ -0,0 +1,232 @@
...
+			return $this->Form($object)->forTemplate();
+		}
+		else {
+			return $this->addToCampaign($object, $this->data['Campaign']);
+		}
+	}
+
+	/**
+	 * Get what ChangeSets are available for an item to be added to by this user
+	 * @return ArrayList[ChangeSet]
```
tractorcow: I don't think you can type-cast non-array collections.

tractorcow:
I'm going to stop reviewing and just push up phpdoc fixes. :D

admin/code/AddToCampaignHandler.php
```diff
@@ -0,0 +1,232 @@
...
+		$inChangeSets = array_unique(ChangeSetItem::get_for_object($object)->column('ChangeSetID'));
+		$changeSets = $this->getAvailableChangeSets()->map();
+
+		$campaignDropdown = DropdownField::create('Campaign', '', $changeSets);
+		$campaignDropdown->addExtraClass('noborder');
+		$campaignDropdown->setDisabledItems($inChangeSets);
+
+		$fields = new FieldList([$campaignDropdown]);
+
+		foreach ($this->data as $k => $v) {
```
tractorcow: I thought you were going to change this? I'm still getting `array to string conversion` when pushing data that is nested array content.

tractorcow: You can change this to a recursive walk (probably in another method) that knows how to reconstruct a set of Name[Field] hiddenfields for each nested array.

admin/client/src/legacy/AddToCampaignForm.js
```diff
@@ -9,7 +9,7 @@ $.entwine('ss', function($){
             if (dialog.length) {
                 dialog.open();
             } else {
-                dialog = $('<div id="add-to-campaign__dialog" />');
+                dialog = $('<div id="add-to-campaign__dialog" class="add-to-campaign__dialog" />');
```
hafriedlander: Why?

admin/code/AddToCampaignHandler.php
```diff
@@ -49,91 +58,96 @@ public function __construct($editForm, $data) {
...
+	 * @return DataObject The referenced DataObject
 	 * @throws SS_HTTPResponse_Exception
 	 */
 	protected function getObject($id, $class) {
 		$id = (int)$id;
 		$class = ClassInfo::class_name($class);
 
 		if (!$class || !is_subclass_of($class, 'DataObject') || !Object::has_extension($class, 'Versioned')) {
-			return $this->editForm->httpError(400, _t(
+			$this->editForm->httpError(400, _t(
```
hafriedlander: Nope, I like the return here. Although httpError is supposed to throw an exception, if it doesn't for some reason (future code change?) the rest of the block would execute without the return.

admin/client/src/legacy/AddToCampaignForm.js
tractorcow: Because we shouldn't rely on IDs for styles. I thought it was bad practice.

admin/code/AddToCampaignHandler.php
tractorcow: httpError doesn't return anything, but it does always throw an exception. You can't return a function that has a void return type (it causes php style warnings).

It would be fine to `return null;` if you're concerned about future changes.

admin/client/src/legacy/AddToCampaignForm.js
hafriedlander: Is this in the BEM style guides somewhere? If there's only ever one of anything I don't see the problem (given it's attached to body and never removed, we _want_ there to only be one, not an ever-increasing number). But if it's against style, sure.

clarkepaul: You shouldn't use id's, we will be using the Airbnb css guides to get going and it also states in there don't use ids https://github.com/airbnb/css and also in BEM docs http://getbem.com/naming/

admin/code/AddToCampaignHandler.php
tractorcow: I've amended to include a specific return, although not in the same statement as the `httpError` invocation.

PR #5369, opened 2016-04-21
[BUG] Updating Upload_Validator file size spelling

timkung:
Changing all cases of "filesize" spelling to "file size" and related translatable files, so spelling correct in the error messages


sminnee:
The change looks good to me, and I think the wording is an improvement, but I can't confirm whether this will persist into transifex properly. @chillu?

chillu:
The `en` master strings will be pushed as part of the normal release process. The translations (e.g. `en_GB`) won't update automatically, unless we specifically push them up. So some of these strings in the commit will eventually be reverted, but every translator sees the master string as changed in transifex and has the option to update their respective translation anyway.

PR #5370, opened 2016-04-21
BUG Fix error when modals are displayed

tractorcow:
Fixes regression in https://github.com/silverstripe/silverstripe-framework/pull/3900


tractorcow:
Once merged https://github.com/silverstripe/silverstripe-cms/pull/1457 should pass.

Issue #5367, opened 2016-04-21
CSS issues on SilverStripe 4.0


ARNHOE:
To start of I am fully aware 4.0 is unreleased and perhaps its a problem on my end. Although I did everything from caching clearing to double checking my composer settings etc...

I am using composer to get the latest changes but since day one I got problems with the admin panel. For some reason loads of CSS issues.

![schermafbeelding 2016-04-21 om 10 52 59](https://cloud.githubusercontent.com/assets/1517053/14704185/b9e45816-07b3-11e6-8147-39b701757d59.png)
Batch Actions is using the font settings of:
`.ui-widget input, .ui-widget select, .ui-widget textarea, .ui-widget button {
    font-family: Verdana,Arial,sans-serif;
    font-size: 1em;
}`

![schermafbeelding 2016-04-21 om 10 58 45](https://cloud.githubusercontent.com/assets/1517053/14704204/d66b6268-07b3-11e6-92b9-2f1234769445.png)

![schermafbeelding 2016-04-21 om 11 09 29](https://cloud.githubusercontent.com/assets/1517053/14704206/d9312ee2-07b3-11e6-96d6-8399e6205752.png)

![schermafbeelding 2016-04-21 om 11 26 22](https://cloud.githubusercontent.com/assets/1517053/14704215/e46ae1cc-07b3-11e6-8529-978c15987bae.png)

Perhaps the above is valuable information.

sminnee:
Yeah, those buttons are not supposed to be like that. It smells like a bug in the compilation of CSS from SCSS.

I see it too on master w/ Chrome

sminnee:
Recompiling the SCSS didn't fix It. @clarkepaul @flashbackzoo, one for you?

dhensby:
Maybe from the reorganisation by @chillu

flashbackzoo:
I can have a look today

flashbackzoo:
Git bisect tells me https://github.com/silverstripe/silverstripe-framework/commit/3a778f23bfdc6d00a716da1fdc18c847129ab2eb

sminnee:
Oh dear, that's not going to tell us much. Looking at the parent commits, my guess would be this one: 35e062d9f32df475eb41c78bd709059b417b8b01

flashbackzoo:
Paul and I will sit down and nut it out now :smile:

flashbackzoo:
https://github.com/silverstripe/silverstripe-framework/pull/5371

[...]

PR #5371, opened 2016-04-21
Fix regression with font-icon styles

flashbackzoo:
Before:
icons-framework-before

After:
fix-icons-framework

The issue was this block of CSS which was added by https://github.com/silverstripe/silverstripe-framework/commit/35e062d9f32df475eb41c78bd709059b417b8b01 
```css
[data-icon]:before {
  font-family: "silverstripe" !important;
  content: attr(data-icon);
  font-style: normal !important;
  font-weight: normal !important;
  font-variant: normal !important;
  text-transform: none !important;
  speak: none;
  line-height: 1;
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
 }
```

JQueryUI adds the `data-icon` attribute to our legacy buttons, so legacy buttons were getting some of these new button styles applied.

I've also moved the font styles back to `_fonts.scss` because we were referencing a dist CSS file in our src SCSS build which is bad.

Also renamed the reference CSS filename to to `styles.css` to match the filename you get in the Fontastic bundle. This will make it less confusing when adding / removing font icons in the future.

I recommend bringing fonts / font-icons into our build tooling at some stage to avoid issues like this. Fontastic added the above rule to their bundle which we download and copy across manually.


flashbackzoo:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/5367

PR #5372, opened 2016-04-21
Rename JS files to comply with AirBnB conventions

chillu:
Merged together with https://github.com/silverstripe/silverstripe-cms/pull/1461 and https://github.com/silverstripe/silverstripe-asset-admin/pull/120

See detailed comments in commit messages.

Before:

```
components/
  grid-field/
    index.js // exports GridFieldComponent
    row.js // exports GridFieldRow
    style.scs
sections/
  campaign-admin/
    controller.js // exports CampaignAdminContainer
    editor.js // exports CampaignAdminEditor (doesn't exist, just an example)
```

After:

```
components/
  GridField/
    GridField.js // exports GridField
    GridFieldRow.js // exports GridFieldRow
    GridField.scss
containers/
  CampaignAdmin/
    CampaignAdmin.js // exports CampaignAdmin
    CampaignAdminEditor.js // exports CampaignAdminEditor
```

The double import reference (`import FormAction from components/FormAction/FormAction`) is a necessary evil of placing each component in its own folder, so we can keep README.md and SCSS on the same level (per component). AirBnB implicitly allows `index.js` if the folder is named after the component, but that would be quite inconsistent in case there's more than one component in a folder. We could look into auto-creating an index.js folder which exports all sibling files (https://github.com/gajus/create-index).

```
// index.js
import GridField from './GridField';
import GridFieldRow from './GridFieldRow';
export GridFieldRow;
export default GridField;

// some-file.js
import { GridField } from 'components/GridField';
```


sminnee:
> The double import reference (import FormAction from components/FormAction/FormAction) is a necessary evil of placing each component in its own folder.

I really don't think this is a particularly big deal and I would suggest we let things settle a bit before pulling in an index.js generator — the build stack is already complex enough!

chillu:
Looked this over with @tractorcow and @flashbackzoo - they've both done a partial peer review. Merging now, since it's starting to inhibit our workflow around other pull requests.

PR #5379, opened 2016-04-22
BUG Fix baseurl in IE missing leading /

tractorcow:
Reformat LeftAndMain.js to new indentation standard, since editorconfig enforces a style different to that it currently follows.


tractorcow:
See https://github.com/silverstripe/silverstripe-framework/pull/5379/files?w=1 for changes without whitespace.

PR #5383, opened 2016-04-23
Insert link select action

jonom:
Currently if you want to link to an `$Action` or `$Action/$ID` on an internal page from a HtmlEditorField instance you need to choose 'External website' as the link type and sacrifice the benefits of linking to a page with a shortcode. Same applies if you want to include any URL query params.

This PR aims to fix that, letting you specify anything that should come after the page URL and before the # fragment identifier by introducing an 'Action' field.

screen shot 2016-04-22 at 8 48 03 pm

When you choose a page to link to, a dropdown field is populated with actions that are allowed on that page, similar to the way anchor selection currently works.

I've also tweaked the behaviour of the anchor dropdown so that it is only shown when it is not empty.

I looked in to writing a behat test for this but wasn't sure how to mimic a page that would have some $allowed_actions on it.

### DIFF
--- a/forms/HtmlEditorField.php
+++ b/forms/HtmlEditorField.php
@@ -159,6 +159,7 @@ class HtmlEditorField_Toolbar extends RequestHandler {
 		'LinkForm',
 		'MediaForm',
 		'viewfile',
+		'getactions',
 		'getanchors'
 	);
 
@@ -238,6 +239,7 @@ class HtmlEditorField_Toolbar extends RequestHandler {
 						array(
 							'internal' => _t('HtmlEditorField.LINKINTERNAL', 'Page on the site'),
 							'external' => _t('HtmlEditorField.LINKEXTERNAL', 'Another website'),
+							'action' => _t('HtmlEditorField.LINKACTION', 'Action on this page'),
 							'anchor' => _t('HtmlEditorField.LINKANCHOR', 'Anchor on this page'),
 							'email' => _t('HtmlEditorField.LINKEMAIL', 'Email address'),
 							'file' => _t('HtmlEditorField.LINKFILE', 'Download a file'),
@@ -252,6 +254,7 @@ class HtmlEditorField_Toolbar extends RequestHandler {
 					TextField::create('external', _t('HtmlEditorField.URL', 'URL'), 'http://'),
 					EmailField::create('email', _t('HtmlEditorField.EMAIL', 'Email address')),
 					$fileField = UploadField::create('file', _t('HtmlEditorField.FILE', 'File')),
+					TextField::create('Action', _t('HtmlEditorField.ACTIONVALUE', 'Action')),
 					TextField::create('Anchor', _t('HtmlEditorField.ANCHORVALUE', 'Anchor')),
 					TextField::create('Subject', _t('HtmlEditorField.SUBJECT', 'Email subject')),
 					TextField::create('Description', _t('HtmlEditorField.LINKDESCR', 'Link description')),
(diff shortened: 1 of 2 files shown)


jonom:
I called the field 'Action' because of SilverStripe terminology but maybe "URL Suffix" would make more sense for users? Probably more accurate too since you can put more than just an Action in there.

forms/HtmlEditorField.php
```diff
@@ -561,6 +564,48 @@ public function viewfile($request) {
...
+					_t(
+						'HtmlEditorField.ANCHORSCANNOTACCESSPAGE',
+						'You are not permitted to access the content of the target page.'
+					),
+					403
+				);
+			}
+
+			// Get lowercase actions
+			// ToDo: look up controller instead of assuming use of _Controller naming convention
```
jonom: I could use some help with this part... I assume Director is able to work out what controller to use for a given class name but I couldn't find where the magic happens.

dhensby:
This is cool, I'm not in a position to review it right now as I'm on my phone. There must be a better way to get the allowed action and the controller name, but we'd ideally like to do it without having to instantiate the controller via `ModelAsController::controller_for()`

kinglozzer:
Hmm, I’m a bit unsure about this one personally. Custom page actions can be ones which both site users and CMS users are completely unaware of the existence of, and often have some sort of access control via permission checks or request checks (AJAX, SecurityID etc). My main concern is that we’re both enabling users to create broken links for many of the actions we give them in the list, and overwhelming them with options that they’re unlikely to use. I also don’t think we can safely assume that all forms handlers are mixed case.

I think we’d be better served by the label “URL suffix”, and forgetting the actions dropdown. How about merging it in with the Anchor option? I.e. rename it to “URL suffix”, remove the auto `#` (if there is one), and keep the “Select an anchor” dropdown just below it.

jonom:
> This is cool, I'm not in a position to review it right now as I'm on my phone

No rush, enjoy your weekend :)

> My main concern is that we’re both enabling users to create broken links for many of the actions we give them in the list, and overwhelming them with options that they’re unlikely to use.

Yeah I wasn't sure if the action selector was a good default inclusion either. It's handy for me as a power user but I can see how it could create confusion for your average CMS user. I could disable this by default with a config setting but I'm not sure people will know they can enable it so maybe best just to remove the dropdown altogether? Would make this change a lot simpler.

>  I think we’d be better served by the label “URL suffix”, and forgetting the actions dropdown. How about merging it in with the Anchor option?

I think it would be nice to have all the URL additions in one field, and the anchor dropdown could just manipulate the applicable portion of the text in that field. I'd need to point this at master instead in that case I think as a potentially breaking change. Maybe we keep them separate in SS3 and combine in to one field in SS4? Or skip SS3?

sminnee:
Because it's a dropdown, the solution that you've proposed doesn't support querystring parameters.

It sounds like this is and advanced user feature. As suggested above, it might be better to make it a text field called "URL Suffix". However, I would also hide it within an "Advanced options" expandable section.

Note: if you put it into an Advanced options section, then that section should be expanded by default when editing a URL that has a suffix value already on it.

Arguably, Anchor could also got into an advanced options section.

jonom:
> Because it's a dropdown, the solution that you've proposed doesn't support querystring parameters.

The dropdown is just a starting point for pre-filling the Action / URL-Suffix field, so query parameters can be added manually. As suggested previously though I think it would be best to remove the dropdown all together.

If we convert the Anchor text field to a URL Suffix field (and let the anchor dropdown simply manipulate the anchor portion of the text in that field) then there's probably no need for an advanced section as we'll be dealing with the same number of fields as we currently have. To make it worthwhile you'd need to put "Link description" and "Open in a new window" under advanced options too I'd say. Which might not be a bad idea.

forms/HtmlEditorField.php
tractorcow: ModelAsController::controller_for()

tractorcow:
Could you implement it as a combo-box that can be edited after the fact with querystring args? That way you could probably make it a grouped combo box that include both anchors and actions in it.

```
[text]
(dropdown
  actions:
    - action 1
    - action 2
  anchors:
    - #anchor1
    - #anchor2
)
```

jonom:
> Could you implement it as a combo-box that can be edited after the fact with querystring args

That's a cool idea! But it won't mitigate Loz's concern which I think is fair:

> Custom page actions can be ones which both site users and CMS users are completely unaware of the existence of, and often have some sort of access control via permission checks or request checks (AJAX, SecurityID etc). My main concern is that we’re both enabling users to create broken links for many of the actions we give them in the list (...)

I think the best solution here is to open a PR against 4 instead that does this:

- Change Anchor text field to a URL Suffix text field
- Anchor dropdown field remains but is only visible when it contains options
- When choosing an anchor from the dropdown it manipulates the value of the URLSuffix field, changing only the anchor portion. So a value of '?start=2#old-anchor' would become '?start=2#new-anchor', or an empty value would become '#new-anchor'

This would give people a lot more freedom with internal URLs without sacrificing anything.

To be honest I'm not sure I'll have time to do this in time for a 4 release so if anyone else wants to have at it feel free! I'm sure I could have it done in time for 5 though ;)

Will close this for now.

PR #5384, opened 2016-04-24
FIX Admin bloacklisted messages using correct $.inArray check

dhensby:
See discussion #5312 

The $.inArray returns the index of the found item and `-1` if there is no match.

### DIFF
--- a/admin/javascript/LeftAndMain.js
+++ b/admin/javascript/LeftAndMain.js
@@ -150,7 +150,7 @@ jQuery.noConflict();
 			}
 
 			// Show message (but ignore aborted requests)
-			if(xhr.status !== 0 && msg && $.inArray(msg, ignoredMessages)) {
+			if(xhr.status !== 0 && msg && $.inArray(msg, ignoredMessages) != -1) {
 				// Decode into UTF-8, HTTP headers don't allow multibyte
 				statusMessage(decodeURIComponent(msg), msgType);
 			}


admin/javascript/LeftAndMain.js
```diff
@@ -150,7 +150,7 @@ jQuery.noConflict();
 			}
 
 			// Show message (but ignore aborted requests)
-			if(xhr.status !== 0 && msg && $.inArray(msg, ignoredMessages)) {
+			if(xhr.status !== 0 && msg && $.inArray(msg, ignoredMessages) != -1) {
```
tractorcow: You said in https://travis-ci.org/silverstripe/silverstripe-framework/builds/125725811 that it would be `== -1` not `!= -1`

tractorcow:
You got your logic backwards. :D

tractorcow:
As a note, I'm merging this up from 3 -> master with this code.

```js
// Show message (but ignore aborted requests)
    if (xhr.status !== 0 && msg && $.inArray(msg, ignoredMessages) === -1) {
      // Decode into UTF-8, HTTP headers don't allow multibyte
      statusMessage(decodeURIComponent(msg), msgType);
    }
```

tractorcow:
Note: Close https://github.com/silverstripe/silverstripe-framework/issues/5281 again once this is merged. :D

tractorcow:
Merged with https://github.com/silverstripe/silverstripe-framework/commit/71a4f1e605d35fec63165a2036110d3cf38d08a1

giankdm:
I've got a similar problem on v3.4. In the dev env the issue doesn't manifest, however in the stage env on the destination server I've got a slew of notifications. It happens in OSX Safari, doesn't happen in OSX Google Chrome. I think a new browser update added to the problem.
screenshot 2016-06-27 alle 16 55 50

PR #5388, opened 2016-04-25
Remove invalid attributes

axllent:
Headers do not have `type` & `name` attributes which are being inherited from the extended classes.


tractorcow:
I'm restarting the test... that cache test is terribly unreliable and needs removing / rewriting.

PR #5391, opened 2016-04-26
Toolbar south fix to make full-width

clarkepaul:
This is branched off https://github.com/open-sausages/silverstripe-framework/tree/pulls/4.0/campaign-admin-add so that should be merged prior.

Issue:
pasted_image_26_04_16__3_18_pm


clarkepaul:
Only the last commit 0bbdc9c is needed for the fix, the rest comes from the original branch.

PR #5392, opened 2016-04-26
Allow THEME_DIR to be customised

wilr:
Adds flexibility for custom setups such as app/themes to be implemented in the case a project wants to provide a custom path.

### DIFF
--- a/core/Constants.php
+++ b/core/Constants.php
@@ -229,10 +229,19 @@ if(!defined('BASE_URL')) {
 		define('BASE_URL', rtrim($dir, DIRECTORY_SEPARATOR));
 	}
 }
-define('MODULES_DIR', 'modules');
+
+if (!defined('MODULES_DIR')) {
+	define('MODULES_DIR', 'modules');
+}
+
 define('MODULES_PATH', BASE_PATH . '/' . MODULES_DIR);
-define('THEMES_DIR', 'themes');
+
+if (!defined('THEMES_DIR')) {
+	define('THEMES_DIR', 'themes');
+}
+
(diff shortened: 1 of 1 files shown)


Zauberfisch:
good to merge.

@wilr though have you searched for hard coded references to `themes/`? I wouldn't be surprised to find a few across framework and cms.

tractorcow:
@wilr you pushed your wip branch directly to origin. :)

core/Constants.php
```diff
@@ -229,10 +229,19 @@ function stripslashes_recursively(&$array) {
 		define('BASE_URL', rtrim($dir, DIRECTORY_SEPARATOR));
 	}
 }
-define('MODULES_DIR', 'modules');
+
+if (!defined('MODULES_DIR')) {
+	define('MODULES_DIR', 'modules');
```
tractorcow: We have NEVER respected this constant before... I'm surprised to see it actually.

tractorcow:
I think you'll need to fix hard-coded themes paths before getting this over the line... I can see a few.

- i18n.php
- i18nTextCollector.php

:D

tractorcow:
I've backed up your work to https://github.com/tractorcow/silverstripe-framework/tree/wilr-custom-theme-dir... I'll remove your branch and let you create another PR from your fork. Is that ok?

wilr:
Didn't we decide to implement merge by default @tractorcow for PRs in the last core chat.. Simple change that had no side effects. Now requires more effort.. #justsaying

tractorcow:
Yeah but there are side effects if you change themes path and there are hard coded references. :)

PR #5394, opened 2016-04-26
FIX: GridFieldPaginator now prevents viewing pages with no results (fixes #3192)

### LINKED ISSUE #3192: searching (find) for file in UploadField does not find file if you are on the wrong page.
Summary: searching (find) for file in UploadField does not find file if you are on the wrong page.

This bug can be replicated as follows:

GO TO:
http://demo.silverstripe.org/admin/pages/edit/show/7197

Click on the "insert Media" button and search for community:
image

Now, move to page "two" and do the same thing:
image

I dont have time right now to investigate in more detail, but when I looked for open issues on Uploadfield it appears to me that a quarter of all open bugs in Framework relate to the Uploadfield: https://github.com/silverstripe/silverstripe-framework/search?q=Uploadfield&ref=cmdform&type=Issues (~450 open issues, ~132 bugs with the word Uploadfield in it).

I'd say it might be a good idea for someone to spend a day just doing those as it would make a lot of people very happy.  I am sure a lot of the uploadfield issues are interrelated. 

I hope this makes sense and I got the facts right ;-)

THANK YOU THANK YOU.

kinglozzer:


kinglozzer:
👍

Issue #5222, opened 2016-03-23
Upload_Validator::getAllowedMaxFileSize() not checking against * property


timkung:
In the getAllowedMaxFileSize method on Upload_Validator, if an extension parameter is passed into the method it does not check the $allowedMaxFileSize array to see if a "*" (asterisk) key has been set to limit the file size on all file types (as I am assuming this is meant to work?).

This means you would have to define all the max file sizes for every type of extension (or category, e.g. using the '[image]' key to capture all the defined image extensions on File::$app_categories), rather than being able to just set a global "2M" for all file extensions through the "*" key.

Should be able to easily fix this by adding the following check after/before the category check:
```
if (isset($this->allowedMaxFileSize['*'])) {
  return $this->allowedMaxFileSize['*'];
}
```

tractorcow:
Or just un-nest the final `[*]' check at the end, so that it runs after the `if($ext)` block without returning immediately.

Pull request please. :D

timkung:
Here we go :)

https://github.com/silverstripe/silverstripe-framework/pull/5227

tractorcow:
Since we couldn't get the test AND the fix in the same pull request, I've merged these two into https://github.com/silverstripe/silverstripe-framework/pull/5444.

kinglozzer:
Closed via https://github.com/silverstripe/silverstripe-framework/commit/36e64168a53854bed8bcd779d9ba6fcac0810c0c

PR #5396, opened 2016-04-26
test to prove #5222

christopherdarling:


tractorcow:
Merged this test with the fix at https://github.com/silverstripe/silverstripe-framework/pull/5444

PR #5402, opened 2016-04-26
Media query to use mixin

clarkepaul:
Swapped out media query for mixin so values are using variables.


sminnee:
👍  from me

chillu:
Note that this is a bootstrap-defined mixin

Issue #5374, opened 2016-04-22
CampaignAdmin only shows preview (no item list) in ~800px window width


chillu:
It makes the whole section unuseable in small resolutions. Should be the other way around: Hide preview, only show list.

/cc @tractorcow @clarkepaul

clarkepaul:
I'm working on a fix for this.

clarkepaul:
Fixed and merged by https://github.com/silverstripe/silverstripe-framework/pull/5405

PR #5405, opened 2016-04-27
Add back button for smaller screens and fix for #5374

clarkepaul:
Added back button for when you can only view campaign items or item preview. This is just the HTML/CSS and doesn't include the JS need to make it work.

Fixes: https://github.com/silverstripe/silverstripe-framework/issues/5374


clarkepaul:
pasted_image_27_04_16__2_11_pm

clarkepaul:
Can you also close https://github.com/silverstripe/silverstripe-framework/issues/5374

PR #5408, opened 2016-04-27
API Implement basic preview behaviour

tractorcow:
Includes a few weird bugfixes:
- Issue with ViewableData::__toString not returning a string (crashes monolog during debugging)
- Cleanup of CMSPreviewable interface so that it required the correct methods, and added this interface to File.

Known issues:
- ~~When you select a campaign item, if you click on the border area it doesn't quite activate that item. Probably due to the parent of the react component having a stupid margin that's then negated by the child, somehow.~~ Fixed below
- Edit button is still not working, but the model layer has been implemented. We just need to update the react components to work with it.

Future improvements:
- Backend has implemented multiple preview actions; We will eventually have live / draft selector when viewing preview.

Also merge with:
- https://github.com/silverstripe/silverstripe-cms/pull/1464
- https://github.com/silverstripe/silverstripe-asset-admin/pull/123


tractorcow:
I also made https://github.com/silverstripe/silverstripe-asset-admin/pull/123 to update the file model view, but it's not critical to this PR.

tractorcow:
Screenshot:

image

tractorcow:
Screenshot of the unwanted border.

image

It'd be better to have the border on the <div> than on the <a> tag.

@clarkepaul can you please have a look at this?

clarkepaul:
Yup I can take a look at this one sometime today.

tractorcow:
I've fixed the click behaviour issue.

clarkepaul:
I've taken a look visually and things are all good, tested in IE10. I've just added a small fix for items which are active and focused as the colours and blurriness clash.

admin/client/src/components/Preview/Preview.js
```diff
@@ -7,25 +7,39 @@ import SilverStripeComponent from 'lib/SilverStripeComponent';
...
-        <iframe className="preview__iframe" src={this.props.previewUrl}></iframe>
+    // @todo - Multiple preview views with toggle slider
+    let body = null;
+    if (!this.props.previewUrl) {
+      body = (
         <div className="preview__overlay">
           <h3 className="preview__overlay-text">There is no preview available for this item.</h3>
         </div>
+      );
+    } else if (this.props.previewUrl.match(/\.((gif)|(jp(e?)g)|(png))($|\?)/i)) {
```
hafriedlander: Little bit ick. We can't pass through a preview Type arg or something? Or have a special handler?

filesystem/File.php
```diff
@@ -1162,4 +1162,16 @@ public function revokeFile() {
...
 		return $this->File->canViewFile();
 	}
+
+	public function CMSEditLink() {
+		$link = null;
+		$this->extend('updateCMSEditLink', $link);
+		return $link;
+	}
+
+	public function PreviewLink($action = null) {
```
hafriedlander: Should be on Image, not File

model/versioning/ChangeSetItem.php
```diff
@@ -345,4 +345,41 @@ public static function get_for_object_by_id($objectID, $objectClass) {
...
+
+	/**
+	 * Get edit link for this item
+	 *
+	 * @return string
+	 */
+	public function CMSEditLink()
+	{
+		$link = $this->getObjectInStage(Versioned::DRAFT);
+		if($link instanceof CMSPreviewable && $link->canEdit()) {
```
hafriedlander: Should this check canEdit, or just stick the link in anyway, and then trigger the "you can't access this item, please log in with a password that can" dialog? Actually, canView might be better anyway (you'd get a read-only form).

tractorcow: Probably the same as Link()... so just return it, and make sure you check visibility before informing the user it's editable.

admin/client/src/components/Preview/Preview.js
tractorcow: I've modified this to pass the mime type as an additional argument. Apparently HAL supports an explicit 'type' property for links, so it's standard. :D

https://tools.ietf.org/html/draft-kelly-json-hal-07#section-5.3

filesystem/File.php
tractorcow: File preview is now blank by default.

model/versioning/ChangeSetItem.php
tractorcow: Now actually removed the check... :D

Issue #5410, opened 2016-04-28
BUG: Error messages aren't properly focused (regression caused by #2397)


patricknelson:
This ticket  #2397 appears to have created a bug whereby editing `DataObjects` via `GridField` using the `GridFieldConfig_RecordEditor` (as an example) will not show that there were any validation errors, if the error occurs on a tab which you're not currently looking at. When I upgraded from SS 3.1 to SS 3.2 I noticed that the error notices no longer display in the top right when there's a validation problem. 

However, this code was written in such a way that it completely removes those notices but doesn't properly find the tab for errors which occur when editing a data object via `GridField`. I think because the particular offending line of code is here: https://github.com/silverstripe/silverstripe-framework/pull/2397/files#diff-5eadc0fb6b04cd16964de972bcf238efR92

In my case, `tabError` doesn't have a closest `.ss-tabset` since a completely different kind of tabs are being used, apparently (don't know the specifics of how it works yet). I'd suggest performing a check **first** to ensure that `tabError.closest('.ss-tabset')` doesn't contain an empty set and, if it does, revert back to old error format (as a stopgap). Or, also attempt to detect the alternative tab format and select that tab. Preferably both solutions would be suitable since if all else fails, the user isn't left thinking "Wait... why didn't I get a confirmation?", or worse, end up losing data thinking that they didn't get an error that it must have worked fine.

**Steps to reproduce:**

1. Setup a `GridField` with editable records via `GridFieldConfig_RecordEditor`
1. Make your `DataObject` have fields on multiple tabs.
1. Ensure you have validation setup for a field on the second tab.
1. Fill out the form ensuring the second tab will be the only one that will trigger an error
1. Switch **back** to the **first** tab and click "Save"

patricknelson:
Debugging this was complete hell due to the cryptic nature of the spaghetti nest of callbacks. Anyway, I think I found a solution to this. If the first attempt to find the nearest `.ss-tabset` fails, try again and look higher by looking for the nearest `.ss-cmstabset`. The only possible issue I can see with this is, due to the lack of a proper SilverStripe-driven API, what if you have multiple nested tabs? You end up with even more of a mess now since then you have to select each tab going up. 

I'll submit a PR soon. It's not a recursive solution, but maybe that's what needs to be done here? Something recursive that goes up the chain and selects each tab. Only problem with **that** then becomes what if there's extra unintended functionality associated with selecting that tab? etc etc ad nauseum! 


[...]

PR #2397, opened 2013-09-10
BUG: Reinstate tab and form focus states (fixes CMS #732 and #817)

adrexia:
* Added methods to track and restore field focus
* Fixed restoreTabState function to track ss-tabset and added a trigger event when tabstate reset

PR #5412, opened 2016-04-28
FIX for #5410 to help focus errors occuring on tabs within GridField controlled DataObjects (et al).

patricknelson:
Initial proof of concept to demonstrate a solution for #5410 for a regression that I believe was caused by PR #2397. By no means final in it's current state.

Some sort of recursive solution may be required, as mentioned in aforementioned issue. Also, using `setTimeout()` like this is obviously not a good idea. Anyone have a better solution? Maybe @wilr or @chillu? Since I know you both chimed in on PR #2397.

### DIFF
--- a/admin/javascript/LeftAndMain.EditForm.js
+++ b/admin/javascript/LeftAndMain.EditForm.js
@@ -46,6 +46,14 @@
 			ChangeTrackerOptions: {
 				ignoreFieldSelector: '.no-change-track, .ss-upload :input, .cms-navigator :input'
 			},
+
+			/**
+			 * Variable: ValidationErrorShown
+			 * Boolean for tracking whether a validation error has been already been shown. Used because tabs can
+			 * sometimes be inadvertently initialised multiple times, but we don't want duplicate messages
+			 * (Boolean)
+			 */
+			ValidationErrorShown: false,
 		
 			/**
 			 * Constructor: onmatch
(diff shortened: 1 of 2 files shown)


patricknelson:
Ok, I tweaked this to ensure that it doesn't accidentally navigate CMS tab links, so there's some better error checking going on there. It's still not recursive however, but it at least fixes my bug so far.

WIP.

tractorcow:
Hope you don't take it the wrong way... not sure about some of those comments in the code. :) Can you improve them please?

I'll treat this as WIP... so... ignoring for now. Let me know when it's working.

patricknelson:
Hah sure that was my intent eventually when (hopefully not "if") I circle back around to this. Basically need a recursive method to find the correct tab. 

By the way my original question: Do you have a better volition there than to use `setTimeout()` per chance?

tractorcow:
I can't think of a better solution... a hack is only bad if there's no elegant way of doing it, and that I can't see I'm afraid.

dhensby:
Seems like a legit approach to me...

patricknelson:
Ok, thinking "out loud" here. I hate using `setTimeout` like this so.. so much. It's like playing some [silly hand stacking game](https://www.youtube.com/watch?v=k6V-X41fjBM) but instead of hands, it's the [JavaScript stack](https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop#Adding_messages). Yuck.

Maybe we could setup an `entwine` `onmatch` listener to then trigger asynchronous events in order to register the toggle functionality at the tab level in a sort of detached way. In this way, we don't play this game of hoping our timeout is after any other timeouts (in case it matters) and simply notify any interested observers about the fact that a tab set has rendered. Granted, this does shift the issue a bit by still having a bit of a tight binding between here and tab sets, however, that already exists to some degree when referencing it by class name. Unfortunately this would require a fairly robust architecture to implement The Right Way™ since this would need a central method/dispatcher to handle my request to trigger my event, which itself is also responsible for handling getting those listeners. I'm positive I'm reinventing some wheel somewhere (e.g. react, promises or some other design pattern, or even entwine itself, but sort of conjoined/squared). 

Anyway, that way it doesn't matter of I call my trigger BEFORE or AFTER the initialization of the tabs. If it's before, the listener just sits and listens, waiting. If it's after tabs have initialized, the central dispatcher will just execute immediately. I'll have to think about it a bit more because this reeks.

patricknelson:
Hmm, what if we extended Entwine to allow for the calling of `.onmatch()` in a "sticky" fashion which then is able to be purged on a global scale (or per element)? For example:

- This code needs to select existing (or future) elements, only initialized with an `onmatch`, with a selector of `.cms-tabset`.
- The code then calls `$('.cms-tabset').onmatch(function() { ... })`.
- That executes immediately on any elements that have already executed their existing/empty `onmatch` method in Entwine.
- Any future elements that get matched will execute (just once) as well.
- User clicks to go to another page, page doesn't *really* refresh (and thus not really clearing out the heap) so, instead, we'll expose another utility method on Entwine which enables the ability to purge those "sticky" `onmatch` events.

Make sense?

kinglozzer:
@patricknelson I’ve removed the use of `setTimeout()` in https://github.com/silverstripe/silverstripe-framework/pull/5534, could you review that and see what you think? :)

patricknelson:
Ok @kinglozzer I was able to take a moment to try this out on my own, both in my own code as well as in a separate bare bones instance of SS 3.3. I think that's a much better solution (short to long term) over using `setTimeout()`. This took me a while to figure out how it works because I saw some (apparently undocumented?) functionality there on line 100 https://github.com/silverstripe/silverstripe-framework/pull/5534/files#diff-5eadc0fb6b04cd16964de972bcf238efR100

Anyway: I can see how this would work better as well, in terms of not mattering when each one's `onadd`/`onmatch` would execute, since the event listener `onafterredrawtabs` here is more declarative in fashion and would be defined very early on and wouldn't have to wait for the current element's `onadd` to fire first before being able to then attach it's listeners. So, presumably, your special syntax is able to define external listeners (from the context of the current form, for example) at a lower level in Entwine before Entwine then moves on to start hashing through it's `onadd` and `onmatch` (etc).

Do you want me to cherry pick that commit, squash into my PR, or will you squash/rebase your PR and then merge it?

patricknelson:
Actually @kinglozzer I've rolled your changes into mine and rebased against `3.3`.  I figured keeping it here would be faster/cleaner anyway. Thank you for your input!

kinglozzer:
> Do you want me to cherry pick that commit, squash into my PR, or will you squash/rebase your PR and then merge it?

Typically I’d say either cherry-pick, or someone (not me, as I authored the second commit) would merge the other PR if they’re happy with it - that way you get the appropriate author in commit history/blame view. As this change is a fairly simple one, and both commits are kinda doing the same thing, we may as well squash into a single commit, so I’ve closed the other PR.

I’ll give someone else a chance to review/merge before I do as I now part-authored this :)

patricknelson:
Sorry about that, good point @kinglozzer. At least the squash helped a bit due to my comments being "in need of improvement" as mentioned by @tractorcow above :joy: I agree on the history though. Given the speed of my SS-related contributions, at least this has been quite fast! Plus SS folks do typically prefer one commit + rebase, so that's at least my thought process originally.

Thanks and cheers to the UK and NZ folks :beers:

tractorcow:
@patricknelson I've manually merged this into master with https://github.com/silverstripe/silverstripe-framework/commit/66ef857cdfaf5e0cc99255a696f0236818a6065f

patricknelson:
2.5 years later and I'm noticing a few problems with this now:

- It doesn't work with nested tabs (my current issue in SS 3.x) which happens to be mentioned in [this comment](https://github.com/silverstripe/silverstripe-framework/issues/5410#issuecomment-215303233) in the original issue 😂 
- It is tightly bound to `.tab` class (which is no longer present in SS 4.x) so `.ui-tabs-panel` should be used instead (since we're still using jQuery UI's Tabs widget)
- Will require a PR to this repo + `silverstripe-admin` where this legacy code has been rolled in and still used for tabs

PR #5415, opened 2016-04-28
Hide preview options button for Alpha1, added BS JS for popovers

clarkepaul:
Commented out the more options button on the preview, added the required JS so popovers and tooltips work in CMS. 

No visual changes except the more options button is gone for Alpha 1.


admin/client/src/boot/index.js
```diff
@@ -37,8 +41,24 @@ function appBoot() {
 
   // Set the initial config state.
   window.store.dispatch(configActions.setConfig(window.ss.config));
+
+
+  // Enable BS Popovers and tooltips
+  $(function () {
```
chillu: I think this should be placed in the respective ReactJS component containing the tooltip logic - not every component will be rendered at the time `boot.js` is run, just happens to be the case here.

admin/client/src/components/Preview/Preview.js
```diff
@@ -20,7 +20,11 @@ class Preview extends SilverStripeComponent {
         <div className="toolbar--south">
           <div className="btn-toolbar">
             <button className="btn btn-secondary btn-secondary" type="button"><i className="font-icon-edit"></i> Edit</button>
-            <button type="button" data-container="body" className="btn btn-link btn--options" data-toggle="popover" title="Page actions" data-placement="top" data-content="<a href=''>Add to campaign</a><a href=''>Remove from campaign</a>"><i className="font-icon-dot-3"></i></button>
+            {/* More options button and popover for Alpha 2
```
tractorcow: I'd prefer to remove this not comment it out, please. :D

admin/client/src/boot/index.js
tractorcow: Also we shouldn't put jquery in react code, right?

tractorcow: If it's just jquery can it go into LeftAndMain.js?

admin/client/src/components/Preview/Preview.js
chillu: We need to ensure the work that has gone into creating this markup (incl. details like setting a tabindex) doesn't get lost. I'm fine with commented out code as long as there's a clear path to enabling it again, which there is in this case since we'll need tie feature soon. If removing it, please create a Github issue which contains the markup as a starting point.

admin/client/src/boot/index.js
clarkepaul: I've moved this stuff to a separate component for Popover and Tooltip.

admin/client/src/components/Preview/Preview.js
clarkepaul: I've deleted the commented out code from this template, but I've kept the HTML as an example in the Popover/Tooltip component which isn't being used anywhere at the moment. 

The Popover/Tooltip component isn't complete but as it isn't used at the moment it hopefully wont hold back this getting merged.

admin/client/src/components/PopoverTooltip/PopoverTooltip.js
```diff
@@ -0,0 +1,46 @@
...
+class PopoverTooltip extends React.Component {
+  componentDidMount() {
+    const el = ReactDOM.findDOMNode(this);
+
+    // Enable BS Popovers and tooltips
+    $(el).find('[data-toggle="tooltip"]').tooltip();
+    $(el).find('[data-toggle="popover"]').popover({ html: true });
+
+    // Accessibility support for popover focus (requires anchor element)
+    $(el).find('.popover-dismiss').popover({ trigger: 'focus' });
```
chillu: These three jQuery selectors don't actually match anything, since they're looking for children rather than on the element itself.

admin/client/src/components/PopoverTooltip/PopoverTooltip.scss
```diff
@@ -0,0 +1,39 @@
...
+		margin-right: -$spacer-x;
+		margin-bottom: 0
+	}
+
+	a {
+		display: block;
+		padding: $spacer-y/4 $spacer-x;
+		color: $dropdown-link-color;
+
+		&:hover {
```
chillu: Has an underline for the three buttons when using the sample markup from the README in `Preview.js`

image

admin/client/src/components/PopoverTooltip/PopoverTooltip.js
```diff
@@ -0,0 +1,46 @@
...
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+class PopoverTooltip extends React.Component {
+  componentDidMount() {
+    const el = ReactDOM.findDOMNode(this);
+
+    // Enable BS Popovers and tooltips
+    $(el).find('[data-toggle="tooltip"]').tooltip();
+    $(el).find('[data-toggle="popover"]').popover({ html: true });
```
chillu: I don't get it, how can this be both a `tooltip()` and a `popover()`? You can either have a `data-toggle="popover"` or a `data-toggle="tooltip"` attribute on the element. How do you see nested elements being both tooltips and popovers? Looking at the docs, we should be fine with just a popover? (http://v4-alpha.getbootstrap.com/components/tooltips/, http://v4-alpha.getbootstrap.com/components/popovers/).

admin/client/src/components/PopoverTooltip/PopoverTooltip.js
```diff
@@ -0,0 +1,46 @@
...
+
+    // Accessibility support for popover focus (requires anchor element)
+    $(el).find('.popover-dismiss').popover({ trigger: 'focus' });
+  }
+
+  // TODO allow for custom popover/tooltip trigger
+  // eg. (classes, contents, data-toggle, placement, icon and text)
+  // Add the possibility to use FormAction component as a trigger
```
chillu: Rather than creating these temporary components with lots to TODOs, I think we should start using react-bootstrap which already has those TODOs resolved ;) https://react-bootstrap.github.io/components.html#popovers
Maybe that's out of scope for the current sprint, but there's not much point going beyond styling then

admin/client/src/components/PopoverTooltip/PopoverTooltip.scss
clarkepaul: I broke it trying to pull the code out into its own component, but seeing as it isn't currently used and not in our current work load I think it is ok to leave as is a pickup when it is relevant (I've created a story to capture the work). Probably an easy one for a dev to figure out but doing React is a bit beyond my capability.

admin/client/src/components/PopoverTooltip/PopoverTooltip.js
clarkepaul: Yeah I agree, I already mention earlier that a story has been created to fix this but it isn't really part of our current work, so I was just trying to clean up the code that that is relevant. I don't think I'm the right person to be doing the React stuff though.

clarkepaul:
Seeing as this is getting further and further away from the sprint goals I will submit another pull request which removes the More Options button, and leave this stuff for Alpha2

tractorcow:
Don't worry about preview buttons; I've removed them at https://github.com/silverstripe/silverstripe-framework/pull/5432

Please merge that one first, I think?

tractorcow:
Closing on @clarkepaul 's request; The buttons have already been updated on master.

tractorcow:
@clarkepaul still need this branch? Remove it if not.

clarkepaul:
I'd like to keep it in the case I need to implement the popups.

PR #5416, opened 2016-04-28
Fix most important documentation error

NightJar:
Lest a riot be started.


chillu:
lol classic. I think that was me getting the NZ sports wrong (I still have no clue, even after 9 years in NZ)

PR #5423, opened 2016-05-02
API LeftAndMain::menu_title can be overridden

tractorcow:
Solves https://github.com/silverstripe/silverstripe-asset-admin/pull/119 without forcing master transifex strings to need to be changed.

Actual fix is at https://github.com/silverstripe/silverstripe-cms/pull/1467, which relies on this framework API being merged first.

### DIFF
--- a/admin/code/LeftAndMain.php
+++ b/admin/code/LeftAndMain.php
@@ -683,14 +683,40 @@ class LeftAndMain extends Controller implements PermissionProvider {
 	}
 
 	/**
-	 * Returns the menu title for the given LeftAndMain subclass.
-	 * Implemented static so that we can get this value without instantiating an object.
-	 * Menu title is *not* internationalised.
+	 * @deprecated 5.0
 	 */
 	public static function menu_title_for_class($class) {
+		Deprecation::notice('5.0', 'Use menu_title() instead');
+		return static::menu_title($class, false);
+	}
+
+	/**
+	 * Get menu title for this section (translated)
+	 *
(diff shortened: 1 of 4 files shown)


admin/code/LeftAndMain.php
```diff
@@ -683,14 +683,40 @@ public function Link($action = null) {
...
+	 * @param bool $localise Determine if menu title should be localised via i18n.
+	 * @return string Menu title for the given class
+	 */
+	public static function menu_title($class = null, $localise = true) {
+		if($class && is_subclass_of($class, __CLASS__)) {
+			// Respect oveloading of menu_title() in subclasses
+			return $class::menu_title(null, $localise);
+		}
+		if(!$class) {
+			$class = get_called_class();
```
chillu: The PHP docs for [get_called_class()](http://php.net/get_called_class) state:

> Gets the name of the class the static method is called in.

My tests on PHP 5.4 (min version for SS4) show that it indeed works with instance methods as well. I still think its a weird mix of using a static method as an instance method (`$this->menu_title()`), but it does the job :)

```
php > class foo { static function bar() {var_dump(get_called_class());}}
php > class subfoo extends foo {};
php > foo::bar();
string(3) "foo"
php > subfoo::bar();
string(6) "subfoo"
php > (new foo())->bar();
string(3) "foo"
php > (new subfoo())->bar();
string(6) "subfoo"
```

tractorcow: > I still think its a weird mix of using a static method as an instance method

It is weird, but it is a valid use case... it's kind of like negative margins. :D

PR #5428, opened 2016-05-02
Added NPM tests

chillu:
Turns out we only ran `npm run sanity`, not `npm run test` - doh.
I'm not running `composer install` for NPM, speeds up that particular build a bit.

Aside: The bash conditionals get a bit messy, but it's still clearer than a separate script IMHO since you can see the output for each line in `.travis.yml` individually. Travis will only output the line which failed. If you route all test setup and invocation through the same script, it'll show all output in case of a test failure, including `composer install` and other gunk. The best solution here would be to separate out the whole JS codebase into a separate repo, publish it on NPM, include it via `package.json` in core, and run tests on a much simpler `.travis.yml` there. But that's more of a long term decision than fixing short term complexity in the `.travis.yml` file.


sminnee:
Minor point but I'd probably refactor this so we have `RUN_PHPUNIT`, `RUN_BEHAT`, and , `RUN_NPMTEST`, and put the `RUN_PHPUNIT` env var on all the "regular" builds. It will simplify the conditionals because you won't need to check for `""` so much.

sminnee:
Other point, if the CMS tests were handled with an `EXTRA_COMPOSER_MODULE=,silverstripe/cms` setting, you wouldn't need a double up of the lines.

sminnee:
Of course, `replace(/RUN_(.*)/, '$1_TEST')` is fine too, if you wanted to keep the variables closer to the current ones.

chillu:
> Minor point but I'd probably refactor this so we have RUN_PHPUNIT, RUN_BEHAT, and , RUN_NPMTEST, and put the RUN_PHPUNIT env var on all the "regular" builds.

Good point, I've introduced `PHPUNIT_TEST` for that purpose. Haven't renamed to `RUN_PHPUNIT` since that'll introduce more obvious naming inconsistencies between all Travis runs between different modules. `PHPUNIT_TEST` is an addition to existing naming, `RUN_*` is a change of all naming - which I don't have time to see through at the moment :)

> Other point, if the CMS tests were handled with an EXTRA_COMPOSER_MODULE=,silverstripe/cms setting, you wouldn't need a double up of the lines.

Only for the `before_script` step, youd' still need a specific flag for `vendor/bin/behat @framework` vs. `vendor/bin/behat @cms` and `vendor/bin/phpunit framework/tests` vs. `vendor/bin/phpunit cms/tests`

tractorcow:
I don't see any output on 'npm run test', but it's not failing either.

tractorcow:
Missing $ on `NPM_TEST`

chillu:
@tractorcow It's all green! And even with running tests, rather than just skipping builds because of a missing `$` ;)

PR #3592, opened 2014-10-30
NEW: check changed state text fields on keyup, rather than just change

unclecheese:
This issue relates to the state of the save/publish buttons in CMSMain. They only update on blur of the text fields, creating an awkward UX, since many users will click one of those actions whilst the text fields are focused.

Minor: update changetracker to use on() and off() instead of bind() and unbind().

PR #5431, opened 2016-05-02
NEW: check changed state text fields on keyup, rather than just change

unclecheese:
This is a continuation of an old PR https://github.com/silverstripe/silverstripe-framework/pull/3592 but against the newer branch.

### DIFF
--- a/admin/javascript/jquery-changetracker/lib/jquery.changetracker.js
+++ b/admin/javascript/jquery-changetracker/lib/jquery.changetracker.js
@@ -46,7 +46,8 @@
 		this.defaults = {
 			fieldSelector: ':input:not(:submit)',
 			ignoreFieldSelector: "",
-			changedCssClass: 'changed'
+			changedCssClass: 'changed',
+			customChangeEvents: {}
 		};
 
 		var options = $.extend({}, this.defaults, _options);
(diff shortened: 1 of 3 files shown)


tractorcow:
I think this fix will add a lot of value. I'm keen on testing it out, but it looks like tests are broken. Do you want to have a go at fixing them?

admin/javascript/jquery-changetracker/lib/jquery.changetracker.js
```diff
@@ -88,9 +89,31 @@
...
+
+			fields.filter(omittedSelectors).on('click.changetracker', onchange);
+			fields.not(omittedSelectors).on('change.changetracker', onchange);
+			
+			// Handle omitted selectors
 			fields.each(function() {
+				// For selectors with custom change events
+				for(var selector in options.customChangeEvents) {
+					var eventName;
+					if(options.customChangeEvents.hasOwnProperty(selector)) {
```
tractorcow: How could this be false? Isn't it a property by virtue of the loop on options.customChangeEvents?

tractorcow:
I'm not smart enough to see any issues in the code itself. :) Good job @unclecheese .

unclecheese:
This whole PR is bunk now, as all of this has moved to `silverstripe-admin`. Will start a new one and close this one off.

PR #5432, opened 2016-05-02
API Implement campaign item edit button

tractorcow:
Makes the "edit" button on campaign preview point to the actual edit url for each item.

Removes the non-functional "extra options" menu for the time being.

See https://silverstripe.atlassian.net/browse/OSS-1687


tractorcow:
Yet again rebased.

admin/client/src/components/Preview/Preview.js
```diff
@@ -9,29 +10,57 @@ class Preview extends SilverStripeComponent {
...
+
+    // Build body
+    if (!previewUrl) {
       body = (
         <div className="preview__overlay">
           <h3 className="preview__overlay-text">There is no preview available for this item.</h3>
         </div>
       );
-    } else if (this.props.previewType.indexOf('image/') === 0) {
+    } else if (previewType.indexOf('image/') === 0) {
```
hafriedlander: previewType could be null here

PR #5433, opened 2016-05-03
Introduce SilverStripe\Control and SilverStripe\Model namespaces

sminnee:
As well as putting the model classes into the `SilverStripe\Model` namespace, this PR highlight the operation of the http://github.com/sminnee/silverstripe-upgrader tool.


sminnee:
Note: before merging this I would like to get a view on whether we should make silverstripe-upgrader a recommended part of the SS3 -> SS4 upgrade path, as that will probably lead to another commit with some silverstripe-upgader meta-data.

sminnee:
In the interest of sanity, we may also want to leave this until after alpha1.

tests/i18n/i18nTextCollectorTestMyObject.php
```diff
@@ -12,7 +12,7 @@ class i18nTextCollectorTestMyObject extends DataObject implements TestOnly {
 	);
 
 	private static $has_many = array(
-		'SilverStripe\Model\Relation' => 'Group'
```
sminnee: Over-eager string replacement. See https://github.com/sminnee/silverstripe-upgrader/issues/1

tests/model/DataQueryTest.php
```diff
@@ -304,7 +304,7 @@ class DataQueryTest_C extends DataObject implements TestOnly {
 class DataQueryTest_D extends DataObject implements TestOnly {
 
 	private static $has_one = array(
-		'SilverStripe\Model\Relation' => 'DataQueryTest_B',
+		'Relation' => 'DataQueryTest_B',
```
sminnee: Over-eager string replacement. See https://github.com/sminnee/silverstripe-upgrader/issues/1

tests/core/ClassInfoTest.php
```diff
@@ -280,6 +280,6 @@ class ClassInfoTest_HasFields extends ClassInfoTest_NoFields {
 class ClassInfoTest_WithRelation extends ClassInfoTest_NoFields {
 
 	private static $has_one = array(
-		'SilverStripe\Model\Relation' => 'ClassInfoTest_HasFields'
```
sminnee: Over-eager string replacement. See https://github.com/sminnee/silverstripe-upgrader/issues/1

tractorcow:
Pull request relevant to this change... https://github.com/silverstripe/silverstripe-framework/pull/5334

We need to look at another solution for "inheritable" configs that don't require us to double-define configs for subclasses and namespaced/unnamespaced versions of each service.

sminnee:
> We need to look at another solution for "inheritable" configs that don't require us to double-define configs for subclasses and namespaced/unnamespaced versions of each service.

I don't believe we need double-defined configs. It will cause more harm than good. It will lead to dumb things like:

 * What do we do if `SilverStripe\Model\DataObject` is defined in core and `DataObject` is defined in project code?
 * What if they're both defined with different values?

My view is that the default services (i.e. the one referenced by `Object::create()`) should match the fully-qualified classname. We may *possibly* accept the unqualified classname as a deprecated syntax, but it would be best to provide a migration tool for this out-of-date syntax.

micmania1:
Another thing to keep in mind here is template names which are mapped to the class name.

eg. 
`Page` -> `templates/Page.ss` now becomes `SilverStripe\Model\Page` -> `templates/SilverStripe\Model\Page.ss`

There could be a fallback which uses the base class name until there's a drive (or need) to expand on this a bit more?

sminnee:
I've added namespaces to the control section. Right now there are some failures - for some reason the CMS code is getting mixed in there. :-/

sminnee:
@micmania1 I haven't added classes to any of the controller subclasses themselves, only controller. So I've had to create `framework/templates/SilverStripe\Control\Controller.ss`, which is a bit perverse, and I guess will break on Windows. Perhaps this is a merge-blocker?

micmania1:
Yeah, I would say so. You can have a Page class/subclass without a controller too (as of 3.2). The logic looks for templates based on the class name of the page, rather than the controller.

This extends to anything that extends ViewableData including DataObjects and form fields.

sminnee:
Have raise #5445, the first part of which probably blocks merge for this.

In short, I think the starting point is to look for the template at `framework/templates/SilverStripe/Control/Controller.ss` - i.e. in a subdirectory. A fancier PSR-4 copycat can come later.

sminnee:
OK the core build is green, but I'll need to upgrade the behat-extension, cms, postgresql, and sqlite builds to get a full passing build. I'll raise the PRs.

But, first, I want to fix the template naming thing.

tractorcow:
> I don't believe we need double-defined configs. It will cause more harm than good.

Then can you remove the duplicate define for the below and just have one.

```yaml
  MySQLDatabase:
    class: 'SilverStripe\Model\Connect\MySQLDatabase'
    properties:
      connector: %$MySQLiConnector
      schemaManager: %$MySQLSchemaManager
      queryBuilder: %$MySQLQueryBuilder
  'SilverStripe\Model\Connect\MySQLDatabase':
    class: 'SilverStripe\Model\Connect\MySQLDatabase'
    properties:
      connector: %$MySQLiConnector
      schemaManager: %$MySQLSchemaManager
      queryBuilder: %$MySQLQueryBuilder
```

tractorcow:
Now that https://github.com/silverstripe/silverstripe-framework/pull/5440 is merged you could just do.


```yaml
MySQLDatabase:
    class: 'SilverStripe\Model\Connect\MySQLDatabase'
    properties:
      connector: %$MySQLiConnector
      schemaManager: %$MySQLSchemaManager
      queryBuilder: %$MySQLQueryBuilder
  'SilverStripe\Model\Connect\MySQLDatabase': '%$MySQLDatabase'
```

sminnee:
I think that this is obsolete now; thanks @tractorcow and @hafriedlander for carrying the torch!

PR #5436, opened 2016-05-03
Initialise React controllers via routes

flashbackzoo:
Removes the dependency on Entwine in React controllers. React controllers (Campaign section) register routes with the new RouteRegister. These routes are applied to Page.js in the `appBoot` boot process.

Merge alongside https://github.com/silverstripe/silverstripe-asset-admin/pull/196

cc: @chillu


admin/client/src/lib/RouteRegister.js
```diff
@@ -2,10 +2,41 @@ import deepFreeze from 'deep-freeze';
 
 /**
  * The register of routes
+ *
  * @private
  */
 let register = deepFreeze({});
 
+/**
+ * RouteRegister is the iterface developers should use to register routes with
```
chillu: Spelling

chillu:
`admin/pages` loads `listview` and `treeview` twice now :( 

image

chillu:
Loading `admin/assets` fails with this: 

image

flashbackzoo:
I'm not getting the `/admin/assets` issue. I'll try updating my AssetAdmin module, could be a change in that module that's causing the issue...

tractorcow:
All these behat tests are passing locally... I can't figure out why they are failing on travis.

flashbackzoo:
The `TypeError` in the behat tests is caused by a conflict between using JavaScript's 'strict mode' and an ES3 fallback in a thirdparty library. Our behat tests use an old version of Firefox which requires the ES3 fallback. But this fallback uses syntax which is not valid in strict mode.

admin/client/src/legacy/LeftAndMain.js
```diff
@@ -370,15 +370,15 @@ $.entwine('ss', function($) {
      * @param object data - Any additional data passed through to `window.history.state`.
      * @param boolean forceReload - Forces the replacement of the current history state, even if the URL is the same, i.e. allows reloading.
      */
-    loadPanel: function (url, title = '', data = {}, forceReload, forceReferer = window.history.state.path) {
+    loadPanel: function (url, title = '', data = {}, forceReload, historyState = window.history.state) {
```
tractorcow: This does make loadPanel a lot harder to use; Before you could pass in a string to the 4th parameter, but now it has to be a state. It seems to be a case of API that serves the api itself, rather than the human using it.

sminnee:
@chillu this is in danger of rotting. are you able to pick it up?

chillu:
I'll put it in the "next" column on Zenhub

chillu:
Behat cms build failure should be fixed through https://github.com/silverstripe/silverstripe-cms/pull/1524

tractorcow:
The functionality works, but there's a critical performance issue where routes are being duplicated several times.

For instance, if I had visited the campaign section 4 times, every page click within that section has routes duplicated 4x

image

In the old code I had an `exit` handler that would disable routes to prevent this. I guess that code (or a variant) needs to be put back.

admin/client/src/legacy/LeftAndMain.js
chillu: The weird thing is that I can't see any changed `loadPanel()` call in this diff, so I've reverted https://github.com/open-sausages/silverstripe-framework/commit/72c92515af69f30631db05d4bd397fdcf71d474d

tractorcow:
Removing deep-freeze broke asset-admin...

tractorcow:
I've fixed these with https://github.com/silverstripe/silverstripe-asset-admin/pull/196.

tractorcow:
Pushed up a fix for missing deep-freeze-strict in shrinkwrap.

tractorcow:
Tests passing. :)

tractorcow:
@chillu please confirm that my fixes to this PR, and to asset-admin are ok, and I'll merge.

admin/client/src/boot/index.js
```diff
@@ -77,22 +77,31 @@ function appBoot() {
...
-        $('.cms-container')
-          .entwine('ss')
-          .handleStateChange(null, ctx.state)
-          .done(next);
-      });
+  const sections = Config.get('sections');
+  Object.keys(sections).forEach((key) => {
+    const sectionConfig = sections[key];
+
+    // Skip react routes which are handled by individual route setup
```
tractorcow: What if a section had both react and non-react routes?

PR #5426, opened 2016-05-02
BUG Fix GridFieldAddExistingAutocompleter

tractorcow:
This fix exchanges references to sub-components by ID with class references
Fixes #5382

Probably best reviewed with '?w=1' to remove whitespaces from diff.

CC @HARVS1789UK  for review. Can you please confirm this fixes your issue?

PR #5437, opened 2016-05-03
BUG Fix GridFieldAddExistingAutocompleter and GridFieldExportButton

### LINKED ISSUE #5267: BUG Can't link existing member to group
GridFieldAddExistingAutocompleter seems to show "Object" instead of the actual item.

Currently affecting master branch.

image

Jira: https://silverstripe.atlassian.net/browse/OSS-1728

tractorcow:
Based on https://github.com/silverstripe/silverstripe-framework/pull/5426
Fixes #5267


hafriedlander:
I don't think it's necessary to replace ID usage with class usage unless we're also changing to BEM. However I'm not going to block merge because of it.

tractorcow:
Annoyingly, this broke master AFTER merge. Investigating.

dhensby:
How did the investigation go?

tractorcow:
We are leaving PHP7 as "allowed_failure" at the moment.

The fix is to properly implement "__isset" for viewable data. See https://github.com/silverstripe/silverstripe-framework/issues/5450

Issue #5424, opened 2016-05-02
Campaign buttons use wrong class, replace .btn-success with .btn-primary


clarkepaul:
After a refactor of the bootstrap button classes we should now use .btn-primary instead of .btn-success which is only being used in a couple places within campaigns. This issues needs https://github.com/silverstripe/silverstripe-framework/pull/5393 to be merged first.

PR #5439, opened 2016-05-03
Swapped a few btn classes for updated ones

clarkepaul:
Also added btn label which I think was missing "Edit" 

(fixes https://github.com/silverstripe/silverstripe-framework/issues/5424)


clarkepaul:
There shouldn't be any major difference in the buttons with the classes changed.

chillu:
Updated dist files and merged

PR #5440, opened 2016-05-03
API Injector dependencies no longer inherit from parent classes

### LINKED ISSUE #4774: Injector service definitions should not inherit to subclasses

tractorcow:
Now injector configs must be explicitly stated as inherting.

A partial resurrection of https://github.com/silverstripe/silverstripe-framework/pull/4580 but for a different reason.

Fixes #4774

### DIFF
--- a/control/injector/SilverStripeServiceConfigurationLocator.php
+++ b/control/injector/SilverStripeServiceConfigurationLocator.php
@@ -19,53 +19,36 @@ class SilverStripeServiceConfigurationLocator extends ServiceConfigurationLocato
 	protected $configs = array();
 
 	public function locateConfigFor($name) {
-
 		// Check direct or cached result
 		$config = $this->configFor($name);
-		if($config !== null) return $config;
-
-		// do parent lookup if it's a class
-		if (class_exists($name)) {
-			$parents = array_reverse(array_values(ClassInfo::ancestry($name)));
-			array_shift($parents);
-
-			foreach ($parents as $parent) {
-				// have we already got for this?
-				$config = $this->configFor($parent);
-				if($config !== null) {
-					// Cache this result
-					$this->configs[$name] = $config;
-					return $config;
-				}
-			}
+		if(!$config) {
+			return null;
 		}
 
-		// there is no parent config, so we'll record that as false so we don't do the expensive
-		// lookup through parents again
-		$this->configs[$name] = false;
+		// If config is in `%$Source` format then inherit from the named config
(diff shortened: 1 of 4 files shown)


tractorcow:
This should be alpha1, can be alpha2 :D

sminnee:
Looks good to me. Keen to get a 👍  from others, maybe @dhensby, @kinglozzer, @hafriedlander or @nyeholt?

control/injector/SilverStripeServiceConfigurationLocator.php
```diff
@@ -19,53 +19,36 @@ class SilverStripeServiceConfigurationLocator extends ServiceConfigurationLocato
...
-				// have we already got for this?
-				$config = $this->configFor($parent);
-				if($config !== null) {
-					// Cache this result
-					$this->configs[$name] = $config;
-					return $config;
-				}
-			}
+		// If config is in `%$Source` format then inherit from the named config
+		if(is_string($config) && stripos($config, '%$') === 0) {
```
kinglozzer: Noice!

kinglozzer:
This gets a 👍  from me. I’ll give it 24 hours in case anyone else has any comments then I’ll merge

kinglozzer:
Merging, nice work @tractorcow!

nyeholt:
Has been merged already, but yes, am happy with this approach :)

patricknelson:
Thanks for this on my end as well. The injector giving me a class I didn't ask for nor directly inject has haunted me for a little while so, for me from an internal/dev perspective, this is an improvement (albeit I know it's a little bit of a gray area).

PR #5443, opened 2016-05-03
Added CodeClimate integration

chillu:
Same as https://github.com/silverstripe/silverstripe-asset-admin/pull/126.
Sorry for the trailing whitespaces fixes, I've turned that off in my IDE now


chillu:
Merging in so I can validate Code Climate is working (we switched the default branch after this pull request was opened). The build fails because of unrelated PHP7 changes which are already addressed elsewhere.

PR #3183, opened 2014-06-02
Lazy Loaded Fields break when initial query runs in default stage

crgwbr:
This pull request fixes issue #3182 and adds a test case for it.

PR #5447, opened 2016-05-04
Fixes #3182 Fixes lazy loading of fields when query was created in default stage

### LINKED ISSUE #3182: Lazy Loaded Fields break when initial query runs in default stage
When a DataObject query is created in the default stage (stage.Stage), then the stage is changed to something else (stage.Live), the query itself runs correctly against stage.Stage. However, lazy loaded fields incorrectly run against stage.Live.  This is due to DataQuery parameters not being saved when the default stage is active.

tractorcow:
Replaces #3183
Fixes #3182

Rebase of https://github.com/silverstripe/silverstripe-framework/pull/3183 onto master.

There is no change to Versioned.php because it's already fixed in master; This rebase just includes the new test to prove the fix works. :)

### DIFF
--- a/tests/model/VersionedTest.php
+++ b/tests/model/VersionedTest.php
@@ -749,6 +749,27 @@ class VersionedTest extends SapphireTest {
 		Versioned::set_reading_mode($originalMode);
 	}
 
+	public function testLazyLoadFieldsRetrieval() {
+		// Set reading mode to Stage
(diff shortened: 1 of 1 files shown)


tractorcow:
PHP7 Failing is an issue with the recent 7.0.6 release. Fixed with https://github.com/silverstripe/silverstripe-framework/pull/5449

tractorcow:
Rebased on top of #5449

PR #5448, opened 2016-05-04
Resolve scss linting issues in new scss within framework

clarkepaul:
Includes new components, campaign admin and base styles like typography


admin/client/src/components/ListGroup/ListGroup.scss
```diff
@@ -56,27 +61,23 @@
...
   display: inline-block;
   transition: opacity .2s ease-in-out;
 }
 
-.list-group-item.active .list-group-item__status {
-	color: #FFF;
-	opacity: 1;
-}
-
-.list-group-item--inactive .list-group-item__status {
```
chillu: `.list-group-item--inactive` is still used in `CampaignAdminList.js` - if its no longer needed, please remove it consistently.

admin/client/src/components/GridField/GridField.scss
```diff
@@ -58,41 +63,43 @@
...
+      background-position-y: -118px;
+    }
+
+    &.ss-gridfield-sorted-desc,
+    &.ss-gridfield-sorted-desc:hover {
+      background-position-y: -78px;
+    }
   }
 
-  thead tr.title th {
```
chillu: I don't see `.title` applied anywhere on the `GridField` ReactJS component, so the styles which used to live on `thead` now won't apply?

chillu:
Merging this, assuming Paul might create some follow-up PRs for the two bits of feedback I had. These whitespace fixes have a habit of messing up our workflow (lots of waste in branch rebasing and merge conflicts), so I don't want to keep it open for too long.

admin/client/src/components/GridField/GridField.scss
clarkepaul: .title isn't used currently anywhere with the new gridfield but if you apply the new gridfield class to the older grid-field it will get the new styles. Kind of a work in progress.

admin/client/src/components/ListGroup/ListGroup.scss
chillu: @clarkepaul Any feedback on the above?

clarkepaul: Sorry I meant to take a look today, it could have been an oversite from me to remove. Will take a look first thing tomorrow.

clarkepaul: Ok I remember why I removed the css and not the html. This feature has just been taken out of Alpha1 so we want it to remain hidden hence we don't want the opacity: 1 at all. I think I'll add a comment to explain this for future reference.

PR #5449, opened 2016-05-04
Excuse php7 errors due to regressions in 7.0.6

tractorcow:
The issue in framework is that ViewableData does not report that getters are set via `__isset`. Up until php 7.0.6 only the last variable within an `isset()` expression would be checked via `__isset`. Now all values in an expression are checked from left to right.

i.e. for the below expression `isset($gridfield->State->Key)` previously php would perfrom `$gridfield->State->__isset('Key')`, but now does `$gridfield->__isset('State') && $gridfield->State->__isset('Key')`


tractorcow:
Raised issue at  https://github.com/silverstripe/silverstripe-framework/issues/5450

PR #5457, opened 2016-05-05
Moved AssetAdmin variables into framework for reuse

clarkepaul:
Also delete a ui-theme font family which was causing linting issues trying to override the style.

This needs merging before AssetAdmin linting fix.


admin/client/src/styles/_variables.scss
```diff
@@ -697,3 +699,36 @@ $cms-panel-xs: 60px;
...
+// Shadows, these need work to make consistent (used in AssetAdmin)
+$shadow-light: rgba(0, 0, 0, .2);
+$dropdown-shadow: 0 0 3px $shadow-light;
+$gallery-item-shadow: 0 1px 4px $gray-light;
+
+// Masks: typically behind modals, drag drop areas
+$page-mask-light: rgba(236, 239, 241, .9);
+
+
+// Variables for use within but not limited to the AssetAdmin module
```
hafriedlander: Doc out of date

admin/client/src/styles/_variables.scss
```diff
@@ -697,3 +699,36 @@ $cms-panel-xs: 60px;
...
+// Gallery item path to framework
+
+$framework-path:     "../../../../framework";
```
hafriedlander: Seems needlessly complicated given we're in framework now :). `../../..`?

clarkepaul: yup I saw that and removed that line, it was originally a straight copy from assetadmin but I've moved that line back.

clarkepaul: Updated already

admin/client/src/styles/_variables.scss
clarkepaul: Whats out of date?

hafriedlander: Why does it mention AssetAdmin module here? They're either appropriate for framework generally or not. I'd just say `// Gallery item sizes` or something.

clarkepaul: They are currently only used in AssetAdmin but they should be generic enough for use elsewhere. The note is there just incase they don't get used again in framework and someone want to delete them—its a reminder to keep them even though they might look like they aren't used.

I can remove comment if you think its not needed?

clarkepaul: FYI Improved comment

hafriedlander:
Looks good to me. Will merge when green.

Issue #5283, opened 2016-04-07
React-based form action loading indicator doesn't animate in IE


chillu:
See `svg` rule in `admin/javascript/src/components/form-action/styles.scss`. Short story: SVG animation in IE sucks.

Will be merged in shortly to `master` via https://github.com/silverstripe/silverstripe-framework/pull/5278/

/cc @scott1702 @flashbackzoo

tractorcow:
Is this fixed with #5278 now?

chillu:
Sorry, my description was a bit misleading there: This problem is *caused* by new code I've merged in with #5278, rather than fixed through it. So it's not fixed yet. @scott1702 any chance you could have a look?

scott1702:
Yep, it's at the top of my open-source todo list.

clarkepaul:
Possibly replace with #7 http://tobiasahlin.com/spinkit/ but test in IE10 first.

PR #5460, opened 2016-05-05
Fix loading icon bug on IE

scott1702:
Unfortunately IE can't handle css animating SVGs, this just turns the <svg> into <span>s so we can animate them in IE.

Fixes https://github.com/silverstripe/silverstripe-framework/issues/5283


chillu:
@scott1702 This hasn't been tested with other button style variations I think, the dots are grey regardless of the button text colour (cc @clarkepaul)

image

scott1702:
Should be good now @chillu

chillu:
Broken up in two lines for IE10 on "Edit Campaign": http://take.ms/8rzqa

scott1702:
Talked this over with @clarkepaul around how we could handle this, came to the decision of using the same indicator with smaller circles. This change has been added and tested with buttons as small as 'icon-only'.

Third time lucky @chillu? 😉

clarkepaul:
Looked good to me and tested in IE10.

admin/client/src/components/FormAction/FormAction.scss
```diff
@@ -143,8 +139,26 @@
...
 
-  svg circle {
-    fill: $body-color;
+  .btn__circle {
+    background: $btn-secondary-color;
+  }
+}
+
+.btn-danger {
+  .btn__circle {
```
dhensby: you're not meant to nest elements in BEM

dhensby:
needs rebasing :/

scott1702:
Rebased :)

dhensby:
Thanks

Issue #5347, opened 2016-04-19
RFC: Remove RestfulService in SS4


sminnee:
RestfulService was written at a time when the state of the art for connecting to an API was direct Curl calls. It was also written at a time when SilverStripe had a more "kitchen sink" mentality.

These days, a developer is likely to have a much better time using something like Guzzle.

Dan has proposed that we drop it, which I agree with.

Specifically:

- deprecate RestfulService in the 3 branch
- update Oembed not to rely on it (a simple direct use of curl is probably fine for now)
- delete RestfulService from the master branch
- close all PRs and issues related to RestfulService

An open question: should we provide something compatible with the RestfulService API that acts as a layer over guzzle, most likely in a module? It doesn't seem worth maintaining so I'm inclined to say "no", but it would make upgrading easier.

sminnee:
@dhensby, this one is for you

micmania1:
Set it on fire!

wilr:
:+1: remove. Implementing via Guzzle is easy enough. I would think it didn't need to be in core. If someone needs a SilverStripe wrapper around Guzzle then a module could be raised but let the need drive the demand for it.

dhensby:
I've thought about providing a polyfill to wrap around guzzle, but I don't think it's worth our time, RestfulService is just not good enough to make it worth the time - and i have other things I'd prefer to do instead :)

bummzack:
I preferred `RestfulService` over raw Curl calls (ease of use), and I have some projects where I made use of `RestfulService`. 

Calling some REST API from your SilverStripe code is quite common and I preferred `RestfulService`, because it didn't add additional dependencies and pretty much did what I wanted in a one-liner.

Using something like guzzle from 4.x onward is fine for me, so I'm also in favor of removing `RestfulService`. I guess a wrapper won't be necessary, guzzle seems simple enough.

kinglozzer:
![atlast](https://cloud.githubusercontent.com/assets/1655548/14656623/860d38fc-0680-11e6-9159-e6d569838b60.gif)

I don’t see any need for maintaining a restfulservice-like wrapper around Guzzle. We may as well just move the class into a “module” for people who want a marginally easier upgrade path and just never enhance it. Not that I’m advocating that, I think we can just drop it.

dhensby:
I'd kill Oembed too ;) (there are thirdparty libs that can do it better!)

tractorcow:
[...]

PR #5461, opened 2016-05-06
RFC #5347 remove RestfulService from SS4.

Firesphere:
Rebuild Oembed on Guzzle 5 or 6. 
Note:
Requires update on DynameDB module to use newer AWS-SDK


composer.json
```diff
@@ -20,7 +20,8 @@
 		"composer/installers": "~1.0",
 		"monolog/monolog": "~1.11",
 		"league/flysystem": "~1.0.12",
-		"symfony/yaml": "~2.7"
+		"symfony/yaml": "~2.7",
+		"guzzlehttp/guzzle": "~5.3|~6.0.1|~6.1"
```
kinglozzer: What’s the thinking behind these constraints? Given that there could be breaking changes between versions 5 and 6 (I haven’t checked, but semver yo), shouldn’t we pick a specific major version?

oembed/Oembed.php
```diff
@@ -103,13 +103,14 @@ protected static function matches_scheme($url, $scheme) {
...
 	{
-		// Fetch the URL (cache for a week by default)
-		$service = new RestfulService($url, 60 * 60 * 24 * 7);
-		$body = $service->request();
-		if (!$body || $body->isError()) {
+		$service = new GuzzleHttp\Client();
+
+		$request = $service->request('GET', $url);
+		// Fetch from Oembed URL (cache for a week by default)
+		if(!$request || $request->getStatusCode() !== 200) {
```
kinglozzer: Comment says “Fetch from Oembed URL (cache for a week by default)”, but there’s no caching logic here any more :(. We could use `SS_Cache` instead, or just leave out the caching and remove this comment

composer.json
Firesphere: There are no breaking changes for this, I choose this, to comply with DynamoDB module, which is going to require AWS SDK 3, which has the same requirements.

As it was a hackday implementation, just copy-pasting the AWS-SDK versions was the fastest way to get it out. I guess just ^6.2 might even work.

oembed/Oembed.php
Firesphere: Same as previous response mainly.
Forgot to take it out because more focused on getting it done than focus on everything around it due to hackday.

Caching is by the way hardly ever of any use, unless the embedded item is embedded a dozen times a minute via the CMS.

Firesphere: By the way, I'm not excusing myself for being sloppy ;) I'm explaining why this is slightly sloppy ;)

Firesphere:
I'll see if I can clean it up a bit over the weekend and squash it :) 
I just wanted it out asap. No excuse :)

Firesphere:
Fixed issues on last force push.

dhensby:
Including guzzle just for oembed is major overkill. I believe the rfc stated just using curl. We could use file_get_contents too.

Firesphere:
Although it's overkill now, for just Oembed, I decided to do include Guzzle, because it's a much used module.
Having it readily available with the framework is a good thing, I believe.
Basically, it's the same as saying "Why remove RestfulService, as it's only used in Oembed". The reason to go with including Guzzle, is to give more flexibility, and possibly/probably use it more in the future and make the obvious switch to something that is more robust.

dhensby:
We want to remove RestfulService so we don't have the maintenance burden of it and we can remove bloat that comes with framework. 

Composer makes including guzzle easy so if devs need it they can include it. Every module we declare as a dependency the core team takes on an obligation to maintain and support that module too (its part of the supported module standard). 

I would personnally pull oembed out of core too, but that's a side point for now. Having to include guzzle to make a GET request is a bit OTT

Firesphere:
From that point of view, indeed it makes sense.
I'm curious of what others think of it? I can easily rewrite it to use either a wget or maybe even remove Oembed? (Although I don't think that's a good idea right now)
To be honest, I think maintaining a Guzzle implementation will cost less overhead than maintaining than using e.g. file_get_contents, which is even in it's documentation stating, it's not a reliable way of getting data.
```This function may return Boolean FALSE, but may also return a non-Boolean value which evaluates to FALSE. Please read the section on Booleans for more information. Use the === operator for testing the return value of this function.```
I do agree however, that it may cause maintenance in the future. But I doubt it'll be much.

**Comment by dhensby** [comment]
I agree not to remove oembed at this point in time. 

We can't refuse to use functions that return explicitly false and falsely values, then we can't use strpos or so on. 

If we use file_get_contents we can actually remove curl asthma dependency of the framework too, which would be cool. 

Given the RFC explicitly mentions a solution for oembed that is not guzzle, I would suggest the concencus on this point has already been reached. Happy for others to chime in and overrule me, though.

**Comment by Firesphere** [comment]
I'll setup a get_contents version and await others opinions. It's not much work to fix.
My personal preference is still using Guzzle though ;)

**Comment by dhensby** [comment]
If we had more than 2 needs in the framework to make HTTP requests, I'd support bringing in guzzle in a flash

**Comment by Firesphere** [comment]
I admit being stubborn here, but would you be upset if we wait for the others to give feedback as well?
I'm more than happy to change it if overruled of course.

**Comment by stevie-mayhew** [comment]
Using guzzle is a good idea, but not for this stage. For now we should just use file_get_contents (or curl as mentioned in the RFC) and then we can look to update to using a guzzle like adapter in the future. This may not be guzzle and should be built using http://www.php-fig.org/psr/psr-7/ so that users can swap their preferred HTTP client in/out. I agree with Dan that we shouldn't have an explicit dependency on guzzle specifically.

**Comment by dhensby** [comment]
@firesphere happy to wait and likewise be overruled

**Comment by Firesphere** [comment]
@stevie-mayhew The "problem" I think, with not using a well established library, is that it'll become ignored or outdated. 
I agree on the swappability though. But using cURL or an equivalent, is not really swappable either.
I've got a cURL and get_contents branch ready for push and pull request. But I'll await the others first.

**Comment by dhensby** [comment]
@silverstripe/core-team ping

**Comment by sminnee** [comment]
OK, so I've looked at a few oEmbed libraries, e.g. https://github.com/oscarotero/Embed and https://github.com/essence/essence. Neither of these require Guzzle, although essence seems to have its own family of packages.

It would be good to replace our oEmbed classes with one of these libraries before the 4.0 stable is out. My vote wold go with the oscarotero package, because it's more popular, seems to be more consistently contributed to, and has fewer external dependencies.

In the short term, we need to get this PR merged. I think it would probably be safer to merge a change that used curl, and then in a subsequent move to a 3rd party oEmbed package, rather than to introduce guzzle as a dependency and then remove it as a dependency later.

@Firesphere are you okay with removing the Guzzle dependency? Do you want to take on refactoring out our hand-rolled oEmbed implementation too?

**Comment by chillu** [comment]
Just noting that guzzle is pretty lightweight as far as libs go: 1MB on disk, four deps.

```
composer global require --prefer-dist guzzlehttp/guzzle
Changed current directory to /Users/ingo/.composer
Using version ^6.2 for guzzlehttp/guzzle
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
  - Installing guzzlehttp/promises (1.1.0)
    Downloading: 100%         

  - Installing psr/http-message (1.0)
    Downloading: 100%         

  - Installing guzzlehttp/psr7 (1.3.0)
    Downloading: 100%         

  - Installing guzzlehttp/guzzle (6.2.0)
    Downloading: 100%         

Writing lock file
Generating autoload files

du -h .composer/vendor/guzzlehttp/
...
968K	.composer/vendor/guzzlehttp/
du -h .composer/vendor/psr/
... 
84K	.composer/vendor/psr/
``

But in the bigger picture of replacing the `OEmbed` logic with a library, I agree that adding the guzzle dependency doesn't make a lot of sense right now - +1 for a simple `curl` based replacement in the meantime.

Firesphere:
Ok. Yep, I'll fix this part to use cURL.
I can start on removing oEmbed completely, but that's a different issue, so I'll get to that on a different branch. I don't think I can make it before Alpha-1 though... That's out today, isn't it?

sminnee:
Alpha2 is fine.

tractorcow:
thanks @Firesphere looking forward to getting this in. :)

Firesphere:
Since the next thing, is strip out oEmbed, I have not taken the time to make it completely and extremely clean (e.g. the `$body = $response` )
But this is with cURL, working mighty fine.

dhensby:
Looks reasonable - how about a mention in upgrade notes that restfulservice is removed and you've created a dead fork for drop-in replacement?

Firesphere:
@dhensby pointing back to the original RFC, as a mention, I can update the docs as well in a separate PR.
I haven't, because it was a hackday project to fix it at first. No excuse obviously, I just didn't look at the docs at all, just focused on getting the RestfulService out.
I've made a note in my private calendar, to remind me the docs needing an update :) I think I'll do that after also fixing the removal of oembed, to make it more a single PR of deprecations for SS4. Is that okay with you guys?

dhensby:
I'm happy with that.

PR #5466, opened 2016-05-06
Allow THEMES_DIR to be modified

wilr:
This allows developers to override THEMES_DIR, allowing source code and themes to remain inside one folder (e.g point to mysite/themes) rather than separate folders for mysite and themes.

### DIFF
--- a/core/Constants.php
+++ b/core/Constants.php
@@ -231,8 +231,13 @@ if(!defined('BASE_URL')) {
 }
 define('MODULES_DIR', 'modules');
 define('MODULES_PATH', BASE_PATH . '/' . MODULES_DIR);
-define('THEMES_DIR', 'themes');
+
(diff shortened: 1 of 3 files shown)


dhensby:
Some strange test failures... Any idea what's going on?

tractorcow:
Looks like issues with the namespace changes made recently. Rebase and try again?

sminnee:
@wilr I think that this has been superceded by the recent work on the themes system? Re-open if you disagree, but the key now is that all the module/mysite templates are treated as "themes" as well, and you can choose any priority order that you like for their application.

My hope is that by the time 4.0 stable comes out, the themes can be stored in the vendor/ folder or your project root, and the `themes/` folder no longer exists.

wilr:
Yep perfect for our use case.

Issue #5397, opened 2016-04-26
Fix breadcrumb display in admin/campaigns list item view


chillu:
It's currently hardcoded to "March Release"

clarkepaul:
Related issue https://github.com/silverstripe/silverstripe-framework/issues/5438

chillu:
For alpha1, simply use a `[placeholder]` text

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-framework/pull/5474

clarkepaul:
The title of this issue was misleading as the list item view has no breadcrumbs, I had the same issue open with the appropriate designs here https://github.com/silverstripe/silverstripe-framework/issues/5438

PR #5474, opened 2016-05-09
API Implement breadcrumbs via controllable state

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/5397

Demonstrates viewing specific record:

image

Not perfect, but could be improved. 💃 

image


admin/client/src/components/Toolbar/Toolbar.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import { connect } from 'react-redux';
```
chillu: This makes the component stateful, and less reuseable - it should be implemented on the container instead (`<CampaignAdmin>`) - having a look at refactoring that now

admin/client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -59,6 +64,35 @@ class CampaignAdmin extends SilverStripeComponent {
...
+    }]);
+
+    switch (view) {
+      case 'show':
+        // NOOP - Lazy loaded in CampaignAdminList.js
+        break;
+      case 'edit':
+        this.props.breadcrumbsActions.pushBreadcrumb({
+          text: i18n._t('Campaigns.EDIT_CAMPAIGN', 'Editing Campaign'),
+          href: `admin/campaigns/set/${id}/edit`,
```
chillu: Should use `sectionConfig.campaignViewRoute`

admin/client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -30,7 +31,16 @@ class CampaignAdminList extends SilverStripeComponent {
...
     const fetchURL = this.props.itemListViewEndpoint.replace(/:id/, this.props.campaignId);
     super.componentDidMount();
-    this.props.recordActions.fetchRecord('ChangeSet', 'get', fetchURL);
+    this.props.recordActions.fetchRecord('ChangeSet', 'get', fetchURL)
+      .then(() => {
+        // Setup breadcrumbs if record is loaded
+        if (this.props.record) {
+          this.props.breadcrumbsActions.pushBreadcrumb({
+            text: this.props.record.Name,
+            href: `admin/campaigns/set/${this.props.record.ID}/show/`,
```
chillu: Should use `sectionConfig.campaignViewRoute`

admin/client/src/state/breadcrumbs/BreadcrumbsReducer.js
```diff
@@ -0,0 +1,31 @@
...
+
+const initialState = deepFreeze({
+  breadcrumbs: [],
+});
+
+function reducer(state = initialState, action) {
+  let breadcrumbs = state.breadcrumbs.slice(0);
+  switch (action.type) {
+
+    case ACTION_TYPES.SET_BREADCRUMS:
```
chillu: Typo

admin/client/src/state/breadcrumbs/BreadcrumbsReducer.js
```diff
@@ -0,0 +1,31 @@
...
+      return deepFreeze(Object.assign({}, state, {
+        breadcrumbs: action.payload.breadcrumbs,
+      }));
+
+    case ACTION_TYPES.PUSH_BREADCRUMB:
+      breadcrumbs.push(action.payload.breadcrumb);
+      return deepFreeze(Object.assign({}, state, { breadcrumbs }));
+
+    case ACTION_TYPES.PUBLISH_CAMPAIGN_REQUEST:
```
chillu: Shouldn't be in here?

admin/client/src/state/breadcrumbs/BreadcrumbsActions.js
```diff
@@ -0,0 +1,40 @@
...
+  };
+}
+
+/**
+ * Adds a breadcrumb to the end of the list
+ *
+ * @param {object} Breadcrumb to add. Has 'text' and optional 'href' property.
+ * @return {function}
+ */
+export function pushBreadcrumb(breadcrumb) {
```
chillu: I would prefer if we just allow setting of breadcrumbs, instead of having it behave like a stack. Every view component which wants to set breadcrumbs should know its full context. For example, a GridField listing folder file contents should know all breadcrumbs required for this (parent folders). If you navigate up to the parent folder, *that* GridField should also know all breadcrumbs, rather than knowing about the navigation "direction" (drill down, go back up).

Additionally, "push" and "pop" aren't "safe" state transformations - if you replay them you'll get different results - which should be avoided in a Flux-style state management.

clarkepaul:
You shouldn't use "Editing campaign" instead of the actual campaign name, use the actual campaign name in the breadcrumb.

tractorcow:
> You shouldn't use "Editing campaign" instead of the actual campaign name, use the actual campaign name in the breadcrumb.


I mentioned that was left as a possible improvement for later, but if we need it for alpha 1 then I'll update it.

admin/client/src/state/breadcrumbs/BreadcrumbsActions.js
tractorcow: The problem is that gridfields attached to records won't know anything about the modeladmin that model is attached to. It won't always know the full context.

How about parent components have a "base" breadcrumbs that are pushed to child components (via properties, not states). Then I'll remove push / pop?

admin/client/src/state/breadcrumbs/BreadcrumbsReducer.js
tractorcow: Oops....

admin/client/src/state/breadcrumbs/BreadcrumbsReducer.js
tractorcow: crums!

admin/client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -59,6 +64,35 @@ class CampaignAdmin extends SilverStripeComponent {
     });
   }
 
+  setupBreadcrumbs(id, view) {
+    // Set root breadcrumb
+    this.props.breadcrumbsActions.setBreadcrumbs([{
+      text: i18n._t('Campaigns.CAMPAIGN', 'Campaigns'),
+      href: 'admin/campaigns/',
```
chillu: Shouldn't be hardcoded

tractorcow:
I've rebased and updated.

I can't get npm tests working; I'll need some help tomorrow with the redux mocking, @chillu ,please.

admin/client/src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -38,4 +39,10 @@ Breadcrumb.propTypes = {
   crumbs: React.PropTypes.array,
 };
 
-export default Breadcrumb;
+function mapStateToProps(state) {
```
chillu: ReactJS components placed in `components/Breadcrumb` should not be tied to Redux state, this should be fed into the component by the "parent" container component.

admin/client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -40,10 +44,13 @@ class CampaignAdmin extends SilverStripeComponent {
...
         // state / redux
-        this.props.actions.showCampaignView(ctx.params.id, ctx.params.view);
+        this.props.campaignActions.showCampaignView(ctx.params.id, ctx.params.view);
+
+        // reset admin to top level of breadcrumbs
+        this.setupBreadcrumbs(ctx.params.id, ctx.params.view);
       } else {
         // If component is not mounted, we need to allow root routes to load
-        // this section in via ajax
+        // this section in via ajaxan
```
chillu: typo

admin/client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -330,12 +372,14 @@ function mapStateToProps(state, ownProps) {
     sectionConfig: state.config.sections[ownProps.sectionConfigKey],
     campaignId: state.campaign.campaignId,
     view: state.campaign.view,
+    breadcrumbs: state.breadcrumbs.breadcrumbs,
```
chillu: Why the double nesting of `breadcrumbs.breadcrumbs`?

admin/client/src/state/breadcrumbs/BreadcrumbsReducer.js
```diff
@@ -0,0 +1,22 @@
...
+const initialState = deepFreeze({
+  breadcrumbs: [],
+});
+
+function reducer(state = initialState, action) {
+  switch (action.type) {
+
+    case ACTION_TYPES.SET_BREADCRUMBS:
+      return deepFreeze(Object.assign({}, state, {
+        breadcrumbs: action.payload.breadcrumbs,
```
chillu: This should just be `deepFreeze(Object.assign({}, action.payload.breadcrumbs))`

admin/client/src/containers/CampaignAdmin/CampaignAdmin.js
tractorcow: Because the state key is breadcrumbs, the state holds a single value.

admin/client/src/state/breadcrumbs/BreadcrumbsReducer.js
tractorcow: Ok, I'll try un-nesting it.

admin/client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -59,6 +66,53 @@ class CampaignAdmin extends SilverStripeComponent {
...
+        break;
+      default:
+        // NOOP
+        break;
+    }
+
+    this.props.breadcrumbsActions.setBreadcrumbs(breadcrumbs);
+  }
+
+  handleBackButtonClick() {
```
chillu: Missing `event`

Issue #5328, opened 2016-04-17
No CSRF protection in "delete campaign" feature (React-based GridField)


chillu:
Requires https://github.com/silverstripe/silverstripe-framework/pull/5316 to be merged first (since it exposes `SecurityID` via Redux state. It should also use `createEndpointFetcher()` in `silverstripe-backend`, rather than passing the endpoint url and method straight through the Redux actions. 

Made a start on this:

```diff
diff --git a/admin/javascript/src/components/grid-field/index.js b/admin/javascript/src/components/grid-field/index.js
index c709e9d..fb8c974 100644
--- a/admin/javascript/src/components/grid-field/index.js
+++ b/admin/javascript/src/components/grid-field/index.js
@@ -136,12 +136,18 @@ GridField.propTypes = {
     headerColumns: React.PropTypes.array,
     collectionReadEndpoint: React.PropTypes.object,
     handleDrillDown: React.PropTypes.func,
+    itemDeleteEndpoint: React.PropTypes.shape({
+      method: React.PropTypes.string,
+      url: React.PropTypes.string,
+    }),
   }),
+  securityID: React.PropTypes.string,
 };
 
 function mapStateToProps(state, ownProps) {
   const recordType = ownProps.data ? ownProps.data.recordType : null;
   return {
+    securityID: state.config.SecurityID,
     records: (state.records && recordType) ? state.records[recordType] : {},
   };
 }
diff --git a/admin/javascript/src/sections/campaign-admin/controller.js b/admin/javascript/src/sections/campaign-admin/controller.js
index 67e5438..dd61237 100644
--- a/admin/javascript/src/sections/campaign-admin/controller.js
+++ b/admin/javascript/src/sections/campaign-admin/controller.js
@@ -151,6 +151,7 @@ class CampaignAdminContainer extends SilverStripeComponent {
 
 CampaignAdminContainer.propTypes = {
   config: React.PropTypes.shape({
+    SecurityID: React.PropTypes.string,
     forms: React.PropTypes.shape({
       editForm: React.PropTypes.shape({
         schemaUrl: React.PropTypes.string,
```

Jira: https://silverstripe.atlassian.net/browse/OSS-1693

PR #5477, opened 2016-05-09
API Add CSRF to Campaign delete

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/5328


tractorcow:
Rebased.

chillu:
I probably would've done this via parsing the HTTP body of the `DELETE` statement instead in the new `SecurityToken-> getRequestToken()` method: HTTP headers are a bit more likely than HTTP body to be sanitised by overzealous proxies and firewalls. Unlikely to be an issue in practice though.

@dhensby You've literally pressed the merge button five seconds before me :D

dhensby:
:)

tractorcow:
I believe I was following the best practices of joomla when implementing this. I think they use the same process.

PR #5480, opened 2016-05-09
Disable subsite filters when running unit tests

mikenz:
If the subsites module is installed when running unit test it causes a number of framework unit tests to fail. This change disables the Subsite filter by default when running tests so they normally. 

Having framework needing to know about another module isn't ideal but I can't see another way of doing this right now.

### DIFF
--- a/dev/SapphireTest.php
+++ b/dev/SapphireTest.php
@@ -207,6 +207,7 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 
 		if(class_exists('RootURLController')) RootURLController::reset();
 		if(class_exists('Translatable')) Translatable::reset();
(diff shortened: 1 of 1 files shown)


tractorcow:
This actually breaks subsites tests in CI. Can we revert this please? Having this kind of interdependency makes testing fragile.


```php
/**
 * Tests the initial state of disable_subsite_filter
 */
function testDisableSubsiteFilter() {
	$this->assertFalse(Subsite::$disable_subsite_filter);
}
```

If framework tests fail because their behaviour is modified by external modules (such that the context of the tests are substantially different) then it's acceptable to have those tests failed. In our internal testing we mute these with explanations on each.

PR #5481, opened 2016-05-09
Explictly set the FileNameFilter replacements before testing them

mikenz:
If the developer overrides the FileNameFilter default_replacements the unit tests will fail. Explicitly set the defaults before testing that they work.

### DIFF
--- a/tests/filesystem/FileNameFilterTest.php
+++ b/tests/filesystem/FileNameFilterTest.php
@@ -5,6 +5,18 @@
  */
 class FileNameFilterTest extends SapphireTest {
 
+    public function setUp() {
+        parent::setUp();
+
+        Config::inst()->update('FileNameFilter', 'default_replacements', array(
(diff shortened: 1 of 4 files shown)


dhensby:
I think we need a better way of doing this than hard coding the defaults into the tests but I'm not sure what...

mikenz:
No arguments there :-) The tests assume a bunch of config is in it's default state but that isn't ever verified.

tractorcow:
Instead of hard coding them into all tests, you can hard code them into a single test (FileNameFilterTest) and simply reference it from there.

E.g.

```php
Config::inst()->update('FileNameFilter', 'default_replacements', FileNameFilterTest::$default_replacements);
```

sminnee:
@tractorcow the negative of that is that we get cross-coupled tests.

I say we merge this, and if someone wants to do a better fix later, so much the better.

Sound good?

PR #5482, opened 2016-05-09
Remember the basic details about a user when they are deleted

mikenz:
When looking a page history, or other actions it is not possible to attribute these to a Member once they are deleted. We simply display "Unknown".

This change creates a DeletedMember object and stores the basic details about a member when they are deleted. This allows other modules to retrieve the name of members after they have been deleted so they can still be 'blamed' for the actions.

### DIFF
--- a/security/DeletedMember.php
+++ b/security/DeletedMember.php
@@ -1,0 +1,40 @@
+<?php
+/**
+ * The delete member class which represents a user that has been deleted from the system
+ */
+class DeletedMember extends DataObject {
+
+    private static $db = array(
+        'FirstName' => 'Varchar',
+        'Surname' => 'Varchar',
+        'Email' => 'Varchar',
+        'NumVisit' => 'Int',
+        'LastVisited' => 'SS_Datetime',
+        'Locale' => 'Varchar(6)',
+        // In ISO format
+        'DateFormat' => 'Varchar(30)',
+        'TimeFormat' => 'Varchar(30)',
+    );
+
+    function getFirstName() {
+        return '(deleted) ' . $this->getField('FirstName');
+    }
+
+    public function getTitle() {
+        return '(deleted) ' . $this->getField('Surname') . ', ' . $this->getField('FirstName');
+    }
+
+    public function getEmail() {
+        return '(deleted) ' . $this->getField('Email');
+    }
+
+    /**
+     * Get the complete name of the member
+     *
+     * @return string Returns the first- and surname of the member.
+     */
+    public function getName() {
+        return ($this->data()->Surname) ? trim($this->data()->FirstName . ' ' . $this->data()->Surname) : $this->data()->FirstName;
+    }
+
+}
--- a/security/Member.php
+++ b/security/Member.php
@@ -921,6 +921,20 @@ class Member extends DataObject implements TemplateGlobalProvider {
 		}
 	}
 
+    /**
+     * Remeber the basic details of a member when they are deleted
+     */
+    public function onBeforeDelete() {
+        $deleted = new DeletedMember();
+        $original = $this->toMap();
+        unset($original['ClassName']);
+        foreach ($original as $k => $v) {
+            $deleted->$k = $v;
+        }
+        $deleted->write();
+        parent::onBeforeDelete();
+    }
+
 	public function onAfterDelete() {
 		parent::onAfterDelete();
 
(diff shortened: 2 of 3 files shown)


tractorcow:
Could we get the same value by, instead of having a separate object, add Versioned to Member, but set it to "version only" (no staging).

This would give us the added benefit of being able to track changes to members over time.

```
class Member extends DataObject {
	private static $extensions = [
		'Versioned('Versioned')' // Versioning without staging
	];
}

class SomeRecord extends DataObject {
	private static $has_one = [
		'Owner' => 'Member
	];
	
	public function Owner() {
		return Versioned::get_latest_version('Member', $this->MemberID);
	}
	
	public function getOwnerName() {
		$owner = $this->Owner();
		if(!$owner) {
			return null;
		}
		if($owner->isOnDraft()) {
			return $owner->Title;
		}
		else {
			return $owner->Title . ' (deleted)';
		}
	}
}
```

mikenz:
I thought about that, but then a new version would be created for every login and failed login attempt when those fields are updated? Is there a way to specify some fields that don't cause a new version to be created?

tractorcow:
> I thought about that, but then a new version would be created for every login and failed login attempt when those fields are updated? Is there a way to specify some fields that don't cause a new version to be created?
 
 
 ```php
 $member->writeWithoutVersion();
 ```

dhensby:
This is a useful feature for *some* projects.I think that this is something that would be better as a module and not in core.

dhensby:
A simpler solution for the history would be to de-normalise the name/email

mikenz:
I would argue (well my users do) that the is needed for _all_ projects. An asset being uploaded by 'unknown' or a page last been edited by 'unknown' after removing an ex-team member is a "bug" in world of my users.

I'm open to re-implementing using another method but it'll be a few months off before I get back to it.

sminnee:
> Could we get the same value by, instead of having a separate object, add Versioned to Member, but set it to "version only" (no staging).

This seems like a much bigger impact, because of the complexity that versioned introduces. However, refactoring this to be a simpler  `RememberDeletedRecords extends DataExtension` would make it easier to apply to objects other than `Member`, which would be helpful.

> I would argue (well my users do) that the is needed for all projects. An asset being uploaded by 'unknown' or a page last been edited by 'unknown' after removing an ex-team member is a "bug" in world of my users.

We'd need to update all the other core code to reference deleted member at the right place. Right now, this doesn't fix the bug. It highlights one difficulty of this approach, that you need to have a bunch of irritating `if()` switches every time you look for a member.

Because of this, it might be better to approach this by adding an `IsDeleted` boolean column to the `Member` table, and to have all queries add `AND IsDeleted = 0` by default (you can do this with a DataExtension). Then we can add a query flag to include deleted records in the query, which would suppress that filter.

Your current solution also doesn't work for custom member fields. What if someone adds a Username field and uses that to identify member? Now they've got to customise both the `Member` and `DeletedMember` tables.

> This is a useful feature for some projects. I think that this is something that would be better as a module and not in core.

My main concern is that introducing a new DataObject is quite a heavy burden for fixing 1 edge-case bug. If we implement this as a DataExtension, we could potentially have it in core but not enabled by default.

axyr:
Is this not a candidate for a SoftDeletable trait?

sminnee:
@axyr a `DataExtension` will give you more ability to manipulate the database queries and schema than a trait would. but yeah, SoftDeletable is a better name than RememberDeletedRecords. ;-)

tractorcow:
So an extension that automatically filters out by `Deleted = true`, except when querying records by ID?

Does this mean it doesn't have to be a core feature, and can be done in a module? If that's the case, we should close this PR and not have it in core at all.

dhensby:
I don't think getting by I'd should change the behaviour of soft delete

axyr:
by id should also filter "where deleted = 0 " by default if you ask me, unless something like :

```php
return Member::get()->includeDeleted()->byID(1);
```

I think it should be in core, but disabled by default.

Just like Created and LastEdited should be in a 'Timestampable' trait/extension if you ask me...

dhensby:
I think a soft delete extension could exist in core. The point of this thread is to not loose information required by core.

But who will volunteer to do it ;)

axyr:
I will take a shot on it this weekend. Seems like a nice kata..

sminnee:
> `return Member::get()->includeDeleted()->byID(1);`

@axyr rather than a new `includeDeleted()` method it's probably better to use `DataList::setDataQueryParam`. See Versioned for an example.

I'm going to close this PR and raise an issue outlining the alternative implementation.

PR #5483, opened 2016-05-09
Don't use Object.freeze on functions (fixes #5473)

### LINKED ISSUE #5473: CampaignAdmin doesn't work in IE10 and IE Edge
`Unhandled promise rejectionTypeError: Object doesn't support this action`. This is a hard one to track down because `Object` is used all over the place. Probably a missing ES6 polyfill. I've tried using bluebird rather than es6-promise to catch unhandled promise rejections, without much success yet.

image

chillu:
We're not using it for any other props passed to ReactJS components,
so there's no reason to do it here. Props are immutable by convention.
While it would be nice to enforce this, its too common to pass through
function objects which aren't supported by IE's Object.freeze().

IE isn't following the spec on how to handle Object.freeze(function() {}).
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze#Notes
> In ES6, a non-object argument will be treated as if it was a  frozen ordinary object, simply return it.

MS docs on https://msdn.microsoft.com/en-us/subscriptions/downloads/ff806186(v=vs.94).aspx
> If the object argument is not an object, a TypeError exception is thrown.


sminnee:
Seems fine to me. Would using the `immutable` npm library be potentially a better long-term solution to this issue (post-alpha, I mean)?

chillu:
Yeah https://github.com/facebook/immutable-js is quite powerful, but also adds ~30kb minified to the build. You can do a lot with `Object.assign()` and the ES6 Object spread operator these days. For now, I'd say we just keep an eye on dev productivity and code readability when writing reducers in native ES6.

PR #5488, opened 2016-05-09
Fix for array to string conversion in DropdownField

### LINKED ISSUE #4835: Array to string conversion notice with PHP > 5.4.x
We're getting Array to string conversion notice on Line 175 in framework/forms/DropdownField.php using php 5.4 or 5.5:

`$selected = ($value === $this->value) || (((string) $value) === ((string) $this->value));`

When $this->value is an array (this happens for example when archiving a blog post with the blog module) this throws a notice with PHP > 5.4.x while it doesn't with PHP 5.3.x

```
[Notice] Array to string conversion
GET /admin/pages/edit/show/92

Line 175 in framework/forms/DropdownField.php

Source

166 			foreach($source as $value => $title) {
167 				$selected = false;
168 				if($value === '' && ($this->value === '' || $this->value === null)) {
169 					$selected = true;
170 				} else {
171 					// check against value, fallback to a type check comparison when !value
172 					if($value) {
173 						$selected = ($value == $this->value);
174 					} else {
175 						$selected = ($value === $this->value) || (((string) $value) === ((string) $this->value));
176 					}
177 
178 					$this->isSelected = $selected;
179 				}
180 
181 				$disabled = false;
```

This with 5.5.30, and I was getting the same also with a version of 5.4 but it is ok with 5.3 (e.g. it is working fine with 5.3.28)

This is our composer.json:

```
"require": {
		"php": ">=5.3.2",
		"silverstripe/cms": "3.2.1",
		"silverstripe/framework": "3.2.1",
		"silverstripe-themes/simple": "*",
		"silverstripe/blog": "~2.3@stable",
		"undefinedoffset/sortablegridfield": "dev-master"
	},
```

Steps to reproduce:
* create a new blog post
* go to /admin/pages and open the blog post to edit it
* archive (delete) the blog post
* you should see the error message in the browser inspector network tab
* you can see the error again by trying to access the deleted page by manually specifying the url (e.g. /admin/pages/edit/show/92)

Thanks.

robbieaverill:
* Resolves #4835

---

### Note:

* There is likely to be a better way to fix this
* I haven't tested any edge cases

### Tested:

* Archiving a blog post

### DIFF
--- a/tests/forms/DropdownFieldTest.php
+++ b/tests/forms/DropdownFieldTest.php
@@ -224,6 +224,32 @@ class DropdownFieldTest extends SapphireTest {
 	}
 
 	/**
+	 * The Field() method should be able to handle arrays as values in an edge case. If it couldn't handle it then
+	 * this test would trigger an array to string conversion PHP notice
+	 *
+	 * @dataProvider arrayValueProvider
+	 */
+	public function testDropdownWithArrayValues($value) {
+		$field = $this->createDropdownField();
+		$field->setValue($value);
+		$this->assertInstanceOf('HTMLText', $field->Field());
+        $this->assertSame($value, $field->Value());
+	}
+
+	/**
+	 * @return array
+	 */
+	public function arrayValueProvider() {
+		return array(
+			array(array()),
+			array(array(0)),
+			array(array(123)),
+			array(array('string')),
+			array('Regression-ish test.')
+		);
+	}
+
+	/**
 	 * Create a test dropdown field, with the option to
 	 * set what source and blank value it should contain
 	 * as optional parameters.
--- a/forms/DropdownField.php
+++ b/forms/DropdownField.php
@@ -172,7 +172,12 @@ class DropdownField extends FormField {
 					if($value) {
 						$selected = ($value == $this->value);
 					} else {
-						$selected = ($value === $this->value) || (((string) $value) === ((string) $this->value));
+						// Safety check against casting arrays as strings in PHP>5.4
+						if(!is_array($value) && !is_array($this->value)) {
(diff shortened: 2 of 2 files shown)


tractorcow:
In cases where either $value or $this->value is an array, what are the contents of these arrays? Unit test would help demonstrate how this fixes the issue.

robbieaverill:
In the case of archiving a blog post (for me anyway) $this->value was just
an empty array. I'll add a test.

robbieaverill:
@tractorcow I've added a test which uses an array as a value. The assumption I've taken with this change is that if there's an array as the value it's an accident, and we should assume that it won't be the "selected value".

tests/forms/DropdownFieldTest.php
```diff
@@ -224,6 +224,31 @@ public function testNumberOfDisabledOptions() {
...
 	/**
+	 * The Field() method should be able to handle arrays as values in an edge case. If it couldn't handle it then
+	 * this test would trigger an array to string conversion PHP notice
+	 *
+	 * @dataProvider arrayValueProvider
+	 */
+	public function testDropdownWithArrayValues($value) {
+		$field = $this->createDropdownField();
+		$field->setValue($value);
+		$this->assertInstanceOf('HTMLText', $field->Field());
```
dhensby: I think we can have a more thorough assertion here to test the behaviour is as expected (ie: value is correctly set) rather than simply that there is no error.

robbieaverill: Hey @dhensby - I've added an assertion for the value against the expectation as well.

dhensby:
waiting for green

dhensby:
that was a long wait :p thanks!

robbieaverill:
No worries!

PR #5491, opened 2016-05-10
API Add empty campaign layout

tractorcow:
image

See https://silverstripe.atlassian.net/browse/OSS-1807


admin/client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -134,27 +134,31 @@ class CampaignAdminList extends SilverStripeComponent {
...
-      ? 'cms-content__split cms-content__split--left-sm'
-      : 'cms-content__split cms-content__split--none';
+    // Set body
+    const body = accordionBlocks.length
+      ? (<Accordion>{accordionBlocks}</Accordion>)
+      : (
+        <div className="alert alert-warning" role="alert">
+          <strong>This campaign is empty.</strong> You can add pages by selecting{' '}
+          <em>Add to campaign</em> from within the <em>More Options</em> popup on{' '}
+          the edit page screen.
```
hafriedlander: Any chance we can make "the edit page screen" a link? Or if not, at least change to "the Pages section" ("edit page" doesn't appear in the CMS)

admin/client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -134,27 +134,31 @@ class CampaignAdminList extends SilverStripeComponent {
...
+        </div>
+      );
 
     return (
-      <div className={classNames}>
+      <div className="cms-content__split cms-content__split--left-sm">
```
hafriedlander: Why are these moved from {classNames} to hard coding like this - unlikely to ever need overriding / adding to?

tractorcow: Before we were hiding the preview, but now if there's no preview, we leave the panel open, but display a message if there is no preview available.

admin/client/src/containers/CampaignAdmin/CampaignAdminList.js
clarkepaul: I would be ok with "Edit this page" or "Edit this file", which then has both the section name and makes sense to a user but I don't really think "the Pages section" is meaningful enough, ie it doesn't say take me to the page with the pages section.

FYI this is just a temporary button for alpha1, well there are plans to make editing availalbe within campaigns.

tractorcow: There is no page, though. This is a campaign with no records attached.

clarkepaul: I miss read that, ignore what I just siad

clarkepaul: If you were to add that link it should be added to the notice rather than the preview.
For the preview message I would either leave the text "no preview available" or remove the text completely.

PR #5499, opened 2016-05-10
FIX: SQLSelect count methods now cast to int (fixes #5498)

### LINKED ISSUE #5498: SQLSelect::count() returns a string, not an integer as documented
The method just needs to be updated to cast to an int. Presumably `unlimitedRowCount()` has the same issue, though I’ve not checked yet

kinglozzer:


dhensby:
Seems to need rebasing :S

PR #5500, opened 2016-05-10
Use props.extraClass on TextField holder

chillu:
This should be separated out to a common FormField class,
but for now we're only using TextField. Can be solved
properly at the same time as switching form fields to
react-bootstrap.

Required for readonly field value alignment in the "campaigns" edit form.


chillu:
Wrong branch...

PR #5505, opened 2016-05-10
Added italian localization to jqueryui datepicker

g4b0:
so it's possible to use ->setConfig('jslocale','it') for DateField


chillu:
Thanks Gabriele!

PR #5511, opened 2016-05-11
Consistent React event method naming

chillu:
Use "on<event>" for props, same as React's own event naming: https://facebook.github.io/react/docs/forms.html#interactive-props
Use "handle<event>" to delineate internal handlers

Merge together with https://github.com/silverstripe/silverstripe-asset-admin/pull/154


tractorcow:
Rebased and merged separately.

tractorcow:
https://github.com/silverstripe/silverstripe-framework/commit/0838770798eb82618c32c8bb803fb94a1566d9fb

Issue #5508, opened 2016-05-10
Campaign title in breadcrumb is not vertically centered


clarkepaul:
![pasted_image_11_05_16__11_49_am](https://cloud.githubusercontent.com/assets/555033/15166002/8f7ec9ba-176e-11e6-9436-db960ceeb83e.png)

PR #5512, opened 2016-05-11
Reorder Campaigns, vert centre only child breadcrumb

clarkepaul:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/5508

Was:
image

After:
pasted_image_11_05_16__3_38_pm


clarkepaul:
Couldn't figure out how to move Campaigns further down but I think this position or after Files would be better than being in first position.
Maybe it was just caching but couldn't find the magic number.

PR #5530, opened 2016-05-12
FIX Empty from clause when joining to same base table

dhensby:
see #3839 

This is a split from the above PR to address only the problem of empty FROM clauses in SQL statements when joining onto the originally queried table.


kinglozzer:
Dan and I have been working through this for a couple of days now. We’re not there yet, but this solves one of the main issues we’d both experienced. Tested this locally on a 3.1 and a 3.3 site and it solves the join/missing FROM problem

tractorcow:
Sorry to arrive late, but I think this particular fix is fine.

dhensby:
@tractorcow I merged up the branches when this went in 😀

I'll try to keep things out of 3.1

PR #5531, opened 2016-05-12
FIX ManyMany relations with shared inheritance

dhensby:
see #3839 

Self referencing relations fail

If you have a relationship between the same ancestor classes (eg: `SomePage` many-many `SomeOtherPage`) it will select base columns from `SiteTree`, try to join SiteTree again (but doesn't because of fix above) and then does a WHERE on SiteTree.

The generated SQL for `SomePage::get()->filter('RelatedPages', array(1,2,3));` is something like:

```sql
SELECT "SiteTree".*
FROM "SiteTree"
INNER JOIN "SomePage_RelatedPages" ON "SomePage_RelatedPages"."SomePageID" = "SiteTree"."ID"
WHERE "SiteTree"."ID" IN (1,2,3) AND "SiteTree"."ClassName" = 'SomePage';
```

This fails to return results as the join is not done properly and should be something closer to:

```sql
SELECT "SomePage".*
FROM "SiteTree" AS "SomePage"
INNER JOIN "SomePage_RelatedPages" ON "SomePage_RelatedPages"."SomePageID" = "SomePage"."ID"
INNER JOIN "SiteTree" AS "RelatedPage" ON "SomePage_RelatedPages"."RelatedPageID" = "RelatedPage"."ID"
WHERE "RelatedPage"."ID" IN (1,2,3) AND "SomePage"."ClassName" = 'SomePage'
```

This is a difficult problem to solve as the `SearchFilter` needs knowledge that the relation has a common inheritance and the where clause will have been aliased out.


model/DataQuery.php
```diff
@@ -13,6 +13,8 @@
  */
 class DataQuery {
 
+	public $lastAlias;
```
kinglozzer: Did you look into storing the alias with `$this->setQueryParam()`? Would mean we don’t need to introduce new things to the public API

dhensby: I haven't looked at that yet, but it would be an interesting alternative

model/DataQuery.php
```diff
@@ -661,6 +664,15 @@ public function applyRelation($relation) {
...
 				}
+				else {
+					$alias = uniqid($componentBaseClass);
+					$this->query->addLeftJoin(
+						$componentBaseClass,
+						"\"$relationTable\".\"$componentField\" = \"$alias\".\"ID\"",
+						$alias
+					);
+					$this->lastAlias = $alias;
+				}
```
kinglozzer: This code to use an alias is in the `else` block, so what we’re saying is if we’re already joined to this table, still perform the join again buy instead use an alias.

I wondered earlier if we could remove the `if (isjoinedto)` statement above this and instead check whether there’s going to be a “duplicate” join, then rename the fields appropriately? I’ve not checked if we have enough info to do that, just thinking aloud :)

tests/model/DataObjectTest.php
```diff
@@ -1532,5 +1541,23 @@ class DataObjectTest_ExtendedTeamComment extends DataObjectTest_TeamComment {
...
+
+class DataObjectTest_B extends DataObjectTest_A {
+	private static $many_many = array(
+		'ManyCs' => 'DataObjectTest_C',
+	);
+}
+
+class DataObjectTest_C extends DataObjectTest_A {
+	private static $belongs_many_many = array(
+		'ManyBs' => 'DataObjectTest_C',
```
kinglozzer: Shouldn’t this be 'ManyBs' => 'DataObjectTest_B'?

dhensby:
Just to clarify - my patch is not compliant with semver and is literally a proof of concept that the query builder can do this.

We need to look into a more approriate way of doing this.

In the log term, I'm not sure that the `dataQuery` class should be so involved in working out the relations and tables and this seems like something the `RelationList`s could do... Somehow, aliased tables need to get to the `SearchFilter` so it can apply conditions on the appropriate tables/aliases

model/DataQuery.php
dhensby: We need to distinguish between already existing joins and whether or not we need a new join because we are joining to a new distinct relationship...

Hard to do this as we need overall context of the entire filter for that :/

tests/model/DataObjectTest.php
dhensby: indeed, looks like it should

model/DataQuery.php
```diff
@@ -661,6 +664,15 @@ public function applyRelation($relation) {
 					$this->query->addLeftJoin($componentBaseClass,
 						"\"$relationTable\".\"$componentField\" = \"$componentBaseClass\".\"ID\"");
 				}
+				else {
+					$alias = uniqid($componentBaseClass);
```
tractorcow: How about the alias being the specific `$relation` up to this point? E.g.

```php
$dataquery->applyRelation(‘Images.Banner.Page’);
```

Will have a table joining the Image, Banner, and Page tables, with aliases Images, Images_Banner, and Images_Banner_Page

model/DataQuery.php
tractorcow: The `isJoinedTo` is going to need to be replaced with something that checks not only the class, but also the alias of the class is correct. Otherwise, cyclic joins aren't going to work. `->applyRelation('Page.Parent.Parent`)`

search/filters/ExactMatchFilter.php
```diff
@@ -50,6 +50,7 @@ protected function applyOne(DataQuery $query) {
 	 */
 	protected function applyMany(DataQuery $query) {
 		$this->model = $query->applyRelation($this->relation);
+		$column = $query->lastAlias ? sprintf('"%s"."%s"', $query->lastAlias, $this->name) : $this->getDbName();
```
tractorcow: I think lastAlias should be returned from applyRelation, since it's specific to that method call, not necessarily to that object, so it shouldn't be a member.

It could either be a part of the return value (array of data including class), or a by-ref parameter.

model/DataQuery.php
tractorcow: Query parameters aren't really the right place for it either. See my comment below. :D

model/DataQuery.php
tractorcow: Tracking isJoinedTo via alias and having predictable aliases could fix this. :)

search/filters/ExactMatchFilter.php
tractorcow: Ok, how about this….

Since one class can have multiple aliases, but not the other way around, applyRelation returns the relation name alias and you can GET the class by $dataQuery->relationClass($alias); 

This would have to be done in master though, and in 3.x we could stick with the property (lastAlias, although maybe stick it in a getter?) and mark it as deprecated in 4.

tractorcow:
@dhensby what do you think we should do with this issue? We can either add a minimal fix to 3.1 (that doesn't introduce any new api, and may have limitations) or we could look at another more complete solution in the 3 branch. I'm not sure that the direction we are heading can really stay open as a 3.1 pull request, so should we close it for the mean time?

I think you've made good progress, but it's a fix that would be better put on the 3 branch. :)

dhensby:
I don't see how we can do this in 3.1 without violating semver... 

I'll look at opening something against 3

sminnee:
Since this PR is  against 3.1, I'll close it for now.

I'd also say that we should fix this in the best way we can, and if that requires moving it to a `master` change I think that'd be better than doing a hackier change that meets minor-release requirements. In the long run, ensuring the ORM remains maintainable is the most important concern.

PR #5534, opened 2016-05-13
Patricknelson issue 5410 error tab focus

kinglozzer:
See https://github.com/silverstripe/silverstripe-framework/pull/5412

### DIFF
--- a/admin/javascript/LeftAndMain.EditForm.js
+++ b/admin/javascript/LeftAndMain.EditForm.js
@@ -89,7 +89,34 @@
 					// Ensure the first validation error is visible
 					var tabError = this.find('.message.validation, .message.required').first().closest('.tab');
 					$('.cms-container').clearCurrentTabState(); // clear state to avoid override later on
-					tabError.closest('.ss-tabset').tabs('option', 'active', tabError.index('.tab'));
+
+					// Selects the discovered tab set (since we'll need to look a few times).
+					var selectErroredTab = function($tabSet) {
+						$tabSet.tabs('option', 'active', tabError.index('.tab'));
+					};
+
+					// Attempt #1: Look for nearest .ss-tabset (usually nested deeper underneath a .cms-tabset).
+					var $tabSet = tabError.closest('.ss-tabset');
+					if ($tabSet.length) {
+						selectErroredTab($tabSet);
+					} else {
+						// TODO: Crappy effort to get this to execute later so we can check for the higher level tab instance that may help us select this error.
+						// TODO: This is necessary since this .onadd() method/event will fire before the higher level .cms-tabset is actually initialized.
+						setTimeout(function() {
+							// Attempt #2: Next level in tab-ception, try to select the tab within this higher level .cms-tabset if possible.
+							$tabSet = tabError.closest('.cms-tabset');
+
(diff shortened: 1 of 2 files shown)


kinglozzer:
@patricknelson Your changes included a check to make sure that no `$('.cms-panel-link')` elements were present (as those trigger navigation), but I couldn’t recreate any scenario in which they caused any problems. Do you remember how that change came about?

patricknelson:
Sure. If the failover to `tabError.closest('.cms-tabset')` ends up needing to run, it's possible that (for some reason or another) the error message itself may not exist inside of a selectable tab set (i.e. these tight bindings mean the classes could possibly change and selectors would fail). This is just in case it ends up failing to properly find the tab set (i.e. future breaking change?) and cause some wacky/magical bug in which a user submits a form and some how it navigates them to the history tab and/or settings tab unpredictably :joy:

admin/javascript/LeftAndMain.EditForm.js
```diff
@@ -77,23 +85,44 @@
...
-				if(this.hasClass('validationerror')) {
-					// Ensure the first validation error is visible
-					var tabError = this.find('.message.validation, .message.required').first().closest('.tab');
-					$('.cms-container').clearCurrentTabState(); // clear state to avoid override later on
-					tabError.closest('.ss-tabset').tabs('option', 'active', tabError.index('.tab'));
-				}
 
 				this._super();
 			},
+			'from .cms-tabset': {
```
patricknelson: Is this sort of nesting of event listeners on *other* selectors some sort of undocumented functionality? How does this work? Is `from` here some sort of keyword?

kinglozzer: I think you already figured it out in #5412, but in case anyone else wonders (there aren’t any docs for this functionality!) - every time the `afterredrawtabs` event is triggered on `.cms-tabset`, we run the `onafterredrawtabs` event handler in the context of `.cms-edit-form`.

It’s essentially equivalent to putting all of that logic in a function named `handleValidationError` and calling `$('.cms-edit-form').handleValidationError()` at the end of the tab redraw logic. Obviously if you wanted 15 different functions from 15 different selectors to trigger after each tab redraw, that would get pretty ugly and hard to maintain, so triggering an event that other selectors can bind to is much cleaner :)

patricknelson: Interesting and I suspected as much. One thing I'm still unsure of: 

Can you decompose the meaning of this object literal key, for me? What significance does the `from` have here in conjunction with the (apparent) `.cms-tabset` selector? Or is this key just a selector? The context of it makes sense to me if:
- It executes with the `this` of any closure defined as a value within this object literal bound to the outer selected `.cms-edit-form`, similar to the above key `onadd`
- The `from` Prefix is magic, similar to `on` since typically definitions here are methods. Instead, `on` converts the closure to an event listener and `from` contorts the context of child keys, with context then reverting back. 

Is that correct! Seems useful... But complicated to explain, no wonder it's undocumented 😂

kinglozzer: > `on` converts the closure to an event listener and `from` contorts the context of child keys, with context then reverting back once you dive into the value again (the closure).

Yep, that’s pretty-much it!

Issue #5487, opened 2016-05-09
Replace oEmbed implementation with 3rd party library


sminnee:
Follow-on from #5461.

There's no need for us to maintain our own oEmbed implementation. Let's move to a 3rd party package for this. https://github.com/oscarotero/Embed looks like a decent implementation. 

https://github.com/essence/essence is another reasonably popular one, although it has more external dependencies, isn't commit to as frequently, and isn't as popular on packagist, so the oscarotero one gets my vote.

Firesphere:
Hmmm, to bad I can not assign myself to this. I've started stripping out oEmbed in favor of oscarotero now, will possibly create a separate branch for essence if wished.

Firesphere:
oscarotero embed requires Guzzle 5.
From the looks of it, essence is slightly less complex. Any preference?
@silverstripe/core-team pong

kinglozzer:
> oscarotero embed requires Guzzle 5

It’s optional - you can [configure how the requests are resolved](https://github.com/oscarotero/Embed#the-request-resolver) (it’s only in `require-dev` in composer.json, not `require`).

+1 for https://github.com/oscarotero/Embed

kinglozzer:
Also, allow me to ping @silverstripe/core-team for you @Firesphere :D

Firesphere:
Pong. Thanks @kinglozzer :D

dhensby:
@kinglozzer is correct, Guzzle is just suggested, not required.

sminnee:
+1 for oscarotero/Embed without Guzzle. essence only looks less complicated because it's spread across 3 packages, I expect ;-)

Firesphere:
I admit I only had a quick look at it. Will go with oscarotero then. On functionality, I doubt it'll be very much different.

Firesphere:
While we're at it...
What about the shortcode parser? I'd think using /var/www/vhosts/TB01-007/2016.stripecon.eu/site/docroot would be a nice way to get those kind of dragging-on things out of the way

dhensby:
Yes please, lets pull all this stuff out :)

Lots of our implementations of these features are quite MVP and have bugs that just never get round to being fixed.

sminnee:
I think that solving the shortcode parser as part of this PR isn't the best idea. Get oEmbed extracted, the PR raised (and mereged), and then move on to shortcodes.

dhensby:
Yes, different pr

Firesphere:
[...]

PR #5536, opened 2016-05-15
RFC #5487 remove oembed

Firesphere:
I only had to move the frontend embedding to ShortcodeParser.
I decided to go with it hardcoded, as the next step is remove the ShortcodeParser.

In the docs, also included the removal of RestfulService (slightly off topic, but it was the best way since I forgot to do it on the original PR where RestfulService was removed)


sminnee:
This isn't quite correct.

oEmbed should still be a mandatory feature, but it should rely on the third party module.

We didn't want to drop the oEmbed feature, just to use someone else's work to power it.

_config.php
```diff
@@ -19,7 +19,6 @@
 
 ShortcodeParser::get('default')
 	->register('file_link', array('File', 'handle_shortcode'))
-	->register('embed', array('Oembed', 'handle_shortcode'))
```
sminnee: We'll need to have a shortcode handler that handles the bridge to the 3rd party module.

_config/Oembed.yml
```diff
@@ -1,40 +0,0 @@
-name: Oembed
----
-Oembed:
```
sminnee: Presumably the 3rd party module comes pre-loaded with this configuration? If not, we'd want to pass it in.

parsers/ShortcodeParser.php
```diff
@@ -21,7 +23,12 @@ public function img_shortcode($attrs) {
 
 	// --------------------------------------------------------------------------------------------------------------
 
-	protected $shortcodes = array();
+	protected $shortcodes = array(
+		'embed' => array(
```
sminnee: This should be provided in the config, not hard-coded into ShortcodeParser.php.

parsers/ShortcodeParser.php
```diff
@@ -612,4 +619,51 @@ function ($matches) use ($tags, $parser) {
...
+	 *
+	 * @param $arguments
+	 * @param $content
+	 * @param $parser
+	 * @param $shortcode
+	 * @param array $extra
+	 *
+	 * @return string
+	 */
+	public static function handle_embed_shortcode($arguments, $content, $parser, $shortcode, $extra = array()) {
```
sminnee: This should be in its own class, e.g. `EmbedShortcodeProvider`. Don't pollute the ShortcodeParser core with oEmbed-specific things.

tests/oembed/OembedTest.php
```diff
@@ -1,153 +0,0 @@
-<?php
```
sminnee: A test for our shortcode provider would be good.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -278,6 +278,14 @@ E.g.
...
         'mysite/js/src/functions.js'
     ]]);
+    
+### RestfulService
+
+* `RestfulService` has been removed. Use Guzzle instead. See Upgrading notes.
+
+### Oembed
+
+* `Oembed` has been removed from the framework in favor of [oscarotero/Embed](https://github.com/oscarotero/Embed)
```
sminnee: I'd make it clearer that oEmbed feature hasn't been dropped, just our code: "Our self-maintained Oembed implementation has been removed, in favour of introducing oscarotero/Embed as a dependency."

sminnee:
Cool, I initially interpreted this PR incorrectly, thinking it dropped oEmbed from core. My main feedback is that the the shortcode handler belongs in its own class.

sminnee:
> I decided to go with it hardcoded, as the next step is remove the ShortcodeParser.

Putting oEmbed-specific code will make it *harder* to swap our ShortcodeParser implementations, which is why I have suggested putting it in another class.

tests/oembed/OembedTest.php
Firesphere: This is the OembedTest, not the ShortcodeParser test.

parsers/ShortcodeParser.php
Firesphere: Same as last remark... but will fix it.

tests/oembed/OembedTest.php
sminnee: OK, then create `EmbedShortcodeProvider` and `EmbedShortcodeProviderTest` ;)

Firesphere:
> Putting oEmbed-specific code will make it harder to swap our ShortcodeParser implementations, which is why I have suggested putting it in another class.

I went for this quick option, because of removing our own shortcode parser was/is literally the next on my to-do list, so I didn't really care enough about it, but agreed, even then, taking shortcuts is bad and I should feel bad about it.

tests/oembed/OembedTest.php
Firesphere: Jep, figured that ;)

sminnee:
We'll still be using shortcodes in some form in the next step, so the `EmbedShortcodeProvider` will be useful there, in some form.

Firesphere:
You want separate providers for the separate shortcodes we're going to support ourselves? Or a generic SilverStripeShortcodeProvider-like method?

sminnee:
For now I'd just do this:

    class EmbedShortcodeProvider {
      public static handleShortcode($arguments, $content, $parser, $shortcode, $extra = array()) { ... }
      public static function embedForTemplate($embed) { ... }
    }

And in _config.php:

    ShortcodeParser::get('default')->register('embed', array('EmbedShortcodeProvider', 'handleShortcode'))

We can decide a better approach once we see what other shortcode engines expect.

Firesphere:
Yep, cool. I think I can push that through tonight.

Firesphere:
Updated. Implemented most feedback. 

*note:
I'll squash into single commit once it's good to merge. For now, I want to keep it on the step-by-step so the changes are better visible.

dhensby:
Looks like something funky is going on here...

sminnee:
Can you please rebase this branch, @Firesphere? `EmbedShortcodeProvider` and its test look pretty good. I think squashing now will be fine.

Firesphere:
This is the rebased branch, something funky happened with rebasing, breaking a totally unrelated test it seems?

dhensby:
Have you recessed on a stale branch or the wrong branch, there are a lot of unrelated commits here.

Firesphere:
Re-re-re-based. This time it should be rebased correctly. I might've accidentally pulled in the wrong master branch, as there are a lot of remotes.

Firesphere:
Error-problem discovered. My attempt to move the config to yml accidentally got pushed.

parsers/ShortcodeParser.php
```diff
@@ -612,4 +620,10 @@ function ($matches) use ($tags, $parser) {
 
 		return $content;
 	}
+
+	public static function getShortcodes($instance = 'default')
```
sminnee: Do we need this method? If not, I would remove it. I think we want to keep the number of static accessors to a minimum as they make it easy to increase coupling.

sminnee:
This is looking really good. I just had 1 more query, commented above.

Has anyone else had a look at it?

parsers/ShortcodeParser.php
Firesphere: Oops.
I added that method for debugging what the hell was going on with HTMLEditorFieldTest failing. Forgot to remove it. Indeed we don't need it at all.

dhensby:
:)

Firesphere:
Also, Happy Birthday @dhensby

dhensby:
Nawwwww. Thanks!

dhensby:
Please can we remove the tests that depend on external sources. This isn't an integration test, it's a unit test. Expected input -> expected output.

if youtube or anyone else changes their responses that doesn't make our code broken it means they're breaking standards/sites.

PR #5538, opened 2016-05-15
Fix bug, export button cannot export value '0'

jallen0927:


sminnee:
Can we have a unit test for this case, please?

jallen0927:
Thanks @sminnee , case for unit test added.

dhensby:
thanks

PR #5540, opened 2016-05-16
BUG Fix singleton('DBLocale')

tractorcow:
With a mandatory first argument DBLocale cannot be created as a singleton, and breaks certain DataObject operations which do reflection on field types.

This seems to appear in translatable running against master, but the bug is in all versions of framework.


sminnee:
Seems okay as a bugfix to me.

PR #5541, opened 2016-05-16
Revert "Disable subsite filters when running unit tests"

tractorcow:
Reverts silverstripe/silverstripe-framework#5480

Fixes some (but not all) CI failures on subsites module.


Firesphere:
I think it's a bit strange, to have the core take care of this. I'd say, the module needs to change something for this.
Even if it's a SilverStripe supported module. Core should not have to contain workarounds for whether a module is installed or not.

dhensby:
Hi team, my bad on merging. I thought this would likely happen :/

@firesphere I agree but as core already does this, I decided it was acceptable. 

We need some kinships pre test hook that allows modules to disable their default settings when system is under test...

tractorcow:
What we need is an events system, so that modules can do specific setup actions during testing.

PR #5545, opened 2016-05-17
BUG Fix DataObject::isChanged() detecting non saveble changes

tractorcow:
When certain properties are saved against a dataobject (e.g. via dependency injection) it often has the effect of incorrectly marking the object as dirty (even if those properties aren't database fields).

This change also removes a lot of duplicate code from updateChanges() that really isn't necessary. If $forceChanges is true, change detection isn't necessary, and updating changes can be done by simply piggy backing off isChanged() which invokes getChangedFields().

If users want to get all changed fields, even if they aren't database fields (which is the exception rather than the rule) they can still use getChangedFields(true) to get all non-db fields also.

This is against 3.x since it is a slight change in behaviour, and has a risk of regression.

I'm not sure if this should go into 3.4.0 or 3.5.0, due to the risk.

### DIFF
--- a/docs/en/04_Changelogs/3.4.0.md
+++ b/docs/en/04_Changelogs/3.4.0.md
@@ -14,3 +14,5 @@ was affected by these:
  * When FormFields are rendered, leading & trailing whitespace is now stripped. The resulting HTML for form fields is
    the same for the default fields, but if you have a custom form field that is relying on trailing whitespace being
    outputted.
+ * DataObject::isChanged() now defaults to only checking database fields. If you rely on this method
+   for checking changes to non-db field properties, use getChangedFields() instead.
--- a/tests/model/DataObjectTest.php
+++ b/tests/model/DataObjectTest.php
@@ -533,29 +533,29 @@ class DataObjectTest extends SapphireTest {
 		$obj->IsRetired = true;
 
 		$this->assertEquals(
-			$obj->getChangedFields(false, 1),
+			$obj->getChangedFields(true, DataObject::CHANGE_STRICT),
 			array(
 				'FirstName' => array(
 					'before' => 'Captain',
 					'after' => 'Captain-changed',
-					'level' => 2
+					'level' => DataObject::CHANGE_VALUE
 				),
 				'IsRetired' => array(
 					'before' => 1,
 					'after' => true,
-					'level' => 1
+					'level' => DataObject::CHANGE_STRICT
 				)
 			),
 			'Changed fields are correctly detected with strict type changes (level=1)'
 		);
 
 		$this->assertEquals(
-			$obj->getChangedFields(false, 2),
+			$obj->getChangedFields(true, DataObject::CHANGE_VALUE),
 			array(
 				'FirstName' => array(
 					'before'=>'Captain',
 					'after'=>'Captain-changed',
-					'level' => 2
+					'level' => DataObject::CHANGE_VALUE
 				)
 			),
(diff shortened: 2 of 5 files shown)


tractorcow:
I'm going to wait for https://github.com/silverstripe/silverstripe-framework/pull/5546 to get merged, then I'll rebase and resolve test issues.

sminnee:
I've merged #5546 now :-)

sminnee:
`DataObject::isChanged()` is supposed to show if things have changed since loading from the database, so returning true on changes to non-database-persisted fields is a bug and is fine for 3.x. I would suggest that we aim to get this into 3.4.

tractorcow:
Local testing detects failures due to lots of dataobject bootstrapping issues. It looks like sapphire test is no longer triggering `onBeforeWrite` on objects created from fixtures, so random things are breaking.

It's possible that the "don't write if not needed" logic has been failing up until now, but we have been relying on that failing. :)

tractorcow:
For instance, in CMSProfileControllerTest, why is $member->Locale blank, when it should be set by Member::onBeforeWrite?

tractorcow:
Oops, the old code that I removed (rightly) treated all fields as changed is `$this->IsInDB()` was false. The code in `getChangedFields` didn't have the same check. :D

tractorcow:
After much hair pulling, I think I got it to work. :)

Passing in $isNewRecord instead of $forceInsert to `updateChanges` means that new records will properly get all records initialised as changed.

tractorcow:
Remaining bug is when modifying a composite field, it's not marking the sub-fields as changed.

Issue to fix another day... this is OK to not put into 3.4.0

tractorcow:
Yay fixed!

docs/en/04_Changelogs/3.4.0.md
```diff
@@ -14,3 +14,6 @@ was affected by these:
  * When FormFields are rendered, leading & trailing whitespace is now stripped. The resulting HTML for form fields is
    the same for the default fields, but if you have a custom form field that is relying on trailing whitespace being
    outputted.
+ * DataObject::isChanged() now defaults to only checking database fields. If you rely on this method
+   for checking changes to non-db field properties, use getChangedFields(true) instead. This will
```
sminnee: Is this supposed to be `getChangedField(false)`?

tests/model/DataObjectTest.php
```diff
@@ -564,50 +564,58 @@ public function testChangedFields() {
...
 				)
 			),
+			$newObj->getChangedFields(true, DataObject::CHANGE_VALUE),
 			'Initialised fields are correctly detected as full changes'
 		);
 	}
 
 	public function testIsChanged() {
 		$obj = $this->objFromFixture('DataObjectTest_Player', 'captain1');
+		$obj->NonDBField = 'bob';
```
sminnee: I think we should test that changes to composite fields are still detected as `isChanged()`.

sminnee:
OK, a few small things but other than that this looks good!

docs/en/04_Changelogs/3.4.0.md
tractorcow: Oops, yes. :D

tests/model/DataObjectTest.php
tractorcow: The test related to this was MoneyTest. The fix was to fix regressions, not a new behaviour.

I can still add a test though to this part.

tractorcow:
isChanged tests added to MoneyTest to test composite DB fields. :)

PR #5546, opened 2016-05-17
BUG Fix filtersOnId ignoring `WHERE "ID" IN ()`

tractorcow:
Fixes bug in subsites where it would apply subsites filter to queries on `ID IN ()`, which should not be subsite filterable.


sminnee:
I'm happy with this. Will merge once the tests pass.

PR #5551, opened 2016-05-17
Updated 01_Data_Model_and_ORM.md for byID

AntonyThorpe:
According to the Framework's API documentation 'byID' is a method of DataList not 'byId'.  Both are used in the documentation.

Actually, both 'byId' and 'byID' methods are used by the Framework and CMS so please correct me if I am wrong about this.


dhensby:
Thanks!

PR #5555, opened 2016-05-18
BUG Fix suppression of display_errors in ErrorControlChain

### LINKED ISSUE #5323: BUG ini_get in ErrorControlChain isn't properly parsing display_errors
See https://github.com/silverstripe/silverstripe-framework/pull/5256#discussion_r59819635

This works fine when set to `0/1` but not `true/false` or `on/off`

tractorcow:
Fixes #5323


core/startup/ErrorControlChain.php
```diff
@@ -46,14 +46,36 @@ public function setErrored($error) {
...
-		// originally allowed to be displayed.
-		if ($this->handleFatalErrors && $this->originalDisplayErrors) {
-			ini_set('display_errors', !$suppression);
+		// If handling fatal errors, conditionally disable, or restore error display
+		// Note: original value of display_errors could also evaluate to "off"
+		if ($this->handleFatalErrors) {
+			if($suppression) {
+				$this->setDisplayErrors(0);
+			} else {
+				$this->setDisplayErrors($this->originalDisplayErrors);
```
halkyon: Could there be a case where this now sets `null` for display_errors since `originalDisplayErrors` wasn't defined?

tractorcow: If that's the case, wouldn't restoring null have the correct behaviour?

PR #5556, opened 2016-05-18
API Enable Debug.friendly_error_httpcode to correctly set HTTP status code for errors

tractorcow:
Relies on https://github.com/silverstripe/silverstripe-framework/pull/5555 to correctly suppress errors from pre-emptively setting "headers_sent" to true.

### DIFF
--- a/dev/Debug.php
+++ b/dev/Debug.php
@@ -41,6 +41,17 @@ class Debug {
 	private static $friendly_error_header = 'There has been an error';
 
 	/**
+	 * Set to true to enable friendly errors to set a http response code corresponding to the error.
+	 * If left false then error pages will be served as HTTP 200.
+	 *
+	 * Will be removed in 4.0, and fixed to on.
+	 *
+	 * @config
+	 * @var bool
+	 */
+	private static $friendly_error_httpcode = false;
+
+	/**
 	 * @config
 	 * @var string The body of the message shown to users on the live site when a fatal error occurs.
 	 */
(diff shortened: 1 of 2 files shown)


dev/Debug.php
```diff
@@ -41,6 +41,17 @@ class Debug {
...
 	/**
+	 * Set to true to enable friendly errors to set a http response code corresponding to the error.
+	 * If left false then error pages will be served as HTTP 200.
+	 *
+	 * Will be removed in 4.0, and fixed to on.
+	 *
+	 * @config
+	 * @var bool
+	 */
+	private static $friendly_error_httpcode = false;
```
halkyon: I would argue this shouldn't even need to be an option, since it's effectively fixing a bug. What do others think?

tractorcow: The risk of regressions is extremely high were it to be fixed to true already. It could potentially interfere with a lot of existing infrastructural, so a phased approach (opt in by setting a config) is likely the least risky.

@hafriedlander can you please weigh in on this?

hafriedlander:
Yeah, although it's "wrong", the current behaviour might be relied on. It doesn't fail, it just does something incorrect. So fixing it might break existing sites, which would be bad.

dev/Debug.php
hafriedlander: Yeah, although it's "wrong", the current behaviour might be relied on. It doesn't fail, it just does something incorrect. So fixing it might break existing sites, which would be bad.

hafriedlander: We could also setup silverstripe-installer to set it to true so new installations get true but existing sites get false (since people rarely update silverstripe-installer). Bit tricky though, and probably breaks SemVer

tractorcow: Yep, I'll cross-post to installer for 3.4.0 release.

tractorcow:
See https://github.com/silverstripe/silverstripe-installer/pull/124

dev/Debug.php
tractorcow: https://github.com/silverstripe/silverstripe-installer/pull/124

tractorcow:
Thank you @hafriedlander !

PR #5559, opened 2016-05-18
FIX Remove old logout icon on retina screens

dhensby:
Remove background rules for old preview icons

Reapply fix to 3 branch that some how got dropped but is in 3.3 and master


kinglozzer:
Thanks @dhensby!

Issue #5562, opened 2016-05-19
Some ClassInfo methods return null instead of an array in certain cases


micschk:
According to their docblocks(?), some of these methods are expected to return an array instead of null. This seems to have caused issues in 3.2 & 3.3, where SiteTree::allowedChildren() tries to loop over null...
- getValidSubClasses
- dataClassesFor
- subclassesFor
- ancestry

Introduced here: https://github.com/silverstripe/silverstripe-framework/commit/1f0602d42fd9e1c0a4268f3a51aa7f483100a935#commitcomment-17538115

Issue #5171, opened 2016-03-12
[3.3.1] allowed children doesn't work properly


t3hn0:
$allowed_children is not respected anymore. If you select "add new page" above site structure in Pages, you get the list of all page types (except the ones which return false in canCreate()).

You have two scenarios:
1) If you select one of allowed pages it works.
2) If you select any other pages which should been disabled, you get an error:
Fatal error: Call to a member function forTemplate() on a non-object in ***/framework/admin/code/LeftAndMain.php on line 593

Btw, filter for allowed children works in context menu.

tractorcow:
So you're saying children don't respect their parents anymore? I'm sure I've heard that elsewhere...

It looks like this shouldn't be too hard to test or debug. I'll try to find some time this week.

Have you tested this in 3.3.0, or 3.2 and found if the issue exists there as well?

t3hn0:
Yes it doesn't respect parent.

I'll try reproducing error in 3.2 tomorrow.

t3hn0:
So I've tested it on SS 3.2.1 and the same problem is present there too.

dhensby:
Ping

tractorcow:
I said I'd do this just before I was off sick for a week... now I don't have time. :D

leclerc:
Having 'none' => 'none' in  $allowed_children  array gave me an error  Invalid argument supplied for foreach() in /cms/code/model/SiteTree.php on line 2692

micschk:
Problem seems to be ClassInfo::subclassesFor() may return NULL instead of an empty array.

A quick fix would be to replace a non-array value with an empty array just before looping over it (SiteTree.php line 2692)
```
if(!is_array($subclasses)) $subclasses = array();
```
The allowedchildren method then becomes
```
public function allowedChildren() {
		$allowedChildren = array();
		$candidates = $this->stat('allowed_children');
		if($candidates && $candidates != "none" && $candidates != "SiteTree_root") {
			foreach($candidates as $candidate) {
				// If a classname is prefixed by "*", such as "*Page", then only that class is allowed - no subclasses.
				// Otherwise, the class and all its subclasses are allowed.
				if(substr($candidate,0,1) == '*') {
					$allowedChildren[] = substr($candidate,1);
				} else {
					$subclasses = ClassInfo::subclassesFor($candidate);
					if(!is_array($subclasses)) $subclasses = array();
					foreach($subclasses as $subclass) {
[...]

PR #5563, opened 2016-05-19
Update ClassInfo to return arrays where expected

micschk:
These methods are expected to return an array instead of null (if $class doesnt exist);
- getValidSubClasses
- dataClassesFor
- subclassesFor
- ancestry

Fixes: #5562
Also see: https://github.com/silverstripe/silverstripe-framework/issues/5171


dhensby:
Thanks

tractorcow:
I would also have accepted exceptions for these, but `return array()` is fine.

PR #5566, opened 2016-05-19
BUGFIX: check token and UID before member autologin

### LINKED ISSUE #5565: autoLogin fails with a non-numeric ID
```php
		list($uid, $token) = explode(':', Cookie::get('alc_enc'), 2);

		$member = DataObject::get_by_id("Member", $uid);
```

This call fails in DataObject if the id is not numeric

```php
		if(!is_numeric($id)) {
			user_error("DataObject::get_by_id passed a non-numeric ID #$id", E_USER_WARNING);
		}
```

stevie-mayhew:
Don't assume that the cookie is correct before trying to log in a member

Fixes #5565

### DIFF
--- a/security/Member.php
+++ b/security/Member.php
@@ -582,6 +582,10 @@ class Member extends DataObject implements TemplateGlobalProvider {
 
 		list($uid, $token) = explode(':', Cookie::get('alc_enc'), 2);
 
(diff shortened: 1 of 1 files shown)


dhensby:
thanks

Issue #5549, opened 2016-05-17
ClassManifest / TokenisedRegularExpression appears to have memory leaks


sminnee:
A frequent source of frustration when setting up new dev environments is that visiting the sites will run out of memory while compiling the initial manifest. This is exacerbated by the fact that the manifest is rebuild for www-data and CLI users, but that's a separate problem.

I believe that the issue is called when the file-by-file parsing needs to happen for all files in the project. This suggests that a memory leak in this subsystem is responsible for the problem.

dhensby:
pull `TokenisedRegularExpression ` out for https://github.com/nikic/PHP-Parser :)

dhensby:
@kinglozzer took a quick look at doing it here: https://github.com/silverstripe/silverstripe-framework/compare/master...kinglozzer:classmanifest

but it takes too long and uses too much memory, I think but the feature of breaking out of the node traverser hasn't been implemented, which I think we can use aggressively to save time (and maybe memory?)

sminnee:
It will be interesting to see if an approach using nikic's parser can be made to work, but until the speed and memory usage issues are solved, it would make this problem worse, not better. It wouldn't surprise me if there were some quick wins with the current codebase and it's probably worth addressing those.

Bringing in nikic/php-parser feels a bit like using a sledgehammer to crack a walnut, but it would be a more robust project to rely on.

dhensby:
It reduces the coffee we have to maintain, but yes it's got issues we'd like to address. 

Copying composer's approach to class finding may be the way to speed this up. We don't need to tokenise more than a few lines of a file to get the info we need so we can look at trashing a ton of the coffee before tokenisation

sminnee:
Composer's approach is inferior and not suited to our use case. It interprets neither the implements not extends keywords. It's a non-starter.

Trying to reliably tokenise a partial file is hard to do reliably, and the code basically starts by scanning for a T_CLASS or T_INTERFACE item.

We could consider moving to a better parser but we're not moving to a worse one.

dhensby:
> Composer's approach is inferior and not suited to our use case. It interprets neither the implements not extends keywords. It's a non-starter.

Sorry, by copying composers approach, I mean aggressively discarding code that we don't need to parse. @kinglozzer has had a go at it and had some success.

> We could consider moving to a better parser but we're not moving to a worse one.

Agreed, but can we get worse than what we've got? ;)

sminnee:
[...]

PR #5567, opened 2016-05-19
Replace TokenizedRegularExpression with PHPParser

dhensby:
see #5549

This is a PoC that we can remove our home baked static tokenizer for a library.

This reduces our code maintenance but at the cost of higher memory and cpu usage... which needs addressing

~~Depends on #5572 (needs merging first)~~


dhensby:
So @kinglozzer has compared this PR with current master and there's a fairly significant impact on time and memory:

image

kinglozzer:
Looks like there could be some memory overhead added by XHProf... using `memory_get_peak_usage(true)` reports 22mb for current master and 25mb for this PR. Still an increase, but around half of what XHProf was reporting.

Using `memory_get_peak_usage(true)` and `time` on CLI reports:

```
Current master:
22mb
real 0m2.620s
user 0m2.463s
sys 0m0.139s

This PR:
25.5mb
real 0m5.068s
user 0m4.843s
sys 0m0.193s
```

SpiritLevel:
@dhensby, @kinglozzer  Dan, Loz, is that a Github tool you are using to compare runs ? Where can I find it ?

dhensby:
No, we are running locally with CLI

dhensby:
OK, here is my bench marking:

```
Test results - tested on XS cloud server by atlantic.net

classmanifest branch:
100 runs of building brand new manifest	: 20m17.015s
100 runs of flushing manifest: 0m42.927s

current master:
100 runs of building brand new manifest	: 7m50.568s
100 runs of flushing manifest: 0m43.211s

diff:
Brand new: 12m26.447s = 7.46s per run
Flushing: -0.284s = negligible

Memory usage: (rm -rf silverstripe-cache/* && framwork/sake /)
classmanifest: 25.75 mb (27000832)
master: 22.25 mb (23330816)
diff: 3.5 mb (3670016)
```

We really need to find a way to optimise this because a 7.5s increase on building the manifest from scratch is not acceptable

dhensby:
Related: https://github.com/nikic/PHP-Parser/issues/278

sminnee:
I'd be surprised if 278 gets results, but it's worth a shot.

In the meantime let's stick with TRE and making minor amendments to that.

tractorcow:
We also have a lot of logic for tokenisation in Object::parse_class_spec. I'd like to, at the very least, see some kind of consolidation for this logic so we don't have custom baked parsers sitting around the code base.

sminnee:
> We also have a lot of logic for tokenisation in Object::parse_class_spec. I'd like to, at the very least, see some kind of consolidation for this logic so we don't have custom baked parsers sitting around the code base.

It's quite a different parsing challenge, although it would probably benefit from using the built-in tokeniser.

sminnee:
@dhensby Given that `php-parser` said that your suggested change wasn't possible; do you think that we're likely to get this to be faster than the current implementation?

dhensby:
OK - so #6615 (and my solution #6619) go to show how hand-rolling our own file parser can cause us maintenance problems. I've had to fudge my fix a little because I can't get it to work the way I'd like (removing references to `T_WHITESPACE` from the TRE definitions - I've had to just mark them as optional for now).

Whilst a 7s increase on initial build of the manifest is a lot, perhaps it's something that isn't sooooo bad if we consider it's only on a completely fresh manifest build (which is quite rare)?

The only other way I can think of getting this faster is working on making https://github.com/nikic/PHP-Parser faster by having it only instantiate objects when entering a syntax tree rather than instantiating the entire AST for a file and then performing the search over it. This is based on the assumption the the extra compute time comes from instantiating the token objects.

If we're only going to enter 1 of the syntax trees, then it's not worth the CPU cycles to instantiate all of them... I'd imagine this is quite a bit of work and would be a 3.x release of PHPParser, so not something that could happen any time soon...

dhensby:
In fact, one of @kinglozzer's original ideas was to only send the tokens through to PHPParser that we care about. so strip out the vast majority of tokens.

And, another point, we don't have 7.x PHP syntax support in our parser, something like this would not be recognised in SS: `use Microsoft\PhpParser\{DiagnosticsProvider, Node, Parser, PositionUtilities};`

dhensby:
OK - with this fix I'm seeing performance improvements over current master of ~1second

dhensby:
so we need to test the php7 style of import for some confidence in this.

src/Core/Manifest/NodeVisitor.php
```diff
@@ -0,0 +1,81 @@
+<?php
+
+namespace SilverStripe\Core\Manifest;
+
+use PhpParser\Node;
+use PhpParser\NodeTraverser;
+use PhpParser\NodeVisitorAbstract;
+
+class NodeVisitor extends NodeVisitorAbstract
```
sminnee: You can pick a better name than this.

tractorcow:
I think cleanFileContents needs some tests of its own, since it's custom code.

src/Core/Manifest/ClassManifest.php
```diff
@@ -644,36 +418,89 @@ public function handleFile($basename, $pathname, $depth)
...
+    }
+
+    /**
+     * Clean out any information from PHP files that is nested more than $cutOffDepth in curly braces
+     *
+     * @param string $contents
+     * @param int $cutOffDepth
+     * @return string
+     */
+    protected function cleanFileContents($contents, $cutOffDepth = 1)
```
sminnee: Add a test specifically for cleanFileContents. Related to that, maybe burying it as a protected method of this class isn't as good as just splitting it out into a separate class.

Maybe `ClassContentRemover` is a better name?

src/Core/Manifest/NodeVisitor.php
dhensby: How about `ManifestVistor`? I was thinking the NS provides enough context here (`SilverStripe\Core\Manifest\NodeVisitor`). I could rename it during import `use SilverStripe\Core\Manifest\NodeVisitor as ManifestNodeVisitor` for more clarity outside the class?

Any other suggestions welcome

dhensby:
Performance test results:

Current master (449ed8d379): 5m37.053s (3.37s avg)
This code: 3min 14 sec (1.94s avg)

~40% reduction in execution time

---

Test script:

```
#!/bin/bash

for i in {1..100}; do
        rm -rf silverstripe-cache/*
        framework/sake / > /dev/null
done
```

sminnee:
Regarding name: maybe ClassManifestVisitor?

dhensby:
OK - I think this is ready now.

I've re run the tests against latest master and can see a speed improvement of 400ms on average for this vs latest master.

src/Core/Manifest/ClassManifest.php
```diff
@@ -195,6 +93,35 @@ public function __construct($base, $includeTests = false, $forceRegen = false, $
         }
     }
 
+    public function getParser()
+    {
+        if (!$this->parser) {
+            $this->parser = (new ParserFactory)->create(ParserFactory::PREFER_PHP5);
```
kinglozzer: Any reason for `PREFER_PHP5` over 7? [PHP-Parser docs](https://github.com/nikic/PHP-Parser/blob/master/doc/2_Usage_of_basic_components.markdown#parsing) say:

> Unless you have strong reason to use something else, `PREFER_PHP7` is a reasonable default.

The PHP 7 grammar looks to have slightly fewer rules, is it possibly marginally more performant that way around?

src/Core/Manifest/ClassManifest.php
```diff
@@ -662,40 +435,42 @@ public function handleFile($basename, $pathname, $depth)
...
-        foreach ($traits as $trait) {
-            $this->traits[strtolower($namespaceBase . $trait['traitName'])] = $pathname;
+        foreach ($traits as $traitName => $traitInfo) {
+            $this->traits[strtolower($traitName)] = $pathname;
         }
     }
 
     /**
-     * Recursively coalesces direct child information into full descendant
+     * Recursively coalthat esces direct child information into full descendant
```
kinglozzer: Typo

dhensby:
OK - ready ready. we are now seeing average improvement on building empty caches of... 1.45 seconds vs current master

kinglozzer:
Here’s an XHProf comparison of current master vs this PR:

screen shot 2017-03-02 at 16 52 11

I don’t know if/how much these results are being skewed by https://github.com/silverstripe/silverstripe-framework/issues/6670

dhensby:
Ping @sminnee

tractorcow:
How well does this code with code such as:

```php
class SomeClass {
	function bob() {
		return '{' . '{' . '{';
	}
}

class AnotherClass {
}
```

dhensby:
It works great, test added as proof.

dhensby:
we have lift off! 🚀

PR #5569, opened 2016-05-19
Switched Diff to the escaped variable created

flamerohr:


dhensby:
Can I get a bit more context on this, please?

flamerohr:
I noticed the $fieldDiff variable was created with field value escaping in mind, but then it's ignored and the logic was duplicated without the value escaping.

dhensby:
So it is. Good spot! Thanks

PR #5572, opened 2016-05-20
Giving composer more responsibility over autoloading

dhensby:
Bring composer's autoloader into the fold as early as possible. This will allow us to stop manual class includes when we are fully up and running with PSR-4 compliant namespaces.

### DIFF
--- a/cli-script.php
+++ b/cli-script.php
@@ -58,6 +58,15 @@ if(isset($_SERVER['argv'][1])) {
 	$_GET['url'] = $_SERVER['argv'][1];
 }
 
+// require composers autoloader
+if (file_exists($autoloadPath = dirname(__DIR__) . '/vendor/autoload.php')) {
+	require_once $autoloadPath;
+}
+else  {
+	echo "Failed to include composer's autoloader, unable to continue\n";
+	exit(1);
+}
+
 /**
  * Include SilverStripe's core code
  */
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,10 @@
 		}
 	},
 	"autoload": {
+		"psr-4": {
+			"SilverStripe\\": ""
+		},
+		"files": ["core/Constants.php"],
 		"classmap": ["tests/behat/features/bootstrap"]
 	}
 }
--- a/core/Core.php
+++ b/core/Core.php
@@ -3,7 +3,6 @@
  * This file is the Framework bootstrap.  It will get your environment ready to call Director::direct().
  *
  * It takes care of:
- *  - Including Constants.php to include _ss_environment and initialise necessary constants
  *  - Checking of PHP memory limit
  *  - Including all the files needed to get the manifest built
  *  - Building and including the manifest
@@ -21,11 +20,6 @@
  * live mode, where reporting is limited to fatal errors and warnings (see later in this file)
  */
 error_reporting(E_ALL | E_STRICT);
-
-/**
- * Include Constants (if it hasn't already been included) to pull in BASE_PATH, etc
- */
-require_once dirname(__FILE__).'/Constants.php';
 
 global $_increase_time_limit_max;
 $_increase_time_limit_max = -1;
@@ -109,11 +103,6 @@ $manifest = new SS_ClassManifest(BASE_PATH, false, $flush);
 $loader = SS_ClassLoader::instance();
 $loader->registerAutoloader();
 $loader->pushManifest($manifest);
-
-// Fall back to Composer's autoloader (e.g. for PHPUnit), if composer is used
-if(file_exists(BASE_PATH . '/vendor/autoload.php')) {
-	require_once BASE_PATH . '/vendor/autoload.php';
-}
 
 // Now that the class manifest is up, load the static configuration
 $configManifest = new SS_ConfigStaticManifest();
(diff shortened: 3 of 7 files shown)


dhensby:
Behat currently fails I assume because it relies on `core/Core.php` bringing in composers autoloader. @chillu Where do we make the https://github.com/silverstripe-labs/silverstripe-behat-extension pull in the autoloader?

sminnee:
> Behat currently fails I assume because it relies on core/Core.php bringing in composers autoloader. Where do we make the https://github.com/silverstripe-labs/silverstripe-behat-extension pull in the autoloader?

Can you give some more details about what's broken here?

As an aside, I'm trying to get this working locally by checking out your framework branch, and I can't work out how to get the new autoloader meta-data pulled into the project. `composer dump-autoload` doesn't seem to do it...

dhensby:
I'll update the behat issue to provide more info. 

As for testing locally, how are you checkout out my branch? 

You'll need to make sure composer is checking it out by having a project composer file similar to: 

```json
{
    "reqiure": {
        "silverstripe/framework": "pulls/4/early-auto-loader-dev as 4.0.x-dev",
        ...
    },
    "repositories": [{
        "type": "vcs",
        "url": "https://githun.com/dhensby/silverstripe-framework"
    }],
    ...
}
```

dhensby:
I think you need to include my branch with composer (detailed above) as opposed to manual checkout because composer may use the lock file instead of the current composer.json file.

dhensby:
I think as a stopgap we can keep the autoloader include in `Core.php` and aim to remove it later?

dhensby:
I think the install script will need the autoloader required too

sminnee:
> I think the install script will need the autoloader required too

Makes sense. Basically any top-level script will need it. We should probably make the presence of `BASE_PATH` optional and have `dirname(__DIR__)` or something get used if its not set...

dhensby:
That's a good point... In fact how is `BASE_PATH` being set at the moment... I know I told composer to bring it in but surely that's only once the autoloader is included?!

sminnee:
Yeah, it's defined by `Constants.php`. The logic worked on the assumption that the framework was always loaded the top of the project root. Once we move modules to the `vendor/` path then we'll probably need some new logic.

dhensby:
@sminnee OK - looking better now.

sminnee:
Yeah I'm happy with this. Merging!

sminnee:
FYI I just checked that CMS is okay with this by re-triggering the cms master build.

All looks good :) https://travis-ci.org/silverstripe/silverstripe-cms/builds/132344719

chillu:
After fetching from upstream on an existing `master` checkout, this fails with `Use of undefined constant BASE_URL`. So it doesn't pull in `core/Constants.php`. Which wouldn't come through for any travis/behat tests since they install from scratch.

I've tried a `composer dump-autoload` on the root project, which doesn't see any need to regenerate - and there's no way to force it.

```
composer dump-autoload -vvv
Running bdb9c5f22f9b4ebcffeff3844c3cb4d1d3232da8 (2016-04-29 16:46:27) with PHP 7.0.3 on Darwin / 15.4.0
Reading ./composer.json
Loading config file /Users/ingo/.composer/config.json
Loading config file /Users/ingo/.composer/auth.json
Loading config file ./composer.json
Checked CA file /private/tmp/composer-cacert-bcf331364fef73d53e4368ba519a6c3fcb21ea7836c3d8efad4ec32731e8152a.pem: valid
Executing command (/Users/ingo/Sites/master): git branch --no-color --no-abbrev -v
Executing command (/Users/ingo/Sites/master): git rev-list 3.1..demos/4hackday
Executing command (/Users/ingo/Sites/master): git rev-list master..demos/4hackday
Reading /Users/ingo/.composer/composer.json
Loading config file /Users/ingo/.composer/config.json
Loading config file /Users/ingo/.composer/auth.json
Loading config file /Users/ingo/.composer/composer.json
Loading config file /Users/ingo/.composer/auth.json
Reading /Users/ingo/.composer/vendor/composer/installed.json
```

So I'm trying a `rm -rf vendor && composer install` instead. This isn't ideal behaviour for anybody who has checked out master already. Maybe a `composer update` does the trick here?

sminnee:
@chillu just tested - `composer update` fixes this. The basic issue is that composer doesn't look in the composer.json files within the actually checked-out modules when building the autoloader. It looks somewhere else, presumably somewhere that is updated at `update` time.

In any case, this highlights that changes to framework's autoloading config are kind of a pain to bubble through, and should be kept to a minimum.

chillu:
I can't reproduce this any more (without trawling through my backup drive), a `rm composer.lock; composer install --prefer-source` did the trick in the end. I haven't tried a `composer update` - I'm trying to avoid running any `composer` commands really because they're so excruciatingly slow (compared to `git` commands). Plus, the changing and underdocumented behaviour on how composer deals with existing git remotes (deleting folders if it feels like it) - frustrating. Yes, I have to suck it up, and risk edge cases when not going through the composer happy path - which seems to have been the case this time.

dhensby:
I presume composer uses the lock file to compile the optimised autoloader and without it it definitely won't be checking the composer file or each module when trying to autoload

My understanding was that composer is a requirement of SS and so no one should experience any pain here as they'd have to use the composer update command to get it

sminnee:
"installed.json should be refreshed on `composer dump-autoload`" is probably worth of a bug report on composer/composer. I'll write it up more fully shortly.

PR #5573, opened 2016-05-20
FIX #5557 Tests with no DB requirements wont create test DB

### LINKED ISSUE #5557: SapphireTest::$usesDatabase seems to be a noop
The bugfix 43ff00ebd60adde181ef70cbd8d1a218c9a53fd9 changed the `SapphireTest` behavior: AFAIU the database is always created, also when you do not want to.

As an example, the following test:
```
class DummyTest extends SapphireTest
{
    protected $usesDatabase = false;

    public function testDummy() {
        $this->assertTrue(true);
    }
}
```

drops from 9.1 seconds to 0.1 seconds if I revert that commit.

dhensby:
- Removed duplicate logic that is already in `using_temp_db`
- Only need to create the temp db if we aren't using one - this logic is not needed to wrap the entire set of logic around fixtures, etc.

### DIFF
--- a/dev/SapphireTest.php
+++ b/dev/SapphireTest.php
@@ -223,13 +223,9 @@ class SapphireTest extends PHPUnit_Framework_TestCase {
 		$this->model = DataModel::inst();
 
 		// Set up fixture
-		if($fixtureFile || $this->usesDatabase || !self::using_temp_db()) {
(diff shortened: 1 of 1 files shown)


kinglozzer:
thumbsup

PR #5579, opened 2016-05-22
BUG Fix getFinalisedQuery not including all queried columns

### LINKED ISSUE #1669: [2013-01-21] getFinalisedQuery misses tables when looking for columns
created by: weboutreach
created at: 2013-01-21
original ticket: http://open.silverstripe.org/ticket/8208

---

The regular expression that extracts the column name from expressions like '"Table"."Column"' returns the table rather than the column.

tractorcow:
Fixes #1669 

My feeling is that in 4.0, we should probably NOT have automatic sub-table inclusion. Having more specific subclasses being added to the join table, rather than simply filtering on that class in the first place, is quite flakey behaviour.

If you want to filter on a subclass field, you should probably write `BlogPost::get()->filter('Author')` instead of `SiteTree::get()->filter('Author')`, and relying on the ORM to "guess" what you meant.

In any case, this should improve the quality of this terrible feature for 3.x.

### DIFF
--- a/model/DataQuery.php
+++ b/model/DataQuery.php
@@ -170,7 +170,9 @@ class DataQuery {
 	 * @return SQLQuery The finalised sql query
 	 */
 	public function getFinalisedQuery($queriedColumns = null) {
-		if(!$queriedColumns) $queriedColumns = $this->queriedColumns;
+		if(!$queriedColumns) {
+			$queriedColumns = $this->queriedColumns;
+		}
 		if($queriedColumns) {
 			$queriedColumns = array_merge($queriedColumns, array('Created', 'LastEdited', 'ClassName'));
 		}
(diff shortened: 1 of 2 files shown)


dhensby:
I had no idea this was possible `SiteTree::get()->filter('Author')`

tractorcow:
Yeah, it joins in any sub-table that has that field. It's kind of lazy and fragile. =(

dhensby:
Quite, I don't think we should do that!

hafriedlander:
It's important when you're trying to filter on a list and you don't know how it was created.

```
function filterByOwner(DataList $list) {
  return $list->filter('Author', $this->Name);
}
```

You shouldn't have to know how the initial list was created, and it's not possible to typecast to `DataList[Author]`.

A better solution would be to explicitly declare the type required though, which I've recommended for 4.

```
function filterByOwner(DataList $list) {
  return $list->whereTypeIs('Blog')->filter('Author', $this->Name);
}
```

or some such.

tractorcow:
I like the subclass-targetting syntax as a solution. It needs to be an explicit include, not an implicit one, as the current behaviour is fraught with edge cases.

PR #5581, opened 2016-05-23
Swap out .Actions class for bootstrap .btn-toolbar

clarkepaul:
Should be reviewed and accepted along with https://github.com/silverstripe/silverstripe-cms/pull/1505.
Replaces all instances of .Action within framework, removed some old styles which are no longer needed. There will only be minor changes to the the positioning of actions but nothing which should cause concern (if I've tested right). I've had to adjust a few related classes and styles to manage the transition of moving to Bootstrap so that the layout remains usable and unaffected.


sminnee:
This looks okay from a code review. Build failure is restricted to behat test.

sminnee:
Did a test on Chrome and can't see any issues.

PR #5586, opened 2016-05-24
Improve performance of TokenisedRegularExpression

kinglozzer:
tldr; `$tokenTypes` contained everything, then relied on `array_keys()` to filter. New approach is to only add to the `$tokenTypes` array if it matches.

Saves ~1.5mb peak memory on a blank install (tested with XHProf).

screen shot 2016-05-24 at 10 18 18

Also tested on the “largest” site (in terms of number of PHP classes) I had on my machine, this time using `time` and `memory_get_peak_usage(true)` instead of XHProf in case of any overhead:

```
<!-- 22 mb -->
real	0m3.215s
user	0m2.838s
sys	0m0.275s


<!-- 20 mb -->
real	0m2.878s
user	0m2.562s
sys	0m0.273s
```

Runs were done by simply hitting the home page after deleting the entire temp directory each time.

### DIFF
--- a/core/manifest/TokenisedRegularExpression.php
+++ b/core/manifest/TokenisedRegularExpression.php
@@ -13,32 +13,40 @@ class TokenisedRegularExpression {
 	 */
 	protected $expression;
 
+	/**
+	 * The first expression to match
+	 */
+	protected $firstMatch;
+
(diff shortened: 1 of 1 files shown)


dhensby:
Nice

PR #5590, opened 2016-05-25
Add namespace for appBoot, and fix syntax error in GridField

flamerohr:
We've found if a module creates another namespace, it triggers the appBoot again and throws an error.

I was debating about whether to use this pull request change or `$(document).ready(appBoot);`

### DIFF
--- a/admin/client/src/boot/index.js
+++ b/admin/client/src/boot/index.js
@@ -45,4 +45,6 @@ function appBoot() {
 // Entwine hooks are being used to set up the <Provider>.
 // `window.onload` happens AFTER these Entwine hooks which means
 // the store is undefined when the <Provider> is constructed.
-$('body').entwine({ onadd: () => appBoot() });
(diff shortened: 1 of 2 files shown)


tractorcow:
So if we have yet another module that creates a namespace (other than 'ss') the problem still exists?

flamerohr:
Yeah, another module that creates a namespace will cause another run of appBoot.

The problem was that the onadd event was not specific to 'ss', so it ended up being a global and was triggered whenever a new namespace was created.

I'm not sure what the benefits are to have `$.entwine({ onadd: appBoot });` over using `$(document).ready(appBoot);`, the latter seems to fix the issue as well and have less confusion with the onadd event.

hafriedlander:
onadd is default called by document.ready on first load anyway, and it's unlikely body could ever be added after the fact. There might be execution order issues though, so for now I'll merge as is.

PR #5592, opened 2016-05-25
Adjusted dbObject order of execution and make ignoreFields chainable

flamerohr:
Moving the check for `db($fieldName)` to the start would help ensure that `dbObject` returns a `DBField` object and not return `$dependencies` injected objects.
This way DataDifferencer would not require a DBField check as well.

ignoreFields `return $this;` so it could be chained e.g.
```
$diffFields = DataDifferencer::create($original, $new)
  ->ignoreFields(array('LastEdited'))
  ->ChangedFields();
```

### DIFF
--- a/ORM/DataObject.php
+++ b/ORM/DataObject.php
@@ -2954,24 +2954,27 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	 * @return DBField The field as a DBField object
 	 */
 	public function dbObject($fieldName) {
+		// Check for field in DB
+		$helper = $this->db($fieldName, true);
+
+		if(!$helper) {
+			return null;
+		}
+
 		$value = isset($this->record[$fieldName])
 			? $this->record[$fieldName]
 			: null;
 
 		// If we have a DBField object in $this->record, then return that
-		if(is_object($value)) {
+		if($value instanceof DBField) {
 			return $value;
 		}
 
-		// Build and populate new field otherwise
-		$helper = $this->db($fieldName, true);
-		if($helper) {
-			list($table, $spec) = explode('.', $helper);
-			$obj = Object::create_from_string($spec, $fieldName);
-			$obj->setTable($table);
-			$obj->setValue($value, $this, false);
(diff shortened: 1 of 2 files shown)


model/DataDifferencer.php
```diff
@@ -197,26 +199,29 @@ public function ChangedFields() {
...
-					$fieldDiff = Diff::compareHTML(
-						$this->fromRecord->$field,
-						$this->toRecord->$field,
-						(!$fieldObj || $fieldObj->stat('escape_type') != 'xml')
-					);
-				} else {
-					if($fieldObj && $fieldObj->stat('escape_type') == 'xml') {
-						$fieldDiff = "<ins>" . $this->toRecord->$field . "</ins>";
+
+				if($fieldObj instanceof DBField) {
```
tractorcow: Why would a return value from dbObject() not be a DBField?

flamerohr: `->toMap()` seems to return classes listed in the `$dependencies` properties as well, which could be regular classes.

Maybe dbObject() could check that before it returns the object found?

sminnee: `$this->record` on DataObject is populated by `DataObject::setField()`, which will be called by any setter, including injected `$depedencies`.

I think that the fix here is to return null if `dbObject()` is called to access something that isn't a DBField. It may even be beneficial to make it stricter, and only return properties that have been sourced from the database.

tractorcow: The easiest way to do this is to move the `$this->db()` call in `dbObject` further up to the top of the method, and return null if `$helper` is empty (even if $record has a value).

flamerohr:
I've rebased and applied a different change which @tractorcow had suggested.

dhensby:
You've just moved the check to the beginning of the method? Is that all you wanted to do?

Also, the comment is now out of context, can that be adjusted, please?

What's the advantage to moving the check to the start?

flamerohr:
adjusted the comments

dhensby:
thanks

PR #5598, opened 2016-05-26
Bugfix: BulkLoader_Result class - Deleted Items

AntonyThorpe:
The Deleted function, within BulkLoader_Result class, calls mapToArrayList, which tries to find the recently deleted dataobjects.
To correct:
- when calling addDeleted, store a map of each dataobject within the $this->deleted array; and,
- have the Deleted function return an ArrayList based upon the $this->deleted array.

### DIFF
--- a/dev/BulkLoader.php
+++ b/dev/BulkLoader.php
@@ -343,7 +343,11 @@ class BulkLoader_Result extends Object {
 	 * @return ArrayList
 	 */
 	public function Deleted() {
-		return $this->mapToArrayList($this->deleted);
+		$set = new ArrayList();
+		foreach ($this->deleted as $arrItem) {
+			$set->push(ArrayData::create($arrItem));
+		}
+		return $set;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


sminnee:
Thanks, @AntonyThorpe! Any chance you can write a test that covers this issue? Something that would fail without your change and passes with your change.

AntonyThorpe:
Tests added as requested.  Unfortunately the Travis CI build failed due to a Behat CMS test (file1.jpg missing in the HTML following inserting from the CMS file store).  Can you please advise the next steps?  Sorry, I am not familiar with Behat tests.  Thank you.

dev/BulkLoader.php
```diff
@@ -391,7 +391,7 @@ public function addUpdated($obj, $message = null) {
 	 */
 	public function addDeleted($obj, $message = null) {
 		$data = $obj->toMap();
-		$data['Message'] = $message;
+		$data['_BulkLoaderMessage'] = $message;
```
sminnee: Why did you change Message to _BuklLoaderMessage? It was Message in the original commit; was this a bug?

sminnee:
I'll re-execute a couple of builds to see if that failure is related to this PR.

dev/BulkLoader.php
sminnee: Oh, I see. It's mimicking the behaviour of `mapToArrayList()`

dhensby:
Thanks for this!

PR #5602, opened 2016-05-27
ActionTabSet, Change position absolute to bottom instead of top

flamerohr:
We have scenarios where the position of the actiontabset is hidden because it overflows behind the edge of the bottom of the window.

http://i.imgur.com/n00vWJa.png

### DIFF
--- a/admin/client/dist/js/LeftAndMain.ActionTabSet.js
+++ b/admin/client/dist/js/LeftAndMain.ActionTabSet.js
@@ -107,7 +107,7 @@
 				} else {
 					this.removeClass('rise-up');
 					if (activeTab.position() !== null) {
-						$(activePanel).css('bottom', '0px');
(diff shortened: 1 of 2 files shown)


sminnee:
Seems fine - can we see a screenshot of the fixed styling?

flamerohr:
Sure, a screenshot of before opening the tabset: http://i.imgur.com/93Glntl.png
and after opening the tabset http://i.imgur.com/RNLHdqV.png

kinglozzer:
Doesn’t that mean that the “More options” button is then hidden behind the menu? Would `bottom: 100%` make more sense?

flamerohr:
Fair point, it'll still be below the margin/padding at the top, but More options won't be hidden anymore.

I'll commit the change later on.

kinglozzer:
Thanks @flamerohr!

PR #5605, opened 2016-05-27
BUG fix CMS_ACCESS permission being ignored if in incorrect order in array

tractorcow:
@dhensby this is a bugfix for that permission you added a while back. :) I found out it's ignored if placed after some other CMS_ACCESS_ permission (which is necessary when writing backwards-compatible security code).

### DIFF
--- a/security/Permission.php
+++ b/security/Permission.php
@@ -185,13 +185,12 @@ class Permission extends DataObject implements TemplateGlobalProvider {
 						}
 					}
 				}
-				elseif (substr($permCode, 0, 11) === 'CMS_ACCESS_') {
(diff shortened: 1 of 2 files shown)


dhensby:
I'm confused, I don't see what this changes? It seems like it just means the `$code` array will have `CMS_ACCESS_LeftAndMain` in it several times...

tractorcow:
@dhensby this fixes a bug where, if you have a certain permission in your list of _any_ permissions, any subsequent items in that list are ignored. That's the problem with the break.

This fix makes the permission order agnostic.

I'll fix the issue with having the code duplicated, but it's minor.

tractorcow:
Ok, I've fixed the duplication error. :)

Do you understand how it can fail though, now? If it helps, try working through the unit tests I've added.

dhensby:
OK - I think I've got it... maybe. I'll let it in :P

PR #5606, opened 2016-05-29
Show the query number as part of debugging

Firesphere:
Show the query number, easier to see the amount of queries executed, and for debugging purposes of a certain slow query, also useful, since a breakpoint can be set, with a parameter saying something in the lines of `$this->queryCount == 50`, so the debugger only breaks on that specific query.

Without `dev` and `showqueries=1`, performance is not impacted according to XHProf.

Behat failures are unrelated. I have no idea where those come from.

### DIFF
--- a/model/connect/Database.php
+++ b/model/connect/Database.php
@@ -15,6 +15,13 @@ abstract class SS_Database {
 	 * @var DBConnector
 	 */
 	protected $connector = null;
+	
+	/**
+	 * Amount of queries executed, for debugging purposes.
+	 * 
+	 * @var int
+	 */
+	protected $queryCount = 0;
 
 	/**
 	 * Get the current connector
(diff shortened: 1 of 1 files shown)


model/connect/Database.php
```diff
@@ -171,14 +178,15 @@ protected function previewWrite($sql) {
...
+			$this->queryCount++;
 			$starttime = microtime(true);
 			$result = $callback($sql);
 			$endtime = round(microtime(true) - $starttime, 4);
 			$message = $sql;
 			if($parameters) {
 				$message .= "\nparams: \"" . implode('", "', $parameters) . '"';
 			}
-			Debug::message("\n{$message}\n{$endtime}s\n", false);
+			Debug::message("\n$this->queryCount: {$message}\n{$endtime}s\n", false);
```
tractorcow: In future please {} surround any string substitutions.

tractorcow:
Tests failed, so I've restarted them. They may have been fixed by now.

tractorcow:
Ok, test failures look unrelated; And framework / cms are probably diverged since this PR so it's unfair to make you take the blame. :)

Framework tests passed so I'll merge.

model/connect/Database.php
Firesphere: Will do, didn't know that actually worked like that :|

PR #5617, opened 2016-05-31
appBoot throwing error when given the same config twice

flamerohr:
I get this on the admin interface if one of my javascript contains an entwine namespace that is not 'ss' (or belonging to the 'ss' namespace)

e.g.
```
$.entwine('custom', () => {
  console.log('custom namespace actions');
});
```

The change in this pull request reduces the chance of the error getting thrown, but still throws the error if the config item given is different to the one existing.


tractorcow:
If it's given the same config twice, shouldn't it fail?

I'd rather track down the cause of the double registration and fix it there. This seems to just hide a true error.

flamerohr:
I don't think it's beneficial for it to fail when it gets the same config twice during production, but I agree it could fail during development builds though.

I can add a "is dev" check with the object check if that helps.

tractorcow:
Not sure how long this PR will remain relevant with the upcoming routing rewrites, but I guess we can leave it open as a placeholder for the issue.

flamerohr:
with the new way routes are working, I think this is safe to Close for now

PR #5619, opened 2016-05-31
BUGFIX: Fixed lookup of next closest visible field for focus restoring (fixes #5618)

### LINKED ISSUE #5618: Nested hidden form fields causes JavaScript error in the CMS
It looks like when you have nested form fields that are not visible a script error is thrown form LeftAndMain.EditForm.js line 350 (in 3.3's version). In short when restoreFieldFocus() tries to find the next visible element it's assuming that the next visible element is the next parent ``.field`` which may not be true in the case of say a FieldGroup inside of a ToggleComposite which is collapsed.

It looks like this issue can occur in [3.2](https://github.com/silverstripe/silverstripe-framework/blob/3.2/admin/javascript/LeftAndMain.EditForm.js#L321), [3.3](https://github.com/silverstripe/silverstripe-framework/blob/3.3/admin/javascript/LeftAndMain.EditForm.js#L350) and [3.4](https://github.com/silverstripe/silverstripe-framework/blob/3.4/admin/javascript/LeftAndMain.EditForm.js#L350).

I think the fix is as simple as adding a ``:visible`` flag to the selector where the nearest visible element is looked up. The diff below was made against 3.3.2, I can open a pull if the change makes sense.

```diff
diff --git a/admin/javascript/LeftAndMain.EditForm.js b/admin/javascript/LeftAndMain.EditForm.js
index a9ee20f..983c18f 100644
--- a/admin/javascript/LeftAndMain.EditForm.js
+++ b/admin/javascript/LeftAndMain.EditForm.js
@@ -318,7 +318,7 @@
 
                    //Fall back to nearest visible element if hidden (for select type fields)
                    if(!$(elementID).is(':visible')){
-                       elementID = '#' + $(elementID).closest('.field').attr('id');
+                       elementID = '#' + $(elementID).closest('.field:visible').attr('id');
                        scrollY = $(elementID).position().top;
                    }
```

UndefinedOffset:
As mentioned in #5618 if a field is not visible LeftAndMain.EditForm.js looks to the immediate parent for the next visible element which may not always be true. This results in a JavaScript error when the vertical position of that element is attempted to be retrieved. This pull request simply changes the selector to look for the closest visible element with the field class.

### DIFF
--- a/admin/javascript/LeftAndMain.EditForm.js
+++ b/admin/javascript/LeftAndMain.EditForm.js
@@ -318,7 +318,7 @@
 
 					//Fall back to nearest visible element if hidden (for select type fields)
 					if(!$(elementID).is(':visible')){
-						elementID = '#' + $(elementID).closest('.field').attr('id');
+						elementID = '#' + $(elementID).closest('.field:visible').attr('id');
 						scrollY = $(elementID).position().top;
 					}
 


UndefinedOffset:
oops disregard... pointed to the wrong merge point

PR #5622, opened 2016-05-31
Moved spinner to right so it does not overlap text

suhongyang:
Spinner is moved to right so it doesn't overlap the text. See attached screen shot to see the issue.

screen shot 2016-05-25 at 9 07 39 am


clarkepaul:
Looks good, I'll see if I can get this bumped through.

clarkepaul:
@tractorcow any chance you could merge this one, pretty simple fix.

admin/client/src/styles/legacy/_forms.scss
```diff
@@ -409,7 +409,7 @@ form.small .field, .field.small {
 				border: none;
 			}
 			&.loading {
-				background: transparent url(../images/network-save.gif) no-repeat $grid-x center;
+				background: transparent url(../images/network-save.gif) no-repeat right $grid-x center;
```
tractorcow: what does the change from `<num> center` to `right <num> center` mean for positioning? Is one of those three variables now needing to be removed?

dhensby:
needs rebasing too :)

admin/client/src/styles/legacy/_forms.scss
clarkepaul: Yup one needs to be removed, I'm not sure which one would render.

suhongyang: @tractorcow 
By changing the `background-position` offset to the 3 value syntax allows to move the ajax spinner background image to the far right, so it does not overlap the CMS action text.

`$grid-x` is 8px;

So `right $grid-x center` means: place the background image **8 pixels from the right**.

The 3rd value is the y-axis, which vertically aligns the image in the middle.

See https://css-tricks.com/almanac/properties/b/background-position/#article-header-id-1

dhensby: #learningeveryday

rebase please

clarkepaul: Me too, don't remember ever seeing that. Nice

suhongyang:
Please review new pull request #5812 with rebase.

PR #5646, opened 2016-06-02
API Allow table_name to be declared / introduce DataObjectSchema

### LINKED ISSUE #5419: table_name property for each DataObject
## Acceptance criteria

 * Data objects can take a `table_name` property.
 * `table_name` can be specified in the yaml config.
 * If `table_name` isn't specified, use SS3 naming convention

## Exclusions

 * Better defaults than SS3 not needed yet (not critical for alpha2)
 * Tables don't need to be automatically renamed (see #5589)

## Description

Once we start putting `DataObject` subclasses into namespaces, we're going to end up with ungainly table names, containing `\` characters. This functions correctly in SilverStripe 3, and some projects have used it, but it's not especially elegant.

I think the best first step is to introduce a `table_name` property, per DataObject class, that will be used in place of the class name in all generated queries.

Anyone who is using hand-rolled queries may trip over this, but I see this as yet another reason not to use hand-rolled queries. Some documentation about alternatives to hand-rolled queries (e.g. starting with a `SQLSelect` generated by the ORM and manipulating it) should be included in the docs.

----

Beyond this, I think it would be good to have some kind of mechanism for swapping out the table name generator functionality. Off the top of my head, I can think of a few examples:

 * Preserve the existing functionality, `\`s and all.
 * Replace `\` with `_` in class names
 * Use only the base name of the class
 * Apply a psr-4 style set of maps between short prefixes and long namespace prefixes
 * Any of the above, converted to lower case

I think that this can be broken out as a separate card, and for our initial namespacing work, we can simply map out bc-preserving `table_name` values in a yaml file.

It's probably worth raising this as a separate ticket, but it might be premature to do so until the table_name functionality is built.

## Related

 * Legacy reference (for SilverStripe internal team only, sorry): https://silverstripe.atlassian.net/browse/OSS-1856

## Tasks

 * [ ] Upgrading notes - https://github.com/silverstripe/silverstripe-framework/pull/5646
 * [ ] Implement `table_name` - https://github.com/silverstripe/silverstripe-framework/pull/5646
 * [ ] Default to current table name for core model
 * [ ] Update hard-coded references to core tables (e.g. SiteTree / Member) so that custom table_name is respected.

tractorcow:
This change introduces a `table_name` config for dataobjects, which can be used to customise the table name for any model. This is necessary as a precursor to namespacing, in order to prevent tables changing as we introduce namespaces for existing models.

I've split this change into a general "cleanup" commit for PHP linting errors I've resolved along the way, as well as the new API.

A minor CMS pull request is at https://github.com/silverstripe/silverstripe-cms/pull/1519

Fixes #5419

### DIFF
--- a/core/ClassInfo.php
+++ b/core/ClassInfo.php
@@ -14,6 +14,8 @@ class ClassInfo {
 
 	/**
 	 * Wrapper for classes getter.
+	 *
+	 * @return array
 	 */
 	public static function allClasses() {
 		return SS_ClassLoader::instance()->getManifest()->getClasses();
@@ -93,7 +95,9 @@ class ClassInfo {
 		);
 
 		foreach ($classes as $class) {
-			if (DataObject::has_own_table($class)) $result[$class] = $class;
+			if (DataObject::has_own_table($class)) {
+				$result[$class] = $class;
+			}
 		}
 
 		return $result;
@@ -143,7 +147,9 @@ class ClassInfo {
 	 * @return array Names of all subclasses as an associative array.
 	 */
 	public static function subclassesFor($class) {
-		if(is_string($class) && !class_exists($class)) return array();
+		if(is_string($class) && !class_exists($class)) {
+			return [];
+		}
 
 		//normalise class case
 		$className = self::class_name($class);
@@ -163,21 +169,14 @@ class ClassInfo {
 	 * eg: self::class_name('dataobJEct'); //returns 'DataObject'
 	 *
 	 * @param string|object $nameOrObject The classname or object you want to normalise
-	 *
 	 * @return string The normalised class name
 	 */
 	public static function class_name($nameOrObject) {
 		if (is_object($nameOrObject)) {
 			return get_class($nameOrObject);
 		} elseif (!self::exists($nameOrObject)) {
-			Deprecation::notice(
-				'4.0',
-				"ClassInfo::class_name() passed a class that doesn't exist. Support for this will be removed in 4.0",
-				Deprecation::SCOPE_GLOBAL
-			);
-			return $nameOrObject;
+			throw new InvalidArgumentException("Class {$nameOrObject} doesn't exist");
 		}
-
 		$reflection = new ReflectionClass($nameOrObject);
 		return $reflection->getName();
 	}
--- a/core/ClassInfo.php
+++ b/core/ClassInfo.php
@@ -78,8 +78,9 @@ class ClassInfo {
 	 * Returns an array of the current class and all its ancestors and children
 	 * which require a DB table.
 	 *
+	 * @todo Move this into {@see DataObjectSchema}
+	 *
 	 * @param string|object $class
-	 * @todo Move this into data object
 	 * @return array
 	 */
 	public static function dataClassesFor($class) {
@@ -104,28 +105,11 @@ class ClassInfo {
 	}
 
 	/**
-	 * Returns the root class (the first to extend from DataObject) for the
-	 * passed class.
-	 *
-	 * @param  string|object $class
-	 * @return string
+	 * @deprecated 4.0..5.0
 	 */
 	public static function baseDataClass($class) {
-		if(is_string($class) && !class_exists($class)) return null;
-
-		$class = self::class_name($class);
-
-		if (!is_subclass_of($class, 'DataObject')) {
-			throw new InvalidArgumentException("$class is not a subclass of DataObject");
-		}
-
-		while ($next = get_parent_class($class)) {
-			if ($next == 'DataObject') {
-				return $class;
-			}
-
-			$class = $next;
-		}
+		Deprecation::notice('5.0', 'Use DataObject::getSchema()->baseDataClass()');
+		return DataObject::getSchema()->baseDataClass($class);
 	}
 
 	/**
@@ -174,8 +158,6 @@ class ClassInfo {
 	public static function class_name($nameOrObject) {
 		if (is_object($nameOrObject)) {
 			return get_class($nameOrObject);
-		} elseif (!self::exists($nameOrObject)) {
-			throw new InvalidArgumentException("Class {$nameOrObject} doesn't exist");
 		}
 		$reflection = new ReflectionClass($nameOrObject);
 		return $reflection->getName();
@@ -291,53 +273,12 @@ class ClassInfo {
 		return strtolower(self::$method_from_cache[$lClass][$lMethod]) == $lCompclass;
 	}
 
-
 	/**
-	 * Returns the table name in the class hierarchy which contains a given
-	 * field column for a {@link DataObject}. If the field does not exist, this
-	 * will return null.
-	 *
-	 * @param string $candidateClass
-	 * @param string $fieldName
-	 *
-	 * @return string
+	 * @deprecated 4.0..5.0
 	 */
 	public static function table_for_object_field($candidateClass, $fieldName) {
-		if(!$candidateClass
-			|| !$fieldName
-			|| !class_exists($candidateClass)
-			|| !is_subclass_of($candidateClass, 'DataObject')
-		) {
-			return null;
-		}
-
-		//normalise class name
-		$candidateClass = self::class_name($candidateClass);
-		$exists = self::exists($candidateClass);
-
-		// Short circuit for fixed fields
-		$fixed = DataObject::config()->fixed_fields;
-		if($exists && isset($fixed[$fieldName])) {
-			return self::baseDataClass($candidateClass);
-		}
-
-		// Find regular field
-		while($candidateClass && $candidateClass != 'DataObject' && $exists) {
-			if( DataObject::has_own_table($candidateClass)
-				&& DataObject::has_own_table_database_field($candidateClass, $fieldName)
-			) {
-				break;
-			}
-
-			$candidateClass = get_parent_class($candidateClass);
-			$exists = $candidateClass && self::exists($candidateClass);
-		}
-
-		if(!$candidateClass || !$exists) {
-			return null;
-		}
-
-		return $candidateClass;
+		Deprecation::notice('5.0', 'Use DataObject::getSchema()->tableForField()');
+		return DataObject::getSchema()->tableForField($candidateClass, $fieldName);
 	}
 }
 
--- a/model/DataList.php
+++ b/model/DataList.php
@@ -111,8 +111,9 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 	 * Note that this function is re-entrant - it's safe to call this inside a callback passed to
 	 * alterDataQuery
 	 *
-	 * @param $callback
+	 * @param callable $callback
 	 * @return DataList
+	 * @throws Exception
 	 */
 	public function alterDataQuery($callback) {
 		if ($this->inAlterDataQueryCall) {
@@ -197,7 +198,7 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 	 * @return DataList
 	 */
 	public function where($filter) {
-		return $this->alterDataQuery(function($query) use ($filter){
+		return $this->alterDataQuery(function(DataQuery $query) use ($filter){
 			$query->where($filter);
 		});
 	}
@@ -215,7 +216,7 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 	 * @return DataList
 	 */
 	public function whereAny($filter) {
-		return $this->alterDataQuery(function($query) use ($filter){
+		return $this->alterDataQuery(function(DataQuery $query) use ($filter){
 			$query->whereAny($filter);
 		});
 	}
@@ -250,9 +251,10 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 	 *
 	 * @param int $limit
 	 * @param int $offset
+	 * @return DataList
 	 */
 	public function limit($limit, $offset = 0) {
-		return $this->alterDataQuery(function($query) use ($limit, $offset){
+		return $this->alterDataQuery(function(DataQuery $query) use ($limit, $offset){
 			$query->limit($limit, $offset);
 		});
 	}
@@ -261,9 +263,10 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 	 * Return a new DataList instance with distinct records or not
 	 *
 	 * @param bool $value
+	 * @return DataList
 	 */
 	public function distinct($value) {
-		return $this->alterDataQuery(function($query) use ($value){
+		return $this->alterDataQuery(function(DataQuery $query) use ($value){
 			$query->distinct($value);
 		});
 	}
@@ -293,9 +296,9 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 			throw new InvalidArgumentException('This method takes zero, one or two arguments');
 		}
 
-		$sort = $col = $dir = null;
-
 		if ($count == 2) {
+			$col = null;
+			$dir = null;
 			list($col, $dir) = func_get_args();
 
 			// Validate direction
@@ -372,6 +375,7 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 	 * Return a new instance of the list with an added filter
 	 *
 	 * @param array $filterArray
+	 * @return DataList
 	 */
 	public function addFilter($filterArray) {
 		$list = $this;
@@ -423,7 +427,7 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 			throw new InvalidArgumentException('Incorrect number of arguments passed to exclude()');
 		}
 
-		return $this->alterDataQuery(function($query, $list) use ($whereArguments) {
+		return $this->alterDataQuery(function(DataQuery $query) use ($whereArguments) {
 			$subquery = $query->disjunctiveGroup();
 
 			foreach($whereArguments as $field => $value) {
@@ -432,8 +436,6 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 				$filterType = array_shift($fieldArgs);
 				$modifiers = $fieldArgs;
 
-				// This is here since PHP 5.3 can't call protected/private methods in a closure.
-				$t = singleton($list->dataClass())->dbObject($field);
 				if($filterType) {
 					$className = "{$filterType}Filter";
 				} else {
@@ -443,8 +445,8 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 					$className = 'ExactMatchFilter';
 					array_unshift($modifiers, $filterType);
 				}
-				$t = new $className($field, $value, $modifiers);
-				$t->apply($subquery);
+				$filter = Injector::inst()->create($className, $field, $value, $modifiers);
+				$filter->apply($subquery);
 			}
 		});
 	}
@@ -465,9 +467,12 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 				gettype($callback)
 			));
 		}
+		/** @var ArrayList $output */
 		$output = ArrayList::create();
 		foreach($this as $item) {
-			if(call_user_func($callback, $item, $this)) $output->push($item);
+			if(call_user_func($callback, $item, $this)) {
+				$output->push($item);
+			}
 		}
 		return $output;
 	}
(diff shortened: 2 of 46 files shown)


core/ClassInfo.php
```diff
@@ -163,21 +153,14 @@ public static function subclassesFor($class) {
...
 		if (is_object($nameOrObject)) {
 			return get_class($nameOrObject);
 		} elseif (!self::exists($nameOrObject)) {
-			Deprecation::notice(
-				'4.0',
-				"ClassInfo::class_name() passed a class that doesn't exist. Support for this will be removed in 4.0",
-				Deprecation::SCOPE_GLOBAL
-			);
-			return $nameOrObject;
+			throw new InvalidArgumentException("Class \"{$nameOrObject}\" doesn't exist");
```
kinglozzer: If you’re tidying this up, can we just remove this entire `elseif` block? `new ReflectionClass` will throw an exception below this anyway :)

model/DataObject.php
```diff
@@ -2187,54 +2105,60 @@ public function manyManyComponent($component) {
 		foreach($classes as $class) {
```
kinglozzer: I feel the need to apologise for this method, even though I didn’t originally write it! The refactor looks good though

model/DataObject.php
```diff
@@ -2808,19 +2737,11 @@ public static function has_own_table_database_field($class, $field) {
...
-				self::$_cache_has_own_table[$dataClass]
-					= Config::inst()->get($dataClass, 'db', Config::UNINHERITED)
-					|| Config::inst()->get($dataClass, 'has_one', Config::UNINHERITED);
-			}
+		if(!is_subclass_of($dataClass, 'DataObject')) {
+			return false;
 		}
-		return self::$_cache_has_own_table[$dataClass];
+		$fields = static::database_fields($dataClass);
+		return !empty($fields);
```
kinglozzer: Nice!

model/DataList.php
```diff
@@ -1075,70 +1082,19 @@ public function removeByID($itemID) {
...
 			$query->reverseSort();
 		});
 	}
 
 	/**
-	 * This method won't function on DataLists due to the specific query that it represent
-	 *
-	 * @param mixed $item
-	 */
-	public function push($item) {
```
sminnee: Why have these been removed?

model/DataObject.php
```diff
@@ -231,6 +238,15 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	protected $unsavedRelations;
 
 	/**
+	 * Get schema object
+	 *
+	 * @return DataObjectSchema
+	 */
+	public static function getSchema() {
+		return DataObjectSchema::singleton();
```
sminnee: Best to use Injector?

model/DataQuery.php
```diff
@@ -846,12 +910,13 @@ public function subtract(DataQuery $subtractQuery, $field='ID') {
...
-	 * @param String $table Unquoted table name (will be escaped automatically)
-	 * @param Array $fields Database column names (will be escaped automatically)
+	 * @param string $table Unquoted table name (will be escaped automatically)
+	 * @param array $fields Database column names (will be escaped automatically)
+	 * @return $this
 	 */
 	public function selectFromTable($table, $fields) {
 		$fieldExpressions = array_map(function($item) use($table) {
-			return "\"$table\".\"$item\"";
+			return "\"{$table}\".\"{$item}\"";
```
kinglozzer:
```php
$schema = DataObject::getSchema();
$fieldExpressions = array_map(function($item) use($table, $schema) {
	return $schema->quotedColumn($table, $item);
}, $fields);
```

? :D

model/DataObjectSchema.php
```diff
@@ -0,0 +1,377 @@
...
+	 * This method will correctly match the column to the data class and table.
+	 *
+	 * Note that any fixed_fields will use the base table. Use quotedID() for joining
+	 * on the ID of non-base tables.
+	 *
+	 * @param string $class Class name
+	 * @param string $column Column name
+	 * @return string The quoted column, if this column belongs to this model
+	 */
+	public function quotedColumn($class, $column) {
```
sminnee: In principle, fields could be remapped as well as tables. To better support this, I think that the 2nd argument of `quotedColumn()` should be `$field`, not `$column`.

 * The inputs, class and field, are names used with the domain model
 * The output, a quoted table.column pair, refers to names used in the database persistence layer 

It doesn't really affect much other than the docblock, but it'll make things clearer for future developers.

search/filters/FulltextFilter.php
```diff
@@ -78,13 +82,14 @@ public function getDbName() {
...
-	*/
+	 */
 	protected function prepareColumns($columns) {
 		$cols = preg_split('/"?\s*,\s*"?/', trim($columns, '(") '));
-		$class = ClassInfo::table_for_object_field($this->model, current($cols));
-		$cols = array_map(function($col) use ($class) {
-			return sprintf('"%s"."%s"', $class, $col);
+		$table = DataObject::getSchema()->tableForField($this->model, current($cols));
+		$cols = array_map(function($col) use ($table) {
+			return sprintf('"%s"."%s"', $table, $col);
```
sminnee: Better to use quotedColumn() here?

search/filters/SearchFilter.php
```diff
@@ -161,29 +161,27 @@ public function setFullName($name) {
...
+		if(!$table) {
 			// fallback to the provided name in the event of a joined column
 			// name (as the candidate class doesn't check joined records)
 			$parts = explode('.', $this->fullName);
 			return '"' . implode('"."', $parts) . '"';
 		}
 
-		return sprintf('"%s"."%s"', $candidateClass, $this->name);
+		return sprintf('"%s"."%s"', $table, $this->name);
 	}
```
sminnee: Better to use quotedColumn() here?

model/Hierarchy.php
```diff
@@ -630,9 +630,12 @@ public function AllHistoricalChildren() {
...
 		}
 
-		$baseClass=ClassInfo::baseDataClass($this->owner->class);
-		return Versioned::get_including_deleted($baseClass,
-			"\"ParentID\" = " . (int)$this->owner->ID, "\"$baseClass\".\"ID\" ASC");
+		$baseTable = $this->owner->baseTable();
+		return Versioned::get_including_deleted(
+			$this->owner->baseClass(),
+			['"ParentID"' => $this->owner->ID],
+			"\"{$baseTable}\".\"ID\" ASC"
```
kinglozzer: `quotedColumn`?

model/ManyManyList.php
```diff
@@ -69,8 +69,11 @@ public function __construct($dataClass, $joinTable, $localKey, $foreignKey, $ext
 	 */
 	protected function linkJoinTable() {
 		// Join to the many-many join table
-		$baseClass = ClassInfo::baseDataClass($this->dataClass);
-		$this->dataQuery->innerJoin($this->joinTable, "\"{$this->joinTable}\".\"{$this->localKey}\" = \"{$baseClass}\".\"ID\"");
+		$dataClassIDColumn = DataObject::getSchema()->quotedColumn($this->dataClass(), 'ID');
+		$this->dataQuery->innerJoin(
+			$this->joinTable,
+			"\"{$this->joinTable}\".\"{$this->localKey}\" = {$dataClassIDColumn}"
```
kinglozzer: `quotedColumn()`?

model/ManyManyList.php
```diff
@@ -292,7 +296,9 @@ public function removeByID($itemID) {
 			user_error("Can't call ManyManyList::remove() until a foreign ID is set", E_USER_WARNING);
 		}
 
-		$query->addWhere(array("\"{$this->localKey}\"" => $itemID));
+		$query->addWhere(array(
+			"\"{$this->joinTable}\".\"{$this->localKey}\"" => $itemID
```
kinglozzer: `quotedColumn()`?

model/ManyManyList.php
```diff
@@ -364,7 +371,7 @@ public function getExtraData($componentName, $itemID) {
 			user_error("Can't call ManyManyList::getExtraData() until a foreign ID is set", E_USER_WARNING);
 		}
 		$query->addWhere(array(
-			"\"{$this->localKey}\"" => $itemID
+			"\"{$this->joinTable}\".\"{$this->localKey}\"" => $itemID
```
kinglozzer: `quotedColumn()`?

model/DataObjectSchema.php
```diff
@@ -0,0 +1,377 @@
...
+		$table = $this->tableForField($class, $column);
+		if($table) {
+			return "\"{$table}\".\"{$column}\"";
+		}
+		return null;
+	}
+
+	/**
+	 * Quotes the ID column of the specific table.
+	 * Unlike quotedColumn, does not force "ID" to the base table.
```
sminnee: I feel that this should be given a name to discourage its use except for in building up the multi-table inheritance joins. E.g. `quotedSubclassID()`?

Regular use should always refer to the base table's ID, so that a missing record in a subtable doesn't break things.

model/versioning/Versioned.php
```diff
@@ -527,52 +537,35 @@ public function augmentLoadLazyFields(SQLSelect &$query, DataQuery &$dataQuery =
...
 		if(
 			!empty($dataObject->Version) &&
 			(!empty($versionedMode) && in_array($versionedMode,$modesToAllowVersioning))
 		) {
 			// This will ensure that augmentSQL will select only the same version as the owner,
 			// regardless of how this object was initially selected
+			$baseTable = $this->baseTable();
 			$dataQuery->where([
-				"\"$dataClass\".\"Version\"" => $dataObject->Version
+				"\"{$baseTable}\".\"Version\"" => $dataObject->Version
```
kinglozzer: `quotedColumn()`?

model/DataQuery.php
```diff
@@ -165,13 +173,18 @@ public function setQueriedColumns($queriedColumns) {
...
+		}
 		if($queriedColumns) {
 			$queriedColumns = array_merge($queriedColumns, array('Created', 'LastEdited', 'ClassName'));
 		}
 
+		$schema = DataObject::getSchema();
 		$query = clone $this->query;
-		$ancestorTables = ClassInfo::ancestry($this->dataClass, true);
+		$baseDataClass = $schema->baseDataClass($this->dataClass());
+		$baseIDColumn = $schema->quotedID($baseDataClass);
```
sminnee: Could also use `quotedColumn($this->dataClass(), 'ID')` and let it look up the base class.

model/versioning/Versioned.php
```diff
@@ -634,87 +634,75 @@ public function augmentDatabase() {
...
+		// If we have a parent table limit orphaned records
+		// to only those that exist in this
+		if(DB::get_schema()->hasTable($baseTable)) {
+			$orphanedQuery
+				->addLeftJoin(
+					$baseTable,
+					"\"{$childTable}\".\"RecordID\" = \"{$baseTable}\".\"RecordID\"
+					AND \"{$childTable}\".\"Version\" = \"{$baseTable}\".\"Version\""
+				)
+				->addWhere("\"{$baseTable}\".\"ID\" IS NULL");
```
kinglozzer: `quotedColumn()`, `quotedColumn()`, `quotedColumn()`, `quotedColumn()`, `quotedColumn()`? :D

I suppose using that API for *everything* will get a bit unwieldy, but then again escaped quotes = just as much ewww

model/DataQuery.php
```diff
@@ -724,18 +757,19 @@ protected function joinHasOneRelation($localClass, $localField, $foreignClass)
...
 		}
 
-		$realModelClass = ClassInfo::table_for_object_field($localClass, "{$localField}ID");
-		$foreignBase = ClassInfo::baseDataClass($foreignClass);
-		$this->query->addLeftJoin(
-			$foreignBase,
-			"\"$foreignBase\".\"ID\" = \"{$realModelClass}\".\"{$localField}ID\""
-		);
+		// Join base table
+		$foreignIDColumn = $schema->quotedID($foreignBaseClass);
```
sminnee: Again, could just use `quotedColumn($foreignClass, 'ID')`

model/versioning/VersionableExtension.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+/**
+ * Minimum level extra fields required by extensions that are versonable
+ */
+interface VersionableExtension {
```
kinglozzer: Does this need to be noted in the changelog?

kinglozzer:
Ignore my repeated comments about `DataObjectSchema::quotedColumn()`, I was misunderstanding where it needs to be used :)

security/Security.php
```diff
@@ -992,33 +992,46 @@ public static function encrypt_password($password, $salt = null, $algorithm = nu
...
 			// Skip test classes, as not all test classes are scaffolded at once
-			if(is_subclass_of($table, 'TestOnly')) continue;
+			if(is_subclass_of($class, 'TestOnly')) {
+				continue;
+			}
 
 			// if any of the tables aren't created in the database
-			if(!ClassInfo::hasTable($table)) return false;
+			$table = DataObject::getSchema()->tableName($class);
+			if(!ClassInfo::hasTable($table)) {
```
kinglozzer: This method is still in `ClassInfo`? :(

model/DataObjectSchema.php
sminnee: Having a look at the other code, I wonder if it's worth having a small object to represent a column reference:

    class SQLColumn {
       __construct($table, $column);
       function getQuoted();
       function getTable();
       function getColumn();
    }

Then you could use the data schema to get a table/column pair once and use it for various comparisons throughout a piece of logic:

    $foreignKey = $schema->columnFor($localClass, "{$localField}ID");
    $baseKey = $schema->columnFor($dataClass, "ID");
    if (!$this->query->isJoinedTo($foreignKey->getTable()) {
        $this->query->addLeftJoin(
			$foreignKey->getTable(),
            $baseKey>getQuoted() . ' = ' . $foreignKey->getQuoted()
		);
    }

kinglozzer:
Very nice work @tractorcow

model/versioning/Versioned.php
```diff
@@ -2126,8 +2118,11 @@ public static function prepopulate_versionnumber_cache($class, $stage, $idList =
 			$parameters = $idList;
 		}
 
-		$baseClass = ClassInfo::baseDataClass($class);
-		$stageTable = ($stage == static::DRAFT) ? $baseClass : "{$baseClass}_{$stage}";
+		$baseClass = DataObject::getSchema()->baseDataClass($class);
+		$stageTable = DataObject::getSchema()->tableName($baseClass);
+		if($stage === static::LIVE) {
```
sminnee: This is a change in logic; it will break if there are any stage other than draft or live (which, admittedly, we never do).

`if($stage !== static::DRAFT)` would preserve the original logic.

sminnee:
Looking good; I've made a number of comments but they're mostly minor points.

Would be good to get a couple more @silverstripe/core-team members to review as it's a significant change.

core/ClassInfo.php
tractorcow: Sure, as long as it promises to break. :D

model/DataList.php
tractorcow: These are throwbacks to 2.4 when ManyManyList extended DataList.

tractorcow: As a note, I've also removed the overrides in ManyManyList that refers to DataList methods that it didn't support, so this is just the other half of that cleanup.

model/DataObject.php
tractorcow: Yes I think so.

model/DataObject.php
tractorcow: If I had more time I'd have rewritten it more.

model/DataObjectSchema.php
tractorcow: > In principle, fields could be remapped as well as tables. To better support this, I think that the 2nd argument of quotedColumn() should be $field, not $column.

I can rename the parameter. Er, the rest is good to think about, but out of scope for this change I'm afraid. :P

tractorcow: @hafriedlander had an idea for using a table casting helper that could quote table.column identifier in tables.

```php
$helper = $obj->getTableHelper();
return "SELECT {$helper->Title} FROM {$helper}";
```

model/DataObjectSchema.php
tractorcow: We can discourage it's use by removing it.

model/DataQuery.php
tractorcow: Yep, I think I'll remove quotedID, since its use is limited.

model/DataQuery.php
tractorcow: Yep.

model/DataQuery.php
tractorcow: quotedColumn() takes a class not a table as a first parameter, though.

model/Hierarchy.php
tractorcow: Yeah I can try that.

model/ManyManyList.php
tractorcow: Doesn't work on tables (e.g. many_many maps) that don't have DataObject backing.

model/ManyManyList.php
tractorcow: Same as above; Doesn't work on non-dataobject tables.

model/ManyManyList.php
tractorcow: same as above.

model/versioning/VersionableExtension.php
tractorcow: Ah, I can add a line item. It's pretty minor, and I don't think anyone is really using this feature. I half felt it could be removed.

model/versioning/Versioned.php
tractorcow: Yeah I think so.

model/versioning/Versioned.php
tractorcow: Nope, doesn't work on `Record_version` tables. RecordID and Version are added to _versions outside of the regular data-model.

The method isn't FOR quoting... it's for mapping classes to tables on a per-column basis.

I'm going to rename quotedColumn() to something else to make it clearer. Maybe `matchTableColumn` or something.

model/versioning/Versioned.php
tractorcow: In the prior update to ChangeSets we vetod stages other than Live or Stage. This change just cleans up some code that missed the original rewrite.

I'll update set_stage with an assert to make it clear.

search/filters/FulltextFilter.php
tractorcow: I really should not have called this method quotedColumn. The intent of that method is to match a class to a specific table where a given field is declared. I can see the confusion this has caused.

Maybe I should just use tableForField() where applicable and make users continue to ansi-sql quote their identifiers.

security/Security.php
tractorcow: I'd have moved it if I had more time.

tractorcow:
It looks like `quotedColumn` might need a re-think. It isn't apparent that it doesn't work with arbitrary table names, only classes. It's also apparent that the MAIN purpose for that method is a wrapper for tableForField() with a bit of hacked on formatting for convenience.

I feel it'd be better to just remove it (and quotedID) and just inline the tableForField() calls where used.

sminnee:
> I feel it'd be better to just remove it (and quotedID) and just inline the tableForField() calls where used.
 Show all checks

What about a value object that's the result of `a tableForField()` has a `dbName()` or `9

model/DataQuery.php
tractorcow: I'm renaming to `sqlColumnForField` and clarifying the need to pass in a class and not a table. It now throws an exception if you pass in a field that doesn't exist on the class, or if you pass in a $class that isn't a class.

tractorcow:
I had a think about it and decided to clarify the purpose of 'quotedColumn' instead of removing it. I've renamed it to sqlColumnForField, to mirror the naming of tableForField

model/DataObjectSchema.php
tractorcow: the return values for each field could be a SQLColumn, as you said above, but we could implement __toString to return the quoted value for each. Then the below will work.

```php
$helper = $obj->getTableHelper();
return "SELECT {$helper->Title} FROM {$helper->Title->Table} INNER JOIN {$helper->BaseTable} ON {$helper->BaseTable->ID} = {$helper->ChildTable->ID}";
```

model/DataObjectSchema.php
tractorcow: It wasn't used in many places, and where it was, it was either not necessary (already on a base class) or its purpose wasn't clear. I've removed it.

model/versioning/Versioned.php
tractorcow: > escaped quotes = just as much ewww

It's the standard for now, and I don't think we can get away from it. I made a failed attempted and realised it was much more practical to keep the convention than try to encapsulate it.

tractorcow:
Ok, updated. :D

Issue #5547, opened 2016-05-17
$Content.Parse breaks [embed] shortcode.


Firesphere:
When using .Parse in a template, to parse for example BBCodes, (giving the BBCodeParser as the to-use parser), the embed shortcode isn't parsed anymore.
Also, it's entirely XML escaped, I guess that's on purpose? If so, may I ask why?

axyr:
This remembers me of the 2.x era ;)

Firesphere:
It looks like a relic from 2.x :D

dhensby:
Delete .parse? 

Make sure parse returns a `HTMLText` object and the shortcode will be parsed and it will be escaped appropriately... I hope

Firesphere:
It's the DBText.Parse() method, which is a relic, that should be moved to the Forum module, since it's the only module that actively uses it.
Also, it relies on the very outdated BBCodeParser.

dhensby:
Then we know what to do

tractorcow:
Will try to address in #5653

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-framework/pull/5653 :D

Issue #3925, opened 2015-02-22
HTMLText un-escaped when shortcode parser is used


dhensby:
```php
$field = DBField::create_field('HTMLText', '&quote;');
$field->XML_val('ATT', null, true)->forTemplate(); // gives '"' instead of '&quote;'
```

However, when turning off shortcode parsing this does not happen

This is because the way the shortcode parser uses `DOMDocument` and how `DOMDocument` will unencode entities it doesn't think need to remain encoded within a `DOMDocument`.

This is fixed by (part of) https://github.com/dhensby/silverstripe-framework/commit/701112898a23d819d2aae70624c74ddab09de902

tractorcow:
`&quote;` Isn't a HTML entity. You mean `&quot;`? That's why. :P

tractorcow:
Unit test that passes:

```
$data = DBField::create_field('HTMLText', '&quot;');
$this->assertEquals('&quot;', $data->ATT());
```

PR #5653, opened 2016-06-03
API Enforce default_cast

tractorcow:
Replaces #3924

Includes lots of cleanup and additional testing.

The main commit addresses several changes all in one go, which were a bit hard to split into individual commits, as they depend on one another:

```
API Enforce default_cast for all field usages
API Introduce HTMLFragment as casting helper for HTMLText with shortcodes disabled
API Introduce DBField::CDATA for XML file value encoding
API RSSFeed now casts from the underlying model rather than by override
API Introduce CustomMethods::getExtraMethodConfig() to allow metadata to be queried
BUG Remove _call hack from VirtualPage
API Remove FormField::$dontEscape
API Introduce HTMLReadonlyField for non-editable readonly HTML
API FormField::Field() now returns string in many cases rather than DBField instance.
API Remove redundant *_val methods from ViewableData
API ViewableData::obj() no longer has a $forceReturnObject parameter as it always returns an object
BUG  Fix issue with ViewableData caching incorrect field values after being modified.
```

CMS pull request https://github.com/silverstripe/silverstripe-cms/pull/1520


tractorcow:
I'd like someone to please do a CMS test; We need to test for UI regressions as a result of this change.

tractorcow:
@dhensby noted that we could probably have cast some of the form field properties to Text  instead of HTMLFragment. I'd accept this as a feedback for possible improvement. :D

dhensby:
I've run a complete test suite of a default install with this (ie: cms, framework, reports and siteconfig) and they all passed.

dhensby:
[holder comment I'll add issues to as I find them]

tractorcow:
Probably need to look at fixing shortcodes at the same time:

- https://github.com/silverstripe/silverstripe-framework/issues/2929
- https://github.com/silverstripe/silverstripe-framework/issues/1381
- https://github.com/silverstripe/silverstripe-framework/issues/3925
- https://github.com/silverstripe/silverstripe-framework/issues/5547

Also will fix the regressions you've noted.

dhensby:
The cms issue is blocking me testing any further

tractorcow:
I've added a new commit for rewriting a bunch of the DBString field summary methods. I haven't got up to implementing these methods for the DBHTMLText field, but I suspect we might simply force these to strip HTML for simplicity. See the @todo

tractorcow:
I think I'm happy with the DBText summary methods now... I'm done with the rewrite, so now I'm just testing and fixing.

@dhensby do you want to check I'm going in the right direction?

Rather than attempting to do XML manipulation in an unsafe way, I'm converting HTML fields to plain text for any summary method.

I'm also looking at a proper XML safe implementation based on http://www.pjgalbraith.com/truncating-text-html-with-php/.

tractorcow:
Rebased and fixed a few unit tests. Still more tests to resolve:


- [ ] RSSFeedTest
- [ ] DBFieldTest
- [ ] DBHTMLTextTest
- [ ] DBTextTest
- [ ] SSViewerTest
- [ ] ViewableDataTest

tractorcow:
Ok, pushed up a fix that fixes ALL of the shortcode issues, except for #5678, which I'll do tomorrow.

tractorcow:
Another squash and rebase, cleanup, fix of last minute DB crashing when regenerating Enum columns.... ugh.

tractorcow:
TestSession module needs update... it still refers to a removed/deprecated method. ;P

tractorcow:
Ok, only broken tests rely on the CMS pr.

@dhensby want to re-review / test to see if the CMS is working now?

model/FieldType/DBComposite.php
```diff
@@ -223,11 +223,14 @@ public function hasField($field) {
 	 * @param string $field
 	 * @param mixed $value
 	 * @param bool $markChanged
+	 * @return $this
 	 */
 	public function setField($field, $value, $markChanged = true) {
+		$this->objCacheClear();
```
sminnee: Do have to clear all cache values here, and not simply the one for key `$field`?

forms/FormField.php
```diff
@@ -255,6 +251,22 @@ class FormField extends RequestHandler {
 	 */
 	protected $schemaData = [];
 
+	private static $casting = array(
+		'FieldHolder' => 'HTMLFragment',
+		'Field' => 'HTMLFragment',
```
sminnee: If Field is already using `renderTemplate()`, which returns an HTMLFragment, why does it casting? Is it just here for the benefit of subclasses that don't use a template?

view/ViewableData.php
```diff
@@ -384,17 +337,32 @@ protected function objCacheName($fieldName, $arguments) {
...
 		$this->objCache[$key] = $value;
+		return $this;
+	}
+
+	/**
+	 * Clear object cache
+	 *
+	 * @return $this
+	 */
+	protected function objCacheClear() {
```
sminnee: Are we able to invalidate caches key-by-key? I'd be a bit worried about the performance impacts of clean-sweep caching...

forms/DropdownField.php
```diff
@@ -110,7 +110,7 @@ protected function getFieldOption($value, $title) {
 
 	/**
 	 * @param array $properties
-	 * @return HTMLText
+	 * @return string
```
sminnee: does this return a string?

forms/FormAction.php
```diff
@@ -83,7 +91,7 @@ public function setFullAction($fullAction) {
 
 	/**
 	 * @param array $properties
-	 * @return HTMLText
```
sminnee: Doesn't this return HTMLFragment?

forms/FormAction.php
```diff
@@ -100,7 +108,7 @@ public function Field($properties = array()) {
...
+	 * @return string
```
sminnee: Doesn't this return HTMLFragment?

forms/FormField.php
```diff
@@ -925,7 +939,6 @@ public function setSmallFieldHolderTemplate($smallFieldHolderTemplate) {
 	 * such as an input tag.
 	 *
 	 * @param array $properties
-	 *
```
sminnee: Doesn't this return HTMLFragment?

view/ViewableData.php
tractorcow: The problem is we don't know what caches will be invalidated. E.g. if you modify URLSegment, then the Link, RelativeLink, etc caches are all invalid.

The use case is VERY limited, as rarely are objects modified during rendering.

forms/FormField.php
tractorcow: Because it doesn't always use renderTemplate(). This is a check to ensure it is forced to HTMLFragment in case subclasses don't.

model/FieldType/DBComposite.php
tractorcow: Copy pasted from below:

The problem is we don't know what caches will be invalidated. E.g. if you modify URLSegment, then the Link, RelativeLink, etc caches are all invalid.

The use case is VERY limited, as rarely are objects modified during rendering.

forms/DropdownField.php
tractorcow: I'm going to change this to string|HTMLText since there are so many variaties of formfields that either do raw string manipulation or use renderTemplate().

tractorcow:
Also, The problem is we don't know what caches will be invalidated. E.g. if you modify URLSegment, then the Link, RelativeLink, etc caches are all invalid.

Also, now that https://github.com/silverstripe/silverstripe-framework/pull/4162#event-706478991 has been merged, we can go back to associative arrays for the DBHTMLField options. @dhensby if you merge up I can rebase this (with the fixes @sminnee suggested above) and then we can clean up a few of these casting helpers.

dhensby:
I've had a play - looking good. The login form message issue still exists, though

dhensby:
Does the `HTMLText::exists` method need updating?

tractorcow:
I've done a bunch of refactoring and implemented the feedback I've been given.

I still need to fix some more test regressions, and update the DBString::parseConstructorOptions() to use proper associative arrays, since object::parse_class_spec now properly supports it. :)

tractorcow:
I got the new DBHTMLText constructor format working, so now you can use normal PHP array syntax in your db casting helpers.

However, I had too much trouble making Field() (and other) methods to return plain strings instead of DBHTMLText instances, so I've left this behaviour in place.

Once we get the new casting behaviour in, changing these methods can be another task, but I'd rather get this in for now, and not try to do too much at once.

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/pull/5798 too

tractorcow:
@dhensby fixed the login form problem. Not in the best way I could think of... we still need a way of setting custom castings on customised templates.

Please let's merge this. :)

dhensby:
Ok. Codeclimate failures are non-issues

Travis failures are cms problems that will be resolved with the cms pr

dhensby:
Top job!

tractorcow:
After more than a year of work we are done. :P

dhensby:
"done"

tractorcow:
Fixes for regressions at https://github.com/silverstripe/silverstripe-cms/pull/1546 and https://github.com/silverstripe/silverstripe-framework/pull/5806

PR #5654, opened 2016-06-03
Reduce number of SHOW queries we run

dhensby:
`SHOW` queries are quite slow to run and we know all the information we need through the class hierarchy.

### DIFF
--- a/core/ClassInfo.php
+++ b/core/ClassInfo.php
@@ -30,11 +30,6 @@ class ClassInfo {
 	}
 
 	/**
-	 * Cache for {@link hasTable()}
-	 */
-	private static $_cache_all_tables = array();
-
-	/**
 	 * @var Array Cache for {@link ancestry()}.
 	 */
 	private static $_cache_ancestry = array();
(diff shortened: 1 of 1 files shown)


tractorcow:
Not quite correct dan... a class can be a subclass of DataObject and it won't have a table, if it has no DB fields.

This method is also used to see if the database schema MATCHES the dataobject schema. The problem with this code is that it assumes that the database is in sync with the data model, and that's not always the case.

For instance, this is what's in Security.php when checking if a table is safe to select from.

```
// if any of the tables aren't created in the database
if(!ClassInfo::hasTable($table)) return false;
```

tractorcow:
See `DataObject::has_own_table()` in DataObject for an existing implementation.

dhensby:
> a class can be a subclass of DataObject and it won't have a table, if it has no DB fields.

Ah, of course

> This method is also used to see if the database schema MATCHES the dataobject schema. The problem with this code is that it assumes that the database is in sync with the data model, and that's not always the case.

`ClassInfo` should not be used to inspect the state of the DB, it's information about the Classes and whether they should have a table or not, it shouldn't be checking to see if there currently is a DB table or not.

I don't think we should be checking on every request that the DB is either built or in sync with the schema (and we don't do that very thoroughly now). Automatically building the DB for users isn't a great idea because it impacts performance and masks bugs. 

Regarding masks bugs, if I import a DB then reload and (for some reason my backup was imported to the wrong DB) SS builds the db and it appears like I've imported an empty DB which is confusing and frustrating.

I think the building the DB should always have to be the job of the user and not automated.

PR #5655, opened 2016-06-03
FIX: ViewableData::setFailover() didn't remove cached methods

kinglozzer:
Hopefully the unit test explains what’s going on here. Turns out @tractorcow was quite correct in his comment in https://github.com/silverstripe/silverstripe-framework/pull/4870#issuecomment-166722469:

> You might also need a mechanism and test to ensure you remove methods from the removed failover, since it looks like the new ones are being merged with the old.

### DIFF
--- a/core/Object.php
+++ b/core/Object.php
@@ -801,6 +801,25 @@ abstract class Object {
 	}
 
 	/**
+	 * @param object $extension
+	 * @return array
+	 */
+	protected function findMethodsFromExtension($extension) {
+		if (method_exists($extension, 'allMethodNames')) {
+			if ($extension instanceof Extension) $extension->setOwner($this);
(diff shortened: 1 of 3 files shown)


dhensby:
That refactor commit needs to go into 3 branch IMO - it's not a bug fix.

dhensby:
In fact, if we want to be real anal, this should all go into 3 and you need to fix this issue in 3.4 but putting all the logic in `setFailover` :/

kinglozzer:
The thing is it *can’t* all go in `setFailover()` - `Object::$extra_methods` is private, so unless we fancy messing around with `ReflectionProperty` I can’t see another way to do this :(

kinglozzer:
I can remove that refactor commit though

dhensby:
Let's just put it in 3?

PR #5658, opened 2016-06-06
FIX: Regression with <% with $Up/$Top %> (fixes #5656)

### LINKED ISSUE #5656: 3.4 regression: <% with $Top %> and <% with $Up %> fails
This works on 3.4 and lower:
```html
$Up.Title
$Top.Title
```

But this only works on 3.3 and lower:
```html
<% with $Up %>
	$Title
<% end_with %>

<% with $Top %>
	$Title
<% end_with %>
```

kinglozzer:


dhensby:
THANSK

jonom:
Thanks Loz!

PR #5659, opened 2016-06-06
[WIP] ManyMany relations with shared inheritance

dhensby:
see #5531


model/DataQuery.php
```diff
@@ -13,6 +13,8 @@
  */
 class DataQuery {
 
+	public $lastAlias;
```
tractorcow: Why public and without any PHPDoc?

model/DataQuery.php
```diff
@@ -702,6 +705,15 @@ public function applyRelation($relation) {
 				$this->query->addLeftJoin($componentBaseClass,
 					"\"$relationTable\".\"$componentField\" = \"$componentBaseClass\".\"ID\"");
 				}
+				else {
+					$alias = uniqid($componentBaseClass);
+					$this->query->addLeftJoin(
```
tractorcow: So we just alias as an unintelligible string? Doesn't seem very fluent to me... you should see my feedback on https://github.com/silverstripe/silverstripe-framework/pull/5531/files#r63108288

tractorcow:
I think using a new alias is fine, but it needs to be deterministic, not some random string that makes no sense to the person debugging their code.

If it's deterministic, it doesn't need to be stored anymore, as it can be simply generated as-needed.

tractorcow:
Any update on this @dhensby ? :P

tractorcow:
Will check back again in another month. :) Maybe you want to ping me tomorrow @dhensby ?

model/DataQuery.php
dhensby: because this was a hacky PoC rather than ever meant to be final

model/DataQuery.php
dhensby: it's not entirely unitelligible because it would be somthing like `SiteTree4b3403665fea6`

tractorcow:
@dhensby you've somehow hosed your pull request... it now contains ONLY test changes. The code is lost?

dhensby:
I've got them locally - I was just preparing to implement a "deterministic" rename (the second commit that was here was just a nasty PoC).

I don't really like the idea of a deterministic naming because it's tight coupling between the two components, but for now we have to work with it, I think.

model/DataQuery.php
```diff
@@ -640,6 +640,61 @@ public function leftJoin($table, $onClause, $alias = null, $order = 20, $paramet
...
+	protected function relationWalk($relations, $cb)
+	{
+		if (is_string($relations)) $relations = explode('.', $relations);
+
+		$modelClass = $this->dataClass;
+
+		foreach ($relations as $relationName) {
+			/** @var DataObject $model */
+			$model = singleton($modelClass);
+			if ($relationClass = $model->hasOneComponent($relationName)) {
```
jonom: Could you make use of [`getRelationClass()`](http://api.silverstripe.org/3.3/source-class-DataObject.html#_getRelationClass) here instead of doing these if statements?

dhensby: Perhaps!

tractorcow:
Every time I come to look at this pull request, I feel like I'm looking down into the depths of a dark, bottomless well. :)

sminnee:
I've marked this as WIP because @dhensby doesn't appear to think this is ready. If I was wrong about that, please revert

dhensby:
@sminnee yep - still WIP this is definitely one difficult issue to resolve - pretty close, though

sminnee:
@dhensby Damian has been re-working many-many relations, specifically creating a more flexible many-many-through system, for SS4. Given that, do you still want to keep working on this for SS3 or should we close it off?

tractorcow:
> @dhensby Damian has been re-working many-many relations, specifically creating a more flexible many-many-through system, for SS4. Given that, do you still want to keep working on this for SS3 or should we close it off?

My understanding is that this work will be needed anyway... it addresses the issue of not being able to walk through any relationship that's mapped onto itself, which is going to be a problem both before and after my work.

tractorcow:
@dhensby I want to take over this pull request and work on it in my own time. Is this ok with you?

tractorcow:
I'm re-assigning this as an enhancement, although it's probably a weakness which could easily be classified as a bug as well. :)

dhensby:
> @dhensby I want to take over this pull request and work on it in my own time. Is this ok with you? I MAY implement it as a 4.0 only solution, however.

sure

tractorcow:
Ok, I've started looking at a new table-prefix system for applying relations.

DataList::applyRelation() looks like this now:

```
// Apply relation
$relationModelName = $query->applyRelation($relations, $linearOnly);
$relationPrefix = $query->applyRelationPrefix($relations);

 // Find the db field the relation belongs to
$columnTable = DataObject::getSchema()->tableForField($relationModelName, $fieldName);
$columnName = "\"{$relationPrefix}{$columnTable}\".\"{$fieldName}\"";
```

tractorcow:
Got it working, just fixing up rest of tests. :)

tractorcow:
Replaced with https://github.com/silverstripe/silverstripe-framework/pull/6939/

Issue #5657, opened 2016-06-06
Missing back button in 3.4


minimalic:
I'm sure we are missing #5009 (commit `f39989a`) in 3.4, so there is no more _visible_ back button in CMS.

dhensby:
That PR is in the 3.4.0 release, though there could have been a problem with the css compilation since?

UndefinedOffset:
Back button is there based on the html, looks to me like the selector ``.font-icon-left-open-big`` is missing out of (_fonts.scss)[https://github.com/silverstripe/silverstripe-framework/blob/3.4/admin/scss/_fonts.scss] in the 3.4 branch where it's there in [3.3](https://github.com/silverstripe/silverstripe-framework/blob/3.3/admin/scss/_fonts.scss#L166)

minimalic:
Yes, all **four** `.font-icon-X-open-big` selectors are no more in the Sass file (admin/scss/_fonts.scss, 3.4 branch), so no css compilation problems.

kinglozzer:
We had this same problem with https://github.com/silverstripe/silverstripe-framework/pull/5006... why are commits going missing? 😢

UndefinedOffset:
I wonder if was a bad merge :S. I've seen commits appear to get overwritten during a merge, though it could be just my overfull head just thinking I did something and didn't lol.

tractorcow:
CSS commits are extremely problematic to merge up... I might put a moratorium on CSS changes that they cannot go into patch branches. :)

tractorcow:
Fixes:
- https://github.com/silverstripe/silverstripe-framework/pull/5769
- https://github.com/silverstripe/silverstripe-framework/pull/5770


Just to make sure there are no merge up dramas, I've created separate versions for each broken branch. :)

tractorcow:
Sorry, hope you don't mind my closing https://github.com/silverstripe/silverstripe-framework/pull/5663 (since only scss files were updated and not re-built).

minimalic:
No problem and thanks for fixes :)

axllent:
Are we please able to get a 3.4.1 release for this fix?

tractorcow:
@dhensby do you have time to do another release?

dhensby:
@tractorcow I'll talk to you about it "offline" :P

axllent:
@dhensby Any particular reason why it needs to be an "offline" discussion? Is it not possible to create a release with the fix?

[...]

PR #5663, opened 2016-06-07
Arrow icons (back button in edit forms)

minimalic:
Back button was no more visible in 3.4, issue #5657


tractorcow:
Replaced with https://github.com/silverstripe/silverstripe-framework/pull/5769, since this is missing the built dist css files.

PR #5666, opened 2016-06-08
BUG Fix tables being created with only ID field

tractorcow:


tractorcow:
codeclimate is being rediculous. =/

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-framework/pull/5646

PR #5675, opened 2016-06-09
Make Forgot Password link work without <base> tag present.

hailwood:
Because /Security/Login/Security/lostpassword is not a thing.

### DIFF
--- a/security/Security.php
+++ b/security/Security.php
@@ -129,6 +129,15 @@ class Security extends Controller implements TemplateGlobalProvider {
 	 */
 	private static $logout_url = "Security/logout";
 
+    /**
+     * The default lost password URL
+     *
+     * @config
+     *
+     * @var string
+     */
+    private static $lost_password_url = "Security/lostpassword";
+
 	/**
 	 * Get location of word list file
 	 *
@@ -650,10 +659,11 @@ class Security extends Controller implements TemplateGlobalProvider {
 	 * @param Member $member Member object associated with this link.
 	 * @param string $autoLoginHash The auto login token.
 	 */
-	public static function getPasswordResetLink($member, $autologinToken) {
-		$autologinToken = urldecode($autologinToken);
+	public static function getPasswordResetLink($member, $autologinToken)
+	{
+		$autologinToken      = urldecode($autologinToken);
 		$selfControllerClass = __CLASS__;
-		$selfController = new $selfControllerClass();
+		$selfController      = new $selfControllerClass();
 		return $selfController->Link('changepassword') . "?m={$member->ID}&t=$autologinToken";
 	}
 
@@ -1108,7 +1118,7 @@ class Security extends Controller implements TemplateGlobalProvider {
 	 * @return string
 	 */
 	public static function login_url() {
-		return self::config()->login_url;
+		return Controller::join_links(Director::baseURL(), self::config()->login_url);
 	}
 
 
@@ -1120,9 +1130,20 @@ class Security extends Controller implements TemplateGlobalProvider {
 	 * @return string
 	 */
 	public static function logout_url() {
-		return self::config()->logout_url;
+		return Controller::join_links(Director::baseURL(), self::config()->logout_url);
 	}
 
+    /**
+     * Get the URL of the logout page.
+     *
+     * To update the logout url use the "Security.logout_url" config setting.
+     *
+     * @return string
+     */
+    public static function lost_password_url()
+    {
+        return Controller::join_links(Director::baseURL(), self::config()->lost_password_url);
+    }
 
 	/**
 	 * Defines global accessible templates variables.
@@ -1133,6 +1154,7 @@ class Security extends Controller implements TemplateGlobalProvider {
 		return array(
 			"LoginURL" => "login_url",
 			"LogoutURL" => "logout_url",
+			"LostPasswordURL" => "lost_password_url"
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


dhensby:
Won't this break sites that are in a subfolder? 

Isn't the fix to add the base tag?

hailwood:
@dhensby Hmm good point on the subfolders.

We disable the base tag as we've run into problems with links generated by plugins etc quite often as the base tag is often not assumed and given that this is the only issue I've found with SilverStripe (at least in the frontend) with not having the base tag I'd much rather just have this work with/without the base tag than try to patch several plugins so enabling the base tag isn't really a suitable fix.

What about an [absolute url](http://api.silverstripe.org/3.3/class-Director.html#_absoluteURL) `Director::absoluteURL('Security/lostpassword', true)`.

If I did that would it be accepted?

tractorcow:
Change the template to a variable `   ... href="$LostPasswordURL.ATT" ... `.

In Security::get_template_global_variables() you can add a `"LostPasswordURL" => 'lost_password_url'` (makes the method available to all templates, like with LoginURL and LogoutURL).

Then in security something like

```php

private static $lost_password_url = "Security/lostpassword";

public static function lost_password_url() {
	return Controller::join_links(
		Director::baseURL(),
		static::config()->lost_password_url
	);
}
```

Or something. :)

We don't want to have to use absolute urls if we can help it.

In SilverStripe the base tag is mandatory, at the moment. The above will make the links work in all situations, though, by ensuring the link is at least root-relative.

flamerohr:
I'm more in favour of using the $controller that was given (which is the Security class)
So results in something like this, which uses the base url:
```
'<p id="ForgotPassword"><a href="'. $controller->Link('lostpassword') .'">'
```

hailwood:
@flamerohr & @tractorcow what about a middle ground based off https://github.com/silverstripe/silverstripe-framework/blob/master/security/Security.php#L655 -

```php
class Security {

public static function lost_password_url(){
    $selfControllerClass = __CLASS__;
    $selfController = new $selfControllerClass();
    return $selfController->Link('changepassword');
}

public static function getPasswordResetLink($member, $autologinToken) {
    $autologinToken = urldecode($autologinToken);
    return self::lost_password_url() . "?m={$member->ID}&t=$autologinToken";
}

public static function get_template_global_variables() {
    return array(
        "LoginURL" => "login_url",
        "LogoutURL" => "logout_url",
        "LostPasswordURL" => "lost_password_url"
    );
}

}
````

and then 

`'<p id="ForgotPassword"><a href="'. Security::lost_password_url() .'">'`

**Comment by sminnee** [comment]
Regarding base tags: although running SilverStripe without a base tag is "an unsupported configuration", I'd be happy with any change that lessens the pain of removing them.

Updating link generators to return root-relative links (i.e. those that start with `/`) that respect subfolders is a change we've made in other places.

See https://github.com/silverstripe/silverstripe-cms/blob/master/code/model/SiteTree.php#L400 for an example of this.

Root-relative links are better than fully absolute links as it makes things easier in, e.g., multi-server static publisher situations. Fully absolute links should only be used when building something specifically for a multi-domain system.

**Comment by hailwood** [comment]
So, do we have a vote on which solution we would like to go for? I'll implement which ever we go for.

--
Matt

**Comment by hailwood** [comment]
@tractorcow are we happy with the implementation now?

File: security/Security.php [hunk 1]
```diff
@@ -1123,6 +1124,18 @@ public static function logout_url() {
...
 
+	/**
+	 * Get the URL of the lost password page
+	 *
+	 * @return string
+	 */
+	public static function lost_password_url()
+	{
+		$selfControllerClass = __CLASS__;
+		$selfController      = new $selfControllerClass();
```
dhensby: `new static()` is nicer, I think?

tractorcow:
> I'm more in favour of using the $controller that was given (which is the Security class)
> So results in something like this, which uses the base url:
> '<p id="ForgotPassword"><a href="'. $controller->Link('lostpassword') .'">'

The issues I have with this are:
- It's not very symmetric with either login_url or logout_url, which are simply config strings.
- It's no longer possible to override this link. You can't, for instance, redirect to a custom password redirection page, as you can with login_url.

Apologies as I have not very much attention for open source issues at the moment. =(

security/Security.php
hailwood: I agree, I actually had that originally bit switched it out for the same method that has been used elsewhere in the file to keep things consistent (https://github.com/silverstripe/silverstripe-framework/blob/ff50b441393ef4950972891dc5b7fee7a9453395/security/Security.php#L656).

I had considered updating both of them but decided against it to keep this pull request focussed.

hailwood:
@tractorcow no problem,
That's not actually the method I have implemented.

I'll work this through with @dhensby and then ping you again when we're totally happy.

dhensby:
@tractorcow makes a good point that these links should be config variables

hailwood:
Hmm you're right of course.
Alright I can update that but it's not going to fix the original issue.

Should I make it a config value to match the other two and then update all three functions so they read as below which should fix the original issue, keep them all as a config value, and keep it consistent?

```php
public static function login_url() {
	return Controller::join_links(Director::baseURL(), self::config()->login_url);
}

public static function logout_url() {
	return Controller::join_links(Director::baseURL(), self::config()->logout_url);
}

public static function lost_password_url() {
	return Controller::join_links(Director::baseURL(), self::config()->lost_password_url);
}
````

--
By the way, I'll eventually squash these updates into one commit before we merge.

dhensby:
I think that solution is acceptable

hailwood:
Alrighty, I'll fire that through tonight after squashing.

hailwood:
And we're green @dhensby please review and if you're happy give tractor a ping.

--
Matt

dhensby:
OK  this is good. unfrotunately it needs to go into 3 branch

dhensby:
merged with ee326f6394ca11ffa5f2ac567fa9861be6b06859

PR #5681, opened 2016-06-10
NEW Use injector for MemberLoginForm fields

dhensby:


security/MemberLoginForm.php
```diff
@@ -110,15 +110,15 @@ public function __construct($controller, $name, $fields = null, $actions = null,
 		}
 
 		if(isset($backURL)) {
-			$fields->push(new HiddenField('BackURL', 'BackURL', $backURL));
+			$fields->push(HiddenField::create('BackURL', 'BackURL', $backURL));
 		}
```
SpiritLevel: BackURL, backURL, bAcKuRl, back_url ? I recall a recent discussion on the naming of this but I can't recall the conclusion.

dhensby: This is right

dhensby: And changing it is out of scope for this PR

PR #5686, opened 2016-06-13
Allow extensions for TextParser

### LINKED ISSUE #5685: Can't extend TextParser
I'm trying to extend a custom TextParser class that I've written, but due to the fact that TextParser doesn't call the Object constructor, it cannot support extensions.

Request to call parent constructor so this is possible.

robbieaverill:
* Call `Object::__construct` to make it extendable
* Resolves #5685


stevie-mayhew:
Thanks @robbieaverill

PR #5687, opened 2016-06-13
remove adding "original classes"

flamerohr:
The sitetree item had the class `disabled` or `edit-disabled` which were still present even after disabled was removed.

I'm not sure what the conflicting classes were, or if the conflict is still applicable?
Otherwise a blacklist like it has for statusflags would work as well


tractorcow:
Looks ok to me, but needs a rebase sorry. :)

Issue #4559, opened 2015-08-29
Simple class rename helper


sminnee:
In SilverStripe 4, a lot of classes are going to be renamed, as we introduce namespacing. Although a script that automatically fixes 100% of these is close enough to NP-hard to be discarded, a script that does the following will cover about 95% of the cases and make the refactoring work a lot less tedious.

I'd expect the script would do something like this:

 * Maintain a list of old class => new class
 * For parse each PHP files
 * Looks for simple references to altered classnames: e.g., `<literal>::` or `new <literal>`.
 * For each occurrence:
   * Replaces the class name with the new one, e.g. `StringField` => `DBString`
   * Adds the relevant `use` statement to the top of the file, e.g. `use SilverStripe\Model\FieldType\DBString;`.

My hope is that we can rely on an PHP parser packages to do this. Presumably this will be a separate packagist package, as we wouldn't want to download all that code for every SilverStripe user.

 - [x] Build tool - https://github.com/silverstripe/silverstripe-upgrader/
 - [ ] Upgrading notes - https://github.com/silverstripe/silverstripe-framework/pull/5695

sminnee:
Note: other common cases I've come across:

 * instanceof checks
 * assertInstanceOf calls with the class name given as a string literal
 * docblock references to classnames

camspiers:
I would have a look https://github.com/endel/galapagos which uses https://github.com/nikic/PHP-Parser

I have used these before and they are really great.

camspiers:
I don't have time to make the whole transform, but here is a start. It isn't at all finished, but it allows you to pass in a map from old class to new class, and it handles class extension replacement, interface implements replacement and static call replacement. With a couple of hours of dev work, it could be made to introduce use statements (for cleaner code), and made to support other cases like `new ClassName`, `ClassName::class` etc.

```php
<?php

use function galapagos\transform_with_visitors;

use PhpParser\Node;
use PhpParser\Node\Stmt;
use PhpParser\Node\Expr;

class RenameClass extends \PHPParser_NodeVisitorAbstract {
  protected $map;
  public function __construct($map) {
    $this->map = $map;
  }
  public function leaveNode(Node $node) {
    if ($node instanceof Stmt\Class_) {
      if ($node->extends !== null) {
        if (array_key_exists($node->extends->parts[0], $this->map)) {
          $node->extends->parts[0] = $this->map[$node->extends->parts[0]];
        }
      }

      if ($node->implements !== null) {
[...]

PR #5695, opened 2016-06-15
Documentation for https://github.com/silverstripe/silverstripe-upgrader/

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/issues/4559

### DIFF
--- a/docs/en/04_Changelogs/4.0.0.md
+++ b/docs/en/04_Changelogs/4.0.0.md
@@ -302,6 +302,26 @@ E.g.
 
 ## Upgrading
 
+### Automatically upgrading
+
+4.0 moves many classes to namespaces, and renames many methods.
+A tool is available to (semi-)automatically update your 3.x code to the new 4.0 names.
+
+See https://github.com/silverstripe/silverstripe-upgrader/ for the latest documentation on 
+how to upgrade your code. Typically the process will be:
+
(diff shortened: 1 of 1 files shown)


docs/en/04_Changelogs/4.0.0.md
```diff
@@ -302,6 +302,26 @@ E.g.
 
 ## Upgrading
 
+### General upgrading
+
+In order to better support migration of legacy modules and code, an automatic class upgrade tool
```
hafriedlander: I don't think this gets to the USP quickly enough. People won't really understand what this tool is without context.

Maybe something like:

4.0 moves many classes to namespaces, and renames many methods. A tool is available to (semi-)automatically update your 3.x code.

sminnee:
It's difficult to test this until we have some upgrade rules added to the framework. Let's get a simple case in and then confirm that these instructions work, before merging them.

hafriedlander:
@sminnee the problem with doing that is that as soon as there are some rules in framework, it means framework is breaking for everyone that hasn't run the upgrader. So I'd rather have the docs in first. But I'll managing merging so there's a PR for the ORM namespacing that these docs can be tested against.

tractorcow:
Rebased to resolve tests fixed upstream.

PR #5700, opened 2016-06-15
Moving a field between tabs, wrong variable name

Ironcheese:
The variable $field should be $content, since its added later as "$content"


dhensby:
Thanks

dhensby:
I've restarted the tests, I can't merge without green ticks :(

PR #5705, opened 2016-06-16
Use new "eslint-2" channel for CodeClimate

chillu:
Should avoid some false positives we're getting from validating with eslint 2
locally (npm dependency), but checking with eslint 1 remotely

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,6 +1,7 @@
 engines:
   eslint:
     enabled: true
+    channel: "eslint-2"
   scss-lint:
     enabled: true
   duplication:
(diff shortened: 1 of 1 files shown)


chillu:
For some reason this is causing a lot more warnings than when I'm running `eslint` locally in `framework`.  I've tried emulating CodeClimate runs more fully through https://github.com/codeclimate/codeclimate, but that times out on my Docker container at the moment. Parking this for now until I have more time to investigate.

tractorcow:
Lots of "unable to resolve path" :P

tractorcow:
Doh, needs a rebase @chillu

chillu:
Needs a rebase, but more importantly resolution on the "unable to resolve path" stuff

sminnee:
This is obsolete now

PR #5708, opened 2016-06-16
FIX: SS_ConfigStaticManifest_Parser failed to handle ::class syntax (fixes #5701)

### LINKED ISSUE #5701: ClassName::class incorrectly handled by SS_ConfigStaticManifest_Parser
Using `::class` for class name resolution breaks `SS_ConfigStaticManifest_Parser` in some situations:

```php
class Test
{
		$foo = Injector::inst()->get(Page::class);
	}
}
```

Put that in a file, run `dev/build?flush` and an error is triggered:

> Fatal error: Couldn\'t parse /path/to/mysite/code/Test.php when building config static manifest in /path/to/framework/core/manifest/ConfigStaticManifest.php on line 247

Changing it to `echo Page::class` doesn’t seem to cause an issue, so it’s probably that it’s expecting whitespace or something.

`SS_ConfigStaticManifest_Parser` has been removed in master, so this issue only applies to 3.x release lines

kinglozzer:
Targeted at 3 as it _technically_ adds new API. Happy to rebase against a different branch if preferred.

### DIFF
--- a/core/manifest/ConfigStaticManifest.php
+++ b/core/manifest/ConfigStaticManifest.php
@@ -199,6 +199,25 @@ class SS_ConfigStaticManifest_Parser {
 	}
 
 	/**
+	 * Get the previous token processed. Does *not* decrement the pointer
+	 *
+	 * @param bool $ignoreWhitespace - if true will skip any whitespace tokens & only return non-whitespace ones
+	 * @return null | mixed - Either the previous token or null if there isn't one
+	 */
+	protected function lastToken($ignoreWhitespace = true) {
+		// Subtract 1 as the pointer is always 1 place ahead of the current token
+		$pos = $this->pos - 1;
+		do {
+			if($pos <= 0) return null;
+			$pos--;
+			$prev = $this->tokens[$pos];
+		}
+		while($ignoreWhitespace && is_array($prev) && $prev[0] == T_WHITESPACE);
+
+		return $prev;
+	}
+
+	/**
 	 * Get the next set of tokens that form a string to process,
 	 * incrementing the pointer
 	 *
@@ -242,6 +261,14 @@ class SS_ConfigStaticManifest_Parser {
 			$type = ($token === (array)$token) ? $token[0] : $token;
 
 			if($type == T_CLASS) {
+				$lastToken = $this->lastToken();
+				$lastType = ($lastToken === (array)$lastToken) ? $lastToken[0] : $lastToken;
+
+				// Ignore class keyword in ClassName::class context
+				if ($lastType === T_PAAMAYIM_NEKUDOTAYIM) {
+					continue;
+				}
+
 				$next = $this->nextString();
 				if($next === null) {
 					user_error("Couldn\'t parse {$this->path} when building config static manifest", E_USER_ERROR);
--- a/tests/core/manifest/ConfigStaticManifestTest.php
+++ b/tests/core/manifest/ConfigStaticManifestTest.php
@@ -257,6 +257,16 @@ DOC;
 
 		$this->assertEquals($expected, $statics[':ss:test2']);
 	}
+
+	public function testParsingClassKeyword() {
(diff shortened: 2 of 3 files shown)


core/manifest/ConfigStaticManifest.php
```diff
@@ -199,6 +199,25 @@ protected function next($ignoreWhitespace = true) {
...
 
 	/**
+	 * Get the previous token processed. Does *not* decrement the pointer
+	 *
+	 * @param bool $ignoreWhitespace - if true will skip any whitespace tokens & only return non-whitespace ones
+	 * @return null | mixed - Either the previous token or null if there isn't one
+	 */
+	protected function lastToken($ignoreWhitespace = true) {
+		// Subtract 1 as the pointer is always 1 place ahead of the current token
+		$pos = $this->pos - 1;
```
kinglozzer: This is because the pointer (`$this->pos`) doesn’t actually match up to the current token being inspected, see comment added below:

```php
protected function next($ignoreWhitespace = true) {
	do {
		if($this->pos >= $this->length) return null;
		// If current position is 1, this will fetch the 1st token, *then* increment the pointer
		$next = $this->tokens[$this->pos++];
	}
	while($ignoreWhitespace && is_array($next) && $next[0] == T_WHITESPACE);

	return $next;
}
```

So `next()` will return the n-th token, but set the pointer to be n+1. I didn’t fancy refactoring this, as changing it affects pretty much everything else in the parser, plus it’s being removed in 4.x anyway.

core/manifest/ConfigStaticManifest.php
```diff
@@ -199,6 +199,25 @@ protected function next($ignoreWhitespace = true) {
...
+		do {
+			if($pos <= 0) return null;
+			$pos--;
+			$prev = $this->tokens[$pos];
```
kinglozzer: Decrementing `$pos` separately before inspecting the `$this->tokens` array for the same reason highlighted above

dhensby: You can do `--$pos` instead of `$pos--`; Fun fact: `--$pos` is slightly faster than `$pos--`!

dhensby:
It would be nice to get this into the 3.2 and up branches.

Can't this problem be solved by amending the `TokenRegexThingMaGiggy` rules? I don't like the idea that a T_CLASS should be ignored JUST because it is after a Paamayim Nekudotayim - we're coupling our business logic to the parser

kinglozzer:
> Can't this problem be solved by amending the `TokenRegexThingMaGiggy` rules?

`TokenisedRegularExpression`? Nope, as `SS_ConfigStaticManifest_Parser` doesn’t use it :(

> I don't like the idea that a T_CLASS should be ignored JUST because it is after a Paamayim Nekudotayim - we're coupling our business logic to the parser

Not sure where the coupling is here? The parser is just trying to determine the “current” class name: it looks for `T_CLASS` tokens, then fetches the next string (the actual class name) it can find and stores it for any statics it subsequently finds.

dhensby:
Ah, ok - then I was misremembering how this class worked (I was just in the process of checking it out to have a look) - otherwise I'm happy.

Is there no way to do the look behind without adding a new API (for the 3.2 and up branches)?

kinglozzer:
> Is there no way to do the look behind without adding a new API (for the 3.2 and up branches)?

I guess the new method could be simplified and added inline instead. I’ll take another look and target 3.2 instead

dhensby:
I'd like this refactor to still go into 3 branch, but it would be nice to get a patch fix for the others :)

tractorcow:
Any update on this? It'd be nice to get it merged.

kinglozzer:
PR for 3.2: https://github.com/silverstripe/silverstripe-framework/pull/5781

I’ve opened a second PR (I was going to reopen this one, but I force pushed 😞) for 3 with the slight refactor: https://github.com/silverstripe/silverstripe-framework/pull/5782. I can rebase that once the 3.2 commit is merged up if it makes life easier :)

PR #5717, opened 2016-06-20
API Add optimistic_live_connect to SS_Database

### LINKED ISSUE #4765: regression problem: database name may not contain underscore or permission issue?
Database names may not contain underscore or you get the following error

[User Error] Uncaught SS_DatabaseException: Couldn't run query: SHOW DATABASES LIKE 'porschefriends_' Access denied; you need (at least one of) the SHOW DATABASES privilege(s) for this operation

It is a problem as my provider does not allow me to choose the database name.

tractorcow:
Fixes #4765

@lienajp, @aledbrown  would you like to test this ?

### DIFF
--- a/model/connect/Database.php
+++ b/model/connect/Database.php
@@ -17,6 +17,15 @@ abstract class SS_Database {
 	protected $connector = null;
 
 	/**
+	 * In cases where your environment does not have 'SHOW DATABASES' permission,
+	 * you can set this to true. Then selectDatabase() will always connect without
+	 * doing databaseExists() check.
+	 *
+	 * @var bool
+	 */
+	private static $optimistic_connect = false;
+
+	/**
 	 * Get the current connector
 	 *
 	 * @return DBConnector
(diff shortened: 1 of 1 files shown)


aledbrown:
Hi

I can confirm once this was set to true that it worked on OVH:

private static $optimistic_live_connect = true;

I manually set that to true in Database.php but I'm guessing it might work from the config file too, will it?

Many thanks for the fix.

Aled

tractorcow:
Yeah, you should be changing this via config yml.

Just need someone to merge / review.

model/connect/Database.php
```diff
@@ -630,18 +641,24 @@ public function databaseList() {
...
-			if (!$create) {
-				if ($errorLevel !== false) {
-					user_error("Attempted to connect to non-existing database \"$name\"", $errorLevel);
-				}
-				// Unselect database
-				$this->connector->unloadDatabase();
-				return false;
+		// In case our live environment is locked down, we can bypass a SHOW DATABASE check
+		$canConnect =
+			(Director::isLive() && Config::inst()->get(get_class($this), 'optimistic_live_connect'))
```
dhensby: Why only on live?

tractorcow: Because it's optimistic_live_connect, not optimistic_connect.

Should I change it to remove the live-only exemption?

dhensby: I think so, otherwise we're just likely to see unexpected behaviour on live releases that were not seen on test sites (or problems on test sites that can't resolve)

tractorcow:
Rebased, and it's now optimistic_connect not optimistic_live_connect

dhensby:
much love for this! thanks

PR #5718, opened 2016-06-20
ENHANCEMENT: Added onBeforeRender extension point to form fields that call renderWith (fixes #4790)

### LINKED ISSUE #4790: $this->extend('onBeforeRender', $this); missing from several FormField Classes Field method.
For several Form Fields - e.g. `OptionSetField`, `ListboxSetField`, `GroupedDropdownField` there is no `Parent::Field` call in the Field method, this means that:

```php
    $this->extend('onBeforeRender', $this); 
```
which resides in `FormField::field` is never called, making it harder to create a DataExtension for these form fields. 

I wanted to create a module that decorates ListField, DropdownField and OptionSetField but that is now a little harder.

UndefinedOffset:
This pull request adds the onBeforeRender extension point to the following form fields, the other built in form fields that are not listed either don't seem to call renderWith and instead use raw html or already have the onBeforeRender extension point in the Field method,.

* CheckboxSetField
* OptionsetField
* TreedropdownField
* TreeMultiselectField
* UploadField
* ListboxField

### DIFF
--- a/forms/CheckboxSetField.php
+++ b/forms/CheckboxSetField.php
@@ -54,6 +54,8 @@ class CheckboxSetField extends OptionsetField {
 			'Options' => $this->getOptions()
 		));
 
+		$this->extend('onBeforeRender', $this);
+
 		return $this->customise($properties)->renderWith(
 			$this->getTemplates()
 		);
--- a/forms/ListboxField.php
+++ b/forms/ListboxField.php
@@ -102,6 +102,8 @@ class ListboxField extends DropdownField {
 			'Options' => new ArrayList($options)
 		));
 
+		$this->extend('onBeforeRender', $this);
+
 		return $this->customise($properties)->renderWith($this->getTemplates());
 	}
 
--- a/forms/OptionsetField.php
+++ b/forms/OptionsetField.php
@@ -83,6 +83,8 @@ class OptionsetField extends DropdownField {
 			'Options' => new ArrayList($options)
 		));
 
+		$this->extend('onBeforeRender', $this);
+
 		return $this->customise($properties)->renderWith(
 			$this->getTemplates()
 		);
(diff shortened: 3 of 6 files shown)


dhensby:
Is this an enhancement or a bug fix? Should it be in 3.2 release line or just 3?

UndefinedOffset:
Ya I'm not sure, given that it adds something previously not available I'd call it a feature. But then again its also possible it could be considered a bug because they probably should have had it when that was originally added. Not sure it's kind of a grey area ;)

dhensby:
Well, the intention of adding this hook to the parent class must have been to make it available to "all" field types, right?

If there are modules that rely on this hook, then it's a broken feature and therefore a bug imo.

Waiting for thoughts from @silverstripe/core-team

dhensby:
why is returning `parent::Field()` not working? It looks like it should from a brief inspection of the code?

I don't like having to add the hook to each child class that overrides the method

kinglozzer:
> Is this an enhancement or a bug fix? Should it be in 3.2 release line or just 3?

I think this is fine for a patch release.

dhensby:
OK - @kinglozzer agrees it's suitable to be a patch release.

I'd really prefer if we can get this working by calling `parent::Field()` instead - is that something that's realistic? I'm not sure why it'd make tests fail from first glance.

@UndefinedOffset are you ok to reopen this against 3.2 potentially with them calling the parent method?

UndefinedOffset:
@dhensby can do I'll make the changes, and open a new pull request. I'll also see if I can figure out that test failure that was spotted in #5780.

dhensby:
@UndefinedOffset much appreciated - if you can't get to the bottom of it, I'll probably accept this as a stop-gap solution and it'd be good to get it working "properly" later

PR #5726, opened 2016-06-22
fix(ArrayList/parseSortColumn): Fix sort with quotes

SilbinaryWolf:
When sorting with something like 'ArrayList->sort('"SiteTree"."Sort"')', ArrayList chokes and throws a 'Invalid sort() column' error. This removes the quotes so that DataList/ArrayList works interchangeably.

### DIFF
--- a/model/ArrayList.php
+++ b/model/ArrayList.php
@@ -360,6 +360,7 @@ class ArrayList extends ViewableData implements SS_List, SS_Filterable, SS_Sorta
 		}
 
 		// Parse column specification, considering possible ansi sql quoting
+		$column = str_replace('"', '', $column); // Remove quotes so sorts like '"SiteTree"."Sort"' work.
 		if(preg_match('/^"?(?<column>[^"\s]+)"?(\s+(?<direction>((asc)|(desc))(ending)?))?$/i', $column, $match)) {
 			$column = $match['column'];
 			if(empty($direction) && !empty($match['direction'])) {


model/ArrayList.php
```diff
@@ -360,6 +360,7 @@ protected function parseSortColumn($column, $direction = null) {
 		}
 
 		// Parse column specification, considering possible ansi sql quoting
+		$column = str_replace('"', '', $column); // Remove quotes so sorts like '"SiteTree"."Sort"' work.
 		if(preg_match('/^"?(?<column>[^"\s]+)"?(\s+(?<direction>((asc)|(desc))(ending)?))?$/i', $column, $match)) {
```
tractorcow: I'm interested why we need the quote removal above, when this regexp already detects quotes.

How does Table.Column sort syntax work on a list with no tables? I'm not able to see how this makes ArrayList sorting work where it didn't before. Maybe if there were some unit tests?

tractorcow: Perhaps you need another capture group to remove (and ignore) the table portion of the column sort? That could work.

SilbinaryWolf:
I have barely any regex experience so doing so isn't really a quick job for me. If I get some spare time, I'll look into adding this fix as practice. :)

dhensby:
@tractorcow can you provide some hints?

tractorcow:
@SilbinaryWolf if you write the unit tests I'll help with the code. :)

Please add some tests that demonstrate how you can sort arraylist with quoted columns.

tractorcow:
Nevermind, did it! https://github.com/silverstripe/silverstripe-framework/pull/5799

PR #5697, opened 2016-06-15
[WIP] Convert forms to use Bootstrap markup

clarkepaul:
Requires the same branch on CMS and AssetAdmin.

See:
 - https://github.com/silverstripe/silverstripe-cms/pull/1530
 - https://github.com/silverstripe/silverstripe-asset-admin/pull/195

PR #5728, opened 2016-06-22
Add page layout based on Bootstrap markup

clarkepaul:
I've simplified the steps styles but added additional responsiveness. It does look a bit different from what we had, and could probably be improved still but this should satisfy the issue of the layout being broken.

Requires https://github.com/silverstripe/silverstripe-framework/pull/5697 merged first.
Requires https://github.com/silverstripe/silverstripe-cms/pull/1533
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1526

Result:
pasted_image_22_06_16__7_12_pm


clarkepaul:
I've cleaned this stuff based on more recent changes and created a new PR here https://github.com/silverstripe/silverstripe-framework/pull/5832

PR #5729, opened 2016-06-22
Fixes issue 5188: X-Forwarded Proto

sekjal:
Removes X-Forwarded-Protocol in favour of the more standard X-Forwarded-Proto in the default Vary header config.
## To do before merge:
- [x] Update to include both `X-Forwarded-Protocol` and `X-Forwarded-Proto` in vary header
## Nice to have
- [x] Remove HTTP_FRONT_END_HTTPS reference from Director.php
- [ ] Throw deprecation warning if HTTP_X_FORWARDED_PROTOCOL is passed
- [ ] Update Vary header (but none of the deprecation changes) in the `3.5` branch.

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -19,4 +19,4 @@ HTTP:
     max-age: 0
     must-revalidate: "true"
     no-transform: "true"
-  vary: "Cookie, X-Forwarded-Protocol, User-Agent, Accept"
+  vary: "Cookie, X-Forwarded-Proto, User-Agent, Accept"


tractorcow:
Wouldn't the robust thing to do is vary on both?

sminnee:
I agree with @tractorcow; the Director.php code will recognise both and so the Vary header should capture this. This would be useful for the `3` branch as well.

The code also looks at `HTTP_FRONT_END_HTTPS` but I think this is an extremely non-standard value and so I would ignore this and remove it entirely from master.

We can probably throw a deprecation warning in SS4 if `X-Forwarded-Protocol` is used, but the Vary header should still contain both values.

chillu:
`HTTP_FRONT_END_HTTPS` seems to be used by MS, although the references are all ten years old: https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Common_non-standard_request_headers. +1 for removing it, since it can be configured.

Given this HTTP header needs to be thought about explicitly by implementors anyway based on their particular infrastructure (https://docs.silverstripe.org/en/3.2/developer_guides/security/secure_coding/#request-hostname-forgery), I think it's fine to merge this into 4.x with both `X-Forwarded-Proto` and `X-Forwarded-Protocol` in the `vary` header. I don't believe we need to throw deprecation warnings for values which aren't part of our API.

dhensby:
@sekjal are you happy to move this forward given the feedback?

sekjal:
Yes, thank you all for the thoughtful analysis. Sounds like a good plan

dhensby:
OK - I don't have permission to push to this PR so I'm opening a new one

Issue #4562, opened 2015-08-30
RFC-6: Namespacing


sminnee:
Author: @sminnee
Status: Pending review
Version: 0.1

## Purpose and outcome

Namespaces have been talked about for a while and we're starting to make pull requests that actually introduce namespaces, such as #4551. This RFC lays out a recommendation with some considerations and justification, so that we can proceed with merging PRs such as these into the master branch. Note that at this stage, the focus is *how* to do namespacing, rather than whether it is a good idea.

Historical discussion of namespacing is here: https://groups.google.com/forum/#!searchin/silverstripe-dev/namespaces/silverstripe-dev/i1voM50oFMk/KvidxVVTf38J

## Why do we need namespaces?

Apart from the absence of namespaces being kind of embarrassing in 2015, the chief practical reason for introducing them is to improve interoperability. It's my vision for the SilverStripe project that we are able to play more effectively with the rest of the PHP ecosystem, and in so doing reduce the extent to which developers have to "choose camps". Although there are a large number of challenges to overcome to make this vision real, it will be very difficult without namespaced code. So, in SilverStripe 4, let's make this happen.

Although it will be of some frustration to developers upgrading from SilverStripe 3, I think we can provide some help to help manage the frustration, and frankly, if we want the SilverStripe project to still be healthy and active in 2020 (which I do), I can't see us permanently avoiding this change.

So, boldly into the future we go!

## Consideration: PSR-4

Namespaces should be used such that the code can be transformed into a PSR-4 compliant layout as part of SilverStripe 4. Right now we don't *need* this because we have the manifest. Although the manifest is helpful, I would like us to incrementally reduce our reliance on it—it slows SilverStripe down and complicates our startup code (e.g. ErrorControlChain).

Because of this, I don't recommend putting all names in a single flat namespace.

Note: Actual adoption of a PSR-4 class loader is a separate topic, probably best attacked after we've refactored the system to allow more flexibility in code locations (e.g., push modules into the default `vendor/` directory). I'd expect that manifest vs. PSR-4 is something that could be selected on a module-by-module basis.

## Consideration: Framework Decomposition

One thing I would like to see is SilverStripe Framework broken down into a few small sub-modules that are included into `silverstripe/framework` as dependencies. The motivation is interoperability: you could, in principle, pull your SilverStripe data-model into another project (e.g. built on another framework). In practise, there's still work to be done to make this work nicely, but breaking into modules will encourage a less tightly coupled architecture.

[...]

PR #5734, opened 2016-06-23
API Apply Framework\ORM Namespace to model

### LINKED ISSUE #5421: Apply namespaces to the rest of the model subsystem
## Description

The model subsystem has already had namespaces introduced to field types, as part of PHP7 compliance.

Applying namespaces to the rest of the model subsystem is a logical next step.

Note that, as long as we don't yet apply namespaces to any of *the subclasses* of DataObject. #5419 needs to be implemented first.

## Exclusions

 * Updating documentation

## Related

 * Legacy reference (for SilverStripe internal team only, sorry): https://silverstripe.atlassian.net/browse/OSS-1855
 * table_name base implementation: https://github.com/silverstripe/silverstripe-framework/issues/5419

## Tasks

 * [x] Rename `model/` to `orm/`
 * [x] Move out DataObject classes to new `model/` dir (`Image`, `Changeset`)
 * [x] Run upgrader tool against `model/` folder, manually fix any issues which break tests
 * [x] Regression test CMS operation
 * [x] Post PR https://github.com/silverstripe/silverstripe-framework/pull/5734
 * [x] Post to dev list about upcoming change after PR raised but before merging https://groups.google.com/forum/#!topic/silverstripe-committers/t27D84dAdMs https://groups.google.com/forum/#!topic/silverstripe-dev/Ig3ODKBEaT4
 * [ ] Merge PR

Closed with #5734

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/issues/4562 for RFC-6 spec.

Closes #5421

This includes a bunch of other fixes:
- Moves all model classes to a new ORM folder to match the namespace.
- Explicitly require standard sql field methods in DBSchemaManager.
- Add a standard rename map to support the [upgrader](https://github.com/silverstripe/silverstripe-upgrader) tool.

Pull requests for other modules incoming.

- CMS: https://github.com/silverstripe/silverstripe-cms/pull/1534
- SiteConfig: https://github.com/silverstripe/silverstripe-siteconfig/pull/27
- Reports: https://github.com/silverstripe-labs/silverstripe-reports/pull/33
- Asset admin: https://github.com/silverstripe/silverstripe-asset-admin/pull/199
- Test Session: https://github.com/silverstripe-labs/silverstripe-testsession/pull/36 (also https://github.com/silverstripe-labs/silverstripe-testsession/pull/37 to maintain 3.x compat)
- Behat extension: https://github.com/silverstripe-labs/silverstripe-behat-extension/pull/117
- SQlite https://github.com/silverstripe-labs/silverstripe-sqlite3/pull/23

Fixes for other DB connectors will be resolved with a future issue. See https://github.com/silverstripe/silverstripe-framework/issues/5738


cli-script.php
```diff
@@ -1,5 +1,8 @@
 <?php
 
+use SilverStripe\ORM\DB;
+use SilverStripe\ORM\DataModel;
+
```
mikenz: Also need to remove the
`require_once("model/DB.php");` 
from this file

tractorcow: I've renamed it to "ORM/DB.php" as it currently is named.

mikenz:
Functionally there's no obvious issues for me. The upgrader worked great on all the modules I use and the unit tests all pass when I run them. Nice work!

mikenz:
SilverStripe\ORM\Versioning\Versioned::augmentWrite()
$manipulation uses the table name as the index now, not the class name.

```
framework$ git diff -w
diff --git a/ORM/Versioning/Versioned.php b/ORM/Versioning/Versioned.php
index 3b5297e..16e5dff 100644
--- a/ORM/Versioning/Versioned.php
+++ b/ORM/Versioning/Versioned.php
@@ -896,7 +896,7 @@ class Versioned extends DataExtension implements TemplateGlobalProvider {

             // If we're editing Live, then use (table)_Live instead of (table)
             if($this->hasStages() && static::get_stage() === static::LIVE) {
-                               $this->augmentWriteStaged($manipulation, $class, $id);
+                $this->augmentWriteStaged($manipulation, $table, $id);
             }
         }

```

tractorcow:
Oh, you're right @mikenz good spotting.

chillu:
- We need to adjust heaps of markdown docs - is there a follow up task for this? It's going to be at least 50% manual I'd say.
- We also need to adjust all the [lessons](http://www.silverstripe.org/learn/lessons/), which will be loads of fun (re-recording videos?). Where's that work tracked?
- I think given how many devs will *need* to read the 4.0 upgrading guide closely now, we should revise this document. It's grown quite organically over time, and doesn't have the changes with the biggest impact easily findable (references to the "upgrader" are hidden halfway through the doc)
- We list renamed classes in other parts of the 4.0 upgrading guide, why don't we do this for all the namespaces? It should at least point to `upgrade.yml`.
- PHPDoc follows the [same namespace rules](https://phpdoc.org/docs/latest/guides/types.html#class-name) as PHP which means we don't need to fix `@param`, `@return` and such. But: `{@link MyClassName}` can link to any class (not required to be listed in `use` statements), so we need to go through those. Could be deferred, as long as its tracked in a ticket?
- `i18nTextCollector` will collect fully qualified "namespaces" through `DataObject->provideI18nEntities()` now (for `{$this->class}.SINGULARNAME`. Is it too hacky to use `DataObject::$table_name` for it, assuming that core classes will get a short table name?
- Having an FQN in ModelAdmin URLs is a bit unwieldy, do we need to allow creation of an alias there? E.g. `admin/test/SilverStripe-FrameworkTest-Model-Employee` should be `admin/test/Employee`. Low priority of course.
- Needing to use FQN for string arguments and the `Config` system sounds a pain in the butt. Have aliases been discussed? Sorry, I'm a bit out of the loop on this. Just interested in thinking so far.

sminnee:
> i18nTextCollector will collect fully qualified "namespaces" through DataObject->provideI18nEntities() now (for {$this->class}.SINGULARNAME. Is it too hacky to use DataObject::$table_name for it, assuming that core classes will get a short table name

My vote goes with "too hacky". My preference would be a yml rewriter that we can pass of the lang files - can we do some kind of extract-transform-load step with transifex as a one-off?

chillu:
> can we do some kind of extract-transform-load step with transifex as a one-off

Yeah. It's also going to mess with everybody else's translations though (not just core), so ideally it's an optional part of the upgrader. Using `<MyModel>.SINGULARNAME` will be quite common in projects, because the key is generated by default on text collection. Or more explicitly mentioned in the upgrade guide as a manual task.

sminnee:
> Needing to use FQN for string arguments and the Config system sounds a pain in the butt. Have aliases been discussed? Sorry, I'm a bit out of the loop on this. Just interested in thinking so far.

The one thing that was rejected (I was one of the first people to say it was a bad idea) was just using short-names for yaml config entries, because it defeats the purposes of introducing namespacing.

So, if you were going to have aliasing, you'd need something like a 'use' statement, which would be yet-more-complexity for the config system.

My inclination would be to park it and wait for it to become an issue. I don't think we need to pre-emptively solve problems that might not be that big of an issue.

tractorcow:
> Yeah. It's also going to mess with everybody else's translations though (not just core), so ideally it's an optional part of the upgrader. Using <MyModel>.SINGULARNAME will be quite common in projects, because the key is generated by default on text collection. Or more explicitly mentioned in the upgrade guide as a manual task.

We'll need to rebuild a new set of master keys for 4.0 (treating 3.x and 4.x strings as independent) but that shouldn't be a problem. This is best done near the end of the namespacing. I'll create another story.

chillu:
Happy with the task priorities and split Sam has proposed. +1 from me on merge, I've discovered some missing `DBForeignKey` `use` statements which Damian fixed up now. Trialled this by running the upgrader on `frameworktest`, which has a few models and form fields interacting with them - looks great!

kinglozzer:
> Needing to use FQN for string arguments and the Config system sounds a pain in the butt. Have aliases been discussed? Sorry, I'm a bit out of the loop on this. Just interested in thinking so far.

I’ve encountered this recently, in PHP-land my personal preference is this:

```php
use Config;
use Vendor\Namespace\SubNamespace\ClassName;
// ...
Config::inst()->get(ClassName::class, 'foo');
```

Doesn’t save anything for one-off calls, but saves typing the namespace repeatedly for multiple calls :)

sminnee:
Also, `$this->config()->foo` or `ClassName::config()->foo` works for those classes using the `Configurable` trait.

PR #5735, opened 2016-06-24
FIX: Saving null values to the _versions table.

frankmullenger:
The isset() check does not return true when a value is set to null, so instead use array_key_exists().


frankmullenger:
@tractorcow implemented your array_key_exists() fix :-)

tractorcow:
It's possible the test that failed was relying on the incorrect behaviour... I'll have a look into it.

dhensby:
bump @tractorcow

tractorcow:
Pushed up a fix. :D

dhensby:
behat brokened :(

tractorcow:
Behat is fixed in another PR.

PR #5737, opened 2016-06-26
Add CIDR range functionality for SS_TRUSTED_PROXY_IPS

brettt89:
## To do before merge

 * [ ] Refactor logic into `ip_in_any_range($ipAddress, $rangeList)`
 * [ ] Add a unit test for `ip_in_any_range()` and/or `ip_in_range()`
 * [ ] Ensure exact-matches on ipv6 addresses (e.g. `::1`) will work.

## Nice to have

 * [ ] Parsing of IPv6 subnet ranges

### DIFF
--- a/core/Constants.php
+++ b/core/Constants.php
@@ -87,6 +87,25 @@ function stripslashes_recursively(&$array) {
 }
 
 /**
+ * Check if a given ip is in a network
+ * @param  string $ip    IP to check in IPV4 format eg. 127.0.0.1
+ * @param  string $range IP/CIDR netmask eg. 127.0.0.0/24, also 127.0.0.1 is accepted and /32 assumed
+ * @return boolean true if the ip is in this range / false if not.
+ */
+function ip_in_range( $ip, $range ) {
+	if ( strpos( $range, '/' ) == false ) {
+		$range .= '/32';
+	}
+	// $range is in IP/CIDR format eg 127.0.0.1/24
+	list( $range, $netmask ) = explode( '/', $range, 2 );
+	$range_decimal = ip2long( $range );
+	$ip_decimal = ip2long( $ip );
+	$wildcard_decimal = pow( 2, ( 32 - $netmask ) ) - 1;
+	$netmask_decimal = ~ $wildcard_decimal;
+	return ( ( $ip_decimal & $netmask_decimal ) == ( $range_decimal & $netmask_decimal ) );
+}
+
+/**
  * Validate whether the request comes directly from a trusted server or not
  * This is necessary to validate whether or not the values of X-Forwarded-
  * or Client-IP HTTP headers can be trusted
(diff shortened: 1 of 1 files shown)


hafriedlander:
I think we need to reverse this and have an "is this IP in this CIDR range" function - otherwise a big enough range (like 10.0.0.0/8) would end up creating an enormous list.

tractorcow:
Can you please update the documentation? Please see https://github.com/silverstripe/silverstripe-framework/blob/master/docs/en/02_Developer_Guides/09_Security/04_Secure_Coding.md#request-hostname-forgery for the current docs links.

Some phpdoc on the new method would be great too thanks. :)

tractorcow:
Also, +1 on what @hafriedlander said; Make it a check rather than an exhaustive allowed value generation. :P

dhensby:
Lot's of examples online how to check IP is in a CIDR declaration (here's one: https://gist.github.com/tott/7684443)

brettt89:
Have made changes as per comments.

core/Constants.php
```diff
@@ -87,6 +87,25 @@ function stripslashes_recursively(&$array) {
 }
 
 /**
+ * Check if a given ip is in a network
+ * @param  string $ip    IP to check in IPV4 format eg. 127.0.0.1
+ * @param  string $range IP/CIDR netmask eg. 127.0.0.0/24, also 127.0.0.1 is accepted and /32 assumed
+ * @return boolean true if the ip is in this range / false if not.
+ */
+function ip_in_range( $ip, $range ) {
+	if ( strpos( $range, '/' ) == false ) {
```
dhensby: If there's no slash, can't we do a basic comparison?

dhensby:
My only question now is how we deal with ipv6? Can we get some unit tests, too?

core/Constants.php
brettt89: Hi @dhensby , What do you mean by Basic Comparison? If there is no slash, it adds /32 onto the end (converts it to CIDR).

dhensby: If there's no slash, then it's not a range, it's a single IP, right? so we can do something like: `$ip == $range`

tractorcow:
According to googling, there is an option for ipv6.

http://stackoverflow.com/questions/7951061/matching-ipv6-address-to-a-cidr-subnet

brettt89:
Yeah, now I just need to figure out how to tie them both together.... Nicely.

brettt89:
@dhensby Where abouts would you put tests for constants?
@tractorcow I have been failing to try and combine the ipv4 and ipv6 check into a single function, do you have any suggestions as to how we could achieve this?

dhensby:
@brettt89 put them in seperate functions and then call the relevant one depending on whether the IP has `:` in it?

As for tests - I don't think we have any method for it :)

core/Constants.php
```diff
@@ -104,7 +123,12 @@ function stripslashes_recursively(&$array) {
 			if(SS_TRUSTED_PROXY_IPS === '*') {
 				$trusted = true;
 			} elseif(isset($_SERVER['REMOTE_ADDR'])) {
```
sminnee: I think you should refactor this to an `ip_in_any_range()` function, then you can add tests of that function.

```php
if(SS_TRUSTED_PROXY_IPS === '*') {
  $trusted = true;
} elseif(isset($_SERVER['REMOTE_ADDR'])) {
  $trusted = ip_in_any_range($_SERVER['REMOTE_ADDR'], SS_TRUSTED_PROXY_IPS);
}
```

Ensure there is a test for exact match on `::1`. [Handling of IPv6 subnets](https://www.ibm.com/support/knowledgecenter/STCMML8/com.ibm.storage.ts3500.doc/opg_3584_IPv4_IPv6_prefix_subnet_mask.html) wouldn't block merge but it might be nice.

core/Constants.php
sminnee: Doing a straight comparison will mean that `::1` can be added.

sminnee:
I've added some tasks to the top of the PR listing merge blockers and nice to haves.

sminnee:
I believe this is superceded by #6062. If @andrewandante and @dhensby can confirm that is the case, can you close this?

PR #5740, opened 2016-06-28
NEW: Add Requirements_Backend.nonce_format config

sminnee:
This config option can be set to ‘md5’ to use nonces based on the
md5-checksum of the content, rather than filemtime. This can be useful
when sessions aren’t blocking (e.g. when using database-backed sessions)

When sessions aren’t blocking, the CMS combined-files calls will
introduce race conditions that lead to unnecessary reloads of combined
files.

### DIFF
--- a/view/Requirements.php
+++ b/view/Requirements.php
@@ -987,7 +987,7 @@ class Requirements_Backend {
 			$mtimesuffix = "";
 			$suffix = '';
 			if($this->suffix_requirements) {
-				$mtimesuffix = "?m=" . filemtime($filePath);
+				$mtimesuffix = "?m=" . $this->generateNonce($filePath);
 				$suffix = '&';
 			}
 			if(strpos($fileOrUrl, '?') !== false) {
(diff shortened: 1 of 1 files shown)


view/Requirements.php
```diff
@@ -1005,6 +1005,16 @@ protected function path_for_file($fileOrUrl) {
 		}
 	}
 
+	protected function generateNonce($filePath) {
+		if(!file_exists($filePath)) return 'notfound';
```
tractorcow: `Director::fileExists` already checks this above.

view/Requirements.php
```diff
@@ -1005,6 +1005,16 @@ protected function path_for_file($fileOrUrl) {
...
+
+		if(Config::inst()->get('Requirements_Backend', 'nonce_format') === 'md5') {
```
tractorcow: The config which turns this on is `suffix_requirements`. Should we use the same naming scheme for the format? E.g. `suffix_format`?

tractorcow: Maybe add a default value on Requirements_Backend above, 

`private static $suffix_format = 'filemtime';`

view/Requirements.php
```diff
@@ -1005,6 +1005,16 @@ protected function path_for_file($fileOrUrl) {
 		}
 	}
 
+	protected function generateNonce($filePath) {
```
tractorcow: phpdoc please. :)

kinglozzer:
@sminnee SS4 appears to use `sha1`, do you think this is worth re-submitting against `4`?

sminnee:
Nah I think I'll cancel this

Issue #5420, opened 2016-04-29
Apply namespaces to Security subsystem


sminnee:
Depends on #5419.

The Security subsystem provides a lot of the out of the box DataObject classes in framework, and so it would be a good place to start the work of introducing namespaces to classes and determining their impact on the ORM. Once #5419 is introduced, we should be able to achieve this without frustrating and unnecessary table renames.

Tasks:
 - [x] Apply upgrader
 - [x] Resolve test regressions
 - [x] Post PR https://github.com/silverstripe/silverstripe-framework/pull/5741
 - [x] Post to silverstripe-committers list https://groups.google.com/forum/#!topic/silverstripe-dev/Ig3ODKBEaT4
 - [ ] Merge PR

Firesphere:
Except for the models, I've done this here:
https://github.com/CasaLaguna/silverstripe-security

(WIP)

tractorcow:
Needs peer review at https://github.com/silverstripe/silverstripe-framework/issues/5741

PR #5741, opened 2016-06-29
API Apply SilverStripe\Security namespace

tractorcow:
Rebased on top of https://github.com/silverstripe/silverstripe-framework/pull/5734; This is to be merged subsequent to that.

Closes https://github.com/silverstripe/silverstripe-framework/issues/5420

This includes a change to FormField() so that the default class property is taken from the basename of the FQN classname, rather than the full class, by default.

Associated pull requests:

 - CMS https://github.com/silverstripe/silverstripe-cms/pull/1536
 - SiteConfig https://github.com/silverstripe/silverstripe-siteconfig/pull/28
 - Reports https://github.com/silverstripe-labs/silverstripe-reports/pull/35
 - Asset Admin https://github.com/silverstripe/silverstripe-asset-admin/pull/201
 - Behat Extension https://github.com/silverstripe-labs/silverstripe-behat-extension/pull/118
 
 
No changes to test session module this time. :D


ORM/ArrayList.php
```diff
@@ -1,4 +1,14 @@
...
+
+namespace SilverStripe\ORM;
+
+use ViewableData;
+use ArrayData;
+use ArrayIterator;
+use Debug;
+use InvalidArgumentException;
+use LogicException;
+
```
Firesphere: Wouldn't it be more logical to use `\ViewableData` as namespace?

ORM/Connect/MySQLSchemaManager.php
```diff
@@ -454,8 +459,8 @@ public function set($values) {
 		//DB::requireField($this->tableName, $this->name, "enum('" . implode("','", $this->enum) . "') character set
 		//utf8 collate utf8_general_ci default '{$this->default}'");
 		$valuesString = implode(",", Convert::raw2sql($values['enums'], true));
-		$charset = Config::inst()->get('MySQLDatabase', 'charset');
-		$collation = Config::inst()->get('MySQLDatabase', 'collation');
+		$charset = Config::inst()->get('SilverStripe\ORM\Connect\MySQLDatabase', 'charset');
+		$collation = Config::inst()->get('SilverStripe\ORM\Connect\MySQLDatabase', 'collation');
 		return "set($valuesString) character set {$charset} collate {$collation}" . $this->defaultClause($values);
```
Firesphere: After seeing a lot of fully namespaced items. I think, it would add value to be able to call Config without using full namespace. Do you see an option there?

ORM/DataObject.php
```diff
@@ -381,7 +410,7 @@ public function __construct($record = null, $isSingleton = false, $model = null,
 		// Identify fields that should be lazy loaded, but only on existing records
 		if(!empty($record['ID'])) {
 			$currentObj = get_class($this);
-			while($currentObj != 'DataObject') {
+			while($currentObj != 'SilverStripe\ORM\DataObject') {
```
Firesphere: Should be used cased? `$currentObj instanceOf DataObject` maybe?

ORM/DataObject.php
```diff
@@ -589,7 +618,7 @@ public function defineMethods() {
 			}
 		}
 
-		if($this->class == 'DataObject') return;
+		if($this->class == 'SilverStripe\ORM\DataObject') return;
```
Firesphere: Shorthand, ouch, but okay in this case. Same as above, is_a would maybe be a better way to do it?

ORM/DataObject.php
```diff
@@ -677,7 +706,8 @@ public function i18n_pluralise($count, $prependNumber = true) {
 	 */
 	public function singular_name() {
 		if(!$name = $this->stat('singular_name')) {
-			$name = ucwords(trim(strtolower(preg_replace('/_?([A-Z])/', ' $1', $this->class))));
+			$reflection = new \ReflectionClass($this);
```
Firesphere: The use-case is missing?

ORM/DataObject.php
```diff
@@ -677,7 +706,8 @@ public function i18n_pluralise($count, $prependNumber = true) {
...
+			$name = ucwords(trim(strtolower(preg_replace('/_?([A-Z])/', ' $1', $reflection->getShortName()))));
```
Firesphere: Not new, but too much nesting of methods. Splitting out would be more readable.

ORM/Connect/MySQLSchemaManager.php
kinglozzer: My suggestion would be to add a use statement and then use `ClassName::class` instead (see https://github.com/silverstripe/silverstripe-framework/pull/5734#issuecomment-229897936). Not sure if this is coming from the automated upgrade tool or not though?

Firesphere:
Double-backslashes are not necessary in single quoted strings. I would suggest either do one, or the other. At the moment it's a mix.

I've not been able to read it all the way through, since it's pretty much midnight ;)

ORM/DataObject.php
dhensby: This is going through the inheritance of  class, I think, so `isntanceof` isn't what we want

ORM/ArrayList.php
tractorcow: I don't understand... you mean put ViewableData into a namespace of the same name, so `SilverStripe\ViewableData\ViewableData`?

This pull request only covers Security namespace for the time being.

tractorcow: Sorry, this commit was merged in another pull request... I'm about to push up a rebased version and this change will disappear. :)

This PR is really intended to cover Security namespacing only.

ORM/Connect/MySQLSchemaManager.php
tractorcow: We can't use ClassName::class unless we bump up the minimum php version to 5.6.

We can't use short name in config because then we'd run into ambiguous class name problems, which is the whole point of using namespaces in the first place.

A better solution is use the `Configurable` trait, and then you can just do ClassName::config()->my_config to access it.

ORM/DataObject.php
tractorcow: $currentObj isn't actually an object, it's a string, so no it won't work.

ORM/DataObject.php
tractorcow: Why, it's just a string comparison. is_a would return true for subclasses, which is not what we want.

ORM/DataObject.php
tractorcow: It's a FQN so it's not missing. We don't strictly require use statements; Using the FQN is fine.

tractorcow:
> Double-backslashes are not necessary in single quoted strings. I would suggest either do one, or the other. At the moment it's a mix.

I think we should always escape slashes, since according to http://php.net/manual/en/language.types.string.php#language.types.string.syntax.single

> To specify a literal backslash, double it (\\).

So technically I would say they are necessary, if we require good coding style.

ORM/Connect/MySQLSchemaManager.php
kinglozzer: `ClassName::class` is PHP 5.5 (http://php.net/manual/en/language.oop5.basic.php#language.oop5.basic.class.class), but I prefer the `Configurable` approach anyway when possible :)

sminnee:
OK I've done a manual code review; I think it's good to merge if the tests pass.

tractorcow:
They don't though lol. And they won't until the cms pr is merged. :)

sminnee:
The behat tests are failing too - the behat support module is hard-coded to look for the Group class.

Do we have a behat PR as well?

PR #5743, opened 2016-06-29
Revert deprecation of Object::get_static()

ntd:
Inside a theme you cannot add code to the controller so I am providing
some sort of customization by leveraging `Object::get_static`, e.g.:

```yaml
SSViewer:
  label_class: col-xs-3
```

```html
  <label class="$get_static(SSViewer,label_class)">...</label>
```

I think it is quite handy and not being able to do such things in 4.0
should be considered a regression.

### DIFF
--- a/core/Object.php
+++ b/core/Object.php
@@ -393,11 +393,7 @@ abstract class Object {
 		return $default;
 	}
 
-	/**
-	 * @deprecated
-	 */
 	public static function get_static($class, $name, $uncached = false) {
-		Deprecation::notice('4.0', 'Replaced by Config#get');
 		return Config::inst()->get($class, $name, Config::FIRST_SET);
 	}
 


kinglozzer:
Hmm, if we’re going to support this I think it would be better to make `Config` implement `TemplateGlobalProvider`. It might look something like:

```php
class Config implements TemplateGlobalProvider {

    // ...

    public static function get_template_global_variables() {
        return array('Config' => 'get');
    }
}
```

```html
<label class="{$Config('SSViewer', 'label_class')}">...</label>
```

dhensby:
I think that is definitely an unintended use of `get_static`

@kinglozzer's suggestion is a good one but one I'd suggest you implement yourself in your own code

ntd:
> ...I'd suggest you implement yourself in your own code

I cannot in a theme... or there is something huge I missed?

In a module, where I can happily add custom PHP code, this could be accomplished in many ways but in a theme AFAIK I do not have access to PHP.

dhensby:
Your theme could have a dependency declared in composer to make this available in the template.

ntd:
Yes, this is what I'm doing right now. The dependency is optional and the helper module is based on `SiteConfig` but the principle is the same.

I still think a configuration layer for themes would be a good addition but this would deserve a more serious propose.

PR #5744, opened 2016-06-29
Update GridFieldAddExistingAutocompleter.php

jonners99:
When saving by hitting enter, was receiving an error message 'Can't handle action "find"'

Traced this back to GridFieldAddExistingAutocompleter setting a 'find' action in getHTMLFragments (line 107) and then not declaring it on line 140


jonners99:
I don't really see how this code change makes the Restore option disappear from the CMS. It appears to be missing even without my change.

Did I send a pull request against the wrong branch?

dhensby:
I restarted the test and it's passed

PR #5751, opened 2016-06-30
BUG Don't hard-code folder into treedropdownfield search hint

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1518

### DIFF
--- a/forms/TreeDropdownField.php
+++ b/forms/TreeDropdownField.php
@@ -217,11 +217,12 @@ class TreeDropdownField extends FormField {
 		Requirements::css(FRAMEWORK_DIR . '/thirdparty/jquery-ui-themes/smoothness/jquery-ui.css');
 		Requirements::css(FRAMEWORK_DIR . '/client/dist/styles/TreeDropdownField.css');
 
-		if($this->showSearch) {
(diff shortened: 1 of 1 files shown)


forms/TreeDropdownField.php
```diff
@@ -217,11 +217,12 @@ public function Field($properties = array()) {
...
-		if($this->showSearch) {
-			$emptyTitle = _t('DropdownField.CHOOSESEARCH', '(choose folder)', 'start value of a dropdown');
-		} else {
-			$emptyTitle = _t('DropdownField.CHOOSE', '(Choose)', 'start value of a dropdown');
-		}
+		$item = DataObject::singleton($this->sourceObject);
+		$emptyTitle = _t(
+			'DropdownField.CHOOSE_MODEL',
+			'(Choose {name})',
+			['name' => $item->i18n_singular_name()]
```
dhensby: Did we make a decision on array declaration style? My understanding was to stick with full `array()` syntax

tractorcow: I think that just applied to existing code; New code is generally being added with the new syntax. :)

dhensby: and a tiny piece of me dies inside

PR #5754, opened 2016-06-30
Wrong date used for <pubDate> in RSSFeed.ss

mschilder:
The default RSSFeed.ss uses $Date to populate <pubDate> in the template. As the latest SilverStripe Blog module uses $PublishDate instead of $Date, it falls back to $Created.

### DIFF
--- a/templates/RSSFeed.ss
+++ b/templates/RSSFeed.ss
@@ -11,7 +11,7 @@
 			<title>$Title.XML</title>
 			<link>$AbsoluteLink</link>
 			<% if $Description %><description>$Description.AbsoluteLinks.XML</description><% end_if %>
-			<% if $Date %><pubDate>$Date.Rfc822</pubDate>
+			<% if $PublishDate %><pubDate>$PublishDate.Rfc822</pubDate>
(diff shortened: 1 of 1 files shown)


tractorcow:
RSSFeed is a generic feed api. It has nothing intrinsically to do with Blog module. It's not a wrong date, it's just not a 1-to-1 map for blog.

This fix would break all reliances on the 'Date' rss feed where it didn't match strictly to 'PublishDate'.

If you have a look at RSSFeed_Item you can see this is the method signature.

```php
/**
 * Create a new RSSFeed entry.
 */
public function __construct($entry, $titleField, $descriptionField, $authorField) {
	// ...
}
```

What we really need is a $dateField property so that blog can specify the field to use for date, in this case, 'PublishedDate'. The default should be left at 'Date' to preserve backwards compatibility.

How does that sound?

mschilder:
Yes, you're right that sounds perfect!

dhensby:
I thought this had already been fixed

yep: https://github.com/silverstripe/silverstripe-blog/pull/403

tractorcow:
I was about to link to that PR... I just saw it. :D

PR #5752, opened 2016-06-30
FIX Parse shortcodes in template helpers

dhensby:
Fixes #5661
Fixes #1381

See #5653 

Basically back-porting the approach from the above PR.

Use `RAW` to parse the shortcodes and rely on RAW for all the other helper methods.

PR #5760, opened 2016-07-01
[3.2] FIX Parse shortcodes in template helpers

dhensby:
see #5752


dhensby:
If this gets accepted as a patch, then we need to adjust `DBField::JSON` when it's merged up into 3.4

PR #5761, opened 2016-07-01
[3.2] FIX Parse shortcodes in template helpers

dhensby:
see #5752


dhensby:
If this gets accepted as a patch, then we need to adjust DBField::JSON when it's merged up into 3.4

tractorcow:
Can you please add a bit of unit test coverage for the new RAW() behaviour, as well as the fix to shortcode parsing?

dhensby:
unit test coverage added :)

`HTMLVarchar` has no test coverage currently, which is a shame - I don't have time to add it now, I'm afraid

Issue #5764, opened 2016-07-04
Time format helper in profile incorrect


flamerohr:
`hh` describes `(00 through 23)` which is actually `(00 through 12)`
`HH` works for the correct `(00 through 23)`

I haven't looked into too much of this yet, but would `hh` = `(00 through 23)` be the correct desired effect, or would it be better to add the description for `HH` and correct `hh` to `(00 through 12)`?
<img width="858" alt="screen shot 2016-07-04 at 2 22 01 pm" src="https://cloud.githubusercontent.com/assets/1064889/16549370/ddfcc318-41f2-11e6-9b48-2a4fb090fe83.png">

tractorcow:
Yep we should fix the docs, and maybe document both HH and hh separately.

I also suggest to separate the time and date format helpers, since it doesn't make much sense to show both for both fields.

PR #5766, opened 2016-07-04
Fix MemberDatetime helper description, and shifted them to templates

flamerohr:
This was in response to https://github.com/silverstripe/silverstripe-framework/issues/5764


tractorcow:
Something weird is going on with these tests.

dhensby:
They all pass?

dhensby:
This is a great fix. Thanks

PR #5775, opened 2016-07-05
FIX calling $record->write() on ListboxField::saveInto breaks other 3rd party fields that write to an UnsavedRelationList.

SilbinaryWolf:
FIX calling $record->write() on ListboxField::saveInto breaks other 3rd party fields that write to an UnsavedRelationList.

Had a quick look at the Git history, Ingo Schommer added that fix in 2012. I'm guessing it was needed in 3.0 but the need for it went away when 3.1 was released in 2013.

(https://www.silverstripe.org/blog/3.1.0-goes-stable/)


dhensby:
Thanks. Good find

PR #5777, opened 2016-07-06
API mute frontend debugging by default

tractorcow:
No more console spam by default, please!


dhensby:
I think I'd probably have it on by default plus add a console log explaining how to turn it off.

dhensby:
Test failure due to NPM issues: https://github.com/npm/npm/issues/13284

tractorcow:
I've changed it to be on by default (thus no change)

PR #5780, opened 2016-07-07
Fix(ListboxField):  Use parent::Field() so that $this->extend('onBeforeRender', $this); gets called.

SilbinaryWolf:
Use parent::Field() so that $this->extend('onBeforeRender', $this); gets called.

### DIFF
--- a/forms/ListboxField.php
+++ b/forms/ListboxField.php
@@ -102,7 +102,7 @@ class ListboxField extends DropdownField {
 			'Options' => new ArrayList($options)
 		));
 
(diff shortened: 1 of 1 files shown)


dhensby:
This is a good fix, but it's causing regressions as can be seen in the test failure.

UndefinedOffset:
@dhensby I had opened a very similar pull request in #5718 though my pull request covers more fields, most of the ones that didn't I think, spotted this issue thanks to display_logic ;). I was also in the process of updating it this am to cover ListboxField as well.

dhensby:
OK, closing in preference for #5718

PR #5781, opened 2016-07-07
FIX: SS_ConfigStaticManifest_Parser failed to handle ::class syntax (fixes #5701)

### LINKED ISSUE #5701: ClassName::class incorrectly handled by SS_ConfigStaticManifest_Parser

kinglozzer:


dhensby:
Thanks @kinglozzer

PR #5783, opened 2016-07-07
BUGFIX: Changed form fields that call renderWith in Field() to call parent::Field() instead

UndefinedOffset:
Per request in #5718 this pull request is to make similar changes to what was in that pull request to the below classes. The main difference with this pull is that ``parent::Field()`` or ``FormField::Field()`` is used instead of calling renderWith and adding the extension point at each of those classes. The reason for using ``FormField::Field()`` in some instances is to jump over the ``parent::Field()`` which in some instances is causing issues like the [failed tests here](https://travis-ci.org/silverstripe/silverstripe-framework/jobs/142966336).

Affected classes:
* CheckboxSetField
* OptionsetField
* TreeDropdownField
* TreeMultiselectField
* UploadField
* ListboxField

### DIFF
--- a/forms/CheckboxSetField.php
+++ b/forms/CheckboxSetField.php
@@ -54,9 +54,7 @@ class CheckboxSetField extends OptionsetField {
 			'Options' => $this->getOptions()
 		));
 
-		return $this->customise($properties)->renderWith(
-			$this->getTemplates()
-		);
+		return FormField::Field($properties);
 	}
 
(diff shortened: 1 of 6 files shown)


UndefinedOffset:
@dhensby per request :)

dhensby:
Ah-ha - that's why the other PR didn't work.

Thanks a ton for this!

PR #5785, opened 2016-07-07
Update 02_Relations

robbieaverill:
* Code formatting on some variable names
* Add example for removing scaffolded parent relation fields


dhensby:
thanks

Issue #5784, opened 2016-07-07
Email setters should use consistent nomenclature (and other cleanup)


patricknelson:
I've only found one method where this is an issue, however: It looks like every single email header that has a method is using the "set" prefixed nomenclature except for the method `->replyTo()` on the `Email` object, see here: https://github.com/silverstripe/silverstripe-framework/blob/master/email/Email.php#L256

Should this be deprecated in favor of a proper `->setReplyTo()` method? We ended up missing this since it wasn't prefixed as expected and ended up putting `->addCustomHeader('Reply-To', $replyTo)` in several places in our code before really diving deeper into why on earth this isn't already a method. Turns out it is!

patricknelson:
Also just wanted to point out that there's plenty of other cleanup opportunity here in this class, including documentation and redundant methods. I'm going to issue a quick PR against `3` to as a proof of concept and will itemize my changes there.

tractorcow:
👍  yes please.

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-framework/pull/5790

PR #5790, opened 2016-07-08
FIX for #5784: Email API improvements, documentation cleanup.

patricknelson:
PR and PoC to address #5784

- Added ->setReplyTo(), deprecated ->replyTo() for API consistency. My original reason for opening #5784. Not sure if this should be set for deprecation on next minor version (3.5) or major (4.0). Please let me know.
- Revamping, fixing, and enhancing internal Email API documentation. There were many instances of `@param` where it should have been `@var`.
- Relocated the `::mailer()` and `::set_mailer()` methods, since it's customary to have the `_construct()` method as first in the list and to help organize the properties together at the very top.
- Provided more detail in externally facing online documentation.
- Other class variable inconsistencies, such as initialization with `null` in some cases but not others (since it's implicit anyway). Normalized that down to just not having the redundant `null`.
- Clarified variable types after doing some digging, added better explanations.
- Simplified code and brought up-to-date with latest standards. For example, `::is_valid_address()` is now completely redundant, since SilverStripe framework requires PHP 5.3+ and the `::validEmailAddress()` method is already setup and using `filter_var()` (available since PHP 5.2) which is itself validating against RFC 822 anyway. Also, that method no longer needs to check for the function existence because of SilverStripe's minimum PHP requirements.

Most important change that might not be suitable for `3` (not sure) since this is a possible bug fix, but more restrictive in API (technically):
- Performed a strict type comparison in the constructor, since the intent was to check `null` variables 
but was using loose type checking. So if `$from` were actually an empty string, it would still end up calling `$this->setFrom()`.

So, please let me know about the deprecation notice and the last bullet here about the strict type checking. Thanks!


email/Email.php
```diff
@@ -25,89 +25,76 @@
...
-
-	/**
-	 * Get the mailer.
-	 *
-	 * @return Mailer
-	 */
-	public static function mailer() {
-		return Injector::inst()->get('Mailer');
-	}
-
```
patricknelson: Actually relocated, see below.

email/Email.php
```diff
@@ -389,15 +496,15 @@ protected function parseVariables($isPlain = false) {
...
+	 * @return bool
 	 */
 	public static function validEmailAddress($address) {
-		if (function_exists('filter_var')) {
-			return filter_var($address, FILTER_VALIDATE_EMAIL);
-		} else {
-			return preg_match('#^([a-zA-Z0-9_+\.\-]+)@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.)'
-				. '|(([a-zA-Z0-9\-]+\.)+))([a-zA-Z]{2,4}|[0-9]{1,3})(\]?)$#', $address);
-		}
+		return filter_var($address, FILTER_VALIDATE_EMAIL) !== false;
```
dhensby: I'm slightly concerned that this counts as an API change more than a bug fix - do we know how this will effect valid/invalid email addresses getting through?

email/Email.php
```diff
@@ -609,37 +716,17 @@ public static function bcc_all_emails_to($emailAddress) {
...
-
-		return preg_match("!^$addr_spec$!", $email) ? 1 : 0;
-	}
-
-	/**
-	 * Encode an email-address to protect it from spambots.
-	 * At the moment only simple string substitutions,
-	 * which are not 100% safe from email harvesting.
+		Deprecation::notice('3.5', 'Use the "validEmailAddress" method instead');
+		return static::validEmailAddress($email);
```
dhensby: Same concern as above about API change vs bug fix

email/Email.php
```diff
@@ -184,103 +204,181 @@ public function setBounceHandlerURL($bounceHandlerURL) {
...
+		return $this;
+	}
+
+	/**
+	 * @param string $email
+	 * @return $this
+	 * @deprecated 3.5 Use the "setReplyTo" method instead
 	 */
 	public function replyTo($email) {
+		Deprecation::notice('3.5', 'Use the "setReplyTo" method instead');
```
dhensby: Should be 4.0, from what I think the general consensus is. though really it's academic as this entire class will hopefully be re-written for 4

dhensby:
Hah - test failure because `filter_var` actaually correctly identifies `example.com` as an invalid host... Test needs upating.

email/Email.php
patricknelson: I'll reply in the regular comment thread since this may be clobbered a bit by a new rewrite/push.

patricknelson:
For the `::validEmailAddress()` and `::is_valid_address()` methods: This may be a bit of an API change since (as you noted later on) the unit test fails. However, it fails because now it actually returns `true` as documented by the docblock. This to me is definitely more of a bug fix as, to my mind, since it shouldn't ever need to return the same email you just passed in; that's just an artifact of how `filter_var` works and that more if an internal implementation of how that validation is happening and shouldn't be transparent (as far as I'm concerned) and offers no utility here anyway. Obviously it's not totally intentional anyway since I think the unit test was originally just written so it'd pass. 

Also, regarding if `::is_valid_address()` also being a bit of an API change; yeah, at least with it being deprecated here. Also, if it really holds true (not 100% of course, haven't tested it with great rigor) that this method is truly RFC 822 compliant, then it's completely redundant as noted in my initial comment. So I figured this would be cleaner :+1:  

Finally, I removed the `function_exists` code as well since it's not necessary if we really require the SilverStripe framework to be compatible with PHP 5.3. Part of the advantage of supporting newer versions anyway.

patricknelson:
Also: Do you unit test deprecated methods? Especially if they're just non-changing/transparent pass-through's to other methods.

patricknelson:
Also x 2 (sorry for the ADD commentary): What do you think about the changes in `__construct()`? That may be a little edgy there too. Wanted to call that out.

dhensby:
> Also: Do you unit test deprecated methods? Especially if they're just non-changing/transparent pass-through's to other methods.

no

> What do you think about the changes in __construct()? That may be a little edgy there too. Wanted to call that out.

I have no problem with that change myself

> This may be a bit of an API change since (as you noted later on) the unit test fails. However, it fails because now it actually returns true as documented by the docblock.

I understand - this is tricky, is the API as documented or as it behaves? It's definitely a lower risk to leave it behaving as is because devs would *have* to make it work the the current return type and you couldn't blame them for relying on it either.

> Also, if it really holds true (not 100% of course, haven't tested it with great rigor) that this method is truly RFC 822 compliant, then it's completely redundant as noted in my initial comment

I would note that `filter_var` does not support non-latic character sets and this is a problem for SS 4, which will need to support those.

@silverstripe/core-team I need some opinions on this please

patricknelson:
Quick note on your point here:

> I would note that filter_var does not support non-latic character sets and this is a problem for SS 4, which will need to support those.

Do you happen to have a source? I'm curious. Maybe a surmountable issue, if indeed it is the case -- but we shall see. Nonetheless, this functionality hasn't changed. Only the return type :smile:

dhensby:
I've tested personally and also: http://php.net/manual/en/function.filter-var.php#111828

patricknelson:
Makes sense. Maybe a separate issue? Or we can address that here too, since that would technically be a change from what we're currently doing at least as far as the `->validEmailAddress()` method is concerned.

dhensby:
We can deal with it in SS 4 :)

email/Email.php
```diff
@@ -184,103 +204,181 @@ public function setBounceHandlerURL($bounceHandlerURL) {
...
+
+	/**
+	 * @param string $email
+	 * @return $this
+	 * @deprecated 3.5 Use the "setReplyTo" method instead
 	 */
 	public function replyTo($email) {
+		Deprecation::notice('4.0', 'Use the "setReplyTo" method instead');
 		$this->addCustomHeader('Reply-To', $email);
 		return $this;
```
kinglozzer:
```php
Deprecation::notice('4.0', 'Use the "setReplyTo" method instead');
return $this->setReplyTo($email);
```

DRY and all that :D

email/Email.php
kinglozzer: Shouldn’t this just `return filter_var($address, FILTER_VALIDATE_EMAIL)`? Why the `!== false`?

email/Email.php
```diff
@@ -609,37 +716,17 @@ public static function bcc_all_emails_to($emailAddress) {
...
-		return preg_match("!^$addr_spec$!", $email) ? 1 : 0;
-	}
-
-	/**
-	 * Encode an email-address to protect it from spambots.
-	 * At the moment only simple string substitutions,
-	 * which are not 100% safe from email harvesting.
+		Deprecation::notice('4.0', 'Use the "validEmailAddress" method instead');
+		return static::validEmailAddress($email);
+	}
```
kinglozzer: How does the old method here differ to `filter_var()`? Perhaps we should just leave the old method alone, but throw a deprecation notice?

email/Email.php
```diff
@@ -184,103 +204,181 @@ public function setBounceHandlerURL($bounceHandlerURL) {
...
+	 */
+	public function setReplyTo($val) {
+		$this->addCustomHeader('Reply-To', $val);
+		return $this;
+	}
+
+	/**
+	 * @param string $email
+	 * @return $this
+	 * @deprecated 3.5 Use the "setReplyTo" method instead
```
dhensby: This comment mis-matches with the actual deprecation declaration

dhensby:
Our convention is to name static methods with underscores so `is_valid_address` shouldn't be the method being deprecated.

Also, please can you check through core for anywhere using the deprecated method and change it to use the new one, please?

kinglozzer:
Just comparing the regex in `Email::is_valid_address()` to `filter_var()`, it looks like the regex we have is more permissive than `filter_var()`, so changing it might be a BC-breaking change 😞. I set up a quick example: https://3v4l.org/LfWYY, it seems that `filter_var()` falls down on whitespace. The [PHP docs](http://php.net/manual/en/filter.filters.validate.php) confirm that:

> In general, this validates e-mail addresses against the syntax in RFC 822, with the exceptions that comments and whitespace folding are not supported.

patricknelson:
@dhensby I suspected, but didn't act on that suspicion (busy, distracted). Thanks for reinforcing that point :smile: 

@kinglozzer In terms of the validation of this static method, I agree that's a problem. For sure then, something needs to change. It could be anywhere from from benign to a breaking change:

- Just update the documentation so developers aren't incorrectly sold on an "RFC822-valid email format" check when that is, in fact, not the case.
- Set it up to simply use `filter_var()` instead (like I'm doing now, albeit in the wrong place).
- Hybrid approaches: Do both. 
  - Update docs on current method and either add a new method that is RFC-compliant (while potentially deprecating `::is_valid_address()` in the process for v4)
  - Update docs on the current method so it doesn't pretend to be RFC-compliant and only update it to actually be RFC-compliant in v4

email/Email.php
patricknelson: Nice catch, did that somewhere else as well. I've fixed this in my latest push.

email/Email.php
patricknelson: Not just DRY, but also "dogfooding" it's own deprecation notice, hah. Not sure how I missed that, so thanks for pointing it out. Fixed in my latest push.

patricknelson:
Alright, @dhensby it looks like there aren't any external usages of `Email`'s method `->validEmailAddress()`.  I've also just switched the methods around (for the time being) to correct for this nomenclature. I admit my changes may not really make it and we may need to revert to the regex version, so at least I succeeded in deprecating one of these redundant methods :+1: 

Also @kinglozzer I failed to ask you: That's an interesting demo. Considering `filter_var()` does still seem to fall a bit short and is more strict in some situations, do you think we should retain the regex? Typically I subscribe to the philosophy of everyone rallying behind shared/common components (like `filter_var()`) which can improve over time and not duplicating code. Maybe in newer versions of PHP, `filter_var()` will improve and SilverStripe will automatically benefit from these improvements. So **the question** is:  Utilize `filter_var()` now (if it's "good enough") for those future benefits, or, use our own (copied) regex implementation?

**Topic shift:** What's interesting there, however, is that I don't think it's actually _validating_ per se, it's more of an extraction/parsing than it is any sort of validation, see: https://github.com/silverstripe/silverstripe-framework/blob/3/email/Mailer.php#L468

So it makes me wonder if that method should be refactored and, if necessary, consideration should be provided to think about if the calling methods (e.g. `->sendPreparedMessage()` should then actually validate first prior to actually attempting to send, once the email address has been extracted (which may help with reducing possible header injection as well at least in those fields, if that protection isn't already well established somewhere else, which it may be).

email/Email.php
```diff
@@ -609,37 +714,20 @@ public static function bcc_all_emails_to($emailAddress) {
...
-		$addr_spec = "$local_part\\x40$domain";
-
-		return preg_match("!^$addr_spec$!", $email) ? 1 : 0;
-	}
-
-	/**
-	 * Encode an email-address to protect it from spambots.
-	 * At the moment only simple string substitutions,
-	 * which are not 100% safe from email harvesting.
+		return filter_var($email, FILTER_VALIDATE_EMAIL) !== false;
```
micmania1: Can this use Email::validEmailAddress() instead and deprecate this function

dhensby:
If the two functions have different behaviours, then we can't just unify them and deprecate, it's not a "non-breaking" API change if the API changes to stop valid emails getting through.

I think we should use the regex for now if it's more accurate...

micmania1:
Ah, I assumed this was 4.0

patricknelson:
Right, that's why I wanted clarification thanks @dhensby -- reverted that code. Updated it so it's more accurate; it may not be technically 100% RFC 822 compliant but as close as we can get for now. Would be great to have a thorough test suite for this (maybe akin to what you could find here: https://github.com/dominicsayers/isemail) but I guess we don't need to be perfect per se; plus that's a lot of work :joy: 

@micmania1 Your follow up statement is confusing me. Do you still need an answer to https://github.com/silverstripe/silverstripe-framework/pull/5790#discussion_r70727183 or are you set?

dhensby:
We don't need to worry about a thorough test suite because mailer will be completely refactored for 4.x :)

patricknelson:
Got it @dhensby - quick question on v4: Are there plans to ensure that new refactoring has an API, or maybe a module that will work as a drop-in replacement, for backward compatibility? Part of why I even looked at this so closely and wanted to clean it up a bit was due to the fact that our site as a whole was using 3 different methods to send email (due to developers, entropy of time, etc...), so I made sure we were all centralized on one solution (the core framework method). 

p.s. I know I need to repair that unit test when I get into the office :tired_face:

dhensby:
I don't think we'll be making a drop-in module to keep the existing APIs, that may be very difficult to do.

I'm looking at integrating `SwiftMailer` as the backend for sending emails, but making it drop-in so you could use anything to send emails (though SwiftMailer does provide a lot of plugins).

Basically we'll be moving to something like this:

```php

$email = new Email();
$email->setTo(...);
...

Mailer::inst()->send($email);
```

I started a while ago but need to find time to get it finished...

patricknelson:
Yeah, that makes more sense to me and, in fact, how I originally expected the API to be at first glance, given I saw a `Mail` and `Email` class, I figured that was how the responsibility was separated. Craft an email message, send it with `Mail` but, turns out, you can send it from `Email` itself too.  Maybe should be more separated :+1:  

Fixed the unit test  by proxy of ensuring the return value is actually `true` or `false` (as originally documented) even though it was explicitly returning `0` or `1`. Should be minor.

dhensby:
Thanks for the work on this

Issue #5789, opened 2016-07-08
Premature return in DevBuildController.php


aNickzz:
There is some dead code near the end of `DevBuildController->build` in `/framework/dev/DevBuildController.php`.
```
public function build($request) {
	if(Director::is_cli()) {
		$da = DatabaseAdmin::create();
		return $da->handleRequest($request, $this->model);
	} else {
		$renderer = DebugView::create();
		$renderer->writeHeader();
		$renderer->writeInfo("Environment Builder", Director::absoluteBaseURL());
		echo "<div class=\"build\">";

		$da = DatabaseAdmin::create();
		return $da->handleRequest($request, $this->model);

		echo "</div>";
		$renderer->writeFooter();
	}
}
```

Suggested fix:
```
public function build($request) {
	if(Director::is_cli()) {
		$da = DatabaseAdmin::create();
		return $da->handleRequest($request, $this->model);
	} else {
		$renderer = DebugView::create();
		$renderer->writeHeader();
		$renderer->writeInfo("Environment Builder", Director::absoluteBaseURL());
		echo "<div class=\"build\">";

		$da = DatabaseAdmin::create();
		$response = $da->handleRequest($request, $this->model);

		echo "</div>";
		$renderer->writeFooter();

		return $response
	}
}
```

dhensby:
Good spot - would you mind creating a PR for this?

PR #5791, opened 2016-07-10
Move return to end of function.

aNickzz:
There is some dead code near the end of `DevBuildController->build` in `/framework/dev/DevBuildController.php`.
[Issue #5789](https://github.com/silverstripe/silverstripe-framework/issues/5789)


dhensby:
Thanks

PR #5793, opened 2016-07-11
Remove unnecessary require calls.

sminnee:
Now that we have PSR-4 compliance on these classes, manual include
statements aren’t needed.


sminnee:
Note that EmbedTest is flappy – a re-execution cause it to succeed.

PR #5796, opened 2016-07-12
API Update react sections to use react-router instead of page.js

### LINKED ISSUE #5711: Refactor frontend routing to react-router
The current [page.js](https://github.com/visionmedia/page.js) based routing was a "least common denominator approach", where routing functionality was stripped back to simple callback handling. See https://github.com/silverstripe/silverstripe-framework/pull/5243 for context. The rationale was that a react based router couldn't handle Entwine/PJAX based legacy sections cleanly. This has caused a lot of work in the last week, combined with a lot of learning about how routing should influence views and redux state.

Some disadvantages:
 * All but the innermost components need to be aware of the routing hierarchy, with switch statements and clumsy route state management in React/Redux vs. page.js. Check the [react-router introduction](https://github.com/reactjs/react-router/blob/master/docs/Introduction.md) to get an understanding how this approach differs for standard callback-based routing, see "Without React Router".
 * `AssetGalleryField` was designed as a standalone, "semi-Entwine" component with its own React root initialisation based on Entwine matching, and its own redux state. This doesn't fit into the bigger picture of how we want the different UIs to work together
 * Since page.js doesn't know about React, it can't support users in nested views - leading to single components with too many responsibilities (the `switch(this.props.view)` conditional in `AssetAdmin` is a good example)
 * Same goes for breadcrumbs - since there's no sense of which part of a route is rendered by a component, there's no support for creating breadcrumbs automatically in page.js
 * Lazy route registration in page.js is possible (after calling `page.start()`), but there's no introspection into existing routes which can be limiting for customisation. With react-router, we can keep all routes in an SS-managed registry and pull them on demand.

We're attempting to get [react-router](https://github.com/reactjs/react-router) into core as an optional router which just kicks in for new react-based sections. We might use [react-router-redux](https://github.com/reactjs/react-router-redux) (not [redux-router](https://github.com/acdlite/redux-router) since its fallen [out of favour](https://github.com/acdlite/redux-router/issues/172)). It's technically not required at the moment (we can pass `props.params` through to children components), but seems likely to be required for any additional components from third party extensions, since those won't have easy access to `props.params` (note that redux-router [recommends](https://github.com/reactjs/react-router-redux#how-do-i-access-router-state-in-a-container-component) not to use this redux state directly though).

Branches:
https://github.com/open-sausages/silverstripe-framework/tree/pulls/4.0/react-router
https://github.com/open-sausages/silverstripe-asset-admin/tree/pulls/4.0/react-router

# Tasks

 * [x] Switch between legacy and react routing on boot
 * [x] Rewrite `lib/RouteRegister` to work with react-router config instead of callbacks
 * [x] Remove PJAX loading of main sections in menu (full page reload required for router switching)
 * [x] Allow injection of components outside `framework` into `FormBuilder`
 * [x] ~~Separate out `AssetGalleryField` logic from `AssetAdmin` container so it can be lazy loaded via `FormBuilder`~~ (we removed AssetGalleryField instead, not relying on form schema for this part)
 * [x] Share redux state and reducers between framework and asset-admin modules
 * [x] Allow lazy registration of routes in react-router configuration (via getChildRoutes callbacks)
 * [x] ~~Lazy registration of `admin/assets/show/:folderId/edit/:fileId` route via `AssetGalleryField`~~ (no longer required since we removed that indirection)
 * [x] Render `Editor` through react-router as a childRoute of `AssetGalleryField`
 * [x] ~~Re-match route once `AssetGalleryField` has lazy registered routes (might need to show `Editor` component)~~
 * [x] Use react-router `:folderId` and `:fileId` props for view initialisation
 * [x] Remove dependance of `AssetAdmin` init on jQuery.entwine, use `FormBuilder` instead
 * [x] Fix `router.show()` calls throughout CMS
 * [x] Use `browserHistory.push()` instead of `router.show()` in `CampaignAdmin` and `AssetAdmin` ([docs](https://github.com/reactjs/react-router/blob/master/docs/guides/NavigatingOutsideOfComponents.md))
 * [x] Update changed `propTypes` on all components

Things that block merge:

 * [x] Introduce `react-router-redux` (Ingo's working this now) to populate redux state from URL (albeit asynchronously so can only be read at certain times)
 * [x] Fix "flash of editor view" on admin/assets load
 * [x] Fix editor now showing in responsive view (probably just a HTML nesting bug)
 * [ ] ~~Sync back Redux state into route, and identify legal times to read react-router-redux state (see [current implementation through refreshUrl](https://github.com/silverstripe/silverstripe-asset-admin/blob/0a725b9db5c4333051328401a59785023ecf66b0/client/src/containers/AssetAdmin/AssetAdmin.js#L76)), and [rationale from Redux creator](https://github.com/reactjs/redux/issues/227#issuecomment-119237073)~~ Pulled out into https://github.com/silverstripe/silverstripe-framework/issues/5797
 * [ ] ~~If "Sync back Redux state into route" is a dead end: Ensure that all existing reducers are accessing router params and not legacy state values (`state.gallery.fileId` and `state.gallery.folderId`)~~ Pulled out into https://github.com/silverstripe/silverstripe-framework/issues/5797
 * [x] Fix `Cannot read property 'replace' of undefined` error on startup (history state doesn't have all required properties)
 * [x] Use persistent `backend.createEndpointFetcher()` rather than recreating on each `render()` call in `AssetAdmin`
 * [x] Write tests for legacy/react router switching
 * [x] Cross-browser testing
 * [x] Test for state management regressions
    * Open a folder, highlight files, navigate to a different folder, highlight other files, delete those => should not delete originally highlighted files
    * Open editor, close it again, ensure the URL updates. Navigate back and forward in browser.
    * Open edit for a file, then also select the checkbox on that file and delete it => the editor should close
 * [x] Fix `AssetAdmin` container layout regression (too much padding). May be "ensure HTML is good and give to Paul"
 * [x] Update routing docs (mention new `reactRouter` boolean flag in `LeftAndMain->getClientConfig()`)

Things that don't block merge:

 * Clean up legacy routing:
   * [ ] Refactor `lib/Router` to avoid wrapping around page.js (ideally we don't need a custom `page.show()`). 
   * [ ] Avoid full page reload in CMS menu when navigating between sections with the same router. Already the case in react router, maybe not in the legacy issues.
 * Tidy up asset breadcrumbs:
   * [x] Fix `back()` navigation on breadcrumbs (use new `browserHistory` object) ([docs](https://github.com/ReactJSTraining/history/blob/master/docs/GettingStarted.md))
   * [x] Infer breadcrumbs from route stack (ideally each component can contribute one or more breadcrumbs, see [example](https://github.com/reactjs/react-router/blob/master/examples/breadcrumbs/app.js)

# Related

 * Remove entwine/PJAX dependency from AssetAdmin boot logic (https://github.com/silverstripe/silverstripe-asset-admin/issues/131)
 * File edit doesn't update URL after navigating to different React section (https://github.com/silverstripe/silverstripe-asset-admin/issues/193)
 * IE10+ doesn't update URL on folder navigation or file display (https://github.com/silverstripe/silverstripe-asset-admin/issues/188)
 * Refactor <Gallery> and <Editor> into containers and presentational components (https://github.com/silverstripe/silverstripe-asset-admin/issues/179)
 * Use FormBuilder in file edit form (https://github.com/silverstripe/silverstripe-asset-admin/issues/171)
 * Implement breadcrumbs (https://github.com/silverstripe/silverstripe-asset-admin/issues/158)

# Recommended Reading

- David/Ingo React research from early 2016 (originally recommended page.js): https://docs.google.com/document/d/1MXDhNPz60Lt5X4VlyB6hFUJGl3H6vIPxPAqV2lmdMnI/edit#
- Intro to react-router-redux: http://jlongster.com/A-Simple-Way-to-Route-with-Redux
- Good full app example: https://github.com/emmenko/redux-react-router-async-example
- Best practice from Redux author on state/route sync: https://github.com/reactjs/redux/issues/227#issuecomment-119237073
- Wordpress data persistence evolution: https://github.com/Automattic/wp-calypso/blob/master/docs/our-approach-to-data.md#third-era-redux-global-state-tree-december-2015---present

tractorcow:
Fixes #5711 
Asset-admin PR: https://github.com/silverstripe/silverstripe-asset-admin/pull/204
Reports PR: https://github.com/silverstripe-labs/silverstripe-reports/pull/38


.eslintrc
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": "airbnb"
+	"extends": "airbnb",
```
dhensby: indentation

tractorcow:
This has been rebased and updated, and is ready for merge.

admin/client/src/lib/Injector.js
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Default data type to component mappings.
+   * Used as a fallback when no component type is provided in the form schema.
+   *
+   * @param string dataType - The data type provided by the form schema.
+   * @return object|null
+   */
+  getComponentByDataType(dataType) {
+    switch (dataType) {
```
sminnee: This is, well, pretty limited. ;-)

Doesn't block merge, but I would be inclined to use registrations of the form:

```js
injector.register('TextField', TextField);
injector.register('DataType.Text', TextField);
injector.register('HiddenField', HiddenField);
injector.register('DataType.Hidden', HiddenField);
injector.register('GridField', GridField);
injector.register('DataType.Custom', GridField);
```

And then getComponentByDataType can be:

```js
getComponentByDataType(dataType) {
  return this.getComponentByName('DataType.' + dataType);
}
```

sminnee:
I've had a read of this; although I don't fully understand it, it looks reasonable.

sminnee:
Failure relates to the  reports module which has a companion PR. https://github.com/silverstripe-labs/silverstripe-reports/pull/38

PR #5798, opened 2016-07-12
BUG Fix issue with HtmlEditorField.js raising javascript errors on pages without editor fields

tractorcow:
Fixes the issue currently on master: https://travis-ci.org/silverstripe/silverstripe-framework/jobs/144031297

Also Convert tabs -> spaces in HtmlEditorField.js


tractorcow:
View with ?w=1 in the querystring to hide the whitespace changes.

https://github.com/silverstripe/silverstripe-framework/pull/5798/files?w=1

dhensby:
Rather than wrapping every call in a conditional, can't we return some kind of dud or empty editor instead so we get the "usual" care-free chaining that most people are used to in JS?

tractorcow:
You mean a monad?

sminnee:
A monad would be slightly different - a wrapper we put the editor inside, that will put the result of any method call inside the same kind of wrapper. Do get the actual editor object you ned to dereference it with `.getValue()` or something.

The issue with that approach is that it would require refactoring of the current code.

Returning some kind of NullEditor is probably better.

sminnee:
Can we please fix the whitespace of src/legacy as a separate commit that does only that. I'm tired of history pollution from whitespace changes.

If we merge this as-is it will be close to impossible to identify this change in a blame.

Separate commit on the same PR is okay.

tractorcow:
I've completely re-approached this problem by addressing the fact that change detection is being invoked on in-bound forms rather than out-bound forms.

The issue was due to navigation invoking multiple handleStateChanged() events on patterns that matched multiple routes. E.g. each of the below would invoke this method for /admin/pages/edit/4

 - `/admin/pages/edit(/*?)?`
 - `/admin/pages(/*?)?`
 
 The first invocation would act normally, and trigger change detection, then subsequently load the page content into the page.
 
 The second one would attempt to do change detection, but it would do it on the immediately added form, which would be in the middle of bootstrapping (tinymce initialises async)
 
 There is yet the possibility that invoking htmleditor API during this time could cause the same error, but it's much more prudent in this case to prevent the incorrect call in the first place. :)

tractorcow:
Remaining issue will be fixed with the default_cast PR over at https://github.com/silverstripe/silverstripe-framework/pull/5653

The bug is that TextareaField.ss has this

```
<textarea $AttributesHTML>$ValueEntities.RAW</textarea>
```

and RAW Has casting = HTMLText, and thus is processing shortcodes, even though our editor requires shortcodes to be left in for editing.

PR #5802, opened 2016-07-13
No need to call twice onBeforeHTTPError

lekoala:


lekoala:
sorry did not notice the do before the extension, i close this

tractorcow:
Thanks anyway.

:)

To be honest, the fact that you made this mistake means the code isn't clear. I would go with `"onBeforeHTTPError{$errorCode}"` to be explicit, personally.

PR #5807, opened 2016-07-14
Re-enable travis artifacts on framework

tractorcow:
This hasn't been updated for a while, and the old keys and defines needed replacing.


tractorcow:
Note that artifacts won't upload on a pull request; It'll only work after merge, so this pull request won't really demonstrate that it fixes the issue. :)

Issue #5772, opened 2016-07-04
Rogue notifications on v3.4 with OSX Safari client on Centos 7 server 


giankdm:
I've got a similar problem to #5384 on v3.4. 

In the dev env (Ubuntu Trusty 14) the issue doesn't manifest, however in the stage env (CentOS 7) on the destination server I've got a slew of notifications. It happens in OSX Safari, doesn't happen in OSX Google Chrome. I think a new browser update added to the problem.
![1f854960-3c88-11e6-8d4a-59af103408c4](https://cloud.githubusercontent.com/assets/3578378/16564608/b60651e2-4208-11e6-8ce2-ac974bd140ce.jpg)

tractorcow:
Can you tell me what X-Status header your responses are sending?

giankdm:
Site is served over SSL. Status of the HTTPS XHR requests is `HTTP/2.0 200`.
I tried disabling SSL: status of the XHR requests becomes `OK`.

So I updated the  `ignoredMessages` array in `framework/admin/javascript/LeftAndMain.js`, added `HTTP/2.0 200` to the list and the popup did not appear.

`ignoredMessages = ['OK', 'success', 'HTTP/2.0 200'];`

dhensby:
@giankdm what software are you using to get HTTP/2 on CentOS 7?

giankdm:
It is a Plesk 12.5 installation over Centos 7, with Apache and nginx managed by Odin, latest updates applied. So it is pretty common. When you install a certificate, apparently it triggers the new HTTP 2.0/SPDY protocol on certain browsers, and it appears to have some slightly different statuses.

dhensby:
HTTP2 doesn't send http status strings, just codes. which is the "problem" - I think this is a bug with jquery / js / safari not with the framework - but we do need to decide if we want to try to mitigate it

giankdm:
Apparently Safari translates the `200` code into a string with `HTTP/2.0 ` attached. However, the same environment in Google Chrome reports a different revenue: the status is reported to be `200`. I think it should be mitigated (Safari has a longer update cycle than Chrome, and if no tests are failed, it is pretty simple to circumvent). Meanwhile we could try to create a minimum code sample to reproduce the bug and submit an Apple Radar and/or a jQuery issue.

giankdm:
Reported on Safari Radar with issue #27171810. Please also add to the discussion there.

dhensby:
[...]

PR #5809, opened 2016-07-14
Quick fix for HTTP/2.0 on OSX/Safari v9.1.1

giankdm:
Workaround to issue #5772 that adds an exception in the `XHR` ignored messages list to account for Safari rewriting status codes in `HTTP/2.0` to string messages.


dhensby:
Perfect, thanks!

PR #5814, opened 2016-07-15
FIX: Fixes support for "inline" form actions (fixes #2534)

### LINKED ISSUE #2534: FormActions/InlineFormAction outside of the main 'actions' FieldList
In discussing an issue with `InlineFormAction` and `CMSForm->validate()`, @Martimiz brought up the point that `InlineFormAction` only really exists so that we can more easily control the physical location of actions in forms.

To better demonstrate what I mean, here's an example issue:

```php
$fields = new FieldList(
	new TextField('Name', 'Name'),
	new InlineFormAction('doNotValidate', 'Do not validate')
);
$actions = new FieldList(new FormAction('doSomeForm', 'Do stuff'));
$validator = new RequiredFields('Name');
$form = new CMSForm($this, 'SomeForm', $fields, $actions, $validator);
$form->setValidationExemptActions(array('doNotValidate'));
```

This doesn't work as expected - `doNotValidate`, as it wasn't passed in to `Form->__construct()` as an action, doesn't skip validation.

Should we be patching this kind of issue and allowing actions to behave consistently anywhere? Or should we be trying to force people to pass all actions to `Form->__construct()`/`Form->setActions()` in the 'actions' `FieldList`?

`Form->httpSubmission()` allows the form to pick up any actions as the 'current' action (by searching the POST/GET data for an `action_xxx` param), regardless of their position in the form, whereas `Form->buttonClicked()` only loops through the forms 'actions' list. Seems like there's a conflict there somewhere.

kinglozzer:


forms/Form.php
```diff
@@ -475,16 +476,28 @@ public function httpSubmission($request) {
...
+		// Always allow actions which map to buttons. See httpSubmission() for further access checks.
+		$fields = $this->fields->dataFields() ?: array();
+		$actions = $this->actions->dataFields() ?: array();
+
+		$fieldsAndActions = array_merge($fields, $actions);
+ 		foreach ($fieldsAndActions as $fieldOrAction) {
+			if ($fieldOrAction instanceof FormAction && $fieldOrAction->actionName() === $action) {
+				return true;
+			}
+		}
```
kinglozzer: This isn’t DRY (it’s copied from `Form::buttonClicked()`), but semver and new apis ¯\\_(ツ)_/¯

dhensby:
Do you want to be in charge of merging it up to 3 and/or making a new PR?

forms/Form.php
```diff
@@ -475,16 +476,28 @@ public function httpSubmission($request) {
...
+			return true;
+		}
+
+		// Always allow actions which map to buttons. See httpSubmission() for further access checks.
+		$fields = $this->fields->dataFields() ?: array();
+		$actions = $this->actions->dataFields() ?: array();
+
+		$fieldsAndActions = array_merge($fields, $actions);
+ 		foreach ($fieldsAndActions as $fieldOrAction) {
+			if ($fieldOrAction instanceof FormAction && $fieldOrAction->actionName() === $action) {
```
tractorcow: I like this a lot better than looking for 'action_$action' field. :D Nice job.

PR #5816, opened 2016-07-15
Fixed PDO returning int as string

trendler:
Without the fix PDO returns integers as strings, which is bad if you write a JSON API. Seems like ATTR_EMULATE_PREPARES has to be explicitly set to false ([see this SO post](http://stackoverflow.com/questions/1197005/how-to-get-numeric-types-from-mysql-using-pdo#answer-1197424)).


dhensby:
Dupe of https://github.com/silverstripe/silverstripe-framework/pull/4919 - but this leads to regressions I can't get to the bottom of it.

tractorcow:
I'm going to close as duplicate, sorry.

PR #5818, opened 2016-07-17
Use columns from gridfield for export

antons-:
This resolves an issue where joined columns inside gridfields aren't used when exporting to a CSV file.

The new `GridFieldExportButton->getExportColumnsForGridField()` that I am proposing basically copies the functionality from `GridFieldPrintButton->getPrintColumnsForGridField()` https://github.com/silverstripe/silverstripe-framework/blob/master/forms/gridfield/GridFieldPrintButton.php#L135

I did not use `$this->exportColumns`.

The old code did not have a case for `$dataCols->getDisplayFields($gridField);`, which was an issue if you were using a data set that had columns obtained via a left join in the gridfield's query. In my case, this was for a report I was building with 2 left joins.

The old code would default to singleton(`$gridField->getModelClass())->summaryFields();`, and in my case, the the columns that were joined aren't in the summary fields so they weren't displayed.

### DIFF
--- a/forms/gridfield/GridFieldExportButton.php
+++ b/forms/gridfield/GridFieldExportButton.php
@@ -96,6 +96,25 @@ class GridFieldExportButton implements GridField_HTMLProvider, GridField_ActionP
 			return SS_HTTPRequest::send_file($fileData, $fileName, 'text/csv');
 		}
 	}
+	
+	/**
+	 * Return the columns to export
+	 * 
+	 * @param GridField $gridField 
+	 * 
+	 * @return array
+	 */
+	protected function getExportColumnsForGridField(GridField $gridField) {
+		if($this->exportColumns) {
+			$exportColumns = $this->exportColumns;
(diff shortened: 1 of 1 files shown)


dhensby:
Looks like a candidate to go against 3 branch. Would you mind rebasing and opening a new PR against 3?

antons-:
@dhensby Can do but I'm a bit of a noob when it comes to rebasing. I'm not having much luck rebasing the commit to the 3 branch. Is a cherry-pick ok or do you have any advice here? Otherwise I'm thinking the easiest way is to create a new branch off 3 and create a PR off that?

dhensby:
@antons- yes, sorry - for you to get this in 3 you'll need to:

1. Close this PR
2. Create a new branch from 3
3. Cherry-pick this commit (5131bf0)
4. Create a new PR against 3 - please reference this one when you create the PR

Thanks!

PR #5821, opened 2016-07-18
BUG Fix regression with uploadfield casting

tractorcow:
Before:

image

After:

image


tractorcow:
Another fix at https://github.com/silverstripe/silverstripe-framework/pull/5861/ was necessary. :)

PR #5822, opened 2016-07-18
Use columns from gridfield for export

antons-:
Was PR #5818. On request, the base fork was changed from master to the 3 branch

This resolves an issue where joined columns inside gridfields aren't used when exporting to a CSV file.

The new `GridFieldExportButton->getExportColumnsForGridField()` that I am proposing basically copies the functionality from `GridFieldPrintButton->getPrintColumnsForGridField()` https://github.com/silverstripe/silverstripe-framework/blob/master/forms/gridfield/GridFieldPrintButton.php#L135

I did not use `$this->exportColumns`.

The old code did not have a case for `$dataCols->getDisplayFields($gridField);`, which was an issue if you were using a data set that had columns obtained via a left join in the gridfield's query. In my case, this was for a report I was building with 2 left joins.

The old code would default to singleton(`$gridField->getModelClass())->summaryFields();`, and in my case, the the columns that were joined aren't in the summary fields so they weren't displayed.

### DIFF
--- a/forms/gridfield/GridFieldExportButton.php
+++ b/forms/gridfield/GridFieldExportButton.php
@@ -91,6 +91,25 @@ class GridFieldExportButton implements GridField_HTMLProvider, GridField_ActionP
 			return SS_HTTPRequest::send_file($fileData, $fileName, 'text/csv');
 		}
 	}
+	
+	/**
+	 * Return the columns to export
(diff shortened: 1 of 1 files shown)


dhensby:
Thanks @antons-

PR #5823, opened 2016-07-18
FIX: Hierarchy was incorrectly unexpanding nodes that had been previously expanded

madmatt:
Original pull-request with comments and notes: https://github.com/silverstripe/silverstripe-framework/pull/5349

This is opened against the 3.2 branch so it gets merged back to earlier versions of framework as well as the latest.


madmatt:
cc @dhensby, @tractorcow 

Will keep an eye on the tests :)

PR #5820, opened 2016-07-18
API Exclude templates in 'themes' directories from $default theme

tractorcow:
Additionally, $themes parameter is now mandatory in TemplateLoader::findTemplate()

Required for https://github.com/silverstripe/silverstripe-framework/issues/5819

PR #5824, opened 2016-07-19
TESTS: code-climate should ignore admin/thirdparty

sminnee:
admin/thirdparty shouldn’t be checked by code climate as it isn’t code
that we maintain.

This causes false failures on tickets such as #5820


sminnee:
FYI: Failing test is pre-existing.

PR #5830, opened 2016-07-21
API Abstract ThemeManifest into ThemeList

tractorcow:
BUG Fix Requirements not resolving $default theme

Includes some additional coverage for themedCSS and themedJavascript, since none exist.


tractorcow:
Broken CMS behat tests are pre-existing and being addressed outside of this PR.

PR #3274, opened 2014-07-08
API Custom admin URL

colymba:
Pull request that follows the short discussion on the mailing list with @chillu here https://groups.google.com/forum/#!topic/silverstripe-dev/gGVRd0wS0U8

This PR goes along with the CMS one here https://github.com/silverstripe/silverstripe-cms/pull/1041
There is probably a few modules that will need updating, not sure which ones yet?

So far still a bit of a work in progress until we are happy with the implementation.

In short, this PR does:
- Use `AdminRootController` as the main class for admin URL configuration etc.
- `AdminRootController.url_base` is the main/only config for custom URL
- This is used during bootstrap to create a `Director.rules` config entry via `AdminRootController::set_admin_route()`
- Remove all other instances of `url_base` configs in other classes
- Makes utility functions and variables available globally:
  - PHP `AdminRootController::adminURL()`
  - SS `$AdminURL`
  - JS `ss.config.adminURL`

To me this seemed like the best approach and made things easier by only having to configure 1 thing once, even though the bootstrap call might not be that elegant? What do you think?

Note: one thing left are `SiteTree.DEFAULTHOMECONTENT` translations that have the admin URL hard-coded, but this might not be an issue?

PR #5834, opened 2016-07-21
BUG Fix regressions in custom admin url from #3274

tractorcow:


admin/code/AdminRootController.php
```diff
@@ -9,28 +9,29 @@
...
+
+	/**
 	 * Convenience function to return the admin route config.
 	 * Looks for the {@link Director::$rules} for the current admin Controller.
+	 *
+	 * @return string
 	 */
 	public static function get_admin_route() {
-		if (Controller::has_curr()) {
-			$routeParams = Controller::curr()->getRequest()->routeParams();
```
tractorcow: We shouldn't be using the current controller because the current controller could be a non-admin one.

admin/templates/Includes/LeftAndMain_MenuStatus.ss
```diff
@@ -3,7 +3,7 @@
 	<% with $CurrentMember %>
 		<span>
 			<%t LeftAndMain_Menu_ss.Hello 'Hi' %>
-			<a href="{$AbsoluteBaseURL}{$AdminURL}admin/myprofile" class="profile-link">
+			<a href="{$AdminURL}myprofile" class="profile-link">
```
dhensby: this probably means the docs need amending too

tractorcow: Which docs? All I've done is remove an unnecessary leading url.

dhensby: Hmm, my bad, I thought there were docs examples showing to do it this way, but on inspection I don't see them

PR #5838, opened 2016-07-25
MINOR Clarify string concatenation to improve readability

tractorcow:
Replaces #5802

cc @lekoala


kinglozzer:
Gleefully merging this, as this has caught me out far too many times too 😅

PR #5845, opened 2016-07-28
Fix "Insert Link" dialog in HTMLEditorField. CSS still needs work.

hafriedlander:


tractorcow:
It doesn't fix all issues, but it does fix a major one with the dialog not working at all.

PR #5847, opened 2016-07-28
FIX ETag header now properly quoted

### LINKED ISSUE #4577: ETag header syntax in HTTP.php not valid
When running a Silverstripe site through [Redbot](https://redbot.org) header checking, ETag syntax is found to be non-standard.  The string returned should be [wrapped in double-quotes](http://stackoverflow.com/questions/6719214/syntax-for-etag).

dhensby:
Fixes #4577

### DIFF
--- a/control/HTTP.php
+++ b/control/HTTP.php
@@ -302,6 +302,9 @@ class HTTP {
 	}
 
 	public static function register_etag($etag) {
+		if (0 !== strpos('"')) {
+			$etag = sprintf('"%s"', $etag);
+		}
 		self::$etag = $etag;
 	}
 
@@ -445,6 +448,11 @@ class HTTP {
 
 		if(self::$etag) {
 			$responseHeaders['ETag'] = self::$etag;
+		}
+
+		// etag needs to be a quoted string according to HTTP spec
+		if (!empty($responseHeaders['ETag']) && 0 !== strpos($responseHeaders['ETag'], '"')) {
+			$responseHeaders['ETag'] = sprintf('"%s"', $responseHeaders['ETag']);
 		}
 
 		// Now that we've generated them, either output them or attach them to the SS_HTTPResponse as appropriate
--- a/control/HTTPResponse.php
+++ b/control/HTTPResponse.php
@@ -254,6 +254,10 @@ EOT
 			if(!headers_sent($file, $line)) {
 				header($_SERVER['SERVER_PROTOCOL'] . " $this->statusCode " . $this->getStatusDescription());
 				foreach($this->headers as $header => $value) {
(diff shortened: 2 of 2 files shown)


sminnee:
This feels a bit loose. It's not clear what the recommended value for a user passing an etag should be.

Should `HTTP:: register_etag()` be given quotes or not? I don't think we should say "you choose!" one of them needs to be the correct approach and the other should throw a warning before auto-correcting.

Ditto `HTTPResponse::addHeader("Etag", x)`. It seems strange that this one weird old header should be fiddled with when outputting a response. Would it be better to say that someone manually manipulating the header is doing it wrong?

My recommendation:

 * `HTTP:: register_etag()` expects an unquoted string. It throws a warning if a quoted string is passed.
 * `HTTPResponse::addHeader()` doesn't perform header correction. It expects callers to this since it's a just a container for headers, not an e-tag parser.
 * We make this change in `3` since we're introducing a warning for diligent developers will have likely passed it a quoted value to get the correct behaviour. That seems okay in a minor release but annoying in a patch release.

dhensby:
When writing this I looked at https://github.com/symfony/http-foundation and they have special handling for etag (https://github.com/symfony/http-foundation/blob/master/Response.php#L864-L877)

I think it's not widely known that an ETag needs to be a quoted string and here we are being a helpful framework by quoting the value if it hasn't been.

> Would it be better to say that someone manually manipulating the header is doing it wrong?

Maybe? I suppose it depends if you want the etag to depend on all the in-built rules that SS has or to make your own. If you want your own, that's not necessarily wrong... but I suppose it is if it's not being quoted properly.

With your suggestion where do we quote the etag value? Keep it at output time as I've got it?

tractorcow:
I would definitely prefer a "set on assignment" approach rather than "attempt to fixup already assigned headers" strategy. As @sminnee suggested, I'd have addressed this in `HTTP::register_etag`.

dhensby:
Is `HTTP::register_etag` the canonical place for setting this header? I don't like that if it is, we should be encouraging setting the header against the response object.

I do think we should be checking this on output as I'd argue that's the point of a framework. To catch common mistakes/bugs/issues and take the tedium of repetitive work away from devs.

We should accept that lots of people don't realise an ETag needs to be a quoted string and it's also probable that this is set directly by some devs via `$response->addHeader`.

I'll move it to register_etag, but I think we should keep the fix on outputs too...

dhensby:
`HTTP::add_cache_headers` doesn't use the `register_etag` function, so we do have to add this logic ourselves there

tractorcow:
I think it adds increased crud to the output method, but we can at least deprecate that in the future if we need to.

Issue #5710, opened 2016-06-17
Batch actions form layout


clarkepaul:
This issue is based on the bootstrap form conversion PR being accepted https://github.com/silverstripe/silverstripe-framework/pull/5697 

![pasted_image_17_06_16__3_00_pm](https://cloud.githubusercontent.com/assets/555033/16139524/702c9272-349c-11e6-877f-fa22f1adf101.png)

Fix at https://github.com/silverstripe/silverstripe-framework/pull/5849

tractorcow:
New PR at https://github.com/silverstripe/silverstripe-framework/pull/5849 needs reviewing to fix this.

clarkepaul:
Merge mine instead :) https://github.com/silverstripe/silverstripe-framework/pull/5850

sminnee:
Fix merged

PR #5850, opened 2016-07-28
Adjust width of select dropdown

clarkepaul:
Simple fix to amend the width of the dropdown, most of the fix has been included already in the forms PR


clarkepaul:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/5710

PR #5851, opened 2016-07-29
BUG Fix hash link navigation in CMS

tractorcow:
This bug can be replicated by clicking the list / tree view toggle in the top right of the pages edit area.

What should happen is that the hash in the url changes, but there is no ajax request. What actually happens is that page.js attemps to trigger a new route.

This fix back ports some of the old behaviour from History.js, which was removed. Now a `handleStateChange` call is only invoked if the path (sans hash) is actually changing (or force reloading).


admin/client/src/boot/BootRoutes.js
```diff
@@ -109,6 +109,7 @@ class BootRoutes {
     // Register all top level routes.
     // This can be removed when top level sections are converted to React,
     // have their own JavaScript controllers, and register their own routes.
+    var lastPath = null;
```
flamerohr: please use `let` instead of `var`

tractorcow:
Fixed.

PR #5853, opened 2016-07-31
Disable datepicker for disabled or readonly fields

flamerohr:
The fields were still editable through the datepicker, which could have undesirable effects.


admin/client/dist/js/LeftAndMain.js
```diff
@@ -289,7 +289,7 @@
         data.__forceReferer = forceReferer;
 
         if (forceReload) {
-          data.__forceReload = Math.random();
+          data.__forceReload = 1 + Math.random();
```
dhensby: This is unrelated and part of another PR (I believe)?

dhensby:
Seems legit - need the sanity checks to pass first

admin/client/dist/js/LeftAndMain.js
flamerohr: yeah, that was related to the hash fix @tractorcow did on Friday

PR #5854, opened 2016-08-01
API Add $action parameter to Controller::Link

tractorcow:
Reverts https://github.com/silverstripe/silverstripe-framework/commit/1ccdcbac5dbabf0cac43004e7c9e8048be95a93e

This is necessary to fix some actions in gridfield, where the form Link() action isn't being correctly concatenated to the parent controller.

Prior to this fix, clicking the gridfield header would cause a HTTP error (HTTP request was to `/admin/reports/EditForm/field/Reports`, instead of `/admin/reports/index/EditForm/field/Reports`)

The problem is that some controllers need to know if they are being appended to nested controllers, and occasionally need to add the full urlsegment in those cases. The 'home' page and the reports 'index' actions are two examples.

### DIFF
--- a/control/Controller.php
+++ b/control/Controller.php
@@ -127,10 +127,11 @@ class Controller extends RequestHandler implements TemplateGlobalProvider {
 	/**
 	 * Returns a link to this controller. Overload with your own Link rules if they exist.
 	 *
+	 * @param string $action Optional action
 	 * @return string
 	 */
-	public function Link() {
-		return get_class($this) .'/';
+	public function Link($action = null) {
+		return Controller::join_links(get_class($this), $action, '/');
 	}
 
 	/**
--- a/docs/en/04_Changelogs/4.0.0.md
+++ b/docs/en/04_Changelogs/4.0.0.md
@@ -56,6 +56,7 @@
  * `CMSMain::buildbrokenlinks()` action is removed.
  * `Folder_UnusedAssetsField` is removed.
  * `UpgradeSiteTreePermissionSchemaTask` is removed.
+ * `$action` parameter to `Controller::Link()` method is standardised.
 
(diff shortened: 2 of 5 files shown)


tractorcow:
There seems to be a regression in GridField... looking into it now.

control/Controller.php
```diff
@@ -127,10 +127,11 @@ public function doInit() {
 	/**
 	 * Returns a link to this controller. Overload with your own Link rules if they exist.
 	 *
+	 * @param string $action Optional action
 	 * @return string
 	 */
-	public function Link() {
-		return get_class($this) .'/';
+	public function Link($action = null) {
+		return Controller::join_links(get_class($this), $action, '/');
```
hafriedlander: 1) Why not just call with get_class + `func_get_args` - $action isn't guaranteed to be an action, it'll just be whatever the caller wants
2) Does it need to finish with a '/'? And if so, can't join_links do that, instead of added it here?

tractorcow: 1) We don't want all args, just the one. See tests below; I'm relying on $action being ignored if empty.
2) Modifying join_links could have a larger range of unexpected consequences, so I've opted for the least disruptive change.

tractorcow:
I've removed the change in behaviour to Form::FormAction() to reduce regressions.

This change will now standardise the behaviour of Controller::link(), but will not increase reliance on this standard behaviour. There were too many regressions in GridField due to the original rewrite.

dhensby:
We need a CMS fix too

tractorcow:
CMS fix at https://github.com/silverstripe/silverstripe-cms/pull/1570 @dhensby , thanks!

tractorcow:
Restarting tests.

PR #5862, opened 2016-08-02
Fix batch action permissions not applied to new nodes loaded

flamerohr:
When new nodes were loaded, checkboxes weren't disabled as was needed.


dhensby:
Is this only a problem in v4?

flamerohr:
I think it goes down to 3, I'll investigate a bit later on.
It may be a separate pull request to the 3 branch?

(note: just added an unbind for `onunmatch` for this pull request)

tractorcow:
This needs a rebase and maybe tests will pass. :)

flamerohr:
I've rebased, hopefully that's all good

dhensby:
thanks!

PR #5863, opened 2016-08-02
Modify getAuthenticator to fall back on default

andrewandante:
Currently it is possible to successfully return `$authenticator`, but if it is not in the array `$authenticators`, you are returned `NULL` instead of falling back to the `default_authenticator`. This PR changes the flow so that `default_authenticator` is returned in that circumstance.

### DIFF
--- a/Security/Security.php
+++ b/Security/Security.php
@@ -371,10 +371,8 @@ class Security extends Controller implements TemplateGlobalProvider {
 			if(in_array($authenticator, $authenticators)) {
 				return $authenticator;
 			}
-		} else {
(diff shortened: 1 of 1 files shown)


dhensby:
seems sensible. This PR is against master, so will only make it in the 4.0 release. Is this something that should be in 3?

andrewandante:
Most likely. Will make an equivalent PR against 3.4 - or would you recommend PRing against 3?

dhensby:
This looks like a good bug fix that should probably be in 3.2 and up, so if you PR against 3.2, then I'd be happy

PR #5864, opened 2016-08-02
BUG Fix unset config options returning isset() = true

### LINKED ISSUE #4791: Class config variables accessed via $this->config() seem empty unless used
After a fresh install of SS 3.2.1 and creating a simple custom class like

    class MyPage extends Page {

      /**
       * @config
       */
      private static $project = 'some string';

      public function init() {

        $a = $this->config()->project;
        $b = $this->config();
        $c = Config::inst()->get('MyPage', 'project');
        $d = Config::inst()->forClass('MyPage');
        var_dump($this->config()->project);
        var_dump(empty($this->config()->project));
        var_dump($a);
        var_dump(empty($a));
        var_dump($b->project);
        var_dump(empty($b->project));
        var_dump($c);
        var_dump(empty($c));
        var_dump($d->project);
        var_dump(empty($d->project));
      
      }
    }

and after calling 

    $mp = new MyPage();
    $mp->init();

the expected output would be:

string(11) "some string"
bool(false)
string(11) "some string"
bool(false)
string(11) "some string"
bool(false)
string(11) "some string"
bool(false)
string(11) "some string"
bool(false)

Instead, the output looks like:

string(11) "some string"
bool(true)
string(11) "some string"
bool(false)
string(11) "some string"
bool(true)
string(11) "some string"
bool(false)
string(11) "some string"
bool(true)

That means calling empty() on a variable accessed directly via $this->config() is not working as expected.
Same situation occurs when getting the config via Config::inst()->forClass('MyPage').

When the value is accessed via Config::inst()->get('MyClass', 'project'), it works as expected.

tractorcow:
Fixes #4791

### DIFF
--- a/core/Config.php
+++ b/core/Config.php
@@ -881,6 +881,7 @@ class Config_ForClass {
 
 	/**
 	 * @param string $name
+	 * @return mixed
 	 */
 	public function __get($name) {
 		return Config::inst()->get($this->class, $name);
@@ -892,6 +893,16 @@ class Config_ForClass {
 	 */
 	public function __set($name, $val) {
 		return Config::inst()->update($this->class, $name, $val);
+	}
+
+	/**
+	 * @param string $name
+	 * @return bool
+	 */
+	public function __isset($name)
+	{
+		$val = $this->__get($name);
+		return isset($val);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


tractorcow:
Wrong branch...

PR #5867, opened 2016-08-02
Modify getAuthenticator to fall back to get_default_authenticator

andrewandante:
As per #5863 PR'ing against 3.2. This will ensure that `getAuthenticator` falls back to `get_default_authenticator` if it fails to find any others.


dhensby:
rganks!

PR #5868, opened 2016-08-04
API Add 'source' and 'disabled' schema data for select fields

tractorcow:
This is necessary to form-schema enable select fields in react.


tractorcow:
Will be merged as a part of another larger pull request.

PR #5869, opened 2016-08-04
FIX don't look in node_modules

micmania1:
Its a scary place. And can sometimes take 5 minutes+ to get back out (that's 5 minutes each time you flush=1).


tractorcow:
'tis a silly place!

assertchris:
@tractorcow the best references.

PR #5872, opened 2016-08-07
FIX Use create syntax for CMSMemberLoginForm remember me form

dhensby:
see #5681 - port to CMSMemberLoginForm

PS: Why is this not in the admin module or cms?


dhensby:
thanks

PR #5873, opened 2016-08-08
BUG Make template lookup use 'type' correctly

tractorcow:
Dependency of https://github.com/silverstripe/silverstripe-cms/issues/1572

Rather than having classes (such as FormField) attempting to do its own path concatenation, pass in the 'type' property correctly to the template finder so that namespaced classes have the correct template paths.

Pulled out of an upcoming CMS namespace fix to resolve issue with namespaced templates.

### DIFF
--- a/tests/core/manifest/ThemeResourceLoaderTest.php
+++ b/tests/core/manifest/ThemeResourceLoaderTest.php
@@ -75,6 +75,54 @@ class ThemeResourceLoaderTest extends SapphireTest {
 		);
 	}
 
+	public function testFindTemplateByType() {
+		// Test that "type" is respected properly
+		$this->assertEquals(
+			"{$this->base}/module/templates/MyNamespace/Layout/MyClass.ss",
+			$this->loader->findTemplate(
+				[
+					[
+						'type' => 'Layout',
+						'MyNamespace/NonExistantTemplate'
+					],
+					[
+						'type' => 'Layout',
+						'MyNamespace/MyClass'
+					],
+					'MyNamespace/MyClass'
+				],
+				[
+					'silverstripe/module:subtheme',
+					'theme',
+					'$default',
+				]
+			)
+		);
+
+		// Non-typed template can be found even if looking for typed theme at a lower priority
+		$this->assertEquals(
+			"{$this->base}/module/templates/MyNamespace/MyClass.ss",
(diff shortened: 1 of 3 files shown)


tests/core/manifest/ThemeResourceLoaderTest.php
```diff
@@ -75,6 +75,54 @@ public function testFindNestedThemeTemplates() {
 		);
 	}
 
+	public function findTemplateByType() {
```
tractorcow: No wonder this test wasn't running... the method wasn't named `test*`... will fix.

tractorcow:
Ok, tests should be finally running against the new code. :D

forms/FormField.php
```diff
@@ -1052,15 +1052,22 @@ private function _templates($customTemplate = null, $customTemplateSuffix = null
 		$matches = array();
 
 		foreach(array_reverse(ClassInfo::ancestry($this)) as $className) {
-			$matches[] = 'forms/'. $className . $customTemplateSuffix;
+			$matches[] = [
+				'type' => 'forms',
```
hafriedlander: Although there's nothing _wrong_ with it, this isn't how I imagined "forms/Whatever.ss" - forms isn't a type, it's a "hollow" namespace (a template namespace that doesn't match any class namespace). That means if we ever did "relative Includes" (like forms/Button.ss checking forms/Includes/Foo.ss first, then Includes/Foo.ss) we could.

tractorcow:
I've reverted the changes to the FormField.php template lookup; This is no longer a dependency of the cms namespace work, but I'd still like to get this merged anyway.

forms/FormField.php
tractorcow: 👍  I understand now, will revert.

view/SSViewer.php
```diff
@@ -880,8 +880,8 @@ public static function get_templates_by_class($className, $suffix = '', $baseCla
 			$template = $class . $suffix;
 			if(SSViewer::hasTemplate($template)) {
 				$templates[] = $template;
-			} elseif(SSViewer::hasTemplate('Includes/'.$template)) {
```
tractorcow: This is still a bug. :) `Includes` is a type not a namespace prefix.

PR #5875, opened 2016-08-08
API Upgrade for CMS namespace

tractorcow:
See https://github.com/silverstripe/silverstripe-cms/issues/1572


hafriedlander:
For later, this is a pretty good list of "places CMS and Framework are undesirably bound together"

view/ViewableData.php
```diff
@@ -541,6 +542,9 @@ public function CSSClasses($stopAtClass = 'ViewableData') {
 			$classes[] = $this->template;
 		}
 
+		// Strip out namespaces
+		$classes = preg_replace('#.*\\\\#', '', $classes);
```
hafriedlander: I think the CSS classes for "MySite\Page" should probably be "Page" and "Mysite-Page"? Otherwise we're not solving the name clash problem for CSS.

(Also, preg_replace is ick. Although it's more work, pushing this into ClassInfo::ancestry is nicer. Might actually be faster too - http://stackoverflow.com/questions/19901850/how-do-i-get-an-objects-unqualified-short-class-name#answer-25472778)

tractorcow: Unfortunately this won't work for namespaced templates, since those won't exist in the class manifest, but will have a namespace (i.e. path).

hafriedlander: This function has nothing to do with namespaced templates. It takes a PHP class as an argument, and returns a list of CSS classes to use to hook CSS selectors on to. So I don't understand your comment, sorry.

hafriedlander: If you mean the code in lines 541 - 543, sure, they might need a regex. But this code still causes all PHP classes and template names to be cut down to their unqualified name only.

tractorcow: CSS classes with slashes are invalid in any case. I can move the code, however, if it'll make it look better for the pull request.

PR #5877, opened 2016-08-09
API remove DataFormatter class and all subclasses

tractorcow:
Cleanup of #5461

These classes aren't necessary to keep around, and were originally added to support the now-defunct RestfulService.


dhensby:
yay - they were always a nightmare anyway!

chillu:
Wait, what? These are required for the still operational `restfulserver` (a supported module), not the `RestfulService` which has been removed from core. See https://github.com/silverstripe/silverstripe-restfulserver/blob/f6025eac1d934a80e650c0aa9b700d35342cb44c/code/RestfulServer.php#L293. Can you please add those to restfulserver to avoid it breaking horribly in 4.x?

chillu:
Also, changelog mention please!

tractorcow:
Logged at https://github.com/silverstripe/silverstripe-framework/issues/6318

PR #5880, opened 2016-08-09
Improve FormBuilder API and added HeaderField and LiteralField

flamerohr:
Touches on:
* FormBuilder props and state merging
* FormReducer fixes lost form data on submit
* Form securityID missing without calling `$form->Fields()`


admin/client/src/components/HeaderField/HeaderField.js
```diff
@@ -0,0 +1,36 @@
...
+    return (
+      <div className="field">
+        <Heading {...this.getInputProps()} >{this.props.data.title}</Heading>
+      </div>
+    );
+  }
+
+  getInputProps() {
+    return {
+      className: ['hidden', this.props.extraClass].join(' '),
```
tractorcow: Probably should not be 'hidden'

forms/HeaderField.php
```diff
@@ -84,4 +70,30 @@ public function getAttributes() {
...
+
+	/**
+	 * Header fields heading level to be set
+	 *
+	 * @return array
+	 */
+	public function getSchemaDataDefaults() {
+		$data = parent::getSchemaDataDefaults();
+
+		$data['headingLevel'] = $this->headingLevel;
```
tractorcow: Properties specific to individual fields should probably be in $data[`data`][someprop]

forms/Form.php
```diff
@@ -286,6 +286,9 @@ public function __construct($controller, $name, FieldList $fields, FieldList $ac
 		$this->securityToken = ($securityEnabled) ? new SecurityToken() : new NullSecurityToken();
 
 		$this->setupDefaultClasses();
+
+		// ensure security token is included in the field list
+		$this->ensureSecurityIDField();
```
chillu: This changes the behaviour of `Form->disableSecurityToken()`. By default, a security token is required in forms, hence the logic would add the field (via `ensureSecurityIDField()`). If it's later disabled on the form instance, you would now also need to remove the field. Which is why we deferred adding the field until the last moment (since `getExtraFields()` is likely called as part of render). Admittedly it's fragile, but your pull request is an API change which makes the logic more consistent in one place, to make it less consistent in another.

forms/HeaderField.php
```diff
@@ -17,31 +17,17 @@ class HeaderField extends DatalessField {
...
+
+	protected $schemaComponent = 'HeaderField';
+
 	/**
 	 * @param string $name
-	 * @param null|string $title
+	 * @param string $title
 	 * @param int $headingLevel
 	 */
-	public function __construct($name, $title = null, $headingLevel = 2) {
```
chillu: Has the `new HeaderField($title, $headlingLevel)` (rather than `new HeaderField($name, $title, $headlingLevel)`) been deprecated? That'll break heaps of existing use cases, from my experience with SilverStripe project code.

flamerohr: probably @tractorcow would be better to answer this, he mentioned that keeping legacy functionality like this isn't ideal?

I had found it using $name as the title if the original title passed in ended up being a numeric value (which can happen, e.g. the case of file names)
originally, I had a `count($args) < 3` check, and maybe even a `is_numeric($args[1]) && $args[1] < 6` would probably suffice

tractorcow: I've checked that there are no core locations where this legacy behaviour was necessary.

It's acceptable for us to make breaking changes it it helps us better support the code in the future. In this case, removing the complex buggy support for legacy constructor is simpler than writing additional fixes to support numeric titles (which this code was crashing on).

flamerohr:
Recreating pull request because travis

Issue #5576, opened 2016-05-21
BUG DataList:canFilterBy() returns false for relations in dot notation


jonom:
`DataList:canFilterBy()` only checks immediate database fields, not fields on relations which the ORM is quite capable of filtering on.

I noticed this when attempting to use a filter on a gridfield that included the Title from a has_one relation in the summary fields. No filter field is displayed for that column. If I amend `DataList:canFilterBy()` to simply return true, the filter field shows up and works perfectly.

@tractorcow you're probably the authority on this subject is this something you could take a look at?

I suspect `DataList:canSortBy()` has a similar limitation.

![silverstripe_-_edit_page](https://cloud.githubusercontent.com/assets/1079425/15450056/4b4ee322-1f45-11e6-93ca-dc627a782f88.png)

tractorcow:
Yep, it could be extended to properly do a check for a composite filter.

What you would want is a variant of DataQuery::applyRelation(), but rather than joining on the entire relation, it simply iterates over the relation with a return true / false at the end. It's a lot of duplicate code though, but it's probably possible.

I suggest a `mirror` method `checkRelation that follows the same logic.

You probably want to restrict filtering to linear relations though, since datalist doesn't do a good job of joining on one-to-many tables (risks duplication of rows). You'll understand once you look at the code. :P

jonom:
@tractorcow thanks for your suggestions, I think I may have come up with something at https://github.com/jonom/silverstripe-framework/commit/1d4a053e84e0caf22d2b5321ebf8af293a6cf75c. For now I've just updated the `canFilterBy()` method but I can break it out so it can be re-used, e.g. for `canSortBy()`.

Questions:

1. Where would be the best place (class) to put a function that checks that a field/relation dot notation string is valid for a class? `DataObject`?
2. I'm thinking that the same check should work for both filtering and sorting... but to your knowledge are there cases where you could filter by a relation but not sort by it, or vice versa?
3. Any glaring issues in the solution I came up with?

tractorcow:
Oh, by the way, @dhensby has been working on something like this. See https://github.com/silverstripe/silverstripe-framework/pull/5659

Also things are not going well and it's been hard getting tests and things to pass...

tractorcow:
1. DataQuery
2. I think it's probably the same rule, so canSortBy can be re-used.
3. Probably won't work out of the box. See 2. as a substitute.

tractorcow:
1. DataQuery
[...]

PR #5884, opened 2016-08-11
FIX include related fields on filter/sort check

jonom:
Fix for #5576. Assume this is too late for 3.4.1 but I can target that branch instead if suitable.

### DIFF
--- a/model/DataList.php
+++ b/model/DataList.php
@@ -233,12 +233,24 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 	}
 
 	/**
+	 * Returns true if this DataList can be filtered by the given field.
 	 *
-	 * @param string $fieldName
+	 * @param string $fieldName (May be a related field in dot notation like Member.FirstName)
 	 * @return boolean
 	 */
 	public function canFilterBy($fieldName) {
-		if($t = singleton($this->dataClass)->hasDatabaseField($fieldName)){
+		$model = singleton($this->dataClass);
+		$relations = explode(".", $fieldName);
+		// First validate the relationships
+		$fieldName = array_pop($relations);
+		foreach ($relations as $r) {
+			$relationClass = $model->getRelationClass($r);
+			if (!$relationClass) return false;
+			$model = singleton($relationClass);
+			if (!$model) return false;
+		}
+		// Then check field
+		if ($model->hasDatabaseField($fieldName)){
 			return true;
 		}
 		return false;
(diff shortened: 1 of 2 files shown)


dhensby:
We'd have to make an rc2 if this went into 3.4.1, if we do a rc2 we'll it'd only be for bugs in rc1 so this will have to miss the 3.4.1 release, I'm afraid

dhensby:
Can we get a test for this, please?

jonom:
Yes, I jumped the gun there. Now that I'm adding tests though I'm getting lost... I though the ORM automatically handled joins etc. so you could just do e.g. `->sort('Member.FirstName ASC')` but doesn't look like that's the case :( It looks like you can do this with filters but not sorting? That seems inconsistent.

jonom:
Because of the lack of support for sorting on implicit joins, I've removed any sorting changes here and focused on filtering only. Think that `sort()` should be updated to support implicit joins though (for consistency and convenience), not sure if that should be a 3.x or 4 thing though? And probably outside of my capabilities.

dhensby:
> I though the ORM automatically handled joins etc

It does, but you need to call `applyRelation` on the data query for it to do the magic. I guess filter calls that method on `DataQuery` but `sort` doesn't?! It'll need to. Though I think the `DataQuery` is meant to ensure the sort is all joined up.

dhensby:
just *thinking* this through I can only imagine that it's going to be fairly complex to do a sort on dot notation (which is probably why it's not supported) becaues the dot notation needs to be interpretted and then converted to the actual table/column name that is joined. this is made extra complicated when taking into account the changes I'm hoping to make to fix bad joins...

jonom:
I'm getting out of my depth but I can open an issue about the lack of sorting on relations with dot notation if that will help :D

I would have imagined that join complexity would be similar between filtering and sorting... but that's too complex for me!

Thanks for the merge!

dhensby:
I think generally this aspect is very complex and that's why some features are missing / bugs are present.

For example, the issue of relations with a common parent table not working is because the work involved to get that to happen is complex and in-depth. The same will be true of sorting.

These really shouldn't be excuses for not having them, but someone does need to find the time to do it and implement it in a bug-free way. that's what I'm trying to do :P

kinglozzer:
Sorting on relations via dot notation exists in master: #4812 :)

jonom:
Hot dog!

dhensby:
I merged it and didn't even know!

tractorcow:
Nice job!

PR #5885, opened 2016-08-11
showqueries debugging tool now inserts parameters in place

dhensby:
Well, here's something to start a lively debate. Whilst I know in master we show the params at the end of the query and that there's an argument users might not realise their queries are parameterised and thus think this is literal queryies being run - I do think this is much more useful for debugging help.

A list of params later on is difficult to reconcile to the right `?` when there are a lot of params. (think `IN` clauses).

If we're really against this, can I suggest that we have `?showqueries=2` which is basically "fudge parameters" (ie: this change)?

see #4582

### DIFF
--- a/model/DB.php
+++ b/model/DB.php
@@ -282,7 +282,7 @@ class DB {
 	 *
 	 * @param array|integer $input An array of items needing placeholders, or a
 	 * number to specify the number of placeholders
-	 * @param string The string to join each placeholder together with
+	 * @param string $join The string to join each placeholder together with
 	 * @return string|null Either a list of placeholders, or null
 	 */
 	public static function placeholders($input, $join = ', ') {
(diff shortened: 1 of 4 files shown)


flamerohr:
I'm in favour of this, makes debugging easier for me where I can investigate the last query that had ran or cherry pick a particular query.

tractorcow:
How can you tell if the query you're debugging is parameterised or not unless you have the original query? This change would actually make it harder to track down errors in certain cases.

It's also misleading to alter what is being debugged during output.

I would suggest to limit this behaviour only to showqueries=inline (explicit option rather than a magic number) so that there's no risk of being mislead.

dhensby:
@tractorcow changed as per your suggestion. including updating docs

model/connect/Database.php
```diff
@@ -174,13 +175,29 @@ protected function previewWrite($sql) {
...
 	 * @return mixed Result of query
 	 */
-	protected function benchmarkQuery($sql, $callback) {
+	protected function benchmarkQuery($sql, $callback, $params = array()) {
 		if (isset($_REQUEST['showqueries']) && Director::isDev()) {
 			$starttime = microtime(true);
 			$result = $callback($sql);
 			$endtime = round(microtime(true) - $starttime, 4);
+			// replace parameters as closely as possible to what we'd expect the DB to put in
+			if (strtolower($_REQUEST['showqueries']) == 'inline') {
```
tractorcow: There's already code for this in https://github.com/silverstripe/silverstripe-framework/blob/3.2/model/queries/SQLQuery.php#L162, to save you having to double-implement this. :)

Your code also doesn't handle literal question marks in queries. E.g.

sql: `WHERE "Title" in ('?', ?)`
params: 'MyTitle'

dhensby: True, it doesn't handle that case well - that function on SQLQuery is, unfortunately, `protected` so I can't use it? but I can copy it...

tractorcow:
Shall we move this huge block (as well as the one in SQLQuery.php) into a DB helper method? `DB::inline_parameters`?

dhensby:
Done

model/DB.php
```diff
@@ -298,6 +298,60 @@ public static function placeholders($input, $join = ', ') {
 	}
 
 	/**
+	 * @param $sql string The parameterised query
```
tractorcow: Not correct phpdoc. :) `@param <type> $<name>`

PR #5886, opened 2016-08-11
BUG Fix usage of $this as closure argument

tractorcow:
Fixes issue in https://github.com/silverstripe-labs/silverstripe-sqlite3/pull/24


dhensby:
That's a bit of a sledge hammer approach... Are they not needed?

tractorcow:
It's a closure; There's no obligation to provide dummy parameters.

PR #5894, opened 2016-08-16
Remove DatabaseAdmin::autoBuild

### LINKED ISSUE #5843: DatabaseAdmin::autoBuild() doesn't work
This is in base 3.4.0, I haven't checked in earlier or later versions yet.

It calls a function called `getClassFile()` which no longer exists in the codebase. The method either needs to get refactored or removed.

tractorcow:
Fixes #5843 

It doesn't even work due to the missing getClassFile which was removed in 3.0.

### DIFF
--- a/ORM/DatabaseAdmin.php
+++ b/ORM/DatabaseAdmin.php
@@ -152,21 +152,6 @@ class DatabaseAdmin extends Controller {
 	}
 
 	/**
-	 * Check if database needs to be built, and build it if it does.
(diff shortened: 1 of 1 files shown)


madmatt:
:+1:

kinglozzer:
Merged as:

- Easy enough for devs to implement if they want it
- It won’t actually do a build if you just add a new file
- It doesn’t work anyway 😛

PR #5895, opened 2016-08-17
Add to campaign modal in React

flamerohr:
A few things covered:
* Added SingleSelectField component
* Add to campaign modal component
* Tweaked PopoverField to reduce duplicated props
* Flattened schema State
* Added FormActions within FormBuilder callback/handler

## TODO

 * [x] Regression: Fix 404 on "add to campaign" form submission in admin/pages
 * [x] Regression: Fix styling of "add to campaign" in admin/pages (looks weird in master with double title styles, looks differently weird in the pull request)
 * [x] More explicit success feedback in "add to campaign" for admin/assets: Slide down a "smiley face" like in admin/pages? At least, append something to body
 * [x] Fix centering of "more options" graphic in grey background
 * [x] Fix spacing in "more options" overlay, too much padding
 * [x] Make "more options" toggle keyboard selectable (with focus ring)
 * [x] Refactor `AddToCampaignModal` to `FormModal` - there's nothing "campaigny" about it, it should be generic. Might be easier to implement if we move the `openAddCampaignModal` redux state into a component state instead so we don't need to name/namespace the component

## Deferred
 * [ ] Disable button when nothing is selected (could be deferred until we have form validation in place)

For disabled button, it currently just prevents submit if no Campaign is selected. Unfortunately it isn't straight forward and a bit of work to have non-schema controlled state.


admin/client/dist/js/AddToCampaignForm.js
```diff
@@ -22,7 +22,7 @@
   }
 
   _jQuery2.default.entwine('ss', function ($) {
-    $('.add-to-campaign-action, #add-to-campaign__action').entwine({
+    $('#add-to-campaign__dialog .add-to-campaign-action,' + '.cms-content-actions .add-to-campaign-action,' + '#add-to-campaign__action').entwine({
```
chillu: Doing JS string concat here in addition to the comma separation mandated by CSS selector syntax seems weird - any particular reason?

chillu:
image


@clarkepaul Could you help with this?

admin/client/dist/js/AddToCampaignForm.js
flamerohr: yeah, linting was complaining about the src file going over 100 characters long for a line, so I broke each selector down to their own line...

I could do `$('select').add('another').add('more')` syntax instead, if that's preferred
it just looks funny with the string concat in the dist file

clarkepaul:
Sure, I reckon I can do that today.

clarkepaul:
@chillu fixed those minor css issues mentioned above

chillu:
Note to self: I think we should make modals reuseable with an approach like this - http://stackoverflow.com/questions/35623656/how-can-i-display-a-modal-dialog-in-redux-that-performs-asynchronous-actions/35641680#35641680

forms/PopoverField.php
```diff
@@ -47,19 +54,33 @@ public function setPopoverTitle($popoverTitle)
...
+	public function getPlacement()
+	{
+		return $this->placement;
+	}
+
+	public function setPlacement($placement)
+	{
+		$valid = ['top', 'right', 'bottom', 'left'];
+
+		if (in_array($placement, $valid)) {
```
chillu: That's a pretty nasty way to hide a fault parameter from devs (silently ignore) - I've made it raise an exception.

PR #5901, opened 2016-08-18
Cleaning up GridFieldAddExistingAutocompleter

dhensby:
- PHPDoc fixes
- Proper use of `Config::nest()`/`Config::unnest()`
- `return $this` on setters
- Minor optimisation on repeated `SSViewer` instantiations

### DIFF
--- a/forms/gridfield/GridFieldAddExistingAutocompleter.php
+++ b/forms/gridfield/GridFieldAddExistingAutocompleter.php
@@ -59,7 +59,7 @@ class GridFieldAddExistingAutocompleter
 	 *  	'Team.Name'
 	 *  )
 	 *
-	 * @var Array
+	 * @var array
 	 */
 	protected $searchFields = array();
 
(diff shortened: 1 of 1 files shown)


forms/gridfield/GridFieldAddExistingAutocompleter.php
```diff
@@ -90,7 +90,7 @@ public function __construct($targetFragment = 'before', $searchFields = null) {
 	/**
 	 *
 	 * @param GridField $gridField
-	 * @return string - HTML
+	 * @return string[] - HTML
```
tractorcow: It's string. array2json returns string right?

dhensby: The compressed diff is misleading you into thinking that the `Convert::array2json` is the return of the method this doc block is for (it's not). [See line 124](https://github.com/silverstripe/silverstripe-framework/pull/5901/files#diff-c2c534971b8d2554baeba1399c70d88aR124).

dhensby:
<3

PR #5904, opened 2016-08-20
'data' method as an alias for get, set and get_all

JorisDebonnet:
I like syntactic sugar. Would it be wrong to start supporting multi-signature methods? I would love to do this for Config as well. They are backwards compatible, and nobody has to use them if they don't like them.

### DIFF
--- a/control/Session.php
+++ b/control/Session.php
@@ -291,7 +291,7 @@ class Session {
 	 * Set a key/value pair in the session
 	 *
 	 * @param string $name Key
-	 * @param string $val Value
+	 * @param mixed $val Value
 	 */
 	public static function set($name, $val) {
 		return self::current_session()->inst_set($name, $val);
@@ -313,6 +313,23 @@ class Session {
 	 */
 	public static function get_all() {
 		return self::current_session()->inst_getAll();
+	}
+	
+	/**
+	 * Get or return variables, depending on method signature:
+	 *
+	 * Session::data()              returns Session::get_all()
+	 * Session::data($name)         returns Session::get($name)
+	 * Session::data($name, $value) returns Session::set($name, $value)
+	 */
+	public static function data() {
+		$args = func_get_args();
+		$num_args = count($args);
+
+		if(!$num_args) return self::get_all();
+		if($num_args == 1) return self::get($args[0]);
+		if($num_args == 2) return self::set($args[0], $args[1]);
+		return null;
 	}
 
 	/**


dhensby:
This is a new API, so at a minimum this needs to go against 3, I've changed the base branch (yay) for this PR to 3.

control/Session.php
```diff
@@ -314,6 +314,23 @@ public static function get($name) {
...
+	/**
+	 * Get or return variables, depending on method signature:
+	 *
+	 * Session::data()              returns Session::get_all()
+	 * Session::data($name)         returns Session::get($name)
+	 * Session::data($name, $value) returns Session::set($name, $value)
+	 */
+	public static function data() {
+		$args = func_get_args();
+		$num_args = count($args);
```
dhensby: `func_num_args()`

dhensby:
Personally, I'm not a huge fan of functions that change behaviour based on number of arguments because it means you need to have a much more extensive knowledge of the codebase to understand what's happening rather than seeing `get` or `update` and being able to guess at what's going on.

I'll wait for thoughts from the @silverstripe/core-team

control/Session.php
JorisDebonnet: Yeah, thought about that, but since I'd already used `func_get_args`, I figured `count` might make more sense ...

unclecheese:
I think the "it's backwards compatible, so take it or leave it" approach establishes a slippery slope toward a more bloated codebase with a bigger API surface and more documentation to maintain. I'm not saying I disagree with this, but I think that's something to consider. I'd be keen to hear from more members of the core team on where else we are doing things like this. Things that come to mind:

`Object::create('ClassName', ...args)` vs `ClassName::create(...args)`
`DataObject::get('ClassName', ...args)` vs `ClassName::get()->...`
`$myList->each($callback)` vs `foreach($myList...)`

In all of the above cases, I find the alias useful in the sense that it allows more declarative code, whereas ambiguous methods like `data()` feel very imperative to me.

flamerohr:
This leaves the code ambiguous if someone were to scan through the code and not necessarily want to dive into the Session class.

Another option, would be to suggest making the Session class extensible so that customisations is possible for the individual developer.
Or possibly using Injector to override the Session class with your custom class containing that function?

It's a nice idea, I don't necessarily agree with it to be in framework.

JorisDebonnet:
I realise the name 'data' is a bit ambiguous; I wanted `Session::var()` but `var` is a protected keyword (at least in php < 7). 

It may actually be more intuitive and consistent if we could just take the Session instance and then use the __set and __get methods (e.g. `$session->some_variable`). That isn't possible right now, right?.. It would make it consistent with most SilverStripe classes that "just hold data", which in the end is all Session does. 

Anyway, as you guessed I'm mostly trying to open the general debate for more intuitive API where possible :) The things @unclecheese mentions are excellent examples: I've been confused in the past why `MyDataObject::get_by_ud($id)` didn't work, and it seems silly to write `MyDataObject::get_by_id('MyDataObject', $id)` (which was my first working solution, although traditionally we'd do it with DataObject at the start).

unclecheese:
You actually wouldn't need a `Session` instance to use the magic of `__get` and `__set`. You could do it through the frontend using `__callStatic`.

`Session::someVariable();`
`Session::someVariable('new value');`

^ that's not going to get a lot of support, but it is an option, and it is more expressive, IMO.

tractorcow:
I dislike the idea of single methods having multiple roles. It means that code analyses requires additional context in order for it to make sense, and dilutes the purposes of any method. If a method can have 3 different behaviours, why not 4 or 5?

https://agiletribe.wordpress.com/2011/11/02/22-avoid-multi-purpose-methods/

ContentController already has a data() method, in case it hasn't been made aware.

I'm afraid I'm against this change.

JorisDebonnet:
It does have a singular role though: it "gives access to the data" stored in the object/Session :) I thought it wouldn't be _that_ controversial because a few other frameworks work in a similar way. 

But I wouldn't mind going for magic getter/setter methods instead, as these are already used in many SilverStripe classes. The only drawback is you need to obtain the instance then,... because the __callStatic solution, although interesting, does feel quite strange :)

dhensby:
OK - it looks like generally the feedback on this particular implementation is negative.

I wouldn't have a problem with those magic getter/setters on `Session` or `Cookie` (`Config` works that way) but because we use `Session` and `Cookie` as statics and not instances, it'd again be a bit counter to what we "normally" have.

PR #5905, opened 2016-08-22
latestPublished doc change

robingram:
I find the doc comment on this method to be particularly misleading. I keep coming back to it expecting it give me the data object that is the last published version but it actually returns a boolean indicating whether the current version is the last published or not.


dhensby:
From reading the code this simply checks if there is not a draft version. "is the latest version of this object published". This isn't what your amended doc implies.

robingram:
Agreed. I've updated the doc comment

PR #5906, opened 2016-08-22
Change order of fixtures in example

robingram:
From my testing it appears that the order in which objects are declared in fixture files is important. If you want to reference another object in the same fixture file it needs to be declared _before_ the referencing object. One of the example fixture files in the documentation would not work because it contains references to objects that have not yet been declared. I've switched the order of the fixtured objects to resolve this.


dhensby:
Thanks. You're right, order is important. It shouldn't be, though... Something to fix one day... :(

PR #5909, opened 2016-08-23
Fixing TabSet.js error when expected elements are not available

timkung:
Updating TabSet.js to open tabs correctly on page load when a hash is added to URL within the CMS.

This was done to fix a JS error found when clicking the "Manage roles" link inside the "Roles" tab when viewing a Group within the CMS. The error was:

`Uncaught TypeError: Cannot read property 'indexOf' of undefined`

It looks like the `openTabFromURL` method was looping over a JS selector that has been updated or is only used in some cases so when it tries to find the `href` property it can't (and hence calling `indexOf` on `this.href` produces the above error).

The suggested fix changes the following line as the `ajaxComplete` event is not being fired so needed to use another appropriate event instead:

`$(window).one('ajaxComplete', function () { ... });`

### DIFF
--- a/javascript/TabSet.js
+++ b/javascript/TabSet.js
@@ -1,5 +1,5 @@
-(function($){
-	$.entwine('ss', function($){
+(function($) {
+	$.entwine('ss', function($) {
 		/**
 		 * Lightweight wrapper around jQuery UI tabs for generic tab set-up
 		 */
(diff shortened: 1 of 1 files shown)


javascript/TabSet.js
```diff
@@ -52,7 +52,7 @@
 				}
 
 				// Switch to the correct tab when AJAX loading completes.
-				$(window).one('ajaxComplete', function () {
+				$(document).ready(function() {
```
dhensby: Is this change necessary? Why is document ready better than ajax complete?

timkung: Mentioned in the description above :)

But it seemed that event was firing either before this code runs, or not at all?

chillu:
@timkung Did you test this in IE8? The ajax bits were in there to ensure a tab remains active after saving (which replaces the entire HTML including tabs). `SecurityAdmin` and `ModelAdmin` shouldn't be your only test cases there, since the "top right" tabs are handled differently to the tabs below (part of the `<form>`. Can you test with a `Page` view that has multiple tabs?

timkung:
@chillu Have just tested this in IE8 and it seems to work as expected. And have also just tested a `Page` that has more than one tab, e.g. a "Root.Test" tab where `#Root_Test` appended to the URL opens the "Test" tab on page load.

chillu:
Yep, confirmed this works locally as well (by installing the https://github.com/silverstripe/silverstripe-frameworktest/ module which has some preconfigured tabs for `Page`)

PR #5917, opened 2016-08-26
Add icons to breadcrumbs

flamerohr:
Can provide an icon class to use and action for when the icon is clicked


dhensby:
Hmm, so duplicate code causes code-climate to fail @chillu can we make code climate a bit more forgiving?

chillu:
@dhensby I've already made it more forgiving with https://github.com/silverstripe/silverstripe-framework/commit/daaff886adea6222719ffc49d81fe2958ece75d5. Maybe Chris hadn't rebased on that fairly recent commit, and it was still using the old values? Anyway, tests are passing now so can't see what `mass_threshold` is was complaining about :)

flamerohr:
Yup, I ended merging two duplicate unit tests it was complaining about, and it needed a rebase to the latest anyway :)

flamerohr:
To give a little bit more context, this facilitates for changes made in https://github.com/silverstripe/silverstripe-asset-admin/pull/233

Giving the ability in breadcrumbs to have an action icon (e.g. an edit button)

PR #5918, opened 2016-08-28
Replace browserify/gulp with Webpack

sminnee:
Details in the commits, feedback welcome!

Pull requests:
https://github.com/silverstripe/silverstripe-cms/pull/1589
https://github.com/silverstripe/silverstripe-asset-admin/pull/249
https://github.com/silverstripe/silverstripe-testsession/pull/41
https://github.com/silverstripe/silverstripe-reports/pull/48
https://github.com/silverstripe-themes/silverstripe-simple/pull/42

## TODO

 * [x] Docs: Fix 4.0 changelog
 * [x] Docs: Fix https://docs.silverstripe.org/en/4.0/contributing/build_tooling/
 * [x] Docs: Fix https://docs.silverstripe.org/en/4.0/developer_guides/customising_the_admin_interface/how_tos/extend_cms_interface/
 * [x] Full Behat test run on CMS (with new framework branch)
 * [x] Fix CSS includes to fix UploadField styling
 * [x] Fix styling on show change password fields as closed by default (not referencing CSS correctly)
 * [x] Fix UploadField upload behaviour
 * [x] Update references to blueimp and jquery.sizes to reference node_module rather than thirdparty. Delete files from thirdparty. See https://github.com/silverstripe/silverstripe-framework/blob/master/gulpfile.js#L90 (note: TinyMCE stays in thirdparty for now)
 * [x] ~~Implement sprite generation~~ (decided it's no longer required)
 * [x] Update docs to mention sprite generation is considered deprecated
 * [x] Document how we use webfonts (SaaS, contact us if you want to add any)
 * [x] Test installer styling
 * [x] Doublecheck we've covered everything and remove Gulpfile.js
 * [x] fix cms "more options" menu popup alignment
 * [x] ~~Document dependencies being copied manually into thirdparty (e.g. TinyMCE)~~ Actually just removed `tinymce` from NPM deps, since we're not using it (only the manual copy in `thirdparty`). Can be added back once we implement my suggestion in comments below.
 * [x] Move JS-related thirdparty/ code into admin/thirdparty
 * [x] Test multilingual CMS

## Notes

 * npm run sanity has been removed, because the manual copying of libraries from src to dist is, by and large, not needed. TinyMCE is still included from `framework/thirdparty`


stevie-mayhew:
One think I notice is that it appears there isn't an explicit set to `NODE_ENV=production` for production builds. Will have a browse through the rest of it today.

chillu:
Hey Sam, this looks awesome!

Could we either move this to the open-sausages repo, or give me temporary commit rights on your fork? I could also send pull requests to your fork, but that seems a bit overkill :) 

Converted `asset-admin`, JS seems to work, figured it's the best way to learn. I can't get the CSS to compile. I've had to expose the new `FormBuilderModal` via `framework`. Also had to `npm-shrinkwrap` in `framework` in order to get `npm install` behaving. Check new commits on the open-sausages framework fork.

https://github.com/open-sausages/silverstripe-framework/tree/webpack
https://github.com/open-sausages/silverstripe-cms/tree/webpack
https://github.com/open-sausages/silverstripe-asset-admin/tree/webpack

I got a bit confused by the `expose` calls hidden away in `bundles/lib.js`, particularly since `webpack.config.js` has the equivalent `export` statements (in order to share modules *within* multiple bundles produced by framework). Should we move this to `webpack.config.js` completely? See https://github.com/webpack/expose-loader

```
module: {
  loaders: [
    { test: require.resolve("react"), loader: "expose?React" }
  ]
}
```

We have to fix the 4.0 changelog referencing build tooling as well as docs references to browserify:
https://docs.silverstripe.org/en/4.0/contributing/build_tooling/
https://docs.silverstripe.org/en/4.0/developer_guides/customising_the_admin_interface/how_tos/extend_cms_interface/

chillu:
@sminnee Done some more work on this (in the open-sausages branches). Kept your commits intact to make review easier, we'll need to squash a bit once you're happy with my modifications.

I now know why you didn't use `loaders` for the `expose` config - you need scripts to be loaded in an order dependant way - all good.

Added some TODOs to the PR description.

chillu:
Talked to Sam and decided to push my commits to `sminnee` instead of the `open-sausages` repo. Removed the `cms` and `framework` repos, kept the `asset-admin` one intact. Current work:

https://github.com/sminnee/silverstripe-framework/tree/webpack
https://github.com/sminnee/silverstripe-cms/tree/webpack
https://github.com/open-sausages/silverstripe-asset-admin/tree/webpack

client/src/legacy/UploadField_downloadtemplate.js
```diff
@@ -1,4 +1,4 @@
-window.tmpl.cache['ss-uploadfield-downloadtemplate'] = tmpl(
+tmpl.cache['ss-uploadfield-downloadtemplate'] = tmpl(
```
flamerohr: Could add `window.tmpl` like in `UploadField_uploadtemplate.js` for consistency, not a breaking change

sminnee:
@chillu I think that the partial revert of the dist rebuild is a bit excessive. Having done a search through the code for `` tags, I see the following:

framework/admin:

 * framework/admin/client/dist/images/spinner.gif (from CMSLoadingScreen.ss)

framework:

 * BBCodeParser references images but this code is broken, so I think it's safe to ignore.

cms:
 * cms/client/dist/images/dialogs/alert.gif (from Install_successfullyinstalled.ss)

installer, asset-admin, siteconfig, reports: none

chillu:
@sminnee I've pushed an update (and squashed). It's less about `` references, most of these images are referenced in SCSS files through `url()`. I've gone through each image in `client/dist/images` and `admin/client/dist/images` and checked for their use, and trimmed down accordingly. Everything that's left in there is actually in use (at least until Paul gets around doing more vectorised implementations). I've removed images from `dist/` which were inlined into the SCSS via data-urls.

Added a few TODOs at the top. What's your take on copying and sanity checking (diffing) third party code? Ideally we can get proper ES6 exports for those, not sure if anything has changed since we last looked at those particular libs.

@unclecheese We've talked about updating build tooling docs for SS4 on Twitter the other day. I've had to rewrite some of this in this PR as part of the Webpack switch, and restructured it a bit while I was going as well. Wanna have a read? Does the `webpack.config.js` setup generally make sense to you?

chillu:
Talked a bit more with Sam offline:

- Remove TinyMCE duplication in thirdparty, reference node_modules if possible
- Note why we removed spriting (new stuff should be webfonts)
- Document how we use webfonts (SaaS, contact us if you want to add any)

We've been able to remove nearly a THOUSAND FRICKEN NPM PACKAGES by switching to Webpack (and simplifying the build, e.g. no more spriting and coffeescript processing) - a full third less.

Before
```
npm ls | wc -l
3189
du -hc node_modules
763M	total
```

After
```
npm ls | wc -l
2301
687M	total
```

chillu:
Further to the point of loading TinyMCE via the PHP GZIP script: An [example of procedural plugin registration](https://github.com/tinymce/tinymce/blob/master/js/tinymce/plugins/autosave/plugin.js#L35), and the corresponding [add() API](https://github.com/tinymce/tinymce/blob/master/js/tinymce/classes/AddOnManager.js#L108).

If we keep a list of pre-bundled default plugins in our own `bundle-lib.js`, we could change `HtmlEditorConfig` to output a `<script>` tag which calls the corresponding [remove() API](https://github.com/tinymce/tinymce/blob/master/js/tinymce/classes/AddOnManager.js#L115) in TinyMCE on plugins which have been removed through PHP or YAML config. New plugins added through `HtmlEditorConfig` would be loaded as separate JS files by default, and use the procedural `add()`

chillu:
@flamerohr added a react-based mini-pattern-lib to frameworktest via https://github.com/silverstripe/silverstripe-frameworktest/pull/22. Which means we should migrate this to Webpack as well

chillu:
Facebook has released a `create-react-app` boilerplate which comes with a default `webpack.config.js` - worth a look: 

https://github.com/facebookincubator/create-react-app/blob/master/config/webpack.config.dev.js
https://github.com/facebookincubator/create-react-app/blob/master/config/webpack.config.prod.js

Forms/UploadField_SelectHandler.php
```diff
@@ -60,7 +60,7 @@ public function __construct($parent, $folderName = null)
 	public function index()
 	{
 		// Requires a separate JS file, because we can't reach into the iframe with entwine.
-		Requirements::javascript(FRAMEWORK_DIR . '/client/dist/js/UploadField_select.js');
+		Requirements::javascript(FRAMEWORK_DIR . '/admin/client/dist/js/UploadField_select.js');
```
tractorcow: webpack this and remove the call here?

tractorcow:
Noted that the cms popup menu is aligned wrong. I've added a bullet to the todos.

image

tractorcow:
I got this error running `npm run build`

```
ERROR in ./admin/client/src/bundles/lib.js
Module not found: Error: Cannot resolve module 'blueimp-file-upload/jquery.fileupload-ui.js' in /Users/dmooyman/Sites/ss40webpack/framework/admin/client/src/bundles
 @ ./admin/client/src/bundles/lib.js 58:0-54
 ```

**Comment by tractorcow** [comment]
For some weird reason, `ontabsbeforeactivate` in LeftAndMain.ActionTabSet.js isn't being triggered anymore.

**Comment by tractorcow** [comment]
It's caused by https://bugs.jqueryui.com/ticket/9316.

widgetEventPrefix is empty, so `onbeforeactivate` is triggered, but not `ontabsbeforeactivate`.

**Comment by chillu** [comment]
@sminnee I think we’re hitting this for Webpack’s TextExtract loader: https://github.com/webpack/extract-text-webpack-plugin/issues/179 - basically it silently fails to merge multiple entry points into a single CSS file. It works with a `[name]` placeholder, but then you end up with `bundle-legacy.css` etc. 

I'll comment out all the `require(*.scss)` statements for now, and move them all into `bundle.scss` as a separate entry point - then we can use `[name]` - it'll only create one CSS file because the other ones don't reference any SCSS

**Comment by sminnee** [comment]
Yeah good approach to get this over the line. We can refine later if we want.

**Comment by tractorcow** [comment]
I've had to back-port the above bugfix into our copy of jquery-ui.js, which is currently at 1.9. I attempted to upgrade to ui 1.10.4, but this would require us to re-do all our code which relied on .data() behaviour from 1.8, which is a breaking change.

**Comment by tractorcow** [comment]
I've fixed pretty much all that I can... the only remaining issue blocking tests passing is the `framework/client/src/styles/legacy` not being included in the dist css files.

**Comment by chillu** [comment]
OMG, I think this is actually ready to merge now! Behat and PHPUnit tests are passing locally for me. There's one test failure which is due to changed paths in the "simple" theme (https://github.com/silverstripe-themes/silverstripe-simple/pull/42, `InvalidArgumentException: File framework/thirdparty/jquery/jquery.js does not exist`).

File: Forms/HTMLEditor/HTMLEditorField_Embed.php [hunk 3]
```diff
@@ -106,7 +106,7 @@ public function getPreviewURL()
 		}
 
 		// Default media
-		return FRAMEWORK_DIR . '/images/default_media.png';
+		return FRAMEWORK_DIR . '/client/dist/images/src/default_media.png';
```
sminnee: should we squash this back to "FIX Webpack handles images & fonts"?

sminnee:
OK I've reviewed Ingo's changes to this and I'm happy, assuming the tests pass.

admin/client/src/legacy/ModelAdmin.js
```diff
@@ -3,6 +3,8 @@
  */
```
sminnee: I can't see how this file is actually getting included, and my suspicion is that the code is so edge-case that tests don't file if we remove it entire.

Can we see some evidence that the `.cms-content-tools #Form_SearchForm` entwine rule has been bundled?

If not, do we need to include these files in the bundle-legacy entrypoint?

sminnee:
OK all tests except the CMS ones pass. @tractorcow do you want to merge this?

Forms/HTMLEditor/HTMLEditorField_Embed.php
chillu: It's a bit hard to tell which commit this belongs to, since that line has been changed in the PR afterwards - I'd say just leave it here, unless you can point out a commit

admin/client/src/legacy/ModelAdmin.js
chillu: Good point, I've added it to `legacy.js` now and verified that the "expand csv stuff" feature works in ModelAdmin (so the JS kicks in)

tractorcow:
I've merged cms + testsession, restarting framework build.

sminnee:
@tractorcow we're green. :-) I'm merging

PR #5919, opened 2016-08-28
Fix Modal response positioning

flamerohr:
Addresses the issue mentioned here https://github.com/silverstripe/silverstripe-asset-admin/issues/234


clarkepaul:
Have you tried to get the success message positioned to the top of the modal (so it covers the modal-header) as in the designs? Might need a different positioning and then the "X" close positioned above the message so it is still accessible.

flamerohr:
I haven't, I'll give it a go when I next get a chance to. It may be that we don't use the modal provided header in that case, but I'll see when I get to it :)

chillu:
General comment: Please add ` (fixes #234)` to the commit message (ideally the summary) - see https://docs.silverstripe.org/en/4.0/contributing/code/#commit-messages. That'll auto-close the issue once the PR is merged which is quite convenient. And ensures there's enough context for later.

chillu:
Double checked in IE10, looks good. I'm not too sold on the whole `display: block` > `display: table` positioning magic (feels fragile), but it's already been set up this way, so won't block merge for it.

PR #5920, opened 2016-08-28
BUGFIX: ensure that tabs functions exists 

### LINKED ISSUE #5301: Reset button not submitting on ModelAdmin SearchForm.
When pressing the 'Reset' button on a ModelAdmin 'SearchForm', the form fails to submit due to the following line:
https://github.com/silverstripe/silverstripe-framework/blob/3/admin/javascript/LeftAndMain.js#L1334

This causes the following error:
`Uncaught Error: cannot call methods on tabs prior to initialization; attempted to call method 'select'`

Which makes perfect sense as there are no tabs on $('.cms-edit-form').
Some kind of check for tab existence should probably be put in place, before '.tabs('select',0)' is called.

stevie-mayhew:
Before trying to call function which uses them. Fixes #5301

### DIFF
--- a/admin/javascript/LeftAndMain.js
+++ b/admin/javascript/LeftAndMain.js
@@ -1334,6 +1334,7 @@ jQuery.noConflict();
 				}
 
 				var container = this.closest('.cms-container');
+				container.find('.cms-edit-form').tabs();
 				container.find('.cms-edit-form').tabs('select',0);  //always switch to the first tab (list view) when searching
 				container.loadPanel(url, "", {}, true);
 


dhensby:
Interesting... How does this solve the issue?

flamerohr:
Maybe check if `('.cms-edit-form')` has tabs applied instead of applying tabs straight out? I think it was `$elem.data("tabs")` for jQueryUI.
I guess this stems from the scaffolder for DataObjects defaults to always having tabs, and not normally seen to have a form without at least 'Root.Main' container

stevie-mayhew:
@dhensby it makes sure a tabset exists for it to query. That tabset can then return false rather than being an undefined function.

@flamerohr I don't know if I want to code in a specific use case action rather than applying a generalised fix which will work in other situations apart from this one?

dhensby:
Would a `try {} catch` block work? It just feel nasty to instantiate it like this...

flamerohr:
I don't know about applying tabs in that area for `.cms-edit-form`, might be a reason it's not instantiated? ;)
Checking if it has been instantiated is less intrusive

But anyway, if it needs to be instantiated always, then I'd rather point to putting it in the `$('.cms-edit-form').entwine({onmatch: function () { /* ...here? */ });` so you'd have keep the tabs manipulation in a similar scope.
Javascript is so evil in that way...

dhensby:
@stevie-mayhew are you in a position to get this over the line or shall we close?

tractorcow:
No reply for a couple of months, I'll close for now, but you can re-open it if you want @stevie-mayhew

PR #5932, opened 2016-08-31
Checks if $this->value is iterable

olyism:
Added additional check to ensure `$this->value` is iterable.

I found this bug when creating a new permission to allow non-ADMIN users to add groups.

When a non-ADMIN user created a new group I got the _invalid argument for foreach_ error message.

In light of this I've applied the check in line 282, as well as 300.

### DIFF
--- a/src/Security/PermissionCheckboxSetField.php
+++ b/src/Security/PermissionCheckboxSetField.php
@@ -9,6 +9,7 @@ use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataObjectInterface;
 use SilverStripe\ORM\SS_List;
+use Traversable;
 
 /**
  * Shows a categorized list of available permissions (through {@link Permission::get_codes()}).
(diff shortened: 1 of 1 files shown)


tractorcow:
It could be iterable, but not an array. E.g. `ArrayList`. You should also check `instanceof Traversable` http://php.net/manual/en/class.traversable.php

dhensby:
It'd be good if framework had some kind of "is iterable" helper, because lots of things can be iterated over

Security/PermissionCheckboxSetField.php
```diff
@@ -279,7 +279,7 @@ public function saveInto(DataObjectInterface $record) {
 		// Remove all "privileged" permissions if the currently logged-in user is not an admin
 		$privilegedPermissions = Permission::config()->privileged_permissions;
 		if(!Permission::check('ADMIN')) {
-			foreach($this->value as $id => $bool) {
+			if (is_array($this->value)) foreach($this->value as $id => $bool) {
```
mikenz: If PSR2 is a goal then this short hand should be phase out. A full `if` block is needed to pass.

tractorcow: Yep that's correct. We aren't PSR-2 yet, so I won't enforce it in this PR.

dhensby:
needs rebasing now

sminnee:
Let's just use `is_array($this->value) || $this->value instanceof Traversable` for now rather than bulk out the global API.

tractorcow:
> Let's just use is_array($this->value) || $this->value instanceof Traversable for now rather than bulk out the global API.

Yeah, this is really something PHP should be addressing itself. Maybe 7.2? :P

dhensby:
bump

tractorcow:
Rebased and switched to 4.0 branch.

tractorcow:
Will merge on green.

PR #5921, opened 2016-08-29
API Apply all SilverStripe namespaces to framework

tractorcow:
Closes #5637

Merge with these:

 - [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/243
 - [ ] https://github.com/silverstripe/silverstripe-cms/pull/1602
 - [ ] https://github.com/silverstripe/silverstripe-reports/pull/45
 - [ ] https://github.com/silverstripe/silverstripe-siteconfig/pull/36
 - [x] https://github.com/silverstripe/silverstripe-sqlite3/pull/25
 - [x] https://github.com/silverstripe/silverstripe-installer/pull/135
 - [x] https://github.com/silverstripe/silverstripe-behat-extension/pull/125
 - [ ] https://github.com/silverstripe/silverstripe-behat-extension/pull/126
 - [x] https://github.com/silverstripe/silverstripe-postgresql/pull/55
 - [ ] https://github.com/silverstripe/silverstripe-postgresql/pull/56
 - [x] https://github.com/silverstripe/silverstripe-mssql/pull/37
 - [ ] https://github.com/silverstripe/silverstripe-frameworktest/pull/21
 - [x] https://github.com/silverstripe/silverstripe-testsession/pull/38

I've created this PR early for peer review at @chillu's request. It's not yet ready to be merged.

This PR includes several smaller pieces of work, including:
 - Remove deprecated code (to save having to upgrade / namespace them)
 - Change in behaviour of FormField::_templates() to use the standard class hierarchy lookup.
 - Change SSViewer::get_templates_by_class() to return un-filtered candidate list (performance reasons)
 - Remove `@package` / `@subpackage` - these are no longer necessary within the namespaced hierarchy.
 - Broke each class into a class-per-file, and named appropriately to support psr-4 autoloading.
 - `Diff` library moved into third party, to avoid having to namespace it. `SilverStripe\View\Parsers\Diff` extends the thirdparty lib to encapsulate ss-specific functionality.
 - Cleanup of incorrect PHPDoc, code style improvements and fixes, some tidy up of inconsistent APIs (e.g. invalid method parameters and return types).
[...]

PR #5940, opened 2016-09-02
API Change behaviour of filter API to support injected search filter classes

tractorcow:
In order to support ambiguous syntax (e.g. `Field:Filter`,  `Field:Filter:modifier`, `Field:modifier`) I've updated the behaviour of SearchFilter to check the case of any supplied argument. Previously, this would use class_exists to resolve ambiguity, but with namespaced classes (and injector aliases) this isn't possible.

Broken out of https://github.com/silverstripe/silverstripe-framework/pull/5921

### DIFF
--- a/ORM/DataList.php
+++ b/ORM/DataList.php
@@ -2,6 +2,9 @@
 
 namespace SilverStripe\ORM;
 
+use BadMethodCallException;
+use SearchFilter;
+use SilverStripe\ORM\Queries\SQLConditionGroup;
 use ViewableData;
 use Exception;
 use InvalidArgumentException;
@@ -34,6 +37,7 @@ use ArrayIterator;
  * @subpackage orm
  */
 class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortable, SS_Limitable {
+
 	/**
 	 * The DataObject class name that this data list is querying
 	 *
@@ -403,12 +407,9 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 	public function addFilter($filterArray) {
 		$list = $this;
 
-		foreach($filterArray as $field => $value) {
-			$fieldArgs = explode(':', $field);
-			$field = array_shift($fieldArgs);
-			$filterType = array_shift($fieldArgs);
-			$modifiers = $fieldArgs;
-			$list = $list->applyFilterContext($field, $filterType, $modifiers, $value);
+		foreach($filterArray as $expression => $value) {
+			$filter = $this->createSearchFilter($expression, $value);
+			$list = $list->alterDataQuery(array($filter, 'apply'));
 		}
 
 		return $list;
@@ -454,21 +455,7 @@ class DataList extends ViewableData implements SS_List, SS_Filterable, SS_Sortab
 			$subquery = $query->disjunctiveGroup();
 
 			foreach($whereArguments as $field => $value) {
-				$fieldArgs = explode(':',$field);
-				$field = array_shift($fieldArgs);
-				$filterType = array_shift($fieldArgs);
-				$modifiers = $fieldArgs;
-
-				if($filterType) {
-					$className = "{$filterType}Filter";
-				} else {
-					$className = 'ExactMatchFilter';
-				}
-				if(!class_exists($className)){
-					$className = 'ExactMatchFilter';
-					array_unshift($modifiers, $filterType);
-				}
-				$filter = Injector::inst()->create($className, $field, $value, $modifiers);
+				$filter = $this->createSearchFilter($field, $value);
 				$filter->apply($subquery);
 			}
 		});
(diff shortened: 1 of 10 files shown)


madmatt:
Using uppercase/lowercase checks seems quite painful/confusing...

Doesn't it make more sense to pass in filter classes if you want something that's more custom than a simple filter? Something like `$list->filter(new NegationFilter(new ExactMatchFilter()), 'test')`. It supports custom filters, and more importantly filter chains, a lot nicer, but to be fair is more painful to use

ORM/DataList.php
```diff
@@ -550,30 +535,48 @@ protected function isValidRelationName($field) {
...
+		$modifiers = $fieldArgs;
+		if(!$secondArg) {
+			// No filter defaults to ExactMatchFilter
+			// E.g. `Name`
+			/** @skipUpgrade */
+			$filterClass = 'ExactMatchFilter';
+		} elseif(ucfirst($secondArg) === $secondArg) {
+			// Upper case pattern means filter class identifier
+			// E.g. `Name:ExactMatch` or `Name:ExactMatchFilter:not`
+			$filterClass = "{$secondArg}Filter";
```
chillu: The second example here wouldn't work, right? It'd become `Name:ExactMatchFilterFilter:not`

docs/en/02_Developer_Guides/00_Model/06_SearchFilters.md
```diff
@@ -36,6 +36,10 @@ These suffixes can also take modifiers themselves. The modifiers currently suppo
 comparison uses the database's default. For MySQL and MSSQL, this is case-insensitive. For PostgreSQL, this is 
 case-sensitive.
 
+Note that all search filters (e.g. `:PartialMatch`) must be written in title case,
```
chillu: I'm with Matt on this one - the upper/lower casing sensitivity is confusing. Looks like you're introducing a new feature which requires this? Before you had to write `MyField:ExactMatch:case`, now you can write `MyField:case`?

chillu: That'll cause hard-to-debug exceptions - for example, a wrong casing like `MyField:partialmatch` would've satisfied `class_exists('partialmatchFilter')` before, but now throws `ExactMatchFilter does not accept partialmatchFilter as modifiers` (http://stackoverflow.com/questions/15488632/php-class-exists-acting-case-sensitive)

See http://stackoverflow.com/questions/15488632/php-class-exists-acting-case-sensitive

docs/en/02_Developer_Guides/12_Search/02_FulltextSearch.md
```diff
@@ -71,11 +71,14 @@ Example DataObject:
 Performing the search:
 
 	:::php
-	SearchableDataObject::get()->filter('SearchFields:fulltext', 'search term');
```
chillu: The fact that even our docs haven't cared about casing should give you an indication of how widespread this problem will be

tractorcow:
Yeah the solution isn't ideal, but with namespacing coming up, we can't rely on simply class_exists() with short user-given strings. It would be nice if there was another solution. Maybe if you provide a modifier, the filter is mandatory? E.g. you couldn't do `Field:not` anymore, but would have to do `Field:ExactMatch:not`? I still feel the "use case to disambiguate" is the simplest of the solutions, though.

docs/en/02_Developer_Guides/00_Model/06_SearchFilters.md
tractorcow: >  Before you had to write MyField:ExactMatch:case, now you can write MyField:case?

No, before you COULD write `MyField:case`, and it would know that `:case` was a modifier because `class_exists('casefilter')` was false.

tractorcow: > That'll cause hard-to-debug exceptions -

The exceptions will give output similar to. E.g. `partialmatch is not a valid modifier on ExactMatchFilter`. That is enough to debug the issue.

ORM/DataList.php
tractorcow: Yes incorrect docs. I'll fix it.

tractorcow:
Second alternative; I change the class_exists() to a check that also includes checking injector registered services.

However, that would then require an `Injector::inst()->has("{$class}Filter")` API.

madmatt:
Of the two options, I'd probably prefer your first option - enforcing the selection of a `SearchFilter`.

So if you previously had `->filter('Title:not', 'A phrase')`, you'd need to do `->filter('Title:ExactMatch:not', 'A phrase')`.

This is less ambiguous, which means clearer code written by developers, and seems like a simple CHANGELOG entry to add.

tractorcow:
@ingo can you please give your vote? I'll update the PR if you are happy to go with this suggestion.

docs/en/02_Developer_Guides/00_Model/06_SearchFilters.md
chillu: OK, I've misunderstood the existing behaviour, but verified now that `MyField:case` is indeed already a feature (https://github.com/silverstripe/silverstripe-framework/blob/master/tests/model/DataListTest.php#L722). In that case, yeah ... I agree with your solution. The alternative would be either introducing new syntax (other than `:`), or using a more explicit object structure as Matt suggests. Neither of these are MVP here, so I'm good with the fix as-is.

sminnee:
-1 for using case sensitivity to mean anything.

`Field:modifier` implicitly means `Field:ExactMatch:modifier`, and the ExactMatch filter only accepts these modifiers: 'not', 'nocase', 'case'

As a short-term fix, can we just have a hard-coded list of 'not', 'nocase', 'case' be special words?

tractorcow:
I guess what we could do is, if the second argument is ambiguous, check if the argument matches any of the modifiers on the default filter (ExactMatchFilter). In other words, as @sminnee suggests, validate the modifier, rather than the class.

sminnee:
As a slightly better fix, perhaps we can add `array SearchFilter::getAllowedModifiers()` and use `ExactMatchFilter::getAllowedModifiers()` to instead of the hardcoded list of `not`, `nocase`, `case`.

Related, there might a better way of defining search filters. A service name of the form `SearchFilter.ExactMatch` might make more sense than `ExactMatchFilter`.

Does the injector give us a way of seeing if a given service is defined? If so, then you use that as the check rather than class_exists

tractorcow:
> Does the injector give us a way of seeing if a given service is defined? If so, then you use that as the check rather than class_exists


No it doesn't, but defining this was one of my suggestions. It would be a bit more complex though.

dhensby:
> As a slightly better fix, perhaps we can add array SearchFilter::getAllowedModifiers()

The problem with this over hard coding the 3 current ones is that no developer will know what the reserved words are anymore and thus avoiding collisions with modified names becomes harder.

tractorcow:
@dhensby I think the risk of collision between the modifiers and the SearchFilter classes is quite low.

tractorcow:
Related : https://github.com/silverstripe/silverstripe-framework/issues/5948

I've spoken with @sminnee and we've opted to use a more injector-heavy solution to this problem.

ORM/DataList.php
```diff
@@ -550,30 +537,40 @@ protected function isValidRelationName($field) {
...
+			// Whether this is a valid modifier on the default filter, or a filter itself.
+			/** @var SearchFilter $defaultFilterInstance */
+			$defaultFilterInstance = Injector::inst()->get('DataListFilter.default');
+			if (in_array(strtolower($secondArg), $defaultFilterInstance->getSupportedModifiers())) {
+				// Treat second (and any subsequent) argument as modifiers, using default filter
+				$filterServiceName = 'DataListFilter.default';
+				array_unshift($modifiers, $secondArg);
+			} else {
+				// Second argument isn't a valid modifier, so assume is filter identifier
+				$filterServiceName = "DataListFilter.{$secondArg}";
```
sminnee: Do you think we should use `Injector::inst()->hasService($filterServiceName)` to throw an error if a service isn't defined?

ORM/DataList.php
```diff
@@ -403,12 +405,10 @@ public function filter() {
...
 
-		foreach($filterArray as $field => $value) {
-			$fieldArgs = explode(':', $field);
-			$field = array_shift($fieldArgs);
-			$filterType = array_shift($fieldArgs);
-			$modifiers = $fieldArgs;
-			$list = $list->applyFilterContext($field, $filterType, $modifiers, $value);
+		foreach($filterArray as $expression => $value) {
+			list($field, $filterClass, $modifiers) = $this->parseFilterArguments($expression);
+			$filter = Injector::inst()->create($filterClass, $field, $value, $modifiers);
```
sminnee: It makes more sense to me to have a single field that take `$key, $value` as an argument and returns a configured SearchFilter object.

E.g.: `$filter = $this->createSearchFilter($expression, value)`

This should work in the 2nd call.

ORM/DataList.php
tractorcow: It doesn't work if the service doesn't exist yet, but a config is available for it.

ORM/DataList.php
tractorcow: Yeah that would be fine. :P

ORM/DataList.php
sminnee: Oh, right, meh. :-P

tractorcow:
Updated based on your feedback @sminnee

sminnee:
As a companion to this PR should we make the injector system case-insensitive, to match the case-insensitivity of PHP classes?

tractorcow:
It would be hard to make case insensitivity work for things such as service identifiers, which don't actually map to physical classes. Otherwise we could normalise them using `ClassInfo::class_name`.

tractorcow:
Green!

chillu:
> Otherwise we could normalise them using ClassInfo::class_name

Aren't we moving away from `ClassInfo` eventually, so we can have fast PSR autoloading without flush? Topic for a different day, but until we've discussed its feasibility I wouldn't add more dependencies on `ClassInfo`.

Anway, looks great otherwise!

PR #5943, opened 2016-09-02
API Update template lookup to late resolution for performance reasons

tractorcow:
This proposed update to templating behaviour addresses a few issues.

Broken out of https://github.com/silverstripe/silverstripe-framework/pull/5921

Merge with https://github.com/silverstripe/silverstripe-cms/pull/1598

*Performance*

As SSViewer::get_templates_by_class() invokes SSViewer::has_template on EVERY item in the chain (even if we have higher precedence templates) it will end up doing redundant work. This is because SSViewer::setTemplate will choose only the first template in any given list that exists within the set of themes.

To address this, instead of doing any pre-filtering, we simply generate the list of all candidate templates, regardless of whether they exist, and rely on setTemplate() to do the final candidate selection.

SSViewer::chooseTemplate() has been refactored out in order to support cases where you need to know if a valid template is available, or to force template choosing.

*Consistency*

Previously, `forms` wasn't a standard type, as `Includes` or `Content` types were. This means that SSviewer::get_templates_by_class() couldn't be used at all for formfields due to it only supporting a template suffix, and not a `forms/` prefix.

When form fields are templated, this would mean that we could have duplicate directories for each namespace (one under a top level `forms/` folder and one at root). This change in behaviour flips the `forms` back into a template type, so that we can use the same lookup inheritance chain to search for templates.

A lot of gridfield components with hard (and semi-hard) coded templates have been updated.

ClassInfo has also been updated to clarify where you can use an instance of an object in place of a class name.

GridFieldItemEditView.ss has been removed as in all cases, this template is overridden and is redundant.


tractorcow:
It looks like this breaks the CMS in a weird way. the save button is gone in profile edit, and you can't scroll.

tractorcow:
Regressions have been resolved; there was a missing 'forms' type required for custom formfield templates, and I wasn't respecting full templates being passed into ThemeResourceLoader::findTemplate() properly (new tests added).

tractorcow:
I've updated this PR to remove the `forms` type from the template lookup.

Merge with https://github.com/silverstripe/silverstripe-cms/pull/1598.

tractorcow:
Green and ready for merge. :)

chillu:
Resolved merge conflict in changelog and force pushed

admin/code/SecurityAdmin.php
```diff
@@ -165,7 +165,7 @@ public function getEditForm($id = null, $fields = null) {
 		// Add import capabilities. Limit to admin since the import logic can affect assigned permissions
 		if(Permission::check('ADMIN')) {
 			$fields->addFieldsToTab('Root.Users', array(
-				new HeaderField(_t('SecurityAdmin.IMPORTUSERS', 'Import users'), 3),
+				new HeaderField('ImportUsersHeader', _t('SecurityAdmin.IMPORTUSERS', 'Import users'), 3),
```
chillu: Technically not part of the PR topic (fix templates), but all good :)

api/RSSFeed.php
```diff
@@ -240,6 +240,22 @@ public function setTemplate($template) {
 	public function getTemplate() {
 		return $this->template;
 	}
+
+	/**
+	 * Returs the ordered list of preferred templates for rendering this form
```
chillu: The PHPDoc is misleading, it refers to `Form` in the `RSSFeed` class (and misspelled "Returns")

forms/gridfield/GridFieldDetailForm.php
```diff
@@ -752,6 +752,21 @@ public function getTemplate() {
 	}
 
 	/**
+	 * Get list of templates to use
+	 *
+	 * @return array
+	 */
+	public function getTemplates()
+	{
+		$templates = SSViewer::get_templates_by_class($this, '', __CLASS__);
```
chillu: Right, `__CLASS__` refers to the base class where the method context was defined (http://php.net/manual/en/language.constants.predefined.php). Was wondering if this breaks on subclasses which wouldn't look for templates matching parent classes, but we're good.

view/SSViewer.php
```diff
@@ -861,49 +861,47 @@ public static function set_theme($theme) {
 	/**
 	 * Traverses the given the given class context looking for templates with the relevant name.
```
chillu: That description is a bit ambiguous now, we should clarify that it returns a list of potentially existing templates - it doesn't actually *look* for templates any more.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -63,6 +63,7 @@
  * Search filter classes (e.g. `ExactMatchFilter`) are now registered with `Injector`
    via a new `DataListFilter.` prefix convention.
    see [search filter documentation](/developer_guides/model/searchfilters) for more information.
+ * FormField templates no longer look in the 'forms' folder for templates.
```
chillu: Should mention the changed semantics of `SSViewer:: get_templates_by_class()` (returning all templates regardless of existence)

chillu:
`Email@parseVariables`is still using `$template = new SSViewer('email/'.$this->ss_template)` - does that need to be fixed?

forms/gridfield/GridFieldDetailForm.php
tractorcow: Yeah, I've started using this extensively to save having to use literal class strings (which are undesirable due to their difficulty to detect and upgrade).

tractorcow:
All above addressed. How does it look @chillu? Do the documentation updates mesh with your view? :D

docs/en/04_Changelogs/4.0.0.md
tractorcow: I've now added this a few lines down.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -415,8 +419,14 @@ require manual intervention, please see the below upgrading notes.
 Templates are now much more strict about their locations. You can no longer put a template in an arbitrary
 folder and have it be found. Case is now also checked on case-sensitive filesystems.
 
-Either include the folder in the template name (`renderWith('Field.ss')` => `renderWith('forms/Field.ss')`), move the 
-template into the correct directory, or both.
+Either include the folder in the template name (`renderWith('MyEmail.ss')` => `renderWith('emails/MyEmail.ss')`),
+move the template into the correct directory, or both.
+
+When using `<% include %>` template tag you should continue to leave out the `Includes` folder.
```
chillu: This advice is only accurate until we implement the suggestions in https://github.com/silverstripe/silverstripe-framework/issues/5952#issuecomment-244881303 (where `<% include Includes/Sidebar %>` will become the new and consistent default)

forms/gridfield/GridFieldPrintButton.php
```diff
@@ -121,8 +121,10 @@ public function handlePrint($gridField, $request = null) {
 		Requirements::css(FRAMEWORK_DIR . '/client/dist/styles/GridField_print.css');
 
 		if($data = $this->generatePrintData($gridField)){
-			return $data->renderWith("GridField_print");
+			return $data->renderWith(get_class($gridField)."_print");
```
chillu: Wouldn't this break on subclasses? `renderWith()` itself does not have a concept of a class hierarchy traversal to find "parent class templates", and `get_class()` always retrieves the concrete class of the instance.

How about this? I don't see a clean way to identify the "base class" of `GridField` without hardcoding here.
```
$templates = SSViewer::get_templates_by_class($gridField, '_print', 'GridField);
return $data->renderWith($templates);
```

Happy to merge either way, given that `GridField` is usually not extended directly it's a bit of an edge case.

PR #5944, opened 2016-09-02
API Allow has_many fixtures to be declared with array format as well as many_many

tractorcow:
CC @dhensby who has struggled with this problem in the past. :D

Broken out of #5921


sminnee:
Seems fine to me.

dhensby:
No back port for 3? :(

tractorcow:
It can be back-ported if you want to. It needs to go into master immediately though.

dhensby:
Does that mean I have to back port? :P

dhensby:
Well, that was easy - it was just your silly short array syntax that broke tests - see PR #5988

PR #5951, opened 2016-09-05
Docs about changed include path use

chillu:


dhensby:
Is this the case or are we peocideing backward compat?

chillu:
Closing this until https://github.com/silverstripe/silverstripe-framework/issues/5952 is resolved

PR #5953, opened 2016-09-05
Added a helpful invalid batch action exception message

assertchris:
screen shot 2016-09-06 at 5 31 00 pm

screen shot 2016-09-06 at 5 38 39 pm

### DIFF
--- a/admin/code/CMSBatchActionHandler.php
+++ b/admin/code/CMSBatchActionHandler.php
@@ -179,7 +179,12 @@ class CMSBatchActionHandler extends RequestHandler {
 		// Find the action handler
 		$actions = $this->batchActions();
 		if(!isset($actions[$name]['class'])) {
-			throw new InvalidArgumentException("Invalid batch action: {$name}");
+			/** @var Closure $message */
+			$message = require __DIR__ . '/messages/InvalidBatchAction.php';
+
+			throw new InvalidArgumentException(
+				$message($actions[$name])
+			);
 		}
 		return $this->buildAction($actions[$name]['class']);
 	}
@@ -248,6 +253,15 @@ class CMSBatchActionHandler extends RequestHandler {
 		$actions = $this->config()->batch_actions;
 		$recordClass = $this->recordClass;
 		$actions = array_filter($actions, function($action) use ($recordClass) {
+			if (!isset($action['recordClass'])) {
+				/** @var Closure $message */
+				$message = require __DIR__ . '/messages/InvalidBatchAction.php';
+
+				throw new InvalidArgumentException(
+					$message($action)
+				);
+			}
+
 			return $action['recordClass'] === $recordClass;
 		});
 		return $actions;
--- a/admin/code/messages/InvalidBatchAction.php
+++ b/admin/code/messages/InvalidBatchAction.php
@@ -1,0 +1,23 @@
+<?php
+
+return function(array $action) {
+	$action = trim(print_r($action, true));
+
+	return "
+
+There seems to be an error with the format of one of the batch actions:
+
+{$action}
+
+This usually happens because the action has been incorrectly formatted, in YAML config. The format should resemble:
+
+my_batch_action:
+  class: BatchActionClass
+  recordClass: ClassToApplyBatchActionTo
+
+Alternatively, you can register the action, in mysite/_config.php, with:
+
+CMSBatchActionHandler::register('my_batch_action', 'BatchActionClass', 'ClassToApplyBatchActionTo');
+
+";
+};
--- a/admin/tests/ExceptionMessagesTest.php
+++ b/admin/tests/ExceptionMessagesTest.php
@@ -1,0 +1,33 @@
+<?php
+
+/**
+ * @package framework
+ * @subpackage tests
+ */
+class ExceptionMessagesTest extends FunctionalTest {
+	public function testInvalidBatchActionExceptionMessage() {
+		$batchActionHandler = new CMSBatchActionHandler(new Controller(), 'test', 'SiteTree');
+
+		Config::inst()->update('CMSBatchActionHandler', 'batch_actions', array(
+			'foo' => array(
+				'invalid' => 'definition',
+			),
+		));
+
+		$throws = false;
+
+		try {
+			$reflection = new ReflectionClass('CMSBatchActionHandler');
+
+			$method = $reflection->getMethod('actionByName');
+			$method->setAccessible(true);
+			$method->invokeArgs($batchActionHandler, array('foo'));
+		}
+		catch (InvalidArgumentException $e) {
+			$throws = true;
+			$this->assertContains('There seems to be an error', $e->getMessage());
+		}
+
+		$this->assertTrue($throws);
+	}
+}


sminnee:
Use of a PHP include as a kind of global function that grabs '$action' from its parent scope is a little strange to me.

Maybe I've been coding too much JS lately but I'd kind of prefer returning a closure that takes $action as an argument.

flamerohr:
I like this, but I only err on the side of using php templates. Where formally the templates folder only contained *.ss files.

I suggest a way to have this as a translation entry? Or perhaps as a SS template?
or maybe a link to a documentation file 😄 ?

assertchris:
context: screen shot 2016-09-06 at 5 16 33 pm

sminnee:
A sketch of returning a callable instead of a string:

```php
// main.php

<?php

echo phpTemplate('template1.php', 'the action');
echo phpTemplate('template2.php', ['action' => 'the action']);

// or

$template = require('template1.php');
echo $template('the action');

function phpTemplate($file) {
	$args = func_get_args();
	array_shift($args);
	$closure = require($file);
	return call_user_func_array($closure, $args);	
}

// --- 
// template1.php

<?php

return function($action) {
	return "This is my template for $action";
};

// --- 
// template2.php

<?php

return function($params) {
	return "This is my template for ${params['action']}\n";
};
```

assertchris:
Related to closures that take arguments: given the required support for 5.3, extracting exception-generating methods into something protected/private is not compatible. A previous version of this PR (sorry Travis!) attempted to use such an approach. The alternative to repeating _a lot_ of message-generation code is to include the message file in multiple places...

admin/code/CMSBatchActionHandler.php
```diff
@@ -179,7 +179,11 @@ protected function actionByName($name) {
 		// Find the action handler
 		$actions = $this->batchActions();
 		if(!isset($actions[$name]['class'])) {
-			throw new InvalidArgumentException("Invalid batch action: {$name}");
+			$action = $actions[$name];
+
+			throw new InvalidArgumentException(
+				require __DIR__ . '/../templates/Exceptions/InvalidBatchActionException.php'
```
sminnee: I don't think the template folders is the right place to keep this. In 99.99% of other cases, templates mean ".ss template rendered by SSViewer used to create html code". At best, it'll be a little confusing. At worst, it will break the template manifest.

Given this is the first instance of this pattern, I'd probably be more comfortable with something a `admin/code/exceptionTemplates` folder.

assertchris: I was a little unsure of where to put this, and I think your suggestion is great. how about `messages`?

sminnee: `messages` works for me!

sminnee:
This doesn't deal with i18n as well as I'd like. Unfortunately it looks like transifex doesn't support multi-line strings.

However, I think blocking improved English help because we're unable to provide it in other languages would be cutting off our nose to spite our face, so I don' think that blocks merge.

admin/code/CMSBatchActionHandler.php
flamerohr: I'd be more comfortable with an absolute path `require FRAMEWORK_DIR . '/templates/Exceptions/InvalidBatchActionException.php'`

FRAMEWORK_DIR is defined somewhere very close to the start of the boot sequence, so should be safe to use here.

assertchris:
I've removed the fencing.

admin/code/messages/InvalidBatchAction.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+return function(array $action) {
```
sminnee: array type hint *might* break on 5.3

admin/code/messages/InvalidBatchAction.php
```diff
@@ -0,0 +1,23 @@
...
+	$action = trim(print_r($action, true));
+
+	return "
```
assertchris: I remove the previous `preg_replace` from here - print_r was just destroying it.

admin/code/messages/InvalidBatchAction.php
assertchris: Works fine with `5.3.29`...

sminnee:
Right this gets a 👍  from me. Will merge tomorrow after the tests pass.

assertchris:
*bump*

dhensby:
Isn't there a nicer way to include this without the manual `require` call? Maybe not in 3...

Should this go against 3.4? Shouldn't it be against 3 as it's a new feature?

assertchris:
@dhensby good questions. 

I've given a bit more detail about the use of `require` in some of the messages above, but the gist of it is that we have 3 options: private/protected functions, code repetition, or `require`. `private`/`protected` isn't compatible with `5.3`. I'm open to either of the other two options, but I would prefer to avoid repetition (especially given the template nature of these verbose messages).

More verbose exception/warning messages aren't a new feature - they're documentation. Many people are still stuck on `^3.2`, and this would be helpful to them now, as opposed to when `4.0` is released and they are able to upgrade. The only functional addition is throwing an exception on line `260` of `CMSBatchActionHandler`, where there was no error checking before. It's arguably a bug fix though, since before it would fail silently for invalid values, upon the assumption that `recordClass` is always defined. In contrast to this, https://github.com/silverstripe/silverstripe-framework/pull/5963 features a replacement of `user_error` message, because the assertion behaviour already exists.

TL;DR I think there's value in this additional documentation being pulled into `3.4`, and I do not believe it qualifies as a breaking change. Since `master` is now `4.0` alpha, I think the `3.4` branch is the most logical place.

assertchris:
Opened a mailing list discussion to draw a bit more discussion and attention to this change: https://groups.google.com/forum/#!topic/silverstripe-dev/14lJBw13jQI

dhensby:
Sorry, I'm suggesting this goes into 3 branch which is non-breaking changes. 3.4 should be for bug fixes. Whilst it may be argued it's documentation (and I'm happy to take a liberal view on that) we've been fairly strict on the technicalities of adding new APIs as guidance to which release branch we should look at. Adding it to 3 won't be detramental to those targeting `^3.2` as it'll make a 3.5 release...

Regarding less "manual `require` calls" - I was wondering if there was some way we could use composer and/or a global `functions.php` to define these callables? What about a specific exception instance that had this logic/look-up baked in?

assertchris:
> Sorry, I'm suggesting this goes into 3 branch which is non-breaking changes. 3.4 should be for bug fixes.

Ah, I see what you're saying. Would you like me to wait for confirmation from some else, or try to change the base branch so long?

> I was wondering if there was some way we could use composer and/or a global functions.php to define these callables?

Seems to me like a huge function of inline code would be a mess, as would a huge function of requires. 

> What about a specific exception instance that had this logic/look-up baked in?

That would be a breaking change.

camspiers:
Another idea would be to make a specific exception for this:

```php
class InvalidBatchActionException extends InvalidArgumentException {
  public function __construct($action) {
    $message = $this->createMessage($action);
    parent::__construct($message);
  }
  protected function createMessage($action) {
    /* do something with $action */
  }
}
```

sminnee:
I've had a look and transifex is able to handle multi-line strings, provided we improve our translation-import scripts in some already planned ways.

Given that, I'd like propse the followwing:

We create a new exception class, e.g. `ExtendedException`, `DescribedException` or `HelpfulExpection` that has a `getDescription()` methods. (It could also be an interface if we want).

We add a method to `i18n` that will attempt to look up a translated string, but if there is any exception failure, just return the default (presumably english) text). E.g.`i18n::_t_safe()`. The translation string finder should also pick up on calls to this method.

These exceptions are be picked up at the top level of SilverStripe, and the result of `getDescription()` printed to the screen after `getMessage()`

On default logger plugins, only `getMessage()` will get logged. If someone builds a fancy graylog plug-in, we can have that send the `getDescription()` through to graylog in such a way that it's registered as a single multi-line log event.

How does that sound?

flamerohr:
The idea of using new Exception classes which draws in a i18n translation which has a safety catch for if the translation fails for whatever reason is quite a nice one.
Even just having the exception class is helpful because it could be referenced/searched for more easily. E.g. get a search engine hit on the documentation for that exception class.

camspiers:
Sound really exciting to me. Only one minor comment:

>We create a new exception class, e.g. ExtendedException, DescribedException or HelpfulExpection that has a getDescription() methods. (It could also be an interface if we want).

A problem with using a common custom base class for these exceptions (e.g. `HelpfulExpection extends Exception`) is that any exception handling code that catches anything more specific than `Exception` will no longer work, but the exception handling could be in user code.

For example:

some existing code to update:

```php
// SilverStripe code
function some_core_silverstripe_function() {
  throw new InvalidArgumentException();
}

// product code
try {
  some_core_silverstripe_function();
} catch (InvalidArgumentException $e) {
  // do something
}
```

when replaced with:

```php
// SilverStripe code
abstract class HelpfulException extends Exception {
  // special translation handling layer
}
class MoreHelpfulException extends HelpfulException {}
function some_core_silverstripe_function() {
  throw new MoreHelpfulException();
}

// product code
try {
  some_core_silverstripe_function();
} catch (InvalidArgumentException $e) {
  // the exception will no longer be caught here
}
```

I think it is better to extend the php exception class that is most appropriate for the situation e.g. `InvalidArgumentException`.

robbieaverill:
This seems to have stagnated. Feel free to reopen if someone wants to revive it =)

PR #5954, opened 2016-09-06
Clarified <% include %> use in SS4 (fixes #5952)

### LINKED ISSUE #5952: Clarify `Includes/` and template path use in SS4
We need some clarity around template references in PHP and SS files, in particular how strong the relationship between folder and file is. At the moment there's ambiguity between forward slash (=folder separator) or back slack (=namespace separator). Template references can be a composite between namespaces and folder paths.

Example: 

```
MyFile.ss
Includes/
  Myfile.ss
MyNamespace/
  MyFile.ss
  Includes/
    MyFile.ss
```

 * `SSViewer` chooses `MyNamespace/MyFile.ss` when passed the `MyNamespace/MyFile` PHP class (higher priority)
 * `<% include MyNamespace/MyFile %>` chooses `MyNamespace/Includes/MyFile.ss`
 * `<% include MyNamespace/Includes/MyFile %>` chooses `MyNamespace/Includes/MyFile.ss`
 * `<% include MyFile %>` chooses `Includes/MyFile.ss`
 * `<% include Includes/MyNamespace/MyFile %>` doesn't work
 * `<% include Includes/MyFile %>` doesn't work

Here's the status quo:
 * SS3 has template inheritance based on class inheritance
 * SS3 has template "types": Subfolders which contain templates accessible via `SSViewer::getTemplateFileByType()`
 * The most commonly used "type" is `Includes`. Templates in this path are included by default in `SSViewer::get_templates_by_class()` (with lower priority than templates without a "type").
 * SS3 has special template placeholders for *some* "types": Templates in `Layout/` and `Content/` subfolders can be used as `$Layout` and `$Content` placeholders
 * SS4 introduces namespaces to templates, matching them to folder paths
 * SS4 supports forward slashes, back slashes and escaped back slashes for template references in `*.ss`
 * SS4 finds namespaced templates by "type" if its the innermost folder (e.g. `MyNamespace/Includes/MyFile.ss`, see `ThemeResourceLoader@findTemplate`)

The [docs](https://github.com/silverstripe/silverstripe-framework/blob/master/docs/en/02_Developer_Guides/01_Templates/01_Syntax.md#includes) mention:

> Note that in SilverStripe 3, you didn't have to specify a namespace in your include tag, as the template engine didn't use namespaces. As of SilverStripe 4, the template namespaces need to match the folder structure of your template files.

They're [ambiguous](https://github.com/silverstripe/silverstripe-framework/blob/master/docs/en/02_Developer_Guides/01_Templates/01_Syntax.md#template-file-location) about how to combine "type" and namespace folders:

> When the class has a namespace, the namespace will be interpreted as subfolder within the templates path. For, example, the class SilverStripe\Control\Controller will be rendered with the templates/SilverStripe/Control/Controller.ss template.

Sam recommended removing the `Includes/` folder (BC only) in SS4: https://github.com/silverstripe/silverstripe-framework/issues/5445

> What is the Includes folder, anyway? With all these new folders to support namespaces, the Includes folder make less sense. Although Layout and Content have special meaning for the template renderer, Includes does not. I recommend we remove this convention and any special supporting code for it.

The [theme stacking/cascading PR](https://github.com/silverstripe/silverstripe-framework/pull/5804) only [mentions](https://github.com/silverstripe/silverstripe-framework/pull/5804/files#diff-e18c85252d7df22a2d833de561771477R398) the following in the upgrading guides, without changing any other docs (see [changelog](https://github.com/silverstripe/silverstripe-framework/blob/master/docs/en/04_Changelogs/4.0.0.md#make-sure-templates-are-in-correct-locations)):

> Make sure templates are in correct locations. Templates are now much more strict about their locations. You can no longer put a template in an arbitrary folder and have it be found. Case is now also checked on case-sensitive filesystems. Either include the folder in the template name (renderWith('Field.ss') => renderWith('forms/Field.ss')), move the template into the correct directory, or both.

chillu:
Blocked until we implement suggestions in https://github.com/silverstripe/silverstripe-framework/issues/5952#issuecomment-244881303 (which also requires docs updates, but reflecting a different outcome)

### DIFF
--- a/docs/en/02_Developer_Guides/01_Templates/01_Syntax.md
+++ b/docs/en/02_Developer_Guides/01_Templates/01_Syntax.md
@@ -49,11 +49,16 @@ text-based format.
 SilverStripe templates are plain text files that have `.ss` extension and located within the `templates` directory of
 a module, theme, or your `mysite` folder.
 
-By default, templates will have the same name as the class they are used to render. So, your Page class will
+By default, templates will have the same name as the class they are used to render. So, your `Page` class will
 be rendered with the `templates/Page.ss` template.
 
-When the class has a namespace, the namespace will be interpreted as subfolder within the `templates` path. For, example, the class `SilverStripe\Control\Controller` will be rendered with the
+When the class has a namespace, the namespace will be interpreted as a subfolder within the `templates` path. 
+For example, the class `SilverStripe\Control\Controller` will be rendered with the
 `templates/SilverStripe/Control/Controller.ss` template.
+
+If you are using template "types" like `Layout` or `Includes`, these are just folders which you need
+to append to your template file location. One exception is the `<% include %>` template tag,
+where you need to leave out the `Includes/` folder.
 
 ## Variables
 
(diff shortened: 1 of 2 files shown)


sminnee:
@chillu can we close this one out for alpha2?

PR #5955, opened 2016-09-06
Fix for spacing of sitetree panel, reportAdmin, and toggle on Member details page

clarkepaul:
Includes improvements to spacing of buttons in the toolbar.


admin/client/src/styles/legacy/_forms.scss
```diff
@@ -494,8 +494,8 @@ button.loading {
 	&[class*="font-icon-"],
 	&[class^="font-icon-"],
 	&.ss-ui-button-constructive[class*="font-icon-"] {
-		padding: 5px 8px;
-		margin-bottom: $grid-y*1.5;
+		padding: 6px 8px;
```
chillu: Why are we breaking out of the grid constants here?

admin/client/src/styles/legacy/_style.scss
```diff
@@ -1614,21 +1614,28 @@ form.member-profile-form {
...
-		.cms-content-view {
-			padding: 0 $grid-gutter-width-half $grid-y*1.5;
-		}
+    > fieldset {
+      margin-top: -$grid-gutter-width-half;
+    }
+
+    .ss-gridfield-buttonrow {
+      margin-left: -20px;
+      margin-right: -20px;
```
chillu: Instead of all of this negative margin malarkey, we should simply remove `.cms-panel-padded` from the container here, right?

admin/client/src/styles/legacy/_style.scss
```diff
@@ -1613,28 +1613,37 @@ form.member-profile-form {
...
 
-	.cms-content-view {
-		padding: 0 $grid-gutter-width-half $grid-y*1.5;
-	}
+  .toolbar--content {
+    margin-left: -$grid-gutter-width-half;
+    margin-right: -$grid-gutter-width-half;
+  }
+
+  &.cms-panel-padded {
```
chillu: Could this be removed if we got rid of `.cms-panel-padded` in ReportAdmin?

admin/client/src/styles/legacy/_forms.scss
clarkepaul: I should create a variable for it actually as its a standard measurement. Same with the line-height, there is already a variable for it. Will rectify.

admin/client/src/styles/legacy/_style.scss
clarkepaul: You could but I think the alternative is worse, if you take a look at the _broken links report_ you will see it has no actions toolbar at the top meaning fields are touching the top. We can't add spacing above the :first-child as it sometimes applies it to hidden messages and fields  and it's hard to tell if they are active or not. 

This probably wouldn't be the way I'd do it if I could adjust all of the HTML but at the moment each section of the ReportAdmin & CMS has completely different markup structure (which we will need to standardise) making it hard to maintain.

The Way I'm approaching this only the first toolbar whether it is part of a gridfield will be negatively spaced, everything else will have appropriate spacing above and to the sides. This means no matter what the panel contains it should look okay.

chillu: Talked to Paul about this, see comment on https://github.com/silverstripe/silverstripe-cms/issues/1491#issuecomment-245161859

clarkepaul:
Rebased with master to fix failing tests, and added a fix to the upload layout as mentioned here https://github.com/silverstripe/silverstripe-cms/issues/1491#event-794987275

clarkepaul:
Added fix for spacing above campaign alert, area was using legacy classes so updated them to the newer panel classes.
After
image

admin/client/src/styles/legacy/_style.scss
```diff
@@ -1,24 +1,24 @@
...
- * To avoid this file getting too large and complicated, it is encouraged to
- * create new SCSS files for larger components like the CMS menu or tree
- * (see _tree.scss and _menu.scss).
- */
-
-/** ----------------------------------------------------
- * Core Styles.
- * ---------------------------------------------------- */
+//
+// This file defines most styles of the CMS: Colors, fonts, backgrounds,
```
flamerohr: How come this changed to `//`?

admin/client/src/styles/legacy/_style.scss
```diff
@@ -1,24 +1,24 @@
...
+// to get started.
+//
+// To avoid this file getting too large and complicated, it is encouraged to
+// create new SCSS files for larger components like the CMS menu or tree
+// (see _tree.scss and _menu.scss).
+
+
+// ----------------------------------------------------
+// Core Styles.
+// ---------------------------------------------------- */
```
flamerohr: still has the `*/` at the end ;)

admin/client/src/styles/legacy/_style.scss
```diff
@@ -761,14 +775,14 @@ body.cms {
  * -------------------------------------------- */
```
flamerohr: seems weird that this wasn't changed to `//` as well?

PR #5959, opened 2016-09-06
FIX Bad strpos call in HTTP::register_etag()

dhensby:


dhensby:
I'll merge up

tractorcow:
That means I need to rebase my mammoth PR again... XD

PR #5962, opened 2016-09-07
Added svg to allowed_extensions array.

thenationofalex:
Added SVG extension back into allowed extensions as it has since been removed after version 3.4.

### DIFF
--- a/filesystem/File.php
+++ b/filesystem/File.php
@@ -159,7 +159,8 @@ class File extends DataObject implements ShortcodeHandler, AssetContainer, Thumb
 		),
 		'document' => array(
 			'css', 'csv', 'doc', 'docx', 'dotm', 'dotx', 'htm', 'html', 'gpx', 'js', 'kml', 'pages', 'pdf',
-			'potm', 'potx', 'pps', 'ppt', 'pptx', 'rtf', 'txt', 'xhtml', 'xls', 'xlsx', 'xltm', 'xltx', 'xml',
+			'potm', 'potx', 'pps', 'ppt', 'pptx', 'rtf', 'svg', 'txt', 'xhtml', 'xls', 'xlsx', 'xltm', 'xltx',
+			'xml',
 		),
 		'image' => array(
 			'alpha', 'als', 'bmp', 'cel', 'gif', 'ico', 'icon', 'jpeg', 'jpg', 'pcx', 'png', 'ps', 'tif', 'tiff',
--- a/filesystem/File.php
+++ b/filesystem/File.php
@@ -140,8 +140,8 @@ class File extends DataObject implements ShortcodeHandler, AssetContainer, Thumb
 		'docx', 'dotx', 'dotm', 'flv', 'gif', 'gpx', 'gz', 'hqx', 'ico', 'jar', 'jpeg', 'jpg', 'js', 'kml',
 		'm4a', 'm4v', 'mid', 'midi', 'mkv', 'mov', 'mp3', 'mp4', 'mpa', 'mpeg', 'mpg', 'ogg', 'ogv', 'pages',
 		'pcx', 'pdf', 'png', 'pps', 'ppt', 'pptx', 'potx', 'potm', 'ra', 'ram', 'rm', 'rtf', 'sit', 'sitx',
-		'tar', 'tgz', 'tif', 'tiff', 'txt', 'wav', 'webm', 'wma', 'wmv', 'xls', 'xlsx', 'xltx', 'xltm', 'zip',
-		'zipx',
+		'svg', 'tar', 'tgz', 'tif', 'tiff', 'txt', 'wav', 'webm', 'wma', 'wmv', 'xls', 'xlsx', 'xltx', 'xltm',
+		'zip', 'zipx',
 	);
 
 	/**


dayneh88:
Great work...really....great work.

kinglozzer:
Hmm, looks like this must’ve been lost on a merge up from 3 to master 😞. Test failures seem related, though?

dhensby:
## Please don't merge

@madmatt I believe the decision had been made to remove SVG from master with the preference for devs to opt in to it given SVGs are XML and can theoretically pose XSS issues if uploaded by untrusted users?

Can we get a public declaration on this and perhaps an update to docs...

thenationofalex:
@dhensby @madmatt -  I understand the security concern here and If the decision is to reject the merge. 

Then there needs to be an easy way to allow for the use of SVGs without modifying the framework. 

Currently if you add: `File::config()->allowed_extensions[] = 'svg';`  to `mysite/_config.php` 

The following error is produced:
 
PHP Notice:  Indirect modification of overloaded property Config_ForClass::$allowed_extensions has no effect in /Applications/MAMP/htdocs/***/mysite/_config.php on`

tractorcow:
@thenationofalex the correct way to update this config is either:

via yaml

```yaml
File:
  allowed_extensions:
    - svg
```

Or via php

```php
Config::inst()->update('File', 'allowed_extensions', ['svg']);
```

I don't think we support arrayaccess for Config_ForClass properly.

Note that soon File is getting namespaced shortly; So you'll need to use the FQN once that is done.

tractorcow:
As discussed, we have explicitly opted to leave out SVG from allowed_extensions in core due to security concerns. Closing this as the correct workaround has been given.

dayneh88:
Really disappointed...really...disappointed.

jaydenseric:
@tractorcow I am on 4.0 Alpha 1 and have not had any success using that YML override.

> can theoretically pose XSS issues if uploaded by untrusted users

> untrusted users

Why would you allow untrusted users to be authoring your content in the first place, that is a greater edge case than being able to add SVG images to a website in 2016.

[silverstripe.com](https://www.silverstripe.com) has an SVG logo.

rosiematt:
@tractorcow outrageous!

nickspiel:
I agree with @jaydenseric if you have allowed a untrusted user in the first place they could do far more damage by other means than uploading a malicious XML file (I didn't even know that was a thing).

I have also have had no success with the YML override.

SilverStripe needs to start treating SVG's seriously. It is 2016 after all.

It would be amazing if you guys got SVGs sorted out for v4.

dhensby:
SVGs could be uploaded via a front end form that has a file attachment field. SVGs are XML and can contain scripts. A malicious SVG could then be linked to or embedded and execute scripts in the host domain's context, potentially leaking information about sessions or login cookies and so on. 

It should be very easy to add svg to the allowed file types of you've evaluated the security risks they pose to your site. 

We are here to provide a secure framework. For those that know this is not a risk factor, it can be enabled easily. We don't support xml uploads by default for the same reason

dhensby:
> We don't support xml uploads by default for the same reason

Hmmm looks like we do... Probably shouldn't? @tractorcow...

tractorcow:
Yeah we should remove that too,

nickspiel:
Why not limit it to CMS forms then? Could be achieved by detect in the `UploadField` if the current user is authenticated before allowing the SVG upload.

dhensby:
We aren't talking malicious cms users, here. We are talking "I have an upload field on my website which places files in a publicly accessible dir that could be linked to by some nefarious user"

We want to support SVGs, but there's a lot more to supporting SVGs than just allowing their upload out of the box... That (should be) configurable. If it's not, that's a bug

jaydenseric:
> It should be very easy to add svg to the allowed file types

https://github.com/restruct/silverstripe-svg-images/issues/1

dhensby:
@nickspiel - sure, you can implement conditional validations of you like

tractorcow:
Hi all, sorry to disappoint anyone. Certainly this is not my intention!

We leave it up to CMS developers to determine who is trusted, and to whitelist potentially risky extensions where they feel that risk is able to be mitigated, or that they trust their own users.

The philosophy we adhere to in these cases is "secure by default" so that out of the box applications do not need to follow any additional steps in order to ensure their sites are secure.

Yes you can trust your users, but if these users fall victim to identity or credential vulnerabilities, keep in mind that even malicious users can masquerade under those indentities. :D

nickspiel:
Makes sense. I guess we just need to make it easier (fix a bug maybe) with allowing the extension.

tractorcow:
> I have also have had no success with the YML override.

I'm going to make sure to get back to you with a fix for this. :)

dayneh88:
Excited...really...excited

tractorcow:
@nickspiel I've had a go at allowing SVG on my local copy. Running current dev-master of CMS and framework and the following code in my project https://gist.github.com/tractorcow/e9dedda8df1b30e3ccf1f971335c29b3 I was successfully able to upload SVG files to my page.

image

Since our Image_Backend implementations don't currently support SVG, I had to upload these as `File` types, not as `Image`.

Was perhaps the issue you struggled with related to uploading SVG into an `Image` dataobject?

tractorcow:
It looks like there is possibly a solution to that Config_ForClass assignment error.

http://stackoverflow.com/questions/13421661/getting-indirect-modification-of-overloaded-property-has-no-effect-notice

sminnee:
Let's ensure that there's a docs note about how to allow non-default page types somewhere.

tractorcow:
There are docs at https://docs.silverstripe.org/en/3.4/developer_guides/security/secure_coding/#user-uploaded-files, but they would benefit from an example config (as per my gist).

tractorcow:
I've found one bug; Folder::validate() should not check against file extensions. Currently it's running the same validation behaviour as File::validate(), and is occasionally raising unexpected "invalid file extension" errors.

tractorcow:
@thenationofalex the issue you have above sometimes occurs due to the folder validation issue. Note your error message has `svg` in the list... this sounds like Folder::validate() getting mistakenly passed through the file name filter.

We are aiming to get a rather large change in core through this week (https://github.com/silverstripe/silverstripe-framework/pull/5921) but once this is done I will address this issue with a separate fix.

Keep in mind that you may soon need to move off alpha1, as alpha2 is nearing release and includes a very large number of fixes over the rather-buggy alpha1. :)

tractorcow:
@thenationofalex I've raised a PR at https://github.com/silverstripe/silverstripe-framework/pull/5970

dhensby:
merged

tractorcow:
Yay thanks. :)

Issue #5960, opened 2016-09-07
Missing Name on CompositeField generates invalid HTML ID


tractorcow:
When generating CompositeField instances, invalid HTML id attributes can occur if setName('SomeValidName') is never invoked on it.

FieldGroup gets around this by doing some convoluted string manipulation on child names, but a standard approach is necessary.

This logic should be moved into FormTemplateHelper::generateFieldID(), so that fields without IDs have a reasonable and predictable fallback logic.

@sminnee suggested something along the lines of https://gist.github.com/sminnee/46bbb377d068d012f130f8601f2b856c.

This has an issue where duplicate titles could potentially (yet rarely) generate duplicate IDs, so another alternative is to look at a positional-based solution.

This feature is necessary so that CompositeField can be used with the form schema within react, which relies on unique ids.

robbieaverill:
>This feature is necessary so that CompositeField can be used with the form schema within react, which relies on unique ids.

Since #5964 we have composite fields in React form schema - is this still required?

I must say I've found it irritating at times that I can't get a composite field by its name

PR #5946, opened 2016-09-05
FIX Show formatting help toggle link

robbieaverill:
* Fixes #5936

PR #5964, opened 2016-09-07
Tabs and CompositeField added to React formfields

flamerohr:
A few changes had happened in the process of adding Bootstrap tabs to the React FormBuilder:

* Switched to `react-bootstrap-ss` for bootstrap4 support (for tabs only currently)
* Added tabs to edit files form (and fixes silverstripe/silverstripe-asset-admin#232)
* Bootstrap4 classes added to legacy tabs
* Added exception in FormBuilder when Component is defined but not found by Injector
* Added check in SingleSelectField for empty value before adding another one in
* Plugged a temporary workaround for CompositeField with no name (addressed in #5960)


admin/client/dist/js/MemberDatetimeOptionsetField.js
```diff
@@ -32,9 +32,9 @@
 
 		$('.memberdatetimeoptionset .toggle').entwine({
 			onclick: function onclick(e) {
-				jQuery(this).closest('.form__field-description').parent().find('.toggle-content').toggle();
+				jQuery(this).closest('.description').parent().find('.toggle-content').toggle();
```
robbieaverill: This regresses #5946 (which was added in the wrong file) and #5955 (which adds it in the right file, but doesn't include a compilation) - is it a deliberate change?

clarkepaul: This is just the dist file again so its automatic that it gets overridden. Fix still hasn't been merged properly yet.

flamerohr: Sorry, dist files are automatically built and isn't recommended to be manually modified, when #5955 gets merged in then the regression will no longer be there. It'll be difficult to keep track of a dist file change, for example if this had been minified.

robbieaverill: That's a good point, what do you guys think about minifying the dist files?

I would have committed into legacy but assumed from its folder name that it
wasn't relevant any more.

A minified file would be impossible to confuse :-)


-- 
*Robbie Averill*
LA PETITE MANOUCHE - NZ Gypsy Jazz Guitar Duo
Phone: +64 27 321 4758
contact@lapetitemanouche.com
www.lapetitemanouche.com

robbieaverill: That's a good point, what do you guys think about minifying the dist files?

I would have committed into legacy but assumed from its folder name `legacy` that it wasn't relevant any more. A minified file would be impossible to confuse :-)

Assets/File.php
```diff
@@ -472,36 +472,36 @@ public function canDelete($member = null) {
...
-				sprintf('%spx, %s', $this->getDimensions(), $this->getSize())
-			));
-			$fields->push(HTMLReadonlyField::create(
-						'ClickableURL',
-						_t('AssetTableField.URL','URL'),
-						sprintf('<a href="%s" target="_blank">%s</a>', $this->Link(), $this->Link())
-			));
-		}
-		$fields->push(HiddenField::create('ID', $this->ID));
+		$width = (int)Config::inst()->get('Image', 'asset_preview_width');
```
chillu: This shouldn't generate thumbnails on `File`, but only on `Image`

Assets/Folder.php
```diff
@@ -194,14 +194,11 @@ public function getCMSFields() {
...
 		// it's too confusing when readonly (makes sense for files only).
 
-		$fields = FieldList::create([
-			HeaderField::create('TitleHeader', $this->Title, 1),
-			LiteralField::create("ImageFull", $this->PreviewThumbnail()),
-			TextField::create("Name", $this->fieldLabel('Filename')),
-			HiddenField::create('ID', $this->ID)
-		]);
-
-		$this->extend('updateCMSFields', $fields);
```
chillu: This changes the API unnecessarily - 3.x was creating a fieldset from scratch: https://github.com/silverstripe/silverstripe-framework/blob/3.4/filesystem/Folder.php#L461

So if we're using `parent::getCMSFields()` here, assumptions made in `File->updateCMSFields()` on 3.x installs will break: It'll add fields to `Folder` which were only intended for `File`.

Assets/File.php
flamerohr: You're right, this only gets the icon of the File/Folder, I hadn't considered removing the Image Scaling calls.

admin/client/dist/js/MemberDatetimeOptionsetField.js
chillu: We're currently not minifying/mangling dist files because it impedes our ability to set breakpoints - some combination of babel/browserify/chrome means that we can't do that on minified files. This should be fixed by the planned switch to Webpack, which will also remove extraneous dist files like the one discussed here - see https://github.com/silverstripe/silverstripe-framework/pull/5918

Assets/Storage/DBFile.php
```diff
@@ -528,4 +528,19 @@ public function canViewFile() {
 				->getStore()
 				->canView($this->Filename, $this->Hash);
 	}
+
+	public function PreviewLink($action = null) {
```
chillu: Currently used as follows:

```php
$width = (int)Image::config()->get('asset_preview_width');
$previewLink = Convert::raw2att($this->ScaleMaxWidth($width)->PreviewLink());
$image = "<img src=\"{$previewLink}\" class=\"editor__thumbnail\" />";
```

We already have the following in `ImageManipulation.php`:

```php
public function PreviewThumbnail() {
	$width = (int)Config::inst()->get(get_class($this), 'asset_preview_width');
	return $this->ScaleWidth($width)  ?: $this->IconTag();
}
```

Took me a while to understand why this method was necessary: Because we want to add a class to the image tag, hence can't use `IconTag()`. And this method was added to `DBFile` rather than `ImageManipulation` because its technically not a manipulation.

All good, apart from some clarification in the commit message which could've simplified the peer review.

Oh, and PHPDoc for the method please.

Forms/CompositeField.php
```diff
@@ -98,6 +102,34 @@ public function getChildren() {
 	}
 
 	/**
+	 * Returns the name (ID) for the element.
+	 * If the CompositeField doesn't have a name, but we still want
+	 * the ID / name to be set. This code, generates the ID from the nested children
+	 *
+	 * TODO this is temporary, and should be removed when FormTemplateHelper is updated to handle ID
```
chillu: Please use `@todo` in PHPDoc blocks

chillu: Also, please use PHPDoc properly (`@return`)

Forms/CompositeField.php
```diff
@@ -98,6 +102,34 @@ public function getChildren() {
 	}
 
 	/**
+	 * Returns the name (ID) for the element.
+	 * If the CompositeField doesn't have a name, but we still want
+	 * the ID / name to be set. This code, generates the ID from the nested children
```
chillu: Spelling

Forms/Tab.php
```diff
@@ -115,7 +122,10 @@ public function getTabSet() {
 	}
 
 	public function extraClass() {
-		return implode(' ', (array)$this->extraClasses);
+		$classes = (array)$this->extraClasses;
+		$classes[] = 'tab-pane';
```
chillu: This adds bootstrap classes to every tab use (even outside of the CMS). It's a shame this can't be implemented through a new `Includes/Tab.ss` template, which we could override with a CMS-specific template - because a tab is rendered through `<div $AttributesHTML>`. `FormField->AttributesHTML()` doesn't allow customisation of the attributes, so we can't merge PHP and template specified classes. And even if we implement that in the `FormField` API, the template parser doesn't support non-string arguments like `<div $AttributesHTML(["class" => "tab-pane"])>` (see `SSTemplateParser.inc.php` `Argument_Lookup()`). That'll be a pain to implement. 

@flamerohr Can you please document the reason this class is on `Tab`?
@tractorcow Do you think this is a worthwhile tradeoff for now?

admin/client/src/components/CompositeField/README.md
```diff
@@ -0,0 +1,19 @@
...
+<CompositeField id="Container">
+    <TextField name="FirstName" />
+    <TextField name="LastName" />
+</Tabs>
+```
+
+## Properties
+
+ * `name` (string) (required): A key to match with the Tabset `activeKey` property to show the content.
+ * `title` (string): The label to display for the Tab, can be set `null` to hide the Tab.
```
chillu: "tab" references don't make sense out of context here (you only mention "composite field" in the intro)

admin/client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -248,13 +261,46 @@ export class FormBuilderComponent extends SilverStripeComponent {
...
 
+  /**
+   * Finds the field with matching id from the schema or state, this is mainly for dealing with
+   * schema's deep nesting of fields.
+   *
+   * @param fields
+   * @param id
+   * @returns {*}
+   */
+  findField(fields, id) {
```
chillu: Unit tests for this please

admin/client/src/components/Tabs/README.md
```diff
@@ -1 +1,37 @@
 # Tabs
+
+For separating content into tabs without the need for separate pages.
+
+This extends from `react-bootstrap` with similar expected behaviours, only difference is that when
+there is only 1 tab (or none) in the Tabset only the content will show without the clickable Tab.
```
chillu: Grammar police: Numbers from one through nine should be spelled out (http://www.grammarbook.com/numbers/numbers.asp). Definitely `1` = `one`. Also, inconsistent capitalisation of `Tab`

admin/client/src/components/Tabs/TabItem.js
```diff
@@ -0,0 +1,33 @@
...
+import SilverStripeComponent from 'lib/SilverStripeComponent';
+import { Tab } from 'react-bootstrap-ss';
+
+class TabItem extends SilverStripeComponent {
+  getTabProps() {
+    return {
+      eventKey: this.props.name,
+      title: this.props.title,
+      disabled: this.props.disabled,
+      className: [this.props.extraClass, this.props.className].join(' '),
```
chillu: We should merge this with other props passed into the component as a standard practice, there's heaps more you can specify via react-bootstrap: https://react-bootstrap.github.io/components.html#tabs. Please implement this on all react-bootstrap wrappers

admin/client/src/components/Tabs/README.md
```diff
@@ -1 +1,37 @@
 # Tabs
+
```
chillu: Awesome, thank you for writing great docs :)

templates/SilverStripe/Forms/TabSet.ss
```diff
@@ -1,10 +1,13 @@
 <div $getAttributesHTML("class") class="ss-tabset $extraClass">
-	<ul>
+	<ul class="nav nav-tabs">
```
chillu: Please implement this in `framework/admin/themes/` instead

admin/client/src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,113 @@
...
+   * Render an individual link for the tabset
+   *
+   * @param child
+   * @returns {ReactElement}
+   */
+  renderTab(child) {
+    if (child.props.title === null) {
+      return null;
+    }
+    return <NavItem eventKey={child.props.name}
```
chillu: Needs parentheses

admin/client/src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import SilverStripeComponent from 'lib/SilverStripeComponent';
+import { Tab, Nav, NavItem } from 'react-bootstrap-ss';
+
+class Tabs extends SilverStripeComponent {
+  /**
+   * Determines a default tab to be opened and validates the given default tab.
```
chillu: Mention that this does not work recursively at the moment

admin/client/src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,113 @@
...
+        .find((child) => child.props.name === this.props.defaultActive);
+
+      if (activeChild) {
+        active = activeChild.props.name
+      }
+    }
+
+    if (typeof active !== 'string') {
+      React.Children.forEach(this.props.children, (child) => {
+        if (active === null) {
```
chillu: This was `active === undefined` for me (merge error?). `active` is always defined. I've changed to `typeof active === 'string'`

Forms/Tab.php
flamerohr: Yes, this is a real shame... :( The alternative way was to use entwine and a lot of "onmatch" and "onunmatch" checking, which I thought was the worse of two evils.
I went with this implementation and got away with only one "onmatch" checking required:
https://github.com/silverstripe/silverstripe-framework/pull/5964/files#diff-8152de84382846fefee15697ffa38b6dR79

admin/client/src/components/Tabs/Tabs.js
flamerohr: I do not understand where recursion comes from?

chillu: Tabs can have children, which can be (nested) Tabs as well, right? This only iterates through the top level

flamerohr: oh, but the child Tab would also have its own `getDefaultActiveKey()` and `activeKey` property.

I think it would lose its self-contained nature if we're extending to that level of control for `defaultActiveKey`/`activeKey`.

clarkepaul:
Seems like there is a missing class (.nav-tabs) on the File/Edit tabs?

flamerohr:
I overlooked the classes defined when adding the new Tab properties

This should fix the `.nav-tabs` class and remove the `.undefined` classes added

clarkepaul:
Looking good @flamerohr, I changed around a bit of legacy styles so they used bootstraps variables more. I didn't do a total clean out of all the legacy styles as that would be a mission. If you can take a quick look at my css. All good by me.

clarkepaul:
Also tested in IE10 btw

chillu:
> API Tweaked API for Folder and Image back to creating fieldset from scratch

Note on that commit message: It's not an API change if you revert to the existing approach within a single PR - devs upgrading don't need to know about the change of direction, it'll confuse them more than it helps. So changes like this should be squashed into the original work. It's good to have them as separate commits for peer review, although I'm not too fussed with you rebasing as you go either.

Assets/ImageManipulation.php
```diff
@@ -156,14 +156,20 @@
...
 	 */
 	private static $asset_preview_width = 400;
 
 	/**
+	 * The height of an image preview in the Asset section
+	 *
+	 * @config
+	 * @var int
+	 */
+	private static $asset_preview_height = 336;
```
chillu: I'll resist the urge to start a retina thumbnail discussion ... :D

Assets/File.php
```diff
@@ -472,36 +473,36 @@ public function canDelete($member = null) {
...
-				sprintf('%spx, %s', $this->getDimensions(), $this->getSize())
-			));
-			$fields->push(HTMLReadonlyField::create(
-						'ClickableURL',
-						_t('AssetTableField.URL','URL'),
-						sprintf('<a href="%s" target="_blank">%s</a>', $this->Link(), $this->Link())
-			));
-		}
-		$fields->push(HiddenField::create('ID', $this->ID));
+		$width = (int)Image::config()->get('asset_preview_width');
```
tractorcow: I'm not sure we should be doing image specific manipulation in File... that should be overridden in subclass, right?

flamerohr: I guess my line of thinking was if `parent::getCMSFields()` were used, then updateCMSFields extension points would run before the image fields are added or removed

Assets/Storage/DBFile.php
```diff
@@ -528,4 +529,27 @@ public function canViewFile() {
...
+	 * have been manipulated e.g. by {@link ImageManipulation}
+	 * Use the 'updatePreviewLink' extension point to customise the link.
+	 *
+	 * @param null $action
+	 * @return bool|string
+	 */
+	public function PreviewLink($action = null) {
+		// Since AbsoluteURL can whitelist protected assets,
+		// do permission check first
+		if (!$this->failover->canView()) {
```
tractorcow: Seems to be a missing check here to see if failover is a file, or even if it's set?

PR #5967, opened 2016-09-08
Insert media into TinyMCE modal, styling fixes and improvements

clarkepaul:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1591

Insert media modal tidy up, required because of the new bootstrap forms styles.
Quite a few general css linting fixes.

There was a minor inconsistency as per below on IE10 but not a show stopper seeing as this modal has a short life span, everything functional.
pasted_image_8_09_16__5_10_pm

Includes toggle fix for https://github.com/silverstripe/silverstripe-framework/pull/5946


flamerohr:
Selected file still shows for "Insert for URL" and ignores the inserted URL if not removed...
not the best UX, but would that be a different story?

screen shot 2016-09-09 at 11 40 26 am

flamerohr:
Issue with when uploading a new file in the modal:
screen shot 2016-09-09 at 1 31 06 pm 2

clarkepaul:
This modal is only a temporary solution and should be reworked before 4 is released. Insert from URL will be a different interface from insert from Files so I would say we could create a new issue with that in mind and then it might not need to fixed.

Upload bug, I missed that one.

clarkepaul:
Okay I think I've resolved the issues, sorry it took longer than I thought. It's still really fragile, with all this css and markup, will be good when we can get rid of it all. 

Uploading and errored uploads styles should be good now.

Ready for PR again

clarkepaul:
Also resolved the issue of not being able to insert from url if previously uploaded an image. Should be able to upload multiple types in one go.

dhensby:
Behat test failed - I've restarted it but the failure did look related.

clarkepaul:
Took a look at behat test but I'm not that familiar with them, things seem to be in the right places. @flamerohr is this something you wouldn't mind checking over for me?

flamerohr:
We've found that behat tests rely on the class `ss-ui-button` to find buttons... which is being deprecated.
So we'll need to work out another logic with how to find buttons, either by an enforced `role="button"` attribute for `<a href="">` or some other means.

My suggestion would be a selector like as follows:
`button, input[type="button"], a[role="button"]`

clarkepaul:
For now I've added back the class .ss-ui-button so that tests will pass. In the long run we will need to remove all of these classes from the system with a solution like @flamerohr has mentioned above. This class is tied to JQueryUI and because of it other classes are automatically added (.ui-button ui-widget ui-state-default ui-corner-all ui-button-text-only) which applies a lot of unwanted styles.

Hoping tests will pass now... ?

clarkepaul:
Waiting tests to pass

flamerohr:
Changes look good, peer review pass for me

A shame about `ss-ui-button`, we'll need to dig into this more in the close future

dhensby:
Maybe we just create a class specifically for behat tests? Seems nasty, but we shouldn't really rely on classes that provide style for those tests...

PR #5969, opened 2016-09-08
Fix case of required paths

tractorcow:


dhensby:
If we namespace can't we remove these requires completely and use the PSR autoloader?

tractorcow:
Yep I think so. Eventually :D

PR #5971, opened 2016-09-08
BUG fix core include

tractorcow:
I'm going to keep pulling until these tests stop breaking. :)


tractorcow:
Merge https://github.com/silverstripe/silverstripe-cms/pull/1604 to fix

PR #5977, opened 2016-09-09
FIX: Remove unnecessary manual includes

sminnee:
Core.php can rely on composer’s autoloader now that classes are PSR-4
compliant.

If you stuck to the core modules, you could even remove ClassLoader’s
autoloader, but this would break any module that hasn’t been updated
to support PSR-4, so I’ve left it in.

In the future, it would be good to apply SilverStripe’s auto-loader
only to those modules that aren’t coded to use PSR-4, as it would
make class loading more predictable.


chillu:
Ooooh 😎 !

dhensby:
Welcome to the future

PR #5982, opened 2016-09-11
Update 01_Error_Handling.md

pandauxstudio:
Indentation error causes yaml parsing error.


dhensby:
In not sure why code climate is complaining. Are you recessed against the latest commit?

chillu:
I think the CodeClimate excludes aren't set up properly: https://github.com/silverstripe/silverstripe-framework/pull/5985

PR #5985, opened 2016-09-12
Improved CodeClimate excludes

chillu:
Exclude lang folders, make other includes less verbose
by using globs, see https://docs.codeclimate.com/docs/excluding-files-and-folders

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -14,9 +14,6 @@ ratings:
   - client/src/**
   - admin/client/src/**
 exclude_paths:
-- "client/dist/**"
-- "admin/client/dist/**"
-- "admin/thirdparty/**"
-- "templates/"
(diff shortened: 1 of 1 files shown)


chillu:
This seems to have done *some* job - now we have less files than expected :D https://codeclimate.com/github/silverstripe/silverstripe-framework/code. I can't make any sense of the pattern of files being included there, so contacted CodeClimate support for help. 

I've also enabled CodeClimate on our fork, but that timed out because it chose the wrong default branch: https://codeclimate.com/github/open-sausages/silverstripe-framework/builds/1 - contacted CodeClimate about that as well

flamerohr:
so with that pull request change, it'd mean duplicate checking is no longer applied to the jest unit tests, is that a desired effect?

PR #5986, opened 2016-09-12
Simpler test run

sminnee:
The goal of this branch is to avoid having to set up a SilverStripe project structure in order to run framework tests. Part of this is to simplify test execution for contributors, but a related motivation is to make the SilverStripe codebase more flexible to structure changes.

In principle, these changes should also pave the way for letting you work on a module by checking out that module and running `composer install` within it.

For actually hosting the site in such a situation, I'm working with @assertchris on https://github.com/assertchris/silverstripe-serve. The idea will be that `silverstripe/serve` can be set up a composer dev-depdendency, and `bin/serve` can be executed to get a working dev server of your module.

This PR isn't ready for merge; it's here to show progress (and get a travis run going)

### Merge blockers

 * [x] Fix broken PHPUnit tests
 * [x] Get behat tests working


Forms/GridField/GridField.php
```diff
@@ -307,12 +307,12 @@ public function getState($getData = true) {
 	 */
 	public function FieldHolder($properties = array()) {
 		Requirements::css(THIRDPARTY_DIR . '/jquery-ui-themes/smoothness/jquery-ui.css');
-		Requirements::css(FRAMEWORK_DIR . '/client/dist/styles/GridField.css');
+		Requirements::css(trim(FRAMEWORK_DIR . '/client/dist/styles/GridField.css', '/'));
```
dhensby: What about introducing a convention of the constants having a trailing slash?

PS: `ltrim`?

tests/bootstrap.php
```diff
@@ -40,6 +38,48 @@
 	$_REQUEST = array_merge($_REQUEST, $_GET);
 }
 
+// DATABASE BOOTSTRAP
+
+if (!defined('SS_ENVIRONMENT_TYPE')) {
+	define('SS_ENVIRONMENT_TYPE', 'dev');
```
dhensby: Not `test` ;)

dhensby:
You'll need to create a `_ss_environment.php` that contains the file to url mapping. This should probably be done in the travis `before_script` part.

It needs to be in the `_ss_environment.php` rather than in the test bootstrapper because the file url mapping is only taken into account when `Constants.php` is included and that'a autoloaded by composer (I assume as soon as the composer autoloader is required?!)

Forms/GridField/GridField.php
sminnee: That'll lead to annoying errors everywhere. I'd prefer to have slightly ugly code until we have a proper solution.

A proper solution, IMO, is going to be to have a way of saying "get me this asset of this module" for either server-side consumption. Important for moving modules to vendor and moving code out of the webroot.

tractorcow:
> You'll need to create a _ss_environment.php that contains the file to url mapping. This should probably be done in the travis before_script part.

This shouldn't be as mandatory as it is; If it's not defined, we should show a warning (once only!) and set it to a reasonable default. e.g. localhost => BASE_DIR

Forms/GridField/GridField.php
tractorcow: Could this logic be stuck in Controller::join_links()?

tractorcow: Also, once we move to code-outside-of-webroot, this is going to need to be changed anyway. The fact that it's currently ugly is temporary. :)

tractorcow:
It looks really good so far @sminnee !

tractorcow:
@sminnee , @dhensby and I were talking earlier about how to simply bootstrapping the environment. Most of what's in installer is code that should be optional, in any case. The mysite/_config.php defines $project = 'mysite', which should probably be moved to a config, and the `include conf/ConfigureFromEnv.php` could be moved into core.

I see that you've eliminated the need for the mysite/_config.php for unit testing, but could we go a step further and make it entirely optional even for running in a local web browser?

sminnee:
> I see that you've eliminated the need for the mysite/_config.php for unit testing, but could we go a step further and make it entirely optional even for running in a local web browser?

I've already done this locally, using the `silverstripe/serve` module that @assertchris has been building. Will share once I've got PHPUnit tests done.

sminnee:
OK. I'm pretty happy with this now. I've rebased and squashed what I think needs to be squashed—I think the renaming commits are best merged as-is.

CMS tests will fail—it's going to be really hard to get these to pass before merging so if possible I'd prefer that we mark them as allowed failures until after I've got the CMS build working with this new mechanism.

In any case, it's ready for more in-depth feedback on what merge blockers there may be.

Forms/InlineFormAction.php
```diff
@@ -37,7 +37,7 @@ public function performReadonlyTransformation() {
 	public function Field($properties = array()) {
 		if($this->includeDefaultJS) {
 			Requirements::javascriptTemplate(
-				FRAMEWORK_DIR . '/client/dist/js/InlineFormAction.js',
+				trim(FRAMEWORK_DIR . '/client/dist/js/InlineFormAction.js', '/'),
```
dhensby: You've used `ltrim` everywhere but here

View/Requirements_Backend.php
```diff
@@ -944,7 +944,7 @@ public function add_i18n_javascript($langDir, $return = false, $langOnly = false
 			// add_i18n_javascript() was called indicates that the methods in
 			// here are needed.
 			if (!$langOnly) {
-				$files[] = FRAMEWORK_DIR . '/client/dist/js/i18n.js';
+				$files[] = trim(FRAMEWORK_DIR . '/client/dist/js/i18n.js', '/');
```
dhensby: `ltrim`

View/Requirements_Backend.php
```diff
@@ -967,7 +967,7 @@ public function add_i18n_javascript($langDir, $return = false, $langOnly = false
 		} else {
 			// Stub i18n implementation for when i18n is disabled.
 			if (!$langOnly) {
-				$files[] = FRAMEWORK_DIR . '/client/dist/js/i18nx.js';
+				$files[] = trim(FRAMEWORK_DIR . '/client/dist/js/i18nx.js', '/');
```
dhensby: `ltrim`

tests/control/HTTPTest.php
```diff
@@ -327,8 +327,9 @@ public function testURISchemes() {
 
 	public function testFilename2url() {
 		$this->withBaseURL('http://www.silverstripe.org/', function($test) {
+			$frameworkTests = trim(FRAMEWORK_DIR . '/tests', '/');
```
dhensby: `ltrim`

composer.json
```diff
@@ -24,7 +24,11 @@
     "embed/embed": "^2.6"
   },
   "require-dev": {
-    "phpunit/PHPUnit": "~4.8"
+    "phpunit/PHPUnit": "~4.8",
+    "silverstripe/behat-extension": "dev-master",
+    "silverstripe/serve": "dev-improve-portability",
```
dhensby: do we want to clean up this version constraint before merge? As well as not using your fork and/or registering it with packagist?

dhensby:
I'd be interested to know how and if this has any impact on `cli-script.php` / `sake`? Do those commands still work/run when run directly from framework? Should they?

composer.json
sminnee: I'm waiting for a PR to be merged, and the merge of that PR was contingent on this and the work in CMS being done. Perhaps we merge like this and I'll clean up once a new release is out?

sminnee:
You can still use sake but not without a project folder. If we make sake a package bin, so you called it as `vendor/bin/sake`, we could fix that.

The changes in this rely on a choosing a composer autoloader before you run any silverstripe stuff, which we do by initiating all calls from a vendor bin: `vendor/bin/behat`, `vendor/bin/serve`, and `vendor/bin/phpunit`. This is why it was important to replace the travis apache with `silverstripe/serve`.

sminnee:
I just want to draw reviewers attention to 417784ccb5e2e1739f97f7d0536a4c8d734e8362. I've had to implement `silverstripe/framework:/admin/themes/cms-forms` as a legal theme name. Allowing for module relative sub-paths was pretty straightforward, but I've also got some hard-code path lookup for CMS and framework based on the constants.

This is pretty ugly, and the real fix is to build a more general module path look-up system, but it's beyond the scope of this initial PR.

sminnee:
OK @dhensby I've moved silverstripe/serve to mainline dev-master now.

sminnee:
OK I've rebased this against the webpack work just pushed to master. Lots of the code I fixed was deleted, which I guess is a good sign! ;-)

tests/control/ControllerTest.php
```diff
@@ -29,6 +29,7 @@ class ControllerTest extends FunctionalTest {
 
 	public function setUp() {
 		parent::setUp();
+        Config::inst()->update('SilverStripe\\Control\\Director', 'alternate_base_url', '/');
```
dhensby: I'm loathed to block merging due to whitespace issues, but this is spaced and the rest of the file is tabbed

tests/security/SecurityTest.php
```diff
@@ -63,6 +63,8 @@ public function setUp() {
 		Member::config()->unique_identifier_field = 'Email';
 
 		parent::setUp();
+
+        Config::inst()->update('SilverStripe\\Control\\Director', 'alternate_base_url', '/');
```
dhensby: Whitespace again

composer.json
```diff
@@ -24,7 +24,11 @@
     "embed/embed": "^2.6"
   },
   "require-dev": {
-    "phpunit/PHPUnit": "~4.8"
+    "phpunit/PHPUnit": "~4.8",
+    "silverstripe/behat-extension": "dev-master",
+    "silverstripe/serve": "dev-master",
+    "silverstripe/testsession": "dev-master",
```
dhensby: It'd be good to get these locked down to actual reeases

.travis.yml
```diff
@@ -30,12 +28,12 @@ matrix:
     - php: 5.6
       env: DB=MYSQL BEHAT_TEST=1
     - php: 5.6
+      env: NPM_TEST=1
+  allow_failures:
```
dhensby: Now that we have more than 1 allowed failure test can we use travis' fast finish feature (https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing)

sminnee:
They're all pretty straightforward fixes except for locking serve. I'm a bit hesitant to lock down serve just yet in case it needs minor tweaks to get cms builds going.

dhensby:
> They're all pretty straightforward fixes except for locking serve. I'm a bit hesitant to lock down serve just yet in case it needs minor tweaks to get cms builds going.

OK - that makes sense

.travis.yml
sminnee: It's a temporary situation only (until I fix cms to also run in this way), and I don't actually think the allowed failures are running at the moment.

tests/security/SecurityTest.php
sminnee: Fixed

composer.json
sminnee: I've locked down behat and testsession but I'm not sufficiently confident that I'm done tinkering with serve and would rather lock that down after I've got the cms build working.

tests/control/ControllerTest.php
sminnee: Fixed.

sminnee:
OK I fixed most of the issues that you've identified and squashed them in. The others I'm deliberately avoiding for now. See above for details.

Would be good to get this merged!

dhensby:
@gregsmirnov thanks for pointing this out, can you please raise a new issue for this regression?

gregsmirnov:
done in #6027

dhensby:
thanks

tractorcow:
There's still a missing dependency on uploads for travis artficats on behat failure. Can we fix those?

PR #5988, opened 2016-09-12
API Allow has_many fixtures to be declared with array format as well …

dhensby:
…as many_many (#5944)

BUG Fix issue with parsing of extrafields in fixtures
BUG Fix issue in duplicate relation name, and ensure FixtureBlueprint fails on these


dhensby:
<3

PR #5993, opened 2016-09-12
Enforce JS var declaration with value

chillu:
This caught us out recently where code did a strict type check for `myVar === undefined`.
The var was defined as `let myVar;`, without a value - so the check returned false (it's `null`).
To avoid this situation, we've decided to enforce declarations with values.
Note that preference should be given to single, immutable assignments via const where possible.

See http://eslint.org/docs/rules/init-declarations


admin/client/src/components/FormBuilder/tests/FormBuilder-test.js
```diff
@@ -55,9 +55,9 @@ describe('FormBuilderComponent', () => {
...
   describe('getFieldValues()', () => {
-    let formBuilder;
-    let fieldValues;
-    let props;
+    let formBuilder = null;
+    let fieldValues = null;
+    let props = null;
 
     it('should retrieve field values based on schema', () => {
       props = {
```
sminnee: You could have put `const props` here.

PR #5996, opened 2016-09-13
API support dblib

tractorcow:
Follow up from https://github.com/silverstripe/silverstripe-mssql/issues/32

Will need to be merged up into master.


tractorcow:
I think I need an abstract PDO DSN builder. It's getting a bit ridiculous hard-coding all these various configurations into one class.

sminnee:
Could we just accept a string DSN arg?

tractorcow:
We could but it would be a little less user friendly to force anyone on PDO to generate their own DSN. It's an option though. ;)

PR #5997, opened 2016-09-13
corrupt YML in italian translation

dimdimych:
when one switches to Italian language a fatal error is thrown


ARNHOE:
This needs to be commited into 3.4 aswell, same problem there.

dimdimych:
this problem exists in all 3.* bramches

PR #6002, opened 2016-09-14
Fixed react-bootstrap tabs warnings

chillu:
We can't use our own "id" values here, since react-bootstrap requires internal consistency
between tab containers and tab panes through its own auto-generated identifiers


flamerohr:
@ingo This looks good, and just need rebase for merging in :D

PR #6005, opened 2016-09-14
API Enable default value to be specified for dbstring types at the db level

### LINKED ISSUE #1409: [2011-07-19] default record for new field
created by: nicolaas
created at: 2011-07-19
original ticket: http://open.silverstripe.org/ticket/6683

---

when I create a new field in a dataobject then I expect the default for the existing records to be set to the default (if a default is defined of course).

```
//just added these two lines to my Data Object:
public static $db = array("MyNewField" => "Varchar");
public static $defaults = array("MyNewField" => "YYY");
```

The same thing should perhaps also apply later, when I add a default like this

```
//I just added this line
public static $defaults = array("MyNewField" => "YYY");
```

that is, if the existing records of the dataobject are NULL or a zero length string then perhaps they should bet set to YYY?

tractorcow:
Fixes #1409

It was easy so I did it. :)

Also added getters and setters to the previously protected member.

### DIFF
--- a/docs/en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
+++ b/docs/en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
@@ -73,7 +73,13 @@ When adding a new `$db` field to a DataObject you can specify a default value
 to be applied to all existing records when the column is added in the database
 for the first time. This will also be applied to any newly created objects
 going forward. You do this be passing an argument for the default value in your 
-`$db` items. For example:
+`$db` items. 
+
+For integer values, the default is the first parameter in the field specification.
+For string values, you will need to declare this default using the options array.
+For enum values, it's the second parameter.
+
+For example:
 
 	:::php
 	<?php
@@ -82,7 +88,8 @@ going forward. You do this be passing an argument for the default value in your
 		
 		private static $db = array(
 			'Wheels' => 'Int(4)',
-			'Condition' => 'Enum(array("New","Fair","Junk"), "New")'
+			'Condition' => 'Enum(array("New","Fair","Junk"), "New")',
+			'Make' => 'Varchar(["default" => "Honda"]),
 		);
 	}
 
--- a/tests/model/DBStringTest.php
+++ b/tests/model/DBStringTest.php
@@ -1,7 +1,7 @@
 <?php
 
 
-
+use SilverStripe\Core\Object;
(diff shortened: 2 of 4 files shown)


docs/en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -73,7 +73,13 @@ When adding a new `$db` field to a DataObject you can specify a default value
 to be applied to all existing records when the column is added in the database
 for the first time. This will also be applied to any newly created objects
 going forward. You do this be passing an argument for the default value in your 
-`$db` items. For example:
+`$db` items. 
+
+For integer values, the default is the first parameter in the field specification.
```
dhensby: I really feel we should remove this feature and make it all conform to using `$defaults`

tractorcow: You mean array option format for int fields? I suppose we could.

dhensby: Yep

tractorcow: Do I need to do that in this PR or can it be another one?

tests/model/DBStringTest.php
```diff
@@ -1,7 +1,8 @@
 <?php
 
 
-
+use SilverStripe\Core\Object;
+use SilverStripe\Dev\Debug;
```
flamerohr: maybe remove `SilverStripe\Dev\Debug` since it was only for Dev

tests/model/DBStringTest.php
```diff
@@ -26,6 +27,18 @@ public function testForTemplate() {
 		);
 	}
 
+	public function testDefault() {
+		/** @var DBString $dbField */
+		$dbField = Object::create_from_string(
+			"StringFieldTest_MyStringField(['default' => 'Here is my default text')",
```
flamerohr: Doesn't seem to break the test, but missing a closing bracket for the array `"StringFieldTest_MyStringField(['default' => 'Here is my default text'])"`

tractorcow:
I've fixed that @flamerohr

chillu:
Last build was passing before @tractorcow made some trivial changes, so I'll merge this. Don't think we need to solve array options for `DBInt` in this PR.

Issue #5976, opened 2016-09-09
Document differnece between get()->byID and DataObject::get_by_id


lekoala:
Something I noticed recently is that doing Member::get()->byID($id) is not cached, whereas DataObject::get_by_id('Member',$id) is.

Is that the desired behaviour ?

Because if it's the case, I think it would make sense to say so in the documentation (if you want to retrieve a cached record, use DataObject::get_by_id..) and use the cache where it make sense.
Actually, I noticed this with Member::currentUser() that is always making a new request, when 99% of the time it's going to return the same record anyway.

kinglozzer:
Something similar (especially the part about `Member::currentUser()`) was raised here: https://github.com/silverstripe/silverstripe-framework/issues/4023. I had a quick play around with implementing the suggestion in that issue, but never tidied it up/wrote test coverage for it: https://github.com/kinglozzer/sapphire/commit/13c03af38d8a31c249f7bad1123fb5dd3aab3f73

dhensby:
I don't think this is "intended" - ideally we'd have a caching system in place to support `ObjectName::get()->byID($id)` but we don't and so the legacy `DataObject::get_by_id('ObjectName', $id)` has stuck around because we haven't got another way to do caching. I don't think we want to encourage the legacy syntax.

lekoala:
Ok, so the goal is to cache every request byID or also any "get_one" query (like how it's done in DataObject::get_one, but it seems to me that it adds a lot of complexity on how to cache things properly).

If it's only byID, maybe it's just a matter of adding a parameter "cache" to the byID method (and set it to true by default?), and having just a static array by class then by id. If the record exists and has not been destroyed, return it from the cache. I'd be happy to write the patch if you agree with this plan.

dhensby:
> If it's only byID, maybe it's just a matter of adding a parameter "cache" to the byID method (and set it to true by default?)

I think we need a more complete caching solution than this - persisting across requests would be ideal but we should aim for something more generic and elegant than a cache flag on one method.

lekoala:
It would be wonderful to have the cache persisting across requests (provided that the caching backend is faster than a database call :-) ...).

dhensby:
> It would be wonderful to have the cache persisting across requests (provided that the caching backend is faster than a database call :-) ...).

It almost certainly would be - I can't think of much that would be slower :P

lekoala:
Well, file cache is not much of a gain compared to a database hit.

[...]

PR #6007, opened 2016-09-14
cache currentUser query

lekoala:
Various modules can call a lot of time Member::currentUser(). We can avoid querying the database multiple times. Cache is implemented as a static array inside the method and store the data byID, in case the currentUserID changes within the same request (not very likely, but..)

### DIFF
--- a/security/Member.php
+++ b/security/Member.php
@@ -834,10 +834,16 @@ class Member extends DataObject implements TemplateGlobalProvider {
 	 * @return Member|null
 	 */
 	public static function currentUser() {
+		static $cache = array();
+		
 		$id = Member::currentUserID();
 
 		if($id) {
-			return Member::get()->byId($id);
+			if(!empty($cache[$id])) {
+				return $cache[$id];
+			}
+			$cache[$id] = Member::get()->byId($id);
+			return $cache[$id];
 		}
 	}
 
--- a/security/Member.php
+++ b/security/Member.php
@@ -837,7 +837,7 @@ class Member extends DataObject implements TemplateGlobalProvider {
 		$id = Member::currentUserID();
 
 		if($id) {
-			DataObject::get_by_id('Member', $id);
+			return DataObject::get_by_id('Member', $id);
 		}
 		return null;
 	}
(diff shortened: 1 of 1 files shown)


Security/Member.php
```diff
@@ -812,10 +812,16 @@ public function getValidator() {
 	 * @return Member
 	 */
 	public static function currentUser() {
+		static $cache = array();
+		
 		$id = Member::currentUserID();
 
 		if($id) {
-			return Member::get()->byID($id);
```
sminnee: replace this with `DataObject::get_by_id()`. It will provide a cache for you.

Then you don't need the caching code.

dhensby:
@sminnee see #5976 

This is a bit of a precursor to removing those legacy methods... Though I suppose needles whilst we still kind of support that syntax :/

lekoala:
At the same time, maybe it's needed to address issues like when a Member is destroyed, like get_one does...
I think creating a singleton and calling cacheKeyComponent just to store a cache by id does not make a lot of sense (and maybe it's too limited anyway, since a proper caching backend would be nicer).

You choose :-) I made the change as well but it's not an easy decision.

dhensby:
@lekoala I think this needs rebasing as tests are failing presumably because `Member` is now namespaced in master.

@sminnee it seems only you can mark this as no longer blocked :/

dhensby:
@lekoala you might want to target 3.4 instead of master for this?

lekoala:
@dhensby yes sorry I did not test on my local computer I did the edit online. Maybe I'll close this one and make a new pull request on 3.4 instead it might be easier than changing everything.
Did you choose to use DataObject::get_by_id('Member',$id) instead of local cache array?

dhensby:
OK - I've rebased this for you and set the target branch to 3.4

dhensby:
OK [according to the docs](https://help.github.com/articles/about-pull-request-reviews/) only @sminnee can approve his review:

> If required reviews are enabled and a collaborator with write, admin, or owner access to the repository submits a review requesting changes, the pull request cannot be merged until the same collaborator submits another review approving the changes in the pull request.

This is a bit silly - I'll merge it anyway as all the CI will fail because I've switched the branch and the CI doesn't switch when that happens.

I pushed to my own fork to get [travis to run](https://travis-ci.org/dhensby/silverstripe-framework/builds/160178078) and it's green so I'm merging.

lekoala:
:-) Great ! Thank you

PR #6010, opened 2016-09-14
Fixing button destroy bug

3Dgoo:
When you want to add a button to the CMS but don't want LeftAndMain to apply jQuery UI button to it we add the `data-button="true"` attribute to our button. 

The `onadd` function checks that this attribute does not exist before calling `this.button()`.

The `onremove` incorrectly checks that this attribute does exist before calling `this.button('destroy')`. This should be checking that this attribute does not exist, just like the `onadd` function.

What this causes is when you have a button with the `data-button` attribute a button is not created, but when you leave the page `destory` gets called on an item which doesn't exist. We end up with the following error:

> Uncaught Error: cannot call methods on button prior to initialization; attempted to call method 'destroy'

The other issue with this logic is buttons are never getting destroyed when `onremove` is called.

The whole issue is caused by a missing `!` in the if statement. This change adds it in to fix the problem.

### DIFF
--- a/admin/javascript/LeftAndMain.js
+++ b/admin/javascript/LeftAndMain.js
@@ -1081,7 +1081,7 @@ jQuery.noConflict();
 				this._super();
 			},
 			onremove: function() {
-				if(this.data('button')) this.button('destroy');
+				if(!this.data('button')) this.button('destroy');
 				this._super();
 			}
 		});


dhensby:
Thanks!

dhensby:
@3Dgoo I've tried merging this up into master but it seems to break the behat tests.

So I'm going to ommit it from my merge up and open a new PR to try to get some feedback from the other core devs.

kinglozzer:
> When you want to add a button to the CMS but don't want LeftAndMain to apply jQuery UI button to it we add the `data-button="true"` attribute to our button.

I don’t think that this is intended behaviour. `if(!this.data('button')) this.button();` is just saying “if we’ve not already initialised a jQuery UI button, initialise it”, the fact that adding a `data-button` attribute stops it being initialised is just a coincidental side-effect of how `$.data` works. If it were intended to work that way, we’d be checking for the attribute directly with `if(!this.attr('data-button'))`.

I think the original behaviour was correct. We’ll likely now get errors because we’re trying to call `this.button('destroy')` on buttons that haven’t been initialised.

dhensby:
ergh - why does behat not complain for 3 but does for master :(

3Dgoo:
Damn. Sorry guys.

PR #6023, opened 2016-09-16
Fix button destroy bug in master

dhensby:
see #6010 

cc @3Dgoo


dhensby:
closing this because it looks like a bad interpretation of the api

PR #6024, opened 2016-09-18
Update file edit form

flamerohr:
* Added Created field to File/Image editor
* Added description element to FieldHolder
* Fix for a few React console errors


tractorcow:
Updated behaviour of html form schema strings. This new behaviour implements https://github.com/silverstripe/silverstripe-framework/issues/4938#textcast, but only for description field at this time.

PR #6031, opened 2016-09-19
BUG Fix error in campaign area

tractorcow:
Fixes alignment of create campaign fields so that they match the edit campaign form.
Removes use of auto scaffolder in ChangeSet to reduce undesirable scaffolding.


flamerohr:
Looks good, good catch with the create form not being inline

PR #6033, opened 2016-09-19
NEW Moving before/after handle request methods to RequestHandler

dhensby:
This builds on #3926

Really, the before/after methods should be on the same class as `handleRequest`, so now they are.

### DIFF
--- a/src/Control/Controller.php
+++ b/src/Control/Controller.php
@@ -67,15 +67,6 @@ class Controller extends RequestHandler implements TemplateGlobalProvider
     protected $basicAuthEnabled = true;
 
     /**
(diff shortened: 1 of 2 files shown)


sminnee:
Can you fix the notice-level errors in the tests

maxime-rainville:
@dhensby Are you still working on this? Presumably it should be against the 4 branch now.

maxime-rainville:
I think Dan lost interest. We'll close for now. We can reopen later if need be.

dhensby:
@maxime-rainville yeah - probably worth closing. This actually was much much harder to get working than expected :(

PR #6036, opened 2016-09-20
BUG fix issue with popup menu appearing on the left

tractorcow:
BUG fix incorrect selector on preview panel toggle
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1581

Requires https://github.com/silverstripe/silverstripe-cms/pull/1618

Before:

image

After:

image


clarkepaul:
The actual problem of the main actions was a missing CSS class (.ui-buttonset) surrounding the "Save" and "Publish" action. Is this something you noticed?

tractorcow:
Those are the old jquery ui styles though... I think we want to get away from relying on those?

tractorcow:
Ok, I've re-implemented the old behaviour, but rather than using jquery-ui styles I've used the bootstrap button row.

clarkepaul:
Looks good and test in IE, ready for merge @chillu

tractorcow:
Rebased to get rid of merge block.

clarkepaul:
Just doing testing on master now and I'm not seeing these changes coming through. Will look into it...

clarkepaul:
the cms branch wasn't was missed during merge, all done now

Issue #6027, opened 2016-09-19
PHPUnit bootstrap regression


gregsmirnov:
setting $_SERVER['HTTP_HOST'] in bootstrap.php:76

```
// Ensure Director::protocolAndHost() works
if (empty($_SERVER['HTTP_HOST'])) {
	$_SERVER['HTTP_HOST'] = 'localhost';
}
```

conflicts with framework/Core/Constants.php:109 where default cli request parametes are bootstrapped.
```
if(!isset($_SERVER['HTTP_HOST'])) {
	// Everything else
	$serverDefaults = array(
		'SERVER_PROTOCOL' => 'HTTP/1.1',
		'HTTP_ACCEPT' => 'text/plain;q=0.5',
		'HTTP_ACCEPT_LANGUAGE' => '*;q=0.5',
		'HTTP_ACCEPT_ENCODING' => '',
		'HTTP_ACCEPT_CHARSET' => 'ISO-8859-1;q=0.5',
		'SERVER_SIGNATURE' => 'Command-line PHP/' . phpversion(),
		'SERVER_SOFTWARE' => 'PHP/' . phpversion(),
		'SERVER_ADDR' => '127.0.0.1',
		'REMOTE_ADDR' => '127.0.0.1',
		'REQUEST_METHOD' => 'GET',
		'HTTP_USER_AGENT' => 'CLI',
	);
	$_SERVER = array_merge($serverDefaults, $_SERVER);
}
```

So other request parameters are missed later (like REQUEST_METHOD) when running from phpunit cli.

dhensby:
cc @sminnee regression from #5986

sminnee:
Happy to address the regression. Could you give an example of a failing test?

sminnee:
In particular, we probably want to start refactoring code to rely on these parameters less and use the HTTPRequest object instead.

gregsmirnov:
I've got a notice in FakeController, that $_SERVER['REQUEST_METHOD'] is not set
```
$request = new HTTPRequest(
	(isset($_SERVER['X-HTTP-Method-Override']))
		? $_SERVER['X-HTTP-Method-Override']
		: $_SERVER['REQUEST_METHOD'],
	'/'
);
```

sminnee:
@gregsmirnov can you give me the command that you're running? Are you running `vendor/bin/phpunit` or just plain `phpunit`? You need to use the former.

sminnee:
[...]

PR #6045, opened 2016-09-20
FIX: Make PR builds work.

sminnee:
PR builds are consistently failing because the version of framework
being used cannot be inferred. This environment variable provides the
answer.

It will need to be fixed for each major release, but I think that’s
less maintenance burden than complicating the build script to read this
from composer.json:extra.branch-alias.dev-master.


sminnee:
I added 2e054af7b1cdd8c647d88cfb0c452b683257c068 to possibly make #6027 clearer too.

tractorcow:
Oh wow, that's weird. :D I read up on this at https://github.com/composer/composer/issues/858. Thanks for looking into this.

PR #6049, opened 2016-09-21
Optimise webpack build time

chillu:
Merge with https://github.com/silverstripe/silverstripe-cms/pull/1617


sminnee:
Code review gets a :+1: from me. Haven't checked it out to try locally yet.

chillu:
Hold off on merge, just fixing some cms deps

admin/code/LeftAndMain.php
```diff
@@ -536,11 +536,10 @@ protected function init() {
 			window.ss.config = " . $this->getCombinedClientConfig() . ";
 		");
 
-		Requirements::javascript(ltrim(FRAMEWORK_ADMIN_DIR . '/client/dist/js/bundle-lib.js', '/'));
+		Requirements::javascript(FRAMEWORK_ADMIN_DIR . '/client/dist/js/vendor.js');
```
tractorcow: Can we use ltrim() to be consistent? @sminnee designed this to work so that we can run tests from the base dir (BASE_DIR === MODULE_DIR).

tractorcow: You can use `File::join_paths` if you want simpler code.

sminnee: It's not strictly needed for FRAMEWORK_ADMIN_DIR. I put this in for the glorious future where admin is also a separate module and we're running tests from its root with framework as a subfolder. ;-)

I wouldn't block merge for this because I didn't consistently apply it to all instances of FRAMEWORK_ADMIN_DIR, and it's a temporary code pattern anyway.

tractorcow: Yeah I guess it can be merged anyway. :D It's not a blocker.

chillu: Was wondering where that came from - I thought it might've been David misunderstanding how the framework constants work...

tractorcow: It's a "once we break out admin module" problem. :D

PR #6051, opened 2016-09-21
FIX: Break dist javascript and onto newlines.

sminnee:
This doesn’t have much impact on resulting JS size, but it will
hopefully make merge conflicts much less frequent.

The CSS growth is a little higher (6.5% increase in size) but is not
material.

If this materially reduces the number of merge conflicts we have, by
letting the git merge tools resolve some dist file mergers, I think it
would be worth it.

Some example changes in file size:

bundle.js 290K -> 301K
vendor.js 1,325K -> 1,321K
bundle.css 628K -> 669K


sminnee:
Here's an example of the diff for a small change: https://gist.github.com/7e19b6a2d9ab4e57c864dc13ac43119d

sminnee:
Here's an example of adding a method: (10 lines altered)
https://gist.github.com/0f37812ebc97ac74f53e4ba2d7d3f110

sminnee:
Here's an example of adding a new file: (154 lines altered)

PR #6053, opened 2016-09-21
Restore CMS build to CI

tractorcow:
Testing to see if these can be built.


chillu:
@tractorcow Now you have them both in the main list, and in allowed_failures - is that intentional?

tractorcow:
@ingo they need to be in both. allowed_failures doesn't add tests, it just makes prior-tests failable. :)

PR #6056, opened 2016-09-22
BUG Fix invalid file uploads not being validated

tractorcow:
Fixes issues in uploading for https://github.com/silverstripe/silverstripe-asset-admin/pull/256


tractorcow:
I would like to see any is_running_tests() invocations substituted with logic that can be explicitly turned on / off by sapphiretest (functionality is modified by tests, rather than core behaviour modifying itself via test detection). E.g. `Director` having a `force_redirect_behaviour` option.

PR #6060, opened 2016-09-22
BUG Fix incorrect ssviewertest path

tractorcow:
Seems to be a regression from the test refactor that incorrectly modified the expected path.


sminnee:
This probably relates to my changes to `themedJavascript()`, where you were now able to write in the full theme-relative URL and not merely the name of the the file within the javascript folder.

Seems fine, assuming the test passes.

PR #6061, opened 2016-09-22
BUG Fix incorrect backslash escaping in htaccess template

tractorcow:
Need to escape these slashes since they were resulting in unescaped `.` in the htaccess regex.

Merge with https://github.com/silverstripe/silverstripe-installer/pull/140


sminnee:
Seems like there's a relevant test failure on this one. Maybe the test is out of date?

tractorcow:
Tests are taking half an hour, but they are green (mostly).

tractorcow:
Now green, fully.

PR #6062, opened 2016-09-22
add IPUtils.php to control and implement symfony IpUtils

andrewandante:
This implements some Symfony functions to check that a given IP is in a range. This can now be used for the TRUSTED_PROXY_IPS setting, so that we can define a range rather than a list of IPs.

### DIFF
--- a/core/Constants.php
+++ b/core/Constants.php
@@ -105,7 +105,10 @@ if(!defined('TRUSTED_PROXY')) {
 			if(SS_TRUSTED_PROXY_IPS === '*') {
 				$trusted = true;
 			} elseif(isset($_SERVER['REMOTE_ADDR'])) {
-				$trusted = in_array($_SERVER['REMOTE_ADDR'], explode(',', SS_TRUSTED_PROXY_IPS));
+				if(!class_exists('SilverStripe\\Control\\Util\\IPUtils')) {
+					require_once FRAMEWORK_PATH . '/control/IPUtils.php';
+				};
+				$trusted = IPUtils::checkIP($_SERVER['REMOTE_ADDR'], explode(',', SS_TRUSTED_PROXY_IPS));
 			}
 		}
 	}
@@ -122,7 +125,7 @@ if(!defined('TRUSTED_PROXY')) {
  */
 if(!isset($_SERVER['HTTP_HOST'])) {
 	// HTTP_HOST, REQUEST_PORT, SCRIPT_NAME, and PHP_SELF
-	global $_FILE_TO_URL_MAPPING;	
+	global $_FILE_TO_URL_MAPPING;
 	if(isset($_FILE_TO_URL_MAPPING)) {
 		$fullPath = $testPath = realpath($_SERVER['SCRIPT_FILENAME']);
 		while($testPath && $testPath != '/' && !preg_match('/^[A-Z]:\\\\$/', $testPath)) {
@@ -182,7 +185,7 @@ if(!isset($_SERVER['HTTP_HOST'])) {
 	$trustedProxyHeader = (defined('SS_TRUSTED_PROXY_HOST_HEADER'))
 		? SS_TRUSTED_PROXY_HOST_HEADER
 		: 'HTTP_X_FORWARDED_HOST';
-		
+
 	if (TRUSTED_PROXY && !empty($_SERVER[$trustedProxyHeader])) {
 		// Get the first host, in case there's multiple separated through commas
 		$_SERVER['HTTP_HOST'] = strtok($_SERVER[$trustedProxyHeader], ',');
--- a/control/IPUtils.php
+++ b/control/IPUtils.php
@@ -1,0 +1,129 @@
+<?php
+/**
+ * These helpful functions were lifted from the Symfony library
+ * https://github.com/symfony/http-foundation/blob/master/LICENSE
+ *
+ * Http utility functions.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+namespace SilverStripe\Control\Util;
+/**
+ * Http utility functions.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class IPUtils {
+	/**
+	 * This class should not be instantiated.
+	 */
+	private function __construct()
+	{
+	}
+	/**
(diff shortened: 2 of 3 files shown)


core/Functions.php
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * @param string       $requestIp IP to check
+ * @param string|array $ips       List of IPs or subnets (can be a string if only a single one)
+ *
+ * @return bool Whether the IP is valid
+ *
+ * @package framework
+ * @subpackage core
+ */
+function checkIp($requestIp, $ips)
```
dhensby: Wrap these in `if (!function_exists())`

main.php
```diff
@@ -55,6 +55,7 @@
 /**
  * Include the defines that set BASE_PATH, etc
  */
+require_once('core/Functions.php');
```
dhensby: This needs adding to `Core.php`

tests/core/FunctionsTest.php
```diff
@@ -0,0 +1,81 @@
...
+ * These helpful tests were lifted from the Symfony library
+ * https://github.com/symfony/http-foundation/blob/master/LICENSE
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+require_once('core/Functions.php');
```
dhensby: Remove this

core/Functions.php
dhensby: `IP` should be all uppercase

stojg:
👍

sminnee:
Can we convert this to a namespaced class that's pulled in as needed via the PSR-4 autoloader, please? If a require statement is needed for SS3 so be it, but I'm reluctant to introduce more always-loaded boilerplate when a lot of our efforts go into making less of it.

tests/core/FunctionsTest.php
```diff
@@ -0,0 +1,79 @@
...
+	 * @expectedException \RuntimeException
+	 * @requires extension sockets
+	 */
+	public function testAnIpv6WithOptionDisabledIpv6()
+	{
+		if (defined('AF_INET6')) {
+			$this->markTestSkipped('Only works when PHP is compiled with the option "disable-ipv6".');
+		}
+
+		checkIp('2a01:198:603:0:396e:4789:8e99:890f', '2a01:198:603:0::/65');
```
halkyon: should be `checkIP` no?

core/Constants.php
```diff
@@ -105,7 +105,7 @@ function stripslashes_recursively(&$array) {
 			if(SS_TRUSTED_PROXY_IPS === '*') {
 				$trusted = true;
 			} elseif(isset($_SERVER['REMOTE_ADDR'])) {
```
sminnee: Put a `require_once()` call here.

core/Core.php
```diff
@@ -25,6 +25,7 @@
 /**
  * Include Constants (if it hasn't already been included) to pull in BASE_PATH, etc
  */
+require_once dirname(__FILE__) . '/Functions.php';
```
sminnee: Remove this require.

main.php
sminnee: It needs removing; the support file should be included immediately preceding the function call.

sminnee:
OK, so it looks like Symfony has already done this: https://github.com/symfony/http-foundation/blob/master/IpUtils.php

I'd suggest `SilverStripe\Control\Util\IpUtils` as the fully qualified classname, and place in "framework/control" in SS3, shifting to "framework/Control/Util" when you merge into SS4, to preserve PSR-4 naming conventions.

Because we're calling the method from Constants.php, we may not be able to rely on the auto-loader. For SS4 (to provide PSR-2 compliance) we'll likely need to refactor the function call out to be further down the call chain, but this is will be easier to do if we're not as hung up on backwards compatibility.

dhensby:
Hey @sminnee 

I've recommended to @andrewandante that we create the `Functions.php` file so that we can lay the groundwork for some better PSR-1/2 compliance, in particular the [side-effects recommendation](http://www.php-fig.org/psr/psr-1/#2-3-side-effects) (tl;dr files shouldn't both declare functions/classes AND make changes to state/vars/constants). I believe having a functions file like this is somewhat standard. The functions file is there to define functions available for the whole framework, not just for `Contstants.php` to use, in a subsequent PR and/or merging into master we can move the other global functions into it too.

I was hesitant to recommend explicitly requiring the functions file inside constants as I felt it was too tightly coupled (as justified by the reason above), though I suppose if we are including it *immediately* before every `Constants.php` require, there's not a big difference ;)

Regarding namespacing, this PR is a fix to keep it inline with how SS3 does it and the above reasoning has been taken to allow an easy upgrade in 4 (we move the `Functions.php` include to composer's autoloader like we have with `Constants.php`) and then these two files form a global set up for functions and constants needed to run the app (including the functions currently defined in `Core.php`). At this point we can implement namespaces and so on (all your points I agree with for implementing in 4, BUT I'd like to see us using htt-foundation so that we don't actually have to hand roll this at all.

I'd like to merge as is and I hope that makes sense from the above.

sminnee:
Keep it as a class. It's a class in Symfony. Seriously we're not Wordpress and it's not 2004.

sminnee:
I get what you mean about separating different concerns into different files, and if you want to move existing global functions into a Functions.php that's fine, but there's no way that we should be adding more global functions, especially not when they're copied from another implementation that does them as a class.

sminnee:
Although we can't add namespaces to existing classes in SS3 there's nothing to stop us putting namespaces on new classes we introduce.

dhensby:
OK - let's go with adding it as a class.

The auto-loader isn't ready by the time `Contants.php` is included so we'll either need to make it ready or we'll need to manually require the class.

Getting the auto-loader ready is out of scope for this PR IMO so we'll go with a manual require, I think.

andrewandante:
Ok, moved it from `core/Functions.php` to `control/IPUtil.php`, refactored to be a class, but at this stage still `require`d in one place.

Suggested namespace `SilverStripe\Control\Util\IpUtils` has been adopted (with uppercase IP)

sminnee:
Cool, just killing time before it's time to change a nappy,will merge in the morning.

PR #6073, opened 2016-09-23
[WIP] Shift admin to a separate module

sminnee:
## To do before merge
- [ ] Write up notes about how to re-generate the admin package so that it can be done at the time this PR is merged (to get any new changes)
- [ ] Move FRAMEWORK_ADMIN_DIR and FRAMEWORK_ADMIN_PATH definitions to the admin module
- [ ] Remove all references to them from framework
- [ ] Move TinyMCEConfig, HTMLEditorField, and UploadField to admin
- [ ] Simplify framework webpack
- [ ] Move CMSSecurity to admin

---

This is an initial proof-of-concept to get feedback, but it's looking functional so far. I deliberately left this out of alpha2 as it would be a distraction.

The content of admin has been shifted to this module: https://github.com/sminnee/silverstripe-admin

It's not been added to packagist yet (I'd want to move it to the silverstripe org before doing that) so to use it, you have to check it out manually into the admin/ subfolder.

A few points:
- TinyMCEConfig, HTMLEditorField, and UploadField depend on admin. Should these be moved to admin?
- CMSSecurity should be moved to admin.
- The webpack config and package.json in framework should be dramatically simplified. It's only use case is to build install.css and debug.css
- Translations have not yet been moved


Core/Constants.php
```diff
@@ -229,8 +229,8 @@
 	throw new Exception("Path error: FRAMEWORK_PATH " . FRAMEWORK_PATH . " not within BASE_PATH " . BASE_PATH);
 }
 
-define('FRAMEWORK_ADMIN_DIR', $frameworkDirSlashSuffix . 'admin');
-define('FRAMEWORK_ADMIN_PATH', FRAMEWORK_PATH . '/admin');
+define('FRAMEWORK_ADMIN_DIR', 'admin');
+define('FRAMEWORK_ADMIN_PATH', BASE_PATH . '/admin');
```
dhensby: This constant should be removed or added to the admin module?

tractorcow:
> TinyMCEConfig, HTMLEditorField, and UploadField depend on admin. Should these be moved to admin?

My feeling says yes. They only work in admin, not on the frontend.

> CMSSecurity should be moved to admin.

Yep. It's CMS only.

> The webpack config and package.json in framework should be dramatically simplified. It's only use case is to build install.css and debug.css

👍 

>Translations have not yet been moved

I think we are at the point where we need to branch translations for 4.0 anyway. We can't keep using the same master file as 3.x for much longer.

chillu:
Yay, we've done this now

Issue #4983, opened 2016-01-26
Varchar default size


JorisDebonnet:
Why is the default size of a Varchar DB field 50? I just wonder where this seemingly arbitrary number comes from. Why not 60, 100 or 250?

I believe that from a technical point of view, the most sensible *default* would be 255. 

An example is the Member.php class where FirstName and Surname use the default length, which is currently 50. This is of course fine in most cases, but a rare case may have their name truncated because of this default. Therefore it seems like a safer choice to go for the largest possible default size that does not come with a technical compromise (256 and beyond use 1 byte extra for storage). It is the most forgiving default for people that do not specify a specific length. 

Any drawbacks to making this change?

tractorcow:
I think we could bump it up for 4.0. want to PR it?

I like 255 personally, but you could put it as a private static on the Varchar class to make it more flexible. How does that sound? :P

dhensby:
This has been raised before... I'm not a fan of having an implied value. It should be set explicitly by the dev.

JorisDebonnet:
Requiring an explicit value would indeed be even more sensible, to avoid any confusion. It would equally fix my initial issue with a plain "Varchar" being much smaller than I would have accepted, so I would be happy if this is how it was done from 4.0 onward.

dhensby:
great :)

I did try to do it with 3.x with a deprecation notice but that means you can't have varchars instantiated like so `DBField::create_field('Varchar')` because it'd throw the notice that no length was supplied.

I guess we just go for straight breaking it in master :P

tractorcow:
It could throw the notice only when Varchar::requireField was invoked, though. It would continue to work fine as a casting helper until then.

JorisDebonnet:
can we have a milestone for 4.0 in any case, though? ^_^

sminnee:
If we don't have a default, then we're going to need to decide what default we put into `SiteTree.Title` and `Member.FirstName` What should our setting for title be?

My assumption is that this is going to come back to "well, we don't really know, so I guess we should make it 255".

That being the case, I would think that setting 255 as the default is better than forcing the same pointless discussion on any maintainer of a reusable module.

+1 for 255 default in master.

sminnee:
@JorisDebonnet I've assigned this to you; could you raise a PR?

JorisDebonnet:
Ooh, my first assigned issue :) fixed with #6083.

PR #6083, opened 2016-09-26
Set default Varchar size to 255, up from 50 (#4983)

JorisDebonnet:


dhensby:
Isnt' 255 as arbitrary as 50 and we're just kicking the can down the road for when people say "oh, why's the default 255"?

I'm still a believe we should just force people to define the length explicitly.

JorisDebonnet:
It may be arbitrary, but it's a lot less arbitrary than 50. 

Varchar(1-255) gets stored in length+1 bytes, while Varchar(256-65535) is stored in length+2 bytes (at least in MySQL). It's a small thing, but what it does show is that it makes no sense to use a hard limiting default less than 255. 

Although I understand your point that having any default is not great, it's certainly better to have it at 255 than at 50. If nothing else, it's more "expected" for historical reasons: "everyone does it". See: http://stackoverflow.com/questions/1217466/is-there-a-good-reason-i-see-varchar255-used-so-often-as-opposed-to-another-l

dhensby:
Sure, I understand the historical context of 255 varchar lengths. But I'm not convinced that 255 is better than 50 because it's the max you can store at length +1 byte.

The best length Varchar is the length you need.

Basically, if you want long text that doesn't need indexing and is of no pre-defined length, use TEXT. If you want to index the column and the length is predefined (or even if it isn't but a low maximum is acceptable) then use Varchar.

Issue #6075, opened 2016-09-23
SilverStripe 4 alpha 2 admin not working on windows


UndefinedOffset:
Looks like due to how the FRAMEWORK_ADMIN_DIR constant is built out and that windows uses backslash as a directory separator the admin just outright crashes on windows with an InvalidArgumentException. I've truncated the exception because of the template appearing in it but this is the gist of it.

```
[Emergency] Uncaught InvalidArgumentException: File \framework/admin/client/dist/js/vendor.js does not exist
GET /silverstripe-4/admin/pages/

Line 1001 in D:\www\silverstripe-4\framework\View\Requirements_Backend.php
Source

992 					$suffix = '?';
993 				}
994 				$suffix .= substr($fileOrUrl, strpos($fileOrUrl, '?') + 1);
995 				$fileOrUrl = substr($fileOrUrl, 0, strpos($fileOrUrl, '?'));
996 			} else {
997 				$suffix = '';
998 			}
999 			return "{$prefix}{$fileOrUrl}{$mtimesuffix}{$suffix}";
1000 		} else {
1001 			throw new InvalidArgumentException("File {$fileOrUrl} does not exist");
1002 		}
1003 	}
1004 
1005 	/**
1006 	 * Concatenate several css or javascript files into a single dynamically generated file. This
1007 	 * increases performance by fewer HTTP requests.

Trace

    SilverStripe\View\Requirements_Backend->pathForFile(\framework/admin/client/dist/js/vendor.js)
    Requirements_Backend.php:760
```

dhensby:
Possibly a regression from #5986??

UndefinedOffset:
Hmm ya looks like it maybe, it does appear to add the lines that are causing the constant to be set with the offending value. Not percicely sure what the fix would be right now. iPads aren't that great for programming ;)

UndefinedOffset:
@dhensby pull opened seems as if it's a simple fix of changing the hard coded forward slash being passed into the trim to the directory separator constant.

UndefinedOffset:
Pull merged so this is addressed

PR #6088, opened 2016-09-26
BUGIX: Fix regression causing the admin to crash on windows due to FRAMEWORK_DIR being prefixed by a backslash

UndefinedOffset:
This fixes the regression causing the admin to crash on windows due to FRAMEWORK_DIR being prefixed by a backslash, for context see #6075.


dhensby:
and now time for alpha3 ;)

PR #6089, opened 2016-09-26
Code coverage via CodeCov.io

chillu:
Needs an installer commit to add uncovered files
(or fixing travis-support to use the module's phpunit.xml).
Or witing for Sam's fix to run phpunit directly on the module
without travis-support. None of these block merge IMHO :)

```
<filter>
	<whitelist addUncoveredFilesFromWhitelist="true">
		<directory suffix=".php">framework/</directory>
		<exclude>
			<directory suffix=".php">framework/thirdparty/</directory>
			<directory suffix=".php">framework/admin/thirdparty/</directory>
		</exclude>
	</whitelist>
</filter>
```

/cc @sminnee

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ env:
     - secure: "SDGv49c2Ee2YBz7dATE3WnHSVSvJiRJ2BVtRasVshdNDNz3NBRzh13C2fDwTGBU1J6PxiQaGTXBy/BGsvbYk2BvdzHVwozkBpHVSaCNdarpCJ5yZZTqKC3mpA1S5353r5tqronwFuMDpftzXnRMfLZGGQ4kYb9hjV55+FPUTFPk=" # Encrypted ARTIFACTS_SECRET
 
 matrix:
+  fast_finish: true
   include:
     - php: 5.5
       env: DB=MYSQL PHPUNIT_TEST=1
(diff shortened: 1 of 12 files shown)


chillu:
Awesome, that worked!

sminnee:
I don't see why should need an installer fix? Simpler-test-run is already operational in the framework build. Am I missing something?

chillu:
Oh right, the change was so subtle in `travis.yml` that I've missed it (particularly since we still `vendor/bin/phpunit cms/tests`). I've added the filter directive as a new commit here.

chillu:
Total runtime was 17 mins (compared to 12 mins without coverage). But since a few builds take longer than 17 mins anyway, it doesn't make a large difference to overall run time of the builds.

dhensby:
We *really* should use `fast_finish` in travis if we're going to start doing code coverage like this

chillu:
@dhensby Sweet, added.

gregsmirnov:
All source code is expected to be in the "code" folder. Why do you add root folder in the whitelist?

tractorcow:
The framework module is the only module we don't follow this rule; As we break this module into it's separate component we expect that each component will adhere to the code-folder pattern.

gregsmirnov:
@tractorcow I was referring to modules only. Like this one [siteconfig](https://github.com/open-sausages/silverstripe-siteconfig/commit/6b751ad47af5b0a801f5359c1778066015a1f4a1)

PR #6091, opened 2016-09-26
FIX Handle validation errors in inline UploadFields

### LINKED ISSUE #6086: Validation failure on Page's has_one Image throws Exception
**SS version:** 3.4.1
---

Given I have a Page that has an Image (such as the BlogPost's FeaturedImage) and I add a field via extension to the Image, when editing inline and saving the Image an exception is thrown if the validation on the new field fails.

Here's a simple example to highlight the issue:

``` php
class MyTextField extends TextField
{
    /**
     * I'm grumpy. All the time.
     * @return false
     */
    public function validate($validator)
    {
        return false;
    }
}
```

And I add this via extension to the Image class:

``` yaml
# File: mysite/_config/app.yml
Image:
  extensions:
    - MyImageExtension
```

``` php
class MyImageExtension extends DataExtension
{
    private static $db = ['MyImageField' => 'Varchar'];
    public function getCMSFields()
    {
        $fields = parent::getCMSFields();
        $fields->addFieldToTab('Root.Main', MyTextField::create('MyImageField', 'My image field'));
        return $fields;
    }
}
```

When I save the inline edit form on the Image from say, a BlogPost or any Page that `has_one` Image, an Exception is thrown:

```
[User Error] Uncaught Exception: Object->__call(): the method 'redirectback' does not exist on 'UploadField_ItemHandler'
POST /admin/pages/edit/EditForm/field/FeaturedImage/item/29/EditForm

Line 780 in /var/www/framework/core/Object.php
...

Trace:

- Object->__call(redirectBack,Array)
    Form.php:541
- UploadField_ItemHandler->redirectBack()
    Form.php:541
- Form->getValidationErrorResponse()
    Form.php:458
...
```

If the validation passes for this same action then the Image is saved successfully.

I assume this also applies to Files.

robbieaverill:
- Implemented redirect method to natively handle validation errors being sent back to the UploadField
- Resolves #6086

I'm happy to discuss this approach - not sure if it's the "best" way to do it - the iframe and duplicated assets scared me off a little ;)

### DIFF
--- a/forms/UploadField.php
+++ b/forms/UploadField.php
@@ -1070,7 +1070,7 @@ class UploadField extends FileField {
 		// Note: Format of posted file parameters in php is a feature of using
 		// <input name='{$Name}[Uploads][]' /> for multiple file uploads
 		$tmpFiles = array();
-		if(	!empty($postVars['tmp_name'])
+		if( !empty($postVars['tmp_name'])
 			&& is_array($postVars['tmp_name'])
 			&& !empty($postVars['tmp_name']['Uploads'])
 		) {
@@ -1430,6 +1430,16 @@ class UploadField_ItemHandler extends RequestHandler {
 	}
 
 	/**
+	 * In place to handle inline UploadForm fields with Form validation errors
+	 *
+	 * @return ViewableData_Customised
+	 */
+	public function redirect()
+	{
+		return $this->edit($this->getRequest());
+	}
+
+	/**
 	 * @return Form
 	 */
 	public function EditForm() {
@@ -1451,6 +1461,7 @@ class UploadField_ItemHandler extends RequestHandler {
 		);
 		$form->loadDataFrom($file);
 		$form->addExtraClass('small');
+		$form->setRedirectToFormOnValidationError(true);
 
 		return $form;
 	}


dhensby:
Interesting, I take it there's no way to make this work without adding the `redirect` method?

We shouldn't really be putting new APIs in to the 3.4 branch...

robbieaverill:
@dhensby you guys would know better than me! [This is the code](https://github.com/silverstripe/silverstripe-framework/blob/3.4/forms/Form.php#L530-L542) that handled the form validation error and does something. I'm guessing that the UploadField might be a bit of a unique case in that it's run inside a frame. Looks like it's being treated as a controller, but doesn't share the same interface methods as a controller.

Agree re: adding methods to that branch - shall I change to `3` and rebase?

robbieaverill:
Hey @dhensby - I've changed to the `3` branch instead. Nice to see that GitHub didn't make me rebase :)

tractorcow:
@robbieaverill can I ask why is the method redirect() vs redirectBack()? I thought that this was the issue causing #6086 

If you are implement the redirect() method then it would pay to follow the same method signature as Controller::redirect, which redirects to the given parameter.

This seems like another side effect of having poorly defined roles for Controller vs RequestHandler.

robbieaverill:
Hey Dan, I used redirect back because is context seemed more relevant to
the fact that the form is an embedded frame.

I haven't done any edge case testing on this, just thought I would propose
it since it does fix the immediate problem (like a band aid I suppose).

I know what you mean re the method signature, however it is not actually a
redirect, so there is a semantic issue there which is more likely to be a
problem in the way that class is architected or implemented. I don't want
to go messing around with that of course. I didn't follow the method
signature because (A) it's not a controller, it's just pretending to be one
to fit into the Form logic and (B) the destination argument has no use at
all in its context. Since it doesn't inherit any kind of interface for the
method I thought it could be treated more as a patch than anything.

TL;DR: it's a quick fix for a bigger problem, one which I don't have time
to look very far into I'm afraid.

I'm happy for you guys to do whatever you see fit with this PR, but the
issue does exist even with a standard FormField with a validation rule,
e.g. Max length as provided out of the box.

Cheers :-)


-- 
*Robbie Averill*
LA PETITE MANOUCHE - NZ Gypsy Jazz Guitar Duo
Phone: +64 27 321 4758
contact@lapetitemanouche.com
www.lapetitemanouche.com

tractorcow:
I had another think about it... and I'm wondering if we don't move some logic up from Controller into RequestHandler: redirect(), redirectBack(), and Link(). That way we can explicitly allow Form::setController() to take a request handler (which we do in many cases), rather than require a controller.

robbieaverill:
Yes please @tractorcow

robbieaverill:
Closing - not sure this is the best way to fix the problem for now.

tractorcow:
I've proposed this in https://github.com/silverstripe/silverstripe-framework/issues/6362

PR #6094, opened 2016-09-26
Update 14_PHP_Coding_Conventions.md

madmatt:
Include a note preferring type-safe identical comparisons (===) over equality (==) where possible.

### DIFF
--- a/docs/en/05_Contributing/14_PHP_Coding_Conventions.md
+++ b/docs/en/05_Contributing/14_PHP_Coding_Conventions.md
@@ -329,6 +329,18 @@ Statements with `if` can be written without braces on a single line as the block
 	if($a == $b) doThis();
 	else doThat();
 
+#### Prefer identical (===) comparisons over equality (==)
+
+Where possible, use type-strict identical comparisons instead of loosely typed equality comparisons.
+Read more in the PHP documentation for [comparison operators](http://php.net/manual/en/language.operators.comparison.php) and [object comparison](http://php.net/manual/en/language.oop5.object-comparison.php).
+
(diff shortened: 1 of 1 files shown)


chillu:
Yeah, PSR-1 and PSR-2 don't mention that - but it's good practice. Can we get at at least three @silverstripe/core-team votes on this to merge? I think its unrealistic to block this doc PR until we've cleaned up equality operators in core, too much work and risk of regressions.

docs/en/05_Contributing/14_PHP_Coding_Conventions.md
```diff
@@ -329,6 +329,18 @@ Statements with `if` can be written without braces on a single line as the block
...
 	else doThat();
 
+#### Prefer identical (===) comparisons over equality (==)
+
+Where possible, use type-strict identical comparisons instead of loosely typed equality comparisons.
+Read more in the PHP documentation for [comparison operators](http://php.net/manual/en/language.operators.comparison.php) and [object comparison](http://php.net/manual/en/language.oop5.object-comparison.php).
+
+	:::php
+	// good - only need to cast to (int) if $a might not already be an int
+	if((int)$a === 100) doThis();
```
jonom: PSR-2 requires `{braces}` for if statements, we should probably extend that to doc examples?

jonom: Also if we do want to follow PSR-2 in code examples there should be a space after the control keyword (if) http://www.php-fig.org/psr/psr-2/#5-control-structures

PR #6100, opened 2016-09-27
Update jquery.js reference

thezenmonkey:
jquery is no longer included at the path specified. Shoudl be in the admin directory


baukezwaan:
I think it also needs to be fixed in [config-form.html](https://github.com/silverstripe/silverstripe-framework/blob/master/Dev/Install/config-form.html#L9)
This way the js-errors on the install.php page are also fixed...

dhensby:
We're looking at pulling admin out into its own module soon so this isn't a long term fix. Better to use Google's CDN?

thezenmonkey:
Which version of jquery shoud be we be fulling from the CDN?

dhensby:
[simple uses 1.7.2](https://github.com/silverstripe-themes/silverstripe-simple/pull/42/files), so probably that...

dhensby:
thanks

dhensby:
@baukezwaan sorry, I missed your comment some how! Would you mind opening a PR to fix config-form.html?

PR #6114, opened 2016-09-29
API Implement many_many through

### LINKED ISSUE #5615: RFC: Version many_many tables
Simple many_many versioning via a versioned mapping table with `has_many_through` option (similar to ruby on rails).

This should function similarly to a has_many table with extra_fields, but allow versioning, and have a physically backed DataObject for the mapping table.

Acceptance criteria:

 * The API syntax in `$many_many` is confirmed
 * Implementation is complete
 * A many-many-through relationship operates as closely to a many-many relationship (properties on the join object should be included in the many-many SS_List, just like many-many-extrafields)
 * Tests demonstrate that these relationships can be versioned
 
----

**UPDATE:** Damian has posted 4 options below for how we could implement this and so it's worth treating this as a small RFC.

----

**UPDATE 2:** PR at https://github.com/silverstripe/silverstripe-framework/pull/6114

tractorcow:
Fixes #5615

Merge with CMS changes: https://github.com/silverstripe/silverstripe-cms/pull/1639

This change introduces a slight refactor in model relationship handling; Validation has been updated, and moved to DataObjectSchema.

The basic implementation of this feature is that a many_many relation now utilises a nested has_many relation, which is rewritten as a joined subquery within it in place of a join table.

In order to ensure that both has_many and many_many queries are joined, and independently augmented correctly, a ManyManyThroughQueryManipulator object has been implemented to merge them during `getFinalisedQuery()`.

I've also taken advantage of the `::class` language feature in PHP 5.5 to replace some class literals.

Includes the following ORM changes:
- API Remove DataObject::validateModelDefinitions, and move to DataObjectSchema
- API Remove deprecated 3.0 syntax for addSelect()
- API made DataList::createDataObject public
- API Move component parsing logic to DataObjectSchema
- API Remove support for triangular has_many / belongs_many relationships

### DIFF
--- a/ORM/DataObject.php
+++ b/ORM/DataObject.php
@@ -1935,35 +1935,10 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	 * format, or RecordClass.FieldClass(args) format if $includeClass is true
 	 */
 	public function db($fieldName = null, $includeClass = false) {
-		$classes = ClassInfo::ancestry($this, true);
-
-		// If we're looking for a specific field, we want to hit subclasses first as they may override field types
-		if($fieldName) {
-			$classes = array_reverse($classes);
-		}
-
-		$db = array();
-		foreach($classes as $class) {
-			// Merge fields with new fields and composite fields
-			$fields = self::database_fields($class);
-			$compositeFields = self::composite_fields($class, false);
-			$db = array_merge($db, $fields, $compositeFields);
-
-			// Check for search field
-			if($fieldName && isset($db[$fieldName])) {
-				// Return found field
-				if(!$includeClass) {
-					return $db[$fieldName];
-				}
-				return $class . "." . $db[$fieldName];
-			}
-		}
-
-		// At end of search complete
-		if($fieldName) {
-			return null;
+		if ($fieldName) {
+			return static::getSchema()->fieldSpecification(static::class, $fieldName, $includeClass);
 		} else {
-			return $db;
+			return static::getSchema()->fieldSpecifications(static::class);
 		}
 	}
 
@@ -2071,7 +2046,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	/**
 	 * Return information about a specific many_many component. Returns a numeric array.
 	 * The first item in the array will be the class name of the relation: either
-	 * RELATION_MANY_MANY or RELATION_MANY_MANY_THROUGH constant value.
+	 * ManyManyList or ManyManyThroughList.
 	 *
 	 * Standard many_many return type is:
 	 *
@@ -3870,10 +3845,19 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	 * Set joining object
 	 *
 	 * @param DataObject $object
+	 * @param string $alias Alias
 	 * @return $this
 	 */
-	public function setJoin(DataObject $object) {
+	public function setJoin(DataObject $object, $alias = null) {
 		$this->joinRecord = $object;
+		if ($alias) {
+			if ($this->db($alias)) {
+				throw new InvalidArgumentException(
+					"Joined record $alias cannot also be a db field"
+				);
+			}
+			$this->record[$alias] = $object;
+		}
 		return $this;
 	}
 
--- a/ORM/DataObject.php
+++ b/ORM/DataObject.php
@@ -154,6 +154,13 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	protected $record;
 
 	/**
+	 * If selected through a many_many through relation, this is the instance of the through record
+	 *
+	 * @var DataObject
+	 */
+	protected $joinRecord;
+
+	/**
 	 * Represents a field that hasn't changed (before === after, thus before == after)
 	 */
 	const CHANGE_NONE = 0;
@@ -1422,13 +1429,16 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	 * @return DataObject $this
 	 */
 	public function writeComponents($recursive = false) {
-		if(!$this->components) {
-			return $this;
+		if($this->components) {
+			foreach ($this->components as $component) {
+				$component->write(false, false, false, $recursive);
+			}
 		}
 
-		foreach($this->components as $component) {
-			$component->write(false, false, false, $recursive);
+		if ($join = $this->getJoin()) {
+			$join->write(false, false, false, $recursive);
 		}
+
 		return $this;
 	}
 
@@ -1513,7 +1523,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 			$joinID    = $this->getField($joinField);
 
 			// Extract class name for polymorphic relations
-			if($class === 'SilverStripe\\ORM\\DataObject') {
+			if($class === __CLASS__) {
 				$class = $this->getField($componentName . 'Class');
 				if(empty($class)) return null;
 			}
@@ -1624,12 +1634,20 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	 * @return string Class name, or null if not found.
 	 */
 	public function getRelationClass($relationName) {
+		// Parse many_many
+		$manyManyComponent = $this->manyManyComponent($relationName);
+		if ($manyManyComponent) {
+			list(
+				$relationClass, $parentClass, $componentClass,
+				$parentField, $childField, $tableOrClass
+			) = $manyManyComponent;
+			return $componentClass;
+		}
+
 		// Go through all relationship configuration fields.
 		$candidates = array_merge(
 			($relations = Config::inst()->get($this->class, 'has_one')) ? $relations : array(),
 			($relations = Config::inst()->get($this->class, 'has_many')) ? $relations : array(),
-			($relations = Config::inst()->get($this->class, 'many_many')) ? $relations : array(),
-			($relations = Config::inst()->get($this->class, 'belongs_many_many')) ? $relations : array(),
 			($relations = Config::inst()->get($this->class, 'belongs_to')) ? $relations : array()
 		);
 
@@ -1682,8 +1700,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	 * @throws InvalidArgumentException
 	 */
 	public function inferReciprocalComponent($remoteClass, $remoteRelation) {
-		/** @var DataObject $remote */
-		$remote = $remoteClass::singleton();
+		$remote = DataObject::singleton($remoteClass);
 		$class = $remote->getRelationClass($remoteRelation);
 
 		// Validate arguments
@@ -1755,13 +1772,15 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 			case 'many_many':
 			case 'belongs_many_many': {
 				// Get components and extra fields from parent
-				list($componentClass, $parentClass, $componentField, $parentField, $table)
+				list($relationClass, $componentClass, $parentClass, $componentField, $parentField, $table)
 					= $remote->manyManyComponent($remoteRelation);
 				$extraFields = $remote->manyManyExtraFieldsForComponent($remoteRelation) ?: array();
 
 				// Reverse parent and component fields and create an inverse ManyManyList
-				/** @var ManyManyList $result */
-				$result = ManyManyList::create($componentClass, $table, $componentField, $parentField, $extraFields);
+				/** @var RelationList $result */
+				$result = Injector::inst()->create(
+					$relationClass, $componentClass, $table, $componentField, $parentField, $extraFields
+				);
 				if($this->model) {
 					$result->setDataModel($this->model);
 				}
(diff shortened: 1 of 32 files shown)


ORM/DataObject.php
```diff
@@ -2149,78 +2069,27 @@ public function manyMany() {
 	}
 
 	/**
-	 * Return information about a specific many_many component. Returns a numeric array of:
+	 * Return information about a specific many_many component. Returns a numeric array.
+	 * The first item in the array will be the class name of the relation: either
+	 * RELATION_MANY_MANY or RELATION_MANY_MANY_THROUGH constant value.
```
kinglozzer: Can’t find these constants, doesn’t it just return the FQCN?

tractorcow: Sorry, those were added but then removed. I'll update the doc.

ORM/DataObjectSchema.php
```diff
@@ -367,4 +368,396 @@ public function classForField($candidateClass, $fieldName)  {
...
+	 * @return string|null
+	 */
+	protected function getManyManyInverseRelationship($childClass, $parentClass)
+	{
+		$otherManyMany = Config::inst()->get($childClass, 'many_many', Config::UNINHERITED);
+		if (!$otherManyMany) {
+			return null;
+		}
+		foreach ($otherManyMany as $inverseComponentName => $nextClass) {
+			if ($nextClass === $parentClass) {
```
kinglozzer: The logic has changed here slightly - it used to also check `|| is_subclass_of($parentClass, $nextClass)`. I’ve run through it mentally a few times and I’m sure the removal of that logic is fine, just noting it :)

ORM/ManyManyList.php
```diff
@@ -210,23 +211,30 @@ public function add($item, $extraFields = array()) {
...
+		}
+		if (empty($itemID)) {
+			throw new InvalidArgumentException("ManyManyList::add() doesn't accept unsaved records");
 		}
 
 		// Validate foreignID
 		$foreignIDs = $this->getForeignID();
 		if(empty($foreignIDs)) {
-			throw new Exception("ManyManyList::add() can't be called until a foreign ID is set", E_USER_WARNING);
+			throw new BadMethodCallException("ManyManyList::add() can't be called until a foreign ID is set", E_USER_WARNING);
```
kinglozzer: Can we remove the `E_USER_WARNING`?

ORM/ManyManyThroughList.php
```diff
@@ -0,0 +1,189 @@
...
+			);
+		}
+		if (empty($itemID)) {
+			throw new InvalidArgumentException("ManyManyThroughList::add() doesn't accept unsaved records");
+		}
+
+		// Validate foreignID
+		$foreignIDs = $this->getForeignID();
+		if (empty($foreignIDs)) {
+			throw new BadMethodCallException("ManyManyList::add() can't be called until a foreign ID is set", E_USER_WARNING);
```
kinglozzer: `ManyManyThroughList`, and `E_USER_WARNING` to be removed?

docs/en/02_Developer_Guides/00_Model/02_Relations.md
```diff
@@ -213,50 +213,139 @@ This is not mandatory unless the relationship would be otherwise ambiguous.
...
+	  ];
+	}
+
+In order to filter on the join table during queries, you can use the "Join" table alias
+for any sql conditions.
+
 
 	:::php
 	$team = Team::get()->byId(1);
+	$supporters = $team->Supporters()->where(['"Join"."Ranking"' => 1]);
```
kinglozzer: Are we able to use the join table name directly here? To use the physician/patient example from [Ruby on Rails](http://guides.rubyonrails.org/association_basics.html#the-has-many-through-association):

```php
$patientsToday = $physician->Patients()->where(['"Appointment"."Date"' => date('Y-m-d')]);
```

Is that possible, or do we have to use the `Join` alias?

docs/en/02_Developer_Guides/00_Model/02_Relations.md
```diff
@@ -213,50 +213,139 @@ This is not mandatory unless the relationship would be otherwise ambiguous.
...
 
-	$supporters = $team->Supporters();
-	// returns a 'ManyManyList' instance.
 
+Note: ->filter() currently does not support joined fields natively.
```
kinglozzer: :( will this be possible in future?

kinglozzer:
Nice work! I think we’ll need consensus on the syntax for declaring the many-many-through relationships, but I like the approaches you’ve used for tackling this

ORM/DataObjectSchema.php
tractorcow: Yeah, supporting for what I called "triangular relationships" was flakey, and led to some weird behaviour infering reciprocal relationships.

ORM/ManyManyList.php
tractorcow: Ah yes that should not be there.

ORM/ManyManyThroughList.php
tractorcow: Will do.

docs/en/02_Developer_Guides/00_Model/02_Relations.md
tractorcow: Hm, possibly, although it would make the api more complex.

docs/en/02_Developer_Guides/00_Model/02_Relations.md
tractorcow: Unfortunately in order to support arbitrary augmentSQL'd join queries, the join is wrapped within a subquery with an alias, so it would be quite difficult for the datalist api to infer from the nested relationship robustly. It would essentially require this solution to be re-implemented with some unknown method.

docs/en/02_Developer_Guides/00_Model/02_Relations.md
tractorcow: On second thoughts, I think it's worth exploring. I'll give it a try. :)

tractorcow:
@kinglozzer I've responded to your feedback.

- Join alias is now the join table name, not just "Join". '$Join' can still be used in templates as well as the class name, however.
- I've added some extra validation to handle clashes between the new alias and other fields on either side of the relation.
- I've moved DataObject::db() into DataObjectSchema()
- I've completely removed the constant that used to be the join alias. It's now a method `ManyManyThroughQueryManipulator::getJoinAlias()`
- I've removed all the extra arguments to exception constructors.

docs/en/02_Developer_Guides/00_Model/02_Relations.md
tractorcow: Done!

ORM/ManyManyThroughList.php
tractorcow: Done!

ORM/ManyManyList.php
tractorcow: Done!

ORM/DataObject.php
tractorcow: Done!

kinglozzer:
Great, thanks! I’ve checked the changes out locally and tested them, and everything works how I was expecting. The lack of `filter()` support is a bummer, but your explanation makes sense - I did play around a little trying to find a solution, but encountered the exact problems you described.

I’m happy with these changes, so I’m going to hit the magic approve button. I won’t merge yet, as I’d like some more eyes on this first :)

tractorcow:
@kinglozzer : @sminnee has looked over this and said that it would be fine if you were to take over merging. ;)

sminnee:
I haven't had a detailed read yet but I'll do so today click a magic approve button when I'm ready.

tractorcow:
Thanks guys. :)

tests/model/ManyManyThroughListTest.php
```diff
@@ -0,0 +1,275 @@
...
+	protected $extraDataObjects = [
+		ManyManyThroughListTest_Item::class,
+		ManyManyThroughListTest_JoinObject::class,
+		ManyManyThroughListTest_Object::class,
+		ManyManyThroughListTest_VersionedItem::class,
+		ManyManyThroughListTest_VersionedJoinObject::class,
+		ManyManyThroughListTest_VersionedObject::class,
+	];
+
+	public function testSelectJoin() {
```
sminnee: Add a test for sorting by `SortOrder`. Sorting by a field on the join object is a pretty critical feature.

ORM/DataObject.php
```diff
@@ -1682,8 +1700,7 @@ public function getRelationType($component) {
 	 * @throws InvalidArgumentException
 	 */
 	public function inferReciprocalComponent($remoteClass, $remoteRelation) {
-		/** @var DataObject $remote */
-		$remote = $remoteClass::singleton();
+		$remote = DataObject::singleton($remoteClass);
```
tractorcow: DataObject::singleton() simple invokes Injector::inst()->get(), but it has `@return static` to enable type-hinting, whereas Injector::inst()->get() has none. I've found that typehinting in code helps uncover bugs a lot easier, so would you agree it's beneficial to leave it?

tractorcow:
I have a feeling the current getSchema() api might need a little bit of think. Perhaps getSchema() should return a schema object scoped to the given parent, similar to Config_forClass?

tractorcow:
I'm going to do a bunch more cleanup of class constants, as well as consolidating some of the DataObjectSchema API and remove methods, as suggested.

I think it's ok for DataObjectSchema to be big, as long as it's well defined, and reduces complexity of DataObject itself; All it does it provide information on how classes, fields, tables, and relations should fit together; DataObject itself still handles physical relationships between instances.

The filtering issue is a tough one and difficult to crack easily in a way that works the way we'd expect it to. The issue I had were to do with the immutability of DataList and dealing with multiple datalists. The analogy I would give is to take a russian doll design pattern and apply it to a binary tree;

The solution to make ->filter() work natively with nested datalists is possible but would just be more complicated. I anticipate that rather than creating the inner HasMany list dynamically, we would need to hold an instance for the lifetime of the outer datalist; Probably with a more complicated alterDataQuery / __clone that would ensure consistency. This is probably something I'll approach when I've more time to come back to it. :D

The downside of not doing this is that you need to write `->where(['"JoinClass"."Field"' => $value])` instead of `->filter('JoinClass.Field', $value)`. It also means needing to express search filters such as BeginsWith / EndsWith etc explicitly via SQL.

ORM/DataObject.php
sminnee: Fair call; fine to leave for now. The lack of type hinting from our DI system can be a problem for another day ;-)

tractorcow:
@sminnee I've added the sorting unit tests. Additionally in a much larger commit I've moved a large quantity of db schema methods to DataObjectSchema.

I spoke with @chillu earlier and we came up with a more useful version of the `db` method that would allow us to supercede a large number of methods on DataObject, meaning we can simplify the class.

tractorcow:
New CMS pr at https://github.com/silverstripe/silverstripe-cms/pull/1639

ORM/DataObject.php
```diff
@@ -3407,10 +3276,8 @@ public function requireTable() {
 
 		$indexes = $this->databaseIndexes();
 
-		// Validate relationship configuration
-		$this->validateModelDefinitions();
 		if($fields) {
-			$hasAutoIncPK = get_parent_class($this) === 'SilverStripe\ORM\DataObject';
+			$hasAutoIncPK = get_parent_class($this) === __CLASS__;
```
tractorcow: Done!

ORM/DataObject.php
```diff
@@ -3421,29 +3288,34 @@ public function requireTable() {
...
-			foreach($manyMany as $relationship => $childClass) {
-				// Build field list
-				if($this->class === $childClass) {
-					$childField = "ChildID";
-				} else {
-					$childTable = $this->getSchema()->tableName($childClass);
-					$childField = "{$childTable}ID";
+			foreach($manyMany as $component => $spec) {
+				// Get many_many spec
+				$manyManyComponent = $this->getSchema()->manyManyComponent(get_class($this), $component);
```
tractorcow: Done.

tests/model/ManyManyThroughListTest.php
tractorcow: Done.

tests/model/ManyManyThroughListTest.php
```diff
@@ -0,0 +1,275 @@
...
+		]
+	];
+}
+
+/**
+ * @property string $Title
+ * @method ManyManyThroughListTest_Object Parent()
+ * @method ManyManyThroughListTest_Item Child()
+ */
+class ManyManyThroughListTest_JoinObject extends DataObject implements TestOnly
```
tractorcow: Done.

ORM/DataObject.php
```diff
@@ -270,88 +283,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
...
-	/**
-	 * Returns the field class if the given db field on the class is a composite field.
-	 * Will check all applicable ancestor classes and aggregate results.
-	 *
-	 * @param string $class Class to check
-	 * @param string $name Field to check
-	 * @param boolean $aggregated True if parent classes should be checked, or false to limit to this class
-	 * @return string|false Class spec name of composite field if it exists, or false if not
-	 */
-	public static function is_composite_field($class, $name, $aggregated = true) {
```
tractorcow: The removal of all of these static methods to a non-static DataObjectSchema service means we can now DI our schema. :)

ORM/DataObjectSchema.php
```diff
@@ -137,6 +138,62 @@ public function baseDataTable($class) {
...
+	 * Get all DB field specifications for a class, including ancestors and composite fields.
+	 *
+	 * @param string|DataObject $classOrInstance
+	 * @param array $options Array of options. Specify any number of the below:
+	 *  - `uninherited`: Set to true to limit to only this table
+	 *  - `dbOnly`: Exclude virtual fields (such as composite fields), and only include fields with a db column.
+	 *  - `includeClass`: If true prefix the field specification with the class name in RecordClass.Column(spec) format.
+	 * @return array
+	 */
+	public function fieldSpecs($classOrInstance, $options = []) {
```
tractorcow: This is what was DataObject::db()

ORM/DataObjectSchema.php
```diff
@@ -212,15 +269,48 @@ protected function buildTableName($class) {
 	 * "ID" will be included on every table.
 	 *
 	 * @param string $class Class name to query from
+	 * @param bool $aggregated Include fields in entire hierarchy, rather than just on this table
 	 * @return array Map of fieldname to specification, similiar to {@link DataObject::$db}.
 	 */
-	public function databaseFields($class) {
+	public function databaseFields($class, $aggregated = true) {
```
tractorcow: Added $aggregate to make this more symmetrical with compositeFields()

tractorcow:
Oh finally, green CI. :)

ORM/DataObjectSchema.php
kinglozzer: Being really nitpicky but... could we use a bitmask instead of an array of case-sensitive strings for passing these options around? Personal preference, so feel free to say no if you disagree!

ORM/DataObjectSchema.php
```diff
@@ -347,24 +450,476 @@ public function tableForField($candidateClass, $fieldName) {
...
 			}
 			$candidateClass = get_parent_class($candidateClass);
 		}
 		return null;
 	}
+
+	/**
+	 * Return information about a specific many_many component. Returns a numeric array.
+	 * The first item in the array will be the class name of the relation: either
+	 * RELATION_MANY_MANY or RELATION_MANY_MANY_THROUGH constant value.
```
kinglozzer: Looks like this docblock was missed earlier - those constants were removed

ORM/DataObject.php
```diff
@@ -410,13 +342,13 @@ public function __construct($record = null, $isSingleton = false, $model = null,
...
 		if(!empty($record['ID'])) {
-			$currentObj = get_class($this);
-			while($currentObj != 'SilverStripe\ORM\DataObject') {
-				$fields = self::custom_database_fields($currentObj);
-				foreach($fields as $field => $type) {
-					if(!array_key_exists($field, $record)) $this->record[$field.'_Lazy'] = $currentObj;
+			// Get all field specs scoped to class for later lazy loading
+			$fields = static::getSchema()->fieldSpecs(static::class, ['includeClass', 'dbOnly']);
+			foreach($fields as $field => $fieldSpec) {
+				$fieldClass = strtok($fieldSpec, ".");
```
kinglozzer: Why is this necessary? I can’t spot where it came from 😅

kinglozzer:
Reviewed the new commit, nothing merge blocking 👍

ORM/DataObjectSchema.php
sminnee: Agree with @kinglozzer: a bitmask based on class constants would be less error-prone and more easily self-documenting.

ORM/DataObjectSchema.php
tractorcow: Must have been a rebase issue...

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/pull/6161

ORM/DataObjectSchema.php
tractorcow: https://github.com/silverstripe/silverstripe-framework/pull/6161

Issue #6058, opened 2016-09-22
Replace success icon within add to campaign modal


clarkepaul:
This icon isn't using the one specified in designs, also animation is pretty rough. Icon should use SilverStripe icon font class="font-icon-happy" or font-icon-sad
![image](https://cloud.githubusercontent.com/assets/555033/18736326/2c984db4-80da-11e6-9aac-edff1010bafd.png)

clarkepaul:
PR https://github.com/silverstripe/silverstripe-framework/pull/6118

clarkepaul:
End result is improved animation and design matches those provided.
![image](https://cloud.githubusercontent.com/assets/555033/18973168/bd7e5b06-86f8-11e6-9d1c-c622ec491865.png)

clarkepaul:
Rebased and ready for review.

tractorcow:
Reviewing it now.

clarkepaul:
Removed legacy styles

PR #6118, opened 2016-09-29
Add to campaign styles UI enhancements

clarkepaul:
Relates to issue https://github.com/silverstripe/silverstripe-framework/issues/6058

### DIFF
--- a/admin/client/src/components/Modal/Modal.scss
+++ b/admin/client/src/components/Modal/Modal.scss
@@ -1,0 +1,112 @@
+// Based on bootstrap modal
+.modal-content {
+  overflow: hidden;
+  border: 0;
+}
+
+
+// Modal header
+// Top section of the modal w/ title and dismiss
+.modal-header {
+  height: $toolbar-total-height;
+
+  // Close icon
+  .close {
+    top: $spacer-sm;
+    right: $spacer-sm;
+    position: absolute;
+    z-index: 2;
+    margin-right: 0;
+    height: 32px;
+
+    span {
+      overflow: hidden;
+      display: block;
+      margin-top: -7px;
+    }
+  }
+}
+
+.modal-body {
+  min-height: 200px;
+}
+
+
+// Modal slide up response
+@keyframes modal--slideup {
+  from {
+    bottom: -100%;
+    opacity: .5;
+  }
+
+  to {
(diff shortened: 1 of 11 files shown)


clarkepaul:
I just remember this isn't rebased, so I will do that shortly.

clarkepaul:
After rebasing I've noticed the file area add to campaign area need updating too. I possibly need to create an asset-admin branch.

clarkepaul:
Requires AssetAdmin branch for add to campaign within the files area. https://github.com/silverstripe/silverstripe-asset-admin/pull/282

admin/client/src/components/Modal/Modal.scss
```diff
@@ -0,0 +1,112 @@
...
+  .chosen-results > li {
+    padding: $spacer-y/1.5 $spacer-x;
+  }
+
+  .disabled-result {
+    font-style: italic;
+
+    &::after {
+      font-style: normal;
+      content: "Already contains item";
```
tractorcow: question: `modal` is a generic term given to popups that block access to the underlying page. This CSS is scoped to `modal__dialog`, but it seems very specific to the `add to campaign` feature, which is really only a single-use-case for modals.

Is there a way that the add-to-campaign content could be separated from the generic modal styles?

tractorcow: Perhaps as a simple win, rename from modal to campaignmodal?

clarkepaul: I was hoping that any styles here could be used generically for other modals. When I create new patterns I would like to create them as a component and not just for one use case like "add to campaign". You are right though in saying that the styles at the bottom are campaign specific with the disabled result—although it doesn't actually work or do anything, at one stage they did work. 

Best I move just styles out. Feels weird deleting it even though they don't work.

clarkepaul:
I've removed the specific styles for campaign admin so everything else should be generic to any modal.

tractorcow:
It looks like you force-pushed over the fix I added to the linting issue.

tractorcow:
I've re-pushed the fix and retested it. Everything is good to merge on green.

tractorcow:
Behat tests run perfectly fine locally.

PR #6119, opened 2016-09-29
API Update form schema to be more responsive to dynamic form changes

tractorcow:
Required for https://github.com/silverstripe/silverstripe-asset-admin/pull/280

### DIFF
--- a/Control/Controller.php
+++ b/Control/Controller.php
@@ -695,12 +695,15 @@ class Controller extends RequestHandler implements TemplateGlobalProvider {
 	 *
 	 * Caution: All parameters are expected to be URI-encoded already.
 	 *
-	 * @param string
-	 *
+	 * @param string|array $arg,.. One or more link segments, or list of link segments as an array
 	 * @return string
 	 */
-	public static function join_links() {
-		$args = func_get_args();
+	public static function join_links($arg = null) {
+		if (func_num_args() === 1 && is_array($arg)) {
+			$args = $arg;
+		} else {
+			$args = func_get_args();
+		}
 		$result = "";
 		$queryargs = array();
 		$fragmentIdentifier = null;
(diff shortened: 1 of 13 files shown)


flamerohr:
with `this.submitApi` is getting more complex, it'd be good to apply some unit tests for some of the expected actions that need to happen.

Control/Controller.php
```diff
@@ -695,12 +695,15 @@ public function setSession(Session $session) {
...
 	 *
-	 * @param string
-	 *
+	 * @param string|array $arg,.. One or more link segments, or list of link segments as an array
 	 * @return string
 	 */
-	public static function join_links() {
-		$args = func_get_args();
+	public static function join_links($arg = null) {
+		if (func_num_args() === 1 && is_array($arg)) {
```
dhensby: I'm not sure I see a benefit to changing the API of `join_links` in this way... What's the thinking?

tractorcow: Because otherwise you can't pass in an arbitrary number of arguments; Each urlsegment needs to be passed explicitly as its own argument.

tests/forms/FormSchemaTest.php
```diff
@@ -1,5 +1,6 @@
 <?php
 
+use SilverStripe\Dev\Debug;
```
flamerohr: Including `Debug` class when not used ;)

tractorcow:
Fixed @flamerohr

flamerohr:
Cool, this passes my review 👍

tractorcow:
I need to rebase.

tractorcow:
Rebased. :)

Control/Controller.php
sminnee: Seems like a reasonable addition.

tractorcow:
I've rebased; Will merge on green, since this has been reviewed and approved by the team.

PR #6120, opened 2016-09-30
Refactored react FormFields to use react-bootstrap FormControl components

flamerohr:
Also have done:

* Addded `CheckboxField`, `CheckboxSetField` and `OptionsetField` (radio) fields
* Refactored documentation for most components
* Split `CastStringToElement` to a library, since it was so handy across multiple components


chillu:
Have you tested `readOnly` and `disable` states? For example, `OptionSetField` converts to a `LookupField`, which isn't implemented in JS yet (I think...). `CheckboxField` converts to `CheckboxField_Readonly` which is more of a `LiteralField` on the client - but needs special handling.

flamerohr:
Yup, I had tested readonly and disabled states on the component itself rather than through a generated schema from SilverStripe.

My main reason is that since the react-bootstrap FormControls support readonly and disabled properties, where the equivalent HTML properties are not supported (e.g. readonly for checkboxes), it would make sense to utilise those if possible.

When I did try testing through a generated schema, I thought the readonly transformation was really weird when I was testing the transformations, e.g. `OptionsetField` (SingleSelect type) to `LookupField` (MultiSelect type) and causes some even more weird functions like this (although that function is still there for the casting to string reason mainly): https://github.com/silverstripe/silverstripe-framework/pull/6120/files#diff-a24423e2714380498612641b9adf1f0dR33

flamerohr:
I had also thought about if someone had chosen to override the react-bootstrap formfield component with their own, they can also decide how to handle readonly through their own component rather than be dictated by the server-side schema.

Up for discussion I guess, maybe this does deserve its own story.

chillu:
Regarding `Form->performReadonlyTransformation()`, that’s a bigger API change (not replacing field identity). I’ll agree that it’s bad, but will leave that alone for now - we need to get forms working as-is - which means implementing LookupField (https://github.com/silverstripe/silverstripe-framework/issues/6148). Not a blocker for this PR though.

flamerohr:
Cool, we'll need to implement the majority of the other fields soon anyway :)

admin/client/src/components/FieldHolder/OptionFieldHolder.js
```diff
@@ -0,0 +1,28 @@
+import fieldHolder from './FieldHolder';
+import SilverStripeComponent from 'lib/SilverStripeComponent';
+
+function optionFieldHolder(Field) {
```
flamerohr: Oops, this was a prototype for `CheckboxField.js`, which I thought had deleted/stopped from committing... Will remove it now.

admin/client/src/lib/CastStringToElement.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
```
flamerohr: hmm yeah, Git doesn't want to detect renamed cases to files... I will need to rebase the initial commit to fix this

admin/client/src/boot/BootInjector.js
```diff
@@ -0,0 +1,38 @@
...
+import PopoverField from 'components/PopoverField/PopoverField';
+import HeaderField from 'components/HeaderField/HeaderField';
+import LiteralField from 'components/LiteralField/LiteralField';
+import HtmlReadonlyField from 'components/HtmlReadonlyField/HtmlReadonlyField';
+import CompositeField from 'components/CompositeField/CompositeField';
+import Tabs from 'components/Tabs/Tabs';
+import TabItem from 'components/Tabs/TabItem';
+import FormAction from 'components/FormAction/FormAction';
+
+class BootInjector {
```
flamerohr: Done, added a comment for removing when dependency injection is in and left a comment that this was mainly to remove clutter from `boot/index.js`

admin/client/src/lib/CastStringToElement.js
chillu: Looks like this is done now

Issue #5187, opened 2016-03-16
Protection against CSRF attacks should be AJAX friendly


uniun:
Protection against CSRF attacks should be AJAX friendly, just like getValidationErrorResponse().
Now it breaks ajax requests if php session is expired.

dhensby:
What way do you want it to be nicer?

tractorcow:
If the PHP Session is expired, then the in-cms authentication should popup, and request the user to login again.

Once that's done, it's meant to re-populate all CSRF hidden inputs in the current page.

If your session is expiring, are you getting the CMS popup requesting you to re-enter your password?

dhensby:
I had thought this was more aimed at front end, not cms, ajax

tractorcow:
Hm, maybe the error condition isn't well catered for... you could be posting to an un-authenticated form, although the CSRF itself could fail. In which case, maybe the CMS re-authentication won't kick in properly. :)

I think in this case I'll need a more concise error reproduction process, if you are able to help, @uniun .

Which form did you post to, what url, what was the expected response, and what was the actual response? Were you expecting json and getting html, right?

uniun:
Sorry for the late response.

@dhensby is right, I was talking about frontend.

Sometimes session lifetime is very short and you cannot increase because you are not allowed to change it. Lets say the session lifetime is just about 10mins and you leave the page with a form for 15mins. Then do AJAX form submission.

Form checks for CSRF attacks and redirects you back instead of rendering a form: [Form.php#L380](https://github.com/silverstripe/silverstripe-framework/blob/3/forms/Form.php#L380).

But I think that it could check for an AJAX request and render a form, just like on [Form.php#L501](https://github.com/silverstripe/silverstripe-framework/blob/3/forms/Form.php#L501).

tractorcow:
Ok, my ideal fix for this is posted on another issue... I've responded to thousands and can't remember... take hard coded logic out of Form.php and put it into a CSRFFormField, and implement validate() logic there.

That way invalid forms will trigger normal formfield validation, rather than having custom handling that breaks cms randomly.

ReeceA:
@tractorcow is this still planned?

In other CMS (Invision Power Suite) for example. They would have a simple function: `$this->getCSRFtoken()` which would output the hidden field, then on the recipient side of things you could have a one-liner `$this->validateCSRFtoken()`

maxime-rainville:
Can someone confirmed if this is still a problem in SS4?

PR #6126, opened 2016-10-01
#5187 Move all framework Shortcode Providers to the same namespace.

Firesphere:


tractorcow:
This code looks fine to me!

Assets/File.php
```diff
@@ -234,92 +234,6 @@ public static function get_shortcodes() {
...
-	 * Replace "[file_link id=n]" shortcode with an anchor tag or link to the file.
-	 *
-	 * @param array $arguments Arguments passed to the parser
-	 * @param string $content Raw shortcode
-	 * @param ShortcodeParser $parser Parser
-	 * @param string $shortcode Name of shortcode used to register this handler
-	 * @param array $extra Extra arguments
-	 * @return string Result of the handled shortcode
-	 */
-	public static function handle_shortcode($arguments, $content, $parser, $shortcode, $extra = array()) {
```
tractorcow: Can you please add this to the 4.0.0 changelog under the appropriate section? (Probably under Filesystem API Additions / Changes).

Forms/HTMLEditor/FileShortcodeProvider.php
```diff
@@ -0,0 +1,128 @@
+<?php
+
+namespace SilverStripe\Forms\HtmlEditor;
```
sminnee: I don't think this is the right namespace – this shortcode handler should be part of the `SilverStripe\Assets` subsystem. If you didn't have `SilverStripe\Assets` you wouldn't have this shortcode provider available. This would be like putting the models of every SilverStripe module into the `SilverStripe\ORM` namespace.

Perhaps this can go into the `SilverStripe\Assets\ViewSupport` namespace, since it's a support class to assist the view?

Obviously you could argue: "if you want to use silverstripe\assets without the silverstripe WYSIWYG editor, then you've got the same problem in reverse". However, I think that we're more likely to require the WYSIWYG editor without assets than the reverse.

Assets/File.php
Firesphere: Yep, will do.

Forms/HTMLEditor/FileShortcodeProvider.php
Firesphere: I went with this namespace, as the previous extraction, the EmbedShortcodeProvider, was decided to go into this namespace.

I'm happy to change it, but I'd like to have them all bundled?

My 2 cents on the namespace: `SilverStripe\View\Provider` or `SilverStripe\View\Shortcode` (Or something among those lines). as they are part of the View, not Forms or Assets as such.

Firesphere:
I'll be adding commits, so it's easier to track the changes, until it's a go, then squash it. That okay with you folks?

Forms/HTMLEditor/FileShortcodeProvider.php
Firesphere: Update on previous comment:
Wouldn't it be more logic to put all the Shortcode Parsers under the same namespace? As the interface is defined under `SilverStripe\View\Parsers`, I would suggest moving all providers to the same namespace?

Assets/File.php
Firesphere: Added changes to the changelog. Including instructions on how to migrate.

Forms/HTMLEditor/FileShortcodeProvider.php
Firesphere: I feel slightly uncomfortable, moving the providers to separate namespaces, as the embed provider would stay in the Form\HtmlEditor?
I've for now moved them all into `SilverStripe\View\Parsers`. Welcome any thoughts? Happy to move them around, but I'd personally prefer to keep all providers in the same namespace, so CMS Providers (like sitetree_link), can go into `SilverStripe\CMS\Chosen\Namespace` as well, to keep it all in a unified place.

tractorcow:
@sminnee I don't mind the new namespace. Is `SilverStripe\View\Parsers` ok with you? If so I'll merge.

Firesphere:
Have rebased and squashed into a single commit. Separately, the tests are fine. Unsure of anything I rebased on broke other tests though.

Firesphere:
@tractorcow It's not a "new" namespace. It's the same namespace as where the ShortcodeParser Interface resides.

sminnee:
My concern here is that this makes the assets module a dependency of the view module, which doesn't make sense to me.

Firesphere:
tl;dr: It makes OembedProvider a bit a hovering orphan that needs a good namespace.

I understand and also partially agree with your concern @sminnee.

What I however wish to accomplish, is unity. Hence also the move of the Oembed provider to the same namespace.

If we put the File and Image providers in Assets\ViewSupport, I think the Oembed one, should also move to a similar namespace instead of HTMLEditor\Form.

I'm happy with the ViewSupport namespace, but unclear on how to make it consistent across all providers (same goes for CMS, which would probably then go to CMS\Model\ViewSupport?)

dhensby:
I'm a bit split on this but I'm learning on agreement with sam here. We should be placing parsers in namespaces relevant to the parser, not lumping them together in one.

Firesphere:
I can live with that @dhensby, I'm trying to find a good concensus, where the View was my first thought on unifying.

I'm just still stuck on where the Oembed provider should go. As having a single `Oembed\ViewSupport` namespace for it, sounds silly to me.

sminnee:
You can leave the Oembed provider where it is if you want.

Firesphere:
To unify, Do you mind if I make the Oembed `SilverStripe\View\ViewSupport` and create a folder for it like that?
I want the end namespace to be as uniform as possible, to avoid confusion.

dhensby:
@Firesphere just do it :) Let's get this merged

Firesphere:
Agreed and fixed your change request @sminnee

dhensby:
needs rebasing

Firesphere:
Will do when I have time tomorrow or this weekend :)

Firesphere:
I completely lost sight of this PR, have rebased but it seems there are still merge conflicts?

chillu:
@Firesphere Any progress here?

unclecheese:
Closing. Reopened as:
* https://github.com/silverstripe/silverstripe-assets/pull/40
* https://github.com/silverstripe/silverstripe-framework/pull/7085

PR #6130, opened 2016-10-03
Allow filtering if a relation is defined or a formatting

lekoala:


dhensby:
Please can we have a test for this? Just a fairly simple test of passing in a dot notation that would previously break would be great :)

lekoala:
is there any test related to the GridFieldFilterHeader class that I can use? Because I don't find any

dhensby:
It doesn't look like it, but you could create it and just test this aspect :)

dhensby:
Right, I've refactored this slightly (to make it testable) and added a test... let's hope it goes green

tests/forms/gridfield/GridFieldFilterHeaderTest.php
```diff
@@ -0,0 +1,27 @@
...
+	protected $extraDataObjects = array(
+		'GridFieldFilterHeaderTest_DataObject',
+	);
+
+	public function testColumnToFilterField() {
+		$class = 'GridFieldFilterHeaderTest_DataObject';
+		$header = new GridFieldFilterHeader();
+		$method = new ReflectionMethod($header, 'columnToFilterField');
+		$method->setAccessible(true);
+		$this->assertEquals('Title', $method->invoke($header, $class,'Title.ATT'));
```
tractorcow: Can we add a test to see how it accepts more than one level deep? E.g. `Parent.Image.LastEdited.Nice`?

tractorcow: I have a sneaky suspicion it's going to break. ;)

dhensby: It could be an accepted limitation? it'll work better than it does now and I'm not going to write a test asserting that should fail ;)

dhensby: Though we do need to test it isn't broken in this instance...

dhensby: I think this will be fine, actually, if the first part is a relation it just returns the whole thing (current behaviour).

dhensby:
@tractorcow added a new assertion

PR #6137, opened 2016-10-04
Have a clear error message

lekoala:
Because it's really annoying not knowing which field causes the error


dhensby:
personally I'm happy to add warnings in patch releases as I don't think it constitutes any kind of API change.

kinglozzer:
Given that it would (presumably) result in a “call to member function on non object” error prior to this change, I don’t see any problem with showing a more useful message. Fatal error -> more helpful fatal error isn’t an API change. I’d rather see an exception than `user_error()`, though...

dhensby:
Good point. I'm not a fan of an exception here and I think a warning is nice as it lets people continue.

PR #6152, opened 2016-10-05
Added development watch command to package.json

flamerohr:
I've created a "dev" config for webpack, for development use only.

This work is a start to improve build time for development, and cuts build time while watching from average 6500ms to average 2500ms.

This does run the risk of developers committing and pushing this development build, which we can setup checks for.
Also a small risk of different behaviours between this development build and the production build, not common or likely but it's difficult to anticipate regressions from Uglify and Source-maps.

I do believe that this is an important step for improving contributor/developer experience, as currently some drawbacks with the current build are:
* Long build time, so if you're making rapid change+saves, it's possible that the last save does not trigger another build so you're missing changes in the build.
* Refreshing the browser too early and you don't get the new build.
* A combination of the two above leads to frustrating: emptying cache, reloading and restarting watch cycle.
* Source maps built for production are redundant because they are in `.gitignore`, only beneficial for development.
* resolve.root performance: https://github.com/webpack/webpack/issues/472#issuecomment-55706013 using resolve.modulesDirectory is slower and not necessary for our setup.

Things I'm looking into for improving this further:
* Dev-server hot-loading new changes. [This gist](https://gist.github.com/MrOrz/4f3da17fd73dbb4c7668) and [this post](http://stackoverflow.com/questions/33233785/installing-webpack-dev-server-with-php-application)
* Possibly a manifest file to separate development and production streams, so it reduces the risk of committing the development build... there are limitations (reference: https://github.com/webpack/webpack/issues/144 )
* Possibly look at upgrading webpack version for its tree-shaking, etc... features
* Looking into [HappyPack](https://github.com/amireh/happypack)

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
   },
   "scripts": {
     "build": "webpack",
+    "watch": "webpack --config ./webpack-dev.config.js --watch",
     "css": "WEBPACK_CHILD=css webpack",
     "lock": "npm-shrinkwrap --dev",
     "test": "NODE_PATH=\"./client/src:./admin/client/src\" jest",
--- a/webpack-dev.config.js
+++ b/webpack-dev.config.js
@@ -1,0 +1,25 @@
+const webpack = require('webpack');
+const Config = require('./webpack.config');
+
+if (Array.isArray(Config)) {
+  const jsConfig = Config.find((item) => item.name === 'js');
+
+  jsConfig.plugins = [
+    new webpack.ProvidePlugin({
+      jQuery: 'jQuery',
+      $: 'jQuery',
+    }),
+    // Most vendor libs are loaded directly into the 'vendor' bundle (through require() calls in vendor.js).
+    // This ensures that any further require() calls in other bundles aren't duplicating libs.
+    new webpack.optimize.CommonsChunkPlugin({
+      name: 'vendor',
+      minChunks: Infinity,
+    }),
+  ];
+
+  for (var i = 0; i < Config.length; i++) {
+    Config[i].devtool = 'source-map';
+  }
+}
+
+module.exports = Config;
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,6 +1,7 @@
 const webpack = require('webpack');
 const autoprefixer = require('autoprefixer');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
+const path = require('path');
 // const SprityWebpackPlugin = require('sprity-webpack-plugin');
 
 const PATHS = {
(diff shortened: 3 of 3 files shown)


anselmdk:
If you haven't already, I certainly recommend looking into Webpack hot reloading. This is what in my opinion really makes it shine. Hot reloading styles should be easier than javascript, but I've previously worked on a React project where Webpack hot reloaded the entire app, getting back to other projects  felt like being thrown back 10 years...

tractorcow:
@anselmdk we'll have a look into it. :)

flamerohr:
Hackday project to add hotloading with proxy settings! :D
Yes, hotloading is part of my todo list in my original post.

chillu:
Awesome change, sorry for overlooking this!

> This does run the risk of developers committing and pushing this development build, which we can setup checks for.

I've already fallen into this trap a few hours after it was merged. It's going to be a frequent cause of frustration I think, particularly since those erroreous changes will be hidden in Github diffs (too large to view). If nothing else, it'll grow the size of our repo because the dist files will go back and forth between different formats. I would've expected only the last line to change (reference to sourcemap), but for some reason the whole file has changed. Any idea why that's the case? I'm not too fussed about a sourcemap reference appearing in our bundle files every now and then, but multiple 100kb diff on accidental commits isn't good.

Also, we've now created an inconsistent experience with `cms` and `asset-admin` - are you going to get those aligned as well?

sminnee:
> Also, we've now created an inconsistent experience with cms and asset-admin - are you going to get those aligned as well?

Let's iron the kinks out of the framework implementation first.

What do we need to do to address @chillu's issue? Generating dev files to .dev.js and patching requirements to include these files instead would be an approach.

chillu:
The problem is that on framework, as a developer you're running `npm run watch`, but on cms and asset-admin you run `npm run build -- --watch`. Maybe it's too early to port over the rest of the approach, but we need a) consistent commands and b) tickets for follow up work. We also need the docs updated with this approach.

I think if we're generating a separate `bundle.dev.js`, we need to have a CI task check if the `bundle.js` is outdated by re-building it. Ideally we can just keep `bundle.js` the same, even leave the `//# sourceMappingURL=bundle.js.map` at the bottom - regardless if its generated with sourcemaps - browsers with dev tools open will simply fail to load these. But we have to sort out why its generating entirely different file structures with sourcemaps enabled first.

flamerohr:
Yes, I'm planning to pull this over to cms/asset-admin/frameworktest when it's a bit more mature, in retrospect, I should've put my notes in an open issue rather than pull request.

I'm not sure why you're getting a differently generated file because sourcemaps weren't enabled? Are you using `webpack -p` or just `npm run build`? I have noticed some builds being pushed up uses the `-p` switch and some do not.
e.g. https://github.com/silverstripe/silverstripe-framework/pull/6180 only changed the sourcemap comment at the bottom (although I'm not sure what is happening with the separate legacy files)

flamerohr:
On the note of a separate output file, I had considered this option, this also present issues in of itself such as:
* Desynchronised files, dev and prod files can be different, I think this point is acceptable until you have to push then it becomes an issue.
* Extra logic on the server-side somewhere to work out which of the two sets of files to use. (e.g. manifests file)
* Build on every branch checked out.

I'm still playing around with this idea a bit more, I did get a nice setup on one of my projects before but it was a lot more simpler ;) and I don't quite see it suitable for a framework implementation.

Issue #6012, opened 2016-09-15
Remove JLayout and replace with flex layout PART 1


clarkepaul:
Acceptance criteria for all parts:

 * Opportunities to split this card have been identified ASAP
 * JLayout is removed
 * HTML is updated to suit flexbox
 * Key layout use-cases are functioning:
   * Sticky-collapsible side menu
   * Sticky-collapsible tree
   * Auto-collapsing preview based on width
   * Bottom buttons fixed to bottom of window
 * Layout regressions logged as tickets

----
PART 1 or 2
 * [x] Remove JLayout
 * [x] Main menu
 * [x] Site tree
 *       Page history
  * [x] DEV js .toggle() of table rows changed to .style.display = 'table-row'  (see description  #below...)
  * [x] CSS layout
 * [x] Page settings
 *      Add page screen
  * [x] DEV Broken add page action 
  * [x] CSS layout
 * [x] Filter/Search dropdown
 * [x] Settings
 * [x] Profile
 * [x] Reports
 * [x] Security
 * [x] Files area
 * [x] Campaigns

IGNORES:
Page preview, see https://github.com/silverstripe/silverstripe-framework/issues/6154
Responsiveness, these are all desktop styles by default, see https://github.com/silverstripe/silverstripe-framework/issues/6164

Before Flex can be tried out properly JLayout needs be removed, there are still many bugs for IE so I've created a demo for how flex could work with the CMS panels. Before implementation refer to Bootstrap docs to ensure existing layout classes don't already exist.

Demo: http://codepen.io/clarkepaul/pen/LRNLgG

chillu:
* Most of the legacy layout logic is contained in https://github.com/silverstripe/silverstripe-framework/blob/master/admin/client/src/legacy/LeftAndMain.Layout.js, and can be removed.
 * The panel toggle/collapse can be implemented via changed classes without JS applying dimensions to DOM elements, but user preferences will need to be restored from 
 * `LayoutOptions.minContentWidth` and `LayoutOptions.minPreviewWidth` in `LeftAndMain.js` probably need to stay in JS, since we can't apply media queries to specific DOM elements (just the whole browser window). We should be able to simplify the dimension calculations though, by applying classes like `.with-preview` to the flexbox container rather than setting dimensions directly via JS.
[...]

PR #6155, opened 2016-10-05
Part 1 of 5, remove JLayout and apply flexbox to pages section (excludes preview)

clarkepaul:
See https://github.com/silverstripe/silverstripe-framework/issues/6012
Requires https://github.com/silverstripe/silverstripe-cms/pull/1638


tractorcow:
Hm, it's not looking good.

image

PR #6157, opened 2016-10-06
Make __call() clearer that a method may be non-public

### LINKED ISSUE #6151: Making a method used in template private leads to inconsistent messaging
Affects latest 3.4, unsure about other versions.

If you specify:
```php
private function Form() {
    return new Form($this, __FUNCTION__, new FieldList(), new FieldList());
}
```

Then call `$Form` in your template, you get "[User Error] Uncaught Exception: Object->__call(): the method 'form' does not exist on 'Flowchart'".

My assumption is we're using reflection somewhere but not checking the visibility property properly before attempting to call the method. Even just updating the error message to say the following would be better: "the method 'form' does not exist (or is protected/private) on the 'Flowchart' class".

Will try and PR for this when I get some spare time :)

madmatt:
Resolves #6151.

### DIFF
--- a/core/Object.php
+++ b/core/Object.php
@@ -777,7 +777,7 @@ abstract class Object {
 		} else {
 			// Please do not change the exception code number below.
 			$class = get_class($this);
-			throw new Exception("Object->__call(): the method '$method' does not exist on '$class'", 2175);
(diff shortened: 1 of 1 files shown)


madmatt:
Check comments on #6151 before merging, there's some discussion there about alternatives to this.

PR #6161, opened 2016-10-06
API Convert fieldSpec options to bitwise operators

### LINKED ISSUE #6159: Use a bitmask for DataObjectSchema::fieldSpecs() options
The options argument for DataObjectSchema::fieldSpecs() is currently an array that looks for specific values. A bitmask composed of DataObjectSchema class constants would be a less error-prone mechanism for this (misspellings throw compile-time errors) and the class constants provide hooks to attach docblocks to.

@kinglozzer and I think this would be the better API. Unless there are objections, can we implement this?

This is a follow-on task from https://github.com/silverstripe/silverstripe-framework/pull/6114

tractorcow:
Fixes #6159

No CMS PR necessary for this one. ;)

### DIFF
--- a/Forms/GridField/GridFieldSortableHeader.php
+++ b/Forms/GridField/GridFieldSortableHeader.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\Forms\GridField;
 
 use SilverStripe\Forms\LiteralField;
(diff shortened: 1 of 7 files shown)


kinglozzer:
❤️

PR #6167, opened 2016-10-11
BUG Fix installer for 4.0

### LINKED ISSUE #6084: not able to install ss4 alpha2 on mamp
trying to install ss4 alpha2 isn't working on mamp because of warning about database credentials (although they are correct). Screenshots attached.
schermafbeelding 2016-09-26 om 09 35 27
schermafbeelding 2016-09-26 om 09 38 24

tractorcow:
Fixes #6084

There were many many issues to resolve in installer. :)

 - [ ] cms https://github.com/silverstripe/silverstripe-cms/pull/1644
 - [ ] postgres https://github.com/silverstripe/silverstripe-postgresql/pull/58
 - [ ] mssql https://github.com/silverstripe/silverstripe-mssql/pull/40
 - [ ] sqlite https://github.com/silverstripe/silverstripe-sqlite3/pull/29

### DIFF
--- a/Dev/Install/config-form.html
+++ b/Dev/Install/config-form.html
@@ -6,7 +6,7 @@
 	<head>
 		<title>SilverStripe CMS / Framework Installation</title>
 		<meta http-equiv="Content-type" content="text/html; charset=utf-8">
-		<script type="application/javascript" src="<?php echo FRAMEWORK_NAME; ?>/admin/thirdparty/jquery/jquery.js"></script>
+		<script type="application/javascript" src="http://code.jquery.com/jquery-1.7.2.min.js"></script>
 		<script type="application/javascript" src="<?php echo FRAMEWORK_NAME; ?>/Dev/Install/client/dist/js/install.js"></script>
 		<link rel="stylesheet" type="text/css" href="<?php echo FRAMEWORK_NAME; ?>/Dev/Install/client/dist/styles/install.css">
 		<link rel="shortcut icon" href="favicon.ico">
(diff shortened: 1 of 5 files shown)


Dev/Install/install.php5
```diff
@@ -816,23 +827,31 @@ class InstallRequirements {
 		$this->testing($testDetails);
 		if(!class_exists($className)) {
 			$this->error($testDetails);
+			return false;
 		} else {
 			return false;
```
tractorcow: Ah, it should be true in that else. Fixed.

tractorcow:
This has been updated and ready to be merged.

Dev/Install/config-form.html
```diff
@@ -6,7 +6,7 @@
 	<head>
 		<title>SilverStripe CMS / Framework Installation</title>
 		<meta http-equiv="Content-type" content="text/html; charset=utf-8">
-		<script type="application/javascript" src="<?php echo FRAMEWORK_NAME; ?>/thirdparty/jquery/jquery.js"></script>
+		<script type="application/javascript" src="<?php echo FRAMEWORK_NAME; ?>/admin/thirdparty/jquery/jquery.js"></script>
```
sminnee: Best bet is to refer to a CDN copy of jQuery: `http://code.jquery.com/jquery-1.7.2.min.js`. Otherwise this will break when we split out admin. It's also an inappropriate coupling of the installer to the admin.

flamerohr: Done, using a CDN loses the ability to use the Installer properly if the user goes offline between downloading SilverStripe and using the installer, which I would imagine isn't common,

PR #6168, opened 2016-10-11
BUG Fix issue with SS_List as datasource for dropdown field

### LINKED ISSUE #6166: CheckboxSetField no longer accepts DataLists as a source argument
Upgrading from 3.1 -> 3.4 I found that adding a `CheckboxSetField` to the CMS to represent a relationship no longer validates when passing a datalist and instead an array is needed.

eg:

```php
CheckboxSetField::create('HasManyRelationName', 'Field Label', RelatedObject::get());
```

In the CMS this will return a validation error of "Please select a value from the list provided. 'RelatedObject and RelatedObject [repeats]' is not a valid option"

tractorcow:
BUG Fix validation issue with CheckboxSetField
Fixes #6166

@dhensby hope this solves all the issues. :)

### DIFF
--- a/forms/CheckboxSetField.php
+++ b/forms/CheckboxSetField.php
@@ -333,21 +333,22 @@ class CheckboxSetField extends OptionsetField {
 			return true;
 		}
 		$sourceArray = $this->getSourceAsArray();
+		$validValues = array_keys($sourceArray);
 		if (is_array($values)) {
-			if (!array_intersect_key($sourceArray, $values)) {
+			if (!array_intersect($validValues, $values)) {
 				$validator->validationError(
 					$this->name,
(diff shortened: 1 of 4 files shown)


tractorcow:
Also removes rule that values need to be a map where keys === values. Now it just checks values (so you can have a non-associative array of valid values).

dhensby:
Looks nice. I'll test it out and merge if it works (And if tests pass)

dhensby:
I've tested locally and it's all good.

You've used short array syntax, which has broken the build - I'm merging by hand - just running tests then will merge

tractorcow:
Darn it, I've been doing too much 4.0!

PR #6169, opened 2016-10-11
API Duplication of many_many relationships is now opt in

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1453

It also works properly on unsaved records.

### DIFF
--- a/docs/en/04_Changelogs/4.0.0.md
+++ b/docs/en/04_Changelogs/4.0.0.md
@@ -1317,6 +1317,8 @@ A very small number of methods were chosen for deprecation, and will be removed 
 * `DataObject::db` now returns composite fields.
 * `DataObject::ClassName` field has been refactored into a `DBClassName` type field.
 * `DataObject::can` has new method signature with `$context` parameter.
+* `DataObject::duplicate` Now requires explicit flag to duplicate belongs_many_many (off by default),
+  but now works with unsaved relations. By default only many_many are duplicated.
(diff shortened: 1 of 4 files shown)


tractorcow:
I've rebased this. Anyone care to review? @dhensby ? :P

tractorcow:
I'm going to revert default to true, to keep from breaking duplication of nested objects in CMS.

tractorcow:
I need to update this PR, block this for now. Don't merge yet!

kinglozzer:
Welp, Github doesn’t let you remove reviews. What needs fixing @tractorcow? I still can’t spot it 😳

tractorcow:
I've changed the behaviour in a way that fixes the original issue, but doesn't completely disable many_many duplication by default.

The issue was with belongs_many_many being duplicated; I've changed it so that it duplicates only many_many by default (but this can be changed).

@dhensby how is this as an alternative? I'll write some unit tests if you like this.

tractorcow:
Ok, tests added. Now I'm happy this can be merged. :)

It does need re-approval of the general logic, however.

tractorcow:
@kinglozzer the use case @dhensby raised earlier was a page having many_many images, and a duplicate suddenly losing all those images. :D

dhensby:
OK - I this makes sense to me

tractorcow:
Fixed linting issue

tractorcow:
Green and can merge. :)

PR #6171, opened 2016-10-11
API Versioned::publishRecursive() now uses a ChangeSet

### LINKED ISSUE #5667: Convert "simple" publishing to use changesets in the backend
As a dev, I don't want duplicate code that does the same thing in different ways, so that I can avoid bugs

Acceptance criteria:

 - Every publication event in the system uses the same / a similar API
 - Simple publications create an "inferred" changeset
 - Inferred changesets have a IsInferred boolean or Type enum to identify them
 - Inferred changesets should appear in the changeset history
 
PR at https://github.com/silverstripe/silverstripe-framework/pull/6171

Excluded:

 - Changeset history pagination

Tasks:

 - [x] Move unlinkDisownedObjects from Versioned::publish() to ChangeSet::publish()
 - [x] change Versioned::publish() to use ChangeSets
 - [x] Ensure auto-generated campaigns are flagged adequately.

---

Right now there are two publish flows. Simple publishing should create a Changeset & immediately publish it so there is only one.

tractorcow:
API Add IsInferred to inferred changesets
API Add SapphireTest::assertNotDOSContains
Fixes #5667


Forms/ReadonlyField.php
```diff
@@ -61,6 +61,12 @@ public function castingHelper($field) {
 		return parent::castingHelper($field);
 	}
 
+	public function getSchemaStateDefaults() {
+		$values = parent::getSchemaStateDefaults();
+		$values['value'] = $this->dataValue();
+		return $values;
```
flamerohr: Maybe leave a comment along the lines of "so we do not end up with `<i>('none')</i>` in the value.

tests/model/VersionedOwnershipTest.php
```diff
@@ -284,6 +287,8 @@ public function testRecursivePublish() {
 		$this->assertDOSEquals($oldLiveBanners, $parentLive->Banners());
 
 		// On publishing of owner, all children should now be updated
+		$this->sleep(0); // Force hard mock of now
+		$now = DBDatetime::now()->Nice();
```
flamerohr: Maybe using `DBDatetime::set_mock_now()` would be a viable option here?

tractorcow: Yeah that's what sleep does but I was lazy. :)

flamerohr:
Looks good to me, straight forward and clean.
Merge on green 👍

tractorcow:
I've fixed an issue due to nested database transactions; I've substituted a ->publishRecursive() with an ancestry walk and ->publish() in File::onBeforePublish().

Issue #5524, opened 2016-05-11
Implement a React form library 


chillu:
PRs:

 * https://github.com/silverstripe/silverstripe-framework/pull/6172
 * https://github.com/silverstripe/silverstripe-asset-admin/pull/287

Acceptance criteria

 * Any forms using FormBuilder will use the new library
 * Form functionality in campaigns and assets are operational using the new library
 * Form data can be loaded, edited, saved
 * Negligible performance regression, e.g. during typing
 * High-level use of the new APIs documented (it's fine to cross-link to other people's docs)

Excluded

 * Validation

----

Some options: https://react.rocks/tag/Form and https://react.rocks/tag/FormGenerator.

 * Compatible with Redux state management
 * Integrates well with the existing `<FormBuilder>` components and the SS form schema
 * Supports creation of custom form fields through `<FormBuilder>`
 * Stores form state separately from client-side models (Redux state tree)
 * Can optionally "bind" form state to client-side models
 * Allows customisation of form behaviour through JS (most likely a layer we need to add on top of a library)
 * Has performant input processing on complex reducer structures and large forms (this might be a blocker for redux forms, see [article about it's performance issues](http://www.position-absolute.com/javascript/my-controlled-input-dilemma-with-react-redux/)). Update: Known [workaround](https://github.com/erikras/redux-form/issues/454#issuecomment-166448837) is a onChange debounce with `setState` on the component.

Legacy cross-ref (for SilverStripe internal only, sorry): https://silverstripe.atlassian.net/browse/OSS-1804

## Tasks

 * [x] Update react to v15.0
 * [x] Pull in Redux Form dependency
 * [x] Set up Redux Form reducer
 * [x] `FormBuilder` to create `reduxForm` with `props.fields` required for Redux Form
 * [x] Set unique form name for Redux Reducer (based on "action" attr?)
 * [x] Set `initialValues` from form schema data
 * [x] Use `onSubmit` (to give form a chance to handle validation)
 * [x] Set button loading state based on `submitting` redux state from React forms (how to determine which button?)
 * [x] ~~Override `getFormState()` with `deep-freeze` behaviour~~ redux-form prefers ImmutableJS, but that's a larger library choice which we shouldn't just introduce here. Given that we've reduced the redraws, it's no longer a high priority for this use case.
[...]

PR #6172, opened 2016-10-12
Use redux-form

chillu:
See https://github.com/silverstripe/silverstripe-framework/issues/5524
Merge with https://github.com/silverstripe/silverstripe-asset-admin/pull/287


admin/client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -250,31 +200,31 @@ export class FormBuilderComponent extends SilverStripeComponent {
...
       .reduce((prev, curr) => {
-        const match = this.findField(fields, curr.id);
+        const match = this.findField(schema.fields, curr.id);
+
         if (!match) {
           return prev;
         }
 
+        // Skip non-data fields
+        if (match.type === 'Structural') {
```
flamerohr: This will affect fields like `HtmlReadonlyField` where it's structural, but does have a value.
Perhaps that field does need a refactor.

admin/client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -373,23 +328,24 @@ export class FormBuilderComponent extends SilverStripeComponent {
...
-      // Events
-      const extraProps = {
-        handleClick: this.handleAction,
-        loading,
-        disabled: loading || action.disabled,
-      };
-
-      // Build child nodes
+      let props = action;
+      props.handleClick = this.handleAction;
```
flamerohr: Could move this and the `if (this.state.submittingAction === action.name) {` below to an `else` statement for `if (action.children) {`.
When reading this code block, it comes as very weird to set the props just to watch them disappear immediately after.

admin/client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -416,32 +372,54 @@ export class FormBuilderComponent extends SilverStripeComponent {
...
+        // Overlap with redux-form prop handling : createFieldProps filters out the 'component' key
+        { schemaComponent: action.component }
+      );
 
-      if (field.children) {
-        return Object.assign({}, data, {
-          children: this.getFieldData(field.children, formState),
-        });
+      if (action.children) {
+        data.children = this.normalizeFields(action.children);
```
flamerohr: Call `normalizeActions(action.children)` instead?

sminnee:
Hehe, CodeCov has pinged us on the lack of unit tests in the PHP files you touched. ;-)

admin/client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -93,154 +87,110 @@ export class FormBuilderComponent extends SilverStripeComponent {
...
+    // Allow custom handlers to cancel event
+    if (!event.isPropagationStopped()) {
+      // Perform state update async to avoid component re-render *during* event bubbling,
+      // which causes the <form> event target to change, and the browser to abort bubbling.
+      // Needs to be shorter than the expected server response to avoid edge cases.
+      this.submittingAction = name;
+      this.submittingSetTimeout = setTimeout(() => this.setState({ submittingAction: name }), 50);
+    }
+
+    return true;
```
chillu: Yeah it's not needed, ancient assumptions about how event handlers used to work (maybe in jQuery?)

flamerohr: Understandable, jQuery had an ancient convention where `return false` = `event.preventDefault() && event.stopPropagation()` in event handlers

admin/client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -250,31 +197,31 @@ export class FormBuilderComponent extends SilverStripeComponent {
...
       .reduce((prev, curr) => {
-        const match = this.findField(fields, curr.id);
+        const match = this.findField(schema.fields, curr.id);
+
         if (!match) {
           return prev;
         }
 
+        // Skip non-data fields
+        if (match.type === 'Structural' && typeof curr.value === 'undefined') {
```
chillu: Checking `curr.value` is pretty hacky - we need a way to either distinguish "structural" from "readonly" fields, or find a different way to populate readonly field values (`props.value` vs. `props.data.content`)

flamerohr: Have a look at `LiteralField` it uses `props.data.content`, so HtmlReadonlyField could follow suit (and we could simplify the React component when we convert it :D )

admin/code/LeftAndMain.php
```diff
@@ -355,10 +356,12 @@ protected function getSchemaResponse($form) {
 	 * based on the request data.
 	 *
 	 * @param Form $form
+	 * @param String $id Optional, will default to the current request URL
 	 * @return array
 	 */
-	protected function getSchemaForForm(Form $form) {
+	protected function getSchemaForForm(Form $form, $id = null) {
 		$request = $this->getRequest();
+		$id = $id ? $id : $request->getURL();
```
flamerohr: I was having some reservations about using the request URL for the "id" of the schema, perhaps `$id` could be mandatory.
`$id` may get removed in the future, so maybe not...
What are your thoughts about this? @chillu

sminnee: Best case is that the same schema is downloaded for every record of a given type, and that dynamic behaviour is done within JavaScript. Not every form will be able to be built this way, but it will be a use-case that will be applied sometimes.

This doesn't preclude us from making id mandatory - it seems like it's usually set to a URL regardless.

sminnee:
OK I've had a review of this and I'm comfortable with the dialog between @chillu and @flamerohr, assuming that both parties are happy with the final changes.

Does this need any squashing before the merge?

flamerohr:
It'll be good to get feedback on the last three comments not replied to, but they're more about opinions than required changes, not blockers from merging.

I don't think this needs anymore squashing and can go ahead with merging :)

admin/client/src/components/FormBuilder/FormBuilder.js
chillu: I've changed this to `if (match.type === 'Structural' && typeof curr.value === 'undefined')` and started another discussion thread about it on this PR

admin/client/src/components/FormBuilder/FormBuilder.js
chillu: Ah yep, changed from `normalizeFields()` to `normalizeActions()` now

admin/client/src/components/FormBuilder/FormBuilder.js
chillu: I'm consistently using `props.value` for all fields now (incl. `LiteralField` and `HTMLReadonlyField`), and modified the check above to `(field.type === 'Structural' || field.readOnly === true)`. This removes the one-off case for `props.data.content` in `LiteralField`.

admin/client/src/containers/FormBuilderLoader/FormBuilderLoader.js
```diff
@@ -0,0 +1,106 @@
...
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+import fetch from 'isomorphic-fetch';
+import { Field as ReduxFormField, reduxForm } from 'redux-form';
+import * as schemaActions from 'state/schema/SchemaActions';
+import Form from 'components/Form/Form';
+import FormBuilder, { basePropTypes, schemaPropType } from 'components/FormBuilder/FormBuilder';
+
+import es6promise from 'es6-promise';
+es6promise.polyfill();
```
chillu: Done

admin/client/src/containers/FormBuilderLoader/README.md
```diff
@@ -0,0 +1,12 @@
+# FormBuilderLoader Component
+
+Used to generate forms, made up of field components and actions, from form schema data.
```
chillu: Changed

chillu:
OK I think I've responded to all of @flamerohr excellent suggestions :) Thanks for fixing up a few of them directly Chris! I've squashed as part of a master rebase. The only outstanding issue is what we should do with schema `id` (see discussion above), but I don't think that needs to block merge.

admin/code/LeftAndMain.php
flamerohr: I like the idea of `Form->Link()`, it's a nice middle ground. Looking at the FormFactory work that was just merged yesterday, this may not work out when Forms are context agnostic.

`schema.action` almost looks to be the same as `Form->Link()`, maybe I'm missing a distinction between the two.

We can pull this discussion/brainstorming out to a separate issue, I agree that it doesn't need to block merge!

Forms/Schema/FormSchema.php
```diff
@@ -26,7 +26,6 @@ public function getSchema(Form $form, $schemaLink) {
 			'id' => $form->FormName(),
 			'action' => $form->FormAction(),
 			'method' => $form->FormMethod(),
-			'schema_url' => $schemaLink,
```
tractorcow: Can we remove $schemaLink from the method argument then?

admin/code/LeftAndMain.php
tractorcow: $form->Link() won't be any more or less distinct that schema link will it?

PR #6173, opened 2016-10-12
Check which table should be used for leftJoin

lekoala:
In the current situation, the join may not be made on the right table in case of a DataObject extending another DataObject. This change use table_for_object_field to ensure that we use the right table for the given field.

### DIFF
--- a/forms/gridfield/GridFieldSortableHeader.php
+++ b/forms/gridfield/GridFieldSortableHeader.php
@@ -233,9 +233,12 @@ class GridFieldSortableHeader implements GridField_HTMLProvider, GridField_DataM
 					// been aliased. In that event, assume what the user has
 					// provided is the correct value
 					if(!$joinClass) $joinClass = $lastAlias;
+					
+					// Check on which table is the sort column
+					$table = ClassInfo::table_for_object_field($tmpItem->class, $parts[sizeof($parts)-1]);
 
 					$dataList = $dataList->leftJoin(
-						$tmpItem->class,
+						$table,
 						'"' . $methodName . '"."ID" = "' . $joinClass . '"."' . $methodName . 'ID"',
 						$methodName
 					);
--- a/forms/gridfield/GridFieldSortableHeader.php
+++ b/forms/gridfield/GridFieldSortableHeader.php
@@ -236,7 +236,11 @@ class GridFieldSortableHeader implements GridField_HTMLProvider, GridField_DataM
 					
 					// Check on which table is the sort column
 					$table = ClassInfo::table_for_object_field($tmpItem->class, $parts[sizeof($parts)-1]);
-
+					// In case the method doesn't return any proper table, use item class as before
+					if(!$table || $table == 'DataObject') {
+						$table = $tmpItem->class;
+					}
+					
 					$dataList = $dataList->leftJoin(
 						$table,
 						'"' . $methodName . '"."ID" = "' . $joinClass . '"."' . $methodName . 'ID"',


lekoala:
as a sidenote of this pull request, I'm very surprised that ClassInfo::table_for_object_field returned "DataObject" as a table (therefore, my other commit).  Maybe there is something else that should be done than checking what's being returned.

dhensby:
Can you provide a more full issue description for what's happening, how to reproduce and how this prevents the issue?

lekoala:
Ok, sample situation:

I have MyMember extending Member
on MyMember, I have the field MobilePhone

Then, I have a Company, with an Admin. Summary fields are these:

```
class MyMember extending Member {
$db = [
'MobilePhone' => 'Varchar'
];
}
class Company {
$has_one = ['Admin' => 'Member'];
$summary_fields=  [
'Title' => 'Company Name', 'Admin.Email' => 'Email', 'Admin.MobilePhone' => 'Mobile Phone'
];
```

Great. Now, I have a GridField with a list of Companies. In can try to sort by Admin.Email and Admin.MobilePhone.

Current situation:
Sorting by MobilePhone works (it makes a leftJoin on MyMember and add the where clause)
Sorting by Email does NOT work (it makes a leftJoin on MyMember that doesn't contain the field)

With my fix, both works (MobilePhone makes a leftJoin on MyMember and Email makes a leftJoin on Member).

As for the last commit, see my comment above

dhensby:
In your example a `Member` doesn't have a `MobileNumber` so sorting on it is somewhat impossible?

I'm not sure why it would be joining on `MyMember` at all given the relation is with `Member` :/

dhensby:
Despite that, this code does seem somewhere sensible @tractorcow any thoughts on the code aspect?

@lekoala are you happy to write a test for this?

lekoala:
Well, it has to join on MyMember (because the Email field is on the Member table and the MobilePhone field is on MyMember table). Otherwise it would be the other use case that fails. My problem is that I'm extending Member, so I have two tables to deal with. 

My only concern with my change is that it was failing without the check for the table returned by table_for_object_field. I don't know what it wouldn't return a valid table at all, that doesn't make a lot of sense.

I'll try to find some time to write tests for my current PR :-) I have to admit I'm really lazy when it comes to unit testing.

forms/gridfield/GridFieldSortableHeader.php
```diff
@@ -233,9 +233,16 @@ public function getManipulatedData(GridField $gridField, SS_List $dataList) {
 					// been aliased. In that event, assume what the user has
 					// provided is the correct value
 					if(!$joinClass) $joinClass = $lastAlias;
-
+					
+					// Check on which table is the sort column
+					$table = ClassInfo::table_for_object_field($tmpItem->class, $parts[sizeof($parts)-1]);
```
tractorcow: Shouldn't this be $methodName? Also, the field will be `$methodName . 'ID'` for has_one relations.

You are using $parts[sizeof($parts)-1] within a loop, which will always join on the last segment in the dot separated column list, even when joining earlier tables.

tractorcow:
Code won't work for sortable columns greater than two parts :D Please also add tests to demonstrate this works.

tractorcow:
I just realised that I solved this problem ages back in 4.x. GridFieldSortableHeader.php now shunts off the work for sorting composite columns to DataList.


```php
public function getManipulatedData(GridField $gridField, SS_List $dataList) {
		if(!$this->checkDataType($dataList)) {
			return $dataList;
		}

		/** @var Sortable $dataList */
		$state = $gridField->State->GridFieldSortableHeader;
		if ($state->SortColumn == "") {
			return $dataList;
		}

		return $dataList->sort($state->SortColumn, $state->SortDirection('asc'));
	}
```

Feel free to go look at the 4.x code if you need inspiration for your 3.x fix. :)

lekoala:
I don't have much time to improve my current proposal or to add tests. I felt it was a good idea to commit my solution because it works on one of my project, but I'm afraid I won't be able to propose a decent, tested proposal in a reasonable amount of time.

tractorcow:
Oh well, maybe we just leave it in 3.x and wait for 4?

forms/gridfield/GridFieldSortableHeader.php
tractorcow: ok, so this is what I would expect to see.

```php
$table = ClassInfo::table_for_object_field(get_class($tmpItem), "{$methodName}ID");
```

tractorcow: And tests. :)

PR #6175, opened 2016-10-12
FIX Dont use superglobals for backurl

dhensby:
This `$data` var comes from the form submission which includes a hiddenfield containing the backurl.

This is the back url a form submission should be using rather than `$_REQUEST`.

This change also allows users to send `$data` outside of a typical form submission flow.

### DIFF
--- a/security/MemberLoginForm.php
+++ b/security/MemberLoginForm.php
@@ -246,7 +246,7 @@ JS;
 			);
 			Session::set("Security.Message.type", "good");
 		}
-		Controller::curr()->redirectBack();
+		return Controller::curr()->redirectBack();
 	}
 
 
@@ -272,9 +272,10 @@ JS;
 	 *                or NULL on failure.
 	 */
 	public function performLogin($data) {
+		/** @var Member $member */
 		$member = call_user_func_array(array($this->authenticator_class, 'authenticate'), array($data, $this));
 		if($member) {
-			$member->LogIn(isset($data['Remember']));
+			$member->logIn(isset($data['Remember']));
 			return $member;
 		} else {
 			$this->extend('authenticationFailed', $data);
--- a/security/MemberLoginForm.php
+++ b/security/MemberLoginForm.php
@@ -201,7 +201,7 @@ JS;
 		Session::clear('SessionForms.MemberLoginForm.Remember');
 
 		if(Member::currentUser()->isPasswordExpired()) {
-			if(isset($_REQUEST['BackURL']) && $backURL = $_REQUEST['BackURL']) {
+			if(isset($data['BackURL']) && $backURL = $data['BackURL']) {
 				Session::set('BackURL', $backURL);
 			}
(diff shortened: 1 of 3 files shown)


sminnee:
Is this fixing a bug or just a refactoring? Is the relevant behaviour covered by tests?

dhensby:
It's a bit of both. I'm working on a MFA module and it's useful to be able to call `logInUserAndRedirect` when not in the middle of a POST request to the login form.

This code actually provides the illusion of doing something it doesn't. Basically the `BackURL` `$_REQUEST` var is a `POST` var from a hidden field in the form (thus it's already in the `$data` that's passed in).

My change just means it's more self contained, portable and testable.

I haven't looked in detail at the tests, tbh - so I can't comment on that aspect.

sminnee:
Given this is a change to support new behaviours, I'd call it a new feature and put it against the `3` branch — I think we do everyone a lot of favours if we limit the patch releases to things that are clearly bugfixes.

Also, let's look into the tests to ensure that this case is covered by a test.

Other than that, the change makes sense.

dhensby:
I'd like to argue that using superglobals is a bug... but I agree that patch releases should probably be more cautious than this.

I've rebased

tractorcow:
@dhensby could you please add a test and I'll merge.

sminnee:
> I'd like to argue that using superglobals is a bug

You can't argue it's a grapefruit if you want; in both cases you're redefining language to suit your purposes. :-P

dhensby:
OK - there are zero tests for `LoginForm`, `MemberLoginForm` or `CMSLoginForm` at the moment. Are you happy with me making a test just for redirects?

sminnee:
Sure but can you structure it with some `@todo`s so it's obvious where to put the other tests?

tractorcow:
Note: already fixed up in 4.x.

dhensby:
tests in

security/MemberLoginForm.php
```diff
@@ -213,8 +213,8 @@ protected function logInUserAndRedirect($data) {
 		}
 
 		// Absolute redirection URLs may cause spoofing
-		if(!empty($_REQUEST['BackURL'])) {
-			$url = $_REQUEST['BackURL'];
+		if(!empty($data['BackURL'])) {
+			$url = $data['BackURL'];
```
tractorcow: If this form is posted, does $data include both $_POST and $_GET vars in the same way that $_REQUEST does?

tractorcow: MemberLoginForm does this

```
// Reduce attack surface by enforcing POST requests
		$this->setFormMethod('POST', true);
```
		
So $data won't include an $_GET vars.

doLogin() is passed the HTTPRequest as a third parameter though. You want to update to use this instead?

tractorcow: Sorry, I just realised that on postback BackURL is a hidden field not a querystring. Probably safe then?

tractorcow:
This can be merged on green.

maxime-rainville:
This has already been fixed in SS4.

PR #6176, opened 2016-10-12
Turn off codecov commenting on PRs

dhensby:
The codecov PR comments add a lot of noise IMO and I'd prefer we just use the review status as we do with travis to see the changes.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,2 +1,3 @@
+comment: false
 codecov:
   branch: master


chillu:
Good call - thanks for sorting that!

chillu:
I'll add it to the other (pending) PRs on modules as well

dhensby:
thanks :)

PR #6178, opened 2016-10-13
API Created a generic DataObject FormBuilder interface

tractorcow:
This is an experimental interface we are looking to introduce as an alternative to the getCMSFields model built on top of dataobjects self-scaffolding their own fields.

Initially we will be using this for AssetAdmin, where we have built specific File, Folder, and Image field scaffolder services.

Implements https://github.com/silverstripe/silverstripe-asset-admin/issues/253

Also includes asset-admin PR at https://github.com/silverstripe/silverstripe-asset-admin/pull/288

### DIFF
--- a/Forms/DefaultFormFactory.php
+++ b/Forms/DefaultFormFactory.php
@@ -1,0 +1,158 @@
+<?php
+
+namespace SilverStripe\Forms;
+
+use SilverStripe\Control\Controller;
+use SilverStripe\Core\Extensible;
+use SilverStripe\ORM\DataObject;
+
+/**
+ * Default form builder class.
+ *
+ * @internal WARNING: Experimental and volatile API.
+ *
+ * Allows extension by either controller or object via the following methods:
+ * - updateFormActions
+ * - updateFormValidator
+ * - updateFormFields
+ * - updateForm
+ */
+class DefaultFormFactory implements FormFactory {
+	use Extensible;
+
+	/**
+	 * @var Controller $controller
+	 */
+	protected $controller;
+
+	/**
+	 * @var DataObject $record
+	 */
+	protected $record;
+
+	/**
+	 * @param Controller $controller
+	 * @param DataObject $model
+	 */
+	public function __construct($controller, $model) {
+		$this->setController($controller);
+		$this->setRecord($model);
+		$this->constructExtensions();
+	}
+
+	/**
+	 * Generates a form using the stored controller and model.
+	 *
+	 * @param string $name
+	 * @return Form $form
+	 */
+	public function getForm($name = 'Form') {
+		$record = $this->getRecord();
+		$fields = $this->getFormFields();
+		$actions = $this->getFormActions();
+		$validator = $this->getFormValidator();
+		$form = Form::create($this->controller, $name, $fields, $actions, $validator);
+
+		// Extend form
+		$this->extendAll('updateForm', $form);
+
+		// Load into form
+		$form->loadDataFrom($record);
+
+		return $form;
+	}
+
+	/**
+	 * Build field list for this form
+	 *
+	 * @return FieldList
+	 */
+	protected function getFormFields() {
+		// Fall back to standard "getCMSFields" which itself uses the FormScaffolder as a fallback
+		$record = $this->getRecord();
+		// @todo Deprecate or formalise support for getCMSFields()
+		$fields = $record->getCMSFields();
+		$this->extendAll('updateFormFields', $fields);
+		return $fields;
+	}
+
+	/**
+	 * Build list of actions for this form
+	 *
+	 * @return FieldList
+	 */
+	protected function getFormActions() {
+		// by default no actions, it's a bit unpredictable
+		$record = $this->getRecord();
+
+		// Support legacy behaviour
+		// @todo Deprecate or formalise support for getCMSActions()
+		$actions = $record->getCMSActions();
+
+		// Extend actions
+		$this->extendAll('updateFormActions', $actions);
+		return $actions;
+	}
+
+	/**
+	 * @return Controller
+	 */
+	public function getController() {
+		return $this->controller;
+	}
+
+	/**
+	 * @param Controller $controller
+	 * @return $this
+	 */
+	public function setController(Controller $controller) {
+		$this->controller = $controller;
+		return $this;
+	}
+
+	/**
+	 * @return DataObject
+	 */
+	public function getRecord() {
+		return $this->record;
+	}
+
+	/**
+	 * @param DataObject $record
+	 * @return $this
+	 */
+	public function setRecord(DataObject $record) {
+		$this->record = $record;
+		return $this;
+	}
+
+	/**
+	 * @return Validator|null
+	 */
+	protected function getFormValidator() {
+		$validator = null;
+		$record = $this->getRecord();
+
+		// Support legacy behaviour
+		if ($record->hasMethod('getCMSValidator')) {
+			// @todo Deprecate or formalise support for getCMSValidator()
+			$validator = $record->getCMSValidator();
+		}
+
+		// Extend validator
+		$this->extendAll('updateFormValidator', $validator);
+		return $validator;
+	}
+
+	/**
+	 * Runs the given extension on the builder, record, and controller
+	 *
+	 * @param string $method name of extension method to call
+	 * @param mixed $object Object parameter
+	 */
+	protected function extendAll($method, &$object) {
+		$this->invokeWithExtensions($method, $object, $this);
+		$this->getRecord()->invokeWithExtensions($method, $object, $this);
+		$this->getController()->invokeWithExtensions($method, $object, $this);
+	}
+}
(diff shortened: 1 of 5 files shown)


Forms/DefaultFormBuilder.php
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	/**
+	 * Perform some action prior to update($name)
+	 *
+	 * @param string $name Identifier for this update
+	 * @param callable $callback Callback to invoke
+	 */
+	protected function beforeUpdate($name, $callback) {
+		$this->beforeExtending("update{$name}", $callback);
```
flamerohr: I think this is an artefact of the `update()` function we refactored out

tractorcow: oops, will remove!

tractorcow: I've removed this method; It's not necessary at all.

Forms/DefaultFormBuilder.php
```diff
@@ -0,0 +1,176 @@
...
+/**
+ * Default form builder class.
+ *
+ * Allows extension by either controller or object via the following methods:
+ * - updateFormActions
+ * - updateFormValidator
+ * - updateFormFields
+ * - updateForm
+ */
+class DefaultFormBuilder implements FormBuilder {
```
chillu: Mark as experimental?

tractorcow: Yes I should.

Forms/FormBuilder.php
```diff
@@ -0,0 +1,30 @@
...
+
+	/**
+	 * @return DataObject
+	 */
+	public function getRecord();
+
+	/**
+	 * @return Controller
+	 */
+	public function getController();
```
chillu: Do you need a controller reference if you want to implement a generic form builder for a record? It's not available in `getCMSFields()` unless you use `Controller::curr()`. The form building might rely on controller context though (request vars, HTTP headers), so there will be form builders which require a reference. Hmmm...

Forms/DefaultFormBuilder.php
tractorcow: Done.

chillu:
One thing to note here (though it doesn't impact the code): In a more API-focused CMS, forms would only be responsible for encapsulating the view, and wouldn't do any processing:

 * No conditional form fields in its `FieldList` creation (go through React and declarative view logic instead)
 * No `save` handlers (go through API instead)
 * No data loading from record (go through API instead)

Now, that's a long way off from where we are, but we should take this use case into account as this experimental API morphs into the preferred API :)

tractorcow:
@chillu sounds good, but definitely another scope above this change;

The main intention of this is to break out the "field builder" logic from DataObject, which is not the correct place for this.

I would like to see a separation of Forms into view / request handler components some time in the future, as the current implementation attempts to fill both roles.

Forms/FormBuilder.php
tractorcow: Controller::curr() is not useful here, in that the form being built may be for one that is not the current controller. :)

The intention of exposing the controller in this way is to make it available to extensions. In extendAll() the current form builder is passed not only to extensions on itself, but also to the controller and the model. There may be times where extensions on the model may need to know which controller, and vice versa, extensions on the controller may need to know which model.

flamerohr:
I would like to work towards a no conditional fieldlist, though it would mean relying on a view handler like react to display the form?
Sounds almost like a dependency there

tractorcow:
At the very simplest, we add a standard "visible" flag to all form fields. That should be simple enough to do.

Forms/FormBuilder.php
```diff
@@ -0,0 +1,30 @@
...
+
+namespace SilverStripe\Forms;
+
+use SilverStripe\Control\Controller;
+use SilverStripe\ORM\DataObject;
+
+/**
+ * A service which can generate a form for a given record and controller
+ */
+interface FormBuilder {
```
sminnee: I assume that we'd want to use this interface in CMSMain as well?

Calling it FormBuilder makes it sound like something tightly coupled to the React FormBuilder system. Perhaps `FormFactory`?

Forms/FormBuilder.php
```diff
@@ -0,0 +1,30 @@
...
+
+	/**
+	 * @return DataObject
+	 */
+	public function getRecord();
```
sminnee: `$record` is probably an argument to getForm(), not a getter. This is a Form factory, not a DataObject factory.

Forms/DefaultFormBuilder.php
```diff
@@ -0,0 +1,178 @@
...
+	 */
+	public function setRecord(DataObject $record) {
+		$this->record = $record;
+		return $this;
+	}
+
+	/**
+	 * @return Validator|null
+	 */
+	public function getFormValidator() {
```
sminnee: I don't think we should have public methods that aren't part of the interface.

chillu:
> At the very simplest, we add a standard "visible" flag to all form fields. That should be simple enough to do.

Just to clear this up: It couldn't be a `FormField::$visible` flag in the PHP class, since the whole point is to make form presentation more deterministic: You shouldn't need to reload the entire form schema and re-render the form view just to load another record into it.

Forms/FormBuilder.php
tractorcow: This method is necessary for extensions on controllers; If generating fields for a form, extensions on the current controller will need to know (in many cases) which record is being scaffolded for. Note that this occurs in many places before the Form itself has been constructed. It's not intended that the builder generates the record itself, only that it holds it.

Forms/DefaultFormBuilder.php
tractorcow: We can protect this.

Forms/DefaultFormBuilder.php
```diff
@@ -0,0 +1,178 @@
...
+		$validator = $this->validator;
+		$record = $this->getRecord();
+
+		// Support legacy behaviour
+		if (!$validator && $record->hasMethod('getCMSValidator')) {
+			Deprecation::notice('5.0', 'Implement updateFormValidator() instead of providing getCMSValidator()');
```
chillu: Might be a bit too early to throw deprecation notices if we're not sure where this API goes?

tractorcow:
> Just to clear this up: It couldn't be a FormField::$visible flag in the PHP class, since the whole point is to make form presentation more deterministic: You shouldn't need to reload the entire form schema and re-render the form view just to load another record into it.


No, but this could be a part of the state couldn't it?

Forms/DefaultFormBuilder.php
tractorcow: Yes I think we should revoke the deprecation notices until we stabilise this API.

Forms/FormBuilder.php
tractorcow: I think the problem here is that we are lacking a way of providing context for a formbuilder. In extensions it will be necessary not to simply know that a form is being scaffolded, but also why and how. For instance, we will need a way to distinguish between a userform scaffolding a front-end form and a cms-edit form. How will extensions on UserForm know how to determine this context and update the form accordingly? In each case, the same record is being provided, but a different Controller (CMSEditPageController vs UserForm_Controller). Thus it would need to query the getController() method on the second argument to the `updateFormFields` method.

Likewise, the context of the record itself is necessary; Certain records will have different fields, and thus extensions on the controller may need to respond. For instance, an extension on ModelAdmin may need to know if the underlying record has versioned or not (and thus provide additional publishing controls), thus it would query the getRecord() method for this use case.

tractorcow: Note that extensions on either the record OR the controller may be invoked at each point, thus the duality of the public getters so that either can query the other component. :)

Forms/FormBuilder.php
tractorcow: FormFactory is great!

chillu:
> Just to clear this up: It couldn't be a FormField::$visible flag in the PHP class, since the whole point is to make form presentation more deterministic: You shouldn't need to reload the entire form schema and re-render the form view just to load another record into it.

> No, but this could be a part of the state couldn't it?

Forms should eventually be populated from API calls which are agnostic to the presentation of the records they return. Any derived data (such as a notice showing "Author X requested publication 4 hours ago") should be computed on the client, and pull in related data as necessary. That's the theory at least, we're far away from that - it's not clear how we can retain the developer experience of `getCMSFields()` in this new model. Plus, `FormField::$visible` fails to take into account the distinction between "can't see" and "can't edit" - SilverStripe forms are abused for field-level access control because there's no other built-in mechanism in the ORM.

tractorcow:
Perhaps rather than `visible` the state is `visibleIf`, and given a set of rules?

tractorcow:
I've made the following updates:
 - Renamed FormBuilder -> FormFactory
 - Removed setValidator, made getValidator and all other non-interface methods protected.
 - Removed $validator parameter from constructor
 - Added unit tests for FormFactory to demonstrate how it can be extended

tractorcow:
I've made the following changes:
 - FormFactory is now a stateless service
 - Scaffolding of forms now takes a $context parameter

sminnee:
OK I'm happy with this now I'm gonna merge it in. 🎉

PR #6179, opened 2016-10-13
API Support fixture paths relative to current directory

tractorcow:
Yes, this includes SapphireTestTest. Nothing cannot be tested.


kinglozzer:
> `SapphireTestTest`

Issue #6170, opened 2016-10-11
Unable to scroll edit page in Firefox


ewweaver:
When editing a page in firefox, you are unable to scroll down the page. You can tab down the form to edit the fields.

The toolbar with the publish / save buttons will not appear higher than the end of either the edit page form or the bottom of the site tree. This means that either is taller than you screen, you are unable to access the toolbar without tabbing all the way down the page to it.

FF 49.0.1

Works fine in Chrome.

kinglozzer:
Which version of SilverStripe are you using? I’m unable to recreate this in 3.4 / Firefox 49.0.1

ewweaver:
This is in master

kinglozzer:
Looks like a regression from the work to replace JLayout with Flexbox, two-part fix at:

- https://github.com/silverstripe/silverstripe-framework/pull/6180
- https://github.com/silverstripe/silverstripe-cms/pull/1647

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-cms/pull/1669

PR #6180, opened 2016-10-13
Fix flexbox 'fill-height' overflowing container

kinglozzer:
First part of a fix for #6170, only seems to affect Firefox.

https://www.fxsitecompat.com/en-CA/docs/2014/initial-value-of-min-width-height-on-flex-items-has-been-reverted-to-auto/

I have no idea why sourcemaps are being removed 😢 I did:

```
$ rm -rf node_modules
$ npm install
$ npm run css
```

Help 😅 

cc @flamerohr, @clarkepaul

### DIFF
--- a/admin/client/dist/AssetUploadField.css
+++ b/admin/client/dist/AssetUploadField.css
@@ -1,41 +1,41 @@
 /******/ (function(modules) { // webpackBootstrap
 /******/ 	// The module cache
 /******/ 	var installedModules = {};
-/******/
+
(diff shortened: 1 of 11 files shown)


flamerohr:
oh, that is interesting! thanks @kinglozzer 

Yeah, I was tweaking around with optimising builds and because sourcemaps were in `.gitignore` they were pulled out of the production webpack config.

I might put it back in for css only, as javascript was the main painpoint.

PR #6186, opened 2016-10-16
BUG Use simplexml_load_string instead of simplexml_load_file

### LINKED ISSUE #6174: simplexml_load_file fails to read locale data
This bug is related to this [forum thread](https://www.silverstripe.org/community/forums/general-questions/show/101293).

On a server with NGINX and php-fpm 5.6.26 I encounter this warning in different silverstripe installations:

```
[Warning] simplexml_load_file(): I/O warning : failed to load external entity ".../framework/thirdparty/Zend/Locale/Data/en.xml"
```

followed by different kind of errors, depending (I think) on the code that triggered the simplexml call, e.g.:

```
[Error] Uncaught Zend_Locale_Exception: Wrong format... missing 0 .../framework/thirdparty/Zend/Locale/Format.php:359
```

```
[Error] Uncaught Zend_Locale_Exception: failed to receive data from locale it_IT .../framework/thirdparty/Zend/Locale/Format.php:1139
```

After **a lot** of research, my guess is this is related to [PHP #62577](https://bugs.php.net/bug.php?id=62577) together with [PHP #64938](https://bugs.php.net/bug.php?id=64938). In other words, when `libxml_disable_entity_loader` is true, the `simplexml_load_file` call just fails.

By leveraging the fact `simplexml_load_string` is not affected, my current workaround (under testing) is:

```
diff --git a/framework/thirdparty/Zend/Locale/Data.php b/framework/thirdparty/Zend/Locale/Data.php
index 826d5ad..4f458d5 100644
--- a/framework/thirdparty/Zend/Locale/Data.php
+++ b/framework/thirdparty/Zend/Locale/Data.php
@@ -158,7 +158,7 @@ class Zend_Locale_Data
                 throw new Zend_Locale_Exception("Missing locale file '$filename' for '$locale' locale.");
             }

-            self::$_ldml[(string) $locale] = simplexml_load_file($filename);
+            self::$_ldml[(string) $locale] = simplexml_load_string(file_get_contents($filename));
         }

         // search for 'alias' tag in the search path for redirection
```

Not sure it is worth a pull request though.

ntd:
PHP #62577 [1] together with PHP #64938 [2] make simplexml_load_file()
fails when the "disable load external entities" flag is set.

As a workaround, use simplexml_load_string() that does succeed
independently of the external entities setting. We are loading internal
XML files after all, so no security implications should arise.

[1] https://bugs.php.net/bug.php?id=62577
[2] https://bugs.php.net/bug.php?id=64938

Fix #6174.

### DIFF
--- a/main.php
+++ b/main.php
@@ -62,6 +62,9 @@ if(!empty($_SERVER['HTTP_X_ORIGINAL_URL'])) {
 	$_SERVER['REQUEST_URI'] = $_SERVER['HTTP_X_ORIGINAL_URL'];
 }
 
+// Enable the entity loader to be able to load XML in Zend_Locale_Data
+libxml_disable_entity_loader(false);
+
 /**
  * Figure out the request URL
  */


dhensby:
Changing the installer requirements feels risky. If you have one function aren't you guaranteed to have the other?

ntd:
> If you have one function aren't you guaranteed to have the other? 

No, but if this PR is accepted I really need to have `simplexml_load_string`, not `simplexml_load_file`.

ntd:
More background to explain why this bug, although highly disruptive, is not so widespread.

To trigger this error you need two things:

1. php in fpm mode (because of PHP bug #64938);
2. a PHP script that leaves the disable entity loader flag set to true.

While the former is a relative common configuration, the latter is quite uncommon. In my case I hit that bug only after installing [phabricator](https://github.com/phacility/phabricator) that, in fact, disable the entity loader on its [startup code](https://github.com/phacility/phabricator/blob/master/support/PhabricatorStartup.php#L394).

dhensby:
OK - I have 2my concerns are 2:

1. People who don't have `simple_xml_load_string` will have broken sites on their hands when they upgrade (this is a SemVer violation and not good for anyone)
2. You're editing `thirdparty` deps, which we shouldn't be doing. (sorry, just noticed this).

If this problem is affecting you mostly because of an unusal set-up, couldn't you resolve the set-up problem? IE: reset the disable entity loader flag?

ntd:
I addressed your two concerns with a new approach: please check it out if this can land upstream.

> If this problem is affecting you mostly because of an unusal set-up, couldn't you resolve the set-up problem? IE: reset the disable entity loader flag?

Here I don't follow: I already worked around this bug. I'd prefer to include the fix upstream for two main reasons: for convenience and to help others having the same issue. If you think too few people are affected, I do not agree but I can understand: feel free to close this pr.

This is not a bug of silverstripe and it is not a bug of phabricator: this is a PHP bug. The best solution would be to have it resolved upstream but... we are in a real world and those bugs are open for more than three years.

dhensby:
I'm happy with this, then.

@sminnee - any thoughts?

sminnee:
I assume that `libxml_disable_entity_loader` is set to false by default, so we're not introducing a security hole with this?

I'm okay with merging this but I'd probably look to remove this in SS4, once we've refactored the locale system. We can set by calling `libxml_disable_entity_loader(true)` and confirming that the system works.

ntd:
> I assume that libxml_disable_entity_loader is set to false by default...

Yes.

> ...so we're not introducing a security hole with this?

Yes, in theory. But AFAIK we are loading XML file only in `Zend_Locale_Data`, and that call loads internal XML files anyway.

```
[nicola@iota sapphire]$ grep -r simplexml_load_file --exclude-dir .git/ .
./dev/TestRunner.php:		$xml = simplexml_load_file($xmlFile);
./dev/install/install.php5:		$this->requireFunction('simplexml_load_file', array(
./thirdparty/Zend/Locale/Data.php:            self::$_ldml[(string) $locale] = simplexml_load_file($filename);
```
I found XML related stuff here and there (`CSSContentParser`, `Convert`, `SS_HTML4Value`) but nothing seems to be loading external XML. A double check could be needed though.

FWIW the new `Zend_Locale_Data` [does not](https://github.com/zf1/zend-locale/blob/master/library/Zend/Locale/Data.php#L160) use `simplexml_load_file` either and they also have a [hack](https://github.com/zendframework/zf1/blob/master/library/Zend/Xml/Security.php#L73) because of PHP #64938.

dhensby:
> We can set by calling libxml_disable_entity_loader(true) and confirming that the system works.

The problem is this is a core PHP bug that's been open 3 years and doesn't look to be getting any love - so not sure we can do much to fix it on our end.

PR #6195, opened 2016-10-18
API Writes to versioned dataobjects always writes to stage

### LINKED ISSUE #6059: Modify versioned dataobject behaviour
Acceptance criteria:

 * When saving content in live mode, both draft and live are updated.
 * Re-label "Archive" as "Delete" in the CMS.
 * Drop the "Delete from draft" action in the CMS.

-----

PRS:
 - [ ] https://github.com/silverstripe/silverstripe-framework/pull/6195
 - [ ] https://github.com/silverstripe/silverstripe-cms/pull/1653
 - [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/291

-----

At the moment we have these kinds of behaviour for versioned dataobjects:

- Save - Modifies record on the current draft only (or creates if not existant)
- Archive - Deletes from both live and draft
- Delete - Deletes from current stage only
- Publish - Writes draft record to live

The problem with this is that, when modifying records on the frontend, no changes to the draft record is effected. E.g. creating new images in live mode, where no draft record ever gets created.

I propose we modify these operations, whereby more operations effect draft in predictable ways.

- Save - On draft mode, this will write the record to draft, and not to live. It will require a publish in the future. However, if on live mode, performing a write will update both stages to the new record.
- Archive - Drop the term as it's confusing.
- Delete - Deletes from both live and draft mode
- Publish - As is

This establishes "draft" version of any record as the source of truth. Additionally, it doesn't matter if a record is created on live, or on draft; It will always be written to both draft, as well as the stage that the record is currently being viewed in.

tractorcow:
This changes removes the separate "archive" action and consolidates it with delete.
Writing records in live mode now also writes to draft, as a source of truth.

Permission behaviour on ChangeSetsItems and Versioned objects tweaked to be slightly more flexible.

Fixes #6059 

Merge with https://github.com/silverstripe/silverstripe-asset-admin/pull/291 and https://github.com/silverstripe/silverstripe-cms/pull/1653

### DIFF
--- a/admin/client/src/legacy/LeftAndMain.BatchActions.js
+++ b/admin/client/src/legacy/LeftAndMain.BatchActions.js
@@ -8,12 +8,12 @@ $.entwine('ss.tree', function($){
 
   /**
    * Class: #Form_BatchActionsForm
-   * 
+   *
    * Batch actions which take a bunch of selected pages,
    * usually from the CMS tree implementation, and perform serverside
    * callbacks on the whole set. We make the tree selectable when the jQuery.UI tab
    * enclosing this form is opened.
-   * 
+   *
    * Events:
    *  register - Called before an action is added.
    *  unregister - Called before an action is removed.
@@ -63,7 +63,7 @@ $.entwine('ss.tree', function($){
      */
     registerDefault: function() {
       // Publish selected pages action
-      this.register(ss.config.adminUrl + 'pages/batchactions/publish', function(ids) {
+      this.register('publish', function(ids) {
         var confirmed = confirm(
           i18n.inject(
             i18n._t(
@@ -77,7 +77,7 @@ $.entwine('ss.tree', function($){
       });
 
       // Unpublish selected pages action
-      this.register(ss.config.adminUrl + 'pages/batchactions/unpublish', function(ids) {
+      this.register('unpublish', function(ids) {
         var confirmed = confirm(
           i18n.inject(
             i18n._t(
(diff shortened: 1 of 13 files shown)


tractorcow:
Updated based on feedback from @clarkepaul

clarkepaul:
Cheers @tractorcow

admin/client/src/legacy/LeftAndMain.BatchActions.js
```diff
@@ -271,39 +242,49 @@ $.entwine('ss.tree', function($){
...
       // apply callback, which might modify the IDs
-      var type = this.find(':input[name=Action]').val();
+      var actionURL = this.find(':input[name=Action]').val();
+      if (!actionURL) {
+        e.preventDefault();
+        return false;
+      }
+
+      // Validate action
+      var type = actionURL.split(/\//).filter(n => true).pop();
```
flamerohr: I'd suggest `.split("/")` for no need for Regex reasons.
What does the `.filter(n => true)` do? To me it isn't doing anything (empty values are retained)

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -892,6 +892,11 @@ A very small number of methods were chosen for deprecation, and will be removed
   * Any extension declared via `versionableExtensions` config on Versioned dataobject must now 
     `VersionableExtension` interface at a minimum. `Translatable` has been removed from default
     `versionableExtensions`
+  * The default CMS delete behaviour for versioned dataobjects is now to delete from both draft
+    and live stages, and to save to the archive. There is no longer a separate "Archive" action.
+  * Any writes to versioned dataobjects (in either Draft or Live) will always write to the draft
+    (main) table as a source of truth. This means that records created in Live mode will always be
+    available to the CMS and in draft mode.
```
chillu: Should this read `...will always be available to the CMS in draft mode.` (without the `and`)?

PR #6199, opened 2016-10-19
FIX Ambiguous column SQL error

jonom:
I noticed that I couldn't use `Max('LastEdited')` with certain Datalists. Seems to be an issue if you use a filter that joins other tables, and results in an error such as:

>  "Column 'LastEdited' in field list is ambiguous"

Have included a test which fails prior to this patch.

In limited testing this seems to fix the problem and not cause side effects. To be honest though I don't understand very well what I've done here, so would appreciate the fine-toothed comb treatment from someone who understand SQL and the ORM better and would know if this is likely to cause regressions.

### DIFF
--- a/model/DataQuery.php
+++ b/model/DataQuery.php
@@ -392,7 +392,8 @@ class DataQuery {
 	 * automatically so must not contain double quotes.
 	 */
 	public function max($field) {
-		return $this->aggregate("MAX(\"$field\")");
+		$table = ClassInfo::table_for_object_field($this->dataClass, $field);
+		return $this->aggregate("MAX(\"$table\".\"$field\")");
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


tractorcow:
Change your test to a numeric column (e.g 'ID'). avg() isn't defined for dates in postgres.

jonom:
Thanks Damian, have updated

tractorcow:
There is no such thing as an average date with a developer is there? :P

dhensby:
great fix

dhensby:
Regression: This causes an error when sorting on `many_many_extraFields`

tractorcow:
We need to take care of the case where table_for_object_field returns DataObject. It's a currently undesirable feature of this method.

jonom:
> Regression: This causes an error when sorting on many_many_extraFields

@dhensby can you give example code to reproduce? It seems to work fine for me, for example `$Coupons.sort('SortOrder').Max('LastEdited')` works in template where SortOrder is defined on many_many_extraFields

dhensby:
@jonom if you look at @tractorcow's commit above, I assume those tests would fail currently

tractorcow:
Fix at https://github.com/silverstripe/silverstripe-framework/pull/6307

jonom:
Thanks for fixing @tractorcow sorry for the regression... bug fixing is a game of whack-a-mole!

PR #6203, opened 2016-10-20
API Rename _versions table to _Versions

### LINKED ISSUE #2365: Inconsistent capitalization of versions table names
Acceptance criteria:

 * Use `_Versions` instead of `_versions` as the suffix
 * Automatically rename the table to `_versions` table to `_Versions`

----

For example, "SiteTree_versions" should be "SiteTree_Versions", to match "SiteTree_Live", "SiteTree_EditorGroups", etc.

This capitalization change would apply to all versions tables.

table-names

tractorcow:
API Support case-sensitive changes to tables
Fixes #2365

Merge with https://github.com/silverstripe/silverstripe-cms/pull/1655


mikenz:
Yay! and oh.... 

The name has always bugged me though.

tractorcow:
It only took a few years!

PR #6204, opened 2016-10-20
BUG Fix cms tests running within framework CI

tractorcow:
Note that CMS tests are currently failing due to existing bugs awaiting merge.

I suggest to merge this change, and remove from `allowed_failures` at a future date once tests are green.

### DIFF
--- a/tests/behat/serve-bootstrap.php
+++ b/tests/behat/serve-bootstrap.php
@@ -1,5 +1,8 @@
 <?php
 
+use SilverStripe\CMS\Model\SiteTree;
+use SilverStripe\CMS\Controllers\ContentController;
+
 // Asset folder
 if(!file_exists(BASE_PATH . '/assets')) {
 	mkdir(BASE_PATH . '/assets', 02775);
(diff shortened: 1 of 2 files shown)


tractorcow:
Once https://github.com/silverstripe/silverstripe-cms/pull/1653 is merged I expect that CMS CI will be green again.

tests/behat/serve-bootstrap.php
```diff
@@ -1,5 +1,8 @@
 <?php
 
+use SilverStripe\CMS\Model\SiteTree;
```
sminnee: I don't think this coupling to CMS makes sense; we probably need a `serve-bootstrap.php` within the CMS module itself.

tests/bootstrap.php
```diff
@@ -45,57 +45,8 @@
 	$_REQUEST = array_merge($_REQUEST, $_GET);
 }
 
-// DATABASE BOOTSTRAP
```
sminnee: Don't quite see how we can ditch this bootstrap?

tests/behat/serve-bootstrap.php
```diff
@@ -37,11 +40,35 @@
...
+// Mock installer 'mysite' ONLY if installed with 'cms'
+$projectPath = BASE_PATH.'/'.$project;
+if (is_dir(BASE_PATH .'/cms') && !is_dir($projectPath)) {
+	$pageBase = SiteTree::class;
+	$controllerBase = ContentController::class;
+	$pageDotPHP = <<<EOS
+<?php
+// Auto-generated via serve-bootstrap.php
+
+class Page extends $pageBase {}
```
sminnee: This might be better done as a .travis.yml step (e.g. a `cp` operation of a Page.php file from tests somewhere) so that it can be loaded into the manifest for behat as well as phpunit?

tests/behat/serve-bootstrap.php
tractorcow: `use` statements are safe to use if the class doesn't exist; SiteTree::class itself doesn't trigger autoloading, so do we feel concerned about this being here?

i.e. `class_exists(SiteTree::class)` will safely return false without the CMS module.

tests/behat/serve-bootstrap.php
tractorcow: I agree; Better to make this an explicit copy instead of an auto-magic one (which could be hard to track down if you are debugging something).

tests/behat/serve-bootstrap.php
tractorcow: I'll implement the below suggestion anyway; Make it an explicit copy, so we won't need this here in any case. :)

tractorcow:
Abandoned in favour of #6257

Issue #5522, opened 2016-05-11
Implement server-side form validation feedback to React


chillu:
## Acceptance criteria

 * Form-level error messages generated by the server are displayed after a react form is submitted
 * Field-specific error messages generated by the server are displayed after a react form is submitted

## Tasks

 - [x] Setup server side validation to work with Form schema
 - [x] Setup Message box component to display errors returned by the schema

## Details

We've made this possible through `Form->setValidationResponseCallback()` on the server, but we need to communicate any server-side validation errors back to the client. This could be scoped to a field (highlight with error state and message), or scoped to the form (add message to top of form, or provide status message).

See https://github.com/silverstripe/silverstripe-framework/issues/5523 for client-side validation

flamerohr:
Pull requests made
https://github.com/silverstripe/silverstripe-framework/pull/6205
https://github.com/silverstripe/silverstripe-asset-admin/pull/292

tractorcow:
Re-updated based on feedback.

flamerohr:
I've switch to the use of throwing a submission error, semantically incorrect though easier to implement than using a separate redux store for messages which is what would be needed.
Ref: https://github.com/erikras/redux-form/issues/94#issuecomment-143398399

PR #6205, opened 2016-10-20
Validation on server side for React forms

flamerohr:
Added logic to display validation messages that have been returned by the server.

This will display a `MessageBox` component either above the form (for form scoped messages) and beneath the field (for field scoped messages).

`MessageBox` extends Bootstrap4 alert messages.

Check https://github.com/silverstripe/silverstripe-framework/issues/5522 for more details


Forms/FormField.php
```diff
@@ -1501,25 +1501,14 @@ public function getSchemaState() {
...
-		}, $errors));
-
 		return [
 			'id' => $this->ID(),
 			'value' => $this->Value(),
-			'valid' => (count($messages) === 0),
-			'messages' => (array)$messages,
+			// @todo - Make form / field messages not always stored as html
+			'message' => [
+				'value' => ['html' => $this->Message()],
```
tractorcow: If $Message is empty, I suggest we set `message` to null, rather than an `['html' => null]`

Forms/Schema/FormSchema.php
```diff
@@ -65,7 +69,8 @@ public function getState(Form $form) {
 
 		if($form->Message()) {
 			$state['messages'][] = [
-				'value' => $form->Message(),
+				// TODO Make form / field messages not always stored as html
+				'value' => ['html' => $form->Message()],
```
tractorcow: I suggest same as above; Nullify if not set.

Also we agreed to change `messages` to a single `message`. Shall we change it for fields as well to simplify things and keep consistency?

Forms/FormField.php
flamerohr: good point, I'll do that!

Forms/Schema/FormSchema.php
flamerohr: yeah, I think keeping this consistent is important, I had left this as `messages` for if more messages were to be added - but that can be for another story.

tractorcow:
Checked with @clarkepaul ; Message needs to be beneath input not beside.

See example from frameworktest.

image

flamerohr:
I think the Password+Confirm fieldgroup has some unique markup which allows it to have space above the validation error message.

When looking at a single field non-fieldgroup, there is no spacing above:
screen shot 2016-10-25 at 1 58 21 pm

tractorcow:
Message without spacing is fine with me. :D

tractorcow:
Something else to consider (and I think this NEEDS to be another story) is whether field level validation should use the bootstrap styles.

https://v4-alpha.getbootstrap.com/components/forms/?#validation

image

clarkepaul:
That would be a nice direction :).

flamerohr:
@sminnee has created https://github.com/silverstripe/silverstripe-framework/issues/6223 for "Improving validation error messages"

Forms/Form.php
```diff
@@ -1377,14 +1370,17 @@ public function getLegend() {
...
 
 	/**
-	 * Returns last result of validation, or null if not validated.
-	 * Note: Does not persist between requests in session.
-	 * @todo Maybe it should?
+	 * Returns boolean if there are any errors set for this form, or null if not validated.
 	 *
 	 * @return bool|null
 	 */
 	public function isValid() {
```
chillu: That's pretty misleading, since `isValid()` relies on `setupFormErrors()` to be called beforehand. Can we avoid creating a new `Form->isValid()` API for this and simply use `Form->validate()`?

tractorcow: This api needs to be separate from ->validate(), because it also needs to determine if the form has been validated already. By calling ->validate() you can no longer determine this.

tractorcow: We could address the prior issue by ensuring that validate() loaded field-level messages without requiring setupFormErrors().

tractorcow: Would a more appropriate name be `isValidated`?

tractorcow: Hm, how about FormStatus, with a return value of one of the following consts:

 - UNSUBMITTED
 - SUBMITTED_VALID
 - SUBMITTED_INVALID

Forms/Schema/FormSchema.php
```diff
@@ -67,10 +67,10 @@ public function getState(Form $form) {
 			$this->getFieldStates($form->Actions())
 		);
 
-		if($form->Message()) {
-			$state['messages'][] = [
```
chillu: I know the PHP `Form` API can only show a single message at the moment, but I'm not sure we should carry through this limitation to `FormSchema` and frontend code as well. I can easily imagine situations where you might want different form messages, e.g. one styled as an error, the other one as a warning. Can we just use an array with a single element instead?

tractorcow: The limitation does keep the API simple though; I will look at allowing multiple messages for the form, but let's keep a single message per field for now.

admin/client/src/state/schema/SchemaReducer.js
```diff
@@ -10,6 +11,16 @@ export default function schemaReducer(state = initialState, action = null) {
       return deepFreeze(Object.assign({}, state, { [action.payload.id]: action.payload }));
     }
 
+    case ACTION_TYPES.CLEAR_MESSAGE: {
```
chillu: I'm not sure we need that - the `state` key in schema is basically considered out-of-date as soon as the form is displayed - form field values, form field validations, and form messages. Since `reduxForm()` is used with `enableReinitialize=false`, you can't propagate form field value changes from schema state to an existing form (and shouldn't need to).

If you carry this reducer-clearing approach through to form fields, you'd change the schema every time a form field validation error is resolved - which causes the whole form `render()` process to kick off (since the `reduxForm()` inputs through `mapStateToProps` changed). This is different from redux-form validation, which only calls `render()` in the individual field.

We could use `setState()` in the `fetch()` response of `FormBuilderLoader` for `state` instead? Or are we just accepting those `render()` calls?

Forms/Schema/FormSchema.php
chillu: Agreed - multiple messages for a form is likely useful, but multiple messages for form fields should be inlined into a single message, or showing the most important message first (like on the http://redux-form.com/6.1.1/examples/syncValidation/ example: `((error && <span>{error}</span>) || (warning && <span>{warning}</span>))`)

sminnee:
I think this needs to be changed to make use of `ValidationResult` objects rather than the session. It may be best to complete and merge in #3937 first and then base these changes on top of this. Right now we're adding layers to shaky foundations.

admin/client/src/state/schema/SchemaReducer.js
flamerohr: A whole form `render()` process happens with either a `setState()` call or a `schema` state change, it's how diffing for any DOM changes work.

I had considered using `setState()` in the `Form` class, which to me made more sense than `FormBuilderLoader`. Using `CLEAR_MESSAGE` in redux state gives a more predictable outcome than setting state, and since the message wasn't showing again until after the form was submitted, in which case you receive a new message and `schema` state to display.

Forms/Schema/FormSchema.php
flamerohr: I guess it's left a little unclear, are both (`Form` and `FormField`) APIs changing to plural messages, or only `Form` messages?

sminnee:
OK I've written #6228 that will address my concerns raised above. My only request for this PR is that we remove the new `isValid()` API as I don't want new public API that we plan on deleting.

Forms/Schema/FormSchema.php
tractorcow: I think just Form for now. I've made the appropriate change.

tractorcow:
Updated:
 - Form message now supports multiple messages (in react / schema, not yet in php)
 - Removed Form::isValid(), and added note into FormSchema::getState()

admin/client/src/state/schema/SchemaReducer.js
sminnee: The other approach here would be to have a value in the state that is the incorrect field value. The error message will only be shown if the field is still set to that value. This means that you don't need a state change to clear the error message, just a re-render:

```
...
{
  message: {
    'value': 'Not a valid email address',
    'for': 'sam@silverstripe,com',
  }
...

```

It will need a re-render to trigger—presumably `onblur` rather than `onchange`

tractorcow:
@flamerohr some weird behaviour I think may be related:
 - go to /admin/campaigns/set/0/create
 - Enter a name and save. You'll see a success message
 - Click "save" again, and nothing will happen.
 - Start typing a new name. The save button will immediately start animating. Weird!

flamerohr:
@tractorcow fixed the issue we had found where `create -> exit to list -> create` again caused an issue.

Forms/FormField.php
```diff
@@ -1501,27 +1501,23 @@ public function getSchemaState() {
...
 			'id' => $this->ID(),
 			'value' => $this->Value(),
-			'valid' => (count($messages) === 0),
-			'messages' => (array)$messages,
+			'message' => null,
 			'data' => [],
 		];
+		
+		if ($message = $this->Message()) {
+			// @todo - Make form / field messages not always stored as html
```
chillu: Inline todos should use `TODO` - no biggie, just sayin. I think this is more commonly picked up by IDEs and other code cleanliness tools. A single comment doesn't start a PHPDoc block, and `@todo` is PHPDoc notation

admin/client/src/components/MessageBox/MessageBox.js
```diff
@@ -0,0 +1,105 @@
+import React, { PropTypes } from 'react';
+import SilverStripeComponent from 'lib/SilverStripeComponent';
+import { Alert } from 'react-bootstrap-ss';
+import castStringToElement from 'lib/castStringToElement';
+
+/**
+ * A wrapper for Alert messages in react-bootstrap.
+ * Displays a given message.
+ */
+class MessageBox extends SilverStripeComponent {
```
chillu: Why are we calling this a `MessageBox` when it's just wrapping an `Alert` bootstrap component? I think if we're just creating "better alert", the file path is enough to convey that. `import Alert from 'components/Alert/Alert'` vs. `import { Alert } from 'react-bootstrap-ss'`. We've used the same approach for `components/PopoverField/PopoverField` vs. `import { Popover } from 'react-bootstrap-ss'`

admin/client/src/components/Form/Form.js
```diff
@@ -1,27 +1,53 @@
...
-  constructor(props) {
-    super(props);
-
-    this.handleSubmit = this.handleSubmit.bind(this);
+  /**
+   * Generates a list of messages if any are available
+   *
+   * @returns {Array|null}
+   */
+  getMessages() {
```
chillu: We should prefix methods returning React components with `render*()`. This is implied by the AirBnB conventions as well: https://github.com/airbnb/javascript/tree/master/react#ordering. So `renderMessages()` in this case.

admin/client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -65,21 +65,17 @@ class FormBuilder extends SilverStripeComponent {
...
     const submitFn = (customData) =>
       this.submitApi(customData || dataWithAction, headers)
         .then(formSchema => {
-          resetSubmittingFn();
+          this.setState({ submittingAction: null });
           return formSchema;
         })
         .catch((reason) => {
           // TODO Generic CMS error reporting
           // TODO Handle validation errors
```
chillu: Remove `TODO Handle validation errors`, that's done right?

admin/client/src/containers/FormBuilderLoader/FormBuilderLoader.js
```diff
@@ -2,7 +2,12 @@ import React, { PropTypes, Component } from 'react';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import fetch from 'isomorphic-fetch';
-import { Field as ReduxFormField, reduxForm } from 'redux-form';
+import {
+  Field as ReduxFormField,
+  reduxForm,
+  SubmissionError,
+  destroy as ReduxDestroyForm,
```
chillu: methods shouldn't be upper camel case.

admin/client/src/containers/FormBuilderLoader/FormBuilderLoader.js
```diff
@@ -25,12 +43,55 @@ class FormBuilderLoader extends Component {
...
+   * @param state
+   * @returns {object}
+   */
+  getMessages(state) {
+    const messages = {};
 
-    if (this.props.onSubmitSuccess) {
-      this.props.onSubmitSuccess(result);
+    // only error messages are collected
+    // TODO define message type as standard "success", "info", "warning" and "danger"
```
chillu: I don't understand how/where this TODO would be resolved - do you mean add an enum to proptypes?

tractorcow: This is due to inconsistencies in the way that silverstripe internally names messages; They are non-bootstrap identifiers.

admin/client/src/components/MessageBox/MessageBox.js
chillu: It's no more related to forms than then underlying `<Alert>` - we just happen to add a way to persist dismissal, and translate our own message type terminology into bootstrap terminology.

chillu:
Found another case where the button loading animation continues: 

 * Cause a field-level validation error (by submitting the form)
 * Submit the form again
 * Fix the error (don't submit)
 * The button loading animation shows, even though no submission is in progress

admin/client/src/components/MessageBox/MessageBox.js
flamerohr: Yeah, I had used `MessageBox` as a placeholder, didn't give a lot of thought about a proper name, `AlertBox` was my alternative.
Anyway, `Alert` can be used instead.

admin/client/src/containers/FormBuilderLoader/FormBuilderLoader.js
flamerohr: Yes, it's a bit scattered in terms of "error types" or "message types", I remember running into a case where some validation errors even return the validation test name rather than "error" as the type which convolutes what to look for even more.

Forms/FormField.php
tractorcow: +1 moved to phpdoc

admin/client/src/components/MessageBox/MessageBox.js
tractorcow: Renaming to FormAlert

admin/client/src/components/Form/Form.js
tractorcow: +1

admin/client/src/components/FormBuilder/FormBuilder.js
tractorcow: +1

PR #3628, opened 2014-11-11
Parsing fix for searching on DateFields

JeremyRoundill:
This fixes an issue where DateFields only accept ISO 8601 dates.

PR #6206, opened 2016-10-20
Prevent undefined index notices

Chibby:
Moving the foreach loop to inside the if statements makes sure the array is actually an array. The old way was generating 'Undefined index' notices.

Perhaps an `isset($params[$field->getName()])` should also be performed.


dhensby:
thanks

dhensby:
this patch needs to be merged into earlier branches :/

dhensby:
#3628 - 3.4+

PR #6211, opened 2016-10-21
BUG Fix : relObject() should return null if one of the node is null

jason-zz:
# Current Problem

`relObject()` didn't check if $component is null in `foreach` loop. In some cases, following fatal error happens.

> ( ! ) Fatal error: Call to a member function hasMethod() on a non-object in /www/framework/model/DataObject.php on line 3073
# Sample Codes

```
class Product extends Page {
    private static $has_many = array (
        'Images' => 'ProductImage',
    );
    private static $summary_fields = array(
        'Images.First.Image.CMSThumbnail'   => 'Image',
        'Title'             => 'Title',
        'URLSegment'        => 'URL'
    );
}
```

```
class ProductImage extends DataObject {
    private static $db = array (
        ....
    );

    private static $has_one = array (
        'Image' => 'Image',
        'Parent' => 'Product'
    );
}
```

```
class ShopAdmin extends VersionedModelAdmin {

    private static $title       = 'Shop';
    private static $menu_title  = 'Shop';
    private static $url_segment = 'shop';          

    private static $managed_models  = array(
        'Product'
    );
}
```
# Steps to reproduce
1. Create Product without adding ProductImage.
2. Go to ShopAdmin and fatal error happens.

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -3069,8 +3069,11 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 			$relations = explode('.', $fieldName);
 			$fieldName = array_pop($relations);
 			foreach($relations as $relation) {
+				// Bail if the component is null
+				if(!$component) {
+					return null;
 				// Inspect $component for element $relation
-				if($component->hasMethod($relation)) {
+				} elseif($component->hasMethod($relation)) {
 					// Check nested method
 					$component = $component->$relation();
 				} elseif($component instanceof SS_List) {


kinglozzer:
I think the example is definitely abusing the `$summary_fields` API to try to do something it’s not supposed to do, but we should let developers know that.

I think it would be better if, rather than just returning null, it raised the `user_error()` beneath that block. Perhaps you could try something like:

```php
elseif($component instanceof DataLast
	&& singleton($component->dataClass())->hasMethod($relation)
) {
	// Select adjacent relation from DataList
	$component = $component->relation($relation);
}
```

tractorcow:
@kinglozzer I disagree; There are valid non-error conditions where a part of a chain could be null. The error is there intentionally when attempting to query a field or method on an object which is invalid, not simply that a part of the chain along the way is null. E.g. a page may query a banner title, but may not be assigned a banner yet; In which case the correct result is null. :)

In any case, @jason-zz , could you please add a unit test to demonstrate your fix?

kinglozzer:
@tractorcow I was looking more at the use of `.First` in the `$summary_fields` definition, but that is actually covered by unit tests so you’re correct!

`has_one` relations return empty objects rather than null, so I think you’d have to be using a `RelationList` to trigger this. I’m not really comfortable with being able to call arbitrary methods like `First()` in there, but that’s the existing behaviour so we should definitely fix it :)

dhensby:
@jason-zz looks like we just need a test :)

dhensby:
@jason-zz PS: we're looking at a release this week - reckon you can do it soon?

dhensby:
This was autoclosed when the old 3.4 branch was deleted (as it's no longer maintained).

This PR still looks useful, we just need that test ;)

jason-zz:
Thanks. @dhensby . I'm not sure how to create proper test for this case. Can you provide some guide or example please ?

dhensby:
So https://github.com/silverstripe/silverstripe-framework/pull/7583 has been merged and it may be worth attempting to backport this approach instead?

tractorcow:
Yes, that approach could be back-ported into 3.7. It was done in as un-breaking a way as possible, so I don't think there are too many risks.

The added benefit of consolidating relObject() and relField() logic makes me feel this is worthwhile doing... should someone want to make the effort. :)

tractorcow:
I think this fix is safe enough that we can merge this into 3.5; I'd prefer it had a test, but we probably would better off merging this as-is than closing without a fix.

PR #6212, opened 2016-10-21
Adjustments for filter panel positioning

clarkepaul:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1622
Requires https://github.com/silverstripe/silverstripe-cms/pull/1658

### DIFF
--- a/admin/client/src/styles/legacy/_style.scss
+++ b/admin/client/src/styles/legacy/_style.scss
@@ -1167,7 +1167,7 @@ body.cms {
   box-sizing: border-box;
   margin-left: -$panel-padding-x;
   margin-right: -$panel-padding-x;
-  height: calc(100% + #{$panel-padding-y});
+  margin-bottom: -$panel-padding-y;
   padding: $panel-padding-y $panel-padding-x;
   background: lighten(#D4E2EC, 4);
   text-shadow: none;
--- a/admin/client/src/styles/legacy/_style.scss
+++ b/admin/client/src/styles/legacy/_style.scss
@@ -1163,13 +1163,13 @@ body.cms {
 }
 
 .cms-tree-filtered {
-  position: absolute;
   margin: 0;
-  width: 100%;
   box-sizing: border-box;
-  margin-left: -$grid-x*2;
-  padding: $grid-y*2 $grid-x*2;
-  background: #D4E2EC;
+  margin-left: -$panel-padding-x;
+  margin-right: -$panel-padding-x;
+  height: calc(100% + #{$panel-padding-y});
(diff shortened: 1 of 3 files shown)


admin/client/src/styles/legacy/_style.scss
```diff
@@ -1163,13 +1163,13 @@ body.cms {
...
-  position: absolute;
   margin: 0;
-  width: 100%;
   box-sizing: border-box;
-  margin-left: -$grid-x*2;
-  padding: $grid-y*2 $grid-x*2;
-  background: #D4E2EC;
+  margin-left: -$panel-padding-x;
+  margin-right: -$panel-padding-x;
+  height: calc(100% + #{$panel-padding-y});
```
flamerohr: This leaves a lot of space below the last page item, and for some reason gives a "snapping"/"jumping" experience when I scroll to the bottom.
I'd suggest a `margin-bottom: -$panel-padding-y;` to give a more consistent experience.
screen shot 2016-10-25 at 10 55 16 am

flamerohr:
Scrolling not enabled for panels by default, unable to scroll if list is long
screen shot 2016-10-25 at 10 53 13 am

clarkepaul:
Yup, looks like I've mucked something up there, probably a case of fixing one thing which breaks another. 
I'm going to leave this for now and hopefully I have time to get back to it later in the week, if you think you have time @flamerohr feel free to fix up but I've noticed the scrolling isn't working either.

clarkepaul:
Actually think I've found the issue, if so I'll push up :)

clarkepaul:
Added a fix to CMS branch, please test again @flamerohr.

flamerohr:
Thanks for the fix @clarkepaul :D

One issue I had with using `calc(100% + {$padding})` was what I had in the screenshot, happens when I resize the height of the browser in chrome.
screen shot 2016-10-26 at 10 12 48 am 2

The way I had fixed it was merged the `.cms-tree-filtered` div with the `.flexbox-area-grow` div container and gave it a negative bottom margin, happy to discuss this and find a better way around this.

clarkepaul:
Yeah doesn't the fix in the CMS branch stop that, I thought I had made that stop?

flamerohr:
The fix in the CMS branch allowed for scrolling.

The screenshote is with the fix in place, but the `calc()` css rule seems to miscalculate at times (note the large size of the panel)

clarkepaul:
Yeah I see what you mean, one view works better than the other. I'm fine with your change meaning the search background doesn't extend all the way to bottom of the panel. It is still a clean result.

Okay from me for how it is now.

Issue #4015, opened 2015-03-16
.Up not finding correct scope in certain circumstances


kinglozzer:
Haven’t got time to investigate this at the moment or write a test case, but:

```
<% with $NestedObjects.First %>
    {$Up.Title} < returns ''
    {$Up.Up.Title} < returns the correct title
<% end_with %>

<% with $NestedObjects.Limit(1).First %>
    {$Up.Title} < returns ''
    {$Up.Up.Title} < returns ''
    {$Up.Up.Up.Title} < returns the correct title
<% end_with %>
```

Same behaviour exists with `<% loop %>`. Looks like it’s just assuming that another `.` means a new scope?

dhensby:
Hmmmmm.

Very interesting.

Intuitively, what you're getting at makes sense - that the `with` or the `loop` should be the next scope. BUT aren't there circumstances where the existing behaviour is actually useful for keeping template syntax clean?

Now, I'd say that current behaviour *is* unexpected - but is it wrong? It actually would allow for cleaner templates with less nesting

kinglozzer:
Yeah it seems to me that it’s `<% with %>` or `<% loop %>` that are actually changing scope, not `.Limit` or `.AnythingElse`.

> BUT aren't there circumstances where the existing behaviour is actually useful for keeping template syntax clean?

Do you have an example of where this might be useful? The only one I can think of is:

```
<% loop $SomeObjects.Limit(1) %>
	{$Title}		<%-- Current object title --%>
	{$Up.Count}		<%-- $SomeObjects count (without the limit) --%>
	{$Up.Up.Title}	<%-- "Parent scope" title --%>
<% end_loop %>
```

Assuming this is indeed a bug, and we intend to fix it, it’d probably be a job for master IMO. I can’t believe I’m the first person ever to encounter this, so I imagine there are a few sites in the wild with extra `$Up`s to compensate for this that we’d break!

kinglozzer:
Just looking at this again, it looks like this behaviour is intentional as we have [test coverage](2/tests/view/SSViewerTest.php#L904-L932) for it. I’m still not sure whether or not this is good or bad!

If someone else sees something I don’t and thinks this is desirable, feel free to close this issue.

dhensby:
Hmm, interesting debate.

Does $Up mean up a block syntax or up a "selector"?

I guess the current behaviour is more powerful, but less expected...

AngryPHPNerd:
This is really confusing ...

[...]

PR #6213, opened 2016-10-21
Specify that the selectors change the scope

ntd:
Follow up of issue #4015.


jonom:
Thanks @ntd!

Template scope probably deserves it's own page rather being under common variables but that's an issue for another day

PR #6214, opened 2016-10-21
DOCS Remove duped content

jonom:
A big block of almost identical content appeared in two places. Better to have it in one place only so they don't get out of sync.


sminnee:
This seems logical to me. Any objections?

PR #6216, opened 2016-10-22
FIX Versioned sort by ID

jonom:
Prevent SQL error from a Versioned DataObject with e.g. `$default_sort = 'ID DESC'`

### DIFF
--- a/model/Versioned.php
+++ b/model/Versioned.php
@@ -313,7 +313,7 @@ class Versioned extends DataExtension implements TemplateGlobalProvider {
 			// Ensure that any sort order referring to this ID is correctly aliased
 			$orders = $query->getOrderBy();
 			foreach($orders as $order => $dir) {
-				if($order === "\"$baseTable\".\"ID\"") {
+				if($order === "\"$baseTable\".\"ID\"" || trim($order, "'\"") === "ID") {
 					unset($orders[$order]);
 					$orders["\"{$baseTable}_versions\".\"RecordID\""] = $dir;
 				}


tractorcow:
Well, yes it does, but strictly Versioned should not be rewriting fields that aren't scoped to a table. I guess in this one case it's probably ok, but you should really use fully qualified sort identifiers.

What if the sort is `$default_sort = '"ID" DESC';`, for instance?

jonom:
> you should really use fully qualified sort identifiers

Maybe I've been doing it wrong. What do you mean by 'fully qualified' here? Unless sorting on a relation I've never seen an example for sorting where you have to include the base table name if that's what you mean, but looking around now it doesn't seem to be very well documented.

If I use an array expression like `$default_sort = array('ID' => 'DESC')` without this patch I still get an SQL error. With the patch things work fine with that format.

If I change it to `$default_sort = array('TheBaseTableForMyClass.ID' => 'DESC')` I don't get an error with or without the patch. But all the examples for sorting are in the format `->sort('FieldName')` (no table prefix) so if the expectation is that this should work, shouldn't the Versioned code account for that?

> What if the sort is `$default_sort = '"ID" DESC'`, for instance?

I've updated the patch to trim quotes.

Here is the format of the error I'm getting by the way:
> no.1 of ORDER BY clause is not in SELECT list, references column 'MyDatabase.MyClass_versions.ID' which is not in SELECT list; this is incompatible with DISTINCT

dhensby:
Maybe the ID field for Versions table should not be `ID` because it makes sense that the Versioned module is able to sort by a field on the original table :/

tractorcow:
Fully qualified means qualified with column table.

Your example `$default_sort = array('TheBaseTableForMyClass.ID' => 'DESC')` isn't ANSI-92 SQL quoted so versioned doesn't know how to rewrite it.

Try

```php
private static $default_sort = '"BaseClass"."ID" DESC';
```

Unfortunately there is ONLY one right way to do this. :D

jonom:
> Your example $default_sort = array('TheBaseTableForMyClass.ID' => 'DESC') isn't ANSI-92 SQL quoted so versioned doesn't know how to rewrite it.

@tractorcow that example I gave *does* work, with or without this patch in place. It doesn't throw an error and sorting works fine. So it appears Versioned *does* know how to rewrite it :D

It's based an [example from the docs](https://docs.silverstripe.org/en/3.4/developer_guides/model/data_model_and_orm/), although it's in a different context:
```php
$players = Players::get()->sort(array(
    'FirstName' => 'ASC',
    'LastName'=>'ASC'
));
```

I made an assumption that the `$default_sort` config variable and `sort()` method would accept the same argument. Since it works it seems to be a fair assumption.

> Try `private static $default_sort = '"BaseClass"."ID" DESC';`

Your suggestion also works fine, with or without this patch 👍 

**So everything works fine if you include the base table, but I don't think you should have to.**

Looking through the core codebase it does seem that most core classes are quoted how you mention e.g. `$default_sort = '"Title"';` but I did find one instance of `$default_sort = 'Title ASC';` and I didn't find a **single** instance of a fully qualified value. SiteTree (the only class that uses Versioned by default) has private static `$default_sort = "\"Sort\"";`.

So I would argue that neither convention nor documentation shows you should need to fully qualify your sort clauses, and thus it's fair to expect that `$default_sort = '"ID" DESC';` should work. It works until you add the Versioned extension, and I don't think Versioned should be allowed to break this.

Sorry for the long explanation, but I hope it's enough to justify this fix. It's just an itty-bitty little change after all 😇 (famous last words). There is already a workaround in place for the ID column, I'm just trying to make it more robust.

Otherwise I know how to solve this problem for myself now at least (thanks!) by fully qualifying the sort clause on Versioned classes, but would be nice to prevent other people encountering this issue if possible.

PR #6217, opened 2016-10-23
OPTMIZATION 200% speed increase with ConfigManifest

SilbinaryWolf:
OPTMIZATION 200% speed increase with ConfigManifest.

By only making the FileFinder search 2 directories deep on a Silverstripe build with ~69 modules, I was able to shave off 1 second on PHP 5.x.

My results (i5-6600K, 3.5ghz, 8GB RAM, 64-bit Windows):
- PHP 5.6.25 -- before -- 1.8809859752655 seconds
- PHP 5.6.25 -- after -- 0.8601610660553 seconds
- PHP 7.0.10 -- before -- 1.6998739242554 seconds
- PHP 7.0.10 -- after -- 0.99255394935608 seconds

Other machine (i7-2620M, 2.7ghz, 8GB RAM, Samsung EVO 840 SSD, 64-bit Windows):
- PHP 5.5.12 -- before -- 2.9904389381409 seconds
- PHP 5.5.12 -- after -- 1.0939319133759 seconds

I also checked the 'count()' of config sources and YAML config fragments to ensure all of them were being picked up.

My testing code:
```
$t = microtime(true);

$finder = new ManifestFileFinder();
$finder->setOptions(array(
	'name_regex'    => '/(^|[\/\\\\])_config.php$/',
	'ignore_tests'  => !$includeTests,
	'file_callback' => array($this, 'addSourceConfigFile'),
	// Cannot be max_depth: 1 due to "/framework/admin/_config.php"
	'max_depth'		=> 2
));
$finder->find($this->base);

$finder = new ManifestFileFinder();
$finder->setOptions(array(
	'name_regex'    => '/\.ya?ml$/',
	'ignore_tests'  => !$includeTests,
	'file_callback' => array($this, 'addYAMLConfigFile'),
	'max_depth'		=> 2
));
$finder->find($this->base);

$t = microtime(true) - $t;
Debug::dump($t);
Debug::dump('Config: '.count($this->phpConfigSources));
Debug::dump('YML: ' .count($this->yamlConfigFragments));
```

### DIFF
--- a/core/manifest/ConfigManifest.php
+++ b/core/manifest/ConfigManifest.php
@@ -202,7 +202,9 @@ class SS_ConfigManifest {
 		$finder->setOptions(array(
 			'name_regex'    => '/(^|[\/\\\\])_config.php$/',
 			'ignore_tests'  => !$includeTests,
-			'file_callback' => array($this, 'addSourceConfigFile')
+			'file_callback' => array($this, 'addSourceConfigFile'),
+			// Cannot be max_depth: 1 due to "/framework/admin/_config.php"
+			'max_depth'     => 2
 		));
 		$finder->find($this->base);
 
@@ -210,7 +212,8 @@ class SS_ConfigManifest {
 		$finder->setOptions(array(
 			'name_regex'    => '/\.ya?ml$/',
 			'ignore_tests'  => !$includeTests,
-			'file_callback' => array($this, 'addYAMLConfigFile')
+			'file_callback' => array($this, 'addYAMLConfigFile'),
+			'max_depth'     => 2
 		));
 		$finder->find($this->base);
 


jonom:
Thanks for this @SilbinaryWolf, always nice to see things getting faster.

This isn't a bug fix so would need to target 3 or master branch instead... Personally I think this should go to master because although the docs don't strictly say you can, I have occasionally grouped code like mini-modules within the project folder to get a structure like `mysite/code/module/_config/`. It works so I've done it, and since you pointed out that framework/admin does that too, I think there's reason to believe this change could cause breakages on some sites.

If you retarget master with this PR, maybe add a note to the docs that config items must be no more than x levels deep to be picked up.

@micmania1 has been [looking in depth](https://groups.google.com/forum/#!topic/silverstripe-dev/q5khashNiuY) at the config system so he might have some thoughts on this too.

micmania1:
I'd also be worried about missing yaml files by adding the max_depth so this is probably best suited to master.

I've been working on integrating https://github.com/micmania1/silverstripe-config with SilverStripe over the past few days. I'm using `symfony/finder` as opposed to `ManifestFileFinder` but that also supports a max depth so anything added here can carried across.

Hopefully i'll be able to share some performance stats over the new week or two when I get time.

SilbinaryWolf:
@jonom You actually made me realize I have an old project that does a similar thing, so master is definitely the best bet.

I think moving forward, I'd rather write a specific file finder that only looks for the _config.php / _config/*.yml files in the specific places the documentation tells you to put them in. (This would extend ManifestFileFinder to use it's current acceptDir() code, unless we're swapping it out for symfony/finder)

For now, would it make more sense to leave this until you've integrated your config system with Silverstripe @micmania1 ?

jonom:
Huh, I didn't realise you could change the target of a PR. That is cool. I've closed this for now since it's pointing at 3.4 again but feel free to re-open.

micmania1:
I'd say this is still worth submitting against master (regardless of my work).

Issue #6164, opened 2016-10-10
Flexbox layout requires responsiveness


flamerohr:
With jLayout removed and flexbox used instead, it (still) lacks responsiveness for mobile and tablets.

Areas such as the Campaigns and Files which were originally responsive have lost this, will need to investigate how responsiveness could be achieved with using flexbox.

Loosely doing responsive improvements to the following areas (trying to get back to the same state as before flexbox being merged):
- [x] Pages area
  - Page history left menu for selecting versions isn't visible on smaller screens, I've opened a separate issue for this https://github.com/silverstripe/silverstripe-cms/issues/1663 but could be done as part of this work if time permits.
- [x] History panel missing versions at small screen sizes
- [x] Files area
- [x] Campaigns
- [x] Other areas shouldn't require responsive adjustments but general browser testing is still required.
- Not doing main left navigation as part of this work.

Campaigns area to do, requires React: 
At smaller screen size we need to deselect the first item so the preview is not visible, although the first item should remain active and show the preview for larger screens for a better experience.
When the "back" button is selected on campaign item preview the preview can be removed/hidden and the selected item from the list should have its active state removed so other items can be selected.

clarkepaul:
Incomplete PR, requires more work in campaigns mentioned above:
https://github.com/silverstripe/silverstripe-framework/pull/6222
https://github.com/silverstripe/silverstripe-cms/pull/1662
https://github.com/silverstripe/silverstripe-asset-admin/pull/293
https://github.com/silverstripe/silverstripe-siteconfig/pull/40/files

clarkepaul:
Still running through some browser testing (done a fair bit already) but I think its safe-ish to merge as continuing with browser testing on master too.

clarkepaul:
Couldn't find a quick fix for "filter and site tree scroll inside scroll" but will open a new issue for it as it is usable so shouldn't be a show stopper for the Alpha release.

clarkepaul:
Ready for review

tractorcow:
Reviewing.

tractorcow:
Feedback left on two pull requests:
- https://github.com/silverstripe/silverstripe-framework/pull/6222#issuecomment-257187548
- https://github.com/silverstripe/silverstripe-cms/pull/1662#issuecomment-257187222

tractorcow:
Merged.

PR #6222, opened 2016-10-25
Enhancements for flexbox on smaller screens

clarkepaul:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/6164

Requires Framework, CMS, AssetAdmin, Site-config branches of features/4.0/responsive-improvements


admin/client/src/components/Preview/Preview.js
```diff
@@ -7,6 +7,19 @@ import SilverStripeComponent from 'lib/SilverStripeComponent';
  */
 class Preview extends SilverStripeComponent {
 
+  constructor(props) {
+    super(props);
+
+    this.handleBackClick = this.handleBackClick.bind(this);
+  }
+
+  handleBackClick(event) {
```
sminnee: If onBack isn't supplied, it will just have a no-op. I think we should throw an error if onBack isn't set and the button is clicked. 

Alternatively, if onBack isn't supplied, remove the back button altogether.

tractorcow:
In the security section, reducing browser width removes the action button column. Perhaps it should remove all except the first / last columns?

image

tractorcow:
I don't know if this is related to the current change but the security cancel button styling is weird.

image

tractorcow:
In the campaign section there is inconsistent placement for the "back" button.

When previewing a page with an available preview the back button appears as this:

image

However if the page has no available preview, the back button appears in another place.

image

clarkepaul:
I've fixed the list view, will push shortly.
My security cancel button looks fine, have you checked out the siteconfig branch too?

clarkepaul:
@tractorcow In regards to showing the first and last column in grid fields, the problem is that all grid fields are composed differently so you can't always say what is first or last (eg page list has toggle first). With the Campaigns grid field I know there are two sets of actions at the end so I can show them, I also know they fit at a mobile screen size where-as other actions are sometimes accompanied by text which can break the layout. Actions are available once you select an item and go deeper so we are not removing functionality.

clarkepaul:
It was a conscious decision to use text to support the back arrow in the Campaign preview because it doesn't sit next to the inner page navigation like breadcrumbs or page name. Not a big deal really if I did remove the text but I think its better to decide that when we introduce the rest of the preview editing actions and draft/publish toggle.

PR #6224, opened 2016-10-25
API Update listview / treeview to use pjax instead of deferred loading

tractorcow:
Merge with https://github.com/silverstripe/silverstripe-cms/pull/1665


tractorcow:
CMS change approved, merging.

PR #6225, opened 2016-10-25
Remove double quotes so example is not parsed.

nfauchelle:
[Notice] Undefined variable: map

Since it's using double quotes it tries to process $map, $key and $value

### DIFF
--- a/model/Map.php
+++ b/model/Map.php
@@ -196,7 +196,7 @@ class SS_Map implements ArrayAccess, Countable, IteratorAggregate {
 		}
 
 		user_error(
-			"SS_Map is read-only. Please use $map->push($key, $value) to append values",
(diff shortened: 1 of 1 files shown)


kinglozzer:
This should go in a 3.x branch as it’s a bug fix, I’m just not sure which ones we maintain anymore @silverstripe/core-team

nfauchelle:
@kinglozzer 
I pushed it to silverstripe:3 - doesn't it just get taken form there and distributed around to whatever 3.x ones are relevant?

kinglozzer:
No as we merge changes up - from 3.1 -> 3.2 -> 3.3 -> 3.4 -> 3 -> master -, but not down (except in a few special cases)

nfauchelle:
ok. I've put it against the 3.1 branch here https://github.com/silverstripe/silverstripe-framework/pull/6238

PR #6229, opened 2016-10-26
FIX: helperPath in _register_database.php

SpiritLevel:
`MySQLDataBaseConfigurationHelper.php` is not in `framework/dev/install` but it is in `framework/Dev/Install` :)


sminnee:
Good find :-) will merge once the tests pass

PR #6230, opened 2016-10-26
Set MySQLPDODatabase as default database class in installer

SpiritLevel:


Dev/Install/install.php5
```diff
@@ -152,7 +152,7 @@ if(isset($_REQUEST['db'])) {
 		$databaseConfig['type'] = $type;
 	}
 } else {
-	$type = $_REQUEST['db']['type'] = defined('SS_DATABASE_CLASS') ? SS_DATABASE_CLASS : 'MySQLDatabase';
+	$type = $_REQUEST['db']['type'] = defined('SS_DATABASE_CLASS') ? SS_DATABASE_CLASS : 'MySQLPDODatabase';
```
tractorcow: This code is hard to read isn't it?

I wonder if we could automatically guess default installer based on dependency detection; It could be safer to fall back to mysqldatabase if the pdo_mysql driver isn't installed.

SpiritLevel: Yes, hard to read :) And the installer is quite lengthy at 1700+ lines! I wonder if it should be broken into pieces, each dealing with only one aspect of the installation ?

It would be good to encourage pdo but fall back to the more common mysqli...how's about detecting pdo with:

`extension_loaded('PDO') && extension_loaded(;pdo_mysql')` 

Here are [more options](http://stackoverflow.com/questions/3131411/php-code-to-test-pdo-is-available).

SpiritLevel: 'twas staring us in the face:

https://github.com/silverstripe/silverstripe-framework/blob/master/_register_database.php#L32

sminnee:
This makes sense to me, as would picking an alternative database driver if PDO wasn't available.

Do you want to make that change? I'd recommended doing it based on the DatabaseAdapterRegistry supported values rather than repeating the check.

SpiritLevel:
OK, I'll have a go at it... :)

SpiritLevel:
Closing. PR submitted here: https://github.com/silverstripe/silverstripe-framework/pull/6274.

PR #6231, opened 2016-10-26
API Shorten overly-verbose invalid extension error

### LINKED ISSUE #6198: Uploading invalid file type brings up a weird message
I think it's due to the vertical-align middle on the item message.

screen shot 2016-10-19 at 4 17 52 pm

tractorcow:
Fixes #6198

I suggest that, instead of a super long error message, there should be another way to pre-determine allowed types. Maybe an info box you can click to get a tooltip?

### DIFF
--- a/Assets/Upload_Validator.php
+++ b/Assets/Upload_Validator.php
@@ -277,10 +277,8 @@ class Upload_Validator
 		// extension validation
 		if (!$this->isValidExtension()) {
 			$this->errors[] = _t(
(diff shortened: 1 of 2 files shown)


dhensby:
hmm - providing a list of allowed extensions is pretty useful :/ why are we taking it out?

sminnee:
It broke the UI I believe.

tractorcow:
Yeah, the old error wasn't usefully presented in any part of the new asset admin.

PR #6233, opened 2016-10-26
FIX: path to ParameterConfirmationToken.php

SpiritLevel:
`ParameterConfirmationToken.php` is in `framework/Core/Startup` not `framework/core/startup`.


sminnee:
Nice find, but it has been fixed in other PR just now!

PR #6234, opened 2016-10-26
BUG Fix clicking "No items found" from causing ajax error	

tractorcow:
Attempted to fix https://github.com/silverstripe/silverstripe-cms/issues/1627 but found the original error had already fixed itself somehow, so I fixed this bug instead.


tractorcow:
Rebased and fixed with your feedback.

tractorcow:
@clarkepaul you're a committer now; Can you merge?

clarkepaul:
ohhh, just because I can doesn't mean I will unless I understand whats going on or reviewed so in this case its all good. 

Such privileges

PR #6236, opened 2016-10-26
FIX TreeDropdownField cached items

nyeholt:
When searching a tree dropdown field, and that dropdown field is looking at
the hierarchy of a class type that has already been iterated on that request,
the cached data present in Hierarchy will mean the search results do not
correctly mark items. Simplest solution is to clear the hierarchy cache.

There's likely a more elegant solution for someone with more time...


nyeholt:
Oops, my bad, meant to put this against 3.4, will re-PR

PR #6238, opened 2016-10-26
Remove double quotes so example is not parsed.

nfauchelle:
[Notice] Undefined variable: map

Since it's using double quotes it tries to process $map, $key and $value

--

This is a re-submit of https://github.com/silverstripe/silverstripe-framework/pull/6225 which was against the wrong branch.

### DIFF
--- a/model/Map.php
+++ b/model/Map.php
@@ -196,7 +196,7 @@ class SS_Map implements ArrayAccess, Countable, IteratorAggregate {
 		}
 
 		user_error(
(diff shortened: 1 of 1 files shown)


nfauchelle:
@kinglozzer 
Here you go.

kinglozzer:
Thanks @nfauchelle, @dhensby or I will merge this up :)

dhensby:
It's been merged up to 3...

PR #6246, opened 2016-10-27
Simpler test run

tractorcow:
Framework changes for https://github.com/silverstripe/silverstripe-cms/pull/1661


tests/bootstrap/init.php
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * This bootstraps the SilverStripe system so that phpunit can be run directly on SilverStripe tests.
+ */
+
+// Make sure display_errors is on
+ini_set('display_errors', 1);
+
+// Asset folder
+if(!file_exists(BASE_PATH . '/assets')) {
```
sminnee: Future idea: have an in-memory asset backend that the tests use.

tests/bootstrap/environment.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+// Bootstrap _ss_environment.php
+
+if (!defined('SS_ENVIRONMENT_TYPE')) {
+	define('SS_ENVIRONMENT_TYPE', 'dev');
+}
+
+if (!defined('SS_DATABASE_CLASS') && !defined('SS_DATABASE_USERNAME')) {
```
sminnee: Future idea: allow these `define()`s to be set by environment variables (probably as part of ConfigureFromEnv) and then put the relevant defines explicitly into .travis.yml

Issue #5523, opened 2016-05-11
Implement client-side form validation feedback in React forms


chillu:
Likely requires us to choose a React forms library first (some options: https://react.rocks/tag/Form). Could use JSON Schema as part of the form field schema we're generating, as a way to communicate validation constraints from the server to the client. Includes styling through Bootstrap (http://v4-alpha.getbootstrap.com/components/forms/#validation).

Redux Form has [pluggable validation](http://redux-form.com/5.2.5/#/examples/synchronous-validation?_k=edvj7l). Note that we can get quite far with powerful declarative validation, as the popularity of https://github.com/sheadawson/silverstripe-zenvalidator proves.

Overview of adding validation to Redux Form: https://medium.com/@rajaraodv/adding-a-robust-form-validation-to-react-redux-apps-616ca240c124#.v8bfb62m7

**TO DO:** Revisit the size after #5524 is complete.

## Tasks
- [x] Require "Name" on File and Folder detail sidebar in AssetAdmin (split out from https://github.com/silverstripe/silverstripe-asset-admin/issues/156)
- [x] Implement new schema for validation
- [x] Research on a validation library to use
- [x] Implement client side validation logic in Form components to work with schema
- [x] Write unit tests for Validator class

# Related
- Server-side validation: https://github.com/silverstripe/silverstripe-framework/issues/5522
- Implement a React form library: https://github.com/silverstripe/silverstripe-framework/issues/5524

flamerohr:
throwing in this `TODO` I noticed while peer reviewing, so it's not forgotten:
https://github.com/silverstripe/silverstripe-asset-admin/pull/287/files#diff-1725a22613a9ff930ae321601182c714R80

flamerohr:
I've been searching around for a validation library that we could use, some criteria I had in mind were:
* Can provide many basic validation rules
* Can be used with ReactJS
* Documentation reasonably covers API.
* No heavy dependencies is a plus, e.g. no jQuery.
* Extensibility is a plus.

## Recommendation
### [Approve.js](https://github.com/CharlGottschalk/approvejs)
Approve.js is a great candidate that would suit our needs.
It was designed to be a non-intrusive layer for frontend/backend validation, the maintainer himself uses it with `Vue.js`.
The library is extensible, so custom validation rules can easily be added in.
Pros:
* Extensible, can add custom validation rules easily.
* Can customise validation messages as required.
* API is basically passing in an object with the rules contained.
[...]

PR #6251, opened 2016-10-30
Implement client-side form validation feedback in React forms

flamerohr:
Now that we're on `redux-form` library with `FormBuilder`, lets have some validation added so it doesn't have to go through the server for obvious issues in validation. (e.g. require)

Check out https://github.com/silverstripe/silverstripe-framework/issues/5523 for more details.

Please merge https://github.com/silverstripe/silverstripe-asset-admin/pull/299 as well after this pull request

### DIFF
--- a/src/ORM/Versioning/ChangeSet.php
+++ b/src/ORM/Versioning/ChangeSet.php
@@ -399,6 +399,7 @@ class ChangeSet extends DataObject {
 		if ($this->isInDB()) {
 			$fields->addFieldToTab('Root.Main', ReadonlyField::create('State', $this->fieldLabel('State')));
 		}
+
 		$this->extend('updateCMSFields', $fields);
 		return $fields;
 	}
--- a/admin/client/dist/js/bundle.js
+++ b/admin/client/dist/js/bundle.js
@@ -58,69 +58,64 @@ return this.fetch(e,{method:"put",credentials:"same-origin",body:s(t),headers:n}
 return this.fetch(e,{method:"delete",credentials:"same-origin",body:s(t),headers:n}).then(a)}}]),e}(),O=new P
 t["default"]=O},function(e,t,n){n(8),e.exports=self.fetch.bind(self)},function(e,t){!function(e){"use strict"
 function t(e){if("string"!=typeof e&&(e=String(e)),/[^a-z0-9\-#$%&'*+.\^_`|~]/i.test(e))throw new TypeError("Invalid character in header field name")
-return e.toLowerCase()}function n(e){return"string"!=typeof e&&(e=String(e)),e}function i(e){var t={next:function(){var t=e.shift()
-return{done:void 0===t,value:t}}}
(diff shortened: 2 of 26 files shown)


tractorcow:
I will review this later today.

tractorcow:
Reviewing now. :)

tractorcow:
Some bugs discovered:

- Fields that have rules but are NOT required somehow become required. E.g. field with `numeric` as a rule but not `required`

image

- Messages for multiple failed rules are not formatted well.

image

tractorcow:
I think the solution to this is to bake into rule lookup some kind of hierarchy:

 - If not required, automatically validate if empty
 - If required AND empty, invalidate, but don't process any other rules.
 - If requried and not empty, process other rules

tractorcow:
Error is traced to schema:
 - form schema ID for the checkboxset field is `Checkboxes`
 - form state ID for the checkbox field is `Form_DetailEditForm_Checkboxes`
 
 So of course it  can't match the IDs to merge state + schema.

tractorcow:
OH! This is because FormField::setSchemaData() acts very bad. If you ever call it, it re-generates the entire schema and stores it. This means that if, later on down the line, the form changes (which would alter the field ID) those changes are no longer reflected.

We can fix this by ensuring that FormField::setSchema() only stores the changed schema, not the full schema.

Before:

```php
public function setSchemaData($schemaData = []) {
	$current = $this->getSchemaData();
	$this->schemaData = array_merge($current, array_intersect_key($schemaData, $current));
	return $this;
}
```

After


```php
public function setSchemaData($schemaData = []) {
	$this->schemaData = array_merge($this->schemaData, array_intersect_key($schemaData, $current));
	return $this;
}
```

Default data will still be merged in getSchemaData(), but not cached.

tractorcow:
Once I fixed that bug, there is still an issue with `required` for multi-select fields.

Checking and unchecking any box treats the field as though it has a value, even though it's empty, and incorrectly allows it to be submitted.


image

flamerohr:
Thanks @tractorcow, I had made the logic simpler "if value is blank and required not defined, don't validate"
and other cases would validate.
And I've fixed the display of each message, they are each on their own line now.

tractorcow:
I've tested and it looks ok.

src/ORM/Versioning/ChangeSet.php
```diff
@@ -399,6 +403,12 @@ public function getCMSFields() {
 		if ($this->isInDB()) {
 			$fields->addFieldToTab('Root.Main', ReadonlyField::create('State', $this->fieldLabel('State')));
 		}
+		$fields->addFieldToTab('Root.Main', $field = new TextField('OptionalField'));
```
tractorcow: Temp code here needs deleting before merge.

flamerohr: Done :D

PR #6256, opened 2016-10-31
OPTMIZATION 200% speed increase with ConfigManifest's use of Manifest…

SilbinaryWolf:
Continuation of https://github.com/silverstripe/silverstripe-framework/pull/6217

Made the changes against 'master' branch this time.

### DIFF
--- a/docs/en/02_Developer_Guides/04_Configuration/00_Configuration.md
+++ b/docs/en/02_Developer_Guides/04_Configuration/00_Configuration.md
@@ -165,6 +165,10 @@ pair in the mask is removed
 
 ## Configuration YAML Syntax and Rules
 
+<div class="alert" markdown="1">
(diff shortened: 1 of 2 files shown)


wilr:
Guess 2 levels would break any modules included via composer (not that SilverStripe supports that officially just yet anyway). Would 3 levels be safer (vendor/<owner>/<repo>/_config.

dhensby:
I'm happy with this.

dhensby:
It does need rebasing, though

chillu:
@SilbinaryWolf Can you please rebase? Also, the docs update is a bit misplaced I think, it should be in the 4.0 changelog with this wording. The configuration docs itself are already correct, stating "Each module can have a directory immediately underneath the main module directory called _config". The fact that you were able to nest modules within modules is a bug IMHO (which we exploit for `framework/admin` at the moment).

SilbinaryWolf:
@chillu Done.

PR #6257, opened 2016-10-31
BUG restore CMS build as required in framework tests

tractorcow:
Also use framework config.yml, as this is the root module for CI tests.


tractorcow:
Ah, only a few easily fixed things left to go. Almost there.

tractorcow:
CMS tests in framework don't seem to like using the framework context; I've created a new cms-config.yml specifically for use when testing cms within framework. Unfortunately the framework path differs, otherwise I'd use the config that comes with cms module.

tractorcow:
Greens!

PR #6263, opened 2016-10-31
NEW Allow setting of unlimited row counts on GridFieldPaginator

dhensby:
This change allows a dev to use `->setItemsPerPage(0)` to create a GridField with unlimited row count.

### DIFF
--- a/forms/gridfield/GridFieldPaginator.php
+++ b/forms/gridfield/GridFieldPaginator.php
@@ -179,15 +179,23 @@ class GridFieldPaginator implements GridField_HTMLProvider, GridField_DataManipu
 		$totalRows = $this->totalItems;
 		if(!$totalRows) return null;
 
-		$totalPages = (int)ceil($totalRows/$this->itemsPerPage);
-		if($totalPages == 0)
-			$totalPages = 1;
-		$firstShownRecord = ($state->currentPage - 1) * $this->itemsPerPage + 1;
-		if($firstShownRecord > $totalRows)
-			$firstShownRecord = $totalRows;
-		$lastShownRecord = $state->currentPage * $this->itemsPerPage;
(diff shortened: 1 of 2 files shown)


tractorcow:
Any chance of a unit test?

robbieaverill:
Just my opinion:

* Setting a null value for a limit can imply a reset to default rather than unlimited
* Perhaps `-1` would be better

sminnee:
It should match the semantics of `DataList::limit()`. The semantics of `DataList::limit()` are really badly documented but I think a limit of `0` specifies unlimited.

robbieaverill:
>It should match the semantics of `DataList::limit()`

Good call.

>...but I think a limit of `0` specifies unlimited.

But zero is a number too! :)

tractorcow:
I wouldn't even mind an unlimit() method (in addition) to explicitly remove any existing limit.

sminnee:
There's a solution that already exists. Robbie doesn't like it, but I don't think that mandates a change.

When I said "It should match the semantics of DataList::limit()" I didn't mean "Let's re-litigate the semantics of DataList::limit()"

dhensby:
The `null` vs `0` discussion is moot because currently the `setItemsPerPage` method doesn't handle an explicit null type and I won't add it to this PR because it's not part of the feature set I'm implementing. 

I'll write a test soon :)

dhensby:
Test added

tractorcow:
It broke. =(

dhensby:
That'll teach me for making up APIs

dhensby:
hopefully that'll fix it...

tractorcow:
It seems to be not. =(

dhensby:
OK - looks like green!

tractorcow:
We need to take care of those.

PR #6266, opened 2016-11-01
Move code files into the src/ folder. Fixes #6098

### LINKED ISSUE #6098: Shift class files into `src/`
In keeping with the conventions of other packages, I think we should move the class files loaded by the PSR-4 autoloader into `src/`.

We should also update our supported module standard to favour `src/` over `code/` but to allow either.

sminnee:


tractorcow:
Bootstrapping Core.php and messing around with the path seems a pain. Maybe make it a proper class with a single static method `init`, and rely on class autoloading?

tractorcow:
You could write it such a way that legacy code still works.


```php
<?php

class Core {
	protected static $init = false;
	public static function init() {
		if (static::$init) {
			return;
		}
		static::$init = true;
		// init everything
	}
}
Core::init();
```

sminnee:
I've gone with `include_path` because although I think that moving towards the use of classes is a good idea, I'd prefer to do that as part of a more considered refactor of the Core.php code. It's a bit soupy, and I'd prefer not to simply pour the soup into a class.

PR #6269, opened 2016-11-01
FIX default_records are no longer inherited to child classes

dhensby:
At the moment `default_records` are inherited by child classes which means if you define a default record on an object or page class all child classes also build the same records.

This feels unintended and unhelpful so this patch only adds the records for the particular class the `default_records` config is defined on


tractorcow:
good idea :)

sminnee:
Yeah agree with this.

dhensby:
test failures due to bug in simple theme (I think) that is causing 3-branch to fail

PR #6272, opened 2016-11-01
BUG Fix regressions from src folder creation

tractorcow:
`framework` is no longer an include_path, so update all dependencies on this to rely on either the `src` or `thirdparty` include_path roots.


src/Dev/TestSession.php
```diff
@@ -239,9 +239,9 @@ public function cssParser() {
 	 * @return SimplePage The response if available
 	 */
 	public function lastPage() {
-		require_once("thirdparty/simpletest/http.php");
```
sminnee: Everyone once and I while I realise that certain pieces of code are still in there :-P

PR #6274, opened 2016-11-02
Setting default db adapter in installation  as PDO MySQL

SpiritLevel:
Following up on https://github.com/silverstripe/silverstripe-framework/pull/6230.

I also added "Re-check requirements" button to CMS Admin Account section of `config-form.html`.

This solution works but is not all that satisfactory to me. There is repetition and I don't know how the error is best handled. More generally, the code between lines 140 and 188 seems awkward.

### DIFF
--- a/_register_database.php
+++ b/_register_database.php
@@ -1,26 +1,11 @@
 <?php
 
-// Register the SilverStripe provided databases
+// Register database adapters available in SilverStripe
 use SilverStripe\Dev\Install\DatabaseAdapterRegistry;
 use SilverStripe\Dev\Install\MySQLDatabaseConfigurationHelper;
 
-// Use MySQLi as default
-DatabaseAdapterRegistry::register(
-	array(
-		/** @skipUpgrade */
-		'class' => 'MySQLDatabase',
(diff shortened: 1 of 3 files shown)


tractorcow:
thanks @SpiritLevel great patch

SpiritLevel:
Good :) I think I might have a go at a refactoring of what's in `framework/src/Dev/Install`...

micmania1:
@SpiritLevel @tractorcow can any of you remember why this line wasn't included here? https://github.com/silverstripe/silverstripe-framework/pull/6230/files#diff-f8f7d711330a3259d575d72fc7e5a5cfR155

This has caused problems for us as local environments were defaulted to PDO (as per .env) but UAT ignored .env and was using the framework default of MySQLi. We found a bug in MySQLi connector which meant we could only re-produce the bug on UAT and it was difficult to track this down as the cause.

At the moment framework defaults to MySQLi and the installer defaults to PDO.

cc @stojg @chillu

micmania1:
I've opened an issue here: https://github.com/silverstripe/silverstripe-framework/issues/8540

PR #6275, opened 2016-11-02
Fixed jsmin path

mikenz:
After the move to src/ jsmin can no longer be found.


kinglozzer:
Test failures are related, I think you can just do:

```php
require_once('jsmin/jsmin.php');
```

sminnee:
Fixed by #6279

PR #6277, opened 2016-11-02
BUG Restore travis artifacts

### LINKED ISSUE #6046: BUG Travis artifact upload broken due to removal of travis-support
We will need to develop another script (inline within .travis.yml, or maybe a separate bashfile) that can be substituted for doing travis artifcact upload on behat error.

https://github.com/silverstripe/silverstripe-framework/blob/08a39f5f93cb546abb6ffdccb7fe4605a96a8ece/.travis.yml#L65

tractorcow:
Fixes #6046

### DIFF
--- a/tests/behat/travis-upload-artifacts.php
+++ b/tests/behat/travis-upload-artifacts.php
@@ -1,0 +1,96 @@
+#!/usr/bin/env php
+<?php
+/**
+ * Creates an index.html with links to all files in a given directory structure, recursively.
+ * This is useful for Amazon S3 uploads with static file hosting, since it doesn't list files by default.
+ *
+ * Assumes to run in a SilverStripe webroot
+ */
+
+/**
+ * Run a command
+ *
+ * @param string $cmd
+ * @param bool $echo
+ */
+function run($cmd, $echo = true) {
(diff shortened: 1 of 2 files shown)


tests/behat/travis-upload-artifacts.php
```diff
@@ -0,0 +1,96 @@
...
+$opts = getopt('', array(
+	'artifacts-path:',
+	'target-path:',
+	'if-env:',
+	'artifacts-base-url:',
+));
+
+// --if-env=BEHAT_TEST means that this script will only be executed if the given environment var is set
+if (empty($opts['if-env'])) {
+	echo "--if-env option is mandatory";
```
tractorcow: Enforced --if-env to prevent this script being invoked outside of the correct context.

sminnee:
Good to merge when green

tractorcow:
Behat tests currently fail on CMS occasionally. Not due to this PR, which is the point of getting artifacts fixed again. :) I suggest we fix this even with the failure.

sminnee:
Merging as this will help diagnose the intermittent failure still present in the CMS build.

PR #6282, opened 2016-11-03
FIX: Ensure that builds use the 3.2 dependencies.

sminnee:


sminnee:
Hold off on this one until #6283 is reviewed. We may have a simpler approach that we can cherry-pick in

sminnee:
OK this is just a cherry-pick of #6283 now

PR #6283, opened 2016-11-03
FIX: Ensure that builds use the 3.4 dependencies.

sminnee:


tractorcow:
Build is still installing 3.x (not 3.4.x) dependencies of siteconfig / reports.

tractorcow:
Build is currently  broken and this fixes it, so I don't think the above issue I noted blocks this merge.

sminnee:
That's because the cms dependencies for those modules are expressed at `~3.4`, rather than `~3.4.0`.

Since the build is passing I'd probably just leave that other issue as-is. In 4.0 we can look at more long-term solutions. IMO, I would probably remove siteconfig and reports as dependencies of cms.

dhensby:
It looks to me that this breaks tests for tags, branches that are named 3.5.x

It also breaks runs on branches that aren't numbers...

PR #6289, opened 2016-11-04
State variables for pages and files

clarkepaul:
Added state variables for the little circles which can be used on files and page icons to indicate whether an item is in draft or modified.


dhensby:
Don't we need the dist files too?

tractorcow:
I think these variables should be substituted; I found a couple hard-coded colour references in `_tree.scss`

```scss
@include tree-status-icon('addedtodraft', #ff7f22, #F46B00, #fff7f2);
@include tree-status-icon('modified', #fff2e8, #F46B00, #fff7f2);
@include tree-status-icon('archived', #5F7688, #455b6c, #f7f7f7);
@include tree-status-icon('deletedonlive', #f7f7f7, #455b6c, #f7f7f7);
@include tree-status-icon('removedfromdraft', #f7f7f7, #455b6c, #f7f7f7);
@include tree-status-icon('workflow-approval', #0070B4, #0070B4, #E8FAFF);
```

clarkepaul:
I didn't do a cleanup of all the legacy variables as yet but it is a plan to remove and migrate them to the variables list.

There are a lot of places were the same colours are being used but I thought it was a good step to get the right variables added to core first so all repos can start to swap out the old ones for the new ones.

@dhensby I will run a build of CSS to check.

clarkepaul:
@dhensby looks like the variables don't have any dist files

clarkepaul:
I've added a separate issue for removing and cleaning up all the related tree variables and hex colours scattered throughout here https://github.com/silverstripe/silverstripe-framework/issues/6295

I know its not going to be a 5min job so I'd prefer to do this work separately. I only needed this merged for the asset admin to be able to access the colours at this stage.

PR #6292, opened 2016-11-04
BUGFIX: Fixed crash when BASE_PATH is set by assuming the location of Constants.php

UndefinedOffset:
This pull fixes a crash when BASE_PATH is set by assuming the location of Constants.php, this appears to have been inadvertently caused by 7a10c194bdd751f74669c07836befecc1fde400e it wasn't taking into account the fact that the path now contains the src folder. Noticed in SilverStripe 4 Alpha 3.

### DIFF
--- a/src/Core/Constants.php
+++ b/src/Core/Constants.php
@@ -188,10 +188,10 @@ if(!defined('BASE_PATH')) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


dhensby:
thanks

PR #6296, opened 2016-11-07
Mark password fields as required if canBeEmpty is false

hailwood:
Because submitting the form to get told it's required is annoying when say the FirstName field told you straight away!

I'm still annoyed that we can't just chuck "Password" in a RequiredFields validator though...

### DIFF
--- a/forms/ConfirmedPasswordField.php
+++ b/forms/ConfirmedPasswordField.php
@@ -166,6 +166,9 @@ class ConfirmedPasswordField extends FormField {
 			/** @var FormField $field */
 			$field->setDisabled($this->isDisabled());
 			$field->setReadonly($this->isReadonly());
+			if(!$this->canBeEmpty && $field->getAttribute('required') === null){
+				$field->setAttribute('required', true);
+			}
 
 			if(count($this->attributes)) {
 				foreach($this->attributes as $name => $value) {


dhensby:
> I'm still annoyed that we can't just chuck "Password" in a RequiredFields validator though...

Can't we? What happens if you do?

forms/ConfirmedPasswordField.php
```diff
@@ -166,6 +166,9 @@ public function Field($properties = array()) {
 			/** @var FormField $field */
 			$field->setDisabled($this->isDisabled());
 			$field->setReadonly($this->isReadonly());
+			if(!$this->canBeEmpty){
+				$field->setAttribute('required', true);
```
dhensby: because this happens at rendertime it means the dev has no way of turning this off if they want to...

ie: it breaks the usecase of iterating over fields to disable the `required` attribute...

Can we check if it has or hasn't already been set?

hailwood: Patch incoming.

hailwood:
> > I'm still annoyed that we can't just chuck "Password" in a RequiredFields validator though...
> 
> Can't we? What happens if you do?

Nothing. I tried 'Password' Password[_Password] and Password_Password to trigger it.

dhensby:
> Nothing. I tried 'Password' Password[_Password] and Password_Password to trigger it.

That seems broken :/

dhensby:
I'd like someone else on the @silverstripe/core-team to give thoughts on this, please

kinglozzer:
Given that the field itself provides that validation already (via `$canBeEmpty`), I don’t think there’s anything we need to patch around `RequiredFields` for this. It should copy the `required`/`aria-required` attributes, though.

We should probably copy the approach in [`FormField::getAttributes()`](https://github.com/silverstripe/silverstripe-framework/blob/4aa50534d566722df81bfac8efd2c3db7ed92596/forms/FormField.php#L565-L568) for this, and only apply the `required` attribute if the field is present in `RequiredFields`. Something like:

```php
if(!$this->canBeEmpty && $this->Required()) {
	$field->setAttribute('required', 'required');
	$field->setAttribute('aria-required',  'true');
}
```

The way to opt-out is to then omit it from `RequiredFields`, same as other fields. Thoughts?

hailwood:
@dhensby If you're happy with the approach @kinglozzer has put forward I'll tests that `$this->Required()` is working as expected on this field and fire through that update.

dhensby:
Maybe it's just better that the `Required` method looks at `canBeEmpty` rather than having that logic in the field rendering method?

At what point does the `required` attribute normally get added? Does it consult the `Required` method? If so that could be a much simpler solution...

kinglozzer:
@dhensby It’s added here: https://github.com/silverstripe/silverstripe-framework/blob/4aa50534d566722df81bfac8efd2c3db7ed92596/forms/FormField.php#L565-L568. The trouble is that `ConfirmedPasswordField` contains nested fields, so we have to copy the attributes across to them manually.

dhensby:
ergh, the `ConfirmedPasswordField` is so broken :(

I think we should both overload `Required` method and we should push the attributes onto child fields.

hailwood:
@dhensby So you're suggesting we overload `Required` to return true if either the validator says it's required, or if `canBeEmpty` is false. Then we also overload getAttributes() to do the pushing of the required attribute to the child fields?

It's annoying - given the option I would remove `canBeEmpty` all together and just rely on the validator, but that would push this patch out to SS4 due to breaking the API.

Later I'm also firing through a PR to extract the logic if `$this->requireStrongPassword === true` to it's own function because having to copy the entire validate function because my client disagrees with what is considered a strong password is balls -- or should I just do that here as well?

dhensby:
Do the strong password thing separately just so we don't hold either up unintentionally. 

If you can patch required fields and that's your preferred method, then I'd tend to agree that's the best approach...

hailwood:
@dhensby,

I wasn't referring to patching `RequiredFields` to apply the required to subfields as we can't be sure what consequences that could have (e.g. a field that has two sub fields but if the field is required then the required attribute should only be added to one of those sub fields) so it would again be a breaking change.

I'll fire through an update a bit later with what I have in mind as we can always rollback.

Issue #5987, opened 2016-09-12
Centered image alignment crashes on save


jaydenseric:
In a `HTMLEditorField`, selecting `Centered, on its own.` for an image's alignment causes the following fatal PHP error on save:

```
PHP Fatal error:  Call to a member function appendChild() on null in /Volumes/Users/foo/Sites/bar/framework/View/Parsers/ShortcodeParser.php on line 187
```

All other alignments seem to save ok.

Observed in the latest dev-master.

clarkepaul:
I can confirm this

tractorcow:
A lot of painful research reveals the problem to be https://github.com/silverstripe/silverstripe-framework/pull/1190

The feature this change attempted to address no longer seems to be necessary; Tinymce shortcode handling has advanced quite considerably since 3.x.

I'm tempted to choose between a huge revert, or just dropping our core shortcode handler and dropping in something like https://github.com/thunderer/Shortcode

tractorcow:
It's hard to come up with a quick fix which isn't simply disable all dom element re-organisation.

sminnee:
Summarising a discussion offline with @tractorcow:

The DOM reorganisation code was designed for when shortcodes were inserted at illegal places in the DOM, eg:

```
<p>hello there</p>
[this_shortcode_generates_a_div]
<p>goodbye there</p>
```

Allegedly, more recent implementations of shortcodes in TinyMCE actually handle this by placing a `<div>` in the WYSIWYG editor and replacing it with the shortcode on save.

If this is the case (and needs some testing) then we could get away with a simple replacement-based shortcode handler, and we could move to the thunderer library.

/cc @Firesphere @hafriedlander

Firesphere:
Another option would maybe be to rewrite the current shortcode parser?

The downside would be, that we would need to maintain it ourselves. The upside would be it's probably easier to update and implement, than a third party module.

Digging deep into my memory, Thunderer wasn't really that easy to setup with SilverStripe. I may have some basic implementation lingering around in some unpushed branch, but it's not the holy grail I think.

thunderer:
Hi everyone, I'm the author of mentioned Shortcode library. I'd like to address some of the concerns raised here and in #5535. I'll link the comments and answer them here:

[...]

Issue #5535, opened 2016-05-14
Replace ShortcodeParser with external library


Firesphere:
Follow-up from #5487

Replace the current ShortcodeParser library with an external, maintained package. Up for the job so far, seems to be thunderer/Shortcode.

Firesphere:
quote @sminnee 
> And in _config.php:
> 
> ``` ShortcodeParser::get('default')->register('embed', array('EmbedShortcodeProvider', 'handleShortcode'))```
> 
> We can decide a better approach once we see what other shortcode engines expect.

I would actually like to have the shortcodes registered via the yml file eventually. That would save a lot I think. Any pros/cons?

sminnee:
> I would actually like to have the shortcodes registered via the yml file eventually. That would save a lot I think. Any pros/cons?

It makes sense, but I'd wait until we have found the right shortcode library to see what it expects.

sminnee:
`thunderer/Shortcode` looks like the only viable option, although it is a 1-person project. It's MIT licensed, which is compatible.

sminnee:
> From @hafriedlander: "However most of the code in SilverStripe's ShortcodeParser is actually to try and adjust the inserted content to still be HTML compliant, not actually parsing. (So if the shortcode tries to insert a p tag into an existing p tag it still works). It doesn't look like thunderer/shortcode attempts that - like most shortcode systems, it's content-type agnostic / ignorant.

(The initial use case, for those interested, was so a shortcode could insert a `<figure><img /><figcaption /></figure>' block which gets floated left or right - these need to be pulled out of any surrounding p tag or most browsers do the wrong thing)."

This situation is covered in `ShortcodeParserTest::testExtract()`. The behaviour is hard-coded to interrogate a `class` argument of any shortcode. It doesn't appear to be used by any core code, but I expect it's been used by modules. It would be an unnecessary regression to drop this functionality.

Firesphere:
From the looks of it, most frameworks actually implement a parser themselves, indeed thunderer seems the only viable option. Considering how much it's used, there are surprisingly few libraries available.

sminnee:
Having looked into this more, I have mixed feelings.

 * The thunderer module is of modest popularity and only has a single maintainer.
 * It's certainly got more features than our shortcode handler, but it lacks a key feature that ours has.
 * Our shortcode handler isn't a *massive* source of open issues - it looks like there are 6 or 8 open features

[...]

PR #6302, opened 2016-11-07
API Added extension points to ShortcodeParser

jonom:
Allows us to solve problems like https://github.com/sheadawson/silverstripe-shortcodable/issues/8

### DIFF
--- a/parsers/ShortcodeParser.php
+++ b/parsers/ShortcodeParser.php
@@ -21,6 +21,13 @@ class ShortcodeParser extends Object {
 
 	// --------------------------------------------------------------------------------------------------------------
 
+	/**
+	 * Registered shortcodes. Items follow this structure:
+	 * [shortcode_name] => Array(
+	 *     [0] => class_containing_handler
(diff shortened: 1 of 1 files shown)


jonom:
To complement this change, might be good to change a lot of the protected stuff to public so that extensions can make use of it. In particular there's no way at the moment to get info on registered shortcodes, data is stored in `protected $shortcodes = array()` with no getter. You can see if a shortcode is registered with `registered($shortcode)` but can't look up the class and method responsible for rendering it.

Would there be any drawback to opening this stuff up?

dhensby:
https://github.com/silverstripe/silverstripe-framework/issues/5535 and https://github.com/silverstripe/silverstripe-framework/issues/5987 are related.

I think adding some public getters makes sense and can go into the 3 branch too

jonom:
@dhensby have added a getter for registered shortcodes which seems like the most useful thing to expose for now and doesn't mess with visibility of existing methods.

jonom:
Thanks Dan

PR #6304, opened 2016-11-09
FIX: Correct return type for Member::currentUser()

kinglozzer:
See [this comment](https://github.com/silverstripe/silverstripe-framework/commit/995d07756d0b7aeef611d61fe3a31663b70e4c51#commitcomment-19754518).

### DIFF
--- a/security/Member.php
+++ b/security/Member.php
@@ -837,9 +837,8 @@ class Member extends DataObject implements TemplateGlobalProvider {
 		$id = Member::currentUserID();
 
 		if($id) {
-			return DataObject::get_by_id('Member', $id);
+			return DataObject::get_by_id('Member', $id) ?: null;
 		}
-		return null;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


dhensby:
i think this needs to go against 3.4, then I can merge 3.4 -> 3.5.0 -> 3

kinglozzer:
So, uh, changing the base branch didn’t quite go as expected... Lemme see if I can force push and fix it without opening a new PR 😝

kinglozzer:
Ah, that’s better

dhensby:
just rebasing and force pushing would work

security/Member.php
```diff
@@ -837,9 +837,9 @@ public static function currentUser() {
 		$id = Member::currentUserID();
 
 		if($id) {
-			return DataObject::get_by_id('Member', $id);
+			$member = DataObject::get_by_id('Member', $id);
```
dhensby: You could avoid this variable assignment and do `DataObject::get_by_id('Member', $id) ?: null;`

PR #6313, opened 2016-11-15
Update CheckboxSetField validate function

ben-w:
I have noticed that if user tick more than one checkbox in the CheckboxSetField, then the validation will always fail. After debugging, it is apparent that $this->value return from the CheckboxSetField is always a string, never an array. Therefore when user tick more than one checkboxes, the validation will always fail at 

if (!in_array($this->value, $validValues)) {
    ...
}

$this->value is now comma separated value. In my case, string(23) "TV,Online,Word of mouth".

Thank you for your time. 
Ben

### DIFF
--- a/forms/CheckboxSetField.php
+++ b/forms/CheckboxSetField.php
@@ -328,40 +328,46 @@ class CheckboxSetField extends OptionsetField {
 	 * @return bool
 	 */
 	public function validate($validator) {
-		$values = $this->value;
-		if (!$values) {
-			return true;
-		}
-		$sourceArray = $this->getSourceAsArray();
-		$validValues = array_keys($sourceArray);
-		if (is_array($values)) {
-			if (!array_intersect($validValues, $values)) {
-				$validator->validationError(
-					$this->name,
-					_t(
-						'CheckboxSetField.SOURCE_VALIDATION',
-						"Please select a value within the list provided. '{value}' is not a valid option",
-						array('value' => implode(' and ', $values))
-					),
-					"validation"
-				);
-				return false;
-			}
-		} else {
-			if (!in_array($this->value, $validValues)) {
-				$validator->validationError(
-					$this->name,
-					_t(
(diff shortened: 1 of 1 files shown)


wilr:
@ben-w the tests failing look directly related to this, could you please investigate and resolve.

ben-w:
Hi @wilr I had look the error, it says
"explode() expects parameter 2 to be string, array given"

This makes sense, as the that particular test is to test for array, hence the name testValidationWithArray.

However, what I am trying to say is, the value returned by the CheckboxSetField is never an array, in my case, it will always return string, not array. Please see attached screenshot, if I tick a single checkbox, var_dump the value to screen, it says string, if I tick more than one checkboxes, it will still return as string, a comma separated string.

screen shot 2016-11-15 at 3 51 59 pm
screen shot 2016-11-15 at 3 52 11 pm
screen shot 2016-11-15 at 3 52 25 pm
screen shot 2016-11-15 at 3 52 35 pm

The old validation code, checks for array and non-array value, but it will never go through the array validation, because $values = $this->value;  is always string, **_please have a look at the code (screenshot below)_**, then you will see what I am trying to say, see where I have added the var_dump, and highlight code will never get executed as the var_dump shows that $values is always a string.

screen shot 2016-11-15 at 4 01 15 pm

If $values = $this->value; is expecting an array or string depend on user selections, then perhaps somewhere else in the Framework is having a bug.

I hope this makes sense. Cheers

ben-w:
Hi @wilr , 
Please ignore this PR, I have done some additional test, after my last comment. It seems that this is coming from the UserForm module. I created a custom form which uses CheckboxSetField directly, all value return is actually array. So I will look into UserFormsCheckboxSetField bit more, before I jump into the conclusion. Really sorry about this.

Ben

kinglozzer:
Thanks for confirming @ben-w. I’ll close this PR for now, but feel free to re-open it if you need to 😃

PR #6315, opened 2016-11-15
API Namespace all tests

tractorcow:
Includes a bugfix to DataObject::dbObject() not setting the correct table for namespaced classes, and an API enhancement to SapphireTest to promote better fixture extensibility.

CMS requires https://github.com/silverstripe/silverstripe-cms/pull/1684 to pass.

In order to prepare for later psr-2 / psr-4 conformance, I have moved all classes to their own file, and grouped fixtures for any test under a sub-namespace of the same name.

E.g. `SilverStripe\ORM\Tests\DataObjectTest` fixtures will be in the `SilverStripe\ORM\Tests\DataObject` namespace. Apparently it's safe to have a namespace the same name as a class. :)

I've removed all `@package` / `@subpackage` since they are superceded by namespaces.


tractorcow:
image

PR #6316, opened 2016-11-15
Add OtherItemID as a schema param

wilr:
History Form needs both an ItemID and a VersionID. Follows the same naming as $ID/$OtherID available in other controllers.


wilr:
Needed for https://github.com/silverstripe/silverstripe-asset-admin/pull/306

tractorcow:
I would merge this, but now there are some CSS changes; I think we need to get the main PR approved before I can merge the framework one.

dhensby:
needs rebasing and rebuilding

PR #6317, opened 2016-11-15
Insert media modal in react

flamerohr:
This is part of the work to give us the new styled insert media modal.
It covers inserting images (embeds are not in this pull request)

For more details, see https://github.com/silverstripe/silverstripe-asset-admin/issues/173

Merge this afterwards: https://github.com/silverstripe/silverstripe-asset-admin/pull/307


admin/client/src/state/formData/FormDataReducer.js
```diff
@@ -0,0 +1,43 @@
...
+
+    case ACTION_TYPES.SET_SCHEMA_LOADING: {
+      return deepFreeze(Object.assign({}, state, {
+        [action.payload.id]: Object.assign({}, state[action.payload.id], {
+          loading: true,
+        }),
+      }));
+    }
+
+    case ACTION_TYPES.SET_SCHEMA_LOADED: {
```
chillu: Distinguishing between `SET_SCHEMA_LOADING` and `SET_SCHEMA_LOADED` seems a bit clunky, easy to misread - could you just pass through a boolean on `SET_SCHEMA_LOADING`?

admin/client/src/containers/FormBuilderLoader/FormBuilderLoader.js
```diff
@@ -161,27 +192,34 @@ FormBuilderLoader.propTypes = Object.assign({}, basePropTypes, {
...
+  const schema = state.schemas[ownProps.schemaUrl];
+  const form = schema && schema.id;
+
+  const reduxFormState = state.form && state.form[ownProps.schemaUrl];
+  const submitting = reduxFormState && reduxFormState.submitting;
+  const values = reduxFormState && reduxFormState.values;
+
+  const formData = state.formData[ownProps.schemaUrl];
+  const stateOverrides = formData && formData.stateOverride;
+  const loading = formData && formData.loading;
```
chillu: Why is the `loading` flag on `state.formData` rather than `state.schemas`? It seems to make it harder to iterate through actual form data, because there's metadata mixed in.

admin/client/src/state/formData/README.md
```diff
@@ -0,0 +1,9 @@
+# Form Data state
+
+Uses `formData` store.
+
+Manages state associated with the FormBuilderLoader.
+
+This sets form data that does not quite belong in schema, as it is client-side generated data, as opposed to server-side generated for schema.
```
chillu: Calling this `formData` is quite misleading, given that the *actual* form data (form field values) is located in `schema`. How about a state structure like `schema: { schema: {}, state: {}, metaData: {} }`? It's hard to name this, since it contains both metadata (`loading`) and data (`stateOverride`).

admin/client/src/legacy/HtmlEditorField.js
```diff
@@ -1566,4 +1582,274 @@ $.entwine('ss', function($) {
...
+
+    Data: {},
+
+    onunmatch() {
+      // solves errors given by ReactDOM "no matched root found" error.
+      this._clearModal();
+    },
+
+    _clearModal() {
+      ReactDOM.unmountComponentAtNode(this[0]);
```
chillu: Can you please run some heap analysis to ensure we don't cause a memory leak with this? Open the modal a few times, navigate to different pages, check if the memory use is roughly the same.

admin/client/src/containers/FormBuilderLoader/FormBuilderLoader.js
flamerohr: I don't think it should go in `state.schemas` that's from the server-side of things and gets overridden as soon as that part is received successfully.
Turns into a nightmare to manage the mingling between the two.

chillu:
@clarkepaul Did you leave out a "close" trigger on the "insert media" panel on purpose? It's a bit weird that you have to click on the ~10px wide background on either side as the only way to get out of the panel, right?

admin/client/src/legacy/HtmlEditorField.js
flamerohr: Memory usage has stayed relatively the same.
Usage jumps up before this was implemented with the entwine implementation, and the behaviour is roughly the same after this is implemented.

flamerohr:
@chillu I was discussing with @clarkepaul about this, at the moment the Modal close button is re-added and now it has a "Close Editor" button under the "Close Modal" button and seems confusing, we're going to try out hiding the "Close Editor" button.

clarkepaul:
@chillu  It was an oversight leaving it out of the designs, added in now in code.

chillu:
Looks good to me, thanks for the modifications Chris!

admin/client/src/state/formData/README.md
chillu: Just to close this off, Chris has now implement this as `metadata` and `stateOverride` keys.

PR #6322, opened 2016-11-20
[Minor] Split finding the member and authenticating the password

Firesphere:
This makes extending the Authentication from the MemberAuthenticator easier, for example if you do want SilverStripe to use the member, but want to validate the password against a different database.

The original authentication is unaffected and still works the same way. It's an implementation with "as few implications as possible"

There are obviously better methods, but this is a first step towards helping MFA modules bypass the SilverStripe authentication with their own.

### DIFF
--- a/src/Security/MemberAuthenticator.php
+++ b/src/Security/MemberAuthenticator.php
@@ -27,6 +27,74 @@ class MemberAuthenticator extends Authenticator {
 		'sha1' => 'sha1_v2.4'
 	);
 
+    /**
+     * @param array $data
+     * @return array(Member|null, boolean)
+     */
+    protected static function find_authentication_member($data)
+    {
+
+        // Attempt to identify by temporary ID
+        $member = null;
+        $email = null;
+
+        if(!empty($data['tempid'])) {
+            // Find user by tempid, in case they are re-validating an existing session
+            $member = Member::member_from_tempid($data['tempid']);
+            if($member) {
+                $email = $member->Email;
+            }
+        }
+
+        // Otherwise, get email from posted value instead
+        /** @skipUpgrade */
+        if(!$member && !empty($data['Email'])) {
+            $email = $data['Email'];
+        }
+
+        // Check default login (see Security::setDefaultAdmin())
+        $asDefaultAdmin = $email === Security::default_admin_username();
+        if($asDefaultAdmin) {
+            // If logging is as default admin, ensure record is setup correctly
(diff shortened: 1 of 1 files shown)


src/Security/MemberAuthenticator.php
```diff
@@ -38,54 +106,14 @@ class MemberAuthenticator extends Authenticator {
...
-			$result = new ValidationResult(false, _t('Member.ERRORWRONGCRED'));
-		}
+        /** @var Member $member */
+        list($member, $asDefaultAdmin) = static::find_authentication_member($data);
+        $success = static::authenticate_member_password($data, $member, $asDefaultAdmin);
 
 		// Emit failure to member and form (if available)
 		if(!$success) {
 			if($member) $member->registerFailedLogin();
-			if($form) $form->sessionMessage($result->message(), 'bad');
```
tractorcow: You lose the context of the $result validation.  This means any password validation custom message is hidden.

Maybe your authenticate_member_password could return a ValidationResult instead? You can always return a success result in place of `return true`.

src/Security/MemberAuthenticator.php
```diff
@@ -38,54 +106,14 @@ class MemberAuthenticator extends Authenticator {
...
+			if($form) $form->sessionMessage(_t('Member.ERRORWRONGCRED'), 'bad');
```
tractorcow: Can you please fill in the default localisation text for this? (even though I realise the original omitted it also).

Firesphere:
Yep, will do @tractorcow 

I'll also have a look at the test-errors. Those didn't show up locally, but as experience go, local and Travis often differ :)

It'll probably around thursday I get to this. But then again, it's not a big thing I guess?

Would you like separate tests for the member and the auth? Or just keep the current ones as they are? While I'm at it, might as well ask huh ;)

tractorcow:
You don't need to add tests if they are already covered by existing tests, although it would help boost the credibility of the PR if you added at least one new one. :)

src/Security/MemberAuthenticator.php
Firesphere: Yes (and yes, indeed it was just a copy paste :D )

chillu:
@Firesphere Is this still relevant with your auth refactor?

Firesphere:
@ingo Considering the WIP is close to it's final stage and will give way more flexibility and all..
Hah, I'm even mildly ashamed of this PR. Closing :D

PR #6325, opened 2016-11-22
API Refactor UploadField, FileField and AssetField into traits

tractorcow:
This is done in preparation for a new FileField which will be used within react, provided by the silverstripe/asset-admin module.

Behaviour which handled the upload of user-submitted files has been refactored into an "Uploadable" trait.

Additional behaviour which handled the upload of user-submitted files, specifically for saving into File dataobjects within has_one / has_many / many_many relations has been refactored into FileUploadable, which extends this base trait.

AssetField and UploadField now no longer extend FileField, but instead directly extend FormField, and have the necessary traits.

### DIFF
--- a/src/Forms/AssetField.php
+++ b/src/Forms/AssetField.php
@@ -26,9 +26,8 @@ use Exception;
  *  - Files can't be edited once uploaded.
  *  - Attached files can only be removed, not deleted.
  */
-class AssetField extends FormField
-{
-	use Uploadable;
+class AssetField extends FormField {
+	use UploadReceiver;
 
 	/**
 	 * @var array
@@ -158,14 +157,13 @@ class AssetField extends FormField
 	 * @param string $name The internal field name, passed to forms.
 	 * @param string $title The field label.
 	 */
-    public function __construct($name, $title = null)
-    {
+	public function __construct($name, $title = null) {
 		$this->addExtraClass('ss-upload'); // class, used by js
 		$this->addExtraClass('ss-uploadfield'); // class, used by css for uploadfield only
 
 		$this->ufConfig = array_merge($this->ufConfig, self::config()->defaultConfig);
 
-		$this->constructUploadable();
+		$this->constructUploadReceiver();
 		parent::__construct($name, $title);
 
 		// AssetField always uses rename replacement method
@@ -188,8 +186,7 @@ class AssetField extends FormField
 	 * @param string
 	 * @return $this
 	 */
-    public function setTemplateFileButtons($template)
-    {
+	public function setTemplateFileButtons($template) {
 		$this->templateFileButtons = $template;
 		return $this;
 	}
@@ -197,8 +194,7 @@ class AssetField extends FormField
 	/**
 	 * @return string
 	 */
-    public function getTemplateFileButtons()
-    {
+	public function getTemplateFileButtons() {
 		return $this->_templates($this->templateFileButtons, '_FileButtons');
 	}
 
@@ -207,8 +203,7 @@ class AssetField extends FormField
 	 *
 	 * @return boolean
 	 */
-    public function canPreviewFolder()
-    {
+	public function canPreviewFolder() {
 		if(!$this->isActive()) {
 			return false;
 		}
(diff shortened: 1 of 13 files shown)


dhensby:
I think the trait `Uploadable` could be named better. These objects are not "Uploadable" they are "upload handlers" no one can upload a `FileField` but they can upload to it.

tractorcow:
Ok, UploadHandler?

tractorcow:
Rebased and renamed. Uploadable -> UploadReceiver

dhensby:
`constructUploadable` :(

tractorcow:
oops

src/Forms/AssetField.php
```diff
@@ -27,7 +26,8 @@
  *  - Files can't be edited once uploaded.
  *  - Attached files can only be removed, not deleted.
  */
-class AssetField extends FileField {
+class AssetField extends FormField {
+	use Uploadable;
```
chillu: Having a separate "trait inheritance tree" doesn't really aid readability, but I can't think of an easy alternative (other than rewriting upload behaviour to use composition and a separate class, which would be quite intrusive)

src/Forms/FileUploadable.php
```diff
@@ -0,0 +1,399 @@
...
+ * Provides operations for reading and writing uploaded files to/from
+ * {@see File} dataobject instances.
+ * Allows writing to a parent record with the following relation types:
+ *   - has_one
+ *   - has_many
+ *   - many_many
+ * Additionally supports writing directly to the File table not attached
+ * to any parent record.
+ *
+ * @mixin FormField
```
chillu: The PHPDoc here should make it clearer that this trait expects to be applied to a `FormField` instance (the `@mixin FormField` just adds the IDE hints)

src/Forms/FileUploadable.php
```diff
@@ -0,0 +1,399 @@
...
+ */
+trait FileUploadable {
+	use Uploadable;
```
chillu: Having a separate "trait inheritance tree" doesn't really aid readability, but I can't think of an easy alternative (other than rewriting upload behaviour to use composition and a separate class, which would be quite intrusive)

src/Forms/FileUploadable.php
```diff
@@ -0,0 +1,399 @@
...
+	 * Each of these methods will update both the items (list of File objects) and the
+	 * field value (list of file ID values).
+	 *
+	 * @param array $value Array of submitted form data, if submitting from a form
+	 * @param array|DataObject|SS_List $record Full source record, either as a DataObject,
+	 * SS_List of items, or an array of submitted form data
+	 * @return $this Self reference
+	 * @throws ValidationException
+	 */
+	public function setValue($value, $record = null) {
```
chillu: This will silently fail if applied to a `FormField` which already implements `setValue()`, due to how [trait precedence](https://www.sitepoint.com/using-traits-in-php-5-4/) works. In the case of `AssetField-> setValue()`, it seems to be overruled on purpose - which makes this even more confusing. Same with `saveInto()`. Shouldn't this live on `UploadField` if its only used there? I'm starting to think the trait does more harm than good in this case: It avoid a bit of trivial duplication at the expense of a much less readable implementation.

chillu:
Sorry I commented on intermediary commits, shown as "chillu commented on outdated diff"

src/Forms/FileUploadable.php
chillu: OK, ignore what I said - `AssetField` doesn't use `FileUploadable`. It does show that the lack of symmetry in naming gets a bit confusing though (`UploadField` and `FileField` use `FileUploadReceiver`, `AssetField` uses `UploadReceiver`).

tractorcow: Yes, hence the division between fileuplodable / uploadable. :)

The best I could do is to maybe split out "uploadable" into an adjacent trait, rather than a nested one. However, you would still need to manually `use` it everywhere you used `FileUploadable`, so I don't see much of a use case where it would be of benefit.

tractorcow: Well, one is in framework, and the other in asset-admin. :) FileField is a field in it's own right, not a feature-crippled subclass of UploadField. The architecture I chose was specifically intended to combat "deep inheritance" that has made the whole field-formfield hierarchy unwieldy.

PR #6330, opened 2016-11-23
Replace hard-coded directory separators constant

assertchris:
This removes warnings when building the manifest cache in Windows environments.

### DIFF
--- a/core/manifest/ManifestCache.php
+++ b/core/manifest/ManifestCache.php
@@ -21,22 +21,22 @@ interface ManifestCache {
  */
 class ManifestCache_File implements ManifestCache {
 	function __construct($name) {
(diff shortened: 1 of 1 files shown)


tractorcow:
Good job @assertchris ; Don't forget to branch and pull from a fork next time. :)

assertchris:
@tractorcow I suggest more specificity in https://github.com/silverstripe/silverstripe-framework/blob/master/docs/en/05_Contributing/01_Code.md#editing-files-directly-on-githubcom, as I would consider the 6 lines change a "small change"...

tractorcow:
Yeah that's odd... I normally would expect that also; I guess that github knows you have permission, so it allows you to do it without a useful prompt? UX fail IMO right? :P

PR #6331, opened 2016-11-23
Replace hard-coded directory separators constant

assertchris:
This removes warnings when building the manifest cache in Windows environments.


tractorcow:
Restarting build; there were un-merged changes in CMS that needed merging to restore tests.

PR #6332, opened 2016-11-24
Apply email defaults to SS_LogEmailWriter

andrewandante:
After #5626 there will be no fallback for the "from" address on SS_LogEmailWriter. This PR will try to use SS_SEND_ALL_EMAILS_FROM as the "from" address in the first case, before falling back on the `send_from` address, then the `admin_email` address.

This means that the expected behaviour of SS_SEND_ALL_EMAILS_FROM is kept intact.

### DIFF
--- a/dev/LogEmailWriter.php
+++ b/dev/LogEmailWriter.php
@@ -60,7 +60,11 @@ class SS_LogEmailWriter extends Zend_Log_Writer_Abstract {
 		$formattedData = $this->_formatter->format($event);
 		$subject = $formattedData['subject'];
 		$data = $formattedData['data'];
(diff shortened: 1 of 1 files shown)


tractorcow:
@dhensby you'll need to manually merge this into rc4 :)

andrewandante:
I've updated the PR to also use the -f option as a first attempt, replicating the behaviour [in Mailer](https://github.com/silverstripe/silverstripe-framework/blob/3.5.0/email/Mailer.php#L225-L230). This was the actual cause of the problem I experienced on our AWS infrastructure.

dhensby:
OK - I'll handle manually merging this into the release

tractorcow:
ta!

Issue #4095, opened 2015-04-22
TinyMCE not able to modify props of embed media (e.g. YouTube videos) and inserts invalid HTML 


patricknelson:
**Issue description**
When editing existing YouTube videos, you will not see the caption (other attributes are retrieved properly). If you enter a caption at all and click "Update," instead of updating the existing elements properly, it will next a new set of elements and retain the existing caption, replacing the current image placeholder and adding the new caption. 

**Note:** Image editing is not affected by this, only `embed` media. I have devised a fix which follows a similar code as the existing code used to modify images, except it has been augmented to be consistent with the different type of media.

**Steps to reproduce**
1. Start with a fresh install of SilverStripe (either branch `3` or `3.1`)
2. Navigate to the homepage.
3. Start blank: Delete all content in the editor.
4. In the content editor, click the media icon, select "From the web" tab.
5. Paste this URL and click "Add url": https://www.youtube.com/watch?v=q9--wqX2TEw
6. Wait for it to finish loading (when you see the dimensions display) **without clicking** "Insert". **Side note:** You can still reproduce this bug if you clicked "Insert" at this stage. Just go back in, add a caption, update and then edit the caption again later.
7. Click the "Edit" drop-down and add your caption text (e.g. "Test caption").
8. Scroll down and click "Update"
9. Click the placeholder image that was inserted (to select it) and then click the media icon to edit.
10. **Note:** The Caption field is blank (**BUG 1**)
11. Enter a _different_ caption here (e.g. "Test caption 2") and click "Update"
12. If you review the code, it will show improperly nested `<div>` tags. (**BUG 2**)

I'll try to get a PR in soon to demonstrate my fix for this. 

**Markup before editing caption:**

``` html
<p>Welcome to SilverStripe! This is the default homepage. You can edit this page by opening <a href="admin/">the CMS</a>.
    You can now access the <a href="http://doc.silverstripe.org">developer documentation</a>, or begin
    <a href="http://doc.silverstripe.org/doku.php?id=tutorials">the tutorials.</a></p>
<div class="captionImage leftAlone" style="width: 480px;">
[...]

PR #6335, opened 2016-11-24
BUGFIX: Fixed issue where a shortcode's location would not get set to split when using the class leftAlone

UndefinedOffset:
In 3.1 to 3.4 at least when using the class or the location ``leftAlone`` in a shortcode the location does not get set to split. Which causes the html tag to not get split up correctly, this is simply caused by the wrong case being used in the check for leftAlone, the check is looking for ``leftALone`` instead.


UndefinedOffset:
I think this either partially fixes  #4095 maybe completely.

PR #6336, opened 2016-11-24
Making testing great again

### LINKED ISSUE #6306: [DEV] Travis builds fail on tags and non-numeric named branches
https://github.com/silverstripe/silverstripe-framework/pull/6283 has caused a regression where builds on tags and forked repos won't complete.

This prevents tests from being run on PR branches before opening a PR and certainty around stability of tags.

dhensby:
Fixes #6306

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ matrix:
 
 before_script:
  - export CORE_RELEASE=$TRAVIS_BRANCH
+ - "if [ \"$TRAVIS_PULL_REQUEST\" == false ]; then git config remote.origin.fetch +refs/heads/*:refs/remotes/origin/*; git fetch; for remote in `git branch -r`; do git branch --track ${remote#origin/} $remote; done; wget https://gist.githubusercontent.com/dhensby/e8c8979ac4fde4b7b348fcf8097df095/raw/bb44afc5ec8befc37777df664203af2d4f3372ff/dna-test.sh; chmod +x dna-test.sh; export CORE_RELEASE=$(./dna-test.sh \"$TRAVIS_BRANCH\"); fi"
  - composer self-update || true
  - phpenv rehash
  - git clone git://github.com/silverstripe-labs/silverstripe-travis-support.git ~/travis-support


sminnee:
I note that the bug is logged here: https://github.com/travis-ci/travis-ci/issues/4745

What happens if we use `CORE_RELEASE=TRAVIS_TAG` in these cases? My instinct about this change is that it's more effort than it "should" be an will lead to future maintainability regrets (see also: behat-extension, travis-support, ErrorControlChain)

If there was a simpler approach that used TRAVIS_TAG in place of TRAVIS_BRANCH when it was provided, and that led to functioning builds, that would be my preference.

dhensby:
This fix isn't designed to address testing tags, however there is a problem with travis-support module and tags - it treats tags as branches which means an invalid version constraint is used for tag test.

This fix is intended to allow feature branches (well, any branch that isn't named after a version) to run travis builds.

sminnee:
Would it be more appropriate to put `dna-test.sh` into the travis-support package?

dhensby:
> Would it be more appropriate to put dna-test.sh into the travis-support package?

I had thought about that, but @tractorcow informed me we are looking at deprecating that

SpiritLevel:
Like the title of this PR :)

sminnee:
@dhensby well we should be deprecating this too. ;-)

Is this code needed when we run tests directly from the travis checkout as we do in ss4 framework and cms? That is the marker point at which we deprecate travis support so if that fixes the problem, putting this into travis support would make sense.

I don't think the ss4 builds use core release any longer.

tractorcow:
I would think the only time we really need CORE_RELEASE is in module tests for multiple core versions, but even that I would like to replace with a new recipe system instead. (`composer require silverstripe/the-recipe 4.0.x-dev` instead of `CORE_RELEASE = master`).

dhensby:
This error only affects SS <4

This change is only needed because we are no longer "hard coding" `CORE_RELEASE` in our travis file (#6283). 

The assumption that the branch name is the same as the core release to test against breaks the build on any non numeric branch or tag. 

So, there are 2 options, revert above PR (which just means we need to be more careful when merging up the 3 line) or we take in something like this which will be unified but more complex. 

I'm happy with either and not particularly precious about this approach, it just keeps the intension of a unified travis file.

sminnee:
Refactor your solution into Travis-support. Although we're deprecating it for SS4, it will remain in use for the lifetime of SS3.

tractorcow:
👍 on moving this to travis support.

tractorcow:
Related : https://github.com/silverstripe/silverstripe-travis-support/issues/33

PR #6337, opened 2016-11-25
NEW replace _ss_environment.php with .env and environment vars

dhensby:
This PR replaces our idiosyncratic `_ss_environment.php` convention with support for first class environment variables with a [`.env`](https://github.com/vlucas/phpdotenv) poly-fill for easy testing/development.

I've also removed the dependency on the `Host` header (`$_SERVER['HTTP_HOST']`) from the bootstrap so we no longer require `$_FILE_TO_URL_MAPPING` global for CLI execution. Devs can now set `SS_HOST` environment variable or `Director.alternate_host` config var.

Docs and tests have been updated


assertchris:
Brilliant change. I'd suggest creating an `env($key, $default = null)` function which proxies to the underlying `env` function, but allows for defaults. Then things like `getenv('SS_PROTECTED_ASSETS_PATH')` can have sensible defaults (like `env('SS_PROTECTED_ASSETS_PATH', 'assets/secure')`).

tractorcow:
Could we not support both getenv() and defined() vars using a core environment helper?

```
class Environment {
	public static function get($var) {
		if (defined($var)) return constant($var);
		return getenv($var);
	}
}
```

This allows certain vars to be fixed (and thus protected from modification). there are security concerns related to allowing core vars to be modified.

dhensby:
@assertchris good idea - I guess I'll just copy laravel ;) - though there is more a favour towards putting these methods into utility classes.

@tractorcow whilst I don't really think there's a significant issue in allowing our pseudo environment vars to be modified I think your suggestion is sound in regards to providing BC support.

One issue there is to address is how we now handle `$_FILE_TO_URL_MAPPING` - though we could just define a new env var for this?

assertchris:
I think it would be better to resolve `.env` variables before constants, but good suggestion @tractorcow.

tractorcow:
I would prefer to see `$_FILE_TO_URL_MAPPING` become less important and opt-in, and instead have a better default.

dhensby:
> I would prefer to see $_FILE_TO_URL_MAPPING become less important and opt-in, and instead have a better default.

I agree - there's currently the ability to set default baseurl and we could easily move this to an env var itself. `.env` won't apply across sites, so there's less of a need for a path to domain map.

tractorcow:
My view is that if you declare security-sensitive vars such as SS_TRUST_* values, attempts to re-declare them via getenv() could potentially be an error condition (i.e. only if defined in multiple places AND differ).

Other values could be safely overridden, however, e.g for testing purposes. :)

I'm very security paranoid though... so I am perhaps arguing for unnecessary strictness in this case.

tractorcow:
> there's currently the ability to set default baseurl 

Maybe just have SS_DEFAULT_BASEURL instead as a substitute? Does that better suit the new architecture?

tests/bootstrap/environment.php
```diff
@@ -2,33 +2,37 @@
 
 // Bootstrap _ss_environment.php
```
SpiritLevel: // Bootstrap environment.php ?

src/Assets/Flysystem/ProtectedAssetAdapter.php
```diff
@@ -34,8 +34,8 @@ protected function findRoot($root) {
         }
 
         // Use environment defined path
-        if(defined('SS_PROTECTED_ASSETS_PATH')) {
-            return SS_PROTECTED_ASSETS_PATH;
+        if(getenv('SS_PROTECTED_ASSETS_PATH')) {
+            return getenv('SS_PROTECTED_ASSETS_PATH');
```
SpiritLevel: Isn't it more efficient to avoid calling `getenv` twice and instead do something like below?
```
if($protetedAssetsPath=getenv('SS_PROTECTED_ASSETS_PATH')) {
   return $protectedAssetsPath;
}

If then there are a number of other places one reduce the calls to `getenv` :)

File: src/Control/Director.php [hunk 3]
```diff
@@ -585,7 +585,7 @@ public static function is_https() {
 		// See https://support.microsoft.com/en-us/kb/307347
 		$headerOverride = false;
 		if (TRUSTED_PROXY) {
-			$headers = (defined('SS_TRUSTED_PROXY_PROTOCOL_HEADER')) ? array(SS_TRUSTED_PROXY_PROTOCOL_HEADER) : null;
+			$headers = (getenv('SS_TRUSTED_PROXY_PROTOCOL_HEADER')) ? array(getenv('SS_TRUSTED_PROXY_PROTOCOL_HEADER')) : null;
 			if (!$headers) {
```
SpiritLevel: Extra call to `getenv` ?

src/Control/HTTPRequest.php
```diff
@@ -708,7 +708,7 @@ public function allParsed() {
 	public function getIP() {
 		$headerOverrideIP = null;
 		if(TRUSTED_PROXY) {
-			$headers = (defined('SS_TRUSTED_PROXY_IP_HEADER')) ? array(SS_TRUSTED_PROXY_IP_HEADER) : null;
+			$headers = (getenv('SS_TRUSTED_PROXY_IP_HEADER')) ? array(getenv('SS_TRUSTED_PROXY_IP_HEADER')) : null;
 			if(!$headers) {
```
SpiritLevel: Extra call to `getenv` ?

src/Core/Constants.php
```diff
@@ -83,18 +40,18 @@
...
-		|| defined('SS_TRUSTED_PROXY_IPS')
+		|| getenv('SS_TRUSTED_PROXY_IPS')
 	) {
 		$trusted = false;
 
-		if(defined('SS_TRUSTED_PROXY_IPS') && SS_TRUSTED_PROXY_IPS !== 'none') {
-			if(SS_TRUSTED_PROXY_IPS === '*') {
+		if(getenv('SS_TRUSTED_PROXY_IPS') !== 'none') {
+			if(getenv('SS_TRUSTED_PROXY_IPS') === '*') {
 				$trusted = true;
```
SpiritLevel: Extra call to getenv ?

src/Core/Constants.php
```diff
@@ -167,8 +123,8 @@
 }
 
 // Filter by configured allowed hosts
-if (defined('SS_ALLOWED_HOSTS') && php_sapi_name() !== "cli") {
-	$all_allowed_hosts = explode(',', SS_ALLOWED_HOSTS);
+if (getenv('SS_ALLOWED_HOSTS') && php_sapi_name() !== "cli") {
+	$all_allowed_hosts = explode(',', getenv('SS_ALLOWED_HOSTS'));
 	if (!isset($_SERVER['HTTP_HOST']) || !in_array($_SERVER['HTTP_HOST'], $all_allowed_hosts)) {
```
SpiritLevel: Extra call to getenv ?

src/Core/Startup/ParameterConfirmationToken.php
```diff
@@ -167,7 +167,7 @@ protected function currentAbsoluteURL() {
 		// See https://support.microsoft.com/en-us/kb/307347
 		$headerOverride = false;
 		if(TRUSTED_PROXY) {
-			$headers = (defined('SS_TRUSTED_PROXY_PROTOCOL_HEADER')) ? array(SS_TRUSTED_PROXY_PROTOCOL_HEADER) : null;
+			$headers = (getenv('SS_TRUSTED_PROXY_PROTOCOL_HEADER')) ? array(getenv('SS_TRUSTED_PROXY_PROTOCOL_HEADER')) : null;
```
SpiritLevel: Extra call to getenv ?

src/Dev/Deprecation.php
```diff
@@ -147,8 +147,8 @@ public static function get_enabled() {
 		if(isset(self::$enabled)) {
 			return self::$enabled;
 		}
-		if(defined('SS_DEPRECATION_ENABLED')) {
-			return SS_DEPRECATION_ENABLED;
+		if(getenv('SS_DEPRECATION_ENABLED')) {
+			return getenv('SS_DEPRECATION_ENABLED');
 		}
```
SpiritLevel: Extra call to getenv ?

src/Dev/Install/config-form.html
```diff
@@ -135,8 +135,8 @@ <h3 class="sectionHeading">Database Configuration <small>Step 2 of 5</small></h3
 													$fieldType = ($fieldName == 'password') ? 'password' : 'text';
 													// values
 													$defaultValue = (isset($fieldSpec['default'])) ? $fieldSpec['default'] : null;
-													if($usingEnv && isset($fieldSpec['envVar']) && defined($fieldSpec['envVar'])) {
-														$value = constant($fieldSpec['envVar']);
+													if($usingEnv && isset($fieldSpec['envVar']) && getenv($fieldSpec['envVar'])) {
+														$value = getenv($fieldSpec['envVar']);
```
SpiritLevel: Extra call to getenv ?

src/Dev/Install/install.php5
```diff
@@ -142,8 +150,8 @@ if(isset($_REQUEST['db'])) {
 	if(isset($_REQUEST['db']['type'])) {
 		$type = $_REQUEST['db']['type'];
 	} else {
-        if( defined('SS_DATABASE_CLASS') ){
-            $type = $_REQUEST['db']['type'] = SS_DATABASE_CLASS;
+        if( getenv('SS_DATABASE_CLASS') ){
+            $type = $_REQUEST['db']['type'] = getenv('SS_DATABASE_CLASS');
```
SpiritLevel: Extra call to getenv ?

src/Dev/Install/install.php5
```diff
@@ -169,8 +177,8 @@ if(isset($_REQUEST['db'])) {
 		$databaseConfig['type'] = $type;
 	}
 } else {
-    if( defined('SS_DATABASE_CLASS') ){
-        $type = $_REQUEST['db']['type'] = SS_DATABASE_CLASS;
+    if( getenv('SS_DATABASE_CLASS') ){
+        $type = $_REQUEST['db']['type'] = getenv('SS_DATABASE_CLASS');
```
SpiritLevel: Extra call to `getenv` ?

src/conf/ConfigureFromEnv.php
```diff
@@ -55,100 +55,82 @@
...
 global $database;
 
 // No database provided
 if(!isset($database) || !$database) {
-	if(defined('SS_DATABASE_NAME')) {
-		$database = SS_DATABASE_NAME;
-	} else if(defined('SS_DATABASE_CHOOSE_NAME') && SS_DATABASE_CHOOSE_NAME) {
-		$loopCount = (int)SS_DATABASE_CHOOSE_NAME;
+	if(getenv('SS_DATABASE_NAME')) {
+		$database = getenv('SS_DATABASE_NAME');
```
SpiritLevel: Extra call to `getenv` ?

src/conf/ConfigureFromEnv.php
```diff
@@ -55,100 +55,82 @@
...
+	} else if(getenv('SS_DATABASE_CHOOSE_NAME')) {
+		$loopCount = (int)getenv('SS_DATABASE_CHOOSE_NAME');
 		$databaseDir = BASE_PATH;
```
SpiritLevel: Extra call to `getenv` ?

src/conf/ConfigureFromEnv.php
```diff
@@ -55,100 +55,82 @@
...
 			. $database
-			. (defined('SS_DATABASE_SUFFIX') ? SS_DATABASE_SUFFIX : ''),
+			. (getenv('SS_DATABASE_SUFFIX') ?: ''),
 	);
 
 	// Set the port if called for
-	if(defined('SS_DATABASE_PORT')) {
-		$databaseConfig['port'] = SS_DATABASE_PORT;
+	if(getenv('SS_DATABASE_PORT')) {
+		$databaseConfig['port'] = getenv('SS_DATABASE_PORT');
```
SpiritLevel: Extra call to `getenv` ?

src/conf/ConfigureFromEnv.php
```diff
@@ -55,100 +55,82 @@
...
 	}
 
 	// Set the timezone if called for
-	if (defined('SS_DATABASE_TIMEZONE')) {
-		$databaseConfig['timezone'] = SS_DATABASE_TIMEZONE;
+	if (getenv('SS_DATABASE_TIMEZONE')) {
+		$databaseConfig['timezone'] = getenv('SS_DATABASE_TIMEZONE');
```
SpiritLevel: Extra call to `getenv` ?

src/conf/ConfigureFromEnv.php
```diff
@@ -55,100 +55,82 @@
...
 	}
 
 	// For schema enabled drivers:
-	if(defined('SS_DATABASE_SCHEMA'))
-		$databaseConfig["schema"] = SS_DATABASE_SCHEMA;
+	if(getenv('SS_DATABASE_SCHEMA'))
+		$databaseConfig["schema"] = getenv('SS_DATABASE_SCHEMA');
```
SpiritLevel: Extra call to `getenv` ?

src/conf/ConfigureFromEnv.php
```diff
@@ -55,100 +55,82 @@
...
 
 	// For SQlite3 memory databases (mainly for testing purposes)
-	if(defined('SS_DATABASE_MEMORY'))
-		$databaseConfig["memory"] = SS_DATABASE_MEMORY;
+	if(getenv('SS_DATABASE_MEMORY'))
+		$databaseConfig["memory"] = getenv('SS_DATABASE_MEMORY');
```
SpiritLevel: Extra call to `getenv` ?

src/conf/ConfigureFromEnv.php
```diff
@@ -55,100 +55,82 @@
...
 }
 
-if(defined('SS_SEND_ALL_EMAILS_TO')) {
-	Email::config()->send_all_emails_to = SS_SEND_ALL_EMAILS_TO;
+if(getenv('SS_SEND_ALL_EMAILS_TO')) {
+	Email::config()->send_all_emails_to = getenv('SS_SEND_ALL_EMAILS_TO');
```
SpiritLevel: Extra call to `getenv` ?

src/conf/ConfigureFromEnv.php
```diff
@@ -55,100 +55,82 @@
...
 }
-if(defined('SS_SEND_ALL_EMAILS_FROM')) {
-	Email::config()->send_all_emails_from = SS_SEND_ALL_EMAILS_FROM;
+if(getenv('SS_SEND_ALL_EMAILS_FROM')) {
+	Email::config()->send_all_emails_from = getenv('SS_SEND_ALL_EMAILS_FROM');
```
SpiritLevel: Extra call to `getenv` ?

src/conf/ConfigureFromEnv.php
```diff
@@ -55,100 +55,82 @@
...
-		user_error("SS_DEFAULT_ADMIN_PASSWORD must be defined in your _ss_environment.php,"
+if(getenv('SS_DEFAULT_ADMIN_USERNAME')) {
+	if(!getenv('SS_DEFAULT_ADMIN_PASSWORD')) {
+		user_error("SS_DEFAULT_ADMIN_PASSWORD must be defined in your environment,"
 			. "if SS_DEFAULT_ADMIN_USERNAME is defined.  See "
 			. "http://doc.silverstripe.org/framework/en/topics/environment-management for more information",
 			E_USER_ERROR);
 	} else {
-		Security::setDefaultAdmin(SS_DEFAULT_ADMIN_USERNAME, SS_DEFAULT_ADMIN_PASSWORD);
+		Security::setDefaultAdmin(getenv('SS_DEFAULT_ADMIN_USERNAME'), getenv('SS_DEFAULT_ADMIN_PASSWORD'));
```
SpiritLevel: Extra call to `getenv` ?

src/conf/ConfigureFromEnv.php
```diff
@@ -55,100 +55,82 @@
...
 	}
 }
-if(defined('SS_USE_BASIC_AUTH') && SS_USE_BASIC_AUTH) {
-	BasicAuth::config()->entire_site_protected = SS_USE_BASIC_AUTH;
+if(getenv('SS_USE_BASIC_AUTH')) {
+	BasicAuth::config()->entire_site_protected = getenv('SS_USE_BASIC_AUTH');
```
SpiritLevel: Extra call to `getenv` ?

src/conf/ConfigureFromEnv.php
```diff
@@ -55,100 +55,82 @@
...
 }
 
-if(defined('SS_ERROR_LOG')) {
+if(getenv('SS_ERROR_LOG')) {
 	$logger = Injector::inst()->get('Logger');
 	if($logger instanceof Logger) {
-		$logger->pushHandler(new StreamHandler(BASE_PATH . '/' . SS_ERROR_LOG, Logger::WARNING));
+		$logger->pushHandler(new StreamHandler(BASE_PATH . '/' . getenv('SS_ERROR_LOG'), Logger::WARNING));
```
SpiritLevel: Extra call to `getenv` ?

dhensby:
> Maybe just have SS_DEFAULT_BASE_URL instead as a substitute? Does that better suit the new architecture? We have BASE_URL but that's a core constant not an environment-level option.

Agreed

dhensby:
@SpiritLevel I guess you're right, there are a lot of double calls to `getenv` which is because I just wrapped `defined` calls and bare constants with `getenv()`. That's because this was more a quick "lets see if the tests pass" thing at first (which they do, yay).

If I take @assertchris' advice about a helper (`getenv($val, $default)`) I could probably remove large chunks of this code.

SpiritLevel:
:)

I like the fact that `_ss_environment.php` is being renamed as a dot file `.env` so that it is hidden. What about doing the same for `silverstripe-cache` ? If I had my hands on the big lever, I think I would use the names `.ss-cache` and `.ss-environment` ;)

dhensby:
You can call either whatever you like. 

With `.env` you can add This to your _config.php:

```php
try {
	(new \Dotenv\Dotenv(BASE_PATH, '.ss-env'))->load();
} catch (\Dotenv\Exception\InvalidPathException $e) {
	// no file found - no big deal
}
```

And with the cache file:

```php
define('TEMP_PATH', BASE_PATH . '/.ss-cache');
```

tractorcow:
Just in case you thought otherwise, dot prefixed files aren't automatically hidden by web servers. it's not really any more secure.

However, I think we DO need a php env file support, as we have relied on this in several places. E.g. for splitting automatically maintained config from manually maintained config, or for conditional configs.

Could we not have separate support for `.env.yml` and `.env.php`?

I also have this problem with phpdotenv.

```
Add your application configuration to a .env file in the root of your project. Make sure the .env file is added to your .gitignore so it is not checked-in the code
```

For some applications we MUST have the enviroment variables outside of the web root. Having it outside of the application root is also preferred, even if the webroot is a sub-directory.

tractorcow:
Also, after re-reading this whole PR again for a second time I see a lot I missed earlier. Sorry for missing these earlier.

SpiritLevel:
Thanks @dhensby. I am sure I noticed that the `silverstripe-cache` folder name was hardcoded somewhere...maybe only as a default though. I can't remember..I'll check. And, since I opened on an efficiency note (:P), isn't the attempt to create a new object, and doing so within a try-catch block, a fairly expensive operation for just checking the existence of a file? A bit picky, I know :)

@tractorcow Yup. I know dot files are not hidden from access but only hidden from view in the file manager and from the `ls` command. It "feels" tidy to remove this noise and have only the essential files and folders visible. I was also going to raise the issue of `.env` being outside of webroot; I put passwords and keys in `_ss_environment.php` above the webroot and have no `_ss_environment.php` in the webroot. I am hoping to be able to do the same thing with `.env` since it appears from Dan's docs that I can create an "extra" `.env` outside the webroot and not have any `.env` in the webroot...is this possible Dan ?

SpiritLevel:
Oh no :(      

From https://github.com/vlucas/phpdotenv:

"phpdotenv is made for development environments, and generally should not be used in production. **In production, the actual environment variables should be set so that there is no overhead of loading the .env file on each request.** This can be achieved via an automated deployment process with tools like Vagrant, chef, or Puppet, or can be set manually with cloud hosts like Pagodabox and Heroku."

Might this be an issue to consider for production mode since you have `.env.` being loaded in `main.php` on each request ?

On the other hand, it looks like `new \Dotenv\Dotenv(PARENT_OF_BASE_PATH, '.ss-env'))->load();` will enable one to put `.ss-env` above the webroot, without having to define an "extra" `.env` file.

dhensby:
> However, I think we DO need a php env file support, as we have relied on this in several places. E.g. for splitting automatically maintained config from manually maintained config, or for conditional configs.

What kind of things are you thinking? something like:

```
if ($_SERVER['REMOTE_ADDR'] == [some ip]) {
    define('SS_ENVIRONMENT_TYPE', 'dev');
}
```

If so, I think this kind of config is really bad practice and something we should be discouraging. 

The aim of moving away from `_ss_environment.php` and to `.env` is to move away from SilverStripe's idiosyncrasies and to a more standardised approach. These constants are a way of doing environment variables without doing environment variables and that should change.



> For some applications we MUST have the enviroment variables outside of the web root

1. Then these variables should probably be real environment variables, not ones stored in a `.env` and polyfilled into environment vars
2. If the above isn't possible, you can move the `.env` to any location you like and pull it in like so:
```php
try {
	(new \Dotenv\Dotenv('/anywhere/I/like'))->load();
} catch (\Dotenv\Exception\InvalidPathException $e) {
	// no file found - no big deal
}
```

The approach to using real environment variables mitigates the need for a `.env` at all - the point of that file is just to make it a bit easier (no restarting apache) during development.

assertchris:
> the point of that file is just to make it a bit easier (no restarting apache) during development

I'd argue that's not the only or even the most important purpose of that file. It documents what environment variables are required (this is usually done with a key-only copy of the file, named `.env.example`) so that newcomers can know waht environment variables they need to set up before the application error log (or hours of debugging) tell them.

sminnee:
I like this change but I'd prefer _ss_environment.php to be deprecated rather than broken.

sminnee:
> Could we not support both getenv() and defined() vars using a core environment helper?

I think that making `getenv()` calls break is A Bad Idea. In moving away from `getenv()` and creating another abstraction we limit our ability to interoperate with the rest of the PHP ecosystem.

I'd do this by something along the lines of:

```php
if(defined('SS_DATABASE_USERNAME')) {
  if(getenv('SS_DATABASE_USERNAME')) {
    cryAndDie();
  }
  setenv('SS_DATABASE_USERNAME', SS_DATABASE_USERNAME);
}
```

This would also mean whitelisting the variables supported.

stojg:
The server guy is here!

I really like the move to using env variables instead of _SS_env. It will be much easier to deploy a silverstripe on different systems (locally, dockers, servers) and follows the 12-factor app methodology nicely.

But it will be tricky for me to build a production environment and knowing before hand if the SS app is going to be 3.0 or 4.0. So, in this case, I would probably duplicate the information into either .env or the webserver env AND _ss_environment.php. 

People might be jumping between branches when deploying to check SS 4 compatibility and I would rather not setup a specific environment for that.

So if both works, it would make my job much easier and cause less confusion during the transition. I would probably deprecate _ss_env.php with the goal of removing support in the next major version of SS

sminnee:
I guess another way to approach it would be to have a shim module that took all the .env content and turned it into `define()`s?

cli-script.php
```diff
@@ -82,23 +82,20 @@
 // Connect to database
 if(!isset($databaseConfig) || !isset($databaseConfig['database']) || !$databaseConfig['database']) {
 	echo "\nPlease configure your database connection details.  You can do this by creating a file
-called _ss_environment.php in either of the following locations:\n\n";
-	echo " - " .  BASE_PATH  . DIRECTORY_SEPARATOR . "_ss_environment.php\n - ";
-	echo dirname(BASE_PATH) . DIRECTORY_SEPARATOR . "_ss_environment.php\n\n";
+called .env in " . BASE_PATH;
```
sminnee: Can we support .env files that aren't in the PATH, please?

 * I have a `~/Sites` folder with 50-odd projects and a single `_ss_environment.php` file. This is a helpful use case
 * We have a lot of sites where `BASE_PATH` is `/sites/something/www` and `_ss_environment.php` is in `/sites/something`. There are a number of benefits to this approach and I wouldn't want to make it impossible.

sminnee:
Related: https://github.com/vlucas/phpdotenv/issues/213

micmania1:
@dhensby do you have xhprof results of before and after (with & without concurrency)? Does this add any overhead if we're running it in prod?

The reason I can see this being a poorer choice over _ss_env is that you lose the benefit of opcache and you'd have to parse the .env file on every request as @SpiritLevel mentioned. 

The benefit I can see is that you could load them into your env vars earlier in your server config and they'd be shared, but some people don't have the ability to configure their server in that way.

cli-script.php
tractorcow: In addition to this, it would be nice if I could split "server level" and "site level" specific configs. E.g. `~/Sites/.env` has my global settings, but I can override specifics with `~/Sites/mysite/.env`.

tractorcow:
```php
if(defined('SS_DATABASE_USERNAME')) {
  if(getenv('SS_DATABASE_USERNAME')) {
    cryAndDie();
  }
  setenv('SS_DATABASE_USERNAME', SS_DATABASE_USERNAME);
}
```

👍  for this idea; Some vars should be allowed to be made immutable, so this addresses it acceptably.

SpiritLevel:
@dhensby you are getting lots of feedback on this one! 

Following on from @micmania1, and the conversations [here](https://github.com/vlucas/phpdotenv/issues/207) and [here](https://github.com/vlucas/phpdotenv/issues/76#issuecomment-87252126), it seems that for production mode it would be a good idea to cache the environment vars loaded from `.env` upon the first request and then use these cached values for all subsequent requests; apparently, this is what Laravel does. And, for test and dev modes, insist that the `.env` file is read every request so that changes take immediate effect, without having to update the cache.  :) I also note there are some security concerns discussed in those above-linked conversations. There is no spoon but there are lots of rabbit holes...

dhensby:
OK - thanks for the feedback, I'm glad to see that this is generally being met with approval. There are some concerns being raised too, which I think I'll summarise as such:

1. No support for "server" wide `.env` file
2. No conditional logic available in `.env` file
3. We shouldn't make the upgrade path to 4 harder than it needs to be and removing `_ss_environment.php` is a too much

So, before I get into responding to these concerns I'd like to lay a few of the objectives of this PR so hopefully we can all consider it from the same standpoint and there's no ambiguity about what we are trying to achieve:

- Standardise SilverStripe so it's more inline with the [12 factor app](https://12factor.net/config)
- Remove (one of) SilverStripe's idiosyncrasies to become more inline with the wider PHP ecosystem
- Remove code complexity and maintenance burden

Also, it's important to remember that `.env` is basically a development environment shim for real environment variables and is not intended for prod.

### 1. No support for "server" wide `.env` file

This is possible and can be done by adding your own snippet as I mentioned [here[(https://github.com/silverstripe/silverstripe-framework/pull/6337#issuecomment-263873656). However, if you have environment variables that are shared amongst your entire server then this sounds like a perfect example of when a environment variable would be best.

### 2. No conditional logic available in `.env` file

I view this as bad practice and only limited to a few valid use cases (all pretty much on development environments):

- Setting a DB name based on site being accessed (this can be done with `SS_DATABASE_CHOOSE_NAME` instead)
- Setting `$_FILE_TO_URL_MAPPING` - we've somewhat agreed this should be deprecated

### 3. Reducing upgrade friction

I agree this will be a major change and so we should probably provide some compatibility for the old files *and* constants to be used but these should definitely be triggering deprecation warnings to encourage people to move away from the _ss_environment.php file.

In the interest of reducing code complexity and maintenance this should be pretty simple poly fill IMO. Something like:

```php
$envFileLocation = findEnvirontmentFileLocation();
if ($envFileLocation) {
    throwDeprecationWarning();
}
$constants = get_defined_constants();
include $envFileLocation;
$newConstants = array_diff_key($constants, get_defined_constants());
foreach ($newConstants as $constant => $value) {
    if (getenv($constant) === null) {
        loadConstantIntoEnvironment($constant, $value);
    }
    else {
        trigger_error('Environment variable "' . $constants . '" already set', E_USER_WARNING);
    }
}
```

If there are other concerns, please raise them and I'll look at incorporating them into the PR.

dhensby:
The more I think about this the less I think providing BC is valuable:

> I have a ~/Sites folder with 50-odd projects and a single _ss_environment.php file. This is a helpful use case

You'll only ever be setting a `.env` on 4.0 sites, it's not like your 50 sites are suddenly going to break because 4.0 doesn't use `_ss_env` anymore. If there's a concern about when you *do* have 50 SS 4 sites, then these shared vars should be first class environment variables on your machine rather than in the `.env` per site.

In SS 4 we're doing a lot of coard cutting in regards to old APIs and I'm not sure why this should be one we retain when it provides a difficult reminder of the past.

sminnee:
If we don't have support for the _ss_environment file in SS4, then we can put both files side by side on a server. That seems of benefit to me and so I'd support that.

I'm not sure what it means for people who have written code that relies on defined constants in their modules, however. As you say, perhaps the upgrader can help, but I'm not sure how it would get a list of defined constants to rewrite.

dhensby:
> If we don't have support for the _ss_environment file in SS4, then we can put both files side by side on a server. That seems of benefit to me and so I'd support that.

This is the conclusion I'm coming to.

As for constants from thirdparty modules, I'm afraid they'd have to switch to env vars or use some custom config option - which will just be as much of the pain as the rest of it, the upgrader can do what I've done in my latest commit - look at the constants immediately added by the `_ss_environment.php` and write them to `.env`

tractorcow:
> 1. No support for server wide .env file
> This is possible and can be done by adding your own snippet as I mentioned here. However, if you have environment variables that are shared amongst your entire server then this sounds like a perfect example of when a environment variable would be best.

I would rather that this was included on-install rather than requiring manual code to enable. Several reasons:

# Developer bootstrapping:

My reasoning is that it should be able to quickly composer create-project a new site without having to go through all the "extra set up steps" in order to get a site running. I never ever need to touch my _ss_environment.php file in root, and I'd hope to not have to do too much additional work in 4.0 either.

I see that you can use global environment variables, but those are nowhere near as flexible or easy to update as a text file. (no rebooting, editing system config files, restarting apache).

# Source control management

We should never have environment configuration in source control. That also means not mixing uncontrolled environment files mixed with other source controlled files in the same directory.

# Security

Config vars, for a similar reason to above, should not be anywhere near web-accessible folders.

# Upgrade paths

Not supporting .env in places where users once relied on _ss_environment.php being accessible means additional upgrade steps.

I would propose something like this is main.php:

```php
// load site-specific config before parent config
$paths = [ BASE_PATH, dirname(BASE_PATH) ];
foreach ($paths as $path) {
	if (file_exists($path . '/.env') {
		(new \Dotenv\Dotenv(BASE_PATH))->load($path, '.env');
	}
}
```

dhensby:
@tractorcow from where I sit what this PR lets us do is almost perfectly analagous to using the `_ss_environment.php` the only difference is where this config sits (a file in a parent directory or a settings file creating real env vars). All of the points you raise are either true now or with this PR or are not problems that should be addressed in this PR.

> I would rather that this was included on-install rather than requiring manual code to enable. Several reasons:

There will be no manual code to enable if you have your environment pre-configured (as is the case now) othwise you'll need to perform a one-time configuration.

> My reasoning is that it should be able to quickly composer create-project a new site without having to go through all the "extra set up steps" in order to get a site running. I never ever need to touch my _ss_environment.php file in root, and I'd hope to not have to do too much additional work in 4.0 either.

This is exactly the same now as it is with this PR. I don't see that there are any extra steps if your server is pre-configured. I also don't follow why if you "never ever need to touch [your environment] file" why a one time set up of server environment variables is seen as such an inconvenience - it adds no extra steps or hassle to the `composer create-project` example and it allows you to put project related config where it belongs (with the project).

> We should never have environment configuration in source control. That also means not mixing uncontrolled environment files mixed with other source controlled files in the same directory.

This isn't a problem for this PR to solve - it seems to be something to address not only in the SilverStripe eco system but PHP and development as a whole. It is something [we do quite extensively](https://github.com/silverstripe/silverstripe-installer/blob/master/.gitignore). The PHP community seems to agree that excluding files using `.gitignore` seems to be somewhat normal and accepted as a means of allowing (sensitive) files to exist on the file system without being added to the VCS.

> Not supporting .env in places where users once relied on _ss_environment.php being accessible means additional upgrade steps.

Yes, this is true - but almost everything we're doing in 4 is adding upgrade steps. I've addressed this in more detail in an [earlier comment](https://github.com/silverstripe/silverstripe-framework/pull/6337#issuecomment-266071543). There will be additional upgrade steps just by virture of including this PR in whatever form.

## Your proposal

If that is the clincher, then I'll make that change though I think it's bad for several reasons:

1. It breaks from the norm of using a `.env` which is confusing to developers and separates our project from the wider PHP community. This retains some idiosyncrasy that we're trying to remove.
2. We have to check one file that we shouldn't be checking as it is, now we'll be checking two. At least with the current set up the idea is that the `_ss_environment.php` should exist so we can justify looking for it, in prod the `.env` should not exist.

Maybe we should wrap the `.env` searcher in `if (getenv('SS_ENVIRONMENT_TYPE') == 'live')`

micmania1:
@dhensby how does this affect performance (see comment: https://github.com/silverstripe/silverstripe-framework/pull/6337#issuecomment-264036980)?

Also, what I haven't seen addressed is that dotenv is not intended to be used in production but it seems your proposal would do that? See comment by @SpiritLevel here which links to some good resources: https://github.com/silverstripe/silverstripe-framework/pull/6337#issuecomment-264050784

tractorcow:
Ok, I'm happy my complaints have been addressed. :) I understand there are still a few other things to consider before merging this PR, but I'll give my approvel. :P

dhensby:
@micmania1 I think there's a performance gain here because we're doing less disk reads to find environment files. If it's not used in production I expect it to be faster than what we currently have - though we're probably talking intangible differences here.

As per my last comment: "Maybe we should wrap the .env searcher in if (getenv('SS_ENVIRONMENT_TYPE') == 'live')" - however I've decided against this - whilst users *shouldn't* use it in prod, shared hosts may have now way at all to set env vars and so this is the last resort and not something I think we should block devs from doing.

Whilst we'd like to enforce best practices, being overly ideological isn't a good thing and more likely to turn people off than see the benefits of doing this properly. Ultimately if the cost is half a millisecond for those "doing it wrong" then that's the choice they've got.

dhensby:
OK - things left to address:

[ ] Replacing `$_FILE_TO_URL_MAPPING`
[ ] PR for https://github.com/silverstripe/silverstripe-installer

Please let me know if there's anything else to add

tractorcow:
Looks good @dhensby :)

tractorcow:
Milestoning at alpha 5, since we need to get alpha 4 out today.

tractorcow:
You could allow your `.env` file to declare a custom php script for inclusion.

`SS_EXTRA_INCLUDE=/var/sites/extra_include.php`

And you can define any non-trivial bootstrapping in there.

tractorcow:
Add to that, SS_EXTRA_INCLUDE could include non-php source files (such as other `.env`s).

dhensby:
@tractorcow I think the `SS_EXTRA_INCLUDE` is flawed for a few reasons - mostly the fact that it can only be set once is the biggest issue.

At the moment we can resolve these problems without needing to modify this PR.

dhensby:
@micmania1 I've also added a new commit that allows a first class env var to be set (`SS_IGNORE_DOT_ENV`) which will skip the loading of environment vars from the `.env` file. This will prevent the needless search for the file and thus mitigate any performance impact of doing so.

Whilst `.env` *shouldn't* be used on live, I don't think we should explicitly block it because

1. It's impossible to reliably determine what environment type we are on this early in the bootstrap
2. It would be useful for those on hosting infrastructure that doesn't allow the setting of first class ENV vars
3. The way this library is integrated will not override existing env vars (so there's little risk in allowing it on live environments)

tractorcow:
> @tractorcow I think the SS_EXTRA_INCLUDE is flawed for a few reasons - mostly the fact that it can only be set once is the biggest issue.


Discussed this with @dhensby and we agreed that it can be added back in with a module, so it doesn't need to be a part of core.

docs/en/05_Contributing/05_Making_A_SilverStripe_Core_Release.md
```diff
@@ -36,30 +36,28 @@ As a core contributor it is necessary to have installed the following set of too
...
+    # DB Credentials
+    SS_DATABASE_CLASS="MySQLDatabase"
+    SS_DATABASE_SERVER="127.0.0.1"
+    SS_DATABASE_USERNAME="root"
+    SS_DATABASE_PASSWORD=""
     
-    // Each release will have its own DB
-    define('SS_DATABASE_CHOOSE_NAME', true);
+    # Each release will have its own DB
+    SS_DATABASE_CHOOSE_NAME="ru"
```
wernerkrauss: typo? should that be "true" instead?

tractorcow:
@dhensby can we use `.env` to set the hostname of the server for CLI mode? It would be a bit more flexible than a hard coded array of phony headers.

docs/en/05_Contributing/05_Making_A_SilverStripe_Core_Release.md
dhensby: whoops

dhensby:
> @dhensby can we use .env to set the hostname of the server for CLI mode? It would be a bit more flexible than a hard coded array of phony headers.

Yes, using `SS_HOST`

tractorcow:
Yes but, what I mean is set that against $_SERVER. At the moment $_SERVER['HOST_NAME'] and getenv('SS_HOST') contradict one another. Unless I've mis-read the code.

main.php
```diff
@@ -71,6 +70,21 @@
 		exit(1);
 	}
 }
+// Allow a first class env var to be set that disables .env file loading
+if (!getenv('SS_IGNORE_DOT_ENV')) {
```
sminnee: Why has this moved from Constants.php to main.php? It's a lot less DRY. You've also forgotten to include it in cli-script.php.

My recommendation would be to shift the env load back to Constants.php after BASE_PATH is defined.

dhensby: Hmm - I had moved this for a reason - but I think that was to do with the host issue that I've now fixed. I'll move it back to Constants.php and see if I remember why I moved it :)

chillu: Could we use `SS_USE_DOT_ENV` instead? I find double negations a bit icky

sminnee: Disagree. We want .env files to be loaded by default, and I strongly disagree that `SS_USE_DOT_ENV = null` will have the opposite behaviour of `SS_USE_DOT_ENV = false`. That's way ickier.

dhensby: We can't have it the other way because this is designed for when setting env vars is tricky, so requiring an env var to be set to load the polyfill would be self-defeating.

chillu: Agree, `null` vs `false` is worse than what we have at the moment, didn't consider the execution order.

chillu:
Looks good! I'd like to see more upgrading notes to make the transition smoother:

 * Needs to link to more detailed docs, and an example on how to rewrite the old format to the new one. 
 * Should mention that devs need to check if their webservers are configured to deny requests to dot-prefixed files if the environment file is located in the webroot (or move it out of the webroot). This wasn't a problem before with PHP files, since they would execute on webserver requests, but not leak any information.
 * Needs to point out conditional logic as bad practice (no longer possible), and suggest a better alternative
 * Mention the removal of `$_FILE_TO_URL_MAPPING` (and how to achieve this now)

For `SS_HOST`, this now requires a project specific `.env` file, correct? It's probably not a big deal, since I can't recall a case where the mapping is actually relevant. I've used it for Behat at some point, but it's best practive to configure the hostname through `behat.yml` anyway.

src/Control/Director.php
```diff
@@ -551,6 +567,49 @@ public static function absoluteURL($url, $relativeParent = self::BASE)
         return Controller::join_links($parent, $url);
     }
 
+    public static function host()
```
tractorcow: Could use some PHPDoc, but not a blocker.

dhensby: added

main.php
dhensby: Moved into `Constants.php`

dhensby:
> Needs to link to more detailed docs, and an example on how to rewrite the old format to the new one.

I've added docs under environment management that show the new format. I've added to the changelog to link to this.

> Should mention that devs need to check if their webservers are configured to deny requests to dot-prefixed files if the environment file is located in the webroot (or move it out of the webroot). This wasn't a problem before with PHP files, since they would execute on webserver requests, but not leak any information.

Added security issues section to environment management docs

> Needs to point out conditional logic as bad practice (no longer possible), and suggest a better alternative

I'm not sure where to put this, it doesn't really make sense to put it in environment management docs ("Don't do this thing you can't do any more coz it's bad but if you have to do it this way").

> Mention the removal of $_FILE_TO_URL_MAPPING (and how to achieve this now)

Added to change log

chillu:
> I'm not sure where to put this, it doesn't really make sense to put it in environment management docs ("Don't do this thing you can't do any more coz it's bad but if you have to do it this way").

In the `4.0.0.md` upgrading guide - doesn't make sense as permanent documentation, but at the point of an upgrade it's relevant.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1327,11 +1328,23 @@ handle field-level and form-level messages. This has the following properties:
...
+### <a name="overview-environment-management"></a>Environment management
+
+See [Environment Management docs](https://docs.silverstripe.org/en/4/getting_started/environment_management/) for full details
+
+#### Environment file changes
+
+* Removed support for _ss_environment.php in favour of .env and first class environment variables
+* Environment variables now can be set in `.env` file placed in webroot or one level above
+* Environment variables will be read from the environment as well
+* `$_FILE_TO_URL_MAPPING` has been removed and replaced with using `Director.aleternate_host` or `SS_HOST` env var
```
chillu: Spelling ("aleternate")

dhensby: What's wrong with my aleternate spelling :)

docs/en/00_Getting_Started/03_Environment_Management.md
```diff
@@ -125,3 +78,13 @@ This is my `_ss_environment.php` file. I have it placed in `/var`, as each of th
...
+| `SS_PROTECTED_ASSETS_PATH` | Path to secured assets - defaults to ASSET_PATH/.protected |
+| `SS_DATABASE_MEMORY` | Used for SQLite3 DBs |
+| `SS_TRUSTED_PROXY_PROTOCOL_HEADER` | Used to define the proxy header to be used to determine HTTPS status |
+| `SS_TRUSTED_PROXY_IP_HEADER` | Used to define the proxy header to be used to determine request IPs |
+| `SS_TRUSTED_PROXY_HOST_HEADER` | Used to define the proxy header to be used to determine the requested host name |
+| `SS_TRUSTED_PROXY_IPS` | IP address or CIDR range to trust proxy headers from |
+| `SS_ALLOWED_HOSTS` | A comma deliminated list of hostnames the site is allowed to respond to |
+| `SS_MANIFESTCACHE` | The manifest cache to use (defaults to file based caching) |
+| `SS_IGNORE_DOT_ENV` | If set the .env file will be ignored. This is good for live to mitigate any performance implications of loading the .env file |
+| `SS_HOST` | The hostname to use when it isn't determinable by other means (eg: for CLI commands) |
```
sminnee: This doesn't appear to let us force the generation of `https://` URLs from CLI scripts. Perhaps we want to allow `SS_HOST` to be a URL fragment as well as a hostname?

sminnee: This could be a future PR.

docs/en/05_Contributing/05_Making_A_SilverStripe_Core_Release.md
```diff
@@ -36,30 +36,28 @@ As a core contributor it is necessary to have installed the following set of too
...
-    // So you can test releases
-    define('SS_DEFAULT_ADMIN_USERNAME', 'admin');
-    define('SS_DEFAULT_ADMIN_PASSWORD', 'password');
+    # So you can test releases
+    SS_DEFAULT_ADMIN_USERNAME="admin"
+    SS_DEFAULT_ADMIN_PASSWORD="password"
     
-    // Basic CLI hostname
+    # Basic CLI hostname
     global $_FILE_TO_URL_MAPPING;
```
sminnee: This is out of date.

src/Core/Constants.php
```diff
@@ -202,6 +85,23 @@
...
     define('BASE_PATH', $candidateBasePath);
 }
+
+// Allow a first class env var to be set that disables .env file loading
+if (!getenv('SS_IGNORE_DOT_ENV')) {
+    foreach (array(
+                 BASE_PATH,
+                 dirname(BASE_PATH),
+             ) as $path) {
+        try {
```
sminnee: Not a blocker, but I feel that an `if(file_exists())` check would be better than throwing and swallowing an exception.

docs/en/00_Getting_Started/03_Environment_Management.md
sminnee: We've also lost the ability to indicate that a SilverStripe site is running in a non-root URL.

Was there a reason that we didn't make this setting `SS_BASE_URL` instead?

dhensby: hmm - good point

docs/en/00_Getting_Started/03_Environment_Management.md
```diff
@@ -1,112 +1,65 @@
...
+Sensitive credentials should not be stored in a VCS or project code and should only be stored on the environment in 
+question. When using live environments the use of `.env` files is discouraged and instead one should use "first class"
+environment variables.
+
+If you do use a `.env` file on your servers, you must ensure that external access to `.env` files is blocked by the
+webserver.
+
+## Managing environment variables with `.env` files
+
+By default the `.env` must be placed in your webroot. If this file exists, it will be automatically loaded by the
```
dhensby: done

docs/en/05_Contributing/05_Making_A_SilverStripe_Core_Release.md
dhensby: fixed

dhensby:
OK - I've addressed all the feedback apart from the URL issues

sminnee:
I'm happy for us to merge as-is and pick up SS_HOST vs SS_BASE_URL as a separate ticket.

src/Core/Constants.php
tractorcow: Exception is a bit of overhead, we can fix in separate ticket.

dhensby:
:') Can't believe it - whoop

dhensby:
I've opened the follow-up issue

chillu:
Awesome to see this merged! Buuuuut: It broke *all* travis builds *apart* from framework, since that's the only one that no longer relies on travis-support: https://github.com/silverstripe/silverstripe-travis-support/issues/41. And funnily enough, we're using conditional logic in the _ss_env there ;) @dhensby, do you have time to look at this?

dhensby:
I'll look at it today.

GrahamCampbell:
The current version of phpdotenv is robust and thread-safe. The following example will work in a multithreaded environment. This avoids using the adapter that would have called `putenv` and `getenv`, which are not threadsafe.

```php
<?php

use Dotenv\Environment\Adapter\EnvConstAdapter;
use Dotenv\Environment\Adapter\ServerConstAdapter;
use Dotenv\Environment\DotenvFactory;
use Dotenv\Dotenv;

$factory = new DotenvFactory([new EnvConstAdapter(), new ServerConstAdapter()]);

Dotenv::create($path, null, $factory)->load();
```

robbieaverill:
Thanks for the update @GrahamCampbell!

@silverstripe/core-team do we want to do an RFC to re-implement the latest version of phpdotenv for SS5?

dhensby:
New issue / thread please :)

maxime-rainville:
I second Dan. Going through dozen of messages from 2 years ago is not conductive to a useful discussion.

GrahamCampbell:
See https://github.com/silverstripe/silverstripe-framework/issues/8764. :)

PR #6339, opened 2016-11-28
API Deprecate unused / undesirable create_new_password implementation

tractorcow:
This method isn't used by any core components any more, and has been since superseded by more secure behaviour.


tractorcow:
Could we merge this?

Issue #6096, opened 2016-09-26
Automated conversion to PSR-1/2 (whitespace, newlines)


chillu:
Our goal for SilverStripe core is PSR-1/2 compliance, eventually. We're much closer now that we have namespaces. Consistent newlines and whitespace is the next step. It's going to be painful at any time, since it'll cause merge conflicts in existing PRs. Merging up from 3.x is already difficult with all the file renames from the namespace work. @sminnee and myself think that now is as good a time as any to perform the automated conversions via http://cs.sensiolabs.org/ over all core modules.

Excludes (can be done later):
- Automating checks via PHP_CodeSniffer or Scrutinizer.
- Changing static method names to `studlyCase()` (5.0 change?)
- Line-based opt-outs (https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-parts-of-a-file)

Reference:
- http://www.php-fig.org/psr/psr-1/
- http://www.php-fig.org/psr/psr-2/

@silverstripe/core-team Any strong objections to this?

sminnee:
This command does a pretty good job of it:

```
phpcbf --standard=PSR2 --sniffs=Generic.Files.ByteOrderMark,Squiz.WhiteSpace.SuperfluousWhitespace,Generic.Files.LineEndings,Generic.WhiteSpace.DisallowTabIndent,Generic.WhiteSpace.ScopeIndent ORM/ Security/ View/ Forms/ Core/ i18n/ 
```

sminnee:
Having run that, this command reports no error (and does report errors prior to the change)

```
phpcs --standard=PSR2 --sniffs=Generic.Files.ByteOrderMark,Squiz.WhiteSpace.SuperfluousWhitespace,Generic.Files.LineEndings,Generic.WhiteSpace.DisallowTabIndent,Generic.WhiteSpace.ScopeIndent ORM/ Security/ View/ Forms/ Core/ i18n/ 
```

sminnee:
FYI I got the sniffs values by reading https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/PSR2/ruleset.xml and https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/PSR1/ruleset.xml

sminnee:
I recommend a PR that includes:
 * Whitespace fixes (1 commit)
 * 2nd commit with new tooling:
[...]

PR #6340, opened 2016-11-28
API Apply psr-2 coding conventions to framework with linting CI check

tractorcow:
Initial commit of https://github.com/silverstripe/silverstripe-framework/issues/6096

Note that all non-automatic fixes have been excluded from this initial build; Manual resolution of all outstanding linting issues are outside of the scope of this introductory pull request.

As requested by @sminnee I've split into two commits, one is an automatic cleanup and the other a linting check.

I have also excluded tests from the current linting, but the plan is to eventually fix and add this folder too.

PHPCS_TEST is run in PHP7 build to ensure optimal coverage of all language constructs.


tractorcow:
PHPCS_TEST is green. :) At least, once I muted all the known broken rules :P

sminnee:
My preference would have been for the whitespace changes to have been applied in their own commit, but it's not a merge blocker if that would be more than an hour's rework.

tractorcow:
All of the automatic fixes are essentially whitespace-only changes. I ran the upgrade again last night with your suggested commands and there was very little difference in output.

I don't think phpcbf can easily fix non-whitespace changes without manual input. I may need to double check though.

sminnee:
The other thing I suggest is setting up a custom script in composer.json so that you can lint by running `composer run-script lint`.

This will make it easier to lint locally without having to copy/paste a command out of .travis.yml.

tractorcow:
Done, and CI uses the composer script to run linting.

PR #6342, opened 2016-11-29
BUG Fix behat tests unable to capture HTML editor fields

tractorcow:
This fix resolves issues where behat tests were capturing the incorrect "Content" field. The basic `findField` method was pulling out the field from the search form in some cases. E.g.

`<input name="q[Term]" class="text" id="Form_SearchForm_q_Term" type="text">`

Note: Not all behat issues are fixed with this patch, but current errors are minimised.

Will need merging up after merge.


tractorcow:
Note that there are some further issues yet to resolve. I don't plan on doing it in this pull request, but I will follow up at a later date.

PR #6345, opened 2016-11-29
FIX: File::find() should respect case-sensitive filesystems properly

madmatt:
Prior to this fix, if you two files with the same case-insensitive filename (e.g. test.jpg and test.JPG), these would be stored correctly on the filesystem and in the database, but when `File::find('test.JPG')` was called, it would return the first matching `File` object, which might point at test.jpg instead of test.JPG.

`File::find()` now properly respects case-sensitive filesystems. This change does mean that you can't search the database in a case-insensitive manner for files, but I can't see a good use-case for this, even of case-insensitive filesystems.

### DIFF
--- a/filesystem/File.php
+++ b/filesystem/File.php
@@ -247,7 +247,7 @@ class File extends DataObject {
 		foreach($parts as $part) {
 			if($part == ASSETS_DIR && !$parentID) continue;
 			$item = File::get()->filter(array(
-				'Name' => $part,
+				'Name:ExactMatch:case' => $part,
 				'ParentID' => $parentID
 			))->first();
 			if(!$item) break;
--- a/tests/filesystem/FileTest.php
+++ b/tests/filesystem/FileTest.php
@@ -276,6 +276,19 @@ class FileTest extends SapphireTest {
 		$this->assertEquals("93132.3 GB", File::format_size(100000000000000));
 	}
 
+	/**
+	 * @covers File::find()
+	 */
+	public function testFind() {
+		$file1 = $this->objFromFixture('File', 'lowercase');
+		$file2 = $this->objFromFixture('File', 'uppercase');
+		$errorMsg = 'File::find() is not finding files in a case-sensitive manner';
+
+		$this->assertSame($file2->ID, File::find($file2->Filename)->ID, $errorMsg);
+		$this->assertSame($file1->ID, File::find($file1->Filename)->ID, $errorMsg);
+
+	}
+
 	public function testDeleteDatabaseOnly() {
 		$file = $this->objFromFixture('File', 'asdf');
 		$fileID = $file->ID;
--- a/tests/filesystem/FileTest.yml
+++ b/tests/filesystem/FileTest.yml
@@ -30,6 +30,14 @@ File:
     Filename: assets/FileTest-folder1/File1.txt
     Name: File1.txt
     ParentID: =>Folder.folder1
+  lowercase:
+    Filename: assets/FileTest-folder1/casetest.txt
+    Name: casetest.txt
+    ParentID: =>Folder.folder1
+  uppercase:
+    Filename: assets/FileTest-folder1/Casetest.TXT
+    Name: Casetest.TXT
+    ParentID: =>Folder.folder1
 Permission:
   admin:
     Code: ADMIN
--- a/tests/filesystem/FolderTest.php
+++ b/tests/filesystem/FolderTest.php
@@ -26,7 +26,7 @@ class FolderTest extends SapphireTest {
 		$file1 = $this->objFromFixture('File', 'file1-folder1');
 
 		$children = $folder1->allChildren();
-		$this->assertEquals(2, $children->Count());
+		$this->assertEquals(4, $children->Count());
 		$this->assertContains($subfolder1->ID, $children->column('ID'));
 		$this->assertContains($file1->ID, $children->column('ID'));
 	}


dhensby:
> but I can't see a good use-case for this, even of case-insensitive filesystems.

Surely an insesitive search on an insensitive system is the exact use-case?

Also, will a CMS user know that their file name is "test.JPG" and not "test.jpg". If I was searching for files and I searched "test" I'd expect to see all files named "test" regardless of case.

I think a better fix is for the CMS to detect collisions using case-insensitivity, ie: we don't allow to files to have the same name.

Just because some file-systems are case sensitive doesn't mean our assets store needs to be.

dhensby:
Also, I assume this test would now break on a case insensitive filesystem...

madmatt:
I did quick test on a (very slow) Windows machine, and also on my Mac which is running a case-insensitive filesystem (default for Macs).

In both of these cases, when uploading a file with the same name but different case (e.g. test.jpg and Test.JPG), on case-insensitive filesystems SilverStripe correctly detects that a file already exists with that name and prompts you to overwrite / rename. Therefore these filenames are no longer the same, and the old `File::find()` code would still work (so this fix only affects case-sensitive filesystems, case-insensitive filesystems are unaffected by both the bug *and* my fix as far as I can tell).

Furthermore, `File::find()` isn't used in the CMS when attempting to find files, it's designed as a method for developers to find a `File` object based on a path. The only use in the CMS as of 3.4.0 is `SiteTreeLinkTracking` and `HtmlEditorField` in framework.

I see there's a failing test, I'll fix that up for `Folder` now (looks to be an issue with sharing fixtures between `FileTest` and `FolderTest`).

tractorcow:
We cannot try to support both case sensitive and case-insensitive behaviour in framework. I propose the most robust solution, which is case insensitivity, even if the underlying filesystem supports sensitivity. I'm afraid this change would most likely, as @dhensby suggested, make case-insensitive filesystems more unstable, such as OSX / Windows, which are the majority of development environments.

madmatt:
So how do you fix that then? If the issue goes away in SS4 due to using flysystem etc. then maybe it's not a massive deal, but as far as I can tell this works fine on both case-sensitive and case-insensitive file systems. The fix I've made doesn't even matter to case-insensitive filesystems, because files that are uploaded are stored in the DB with the correct case (that matches the filesystem), it's just that on case-sensitive filesystems when two files have the same name except for case, the previous code would pick the wrong file 50% of the time.

How else do you propose fixing this? Always enforcing unique file & folder names, even on filesystems than support case-sensitivity? I'm all for that, but that's going to be a pain to actually implement, because `file_exists()` doesn't support passing a `$assumeCaseInsensitive` flag, so on a case-sensitive FS you'd have to check any variation of case? I don't think I could support having `Upload` only check the database to see if a file exists before saving it, that seems like a worse backwards step given the number of times files are stored in `assets` that aren't in the DB (or aren't unlinked when they're deleted for whatever reason). I see it on a lot of projects, despite the fact that it shouldn't happen...

Ultimately, this fix appears to work correctly on case-insensitive filesystems for me, as well as on a case-sensitive one (where it fixes the bug that was previously occurring). FWIW, the bug I was fixing was due to a file being returned by `secureassets` that wasn't supposed to be, and that's a potentially large security issue. I can't see a better way around this except for hacking secureassets instead to always check case, but that still leaves a bug in this method.

tractorcow:
> Prior to this fix, if you two files with the same case-insensitive filename (e.g. test.jpg and test.JPG), these would be stored correctly on the filesystem and in the database

Purely my opinion here, but this seems to be the failure of the detection of existing uploads. What should happen is that when you upload "test.JPG" it should pop up a warning in your uploadfield that an existing file is there, and it can be either overwritten or renamed.

I'm not sure the problem is fixed in 4.0.

Another possible, and more complete yet complex solution, is to look at detecting case sensitivity of the underlying filesystem and having the framework adjust accordingly, or in some cases allowing it to be forced to one or the other.

madmatt:
Sure, but how do you actually detect that this newly-uploaded file collides with another file?

On a case-insensitive filesystem, it's easy, because `file_exists($filename)` will return true for test.jpg vs. Test.JPG.

But on case-sensitive filesystem the easiest way I can see to do it is a loop over the output of `scandir($dir)`, and comparing e.g. `strtolower()` of the existing files vs. the new filename. That just seems dumb when the filesystem is actually case-sensitive and is totally fine with files that are named the same except for their case.

I guess if we want to 'dumb down' the interface to the lowest common denominator (case-insensitive) then it makes sense, but that's a much larger pull request.

Sorry - I just can't see why this is such a bad change. As far as I can tell it doesn't cause any issues with case-insensitive filesystems, and fixes a real bug on case-sensitive filesystems. If the issue here is "this is a hack and we should store all files case-insensitive" then that's fine, but that's a problem for 4.x or 3.6.0, and I'd argue this patch is still good regardless, to ensure you specify filenames correctly.

Would it help if I changed it so that `File::find()` does the existing broken (IMO) implementation, except if you pass a second `$strict = true` argument, where it matches names strictly? Then we'd have to make a PR on securefiles as well, but at least that way there's no potential for breaking existing sites calling `File::find()` with an incorrectly-cased filename?

madmatt:
(I see builds are failing, but I don't want to figure out what's wrong with behat until we can agree on how it's best to solve this bug) :)

tractorcow:
> Sure, but how do you actually detect that this newly-uploaded file collides with another file?

As a fallback to file_exists, we could use a case-insensitive DB search (similar to the code that you are above using) and use that to detect any existing files.

madmatt:
Yeah, that was the other way I thought of (and mentioned in an earlier comment), but in my experience with SS sites there are a lot of sites that place files in assets that don't get added to the database (or vice versa - add a file but then remove the DB record without removing the file).

Obviously this isn't a good idea and it's not necessarily supported, but on the other hand I'd point out that you're talking about performing a check against something that isn't the canonical source of truth (the canonical source would be the filesystem itself).

What about the idea to introduce a `$strict` param to `File::find()` and leave it false by default? That way, things that actually care about case-sensitivity (like secureassets, where it serves the wrong file otherwise) can opt-in to it, but we don't break BC at all. I'm not well versed on semver, can we add new optional params without breaking semver? I think so, right?

tractorcow:
> What about the idea to introduce a $strict param to File::find() and leave it false by default? 

I think this is a better idea... based on what you WANT the result of this method to do, you might need to do different things.

E.g. when checking if a file name is safe to write to, maybe it should be case-insensitive. When it's doing validation to check the db record matches the physical file, maybe it should be case-sensitive?

dhensby:
How has the system got to the point of having 2 files that are the same name but have different casing if the upload system should detect these collisions anyway?

Is the bug here that `find` is case insensitive or that your filesystem is in an unexpected state?

It is tricky to solve this problem - especially in a patch release and without re-thinking how assets work. Other approaches I can think of would be to ensure all files are named in lowercase on upload - the problem here is that whatever has caused you to have 2 same named files is likely to mean that not all files will be forced to lowercase on upload. I think it's completely unavoidable that if the filesystem and DB are out of sync that we will encounter inconsistent 

I'm not a fan of adding a `$strict` flag and this change is a behaviour change, so we'd have to be cautious about which release it goes into... It may be best to patch secureassets first and then we think of a long-term fix for core.

tractorcow:
> How has the system got to the point of having 2 files that are the same name but have different casing if the upload system should detect these collisions anyway?

This is my point, is that we should be case-insensitive when detecting clashes.

madmatt:
The upload system as far as I know doesn't detect these collisions. If it's supposed to, then it fails.

**Case-sensitive filesystem:**
 - I upload `test.jpg` fine
 - Now I upload `Test.JPG`, it works fine (because `file_exists('Test.JPG')` returns false I assume)

**Case-insensitive filesystem:**
 - I upload `test.jpg` fine
 - I now upload `Test.JPG` and get prompted to either overwrite or rename.

IMO, this approach above is perfectly fine, and I don't think we should change it. If your filesystem allows you to have case-sensitive filenames then we should allow it.

There is **no way** that we should change files to be stored in lowercase as floated earlier - there is a much higher chance that this breaks something for someone who is expecting filenames in a certain format.

I still think the best solution here, if we don't want to just fix this bug and move on, is to introduce a new `$strict = false` param to `File::find` and use that instead.

So probably tomorrow I'll force-push a change to do just that, and then we can discuss that commit instead :)

dhensby:
> If your filesystem allows you to have case-sensitive filenames then we should allow it.

I disagree because this breaks portability of your website which is one of the core aims of a framework (to work out the inconsistencies between environments). One should be able to pick up an SS site, move it to their local dev environment, a test server or a new prod server and not encounter issues that are environment specific.

Just like we support 5.3+ we support case insensitive file systems. That means if you move a site from a sensitive to insensitive system the site should work, just like if you move it from 5.6 to 5.3 it should work.

tractorcow:
Sorry @madmatt but I'm going to have to close this as rejected. I'm happy to re-visit this issue again in the future, but the consensus is no to forcing case sensitivity for file searches.

PR #6348, opened 2016-11-29
ENHANCEMENT Prevent test DBs persisting after testing

tractorcow:


dhensby:
Hmm - keeping test DBs on failure can be a useful debugging tool - would it be better to introduce a task one could run manually - or maybe only kill the DB if there's no failure?

tractorcow:
We have a task, but then you have to manually run it yourself. A good testing environment leaves no additional artifacts to clean up; Scouts rule.

dhensby:
Can we pass any additional flags to phpunit like `--keep-db` ?

tractorcow:
Possibly! it might be a bit more architecture to build in cli args -> environment args.

sminnee:
> Hmm - keeping test DBs on failure can be a useful debugging tool - would it be better to introduce a task one could run manually - or maybe only kill the DB if there's no failure?

This was never the intended use-case, this PR just fixes a bug. 👍  to thies patch from me, assuming that the tests pass.

`--keep-db` could be an extra feature but given that this patch fixes a bug in the current system I'm inclined to merge this one first.

tractorcow:
Tests are green so . :D

PR #6351, opened 2016-11-30
Update Versioned.php

steve-silverstripe:
Change isset() to array_key_exists(). In the CMS, if you have a FormField such as 'My value', then you change it to '' (blank), isset() will return false because even though the key exists, the value === null by the time it reaches this point in the code, and isset() return false on null values. array_key_exists will return true with a null value

All FormFields with a blank value such as empty string get converted to null in DBField.php in the function writeToManipulation()

### DIFF
--- a/model/Versioned.php
+++ b/model/Versioned.php
@@ -607,7 +607,7 @@ class Versioned extends DataExtension implements TemplateGlobalProvider {
 				$data = array_intersect_key($data, $fields);
 
 				foreach ($data as $k => $v) {
-					if (!isset($newManipulation['fields'][$k])) {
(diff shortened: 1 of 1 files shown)


tractorcow:
This fix was previously applied to 4.0, but it's good to back port it this way. You'll need to fix the tests too.

See https://github.com/silverstripe/silverstripe-framework/pull/5735

tractorcow:
@dhensby thanks

dhensby:
During the merge commit I made sure to add the missing tests and fix the issues. I basically back-ported the code in 4.x

tractorcow:
I noticed that. :D I was going to complain, but I dug a bit and saw your extra work. Hence the thanks. :)

PR #6356, opened 2016-12-01
API Refactor form validation handling

### LINKED ISSUE #6228: Don't use session and FormSchema to manage server-side React validation responses
## Acceptance Criteria
- Models can throw ValidationExceptions to do model validation, and the form system consistently uses ValidationResult for managing validation results (#3937 is reviewed and merged)
- React form submissions return their validation errors in direct response the submission HTTP request without modifying the session
- The validation error response is derived directly from a ValidationResult
- You don't need to re-send the entire FormSchema back to the client
- There is no stateful value of "isValid" stored on a form; forms are neither valid nor invalid, only their submissions are
- In addition to validation responses, it should be possible for form submissions to supply modified form state. When the form state is the same as what the client has supplied, this would be redundant, but for server-modified data (e.g. ensuring that a URLSegment is unique) this would be of use.

Tasks:

 - [X] Rebase #3937
 - [x] Code Review / Upgrade
 - [x] LeftAndMain::getSchemaResponse() to support ValidationResponse input. Update React form handling to support response.
 - [x] X-FormSchema-Request option for "messages" or "schema" for messages-only or full schema response options.
 - [x] Fix / Create Tests
 - [x] Upgrading docs
 
Pull requests:
 - [ ] https://github.com/silverstripe/silverstripe-framework/pull/6356
 - [ ] https://github.com/silverstripe/silverstripe-siteconfig/pull/43
 - [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/322
 - [ ] https://github.com/silverstripe/silverstripe-cms/pull/1691

tractorcow:
Resolves #6228
Includes rebase of #3937 as a separate commit

Merge with:
 - [ ] https://github.com/silverstripe/silverstripe-siteconfig/pull/43
 - [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/322
 - [ ] https://github.com/silverstripe/silverstripe-cms/pull/1691

The major features of this change:

 - When a form fails validation in a react admin area, it's possible for only error messages to be sent in the response, meaning simpler and more lightweight http responses.
 - If a ValidationException is thrown in any form action (e.g. in a DataObject::save) then these will now be safely caught by the Form and will be setup as nice error messages.
 - ValidationResult is now serialiseable 
 - Form messages can now be individually cast as either html / plain text
 - Form and FormField message methods moved into a `FormMessage` trait since the code was duplicated.
 - Form::setValidationResponseCallback callback now takes a parameter to the ValidationResult containing any errors.
 - Cleanup, cleanup, cleanup


src/Forms/Schema/FormSchema.php
```diff
@@ -2,10 +2,10 @@
...
 namespace SilverStripe\Forms\Schema;
 
-use SilverStripe\Control\Session;
 use SilverStripe\Forms\CompositeField;
 use SilverStripe\Forms\Form;
 use SilverStripe\Forms\FormField;
+use SilverStripe\ORM\ValidationResult;
 
 /**
  * Represents a {@link Form} as structured data which allows a frontend library to render it.
```
sminnee: This comment isn't really correct. FormSchema doesn't represent a form, because it doesn't any state. It's more like a service/factory that builds JSON representations of a Form Schema on demand. It's not related to this PR but it confused me a bit.

admin/code/LeftAndMain.php
```diff
@@ -356,37 +355,40 @@ protected function getSchemaResponse($form, $id = null) {
...
+		$auto = in_array('auto', $schemaParts);
 
-		$return = ['id' => $id];
+		// Require ID
+		if (empty($schemaID)) {
+			throw new InvalidArgumentException("schemaID is required");
+		}
+		$return = ['id' => $schemaID];
 
-		if (in_array('schema', $schemaParts)) {
```
sminnee: I wonder if this logic should be refactored out of `LeftAndMain` and into `FormSchema`, as it's likely to be a part of any interaction between a rich JS form and a supporting controller.

Something like this:

```php
return $this->schema->getMultipartSchema($schemaID, $schemaParts, $form, $errors);
```

Also I think that 'result' would be a better name than 'errors' for component generated by the ValidationResult, as a ValidationResult might also say "valid".

tractorcow: Yes I think so. Will do that. :)

tractorcow:
@sminnee see https://github.com/silverstripe/silverstripe-framework/pull/6356/commits/368a7f0644a5f2264ceb7b2fb676b5bdeb3f0c01 :D

admin/code/LeftAndMain.php
tractorcow: https://github.com/silverstripe/silverstripe-framework/pull/6356/commits/368a7f0644a5f2264ceb7b2fb676b5bdeb3f0c01

src/Forms/Schema/FormSchema.php
```diff
@@ -14,6 +16,47 @@
...
+	 * Returns a representation of the provided {@link Form} as structured data,
+	 * based on the request data.
+	 *
+     * @param HTTPRequest $request
+	 * @param string $schemaID ID for this schema. Required.
+	 * @param Form $form Required for 'state' or 'schema' response
+	 * @param ValidationResult $result Required for 'error' response
+	 * @return array
+	 */
+	public function getMultipartSchema($request, $schemaID, Form $form = null, ValidationResult $result = null) {
```
chillu: I think this particular method shouldn't concern itself with *how* the schema parts are requested - if we switch this to GraphQL sometime, it likely would be contained in the GraphQL query itself, rather than `$request`. Maybe we could accept an array of parts here instead, and have another helper method to extract that from a `HTTPRequest`?

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1095,6 +1095,55 @@ Some methods on `Requirements` have had their method signatures changed:
...
+  persisted in the state by default, unless a redirection occurs. You can also save any response in the
+  state by manually invoking `saveFormState` inside a custom validation response handler.
+* `setupFormErrors` renamed to `restoreFormState`
+* `resetValidation` renamed to `clearFormState`
+* `loadMessagesFrom` method created to load a ValidationResult into a form.
+* `setMessage`. third parameter is now $cast type
+* `messageForForm`. removed
+* `getSessionValidationResult` / `setSessionValidationResult` used to get / set session errors
+* `getSessionData` / `setSessionData` used to get / set field values cached in the session
+* `getAjaxErrorResponse` and `getRedirectReferer` created to simplify `getValidationErrorResponse`
```
sminnee: We could also make them `private`.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1095,6 +1095,55 @@ Some methods on `Requirements` have had their method signatures changed:
...
+`ValidationResult` now has these methods:
+
+* `serialize` / `unserialize` for saving within session state
+* `messageList` renamed to `getMessages`
+* `getFieldMessages` created as a subset of `getMessages` for fields only.
+* `getOverallMessages` created as a subset of `getMessages` for non-fields only.
+* `codeList` renamed to `getMessageCodes`
+* `message` renamed to `joinedMessage`
+* `joinedOverallMessage` added.
+* `starredList` renamed to `starredMessage`
```
sminnee: Agreed. It's a hangover from the kitchen-sink mentality that SilverStripe is moving away from.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1095,6 +1095,55 @@ Some methods on `Requirements` have had their method signatures changed:
...
+* `setMessage` to assign a message, type, and cast
+* `getMessage` retrieves the message string
+* `getMessageType` retrieves the message type (E.g. error, good, info)
+* `getMessageCast` retrieves the cast type
+* `getMessageCastingHelper` retrieves the DBField cast to use for the appropriate message cast
+* `getSchemaMessage` encodes this message for form schema use in ReactJS.
+
+`Form` methods have been changed:
+
+* `validate` method now returns a `ValidationResult` instance. This is no longer automatically
```
sminnee: Perhaps we want to:

 * Create a 2nd method with a new name. `getValidatonResult()`?
 * Rewrite `validate()` as a wrapper around this method that returns a boolean and throws a deprecation?

sminnee:
> Nevertheless, I'm a bit worried about the amount of API changes - given how prevalent form validation is in your average project, this could easily double the upgrade effort on some projects. We can use silverstripe-upgrader for some of this, but there's subtleties like the changed return value in validate() which will need a dev to look at things.

Can we get more of a sense of what the common use-cases are and how they need to change? I've made a specific comment about the `validate()` method above – it may be better to deprecate this method and introduce a new one rather than change its return type.

In fact, even removing the method and creating a new one would be better since then you'd get a fatal error.

Code of the form `if(!$this->validate())` would silently break, which could be dangerous.

docs/en/04_Changelogs/4.0.0.md
tractorcow: Inline methods increases code complexity (of the combined block) and reduces readability. These methods are in fact not temporary; All that will change is that they will be moved to FormResponseHandler, not removed.

tractorcow:
Updated based on feedback:

 - Form::validate() removed, now use Form::validationResponse() to generate a new `ValidationResponse` object.
 - Removed extraneous methods on ValidationResult

PR #6359, opened 2016-12-02
HTMLEditorField default alignment setting

jonom:
A config setting to control the default alignment of embedded media and images, because people probably often want it to be center instead of leftAlone.

### DIFF
--- a/forms/HtmlEditorField.php
+++ b/forms/HtmlEditorField.php
@@ -22,6 +22,12 @@ class HtmlEditorField extends TextareaField {
 
 	/**
 	 * @config
+	 * @var string Default alignment for Images and Media. Options: leftAlone|center|left|right
+	 */
+	private static $media_alignment = 'leftAlone';
+
+	/**
+	 * @config
 	 * @var bool Should we check the valid_elements (& extended_valid_elements) rules from HtmlEditorConfig server side?
 	 */
 	private static $sanitise_server_side = false;
@@ -674,7 +680,9 @@ class HtmlEditorField_Toolbar extends RequestHandler {
 					'center' => _t('HtmlEditorField.CSSCLASSCENTER', 'Centered, on its own.'),
 					'left' => _t('HtmlEditorField.CSSCLASSLEFT', 'On the left, with text wrapping around.'),
 					'right' => _t('HtmlEditorField.CSSCLASSRIGHT', 'On the right, with text wrapping around.')
-				)
+				),
+				HtmlEditorField::config()->get('media_alignment')
+
 			)->addExtraClass('last')
 		);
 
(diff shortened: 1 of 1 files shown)


forms/HtmlEditorField.php
```diff
@@ -674,7 +680,9 @@ protected function getFieldsForOembed($url, $file) {
 					'center' => _t('HtmlEditorField.CSSCLASSCENTER', 'Centered, on its own.'),
 					'left' => _t('HtmlEditorField.CSSCLASSLEFT', 'On the left, with text wrapping around.'),
 					'right' => _t('HtmlEditorField.CSSCLASSRIGHT', 'On the right, with text wrapping around.')
-				)
+				),
+				Config::inst()->get('HtmlEditorField', 'media_alignment')
```
tractorcow: I would put this config on the HtmlEditorField_Toolbar, since it is the sole consumer of this option.

Then you can use `static::config()->get('media_alignment')` and we won't have to manually rewrite the class name on merge to 4. :)

forms/HtmlEditorField.php
```diff
@@ -22,6 +22,12 @@ class HtmlEditorField extends TextareaField {
 
 	/**
 	 * @config
+	 * @var String Default alignment for Images and Media. Options: leftAlone|center|left|right
```
tractorcow: Lowercase `string` please. :)

tractorcow:
Ignore the cms behat failures; These are known and not the fault of this pr. :)

forms/HtmlEditorField.php
jonom: I considered that but these settings can be a bit hard to find and I feel like people would be more inclined to look for this on HTMLEditorField rather than HtmlEditorField_Toolbar? `insert_width` is a very closely related setting so it would be nice for them to be grouped together on the base class I think.

BTW is there any way to view all the available config settings in SilverStripe? I feel like they should be exposed automatically in the API docs but the only way I know to find these kind of things is to go digging in the source code.

> Then you can use static::config()->get('media_alignment') and we won't have to manually rewrite the class name on merge to 4. :)

Ah, I wasn't aware of this shorthand. Thanks!

jonom:
@tractorcow thanks for your feedback, I replied above but you'll have to hit 'Show outdated' to see

forms/HtmlEditorField.php
tractorcow: > BTW is there any way to view all the available config settings in SilverStripe? I feel like they should be exposed automatically in the API docs but the only way I know to find these kind of things is to go digging in the source code.


Not yet but I'm hoping to get something built for api.silverstripe.org that will let us extract them.

dhensby:
@tractorcow I'm trying to merge up from 3 -> master  and I don't see anywhere that this merges into...

jonom:
@dhensby looks like this needs to go in `HTMLEditorField_File` in master. Is it easier if I open a separate PR for that?

tractorcow:
Yep that's it. :)

PR #6368, opened 2016-12-06
Text supports up to 16 megabytes, not 2

anselmdk:
See the `requireField` method that defines it as `mediumtext`.

### DIFF
--- a/src/ORM/FieldType/DBText.php
+++ b/src/ORM/FieldType/DBText.php
@@ -12,7 +12,7 @@ use SilverStripe\ORM\DB;
 use InvalidArgumentException;
 
 /**
- * Represents a variable-length string of up to 2 megabytes, designed to store raw text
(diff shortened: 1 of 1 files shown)


tractorcow:
I'd rather remove the size constraint at all; It's not correct to declare db-level storage limitations at the orm level.

assertchris:
Agree with @tractorcow - removing these constraints (yet keeping the ability to constrain) seems like a good idea to me.

anselmdk:
I agree that making them configurable would be the best option. Maybe just through `config.yml`.

sminnee:
This discussion is completely unrelated to the PR which is correcting some docs.

Removing these constraints requires a patch to MySQL.

I've confirmed that the limit is 16mb on MSSQL. I can't get clarity on PostgreSQL's limits; as best I can see it's unrestricted. SQLite has a 1gb limit.

I'm merging this and closing the PR.

PR #6369, opened 2016-12-06
Move travis env vars out of .travis.yml and into travis CI dashboard

dhensby:
These variables are now stored in our travis dashboard.

These variables add noise to the OS world and can cause problems if improperly merged up.


tractorcow:
Current failures are known

PR #6370, opened 2016-12-07
ENHANCEMENT Move temporary JSON block into standard component

tractorcow:


admin/code/CampaignAdminList.php
```diff
@@ -0,0 +1,62 @@
...
+    {
+        $data = parent::getSchemaDataDefaults();
+
+        // Set admin section options
+        $adminURL = AdminRootController::admin_url();
+        $data['data']['recordType'] = CampaignAdmin::config()->get('tree_class');
+
+        // Merge custom endpoints
+        $data['data']['collectionReadEndpoint'] = [
+            "url" => "{$adminURL}campaigns/sets",
```
chillu: `campaigns` shouldn't be hardcoded here, the URL route can be changed through config

src/ORM/Versioning/ChangeSet.php
```diff
@@ -535,6 +541,20 @@ public function getDescription()
         return $string;
     }
 
+    /**
+     * Get number of changes
+     *
+     * @return int
+     */
+    public function getChangesCount()
```
chillu: Why isn't `Changes()` always returning an empty list, like other ORM relations? In that case the method wouldn't be needed

tractorcow: The method is needed anyway because, I couldn't use `Changes.Count` due to the fact that React gridfield component doesn't yet support composite summary fields.

tractorcow:
I've added tests to resolve the complaining lack of coverage in diff. :)

Issue #5277, opened 2016-04-05
FIeldnames cannot contain [ if value contains & character (e.g. <p>apple &amp; orange</p>)


hailwood:
TLDR: field values are truncated at an ampersand if the fieldname contains [

**Expected Output**

``` html
"<p>apple &amp; banana</p>"
```

**Actual Output**

``` html
"<p>apple "
```

**How to reproduce**

add the following to a form on a dataobject.

``` php
HtmlEditorField::create('Geocontent[AU]', 'Content for Australia');
```

In the form set the value to "apple & banana" - html view should show

``` html
<p>apple &amp; banana</p>
```

In onbeforeWrite attempt to read the value

``` php
$val = $this->{"Geocontent[AU]"};
die($val);
```

**What's happening?**
This is the offending line https://github.com/silverstripe/silverstripe-framework/blob/3ee8f505b75af39817d394bcf48623e092f43ed2/forms/Form.php#L1431

From what I can tell it's splitting the string on the `&` to get the parts, but due to `Title=Awesome Page&GeoContent[AU]=<p>Apple &amp; Orange</p>` it's truncating the data.

tractorcow:
Question : Does it matter if the [] has a value? e.g.

`Geocontent[]` vs `Geocontent[AU]`? Perhaps the code works for the former, but not the latter?

hailwood:
Hey @tractorcow I don't believe so as the part that causes the issue occurs based purely on the existence of [ in the name.

My thoughts are the intention was for this feature to be used for checkboxes etc where the value would be 1/0 

I've gotted around the issue by changing my fieldname to just GeoContent_AU but I do much prefer the other syntax mostly for the post value of `$gc = $request->postVar('GeoContent'); echo $gc['AU'];` so would be good to get it fixed, but if not then at a minimum I believe any field that can contain an non-numeric/boolean value should throw a warning/error if the fieldname contains [.

quamsta:
This causes major issues with ManyManyExtraFields added to the getCMSFields FieldList: eg:

`$fields->addFieldToTab('Root.Main', HTMLEditorField::create(
				'ManyMany[ArticleNote]'));`

tractorcow:
I see, well we are probably going to deprecate many_many_extraFields, now that we have many_many through relations.

dhensby:
[...]

PR #6375, opened 2016-12-08
FIX Handle fields with square brackets

dhensby:
Potential fix to #5277 

A more robust way to handle array like field names

### DIFF
--- a/forms/Form.php
+++ b/forms/Form.php
@@ -1418,6 +1418,7 @@ class Form extends RequestHandler {
 		if(is_object($data)) $this->record = $data;
 
 		// dont include fields without data
+		/** @var FormField[] $dataFields */
 		$dataFields = $this->Fields()->dataFields();
 		if($dataFields) foreach($dataFields as $field) {
 			$name = $field->getName();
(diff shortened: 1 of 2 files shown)


tractorcow:
A little hard to follow... is there a test you can add that demonstrates where this fixes the issue?

dhensby:
@tractorcow see https://github.com/silverstripe/silverstripe-framework/issues/5277

Do we even think we should support this kind of "broken" syntax for form fields?

tractorcow:
Ok, looking over the code, I think that your solution is better than the existing one. The existing regexp code is kind of flakey, and has caused other problems in the past.

I suggest cleaning up with some unit tests and we can merge it in.

dhensby:
I think this test provides suitable coverage: https://github.com/silverstripe/silverstripe-framework/blob/3.4.2/tests/forms/FormTest.php#L27-L65

As for "Should we support it" - I mean supporting this bad use of array syntax in requests and then reverse engineering it from the form names. Wouldn't dot notation be better?

dhensby:
Test failures look related :P

dhensby:
@tractorcow ready

tractorcow:
Restarting behat tests...

tractorcow:
Ok, I guess these behat issues are pre-existing.

PR #6378, opened 2016-12-08
Clarify UK-english in coding conventions

chillu:
Discussed with @tractorcow - we're mostly using NZ spelling already, e.g. `ViewableData->customise()` and `CLI::supports_colour()`. It's not entirely consistent (e.g. `ShortcodeParser::$error_behavior` is US spelling)

https://docs.silverstripe.org/en/4/contributing/documentation already contains this:

> Use UK English and not US English. SilverStripe is proudly a New Zealand open source project we use the UK spelling and forms of English. The most common of these differences are -ize vs -ise, or -or vs our (eg color vs colour).

### DIFF
--- a/docs/en/05_Contributing/12_JavaScript_Coding_Conventions.md
+++ b/docs/en/05_Contributing/12_JavaScript_Coding_Conventions.md
@@ -21,6 +21,11 @@ as well as the [AirBnB React Conventions](https://github.com/airbnb/javascript/t
 A lot of their rules can be validated via [ESLint](http://eslint.org/),
 and can be checked locally via `npm run lint`.
 
+## Spelling
+
+All symbols and documentation should use UK-English spelling (e.g. "behaviour" instead of "behavior"),
+except when necessitated by third party conventions (e.g using "color" for CSS styles).
+
 ## File and Folder Naming
 
 - All frontend files (CSS, JavaScript, images) should be placed in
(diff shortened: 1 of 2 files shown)


tractorcow:
I would add a caveat "except when necessitated by third party or platform conventions. E.g using `color` for css styles."

chillu:
Done. I've also changed to "UK-English" since that was already contained in the documentation guidelines elsewhere. Shouldn't make much difference in practice, and UK spelling will cause less uncertainty.

sminnee:
This gets my vote but would be good to get some yes/no votes from others.

The main alternative would be US English. This is probably more standard in other tech projects but given our stronger uptake in Europe / Oceania I think UK English is a good standard.

chillu:
I'm basically documenting the status quo. More occurrences:

 * `i18n::pluralise()`
 * `SSTemplateParser::finalise()`
 * `HTMLEditorSanitiser`

sminnee:
@silverstripe/core-team any objections to blocking in UK english as our convention?

clarkepaul:
I have a biased opinion... UK English makes sense especially if it already has the majority.

stevie-mayhew:
I think that using UK English is the best choice.

sminnee:
Is UK English more natural for Europeans than US English?

dhensby:
> Is UK English more natural for Europeans than US English?

Probably not? But what's the point in documenting something that's wrong and where's the benefit in changing all the spelling now? Efforts could be better spent else-where.

I'm :+1: for keeping the status quo

Issue #6376, opened 2016-12-08
Change code examples in docs to match PSR-2


chillu:
* Bracket positioning
 * Form method names should be lowerCamelCase
 * Move `Page_Controller` into separate file
 * etc

chillu:
Closing as duplicate of https://github.com/silverstripe/silverstripe-framework/issues/5757

PR #6379, opened 2016-12-08
Updated coding conventions for PSR-2

chillu:
See https://github.com/silverstripe/silverstripe-framework/pull/6340
See https://github.com/silverstripe/silverstripe-framework/issues/6376

I've removed the conventions for underscores in class names (e.g. `Page_Controller`), which might be contentious - how are we expecting auto-mapping between page types and their controller in a PSR-2 world? `Page` => `PageController`?


tractorcow:
Merged in changes, waiting for tests to finish.

dhensby:
merging in changes ain't the best as they won't be PSR-2?

@ingo regarding controller name changes, that's fine IMO we can provide BC through the [SiteTree@getControllerName method](https://github.com/silverstripe/silverstripe-cms/blob/4.0.0-alpha3/code/Model/SiteTree.php#L2714-L2734) if we like?

You'll need to patch that method, though

chillu:
I've created a ticket for fixing `SiteTree->getControllerName()`: https://github.com/silverstripe/silverstripe-framework/issues/6403

PR #6390, opened 2016-12-13
Outputting the Exception message too

catharsisjelly:
I just encountered a problem with this bit of code that forced me to edit the framework to see why my Import failed. I think by adding the exception message will help developers in the future. My problem was that the custom loader had a bad definition so it throw an error like this

```
Error at framework/model/DataObject.php line 1570: Uncaught Exception: DataObject->getComponent(): Could not find component 'Tags'.
```

There also seems to be no tests based around this so it looks like it could fail for any number of reasons. Maybe in updating this the tests should then include testing when it should throw an Exception?

I think this could also be done in 3.x too just to give a better output

### DIFF
--- a/src/Dev/CsvBulkLoader.php
+++ b/src/Dev/CsvBulkLoader.php
@@ -88,7 +88,7 @@ class CsvBulkLoader extends BulkLoader
                 @unlink($file);
             }
         } catch (Exception $e) {
-            print "Failed to parse {$last}\n";
+            printf("Failed to parse %s because %s\n", $last, $e->getMessage());
         }
 
         return $result;


dhensby:
@catharsisjelly this is only going to go into v4, if that's what you want this looks acceptable, but you may wish to put it into 3.4 and then it'll get merged up into master later

catharsisjelly:
@dhensby fine, I will close this PR and amend the same thing in 3.4, would there be any point in adding tests into 3.4 too? I think it would be wise to ensure that this exception is tested, and/or use a custom exception instead of a standard one

dhensby:
Hmm, adding a test will be tricky because no exception is thrown, just a `printf` is used.

If you'd like to tackle adding a test, then by all means that will be welcome.

catharsisjelly:
I was thinking of just letting it throw the exception and remove the printf entirely but if this is not what the norm is for silverstripe I'll just close this PR and keep #6391 open

dhensby:
I'd leave it for now :)

PR #6391, opened 2016-12-13
Altering error message to show the exception

catharsisjelly:
See #6390


tractorcow:
It's too risky to be displaying error messages contained in exceptions outside of dev mode, as these could contain potentially sensitive information.

You can get around this with a conditional `Director::isDev() ? print('explicit message') : print('simple message')`

catharsisjelly:
@tractorcow Made a change based on your suggestion to this PR

tractorcow:
Nice job. :D

PR #6392, opened 2016-12-13
ENHANCEMENT Enable UploadField.js within non-react sections

tractorcow:
Required for https://github.com/silverstripe/silverstripe-asset-admin/pull/331


admin/client/src/boot/index.js
```diff
@@ -68,6 +68,10 @@ function appBoot() {
   // Bootstrap routing
   const routes = new BootRoutes(store);
   routes.start(window.location.pathname);
+
+  // @TODO - Remove once we remove entwine
+  // Enable top-level css selectors for react-dependant entwine sections
+  window.jQuery('body').addClass('react-boot');
```
dhensby: please can we use a `.js-` prefix on this?

http://csswizardry.com/2015/03/more-transparent-ui-code-with-namespaces/#javascript-namespaces-js-

flamerohr: I don't think it's necessary for this instance, seems to act more like a modernizr for react booting.

Although a check for `window.jQuery` existing would be nice.

tractorcow:
Feedback implemented

dhensby:
NPM failure pre-existing.

PR #6393, opened 2016-12-14
NEW: Added ImagickBackend::crop() for compatibility with GDBackend

UndefinedOffset:
This pull request adds the crop method to ImagickBackend for compatibility with GDBackend, this was suggested to be added to the core in jonom/silverstripe-focuspoint#37. I've based this pull off of the 3 branch, I do have a [branch here](https://github.com/webbuilders-group/silverstripe-framework/tree/image-backend-crop) that is compatible with master but also adds the crop method stub to the interface Image_Backend. I can open a pull against master as well if it's desired.

PR #6394, opened 2016-12-14
NEW: Image_Backend now includes a crop method stub for compatibility with GDBackend

UndefinedOffset:
Per #6393 This adds the crop method stub to Image_Backend however it also does what the #6393's pull did and adds the crop method to ImagickBackend. I can adjust to remove the method from this pull if it makes the merge easier ;). But it will break ImagickBackend on master until 3 is merged.

### DIFF
--- a/src/Assets/Image_Backend.php
+++ b/src/Assets/Image_Backend.php
@@ -131,4 +131,14 @@ interface Image_Backend
      * @return static
      */
     public function croppedResize($width, $height);
+    
+    /**
+     * Crop's part of image.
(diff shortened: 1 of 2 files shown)


dhensby:
I merged your original PR up and into master so this is now redundant. thanks

UndefinedOffset:
@dhensby only thing this one does different is it amends the Image_Backend interface to include a stub for the crop method to force all backends to have it.

dhensby:
yep - I included/added that in the merge up :)

UndefinedOffset:
ah nice :)

PR #6395, opened 2016-12-14
Proper namespace for IPUtils and path correction

andrewandante:
On a non-optimised composer update (i.e. not composer update -o), if you have `SS_TRUSTED_PROXY_IPS` defined, your site will break on two errors:

a) `FRAMEWORK_PATH` not defined
b) Couldn't find class `IPUtils`

These are both remedied in this commit.


dhensby:
this has been resolved by merging #6401 up and into master

PR #6405, opened 2016-12-15
API Add DataObject::getTableName to retrieve static $table_name

robbieaverill:


tractorcow:
table_name is a config, not a protected static. Please use DataObjectSchema for getting the table name.

robbieaverill:
Thanks @tractorcow - equivalent should be `DataObject::getSchema()->tableName(MyDataObject::class)` ?

tractorcow:
Yep that's it.

tractorcow:
In the future I was thinking of adding something like DataObjectSchema_ForClass, similar to how we have Config_ForClass. That might simplify the syntax a bit.

PR #6406, opened 2016-12-16
FIX Update links to docs.ss for default template

robbieaverill:


robbieaverill:
Actually, perhaps this should go back into `3`

PR #6410, opened 2016-12-16
BUG Suppress HTMLEditorField casting

### LINKED ISSUE #6396: Shortcode parsed in HTMLEditorField
Shortcodes are now being rendered in the HTMLEditorField. It's possibly due to the addition of the casting variable on Value. The correct value is stored in the database, but when the field renders in the CMS, it parses the shortcode and renders the parsed shortcode value.
This effects site links, file links, and custom shortcodes

tractorcow:
Fixes #6396

Note that `.RAW` no longer works, as it now parses shortcodes. The second `.Value` invokes the HTMLText::getValue() method, which bypasses shortcodes.

### DIFF
--- a/forms/HtmlEditorField.php
+++ b/forms/HtmlEditorField.php
@@ -26,14 +26,6 @@ class HtmlEditorField extends TextareaField {
 	 */
 	private static $sanitise_server_side = false;
 
-	/**
-	 * @config
-	 * @var array
-	 */
-	private static $casting = array(
-		'Value' => 'HTMLText',
-	);
-
 	protected $rows = 30;
 
 	/**
--- a/forms/TextareaField.php
+++ b/forms/TextareaField.php
@@ -20,7 +20,7 @@
 class TextareaField extends FormField {
 
 	private static $casting = array(
-		'Value' => 'HTMLText',
+		'Value' => 'HTMLText(array("shortcodes" => 0))',
 	);
 
 	/**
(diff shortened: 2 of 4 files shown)


dhensby:
Hmm - this is something we wanted to avoid (having to change templates).

Would it be possible to turn off shortcode parsing when the content is loaded into the editor/textarea?

tractorcow:
hmmmmmmmm

I could maybe, cast HTMLEditorField::Value to a HTMLText that has shortcodes disabled. That's a lot harder though. I'll give it a go.

tractorcow:
@dhensby how's this?

tractorcow:
v3 of this fix; Addresses casting issues in both HTMLEditorField and TextareaField as noted by @nickspiel  https://github.com/silverstripe/silverstripe-framework/issues/6396#issuecomment-267854588

sminnee:
This is a regression between 3.3 and 3.4? If so I think we should get a patch release out sooner rather than later, once it's merged

tractorcow:
@dhensby is doing a new release of 3.4.3 and 3.5.1

dhensby:
yep, this is in the RCs

gregsmirnov:
Something wrong with merge:

git diff origin/3.5..origin/3.4 forms/TextareaField.php

```
-               'Value' => 'HTMLText(array(\'shortcodes\' => false))',
+               'Value' => 'HTMLText(array("shortcodes" => 0))',

```

tractorcow:
Yeah, they are functionally the same... we implemented the same fix at different times.

PR #6413, opened 2016-12-18
search term highlighting without altering case (fixes #1400)

### LINKED ISSUE #1400: [2011-04-19] Text.ContextSummary alters case
created by: dave@room9.co.nz
created at: 2011-04-19
original ticket: http://open.silverstripe.org/ticket/6605

---

Text.ContextSummary changes case of the original text. I.e. A search for "John Key" will highlight as "john key". This is a problem with proper names and brand names.

The search term should highlighted without changing the inner text.

volkerloew:
using preg_replace instead of str_ireplace

### DIFF
--- a/model/fieldtypes/Text.php
+++ b/model/fieldtypes/Text.php
@@ -265,8 +265,7 @@ class Text extends StringField {
 
 				foreach($stringPieces as $stringPiece) {
 					if(strlen($stringPiece) > 2) {
-						$summary = str_ireplace($stringPiece, "<span class=\"highlight\">$stringPiece</span>",
-							$summary);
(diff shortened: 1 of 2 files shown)


dhensby:
cc @tractorcow

tractorcow:
Could you please add a unit test which shows how the case remains unaltered?

tractorcow:
E.g.

Text: `Here is some Testing text`
Search: `testing`
Expected: `Here is some <span class="highlight">Testing</span> text`

volkerloew:
Sorry, missed that initially. I updated an existing test to reflect the change.

Issue #6412, opened 2016-12-16
Pagination Summary is buggy


carpentiers:
Hi. I'd like to point out there's a serious bug with the pagination. I followed what's said in Silverstripe's website tutorial and I remarked that if I change the PaginationSummary's value, I sometimes get a number of pages different from what I assigned and sometimes bizarre behaviours, too.
I will give you two examples. To put you in context, in my tests, I have a total of 6 pages. The only thing that will change is the summary of those pages.

1st example with PaginationSummary(3):
When on the 1st page, I get pages 1, 2, 3, ..., 6.
On the 2nd page, I get pages 1, 2, 3, ..., 6.
On the 3rd page, pages 1, 2, 3, 4, ..., 6. (You can notice a difference in the number of pages, it's not a summary of 3 pages, but 4.)
On the 4th page, pages 1, ..., 3, 4, 5, 6. (Same behaviour.)
On the 5th page, pages 1, ..., 4, 5, 6. (Back to normal.)
On the 6th page, pages 1, ..., 4, 5, 6.
In summary, we see an inconsistent number of pages shown which can affect the layout of the page, but still isn't that bad as the pagination still makes sense which isn't the case in the next example.

2nd example with PaginationSummary(4):
On the 1st page, I get pages 1, 2, 3, 4, 5, 6. (Quite long for what I asked for.)
On the 2nd page, pages 1, 2, 3, 4, 6. (Surprise! No ellipsis, but page 5 is missing, too!)
On the 3rd page, pages 1, 2, 3, 4, 5, 6. 
On the 4th page, pages 1, 2, 3, 4, 5, 6. 
On the 5th page, pages 1, 3, 4, 5, 6. (No ellipsis and page 2 is missing.)
On the 6th page, pages 1, 2, 3, 4, 5, 6.

The loop and the PaginationSummary function are obviously failing somewhere. It seems like PaginationSummary reaches it's end at some point, but doesn't return the value false, so the condition can't get into the ELSE which returns the ellipsis.What do we get then? No ellipsis and a missing page. That seems to occur only when the ellipsis is supposed to be there to replace only 1 single missing page as if the function confuses that page for the first or the last page. The number of pages presented by the pagination summary is inconsistent, too.

dhensby:
The first example seems fine to me. Always show the first and last page and then 3 pages around the current page (one either side of and including the current page).

The second one is clearly doing something wrong for 2nd and 4th instances but all others are as I'd expect.

[...]

PR #6414, opened 2016-12-18
Fixed a pagination bug

PingMetal:
#6412 Depending on the current page, a page not supposed to show up in the pagination summary can be skipped without returning a null value. It then makes it difficult to know it is existent, but just skipped.


dhensby:
thanks

Issue #6383, opened 2016-12-09
BUG: cache constantly being cleared on Redis / Memcached / LibMemcached / WinCache / APC / XCache


sunnysideup:
**From what I understand this any site running APC, Libmemcached/Memcached, Redis or WinCache as a cache for Silvertripe: _the entire cache gets cleared every time ANY dataobject is written_ **

This is obviously not workable on any site that regularly writes to the database (userforms, e-commerce, logs, any user logging in and doing stuff)

Here is the evidence:

----

1. we write a data object:

https://github.com/silverstripe/silverstripe-framework/blob/3.5/model/DataObject.php#L1366

2. as the last step of the write function, the cache is flushed

https://github.com/silverstripe/silverstripe-framework/blob/3.5/model/DataObject.php#L1407

3. in the first line, Aggregate::flushCache is called

https://github.com/silverstripe/silverstripe-framework/blob/3.5/model/DataObject.php#L3257

4. in this function, if the back-end does not support "tags" then the whole cache gets cleaned!

https://github.com/silverstripe/silverstripe-framework/blob/3.5/model/Aggregate.php#L51-L65

We are running a site that was clearing the cache every half second or so. 

----

caches affected are:

* Libmemcached

* Memcached

* Redis  .... I think Redis actually supports tags, but the library I was looking at does not: https://github.com/andrelohmann/silverstripe-cache-backends/blob/master/thirdparty/Zend/Cache/Backend/Redis.php#L381)

* WinCache

* APC

 * Xcache
caches unaffected are:

* files

* sqllite

* mongodb

tractorcow:
Aggegate is removed in 4.0, so just to clarify this is a 3.x only issue.

sunnysideup:
> Aggegate is removed in 4.0, so just to clarify this is a 3.x only issue.

Yes, but maybe the flushcache method has been moved somewhere else? I think it would make more sense for this to be looked at by core team.

sunnysideup:
Basically, since 22 Nov 2013, the above caches have been pretty much useless (I can provide more detail if needed).  I have fixed this by adding the below in my composer.json script:

```yml
        "scripts": {
            "post-install-cmd": [
                "sed -i '/CLEANING_MODE_ALL/d' ./framework/model/Aggregate.php"
            ],
            "post-update-cmd": [
                "sed -i '/CLEANING_MODE_ALL/d' ./framework/model/Aggregate.php"
            ]
        },
```

[...]

PR #6415, opened 2016-12-18
FIX: avoid overzealous flushing of cache

sunnysideup:
This fix sets the cache for Aggregate to File.  This avoids the cache being flushed on any DataObject::write when using a non-file based caching system (see: https://github.com/silverstripe/silverstripe-framework/issues/6383 for details).

### DIFF
--- a/model/Aggregate.php
+++ b/model/Aggregate.php
@@ -64,6 +64,7 @@ class Aggregate extends ViewableData {
 		}
 	}
 
+	private static $_cache_has_been_set = false;
 	/**
 	 * Constructor
 	 *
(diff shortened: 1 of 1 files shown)


sunnysideup:
see: https://github.com/silverstripe/silverstripe-framework/issues/6383...  I am not sure if this fix works (untested!) but it would be good to get some feedback first. This is only for 3.5+ (and maybe 3.4??) and not for 4.0  as this issue no longer exists in 4.0.

tractorcow:
A risk here is that Aggegate::cache() is invoked BEFORE any instance is constructed, as it is as static method.

Perhaps moving the backend pickup into the cache() method would resolve this?

tractorcow:
I hope I didn't put you off your suggestion... I think it's a good one @sunnysideup . :)

sunnysideup:
@tractorcow  ... NO you did not put me off - you raised a very valid point!  I think my proposed solution was silly, because the ?flush would still kill the entire cache.  Here is my replaced one: https://github.com/silverstripe/silverstripe-framework/pull/6416.  Travis seems to fail, but when I look at it, it just appears to be behat - I am not sure if this is related to my suggested change: https://travis-ci.org/silverstripe/silverstripe-framework/jobs/185015501

PR #6416, opened 2016-12-18
FIX: avoid Aggregate clearing entire cache on DataObject::write

sunnysideup:
See https://github.com/silverstripe/silverstripe-framework/issues/6383 for more details.

### DIFF
--- a/_config.php
+++ b/_config.php
@@ -43,5 +43,14 @@ SS_Cache::set_cache_lifetime('GDBackend_Manipulations', null, 100);
 // If you don't want to see deprecation errors for the new APIs, change this to 3.2.0-dev.
 Deprecation::notification_version('3.2.0');
 
+$cachedir = TEMP_FOLDER . DIRECTORY_SEPARATOR . 'aggregatecache';
+
+if (!is_dir($cachedir)) {
+	mkdir($cachedir);
+}
+
+SS_Cache::add_backend('aggregatecache', 'File', array('cache_dir' => $cachedir));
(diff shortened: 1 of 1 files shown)


tractorcow:
This is probably ok for 3.2, but I would mark it as deprecated at 4.0, as with aggregate API.

Test failures are pre-existing. @dhensby do you want to double check this is ok style-wise to merge?

sunnysideup:
@tractorcow are you saying we need to add a deprecate statement in the _config.php file?  How and when would that be used / useful.  Not a rhetorical question - just trying to understand it.  Thank you, as always ;-)

tractorcow:
This is purely my own opinion, but when we add code into 3.x that we don't want bubbling up into 4.0 it's ok to add with an immediate deprecation.

It might be safer to put this code into Aggregate class somewhere, and simply call it from _config.php, rather than inlining it. This would mean it's automatically deprecated.

dhensby:
> It might be safer to put this code into Aggregate class somewhere, and simply call it from _config.php, rather than inlining it. This would mean it's automatically removed on merge up to master.

This is a good idea, can we do it?

tractorcow:
Yeah @sunnysideup please make the suggested tweak, if that's ok :) Aggregate class was deleted, so when we merge we won't end up with artifacts in _config.php that don't do anything.

sunnysideup:
Ok - sorry guys.  I have been super busy with ssmods.com so I have not been able to do any other open source work, but I can perhaps have a go at this in the next few days.

dhensby:
moved to #6569

PR #6417, opened 2016-12-18
Apply PSR2 / Namespace to remaining admin / tests

tractorcow:
This also resolves a bunch of namespaces to framework/tests which were overlooked during the last test namespacing.

Applies the same PSR-2 tests to admin/code, admin/tests and tests/ that were previously only being enforced on /src folder.


sminnee:
I think that we should exclude `tests/php/Core/Manifest/fixtures/classmanifest/` from the PSR-2 syntax checking and leave these classes as-is. We don't want to only test parsing of PSR-2 compliant code.

tractorcow:
I've added `*/fixtures/*` to the global exclusion list. Is this what you were referring to?

sminnee:
@tractorcow files int that classmanifest folder have been altered, presumably by phpcbf. I think those changes should be reverted.

Also - it looks like something is up with the node build. Perhaps TRAVIS_NODE_VERSION is no longer set?

tractorcow:
Something happened within the last week due to merging up from 3 -> master. @dhensby has the deetails.

I'll revert the necessary changes in fixtures. :)

tractorcow:
I've reverted psr2 reformatting to fixtures folder, and added a separate commit to fix the missing TRAVIS_NODE_VERSION env.

tractorcow:
Yay! Thanks @sminnee

PR #6424, opened 2016-12-19
Optimised keyedColumn() and map() for MySQLQuery

mikenz:
I'm not sure if you want DB specific implementations like this but thought I'd create a PR in case there was interest.

### DIFF
--- a/src/ORM/Connect/MySQLQuery.php
+++ b/src/ORM/Connect/MySQLQuery.php
@@ -95,5 +95,5 @@ class MySQLQuery extends Query
             }
         }
         return $column;
-    }   
+    }
 }
--- a/src/ORM/Connect/MySQLQuery.php
+++ b/src/ORM/Connect/MySQLQuery.php
@@ -60,4 +60,40 @@ class MySQLQuery extends Query
             return false;
         }
     }
+
+    /**
+     * Return an array containing all values in the leftmost column, where the keys are the
+     * same as the values.
+     *
+     * @return array
+     */
+    public function keyedColumn()
+    {
+        $column = [];
+        if (is_object($this->handle)) {
+            foreach ($this->handle->fetch_all(MYSQLI_ASSOC) as $record) {
+                $val = $record[key($record)];
+                $column[$val] = $val;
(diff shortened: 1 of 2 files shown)


dhensby:
I don't think we mind MySQL-centric syntax in the MySQL classes.

What kind of performance increase does this give and why is it more performant than what we have?

mikenz:
before

after

dhensby:
Well look at that. @kinglozzer @tractorcow

Please can we get some test coverage (skip the tests if not a mysql driver)? Then I'm happy :)

mikenz:
I guess there's no existing test that uses keyedColumn so the current code isn't tested either. I'll write a test this week and added it to this PR.

kinglozzer:
Makes sense to me, but I think we need our database overlord @tractorcow to sign this off 😄

dhensby:
testfailure on postgres - the keys (IDs) are different. We need to make sure the test doesn't make assumptions about what the IDs are and instead pull them out of the DB...

mikenz:
Yeah, I'll change the test to use the Sort column instead of the ID column - an equally valid test and the Sort values are defined in the fixture.

dhensby:
Oh, whoops, sorry I just pushed a change up (I thought you'd be sleeping)

dhensby:
Thanks for the test coverage, super helpful. Once this is green, I'll merge. It's going into master and so @tractorcow can revert if he doesn't like it.

Can we make any of these kinds of improvements to other parts of core?

mikenz:
I'm working on performance of our installation all this week so where I can (and think it'll get merged) I'll raise PRs. Some of my improvements only work because I can make assumptions that won't hold true for everyone.

dhensby:
OK - great, thanks - really appreciate that :)

dhensby:
@mikenz also, if you find any areas that are bottlenecks but can't submit a PR for them, please raise them as issues.

tractorcow:
Sorry, this needs to be reverted since it's not compatible with mysql-nd, and is breaking sites. =(

PR #6425, opened 2016-12-19
remove Created field on duplicate

andrewandante:
Currently, using the `duplicate()` method on a Page will also duplicate the 'Created' date. This is counter-intuitive and should be fixed to represent the creation date of the new record. This PR will unset the 'Created' field from the map, allowing it to be set on `write()`.

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -531,7 +531,9 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	 */
 	public function duplicate($doWrite = true) {
 		$className = $this->class;
-		$clone = new $className( $this->toMap(), false, $this->model );
+		$map = $this->toMap();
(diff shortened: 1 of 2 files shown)


dhensby:
I'm happy with this as I feel the created date being copied over on duplicate is a bug not a feature. The `ID`, `Created` and `LastEdited` values are system values not ones that the user is duplicating as part of content.

As I see it `Created` is the time when this particular row was added to the DB. If we duplicate `Created` we should also be duplicating history too (which we shouldn't)

tractorcow:
Yes this is fine with me also. :)

tractorcow:
We should also clear the `Version` field, as the new duplicate will be versioned at 1 again with a new ID.

tractorcow:
Sorry, Versioned extension already does this in onBeforeDuplicate. :)

PR #6432, opened 2016-12-20
FIX UploadField FileExists check

jonom:
UploadField in asset admin area would only check root folder for existing files (Fixes https://github.com/silverstripe/silverstripe-cms/issues/950)

ping @dhensby


dhensby:
:(

jonom:
Point friendly bugfix on the way :)

PR #6433, opened 2016-12-20
FIX UploadField FileExists check

jonom:
UploadField in asset admin area would only check root folder for existing files (Fixes https://github.com/silverstripe/silverstripe-cms/issues/950)

ping @dhensby 

Copied the filtered name check functionality to the Action so we can do this as a point fix.

It all seems a bit funky... I'd prefer to use an ID for folder but since this is a bugfix I'm trying not to stray to far from the existing structure.

Is it okay to target 3.5.1 or should this go somewhere else, like 3.4?

### DIFF
--- a/forms/UploadField.php
+++ b/forms/UploadField.php
@@ -921,6 +921,7 @@ class UploadField extends FileField {
 			'urlSelectDialog' => $this->Link('select'),
 			'urlAttach' => $this->Link('attach'),
 			'urlFileExists' => $this->link('fileexists'),
+			'folderName' => $this->getFolderName(),
 			'acceptFileTypes' => '.+$',
 			// Fileupload treats maxNumberOfFiles as the max number of _additional_ items allowed
 			'maxNumberOfFiles' => $allowedMaxFileNumber ? ($allowedMaxFileNumber - count($this->getItemIDs())) : null,
@@ -1228,6 +1229,7 @@ class UploadField extends FileField {
 	 *
 	 * @param string $originalFile Filename
 	 * @return bool
+	 * @deprecated 4.0
 	 */
 	protected function checkFileExists($originalFile) {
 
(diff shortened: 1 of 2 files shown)


dhensby:
If it's a bug in the 3.4 line, please target 3.4

tractorcow:
Can you tell why the `getFolderName` in `checkFileExists` doesn't work, but the one in `Field` does exist?

jonom:
For getFolderName() to work as expected, setFolderName() has to have been called first. The problem is that in the asset admin upload field, a URL param is relied on to indicate the upload folder and configure the form field. The fileexists ajax request only passes the name of the file, not the full target path, so there is not enough information being supplied and the folder that gets checked ends up being root.

tractorcow:
Oh I see, so the issue is specific to asset-admin. Maybe we should fix it there also, instead of risking regressions in UploadField?

tractorcow:
CMSFileAddController relies on `$folder = $this->currentPage();`. That should be in the Session in the `Folder.currentPage` key shouldn't it, even when doing fileexists checks?

jonom:
Yeah, I removed the reliance on session data previously to allow for a direct file edit URL. Relying on a session variable is messy. The Ajax request should just include all the necessary information to check if a file exists I think (file name + path)

tractorcow:
> Yeah, I removed the reliance on session data previously to allow for a direct file edit URL. Relying on a session variable is messy. 

Yes but it's the way that 3.x works. :) If you remove it you'll get issues such as this.

When did you remove it?

jonom:
The session thing was originally added as a bug fix only, but I think this is a more robust approach. Sorry can't give more info right now, typing on a phone.

tractorcow:
ok, send me a link when you get the time. :)

jonom:
@tractorcow https://github.com/silverstripe/silverstripe-cms/issues/950 was an issue that was fixed by introducing session data in https://github.com/silverstripe/silverstripe-cms/pull/951/commits/dc865399dcbc0c485eb42fbefc141bad733d1c55. I unwittingly re-introduced the issue with https://github.com/silverstripe/silverstripe-cms/pull/1566/commits/c965133da03941fc736384d17518dc37be3d0e07 when I removed the reading of the session data. I did that because it wasn't an ideal fix, from memory the session var wasn't getting cleared so would gum things up and be referred to out of context. I saw that as an issue because it meant you couldn't reliably link directly to a file for editing in the CMS.

The root problem is that the UploadField fileexists function doesn't get all the information it needs in the request, so it has to work out what the target directory for the upload is. I've tried to address that by ensuring the target folder is included in the configuration data on the front-end UploadField, and including it in the ajax request as an additional param. If that param is missing though, the old behaviour of trying to infer the target folder kicks in, so should be backwards compatible.

tractorcow:
Oh, I just remembered, I wrote a module a while back that added custom folder selection to uploadfield.

https://github.com/silverstripe/silverstripe-selectupload/blob/master/code/SelectUploadField.php

Same kind of solution?

tractorcow:
I guess since asset-admin is going away, we don't need to spend that much time trying to fix it. The session id SHOULD still be kept, regardless of how hackey it is; LeftAndMain relies on it by principle.

Can you please check that the `upload` handler also gets the right folder name, even if it's not in the session? It's the other endpoint that needs the correct folder to be assigned.

jonom:
The session ID is still set, I kept it as you noted, in case other functions rely on it. I just didn't want to keep reading the session ID, specifically for this reason: if you visit a folder in asset admin the folder ID is saved in session, then if you try to directly visit a file in asset admin by URL, the session ID doesn't match the parent folder of the file you're trying to access and it blows up. When you're viewing a file in asset admin we know the file ID so we can deduce the parent ID, there is no need to read this from session.

I'm not actually sure how/where to check the upload handler, but in all the testing I did the uploaded files end up in the correct directory so I assume this is still working? 3.4 has been out in the wild for a while so I think we would have had some complaints by now if uploads were ending up in the wrong place.

The bug I'm trying to fix right now is purely for the fileexists ajax request - which probably is probably employed on very few sites because the default action is not to overwrite files but to rename the new upload with a number on the end, so the fileexists check never happens by default.

Issue #6422, opened 2016-12-19
[GRIDFIELD] No titles or labels for inputs and buttons in gridfield search


clarkepaul:
From what I can tell some of the actions have "title" in code but it doesn't seem to be in the output of markup. The inputs should have labels, but failing that at least a title.

![image](https://cloud.githubusercontent.com/assets/555033/21301402/163d9146-c612-11e6-95d9-ef299b90be89.png)

lukereative:
I think this will be made redundant with #8309

lukereative:
Closing as legacy filter header is being replaced by the new search #8309

PR #6354, opened 2016-11-30
Update buttons to new flat bootstrap style

wilr:
NOTE: MOVED TO OPENSOURCERERS REPO!

Upgrading the CMS admin interface to use the new buttons and remove `ss-ui` references. Also upgrading to use the new font icon style.

Removes the use of jQuery UI Button (yay!)

Have left the old styles in the stylesheet to make sure older buttons will have some form of styling.

Old: screen shot 2016-11-30 at 11 34 19 am

New: screen shot 2016-12-01 at 9 29 33 am

screen shot 2016-12-01 at 9 20 43 am

Marking as WIP for @clarkepaul to review. Requires branches for:

~~CMS: https://github.com/wilr/silverstripe-cms/tree/feature/ui-button-refresh~~
~~SiteConfig: https://github.com/silverstripe/silverstripe-siteconfig/pull/42~~

PR's moved to Opensourcers repo:
https://github.com/open-sausages/silverstripe-siteconfig/tree/features/4.0/ui-button-refresh
https://github.com/open-sausages/silverstripe-cms/tree/features/4.0/ui-button-refresh
https://github.com/open-sausages/silverstripe-framework/tree/features/4.0/ui-button-refresh
*Also the same branch names on:*
frameworktest
reports
assetadmin

## Todo

- [x] Save / Publish button transitions

Issue #6431, opened 2016-12-20
Move GridField "Import CSV" into modal, with better UX


clarkepaul:
A partial fix for this issue is included in this PR https://github.com/silverstripe/silverstripe-framework/pull/6354 but it doesn't resolve all the experience completely.

Having it in the modal has improved the usability somewhat since it was sitting loose under gridfields but still room for improvements.

Current implementation:
![image](https://cloud.githubusercontent.com/assets/555033/21337839/a5c915f6-c6d5-11e6-9122-7536eb05eb4b.png)

Proposed UX and designs improvements:
https://projects.invisionapp.com/boards/7XX68PM4TPCR/

sminnee:
@clarkepaul can you elaborate on what "needs some love" might mean? :-)

clarkepaul:
This UI doesn't really make sense if you go through the steps to import. The "Import from CSV" button looks like it does the same action as "Choose file"—can we do away with the "Import from CSV" button completely? This could also be a candidate for a drop zone. Once you have imported from CSV is there any way to undo (doubtful but thought it was worth clarifying in the off chance)?

The "Done" action is very out of place compared the rest of the CMS UI, so it might need some thought as to what benefit it provides over providing a success message and closing the modal automatically (it is a duplicate action as the X button as well).

It would be good to review the text/labels to see if we can improve them in any way also.

clarkepaul:
This issue is marked as "Type/UX" as it needs some time from a designer or dev willing to provide a nicer solution than the current experience (prior to development).

jamessilverstripe:
![ss import csv modal - initial](https://user-images.githubusercontent.com/29618750/27415739-c43735dc-575c-11e7-81eb-8ec87a6e9388.png)
![ss import csv modal - uploaded](https://user-images.githubusercontent.com/29618750/27415740-c44e18c4-575c-11e7-9321-75bd87ed326b.png)
![ss import csv modal - done](https://user-images.githubusercontent.com/29618750/27415738-c4034574-575c-11e7-9a8e-0eb4beeb1da0.png)

"Add from files" would only show CSV files as selectable

clarkepaul:
UX completed as above, ready for sizing and development.

clarkepaul:
Comment from @lerni 
> CSV-Import is also hard to find. And if you import, the Success Message like "Imported 20 records." gets hidden in the collapsed Search, since "Import from CSV" collapses Search Panel.

[...]

PR #6434, opened 2016-12-20
Refresh CMS UI buttons to new flat style and bootstrap classes

wilr:
See discussion on #6354.

Related pull requests:
 - [ ] https://github.com/silverstripe/silverstripe-cms/pull/1699
 - [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/359
 - [ ] https://github.com/silverstripe/silverstripe-reports/pull/57
 - [ ] https://github.com/silverstripe/silverstripe-siteconfig/pull/46
 - [x] https://github.com/silverstripe/silverstripe-behat-extension/pull/140


clarkepaul:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/6422

admin/client/src/components/GridField/GridFieldAction.js
```diff
@@ -17,6 +18,9 @@ class GridFieldAction extends SilverStripeComponent {
   }
 
   handleClick(event) {
+    console.log('handle click');
+
+    return false;
```
flamerohr: this almost looks like test data?

admin/client/src/legacy/GridField.js
```diff
@@ -133,21 +134,52 @@ $.entwine('ss', function($) {
...
+      var backdrop = $('.modal-backdrop');
+
+      if(backdrop.length < 1) {
+        backdrop = $('<div class="modal-backdrop fade in"></div>');
+        $('body').append(backdrop);
+      } else {
+        backdrop.addClass('fade in').fadeIn();
+      }
+
+      $(this.data('target')).addClass('in');
```
flamerohr: save `$(this.data('target'))` to a variable, so you won't need to call it again below

admin/client/src/legacy/GridField.js
```diff
@@ -133,21 +134,52 @@ $.entwine('ss', function($) {
...
+
+      // workaround until GridField rewritten into react.
+      var self = this;
+
+      $(this.data('target')).find('[data-dismiss]').on('click', function() {
+        backdrop.fadeOut(function() {
+          $(this).removeClass('in');
```
flamerohr: use `backdrop` rather than `$(this)`, makes it easier to predict what's happening at a glance.

admin/client/src/state/records/RecordsReducer.js
```diff
@@ -7,8 +7,12 @@ function recordsReducer(state = initialState, action) {
   let records = null;
   let recordType = null;
   let record = null;
+  let apolloRecordRelation = {};
 
   switch (action.type) {
+    case 'APOLLO_QUERY_RESULT':
```
flamerohr: I think this leaked from the GraphQL changes, don't think this belongs in this PR :D

flamerohr:
The Done button seems out of place at the bottom, when it's the functionality as the Close button?
screen shot 2016-12-21 at 2 32 47 pm

flamerohr:
I get a Forbidden error when importing, is `import` not supposed to be in the `$allowed_actions` list by default?

clarkepaul:
@flamerohr I've opened a issue based on the modal for importing of csv files https://github.com/silverstripe/silverstripe-framework/issues/6431 although it should be working before merge (if it actually doesn't work).

I will update label of button to "close", as a temporary measure.

flamerohr:
@clarkepaul Oh sweet, I'll leave that out for now then
Still reviewing, just working around a few blockers currently

flamerohr:
A bit of jumping around with loading indicators in the Add Page section https://vimeo.com/196661702

flamerohr:
Changes look good, will create a new issue for the video linked

tractorcow:
I'm taking over peer review.

tractorcow:
I've fixed behat tests, and added a patch to some issues with the import form. GridFieldImportButton wasn't working outside of the ModelAdmin (e.g. SecurityAdmin, which is a LeftAndMain subclass).

admin/client/src/legacy/GridField.js
```diff
@@ -137,18 +137,37 @@ $.entwine('ss', function($) {
...
+      }
+		},
+		onunmatch: function() {
+			this._super();
+		},
 
-      var backdrop = $('.modal-backdrop');
+    openmodal: function() {
+      // Remove existing modal
+      let modal = $(this.data('target'));
```
flamerohr: Was this intended for a pseudo-reload of the modal html?
Why couldn't it re-use the existing modal?

tractorcow: If the inner template has been replaced, we need to update the content.

I could maybe replace the inner content. Would that be ok?

flamerohr: I think it's fine, just checking if replacing the modal was intentional, that pjax reloading throws me off sometimes... :/

tractorcow: I've implemented a fix in any case; we now re-use the outer container, but update the inner container on reload.

admin/client/src/legacy/GridField.js
```diff
@@ -153,29 +153,38 @@ $.entwine('ss', function($) {
...
       modal.find('[data-dismiss]').on('click', function() {
-        backdrop.fadeOut(function() {
-          backdrop.removeClass('in');
-        });
-
+        backdrop.removeClass('in');
         modal.removeClass('in');
+        setTimeout(function() {
+          backdrop.remove();
+        }, 0.2)
```
flamerohr: `setTimeout()` uses milliseconds, so 0.2 is equivalent to 0 since it's so short anyway.

PR #6435, opened 2016-12-21
Revert "Optimised keyedColumn() and map() for MySQLQuery"

tractorcow:
Reverts silverstripe/silverstripe-framework#6424

Sorry, this includes non-mysqlnd compatible API. =(

### DIFF
--- a/src/ORM/Connect/MySQLQuery.php
+++ b/src/ORM/Connect/MySQLQuery.php
@@ -60,40 +60,4 @@ class MySQLQuery extends Query
             return false;
         }
     }
-
-    /**
-     * Return an array containing all values in the leftmost column, where the keys are the
-     * same as the values.
-     *
-     * @return array
-     */
-    public function keyedColumn()
-    {
-        $column = [];
-        if (is_object($this->handle)) {
-            foreach ($this->handle->fetch_all(MYSQLI_ASSOC) as $record) {
-                $val = $record[key($record)];
-                $column[$val] = $val;
-            }
-        }
-        return $column;
-    }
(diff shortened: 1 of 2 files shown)


tractorcow:
Sorry, in-github revert button creates a branch in the same repo by default.

dhensby:
can we create a query class for mysqlnd? Perhaps we can drop support for `libmysql` - it is deprecated as of php 5.5?

dhensby:
our minimum support for ss 4 is 5.5+ and I believe we are removing deprecated libs and APIs from SS4, so this should be acceptable?

tractorcow:
Where is libmysql being deprecated? http://php.net/manual/en/mysqlinfo.library.choosing.php implies both are actively supported and have no EOL set.

dhensby:
> This extension is deprecated as of PHP 5.5.0, and has been removed as of PHP 7.0.0.

http://php.net/manual/en/intro.mysql.php

dhensby:
Perhaps I am confused between `mysql` and `libmysql`

tractorcow:
Oh right, this is the `mysql` library being deprecated in favour of `mysqli`, which is provided by both mysqlnd and libmyself. :)

dhensby:
Ah, ok.

So one solution is to make a mysqlnd only drive?

tractorcow:
Yes, you could do that. Alternatively, have one driver which separate code paths for each of the two libraries.

dhensby:
@mikenz do you have any desire to create a DB driver for mysqlnd?

PR #6436, opened 2016-12-21
BUG Temp disable shortcode SPLIT behaviour due to crash

### LINKED ISSUE #5987: Centered image alignment crashes on save

tractorcow:
Fixes #5987


tractorcow:
Thanks @sminnee ! 3.5.1 will tag tomorrow :)

PR #6440, opened 2016-12-23
ENHANCEMENT: Debug class emits plain text for application/json requests

tractorcow:
Also see https://github.com/silverstripe/silverstripe-admin/pull/137

I've majorly cleaned up and refactored the html generation from Debug class into either CliDebugView or DebugView

### DIFF
--- a/src/Control/HTTPRequest.php
+++ b/src/Control/HTTPRequest.php
@@ -428,7 +428,7 @@ class HTTPRequest implements ArrayAccess
     {
         return (
             $this->requestVar('ajax') ||
-            $this->getHeader('X-Requested-With') && $this->getHeader('X-Requested-With') == "XMLHttpRequest"
+            $this->getHeader('X-Requested-With') === "XMLHttpRequest"
         );
     }
 
--- a/src/Dev/CliDebugView.php
+++ b/src/Dev/CliDebugView.php
@@ -3,6 +3,8 @@
 namespace SilverStripe\Dev;
 
 use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Core\ClassInfo;
+use SilverStripe\Core\Convert;
 
 /**
  * A basic HTML wrapper for stylish rendering of a developement info view.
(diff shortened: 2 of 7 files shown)


chillu:
Copying my comment to Damian from Slack: I’m not sure its the right level of abstraction (has nothing to do with Ajax), but then again that’s already a problem in core. I think when you requested application/json, plaintext errors are always better than HTML errors. But this might cause regressions for devs who relied on `isAjax()` for app logic

dhensby:
Rather than saying that it's an "ajax" request - which is a bad way of determining how to respond IMO - we should have a better way of returning different response formats.

Can't the error handler look at the `Accept` header and respond appropriately instead of always producing HTML responses?

src/Control/HTTPRequest.php
```diff
@@ -406,8 +406,11 @@ public function getURL($includeGetVars = false)
      */
     public function isAjax()
     {
+        $contentType = $this->getHeader('Content-Type') ?: $this->getHeader('content-type');
```
dhensby: Shouldn't this be the `accept` header?

Better to use `$this->getAcceptMimetypes()`?

tractorcow:
Request body can also have a content-type. :) If you are posting json_data this assumes ajax.

I wanted to use accept, but in my case the accept is `*/*` which is useless.

tractorcow:
I'll see if I can fix graphql to send the correct accept header.

tractorcow:
I've updated graphql to send the appropriate accept header, and updated the code to respect this now.

@dhensby how does this look now? :D

tractorcow:
The critique has been raised that we probably shouldn't be using is_ajax in places where we currently do for certain purposes.

It may be better to forgo this fix in favour of solution-specific patches to each area.

The use case I had in mind was Debug::create_debug_view() correctly rendering text errors for ajax requests; However it should probably be comparing the `accept:` header in this case for text/html, and inverting the check.

dhensby:
I think you're right - we're doing this backwards.

Better to check the Accept header for content types than infer format from request "type".

tractorcow:
I will update this PR once I get more time... maybe later this week. :) /remindme 4 days

tractorcow:
I've pushed up a new proof of concept... I need to write tests still before we can merge it.

tractorcow:
Note: Still working on this, will hope to update shortly (this week, not today).

tractorcow:
I'll need to update this, maybe beta2.

tractorcow:
ok, looking to pick this up again. :)

tractorcow:
@dhensby I've updated this PR with some cleanup of Debug HTML generation.

tests/php/Dev/DebugViewTest.php
```diff
@@ -0,0 +1,73 @@
+<?php
+
+namespace SilverStripe\Dev\Tests;
+
+use SilverStripe\Dev\DebugView;
+use SilverStripe\Dev\SapphireTest;
+use SilverStripe\Dev\Tests\DebugViewTest\ObjectWithDebug;
+
+class DebugViewTest extends SapphireTest
```
tractorcow: Oh look a test for a class we've had 10 years.

src/Dev/CliDebugView.php
```diff
@@ -129,4 +131,76 @@ public function renderVariable($val, $caller)
...
+     * Get debug text for this object
+     *
+     * @param mixed $val
+     * @return string
+     */
+    public function debugVariableText($val)
+    {
+        // Check debug
+        if (ClassInfo::hasMethod($val, 'debug')) {
+            return $val->debug();
```
dhensby: A lot of the time these methods hard code html into the output - is there a way we could either strip it or pass a parameter to indicate we don't want html?

src/Dev/Debug.php
```diff
@@ -88,118 +76,136 @@ public static function caller()
...
-    public static function endshow($val)
+    public static function endshow($val, $showHeader = true, HTTPRequest $request = null)
     {
-        if (!Director::isLive()) {
-            $caller = Debug::caller();
-            echo "<hr>\n<h3>Debug \n<span style=\"font-size: 65%\">($caller[class]$caller[type]$caller[function]()"
-                . " \nin " . basename($caller['file']) . ":$caller[line])</span>\n</h3>\n";
-            echo Debug::text($val);
-            die();
+        // Don't show on live
```
dhensby: This method could just be `static::show(); die;`?

dhensby:
I like it, a couple of points made that could be ignored if you just want to get this over the line

src/Dev/Debug.php
tractorcow: static::caller() doesn't work since it is another level deep.

src/Dev/CliDebugView.php
tractorcow: Unfortunately debug() doesn't allow context to be provided; We simply (and have to) assume that user-implemented debug() methods will output in the correct format.

User code can use Debug::create_debug_view() internally to achieve this, but we can't easily and probably shouldn't enforce it.

src/Dev/Debug.php
dhensby: ah - ok

Issue #6397, opened 2016-12-14
Insecure random number generators in RandomGenerator should require explicit enabling


madmatt:
Ideally we can add a Config variable, that defaults to false, that prevents the use of the last two random number generators (Capicom for Windows and `mt_rand()`) unless the config var is changed to `true`, so it is opt-in to produce non-psuedo-random numbers.

Probably a v4 change given it would change the output of the method (might throw an Exception where none was thrown before).

Reference: https://github.com/silverstripe/silverstripe-framework/blob/d38097ec146d5509fb397d930f069a313093ac13/src/Security/RandomGenerator.php#L56

dhensby:
Are there any tangible risks to how it is at the moment?

tractorcow:
The method comment states `     * Note: Returned values are not guaranteed to be crypto-safe,`. This method should not be used in any place where it could potentially raise a security risk at all.

I would rather see this left in place, and we implemented a proper crypto-safe random number generator instead. php 5.5 (supported in 4.0) has better crypto-random generation. e.g. http://php.net/manual/en/function.hash-pbkdf2.php

robbieaverill:
Sort of related - `Function mcrypt_create_iv() is deprecated` in `framework/src/Security/RandomGenerator.php:30` during a module's Travis build on PHP 7.1. If we want to support 7.1 for SS 4 we'll probably need to replace it.

robbieaverill:
Regarding dropping the deprecated (in 7.1) mcrypt library use, [here is the RFC that drops it](https://wiki.php.net/rfc/mcrypt-viking-funeral) (for reference). It seems that the suitable replacement would be `openssl_random_pseudo_bytes' which is already implemented in this method, just a lower priority than `mcrypt_create_iv`.

I propose we simply remove the mcrypt use, leave the rest and give a higher priority for Windows to the method that @tractorcow shared over the existing code. Personally, I don't see any reason why we shouldn't remove old Windows code now (4.x) in favour of more modern, cross platform methods.

dhensby:
@robbieaverill from a quick look that commit looks good - but is there a thirdparty library we can use that does the same thing?

dhensby:
https://github.com/ircmaxell/RandomLib this looks good, well maintained and regularly contributed to

robbieaverill:
@dhensby yeah cool. As long as we are ok with replacing the functionality with a new backend essentially then I'll have a play with that instead

[...]

PR #6442, opened 2016-12-29
PHP 7.1 compatibility tweaks

robbieaverill:
* Replace the soft string to int casting in ini2bytes with a more explicit equivalent, update tests
* Consolidate some temporary file creation and deletion logic in UploadTest
* Allow PHP 7.x to use `random_bytes` over mcrypt since it is deprecated in 7.1 (relates to #6397)
* Update Travis configuration to build 7.1 as the 7.x default, while still running 7.0


src/Assets/File.php
```diff
@@ -1040,22 +1038,29 @@ public static function format_size($size)
...
-     * @param string $iniValue
+     * @param  string $iniValue
      * @return int
      */
     public static function ini2bytes($iniValue)
     {
-        switch (strtolower(substr(trim($iniValue), -1))) {
+        $iniValues = array_filter(str_split(trim($iniValue)));
+        $unit = strtolower(array_pop($iniValues));
+        $quantity = (int) implode($iniValues);
```
dhensby: `implode` should have the two params

dhensby:
thanks :)

PR #6444, opened 2016-12-29
TEST Add Group tests for collating family IDs and getting all children

robbieaverill:


tests/php/Security/GroupTest.php
```diff
@@ -136,6 +137,38 @@ public function testCollateAncestorIDs()
         $this->assertContains($orphanGroup->ID, $orphanGroup->collateAncestorIDs());
     }
 
+    /**
+     * Test that Groups including their children (recursively) are collated and returned
+     */
+    public function testCollateFamilyIds()
+    {
+        $group = $this->objFromFixture(Group::class, 'parentgroup');
+        $this->assertSame([2, 3, 4], $group->collateFamilyIDs());
```
dhensby: You mustn't hard code IDs because postgres doesn't seem to reset the IDs for each test.

You could use `$this->allFixtureIDs('Group')` perhaps?

robbieaverill: Updated, thanks

dhensby:
That's an annoying failure :/

robbieaverill:
I loosened the assertion a little, green now

Issue #6441, opened 2016-12-23
Improve 4.0 upgrading guide with Robbie's feedback


chillu:
Robbie's notes at  https://docs.google.com/document/d/1DI3Cf111SWyEVdIUENSRaYGt3pwBR9FwF5ZyYZlF02w/edit#heading=h.vf0es2ofzk5e

Reflect those in https://github.com/silverstripe/silverstripe-framework/blob/master/docs/en/04_Changelogs/4.0.0.md

## Pull requests
- [x] https://github.com/silverstripe/silverstripe-framework/pull/7169

chillu:
Ensure you mention DataObject ownership more clearly (particularly since you have to change any objects relating to `File` in your code now): https://docs.silverstripe.org/en/4/developer_guides/model/versioning/#dataobject-ownership

PR #6446, opened 2016-12-29
API Allow controller discovery without underscores (PSR-2 compliance)

### LINKED ISSUE #6403: Allow controller discovery without underscore (PSR-2 compliance)
The `SiteTree->getControllerName()` (https://github.com/silverstripe/silverstripe-cms/blob/4.0.0-alpha3/code/Model/SiteTree.php#L2714-L2734) assumes an underscore in the class name, e.g. `Page_Controller`, which goes against the PSR-2 conventions (see [discussion](https://github.com/silverstripe/silverstripe-framework/pull/6379#issuecomment-267014124)).

Change the method to allow for non-underscored controller names, and deprecate the use of underscores (but keep it functional).

There's a separate PR for changing docs code examples accordingly: https://github.com/silverstripe/silverstripe-framework/issues/6376

robbieaverill:
* Update docs that reference `*_Controller` to `*Controller`.
* Reformat some docs for PSR-2 compat (relates to #6376). I didn't reformat them all, just a few of them. Still a fair bit of work needed for this issue.
* Update `Security` to pull `Page` and `PageController` from the global namespace instead of importing, in case the CMS is not present.
* Update `SSViewer` to resolve template names for `_Controller` and `Controller`. Use regex match to prevent classes like `SilverStripe\Control\Controller` from being matched by accident.
* Remove some legacy PHP code from the `SSViewerTest` class (e.g. `$self = $this` for closures).

---

* Resolves #6403.

## Blockers

- [ ] silverstripe/silverstripe-cms#1703

## Blocks

- [ ] silverstripe/silverstripe-installer#146

### DIFF
--- a/src/View/SSViewer.php
+++ b/src/View/SSViewer.php
@@ -254,7 +254,7 @@ class SSViewer implements Flushable
             $templates[] = ['type' => 'Includes', $template];
 
             // If the class is "PageController" (PSR-2 compatibility) or "Page_Controller" (legacy), look for Page.ss
-            if (preg_match('/^(?<name>.+[^\\])_?Controller$/i', $class, $matches)) {
+            if (preg_match('/^(?<name>.+[^\\\\])_?Controller$/iU', $class, $matches)) {
                 $templates[] = $matches['name'] . $suffix;
             }
 
--- a/src/View/SSViewer.php
+++ b/src/View/SSViewer.php
@@ -253,9 +253,9 @@ class SSViewer implements Flushable
             $templates[] = $template;
             $templates[] = ['type' => 'Includes', $template];
 
-            // If the class is "Page_Controller", look for Page.ss
-            if (stripos($class, '_controller') !== false) {
-                $templates[] = str_ireplace('_controller', '', $class) . $suffix;
+            // If the class is "PageController" (PSR-2 compatibility) or "Page_Controller" (legacy), look for Page.ss
+            if (preg_match('/^(?<name>.+[^\\])_?Controller$/i', $class, $matches)) {
+                $templates[] = $matches['name'] . $suffix;
             }
 
             if ($baseClass && $class == $baseClass) {
--- a/.upgrade.yml
+++ b/.upgrade.yml
@@ -1087,6 +1087,7 @@ mappings:
   ConfigTest_TestNest: SilverStripe\Core\Tests\Config\ConfigTest\TestNest
   ConfigTest: SilverStripe\Core\Tests\Config\ConfigTest
   ConfigTest_Config_MemCache: SilverStripe\Core\Tests\Config\ConfigTest\ConfigTestMemCache
+  Page_Controller: PageController
 skipConfigs:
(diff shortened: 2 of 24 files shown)


src/View/SSViewer.php
```diff
@@ -253,9 +253,9 @@ public static function get_templates_by_class($classOrObject, $suffix = '', $bas
             $templates[] = $template;
             $templates[] = ['type' => 'Includes', $template];
 
-            // If the class is "Page_Controller", look for Page.ss
-            if (stripos($class, '_controller') !== false) {
-                $templates[] = str_ireplace('_controller', '', $class) . $suffix;
+            // If the class is "PageController" or "Page_Controller" (legacy), look for Page.ss
+            if (preg_match('/\w+_?Controller$/i', $class)) {
```
tractorcow: If you are using a regexp to match, use the same regexp to replace.

```php
if (preg_match('/^(?<name>\w+)(_?Controller)$/i', $class, $matches)) {
	$templates[] = $matches['name'] . $suffix;
}
```

robbieaverill: Makes sense @tractorcow - I've pushed an updated commit

tractorcow:
I think I made one mistake... `\w` isn't going to capture namespace separators in the FQN. :) You might need to swap that for something more optimistic. Maybe just `.+`?

sminnee:
The upgrader meta-data file should have an instruction to rewrite `Page_Controller` to `PageController`.

robbieaverill:
@sminnee I've added the upgrade map, rebased and pushed

src/View/SSViewer.php
```diff
@@ -253,9 +253,9 @@ public static function get_templates_by_class($classOrObject, $suffix = '', $bas
             $templates[] = $template;
             $templates[] = ['type' => 'Includes', $template];
 
-            // If the class is "Page_Controller", look for Page.ss
-            if (stripos($class, '_controller') !== false) {
-                $templates[] = str_ireplace('_controller', '', $class) . $suffix;
+            // If the class is "PageController" (PSR-2 compatibility) or "Page_Controller" (legacy), look for Page.ss
+            if (preg_match('/^(?<name>.+[^\\])_?Controller$/i', $class, $matches)) {
```
tractorcow: Sorry, you'll need to quad-escape this `[^\\\\]`. Which collapses to `[^\\]` when evaluated by PHP, and `[^\]` when evaluated by regexp :D

tractorcow:
Cool, will merge if these pass. :)

robbieaverill:
@tractorcow I'd assumed it requires silverstripe/silverstripe-cms#1703 first, but we'll see :)

chillu:
@robbieaverill Do we need to mention this in the upgrading guide?

robbieaverill:
@chillu yes, I'd say so, since it's a deprecating change. I'll be looking at those docs at some stage this week, but will update my Google doc (ref: #6441) for now

PR #6449, opened 2017-01-03
ENHANCEMENT: Show the query count/query number when using showqueries

mikenz:


dhensby:
What is the `queryCount` in this context?

mikenz:
Just went back through the blames. It was previously outputing the query number:

https://github.com/silverstripe/silverstripe-framework/commit/16f876f64713668b246900a2e2ff5c043c32ec65

So I guess this is now a bug fix PR.

PR #6450, opened 2017-01-03
ENHANCEMENT: Use SiteTree instead of Page class in more places

mikenz:
This allows developers to extend SiteTree directly, instead of Page.


dhensby:
Hmm - surely framework shouldn't have any dependency on CMS at all...

It'd be better to remove these dependencies and replace them with hooks for the CMS module to use

mikenz:
I totally agree - but I'm not feeling that brave :-)

PR #6453, opened 2017-01-05
Switch from using ORM in ForeignKey to SQL query

textagroup:
Using a SQLSelect object to grab the dropdown list as a standard array seems to improve the memory usage and result in less DB queries, it also overcomes the problem of the dropdown being replaced with a numeric field when more then 100 results are turned.

The results I received from memory_get_usage and showqueries are below.

**Current result**
Memory: 27540712 (26.26 mb)
Main SQL Queries:
SELECT DISTINCT count(DISTINCT "SiteTree"."ID") AS "Count" FROM "SiteTree" 0.0001s

SELECT DISTINCT "SiteTree"."ClassName", "SiteTree"."LastEdited", "SiteTree"."Created", "SiteTree"."URLSegment", "SiteTree"."Title", "SiteTree"."MenuTitle", "SiteTree"."Content", "SiteTree"."MetaDescription", "SiteTree"."ExtraMeta", "SiteTree"."ShowInMenus", "SiteTree"."
ShowInSearch", "SiteTree"."Sort", "SiteTree"."HasBrokenFile", "SiteTree"."HasBrokenLink", "SiteTree"."ReportClass", "SiteTree"."CanViewType", "SiteTree"."CanEditType", "SiteTree"."Version", "SiteTree"."ParentID", "SiteTree"."ID", CASE WHEN "SiteTree"."ClassName" IS NOT
NULL THEN "SiteTree"."ClassName" ELSE 'SiteTree' END AS "RecordClassName" FROM "SiteTree" ORDER BY "SiteTree"."Sort" ASC 0.0003s

**With applied patch results**
Memory: 26630040 (25.39 mb)

Main SQL Queries:
SELECT * FROM SiteTree 0.0003s

### DIFF
--- a/model/fieldtypes/ForeignKey.php
+++ b/model/fieldtypes/ForeignKey.php
@@ -37,17 +37,16 @@ class ForeignKey extends Int {
 			$field = new UploadField($relationName, $title);
 		} else {
 			$titleField = (singleton($hasOneClass)->hasField('Title')) ? "Title" : "Name";
-			// Use a SQL query instead of the ORM to save on PHP memory, DB objects and queries
-			$list = array();
-			$sqlQuery = new SQLSelect();
-			$sqlQuery->setFrom($hasOneClass);
-
-			$result = $sqlQuery->execute();
-			foreach ($result as $row) {
-				$list[$row['ID']] = $row[$titleField];
+			$list = DataList::create($hasOneClass);
+			// Don't scaffold a dropdown for large tables, as making the list concrete
+			// might exceed the available PHP memory in creating too many DataObject instances
+			if($list->count() < 100) {
+				$field = new DropdownField($this->name, $title, $list->map('ID', $titleField));
+				$field->setEmptyString(' ');
+			} else {
+				$field = null;
 			}
-			$field = new DropdownField($this->name, $title, $list);
-			$field->setEmptyString(' ');
+
 		}
 
 		return $field;
--- a/model/fieldtypes/ForeignKey.php
+++ b/model/fieldtypes/ForeignKey.php
@@ -37,16 +37,17 @@ class ForeignKey extends Int {
 			$field = new UploadField($relationName, $title);
 		} else {
 			$titleField = (singleton($hasOneClass)->hasField('Title')) ? "Title" : "Name";
-			$list = DataList::create($hasOneClass);
-			// Don't scaffold a dropdown for large tables, as making the list concrete
-			// might exceed the available PHP memory in creating too many DataObject instances
-			if($list->count() < 100) {
-				$field = new DropdownField($this->name, $title, $list->map('ID', $titleField));
-				$field->setEmptyString(' ');
-			} else {
-				$field = new NumericField($this->name, $title);
-			}
+			// Use a SQL query instead of the ORM to save on PHP memory, DB objects and queries
+			$list = array();
+			$sqlQuery = new SQLSelect();
+			$sqlQuery->setFrom($hasOneClass);
 
+			$result = $sqlQuery->execute();
+			foreach ($result as $row) {
+				$list[$row['ID']] = $row[$titleField];
+			}
+			$field = new DropdownField($this->name, $title, $list);
+			$field->setEmptyString(' ');
 		}
 
 		return $field;


dhensby:
The issue here is that you're now missing out core logic that may be useful:

1. What if there is a `getName` method on the object that is meant to change the presentation of the value
2. What if the field is on another table that needs to be joined

It's definitely worth pursuing how we can keep the memory footprint down and trying to create sparsely populated objects if we can, but this I fear is cutting out too much core logic...

tractorcow:
Also what if there are 1000 rows in this table, and now you are iterating over all of them? The code here is attempting to prevent unnecessarily large tables from breaching memory limits. This is the purpose of the second query.

Your code fails if the class is `Page` but the `Title` field is on the base class (`SiteTree`), which is the issue @dhensby raised in his second point.

textagroup:
Your right they are valid points.
The main issue I was trying to overcome was the DataObject limit which I have come across a couple of times when working on sites with large SiteTrees.

I am wondering at the very least if we are going to limit the DataObjects and switch to a NumericField if we should at least explain this by setting a description on the numeric field to explain why the user is getting the following behaviour?

The way it currently works does confuse the user and also any developers who end up trying to troubleshoot the issue.

dhensby:
The current behaviour is not idea, I agree, it would be best if we did not do this and could efficiently iterate over large sets - this is probably best done within a separate thread or some performant way.

Holding thousands of rows as ID => Value shouldn't cause a big memory issue - the issue comes from instantiating the objects and holding them in memory - this is partly required and so we should see how we can do this effectively.

dhensby:
relevant: https://github.com/silverstripe/silverstripe-framework/pull/6501

tractorcow:
As of a previous change, it's now allowed to return null from a scaffold field method. We should consider that as an option, rather than scaffolding something that's not the least bit useful (text field for a foreign key ID).

textagroup:
Have experimented with a simple way of loading one object at a time with the following https://gist.github.com/textagroup/0a1add09a4bc6e86ea412ebb26a7de44

However I am guessing it would be quite possible on large datasets to simply time out.
Ideally a AJAX request may be a option will look into trying something with TreeDropdownField

tractorcow:
You could do something like:
 - query with a hard limit (e.g. 100 rows)
 - on result, check number of items in result. If count >= 100, indicating a potential partial return set, return null.
 - otherwise, iterate list and return results as list
 
 This will at most load in 100 items into memory, will only ever perform a single query, and will never return partial sets.

tractorcow:
👍  from me :D

model/fieldtypes/ForeignKey.php
```diff
@@ -37,17 +37,16 @@ public function scaffoldFormField($title = null, $params = null) {
...
-			foreach ($result as $row) {
-				$list[$row['ID']] = $row[$titleField];
+			$list = DataList::create($hasOneClass);
+			// Don't scaffold a dropdown for large tables, as making the list concrete
+			// might exceed the available PHP memory in creating too many DataObject instances
+			if($list->count() < 100) {
+				$field = new DropdownField($this->name, $title, $list->map('ID', $titleField));
+				$field->setEmptyString(' ');
+			} else {
+				$field = null;
```
textagroup: @tractorcow Updated the pull request to set the $field to null if it is equal or greater then 100.
However I am wondering if we should at least return a LabelField or LiteralField giving some info on the field being dropped.

The majority of general CMS users would be confused by this, so maybe we could use a LiteralField with a HTML comment which could be useful for a developer.

tractorcow: If it gets to that point, devs really need to start implementing their own getCMSFields for their models, since the auto-scaffolding can only go so far.

tractorcow:
I just realised that this is on the 3.5 branch; We can't remove a feature except in master, so I'm switching the target branch.

model/fieldtypes/ForeignKey.php
dhensby: So in the CMS if a object has over 100 records we just won't show a field? I don't think that's acceptable. With the new work using generators instead of iterators building large lists for forms should be much more efficient now...

Just removing a field at (what will appear to be) random isn't a good experience for a CMS user...

kinglozzer: I’m inclined to agree. If a site - say, 1 year after launching - hits the magic threshold of 100 objects, the field would disappear entirely from the CMS without warning. I think the current behaviour is far better (though still flawed)

dhensby: My view is that it'd be better for the default to load everything and for devs to have to set a limit themselves than for us to make that decision for them.

jonom:
I think we need a smarter form field built in to the framework for this to work well. Form scaffolding is super handy but seeing an ID number or null is not. Some kind of a searchable drop down maybe, like what's seen in grid field for adding existing, combined with a paginated list? Has_one is has_many's poor cousin in SilverStripe CRUD atm, except for file relations.

model/fieldtypes/ForeignKey.php
tractorcow: That's fine too. My concern is the pointlessness in showing a numeric field as though it were an acceptable fallback.

tractorcow:
I wouldn't mind an autocomplete-combobox field either. ;)

jonom:
A consistent experience for 'choose from existing' for has_one, has_many and many_many would be great. I.e. autocomplete searches on the same fields and results are displayed in the same format.

Maybe we could drop Chosen from Master and use [Select2](https://select2.github.io/) instead? I feel like this has been discussed a lot somewhere but I couldn't find anything :) It seems to have feature parity with Chosen plus a lot extra, including remote dataset support.

Sorry I'm straying from specifically addressing the contents of this PR but I think it would address the root problem. If we want CRUD to be reliable and scalable in SilverStripe out of the box I think we need something like this.

tractorcow:
Dropping chosen for select2 could be done in 4; Time is limited though, since alpha 5 is coming up in a few weeks.

sminnee:
@jonom with the focus on React in SS4, it looks like https://github.com/JedWatson/react-select is going to be introduced. If you have compelling reasons why Select2 would be better now would be the time to raise it ;-)

jonom:
@sminnee main weakness with Chosen was lack of remote dataset support, react-select has that and looks like quite a swiss army knife select field so looks good to me! Should let us improve the default form scaffolding when editing relationships where lots of records exist, which was the main priority for me.

tractorcow:
react-select is now in core.

tractorcow:
I think the scope of this issue has changed. It's now "we need a lazy-load single select front end field". I'll close this for now, unless are core committer wishes to re-open this issue (as is).

PR #6456, opened 2017-01-06
 FEATURE (update form hook): add extension hook to update form

Neumes:
Okay, I understand that this one's going to open a huge can of worms, and I can definitely see how it's going to be a less than ideal implementation, however if it's at least canned, the discussion is here for historical purposes.

There are some circumstances in the past where on work that I've done, it's been required to update to the form fields of all forms in a site, so this was patched in. A lot of the work done was on instances of the base form class, rather than an extended form object (e.g. CMSForm).

I'm aware that having a hook like this can be undesirable as it will also hook into CMSForm, GridfieldDetailForm, etc and can produce undesirable or unexpected results.

If, as I'm guessing this PR isn't going to make it through due to certain complications arising, I'd like thoughts on a better approach.

### DIFF
--- a/forms/Form.php
+++ b/forms/Form.php
@@ -272,6 +272,8 @@ class Form extends RequestHandler {
 		$this->securityToken = ($securityEnabled) ? new SecurityToken() : new NullSecurityToken();
 
 		$this->setupDefaultClasses();
+
+		$this->extend('updateForm');
 	}
 
 	/**


dhensby:
I think the better approach here is to use injector to override the class that `Form::create()` returns and then you can customise it as you like.

The PR for core would then be to make sure we use `Form::create()` instead of `new Form()` in any places we were still using direct instantiation.

Adding a hook to the constructor is too expensive an operation IMO.

SilbinaryWolf:
Perhaps a config then to enable the hook? The problem with your method is that modules that subclass "Form" wont have spam protection OOTB. (Which is the point of the PR)

The $this->extend() call is extremely light assuming the 'Form' class has zero extension_instances, so I don't see how it's an expensive operation.

dhensby:
What's the problem with how the existing spam protection extension works?

Neumes:
Honestly in this day and age with how many sites require forms, it's a bit silly to have spam protection outside of core functionality. The fact that two extra modules for one feature is getting a little ridiculous. Spam protection, not just csrf protection, should be a core feature if forms themselves are going to be a core feature. On top of this, there shouldn't be a requirement for explicitly enabling things like spam protection, or other fields, per form where for 90% of things are going to require it (maybe not cms side, but who knows really).

Second to this, overriding what `Form::create()` returns won't work for a form object which extends Form, unless the magic underpinning injector logic has changed to support this in 4, but I can't see a way to make it work via inheritance for objects by doing that.

tractorcow:
> it's a bit silly to have spam protection outside of core functionality. 

Having a feature in core locks users to use that single feature and removes optionality. That's the purpose of modularisation, so that you don't need to install components you don't need, or want to replace.

If you need injector modifications to all forms then you should use injector with custom code in your project.

```yaml
  SilverStripe\Forms\Form:
    calls:
      extend: [ extend, ['updateForm']]
```

PR #6459, opened 2017-01-09
Add X-Robots-Tag noindex,nofollow header from Security controller to prevent indexing

robbieaverill:
Search engines will index the `/Security*` routes, which is probably not wanted in most installations.

This PR adds the `X-Robots-Tag` header with `nofollow` and `noindex`.

Reference: https://developers.google.com/webmasters/control-crawl-index/docs/robots_meta_tag

### DIFF
--- a/security/Security.php
+++ b/security/Security.php
@@ -147,6 +147,14 @@ class Security extends Controller implements TemplateGlobalProvider {
 	private static $frame_options = 'SAMEORIGIN';
 
 	/**
+	 * Value of the X-Robots-Tag header (for the Security section)
+	 *
+	 * @config
+	 * @var string
+	 */
+	private static $robots_tag = 'noindex, nofollow';
+
+	/**
 	 * Get location of word list file
 	 *
 	 * @deprecated 4.0 Use the "Security.word_list" config setting instead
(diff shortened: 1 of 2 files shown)


security/Security.php
```diff
@@ -147,6 +147,14 @@ class Security extends Controller implements TemplateGlobalProvider {
 	private static $frame_options = 'SAMEORIGIN';
 
 	/**
+	 * Value of the X-Robots-Tag header (for the Security section)
+	 *
+	 * @config
+	 * @var string
+	 */
+	private static $robots_tag = 'noindex, nofollow';
```
dhensby: this should just be "none" - which is both "noindex" and "nofollow".

security/Security.php
```diff
@@ -326,6 +334,9 @@ public function init() {
 
 		// Prevent clickjacking, see https://developer.mozilla.org/en-US/docs/HTTP/X-Frame-Options
 		$this->getResponse()->addHeader('X-Frame-Options', $this->config()->frame_options);
+
+		// Prevent search engines from indexing the login page
+		$this->getResponse()->addHeader('X-Robots-Tag', $this->config()->robots_tag);
```
dhensby: Wrap this in an `if ($this->config()->robots_tag)` so devs can unset it without an empty header being sent

security/Security.php
robbieaverill: I'd read somewhere that the more explicit version has better browser support, will see if I can find it

robbieaverill: https://yoast.com/robots-meta-tags/

robbieaverill:
Gah - damn behat tests!

security/Security.php
dhensby: I totally don't buy that it has no effect with major search engines...

dhensby: But sure. We can stick with

dhensby:
Behat is broken on 3

robbieaverill:
I can't rerun the tests, but if you can Dan they should pass now that Damian has fixed stuff in the behat extension

Issue #6460, opened 2017-01-09
[ORM] <dataobject subclass>::create($record) doesn't call setters


chillu:
This is done on purpose, since `$record` should only be used to rehydrate an existing object from the database. The underlying assumption here is that values stores in the database have already gone through validation and setters before being saved in the database.

I've wrongly used a shorthand to create objects through `MyObject::create(['MyField' => 'MyValue'])` many times before - mostly in manually creating unit test fixtures. It's not very clear that its dangerious to use a DataObject constructor in this way, particularly if your setters have side effects or data sanitisation.

We should change the constructor to call setters, and use a more explicit API for rehydrating records from a database row (which can bypass setters).

Note that `DataObject->update()` *does* call setters.

Pull request with a 3.x PHPDoc update at https://github.com/silverstripe/silverstripe-framework/pull/6461 - it's documenting the status quo for now.

Related

 * [Missing abstraction layer between database and object values](https://github.com/silverstripe/silverstripe-framework/issues/4018)
 * [RFC: Consistent init methods](https://github.com/silverstripe/silverstripe-framework/issues/4916)

# Pull requests
* [x] https://github.com/silverstripe/silverstripe-framework/pull/8591

sminnee:
Potentially we could introduce a non-constructor approach for rehydrating an object.

sminnee:
My recommendation here would be to favour `<dataobjectclass>::create([ k => v, k => v]);` as the approach for creating new records.

We can then supply a factory to Injector to go through a path other than the current constructor.

In the meantime, we can say that failure to populate defaults or call setters is a bug, and resolve as a minor change.

sminnee:
Potentially the $isSingleton argument could be changed into a 3 option argument, $creationType:

 * DataObject::CREATE (default) - create a new object
 * DataObject::HYRDRATE - hydrate an object from the database
 * DataObject::SINGLETON or `true` - create a singleton for access to type data

I can't see a good way in PHP of having alternative constructors that bypass __construct().

sminnee:
We would want to resolve #1575 as well, to avoid adding further changes to singletons.

sminnee:
OK with #1575 all sorted, I've made a PR for a tri-state alternative to isSingleton at #8591

maxime-rainville:
PR has been merged.

PR #6461, opened 2017-01-09
Clarify DataObject::__construct()

chillu:
The $record argument should NOT be used for object construction, since it bypasses setters.
While the original PHPDoc hints at this (“internal systems”), it doesn’t make it clear enough.

See https://github.com/silverstripe/silverstripe-framework/issues/6460


tractorcow:
Test failures are pre-existing due to deterioration of behat on the 3 branch. Merging.

dhensby:
You'd use `update` to populate an object from an array

```php
$obj = new DataObject();
$obj->update($record);
```

This has always confused anyone I show :P

Issue #5638, opened 2016-06-01
Class-generated form HTML IDs


sminnee:
- Agree on new approach
- Implement

tractorcow:
In the security namespace rewrite I've opted to use the base name of the FQN to maintain backwards compatibilitity.

https://github.com/open-sausages/silverstripe-framework/blob/f6ea54632bdb0e3805f0e337548ffbaa67d383d8/forms/FormTemplateHelper.php#L46

Whatever we decide can replace this.

sminnee:
Closing as I don't think this is a major problem now that @tractorcow has a solution in place.

PR #6462, opened 2017-01-09
Fix for SiteTree icons within TreeDropdownField

colintucker:
(see also silverstripe/silverstripe-cms#980)

This PR adds a `raw2htmlclass()` static method to `SilverStripe\Core\Convert`, which converts a mixed-case, namespaced class such as:

`SilverStripe\CMS\Model\ErrorPage`

to:

`silverstripe-cms-model-errorpage`

The resulting string (or array of strings) is CSS-friendly, maintains the SS convention of lowercase CSS class names, and rectifies the CMS issue of broken namespaced `SiteTree` icons (silverstripe/silverstripe-cms#980) when used in conjuction with the accompanying CMS PR.

`SilverStripe\Forms\TreeDropdownField` has also been patched to make use of the new conversion method.

### DIFF
--- a/src/Core/Convert.php
+++ b/src/Core/Convert.php
@@ -103,7 +103,29 @@ class Convert
             );
         }
     }
-
+    
+    /**
+     * Convert a value to be suitable for an HTML class attribute. Replaces non
+     * supported characters with a hyphen and converts to lowercase.
+     *
+     * @param array|string $val String to escape, or array of strings
+     *
+     * @return array|string
+     */
+    public static function raw2htmlclass($val)
+    {
+        $val = self::raw2htmlid($val);
+        
+        if (is_array($val)) {
+            foreach ($val as $k => $v) {
+                $val[$k] = strtolower(str_replace('_', '-', $v));
+            }
+            return $val;
+        } else {
+            return strtolower(str_replace('_', '-', $val));
+        }
+    }
+    
     /**
      * Ensure that text is properly escaped for XML.
      *
--- a/src/Forms/TreeDropdownField.php
+++ b/src/Forms/TreeDropdownField.php
@@ -368,7 +368,7 @@ class TreeDropdownField extends FormField
                 Convert::raw2xml($self->getName()),
                 Convert::raw2xml($child->$keyField),
                 Convert::raw2xml($child->$keyField),
-                Convert::raw2xml($child->class),
+                Convert::raw2htmlclass($child->class),
                 Convert::raw2xml($child->markingClasses($self->numChildrenMethod)),
                 ($self->nodeIsDisabled($child)) ? 'disabled' : '',
                 (int)$child->ID,


dhensby:
This idea was previously rejected in https://github.com/silverstripe/silverstripe-framework/pull/4402

mikenz:
The default page icon class names will need to be updated in https://github.com/silverstripe/silverstripe-framework/blob/master/admin/client/src/styles/legacy/_style.scss#L941

colintucker:
@dhensby So it looks like a no to adding additional statics to `Convert`?  Nonetheless, we still need to process the namespaced class names somehow that appear in the HTML, because the backslashes will not work correctly to assign styles.  IMHO it's a good idea to keep this functionality in the one spot, hence `Convert`, rather than c+p the same conversion code in multiple locations. Suggestions?

dhensby:
I somewhat agree - @chillu do you care to reconsider your position?

chillu:
`raw2htmlclass()` is a bit more general purpose than `fqcn2htmlname()`, so that's a good start. It's a bit odd though that we use underscores in `raw2htmlid()`, but then dashes in `raw2htmlclass`. That seems like a stylistic preference (CSS more commonly uses dashes), rather than anything to do with the `id` vs `class` attributes on a HTML element. Running a BEM-style class through this converter will break your CSS for example, since it relies on underscores (which is perfectly valid in class names).

Both are ambiguous converters, since there's different constraints in HTML4 and HTML5 (see http://stackoverflow.com/questions/70579/what-are-valid-values-for-the-id-attribute-in-html#79022). HTML4 disallows backslashes for example, but HTML5 doesn't. When using those class names in CSS, further constraints apply (backslashes disallowed). 

There's also `raw2att()` ("Convert a value to be suitable for an XML attribute") and `raw2htmlatt()` ("Convert a value to be suitable for an HTML attribute") already. Internally they use `raw2xml()`, which is probably a bit broad (just escaping quotes). 

Oh, and `raw2htmlname()`, which states "Convert a value to be suitable for an HTML ID attribute", so directly conflicts with `raw2htmlid()` - and does different stuff. 

So we need to figure out if the purpose of these converters is to "comply with spec", or "enforce common practice". At the moment it's very messy, and this addition just makes it slightly worse.

chillu:
Some historical context: https://github.com/silverstripe/silverstripe-framework/issues/5638

colintucker:
Cheers @chillu for the update.  The underscores vs dashes thing really was to maintain the convention of SS typically using mixed case and underscores for ids, and lower case with dashes for classes, i.e.:

```html
<div id="Form_EditForm_Title_Holder" class="form-group field text">
```

That still appears to be the case in v4, at least for now.

>Oh, and raw2htmlname(), which states "Convert a value to be suitable for an HTML ID attribute", so directly conflicts with raw2htmlid() - and does different stuff.

Looks like a copy+paste doc block error. 😉

Agreed `Convert` is a bit messy right now, but I'd really like to see a centralised place where class names can be safely converted to attributes, without having to c+p conversion one-liners everywhere.

chillu:
From the outside, it looks like `raw2htmlclass()` should do the same thing as `raw2htmlid()`, but it's much more opinionated: It claims to "replace unsupported characters", yet replaces underscores as well, which are definitely supported. And the lowercase conversion is purely a convention we happen to have in the CMS. Could you please make this clear in the docs? Say something about "if you just want to sanitise user input, use raw2htmlatt()`. Also, test coverage please :)

flamerohr:
Closing, as the pull request which would have utilised this has closed, and the `TreeDropdownField::tree()` method has changed drastically enough that it no longer needs the new `raw2htmlclass()` method either.

Feel free to open again if there is a valid need for this, I don't really agree with replacing `_` for `-` in class names, sorry.

PR #6463, opened 2017-01-09
Fix for v4: Admin returns "Too many pages" for subpages below top level

oilee80:
Simple logic error during porting to new version it appears

If Number of Children is greater than threshold then return the Too Many pages response


oilee80:
Closing as will branch off master and re-submit

PR #6466, opened 2017-01-09
NEW Move to SwiftMailer powered Emails

dhensby:
This PR rewrites our `Mailer` and `Email` classes to be powered by [SwiftMailer](https://github.com/swiftmailer/swiftmailer/).

`Email` now acts as a proxy for `Swift_Message` and adds our framework's rendering over the top.

`Mailer` converts an `Email` object into a `Swift_Message` and sends it using `Swift_Mailer` using the `Swift_MailTransport` by default (php's `mail` function) just to keep things the same as they are now.

todo:

- [x] Docs


src/Control/Email/Email.php
```diff
@@ -3,766 +3,618 @@
...
+        $message = static::create();
+        $message->setTemplate($template);
+        $message->setData($data);
+
+        // if a callback is provided, call it against the message
+        if (is_callable($callback)) {
+            $callback($message);
         }
 
+        $message->render();
```
kinglozzer: Do devs have to call this manually now? Shouldn’t this just be called by `send()`?

src/Control/Email/Email.php
```diff
@@ -3,766 +3,618 @@
...
     {
-        $this->attachments[] = array(
-            'contents' => $data,
-            'filename' => $attachedFilename,
-            'mimetype' => $mimeType,
-        );
+        $swiftMessage->setDate(DBDatetime::now()->Format('U'));
+        if ($defaultFrom = $this->config()->admin_email) {
+            $swiftMessage->setFrom($defaultFrom);
+        }
```
kinglozzer: Won’t this mean that if you use this method, you’re forced to use `admin_email` (if set)? i.e. if I set a default from address, how can I override it on a per-email basis? I think this needs to check if a from address has already been set

src/Control/Email/Email.php
dhensby: `render` is the replacement of `populateTemplate` which had to be called to "render" the email in v3.

So yes, devs do have to call this themselves if they don't use the callback method for creating emails, but this is not a change in behaviour IMO.

The problem with `send` calling render is what if a dev has manually set the body of the email? it'll be overridden and it'd be impossible to send a manually constructed email.

src/Control/Email/Mailer.php
```diff
@@ -2,10 +2,10 @@
 
 namespace SilverStripe\Control\Email;
 
-use InvalidArgumentException;
-use SilverStripe\Control\HTTP;
-use SilverStripe\Core\Convert;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Object;
+use Swift_Mailer;
+use Swift_MailTransport;
```
kinglozzer: There’s a lot of Swift-specific code in this class, which shouldn’t really be inherited by modules that provide other mailers. Would it be better to make `Mailer` an interface?

```php
interface Mailer
{
    public function send($message);
}

class SwiftMailer implements Mailer
{
    public function send($message)
    {
        $swiftMessage = $message->getSwiftMessage();
        return $this->sendSwift($swiftMessage);
    }
}

// Email.php
Injector::inst()->get('SilverStripe\Email\Mailer')->send($this);
```

src/Control/Email/Email.php
dhensby: That's a good point. I'll update this

kinglozzer:
So I’ve just had a look at SwiftMailer in a bit more detail (I’d never used it before), and I’m guessing the idea here is that you’d write or use a transporter for Swift rather than your own `Mailer` implementation? How easy is it to do?

src/Control/Email/Mailer.php
dhensby: So I've taken a bit of a look at how Laravel does it and it's quite closely coupled - mostly because devs can write their own transports (which is basically the bulk of what they'd probably need to do).

There are a lot of out of the box transports, too, which would likely cover most of what people need.

src/Control/Email/Email.php
kinglozzer: How about if we just check if the body is empty in `render()`? It looks like the old implementation would’ve overridden the message body: https://github.com/dhensby/silverstripe-framework/blob/4b7d31377bf8e4e5415da8dd9b5f172967a407a9/src/Control/Email/Email.php#L620-L624 and https://github.com/dhensby/silverstripe-framework/blob/4b7d31377bf8e4e5415da8dd9b5f172967a407a9/src/Control/Email/Email.php#L494-L526

dhensby:
> I’m guessing the idea here is that you’d write or use a transporter for Swift rather than your own Mailer implementation?

Yep - pretty much - SwiftMailer is pretty popular and there are transports for all the major mail services.

src/Control/Email/Email.php
dhensby: Interesting - we could check if the body is populated or not, though I do like the certainty of basically saying "you have to render your email" over the ambiguity of it...

dhensby:
I've addressed the 2 issues and pushed up

docs/en/02_Developer_Guides/10_Email/index.md
```diff
@@ -7,10 +7,22 @@ covers how to create an `Email` instance, customise it with a HTML template, the
...
-server configuration or use one of the third party SMTP services like [SparkPost](https://github.com/lekoala/silverstripe-sparkpost) 
-and [Postmark](https://github.com/fullscreeninteractive/silverstripe-postmarkmailer).
+SilverStripe provides an API over the top of the [SwiftMailer](http://swiftmailer.org/) PHP library which comes with an
+extensive list of "transports" for sending mail via different services. 
+
+Out of the box, SilverStripe will use the built-in PHP `mail()` command via the `Swift_MailTransport` class. If you'd
+like to use a more robust transport to send mail you can swap out the transport used by the `Mailer` via config:
+
+```yml
+Mailer:
```
tractorcow: Needs the full namespace: `SilverStripe\Control\Email\Mailer:`

src/Control/Email/Mailer.php
```diff
@@ -15,513 +13,124 @@ class Mailer extends Object
...
+    public function getSwiftMailer()
     {
-        $result = '';
-        foreach ($headers as $key => $value) {
-            $result .= "$key: $value\n";
-        }
-        if ($body) {
-            $result .= "\n$body";
+        if (!$this->swift) {
+            $this->setSwiftMailer(new \Swift_Mailer($this->getTransport()));
```
tractorcow: We could inject this with dependencies, instead of via PHP.

```
Injector:
  Swift_Transport: Swift_MailTransport
  Swift_Mailer:
    constructor:
      - '%$Swift_Transport'
  SilverStripe\Control\Email\Mailer:
    properties:
      SwiftMailer: '%$Swift_Mailer'
```

How about this?

src/Control/Email/Mailer.php
tractorcow: I second @kinglozzer's suggestion, however I don't think it's critical to fully-decouple from swiftmailer for the sake of the initial implementation. It's not a merge blocker. :)

src/Control/Email/Mailer.php
```diff
@@ -15,513 +13,124 @@ class Mailer extends Object
...
-     *
-     * @param string $to To
-     * @param string $from From
-     * @param string $subject Subject
-     * @param array $attachedFiles List of attachments
-     * @param array $customHeaders User specified headers
-     * @param string $fullBody Prepared message
-     * @param array $headers Prepared headers
-     * @return mixed Return false if failure, or list of arguments if success
+     * @return mixed
```
tractorcow: Please use `@return static` or `@return Mailer` for correct type hinting. :)

src/Control/Email/Email.php
```diff
@@ -31,738 +16,646 @@ class Email extends ViewableData
...
      */
-    protected $parseVariables_done = false;
+    private $swiftMessage;
 
     /**
-     * @var string $ss_template The name of the used template (without *.ss extension)
+     * @var string The name of the used template (without *.ss extension)
      */
-    protected $ss_template = 'GenericEmail';
+    private $template = "SilverStripe\\Email\\Email";
```
tractorcow: Can you please use the full qualified path for the template? The convention is that it should match the same class name.

I.e.

`private static $template = 'SilverStripe\Control\Email\Email`, or just `private static $template = self::class`.

src/Control/Email/Email.php
```diff
@@ -31,738 +16,646 @@ class Email extends ViewableData
...
+        if ($mime) {
+            $attachment->setContentType($mime);
+        }
+        $this->getSwiftMessage()->attach($attachment);
+
+        return $this;
+    }
 
-            // Rewrite relative URLs
-            $this->body = HTTP::absoluteURLs($fullBody);
```
tractorcow: Very easy to miss, but we need to ensure we still do this somewhere. Perhaps convert absoluteURLs() in Email::render() ? Relative links won't work in email content once sent.

src/Control/Email/Email.php
```diff
@@ -31,738 +16,646 @@ class Email extends ViewableData
...
+    public function addAttachmentFromData($data, $name, $mime = null)
```
tractorcow: PHPDoc missing. :)

src/Control/Email/Email.php
```diff
@@ -31,738 +16,646 @@ class Email extends ViewableData
...
     {
-        return $this->subject;
+        $this->getSwiftMessage()->addFrom($address, $name);
+
+        return $this;
     }
 
     /**
-     * @return string|null
+     * @return array
```
tractorcow: Should be `@return string`

src/Control/Email/Email.php
```diff
@@ -31,738 +16,646 @@ class Email extends ViewableData
...
     }
 
     /**
-     * Load all the template variables into the internal variables, including
-     * the template into body.  Called before send() or debugSend()
-     * $isPlain=true will cause the template to be ignored, otherwise the GenericEmail template will be used
-     * and it won't be plain email :)
-     *
-     * @param bool $isPlain
+     * @param string $path Path to file
```
tractorcow: Please phpdoc $alias and $mime arguments too.

src/Control/Email/Email.php
```diff
@@ -31,738 +16,646 @@ class Email extends ViewableData
...
-            unset($headers['Cc']);
-            unset($headers['Bcc']);
-        } else {
-            if ($this->cc) {
-                $headers['Cc'] = $this->cc;
-            }
-            if ($this->bcc) {
-                $headers['Bcc'] = $this->bcc;
-            }
+    /**
```
tractorcow: I think this needs a good description. Also please document the method signature for the `$callback`

src/Control/Email/Email.php
```diff
@@ -31,738 +16,646 @@ class Email extends ViewableData
...
+     * @param string $template
+     * @param array|ViewableData $data
+     * @param callable|null $callback
+     * @return static
+     */
+    public static function create_from_callback($template, $data, $callback = null)
```
tractorcow: I'm not sure why this is valuable; All it does is wrap Email::create()->setArg()->setArg()? All of these setters return $this, so there's nothing preventing them from being chained. I think we should just drop this method altogether.

src/Control/Email/Email.php
tractorcow: On second thoughts, see my below comment.

src/Control/Email/Email.php
```diff
@@ -31,738 +16,646 @@ class Email extends ViewableData
...
-        if ($bccAllTo = $this->config()->bcc_all_emails_to) {
-            if (!empty($headers['Bcc']) && trim($headers['Bcc'])) {
-                $headers['Bcc'] .= ', ' . $bccAllTo;
-            } else {
-                $headers['Bcc'] = $bccAllTo;
-            }
+    /**
+     * Send the message to the recipients
+     *
+     * @return array|bool true if successful or array of failed recipients
```
tractorcow: It might be more concise to have a simple boolean response, and document that the user can call $message->getFailedRecipients() to get the list of failed recipients.

src/Control/Email/Mailer.php
```diff
@@ -15,513 +13,124 @@ class Mailer extends Object
...
-        $htmlPart = $this->prepareHTMLSubmessage($htmlContent);
-        list($fullBody, $headers) = $this->encodeMultipart(
-            array($plainPart, $htmlPart),
-            "multipart/alternative"
-        );
-
-        // Send prepared message
-        return $this->sendPreparedMessage($to, $from, $subject, $attachedFiles, $customHeaders, $fullBody, $headers);
-    }
+    private $failedRecipients = array();
```
tractorcow: I don't think that the service should be storing data related to specific invocations of Mailer::sendSwift(); I suggest to pass this back to Email::send() and have it stored on the email instance instead. Otherwise it dilutes the role of Mailer as a service.

src/Control/Email/Email.php
tractorcow: Sorry, `private $template` not `private static $template` in this case. :)

src/Security/Member.php
```diff
@@ -953,17 +954,16 @@ public function onBeforeWrite()
 
         // We don't send emails out on dev/tests sites to prevent accidentally spamming users.
         // However, if TestMailer is in use this isn't a risk.
-        if ((Director::isLive() || Email::mailer() instanceof TestMailer)
+        if ((Director::isLive() || Mailer::get_inst() instanceof TestMailer)
```
sminnee: Would it be better to rely on `Injector` for these global services?

src/Control/Email/Mailer.php
```diff
@@ -15,513 +13,124 @@ class Mailer extends Object
 {
```
sminnee: May as well remove 'extends Object' while we're tidying up this class.

src/Control/Email/Mailer.php
sminnee: +1. This approach would be in keeping with what we've done for assets and logging.

docs/en/02_Developer_Guides/10_Email/index.md
```diff
@@ -7,10 +7,22 @@ covers how to create an `Email` instance, customise it with a HTML template, the
...
+  swift_transport: Swift_SendmailTransport
+```
+
+or in PHP
+
+```php
```
sminnee: Don't encourage the use of PHP-based configuration I think.

src/Control/Email/Email.php
```diff
@@ -31,738 +16,646 @@ class Email extends ViewableData
...
+        if (!$this->getBody()) {
+            $this->render();
         }
-
-        if ($sendAllfrom = $this->config()->send_all_emails_from) {
-            if ($from) {
-                $subject .= " [from $from]";
-            }
-            $from = $sendAllfrom;
+        $numFailed = Mailer::send_message($this);
```
sminnee: Would change this to `Injector::inst()->get(Mailer::class)->send($this)` if Mailer is to become an interface.

src/Control/Email/Email.php
```diff
@@ -31,738 +16,646 @@ class Email extends ViewableData
...
+        if ($numFailed) {
+            return Mailer::get_inst()->getFailedRecipients();
```
sminnee: Would change this to `Injector::inst()->get(Mailer::class)->getFailedRecipients()` if Mailer is to become an interface.

src/Control/Email/Mailer.php
sminnee: In addition the plugins could be provided to Swift_Mailer with the Injector's 'calls' option.

I think that SilverStripe\Control\Email\Mailer could be an interface, with the current impelmentation called `SilverStripe\Control\Email\SwiftMailer`or something.

The interface would probably be very simple:

```php
namespace SilverStripe\Control\Email;
interface Mailer {
  public function send(Email $message);
  public function getFailedRecipients();
}  
```

That being the case, rather than `Mailer::get_inst()` we'd probably want to use `Injector::inst()->get(Mailer::class)`.

Alternatively, we could do away with Mailer altogether and use Injector to provide a `Swift_Mailer` class? We can use the yaml config to add the transport and plugins.

For test mailing we could make a test transport instead?

dhensby: >Alternatively, we could do away with Mailer altogether and use Injector to provide a Swift_Mailer class? We can use the yaml config to add the transport and plugins.

The problem with this is we can't send `Email` objects directly to the `Mailer` for sending as we can now.

We'd basically just be implementing raw `SwiftMailer` without any convenience methods.

src/Control/Email/Mailer.php
dhensby: Ok - good point.

src/Control/Email/Email.php
dhensby: it returns an array see my test `testGetSetSender`

src/Control/Email/Email.php
dhensby: phew - good spot. added

dhensby: Actually, I've moved it to `setBody` so we always do this on the body regardless of render. However, maybe the `Mailer` or even `SwiftPlugin` should do it to absolutely make sure nothing can get through?

src/Control/Email/Email.php
dhensby: Hmm - originally it was a nice convenience method which would call `render` for you (it had been mandatory to call `render` by hand otherwise) but after @kinglozzer's suggestion of checking if the body is empty, this may be less desirable...

src/Control/Email/Email.php
dhensby: The problem is some recipients could be successful and some rejected which means that a true/false return value is misleading. If one recipient is rejected then we should return false, but that doesn't mean a complete failure and so more debugging is needed.

I suppose that just means better documentation is required...

src/Security/Member.php
dhensby: I'd like to get rid of these references completely. Are tests not run under "live" mode?

dhensby:
I've implemented pretty much all the feedback.

> @sminnee In addition the plugins could be provided to Swift_Mailer with the Injector's 'calls' option.

How could we do this and allow devs to specify many plugins that should be loaded?

This could work for our case, but doesn't provide a nice way for devs to add extra plugins:

```yml
  Swift_Plugin: SilverStripe\Control\Email\SwiftPlugin
Swift_Mailer:
  calls:
    SwiftPlugin:
      - registerPlugin
      - [ '%$Swift_Plugin' ]
```

todo:

- [ ] More docs highlighting how to do lots of this stuff with Injector

dhensby:
Looks like https://github.com/silverstripe/silverstripe-behat-extension's `TestMailer` needs re-writing

src/Control/Email/SwiftMailer.php
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * @var Swift_Mailer
+     */
+    private $swift;
+
+    /**
+     * @return static
+     */
+    public static function get_inst()
```
sminnee: Delete this.

src/Control/Email/SwiftMailer.php
```diff
@@ -0,0 +1,96 @@
...
+    {
+        return Injector::inst()->get(Mailer::class);
+    }
+
+    /**
+     * @param Email $message
+     * @return int
+     */
+    public static function send_message($message)
```
sminnee: Delete this.

src/Control/Email/SwiftMailer.php
```diff
@@ -0,0 +1,96 @@
...
+        $message->setFailedRecipients($failedRecipients);
+
+        return $result != 0;
+    }
+
+    /**
+     * @param Swift_Message $message
+     * @return int
+     */
+    public function sendSwift($message, &$failedRecipients = null)
```
sminnee: I think it would be tidier if this was protected because it's not part of the interface-provided API. If people start calling it directly it becomes harder for another implementation of the Mailer interface to be dropped in as a replacement.

docs/en/02_Developer_Guides/10_Email/index.md
```diff
@@ -7,10 +7,16 @@ covers how to create an `Email` instance, customise it with a HTML template, the
...
-and [Postmark](https://github.com/fullscreeninteractive/silverstripe-postmarkmailer).
+SilverStripe provides an API over the top of the [SwiftMailer](http://swiftmailer.org/) PHP library which comes with an
+extensive list of "transports" for sending mail via different services. 
+
+Out of the box, SilverStripe will use the built-in PHP `mail()` command via the `Swift_MailTransport` class. If you'd
+like to use a more robust transport to send mail you can swap out the transport used by the `Mailer` via config:
+
+```yml
+SilverStripe\Control\Email\Mailer:
+  swift_transport: Swift_SendmailTransport
```
sminnee: Would it be better to recommend that developers override the `Swift_Transport` service instead? Otherwise the resulting service definition list will include a service called "Swift_Transport"' that's not actually the swift transport being used...

```yml
Swift_Transport:
  class: Swift_SendmailTransport
```

sminnee:
There are styleguide breakages that need fixing. Call `composer run-script lint` to get details.

sminnee:
> How could we do this and allow devs to specify many plugins that should be loaded?
> This could work for our case, but doesn't provide a nice way for devs to add extra plugins:

Fair enough; happy to leave it for now.

src/Control/Email/Email.php
tractorcow: Yep good job.

sminnee:
@dhensby if you can rebase to the latest master and fix up src/Dev/TestMailer.php, behat should Just Work(tm)

src/Control/Email/SwiftMailer.php
dhensby: good point

docs/en/02_Developer_Guides/10_Email/index.md
dhensby: Yes, I'll amend

_config/config.yml
```diff
@@ -6,6 +6,19 @@ SilverStripe\Core\Injector\Injector:
     class: SilverStripe\Forms\Schema\FormSchema
   FixtureFactory:
     class: SilverStripe\Dev\FixtureFactory
+  Swift_Transport: Swift_MailTransport
+  Swift_Plugin: SilverStripe\Control\Email\SwiftPlugin
+  Swift_Mailer:
+    constructor:
+      - '%$Swift_Transport'
+    calls:
+      SwfitPlugin:
```
tractorcow: Typo: `SwiftPlugin`

src/Control/Email/Email.php
```diff
@@ -31,738 +18,641 @@ class Email extends ViewableData
...
-    public function attachFileFromString($data, $attachedFilename, $mimeType = null)
+    public function setSwiftMessage($swiftMessage)
     {
-        $this->attachments[] = array(
-            'contents' => $data,
-            'filename' => $attachedFilename,
-            'mimetype' => $mimeType,
-        );
+        $swiftMessage->setDate(DBDatetime::now()->Format('U'));
+        if (!$swiftMessage->getFrom() && $defaultFrom = $this->config()->admin_email) {
```
tractorcow: Operator precedence; Should the assignment `=` be surrounded with braces, since it's lower than the `&&`?

src/Control/Email/Email.php
tractorcow: If I look at SimpleMessage::getSender(), it says `@return string`, so something is wrong here.

tractorcow: There is no testGetSetSender =(

tractorcow: I guess SimpleMessage::getSender() is wrong?

src/Control/Email/SwiftMailer.php
```diff
@@ -0,0 +1,96 @@
...
+use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\Core\Injector\Injector;
+use Swift_Mailer;
+use Swift_Message;
+
+/**
+ * Mailer objects are responsible for actually sending emails.
+ * The default Mailer class will use PHP's mail() function.
+ */
+class SwiftMailer
```
tractorcow: You forgot `implements Mailer` interface.

src/Control/Email/Mailer.php
```diff
@@ -2,526 +2,17 @@
...
-
-        // Ensure name component cannot be mistaken for an email address
-        if ($openBracket) {
-            $emailAddress = str_replace('@', '', substr($emailAddress, 0, $openBracket))
-                . substr($emailAddress, $openBracket);
-        }
-
-        return $emailAddress;
-    }
+    public function getFailedRecipients();
```
tractorcow: Was this going to be moved to Email? `SwiftMailer` doesn't implement this, so we can just remove this I guess.

src/Control/Email/SwiftMailer.php
tractorcow: Missing PHPDoc.

src/Control/Email/SwiftMailer.php
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * @param Swift_Mailer $swift
+     * @return $this
+     */
+    public function setSwiftMailer($swift)
+    {
+        // register any required plugins
+        foreach ($this->config()->get('swift_plugins') as $plugin) {
```
tractorcow: This needs to be a simple setter. Use YML injection to set these plugins on the injected $swift object.

src/Control/Email/Email.php
```diff
@@ -31,738 +18,641 @@ class Email extends ViewableData
...
-        return preg_match("!^$addr_spec$!", $email) === 1;
+    /**
+     * @return array|bool
+     */
+    public function sendPlain()
+    {
+        $this->getSwiftMessage()->setContentType('text/plain');
+        if (!$this->getBody()) {
+            $this->render();
+            $this->setBody(Convert::xml2raw($this->getBody()));
```
tractorcow: This code looks super risky, especially if you are sending an email multiple times, you are at risk of double-encoding.

Why not have separate messages for plain / html (with separate setBody / setPlainBody) and allow them to be set independently? You can always convert between the two if attempting to send one which is blank. :)

tractorcow: Use cases:

 - Send a HTML message with a html template, I want the system to automatically pick a plain text version for me (as current implementation)
 - Send a HTML message with a html template, and separate plain text string provided for plain text version
 - Sending a plain text message only with a setPlainBody().
 - Sending a html and plain message, I will provide setBody() / setPlainBody() for each.

src/Control/Email/SwiftMailer.php
sminnee: The question is how does that let site developers add 1 more plug-in in addition to what's currently existing.

src/Control/Email/Email.php
sminnee: Yeah, agreed. Perhaps there should be a `getPlainBody()` / `setPlainBody()` pair, and this transformation is called is PlainBody isn't set explicitly?

sminnee: Alternatively `getPlainTextSwiftMessage()` could be a second getter that transforms the email prior to passing it to the mailers. This, however, would mean that the mailer would need to accept a swift message object rather than our email. I don't have a problem with that, but it's another tweak to the interface.

src/Control/Email/Email.php
dhensby: I think so!

src/Control/Email/Email.php
dhensby: Because the assignment is the last condition it won't be messed up an assign a boolean value, however I'll wrap it in parentheses for clarity.

src/Control/Email/SwiftMailer.php
dhensby: thanks

src/Control/Email/SwiftMailer.php
dhensby: @tractorcow see https://github.com/silverstripe/silverstripe-framework/pull/6466#issuecomment-271880190

src/Control/Email/Mailer.php
dhensby: yep - removed

src/Control/Email/Email.php
dhensby: If the `Convert::xml2plain` is inside the `if (!$this->getBody())` block then there's no risk of double encoding because we render and then take a copy we don't repeatedly set the plain part.

Adding the plain part automatically is all round risky. I don't like that the `send` method does anything other than send the email (it tries to guess if it should be rendered and then attempts to render a plain part too).

In Laravel you have to set the plain part explicitly and can even define a template for the plain version - perhaps we should do something similar.

We could have something like.

```php
$email = Email::create()
    ->setHTMLTemplate('MyTemplate')
    ->setPlainTemplate('MyTemplate_plain')
    ->send(); //still performs a render
```

Or if there's no plain template we create a `generatePlainPart` method that does what we have at the moment.

I think the biggest challenge is with reusing `Email` objects because I'm not so sure how easy it is to detach the previously generated plain part... I'll look into it.

src/Control/Email/SwiftMailer.php
tractorcow: Ok, that's good enough I guess.

src/Control/Email/Email.php
tractorcow: I have code in HTMLText that generates plain text from HTML if you want to re-use that. :)

tractorcow:
```php
/**
     * Get plain-text version
     *
     * @return string
     */
    public function Plain()
    {
        // Preserve line breaks
        $text = preg_replace('/\<br(\s*)?\/?\>/i', "\n", $this->RAW());

        // Convert paragraph breaks to multi-lines
        $text = preg_replace('/\<\/p\>/i', "\n\n", $text);

        // Strip out HTML tags
        $text = strip_tags($text);

        // Implode >3 consecutive linebreaks into 2
        $text = preg_replace('~(\R){2,}~', "\n\n", $text);

        // Decode HTML entities back to plain text
        return trim(Convert::xml2raw($text));
    }
```

_config/config.yml
tractorcow: Typo still here in your last push :)

src/Dev/TestMailer.php
```diff
@@ -2,88 +2,30 @@
...
-            'PlainContent' => $plainContent,
-            'HtmlContent' => $htmlContent,
-
-            'AttachedFiles' => $attachedFiles,
-            'CustomHeaders' => $customHeaders,
-            'InlineImages' => $inlineImages,
-        ]);
+    public function send($message)
+    {
+        $this->saveEmail($message);
```
sminnee: Did you *have* to change the API for saveEmail? Now the behat test mailer is broken again :-(

dhensby: :(

An email isn't really suitably represented by the old parameters and I thought best that we just store the email object instead...

This test suite is passing - where are the failures coming from?

tractorcow: `$state->emails[] = array_filter($data);` in TestMailer::saveEmail() in behat extension. array_filter will fail on an Email instance.

tractorcow: TestSession state needs to be serialized, which is the main reason these were originally array format. I think maybe we should revert this back?

dhensby: Whatever you thinks best :)

dhensby: But if they rely on some of the specific data (like `Type`) then behat tests will need changing because that's not really a thing anymore.... or we'll need to have a "best guess" at type.

tractorcow: All of my guesses are the best.

tractorcow:
@sminnee I've pushed a change to revert api changes to TestMailer.

tractorcow:
Happy to merge once CI passes.

sminnee:
OK I'm happy with this. Say the word and I'll squash & merge

PR #6472, opened 2017-01-11
feat(FieldList): Add nestedFields() function to return a flat array of all deeply nested fields in a FieldList.

silbinarywolf:
feat(FieldList): Add nestedFields() function to return a flat array of all deeply nested fields in a FieldList.

### DIFF
--- a/src/Forms/FieldList.php
+++ b/src/Forms/FieldList.php
@@ -59,6 +59,38 @@ class FieldList extends ArrayList
     }
 
     /**
+     * Iterate over each field in the current list recursively
+     *
+     * @param callable $callback
+     */
+    public function recursiveWalk(callable $callback)
+    {
+        $stack = $this->toArray();
+        while (!empty($stack)) {
+            /** @var FormField $field */
+            $field = array_shift($stack);
+            $callback($field);
+            if ($field instanceof CompositeField) {
+                $stack = array_merge($field->getChildren()->toArray(), $stack);
+            }
+        }
+    }
+
+    /**
+     * Return a flattened list of all fields
+     *
+     * @return static
+     */
+    public function flattenFields()
+    {
+        $fields = [];
+        $this->recursiveWalk(function (FormField $field) use (&$fields) {
+            $fields[] = $field;
+        });
+        return static::create($fields);
+    }
+
+    /**
      * Return a sequential set of all fields that have data.  This excludes wrapper composite fields
      * as well as heading / help text fields.
      *
@@ -66,8 +98,19 @@ class FieldList extends ArrayList
      */
     public function dataFields()
     {
-        if (!$this->sequentialSet) {
-            $this->collateDataFields($this->sequentialSet);
+        if (empty($this->sequentialSet)) {
+            $fields = [];
+            $this->recursiveWalk(function (FormField $field) use (&$fields) {
+                if (!$field->hasData()) {
+                    return;
+                }
+                $name = $field->getName();
+                if (isset($fields[$name])) {
+                    $this->fieldNameError($field, __FUNCTION__);
+                }
+                $fields[$name] = $field;
+            });
+            $this->sequentialSet = $fields;
         }
         return $this->sequentialSet;
     }
@@ -77,10 +120,60 @@ class FieldList extends ArrayList
      */
     public function saveableFields()
     {
-        if (!$this->sequentialSaveableSet) {
-            $this->collateDataFields($this->sequentialSaveableSet, true);
+        if (empty($this->sequentialSaveableSet)) {
+            $fields = [];
+            $this->recursiveWalk(function (FormField $field) use (&$fields) {
+                if (!$field->canSubmitValue()) {
+                    return;
+                }
+                $name = $field->getName();
+                if (isset($fields[$name])) {
+                    $this->fieldNameError($field, __FUNCTION__);
+                }
+                $fields[$name] = $field;
+            });
+            $this->sequentialSaveableSet = $fields;
         }
         return $this->sequentialSaveableSet;
+    }
+
+    /**
+     * Return array of all field names
+     *
+     * @return array
+     */
+    public function dataFieldNames()
+    {
+        return array_keys($this->dataFields());
+    }
+
+    /**
+     * Trigger an error for duplicate field names
+     *
+     * @param FormField $field
+     * @param $functionName
+     */
+    protected function fieldNameError(FormField $field, $functionName)
+    {
+        if ($this->form) {
+            $errorSuffix = sprintf(
+                " in your '%s' form called '%s'",
+                get_class($this->form),
+                $this->form->getName()
+            );
+        } else {
+            $errorSuffix = '';
+        }
+
+        user_error(
+            sprintf(
+                "%s() I noticed that a field called '%s' appears twice%s",
+                $functionName,
+                $field->getName(),
+                $errorSuffix
+            ),
+            E_USER_ERROR
+        );
     }
 
     protected function flushFieldsCache()
@@ -89,8 +182,14 @@ class FieldList extends ArrayList
         $this->sequentialSaveableSet = null;
     }
 
+    /**
+     * @deprecated 4.1..5.0 Please use dataFields or saveableFields
+     * @param $list
+     * @param bool $saveableOnly
+     */
     protected function collateDataFields(&$list, $saveableOnly = false)
     {
+        Deprecation::notice('5.0', 'Please use dataFields or SaveableFields');
         if (!isset($list)) {
             $list = array();
         }
--- a/tests/php/Forms/FieldListTest.php
+++ b/tests/php/Forms/FieldListTest.php
@@ -3,7 +3,10 @@
 namespace SilverStripe\Forms\Tests;
 
 use SilverStripe\Dev\SapphireTest;
+use SilverStripe\Forms\ConfirmedPasswordField;
 use SilverStripe\Forms\FieldList;
+use SilverStripe\Forms\FormField;
+use SilverStripe\Forms\LiteralField;
 use SilverStripe\Forms\Tab;
 use SilverStripe\Forms\TextField;
 use SilverStripe\Forms\EmailField;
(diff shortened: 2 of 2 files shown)


dhensby:
This is effectively a "flatten" method? Can you not make use of `collateDataFields`?

SilbinaryWolf:
I can't if I want to modify FieldGroup's nested (or any other CompositeField for that matter).

sminnee:
Having both `collateDataFields()` as this method seems like a bloated API.

I would probably:

 * Switch to `master`
 * Have it return a new FieldList object for consistency, and have it make use of a generator so that we can efficiently chain things.
 * Update `ArrayList::filterByCallback()` to use a generator
 * Update `collateDataFields()` or its caller to use `nestedFields()` chained with some kind of `filterByCallback()` call.

sminnee:
Sorry - the public APIs are `dataFields()` and `saveableFields()`. I think both of these should be deprecated in favour of a filter chained onto `nestedFields()`

sminnee:
It would probably be worth adding `FieldList::fieldNames()` that does the equivalent of `array_keys($fields->dataFields())`

So this:
```php
array_keys($fields->saveableFields())
```

Would become:
```php
$fields->nestedFields()->filterByCallback(function ($field) { return $fields->canSubmitValue(); })->fieldNames();
```

Which is a bit of a mouthful. We could add filterByMethod?

```php
// saveableFields()
$fields->nestedFields()->filterByMethod('canSubmitValue', true)->fieldNames();

// dataFIelds()
$fields->nestedFields()->filterByMethod('hasData', true)->fieldNames();
```

Arguably the `true` could be optional

SilbinaryWolf:
I'm not sure about deprecating dataFields(), it's nice sugar and it's what you want to work with *most* of the time.

I'll still look toward making it work as you suggest, seems nice and clean.

SilbinaryWolf:
I've done some of what you requested,
However for the 'fieldNames' part, I feel like DataList/ArrayList would benefit from a function that turns the results into an associative array, ie:
```
/**
 * Use Cases: - Get records where the key is ID
			  - Get dataFields ('Name' must be unique)
 */
function toUniqueAssoc($fieldName) {
	foreach ($this as $record) {
		$key = $record->$fieldName;
		if (isset($result[$key])) {
			throw new LogicException(__FUNCTION__."() The property '$fieldName' with value '$field->getName()' appears twice.");
		}
		$result[$key] = $record;
	}
}
```

```
/**
 * Use Case: Group a list of records by their has_one (ie. Group)
 */
function toAssoc($fieldName) {
	foreach ($this as $record) {
		$key = $record->$fieldName;
		$result[$key][] = $record;
	}
}

**Comment by SilbinaryWolf** [comment]
@sminnee Just need further feedback on this one.

File: src/Forms/FieldList.php [hunk 1]
```diff
@@ -59,28 +59,86 @@ public function __clone()
...
+            }
+            $fields[$name] = $field;
+        }
+        return $this->sequentialSaveableSet = $fields;
+    }
+    
+    protected function fieldNameError(FormField $field, $functionName)
+    {
+        if ($this->form) {
+            $errSuffix = " in your '{$this->form->class}' form called '" . $this->form->Name() . "'";
```
dhensby: Please use one form of injecting variables into strings. Either concatenate or variable expansion, but not both.

src/Forms/FieldList.php
```diff
@@ -59,28 +59,86 @@ public function __clone()
...
+        } else {
+            $errSuffix = '';
         }
-        return $this->sequentialSaveableSet;
+        user_error(
+            $functionName."() I noticed that a field called '$field->getName()' appears twice$errSuffix.",
```
dhensby: Please use one form of injecting variables into strings. Either concatenate or variable expansion, but not both.

src/Forms/FieldList.php
```diff
@@ -59,28 +59,86 @@ public function __clone()
     }
 
     /**
+     * Return an array of all fields, including deeply nested fields.
+     *
+     * @var FieldList
+     */
+    public function nestedFields()
```
dhensby: I  think this method name is ambiguous, wouldn't something like "flatten" be better?

SilbinaryWolf: I'll see what the community says on the name of this.

src/Forms/FieldList.php
SilbinaryWolf: Will do.

src/Forms/FieldList.php
SilbinaryWolf: Will do.

src/Forms/FieldList.php
colintucker: I prefer `flattenFields()` 👍

SilbinaryWolf: @michalkleiner +1'd in the slack chat. So yeah, I'll do flattenFields()

SilbinaryWolf:
Adjusted according to feedback.
However, I'm not sure 'recursiveWalk' seems to be the way to go (or at least the approach I took).

I get the feeling this would be way less performant than what I had before as it'll call a function per field now. It didn't really make the code anymore concise so I'm not convinced the cost is worth it.

dhensby:
This looks good. I'd be very surprised if looping one the fields once was slower than doing it twice, tbh. 

There's just a few code style issues to fix. As you've removed collatedatafields, can you put this into the upgrade guide, please.

SilbinaryWolf:
@dhensby collateDataFields() was a protected method, just confirming to make sure you still want it documented regardless :)

dhensby:
> collateDataFields() was a protected method, just confirming to make sure you still want it documented regardless :)

I think it's important for anyone who's made custom form fields, so yes please

SilbinaryWolf:
Wasn't aware that people extended the FieldList class. What file am I meant to modify for the upgrade guide?

dhensby:
This is where you should note BC breakages: https://github.com/silverstripe/silverstripe-framework/blob/master/docs/en/04_Changelogs/4.0.0.md

SilbinaryWolf:
Thanks! Done!

SilbinaryWolf:
@dhensby @sminnee Not sure what else is required for this one.

dhensby:
@SilbinaryWolf sorry for the silence on this - I'm going to rebase this and have a final read over

dhensby:
OK - so we're past the point of being able to remove APIs from 4.x - so we'll need to re-instate `collateDataFields` I'm afraid :/

silbinarywolf:
@dhensby Happy to re-instate `collateDataFields` on this, but that's all the effort I'm willing to put forward on this as this was done in my personal time.

If this is all we need to get it in, then I'll happily finish it off. If some sort of tests / etc need to be written to hit a specific coverage %, then I'll leave this for someone else to pick up and finish.

dhensby:
OK - I've added some tests and deprecated `collateDataFields`,

dhensby:
lots of failures! what have i done?!

silbinarywolf:
@dhensby Thanks for picking this up. I'm glad the effort isn't going to waste.

dhensby:
🤞 green ticks coming

src/Forms/FieldList.php
```diff
@@ -58,6 +58,38 @@ public function __clone()
...
+            $field = array_shift($stack);
+            $callback($field);
+            if ($field instanceof CompositeField) {
+                $stack = array_merge($field->getChildren()->toArray(), $stack);
+            }
+        }
+    }
+
+    /**
+     * Return a flattened array of all fields
```
tractorcow: Doesn't return an array; It returns a new fieldlist object.

src/Forms/FieldList.php
```diff
@@ -58,6 +58,38 @@ public function __clone()
...
+     *
+     * @return static
+     */
+    public function flattenFields()
+    {
+        $fields = [];
+        $this->recursiveWalk(function (FormField $field) use (&$fields) {
+            $fields[] = $field;
```
tractorcow: If we are flattening fields, should we skip composite fields? E.g. merge in fields in nested FieldGroups, but not the FieldGroup themselves?

src/Forms/FieldList.php
tractorcow: Doesn't return an array; It returns a new fieldlist object.

src/Forms/FieldList.php
```diff
@@ -77,18 +120,67 @@ public function dataFields()
...
+     * Return array of all field names
+     *
+     * @return array
+     */
+    public function fieldNames()
+    {
+        return array_keys($this->dataFields());
+    }
+
+    protected function fieldNameError(FormField $field, $functionName)
```
tractorcow: PHPDoc please. :)

src/Forms/FieldList.php
```diff
@@ -77,18 +120,67 @@ public function dataFields()
...
+                " in your '%s' form called '%s'",
+                get_class($this->form),
+                $this->form->getName()
+            );
+        } else {
+            $errorSuffix = '';
+        }
+
+        user_error(
+            sprintf(
```
tractorcow: Should we use exceptions, so that user-code can catch this?

src/Forms/FieldList.php
```diff
@@ -77,18 +120,67 @@ public function dataFields()
...
     }
 
+    /**
+     * @deprecated 4.1..5.0 Please use dataFields or saveableFields
+     * @param $list
+     * @param bool $saveableOnly
+     */
     protected function collateDataFields(&$list, $saveableOnly = false)
     {
         if (!isset($list)) {
```
tractorcow: Can you add a `Deprecation::notice('5.0', 'message');` as well? :P

src/Forms/FieldList.php
```diff
@@ -77,18 +120,67 @@ public function dataFields()
...
+                }
+                $fields[$name] = $field;
+            });
+            $this->sequentialSaveableSet = $fields;
         }
         return $this->sequentialSaveableSet;
     }
 
+    /**
+     * Return array of all field names
```
tractorcow: Actually, only data field names. Should this method be `dataFieldNames()` ? Or should $dataFieldsOnly be a flag maybe? Up to you. :)

src/Forms/FieldList.php
dhensby: Not for flattening explicitly,  don't think... it is a field regardless of it's type. I'd much rather `flattenFields` be agnostic about this point...

src/Forms/FieldList.php
dhensby: tbh, the only reason I've added this is because it was requested here: https://github.com/silverstripe/silverstripe-framework/pull/6472#issuecomment-272305378

I'm not sure it adds a huge amount of value - do you think it's worth adding?

src/Forms/FieldList.php
dhensby: Possibly, what exception? `DuplicateFieldNameException`?

dhensby: is that a breaking change? maybe we can't do that...

tractorcow: Well, this error breaks the code, so it's going to break anyway.

tractorcow: `Exception` is fine, or even `LogicException`. You don't need a separate class.

tractorcow:
Close enough I guess!

tests/php/Forms/FieldListTest.php
```diff
@@ -32,6 +35,90 @@
...
+            new LiteralField('Literal', 'Literal content'),
+            new CompositeField(
+                $day = new TextField('Day'),
+                $month = new TextField('Month'),
+                $year = new TextField('Year')
+            ),
+        );
+        $fieldList = new FieldList($fields);
+
+        $this->assertEquals(['Name', 'Email', 'Hidden', 'Day', 'Month', 'Year'], $fieldList->fieldNames());
```
tractorcow: Change to `-> dataFieldNames()`

dhensby: fixed

tractorcow:
Thanks everyone for bearing with us on this issue for so long. :)

PR #6475, opened 2017-01-11
ENHANCEMENT Remove jquery-ui button() api from default HTML editor dialog

tractorcow:
Necessary to fix CMS behat tests, now that we have merged the necessary changes into framework module.

Merge this before https://github.com/silverstripe/silverstripe-cms/pull/1699 to get those tests working again.

### DIFF
--- a/admin/client/src/legacy/HtmlEditorField.js
+++ b/admin/client/src/legacy/HtmlEditorField.js
@@ -389,7 +389,7 @@ $.entwine('ss', function($) {
                 'Insert'
               ),
               'data-icon': 'accept',
-              class: 'ss-ui-action-constructive media-insert',
+              class: 'btn action btn-primary media-insert',
               click: function() {
                 $(this).find('form').submit();
               }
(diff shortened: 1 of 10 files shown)


tractorcow:
Note: Please do not merge until I've managed to confirm these fixes resolve the outstanding behat issues.

admin/client/src/legacy/HtmlEditorField.js
```diff
@@ -991,10 +991,12 @@ $.entwine('ss', function($) {
...
       // Disable "insert" button if no files are selected
-      this.closest('ui-dialog')
-        .find('ui-dialog-buttonpane .media-insert')
-        .button(hasItems ? 'enable' : 'disable')
-        .toggleClass('ui-state-disabled', !hasItems);
+      this.closest('.ui-dialog')
+        .find('.ui-dialog-buttonpane .media-insert')
+        .toggleClass('ui-state-disabled', !hasItems)
+        .each(function() {
+          this.disabled = !hasItems;
```
flamerohr: `$(this).prop('disabled', !hasItems);` would be a safer

admin/client/src/legacy/HtmlEditorField.js
```diff
@@ -1174,14 +1176,20 @@ $.entwine('ss', function($) {
...
       val = val.replace(/^https?:\/\//i, '');
-      if (orig !== val) this.val(val);
+      if (orig !== val) {
+        this.val(val);
+      }
 
-      this.getAddButton().button(!!val ? 'enable' : 'disable');
-      return !!val;
+      this.getAddButton().each(function() {
+        this.disabled = !hasVal;
```
flamerohr: same as above `$(this).prop('disabled', !hasVal);` would be a safer

tractorcow:
ok, tests are passing; Once this is merged https://github.com/silverstripe/silverstripe-cms/pull/1699 should be repaired.

chillu:
The broken tests on Travis are due to the fixed logic targeting https://github.com/silverstripe/silverstripe-cms/pull/1699. I've run some spot checks locally and looks good. Notably, this behaviour fixed the *old* "insert media" dialog which is only available when you don't have asset-admin installed.

PR #6482, opened 2017-01-11
Client-side email validation rule

chillu:
This fixes the "admin/test-react" UI for frameworktest,
which we're using frequently to sanity check our stuff.
Not having this UI available is impacting our ability to determien regressions,
e.g. from styling merges.


tractorcow:
Tests still expect an error being thrown; Can you do `expects(console.warn).toBeCalled()` ?

chillu:
Sweet, updated

PR #6484, opened 2017-01-11
Enforce unique Folder names, use AssetNameGenerator

chillu:
Regression from 3.x. Also makes behaviour between File->write()
and Upload->load() more consistent. The latter was using AssetNameGenerator,
which suffixed "-v2", but File->setName() was using a "-2" suffix.

This also fixes a bug where folder duplicates try to set an extension,
so creating a duplicate folder called "test" results in "test-2."
with a suffixed dot.

Merge together with https://github.com/silverstripe/silverstripe-graphql/pull/49

### DIFF
--- a/src/Assets/File.php
+++ b/src/Assets/File.php
@@ -10,6 +10,7 @@ use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Core\Convert;
 use SilverStripe\Control\Director;
 use SilverStripe\Control\Controller;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\DatetimeField;
 use SilverStripe\Forms\FieldList;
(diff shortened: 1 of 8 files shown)


src/Assets/File.php
```diff
@@ -741,24 +742,26 @@ public function setName($name)
 
         // We might have just turned it blank, so check again.
         if (!$name) {
-            $name = 'new-folder';
+            $name = 'new-' . strtolower(__CLASS__);
```
tractorcow: Please use basename, since `__CLASS__` has slashes in it (FQN).

tractorcow:
Question  @chillu ; Can you ensure you set ParentID before Name when creating folders?

sminnee:
> Question @chillu ; Can you ensure you set ParentID before Name when creating folders?

I don't think it's realistic to enforce that across all apps. Perhaps it's better to add the uniqueness suffix as part of onBeforeWrite()?

sminnee:
onBeforeWrite() would also pave the way for us addressing this in a way that doesn't suffer from race conditions. However, it would necessitate that we obtain a write lock prior to querying for duplicates....

We don't currently have any good scheme for requesting a write lock within `onBeforeWrite()` that will be released once the write has successfully completed.

chillu:
> Perhaps it's better to add the uniqueness suffix as part of onBeforeWrite()?

@sminnee Good call. I was approaching this from the perspective of a bugfix, hence tried to minimise the change. But the circumstances have changed since 3.x, from memory there was a reason why we enforced uniqueness earlier (something with filesystem syncing order?). Anyway, onBeforeWrite makes this a lot more readable.

Regarding read locks, we're already in a much better position now - the timespan between `onBeforeWrite()` and `write()` is usually shorter than between `setName()` and `write()`. Introducing read locks would be nice, but I'd consider it scope creep for this PR.

src/Assets/File.php
```diff
@@ -639,12 +640,60 @@ protected function onBeforeWrite()
...
+        $name = $filter->filter($name);
+
+        // We might have just turned it blank, so check again.
+        if (!$name) {
+            $name = 'new-' . strtolower(basename($this->class));
         }
 
-        // Propegate changes to the AssetStore and update the DBFile field
+        // Check for duplicates when the name has changed (or is set for the first time)
+        if ($this->isChanged('Name')) {
```
tractorcow: Between the above `$name = ` assignments and this line no change detection will occur. If $name was blank earlier `isChanged('Name')` will be false.

src/Assets/Folder.php
```diff
@@ -129,7 +129,8 @@ public function getIcon()
      */
     public function setTitle($title)
     {
-        $this->setName($title);
+        $this->setField('Name', $title);
```
tractorcow: You need a setField('Title') as well, since that's removed from setName().

src/Assets/File.php
```diff
@@ -639,12 +640,65 @@ protected function onBeforeWrite()
...
+        }
+
+        $filter = FileNameFilter::create();
+        if ($name) {
+            // Fix illegal characters
+            $name = $filter->filter($name);
+        } else {
+            // Default to file name
+            $changed = true;
+            $name = $this->i18n_singular_name();
```
chillu: I’m using singular_name() rather than basename(__CLASS__), since that’s abusing basename() for FQCN stripping, and doesn’t work consistently (backslashes only stripped on windows?)

src/Assets/File.php
```diff
@@ -639,12 +640,65 @@ protected function onBeforeWrite()
...
+            $name = $filter->filter($name);
```
chillu: This filters out non-ASCII characters in localised file names, but I can live with that (given that normal users will be forced to set a file name by virtue of uploading files)

src/Assets/File.php
tractorcow: That's fine. :)

PR #6487, opened 2017-01-11
FIX Remap versioned ClassNames during build process

robbieaverill:
## Problem

Missed ClassName remapping causes the default "Getting Started" page to render if the home page is a custom class (extending Page). Since the Live table will use `HomePage` and not be able to resolve it to a controller, then falls back to `SiteTree`  and renders a default template.

Re-publishing the page from admin fixes the problem, as the ClassName is updated to a FQCN.

---

## Fix

This prevents the current behaviour from happening during `DatabaseAdmin::doBuild`:

```
mysql> select a.ClassName, b.ClassName, c.ClassName from SiteTree as a inner join SiteTree_Versions as b on a.ID = b.ID inner join SiteTree_Live as c on a.ID = c.ID where a.URLSegment = 'home';
+--------------------------+-----------+-----------+
| ClassName                | ClassName | ClassName |
+--------------------------+-----------+-----------+
| MyProject\Model\HomePage | Page      | HomePage  |
+--------------------------+-----------+-----------+
1 row in set (0.00 sec)
```

Updated behaviour:

```
mysql> select a.ClassName, b.ClassName, c.ClassName from SiteTree as a inner join SiteTree_Versions as b on a.ID = b.ID inner join SiteTree_Live as c on a.ID = c.ID where a.URLSegment = 'home';
+--------------------------+-----------+--------------------------+
| ClassName                | ClassName | ClassName                |
+--------------------------+-----------+--------------------------+
| MyProject\Model\HomePage | Page      | MyProject\Model\HomePage |
+--------------------------+-----------+--------------------------+
```

### DIFF
--- a/src/ORM/DatabaseAdmin.php
+++ b/src/ORM/DatabaseAdmin.php
@@ -9,6 +9,7 @@ use SilverStripe\Core\Manifest\ClassLoader;
 use SilverStripe\Dev\SapphireTest;
 use SilverStripe\Dev\TestOnly;
 use SilverStripe\Dev\Deprecation;
+use SilverStripe\ORM\Versioning\Versioned;
 use SilverStripe\Security\Security;
 use SilverStripe\Security\Permission;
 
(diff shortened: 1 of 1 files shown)


dhensby:
This seems a pretty delicate operation - is it possible to unit test this?

robbieaverill:
@dhensby I guess if I manually create some dodgy database records and trigger a build within a SapphireTest, that might do it?

dhensby:
Yes, but might also be a massive pain.. if you're feeling brave....

robbieaverill:
I'm not, at the moment :)

src/ORM/DatabaseAdmin.php
```diff
@@ -325,10 +326,19 @@ public function doBuild($quiet = false, $populate = true, $testMode = false)
...
-                    DB::prepared_query(
-                        "UPDATE \"$table\" SET \"ClassName\" = ? WHERE \"ClassName\" = ?",
-                        [ $newClassName, $oldClassName ]
-                    );
+
+                    $updateQuery = "UPDATE \"$table%s\" SET \"ClassName\" = ? WHERE \"ClassName\" = ?";
+                    $updateQueries = [sprintf($updateQuery, '')];
+
+                    // Remap versioned table ClassName values as well
+                    if (singleton($newClassName)->has_extension(Versioned::class)) {
```
tractorcow: Not all versioned classes have a _Live table. Some have _Versions only.

See

```php
/**
     * Constructor arg to specify that versioning only is active on this record.
     */
    const VERSIONED = 'Versioned';
```

PR #6489, opened 2017-01-12
FIX Regression in using template_main to render the Security area

robbieaverill:
* Introduced in 0cf477d36
* Should use the "template_main" static to decide which model to use for rendering this page
* Assuming SiteTree means any requirements added in Page and assumed to work everywhere will not

### DIFF
--- a/src/Security/Security.php
+++ b/src/Security/Security.php
@@ -514,15 +514,17 @@ class Security extends Controller implements TemplateGlobalProvider
             return $this;
         }
 
-        // Use sitetree pages to render the security page
-        $tmpPage = new SiteTree();
+        // Use the default setting for which Page to use to render the security page
+        $pageClass = (string) $this->stat('template_main');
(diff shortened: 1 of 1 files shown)


src/Security/Security.php
```diff
@@ -514,15 +514,17 @@ protected function getResponseController($title)
             return $this;
         }
 
-        // Use sitetree pages to render the security page
-        $tmpPage = new SiteTree();
+        // Use the default setting for which Page to use to render the security page
+        $pageClass = (string) $this->stat('template_main');
```
tractorcow: Oh, risky. template_main could be a path to a template file (e.g. `/themes/mytheme/templates/Security.ss`.

I think this change is probably safe for now, but we should use a class_exists() and document template_main appropriately.

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/pull/6492

PR #6496, opened 2017-01-13
API Enhancement: Allow "removeComponentsByType" to remove multiple component"s"

robbieaverill:
As the method name `removeComponentsByType` suggests, this PR allows the method to remove component"s".
Existing behaviour is not affected.

### DIFF
--- a/src/Forms/GridField/GridFieldConfig.php
+++ b/src/Forms/GridField/GridFieldConfig.php
@@ -91,15 +91,22 @@ class GridFieldConfig extends Object
     }
 
     /**
-     * @param string $type Class name or interface
+     * @param string|string[] $types Class name or interface, or an array of the same
      * @return $this
      */
-    public function removeComponentsByType($type)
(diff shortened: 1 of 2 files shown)


src/Forms/GridField/GridFieldConfig.php
```diff
@@ -91,12 +91,19 @@ public function removeComponent(GridFieldComponent $component)
...
     /**
-     * @param string $type Class name or interface
+     * @param string|string[] $types Class name or interface, or an array of the same
      * @return $this
      */
-    public function removeComponentsByType($type)
+    public function removeComponentsByType($types)
     {
-        $components = $this->getComponentsByType($type);
+        if (is_array($types)) {
```
tractorcow: One code path to do one task please... just convert the string to a single-item array and then we don't duplicate logic. :)

robbieaverill: Updated, thanks!

PR #6498, opened 2017-01-13
API Allow users to act-as another

tractorcow:


src/Security/Member.php
```diff
@@ -872,12 +872,50 @@ public static function currentUser()
...
+     * @param $callback
+     */
+    public static function actAs($member, $callback)
+    {
+        $id = ($member instanceof Member ? $member->ID : $member) ?: 0;
+        $previousID = static::$overrideID;
+        static::$overrideID = $id;
+        try {
+            return $callback();
+        } finally {
```
sminnee: So finally happens before the return?

tractorcow: It happens after `return` is invoked, but before the method actually returns. Weird I know!

sminnee: #learning. Good solution!

tests/php/Security/MemberTest.php
```diff
@@ -1402,4 +1402,66 @@ public function testCurrentUser()
         $userFromSession = Member::currentUser();
         $this->assertEquals($adminMember->ID, $userFromSession->ID);
     }
+
+    /**
+     * @covers Member::actAs()
```
dhensby: This is breaking the coverage test run

kinglozzer: Does it need the FQCN maybe?

tractorcow:
I've fixed to use FQN.

src/Security/Member.php
chillu: My only gripe here is that any associated error handling is executed in the current user context as well - project might've been configured to output stack traces based on user permissions, rather than environment context. But that's enough of an edge case to ignore it.

PR #6499, opened 2017-01-13
feat(ListDecorator): Add setList() function. Useful for keeping a decorator/paginated list configuration intact while modifying the underlying list.

SilbinaryWolf:
feat(ListDecorator): Add setList() function. Useful for keeping a decorator/paginated list configuration intact while modifying the underlying list.

### DIFF
--- a/model/ListDecorator.php
+++ b/model/ListDecorator.php
@@ -15,8 +15,7 @@ abstract class SS_ListDecorator extends ViewableData implements SS_List, SS_Sort
 	protected $list;
 
 	public function __construct(SS_List $list) {
-		$this->list     = $list;
-		$this->failover = $this->list;
+		$this->setList($list);
 
 		parent::__construct();
 	}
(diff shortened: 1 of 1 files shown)


model/ListDecorator.php
```diff
@@ -30,6 +29,21 @@ public function getList() {
 		return $this->list;
 	}
 
+    /**
+     * Set the list this decorator wraps around.
+     *
+     * Useful for keeping a decorator/paginated list configuration intact while modifying
+     * the underlying list.
+     *
+     * @return SS_List
```
tractorcow: `@return $this` is more correct :P but so minor I'm not going to block merge.

tractorcow:
Please use correct git tags in future. E.g. `BUG <desc>`, `ENHANCEMENT <desc>`, `API <desc>`.

 `type(class)` isn't a bad standard, but it's not the one we'd like and our release notes generation doesn't support it sorry. =(

SilbinaryWolf:
Noted. After I finish this PR: https://github.com/silverstripe/silverstripe-framework/pull/6472

I'll make sure to follow your standards.

PR #6500, opened 2017-01-13
Fix shift graphql to use post data 

flamerohr:
See https://github.com/silverstripe/silverstripe-graphql/issues/48 for more details


dhensby:
some linting issues

PR #6501, opened 2017-01-13
DON’T MERGE - NEW: Make DataList::getIterator a generator

kinglozzer:
Huge hat tip to @SilbinaryWolf for suggesting this.

---

Currently, `DataList::getIterator()` wraps the result of `DataList::toArray()` in an `ArrayIterator`. `DataList::toArray()` will load every single row in the result into memory at once, creating a `DataObject` instance for each of them.

After this change, only the current row will be loaded into a `DataObject` and then discarded before the next one is loaded.

Crudely written benchmarks below:

### Small collections
No discernible difference in CPU or memory.
```php
$pages = Page::get()->limit(20);
$count = 0;
foreach ($pages as $page) {
	$count++;
}
echo $count;
```
screen shot 2017-01-13 at 10 01 30

### Large collections
Significant memory saving, some CPU sacrificed.
```php
$pages = Page::get()->limit(2000);
$count = 0;
foreach ($pages as $page) {
	$count++;
}
echo $count;
```
screen shot 2017-01-13 at 11 25 25

Happy to do some more benchmarks if anyone has anything they’d like to see

### DIFF
--- a/src/ORM/DataList.php
+++ b/src/ORM/DataList.php
@@ -815,11 +815,13 @@ class DataList extends ViewableData implements SS_List, Filterable, Sortable, Li
      * Returns an Iterator for this DataList.
      * This function allows you to use DataLists in foreach loops
      *
-     * @return ArrayIterator
+     * @return Generator
      */
     public function getIterator()
     {
-        return new ArrayIterator($this->toArray());
+        foreach ($this->dataQuery->query()->execute() as $row) {
+            yield $this->createDataObject($row);
+        }
     }
 
     /**


kinglozzer:
The increase in CPU is the result of a lot of calls to `Query::seek()` (one for each row), which in turn will act on the query result object to seek to the given row. That _can_ be avoided by immediately loading all of the database rows in `DataListIterator::__construct()`, but still “lazily” constructing the DataObject instances, but that approach wipes out a significant amount of the memory saving.

kinglozzer:
Current blockers:

- SQLite3 (non-PDO) CPU usage skyrockets. There’s no `SQLite3Result::seek()`, so it currently just loops every row until it reaches the given offset. If there are 2000 rows, it has too loop over 1 row, then 2, then 3, all the way up to 2000 when we try to seek the last row. One possible solution would be to fetch and store every row once, then `seek()`, `numRecords()` etc could inspect that array. Downside: memory savings are all but gone, as every row now sits in memory. SQLite3 in PDO mode is unaffected by this.
- MSSQL module is [currently unable to seek](https://github.com/silverstripe/silverstripe-mssql/blob/master/code/SQLServerQuery.php#L46-L67) in non-PDO mode. Looking at the PHP docs for [`sqlsrv_fetch_array`](http://php.net/manual/en/function.sqlsrv-fetch-array.php), we _should_ be able to implement that ability... I just don’t currently have an environment to test this

kinglozzer:
I’ve tentatively pushed a third commit with some changes to `PDOQuery` that stops it from loading every result immediately, and instead allows it to seek through rows in the same way that other drivers do. Compared to the original benchmarks above, it saves an additional 4mb of peak memory usage.

dhensby:
I can only say how awesome this is.

I reckon the reason we don't see any memory gains on few records (~20) is because the garbage collector isn't being triggered.

sminnee:
Perhaps DataList::getIterator() could simply be a generator?

```php
function getIterator() {
  foreach($this->dataQuery as $row) {
    yield $this->createDataObject($row);
  }
}
```

It's possible that this will remove the use of `rewind()`; if not and we've still got issues with SQLite and MSSQL I'd suggest coding this in a way where the original behaviour is used for those databases.

sminnee:
Note: generators are awesome and much of the ORM could be rewritten to use them, now that we're 5.5+.

kinglozzer:
What an awesome idea! Look at that, 3 lines of code changed and a ~30% memory saving. ❤️

I’ve tested again on MySQL, MySQL+PDO, SQLite3, SQLite3+PDO and PostgreSQL, and performance is consistent across all of them.

sminnee:
Generators. Love 'em!

kinglozzer:
Adding a precautionary “DO NOT MERGE”. Travis is building against the wrong commit for some reason, there are plenty of failures.

sminnee:
I've cherry-picked this into #6518, so I can close this PR.

PR #6505, opened 2017-01-14
BUG Ensure correct regeneration of ConfigManifest if only one of the …

### LINKED ISSUE #6467: ConfigManifest does not proper validate existence of cache files
As reported by @nyeholt there are [cases](https://github.com/silverstripe/silverstripe-framework/issues/2706#issuecomment-29967406) when the cache files of ConfigManifest are getting deleted. If only the php_config_source file is deleted, but not the variant_key_spec, there will be a fatal error on the website: As the phpConfigSources is not populated, we will see an error in activateConfig. 
It can be only recovered manually by flushing the cache. From my point of view we cannot assume that both cache files are always delete together. We just recently faced this scenario on one of our websites. I propose to add the [suggested fix](https://github.com/silverstripe/silverstripe-framework/issues/2706#issuecomment-29967406) to ConfigManifest.

_edit: mixed the two files, added information what causes an error_

sb-relaxt-at:
…cache files is missing (fixes #6467)

### DIFF
--- a/core/manifest/ConfigManifest.php
+++ b/core/manifest/ConfigManifest.php
@@ -99,7 +99,7 @@ class SS_ConfigManifest {
 		}
 
 		// If we don't have a variantKeySpec (because we're forcing regen, or it just wasn't in the cache), generate it
-		if (false === $this->variantKeySpec) {
+		if (false === $this->phpConfigSources || false === $this->variantKeySpec) {
 			$this->regenerate($includeTests);
 		}
 
--- a/tests/core/manifest/ConfigManifestTest.php
+++ b/tests/core/manifest/ConfigManifestTest.php
@@ -153,6 +153,83 @@ class ConfigManifestTest extends SapphireTest {
 	}
 
 	/**
+	 * Test cache regeneration if all or some of the cache files are missing
+	 *
+	 * 1. Test regeneration if all cache files are missing
+	 * 2. Test regeneration if 'variant_key_spec' cache file is missing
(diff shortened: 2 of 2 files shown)


kinglozzer:
You’ve even managed to write a test for it, nice job!

sb-relaxt-at:
@kinglozzer thanks :) Yeah writing the test took a lot more time than fixing the issue ;)

dhensby:
If this bug affects 3.4 it'd be good to have it back ported so the issue is fixed in all our supported versions

kinglozzer:
Damn, sorry, I should’ve spotted that

sb-relaxt-at:
@dhensby: yes it affects 3.4, any special process or just another pr against 3.4?

Is 3.4 still in "full support"? The [timeline](https://www.silverstripe.org/software/roadmap/#title) suggests that it is only "limited support" with security fixes only.

dhensby:
Hmm - good point, though that timeline is quite fluid, you can see we are behind on the v4 part of it.

I suppose we can leave this out of 3.4 for now, then. Though I'd probably have merged it had it been against 3.4!

sminnee:
Yeah the timeline is more about the minimum level of support we commit to, rather than blocking additional contributions by individuals to the earlier versions. Regardless, it's fine to raise this here.

sb-relaxt-at:
Thanks for the additional informations! I have made a second PR #6513 in case we want to fix it in 3.4 as well.

PR #6510, opened 2017-01-16
BUG Fix root folder getFilename() returning incorrect path

tractorcow:
Root folder should not have a non-empty filename


tractorcow:
@chillu I have a feeling that the issue pre-dated that PR, but it needs to be fixed anyway. The default Name isn't necessary since you added the onBeforeWrite check.

PR #6511, opened 2017-01-16
FIX Reintroduce lost changes to CompositeField_holder.ss after namespacing

robbieaverill:
I'm not sure if this was deliberate or not, but a large amount of the CompositeField_holder.ss template was lost during namespacing (commit #8dd644d).

See [3.5](https://github.com/silverstripe/silverstripe-framework/blob/3ee8f505b75af39817d394bcf48623e092f43ed2/templates/forms/CompositeField_holder.ss) version compared with [current master](https://github.com/silverstripe/silverstripe-framework/blob/8dd644d25d4e01b6853922164440f8ac9c824f22/templates/SilverStripe/Forms/CompositeField_holder.ss).

In its current state, frontend composite fields do not render correctly. No labels are displayed as the individual field's holder templates are not used.

If I've mistaken this, and this was a deliberate change, please feel free to close this pull request!

### DIFF
--- a/templates/SilverStripe/Forms/CompositeField.ss
+++ b/templates/SilverStripe/Forms/CompositeField.ss
@@ -1,9 +1,9 @@
 <% loop $FieldList %>
 	<% if $ColumnCount %>
 		<div class="column-{$ColumnCount} $FirstLast">
-			$Field
+			$FieldHolder
 		</div>
 	<% else %>
-		$Field
+		$FieldHolder
 	<% end_if %>
 <% end_loop %>


robbieaverill:
cc @tractorcow

dhensby:
Test failures related...

robbieaverill:
Ok, those errors are happening because `SelectionGroup` extends `CompositeField` so will be inheriting the changed template.

@tractorcow are you happy that me "fixing" this regression is correct in this case? If so, I'll add the old version of this file as a template for `SelectionGroup_holder` to prevent this change from affecting it as well.

tractorcow:
These changes weren't lost after namespacing, they were intentionally made during the bootrapping process.

Each field is built with a fixed <Class>_holder.ss, and a <Class>.ss. Typically the holder includes the inner field, but in some cases this was not done consistently. As of 4.0, it is.

If you take a look at ComposetField.ss, you'll see this contains the missing code you were looking for, and this is what would be included with $Field in CompositeField_holder.ss

If fields are not displayed correctly on the front end, then that is a real bug. Can you please show me the code you used to scaffold such a broken form @robbieaverill ?

tractorcow:
Ok, I've tested the code locally on both 3.5 and 4.0. There is a bug, but it can be fixed by changing `$Field` to `$FieldHolder` in `CompositeField.ss`, rather than reverting the entire `CompositeField_holder.ss` template.

The bootstrap version of this field / holder already does this properly. We just need to update the default template.

PR #6513, opened 2017-01-16
BUG Ensure correct regeneration of ConfigManifest if only one of the …

sb-relaxt-at:
…cache files is missing

Backport of #6467


dhensby:
thanks

PR #6519, opened 2017-01-17
TinyMCE in React form

flamerohr:
I've created a HtmlEditorField component, which handles loading TinyMCE and the Insert Media modal plugins. Functionally works. The loading is handled by polling for the tinymce object in the global space.

Currently uses TinyMCE's built-in link plugin instead of SilverStripe's link plugin, I didn't have the time to build an entire Link modal in react, (and kind of out of the scope for which I was aiming for).

I ran out of time to add documentation or tests for this before my trip, but I put this here in hopes that someone could help with furthering this to get it merged!

NOTE: Field cannot be in the asset-admin seciton, it'll have some really bad side-effects when the Insert Media modal opens. Asset-admin would need to be updated to namespace the "type" it is opened as for this to work there.


sminnee:
Does this integrate with any FormSchema builders yet? If not, how do you use it? If it is, where is the code that hooks it in?

flamerohr:
It integrates with the FormBuilder like other FormFields in the Framework using react, if desired, can be added to the Form/FieldList on the php side `$fields->push(HTMLEditorField::create('Content', 'My Content'));`

It is hooked in by the schemaBuilder like other FormFields by defining a `getSchemaDataDefaults` and the main activating trigger is the `protected $schemaComponent = 'HtmlEditorField';` attribute.

On the react side, there's a tinymceLoader, which checks that tinymce is loaded before `render()` returns a non-null value. There is a check for the tinymce `<script></script>` tag, and adds that to the markup to download the tinymce library from framework before it starts polling.
Then the tinymcePluginsLoader is loaded, and it begins loading plugins in the plugins property (by default it's the TinyMCEInsertMediaPlugin component).
Finally after tinymce and plugins are loaded, it will display the tinymce editor with the plugins loaded.

flamerohr:
I'm going to close this and re-open later :)
The proof of concept is here, I think after this much time has past, I'd take a different approach (not by much though)
I don't like how TinyMCE has its own async methods, makes plugging in much harder.

dhensby:
@flamerohr do you have any plans to revive this?

flamerohr:
Might be for 5.0, in the end I had found this solution has many race conditions and haven't found a suitable solution for yet.

Did you want to give it a try @dhensby ?

dhensby:
Not something I have the time or deep knowledge for at the moment :(

I was just following up on my unread notifications

PR #6522, opened 2017-01-17
Adding some PHPDoc

catharsisjelly:


dhensby:
thanks

dhensby:
FYI - worth noting this class has been completely re-written for v4 :)

catharsisjelly:
I thought that might be the case, I was going to look for it but have not had the time. Thanks for merging it

PR #6528, opened 2017-01-18
Simplify permissions displayed when admin (fixes #3310)

### LINKED ISSUE #3310: Security admin shows confusing permissions for super admin
I believe it would be better if we can have all boxes under the administrator green or at least greyed out when the user is a super user.

@clarkepaul, do you have any opinions / preferences?

This is the view when looking at the Members permissions
screen shot 2014-03-28 at 3 33 25 pm

It' looks better when looking at the permissions tab in the groups

I noticed that the group permissions shows the permissions as greyed out:

screenshot 2014-07-18 11 05 41

Ticket moved from https://github.com/silverstripe/silverstripe-cms/issues/985

zacps:
This changes permissions inherited by being a superuser to a grey x. Fixes #3310.
screenshot from 2017-01-18 14-23-27

### DIFF
--- a/admin/client/dist/styles/bundle.css
+++ b/admin/client/dist/styles/bundle.css
@@ -7639,7 +7639,7 @@ h1{
   content:"z";
 }
 
-.font-icon-cancel-circled:before{
+.font-icon-cancel-circled:before,.font-icon-disable-circled:before{
   content:"Q";
 }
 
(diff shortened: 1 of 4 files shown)


dhensby:
I like this - but may defer to @clarkepaul to merge

clarkepaul:
Looks good, I'm not actively working on CMS stuff at the moment so might be hanging for a bit. Happy for someone else to merge, check if still relevant etc.

dhensby:
Let's get it in :)

PR #6529, opened 2017-01-18
Fix bug: In addOrderBy method, _SortColumn will only keep the last one if there are more than 1 multi-word columns

shlinx:
Fix bug: In addOrderBy method, _SortColumn will only keep the last one if there are more than 1 multi-word columns

### DIFF
--- a/model/queries/SQLSelect.php
+++ b/model/queries/SQLSelect.php
@@ -334,10 +334,12 @@ class SQLSelect extends SQLConditionalExpression {
 
 					// Move the clause to the select fragment, substituting a placeholder column in the sort fragment.
 					$clause = trim($clause);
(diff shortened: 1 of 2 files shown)


shlinx:
Bug example:
```php
        $pages = Page::get()->sort(array(
            '(ID % 2)  = 0' => 'ASC',
            'CASE WHEN LastEdited = Created THEN Created END' => 'ASC'
        ));
        
        var_dump($pages->sql());
```

You can see only the second clause is taking effect.

dhensby:
Please can you add a test with an example that was previously failing?

shlinx:
@dhensby Sorry for the late response. Just got the chance for this. I've just added test function so it will catch the bug.

dhensby:
thanks

Issue #6506, opened 2017-01-15
RFC: Create a UnionList class - An SS_List that contains multiple SS_Lists to iterate over.


SilbinaryWolf:
ASSIGNING TO MYSELF TO WORK ON. THIS IS FOR PERSONAL AND COMMUNITY REFERENCE.

Discussed with @sminnee very briefly on the community Slack chat. 
The use case is:
"I have a complex query to get the first 3 relevant items, and then get the other items.
How can I get all the SQL logic so it comes out as one DataList / query."

Now, you can usually work around this use-case trivially, but things get tricky when you want pagination to come into the picture or you don't have much control over how a module / third-party code.

To begin with, limit(), sort(), and filter() will throw exception or user_error (like DataList) and won't work.  But if it proves to be easy to reason about a grouped limit, sort or filter, we can add that functionality in later.

```php
class UnionList implements SS_List {
function __construct(array $lists) {
  $this->lists = $list;
}

function getIterator() {
  foreach($this->lists as $list) {
    foreach($list as $item) {
      yield $item;
    }
  }
}
}
```

tractorcow:
This is a pretty good feature, +1 for this. :)

sminnee:
> To begin with, limit(), sort(), and filter() will throw exception or user_error (like DataList) and won't work. But if it proves to be easy to reason about a grouped limit, sort or filter, we can add that functionality in later.

Rather than throwing an exception can simply not implement the `Sortable`, `Filterable`, and `Limitable` interfaces and just implement the core `SS_List` interface instead.

tractorcow:
Yes that's better. :)

You could even optimise it so that lists of similar types (e.g. ArrayUnionList, DataUnionList) have the necessary logic to internally implement those extra methods.

SilbinaryWolf:
Added the base list here:
https://github.com/silverstripe/silverstripe-framework/pull/6540

PR #6540, opened 2017-01-19
ENHANCEMENT Add UnionList, a SS_List type that iterators over multiple SS_Lists

silbinarywolf:
ENHANCEMENT Add UnionList, a SS_List type that iterators over multiple SS_Lists.

See https://github.com/silverstripe/silverstripe-framework/issues/6506

Run Test command:
```
vendor/bin/phpunit.bat --filter UnionListTest framework/tests
```


robbieaverill:
Just a quick comment - this PR would have a lot more confidence if it had documentation and unit tests

SilbinaryWolf:
Yep, as i said on the issue, this is just a "first pass".
Unit tests aren't something I'm super experienced in so it's something that'll require me to look into over the weekend.

SilbinaryWolf:
Added better documentation and unit tests now.

**Todo:**
- Test first()
- Test last()
- Test exists()
- Test iterating
- Test toArray()
- Test toNestedArray()

Am I missing anything?

tests/php/ORM/UnionListTest.php
```diff
@@ -0,0 +1,78 @@
...
+        $obj2->Name = 'test obj 2';
+        $obj2->write();
+        $this->assertTrue($obj2->isInDB());
+       
+        $list1 = ValidatedObject::get()->filter(array('Name' => 'test obj 1'));
+        $list2 = ValidatedObject::get()->filter(array('Name' => 'test obj 2'));
+        $list3 = ValidatedObject::get();
+        $unionList = UnionList::create(array($list1, $list2, $list3));
+        return $unionList;
+    }
```
robbieaverill: This would probably be "best" done in the `setUp` method, e.g.:

```php
protected $unionList;
public function setUp()
{
    parent::setUp();
    $this->unionList = new ValidatedObject();
    // ...
}
```

Then use `$this->unionList` as your test subject

tests/php/ORM/UnionListTest.php
```diff
@@ -0,0 +1,78 @@
...
+use SilverStripe\ORM\Filterable;
+use SilverStripe\Dev\SapphireTest;
+use SilverStripe\ORM\Tests\DataObjectTest\EquipmentCompany;
+use SilverStripe\ORM\Tests\DataObjectTest\Fan;
+use SilverStripe\ORM\Tests\DataObjectTest\Player;
+use SilverStripe\ORM\Tests\DataObjectTest\Sortable;
+use SilverStripe\ORM\Tests\DataObjectTest\SubTeam;
+use SilverStripe\ORM\Tests\DataObjectTest\Team;
+use SilverStripe\ORM\Tests\DataObjectTest\TeamComment;
+use SilverStripe\ORM\Tests\DataObjectTest\ValidatedObject;
```
robbieaverill: There's probably a few of these that aren't used in this test class

tests/php/ORM/UnionListTest.php
```diff
@@ -0,0 +1,78 @@
...
+
+class UnionListTest extends SapphireTest
+{
+
+    // Borrow the model from DataObjectTest
```
robbieaverill:
```php
/**
 * Borrow the model from DataObjectTest
 * {@inheritDoc}
 */
```

SilbinaryWolf: This one was literally a copy-paste from DataListTest :)
But I'll fix.

tests/php/ORM/UnionListTest.php
SilbinaryWolf: Sweet. I thought there was a better place to handle this shared logic :)

tests/php/ORM/UnionListTest.php
SilbinaryWolf: Yup, was going to remove them at the end. IDE doesn't automatically add "use". Will need to look into that.

dhensby:
Minor code style issue

SilbinaryWolf:
Hm?

dhensby:
check out the build failures: https://travis-ci.org/silverstripe/silverstripe-framework/builds/195039074

SilbinaryWolf:
Oh right. I'm aware it's failing, was going to look into it in the future (now)

SilbinaryWolf:
All I see is:
```
Script phpcs --standard=tests/phpcs/ruleset.xml src/ admin/code/ tests/ admin/tests/ handling the lint event returned with error code 1
The command "if [ "$PHPCS_TEST" = "1" ]; then composer run-script lint; fi" exited with 1.
```

I'm assuming PHP 5.6 is busted with Travis CI?
In any case, I'm done with this unless you guys have further feedback!

robbieaverill:
Hey @SilbinaryWolf - that's a codesniffer violation that's failing the build:

```
FILE: ...verstripe/silverstripe-framework/tests/php/ORM/UnionListTest.php
 159 | ERROR | [x] Expected 1 newline at end of file; 0 found
```

SilbinaryWolf:
Ah right you are. I assumed it would be highlighted in red like the other bits!

tests/php/ORM/UnionListTest.php
```diff
@@ -0,0 +1,159 @@
+<?php
+
+namespace SilverStripe\ORM\Tests;
+
+use SilverStripe\Dev\Debug;
```
dhensby: This import is not required

src/ORM/UnionList.php
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+     /**
+     * Returns true if one of the SS_List's have an item.
+     *
+     * @return bool
+     */
+    public function exists()
+    {
+        return $this->count() > 0;
```
dhensby: There may be a more performant way of determining if the list exists without counting every item just to see if it's `>0`.

src/ORM/UnionList.php
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    /**
+     * Iterate over each SS_List, one after the other.
+     *
+     * @return ArrayIterator
+     */
+    public function getIterator()
+    {
+        foreach ($this->lists as $i => $list) {
```
dhensby: `$i` is not used

src/ORM/UnionList.php
```diff
@@ -0,0 +1,226 @@
...
+
+    /**
+     * Get array of each record in each SS_List.
+     *
+     * @return array
+     */
+    public function toArray()
+    {
+        $result = array();
+        foreach ($this->lists as $i => $list) {
```
dhensby: `$i` is not used

src/ORM/UnionList.php
```diff
@@ -0,0 +1,226 @@
...
+
+    /**
+     * Return this list as an array and every object it as an sub array as well
+     *
+     * @return array
+     */
+    public function toNestedArray()
+    {
+        $result = array();
+        foreach ($this->lists as $i => $list) {
```
dhensby: `$i` is not used

src/ORM/UnionList.php
```diff
@@ -0,0 +1,226 @@
...
+            foreach ($list as $item) {
+                if (is_object($item)) {
+                    if (method_exists($item, 'toMap')) {
+                        $result[] = $item->toMap();
```
dhensby: does `toMap` return an `array` or an `SS_Map` object?

src/ORM/UnionList.php
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    /**
+     * Get first item in the first SS_List.
+     *
+     * @return array
+     */
+    public function first()
+    {
+        $list = isset($this->lists[0]) ? $this->lists[0] : null;
```
dhensby: This may be more robust:

```php
$list = reset($this->lists);
return $list ? $list->first() : null;
```

src/ORM/UnionList.php
```diff
@@ -0,0 +1,226 @@
...
+     /**
+     * Returns an array of a single field value for each item in each list.
+     *
+     * @param string $colName
+     * @return array
+     */
+    public function column($colName = 'ID')
+    {
+        $result = array();
+        foreach ($this->lists as $i => $list) {
```
dhensby: `$i` is not used

src/ORM/UnionList.php
```diff
@@ -0,0 +1,226 @@
...
+            $result = array_merge($result, $list->column($colName));
+        }
+        return $result;
+    }
+
+    public function map($keyfield = 'ID', $titlefield = 'Title')
+    {
+        throw new LogicException(
```
dhensby: Why is `map` not allowed?

src/ORM/UnionList.php
```diff
@@ -0,0 +1,226 @@
...
+            "UnionList::".__FUNCTION__."() is not allowed."
+        );
+    }
+
+    public function find($key, $value)
+    {
+        throw new LogicException(
```
dhensby: Why is `find` not allowed?

tests/php/ORM/UnionListTest.php
```diff
@@ -0,0 +1,159 @@
...
+use SilverStripe\Dev\SapphireTest;
+use SilverStripe\ORM\Tests\DataObjectTest\ValidatedObject;
+
+class UnionListTest extends SapphireTest
+{
+    /**
+     * Borrow the model from DataObjectTest
+     * {@inheritDoc}
+     */
+    protected static $fixture_file = 'DataObjectTest.yml';
```
dhensby: Your using this fixture file but never actually use any of the fixtures.

tests/php/ORM/UnionListTest.php
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * @var UnionList
+     */
+    protected $unionList = null;
+
+    protected function getExtraDataObjects()
+    {
+        return array_merge(
+            DataObjectTest::$extra_data_objects,
```
dhensby: Do you need all these extra data objects? You don't appear to use them.

tests/php/ORM/UnionListTest.php
```diff
@@ -0,0 +1,159 @@
...
+            ManyManyListTest::$extra_data_objects
+        );
+    }
+
+    public function setUp()
+    {
+        parent::setUp();
+        // create an object to test with
+        $obj1 = new ValidatedObject();
```
dhensby: Rather than writing all of these in the test, you should create a fixture file

tests/php/ORM/UnionListTest.php
```diff
@@ -0,0 +1,159 @@
...
+        $this->assertEquals('test obj 1', $recordSet[0]->Name);
+        $this->assertEquals('test obj 2', $recordSet[1]->Name);
+        $this->assertEquals('test obj 3', $recordSet[2]->Name);
+        $this->assertEquals('test obj 4', $recordSet[3]->Name);
+    }
+
+    public function testExists()
+    {
+        $unionList = $this->unionList;
+        $this->assertEquals(true, $unionList->exists());
```
dhensby: better to use `assertTrue`

tests/php/ORM/UnionListTest.php
```diff
@@ -0,0 +1,159 @@
...
+        $unionList = UnionList::create(array(ValidatedObject::get()->filter(array('Name' => 'non existant'))));
+        $this->assertEquals(false, $unionList->exists());
```
dhensby: Better to use `assertFalse`

tests/php/ORM/UnionListTest.php
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    public function testToNestedArray()
+    {
+        $unionList = $this->unionList;
+        $recordSet = $unionList->toNestedArray();
+        $expected = [
```
dhensby: I think you could make this test a bit more elegant, especially once you start using fixtures.

src/ORM/UnionList.php
SilbinaryWolf: Because if you you have two DataLists, one of say "Team" DataObject and one of "SiteTree" DataObject, the IDs aren't going to merge cleanly. Open to suggestions though on how this might work.

src/ORM/UnionList.php
SilbinaryWolf: Because if you want to ::find('ID', 1) and you're iterating over two different DataList types, it'll not work as you expect. IMO, It's something that can be looked at if people want to use it.

src/ORM/UnionList.php
SilbinaryWolf: No idea. This is essentially a copy-paste of how ArrayList works.

src/ORM/UnionList.php
SilbinaryWolf: It could be faster *sometimes* if it iterated over each list seperately and early exited. I'd prefer performance to be predictable / show the worst possible case though when profiling.

Unless you have a better optimization idea?

src/ORM/UnionList.php
SilbinaryWolf: My version saves a function call and is more performant. However for consistency and readability's sake, I'll do the reset()

tests/php/ORM/UnionListTest.php
SilbinaryWolf: Getting rid of this and 'getExtraDataObjects()' caused by tests to fallover in an odd way:
"42S02-1146: Table 'ss_tmpdb2509246.dataobjecttest_validatedobject' doesn't exist"

Might be a Windows thing or slight configuration issue. Anyway, I've pushed up some of the latest changes. If you spot what I'm doing wrong, that'd be awesome!

robbieaverill: If you remove the fixture but still create stub objects you need to add `protected $usesDatabase = true;` - otherwise it won't create a test DB/table

src/ORM/UnionList.php
dhensby: OK - maybe an enhancement for later - but if a user selects a non `ID` field, this function would work fine. I think if you use `ID` then we can say it's expected behaviour for duplicate IDs to be overridden.

src/ORM/UnionList.php
dhensby: As above, we can enhance later. But my view is just because it doesn't work reliably for one field doesn't mean it shouldn't work for any.

src/ORM/UnionList.php
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    /**
+     * Get first item in the first SS_List.
+     *
+     * @return array
+     */
+    public function first()
+    {
+        $list = resets($this->lists);
```
dhensby: [`reset`](http://php.net/manual/en/function.reset.php) not `resets`

tests/php/ORM/UnionListTest.php
```diff
@@ -0,0 +1,152 @@
...
+     * @var UnionList
+     */
+    protected $unionList = null;
+
+    public function setUp()
+    {
+        parent::setUp();
+        
+        // create an object to test with
+        /*$obj1 = new ValidatedObject();
```
dhensby: Please don't keep commented code

tests/php/ORM/UnionListTest.php
```diff
@@ -0,0 +1,152 @@
...
+        $obj3 = new ValidatedObject();
+        $obj3->Name = 'test obj 3';
+        $obj3->write();
+        $this->assertTrue($obj3->isInDB());
+
+        $obj4 = new ValidatedObject();
+        $obj4->Name = 'test obj 4';
+        $obj4->write();
+        $this->assertTrue($obj4->isInDB());*/
+        $obj3 = ValidatedObject::get()->find('Name', 'test obj 3');
```
dhensby: you can use `$this->objFromFixtureValidatedObject::class, 'numeric1')` - this makes the tests a little more versatile.

tests/php/ORM/UnionListTest.php
SilbinaryWolf: I'm not.

src/ORM/UnionList.php
SilbinaryWolf: I know.

Just an FYI, I'm generally working on unit tests before I go out, and the easiest way to ensure I can access it from a different house/computer is just pushing it up. Didn't need feedback here :)

tests/php/ORM/UnionListTest.php
SilbinaryWolf: I'm not sure fixtures are testing what I want them to test. I assumed they were just DataObjects written to the DB automatically but that doesn't seem to be the case? Are they just DataObjects created via a YML configuration (without a ->write() call)?

src/ORM/UnionList.php
SilbinaryWolf: Thinking now. We should just wait to see how modules/user-code uses this in the wild, then optimize for the majority case. If people use ArrayList with it a lot, a nice optimization is just checking the $lists for an ArrayList->count() > 0 to avoid a DB call.

dhensby: I like your ideas of iterating the lists as breaking early. There's no need to force worst case on people.

tests/php/ORM/UnionListTest.php
dhensby: That is what they are. The tests weren't failing because of fixture failures, I don't think...

SilbinaryWolf: It was due to me using getExtraDataObjects() wrong. Replaced all that with fixtures now.

SilbinaryWolf:
Is all that's left making ->exists() more efficient?
I'll see if I can make time this week.

SilbinaryWolf:
I'll look at codecov.io soon. I'll probably see if I can implement map() and filter() functions.

src/ORM/UnionList.php
```diff
@@ -0,0 +1,238 @@
...
+     * @return bool
+     */
+    public function exists()
+    {
+        foreach ($this->lists as $list) {
+            if ($list instanceof ArrayList && $list->exists()) {
+                return true;
+            }
+        }
+        foreach ($this->lists as $list) {
```
dhensby: The second loop here isn't needed - this can be an esleif inside the same loop.

It appears there's nothing to force to be an object - that should probably be caught somewhere.

Also, if the `$list` is not an `instanceof ArrayList` then there's no guarantee that `->exists()` will be a defined method.

SilbinaryWolf: Yep you're right. Though I'd just do an "if" underneath, that's more performant.

Not sure what you mean by "nothing to force to be an object"?

Ah. Good catch. Why isn't that in the SS_List interface?

SilbinaryWolf:
Note to self: I want to also test paginated list behaviour to ensure it works with UnionList as expected. I'l write a unit test if necessary.

SilbinaryWolf:
I'd like to test this exists() function to ensure if an ArrayList() is at the end of the UnionList, it'll still not hit the DB at all. How can I check for a DB-hit?

SilbinaryWolf:
This is mostly ready to go, other than me wanting to check exists() for the DB-hit. But not sure the best way to do that so... unless you guys have suggestions, PR this in please.

dhensby:
I think the DB hit is acceptable.

I think that the `exists` method only needs the second conditional it would work on both `ArrayList` and other instances...

dhensby:
Oh - and can you please squash the committs - or we can do it on merge...

SilbinaryWolf:
Just squash on merge if you could.

src/ORM/UnionList.php
```diff
@@ -0,0 +1,261 @@
...
+     * @return bool
+     */
+    public function exists()
+    {
+        foreach ($this->lists as $list) {
+            if ($list instanceof ArrayList) {
+                if ($list->exists()) {
+                    return true;
+                }
+            } elseif (method_exists($list, 'exists') && $list->exists()) {
```
dhensby: This condition can be used alone, we don't need to check if the list is an `instanceof ArrayList` above because this check will handle that situation

dhensby:
I'm happy with this from a code point of view but have two concerns:

1. Code coverage has dropped by 4% - we should try to maintain the previous coverage (ie: you need 100% coverage for your new code)
2. Throwing exceptions on methods that are valid on a class/interface feels very bad to me. If you do `instanceof SS_List` (or any other class/interface) you should be able to assume that's a safe call to make and that you won't get logic exceptions thrown at you.

SilbinaryWolf:
1) How am I meant to cover exception throwing code? I could refactor to merely throw user_error() like DataList does on offsetSet()?
2) Is throwing a user_error() better? I understand your concern, but I'd rather leave these functions undefined for now, rather than implementing a half-baked idea or something that'll cause user-code pain.

Keep in mind, DataList::add() is simply blank (not sure that is good either) and has unimplemented behaviour in offsetSet/offsetUnset.

robbieaverill:
>How am I meant to cover exception throwing code?

You can assert that exceptions are thrown (however it sounds like @dhensby is suggesting you remove the exceptions anyway) - https://phpunit.de/manual/5.7/en/appendixes.annotations.html#appendixes.annotations.expectedException

src/ORM/UnionList.php
```diff
@@ -0,0 +1,257 @@
...
+
+    /**
+     * Get first item in the first SS_List.
+     *
+     * @return array
+     */
+    public function first()
+    {
+        $list = reset($this->lists);
+        return $list ? $list->first() : null;
```
micmania1: What if the first list didn't have any items but the second list did? Then this function would return null, but that wouldn't be correct. It should probably loop through all of the lists and return the first result in any list.

src/ORM/UnionList.php
```diff
@@ -0,0 +1,257 @@
...
+
+    /**
+     * Get last item in the last SS_List.
+     *
+     * @return array
+     */
+    public function last()
+    {
+        $list = end($this->lists);
+        return $list ? $list->last() : null;
```
micmania1: Same as the first comment

SilbinaryWolf:
Considering I missed this pretty big issue with UnionList and that the idea hasn't really been tested much in real world scenarios.

I'm thinking perhaps I move this PR and all the work to my own module for the time being before it goes into core.

Agree? Disagree?

dhensby:
Perhaps - this feature seems suitable to a module as it's quite niche and, perhaps, not particularly performant so may be worth discouraging the use of...

Though, we have come a long way... it just needs some solid test cases

dhensby:
this may want to target `4` branch now.

SilbinaryWolf:
@dhensby @sminnee Im not going to put any more effort towards this as this was a personal time piece of work and I have other things to prioritize.

So options are:
- If someone else wants to finish this off, they can.
- Close off PR.

robbieaverill:
I'm going to close this for now - feel free to reopen if someone wants to revive this

robbieaverill:
Ah, great! Thanks

Issue #6549, opened 2017-01-22
Controllers for SiteTree instances should be configurable


robbieaverill:
At the current state of 4.0.0-alpha4 the controller for a SiteTree model must live in the same namespace as the model itself. This leads to controllers that are namespaced like `Batman\Batcave\Model\MyPageController`, where it might be more relevant to put it into the `Batman\Batcave\Controllers` namespace.

Via Slack it was briefly discussed, and an option is to add a configuration static to SiteTree which can define the controller name.

Options:

* The property (e.g. `$frontend_controller`) would be null by default
* You can define it for your own SiteTree classes
* If not defined, the SiteTree `__construct` would define it as the current behaviour, which is the FQ class name with `Controller` on the end (and `_Controller`) for deprecated support
* The logic from `getControllerName` can be moved to `SiteTree::__construct` in this case

Considerations:

* If the controller name is defined for a SiteTree instance, and it can't be found, we should throw an exception with a useful error message. Something like "Check your model's `$frontend_controller` property - it was defined but could not be found!"
* If we use a configuration static for this property it gives the benefit of being able to configure it in YAML, however;
* If we use a configuration static for this property, it opens the way for a developer to entirely replace a controller from a different module with their own. I personally like this way of extending functionality, but it's not something that exists in SilverStripe at this stage outside of the Injector - so do we want to introduce it for the sake of this?
* If not, we should use a protected property instead - this way you'd need to use the Injector to achieve the above behaviour (as you would currently anyway)

sminnee:
I generally think this is a good idea but I think that extra logic in `SiteTree::__construct()` is a really bad idea: it's called literally hundreds of times on every pageview.

Instead, I think that a new method, `SiteTree::getFrontendController()` should implement both the config lookup and the legacy functionality, and return a controller instance.

robbieaverill:
And how about the considerations around a member property vs a config static?

sminnee:
use a config static i think, and pass its value to `Injector::inst()->get($value)` to get the controller, rather than using `new`

robbieaverill:
Sorry, this should have been logged on the CMS module.

dhensby:
From a purist's point of view shouldn't `Controller`s nominate the `Model` they handle and not the other way round? `Controller`s act on `Model`s and so need knowledge about the `Model` but the `Model` doesn't act on a control and should be agnostic to what `Controller`(s) can "handle" it?

[...]

PR #6552, opened 2017-01-22
Use customizable controller name method in Security::getResponseController

robbieaverill:
This is a follow up PR for consistency, from silverstripe/silverstripe-cms#1722 to ensure we use a consistent public API throughout.

The builds shouldn't pass until that PR is merged.

Part of #6549


robbieaverill:
Closing pending further discussion on #6549

Issue #6493, opened 2017-01-12
Aggregate filters in DataList::filter


sminnee:
For example, do get members who are in more than 3 groups:

```php
$socialites = Member::get()-filter(['Groups.Count():GreaterThan' => 3])
```
The core would be that the column name would end in a parenthetic expression, which would indicate that it's not a regular column:

 * `count()`
 * `min(Column)`
 * `max(Column)`
 * `avg(Column)`
 * `sum(Column)`

## Implementation notes

For a method like `Things.Count:GreaterThan`, “Count” will wind up getting passed to `SearchFilter::getDbName()`. The existing logic for joining to things will already be used.

So you could put something in there to parse `Count()`, `Max(Column)`, `Min(Column)` etc. A simple check of whether it contained `(` should be enough to distinguish it from a regular column, and then instead of `"$table"."$column"` you could return `COUNT("$table".*)`.

The final thing you’d have to do is attach these queries to `having()` rather than `where()`.

This is done in each SearchFilter subclass so you’d need to have some flag passed back that said “does this filter need to be applied in WHERE or HAVING?" and then make use of that flag in applyOne, applyMany, excludeOne, excludeMany

sminnee:
@tractorcow oh ye of the ORM, see any issues with this? I think @unclecheese might give it a go.

sminnee:
Great to see this implemented :-)

@unclecheese do we have docs on this? didn't notice them in the linked PR...

PR #6553, opened 2017-01-22
Feature/aggregate data filters

unclecheese:
Per https://github.com/silverstripe/silverstripe-framework/issues/6493

### DIFF
--- a/src/ORM/Filters/SearchFilter.php
+++ b/src/ORM/Filters/SearchFilter.php
@@ -58,6 +58,8 @@ abstract class SearchFilter extends Object
      */
     protected $relation;
 
+    protected $aggregate;
+
     /**
      * @param string $fullName Determines the name of the field, as well as the searched database
      *  column. Can contain a relation name in dot notation, which will automatically join
@@ -73,6 +75,7 @@ abstract class SearchFilter extends Object
 
         // sets $this->name and $this->relation
         $this->addRelation($fullName);
+        $this->addAggregate($fullName);
         $this->value = $value;
         $this->setModifiers($modifiers);
     }
@@ -92,6 +95,23 @@ abstract class SearchFilter extends Object
         } else {
             $this->name = $name;
         }
+    }
+
+
+    protected function addAggregate($name)
+    {
+        if(!$this->relation) {
+        	return;
+        }
+
+        if(!preg_match('/([A-Za-z]+)\((?:([A-Za-z_*][A-Za-z0-9_]*))?\)$/', $name, $matches)) {
+        	return;
+        }
+
+        $this->aggregate = [
+        	'function' => strtoupper($matches[1]),
+        	'column' => isset($matches[2]) ? $matches[2] : null
+        ];
     }
 
     /**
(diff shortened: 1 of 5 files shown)


dhensby:
Postgres issues. Also don't hardcode IDs into tests because postgres won't like that, either.

Linting issues too.

tests/php/ORM/DataListTest.php
```diff
@@ -1220,6 +1222,107 @@ public function testFilterByNullCase()
...
+    	}
+    }
+
+    public function testAggregateFilterExceptions()
+    {
+		$ex = null;
+		try {
+			$filter = new ExactMatchFilter('Comments.Max( This will not parse! )');
+		} catch(\Exception $e) {
+			$ex = $e;
```
dhensby: It looks like this assignment is unnecessary as the resulting var is never used.

tests/php/ORM/DataListTest.php
```diff
@@ -1220,6 +1222,107 @@ public function testFilterByNullCase()
...
+		$this->assertInstanceOf(\InvalidArgumentException::class, $e);
+		$this->assertRegExp('/Invalid column/', $e->getMessage());
+    }
+
+    public function testAggregateFilters()
+    {    	
+    	$teams = Team::get()->filter('Comments.Count()', 2);
+    	$team1 = $this->objFromFixture(Team::class, 'team1');
+
+    	$this->assertEquals(1, $teams->count());
```
dhensby: `$this->assertCount(1, $teams)`

tests/php/ORM/DataListTest.php
```diff
@@ -1220,6 +1222,107 @@ public function testFilterByNullCase()
...
+    	$this->assertEquals($team1->ID, $teams->first()->ID);
+
+    	$teams = Team::get()->filter('Comments.Count()', [1,2]);
+    	
+    	$this->assertEquals(2, $teams->count());
+    	$this->assertEquals([], array_diff([1,2], $teams->column('ID')));
```
dhensby: Don't hard code the IDs as postgres doesn't reset the auto_increment with each test.

My approach tends to be:

```php
$expectedTeam1 = $this->objFromFixture(Team::class, 'team1');
$expectedTeam2 = $this->objFromFixture(Team::class, 'team2');

$this->assertCount(2, $teams);
$this->assertContains($expectedTeam1->ID, $teams->column('ID'));
$this->assertContains($expectedTeam2->ID, $teams->column('ID'));
```

unclecheese: Thanks, Dan. I thought about that, but it gets awfully long-winded if you have more than a few records in the list. You need an assertion for every ID. Part of the fun, I guess. Thanks for the tip on Postgres. Did not know that.

unclecheese:
Have revised per your review, @dhensby

tests/php/ORM/DataListTest.php
unclecheese: I've updated this to use the `$ex` var instead.

dhensby:
Still lots of code style errors: https://travis-ci.org/silverstripe/silverstripe-framework/jobs/194647196#L641-L1016

Looks like the `GROUP BY` clause is not correctly quoting the fields (they aren't wrapped in `"`)

src/ORM/Filters/SearchFilter.php
```diff
@@ -244,13 +310,34 @@ public function getDbFormattedValue()
...
+     * @return DataQuery
+     */
+    public function applyAggregate(DataQuery $query, $having)
+    {
+    	$schema = DataObject::getSchema();
+    	$baseTable = $schema->baseDataTable($query->dataClass());
+    	
+    	return $query
+        	->having($having)
+        	->groupby("{$baseTable}.ID");
```
dhensby: This line needs to be: `->groupby("\"{$baseTable}\".\"ID\"");`

dhensby:
almost there!

unclecheese:
All sorted, Dan. Cleaned up the code formatting and added those quotes.

dhensby:
Great, waiting for green :)

dhensby:
got a test failure :(

dhensby:
It's postgres - https://travis-ci.org/silverstripe/silverstripe-framework/jobs/195467893#L595

dhensby:
A few CS issues too

unclecheese:
OK, it finally passed. :)

tractorcow:
I know this is a bit late @unclecheese but can we have some user documentation for this? I tried to look it up and I can't find any anywhere.

unclecheese:
Docs are here https://github.com/silverstripe/silverstripe-framework/blob/4/docs/en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md#filtering-by-aggregates

Issue #6221, opened 2016-10-25
RFC Replace or upgrade Zend_Translate


chillu:
Split out from [RFC Replace Zend_Locale with new i18n backend](https://github.com/silverstripe/silverstripe-framework/issues/6194)
# Introduction

SilverStripe uses Zend Framework v1 for message translations (`i18n::_t()`) through `Zend_Translate`. This dependency is out of date, Zend has declared [EOL](https://framework.zend.com/blog/2016-06-28-zf1-eol.html) in September 2016. Our goal is to either upgrade or replace the library.

We've written a custom [YAML adapter](https://github.com/silverstripe/silverstripe-framework/blob/master/thirdparty/zend_translate_railsyaml/library/Translate/Adapter/RailsYAML.php), since `Zend_Translate` doesn't support the format natively. The `Zend_Translate` API is used both indirectly in `i18n::_t()`, and directly through `i18n:: register_translator(<Zend_Translate>)`.
# Solution Requirements

Note: Some of these requirements might be fulfilled by the existing `i18n` class.
- Include translations by module (`i18n:: include_by_locale()`)
- Translation fallbacks from full locale to lang (e.g. `de_AT` => `de`)
- Translation fallbacks to `i18n::$default_locale`
- Caching of parsed translation files (e.g. YAML)
- Lookup performance: `i18n::_t()` is called dozens of times on each CMS render
- Default message fallback defined on `_t()` call
- Sprintf-style string interpolation (e.g. `_t('There are %s monkeys', [5])`)
- Translator cascade for multiple format support (not really used in practice, was an effort to smoothly upgrade from PHP to YAML based translations)
- Format compatibility with [Transifex translation service](http://docs.transifex.com/formats/)
- Message "groups" in translation files through dot notation (e.g. `LeftAndMain.SAVE`)
- Rails YAML format support (or rewriting our existing [RailsYaml.php adapter](https://github.com/silverstripe/silverstripe-framework/blob/master/thirdparty/zend_translate_railsyaml/library/Translate/Adapter/RailsYAML.php)). If we switch to a different format, we also need to fix [text collection](https://github.com/silverstripe/silverstripe-framework/blob/master/i18n/i18nTextCollector.php)
# Options
## 1. Symfony Translate

Currently at version 3.1. Seems to have native support for string placeholders. Initial impression is that the symfony localisation is slightly more polished and feature-rich than zend. Documentation http://symfony.com/doc/current/translation.html

[...]

Issue #6194, opened 2016-10-17
RFC Replace Zend_Locale with new i18n backend


tractorcow:
See https://github.com/silverstripe/silverstripe-framework/issues/5359
dev list post: https://groups.google.com/forum/#!topic/silverstripe-dev/BVgFrxb7ISw

# Introduction

Currently SilverStripe framework relies on Zend framework v1 for several core localisation features. This includes the below Zend components:
- Zend_Locale
  - Text direction
  - Geographic information
  - Numeric formatting
  - Date formatting
- Zend_Currency
  - Currency formatting
- Zend_Date
  - Date formatting

The libraries these features are built on top of are several years out of date and require upgrading.
# Exclusions

Note that the following other zend components remain a dependency of other parts of the CMS and are not covered by this RFC, even if these may be affected by the outcome of the decided direction.
- ~~Zend_Translate~~ Moved to https://github.com/silverstripe/silverstripe-framework/issues/6221
- Zend_Cache
- Zend_Loader (only exists as a dependency of the prior Zend components)
# Solution Requirements:

The chosen solution will need to either provide the below functionality, or have added on top of the core library to meet the necessary functionality:
- Support for mi_NZ locale
- Date formatting
- Numeric formatting
- Currency formatting
# PHP INTL Dependency

Many modern i18n Solutions require the php intl library. Currently SilverStripe does not, although evading this as a requirement in future releases will limit our options.

Symfony currently provides a polyfill for intl https://github.com/symfony/polyfill.

Our options are:
1. Choose a solution that does rely on php-intl and require it as a core dependency
2. Choose a solution that does rely on php-intl, but rely on a polyfill
3. Choose a solution that does not rely on php-intl
4. Require php-intl to be installed to support i18n, but allow i18n to be disabled.

Investigation of symfony's polyfill implementation seems extremely incomplete, or implemented at a very shallow level only. E.g. stubs containing `MethodNotImplementedException` and hard-coded support for `en` locale only. Supporting or offering an incomplete solution in my mind is worse than no solution at all.

I recommend option 4, as it allows us to build a full i18n backend that will operate with all necessary features, but provides room for the community to develop a custom backend (or backends) to support environments without php-intl.
# Example i18n Options:

Given that there are a huge number of possible i18n solution libraries, below is a non-exhaustive list of examples of solutions that could be considered.
## 1. Write i18n wrapper over php-intl

[...]

PR #6558, opened 2017-01-24
API Replace i18n message localisation with symfony/translation

tractorcow:
Closes https://github.com/silverstripe/silverstripe-framework/issues/6221

Requires:

 - [x] https://github.com/silverstripe/silverstripe-cms/pull/1724 (can be merged without framework changes)
 - [x] https://github.com/silverstripe/silverstripe-reports/pull/58

This feature includes advanced pluralisation based on http://symfony.com/doc/current/components/translation/usage.html#component-translation-pluralization

An important upgrading note for this change is that any localisation calls to _t() without a default will now trigger a warning.

All existing Zend_Translate libraries and dependencies have been removed.

This does not replace Zend_Locale, however, which will be addressed later with https://github.com/silverstripe/silverstripe-framework/issues/6194

There are some potential performance issues with this implementation: I will post benchmarking analysis later.


tractorcow:
I also have some preliminary feedback from @sminnee which I need to implement:

 
 - i18nTextCollector should collect strings, even if they do not have a default defined, to ensure they show up as editable within transifex.
 - usage of pluralised strings with i18n::_t (rather than i18n::pluralise) should be a warning.

tractorcow:
Phew, travis-ci isn't slower, as I had feared. :) Still about ~12 minutes per test run.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -212,6 +214,104 @@ instead, or if used in an actual XML file use `.CDATA` (see [template casting](/
 
 Where your code once used SQLQuery you should now use SQLSelect in all cases, as this has been removed (check the [3.2.0](3.2.0) upgrading notes).
 
+#### Upgrade code that uses i18n
+
+In many cases, localisation strings which worked in 3.x will continue to work in 4.0, however certain patterns
+have been deprecated and will be removed in 5.0. These include:
+
+ - _t calls with a $context parameter, which is ignored.
```
chillu: Hmm, this was a useful for translators, since text collector picks it up - we just stopped using it because transifex didn't support it at the time. Looks like support for comments has been added now: https://docs.transifex.com/formats/yaml. I originally considered this a regression when moving from the old getlocalization.com service over to transifex.

Context is fairly essential for good translation quality - for example, an identifier like `CMSMain.HANDLE="handle"` doesn't make it clear to a translator if it's used a verb or noun. The fact that we haven't been including comments is unlikely to generate a lot of noise with translators, but that's mainly because we don't *talk* to translators enough (rather than them not valuing the feature).

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -212,6 +214,104 @@ instead, or if used in an actual XML file use `.CDATA` (see [template casting](/
...
+ - _t calls with sprintf-style placeholders (`%s`). Replace with named placeholders instead.
+ - _t calls with non-associative injection arguments. Please use an associative array for all arguments.
+ - _t calls which do not include a default value.
```
chillu: The intention here was to avoid duplication - you can use the same `_t()` call multiple times in your code, without duplicating the default value and having them diverge over time. But relying on one "master" `_t()` call is just as prone to human error, since you might remove from the code later on, only leaving the `_t()` call without the default value. In this case, text collection might set the value to null.

I think it's a per-project decision though: Some translation workflows will rely on developers maintaining YAML/JSON files directly, rather than relying on text collection - so you shouldn't force them to duplicate their efforts by setting default values in `_t()` calls.

docs/en/02_Developer_Guides/13_i18n/index.md
```diff
@@ -166,12 +166,61 @@ All strings passed through the `_t()` function will be collected in a separate l
...
+
+## Pluralisation
+
+i18n also supports locale-respective pluralisation rules. Many languages have more than two plural forms,
+unlike English which has two only; One for the singular, and another for any other number.
+
+More information on what forms these plurals can take for various locales can be found on the
+[CLDR documentation](http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html)
+
+The ability to pluralise strings is provided through the `i18n::pluaralise` method, which is similar to the
```
chillu: Spelling

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1179,15 +1279,34 @@ handle field-level and form-level messages. This has the following properties:
...
-### <a name="overview-mailer"></a>Email and Mailer
+### <a name="overview-i18n"></a>i18n API
 
-* `Mailer` converted to an interface
-* `SwfitMailer` added as new default mailer
-* `Email` re-written to be powered by [SwiftMailer](https://github.com/swiftmailer/swiftmailer)
+#### <a name="overview-i18n-api"></a>i18n API Additions / Changes
+
+* Upgrade of i18n to symfony/translation
+* Localisation based on language-only (without any specific locale) is now supported
```
chillu: Wasn't that supported already? For example, we've got `es.yml` for generic Spanish, and `es_MX.yml` for Mexican Spanish.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1179,15 +1279,34 @@ handle field-level and form-level messages. This has the following properties:
...
+* i18nEntityProvider::provideI18nEntities() Now is expected to return only a single array
+  map of key to default values.
+* i18n keys for '.PLURAL_NAME' and '.SINGULAR_NAME' have been changed back to FQN class names
```
chillu: "FQN" isn't a commonly used term - just say "include their namespace"

docs/en/02_Developer_Guides/05_Extending/05_Injector.md
```diff
@@ -121,6 +121,21 @@ As well as properties, method calls can also be specified:
 	      - [ pushHandler, [ %$DefaultHandler ] ]
 
 
+## Using constants as variables
+
+Any of the core constants can be used as a service argument by quoting with back ticks "`".
```
chillu: Awesome change, that would've come in handy a couple of times already!

lang/en.yml
```diff
@@ -665,45 +667,87 @@ en:
...
     PLURALNAME: 'Permission Role Codes'
+    PLURALS:
+      one: 'A Permission Role Code'
+      other: '{count} Permission Role Codes'
     SINGULARNAME: 'Permission Role Code'
   SilverStripe\Security\RememberLoginHash:
     PLURALNAME: 'Remember Login Hashs'
+    PLURALS:
+      one: 'A Remember Login Hash'
+      other: '{count} Remember Login Hashs'
```
chillu: Hashs -> Hashes

src/Dev/Backtrace.php
```diff
@@ -155,7 +155,7 @@ public static function backtrace($returnVal = false, $ignoreAjax = false, $ignor
      * @param int $argCharLimit
      * @return string
      */
-    public static function full_func_name($item, $showArgs = false, $argCharLimit = 10000)
+    public static function full_func_name($item, $showArgs = false, $argCharLimit = 100)
```
chillu: What does that have to do with i18n?

src/Dev/Tasks/i18nTextCollectorTask.php
```diff
@@ -6,7 +6,7 @@
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Dev\Debug;
 use SilverStripe\Dev\BuildTask;
-use SilverStripe\i18n\i18nTextCollector;
+use SilverStripe\i18n\TextCollection\i18nTextCollector;
```
chillu: What do you think about moving text collection to its own module? As I've outlined in my comments around the _t() default value, I don't think text collection is the *only* translation workflow we should support - some teams will just manage their YAML/JSON separately. The reason I'm bringing this up is namespacing: If it's a separate module, we'd probably use `SilverStripe\i18nTextCollection\i18nTextCollector` instead of `SilverStripe\i18n\TextCollection\i18nTextCollector` (since the `SilverStripe\i18n` namespace should be core only)

src/i18n/Messages/Symfony/FlushInvalidatedResource.php
```diff
@@ -0,0 +1,76 @@
...
+
+use SilverStripe\Core\Flushable;
+use Symfony\Component\Config\Resource\DirectoryResource;
+use Symfony\Component\Config\Resource\SelfCheckingResourceInterface;
+
+/**
+ * Some arbitrary resource which expires when flush is invoked.
+ * Uses a canary file to mark future freshness requests as stale.
+ *
+ * @link https://media.giphy.com/media/fRRD3T37DeY6Y/giphy.gif for use case
```
chillu: hah, nice easter egg

src/i18n/Messages/Symfony/FlushInvalidatedResource.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * Path to i18n canary
+     *
+     * @return string
+     */
+    protected static function canary()
+    {
+        return TEMP_FOLDER . '/catalog.i18n_canary';
```
chillu: The class name and PHPDoc ("Some arbitrary resource") implies this is a generic helper, yet the canary file name is hardcoded to `catalog.i18n_canary`. Shouldn't it take a mandatory identifier as a constructor arg?

docs/en/04_Changelogs/4.0.0.md
sminnee: Perhaps `warn_on_missing_default` is global config option set to true by default?

docs/en/04_Changelogs/4.0.0.md
sminnee: OK - if there's a reason we're not using it that's no longer relevant, then maybe it's worth keeping.

chillu: Sorry for forking the discussion, I've commented a bit more on the original RFC: https://github.com/silverstripe/silverstripe-framework/issues/6221#issuecomment-274708979

docs/en/04_Changelogs/4.0.0.md
chillu: Sorry for forking the discussion, I've commented a bit more on the original RFC: https://github.com/silverstripe/silverstripe-framework/issues/6221#issuecomment-274708979

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1179,15 +1279,34 @@ handle field-level and form-level messages. This has the following properties:
...
+  for all DataObject subclasses.
+* i18n key for locale-respective pluralisation rules added as '.PLURALS'. These can be configured
+  within yaml in array format as per [ruby i18n pluralization rules](http://guides.rubyonrails.org/i18n.html#pluralization).
+
+#### <a name="overview-i18n-removed"></a>i18n API Removed API
 
+* Zend_Translate removed
+* i18n::_t $context parameter deprecated
+* i18n::_t Support for sprintf-style `%s` arguments deprecated
+* i18n::_t Using non-associative injection with named parameters is now an error
```
robbieaverill: Minor: Can you either wrap things like this in backticks or escape the markdown characters to prevent the formatting breaking?

docs/en/04_Changelogs/4.0.0.md
dhensby: FQCN is probably what you mean, but I agree it's best to avoid these abreviations where possible

composer.json
```diff
@@ -22,7 +22,9 @@
     "league/flysystem": "~1.0.12",
     "symfony/yaml": "~2.7",
     "embed/embed": "^2.6",
-    "swiftmailer/swiftmailer": "~5.4"
+    "swiftmailer/swiftmailer": "~5.4",
+		"symfony/config": "^2.8|^3",
```
tractorcow: More like, in-don't-ation right?

docs/en/04_Changelogs/4.0.0.md
tractorcow: Added this back.

docs/en/04_Changelogs/4.0.0.md
tractorcow: Done as @sminnee suggested. `missing_default_warning` added.

docs/en/04_Changelogs/4.0.0.md
tractorcow: I think what's changed is that you can do `i18n::set_locale('es')` now, where before you could have a set of es strings, but you'd still need to set `i18n::set_locale('es_MX')`.

docs/en/04_Changelogs/4.0.0.md
tractorcow: I'll use `<Namespaced\ClassName>.PLURALS`

src/Dev/Backtrace.php
tractorcow: Oops, this was just debugging changes I forgot to revert. :)

src/Dev/Tasks/i18nTextCollectorTask.php
tractorcow: Sure, but in another change. :)

src/i18n/Messages/Symfony/FlushInvalidatedResource.php
tractorcow: It needs to be accessible to flush handler, which is a static method, otherwise I'd make it a constructor argument.

tractorcow:
Updated based on feedback, as well as result of conversation over at https://github.com/silverstripe/silverstripe-framework/issues/6221#issuecomment-275009696

tractorcow:
Note: CMS build can be merged ahead of framework build, so we can restart and see if these tests pass.

https://github.com/silverstripe/silverstripe-cms/pull/1724

tractorcow:
https://github.com/silverstripe/silverstripe-cms/pull/1724 is green

src/i18n/i18n.php
```diff
@@ -1970,256 +1956,167 @@ public static function get_cache()
...
+        // Encourage the provision of default values so that text collector can discover new strings
+        if (!$default && static::config()->get('missing_default_warning')) {
+            user_error("Missing default for localisation key $entity", E_USER_WARNING);
+        }
 
+        // Deprecate legacy injection format (`string %s, %d`)
+        // inject the variables from injectionArray (if present)
+        $sprintfArgs = [];
+        if ($default && !preg_match('/\{[\w\d]*\}/i', $default) && preg_match('/%[s,d]/', $default)) {
+            Deprecation::notice('5.0', 'sprintf style localisation variables are deprecated');
```
chillu: We also need to replace existing core uses: https://github.com/silverstripe/silverstripe-framework/issues/6582.

PR #6559, opened 2017-01-24
DOCS Add 4.0 upgrade guide, update "composer" and "modules" to include 4.x recommended practices

robbieaverill:


docs/en/00_Getting_Started/02_Composer.md
```diff
@@ -56,48 +64,57 @@ see [Using development versions](#using-development-versions).
 
 Composer isn't only used to download SilverStripe CMS, it can also be used to manage all SilverStripe modules.  Installing a module can be done with the following command:
 
-	composer require "silverstripe/forum:*"
+```
+composer require "silverstripe/forum:*"
+```
+
+This will install the forum module in the latest compatible version. You can also omit the `:*` if you simply want the latest compatible version of the module.
```
stevie-mayhew: Do we need to explain that omitting will result in the same action? If a user has `minimum-stability:dev` set in their `composer.json` it may also install master, which might not be compatible with their version (for various reasons)

docs/en/00_Getting_Started/02_Composer.md
```diff
@@ -56,48 +64,57 @@ see [Using development versions](#using-development-versions).
...
+The second part after the colon, `*`, is a version string.  `*` is a good default: it will give you the latest version that works with the other modules you have installed. Alternatively you can specify a specific version, or a constraint such as `^4.0`. For more information, read the [Composer documentation](http://getcomposer.org/doc/01-basic-usage.md#the-require-key).
 
 <div class="warning" markdown="1">
-`master` is not a legal version string - it's a branch name.  These are different things.  The version string that would get you the branch is `dev-master`.  The version string that would get you a numeric branch is a little different.  The version string for the `3.0` branch is `3.0.x-dev`. 
+`master` is not a legal version string - it's a branch name. These are different things. The version string that would get you the branch is `dev-master`. The version string that would get you a numeric branch is a little different. The version string for the `3` branch is `3.x-dev`.
```
stevie-mayhew: Update `3` and `3.x-dev` to 4 for consistency?

docs/en/00_Getting_Started/02_Composer.md
```diff
@@ -56,48 +64,57 @@ see [Using development versions](#using-development-versions).
...
+The `composer.lock` file helps with this. It references the specific commits that have been checked out, rather than the version string. You can run `composer install` to install dependencies from this rather than `composer.json`.
 
-So, your deployment process, as it relates to Composer, should be as follows:
+So your deployment process, as it relates to Composer, should be as follows:
 
- * Run `composer update` on your development version before you start whatever testing you have planned.  Perform all the necessary testing.
+ * Run `composer update` on your development version before you start whatever testing you have planned. Perform all the necessary testing.
  * Check `composer.lock` into your repository.
  * Deploy your project code base, using the deployment tool of your choice.
  * Run `composer install --no-dev -o` on your production version.
```
stevie-mayhew: Can we use the full `--optimize-autoloader` here so that it isn't a magic command for those who don't know it?

docs/en/00_Getting_Started/02_Composer.md
robbieaverill: I thought I'd leave this as 3 incase someone wants to install an old version, especially since `4` isn't a branch name yet

docs/en/00_Getting_Started/02_Composer.md
robbieaverill: The same is true for `*` thought right, unless you use `*@stable`? I think there's an example of that further down this page

docs/en/00_Getting_Started/02_Composer.md
robbieaverill: I might defer this one to @tractorcow. I'm not sure whether it would be considered as a "recommended practice" by us or not, but I'm sure he will :)

docs/en/00_Getting_Started/02_Composer.md
wernerkrauss: IMHO using "*" as a version is a bad practice. If all modules are using proper semver you should use ^1.2.3 to update in the v1 branch automatically (without updating your composer json) but use minimal v1.2.3, or 1.2.x to use the latest tagged version of v1.2, but not update to 2.

@stevie-mayhew From my experience composer will install the latest compatible version or throw an error, even with minimum-stability:dev, it will only result in incompatible code if the module's composer.json doesn't have the right requirements to the framework etc.

robbieaverill: >IMHO using "*" as a version is a bad practice.

While I agree completely, I think this example is trying to cater for people who don't know which major version they want to install.

Perhaps we should encourage them to update their constraint in `composer.json` and update again once they've found the latest stable version, or just omit the `*` to install (and track) the latest version by default?

wernerkrauss: good point. For the moment * seems to be a time saver, but later, when you want to update your site/app it's installing the latest stable compatible version - which might have breaking changes and unintended side effects to your existing app. This is a problem if your code relies on the installed module.

robbieaverill: Yeah. It's probably safer to encourage it to be omitted if you don't know the version you want to prevent this from being an issue later..

dhensby: I think we should drop references to `*` as it will be confusing for those who know a little about version constraints but not why `*` is bad. ie: the nuance here will be lost on them

docs/en/00_Getting_Started/02_Composer.md
dhensby: You're referencing master as the branch so I'd probably reference version 4.

docs/en/00_Getting_Started/02_Composer.md
dhensby: using the optimised auto-loader should definitely be our recommended practice.

Perhaps even `--classmap-authoritative`!

But I'm not sure that having the full `--optimize-autoloader` is needed - people can look it up if they need to, but teaching people to type a tone of characters they don't need seems a bit mean

stevie-mayhew: @dhensby if we stick with `-o` I'd like to see it explained. Teaching people to do something without explanation of what the functionality represents is also a bit mean 😄

docs/en/00_Getting_Started/02_Composer.md
robbieaverill: Yeah, good call

robbieaverill:
@dhensby @stevie-mayhew I've pushed a commit which updates the parts you both mentioned. I've added a little bit of an explanation around with `--no-dev` and `-o` to in the deployment section. If you're happy, I'll squash the commits.

stevie-mayhew:
Alright, I'm happy with that.

Issue #6562, opened 2017-01-24
LeftAndMain.js assumes iframe postmessages are strings


roed:
In LeftAndMain.js, there is a method that handles messages triggered in iframes (postMessage):

````
$(window).on("message", function(e) {
    var target,
        event = e.originalEvent,
        data = JSON.parse(event.data);
````
I tried to integrate a third party app that uses such messages, but the event.data is an object, not a string. This causes an error, because it tries to parse an object. Perhaps it is better to check if event.data is a string, and, if so, parse it:
````
data = typeof event.data == 'string' ? JSON.parse(event.data) : event.data;
````

dhensby:
want to open a PR?

roed:
Was already fixed in 4.0, but still made a PR.

dhensby:
thanks

michalkleiner:
Came across a Chrome plugin for Grammarly using this mechanism.
It sends string messages in `event.data`. The JS here fails to `JSON.parse` them. The logic needs further strengthening.

PR #6563, opened 2017-01-24
Fixed iframe postmessage breaking non-string messages

roed:
See issue: https://github.com/silverstripe/silverstripe-framework/issues/6562

This is a fix for that, which was already implemented in the code for version 4.0.

### DIFF
--- a/admin/javascript/LeftAndMain.js
+++ b/admin/javascript/LeftAndMain.js
@@ -73,7 +73,7 @@ jQuery.noConflict();
 		$(window).on("message", function(e) {
 			var target,
 				event = e.originalEvent,
-				data = JSON.parse(event.data);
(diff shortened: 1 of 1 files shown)


dhensby:
If this affects 3.4 I'll rebase it against that

dhensby:
hmm - you've opened a PR from your 3.5 branch, so we can't rebase against 3.4 :P

roed:
I am sorry. How can I fix this?

dhensby:
> I am sorry. How can I fix this?

You should only open PRs using branches intended for the PR. But it's ok we just won't fix it in 3.4 - it's a very low risk issue and 3.4 is technically only security fixes

roed:
Ok, so I should have made a feature/fix branch?

dhensby:
yes

roed:
Ok, will do that the next time

PR #6565, opened 2017-01-24
BUG Ensure root path of any local adapter is safely created and mapped from symlink

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/362


tractorcow:
I would like @robbieaverill to please confirm this fixes his issue before merging.

tractorcow:
Updated based on further exploration.

PR #6566, opened 2017-01-25
add segment config var to buildtask

### LINKED ISSUE #6556: Namespaced BuildTasks should be able to define a simpler URL
When creating a new `BuildTask`, the URL used to run the task is auto-generated based on the `ClassName`. However, with the introduction of namespacing, this means you end up with clunky URLs such as:

```
mysite.example.com/dev/tasks/andante-newproject-americasnexttopmodule-tasks-StillInTheRunningTask
```

I think there should be the ability to set the command URL, something like:

```php
public function getCommandURL()
{
  return "StillInTheRunningTask";
}
```

Which appends it to `dev/tasks`.

andrewandante:
Resolves #6556 

Backport of the SS4 `$segment` config variable to allow for namespaced `Buildtask`s to not have awful URLs.

### DIFF
--- a/dev/BuildTask.php
+++ b/dev/BuildTask.php
@@ -13,6 +13,14 @@
 abstract class BuildTask extends Object {
 
 	/**
+	 * Set a custom url segment (to follow dev/tasks/)
+	 *
+	 * @config
+	 * @var string
+	 */
+	private static $segment = null;
+
+	/**
 	 * @var bool $enabled If set to FALSE, keep it from showing in the list
 	 * and from being executable through URL or CLI.
 	 */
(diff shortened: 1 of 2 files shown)


dhensby:
thanks

dhensby:
were there any tests added in the 4.x change?

andrewandante:
Doesn't appear to be.

andrewandante:
Needs to be merged up the branches please @dhensby :)

dhensby:
3 only gets merged into master. we won't be adding this to any of the current minor releases, I'm afraid

PR #6567, opened 2017-01-25
 BUG Ensure that tests run with flush=1 clean Flushables

tractorcow:
flush=all will set a flushed test manifest, but no other flushables are invoked.

### DIFF
--- a/src/Dev/SapphireTest.php
+++ b/src/Dev/SapphireTest.php
@@ -12,6 +12,7 @@ use SilverStripe\Control\Controller;
 use SilverStripe\Control\Director;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\ClassInfo;
+use SilverStripe\Core\Flushable;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Manifest\ClassManifest;
 use SilverStripe\Core\Manifest\ClassLoader;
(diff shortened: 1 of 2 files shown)


dhensby:
So every test will be flushed with this change, is that expected?

tractorcow:
Oops, I screwed up the creation of this PR. Sorry, this was my attempt at breaking up a larger pull request. I'll re-pull.

tractorcow:
Sorry, I think you're right. flush is ment to be opt-out, not opt-in for tests. I need to exclude the `flush=0` condition.

tractorcow:
Ok, updated. ;)

dhensby:
Flushing being opt-out for tests was removed in 3.1: https://github.com/silverstripe/silverstripe-framework/pull/3813

dhensby:
test failures related

tractorcow:
Oh darn it, I see. The manifest was still being flushed every time, however.

tractorcow:
This code is currently in 4.0 (compare it to that pull request you linked).

```
public static function use_test_manifest()
    {
        $flush = true;
        if (isset($_GET['flush']) && $_GET['flush'] === '0') {
            $flush = false;
        }
```

How did that change get reverted between 3.1 and here?

tractorcow:
Answer, refactor at https://github.com/silverstripe/silverstripe-framework/pull/4404 somehow re-introduced it. :P

tractorcow:
Ok, fixed the broken test, and restored opt-in flushing.

PR #6569, opened 2017-01-26
BUGFIX: stop clear full cache on DataObject::write

sunnysideup:
I hope this is now done correctly (I did NOT test it), if not then I'd like to ask someone from the core team to do this as I dont really have the time to do this.

### DIFF
--- a/model/Aggregate.php
+++ b/model/Aggregate.php
@@ -1,5 +1,6 @@
 <?php
 
+
 /**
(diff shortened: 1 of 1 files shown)


dhensby:
Has to go against 3 because of new APIs

sunnysideup:
moved to https://github.com/silverstripe/silverstripe-framework/pull/6570

PR #6571, opened 2017-01-26
Cast SmallFieldHolder to HTMLFragment - fixes #6568

### LINKED ISSUE #6568: Bug with escaped HTML from LiteralField within FieldGroup
Noticed while trying to add a `LiteralField` to a `FieldGroup` that this happens:

image

Not 100% sure of the mechanics of `FormField` in SS4 yet, but since SS4 is escaping all strings that are not explicitly casted as HTML, does this mean that `SmallFieldHolder` needs to be added to `$casting` in `FormField`?  i.e.

```php
private static $casting = array(
    'SmallFieldHolder' => 'HTMLFragment'
);
```

I noticed that `FieldHolder` is already in `$casting` for `FormField`.  If that's the case, happy to submit a PR.

colintucker:
`LiteralField` should now render correctly in `FieldGroup`.

### DIFF
--- a/src/Forms/FormField.php
+++ b/src/Forms/FormField.php
@@ -255,6 +255,7 @@ class FormField extends RequestHandler
 
     private static $casting = array(
         'FieldHolder' => 'HTMLFragment',
+        'SmallFieldHolder' => 'HTMLFragment',
         'Field' => 'HTMLFragment',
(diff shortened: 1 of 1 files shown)


tractorcow:
Good catch.

PR #6575, opened 2017-01-29
FIX Ensure that Zend_ classes can still be autoloaded

### LINKED ISSUE #6572: [4.0] dev/build fails when requireDefaultRecords uses a Zend class
**Version:** dev-master#1c36ef9

To reproduce, add something like this to a Page class:

```php
public function requireDefaultRecords()
{
    $list = \SilverStripe\Forms\CountryDropdownField::create('dummy')->getSource();
}
```

Then run a dev/build:

```
CREATING DATABASE RECORDS

 * SilverStripe\SiteConfig\SiteConfig
 * SilverStripe\CMS\Model\SiteTree
 * Page
PHP Fatal error:  Class 'Zend_Locale' not found in /path/to/ss40/framework/src/Forms/CountryDropdownField.php on line 58

Fatal error: Class 'Zend_Locale' not found in /path/to/ss40/framework/src/Forms/CountryDropdownField.php on line 58
ERROR [Alert]: Class 'Zend_Locale' not found
IN GET /dev/build
Line 58 in /path/to/ss40/framework/src/Forms/CountryDropdownField.php
```

robbieaverill:
cc @tractorcow

This is a temporary change to support the autoloading of `Zend_*` classes since the removal of some of them during the Zend -> Symfony translation PR.

This change ensures that the Zend autoloader is instantiated early on in the process.

Resolves #6572.

### DIFF
--- a/src/Core/Core.php
+++ b/src/Core/Core.php
@@ -54,6 +54,14 @@ mb_regex_encoding('UTF-8');
  */
 gc_enable();
 
+/**
+ * Include the Zend autoloader. This will be removed in the near future.
+ */
+if (file_exists('thirdparty/Zend/Loader/Autoloader.php')) {
+    require_once 'thirdparty/Zend/Loader/Autoloader.php';
+    Zend_Loader_Autoloader::getInstance();
+}
+
 // Initialise the dependency injector as soon as possible, as it is
 // subsequently used by some of the following code
 $injector = new Injector(array('locator' => 'SilverStripe\\Core\\Injector\\SilverStripeServiceConfigurationLocator'));


tractorcow:
Hm, it's a bit risky to add a new autoloader config. Were we doing this earlier? I thought all we needed was an include_once in a few places to specific zend libraries.

robbieaverill:
This seemed like a tidier solution than adding that to dozens of classes, but I'm happy to do that instead for the sake of being explicit about it.

robbieaverill:
Seeing as this appears to only affect the command line - would you rather than I:

* Move the Zend autoloader to `cli-script.php`
* Remove it, and add `require_once` calls to each class that uses `Zend_Locale`
* Do the above for everything that uses `Zend_*` throughout the framework (21 files)

tractorcow:
Temp code, it can be sub-optimal for the time being.

PR #6579, opened 2017-01-30
FIX Update to ensure ThemeDir can handle cascading themes

### LINKED ISSUE #6578: [4.0] ViewableData::ThemeDir doesn't support cascading themes
`ViewableData::ThemeDir` hasn't been updated to support a `SilverStripe\View\SSViewer.themes` property - is still looking for the string value of `.theme`

robbieaverill:
* Return the first defined theme from the stack
* Suggest using the ThemeResourceLoader instead if looking for a specific file across themes

---

Resolves #6578


robbieaverill:
Closing as we will remove the method instead.

PR #6583, opened 2017-01-30
API Remove ViewableData::ThemeDir, update changelog to reflect suggested replacement

### LINKED ISSUE #6578: [4.0] ViewableData::ThemeDir doesn't support cascading themes

robbieaverill:
* Remove `ThemeDir` from `ViewableData`
* Add note to changelog regarding removal, and suggest using `ThemeResourceLoader::findThemedResource` in conjunction with `SSViewer::get_themes` instead.

Resolves #6578

### DIFF
--- a/docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
+++ b/docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
@@ -211,13 +211,17 @@ and production environments.
 You can also combine CSS files into a media-specific stylesheets as you would with the `Requirements::css` call - use
 the third paramter of the `combine_files` function:
 
-	:::php
-	$printStylesheets = array(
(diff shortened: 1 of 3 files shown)


robbieaverill:
Since it can't be done in a single call from templates without a controller method to wrap, should we add a similar method to `requireCSS` somewhere?

tractorcow:
Do you mean adding a themedImage function? Probably just best to hard-code it for that theme, to be honest.

robbieaverill:
True- it's probably a unique file to that theme

robbieaverill:
@tractorcow both points updated

PR #6586, opened 2017-01-31
BUG Use correct timeout for modal fade out

tractorcow:


tractorcow:
This file is in admin module now, doh.

tractorcow:
ok, moved to https://github.com/silverstripe/silverstripe-admin/pull/72

PR #6587, opened 2017-02-01
Removed references to an archived module

zanderwar:
Removed references to the [BBCodeParser module](https://github.com/silverstripe-archive/silverstripe-bbcodeparser) as it has been removed from SS4

> Removed TextParser and BBCodeParser. These are available in an archived module, silverstripe-archive/bbcodeparser
> We recommend that you update your code to make use of another BBCode parsing library.

### DIFF
--- a/.upgrade.yml
+++ b/.upgrade.yml
@@ -497,7 +497,6 @@ mappings:
   ViewableData: SilverStripe\View\ViewableData
   ViewableData_Customised: SilverStripe\View\ViewableData_Customised
   ViewableData_Debugger: SilverStripe\View\ViewableData_Debugger
-  BBCodeParser: SilverStripe\View\Parsers\BBCodeParser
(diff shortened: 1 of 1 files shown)


dhensby:
I'm not sure it's right to remove this from the upgrader. if people rely on it then they should be able to include the module and then run the upgrader all the same...

Perhaps these would fit better in the module itself, though?

zanderwar:
I definitely agree that these should be in the module itself. However the module is EOL due to it's recommendation to use something else.

I can move them across if you like @dhensby

robbieaverill:
>Perhaps these would fit better in the module itself, though?

Yeah - I think they should be moved over to the archived module. Otherwise people will upgrade their code from the framework mapping and hit errors when it doesn't exist. At least this way it won't be namespaced so you'll notice immediately that it's not in the upgrade map.

dhensby:
thanks

PR #6589, opened 2017-02-01
Fix pipe use in new composer constraint

chillu:
Single pipes aren't defined in the composer spec,
and cause issues when installing other dependencies like silverstripe/behat-extension.

See https://getcomposer.org/doc/articles/versions.md#range


chillu:
Closing in favour of https://github.com/silverstripe/silverstripe-framework/pull/6590

PR #6593, opened 2017-02-02
Missing .env error detection

sminnee:
If your database configuration isn’t found, it’s quite likely that this
is because you have a stray _ss_environment.php file that you haven’t
converted to a .env file.

This patch detects this situation and provides a more helpful error.

I also had to clean up DebugView a bit.


madmatt:
LGTM, don't know why git diff is complaining in Travis. No tests, but I'm not sure if it's complicated enough (or going to change enough) to make them worthwhile? There's a lot of other stuff in main.php that isn't tested either (at least, not directly).

dhensby:
The failure is because the `npm run build` command is building dist files that don't match the current repository...

It appears that the current master dist files have map files referenced but the travis build isn't referencing them :/

dhensby:
oh - ok, I see - it's pretty obvious, @sminnee's build is creating map files, which apparently isn't the standard=

dhensby:
Right, I've updated this code so that it more closely represents the old behaviour when looking for `_ss_environment.php` files - though I'd be tempted to actually avoid this completely and just make the existing error more relevant. `"SilverStripe Framework requires a $databaseConfig defined. Perhaps you need to set up a .env file, see [environment management docs]"`.

I've rebuilt the dist files and pushed them into the relevant commit. NPM build now passes.

I've also just noticed that there is no error response code when this new error message is thrown.

sminnee:
@dhensby my substantially simpler implementation covers 95% of the cases that people will run into. If we go with your more complete implementation then I'd be inclined to shift the file finder logic into a little generator and iterate on that. Or we could just look in BASE_PATH and its parent, as we do for .env files?

sminnee:
Adde HTTP 500 header and corrected dist files in my original change. I think it's a reasonable balance of bloat and functionality. Would be good to get this merged before alpha5 as the .env files are a common tripping point for people using ss4.

sminnee:
Failures are due to a lack of testing coverage in this code. I don't really see that as a blocker for the main.php code; DebugView could arguably have some test coverage as it's at least a modular unit, but not sure whether this blocks merge?

chillu:
Looking good (verified operation locally) - merging.

PR #6598, opened 2017-02-04
FIX Replace deprecated %s placeholders in translations with named placeholders

### LINKED ISSUE #6582: Replace deprecated sprintf rules in core translations
We've deprecated the use of `sprintf()` style placeholders (e.g. `%s`) in `_t()` calls with https://github.com/silverstripe/silverstripe-framework/pull/6558 - but core still uses it. There's around two dozen strings in core. Ideally we also replace this in the existing translations as well, so the translation strings don't get invalidated (we either have to fix them or remove them, otherwise will see stray `%s` in the UI)

robbieaverill:
This pull request largely removed the use of `sprintf(_t(TRANSNAME, '... %s'), $myVar)` in favour of the named placeholder equivalent `_t(TRANSNAME, '... {myvar}', ['myvar' => $myVar])`.

All translations have been updated to ensure they don't break before in the short term.

There are some areas especially in the tests that use this deliberately to test against, so I've left them alone for now, but perhaps we should remove the syntax from the tests and not test the behaviour any more?

Resolves #6582


robbieaverill:
@tractorcow I'll rebase this branch shortly.

Regarding the deprecation/removal of that method we talked about (#vague) - the changelog currently says that the `%s` format is deprecated and will be removed in 5.0, so perhaps we should deprecate that method too instead of removing it?

tractorcow:
The plan is to fork translations at some point in the future, and then merge this PR into that, before pushing up to transifex. That will allow us to replace upstream strings without affecting 3.x releases.

dhensby:
what's the progress on this? @robbieaverill it'll need a rebase at some point

robbieaverill:
>what's the progress on this? @robbieaverill it'll need a rebase at some point

@dhensby yeah I'll rebase it when we're ready to look at it, otherwise it'll just keep needing to be rebased as the translations change :D

PR #6601, opened 2017-02-06
BUG: Fix GridFieldExportButton column selection for export

madmatt:
`GridFieldPrintButton` uses a three-phase approach to finding columns to export:

 1. Look for `GridFieldPrintButton->printColumns` set directly on the button
 2. Look for the presence of `GridFieldDataColumns` config on the `GridField` to determine the columns included in the browser view
 3. Fallback on the `DataObject->summaryFields()` for the given object to export

By contrast, `GridFieldExportButton` only looks at (1) and (3) above. This isn't consistent across both buttons, which leads to a lot of people defining either their own implementation of `GridFieldExportButton`, or explicitly setting fields on the export button, but not needing to for the print button.

This fixes that, but making sure that `GridFieldExportButton` is updated to look in the same 3 places (in the same order) as `GridFieldPrintButton`.

### DIFF
--- a/forms/gridfield/GridFieldExportButton.php
+++ b/forms/gridfield/GridFieldExportButton.php
@@ -98,11 +98,16 @@ class GridFieldExportButton implements GridField_HTMLProvider, GridField_ActionP
 	 * @return array
 	 */
 	public function generateExportFileData($gridField) {
-		$separator = $this->csvSeparator;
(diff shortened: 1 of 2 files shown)


madmatt:
Never mind - this is fixed already in a later version of framework than I had locally (:sheep:ish)

PR #6604, opened 2017-02-07
FIX load fields if lazy ones exists

### LINKED ISSUE #6603: dbObject() method doesn't hydrate Lazy Fields
Discovered when using Blog module. There is a `canView()` method which relies on the following logic:

```php
$publishDate = $this->dbObject('PublishDate');

if (!$publishDate->exists()) {
    return false;
}
return !$publishDate->InFuture();
```

However, `$publishDate` never has a value, because the `dbObject()` method is unable to use `PublishDate_Lazy` as the value for the new object. Meaning no-one `canView()` if they aren't logged in.

Hotfix for that is to `setValue()` once the object is created, but that shouldn't be necessary.

andrewandante:
Resolves #6603 

Bonus test for good measure.

### DIFF
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -2721,6 +2721,11 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
             return null;
         }
 
+        if (!isset($this->record[$fieldName]) && isset($this->record[$fieldName . '_Lazy'])) {
+            $tableClass = $this->record[$fieldName . '_Lazy'];
+            $this->loadLazyFields($tableClass);
+        }
+
         $value = isset($this->record[$fieldName])
             ? $this->record[$fieldName]
             : null;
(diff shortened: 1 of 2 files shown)


src/ORM/DataObject.php
```diff
@@ -2725,6 +2725,13 @@ public function dbObject($fieldName)
             ? $this->record[$fieldName]
             : null;
 
+        if (!$value && isset($this->record[$fieldName . '_Lazy'])) {
+            $tableClass = $this->record[$fieldName . '_Lazy'];
+            $this->loadLazyFields($tableClass);
+
+            $value = isset($this->record[$fieldName]) ? $this->record[$fieldName] : null;
```
dhensby: I think we do need to investigate how a `$field . '_Lazy'` can still be unset after calling `loadLazyFields`

sminnee: `!$value` matches false. would suggest using  `!isset($this->record[$fieldName])` in this if block instead, and moving the value-setter on line 2724 to below this if clause. then line 2732 isn't needed.

andrewandante:
Reordered as per @sminnee recommendations - failing test is the code coverage one.

PR #6607, opened 2017-02-08
API Substitute Zend_Locale / Zend_Date with php-intl

### LINKED ISSUE #6194: RFC Replace Zend_Locale with new i18n backend

tractorcow:
Fixes #6194

Requires:

- [ ]  https://github.com/silverstripe/silverstripe-reports/pull/59
- [ ] https://github.com/silverstripe/silverstripe-cms/pull/1734
- [ ] https://github.com/silverstripe/silverstripe-testsession/pull/43

Note: This PR removes about 1/4 of a million lines of code. Yay. :D

Includes a few other enhancements:
 -  i18nTextCollector supports `__CLASS__`
 - Update FormField with a `setSubmittedValue` to differentiate between loading values from the database and loading values from user-submitted field values.
 - Add `Resettable` interface in place of having to manually invoke `reset()` on explicit classes between test resets.
 
 Some minor functionality around currency names could not be ported across, and have been removed.


tractorcow:
Some immediate feedback is that Resettable is a bad design pattern. Perhaps we should address this at a later date with better locale caching.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1051,6 +1052,26 @@ A very small number of methods were chosen for deprecation, and will be removed
...
+  will be formatted in the default locale.
+
+The below methods have been added or had their functionality updated to `DBDate`, `DBTime` and `DBDatetime`
+* `getTimestamp()` added to get the respective date / time as unix timestamp (seconds since 1970-01-01)
+* `Format()` method now use CLDR format strings, rather than PHP format strings.
+  See http://userguide.icu-project.org/formatparse/datetime.
+* getISOFormat() added which returns the standard date/time ISO 8601 pattern in CLDR format.
+* Dates passed in m/d/y format will now raise a notice but will be parsed.
+  Dates passed to constructors should follow ISO 8601 (y-m-d).
+* 2-digit years will raise a notice.
```
chillu: "2-digit years in `setValue()` will raise a notice, since they're ambiguous without additional formatting hints"

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -941,6 +941,7 @@ specific functions.
   service name from the filename. Now the service name will either
   by the array key, or the `class` parameter value.
 * Uniqueness checks for `File.Name` is performed on write only (not in `setName()`)
+* Created `Resettable` interface to better declare objects which should be reset between tests.
```
chillu: Add the following to the highlevel changelog summary:

 * Added a server requirement for the [php-intl](http://php.net/manual/en/book.intl.php) extension (shipped by default with most PHP distributions)
 * Replaced Zend_Date and Zend_Locale with the [php-intl](http://php.net/manual/en/book.intl.php) extension.
 * Consistently use CLDR date formats (rather than a mix of CLDR and date() formats)

Also, you need to update server-requirements.md

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1051,6 +1052,26 @@ A very small number of methods were chosen for deprecation, and will be removed
...
+* 2-digit years will raise a notice.
+* `FormatFromSettings` will default to `Nice()` format if no member is logged in.
+* `Nice`, `Long` and `Full` methods will now follow standard formatting rules for the
+  current locale, rather than pre-defined formats.
+* `Short` added to format concise date/time values, as a shorter version than `Nice`
+* `getFormatter` method added, which returns a locale-specific date/time formatter.
+
+`DBTime` specific changes:
+* Added `DBTime::FormatFromSettings`
+* Added `DBTime::Nice12`
```
chillu: Shouldn't we remove `Nice24` rather than adding `Nice12`, given that you do all of this via custom formatting now? The original purpose for this stuff was the lack of passing complex arguments via SSViewer, but that's since been resolved.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1275,6 +1314,28 @@ handle field-level and form-level messages. This has the following properties:
 
 * `$message` second constructor parameter is removed. Constructor only accepts `$result`,
   which may be a string, and optional `$code`
+  
+New `DatetimeField` methods replace `getConfig()` / `setConfig()`:
```
chillu: What happened with `datavalueformat`?

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1320,12 +1383,30 @@ handle field-level and form-level messages. This has the following properties:
   for all DataObject subclasses, rather than just the basename without namespace.
 * i18n key for locale-respective pluralisation rules added as '.PLURALS'. These can be configured
   within yaml in array format as per [ruby i18n pluralization rules](http://guides.rubyonrails.org/i18n.html#pluralization).
+* `i18n.all_locales` config moved to `Locales.locales`
```
chillu: All of these should mention the correct namespace, otherwise it's harder to fix via copy/paste.

src/Core/Resettable.php
```diff
@@ -0,0 +1,17 @@
...
+
+namespace SilverStripe\Core;
+
+/**
+ * Represents a class with a local cache which normally invalidates itself between requests.
+ *
+ * Designed so that tests can automatically flush these objects between runs in lieu
+ * of a real change.
+ */
+interface Resettable
```
chillu: Shouldn't we move this stuff to explicit registration via YAML config? My understanding was that `ClassInfo::implementorsOf()` is on it's way out, because it forces us to parse all PHP files on a flush (which is no longer required for autoloading thanks to PSR-7). The `interface Resettable` is fine, but I wouldn't use `ClassInfo::implementorsOf('Resettable')` to retrieve those classes.

src/Forms/DateField.php
```diff
@@ -58,175 +33,320 @@
...
+    protected $dateFormat = null;
 
     /**
-     * @var array
+     * Set if js calendar should popup
+     *
+     * @var bool
      */
-    protected $config;
+    protected $showCalendar = false;
```
chillu: One of the reasons I've avoided elevating these settings to class properties is that I didn't want to lock in the API to that extent - it's a bit like using `FormField->setAttributes()` to influence client-side react component presentation.

So rather than increasing the API surface of `DateField`, what do you think about separating out `DateFieldSeparated extends DateField` and `DateFieldJQuery extends DateField`? Then we can set the `DateField` injector service name to resolve to`DateFieldJQuery` by default, in order to get correct behaviour for built-in features like `DBDate->scaffoldFormField()`

docs/en/04_Changelogs/4.0.0.md
tractorcow: It's fixed to iso format. No more custom format for internal values.

src/Forms/MoneyField.php
```diff
@@ -211,15 +293,27 @@ public function getAllowedCurrencies()
...
+    /**
+     * Assign locale to format this currency in
+     *
+     * @param string $locale
+     * @return $this
+     */
     public function setLocale($locale)
     {
-        $this->_locale = $locale;
+        $this->fieldAmount->setLocale($locale);
```
chillu: This breaks for readonly versions of this field:

> [Emergency] Uncaught BadMethodCallException: Object->__call(): the method 'getLocale' does not exist on 'SilverStripe\Forms\NumericField_Readonly'

src/Forms/DateField.php
```diff
@@ -58,175 +33,320 @@
...
-            $fields[stripos($format, 'y')] = $fieldYear->Field();
-            ksort($fields);
-            $html = implode($sep, $fields);
-
-            // dmyfields doesn't work with showcalendar
-            $this->setConfig('showcalendar', false);
-        } // Default text input field
-        else {
-            $html = parent::Field();
+        $valArr = $this->iso8601ToArray($this->Value());
```
chillu: This breaks when `locale` isn't `en_us`, since `Value()` uses `iso8601ToLocalised()`, but `iso8601ToArray()` expects ISO

src/Forms/NumericField.php
```diff
@@ -132,71 +205,105 @@ public function getSchemaValidation()
...
     }
 
     /**
-     * Creates a read-only version of the field.
+     * Determine if we should use html5 number input
      *
-     * @return NumericField_Readonly
+     * @return bool
      */
-    public function performReadonlyTransformation()
```
chillu: Somehow the readonly version doesn't respect `setLocale('de')` or `setScale(4)`. The disabled version is fine, weird. Example from a slightly modified frameworktest:

image

src/Forms/NumericField.php
```diff
@@ -132,71 +205,105 @@ public function getSchemaValidation()
...
+
+    /**
+     * Get number of digits to show to the right of the decimal point.
+     * 0 for integer, any number for floating point, or null to flexible
      *
+     * @param int|null $scale
      * @return $this
      */
-    public function setLocale($locale)
+    public function setScale($scale)
```
chillu: Where did you get the "scale" terminology from? The ICU docs behind PHP's `NumberFormatter` refer to this as the "significant digit": http://www.icu-project.org/apiref/icu4c/classDecimalFormat.html#details

src/ORM/FieldType/DBDate.php
```diff
@@ -271,18 +309,23 @@ public function Rfc822()
     public function Rfc2822()
     {
         if ($this->value) {
-            return date('Y-m-d H:i:s', strtotime($this->value));
+            return date('Y-m-d H:i:s', $this->getTimestamp());
```
chillu: Shouldn't we use `getFormatter()` consistently here, rather than relying on `date()` for some of the features? It'll be confusing to refer back to two different symbol systems (`date()` and ICU)

src/ORM/FieldType/DBDate.php
```diff
@@ -432,85 +501,129 @@ public function InFuture()
...
      */
-    public static function past_date($fmonth, $fday = 1, $fyear = null)
+    protected function explodeDateString($value)
     {
-        if (!$fyear) {
-            $fyear = date('Y');
+        // US date format with 4-digit year first
+        if (preg_match('#^(?<year>\\d{4})/(?<day>\\d+)/(?<month>\\d+)(?<time>.*)$#', $value, $matches)) {
+            trigger_error(
+                "Implicit y/d/m conversion. Use " . self::ISO_DATE . " to prevent this notice.",
```
chillu: Add value to error message (since `y/d/m` is more of a guess at this point, could well be `y/m/d`)

src/ORM/FieldType/DBDate.php
```diff
@@ -432,85 +501,129 @@ public function InFuture()
...
+            return [$matches['day'], $matches['month'], $matches['year'], $matches['time']];
         }
-        $fday = (int) $fday;
-        $fmonth = (int) $fmonth;
-        $fyear = (int) $fyear;
 
-        $pastDate = mktime(0, 0, 0, $fmonth, $fday, $fyear);
-        $curDate = mktime(0, 0, 0, date('m'), date('d'), $fyear);
+        // US date format without 4-digit year first: assume m/d/y
+        if (preg_match('#^(?<month>\\d+)/(?<day>\\d+)/(?<year>\\d+)(?<time>.*)$#', $value, $matches)) {
```
chillu: How do you know that it's `m/d/y` here? Could be NZ date format with a short year (`d/m/y`), e.g. `31/07/17`. I'd change the error message to "Ambiguous short year format detected. Use <ISO_DATE> to prevent this notice". Also, add the value to the error message.

src/ORM/FieldType/DBDate.php
```diff
@@ -432,85 +501,129 @@ public function InFuture()
...
      *
-     * @param $fmonth int The number of the month (e.g. 3 is March, 4 is April)
-     * @param $fday int The day of the month
-     * @param $fyear int Determine historical value
-     * @return string Date in YYYY-MM-DD format
+     * @param string $value
+     * @return array
      */
-    public static function past_date($fmonth, $fday = 1, $fyear = null)
+    protected function explodeDateString($value)
```
chillu: Why did you replace the `new DateTime($value)` based parsing with regexes here? `DateTime` is based on `strtotime()`, which supports [more formats](http://php.net/manual/en/datetime.formats.date.php) than you cover here. Is this an attempt to warn when passing invalid or ambiguous formats like two-digit years, and narrow down edge cases? Seems reasonable. but will require a changelog entry. The PHPDoc should also explain why it's not using strtotime()

src/i18n/Data/Intl/IntlLocales.php
```diff
@@ -0,0 +1,1622 @@
...
+    {
+        if (!$locale) {
+            return false;
+        }
+        $lang = $this->langFromLocale($locale);
+        $region = $this->countryFromLocale($locale);
+        if (!$lang || !$region) {
+            return false;
+        }
+        return strcasecmp($lang, $region)
```
chillu: Is multiplication really what you want here? Seems like a roundabout way of expressing a conditional

src/i18n/TextCollection/i18nTextCollector.php
```diff
@@ -548,12 +549,37 @@ public function collectFromCode($content, $module)
...
+        $inNamespace = false;
+        $inClass = false;
         $inArrayClosedBy = false; // Set to the expected closing token, or false if not in array
         $currentEntity = array();
+        $currentClass = []; // Class components
         foreach ($tokens as $token) {
             if (is_array($token)) {
                 list($id, $text) = $token;
 
+                // Check class
```
chillu: What are these text collector changes about?

src/i18n/i18n.php
```diff
@@ -2396,97 +355,7 @@ public static function get_locale()
...
-            }
-            $path = $locator->getPath($theme);
-            $langPath = "{$path}/lang/";
-            if (is_dir($langPath)) {
-                $paths[] = $langPath;
-            }
-        }
-
-        return $paths;
+        return static::getData()->scriptDirection($locale);
```
chillu: Why didn't you move this to `Locale` as well?

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1051,6 +1052,26 @@ A very small number of methods were chosen for deprecation, and will be removed
 * `ChangeSet` and `ChangeSetItem` have been added for batch publishing of versioned dataobjects.
 * `DataObject.table_name` config can now be used to customise the database table for any record.
 * `DataObjectSchema` class added to assist with mapping between classes and tables.
+* `DBMoney` values are now treated as empty only Amount is null. Values without Currency
```
tractorcow: Updated to
```
* `DBMoney` values are now treated as empty only if `Amount` field is null. If an `Amount` value
  is provided without a `Currency` specified, it will be formatted as per the current locale.
```

docs/en/04_Changelogs/4.0.0.md
tractorcow: Sure. :)

docs/en/04_Changelogs/4.0.0.md
tractorcow: Added `* `datavaluefield` config is removed as internal data value is now fixed to ISO 8601 format`

src/Core/Resettable.php
tractorcow: That's a problem I'm not sure we can solve in this pull request.

src/Forms/DateField.php
tractorcow: I've split out DateFieldSeparated, although I'm a bit hesitant to make a jquery-specific class which may limit usability (e.g. in react-formschema environments).

src/Forms/FormField.php
```diff
@@ -769,17 +770,34 @@ public function attrValue()
...
-     * @param null|array|DataObject $data {@see Form::loadDataFrom}
+     * If a FormField requires specific behaviour for loading content from either the database
+     * or a submitted form value they should override setSubmittedValue() instead.
+     *
+     * @param mixed $value Either the parent object, or array of source data being loaded
+     * @param array|DataObject $data {@see Form::loadDataFrom}
      * @return $this
      */
-    public function setValue($value)
+    public function setValue($value, $data = null)
```
tractorcow: Not sure where forms.md is; I'll have a look at refreshing the forms docs dir though.

tractorcow: Found and fixed. :)

src/Forms/MemberDatetimeOptionsetField.php
```diff
@@ -134,9 +153,18 @@ public function setValue($value)
...
         }
 
-        return parent::setValue($value);
+        return parent::setSubmittedValue($value);
+    }
+
+    public function setValue($value, $data = null)
+    {
+        if (is_array($value)) {
+            throw new InvalidArgumentException("Invalid value");
```
tractorcow: `throw new InvalidArgumentException("Invalid array value: Expected string");`

src/Forms/MoneyField.php
```diff
@@ -86,48 +84,124 @@ public function __clone()
...
     {
         $name = $this->getName();
+
+        // Validate allowed currencies
+        $currencyValue = $this->fieldCurrency ? $this->fieldCurrency->dataValue() : null;
         $allowedCurrencies = $this->getAllowedCurrencies();
-        if ($allowedCurrencies) {
-            $field = new DropdownField(
+        if (count($allowedCurrencies) === 1) {
+            // Dropdown field for multiple currencies
```
tractorcow: Oh dear, will fix!

src/Forms/MoneyField.php
```diff
@@ -86,48 +84,124 @@ public function __clone()
...
+            $this->value = null;
+            $this->fieldCurrency->setValue(null);
+            $this->fieldAmount->setValue(null);
+            return $this;
+        }
+
+        // Convert string to array
+        // E.g. `44.00 NZD`
+        if (is_string($value) &&
+            preg_match('/^(?<amount>[\\d\\.]+)( (?<currency>\w{3}))?$/i', $value, $matches)
```
tractorcow: >  It also means you can't properly prepopulate a field with values, without casting into DBMoney first.

You can pass in an array with `Amount` and `Currency` keys.

tractorcow: Oh dear, I'm throwing an exception... I'd intended to support that array format, sorry!

tractorcow: Changed error case to

```php
} elseif (!is_array($value)) {
    throw new InvalidArgumentException("Invalid currency format");
}
```

src/Forms/MoneyField.php
tractorcow: I've modified this so that fieldAmount remains a NumericField, but is assigned setReadonly(true) instead of being type-shifted.

src/Forms/DateField.php
tractorcow: Oops! Can fix with $this->dataValue() instead .:)

src/Forms/NumericField.php
tractorcow: Deleting NumericField_Readonly helped. :)

src/Forms/NumericField.php
tractorcow: It was a bit of digging, but http://math.stackexchange.com/questions/64042/what-are-the-numbers-before-and-after-the-decimal-point-referred-to-in-mathemati

tractorcow: Nope nothing to do with significant digits. scale is number of digits to the RIGHT of the decimal point. significant digits means non-zero digits from the left of the number.

tractorcow: ok, it looks like the correct term is `mantissa`.

tractorcow: I can't find the correct term, to be honest. Closest reference I could find is https://msdn.microsoft.com/en-us/library/ms190476.aspx

`Scale is the number of digits to the right of the decimal point in a number. For example, the number 123.45 has a precision of 5 and a scale of 2.`

src/ORM/FieldType/DBDate.php
tractorcow: I was too lazy to re-write the strings...

src/ORM/FieldType/DBDate.php
tractorcow: As with DateField, I'm not trying to support arbitrary custom formats. DBDate / DBDatetime are intended to load records from the database, and I've chosen a restrictive set of date formats based on this restriction.

If you are passing in raw user into into DBDate / DBDatetime then something is wrong.

src/ORM/FieldType/DBDate.php
tractorcow: who would do such a thing though? =(

tractorcow: I'm just going to stop falling back and fail any non-iso 8601 format. Too much maintenance.

tractorcow: `DBDate / DBDatetime now fail if passed not-iso8601 date as values`

src/ORM/FieldType/DBDate.php
tractorcow: Slashes = US date.

tractorcow: http://php.net/manual/en/function.strtotime.php#100144


```
The best way to compensate for this is by modifying your joining characters. Forward slash (/) signifies American M/D/Y formatting, a dash (-) signifies European D-M-Y and a period (.) signifies ISO Y.M.D. 
```

src/Core/Resettable.php
chillu: My point is that introducing *more* use cases is taking it in the wrong direction. But you're right, this is best addressed separately - the interface itself is good, we might just move all those to explicit registration at some point.

src/i18n/Data/Intl/IntlLocales.php
tractorcow: I can make it more explicit.

src/i18n/i18n.php
tractorcow: Because i18n has a global template provider that needs this method.

src/i18n/TextCollection/i18nTextCollector.php
tractorcow: I'll add to 4.0.0.md. This is so that you can use `_t(__CLASS__.'.SOMEKEY')` instead of the FQN.

tractorcow:
I've added another commit with changes requested. Will fix tests tomorrow.

src/ORM/FieldType/DBDate.php
tractorcow: I'm going to simplify the hell out of this.

src/Forms/NumericField.php
chillu: Actually, both left/right digits are "significant" - quoting https://en.wikipedia.org/wiki/Significant_figures#Rounding_and_decimal_places

> Trailing zeros in a number containing a decimal point are significant. For example, 12.2300 has six significant figures: 1, 2, 2, 3, 0 and 0

But yeah, you're correct that "significant digits" is the wrong term here. I'm used to "decimal places", but since you've documented it well, I'm also happy with "scale"

src/ORM/FieldType/DBDate.php
chillu: Fair enough, there's still around two dozen occurrences of `date()` in core - low priority, not a merge blocker.

src/Forms/NumericField.php
tractorcow: Yes, significant figures are counted FROM the left, but potentially including digits on the right. Not ONLY whole number values. :)

tractorcow:
Ok, if tests pass this is good to merge.

tractorcow:
ok, all non-cms tests passing. CMS tests will pass with cms pr merged.

src/Core/Resettable.php
sminnee: My thinking is that we should introduce some kind of `Application` object that contains any in-memory caches and other objects that we might want to re-set, and have our tests create new Application objects. But it's a separate PR.

src/Forms/HTMLEditor/TinyMCEConfig.php
```diff
@@ -15,6 +16,172 @@
  */
 class TinyMCEConfig extends HTMLEditorConfig
 {
+    /**
+     * @config
+     * @var array
+     */
+    private static $tinymce_lang = [
```
sminnee: It seems like this static could only list the exceptions to the rule `strtok($input, '_')` and be a lot shorter.

src/i18n/TextCollection/i18nTextCollector.php
sminnee: i18nTextCollector could probably use a rewrite to make use of php-parser. Separate optional PR, though.

docs/en/02_Developer_Guides/03_Forms/Field_types/02_DateField.md
```diff
@@ -51,19 +51,16 @@ strtotime()).
...
-		->setConfig('min', '-7 days')
-		->setConfig('max', '2012-12-31')
+		->setMinDate('-7 days')
+		->setMaxDate'2012-12-31')
 		
 ## Separate Day / Month / Year Fields
 
-The following setting will display your DateField as three input fields for day, month and year separately. HTML5 
-placeholders 'day', 'month' and 'year' are enabled by default. 
+To display separate input fields for day, month and year separately you can use the `DateFieldSeparated` subclass`.
```
chillu: This references `DateFieldSeparated` rather than `SeparatedDateField` - I've fixed this in a follow-up commit.

src/Forms/SeparatedDateField.php
```diff
@@ -0,0 +1,136 @@
...
+
+        // Set placeholders
+        if ($this->getPlaceholders()) {
+            $fieldDay->setAttribute('placeholder', _t(__CLASS__ . '.DAY', 'Day'));
+            $fieldMonth->setAttribute('placeholder', _t(__CLASS__ . '.MONTH', 'Month'));
+            $fieldYear->setAttribute('placeholder', _t(__CLASS__ . '.YEAR', 'Year'));
+        }
+
+        // Join all fields
+        // @todo custom ordering based on locale
```
chillu: I'm just implementing this - it's a needless feature regression compared to 3.x

src/Forms/DateField.php
chillu: OK, I've reinstated format-based ordering and custom separators now

src/ORM/FieldType/DBDate.php
chillu: We've discussed this a bit more on Slack, and Damian confirmed that the intent was to narrow down supported cases. Given this is populated from the database or somewhat filtered/sanitized user input, that's OK

dhensby:
whoop

chillu:
Created a follow up issue with https://github.com/silverstripe/silverstripe-framework/issues/6626

PR #6612, opened 2017-02-09
Fixes for using dblib PDO driver.

ajoneil:
These fixes allow *nix environments to connect to SQL Server using
the dblib PDO driver and the silverstripe mssql module.

  - Only set MYSQL_ATTR_INIT_COMMAND when using the mysql driver, this
    constant isn't defined if the mysql pdo driver isn't installed
  - Supress warnings on getting the server version, attributes aren't
    supported by the dblib driver
  - Explicitly check for errors on sql exec, checking the return
    value isn't reliable for statements with no return value (e.g.
    USE database)


model/connect/PDOConnector.php
```diff
@@ -227,7 +228,7 @@ public function exec($sql, $errorLevel = E_USER_ERROR) {
 		$result = $this->pdoConnection->exec($sql);
 
 		// Check for errors
-		if ($result !== false) {
+		if (!$this->hasError($result)) {
```
tractorcow: The problem is that `false` as a return value now reports hasError() = false, when before it was false. Docs on this method:

```
PDO::exec() returns the number of rows that were modified or deleted by the SQL statement you issued. If no rows were affected, PDO::exec() returns 0.
```

I think the `!== false` check was correct.

tractorcow: You could also fix it by making hasError() return true for `=== false`

ajoneil: I'm not sure what you mean by the `false` return value here - this still returns the result set if the query was successful, or `null` if there was an error.

The `PDO::exec()` docs don't match with what I'm seeing - this is returning `false` with a `USE database` statement with the dblib driver, but the database select is successful.

ajoneil: I reorganised the code a little here, so it's easier to read.

dhensby:
Although this is something we can only do in 4, really anything we are finding we have to wrap in "is this a xxx kind of driver" should be logic that each connector provides.

Each connector should provide the connection options, the DSN, and so on.

model/connect/PDOConnector.php
```diff
@@ -178,7 +179,7 @@ public function connect($parameters, $selectDB = false) {
 	}
 
 	public function getVersion() {
-		return $this->pdoConnection->getAttribute(PDO::ATTR_SERVER_VERSION);
+		return @$this->pdoConnection->getAttribute(PDO::ATTR_SERVER_VERSION);
```
dhensby: Why do we need to suppress errors here?

ajoneil: "Supress warnings on getting the server version, attributes aren't supported by the dblib driver"

tractorcow:
Damn it, I wrote some comments 6 days ago but forgot to submit review (so they were hidden). Sorry!

tractorcow:
I think we need to use `hasError($this->pdoConnection)` instead of `hasError($result)` here.

ajoneil:
Looks like you're right with checking the error on the connection, I'll fix that up.

chillu:
Yay a wild Andrew appears!

tractorcow:
Looks good; Let's see if travis wants to work today.

sminnee:
@tractorcow are you happy with the latest incarnation of this?

tractorcow:
Yes. :)

PR #6616, opened 2017-02-12
Cyrillic Transliteration

flamytwista:


flamytwista:
I think it's comfortable to transit greek, cyrillic and other charsets in one way.
Transliteration is particularly used in cms when creating URLSegment(slug) based on title.
Content manager potentially can insert any symbols in title field, and title must be translited into latin.

For example russian and ukrainian alphabets has differences in just 2 symbols and both can be used in one site.


SHSH mistake is fixed


"The Travis CI build failed" - is it important for this pull request?
I've used the same code convention as used in current file.

dhensby:
This can't go in the 3.5.3 release and it probably can't go in 3.5, it needs to go in 3.

Our transliterator is not meant to be definitive, just provide some help - people can correct their URLs if the need to.

Thanks for adding Cyrillic @flamytwista - I'm not sure why the CMS test is broken, but I'm going to move this over to 3 branch and we'll go from there.

gregsmirnov:
We just use custom class to support more languages with `Object::useCustomClass('SS_Transliterator', 'TransliteratorExtended');`. I completely agree with @flamytwista with Cyrilic transliteration. And @dhensby, I can PR Greek mapping as well.

gregsmirnov:
added PR for Greek as well #6620

PR #6619, opened 2017-02-13
FIX Comments ignored by classmanifest

### LINKED ISSUE #6615: Class not found if the class line ends with a comment
SilverStripe version: 3.4.1

I run into this issue when I added a comment to one of my Controller classes which I knew was previously working well. I haven't tested this with other kind of classes (i.e. DataObjects), but I think it would be reasonable if the same issue affects other classes too.

1. Create a new Controller class and add a single line comment:
```PHP
class MyController extends Controller //Some comment
{
    private static $allowed_actions = array(
        'test' => true,
    );
	
    private static $url_handlers = array(
        'test'	=> 'test',
    );

    public function test()
    {
        return 'MyController works.';
    }
}
```

2. Define the route:
```YAML
Director:
  rules:
    'MyController//$Action': 'MyController'
```

3. Run `/dev/build?flush=all`

4. Run `/MyController/test`. You should now get this error: `[User Error] Uncaught ReflectionException: Class MyController does not exist`.

(A couple of side notes: I usually name my controllers like `Something_Controller` but in this example I didn't use that pattern. Also it's not a huge issue being not able to put comments in the same line with the `class` keyword, but I just wanted to open this issue so that if someone else is wondering why their class is not found, then perhaps this issue can be of help. The reason why I originally put the comment to that line was that I wanted to write a note about why I was extending another custom controller instead of straight the base Controller class, so I thought it would be reasonable to put the comment near the `extends` keyword. Well, I can place it elsewhere though :).)

dhensby:
Fixes #6615

`TokenizedRegularExpression` now takes an optional list of tokens to ignore when searching for matches.

This means that an unexpected comment or whitespace sprinkled between the expected structure of the TRE is ignored rather than causing silen failures or breakages.

### DIFF
--- a/core/manifest/ClassManifest.php
+++ b/core/manifest/ClassManifest.php
@@ -61,16 +61,16 @@ class SS_ClassManifest {
 	public static function get_namespaced_class_parser() {
 		return new TokenisedRegularExpression(array(
 			0 => T_CLASS,
-			1 => T_WHITESPACE,
+			1 => array(T_WHITESPACE, 'optional' => true),
 			2 => array(T_STRING, 'can_jump_to' => array(8, 16), 'save_to' => 'className'),
-			3 => T_WHITESPACE,
+			3 => array(T_WHITESPACE, 'optional' => true),
 			4 => T_EXTENDS,
-			5 => T_WHITESPACE,
+			5 => array(T_WHITESPACE, 'optional' => true),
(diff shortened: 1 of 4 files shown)


dhensby:
There is something weird going on - I was testing this locally and all fine. then it seems to break when I checkout again and re-test.

Maybe the tests are caching the manifests?!

PR #6621, opened 2017-02-14
Original File Title charset on upload.

flamytwista:
For example content-manager uploads "мой файл.jpg"
File name will be "moi-fail.jpg"
but file title will be "мой файл".

For now framework makes file title like "moi-fail".

I think it's comfortable to left original filename's charset for file title.
What do you think about it?

### DIFF
--- a/filesystem/Upload.php
+++ b/filesystem/Upload.php
@@ -208,6 +208,7 @@ class Upload extends Controller {
 			$this->file->ParentID = $parentFolder ? $parentFolder->ID : 0;
 			// This is to prevent it from trying to rename the file
 			$this->file->Name = basename($relativeFilePath);
+			$this->file->Title = preg_replace("/{$fileSuffix}$/", '', $tmpFile['name']);
 			$this->file->write();
 			$this->file->onAfterUpload();
 			$this->extend('onAfterLoad', $this->file, $tmpFile);   //to allow extensions to e.g. create a version after an upload


filesystem/Upload.php
```diff
@@ -208,6 +208,7 @@ public function load($tmpFile, $folderPath = false) {
 			$this->file->ParentID = $parentFolder ? $parentFolder->ID : 0;
 			// This is to prevent it from trying to rename the file
 			$this->file->Name = basename($relativeFilePath);
+			$this->file->Title = preg_replace("/{$fileSuffix}$/", '', $tmpFile['name']);
```
dhensby: `$fileSuffix` needs to be correctly escaped for use in a regular expression. I'm not sure a regular expression is even needed here.

flamytwista: @dhensby 
Potentially filename can be file.jpgname.jpg
str_replace is not suitable in this case:
regular expression is needed for check extension only in the end of string.

I have no idea how to correctly escape $fileSuffix for regexp.
Can you advice some solution?

dhensby: `preg_quote` is your friend.

However, isn't `$fileSuffix` already transliterated by this point and so won't always match `$tmpFile['name']`? For example  `мой.файл.jpg` wouldn't have the expected results (`.файл.jpg` is the file suffix in this case, which will have been transliterated to `.fail.jpg`).

To be honest, if you want this behaviour you should just turn the transliteration off `FileNameFilter.default_use_transliterator = false`

flamytwista:
@dhensby 
FileNameFilter.default_use_transliterator = false
In this case transliteration of file name not works: "мой файл.jpg" will be "1.jpg", "myфайл" will be "my.jpg"

The goal is to translit filename and left original charset in file title.
Goal is on these screenshots.
https://yadi.sk/i/yIaZxNIP3E5KvP
https://yadi.sk/i/uYO-nUhI3E5LKX

filesystem/Upload.php
flamytwista: @dhensby
Thanks for `preg_quote`

`FileNameFilter.default_use_transliterator = false`  isn't a solution. 
Look here.
https://github.com/silverstripe/silverstripe-framework/pull/6621#issuecomment-279725738

Yes, $fileSuffix is already transliterated by this point and situation you've described is possible.

For now I haven't graceful way how to reach a goal and not break something else.

So, what will we do with this PR?
Maybe we'll close it and I'll create new PR next time?
Or can I commit to current PR?

dhensby:
Whilst we don't have a resolution we can close this. It's a shame non-latin chars are stripped no matter what...

flamytwista:
@dhensby 
Close this please.

For now I will use my not perfect solution with extension and hook.
I hope me or somebody else will find graceful solution later.

PR #6622, opened 2017-02-15
FIX Array to string conversion message after CSV export

### LINKED ISSUE #6617: Array to string conversion message after CSV export
SilverStripe version: 3.3.1

If you use `Formatting` or `Casting` options on a report column definition, then you're unable to export the data to CSV using the `GridFieldExportButton`.

1. Create a new report that sets the columns to be exported using the `setExportColumns` function of the `GridFieldExportButton` type.
2. Override the `columns`  function and add a column definition that uses the formatting- or casting option.
3. Generate the report and try to export the data using the '`Export to CSV`' button. The file will be generated but won't contain any data. Instead it contains the following error: `Array to string conversion` in `Line` **116** `in` **/framework/forms/gridfield/GridFieldExportButton.php**.

```php
<?php

class TestReport extends SS_Report {

    public function title() {
        return 'Test report';
    }

    public function getReportField() {
        $field = parent::getReportField();

        $gridFieldConfig = $field->getConfig();

        $exportButton = $gridFieldConfig->getComponentByType('GridFieldExportButton');
        $exportButton->setExportColumns($this->columns());

        return $field;
    }

    public function sourceRecords($params = NULL) {
        $q = DB::query('SELECT 1 AS TestField');

        $result = new ArrayList();

        foreach ($q as $item) {
            $temp = new ViewableData();
            $temp->TestField = $item["TestField"];
            $result->push($temp);
        }

        return $result;
    }

    public function columns() {
        $fields = array(
            "TestField" => array(
                'title' => 'Title',
                'formatting' => function($value, $item) {
                    return sprintf('%s', $value);
                }
            )
        );

        return $fields;
    }
}
```

jvdanker:
Fixes #6617 - Rebased.
Added a check where an array column definition is used instead of just a string.

### DIFF
--- a/forms/gridfield/GridFieldExportButton.php
+++ b/forms/gridfield/GridFieldExportButton.php
@@ -111,7 +111,11 @@ class GridFieldExportButton implements GridField_HTMLProvider, GridField_ActionP
 			// determine the CSV headers. If a field is callable (e.g. anonymous function) then use the
 			// source name as the header instead
 			foreach($csvColumns as $columnSource => $columnHeader) {
-				$headers[] = (!is_string($columnHeader) && is_callable($columnHeader)) ? $columnSource : $columnHeader;
+				if (is_array($columnHeader) && array_key_exists('title', $columnHeader)) { 
+					$headers[] = $columnHeader['title'];
+				 } else { 
+					$headers[] = (!is_string($columnHeader) && is_callable($columnHeader)) ? $columnSource : $columnHeader; 
+				}
 			}
 
 			$fileData .= "\"" . implode("\"{$separator}\"", array_values($headers)) . "\"";
--- a/forms/gridfield/GridFieldExportButton.php
+++ b/forms/gridfield/GridFieldExportButton.php
@@ -110,11 +110,12 @@ class GridFieldExportButton implements GridField_HTMLProvider, GridField_ActionP
 
 			// determine the CSV headers. If a field is callable (e.g. anonymous function) then use the
 			// source name as the header instead
+
 			foreach($csvColumns as $columnSource => $columnHeader) {
-				if (is_array($columnHeader) && array_key_exists('title', $columnHeader)) { 
+				if (is_array($columnHeader) && array_key_exists('title', $columnHeader)) {
 					$headers[] = $columnHeader['title'];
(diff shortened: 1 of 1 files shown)


forms/gridfield/GridFieldExportButton.php
```diff
@@ -111,7 +111,11 @@ public function generateExportFileData($gridField) {
 			// determine the CSV headers. If a field is callable (e.g. anonymous function) then use the
 			// source name as the header instead
 			foreach($csvColumns as $columnSource => $columnHeader) {
-				$headers[] = (!is_string($columnHeader) && is_callable($columnHeader)) ? $columnSource : $columnHeader;
+				if (is_array($columnHeader) && array_key_exists('title', $columnHeader)) {
```
dhensby: You have some weird whitespace chars at the end of you lines

forms/gridfield/GridFieldExportButton.php
```diff
@@ -111,7 +111,11 @@ public function generateExportFileData($gridField) {
...

+					$headers[] = $columnHeader['title'];
+				
} else {

+					$headers[] = (!is_string($columnHeader) && is_callable($columnHeader)) ? $columnSource : $columnHeader;
```
dhensby: isn't `is_callable` enough? the `is_string` check is redundant

jvdanker: I don't think it is. `is_callable` will also return `true` if the name is an existing function. I think that's why the original author of this code used `is_string` as well.

forms/gridfield/GridFieldExportButton.php
jvdanker: Weird... Those are unicode line separators (U+2028) but I can't find out why those are being used... I've will remove those and squash the commits onto a new branch.

forms/gridfield/GridFieldExportButton.php
dhensby: ah - good point

forms/gridfield/GridFieldExportButton.php
jvdanker: Those characters broke the travis build as well.. but now it's passing again.

PR #6623, opened 2017-02-15
avoid php warnings if markup is invalid

lekoala:
I had this recently on one of my project. not quite sure how my client managed to generate invalid markup in the first place, but here we go. it's better to ignore parsing errors anyway to handle failure more gracefuly.

### DIFF
--- a/model/HTMLValue.php
+++ b/model/HTMLValue.php
@@ -24,7 +24,11 @@ abstract class SS_HTMLValue extends ViewableData {
 	 * @return string
 	 */
 	public function getContent() {
-		$doc = clone $this->getDocument();
+		$state = libxml_use_internal_errors();
+		// Cloning a document with invalid markup (like duplicated ids) could cause a PHP warning
+        libxml_use_internal_errors(true);
+        $doc = clone $this->getDocument();
+        libxml_use_internal_errors($state);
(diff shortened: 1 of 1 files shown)


dhensby:
This needs to go in 3.5 not 3.5.3

I'm not sure there's value in suppressing errors - this is basically the same as adding `@` to suppress errors.

Unless it's a fatal error, this isn't a big deal and will just be logged (unless you're on a site that displays errors) and then it's probably best to fix the root problem (invalid html) rather than have silent error failures.

lekoala:
I think you are right, it's easier to add a @ in front. Actually, it's the ShortcodeParser that triggers the error, so I can just stick the @ there. I'll make a new pull requests.

Although the warning don't show on my production website, it's still really annoying since I have a local copy in dev mode where one of the pages is not working.

lekoala:
https://github.com/silverstripe/silverstripe-framework/pull/6625

just for reference, the new pull request

dhensby:
FYI - I wasn't suggesting using the error silencing was a good thing - we're working on removing that from our code base wherever we can and a PR to add one won't be accepted.

Whilst I understand it's annoying on the dev site, it may be better to fix the bad HTML

PR #6625, opened 2017-02-15
avoid potential php warning

lekoala:
If markup is invalid, this will trigger a warning. Adding a @ suppress the error.


dhensby:
see https://github.com/silverstripe/silverstripe-framework/pull/6623#issuecomment-280038855

tl;dr error suppression is not an acceptable remedy to this issue

lekoala:
well, I agree, but to a certain extent : I'm not even sure how it was possible to generate duplicated ids within tinymce in the first place. How could you make sure that the client will generate valid markup that is not going to trigger a warning?

but then what I'll do is probably simply run some checks on my page before saving to ensure html is correct.

PR #6628, opened 2017-02-16
Absolute paths for theme lang dirs (fixes #1354)

### LINKED ISSUE #1354: Template translate does not work as expected
What stated in documentation http://doc.silverstripe.org/framework/en/3.1/topics/i18n#usage-in-template-files doesn't work.
- Template file: http://pastie.org/7280012
- # sake /dev/tasks/i18nTextCollectorTask --- Running Task i18n Textcollector Task ---  (no more output)
- themes/mySimple/lang/en.yml is updated like this: http://pastie.org/7280023
- copied it into it.yml: http://pastie.org/7280032
- # sake /dev/build "flush=all"

not working...

chillu:
Regression in 4.x i18n rewrite. Can’t test this since ThemeResourceLoader hardcodes constants like THEME_DIR,
so its impossible to mock in different directory structures. Ideally this would use an application object with access
to environment variables, but that’s too much of a refactor to validate this simple fix.

### DIFF
--- a/src/i18n/Data/Sources.php
+++ b/src/i18n/Data/Sources.php
@@ -91,7 +91,7 @@ class Sources implements Resettable
         }
         $paths = [];
 
-        // Search sorted modules
+        // Search sorted modules (receives absolute paths)
         foreach ($this->getSortedModules() as $module => $path) {
             $langPath = "{$path}/lang/";
             if (is_dir($langPath)) {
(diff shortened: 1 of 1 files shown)


dhensby:
Failures related, hb

src/i18n/Data/Sources.php
```diff
@@ -91,23 +91,23 @@ public function getLangDirs()
...
         $locator = ThemeResourceLoader::instance();
         foreach (SSViewer::get_themes() as $theme) {
             if ($locator->getSet($theme)) {
                 continue;
             }
             $path = $locator->getPath($theme);
-            $langPath = "{$path}/lang/";
-            if (is_dir($langPath)) {
+            $langPath = BASE_PATH . "/{$path}/lang/";
+            if ($langPath) {
```
tractorcow: Er, this condition will always be true. Did you mean is_dir()?

chillu: Oh my, yes. Illustrates the danger of lacking test coverage to directly validate a fix against - although thankfully it seems like other tests using i18n have caught this implicitly. `vendor/bin/phpunit framework/tests/php/i18n/` is passing for me locally now, updated the commit.

Issue #3259, opened 2014-07-03
Update: make UploadField.ss more friendly


rpaget:
framework\templates\UploadField.ss

Currently when the mouse scrolls over "Attach a file", the pointer changes form, but clicking does not pop up the dialog box. The mouse has to scroll over "From your computer", for the dialog window to pop up. This inconsistency can be easily fixed by replacing:

line 40: <label class="ss-uploadfield-item-name">
With: <label class="ss-uploadfield-item-name" for="$id">

pedro2555:
@rpaget Can you confirm this against 3.1?

I'm using Chrome 40, and the pointer does not change when hovering 'Attach a file'.

rpaget:
@pedro2555 I am using 3.1 on firefox, but I also checked chrome. In both cases, the pointer changes to a clickable form (ie a little hand) while hovering over "Attach a file", but clicking does not pop up the dialog box when clicked. The mouse has to scroll over "From your computer", for the dialog window to pop up when the mouse is clicked.

kinglozzer:
“Attach a file” is a label, which apparently can trigger a `cursor: pointer` in some browsers. As the label doesn’t actually point to anything, we just need to add `cursor: default;` to [this selector](https://github.com/silverstripe/silverstripe-framework/blob/292378735251afe221f77c097095c859dc49caf5/scss/UploadField.scss#L71).

PR #6631, opened 2017-02-17
#3259 Make UploadField.ss more friendly

mikaelayoung:
Make cursor default when hovering over the label [#3259](https://github.com/silverstripe/silverstripe-framework/issues/3259)


dhensby:
Thanks a bunch for these UX improvements. Often the things that don't get actioned bit add lots of value to end users

PR #6633, opened 2017-02-17
Fix for bug introduced in 5e19d905

oilee80:
JS selector wrongly scoped to remove element

Introduced in 5e19d905


dhensby:
Thsnks

PR #6636, opened 2017-02-21
Change to use injector for file upload fields in framework

flamerohr:
Created a FieldHandleField interface to use in place of UploadField (which had been pushed to a dependency), and use FileField by default in framework.

Requires:
 - [x] https://github.com/silverstripe/silverstripe-behat-extension/pull/145
 - [ ] https://github.com/silverstripe/silverstripe-cms/pull/1740 (depends on behat pull request)
 - [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/387 (but merge this AFTER the above two, and THIS framework PR)


dhensby:
The CMS behat tests are failing - I guess this is related?

tractorcow:
CMS tests need to be fixed separately. I'll come up with a patch.

What we want is an annotation on cms behat fixtures that disables certain tests when run with asset-admin installed, and then focus those tests (e.g. on upload field) for running in isolation.

tractorcow:
E.g.

```
@assets
@excludeIfModule asset-admin
Feature: Insert an image into a page
  As a cms author
  I want to insert an image into a page
  So that I can insert them into my content efficiently
```

tractorcow:
If @sminnee can approve this, I've opted to remove this feature from the default insert link dialog. The issue is that react components don't work within any jquery-ui dialog due to the way that events are blocked by jquery UI (in a way that cannot be fixed without modifying jquery-ui directly).

The affected CMS tests are removed with https://github.com/silverstripe/silverstripe-cms/pull/1740

tractorcow:
File media dialog is now removed; asset-admin module adds it back in. See https://github.com/silverstripe/silverstripe-asset-admin/pull/387

Note that the CMS test removes the insert-media test fixtures, as these are now both implemented and tested via asset-admin.

tractorcow:
Added fix for behat https://github.com/silverstripe/silverstripe-behat-extension/pull/145

admin/client/src/legacy/TinyMCE_sslink.js
```diff
@@ -0,0 +1,55 @@
...
+(() => {
+  const sslink = {
+    /**
+     * Returns information about the plugin as a name/value array.
+     * The current keys are longname, author, authorurl, infourl and version.
+     *
+     * @returns Name/value array containing information about the plugin.
+     * @type Array
+     */
+    getInfo() {
```
flamerohr: This property is unused and unaccessible, we could remove this

admin/client/src/legacy/TinyMCE_ssmedia.js
```diff
@@ -0,0 +1,159 @@
...
+(() => {
+  const ssmedia = {
+    /**
+     * Returns information about the plugin as a name/value array.
+     * The current keys are longname, author, authorurl, infourl and version.
+     *
+     * @returns Name/value array containing information about the plugin.
+     * @type Array
+     */
+    getInfo() {
```
flamerohr: This is also unused and inaccessible, could remove it

webpack.config.js
```diff
@@ -45,7 +45,8 @@ const config = [
       leaktools: `${PATHS.ADMIN_JS_SRC}/legacy/leaktools.js`,
       MemberImportForm: `${PATHS.ADMIN_JS_SRC}/legacy/MemberImportForm.js`,
       UploadField_select: `${PATHS.ADMIN_JS_SRC}/legacy/UploadField_select.js`,
-      TinyMCE_SSPlugin: `${PATHS.ADMIN_JS_SRC}/legacy/TinyMCE_SSPlugin.js`,
+      TinyMCE_ssmedia: `${PATHS.ADMIN_JS_SRC}/legacy/TinyMCE_ssmedia.js`,
```
flamerohr: we could move this to the asset-admin repo instead right?

tractorcow:
Feedback addressed; moved the plugin to asset-admin module.

flamerohr:
@tractorcow 's changes to my pull request look good, I think this is ready for merging

PR #6638, opened 2017-02-21
Fixes for issues introduced in 5e19d905

oilee80:
As the "onunmatch" is triggered AFTER the element is removed
  from the DOM it never finds the `closest('div')` when switching
  screens. E.g. when using GridField edit
  So when you go from the GridField edit screen back to the page
  you end up with the page loaded but the spinner doesn't disappear
Other option would be to do `$('body').find('.cms-content-loading-overlay,.cms-content-loading-spinner').remove()`
  but this is probably a bit drastic


oilee80:
Thanks @dhensby 👍

Issue #6477, opened 2017-01-11
Switch to micmania1's config implementation


sminnee:
## Acceptance criteria

 * Config uses silverstripe/config module
 * Hard-coded configs can be easily built as a wrapper around an array
 * Config performance is faster

Ingo to talk to Michael about this.

PRS:

 - [x] https://github.com/silverstripe/silverstripe-framework/pull/6641
 - [x] https://github.com/silverstripe/silverstripe-config/pull/4
 - [x] https://github.com/silverstripe/silverstripe-cms/pull/1739
 - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/390

chillu:
@micmania1 FYI, Sam has assigned this to 4.0.0-alpha5 at the moment, which is planned for around a month from today. Are there any documented blockers or TODOs? Existing discussion: https://groups.google.com/forum/#!searchin/silverstripe-dev/config%7Csort:relevance/silverstripe-dev/q5khashNiuY/0Q7zkKwwAwAJ

micmania1:
@chillu The new implementation relies on PSR-6 cache interface which I believe is still up for discussion?

tractorcow:
I'm going to keep a list of todos based on my review (as I go along):

# Caching

We can probably formalise this when we substitute the current cache system.

# Reproduce lookup options (e.g. inheritance, first_set)

We could probably move the lookup behaviour to a level above the core config (which can be pure key-value pairs). E.g. Some config lookup helper that applies class-hierarchy and applied extensions on top of the config.
 
That way these extra functions aren't a part of the config itself, they are just extra tools the framework uses to lookup config.
 
# Support nested config collections
 
I suggest flipping the class construction around to something similar to what @sminnee 
 proposed in the above discussion.
 
 ```
 new CachedConfig(
    new CascadedConfig([
      new YamlConfig(...),
      new PrivateStatticConfig(...),
    ]),
    new PSR6CacheHandler(...)
  );
 ```
 
This would make ConfigCollections responsible for invoking ->transform() on their config sources as-needed.


# Cacheable user-modified config

Just brainstorming some ideas to increase performance where user-modified code provides updated config.

I see these as a great substitute for having to invoke Config::inst()->set(). I imagine a `UserModifiedConfig` store which can be provided a list of callbacks which are conditionally invoked when the cache is regenerated. Alternatively, the non-cached ConfigCollection could just store these on ->set() and invoke the callback on ->get().

[...]

Issue #6252, opened 2016-10-31
SS4 caching: update Zend_Cache or PSR-6?


chillu:
*UPDATE: Ticket description has been extended/replaced by @chillu on 10th of Feb*

## Tasks

 * [x] Remove Zend_Cache dep
 * [x] Use DI for cache configuration
 * [x] Create factory for retrieving caches
 * [x] Fix tests
 * [x] Update documentation
 * [x] Summarise changes in changelog
 * [x] Create ticket for ManifestCache rewrite (should use same cache)
 * [x] Create ticket with symfony/cache about filesystem cache garbage collection (https://github.com/symfony/symfony/issues/21764)
 * [x] Investigate using PhpFile adapter rather than Filesystem (or use [createSystemCache](https://github.com/symfony/cache/blob/master/Adapter/AbstractAdapter.php#L76))
 * [x] Test PHP7 PhpFileCache operation (is 64M enough cache memory by default?)
 * [ ] Resolve conflicts with https://github.com/silverstripe/silverstripe-framework/pull/6641 (if that one is merged first)

## Goals

 * Replace outdated Zend_Cache implementation
 * Use a standards-based implementation ([PSR-6](http://www.php-fig.org/psr/psr-6/) or [PSR-16](http://www.php-fig.org/psr/psr-16/))
 * Minimise upgrade pain (changes to non-core cache uses)
 * Retain or improve cache access and write performance

## Scope

 * Required
    * PSR-6/PSR-16: Store data in a cache pools
    * PSR-6/PSR-16: Store and retrieve common data types via cache keys
    * PSR-6/PSR-16: Define cache expiry on each item
    * PSR-6/PSR-16: Batched cache writes
    * PSR-6/PSR-16: Hit/miss detection
    * PSR-6/PSR-16: Invalidate all caches in a pool
    * Define named cache pools
    * Configure cache drivers (e.g. connection details)
    * Support different drivers
    * Configure through YAML (custom code can modify core caching behaviour)
    * Use for SilverStripe manifests (prior to YAML config being available)
 * Desireable
    * Configure cache drivers per cache pool
    * Cascading cache drivers
    * [Stampede Protection](http://www.stashphp.com/Invalidation.html#stampede-protection)
    * [Cache Key Grouping](http://www.stashphp.com/Grouping.html)
    * Support for at least one driver which can be shared between distributed webservers (e.g. memcache)
    * PSR-6: Deferred cache writes (separate commit phase)
[...]

PR #6641, opened 2017-02-22
API Substitute silverstripe/config module for core config

tractorcow:
Implements https://github.com/silverstripe/silverstripe-framework/issues/6477

Config module work available for review at https://github.com/silverstripe/silverstripe-config/pull/4

Note that this change includes a few other api enhancements and changes designed to support the updated API. This includes a new module registration system (allowing modules to be referenced by full composer name), and changes to Director.

Non-trivial documentation updates will follow.

other PR:

 - [x] https://github.com/silverstripe/silverstripe-testsession/pull/45
 - [x] https://github.com/silverstripe/silverstripe-config/pull/4
 - [ ] https://github.com/silverstripe/silverstripe-graphql/pull/81
 - [ ] https://github.com/silverstripe/silverstripe-cms/pull/1739
 - [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/390
 - [ ] https://github.com/silverstripe/silverstripe-config/pull/5 (bugfix for infinity loop)
 - [ ] https://github.com/silverstripe/silverstripe-behat-extension/pull/148 (bugfix for missing `.env` handling)


src/Core/Config/CoreConfigCreator.php
```diff
@@ -0,0 +1,171 @@
...
+use SilverStripe\Core\Config\Middleware\InheritanceMiddleware;
+use SilverStripe\Core\Manifest\ClassLoader;
+use SilverStripe\Core\Manifest\ModuleLoader;
+use Symfony\Component\Cache\Adapter\FilesystemAdapter;
+use Symfony\Component\Finder\Finder;
+
+/**
+ * Factory for silverstripe configs
+ */
+class CoreConfigCreator
```
chillu: What's the difference between a "factory" and a "creator"? We seem to use both terms in core

tractorcow: Synonyms.

src/Core/Config/CoreConfigCreator.php
```diff
@@ -0,0 +1,171 @@
...
+     *
+     * @param bool $flush
+     * @return CachedConfigCollection
+     */
+    public function createRoot($flush)
+    {
+        $instance = new CachedConfigCollection();
+
+        // Set root cache
+        $instance->setPool(new FilesystemAdapter('configcache', 0, getTempFolder()));
```
chillu: Is there a plan for how to make this more pluggable? I guess it'll be easier once we create an `App` object as part of `main.php` bootstrap, where you could manually set services before relying on DI?

micmania1: Problem with that is the config is already built before it gets to the users code. You'd need to restructure how SS is bootstrapped.

micmania1: There's a little explanation and discussion here: https://github.com/silverstripe/silverstripe-framework/issues/6252#issuecomment-273914525

tractorcow: We do setup `.env` prior to this stage, so it's possible we could use non-yml/statics to configure this. maybe use `getenv`?

micmania1: The problem is that different backends have different constructors. eg. filesystem has none, redis has url, username, password etc.

sminnee: I recommend that we park config-cache-pluggability until after we have an app object.

src/Core/Config/CoreConfigCreator.php
sminnee: Preference is to say "Factory"

src/Core/Config/Middleware/ExtensionMiddleware.php
```diff
@@ -0,0 +1,84 @@
...
+{
+    use MiddlewareCommon;
+
+    /**
+     * Get config for a class
+     *
+     * @param string $class
+     * @param mixed $options
+     * @param callable $next
+     * @return string
```
sminnee: Return string|array|boolean|numeric I assume?

sminnee:
@chillu Once https://github.com/silverstripe/silverstripe-config/pull/4 is merged, can you nudge the tests on this and assuming they pass merge it on Monday?

tractorcow:
@sminnee I'll do a rebase and do the factory class rename at the same time.

tractorcow:
Rebased and renamed that class.

tractorcow:
Oops, rebase didn't remove Zend_Cache from new ModuleManifest. Fixed now. :D

tractorcow:
Need a merge on https://github.com/silverstripe/silverstripe-config/pull/4 then we can restart this build.

tractorcow:
Restarting with fix applied to testsession module

tractorcow:
Please note that I've added a few "experimental" commits that should NOT be merged into the final PR. I'll remove these once I get tests green.

tractorcow:
Added a new fix for https://github.com/silverstripe/silverstripe-config/pull/5

tractorcow:
I've added two new bugfix PRs (see first comment) and pushed another change to this PR that should address certain issues.

sminnee:
I've merged config#5 and re-triggered one of the builds on this PR to confirm that it doesn't break anything: https://travis-ci.org/silverstripe/silverstripe-framework/jobs/205966670#L210

sminnee:
> Please note that I've added a few "experimental" commits that should NOT be merged into the final PR. I'll remove these once I get tests green.

@tractorcow can you confirm that these have now been removed and you're in principle happy to see this merged?

tractorcow:
Yes those are removed. :)

PR #6648, opened 2017-02-23
PSR-16 cache

chillu:
See #6252. Merge with https://github.com/silverstripe/silverstripe-cms/pull/1741


src/Core/Cache/DefaultCacheFactory.php
```diff
@@ -0,0 +1,79 @@
...
+        if (null === $apcuSupported) {
+            $apcuSupported = ApcuAdapter::isSupported();
+        }
+
+        if (!$apcuSupported && null === $phpFilesSupported) {
+            $phpFilesSupported = PhpFilesAdapter::isSupported();
+        }
+
+        if ($phpFilesSupported) {
+            $opcache = Injector::inst()->get(PhpFilesCache::class, false, [$namespace, $defaultLifetime, $directory]);
```
tractorcow: Factory constructor needs to use ->create() internally, not ->get() (with args). This is because items are keyed by class only.

The named services themselves will only invoke this factory when an item needs creating, so it's still a single service per top level key.

src/Core/Cache/CacheFactory.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace SilverStripe\Core\Cache;
+
+use Psr\SimpleCache\CacheInterface;
+
+interface CacheFactory
```
tractorcow: We don't need a `CacheFactory`. Just use `SilverStripe\Core\Injector\Factory` directly.

tractorcow: Sorry, I revoke that. CacheFactory just needs to extend Factory :) Having `CacheFactory` as a service name is useful.

src/View/SSTemplateParser.php
```diff
@@ -4020,10 +4020,10 @@ function CacheBlock_CacheBlockTemplate(&$res, $sub)
         // Get any condition
         $condition = isset($res['condition']) ? $res['condition'] : '';
```
tractorcow: These changes need to be made in SSTemplateParser.peg as well, otherwise it'll get replaced later.

composer.json
```diff
@@ -23,6 +23,7 @@
     "symfony/yaml": "~2.7",
     "embed/embed": "^2.6",
     "swiftmailer/swiftmailer": "~5.4",
+    "symfony/cache": "3.3.x-dev",
```
tractorcow: Can we change to `^3.3@dev` which will still match the dev branch, but also allows stable to be installed once a tag is available.

tractorcow:
Review is done. Please address the above + wait until green, then I approve the merge of this PR.

src/Core/Cache/DefaultCacheFactory.php
chillu: Great catch, fixed now. I've left the PHPDoc for `create()` in `interface CacheFactory extends InjectorFactory`, to clarify the point above.

src/Core/Cache/CacheFactory.php
chillu: Done

src/View/SSTemplateParser.php
chillu: Good point - maybe I do need to start looking for occurrences in `*.*` rather than `*.php` for these cases...

composer.json
chillu: Done

chillu:
OK, I've tried this with PHP7.1 and opcache enabled. On a base install, I've navigated around each CMS section and performed some basic user actions to fill the cache. I get the following stats of `opcache_get_stats()`:

```
array (
  'opcache_enabled' => true,
  'cache_full' => false,
  'restart_pending' => false,
  'restart_in_progress' => false,
  'memory_usage' => 
  array (
    'used_memory' => 23541384,
    'free_memory' => 43546832,
    'wasted_memory' => 20648,
    'current_wasted_percentage' => 0.03076791763305664,
  ),
  'interned_strings_usage' => 
  array (
    'buffer_size' => 4194304,
    'used_memory' => 2035240,
    'free_memory' => 2159064,
    'number_of_strings' => 39877,
  ),
  'opcache_statistics' => 
  array (
    'num_cached_scripts' => 693,
    'num_cached_keys' => 1315,
    'max_cached_keys' => 3907,
    'hits' => 10542,
    'start_time' => 1488058755,
    'last_restart_time' => 0,
    'oom_restarts' => 0,
    'hash_restarts' => 0,
    'manual_restarts' => 0,
    'misses' => 959,
    'blacklist_misses' => 0,
    'blacklist_miss_ratio' => 0.0,
    'opcache_hit_rate' => 91.661594643944,
  ),
```

In its default configuration of 64M, the bytecode and custom caches in core fill up 30% of the available memory. I've noted that in the caching documentation, recommending to increase memory or enable the filesystem fallback for larger apps.

I've also benchmarked locally a bit (homepage with default theme), with `opcache.enabled=1` I get 170ms mean response time, without I get 245ms mean response time. Without opcache means it's using filesystem cache (since I don't have APC installed).

chillu:
I've written a ticket on symfony/cache filesystem purge: https://github.com/symfony/symfony/issues/21764

sminnee:
OK this is passing (except for CMS) and it looks like @tractorcow's comments have been addressed. I'm merging!

tractorcow:
Excellent work @chillu . I just sat down to review and it'd been merged. :)

PR #6650, opened 2017-02-25
Small translation misprint fix.

flamytwista:


tractorcow:
I've fixed this in transifex upstream. https://www.transifex.com/silverstripe/silverstripe-framework/translate/#ru/master/15329145?key=Form.FIELDISREQUIRED

PR #6654, opened 2017-02-26
Prevent unsaved changes alert when clicking a save button

flamerohr:
The "Are you sure you want to leave?" alert shows up for ModelAdmin under certain conditions:
* It is a create form
* for me, it only happens when I have Html field
* Click save


tractorcow:
Restarting tests since we merged some CMS changes. Should be passing now.

PR #6655, opened 2017-02-27
DOCS Swap legacy code block syntax for GFM in "Requirements"

robbieaverill:
This page (currently) for example breaks the code block whenever you mix spaces and tabs for indentation (which is currently part of the 3.x style guideline) - it makes updating existing code examples in the docs quite difficult: https://docs.silverstripe.org/en/3/developer_guides/templates/requirements/

GFM isn't as picky - this PR replaces with it.

### DIFF
--- a/docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
+++ b/docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
@@ -11,10 +11,11 @@ coding any references in the `<head>` tag of your template, as it enables a more
 
 **mysite/templates/Page.ss**
 
-	:::ss
-	<% require css("cms/css/TreeSelector.css") %>
-	<% require themedCSS("TreeSelector") %>
-	<% require javascript("cms/javascript/LeftAndMain.js") %>
+```
+<% require css("cms/css/TreeSelector.css") %>
+<% require themedCSS("TreeSelector") %>
+<% require javascript("cms/javascript/LeftAndMain.js") %>
+```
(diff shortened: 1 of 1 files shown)


chillu:
So we support Github Flavoured Markdown in docsviewer now? It looks like we still use https://github.com/erusev/parsedown-extra, which doesn't explicitly declare support for this syntax.

robbieaverill:
We have been moving towards the GFM syntax for a while. I haven't seen any problems with it on parsedown so far, but docsviewer has recently changed to using the GitHub markdown API - not sure if it has been deployed yet though

robbieaverill:
Sorry- I'm thinking of the addons site.

robbieaverill:
@chillu here's the build context:

* Docsviewer has [a custom parser](https://github.com/silverstripe/silverstripe-docsviewer/blob/master/code/DocumentationParser.php#L86-L95) which converts `:::*` to backticks
* There are cases where `:::` isn't removed from the output, e.g. [here](https://docs.silverstripe.org/en/4/changelogs/4.0.0/#upgrade-code-that-uses-uploadfield)
* Triple backtick format in GitHub Flavour is [supported by erusev/parsedown](https://github.com/erusev/parsedown#features) which we use to parse markdown in docsviewer
* Use of GFM code blocks [renders correctly](https://docs.silverstripe.org/en/4/changelogs/4.0.0/#explicit-text-casting-is-now-enforced-on-all-template-variables)

PR #6658, opened 2017-02-27
FIX Confirm delete actions in React GridFields (e.g. Campaign admin)

### LINKED ISSUE #5398: Ask before deleting a campaign
Acceptance criteria:
- Before deleting a campaign, a `confirm()` checks that you want to delete
- If you click cancel, the campaign is not deleted

---

Can be a simple browser dialog for now

Jira: https://silverstripe.atlassian.net/browse/OSS-1797

robbieaverill:
Fixes #5398

## Note:

This component needs to support localisation - unless there's something magic happening it currently has a couple of messages written in English. I'm unsure how to add localisation at this point, so followed with the previous cases in this file.

This file has a message in it which is specific to campaigns, yet is named broadly as "GridField" - something to keep in mind when using it for the next record type :)

I've disabled the `no-alert` eslint rule for the line I'm using the confirm - obviously a modal would be nicer, but for now a confirm works but throws a lint warning.


flamerohr:
Thanks @robbieaverill!
The translation files are in `framework/admin/client/lang/src` and `framework/admin/client/lang`, self-explanatory once you see inside and needs both respective files to be changed.

Also, please remove the `return true` at the bottom, event handlers do not/should not need a return value :)

robbieaverill:
@flamerohr the return true was for the eslint rule for "consistent return value"

flamerohr:
@robbieaverill if you use `return;` in the condition, I believe eslint will no longer complain

robbieaverill:
@flamerohr I've updated to include localised strings where they were present in the file. I've also left the confirm eslint warning in place, so we don't forget to replace it later on :)

flamerohr:
thanks @robbieaverill, this looks good to merge :D

I've added a line for disabling the eslint, we've got a few confirms/alerts around so if we had left all of them, it would be distracting to errors that may appear

PR #6666, opened 2017-02-28
BUG Prevent obsolete class cache breaking autoload

tractorcow:
Remove hard-coded class paths. These are no longer necessary since we moved the source file to a `.peg` extension, and the class file finder only looks at `.php` files.

### DIFF
--- a/src/Core/Manifest/ClassManifest.php
+++ b/src/Core/Manifest/ClassManifest.php
@@ -206,12 +206,14 @@ class ClassManifest
     {
         $name = strtolower($name);
 
-        if (isset($this->classes[$name])) {
-            return $this->classes[$name];
-        } elseif (isset($this->interfaces[$name])) {
-            return $this->interfaces[$name];
-        } elseif (isset($this->traits[$name])) {
-            return $this->traits[$name];
+        foreach ([
+            $this->classes,
+            $this->interfaces,
+            $this->traits
+        ] as $source) {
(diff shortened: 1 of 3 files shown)


chillu:
Doesn't fix the issue, talking to Damian

tractorcow:
I've updated this fix. Can confirm that with this + https://github.com/silverstripe/silverstripe-config/pull/6, upgrading from alpha2 -> alpha5 works.

tractorcow:
Just to confirm that I get no change in memory usage after this change. The "performance issues" don't seem to be real. :D

src/Core/Manifest/ClassManifest.php
```diff
@@ -382,15 +384,6 @@ public function getOwnerModule($class)
     }
 
     /**
-     * Used to set up files that we want to exclude from parsing for performance reasons.
```
chillu: Damian confirmed that this is no longer the case (memory usage on flush doesn't go up). I'm unsure why that would be, since we haven't really fixed the underlying issue (tokenizer running over a class with heaps of cyclomatic complexity). It might be that the tokenizer in PHP 5.6 is more memory efficient.

tractorcow: @dhensby wrote an improved class parser recently that discarded class-body. This is probably the reason. :D

dhensby: It's not been merged yet :(

Issue #6476, opened 2017-01-11
BUG Default UploadField styling / edit form is broken


tractorcow:
# Tasks

* [x] Create a `FileHandleField` interface, so that asset-admin can utilise DI.
* [x] Replace places where UploadField is used with the FileSelector interface in framework.

# Background

Due to changes we have made to asset management and form styling, the UploadField (framework not asset-admin) has had both serious functionality regressions as well as styling deterioration.

![image](https://cloud.githubusercontent.com/assets/936064/21835677/d3d72c44-d823-11e6-8a1a-0ef5697630f3.png)

![image](https://cloud.githubusercontent.com/assets/936064/21835686/e9ad9d96-d823-11e6-8b1c-abaec9dead5d.png)

chillu:
I propose we don't fix this, but rather remove the legacy UploadField in favour of the new React-powered one: https://github.com/silverstripe/silverstripe-framework/issues/6481. What's the point of keeping old code around when we've written a replacement? Particularly since that'll lock us in to maintaining it until fricken 2020.

tractorcow:
Kill it!

flamerohr:
I've only replaced the UploadFields used in framework with these pull requests:
https://github.com/silverstripe/silverstripe-framework/pull/6636
https://github.com/silverstripe/silverstripe-asset-admin/pull/387

I've left the removing to the other very related issue.

tractorcow:
All merged.

PR #6667, opened 2017-03-01
Remove UploadField, AssetField and associated files

flamerohr:
A bit of clean up following https://github.com/silverstripe/silverstripe-framework/issues/6476

Which removes the now no longer useable UploadField


tractorcow:
We need an upgrading uploadfield section in the 4.0.0.md upgrading notes.

Why did the silverstripe font get removed? It's used in places other than uploadfield.

PR #6668, opened 2017-03-01
Updating title attribute on ModelAdmin_Content template

timkung:
Updating title attribute on managed model tab list links to the actual title of each model


kinglozzer:
This has broken some ModelAdmin styling. Before:
screen shot 2017-04-05 at 14 56 33

After:
screen shot 2017-04-05 at 14 56 47

Very easily missed - jQuery UI’s tabs component has some strange behaviour where it relies on the `title` attribute for the link, and the `id` attribute for the tab matching up. Thankfully there seems to be a workaround, I’ll PR it 😄

PR #6669, opened 2017-03-01
Updating GridField.scss to fix sort icon direction

timkung:
Switching the ascending and descending sprite background positioning for the grid field sort icon to be correct


dhensby:
thanks

PR #6671, opened 2017-03-02
BUG Mark $instance as internal to prevent being saved to config

tractorcow:
Merge with https://github.com/silverstripe/silverstripe-config/pull/7

### DIFF
--- a/src/Core/Config/ConfigLoader.php
+++ b/src/Core/Config/ConfigLoader.php
@@ -10,6 +10,7 @@ use SilverStripe\Config\Collections\ConfigCollectionInterface;
 class ConfigLoader
 {
     /**
+     * @internal
      * @var self
      */
     private static $instance;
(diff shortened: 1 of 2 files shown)


chillu:
If we decide that's the way to go, it'll need docs. I'm not convinced that coupling private statics to config keys is the best long-term approach for SilverStripe, but the alternative (independant config key structures) would necessitate some form of schema declaration to keep config manageable (see [symfony/config](http://symfony.com/doc/current/components/config/definition.html#adding-node-definitions-to-the-tree)). And it would either water down or change the way we define stuff like `DataObject::$db`. So a bit of a bigger discussion than post-alpha5...

tractorcow:
Sorry, I should have opened this more to discussion before merge.

However, I'm happy to discuss and revert / replace with another solution, if you like @chillu :D

chillu:
It's fine for now, relatively low impact change that fixes a high impact regression :)

PR #6672, opened 2017-03-02
FIX/apcu cache errors

andrewandante:
Fixes error:

`Fatal error: Uncaught exception 'Symfony\Component\Cache\Exception\InvalidArgumentException' with message 'Cache key must be string, "boolean" given`

when using apcu.


tractorcow:
Oh, args were correct for ->get(), but we changed those to ->create(). Good catch.

PR #6674, opened 2017-03-03
Fix for APCu cache otherwise you get 'Cache key must be string, "bool…

oilee80:
…ean" given'

Also corrects parameter variable name

### DIFF
--- a/src/Core/Cache/DefaultCacheFactory.php
+++ b/src/Core/Cache/DefaultCacheFactory.php
@@ -44,7 +44,7 @@ class DefaultCacheFactory implements CacheFactory
     /**
      * @inheritdoc
      */
-    public function create($service, array $params = array())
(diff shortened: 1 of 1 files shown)


dhensby:
thanks

tungdt-90:
Hi,

Can you merge this commit for Silverstripe 4.0.0-alpha5 ?

I'm testing the Alpha version, then I got this issue. And this commit fixed it.

dhensby:
@tungdt-90 once can't merge fixes into tags as tags are static and don't change.

If you want this fix you'll need to run 4.0.x-dev or wait for the next release.

tungdt-90:
@dhensby Thanks for your answer, actually i'm not really in such a hurry, so i guess i can wait for a proper release.

But at least, i can confirm that this fix works for SS 4.

Issue #6022, opened 2016-09-16
Switch to node 6 & yarn


sminnee:
## Acceptance criteria

 * NodeJS asset builds for admin, cms, asset-admin, siteconfig, graph-dev tools are updated to work with and/or rely on Node 6 and Yarn
 * Dev docs are updated to indicate this

**Really** nice to have:

 * If you try and use npm or node 4 you'll get a helpful error message

----

## Pull requests

 - [x] https://github.com/silverstripe/silverstripe-framework/pull/6679
 - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/398
 - [x] https://github.com/silverstripe/silverstripe-cms/pull/1754
 - [x] https://github.com/silverstripe/silverstripe-graphql-devtools/pull/4

---

## Background

Looking at https://github.com/nodejs/LTS#lts_schedule, it seems that node 4 is going to go out of support before SilverStripe 4.

At some stage before 4.0-beta1 I think we'll want to switch to node 6 (which I believe requires npm 3) as the recommended build toolchain.

npm 3 will mean that uber's shrinkwrap will need to be dropped. My understanding is that the worst of the shrinkwrap bugs have been resolved, but YMMV.

sminnee:
Incidentally, this is a test case for triggering a lot of the potential shrinkwrap bugs:

```
npm install
npm shrinkwrap
rm -rf node_modules
npm install
npm shrinkwrap
```

If we do try to upgrade to npm3, we'd want to test that this works.

sminnee:
NOTE: We may be better placed to switch to yarn instead, as it's shrinkwrap-equivalent is a core feature (like composer) rather than an afterthought.

It would require a test but it's worth investigating.

flamerohr:
Also would need to investigate migrating to Webpack 2 (or Gulp 4) for the same reason, support isn't likely to be around when SilverStripe 4 is out.

Also other benefits/features that would help our releases, such as reduced file sizes via tree shaking.

Would that be part of this issue, or a separate one?

sminnee:
I'd be deeply concerned about a webpack -> gulp 4 migration.

sminnee:
I think that a webpack 2 w/ tree-shaking upgrade is def worth considering but i'd make it a separate issue against the same epic.

chillu:
Created a separate issue for Webpack 2: https://github.com/silverstripe/silverstripe-framework/issues/6478

PR #6679, opened 2017-03-06
Upgrade to Node6 and yarn

flamerohr:
Please see for more details: 
https://github.com/silverstripe/silverstripe-framework/issues/6022

Upgraded `package.json` to require node6 as a minimum.
`yarn.lock` added to replace `npm-shrinkwrap.json`.
As a result of the upgrade, jest unit tests have been updated to a version which works with node6, this also required some code refactoring and removal of some unit-test-specific-hacks.

Related pull requests
https://github.com/silverstripe/silverstripe-asset-admin/pull/398

Not dependent pull requests
https://github.com/silverstripe/silverstripe-cms/pull/1754
https://github.com/silverstripe/silverstripe-graphql-devtools/pull/4


docs/en/05_Contributing/02_Build_Tooling.md
```diff
@@ -21,14 +21,17 @@ Since we're compiling SVG icons, you'll also need to compile native Node addons,
 which requires `gcc` or a similar compiler - see [node-gyp](https://github.com/nodejs/node-gyp#installation)
 for instructions on how to get a compiler running on your platform.
 
-[npm](https://www.npmjs.com/) is the package manager we use for JavaScript dependencies.
-It comes bundled with NodeJS so should already have it installed if you have Node.
+[yarn](https://yarnpkg.com/) is the package manager we use for JavaScript dependencies.
 The configuration for an npm package goes in `package.json`.
+You'll need to install after [Node.js is installed](https://yarnpkg.com/en/docs/install), we recommend using `npm` which comes with Node.js to install it globally.
```
chillu: Why do you link to yarnpkg.com for a "node.js is installed" link title?

package.json
```diff
@@ -7,13 +7,12 @@
     "test": "tests"
   },
   "engines": {
-    "node": "^4.2.0"
+    "node": "^6.10.0"
```
chillu: Does yarn require 6.10? Don't assume its easy for devs to update minor releases (e.g. when using a SaaS provider which might lag a bit). Can we just say 6.x? Yarn itself claims to work with >4.0.0 (https://github.com/yarnpkg/yarn/blob/0.22-stable/package.json#L79)

.sass-lint.yml
```diff
@@ -4,6 +4,7 @@ files:
   ignore:
     - 'client/src/styles/legacy/*'
     - 'admin/client/src/styles/legacy/*'
+    - 'src/Dev/**/*'
```
chillu: Shouldn't it just ignore all of `src/`, since we don't expect CSS or JS files there?

docs/en/05_Contributing/02_Build_Tooling.md
```diff
@@ -78,6 +81,13 @@ The first `--` separator is required to separate arguments from NPM's own ones.
 $ npm run build -- --watch
 ```
 
+*For development only*:
+Run this to keep webpack automatically rebuilding your file changes, this will also include *.map files for easier debugging.
+It is important to note that this should not be used for pushing up changes, and you should run `npm run build` after you're done.
+```
+$ npm run watch
```
chillu: We should use `yarn run` rather than `npm run` consistently here. I'm assuming those have the same effect

chillu: There's 18 occurrences of "npm run" in framework/docs - please go through these and make them "yarn run"

package.json
```diff
@@ -7,13 +7,13 @@
     "test": "tests"
   },
   "engines": {
-    "node": "^6.10.0"
+    "node": "^6.x"
```
sminnee: Isn't `^6.0.0` better?

.sass-lint.yml
tractorcow: There are js files in `src/Dev/Install/client/src`.

sminnee: Doesn't need to be fixed in this PR, but this is a surprising place to put javascript. `src/` is for PHP.

tractorcow:
I've updated feedback and rebased. Now ready for peer review.

tractorcow:
Some issue with yarn and jquery. Any ideas @flamerohr ? https://travis-ci.org/silverstripe/silverstripe-framework/jobs/209550835

tractorcow:
It looks like the issue is related to https://github.com/yarnpkg/yarn/issues/2629

PR #6684, opened 2017-03-07
FIX Installer no longer causes recursion in yml config

dhensby:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1752

The config.yml file that the `install.php5` file creates is now inline with the same file inside our installer module.


tractorcow:
I think it's about time we add install.php to the CI testing.

PR #6687, opened 2017-03-08
Do not default to locale if hasEmptyDefault is true

hailwood:
Currently there is no per-instance way to say "don't default to the current locale".
If you've explicitly said "This field right here, it has an empty default value" then we should probably respect that.

### DIFF
--- a/forms/CountryDropdownField.php
+++ b/forms/CountryDropdownField.php
@@ -58,16 +58,18 @@ class CountryDropdownField extends DropdownField {
 	public function Field($properties = array()) {
 		$source = $this->getSource();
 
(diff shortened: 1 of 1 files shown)


dhensby:
It looks like the `$source = $this->getSource()` can be inside the condition block too

dhensby:
Also, can we add a test case for this, please?

hailwood:
@dhensby You're right it can, also, do we have any tests at all for the country dropdown field? Wen't to modify them but couldn't find them?

dhensby:
There may well be no tests - don't feel like you need to add coverage for everything just this, please 👼

Issue #6362, opened 2016-12-04
RFC Separate Form Schema / Controller #19


tractorcow:
See https://github.com/silverstripe/silverstripe-framework/issues/6334 for initial story.

# Introduction

Currently the `SilverStripe\Forms\Form` class has a variety of roles; It handles storage of the form schema (fields, layout, etc), handles form and nested field requests, validation, error handling, and rendering.

The basis of this pull request is to separate this into distinct model (i.e. schema) and controller components.

# Background

This RFC is being developed in preparation for some future work planned around abstracting form schema and submission handling in GraphQL. This work necessitates that the current "parent controller" dependency on forms is removed, and instead we move to form scaffolding based on the `FormBuilder` api.

However, the goal of this is to not break or change existing form handling, but to simply segment the current API into separate pieces that can be substituted on a case by case basis.

Note that this RFC explicitly does not dictate how GraphQL based form submission will operate, but instead focuses on any necessary refactor to the current API without significant change in behaviour for traditional forms.

Currently the Form API looks similar to the below.

**Old Form.php**

```php
class Form extends RequestHandler {
	public function __construct($controller, $name, FieldList $fields, FieldList $actions, Validator $validator = null) {}
	
	public function httpSubmission($request) {}
}
```

The issue with this model is that it requires both the controller and the name of the form to be known, meaning that it's impossible to create an abstract from unassigned to any particular parent controller.

# Proposal

This RFC proposes that Form becomes a direct subclass of ViewableData. Instead the request handler may be requested on a case by case basis for a form via `Form::getRequestHandler()` method. In order to ensure that quickly scaffolding up any form remains an easy task, the default behaviour of this method will be to return a response handler which behaves as per standard forms.

Note: It is not necessary for a form to have a response handler in all cases, as external API may in fact  provide their own response handler objects for any given form.

The new API would look similar to the below:

**New Form.php**

```php
class Form extends ViewableData implements HasRequestHandler {
	public function __construct(RequestHandler $parent = null, $name = 'Form', FieldList $fields = null, FieldList $actions = null, Validator $validator = null) {}
	
	/**
	 * @return FormRequestHandler
	 */
	public function getRequestHandler() {
		return Injector::inst()->create(FormRequestHandler::class, $this);
	}
}
```

**New FormRequestHandler.php**

```php
class FormRequestHandler extends RequestHandler {
	public function __construct(Form $form) {}
	
	public function httpSubmission($request) {}
}
```

[...]

Issue #5639, opened 2016-06-01
Fix class to default-route mapping


sminnee:
- Decide what to do about default-routes (do they still make sense?) 
- Ensure that namespaced controllers can have manually-configured routes assigned to them

tractorcow:
Add to this to remove the default Controller::Link() behaviour; I'd prefer it was an abstract method, thus ensuring that only explicit routes exist.

For testing we should disable default route registration, which I've found is still turned on. Testing should be the same as live, where controllers aren't automatically available.

Namespaced controllers can have routes. [x]

sminnee:
Consensus is to drop default URL-to-class mapping; it's an unnecessarily lazy & magical relic of SS2.

tractorcow:
PR at https://github.com/silverstripe/silverstripe-framework/pull/6692

PR #6692, opened 2017-03-09
API Refactor Form request handling into FormRequestHandler

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/5639
Fixes https://github.com/silverstripe/silverstripe-framework/issues/6362

Requires https://github.com/silverstripe/silverstripe-cms/pull/1758

### DIFF
--- a/src/Control/RequestHandler.php
+++ b/src/Control/RequestHandler.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\Control;
 
 use InvalidArgumentException;
+use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Object;
 use SilverStripe\Dev\Debug;
@@ -47,6 +48,13 @@ use BadMethodCallException;
  */
 class RequestHandler extends ViewableData
 {
+    /**
+     * Optional url_segment for this request handler
+     *
+     * @config
+     * @var string|null
+     */
+    private static $url_segment = null;
 
     /**
      * @var HTTPRequest $request The request object that the controller was called with.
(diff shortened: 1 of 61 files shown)


dhensby:
There are some CS failures

src/Control/RequestHandler.php
```diff
@@ -546,4 +547,114 @@ public function setRequest($request)
...
+    {
+        $request = $this->getRequest();
+        if (!$request) {
+            return null;
+        }
+        $pageURL = $request->getHeader('Referer');
+        if (!$pageURL) {
+            return null;
+        }
+        if (Director::is_site_url($pageURL)) {
```
dhensby: Why should the `getReferer` be opinionated about whether the url is a site url or not?

src/Security/Security.php
```diff
@@ -2,17 +2,15 @@
 
 namespace SilverStripe\Security;
 
+use Page;
```
dhensby: `Page` is never used in this class (apart from for a phpdoc type hint)

src/Control/RequestHandler.php
tractorcow: Well, I can call it `getInternalReferer`, but the intent of this method is to get only a referer that's safe to redirect to.

I could add a flag to it to allow external urls to be allowed?

tractorcow:
```
/**
 * Get referrer to redirect back to and safely validates it
 */
```

src/Security/Security.php
tractorcow: I'll change this reference to a class reference.
`private static $page_class = Page::class;`

src/Control/RequestHandler.php
tractorcow: I'll split it into getReferer() (raw) and getReturnReferer() (safely validates)

tractorcow:
Fixed up linting issues, split getReferer() into getReferer() and getReturnReferer(), and replaced a string literal with class constant in Security.php.

src/Control/RequestHandler.php
dhensby: Maybe `getSafeReferer` is all we need as the code to get the referer header is a one liner anyway?

dhensby:
The CMS PR needs fixing before we can merge

tractorcow:
Fixed @dhensby

Issue #6670, opened 2017-03-02
Memory usage regression after Config API merge


chillu:
Loading the default theme homepage took 10MB peak memory on alpha4, and 27MB peak memory on alpha5. Most likely culprit is the new config system.

![image](https://cloud.githubusercontent.com/assets/111025/23490187/b2ffd016-ff5b-11e6-800c-69a3c4d921ce.png)

```
$ php -v
PHP 5.6.29 (cli) (built: Dec 19 2016 23:48:06) 
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2016, by Zend Technologies
    with Xdebug v2.5.0, Copyright (c) 2002-2016, by Derick Rethans
```

```
opcache.blacklist_filename => no value => no value
opcache.consistency_checks => 0 => 0
opcache.dups_fix => Off => Off
opcache.enable => On => On
opcache.enable_cli => Off => Off
opcache.enable_file_override => Off => Off
opcache.error_log => no value => no value
opcache.fast_shutdown => 0 => 0
opcache.file_update_protection => 2 => 2
opcache.force_restart_timeout => 180 => 180
opcache.inherited_hack => On => On
opcache.interned_strings_buffer => 4 => 4
opcache.load_comments => 1 => 1
opcache.log_verbosity_level => 1 => 1
opcache.max_accelerated_files => 2000 => 2000
opcache.max_file_size => 0 => 0
opcache.max_wasted_percentage => 5 => 5
opcache.memory_consumption => 64 => 64
opcache.optimization_level => 0xffffffff => 0xffffffff
opcache.preferred_memory_model => no value => no value
opcache.protect_memory => 0 => 0
opcache.restrict_api => no value => no value
opcache.revalidate_freq => 2 => 2
opcache.revalidate_path => Off => Off
opcache.save_comments => 1 => 1
opcache.use_cwd => On => On
opcache.validate_permission => Off => Off
opcache.validate_root => Off => Off
opcache.validate_timestamps => On => On
```

kinglozzer:
I’ve done a quick bit of digging, I see a tonne of YAML parser/transformer related stuff on every request, instead of just `dev/build` or `?flush`. The `configcache` directory is also empty

kinglozzer:
[...]

PR #6693, opened 2017-03-09
API Apply logging to cache system

tractorcow:
Follow up from https://github.com/silverstripe/silverstripe-framework/issues/6670: Errors were causing performance issues, but were not surfacing.

Actual cache failure was resolved with https://github.com/silverstripe/silverstripe-framework/pull/6671.

This fix ensures that any future cache failures are caught by the logging system.

Note that while we apply the logger to the cache factory (for simplicity), a cache that necessitates a custom logger (or no logger) can be constructed by creating another cache factory, of the same factory class, with no logger assigned.

### DIFF
--- a/src/Core/Cache/DefaultCacheFactory.php
+++ b/src/Core/Cache/DefaultCacheFactory.php
@@ -2,6 +2,9 @@
 
 namespace SilverStripe\Core\Cache;
 
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerInterface;
+use Psr\SimpleCache\CacheInterface;
 use SilverStripe\Core\Injector\Injector;
 use Symfony\Component\Cache\Simple\FilesystemCache;
 use Symfony\Component\Cache\Simple\ApcuCache;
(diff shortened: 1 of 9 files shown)


src/Core/Cache/DefaultCacheFactory.php
```diff
@@ -20,59 +23,101 @@
...
+        $version = isset($args['version']) ? $args['version'] : null;
+
+        // Check support
+        $apcuSupported = $this->isAPCUSupported();
+        $phpFilesSupported = $this->isPHPFilesSupported();
+
+        // If apcu isn't supported, phpfiles is the next best preference
+        if (!$apcuSupported && $phpFilesSupported) {
+            return $this->createCache(PhpFilesCache::class, [$namespace, $defaultLifetime, $directory]);
+        }
```
micmania1: I haven't tested but in PHP 7 you can use PhpArrayAdapter: https://github.com/symfony/cache/blob/master/Adapter/PhpArrayAdapter.php

tractorcow:
This PR might need a rebase. A lot is getting merged today.

tractorcow:
It looks as though the old config was being pulled in; It needed a new tag to ensure it was installed during testing.

tractorcow:
Tests are green, ready to merge.

tractorcow:
Rebased post admin split.

PR #6694, opened 2017-03-10
BUG work-around for form property on formfields not working

tractorcow:
Fix for https://github.com/silverstripe/silverstripe-asset-admin/pull/398


tractorcow:
Sorry, making a call to merge this myself; asset-admin will be broken until we get both modules merged together.

Issue #5758, opened 2016-06-30
Re-code i18n keys to have namespaces


sminnee:
## Acceptance criteria

 * [x] silverstripe-upgrader will replace classnames in strings in _t() calls
 * [x] Existing translations are migrated in transifex

See https://github.com/silverstripe/silverstripe-framework/issues/6726 for work to split the translation keys into the correct module.

----

## PRs

 - [x] https://github.com/silverstripe/silverstripe-admin/pull/38
 - [x] https://github.com/silverstripe/silverstripe-framework/pull/6834
 - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/444
 - [x] https://github.com/silverstripe/silverstripe-cms/pull/1802
 - [x]  https://github.com/silverstripe/silverstripe-campaign-admin/pull/11
 - [x] https://github.com/silverstripe/silverstripe-framework/pull/6852
 - [x] https://github.com/silverstripe/cow/pull/49
 - [x] https://github.com/silverstripe/silverstripe-admin/pull/48

----

## Tasks

 - [x] Remove `.tx` from 3.x branches (all) in preparation for localisation freeze
 
Rest of tasks are 4.x:
 
 - [x] Update locale localisations (4.x) with `tx pull -a -s -f` to ensure we have current localisations from transifex.
 - [x] Run i18nTextCollector on all core modules (4.x) with merge DISABLED `?merge=false` to clear legacy strings.
 - [x] Run [upgrader](https://github.com/silverstripe/silverstripe-upgrader) with `--rule=lang` to update localisations.
 - [x] Check en.yml for un-upgraded strings (e.g. 'Permissions') and add these to the `.upgrade.yml` for the parent module ('Permissions' => 'SilverStripe\Security\Permission'). Then re-run the upgrader again.
 - [x] Manually upgrade all i18n API calls in *.js files (since upgrader can't rewrite these) (Use `ModuleName.ENTITY` convention, e.g. `AssetAdmin.FOLDER_NAME`)
 - [x] Check en.js for un-upgraded strings. Manually upgrade these (more searching and replacing in *.js files).
 - [x] Double check that all localisation files (yml and js, both source and non-en files) have been upgraded, and then install the [transifex client](https://docs.transifex.com/client/introduction). Get auth details from Damian. Push up all files for all modules (one at a time) to transifex with `tx push -s -f && tx push -t -f`. (push localisations, sources, and force).

---
From Ingo's comments on #5734:

[...]

PR #6695, opened 2017-03-10
API Move admin to separate module

tractorcow:
Module has been history pruned and refactored out to https://github.com/silverstripe/silverstripe-admin

 - [ ] CMS PR https://github.com/silverstripe/silverstripe-cms/pull/1760
 - [ ] Installer PR https://github.com/silverstripe/silverstripe-installer/pull/156
 - [ ] Reports https://github.com/silverstripe/silverstripe-reports/pull/62


chillu:
@tractorcow The admin module weighs in at 30.95MB - we need to prune that.

```
git clone https://github.com/silverstripe/silverstripe-admin.git
Cloning into 'silverstripe-admin'...
remote: Counting objects: 20427, done.
remote: Compressing objects: 100% (6493/6493), done.
remote: Total 20427 (delta 13632), reused 20427 (delta 13632), pack-reused 0
Receiving objects: 100% (20427/20427), 30.95 MiB | 1.66 MiB/s, done.
Resolving deltas: 100% (13632/13632), done.
```

chillu:
The actual weight (without history) is ~12MB

```
du -hc -d 1
 32M	./.git
4.0K	./_config
5.4M	./client
236K	./code
104K	./templates
100K	./tests
120K	./themes
6.1M	./thirdparty
 44M	.
 44M	total
```

I wonder if this would be a good opportunity to get some thirdparty deps out of code, and avoid bloating the download. It'll cause every `composer create-project` to download an additional 6.1MB, which could be done a lot more efficiently through pre-bundling into a `dist/` folder. We have to keep tinymce in there, since that's compiled on demand (at least for now).

```
du -hc -d 1 thirdparty/
348K	thirdparty//jquery
556K	thirdparty//jquery-changetracker
 12K	thirdparty//jquery-cookie
760K	thirdparty//jquery-entwine
 40K	thirdparty//jquery-form
 12K	thirdparty//jquery-hoverIntent
 12K	thirdparty//jquery-notice
8.0K	thirdparty//jquery-ondemand
8.0K	thirdparty//jquery-query
724K	thirdparty//jquery-ui
132K	thirdparty//jquery-ui-themes
792K	thirdparty//jstree
2.7M	thirdparty//tinymce
 72K	thirdparty//tinymce-spellchecker
6.1M	thirdparty/
6.1M	total
```

tractorcow:
I trimmed history from the point where we aliased master as 4.0, and trimmed off secondary trees for merge commits. It only reduced the repo size a little.

```
git clone https://github.com/silverstripe/silverstripe-admin.git
Cloning into 'silverstripe-admin'...
remote: Counting objects: 9363, done.
remote: Compressing objects: 100% (3493/3493), done.
remote: Total 9363 (delta 5591), reused 9363 (delta 5591), pack-reused 0
Receiving objects: 100% (9363/9363), 25.42 MiB | 4.90 MiB/s, done.
Resolving deltas: 100% (5591/5591), done.
```

tractorcow:
I've trimmed all the way up until the release of 4.0.0-alpha5 (so we have no duplicates commits between tags). Down to 3.6 mb compressed.

```
git clone git@github.com:silverstripe/silverstripe-admin.git
Cloning into 'silverstripe-admin'...
remote: Counting objects: 1346, done.
remote: Compressing objects: 100% (1064/1064), done.
remote: Total 1346 (delta 231), reused 1346 (delta 231), pack-reused 0
Receiving objects: 100% (1346/1346), 3.60 MiB | 1.43 MiB/s, done.
Resolving deltas: 100% (231/231), done.
Checking connectivity... done.
```

tractorcow:
Note that I've used the below to flatten the silverstripe-admin module history.

http://stackoverflow.com/questions/17988099/how-do-i-trivially-linearize-my-git-history

src/ORM/Hierarchy/Hierarchy.php
```diff
@@ -770,7 +770,7 @@ public function numChildren($cache = true)
      */
     public function showingCMSTree()
     {
-        if (!Controller::has_curr()) {
+        if (!Controller::has_curr() || !class_exists(LeftAndMain::class)) {
```
chillu: Heh I didn't know you could do `class_exists(UnknownClass::class)` without a fatal error, but apparently that works

chillu:
Awesome work, SO CLOSE! :)

Blockers:

 * New module doesn't run NPM tests or linter
 * framework runs linter on `silverstripe-admin` subdirectory (which fails the build)
 * PHPCS tests in framework fail because of a non-existant folder

Can be fixed after merge:

 * All Behat tests need to move over into the new module (we could keep some login tests in framework, but I think that's fairly low value)
 * `/lang` keys aren't untangled (need new transifex module)
 * Module doesn't have a README or LICENSE file
 * Module doesn't expose build status indicators in README (https://travis-ci.org/silverstripe/silverstripe-admin)
 * Framework still tries to run `yarn run build` in CI (even though there's no tests)
 * I wonder if it's worth installing 62MB of node_modules and an entire webpack build toolchain for compiling two SCSS files (./client/src/styles/debug.scss and ./src/Dev/Install/client/src/styles/install.scss). It's probably OK since most contributors will never need to install it (unlike admin, cms and asset-admin, where heavy use of JS and SCSS mandates it)
 * `GridFieldDetailForm_ItemRequest` refers to `LeftAndMain` without `class_exists()`
 * `GridFieldPrintButton` refers to an `admin/` css file
 * TinyMCE refers to assets in `FRAMEWORK_ADMIN_DIR`
 * The `CMSSecurity` class is still in `framework`?
 * `GenericTemplateProvider` still has `'frameworkadmin' => FRAMEWORK_ADMIN_DIR,` in framework
 * Do we need to move `upgrade.yml` definitions over?
 * Module needs to be added to the build dashboard at github-dashing.herokuapp.com

tractorcow:
> All Behat tests need to move over into the new module (we could keep some login tests in framework, but I think that's fairly low value)

Yes, that would be a little bit of work though :) Maybe when behat 3 is a bit closer?

> /lang keys aren't untangled (need new transifex module)

We could make this another story. Maybe when we implement https://github.com/silverstripe/silverstripe-framework/issues/5758?

> I wonder if it's worth installing 62MB of node_modules and an entire webpack build toolchain for compiling two SCSS files (./client/src/styles/debug.scss and ./src/Dev/Install/client/src/styles/install.scss). It's probably OK since most contributors will never need to install it (unlike admin, cms and asset-admin, where heavy use of JS and SCSS mandates it)

I wasn't really sure what would be necessary to leave here; Maybe we could simplify it a bit? I need a bit of front end guidance on this. :)

>  GridFieldDetailForm_ItemRequest refers to LeftAndMain without class_exists()

`instanceof` works even if the class doesn't exist (returns false)

>  GridFieldPrintButton refers to an admin/ css file

Yep needs fixing.

>  TinyMCE refers to assets in FRAMEWORK_ADMIN_DIR

Needs fixing.

>  The CMSSecurity class is still in framework?

Well, it's a security namespace class; It doesn't have a direct dependency on admin module, so it should be safe. I probably need a class_exists around AdminRootController though.

>  GenericTemplateProvider still has 'frameworkadmin' => FRAMEWORK_ADMIN_DIR, in framework

Needs fixing.

>  Do we need to move upgrade.yml definitions over?

Yes.

Question: Should we just move gridfield and tinymce to admin module? Move them under the `SilverStripe\Admin\Forms` namespace?

src/ORM/Hierarchy/Hierarchy.php
tractorcow: also `instanceof NotAClass` works too.

Issue #6647, opened 2017-02-23
Replace ManifestCache with PSR-16


chillu:
Follow up from #6252. These caches were kept separate from Zend_Cache because we needed to instanciate them before the app had a chance to configure anything. The only way to configure a ManifestCache is through an `SS_MANIFESTCACHE` class name. 

I suggest we replace this with standard symfony/cache, and make them configurable by manually setting injector services during a customised boot process (e.g. a custom `main.php`). Sam's got some ideas regarding an `App` object which feeds into this approach.

Roughly speaking, this would be the content of an `index.php` with custom caches:

```php
require_once __DIR__ . '../vendor/autoload.php';

use Psr\SimpleCache\CacheInterface;
use Symfony\Component\Cache\Simple\MemcachedCache;
use SilverStripe\Core\DefaultAppFactory;
use SilverStripe\Control\DefaultRequestFactory;

$app = DefaultAppFactory::create();

// custom
$memcached = new Memcached();
$app->setService(new MemcachedCache(
  $memcached,
  'thememanifest',
), CacheInterface::class . '.thememanifest');
$app->setService(new MemcachedCache(
  $memcached,
  'classmanifest',
), CacheInterface::class . '.classmanifest');
$app->setService(new MemcachedCache(
  $memcached,
  'configmanifest',
), CacheInterface::class . '.configmanifest');

$request = DefaultRequestFactory::create();
$response = $app->getDirector()->direct($request);
$response->output();
```

There's some new bits in here (e.g. `DefaultAppFactory`) which should be solved first, in a separate ticket.

tractorcow:
You could also inject custom code with a `files` composer-included file. This is the same way that framework injects Constants.php (even when running in other apps, e.g. phpunit).

chillu:
Related: https://github.com/silverstripe/silverstripe-framework/issues/3940

tractorcow:
WIP: 
 - https://github.com/open-sausages/silverstripe-framework/commits/pulls/4.0/manifest-cache
 - https://github.com/open-sausages/silverstripe-config/commits/pulls/1.0/simplecache

tractorcow:
PR at https://github.com/silverstripe/silverstripe-framework/pull/6706

PR #6706, opened 2017-03-15
API Build new ManifestCache based on PSR-16 SimpleCache

tractorcow:
Rebased on top of https://github.com/silverstripe/silverstripe-framework/pull/6693

Fixes https://github.com/silverstripe/silverstripe-framework/issues/6647

Merge with:
 - [x] https://github.com/silverstripe/silverstripe-config/pull/8 (requires tag before framework CI will pull in changes)
 - [ ] https://github.com/silverstripe/silverstripe-admin/pull/2
 - [x] https://github.com/silverstripe/silverstripe-cms/pull/1762 
 
 
The main reason for the size of this change is that this change includes additional consolidation of getModules() duplicated between ClassManifest and ModuleManifest. I've remove this from ClassManifest, and added full tests in ModuleManifestTest.

I've updated various other classes, such as i18nTextCollector, to use the new Module manifest, and remove instances of the legacy module registration system.

I've noticed significant performance increases after switching to the optimised ManifestCacheFactory. No official benchmarks as of yet, but it is observably faster. :D


src/Core/Manifest/ClassManifest.php
```diff
@@ -496,4 +509,30 @@ protected function coalesceDescendants($class)
...
+            // Must be set
+            if (!isset($data[$key])) {
+                return false;
+            }
+            // and an array
+            if (!is_array($data[$key])) {
+                return false;
+            }
+            // No numeric keys
+            $numericKeys = array_filter(array_keys($data), 'is_numeric');
```
dhensby: how important is this check? I feel like this isn't a particularly optimal way to test for valid cache items.

We have to visit every array item to pass the check.

src/Logging/Log.php
```diff
@@ -36,8 +36,8 @@ class Log
      */
     public static function get_logger()
     {
-        Deprecation::notice('5.0', 'Use Injector::inst()->get(\'Logger\') instead');
-        return Injector::inst()->get('Logger');
+        Deprecation::notice('5.0', 'Use Injector::inst()->get(LoggerInterface::class) instead');
```
dhensby: I feel like deprecating APIs for 5 before 4 is even released is a bit silly - we should trash them now.

dhensby:
OK - needs rebasing

src/Core/Core.php
```diff
@@ -70,20 +71,26 @@
 $requestURL = isset($_REQUEST['url']) ? trim($_REQUEST['url'], '/') : false;
 $flush = (isset($_GET['flush']) || $requestURL === trim(BASE_URL . '/dev/build', '/'));
 
-global $manifest;
-$manifest = new ClassManifest(BASE_PATH, false, $flush);
+// Manifest cache factory
+$manifestCacheFactory = new ManifestCacheFactory([
+    'namespace' => 'manifestcache',
+    'directory' => getTempFolder(),
```
tractorcow: It's just a default argument; You can ignore it if you are using an in-memory cache (set via SS_MANIFESTCACHE).

src/Core/Manifest/ClassManifest.php
tractorcow: It's important because upgrades from alpha4 were running into issues with numeric keys, for some reason.

src/Logging/Log.php
tractorcow: Sure I can just drop them.

tractorcow: Let's remove the whole class; It has no non-deprecated methods.

robbieaverill:
Ignore that review - it was for 95a48aa only

tractorcow:
Rebased.

src/Core/Manifest/ClassManifest.php
dhensby: We shouldn't have code that covers upgrades from unstable to unstable versions...

tractorcow:
Rebased again.

src/Core/Manifest/ClassManifest.php
dhensby: bump

dhensby:
test failures related?

tractorcow:
No, config module was moved from PSR-6 to PSR-16 cache, so we can't cleanly run CI until everything is merged.

dhensby:
ah - ok

src/Core/Manifest/ClassManifest.php
tractorcow: It's not to cover just upgrades; It validates the cache is valid. It could be invalid for some OTHER reason.

sminnee: This validation check is only called when actually flushing (as it's to do with the look up of the cache meta-data of a single file), and so I think a minor slow-down is okay. Other approaches to this could include versioning the cached meta-data, but I don't think it blocks PR.

sminnee: > We shouldn't have code that covers upgrades from unstable to unstable versions...

Disagree. We caused substantial pain for a lot of ss4 users with this.

tractorcow: ok, we've decided to simplify this to just check the first key to fix upgrades from alpha5.

sminnee: Summary of offline discussion: @tractorcow will update this code to only check the first element of `$data`, which will detect and reject old cache content much more quickly. A comment will be added to note that the purpose of this is to detect out-of-date caches.

src/Core/Manifest/ClassManifest.php
```diff
@@ -496,4 +509,30 @@ protected function coalesceDescendants($class)
...
+        foreach (['classes', 'interfaces', 'traits'] as $key) {
+            // Must be set
+            if (!isset($data[$key])) {
+                return false;
+            }
+            // and an array
+            if (!is_array($data[$key])) {
+                return false;
+            }
+            // Detect legacy cache keys (non-associative)
```
sminnee: Looks good to me!

sminnee:
Merged the config patch and tagged as 1.0.0-alpha7, re-running tests now.

sminnee:
@tractorcow is there a patch to behat-extension too?

sminnee:
```
Fatal error: Call to undefined method SilverStripe\Core\Manifest\ClassManifest::getModules() in
/vendor/silverstripe/behat-extension/src/SilverStripe/BehatExtension/Console/Processor/LocatorProcessor.php
on line 64
```

sminnee:
I'm merging this, as the merge of config broken master installs.

sminnee:
@tractorcow please push a behat PR asap

tractorcow:
PR for behat at https://github.com/silverstripe/silverstripe-behat-extension/pull/149

src/Core/Core.php
chillu: What if you need *additional* args like a Redis/Memcached url and port? I guess that can wait until we have an app object, and a way to create custom `index.php` (where you could define those services prior to their use, and add them to the injection container manually - without relying on YAML configuration.

PR #6708, opened 2017-03-15
Improved spanish translation

filisko:


dhensby:
Thanks for contributing to our translations. However our translations are managed through [Transifex](https://www.transifex.com/silverstripe/). You'll need to sign up for an account and then submit your translations to the relevant module.

Thanks!

PR #6712, opened 2017-03-16
Fix dev CI builds

tractorcow:
Move silverstripe/config release responsibility to installer
Remove admin artifacts in composer.json

Merge with https://github.com/silverstripe/silverstripe-installer/pull/157


dhensby:
it's working!

PR #6718, opened 2017-03-19
Drop PHP 5.5 support, limit PHP 5.6 to 2018

### LINKED ISSUE #6705: Drop PHP 5.5 support in SilverStripe 4.0
PHP 5.5 has been End Of Life (EOL) since July 2016, meaning it no longer receives bugfixes or security fixes. Anybody running it is effectively on insecure software - see http://php.net/eol.php

According to packagist installs as of Nov 2016 PHP 5.5 and lower made up 27.29%. That's rapidly trending downward (-12% in 6 months): https://seld.be/notes/php-versions-stats-2016-2-edition

This is also not an isolated decision for our framework - we're increasingly relying on third party dependencies like Symfony, which are very unlikely to continue development on their releases for unsupported PHP versions. Case in point: We want to upgrade to PHPUnit 5.x, which was released 18 months ago, and dropped PHP 5.5 support back then already: https://github.com/silverstripe/silverstripe-framework/issues/6704

I'm proposing that we drop PHP 5.5 for SilverStripe 4.0, which would make PHP 5.6 the minimum required version.

Please vote with 👍  if you are in favour of dropping PHP 5.5, or 👎  if you want to keep PHP 5.5 support. In this case, please add a comment clarifying why that's important to you.

/cc @silverstripe/core-team

chillu:
Fixes #6705

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: 5.5
+    - php: 5.6
       env: DB=PGSQL PHPUNIT_TEST=1
     - php: 5.6
       env: DB=MYSQL PHPUNIT_TEST=1 PHPCS_TEST=1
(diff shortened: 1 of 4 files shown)


dhensby:
we need to drop the 5.5 travis test too

PR #6720, opened 2017-03-19
Reinstate 7.0 support, clarify policy

chillu:
PLEASE WAIT FOR A VOTE UNTIL MERGE

Vote with 👍  or 👎 

We’ve discussed dropping support for 5.6 once it’s EOL on https://github.com/silverstripe/silverstripe-framework/issues/6705.
The same question needs to be asked for 7.0, which goes EOL in the same month. And should be generalised to future PHP versions.
Follow up to https://github.com/silverstripe/silverstripe-framework/pull/6718/

### DIFF
--- a/docs/en/00_Getting_Started/00_Server_Requirements.md
+++ b/docs/en/00_Getting_Started/00_Server_Requirements.md
@@ -8,8 +8,9 @@ Our web-based [PHP installer](installation/) can check if you meet the requireme
 
 ## Web server software requirements
 
- * PHP 5.6 and PHP 7.1+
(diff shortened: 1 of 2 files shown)


tractorcow:
Note: We are looking for feedback on this PR before we merge. Once we agree we will merge and make this the new standard policy for PHP version support.

sminnee:
@chillu I used the "request review" feature to get the as-yet-unvoted core committers to see this ;)

sminnee:
We have a good quorum, merged :-)

PR #6721, opened 2017-03-20
Remove spaces in elseif statements

flamerohr:
Looking at the note after the first example here: http://php.net/manual/en/control-structures.elseif.php

the `else if` would be using `elseif` now (due to using colons)


dhensby:
I thought I'd caught these - odd - sorry.

PR #6722, opened 2017-03-20
Better HTML generation behaviour for Requirements_Backend

tractorcow:
Because string concatenation in Requirements_Backend is gross.

Merge with https://github.com/silverstripe/silverstripe-cms/pull/1793

### DIFF
--- a/src/Forms/GridField/GridField.php
+++ b/src/Forms/GridField/GridField.php
@@ -19,6 +19,7 @@ use SilverStripe\Forms\FormField;
 use SilverStripe\Forms\Form;
 use LogicException;
 use InvalidArgumentException;
+use SilverStripe\View\HTML;
 
 /**
  * Displays a {@link SS_List} in a grid format.
@@ -466,9 +467,8 @@ class GridField extends FormField
         }
 
         // Display a message when the grid field is empty.
-
         if (empty($content['body'])) {
-            $cell = FormField::create_tag(
+            $cell = HTML::createTag(
                 'td',
                 array(
                     'colspan' => count($columns),
@@ -476,7 +476,7 @@ class GridField extends FormField
                 _t('SilverStripe\\Forms\\GridField\\GridField.NoItemsFound', 'No items found')
             );
 
-            $row = FormField::create_tag(
+            $row = HTML::createTag(
                 'tr',
                 array(
                     'class' => 'ss-gridfield-item ss-gridfield-no-items',
@@ -518,20 +518,20 @@ class GridField extends FormField
         );
 
         if ($this->getDescription()) {
-            $content['after'] .= FormField::create_tag(
+            $content['after'] .= HTML::createTag(
                 'span',
                 array('class' => 'description'),
                 $this->getDescription()
             );
         }
 
-        $table = FormField::create_tag(
+        $table = HTML::createTag(
             'table',
             $tableAttributes,
             $header . "\n" . $footer . "\n" . $body
         );
 
-        return FormField::create_tag(
+        return HTML::createTag(
             'fieldset',
             $fieldsetAttributes,
             $content['before'] . $table . $content['after']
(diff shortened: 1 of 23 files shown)


dhensby:
I don't like that we use `Form` to build HTML - can't we have an HTML builder or something?

tractorcow:
Perhaps something is SilverStripe\View?

dhensby:
> Perhaps something in SilverStripe\View?

or a library if there's a good one?

sminnee:
My recommendation is that we push `create_tag()` to something in view, and have the form system call that instead of `FormField::create_tag()`. Form depending on View is better than the reverse.

tractorcow:
I've moved create_tag to HTMLValue and made it non-static; This means that if we wanted to set a custom HTMLValue subclass, we could customise tag creation. :)

tractorcow:
Note: CMS changes need merging too at https://github.com/silverstripe/silverstripe-cms/pull/1793, since this change is now breaking.

dhensby:
looks much nicer

tractorcow:
Is this mergable if we like the API?

tractorcow:
Rebased.

sminnee:
> I've moved create_tag to HTMLValue and made it non-static

I'm not convinced that the non-static API here is an improvement, as the method is essentially a stateless function.

If we do want to make it an instance method it should be a method of an `HTMLGenerator` service or something — it's not nothing to do with the responsibility of HTMLValue. At best, it a type of HTMLValueFactory.

src/Forms/GridField/GridField.php
```diff
@@ -466,17 +467,17 @@ public function FieldHolder($properties = array())
         }
 
         // Display a message when the grid field is empty.
-
+        $htmlValue = HTMLValue::create();
```
flamerohr: would `$htmlValue = HTMLValue::singleton();` be better?

flamerohr:
Was the concensus to move the `createTag` method to a factory or provide?

dhensby:
I agree with Sam that the method and best as a static or on a service

tractorcow:
Ok, so create new service, and make it static (similar to `Convert::raw2sql`)

tractorcow:
Ok, done @dhensby :D Can you please review and make sure this new approach is better?

src/View/HTML.php
```diff
@@ -0,0 +1,77 @@
...
+    use Configurable;
+
+    /**
+     * List of HTML5 void elements
+     *
+     * @see https://www.w3.org/TR/html51/syntax.html#void-elements
+     * @config
+     * @var array
+     */
+    private static $void_elements = [
```
flamerohr: very nice, extendable by config :D

tests/php/View/HTMLTest.php
```diff
@@ -0,0 +1,45 @@
...
+
+    public function testVoidContentError()
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage("Void element \"link\" cannot have content");
+        $tag = HTML::createTag('link', [
+            'title' => 'HTML & Text',
+            'nullattr' => null,
+        ], 'Some <strong>content!</strong>');
+        $this->assertEquals('<a title="HTML &amp; Text">Some <strong>content!</strong></a>', $tag);
```
flamerohr: I'm confused with why it's asserting `'<a title="HTML &amp; Text">Some <strong>content!</strong></a>'`, does this assert even trigger?
if that's the case, perhaps add a comment above to state that.

Although I've noticed another `expectException()` below and doesn't have any asserts following the code which throws an exception

tractorcow:
There was an issue where empty attributes were being generated where none should be (e.g. `disabled=""`) which broke behat.

flamerohr:
I'll restart tests when the cms part is merged :)

PR #6723, opened 2017-03-20
API Split SilverStripe\Assets into separate module

### LINKED ISSUE #5289: RFC: Create module silverstripe-assets
Create new `silverstripe/silverstripe-assets` module with the asset code. Use git `filter-branch` to preserve a filtered history (https://help.github.com/articles/splitting-a-subfolder-out-into-a-new-repository/) 

(This module can be renamed to `silverstripe/assets` once we are running modules in the vendor folder.)

Include at a minimum the following code:
- File
- Folder
- AssetAdmin
- DBFile
- Asset backend
- UploadField
- Image
- Image manipulation

---

_Original text_

Right now the underlying API for managing files is part of framework, but the UI is separate. This isn't a good long-term approach.

The new SS4 UI is being developed in https://github.com/silverstripe/silverstripe-asset-admin/

I see there as being two major options:
- Move back-end code into https://github.com/silverstripe/silverstripe-asset-admin/ and rename it to https://github.com/silverstripe/silverstripe-assets/
- Create an additional module, https://github.com/silverstripe/silverstripe-assets/, that has no UI

In my view, it would be cleaner to have a single module, but having separate `assets` and `asset-admin` modules has a couple of advantages:
- You can build custom asset UIs without leaving the default UI in your project
- It will be easier to preserve the history of the relevant classes

Although the SS3 file system had no versioning, I think it's okay to make `silverstripe/versioning` a requirement of `silverstripe/assets`, as to make versioning optional will introduce a number of parallel code paths that would be likely to increase the number of bugs.

tractorcow:
See https://github.com/silverstripe/silverstripe-assets

Fixes #5289

Requires https://github.com/silverstripe/silverstripe-installer/pull/158


composer.json
```diff
@@ -23,6 +23,7 @@
     "monolog/monolog": "~1.11",
     "nikic/php-parser": "^2 || ^3",
     "silverstripe/config": "^1@dev",
+    "silverstripe/assets": "^1@dev",
```
sminnee: This should be a requirement of cms, not framework.

sminnee: OK while there's coupling still in place this is probably okay as a first step.

tractorcow: Yeah, there is quite a bit of work to de-couple that at this stage. It may be closer to a 5.0 / 4.late goal. :)

sminnee:
@silverstripe/core-team any objections?

sminnee:
I feel like we've got a sufficient quorum so I'm merging!

PR #6728, opened 2017-03-22
API Use mysql-safe table namespace separator

tractorcow:
The idea behind the original behaviour is that pre-namespaced tables in 3.x would continue to work.

However, namespaced dataobjects were not feasibly in 3.x. They were theoretically possible, but ran into several edge cases which made it impossible in practice. I don't believe it's worth trying to soft-support automatic upgrades for these classes in this case.

As a result of this change, all new namespaced classes (without any table_name) will be buildable in mysql.

Note: Please leave this PR open for a few days for public approval. There are a few opinions that need to be considered on the matter. :)

### DIFF
--- a/src/ORM/DataObjectSchema.php
+++ b/src/ORM/DataObjectSchema.php
@@ -24,12 +24,10 @@ class DataObjectSchema
      * Default separate for table namespaces. Can be set to any string for
      * databases that do not support some characters.
      *
-     * Defaults to \ to to conform to 3.x convention.
-     *
      * @config
      * @var string
      */
-    private static $table_namespace_separator = '\\';
+    private static $table_namespace_separator = '_';
 
     /**
      * Cache of database fields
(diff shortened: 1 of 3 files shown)


sminnee:
Oops I clicked merge before reading that last sentence. But it's been a couple of days and there's only been 👍 s

tractorcow:
Nothing can't be changed after it's merged. :)

dhensby:
Only just seen this ;)

My comment would be that we already use `_` as a separator for `_Live` and `_Versions` (and so on) so this could technically allow table name collisions...

Would a `-` be more suitable as a default? It can't be used in PHP classnames so we're less likely to see a collision...

sminnee:
A table name collision is unlikely and if it happens there's a workaround practise:
 * dev/build throws a fatal error
 * developer puts a table_name setting into yml or private static

So I think it's fine.

sminnee:
I think the main thing is that the fatal error on a table name collision should give people a bit of instructions about the table_name config option.

dhensby:
It might be rare, but it's a possibility we can move much closer to zero by changing one character in our code...

If you're running a site, then change the separator, will a dev/build migrate the tables? I guess not because we couldn't know what it changed from and to... So this could cause a lot of pain for someone that happened to have a name collision with a join table.

sminnee:
hyphen requires escaping in MySQL queries, underscore doesn't. We'd be replacing a problem that hardly ever happens with one that happens no every query.

👎  from me

tractorcow:
Underscore is reasonably safe. We don't assume it is always a table suffix, but we test all candidate table -> class mappings against the list of known tables. The worst case scenario is that someone makes a `Page/Live` class, which will fail to build before it has a chance to silently cause issues.

dhensby:
I'm more concerned about relation conflicts as they are both defined outside the framework and could cause issues.

The fact we hand build DB identifiers rather than use `Convert::symbol2sql` makes using a `-` much harder than it should be 😞

PR #6729, opened 2017-03-22
fix glitch duplicate jstree-icon in nested pages

leonixyz:
Nested pages having their own children are shown in the site tree with a duplicate `jstree-icon`. I have no idea why there are two icons in this template file, however, by removing the first one the glitch doesn't happen.

screen shot 2017-03-22 at 13 39 57


kinglozzer:
https://github.com/silverstripe/silverstripe-framework/pull/6610 was supposed to have resolved this issue, have you tested on version 3.5.3? I can’t recreate it

leonixyz:
No, sorry, I'm still at version 3.5.2

kinglozzer:
Thanks for confirming 👍

PR #6731, opened 2017-03-23
FIX: many_many_extraFields breaks _SortColumn0 ordering (fixes #6730)

### LINKED ISSUE #6730: ManyManyList::appendExtraFieldsToQuery() breaks sort orders that rely on _SortColumn0
Boy was this fun to track down 🙄😅. Discovered this with the blog module - you’re unable to use `$many_many_extraFields` with it (e.g. I wanted sortable “Related news” on pages).

Simplified example that doesn’t rely on blog below. Tested on 3.x, but a glance at 4.x suggests the behaviour around this hasn’t changed.

```php
class Page extends SiteTree
{
    private static $many_many = [
        'TestPages' => 'TestPage'
    ];

    private static $many_many_extraFields = [
        'TestPages' => [
            'Foo' => 'Varchar' // Can be anything, doesn't matter
        ]
    ];
}

class Page_Controller extends ContentController
{
    public function init()
    {
        parent::init();

        foreach ($this->TestPages() as $page) {
            die('winrar');
        }
    }
}

class TestPage extends Page
{
    // Any clause that contains a space or parenthesis
    private static $default_sort = '"Title" IS NOT NULL DESC';
}
```

## What happens

- The query is set up, the `$default_sort` clause is added and uses `_SortColumn0` as an alias if the clause contains spaces or parentheses ([see here](https://github.com/silverstripe/silverstripe-framework/blob/3.5.3/model/queries/SQLSelect.php#L332-L342))
- An extra `SELECT` statement is added (e.g. `SELECT "Title" IS NOT NULL DESC AS _SortColumn0`)
- `ManyManyList::appendExtraFieldsToQuery()` calls `DataQuery::selectFromTable()`, which then wipes the list of fields being selected - meaning the `_SortColumn0` select statement is gone
- Errors happen:

> `Couldn't run query: SELECT [...] Unknown column '_SortColumn0' in 'order clause'`

## Fixing it

Perhaps make `ManyManyList::appendExtraFieldsToQuery()` instead call a new method `DataQuery::addSelectFromTable()`, rather than altering the behaviour of `DataQuery::selectFromTable()`. The naming of the `appendExtraFieldsToQuery()` method suggests that’s what it should be doing, rather than wiping existing selects out and starting again

kinglozzer:
Proposed solution to #6730. The new method `DataQuery:: addSelectFromTable()` is essentially a clone of `DataQuery:: selectFromTable()`, but using `SQLQuery::addSelect` instead of `SQLQuery::setSelect()`. I didn’t want to alter the behaviour of an existing method for a bugfix/patch release.

### DIFF
--- a/model/DataQuery.php
+++ b/model/DataQuery.php
@@ -763,17 +763,33 @@ class DataQuery {
 	}
 
 	/**
-	 * Select the given fields from the given table.
+	 * Select only the given fields from the given table.
 	 *
 	 * @param String $table Unquoted table name (will be escaped automatically)
 	 * @param Array $fields Database column names (will be escaped automatically)
 	 */
 	public function selectFromTable($table, $fields) {
 		$fieldExpressions = array_map(function($item) use($table) {
-			return "\"$table\".\"$item\"";
+			return Convert::symbol2sql("{$table}.{$item}");
 		}, $fields);
 
 		$this->query->setSelect($fieldExpressions);
+
+		return $this;
+	}
+
+	/**
+	 * Add the given fields from the given table to the select statement.
+	 *
+	 * @param String $table Unquoted table name (will be escaped automatically)
+	 * @param Array $fields Database column names (will be escaped automatically)
+	 */
+	public function addSelectFromTable($table, $fields) {
+		$fieldExpressions = array_map(function($item) use($table) {
+			return Convert::symbol2sql("{$table}.{$item}");
+		}, $fields);
+
+		$this->query->addSelect($fieldExpressions);
 
 		return $this;
 	}
--- a/model/ManyManyList.php
+++ b/model/ManyManyList.php
@@ -104,7 +104,7 @@ class ManyManyList extends RelationList {
 			}
 		}
 
-		$this->dataQuery->selectFromTable($this->joinTable, $finalized);
+		$this->dataQuery->addSelectFromTable($this->joinTable, $finalized);
 	}
 
 	/**
@@ -352,7 +352,7 @@ class ManyManyList extends RelationList {
 	 */
 	public function getExtraData($componentName, $itemID) {
 		$result = array();
-		
+
 		// Skip if no extrafields or unsaved record
 		if(empty($this->extraFields) || empty($itemID)) {
 			return $result;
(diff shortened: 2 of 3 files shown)


model/DataQuery.php
```diff
@@ -779,6 +779,22 @@ public function selectFromTable($table, $fields) {
...
 
 	/**
+	 * Add the given fields from the given table to the select statement.
+	 *
+	 * @param String $table Unquoted table name (will be escaped automatically)
+	 * @param Array $fields Database column names (will be escaped automatically)
+	 */
+	public function addSelectFromTable($table, $fields) {
+		$fieldExpressions = array_map(function($item) use($table) {
+			return "\"$table\".\"$item\"";
```
dhensby: Can we start using `Convert::symbol2sql` as a matter of course now, please?

kinglozzer: Sure, I’ll update the existing method that I copied this from too

tractorcow: Can we please not add in new conventions into 3.x? All of our core code is ansi-92 sql standard, and a convention change (especially in a non-major branch) means we've lost all of our consistency.

You're also breaking all augmentSQL rewrites on db connectors that support multiple table quotation styles.

If you want to change this convention, it's worth doing as a part of an RFC and a complete work (rewrite of core) plus collecting feedback on impact of this change.

tractorcow: At the moment symbol2sql seems hard coded to ansi-quotes, so it's probably safe for the time being. I wish I never added this method though; It's made the ORM api a lot more inconsistent.

model/DataQuery.php
```diff
@@ -779,6 +779,22 @@ public function selectFromTable($table, $fields) {
 	}
 
 	/**
+	 * Add the given fields from the given table to the select statement.
+	 *
+	 * @param String $table Unquoted table name (will be escaped automatically)
+	 * @param Array $fields Database column names (will be escaped automatically)
+	 */
+	public function addSelectFromTable($table, $fields) {
```
tractorcow: New public method probably means it needs to be a new minor (feature) version. Can we PR to 3 branche instead of 3.5?

kinglozzer: Done 😄

model/DataQuery.php
dhensby: Ah, sorry - I probably thought this was against master. If we're connecting in ansi mode then `quoteSymbol` should quote in ansi mode otherwise it wouldn't be doing a very good job!

Issue #6243, opened 2016-10-27
RFC: Package resource management


sminnee:
In SilverStripe 3 client assets were typically referenced by their path relative to the `BASE_PATH`. This includes the module folder, which is typically hardcoded.

This has simple approach has served us for a number of years but it makes a number of things difficult:
- Moving the webroot to something other than the project root, which is a common security practise.
- Moving modules into the `vendor/` folder, which would make it easier to work with a larger number of modules, and therefore create smaller, more single-purpose modules.
- Making use of modules without carefully setting up a project tree. The "simpler test run" branches of framework and cms hit this, and overcame them by refining the use of constants like `CMS_DIR`. However it's awkward to scale this out to every module.

We'd like to improve on this.

## Recommendation

I recommend that we introduce a "resource locator" as an injected service. The service would be responsible for handling both server-accessed and client-accessed resources.

A mock-up of the relevant interfaces are here: https://gist.github.com/sminnee/92d5486ba9d23d12ef4253b5ef274bf7

The Injector would be responsible for providing one of these.

One key aspect of these interfaces is that modules are referenced by their full composer package names.

### Requirements

Right now we have a call such as:

``` php
Requirements::javascript(FRAMEWORK_ADMIN_DIR . '/client/dist/js/vendor.js');
```

This could be replaced with an inlined call to the ClientResourceLocator, but this is very verbose:

``` php
Requirements::javascript(
  Injector::inst()
    ->get('SilverStripe\\Core\\ResourceLocator')
    ->getResource('framework/admin', 'client/dist/js/vendor.js')
    ->getURL()
);
```

So my recommendation is that we have additional support methods in Requirements that handle this:

``` php
Requirements::packageJavascript('framework/admin', 'client/dist/js/vendor.js');
```

Or in a template:

```
<% require packageJavascript('framework/admin', 'client/dist/js/vendor.js') %>
```

### Finding modules

For finding the location of the packages, I would expect a relatively simple implementation such as this one would suffice by default. Potentially, package searching could also be a pluggable Injector service.

``` php
function pathFor($package) {
   if(file_exists(BASE_PATH . '/vendor/' . $package)) {
    return BASE_PATH . '/vendor/' . $package;
   }

  $shortname = substr($package, strpos($package)+1);
   if(file_exists(BASE_PATH . '/' . $shortname)) {
     return BASE_PATH . '/' . $shortname;
   }

[...]

PR #6733, opened 2017-03-23
API Add experimental getResource() to Module

tractorcow:
Partial implementation of https://github.com/silverstripe/silverstripe-framework/issues/6243

Also required for https://github.com/silverstripe/silverstripe-framework/issues/6735

### DIFF
--- a/src/Core/Manifest/Module.php
+++ b/src/Core/Manifest/Module.php
@@ -4,7 +4,6 @@ namespace SilverStripe\Core\Manifest;
 
 use Exception;
 use Serializable;
-use SilverStripe\Assets\File;
 
 class Module implements Serializable
 {
@@ -163,7 +162,9 @@ class Module implements Serializable
      */
     public function getResourcePath($path)
     {
-        return File::join_paths($this->getRelativePath(), $path);
+        $base = rtrim($this->getRelativePath(), '/\\');
+        $path = rtrim($path, '/\\');
+        return "{$base}/{$path}";
     }
 
     /**
--- a/src/Core/Manifest/Module.php
+++ b/src/Core/Manifest/Module.php
@@ -4,6 +4,7 @@ namespace SilverStripe\Core\Manifest;
 
 use Exception;
 use Serializable;
+use SilverStripe\Assets\File;
 
 class Module implements Serializable
 {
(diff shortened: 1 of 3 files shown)


tractorcow:
Rebased.

sminnee:
Looks good. Presumably if we updated Requirements to include assets through this method (it might need a new method signature) we could push modules into `vendor/`?

src/Core/Manifest/Module.php
```diff
@@ -148,4 +149,33 @@ protected function loadComposer()
...
+     * This method makes no distinction between public / local resources,
+     * which may change in the near future.
+     *
+     * @internal Experimental API and may change
+     * @param string $path File or directory path relative to module directory
+     * @return string Path relative to base directory
+     */
+    public function getResource($path)
+    {
+        return File::join_paths($this->getRelativePath(), $path);
```
sminnee: Let's remove the dependence on File, it's way bigger than is needed here.

src/Core/Manifest/Module.php
```diff
@@ -148,4 +149,33 @@ protected function loadComposer()
...
+     * Directories included
+     *
+     * This method makes no distinction between public / local resources,
+     * which may change in the near future.
+     *
+     * @internal Experimental API and may change
+     * @param string $path File or directory path relative to module directory
+     * @return string Path relative to base directory
+     */
+    public function getResource($path)
```
sminnee: Would suggest `getResourcePath()`?

src/Core/Manifest/Module.php
chillu: OK, using `rtrim()` instead

sminnee: Maybe we could transpile `left-pad` for this?

src/Core/Manifest/Module.php
```diff
@@ -148,4 +148,35 @@ protected function loadComposer()
...
+     *
+     * @internal Experimental API and may change
+     * @param string $path File or directory path relative to module directory
+     * @return string Path relative to base directory
+     */
+    public function getResourcePath($path)
+    {
+        $base = rtrim($this->getRelativePath(), '/\\');
+        $path = rtrim($path, '/\\');
+        return "{$base}/{$path}";
```
tractorcow: ltrim this please; Not accounting for case where `$base` is empty string.

dhensby: Who's going to PR this?

chillu: https://github.com/silverstripe/silverstripe-framework/pull/6750

PR #6734, opened 2017-03-23
Respecting PHPUnit APIs in SapphireTest

dhensby:
- Removing `TestListener` as it's not required
- Make use of core PHPUnit features:
  - ~~`backupGlobals` (we have too much core test code that stores global state just to reset it - this is built in)~~
  - make use of `tearDownBefore/AfterClass()` in place of `setUp/tearDownOnce()`

Merge with:

- [ ] https://github.com/silverstripe/silverstripe-versioned/pull/5
- [ ] https://github.com/silverstripe/silverstripe-siteconfig/pull/49
- [ ] https://github.com/silverstripe/silverstripe-assets/pull/4
- [ ] https://github.com/silverstripe/silverstripe-admin/pull/10
- [ ] https://github.com/silverstripe/silverstripe-graphql/pull/91


sminnee:
@dhensby the PR notes have nothing to do with the commit...?

dhensby:
> @dhensby the PR notes have nothing to do with the commit...?

I'm working on it ;)

sminnee:
Flagged as WIP then :)

dhensby:
Right - I've moved the bulk of this to #6736 

I'll leave this to deal with the difficult task of removing our TestListener

dhensby:
It may be better for us to fully deprecate setUp/tearDownOnce() and make people use the static method setUpBeforeClass()/tearDownAfterClass(). From what I can see our core logic in these methods can be made static fairly easily.

src/Dev/SapphireTest.php
```diff
@@ -70,6 +72,16 @@ class SapphireTest extends PHPUnit_Framework_TestCase
      */
     protected $fixtureFactory;
 
+    protected $backupGlobalsBlacklist = array(
+        'injector',
```
tractorcow: I'm a bit confused about these. Are all top level variables automatically global? We use `$injector` not `global $injector` in our code. I had a look at the phpunit code, and it seems to piggy-back off `$GLOBALS` which doesn't seem to include these variables.

dhensby: Any variable declared in the global scope: http://php.net/manual/en/reserved.variables.globals.php

In PHP all variables that aren't declared in functions/classes are automatically in the global scope. You can access these vars using the `global` keyword.

On another point, there is a risk that modules that declare vars in the global scope will then have to construct their own blacklist - so maybe we create an API for that?

tractorcow: Ok, they are all globals. I learn something new every day.

src/Dev/SapphireTest.php
```diff
@@ -327,6 +333,12 @@ protected function setUp()
         Email::config()->remove('bcc_all_emails_to');
     }
 
+    public static function setUpBeforeClass()
+    {
+        self::$test_singleton = singleton(static::class);
```
tractorcow: It may be wise to put `static::start()` in here to ensure that the manifest is configured for tests prior to any code touching injector (as `singleton()` does).

src/Dev/SapphireTest.php
```diff
@@ -402,6 +414,12 @@ public function setUpOnce()
         }
     }
 
+    public static function tearDownAfterClass()
+    {
+        self::$test_singleton->tearDownOnce();
```
tractorcow: Can we use `singleton` here too in place of more static variables?

dhensby: Nope, we can't use singleton because `setUpOnce` stores state on the object instance so we need to keep a reference to the object instance.

src/Dev/SapphireTest.php
dhensby: done

dhensby:
OK - I've dropped backupGlobals from this PR.

sminnee:
Looks like we need to note the API changes to SapphireTest in README.md; apart from that it looks good.

dhensby:
Docs added

src/Dev/SapphireTest.php
```diff
@@ -159,12 +157,14 @@ class SapphireTest extends PHPUnit_Framework_TestCase
     protected $backupGlobals = false;
 
     /**
-     * Helper arrays for illegalExtensions/requiredExtensions code
+     * Helper arrays for illegal_extensions/required_extensions code
      */
-    private $extensionsToReapply = array(), $extensionsToRemove = array();
+    private static $extensions_to_reapply = [];
+
+    private static $extensions_to_remove = [];
```
tractorcow: Protip; You can add `@internal` on any private static to make it non-config. :)

PR #6744, opened 2017-03-27
fixing val reference to url in https hotlink and allowes https hotlinking on TinyMCE

ishannz:


ishannz:
Removes https:// overwrite on images URLs, so http: or https images can be linked.

sminnee:
Thanks, @ishannz! Can you update the ticket description and title to explain the bug you're fixing in a little more detail, without reference to dev/pharmac?

ishannz:
there have two commits in this merge request, the originals commit belong to @d-rivera-c 
@d-rivera-c  can you please give some more information that explains the bug your are fixing.

https://github.com/silverstripe/silverstripe-framework/pull/6744/commits/22b3a71ec0c8cd8c38030fa0bf5449abefafe8a3 is 
fixing val reference to url in https hotlink 

https://github.com/silverstripe/silverstripe-framework/pull/6744/commits/5612b498b9d25f7bc348cf83ce927c5834552deb
Removes http:// overwrite on file URLs, so http: or https files can be linked.

dhensby:
Hi @ishannz - I think what @sminnee is asking is for a bit more detail on the issue you're having and how this patch resolves it. If you could explain in the form of "when I try to do [action] in the CMS I can't because [reason], this patch addresses it by [solution]"

tractorcow:
I think the bug is that even if you use a `https://` url, tinymce calls showFileView with the `http://` protocol url.

PR #6745, opened 2017-03-28
Require LICENSE in supported modules (no *.md)

chillu:
It's more standard to have this file in the webroot.
It's technically markdown compatible text (e.g. asterisk bullet points),
but there's not much point in rendering it via markdown.

If you use the Github "new repo" dialog, it'll create the file without
an extension, so that's pretty much considered the standard.


sminnee:
Let's make framework comply with this guideline as part of this PR, eh? ;)

PR #6746, opened 2017-03-28
FIX: Don't mistake \ for _ in dev/build

sminnee:
The query SHOW TABLES LIKE 'some_thing' will match the table some\thing.

This causes issues when the namespace separator has changed.

The fix is to escape the _s in this LIKE statement, as done here.


tractorcow:
Test failures pre-existing.

PR #6748, opened 2017-03-28
FIX Update config API reference in FulltextSearchable and use namespaced imports for class names

robbieaverill:


tractorcow:
Other issues are pre-existing.

PR #6750, opened 2017-03-28
Root-level module support in Module->getResourcePath()

chillu:
See https://github.com/silverstripe/silverstripe-framework/pull/6733


tractorcow:
Only fails due to pre-existing segfault error.

PR #6751, opened 2017-03-28
Add campaign-admin to CMS_TEST=1 on travis

chillu:
We want to remove this implicit dependency (through cms)
by fully decoupling campaign-admin (https://github.com/silverstripe/silverstripe-campaign-admin/issues/1),
but for now this is required.


tractorcow:
Improves tests even though it doesn't fix all of them.

PR #6752, opened 2017-03-28
DONTMERGE Disable opcache to get PHP7 coverage passing

chillu:
See https://github.com/travis-ci/travis-ci/issues/4091
https://github.com/travis-ci/travis-ci/issues/5323
https://github.com/silverstripe/silverstripe-cms/pull/1761


chillu:
OK, the build has been running for 30 mins (without segfault), which is slower than on PHP 5 - so that's not a viable solution

tractorcow:
I prefer moving coverage to php5 for now.

PR #6756, opened 2017-03-29
FIX: Add separate license file to match module standard.

sminnee:


tractorcow:
Not going to blame a moved markdown file for a segfault.

dhensby:
Should the attributions remain in the README? Not that it's a huge issue but GitHub isn't picking this repo up as BSD-3 as it has done with the others that have had the licence file moved.

PR #6764, opened 2017-03-30
API Drop IE10 support (fixes #6321)

### LINKED ISSUE #6321: VOTE: Do you or your clients use IE10?
Currently we have agreed to drop IE8 and IE9 support for SilverStripe 4. However, we had not yet decided whether IE10 or IE11 would be the minimum version.

Please note that when talking about browser support **we refer only to CMS usage** and not to the browser support of websites built with SilverStripe.

We will definitely support IE11 in SilverStripe as it is not possible to run newer versions of IE on Windows 7, and this is a common configuration for corporate IT environments.

IE11 is supported in Windows 7 service pack 1, and as best we can tell IE10 will auto-upgrade to IE11, so it is likely that IE10 use is limited, even in corporate environments.

We would like to get feedback on whether or not the SilverStripe community currently relies on IE10 support via votes on this card.

Please vote 👍 to one of the comments below to cast your vote.

chillu:


sminnee:
Cancelling a build like a cowboy, since it's just a doc change.

Issue #6626, opened 2017-02-15
Remove jQuery calendar from DateField


chillu:
Followup from #6194, where @tractorcow has refactored `DateField->setConfig('showcalendar', true)` to `DateField->setShowCalendar(true)`. This triggered a discussion whether the PHP API should care about these view aspects, in particular around the "sub views" implemented in DateField to support the CMS use case (`DateField_View_JQuery`). Since the introduction of this functionality, we've gained a [HTML5 date type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) which comes with built-in, localised date pickers in modern browsers like Chrome, IE Edge and Safari ([supported browser overview](http://caniuse.com/#search=date)). Given we want to remove jQuery UI from core in favour of React components, this was a good point to revisit the client-side library use as well.

## Merge Process

 * [ ] https://github.com/silverstripe/silverstripe-framework/pull/6766
 * [ ] https://github.com/silverstripe/silverstripe-admin/pull/12
 * [ ] https://github.com/silverstripe/silverstripe-cms/pull/1775
 * [ ] https://github.com/silverstripe/silverstripe-frameworktest/pull/32
 * [ ] https://github.com/silverstripe/silverstripe-testsession/pull/46

## Tasks

 * [x] Remove `DateField_View_JQuery` and jQuery datepicker related logic
 * [x] Remove `DateField.js` wrapper for jQuery UI
 * [x] Remove `DateField->setShowCalendar()`
 * [x] Add `DateField->setHTML5()` (consistent with `NumericField->setHTML5()`)
 * [x] Document the limitations of `DateField->setLocale()` and `setFormat()` (will be overruled by browser on presentation on `setHTML5(true)`). Throw exception if locale or format isn't `null` or ISO on `setHTML5(true)`
 * [x] Force format to ISO on `setHTML5(true)`
 * [ ] ~~Point to type=date polyfill if CMS users are stuck on older browsers (https://github.com/jonstipe/date-polyfill)~~ (building a MomentJS polyfill instead)
 * [ ] ~~Replicate logic to `DatetimeField` (with type=datetime-local)~~ (nice to have)
 * [x] Replicate logic to `TimeField` (with type=time)
 * [x] Add default placeholder to DateField documenting date ordering for usability (if type=date isn’t supported)
 * [x] Make scaffolded and core DateField use HTML5 by default
 * [x] Pass through min/max as properties on DateField
[...]

PR #6766, opened 2017-04-01
HTML5 Date Fields

chillu:
See https://github.com/silverstripe/silverstripe-framework/issues/6626


src/Forms/DateField.php
```diff
@@ -226,8 +222,20 @@ protected function getFormatter()
             IntlDateFormatter::NONE
         );
 
-        // Don't invoke getDateFormat() directly to avoid infinite loop
-        if ($this->dateFormat) {
+        $isoFormat = 'y-MM-dd';
```
tractorcow: `DBDate::ISO_DATE`

src/Forms/DateField.php
```diff
@@ -243,59 +251,38 @@ protected function getFormatter()
...
-     */
-    protected function renderWithClientView($callback)
-    {
-        $clientView = null;
-        if ($this->getShowCalendar()) {
-            $clientView = $this->getClientView();
-            $clientView->onBeforeRender();
+        if ($this->getHTML5()) {
+            // Browsers expect ISO 8601 dates, localisation is handled on the client
+            $this->setDateFormat('y-MM-dd');
```
tractorcow: Instead of modifying the state on render, can we move this logic to getDateFormat()? I realise we violate this in many places, but we should probably make fields immutable during render.

src/Forms/TimeField.php
```diff
@@ -58,6 +58,31 @@ class TimeField extends TextField
     protected $timezone = null;
 
     /**
+     * Use HTML5-based input fields (and force ISO 8601 date formats).
+     *
+     * @var bool
+     */
+    protected $html5 = false;
```
tractorcow: As above, should this be true by default?

src/Forms/TimeField.php
```diff
@@ -140,8 +165,20 @@ protected function getFormatter()
             $this->getTimezone()
         );
 
-        // Don't invoke getDateFormat() directly to avoid infinite loop
-        if ($this->timeFormat) {
+        $isoFormat = 'HH:mm:ss';
```
tractorcow: `DBTime::ISO_TIME`

src/Forms/TimeField.php
```diff
@@ -337,9 +360,19 @@ public function localisedToISO8601($time)
         $fromFormatter = $this->getFormatter();
         $toFormatter = $this->getISO8601Formatter();
         $timestamp = $fromFormatter->parse($time);
+
+        // Try to parse time without seconds, since that's a valid HTML5 submission format
+        // See https://html.spec.whatwg.org/multipage/infrastructure.html#times
+        if ($timestamp === false && $this->setHTML5(true)) {
```
tractorcow: Did you mean to use `$this->getHTML5()` as a condition? This method shouldn't mutate the object.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1495,19 +1507,27 @@ New `DatetimeField` methods replace `getConfig()` / `setConfig()`:
...
+* `getPlaceholders()` / `setPlaceholders()` moved to a new `SeparatedDateField` class
+* `getClientLocale` / `setClientLocale` have been removed (handled by `DateField->locale` and browser settings)
 
 New `TimeField` methods replace `getConfig()` / `setConfig()`
 
 * `getTimeFormat()` / `setTimeFormat()`
 * `getLocale()` / `setLocale()`
+* `getClientConfig()` has been removed (in favour of `setHTML5()`)
 
 #### <a name="overview-template-removed"></a>Template and Form Removed API
```
tractorcow: Need to add notes on the removal of `MemberDatetimeOptionsetField` to this section.

tractorcow: Maybe mention `DateField_View_JQuery` too here.

src/Security/Member.php
```diff
@@ -81,9 +80,6 @@ class Member extends DataObject implements TemplateGlobalProvider
         'Locale' => 'Varchar(6)',
         // handled in registerFailedLogin(), only used if $lock_out_after_incorrect_logins is set
         'FailedLoginCount' => 'Int',
-        // In ISO format
-        'DateFormat' => 'Varchar(30)',
```
tractorcow: Good riddance. :)

src/Security/Member.php
```diff
@@ -1315,19 +1311,16 @@ public function splitName($name)
...
      * @return string ISO date format
      */
     public function getDateFormat()
     {
-        $format = $this->getField('DateFormat');
-        if ($format) {
-            return $format;
-        }
-        return $this->getDefaultDateFormat();
+        $format = $this->getDefaultDateFormat();
```
tractorcow: We can probably just merge getDefaultDateFormat and getDateFormat

src/Security/Member.php
```diff
@@ -1343,19 +1336,17 @@ public function getLocale()
...
      * @return string ISO date format
      */
     public function getTimeFormat()
     {
-        $timeFormat = $this->getField('TimeFormat');
-        if ($timeFormat) {
-            return $timeFormat;
-        }
-        return $this->getDefaultTimeFormat();
+        $format = $this->getDefaultTimeFormat();
```
tractorcow: As above, we may as well merge these. The other place getDefault* methods were called are gone.

src/Forms/DateField.php
chillu: Heh I was looking for a ICU constant for this but couldn't find it - good call!

src/Forms/DateField.php
```diff
@@ -122,25 +115,27 @@ class DateField extends TextField
     protected $rawValue = null;
 
     /**
-     * Check if calendar should be shown on the frontend
+     * Use HTML5-based input fields (and force ISO 8601 date formats).
      *
+     * @var bool
+     */
+    protected $html5 = false;
```
chillu: Hmmm yeah I guess for most frontend implementations, it'll be a better default (opt-out of HTML5 rather than opt-in). For the CMS, it'll definitely be a better default, since we have a polyfill there. And it would save devs remembering that when writing `getCMSFields()`. I'll change it, and update the changelog

src/Forms/TimeField.php
chillu: Agreed, it's a better default - I'll change.

src/Forms/TimeField.php
chillu: Done

src/Forms/TimeField.php
chillu: Derp derp

docs/en/04_Changelogs/4.0.0.md
chillu: Done

src/Security/Member.php
chillu: Done

src/Security/Member.php
chillu: Done

src/Forms/DateField.php
chillu: Good idea. I want to perform it as late as possible, but overriding the getter behaviour is more internally consistent.

tractorcow:
@chillu your tests seem broken. Do they pass locally?

Issue #6759, opened 2017-03-29
RFC: PHP7 support in SilverStripe 3


sminnee:
## Problem: we only have ~15 months between SS4 stable release and SS3 end-of-life.

Right now SS3's end-of-life is tied to the end-of-life of the PHP version that it requires - 5.6. This is December 2018.
We're targeting an October 2017 release for SS4 stable, which only leaves a 15 months window for people to upgrade.

For comparison, we had 33 months (2 years, 9 months) between 3.0 being released and 2.4 being end-of-life'd.

15 months is too short a window and I would like a minimum of 2 years.

## Solution: PHP7 support in SS3

In order to resolve this, we want to support PHP7 in SS3. This could start at 3.6. However, as this requires an API breakage, we would be breaking semantic versioning to do this.

Having discussed it as a core team, we don't want to have semantic versioning be something that we do when convenient, and so we've come up with another solution: we will put the breaking version of SS3 into a separate package, `silverstripe/framework3lts`

What this will means is that the version list will look a bit unusual:

 * SilverStripe 2.0 - 3.5: `silverstripe/framework`
 * SilverStripe 3.6 - 3.999: `silverstripe/framework3lts`
 * SilverStripe 4.0+: `silverstripe/framework`

If you don't want to worry about this, you can skip straight from 3.5 to 4.0. But if you want to get onto PHP7 before upgrading to SS4, you will have a way forward that doesn't require pulling in patches to your system.

## Specifics

* 3.5 is the last minor 3.x release on the `silverstripe/framework` package. We support this minor release with security fixes until Dec 2018
* We create package silverstripe/framework3lts, by forking the `3` branch of silverstripe/framework.
  * composer.json will indicate provides = silverstripe/framework, so we can 
  * php requirements will be PHP 5.5 - 7.1,
  * We rename the Int class to DBInt, the Float class to DBFloat, and make use of Injector for instantiating field types, with "Int" and "Float" set up as service names.
* 3.6 is the first release on `silverstripe/framework3lts`.
* It’s supported with minor releases until Dec 2019
* From version 3.6 onwards, silverstripe/installer includes `silverstripe/framework3lts`, so that newly created projects start on this version.

## What are the API breakages?

They're pretty small:

 * Class `Int` must become `DBInt`
 * Class `Float` must become `DBFloat`
 * The `$db` and `$casting` variables will be reinterpreted to point to Injector services rather than classes
[...]

PR #6771, opened 2017-04-03
PHP 7 compatibility

kinglozzer:
Proposed solution to https://github.com/silverstripe/silverstripe-framework/issues/6759

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,11 +17,15 @@ matrix:
       env: DB=SQLITE
     - php: 5.6
       env: DB=MYSQL PDO=1
+    - php: 7.0
+      env: DB=MYSQL
+    - php: 7.1
+      env: DB=MYSQL PDO=1
     - php: 5.6
       env: DB=MYSQL BEHAT_TEST=1
-    - php: 5.6
+    - php: 7.0
       env: DB=MYSQL CMS_TEST=1
-    - php: 5.6
+    - php: 7.0
       env: DB=MYSQL BEHAT_TEST=1 CMS_TEST=1
 
 before_script:
--- a/model/fieldtypes/DBFloat.php
+++ b/model/fieldtypes/DBFloat.php
@@ -5,7 +5,7 @@
  * @package framework
  * @subpackage model
  */
-class Float extends DBField {
+class DBFloat extends DBField {
 
 	public function __construct($name = null, $defaultVal = 0) {
 		$this->defaultVal = is_float($defaultVal) ? $defaultVal : (float) 0;
--- a/model/fieldtypes/compat/Float.php
+++ b/model/fieldtypes/compat/Float.php
@@ -1,0 +1,6 @@
+<?php
+
+class Float extends DBFloat
+{
+
+}
--- a/model/fieldtypes/compat/autoload.php
+++ b/model/fieldtypes/compat/autoload.php
@@ -1,0 +1,17 @@
+<?php
+
+if (PHP_MAJOR_VERSION < 7) {
+	spl_autoload_register('php5_compat_autoloader');
+}
+
+function php5_compat_autoloader($classname) {
+	$classMap = array(
+		"int" => "/framework/model/fieldtypes/compat/Int.php",
+		"float" => "/framework/model/fieldtypes/compat/Float.php",
+	);
+
+	$classname = strtolower($classname);
+	if(isset($classMap[$classname])) {
+		require_once BASE_PATH . $classMap[$classname];
+	}
+}
--- a/view/SSViewer.php
+++ b/view/SSViewer.php
@@ -536,7 +536,7 @@ class SSViewer_DataPresenter extends SSViewer_Scope {
 					// If not provided, use default
 					if (!$casting) $casting = Config::inst()->get('ViewableData', 'default_cast', Config::FIRST_SET);
 
-					$obj = new $casting($property);
+					$obj = Injector::inst()->create($casting, false, array($property));
 					$obj->setValue($res['value']);
 
 					$res['obj'] = $obj;
--- a/view/ViewableData.php
+++ b/view/ViewableData.php
@@ -93,7 +93,18 @@ class ViewableData extends Object implements IteratorAggregate {
 	 * @return bool
 	 */
 	public function __isset($property) {
-		return $this->hasField($property) || ($this->failover && $this->failover->hasField($property));
+		// getField() isn't a field-specific getter and shouldn't be treated as such
+		if (strtolower($property) !== 'field' && $this->hasMethod($method = "get$property")) {
+			return true;
+
+		} elseif ($this->hasField($property)) {
+			return true;
+
+		} elseif ($this->failover) {
+			return isset($this->failover->$property);
+		}
+
+		return false;
 	}
 
 	/**
@@ -104,13 +115,17 @@ class ViewableData extends Object implements IteratorAggregate {
 	 * @return mixed
 	 */
 	public function __get($property) {
-		if($this->hasMethod($method = "get$property")) {
+		// getField() isn't a field-specific getter and shouldn't be treated as such
+		if (strtolower($property) !== 'field' && $this->hasMethod($method = "get$property")) {
 			return $this->$method();
-		} elseif($this->hasField($property)) {
+
+		} elseif ($this->hasField($property)) {
 			return $this->getField($property);
-		} elseif($this->failover) {
+
+		} elseif ($this->failover) {
 			return $this->failover->$property;
 		}
+		return null;
 	}
 
 	/**
--- a/view/ViewableData.php
+++ b/view/ViewableData.php
@@ -331,7 +331,7 @@ class ViewableData extends Object implements IteratorAggregate {
 	public function escapeTypeForField($field) {
 		$class = $this->castingClass($field) ?: $this->config()->default_cast;
 
-		return Config::inst()->get($class, 'escape_type', Config::FIRST_SET);
+		return Injector::inst()->get($class, true)->config()->escape_type;
 	}
 
 	/**
--- a/_config/database.yml
+++ b/_config/database.yml
@@ -20,3 +20,7 @@ Injector:
   PDOConnector:
     class: 'PDOConnector'
     type: prototype
+  Int:
+    class: DBInt
+  Float:
+    class: DBFloat
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
 		}
 	],
 	"require": {
-		"php": ">=5.3.3,<7",
+		"php": ">= 5.3.3, <7.2",
 		"composer/installers": "~1.0"
 	},
 	"require-dev": {
--- a/core/Config.php
+++ b/core/Config.php
@@ -704,8 +704,8 @@ class Config_LRU {
 
 	public function __construct() {
 		Deprecation::notice('4.0', 'Please use Config_MemCache instead', Deprecation::SCOPE_CLASS);
-		if (version_compare(PHP_VERSION, '5.3.7', '<')) {
-			// SplFixedArray causes seg faults before PHP 5.3.7
+		if (version_compare(PHP_VERSION, '5.3.7', '<') || version_compare(PHP_VERSION, '6.99.99', '>')) {
+			// SplFixedArray causes seg faults before PHP 5.3.7, and also in 7.0.0-RC1
 			$this->cache = array();
 		}
 		else {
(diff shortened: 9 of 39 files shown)


kinglozzer:
Okay, I’ve gone far enough down the rabbit hole for this evening. Outstanding issues are:

- segfault on 7.1 at ~25% test completion. I don’t know which test yet
- version constraints on reports module prevents CMS builds running

sminnee:
* Pin the 7.1 tests to 7.1.2. We've had some issues with 7.1.3 on master.
 * Let's relax the version constraints for the supporting modules and re-run this.

model/fieldtypes/compat/Float.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
```
tractorcow: Should we put a conditional on PHP_MAJOR_VERSION in here too? I'd move it to the start of this code from the DBFLoat.php file.


```
<?php

if (PHP_MAJOR_VERSION < 7) {
	return;
}

class Float extends DBFloat {
}
```

view/ViewableData.php
```diff
@@ -331,7 +331,7 @@ public function castingClass($field) {
 	public function escapeTypeForField($field) {
 		$class = $this->castingClass($field) ?: $this->config()->default_cast;
 
-		return Config::inst()->get($class, 'escape_type', Config::FIRST_SET);
+		return Injector::inst()->get($class, true)->config()->escape_type;
```
tractorcow: Could we just use singleton()?

view/SSViewer.php
```diff
@@ -536,7 +536,7 @@ public function getInjectedValue($property, $params, $cast = true) {
 					// If not provided, use default
 					if (!$casting) $casting = Config::inst()->get('ViewableData', 'default_cast', Config::FIRST_SET);
 
-					$obj = new $casting($property);
+					$obj = Injector::inst()->get($casting, false, array($property));
```
tractorcow: Shall we just use ->create() here? We normally use ->get() for singletons (just for consistency of style).

.travis.yml
```diff
@@ -18,10 +18,12 @@ matrix:
     - php: 5.6
       env: DB=MYSQL PDO=1
     - php: 5.6
-      env: DB=MYSQL BEHAT_TEST=1
-    - php: 5.6
+      env: DB=MYSQL BEHAT_TEST=1 CMS_TEST=1
```
tractorcow: Did you mean to remove the framework behat tests here? You're running cms behat tests twice now.

model/fieldtypes/compat/Float.php
sminnee: I don't think this will do anything as if the file is loaded you've already lost — with a compile-time error.

view/ViewableData.php
sminnee: I prefer this style myself.

view/SSViewer.php
sminnee: Agreed.

sminnee:
^ PR for reports above.

kinglozzer:
I also need to update the PHP version constraint to `<7.2`

model/fieldtypes/compat/Float.php
tractorcow: I'd rather a `class_exists` return false in PHP7 than an error be emitted.

view/ViewableData.php
tractorcow: No problem then. :)

sminnee:
@kinglozzer pushed a few commits

sminnee:
One thing I'd note with this PR is that we now have 9 build-jobs, which feels a bit bloatly.

 * Shall we drop the PHP 5.4 and PHP 5.5 and SQLite builds? Keeping 5.3 and 5.6 should pick up most issues that affect 5.4 or 5.5.

model/fieldtypes/DBFloat.php
```diff
@@ -60,3 +60,7 @@ public function prepValueForDB($value) {
 	}
 
 }
+
+if (PHP_MAJOR_VERSION < 7) {
+	include_once 'compat/Float.php';
```
tractorcow: I still don't like classes being manually included in this way. It's not the job of DBFloat to trigger loading of Float.php.

model/fieldtypes/compat/Float.php
tractorcow: See my comment above. My major dislike is the roles that the php-7 equivalent fields in trying to load these deprecated classes.

model/fieldtypes/DBFloat.php
sminnee: I like this better than a manifest hack tbh

tractorcow: I don't think we need to hack manifest.

sminnee:
On the segfault, I believe that this has something to do with code in the LRU cache for Config. I had to patch this when getting PHP7 support working in SS4

model/fieldtypes/DBFloat.php
sminnee: What's your solution? Note that putting the check into the file won't block the parse error that PHP7 will throw.

tractorcow: Oh, I hadn't realised. I thought PHP7 would safely escape at the "return". Can someone confirm this?

sminnee: Confirmed with this test script:

```php
<?php

if(PHP_MAJOR_VERSION >= 7) {
	return;
}

if(PHP_MAJOR_VERSION < 7) {
	class Int {
	  function go() { echo "good"; }
	}
}

// $i = new Int();
// $i->go();
```

Results in

```
Fatal error: Cannot use 'Int' as class name as it is reserved in /Users/sminnee/Sites/ss3/framework/test-int.php on line 8
```

sminnee: One issue with Loz' patch is that calling class_exists('Int') or a before class_exists('DBInt') or similar will return false, as the manifest won't pick up Int. Ditto subclassing Int, etc.

If that's a showstopper, one solution would be to load an additional autoloader if PHP < 7, which would load the Int and Float classes as needed.

I'd prefer an additional autoloader to hacking manifest, simply because it reduces the impact on pre-existing code.

sminnee:
FYI I think this is the segfault fix: https://github.com/silverstripe/silverstripe-framework/pull/4551/commits/f5af0c85bade63174d282162a6c99fb37461b433

sminnee:
We'll also need https://github.com/silverstripe/silverstripe-framework/pull/6232/commits/c80417a949c7f2821ab4ce1f76ccbc5b6649fcaf to fix the GridField test, but first we need a quick semver check.

model/fieldtypes/DBFloat.php
tractorcow: With a hammer-proof helmet on I present this solution.


```
<?php

if (PHP_MAJOR_VERSION < 7) {
    eval('class Int extends DBInt {}');
}
```

Still requires inclusion somewhere, as class manifest won't pick it up.

model/fieldtypes/compat/Float.php
sminnee: @kinglozzer @dhensby what do you think of this approach? I've added a separate autoloader for the Int/Float classes so as not to pollute manifest: https://github.com/silverstripe/silverstripe-framework/pull/6771/commits/c3ed81b9df0e278b85143fff59836b41ce3cdbb4

tractorcow:
I propose merge on green. :) I'm happy with the solution now.

model/fieldtypes/compat/autoload.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+if (PHP_MAJOR_VERSION < 7) {
+	spl_autoload_register('php5_compat_autoloader');
+}
+
+function php5_compat_autoloader($classname) {
+	$classMap = [
```
tractorcow: Can't use php5 short array syntax in php5.3.

tractorcow:
GridFieldTest::testGetStateData() is failing. I have a note in master regarding this test:

```
// @todo - PHP 7.0.6 change requires __isset() to return true
// for each reference from left to right along an isset() invocation.
// See https://bugs.php.net/bug.php?id=62059
```

Failing on https://github.com/kinglozzer/sapphire/blob/c3ed81b9df0e278b85143fff59836b41ce3cdbb4/tests/forms/GridFieldTest.php#L130

sminnee:
@kinglozzer @dhensby I've made a lot of pushes to this branch to carry this forward.

The main things I did:
 * copy a few fixes from the SS4 compat work needed to get the tests passing (ViewableData changes, and removed use of SplFixedArray)
 * add an extra autoloader for Int and Float so that no failures occur if these classes are referenced before DBInt / DBFloat is loaded. added a version-aware test for this stuff too
 * addressed a few of Damian's suggestions

Still waiting for Travis but *hopefully* it's green. Feel free to rebase/squash this as you see fit if you want to keep hacking on it while I'm asleep. ;-)

kinglozzer:
Looks like the current failures (a few cases of `Undefined property: X`) could have been caused by the magic getters behaviour change, that does make me a little nervous about it...

model/fieldtypes/compat/autoload.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+if (PHP_MAJOR_VERSION < 7) {
+	spl_autoload_register('php5_compat_autoloader');
```
kinglozzer: 👍

model/fieldtypes/compat/Float.php
dhensby: Wrapping in a condition like that doesn't actually work. PHP 7 seems to parse the file and throw a "can't name class Float" error before even evaluating the condition.

What benefit does the new auto-loader add? Just means we don't have a conditional include of a stub? I suppose that makes sense.

dhensby:
> One thing I'd note with this PR is that we now have 9 build-jobs, which feels a bit bloatly.

@sminnee we can consolidate these because we have about 4 5.6 builds.. we only need one and can move the behat tests and cms tests about...

dhensby:
> We'll also need c80417a to fix the GridField test, but first we need a quick semver check.

I'm happy this isn't a breaking change... we're just stopping bad edge-cases, aren't we?

kinglozzer:
Right, more commits added:

- https://github.com/silverstripe/silverstripe-framework/pull/6771/commits/314fd406809e14ccd904d1031058560e1d49a70b - fixes an issue raised by the changes to `ViewableData::__isset()` as mentioned above. Similar fixes need to be added to CMS
- https://github.com/silverstripe/silverstripe-framework/pull/6771/commits/2c1736ae08f49e5f9fa723d53f27e933e595482f - copied from the `silverstripe/silverstripe-assets` module, stops PHP 7 complaining about performing numeric operations on strings like `512M`
- https://github.com/silverstripe/silverstripe-framework/pull/6771/commits/788ca7f7b8516f3ca18ef2d56d5637578960a6a0 - Fixes PHP 7.1+ `non-numeric value was encountered` error: https://travis-ci.org/silverstripe/silverstripe-framework/jobs/218382148#L550. tdlr; a unit test expects that calling `Image::SetHeight('String')` will raise an exception, it never got that far on PHP 7.1 because of the above warnings. `intval()` is equivalent to what PHP does if the error is suppressed anyway.

model/fieldtypes/compat/Float.php
sminnee: It means that you won't get a fatal error if you refer to class int without first referring to class dbint. The original approach suffered from that weakness.

sminnee:
@kinglozzer you wanna do a similar PR for CMS. We're close!

sminnee:
I'm comfortable with all Loz' commits; shall we git rebase -i this into a mergable shape?

sminnee:
I recommend that Dan/Loz polish this today and then @tractor merges in the NZ morning.
#excite!

dhensby:
@sminnee I'm working on the CMS fixes at the moment - it may not be done in time, though :(

sminnee:
Yeah, it's a bit surprising that `$this->CopyContentFromID` on VirtualPage throws a warning — that's a has_one relationship on the object so it should return a value without a warning.

kinglozzer:
I think we might be able to remove the `7.1.2` travis constraint - the segfault turned out to be unrelated to `7.1.3` specifically. I’ll let you guys take over again while I sleep 😄

sminnee:
Ok I've pushed the 7.1 dep back to "7.1" and squashed some of the the fixup commits. I've left the commits that cherry-pick from SS4 broken out separately, to assist with tracking. I also added a bit more text on the initial commit.

muskie9:
I like it if no major breakages... will likely move clients to a 7 environment once this is all sorted.

kinglozzer:
:tada:🍾 thanks for the help @sminnee @dhensby @tractorcow

PR #6774, opened 2017-04-04
Update test configuration

tractorcow:
See https://github.com/silverstripe/silverstripe-installer/pull/168

I've updated the travis yml to use more readable format, avoiding double quotes which would require us to have to escape the entire code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,62 +10,78 @@ addons:
 env:
   global:
     - COMPOSER_ROOT_VERSION=4.0.x-dev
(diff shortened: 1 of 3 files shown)


tractorcow:
Segfault and outstanding issue on date fields used in behat.

will attempt to fix with:
 - coverage without phpdbg
 - adjust behat code to interact better with date field polyfill

tractorcow:
PHP 7 build fixed.

CMS behat build fixed with https://github.com/silverstripe/silverstripe-admin/pull/16

PR #6777, opened 2017-04-05
FIX Add transparency percent argument to Image::generatePad to ensure transparency works from ::Pad

robbieaverill:
[This example](https://docs.silverstripe.org/en/3/developer_guides/files/images/#padded-image-resize) doesn't actually work, as it doesn't pass the `$transparencyPercent` variable to the image backend.


robbieaverill:
Yep, it is a fix - I'll rebase against whichever branch you suggest - perhaps 3.5?

dhensby:
whoop!

PR #6780, opened 2017-04-06
FIX Do not insert requirements more than once

### LINKED ISSUE #6779: [3.x] Requirements::includeInHTML inserts things into more than one </head> tag
For example: if I insert a third-party Javascript snippet (like an A/B tester, etc) into my page via:

```php
Requirements::customScript('//.....');
```

... and my custom script has a _string literal_ like `var whatever = "<head></head>"`, then `Requirements::includeInHTML` will inject requirements into that `</head>` as well as the closing tag that is actually part of the DOM. This theoretically would also affect code documentation examples that have a `</head>` in them (that aren't HTML escaped).

Adding a limit of 1 to the `preg_replace` calls seems safe and fixes the problem (unless you had the same thing happening in the head tag itself).

robbieaverill:
This change consolidates the string replacements used to insert requirements into the page content to help ensure
that they are not compounding and overwriting eachother.
    
The added test case includes where a user may have a Javascript snippet that contains a closing head tag, and the
test ensures that it does not get injected with requirements as well as the actual head tag in the DOM.

Resolves #6779

### DIFF
--- a/view/Requirements.php
+++ b/view/Requirements.php
@@ -863,19 +863,18 @@ class Requirements_Backend {
 				$requirements .= "$customHeadTag\n";
 			}
 
+			$replacements = array();
 			if ($this->force_js_to_bottom) {
-				// Remove all newlines from code to preserve layout
-				$jsRequirements = preg_replace('/>\n*/', '>', $jsRequirements);
+				$jsRequirements = $this->removeNewlinesFromCode($jsRequirements);
 
 				// Forcefully put the scripts at the bottom of the body instead of before the first
 				// script tag.
-				$content = preg_replace("/(<\/body[^>]*>)/i", $jsRequirements . "\\1", $content);
+				$replacements["/(<\/body[^>]*>)/i"] = $jsRequirements . "\\1";
 
 				// Put CSS at the bottom of the head
-				$content = preg_replace("/(<\/head>)/i", $requirements . "\\1", $content);
-			} elseif($this->write_js_to_body) {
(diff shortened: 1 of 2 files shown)


sminnee:
Have discussed an improvement to this with Robbie where we pass all the search/replace patterns into a single preg_replace call; this should give a robust & mergable approach.

It's a bugfix so good for a `3.5` merge IMO

robbieaverill:
@sminnee updated and rebased onto `3.5`

view/Requirements.php
```diff
@@ -892,24 +891,37 @@ public function includeInHTML($templateFile, $content) {
...
+				$replacements["/(<\/head>)/i"] = $requirements . "\\1";
 			} else {
-				$content = preg_replace("/(<\/head>)/i", $requirements . "\\1", $content);
-				$content = preg_replace("/(<\/head>)/i", $jsRequirements . "\\1", $content);
+				// Put CSS and Javascript together before the closing head tag
+				$replacements["/(<\/head>)/i"] = $requirements . $jsRequirements. "\\1";
 			}
+
+			// Replace everything at once (only once)
+			$content = preg_replace(array_keys($replacements), array_values($replacements), $content, 1);
```
sminnee: Maybe add `if($replacements)` here? If `$canWriteToBody and $this->write_js_to_body` are both set I think that replacements will be empty?

robbieaverill: There will always be something in the array since each condition adds to it, but it doesn't hurt to have it there anyway right? Pushing now.

Issue #6140, opened 2016-10-04
React form support for Date/Datetime fields


chillu:
## Acceptance criteria

 * Native HTML5 date field inserted into React forms where a formschema specifies a date field
 * Placeholder text is shown for non-compliant browsers
 * Placeholder text can be localised to the user's preferred date format – dd/mm/yyyy (using the browser or the member profile data can support this)

----

See https://github.com/silverstripe/silverstripe-framework/issues/5521 for context.
Preferrably while also removing the jQuery UI dependency (find a react-based date picker component)

## Tasks

- [x] Implement https://github.com/silverstripe/silverstripe-framework/issues/6626 (blocker)
- [x] investigate into date/time input native browser support
- [x] create date field component (include injector)
  - [x] update injector
  - [x] create react component
  - [x] write docs
  - [x] write tests
- [x] create time field component (include injector)
  - [x] update injector
  - [x] create react component
  - [x] write docs
  - [x] write tests
- [x] refactor back-end datetime field to not use date and time fields
  - [x] refactor
  - [x] update tests
  - [x] update changelogs
- [x] create datetime field component - single field (include injector)
  - [x] update injector
  - [x] create react component
  - [x] write docs
  - [x] write tests
- [x] Legacy (jQuery/entwine) field for date time input
---

## Pull request

- [x] https://github.com/silverstripe/silverstripe-framework/pull/6783
- [x] https://github.com/silverstripe/silverstripe-admin/pull/21
- [x] https://github.com/silverstripe/silverstripe-frameworktest/pull/34

chillu:
This should be extremely trivial if we use native browser functionality like suggested in https://github.com/silverstripe/silverstripe-framework/issues/6626

chillu:
I've thought about this a bit more, and think we can create an acceptable user experience without pulling in a whole bunch of code for a date picker UI - I'm hoping to get rid of jQuery UI in the lifespan of 4.x.

Additional research:
- Current uses of jQuery UI: Non-React tabs, toggle fields. We've removed the need for jQuery UI Dialogs (all React now), and JSTree never needed it. It seems possible to phase out jQuery UI, unless we commit to it in 4.x through date polyfills. Changes to browser support would be considered an API breakages, so we're effectively supporting IE10 without HTML5 date support until 2020.
- With our current bundling strategy, 85% of HTML5 compatible browser will receive an extra 70kb of jQuery UI even though they don't need it. Splitting this out would require server-side user agent checks.
[...]

PR #6783, opened 2017-04-06
Add 'lang' attribute to front-end date field schema

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/6140

### DIFF
--- a/src/Forms/DatetimeField.php
+++ b/src/Forms/DatetimeField.php
@@ -8,16 +8,10 @@ use SilverStripe\i18n\i18n;
 use SilverStripe\ORM\FieldType\DBDatetime;
 
 /**
- * Form field used for editing date time string
- *
- * # Localization
- *
- * See {@link DateField}
- *
- * # Configuration
- *
- * - "timezone": Set a different timezone for viewing. {@link dataValue()} will still save
- * the time in PHP's default timezone (date_default_timezone_get()), its only a view setting.
+ * Form field used for editing date time strings.
+ * By default, the field handles strings in normalised ISO 8601 format,
+ * for example 2017-04-26T23:59:59. The "T" separator can be replaced with a whitespace for value setting.
+ * Data is passed on via {@link dataValue()} with whitespace separators.
  */
 class DatetimeField extends TextField
 {
@@ -79,6 +73,9 @@ class DatetimeField extends TextField
      */
     protected $rawValue = null;
 
+    /**
+     * @inheritDoc
+     */
     protected $schemaDataType = FormField::SCHEMA_DATA_TYPE_DATETIME;
 
     /**
@@ -88,10 +85,12 @@ class DatetimeField extends TextField
      */
     protected $dateTimeOrder = '{date} {time}';
 
-    public function __construct($name, $title = null, $value = "")
-    {
-        parent::__construct($name, $title, $value);
-    }
+    /**
+     * Custom timezone
+     *
+     * @var string
+     */
+    protected $timezone = null;
 
     public function setForm($form)
     {
@@ -171,7 +170,7 @@ class DatetimeField extends TextField
     /**
      * Convert date localised in the current locale to ISO 8601 date
      *
-     * @param string $date
+     * @param string $datetime
      * @return string The formatted date, or null if not a valid date
      */
     public function localisedToISO8601($datetime)
@@ -181,6 +180,13 @@ class DatetimeField extends TextField
         }
         $fromFormatter = $this->getFormatter();
         $toFormatter = $this->getISO8601Formatter();
+
+        // Remove 'T' date and time separator before parsing (required by W3C HTML5 fields)
+        if ($this->getHTML5()) {
+            $datetime = str_replace('T', ' ', $datetime);
+        }
+
+        // Try to parse time with seconds
         $timestamp = $fromFormatter->parse($datetime);
 
         // Try to parse time without seconds, since that's a valid HTML5 submission format
@@ -204,7 +210,7 @@ class DatetimeField extends TextField
      */
     protected function getFormatter()
     {
-        if ($this->getHTML5() && $this->datetimeFormat && $this->datetimeFormat !== DBDatetime::ISO_DATETIME) {
+        if ($this->getHTML5() && $this->datetimeFormat && $this->datetimeFormat !== DBDatetime::ISO_DATETIME_NORMALISED) {
             throw new \LogicException(
                 'Please opt-out of HTML5 processing of ISO 8601 dates via setHTML5(false) if using setDatetimeFormat()'
             );
@@ -230,8 +236,9 @@ class DatetimeField extends TextField
         );
 
         if ($this->getHTML5()) {
-            // Browsers expect ISO 8601 dates, localisation is handled on the client
-            $formatter->setPattern(DBDatetime::ISO_DATETIME);
+            // Browsers expect ISO 8601 dates, localisation is handled on the client.
+            // Add 'T' date and time separator to create W3C compliant format
+            $formatter->setPattern(DBDatetime::ISO_DATETIME_NORMALISED);
         } elseif ($this->datetimeFormat) {
             // Don't invoke getDatetimeFormat() directly to avoid infinite loop
             $ok = $formatter->setPattern($this->datetimeFormat);
@@ -254,7 +261,7 @@ class DatetimeField extends TextField
     {
         if ($this->getHTML5()) {
             // Browsers expect ISO 8601 dates, localisation is handled on the client
-            $this->setDatetimeFormat(DBDatetime::ISO_DATETIME);
+            $this->setDatetimeFormat(DBDatetime::ISO_DATETIME_NORMALISED);
         }
 
         if ($this->datetimeFormat) {
@@ -295,6 +302,11 @@ class DatetimeField extends TextField
         // Build new formatter with the altered timezone
         $formatter = clone $this->getISO8601Formatter();
         $formatter->setTimeZone($timezone);
+
+        // ISO8601 date with a standard "T" separator (W3C standard).
+        // Note we omit timezone from this format, and we assume server TZ always.
+        $formatter->setPattern(DBDatetime::ISO_DATETIME_NORMALISED);
+
         return $formatter;
     }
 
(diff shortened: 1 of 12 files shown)


tractorcow:
I'm giving this a review now.

src/Forms/DatetimeField.php
```diff
@@ -181,6 +180,13 @@ public function localisedToISO8601($datetime)
         }
         $fromFormatter = $this->getFormatter();
         $toFormatter = $this->getISO8601Formatter();
+
+        // Remove 'T' date and time separator before parsing (required by W3C HTML5 fields)
+        if ($this->getHTML5()) {
```
tractorcow: I've tested that `$formatter->setLenient(true);` means you don't need the str_replace.

tractorcow: Ignore this, @chillu has indicated this would allow invalid dates to be passed in.

tractorcow: After some review, html5 date submissions should always include the `T`. We should test removing this line of code, and enforcing the existence of the `T`.

src/Forms/DatetimeField.php
```diff
@@ -58,59 +85,204 @@ class DatetimeField extends FormField
...
+     * This can be set explicitly. If not, this will be generated from the current locale
+     * with the current date length.
+     *
+     * @see http://userguide.icu-project.org/formatparse/datetime#TOC-Date-Field-Symbol-Table
+     */
+    public function getDatetimeFormat()
+    {
+        if ($this->getHTML5()) {
+            // Browsers expect ISO 8601 dates, localisation is handled on the client
+            $this->setDatetimeFormat(DBDatetime::ISO_DATETIME_NORMALISED);
```
tractorcow: Getters shouldn't modify state in unnecessary ways. Just return the constant directly here.

src/Forms/DatetimeField.php
```diff
@@ -58,59 +85,204 @@ class DatetimeField extends FormField
...
-        parent::__construct($name, $title, $value);
-    }
+    /**
+     * Custom timezone
+     *
+     * @var string
+     */
+    protected $timezone = null;
 
     public function setForm($form)
```
tractorcow: Delete this whole method. :)

src/Forms/DatetimeField.php
```diff
@@ -58,59 +85,204 @@ class DatetimeField extends FormField
...
     {
         parent::setForm($form);
-        $this->dateField->setForm($form);
-        $this->timeField->setForm($form);
         return $this;
     }
 
     public function setName($name)
     {
```
tractorcow: Can delete this method now.

src/Forms/DatetimeField.php
```diff
@@ -58,59 +85,204 @@ class DatetimeField extends FormField
...
+
+    public function getAttributes()
+    {
+        $attributes = parent::getAttributes();
+
+        $attributes['lang'] = i18n::convert_rfc1766($this->getLocale());
+
+        if ($this->getHTML5()) {
+            $attributes['type'] = 'datetime-local';
+            $attributes['min'] = $this->getMinDatetime();
```
tractorcow: min is currently in ISO8601 (not normalised). You need to convert these from 8601 to 8601(normalised), probably just in getAttributes() method. Minor distinction I know.

src/Forms/DatetimeField.php
```diff
@@ -58,59 +85,204 @@ class DatetimeField extends FormField
...
+            $attributes['max'] = $this->getMaxDatetime();
+        }
+
+        return $attributes;
+    }
+
+    public function getSchemaDataDefaults()
+    {
+        $defaults = parent::getSchemaDataDefaults();
```
tractorcow: We should add the min / max here. Please add these (and move the `html5` key) to the `data` property, rather than a top level key; `data` is intended for field-specific values, and top level keys are common across all fields.

src/Forms/DatetimeField.php
```diff
@@ -58,59 +85,204 @@ class DatetimeField extends FormField
...
+            'html5' => $this->getHTML5()
+        ]);
+    }
+
+    public function Type()
+    {
+        return 'text datetime';
+    }
+
+    public function getHTML5()
```
tractorcow: PHPdoc please. :)

src/Forms/DatetimeField.php
```diff
@@ -189,141 +375,167 @@ public function setValue($value, $data = null)
...
     {
-        $date = $this->dateField->Value();
-        $time = $this->timeField->Value();
-        return $this->joinDateTime($date, $time);
+        return $this->iso8601ToLocalised($this->value);
+    }
+
+    public function dataValue()
+    {
+        return $this->iso8601ToDataValue($this->value);
```
tractorcow: `$this->value` should only be saved in ISO8601. The `iso8601ToDataValue` method should not be necessary. Instead, we ensure that we only assign `$this->value = $some8601ConvertedVal`.

src/Forms/DatetimeField.php
```diff
@@ -162,11 +341,12 @@ protected function getISO8601Formatter()
...
+        $this->rawValue = $value;
+
         // Empty value
         if (empty($value)) {
             $this->value = null;
-            $this->dateField->setValue(null);
-            $this->timeField->setValue(null);
             return $this;
         }
         if (is_array($value)) {
```
tractorcow: setSubmittedValue() doesn't support array format anymore. We should fail if not a string.

src/Forms/DatetimeField.php
```diff
@@ -147,9 +324,11 @@ protected function getISO8601Formatter()
             date_default_timezone_get() // Default to server timezone
         );
         $formatter->setLenient(false);
-        // CLDR iso8601 date.
-        // Note we omit timezone from this format, and we assume server TZ always.
-        $formatter->setPattern('y-MM-dd HH:mm:ss');
+
+        // Note we omit timezone from this format, and we always assume server TZ
+        // ISO8601 date with a standard "T" separator (W3C standard).
+        $formatter->setPattern(DBDatetime::ISO_DATETIME_NORMALISED);
```
tractorcow: We should only use ISO_DATETIME_NORMALISED for the localised getFormatter(), and only if html5; Through the rest of framework we assume ISO_DATETIME.

src/Forms/DatetimeField.php
```diff
@@ -177,9 +357,15 @@ public function setValue($value, $data = null)
         // If invalid, assign for later validation failure
         $isoFormatter = $this->getISO8601Formatter();
         $timestamp = $isoFormatter->parse($value);
+
+        // Retry without "T" separator
+        if (!$timestamp) {
```
tractorcow: See above, getISO8601Formatter() should probably be fixed to ISO_DATETIME, then we won't need a fallback.

src/Forms/DatetimeField.php
```diff
@@ -189,141 +375,167 @@ public function setValue($value, $data = null)
...
-     * @param string $date
-     * @param string $time
-     * @return string
+     * Convert an ISO 8601 localised datetime into a saveable representation.
+     *
+     * @param string $datetime
+     * @return string The formatted date and time, or null if not a valid date and time
      */
-    protected function joinDateTime($date, $time)
+    public function iso8601ToDataValue($datetime)
```
tractorcow: See comment above; this method shouldn't be necessary, and we should save $value in ready-to-save-to-db format always.

src/Forms/TimeField.php
```diff
@@ -231,6 +231,14 @@ public function getAttributes()
         return $attributes;
     }
 
+    public function getSchemaDataDefaults()
+    {
+        $defaults = parent::getSchemaDataDefaults();
+        return array_merge($defaults, [
+            'html5' => $this->getHTML5()
```
tractorcow: Please nest non-standard schema properties within the `data` top level key.

tractorcow:
I'm also open to the idea of renaming some of these get*formatter methods.
getISO8601Formatter() could be getInternalFormatter(), for clarity of purpose.

src/Forms/DatetimeField.php
chillu: Good point, I overlooked that in my peer review - fixed now.

src/Forms/DatetimeField.php
chillu: Done

src/Forms/DatetimeField.php
chillu: Done

src/Forms/DatetimeField.php
chillu: Done

src/Forms/DatetimeField.php
chillu: That's done through `tidyInternal()` in `setMinDate()` now (which relies on `getInternalFormatter()` that's always set to `DBDatetime::ISO_DATETIME_NORMALISED`)

src/Forms/DatetimeField.php
chillu: Done

src/Forms/DatetimeField.php
chillu: Done

src/Forms/DatetimeField.php
chillu: Yeah I've picked that up in the meantime as well heh

src/Forms/DatetimeField.php
chillu: Hmm, `html5` defaults to `true` in the new implementation, which means we'd expect values in the new normalised ISO format. But lots of 3.x implementations will have called `setValue()` with an whitespace separated ISO date before. It'll be subtle breakages that I'd prefer to avoid.

src/Forms/TimeField.php
chillu: Done

src/Forms/DatetimeField.php
chillu: Yeah that simplifies the logic a lot, good call

src/Forms/DatetimeField.php
chillu: Done

src/Forms/DatetimeField.php
tractorcow: > in the new implementation, which means we'd expect values in the new normalised ISO format

Only from setSubmittedValue, though, right?

tractorcow:
We can delete DatetimeField::getTimezoneFormatter() now :)

tests/php/Forms/FormSchemaTest.php
```diff
@@ -462,6 +462,8 @@ public function testSchemaValidation()
                         'extraClass' => 'date text',
                         'description' => null,
                         'rightTitle' => null,
+                        'html5' => true,
```
tractorcow: This was moved to data right?

chillu: Fixed

PR #6786, opened 2017-04-07
Support booleans in SilverStripe\View\Parsers\Diff::getHTMLChunks

mikenz:
Convert booleans to strings ("true", "false") so they can be diff'd for display.

### DIFF
--- a/src/View/Parsers/Diff.php
+++ b/src/View/Parsers/Diff.php
@@ -161,15 +161,20 @@ class Diff extends \Diff
     }
 
     /**
-     * @param string|array $content If passed as an array, values will be concatenated with a comma.
+     * @param string|bool|array $content If passed as an array, values will be concatenated with a comma.
      * @return array
      */
     public static function getHTMLChunks($content)
     {
-        if ($content && !is_string($content) && !is_array($content) && !is_numeric($content)) {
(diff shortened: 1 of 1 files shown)


src/View/Parsers/Diff.php
```diff
@@ -161,15 +161,21 @@ public static function compareHTML($from, $to, $escape = false)
...
      */
     public static function getHTMLChunks($content)
     {
+        if (is_bool($content)) {
+            // Convert boolean to strings
+            $content = $content ? "true" : "false";
+        }
+
         if ($content && !is_string($content) && !is_array($content) && !is_numeric($content)) {
             throw new InvalidArgumentException('$content parameter needs to be a string or array');
```
flamerohr: Could we throw the `is_bool()` check to this `InvalidArgumentException`? Could move the conversion to below this, and I think this would be a good addition.

tractorcow:
Re-opening to froce tests to run again.

src/View/Parsers/Diff.php
tractorcow: It's not a merge blocker if we don't change this. I'll merge if green.

mikenz: I've change this tomorrow.

mikenz:
Reordered as per feedback

PR #6791, opened 2017-04-10
Documentation: Updated Index & Secure Coding Practices

caffeineinc:
Indexes: 
I've revamped the indexes page so it's a little easier to follow, and when you should use indexes in your SilverStripe application: 
- Removed the reference about the ability to use raw SQL (the framework doesn't support this). 
- Added documentation on what indexes the framework does add
- Added some information on composite indexes.
- Added some information about when indexes get created/updated & how.

Secure Coding Practices: 
I've added to the documentation to include
- Information about security headers, 
- The use of HTTPS, and forcing TLS.
- Using secure cookies (framework defaults to un-secure cookies)

### DIFF
--- a/docs/en/02_Developer_Guides/00_Model/12_Indexes.md
+++ b/docs/en/02_Developer_Guides/00_Model/12_Indexes.md
@@ -6,7 +6,8 @@ Indexes are a great way to improve performance in your application, especially a
 data model you can reduce the time taken for the framework to find and filter data objects. 
 
 The addition of an indexes should be carefully evaluated as they can also increase the cost of other operations such as 
-`UPDATE`/`INSERT` and `DELETE`. An index which has the same cardinality as the table will actually cost you performance. 
+`UPDATE`/`INSERT` and `DELETE`. An index on a column whose data is non unique will actually cost you performance.
+E.g. In most cases an index on `boolean` status flag, or `ENUM` state will not increase query performance.
 
 It's important to find the right balance to achieve fast queries using the optimal set of indexes; For SilverStripe 
 applications it's a good practice to: 
@@ -15,12 +16,12 @@ applications it's a good practice to:
 
 The SilverStripe framework already places certain indexes for you by default:
 - The primary key for each model has a `PRIMARY KEY` unique index
-- The `ClassName` column if your model is a direct decedent from `DataObject`
+- The `ClassName` column if your model inherits from `DataObject`
 - All relationships defined in the model have indexes for their `has_one` entity (for `many_many` relationships 
 this index is present on the associative entity).
 
 ## Defining an index
-Indexes are represented on a data object through the `DataObject::$indexes` array which maps index names to a 
+Indexes are represented on a `DataObject` through the `DataObject::$indexes` array which maps index names to a 
 descriptor. There are several supported notations:
 
 	:::php
@@ -40,7 +41,7 @@ descriptor. There are several supported notations:
 
 The `<column-name>` is used to put a standard non-unique index on the column specified. For complex or large tables 
 we recommend building the index to suite the requirements of your data.
- 
+
 The `<index-name>` can be an arbitrary identifier in order to allow for more than one index on a specific database 
 column. The "advanced" notation supports more `<type>` notations. These vary between database drivers, but all of them 
 support the following:
@@ -73,7 +74,7 @@ support the following:
 For complex queries it may be necessary to define a complex or composite index on the supporting object. To create a 
 composite index, define the fields in the index order as a comma separated list. 
 
-*Note* Most DBMSs only use the leftmost prefix to optimise the query, try to ensure the order of the index and your 
+*Note* Most databases only use the leftmost prefix to optimise the query, try to ensure the order of the index and your 
 query parameters are the same. e.g.
 - index (col1) - `WHERE col1 = ?`
 - index (col1, col2) = `WHERE (col1 = ? AND col2 = ?)`
--- a/docs/en/02_Developer_Guides/00_Model/12_Indexes.md
+++ b/docs/en/02_Developer_Guides/00_Model/12_Indexes.md
@@ -2,10 +2,26 @@ title: Indexes
 summary: Add Indexes to your Data Model to optimize database queries.
 
 # Indexes
+Indexes are a great way to improve performance in your application, especially as it grows. By adding indexes to your 
+data model you can reduce the time taken for the framework to find and filter data objects. 
 
-It is sometimes desirable to add indexes to your data model, whether to optimize queries or add a uniqueness constraint 
-to a field. This is done through the `DataObject::$indexes` map, which maps index names to descriptor arrays that 
-represent each index. There're several supported notations:
+The addition of an indexes should be carefully evaluated as they can also increase the cost of other operations such as 
+`UPDATE`/`INSERT` and `DELETE`. An index which has the same cardinality as the table will actually cost you performance. 
+
+It's important to find the right balance to achieve fast queries using the optimal set of indexes; For SilverStripe 
+applications it's a good practice to: 
+- add indexes on columns which are frequently used in `filter`, `where` or `orderBy` statements
+- for these, only include indexes for columns which are the most restrictive (return the least number of rows)
+
+The SilverStripe framework already places certain indexes for you by default:
+- The primary key for each model has a `PRIMARY KEY` unique index
+- The `ClassName` column if your model is a direct decedent from `DataObject`
+- All relationships defined in the model have indexes for their `has_one` entity (for `many_many` relationships 
+this index is present on the associative entity).
+
+## Defining an index
+Indexes are represented on a data object through the `DataObject::$indexes` array which maps index names to a 
+descriptor. There are several supported notations:
 
 	:::php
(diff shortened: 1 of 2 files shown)


docs/en/02_Developer_Guides/09_Security/04_Secure_Coding.md
```diff
@@ -585,6 +583,75 @@ In a future release this behaviour will be changed to be on by default, and this
...
+	class MySecureController extends Controller {
+		
+		public function init() {
+			parent::init();
+        
+			// max-age can be configured to anything in seconds: max-age=31536000 (1 year), 
+			// for rollout, consider something lower
+			// includeSubDomains to ensure all present and future sub domains will also be HTTPS
+			$this->response->addHeader('Strict-Transport-Security', 'max-age=86400; includeSubDomains');
+			$this->response->addHeader('Cache-control', 'max-age=0, must-revalidate, no-transform, no-cache, no-store');
```
tractorcow: I wonder if we should document `HTTP::add_cache_headers` here instead, since that's our preferred approach for setting cache-control headers.

caffeineinc: This is actually removing the cache headers so sensitive content _isnt_ cached... 

Do you mean setting `HTTP::$cache_age` to zero. and then calling `HTTP:: add_cache_headers` ?

tractorcow: I'd assume the same API should allow assignment of no-cache.

caffeineinc: @tractorcow I've changed it to use the `HTTP` interface - unfortunately it doesn't add all of the cache-control elements like `max-age=0, must-revalidate, no-transform` so i have to add these manually.

docs/en/02_Developer_Guides/09_Security/04_Secure_Coding.md
```diff
@@ -585,6 +583,80 @@ In a future release this behaviour will be changed to be on by default, and this
...
+
+	:::php
+
+	class MySecureController extends Controller {
+		
+		public function init() {
+			parent::init();
+        	
+			// Add cache headers to ensure sensitive content isn't cached.
+			$this->response->addHeader('Cache-control', 'max-age=0, must-revalidate, no-transform');
```
tractorcow: What's the difference between no-cache and max-age=0?

Take care, as we use `Cache-Control` (different case) in HTTP class.

caffeineinc: Hey @tractorcow
Good catch, i've fixed the case problem.
 `no-cache` - Forces caches to submit the request to the origin server for validation before releasing a cached copy.
`max-age` - Specifies the maximum amount of time a resource will be considered fresh. Contrary to Expires, this directive is relative to the time of the request. Max age is generally used for assets and is relative to the  but i'm following best practices recommended from a security review.

In theory no-cache should be enough (but that's being added by the `add_cache_headers` call now. The other headers just define more about the request to ensure nothing is reused, and everything should be re-requested, and some behaviour at WAF's vary depending on other headers used. By including them all (including pragma) I'm trying to cover all HTTP versions and providers and possible platforms.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
https://www.owasp.org/index.php/HTTP_Strict_Transport_Security_Cheat_Sheet

docs/en/02_Developer_Guides/09_Security/04_Secure_Coding.md
madmatt: @tractorcow Why are we suggesting to use static method calls instead of setting values on a Response object? I would have thought the static method calls is the outdated way of doing things, given the relative difficulty of testing those in isolation?

docs/en/02_Developer_Guides/09_Security/04_Secure_Coding.md
```diff
@@ -585,6 +583,75 @@ In a future release this behaviour will be changed to be on by default, and this
...
+
+	if (!Director::isDev()) {
+		Director::forceSSL();
+	}
+
+Forcing HTTPS so requires a certificate to be purchased or obtained through a vendor such as 
+[lets encrypt](https://letsencrypt.org/) and configured on your web server.
+
+In addition to this ensure the users are provided with secure cookies by setting the recommended "Secure" and 
+"HTTPOnly" flags. These flags prevent them from being stolen by an attacker through javascript.
```
caffeineinc: thanks, will do.

docs/en/02_Developer_Guides/00_Model/12_Indexes.md
```diff
@@ -2,10 +2,26 @@ title: Indexes
...
 
 # Indexes
+Indexes are a great way to improve performance in your application, especially as it grows. By adding indexes to your 
+data model you can reduce the time taken for the framework to find and filter data objects. 
 
-It is sometimes desirable to add indexes to your data model, whether to optimize queries or add a uniqueness constraint 
-to a field. This is done through the `DataObject::$indexes` map, which maps index names to descriptor arrays that 
-represent each index. There're several supported notations:
+The addition of an indexes should be carefully evaluated as they can also increase the cost of other operations such as 
+`UPDATE`/`INSERT` and `DELETE`. An index which has the same cardinality as the table will actually cost you performance.
```
chillu: "cardinality" is quite specialised terminology, is there a simpler term?

docs/en/02_Developer_Guides/00_Model/12_Indexes.md
```diff
@@ -2,10 +2,26 @@ title: Indexes
...
+`UPDATE`/`INSERT` and `DELETE`. An index which has the same cardinality as the table will actually cost you performance. 
+
+It's important to find the right balance to achieve fast queries using the optimal set of indexes; For SilverStripe 
+applications it's a good practice to: 
+- add indexes on columns which are frequently used in `filter`, `where` or `orderBy` statements
+- for these, only include indexes for columns which are the most restrictive (return the least number of rows)
+
+The SilverStripe framework already places certain indexes for you by default:
+- The primary key for each model has a `PRIMARY KEY` unique index
+- The `ClassName` column if your model is a direct decedent from `DataObject`
```
chillu: "direct decendant" is inaccurate - say "inherits from DataObject"

docs/en/02_Developer_Guides/00_Model/12_Indexes.md
```diff
@@ -2,10 +2,26 @@ title: Indexes
...
+- All relationships defined in the model have indexes for their `has_one` entity (for `many_many` relationships 
+this index is present on the associative entity).
+
+## Defining an index
+Indexes are represented on a data object through the `DataObject::$indexes` array which maps index names to a
```
chillu: Consistent spelling - use `DataObject`

docs/en/02_Developer_Guides/00_Model/12_Indexes.md
```diff
@@ -50,6 +69,25 @@ In order to use more database specific or complex index notations, we also suppo
 		);
 	}
 
+## Complex/Composite Indexes
+For complex queries it may be necessary to define a complex or composite index on the supporting object. To create a 
+composite index, define the fields in the index order as a comma separated list. 
+
+*Note* Most DBMSs only use the leftmost prefix to optimise the query, try to ensure the order of the index and your
```
chillu: Say "databases" rather the "DMBSs"

caffeineinc:
Thanks @tractorcow, @chillu, @madmatt  I've updated the PR with your changes.

docs/en/02_Developer_Guides/09_Security/04_Secure_Coding.md
tractorcow: > @tractorcow Why are we suggesting to use static method calls instead of setting values on a Response object? I would have thought the static method calls is the outdated way of doing things, given the relative difficulty of testing those in isolation?

For us to outdate this behaviour, we would need to refactor this to a non-static method, and add a deprecation to the HTTP class. Although this would be ideas, we haven't yet done this, so until then it's still standard practice.

docs/en/02_Developer_Guides/09_Security/04_Secure_Coding.md
```diff
@@ -585,6 +583,88 @@ In a future release this behaviour will be changed to be on by default, and this
...
+        	
+			// Add cache headers to ensure sensitive content isn't cached.
+			$this->response->addHeader('Cache-Control', 'max-age=0, must-revalidate, no-transform');
+			$this->response->addHeader('Pragma', 'no-cache'); // for HTTP 1.0 support
+
+			HTTP::set_cache_age(0);
+			HTTP::add_cache_headers($this->response);
+			
+			// Add HSTS header to force TLS for document content
+			$this->response->addHeader('Strict-Transport-Security', 'max-age=86400; includeSubDomains');
```
dhensby: I'd note that the use of `includeSubDomains` is quite risky and  I've had this catch me out before. Lots of devs will work on projects where they don't have control or the ability to enable https on all of an organisations sub-domains so this can be risky.

PR #6795, opened 2017-04-11
FIX Ensure that mysite test boostrap configuration is loaded after core and before the database connection

robbieaverill:
To reproduce the issue define `SS_SEND_ALL_EMAILS_TO="foo@example.com"` in your `.env` and run phpunit:

```
Notice: Undefined offset: -1 in .../framework/src/Core/Config/ConfigLoader.php on line 39
Notice: Undefined offset: -1 in .../framework/src/Core/Config/ConfigLoader.php on line 39
Fatal error: Call to a member function nest() on null in .../framework/src/Core/Config/Config.php on line 75
```

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -3,5 +3,4 @@
 require __DIR__ . '/bootstrap/init.php';
 require __DIR__ . '/bootstrap/cli.php';
 require __DIR__ . '/bootstrap/environment.php';
-require __DIR__ . '/bootstrap/mysite.php';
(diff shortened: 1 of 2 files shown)


tractorcow:
@robbieaverill that build failure is due to localisations in the recent release breaking the cms strings. This is fixed in another PR.

tractorcow:
@robbieaverill CMS module also has a bootstrap.php. it probably needs the same fix.

robbieaverill:
Re-triggering tests...

Issue #3420, opened 2014-08-21
search button for groups is meaningless... 


sunnysideup:
when you click on search in the group model admin in the CMS there is no field to search...  This should allow you to search for the group name:

![image](https://cloud.githubusercontent.com/assets/167154/3994480/cc52a404-291a-11e4-9ec7-37f6acb21558.png)

tractorcow:
You can set the summary_fields to Title, but then override the gridfield cell generation for that field to display the breadcrumbs.

SpiritLevel:
@kinglozzer @tractorcow @sunnysideup

Loz's solution (changing L97 and L100) works perfectly well. 

However, if we just delete lines 95-103 altogether then we get a filter input field for Group name and one for Description, and partial matching works for both as expected. Having a searchable description seems potentially useful. What do you think ? 

And, unless I am missing something, the lines 95-103 are completely superfluous anyways; breadcrumbs still display properly on the top left corner of Main and the grid field columns display with the correct heading. It seems that `$columns` is not used anywhere else in `SecurityAdmin`.

I propose we just delete lines 95-103 as the solution to this issue.

I quickly experimented with @tractorcow's idea of using `$summary_fields` but couldn't get anything to work. Maybe I am missing something here ? Or was this an idea for searching on subgroups ? 

I did notice one other minor quirk: if the `GridFieldFilterHeaders` are closed in all tabs (Users, Groups, Roles), then opening `GridFieldFilterHeaders` in one tab opens it in all tabs. However, when the `GridFieldFilterHeaders` are open in all tabs, closing it in one tab only closes it in that tab; it remains open in all other tabs. Should I submit an issue for this ?

@sunnysideup: There will always be at least one in the Administrators group, will there not (the Default Admin)? So I suppose there is always something to search on :-). 

Anyhow, let me know what you think. I'll submit a PR after consensus reached :)

sunnysideup:
@SpiritLevel , what I mean is that the search icons displays, but is not actually operational.  I guess I am wondering how the search icons display when there is no searchable field (as opposed to searchable rows). 

Of course, it is great to be able to search the groups, but in its current setup without a searchable field, it should look like this:

![image](https://cloud.githubusercontent.com/assets/167154/20652736/c1de455a-b563-11e6-9ea6-a48251ac1d02.png). There will always be gridfields that are "unsearchable" and so maybe the GridFieldFilterHeader should allow for this.

Does that make sense?

[...]

PR #6797, opened 2017-04-12
Fixing broken search in SecurityAdmin Groups field 

halkyon:
Breadcrumbs is not a database field, so can't be searched.

Before and after:

before
after

### DIFF
--- a/admin/code/SecurityAdmin.php
+++ b/admin/code/SecurityAdmin.php
@@ -94,10 +94,10 @@ class SecurityAdmin extends LeftAndMain implements PermissionProvider {
 		);
 		$columns = $groupList->getConfig()->getComponentByType('GridFieldDataColumns');
 		$columns->setDisplayFields(array(
-			'Breadcrumbs' => singleton('Group')->fieldLabel('Title')
+			'Title' => singleton('Group')->fieldLabel('Title')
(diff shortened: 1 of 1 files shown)


chillu:
Hmm, I don't think this is the right fix - groups can be nested, hence the breadcrumb display. We should rather remove the search functionality for this column.

kinglozzer:
Related: https://github.com/silverstripe/silverstripe-framework/issues/3420. Without significant work, it’s a choice between:

- No search at all
- Searching by group title. This won’t match the title of any parent groups in the breadcrumb (see https://github.com/silverstripe/silverstripe-framework/issues/3420#issuecomment-60502090)

tractorcow:
> Hmm, I don't think this is the right fix - groups can be nested, hence the breadcrumb display. We should rather remove the search functionality for this column.

The breadcrumbs are displayed. It is named "Title", meaning the backend search still searches on the Title field, but applies the breadcrumb nesting for user-visible value.

I should test this PR out but it does seem to solve the issue.

Making search work on the full chain (including parent groups) isn't necessary to fix this issue.

tractorcow:
It looks fine to me

image

halkyon:
🎉

PR #6798, opened 2017-04-13
API Replace SS_HOST with SS_BASE_URL

### LINKED ISSUE #6588: Better support for custom host/base url in CLI / env vars
Following on from #6337 we need to make sure CLI tasks can correctly work out their intended full URL.

Previously this was doing using `$_FILE_TO_URL_MAPPING` which has now been removed.

The environment variable `SS_HOST` has been added, but this only addresses the host name and not the protocol or base url (the sub directory that SS may be running in on a domain).

We currently have the ability to set `Director.alternate_base_url` as well as `alternate_protocol` and `alternate_host`. These can be used in combination but their effect isn't limited solely to CLI and this may be undesirable or inflexible on sites that require multiple domains to run.

Therefore, some environment variable(s) should be made available to provide a fallback when there is no other means to determine the host/protocol/base url.

---

Side note: it is currently possible to set the `alternate_base_url` to a full url (eg: https://example.com/base/). I feel this is confusing and we complicates our code and makes the behaviour less predictable. It may be worth deprecating this variable and replacing it with `base_path` as it appears the original intention was really only to set the base directory. It also leads to unexpected behaviours; if `alternate_protocol` is set, which protocol should be used - the one from the `alternate_base_url` or from `alternate_protocol`?

tractorcow:
API Remove Director::$test_servers / $dev_servers
API Remove MODULES_PATH / MODULES_DIR constants
ENHANCEMENT Injector backtick syntax now supports environment variables as well as constants
Fixes #6588

Also merge with https://github.com/silverstripe/silverstripe-installer/pull/171 to remove .htaccess declaration

CMS pr https://github.com/silverstripe/silverstripe-cms/pull/1794

### DIFF
--- a/docs/en/00_Getting_Started/03_Environment_Management.md
+++ b/docs/en/00_Getting_Started/03_Environment_Management.md
@@ -83,8 +83,8 @@ SilverStripe core environment variables are listed here, though you're free to d
 | `SS_TRUSTED_PROXY_PROTOCOL_HEADER` | Used to define the proxy header to be used to determine HTTPS status |
 | `SS_TRUSTED_PROXY_IP_HEADER` | Used to define the proxy header to be used to determine request IPs |
 | `SS_TRUSTED_PROXY_HOST_HEADER` | Used to define the proxy header to be used to determine the requested host name |
-| `SS_TRUSTED_PROXY_IPS` | IP address or CIDR range to trust proxy headers from |
+| `SS_TRUSTED_PROXY_IPS` | IP address or CIDR range to trust proxy headers from. If left blank no proxy headers are trusted. Can be set to 'none' (trust none) or '*' (trust all) |
 | `SS_ALLOWED_HOSTS` | A comma deliminated list of hostnames the site is allowed to respond to |
 | `SS_MANIFESTCACHE` | The manifest cache to use (defaults to file based caching). Must be a CacheInterface or CacheFactory class name |
 | `SS_IGNORE_DOT_ENV` | If set the .env file will be ignored. This is good for live to mitigate any performance implications of loading the .env file |
-| `SS_HOST` | The hostname to use when it isn't determinable by other means (eg: for CLI commands) |
+| `SS_BASE_URL` | The url to use when it isn't determinable by other means (eg: for CLI commands) |
--- a/src/Control/Director.php
+++ b/src/Control/Director.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\Control;
 
+use InvalidArgumentException;
 use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Config\Configurable;
@@ -54,62 +55,49 @@ class Director implements TemplateGlobalProvider
     const REQUEST = 'REQUEST';
 
     /**
+     * @config
      * @var array
      */
-    static private $urlParams;
+    private static $rules = array();
 
     /**
-     * @var array
-     */
-    static private $rules = array();
-
-    /**
+     * Set current page
+     *
+     * @internal
      * @var SiteTree
      */
     private static $current_page;
 
     /**
      * @config
-     *
      * @var string
      */
     private static $alternate_base_folder;
 
     /**
-     * @config
-     *
-     * @var array
-     */
-    protected static $dev_servers = array();
-
-    /**
-     * @config
-     *
-     * @var array
-     */
-    protected static $test_servers = array();
-
-    /**
-     * Setting this explicitly specifies the protocol ("http" or "https") used, overriding the normal
-     * behaviour of Director::is_https introspecting it from the request. False values imply default
-     * introspection.
+     * Force the base_url to a specific value.
+     * If assigned, default_base_url and the value in the $_SERVER
+     * global is ignored.
+     * Supports back-ticked vars; E.g. '`SS_BASE_URL`'
      *
      * @config
-     *
-     * @var string
-     */
-    private static $alternate_protocol;
-
-    /**
-     * @config
-     *
      * @var string
      */
     private static $alternate_base_url;
 
     /**
-     * @config
+     * Base url to populate if cannot be determined otherwise.
+     * Supports back-ticked vars; E.g. '`SS_BASE_URL`'
      *
+     * @config
+     * @var string
+     */
+    private static $default_base_url = '`SS_BASE_URL`';
+
+    /**
+     * Assigned environment type
+     *
+     * @internal
      * @var string
      */
     protected static $environment_type;
@@ -279,6 +267,7 @@ class Director implements TemplateGlobalProvider
 
         // These are needed so that calling Director::test() does not muck with whoever is calling it.
         // Really, it's some inappropriate coupling and should be resolved by making less use of statics.
+        $oldReadingMode = null;
         if (class_exists(Versioned::class)) {
             $oldReadingMode = Versioned::get_reading_mode();
         }
@@ -289,11 +278,11 @@ class Director implements TemplateGlobalProvider
         }
 
         if (!$session) {
-            $session = Injector::inst()->create('SilverStripe\\Control\\Session', array());
+            $session = Session::create([]);
         }
         $cookieJar = $cookies instanceof Cookie_Backend
             ? $cookies
-            : Injector::inst()->createWithArgs('SilverStripe\\Control\\Cookie_Backend', array($cookies ?: array()));
+            : Injector::inst()->createWithArgs(Cookie_Backend::class, array($cookies ?: []));
 
         // Back up the current values of the superglobals
         $existingRequestVars = isset($_REQUEST) ? $_REQUEST : array();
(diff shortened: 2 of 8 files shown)


src/Core/Injector/Injector.php
```diff
@@ -527,7 +527,13 @@ public function convertServiceProperty($value)
 
         // Evaluate constants surrounded by back ticks
         if (preg_match('/^`(?<name>[^`]+)`$/', $value, $matches)) {
-            $value = defined($matches['name']) ? constant($matches['name']) : null;
+            if (defined($matches['name'])) {
+                $value = constant($matches['name']);
+            } elseif (getenv($matches['name'])) {
```
sminnee: I think this should be a `!== false` check. getenv will return false if the var isn't set; if it's set to a blank it will return an empty string.

Tested with this:

```php
<?php
echo getenv("TEST") === false ? "F" : (getenv("TEST") === "" ? "Emptystring" : "Neither F nor emptystring");
```
Result:
```
> env TEST= php test.php 
Emptystring
> env TEST=false php test.php 
Neither F nor emptystring
> php test.php 
F
> env TEST=hello php test.php 
Neither F nor emptystring

Comment (sminnee) [inline src/Core/Injector/Injector.php]: I also think that env vars should take precedence over constants, so put that code first.

**Comment by tractorcow** [comment]
@sminnee updated exactly as requested. :) Good feedback thank you!

File: src/Core/Injector/Injector.php (hunk 1 above)
Comment (tractorcow) [inline src/Core/Injector/Injector.php]: Sounds good to me.

**Comment by tractorcow** [comment]
@dhensby as an aside, I removed many global variables from Core.php by using closures for generating default constant values. :)

File: src/Control/Director.php [hunk 2]
```diff
@@ -647,64 +653,78 @@ public static function protocol()
      */
     public static function is_https()
     {
+        // Check override from alternate_protocol
+        if ($protocol = self::config()->get('alternate_protocol')) {
```
sminnee: I think we should remove alternate_protocol in favour of alternate_base_url.

src/Control/Director.php
```diff
@@ -76,40 +80,47 @@ class Director implements TemplateGlobalProvider
...
     /**
-     * @config
+     * Setting this overrides the host name
      *
-     * @var array
+     * @config
+     * @var string
      */
-    protected static $test_servers = array();
+    private static $alternate_host;
```
sminnee: I think we should remove alternate_host in favour of alternate_base_url.

Note that this doesn't need to be logged as an API change as it was introduced in 4.

src/Control/Director.php
```diff
@@ -76,40 +80,47 @@ class Director implements TemplateGlobalProvider
     private static $alternate_base_folder;
```
sminnee: I think we should remove alternate_base_folder in favour of alternate_base_url.

Note that this doesn't need to be logged as an API change as it was introduced in 4.

tractorcow: folders are urls are different though. We need to set a base folder when mocking a subdirectory, which doesn't mean we want to change the base url.

src/Control/Director.php
tractorcow: ok.

src/Control/Director.php
tractorcow: Sure.

src/Control/Director.php
dhensby: Having these separate configs makes things confusing unless we don't use this base_folder to construct a URL...

docs/en/00_Getting_Started/03_Environment_Management.md
```diff
@@ -83,8 +83,8 @@ SilverStripe core environment variables are listed here, though you're free to d
...
 | `SS_TRUSTED_PROXY_IP_HEADER` | Used to define the proxy header to be used to determine request IPs |
 | `SS_TRUSTED_PROXY_HOST_HEADER` | Used to define the proxy header to be used to determine the requested host name |
-| `SS_TRUSTED_PROXY_IPS` | IP address or CIDR range to trust proxy headers from |
+| `SS_TRUSTED_PROXY_IPS` | IP address or CIDR range to trust proxy headers from. If left blank no proxy headers are trusted. Can be set to 'none' (trust none) or '*' (trust all) |
 | `SS_ALLOWED_HOSTS` | A comma deliminated list of hostnames the site is allowed to respond to |
 | `SS_MANIFESTCACHE` | The manifest cache to use (defaults to file based caching). Must be a CacheInterface or CacheFactory class name |
 | `SS_IGNORE_DOT_ENV` | If set the .env file will be ignored. This is good for live to mitigate any performance implications of loading the .env file |
-| `SS_HOST` | The hostname to use when it isn't determinable by other means (eg: for CLI commands) |
+| `SS_BASE_URL` | The url to use when it isn't determinable by other means (eg: for CLI commands). Protocol is optional. E.g. `//mysite.com/base-path` or `https://mysite.com/base-path` |
```
sminnee: I'm not sure that protocol should be optional as I don't see what the behaviour would be in CLI situations.

src/Control/Director.php
```diff
@@ -76,40 +80,47 @@ class Director implements TemplateGlobalProvider
...
+     * If assigned, default_base_url and the value in the $_SERVER
+     * global is ignored.
+     * Supports back-ticked vars; E.g. '`SS_BASE_URL`'
      *
      * @config
-     *
      * @var string
      */
-    private static $alternate_protocol;
+    private static $alternate_base_url;
```
sminnee: Default base URL is only used when the URL can't be inferred from the request, e.g. in a CLI call. This ensures that websites can function on >1 domain.

Alternate base URL is used in all cases and is intended for cases like multi-server static publisher setups and subsites where you want to deliberately override.

Admittedly it could be tidier eg we could have a Boolean "force_base_url" flag that switches to the latter behaviour.

src/Control/Director.php
sminnee: Maybe I misunderstood when I commented - is this the local file system path rather than a portion of the URL? If that's the case then I agree it can be included

tractorcow: > unless we don't use this base_folder to construct a URL...

We don't. :)

```php
/**
 * Returns the root filesystem folder for the site. It will be automatically calculated unless
 * it is overridden with {@link setBaseFolder()}.
 *
 * @return string
 */
public static function baseFolder()
{
    $alternate = Director::config()->uninherited('alternate_base_folder');
    return ($alternate) ? $alternate : BASE_PATH;
}
```

Nothing to do with URLS.

docs/en/00_Getting_Started/03_Environment_Management.md
tractorcow: The site could have an SSL and a non-ssl base url, in which case you may not want to fix it to one or the other.

src/Control/Director.php
```diff
@@ -566,56 +565,63 @@ public static function absoluteURL($url, $relativeParent = self::BASE)
...
+        // Check base url
+        if ($baseURL = self::config()->uninherited('default_base_url')) {
+            $baseURL = self::resolveSubstitutions($baseURL);
+            if (strstr($baseURL, '//')) {
+                return parse_url($baseURL, PHP_URL_HOST);
+            }
         }
 
+        // Legacy ss4-alpha environment var
+        $legacyHostname = getenv('SS_HOST');
```
tractorcow: Ok.

tractorcow: I disagree; It will mean that users who have just upgraded to alpha6 will have additional pains on this alpha7. We could remove this at stable, but removing it now makes onboarding/upgrading more frustrating to users. I want to make the adoption process as easy and push button for these users as possible. I can mark it as deprecated for removal at 4.0 beta, but I'd rather not immediately break it.

src/Control/Director.php
sminnee: 👍

docs/en/00_Getting_Started/03_Environment_Management.md
sminnee: Sure, but the sysadmin needs to pick one to use when there's no information available — the primary use-case of this setting is for generating absolute URLs from CLI-scripts and cronjobs.

Having the absolute URLs sent out in emails, etc, as "<a href="//silverstripe.com>" doesn't seem like a good idea, does it?

src/Control/Director.php
```diff
@@ -76,40 +80,47 @@ class Director implements TemplateGlobalProvider
...
+     * Setting this explicitly specifies the protocol ("http" or "https") used, overriding the normal
+     * behaviour of Director::is_https introspecting it from the request
+     *
      * @config
      *
-     * @var array
+     * @var string
      */
-    protected static $dev_servers = array();
+    private static $alternate_protocol;
```
sminnee: 👍

src/Control/Director.php
sminnee: Can you put `// @todo delete this before tagging 4.0.0-beta1` on this then? :)

docs/en/00_Getting_Started/03_Environment_Management.md
tractorcow: Fair enough, I agree. :)

tractorcow:
Updated with all the above requested changes. I've also found some more Director API that can be removed ($urlParams).

sminnee:
The CMS failure is non-benign. It appears as though `Director::baseURL()` is returning an absolute rather than relative URL in the test case. :-/

tractorcow:
CMS fix at https://github.com/silverstripe/silverstripe-cms/pull/1794

Issue #6788, opened 2017-04-08
[bug] Multiple login methods impossible


Firesphere:
In the template `Security_MultiAuthenticatorLogin`, the method h`$AuthenticatorName` is called, which returns null.

This method isn't called when there is only one LoginForm, this seems to be related to the creation of the tabset?

Firesphere:
I'm working on this by the way. Would it be an asset to have a LoginForm::getTitle method? That would make the process a lot easier!

robbieaverill:
@Firesphere it's worth mentioning that I've just looked at something similar with the active directory module in that `LoginForm` instances now implement a `buildRequestHandler` method which returns an instance of a `FormRequestHandler` class, and that tends to be where processing logic is stored. This isn't very well covered in the changelogs. See the login form and authenticator changes [in this PR](https://github.com/silverstripe/silverstripe-activedirectory/pull/100).

The framework still doesn't allow you to have multiple login forms/authenticators at once though:

```
Fatal error: Class name must be a valid object or a string in .../framework/src/Security/LoginForm.php on line 55
```

Firesphere:
This has been fixed in #6800

PR #6800, opened 2017-04-14
[bug] Pass the AuthenticationMethod to the controller

Firesphere:
Resolves issue #6788
The AuthenticationMethod is passed in via hidden field as per usual, but due to changes, the fallback authenticator was always the MemberAuthenticator and the actual passed in authenticator was defaulting to an empty string.
This causes an issue when there are multiple authenticators and the default authenticator is _not_ in the allowed authenticators, but is still the default. It caused the getAuthenticator method to return the default MemberAuthenticator to be returned, despite it being disabled.
A second issue around multiple authenticators, was the template using a no-longer used method `getAuthenticatorName`. This method returned a null on the default MemberLoginForm (as nothing was set), causing a Warning.
Because the getAuthenticator and getAuthenticatorName in the Authenticator are no longer in use, I've opted to replace these with a translatable string `getAuthenticatorName` in the Form, to display the title of the form on the tabs, as per the tabset on Security_MultiAuthenticatorLogin template.

### DIFF
--- a/src/Security/Authenticator.php
+++ b/src/Security/Authenticator.php
@@ -79,15 +79,6 @@ abstract class Authenticator extends Object
     }
 
 
-    /**
-     * Get the name of the authentication method
-     *
-     * @return string Returns the name of the authentication method.
-     */
-    public static function get_name()
-    {
-    }
-
     public static function register($authenticator)
     {
         self::register_authenticator($authenticator);
--- a/src/Security/CMSMemberLoginForm.php
+++ b/src/Security/CMSMemberLoginForm.php
@@ -79,4 +79,12 @@ class CMSMemberLoginForm extends LoginForm
     {
         return CMSMemberLoginHandler::create($this);
     }
+
+    /**
+     * @return string
+     */
+    public function getAuthenticatorName()
+    {
+        return _t('CMSMemberLoginForm.AUTHENTICATORNAME', 'CMS Member Login Form');
+    }
 }
--- a/src/Security/LoginForm.php
+++ b/src/Security/LoginForm.php
@@ -23,31 +23,13 @@ abstract class LoginForm extends Form
      * form.
      * @var string
      */
-
     protected $authenticator_class;
 
     /**
-     * Get the authenticator instance
-     *
-     * @return Authenticator Returns the authenticator instance for this login form.
+     * Return the title of the form for use in the frontend
+     * For tabs with multiple login methods, for example.
+     * This replaces the old `get_name` method
+     * @return string
      */
-    public function getAuthenticator()
-    {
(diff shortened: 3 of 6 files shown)


Firesphere:
I'm working on updating the documentation around adding/removing authenticators. As there is currently not much available, it'll be a separate branch with a set of updated docs.

sminnee:
If we're going to stop using `Authenticator::get_name()` then I think we should get rid of it altogether — having this configuration in two places is a bad idea. The docs will need to be updated with this API change.

I think that `LoginForm::getTitle()` is too generic a method for returning "short name of the authentication method as presented to a user" — the risk of conflict with some other feature of forms seems high enough to worry about. I'd probably with `LoginForm::getAuthenticatorTitle()`

_config/security.yml
```diff
@@ -0,0 +1,2 @@
+SilverStripe\Security\MemberLoginForm:
```
sminnee: This is a bidirectional coupling between MemberLoginForm and MemberAuthenticator and that feels messy.

Since MemberAuthenticator is responsible for constructing MemberLoginForm, I think it would be better if this became a constructor arg to MemberLoginForm, and `MemberAuthenticator::get_login_form()` can pass `self::class` as an argument.

Firesphere:
Agreed on both, except I feel "AuthenticatorTitle" is referring to the authenticator, and my gut tells me it's the _Form_ that has the title. I'm okay with either, but prefer `getAuthenticatorformTitle`?

Just to be sure, get_name() isn't used anywhere except in the template, where it breaks because it's missing currently from the MemberAuthenticator.

Firesphere:
There is, by the way, currently _no_ documentation on the `get_name` method. Would you mind if I leave the documentation out for now, and write a more extensive documentation covering Multiple Auth methods etc. in a separate commit?

sminnee:
> Agreed on both, except I feel "AuthenticatorTitle" is referring to the authenticator, and my gut tells me it's the Form that has the title. I'm okay with either, but prefer getAuthenticatorformTitle?

Right now it's called getAuthenticatorName — this is the existing API that login templates will make use of. I don't see a good reason to change it, so I think we should keep it as getAuthenticatorName.

Firesphere:
Already withdrew my comment and updated. ;)

Firesphere:
Scrutinizer is errorring for unknown reasons, enlighten me?

Firesphere:
Requested changes and docblocks implemented. Scrutinizer failure for unknown reasons though.

Firesphere:
The switch from configuration to using `self::class` resolves one of my biggest issues by the way. I was looking for a way to disable bypassing the set Authenticator (as in my original description). Using self::class in the code both prevents as well as secures that issue.

Ow Captain Hindsight, sometimes you're nice :)

sminnee:
Merged and updated the changelog in 9b1baa9503126607e57683713aaa10c13252c264

PR #6801, opened 2017-04-14
Improve the default Logins

Firesphere:
- MemberLoginForm has the fields and actions extracted to a separate method, so it's more easily overridable
- Moved the global variable $_REQUEST to getting the info from the controller
- Updated string variables to `::class`
- Updated RequiredFields to be a static, so it's overridable/updatable from either Config or code

### DIFF
--- a/src/Security/MemberLoginForm.php
+++ b/src/Security/MemberLoginForm.php
@@ -36,6 +36,12 @@ class MemberLoginForm extends LoginForm
     public $loggedInAsField = 'FirstName';
 
     /**
+     * Required fields for validation
+     * @var array
+     */
+    private static $required_fields;
+
+    /**
      * Constructor
      *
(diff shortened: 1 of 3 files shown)


src/Security/MemberLoginForm.php
```diff
@@ -135,16 +109,69 @@ public function __construct(
 
         parent::__construct($controller, $name, $fields, $actions);
 
-        $this->setValidator(RequiredFields::create('Email', 'Password'));
+        $this->setValidator(RequiredFields::create(self::$required_fields));
```
sminnee: If this is going to be a config-overridable let's access it with the config system.

Firesphere:
I think Scrutinizer is broken. Changes implemented though, I should've done it that way in the first place.

PR #6802, opened 2017-04-16
[WIP][feature] Log queries to a logfile

Firesphere:
This feature enables logging the queries to a file. Although this has an impact on performance, it's only enabled in `dev` mode anyway, so there are no big issues there.

Reason for this change, is I found it incredibly useful to be able to log my queries to a file and compare them to MySQL's `slow_queries` log and by some deduction, improve performance of the whole.

I've opted to use inline and log to both try and make the best of the queries, because otherwise, the log file is slightly unreadable after the logging.

The added documentation contains a suggested use of `time()`

Also updated the output to have a precision of 5, as per the microtime() call, instead of the original 4.

I'm happy to add tests for this, but it as it is only used in dev-mode, I'm unsure if the tests actually run (hence, no tests yet)

todo: use the logger instead of manually writing

### DIFF
--- a/docs/en/02_Developer_Guides/07_Debugging/02_URL_Variable_Tools.md
+++ b/docs/en/02_Developer_Guides/07_Debugging/02_URL_Variable_Tools.md
@@ -38,9 +38,19 @@ Append the option and corresponding value to your URL in your browser's address 
 ## Database
 
  | URL Variable | | Values | | Description                                                                                  | 
- | ------------ | | --------- | | -----------                                                                                  | 
- | showqueries  | | 1\|inline | | List all SQL queries executed, the `inline` option will do a fudge replacement of parameterised queries          | 
- | previewwrite | | 1      | | List all insert / update SQL queries, and **don't** execute them.  Useful for previewing writes to the database. | 
+ | ------------ | | -------------- | | -----------                                                                                  | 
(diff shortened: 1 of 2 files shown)


Firesphere:
Just to whoever is reviewing this, Scrutinizer is bugging out, it's not an actual failure of code. Slightly annoying.

sminnee:
I don't think that the ORM should be responsible for log-file management.

Can we instead structure this as the ORM does DEBUG-level or TRACE-level PSR-3 logging, and then that can be sent to a logfile instead?

PSR-3 all the things, plz!

sminnee:
If there was a way of getting your PSR-3 logs sent to the chrome dev console, that'd be really nice too. Presumably that part would be a module and not part of core.

sminnee:
If you want to make it easier for people to send these logs somewhere other than the main ones, you could put a dot-suffix on the service name:

```php
Injector::inst()->get(LoggerInterface::class . '.ORM')->debug($message)
```

This will default to the main log destination unless someone specifically defines a separate `Psr\Log\LoggerInterface.ORM` service.

sminnee:
Having dug into this more, #6806 might be a better way of attacking this...

Firesphere:
Absolutely agreed (and admitted, this initial mock-up is dirty and not core-worthy, but it was more to get a thought back and maybe update.)
Closing in favor of #6806

PR #6804, opened 2017-04-17
Use `Config` for authenticator settings

Firesphere:
Unless there's a specific reason for not using `Config`? 
This change reads the authentication methods etc. from config instead of static.

### DIFF
--- a/src/Security/Authenticator.php
+++ b/src/Security/Authenticator.php
@@ -4,6 +4,7 @@ namespace SilverStripe\Security;
 
 use SilverStripe\Core\Object;
 use SilverStripe\Control\Controller;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\Form;
 
 /**
@@ -22,7 +23,7 @@ abstract class Authenticator extends Object
      *
      * @var array
      */
-    private static $authenticators = array(MemberAuthenticator::class);
+    private static $authenticators = [];
 
     /**
      * Used to influence the order of authenticators on the login-screen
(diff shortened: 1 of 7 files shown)


src/Security/Authenticator.php
```diff
@@ -117,9 +117,10 @@ public static function register_authenticator($authenticator)
             return false;
         }
 
-        if (in_array($authenticator, self::$authenticators) == false) {
-            if (call_user_func(array($authenticator, 'on_register')) === true) {
-                array_push(self::$authenticators, $authenticator);
+        $authenticators = self::config()->get('authenticators');
+        if (in_array($authenticator, $authenticators) == false) {
+            if ($authenticator::on_register() === true) {
+                self::config()->merge('authenticators', [$authenticator]);
```
tractorcow: Any time that configs are modified, you immediately lose a level of performance, since the immutable config collection needs to be cloned before update.

Config should not be used as a cache; Can we put this somewhere else?

Firesphere: Yep, I added this more as a fallback so it wouldn't immediately break any existing `register()` calls from other code, but the idea is to deprecate it and move to `Config` only. I'm in that regard slightly concerned about the `unregister` method though.

tractorcow: Let's deprecate unregister, and move to config-only registration? That way we can add a cache on top of config rather than through it.

Firesphere: How would you unregister a login method then? If I want my custom method applied and not use MemberAuthenticator anymore, how do I unregister via YML?

tractorcow: Remove the default authenticator from the Authenticators config, and then make `get_authenticators` conditionally return the list of authenticators (which is empty by default), but return a single-length default authenticator array if empty. That way as soon as you add a single item to authenticators, it disables the default.

tractorcow:
```php
/**
     * This variable holds all authenticators that should be used.
     * If left empty, will fail over to default_authenticator. Add at least one item to disable the default authenticator.
     *
     * @var array
     */
    private static $authenticators = [];

    /**
     * Used to influence the order of authenticators on the login-screen
     * (default shows first).
     *
     * @var string
     */
    private static $default_authenticator = MemberAuthenticator::class;
```

Firesphere: Right, yep, that makes sense.

Leave the old "set/unset" methods in with a 4.0 deprecation notice, and update the docs to match the new Config.

Note to self: Add a check if the returned Authenticator in get_authenticators is actually a subclass of Authenticator.

Firesphere:
Follow up, Authenticator:204 and following. How often are those methods actually implemented? Is it worth cleaning up?

tractorcow:
Yes, we should remove any PHP methods that directly alter config. Users can use the config api instead.

Firesphere:
Okay, updated with the config, added the abstract methods from the previous PR and unified it across the LoginForm, MemberLoginForm and CMSMemberLoginForm.
Cleaned up the unused on/un register methods (they were awkwardly unused and applied nowhere and simply returned true)
Updated the tests to reflect these changes. Awaiting Travis.

Firesphere:
For updates on the documentation, please see here: https://github.com/Firesphere/silverstripe-framework/tree/pulls/authenticator_docs
I didn't want to do them on the same branch, as by switching, I might get them confused and do the wrong docs. When all changes are merged or rejected, I'll finish up that branch for final documentation

Firesphere:
Update: I have looked through all of my history of custom Authenticators, and not once have I actually encountered the use of `on_register` or `un_register`. I will document the removal in my documentation-branch

PR #6810, opened 2017-04-19
suffix subfolder in silverstripe-chache with php-version

lerni:
If the cache-folder is in system-tmp php-version is added to the path
https://github.com/silverstripe/silverstripe-framework/pull/3882

...but not, if silverstripe-cache is in webroot
https://github.com/silverstripe/silverstripe-framework/issues/2986

something like that would prevent a crash if minor php-version changes without ?flush and silverstripe-cache is in webroot and would also solve the CLI vs. HTTP PHP-Version dilemma in those cases.

### DIFF
--- a/core/TempPath.php
+++ b/core/TempPath.php
@@ -11,8 +11,9 @@
 function getTempFolder($base = null) {
 	$parent = getTempParentFolder($base);
 
-	// The actual temp folder is a subfolder of getTempParentFolder(), named by username
-	$subfolder = $parent . DIRECTORY_SEPARATOR . getTempFolderUsername();
+	// The actual temp folder is a subfolder of getTempParentFolder(), named by username and suffixed with currently used php-version
+	$phpversion = '-php' . preg_replace('/[^\w-\.+]+/', '-', PHP_VERSION);
+	$subfolder = $parent . DIRECTORY_SEPARATOR . getTempFolderUsername() . $phpversion;
 
 	if(!@file_exists($subfolder)) {
 		mkdir($subfolder);
--- a/core/TempPath.php
+++ b/core/TempPath.php
@@ -65,7 +65,7 @@ function getTempParentFolder($base = null) {
 
 	// failing the above, try finding a namespaced silverstripe-cache dir in the system temp
 	$tempPath = sys_get_temp_dir() . DIRECTORY_SEPARATOR .
-		'silverstripe-cache-php' . preg_replace('/[^\w-\.+]+/', '-', PHP_VERSION) .
+		'silverstripe-cache' .
 		str_replace(array(' ', '/', ':', '\\'), '-', $base);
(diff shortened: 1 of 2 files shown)


robbieaverill:
Hey @Lerni - would you mind squashing your commits and pushing a single, meaningful commit message up?

lerni:
@robbieaverill I've tried but it seams I'm stucked with git for this. In https://github.com/silverstripe/silverstripe-framework/pull/6810/commits/16767109a7cb55846e519aeb9708ee1aa497abce is already everything in. If this is acceptable, can't you just "squash on merge"?

core/TempPath.php
```diff
@@ -11,8 +11,9 @@
 function getTempFolder($base = null) {
 	$parent = getTempParentFolder($base);
 
-	// The actual temp folder is a subfolder of getTempParentFolder(), named by username
-	$subfolder = $parent . DIRECTORY_SEPARATOR . getTempFolderUsername();
+	// The actual temp folder is a subfolder of getTempParentFolder(), named by username and suffixed with currently used php-version
+	$phpversion = '-php' . preg_replace('/[^\w-\.+]+/', '-', PHP_VERSION);
+	$subfolder = $parent . DIRECTORY_SEPARATOR . getTempFolderUsername() . $phpversion;
```
sminnee: In `/tmp` this will end up including the PHP version twice.

I think you should remove the PHP version logic from `getTempParentFolder()`

lerni: @sminnee - done. I promise 'll run test locally next time. Sorry for the commit-mess.

tractorcow: I suggest to use DIRECTORY_SEPARATOR between silverstripe-cache and $phpversion. This is necessary so that we can have a parent (and predictable) folder name which can be controlled by a hard-coded `.htaccess` rewrite rule. If you make this variable then it's riskier to secure this folder.

lerni:
the path looks like:

in webroot:
/silverstripe-cache/USER-php7.1.5/

in tmp:
/tmp/silverstripe-cache-Users-FOO-webroot-BAR/USER-php7.1.5

@tractorcow .htaccess is needed for webroot but not for tmp and it's cascading. Wouldn't /silverstripe-cache/ be the place to put it?

tractorcow:
@lerni that looks good to me.

@sminnee can you confirm you are happy for this to get merged?

sminnee:
I note that the temp path is going to change. I think this is fine for the 3 branch as I don't think that's an API breakage. However, I wouldn't put it into a patch release.

Happy for it to be squashed in a single commit and merge.

sminnee:
Done!

lerni:
PR as composer-patch:
[https://github.com/lerni/silverstripe3-cache-folder-php-version-prefix](https://github.com/lerni/silverstripe3-cache-folder-php-version-prefix)

PR #6817, opened 2017-04-20
FIX: Add PHP extension requirements to composer.

sminnee:
This will raise incompatibilities sooner rather than relying on the 
installer.

GD is excluded from this as it’s a requirement of silverstripe/assets

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,6 @@
     }
   ],
   "require": {
-    "php": ">=5.6.0",
     "composer/installers": "~1.0",
     "embed/embed": "^3.0",
     "league/flysystem": "~1.0.12",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -29,7 +28,18 @@
     "symfony/config": "^2.8",
     "symfony/translation": "^2.8",
     "symfony/yaml": "~2.7",
-    "vlucas/phpdotenv": "^2.4"
+    "vlucas/phpdotenv": "^2.4",
+
```
tractorcow: just a heads up, whitespaces in composer.json will get smashed by auto-rewrite tools (cow is one of them) :)

sminnee:
Robbie and Chris approved so I merged

PR #6818, opened 2017-04-20
js error fixed in admin when selects from the drop down fields in model admin search

priyashantha:
Error was `uncaught exception: cannot call methods on tabs prior to initialization` 

Tabs must be initialized before you can work on them


kinglozzer:
Duplicate of https://github.com/silverstripe/silverstripe-framework/pull/5920, sorry @priyashantha 😅

PR #6824, opened 2017-04-21
DOCS Corrected logger documentation

micmania1:


flamerohr:
I've found references to `Injector: Logger:` in 05_Injector.md that do not use the FQN. pinging @tractorcow if that would be relevant to have fixed for this pull request

PR #6825, opened 2017-04-21
Don't fail dev/build without phpunit

chillu:
When you install a SilverStripe project with "composer install --no-dev",
the PHPUnit dependency gets skipped. Which means the PHPUnit_Framework_TestListener
interface doesn't exist. The SilverStripe Classloader might still include
SapphireTestReporter which relies on this interface, which then breaks execution.

SS3 fixed this by NOT defining the class in the first place.
This has been removed in https://github.com/silverstripe/silverstripe-framework/commit/2fdc96a0de33c34b4dbcdeb971cd4ccf319d64a0#diff-82b3f89e8e5ae090c93e9c3a2ba8aa36L3,
as part of a PHPUnit version upgrade - but without an apparent fix to replace this.


chillu:
Heh fixed now

sminnee:
Can we delete this test reporter altogether? My understanding Ian that @dhensby removed the need for this in SS4

chillu:
Very good point @sminnee - I've removed the `PhpUnitWrapper` which would instantiate these reporters a while ago: https://github.com/silverstripe/silverstripe-framework/commit/a16588aac3b5b3ea0228d13f5c1f9cf437ac9582

Updated the PR to remove both remaining reporters. Note that I'm also removing `TeamCityListener` with https://github.com/silverstripe/silverstripe-framework/pull/6823

dhensby:
awwwwwww yess

PR #6832, opened 2017-04-23
add requiredField class to field holder of required fields

xini:
can be used to style required fields and their labels differently

### DIFF
--- a/templates/forms/FormField_holder.ss
+++ b/templates/forms/FormField_holder.ss
@@ -1,4 +1,4 @@
-<div id="$HolderID" class="field<% if $extraClass %> $extraClass<% end_if %>">
+<div id="$HolderID" class="field<% if $extraClass %> $extraClass<% end_if %><% if $Required %> requiredField<% end_if %>">
 	<% if $Title %><label class="left" for="$ID">$Title</label><% end_if %>
 	<div class="middleColumn">
 		$Field
--- a/templates/forms/FormField_holder.ss
+++ b/templates/forms/FormField_holder.ss
@@ -1,4 +1,4 @@
-<div id="$HolderID" class="field<% if $extraClass %> $extraClass<% end_if %><% if $Required %> requiredField<% end_if %>">
+<div id="$HolderID" class="field<% if $extraClass %> $extraClass<% end_if %>">
 	<% if $Title %><label class="left" for="$ID">$Title</label><% end_if %>
 	<div class="middleColumn">
 		$Field
--- a/forms/FormField.php
+++ b/forms/FormField.php
@@ -557,6 +557,10 @@ class FormField extends RequestHandler {
 			$classes[] = 'nolabel';
 		}
 
+		if($this->Required()) {
+			$classes[] = 'requiredField';
+		}
+
 		// Allow custom styling of any element in the container based on validation errors,
 		// e.g. red borders on input tags.
 		//


templates/forms/FormField_holder.ss
```diff
@@ -1,4 +1,4 @@
-<div id="$HolderID" class="field<% if $extraClass %> $extraClass<% end_if %>">
+<div id="$HolderID" class="field<% if $extraClass %> $extraClass<% end_if %><% if $Required %> requiredField<% end_if %>">
```
robbieaverill: This could also be achieved by adding an extension on FormField that adds `requiredField` as an extra CSS class via `if ($formField->Required()) { $formField->addExtraClass('requiredField'); }`.

If this was going to be made a default CSS class for form fields I think it should probably be done in `FormField::extraClass` - there's an example there already of it adding `nolabel` when there's no title on the field.

xini:
Hi Robbie,
I don't mind. Probably makes sense, yes. Here you go.
Cheers, Flo

xini:
Can you please merge this in? Thanks.

tractorcow:
Please defer to @clarkepaul for approval of this. Is this going to interfere with any of our bootstrap styling work?

dhensby:
I would much rather leave this to devs to customise their own templates rather than us increase coupling and custom class names. these internal classnames are already pretty bad as it is...

tractorcow:
Yeah, I think given @robbieaverill and @dhensby 's comments I'll close this.

xini:
I strongly disagree. @robbieaverill's concern has been addressed and the code changed accordingly. 
In userforms we have a similar functionality in https://github.com/silverstripe/silverstripe-userforms/blob/4/code/model/editableformfields/EditableFormField.php#L850 and this PR is porting this in for other forms. Please reconsider.

tractorcow:
That example you linked us to shows another example of how to achieve the same solution by calling `$field->addExtraClass('requiredField');` on the field directly.

You could also add an extension to form field and implement updateAttributes() to update `class` property with the extra classes.

If you like, I would accept a change to FormField::extraClass() method with the below code.

```php
$this->extend('updateExtraClass', $classes);
```

So you could update that property directly.

Would that better suit your extension requirements?

xini:
Damian. Why do visually impaired users get all the bells and whistles to make a required field accessible to them, but to add an * to a label to make it visually stand out, a simple class is missing?!
Of course we can add an extension hook and keep patching the patches. But maybe it would be a better solution if the field took care of the validator properly. 
also, why have it in userforms and not other forms by default?

dhensby:
@xini I've achieved this easily on many projects by amending the template. Having this class injected would only make my life harder as a dev as I'd have class pollution and potentially conflict and no one needs that.

The best thing to do... is create your own templates

xini:
@dhensby I know. And you pretty much prove my point. You do it, I do it, probably everyone does it. In every project. 
userform does it out of the box which makes the experience inconsistent.
And re class conflicts: how ambiguous is a class 'requiredField' or 'required' on a form field?

dhensby:
For my projects it's an inappropriate class name because it didn't confirm to the CSS conventions I use

xini:
so that means you can't use userforms. great.

xini:
@dhensby what class name would you suggest to get this merged? I'm happy to submit a PR to userforms as well if you want the class to be consistent.

tractorcow:
I mentioned above I would accept an extension point so you could easily add this functionality to your own project.

> If you like, I would accept a change to FormField::extraClass() method with the below code.
> `$this->extend('updateExtraClass', $classes);`
> So you could update that property directly.

Issue #6835, opened 2017-04-24
CLI errors on live environments using friendly, rather than detailed, view


sminnee:
The selection of error reporter is now controlled by "only" / "except" settings on environment type in _config/logging.yml

This means that if CLI is used on a live environment, the DebugViewFriendlyErrorFormatter handles the errors, which supresses the error information from being shared on the command-line.

This a regression, and, for example, prevents diagnostic data from being sent in deploy scripts. It appears to have been introduced recently only because the "environment" flag in "only" was previously broken.

The detailed error formatter should be configured to display on either CLI or dev mode environments.

PR #6836, opened 2017-04-24
FIX: Show detailed errors on CLI for live environments

sminnee:
API: Add HTTPOutputHandler::setCliFormatter

Fixes https://github.com/silverstripe/silverstripe-framework/issues/6835

This provides detailed errors (but not warnings or notices) in CLI calls
on live environments.

It does this by adding a 2nd argument to our output handler,
CliFormatter. This formatter will be used when Director::is_cli() is
true.

### DIFF
--- a/_config/logging.yml
+++ b/_config/logging.yml
@@ -37,8 +37,9 @@ SilverStripe\Core\Injector\Injector:
     constructor:
       - "error"
     properties:
-      Formatter: %$FriendlyErrorFormatter
-  FriendlyErrorFormatter:
+      Formatter: %$SilverStripe\Logging\DebugViewFriendlyErrorFormatter
+      CLIFormatter: %$SilverStripe\Logging\DetailedErrorFormatter
+  SilverStripe\Logging\DebugViewFriendlyErrorFormatter:
     class: SilverStripe\Logging\DebugViewFriendlyErrorFormatter
     properties:
       Title: "There has been an error"
--- a/src/Logging/HTTPOutputHandler.php
+++ b/src/Logging/HTTPOutputHandler.php
@@ -3,8 +3,10 @@
 namespace SilverStripe\Logging;
 
 use SilverStripe\Control\Controller;
+use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPResponse;
 use Monolog\Handler\AbstractProcessingHandler;
+use Monolog\Formatter\FormatterInterface;
 
 /**
  * Output the error to the browser, with the given HTTP status code.
@@ -24,6 +26,11 @@ class HTTPOutputHandler extends AbstractProcessingHandler
     private $statusCode = 500;
 
     /**
+     * @var FormatterInterface
+     */
+    private $cliFormatter = null;
+
+    /**
      * Get the mime type to use when displaying this error.
      *
      * @return string
@@ -38,7 +45,7 @@ class HTTPOutputHandler extends AbstractProcessingHandler
      * Default text/html
      *
      * @param string $contentType
-     * @return $this
+     * @return HTTPOutputHandler Return $this to allow chainable calls
      */
     public function setContentType($contentType)
     {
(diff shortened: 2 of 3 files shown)


src/Logging/HTTPOutputHandler.php
```diff
@@ -70,6 +72,38 @@ public function setStatusCode($statusCode)
     }
 
     /**
+     * Set a formatter to use if Director::is_cli() is true
```
robbieaverill: These methods could use some PHPDoc params etc

src/Logging/HTTPOutputHandler.php
```diff
@@ -70,6 +72,38 @@ public function setStatusCode($statusCode)
...
+     */
+    public function setCliFormatter(FormatterInterface $cliFormatter)
+    {
+        $this->cliFormatter = $cliFormatter;
```
robbieaverill: Is the property undefined at this point?

src/Logging/HTTPOutputHandler.php
```diff
@@ -70,6 +72,38 @@ public function setStatusCode($statusCode)
...
+        return $this->cliFormatter;
+    }
+
+    /**
+     * Return the formatter to use in this case.
+     * May be the getCliFormatter() value if one is provided and Director::is_cli() is true.
+     */
+    public function getFormatter()
+    {
+        if (Director::is_cli() && $this->cliFormatter) {
```
chillu: Shouldn't this use the getter?

src/Logging/HTTPOutputHandler.php
chillu: Ditto, they all need PHPDoc

chillu:
Can you please update the docs at https://github.com/silverstripe/silverstripe-framework/blob/master/docs/en/02_Developer_Guides/07_Debugging/01_Error_Handling.md?

src/Logging/HTTPOutputHandler.php
```diff
@@ -70,6 +72,38 @@ public function setStatusCode($statusCode)
     }
 
     /**
+     * Set a formatter to use if Director::is_cli() is true
+     */
+    public function setCliFormatter(FormatterInterface $cliFormatter)
```
tractorcow: Our current convention is to fully capitalise initials. Should this be `setCLIFormatter`?

_config/logging.yml
```diff
@@ -38,6 +38,7 @@ SilverStripe\Core\Injector\Injector:
       - "error"
     properties:
       Formatter: %$FriendlyErrorFormatter
```
tractorcow: To maintain the distinction, should this be HTTPFormatter?

chillu:
@sminnee FYI this is "blocking" deployment of ss4-demo.silverstripe.org (because I can't figure out the error without this patch). Blocking in quotation marks because I could unblock it in various ways, but in case you plan to work on this ... :D

src/Logging/HTTPOutputHandler.php
sminnee: Just `* @param $cliFormatter` eh?

 * The purpose of the method is defined in the method description
 * The type is in the type hint

src/Logging/HTTPOutputHandler.php
sminnee: I'd follow PSR-2 I'll check what it says

sminnee: Not specifically mentioned in PSR-1 but these two references suggest that `setCliFormatter()` would be better:

 * https://groups.google.com/forum/#!topic/php-fig/rGopcf4WyyU
 * http://stackoverflow.com/a/15526526

_config/logging.yml
sminnee: I could just use `SilverStripe\Logging\DebugViewFriendlyErrorFormatter` as the injection point

src/Logging/HTTPOutputHandler.php
robbieaverill: `@param [type] $varName` (PHPStorm does it wrong)

src/Logging/HTTPOutputHandler.php
tractorcow: psr-2 doesn't strictly declare, apparently. https://groups.google.com/forum/#!topic/php-fig/rGopcf4WyyU

I suggest we stick with our own convention until such time as it's clarified.

_config/logging.yml
tractorcow: Or given that the interface is `Monolog\Formatter\FormatterInterface`, we could use dot notation to declare the named variant of the interface provided.

`Monolog\Formatter\FormatterInterface.friendly`

and

`Monolog\Formatter\FormatterInterface.cli`

Depending on how pure you want your treatment of interfaces to be. :)

sminnee:
@chillu force pushed - wanna review?

tractorcow:
It's not critical (but best practice) to have `@return <type>` phpdocs on the getters too.

src/Logging/HTTPOutputHandler.php
sminnee: Type violates DRY

sminnee:
Added `@return`s

src/Logging/HTTPOutputHandler.php
tractorcow: I think @sminnee may be right... maybe I've been doing it wrong all these years? :P

sminnee:
@chillu ready to merge... ;-)

PR #6839, opened 2017-04-24
FIX Create Image_Cached with Injector.

### LINKED ISSUE #6838: [3.x] Image and Image_Cached classes cannot be substituted with Injector
There are many modules that extend core Image class, for example in [kinglozer/silverstripe-tinypng](https://github.com/kinglozzer/silverstripe-tinypng).
Using this configuration:

```yml
Injector:
  Image:
    class: TinyPngImage
  Image_Cached:
    class: TinyPngImage_Cached
```

However `Image::getFormattedImage()` creates `Image_Cached` with **new** operator.

It should be just one line fix for Injector:

```
 - $cached = new Image_Cached($cacheFile);
+ $cached = Injector::inst()->createWithArgs('Image_Cached', array($cacheFile));
```

Another improvement is dynamic tag generation, with API to set/remove any attribute, and IMG tag itself (alternatives - amp-img and picture). This would allow modules to extend Image class in all possible ways.

Would it be possible to add this change in 3.6.0 release?

gregsmirnov:
Fixes #6838

### DIFF
--- a/model/Image.php
+++ b/model/Image.php
@@ -730,7 +730,7 @@ class Image extends File implements Flushable {
 				call_user_func_array(array($this, "generateFormattedImage"), $args);
 			}
 
-			$cached = new Image_Cached($cacheFile, false, $this);
+			$cached = Injector::inst()->createWithArgs('Image_Cached', array($cacheFile, false, $this));
 			return $cached;
 		}
 	}


gregsmirnov:
added another pull request rebased on top of 3.4 branch

robbieaverill:
FYI you can switch the target branch for a PR in GitHub

Issue #6210, opened 2016-10-21
RFC Extensions must whitelist provided methods


tractorcow:
Extensions currently allow method extensions in the following ways: Either by methods that can be invoked via `$obj->extend('method')`, or by providing a default method implementation that can be invoked if the parent object does not have a definition for `$obj->method()`.

However there is no distinction between the two forms of methods, meaning that it's impossible to provide an extension point without also providing a default method implementation. This can be hazardous for extensions which are ONLY intended to work with `->extend` and should not increase the parent object's surface area.

I propose that we add `provide_methods` config to extensions. By default methods on extensions can only be invoked via `->extend`. However, a default method can also be provided, given that it's registered.

Additionally, the `provides_methods` can support aliases, so you can have two different method implementations.

E.g. Versioned could look like the below:

``` php


class Versioned extends Extension {
    private static $provides_methods = [
        'canArchive' => 'defaultCanArchive'
    ];


    /**
     * Can be invoked on parent object via $obj->canArchive()
     */
    public function defaultCanArchive($member = null) {
        // Default method implementation
        return $this->extendedCan('canArchive');
    }

    /**
     * can only be invoked via $obj->extend('canArchive')
     */
    public function canArchive($method = null) {
        return null;
    }
}

```

provides_methods could also support non-associative values, which are a 1-to-1 mapping.

A downside to this I can see is that it breaks php type and method hinting, which previously was supported via `@mixin Versioned`, so some more thought is necessary here. :)

# PRs created:

 - [ ] https://github.com/silverstripe/silverstripe-framework/pull/7752
 - [ ] https://github.com/silverstripe/silverstripe-versioned/pull/84
 - [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/723
 - [x] https://github.com/silverstripe/silverstripe-testsession/pull/55

kinglozzer:
Alternative idea: make the “provided methods” public, but allow `->extend()` calls to call protected methods.

```php
class MyExtension extends Extension
{
    private static $db = [
        'Test' => 'Varchar'
    ];

    protected function updateCMSFields($fields)
    {
        $fields->push(TextField::create('Test'));
    }

    public function FormattedTest()
    {
        return str_replace('Foo', 'Bar', $this->owner->Test);
    }
}

$inst = MyDataObject::get()->first(); // Has the above extension applied
$inst->FormattedTest(); // Returns the value
$inst->updateCMSFields(); // lolnope
[...]

PR #6847, opened 2017-04-26
NEW Use splat operator for extensible calls

dhensby:
As we're now php 5.6+ in master we can use the ellipsis operator on the extend calls.

### DIFF
--- a/src/Core/Extensible.php
+++ b/src/Core/Extensible.php
@@ -405,25 +405,19 @@ trait Extensible
      * all results into an array
      *
      * @param string $method the method name to call
-     * @param mixed $a1
-     * @param mixed $a2
-     * @param mixed $a3
-     * @param mixed $a4
-     * @param mixed $a5
-     * @param mixed $a6
-     * @param mixed $a7
+     * @param mixed ...$a
      * @return array List of results with nulls filtered out
      */
-    public function invokeWithExtensions($method, &$a1 = null, &$a2 = null, &$a3 = null, &$a4 = null, &$a5 = null, &$a6 = null, &$a7 = null)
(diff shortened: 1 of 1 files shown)


flamerohr:
There are cases where some hacks check for 8 args to understand if the function was called by an `extend()`
do you think those in core have been addressed?

tractorcow:
Yep, we have this in Versioned.php

```php
public function canPublish($member = null)
    {
        // Skip if invoked by extendedCan()
        if (func_num_args() > 4) {
            return null;
        }
```

I raised this issue over at https://github.com/silverstripe/silverstripe-framework/issues/6210.

dhensby:
> There are cases where some hacks check for 8 args to understand if the function was called by an extend() do you think those in core have been addressed?

🤦‍♂️ 

I'll look into this....

tractorcow:
@sminnee has confirmed he's happy to split canView() into canView and extendsCanView(). (see https://github.com/silverstripe/silverstripe-framework/issues/6210). That will unblock this story, although it'll need to be done at the same time, or earlier.

tractorcow:
Blocked by https://github.com/silverstripe/silverstripe-framework/issues/6210

tractorcow:
Shifting to 5.x milestone as there are no more API breakages to go into 4.x at the moment.

dhensby:
😱 is this going to be an API breakage? is it because we need to change how extensions can avoid self-calling themselves?

tractorcow:
Yes that's right.

tractorcow:
@dhensby replaced with:

 - https://github.com/silverstripe/silverstripe-framework/pull/7752
 - https://github.com/silverstripe/silverstripe-versioned/pull/84

PR #6850, opened 2017-04-27
FIX IntlLocales::validate when lang and region are the same e.g. de_DE

robbieaverill:
Provided test will fail locales that have the same language and region e.g. de_DE and fr_FR when tested against the current dev-master. This change also handles locale codes that are all lowercase, eg. zh_cmn

### DIFF
--- a/src/i18n/Data/Intl/IntlLocales.php
+++ b/src/i18n/Data/Intl/IntlLocales.php
@@ -1605,6 +1605,18 @@ class IntlLocales implements Locales, Resettable
         if (!$lang || !$region) {
             return false;
         }
+
+        // Check the configurable whitelist
+        $localeCode = strtolower($lang) . '_' . strtoupper($region);
+        $locales = $this->getLocales();
+
+        if (array_key_exists($localeCode, $locales)
+            || array_key_exists(strtolower($localeCode), $locales)
+        ) {
+            return true;
+        }
+
+        // Fallback
         return strcasecmp($lang, $region)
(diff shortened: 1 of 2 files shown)


src/i18n/Data/Intl/IntlLocales.php
```diff
@@ -1605,9 +1605,11 @@ public function validate($locale)
         if (!$lang || !$region) {
             return false;
         }
-        return strcasecmp($lang, $region)
-            && strcasecmp($lang, $locale)
-            && strcasecmp($region, $locale);
+        $localeCode = strtolower($lang) . '_' . strtoupper($region);
+        $locales = $this->getLocales();
+
+        return array_key_exists($localeCode, $locales)
```
tractorcow: This code means that if a locale is valid on the current system, but we haven't added it to our config, we consider it invalid. We should probably not fail an otherwise valid locale just because of this.

src/i18n/Data/Intl/IntlLocales.php
```diff
@@ -1605,9 +1605,11 @@ public function validate($locale)
         if (!$lang || !$region) {
             return false;
         }
-        return strcasecmp($lang, $region)
```
tractorcow: Removing the $lang and $region comparison should be all you need to do to fix this issue.

robbieaverill: If you're happy with that! I don't know the context so thought this would suffice. Will re-push shortly.

PR #6854, opened 2017-04-28
Prefer PHP 5.6 in sake

mobiusnz:
php5.6 is more or less an afterthought in current operating system versions, and as Silverstripe 3 requires it, we need to adjust the tools to prefer 5.6 over the more common 7.x

### DIFF
--- a/sake
+++ b/sake
@@ -45,7 +45,7 @@ else
 fi
 
 # Find the PHP binary
-for candidatephp in php php5; do
+for candidatephp in php5.6 php php5; do
 	if [ `which $candidatephp 2>/dev/null` -a -f `which $candidatephp 2>/dev/null` ]; then
 		php=`which $candidatephp 2>/dev/null`
 		break


tractorcow:
Don't most bin aliases for php exclude periods? e.g. `php56` vs `php5.6`

wernerkrauss:
Just a note: On my managed server i can choose the php version per user / domain.

When running dev/build with another php version than my webserver runs, though it's the same user, I get a bad manifest and a whitescreen of death.

This happened on SS3.x with the php7 patch installed.

I usually symlink "php" to the php version used by the webserver.

IMHO it's bad to prefer `php56` over `php` because you don't know what php version is used by the current installation - now with SS3.6 around the corner it might also be php7 out of the box.

sminnee:
Given that SilverStripe 3.6 will work with PHP 7 I think the primary purpose of this is obsolete, and I wouldn't generally make this kind of change in a patch release.

Perhaps your best bet will be to upgrade to 3.6.0-beta1! You can expect 3.6 stable by end of May. :-)

tractorcow:
Posting my random thought here; I think sake should be moved to /vendor/bin, and probably given a modernisation.

sminnee:
There are community projects afoot that will fill this need :-)

PR #6856, opened 2017-04-28
FIX #6855: Mangled JS in Requirements, escaping replacement values prior to passing to preg_replace().

### LINKED ISSUE #6855: SS 3.x: Requirements::customScript() + force_js_to_bottom is butchering content.
I noticed that when I had JSON-encoded strings like the following (cutting a snippet):

```js
var data = {"stats":[{"value":"$8,871,721"}]};
```

It would get mangled to:

```js
var data = {"stats":[{"value":",871,721"}]};
```

And this happened eventually until I disabled `force_js_to_bottom`. You can easily reproduce this by simply using the following PHP code. This is happening due to `$n` looking like regular expression back references, so obviously this method is going to cause issues:

```php
$jsRequirements = '<script>var data = {"stats":[{"value":"$8,871,721"}]};</script>';
$jsRequirements = preg_replace('/>\n*/', '>', $jsRequirements);
$content = '</body>';
$content = preg_replace("/(<\/body[^>]*>)/i", $jsRequirements . "\\1", $content);
```

Output is:
```html
<script>var data = {"stats":[{"value":",871,721"}]};</script></body>
```

Instead of:

```html
<script>var data = {"stats":[{"value":"$8,871,721"}]};</script></body>
```

Maybe just escape the `$jsRequirements` propr to passing it into `preg_replace`? I checked the `3` branch and [this is still happening](https://github.com/silverstripe/silverstripe-framework/blob/3/view/Requirements.php#L872) (I'm on an older version of SS but I tested this file). It is, however, being taken care of properly in `master` since I see it's being passed through to `->escapeReplacement(...)` ([link](https://github.com/silverstripe/silverstripe-framework/blob/master/src/View/Requirements_Backend.php#L817))

patricknelson:
PR to address issue #6855 where SS 3.x is currently mangling JavaScript code when forcing to bottom. This is due to using `preg_replace()` and interpreting `$n` (such as `$8`) as back references.

Copied `->escapeReplacement()` from `master` and  attempted to place it in a spot least likely to require effort when merging up :crossed_fingers:

### DIFF
--- a/view/Requirements.php
+++ b/view/Requirements.php
@@ -869,10 +869,10 @@ class Requirements_Backend {
 
 				// Forcefully put the scripts at the bottom of the body instead of before the first
 				// script tag.
-				$replacements["/(<\/body[^>]*>)/i"] = $jsRequirements . "\\1";
+				$replacements["/(<\/body[^>]*>)/i"] = $this->escapeReplacement($jsRequirements) . "\\1";
 
 				// Put CSS at the bottom of the head
-				$replacements["/(<\/head>)/i"] = $requirements . "\\1";
+				$replacements["/(<\/head>)/i"] = $this->escapeReplacement($requirements) . "\\1";
 			} elseif ($this->write_js_to_body) {
 				$jsRequirements = $this->removeNewlinesFromCode($jsRequirements);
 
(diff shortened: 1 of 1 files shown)


sminnee:
@patricknelson can you address the failing tests?

patricknelson:
Ugh, talk about mangling :man_facepalming: sorry about that, here's an updated version. I was also in such a hurry that I forgot to put this in a separate branch. Anyway, I appreciate the quick response!

PR #6865, opened 2017-05-02
BUG Fix ApcuCache namespace

tractorcow:
@flamerohr could you please review and test that this fixes your issue?

### DIFF
--- a/src/Core/Cache/ApcuCacheFactory.php
+++ b/src/Core/Cache/ApcuCacheFactory.php
@@ -27,9 +27,13 @@ class ApcuCacheFactory implements CacheFactory
      */
     public function create($service, array $params = array())
     {
-        return Injector::inst()->create(ApcuCache::class, false, [
-            (isset($args['namespace'])) ? $args['namespace'] : '',
-            (isset($args['defaultLifetime'])) ? $args['defaultLifetime'] : 0,
+        $namespace = isset($params['namespace'])
+            ? $params['namespace'] . '_' . md5(BASE_PATH)
+            : md5(BASE_PATH);
+        $defaultLifetime = isset($params['defaultLifetime']) ? $params['defaultLifetime'] : 0;
+        return Injector::inst()->createWithArgs(ApcuCache::class, [
+            $namespace,
+            $defaultLifetime,
             $this->version
         ]);
     }
--- a/src/Core/Cache/DefaultCacheFactory.php
+++ b/src/Core/Cache/DefaultCacheFactory.php
@@ -71,7 +71,8 @@ class DefaultCacheFactory implements CacheFactory
         }
 
         // Chain this cache with ApcuCache
-        $apcu = $this->createCache(ApcuCache::class, [$namespace, (int) $defaultLifetime / 5, $version]);
+        $apcuNamespace = $namespace . ($namespace ? '_' : '') . md5(BASE_PATH);
+        $apcu = $this->createCache(ApcuCache::class, [$apcuNamespace, (int) $defaultLifetime / 5, $version]);
         return $this->createCache(ChainCache::class, [[$apcu, $fs]]);
     }
 
--- a/src/Core/Cache/MemcachedCacheFactory.php
+++ b/src/Core/Cache/MemcachedCacheFactory.php
@@ -17,7 +17,7 @@ class MemcachedCacheFactory implements CacheFactory
     /**
      * @param Memcached $memcachedClient
      */
-    public function __construct(Memcached $memcachedClient)
+    public function __construct(Memcached $memcachedClient = null)
     {
         $this->memcachedClient = $memcachedClient;
     }
@@ -27,10 +27,14 @@ class MemcachedCacheFactory implements CacheFactory
      */
     public function create($service, array $params = array())
     {
-        return Injector::inst()->create(MemcachedCache::class, false, [
+        $namespace = isset($params['namespace'])
+            ? $params['namespace'] . '_' . md5(BASE_PATH)
+            : md5(BASE_PATH);
+        $defaultLifetime = isset($params['defaultLifetime']) ? $params['defaultLifetime'] : 0;
+        return Injector::inst()->createWithArgs(MemcachedCache::class, [
             $this->memcachedClient,
-            (isset($args['namespace'])) ? $args['namespace'] : '',
-            (isset($args['defaultLifetime'])) ? $args['defaultLifetime'] : 0
+            $namespace,
+            $defaultLifetime
         ]);
     }
 }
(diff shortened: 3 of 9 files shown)


flamerohr:
This fix works! Maybe check out `ApcuCacheFactory` as well, I don't know if it's even used.

Possibly other state-shared cache systems like Memcache and Redis? Although those ones will probably need more thought put into it, as it'll be across systems... Possibly a `.env` variable to set a proper shared namespace?

tractorcow:
Those caches will need the cache constructor to pass in a necessarily distinct key. Since there's no direct references to those in core (they exist but aren't used) there isn't any code for me to update. It'll have to be something users use their discretion to implement.

tractorcow:
I've updated non-filesystem caches so that they use the directory to distinguish their namespaces.

src/Core/Cache/ApcuCacheFactory.php
```diff
@@ -27,9 +27,13 @@ public function __construct($version = null)
      */
     public function create($service, array $params = array())
     {
+        $namespace = isset($params['namespace']) ? $params['namespace'] : '';
+        $defaultLifetime = isset($params['defaultLifetime']) ? $params['defaultLifetime'] : 0;
+        $directory = isset($params['directory']) ? $params['directory'] : null;
+        $apcuNamespace = $namespace . '_' . substr(md5($directory), 0, 7);
         return Injector::inst()->create(ApcuCache::class, false, [
-            (isset($args['namespace'])) ? $args['namespace'] : '',
```
tractorcow: Existing code used `$args` but passed in `$params`. oops :P

src/Core/Cache/ApcuCacheFactory.php
```diff
@@ -27,9 +27,13 @@ public function __construct($version = null)
      */
     public function create($service, array $params = array())
     {
+        $namespace = isset($params['namespace']) ? $params['namespace'] : '';
+        $defaultLifetime = isset($params['defaultLifetime']) ? $params['defaultLifetime'] : 0;
+        $directory = isset($params['directory']) ? $params['directory'] : null;
+        $apcuNamespace = $namespace . '_' . substr(md5($directory), 0, 7);
```
flamerohr: leaves a dangling `_` if directory isn't defined

src/Core/Cache/ApcuCacheFactory.php
```diff
@@ -27,9 +27,13 @@ public function __construct($version = null)
      */
     public function create($service, array $params = array())
     {
+        $namespace = isset($params['namespace']) ? $params['namespace'] : '';
+        $defaultLifetime = isset($params['defaultLifetime']) ? $params['defaultLifetime'] : 0;
+        $directory = isset($params['directory']) ? $params['directory'] : null;
```
sminnee: I don't think we can rely on directly being passed, since this isn't a filesystem-backed cache and we'd be relying on users to set this manually, which is bad dev-experience.

I'd recommend adding BASE_PATH to the cache key instead.

src/Core/Cache/ApcuCacheFactory.php
sminnee: Are namespace strings length-limited? Is this substr necessary? It just risks a collision which I admit is rare but I'd only do it if it's critical.

Alternatively you could strip out bad chars from the directory.

src/Core/Cache/ApcuCacheFactory.php
tractorcow: I see, I think you're right.

src/Core/Cache/DefaultCacheFactory.php
```diff
@@ -71,7 +71,8 @@ public function create($service, array $args = array())
         }
 
         // Chain this cache with ApcuCache
-        $apcu = $this->createCache(ApcuCache::class, [$namespace, (int) $defaultLifetime / 5, $version]);
+        $apcuNamespace = $namespace . '_' . substr(md5($directory), 0, 7);
```
sminnee: Ditto on the need for substr

src/Core/Cache/MemcachedCacheFactory.php
```diff
@@ -27,10 +27,14 @@ public function __construct(Memcached $memcachedClient)
      */
     public function create($service, array $params = array())
     {
+        $namespace = isset($params['namespace']) ? $params['namespace'] : '';
+        $defaultLifetime = isset($params['defaultLifetime']) ? $params['defaultLifetime'] : 0;
+        $directory = isset($params['directory']) ? $params['directory'] : null;
+        $memCachedNamespace = $namespace . '_' . substr(md5($directory), 0, 7);
         return Injector::inst()->create(MemcachedCache::class, false, [
```
flamerohr: I think this should be `Injector::inst()->createWithArgs();` in both factories

src/Core/Cache/ApcuCacheFactory.php
tractorcow: 7 characters on an md5 is sufficiently rare that no two instances of silverstripe on a site will collide. there won't be any bad characters, but I can increase it if needed.

tractorcow:
Updated with tests to cover actual factory behaviour.

tractorcow:
I've had to fix FormSchemaTest linting issue by shifting the json to actual .json files instead of inlining in PHP as arrays.

tests/php/Forms/FormSchemaTest.php
```diff
@@ -22,43 +21,7 @@ public function testGetSchema()
...
-                    'disabled' => false,
-                    'customValidationMessage' => '',
-                    'attributes' => [],
-                    'data' => [],
-                    'validation' => [],
-                ],
-            ],
-            'actions' => []
-        ];
+        $expected = json_decode(file_get_contents(__DIR__.'/FormSchemaTest/testGetSchema.json'), true);
```
flamerohr: maybe remove the `json_decode()` and the `json_encode($expected)` below?
encoding for the sake of decoding :D

tests/php/Forms/FormSchemaTest.php
```diff
@@ -183,188 +146,7 @@ public function testGetNestedSchema()
...
-                                'icon' => null,
-                            ],
-                            'validation' => [],
-                        ],
-                    ]
-                ]
-            ]
-        ];
-
+        $expected = json_decode(file_get_contents(__DIR__.'/FormSchemaTest/testGetNestedSchema.json'), true);
```
flamerohr: same as above

tests/php/Forms/FormSchemaTest.php
```diff
@@ -406,154 +188,7 @@ public function testSchemaValidation()
...
-                            [],
-                        'data' =>
-                            [],
-                    ],
-                ],
-            'actions' =>
-                [],
-        ];
-
+        $expected = json_decode(file_get_contents(__DIR__.'/FormSchemaTest/testSchemaValidation.json'), true);
```
flamerohr: same as above

PR #6867, opened 2017-05-03
Fix class name reference consistency

gregsmirnov:
Simple patch that helps Object::__call() to be independent of `$this->class`.
The latest is used frequently for ORM. For example BetterImage can replace Image::getTag() but still act as Image class with explicitly set `$this->class = 'Image';`


tractorcow:
The fact that you can set `$obj->class => 'SomeNotObjClass'` is the reason I want to do away with that property entirely. :)

PR #6872, opened 2017-05-04
NEW: Allow SapphireTest::objFromFixture() to accept either table or class

sminnee:
Right now SapphireTest::objFromFixture() requires a class as the first
argument. This is fine when your fixture file uses classes as the keys,
but if populating a fixture via tables, objFromFixture() won’t work.

This patch lets you specify either the table name or the class name as
the key.

The benefit here is that you can build fixtures as raw inserts, which is
substantially quicker, and is likely to be a useful tool in building
more efficient test suites.

### DIFF
--- a/src/Dev/FixtureFactory.php
+++ b/src/Dev/FixtureFactory.php
@@ -135,7 +135,7 @@ class FixtureFactory
     /**
      * Return all of the IDs in the fixture of a particular class name.
      *
-     * @param string $class
+     * @param string $class The data class or table name
      * @return array|false A map of fixture-identifier => object-id
      */
     public function getIds($class)
@@ -162,7 +162,7 @@ class FixtureFactory
     /**
      * Get an object from the fixture.
      *
-     * @param string $class The data class, as specified in your fixture file.  Parent classes won't work
+     * @param string $class The data class or table name, as specified in your fixture file.  Parent classes won't work
      * @param string $identifier The identifier string, as provided in your fixture file
      * @return DataObject
      */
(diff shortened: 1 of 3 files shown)


sminnee:
This came up when I was looking into why blog tests were so slow. It had the following improvements:

 * Fixture set-up time: 4s => 0.08s
 * Text execution time: 170 seconds => 30 seconds.

You can see what I had to change here: https://github.com/silverstripe/silverstripe-blog/pull/438/commits/c4abe91b4b07ec26398ba394d545f3e6ca900674

The yaml is admittedly more verbose but a reasonable execution time on tests is worth the effort. Longer term we might want to find alternatives to yml fixtures as they encourage some bad habits, but that's for another day.

tractorcow:
> The benefit here is that you can build fixtures as raw inserts, which is substantially quicker

Yes but it also bypasses all insert hooks, doesn't populate subtables, versions table, etc and bypasses error checking and validation. My feeling is that this introduces risk of errors silently creeping into test state. How do you feel about this?

sminnee:
> Yes but it also bypasses all insert hooks, doesn't populate subtables, versions table, etc and bypasses error checking and validation. My feeling is that this introduces risk of errors silently creeping into test state. How do you feel about this?

I think that in many cases the benefits outweigh the costs, and that the costs are not so great that we should make the approach I've suggested impossible.

As an example: it won't be feasible to have code coverage for the blog module without this.

sminnee:
In the big picture I think I'd like to:

 * Investigate the creation time of SiteTree to see why they're so much slower than raw inserts
 * Move away from using fixture files as they encourage the creation of "kitchen sink" fixtures that slow down creation-time and instead have a DSL for scaffolding data on a per-test-method basis

However both of these are more work than the improvement allowed by this patch.

PR #6873, opened 2017-05-04
Configurable Build Task

silverstripe-elliot:
This change will allow Build Tasks to be enabled or disabled using the Config API.

I noticed on another project that certain tasks (such as those from modules) were not able to be enabled or disabled via the Config system. This change will allow those tasks to be disabled by a developer by setting the `$enable` variable to `true` or `false` in their project's `.yml` files.

It should work on older versions of SilverStripe as well, but since I had to change the visibility and scope of the $enable variable, it's a potentially breaking change for legacy code.

### DIFF
--- a/src/Dev/BuildTask.php
+++ b/src/Dev/BuildTask.php
@@ -9,8 +9,8 @@ use SilverStripe\Core\Object;
  * Interface for a generic build task. Does not support dependencies. This will simply
  * run a chunk of code when called.
  *
- * To disable the task (in the case of potentially destructive updates or deletes), declare
- * the $Disabled property on the subclass.
+ * To disable the task (in the case of potentially destructive updates or deletes), change the
+ * `enabled` property on the subclass to true.
  */
 abstract class BuildTask extends Object
 {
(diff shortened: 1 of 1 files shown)


silverstripe-elliot:
I'll have a look at the failing tests over the weekend

dhensby:
how was the weekend :)

tractorcow:
@silverstripe-elliot can you please tag with `API` label so it'll go into our changelog notice too? Thanks. :)

Fixing the tests is as simple as changing the test classes `protected $enabled` to `private static $enabled` too.

Also fix `TaskRunner`

```diff
protected function taskEnabled($class)
{
-    $reflectionClass = new ReflectionClass($class);
-    if ($reflectionClass->isAbstract()) {
-        return false;
-    } elseif (!singleton($class)->isEnabled()) {
-        return false;
-    }
-
-    return true;
+   return Config::inst()->get($class, 'enabled');
}
```

maxime-rainville:
Hi @silverstripe-elliot. This looks like its was targeted to SilverStripe 3.

If you feel this is still needed, please open a PR against the 4 branch.

Issue #6063, opened 2016-09-22
Introduce asset permission model


sminnee:
## Acceptance criteria

 * "Viewers" and "Editors" are properties on Folder and Files, with the default to inherit from parent, just like Pages
 * These fields appear on the folder/file detail froms
 * Default canView() and canEdit() implementations make use of these
 * 100,000 files, 1,000 per folder, are supported
 * For debate: should we use new SiteConfig fields for root permission or separate fields for pages and files?
 * Files with canView() = false for the current user aren't displayed in asset admin
 * Files with canEdit() = false for the current user show a readonly from in asset admin

Assumption: existing level of optimisation in `SiteTree::batch_permission_check()` is sufficient.

## Tasks

 - [x] Implement abstract hierarchal permission calculator based on current SiteTree permissions
 - [x] Update SiteTree to use new permission model
 - [x] Update File to use this permission module
 - [x] Update SiteTree unit tests
 - [x] Permission model unit tests
 - [x] Asset admin behat tests for permissions / readonly
 - [x] Verify NFR (100,000 files, 1,000 folders)
 - [x] Ensure readonly mode is visible in the asset-admin
 - [x] Ensure view / editor groups are editable within the asset edit form for files and folders
 - [x] Ensure that multi-select of groups can be done in react (upgrade or create new react component).
 
# Pull Requests

- [x] https://github.com/silverstripe/silverstripe-framework/pull/6877
- [x] https://github.com/silverstripe/silverstripe-assets/pull/13
- [x] https://github.com/silverstripe/silverstripe-cms/pull/1811
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/465
- [x] https://github.com/silverstripe/silverstripe-admin/pull/54
- [x] https://github.com/silverstripe/silverstripe-siteconfig/pull/57

----

Right now asset-admin is built to respect `canView()` and `canEdit()` permissions of files, however there is no default permission model provided.

We should have some permission model built in that allows protected files and folders, similar in concept to the default page permissions.

We may wish to make this functionality provided-by-default but able-to-be-removed, so that developers can replace the default scheme in its entirety with something else.

tractorcow:
👍 

Something like an extremely bare-bones secureassets would be nice.

chillu:
[...]

PR #6877, opened 2017-05-05
API Implement InheritedPermission calculator

tractorcow:
Refactor of SiteTree inheritable permissions to a core service.

See https://github.com/silverstripe/silverstripe-framework/issues/6063

### DIFF
--- a/src/Security/InheritedPermissions.php
+++ b/src/Security/InheritedPermissions.php
@@ -1,0 +1,632 @@
+<?php
+
+namespace SilverStripe\Security;
+
+use InvalidArgumentException;
+use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\ORM\DataList;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\Hierarchy\Hierarchy;
+use SilverStripe\Versioned\Versioned;
+
+/**
+ * Calculates batch permissions for nested objects for:
+ *  - canView: Supports 'Anyone' type
+ *  - canEdit
+ *  - canDelete: Includes special logic for ensuring parent objects can only be deleted if their children can
+ *    be deleted also.
+ */
+class InheritedPermissions
+{
+    use Injectable;
+
+    const DELETE = 'delete';
+    const VIEW = 'view';
+    const EDIT = 'edit';
+
+    /**
+     * Anyone canView permission
+     */
+    const ANYONE = 'Anyone';
+
+    /**
+     * Restrict to logged in users
+     */
+    const LOGGED_IN_USERS = 'LoggedInUsers';
+
+    /**
+     * Restrict to specific groups
+     */
+    const ONLY_THESE_USERS = 'OnlyTheseUsers';
+
+    /**
+     * Inherit from parent
+     */
+    const INHERIT = 'Inherit';
+
+    /**
+     * Class name
+     *
+     * @var string
+     */
+    protected $baseClass = null;
+
+    /**
+     * Object for evaluating top level permissions designed as "Inherit"
+     *
+     * @var RootPermissions
+     */
+    protected $rootPermissions = null;
+
+    /**
+     * Global permissions required to edit.
+     * If empty no global permissions are required
+     *
+     * @var array
+     */
+    protected $globalEditPermissions = [];
+
+    /**
+     * Cache of permissions
+     *
+     * @var array
+     */
+    protected $cachePermissions = [];
+
+    /**
+     * @param RootPermissions $callback
+     * @return $this
+     */
+    public function setRootPermissions(RootPermissions $callback)
+    {
+        $this->rootPermissions = $callback;
+        return $this;
+    }
+
+    /**
+     * Global permissions required to edit
+     *
+     * @param array $permissions
+     * @return $this
+     */
+    public function setGlobalEditPermissions($permissions)
+    {
+        $this->globalEditPermissions = $permissions;
+        return $this;
+    }
+
+    /**
+     * @return array
+     */
+    public function getGlobalEditPermissions()
+    {
+        return $this->globalEditPermissions;
+    }
+
+    /**
+     * Get root permissions handler, or null if no handler
+     *
+     * @return RootPermissions|null
+     */
+    public function getRootPermissions()
+    {
+        return $this->rootPermissions;
+    }
+
+    /**
+     * Get base class
+     *
+     * @return string
+     */
+    public function getBaseClass()
+    {
+        return $this->baseClass;
+    }
+
+    /**
+     * Set base class
+     *
+     * @param string $baseClass
+     * @return $this
+     */
+    public function setBaseClass($baseClass)
+    {
+        if (!is_a($baseClass, DataObject::class, true)) {
+            throw new InvalidArgumentException('Invalid DataObject class: ' . $baseClass);
+        }
+        $this->baseClass = $baseClass;
+        return $this;
+    }
+
+    /**
+     * Pre-populate the cache of canEdit, canView, canDelete, canPublish permissions. This method will use the static
+     * can_(perm)_multiple method for efficiency.
+     *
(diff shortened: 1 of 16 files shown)


tests/php/Security/InheritedPermissionsTest.php
```diff
@@ -0,0 +1,269 @@
...
+        $this->assertFalse($product4->canEdit($editor));
+
+        // Test that root node respects root permissions
+        $this->assertTrue($history->canEdit($editor));
+
+        TestPermissionNode::getInheritedPermissions()->clearCache();
+        $this->rootPermissions->setCanEdit(false);
+
+        // With root edit false, permissions are now denied for CanEditType = Inherit
+        $this->assertFalse($history->canEdit($editor));
```
unclecheese: No `canCreate()` tests?

unclecheese:
@tractorcow Is there a reason why `canCreate()` is never tested in this PR?

tractorcow:
Inherited permissions doesn't cover canCreate. It's left up to individual dataobjects.

src/Security/InheritedPermissions.php
```diff
@@ -0,0 +1,632 @@
...
+            $cachedValues = array_intersect_key($this->cachePermissions[$cacheKey], $result);
+
+            // If we can't find everything in the cache, then look up the remainder separately
+            $uncachedIDs = array_keys(array_diff_key($result, $this->cachePermissions[$cacheKey]));
+            if ($uncachedIDs) {
+                $uncachedValues = $this->canDeleteMultiple($uncachedIDs, $member, false);
+                return $cachedValues + $uncachedValues;
+            }
+            return $cachedValues;
+        }
```
unclecheese: @tractorcow What do you think of this?

src/Security/InheritedPermissions.php
```diff
@@ -0,0 +1,632 @@
...
+     *
+     * Caches based on $typeField data. To invalidate the cache, use {@link SiteTree::reset()} or set the $useCached
+     * property to FALSE.
+     *
+     * @param string $type Either edit, view, or create
+     * @param array $ids Array of IDs
+     * @param Member $member Member
+     * @param array $globalPermission If the member doesn't have this permission code, don't bother iterating deeper
+     * @param bool $useCached Enables use of cache. Cache will be populated even if this is false.
+     * @return array An map of <a href='psi_element://SiteTree'>SiteTree</a> ID keys to boolean values
```
chillu: What's this `psi_element` stuff about?

src/Security/InheritedPermissions.php
```diff
@@ -0,0 +1,632 @@
...
+        return $this->baseClass;
+    }
+
+    /**
+     * Set base class
+     *
+     * @param string $baseClass
+     * @return $this
+     */
+    public function setBaseClass($baseClass)
```
chillu: Shouldn't this be a constructor argument? The implementation will fail without this being set, right?

src/Security/InheritedPermissions.php
```diff
@@ -0,0 +1,632 @@
...
+     * @return string
+     */
+    protected function getJoinTable($type)
+    {
+        $table = DataObject::getSchema()->tableName($this->baseClass);
+        switch ($type) {
+            case self::DELETE:
+                // Delete uses edit type - Drop through
+            case self::EDIT:
+                return "{$table}_EditorGroups";
```
chillu: Should we make the join tables configurable to make this class a bit more reuseable?

src/Security/RootPermissions.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace SilverStripe\Security;
+
+/**
+ * For root objects that have permission type "Inherit" determine the result of certain permission checks
```
chillu: I had a hard time understanding the purpose of this class. Suggested alternative:

> Allows objects to enforce permissions for the "root" level, where permissions can not be tied to a particular database record. Objects below the "root" level should use their own can*() implementations instead of this interface.

chillu: Also, I'm wondering if we should just name this interface `PermissionChecked` and apply it to `DataObject` as well? There's nothing in the interface which makes this "root only", right? Which also means that you assign the object class itself here, which means you don't want to distinguish between "root level" and "instance level" permission checks in your implementation.

chillu:
> Inherited permissions doesn't cover canCreate. It's left up to individual dataobjects. It's not even a method on the new class. :)

Yeah I guess `canCreate()` doesn't make sense on the root, since it's a "virtual" object. When you create an object, the intent is to save it as a database record, in which case it's by definition not hte "root" object.

src/Security/InheritedPermissions.php
tractorcow: The code blocks are similar but not quite the same; There are some subtle differences that would make refactoring less useful for de-duplication. In this case, it was simpler and more readable to simply repeat the pattern.

src/Security/InheritedPermissions.php
tractorcow: Sorry, that must have been copied from SiteTree. That makes no sense to me!

src/Security/RootPermissions.php
tractorcow: I think RootPermissions is incorrect, as the permissions actually accounts for the -1 level. "root" permissions applies to all items at the root, which are actual instances of DataObjects.

Consider this handler the "imaginary number" equivalent of permissions, which attempts to generalise permissions for objects outside the natural bounds within which objects can be created (root, child, nested). This is why it doesn't apply directly to permissions for actual DataObject instances.

tractorcow: A more correct name is "ParentOfRootPermissions" but that's a bit of a mounthful.

src/Security/InheritedPermissions.php
```diff
@@ -0,0 +1,632 @@
...
+use SilverStripe\Versioned\Versioned;
+
+/**
+ * Calculates batch permissions for nested objects for:
+ *  - canView: Supports 'Anyone' type
+ *  - canEdit
+ *  - canDelete: Includes special logic for ensuring parent objects can only be deleted if their children can
+ *    be deleted also.
+ */
+class InheritedPermissions
```
sminnee: I think that this would make more sense if it implemented a more generic interface such as `PermissionChecker`, `AccessChecker`, `PermissionStore`, etc.

The interface that I would recommend is probably:

 * checkOne($permission, $id, $member);
 * checkMany($permission, $ids, $member);

That means that an inherited permission model is simply 1 approach that we're plugging into this; you could provide other implementations of the same interface.

It will be a good starting point for ACL work in the future, without requiring too much more work in the short term.

tractorcow: Well, we have 
checkOne($permission, $id, $member)
checkMany($permission, $ids, $member)

and we have
check($permission, $member) which is self-scoped to a record. This covers both Datobjects themselves, and the RootPermissions implementors.

I'm having trouble thinking of names to distinguish these interfaces that makes their purpose clear.

tractorcow: How about PermissionChecker interface (implemented by InheritedPermissionChecker) and ScopedPermissionChecker (inherited by the root permission classes, each of which will be renamed)

the difference being ScopedPermissionChecker doesn't take any IDs

src/Security/InheritedPermissions.php
tractorcow: This should cover 99% of use cases. As @unclecheese suggested I'm moving these fields to a standard extension, so if users want to use their own fields, they can subclass InheritedPermissions.

tractorcow:
Feedback implemented. I've created a new interface PermissionChecker and updated all references to InheritedPermissions to use the abstract interface as a service reference.

The root permissions interface has also been renamed.

tractorcow:
Since these tests are green I suggest to merge this first, and then restart silverstripe/assets module tests, @unclecheese .

PR #6879, opened 2017-05-08
MINOR: Fix mismatched brackets

mikenz:
`Too many children (1897}` just didn't look right....


mikenz:
@robbieaverill Agreed, but that means I'd have to checkout a clean framework and do the change properly - not just use the online editor on github. And it didn't seem worth that effort to me.

robbieaverill:
@mikenz ha! It can be squashed on merge anyway.

mikenz:
squashed...

PR #6884, opened 2017-05-08
Remove APIs with target deprecation of 4.0

chillu:
Merge together with https://github.com/silverstripe/silverstripe-cms/pull/1813


tractorcow:
Approved, can be merged on green @chillu if you like.

PR #6887, opened 2017-05-08
Doc dateformats with calendar year

chillu:
https://github.com/silverstripe/silverstripe-framework/issues/3749
http://stackoverflow.com/questions/1978051/zend-datetostring-outputs-the-wrong-year-bug-in-my-code-or-zend-date
https://en.wikipedia.org/wiki/ISO_week_date#Disadvantages


chillu:
Merging, this sure as hell didn't break Behat.

PR #6892, opened 2017-05-08
ENHANCEMENT groupby() is now accessible from on DataList

Hambrook:


tractorcow:
Can you please include some tests that demonstrate placed this feature would add value? The ability to group datalist means that there may be a larger range of edge (and potential error) cases that the ORM now needs to adapt for; This makes me cautious about introducing this feature.

Hambrook:
I've added tests (but somehow pushed before I'd finished them). Finishing them now. 

This is the code that led me to believe that GroupBy should be accessible from DataList

```
public function CarouselEvents($limit=3) {
	$dataQuery = DataObject::get('EventInstance')
		->dataQuery()
			->groupBy('EventID');
	return DataObject::get('EventInstance')
		->setDataQuery($dataQuery)
		->filter('Featured', 1)
		->sort('StartDate ASC')
		->limit($limit);
}
```

tractorcow:
If you are grouping on event ID, don't you need an aggregation method for all non-grouped fields?

This will cause broken queries on mysql 5.7 due to sqlmode_only_full_group_by

https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_only_full_group_by

kinglozzer:
I don’t think this is something we should provide in the `DataList` API. `DataList` will select *all* the fields for a given model, which as @tractorcow points out will result in errors in MySQL 5.7+ (and most other database drivers too) as they’re not aggregated. Even in environments where this doesn’t cause errors, it’s going to cause weird behaviour/edge cases:

| EventID | Title | Featured | StartDate |
| ------- | ----- | --------- | ----------- |
| 1 | Event one | 1 | 2017-05-01 |
| 1 | Event two | 1 | 2017-06-01 |
| 2 | Event three | 0 | 2017-09-01 |
| 3 | Event four | 1 | 2017-11-01 |

Given the example code above (grouping by EventID only), would “Event one” or “Event two” be included in the result set?

Hambrook:
Only "Event one" as it's the first (by StartDate) for that event.

It's all fine if you feel this isn't a fit for the project. I just noticed that I had to jump through some hoops to set a GroupBy and believed it to be an oversight so thought I'd help.

dhensby:
You probably want to build up a custom `DataQuery` if you need this level of control

kinglozzer:
> Only "Event one" as it's the first (by StartDate) for that event.

The problem is that it _may_ work like that for a given example, but it’s not guaranteed to behave in that way (like if the StartDate values were the same) - which is why newer MySQL versions/other drivers will error for that query by default.

As @dhensby suggested, building a custom query is probably best for maximum compatibility. If that’s not of concern then you should still be able to achieve the same thing by calling `alterDataQuery()` on your list directly rather than getting/setting the DataQuery manually:

```php
public function CarouselEvents($limit=3) {
    return EventInstance::get()
        ->alterDataQuery(function($query) {
            $query->groupby('EventID');
        })
        ->filter('Featured', 1)
        ->sort('StartDate ASC')
        ->limit($limit);
}
```

Thanks for submitting though @Hambrook 😃

tractorcow:
You could also join on a subquery, and group on the subquery, filtering by that.

Hambrook:
Thanks for your feedback everyone

PR #6895, opened 2017-05-09
Update index.md

nickbyfleet:
Fixing broken link.


dhensby:
ping @robbieaverill

dhensby:
@nickbyfleet this isn't the right way to fix this. it looks like our API parser might be broken if this link isn't working

robbieaverill:
Yeah, these should be automatically determined by the shortcode syntax rather than direct links

SpiritLevel:
The parsing of [api:Director] style short codes seems to be generating the correct query:

api.silverstripe.org/search/lookup/?q=Director&version=3&module=framework

The problem is that the api site no longer processes such queries. These queries used to work :(

sminnee:
Have created https://github.com/silverstripe/api.silverstripe.org/issues/33 to cover that

PR #6896, opened 2017-05-09
GridField and GridState_component bug

t3hn0:
Ensuirng GridState_Component is added to GridField config even if we set config with GridField::setConfig


t3hn0:
Done :)

forms/gridfield/GridField.php
```diff
@@ -177,6 +174,11 @@ public function getConfig() {
 	public function setConfig(GridFieldConfig $config) {
 		$this->config = $config;
 
+		if (!$this->config->getComponentByType('GridState_Component')) {
+			$this->config->addComponent(new GridState_Component());
+			$this->state = new GridState($this);
```
tractorcow: Should the state itself stay in constructor?

dhensby: I think we need to keep it in the constructor and then ensure it's retained when a new config is assigned...

t3hn0: Looks like online edit works :) 
@tractorcow and @dhensby I moved the line back to constructor.

tractorcow:
@dhensby I don't think we can merge this into 3.6.0; Which branch should I switch to?

dhensby:
I'd class this as a bug, so I think it could go into 3.4? 3.6.0 would work, but then we'd need to hand port it to 3.4+

dhensby:
@t3hn0 thanks for the latest commit, but I think we're still ending up with something inconsistent here.

- GridField seems to store a property `state` that stores the current state object.
- A GridFieldConfig holds these GridFieldState components.
- If we set a new config and it doesn't have a gridfieldstate we should push our existing state object into it.
- If the new config *does* have a GridFieldState component then we need to decide if we store that as the new state property or if that should be discarded in favour of our current state object.

Considering it appears very rare that any state component is added as part of a config (because it's not done at the moment) we can probably allow the passed state object to override ours as that gives power users most control.

sminnee:
@dhensby can you please use the "changes requested" review feature when pushing back a PR?

tractorcow:
@dhensby 

> If we set a new config and it doesn't have a gridfieldstate we should push our existing state object into it.

The state isn't actually stored in the GridState_Component; It's a mere adaptor for serialising the gridfield's state into the form. In other words, you CAN'T push the state into the GridState_Component because it only sits on the gridfield itself.

Essentially, state and gridstate_component can be independently assigned to the gridfield and gridfield config safely. GridState_component.php is literally just this.

```php
<?php

namespace SilverStripe\Forms\GridField;

/**
 * @see GridState
 */
class GridState_Component implements GridField_HTMLProvider
{

    public function getHTMLFragments($gridField)
    {
        return array(
            'before' => $gridField->getState(false)->Field()
        );
    }
}

```

With that in mind, can we merge this? I think the fix is valid.

tractorcow:
Ok, I tried to change the base to 3.4 and it messed up the PR. =(

tractorcow:
Rebased on 3.4 and replaced with https://github.com/silverstripe/silverstripe-framework/pull/6942

PR #6898, opened 2017-05-09
Cleanup SapphireTest and time related tests

dhensby:
A follow up to #6885 

1. Cleanup SapphireTest so it no longer users string FQCNs and usesed `ClassName::class` instead
2. Change format of ss_tmpdb name to be `[prefix]tpmdb_[timestamp]_[rand number]`
3. Don't use `DataObject::get_one()` in `SapphireTest`
4. Fix tests relying on specific capitalisation of am/pm in time formatting tests
5. Restore the phpunit error handler properly

Regarding point 5. PHPUnit only assings it's error handler if the `convertErrorsToExceptions` setting is true (default) so we shouldn't blindly restore it. This patch just sets the handler to default and then resets it to the previous handler (whatever that may have been)

### DIFF
--- a/src/Dev/SapphireTest.php
+++ b/src/Dev/SapphireTest.php
@@ -23,6 +23,7 @@ use SilverStripe\Core\Manifest\ClassManifest;
 use SilverStripe\Core\Manifest\ClassLoader;
 use SilverStripe\Core\Resettable;
 use SilverStripe\i18n\i18n;
+use SilverStripe\ORM\DataExtension;
 use SilverStripe\ORM\SS_List;
 use SilverStripe\Versioned\Versioned;
 use SilverStripe\ORM\DataObject;
@@ -461,7 +462,7 @@ class SapphireTest extends PHPUnit_Framework_TestCase
     public function getFixtureFactory()
     {
         if (!$this->fixtureFactory) {
-            $this->fixtureFactory = Injector::inst()->create('SilverStripe\\Dev\\FixtureFactory');
+            $this->fixtureFactory = Injector::inst()->create(FixtureFactory::class);
         }
         return $this->fixtureFactory;
     }
@@ -538,7 +539,7 @@ class SapphireTest extends PHPUnit_Framework_TestCase
      */
     public function loadFixture($fixtureFile)
     {
-        $fixture = Injector::inst()->create('SilverStripe\\Dev\\YamlFixture', $fixtureFile);
+        $fixture = Injector::inst()->create(YamlFixture::class, $fixtureFile);
         $fixture->writeInto($this->getFixtureFactory());
     }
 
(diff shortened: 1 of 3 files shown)


src/Dev/SapphireTest.php
```diff
@@ -1061,8 +1062,7 @@ public static function using_temp_db()
     {
         $dbConn = DB::get_conn();
         $prefix = getenv('SS_DATABASE_PREFIX') ?: 'ss_';
-        return $dbConn && (substr($dbConn->getSelectedDatabase(), 0, strlen($prefix) + 5)
-            == strtolower(sprintf('%stmpdb', $prefix)));
+        return 1 === preg_match(sprintf('/^%stmpdb_[0-9]+_[0-9]+$/i', $prefix), $dbConn->getSelectedDatabase());
```
sminnee: Safer to use `preg_quote($prefix, '/')` in the sprintf

src/Dev/SapphireTest.php
```diff
@@ -1232,9 +1231,9 @@ public function logInWithPermission($permCode = "ADMIN")
                 $group->Permissions()->add($permission);
             }
 
-            $member = DataObject::get_one('SilverStripe\\Security\\Member', array(
-                '"Member"."Email"' => "$permCode@example.org"
-            ));
+            $member = Member::get()->filter([
```
sminnee: This call won't be cached. get_one will. Not sure if that's a good idea?

sminnee: We made a mistake in 3.0 or 3.1 being gung ho about deprecated get_one without a replacement for its caching behaviour.

src/Dev/SapphireTest.php
dhensby: Good shout

src/Dev/SapphireTest.php
dhensby: After each test `DataObject::singleton()->flushCache();` is called so the cache is discarded for get_ones anyway.

dhensby: Sorry, it's before each test - a few lines before this call

dhensby:
Hmm - looks like behat is sensitive to the naming of the test DB :/

dhensby:
see https://github.com/silverstripe/silverstripe-testsession/pull/48

tractorcow:
2.0.0-alpha6 is also hard-coded in behat-extension. We'll need a release there too.

sminnee:
FYI: we hardcoded the alpha releases rather than using ^ because more often than not, subsequent alphas broke things ;-)

We may want to consider tagging stable releases of our supporting tools if we feel that we're going to stop the API breakages in them.

dhensby:
@tractorcow we are green

tractorcow:
Lots of commits so I used squash.

tractorcow:
Yeah sorry. Maybe I should have left them un-squashed!

Issue #6869, opened 2017-05-04
BUG PHPUnit shiv breaks code inspection in all unit tests


tractorcow:
![image](https://cloud.githubusercontent.com/assets/936064/25687171/4242eef0-30c8-11e7-8de4-268f1e213096.png)

This breaks all code auto-inspection in unit tests.

Related to https://github.com/silverstripe/silverstripe-framework/pull/6833.

Ideal fix would continue to address the original issue without the unfortunate side effects.

chillu:
It's an annoyance, but much less important than other bugs in the beta milestone - removing milestone.

tractorcow:
This file has been removed with the SapphireTest changes a while back; No longer an annoyance!

Issue #6681, opened 2017-03-06
RFC: App object to encapsulate global state and services


chillu:
## Acceptance Criteria

  - A kernel object exists, which can be easily overridden in user code, to control the boot process of any silverstripe application.
  - An application object exists, which can be easily overridden in user code, to control the execution of any request in a silverstripe application.
  - All components which make up the state of the Kernel object are available via Injector service, even if these components are initialised pre-config (including the Kernel itself).
  - Kernel has no static members.
  - The Kernel object can be nested (and un-nested), which internally nests all core services, as well as injector / config. However, this nesting is performed per-instance and not via static accessors as with config / injector.
  - Cookie, Requirements, Config::inst() and Injector::inst() functionality remains unaltered.
  - HTTPRequest is not a part of kernel, nor are any request-specific services.
  - Session should accessible through the HTTPRequest (no longer static).
  - Development experience should not be greatly adversely affected for classes which use the main core traits.
  - If a non-current instance of Injector / Config is ever invoked, it must raise a logic error to enforce application consistency.
  - The concept is applied to all core modules
  - The upgrade path is clearly documented

## Excludes

 * Converting static methods on core classes to instance methods
 * Converting core classes to interfaces
 * Global variable state (e.g. $_SESSION, $_REQUEST) for initial implementation.

## Notes

 * PSR-11 [discourages](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-11-container-meta.md#4-recommended-usage-container-psr-and-the-service-locator) the direct use of service locators in implementations
 * Symfony exposes [services](http://symfony.com/doc/current/service_container.html) service locators (e.g. `$this->app->get('my-dependency')`)

## RFC Versions

 - V1 by @sminnee - https://gist.github.com/sminnee/7fd620a026d551730a088d5e82e542f0
 - V2 by @tractorcow -   https://gist.github.com/tractorcow/c0656a334c13eac8ae252610d226a1c6
 - V3 by @tractorcow / @chillu -
 https://gist.github.com/tractorcow/8be54d6b019412c037049c8caddf39eb
 - V4 by @tractorcow - https://gist.github.com/tractorcow/2bb76d12074b885b9b2923dafa07ae88

## Related

 * [Issue: Tighten up usage of dependency injection in core](https://github.com/silverstripe/silverstripe-framework/issues/1678)
 * [RFC: Make Injector comply with PSR-11](https://github.com/silverstripe/silverstripe-framework/issues/6594)
[...]

PR #6899, opened 2017-05-09
Remove reliance on SapphireTest::is_running_test

dhensby:
A first attempt at removing the coupling of SapphireTest with core code.

This should allow us to remove the `PHPUnitShim` file

### DIFF
--- a/src/Control/Director.php
+++ b/src/Control/Director.php
@@ -8,7 +8,6 @@ use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Config\Configurable;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Dev\Deprecation;
-use SilverStripe\Dev\SapphireTest;
 use SilverStripe\ORM\ArrayLib;
 use SilverStripe\ORM\DataModel;
 use SilverStripe\Versioned\Versioned;
@@ -101,6 +100,13 @@ class Director implements TemplateGlobalProvider
      * @var string
      */
     protected static $environment_type;
+
+    /**
+     * Flag if the system is under test
+     *
+     * @var bool
+     */
+    protected static $system_under_test = false;
 
     /**
      * Process the given URL, creating the appropriate controller and executing it.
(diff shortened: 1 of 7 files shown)


tractorcow:
Also https://github.com/silverstripe/silverstripe-framework/issues/6869

src/Control/Director.php
```diff
@@ -1207,6 +1213,25 @@ public static function session_environment()
     }
 
     /**
+     * Determines if unit tests are currently run, flag set during test bootstrap.
```
sminnee: More Director bloat feels like a step in the wrong direction.

tractorcow:
> It probably would logically live in the same place as wherever isDev() / isTest() / isLive() belong, which should also be refactored out of director.

It needs to go into App object doesn't it @sminnee ? What should we do in the mean time; Add it as a future story for once app is in core?

sminnee:
> It needs to go into App object doesn't it @sminnee ? What should we do in the mean time; Add it as a future story for once app is in core?

I guess that would be the logical place — maybe just note that this method should be shifted on the App object card? Or park this work until App object is done?

tractorcow:
I think parking it is best for the sake of velocity.

dhensby:
I wanted to get this in place to be able to remove our PHPUnit shim - if the App object is likely to come soon then we can put this on hold

tractorcow:
It's marked for beta1. https://github.com/silverstripe/silverstripe-framework/issues/6681

Once I finish the injector service upgrade work I'll jump onto it.

tractorcow:
@dhensby might be worth thinking about what other parts of Director should be in app :)

dhensby:
> @dhensby might be worth thinking about what other parts of Director should be in app :)

all of it? :P

tractorcow:
Well, how about anything not routing / url related?

 - url generation, yes?
 - environment state, no
 - routing, yes
 - file paths/handling, no (we need a new director equivalent for file paths IMO).

tractorcow:
This API has been superseded by already merged changes. is_running_tests() has been removed, rather than replaced. :) Any code which relies on configurable behaviour has been moved to be controlled by config, which tests can set on an as-needed basis.

PR #6901, opened 2017-05-09
BUGFIX: Prevent infinite loop

mikenz:
Prevent infinite loop when a < is present without a matching >


mikenz:
The github online editor doesn't have fancy options like that. 

Whitespace now cleaned up.

PR #6902, opened 2017-05-10
Pulls/4/remove jsmin

unclecheese:


tractorcow:
Please remove the actual JSMinifier class too.

unclecheese:
@chillu That makes sense, but minification lends itself nicely to a single injectable service. If we're just talking about "processing," it seems to me that the API needs to allow for multiple services to be injected.

```
  SilverStripe\View\Requirements_Backend:
    properties:
      processors:
        - %$MinificationService
        - %$VariableReplacementService
        - ....
```

Or we could keep it simple and if the user wants multiple processors, you just chuck them all in one processor service, but that limits configurability.

tractorcow:
Given we are removing and making a core function injectable, I don't think it's necessary to rename the service.

chillu:
It seems weird that we creating new APIs for an approach we want to phase out. But I take your point about not over-optimizing this at the moment with creating a new processor API. How do you feel about marking this API as deprecated for 5.x?

tractorcow:
> It seems weird that we creating new APIs for an approach we want to phase out. But I take your point about not over-optimizing this at the moment with creating a new processor API. How do you feel about marking this API as deprecated for 5.x?


Great idea. @unclecheese can you action this?

unclecheese:
Sure thing.

PR #6903, opened 2017-05-10
Find root modules with _config.php

chillu:
When modules are installed as the webroot,
manifest generation should behave the same way as when they're in a subfolder.
Which means accepting the module folder both with a _config/ folder
and a _config.php file present.


tractorcow:
Please fix the test and we can merge.

tractorcow:
because you can't commit an empty `_config` folder to git. :)

tractorcow:
Rebased; Hope this greens!

PR #6905, opened 2017-05-11
Fix for void element closing tags

axllent:
Prevent html errors when FormField::create_tag('meta') is called from $MetaTags() so 
```
<meta name="generator" content="SilverStripe - http://silverstripe.org"></meta>
```
becomes
```
<meta name="generator" content="SilverStripe - http://silverstripe.org" />
```

### DIFF
--- a/src/Forms/FormField.php
+++ b/src/Forms/FormField.php
@@ -333,7 +333,13 @@ class FormField extends RequestHandler
             }
         }
 
-        if ($content || $tag != 'input') {
+        if (
+            $content ||
+            !in_array($tag, array(
+                'area', 'base', 'br', 'col', 'embed', 'hr', 'img',
+                'input', 'link', 'meta', 'param', 'source', 'track', 'wbr'
+            ))
+        ) {
             return sprintf(
                 '<%s%s>%s</%s>',
                 $tag,


src/Forms/FormField.php
```diff
@@ -333,7 +333,7 @@ public static function create_tag($tag, $attributes, $content = null)
             }
         }
 
-        if ($content || $tag != 'input') {
+        if ($content || !in_array($tag, array('input', 'meta'))) {
```
dhensby: Can we add all the void elements to this array? see http://w3c.github.io/html/syntax.html#void-elements

tractorcow:
This method is getting a refactor soon.

See https://github.com/silverstripe/silverstripe-framework/pull/6722

The solution in my Pr was to explicitly pass in `false` to body to force a self-closing tag.

tractorcow:
> Can we add all the void elements to this array? see http://w3c.github.io/html/syntax.html#void-elements

If this is html version specific (4 vs 5 vs 5.1 vs 5.2) my suggestion is to abstract that into the HTMLValue class (and html version specific subclasses). Do you think this is a good approach?

dhensby:
> If this is html version specific (4 vs 5 vs 5.1 vs 5.2) my suggestion is to abstract that into the HTMLValue class (and html version specific subclasses). Do you think this is a good approach?

The HTML specs only ADD new tags, not remove them (well, they get deprecated) but we won't see tags suddenly become non-void when they once were so I think it's fine to keep  a complete list and not worry about grouping by HTML version. HTML is a robustly backwards compatible spec.

axllent:
@dhensby No problem, and you're right, all those elements should be in there. I've updated / squashed and re-pushed, so (assuming I did it correctly) the patch should have updated into one single merge request. If not then please yell.

tractorcow:
Change is good with me then, after tests are added and fixed. :D

axllent:
There aren't any php linting issues though?

As for the unit test, shouldn't that be done as a separate merge, as there doesn't even appear to be one for `input` either - or did I miss something?

tractorcow:
This is the issue I noted.

```
The command "if [[ $BEHAT_TEST == cms ]]; then vendor/bin/behat @cms; fi" exited with 0.
53.33s$ if [[ $PHPCS_TEST ]]; then composer run-script lint; fi
> phpcs --standard=tests/phpcs/ruleset.xml src/ tests/php
FILE: ...uild/silverstripe/silverstripe-framework/src/Forms/FormField.php
 336 | ERROR | [x] Expected 0 spaces after opening bracket; newline
     |       |     found
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
Time: 52.6 secs; Memory: 59.75Mb
Script phpcs --standard=tests/phpcs/ruleset.xml src/ tests/php handling the lint event returned with error code 2
```

Regarding tests, there may be some tests that cover this indirectly in forms, but any new change should probably include at least one test change to demonstrate a previously un-tested error condition is now tested (in this case, meta tags).

dhensby:
I merged this by hand, fixing the linting issues and adding a test

Issue #6439, opened 2016-12-22
Button (with loading indicator) while loading doesn't hold its width


clarkepaul:
#The page save and publish buttons have loading indicators added as part of https://github.com/silverstripe/silverstripe-framework/pull/6434 but the button width is not retained when clicked.

There are two ways, set visibility of text to "0" or check width of button before its clicked and set it at that width.

In campaigns we are already are using buttons with an additional span around the text and setting visibility to hidden.

## Pull requests

- [ ] https://github.com/silverstripe/silverstripe-admin/pull/62
- [ ] https://github.com/silverstripe/silverstripe-cms/pull/1819
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/6907

clarkepaul:
![button-loading](https://cloud.githubusercontent.com/assets/555033/24382090/8fdb24b4-13b0-11e7-9bc3-425e2ba09c9e.gif)

tractorcow:
I've left some feedback on the action templates.

PR #6907, opened 2017-05-12
Fix update ss template FormActions

flamerohr:
To be closer to React's implementation

Fixes https://github.com/silverstripe/silverstripe-framework/issues/6439

Dependencies
https://github.com/silverstripe/silverstripe-cms/pull/1819
https://github.com/silverstripe/silverstripe-admin/pull/62

### DIFF
--- a/templates/SilverStripe/Forms/FormAction.ss
+++ b/templates/SilverStripe/Forms/FormAction.ss
@@ -1,6 +1,6 @@
 <% if $UseButtonTag %>
 	<button $AttributesHTML>
-		<% if $ButtonContent %>$ButtonContent<% else %>$Title.XML<% end_if %>
+		<% if $ButtonContent %>$ButtonContent<% else %><span>$Title.XML</span><% end_if %>
 	</button>
 <% else %>
 	<input $AttributesHTML />


tractorcow:
ButtonContent is raw HTML: Should we only add the span if using default $Title.XML field?

flamerohr:
@tractorcow good point, just looking at the React `FormAction`, looks like that doesn't support anything but simple text for title/content.

I'll update both to be similar in behaviour

templates/SilverStripe/Forms/FormAction.ss
```diff
@@ -1,6 +1,6 @@
 <% if $UseButtonTag %>
 	<button $AttributesHTML>
-		<% if $ButtonContent %>$ButtonContent<% else %>$Title.XML<% end_if %>
+		<span><% if $ButtonContent %>$ButtonContent<% else %>$Title.XML<% end_if %></span>
```
tractorcow: (Updating comment to be review)

If we set html directly via $ButtonContent, then we shouldn't wrap that in a span.

tractorcow:
> @tractorcow good point, just looking at the React FormAction, looks like that doesn't support anything but simple text for title/content.

We don't need to support react buttoncontent just yet, but we can't break non-react form actions.

flamerohr:
@tractorcow I've updated the templates not wrap `$ButtonContent` in a `<span>` and React now uses `castStringToElement()` instead of hardcoded `<span>`

Issue #5462, opened 2016-05-06
New flysystem abstraction doesn't handle large files locally


madmatt:
Using the built-in `Local` flysystem adapter, it looks like all assets are passed through a SilverStripe controller, regardless of whether they are protected or not. This is a potential performance problem.

Beyond that though, the `FlysystemAssetStore` is calling `read()` which reads the entire file into memory before creating an `SS_HTTPResponse` to send the response to the user. This is bad for large files, as you need a `memory_limit` bigger than the largest asset that you have on your site.

Code is here: https://github.com/silverstripe/silverstripe-framework/blob/master/filesystem/flysystem/FlysystemAssetStore.php#L816

We could maybe just use PHP's native `readfile()` function directly, or look at using `FileSystemInterface->readStream()` (https://github.com/thephpleague/flysystem/blob/master/src/FilesystemInterface.php#L36)

chillu:
@micmania1 You raised this before as well, right? Was there any issues raised for it on Github?

micmania1:
It was discussed on the main filesystem ticket. Something like stream_copy_to_stream() would work here as the data can be passed straight to `STDOUT` instead of being stored in memory. @tractorcow was talking about implementing something like this in `SS_HTTPResponse` I believe.

This still leaves a problem where large files will take a long time to download so max time limit could also be an issue here.

From what I remember, unprotected files weren't being passed through SilverStripe but I could be wrong. That would seem like a separate problem if true.

tractorcow:
I've a stash of some WIP for a streamed file response object. See https://github.com/tractorcow/silverstripe-framework/commit/afb65bd003205edadd7fb349e9ecf40caed042b2

Our internal jira ticket for this feature is at https://silverstripe.atlassian.net/browse/OSS-1365

chillu:
There's discussion on how to generally improve file access performance through dynamic requests: https://github.com/silverstripe/silverstripe-framework/issues/6457

tractorcow:
Fix PRs:

 - [ ] https://github.com/silverstripe/silverstripe-framework/pull/6909
 - [ ] https://github.com/silverstripe/silverstripe-assets/pull/18

PR #6909, opened 2017-05-12
API Add streamable response object

tractorcow:
Part 1/2 of fix for https://github.com/silverstripe/silverstripe-framework/issues/5462

Part 2 at https://github.com/silverstripe/silverstripe-assets/pull/18

### DIFF
--- a/tests/php/Control/HTTPStreamResponseTest.php
+++ b/tests/php/Control/HTTPStreamResponseTest.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace SilverStripe\Control\Tests;
+
+use SilverStripe\Control\HTTPStreamResponse;
+use SilverStripe\Dev\SapphireTest;
+
+class HTTPStreamResponseTest extends SapphireTest
+{
+    /**
+     * Test replaying of stream from memory
+     */
+    public function testReplayStream()
+    {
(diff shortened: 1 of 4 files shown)


tests/php/Control/HTTPStreamResponseTest.php
```diff
@@ -0,0 +1,61 @@
...
+        ob_start();
+        $response->output();
+        $result = ob_get_contents();
+        ob_end_clean();
+
+        $this->assertEquals(12, $response->getHeader('Content-Length'));
+        $this->assertEquals("Test output\n", $result);
+        $this->assertEmpty($response->getSavedBody(), 'Body of seekable stream is un-cached');
+
+        // Seekable stream can be repeated
```
flamerohr: maybe add a `$metadata = stream_get_meta_data($stream); assertTrue($metadata['seekable']);`?
I don't really understand how the stream is seekable, so running the test twice doesn't seem intuitive to me currently.

And is there a case for when seekable is false?

tractorcow: I couldn't figure out how to mock an unseekable stream, but I will add the suggested assertion.

If the stream is seekable, the test is to ensure it doesn't unnecessarily cache the body on a second invocation.

tractorcow:
I've added that assertion; I don't think tests will pass for a while until travis gets better. =(

PR #6911, opened 2017-05-12
Fix numeric field localisation on validation error

### LINKED ISSUE #6910: [FORMS] Invalid form causes NumericField with localised number format to be populated in wrong format after validation
If a form validation fails, the form is populated with the submitted values. The submitted values are determined by calling the [getData method](https://github.com/silverstripe/silverstripe-framework/blob/3.5.3/forms/Form.php#L1523). This method iterates over all data fields and determines their value by calling dataValue. The NumericField returns as dataValue the normalised format of its value (e.g. '4.2'). The type of the returned value is a string (as defined in the PHPDocs). When loading this value into the form in [setupFormErrors](https://github.com/silverstripe/silverstripe-framework/blob/3.5.3/forms/Form.php#L309), the setValue method in NumericField gets passed the normalised value as string as well as the array of all submitted data. As the value is neither of type int or float, nor is the data passed as second argument a DataObject, the [non localised value is set](https://github.com/silverstripe/silverstripe-framework/blob/3.5.3/forms/NumericField.php#L33). This results in NumericFields, where all values are non localised. Saving the form again results in a form validation error, as the non-localised values are not valid.

This is a 3.5.x issue.

dhensby:
Fixes #6910

### DIFF
--- a/forms/NumericField.php
+++ b/forms/NumericField.php
@@ -30,7 +30,7 @@ class NumericField extends TextField {
 		// If passing in a non-string number, or a value
 		// directly from a DataObject then localise this number
 
-		if(is_int($value) || is_float($value) || $data instanceof DataObject) {
+		if(is_numeric($value) || is_int($value) || is_float($value) || $data instanceof DataObject) {
 			$locale = new Zend_Locale($this->getLocale());
 
 			$this->value = Zend_Locale_Format::toNumber(


sb-relaxt-at:
I am afraid this breaks functionality, as surfaced in some testcases. Example given:
If someone passes '42.000' with local 'de' this means '42000' but would be interpreted as '42'.

I have two different suggestions for fixing #6910 (not sure if they cause any other unwanted side effects): 
1. Return a float/int from NumericField#dataValue
2. When loading the data into the form do not use the raw values, but the localised values by accessing 'Value' instead of 'dataValue'

dhensby:
Hmm, yes. Would you mind creating an alternative PR?

forms/NumericField.php
```diff
@@ -30,7 +30,7 @@ public function setValue($value, $data = array()) {
 		// If passing in a non-string number, or a value
 		// directly from a DataObject then localise this number
 
-		if(is_int($value) || is_float($value) || $data instanceof DataObject) {
+		if(is_numeric($value) || is_int($value) || is_float($value) || $data instanceof DataObject) {
```
tractorcow: The way it was is intentional; A numeric string should be parsed in the current locale. the is_int case is to explicitly capture REAL integers that do not not need to be parsed.

sb-relaxt-at: I do agree, as mentioned above, changing this causes issues with localisation. I will try to suggest a different PR.

tractorcow:
We could fix this by making the session errors data array hold the submitted value, rather than the data value. It needs a bit more investigation.

sb-relaxt-at:
I have tried this (option 2) and in my very limited testing it seemed to work properly. If you basically approve this idea, I can prepare a PR.

tractorcow:
I think you should look at your `2.` option:

> Return a float/int from NumericField#dataValue (which feels like a workaround...)

Perhaps this is what we should do in the first place?

dhensby:
`dataValue` should return an int/float and `Value` should return the localised string (what should display in the field)...

sb-relaxt-at:
I do agree, this seems to be the best fix. It just seemed curious to have a subclass of TextField return a float/int; but it works the same way in CurrencyField.

Regarding the idea of populating the form by using Value instead of dataValue - any thoughts about that?

dhensby:
> Regarding the idea of populating the form by using Value instead of dataValue - any thoughts about that?

If it's not doing that already, don't worry about it (unless you have to get it working)

PR #6914, opened 2017-05-14
Fix translations will use namespaced keys

flamerohr:
Related to changes for https://github.com/silverstripe/silverstripe-cms/pull/1815 and https://github.com/silverstripe/silverstripe-admin/pull/58


tractorcow:
moved strings out of framework module; This PR isn't needed.

PR #6915, opened 2017-05-15
TreeDropdownField API for root option

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/217

Adds API required by `TreeDropdownField` to allow for a "root" option, rather than implicitly select root when clearing field value

Related pull requests
https://github.com/silverstripe/silverstripe-admin/pull/67
https://github.com/silverstripe/silverstripe-asset-admin/pull/476

### DIFF
--- a/src/Forms/TreeDropdownField.php
+++ b/src/Forms/TreeDropdownField.php
@@ -69,19 +69,14 @@ class TreeDropdownField extends FormField
     );
     
     /**
-     * Title to display when the value is empty
-     *
      * @var string
      */
-    protected $emptyTitle = null;
+    protected $emptyString = null;
     
     /**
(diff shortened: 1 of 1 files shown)


src/Forms/TreeDropdownField.php
```diff
@@ -649,14 +666,64 @@ public function getSchemaDataDefaults()
...
+        
+        return $this;
+    }
+    
+    /**
+     * Get whether to show the Empty title as the option for root
+     *
+     * @return bool
+     */
+    public function getShowRootOption()
```
tractorcow: I think this is a duplication of `getHasEmptyDefault`, can we consolidate the API here?

flamerohr: sure, good idea

Issue #6642, opened 2017-02-22
Consistently reference injector services in config


chillu:
## Acceptance Criteria

 * As a SilverStripe developer, I can predict the config path of a value if I know the PHP class/interface it relates to 
 * As a SilverStripe developer, I can determine any naming changes from the upgrade guide
 * Where existing config keys are changed and match class names, the silverstripe/upgrader will automatically rewrite my YAML configurations

## Notes

 * YAML rewriting is [already implemented](https://github.com/silverstripe/silverstripe-upgrader/blob/master/src/UpgradeRule/YML/UpdateConfigClasses.php) in silverstripe/upgrader

## Background

At the moment we're using different ways to name injectable services:

- Short concrete class names without namespaces (`FormSchema`)
- FQCN of PHP interfaces (`SilverStripe\Control\Cookie_Backend`, `SilverStripe\Control\Email\Mailer`)
- Aliases without a direct corelation to a class (`FlysystemPublicAdapter`, `ErrorHandler`)

This will be confusing to SS devs, and make it harder to convince them that DI is a good idea in general.

@sminnee and @tractorcow think service names should be FQCN of PHP interfaces, which means creating those interfaces where they're missing.

Any `Injector::inst()` uses should refer to the FQCN through a `use` statement, combined with `::class`. Example:

```php
use SilverStripe\Assets\Storage\AssetStore;

$store = Injector::inst()->create(AssetStore::class);
```

## Exclusions

 * Create new PHP interfaces (use base classes if none exists)
 * Create service name aliases for FQCN
 * Change handling of dependencies as instance properties
 * Upgrader changes (relies on classname rename functionality in YAML)
 
# Tasks

 - [x] Upgrade YML files for each core module (including modules installed as a part of core tests such as testsession).
 - [x] Upgrade all references to named configs
 - [x] Ensure all renamed services are registered with `.upgrade.yml` and will be invoked via upgrader (exclusions pending).
 
# Pull Requests

Note: Merge in order for best results

 - [x] https://github.com/silverstripe/silverstripe-assets/pull/19
 - [x] https://github.com/silverstripe/silverstripe-behat-extension/pull/157
 - [x] https://github.com/silverstripe/silverstripe-framework/pull/6916
 - [x] https://github.com/silverstripe/silverstripe-cms/pull/1821
 - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/477

tractorcow:
[...]

PR #6916, opened 2017-05-15
API Rename services to match FQN of interface / classes

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/issues/6642 for parent issue

Relies on changes to silverstripe-assets, and is a dependency for changes in cms and asset-admin modules (coming up).

Includes missing test coverage for SilverStripe\Logging namespace.

Fixes some issues with CLI debug view echoing out HTML incorrectly.

I suggest to merge https://github.com/silverstripe/silverstripe-assets/pull/19 first and restart framework tests.

After framework, restart https://github.com/silverstripe/silverstripe-cms/pull/1821 tests.

### DIFF
--- a/src/Control/HTTPResponse.php
+++ b/src/Control/HTTPResponse.php
@@ -3,6 +3,8 @@
 namespace SilverStripe\Control;
 
 use InvalidArgumentException;
+use Monolog\Formatter\FormatterInterface;
+use Monolog\Handler\HandlerInterface;
 use SilverStripe\Core\Convert;
 use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\Core\Injector\Injector;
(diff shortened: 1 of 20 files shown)


src/Control/HTTPResponse.php
```diff
@@ -3,6 +3,8 @@
 namespace SilverStripe\Control;
 
 use InvalidArgumentException;
+use Monolog\Formatter\FormatterInterface;
```
flamerohr: I don't see `FormatterInterface` used anywhere in the file, perhaps remove it?

tractorcow: It can be removed.

tractorcow:
Only CMS tests are failing; We can merge this now.

tests/php/Core/Cache/CacheTest.php
```diff
@@ -18,38 +17,29 @@ protected function setUp()
...
-                ]
-            )
-            ->set(
-                Injector::class,
-                CacheInterface::class . '.TestApcuCache',
-                [
+        Injector::inst()
+            ->load([
+                ApcuCacheFactory::class => [
+                    'constructor' => [ 'version' => 'ss40test' ]
```
chillu: What's the deal with replacing `version => 4400` with `version => ss40test`?

tractorcow: Because apcucache uses version as a cache prefix, and it physically breaks on numeric keys.

PR #6917, opened 2017-05-15
Fix for prototype `evalResponse` detection

oilee80:
If the response header has a `chartset` in the *Content-Type* then it doesn't match and `evalResponse()` doesn't get run


tractorcow:
Sorry, we are not accepting any more pull requests for 2.x, but thank you for the contribution anyway.

PR #6918, opened 2017-05-15
Check for \r\n as well as just \n\n in FirstParagraph

hailwood:
Makes the `FirstParagraph` check a bit more robust depending on the OS used when creating the content.

### DIFF
--- a/model/fieldtypes/Text.php
+++ b/model/fieldtypes/Text.php
@@ -203,7 +203,10 @@ class Text extends StringField {
 			if(!$data) return "";
 
 			// grab the first paragraph, or, failing that, the whole content
-			$pos = strpos($data, "\n\n");
+			$pos = max(array(
+				strpos($data, "\n\n"),
+				strpos($data, "\r\n"),
+			));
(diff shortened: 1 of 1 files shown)


model/fieldtypes/Text.php
```diff
@@ -203,7 +203,10 @@ public function FirstParagraph($plain = 1) {
 			if(!$data) return "";
 
 			// grab the first paragraph, or, failing that, the whole content
-			$pos = strpos($data, "\n\n");
+			$pos = max(array(
+				strpos($data, "\n\n"),
```
tractorcow: I'm not keen on using max to filter out false values, given that you are also applying this to non-falsey values (max of 20 and 30 is 30, when we want 20).

I suggest using Convert::nl2os($data, "\n") to normalise newlines prior to this operation, and then splitting.

Also please note the logical error in your code; `\r\n` is a single line break in some OS, when we want to detect the first line break with 2 or more breaks.

tractorcow: Please also add a unit test. :)

dhensby:
closing as inactive

PR #6931, opened 2017-05-18
API PSR-11 compliance (fixes #6594)

### LINKED ISSUE #6594: RFC: Make Injector comply with PSR-11
PSR-11 covers service containers for dependency injection; Injector is one of these.

https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-11-container.md
https://packagist.org/packages/psr/container

My short-term proposal is that we update Injector to make use of this interface:

 * Replace `hasService()` with a boolean-returning `has()`.
 * Allow any `Psr\Container\ContainerInterface` to be passed to `Injector::set_inst()`

My long-term proposal is that we remove usage of the Injector that doesn't comply with this PSR:

 * Shift away from widespread use of `Injector::create()`, `Injector::createWithArgs()`, and `Injectable::create()`, instead favour factories that are provided by `Injector::get()`.

This isn't something that I would expect to see ready by SS4 but that perhaps by SS5 we can deprecate these methods.

The benefits of these changes are as follows:

 * `Injector` can be passed to other PSR-11-compliant frameworks to provide services. My short-term proposal allows for this.
 * Other PSR-11-complaint frameworks. My long-term proposal allows for this.

## Alternatives

An alternative approach to `create()` and `createWithArgs()` would be to let developers supply both a PSR-11-compliant container and a "Generic Factory". `get()` would be routed to the container and `create()` / `createWithArgs()` would be routed to the generic factory.

As a tweak on this, GenericFactory could be shifted to being a service that the parent Injector provides.

```php
Injector::inst()->get('GenericFactory')->create('<class>');
```

This would make it easier to make use of 3rd party service containers before the long-term factory has completed. However, it's a bit messier.

chillu:
Note that our usage of `$asSingleton` in `get()` is fine. Quote from the PSR:

> Two successive calls to get with the same identifier SHOULD return the same value. However, depending on the implementor design and/or user configuration, different values might be returned, so user SHOULD NOT rely on getting the same value on 2 successive calls.

### DIFF
--- a/src/Core/Injector/Injector.php
+++ b/src/Core/Injector/Injector.php
@@ -2,11 +2,14 @@
 
 namespace SilverStripe\Core\Injector;
 
+use Psr\Container\NotFoundExceptionInterface;
 use SilverStripe\Core\Config\Config;
 use ReflectionProperty;
 use ArrayObject;
 use ReflectionObject;
 use ReflectionMethod;
+use Psr\Container\ContainerInterface;
+use SilverStripe\Dev\Deprecation;
 
 /**
  * A simple injection manager that manages creating objects and injecting
@@ -122,7 +125,7 @@ use ReflectionMethod;
  * @author marcus@silverstripe.com.au
  * @license BSD License http://silverstripe.org/bsd-license/
  */
-class Injector
+class Injector implements ContainerInterface
 {
 
     /**
@@ -233,7 +236,7 @@ class Injector
      * If a user wants to use the injector as a static reference
      *
      * @param array $config
-     * @return Injector
+     * @return ContainerInterface
      */
     public static function inst($config = null)
     {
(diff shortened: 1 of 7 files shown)


src/Core/Injector/Injector.php
```diff
@@ -876,8 +909,10 @@ public function unregisterAllObjects()
      */
     public function get($name, $asSingleton = true, $constructorArgs = null)
     {
+        $object = null;
```
chillu: I'm not too happy with all the `if(!$object)` conditionals, but the only alternative I can see is extracting the `get()` implementation into a new protected method which is then called by a much leaner `get()`

tractorcow: I would highly recommend to do that; It'll result in more readable code and less extraneous conditionals through the code, not to mention we can un-do all this nesting.

tractorcow: Copied from slack:

abstract it to
 - find service spec (either by hasService() or locateConfigFor()
 - construct found service

Turn

```if (A) {
 construct(A)
}
elseif (B) {
  construct(B)
}
```

into

```spec = A || B
construct(spec)
```

you know that Injector::get() (current implementation) calls ->instantiate() five times right?

tractorcow:
Looks good to merge. @sminnee do you want to weigh in on the new PSR interface?

PR #6934, opened 2017-05-19
API Consistent use of inst() naming across framework

### LINKED ISSUE #6851: Consistent use of inst() naming across framework
We've historically been using `inst()` as a short name for `instance()` starting with the widely used `Injector` class some time in 3.x. Since then, more classes have been introduced in 4.x with the long form of `instance()`, making APIs inconsistent. 

 * ThemeResourceLoader
 * ModuleLoader
 * ConfigLoader
 * ClassLoader

While this should all be unnecessary through proper injector use and the new [app object](https://github.com/silverstripe/silverstripe-framework/issues/6681). that's unlikely to be fully realised in 4.x - so we should avoid API inconsistencies in this release.

The consensus from a Slack conversation on `#core-committers` is that we should stick with `inst()`.

```
@channel for static access to singleton objects, do we want to continue with the pattern `Injector::inst()`switch to `Injector::instance()`? We’ll be looking to reduce the amount that it’s necessary to use it but it will probably still be hanging around a bit in SS4

:+1: for inst() - 5
:-1: for instance() - 3
 

unclecheese [9:27 AM] 
So, one more API change, more upgrade pain.. what's the benefit?

tractorcow [9:28 AM] 
we also have `::singleton()`

[9:29] 
Does your change cover this?

sminnee [9:41 AM] 
@tractorcow it’s your change, not mine :wink:

[9:41] 
my recommendation is that we leave it as `inst()` you were recommending that we change it to instance.

[9:42] 
@unclecheese in fairness, by the time SS4 goes stable I’m expecting it’d be `App::inst()->getInjector()` anyway. We might then have `Injector::inst()` as a deprecated fallback, though

unclecheese [9:43 AM] 
yeah, i got caught up on the internal thread.. makes sense

sminnee [9:43 AM] 
and having `Injector::instance()` as a deprecated fallback would make no sense whatsoever

unclecheese [9:43 AM] 
Given that, my vote is for inst()

sminnee [9:43 AM] 
so you’d have `App::instance()` and `Injector::inst()` if we changed.

[9:44] 
Unless we deleted `Injector::inst()` altogether and force the use of `App::inst()->getInjector()`

unclecheese [9:45 AM] 
So it's basically an API change for something that will be deprecated, or at least discouraged

[9:45] 
I don't see the point in that

sminnee [9:45 AM] 
More or less, yep.

tractorcow [9:45 AM] 
Removing things in 4.0 is fine :smile:

[9:46] 
how would config look?

[9:46] 
We have Config class (handles nesting) and ConfigCollectionInterface (collection at each nesting level) (edited)

tractorcow [10:02 AM] 
`App::inst()->getConfig()->nest()`
`App::inst()->getConfig()->get('Class', 'property')`

sminnee [10:08 AM] 
I would expect that `nest()` would shift to the application object, but we’d need to validate what the use-case of nesting actually is

[10:08] 
My understanding was that it was pervasive things like test-execution (edited)

unclecheese [10:16 AM] 
it's mostly for testing

dhensby 
[8:54 PM] 
> by the time SS4 goes stable I’m expecting it’d be `App::inst()->getInjector()` anyway. We might then have `Injector::inst()` as a deprecated fallback, though
```

robbieaverill:
Issue: #6851

## Blockers

* [x] https://github.com/silverstripe/silverstripe-behat-extension/pull/158 (also requires this PR)

## Blocks

* [ ] https://github.com/silverstripe/silverstripe-admin/issues/79

### DIFF
--- a/docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
+++ b/docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
@@ -212,7 +212,7 @@ You can also combine CSS files into a media-specific stylesheets as you would wi
 the third paramter of the `combine_files` function:
 
 ```php
-$loader = ThemeResourceLoader::instance();
+$loader = ThemeResourceLoader::inst();
 $themes = SSViewer::get_themes();
 
 $printStylesheets = array(
@@ -237,7 +237,7 @@ You can also add the 'async' and/or 'defer' attributes to combined Javascript fi
 `Requirements::javascript` call - use the third paramter of the `combine_files` function:
 
 ```php
-$loader = ThemeResourceLoader::instance();
+$loader = ThemeResourceLoader::inst();
 $themes = SSViewer::get_themes();
 
 $scripts = array(
--- a/docs/en/02_Developer_Guides/16_Execution_Pipeline/02_Manifests.md
+++ b/docs/en/02_Developer_Guides/16_Execution_Pipeline/02_Manifests.md
@@ -37,7 +37,7 @@ It finds the following information:
  * All implementors of an interface
  * All module configuration files
 
-The gathered information can be accessed through [api:ClassLoader::instance()],
+The gathered information can be accessed through [api:ClassLoader::inst()],
 as well as [api:ClassInfo]. Some useful commands of the `ClassInfo` API:
 
  * `ClassInfo::subclassesFor($class)`: Returns a list of classes that inherit from the given class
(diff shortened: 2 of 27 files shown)


robbieaverill:
Will update behat now

tractorcow:
I'm trying to get behat extension working first. Then we'll merge and tag that, then restart this build.

tractorcow:
Can be merged on green.

PR #6937, opened 2017-05-19
CheckboxField creates invalid HTML when required #2939

### LINKED ISSUE #2939: CheckboxField creates invalid HTML when required.
Attribute aria-required not allowed on element input at this point.

…ox" id="ApplicationForm_Form_Terms" required="required" aria-required="true" />

caffeineinc:
Fixes #2939

Semantically, it doesn't make sense to have a required attribute on a 
field in which both checked and unchecked attributes are 'valid',

For checkboxes, required might make more sense where it is the 
equivalent of "check this checkbox" (for example accepting license 
terms).but we can't use the aria-required attribute here. 

For a radio button grouping OptionSet, the "required" attribute applies 
to the group and not to the individual items. It is impossible to 
select all the radio buttons in a group for example but it does make 
sense to have a grouping with no default value and expect the user to 
choose one value. Here we can enforce the user to select an option, 
without selecting one first. 

For Dropdowns, we have to include an empty option, so the user must
 select a valid option from the dropdown, otherwise the first option 
 would be required and might not be a valid user-input. 

https://www.w3.org/TR/wai-aria/states_and_properties#aria-required

relates to #4901

### DIFF
--- a/src/Forms/CheckboxField.php
+++ b/src/Forms/CheckboxField.php
@@ -28,10 +28,16 @@ class CheckboxField extends FormField
 
     public function getAttributes()
     {
-        $attrs = parent::getAttributes();
-        $attrs['value'] = 1;
+        $attributes = parent::getAttributes();
+        $attributes['value'] = 1;
+        if ($this->Required()) {
+            // Semantically, it doesn't make sense to have a required attribute
+            // on a field in which both checked and unchecked are allowable.
+            unset($attributes['aria-required']);
+        }
+
         return array_merge(
-            $attrs,
+            $attributes,
             array(
                 'checked' => ($this->Value()) ? 'checked' : null,
                 'type' => 'checkbox',
--- a/src/Forms/CheckboxSetField.php
+++ b/src/Forms/CheckboxSetField.php
@@ -104,4 +104,18 @@ class CheckboxSetField extends MultiSelectField
     {
         return 'optionset checkboxset';
     }
+
+    public function getAttributes()
+    {
+        $attributes = array_merge(
+            parent::getAttributes(),
+            array('role' => 'listbox')
+        );
+
+        // Remove invalid attributes from wrapper.
+        unset($attributes['name']);
+        unset($attributes['required']);
+        unset($attributes['aria-required']);
+        return $attributes;
+    }
 }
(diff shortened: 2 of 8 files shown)


tractorcow:
I understand that a checkbox can have two valid states, but my understanding is that, in most real world practical examples, a "required" checkbox is one that must be set to true to validate. E.g.

 - [ ] I accept the terms and conditions (required)
 
My understanding is that "required" implies "required to be checked". I wouldn't interpret this as "a value must be selected, whether true / false". However, maybe I misunderstand the specification.

> but we can't use the aria-required attribute here.

Can you please elaborate on why? I had a cursory investigation of the link you provided https://www.w3.org/TR/wai-aria/states_and_properties#aria-required but it wasn't clear to me. :)

I ran the below through https://validator.w3.org/nu/ and it came up clean.

```html
<!DOCTYPE html>
<html>
<head>
<title>Hello</title>
</head>
<body>
<input type="checkbox" aria-required="true" />
</body>
</html>
```

caffeineinc:
Yeah, @tractorcow it's having _both_ required="required" and `aria-required="true"` According to HTML5 + ARIA. The spec says aria-required semantically means the form must submit something valid, when that something has both states regardless of user input. (I know it's confusing). I don't think that validator includes validation of HTML5 + ARIA?  Try: https://html5.validator.nu/  Or something that validates against HTML5 + ARIA 

For this example: The checkbox can still be `required` using required="required" which denotes the requirement of having a `value` (which validates). but can't have aria-required="true", as it doesn't make sense semantically. 

*Note*: the required attribute still prevents submission in this nature, as it must be "checked"

The required instance in this case would output which validates:

 ```<!DOCTYPE html>
<html>
<head>
<title>Hello</title>
</head>
<body>
<input type="checkbox" name="myfield" value="1" class="checkbox" id="Form_form_myfield" required="required"></body>
</html>``` 

Whereas: this is invalid: 

```<!DOCTYPE html>
<html>
<head>
<title>Hello</title>
</head>
<body>
<input type="checkbox" name="myfield" class="checkbox" id="Form_form_myfield" required="required" aria-required="true" />
</body>
</html>``` 

https://html5.validator.nu/  fails correctly with the above: `Error: Attribute aria-required not allowed on element input at this point.`

tractorcow:
That's all I was looking to see; Thanks for reassuring my concerns!

caffeineinc:
@tractorcow updated. Not quite sure how my editor got out of PSR-2 ...

PR #6939, opened 2017-05-19
API Ensure that all DataQuery joins are aliased based on relationship name

tractorcow:
Replaces #5659

How's this @dhensby ?

### DIFF
--- a/src/ORM/DataList.php
+++ b/src/ORM/DataList.php
@@ -543,11 +543,13 @@ class DataList extends ViewableData implements SS_List, Filterable, Sortable, Li
                 $relations = explode('.', $field);
                 $fieldName = array_pop($relations);
 
-                // Apply
+                // Apply relation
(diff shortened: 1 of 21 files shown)


tractorcow:
Postgres fix at https://github.com/silverstripe/silverstripe-postgresql/pull/67

chillu:
@tractorcow Is this an API change, hence need to go in before beta1?

dhensby:
From the look of the tests, this looks good. Did you port over my original failing test to confirm it fixes the issue as reported?

tractorcow:
@chillu yes it changes the way that isJoinedTo will work, since it no longer reports true for the table directly, and you must use the alias to get the correct response.

@dhensby I'll look at the test.

tractorcow:
@dhensby I've migrated your tests over to this PR.

tractorcow:
Now depends on https://github.com/silverstripe/silverstripe-asset-admin/pull/487 which fixes breaking JS changes.

PR #6941, opened 2017-05-21
[Improvement|testing] Use scriptfusion to immediately print PHPUnit output

Firesphere:
Props to @madmatt for the idea.

This change updates the output of the unit tests, to use the pretty-print method, which outputs errors immediately (no waiting for all tests to finish), and show a better progress of the tests.

More information:
https://github.com/ScriptFUSION/PHPUnit-Immediate-Exception-Printer

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,6 +45,7 @@
   },
   "require-dev": {
     "phpunit/phpunit": "^5.7",
+    "scriptfusion/phpunit-immediate-exception-printer": "^1",
     "silverstripe/versioned": "^1.0@dev",
     "silverstripe/behat-extension": "^3",
     "silverstripe/serve": "dev-master",
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -3,7 +3,9 @@
 Standard module phpunit configuration.
 Requires PHPUnit ^5.7
 -->
-<phpunit bootstrap="tests/bootstrap.php" colors="true">
+<phpunit printerClass="ScriptFUSION\PHPUnitImmediateExceptionPrinter\ImmediateExceptionPrinter"
+             bootstrap="tests/bootstrap.php"
+             colors="true">
 	<testsuite name="Default">
 		<directory>tests/php</directory>
     </testsuite>


tractorcow:
I think I should revert this for now; It's had the effect of creating excessively verbose output.

https://github.com/silverstripe/silverstripe-framework/pull/6950

If we can get this to work on errors only that would be good, thanks @Firesphere :)

sminnee:
Or errors-and-warnings (including those warnings about slow tests)

Bilge:
It's not really any of my business, and I'm glad you've taken an interest in the plugin, but what's the problem with verbose output? I understand it might not be what you're used to seeing, but that's the whole point, and it comes with added benefits. For example, it's easy to check a specific test actually ran. Sometimes we might misconfigure PHPUnit or put a test in the wrong directory or give it an incorrect name. With the *verbose* output we can easily search for a test to make sure it was actually executed, just by way of example.

tractorcow:
> It's not really any of my business, and I'm glad you've taken an interest in the plugin, but what's the problem with verbose output? I understand it might not be what you're used to seeing, but that's the whole point, and it comes with added benefits. For example, it's easy to check a specific test actually ran. Sometimes we might misconfigure PHPUnit or put a test in the wrong directory or give it an incorrect name. With the verbose output we can easily search for a test to make sure it was actually executed, just by way of example.

That's ok, it's open source, so we love any feedback and aren't precious about our code. :)

The issue is that our tests runs are 2000+ tests, and we rely on build logs for debugging. When we increase the ratio of useless information / useful debug information, it becomes more maintenance burden. I see the reversal of this code as temporary until we can refine the solution to limit debugging to verbose errors, and non-verbose non-error logging.

All of the use cases you presented are perfectly valid, but they can be run locally by adding the plugin manually, and do not need to be included by default in our CI test suite.

@Bilge It's good to see you taking an interested in our ticket at https://github.com/ScriptFUSION/PHPUnit-Immediate-Exception-Printer/issues/5; As you pointed out, every project seems to have their own vision and goals for their testing suite, and as an open source project we respect that it's important for you to remain true to the goals of your own project, and not try to please everyone. :) We have that same issue, and respect those challenges.

As you quoted:

> I think it should be a priority to create a PR to pass options to printers in PHPUnit so we can configure them to our heart's content and make everyone happy in a maintainable way.

As we have a project have improved our understanding of the community, we have found that making everyone happy is not only impossible at times, it probably shouldn't be a goal at all. A better goal is to allow users to build the tools themselves they need to solve their problems, rather than taking on responsibility for that directly. The challenge here is to learn to address the root problems that our users suffer, rather than trying to simply implement the solutions people want.

I would love to see some of the solutions and suggestions that come out of the above discussion; I share your idea that we should not be suppressing success at all.

Bilge:
>When we increase the ratio of useless information / useful debug information, it becomes more maintenance burden. I see the reversal of this code as temporary until we can refine the solution to limit debugging to verbose errors

We [still get the skips, warnings and errors summary at the end of the test run](https://travis-ci.org/silverstripe/silverstripe-framework/jobs/234464671#L2406) exactly the same as before. If that's all you're interested in just skip to the end encounter no more burden than before.

Issue #6361, opened 2016-12-04
Fatal error on login if SS_DEFAULT_ADMIN_PASSWORD is unset


JorisDebonnet:
This was rather counter-intuitive for me... 

This is the first project I started using `_ss_environment.php` on - certainly handy - but after developing for a while, I didn't particularly like that my password was in that file as plain text (even though it's only on my local machine and it's not checked into any repository), so I just cleared it out. I figured it didn't really need it anymore, as the default admin was already set up in the database. 

But then I started getting `"Fatal error: Call to a member function isLockedOut()"` on MemberAuthenticator->authenticate_member. I saw this the first time after updating to SS 3.5, so of course I first checked that. But it turns out I just wasn't allowed to change that password in `_ss_environment.php`! 

I would like to suggest either making that check less strict, or making the error message more descriptive in a case like this.

tractorcow:
I would say that if SS_DEFAULT_ADMIN_USERNAME is declared but SS_DEFAULT_ADMIN_PASSWORD then it should block access to the admin.

The problem is this

```
// Check default login (see Security::setDefaultAdmin())
        $asDefaultAdmin = $email === Security::default_admin_username();
        if ($asDefaultAdmin) {
            // If logging is as default admin, ensure record is setup correctly
            $member = Member::default_admin();
// ...
```

default_admin_username() returning a value assumes that default_admin() won't return null, in your case it did.

maybe change it to something like

```
// Check default login (see Security::setDefaultAdmin())
$asDefaultAdmin = Security::has_default_admin() && ($email === Security::default_admin_username());
        if ($asDefaultAdmin) {
            // If logging is as default admin, ensure record is setup correctly
            $member = Member::default_admin();
```

dhensby:
so the password for default admins isn't stored in the DB, which I think is the confusion here.

We probably should solve this problem as @tractorcow suggests, but you are doing it "wrong" by not having a password for your default admin.

dhensby:
fixed by #7012

PR #6944, opened 2017-05-22
Fix Member Authentication Edgecase

Cheddam:
There exists a scenario in which:

- You've successfully authenticated to the CMS
- You haven't accessed the CMS for more than 3 days
- You don't have a Default Admin user configured in your environment

In this scenario, trying to re-authenticate via the provided CMS dialog will result in a Fatal Error as of the tip of the SilverStripe 3 branch. This PR shifts the Default Admin authentication attempt inside a conditional to avoid comparing a `null` email address to a `null` default admin username.

This was discussed in #6361.

Test coverage included. 👍


dhensby:
I believe the issue has be fixed by #7012 

However, this has tests, and I like tests. so I'll update this PR just to have the test in it :)

PR #6945, opened 2017-05-22
Revert injector type hint to Injector

tractorcow:
Fixes issues across the framework where `Injector::inst()-create()` doesn't resolve to anything.

### DIFF
--- a/src/Core/Injector/Injector.php
+++ b/src/Core/Injector/Injector.php
@@ -236,7 +236,7 @@ class Injector implements ContainerInterface
      * If a user wants to use the injector as a static reference
      *
      * @param array $config
-     * @return ContainerInterface
+     * @return Injector
      */
     public static function inst($config = null)
     {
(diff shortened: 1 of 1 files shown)


sminnee:
Yeah — I guess this is is an important point: SilverStripe expects more than a PSR-11 object in place in order to function. It's both a DI container (get) and a generic factory (create).

tractorcow:
What suddenly is going on with travis?

image

sminnee:
Simon E added that, check recently merged PRs

tractorcow:
Oh no, what have I done. I may revert that PR. I should have inspected the output in more detail.

tractorcow:
https://github.com/silverstripe/silverstripe-framework/pull/6948 to fix.

sminnee:
You don't like the more verbose output, then?

tractorcow:
I would if it were print on error only lol

sminnee:
Or at least error-or-slow. I wonder if it can be configured?

tractorcow:
According to @Firesphere there is an open issue to get that resolved, but it's not yet possible.

tractorcow:
See https://github.com/ScriptFUSION/PHPUnit-Immediate-Exception-Printer/issues/5

PR #6946, opened 2017-05-22
Bugfix: i18nTextCollector merge direction is incorrect and overwrites existing values

mikenz:
Also update comment that it is actually enabled by default, just wasn't working so no one noticed?

### DIFF
--- a/src/Dev/Tasks/i18nTextCollectorTask.php
+++ b/src/Dev/Tasks/i18nTextCollectorTask.php
@@ -26,7 +26,7 @@ class i18nTextCollectorTask extends BuildTask
 		- locale: Sets default locale
 		- writer: Custom writer class (defaults to i18nTextCollector_Writer_RailsYaml)
 		- module: One or more modules to limit collection (comma-separated)
-		- merge: Merge new strings with existing ones already defined in language files (default: FALSE)
+		- merge: Merge new strings with existing ones already defined in language files (default: TRUE)
 	";
 
     /**
--- a/src/i18n/TextCollection/i18nTextCollector.php
+++ b/src/i18n/TextCollection/i18nTextCollector.php
@@ -378,8 +378,8 @@ class i18nTextCollector
             // Merge
             if ($existingMessages) {
                 $entitiesByModule[$module] = array_merge(
-                    $existingMessages,
-                    $messages
+                    $messages,
+                    $existingMessages
                 );
             }
         }


src/i18n/TextCollection/i18nTextCollector.php
```diff
@@ -378,8 +378,8 @@ protected function mergeWithExisting($entitiesByModule)
             // Merge
             if ($existingMessages) {
                 $entitiesByModule[$module] = array_merge(
-                    $existingMessages,
-                    $messages
+                    $messages,
```
tractorcow: If we do it this way, any existing messages with new defaults won't be updated, won't they?

src/Dev/Tasks/i18nTextCollectorTask.php
```diff
@@ -26,7 +26,7 @@ class i18nTextCollectorTask extends BuildTask
 		- locale: Sets default locale
 		- writer: Custom writer class (defaults to i18nTextCollector_Writer_RailsYaml)
 		- module: One or more modules to limit collection (comma-separated)
-		- merge: Merge new strings with existing ones already defined in language files (default: FALSE)
+		- merge: Merge new strings with existing ones already defined in language files (default: TRUE)
```
tractorcow: Good spotting :)

src/i18n/TextCollection/i18nTextCollector.php
tractorcow: Sorry, looking at the context of the PR title, I think we might have different understandings of this behaviour. The intent of merge is to not delete any keys not present in the new strings (in case older versions still need to localise those values), but not to prevent changed ones from being updated.

If you do merge=0, keys that aren't found are deleted, meaning any subsequent push to transifex deletes both the master, and all localisations for that key. It's normally important to keep those keys if we release, for instance, a new 3.4.x tag, even though a new 3.6.x tag might have removed that key.

mikenz: Correct - you don't want to be merging then.

If I call it /dev/tasks/i18nTextCollectorTask/?module=themes&locale=ja_JP I expect it to keep my existing ja_JP translations and add any new messages with their defaults into the file. Currently it'd lose all my ja_JP translations as they'd be overwritten by the defaults.

mikenz: Oh, this is all about merging keys, I don't care about keys, I care about values. Seems we need a different mode for what I want.

tractorcow: There is a very good case for adding a merge mode parameter. :) Do you want to suggest one?

merge=0 (clear and replace all) (current merge=false)
merge=1 (update and add) (current merge=true)
merge=2 (add new only)

As an example.

dhensby:
@mikenz do you expect to address the feedback in this PR?

DrMartinGonzo:
@dhensby, @mikenz Any news on this ? This is really hard to translate because of this since it resets everything every time :)

dhensby:
I think we may have to intervene to get it over the line.

lekoala:
Meanwhile if someone needs a solution (like I do), I've been building this
https://github.com/lekoala/silverstripe-base/blob/master/code/i18n/TextCollector.php

Should be fairly easy to extract and include in your own projects

maxime-rainville:
Looks like people lost interest. If someone feels like picking this up, just let us know and we'll reopen.

PR #6948, opened 2017-05-22
Revert "[Improvement|testing] Use scriptfusion to immediately print PHPUnit output"

tractorcow:
Reverts silverstripe/silverstripe-framework#6941

I merged this PR a little too hastily, I'm afraid.

PR #6950, opened 2017-05-22
Revert 6941 pulls/phpunit pretty print

tractorcow:
Replaces https://github.com/silverstripe/silverstripe-framework/pull/6948


flamerohr:
We'll bring this back when we can configure it to only give errors+warnings :) @Firesphere

Issue #2494, opened 2013-10-04
Silverstripe 3.1: Warning: posix_getpwuid() has been disabled for security reasons


rpaget:
In Silverstripe 3.1, on some shared hosts the following bug can occur:
Warning: posix_getpwuid() has been disabled for security reasons

Solved by replacing 
framework/core/TempPath.php line 28 with:

if(!$user && function_exists('posix_getpwuid') && function_exists('posix_getuid')) {

PR #6951, opened 2017-05-22
FIX: Don’t assume posix_getpwuid is available.

sminnee:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/2494

In Silverstripe 3.1, on some shared hosts the following bug can occur:
Warning: posix_getpwuid() has been disabled for security reasons


kenlog:
Why was this merger not included in the 3.7 branch? Thank you

PR #6957, opened 2017-05-23
Docs for React DI

unclecheese:
Added docs for React dependency injection.  Dependency: https://github.com/silverstripe/silverstripe-admin/pull/70


docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -204,26 +204,340 @@ To make the actions more user-friendly you can also use alternating buttons as
...
+The remainder of this tutorial is written in [ECMAScript 6](http://es6-features.org/#Constants), or _ES6_
+for short. This is the new spec for Javascript (currently ES5) that is as of this writing
+only partially implmented in modern browsers. Because it doesn't yet enjoy vast native support,
+it has to be [transpiled](https://www.stevefenton.co.uk/2012/11/compiling-vs-transpiling/) in order to work
+in a browser. This transpiling can be done using a variety of toolchains, but the basic
+ principle is that a browser-ready, ES5 version of your code is generated in your dev
+ environment as part of your workflow. Often called a "bundle," you should rarely have 
+ to see this file. It is effectively an invisible layer that translates modern ES6
+ code into something a browser can parse. As browsers evolve, this step will become less
+ necessary. (Although, it is worth noting that because transpiling comes at such a low cost,
```
chillu: Might be a bit wordy for a preface about ES6

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -204,26 +204,340 @@ To make the actions more user-friendly you can also use alternating buttons as
...
+```
+
+Much like the configuration layer, we need to specify a name for this mutation. This
+will help other modules negotiate their priority over the injector in relation to yours.
+
+The second parameter of the `update` argument is a callback which receives a `wrap()` function
+that allows you to mutate the DI container with a wrapper for the component. Remember, this function does not _replace_
+the component -- it enhances it with new functionality.
+
+The last thing we'll have to do is make sure this script gets loaded into the admin
```
chillu: You should note that this code needs to be compiled by a bundler somehow - the earlier pointers might be helpful, but it feels like there's a gap here

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -204,26 +204,340 @@ To make the actions more user-friendly you can also use alternating buttons as
...
+higher order component.
+
+__my-module/js/components/Gallery.js__
+```js
+import React from 'react';
+import { withInjector } from 'lib/Injector';
+
+class Gallery extends React.Component {
+  render() {
+    const GalleryItem = this.context.injector.get('GalleryItem');
```
chillu: It might be useful to show this line earlier on? Up until this point, any React dev will scratch their head on how all these customisations are actually applied.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -204,26 +204,340 @@ To make the actions more user-friendly you can also use alternating buttons as
...
+<EnhancedPhotoItem isNew={true} size={300} />
+```
 
-class MyComponent extends SilverStripeComponent {
+Alternatively, this component could be expressed with an ES6 class, rather than a simple
+function.
+
+```js
+const enhancePhoto = (PhotoItem) => {
+	return class EnhancedPhotoItem extends React.Component {
```
chillu: Oooh, looks like you've done something smart with `createDisplayName` there. Using the module name as a fallback kind of works, but we shouldn't encourage skipping `displayName`. `my-other-module(my-module(TextField)))` isn't as expressive as `TextLengthChecker(CharacterCounter(TextField)))`

unclecheese: Yeah, the problem is, no one uses `displayName`, so this is a bit of a compromise. I'm not sure how we would enforce it, either. Anonymous functions can have `displayNames`, though. 

```
const HOC = (Component) => (props) => ...;
HOC.displayName = 'foo';`
```

Our linting rules have `react/prefer-stateless-functions` enabled, so I'm not sure we're in a position to be recommending ES6 classes?

chillu:
The ACs state "Example for: Block opening the detail form of a gallery tile if the file is in draft mode (event overrides)". Is there an easy way to get an example with an event handler in there?

unclecheese:
@chillu Is this sufficient? https://github.com/open-sausages/experimental-react-di/blob/master/js/src/FormActionAwesomenessCreator.js

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
chillu: Hah, caught out by our own linter rules ;) I don't think we should override the displayName completely, but rather add our own. So `MyEnhancedComponent(Component)`. See https://github.com/apollographql/react-apollo/blob/master/src/graphql.tsx#L135. Maybe we need a `getDisplayName()` getter as well?

chillu:
Yeah that example `FormActionAwesomenessCreator ` is great, could you weave that into the docs please?

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -204,26 +204,340 @@ To make the actions more user-friendly you can also use alternating buttons as
...
+
+### Using the injector within your component
+
+If your component has dependencies, you can add the injector via context using the `withInjector`
+higher order component.
+
+__my-module/js/components/Gallery.js__
+```js
+import React from 'react';
+import { withInjector } from 'lib/Injector';
```
chillu: Could you please also document how to inject dependencies into the component as props through `withInjector`?

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
unclecheese: Have added this in the new API. `update()` takes a third parameter you can use to name the HOC. If not defined, it falls back on the transformation name.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -204,26 +204,443 @@ To make the actions more user-friendly you can also use alternating buttons as
...
+the prop names and the service names will mirror each other.
+
+__my-module/js/components/Gallery.js__
+```js
+import React from 'react';
+import { inject } from 'lib/Injector';
+
+class Gallery extends React.Component {
+  render() {
+    const { SearchComponent, ItemComponent } = this.props;
```
sminnee: Is there a reason you have to call this ItemComponent and not simply GalleryItem? I get that it's injectable and can be replaced, but, for example, `EmailField::create()` also uses the default implementation as the name.

sminnee: The issue that I have is that we've created multiple arbitrary names for the same thing, which seems like it would add confusion.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -204,26 +204,443 @@ To make the actions more user-friendly you can also use alternating buttons as
...
 
-__my-component.js__
-```javascript
-import SilverStripeComponent from 'silverstripe-component';
+### Customising React components
 
-class MyComponent extends SilverStripeComponent {
+React components can be customised in a similar way to PHP classes, using a dependency
+injection API. The key difference is that components are not overriden the way backend
+services are. Rather, new components are composed using [higher order components](https://facebook.github.io/react/docs/higher-order-components.html).
```
sminnee: Maybe say "Higher order components have a similar role as Decorators in object-oriented code such as PHP"?

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -204,26 +204,443 @@ To make the actions more user-friendly you can also use alternating buttons as
 detailed in the [CMS Alternating Button](cms_alternating_button)
 how-to.
 
-## ReactJS in SilverStripe
+## React components
```
sminnee: I wonder if we should turn this into its own page? It's quite a lot of content.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -204,26 +204,443 @@ To make the actions more user-friendly you can also use alternating buttons as
...
+```
+
+Much like the configuration layer, we need to specify a name for this transformation. This
+will help other modules negotiate their priority over the injector in relation to yours.
+
+The second parameter of the `transform` argument is a callback which receives a `update()` function
+that allows you to mutate the DI container with a wrapper for the component. Remember, this function does not _replace_
+the component -- it enhances it with new functionality.
+
+The last thing we'll have to do is transpile our code and load the resulting bundle file
```
sminnee: We've got no instructions on how to do this transpiling. Is it feasible to provide a simple webpack boilerplate fro this?

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -204,26 +204,443 @@ To make the actions more user-friendly you can also use alternating buttons as
...
+We'll apply this one to the injector as well, but let's do it under a different name.
+For the purposes of demonstration, let's imagine this customisation comes from another
+module.
+
+__my-module/js/main.js__
+```js
+import Injector from 'lib/Injector';
+import TextLengthChecker from './components/TextLengthChecker';
+
+Injector.transform('my-other-transformation', (update) => {
```
sminnee: As best I can see you've deleted the CharacterCounter reference in this example. Does it need to be added back?

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -204,26 +204,443 @@ To make the actions more user-friendly you can also use alternating buttons as
...
+}
+
+export default CharacterCounter;
+```
+
+Now, when we apply this customisation, we need to be sure it loads _after_ the length
+checker in the middleware chain, as it relies on the prop `limit`. We can do that by specifying priority using `before` and `after`
+metadata to the customisation.
+
+For this example, we'll imagine these two enhancements come from different modules.
```
sminnee: I assume that you're using this to show how you can have multiple modules, as well as interacting enhancements.

I'd probably break this into separate sections, each demonstrating one of these things.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -204,26 +204,443 @@ To make the actions more user-friendly you can also use alternating buttons as
...
+  (update) => update('TextField', FinalTransform),
+  { after: '*' }
+);
+```
+**Note**: This flag can only be used once per transformation.
+The following are not allowed:
+* `{ before: ['*', 'something-else'] }`
+* `{ after: '*', before: 'something-else' }`
+
+### Helpful tip: Name your higher order components
```
sminnee: It might be worth including the name property in the examples above, but saying "we'll explain that 3rd argument below" or something, and then including the explanation here.

The issue being that people will likely just copy & paste the earlier examples as boilerplate.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -204,26 +204,443 @@ To make the actions more user-friendly you can also use alternating buttons as
...
+
+__my-module/js/main.js__
+```js
+import ConfirmingFormButton from './components/ConfirmingFormButton';
+
+Injector.transform('my-transformation', (update) => {
+  update('FormAction', ConfirmingFormButton, 'ConfirmingFormButton');
+});
+```
+### Registering new React components
```
sminnee: Maybe this could be put at the end of the Gallery example at the top? I kind of read that and wondered, "okay, so, now what? I've created a component but I can't use it anywhere."

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -204,26 +204,443 @@ To make the actions more user-friendly you can also use alternating buttons as
...
+        return <Component title={item.title} image={item.image} />
+      })}
+    </div>
+  }
+}
+
+export default withInjector(MyGallery);
+```
 
 ### Interfacing with legacy CMS JavaScript
```
sminnee: Is this next section still relevant? It's confusing and seems out of date.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
unclecheese: Yeah, I see that. We need to show the utility of the mapping function to assign the services to arbitrary variables. It's the equivalent of `properties: { someVar: %$SomeService } ` in YAML. 

If left undefined, the mapping function assigns the props identical to the service name. I do think that's an important thing to show off, because it's less magic. Plus, it's more IDE friendly to declare explicitly the props you're using.

I think what I can do is show the example without the mapping function first, where the prop names magically mirror the service names, and then show an enhancement where they're mapped, which would be the "we recommend this" method.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
unclecheese: You don't think the bit above this recommending webpack/babel is sufficient? I just don't want to get bogged down in duplicating documentation from third parties that will go stale in six months time.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
sminnee: >  We need to show the utility of the mapping function to assign the services to arbitrary variables

I think you need to shift this kind of stuff to an "advanced" page. For a newcomer it's a bit distracting.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
unclecheese: Likely not, but getting into that type of cleanup is out of scope for this story. I think it's fair to say that this section of the documentation needs a thorough review.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
unclecheese: Yeah, fair enough. That's a good change.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
sminnee: Is there something simple that can be a copy/paste starting point?

Right now I can follow the tutorial via copy/paste, go to Webpack University, learn how to apply its nontrivial set of config options to this specific case, write a webpack config file, and carry on with my day.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
chillu: It could use a `// ...`, yes

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
flamerohr: I think we can do this starting point later on after the webpack2 upgrade work has been done. As that contains a lot of the boilerplate work needed to hit the ground running for webpack in modules.

PR #6962, opened 2017-05-24
BUG Fix minor accessors of legacy ->class property

tractorcow:


tests/php/Forms/DatetimeFieldTest.php
```diff
@@ -23,11 +23,14 @@ protected function setUp()
...
         parent::setUp();
         i18n::set_locale('en_NZ');
+        // Fix now to prevent race conditions
+        DBDatetime::set_mock_now('2010-04-04');
         $this->timezone = date_default_timezone_get();
     }
 
     protected function tearDown()
     {
+        DBDatetime::clear_mock_now();
```
dhensby: I think this is fine in sapphiretest

Issue #6740, opened 2017-03-26
New UX insert link dialog - internal URLs


tractorcow:
## Acceptance criteria

 * Links to pages can be created through TinyMCE
 * Users can add link descriptions to all media types
 * When the cursor is on an existing internal page link, opening the "insert link" panel allows editing of the link
 * I can insert a link with a custom title set in the dialog when no selection is made
 * When editing an existing link, clearing the selection removes the link
 * I can explicitly remove a link from the dialog

## Tasks

* [x] Create Internal URL form in React
* [x] Create entwine script to inject in to tinymce
* [ ] Re-instate commented out behat tests

## Resources

 * Related ticket: #5947
 * Precondition: #6141 
 * Designs: https://projects.invisionapp.com/share/BGAYOTRFY#/screens/224873109

## Related

 * [Inline edit/remove of links in HTML editor](https://github.com/silverstripe/silverstripe-admin/issues/66)
 * [New UX insert link dialog - Insert anchors](https://github.com/silverstripe/silverstripe-admin/issues/65)

## Exclusions

- Create AnchorLink (dropdown list) component for the form, refer to `HTMLEditorField_Toolbar::getanchors()`
- Users do not have the ability to edit linked file descriptions from the insert link modal

## PRs:

 - [x] https://github.com/silverstripe/webpack-config/pull/2 (critical , please merge first and update NPM)
 - [x] https://github.com/silverstripe/silverstripe-framework/pull/6969
 - [x] https://github.com/silverstripe/silverstripe-cms/pull/1834
 - [x] https://github.com/silverstripe/silverstripe-admin/pull/88
 - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/491

tractorcow:
WIP is at https://github.com/open-sausages/silverstripe-admin/tree/pulls/4.0/tinymce-link-dialog

I've dropped priority of this story in lieu of some confirmation with @sminnee

clarkepaul:
I don't think this covers all of the dev from #5947 as it misses: link to file, link to email, and link to external url. @tractorcow is all of that work separated out into other issues?

tractorcow:
Yes @sminnee split the work into separate stories.

tractorcow:
See

-  https://github.com/silverstripe/silverstripe-framework/issues/6743
-  https://github.com/silverstripe/silverstripe-framework/issues/6740
-  https://github.com/silverstripe/silverstripe-framework/issues/6141

[...]

PR #6969, opened 2017-05-26
API Remove legacy HTMLEditor classes

tractorcow:
Required for https://github.com/silverstripe/silverstripe-framework/issues/6740

Ready for review, not ready for merge (need to update tests)


src/Forms/Schema/FormSchema.php
```diff
@@ -16,6 +17,8 @@
  */
 class FormSchema
 {
+    use Injectable;
```
tractorcow: I use FormSchema::singleton() outside of this class (asset-admin).

phalkunz: 👍

PR #6973, opened 2017-05-29
Fix for badly escaped values in CSV export

zanderwar:
The current functionality wouldn't escape characters correctly, have switched this over to `fputcsv` and result is much better inclusive of the fact that values will only be encapsulated with quotes IF required etc.

I couldn't discover why this was never used in the first place

Aside, I tried to refrain from using `ob_start()` and `ob_get_clean()` but `stream_get_contents($stream)` had unexpected results (the page would render unstyled instead of downloading the CSV)

### DIFF
--- a/forms/gridfield/GridFieldExportButton.php
+++ b/forms/gridfield/GridFieldExportButton.php
@@ -124,12 +124,12 @@ class GridFieldExportButton implements GridField_HTMLProvider, GridField_ActionP
 	 * Generate export fields for CSV.
 	 *
 	 * @param GridField $gridField
-	 * @return array
+	 * @return string
 	 */
 	public function generateExportFileData($gridField) {
-		$separator = $this->csvSeparator;
+		$separator = $this->getCsvSeparator();
 		$csvColumns = $this->getExportColumnsForGridField($gridField);
-		$fileData = '';
+		$fileData = array();
 
 		if($this->csvHasHeader) {
 			$headers = array();
@@ -145,8 +145,7 @@ class GridFieldExportButton implements GridField_HTMLProvider, GridField_ActionP
 				}
 			}
 
-			$fileData .= "\"" . implode("\"{$separator}\"", array_values($headers)) . "\"";
-			$fileData .= "\n";
+			$fileData[] = $headers;
 		}
 
 		//Remove GridFieldPaginator as we're going to export the entire list.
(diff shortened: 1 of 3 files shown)


zanderwar:
Will fix tests/issue shortly when possible, but open for discussion

forms/gridfield/GridFieldExportButton.php
```diff
@@ -115,12 +115,14 @@ protected function getExportColumnsForGridField(GridField $gridField) {
...
 	 * @param GridField $gridField
-	 * @return array
+	 * @return string
 	 */
 	public function generateExportFileData($gridField) {
 		$separator = $this->csvSeparator;
 		$csvColumns = $this->getExportColumnsForGridField($gridField);
-		$fileData = '';
+
+		$stream = fopen('php://memory', 'w');
```
tractorcow: My suggestion is to put this entire structure into a try {} finally {}, so you can safely nest output buffering and cleanup in finally.

If you are writing directly to a memory stream, you shouldn't need to use the output buffering feature, should you? As I understand it, ob_start(); affects the `php://output stream`.

tractorcow: Also see http://www.ozonesolutions.com/programming/2011/10/use-fputcsv-to-make-csv-output-to-browser-or-variable/

tractorcow: I think the issue you had is that headers_sent() was getting flagged to true, even though you weren't really sending any content from this block. Maybe double check your output when downloading locally?

If you are still having issues, maybe just try writing it to a temp file.

dhensby: can't use `try {} finally {}` as that's php 5.6 and 3.x has to support 5.3.3+

tractorcow: > can't use try {} finally {} as that's php 5.6 and 3.x has to support 5.3.3+

Oh darn, that's my bad. =( You will need to mock it with a try {} catch {} then.

tractorcow:
Feedback still stands... also needs a rebase @zanderwar if you're still up for it. :)

dhensby:
Right, I've decided to cherry-pick the commit from https://github.com/silverstripe/silverstripe-framework/pull/4432 because this does what we need but was applied to 4.0.

I've taken out the adding of new APIs so this is semver compliant

PR #6975, opened 2017-05-29
Fix $class variable from being clobbered

nfauchelle:
The $class variable gets overwritten in the function.

This causes error messages to be less helpful. For example if you setup a has_many but forget the has_one on the other side the error will look something like

`[Emergency] Uncaught Exception: No has_one found on class 'SomeObject', the has_many relation from 'SilverStripe\View\ViewableData' to 'SomeObject' requires a has_one on 'SomeObject'`

fixing this gives a more useful error, like

`[Emergency] Uncaught Exception: No has_one found on class 'SomeObject', the has_many relation from 'Page' to 'SomeObject' requires a has_one on 'SomeObject'`


tractorcow:
Good job @nfauchelle appreciate the fix.

PR #6981, opened 2017-05-31
Docs: Correct Stevie's name on committers page

edlinklater:
A bit over-zealous with the ending Ws.


micmania1:
@edlinklater tests? :p

PR #6982, opened 2017-05-31
Enforce type integer in DataLists byID and byIDs methods

ichaber:
Enforce type integer in DataLists byID and byIDs methods and throws exception if invalid type is given.
This is to prevent unexpected behaviour when filtering by ID without checking the type of params.
E.g. If using the get parameter **1;\"}};alert("ALERT");** unfiltered as parameter in byID it would return the entry with ID 1.

### DIFF
--- a/src/ORM/DataList.php
+++ b/src/ORM/DataList.php
@@ -948,15 +948,17 @@ class DataList extends ViewableData implements SS_List, Filterable, Sortable, Li
      */
     public function byIDs($ids)
     {
-        foreach ($ids AS $id) {
-            if (!is_int($id)) {
+        $intIDs = array();
+        foreach ($ids as $id) {
+            if (!is_numeric($id)) {
                 throw new InvalidArgumentException(
-                    'Invalid value passed to byIDs() in param array. All params have to be of type Integer.'
+                    'Invalid value passed to byIDs() in param array. All params have to be numeric or of type Integer.'
                 );
             }
+            $intIDs[] = intval($id, 10);
         }
 
-        return $this->filter('ID', $ids);
+        return $this->filter('ID', $intIDs);
     }
 
     /**
@@ -968,11 +970,12 @@ class DataList extends ViewableData implements SS_List, Filterable, Sortable, Li
      */
     public function byID($id)
     {
-        if (is_int($id)) {
-            return $this->filter('ID', (int)$id)->first();
-        } else {
-            throw new InvalidArgumentException('Incorrect param type for $id passed to byID(). Integer is expected.');
+        if (!is_numeric($id)) {
+            throw new InvalidArgumentException(
+                'Incorrect param type for $id passed to byID(). Numeric value is expected.'
+            );
         }
+        return $this->filter('ID', intval($id, 10))->first();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


tractorcow:
PHP uses type juggling, so we treat `(string)"1"` as a perfectly valid database ID identifier. In addition, various DB connectors return different casting for integer values retrieved from sql queries. Sometimes they are type-safe (int), sometimes they are pure strings.

We could just use `is_numeric` to get the same value without the strictness.

src/ORM/DataList.php
```diff
@@ -943,10 +943,19 @@ public function setQueriedColumns($queriedColumns)
      * Filter this list to only contain the given Primary IDs
      *
      * @param array $ids Array of integers
+     * @throws InvalidArgumentException
      * @return $this
      */
     public function byIDs($ids)
     {
+        foreach ($ids AS $id) {
+            if (!is_int($id)) {
```
tractorcow: please use `is_numeric` as these will fail on integer strings.

tractorcow: please use `is_numeric` as these will fail on integer strings.

src/ORM/DataList.php
```diff
@@ -943,10 +943,19 @@ public function setQueriedColumns($queriedColumns)
      * Filter this list to only contain the given Primary IDs
      *
      * @param array $ids Array of integers
+     * @throws InvalidArgumentException
      * @return $this
      */
     public function byIDs($ids)
     {
+        foreach ($ids AS $id) {
```
kinglozzer: Lowercase `as` please 😁

src/ORM/DataList.php
```diff
@@ -955,10 +964,15 @@ public function byIDs($ids)
...
+     * @throws InvalidArgumentException
      */
     public function byID($id)
     {
-        return $this->filter('ID', $id)->first();
+        if (is_int($id)) {
+            return $this->filter('ID', (int)$id)->first();
+        } else {
+            throw new InvalidArgumentException('Incorrect param type for $id passed to byID(). Integer is expected.');
+        }
```
kinglozzer: Could you please refactor this so that the `return` statement comes last? You can also remove the `else` statement:

```
if (somethingwrong) {
    throw new InvalidArgumentException('...
}

return $this->filter(...
```

tractorcow: Right, that way if you need another validation check, you can do it with pure addition and not having to rewrite the code. :D

src/ORM/DataList.php
ichaber: Sure, I would use is_numeric of the check and then parse all params to _int_. 
The only disadvantage to that is, that a string value like "0.1" is numeric and would parse to 0 using _intval_.
In addition to that, according to the docs (http://php.net/manual/en/function.intval.php) it behaves a bit differently for PHP7.1 with values like "1e5".

src/ORM/DataList.php
ichaber: Whoops, I used code autocomplete snippets :blush:. Will change that.

src/ORM/DataList.php
ichaber: :+1: I'll flip that around.

src/ORM/DataList.php
dhensby: You can check that the card value is the same as the original.

```php
If (is_numeric($val) && intval($val) == $val)
```

This should do strings getting through as ints and floats being cast to int.

tractorcow:
Thanks @ichaber good job.

PR #6983, opened 2017-05-31
Update to 00_CSV_Import.md

JustinTBrown:
Adding further explanation for using a custom CsvBulkLoader in ModelAdmin instead of the default one. I think some people might be able to guess at this, but others (like me) might benefit from making things a bit more explicit. This is a follow up from my [question on StackOverflow](https://stackoverflow.com/questions/44271755/adding-custom-csvbulkuploader-to-modeladmin-in-silverstripe).


dhensby:
thanks

PR #6984, opened 2017-06-01
NEW DB Driver defaults to PDO

dhensby:
We should default to the PDO driver out of the box as it has better performance.

This wasn't done in 3.x for backwards-compat reasons.

### DIFF
--- a/src/conf/ConfigureFromEnv.php
+++ b/src/conf/ConfigureFromEnv.php
@@ -8,7 +8,7 @@
  *
  * Your database connection will be set up using these defines:
  *  - SS_DATABASE_CLASS:    The database class to use, MySQLDatabase, MSSQLDatabase, etc. defaults to
- *                          MySQLDatabase
+ *                          MySQLPDODatabase
  *  - SS_DATABASE_SERVER:   The database server to use, defaulting to localhost
  *  - SS_DATABASE_USERNAME: The database username (mandatory)
  *  - SS_DATABASE_PASSWORD: The database password (mandatory)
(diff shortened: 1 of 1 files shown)


tractorcow:
Can we do a tiny bit of extension detection, maybe copy a bit of code from https://github.com/silverstripe/silverstripe-framework/blob/master/_register_database.php#L17

```php
$default = (class_exists('PDO') && in_array('mysql', PDO::getAvailableDrivers()))
    ? 'MySQLPDODatabase'
    : 'MySQLDatabase';
```

If we don't, then we should add ext-pdo to https://github.com/silverstripe/silverstripe-framework/blob/master/composer.json.

dhensby:
We don't currently check that mysqli extension is installed when we default to mysqli so I'm not sure we need to do the same for PDO. If someones not running with the recommended extensions they'll need to make manual adjustments.

A site can run on mysqli or PDO so neither should to be enforced in the composer.json - perhaps they could be recommendations?

tractorcow:
Ok, then don't make it a requirement, and use the environment checking as suggested?

dhensby:
What happens if neither PDO nor mysqli are available? At the moment we don't check the default is available, if someone doesn't have it they need to make an adjustment.

dhensby:
It also feels like we're tightly coupling our knowledge of what libraries the classes are using when this should only be information the classes themselves need to know about...

dhensby:
We'd also slow down sites for people relying on the default by calling `in_array` on every request

tractorcow:
> What happens if neither PDO nor mysqli are available?

Nice error message?

tractorcow:
> We'd also slow down sites for people relying on the default by calling in_array on every request

You can rewrite the code so it only runs if driver isn't selected.

dhensby:
I invoke the mantra "if it's an improvement it should be merged".

This PR changes nothing except that PDO is the preferred MySQL driver, which was my intend.

Adding checks that PDO/mysqli are installer and/or that PDO has the mysql driver available and/or that there's a nice error if no compatible lib is available are all nice additions but ultimately outside the scope of what I intend to achieve which is PDO drivers are our default option.

tractorcow:
It was an improvement so I merged it.

Issue #6632, opened 2017-02-17
Replace Database layer with doctrine/dbal


dhensby:
More and more we're seeing issues raised about our DB layer and it not working under specific circumstances. Our DB layer is also in need of a slight refactor to properly separate concerns and stop us hard coding logic related to DB drivers into our generic connectors.

Rather than spending our time re-factoring our DB connectors and maintaining an incomplete set of DB drivers I suggest we move our DB layer to a thirdparty - this will have the advantages of:

1. Reducing maintenance burden on core team - We currently struggle with the maintenance of external DB adapters especially MSSQL
2. Increase support of other DB drivers - The number of DB drivers we offer is limited and can be increased by moving to a thirdparty lib

I've looked at [Laravel's Illuminate](https://github.com/illuminate/database) and [Doctrine's DBAL](https://github.com/doctrine/dbal). Laravel's database component is far too tightly coupled to the Eloquent ORM - we'd have to pull it in even though we aren't going to use it.

Looking into doctrine/dbal and it's got a really nice API and a very similar architecture to our current DB layer. It looks like a pretty awesome and well architected DB layer as well as giving us the advantage of a lot of stock drivers and any other open-source drivers that have subsequently been written.

If we're able to swap out our DB layer, we'll drop a lot of hand-rolled code whilst improving functionality and focusing on adding value.

Other advantages:

1. More easily move to database migrations (https://github.com/doctrine/migrations/)
2. Defer DB connections to be "on demand" rather than connecting on every request

Issues:

1. There is no Enum support in doctrine/dbal and so we'll have to move this logic to the application layer.\*

\*Personally, I think this is something that we should be doing anyway as Enums can be painful for a number of reasons see [doctrine's docs on enums](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/cookbook/mysql-enums.html)

sminnee:
Dan, could you outline which classes you'd see as being replaced by dbal, and where the coupling points to dbal woud be?

dhensby:
I think this work is something that will need to be done in steps. In the long term I'd like to see pretty much everything in our `SilverStripe\ORM\Connect` namespace go. However, this is a lot of work and I'm starting by aiming at keeping most of our SQL generation and simply using doctrine/dbal to execute the queries for us.

[...]

PR #6985, opened 2017-06-01
NEW Don't hard code SQL identifier quotes

dhensby:
This is a first stab at removing hard-coded quotes from SQL identifiers in core.

This is a first step to decoupling from hand written SQL in core.


tractorcow:
I don't agree with this being merged at this stage.

At the moment, ansi sql-92 is our convention, and our SQL rewriting and extension behaviour relies on this.

If this were to be merged WITH the dbal rewrite in place, I could consider it, but as a stand alone change this immediately violates necessary conventions that are relied upon in other parts of the core.

For example, config vars (such as default_sort) can no longer have properly-escaped SQL identifiers, since it's now an expression not a constant.

It introduces extra fragility in DB connectors which may not use double quotes to escape identifiers; Some part of the code base which relied on double quotes will suddenly break. Having a single convention against which each part of the code can be tested against, in effect, means you don't need to test each part against every other part.

For example, the below test has one points of failure.


```php
$this->assertEquals('"Table"."ID"', $this->getIdentifier());
```

Whereas the below has two points of failure, and a possible false positive.

```php
$this->assertEquals(Convert::symbol2sql('Table.ID'), $this->getIdentifier());
```

If we absolutely have to abstract this, given the risks I've noted above, I would expect that at the very least we re-introduce SQLITE to the CI matrix to validate we properly test the variability of escaping syntax.

dhensby:
This PR doesn't change our convention of using ANSI - it stops us hard coding `"` in our queries. At the moment the quote identifier is hard coded to `"`.

Sort configs shouldn't be referencing quoted SQL, they should be referencing fields as defined in our model. Having this kind of cross over of knowledge of how the models are queried in our storage backend is nasty.

Quoting identifiers this way also means in the future it *can* change without a large and painful refactor because we will be following a convention that provides us future proofing and extensiblity.

Overall our tests and some core code rely far too much on SQL and asserting SQL is produced in very specific ways which is fragile and is something we should move away from, rather than double down on

tractorcow:
> Sort configs shouldn't be referencing quoted SQL, they should be referencing fields as defined in our model. Having this kind of cross over of knowledge of how the models are queried in our storage backend is nasty.

The problem is that you can't guess what it is; SQL sort could be an expression, or subquery, and those require the syntax to be in raw SQL, so it's treated as such to be consistent.

> Quoting identifiers this way also means in the future it can change without a large and painful refactor because we will be following a convention that provides us future proofing and extensiblity.

To allow this to change in the future requires a lot more than simply changing the quoting behaviour; This is why I'd want to see this as a part of the DBAL work in context, but not as a standalone component.

dhensby:
> The problem is that you can't guess what it is; SQL sort could be an expression, or subquery, and those require the syntax to be in raw SQL, so it's treated as such to be consistent.

I've not seen people sort by subquery in config. Sorting by subquery in config should not be possible imo, or any complex expression.

> To allow this to change in the future requires a lot more than simply changing the quoting behaviour; This is why I'd want to see this as a part of the DBAL work in context, but not as a standalone component.

This is a change that can be made standalone with no impact other than tightening up our SQL generation behaviour. Hard-coding delimiters is bad whether it's SQL or CSV.

Is there a solid argument as to why quoting identifiers shouldn't be programmatic?

tractorcow:
> I've not seen people sort by subquery in config. Sorting by subquery in config should not be possible imo, or any complex expression.

I have :D not in core, but in other modules. Some modules use `CASE` statements to do conditional sort based on null state of fields.

Ok, how about a compromise... we get the bulk of this PR in, but:

 - We don't rewrite private statics (incl default_sort), and leave them as SQL quoted for now
 - Add SQLITE back to the build matrix to improve regression testing against DB specific risks.
 - Fix Convert::sql2identifier() to double quote for now, and add a test with a comment for now, something ilke the below
 
```php
public function testConvertSQL2Identifier() {
	// Currenly fixed to ansi92 sql standard; Will change in the future
	// See https://github.com/silverstripe/silverstripe-framework/pull/6985/
	$this->assertEquals('"Table"."Column"', Convert::sql2identifier('Table.Column'));
}
```

We can look at fixing the privet static SQL blocks some other way. I might suggest, maybe some normalisation tool? Something that converts ansi-quotes to proper-quotes? That's the kind of thing I'd like to see come out of the final product.

src/Core/Convert.php
```diff
@@ -214,6 +214,24 @@ public static function symbol2sql($identifier, $separator = '.')
     }
 
     /**
+     * @param string $identifier
+     * @param string $separator
+     * @return string
+     */
+    public static function sql2symbol($identifier, $separator = '.')
+    {
+        $q = DB::get_conn()->getIdentifierQuoteCharacter();
```
Firesphere: The usage of `$q` is pretty unreadable. Would it make a difference if it's changed to `$quoteChar`?

src/ORM/Connect/Database.php
```diff
@@ -246,6 +246,16 @@ public function isActive()
     }
 
     /**
+     * The identifier quote char. For ANSI this is " and usually is ` for others
+     *
+     * @return string
+     */
+    public function getIdentifierQuoteCharacter()
+    {
+        return '"';
```
Firesphere: Shouldn't this be configurable?

src/ORM/Connect/Database.php
```diff
@@ -279,13 +289,18 @@ public function quoteString($value)
      */
     public function escapeIdentifier($value, $separator = '.')
     {
+        $q = $this->getIdentifierQuoteCharacter();
```
Firesphere: As above, I'm not a fan of undescriptive vars.

src/ORM/Connect/Database.php
tractorcow: As I said above, we need to fix this to `"` for now. We aren't able to yet soft-code this.

maxime-rainville:
Presumably, this has zero chance of happening in the 4 release line as it would introduce potentially breaking changes.

Shall we consider this for SS5 as part of https://github.com/silverstripe/silverstripe-framework/issues/6632?

maxime-rainville:
Doesn't look like there's much interest for this PR anymore. I'll close it for now, but please reopen if you feel it's still relevant.

PR #6986, opened 2017-06-02
Fix tinymce image selection issue in newer versions of Chrome

flamerohr:
Please see https://github.com/silverstripe/silverstripe-admin/issues/84 for more details


thirdparty/tinymce/tiny_mce_gzip.php
```diff
@@ -19,7 +19,7 @@
 	// Default settings
 	$tinyMCECompressor = new TinyMCE_Compressor(array(
 		// CUSTOM SilverStripe
-		'cache_dir' => TEMP_FOLDER
+		'cache_dir' => TEMP_FOLDER.'/tinymce-cache'
```
Firesphere: Could this maybe be a constant?

flamerohr: It could be, but all I wanted was to invalidate cache for this instance, so you don't need to wait up to 30 days for the fix to take effect.

kinglozzer: This directory needs to be created first, otherwise the cache is never written 😞

Firesphere: add spaces, while at it?

flamerohr: ooh, I hadn't realised it didn't get written :( I'll make the relevant changes soon

Issue #3239, opened 2014-06-27
DataList / ArrayList treat unfound entries with first/last differently


willmorgan:
For example, if you have an `ArrayList` and a `DataList` each with no items in them, if you call `$list->first()` on each, you'll get `false` or `null` respectively.

Should this not be standardised so it's null across all lists?

Really obscure one - found it while unit testing my work earlier.

phptek:
`ArrayList::first()` calls the native `reset()` function which can return a bool `false`, whereas `DataList` would simply return a `null`. So I'd argue it'd be the the other way around: They might both return a boolean. 

FYI looking at the `SS_List` interface, the phpDoc comments don't dictate a type to return (mixed). I guess as long as you're checking the return values using `!` or `empty()` you should be OK.

kinglozzer:
+1 for making `ArrayList` return `null`, updating the `List` interface to match is probably a good idea too

dhensby:
> +1 for making ArrayList return null, updating the List interface to match is probably a good idea too

Agreed - just because `reset()` returns a bool doesn't mean we should return a bool...

phalkunz:
Pull request: https://github.com/silverstripe/silverstripe-framework/pull/6987

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-framework/pull/6987

PR #6987, opened 2017-06-02
Consistent return values for first and last methods

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/3239

### DIFF
--- a/src/ORM/ArrayList.php
+++ b/src/ORM/ArrayList.php
@@ -302,6 +302,10 @@ class ArrayList extends ViewableData implements SS_List, Filterable, Sortable, L
      */
     public function first()
     {
+        if (empty($this->items)) {
+            return null;
+        }
+
         return reset($this->items);
     }
 
@@ -312,6 +316,10 @@ class ArrayList extends ViewableData implements SS_List, Filterable, Sortable, L
      */
     public function last()
     {
+        if (empty($this->items)) {
+            return null;
+        }
+
         return end($this->items);
     }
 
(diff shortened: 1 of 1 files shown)


flamerohr:
Since this is quite a core API change, could we add some unit tests to make sure we're covered for future changes in this?

src/ORM/ArrayList.php
```diff
@@ -302,7 +302,8 @@ public function shift()
      */
     public function first()
     {
-        return reset($this->items);
+        $first = reset($this->items);
+        return $first === false ? null : $first;
```
tractorcow: What if `$this->items` is `[false]`? :P

tractorcow: Maybe we should have a conditional on `empty($this->items)` instead?

src/ORM/ArrayList.php
```diff
@@ -516,7 +518,7 @@ public function canFilterBy($by)
     {
         $firstRecord = $this->first();
 
-        if ($firstRecord === false) {
+        if ($firstRecord === null) {
```
tractorcow: If the first record is false OR null then we can't filter.

tractorcow:
Note: `canFilterBy` should, but doesn't, use the same logic as `extractValue`. `extractValue` supports both array / dataobject, but `canFilterBy` only supports array.

src/ORM/ArrayList.php
phalkunz: Ah, true. 👍

src/ORM/ArrayList.php
```diff
@@ -516,7 +524,7 @@ public function canFilterBy($by)
     {
         $firstRecord = $this->first();
 
-        if ($firstRecord === false) {
+        if(empty($this->items)) {
```
tractorcow: Probably should do this before the `->first()` since it's wasted effort.

src/ORM/ArrayList.php
```diff
@@ -312,6 +316,10 @@ public function first()
      */
     public function last()
     {
+        if(empty($this->items)) {
```
tractorcow: Check linting.

src/ORM/ArrayList.php
tractorcow: or do `if (empty($firstRecord))`

PR #6988, opened 2017-06-02
Fixes printing from crashing

flamerohr:
because the css file has moved


tractorcow:
Do you need an extension in the silverstripe\admin module with the extension? The `silverstripe-admin/client/dist/styles/GridField_print.css` file does exist

tractorcow:
Sorry, just found https://github.com/silverstripe/silverstripe-admin/pull/96 :D

Issue #6387, opened 2016-12-12
Timeout / Login window stlying


clarkepaul:
## Acceptance criteria
* If logged out during a session due to a certain amount of time of inactivity the user should be presented with a quick way to log back in to continue what they were doing in the CMS in the current session
* Only the password is required (no email field needed) and must be used to gain access to CMS
* Changes are not lost due to being timed out
* If the password does not match the username/email then an appropriate validation and message are presented.
* If the correct password is entered then the user is presented with a success indication and automatically taken to the last place they were working on in the CMS so they can continue their work.
* There is no "close" trigger for the modal and the modal should not have the ability to use keyboard triggers to close the modal eg. escape key
* If the "Keep me logged in" checkbox is selected the timeout modal is never presented unless the user logs out completely (rather than just closing the tab/window). 
* We do not limit the number of login attempts as part of this issue.

## Design tasks
* [x] Define steps
* [x] Define text
* [x] Create designs
* [x] Team member peer review (provided by PK)
* [x] PO has had a visual pass

Mockups https://invis.io/EJBQUNR4Z#/234105662_Login_-_Session_Timeout

## Development tasks
* [x] Custom CMS template for CMSSecurity popup
* [x] Update CSS For modal popup
* [x] Better form-error validation messages (appear nicely as per design)
* [x] Fix login "loading" indicator (after submit)
* [X] Behat tests (nice to have, highly desired!)

## PRs

 - [x] https://github.com/silverstripe/silverstripe-admin/pull/97 (merge first)
 - [x] https://github.com/silverstripe/silverstripe-framework/pull/7026 (merge first)
 - [x] https://github.com/silverstripe/silverstripe-behat-extension/pull/161
 - [ ] https://github.com/silverstripe/silverstripe-framework/pull/6989 (restart behat tests)

## Notes:
In SilverStripe 3 we had a "Remember me" checkbox which never worked for me, often this pattern only saves the email address but sometimes auto fills the form. In 4.0 we have instead "Keep me logged in" which bypasses the login screen altogether so I have kept with this pattern as it seems to be more intuitive as to what the user is getting. 

We have no indication for how long the "Keep me logged in" action will work for but I presume in this UI it will continue to work until the user logs out rather than closing the tab/window. The alternative to this pattern is that the "Keep me logged in" only works for a set time period before the user needs to log in properly.

jaylou:
[...]

PR #6989, opened 2017-06-02
ENHANCEMENT Update style of CMSLogin form

tractorcow:
Closes https://github.com/silverstripe/silverstripe-framework/issues/6387

Requires https://github.com/silverstripe/silverstripe-admin/pull/97 first. After this is merged start behat tests.

Moved templates from admin module to framework, so that the templates could live in the same namespace as the PHP code.

Screenshots:

 
image

image

(success message was hard to take since it only appears for about 0.1 seconds)

image

### DIFF
--- a/src/Security/CMSSecurity.php
+++ b/src/Security/CMSSecurity.php
@@ -5,22 +5,21 @@ namespace SilverStripe\Security;
 use SilverStripe\Admin\AdminRootController;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\Director;
+use SilverStripe\Admin\LeftAndMain;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\Session;
 use SilverStripe\Core\Convert;
+use SilverStripe\Core\Manifest\ModuleLoader;
 use SilverStripe\ORM\FieldType\DBField;
+use SilverStripe\ORM\FieldType\DBHTMLText;
 use SilverStripe\View\Requirements;
+use SilverStripe\View\SSViewer;
 
 /**
  * Provides a security interface functionality within the cms
  */
 class CMSSecurity extends Security
 {
-
-    private static $casting = array(
-        'Title' => 'HTMLFragment'
-    );
-
     private static $allowed_actions = array(
         'login',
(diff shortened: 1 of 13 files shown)


tractorcow:
Including @Firesphere so that he gets alerted when this is merged. ;)

clarkepaul:
Is it guaranteed that once logged in the user is taken back to the CMS? If so I would prefer not having the "login successful" screen and just have the loading within the button. eg
image

Also mentioned here:
pasted_image_6_06_17__11_05_am

If we need to keep the "Successful login" screen there are inconsistencies with the text, I can provide you with the spinner/loading icon too if it isn't already part of the set.

tractorcow:
> Is it guaranteed that once logged in the user is taken back to the CMS? If so I would prefer not having the "login successful" screen and just have the loading within the button. eg

It's 99% guaranteed; The 1% could come from ad-blocker or some client-side weirdness.

tests/behat/src/CmsFormsContext.php
```diff
@@ -293,4 +293,31 @@ public function assertTheFieldHasProperty($name, $cond, $property)
                 throw new BadMethodCallException("Invalid condition");
         }
     }
+
+    /**
+     * @When /^I switch to the "([^"]*)" iframe$/
+     * @param string $id iframe id property
+     */
+    public function assertElementContainsText($id)
```
tractorcow: Yikes, that's a pretty bad oversight. =(

dhensby:
behat tests failing to pass after restart

tractorcow:
I'm attempting to debug the travis errors at https://github.com/open-sausages/silverstripe-framework/pull/19

src/Security/CMSSecurity.php
```diff
@@ -85,36 +91,26 @@ public function getResponseController($title)
...
+        );
     }
 
-    public function getTitle()
+    /**
+     * Check if there is a logged in member
+     *
+     * @return bool
+     */
+    public function getIsloggedIn()
```
flamerohr: I won't let this block merge, but is this used anywhere?

PR #6991, opened 2017-06-05
NEW Add 'updateURL' extension hook to File::getURL()

jacobbuck:


filesystem/File.php
```diff
@@ -719,7 +719,8 @@ public function getAbsoluteURL() {
 	 * @return string
 	 */
 	public function getURL() {
-		return Controller::join_links(Director::baseURL(), $this->getFilename());
+		$url = Controller::join_links(Director::baseURL(), $this->getFilename());
+		return $this->extend('updateURL', $url);
```
tractorcow: Be careful, ->extend returns an array of all return types. It should not return anything, but instead we return the modified (by ref) `$url`.

```php
$url = Controller::join_links(Director::baseURL(), $this->getFilename());
$this->extend('updateURL', $url);
return $url;
```

Additionally, since this is a new feature I need to switch this to the 3 (feature) branch.

tractorcow:
You can ignore the other commits that appeared. They should disappear once we merge them up into 3.

filesystem/File.php
jacobbuck: Fix that now.

jacobbuck:
Thanks @tractorcow, got that fixed up now 😄

PR #6998, opened 2017-06-06
Updated Form.php & 04_Form_Security.md  - strictFormMethodCheck to true

AntonyThorpe:
Changed the `strictFormMethodCheck` protected property from `false` to `true` to step out on the front foot with this security setting.  In the documentation under the title [Cross-Site Request Forgery](https://github.com/silverstripe/silverstripe-framework/blob/master/docs/en/02_Developer_Guides/09_Security/04_Secure_Coding.md#cross-site-request-forgery-csrf) it states, "it is also recommended to limit form submissions to the intended HTTP verb (mostly GET or POST) through [api:Form::setStrictFormMethodCheck()]."  The same advice is noted in [Form Security](https://github.com/silverstripe/silverstripe-framework/blob/c2292a4cc1649448b1eac8b651dba03466da8be5/docs/en/02_Developer_Guides/03_Forms/04_Form_Security.md#strict-form-submission).

Why not make this the default behaviour?  Is there a scenario where this would cause a problem?  Have manually tested in the CMS (alpha7) and is working fine.

Note: Original commit that establised the API Form::setStrictFormMethodCheck is 14c59be8.

Thank you.

### DIFF
--- a/docs/en/02_Developer_Guides/03_Forms/04_Form_Security.md
+++ b/docs/en/02_Developer_Guides/03_Forms/04_Form_Security.md
@@ -47,18 +47,18 @@ application up to `CSRF` security holes.
 
 ## Strict Form Submission
 
-Forms should be limited to the intended HTTP verb (mostly `GET` or `POST`) to further reduce attack exposure. Without 
+To reduce attack exposure forms are limited, by default, to the intended HTTP verb (mostly `GET` or `POST`). Without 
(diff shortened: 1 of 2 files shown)


dhensby:
I agree this should be on by default - I'm surprised it isn't.

chillu:
@AntonyThorpe Looks like you caused a regression with this: https://github.com/silverstripe/silverstripe-cms/issues/1846. Would you be able to look into it?

AntonyThorpe:
@chillu will do.

PR #7000, opened 2017-06-06
FIX: Upload_Validator failed to fetch max size from PHP ini values (fixes #6999)

### LINKED ISSUE #6999: UploadField fails to detect max file size from PHP values
Looks like a regression from https://github.com/silverstripe/silverstripe-framework/pull/4052. If no file size limits are set in config, it _should_ fall back to the minimum of either `upload_max_filesize` or `post_max_size`. However, we call `isset()` on the config value, which will return true on an empty array (the default value), so this never happens.

See https://github.com/silverstripe/silverstripe-framework/blob/ae8dbe309b797a20a91088351d70486cf3dfea79/filesystem/Upload.php#L377-L385

kinglozzer:


kinglozzer:
Also, 7000 get :tada:🍾

PR #7004, opened 2017-06-09
Fix change to latest changes for behat-extension

flamerohr:
It is missing the latest changes made and recently merged


dhensby:
The solution is to tag the behat module, I believe.

flamerohr:
hmm yes, that would help fix it :D

dhensby:
tagged

flamerohr:
thanks! I'll keep that in mind for bhehat in future

PR #7006, opened 2017-06-10
[FIX] Not `CascadeLogInTo` anymore, but `CascadeInTo`

Firesphere:
I'm mildly surprised this didn't break. I changed it to CascadeInTo, as the logout action needs to cascade into the session as well.


tractorcow:
This to be highlights the need for injector to be better at detecting vars. it shouldn't be silently ignoring dependencies that aren't properties.

tractorcow:
Please in future tag your commits with the correct BUG / FIX tag for bugs.

PR #7007, opened 2017-06-10
[FIX] CMSSecurity doesn't have Authenticators assigned.

Firesphere:
Switching to Security::singleton()->getApplicableAuthenticators() and Security::singleton()->getAuthenticators() fixes this and makes it more robust.

### DIFF
--- a/_config/security.yml
+++ b/_config/security.yml
@@ -31,5 +31,8 @@ SilverStripe\Core\Injector\Injector:
     properties:
       Authenticators:
         default: %$SilverStripe\Security\MemberAuthenticator\MemberAuthenticator
+  SilverStripe\Security\CMSSecurity:
+    properties:
+      Authenticators:
         cms: %$SilverStripe\Security\MemberAuthenticator\CMSMemberAuthenticator
   SilverStripe\Security\IdentityStore: %$SilverStripe\Security\AuthenticationHandler
--- a/src/Security/CMSSecurity.php
+++ b/src/Security/CMSSecurity.php
@@ -165,14 +165,14 @@ PHP
      *
      * @return bool
      */
-    public static function enabled()
+    public function enabled()
     {
         // Disable shortcut
         if (!static::config()->get('reauth_enabled')) {
             return false;
         }
 
-        return count(Security::singleton()->getApplicableAuthenticators(Authenticator::CMS_LOGIN)) > 0;
+        return count($this->getApplicableAuthenticators(Authenticator::CMS_LOGIN)) > 0;
     }
 
     /**
--- a/src/Security/Security.php
+++ b/src/Security/Security.php
@@ -356,7 +356,7 @@ class Security extends Controller implements TemplateGlobalProvider
                     _t('SilverStripe\\CMS\\Controllers\\ContentController.NOTLOGGEDIN', 'Not logged in')
                 );
                 // Tell the CMS to allow re-authentication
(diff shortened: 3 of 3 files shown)


src/Security/Security.php
```diff
@@ -269,7 +269,7 @@ public function index()
      */
     protected function getAuthenticator($name = 'default')
     {
-        $authenticators = $this->authenticators;
+        $authenticators = Security::singleton()->getAuthenticators();
```
tractorcow: getAuthenticators() is already an instance method on the service with the assigned dependencies. Don't re-invoke singleton() here. I can understand it on CMSSecurity though, but not in Security, please.

src/Security/Security.php
```diff
@@ -286,7 +286,7 @@ protected function getAuthenticator($name = 'default')
      */
     public function getApplicableAuthenticators($service = Authenticator::LOGIN)
     {
-        $authenticators = $this->authenticators;
+        $authenticators = Security::singleton()->getAuthenticators();
```
tractorcow: Same as above.

src/Security/Security.php
```diff
@@ -307,7 +307,7 @@ public function getApplicableAuthenticators($service = Authenticator::LOGIN)
      */
     public function hasAuthenticator($authenticator)
     {
-        $authenticators = $this->authenticators;
+        $authenticators = Security::singleton()->getAuthenticators();
```
tractorcow: Same as above.

tractorcow:
I changed my mind,

You need to fix this by correctly injecting CMSSecurity with it's own YML config. Doing it this way is going to be disasterous.

Firesphere:
Agreed, good point, didn't think of it that way, gimme a few :)

Firesphere:
Travis is green again, so, here ya go. This should be more robust, more logical with each Controller having it's own Authenticator(s) assigned where possible.

_config/security.yml
```diff
@@ -31,5 +31,8 @@ SilverStripe\Core\Injector\Injector:
     properties:
       Authenticators:
         default: %$SilverStripe\Security\MemberAuthenticator\MemberAuthenticator
+  SilverStripe\Security\CMSSecurity:
```
tractorcow: Do we want to go ahead and just have one-authenticator per Security class?

Firesphere: For the defaults, I think so. Because we don't want the standard `Security` to have the `CMS` login option, as it is unneeded in this case.

We _can_ have multiple per `Security`-class easily, I can confirm this works (though my tests are failing because it's actual two step now, instead of "try in a single step")

src/Security/CMSSecurity.php
```diff
@@ -165,14 +165,14 @@ protected function preLogin()
...
-    public static function enabled()
+    public function enabled()
     {
         // Disable shortcut
         if (!static::config()->get('reauth_enabled')) {
             return false;
         }
 
-        return count(Security::singleton()->getApplicableAuthenticators(Authenticator::CMS_LOGIN)) > 0;
+        return count($this->getApplicableAuthenticators(Authenticator::CMS_LOGIN)) > 0;
```
tractorcow: Might need to talk about this on Monday, but I think the CMS_LOGIN key is probably redundant. We probably over-engineered the authenticator assignment system in the initial implementation.

tractorcow:
Will merge for now since it fixes the bug. We can discuss API later. :)

PR #7011, opened 2017-06-12
Ensure a member exists before calling `isLockedOut`

jakedaleweb:
Keep seeing `PHP Fatal error: Call to a member function isLockedOut() on null` in logs, this should fix that.

### DIFF
--- a/security/MemberAuthenticator.php
+++ b/security/MemberAuthenticator.php
@@ -51,7 +51,9 @@ class MemberAuthenticator extends Authenticator {
 		if($asDefaultAdmin) {
 			// If logging is as default admin, ensure record is setup correctly
 			$member = Member::default_admin();
-			$success = !$member->isLockedOut() && Security::check_default_admin($email, $data['Password']);
+			if($member && !is_null($member)) {
+				$success = !$member->isLockedOut() && Security::check_default_admin($email, $data['Password']);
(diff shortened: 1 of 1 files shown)


security/MemberAuthenticator.php
```diff
@@ -51,7 +51,9 @@ protected static function authenticate_member($data, $form, &$success) {
 		if($asDefaultAdmin) {
 			// If logging is as default admin, ensure record is setup correctly
 			$member = Member::default_admin();
```
tractorcow: I think the fix is in the wrong place. Member::default_admin() should NOT be returning null if a default admin is specified. that method is meant to be guaranteed to return an instance. Can you please investigate why this is not happening?

security/MemberAuthenticator.php
```diff
@@ -51,7 +51,9 @@ protected static function authenticate_member($data, $form, &$success) {
...
-			$success = !$member->isLockedOut() && Security::check_default_admin($email, $data['Password']);
+			if($member && !is_null($member)) {
```
tractorcow: The right hand condition is redundant here, in any case.

dhensby:
I think this is just an order of comparison issue.

dhensby:
related #6944

PR #7012, opened 2017-06-12
FIX Order of conditionals for getting default admin

dhensby:
replaces #7011

### DIFF
--- a/security/MemberAuthenticator.php
+++ b/security/MemberAuthenticator.php
@@ -47,11 +47,11 @@ class MemberAuthenticator extends Authenticator {
 		}
 
 		// Check default login (see Security::setDefaultAdmin())
-		$asDefaultAdmin = $email === Security::default_admin_username();
+		$asDefaultAdmin = Security::has_default_admin() && $email === Security::default_admin_username();
 		if($asDefaultAdmin) {
 			// If logging is as default admin, ensure record is setup correctly
 			$member = Member::default_admin();
-			$success = !$member->isLockedOut() && Security::check_default_admin($email, $data['Password']);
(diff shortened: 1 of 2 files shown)


security/MemberAuthenticator.php
```diff
@@ -51,7 +51,7 @@ protected static function authenticate_member($data, $form, &$success) {
 		if($asDefaultAdmin) {
 			// If logging is as default admin, ensure record is setup correctly
 			$member = Member::default_admin();
-			$success = !$member->isLockedOut() && Security::check_default_admin($email, $data['Password']);
+			$success = Security::check_default_admin($email, $data['Password']) && !$member->isLockedOut();
```
kinglozzer: IMO the issue described in #7011 shows that we should code defensively and check that `Member::default_admin()` doesn’t return `null` here too:

```php
$success = Security::check_default_admin($email, $data['Password']);
if ($success && $member && !$member->isLockedOut() {
    return $member;
}
```

Thoughts?

dhensby: I don't think we should change the meeaning of the `$success` var as that will likely have cascading impacts.

I've changed it so `$asDefaultAdmin` won't be true if `Security::has_default_admin` is false. I've also added the `$member` truthy test to the `$success` assignment.

dhensby:
Ok, so to answer "how does this happen"

`$email` is initially set to `null` in the function, it's then only set if the `$data` array has an appropriate `Email` key. Perhaps a malformed or bad form submission is causing this problem? But these changes will meant that a `null` `$email` cannot match a `null` set of default admin credentials

PR #7016, opened 2017-06-12
Fix changetracker checkbox bugs

bcairns:
Checkboxes that are initially unchecked, do not work with the changetracker correctly (eg activating/deactivating "Save draft" and "Save & publish" buttons properly).

There are two bugs, both in the “setup original values” code (jquery.changetracker.js, line ~95):
1) The radio/checkbox jQuery selector does not use a quoted attribute for searching the name field, which breaks on CheckboxsetField’s bracketed[] names
2) The “has changed” logic for checkboxes treats all new values as 1 or 0, however the initialization ends up setting the original values for unchecked boxes to `undefined`, which will not match the 0 value when the box is unchecked again

This commit fixes both of these issues.


kinglozzer:
Closing in favour of https://github.com/silverstripe/silverstripe-framework/pull/7015, it will be merged up anyway :)

Issue #4149, opened 2015-05-01
RFC-2 Error page simplification


jonom:
# RFC-2 Move ErrorPage in to optional module

Author: @jonom
Status: Pending review
Version: 1.2.2 (2016-03-20)
## Purpose and outcome

This RFC aims to address a few issues with the current state of error page responses and error page content management in SilverStripe CMS:
1. **Flaws** Statically generated error pages are under-utilised and go stale quickly.
2. **Inflexibility** 'Error content' management is not a requirement for many websites, and a different approach may be preferred where it is.
3. **Confusion** The ErrorPage content management model leads to a confusing CMS experience.

Since the current `ErrorPage` implementation is suitable for some sites but not all, this proposal recommends removing the `ErrorPage` class from core and moving it in to a separate optional module. This would extend upon a simpler error response model in core, and pave the way for developers to create alternate solutions for managing, displaying, and caching error response content that suits their needs better.

Community is welcome to give feedback. The proposal will be submitted to the core-committers for approval.
### Static error pages easily become stale and broken

They are generated on dev/build but only if they don't already exist. After that they are only regenerated when re-published from the CMS. This means the static pages can easily become very out of date, and even include dev-environment content rendered when the website project was first initiated.
### Static error pages are not widely utilised

On a regular SilverStripe installation directives in the .htaccess file suggest that error responses will be handled efficiently by Apache serving up a static error page. This is true for a 500 response, but all 404 responses in fact go through the full SilverStripe request cycle and a dynamic error response is generated, meaning the static 404 file is never used. In the case of a fatal error Apache serves the 500 error page because PHP has aborted the response. Hopefully this situation doesn't happen often, but when it does it is likely that a stale error page is being served.
### Inflexible error response model
- The use of ErrorPage content [removes the ability to set a custom error message](https://www.silverstripe.org/community/forums/general-questions/show/15289) from code using `ContentController::httpError($code, $message)` (`$message` is ignored)
- Error responses which don't have a corresponding ErrorPage published in the CMS will show the content for a 500 Server Error which is misleading.
- Static error pages are only generated for the default language (can anyone confirm this?) and primary Subsite.
### ErrorPages confuse regular (non-technical) CMS users
- It doesn't make sense to show error pages in the Site Tree and allow drag and drop re-ordering, since they aren't part of a website's menu structure or information architecture.
- Non-technical CMS users are generally not interested in editing error page content
[...]

PR #7017, opened 2017-06-13
Removed ErrorPage

jonom:
Removed references to ErrorPage so it can be split off in to a separate module. Most of this happens in the CMS (https://github.com/silverstripe/silverstripe-cms/pull/1849)

Related RFC: https://github.com/silverstripe/silverstripe-framework/issues/4149

### DIFF
--- a/tests/php/Control/RequestHandlingTest.php
+++ b/tests/php/Control/RequestHandlingTest.php
@@ -3,7 +3,7 @@
 namespace SilverStripe\Control\Tests;
 
 use SilverStripe\Admin\LeftAndMain;
-use SilverStripe\CMS\Controllers\ErrorPageControllerExtension;
+use SilverStripe\ErrorPage\ErrorPageControllerExtension;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\Tests\RequestHandlingTest\AllowedController;
 use SilverStripe\Control\Tests\RequestHandlingTest\ControllerFormWithAllowedActions;
--- a/docs/en/01_Tutorials/01_Building_A_Basic_Site.md
+++ b/docs/en/01_Tutorials/01_Building_A_Basic_Site.md
@@ -71,7 +71,7 @@ should see the CMS interface with a list of the pages currently on your website 
 
 ### Try it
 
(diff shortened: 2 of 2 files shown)


tests/php/Control/RequestHandlingTest.php
```diff
@@ -25,16 +24,6 @@ class RequestHandlingTest extends FunctionalTest
     protected static $fixture_file = null;
 
     protected static $illegal_extensions = array(
-        // Suppress CMS error page handling
-        Controller::class => array(
-            ErrorPageControllerExtension::class,
-        ),
-        Form::class => array(
```
tractorcow: Instead of removing these, we should switch to the new namespace of the new module. Is that ok @jonom ?

jonom: Can do. I will hardcode a string rather than using `::class` as we'd want to avoid a tight coupling to the ErrorPage module, right?

tractorcow: ::class is ok. It works even if the module isn't installed.

jonom: 'kay that's done

tractorcow:
Don't merge until CMS + errorpage is done.

chillu:
Very cool, great effort and persistence, Jono! :) Does this *need* to be in beta1 (= is it an API change?). I can't see any documented in the upgrading guide here in this PR. If it's an API change, you've just got around a day to get this over the line - we're trying to stabilise master now for a beta1 release in the next few days.

jonom:
Aaargh, pushed some changes to my branch but they weren't getting picked up here so I closed and now can't reopen due to force push. **New PR at https://github.com/silverstripe/silverstripe-framework/pull/7061.**

@chillu yes it's an API change in CMS due to removal of ErrorPage class. I've added some notes to the changelog. This should be all good to go I think, module set up is done, just awaiting review and merge of [PR](https://github.com/silverstripe/silverstripe-errorpage/pull/2). Primary discussion and overview of work is at https://github.com/silverstripe/silverstripe-errorpage/issues/1

Issue #7021, opened 2017-06-14
[BUG] PNG images transparency after GDBackend::rotatePixelByPixel()


sanderha:
When manipulating a transparent PNG image with ``` GDBackend::rotatePixelByPixel() ```,  a white background is generated. 

By adding ```imagesavealpha($rotate, true); ``` after ```imagealphablending($rotate, false);``` (line 248), the issue is fixed.

chillu:
Can you send a pull request please?

sanderha:
I made a PR to 3.6 here https://github.com/silverstripe/silverstripe-framework/pull/7022

robbieaverill:
#7022 is merged into the 3.5 release line

PR #7022, opened 2017-06-14
keep PNG transparancy on GDBackend::rotatePixelByPixel

sanderha:
#7021

### DIFF
--- a/filesystem/GD.php
+++ b/filesystem/GD.php
@@ -321,6 +321,7 @@ class GDBackend extends Object implements Image_Backend {
 		}
 		$rotate=imagecreatetruecolor($destWidth,$destHeight);
 		imagealphablending($rotate, false);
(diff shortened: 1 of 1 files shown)


filesystem/GD.php
```diff
@@ -323,6 +323,7 @@ public function rotatePixelByPixel($angle) {
 		}
 		$rotate=imagecreatetruecolor($destWidth,$destHeight);
 		imagealphablending($rotate, false);
+        imagesavealpha($rotate, true); // to maintain PNG transparency
```
kinglozzer: I hate to be a pain in the butt... but could you change the indentation to tabs? SilverStripe 3 is still in the dark ages I’m afraid 😄

sanderha: Should be fixed now :)

dhensby:
I've rebased and squashed the commits so this should be good to go

PR #7028, opened 2017-06-15
Add CSRF token to logout action

kinglozzer:


kinglozzer:
Behat failures are related: I need to update `stepIAmNotLoggedIn()` to use `Security::logout_url()` so it gets a token

src/Security/Security.php
```diff
@@ -498,9 +498,35 @@ public function ping()
...
+        // Check pre-logout process
+        if ($response = $this->preLogout()) {
+            return $response;
+        }
+
         $member = static::getCurrentUser();
 
+        // If the user doesn't have a security token, show them a form where they can get one.
+        // This protects against nuisance CSRF attacks to log out users.
+        if ($member && !SecurityToken::inst()->checkRequest($this->getRequest())) {
```
Firesphere: This should be in the LogoutHandler, for consistency and handled below, where the authenticators are checked and validated for the required supportedService.

src/Security/Security.php
```diff
@@ -570,6 +596,32 @@ protected function preLogin()
     }
 
     /**
+     * Perform pre-logout checking and prepare a response if available prior to logout
+     *
+     * @return HTTPResponse Substitute response object if the logout process should be curcumvented.
+     * Returns null if should proceed as normal.
+     */
+    protected function preLogout()
```
Firesphere: I would expect the same as above, have this in the LogoutHandler?

src/Security/LogoutForm.php
```diff
@@ -0,0 +1,77 @@
...
+use SilverStripe\Forms\FormAction;
+use SilverStripe\Forms\HiddenField;
+use SilverStripe\Forms\Validator;
+
+/**
+ * Log out form to display to users who arrive at 'Security/logout' without a
+ * CSRF token. It's preferable to link to {@link Security::logout_url()}
+ * directly - we only use a form so that we can preserve the "BackURL" if set
+ */
+class LogoutForm extends Form
```
Firesphere: 👍 But this needs to be consistently implemented on CMS as well, right?

src/Security/Security.php
kinglozzer: I added this here so it would be consistent with `Security::login()` - that method calls `preLogin()` before delegating to the handlers. If I move the token checking to `LogoutHandler` (as per above comment), then this _could_ be removed... I think it’s still potentially useful though

Firesphere: I would like to have it moved to a central handler. And all other handling extensions should extend from that one handler. It's been on my radar to clean up and centralize it.

In short:
I do not believe pre/post methods should live in Security, but in the handler. Details will follow.

src/Security/LogoutForm.php
kinglozzer: I’m not sure how this affects the CMS? Login is only different for re-authenticating, that doesn’t really apply to logging out

kinglozzer:
Scope creep ftw. After a brief look and discussion with @Firesphere it became quite clear that this couldn’t be achieved without refactoring a few bits.

I’ve done a few things to `Security`:

- Updated `delegateToMultipleHandlers()` to be less opinionated on what gets passed through to the “wrapped” controllers. Moved the opinionated bits to a new `aggregateTabbedForms()` method.
- Abstracted out a few bits of `login()` so that they can be re-used for `logout()`
- Updated references to “login” (e.g. `setLoginMessage()`) to be more generically named, as they’re not just used for login

Not proposing this as the finished solution (I’ve got unit tests to do still), just pushing it for discussion

src/Security/LogoutForm.php
```diff
@@ -0,0 +1,77 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct(
+        RequestHandler $controller = null,
+        $name = self::DEFAULT_NAME,
```
dhensby: `self::DEFAULT_NAME` doesn't appear to be defined anywhere?

src/Security/LogoutForm.php
```diff
@@ -0,0 +1,77 @@
...
+        FieldList $fields = null,
+        FieldList $actions = null,
+        Validator $validator = null
+    ) {
+        $this->setController($controller);
+
+        $fields = $this->getFormFields();
```
dhensby: we shouldn't accept `$fields` as an argument if it's going to be discarded. Perhaps only use the classes fields if non were supplied?

src/Security/MemberAuthenticator/LogoutHandler.php
```diff
@@ -45,20 +49,64 @@ public function logout()
...
+
         if ($member instanceof Member) {
             Injector::inst()->get(IdentityStore::class)->logOut($this->getRequest());
         }
 
-        return true;
+        if (Security::getCurrentUser()) {
+            $this->extend('failedLogout');
+        } else {
+            $this->extend('afterLogin');
```
dhensby: `afterLogout` perhaps?

tests/php/Security/SecurityTest.php
```diff
@@ -277,6 +278,60 @@ public function testLoginUsernamePersists()
...
+        );
+
+        /* Make sure the user is still logged in */
+        $this->assertNotNull(Security::getCurrentUser(), 'User is still logged in.');
+
+        $token = $this->cssParser()->getBySelector('#LogoutForm_Form #LogoutForm_Form_SecurityID');
+        $actions = $this->cssParser()->getBySelector('#LogoutForm_Form input.action');
+
+        /* We have a security token, and an action to allow the user to log out */
+        $this->assertEquals(count($token), 1, 'There is a hidden field containing a security token.');
```
dhensby: `assertCount(1, $token)`

Argument order is ($expected, $actual)

src/Security/LogoutForm.php
kinglozzer: It’s inherited from `Form`

dhensby: derp

src/Security/LogoutForm.php
Firesphere: I think you'll get a strict warning if you don't add them?

tractorcow:
Given we have good pre-existing behat coverage of user authentication, I would like to see a small test to cover the new functionality that demonstrates the user interface works successfully. See https://github.com/silverstripe/silverstripe-framework/blob/master/tests/behat/features/login.feature

Something like:

 - I go to the logout url
 - I see the logout form
 - I click the "Logout" button
 - I see the message "You are now logged out"
 
 Or something to that effect. :)

src/Security/LogoutForm.php
kinglozzer: I’ve now wrapped this in an `if (!$fields)` conditional, which is what we do for `MemberLoginForm`

kinglozzer:
Added a behat scenario for this, depends on https://github.com/silverstripe/silverstripe-behat-extension/pull/164

dhensby:
I'm cancelling the behat tests for now (as they'll fail) and once the behat fix is merged and tagged we can restart the cancelled tests

dhensby:
Behat fix merged and tagged - behat tests restarted

dhensby:
@kinglozzer failure: https://travis-ci.org/silverstripe/silverstripe-framework/jobs/245314715#L647

tractorcow:
Looks good, @kinglozzer can you PR your change from https://github.com/kinglozzer/silverstripe-behat-extension/commit/987a029b51f747af1f9fd65b1d61c4b412e01d7c and we'll tag alpha8 of behat-extension.

kinglozzer:
Behat PR: https://github.com/silverstripe/silverstripe-behat-extension/pull/165. Bear in mind that merging & tagging it will break new builds until this PR is merged.

For posterity, the passing build is here: https://travis-ci.org/silverstripe/silverstripe-framework/builds/245399593

dhensby:
merged  and tagged behat

dhensby:
I've restarted the behat tests

dhensby:
we are green :)
Great job @kinglozzer

Firesphere:
Nice addition @kinglozzer. 👍

PR #7035, opened 2017-06-16
TreeDropDown performance boost.

mfendeksilverstripe:
TreeDropDown has very poor performance when dealing with thousands of pages. The main problem is that filtering and applying limits happen only after all pages are loaded which results in terrible performance.  This patch addresses both issues and filters and limits lists before further processing is done.

This patch aims to preserve original functionality. The only exception is that it changes the default children method from `AllChildrenIncludingDeleted` to `AllChildren` on `TreeDropDownFIeld` so we can use `DataList` instead of `ArrayList` in the processing.

### DIFF
--- a/forms/TreeDropdownField.php
+++ b/forms/TreeDropdownField.php
@@ -107,6 +107,13 @@ class TreeDropdownField extends FormField {
 			$this->childrenMethod = 'ChildFolders';
 			$this->numChildrenMethod = 'numChildFolders';
 		}
+		else {
+			// AllChildren should be used as a default because
+			// if you use AllChildrenIncludingDeleted instead, the filters will be much more demanding
+			// this is because instead of working with DataList you will be working with ArrayList
+			// AllChildren should be always used when dealing with more pages than the node leaf limit
+			$this->childrenMethod = 'AllChildren';
+		}
 
 		$this->addExtraClass('single');
 
@@ -303,16 +310,57 @@ class TreeDropdownField extends FormField {
 		if ($this->filterCallback || $this->search != "" )
 			$obj->setMarkingFilterFunction(array($this, "filterMarking"));
 
-		$obj->markPartialTree($nodeCountThreshold = 30, $context = null,
-			$this->childrenMethod, $this->numChildrenMethod);
+		// prepare a filter by ids, this is very useful when we have thousands of child pages under common parent
+		$filteredIds = ($this->search != '' && !is_null($this->searchIds) && is_array($this->searchIds))
+			? array_keys($this->searchIds) : array();
+
+		// collect IDs of currently selected pages (this includes hierarchy path)
+		$forceValue = $request->requestVar('forceValue');
+		if (!is_null($forceValue) || $this->value) {
+			$fieldValue = (!is_null($forceValue) ? $forceValue : $this->value);
+
+			if (($values = preg_split('/,\s*/', $fieldValue)) && count($values)) {
+				foreach ($values as $value) {
+					if (!$value || $value == 'unchanged') {
+						continue;
+					}
+
+					$selectedObject = $this->objectForKey($value);
+
+					// add all pages along the hierarchy path
+					$stack = $selectedObject->parentStack();
+					foreach ($stack as $stackItem) {
+						$filteredIds[] = $stackItem->ID;
+					}
+				}
+			}
+		}
+
+		// main node leaf limit configuration
+		$nodeCountThreshold = 30;
+		$obj->markPartialTree(
+			$nodeCountThreshold, $context = null,
+			$this->childrenMethod, $this->numChildrenMethod, $filteredIds
+		);
 
 		// allow to pass values to be selected within the ajax request
-		if( isset($_REQUEST['forceValue']) || $this->value ) {
-			$forceValue = ( isset($_REQUEST['forceValue']) ? $_REQUEST['forceValue'] : $this->value);
-			if(($values = preg_split('/,\s*/', $forceValue)) && count($values)) foreach($values as $value) {
-				if(!$value || $value == 'unchanged') continue;
+		if (!is_null($forceValue) || $this->value) {
+			$fieldValue = (!is_null($forceValue) ? $forceValue : $this->value);
 
-				$obj->markToExpose($this->objectForKey($value));
+			if (($values = preg_split('/,\s*/', $fieldValue)) && count($values)) {
+				foreach ($values as $value) {
+					if (!$value || $value == 'unchanged') {
+						continue;
+					}
+
+					$obj->markToExpose(
+						$this->objectForKey($value),
+						$this->childrenMethod,
+						$this->numChildrenMethod,
+						$nodeCountThreshold,
+						$filteredIds
+					);
+				}
 			}
 		}
 
@@ -337,7 +385,7 @@ class TreeDropdownField extends FormField {
 		// Skip the check if we're filtering the tree, since its not clear how many children will
 		// match the filter criteria until they're queried (and matched up with previously marked nodes).
 		$nodeThresholdLeaf = Config::inst()->get('Hierarchy', 'node_threshold_leaf');
-		if($nodeThresholdLeaf && !$this->filterCallback && !$this->search) {
+		if($nodeThresholdLeaf && !$this->filterCallback && !$this->search && count($filteredIds) == 0) {
 			$className = $this->sourceObject;
 			$nodeCountCallback = function($parent, $numChildren) use($className, $nodeThresholdLeaf) {
 				if($className == 'SiteTree' && $parent->ID && $numChildren > $nodeThresholdLeaf) {
@@ -361,7 +409,8 @@ class TreeDropdownField extends FormField {
 				$this->numChildrenMethod,
 				true, // root call
 				null,
-				$nodeCountCallback
+				$nodeCountCallback,
+                $filteredIds
 			);
 			return substr(trim($html), 4, -5);
 		} else {
@@ -374,7 +423,8 @@ class TreeDropdownField extends FormField {
 				$this->numChildrenMethod,
 				true, // root call
 				null,
-				$nodeCountCallback
+				$nodeCountCallback,
+                $filteredIds
 			);
 			return $html;
 		}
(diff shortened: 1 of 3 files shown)


model/Hierarchy.php
```diff
@@ -149,6 +149,17 @@ public function getChildrenAsUL($attributes = "", $titleEval = '"<li>" . $child-
 
 		if($this->owner->hasMethod($childrenMethod)) {
 			$children = $this->owner->$childrenMethod($extraArg);
+
+            // apply filter before any further processing
+			if (count($filteredIds) > 0) {
+                $children = $children->filter(['ID' => $filteredIds]);
```
dhensby: we support php 5.3 so short array syntax can't be used

model/Hierarchy.php
```diff
@@ -302,9 +323,15 @@ public function markingFilterMatches($node) {
...
 	public function markChildren($node, $context = null, $childrenMethod = "AllChildrenIncludingDeleted",
-			$numChildrenMethod = "numChildren") {
+			$numChildrenMethod = "numChildren", $filteredIds = array()) {
+
 		if($node->hasMethod($childrenMethod)) {
 			$children = $node->$childrenMethod($context);
+
+			// apply filter before any further processing
+			if (count($filteredIds) > 0) {
+                $children = $children->filter(['ID' => $filteredIds]);
```
dhensby: ditto short array syntax is not allowed

dhensby:
This change adds a new API so it'll need to go into the 3 branch.

dhensby:
>If there are too many pages the processing is simply skipped and user will see "Too many pages" message.

Does this mean the field is completely unusable if there are an arbitrary number of pages?

mfendeksilverstripe:
I'm testing the TreeDropdownField with ~4000 pages. Before, this change the performance was quite poor:

- it takes about 25 seconds to create a link in the WYSIWYG editor in the CMS
- it takes about 160MB of memory to process such request

It the best case this results in poor performance, in worst case it breaks the application as it hits the memory limit (this depends on your configuration).

After this patch the performance is as follows:

- it takes about 5 seconds to create a link in the WYSIWYG editor in the CMS
- it takes about 60MB of memory to process such request (which is roughly the same as other requests in the CMS) and the amount of memory no longer increases with addition of new pages

forms/TreeDropdownField.php
```diff
@@ -303,17 +310,57 @@ public function tree(SS_HTTPRequest $request) {
...
 			$obj->setMarkingFilterFunction(array($this, "filterMarking"));
 
-		$obj->markPartialTree($nodeCountThreshold = 30, $context = null,
-			$this->childrenMethod, $this->numChildrenMethod);
+		// prepare a filter by ids, this is very useful when we have thousands of child pages under common parent
+		$filteredIds = ($this->search != '' && !is_null($this->searchIds) && is_array($this->searchIds))
+            ? array_keys($this->searchIds) : array();
+
+        // collect IDs of currently selected pages (this includes hierarchy path)
+        if (isset($_REQUEST['forceValue']) || $this->value) {
```
Firesphere: If possible, avoid using global variables

forms/TreeDropdownField.php
```diff
@@ -303,17 +310,57 @@ public function tree(SS_HTTPRequest $request) {
...
+                    foreach ($stack as $stackItem) {
+                        $filteredIds[] = $stackItem->ID;
+                    }
+                }
+            }
+        }
+
+        // main node leaf limit configuration
+        $nodeCountThreshold = 30;
+		$obj->markPartialTree(
```
Firesphere: Please make sure indentation is consistent

forms/TreeDropdownField.php
```diff
@@ -303,17 +310,57 @@ public function tree(SS_HTTPRequest $request) {
...
 
 		// allow to pass values to be selected within the ajax request
-		if( isset($_REQUEST['forceValue']) || $this->value ) {
-			$forceValue = ( isset($_REQUEST['forceValue']) ? $_REQUEST['forceValue'] : $this->value);
-			if(($values = preg_split('/,\s*/', $forceValue)) && count($values)) foreach($values as $value) {
-				if(!$value || $value == 'unchanged') continue;
-
-				$obj->markToExpose($this->objectForKey($value));
-			}
+		if (isset($_REQUEST['forceValue']) || $this->value) {
```
Firesphere: Ditto, if possible, avoid globals

model/Hierarchy.php
```diff
@@ -375,11 +414,18 @@ public function markingClasses($numChildrenMethod="numChildren") {
...
 	 * @param int  $id   ID of parent node
 	 * @param bool $open If this is true, mark the parent node as opened
+     * @param string $childrenMethod The name of the instance method to call to get the object's list of children
+     * @param string $numChildrenMethod The name of the instance method to call to count the object's children
+     * @param int $nodeCountThreshold
+     * @param array $filteredIds list of ids that will be used to filter list items (reduce the result set before processing)
 	 * @return bool
 	 */
-	public function markById($id, $open = false) {
+	public function markById($id, $open = false, $childrenMethod = "AllChildrenIncludingDeleted",
```
Firesphere: Are you sure you want to default to `AllChildrenIncludingDeleted`?

mfendeksilverstripe: Yes, because I want to be consistent with the rest of the code in the `Hierarchy` class. Please note that $childrenMethod is correctly set in the `TreeDropDown` and propagated to the rest ofthe functionality, so it's not an issue.

I agree that this default should be changed to `AllChildren` in all instances of the code, but this change is beyond scope of this PR, thus I don't think it should be included.

Firesphere:
> Does this mean the field is completely unusable if there are an arbitrary number of pages?

No, it is the same as SiteTree currently does, when there are too many leafs to show, it'll show a message. A search for a specific page, or a part of it, will limit the amount of results and therefore will return a useful list.

model/Hierarchy.php
```diff
@@ -299,12 +330,20 @@ public function markingFilterMatches($node) {
 	 * @param mixed      $context
 	 * @param string     $childrenMethod    The name of the instance method to call to get the object's list of children
 	 * @param string     $numChildrenMethod The name of the instance method to call to count the object's children
+	 * @param int        $nodeCountThreshold
+	 * @param array      $filteredIds list of ids that will be used to filter list items (reduce the result set before processing)
 	 * @return DataList
 	 */
 	public function markChildren($node, $context = null, $childrenMethod = "AllChildrenIncludingDeleted",
-			$numChildrenMethod = "numChildren") {
+			$numChildrenMethod = "numChildren", $nodeCountThreshold = null, array $filteredIds = array()) {
```
Firesphere: Would it make sense to move the nodeCountThreshold to a config instead of passing it around? Would also save a bunch of null's being passed around

model/Hierarchy.php
Firesphere: Cool, thanks for the clarification :)

model/Hierarchy.php
mfendeksilverstripe: I think the main reason for this is that `Hierarchy` is considered a generic component that provides functionality to other components like `TreeDropDown`, `LeftAndMain` and so on. Each component has it's own config which is passed via params to the Hierarchy. This keeps the hierarchy code clean of references to other components.

dhensby:
thanks :)

dhensby:
@mfendeksilverstripe I've been looking at merging this work up into master (v4) but there's been quite a big re-write of that work and it *looks* like it'll be similar in terms of efficiency as to what you have.

The only issue is that the `$filteredIDs` is not passed into our `MarkedSet` object and so this information is not acted on...

It's too complex for me to merge this up myself, I'm afraid

hailwood:
@mfendeksilverstripe you're my hero buddy.
Working with a site-tree where one node has > 13,000 children and this performs perfectly.

--
Matt

mfendeksilverstripe:
@hailwood I'm glad you like it :)

mfendeksilverstripe:
@dhensby I would expect SilverStripe 4 to have this case already covered. Link editing in the CMS is a core functionality and it has to work properly even if there are thousands of pages.

The idea behind my changes is not that too complex - I only push filters and limitations before we actually change the `DataList` into `ArrayList`. Even if the implementation on SS4 is different the same idea should be used there as well.

dhensby:
@mfendeksilverstripe sure - but the effort to apply it against the diff was too great given my time constraints and context.

Would you have time to look into it in the 4 release?

mfendeksilverstripe:
@dhensby Sure, I'll have a look.

mfendeksilverstripe:
@dhensby I was able to integrate these changes to SS4. Please see following PRs:

https://github.com/silverstripe/silverstripe-framework/pull/7129
https://github.com/silverstripe/silverstripe-cms/pull/1882
https://github.com/silverstripe/silverstripe-lumberjack/pull/57

To get a functional site that has ~4000 pages, you need to have all three components.

dhensby:
thanks so much :)

jonom:
> This change adds a new API so it'll need to go into the 3 branch.

This got merged in to 3.5 and has since made it in to 3.6... was that an accident?

dhensby:
gah :/ yes - it should have gone into 3 and not 3.5 / 3.6 if they had been released :(

Sorry - we can revert it and merge into 3 instead - but it may be a little late for that now...

jonom:
I don't think this has made it in to a tagged release yet - it's waiting in the 3.5 and 3.6 branch. So from a SemVer perspective it should be possible to revert it without impacting anyone. Will that cause a big mess / be very difficult?

dhensby:
yep, you're right - this can be reverted and moved into `3`

Firesphere:
Well, it can indeed, but it will cause a mess for the people who use this change and actually profit from it?

Not disagreeing, but a way of applying the patch manually as a backup option would be a good idea maybe?

dhensby:
It's caused a regression as reported by @jonom - it's a new API and so shouldn't make a patch release. People who "rely" on dev branches for their code need to accept that comes with some risk. We can't wrap this in some "backup" option as that's just more semver violations.

If users need this fix they can get it from 3.x-dev or 3.7.x-dev instead of 3.5.x-dev or 3.6.x-dev. I think that's acceptable.

We should aim to investigate the regression a bit more as if it's causing breakages it'll need to be removed from 3.x-dev too :/

dhensby:
we are seeing issues being raised because of this change:

- https://github.com/silverstripe/silverstripe-subsites/issues/286
- https://github.com/silverstripe/silverstripe-framework/pull/7182

@mfendeksilverstripe do you have any desire to look into the cause of these problems as if there's a regression in the 3.x line when it comes to release 3.7 we'll have to revert this change...

PR #7036, opened 2017-06-18
Fix ImportButton not opening the modal

wilr:
The no-ajax class forces a window.location.href which prevents the Modal.


dhensby:
I assume this wants to target 4 now and not master - @tractorcow any thoughts on this PR?

tractorcow:
It fixes the issue, but what's up with the weird deprecation errors in CI?

PR #7037, opened 2017-06-19
[WIP] API App object implementation

tractorcow:
Don't merge yet, posted as request to get early peer review underway.

See https://github.com/silverstripe/silverstripe-framework/issues/6681

Other PRs:

 - [x] https://github.com/silverstripe/silverstripe-cms/pull/1854
 - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/505
 - [x] https://github.com/silverstripe/silverstripe-admin/pull/110
 - [x] https://github.com/silverstripe/silverstripe-assets/pull/25
 - [x] https://github.com/silverstripe/silverstripe-versioned/pull/16
 - [x] https://github.com/silverstripe/silverstripe-testsession/pull/49
 - [x] https://github.com/silverstripe/silverstripe-installer/pull/183
 - [x] https://github.com/silverstripe/silverstripe-reports/pull/74
 - [x] https://github.com/silverstripe/silverstripe-siteconfig/pull/63
 - [x] https://github.com/silverstripe/silverstripe-graphql/pull/104
 - [x] https://github.com/silverstripe/silverstripe-behat-extension/pull/163


cli-script.php
```diff
@@ -1,152 +1,19 @@
...
+use SilverStripe\Core\AppKernel;
+use SilverStripe\Core\HTTPApplication;
+use SilverStripe\Core\Startup\OutputMiddleware;
+use SilverStripe\Control\HTTPRequest;
+
+require __DIR__ . '/src/includes/cli.php';
+require __DIR__ . '/src/includes/autoload.php';
+
+// Build request and detect flush
+$request = HTTPRequest::createFromEnvironment();
```
sminnee: Rather than using cli.php fake the environment and then calling this method, would it make more sense to have a separate class/method that will create a request from command-line args?

cli-script.php
```diff
@@ -1,152 +1,19 @@
...
+
+// Default application
+$kernel = new AppKernel();
```
sminnee: Is it worth making BASE_PATH a constructor arg to AppKernel rather than having AppKernel directly reference the magic constant? I'm not sure we'd entirely get rid of BASE_PATH but it might be good to start moving away from its use in favour of kernel properties.

main.php
```diff
@@ -1,226 +1,19 @@
...
+
+require __DIR__ . '/src/includes/autoload.php';
+
+// Build request and detect flush
+$request = HTTPRequest::createFromEnvironment();
+
+// Default application
+$kernel = new AppKernel();
+$app = new HTTPApplication($kernel);
+$app->addMiddleware(new OutputMiddleware());
```
sminnee: I think Ingo's mentioned this but I would have expected $app->handle simply to return an HTTPResponse object... I don't have strong views one way or another but following the patterns of other projects is probably a good idea.

main.php
```diff
@@ -1,226 +1,19 @@
...
+$app->addMiddleware(new ErrorControlChainMiddleware($app, $request));
```
sminnee: Can we leave this one out by default?

main.php
```diff
@@ -1,226 +1,19 @@
...
+use SilverStripe\Core\Startup\OutputMiddleware;
+use SilverStripe\Control\HTTPRequest;
+
+require __DIR__ . '/src/includes/autoload.php';
+
+// Build request and detect flush
+$request = HTTPRequest::createFromEnvironment();
+
+// Default application
+$kernel = new AppKernel();
```
sminnee: BASE_PATH as arg? (also mentioned on cli-script.php)

main.php
sminnee: ...or does this provide the security for ?isDev and ?flush. If so, we should leave it in.

src/Core/AppKernel.php
```diff
@@ -0,0 +1,416 @@
...
+    }
+
+    /**
+     * Check or update any temporary environment specified in the session.
+     *
+     * @return null|string
+     */
+    protected function sessionEnvironment()
+    {
+        // Check isDev in querystring
```
sminnee: This feels like the wrong abstraction:

 * Direct reference to superglobals
 * Security of these values is in a whole separate subsystem; if we remove that it's a security hole

This feels like something that should be a routing middleware that relies on the regular HTTPRequest object and then maybe calls Kernel::setEnvironment()?

src/Core/HTTPApplication.php
```diff
@@ -0,0 +1,122 @@
...
+        return $this;
+    }
+
+    /**
+     * Call middleware
+     *
+     * @param callable $last Last config to call
+     * @return HTTPResponse
+     */
+    protected function callMiddleware($last)
```
sminnee: It's a little surprising that the request and/or application objects aren't available to the application middleware.

src/Control/HTTPRequest.php
```diff
@@ -138,13 +144,185 @@ public function __construct($httpMethod, $url, $getVars = array(), $postVars = a
...
         $this->httpMethod = strtoupper(self::detect_method($httpMethod, $postVars));
         $this->setUrl($url);
-
         $this->getVars = (array) $getVars;
         $this->postVars = (array) $postVars;
         $this->body = $body;
     }
 
     /**
+     * Create HTTPRequest instance from the current environment variables.
```
sminnee: This is feels like the "lots of static methods representing something that should be a separate class" anti-pattern. Maybe you want an `HTTPRequestBuilder` or `HTTPRequestFactory` here?

src/includes/functions.php
```diff
@@ -322,3 +206,114 @@ function get_increase_time_limit_max()
...
 }
+
+
+/**
+ * Returns the temporary folder path that silverstripe should use for its cache files.
+ *
+ * @param string $base The base path to use for determining the temporary path
+ * @return string Path to temp
+ */
+function getTempFolder($base = null)
```
sminnee: There 3 functions can be probably be pushed into a single class with 3 static methods, of which only the first one (this one) is public. `TempFolderCreator` or something (since it creates as well as finds the temp folder).

Benefits:
 * cleaner root namespace
 * smaller public API (only the first method)
 * can use autoloading

src/Core/HTTPApplication.php
tractorcow: Some middleware (e.g. ErrorControlChainMiddleware) take the app as a constructor args. I don't want to add a lot of API parameters to the core interface unless explicitly needed.

src/Core/AppKernel.php
tractorcow: This is existing code refactored from Director. I could just disable persistent isDev=1, if the loss of that functionality is worth the reduction in globals coupling.

main.php
tractorcow: I leave it out in cli-main.php :) It's essentially the entry point for errorcontrol chain, and making it a middleware makes it easy to skip.

main.php
tractorcow: My response to ingo is that this would weaken the ability of middleware to control response handling. E.g. what if a middleware needed to perform some action after ->output()?

src/Core/HTTPApplication.php
sminnee: Fair call. What about the request object?

src/Core/AppKernel.php
sminnee: Maybe use this and look at it in a separate PR, then? I think that the future is probably useful (if you need the post-response to run in dev mode, $_SESSION[isDev] is important), but it could be less hacky

src/includes/functions.php
tractorcow: That would be nice.

src/Control/HTTPRequest.php
tractorcow: Yes I think so.

cli-script.php
tractorcow: Yes I think so.

cli-script.php
tractorcow: I can roll it into the discussed HTTPRequestBuilder class.

main.php
chillu: Here's what I've written on Slack before. Sorry to bring this up a third time Damian, discussions can easily trail off on Slack (less persistent than on a pull request).

---

Current approach

```$kernel = new AppKernel($flush);
$app = new HTTPApplication($kernel);
$app->addMiddleware(new OutputMiddleware());
$app->addMiddleware(new ErrorControlChainMiddleware($app, $request));
$app->handle($request);
```

How is this better than the following approach which is modeled after Symfony? https://github.com/symfony/symfony-demo/blob/master/web/app.php. Output isn’t a middleware in my opinion, it’s a function to be invoked on the resulting response. Or not, in case you’re doing testing. As far as I can tell, you can still catch exceptions in the same way?

```$kernel = new AppKernel($flush);
$app = new HTTPApplication($kernel);
$app->addMiddleware(new ErrorControlChainMiddleware($app, $request));
$response = $app->handle($request);
$response->output();
```

I’ve seen your `defaultResponse` logic in https://github.com/open-sausages/silverstripe-framework/blob/af601ca82c57d6c4d1c68a2feda05e5abf1d4a02/src/Core/Startup/OutputMiddleware.php. Symfony has the exception handling and fallback responses in `handle()`: https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/HttpKernel/HttpKernel.php#L63

All in all, I think Symfony’s HTTPFoundation stuff is solid. It’s used in other frameworks like Laravel. For various reasons we can’t just use it directly. But I think we should stick to that model as much as we can. There’s been a lot of thought into the Symfony code. We shouldn’t try to invent something different like `OutputMiddleware` unless there’s a really good reason for it.

For example, in `HttpKernel->handle()` you can opt out of the built-in exception handling via a method arg, and override `finishRequest()` in your own `AppKernel`. These are all things we’ll eventually have to do as well, so we might as well take inspiration from a project that has gone down that road already

---

Damian responded on Slack:

> I think it fits as a middleware. Moving it out, again, means you can't have "after response" middleware around request outputting. You can still easily do exactly what you proposed without any API changes (just modify index.php) so I don't see this approach as a limitation.

What's an example of this? If you want pseudo-async jobs, `register_shutdown_handler()` after output is a separate concern. And in terms of a *middle*ware for HTTP request/response, they shouldn't be involved after the output has taken place. For all you know, the response has been sent (ignoring output buffering), you shouldn't mess with it any more. In general, I see all of this being customised in a project specific `AppKernel`, rather than relying on declarative stuff like `$defaultResponse` arguments in a `OutputMiddleware`

tractorcow: Well I won't change my opinion, but I will accept that the majority view is that it should not be a middleware. It's not critical to the success of app. I'll adjust this and remove the output middleware tomorrow.

src/Control/Director.php
```diff
@@ -1073,47 +992,7 @@ public static function is_ajax()
...
-     * can turn it back.
-     *
-     * Generally speaking, these methods will be called from your _config.php file.
-     *
-     * Once the environment type is set, it can be checked with {@link Director::isDev()},
-     * {@link Director::isTest()}, and {@link Director::isLive()}.
-     *
-     * @param string $environment
-     */
-    public static function set_environment_type($environment)
```
chillu: Note the removal of this API, and how to replace it

src/Control/HTTPRequest.php
chillu: Ditto, see https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/HttpFoundation/Request.php#L302

src/Control/Session.php
```diff
@@ -493,55 +516,11 @@ protected function recursivelyApply($data, &$dest)
...
-
-    /**
-    * Sets the appropriate form message in session, with type. This will be shown once,
-    * for the form specified.
-    *
-    * @param string $formname the form name you wish to use ( usually $form->FormName() )
-    * @param string $message the message you wish to add to it
-    * @param string $type the type of message
-    */
-    public static function setFormMessage($formname, $message, $type)
```
chillu: Document removal

src/Core/AppKernel.php
```diff
@@ -0,0 +1,416 @@
...
+use SilverStripe\Core\Manifest\ModuleLoader;
+use SilverStripe\Core\Manifest\ModuleManifest;
+use SilverStripe\Dev\DebugView;
+use SilverStripe\Dev\Install\DatabaseAdapterRegistry;
+use SilverStripe\Logging\ErrorHandler;
+use SilverStripe\ORM\DB;
+use SilverStripe\View\ThemeManifest;
+use SilverStripe\View\ThemeResourceLoader;
+
+class AppKernel extends CoreKernel
```
chillu: If we call the core class `AppKernel`, what would a subclass of this be called in a default installer project? In Symfony, `AppKernel extends Kernel`, and `AppKernel.php` lives in project code. I think the `App` naming prefix should stay reserved for project code. How do you see `CoreKernel` being used other than extended through `AppKernel`? I'm wondering if all code currently in `AppKernel` should simply move into `CoreKernel`.

src/Core/HTTPApplication.php
```diff
@@ -0,0 +1,125 @@
...
+        }
+        return call_user_func($next, $request);
+    }
+
+    /**
+     * @var Kernel
+     */
+    protected $kernel;
+
+    public function __construct(Kernel $kernel)
```
chillu: Weird function ordering, shouldn't var declarations come first, then `__dconstruct()`?

src/Dev/SapphireTest.php
```diff
@@ -2,48 +2,43 @@
 
 namespace SilverStripe\Dev;
 
+use Exception;
```
chillu: OMG awesome work in `SapphireTest`, so much simpler! 💯 🥇

src/includes/functions.php
chillu: Yeah scope creep :)

cli-script.php
tractorcow: Ok, removed cli.php, and created HTTPRequestBuilder / CLIRequestBuilder.

cli-script.php
tractorcow: Done.

main.php
tractorcow: Done.

main.php
tractorcow: Done.

src/Control/HTTPRequest.php
tractorcow: Moved to HTTPRequestBuilder and CLIRequestBuilder

src/Control/Director.php
```diff
@@ -120,112 +117,40 @@ class Director implements TemplateGlobalProvider
...
-                foreach ($v as $tmpFile) {
-                    if ($tmpFile && !is_uploaded_file($tmpFile)) {
-                        user_error("File upload '$k' doesn't appear to be a valid upload", E_USER_ERROR);
-                    }
-                }
-            } else {
-                if ($v['tmp_name'] && !is_uploaded_file($v['tmp_name'])) {
-                    user_error("File upload '$k' doesn't appear to be a valid upload", E_USER_ERROR);
-                }
+        if (getenv('SS_ALLOWED_HOSTS') && !static::is_cli()) {
```
tractorcow: @chillu I've decided to keep this logic exactly as it was in the same location.

chillu: Sweet, thanks :) We can make it more flexible later on, it's not going to be the only/last feature which is prone to be moved to a middleware

tractorcow:
In my most recent push I've consolidated both php-ini modification global methods, as well as direct global access into a new `Environment` class, and marked it as to be used sparingly.

The logic for constructing HTTPRequest objects has been moved into HTPRequestBuilder, and a CLIRequestBuilder supercedes the original `cli.php` bootstrap file.

src/Core/AppKernel.php
tractorcow: I'd be happy to merge the classes, given CoreKernel by itself is useless. @sminnee want to weigh in?

src/Core/HTTPApplication.php
tractorcow: We've added the request to the signature, following the PSR-15 middleware API. It's not a direct implementation, but follows the same method arguments / ordering. Once we implement PSR-7 it should be a simple shift to also support PSR-15.

src/includes/functions.php
tractorcow: Moved to TempFolder

src/Core/AppKernel.php
sminnee: Seems fine to me to merge the two classes.

dhensby:
I haven't fully reviewed this, but where is `setUpOnce` called and how come we're now going back to this pattern?

tractorcow:
@dhensby from setupBeforeClass.


```php
public static function setUpBeforeClass()
    {
        // Start tests
        static::start();

        // Call state helpers
        static::$state->setUpOnce(static::class);
        ...
```

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -245,6 +245,30 @@ Extensions
...
+    public function httpSubmission($data, $form, $request) {
+        Session::set('loggedIn', null);
+    }
+
+
+After:
+
+    :::php
+    public function httpSubmission($data, $form, $request) {
+        $requset->getSession()->set('loggedIn', null);
```
sminnee: typo in $request

src/Control/Director.php
sminnee: done here? https://github.com/silverstripe/silverstripe-framework/pull/7037/files#diff-e18c85252d7df22a2d833de561771477R1373

src/Core/HTTPMiddleware.php
```diff
@@ -0,0 +1,22 @@
...
+namespace SilverStripe\Core;
+
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
+
+/**
+ * HTTP Request middleware
+ * Based on https://github.com/php-fig/fig-standards/blob/master/proposed/http-middleware/middleware.md#21-psrhttpservermiddlewaremiddlewareinterface
+ */
+interface HTTPMiddleware
```
sminnee: I wonder if this should be part of SilverStripe\Control, since ti depends on HTTPRequest response?

src/Core/HTTPApplication.php
```diff
@@ -0,0 +1,125 @@
...
+namespace SilverStripe\Core;
+
+use SilverStripe\Control\Director;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
+
+/**
+ * Invokes the HTTP application within an ErrorControlChain
+ */
+class HTTPApplication implements Application
```
sminnee: I wonder if this should be part of SilverStripe\Control, since ti depends on HTTPRequest response?

sminnee: To be clear: I think that Kernel should definitely be part of core. I'm not sue about the Application interface; that might stay in core?

HTTPApplication isn't likely to be useful if you're using another controller system; in those cases you'd likely plug in the SilverStripe Kernel as a service another framework's app object.

tractorcow:
I'll move all http* classes to control @sminnee

src/Core/Config/ConfigLoader.php
```diff
@@ -78,14 +82,31 @@ public function countManifests()
...
     public function nest()
     {
-        $manifest = $this->getManifest()->nest();
-        $this->pushManifest($manifest);
-        return $manifest;
+        // Nest config
+        $manifest = clone $this->getManifest();
+
+        // Create new blank loader with new stack (top level nesting)
+        $newLoader = new self;
```
robbieaverill: In `activate()` we're using `static` - this won't consider late static binding - should it?

chillu:
Resolved conflicts locally, and merged with `git merge --squash` through 3873e4ba008cfc2af7e26ca86665affc289cd677. Now merging other requests and keeping an eye on builds. It's going to be a bit of a rocky ride in terms of build failures over the next ~16h. Happy to take a look in the NZ morning, unless @dhensby feels like checking build results as they come in.

PR #7042, opened 2017-06-20
Custom redux state in injector, updated docs

flamerohr:
Update docs for new Injector js structure and description for injector reducer API

Relies on https://github.com/silverstripe/silverstripe-admin/pull/114


docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -642,6 +642,146 @@ class MyGallery extends React.Component {
 export default withInjector(MyGallery);
 ```
 
+### Using injector to customise redux state data
+
+It is recommended to be familiar with the concepts of Immutability and Redux before starting this.
```
chillu: Link to good starting points?

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -642,6 +642,146 @@ class MyGallery extends React.Component {
...
+
+The examples use `Spread in object literals` which is at this moment in Stage 3 Proposal. If you're more comfortable with using
```
chillu: Link to description of this syntax? It's a bit hard to tell what you mean without a code example. Either point out that it's `{...}` syntax, or only mention this after first using the syntax in an example. At the moment, it'll feel to devs like they're missing something important that's hidden in the examples.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -642,6 +642,146 @@ class MyGallery extends React.Component {
...
+ the `Object.assign()` API that shouldn't present any problems and should work the same.
+
+To start customising, you'll need to transform an existing registered reducer.
```
chillu: How do you discover which reducer names are available?

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -642,6 +642,146 @@ class MyGallery extends React.Component {
...
+- `state` - current state of the current scope, this is what should be used to form the new State.
+- `type` - the action to fire, like in any reducer in redux, this helps determine if your transformer should do anything
+- `payload` - the new data sent with the action to mutate the redux store.
+
+```js
+const MyReducerTransformer = (originalReducer) => (globalState) => (state, { type, payload }) => {
+  switch (type) {
+    case 'MY_ACTION': {
+      // recommended to call and return the originalReducer with the payload changed by the transformer
+      /* return action to call here; */
```
chillu: It's hard to tell the difference between the first and second `case` here. Maybe rename to `CUSTOM_ACTION` and `EXISTING_ACTION`?

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -642,6 +642,146 @@ class MyGallery extends React.Component {
...
+      // it is important to return the originalReducer with original redux parameters.
+      return originalReducer(state, { type, payload });
+    }
+  }
+}
+```
+
+#### A basic transformation
+
+We manipulate the payload data to show something else instead of "Files" before calling the originalReducer
```
chillu: It's jumping into a concrete example (breadcrumbs and files) a bit too fast here - maybe add a sentence to prepare the reader?

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -642,6 +642,146 @@ class MyGallery extends React.Component {
...
+  
+  switch (type) {
+    /* ... cases here ... */
+  }
+}
+```
+
+#### Setting a different initial state
+
+We can easily define a new initial state by providing the `state` param with a default value,
```
chillu: Should be a full stop before "This does not work..."

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
```diff
@@ -642,6 +642,146 @@ class MyGallery extends React.Component {
...
+ this does not work if an earlier transformer in the chain has defined an initial state already.
+
+```js
+const MyReducerTransformer = (originalReducer) => () => (state = { myCustom: 'initial state here'}, { type, payload }) => {
```
chillu: That's a bit limiting, isn't it? It'll only set the `myCustom` key if the *entire* state tree has been undefined at this point. What happens when two transformers want to add initial state? Can't we just instruct the reader to set the initial state in the body?

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
```diff
@@ -0,0 +1,666 @@
...
+				]
+			}
+		]
+	}
+}
+```
+
+On its own, GraphQL offers nothing functional, as it's just a query language. You still need a service that will invoke queries and map their results to UI. For that, SilverStripe uses an implementation of [Apollo](http://dev.apollodata.com/) that works with React.
+
+## For more information
```
chillu: Awesome one-pager intro for these concepts, thanks Aaron! :)

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
```diff
@@ -0,0 +1,666 @@
...
+* [Build With React](http://buildwithreact.com/tutorial)
+* [Getting Started with Redux](https://egghead.io/courses/getting-started-with-redux)
+* [The React Apollo docs](http://dev.apollodata.com/react/)
+
+# The Injector API
+
+Much like customisation is done in the backend through [Injector](../../extending/injector), the client-side framework has its own implementation of dependency injection known as `Injector`. Using Injector, you can register new services, and transform existing services.
+
+Injector is broken up into three sub-APIs:
+* `Injector.react` for React UI components
```
chillu: It's a bit weird that we're naming one API after a library (`react`), and the other one after a concept (`reducer`). Why not `component` and `reducer`? You've probably discussed this already, just wondering.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
```diff
@@ -0,0 +1,666 @@
...
+This documentation will stop short of explaining React, Redux, and GraphQL/Apollo in-depth, as there is much better
+documentation available all over the web. We recommend:
+* [The Official React Tutorial](https://facebook.github.io/react/tutorial/tutorial.html)
+* [Build With React](http://buildwithreact.com/tutorial)
+* [Getting Started with Redux](https://egghead.io/courses/getting-started-with-redux)
+* [The React Apollo docs](http://dev.apollodata.com/react/)
+
+# The Injector API
+
+Much like customisation is done in the backend through [Injector](../../extending/injector), the client-side framework has its own implementation of dependency injection known as `Injector`. Using Injector, you can register new services, and transform existing services.
```
chillu: "backend" might be a bit confusing in this context, I'd say "Much like SilverStripe's Injector API in PHP"

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
```diff
@@ -0,0 +1,666 @@
...
+```
+
+Now, let's add some middleware to that service. The signature of middleware is: 
+```js
+(next) => (args) => next(args)
+```
+Where `next()` is the next customisation in the "chain" of middleware. Before invoking the next implementation, you can add whatever customisations you need. Here's how we would use middleware to enhance `LoggingService`.
+
+```js
+(LoggingService) => (error) => {
```
chillu: It's a bit weird to write invalid JS here. Which I guess you've done because `Injector.transform()` hasn't been introduced. How about moving this middleware section underneath "registering new services with Injector", and folding the middleware intro into "Transforming services using middleware"? Alternatively, you'll need to mention why this JS is invalid, and that we'll complete it later.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
```diff
@@ -0,0 +1,666 @@
...
+
+If you've created a module using React, it's a good idea to afford other developers an 
+API to enhance those components, forms, and state. To do that, simply register them with `Injector`.
+
+__my-public-module/js/main.js__
+```js
+import Injector from 'lib/Injector';
+
+Injector.react.register('MyComponent', MyComponent);
+Injector.reducer.register('myCustom', MyReducer);
```
chillu: This should also mention how the injector services can be used then, otherwise it'll be hard to understand why this is needed in the first place. Probably through the `MyComponent` example, since that's the easiest to understand for anybody with React context.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
```diff
@@ -0,0 +1,666 @@
...
+ open your React Developer Tools (see above) window and inspect the component name. The
+ context of the component is displayed between two square brackets, appended to the original name, for example:
+ `TextField[TextField.AssetAdmin.FileEditForm.Title]`. The context description is hierarchical, starting
+ with the most general category (in this case, "Admin") and working its way down to the most specific
+ category (Name = 'Title'). You can use Injector to hook into the level of specificity that you want.
+
+
+# Customising React components with Injector
+
+When middleware is used to customise a React component, it is known as a [higher order component](https://facebook.github.io/react/docs/higher-order-components.html).
```
chillu: You mention HoCs multiple times before this - put the link and explanation in earlier?

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
```diff
@@ -0,0 +1,666 @@
...
+
+If your component has dependencies, you can add them via the injector using the `inject()`
+higher order component. The function accepts the following arguments:
+
+```js
+inject(Component, [dependencies], mapDependenciesToProps)
+```
+* **Component** The component definition to inject into
+* **[dependencies]**: An array of dependencies (or a string, if just one)
+* **mapDependenciesToProps**: (optional) All depdencies are passed into this function as params. The function
```
chillu: Have you checked if ES6 has a way to get argument names, which would save us from repeating those? I assume that's not possible though

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
unclecheese: I totally agree. I don't like the idea of naming an API after a library. Will discuss with @flamerohr

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
flamerohr: good point, I've linked to https://facebook.github.io/immutable-js/ and http://redux.js.org/

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
flamerohr: I've added examples `newObject = { ...originalObject }` for object spread and `newObject = Object.assign({}, originalObject)` for the `Object.assign()` API

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
```diff
@@ -0,0 +1,666 @@
...
+	updater.react('TextField.AssetAdmin.FileEditForm', MyTextField);
+
+	// Applies to any textfield named "Title" in AssetAdmin
+	updater.react('TextField.AssetAdmin.*.Title', MyTextField);
+
+	// Applies to any textfield named "Title" in any admin
+	updater.react('TextField.*.*.Title', MyTextField);
+
+	// Applies to all text fields in AssetAdmin
+	updater.react('TextField.AssetAdmin', MyTextField);
```
chillu: Duplicate from first line in this function body

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
```diff
@@ -0,0 +1,666 @@
...
+
+### Updating the form schema
+
+Most behavioural and aesthetic customisations will happen via a mutation of the form schema. For this, we'll use the `updater.form.alterSchema()` function.
+
+```js
+Injector.transform(
+	'my-custom-form',
+	(updater) => {
+		update.form.alterSchema(
```
chillu: Syntax error, should be `updater`

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
```diff
@@ -0,0 +1,666 @@
...
+* `updateField(fieldName:string, updates:object)`
+* `updateFields({ myFieldName: updates:object })`
+* `mutateField(fieldName:string, callback:function)`
+* `setFieldComponent(fieldName:string, componentName:string)`
+* `setFieldClass(fieldName:string, cssClassName:string, active:boolean)
+* `addFieldClass(fieldName:string, cssClassName:string)`
+* `removeFieldClass(fieldName:string, cssClassName:string)`
+
+<div class="notice" markdown="1">
+It is critical that you end series of mutation calls with `getState()`. The form schema middleware expects an object, and will not handle an instance of `SchemaStateManager`.
```
chillu: > and will not handle an instance of `SchemaStateManager`

It's a bit confusing to mention a wrong return value by name here. In such a deeply nested callback and argument structure that confused me at least. I'd just leave out the remark

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
```diff
@@ -0,0 +1,666 @@
...
+</div>
+
+### Adding validation to a form
+
+Validation for React-rendered forms is handled by the `redux-form` package. You can inject your own middleware to add custom validation rules using the `updater.form.addValidation()` function.
```
chillu: Link to the lib here

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
```diff
@@ -0,0 +1,666 @@
...
+				)
+			)
+		)
+	}
+);
+```
+
+The `alterSchema()` function takes a callback, with an instance of `SchemaStateManager` (`form` in the above example), and a map of the current form values as parameters. `SchemaStateMangaer` allows you to declaratively update the form schema API using several helper methods, including:
+
+* `updateField(fieldName:string, updates:object)`
```
chillu: Mention which props are available from the redux-forms docs?

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
```diff
@@ -0,0 +1,666 @@
+title: React, Redux, and GraphQL
+summary: Learn how to extend and customise the technologies we use for application state and client-rendered UI.
+
+# Introduction to the "React" layer
```
chillu: Find a good place in this intro doc to link to the related how-tos?

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Customise_React_Components.md
```diff
@@ -0,0 +1,229 @@
+# Customising React Components
+
+In this tutorial, we'll customise some form elements rendered with React to have some new features.
+
+## An enhanced TextField
+
+Let's add a character count to the `TextField` component. Because the `TextField` component is fetched 
+through Injector, we can override it and augment it with our own functionality.
+
+First, let's create our higher order component.
```
chillu: Link back to the basics doc on what a HoC is

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Customise_React_Components.md
```diff
@@ -0,0 +1,229 @@
...
+            <TextField {...props} />
+            <small>Character count: {props.value.length}</small>
+        </div>
+    );
+}
+
+export default CharacterCounter;
+```
+
+Now let's add this higher order component to Injector.
```
chillu: Link to injector docs

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Customise_React_Components.md
```diff
@@ -0,0 +1,229 @@
+# Customising React Components
+
+In this tutorial, we'll customise some form elements rendered with React to have some new features.
+
+## An enhanced TextField
+
+Let's add a character count to the `TextField` component. Because the `TextField` component is fetched
```
chillu: Note that `TextField` is a built-in component in the admin area

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Customise_React_Components.md
```diff
@@ -0,0 +1,229 @@
...
+```js
+import Injector from 'lib/Injector';
+import CharacterCounter from './components/CharacterCounter';
+
+Injector.transform('my-transformation', (update) => {
+  update.react('TextField', CharacterCounter);
+});
+```
+
+The last thing we'll have to do is transpile our code and load the resulting bundle file
```
chillu: Link to existing "building with webpack" docs.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Customise_React_Components.md
```diff
@@ -0,0 +1,229 @@
...
+Now, when we apply this customisation, we need to be sure it loads _after_ the length
+checker in the middleware chain, as it relies on the prop `limit`.
+
+For this example, we'll imagine these two enhancements come from different modules.
+
+__module-a/js/main.js__
+```js
+import Injector from 'lib/Injector';
+import CharacterCounter from './components/CharacterCounter';
+Injector.transform(
```
chillu: Call the transformations something more meaningful and identifiable

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
flamerohr: Add a note about using `Injector.reducer.getAll()` to aid discovery.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Customise_React_Forms.md
```diff
@@ -0,0 +1,254 @@
+# Customising React Forms
```
chillu: It's a bit weird that you customise a form field in the "Customise React components" howto, including stuff like `handleClick` on buttons. Most readers would have expected that on "Customise React Forms"? Maybe add some cross-linking?

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Customise_React_Forms.md
```diff
@@ -0,0 +1,254 @@
...
+  (updater) => {
+    updater.react('ReduxFormField', HideableComponentCreator);
+  }
+);
+```
+
+Lastly, we need to apply a schema transformation using `updater.form.alterSchema()`.
+
+```js
+Injector.transform(
```
chillu: Looking at this code example, I think we need proper API docs on our JS (and/or type hinting) pretty soon. It's not clear at all any more what the arguments are here, and it's hard to trace that back to the actual API in this nested callback format. Just an observation, nothing we can fix with docs.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Customise_React_Forms.md
```diff
@@ -0,0 +1,254 @@
...
+      }
+    )
+  }
+);
+
+```
+
+## Conditionally adding a CSS class to a form field
+
+In this exmaple, we'll add the class "danger" to the `Price` field when `TicketsRemaining` is less than 10.
```
chillu: Spelling

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Extend_CMS_Interface.md
flamerohr: There are a few ways to do this, such as checking if `myCustom` is `undefined` then set it before any `switch (type)` statements.

hmm...
I caution setting initialState this way anyway, and would recommend this... I'll add it to the docs
```
if (typeof state === 'undefined') {
  return { ...orginalState(state, {type, payload}), myCustom: 'initial state here' };
}
```

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
flamerohr: to both of us `component` was too generic for use, whereas `reducer` is quite specific to `redux`

I think we spent a good night discussing the intricate naming, lol

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
unclecheese: The mapping function is optional. If it's not supplied, the prop names mirror the dependency names. Is that what you're talking about?

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
chillu: Fair enough, if you both agree on the naming, and have considered the tradeoffs, I'm good

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
chillu: Ah right, only required when your prop names are different. Then we're good.

docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/Customise_React_Forms.md
unclecheese: Yeah, this is a mess. Very confusing for a new dev. I've been chatting to @flamerohr about how we can simplify this. The fundamental issue is that these form customisations are not middleware in the pure sense. They're more like reducers. I'm going to do a bit of investigation to see how we can get that callback signature to look more like you'd expect:

```js
(formValues, schemaUpdater) => schemaUpdater.getState();
```
without all of the `updateSchema` bloat.

unclecheese: Update: I've started a branch that cleans this up. Do not intend to get it in for beta 1, but worth noting that this complexity is transitional.

PR #7045, opened 2017-06-21
BUG Add flag on form whether to notify user when there's unsaved changes

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-admin/issues/90

### DIFF
--- a/src/Forms/Form.php
+++ b/src/Forms/Form.php
@@ -263,6 +263,11 @@ class Form extends ViewableData implements HasRequestHandler
     protected $securityTokenAdded = false;
 
     /**
+     * @var bool
+     */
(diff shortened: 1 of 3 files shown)


dhensby:
This seems like very CMS specific functionality to have on all forms - is there not a nicer way to inject this just for CMS forms?

flamerohr:
I contemplated this for a while before merging...
We might have been able to achieve the same effect with extensions at different parts, I think it was something like:
1. add extension to provide the state functionality
2. add extension hooks to the Form schema generation to add the given data to the schema
3. have method checks in the FormFactories before calling the state functionality

It's quite a bit more overhead for a state flag, so I thought this would be a cleaner approach.

PR #7046, opened 2017-06-21
add inGroup(s) methods to Group

andrewandante:


src/Security/Group.php
```diff
@@ -377,6 +377,38 @@ public function collateAncestorIDs()
     }
 
     /**
+     * @param string|int|Group $group
+     * @return bool
+     */
+    public function inGroup($group) {
```
tractorcow: Would be good if the phpdoc told me what this method did. :)

Tests would also be nice.

src/Security/Group.php
```diff
@@ -377,6 +377,38 @@ public function collateAncestorIDs()
...
+        if (is_numeric($group)) {
+            $groupCheckObj = Group::get()->byID($group);
+        } elseif (is_string($group)) {
+            $SQL_group = Convert::raw2sql($group);
+            $groupCheckObj = Group::get_one("\"Code\" = '{$SQL_group}'");
+        } elseif ($group instanceof Group) {
+            $groupCheckObj = $group;
+        } else {
+            user_error('Member::inGroup(): Wrong format for $group parameter', E_USER_ERROR);
```
tractorcow: InvalidArgumentException is preferred response in this case.

src/Security/Group.php
```diff
@@ -377,6 +377,38 @@ public function collateAncestorIDs()
...
+        }
+
+        if(!$groupCheckObj) return false;
```
tractorcow: psr-2 will fail here. All conditionals need `{ }`

src/Security/Group.php
```diff
@@ -377,6 +377,38 @@ public function collateAncestorIDs()
...
     /**
+     * @param string|int|Group $group
+     * @return bool
+     */
+    public function inGroup($group) {
+        if (is_numeric($group)) {
+            $groupCheckObj = Group::get()->byID($group);
+        } elseif (is_string($group)) {
+            $SQL_group = Convert::raw2sql($group);
+            $groupCheckObj = Group::get_one("\"Code\" = '{$SQL_group}'");
```
tractorcow: Please use ORM filter API instead of raw SQL please.

src/Security/Group.php
```diff
@@ -377,6 +377,53 @@ public function collateAncestorIDs()
...
+     * @param string|int|Group $group Group instance, Group Code or ID
+     * @return bool Returns TRUE if the Group is a child of the given group, otherwise FALSE
+     */
+    public function inGroup($group)
+    {
+        if (is_numeric($group)) {
+            $groupCheckObj = Group::get()->byID($group);
+        } elseif (is_string($group)) {
+            $SQL_group = Convert::raw2sql($group);
+            $groupCheckObj = Group::get()->filter(['Code' => $SQL_group])->first();
```
tractorcow: Don't raw2sql when using filter api; It'll double encode.

```php
Group::get()->filter('Code', $group)->first();
```

is all you need

src/Security/Group.php
```diff
@@ -377,6 +377,53 @@ public function collateAncestorIDs()
     }
 
     /**
+     * Check if the group is a child of the given group or any parent groups
+     *
+     * @param string|int|Group $group Group instance, Group Code or ID
+     * @return bool Returns TRUE if the Group is a child of the given group, otherwise FALSE
+     */
+    public function inGroup($group)
```
tractorcow: Edge case, is `$group->inGroup($group)` ?

src/Security/Group.php
```diff
@@ -377,6 +377,53 @@ public function collateAncestorIDs()
...
+        }
+
+        return in_array($groupCheckObj->ID, $this->collateAncestorIDs());
+    }
+
+    /**
+     * Check if the group is a child of the given groups or any parent groups
+     *
+     * @param array $groups
+     * @param bool $strict set to TRUE if must be in ALL groups, or FALSE if must be in ANY
```
tractorcow: Not a well named arg; `$strict` means something other than what you're using it for. Did you mean to flag as conjunctive (and) or disjunctive (or)? Perhaps `$conjunctive = false` is better?

src/Security/Group.php
```diff
@@ -377,6 +377,53 @@ public function collateAncestorIDs()
...
+     * @return bool Returns TRUE if the Group is a child of any of the given groups, otherwise FALSE
+     */
+    public function inGroups($groups, $strict = false)
```
tractorcow: Edge case: is `$group->inGroups([]);` true?

src/Security/Group.php
```diff
@@ -377,6 +377,53 @@ public function collateAncestorIDs()
...
+    {
+        foreach ($groups as $group) {
+            if ($this->inGroup($group)) {
```
tractorcow: you will be calling $this->collateAncestorIDs() once for each loop. Why not inline inGroup() into this method and optimise it a little better? You could create a set of all groupIDs, and another set of collateAncestorIDs(), then do  subset comparison (array_intersect). You'd just need to count the intersection to determine if inGroups is true (intersect > 0 for disjunctive search, intersect === total for conjunctive search).

src/Security/Group.php
andrewandante: I think it should return true. Added a test.

src/Security/Group.php
andrewandante: Have reworked this, I _think_ I understood what you meant? Let me know if I've gone down the wrong rabbithole 🐰

src/Security/Group.php
```diff
@@ -377,6 +377,68 @@ public function collateAncestorIDs()
...
+    /**
+     * Turn a string|int|Group into a Group object
+     *
+     * @param string|int|Group $group Group instance, Group Code or ID
+     * @return Group|bool Returns FALSE if the Group doesn't exit, otherwise returns the Group Object
+     */
+    protected function identifierToGroup($group)
+    {
+        if (is_numeric($group)) {
+            $groupCheckObj = Group::get()->byID($group);
```
tractorcow: You are converting the id to a group, and then to an id again. That's a very inefficient SQL overhead.

src/Security/Group.php
```diff
@@ -377,6 +377,68 @@ public function collateAncestorIDs()
...
+        $candidateIDs = [];
+        foreach ($groups as $group) {
+            $candidate = $this->identifierToGroup($group);
+            if (!$candidate) {
+                continue;
+            }
+            $candidateIDs[] = $candidate->ID;
+        }
+        if ($requireAll) {
+            return array_intersect($candidateIDs, $ancestorIDs) === $ancestorIDs;
```
tractorcow: Sorting issues `[1, 2] !== [2, 1]`. You should only need to measure size of each group.

You should also be comparing to $candidateIDs not to $ancestorIDs, since $ancestorIDs may contain groups that aren't being compared to.

src/Security/Group.php
```diff
@@ -377,6 +377,68 @@ public function collateAncestorIDs()
...
+     *
+     * @param string|int|Group $group Group instance, Group Code or ID
+     * @return Group|bool Returns FALSE if the Group doesn't exit, otherwise returns the Group Object
+     */
+    protected function identifierToGroup($group)
+    {
+        if (is_numeric($group)) {
+            $groupCheckObj = Group::get()->byID($group);
+        } elseif (is_string($group)) {
+            $groupCheckObj = Group::get()->filter('Code', $group)->first();
```
tractorcow: If you require a group `somecode` that doesn't exist, your code will silently ignore it and pass.

tests/php/Security/GroupTest.php
```diff
@@ -171,6 +171,60 @@ public function testGetAllChildren()
...
+        $group->inGroup('NotARealGroup');
+    }
+
+    /**
+     * @expectedException \InvalidArgumentException
+     */
+    public function testGroupInGroupExceptionForFakeGroupID()
+    {
+        $group = new Group();
+        $group->inGroup(99999999);
```
dhensby: I'm not sure an invalid argument exception should be thrown when a correct type is passed

tractorcow: I think it should be an exception, but the wrong exception message is thrown. It says "invalid type" when the type is valid, but the ID isn't a real record.

andrewandante: I've updated the exception to cover all the bases - i.e. must be $correctType and exist

src/Security/Group.php
```diff
@@ -377,6 +377,79 @@ public function collateAncestorIDs()
...
+            $group = Group::get()->byID($groupID);
+        } elseif (is_string($groupID)) {
+            $group = Group::get()->filter(['Code' => $groupID])->first();
+        } elseif ($groupID instanceof Group && $groupID->exists()) {
+            $group = $groupID;
+        }
+        if ($group) {
+            return $group->ID;
+        }
+        if ($groupID instanceof DataObject) {
```
dhensby: This fails to catch objects that don't extend from `DataObject`

src/Security/Group.php
```diff
@@ -377,6 +377,70 @@ public function collateAncestorIDs()
...
+    {
+        if (empty($groups)) {
+            return false;
+        }
+        $ancestorIDs = $this->collateAncestorIDs();
+        $candidateIDs = [];
+        foreach ($groups as $group) {
+            try {
+                $candidateIDs[] = $this->identifierToGroupID($group);
+            } catch (\InvalidArgumentException $e) {
```
tractorcow: If you're catching and returning false, then it's not really an error. Exceptions add alot of overhead (processing time) and shouldn't be used unless it's a real error.

src/Security/Group.php
```diff
@@ -377,6 +377,70 @@ public function collateAncestorIDs()
...
+                if ($requireAll) {
+                    return false;
+                }
+            }
+        }
+        if ($requireAll) {
+            $matches = array_intersect($candidateIDs, $ancestorIDs);
+            return count($candidateIDs) === count($matches);
+        }
+        return count(array_intersect($candidateIDs, $ancestorIDs)) > 0;
```
tractorcow: same expression repeated above. Why not `count($matches) > 0`?

src/Security/Group.php
tractorcow: Going back to @dhensby 's feedback

> I think trying to have a custom error message for every type of invalid argument that's sent in is overkill. It'll be less logic and code just to use a generic message for the exceptions.
> Lets keep code complexity down

It's too hard to get this right just drop exceptions and handle the logic here. If it's $requireAll and null, return false, otherwise if it's null and not $requireAll, don't add it to $candidateIDs. you'll need to move your empty() check to below this loop though, if you could end up with empty $candidateIDs.

src/Security/Group.php
```diff
@@ -377,6 +377,64 @@ public function collateAncestorIDs()
...
+            }
+        }
+        if (empty($candidateIDs)) {
+            return false;
+        }
+        $matches = array_intersect($candidateIDs, $ancestorIDs);
+        if ($requireAll) {
+            return count($candidateIDs) === count($matches);
+        }
+        return count($matches) > 0;
```
dhensby: `!empty()`

tractorcow:
Thanks @dhensby and @andrewandante

PR #7050, opened 2017-06-22
DOCS InterventionBackend updates

dhensby:
see https://github.com/silverstripe/silverstripe-assets/pull/26


dhensby:
These tests pass locally... not entirely sure where the reflection reference to GDBackend_Manipulations is coming from the fact the test suite is pulling in `silverstripe/assets`

PR #7053, opened 2017-06-23
NEW restored Extension::__construct()

### LINKED ISSUE #6965: MINOR Missing Extension::__construct() breaks existing parent::__construct() in subclasses
To reduce breakage, and support for future constructor behaviour, @dhensby and I agree we should re-introduce the base constructor to Extension (recently removed).

fspringveldt:
Fixes #6965

### DIFF
--- a/src/Core/Extension.php
+++ b/src/Core/Extension.php
@@ -44,6 +44,10 @@ abstract class Extension
      */
     private $ownerStack = [];
 
+    public function __construct()
+    {
+    }
+
     /**
      * Called when this extension is added to a particular class
      *
--- a/src/ORM/Search/FulltextSearchable.php
+++ b/src/ORM/Search/FulltextSearchable.php
@@ -88,6 +88,7 @@ class FulltextSearchable extends DataExtension
      */
     public function __construct($searchFields = array())
     {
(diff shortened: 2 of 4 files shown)


tractorcow:
you'll need to add parent::__construct() to all the places it was removed, as well.

Also, we've removed the $class field, that shouldn't be re-introduced.

src/Core/Extension.php
```diff
@@ -44,6 +44,13 @@
      */
     private $ownerStack = [];
 
+    public $class;
+
+    public function __construct()
+    {
+        $this->class = get_class($this);
```
tractorcow: you'll need to add parent::__construct() to all the places it was removed, as well.

Also, we've removed the $class field, that shouldn't be re-introduced.

tractorcow:
One linting bug @fspringveldt 

```
FILE: ...build/silverstripe/silverstripe-framework/src/Core/Extension.php
 50 | ERROR | [x] Function closing brace must go on the next line
    |       |     following the body; found 1 blank lines before
    |       |     brace
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
```

PR #7054, opened 2017-06-23
Initially set a default height for the html editor to 350px

ssmarco:
BUG TinyMCE WYSIWYG loses height after saving page #1831

This merge request will set the initial tinymce editor height
See https://github.com/silverstripe/silverstripe-cms/issues/1831

### DIFF
--- a/src/Forms/HTMLEditor/TinyMCEConfig.php
+++ b/src/Forms/HTMLEditor/TinyMCEConfig.php
@@ -331,7 +331,8 @@ class TinyMCEConfig extends HTMLEditorConfig
     {
         return [
             'data-editor' => 'tinyMCE', // Register ss.editorWrappers.tinyMCE
-            'data-config' => Convert::array2json($this->getConfig())
+            'data-config' => Convert::array2json($this->getConfig()),
+            'style' => 'height:350px'
         ];
     }
 


src/Forms/HTMLEditor/TinyMCEConfig.php
```diff
@@ -320,7 +321,8 @@ public function getAttributes()
     {
         return [
             'data-editor' => 'tinyMCE', // Register ss.editorWrappers.tinyMCE
-            'data-config' => Convert::array2json($this->getConfig())
+            'data-config' => Convert::array2json($this->getConfig()),
+            'style' => 'height:350px'
```
flamerohr: don't need the `style` attribute if you're already setting `min_height` in the config setting already :D

flamerohr: Oh, or maybe remove `min_height` setting and keep the `style` :)
I don't believe both is needed, either way.

ssmarco: Hey, thanks for looking into this. The min_height works on first loading the page.
For some reason, when you save the document, TinyMCE re-renders and does not follow original height. Thus adding 'style' here works for me during that time.

ssmarco: I agree with this. Let me update my pull request. Thanks again.

flamerohr: Ah cool, have you tested with just `style`? it seems to work just as well with and without the `min_height`, I think
The main reason I'm looking at removing `min_height` is because I can resize the box to be smaller than 350px if I wanted to.

ssmarco: Yes, tested it and it works as expected.

flamerohr:
Rebased, this can be merged on green

flamerohr:
Thanks for the pull request @ssmarco :)

tractorcow:
I've tweaked this solution over at https://github.com/silverstripe/silverstripe-framework/pull/7228

PR #7055, opened 2017-06-23
NEW Ensure polymorphic has_one fields are indexed

### LINKED ISSUE #6966: PolymorphicForeignKeys are not made into DB indexes automatically
`DBForeignKey` get's an automatic index and so should our polymorphic ones.

robbieaverill:
I'm not sure whether this is the best way to tackle this, but at the point where you're looping database fields there's nothing to indicate that the relationship is polymorphic other than it is a ClassName enum. I've used that to look for a matching field name with the "ID" suffix and treat those as polymorphic.

Resolves #6966

### DIFF
--- a/src/ORM/DataObjectSchema.php
+++ b/src/ORM/DataObjectSchema.php
@@ -44,6 +44,13 @@ class DataObjectSchema
     protected $databaseIndexes = [];
 
     /**
+     * Fields that should be indexed, by class name
+     *
+     * @var array
+     */
+    protected $defaultDatabaseIndexes = [];
+
+    /**
      * Cache of composite database field
      *
      * @var array
@@ -65,6 +72,7 @@ class DataObjectSchema
         $this->tableNames = [];
         $this->databaseFields = [];
         $this->databaseIndexes = [];
+        $this->defaultDatabaseIndexes = [];
         $this->compositeFields = [];
     }
 
@@ -193,7 +201,6 @@ class DataObjectSchema
         foreach ($classes as $tableClass) {
             // Find all fields on this class
             $fields = $this->databaseFields($tableClass, false);
-
             // Merge with composite fields
             if (!$dbOnly) {
                 $compositeFields = $this->compositeFields($tableClass, false);
@@ -486,30 +493,59 @@ class DataObjectSchema
     }
 
     /**
-     * Cache all indexes for the given class.
-     * Will do nothing if already cached
+     * Cache all indexes for the given class. Will do nothing if already cached.
      *
      * @param $class
      */
     protected function cacheDatabaseIndexes($class)
     {
-        if (array_key_exists($class, $this->databaseIndexes)) {
-            return;
+        if (!array_key_exists($class, $this->databaseIndexes)) {
+            $this->databaseIndexes[$class] = array_merge(
+                $this->cacheDefaultDatabaseIndexes($class),
+                $this->buildCustomDatabaseIndexes($class)
+            );
         }
-        $indexes = [];
+    }
 
-        // look for indexable field types
-        foreach ($this->databaseFields($class, false) as $field => $type) {
-            /** @skipUpgrade */
-            if ($type === 'ForeignKey' || $type === 'DBClassName') {
-                $indexes[$field] = [
+    /**
+     * Get "default" database indexable field types
+     *
+     * @param  string $class
+     * @return array
+     */
+    protected function cacheDefaultDatabaseIndexes($class)
+    {
+        $indexes = [];
+        if (array_key_exists($class, $this->defaultDatabaseIndexes)) {
+            return $this->defaultDatabaseIndexes[$class];
+        }
+        $this->defaultDatabaseIndexes[$class] = [];
+
+        $fieldSpecs = $this->fieldSpecs($class, self::UNINHERITED);
+        foreach ($fieldSpecs as $field => $spec) {
+            /** @var DBField $fieldObj */
+            $fieldObj = Injector::inst()->create($spec, $field);
+            if ($fieldObj->getIndexed()) {
+                $this->defaultDatabaseIndexes[$class][$field] = [
                     'type' => 'index',
-                    'columns' => [$field],
+                    'columns' => $fieldObj->getIndexSpecs(),
                 ];
             }
         }
+        return $this->defaultDatabaseIndexes[$class];
+    }
 
-        // look for custom indexes declared on the class
+    /**
+     * Look for custom indexes declared on the class
+     *
+     * @param  string $class
+     * @return array
+     * @throws InvalidArgumentException If an index already exists on the class
+     * @throws InvalidArgumentException If a custom index format is not valid
+     */
+    protected function buildCustomDatabaseIndexes($class)
+    {
+        $indexes = [];
         $classIndexes = Config::inst()->get($class, 'indexes', Config::UNINHERITED) ?: [];
         foreach ($classIndexes as $indexName => $indexSpec) {
             if (array_key_exists($indexName, $indexes)) {
@@ -546,8 +582,7 @@ class DataObjectSchema
             }
             $indexes[$indexName] = $indexSpec;
         }
-
-        $this->databaseIndexes[$class] = $indexes;
+        return $indexes;
     }
 
     /**
--- a/src/ORM/DataObjectSchema.php
+++ b/src/ORM/DataObjectSchema.php
@@ -525,9 +525,9 @@ class DataObjectSchema
         foreach ($fieldSpecs as $field => $spec) {
             /** @var DBField $fieldObj */
             $fieldObj = Injector::inst()->create($spec, $field);
-            if ($fieldObj->getIndexed()) {
+            if ($type = $fieldObj->getIndexType()) {
                 $this->defaultDatabaseIndexes[$class][$field] = [
-                    'type' => 'index',
+                    'type' => $type,
                     'columns' => $fieldObj->getIndexSpecs(),
                 ];
             }
--- a/src/ORM/DataObjectSchema.php
+++ b/src/ORM/DataObjectSchema.php
@@ -525,11 +525,8 @@ class DataObjectSchema
         foreach ($fieldSpecs as $field => $spec) {
             /** @var DBField $fieldObj */
             $fieldObj = Injector::inst()->create($spec, $field);
-            if ($type = $fieldObj->getIndexType()) {
-                $this->defaultDatabaseIndexes[$class][$field] = [
-                    'type' => $type,
-                    'columns' => $fieldObj->getIndexSpecs(),
-                ];
+            if ($indexSpecs = $fieldObj->getIndexSpecs()) {
+                $this->defaultDatabaseIndexes[$class][$field] = $indexSpecs;
             }
         }
         return $this->defaultDatabaseIndexes[$class];
--- a/src/ORM/FieldType/DBClassName.php
+++ b/src/ORM/FieldType/DBClassName.php
@@ -37,6 +37,8 @@ class DBClassName extends DBEnum
      */
     protected static $classname_cache = array();
 
+    private static $indexed = true;
+
     /**
      * Clear all cached classname specs. It's necessary to clear all cached subclassed names
      * for any classes if a new class manifest is generated.
@@ -49,13 +51,14 @@ class DBClassName extends DBEnum
     /**
      * Create a new DBClassName field
      *
-     * @param string $name Name of field
+     * @param string      $name      Name of field
      * @param string|null $baseClass Optional base class to limit selections
+     * @param array       $options   Optional parameters for this DBField instance
      */
-    public function __construct($name = null, $baseClass = null)
+    public function __construct($name = null, $baseClass = null, $options = [])
     {
         $this->setBaseClass($baseClass);
-        parent::__construct($name);
+        parent::__construct($name, null, null, $options);
     }
 
     /**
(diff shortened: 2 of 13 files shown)


src/ORM/DataObjectSchema.php
```diff
@@ -486,29 +486,76 @@ protected function cacheDatabaseFields($class)
...
+     * Get "default" database indexable field types
+     *
+     * @param  string $class
+     * @return array
+     */
+    protected function defaultDatabaseIndexes($class)
+    {
+        $indexes = [];
+        $databaseFields = $this->databaseFields($class, false);
+        foreach ($databaseFields as $field => $type) {
```
tractorcow: Would it be more efficient to simply loop over all has_one? At least you would have exactly one index per loop. Then you only need to compare the right-hand side of the has_one to check if it's polymorphic or not. I'd abandon the old implementation and start afresh.

dhensby:
In my dbal work I did the following:

```php
                /** @var DBField $field */
                $fieldSpec = ClassInfo::parse_class_spec($fieldType);
                array_unshift($fieldSpec[1], $fieldName);
                $field = Injector::inst()->createWithArgs($fieldSpec[0], $fieldSpec[1]);
                $field->augmentDBTable($dbTable);
```

You could do this and then perform an `instanceof` or, better yet, let the field type itself determine if there's an index to add... `$field->augmentDBIndex($indexes);`

tractorcow:
I was going to say parse_class_specs might be a bit inefficient for has_ones, but I guess this is only done on schema update right (dev/build).

@dhensby what was your outer loop for that code?

robbieaverill:
@tractorcow @dhensby I've pushed a different approach which hopefully is closer to what you've both suggested.

Couple of notes:

* I've added some logic to track polymorphic indexes in the `cacheDatabaseFields` method, since that is the point where PolymorphicForeignKey instances are converted to Int and ClassName and are difficult to detect afterwards
* Have made "indexed" configurable
* Still need to loop `databaseFields()` as we need to add indexes on some fixed fields like ClassName which wouldn't be picked up if we simply looped the has_ones

dhensby:
I'm a little unclear on this because I'd expect that we use the `composite_db` config on composite fields to build the index but I don't see that happening anywhere?

tractorcow:
Well, not every composite field is automatically an index in the same way that has_ones are.

src/ORM/DataObjectSchema.php
```diff
@@ -486,29 +500,65 @@ protected function cacheDatabaseFields($class)
...
         }
+    }
+
+    /**
+     * Get "default" database indexable field types
+     *
+     * @param  string $class
+     * @return array
+     */
+    protected function defaultDatabaseIndexes($class)
```
tractorcow: Should be called `cacheDefaultDatabaseIndexes` to distinguish this from the public getter and keep its purpose clear. This should also be the only place that defaultDatabaseIndexes is populated, since it seems part of this method has migrated to cacheDatabaseFields.

src/ORM/DataObjectSchema.php
```diff
@@ -486,29 +500,65 @@ protected function cacheDatabaseFields($class)
...
+
+    /**
+     * Look for custom indexes declared on the class
+     *
+     * @param  string $class
+     * @return array
+     * @throws InvalidArgumentException If an index already exists on the class
+     * @throws InvalidArgumentException If a custom index format is not valid
+     */
+    protected function customDatabaseIndexes($class)
```
tractorcow: As with `buildTableName` being called from `cacheTableNames`, this should be called `buildCustomDatabaseIndexes` as it's called from `cacheDatabaseIndexes`, so that it doesn't look like the standard getter.

src/ORM/DataObjectSchema.php
```diff
@@ -471,7 +479,13 @@ protected function cacheDatabaseFields($class)
             $fieldObj->setTable($class);
             $nestedFields = $fieldObj->compositeDatabaseFields();
             foreach ($nestedFields as $nestedName => $nestedSpec) {
-                $dbFields["{$fieldName}{$nestedName}"] = $nestedSpec;
+                $compositeName = $fieldName . $nestedName;
+                $dbFields[$compositeName] = $nestedSpec;
+
+                // Handle configurable indexes, e.g. for polymorphic keys
```
tractorcow: This logic should probably be in defaultDatabaseIndexes shouldn't it?

src/ORM/DataObjectSchema.php
```diff
@@ -486,29 +500,65 @@ protected function cacheDatabaseFields($class)
...
 
-        // look for indexable field types
-        foreach ($this->databaseFields($class, false) as $field => $type) {
-            if ($type === 'ForeignKey' || $type === 'DBClassName') {
-                $indexes[$field] = [
-                    'type' => 'index',
-                    'columns' => [$field],
-                ];
+        // Ensure that fixed field types, e.g. ClassName can be indexed
+        $databaseFields = $this->databaseFields($class, false);
```
tractorcow: If you use `$this->fieldSpecs($class, UNINHERITED)` here you would get all DBField instances, including DBPolymorphicForeignKey, DBForeignKey, and other instances.

All you need is the class of the field, and the `indexed` config option for that class right? then all you need is a method that gets the list of fields to index for a given DBField instance.

src/ORM/DataObjectSchema.php
```diff
@@ -486,29 +500,65 @@ protected function cacheDatabaseFields($class)
...
+            /** @var DBField $fieldObj */
+            $fieldObj = Injector::inst()->create($type, $name);
+            if ($fieldObj->config()->get('indexed')) {
+                $this->defaultDatabaseIndexes[$class][] = $name;
             }
         }
 
-        // look for custom indexes declared on the class
+        // Return the registered default indexes
+        foreach ($this->defaultDatabaseIndexes[$class] as $field) {
```
tractorcow: What if $field is a composite field? `colums` should be as many fields as that composite field has right?

tractorcow:
This needs some tests with polymorphic has_ones, normal composite fields, and maybe DBClassName for good measure, to make sure the appropriate indexes are built. :) I've left all feedback I could!

Good job so far.

src/ORM/DataObjectSchema.php
robbieaverill: Yes ideally, but at that point you don't know whether it's a polymorphic key because at this point it gets changed into an Int and a DBClassName instead

robbieaverill:
Thanks, will add some more tests when I get a chance

src/ORM/DataObjectSchema.php
tractorcow: My suggestion for using fieldSpecs should pull those out in the result. Although, you'll get both the composite field parent, and all sub-fields in the result. Not an issue unless a composite field has 'indexed' and one of it's components ALSO has 'indexed' though right?

dhensby:
> Well, not every composite field is automatically an index in the same way that has_ones are.

Why is this? If it's not then why are we building this into core :/ - I'm probably missing something fairly obvious here, sorry.

I'd have thought that the composite field ID/ClassName should always be indexed

tractorcow:
Yes, but composite Money Field (currency, amount) shouldn't automatically be indexed.

src/ORM/DataObjectSchema.php
robbieaverill: Actually, I'm not sure we can do this just with config - `DBPolymorphicForeignKey` is not a FQCN. We could define unqualified class name keys in a config file somewhere but then they wouldn't be bound to a concrete class. Dynamic options like `$className::class` don't work.

robbieaverill: Oh, this can be looked up from injector config

dhensby:
ah - ok

src/ORM/DataObjectSchema.php
robbieaverill: No - we need a class instance to check if it's a composite field or not so composite fields can have their own fields indexed

robbieaverill:
Ok, I've rebased and pushed a new commit that will allow composite fields to be indexed with their fields together, and treats polymorphic has_ones as a composite field in that regard.

I've updated some of the bits as per @tractorcow's suggestions.

Need to add tests another day.

robbieaverill:
Ok, ready for re-review

src/ORM/DataObjectSchema.php
```diff
@@ -486,30 +494,72 @@ protected function cacheDatabaseFields($class)
...
+        $this->defaultDatabaseIndexes[$class] = [];
+
+        $fieldSpecs = $this->fieldSpecs($class, self::UNINHERITED);
+        foreach ($fieldSpecs as $field => $spec) {
+            list($fieldClass, $arguments) = ClassInfo::parse_class_spec($spec);
+            /** @var DBField $fieldObj */
+            $fieldObj = Injector::inst()->create($fieldClass, $field);
+
+            // Polymorphic indexes are handled as a DBComposite, check to avoid adding them twice
+            if ($fieldObj->config()->get('indexed') && array_shift($arguments) !== DataObject::class) {
```
tractorcow: What was that dataobject class check for? Is that `DBClassName(DataObject)`?

robbieaverill: It's for catching polymorphics that have a class name added in - we don't need to index `PolymorphicClassName` on its own as well as the combination of PolymorphicID and *ClassName

src/ORM/DataObjectSchema.php
```diff
@@ -486,30 +494,72 @@ protected function cacheDatabaseFields($class)
...
+                // Either index a composite field's fields, or just the field itself
+                if ($fieldObj instanceof DBComposite) {
```
tractorcow: Could we move this logic into DBField itself? It might be useful for DBFields subclasses to customise their own index specs.

src/ORM/DataObjectSchema.php
tractorcow: Oh, I see. In that case, I suggest moving field-specific index generation (such as this) to the DBPolymorphicForeignKey instance. See my other suggestion.

tractorcow: Hm, maybe some check that skips this field if it's a member of a composite field?

tractorcow: Suggest to control it via spec arguments directly.

```php
```private static $composite_db = array(
        'ID' => 'Int',
        'Class' => "DBClassName('SilverStripe\\ORM\\DataObject', [ 'indexed' => false ])"
    );
```

Similarly to how we specify options for HTMLText.

robbieaverill:
Action points:

* Remove polymorphic check from DataObjectSchema
* Add constructor arg to DBField for setting bool indexed with setter/getter
* Constructor arg or `setIndexed()` value takes precendence over global config for DBField type
* Move logic for "get fields to include in index" to DBField and implement in DBComposite to use `composite_db`, default value in DBField would be the field name

tractorcow:
> Add constructor arg to DBField for setting bool indexed with setter/getter

Be careful, since there's already an $options in DBString. Should that just go up one level to DBField?

robbieaverill:
>Be careful, since there's already an $options in DBString. Should that just go up one level to DBField?

I think that makes sense

robbieaverill:
Ok, good to go again. This isn't really a patch fix any more - I can squash and remove "FIX" from one of the commits when we're good to go!

tractorcow:
It's looking really good @robbieaverill . I'll do some more peer review later this week when I check it out and test it.

src/ORM/DataObjectSchema.php
```diff
@@ -486,30 +493,59 @@ protected function cacheDatabaseFields($class)
...
+        }
+        $this->defaultDatabaseIndexes[$class] = [];
+
+        $fieldSpecs = $this->fieldSpecs($class, self::UNINHERITED);
+        foreach ($fieldSpecs as $field => $spec) {
+            /** @var DBField $fieldObj */
+            $fieldObj = Injector::inst()->create($spec, $field);
+            if ($fieldObj->getIndexed()) {
+                $this->defaultDatabaseIndexes[$class][$field] = [
                     'type' => 'index',
```
tractorcow: If you move the whole array syntax to the DBField, theoretically you could have field-specific index types (e.g. unique, index, fulltext).

robbieaverill: Good call - updated in Good call - updated in 345357f which integrates this into the original approach to keep it simple yet more flexible

dhensby:
@tractorcow should this make the 4.0 release or 4.1? If 4.0 then it needs to go into the 4.0 branch

dhensby:
Nice - this is starting to look a lot like my DBAL approach ;)

robbieaverill:
Happy to rebase and squash as appropriate when requested. Will leave the commits as is for now to help with following peer review

robbieaverill:
Rebased

tractorcow:
@robbieaverill where is 345357f?

robbieaverill:
@tractorcow - sorry, it got killed in the rebase

tractorcow:
@robbie what I was suggesting was
```
$this->defaultDatabaseIndexes[$class][$field] = fieldObj->getIndexSpecs();
```

not

```
$this->defaultDatabaseIndexes[$class][$field] = [
    'type' => $type,
    'columns' => $fieldObj->getIndexSpecs(),
];
```

robbieaverill:
@tractorcow thanks for the review. Implementors of the `DBIndexable` interface are now responsible for returning a full and complete index spec (as validated by `DBSchemaManager::convertIndexSpec`).

PR #7058, opened 2017-06-25
FIX: Allow DB::setConfig() in _config.php

sminnee:
This wasn’t working for two reasons
 - bootDatabase was running before the _config files were loaded
 - It was looking for databaseConfig even if DB::setConfig

This is how the installer sets config so this is an important fix.

### DIFF
--- a/src/Core/CoreKernel.php
+++ b/src/Core/CoreKernel.php
@@ -185,8 +185,10 @@ class CoreKernel implements Kernel
         $this->bootPHP();
         $this->bootManifests($flush);
         $this->bootErrorHandling();
-        $this->bootDatabase();
+        $this->bootDatabaseEnvVars();
         $this->bootConfigs();
+        $this->bootDatabaseGlobals();
+        $this->validateDatabase();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


sminnee:
Added to milestone as it breaks the installer.

tractorcow:
I know it seems odd, but the way it is is necessary for _config.php to be able to augment the existing DB.

The way it works is:

 - bootDatabase() -> loads config from .env into db::setConfig()
 - _config.php -> loads additional settings into db::setConfig()
 - (next request that requires a connection) -> creates connection from current db::getConfig()
 
 If you swap the order then you can't do anything that touches DB in _config.php, which breaks testsession module which does it's DB augmentation in it's _config.php. Testsession requires the existing DB configuration, which would not be available if you switch the order.
 
Remember that db config is lazy loaded; All bootDatabase() does is setup the default connection details. It doesn't commit these to a connection until the first query is made.

> This is how the installer sets config so this is an important fix.

Installer works fine because it's only overwriting any `.env` setting with custom details set in the install form, which of course don't exist if you are setting up with installer. It should be writing to an `.env` file directly, but that can be a later improvement.

tractorcow:
_config.php probably needs to be made obsolete, perhaps in favour of some kind of ModuleKernel system? :P That way you can control module specific boot steps. (pre-db, post-db, etc).

After config is initialised (pretty early on) you could register these via config.

sminnee:
Ok in that I think we need to go


Boot database
Boot configs
Check database config completeness

sminnee:
OK, lol so the testsession code literally only works because "testsession" comes after "app" and "mysite" in the alphabet. :-P It's a circular dependency and is basically in violation of our public API.

We have a number of different config approaches:

 * Env variables
 * DB::setConn() in _config.php
 * $databaseConfig and optionally $database globals in _config.php

Making them all work is a little challenging to say the least, but I'll patch as best I can.

sminnee:
OK — pushed a fix. `bootDatabaseGlobals()` is pretty messy but theres' not a lot of way around that without dropping functionality, and at least the messiness is partitioned into a single function.

Just to reiterate: this fixes a critical bug. Currently .env / env-vars is the only functioning way of configuring database — everything else is broken.

tractorcow:
> Just to reiterate: this fixes a critical bug. Currently .env / env-vars is the only functioning way of configuring database — everything else is broken.

What about using DB::setConfig() in _config.php? I tested that and it worked fine.

Remember globals are deprecated, now, so I'm not including those in acceptance critera.

src/Core/CoreKernel.php
```diff
@@ -199,28 +201,64 @@ protected function bootConfigs()
...
+        global $database;
 
-        // Gracefully fail if no DB is configured
-        if (empty($name)) {
-            $this->detectLegacyEnvironment();
-            $this->redirectToInstaller();
+        $prefix = getenv('SS_DATABASE_PREFIX') ?: 'SS_';
+
+        // Case 1: $databaseConfig global exists. Merge $database in as needed
+        if (!empty($databaseConfig)) {
```
tractorcow: Ok, I think this is a good way to support legacy globals. My support. :D

tractorcow:
Ok, I am happy. I see the bug you were trying to fix (legacy deprecated globals simply break altogether). Even though I want to deprecate these, we still want them to work right? Better upgrade path.

sminnee:
It wasn't just use of the databaseConfig global that was broken - DB::setConfig in _config.php was also broken - not sure why it passed your testing. 

The issue was that the validation failure occurred before _config.php was loaded.

sminnee:
Anyway good to see it merged :-)

tractorcow:
Oh, I see! Thanks for explaining it so that even I could understand it haha. I'm a bit too slow sometimes.

sminnee:
Lol, these things are complicated enough even when we're not in middle of heavily changing them ;-)

PR #7060, opened 2017-06-26
Simplify SecurityTest redirect assertion

chillu:
It causes inexplicable failures in the installer test,
while it's passing running in framework standalone.
This might be related to the cms module being present (with framework tests executed).

Either way, testing that FunctionalTest can follow a redirect
and correctly display the "/" homepage isn't really the point of this test. It should limit its assertion to the presence of a redirect.

See https://travis-ci.org/silverstripe/silverstripe-installer/jobs/246703789

### DIFF
--- a/tests/php/Security/SecurityTest.php
+++ b/tests/php/Security/SecurityTest.php
@@ -202,8 +202,6 @@ class SecurityTest extends FunctionalTest
         /* We have only 1 input, one to allow the user to log in as someone else */
         $this->assertEquals(count($items), 1, 'There is 1 input, allowing the user to log in as someone else.');
 
-        $this->autoFollowRedirection = true;
-
         /* Submit the form, using only the logout action and a hidden field for the authenticator */
         $response = $this->submitForm(
             'MemberLoginForm_LoginForm',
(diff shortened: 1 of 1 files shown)


tests/php/Security/SecurityTest.php
```diff
@@ -214,11 +214,12 @@ public function testLogInAsSomeoneElse()
...
 
         /* We get a good response */
-        $this->assertEquals($response->getStatusCode(), 200, 'We have a 200 OK response');
-        $this->assertNotNull($response->getBody(), 'There is body content on the page');
+        $this->assertEquals($response->getStatusCode(), 302, 'We have a redirection response');
 
         /* Log the user out */
         Security::setCurrentUser(null);
+
+        $this->autoFollowRedirection = true;
```
flamerohr: is this set in `setUp()` or `tearDown()`, then this line is not needed?
To me it seems like a weird side-effect to not capture the previous redirection value and re-set that value here, but instead force to `true`.

chillu: You're right - I've followed a pattern from another test case in the same class. It turned out to be unnecessary because the line in the middle of this test case was removed, making the "outer" ones obsolete. I've removed them now.

dhensby:
CMS behat tests are broken

chillu:
Yeah, it's because @flamerohr did a partial merge to core ;)

sminnee:
I'm merging this as it's unrelated to the failure.

sminnee:
I've re-started the 2 broken builds on https://travis-ci.org/silverstripe/silverstripe-installer/builds/246961444 to see if this fixes it as Ingo has alleged.

flamerohr:
apologies, getting that up sooon!

PR #7061, opened 2017-06-26
Removed ErrorPage

jonom:
(Replaces #7017)

Removed references to ErrorPage so it can be split off in to a separate module. Most of this happens in the CMS (https://github.com/silverstripe/silverstripe-cms/pull/1849)

Related RFC: https://github.com/silverstripe/silverstripe-framework/issues/4149


jonom:
Couple of behat tests failed but they don't look like they're due to this PR? They were passing before...

chillu:
Yeah @flamerohr is fixing core now, sorry for that!

flamerohr:
I've fixed the javascript side, but @tractorcow merged something recently that broke a fair amount of the php side :D

PR #7062, opened 2017-06-26
[WIP] Move ThemeResourceLoader to ResourceLoader

sminnee:
I've tried to keep this minimal so as not to completely hose beta1, but this is a minimal set of changes to provide an API that would let us be more flexible in where modules are saved.

In particular, this will let use move modules to vendor in beta2. Specifically, we could add this as an opt-in feature and have ResourceLoader transparently return the right values.

I'm not sure if we could also introduce BASE_PATH != webroot without breaking APIs, but hopefully that would be possible too.

My goal has been to provide an API for this that was a minimal distance from what we already have, rather than designing the deal API.


sminnee:
OK, so a major weakness of this is that it overlaps with ModuleLoader / ModuleManifest.

sminnee:
I'm starting over with this.

sminnee:
Went back to the drawing board and did #7065 instead.

PR #7063, opened 2017-06-26
Support empty DATABASE_PREFIX

wilr:
Following on from discussion with @tractorcow noticed using latest `master` that a DB Prefix of '' fails.. Causing a database name to contain `SS_`. This would be more elegant in PHP7 using the Null Coalescing operator (`getenv('SS_DATABASE_PREFIX') ?? 'SS_'`) but I take it we still support `5.6` for 4 at least.

### DIFF
--- a/src/Core/CoreKernel.php
+++ b/src/Core/CoreKernel.php
@@ -209,17 +209,16 @@ class CoreKernel implements Kernel
         global $databaseConfig;
         global $database;
 
-        $prefix = getenv('SS_DATABASE_PREFIX') ?: 'SS_';
-
         // Case 1: $databaseConfig global exists. Merge $database in as needed
         if (!empty($databaseConfig)) {
             if (!empty($database)) {
-                $databaseConfig['database'] = $prefix . $database;
+                $databaseConfig['database'] =  $this->getDatabasePrefix() . $database;
             }
 
             // Only set it if its valid, otherwise ignore $databaseConfig entirely
             if (!empty($databaseConfig['database'])) {
                 DB::setConfig($databaseConfig);
+
                 return;
             }
         }
@@ -227,7 +226,8 @@ class CoreKernel implements Kernel
         // Case 2: $database merged into existing config
         if (!empty($database)) {
             $existing = DB::getConfig();
-            $existing['database'] = $prefix . $database;
+            $existing['database'] = $this->getDatabasePrefix() . $database;
+
             DB::setConfig($existing);
         }
     }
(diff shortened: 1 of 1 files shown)


tractorcow:
Looks good to me.

src/Core/CoreKernel.php
```diff
@@ -209,7 +209,7 @@ protected function bootDatabaseGlobals()
         global $databaseConfig;
         global $database;
 
-        $prefix = getenv('SS_DATABASE_PREFIX') ?: 'SS_';
+        $prefix = $this->getDatabasePrefix();
```
sminnee: Remove this variable definition and just put `$this->getDatabasePrefix()` in its place. This will mean it's not called unnecessarily.

wilr:
Implemented feedback as per @sminnee

src/Core/CoreKernel.php
```diff
@@ -357,28 +357,47 @@ protected function getDatabaseConfig()
...
 
     /**
+     * @return string
+     */
+    protected function getDatabasePrefix()
+    {
+        $prefix = getenv('SS_DATABASE_PREFIX');
+
+        if ($prefix === false) {
+            return '';
```
tractorcow: if prefix is false (not declared) shouldn't it be `SS_`? Why does null map to `SS_` but not specifying it maps to empty string?

tractorcow: It seems like you are treating `null` as the result for a missing env var. Check http://php.net/manual/en/function.getenv.php: false means not declared, so should be default value here.

wilr:
@tractorcow sorry it had that initially, flipped it when Sam wanted feedback.

If false (undefined) should default to `SS_` as per 3 did. If set to "' it should be ''

tractorcow:
Ok approved again. :D

kinglozzer:
Why is `SS_` the default prefix now? It doesn’t look like it was in 3.x: https://github.com/silverstripe/silverstripe-framework/blob/3.6.0/conf/ConfigureFromEnv.php#L97

tractorcow:
Well, there is this
https://github.com/silverstripe/silverstripe-framework/blob/3.6.0/conf/ConfigureFromEnv.php#L85

tractorcow:
Hm, maybe it should only be a default `SS_` for SS_DATABASE_CHOOSE_NAME ? @sminnee views?

dhensby:
> Hm, maybe it should only be a default SS_ for SS_DATABASE_CHOOSE_NAME ? @sminnee views?

This was the old behaviour and I think it should be retained...

tractorcow:
Ok, back to old behaviour then. :)

PR #7065, opened 2017-06-27
Abstraction of resource loading

sminnee:
This has 3 key changes:

 * API: ModulePath template global now takes any composer package name.
 * NEW: URL generation now handled by pluggable ResourceURLGenerator service.
 * NEW: Requirements::javascript() and Requirements::css() now support “vendor/package:resource” syntax.

These changes will make it easier to us to fully abstract the following in the future without API breakages:

 * file access from module location
 * file location from URL generation

The goal is that we can do the following after beta1:

 * Allow the optional moving of modules into the vendor/ path
 * Allow the optional moving of the webroot to somewhere other than BASE_PATH (e.g. a public_html subfolder)

----

After merging this PR there are no _necessary_ changes, but you can merge all these, in this order:

 * [ ] https://github.com/silverstripe/silverstripe-siteconfig/pull/66
 * [ ] https://github.com/silverstripe/silverstripe-reports/pull/78
 * [ ] https://github.com/silverstripe/silverstripe-cms/pull/1871
 * [ ] https://github.com/silverstripe/silverstripe-admin/pull/121
 * [ ] https://github.com/silverstripe/silverstripe-assets/pull/37
 * [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/514
 * [ ] https://github.com/silverstripe/silverstripe-campaign-admin/pull/32

### DIFF
--- a/src/Control/SimpleResourceURLGenerator.php
+++ b/src/Control/SimpleResourceURLGenerator.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace SilverStripe\Control;
+
+use SilverStripe\Core\Manifest\ResourceURLGenerator;
+
+/**
+ * Generate URLs assuming that BASE_PATH is also the webroot
+ * Standard SilverStripe 3 operation
+ */
+class SimpleResourceURLGenerator implements ResourceURLGenerator
+{
+    /*
+     * @var string
+     */
+    private $nonceStyle;
+
+    /*
+     * Get the style of nonce-suffixes to use, or null if disabled
+     *
+     * @return string|null
+     */
+    public function getNonceStyle()
+    {
+        return $this->nonceStyle;
+    }
+
+    /*
+     * Set the style of nonce-suffixes to use, or null to disable
+     * Currently only "mtime" is allowed
+     *
+     * @param string|null $nonceStyle The style of nonces to apply, or null to disable
+     */
+    public function setNonceStyle($nonceStyle)
+    {
+        if ($nonceStyle && $nonceStyle !== 'mtime') {
+            throw new InvalidArgumentException('The only allowed NonceStyle is mtime');
+        }
+        $this->nonceStyle = $nonceStyle;
+    }
+
+    /**
+     * Return the URL for a resource, prefixing with Director::baseURL() and suffixing with a nonce
+     *
+     * @param string $relativePath File or directory path relative to BASE_PATH
+     * @return string Doman-relative URL
+     * @throws InvalidArgumentException If the resource doesn't exist
+     */
+    public function urlForResource($relativePath)
+    {
+        $absolutePath = preg_replace('/\?.*/', '', Director::baseFolder() . '/' . $relativePath);
+
+        if (!file_exists($absolutePath)) {
+            throw new InvalidArgumentException("File {$relativePath} does not exist");
+        }
+
+        $nonce = '';
+        if ($this->nonceStyle) {
+            $nonce = (strpos($relativePath, '?') === false) ? '?' : '&';
+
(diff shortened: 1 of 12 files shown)


sminnee:
We can see a better example of how this is used at https://github.com/silverstripe/silverstripe-admin/issues/121

dhensby:
awesome

src/Control/SimpleResourceURLGenerator.php
```diff
@@ -0,0 +1,68 @@
...
+        if ($nonceStyle && $nonceStyle !== 'mtime') {
+            throw new InvalidArgumentException('The only allowed NonceStyle is mtime');
+        }
+        $this->nonceStyle = $nonceStyle;
+    }
+
+    /**
+     * Return the URL for a resource, prefixing with Director::baseURL() and suffixing with a nonce
+     *
+     * @inherit
```
tractorcow: This isn't actually a phpdoc feature. There is `@inheritdoc` but that only inherits the description, not the args / return.

src/Forms/HTMLEditor/TinyMCEConfig.php
```diff
@@ -618,7 +615,7 @@ protected function getEditorCSS()
         $editor = array();
 
         // Add standard editor.css
-        $editor[] = Director::absoluteURL(ltrim($this->getAdminPath() . '/client/dist/styles/editor.css', '/'));
```
tractorcow: getAdminPath can be removed.

src/View/Requirements_Backend.php
```diff
@@ -295,7 +298,7 @@ public function getCombinedFilesFolder()
         return Config::inst()->get(__CLASS__, 'default_combined_files_folder');
     }
 
-    /**
```
tractorcow: Um! Trying not to be pedantic here (I lie I am), but `/**` is required for phpdoc. :)

src/View/Requirements_Backend.php
```diff
@@ -398,6 +401,16 @@ public function setMinifyCombinedFiles($minify)
      */
     public function javascript($file, $options = array())
     {
+        // String of the form vendor/package:resource. Excludes "http://bla" as that's an absolute URL
```
tractorcow: We should put this and the below block into a single method to reduce duplication.

sminnee:
OK @tractorcow @chillu I have (touch wood) fixed all the tests in this now, and it would be good to get your feedback.

tractorcow:
I can smell slightly-incorrect-phpdoc from a mile away. :D

src/Control/SimpleResourceURLGenerator.php
sminnee: will fix

sminnee:
@tractorcow I need to leave phpdoc errors in there so I know that's been reviewed ;-)

I've addressed your fixes and pushed up a separate patch

tractorcow:
Oh no what have I done.

sminnee:
Squashed it all

chillu:
Very very cool, API looks great. We'll need to update the changelog to document some before/after cases, and update our docs to match this approach - but that doesn't block this PR.

tractorcow:
I've done some local testing and seems ok.

Once all is green on behat, merge all.

chillu:
Sweet, merged! I've created a new issue for fixing up the docs: https://github.com/silverstripe/silverstripe-framework/issues/7075

chillu:
@sminnee I've created a PR for the upgrading docs: https://github.com/silverstripe/silverstripe-framework/pull/7076

Issue #7025, opened 2017-06-14
Merge up 3.x security fixes into master


chillu:
We need to do a wider merge from 3.x into master, but the important bit is that we don't release 4.0.0-beta1 with security issues that we've already fixed in the 3.x release line. 

My gut feel: We should try to cherry pick those. A wider merge of 3.x into master seems risky at this point due to the module splits - it's going to be very easy to get things wrong and cause regressions in 4.x.

@dhensby reckons that https://github.com/silverstripe/silverstripe-framework/commit/f225b83e2b81dc544ecfe1e837f954354471957c is the last merge up from 3.x to master (in March).

tasks:

 - [x] [SS-2017-002](https://github.com/silverstripe/silverstripe-framework/commit/f71efb5063c57d823dd130b9bfd018f6ef903d49)
 - [x] [ss-2017-003](https://github.com/silverstripe/silverstripe-cms/commit/41270fcf9980c4be2529d2750c717675548eb617)
 - [x] [ss-2017-004](https://github.com/silverstripe/silverstripe-framework/commit/2b72c0f73b668ddf7c059319da915a6c08652278)
 
 PRs:
 
  - [x] https://github.com/silverstripe/silverstripe-framework/pull/7067
  - [x] https://github.com/silverstripe/silverstripe-cms/pull/1868
  - [x] https://github.com/silverstripe/silverstripe-versioned/pull/19

PR #7067, opened 2017-06-27
[SS-2017-002] FIX Lock out users who dont exist in the DB

tractorcow:
Merge up from 3 for https://github.com/silverstripe/silverstripe-framework/issues/7025


tractorcow:
ok, fixed!

PR #7073, opened 2017-06-28
[WIP] API: Don’t auto-convert update to an insert in DB::manipulate()

### LINKED ISSUE #7072: Unnecessary read query on ORM insert
This goes way back to 2008: 2e955b498e0438bbe24f953f6e670cf069794339...

In order to correct a mistaken update query that should be an insert, there's an extra read query to check if the value is already in the database.

Instead, this should probably use the `Database::affectedRows()` instead.

sminnee:
This feature creates an extra SELECT on every update and its 
only benefit is to paper over other bugs that should be fixed
with better data integrity.

Fixes #7072

### DIFF
--- a/src/ORM/Connect/Database.php
+++ b/src/ORM/Connect/Database.php
@@ -343,12 +343,8 @@ abstract class Database
                         $query->addWhere(array('"ID"' => $writeInfo['id']));
                     }
 
-                    // Test to see if this update query shouldn't, in fact, be an insert
-                    if ($query->toSelect()->count()) {
-                        $query->execute();
-                        break;
-                    }
-                    // ...if not, we'll skip on to the insert code
+                    $query->execute();
+                    break;
 
                 case "insert":
                     // Ensure that the ID clause is given if possible


src/ORM/Connect/Database.php
```diff
@@ -343,12 +343,8 @@ public function manipulate($manipulation)
                         $query->addWhere(array('"ID"' => $writeInfo['id']));
                     }
 
-                    // Test to see if this update query shouldn't, in fact, be an insert
-                    if ($query->toSelect()->count()) {
```
tractorcow: This by itself isn't going to be enough. The insert logic in DataObject::write() initially writes to the base table to get the ID, and that relies on the base table insert being converted to an update here.

I thought the fix was to use affectedRows() to trigger the fallback to update?

sminnee: `affectedRows()` seems to return false if you run an update query that doesn't change any columns.

Well, I know that it erroneously returns false sometimes, I'm guessing that that's the reason why.

sminnee: In any case #7073 has more details.

Imma close this PR as it adds little value keeping it open.

PR #7076, opened 2017-06-28
Docs for resource loading upgrade path

chillu:
See https://github.com/silverstripe/silverstripe-framework/pull/7065

### DIFF
--- a/docs/en/04_Changelogs/4.0.0.md
+++ b/docs/en/04_Changelogs/4.0.0.md
@@ -50,6 +50,7 @@ guide developers in preparing existing 3.x code for compatibility with 4.0
   unless explicitly opted out.
 * Themes are now configured to cascade, where you can specify a list of themes, and have the template engine
   search programatically through a prioritised list when resolving template and CSS file paths.
+* Removed module path constants (e.g. `FRAMWORK_PATH`) and support for hardcoded file paths (e.g. `mysite/css/styles.css`)
 * i18n Updated to use symfony/translation over zend Framework 1. Zend_Translate has been removed.
 * Replaced `Zend_Cache` and the `Cache` API with a PSR-16 implementation (symfony/cache)
 * _ss_environment.php files have been removed in favour of `.env` and "real" environment variables.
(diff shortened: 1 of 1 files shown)


docs/en/04_Changelogs/4.0.0.md
```diff
@@ -108,6 +109,65 @@ Core template locations have moved - if you're including or overriding these
...
+// before
+Requirements::css('framework/css/styles.css');
+
+// after
+Requirements::css('silverstripe/framework: css/styles.css');
+```
+
+Usage with custom logic:
+
+```php
```
chillu: I really wish we could use `diff` syntax here, but the underlying syntax highlighter used for docs.ss.org doesn't support it...

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -108,6 +109,65 @@ Core template locations have moved - if you're including or overriding these
...
+$baseFilePath = BASE_PATH . '/composer.json'; 
+$mysiteFilePath = 'mysite/css/styles.css';
+$themesFilePath = SSViewer::get_theme_folder() . '/css/styles.css';
+$themeFolderPath = THEMES_DIR . '/simple';
+
+// after
+use SilverStripe\Core\Manifest\ModuleLoader;
+use SilverStripe\View\ThemeResourceLoader;
+$moduleFilePath = ModuleLoader::getModule('silverstripe/framework')->getRelativeResourcePath('MyFile.php');
+$baseFilePath = Director::baseFolder() . '/composer.json';
```
chillu: Not quite sure if this is the right way to do this now? It sure beats `BASE_PATH`, but it's still static.

docs/en/04_Changelogs/4.0.0.md
robbieaverill: Bah - does parsedown not support diff?

chillu: Parsedown doesn't do syntax highlighting? We're using https://github.com/google/code-prettify as a JS lib, and that doesn't do diff: https://github.com/google/code-prettify/issues/62. Ideas?

robbieaverill: Doh, of course. I feel like we have had this chat before, would need a separate JS lib just for diff highlighting

robbieaverill: [highlight.js](https://highlightjs.org) supports a large range of syntaxes, including diff - perhaps we could switch?

chillu: https://github.com/silverstripe/silverstripe-docsviewer/pull/134 :)

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -108,6 +109,65 @@ Core template locations have moved - if you're including or overriding these
...
+$mysiteFilePath = ModuleLoader::getModule('mysite')->getRelativeResourcePath('css/styles.css');
+$themesFilePath = ThemeResourceLoader::inst()->findThemedResource('css/styles.css');
+$themeFolderPath = ThemeResourceLoader::inst()->getPath('simple');
+```
+
+To ensure consistency, we've also removed support for path constants:
```
tractorcow: We haven't removed these.

chillu: Replaced "removed" with "deprecated"

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -108,6 +109,65 @@ Core template locations have moved - if you're including or overriding these
...
+$themeFolderPath = ThemeResourceLoader::inst()->getPath('simple');
+```
+
+To ensure consistency, we've also deprecated support for path constants:
+
+ * `FRAMEWORK_DIR` and `FRAMEWORK_PATH`
+ * `FRAMEWORK_ADMIN_DIR` and `FRAMEWORK_ADMIN_PATH`
+ * `FRAMEWORK_ADMIN_THIRDPARTY_DIR` and `FRAMEWORK_ADMIN_THIRDPARTY_PATH`
+ * `THIRDPARTY_DIR` and `THIRDPARTY_PATH`
+ * `ASSETS_DIR` and `ASSETS_PATH`
```
tractorcow: I think we have a "`ASSETS_PATH`" env lookup in config f or assets store configuration, but we could still deprecate it I think.

chillu: You're right, it's too early to deprecate that - there's no appropriate getter. Which we'll need to solve, but it's out of scope of this docs-focused PR

Issue #5441, opened 2016-05-03
DataObject::get_one() return value (boolean vs. null)


patricknelson:
Hi, this may be an interesting discussion... but since I upgraded to SS `3.3` from `3.1` (since there used to be a bug in the way type hinting was being handled, i.e. ignored) I've been encountering bugs that would otherwise have arisen before due to the return value from this method. Basically if it doesn't find anything it returns either a boolean (only ever `false`) or an instance of the particular `DataObject` child in question.

Primary line responsible: https://github.com/silverstripe/silverstripe-framework/blob/master/model/DataObject.php#L3204

``` php
DataObject::$_cache_get_one[$callerClass][$cacheKey] = false;
```

I'd propose that we do one of the following:
1. Change the return (when not found), and thus the type, from `false` to `null` and document this as a nullable return value, e.g. `@return DataObject|null` or
2. At least document the fact that the return value has two types and clarify in the documentation that `bool` is always `false`, i.e. `@return DataObject|bool False if not found.`

What do you think? Is my intuition off by assuming that returning `null` is better in this circumstances since then type-hints will not fail if they're [nullable](https://en.wikipedia.org/wiki/Nullable_type)?

sminnee:
I expect that the change from `null` to `false` for an unfound value is an oversight.

The main thing is to ensure that such misses are cached. I think that putting `false` into the cache makes it easier to distinguish from no cached data than `null`.

I'd leave line 3204 as-is, but have the return value of `DataObject::get_one()` be `null` if the cache value is `false`.

sminnee:
Alternatively I guess you could use `array_key_exists()` rather than `isset()` on line 3197, and set the cache to `null`.

At any rate, I would add a type-sensitive test for this behaviour, so we don't miss it again.

kinglozzer:
I’d be happy to see this changed to `null` in master, it’d match `DataList::first()` and co (as an aside: the docs for those methods should probably be changed to indicate that `null` will be returned if no results are found).

Personal preference is for `array_key_exists()` and storing `null` in the cache. IIRC it’s marginally slower, but better than maintaining extra logic to swap between `null` and `false`.

patricknelson:
[...]

PR #7083, opened 2017-06-29
FIX: DataObject::get_one() misses return null, not false

sminnee:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/5441

Also merge:

 * [ ] https://github.com/silverstripe/silverstripe-cms/pull/1874


sminnee:
OK everything but the CMS test is passing — the CMS test requires the aforementioned PR.

PR #7084, opened 2017-06-29
API: Don’t allow modules in 2nd level directories

sminnee:
This was a setting that was only used for framework/admin, which has
since been removed. I’ve patched this so that we don’t need to support
this feature for the 4.x lifetime.

Post beta1 we’ll likely add support for modules in vendor, but that’s
a new feature so doesn’t require a major change.


sminnee:
This was bogus

PR #7085, opened 2017-06-29
Pulls/4.0/shortcode namespacing

unclecheese:
Per https://github.com/silverstripe/silverstripe-framework/pull/6126

### DIFF
--- a/_config.php
+++ b/_config.php
@@ -1,9 +1,9 @@
 <?php
 
-use SilverStripe\Assets\File;
-use SilverStripe\Assets\Image;
 use SilverStripe\Dev\Deprecation;
-use SilverStripe\Forms\HtmlEditor\EmbedShortcodeProvider;
+use SilverStripe\View\ViewSupport\EmbedShortcodeProvider;
+use SilverStripe\Assets\ViewSupport\FileShortcodeProvider;
+use SilverStripe\Assets\ViewSupport\ImageShortcodeProvider;
 use SilverStripe\View\Parsers\ShortcodeParser;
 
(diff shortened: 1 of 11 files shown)


composer.json
```diff
@@ -79,6 +79,7 @@
       "SilverStripe\\Security\\Tests\\": "tests/php/Security/",
       "SilverStripe\\View\\": "src/View/",
       "SilverStripe\\View\\Tests\\": "tests/php/View/",
+      "SilverStripe\\View\\ViewSupport\\Tests\\": "tests/php/View/ViewSupport",
```
tractorcow: Namespace should be `SilverStripe\View\Tests\<others>` and thus not need a separate composer.json declaration.

_config.php
```diff
@@ -15,13 +15,13 @@
  */
 
 ShortcodeParser::get('default')
-	->register('file_link', array(File::class, 'handle_shortcode'))
-	->register('embed', array(EmbedShortcodeProvider::class, 'handle_shortcode'))
-	->register('image', array(Image::class, 'handle_shortcode'));
+	->register('file_link', [FileShortcodeProvider::class, 'handle_shortcode'])
```
tractorcow: Can we move these to assets module.

tractorcow: Leave EmbedShortcodeProvider in framework, since it's a non-assets module class.

tractorcow: Also looks like you're using tabs for indentation here.

tractorcow:
All dependencies have been merged. Please merge this PR on green.

Issue #7051, opened 2017-06-22
`?flush` is broken, `?flush=1` works


sminnee:
Since the app object refactor, the short syntax of `?flush` no longer works.

robbieaverill:
@sminnee seems to work OK for me testing like this:

```diff
diff --git a/src/Control/FlushRequestFilter.php b/src/Control/FlushRequestFilter.php
index 2f7c0a694..e0db27017 100644
--- a/src/Control/FlushRequestFilter.php
+++ b/src/Control/FlushRequestFilter.php
@@ -13,10 +13,12 @@ class FlushRequestFilter implements RequestFilter
     public function preRequest(HTTPRequest $request)
     {
         if (array_key_exists('flush', $request->getVars())) {
+            die('yes');
             foreach (ClassInfo::implementorsOf(Flushable::class) as $class) {
                 $class::flush();
             }
         }
+        die('no');
         return true;
     }
 
```

sminnee:
This isn't used for all flush tasks — the config flush is different code.

PR #7090, opened 2017-06-29
Fix enable ?flush rather than just ?flush=1

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/7051


tractorcow:
Tested locally and works fine.

PR #7091, opened 2017-06-29
Move to travis trusty dist

dhensby:
I've seen some speed improvements on trusty, lets see if we get them for framework

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: php
 
+dist: trusty
+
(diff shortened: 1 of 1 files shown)


dhensby:
gah - well, it's about twice as fast but we can't run php 5.3 builds on it

tractorcow:
> gah - well, it's about twice as fast but we can't run php 5.3 builds on it


Oh, how about master?

PR #7092, opened 2017-06-29
Move to travis trusty dist

dhensby:
#7091 but on master

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,7 @@
         "silverstripe/versioned": "^1.0@dev",
         "silverstripe/behat-extension": "^3",
         "silverstripe/serve": "dev-master",
-        "se/selenium-server-standalone": "2.41.0"
+        "se/selenium-server-standalone": "^3.4"
     },
     "provide": {
         "psr/container-implementation": "1.0.0"
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,6 +86,10 @@ before_script:
   - if [[ $BEHAT_TEST ]]; then cp composer.lock artifacts/; fi
   - if [[ $BEHAT_TEST ]]; then sh -e /etc/init.d/xvfb start; sleep 3; fi
   - if [[ $BEHAT_TEST ]]; then (vendor/bin/selenium-server-standalone > artifacts/selenium.log 2>&1 &); fi
+  - if [[ $BEHAT_TEST ]]; then export GECKODRIVER_DOWNLOAD=`curl -s 'https://api.github.com/repos/mozilla/geckodriver/releases/latest' | python -c "import sys, json; r = json.load(sys.stdin); print([a for a in r['assets'] if 'linux64' in a['name']][0]['browser_download_url']);"`; fi
+  - if [[ $BEHAT_TEST ]]; then curl -L -o geckodriver.tar.gz $GECKODRIVER_DOWNLOAD; fi
+  - if [[ $BEHAT_TEST ]]; then gunzip -c geckodriver.tar.gz | tar xopf -; fi
+  - if [[ $BEHAT_TEST ]]; then chmod +x geckodriver && export PATH=`pwd`/geckodriver:$PATH; fi
   - if [[ $BEHAT_TEST == framework ]]; then (vendor/bin/serve --bootstrap-file tests/behat/serve-bootstrap.php &> artifacts/serve.log &); fi
   - if [[ $BEHAT_TEST == cms ]]; then (vendor/bin/serve --bootstrap-file cms/tests/behat/serve-bootstrap.php &> artifacts/serve.log &); fi
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,13 @@ addons:
 env:
   global:
     - COMPOSER_ROOT_VERSION=4.0.x-dev
-    - DISPLAY=":99"
-    - XVFBARGS=":99 -ac -screen 0 1024x768x16"
+    - DISPLAY=:99.0
+    - XVFBARGS=":99.0 -ac -screen 0 1024x768x16"
+
+_browsers:
+  firefox: &firefox-esr
+    addons:
+      firefox: latest-esr
 
 matrix:
   fast_finish: true
@@ -40,6 +45,7 @@ matrix:
       env:
        - DB=MYSQL
        - BEHAT_TEST=framework
+      <<: *firefox-esr
     - php: 7.0
       env:
         - DB=MYSQL
@@ -48,6 +54,7 @@ matrix:
       env:
         - DB=MYSQL
         - BEHAT_TEST=cms
+      <<: *firefox-esr
 
 before_script:
 # Init PHP
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: php
 
-dist: precise
+dist: trusty
 
 sudo: false
 


tractorcow:
Oh cool :D

dhensby:
It looked like behat had problems too - lets see if it works on master

tractorcow:
Oh no, pear again. =(

dhensby:
Selenium doesn't seem to be working either :/

dhensby:
pear install seems fine

tractorcow:
If you push to the open-sourcerer fork, you'll get artifacts upload to aws. Then you can download selenium.log.

tractorcow:
Ok, I've re-enabled travis and added you with write permissions to https://github.com/open-sausages/silverstripe-framework

dhensby:
gah - I'm not sure I have the motivation to look into this. the speed improvements on trusty look variable at best...

I think I'll have more luck trying to get selenium running on a local trusty vm first

tractorcow:
I've found that there are some additional behat breakages caused by a recent release of php-webdriver module.

Fix is at https://github.com/instaclick/php-webdriver/pull/79

dhensby:
still no luck

tractorcow:
Errors remind me why I stopped using pear.

dhensby:
the behat tests not running are really the big issue here, but I've pushed an attempted fix to use composer to install codesniffer

dhensby:
I've tried to get selenium running on a local trust vm and can't - so I'm going to bail at this point. If anyone has any ideas of how to fix this on travis, then please take a look, otherwise I'll close this.

dhensby:
@chillu do you have any interest in investigating the selenium issues?

composer.json
```diff
@@ -94,8 +95,8 @@
         "thirdparty/"
     ],
     "scripts": {
-        "lint": "phpcs src/ tests/php",
-        "lint-clean": "phpcbf src/ tests/php",
+        "lint": "vendor/bin/phpcs src/ tests/php",
```
tractorcow: This isn't going to work if framework isn't root, meaning these scripts are no longer useful to devs.

dhensby: Sure

dhensby:
So FF52 is broken with selenium 3.4.0 (we need 3.4.1 to be released - https://github.com/mozilla/geckodriver/issues/676).. so this is stuck until then

dhensby:
3.5.0 has been released - now we wait for standalone server to be released (https://github.com/sveneisenschmidt/selenium-server-standalone/issues/29)

tractorcow:
mink-selenium2-driver doesn't even have firefox support yet.

https://github.com/minkphp/MinkSelenium2Driver/issues/254

Maybe we shift to chrome for now?

tractorcow:
Selenium is falling apart, so I'm triaging and milestoning this for 4.0 stable.

tractorcow:
Looks like we need to install geckodriver to get support for firefox... I'll try to set this up before I dismiss the possibility immediately. :)

tractorcow:
Latest ESR IS 52 of firefox.... we may need to lock our expected browser requirements to that (currently it's requesting v 9!)

```

  ┌─ @BeforeScenario # SilverStripe\Framework\Tests\Behaviour\FeatureContext::before()
  │
  ╳  Could not open connection: TypeError: Given browserVersion [object String] "9", but current version is [object String] "55.0.3"
  ╳  Build info: version: '3.5.3', revision: 'a88d25fe6b', time: '2017-08-29T12:54:15.039Z'
  ```

**Comment by dhensby** [comment]
OK - I've rerun the suite and it's still failing - I'm just at the point where I'm not really sure how to proceed... @tractorcow do you want to take it on from here?

**Comment by tractorcow** [comment]
Yep I'm still working on it... there is quite a bit left to do (E.g. setup geckdriver, mink configuration upgrades)

**Comment by tractorcow** [comment]
It looks like this could be the cause of recent failures on travis precise dist:

```
With all incoming jobs defaulted to dist: trusty, the remaining transition we will be focusing on is phasing out our container-based infrastructure for dist: precise jobs. This rollout will be similar to the dist: trusty defaulting, but instead of setting a default value, we will begin ignoring sudo: false when we find an explicit dist: precise.
```

From https://blog.travis-ci.com/

**Comment by dhensby** [comment]
what a pain :/

**Comment by tractorcow** [comment]
New issue, https://github.com/instaclick/php-webdriver doesn't work with selenium 3.

**Comment by tractorcow** [comment]
My next step would be to fork MinkSelenium2Driver module and update support for selenium3. Not going to do that now though; Maybe next hackday?

**Comment by tractorcow** [comment]
My progress on upgrading to facebook webdriver:

image

**Comment by chillu** [comment]
Just getting my head around this - here's the chain of events as far as I can tell:

1. Travis limited precise support ([removed sudo:false](https://blog.travis-ci.com/2017-08-31-trusty-as-default-status)), need to move to trusty. Travis will remove precise support completely next April.
1. We need `sudo: false` to run on container-based infrastructure for fast builds ([explanation](https://docs.travis-ci.com/user/reference/trusty/#Container-based-with-sudo%3A-false)). We could switch to `sudo: required` to stay on trusty, but that'll slow down builds
1. trusty doesn’t support old Firefox releases, need to move to ESR
1. Selenium 2 doesn't support Firefox ESR, need to update to Selenium 3
1. mink-selenium2-driver doesn’t support Firefox on Selenium 3, shift to Chrome?
1. php-webdriver doesn’t support Selenium 3, fork and fix?

@tractorcow The above is an updated version compared to what I posted on Slack.

**Comment by tractorcow** [comment]
It looks like some of the issues I noticed upgrading to selenium3 were my own fault; Regarding https://github.com/facebook/php-webdriver/issues/468, you should only need to set `--enablePassThrough false` when starting selenium.

**Comment by tractorcow** [comment]
I have most framework tests passing, but a few iframe switching bugs.

Working branches:

 - https://github.com/open-sausages/MinkSelenium2Driver/commits/master
 - https://github.com/open-sausages/silverstripe-behat-extension/commits/pulls/4.0/selenium-upgrade

**Comment by tractorcow** [comment]
Also followed https://github.com/minkphp/Mink/issues/158; The lack of this core API means having to upgrade various behat contexts to support facebook/webdriver directly (and undesirably).

**Comment by chillu** [comment]
Note that I've switched to trusty without any other changes on cms, with successful builds: https://travis-ci.org/silverstripe/silverstripe-cms/jobs/274442114

I've created PRs to do a straight-up switch to precise for framework - just to see what the actual problem is (that's not mentioned in this thread AFAICT):

3.x PR for framework: https://github.com/silverstripe/silverstripe-framework/pull/7366
4.x PR for framework: https://github.com/silverstripe/silverstripe-framework/pull/7367

My theory is that we just need to pin the Firefox dependency to something that still supports Selenium 2.

**Comment by chillu** [comment]
So those builds both fail with:

```
[Behat\Mink\Exception\DriverException]                                                                                         
  Exception has been thrown in "beforeScenario" hook, defined in SilverStripe\Framework\Test\Behaviour\FeatureContext::before()  
  Could not open connection: Unable to connect to host localhost on port 7055 after 45000 ms. Firefox console output:            
  nf-WARNING **: Client failed to connect to the D-BUS daemon:                                                                   
  //bin/dbus-launch terminated abnormally without any error message      
```

3.x: https://travis-ci.org/silverstripe/silverstripe-framework/jobs/274445959
4.x: https://travis-ci.org/silverstripe/silverstripe-framework/jobs/274446637

Now I'm updating them with a pinned Firefox version (31.0, same as [cms travis config](https://github.com/silverstripe/silverstripe-cms/blob/4/.travis.yml#L13)) and see if that's any different.

tractorcow:
More bad news, looks like upgrading to facebook/php-webdriver is blocked by https://github.com/facebook/php-webdriver/issues/462.

chillu:
OK, both those builds have passed - so the only thing that was necessary to fix this (apart from switching to trusty in the first place) was pinning of the FIrefox version we already had on CMS.

Selenium [states](http://www.seleniumhq.org/about/platforms.jsp) that it's supporting the current Firefox ESR - which would be v52 at the moment. I haven't seen any notice about Selenium 2 being unsupported (nothing mentioned in the [announcement](https://seleniumhq.wordpress.com/2016/10/13/selenium-3-0-out-now/)). It'll become increasingly infeasible to stay on Selenium 2, but looks like we don't have an immediate problem there.

There's a [PoC with Behat/Mink in Chrome Headless](https://www.reddit.com/r/PHP/comments/6axmyu/chrome_headless_support_without_selenium_for_mink/), which is said to be 3x as fast as the equivalent through Selenium - so that's promising. And it would mean we skip Selenium entirely (no longer needing Java would be nice...). I'm happy to keep going on Selenium 2 for a while and hope that someone gets to a stable version of that PoC in the meantime :D

tractorcow:
Replaced with https://github.com/silverstripe/silverstripe-framework/pull/7367

PR #7095, opened 2017-06-30
Update nginx configuration

oddnoc:
- Add buffer parameters without which some parts of the CMS fail with "too big header"
- Use RFC-2606-compliant example domains
- Remove configuration that allows arbitrary PHP execution
- Improve sample denials to show how to secure modules
- Restore missing boilerplate for MIME types

### DIFF
--- a/docs/en/00_Getting_Started/01_Installation/How_To/Configure_Nginx.md
+++ b/docs/en/00_Getting_Started/01_Installation/How_To/Configure_Nginx.md
@@ -18,83 +18,75 @@ Especially be aware of [accidental php-execution](https://nealpoole.com/blog/201
 
 But enough of the disclaimer, on to the actual configuration — typically in `nginx.conf`:
 
-	server {
-		include mime.types;
-		default_type  application/octet-stream;
-		client_max_body_size 0; # Manage this in php.ini
-		listen 80;
-		root /path/to/ss/folder;
-		server_name example.com www.example.com;
+```nginx
+server {
+  include mime.types;
+  default_type  application/octet-stream;
+  client_max_body_size 0; # Manage this in php.ini
+  listen 80;
+  root /path/to/ss/folder;
+  server_name example.com www.example.com;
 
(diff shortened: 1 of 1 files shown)


docs/en/00_Getting_Started/01_Installation/How_To/Configure_Nginx.md
```diff
@@ -19,88 +19,86 @@ Especially be aware of [accidental php-execution](https://nealpoole.com/blog/201
...
-The above configuration sets up a virtual host `site.com` with
-rewrite rules suited for SilverStripe. The location block for php files
-passes all php scripts to the FastCGI-wrapper via a TCP socket.
+		# Core denial (change mysite if you use a different name)
+        location ~ /(mysite|framework|cms)/.*\.(php|php3|php4|php5|phtml|inc)$ {
+            deny all;
+        }
+
+        # Modules denial (edit the regex to match your installed modules)
+        location ~ /(buildtools|colorpicker|docsviewer|editlock|geoip|googlesitemaps|mathspamprotection|sortablegridfield|spamprotection|testsession|userforms)/.*\.(php|php3|php4|php5|phtml|inc)$ {
```
robbieaverill: I'm not a very big fan if whitelisting modules like this. Can we just include the same thing in the clause above with maybe the "userforms" module since it's one of the most popular, and include the same comment "edit the regex to match your installed modules"?

dhensby:
Would a better approach be just to whitelist framework/main.php (and the tinymce scripts) and block all other PHP scripts... If devs have PHP scripts they need to run outside of SilverStripe they should be manually whitelisting them, rather than trying to blacklist every module and folder

docs/en/00_Getting_Started/01_Installation/How_To/Configure_Nginx.md
oddnoc: Agreed. Turns out the blacklist of modules isn't needed. Removed it in a new commit.

oddnoc:
Shall I squash these into a single commit?

PR #7103, opened 2017-07-02
added null check, so that zeros don’t turn up unexpectantly

pitchandtone:


tractorcow:
It seems ok, but it would be good to have a test. @pitchandtone could you add a check for empty strings and nulls being converted to null?

pitchandtone:
Tests added.

tests/php/Forms/NumericFieldTest.php
```diff
@@ -127,6 +127,16 @@ public function testNumberTypeOnInputHtml()
         $this->assertContains('type="text"', $html, 'number type not set');
     }
 
+    public function testNullSet() {
+        $field = new NumericField('Number');
+        $field->setValue('');
+        $this->assertEquals($field->Value(), null);
```
tractorcow: Sorry to be pedantic, but the expected value is on the left, expression on the right. :)

src/Forms/NumericField.php
```diff
@@ -154,6 +154,9 @@ public function setValue($value, $data = null)
      */
     protected function cast($value)
     {
+        if (strlen($value) === 0) {
+            return false;
```
tractorcow: Why is false being cast for a numeric field? I would expect `null` to be used here.

tests/php/Forms/NumericFieldTest.php
```diff
@@ -127,6 +127,16 @@ public function testNumberTypeOnInputHtml()
...
+        $field->setValue(null);
+        $this->assertEquals($field->Value(), null);
+        $field->setValue(0);
+        $this->assertEquals($field->Value(), 0);
```
tractorcow: Can you also please test dataValue()

src/Forms/NumericField.php
pitchandtone: It's set to false, so that in the Value function it doesn't end up being pushed into a formatter.

I can update the Value function to check is null and false if you'd prefer.

tractorcow: Let's try null and see if we get better value for money :) It feels more semantically correct than false for empty numeric values.

pitchandtone:
Reworked now with null check.

tests/php/Forms/NumericFieldTest.php
```diff
@@ -127,6 +127,19 @@ public function testNumberTypeOnInputHtml()
         $this->assertContains('type="text"', $html, 'number type not set');
     }
 
+    public function testNullSet() {
+        $field = new NumericField('Number');
+        $field->setValue('');
+        $this->assertEquals(null, $field->Value());
```
dhensby: `assertNull` would be a more appropriate assertion here - but I won't block merge over it

dhensby:
I've rebased this onto 4.0 so that it makes the 4.0 release

tests/php/Forms/NumericFieldTest.php
dhensby: OK - I fixed this as part of the rebase

dhensby:
It appears the tests aren't passing :/

tests/php/Forms/NumericFieldTest.php
```diff
@@ -127,6 +127,19 @@ public function testNumberTypeOnInputHtml()
         $this->assertContains('type="text"', $html, 'number type not set');
     }
 
+    public function testNullSet() {
+        $field = new NumericField('Number');
+        $field->setValue('');
+        $this->assertNull($field->Value());
```
tractorcow: It's coming up as `''`, which is probably should be the expected value for this test. Only the datavalue() should be null, although it emits an empty string to the user through the front-end Value() method

I suggest:
 - changing setValue() to setSubmittedValue() (to simulate a user submitting an empty string)
 - expect these invalid strings to be returned as-is ($this->originalValue).

dhensby: I'll leave this for @pitchandtone to tidy up. Remember to pull the commit locally that I've pushed up...

pitchandtone:
@dhensby I've done the rebase on 4.0 here, and fixed that text, let me know if there's anything missing.

dhensby:
There's a code style issue, but I'll merge by hand and fix it.

tractorcow:
Good job @pitchandtone and thanks for picking up the extra cleanup @dhensby

Issue #7104, opened 2017-07-02
Ensure upgrader is tested with master


tractorcow:
Task before beta1: Ensure that upgrader can be safely run on code prior to tagging beta1.

Acceptance criteria:
 - Running upgrader doesn't break anything
 - Upgrader doesn't upgrade to any classes that don't exist
 
PRS:
 - [x] https://github.com/silverstripe/silverstripe-upgrader/pull/36
 - [x] https://github.com/silverstripe/silverstripe-admin/pull/133 (merge first)
 
Then:
 
  - [x] https://github.com/silverstripe/silverstripe-assets/pull/41
  - [x] https://github.com/silverstripe/silverstripe-framework/pull/7106
  - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/521
  - [x] https://github.com/silverstripe/silverstripe-testsession/pull/52
  - [x] https://github.com/silverstripe/silverstripe-versioned/pull/20
  - [x] https://github.com/silverstripe/silverstripe-graphql/pull/107
  - [x] https://github.com/silverstripe/silverstripe-reports/pull/79
  - [x] https://github.com/silverstripe/silverstripe-cms/pull/1879
  - [x] https://github.com/silverstripe/silverstripe-campaign-admin/pull/33
  - [x] https://github.com/silverstripe/silverstripe-framework/pull/7109

PR #7106, opened 2017-07-03
BUG Implement or exclude all pending upgrader deltas

tractorcow:
Part of https://github.com/silverstripe/silverstripe-framework/issues/7104


_config/logging.yml
```diff
@@ -17,7 +17,7 @@ SilverStripe\Core\Injector\Injector:
     constructor:
       - "error-log"
     calls:
-      DisplayErrorHandler: [ pushHandler, [ %$Monolog\Handler\HandlerInterface ] ]
+      pushDisplayErrorHandler: [ pushHandler, [ %$Monolog\Handler\HandlerInterface ] ]
```
flamerohr: I'd like to see the docs updated for this as well, as it's an unclear part of the API :D

tractorcow: PR coming separately

PR #7118, opened 2017-07-04
FIX: Fixes #7116 Improves server requirements docs viz: OpCaches.

### LINKED ISSUE #7116: Incomplete OpCache docs
The current `00_Server_Requirements.md` docs for both 3.6 and 4.x make no mention of Zend OpCache and its being installed by default as of PHP >= 5.5 - it may also help to mention that enabling >1 such cache can cause "undefined" problems.

See: https://docs.silverstripe.org/en/3/getting_started/server_requirements/ and https://docs.silverstripe.org/en/4/getting_started/server_requirements/

phptek:


kinglozzer:
>If you're using PHP >= 5.5 then Zend OpCache is installed by default in managed package repos e.g. apt

Given that, I wonder if we should just remove the suggestion of installing an opcache entirely. SS4 only supports 5.6+ anyway, so everyone will get an opcache unless they’re building from source right?

docs/en/00_Getting_Started/00_Server_Requirements.md
```diff
@@ -12,6 +12,7 @@ Our web-based [PHP installer](installation/) can check if you meet the requireme
  * Once PHP versions become [unsupported by the PHP Project](http://php.net/supported-versions.php),
    we drop support for those versions in the [next minor release](/contributing/release-process). This means that PHP 5.6 support may be dropped in a 4.x minor release after December 2018.
  * We recommend using a PHP accelerator or opcode cache, such as [xcache](http://xcache.lighttpd.net/) or [WinCache](http://www.iis.net/download/wincacheforphp).
+     * If you're using PHP >= 5.5 then [Zend OpCache](http://php.net/manual/en/book.opcache.php) is installed by default in managed package repos e.g. apt (You may need to manually enable it in php.ini). Do not try and run additional opcaches alongside Zend OpCache without first disabling it.
```
dhensby: This definitely is not true for CentOS - I have to install it manually to get it in (`yum install php-opcache`)

phptek:
True. But the docs for 3.x should keep this. Also note it may still be useful for those that may be lumbered with xcache and the friendly sysadmin is left scratching his head when upgrading from <5.5 to >5.5 and wondering why his devs are reporting non-app specific weird behavior..happened only about 2 months ago!

phptek:
@dhensby good spot. What say this patch is modified as a side note via an asterisk, and made more general i.e "Some package managers such as apt, enable OpCache by default...." ?

dhensby:
> @dhensby good spot. What say this patch is modified as a side note via an asterisk, and made more general i.e "Some package managers such as apt, enable OpCache by default...." ?

I don't think it has anything to do with the package manager itself, it's up to the repository you're installing from. For example it may be installed by default for Ubuntu's own repo but not for Debian's.

I'd probably be very generic here and say something to the effect of "in some cases it is installed by default and you should make sure you don't run two side-by-side as it'll have unexpected consequences"...

phptek:
I have re-written things in line with the discussion so far, let me know your collective thoughts.

dhensby:
thanks - now, should this go in 3 branch ?

flamerohr:
This looks suitable for both 3 and 4, though maybe double check the PHP 5.5 for 3? :D

flamerohr:
Changed to the `3.5` branch and rebased

flamerohr:
Merging, doc changes shouldn't affect tests :) ...I hope...

Issue #6880, opened 2017-05-08
Make CMS logo title configurable


robbieaverill:
**Version:** 3.x

As discussed in the community slack channel, for CWP it would be nice to show the CWP recipe version when hovering over the SS logo in the top left of the CMS:

![silverstripe hover](https://cloud.githubusercontent.com/assets/5170590/25791247/16e7cac0-3413-11e7-87f9-bb74151ef5e8.png)

`LeftAndMain::CMSVersion` could be updated to use a configurable list of modules to add, so cwp/cwp-core could add itself via YAML configuration.

### Question:

Potentially we could remove the use of the `silverstripe_version` file and rely on reading the `composer.lock` file instead (it's probably reasonably safe, but not entirely safe to assume that SilverStripe sites will be installed with Composer nowadays?).

---

At the same time, if the framework and CMS are both installed, just show the CMS version rather than both.

sminnee:
Assuming that composer is installed for anything related to modules/versioning is okay as long as it's not a critical feature. Support for non-composer installs is an edge case.

In SS4 non-composer installs are completely unsupported, but for SS3 we still allow them.

They're certainly more supported than silverstripe_version.

robbieaverill:
@sminnee how do you feel about an approach like this? https://gist.github.com/robbieaverill/4410297b04bd2df19f353b9e1e4f8ee7

It feels like it could definitely go down the configuration route, but would possibly need to be in SS4 to remove the support for the silverstripe_version file.

sminnee:
I'd suggest using an Injector service such as VersionNumberProvider rather than extensions.

Have an interface and a default implementation, and potentially provide config setters for the default implementation that let you choose which modules to interrogate.

sminnee:
Not sure whether a service should be "provide version string" or "provide app metdata". Other metadata:

 * App name
 * App link
 * App logo

robbieaverill:
You'd probably want to avoid having each module parse and read composer.lock, but I like the service approach

sminnee:
Yeah - one service for the entire project. It can do whatever iteration over modules it thinks it needs.

A composite system where you have multiple version providers that are aggregated is probably overengineering. The nice thing about a service system is that someone could add that as a module if they really needed it, so we can ignore the use-case in core.

robbieaverill:
Ok initial PR at #7125

chillu:
@robbieaverill Does that PR fix the issue here?

[...]

PR #7125, opened 2017-07-04
NEW Add new SilverStripeVersionProvider to provider module versions

robbieaverill:
Issue: #6880 

This PR adds a new service provider for gathering module versions from the composer lock file, and replaces the logic in `LeftAndMain::CMSVersion`.

Example configuration to add modules:

```yaml
SilverStripeVersionProvider:
  modules:
    silverstripe/framework: Framework
    silverstripe/somemodule: Some Module
```

The `getVersion` method would return `Framework: 3.7.0, Some Module: 1.2.3` for example.

The CMS (see PR) can block the framework version from being displayed in favour of the CMS. Exaple `CMS: 3.7.0` instead of `Framework: 3.7.0, CMS: 3.7.0` - assumption being that in 99% of cases the CMS and framework versions will be identical. This will also allow CWP to add its version to the tooltip as well, e.g. `SilverStripe: 3.7.0, CWP recipe: 1.7.0` for example.

---

It's worth mentioning that `LeftAndMain::CMSVersion` used to try and fall back to loading the `silverstripe_version` file if a composer lock doesn't exist and get the version from that. I didn't port this change over because the file is currently empty, so it wouldn't result in different behaviour from 3.7.x onwards - and most installs would be with composer now.

---

Related PR for CMS: https://github.com/silverstripe/silverstripe-cms/pull/1881

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -17,3 +17,9 @@ HTTP:
     must-revalidate: "true"
     no-transform: "true"
   vary: "Cookie, X-Forwarded-Protocol, User-Agent, Accept"
+LeftAndMain:
+  dependencies:
+    versionProvider: %$SilverStripeVersionProvider
+SilverStripeVersionProvider:
+  modules:
+    silverstripe/framework: Framework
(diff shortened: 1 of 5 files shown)


robbieaverill:
Yeah, maybe I'll rewrite a little and use the caching to mock it

robbieaverill:
I'll move the "load composer lock contents" logic to another method then just mock that

robbieaverill:
@dhensby test added

PR #7127, opened 2017-07-04
Create a more generic travis installer

dhensby:
At the moment travis tests explicitly require a `.x-dev` branch to make sure we test the latest code. This means that when we add new branches and aliases we then need to update our travis configs - this is (hopefully) unnecessary.

This patch should mean less maintenance overhead for travis config

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,11 +57,13 @@ before_script:
 
 # Install composer dependencies
   - composer validate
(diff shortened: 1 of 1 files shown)


dhensby:
hmm - well, this doesn't quite work :(

dhensby:
not working :(

PR #7133, opened 2017-07-05
Cache ClassInfo::class_name() calls

kinglozzer:
A lot of these calls, especially in the admin area, are repeated. Memory impact is tiny as it’s just an array of string values.

Examples:

- `admin/pages` on a clean install:
  - saves ~18,000 `(new ReflectionClass($class)->getName())` calls
  - 14kb memory increase
- home page on a clean install:
  - saves ~1,800 `(new ReflectionClass($class)->getName())` calls
  - 6kb memory increase

### DIFF
--- a/src/Core/ClassInfo.php
+++ b/src/Core/ClassInfo.php
@@ -65,6 +65,14 @@ class ClassInfo
     private static $_cache_parse = [];
 
     /**
+     * Cache for class_name
+     *
+     * @internal
+     * @var array
+     */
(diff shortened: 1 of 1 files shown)


dhensby:
what are the disadvantages in using a persistent cache / the build in cache mechanisms?

tractorcow:
That would probably be less performant than using ReflectionClass.

dhensby:
I've restarted bad builds - merge on green

PR #7141, opened 2017-07-06
Fix OS X issue with some string methods.

### LINKED ISSUE #7132: BUG: String tokenizing methods return garbled content with some UTF-8 characters

bummzack:
Fix OS X issue with `Convert::html2raw`, `HTMLText::FirstSentence`, `HTMLText::Summary` and `Text::FirstSentence`.
Use unicode modifier for regular expressions that deal with whitespace.
Added unit-tests to ensure no invalud utf-8 gets generated by these methods.
Fixes #7132

### DIFF
--- a/tests/core/ConvertTest.php
+++ b/tests/core/ConvertTest.php
@@ -10,6 +10,24 @@ class ConvertTest extends SapphireTest {
 
 	protected $usesDatabase = false;
 
+	private $previousLocaleSetting = null;
+
+	public function setUp()
+	{
+		parent::setUp();
+		// clear the previous locale setting
+		$this->previousLocaleSetting = null;
+	}
+
+	public function tearDown()
+	{
+		parent::tearDown();
+		// If a test sets the locale, reset it on teardown
+		if ($this->previousLocaleSetting) {
+			setlocale(LC_CTYPE, $this->previousLocaleSetting);
+		}
+	}
+
 	/**
 	 * Tests {@link Convert::raw2att()}
 	 */
@@ -395,5 +413,29 @@ XML
 			$data,
 			Convert::base64url_decode(Convert::base64url_encode($data))
 		);
+	}
+
+	public function testValidUtf8()
+	{
+		// Install a UTF-8 locale
+		$this->previousLocaleSetting = setlocale(LC_CTYPE, 0);
+
+		$locales = array('en_US.UTF-8', 'en_NZ.UTF-8', 'de_DE.UTF-8');
+		$localeInstalled = false;
+		foreach ($locales as $locale) {
+			if ($localeInstalled = setlocale(LC_CTYPE, $locale)) {
+				break;
+			}
+		}
+
+		// If the system doesn't have any of the UTF-8 locales, exit early
+		if ($localeInstalled === false) {
+			$this->markTestIncomplete('Unable to run this test because of missing locale!');
+			return;
+		}
+
+		$problematicText = html_entity_decode('<p>This is a&nbsp;Test with non-breaking&nbsp;space!</p>', ENT_COMPAT, 'UTF-8');
+
+		$this->assertTrue(mb_check_encoding(Convert::html2raw($problematicText), 'UTF-8'));
 	}
 }
--- a/tests/model/HTMLTextTest.php
+++ b/tests/model/HTMLTextTest.php
@@ -5,6 +5,24 @@
  */
 class HTMLTextTest extends SapphireTest {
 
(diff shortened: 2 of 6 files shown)


tests/core/ConvertTest.php
```diff
@@ -396,4 +396,30 @@ public function testBase64url() {
...
+			$this->markTestIncomplete('Unable to run this test because of missing locale!');
+			return;
+		}
+
+		$problematicText = html_entity_decode('<p>This is a&nbsp;Test with non-breaking&nbsp;space!</p>', ENT_COMPAT, 'UTF-8');
+
+		$this->assertTrue(mb_check_encoding(Convert::html2raw($problematicText), 'UTF-8'));
+
+		// reset locale
+		setlocale(LC_ALL, $currentLocale);
```
tractorcow: This reset needs to be in tearDown(), not in the test itself. A failed test will not reset the locale correctly.

tests/model/HTMLTextTest.php
```diff
@@ -314,4 +314,34 @@ public function testShortCodeParsedInTemplateHelpers() {
...
+		$problematicText = html_entity_decode('<p>This is a&nbsp;Test with non-breaking&nbsp;space!</p>', ENT_COMPAT, 'UTF-8');
+
+		$textObj = new HTMLText('Test');
+		$textObj->setValue($problematicText);
+
+		$this->assertTrue(mb_check_encoding($textObj->FirstSentence(), 'UTF-8'));
+		$this->assertTrue(mb_check_encoding($textObj->Summary(), 'UTF-8'));
+
+		// reset locale
+		setlocale(LC_ALL, $currentLocale);
```
tractorcow: As above, please use tearDown() to reset.

tests/core/ConvertTest.php
tractorcow: Also please see http://php.net/manual/en/function.setlocale.php#106811; setlocale(LC_ALL, 0) can return multiple locales.

tractorcow:
Note : This PR will need to be merged up to 4 branch manually.

tests/core/ConvertTest.php
bummzack: @tractorcow I see. Should I implement the locale set/reset as outlined in that user-comment? Would it hurt to set/clear the locale for all tests in these classes? Eg. run `setlocale` in `setUp` and reset in `tearDown`?

bummzack: I've updated the tests with a tearDown implementation. Changed `LC_ALL` to `LC_CTYPE`, as this setting is enough for the test in question and prevents getting multiple locales from `setlocale`.

PR #7142, opened 2017-07-06
DOCS Add reference in performance -> profiling to SilverStripe DebugBar module

robbieaverill:
Add a short section referencing the [SilverStripe DebugBar](https://github.com/lekoala/silverstripe-debugbar) to the "Performance -> Profiling" section of the 3.x dev docs.

### DIFF
--- a/docs/en/02_Developer_Guides/08_Performance/03_Profiling.md
+++ b/docs/en/02_Developer_Guides/08_Performance/03_Profiling.md
@@ -3,11 +3,19 @@ summary: Identify bottlenecks within your application.
 
 # Profiling
 
-Profiling is the best way to identify bottle necks and other slow moving parts of your application prime for 
-optimization. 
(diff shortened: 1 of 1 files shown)


tractorcow:
I realise this module may be useful, but I'm not sure if we should adopt the practice of linking to helpful third party modules. Adding coupling to third party modules increases the maintenance burden of our own docs, and isn't really the goal of the documentation.

As an alternative, could we not link to addons with a search time like `http://addons.silverstripe.org/add-ons?search=logging+debug&type=module`?

robbieaverill:
@tractorcow good point re: doc rot. This module will become a "supported dependency" via our team - does that change anything?

robbieaverill:
We'll just put this into the CWP docs, since the intention is that it'll become part of the dev recipe which makes it more relevant there. Thanks @tractorcow!

PR #7149, opened 2017-07-10
[BUG] Fix issues with multiple authenticators for a single task

Firesphere:
Using multiple 2FA authenticators, logging out, resetting password etc. proved to be handled wrong.
Example scenario:
### Person logs out while multiple authenticators support logout
The result is an error, because the `renderWrappedController` was called, despite the responses being a set of either array with Content or Form, or a redirect action.

The default action should be followed and not try to render if there is nothing to render

Because the logout (or changepassword, or resetpassword, etc.) has already been handled, the first response is the default authenticator's response. This _could_ be a form (in case of logout without valid token), a content set (reset password) or a form (change password).

This edge case only happens when there are multiple authenticators supporting the requested method that is _not_ login.

### DIFF
--- a/src/Security/MemberAuthenticator/LogoutHandler.php
+++ b/src/Security/MemberAuthenticator/LogoutHandler.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\Security\MemberAuthenticator;
 
 use SilverStripe\Control\Director;
+use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\RequestHandler;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\ORM\ValidationResult;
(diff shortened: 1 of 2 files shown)


Firesphere:
Explanation:

My test case is based on the usage of @camfindlay his 2FA method side-by-side with my YubiAuth module.
What happens on logout, is the request is delegated to the aggregateTabbed, which will fail, because both modules will respond with an HTTPResponse.

The reason for failure, is that both Authenticators support logout (or reset password, or what else), and therefore, will respond with a valid HTTPResponse the Security::handleMultiple doesn't support.

This PR resolves the handle multiple, by returning the first successful response. This works, because the user at this point is already logged out, so there's no further need for confirmation. With the exception of a user with an expired logout token. In which case, the confirmation screen by @kinglozzer is shown.

Firesphere:
No new tests, as there is no new scenario to test. Happy to add if requested!

PR #7150, opened 2017-07-10
WIP: Only remove XDebug when not running coverage builds

robbieaverill:


robbieaverill:
Ok, this will need more thought

PR #7151, opened 2017-07-10
NEW Allow SSViewer to be injectable

robbieaverill:
Unless there's a good reason why SSViewer isn't injectable by default, how do we feel about this change? If this gets approved then I've got a couple of other module PRs to smash up afterwards.

### DIFF
--- a/src/Control/Controller.php
+++ b/src/Control/Controller.php
@@ -414,7 +414,7 @@ class Controller extends RequestHandler implements TemplateGlobalProvider
             $templates = array_unique(array_merge($actionTemplates, $classTemplates));
         }
 
-        return new SSViewer($templates);
+        return SSViewer::create($templates);
     }
 
     /**
--- a/src/Forms/Form.php
+++ b/src/Forms/Form.php
@@ -1582,7 +1582,7 @@ class Form extends ViewableData implements HasRequestHandler
      */
     public function forAjaxTemplate()
     {
-        $view = new SSViewer($this->getTemplates());
+        $view = SSViewer::create($this->getTemplates());
 
         $return = $view->dontRewriteHashlinks()->process($this);
 
@@ -1631,7 +1631,7 @@ class Form extends ViewableData implements HasRequestHandler
         ));
 
         if (is_string($template)) {
-            $template = new SSViewer($template);
+            $template = SSViewer::create($template);
         }
 
         return $template->process($custom);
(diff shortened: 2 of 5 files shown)


sminnee:
@robbieaverill what would the other module PRs be?

sminnee:
This change is okay, although as far as extensibility APIs go, we could potentially do better.

E.g. something like:

```php
interface Template
{
  /**
   * Populate this template with the given data and return the result as a string
   */
  function process($data);
}

// Remains as is expect for implementing the new interface
class SSViewer implements Template
{ ... }


interface TemplateFinder
{
  function getTemplateForIdentifierList($templates);
}
```

sminnee:
That interface pair is just a quick sketch but the main thing is having an object that represents a single template ready to be rendered, and then another which is a factory for getting the right template based on the identifier list. The latter one would be an injector service.

In principle you could use this to plug in support for other template engines, including having some kind of dispatch that checked what kind of template was used for the given identifier

robbieaverill:
Sounds like a good idea. I wish I had a use case to test this with!

sminnee:
Yeah – that's what I was going to ask. Were the aforementioned "other PRs" test use-cases?

robbieaverill:
Ah, ok - so the "other PRs" were the other core modules that use `new SSViewer` -> `SSViewer::create()`. The reason I'm suggesting this change is to allow us to add a proxy layer over SSViewer to monitor template usage etc.

sminnee:
...like a middleware added to the TemplateFinder? :-D

robbieaverill:
That sounds nice 😄

sminnee:
...it's hackday next Friday and that's not a breaking API. ;-)

tractorcow:
@robbieaverill I suggest making SSViewer_FromString injectable too.

```php
$viewer = new SSViewer_FromString($content);
```

sminnee:
I'd probably approach that like...

```
interface TemplateFinder
{
  function getTemplateForIdentifierList($templates);
  function getTemplateFromString($templateContent);
}
```

robbieaverill:
Would you both be OK with this PR making things injectable and another to add some interface structures?

sminnee:
Yeah I'm okay with that if other core team agree.

robbieaverill:
Cool - I'll change the title to WIP then add injection for the other SSViewer classes too

robbieaverill:
Ok, done

tractorcow:
@sminnee given we discussed the interface approach being another set of work, would it be ok to merge this? I've given in my blessing. :D

robbieaverill:
Thanks team - the other PRs are open (listed above)

sminnee:
Merged all 3.

PR #7153, opened 2017-07-10
Fix: Make File::ini2bytes() compliant with binary prefixes (fixes #7145)

### LINKED ISSUE #7145: File uploads failing with new ini2bytes logic
The  new ini2bytes logic introduced [here](https://github.com/silverstripe/silverstripe-framework/commit/f101697f8ef5dac427c7c3b65c457f5c6c1ab090) drops the final character when not specifying a unit.

`$unit = strtolower(array_pop($iniValues));` is assuming the last character in the array is a unit. This means code that was already in bytes, eg: `5242880` is now returned as `524288`

kinglozzer:
Lifted from Drupal core https://github.com/drupal/drupal/blob/8.4.x/core/lib/Drupal/Component/Utility/Bytes.php

### DIFF
--- a/filesystem/File.php
+++ b/filesystem/File.php
@@ -887,24 +887,21 @@ class File extends DataObject {
 	 * @return int
 	 */
 	public static function ini2bytes($iniValue) {
-		$iniValues = str_split(trim($iniValue));
-		$unit = strtolower(array_pop($iniValues));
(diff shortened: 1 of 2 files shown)


tractorcow:
I just realised I was referring to 4.x API. Ignore that.

Issue #7115, opened 2017-07-04
$project other than "" or "mysite" broken


sminnee:
There's currently no way to define a project() value other than mysite. This is because ThemeManifest is initialised in CoreKernel::__construct(), prior to bootConfigs() being called where $project could be defined.

If your project is called mysite, it still works, because `project()` defaults to returning mysite if that folder exists and the global isn't defined.

This could be fixed by adding the default set to ThemeResourceLoader after `bootConfigs()` in the boot.

**Note:** `project()`'s code that defaults to mysite hides this bug and should probably be removed If we wanted to remove the need for the global this value could probably be placed in yaml config instead. It would also highlight that the only use of this odd feature is to put 1 of the modules as top priority in the default ordering.

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/7154
- [x] https://github.com/silverstripe/silverstripe-admin/pull/148
- [x] https://github.com/silverstripe/silverstripe-assets/pull/45
- [x] https://github.com/silverstripe/silverstripe-cms/pull/1888

tractorcow:
Could we fix it by making `project` a config instead?

tractorcow:
Actually, why not just ditch project altogether and have a module priority order, like i18n does?

```yaml
---
Name: cmsi18n
Before: '/i18n'
After: '/i18n#basei18n'
---
SilverStripe\i18n\i18n:
  module_priority:
    - silverstripe\cms
```

sminnee:
> Actually, why not just ditch project altogether and have a module priority order, like i18n does?

Yeah something like that for SSViewer could work.

unclecheese:
Per my discussion with @tractorcow, we've agreed that the outcome of this should be to allow the user to set priority for template lookups, using a system similar to i18n. Further, we believe the i18n module sorting and the theme sorting should be one in the same.

## Tasks (TBD)
* Make `project` an `SSViewer` config setting
* Add `module_priority` to `SSViewer`, which can consume the the soft coded `project`setting:

```
SilverStripe\View\SSViewer:
  project: 'myapp'
  module_priority:
    - my-really-important-app
    - $project
    - blog
    - ecommerce
    - $other_modules
```

* Create a service based on `SilverStripe\i18n\Data\Sources::getSortedModules()` that can be composed to sort modules.
[...]

PR #7154, opened 2017-07-10
Pulls/4.0/project no more project

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-framework/issues/7115


_config/config.yml
```diff
@@ -6,9 +6,18 @@ SilverStripe\Core\Injector\Injector:
...
+    properties:
+      RestKey: 'other_modules'
 SilverStripe\Control\HTTP:
   cache_control:
     max-age: 0
     must-revalidate: "true"
     no-transform: "true"
   vary: "Cookie, X-Forwarded-Protocol, User-Agent, Accept"
+SilverStripe\Core\Manifest\ModuleManifest:
+  project: ''
```
tractorcow: We should set this to the current `mysite` default.

_config/config.yml
```diff
@@ -6,9 +6,18 @@ SilverStripe\Core\Injector\Injector:
...
+  module_priority:
+    - 'other_modules'
```
tractorcow: Why is $project lower in the priority than all other modules?

src/Core/Manifest/Module.php
```diff
@@ -127,9 +128,15 @@ public function unserialize($serialized)
...
     {
-        $config = "{$this->path}/_config.php";
-        if (file_exists($config)) {
-            require_once $config;
+        global $project;
+        $config = ModuleManifest::config();
+        $configPath = "{$this->path}/_config.php";
+        if (file_exists($configPath)) {
+            require_once $configPath;
+            if ($project !== $config->get('project')) {
```
tractorcow: Risk is that `global $project` is already different to the config project before the _config.php is included. I suggest to assign the global to a temp variable, include _config.php, and then compare back at that point.

tractorcow: I suggest to move this whole logic out of Module.php and put it in the Kernel code.


```php
/**
     * Include all _config.php files
     */
    protected function bootConfigs()
    {
    	global $project;
    	$projectBefore = $project;
        // After loading all other app manifests, include _config.php files
        $this->getModuleLoader()->getManifest()->activateConfig();
        if ($project && $project !== $projectBefore) {
        	// deprecation, etc
        }
    }
```

src/Core/Manifest/ModuleManifest.php
```diff
@@ -225,4 +238,31 @@ public function getModules()
...
+
+            $sorter = Injector::inst()->createWithArgs(
+                PrioritySorter::class.'.modulesorter',
+                [
+                    $this->modules,
+                    $order
+                ]
+            )
+            ->setVariable(self::PROJECT_KEY, $project)
+            ->setDefaultTop(self::PROJECT_KEY);
```
tractorcow: "top" should be controlled by config. See my earlier comment on ordering of $project. I don't think there should be a get/setDefaultTop at all.

I suggest to break out the YML config for ordering of modules as below:

file `framework/_config/modules.yml`

```yaml
---
name: modulepriorityproject
before: *
---
SilverStripe\Core\Manifest\ModuleManifest
  module_priority:
    - $project
---
name: modulepriorityother
---
SilverStripe\Core\Manifest\ModuleManifest
  module_priority:
    - $other_modules
```

We should probably use `$other_modules` for consistency with the `$projects` placeholder, since they are both placeholders.

src/Core/Manifest/PrioritySorter.php
```diff
@@ -0,0 +1,216 @@
+<?php
+
+namespace SilverStripe\Core\Manifest;
+
+use SilverStripe\Core\Injector\Injectable;
+
+/**
+ * Class PrioritySorter
```
tractorcow: Can you document the purpose of this class here and maybe give an example usage? we don't use @package anymore.

src/Core/Manifest/PrioritySorter.php
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * A map of variables to their values
+     * @var array
+     */
+    protected $variables = [];
+
+    /**
+     * @var array
+     */
+    protected $items;
```
tractorcow: A description of the format of these arrays helps with readability :) What is it a map from => to?

src/Core/Manifest/PrioritySorter.php
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    /**
+     * Sorts the items and returns a new version of $this->items
+     *
+     * @return array
+     */
+    public function getSortedList()
+    {
+        $this->addVariables();
```
tractorcow: This method is only ever called once, and it's protected. I wonder if it should just be inlined?

src/Core/Manifest/PrioritySorter.php
```diff
@@ -0,0 +1,216 @@
...
+
+        // Find all items that don't have their order specified by the config system
+        $unspecified = array_diff($this->names, $this->priorities);
+
+        if ($this->restKey && !empty($unspecified)) {
+            $this->includeRest($unspecified);
+        }
+
+        if ($this->defaultTop) {
```
tractorcow: Suggested earlier, I don't think we need a defaultTop if variable is included in priority in correct place.

src/Core/Manifest/PrioritySorter.php
```diff
@@ -0,0 +1,216 @@
...
+            $this->includeDefaultTop();
+        }
+
+        $sortedList = [];
+        foreach ($this->priorities as $itemName) {
+            if (isset($this->items[$itemName])) {
+                $sortedList[$itemName] = $this->items[$itemName];
+            }
+        }
+        $sortedList = array_reverse($sortedList, true);
```
tractorcow: I think we are double-reversing this list. If you put `$project` earlier in the config, you don't need to reverse the array here.

Also my testing is showing your $sortedList as 

```

[0] =>
--
  | string(22) "silverstripe/versioned"
  | [1] =>
  | string(23) "silverstripe/siteconfig"
  | [2] =>
  | string(22) "silverstripe/errorpage"
  | [3] =>
  | string(20) "silverstripe/reports"
  | [4] =>
  | string(19) "silverstripe/assets"
  | [5] =>
  | string(18) "silverstripe/admin"
  | [6] =>
  | string(6) "mysite"
  | [7] =>
  | string(20) "silverstripe/graphql"
  | [8] =>
  | string(22) "silverstripe/framework"
  | [9] =>
  | string(16) "silverstripe/cms"
  | [10] =>
  | string(27) "silverstripe/campaign-admin"
  | [11] =>
  | string(24) "silverstripe/asset-admin"

```

Not sure why mysite is ending up in the middle.

src/Dev/Deprecation.php
```diff
@@ -110,7 +110,10 @@ protected static function get_calling_module_from_trace($backtrace)
 
         $modules = ModuleLoader::inst()->getManifest()->getModules();
         foreach ($modules as $module) {
+            echo "checking $callingfile in {realpath($module->getPath())}\n";
```
tractorcow: debug code!

src/View/ThemeManifest.php
```diff
@@ -156,19 +164,17 @@ public function handleDirectory($basename, $pathname, $depth)
...
-        // If this is in the project, add to beginning of list. Else add to end.
-        if ($themeParts && $themeParts[0] == $this->project) {
-            array_unshift($this->themes, $path);
-        } else {
-            array_push($this->themes, $path);
-        }
+    /**
+     * Sets the project
+     *
+     * @param $project
```
tractorcow: phpdoc types too please.

src/View/ThemeManifest.php
```diff
@@ -156,19 +164,17 @@ public function handleDirectory($basename, $pathname, $depth)
...
+     */
+    public function setProject($project)
+    {
+        $this->project = $project;
```
tractorcow: `return $this;` on fluent setters

src/Core/Manifest/ModuleManifest.php
```diff
@@ -225,4 +238,31 @@ public function getModules()
     {
         return $this->modules;
     }
+
+    /**
+     * Get modules sorted by priority
+     *
+     * @return Module[]
+     */
+    public function getSortedModules()
```
tractorcow: Let's not have a getModules() and getSortedModules(); Just make getModules() return sorted list?

tractorcow: I suggest to do the sorting in ModuleManfiest::regenerate(); That way you can cache the sorting as well (yay performance).

src/Core/Manifest/PrioritySorter.php
tractorcow: result of missing $project default value.

src/Core/Manifest/ModuleManifest.php
unclecheese: If we remove `defaultTop`, then there is a scenario in which you can define `project`, and its effect would be meaningless. Currently what makes `project` special is that even if you don't specify its priority, it will ensure that it comes out on top. Maybe that's a good thing. I like the idea of being explicit about the ordering.

With regard to `$other_modules`, I very much wanted to do that, but I feared it would be met with resistance as an API change. Do you see it as one?

src/Core/Manifest/ModuleManifest.php
unclecheese: Per our discussion, `getModules()` is necessary as an option that is not `Config` dependent. This is vital during the boot process when `Config` may not be ready.

Moving this to `regenerate()` may circumvent the issue entirely. `getModules()` would return an unsorted list pre-boot and a sorted one post-boot.

src/Core/Manifest/Module.php
unclecheese: This is beautiful.

_config/config.yml
unclecheese: Discussed outside this thread, but we should figure out if top-to-bottom or bottom-to-top is more semantically correct.

_config/config.yml
unclecheese: Per @sminnee this will move to the core recipe, which makes a lot more sense. That way the module is guaranteed to exist.

src/Core/Manifest/PrioritySorter.php
unclecheese: Could be.. the method was getting quite long.

src/Core/Manifest/ModuleManifest.php
unclecheese: Things get really stuffed up when multiple configs specify `module_priority`, as the arrays merge, and you get something like:

```
[
  '$project',
  'blog',
  '$project',
  '$project',
  '$other_modules',
  'app',
  '$other_modules'
]
```

 I'm not sure there's a great way to solve this, short of making the priority a scalar value.

src/Core/Manifest/ModuleManifest.php
```diff
@@ -42,7 +48,14 @@ class ModuleManifest
...
      */
-    protected $modules = array();
+    protected $modules = [];
+
+    /**
+     * A cached list of sorted modules
+     *
+     * @var Module[]
+     */
+    protected $sortedModules;
```
kinglozzer: Looks like this can be removed now?

src/Core/CoreKernel.php
```diff
@@ -498,10 +506,15 @@ protected function bootManifests($flush)
                 $config->setFlush(true);
             }
         }
+        // tell modules to sort, now that config is available
+        $this->getModuleLoader()->getManifest()->sort();
```
kinglozzer: Looks like this will impact the order in which `_config.php` files are loaded when `bootConfigs()` is called. This will mean that `$project/_config.php` will be loaded earlier if it’s a higher priority (which it usually will be). Should this order be reversed for `_config.php` files? e.g. if I set my module to have a higher priority than `framework`, I can override templates but I can’t override the `embed` shortcode parser

unclecheese: Good point. We may need to update `ModuleManifest::activateConfig()` to work on a reversed array.

sminnee: >  Should this order be reversed for _config.php files? e.g. if I set my module to have a higher priority than framework, I can override templates but I can’t override the embed shortcode parser.

👍  from me just make sure you document it clearly.

docs/en/02_Developer_Guides/01_Templates/05_Template_Inheritance.md
```diff
@@ -43,20 +43,59 @@ The manifest is created whenever you flush your SilverStripe cache by appending
...
-4. modules  (e.g. blog)
-5. framework
-
 <div class="warning">
 Whenever you add or remove template files, rebuild the manifest by visiting `http://yoursite.com/?flush=1`. You can 
 flush the cache from any page, (.com/home?flush=1, .com/admin?flush=1, etc.). Flushing the cache can be slow, so you 
 only need to do it when you're developing new templates.
 </div>
 
+## Template Priority
```
chillu: @unclecheese I think you need to fix up the docs in a similar way in `i18n.md`. Can you please also check `Directory_Structure.md` and ensure those instructions still make sense?

Is there no need for mentioning this in the upgrade guide? Will template and i18n inheritance work like in 3.x by default?

docs/en/02_Developer_Guides/01_Templates/05_Template_Inheritance.md
```diff
@@ -43,20 +43,59 @@ The manifest is created whenever you flush your SilverStripe cache by appending
...
+### Defining a "project"
+
+It is a good idea to define one of your modules as the `project`. Commonly, this is the
+`mysite/` module, but there is nothing compulsory about that module name. The "project"
+module can be specified as a variable in the `module_priorities` list, as well. 
+
+*some-module/_config.yml*
+```yml
+SilverStripe\Core\Manifest\ModuleManifest:
+  project: 'myapp'
```
chillu: You shouldn't reference this as `mysite` in the intro, and then call it `myapp` here.

PR #7155, opened 2017-07-10
Remove unnecessary DataObject::write in fixture blueprint

robbieaverill:
This line doesn't seem necessary, and shaves ~1 min off overall build time without it.

### DIFF
--- a/src/Dev/FixtureBlueprint.php
+++ b/src/Dev/FixtureBlueprint.php
@@ -143,8 +143,6 @@ class FixtureBlueprint
                 }
             }
 
-            $obj->write();
-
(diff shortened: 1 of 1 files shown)


dhensby:
seems legit - would it be worth doing this in the 3.5 line?

robbieaverill:
I suppose so. I assume it won't be cherry pickable but feel free to copy it through

tractorcow:
How does this line work if you don't write it?

```php
$fixtures[$class][$identifier] = $obj->ID;
```

robbieaverill:
@tractorcow I guess it doesn't. I based this off the test suite passing. I'm happy to close it as well since it doesn't add that much benefit.

PR #7163, opened 2017-07-12
NEW add web accessible colours to web view dev/build

### LINKED ISSUE #7144: SS4: build info colour is hard to read
The colour contrast between the grey background and #5bc0de is making it fairly uneasy on the eyes to read. Using bootstrap's #31708f code is softer on the eyes

sachajudd:
Fixes #7144 After discussion with @clarkepaul we decided to change a few colours to meet WCAG AA colour contrast standards.

### DIFF
--- a/client/styles/debug.css
+++ b/client/styles/debug.css
@@ -4,7 +4,7 @@ body {
     margin: 0;
     overflow-x: hidden;
     padding: 0;
-    color: #66727d;
+    color: #4F5861;
     background: #f6f7f8;
     font-size: 15px;
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
@@ -31,6 +31,19 @@ ul {
     padding: 0 0 0 18px;
 }
 
+li {
+    padding: 2px 0;
+}
+
+li.info {
+    word-wrap: break-word;
+    padding: 0;
+}
+
+.build-info-before {
+    color: #66727d;
+}
+
 b {
     color: #4f5861;
     font-weight: bold;
@@ -49,7 +62,7 @@ a:active {
 .info {
     position: relative;
     z-index: 9999;
-    padding: 20px;
+    padding: 15px 15px 0;
 }
 
 .info h1 {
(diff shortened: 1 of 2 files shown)


sachajudd:
Hey @kinglozzer yes, this WIP will be adding a few more colours to improve accessibility. I'll let you know when I submit a new P.R, would be great to have your feedback 🙂

kinglozzer:
@sachajudd Great, I’ll hold off merging this for now - it’s probably easier to add more commits to this PR than open another one 😄

sachajudd:
@kinglozzer would you mind taking a look at this P.R again? 🙂  

old

new

client/styles/debug.css
```diff
@@ -49,7 +61,6 @@ a:active {
 .info {
     position: relative;
     z-index: 9999;
-    padding: 20px;
```
kinglozzer: I think we do need this padding - `debug.css` is also used for PHP errors, and without this padding the error message and stack trace touch the edges of the browser window, which looks a little odd:

screen shot 2017-07-17 at 09 24 18

client/styles/debug.css
```diff
@@ -71,14 +82,25 @@ a:active {
     border-bottom: solid 1px #004e7f;
     color: white;
     background: #005a93;
+    padding: 15px;
+}
+
+.header.info.error {
+    margin: 15px -15px 15px -33px;
```
kinglozzer: I think these margins will need tweaking, see the header in the screenshot above

kinglozzer: Ah, I see why this was added - to tidy up fatal errors during a `dev/build`. I wouldn’t worry about that scenario too much as we’re currently outputting another `<html><body><link rel="stylesheet"...` in the middle of the build log when that happens, so that will need to be addressed properly at some point 😃

dhensby:
@kinglozzer any more thoughts? :)

kinglozzer:
Nope, I just hadn’t seen that another commit had been added 😅. Looking great, thanks @sachajudd

Issue #6842, opened 2017-04-26
TreeDropdownField - React Search Feature


tractorcow:
## Acceptance Criteria

 * I can search for an item by partial title match
 * I can find items quickly regardless of their nesting level
 * I can determine when the search is still loading
 * I can select an item in the result
 * I can cancel a search in a user friendly way
 * I can get context of where in the hierarchy a search result is related
 * I can easily browse through dozens of search results
 * The component is keyboard accessible and has correct accessibility hints

## Tasks
- [x] store and retrieve the search term used
- [x] update `tree` field API to account for search, and showing in a flat list.
- [x] clear search when field no longer has focus
- [x] implement result item context display
- [x] Unit tests
- [x] Documentation
- [x] Design appropriate fields and add to CMS style guide (@clarkepaul)

## Design
https://projects.invisionapp.com/boards/D3X68PS46K28/

## Nice-to-have (?)

- [ ] Show full path for selected item (and remove Path field)

## Out of scope

 * Multiselect

## Notes

 * Broken out of https://github.com/silverstripe/silverstripe-framework/issues/6141
 * Design pending for this component.

## Related

 * [Use react-select for react-based dropdowns (SingleSelectField and MultiSelectField)](https://github.com/silverstripe/silverstripe-admin/issues/52)
 * Implement a TreeMultiSelectField in React

## Pull requests
- [x] https://github.com/silverstripe/silverstripe-admin/pull/150
- [x] https://github.com/silverstripe/silverstripe-framework/pull/7167
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/544

clarkepaul:
@flamerohr I'd like to discuss what the UI looks like with this one as I haven't supplied any design since a long time and it looks like the direction has changed a fair bit.

flamerohr:
@clarkepaul sounds good, I had wanted to touch base about the design soon too :)

chillu:
@flamerohr Just a note that I think find-as-you-type is out of scope here (not in ACs), because it's very likely too slow in the current implementation. Also, please doublecheck that you've implemented all of the ACs, the tasks don't reflect stuff like keyboard accessibility.

@clarkepaul Where do you see the "stats" counters being used? If they're counting recursively, they're going to be resource intensive to create. And I don't see the need for them, given we have a flattened search now.

flamerohr:
@chillu sorry, I'm not sure what you're referring to with "find-as-you-type"?
[...]

PR #7167, opened 2017-07-13
API TreeDropdown tree parameter

flamerohr:
Added flatList parameter for generating the json tree list into a flat list of options with a context string property instead of a nested hierarchy like it traditionally does.

This is to cater for searching and returning the results in a presentable way without the need to open each ancestor before getting to a result

Please see https://github.com/silverstripe/silverstripe-framework/issues/6842 for more details

### DIFF
--- a/src/Forms/TreeDropdownField.php
+++ b/src/Forms/TreeDropdownField.php
@@ -154,10 +154,19 @@ class TreeDropdownField extends FormField
 
     /**
      * List of ids in current search result (keys are ids, values are true)
+     * This includes parents of search result children which may not be an actual result
      *
      * @var array
      */
     protected $searchIds = [];
+    
+    /**
+     * List of ids which matches the search result
+     * This excludes parents of search result children
+     *
+     * @var array
+     */
+    protected $realSearchIds = [];
 
     /**
      * Determine if search should be shown
@@ -378,9 +387,15 @@ class TreeDropdownField extends FormField
         $isSubTree = false;
 
         $this->search = $request->requestVar('search');
+        $flatlist = $request->requestVar('flatList');
         $id = (is_numeric($request->latestParam('ID')))
             ? (int)$request->latestParam('ID')
             : (int)$request->requestVar('ID');
+    
+        // pre-process the tree - search needs to operate globally, not locally as marking filter does
+        if ($this->search) {
+            $this->populateIDs();
+        }
 
         /** @var DataObject|Hierarchy $obj */
         $obj = null;
@@ -400,11 +415,6 @@ class TreeDropdownField extends FormField
             if (!$this->baseID || !$obj) {
                 $obj = DataObject::singleton($this->sourceObject);
             }
-        }
-
-        // pre-process the tree - search needs to operate globally, not locally as marking filter does
-        if ($this->search) {
-            $this->populateIDs();
         }
 
         // Create marking set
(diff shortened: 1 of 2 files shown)


src/Forms/TreeDropdownField.php
```diff
@@ -555,6 +570,27 @@ public function getSourceObject()
         return $this->sourceObject;
     }
 
+    protected function flattenChildrenArray($children, $parentTitles = [])
```
tractorcow: Can you document this?

Also, it's not apparent that this method filters by realSearchIds.

src/Forms/TreeDropdownField.php
```diff
@@ -555,6 +570,27 @@ public function getSourceObject()
...
+    
+        foreach ($children as $child) {
+            $childTitles = array_merge($parentTitles, [$child['title']]);
+            $grandChildren = $child['children'];
+            $contextString = implode('/', $parentTitles);
+            
+            $child['contextString'] = ($contextString !== '') ? $contextString .'/' : '';
+            $child['children'] = [];
+    
+            if (in_array($child['id'], $this->realSearchIds)) {
```
tractorcow: Probably should also include this child if `!$this->search || in_array($child['id'], $this->realSearchIds)`, otherwise output will be empty.

src/Forms/TreeDropdownField.php
```diff
@@ -563,6 +599,44 @@ public function getSourceObject()
     protected function populateIDs()
     {
         // get all the leaves to be displayed
+        $res = $this->getSearchResults();
+        
+        if ($res) {
```
tractorcow: can you flip this

```php
if (!$res) {
	return;
}
```

src/Forms/TreeDropdownField.php
```diff
@@ -563,6 +599,44 @@ public function getSourceObject()
...
+            $this->realSearchIds = $res->column();
```
tractorcow: Move this to below the foreach() to save an extra query.


```php
foreach ($res as $row) {}
$this->realSearchIds =  $this->searchIds;
```

src/Forms/TreeDropdownField.php
```diff
@@ -563,6 +599,44 @@ public function getSourceObject()
...
+    }
+    
+    /**
+     * Get the DataObjects that matches the searched parameter.
+     *
+     * @return mixed
+     */
+    protected function getSearchResults()
+    {
         if ($this->searchCallback) {
```
tractorcow: You can reduce nesting by not using a temp variable.

```php
if ($this->searchCallback) {
	return call_user_func($this->searchCallback, $this->sourceObject, $this->labelField, $this->search);
}

$sourceObject = $this->sourceObject;
```

src/Forms/TreeDropdownField.php
```diff
@@ -563,6 +599,44 @@ public function getSourceObject()
...
+                    $this->searchExpanded[$item->ID] = true;
+                }
+            }
+        }
+    }
+    
+    /**
+     * Get the DataObjects that matches the searched parameter.
+     *
+     * @return mixed
```
tractorcow: Better to use `DataObject[]` or `SS_List` as a typehint for return.

src/Forms/TreeDropdownField.php
```diff
@@ -455,6 +465,11 @@ public function tree(HTTPRequest $request)
             // Format JSON output
             $json = $markingSet
                 ->getChildrenAsArray($customised);
+            
+            if ($flatlist) {
```
tractorcow: Can you add a test in TreeDropdownTest with flatlist?

flamerohr:
thanks @tractorcow and @phalkunz

I've added unit tests :)

PR #7169, opened 2017-07-13
Upgrade the upgrading docs

unclecheese:
Addresses https://github.com/silverstripe/silverstripe-framework/issues/6441

### DIFF
--- a/docs/en/04_Changelogs/4.0.0.md
+++ b/docs/en/04_Changelogs/4.0.0.md
@@ -152,15 +152,15 @@ public function countDuplicates($model, $fieldToCheck)
 
 ##### Class name remapping
 
-If you've namespaced one of your custom page types then loaded up the CMS and see 
-a message telling you it's obsolete. This is likely because the `ClassName` 
+If you've namespaced one of your custom page types, you may notice a message in the CMS 
+ telling you it's obsolete. This is likely because the `ClassName` 
 field in the `SiteTree` table still contains the singular model name, e.g. `BlogPost`
 and that when you change it to `SilverStripe\Blog\Model\BlogPost` then everything 
 works again.
 
 Luckily the `dev/build` task is configured to look for a legacy class name mapping 
 configuration setting and will update this for you automatically. For an example 
-take a look at `_config/legacy.yml` in the CMS module. For the Blog module we used 
+take a look at `_config/legacy.yml` in the CMS module. The Blog module, for instance, uses 
 this:
 
 ```yaml
@@ -221,9 +221,9 @@ A better option is to replace it with `DataObject::getSchema()->tableName(BlogPo
 
 The convention for naming controllers is now `[MyPageType]Controller`, where it used to be `[MyPageType]_Controller`. This change was made to be more compatible with the PSR-2 standards.
 
-You can still use, for example, `Page_Controller`, but you will get a deprecation notice. Best to change it to `PageController` during your upgrade process.
+You can still use, for example, `Page_Controller`, but you will get a deprecation notice. It is best to change it to `PageController` during your upgrade process.
 
-By default, a controller for a page type *must* reside in the same namespace as its page. To break this convention, override `SiteTree::getControllerName()`.
+By default, a controller for a page type *must* reside in the same namespace as its page. To use different logic, override `SiteTree::getControllerName()`.
 
 #### Upgrade template locations and references
 
@@ -244,8 +244,8 @@ the `SilverStripe\Forms` namespace.
 
 #### Upgrade your member statics
 
-If you've got some class configuration statics defined and they aren't private, 
-you may find that they don't register any more. For example, this code, taken from 
+If you have some class configuration statics defined and they aren't private, 
+you may find that they don't register anymore. For example, this code, taken from 
 the `silverstripe/tagfield` module will no longer work in SilverStripe 4.0.
 ```php
 public static $allowed_actions = [
@@ -272,7 +272,7 @@ private static $allowed_actions = [
 While not critical to an upgrade, SilverStripe 4.0 has adopted the [PS-4 autoloading](http://www.php-fig.org/psr/psr-4/)
 standard for the core modules, so it's probably a good idea to be consistent. 
 
-You can implement this into your composer configuration like so:
+You can implement this in your composer configuration like so:
 
 ```js
     ...
@@ -293,14 +293,15 @@ have also started to use `src/` as a default folder location instead of
 `code/`. If you’re going to change, it's probably a good time to do it while you're
 upgrading.
 
-For examples, take a look at the file/folder structure compared to the 
+For examples, take a look at the file/folder structure and to the 
 `composer.json` configuration in either the `framework` or `cms` modules.
 
 Please note that there are changes to template structure which in some cases 
-require templates to be in a folder location that matches the class's namespace 
+require templates to be in a folder location that matches the namespace of the class 
 that it belongs to, e.g. `themes/mytheme/templates/MyVendor/Foobar/Model/MyModel.ss`.
 
 #### Get on board with PSR-# logging, too
+
 One of the great changes that comes with SilverStripe 4 is the introduction of 
 [PSR-3](http://www.php-fig.org/psr/psr-3/) compatible logger interfaces. This 
 means we can use thirdparty services like Monolog.
@@ -322,7 +323,7 @@ in this case!
 
 ##### Unit tests and PSR-3 loggers
 
-If you've got some code that you've tested originally using `SS_Log`, and now moved 
+If you have some code that you've tested originally using `SS_Log`, and now moved 
 to using a Monolog Logger then you may notice occasions where your unit tests pass 
 but output a bunch of warnings, debug information, etc. to `stderr`. This is because
  Monolog will use `stderr` as the default Handler unless you specify one.
@@ -1317,10 +1318,45 @@ These methods are deprecated:
 #### Implementation of ownership API
 
 In order to support the recursive publishing of dataobjects, a new API has been developed to allow
-developers to declare dependencies between objects. See the
-[versioned documentation](/developer_guides/model/versioning) for more information.
+developers to declare dependencies between objects. This is done to ensure that the published state
+of linked components are consistent with their "owner." Without the concept of ownership, these linked
+components could be implicitly exposed on the frontend, which may not align with the intent of the
+content author.
 
-By default all versioned dataobjects will automatically publish objects that they own.
+For instance, on a products page which has a list of products, the products should not be published or
+ unpublished unless the products page is, too. The ownership API solves this by allowing you to declare
+ a two-way relationship between objects, typically, but not necessarily, linked by a database relationship
+ (`has_many`, `many_many`, etc.).
+
+```php
+class ProductPage extends Page {
+    private static $has_many = [
+        'Products' => Product::class
+    ];
+    
+    private static $owns = [
+        'Products'
+    ];
+}
+
+class Product extends DataObject {
+    private static $extensions = [
+        Versioned::class
+    ];
+    
+    private static $has_one = [
+        'Parent' => ProductPage::class
+    ];
+}
+```
+If your objects are linked by something other than a database relationship, for instance, a custom
+getter that is computed at runtime, the same rules can be applied, as long as you provide an `$owned_by`
+setting on the child object.
+
+For more information, see:
+
+* The [DataObject ownership](https://docs.silverstripe.org/en/4/developer_guides/model/versioning/#dataobject-ownership) documentation
+* The [versioned](/developer_guides/model/versioning) documentation
 
 #### ChangeSet batch publishing
 
--- a/docs/en/04_Changelogs/4.0.0.md
+++ b/docs/en/04_Changelogs/4.0.0.md
@@ -201,6 +201,7 @@ the `SilverStripe\Forms` namespace.
 If you have some class configuration statics defined and they aren't private, 
 you may find that they don't register anymore. For example, this code, taken from 
 the `silverstripe/tagfield` module will no longer work in SilverStripe 4.0.
+
 ```php
 public static $allowed_actions = [
     'suggest'
@@ -210,13 +211,9 @@ public static $allowed_actions = [
 Changing the visibility to `private` (as per `RequestHandler::$allowed_actions` 
 visibility) will make it 4.0 compatible.
 
-```ss
-<!-- before -->
-public static $allowed_actions = [
-    'suggest'
-];
-<!-- after -->
-private static $allowed_actions = [
+```diff
+-public static $allowed_actions = [
++private static $allowed_actions = [
     'suggest'
 ];
 ```
--- a/docs/en/04_Changelogs/4.0.0.md
+++ b/docs/en/04_Changelogs/4.0.0.md
@@ -59,6 +59,7 @@ guide developers in preparing existing 3.x code for compatibility with 4.0
   [behat extension](https://github.com/silverstripe/silverstripe-behat-extension) for more information.
 * The `GDBackend` and `ImagickBackend` classes have been replaced by a unified `InterventionBackend` which uses the
   [intervention/image](https://github.com/intervention/image) library to power manipualations.
+* Page types and their controllers can no longer be collated in a single file
 
 ## <a name="upgrading"></a>Upgrading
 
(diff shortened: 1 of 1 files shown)


docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +93,138 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
 
 The rename won't affect class-based permission codes or database table names.
 
+##### Use ::class whenever possible
+
+Prefer:
+ ```php
```
chillu: Can you please use `diff` format here? It's more succinct, easier to read.

unclecheese: I avoided that because it's not old/new .. it's a recommended pattern. I didn't want to confuse the reader into thinking one way was deprecated or wrong.

tractorcow: I think the way it's written better supports separate blocks.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +93,138 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+A better option is to replace it with `DataObject::getSchema()->tableName(BlogPost::class)` as long as the class has already been imported into your current namespace.
+
+
+
+
+#### Migrate your controllers to their own files
+
+The convention for naming controllers is now `[MyPageType]Controller`, where it used to be `[MyPageType]_Controller`. This change was made to be more compatible with the PSR-2 standards.
+
+You can still use, for example, `Page_Controller`, but you will get a deprecation notice. It is best to change it to `PageController` during your upgrade process.
```
robbieaverill: This is true, but any module depending on the page controller will look for `PageController`

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -59,6 +59,7 @@ guide developers in preparing existing 3.x code for compatibility with 4.0
   [behat extension](https://github.com/silverstripe/silverstripe-behat-extension) for more information.
 * The `GDBackend` and `ImagickBackend` classes have been replaced by a unified `InterventionBackend` which uses the
   [intervention/image](https://github.com/intervention/image) library to power manipualations.
+* Page types and their controllers can no longer be collated in a single file
```
tractorcow: Well, they can be, they just shouldn't.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +93,138 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
 
 The rename won't affect class-based permission codes or database table names.
 
+##### Use ::class whenever possible
```
tractorcow: These upgrade steps are very low-priority. I suggest you put higher-priority important upgrade tasks at the top, and more "style guide recommendations" near the bottom of the list.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -109,6 +242,107 @@ Core template locations have moved - if you're including or overriding these
...
+
+Note that the old `SS_Log` class has been renamed to `SilverStripe\Logging\Log`, 
+but has also had most of its logic stripped out and the remainder is marked as 
+deprecated.
+
+```ss
+<!-- before -->
+SS_Log::log('My error message', SS_Log::ERR);
+<!-- after -->
+Injector::inst()->get('Logger')->error('My error message');
```
robbieaverill: This needs to be `'Psr\Log\LoggerInterface'` or `LoggerInterface::class` with an import now

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +93,138 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+
+##### Class name remapping
+
+If you've namespaced one of your custom page types, you may notice a message in the CMS 
+ telling you it's obsolete. This is likely because the `ClassName` 
+field in the `SiteTree` table still contains the singular model name, e.g. `BlogPost`
+and that when you change it to `SilverStripe\Blog\Model\BlogPost` then everything 
+works again.
+
+Luckily the `dev/build` task is configured to look for a legacy class name mapping
```
tractorcow: Not sure this is the best writing style for a technical upgrade manual. Nothing is "luckily" there, is it?

We also need to clarify the difference between class names changed in code (requires upgrader) and class names in database (this documentation block). The title "Class name remapping" doesn't explain that this is specific to database DBClassname values.

You may also need to clarify the distinction between user tasks and the explanation of the already implemented behaviour; This seems like the user is expected to write an upgrade yml, when in reality all they need to do is "run dev/build". The point of this block should focus on the "what users should do" rather than explaining the user story we took te implementing this feature.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +93,138 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+```yaml
+SilverStripe\ORM\DatabaseAdmin:
+  classname_value_remapping:
+    Blog: SilverStripe\Blog\Model\Blog
+    BlogCategory: SilverStripe\Blog\Model\BlogCategory
+    BlogPost: SilverStripe\Blog\Model\BlogPost
+    BlogTag: SilverStripe\Blog\Model\BlogTag
+```
+
+##### Other namespacing gotchas
```
tractorcow: Can you name these title more specifically? People aren't going to search for "gotchas" and it doesn't explain the purpose of the guide.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -109,6 +242,107 @@ Core template locations have moved - if you're including or overriding these
 no longer exists, and instead template locations will be placed in paths that match
 the `SilverStripe\Forms` namespace.
 
+#### Upgrade your member statics
```
tractorcow: `member` implies `Member` dataobject; Misleading title.

How about `Upgrade class config static properties to private`?

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -109,6 +242,107 @@ Core template locations have moved - if you're including or overriding these
...
+public static $allowed_actions = [
+    'suggest'
+];
+```
+
+Changing the visibility to `private` (as per `RequestHandler::$allowed_actions` 
+visibility) will make it 4.0 compatible.
+
+```ss
+<!-- before -->
```
tractorcow: `diff` style here please.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +93,138 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+```
+
+##### Use Injector wherever you can
+
+This can help with reducing changes from 3.x to 4.x by allowing your code to 
+continue to use `MyClassName`, while you tell the injector what the FQ class name is:
+
+```yaml
+SilverStripe\Core\Injector\Injector:
+  MyClassName:
```
tractorcow: This seems to be more of a "how to namespace your code" task than a "upgrade task". Upgrading to ss4 doesn't automatically namespace your code, nor is it required.

A better example is injector usage to construct DB Field types, eg. 


```diff
- $field = new Varchar();
+ $field = Injector::inst()->create('Varchar');
```

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +93,138 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+    class: Me\MyModule\MyClassName # fully qualified!
+```
+
+##### Namespaced table names
+
+Keep in mind that table names are namespaced too. To ease data migrations, use the `$table_name` config property of your `DataObject` subclasses.
```
tractorcow: Again this isn't "upgrading to silverstripe 4" this is "how to namespace your code"

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +93,138 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+##### Other namespacing gotchas
+
+You'll need to update any strings that represent class names and make sure they're fully
+qualified. In particular, relationship definitions such as `has_one` and `has_many` will need
+to be updated to refer to fully qualified class names.
+
+```ss
+<!-- before -->
+private static $has_one = [
+  'MyRelation' => 'MyObject',
```
tractorcow: Upgrader does a lot of this automatically. I strongly suggest to put this into a "how to use upgrader to automatically upgrade your code" block.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +93,138 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+names expressed literally. The upgrader will automatically replace anything that looks
+like an unqualified class name with a fully qualified one, but a fully qualified
+class name may not be the same as the table name, as explained above, in the case of
+`$table_name`.
+
+For example:
+```php
+->innerJoin('BlogPost', sprintf('"BlogPost%s"."ID" = "SiteTree%s"."ID"', $stage, $stage))
+```
+The upgrader will replace `BlogPost` with `SilverStripe\\Blog\\Model\BlogPost` as the fully qualified class name. In the case of innerJoin(), it asks for a table name which have remained the same as SilverStripe 3.x, so this should not be namespaced. This of course depends on what you’ve configured your table name to be.
```
tractorcow: It'll replace `'BlogPost'` with `BlogPost::class` and adding `use ...` actually, but not in substrings. You'll need to make this clear.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +93,138 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+
+A better option is to replace it with `DataObject::getSchema()->tableName(BlogPost::class)` as long as the class has already been imported into your current namespace.
+
+
+
+
+#### Migrate your controllers to their own files
```
tractorcow: You don't mention anything about moving them to their own files in this section though.

I would recommend to note that this is not mandatory, but is recommended.

We should group up "Upgrading your code to psr-2 standard" sections together maybe?

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -109,6 +242,107 @@ Core template locations have moved - if you're including or overriding these
...
+ Monolog will use `stderr` as the default Handler unless you specify one.
+
+This isn't so much a regression as simply that you're now seeing things happening 
+in your code that didn't used to happen. You have a couple of options here. You can either:
+
+* Address that there's possibly a logic problem in your unit tests
+* Push a conveniently named `NullHandler` to monolog to quiet it.
+
+```php
+$logger = Injector::inst()->get('Logger');
```
tractorcow: 'Logger' service is out of date; Please use `Psr\Log\LoggerInterface::class` instead

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -109,6 +242,107 @@ Core template locations have moved - if you're including or overriding these
...
+Injector::inst()->get('Logger')->error('My error message');
+```
+
+If you call that multiple times in one class, you might want to add a getter 
+method to centralize it a little - but you probably should've had that already 
+in this case!
+
+##### Unit tests and PSR-3 loggers
+
+If you have some code that you've tested originally using `SS_Log`, and now moved
```
tractorcow: Have many people actually been having this issue? Maybe we should provide a more generic "how to customise logger" guide rather than "how to suppress logging output in this one very specific situation".

docs/en/04_Changelogs/4.0.0.md
chillu: I take your point on this particular example, but there's lot of others further down with a clearer before/after relationship. Also, "prefer/instead" feels like the wrong way around for code examples. If you compare it to `.env`, we've first documented the old way, then the new way.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +93,137 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+```
+
+##### Other namespacing gotchas
+
+You'll need to update any strings that represent class names and make sure they're fully
+qualified. In particular, relationship definitions such as `has_one` and `has_many` will need
+to be updated to refer to fully qualified class names.
+
+```ss
+<!-- before -->
```
chillu: Nitpicking, but this isn't HTML - you should PHP commenting syntax.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +93,137 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+];
+<!-- after -->
+private static $has_one = [
+  'MyRelation' => MyObject::class,
+];
+```
+ 
+In the context of YAML, the magic constant `::class` does not apply. Class names must be hard coded.
+
+```ss
```
chillu: Same here, you should use YAML commenting syntax

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +93,137 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+
+For example:
+```php
+->innerJoin('BlogPost', sprintf('"BlogPost%s"."ID" = "SiteTree%s"."ID"', $stage, $stage))
+```
+The upgrader will replace `BlogPost` with `SilverStripe\\Blog\\Model\BlogPost` as the fully qualified class name. In the case of innerJoin(), it asks for a table name which have remained the same as SilverStripe 3.x, so this should not be namespaced. This of course depends on what you’ve configured your table name to be.
+
+A better option is to replace it with `DataObject::getSchema()->tableName(BlogPost::class)` as long as the class has already been imported into your current namespace.
+
+##### Namespacing tip: use ::class whenever possible
```
chillu: Make `Use ...` uppercase (proper sentence case): Hard to read this way

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +93,137 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+ ```
+ to
+ 
+ ```php
+private static $has_one = [
+  'MyObject' => 'My\Project\MyObject'
+];
+```
+
+##### Namespacing tip: use Injector whenever possible
```
chillu: Same here, sentence case

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -109,6 +241,107 @@ Core template locations have moved - if you're including or overriding these
...
+];
+<!-- after -->
+private static $allowed_actions = [
+    'suggest'
+];
+```
+
+#### Get on board with PSR-4 autoloading
+
+While not critical to an upgrade, SilverStripe 4.0 has adopted the [PS-4 autoloading](http://www.php-fig.org/psr/psr-4/)
```
chillu: Missing an "R"

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -109,6 +241,107 @@ Core template locations have moved - if you're including or overriding these
...
+upgrading.
+
+For examples, take a look at the file/folder structure and to the 
+`composer.json` configuration in either the `framework` or `cms` modules.
+
+Please note that there are changes to template structure which in some cases 
+require templates to be in a folder location that matches the namespace of the class 
+that it belongs to, e.g. `themes/mytheme/templates/MyVendor/Foobar/Model/MyModel.ss`.
+
+#### Get on board with PSR-# logging, too
```
chillu: You mean PSR-3?

docs/en/04_Changelogs/4.0.0.md
chillu: Yeah that seems like a bit too much noise so early on in the upgrading guide

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1105,10 +1255,45 @@ These methods are deprecated:
...
 In order to support the recursive publishing of dataobjects, a new API has been developed to allow
-developers to declare dependencies between objects. See the
-[versioned documentation](/developer_guides/model/versioning) for more information.
+developers to declare dependencies between objects. This is done to ensure that the published state
+of linked components are consistent with their "owner." Without the concept of ownership, these linked
+components could be implicitly exposed on the frontend, which may not align with the intent of the
+content author.
 
-By default all versioned dataobjects will automatically publish objects that they own.
+For instance, on a products page which has a list of products, the products should not be published or
```
chillu: "published **or unpublished**" is incorrect. We're still discussing how cascading unpublish would work, so the ownership API doesn't "solve this" at the moment.

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1105,10 +1255,45 @@ These methods are deprecated:
...
+    private static $has_one = [
+        'Parent' => ProductPage::class
+    ];
+}
+```
+If your objects are linked by something other than a database relationship, for instance, a custom
+getter that is computed at runtime, the same rules can be applied, as long as you provide an `$owned_by`
+setting on the child object.
+
+For more information, see:
```
chillu: Please be careful with excessive whitespace (like two-element lists) - I'd suggest that's inlined. Too hard to skim read the guide otherwise

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1335,6 +1517,158 @@ SilverStripe\Core\Injector\Injector:
...
+
+#### Upgrading user-code to use namespaces
+
+Upgrading code to use namespaces is quite a complex process, and as such we have provided
+several development tools and processes to help make the upgrade user friendly and as
+automated as possible.
+
+##### Using the upgrader tool to automatically apply namespaces
+
+HThe [upgrader tool](https://github.com/silverstripe/silverstripe-upgrader/) provides a feature
```
chillu: Spelling

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -18,18 +19,21 @@ guide developers in preparing existing 3.x code for compatibility with 4.0
...
+  * [Security](#overview-security)
 
 ## <a name="overview"></a>Highlights of major changes
 
 * Minimum version dependencies have increased; PHP 5.5 and Internet Explorer 10 (or other modern browser)
   is required.
 * All code earlier marked as deprecated for 4.0 has now been removed. See
   [deprecation documentation](/contributing/release_process) for information on code deprecation.
-* All SilverStripe classes are now namespaced, and some have been renamed. This has major implications for
+* All code has been migrated to follow the PSR-2 coding standard. most significantly, all SilverStripe
```
robbieaverill: Capital M on most

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -18,18 +19,21 @@ guide developers in preparing existing 3.x code for compatibility with 4.0
...
+ classes are now namespaced, and some have been renamed. This has major implications for
   arrangement of templates, as well as other references to classes via string literals or configuration.
   Automatic upgrading tools have been developed to cope with the bulk of these changes (see
-  [upgrading notes](#upgrading)).
+  [upgrading notes](#upgrading)). Each class has been arrange in a single file per class, and it is recommended
```
robbieaverill: arranged* in/into*(?)

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +103,82 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
 
 The rename won't affect class-based permission codes or database table names.
 
+##### Upgrade references to literal or class table names
+
+In 3.x the class name of any dataobject matched the table name, but in 4.x all classes are namespaced, and it is
```
robbieaverill: DataObject

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +103,82 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+}
+```
+
+##### Replacing literal strings with fully qualified class name, or ::class constant
+
+You'll need to update any strings that represent class names and make sure they're fully
+qualified. In particular, relationship definitions such as `has_one` and `has_many` will need
+to be updated to refer to fully qualified class names.
+
+In configs and with literal php strings it is recommended to use the php `::class` constant,
```
robbieaverill: PHP

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +103,82 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+```diff
+<?php
++ use SilverStripe\ORM\DataObject;
++ use SilverStripe\Security\Member;
+class MyClass extends DataObject {
+    private static $has_one = [
+    -  'Author' => 'Member',
+    +  'Author' => Member::class,
+    ];
+}
```
robbieaverill: Should follow PSR-2

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +103,82 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+```
+ 
+In the context of YAML, the magic constant `::class` does not apply. Class names must be hard coded.
```
robbieaverill: Maybe this could say "Fully qualified class names must be hard coded"?

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +103,82 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+```diff
+-MyObject:
++My\Project\MyObject:
+  property: value
+```
+
+##### Update references to literal classes to use Injector
+
+In many places certain classes namespaced in SilverStripe 4.x will still have a non-namespaced service
+name reference that can be referred to with injector. You should upgrade direct object constructors with
```
robbieaverill: That's a hell of a sentence :)

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -92,6 +103,82 @@ For a full list of renamed classes, check the `.upgrade.yml` definitions in each
...
+Injector api.
```
robbieaverill: API

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -629,7 +812,7 @@ this task is run manually during an explicit migration process, as this process
 large amounts of memory and run for an extended time.
 
 
-```yaml
+```yml
```
robbieaverill: FYI in GFM they're both valid

docs/en/04_Changelogs/4.0.0.md
```diff
@@ -850,14 +1033,14 @@ use SilverStripe\ORM\FieldType\DBComposite;
 
 class MyAddressField extends DBComposite
 {
-	private static $composite_db = ]
+	private static $composite_db = [
```
robbieaverill: lol

docs/en/04_Changelogs/4.0.0.md
tractorcow: I was trying to fix some markdown rendering issues; I didn't. =(

PR #7170, opened 2017-07-13
FIX Ignore exceptions thrown when deleting test databases

### LINKED ISSUE #7164: [4.x] Code coverage builds die with "MySQL server has gone away"
Seemingly sporadically if I run a module through Travis CI with a code coverage run enabled, it will get to generating the coverage report, then after a while it will die with MySQL server has gone away.

The following stack trace indicates that this is probably being caused by `TempDatabase::build()` which sets a shutdown handler to kill the temporary database:

```php
// Ensure test db is killed on exit
register_shutdown_function(function () {
    $this->kill();
});
```

Stack trace:

```
Generating code coverage report in Clover XML format ... done
PHP Fatal error:  Uncaught PDO::query(): MySQL server has gone away
/home/travis/build/lekoala/silverstripe-debugbar/framework/src/ORM/Connect/PDOConnector.php:260
/home/travis/build/lekoala/silverstripe-debugbar/framework/src/ORM/Connect/Database.php:142
/home/travis/build/lekoala/silverstripe-debugbar/framework/src/ORM/Connect/Database.php:222
/home/travis/build/lekoala/silverstripe-debugbar/framework/src/ORM/Connect/Database.php:143
/home/travis/build/lekoala/silverstripe-debugbar/framework/src/ORM/Connect/DBSchemaManager.php:104
/home/travis/build/lekoala/silverstripe-debugbar/framework/src/ORM/Connect/MySQLSchemaManager.php:211
/home/travis/build/lekoala/silverstripe-debugbar/framework/src/ORM/Connect/Database.php:763
/home/travis/build/lekoala/silverstripe-debugbar/framework/src/ORM/Connect/TempDatabase.php:82
/home/travis/build/lekoala/silverstripe-debugbar/framework/src/ORM/Connect/TempDatabase.php:150 <------------
```

I assume that the wait time is too great and MySQL disconnects before the temp DB can be killed.

There are some configuration options for mysqli and methods like `mysqli_ping`, but not for PDO.

robbieaverill:
This will prevent long runnings builds (e.g. code coverage) from failing when the test database connection is gone (MySQL server has gone away) by the time the shutdown handler runs.

---

I've made an assumption here that an exception thrown at this point (after the test suite has finished running and we're in tidy up mode) should be ignored instead of causing the build to fail, since this has nothing to do with the actual test suite results.

This approach was cleaner than adding in some conditional logic to check for "MySQL server has gone away" exception messages and trying to reconnect, then kill again. If the team thinks it would be preferable to do that I've got an alternative branch that does that as well, but I suspect this approach is better in the long run.

Resolves #7164

### DIFF
--- a/src/ORM/Connect/TempDatabase.php
+++ b/src/ORM/Connect/TempDatabase.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\ORM\Connect;
 
+use Exception;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\Core\Injector\Injector;
@@ -92,7 +93,6 @@ class TempDatabase
             }
         }
 
-        // echo "Deleted temp database " . $dbConn->currentDatabase() . "\n";
         $dbConn->dropSelectedDatabase();
     }
 
@@ -147,7 +147,11 @@ class TempDatabase
 
         // Ensure test db is killed on exit
         register_shutdown_function(function () {
-            $this->kill();
+            try {
+                $this->kill();
+            } catch (Exception $ex) {
+                // An exception thrown while trying to remove a test database shouldn't fail a build, ignore
+            }
         });
 
         return $dbname;


dhensby:
thanks

PR #7171, opened 2017-07-13
Fix SearchEngine to use quoted table names

Martimiz:
This fixes MySQLDatabase::searchEngine(). Tablenames were not quoted and this meant that SQLExpression::sql() couldn't replace them depending on the actual stage. 

So even when Stage = Live, the query would still be performed on SiteTree, although other modules (Translatable) would  add filters based on SiteTree_live, resulting in errors.

### DIFF
--- a/src/ORM/Connect/MySQLDatabase.php
+++ b/src/ORM/Connect/MySQLDatabase.php
@@ -226,10 +226,10 @@ class MySQLDatabase extends Database
 
         // Generate initial DataLists and base table names
         $lists = array();
-        $baseClasses = array($pageClass => '', $fileClass => '');
+        $sqlTables = array($pageClass => '', $fileClass => '');
         foreach ($classesToSearch as $class) {
             $lists[$class] = DataList::create($class)->where($notMatch . $match[$class] . $extraFilters[$class]);
-            $baseClasses[$class] = '"' . $class . '"';
+            $sqlTables[$class] = '"' . DataObject::getSchema()->tableName($class) . '"';
         }
 
         $charset = static::config()->get('charset');
@@ -237,14 +237,14 @@ class MySQLDatabase extends Database
         // Make column selection lists
         $select = array(
             $pageClass => array(
-                "ClassName", "{$pageTable}.\"ID\"", "ParentID",
+                "ClassName", "{$sqlTables[$pageClass]}.\"ID\"", "ParentID",
                 "Title", "MenuTitle", "URLSegment", "Content",
                 "LastEdited", "Created",
                 "Name" => "_{$charset}''",
                 "Relevance" => $relevance[$pageClass], "CanViewType"
             ),
             $fileClass => array(
-                "ClassName", "{$fileTable}.\"ID\"", "ParentID",
+                "ClassName", "{$sqlTables[$fileClass]}.\"ID\"", "ParentID",
                 "Title", "MenuTitle" => "_{$charset}''", "URLSegment" => "_{$charset}''", "Content" => "_{$charset}''",
                 "LastEdited", "Created",
                 "Name",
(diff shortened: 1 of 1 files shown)


robbieaverill:
Those build failures are my fault, if you rebase onto the latest 4 branch they should pass

src/ORM/Connect/MySQLDatabase.php
```diff
@@ -226,25 +226,25 @@ public function searchEngine(
 
         // Generate initial DataLists and base table names
         $lists = array();
-        $baseClasses = array($pageClass => '', $fileClass => '');
```
tractorcow: `$baseClasses` wasn't even used. Just delete it instead of fixing it?
We already have $pageTable and $fileTable, so we don't need to do it again here.

src/ORM/Connect/MySQLDatabase.php
```diff
@@ -226,25 +226,25 @@ public function searchEngine(
...
             $pageClass => array(
-                "ClassName", "{$pageTable}.\"ID\"", "ParentID",
+                "ClassName", "{$sqlTables[$pageClass]}.\"ID\"", "ParentID",
                 "Title", "MenuTitle", "URLSegment", "Content",
                 "LastEdited", "Created",
                 "Name" => "_{$charset}''",
                 "Relevance" => $relevance[$pageClass], "CanViewType"
             ),
             $fileClass => array(
-                "ClassName", "{$fileTable}.\"ID\"", "ParentID",
```
tractorcow: It seems like the only fix needed here is:

```diff
- "ClassName", "{$fileTable}.\"ID\"", "ParentID",
+ "ClassName", "\"{$fileTable}\".\"ID\"", "ParentID",
```

And likewise for pageTable.

Martimiz:
$baseClasses & $classesToSearch were supposed to enable searching custom classes, but this option sort of remained in twilight mode for a loooong time :)

My suggestion leaves this as it is, open for future improvement. Your idea, I suppose, is to remove that possibility once and for all? In that case we might comment the method to make that clear.  We'd still also need this (#265):

    $query->setFrom("\"$table"\");

Happy to alter the pull in that direction - if that's what you guys decide...

Martimiz:
One (probably stupid) question: Are we to aim pull requests at master or at branch 4 now?

dhensby:
master is v5 now, so 4 if you want it to make 4.0

Martimiz:
OK, so to be clear:

- go for my original version that just fixes the problem, leaving functionality as it is?
- move forward to make this permanently a SiteTree/File search engine, as per your suggestion?

tractorcow:
Nothing is permanent; Just because we don't change it here, doesn't mean we can't change it in the future. You could even raise a separate pull request just for that. However, the core issue for this PR is the missing quote, so it'd be better to just focus on that.

Making this code work for classes other than just File and SiteTree is a lot more work.

Martimiz:
OK - help please, I'm a gitmoron :( 

I wanted this one aimed at 4, but didn't realize master was now 5 (should be in 5 as well though?). 

So if I rebase to 4 (as Robbie said), the pull request would still be aimed at master... this wouldn't be good, would it? 

Should I create a new branch for a pull on 4, and rebase the existing pull request to master and then change it accordingly? :-\

kinglozzer:
@Martimiz If you weren’t already aware, you can change the target branch in Github’s UI now. Next to the pull request title there’s an edit button, you should be able to change it to target `4` in there :)

tractorcow:
Ah, I tried to move it to 4 but it pulled in a lot of extra commits. =(

dhensby:
I'll sort it out :)

dhensby:
done :)

Martimiz:
@dhensby  - is there anything I should do next? Sorry for the confusion, it was meant to be a simple fix, untill I got the branches mixed up.

Tried to rebase to 4, but got into conflicts in composer.json that I couldn't seem to solve. I cóuld just make a simple new request... :-/

dhensby:
tbh I'm not following this in a lot of detail - @tractorcow anything left?

tractorcow:
It looks good.

PR #7175, opened 2017-07-15
FIX Version provider uses non LSB config getters, move LeftAndMain config to admin module

robbieaverill:
Follows merge up of https://github.com/silverstripe/silverstripe-framework/pull/7125

The skipped `testGetModulesFromComposerLock` test showed that the config for the version provider needed to be gathered from the base class, not that provided from late static binding.

Moved LeftAndMain configuration to admin module: https://github.com/silverstripe/silverstripe-admin/pull/153

cc @dhensby


dhensby:
thanks

PR #7176, opened 2017-07-16
FIX Enable code coverage builds with phpdbg and 7.1.7

### LINKED ISSUE #7156: [4.x] Code coverage is not running
In the 4 branch we're removing xdebug as one of the Travis build steps, which means that when the coverage run runs it can't run coverage since it has no coverage driver. It also needs a coverage whitelist.

I did a quick test (#7150) and these are the following problems:

* With xdebug enabled and a whitelist, the build times out before finishing the test suite
* With phpdbg enabled and a whitelist, generating the coverage report segfaults on PHP 7.1.2

Anyway, at the moment the coverage reports aren't being generated.

robbieaverill:
PHP 7.1.7 is currently not segfaulting when generating coverage reports with phpdbg

~~Due to some weird behaviour happening with this combination on TravisCI not setting the base URL correctly, I've moved the SS_BASE_URL environment variable from being a fallback option to being the priority if it's defined - in constants.php.~~ Reverted

Resolves #7156

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,17 +19,17 @@ matrix:
     - php: 5.6
       env:
         - DB=MYSQL
-        - PHPUNIT_COVERAGE_TEST=framework
         - PHPCS_TEST=1
+        - PHPUNIT_TEST=framework
     - php: 7.0
       env:
         - DB=PGSQL
         - PHPUNIT_TEST=framework
-    - php: 7.1.2
+    - php: 7.1
       env:
         - DB=MYSQL
         - PDO=1
-        - PHPUNIT_TEST=framework
+        - PHPUNIT_COVERAGE_TEST=framework
     - php: 7.0
       env:
        - DB=MYSQL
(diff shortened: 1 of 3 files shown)


dhensby:
> Due to some weird behaviour happening with this combination on TravisCI not setting the base URL correctly, I've moved the SS_BASE_URL environment variable from being a fallback option to being the priority if it's defined - in constants.php.

@silverstripe/core-team this is a subtle but perhaps significant change - any thoughts?

robbieaverill:
This is actually caused by using phpdbg- fix at #7179 means I can revert that change

dhensby:
OK = #7179 is merged, please can you revert the base_url change :)

robbieaverill:
Thanks, updated @dhensby :)

sminnee:
This might help https://gist.github.com/61edb321955bf0d6638d3aba6bedeceb

robbieaverill:
Seems legit @sminnee - thanks!

sminnee:
Travis passed - codecov was failing but that's expected for codecov-fixing PRs. Merged!

PR #7177, opened 2017-07-16
NEW Add Behat CMS header tab context methods

robbieaverill:


tests/behat/src/CmsUiContext.php
```diff
@@ -220,6 +229,40 @@ public function stepIClickOnElementInTheTree($method, $text)
...
+    {
+        assertTrue($this->getCmsTabElement($text)->hasClass('active'));
+    }
+
+    /**
+     * @Then the :text header tab should not be active
+     */
+    public function theHeaderTabShouldNotBeActive($text)
+    {
+        assertFalse($this->getCmsTabElement($text)->hasClass('active'));
```
tractorcow: Please check for null from getCmsTabElement() before calling hasClass.

robbieaverill: Updated

PR #7182, opened 2017-07-18
Fixes baked-in node_threshold_total for TreedropdownField

phptek:
node_threshold_total baked-in as 30 in 3.6

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -17,3 +17,5 @@ HTTP:
     must-revalidate: "true"
     no-transform: "true"
   vary: "Cookie, X-Forwarded-Protocol, User-Agent, Accept"
+Hierarchy:
+  node_threshold_total: 30
--- a/forms/TreeDropdownField.php
+++ b/forms/TreeDropdownField.php
@@ -337,9 +337,8 @@ class TreeDropdownField extends FormField {
 		}
 
 		// main node leaf limit configuration
-		$nodeCountThreshold = 30;
 		$obj->markPartialTree(
-			$nodeCountThreshold, $context = null,
+			$nodeCountThreshold = null, $context = null,
 			$this->childrenMethod, $this->numChildrenMethod, $filteredIds
 		);
 
--- a/model/Hierarchy.php
+++ b/model/Hierarchy.php
@@ -237,10 +237,12 @@ class Hierarchy extends DataExtension {
 	 * @param array $filteredIds list of ids that will be used to filter list items (reduce the result set before processing)
 	 * @return int The actual number of nodes marked.
 	 */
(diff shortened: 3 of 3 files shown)


_config/config.yml
```diff
@@ -17,3 +17,5 @@ HTTP:
     must-revalidate: "true"
     no-transform: "true"
   vary: "Cookie, X-Forwarded-Protocol, User-Agent, Accept"
+Hierarchy:
+  node_threshold_total: 30
```
tractorcow: For default values it's better to put these in private statics. YML is more for overrides and injector configs.

forms/TreeDropdownField.php
```diff
@@ -337,9 +337,8 @@ public function tree(SS_HTTPRequest $request) {
 		}
 
 		// main node leaf limit configuration
-		$nodeCountThreshold = 30;
```
tractorcow: You may want a different threshold for treedropdown field than you might use for, say, the sitetree hierarchy view.

_config/config.yml
dhensby: this is already set as a private static so shouldn't be changed here

tractorcow: This is the framework module, why should the framework set a default for it's own class to two different values in two places?

```
private static $node_threshold_total = 50;
```

This PR changes the default to 30. This above 50 is now redundant.

phptek:
@tractorcow simple. I hadn't spotted the private static! I'll amend ASAP.

phptek:
Since re-testing with 3.6-latest, the `TreeDropdownField` "performance" improvement changes that were reverted - means this issue as reported is no longer a problem. Thanks for your feedback, closing PR.

dhensby:
FYI @phptek the performance improvements still live in the 3 branch which means this will become a problem again in 3.7

phptek:
Hmmm what is the merge process regarding the 3 branch? That confused me. Regardless, I'll re-submit the PR

dhensby:
@phptek the way it works is:

Patch branches (X.Y - 3.5, 3.6) are for patches only, only bug fixes should be going into them. Every so often they are then tagged as the latest "patch" release (X.Y.Z - 3.5.2, etc). Changes to these branches are then merged up into the X branch (3, 4, master). Patches in 3.5 will also be merged into 3.6 and then into 3 to ensure all code fixes make their way into all releases.

Minor branches (X - 3, 4, master) are used to receive new (non-breaking) features which are released less frequently as new "minor" releases (X.Y - 3.6, 3.7, etc). These will contain patches from the previous versions along with new features.

phptek:
@dhensby So should feature-branches for bug-fixes for releases 3.6.x be based off the `3.6` branch or the `3.0` branch then?

PR #7185, opened 2017-07-18
FIX Exclude thirdparty dir from code coverage

robbieaverill:


dhensby:
we probably want to enable codecov on 4 branch

robbieaverill:
Looks like this is for the default branch to display on the dashboard rather than disabling others - https://docs.codecov.io/docs/codecov-yaml#section-default-branch

dhensby:
oh, ok. cool

PR #7186, opened 2017-07-18
FIX Add unique prefix to cache stores to prevent cache leak

### LINKED ISSUE #6358: caching on shared environment - add base URL to cachekey to prevent cross-pollination
We have been using caching a lot lately - like this:

https://github.com/silverstripe/silverstripe-framework/blob/67496d3eb017dae09d0e55c607d2e80a60654e70/tests/php/Core/CacheTest.php#L10-L15

That works great - but there is a catch (there are many, but this is a big one):

Unless you specifically make sure that each site uses its own caching system (be it APC / temp files / Memcache) any matching cache keys are going to cross-pollinate.  Of course this only applies if two sites are running on the same server and are using the same key - but it still applies even if they use separate file systems (apc / memcache). 

I confirmed this theory by creating the following task:

```php

<?php


class CacheTest extends BuildTask
{
    protected $title = 'Test Silverstripe Cache';

    protected $description = 'Basic test for the silverstripe cache. It will show the date and time the cache was made.';

    public function run($request)
    {
        $cachekey = 'foo';
        $cache = SS_Cache::factory($cachekey);
        if (!($result = $cache->load($cachekey))) {
            $result = date('Y-m-d H:i:s');;
            $cache->save($result, $cachekey);
        }
        echo $result;
    }

}

```
you can view this task by visiting: www.mysite.co.nz/dev/tasks/CacheTest.

This task returns the exact same second, if:

 - websites are the same server even if there are  in completely separate accounts
 - you are running memcache cache or apc cache (not tested for file based caching) 

So, what I would recommend is that we simply add the URL to the caching key, or at least a simplified version of it.

https://github.com/silverstripe/silverstripe-framework/blob/67496d3eb017dae09d0e55c607d2e80a60654e70/src/Core/Cache.php#L175

and add to the caching id ($for): 

```php
$prepend = hash("crc32b", Director::absoluteBaseURL());
$for = $prepend.$for;
```

As a second rate alternative, I feel that the documentation should at least point out this risk.

My apologies in advance if I have not completely understood how this works.

dhensby:
fixes #6358 

This adds a default prefix to cache identifiers so that servers with multiple sites have a lower likelihood of accessing each-others caches.

### DIFF
--- a/cache/Cache.php
+++ b/cache/Cache.php
@@ -174,7 +174,9 @@ class SS_Cache {
 
 		$backend = self::$backends[$backend_name];
 
-		$basicOptions = array('cache_id_prefix' => $for);
+		$basicOptions = array(
+		    'cache_id_prefix' => $for . '_' . md5(BASE_PATH) . '_',
+        );
 
 		if ($cache_lifetime >= 0) {
 			$basicOptions['lifetime'] = $cache_lifetime;
--- a/core/manifest/ConfigManifest.php
+++ b/core/manifest/ConfigManifest.php
@@ -115,7 +115,8 @@ class SS_ConfigManifest {
 	{
 		return SS_Cache::factory('SS_Configuration', 'Core', array(
 			'automatic_serialization' => true,
-			'lifetime' => null
+			'lifetime' => null,
+			'cache_id_prefix' => 'SS_Configuration_',
 		));
 	}
 


kinglozzer:
Will leave this open for a little while in case anyone has any other concerns with this :)

dhensby:
For config cache we currently generate our own unique hash so we need to incorporate that into this fix

dhensby:
Added fix to config to use custom prefix instead of the auto one

tractorcow:
Looks good thanks @dhensby

PR #7190, opened 2017-07-19
Revert "TreeDropDown performance boost."

dhensby:
Reverts silverstripe/silverstripe-framework#7035

This PR should have gone into the 3 branch.


dhensby:
FYI - this PR was already merged up from 3.5 -> 3 so we don't need to re open a PR for 3 we just shouldn't merge up this revert into 3

tractorcow:
Or merge up, and revert the revert? :P

PR #7193, opened 2017-07-20
TEST Make travis install codesniffer with composer global

dhensby:
`pyrus` doesn't seem to be supported in PHP 5 travis builds anymore.


dhensby:
Taken from #7092

PR #7195, opened 2017-07-20
BUG Fix unassigned nestedFrom

### LINKED ISSUE #7194: Kernel::getNestedFrom does nothing
When nesting a kernel the `nestedfrom` property is never set and so `getNestedFrom` is always null.

This leaves no way to unnest a nested Kernel without keeping a local reference to it.

tractorcow:
Fixes #7194


tractorcow:
@dhensby after a few restarts tests are green. :)

PR #7196, opened 2017-07-20
Add step definition for cms list

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-cms/issues/632

### DIFF
--- a/tests/behat/src/CmsUiContext.php
+++ b/tests/behat/src/CmsUiContext.php
@@ -122,20 +122,6 @@ class CmsUiContext implements Context
         return $cms_tree_element;
     }
 
-    protected function getCmsListElement()
-    {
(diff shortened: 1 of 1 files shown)


tests/behat/src/CmsUiContext.php
```diff
@@ -156,6 +170,28 @@ public function stepIShouldNotSeeInCmsTree($text)
     }
 
     /**
+     * @When /^I should see "([^"]*)" in the list$/
```
tractorcow: Can you please make sure the declaration matches the specifics? This fixture isn't useful for any lists, but exclusively "page list view".

tractorcow:
Merge on green, then we'll re-run CMS tests.

PR #7197, opened 2017-07-21
NEW Add updateRules extension point to Director::handleRequest

robbieaverill:
Allows user code to modify routing rules before they are processed. Middleware is executed after the rules are processed, so is too late.

### DIFF
--- a/src/Control/Director.php
+++ b/src/Control/Director.php
@@ -6,6 +6,7 @@ use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Control\Middleware\HTTPMiddlewareAware;
 use SilverStripe\Core\Config\Configurable;
 use SilverStripe\Core\Environment;
+use SilverStripe\Core\Extensible;
 use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Kernel;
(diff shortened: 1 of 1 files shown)


tractorcow:
@sminnee how do you feel about either this PR, or moving the rules config + main loop into inside the middleware callback?

sminnee:
There's only a single Director created per request, so the performance impact of making it extensive is negligible.

I'd probably shift the logic into a protected getRules() method.

I'm curious to know what the use-case for dynamic rules are, rather than just a handler, but I wouldn't block the PR due to my curiosity. ;-)

I think encouraging middlewares to mess with more of the global state would be a really bad idea. In the idealised case, the middlewares would only mess with the request, and a bunch of stuff that is currently global state (eg current user) would become request attributes. That's not going to happen in SS4 but as a best practise I would limit the scope of middlewares to that principle, and modifying global rules would be a violation.

A better approach to "middlewaring" custom ruled would be to have the entire rules processor available to place within a middleware, but that's a bigger change and probably of questionable value while use of the pattern is still so young in SS.

sminnee:
To be clear: I think a Director extension as proposed in this PR is the most appropriate way of allowing dynamic rules

tractorcow:
Thanks for the feedback @sminnee; I agree with your observation.

PR #7198, opened 2017-07-21
BUG fix config rules to match updated `Except`

tractorcow:
Except rules were meant to be exclusive, not inclusive. Patched in upstream config module, but framework tests need updating.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: php
 
 sudo: false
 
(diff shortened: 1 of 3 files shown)


tractorcow:
Ok, so travis moved to trusty by default which we can't use with selenium. I locked to precise for now.

@dhensby is this ok?

dhensby:
Sure

PR #7200, opened 2017-07-21
FIX Deprecated yml syntax

dhensby:
Since 3.1 of symfony/yaml having unquoted strings starting with `%` is deprecated (https://github.com/symfony/yaml/blob/v3.1.0/Inline.php#L301-L303)

Since 3.2 having duplicate keys fail silently is deprecated (https://github.com/symfony/yaml/blob/v3.2.0/Parser.php#L246)

This PR aims to address those issues.

---

Related PRs:

- [x] https://github.com/silverstripe/silverstripe-cms/pull/1908
- [x] https://github.com/silverstripe/silverstripe-assets/pull/54
- [x] https://github.com/silverstripe/silverstripe-versioned/pull/30

### DIFF
--- a/_config/i18n.yml
+++ b/_config/i18n.yml
@@ -24,14 +24,14 @@ SilverStripe\Core\Injector\Injector:
   Symfony\Component\Translation\Loader\LoaderInterface:
     class: SilverStripe\i18n\Messages\Symfony\ModuleYamlLoader
     properties:
-      Reader: %$SilverStripe\i18n\Messages\Reader
+      Reader: '%$SilverStripe\i18n\Messages\Reader'
   # Ensure our cache respects ModuleYamlLoader's self-invalidation
   # @see DirectoryListResource::isFresh()
   # Note: This could be replaced with a more aggressive cache if necessary on a live environment
   Symfony\Component\Config\ConfigCacheFactoryInterface:
     class: Symfony\Component\Config\ResourceCheckerConfigCacheFactory
     constructor:
-      0: [ %$Symfony\Component\Config\Resource\SelfCheckingResourceChecker ]
+      0: [ '%$Symfony\Component\Config\Resource\SelfCheckingResourceChecker' ]
   # Create default translator with standard cache path and our custom loader
   Symfony\Component\Translation\TranslatorInterface:
     class: Symfony\Component\Translation\Translator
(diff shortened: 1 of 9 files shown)


_config/mimetypes.yml
```diff
@@ -756,7 +756,6 @@ SilverStripe\Control\HTTP:
     stl: application/vnd.ms-pki.stl
     str: application/vnd.pg.format
     stw: application/vnd.sun.xml.writer.template
-    sub: image/vnd.dvb.subtitle
```
dhensby: We may need to come up with a way of handling multiple mime types for a single extension

_config/model.yml
```diff
@@ -87,13 +87,14 @@ SilverStripe\Core\Injector\Injector:
...
 ---
 SilverStripe\Core\Injector\Injector:
-  EndsWithFilter: %$DataListFilter.EndsWith
-  ExactMatchFilter: %$DataListFilter.ExactMatch
-  FulltextFilter: %$DataListFilter.Fulltext
+  EndsWithFilter: "%$DataListFilter.EndsWith"
+  ExactMatchFilter: "%$DataListFilter.ExactMatch"
+  FulltextFilter: "%$DataListFilter.Fulltext"
+  # is this meant to be `&$`?
   GreaterThanFilter: $&DataListFilter.GreaterThan
```
dhensby: I'm not sure if this is an error... Should this be `%$`?

tractorcow: You're right it's backwards.

_config/i18n.yml
```diff
@@ -40,23 +40,23 @@ SilverStripe\Core\Injector\Injector:
...
+      Translator: "%$Symfony\\Component\\Translation\\TranslatorInterface"
 ---
 Name: textcollector
 ---
 SilverStripe\Core\Injector\Injector:
   SilverStripe\i18n\TextCollection\i18nTextCollector:
     properties:
-      Reader: %$SilverStripe\i18n\Messages\Reader
-      Writer: %$SilverStripe\i18n\Messages\Writer
+      Reader: "%$SilverStripe\\i18n\\Messages\\Reader"
```
tractorcow: Can we use single quotes and not have to double escape slash? It's simpler and more standard across all our yml.

dhensby: :(

dhensby: Fixed

_config/model.yml
dhensby: Fixed

_config/mimetypes.yml
kinglozzer: Given that this is for HTTP responses, and the header only supports one type at a time (http://greenbytes.de/tech/webdav/rfc2616.html#rfc.section.14.17), can we just discard these duplicates and forget about it?

dhensby: I think we can as everything seems to have been working up to this point and there would have been an overwritten value anyway

PR #7201, opened 2017-07-21
Ensure ClassManifest isn't flushed twice on build

kinglozzer:
Shaves ~10% off the wall time of a `dev/build` for a “blank” install

screen shot 2017-07-21 at 14 39 30

### DIFF
--- a/src/ORM/DatabaseAdmin.php
+++ b/src/ORM/DatabaseAdmin.php
@@ -122,8 +122,12 @@ class DatabaseAdmin extends Controller
         // The default time limit of 30 seconds is normally not enough
         Environment::increaseTimeLimitTo(600);
 
-        // Get all our classes
-        ClassLoader::inst()->getManifest()->regenerate(false);
+        // If this code is being run outside of a dev/build or without a ?flush query string param,
+        // the class manifest hasn't been flushed, so do it here
+        $request = $this->getRequest();
+        if (!array_key_exists('flush', $request->getVars()) && strpos($request->getURL(), 'dev/build') !== 0) {
+            ClassLoader::inst()->getManifest()->regenerate(false);
+        }
 
         $url = $this->getReturnURL();
         if ($url) {


src/ORM/Connect/DBSchemaManager.php
```diff
@@ -640,8 +650,15 @@ public function requireField($table, $field, $spec)
             $specValue = $this->IdColumn(true);
         }
 
+        // Cache of fields
+        static $fieldListCache = [];
+
         if (!$newTable) {
-            $fieldList = $this->fieldList($table);
+            if (!isset($fieldListCache[$table])) {
+                $fieldListCache[$table] = $this->fieldList($table);
```
kinglozzer: On a “blank” install (framework + cms + asset-admin), this + the index list cache above saves around 300 queries per `dev/build` - we’re currently running a query to fetch a full list of fields for _each_ field we require.

src/ORM/DatabaseAdmin.php
```diff
@@ -122,8 +124,11 @@ public function build()
         // The default time limit of 30 seconds is normally not enough
         Environment::increaseTimeLimitTo(600);
 
-        // Get all our classes
-        ClassLoader::inst()->getManifest()->regenerate(false);
+        // Flush the manifest if we haven't already
+        $kernel = Injector::inst()->get(Kernel::class);
+        if (!$kernel->getFlushedOnBoot()) {
+            ClassLoader::inst()->getManifest()->regenerate(false);
+        }
```
kinglozzer: We don’t really need to keep this at all, I’m just being defensive in case anyone wants to run a build programatically like this

dhensby:
I think there needs to be some way to flush the cache during tests

src/ORM/Connect/DBSchemaManager.php
kinglozzer: Removed this, it completely breaks tests as we remove and re-build tables for each test we run... will need more thought if it’s to be implemented

dhensby:
you need a rebase

src/ORM/DatabaseAdmin.php
tractorcow: Instead of having new API in CoreKernel, why can't you just check to see if `flush` queryparam is set, and then flush if it's not?

kinglozzer: My intention was to store whether a flush has been performed in the kernel, and then look to refactor other places in core where we have hardcoded URLs/query strings to fetch that info from the kernel instead. Not necessary for this PR though, removed it

dhensby:
we are green

PR #7206, opened 2017-07-25
Enhancement add support for TreeMultiselectField in react

flamerohr:
Please see https://github.com/silverstripe/silverstripe-admin/issues/53 for more details


tractorcow:
image

missing `t` in `geMenuTitle`.

I'll push up a fix. :)

tractorcow:
Second bug: Selecting a parent stops you from being allowed to select the children.

image

I think the fix is to NOT remove selected items from the list.

PR #7207, opened 2017-07-25
Add title, classes, and icon for action in saved version

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-campaign-admin/issues/16

### DIFF
--- a/src/Forms/FormAction.php
+++ b/src/Forms/FormAction.php
@@ -70,6 +70,10 @@ class FormAction extends FormField
      */
     protected $validationExempt = false;
 
+    protected $savedTitle = null;
+    protected $savedClasses = null;
+    protected $savedIcon = null;
+
     /**
      * Create a new action button.
      *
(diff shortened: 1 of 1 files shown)


src/Forms/FormAction.php
```diff
@@ -93,9 +97,45 @@ public function getSchemaDataDefaults()
         $defaults = parent::getSchemaDataDefaults();
         $defaults['attributes']['type'] = $this->getUseButtonTag() ? 'button' : 'submit';
         $defaults['data']['icon'] = $this->getIcon();
+        $defaults['data']['savedTitle'] = $this->getSavedTitle();
+        $defaults['data']['savedClasses'] = $this->getSavedClasses();
+        $defaults['data']['savedIcon'] = $this->getSavedIcon();
         return $defaults;
     }
 
+    public function setSavedTitle($title)
```
tractorcow: None of these methods describe what each of these are used for; They also seem to be very specific use cases baked into the generic form action class, which feels a little over-zealous.

You could generalise it, so you don't need so many extra methods into a saved state?

```php
FormAction::create('save', _t(__CLASS__.'.SAVE', 'Save'))
    ->setIcon('save')
    ->setSavedState([
    	'title' => _t(__CLASS__.'.SAVED', 'Saved'),
    	'icon' => 'tick',
    	'extraClass' => 'primary-outline',
   	]);
```
and in js
```js
getProps() {
	let props = Object.assign({}, this.props, this.props.data);
	if (this.changed) {
		props = Object.assign({}, this.props.data.savedState);
	}
	return props;
}
```

Or something :P

phalkunz: @tractorcow Thanks :) Will fix it now

tractorcow:
For context, this is what we do in SiteTree at the moment (entwine styles)

```php
FormAction::create('save', _t(__CLASS__.'.BUTTONSAVED', 'Saved'))
                    ->addExtraClass('btn-secondary-outline font-icon-check-mark')
                    ->setAttribute('data-btn-alternate', 'btn action btn-primary font-icon-save')
                    ->setUseButtonTag(true)
                    ->setAttribute('data-text-alternate', _t('SilverStripe\\CMS\\Controllers\\CMSMain.SAVEDRAFT', 'Save draft'))
```

src/Forms/FormAction.php
flamerohr: I don't recommend mixing `this.props` and `this.props.data` like that `getProps()` suggestion...
if the data cannot be clearly discovered in a given path and needs to be explored in multiple areas, maybe the API needs to be re-assessed.

flamerohr:
Closing, as the issue has been solved

PR #7211, opened 2017-07-26
Fix: Use Config API for MemberAuthenticator::$migrate_legacy_hashes (fixes #7208)

### LINKED ISSUE #7208: MemberAuthenticator::$migrate_legacy_hashes should use Config API

kinglozzer:


kinglozzer:
Do we want to cater for people (like me) who have worked around this by using `ReflectionProperty` to modify the property directly? If we do, it’s a one-liner:

```php
$migrateLegacyHashes = array_merge($migrateLegacyHashes, self::$migrate_legacy_hashes);
```

I don’t feel strongly about it at all, you take your chances with hacks like that 😉

dhensby:
Is this correct:

At the moment it's impossible to change the values in the legacy hashes (unless you hack it together using reflection as you have). This PR allows users who have (or want to) use the Config to set the legacy migrations?

kinglozzer:
That’s correct @dhensby

dhensby:
I'm happy - using reflection to edit private statics is not a public API ;)

PR #7215, opened 2017-07-26
check SERVER for HTTPS so _config.php doesn't infinite loop

### LINKED ISSUE #7214: Director::forceSSL(); from _config.php results infinite redirections SS 4 Beta 1
This happens when the `\SilverStripe\Control\Director::forceSSL();` with mysite/_config.php file. 

The main reason behind is that `static::currentRequest($request);` returns null. 

However calling it via a Controller::init function works.

andrewandante:
Re-adds the check in $_SERVER so that using `Director::forceSSL()` in `_config.php` as recommended doesn't result in an infinite loop.

Fixes #7214

### DIFF
--- a/src/Control/Director.php
+++ b/src/Control/Director.php
@@ -875,6 +875,10 @@ class Director implements TemplateGlobalProvider
     {
         // if no request i.e. has been set in _config.php - prevents loops
         if (!static::currentRequest()) {
+            user_error(
+                'Director::forceSSL() requires a current request. This usually means you have tried to  call this from _config.php. If so, please add this to the init() method of your PageController.',
+                E_USER_WARNING
+            );
             return false;
         }
 
@@ -922,6 +926,10 @@ class Director implements TemplateGlobalProvider
     {
         // if no request i.e. has been set in _config.php - prevents loops
         if (!static::currentRequest()) {
+            user_error(
+                'Director::forceWWW() requires a current request. This usually means you have tried to call this from _config.php. If so, please add this to the init() method of your PageController.',
+                E_USER_WARNING
+            );
             return false;
         }
 
--- a/src/Control/Director.php
+++ b/src/Control/Director.php
@@ -843,7 +843,7 @@ class Director implements TemplateGlobalProvider
     /**
      * Force the site to run on SSL.
      *
-     * To use, call from _config.php. For example:
+     * To use, call from the init() method of your PageController. For example:
      * <code>
      * if (Director::isLive()) Director::forceSSL();
      * </code>
@@ -873,6 +873,11 @@ class Director implements TemplateGlobalProvider
      */
     public static function forceSSL($patterns = null, $secureDomain = null)
     {
+        // if no request i.e. has been set in _config.php - prevents loops
+        if (!static::currentRequest()) {
+            return false;
+        }
+
         // Already on SSL
         if (static::is_https()) {
             return true;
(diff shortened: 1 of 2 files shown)


dhensby:
we are trying to decouple from the server globals - this should be checked against the current HTTPRequest rather than some global state.

The real issue is that the force SSL method should either be removed or we should provide a middleware for people to add to the HTTPApplication.

tractorcow:
> we are trying to decouple from the server globals - this should be checked against the current HTTPRequest rather than some global state.

The issue is that when _config.php is included this is too early for the HTTPRequest to be available.

src/Control/Director.php
```diff
@@ -563,6 +563,11 @@ public static function is_https(HTTPRequest $request = null)
             }
         }
 
+        if (!empty($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != 'off') {
```
tractorcow: This code needs to be higher than `default_base_url` check, but lower than the currentRequest(), since `default_base_url` is almost never going to be empty, and will almost always be a poor approximation.

tractorcow:
Needs to be ported to 4 branch on merge (or rebased against 4, @andrewandante )

dhensby:
I don't see how moving this to a middleware didn't solve the problem? Devs can add a HTTPS middleware to add this behaviour...

tractorcow:
Hm, ok... I'll add a changelog note to this then.

dhensby:
We should discourage the use of PHP, but especially Director for forcing SSL (and WWW subdomain). I'm disappointed these APIs made the beta :/

tractorcow:
>  We should discourage the use of PHP, but especially Director for forcing SSL (and WWW subdomain). I'm disappointed these APIs made the beta :/

These were always in Director though. They just rely less on globals.

dhensby:
I mean I'm disappointed they weren't removed

tractorcow:
How about we check for a current HTTPRequest, and make forceWWW / forceSSL methods fail if null?

dhensby:
Sure... Or remove them ;)

tractorcow:
Too late to remove API sorry. =(

andrewandante:
Have updated PR to check for a current request. Likely to need a notice/warning better than an updated docblock - I believe @dhensby is advocating for removal in 5.0

dhensby:
tests need updating, it seems :/

dhensby:
Here's a naive middleware to force ssl:

```php

class SSLMiddleware implements \SilverStripe\Control\Middleware\HTTPMiddleware
{
    public function process(\SilverStripe\Control\HTTPRequest $request, callable $delegate)
    {
        $result = null;

        if (!$request->getScheme() !== 'https') {
            $result = \SilverStripe\Control\HTTPResponse::create(null, 301);
            $result->addHeader('Location', 'https://' . $request->getHeader('Host') . $request->getURL(true));
        } else {
            $result = call_user_func($delegate, $request);
        }

        return $result;
    }
}
```

It seems you need to override main.php as I can't see a way to load middlewares without having your own bootstrap.

andrewandante:
_shakes fist at behat test_

src/Control/Director.php
```diff
@@ -873,6 +873,11 @@ protected static function force_redirect($destURL)
      */
     public static function forceSSL($patterns = null, $secureDomain = null)
     {
+        // if no request i.e. has been set in _config.php - prevents loops
+        if (!static::currentRequest()) {
+            return false;
```
tractorcow: Can you make this a warning? If you hit this it's really an error condition that should be fixed in code.

tractorcow:
Good fix just one more suggestion please. :)

andrewandante:
Added in warnings - is that the "correct" SilverStripe way to do it? I copied it from DataObject :)

tractorcow:
I don't think behat failure is your fault... let me restart anyway. We need to adjust framework behat timeout I think.

PR #7216, opened 2017-07-26
Optimise travis composer installation

dhensby:
ping @tractorcow 

We had previously attempted to add caching to travis but given up on it because it caused cross build pollination due to the way our travis-support module worked.

However, we now don't use travis-support and use the more conventional composer installation approach which should make this caching safer and we can start to see some performance gains 🤞 

---
Postgres build on PHP7:

Branch | Time
--- | ---
Current 4 branch | 56.5s
Optimised installation | 36s
Travis caching | 40s

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,12 +59,12 @@ before_script:
 # Install composer dependencies
   - export PATH=~/.composer/vendor/bin:$PATH
   - composer validate
-  - composer install --prefer-dist
-  - if [[ $DB == PGSQL ]]; then composer require silverstripe/postgresql:2.0.x-dev --prefer-dist; fi
-  - if [[ $DB == SQLITE ]]; then composer require silverstripe/sqlite3:2.0.x-dev --prefer-dist; fi
-  - composer require silverstripe/config:1.0.x-dev silverstripe/admin:1.0.x-dev silverstripe/assets:1.0.x-dev silverstripe/versioned:1.0.x-dev --prefer-dist
-  - if [[ $PHPUNIT_TEST == cms ]] || [[ $BEHAT_TEST == cms ]]; then composer require silverstripe/cms:4.0.x-dev silverstripe/campaign-admin:1.0.x-dev silverstripe/siteconfig:4.0.x-dev silverstripe/reports:4.0.x-dev --prefer-dist; fi
+  - if [[ $DB == PGSQL ]]; then composer require silverstripe/postgresql:2.0.x-dev --no-update; fi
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -59,12 +63,12 @@ before_script:
...
-  - if [[ $PHPUNIT_TEST == cms ]] || [[ $BEHAT_TEST == cms ]]; then composer require silverstripe/cms:4.0.x-dev silverstripe/campaign-admin:1.0.x-dev silverstripe/siteconfig:4.0.x-dev silverstripe/reports:4.0.x-dev --prefer-dist; fi
+  - if [[ $DB == PGSQL ]]; then composer require silverstripe/postgresql:2.0.x-dev --no-update; fi
+  - if [[ $DB == SQLITE ]]; then composer require silverstripe/sqlite3:2.0.x-dev --no-update; fi
+  - composer require silverstripe/config:1.0.x-dev silverstripe/admin:1.0.x-dev silverstripe/assets:1.0.x-dev silverstripe/versioned:1.0.x-dev --no-update
+  - if [[ $PHPUNIT_TEST == cms ]] || [[ $BEHAT_TEST == cms ]]; then composer require silverstripe/cms:4.0.x-dev silverstripe/campaign-admin:1.0.x-dev silverstripe/siteconfig:4.0.x-dev silverstripe/reports:4.0.x-dev --no-update; fi
   - if [[ $PHPCS_TEST ]]; then composer global require squizlabs/php_codesniffer:^3 --prefer-dist --no-interaction --no-progress --no-suggest -o; fi
+  - composer install --prefer-dist --no-interaction --no-progress --no-suggest --optimize-autoloader --verbose --profile
```
tractorcow: 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏 🎏

tractorcow:
I remember trying this a few years ago; I think your approach is pretty precise (caching files, not metadata).

tractorcow:
Now we just need PHP coverage tests running in less than 45 minutes.

dhensby:
I'm not sure the caching actually has any tangible benefits to speed because the time it takes to download the cache and set it up is about as long as the downloads through composer...

Oh well

dhensby:
No matter what composer seems to have a 25-30 second pause between calculating the dependencies and installing the first package :/

PR #7219, opened 2017-07-27
BUGFIX: DBComposite doesn't allow arbitrary property assignment

unclecheese:
To be more consistent with `ViewableData`, whose `setField()` method will fallback on [assigning properties arbitrarily](https://github.com/silverstripe/silverstripe-framework/blob/4/src/View/ViewableData.php#L213), `DBComposite` shouldn't bail out of `setField` when the field specified isn't in the record.

Arbitrary property assignment is particularly important in injection.

```yaml
SilverStripe\ORM\FieldType\DBComposite:
  dependencies:
    myService: %$Service
```

Right now, that fails, because `$obj->myService = Service` invokes `__set()` which calls `setField()` which refuses the assignment when `myService`is not in the record.

### DIFF
--- a/src/ORM/FieldType/DBComposite.php
+++ b/src/ORM/FieldType/DBComposite.php
@@ -240,8 +240,9 @@ abstract class DBComposite extends DBField
     {
         $this->objCacheClear();
 
-        // Skip non-db fields
+        // Non-db fields get assigned as normal properties
         if (!$this->hasField($field)) {
+            $this->$field = $value;
             return $this;
         }
 
--- a/src/ORM/FieldType/DBComposite.php
+++ b/src/ORM/FieldType/DBComposite.php
@@ -242,7 +242,8 @@ abstract class DBComposite extends DBField
 
         // Non-db fields get assigned as normal properties
         if (!$this->hasField($field)) {
-            $this->$field = $value;
+            parent::setField($field, $value);
+            
             return $this;
         }
 


src/ORM/FieldType/DBComposite.php
```diff
@@ -240,8 +240,9 @@ public function setField($field, $value, $markChanged = true)
     {
         $this->objCacheClear();
 
-        // Skip non-db fields
+        // Non-db fields get assigned as normal properties
         if (!$this->hasField($field)) {
+            $this->$field = $value;
```
tractorcow: I really dislike this change, as it encourages poor style. You shouldn't be using setField() to shiv in public access to potentially protected members, as it violates the open closed principle. ViewableData suffers from being an overly-generic "god" parent class in much the way that Object once was, but I don't think we should be shuffling it's behaviour into subclasses.

If you have a service you need injected, you should be implementing service endpoint getters and setters instead for those instead.

E.g.

```php
public function setMyService($service) {
	$this->service = $service
};
```

Not

```php
protected $myService; // protected, but will be set by injector anyway as though it was public
```

dhensby: Yep, use setters and getters

dhensby:
Hmmm, though what if you want to inject properties on an object you can't add setters and getters to?

kinglozzer:
> Hmmm, though what if you want to inject properties on an object you can't add setters and getters to?

How would you add the alternative: `protected $foo`? I don’t think we can support hacking in extra dependencies to core classes like that

dhensby:
Sure, it would be a defacto public property - but the point is that setField wouldn't let that happen, right?

unclecheese:
That's just it, though. You can't add getters and setters to core classes. So while I agree it's bad practice, it's a practice that already accepted and forgiven by ViewableData, so I don't understand why you would revoke that for just this one of many dozens of decendants of that class. That's just inconsistent and frustrating.

Would a more acceptable pattern be to add the setter via an extension, and store the service on the extension instance?

dhensby:
> Would a more acceptable pattern be to add the setter via an extension, and store the service on the extension instance?

You almost certainly wouldn't want to store it on the extension instance because I *believe* the same extension instance is shared between all classes with the extension enabled and the `$owner` property is just changed...

unclecheese:
Yup. That's what I thought. So that leaves the same option. $this->owner->service = $service. Where "service" is an undeclared member property. Is that any less ugly?

dhensby:
> Yup. That's what I thought. So that leaves the same option. $this->owner->service = $service. Where "service" is an undeclared member property. Is that any less ugly?

Would that not use `setField` and so break?

I'm leaning towards agreeing that a subclass shouldn't be breaking the bad behaviour of `ViewableData` - we should fix the root problem but not block it's continuation in the meantime, especially if there are tangible issues with doing so.

unclecheese:
> Would that not use setField and so break?

In the case of `DBComposite`, yes. But if we're talking about the general practice of extending core classes with dependencies, then it seems like whether you do it via config or extensions, it's really all the same type of violation.

The only other thing that comes to mind is if I customised the field.

```yaml
  DBComposite:
    class: My\Custom\DBComposite
My\Custom\DBComposite:
  properties:
    MyService: %$Service
```

```php
class DBComposite extends BaseDBComposite
{
  public function setMyService($service);
}
```

kinglozzer:
@unclecheese I’m struggling to understand the point of adding new dependencies to core classes, surely there’s no code that can actually _use_ the extra dependencies? If there is, it probably lives in a subclass, an extension, or a dependent class so isn’t the dependency better off injected there?

kinglozzer:
> That's just it, though. You can't add getters and setters to core classes. So while I agree it's bad practice, it's a practice that already accepted and forgiven by ViewableData, so I don't understand why you would revoke that for just this one of many dozens of decendants of that class. That's just inconsistent and frustrating.

I wouldn’t say it’s an accepted practice, it’s a bug in `ViewableData`. It’s the same issue we have with `ViewableData::__get()` and `ViewableData::getField()` - protected properties are usually protected for a reason, and we shouldn’t be able to access them, but can because of that bug. IMO doing extra work to facilitate this is going to tell people that it _is_ accepted practice

unclecheese:
It's more than just ViewableData, though. The whole reason this came up is because it's something Injector [wants to do](https://github.com/silverstripe/silverstripe-framework/blob/4/src/Core/Injector/Injector.php#L767). If we were to fix the "bug" in `ViewableData`, then one of the two ways that `Injector` injects dependencies (via `set...()` or via direct property assignment) would no longer work with any ViewableData instance.

This is a pattern we have implicitly encouraged, so the way I see it, if we're going to fix the bug in `ViewableData`, the solution can't simply be to refuse direct property assignment in `__set()`. We'd need to accomodate that somehow.

kinglozzer:
I’ve just been comparing the behaviour of this/Injector with PHP’s “native” behaviour around visibility and I’ve proven myself wrong. The original code snippet in this PR **should** work, provided there’s no property named `$myService`: https://3v4l.org/fdMbu.

So I’m going to do a full-180 and say that we should fix this, sorry @unclecheese!

tractorcow:
> Hmmm, though what if you want to inject properties on an object you can't add setters and getters to?

You can use an extension.

>  You almost certainly wouldn't want to store it on the extension instance because I believe the same extension instance is shared between all classes with the extension enabled and the $owner property is just changed...

Actually they are per-instance. I used to think the above, but I was incorrect. It's safe to store services on extensions.

On the other hand, that begs a question. If an object doesn't know about a service, how is it going to use it? It's not really a dependency in that case is it?

> That's just it, though. You can't add getters and setters to core classes. So while I agree it's bad practice, it's a practice that already accepted and forgiven by ViewableData, so I don't understand why you would revoke that for just this one of many dozens of decendants of that class. That's just inconsistent and frustrating.

I don't think ViewableData should do it either. A single instance of bad behaviour doesn't eternally justify it.

src/ORM/FieldType/DBComposite.php
tractorcow: Can we at least use `return parent::setField($field, $value)` and minimise direct access to unknown variables.

unclecheese:
I'm still not clear on why Injector uses arbitrary property assignment as a main way of injecting dependencies if it's such a bad thing to do:

```
    protected function setObjectProperty($object, $name, $value)
    {
        if (method_exists($object, 'set'.$name)) {
            $object->{'set'.$name}($value);
        } else {
            $object->$name = $value;
        }
    }
```

I think we're pretty clear on the issues with `ViewableData`. My concern is that if we fix the `ViewableData` "bug", it's no longer 100% compatible with Injector.

tractorcow:
@kinglozzer I think you aren't testing the correct concerns.

Please see https://3v4l.org/f16oo

unclecheese:
Injector uses reflection to make sure it only writes to public properties. Maybe we need something like that in `ViewableData`, too?

tractorcow:
Yes exactly @unclecheese that would fix ViewableData.

For now, you can use parent::setField(), so at least this bug stays in one place.

tractorcow:
> Injector uses reflection to make sure it only writes to public properties

It doesn't seem to be no.

unclecheese:
Then what's going on here? https://github.com/silverstripe/silverstripe-framework/blob/4/src/Core/Injector/Injector.php#L679

tractorcow:
That's only for autoScanProperties, which is false by default.

unclecheese:
Just to close this up:

* If extension instances are per-instance, then putting the services into the extension along with getters and setters seems like an ideal solution.
* If I were to submit a patch to `ViewableData` that checked the visibility of the property before setting it, would that, in theory, be a suitable fix?
* If that got merged, could we proceed with this one? (although technically they don't have to happen in any particular order)

tractorcow:
> If that got merged, could we proceed with this one? (although technically they don't have to happen in any particular order)

MVP to get this pr over the line is to use the parent::setField() call instead; I don't mind if you follow up with a ViewableData change later, but it's not as critical (it's been a thorn in our side for years it won't hurt it to sit a while longer).

PR #7221, opened 2017-07-28
Fix add function for selecting a value in the new tree dropdown

flamerohr:
This fixes the behat tests for https://github.com/silverstripe/silverstripe-asset-admin/pull/564


tractorcow:
@flamerohr when this is green please merge, and restart the asset-admin PR.

PR #7224, opened 2017-07-28
Move spyc dependency to composer

### LINKED ISSUE #7223: PHP Fatal error:  Cannot redeclare class Spyc
I've added a new dependency via composer, specifically:
`piwik/device-detector`

This has a dependency in it's composer of `mustangostang/spyc`

When I run my code in the browser, it all works fine, however when I want to run any PHPUnit tests, I get the following error:
`PHP Fatal error:  Cannot redeclare class Spyc`

Further investigation reveals that the dependency `mustangostang/spyc` redeclares the class Spyc.

This means, no phpunit tests can be run at all.

I'm not having any luck trying to exclude this depedency, possibly some phpunit bootstrap or configuration?

SS 3.6.1

dhensby:
Fixes #7223


oddnoc:
This change caused a regression in a site I maintain where `thirdparty/spyc/spyc.php` was being required.

dhensby:
Where was it being required?

oddnoc:
In `mysite/_config.php`. In my particular case, that code was no longer needed (the developer who put it there had removed the actual Spyc calls), so it does not impact me directly. Nonetheless, this could be interpreted as a semver violation.

dhensby:
@oddnoc ok - thanks for explaining.

The alternative approach would have been to keep it in thirdparty whilst wrapping the `require_once` in a `!class_exists()` but I felt at the time that the likelihood that this lib was required by other developers (and that the APIs haven't changed and that we don't typically judge file location as an API unless we are documenting that people should rely on it) I felt this was an acceptable change...

I'm happy to reconsider if others think there's an issue.

PR #7228, opened 2017-07-31
API Replace legacy tiny_mce_gzip compressor with asset generator

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-admin/issues/74


flamerohr:
Failing phpunit tests, perhaps the `hash` part of the assert shouldn't be hardcoded? Or ensure config is the same for each unit test.

tractorcow:
Tests needed to set a fixed baseurl, otherwise hash would be different (intentional, but not saved per test).

tractorcow:
Ok, this should fix test state.

flamerohr:
heheh, I need to think about my complaint rather than just "it's way too big" :D
but yeah, the 400 height is better than the 400 + 120...

flamerohr:
heheh, I need to think about my complaint rather than just "it's way too big" :D
but yeah, the 400 height is better than the 400 + 120...

ssmarco:
Nice Solution 👍

tractorcow:
ok, found the issue was due to my local theme breaking tests. Disabling theme for the test fixes this.

PR #7233, opened 2017-08-02
NEW Added MySQL SSL PDO Support (fixes #7077)

### LINKED ISSUE #7077: SSL capability for PDO MySQL driver
Hi all,

There doesn't seem to be a default way to include SSL capabilities for PDO MySQL remote database connections.

Please correct me if I'm wrong.

Thanks,

John

johndalangin:
Modified ConfigureFromEnv.php to parse SS_DATABASE_SSL variables (also added a bit of documentation)
Modified PDOConnector.php to implement variables set in ConfigureFromEnv if exists
Modified install files MySQLDatabaseConfigurationHelper and install.php5 to accept and implement SS_DATABASE_SSL variables set in _ss_environment.php

TODO: Add documentation

### DIFF
--- a/conf/ConfigureFromEnv.php
+++ b/conf/ConfigureFromEnv.php
@@ -25,7 +25,7 @@
  * - SS_DATABASE_SSL_KEY:		Path to SSL private key file
  * - SS_DATABASE_SSL_CERT:		Path to SSL certificate file
  * - SS_DATABASE_SSL_CA:		Path to SSL CA file
- * - SS_DATABASE_SSL_CIPHER: 	Alternative cipher (defaults to DHE-RSA-AES256-SHA)
+ * - SS_DATABASE_SSL_CIPHER: 	Override default cipher (defaults are set per database connector)
  *
  *
  * There is one more setting that is intended to be used by people who work on SilverStripe.
@@ -125,23 +125,28 @@ if(defined('SS_DATABASE_USERNAME') && defined('SS_DATABASE_PASSWORD')) {
 	if(defined('SS_DATABASE_MEMORY'))
 		$databaseConfig["memory"] = SS_DATABASE_MEMORY;
 
-	// PDO MySQL SSL parameters
-	if(defined('SS_DATABASE_CLASS') && SS_DATABASE_CLASS === 'MySQLPDODatabase') {
+	// Add ssl parameters to databaseConfig if these are defined
+	if(
+		defined('SS_DATABASE_SSL_KEY') && 
+		defined('SS_DATABASE_SSL_CERT')
+		) {
 
-		// add ssl parameters if these are defined
-		if(
-			defined('SS_DATABASE_SSL_KEY') && 
-			defined('SS_DATABASE_SSL_CERT') &&
-			defined('SS_DATABASE_SSL_CA')
-			) {
+		$databaseConfig['ssl_key'] = SS_DATABASE_SSL_KEY;
+		$databaseConfig['ssl_cert'] = SS_DATABASE_SSL_CERT;
 
-			$databaseConfig['ssl_key'] = SS_DATABASE_SSL_KEY;
-			$databaseConfig['ssl_cert'] = SS_DATABASE_SSL_CERT;
-			$databaseConfig['ssl_ca'] = SS_DATABASE_SSL_CA;
-			$databaseConfig['ssl_cipher'] = defined('SS_DATABASE_SSL_CIPHER') ? SS_DATABASE_SSL_CIPHER : 'DHE-RSA-AES256-SHA';
-
-		}
 	}
+
+	// Some databases do not require a CA so this is optional
+	if(defined('SS_DATABASE_SSL_CA')) {
+		$databaseConfig['ssl_ca'] = SS_DATABASE_SSL_CA;		
+	}
+
+	// Cipher defaults should be set per connector
+	if(defined('SS_DATABASE_SSL_CIPHER')) {
+		$databaseConfig['ssl_cipher'] = SS_DATABASE_SSL_CIPHER;
+	}
+
+
 
 }
 
--- a/conf/ConfigureFromEnv.php
+++ b/conf/ConfigureFromEnv.php
@@ -20,6 +20,14 @@
  *  - SS_DATABASE_MEMORY:   Use in-memory state if possible. Useful for testing, currently only
  *                          supported by the SQLite database adapter.
  *
+ * SSL Support (for MySQLPDODatabase)
+ *
+ * - SS_DATABASE_SSL_KEY:		Path to SSL private key file
+ * - SS_DATABASE_SSL_CERT:		Path to SSL certificate file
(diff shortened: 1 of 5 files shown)


johndalangin:
@dhensby fixes #7077 

Thank you.

conf/ConfigureFromEnv.php
```diff
@@ -116,6 +124,25 @@
 	// For SQlite3 memory databases (mainly for testing purposes)
 	if(defined('SS_DATABASE_MEMORY'))
 		$databaseConfig["memory"] = SS_DATABASE_MEMORY;
+
+	// PDO MySQL SSL parameters
+	if(defined('SS_DATABASE_CLASS') && SS_DATABASE_CLASS === 'MySQLPDODatabase') {
```
dhensby: Is this only for MySQL using PDO or do all PDO drivers support this?

dhensby:
I'd like @tractorcow's thoughts, again

johndalangin:
@tractorcow @dhensby Thank you for your prompt response. I checked the contents of `/model/connect` and it seems that the only 2 DB connectors included by default are MySQLi and MySQLPDO.

I've taken a look at the MySQLi documentation and it seems that there might be a way to implement SSL via the `mysqli::ssl_set` directive. (http://php.net/manual/en/mysqli.ssl-set.php). I can create another issue and work on this as well.

I don't have any experience working with PostgreSQL or MSSQL with Silverstripe so it might take me a while to get working on an implementation. Maybe we can indicate in the documentation that SSL is only supported on MySQL via MySQLi and PDOMySQL in the mean time so we can push this forward and worry about tight coupling later? I can work on the documentation as well. 

Let me know your thoughts.

Thank you.

conf/ConfigureFromEnv.php
johndalangin: For now it's only for MySQL using PDO.

conf/ConfigureFromEnv.php
```diff
@@ -116,6 +124,25 @@
...
+
+		// add ssl parameters if these are defined
+		if(
+			defined('SS_DATABASE_SSL_KEY') && 
+			defined('SS_DATABASE_SSL_CERT') &&
+			defined('SS_DATABASE_SSL_CA')
+			) {
+
+			$databaseConfig['ssl_key'] = SS_DATABASE_SSL_KEY;
```
tractorcow: My feeling is that perhaps we should set these for all connectors, and we allow classes that support these options to configure them internally. Otherwise, $databaseConfig will keep these values but will be ignored by unsupported backends.

As it stands I wouldn't be able to support a custom connector with SSL.

dev/install/install.php5
```diff
@@ -132,6 +132,20 @@ if(isset($_REQUEST['db'])) {
...
+		if(
+			defined('SS_DATABASE_SSL_KEY') && 
+			defined('SS_DATABASE_SSL_CERT') &&
+			defined('SS_DATABASE_SSL_CA')
+			) {
+
+			$databaseConfig['ssl_key'] = SS_DATABASE_SSL_KEY;
+			$databaseConfig['ssl_cert'] = SS_DATABASE_SSL_CERT;
+			$databaseConfig['ssl_ca'] = SS_DATABASE_SSL_CA;
+			$databaseConfig['ssl_cipher'] = defined('SS_DATABASE_SSL_CIPHER') ? SS_DATABASE_SSL_CIPHER : 'DHE-RSA-AES256-SHA';
```
tractorcow: Instead of setting the default cipher here, let's leave it up to the DB connection to choose it's own default.

conf/ConfigureFromEnv.php
```diff
@@ -116,6 +124,25 @@
...
+			$databaseConfig['ssl_cert'] = SS_DATABASE_SSL_CERT;
+			$databaseConfig['ssl_ca'] = SS_DATABASE_SSL_CA;
+			$databaseConfig['ssl_cipher'] = defined('SS_DATABASE_SSL_CIPHER') ? SS_DATABASE_SSL_CIPHER : 'DHE-RSA-AES256-SHA';
```
tractorcow: Also here, let's leave it blank if not defined, and let the connector choose it's own default.

model/connect/PDOConnector.php
```diff
@@ -159,6 +159,22 @@ public function connect($parameters, $selectDB = false) {
...
+			if(
+				array_key_exists('ssl_key', $parameters) &&
+				array_key_exists('ssl_cert', $parameters) &&
+				array_key_exists('ssl_ca', $parameters) &&
+				array_key_exists('ssl_cipher', $parameters)) {
+
+					$options[PDO::MYSQL_ATTR_SSL_KEY] = $parameters['ssl_key'];
+					$options[PDO::MYSQL_ATTR_SSL_CERT] =$parameters['ssl_cert'];
+					$options[PDO::MYSQL_ATTR_SSL_CA] = $parameters['ssl_ca'];
+					$options[PDO::MYSQL_ATTR_SSL_CIPHER] = $parameters['ssl_cipher'];
```
tractorcow: Lets make this a config rather than hard coding it.

```php
private static $ssl_cipher_default = 'DHE-RSA-AES256-SHA';
// ...
$options[PDO::MYSQL_ATTR_SSL_CIPHER] = $parameters['ssl_cipher'] ?: $this->config()->ssl_cipher_default;
```

conf/ConfigureFromEnv.php
johndalangin: Understood. The reason why I encapsulated setting the SSL variables to PDO is because I feared I would break something with the other databases. If the setting the variables will be innocuous for other database, I can remove the condition.

conf/ConfigureFromEnv.php
johndalangin: Noted.

dev/install/install.php5
johndalangin: Noted.

model/connect/PDOConnector.php
johndalangin: Understood.

johndalangin:
@tractorcow I'll have the changes ready within the day. Thank you for your feedback.

tractorcow:
That's fine @johndalangin, thanks for the great feature!

johndalangin:
@tractorcow No worries! It's my pleasure to contribute. We've been working with Silverstripe for a couple of years now and we'd love to contribute every change we'd get.

The requested revisions have been pushed. Feel free to take a look and let me know if you have any more comments.

tractorcow:
I'm happy. @dhensby you have any thoughts or shall we just merge it?

johndalangin:
@tractorcow @dhensby Thanks for your feedback! I'll work on the MySQLi connector integration so in a separate issue. Do you have a release schedule for the next minor version? I can work on the SSL MySQLi integration by next week but if you need it faster, I can adjust my schedule.

I'll also do the documentation in a separate issue which will most likely affect `Security`, `Models and Databases`, and `Configuration`.

Let me know if you have concerns.

dhensby:
I'm happy

PR #7238, opened 2017-08-03
ENHANCEMENT Ensure flush destroys temp tinymce files

tractorcow:


tests/php/Forms/HTMLEditor/TinyMCECombinedGeneratorTest.php
```diff
@@ -19,10 +19,17 @@ protected function setUp()
...
-        SSViewer::config()->set('themes', [ SSViewer::DEFAULT_THEME ]);
+        SSViewer::config()->set('themes', [SSViewer::DEFAULT_THEME]);
         TinyMCEConfig::config()->set('base_dir', 'tinymce');
     }
 
+    protected function tearDown()
+    {
+        parent::tearDown();
+        // Flush test configs
+        HTMLEditorConfig::set_config('testconfig', null);
```
flamerohr: don't need to include `null` in the params, it's the default ;)

tractorcow: It's better to be explicit about this. The purpose of the default being null is to bypass the strict typehint on that argument, rather than to allow the method to be called with a single arg.

With this signature
```php
public static function set_config($identifier, HTMLEditorConfig $config)
```

calling `::set_config($name, null)` is a php error. With a default argument of null, it becomes nullable.

flamerohr: oh ok, I wasn't familiar with that

PR #7241, opened 2017-08-03
BUG Fix flushing on live mode

### LINKED ISSUE #7217: SS4-beta1: Unable to flush in live environment
SilverStripe-4-beta1

Flushing via the URL `<website-url>/?flush`,  `<website-url>/?flush=1` or  `<website-url>/?flush=all` doesn't work, instead redirecting the **logged-in administrator** to the `/Security/login?BackURL=` page with the message

```
You're logged in as Default Admin.
[ Log in as someone else]
```

Flushing in the same manner works as expected with `SS_ENVIRONMENT_TYPE="dev"` in `.env`

No errors or warnings in the apache log, and tested on php 5.6.31 & 7.1.7.

This may or not be related (else I can report as a separate issue), when running `sudo -u www-data ./framework/sake / flush=all` on the commandline I get about 200 warnings along the lines of:
```
[2017-07-27 10:52:46] manifestcache-log.WARNING: Failed to save values {"keys":["AddToCampaignHandler_php_57a8bc57e6cbe4f681b527fed47c439c"],"exception":null} []
[2017-07-27 10:52:46] manifestcache-log.WARNING: Failed to save values {"keys":["AddToCampaignHandler_FormAction_php_c1363c50f9e960cbf8619619060ca63b"],"exception":null} []
[2017-07-27 10:52:46] manifestcache-log.WARNING: Failed to save values {"keys":["CampaignAdmin_php_03d12d59f12b174bb7655faa226d51c9"],"exception":null} []
[2017-07-27 10:52:46] manifestcache-log.WARNING: Failed to save values {"keys":["CampaignAdminList_php_8fe6b410624da8ed3b4a30f44c33e16f"],"exception":null} []
[2017-07-27 10:52:46] manifestcache-log.WARNING: Failed to save values {"keys":["Controller_php_dc27224d8eec716b4c660acde6c8faca"],"exception":null} []
[2017-07-27 10:52:46] manifestcache-log.WARNING: Failed to save values {"keys":["DataObjectInterfaceTypeCreator_php_4b252ad9c0bb030acc5db20c7c304e5f"],"exception":null} []
[2017-07-27 10:52:46] manifestcache-log.WARNING: Failed to save values {"keys":["FieldCreator_php_8887cd0a643ba1a90e7d5ee24bd91335"],"exception":null} []
[2017-07-27 10:52:46] manifestcache-log.WARNING: Failed to save values {"keys":["InterfaceTypeCreator_php_ff560fd328824d3ddb82f9ca2841eaa0"],"exception":null} []
[2017-07-27 10:52:46] manifestcache-log.WARNING: Failed to save values {"keys":["Manager_php_d53eb5861f39123fed3778d3c4d9a4d9"],"exception":null} []
[2017-07-27 10:52:46] manifestcache-log.WARNING: Failed to save values {"keys":["MutationCreator_php_2fb9225d673682584bb7bcf42cf05def"],"exception":null} []
[2017-07-27 10:52:46] manifestcache-log.WARNING: Failed to save values {"keys":["OperationResolver_php_2fd7e964309ff64e73a0e2d5a6376e99"],"exception":null} []
...
```

Again, this only happens with `SS_ENVIRONMENT_TYPE="live"`

tractorcow:
Fixes #7217

CC @sminnee @Firesphere @dhensby for review please

This pull request involves some serious hacking to get an injectable ErrorControlChain to support authentication. It works locally, as well as unit tested, but it needs some additional review by those close to the authentication logic.

### DIFF
--- a/_config/requestprocessors.yml
+++ b/_config/requestprocessors.yml
@@ -3,6 +3,7 @@ Name: requestprocessors
 ---
 SilverStripe\Core\Injector\Injector:
   SilverStripe\Control\Director:
+    # Note: Don't add 'class' config here, as it will affect ErrorDirector as well
     properties:
       Middlewares:
         TrustedProxyMiddleware: '%$SilverStripe\Control\Middleware\TrustedProxyMiddleware'
@@ -22,4 +23,5 @@ After:
   - requestprocessors
 ---
 SilverStripe\Core\Injector\Injector:
+  # Note: If Director config changes, take note it will affect this config too
   SilverStripe\Core\Startup\ErrorDirector: '%$SilverStripe\Control\Director'
--- a/_config/requestprocessors.yml
+++ b/_config/requestprocessors.yml
@@ -16,3 +16,10 @@ SilverStripe\Core\Injector\Injector:
   SilverStripe\Control\TrustedProxyMiddleware:
     properties:
       TrustedProxyIPs: "`SS_TRUSTED_PROXY_IPS`"
+---
+Name: errorrequestprocessors
+After:
+  - requestprocessors
+---
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\Core\Startup\ErrorDirector: '%$SilverStripe\Control\Director'
--- a/src/Core/Startup/ErrorControlChainMiddleware.php
+++ b/src/Core/Startup/ErrorControlChainMiddleware.php
@@ -6,9 +6,8 @@ use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\HTTPResponse_Exception;
-use SilverStripe\Core\Application;
 use SilverStripe\Control\Middleware\HTTPMiddleware;
-use SilverStripe\Security\Permission;
+use SilverStripe\Core\Application;
 use SilverStripe\Security\Security;
 
 /**
@@ -87,9 +86,10 @@ class ErrorControlChainMiddleware implements HTTPMiddleware
         // Ensure session is started
         $request->getSession()->init($request);
 
-        // Next, check if we're in dev mode, or the database doesn't have any security data, or we are admin
-        if (Director::isDev() || !Security::database_is_ready() || Permission::check('ADMIN')) {
-            return $reloadToken->reloadWithToken();
+        // Request with ErrorDirector
+        $result = ErrorDirector::singleton()->handleRequestWithToken($request, $reloadToken, $this->getApplication()->getKernel());
+        if ($result) {
+            return $result;
         }
 
         // Fail and redirect the user to the login page
--- a/src/Core/Startup/ErrorDirector.php
+++ b/src/Core/Startup/ErrorDirector.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace SilverStripe\Core\Startup;
+
+use SilverStripe\Control\Director;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Core\Kernel;
+use SilverStripe\Security\Permission;
+use SilverStripe\Security\Security;
+
+/**
+ * Specialised Director class used by ErrorControlChain to handle error and redirect conditions
+ *
+ * @internal This class is experimental API and may change without warning
+ */
+class ErrorDirector extends Director
+{
+    /**
+     * Redirect with token if allowed, or null if not allowed
+     *
+     * @param HTTPRequest $request
+     * @param ParameterConfirmationToken $token
+     * @param Kernel $kernel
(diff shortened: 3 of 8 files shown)


tractorcow:
Ah, I need to fix that base url issue in the test.

src/Core/Startup/ErrorControlChainMiddleware.php
```diff
@@ -88,8 +94,19 @@ protected function safeReloadWithToken(HTTPRequest $request, $reloadToken)
...
+        $reload = function (HTTPRequest $request) use ($reloadToken) {
+            $env = $this->getApplication()->getKernel()->getEnvironment();
+            if ($env === Kernel::DEV || !Security::database_is_ready() || Permission::check('ADMIN')) {
+                return $reloadToken->reloadWithToken();
+            }
+            return null;
+        };
+
+        // Process middleware
+        $chainWithMiddleware = static::create($this->getApplication());
```
tractorcow: This is the part of this logic I'm least happy with... I have to construct a second DI version of ErrrorControlChainMiddlware here.

I'd have used Director::singleton() instead, but callMiddleware isn't public.

src/Core/Startup/ErrorControlChainMiddleware.php
```diff
@@ -88,8 +94,19 @@ protected function safeReloadWithToken(HTTPRequest $request, $reloadToken)
         $request->getSession()->init($request);
 
         // Next, check if we're in dev mode, or the database doesn't have any security data, or we are admin
-        if (Director::isDev() || !Security::database_is_ready() || Permission::check('ADMIN')) {
-            return $reloadToken->reloadWithToken();
+        $reload = function (HTTPRequest $request) use ($reloadToken) {
```
Firesphere: This feels a bit dirty, but seems to work

tractorcow: I think if any part changes, it won't be this. Using a callback as the 'action' is pretty standard for the middleware pattern.

tractorcow:
Rebased with fixed test.

tractorcow:
I think I've addressed the architectural issues. I've shifted the middleware handling to a Director subclass, and set this to inherit the same middleware from the base Director.

src/Core/Startup/ErrorDirector.php
```diff
@@ -0,0 +1,45 @@
...
+use SilverStripe\Control\Director;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Core\Kernel;
+use SilverStripe\Security\Permission;
+use SilverStripe\Security\Security;
+
+/**
+ * Specialised Director class used by ErrorControlChain to handle error and redirect conditions
```
sminnee: Can we mark this whole class with `@internal This is an internal support class and may be changed/removed in minor releases`? I wouldn't want to be supporting people's subclasses of this.

_config/requestprocessors.yml
```diff
@@ -16,3 +16,10 @@ SilverStripe\Core\Injector\Injector:
   SilverStripe\Control\TrustedProxyMiddleware:
     properties:
       TrustedProxyIPs: "`SS_TRUSTED_PROXY_IPS`"
+---
+Name: errorrequestprocessors
+After:
+  - requestprocessors
+---
+SilverStripe\Core\Injector\Injector:
```
sminnee: I don't understand why this is necessary? It appears to disable the use of ErrorDirector?

src/Core/Startup/ErrorControlChainMiddleware.php
```diff
@@ -6,9 +6,8 @@
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\HTTPResponse_Exception;
-use SilverStripe\Core\Application;
 use SilverStripe\Control\Middleware\HTTPMiddleware;
-use SilverStripe\Security\Permission;
+use SilverStripe\Core\Application;
 use SilverStripe\Security\Security;
 
 /**
```
sminnee: Maybe mark this as @internal too?

_config/requestprocessors.yml
tractorcow: It just imports the same dependencies (Middlewares) but doesn't override the class property (director config doesn't specify it).

src/Core/Startup/ErrorControlChainMiddleware.php
tractorcow: Already in beta1, might be too late to @internal it.

src/Core/Startup/ErrorDirector.php
tractorcow: Done.

_config/requestprocessors.yml
tractorcow: I've clarified the intricacies of injector config with additional docs on `class` property.

sminnee:
@unclecheese can you merge?

PR #7246, opened 2017-08-06
BUG Fix install issue with IIS

### LINKED ISSUE #7218: SS4 beta 1 - Installer not working
Hello

I tried to install the SS4 beta 1, but the installer it not working.

Environment:
- Window 10
- IIS 
- PHP 5.6

I am using environment variables for configuration.

Problem 1:
SS_DATABASE_NAME does not actually set the database name in the installer.

Problem 2
SilverStripe\Dev\Install\InstallRequirements::isIIS returns false. So no web.config file is created.

Installer output:
```
Installing SilverStripe...
I am now running through the installation steps (this should take about 30 seconds)

If you receive a fatal error, refresh this page to continue the installation

Setting up 'mysite/_config.php' for use with environment variables...
Setting up ***/mysite/_config.php
Setting up 'mysite/_config/config.yml'
Setting up ***/mysite/_config/config.yml
Building database schema...
Creating default CMS admin account...
Checking that friendly URLs work...
Friendly URLs are not working. This is most likely because a rewrite module isn't configured correctly on your site. You may need to get your web host or server administrator to do this for you:
mod_rewrite or other rewrite module is enabled on your web server
AllowOverride All is set for the directory where SilverStripe is installed
```

Jonatan

tractorcow:
Fixes #7218

### DIFF
--- a/src/Dev/Install/InstallRequirements.php
+++ b/src/Dev/Install/InstallRequirements.php
@@ -142,24 +142,25 @@ class InstallRequirements
      */
     public function isIIS($fromVersion = 7)
     {
-        if (strpos($this->findWebserver(), 'IIS/') === false) {
-            return false;
+        $webserver = $this->findWebserver();
+        if (preg_match('#.*IIS/(?<version>[.\\d]+)$#', $webserver, $matches)) {
+            return version_compare($matches['version'], $fromVersion, '>=');
         }
-        return substr(strstr($this->findWebserver(), '/'), -3, 1) >= $fromVersion;
+        return false;
     }
 
+    /**
+     * @return bool
+     */
     public function isApache()
     {
-        if (strpos($this->findWebserver(), 'Apache') !== false) {
-            return true;
-        } else {
-            return false;
-        }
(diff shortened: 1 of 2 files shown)


src/Dev/Install/InstallRequirements.php
```diff
@@ -142,24 +142,25 @@ public function checkAdminConfig($adminConfig)
      */
     public function isIIS($fromVersion = 7)
     {
-        if (strpos($this->findWebserver(), 'IIS/') === false) {
-            return false;
+        $webserver = $this->findWebserver();
+        if (preg_match('#.*IIS/(?<version>[.\\d]+)$#', $webserver, $matches)) {
+            return version_compare($matches['version'], $fromVersion, '>=');
         }
-        return substr(strstr($this->findWebserver(), '/'), -3, 1) >= $fromVersion;
```
flamerohr: it's like the Y2K bug :D

tractorcow:
Note: I'm not backing up / restoring the $_SERVER global as this is handled via GlobalsTestState already.

PR #7247, opened 2017-08-06
Issue/7181

### LINKED ISSUE #7181: node_threshold_total baked-in as 30 in 3.6
Userland config for SS" config system should allow dev's to configure a default value for the no. nodes able to be shown in a `TreeDropdownField` that suits their application and use-case. Currently, the frmework bakes-in as "30" as the default with no recourse to the YML config system to modify it. (see: `TreeDropdownField.php:342`).

In my use-case it means I cannot show my list of 42 security `Group` records in a `TreeMultiSelect` menu via the advancedworkflow module. The logic bails-out at 30, marks all nodes as "closed" and refused to return _anything_ to the menu.

See the following issue for more detail: https://github.com/symbiote/silverstripe-advancedworkflow/issues/170

PR to follow.

phptek:
FIX: Fixes #7181. Ensures node_threshold_total config value is respected rather than baked-in (and mis-matching) defaults.


phptek:
Just spotted something that wouldn't be too good, closing.

PR #7250, opened 2017-08-07
FIX: Fixes #7181 Use config system for userland config of node_threshold_total

### LINKED ISSUE #7181: node_threshold_total baked-in as 30 in 3.6

phptek:
Substitutes baked-in defaults for x2 methods on `TreeDropdownField` and `Hierarchy` for configurable defaults via a `node_threshold_total` private static.

### DIFF
--- a/forms/TreeDropdownField.php
+++ b/forms/TreeDropdownField.php
@@ -51,6 +51,13 @@ class TreeDropdownField extends FormField {
 	);
 
 	/**
+	 * @config
+	 * @var int
+	 * @see {@link Hierarchy::$node_threshold_total}.
+	 */
+	private static $node_threshold_total = 30;
+
+	/**
 	 * @ignore
 	 */
 	protected $sourceObject, $keyField, $labelField, $filterCallback,
@@ -303,7 +310,8 @@ class TreeDropdownField extends FormField {
 		if ($this->filterCallback || $this->search != "" )
 			$obj->setMarkingFilterFunction(array($this, "filterMarking"));
 
-		$obj->markPartialTree($nodeCountThreshold = 30, $context = null,
+		$nodeCountThreshold = Config::inst()->get('TreeDropdownField', 'node_threshold_total');
+		$obj->markPartialTree($nodeCountThreshold, $context = null,
 			$this->childrenMethod, $this->numChildrenMethod);
 
 		// allow to pass values to be selected within the ajax request
--- a/model/Hierarchy.php
+++ b/model/Hierarchy.php
@@ -218,10 +218,12 @@ class Hierarchy extends DataExtension {
 	 * @param int $nodeCountThreshold See {@link getChildrenAsUL()}
 	 * @return int The actual number of nodes marked.
 	 */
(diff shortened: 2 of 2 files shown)


model/Hierarchy.php
```diff
@@ -218,10 +218,12 @@ public function getChildrenAsUL($attributes = "", $titleEval = '"<li>" . $child-
 	 * @param int $nodeCountThreshold See {@link getChildrenAsUL()}
 	 * @return int The actual number of nodes marked.
 	 */
-	public function markPartialTree($nodeCountThreshold = 30, $context = null,
+	public function markPartialTree($nodeCountThreshold = null, $context = null,
 			$childrenMethod = "AllChildrenIncludingDeleted", $numChildrenMethod = "numChildren") {
 
-		if(!is_numeric($nodeCountThreshold)) $nodeCountThreshold = 30;
+		if(!is_numeric($nodeCountThreshold)) {
+			$nodeCountThreshold = Config::inst()->get('Hierarchy', 'node_threshold_total');
```
tractorcow: Can we use `$this->owner->config()->get('node_threshold_total')` so that we can configure these on a per-object basis?

forms/TreeDropdownField.php
```diff
@@ -303,7 +310,8 @@ public function tree(SS_HTTPRequest $request) {
 		if ($this->filterCallback || $this->search != "" )
 			$obj->setMarkingFilterFunction(array($this, "filterMarking"));
 
-		$obj->markPartialTree($nodeCountThreshold = 30, $context = null,
+		$nodeCountThreshold = Config::inst()->get('TreeDropdownField', 'node_threshold_total');
```
tractorcow: $this->config() is prefferred to config::inst()->get(), if you are making the other change I requested at the same time. :)

model/Hierarchy.php
phptek: I would love to have done, but `Config::inst()->get('Hierarchy', 'node_threshold_total')` is used elsewhere in the same class. If it were to be included, it would be an unrelated change being included in this PR. Are you happy for other refs to it to be changed also?

forms/TreeDropdownField.php
phptek: Happy to do so for `TreeDropdownField`, see related comment on `Hierarchy` though.

model/Hierarchy.php
tractorcow: Ah, in that case, better leave it...

PR #7257, opened 2017-08-08
API Implement cascade_deletes

tractorcow:
Framework component of https://github.com/silverstripe/silverstripe-versioned/issues/23

CMS fixes at https://github.com/silverstripe/silverstripe-cms/pull/1919

### DIFF
--- a/docs/en/02_Developer_Guides/00_Model/02_Relations.md
+++ b/docs/en/02_Developer_Guides/00_Model/02_Relations.md
@@ -406,6 +406,32 @@ the best way to think about it is that the object where the relationship will be
 Product => Categories, the `Product` should contain the `many_many`, because it is much
 more likely that the user will select Categories for a Product than vice-versa.
 
+## Cascading deletions
+
+Relationships between objects can cause cascading deletions, if necessary, through configuration of the
+`cascade_deletes` config on the parent class.
+
+```php
+class ParentObject extends DataObject {
+    private static $has_one = [
+        'Child' => ChildObject::class,
+    ];
+    private static $cascade_deletes = [
+        'Child',
+    ];
+}
+class ChildObject extends DataObject {
+}
+```
+
+In this example, when the Parent object is deleted, the Child specified by the has_one relation will also
+be deleted. Note that all relation types (has_many, many_many, belongs_many_many, belongs_to, and has_one)
+are supported, as are methods that return lists of objects but do not correspond to a physical database relation.
+
+If your object is versioned, cascade_deletes will also act as "cascade unpublish", such that any unpublish
+on a parent object will trigger unpublish on the child, similarly to how `owns` causes triggered publishing.
+See the [versioning docs](/developer_guides/versioning) for more information on ownership.
+
 ## Adding relations
 
 Adding new items to a relations works the same, regardless if you're editing a **has_many** or a **many_many**. They are 
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -267,6 +267,17 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
     protected $unsavedRelations;
 
     /**
+     * List of relations that should be cascade deleted, similar to `owns`
+     * Note: This will trigger delete on many_many objects, not only the mapping table.
+     * For many_many through you can specify the components you want to delete separately
+     * (many_many or has_many sub-component)
+     *
+     * @config
+     * @var array
+     */
+    private static $cascade_deletes = [];
+
+    /**
      * Get schema object
      *
      * @return DataObjectSchema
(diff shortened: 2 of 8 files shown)


tractorcow:
Looks like I need to take a double check of those CMS tests.

docs/en/02_Developer_Guides/00_Model/02_Relations.md
```diff
@@ -406,6 +406,31 @@ the best way to think about it is that the object where the relationship will be
...
+    private static $cascade_deletes = [
+        'Child',
+    ];
+}
+class ChildObject extends DataObject {
+}
+```
+
+In this example, when the Parent object is deleted, the Child specified by the has_one relation will also
+be deleted.
```
flamerohr: Does this also apply to `many_many`, `belongs_many_many`, `has_many` and `many_many through` as well?
It seems to according to the docblock in `DataObject`
Perhaps add a note about those types of relationships here to be consistent.

src/ORM/DataObject.php
```diff
@@ -1044,6 +1055,23 @@ protected function onAfterWrite()
     }
 
     /**
+     * Find all objects that will be cascade deleted if this object is deleted
+     *
+     * Notes:
+     *   - Objects will only be searched in the same stage as the given record.
```
flamerohr: Add note:
> If this Object is versioned

src/ORM/DataObject.php
```diff
@@ -3680,4 +3714,103 @@ public function setJoin(DataObject $object, $alias = null)
         }
         return $this;
     }
+
+    /**
+     * Find objects in the given relationships, merging them into the given list
+     *
+     * @param string $source Config property to extract relationships from
+     * @param bool $recursive True if recursive
+     * @param ArrayList $list Optional list to add items to
```
flamerohr: is it more accurate to say this?
> list to add to for all items found

Maybe a note that this is used mainly to easily enable recursion?

src/ORM/DataObject.php
```diff
@@ -3680,4 +3714,103 @@ public function setJoin(DataObject $object, $alias = null)
...
+
+        /** @var DataObject $item */
+        foreach ($items as $item) {
+            $this->mergeRelatedObject($list, $added, $item);
+        }
+        return $added;
+    }
+
+    /**
+     * Merge single object into a list
```
flamerohr: Missing:
> if object already exists, it will not be added again

src/ORM/DataObject.php
```diff
@@ -3680,4 +3714,103 @@ public function setJoin(DataObject $object, $alias = null)
...
+     *
+     * @param ArrayList $list Global list. Object will not be added if already added to this list.
+     * @param ArrayList $added Additional list to insert into
+     * @param DataObject $item Item to add
+     */
+    public function mergeRelatedObject($list, $added, $item)
```
flamerohr: should this method be public? seemed to be quite a special case method

src/ORM/DataObject.php
```diff
@@ -3680,4 +3714,103 @@ public function setJoin(DataObject $object, $alias = null)
...
+
+    /**
+     * Helper method to merge owned/owning items into a list.
+     * Items already present in the list will be skipped.
+     *
+     * @param ArrayList $list Items to merge into
+     * @param mixed $items List of new items to merge
+     * @return ArrayList List of all newly added items that did not already exist in $list
+     */
+    public function mergeRelatedObjects($list, $items)
```
flamerohr: should this method be public?

src/ORM/DataObject.php
```diff
@@ -3680,4 +3714,103 @@ public function setJoin(DataObject $object, $alias = null)
...
+
+    /**
+     * Find objects in the given relationships, merging them into the given list
+     *
+     * @param string $source Config property to extract relationships from
+     * @param bool $recursive True if recursive
+     * @param ArrayList $list Optional list to add items to
+     * @return ArrayList The list
+     */
+    public function findRelatedObjects($source, $recursive = true, $list = null)
```
flamerohr: I'm not sure if this method needs to be public?

tractorcow: It needs to be public because Versioned extension calls it.

src/ORM/DataObject.php
tractorcow: As above, Versioned extension needs to call it. It used to be protected and a member of Versioned, but it needs to be exposed now.

src/ORM/DataObject.php
tractorcow: Same as above.

tractorcow: Ok, I think I can make just this one method protected, but not the others.

flamerohr: Fair enough, I'm probably too critical about what's public, only half way through the Versioned pull request

tractorcow:
Updated based on feedback.

src/ORM/DataObject.php
flamerohr: Just been through the other pull request, and yeah I agree that only `mergeRelatedObject()` could be made protected.

PR #7261, opened 2017-08-10
Remove `use_gzip` from HTMLEditorField since it's been removed by TinyMCE codebase

phalkunz:
Original part of https://github.com/silverstripe/silverstripe-admin/issues/158

FYI: TinyMCE has removed it https://github.com/tinymce/tinymce-docs/pull/460

### DIFF
--- a/src/Forms/HTMLEditor/HTMLEditorField.php
+++ b/src/Forms/HTMLEditor/HTMLEditorField.php
@@ -29,14 +29,6 @@ class HTMLEditorField extends TextareaField
     protected $schemaComponent = 'HtmlEditorField';
 
     /**
-     * Use TinyMCE's GZIP compressor
-     *
-     * @config
-     * @var bool
-     */
-    private static $use_gzip = true;
-
-    /**
      * @config
      * @var string Default alignment for Images and Media. Options: leftAlone|center|left|right
      */
--- a/docs/en/04_Changelogs/4.5.0.md
+++ b/docs/en/04_Changelogs/4.5.0.md
@@ -2,9 +2,11 @@
 
 ## Overview {#overview}
 
-TBC.
+ * Removed `use_gzip` option on `HtmlEditorField` which used to compress the rich text editor dependency. 
+   No longer required since compression is performed as part of the CMS build automatically.
+   See (#832)(https://github.com/silverstripe/silverstripe-admin/issues/832)
 
 ## Deprecation
 
-* `PasswordValidator` methods `minLength`, `characterStrength`, and `checkHistoricalPasswords` are now deprecated from
+ * `PasswordValidator` methods `minLength`, `characterStrength`, and `checkHistoricalPasswords` are now deprecated from
   4.5.0 onwards (previously 5.0).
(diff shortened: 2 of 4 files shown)


tractorcow:
I'm not sure we can remove this in beta. Removal of API / features is now 5.0 at the earliest. In which case, we should remove that entire class (as it's redundant).

flamerohr:
Cleaned up rebase to master :)

dhensby:
> I'm not sure we can remove this in beta. Removal of API / features is now 5.0 at the earliest. In which case, we should remove that entire class (as it's redundant).

if this is a defunct API which has no impact what-so-ever then I feel we should remove it despite being in beta.

The beta should give us some flexibility around this and keeping a useless API around for 3 years feels like dogma over reason.

tractorcow:
@dhensby if you get another core committer to second that then we can remove it in 4. I will admit to being a bit dogmatic about semver. :)

dhensby:
@tractorcow I never got round to reading this notification until now and I guess now we're in RC phase it's going to be even harder to get this removed from 4.x ;)

Are `private statics` an API at all? I still think this can be removed in the RC...

@chillu - what do you think to this?

tractorcow:
> Are private statics an API at all? I still think this can be removed in the RC...

Yes it's an API. I'd feel better about removing this on master branch.

sminnee:
If this is broken in SS4 then I think removing it is a bugfix, not an API change. I'd :+1: to removing it and the class. The config option can still be set and perform a no-op if the private static is removed so that's fine.

If we really want to avoid deleting the class, then I would just clearing out the classes content to only throw a deprecation warning saying "this doesn't work" and deferring to default behaviour.

tractorcow:
Is there an actual verified report that it's broken? I understood this as a removal of an undesirable feature.

sminnee:
Oh, right, I just assumed. If it's still working then I would bump to master too.

dhensby:
The feature has been removed by tinymce, so whilst our class does *something* it's not usable for its intended purpose.

tractorcow:
> The feature has been removed by tinymce, so whilst our class does something it's not usable for its intended purpose.

Are you talking about tiny_mce_gzip.php? It's still in thirdparty dir, and it is usable for it's intended purpose. Can you please elaborate on "not usable"?

dhensby:
See https://github.com/tinymce/tinymce-docs/pull/460

tractorcow:
I've seen that; That's not retroactively removing / rendering inoperable the php file from every current tinymce installation out there, including the one bundled with our last beta.

maxime-rainville:
Related to https://github.com/silverstripe/silverstripe-admin/issues/832

src/Forms/HTMLEditor/HTMLEditorField.php
```diff
@@ -28,14 +28,6 @@ class HTMLEditorField extends TextareaField
 
     protected $schemaComponent = 'HtmlEditorField';
 
-    /**
-     * Use TinyMCE's GZIP compressor
-     *
-     * @config
-     * @var bool
-     */
-    private static $use_gzip = true;
```
maxime-rainville: I think removing this line is OK as there's no way to access it directly and setting a config flag for a value that doesn't exist doesn't cause an error.

chillu:
Tested locally, added changelog entry, and merged.

PR #7270, opened 2017-08-13
FIX Files without extensions (folders) do not have a trailing period added

### LINKED ISSUE #7265: BUG: weird dot at the end of Folder Name
I tried to add a bunch of `test` folders to the Silverstripe demo site and they got a "dot" added to the end of their name: 

image

robbieaverill:
Fixes #7265

The issue is that `Folder::setName` calls `File::setName`, which assumes there will always be an extension and so returns `original-2.` in the format of `[original]-[unique].[extension]`, which is invalid for folders, and files without an extension in principle.

### DIFF
--- a/tests/filesystem/FileTest.php
+++ b/tests/filesystem/FileTest.php
@@ -187,6 +187,30 @@ class FileTest extends SapphireTest {
 		}
 	}
 
+	/**
+	 * Uses fixtures Folder.folder1 and File.setfromname
+	 * @dataProvider setNameFileProvider
+	 */
+	public function testSetNameAddsUniqueSuffixWhenFilenameAlreadyExists($name, $expected)
+	{
+		$duplicate = new Folder;
+		$duplicate->setName($name);
+		$duplicate->write();
+
+		$this->assertSame($expected, $duplicate->Name);
+	}
+
+	/**
+	 * @return array[]
+	 */
+	public function setNameFileProvider()
+	{
+		return array(
+			array('FileTest-folder1', 'FileTest-folder1-2'),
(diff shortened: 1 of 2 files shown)


tests/filesystem/FileTest.php
```diff
@@ -187,6 +187,30 @@ public function testSetNameWithInvalidExtensionDoesntChangeFilesystem() {
...
+		$duplicate->setName($name);
+		$duplicate->write();
+
+		$this->assertSame($expected, $duplicate->Name);
+	}
+
+	/**
+	 * @return array[]
+	 */
+	public function setNameFileProvider()
```
dhensby: non-test methods should be `protected` or `private`

robbieaverill: It's a provider - they need to be public don't they?

robbieaverill:
It's probably affecting 3.5 as well - want me to switch and rebase?

tests/filesystem/FileTest.php
dhensby: I don't believe so.

dhensby:
> It's probably affecting 3.5 as well - want me to switch and rebase?

yes please

tests/filesystem/FileTest.php
robbieaverill: [They do](https://phpunit.de/manual/5.7/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers) in PHPUnit 5.x at least, may have changed in 6.x

robbieaverill:
@dhensby done, thanks :)

tractorcow:
Test failures are due to behat / precise.

PR #7281, opened 2017-08-14
NEW: Test on php 7.2 

sminnee:
**UPDATE:*  It turns out that nightly is now running PHP 7.3 and there's a separate PHP 7.2 config.

It’s important that SS4 runs on PHP 7.2, and PHP 7.2 isn’t released
until November, which we expect will be later than SS4 stable. To 
maximise the chance that SS 4.0.0 works with PHP 7.2 I’d think we should
start testing on the PHP nightly, and change this to a 7.2 build once
it’s released.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,11 @@ matrix:
         - DB=MYSQL
         - PDO=1
         - PHPUNIT_COVERAGE_TEST=framework
+    - php: 7.2
+      env:
+        - DB=MYSQL
+        - PDO=1
+        - PHPUNIT_TEST=framework
     - php: 7.0
       env:
        - DB=MYSQL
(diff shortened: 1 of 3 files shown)


sminnee:
I don't think we should add nightlies until PHP 7.3 is a little closer to release. It's an extra run in an already heavy build-burden, there are chances that nightlies failures might introduce false-positive build failures, and there's little current value in explicitly offering PHP 7.3 support.

My recommendation would be to add a PHP 7.3 build target once it's in beta. At that time we can probably drop 7.0 or 7.1 as well, reasoning that if something works in 5.6 and 7.1 then it probably works in 7.0 as well.

I'd expect PHP 7.2 beta to be about 9-12 months away and I don't think we need to keep an open ticket for this.

dhensby:
I  was thinking as an allowed failure with fast-fail on for travis, but it's still a heavy burden and it shouldn't block merge

PR #7285, opened 2017-08-15
Enhancement Add a path option for the schema data,  …

flamerohr:
so a full tree is not required for the full path data


tractorcow:
Merge on green.

PR #7287, opened 2017-08-16
BUG Fix issue with multiple editors breaking plugins

tractorcow:
For the attention of @mikenz . :)


src/Forms/HTMLEditor/TinyMCECombinedGenerator.php
```diff
@@ -137,18 +136,38 @@ public function generateContent(TinyMCEConfig $config)
...
-        // Set base URL for where tinymce is loaded from
-        $buffer = "var tinyMCEPreInit={base:'" . Convert::raw2js($tinymceDir) . "',suffix:'.min'};\n";
+        $libContent = $this->getFileContents(Director::baseFolder() . '/' . $tinymceDir . '/tinymce.min.js');
+
+        // Register vars for config
+        $baseDirJS = Convert::raw2js(Director::absoluteBaseURL());
+        $buffer = [
+            <<<SCRIPT
+var baseTag = window.document.getElementsByTagName('base');
+var baseURL = baseTag.length ? baseTag[0].baseURI : '$baseDirJS';
```
flamerohr: short-circuit these variables, so you don't need to compute them again if they're declared more than once.

e.g. `var baseTag = baseTag || window.document.getElementsByTagName('base');`

PR #7288, opened 2017-08-16
FIX Don't update LastEdited on skipped writes

### LINKED ISSUE #3642: DataObject->write() without changes does not update LastEdited in the Database, but does on the instance
write() updates LastEdited on the object itself, but only writes it to database if the DataObject has changes.

```
$foo = Foo::get()->first();
echo $foo->LastEdited; // 2014-11-14 18:46:14
$foo->write();
echo $foo->LastEdited; // 2014-11-14 22:00:32
$foo = Foo::get()->first();
echo $foo->LastEdited; // 2014-11-14 18:46:14
```

dhensby:
fixes #3642

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -1377,6 +1377,12 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 		// Check changes exist, abort if there are none
 		$hasChanges = $this->updateChanges($isNewRecord);
 		if($hasChanges || $forceWrite || $isNewRecord) {
+
+			// Ensure Created and LastEdited are populated
+			if(!isset($this->record['Created'])) {
+				$this->record['Created'] = $now;
+			}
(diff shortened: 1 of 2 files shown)


dhensby:
In my testing of this I've noticed that the `FixtureBlueprintTest` has a test that tests a feature that implies the setting of `Created` in fixture files will be persisted in the DB (but they won't).

The [`FixtureBlueprintTest::testCreateWithClassAncestry`](https://github.com/silverstripe/silverstripe-framework/blob/69451790d6ffdda89c6222796b188a7f500dd915/tests/dev/FixtureBlueprintTest.php#L216-L237) checks that a fixture written with data which specifies a `Created` field will report that field back - however this checks the current object in memory and not the data written to the DB or that is pulled out when a `DataObject::get()` is performed...

This leads to a false positive implying the `Created` date can be set in a fixture when it cannot.

PR #7291, opened 2017-08-21
Fix trim accept header parts

flamerohr:


tractorcow:
Can be merged on green.

PR #7292, opened 2017-08-21
FIX: Prevent repeated lookup of obj.dependencies by Injector

sminnee:
This unnecessary repeated call to Injector slows down the construction
of frequently instantiated classes.

On admin/pages, this reduced execution from 1.67s to 1.56s, and it
reduced the impact of having an extension added to DBField by 33%
(from 100ms to 67ms)


tractorcow:
Looks good to me!

tractorcow:
Parent issue https://github.com/silverstripe/silverstripe-admin/issues/199

tractorcow:
I've pushed up a linting fix. Will merge once it's green.

PR #7293, opened 2017-08-21
Major performance improvements

tractorcow:
Based on results of config profiling at https://github.com/silverstripe/silverstripe-admin/issues/199 I've made the following improvements to cut down uncessary config API calls:

 - i18n.default_locale was invoked a huge amount of times. This change reduces overall page load by about 20ms in the CMS
 - Extension construction being done excessively caused a massive amount of performance degradation, especially with 4 times the DBField's being constructed due to default_cast changes
 - Stopped merging in config from Extension and DataExtension which provide no config worth merging in.

### DIFF
--- a/src/Control/Director.php
+++ b/src/Control/Director.php
@@ -108,7 +108,6 @@ class Director implements TemplateGlobalProvider
 
     public function __construct()
     {
-        $this->constructExtensions();
     }
 
     /**
--- a/src/Core/Extensible.php
+++ b/src/Core/Extensible.php
@@ -6,29 +6,18 @@ use InvalidArgumentException;
 use SilverStripe\Control\RequestHandler;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Dev\Deprecation;
+use SilverStripe\ORM\DataExtension;
+use SilverStripe\ORM\DataObject;
 use SilverStripe\View\ViewableData;
 
 /**
  * Allows an object to have extensions applied to it.
- *
- * Bootstrap by calling $this->constructExtensions() in your class constructor.
- *
- * Requires CustomMethods trait
  */
 trait Extensible
 {
     use CustomMethods {
-        getExtraMethodConfig as getCustomMethodsConfig;
-    }
-
-    protected function getExtraMethodConfig($method)
-    {
-        $config = $this->getCustomMethodsConfig($method);
-        // Ensure extension instances are populated before being accessed
-        if (isset($config['property']) && $config['property'] === 'extension_instances') {
-            $this->getExtensionInstances();
-        }
-        return $config;
+        defineMethods as defineMethodsCustom;
     }
 
     /**
@@ -51,8 +40,6 @@ trait Extensible
      * @config
      */
     private static $extensions = [];
-
-    private static $classes_constructed = array();
 
     /**
      * Classes that cannot be extended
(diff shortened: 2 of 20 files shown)


tractorcow:
load times of admin/pages/

Before:

image

After:

image

src/Core/Extensible.php
```diff
@@ -18,7 +17,18 @@
  */
 trait Extensible
 {
-    use CustomMethods;
+    use CustomMethods {
```
sminnee: The callSetOwnerFirst option of CustomMethods is already tightly coupled to extensions; you could potentially just put this logic there and keep the call-stack from getting too much deeper.

tractorcow: The problem is that the code would error before it got to this point.

This line is called to get $obj and would fail.

```php
$obj = $config['index'] !== null ?
                    $this->{$config['property']}[$config['index']] :
                    $this->{$config['property']};
```

$config['property'] would be `extension_instances` so we need to force that property to be populated by that point.

tractorcow: Alternatively, I could put it into getExtraMethodConfig() instead of the __call override.

tractorcow: Ok pushed. :D

sminnee: I was suggesting that you edit CustomMethods::call:

```php
        switch (true) {
            case isset($config['property']) : {
+                if (!empty($config['warmPropertyMethod'])) {
+                   if (!method_exists($this, $method)) {
+                      $methodName = $config['warmPropertyMethod'];
+                      $this->$methodName();
+                   }
+                }

                $obj = $config['index'] !== null ?
                    $this->{$config['property']}[$config['index']] :
                    $this->{$config['property']};

                if ($obj) {
                    if (!empty($config['callSetOwnerFirst'])) {
                        /** @var Extension $obj */
                        $obj->setOwner($this);
                    }
                    $retVal = call_user_func_array(array($obj, $method), $arguments);
                    if (!empty($config['callSetOwnerFirst'])) {
                        /** @var Extension $obj */
                        $obj->clearOwner();
                    }
                    return $retVal;
                }
```

And then set warmPropertyMethod in CustomMethod::addMethodsFrom() the same way that callSetOwnerFirst is set.

It's not as loosely coupled but reducing the method-call count would be good in such a critical loop

sminnee: The other thing that might be worth doing (not sure of perf impact) would be replacing the `empty` calls with a basic if, and ensure that these values are set to true/false in the config rather than omitted.

sminnee: OK that last point is a no-go — `empty()` is very fast.

sminnee: It's not a complete solution – adding warmPropertyMethod might be better – but this shaved about 40ms https://gist.github.com/f69bad13d28e1e1ca2daf3161e4142e9

The key IMO is avoiding an additional method call.

tractorcow: Unfortunately getExtensionInstances isn't a method on CustomMethods. It'll need refactoring.

sminnee: Yeah — what did you think about including a "warmPropertyMethod" option that can be set to "getExtensionInstances" on an Extensible object?

sminnee: However, `callSetOwnerFirst` already requires an Extensible object so the CustomMethod code already depends on Extensible being applied.

tractorcow:
@sminnee I've re-worked both extensions and now neither defineMethods() nor constructExtensions() are invoked at all on object construction. These are now safely lazy-loaded.

As a result I can drop the objects_constructed private property, and rely on lazy loading to bootstrap these objects.

src/Core/Extensible.php
tractorcow: I've shifted from a warming-based approach to lazy-loading for methods. We should deprecate callSetOwner in favour of the new extension method registration in my most recent commit.

sminnee:
All good from me.

tractorcow:
It's green, can it be merged?

src/Control/Director.php
```diff
@@ -108,7 +108,6 @@ class Director implements TemplateGlobalProvider
 
     public function __construct()
     {
-        $this->constructExtensions();
```
robbieaverill: Does this change mean that Director is no longer extensible, or that is doesn't need to call constructExtensions any more?

tractorcow: It doesn't need it.

PR #7300, opened 2017-08-22
BUG Capture errors after a reload token redirect to login url

### LINKED ISSUE #7289: BUG Some error conditions may forbid flushing on live mode
E.g. when requesting `url/flush=all`, the middleware may redirect to `Security/login` to authenticate on live mode.

However, occasionally an error may occur not on the first request, but on the second. In this case, perhaps ErrorControlChainMiddleware needs to be aware of nested parameters in the BackURL querystring arg.

I.e. if an error occurrs on `Security/login?BackURL=url%2F%3Fflush%3D1` then it should trigger the flush.

PR: https://github.com/silverstripe/silverstripe-framework/pull/7300

tractorcow:
Fixes #7289

### DIFF
--- a/src/Core/Startup/ParameterConfirmationToken.php
+++ b/src/Core/Startup/ParameterConfirmationToken.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\Core\Startup;
 
+use function GuzzleHttp\Psr7\parse_query;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
@@ -34,11 +35,18 @@ class ParameterConfirmationToken
     protected $request = null;
 
     /**
-     * The parameter given
+     * The parameter given in the main request
      *
      * @var string|null The string value, or null if not provided
      */
     protected $parameter = null;
+
+    /**
+     * The parameter given in the backURL
+     *
+     * @var string|null
+     */
+    protected $parameterBackURL = null;
 
     /**
      * The validated and checked token for this parameter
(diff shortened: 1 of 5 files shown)


src/Core/Startup/ParameterConfirmationToken.php
```diff
@@ -136,6 +168,16 @@ public function parameterProvided()
     }
 
     /**
+     * Is the parmeter requested in a BackURL param?
+     *
+     * @return bool
+     */
+    public function parameterProvidedBackURL()
```
flamerohr: This naming is a bit weird, `hasParameterBackURL()` or `hasBackURL()` would be more suited?

Or if you'd like stick to "Provided" perhaps `parameterBackURLProvided()`?

src/Core/Startup/ParameterConfirmationToken.php
```diff
@@ -157,6 +199,18 @@ public function reloadRequired()
     }
 
     /**
+     * Check if this token is provided either in the backurl, or directly,
+     * but without a token
+     *
+     * @return bool
+     */
+    public function reloadOnError()
```
flamerohr: the name reads like a command to action something rather than a checker, `shouldReloadOnError` perhaps?

src/Core/Startup/ParameterConfirmationToken.php
tractorcow: I've renamed to `existsInReferer`

src/Core/Startup/ParameterConfirmationToken.php
tractorcow: This method should match `reloadRequired`, so I've renamed to `reloadRequiredIfError` to be more explicit.

tractorcow:
@flamerohr I've changed constants.php to prefer SS_BASE_URL instead of using the SCRIPT_FILENAME check, as it's not the least bit reliable on CLI.

What I've discovered is that running this on `phpdbg` gives the wrong BASE_URL at all.

flamerohr:
Merging, travis is green but github hasn't updated

PR #7301, opened 2017-08-23
ENHANCEMENT Add HTTPRequest::inst() to simplify usercode access to current request

tractorcow:
At request of @zanderwar

### DIFF
--- a/src/Control/HTTPRequest.php
+++ b/src/Control/HTTPRequest.php
@@ -6,6 +6,7 @@ use ArrayAccess;
 use BadMethodCallException;
 use InvalidArgumentException;
 use SilverStripe\Core\ClassInfo;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\ORM\ArrayLib;
 
 /**
@@ -895,5 +896,20 @@ class HTTPRequest implements ArrayAccess
     {
         $this->session = $session;
         return $this;
+    }
+
+    /**
+     * Return injector-registered current request, if one exists
+     *
+     * @return static
+     */
+    public static function inst()
+    {
+        // Ensure we only use a registered HTTPRequest and don't
+        // incidentally construct a singleton
+        if (Injector::inst()->has(static::class)) {
+            return Injector::inst()->get(static::class);
+        }
+        return null;
     }
 }


tractorcow:
Requesting @chillu review too to make sure this fits with our intended patterns.

We can then replace a lot of the Controller::curr() references, plus un-typehinted Injector calls.

sminnee:
This seems okay but I'd still recommend that people use `$this->request` wherever possible...

zanderwar:
@sminnee The concern was with the length of the code required to modify session variables, outside the scope of `$this` where request is available

```
Session::set('ACTIVE_STORE', $store->ID);
// vs
Controller::curr()->getRequest()->getSession()->set('ACTIVE_STORE', $store->ID);
```

sminnee:
Yeah. I'd support this addition but needing to change the session when a request isn't available feels like a bit of a code smell - as in, if there's no request available why are you mucking with the session?

I still think that HTTPRequest::inst() is a pragmatic addition for SS4, but it would be good to get more insight as to what kind of things touch session outside of controllers.

chillu:
This is the alternative access path documented in `02_Sessions.md`:

```
$request = Injector::inst()->get(HTTPRequest::class)->getSession();
$session = $request->getSession();
```

I think that's a feasible amount of code for the case where you're following bad practices and relying on global session access outside of a controller context. You don't need the lengthy `Controller::curr()` call for this. My concern with creating a shortcut here is that we're coupling `HTTPRequest` to the `Injector` class. We're trying to reduce those couplings, not increase them, right?

sminnee:
The bug with that approach is that it creates a blank object if none has been loaded, which is why that "has" is needed.

chillu:
Some context from the community Slack:

image

chillu:
If you are putting this code into an execution context that might not even have a HTTPRequest (e.g. when run on CLI, or very early during boot) then it's quite a fragile piece of logic to start with. Early boot session setting could happen through HTTP middleware or HTTP request filters, but both of these have the request available to retrieve a session (no need for globals). If you're using session on CLI, you're doing it wrong and we shouldn't support you with shortcuts.

tractorcow:
@chillu we haven't made the decision that Session object isn't available in CLI. I agree that you shouldn't use a session when one isn't available, but this code is not intended for those times, but for code that exists within the normal request handling loop. This isn't a CLI-specific shortcut.

chillu:
> but this code is not intended for those times, but for code that exists within the normal request handling loop

Where you should be able to use `$request->getSession()` or `$this->request->getSession()`. If request isn't available, you're likely writing bad code. But I digress, I feel more strongly about the resulting bad code than about the fact that we'll make it easier to write with this shortcut. If @sminnee and you think it's a good idea, I'm happy to be overruled with this merge :)

sminnee:
FYI — I had a DM chat with @zanderwar afterwards and identified that the best solution to his problem was to create an HTTPMiddleware.

I'd be okay with deferring this improvement for a bit to see if it's easy enough for people to find "proper" workarounds. However, if people are putting some effort into "doing this properly" and still find it necessary to statically reference the HTTPRequest, then I think we need to make it a reasonable experience rather than scold developers. I don't think we have enough data yet.

The other solution to this problem would be to have a "get-but-don't-automatically-create" accessor from Injector, or something in the yaml to block its automatic creation. Then just using Injector::get() would be fine.

sminnee:
@adrexia as the other person who :+1:d this card, can you outline the places where you find it necessary to statically reference the request? Feel free to DM On silverstripe-users if it's too sensitive for public channels. ;)

tractorcow:
> Where you should be able to use $request->getSession() or $this->request->getSession(). If request isn't available, you're likely writing bad code. But I digress, I feel more strongly about the resulting bad code than about the fact that we'll make it easier to write with this shortcut. If @sminnee and you think it's a good idea, I'm happy to be overruled with this merge :)

We already have bad code; See entire of Director.php.

adrexia:
I +1ed because I was around when @zanderwar had the issue on slack and this changed fixed it for him. If there are other solutions to that problem, then I have no particular stake in this. :)

flamerohr:
I'm for deferring this to later, depending on what people's feedback are on needing something like this.
This does take us a step back from not relying on a global session/request access, and doesn't seem like a clear need yet.

dhensby:
I'm not for this approach and it seems the consensus is forming around putting this off if middleware solves the problem

tractorcow:
Middleware doesn't solve the problem; As I've stated, you shouldn't need this code in middleware (as it's not necessary there).

However, if you look at all of the Form session state logic, that occurs at a much higher level than middleware, and is baked into the static state provided by Controller::curr(). The point of this PR is to avoid the reliance on this API.

Instead, what we will have to do is inline Injector::has() / Injector::get() everywhere.

tractorcow:
Here's the code in question.

```php
    /**
     * Helper to get current request for this form
     *
     * @return HTTPRequest
     */
    protected function getRequest()
    {
        // Check if current request handler has a request object
        $controller = $this->getController();
        if ($controller && !($controller->getRequest() instanceof NullHTTPRequest)) {
            return $controller->getRequest();
        }
        // Fall back to current controller
        if (Controller::has_curr() && !(Controller::curr()->getRequest() instanceof NullHTTPRequest)) {
            return Controller::curr()->getRequest();
        }
        return null;
    }
```

dhensby:
Ok - sorry for the misunderstanding.

I do think we should avoid adding an `inst` method to `HTTPRequest` though. I think we should be aiming to decouple from our `HTTPRequest` objects not tighten it.

Something SilverStripe doesn't really have at the moment is helper methods or utilities that could more easily solve this. Some global (or namespaced) method like `get_current_request()` which would do this repeated code for us would be an alternative approach

sminnee:
-100 on helper methods. They create an arbitrary set of different names for things, they aren't compatible with an autoloader.

::inst() isn't great but it's a step above helper methods.

Making injector::inst()->get(HTTPRequest::class) reliable and limiting the number of places where it's actually necessary gets my vote.

zanderwar:
Not a fan of injector::inst()->get(HTTPRequest::class) to be honest. Only cause an IDE won't know what that method returns and lose autocomplete without adding /** @var HTTPRequest $myrequest **/

tractorcow:
Can't we make the existing ::singleton() support nullable service names? It's the preferred type-safe variant of `singelton()` / `Injector::inst()->get()`

tractorcow:
Closing in favour of https://github.com/silverstripe/silverstripe-framework/issues/7316

@sminnee I believe you've suggested a possible fix to framework which would implement this. Could you please find the code you were suggesting and add it as a link?

chillu:
> Not a fan of injector::inst()->get(HTTPRequest::class) to be honest. Only cause an IDE won't know what that method returns and lose autocomplete

@zanderwar That's going to be an issue for every usage of `Injector`, and can be solved by using dependency injection on your own classes as much as possible (since you can indicate the `class` or `interface` on your constructor arguments or class attributes.

tractorcow:
Or `TheClass::singleton()` which is my plan :)

PR #7302, opened 2017-08-23
API Make FormField::hasClass return a boolean instead of an int

robbieaverill:
It's always bothered me that this method returns an integer. If nobody has a concern with this change being made, this PR makes it return a boolean.

### DIFF
--- a/src/Forms/FormField.php
+++ b/src/Forms/FormField.php
@@ -1247,17 +1247,16 @@ class FormField extends RequestHandler
     }
 
     /**
+     * Returns whether the current field has the given class added
+     *
      * @param string $class
      *
-     * @return int
+     * @return bool
      */
     public function hasClass($class)
     {
-        $patten = '/' . strtolower($class) . '/i';
-
-        $subject = strtolower(static::class . ' ' . $this->extraClass());
-
-        return preg_match($patten, $subject);
+        $classes = explode(' ', strtolower($this->extraClass()));
+        return in_array(strtolower(trim($class)), $classes);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Forms/FormField.php
```diff
@@ -1247,17 +1247,19 @@ public function transform(FormTransformation $transformation)
...
+     * Returns whether the current field has the given class added
+     *
      * @param string $class
      *
-     * @return int
+     * @return bool
      */
     public function hasClass($class)
     {
         $patten = '/' . strtolower($class) . '/i';
```
tractorcow: I'm surprised a method this bad still exists; Not that I'm blaming you in any way @robbieaverill :).

I notice it assumes `magic-bananas` contains the class `banana`. =/ Easier to split on whitespace and just use in_array, wouldn't it?

The `static::class` exception is also super wrong and out of date.

robbieaverill: Good time to update! Coming shortly...

tractorcow:
Can be merged on green by anyone with privileges. :P

tests/php/Forms/FormFieldTest.php
```diff
@@ -373,4 +373,15 @@ public function testGetSchemaStateWithFormValidation()
...
     }
+
+    public function testHasClass()
+    {
+        $field = new FormField('Test');
+        $field->addExtraClass('foo BAr cool-banana');
+
+        $this->assertTrue($field->hasClass('foo'));
+        $this->assertTrue($field->hasClass('bAr'));
+        $this->assertFalse($field->hasClass('banana'));
```
tractorcow: Yes, we have no banana.

PR #7311, opened 2017-08-25
FIX: Removed unnecessary database_is_ready call.

sminnee:
This shaves about 45ms from every request (PHP 7.1 on a 2013 rMBP), 
cutting down execution time of a “hello world” controller by about 33%.

database_is_ready is still used in dev/build and ?flush=1 to stop people
from people bypassing security by DOSing the database or otherwise
forcing a DatabaseException


tractorcow:
Much better!

PR #7312, opened 2017-08-25
[ENHANCEMENT] Add the ability to redirect a user to a custom page

Firesphere:
User currently is either directed to the backurl or the default login url.
This adds the option of setting a config on Security redirecting the user to a custom page

### DIFF
--- a/src/Security/MemberAuthenticator/ChangePasswordHandler.php
+++ b/src/Security/MemberAuthenticator/ChangePasswordHandler.php
@@ -3,12 +3,15 @@
 
 namespace SilverStripe\Security\MemberAuthenticator;
 
+use Psr\Container\NotFoundExceptionInterface;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\RequestHandler;
+use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\ORM\FieldType\DBDatetime;
 use SilverStripe\ORM\FieldType\DBField;
+use SilverStripe\ORM\ValidationException;
 use SilverStripe\Security\Authenticator;
 use SilverStripe\Security\IdentityStore;
 use SilverStripe\Security\Member;
@@ -195,6 +198,8 @@ class ChangePasswordHandler extends RequestHandler
      * @param array $data The user submitted data
      * @param ChangePasswordForm $form
      * @return HTTPResponse
+     * @throws ValidationException
+     * @throws NotFoundExceptionInterface
      */
     public function doChangePassword(array $data, $form)
     {
@@ -272,13 +277,12 @@ class ChangePasswordHandler extends RequestHandler
         $member->AutoLoginExpired = DBDatetime::create()->now();
         $member->write();
 
-        if ($member->canLogIn()) {
+        if ($member->canLogin()) {
             /** @var IdentityStore $identityStore */
             $identityStore = Injector::inst()->get(IdentityStore::class);
             $identityStore->logIn($member, false, $this->getRequest());
         }
 
-        // TODO Add confirmation message to login redirect
         $session->clear('AutoLoginHash');
 
         // Redirect to backurl
@@ -290,6 +294,10 @@ class ChangePasswordHandler extends RequestHandler
             return $this->redirect($backURL);
         }
 
+        $backURL = Security::config()->get('default_reset_password_dest');
+        if ($backURL) {
+            return $this->redirect($backURL);
+        }
         // Redirect to default location - the login form saying "You are logged in as..."
         $url = Security::singleton()->Link('login');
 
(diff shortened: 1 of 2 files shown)


dhensby:
seems like a fair addition... @tractorcow ?

Firesphere:
It actually fixes a ToDo I ran in to last week, that annoyed the hell outta me :D Hence, the PR :)

src/Security/MemberAuthenticator/ChangePasswordHandler.php
```diff
@@ -272,13 +275,12 @@ public function doChangePassword(array $data, $form)
         $member->AutoLoginExpired = DBDatetime::create()->now();
         $member->write();
 
-        if ($member->canLogIn()) {
+        if ($member->canLogin()) {
```
Firesphere: This is just an irrelevant PHPStorm complaining fix, irrelevant and not hurting the actual API. It just makes it case-sensitive.

src/Security/Security.php
```diff
@@ -1266,6 +1266,14 @@ public static function force_database_is_ready($isReady)
      */
     private static $default_login_dest = "";
 
+    /**
+     * @config
+     * @var string Set the default reset password destination
+     * This is the URL that users will be redirected to after they change their password,
+     * By default, it's redirecting to {@link $login}.
+     */
+    private static $default_reset_password_dest = false;
```
Firesphere: Set to `false`, so the actual handler will fall back to the login link instead of breaking when it's an empty string (happens on strict type-checks)

src/Security/MemberAuthenticator/ChangePasswordHandler.php
```diff
@@ -287,6 +289,10 @@ public function doChangePassword(array $data, $form)
             return $this->redirect($backURL);
         }
 
+        $backURL = Security::config()->get('default_reset_password_dest');
+        if ($backURL) {
```
Firesphere: Not using strict type checking, but depending on the PHP setting, it might. Hence the explicit false down in Security.

src/Security/MemberAuthenticator/ChangePasswordHandler.php
```diff
@@ -195,6 +196,8 @@ public function changePasswordForm()
      * @param array $data The user submitted data
      * @param ChangePasswordForm $form
      * @return HTTPResponse
+     * @throws \SilverStripe\ORM\ValidationException
+     * @throws \Psr\Container\NotFoundExceptionInterface
```
Firesphere: Just some docblocks that needed to be added to shut up PHPStorm :)

tractorcow: Please use `use` to simplify these FQN please.

src/Security/MemberAuthenticator/ChangePasswordHandler.php
tractorcow: `null` is preferred for unset values. We've had this issue with get_by_id() returning false for non-matches as well.

src/Security/Security.php
tractorcow: Have you verified this?

src/Security/MemberAuthenticator/ChangePasswordHandler.php
Firesphere: Sure. Could you please give me a heads up to rework this? I've got too much going to keep track of everything :D

src/Security/Security.php
Firesphere: It happened on my machine earlier, can't reproduce this anymore, so `null` is the way to go then :)

src/Security/MemberAuthenticator/ChangePasswordHandler.php
Firesphere: 👍

Firesphere:
Done @tractorcow

tractorcow:
Looks like it needs a rebase on the correct branch @Firesphere

Firesphere:
I just rebased on branch 4.0, I think it was the base it wanted to merge in to was wrong? This should be the right one, right?

tractorcow:
@Firesphere can you please make sure you tag your commits with the correct change type? E.g. `ENHANCEMENT` / `API` / `BUG`

dhensby:
shouldn't this target the 4 branch as it's a new API/feature?

@Firesphere renaming the issue isn't the same as amending your commit message ;)

tractorcow:
Oh yeah, maybe 4 is better.

Firesphere:
Yep, sure enough :)

Firesphere:
Gah, rebasing once again does weird things :| I'll fix that tonight

Firesphere:
@GitHub, Remind me next weekend (not the upcoming one) to fix this.

tractorcow:
@Firesphere does that reminder even work? I'd love it if it did lol.

tractorcow:
@Firesphere do you want me to rebase this for you and get it over the line, or will you do it?

Firesphere:
@tractorcow I hoped it did, but it didn't. I'll have to rebase and fix this though :| still :|

Firesphere:
You're free to rebase it and get it over the line :) I just have too much to do in too little time

Firesphere:
@tractorcow There you go :)

Issue #2995, opened 2014-03-31
Requirements::clear() does not clear Requirements_Backend->combine_files


Zauberfisch:
the following code:

```
    Requirements::combine_files('test1.js', array(
        THIRDPARTY_DIR . '/jquery/jquery.min.js',
        THIRDPARTY_DIR . '/jquery-entwine/dist/jquery.entwine-dist.js',
    ));
    Requirements::clear();
    Requirements::combine_files('test2.js', array(
        THIRDPARTY_DIR . '/jquery/jquery.min.js',
        THIRDPARTY_DIR . '/jquery-entwine/dist/jquery.entwine-dist.js',
    ));
```

currently results in the error:

```
Requirements_Backend::combine_files(): Already included files framework/thirdparty/jquery/jquery.min.js,framework/thirdparty/jquery-entwine/dist/jquery.entwine-dist.js in combined file 'test1.js'
```

But should in fact work. 
The fix should be rather easy, I think we only need to change `Requirements::clear()` to also clear `Requirements_Backend->combine_files`.

(This test case might look silly, but there is a valid use-case for wanting to include a file into multiple combined files.
lets assume we have a site that has a special page that is (in frontend terms) completely unrelated to the rest of the site. in this case I might wish to remove existing requirements and include a different combined file because the common ground is to small to split it up into two files.)

tractorcow:
Wow, this a bug still in 4.x too. =( Should be easy fix?

dhensby:
Bump

flamerohr:
Pull requests made to address this:
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/7318
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/7319

tractorcow:
Fixed. :D

PR #7318, opened 2017-08-28
Fix add combinedFiles to clear logic - ss4

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/2995 for SilverStripe 4

### DIFF
--- a/src/View/Requirements_Backend.php
+++ b/src/View/Requirements_Backend.php
@@ -704,25 +704,24 @@ class Requirements_Backend
      */
     public function clear($fileOrID = null)
     {
-        if ($fileOrID) {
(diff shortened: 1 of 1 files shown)


src/View/Requirements_Backend.php
```diff
@@ -704,25 +704,24 @@ public function getBlocked()
...
+            if ($fileOrID) {
                 if (isset($this->{$type}[$fileOrID])) {
                     $this->disabled[$type][$fileOrID] = $this->{$type}[$fileOrID];
                     unset($this->{$type}[$fileOrID]);
                 }
+            } else {
+                $this->disabled[$type] = $this->{$type};
+                $this->{$type} = [];
             }
-        } else {
```
tractorcow: Please update restore() as well.

PR #7321, opened 2017-08-29
Bump deprecation to 4.0 in preparation for rc and stable tags

tractorcow:
The effect of this change is that all `Deprecation::notice('4.0')` calls will now display as actual deprecation notices. This is best done before RC in order to capture any un-upgraded APIs marked for removal at 4.0.


flamerohr:
We're heading to release candidates soon, so... let the fun begin! (for modules)

PR #7336, opened 2017-09-04
Fix add schema to the "auto" parts request

flamerohr:
A part of fixing https://github.com/silverstripe/silverstripe-asset-admin/issues/163

### DIFF
--- a/src/Forms/Schema/FormSchema.php
+++ b/src/Forms/Schema/FormSchema.php
@@ -66,7 +66,7 @@ class FormSchema
         $return = ['id' => $schemaID];
 
         // Default to schema if not set
-        if ($form && ($wantSchema || empty($schemaParts))) {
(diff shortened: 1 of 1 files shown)


tractorcow:
On submission, are you sure you want to re-request the schema? Success state should be included in the state not the schema.

flamerohr:
Yes, this is because the form schema could change after submission... e.g. if a file is published, we want to start showing the unpublish button, and vice versa.

Unfortunately, I wasn't able to find another way to provide such an information change

tractorcow:
We may as well just get rid of auto then. Do you agree?

flamerohr:
That might be an idea... however getting rid of it seems like a bigger architectural decision which I'll need more time to think about :)

Currently, I think including the schema in "auto" is fine.

PR #7337, opened 2017-09-04
BUG Fix typing of DB types for supported drivers

### LINKED ISSUE #7039: [ORM] Inconsistent DB ORM casting of int values
This covers a range of issues, but is essentially caused by the lack of consistency between the various DB connector classes in the way that they encode integer values. Some encode them as strings, others as real integer.

This issue causes increasing breakage in client scripts where we work with de-serialised client side models, but in a type-strict way. E.g. `this.props.id === checkID`, which may not have consistent behaviour when using another DB backend.

I won't propose a desired resolution, but previously discussed suggestions included:
 - Enforce strictly-typed DB connectors (e.g. PDO, move to DBAL (https://github.com/silverstripe/silverstripe-framework/issues/6632))
 - Cast int fields at the DataObject level (cast on write, store normalised values)
 - Cast values during serialisation for the front-end
 
 This will also fix:

 * #3473
 * #6709

# Related Pull Requests
* [x] https://github.com/silverstripe/silverstripe-postgresql/pull/91
* [ ] https://github.com/silverstripe/silverstripe-framework/pull/8448

tractorcow:
Fixes #7039

Unfortunately, only the MysqlND driver supports native type returns (both PDO / mysqli version), but this change will support those options if available.

I've found that with mysqlnd, it's necessary to set ATTR_EMULATE_PREPARES to false explicitly; It's not enough to simply not set it. Given that there are situations where we need to avoid setting this, I've allowed this to be set to `null` to bypass assignment from the default `false`.

### DIFF
--- a/src/ORM/Connect/MySQLiConnector.php
+++ b/src/ORM/Connect/MySQLiConnector.php
@@ -73,10 +73,15 @@ class MySQLiConnector extends DBConnector
         $selectedDB = ($selectDB && !empty($parameters['database'])) ? $parameters['database'] : null;
 
         // Connection charset and collation
-        $connCharset = Config::inst()->get('SilverStripe\ORM\Connect\MySQLDatabase', 'connection_charset');
-        $connCollation = Config::inst()->get('SilverStripe\ORM\Connect\MySQLDatabase', 'connection_collation');
+        $connCharset = $this->config()->get('connection_charset');
+        $connCollation = $this->config()->get('connection_collation');
 
         $this->dbConn = mysqli_init();
+
+        // Use native types (MysqlND only)
+        if (defined('MYSQLI_OPT_INT_AND_FLOAT_NATIVE')) {
+            $this->dbConn->options(MYSQLI_OPT_INT_AND_FLOAT_NATIVE, true);
+        }
 
         // Set SSL parameters if they exist. All parameters are required.
         if (array_key_exists('ssl_key', $parameters) &&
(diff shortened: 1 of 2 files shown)


tractorcow:
@dhensby do you want to look at this for me? :D

tractorcow:
Nope, not as easy as I thought it'd be. =(

tractorcow:
As an aside, I've noticed a lot of errors running CMS tests w/ PDO, although the CI doesn't cover this.

dhensby:
see https://github.com/silverstripe/silverstripe-framework/pull/4919

I came to the conclusion this shouldn't be done by the DB connector, it should be the ORM - that's the approach that DBAL takes - the DBField takes responsibility for casting the type

tractorcow:
It looks like I ended up taking exactly the same path you tried, too bad!

It seems to work with mysqli at least. Maybe I'll restrict my fix to that?

dhensby:
> It seems to work with mysqli at least. Maybe I'll restrict my fix to that?

If we don't get the ORM to do the typing, then I wouldn't bother doing it at all...

tractorcow:
> I wouldn't bother doing it at all...

This is what I'm thinking... PHP is dynamically typed, and even if we could protect against loading mis-typed values from the DB, it would be a lot of overhead to enforce strict typing, and would go against the general expectations of coding in PHP (that types are inherently loose).

PR #7342, opened 2017-09-05
Fix permission check for admin role

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1856


tractorcow:
CMS behat tests are failing unrelated to this change.

PR #7343, opened 2017-09-06
NEW Allow GridFieldEditButton to have configurable HTML classes. Change edit icon.

sachajudd:


tractorcow:
@sachajudd could you please add a quick before / after screenshot for review? That'd be fab thanks. ;)

sachajudd:
@tractorcow Before: screen shot 2009-13-03 at 1 53 57 pm
After: 
screen shot 2009-13-03 at 1 54 21 pm

src/Forms/GridField/GridFieldEditButton.php
```diff
@@ -95,14 +106,63 @@ public function getColumnContent($gridField, $record, $columnName)
...
     }
 
     /**
+     * Get the extra HTML classes to add for edit buttons
+     *
+     * @return string
+     */
+    public function getExtraClass()
+    {
+        return implode(' ', Convert::raw2att($this->extraClass));
```
tractorcow: The `default_cast` should have already cast these as safely escaped text. The convert is not needed here.

src/Forms/GridField/GridFieldEditButton.php
```diff
@@ -95,14 +106,63 @@ public function getColumnContent($gridField, $record, $columnName)
...
+        return $this;
+    }
+
+    /**
+     * Checks whether the button has the given HTML class
+     *
+     * @param string $class
+     * @return boolean
+     */
+    public function hasExtraClass($class = '')
```
tractorcow: Is this method necessary?

src/Forms/GridField/GridFieldEditButton.php
```diff
@@ -95,14 +106,63 @@ public function getColumnContent($gridField, $record, $columnName)
...
+        return implode(' ', Convert::raw2att($this->extraClass));
+    }
+
+    /**
+     * Add an extra HTML class
+     *
+     * @param string $class
+     * @return $this
+     */
+    public function addExtraClass($class = '')
```
tractorcow: Default arg for addExtraClass should never be used; It's always mandatory right?

src/Forms/GridField/GridFieldEditButton.php
```diff
@@ -95,14 +106,63 @@ public function getColumnContent($gridField, $record, $columnName)
...
+    /**
+     * Remove an HTML class
+     *
+     * @param string $class
+     * @return $this
+     */
+    public function removeExtraClass($class)
+    {
+        $key = array_search($class, $this->extraClass);
+        unset($this->extraClass[$key]);
```
tractorcow: Please check for `$key !== false` before doing unset.

src/Forms/GridField/GridFieldEditButton.php
tractorcow: For your tests, I'm happy if you just check getExtraClass has the string (or lack thereof)

src/Forms/GridField/GridFieldEditButton.php
dhensby: `unset` doesn't throw errors for undefined vars - so this is safe and going to be slightly faster than adding a condition.

tractorcow:
Rebased, can merge on green.

tractorcow:
Whoops my push failed.

sachajudd:
Thanks, @tractorcow and thanks @robbieaverill for your help 🎉

PR #7345, opened 2017-09-06
TreeDropdownField: replace onadd by onmatch

schellmax:
Makes TreeDropdownField work in situations where it's moved around in the DOM (e.g. when inside a drag'n'drop container)

### DIFF
--- a/javascript/TreeDropdownField.js
+++ b/javascript/TreeDropdownField.js
@@ -47,7 +47,7 @@
 			// XMLHttpRequest
 			CurrentXhr: null,
 
-			onadd: function() {
+			onmatch: function() {
 				this.append(
 					'<span class="treedropdownfield-title"></span>' +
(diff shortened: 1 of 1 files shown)


th3hamm0r:
@dhensby Do you plan to merge this in the near future? The travis errors look like they are unrelated.

dhensby:
@th3hamm0r I haven't replicated nor tested this issue - in fact there's no linked issue or steps to replicate this - so I'm not confident about merging it right now.

@schellmax Do you have a more detailed issue report / steps to replicate?

th3hamm0r:
@dhensby Oh, that's right, there is no issue.
We've an inline editor where you can drag-and-drop content blocks, that use standard `FormFields`, like the `TreeDropdownField`.
All FormFields of SilverStripe have worked out of the box, except the `TreeDropdownField`, which whenever you move it, duplicates its dropdown field in the frontend (because the "onadd" trigger is called on every drag-and-drop). This commit fixes this issue.

PR #7346, opened 2017-09-06
[WIP] search for path or host for proper BASE_URL detection

andrewandante:
Currently returns an empty string if you declare an SS_BASE_URL with a protocol of any form.

This explodes the URL out into it's parts and checks for the existence of path/host as appropriate.


dhensby:
This feels super risky and a change to the intended value of `BASE_URL`

andrewandante:
Replaced with #7349

PR #7347, opened 2017-09-06
Feature: Convert::raw2filename

wernerkrauss:
This might be a handy shortcut for sanitising a filename.

See https://github.com/silverstripe/silverstripe-errorpage/issues/8
See https://github.com/silverstripe/silverstripe-subsites/issues/299

### DIFF
--- a/src/Core/Convert.php
+++ b/src/Core/Convert.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\Core;
 
+use SilverStripe\Assets\FileNameFilter;
 use SilverStripe\ORM\DB;
 use SilverStripe\View\Parsers\URLSegmentFilter;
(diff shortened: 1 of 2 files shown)


dhensby:
So we are replacing `FileNameFilter::create()->filter($f)` with `Convert::raw2filename($f)`?

I'm not sure there's a big incentive for this

tractorcow:
Thanks for the suggestion, but I don't think the additional reference is necessary for this functionality.

wernerkrauss:
@dhensby it's a similar shortcut like `Convert::raw2url()`, which ishandy and easier to find for new developers.

tractorcow:
Sorry it looks like there isn't enough core member support for this feature.

PR #7349, opened 2017-09-06
FIX use correct namespaces for middleware injection

### LINKED ISSUE #7348: SS_TRUSTED_PROXY_IPS is not applied to middleware
When setting the envvar SS_TRUSTED_PROXY_IPS the value is not passed through to the middleware. This means (for me) that the trusted header X_FORWARDED_PROTOCOL is not used to determine the request protocol, and therefore there is a mismatch between the URL (https://my.site.com) and the base_tag (http://my.site.com) which results in an unstyled CMS and mixed-protocol warnings.

This appears to be because it is not parsed correctly out of the config: https://github.com/silverstripe/silverstripe-framework/blob/4/_config/requestprocessors.yml#L19

andrewandante:
Fixes #7348


dhensby:
Should `Injector` be throwing errors if there are service definitions for non-existent classes?

PR #7354, opened 2017-09-08
Add 'required' attribute to front end field schema

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-admin/issues/160

### DIFF
--- a/src/Forms/FormField.php
+++ b/src/Forms/FormField.php
@@ -1583,6 +1583,7 @@ class FormField extends RequestHandler
             'value' => $this->Value(),
             'message' => $this->getSchemaMessage(),
             'data' => [],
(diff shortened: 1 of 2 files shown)


src/Forms/FormField.php
```diff
@@ -1583,6 +1583,7 @@ public function getSchemaStateDefaults()
             'value' => $this->Value(),
             'message' => $this->getSchemaMessage(),
             'data' => [],
+            'required' => $this->Required(),
```
flamerohr: I don't agree with this... could we discuss why this is here?
you know there's already a validation property for formfields, right?

phalkunz:
Discussed and decided to drop this pr.

PR #7355, opened 2017-09-08
Make auto login token expiry configurable (closes #7278)

### LINKED ISSUE #7278: Password reset “token” expiry time should be configurable
The current expiry time is 2 days, which isn’t a bad default but it should be configurable regardless.

I’m happy to patch this. Given that any change to [this method](https://github.com/silverstripe/silverstripe-framework/blob/3.6.1/security/Member.php#L716-L741) (in 3.x or 4.x, they’re identical) would be an API breakage I’m thinking it might be easier to add a new method to handle this, which takes the expiry time from a `Member.password_reset_expiry` config setting.

kinglozzer:


dhensby:
good idea - though 2 days seems like a long default (I know you haven't changed this), should we consider lowering this? I'd expect someone who's asked for a reset to use it pretty quickly (minutes to hours) not after a day or so.

PR #7356, opened 2017-09-10
Toggle Lost Password Link & Functionality from Security/login

zanderwar:
This will allow a developer to easily hide the "I've lost my password" link from the Security/login page, and return 404's for all the url_handlers used by that functionality

```yml
SilverStripe\Security\Security:
  # default is true
  enable_lost_password: false
```

### DIFF
--- a/src/Security/MemberAuthenticator/LostPasswordHandler.php
+++ b/src/Security/MemberAuthenticator/LostPasswordHandler.php
@@ -78,6 +78,9 @@ class LostPasswordHandler extends RequestHandler
      */
     public function lostpassword()
     {
+        if (!Security::config()->get('enable_lost_password')) {
+            return $this->httpError(404);
+        }
 
         $message = _t(
             'SilverStripe\\Security\\Security.NOTERESETPASSWORD',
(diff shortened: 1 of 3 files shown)


dhensby:
@Firesphere you happy with this?

Firesphere:
I've not even had a look at it properly, but there are a few things I noticed immediately.
1. `enable_lost_password` sounds like the opposite of what we'd like to achieve. It should be enabled by default, but optionally disabled
2. Where are the tests
3. The `if` statement in get_actions is nice and all, but does not actively push the "I've lost my password" field

What I would suggest, is instead of checking for enabled, would be to see if it was disabled and act on that.
That would limit the resources used as well (if(not_enabled){ do stuff} ) versus (if(disabled){ do stuff})). 

I do agree with the gist of the PR though. And I know I'm being quite picky here :)

dhensby:
behat failure looks related

robbieaverill:
@zanderwar do you want to have a look at this PR again?

tractorcow:
> enable_lost_password sounds like the opposite of what we'd like to achieve. It should be enabled by default, but optionally disabled

With 4.x onwards it's fine for all options to be truthful by default, since they can be turned off via yml (where we couldn't in 3.x) so I think this is fine.

dhensby:
The down stream repo has been deleted do this PR can't take modifications. We either merge by hand and add tests, etc at the same time or we need a new PR with someone else trashing ownership.

I don't think the former will happen so I'm closing this and someone can take it on if they want

PR #7366, opened 2017-09-12
Move to travis trusty dist (3.x)

chillu:
See https://github.com/silverstripe/silverstripe-framework/pull/7092 for context.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: php
 
-dist: precise
-
-sudo: false
+dist: trusty
 
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ addons:
   apt:
     packages:
       - tidy
+  firefox: "31.0"
 
 matrix:
   include:


dhensby:
We can't use trusty for the 3.x branch because 5.3 is not supported

chillu:
Oh yep - the 5.3 build fails: https://travis-ci.org/silverstripe/silverstripe-framework/builds/274452277

https://docs.travis-ci.com/user/reference/trusty/#PHP-images

> Note: We do not support PHP versions 5.2.x and 5.3.x on Trusty. Specifying it will result in build failure. If you need to test with these versions, use Precise.

So given that Travis is (tentatively) [ending precise support next April](https://blog.travis-ci.com/2017-08-31-trusty-as-default-status), but we want to support 3.x until late 2019, we have a bit of a problem. PHP 5.3 is already [end of life](http://php.net/supported-versions.php) though, so I'd say we just remove 5.3 builds now and make its support a "best effort" basis. If you're still running 5.3 now in production, you've got other problems than failing builds on framework. @silverstripe/core-team thoughts?

kinglozzer:
> just remove 5.3 builds now and make its support a "best effort" basis. If you're still running 5.3 now in production, you've got other problems

Agreed 👍.

sminnee:
Seems okay but not a priority until February ;-)

dhensby:
Yeo - let's keep 5.3 builds for as long as we can and take them off when we have to.

I'm also not against dropping 5.3 support in a minor release, so we can always discuss that when the time comes ;)

sminnee:
This would be a great time to have live usage stats being beamed to us, e.g. via a module that phoned home to check for module updates.

PR #7370, opened 2017-09-12
FIX SSViewer::add_themes() to properly prepend

### LINKED ISSUE #7369: SSViewer::add_themes() appends theme to list rather than prepends
Replication code:

```php
public function testThemesHelpers()
    {
        SSViewer::config()->remove('themes');
        SSViewer::set_themes(['mytheme', '$default']);
        $this->assertEquals(['mytheme', '$default'], SSViewer::get_themes());

        SSViewer::add_themes(['my_more_important_theme']);
        $this->assertEquals(['my_more_important_theme', 'mytheme', '$default'], SSViewer::get_themes());
    }
```
Second assertion returns:
```php
[
    0 => 'mytheme'
    1 => '$default'
    2 => 'my_more_important_theme'
]
```
Detailed in the [RFC](https://github.com/silverstripe/silverstripe-framework/issues/5604) is that:

> `SSViewer::add_themes()` to prepend additional themes onto the current stack. If add_themes() is passed a theme already in the cascade, then the given theme is pulled to a higher precedence place in the list.

Currently the new theme is appended instead. This means that adding a theme doesn't automatically apply it, which is the expected behaviour.

andrewandante:
Fixes #7369

### DIFF
--- a/src/View/SSViewer.php
+++ b/src/View/SSViewer.php
@@ -165,7 +165,10 @@ class SSViewer implements Flushable
 
     public static function add_themes($themes = [])
     {
-        SSViewer::config()->merge('themes', $themes);
+        $currentThemes = SSViewer::get_themes();
+        $finalThemes = array_merge($themes, $currentThemes);
+        // array_values is used to ensure sequential array keys as array_unique can leave gaps
+        SSViewer::set_themes(array_values(array_unique($finalThemes)));
     }
 
     public static function get_themes()
--- a/tests/php/View/SSViewerTest.php
+++ b/tests/php/View/SSViewerTest.php
@@ -81,6 +81,25 @@ class SSViewerTest extends SapphireTest
     }
 
     /**
+     * Tests for themes helper functions, ensuring they behave as defined in the RFC at
+     * https://github.com/silverstripe/silverstripe-framework/issues/5604
+     */
+    public function testThemesHelpers()
+    {
+        // Test set_themes()
+        SSViewer::set_themes(['mytheme', '$default']);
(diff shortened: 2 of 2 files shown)


src/View/SSViewer.php
```diff
@@ -165,7 +165,10 @@ public static function set_themes($themes = [])
 
     public static function add_themes($themes = [])
     {
-        SSViewer::config()->merge('themes', $themes);
+        $currentThemes = SSViewer::get_themes();
+        $finalThemes = array_merge($themes, $currentThemes);
+        SSViewer::config()->remove('themes');
```
flamerohr: `->remove()` is redundant, since `->set_themes()` uses `->set()` inside anyway.

Totally valid if it was using `->merge()` inside.

andrewandante: Thanks, have updated both this and the test 👍

andrewandante:
Removed a stray newline 😁

src/View/SSViewer.php
```diff
@@ -165,7 +165,9 @@ public static function set_themes($themes = [])
 
     public static function add_themes($themes = [])
     {
-        SSViewer::config()->merge('themes', $themes);
+        $currentThemes = SSViewer::get_themes();
+        $finalThemes = array_merge($themes, $currentThemes);
+        SSViewer::set_themes(array_values(array_unique($finalThemes)));
```
dhensby: I'd probably add a comment about why we're using `array_values` in case it's not clear to those who come along later

andrewandante: Comment added

dhensby:
OK - I'm happy, but I'll leave merging so others have a chance to review

PR #7373, opened 2017-09-13
NEW RateLimiter for Security controller

dhensby:
This adds request rate limiting to the Security controller. Currently seeking feedback.

todo:
- [x] Tests
- [x] Make RateLimiter parameters injectable

This has been heavily influenced by [Laravel's ThrottleRequests middleware](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Routing/Middleware/ThrottleRequests.php) see also [Laravel's RateLimiter](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Cache/RateLimiter.php).


src/Control/Middleware/RateLimitMiddleware.php
```diff
@@ -0,0 +1,71 @@
...
+        return md5($domain . '-' . $request->getIP());
+    }
+
+    /**
+     * @return HTTPResponse
+     */
+    protected function getErrorHTTPResponse()
+    {
+        $response = null;
+        if (class_exists(ErrorPage::class)) {
```
dhensby: I'm not sure if it's worth adding this. Perhaps the error page module should have it's own middleware that reformats error responses rather than coupling like this...

It also doesn't work if there's no error page in the CMS for the 429 response anyway

kinglozzer:
It could be that I’m misunderstanding “how things work”, but after looking at this it seems to me that it’d be really nice to be able to apply middlewares directly to `RequestHandler` classes. It would allow things like:

```yml
# Core config
  SilverStripe\Security\Security:
    properties:
      Middlewares:
        - '%$SecurityRateLimiter'
  SecurityRateLimiter:
    class: SilverStripe\Control\Middleware\RateLimitMiddleware
    constructor:
      args:
        max_attempts: 10
        decay: 1

# Mysite config
  Mysite\Api\Controller:
    properties:
      Middlewares:
        - '%$RelaxedRateLimiter'
  RelaxedApiRateLimiter:
    class: SilverStripe\Control\Middleware\RateLimitMiddleware
    constructor:
      args:
        max_attempts: 10000
        decay: 1
```

Docs suggest this is already possible to a degree, via [route-specific middleware](https://docs.silverstripe.org/en/4/developer_guides/controllers/middlewares/#route-specific-middleware). Thoughts?

tractorcow:
I think that we should be able to approach these issues with middleware-only; The RequestHandlerMiddlewareAdapter doesn't seem to be included in this pull request, so I can't review it's purpose, but it probably can be done without.

@kinglozzer regarding route-specific middleware, I think the approach we decided on was to apply middleware at the lower director level.

dhensby:
[RequestHandlerMiddlewareAdapter](https://github.com/silverstripe/silverstripe-framework/blob/4/src/Control/Middleware/RequestHandlerMiddlewareAdapter.php) is an existing class which is referenced from the docs about how to apply route specific middleware: https://docs.silverstripe.org/en/4/developer_guides/controllers/middlewares/#route-specific-middleware

tractorcow:
Oh, and look who wrote that code... and then couldn't even figure out where it came from (me).

src/Dev/SapphireTest.php
```diff
@@ -1090,7 +1090,7 @@ public static function getExtraDataObjects()
      */
     public static function getExtraControllers()
     {
-        return static::$extra_controllers;
+        return array_merge([Security::class], static::$extra_controllers);
```
tractorcow: I'm not sure this should be in here. Security has it's own routes that should be routed by default via YAML.

dhensby: This change makes requests to the security controller go directly to the controller rather than via the ratelimiter.

The routes are, by default, defined in yaml (to go via the rate limited middleware). This just replaces that definition.

dhensby:
> Also ready for those tests. :)

I want to get the current suite passing first - which means finding a way to turn off the rate limiting for behat tests...

src/Dev/SapphireTest.php
tractorcow: Can you move the rate limit enable / disable to a RateLimitDisableTestState? If you really need to disable it for all tests that would seem the most injectable, instead of baking this into the sapphiretest class.

tractorcow:
Why don't you put the rate limiting registration into a conditional YML based on `environment: live`? That seems like the most appropriate fix.

tractorcow:
```yaml
---
Name: security-limited
Except:
  environment: dev
---
SilverStripe\Control\Director:
  rules:
    'Security//$Action/$ID/$OtherID': '%$RateLimitedSecurityController'
---
Name: security-unlimited
Only:
  environment: dev
---
SilverStripe\Control\Director:
  rules:
    'Security//$Action/$ID/$OtherID': 'Security'
```

src/Dev/SapphireTest.php
tractorcow: See below.

dhensby:
> Why don't you put the rate limiting registration into a conditional YML based on environment: dev? That seems like the most appropriate fix.

I thought about putting this on live environments only or even, as you've suggested, not on dev environments but...

 - We really want rate limiting on always to help with debugging, testing and any instances where devs accidentally deploy sites in dev mode
 - Whilst maybe our test runs are done in dev mode, is there any logic to force this? This effectively will make test runs done under any other environment type fail, which feels undesirable.

tractorcow:
> We really want rate limiting on always to help with debugging, testing and any instances where devs accidentally deploy sites in dev mode

Then you can turn it on with custom YML. I don't want to return to the days of "is_running_tests()" scattered through our codebase.

> Whilst maybe our test runs are done in dev mode, is there any logic to force this? This 

Yes behat will fail if you don't set it to DEV.

_config/routes.yml
```diff
@@ -11,7 +11,7 @@ After:
 ---
 SilverStripe\Control\Director:
   rules:
-    'Security//$Action/$ID/$OtherID': SilverStripe\Security\Security
+    'Security//$Action/$ID/$OtherID': '%$RateLimitedSecurityController'
```
chillu: Should we allow adding middleware to route definitions directly? It's not really composable this way. Might also be worth pointing out in the upgrade guide, since I've seen devs overwriting their `Security` director rules to direct to custom controllers.

src/Control/Middleware/RateLimitMiddleware.php
```diff
@@ -0,0 +1,72 @@
...
+        }
+        $this->addHeadersToResponse($response, $limiter);
+        return $response;
+    }
+
+    /**
+     * @param HTTPRequest $request
+     * @return string
+     */
+    protected function getKeyFromRequest($request)
```
chillu: The rate limiting is applied to a particular route, but seems to be global in it's effect. AFAICT, it doesn't take the applied route or current URL path into account - should it? We couldn't include GET params there, since that would make the implementation vulnerable to randomisation. But if URL routing is implemented correctly, there should only be one path to get to a certain controller.

_config/routes.yml
tractorcow: We could nest this as another composable level: e.g.


```yaml
SilverStripe\Control\Director:
  rules:
    'Security//$Action/$ID/$OtherID': '%$SecurityController'
---
Name: securitystuff
---
  SecurityController: '%$RateLimitedSecurityController'
```

dhensby:
@tractorcow, ok so there's no way for a behat suite to apply a global config to a site for a test run?

I'm definitely not advocating to wrap this in a "system under test" condition.

If there's no alternative other than the conditional yaml config then I'll have to go with that, but it feels bad to me.

tractorcow:
You could add an option to ConfigContext, and then have this config activated in a `@beforeScenario` step, but honestly the dev conditional config is so much less invisible and easier for users to read / modify.

src/Control/Middleware/RateLimitMiddleware.php
chillu: This is blocking the merge from my perspective.

src/Control/Middleware/RateLimitMiddleware.php
```diff
@@ -0,0 +1,72 @@
...
+{
+
+    /**
+     * @param HTTPRequest $request
+     * @param callable $delegate
+     * @return HTTPResponse
+     */
+    public function process(HTTPRequest $request, callable $delegate)
+    {
+        $limiter = new RateLimiter($this->getKeyFromRequest($request), 10, 1);
```
chillu: I assume the max attempts and decay constructor args aren't YAML configurable because this happens before YAML config is available? Maybe note that here?

src/Core/Cache/RateLimiter.php
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * @var int Maximum number of attempts allowed
+     */
+    private $maxAttempts;
+
+    /**
+     * @var int How long the rate limit lasts for
+     */
+    private $decay;
```
chillu: Should state the unit (seconds)

_config/routes.yml
chillu: However you configure the class use (route or injector), it's still a different class that you might accidentally overwrite as a dev without realising what you're doing. Our YAML-based dependency injection is powerful, but not very transparent - there's no good way to list the "final" injection rules and determine if you've overwritten another rule, right?

I don't see this as a controller wrapper, but as a route specific middleware, which is a feature that doesn't exist in core at the moment. Sorry, but I think we should defer this implementation until it does.

chillu:
I'm fine with a dev conditional for this. If you're load testing your app, you should do it in "live" mode. If you accidentally deploy in dev mode, you have bigger problems than rate limiting. The only way I can see this tripping up developers is when they run end-to-end tests in "test" rather than "dev" mode. 

If this doesn't get in during in the next ~48h, it's dropping out of the beta4 milestone, our last beta release. And since we're not doing new features in the RC phase, that means it'll have to wait until 4.1. Regarding my comments around route specific middleware, that seems the more likely path at the moment, although I'm keen to hear from other core devs if they're fine with the current "controller wrapper" approach.

_config/routes.yml
dhensby: >I don't see this as a controller wrapper, but as a route specific middleware, which is a feature that doesn't exist in core at the moment. Sorry, but I think we should defer this implementation until it does.

This is literally in the docs as how to implement route-specific middleware. I don't see that it makes sense to block this PR because the current way to implement route-specific middleware isn't deemed to be suitable.

We can address this in 4.x branch once 4.0 is out if we are concerned about the implementation of route-specific middlewares, but to block a fairly major security enhancement (which will make SS more secure out of the box) on these grounds seems to go against what's best for the framework and our users.

src/Control/Middleware/RateLimitMiddleware.php
dhensby: I'm removing this

src/Control/Middleware/RateLimitMiddleware.php
dhensby: OK - I've added a change so that the rate limit middleware accepts some extra data to be added to the key. This allows the middleware to be used across controllers/routes/etc.

src/Control/Middleware/RateLimitMiddleware.php
dhensby: It's just because I wanted to do it later, tbh. I'll get it in now.

dhensby: added

src/Core/Cache/RateLimiter.php
dhensby: done

dhensby:
Right, I've added the tests, which hopefully cover all that we need. Now onto docs...

dhensby:
Docs are in

_config/routes.yml
```diff
@@ -18,3 +18,11 @@ SilverStripe\Control\Director:
     'InstallerTest//$Action/$ID/$OtherID': SilverStripe\Dev\InstallerTest
     'SapphireInfo//$Action/$ID/$OtherID': SilverStripe\Dev\SapphireInfo
     'SapphireREPL//$Action/$ID/$OtherID': SilverStripe\Dev\SapphireREPL
+---
+Name: security-limited
+Except:
+  environment: dev
```
tractorcow: We should probably add an explicit `after: rootroutes` here.

docs/en/02_Developer_Guides/09_Security/05_Rate_Limiting.md
```diff
@@ -0,0 +1,72 @@
...
+
+You may already solve the rate limiting problem on a server level and the built in rate limiting may well be redundant.
+If this is the case you can turn off the rate limiting middleware by redefining the URL rules for the Security controller.
+
+Add the following to your config.yml:
+
+```yml
+SilverStripe\Control\Director:
+  rules:
+    'Security//$Action/$ID/$OtherID': SilverStripe\Security\Security
```
chillu: Does the value need to be quoted due to the backslashes?

chillu:
Apart from one small docs comment, this looks good to me now - great work!

docs/en/02_Developer_Guides/09_Security/05_Rate_Limiting.md
tractorcow: No it's safe :)

dhensby:
Yay :) thanks team

PR #7374, opened 2017-09-13
FIX HTTP Headers are case insensitive

dhensby:
HTTP Headers are case insensitive but the framework treats them as case sensitive which means setting headers can be risky as "content-type" is treated as a distinct header from "Content-Type", which is wrong.

### DIFF
--- a/src/Control/HTTPRequest.php
+++ b/src/Control/HTTPRequest.php
@@ -349,12 +349,14 @@ class HTTPRequest implements ArrayAccess
     /**
      * Add a HTTP header to the response, replacing any header of the same name.
      *
-     * @param string $header Example: "Content-Type"
+     * @param string $header Example: "content-type"
      * @param string $value Example: "text/xml"
      */
     public function addHeader($header, $value)
     {
+        $header = strtolower($header);
         $this->headers[$header] = $value;
+        return $this;
     }
 
     /**
@@ -373,6 +375,7 @@ class HTTPRequest implements ArrayAccess
      */
     public function getHeader($header)
     {
+        $header = strtolower($header);
         return (isset($this->headers[$header])) ? $this->headers[$header] : null;
     }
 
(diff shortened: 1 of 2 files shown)


tractorcow:
The problem with this is that now it forces all http headers to be lowercase on output, which is not simply the same as making header methods case insensitive.

That might not be fatal to this fix though. We should probably do it anyway.

dhensby:
> The problem with this is that now it forces all http headers to be lowercase on output, which is not simply the same as making header methods case insensitive.

True, but that's not a problem because they are case insensitive :)

tractorcow:
It's only a problem to my pedanticalness.

madmatt:
I've seen instances where older versions of IE assumed case-sensitive headers (specifically `Content-Type` rather than `Content-type`. Was this checked to make sure it wouldn't break anything? Given the significant amount of browsers, web servers, proxies and clients around, simply outputting everything as lowercase seems a bit heavy-handed and may cause regressions.

Just a thought anyway, I can't find concrete examples of issues that are easy to test unfortunately.

I agree that we should compare headers case-insensitively as per RFC2616/RFC7230 spec, but ideally use original case when outputting, allowing the dev to override if necessary for their specific issue.

Examples: [1](https://github.com/nodejs/node-v0.x-archive/issues/1954#issuecomment-162250909)

tractorcow:
@madmatt I think we could do this, but maybe let's look at this in 4.1 if we find it necessary. If needed we could have a case-mapping variable for all headers as a separate array, so that lookup remains fast. :)

dhensby:
in http2 all the headers are dealt with in lowercase regardless of output, so this is going to become more and more a legacy problem (if it even is one now).

We could look at adjusting the output of headers so that they `ucwords` the headers? ie: `ucwords($headerName, '-')` to get content-type -> Content-Type and so on

Issue #7385, opened 2017-09-19
Performance recommendation: Ensure composer autoload takes priority over classloader


tractorcow:
Parent issue: https://github.com/silverstripe/silverstripe-admin/issues/199

Research shows that the core ClassLoader is still taking responsibility for about 1/4 of all class loading; With psr-4 this means we are sacrificing a huge amount of composer optimisation potential.

![image](https://user-images.githubusercontent.com/936064/30571595-03052c58-9d3d-11e7-8fd0-a8114fc8a50d.png)

https://blackfire.io/profiles/5ab08678-8c9a-4177-87cd-0ba10f3e4310/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=SilverStripe%5CCore%5CManifest%5CClassLoader%3A%3AloadClass&callname=main()

# PRs

- [ ] https://github.com/silverstripe/silverstripe-framework/pull/7386
- [x] https://github.com/silverstripe/silverstripe-config/pull/21
- [x] https://github.com/silverstripe/silverstripe-cms/pull/1973

tractorcow:
Update to this, it seems that PSR-4 paths are actually being loaded by composer first, but are failing. Investigating.

tractorcow:
The issue I've discovered is due to two major components forcing classes to lowercase prior to loading:
 - ClassManifest
 - silverstripe/config module

I'll update these to still respect case-insensitivity, but not to force all classes to lowercase on output.

sminnee:
Another option to consider for this would be to limit the manifest-based class loader to only interrogate modules that don't have psr-4 configuration.

PR #7386, opened 2017-09-19
ENHANCEMENT Don't force all class names to lowercase

tractorcow:
Speeds up autoloading because composer psr-4 works properly now. (~40ms on assets section load).

Requires https://github.com/silverstripe/silverstripe-config/pull/21

Fixes: https://github.com/silverstripe/silverstripe-framework/issues/7385

Parent story: https://github.com/silverstripe/silverstripe-admin/issues/199

CMS tests fixed with https://github.com/silverstripe/silverstripe-cms/pull/1973

Before:

image

image

After:

image

image

### DIFF
--- a/src/Core/ClassInfo.php
+++ b/src/Core/ClassInfo.php
@@ -3,13 +3,13 @@
 namespace SilverStripe\Core;
 
 use Exception;
+use ReflectionClass;
+use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Control\Director;
 use SilverStripe\Core\Manifest\ClassLoader;
 use SilverStripe\Dev\Deprecation;
-use SilverStripe\ORM\ArrayLib;
-use SilverStripe\ORM\DB;
 use SilverStripe\ORM\DataObject;
-use ReflectionClass;
+use SilverStripe\ORM\DB;
 
 /**
  * Provides introspection information about the class tree.
@@ -20,28 +20,6 @@ use ReflectionClass;
  */
 class ClassInfo
 {
-
-    /**
-     * Wrapper for classes getter.
-     *
-     * @return array
-     */
-    public static function allClasses()
-    {
-        return ClassLoader::inst()->getManifest()->getClasses();
-    }
-
-    /**
-     * Returns true if a class or interface name exists.
-     *
-     * @param  string $class
-     * @return bool
-     */
-    public static function exists($class)
-    {
-        return class_exists($class, false) || interface_exists($class, false) || ClassLoader::inst()->getItemPath($class);
-    }
-
     /**
      * Cache for {@link hasTable()}
      *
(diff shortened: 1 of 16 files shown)


src/Core/ClassInfo.php
```diff
@@ -364,8 +383,9 @@ public static function table_for_object_field($candidateClass, $fieldName)
      */
     public static function shortName($nameOrObject)
     {
-        $reflection = new ReflectionClass($nameOrObject);
-        return $reflection->getShortName();
+        $name = static::class_name($nameOrObject);
+        $parts = explode('\\', $name);
+        return end($parts);
```
flamerohr: looks worse than using `ReflectionClass::getShortName()` but is there a performance gain from this?

chillu:
Wow, awesome work! Does this need a mention in the upgrading guide? If you've misspelled class capitalisation in your own PHP before, it'll break now, right?

src/Core/Manifest/ClassManifest.php
```diff
@@ -489,23 +583,57 @@ public function handleFile($basename, $pathname, $includeTests)
...
-                    array($class),
-                    $this->coalesceDescendants($class)
-                );
+    /**
+     * Reload state from given cache data
+     *
+     * @param array $data
+     * @return bool True if cache was valid and successfully loaded
+     */
+    protected function loadFromCache($data)
```
flamerohr: what about naming this `loadToState`?
I was confused with the current naming, as "from" implies you're returning data from the cache, as opposed to pushing data to the state
This might be because I wasn't thinking about the cache being external...

src/Core/Manifest/ClassManifest.php
```diff
@@ -489,23 +583,57 @@ public function handleFile($basename, $pathname, $includeTests)
...
 
-            return $this->descendants[$lClass];
-        } else {
-            return array();
+    /**
+     * Load current state into an array of data
+     *
+     * @return array
+     */
+    protected function getCacheableData()
```
flamerohr: I'd almost name this `getCurrentState()`, but I think this is fine actually

src/Core/Manifest/ClassManifest.php
```diff
@@ -516,6 +644,12 @@ protected function coalesceDescendants($class)
      */
     protected function validateItemCache($data)
     {
+        if (!$data) {
+            return false;
+        }
+        if (is_array($data)) {
```
flamerohr: shouldn't this be `if (!is_array($data)) {`? at least that's what it looks like from the code below

tractorcow:
> Wow, awesome work! Does this need a mention in the upgrading guide? If you've misspelled class capitalisation in your own PHP before, it'll break now, right?

No, the only cost is that it'll probably go through classloader autoloader, instead of composer autoloader, but only for that one instance. :D

src/Core/ClassInfo.php
tractorcow: string manipulation is less expensive than creating a new object.

tractorcow:
Fixed some bugs and it's faster again.

image

tractorcow:
CMS Failures should be fixed with https://github.com/silverstripe/silverstripe-cms/pull/1973

PR #7389, opened 2017-09-20
BUG Fix sub-template lookup for includes

### LINKED ISSUE #7383: Performance recommendation: Lazy-evaluated template properties
Parent issues:
 - https://github.com/silverstripe/silverstripe-framework/issues/7378
 - https://github.com/silverstripe/silverstripe-admin/issues/199
  
When rendering templates, it can sometimes be necessary to provide variable values. E.g. via ->customise(), ArrayData(), etc.

Additionally SSViewer::process() will merge in various other nested process calls to provide the $Layout and $Content properties. (see https://github.com/silverstripe/silverstripe-framework/blob/0aabcac7479f961b9b2d57a7baa3a395c68907ac/src/View/SSViewer.php#L538-L540).

The issue is that even if we don't use these variables, they still need to be evaluated in advance. This causes excessive and unnecessary pre-rendering.

In order to address this I propose that ViewableData be given lazy-evaluated property support, so that callables are safely evaluated just before inserting into templates.

Changes needed:

- SSViewer::obj() supports callables
- SSViewer_DataPresenter::getInjectedValue() supports callables
- SSViewer::render() assigns sub-templates via callables
- ArrayData::getField() is updated to support callables
- Usages of ArrayData that make expensive calls should be updated to use callables
- Others? (TBC)

# PRs

 - [ ] https://github.com/silverstripe/silverstripe-framework/pull/7389

tractorcow:
Fixes #7383

### DIFF
--- a/src/View/SSViewer.php
+++ b/src/View/SSViewer.php
@@ -474,6 +474,8 @@ class SSViewer implements Flushable
         $scope = new SSViewer_DataPresenter($item, $overlay, $underlay, $inheritedScope);
         $val = '';
 
+        // Placeholder for values exposed to $cacheFile
+        [$cache, $scope, $val];
         include($cacheFile);
 
         return $val;
@@ -519,26 +521,26 @@ class SSViewer implements Flushable
         // Makes the rendered sub-templates available on the parent item,
         // through $Content and $Layout placeholders.
         foreach (array('Content', 'Layout') as $subtemplate) {
-            $sub = null;
-            if (isset($this->subTemplates[$subtemplate])) {
-                $sub = $this->subTemplates[$subtemplate];
-            } elseif (!is_array($this->templates)) {
-                $sub = ['type' => $subtemplate, $this->templates];
-            } elseif (!array_key_exists('type', $this->templates) || !$this->templates['type']) {
-                $sub = array_merge($this->templates, ['type' => $subtemplate]);
+            // Detect sub-template to use
+            $sub = $this->getSubtemplateFor($subtemplate);
+            if (!$sub) {
+                continue;
             }
 
-            if ($sub) {
+            // Create lazy-evaluated underlay for this subtemplate
+            $underlay[$subtemplate] = function () use ($item, $arguments, $sub) {
                 $subtemplateViewer = clone $this;
                 // Disable requirements - this will be handled by the parent template
                 $subtemplateViewer->includeRequirements(false);
                 // Select the right template
                 $subtemplateViewer->setTemplate($sub);
 
+                // Render if available
                 if ($subtemplateViewer->exists()) {
-                    $underlay[$subtemplate] = $subtemplateViewer->process($item, $arguments);
+                    return $subtemplateViewer->process($item, $arguments);
                 }
-            }
+                return null;
+            };
         }
 
         $output = $this->includeGeneratedTemplate($cacheFile, $item, $arguments, $underlay, $inheritedScope);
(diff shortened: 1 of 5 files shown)


src/View/SSViewer.php
```diff
@@ -474,6 +474,8 @@ protected function includeGeneratedTemplate($cacheFile, $item, $overlay, $underl
         $scope = new SSViewer_DataPresenter($item, $overlay, $underlay, $inheritedScope);
         $val = '';
 
+        // Placeholder for values exposed to $cacheFile
+        [$cache, $scope, $val];
```
kinglozzer: What difference does this change make? We could also consider the [composer autoloader approach](https://github.com/composer/composer/blob/e42e1156d51ca5494b058b1a7b480bc703c4f57a/src/Composer/Autoload/ClassLoader.php#L437-L445) if isolating this include is a concern for any reason.

src/View/SSViewer_DataPresenter.php
```diff
@@ -117,85 +130,29 @@ protected function createCallableArray(&$extraArray, $interfaceToQuery, $variabl
...
-                    $obj = Injector::inst()->get($casting, false, array($property));
-                    $obj->setValue($res['value']);
-
-                    $res['obj'] = $obj;
-                }
-            }
+        // Look up the value - either from a callable, or from a directly provided value
+        $res = [];
+        if (isset($source['callable'])) {
+            $res['value'] = call_user_func_array($source['callable'], $params);
```
kinglozzer: Appreciate you didn’t write this, but we can now use (the apparently slightly faster) argument unpacking:

```php
$callable = $source['callable'];
$res['value'] = $callable(...$params);
```

src/View/SSViewer_DataPresenter.php
```diff
@@ -316,4 +276,110 @@ public function __call($name, $arguments)
...
+     * Determine source to use for getInjectedValue
+     *
+     * @param string $property
+     * @return array|null
+     */
+    protected function getValueSource($property)
+    {
+        // Check for a presenter-specific override
+        $overlay = $this->processTemplateOverride($property, $this->overlay);
+        if (isset($overlay)) {
```
kinglozzer: Can’t these just be `if ($overlay)`?

src/View/SSViewer_DataPresenter.php
```diff
@@ -316,4 +276,110 @@ public function __call($name, $arguments)
...
+        // Check if the method to-be-called exists on the target object - if so, don't check any further
+        // injection locations
+        $on = $this->itemIterator ? $this->itemIterator->current() : $this->item;
+        if (isset($on->$property) || method_exists($on, $property)) {
+            return null;
+        }
+
+        // Check for a presenter-specific override
+        $underlay = $this->processTemplateOverride($property, $this->underlay);
+        if (isset($underlay)) {
```
kinglozzer: As above

kinglozzer:
Added a couple of minor comments, nothing that should block this getting over the line. Happy for this to be merged

src/View/SSViewer.php
tractorcow: This is to protect these variables from being accidentally deleted, as the $cacheFile will require these. You'll notice I've taken a similar approach in the installer.

https://github.com/silverstripe/silverstripe-framework/blob/c939737e5c8c484a02a61bcfbc7737f1da48224d/src/Dev/Install/install5.php#L111

src/View/SSViewer_DataPresenter.php
tractorcow: Won't work if using `[$this, 'methodName']` approach I think. I use call_user_func_array for legacy purposes.

src/View/SSViewer_DataPresenter.php
tractorcow: The template could be a valid empty string, so I use nulls as explicit non-value.

PR #7390, opened 2017-09-20
Added unit tests for updating Config values on extensions

wilr:
Demonstrates an issue with extension values not being able to be overridden.

This is blocking the 4.0 upgrade for comments as values on the comments extension are unable to be updated via the Config API.

### DIFF
--- a/tests/php/Core/Config/ConfigTest.php
+++ b/tests/php/Core/Config/ConfigTest.php
@@ -145,6 +145,42 @@ class ConfigTest extends SapphireTest
         $this->assertNull(Config::inst()->get(ConfigTest\First::class, 'nullable'));
         Config::modify()->merge(ConfigTest\First::class, 'nullable', 'value');
         $this->assertEquals('value', Config::inst()->get(ConfigTest\First::class, 'nullable'));
+
+        // arrays of false-y
+        $array = Config::inst()->get(ConfigTest\First::class, 'default_array');
+        $this->assertTrue($array['default_true']);
+        $this->assertFalse($array['default_false']);
+
+        Config::modify()->merge(ConfigTest\First::class, 'default_array', [
+            'default_true' => false
+        ]);
+
+        $array = Config::inst()->get(ConfigTest\First::class, 'default_array');
+        $this->assertFalse($array['default_true']);
+        $this->assertFalse($array['default_false']);
+
+        Config::modify()->merge(ConfigTest\First::class, 'default_array', [
+            'default_false' => true
+        ]);
+
+        $array = Config::inst()->get(ConfigTest\First::class, 'default_array');
+        $this->assertFalse($array['default_true']);
+        $this->assertTrue($array['default_false']);
+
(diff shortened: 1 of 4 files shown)


wilr:
@tractorcow I think `SilverStripe\Config\MiddlewareAware` is not running the last value correct. Doesn't appear to be any tests around how `MiddlewareAware::callMiddleware` functions but I confirm the config value is correct in the `$last` method but `$result` returns the incorrect one.

tractorcow:
You need to apply the config::update on the extension itself, I'm afraid.

I don't think the new architecture can support local modifications to the base object overwriting extensions anymore; It probably just needs to be documented, but the extensions are higher level and must be modified separately.

wilr:
We can't do Config update on the extension since the each object that has the extension may have unique config.

tractorcow:
More explanation:

Config is sourced from a collection on a per-class basis, which is merged together via middlewares; Firstly via inheritance, and then via extensions.

However, config modifications are applied to the root config for each class before these are applied; Meaning that even those can be overwritten by extensions.

The old config used to have a base config, with a set of modifications ("add", "delete" etc) which would then be applied on top of the merged config each time it was read. This was initially discarded because it was complex and was the source of some performance issues.

Possibly we could go back to a middle ground, but at the moment the architecture doesn't allow it.

Theoretically, you could apply modifications as a middleware itself; And shift all non-persistent class modifications to that new middleware instead of modifying the base. The new middleware would store all the modifications in order (add, delete, merge, etc) and apply them post-class merge.

However, this would probably slow the config API down yet again.

In core we have opted to avoid config modification on all critical paths completely; If you have seen https://github.com/silverstripe/silverstripe-framework/pull/7392 you'll note that we have removed certain config modification calls in preference to using getters / setters that draw their defaults from the config instead.

tractorcow:
> This is blocking the 4.0 upgrade for comments as values on the comments extension are unable to be updated via the Config API.

I wonder if there's a way you could use get_extra_config() to work around this?

tractorcow:
I've had a look at the config, and I do admit this is an annoying conundrum. 

You could possibly do this to ensure config is overwritten:

```
public function get_extra_config($class) 
{
	// Overwrite CommentsExtension.comments config with base class values
	return Config::inst()->get($class, null, true);
}
```

get_extra_config() is merged into the config after the extension's own config, so it should work.

tractorcow:
I've had a chat with @chillu and our initial feeling is that the system is not working as intended; I suspect we may need to look into the local-modifications middleware after all!

tractorcow:
I've raised this as a bug over at https://github.com/silverstripe/silverstripe-config/issues/22

tractorcow:
@wilr can you please have a look at https://github.com/silverstripe/silverstripe-config/issues/22 to confirm this fixes your issues.

wilr:
@tractorcow Sweet thanks! the tests in this PR pass with #7407 applied.

PR #7392, opened 2017-09-21
Enhancement Add setter and getter for certain classes

flamerohr:
so that LeftAndMain no longer updates config during init

Please check https://github.com/silverstripe/silverstripe-admin/issues/264 for more details

### DIFF
--- a/src/Control/ContentNegotiator.php
+++ b/src/Control/ContentNegotiator.php
@@ -39,35 +39,29 @@ class ContentNegotiator
 
     /**
      * @config
-     *
      * @var string
      */
     private static $content_type = '';
 
     /**
      * @config
-     *
      * @var string
      */
     private static $encoding = 'utf-8';
 
     /**
      * @config
-     *
      * @var bool
      */
     private static $enabled = false;
-    
+
     /**
-     * @config
-     *
      * @var bool
      */
     protected static $current_enabled = null;
 
     /**
      * @config
-     *
      * @var string
      */
     private static $default_format = 'html';
@@ -97,7 +91,7 @@ class ContentNegotiator
             return (substr($response->getBody(), 0, 5) == '<' . '?xml');
         }
     }
-    
+
     /**
      * Gets the current enabled status, if it is not set this will fallback to config
      *
@@ -110,7 +104,7 @@ class ContentNegotiator
         }
         return Config::inst()->get(static::class, 'enabled');
     }
-    
+
     /**
      * Sets the current enabled status
      *
--- a/src/Control/ContentNegotiator.php
+++ b/src/Control/ContentNegotiator.php
@@ -57,6 +57,13 @@ class ContentNegotiator
      * @var bool
      */
     private static $enabled = false;
+    
+    /**
+     * @config
+     *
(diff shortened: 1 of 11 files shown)


src/View/SSViewer.php
```diff
@@ -332,6 +332,8 @@ public static function hasTemplate($templates)
      * @var boolean
      */
```
tractorcow: I think we should have a similar property for `set_themes`; `$current_themes`?

src/Control/ContentNegotiator.php
```diff
@@ -84,12 +91,37 @@ public static function enabled_for($response)
...
+    
+    /**
+     * Gets the current enabled status, if it is not set this will fallback to config
+     *
+     * @return bool
+     */
+    public static function getEnabled()
+    {
+        if (!isset(self::$current_enabled)) {
+            self::$current_enabled = ContentNegotiator::config()->get('enabled');
```
tractorcow: I want to change my mind about something; I think we should not modify state inside the getters (even though i18n does this). The justification for i18n was due to performance and the fact that i18n::get_theme() is called hundreds of times.

For less-frequently accessed properties, shall we use
```
if (!isset(self::$current_enabled)) {
  return ContentNegotiator::config()->get('enabled');
}
```

Instead?

src/View/SSViewer.php
flamerohr: yeah, I remember the documentation recommended setting theme in the `_config.php` files, so making sure the recommendation uses `SSViewer::set_themes()` along with the `$current_themes` static property would be a good idea :)

src/Control/ContentNegotiator.php
flamerohr: hmm, I kind of agree with your proposal about the less frequently accessed properties, but I'm wary of setting another paradigm for them just because of the fact that they're not used frequently
reference:
```
if (!isset(self::$current_enabled)) {
  return ContentNegotiator::config()->get('enabled');
}
return self::$current_enabled;
```
would you be opposed to updating `i18n::get_locale()` to follow?
a config read should be cheap right? and the whole purpose of this pull request is to remove set/merge calls to the config in Core

flamerohr: We've agreed to update i18n as well so we can be consistent with this suggestion - in such a way that it doesn't use `Config_ForClass` which was the root cause of the slow down for i18n

flamerohr:
Looks good, do you think it'd be worth updating `i18n` to have this behaviour as well? @tractorcow

tractorcow:
@flamerohr i18n already has a static var to override the config; It's probably ok as it is for now.

tractorcow:
@flamerohr updated, but I've restored the non-static $rewriteHashlinks variable in SSViewer.

tractorcow:
Also, dontRewriteHashlinks() no longer irrepairably modifies global state :)

tractorcow:
Ok, cause of the issue was test state no longer being resettable solely via config nest / unnest; I've resolved this by adding a new SSViewerTestState to reset the new vars. :)

Issue #7358, opened 2017-09-10
Improve documentation for custom theme locations


tractorcow:
E.g. `SSViewer.themes` supports a variety of syntaxes, but these are not well documented at https://docs.silverstripe.org/en/4/developer_guides/templates/themes/

E.g.
- `silverstripe/module:subtheme`
- `silverstripe/module:/nestedpath`
- `silverstripe/module`
- `theme`
- `/pathfromroot`

It's also necessary to demonstrate how to override themes via `mysite/templates` folder.

tractorcow:
For reference the logic to document is implemented in `ThemeResourceLoader::getPath($theme)`

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-framework/pull/7395

PR #7395, opened 2017-09-22
WIP Vendor modules proof of concept

tractorcow:
Parent issue: https://github.com/silverstripe/silverstripe-framework/issues/7360

Also fixes https://github.com/silverstripe/silverstripe-framework/issues/7358 and includes accurate documentation on how to configure themes.

See two new projects (cli tool and composer plugin):

 - [x] https://github.com/silverstripe/vendor-plugin
 - [x] https://github.com/silverstripe/vendor-plugin-helper
 
 This can be tested with the below module:
 
 `composer require tractorcow/test-vendor-module ^1.0`

This demonstrates a working proof of concept for supporting silverstripe modules in /vendor dir.

The major changes include an updated module manifest, as well as additional URLGenerator which supports "remapping" of files from vendor to a top level 'resources' folder.

The idea is that, although not explicitly supported yet, this should allow us to move to a `public` folder later on down the track when we dis-associate the BASE_PATH with the web root folder.

Remapping can be configured to one or more of the following methods:
 - symlink to vendor
 - copy assets from vendor
 - url rewriting only
 
By default the first two options are configured, meaning that a failed symlink will fail over to a file copy. This does require that the 'resources' folder has write access, similar to 'assets'.

Modules in vendor are not automatically exposed; It's necessary for such a module to declare which folders are public.

```json
{
    "name": "tractorcow/anothermodule",
    "description": "a test module",
    "type": "silverstripe-vendormodule",
    "extra": {
        "expose": [
            "client"
        ]
    },
    "require": {
        "silverstripe/framework": "^4.0"
    }
}
```

This will trigger a symlink at the location `resources/tractorcow/anothermodule/client` linking to the vendor dir `vendor/tractorcow/anothermodule/client`.

Remapping is performed JIT if a url to a resource isn't available, and additionally on flush. I had considered doing this on dev/build, but it was necessary to build this into error-handling, such that ErrorControlChain will allow the remapping to occur if a critical resource is unavailable (e.g. css on the login page).


tractorcow:
Note: No changes to composer/installers necessary; As long as it doesn't match `silverstripe-module` it'll be installed normally.

src/Control/SimpleResourceURLGenerator.php
```diff
@@ -45,18 +78,42 @@ public function setNonceStyle($nonceStyle)
...
+            $exists = file_exists($absolutePath);
+        }
+        if (!$exists) {
             throw new InvalidArgumentException("File {$relativePath} does not exist");
         }
 
+        // Delegate path rewrites to FileRemapper if available
+        $fileRemapper = $this->getFileRemapper();
+        if ($fileRemapper) {
+            $relativePath = $fileRemapper->rewritePath($relativePath, $resource);
```
sminnee: Suggest FileRemapper is an interface exposing `rewritePath()`

src/Control/FileRemapper.php
```diff
@@ -0,0 +1,274 @@
...
+use SilverStripe\Core\Flushable;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Core\Manifest\Module;
+use SilverStripe\Core\Manifest\ModuleLoader;
+use SilverStripe\Core\Manifest\ModuleResource;
+
+/**
+ * Assists with exposing resources from vendor folder to public
+ */
+class FileRemapper implements Flushable
```
sminnee: I would call the default implementation "ComposerFileRemapper" or something.

src/Control/FileRemapper.php
```diff
@@ -0,0 +1,274 @@
...
+     *
+     * @param Module $module
+     */
+    protected function remapModule(Module $module)
+    {
+        if (!$this->hasRemapActions()) {
+            return;
+        }
+        // Get list of public folders to expose
+        $exposed = $module->getExposedFolders();
```
sminnee: IMO module should expose `getComposerData()` or `getComposerExtraData($key)`. Let this class take responsibility for deciding where in the composer file the settings are kept.

tractorcow:
@sminnee I agree; We are now thinking of moving this code out of framework, and putting it into a plugin.

Further discussion at https://github.com/silverstripe/silverstripe-framework/issues/7360#issuecomment-331322934

tractorcow:
Ok, a big update; All file-mapping has been shifted out of framework, and put into a new `silverstripe/vendor-plugin`. Any `silverstripe-vendormodule` which requires this plugin will automatically have their exposed folders automatically mapped.

Additionally, to support `--no-plugins` installations you can now use the new `silverstripe/vendor-helper` CLI tool to link these dirs after the fact (internally they use the same code, but the CLI has slightly heavier dependencies).

See:

 - https://github.com/open-sausages/vendor-helper
 - https://github.com/open-sausages/vendor-plugin

tractorcow:
@chillu would you like to review to check my approach works and fits with platform?

src/Core/Manifest/ModuleResource.php
```diff
@@ -0,0 +1,109 @@
...
+
+namespace SilverStripe\Core\Manifest;
+
+use InvalidArgumentException;
+use SilverStripe\Core\Injector\Injector;
+
+/**
+ * Experimental API; This may change at any time
+ */
+class ModuleResource
```
chillu: Describe what this API does anyway? :D

src/Core/Manifest/ModuleResource.php
```diff
@@ -0,0 +1,109 @@
...
+        $this->relativePath = ltrim($relativePath, Module::TRIM_CHARS);
+        if (empty($this->relativePath)) {
+            throw new InvalidArgumentException("Resource cannot have empty path");
+        }
+    }
+
+    /**
+     * @return string
+     */
+    public function getPath()
```
chillu: Note that for path operations, it'll get you the *original* file, not the copied one.

tests/php/i18n/i18nTextCollectorTest.php
```diff
@@ -712,11 +712,11 @@ public function testModuleDetection()
         $collector = new Collector();
         $modules = ModuleLoader::inst()->getManifest()->getModules();
         $this->assertEquals(
-            array(
+            [
                 'i18nnonstandardmodule',
+                'i18nothermodule',
```
chillu: Why has the order changed here? I assume that doesn't change the YAML config order, since that's determined independently of module loading order. Is there anything else that relies on it though?

chillu:
Holy crap, awesome work! I've looked through the two new modules on open-sausages, and they look solid.

I've had a look at how [laravel packages](https://laravel.com/docs/5.5/packages#public-assets) handle this again, and they also publish config files and other stuff that's not intended for website use. So you might copy `vendor/silverstripe/fluent/_config/default.yml` to `mysite/_config/fluent.yml`. Which means the mapped path depends on the file, and wouldn't be set globally to `resources/`. I guess there's a bit of conceptual overlap with the [recipe-plugin](https://github.com/silverstripe/recipe-plugin) and its `extra.project-files` composer config. I know this isn't part of the AC, and I can't think of a good use case outside recipes. Just thought it's worth a mention.

Also, since the process doesn't actually boot a SilverStripe app, there's no way to conditionally load a module. We don't do that at the moment, but I could see us getting there long term. For example, conditional on a feature flag in that particular environment. So the composer package would always be installed, but the module isn't determined as "active" during boot - hence no files should be copied either. That's the tradeoff keeping the script separate (and compatible with `composer --no-scripts --no-plugins`).

chillu:
A few minor bugs. I don't think any of them is blocking the merge here, but they're worth pointing out in the docs as gotchas:

 * If you switch from symlink to copy methods, it'll silently fail to add new files: https://github.com/open-sausages/vendor-plugin/blob/master/src/Methods/CopyMethod.php#L28
 * If you use copy, it won't delete files that have disappeared from the original vendor module.
 * If you use either copy or symlink, removing the module won't remove the files (there's a `pre-package-uninstall` [event](https://getcomposer.org/doc/articles/scripts.md#event-names))

src/Core/Manifest/ModuleResource.php
tractorcow: Since resources became more complex than a simple path, I've shifted it to an object that can be physically passed around independently of the module itself.

I'll add a class description.

src/Core/Manifest/ModuleResource.php
tractorcow: Yep, I'll document this in the code a bit more concisely.

tests/php/i18n/i18nTextCollectorTest.php
tractorcow: Alphabetisation is the default; I'm not sure WHY it was the way it was earlier. Perhaps it wasn't alphabetised properly?

Note: Module sorting is still applied after this and can be controlled via config.

tractorcow:
> I've had a look at how laravel packages handle this again, and they also publish config files and other stuff that's not intended for website use.

The way that modules work, these files will work fine regardless of whether they are in the root or not; Vendor .yml config should be picked up, but I will double check to make sure the non-module manifests are working properly. :)

tractorcow:
Task for today is to build a vendor-module and test it out to ensure it works properly.

tractorcow:
> If you switch from symlink to copy methods, it'll silently fail to add new files: https://github.com/open-sausages/vendor-plugin/blob/master/src/Methods/CopyMethod.php#L28

Ok, I'll remove the symlink in this case.

tractorcow:
You can check this out, and then install `https://github.com/tractorcow/test-vendor-module`

`composer require tractorcow/test-vendor-module ^1.0` and it will bootstrap a test page with a bunch of tests. :)

tractorcow:
> If you switch from symlink to copy methods, it'll silently fail to add new files: https://github.com/open-sausages/vendor-plugin/blob/master/src/Methods/CopyMethod.php#L28
> If you use copy, it won't delete files that have disappeared from the original vendor module.
If you use either copy or symlink, removing the module won't remove the files (there's a pre-package-uninstall event)

@chillu all of this feedback has been addressed. :)

tractorcow:
Updated with tests and documentation on themes.

docs/en/02_Developer_Guides/05_Extending/How_Tos/01_Publish_a_Module.md
```diff
@@ -17,39 +17,87 @@ this:
...
+```
+
+Note that these modules have the following distinct characteristics:
+
+ - Library type is `silverstripe-vendormodule`
+ - Any folder which should be exposed to the public webroot must be declared in the `extra.expose` config.
+   These paths will be automatically rewritten to public urls which don't directly serve files from the `vendor`
+   folder. For instance, `vendor/tractorcow/test-vendor-module/client` will be rewritten to
+   `resources/tractorcow/test-vendor-module/client`.
+ - Any module which uses the folder expose feature must require `silverstripe/vendor-plugin` in order to
```
chillu: Hmm, is that a requirement because of composer hooks? It feels a bit noisy to have this dependency eventually end up in every SilverStripe module. I guess it makes the semver constraints clearer than hiding it as an implicit dependency in core, but I don't see us changing the `expose` JSON format anytime soon - and that's really the only "API surface".

src/Core/Manifest/Module.php
```diff
@@ -152,63 +181,69 @@ protected function loadComposer()
...
-     * which may change in the near future.
-     *
-     * @internal Experimental API and may change
-     * @param string $path File or directory path relative to module directory
-     * @return string Path relative to base directory
+     * @deprecated 4.0...5.0 Use getResource($path)->getRelativePath() instead
+     * @param string $path
+     * @return string
+     */
+    public function getRelativeResourcePath($path)
```
chillu: Do we need to change other core modules to not use this deprecated API? :D

robbieaverill: Gah.

docs/en/02_Developer_Guides/05_Extending/How_Tos/01_Publish_a_Module.md
tractorcow: Yes that's right. We could move the dependency to framework, but it'll still be a module dependency in practice.

src/Core/Manifest/Module.php
tractorcow: We could but it's not critical; We should be forgiving if people treat previous betas as "final api";

chillu:
Our modules have the following structure (e.g. `silverstripe/campaign-admin`)

```
client/
  dist/
  lang/
    src/
  src/
```

So we'd expose `client/dist` and `client/lang` separately here, mapping to `resources/silverstripe-campaign-admin/dist` and `resources/silverstripe-campaign-admin/lang`? Technically that would also copy the `client/lang/src` folder, but that's probably due to bad folder design in the module rather than shortcomings of the vendor approach itself. An exclusion regex list would be useful at some point, but doesn't seem to be block merge.

tractorcow:
`resources/silverstripe-campaign-admin/client/dist` rather than `resources/silverstripe-campaign-admin/dist`

src/Core/Manifest/Module.php
chillu: Works for me, as long as you track that remaining work against the 4.0 stable milestone as a new ticket :)

tractorcow:
I just realised ALL of the 'vendor' test fixtures got lost due to `.gitignore`. Fixed now. :D

tractorcow:
@chillu it's green :)

tractorcow:
Rebased.

PR #7399, opened 2017-09-25
Reduce dependence on session state for accessing draft stages

dhensby:
This attempts to lay the foundations to move away from using the session to store the current stage.

This is an issue which is raised by clients regarding unexpectedly seeing staged content when navigating the site outside of the CMS or when wanting to open multiple tabs to see one draft and one live site in each.

### DIFF
--- a/src/Control/RequestHandler.php
+++ b/src/Control/RequestHandler.php
@@ -560,9 +560,11 @@ class RequestHandler extends ViewableData
         // Check configured url_segment
         $url = $this->config()->get('url_segment');
         if ($url) {
+            $link = Controller::join_links($url, $action, '/');
+
             // Give extensions the chance to modify by reference
             $this->extend('updateLink', $link, $action);
-            return Controller::join_links($url, $action, '/');
+            return $link;
         }
 
         // no link defined by default
--- a/src/Control/RequestHandler.php
+++ b/src/Control/RequestHandler.php
@@ -560,6 +560,8 @@ class RequestHandler extends ViewableData
         // Check configured url_segment
         $url = $this->config()->get('url_segment');
         if ($url) {
+            // Give extensions the chance to modify by reference
+            $this->extend('updateLink', $link, $action);
             return Controller::join_links($url, $action, '/');
         }
 
--- a/src/Dev/FunctionalTest.php
+++ b/src/Dev/FunctionalTest.php
@@ -4,6 +4,7 @@ namespace SilverStripe\Dev;
 
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\Director;
+use SilverStripe\Control\HTTP;
 use SilverStripe\Control\Session;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Core\Config\Config;
@@ -12,6 +13,7 @@ use SilverStripe\Security\BasicAuth;
 use SilverStripe\Security\Member;
 use SilverStripe\Security\Security;
 use SilverStripe\Security\SecurityToken;
+use SilverStripe\Versioned\Versioned;
 use SilverStripe\View\SSViewer;
 use PHPUnit_Framework_AssertionFailedError;
 use SimpleXMLElement;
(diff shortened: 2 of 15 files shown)


dhensby:
There's a bit of cross dependency on https://github.com/silverstripe/silverstripe-versioned/pull/45 because this requires the change in logic for permission checks.

src/Dev/FunctionalTest.php
```diff
@@ -160,6 +162,9 @@ protected function withBaseFolder($folder, $callback)
     public function get($url, $session = null, $headers = null, $cookies = null)
     {
         $this->cssParser = null;
+        if (Versioned::get_stage() === Versioned::DRAFT) {
```
tractorcow: Versioned isn't a dependency of framework, FYI. :D This will break in some framework-only installations.

dhensby: It is a Dev dependency and the test suite makes direct assumptions about it. So yes, whilst this will break if you don't have versioned installed the tests won't run anyway (as we've found in a number of modules)

dhensby: I've adjusted it slightly so there's less dependence on `Versioned`

tractorcow:
I left some comments over on https://github.com/silverstripe/silverstripe-versioned/pull/45.

Could this rewrite be a middleware instead of using php url_rewrite?

dhensby:
I've rebased this PR and tested by hand - it's working well IMO.

> Could this rewrite be a middleware instead of using php url_rewrite?

Using a middleware will mean either moving the `url_rewrite` call to that middleware (not changing much as it's global state) or writing our own HTML parser / URL re-writer, which I'm sure would be less efficient that a built in PHP method aimed at re-writing the output buffer and that is used to handle the appending of session vars when cookies aren't available.

tractorcow:
@dhensby I think we can keep using url_rewriter, but let's shift it into it's own middleware rather than tying it to versioned.

https://gist.github.com/tractorcow/f5b436a2d378cd5ed4ab51375bd97f24

This solution respects the middleware pattern, rather than relying no changes to global state being "eventually" evaluated.

tractorcow:
The next question is how we get versioned to "register" rewrites. It could talk to the service directly via injector, I guess, or it could register it directly on the HTTPResponse object.

That's what I'm hoping we can come up with a solution for together. :)

src/Control/RequestHandler.php
```diff
@@ -560,6 +560,8 @@ public function Link($action = null)
         // Check configured url_segment
         $url = $this->config()->get('url_segment');
         if ($url) {
+            // Give extensions the chance to modify by reference
+            $this->extend('updateLink', $link, $action);
```
tractorcow: oops :)

chillu: Derp derp

tractorcow:
I think we need to deprecate useDraftSite(), and rely on stage querystring from here on. It's becoming a real problem trying to support both. It's causing lots of CMS test regressions.

tractorcow:
Is this ok @chillu ?

```php
/**
     * Set this to true on your sub-class to use the draft site by default for every test in this class.
     *
     * @deprecated 4.2..5.0 Use ?stage=Stage in your ->get() querystring requests instead
     * @var bool
     */
    protected static $use_draft_site = false;
```

And same deprecation notice on useDraftSite()

sminnee:
@tractorcow potentially both the permanently-querystringed approach to versioning and the session-based approach to versioning could be 2 separate pieces of router middleware? Then you could pick one or the other rather having both to tread on each other's toes?

sminnee:
> Using a middleware will mean either moving the url_rewrite call to that middleware (not changing much as it's global state) or writing our own HTML parser / URL re-writer, which I'm sure would be less efficient that a built in PHP method aimed at re-writing the output buffer and that is used to handle the appending of session vars when cookies aren't available.

Rather than rewriting the outgoing HTML, could we override the behaviour of `SiteTree::Link()`?

chillu:
> protected static $use_draft_site = false;

So that's in `FunctionalTest`. I think it's fine to rely on query strings for fake HTTP requests there, instead of allowing to call `useDraftSite()` procedurally.

> potentially both the permanently-querystringed approach to versioning and the session-based approach to versioning could be 2 separate pieces of router middleware?

Where is the advantage of making this a middleware rather than a config switch? Neither of these would directly modify the HTTP request or response, but rather perform config settings (session or class statics) which indirectly influence the output (alongside a myriad other settings). I don't see that as good or consistent use of middleware?

tractorcow:
Ok, after some intensive testing I've got this more or less working with most of the modules green (all of the ones I've tested at least).

Tomorrow I'll finish testing framework and run behat and we can merge this in.

PR #7401, opened 2017-09-26
FIX Class name in _t() call in installer and run text collector

robbieaverill:


tractorcow:
Can be merged once tests are green.

PR #7403, opened 2017-09-26
NEW Add actWithPermission to SapphireTest

dhensby:
Currently you need to be logged in to programmatically publish objects/changesets.

In a test suite this can be a bit annoying as it either means creating a specific user and using `Member::actAs` or you have to use the `logInWithPermission` helper in sapphiretest.  Both can be undesirable as the former requires creating member objects in test code, which is noisy and wasteful, and the latter means if you're logged in already you'll have to re-login again.

I've added a method `actWithPermission` which is a helper for using `Member::actAs` with a single permission code rather than using a Member object.

I've also split out the logic for creating Group/Member objects from the `logInWithPermission` method so it can be re-used.

### DIFF
--- a/src/Dev/SapphireTest.php
+++ b/src/Dev/SapphireTest.php
@@ -950,13 +950,24 @@ class SapphireTest extends PHPUnit_Framework_TestCase implements TestOnly
     }
 
     /**
-     * Create a member and group with the given permission code, and log in with it.
-     * Returns the member ID.
+     * A wrapper for automatically performing callbacks as a user with a specific permission
      *
-     * @param string|array $permCode Either a permission, or list of permissions
-     * @return int Member ID
+     * @param string|array $permCode
+     * @param callable $callback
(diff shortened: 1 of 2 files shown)


tractorcow:
Looks good to me.

wernerkrauss:
@dhensby the `testCreateMemberWithPermission` runs well the first time you run `SapphireTestTest`, but re-running it in the same environment doesn't clear the DB, therefor it fails:

> Failed asserting that actual size 1 matches expected size 0.
>  /vagrant/www/framework/tests/php/Dev/SapphireTestTest.php:49
>  /vagrant/www/vendor/phpunit/phpunit/phpunit:52

Any chance to clear the DB for this one test only?

Adding "uses_database" to the test increases the time for the tests from about 1.6s to 6.5s...

dhensby:
hmmm - which DB is it writing to?

Issue #5900, opened 2016-08-18
Alphabetical Sort


thezenmonkey:
I've noticed this in Sorting in ArrayList, though it possbily applies to others. When doing a sort on Varchar or Text column it should use natural order. Currently` ArrayList()->sort('Title ASC')` will return the list as case sensitive so you get A-Z followed by a-. Expected reslut shoudl be (A|a)-(Z|z) 
{edit typos}

PR #7406, opened 2017-09-26
[DO NOT MERGE] Fix ArrayList sort error with old (supported) PHP

NightJar:
PHP 5.3 at least (the reported and tested against version) requires arguments to `call_user_func_array` to be passed by reference. There exists a note as a comment in the code, but was unfortunately overlooked in a previous commit to fix case sensitive sorting https://github.com/silverstripe/silverstripe-framework/commit/4998b8044530a83c617194d544b76a98f742386e#diff-6ba746c3d31fd6b4c4a99d7efe35eb21L442

To solve this issue we simply first assign the constant to a variable, so we can then pass that by reference. This has no functional impact, however fixes an issue for users locked in to old PHP versions which we still list as supported (https://docs.silverstripe.org/en/3/getting_started/server_requirements/#web-server-software-requirements).

As requested by @dhensby please tag this PR as DO NOT MERGE

### DIFF
--- a/model/ArrayList.php
+++ b/model/ArrayList.php
@@ -448,7 +448,8 @@ class ArrayList extends ViewableData implements SS_List, SS_Filterable, SS_Sorta
 			// First argument is the direction to be sorted,
 			$multisortArgs[] = &$sortDirection[$column];
 			if ($firstRun) {
-				$multisortArgs[] = SORT_REGULAR;
+				$sortArg = SORT_REGULAR;
+				$multisortArgs[] = &$sortArg;
 			}
 			$firstRun = false;
 		}


dhensby:
Original PR: https://github.com/silverstripe/silverstripe-framework/pull/5903

Original issue: https://github.com/silverstripe/silverstripe-framework/issues/5900

@jonom - I know you looked into regressions with my changes quite a bit - would you mind casting an eye on this?

@kinglozzer any thoughts given you were also a little involved in my original patch?

dhensby:
@NightJar how does this manifest itself on old versions of PHP? the test suite runs against 5.3 and doesn't appear to have any existing issues.

NightJar:
capture

Yeah it's certainly a thing.
Apache 2.2, PHP 5.3.3

capture

Why Travis is so awesome it appears to be entirely backwards, I'm not sure. Will investigate.

NightJar:
So Travis has failed this PR on PHP 5.6 and PHP 7, Because some part of the harness failed to connect to the D-Bus daemon. This _appears_ wholey unrelated.

```                                                                                                                                
  [Behat\Mink\Exception\DriverException]                                                                                         
  Exception has been thrown in "beforeScenario" hook, defined in SilverStripe\Framework\Test\Behaviour\FeatureContext::before()  
  Could not open connection: Unable to connect to host localhost on port 7055 after 45000 ms. Firefox console output:            
  //bin/dbus-launch terminated abnormally without any error message                                                              
  (firefox:8417): GConf-WARNING **: Client failed to connect to the D-BUS daemon:                                                
  //bin/dbus-launch terminated abnormally without any error message 
```

So seems like it's the tests that aren't good enough.
`49 tests run: 40 passes, 0 failures, and 0 incomplete with 9 errors`
Should be 9 failures. Although they're only warnings, the result is that the sort does not happen.
Digging further.

NightJar:
Right, updated tests to hard fail tests in the case I'm also fixing.
Ultimately I guess it's Travis (?) that should be altered to fail a build on any Error output from PHPUnit, not only on Failures. Not sure how to approach that though.

dhensby:
OK - testing locally I get this output on my centos 6 machine ;)

```sh
# vendor/bin/phpunit framework/tests/model/ArrayListTest.php
PHPUnit 3.7.38 by Sebastian Bergmann.

Configuration read from /var/www/html/phpunit.xml.dist

...................EEEEEEEEE.....................

Time: 214 ms, Memory: 24.00MB

There were 9 errors:

1) ArrayListTest::testSortSimpleDefaultIsSortedASC
Parameter 3 to array_multisort() expected to be a reference, value given

/var/www/html/framework/model/ArrayList.php:461
/var/www/html/framework/tests/model/ArrayListTest.php:255
/var/www/html/vendor/phpunit/phpunit/composer/bin/phpunit:63

2) ArrayListTest::testMixedCaseSort
Parameter 3 to array_multisort() expected to be a reference, value given

/var/www/html/framework/model/ArrayList.php:461
/var/www/html/framework/tests/model/ArrayListTest.php:344
/var/www/html/vendor/phpunit/phpunit/composer/bin/phpunit:63

3) ArrayListTest::testSortSimpleASCOrder
Parameter 3 to array_multisort() expected to be a reference, value given

/var/www/html/framework/model/ArrayList.php:461
/var/www/html/framework/tests/model/ArrayListTest.php:368
/var/www/html/vendor/phpunit/phpunit/composer/bin/phpunit:63

4) ArrayListTest::testSortSimpleDESCOrder
Parameter 3 to array_multisort() expected to be a reference, value given

/var/www/html/framework/model/ArrayList.php:461
/var/www/html/framework/tests/model/ArrayListTest.php:415
/var/www/html/vendor/phpunit/phpunit/composer/bin/phpunit:63

5) ArrayListTest::testSortNumeric
Parameter 3 to array_multisort() expected to be a reference, value given

/var/www/html/framework/model/ArrayList.php:461
/var/www/html/framework/tests/model/ArrayListTest.php:466
/var/www/html/vendor/phpunit/phpunit/composer/bin/phpunit:63

6) ArrayListTest::testReverse
Parameter 3 to array_multisort() expected to be a reference, value given

/var/www/html/framework/model/ArrayList.php:461
/var/www/html/framework/tests/model/ArrayListTest.php:497
/var/www/html/vendor/phpunit/phpunit/composer/bin/phpunit:63

7) ArrayListTest::testSimpleMultiSort
Parameter 3 to array_multisort() expected to be a reference, value given

/var/www/html/framework/model/ArrayList.php:461
/var/www/html/framework/tests/model/ArrayListTest.php:514
/var/www/html/vendor/phpunit/phpunit/composer/bin/phpunit:63

8) ArrayListTest::testMultiSort
Parameter 3 to array_multisort() expected to be a reference, value given

/var/www/html/framework/model/ArrayList.php:461
/var/www/html/framework/tests/model/ArrayListTest.php:530
/var/www/html/vendor/phpunit/phpunit/composer/bin/phpunit:63

9) ArrayListTest::testSortWithCircularDependencies
Parameter 3 to array_multisort() expected to be a reference, value given

/var/www/html/framework/model/ArrayList.php:461
/var/www/html/framework/tests/model/ArrayListTest.php:560
/var/www/html/vendor/phpunit/phpunit/composer/bin/phpunit:63

FAILURES!
Tests: 49, Assertions: 120, Errors: 9.
```

dhensby:
ok - so I compiled 5.3.29 (which is what travis runs) and it's passing without your patch:

```sh
# vendor/bin/phpunit framework/tests/model/ArrayListTest.php
PHPUnit 3.7.38 by Sebastian Bergmann.

Configuration read from /var/www/html/phpunit.xml.dist

.................................................

Time: 520 ms, Memory: 23.50MB

OK (49 tests, 152 assertions)
```

dhensby:
I can confirm this patch fixes the error on 5.3.3

NightJar:
> @NightJar that change won't make any difference because if warnings are already converted to exceptions they will force an error in the test runner.
>
> I've tested locally with the same version of PHPUnit and warnings cause test errors.
>
> The only things that could be different is if the builds on travis are turning off this feature... but that would be nuts :/

Indeed it would be nuts. Also nuts is that it appears by default (at least with the version of PHPUnit installed with 3.6.1, tested locally also) that a Warning level error does not appear to count as a failure. So you get 9 unsuccessful tests, but no failures. The PHPUnit results still come back red as opposed to green, but the travis build passes... so... ¯\\_(ツ)_/¯

As you say it doesn't make a difference overall, but it would make the build fail I should hope :)

dhensby:
The warnings mark the tests as "errored" which fails the test suite - it's all good. The reason travis passes is because it's not testing a version of PHP 5.3 that has this issue.

PR #7407, opened 2017-09-27
BUG Config updates are now applied after middleware not before

tractorcow:
Requires https://github.com/silverstripe/silverstripe-config/pull/23
Parent issue: https://github.com/silverstripe/silverstripe-config/issues/22
Fixes issue found in https://github.com/silverstripe/silverstripe-framework/pull/7390


tractorcow:
I've shifted the middleware flag to framework, and this is now injected by CoreConfigFactory (which itself injects a callback as a nest middleware).

wilr:
Have also reviewed and the original issue is resolved with this 👍  Not sure about the test failure. Seems unrelated and odd.

tractorcow:
Bug was from a regression in ExtensionMiddleware.

```diff
-$extensionSourceConfig = Config::inst()->get($class, null, $excludeMiddleware | $this->disableFlag);
+$extensionSourceConfig = Config::inst()->get($class, null, Config::UNINHERITED | $excludeMiddleware | $this->disableFlag);
```

Fixed it.

PR #7408, opened 2017-09-27
Fix links on paginated lists when there are GET vars

ajoneil:
Prior to this change, if there were already GET vars on a page
with a PaginatedList, the links would include a mix of `&` and `&amp;`.


dhensby:
OK - I've merged this by hand and fixed the test assertions (I made them a bit more "robust" - in that they don't care about the complete URL and just that the query params exist correctly) and I've fixed the code style issues too.

Thanks!

Issue #7202, opened 2017-07-21
RFC Move test suite to Codeception


dhensby:
## Problem

Currently our test suites are either mangled together (unit tests and functional tests all run off the same core `SapphireTest` dependency) or completely separate (behat).

Whilst our unit and functional tests are built on top of PHPUnit, our use of `SapphireTest` is still somewhat idiosyncratic and confusing for newcomers. Behat tests are generally seen by developers/the community as difficult to get running and a scary/different method to write.

The separation of our behat tests from our other tests also means we have separate ways of setting up fixtures for tests, managing sessions, and  managing test behaviour. Of course, there's also a new syntax (Gherkin) to learn for writing behat tests which requires a good knowledge of SilverStripe's pre-defined features and contexts.

Summary:
 - High barrier to entry for acceptance testing
 - Poor documentation around testing
 - New syntax and approach needs to be learnt for each test type

### Deliver a phased approach to move over to Codeception as preferred test suite throughout 4.x lifetime.

I suggest a phased approach where we aim to move first unit tests (4.1), then functional tests (4.2), then acceptance tests (4.3) over to SilverStripe.

We would keep `SapphireTest` for the life of 4.x so that external projects that depend on it are not forced to migrate and legacy tests continue to run.

As we phase our test suite over new tests (new test classes) will need to be written in Codeception and amendments to existing tests will have to be made to codeception tests if they've been migrated.

Pros:
 - Unified test suite, syntax and approach across all test suites
 - Better external documentation to guide developers
 - More powerful and modular testing tools
 - Lower barrier to entry for all types of tests
 - Can retain gherkin tests

Cons:
 - Migration of tests to new format
 - Migration of behat fixtures/contexts 

### Move to Codeception in SS 5.x

I feel this isn't feasible as the 5.x delivery is so far away someone will have to commit to re-writing every new test into codeception for 5.x which is years of commitment. Otherwise it means putting this off for years.

Pros:
 - We won't run two test suites side-by-side

Cons:
 - Big delay to core support
 - Large amount of effort to migrate new tests over the coming years

### Migrate smaller core modules to Codeception and leave framework to last

We could focus on migrating smaller modules to Codeception over framework which would allow us to build the support for all three test types more quickly.

Pros:
 - All 3 test types supported relatively quickly

Cons:
 - Fragmented testing standards across modules
 - Dual support of fixtures/contexts across modules and core

#### Terminology

[...]

PR #7409, opened 2017-09-27
WIP: Rename assert dos to assert list

wernerkrauss:
As discussed in [Slack](https://silverstripe-users.slack.com/archives/C6PLF83H9/p1506493049000056) I renamed SapphireTest's `assertDOS*()` methods to `assertList*()` 

Todo:
- [x] write tests for that assertions
- [x] make assert* static methods
- [x] refactor assert* to show a meaningful failure message and the actual test location (throw exception)
- - [x] assertListAllMatch
- - [x] assertListContains
- - [x] assertListNotContains
- - [x] assertListEquals
- [x] write test for new Constraint classes
- [ ] should we rename and deprecate SapphireTest also, as this is a relict from good ol' 2.x days?

### DIFF
--- a/src/Dev/Constraint/SSListContains.php
+++ b/src/Dev/Constraint/SSListContains.php
@@ -6,6 +6,7 @@ use PHPUnit_Framework_Constraint;
 use PHPUnit_Framework_ExpectationFailedException;
 use SilverStripe\Dev\SSListExporter;
 use SilverStripe\Dev\TestOnly;
+use SilverStripe\ORM\SS_List;
 use SilverStripe\View\ViewableData;
 
 if (!class_exists(PHPUnit_Framework_Constraint::class)) {
@@ -48,7 +49,7 @@ class SSListContains extends PHPUnit_Framework_Constraint implements TestOnly
      * a boolean value instead: true in case of success, false in case of a
      * failure.
      *
-     * @param mixed $other Value or object to evaluate.
+     * @param SS_List $other Value or object to evaluate.
      * @param string $description Additional information about the test
      * @param bool $returnResult Whether to return a result or throw an exception
      *
@@ -61,7 +62,7 @@ class SSListContains extends PHPUnit_Framework_Constraint implements TestOnly
         $success = true;
 
         foreach ($other as $item) {
-            $this->checkIfItemEvaltuatesRemainingMatches($item);
+            $this->checkIfItemEvaluatesRemainingMatches($item);
         }
 
         //we have remaining matches?
@@ -85,7 +86,7 @@ class SSListContains extends PHPUnit_Framework_Constraint implements TestOnly
      * @param ViewableData $item
      * @return bool
      */
-    protected function checkIfItemEvaltuatesRemainingMatches(ViewableData $item)
+    protected function checkIfItemEvaluatesRemainingMatches(ViewableData $item)
     {
         $success = false;
         foreach ($this->matches as $key => $match) {
--- a/src/Dev/Constraint/SSListContains.php
+++ b/src/Dev/Constraint/SSListContains.php
@@ -7,15 +7,13 @@ use SilverStripe\Dev\SSListExporter;
 use SilverStripe\View\ViewableData;
 
 /**
- * Constraint for checking if a SS_List contains only items matching the given
- * key-value pairs.  Each match must correspond to 1 distinct record.
- *
- * @todo can this be solved more elegantly using a Comparator?
+ * Constraint for checking if a SS_List contains items matching the given
+ * key-value pairs.
  *
  * Class SSListContainsOnly
  * @package SilverStripe\Dev\Constraint
  */
-class SSListContainsOnly extends \PHPUnit_Framework_Constraint
+class SSListContains extends \PHPUnit_Framework_Constraint
 {
 
     private $matches = [];
@@ -55,15 +53,11 @@ class SSListContainsOnly extends \PHPUnit_Framework_Constraint
         $success = true;
 
         foreach ($other as $item) {
-            if (!$this->checkIfItemEvaltuatesRemainingMatches($item)) {
-                $this->item_not_matching = true;
-                $success = false;
-                break;
-            }
+            $this->checkIfItemEvaltuatesRemainingMatches($item);
         }
 
         //we have remaining matches?
-        if (!$this->item_not_matching && count($this->matches) !== 0) {
+        if (count($this->matches) !== 0) {
             $success = false;
             $this->has_leftover_items = true;
         }
@@ -105,10 +99,7 @@ class SSListContainsOnly extends \PHPUnit_Framework_Constraint
      */
     public function toString()
     {
-        $stub = $this->item_not_matching
-            ? ' contains an item matching '
-            : " contained only the given items, the following items were left over:\n";
-
+        $stub = ' contains an item matching ';
 
         $matchToString = function ($key, $value) {
             return ' "' . $key . '" is "' . $value . '"';
--- a/src/Dev/Constraint/SSListContains.php
+++ b/src/Dev/Constraint/SSListContains.php
@@ -9,9 +9,6 @@ use SilverStripe\View\ViewableData;
 /**
  * Constraint for checking if a SS_List contains items matching the given
  * key-value pairs.
- *
- * Class SSListContainsOnly
- * @package SilverStripe\Dev\Constraint
  */
 class SSListContains extends \PHPUnit_Framework_Constraint
 {
--- a/src/Dev/Constraint/SSListContains.php
+++ b/src/Dev/Constraint/SSListContains.php
@@ -52,7 +52,7 @@ class SSListContains extends PHPUnit_Framework_Constraint implements TestOnly
      * @param string $description Additional information about the test
      * @param bool $returnResult Whether to return a result or throw an exception
      *
-     * @return mixed
+     * @return null|bool
      *
      * @throws PHPUnit_Framework_ExpectationFailedException
      */
@@ -77,6 +77,8 @@ class SSListContains extends PHPUnit_Framework_Constraint implements TestOnly
         if (!$success) {
             $this->fail($other, $description);
         }
+
+        return null;
     }
 
     /**
--- a/src/Dev/Constraint/SSListContains.php
+++ b/src/Dev/Constraint/SSListContains.php
@@ -18,9 +18,17 @@ if (!class_exists(PHPUnit_Framework_Constraint::class)) {
  */
 class SSListContains extends PHPUnit_Framework_Constraint implements TestOnly
 {
+    /**
+     * @var array
+     */
     protected $matches = [];
 
-    protected $has_leftover_items = false;
+    /**
+     * Check if the list has left over items that don't match
+     *
+     * @var bool
+     */
+    protected $hasLeftoverItems = false;
 
     public function __construct($matches)
     {
@@ -59,7 +67,7 @@ class SSListContains extends PHPUnit_Framework_Constraint implements TestOnly
         //we have remaining matches?
         if (count($this->matches) !== 0) {
             $success = false;
-            $this->has_leftover_items = true;
+            $this->hasLeftoverItems = true;
         }
 
         if ($returnResult) {
--- a/src/Dev/Constraint/SSListContains.php
+++ b/src/Dev/Constraint/SSListContains.php
@@ -2,17 +2,22 @@
 
 namespace SilverStripe\Dev\Constraint;
 
+use PHPUnit_Framework_Constraint;
 use PHPUnit_Framework_ExpectationFailedException;
 use SilverStripe\Dev\SSListExporter;
+use SilverStripe\Dev\TestOnly;
 use SilverStripe\View\ViewableData;
+
+if (!class_exists(PHPUnit_Framework_Constraint::class)) {
+    return;
+}
 
 /**
  * Constraint for checking if a SS_List contains items matching the given
  * key-value pairs.
  */
-class SSListContains extends \PHPUnit_Framework_Constraint
+class SSListContains extends PHPUnit_Framework_Constraint implements TestOnly
 {
-
     protected $matches = [];
 
     protected $has_leftover_items = false;
--- a/src/Dev/Constraint/SSListContains.php
+++ b/src/Dev/Constraint/SSListContains.php
@@ -16,11 +16,9 @@ use SilverStripe\View\ViewableData;
 class SSListContains extends \PHPUnit_Framework_Constraint
 {
 
-    private $matches = [];
+    protected $matches = [];
 
-    private $item_not_matching = false;
-
-    private $has_leftover_items = false;
+    protected $has_leftover_items = false;
 
     public function __construct($matches)
     {
@@ -75,7 +73,7 @@ class SSListContains extends \PHPUnit_Framework_Constraint
      * @param ViewableData $item
      * @return bool
      */
-    private function checkIfItemEvaltuatesRemainingMatches(ViewableData $item)
+    protected function checkIfItemEvaltuatesRemainingMatches(ViewableData $item)
     {
         $success = false;
         foreach ($this->matches as $key => $match) {
@@ -99,8 +97,6 @@ class SSListContains extends \PHPUnit_Framework_Constraint
      */
     public function toString()
     {
-        $stub = ' contains an item matching ';
-
         $matchToString = function ($key, $value) {
             return ' "' . $key . '" is "' . $value . '"';
         };
@@ -120,6 +116,10 @@ class SSListContains extends \PHPUnit_Framework_Constraint
             array_map($matchesToString, $this->matches));
 
 
-        return $stub . $allMatchesAsString;
+        return $this->getStubForToString() . $allMatchesAsString;
+    }
+
+    protected function getStubForToString() {
+        return ' contains an item matching ';
     }
 }
--- a/src/Dev/Constraint/SSListContains.php
+++ b/src/Dev/Constraint/SSListContains.php
@@ -113,13 +113,15 @@ class SSListContains extends \PHPUnit_Framework_Constraint
 
         $allMatchesAsString = implode(
             "\n or ",
-            array_map($matchesToString, $this->matches));
+            array_map($matchesToString, $this->matches)
+        );
 
 
         return $this->getStubForToString() . $allMatchesAsString;
     }
 
-    protected function getStubForToString() {
+    protected function getStubForToString()
+    {
         return ' contains an item matching ';
     }
 }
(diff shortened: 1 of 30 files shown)


dhensby:
I've recently looked at the `assertDOS*` methods and found that (in core) they are rarely used and, where they are, they can be pretty easily replaced with a vanilla assertion.

Is there a wider motivation for providing this functionality over encouraging the use of vanilla assertions?

src/Dev/SapphireTest.php
```diff
@@ -594,24 +594,24 @@ public function assertEmailSent($to, $from = null, $subject = null, $content = n
...
-     *         array('Email' => '...@example.com'),
-     *         array('Email' => 'i...@example.com'),
-     *      ), $members);
+     *      $this->assertListContains([
+     *         ['Email' => '...@example.com'],
+     *         ['Email' => 'i...@example.com'],
+     *      ], $members);
      */
-    public function assertDOSContains($matches, $dataObjectSet)
+    public function assertListContains($matches, $list)
```
dhensby: assertions should be static methods

wernerkrauss:
> Is there a wider motivation for providing this functionality over encouraging the use of vanilla assertions?

Of course it's basically just syntactic sugar, but

* It's already there
* It makes your tests more readable, clear and understandable when you don't loop over your list in every test; e.g. testing a `->filter()` in your getter method becomes more easy.
* Core might not use that functionality so much, other modules rely more on DataObjects and SS_Lists, therefore a shortcut is handy
* If people don't know about it, they don't use it; hence the initial renaming

wernerkrauss:
> Is there a wider motivation for providing this functionality over encouraging the use of vanilla assertions?

Of course it's just syntactic sugar, but

* It's already there
* It makes your tests more readable, clear and understandable when you don't loop over your list in every test; e.g. testing a `->filter()` in your getter method becomes much less code.
* Core might not use that functionality so much, other modules rely more on DataObjects and SS_Lists, therefore a shortcut is handy
* If people don't know about it, they don't use it; hence the initial renaming

dhensby:
hmm - the premise of those points is that this is a good feature and therefore we should update because of those reasons - but I feel like it's a bad feature and actually a bit opaque as to what's going on. It hides the logic of the assertion because it most definitely doesn't assert lists are equal in the sense that `assertEquals` does.

It's badly named as it's really `assertListABitLikeThis` and I feel doing something like:

```php
$this->assertEquals([ 'ID' => 3], $list->map('ID')->toArray());
```

is actually far clearer for showing what's going on.

This is somewhat related to #7202 where hopefully we'll move away from our own base SapphireTest that contains a lot of custom logic and tight coupling and move to something a bit more generic

sminnee:
Even better:

```php
$this->assertEquals([ 'ID' => 3], $list->column('ID'));
```

Perhaps we deprecate it entirely recommending this instead?

The rest of the PR tidying up the assertion methods makes sense, though.

dhensby:
@sminee - your example doesn't retain order so it's not quite as good ;)

As for tidy up, yep - happy with that.

I'm in favour of deprecating this behaviour.

tractorcow:
assertDOSEquals is used at least 50 times in Versioned module for tests; It's necessary to do partial matching on sets of objects matched by textual field values; This feature provides a necessary behaviour that allows you to match lists by human-readable properties; Matching by ID only makes tests a lot harder to follow.

E.g.

```
$this->assertDOSEquals(
    [
        ['Title' => 'Page 2a - v2'],
        ['Title' => 'Page 2b'],
    ],
    $page2v2->Children()
);
```

vs


```
$this->assertDOSEquals(
    [
        4,
        5
    ],
    $page2v2->Children()->column('ID')
);
```

And also misses the ability to match multiple fields safely:

```php
$this->assertDOSEquals(
    [
        [
        	'Title' => 'Page 2a - v2',
        	'Version' => 2
       	],
        [
       		'Title' => 'Page 2a',
       		'Version' => 1
      	],
    ],
    $page2v2->AllVersions()
);
```

src/Dev/SapphireTest.php
```diff
@@ -628,52 +628,64 @@ public function assertDOSContains($matches, $dataObjectSet)
...
         }
     }
+
+    /**
+     * @deprecated 4.0.0:5.0.0 Use assertListContains() instead
+     *
+     * @param $matches
+     * @param $dataObjectSet
+     */
+    public static function assertDOSContains($matches, $dataObjectSet) {
```
tractorcow: Shall we leave the deprecated methods with the same type? (i.e. non-static). It makes the upgrade path slightly safer. :)

src/Dev/SapphireTest.php
```diff
@@ -871,16 +920,16 @@ private function dataObjectArrayMatch($item, $match)
...
-     * Helper function for the DOS matchers
+     * Helper function for the List matchers
      *
-     * @param SS_List|array $dataObjectSet
+     * @param SS_List|array $list
      * @param array $match
      * @return string
      */
-    private function DOSSummaryForMatch($dataObjectSet, $match)
+    private static function ListSummaryForMatch($list, $match)
```
tractorcow: lowerCamelCase method name please. ;)

src/Dev/SapphireTest.php
```diff
@@ -594,32 +594,32 @@ public function assertEmailSent($to, $from = null, $subject = null, $content = n
...
-     *         array('Email' => '...@example.com'),
-     *         array('Email' => 'i...@example.com'),
-     *      ), $members);
+     *      $this->assertListContains([
+     *         ['Email' => '...@example.com'],
+     *         ['Email' => 'i...@example.com'],
+     *      ], $members);
      */
-    public function assertDOSContains($matches, $dataObjectSet)
+    public static function assertListContains($matches, $list)
```
tractorcow: Most phpunit assertions allow you to customise the error message; Shall we add this in while we are adding a new method?.

tests/php/Dev/SapphireTestTest.php
```diff
@@ -47,8 +53,155 @@ public function testActWithPermission()
...
     {
-        $this->assertCount(0, Member::get()->filter([ 'Email' => 'TESTPERM@example.org' ]));
+        $this->assertCount(0, Member::get()->filter(['Email' => 'TESTPERM@example.org']));
         $this->createMemberWithPermission('TESTPERM');
-        $this->assertCount(1, Member::get()->filter([ 'Email' => 'TESTPERM@example.org' ]));
+        $this->assertCount(1, Member::get()->filter(['Email' => 'TESTPERM@example.org']));
+    }
+
+    /**
+     * @dataProvider \SilverStripe\Dev\Tests\SapphireTestTest\DataProvider::provideAllMatchingList()
```
tractorcow: I like this good job. :)

tests/php/i18n/i18nTextCollectorTest.php
```diff
@@ -71,6 +71,10 @@ public function testConcatenationInEntityValues()
         );
     }
 
+    /**
+     * @expectedException PHPUnit_Framework_Error_Notice
+     * @expectedExceptionMessage Missing localisation default for key i18nTestModule.INJECTIONS_3
```
tractorcow: Since this test mixes both asserts and expected exceptions, I suggest replacing this with `$this->expectException() / $this->expectExceptionMessage()` rather than annotations.

tests/php/ORM/DataObjectTest.php
```diff
@@ -1405,27 +1411,37 @@ public function testPopulateDefaults()
...
     public function testValidateModelDefinitionsFailsWithIntValue()
     {
         Config::modify()->merge(DataObjectTest\Team::class, 'many_many', array('Players' => 12));
-        $this->setExpectedException(InvalidArgumentException::class);
         DataObject::getSchema()->manyManyComponent(DataObjectTest\Team::class, 'Players');
     }
 
+    /**
+     * @expectedException \InvalidArgumentException
+     * @expectedExceptionMessage Controller is not a valid subclass of DataObject
```
tractorcow: As mentioned elsewhere, for mixed tests it's advisable to PHP api to set expected exception prior to the throwing expression, rather than annotations.

wernerkrauss:
@tractorcow regarding custom error message: the custom assertions already have predefined error messages, should the custom message overwrite them? Also should it output the expected and actual list?

tractorcow:
Yeah the message can fall back to the auto-generated one if not provided (I would expect).

Ideally we would output the diff between actual-expected; Similar to how PHPUnit does it internally, but I'll accept this maybe a lot of work to perfect so I wouldn't expect you to go so far. :)

wernerkrauss:
Current state of progress on adding meaningful error messages:

```
Failed asserting that SilverStripe\ORM\ArrayList &00000000150163e400000000796f9c33 (
        SilverStripe\Security\Member => Array &0 (
            'FirstName' => 'Ingo'
            'Surname' => 'Schommer'
            'Locale' => 'en_US'
        )
 
        SilverStripe\Security\Member => Array &0 (
            'FirstName' => 'Sam'
            'Surname' => 'Minnee'
            'Locale' => 'en_US'
        )
 
)  contains an item matching ( "FirstName" is "IngoXXX" and  "Surname" is "Schommer")
 or ( "FirstName" is "Sam" and  "Surname" is "MinneeXXX").
```

tractorcow:
> Current state of progress on adding meaningful error messages:

That's looking really good @wernerkrauss I'm impressed with your work so far. :)

src/Dev/Constraint/SSListContains.php
```diff
@@ -0,0 +1,126 @@
...
+
+use PHPUnit_Framework_ExpectationFailedException;
+use SilverStripe\Dev\SSListExporter;
+use SilverStripe\View\ViewableData;
+
+/**
+ * Constraint for checking if a SS_List contains items matching the given
+ * key-value pairs.
+ *
+ * Class SSListContainsOnly
```
tractorcow: For consistency can we please remove `@package` and all the IDE-added headers to these classes? We are trying to arrange our PHP api docs (auto-built) by namespace rather than package these days.

src/Dev/Constraint/SSListContains.php
```diff
@@ -0,0 +1,126 @@
...
+ * @package SilverStripe\Dev\Constraint
+ */
+class SSListContains extends \PHPUnit_Framework_Constraint
```
tractorcow: If phpunit isn't installed this code may run the risk of breaking a production environment due to the manifest builder.

Any code outside of the `tests` folder is pretty much at risk of this.

See the fixes we added to SapphireTest that protect against this:

 - Add `TestOnly` interface to all phpunit-dependant classes. This will need to be on all classes down the hierarchy, in case you subclass this. :)
 - Add `if (!class_exists(PHPUnit_Framework_Constraint::class)) { return; }` before the class definition.

src/Dev/Constraint/SSListContains.php
```diff
@@ -0,0 +1,126 @@
...
+{
+
+    protected $matches = [];
+
+    protected $has_leftover_items = false;
```
tractorcow: Can we please use lowerCamelCase (psr-2) for new protected vars, and a phpdoc would be great if you like. :)

src/Dev/Constraint/SSListContainsOnly.php
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Constraint for checking if a SS_List contains only items matching the given
+ * key-value pairs.  Each match must correspond to 1 distinct record.
+ *
+ * @todo can this be solved more elegantly using a Comparator?
+ *
+ * Class SSListContainsOnly
+ * @package SilverStripe\Dev\Constraint
+ */
+class SSListContainsOnly extends SSListContains
```
tractorcow: As above, `TestOnly` and conditional. :)

src/Dev/Constraint/ViewableDataContains.php
```diff
@@ -0,0 +1,76 @@
+<?php
+
+namespace SilverStripe\Dev\Constraint;
+
+use PHPUnit_Util_InvalidArgumentHelper;
+
+class ViewableDataContains extends \PHPUnit_Framework_Constraint
```
tractorcow: As above, `TestOnly` and conditional. :)

src/Dev/SapphireTest.php
```diff
@@ -280,6 +285,50 @@ protected function setUp()
...
 
+
+
+    /**
+     * Helper method to determine if the current test should enable a test database
+     *
+     * @param $fixtureFiles
+     * @return bool
+     */
+    protected function setupDatabaseForCurrentTest($fixtureFiles): bool
```
tractorcow: PHP 5.6 compatible syntax please. :)

src/Dev/SapphireTest.php
```diff
@@ -280,6 +285,50 @@ protected function setUp()
...
+     * Helper method to check, if the current test uses the database.
+     * This can be switched on with the annotation "@useDatabase false"
+     *
+     * @return bool
+     */
+    protected function currentTestDisablesDatabase()
+    {
+        $annotations = $this->getAnnotations();
+
+        return array_key_exists('useDatabase', $annotations['method'])
```
tractorcow: This method is exactly the same as above, with an inverse check. I'd rather a single method that returned either a boolean, or `null` if not declared. These could even be combined with the prior method, as these effectively resolve a single value.

src/Dev/SapphireTest.php
tractorcow: `setup` replies a verb rather than an answer to a question. Perhaps `isDatabaseRequired()`?

src/Dev/SapphireTest.php
```diff
@@ -594,91 +643,101 @@ public function assertEmailSent($to, $from = null, $subject = null, $content = n
...
-     *         array('Email' => '...@example.com'),
-     *         array('Email' => 'i...@example.com'),
-     *      ), $members);
+     *      $this->assertListContains([
+     *         ['Email' => '...@example.com'],
+     *         ['Email' => 'i...@example.com'],
+     *      ], $members);
      */
-    public function assertDOSContains($matches, $dataObjectSet)
+    public static function assertListContains($matches, SS_List $list, $message = '')
```
tractorcow: Please add `$message` to phpdoc.

src/Dev/SapphireTest.php
```diff
@@ -594,91 +643,101 @@ public function assertEmailSent($to, $from = null, $subject = null, $content = n
...
-     *         array('Email' => '...@example.com'),
-     *         array('Email' => 'i...@example.com'),
-     *      ), $members);
+     *      $this->assertListNotContains([
+     *          ['Email' => '...@example.com'],
+     *          ['Email' => 'i...@example.com'],
+     *      ], $members);
      */
-    public function assertNotDOSContains($matches, $dataObjectSet)
+    public static function assertListNotContains($matches, SS_List $list, $message = '')
```
tractorcow: Please add '$message` to phpdoc.

src/Dev/SapphireTest.php
```diff
@@ -689,85 +748,87 @@ public function assertNotDOSContains($matches, $dataObjectSet)
...
+     *        ['FirstName' => 'Ingo', 'Surname' => 'Schommer'],
+     *      ], $members);
      *
      * @param mixed $matches The patterns to match.  Each pattern is a map of key-value pairs.  You can
      * either pass a single pattern or an array of patterns.
-     * @param mixed $dataObjectSet The {@link SS_List} to test.
+     * @param mixed $list The {@link SS_List} to test.
      */
-    public function assertDOSEquals($matches, $dataObjectSet)
+    public static function assertListEquals($matches, SS_List $list, $message = '')
```
tractorcow: Please add `$message` to phpdoc.

tractorcow: Should we have an assertListNotEquals() to match all the pairs of tests?

src/Dev/Constraint/SSListContainsOnlyMatchingItems.php
```diff
@@ -0,0 +1,76 @@
...
+     * a boolean value instead: true in case of success, false in case of a
+     * failure.
+     *
+     * @param mixed $other Value or object to evaluate.
+     * @param string $description Additional information about the test
+     * @param bool $returnResult Whether to return a result or throw an exception
+     *
+     * @return mixed
+     *
+     * @throws PHPUnit_Framework_ExpectationFailedException
```
tractorcow: For phpdoc types please also add the `use` declarations.

src/Dev/Constraint/SSListContainsOnly.php
```diff
@@ -0,0 +1,72 @@
...
+            $this->has_leftover_items = true;
+        }
+
+        if ($returnResult) {
+            return $success;
+        }
+
+        if (!$success) {
+            $this->fail($other, $description);
+        }
```
tractorcow: For methods which return a value can we please add a `return null;` at the end to demonstrate the explicit behaviour? I also prefer `@return null|bool` to `@return mixed` to assist with users reading this code. :)

src/Dev/Constraint/SSListContainsOnly.php
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    protected function getStubForToString()
+    {
+        $this->item_not_matching
```
tractorcow: Is there a missing `return` here?

src/Dev/Constraint/SSListContainsOnly.php
wernerkrauss: ok, done. I also ran PHPStorms inspection and specified the docblocs where possible.

src/Dev/SapphireTest.php
wernerkrauss: Somehow I don't like it to return `null` for "do nothing", as it might be harder to check above in the setup method where this helpers are used to create cleaner code.

is this only used on setup or do we need it also on tear down? We could of course combine those three methods if it's only used for setup (and tear down checks differently if a DB was used)

tractorcow: Fair enough.

src/Dev/Constraint/SSListContainsOnly.php
```diff
@@ -0,0 +1,79 @@
...
+
+if (!class_exists(PHPUnit_Framework_Constraint::class)) {
+    return;
+}
+
+/**
+ * Constraint for checking if a SS_List contains only items matching the given
+ * key-value pairs.  Each match must correspond to 1 distinct record.
+ */
+class SSListContainsOnly extends SSListContains
```
tractorcow: We need `implements TestOnly` on this too.

tractorcow:
I've rebased this at https://github.com/silverstripe/silverstripe-framework/pull/7450; I'll merge it once it's green.

src/Dev/Constraint/SSListContainsOnly.php
tractorcow: I've fixed this in my rebase.

tractorcow:
Merged with https://github.com/silverstripe/silverstripe-framework/pull/7450

PR #7422, opened 2017-10-03
Add gridfield versioned columns

phalkunz:
Fixes https://github.com/silverstripe/silverstripe-versioned/issues/22

### DIFF
--- a/src/Forms/GridField/GridFieldVersionedState.php
+++ b/src/Forms/GridField/GridFieldVersionedState.php
@@ -1,8 +1,6 @@
 <?php
 namespace SilverStripe\Forms\GridField;
 
-use SilverStripe\Forms\GridField\GridField;
-use SilverStripe\Forms\GridField\GridField_ColumnProvider;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\Versioned\Versioned;
 use SilverStripe\Core\Convert;
@@ -17,8 +15,9 @@ class GridFieldVersionedState implements GridField_ColumnProvider
      */
     protected $versionedLabelFields = ['Title'];
 
-    public function __construct($versionedLabelFields = null) {
-        if($versionedLabelFields) {
+    public function __construct($versionedLabelFields = null)
+    {
+        if ($versionedLabelFields) {
             $this->versionedLabelFields = $versionedLabelFields;
         }
     }
@@ -37,7 +36,7 @@ class GridFieldVersionedState implements GridField_ColumnProvider
         $model = $gridField->getModelClass();
         $isModelVersioned = $model::has_extension(Versioned::class);
 
-        if(!$isModelVersioned) {
+        if (!$isModelVersioned) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


src/Forms/GridField/GridFieldVersionedState.php
```diff
@@ -0,0 +1,111 @@
...
+     * @param  GridField $gridField
+     * @param  DataObject $record - Record displayed in this row
+     * @param  string $columnName
+     * @return string - HTML for the column. Return NULL to skip.
+     */
+    public function getColumnContent($gridField, $record, $columnName)
+    {
+
+        $flagContent = '';
+        $flags = $record->getStatusFlags();
```
flamerohr: From memory `getStatusFlags()` is specific to `SiteTree` and other Versioned `DataObject` do not have this method.
The `getStatusFlags()` defined in the `Versioned` extension doesn't return the same data signature either, so it's quite confusing to follow.
SiteTree returns:
```
$status = [
  'draft' => ['title' => 'Bob', 'text' => 'Bob Text'],
];
```
Versioned returns:
```
$status = ['draft'];
```

I suggest building a `getStatusFlag()` protected method in this class and standardise the data which this `getColumnContent()` method could use rather than jump around so many variables which `SiteTree` introduces...
Focus on only showing `draft` and `modified` statuses, and maybe have `archived`/`removedfromdraft` as a nice-to-have.
I should be able to replace `SiteTree` with another Versioned `DataObject` like `File` or (custom) `MyVersionedProduct`. This is important for contentblocks/elemental, for example.

src/Forms/GridField/GridFieldVersionedState.php
```diff
@@ -0,0 +1,111 @@
...
+use SilverStripe\Forms\GridField\GridField_ColumnProvider;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\Versioned\Versioned;
+use SilverStripe\Core\Convert;
+
+class GridFieldVersionedState implements GridField_ColumnProvider
+{
+    protected $column = null;
+
+    protected $versionedLabelFields = ['Name', 'Title'];
```
flamerohr: these could be set in a constructor and updated with setter/getters so it's more versatile for changes

src/Forms/GridField/GridFieldVersionedState.php
```diff
@@ -11,7 +11,11 @@ class GridFieldVersionedState implements GridField_ColumnProvider
 {
     protected $column = null;
 
-    protected $versionedLabelFields = ['Name', 'Title'];
+    protected $versionedLabelFields = [];
+
+    public function __construct($versionedLabelFields = []) {
```
flamerohr: lets have `[ 'Title' ]` as the default, as all `DataObject` have that field defined.

phalkunz: 👍

tractorcow:
This change added a hard dependency on framework with versioned... not ideal!

flamerohr:
true, didn't realise it at the time

PR #7424, opened 2017-10-03
NEW Make Member::changePassword extensible

robbieaverill:
PR allows extensions to perform logic after core when `Member::changePassword` is called. Extensions are provided the member instance, the new password and the current validation result.

If an extension returns a ValidationResult it will be merged and returned with the original.

Use case: the LDAP module's reset password functionality doesn't need to completely override the `doChangePassword` functionality, but it also can't really just call `parent::doChangePassword` then handle the return since it's always a HTTPResponse, regardless of the result of the operation. This change would mean that it can simply add `onAfterChangePassword(...)` to LDAPMemberExtension to sync the new password to an AD server instead.

Another option might be to refactor `doChangePassword::doChangePassword` into smaller (more overloadable) methods. I'm open to suggestions here!

### DIFF
--- a/src/Security/Member.php
+++ b/src/Security/Member.php
@@ -1680,8 +1680,10 @@ class Member extends DataObject
     }
 
     /**
-     * Change password. This will cause rehashing according to
-     * the `PasswordEncryption` property.
+     * Change password. This will cause rehashing according to the `PasswordEncryption` property. This method will
+     * allow extensions to perform actions and augment the validation result if required before the password is written
+     * and can check it after the write also. Note that the onAfterChangePassword extension point receives a clone of
+     * the validation result which cannot be modified.
      *
      * @param string $password Cleartext password
      * @return ValidationResult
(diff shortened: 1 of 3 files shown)


src/Security/Member.php
```diff
@@ -1696,6 +1696,13 @@ public function changePassword($password)
             $this->write();
         }
 
+        $resultFromExtensions = $this->extend('onAfterChangePassword', $password, $valid);
```
tractorcow: If you modify the result from valid to invalid here, you have still written the member to disk in a supposedly invalid state.

I suggest 'onBeforeChangePassword' (takes $valid result and can modify it before $valid->isValid() is called), and an `onAfterChangePassword` which can take the result but not modify it.

Your extensions should instead be responsible for doing combineAnd() not this code. This seems like baked-in logic for your specific use case.

src/Security/Member.php
```diff
@@ -1691,11 +1693,15 @@ public function changePassword($password)
...
         $valid = $this->validate();
 
+        $this->extend('onBeforeChangePassword', $password, $valid);
+
         if ($valid->isValid()) {
             $this->AutoLoginHash = null;
             $this->write();
         }
 
+        $this->extend('onAfterChangePassword', $password, clone $valid);
```
tractorcow: It's a nice idea but unnecessary to clone.

robbieaverill: Was following your comment "and an onAfterChangePassword which can take the result but not modify it" - will remove

tractorcow:
Just needs a rebase to get tests passing.

PR #7426, opened 2017-10-03
NEW: Implement accept attribute in FileField (closes #7279)

### LINKED ISSUE #7279: FileField should make use of accept attribute
This helps users by limiting the types of files they can select in the file dialog (though most browsers still allow a user to switch to “All files” and select anything). For example, `<input type="file" accept="jpeg,jpg,image/jpeg" name="File" />`. [Recommended approach](https://www.w3.org/TR/html5/forms.html#attr-input-accept) is to specify _both_ file extension and mime type. Psuedo-code from a project I’ve added this to:

```php
/**
 * {@inheritdoc}
 */
public function getAttributes()
{
    $attributes = parent::getAttributes();

    $mimeTypes = Config::inst()->get('HTTP', 'MimeTypes');
    $extensions = $this->getValidator()->getAllowedExtensions();
    $accept = [];
    foreach ($extensions as $extension) {
        $accept[] = $extension;
        if (isset($mimeTypes[$extension])) {
            $accept[] = $mimeTypes[$extension];
        }
    }

    if (!empty($accept)) {
        $attributes = array_merge($attributes, ['accept' => implode(',', array_unique($accept))]);
    }

    return $attributes;
}
```

kinglozzer:


src/Forms/FileField.php
```diff
@@ -85,6 +86,43 @@ public function Field($properties = array())
...
+        if (!$extensions) {
+            return [];
+        }
+
+        $accept = [];
+        $mimeTypes = HTTP::config()->uninherited('MimeTypes');
+        foreach ($extensions as $extension) {
+            $accept[] = ".{$extension}";
+            // Check for corresponding mime type
+            if (isset($mimeTypes[$extension])) {
```
tractorcow: Multiple extensions could share the same mime type; You should ensure `$accept` is unique before returning. Maybe associative array?

kinglozzer: Just below: `return array_unique($accept);` 😉

dhensby:
next time push to your fork ;)

kinglozzer:
Ugh, I ran a `composer update` that moved framework into `vendor/`, which must’ve re-created the git directory with the wrong remotes. Good job I didn’t push to any of the main branches 😛

src/Forms/FileField.php
tractorcow: Ah sorry. :D

PR #7428, opened 2017-10-04
MINOR Separate some areas of logic in LostPasswordHandler to make them more overridable

robbieaverill:
This PR does this:

* Don't redirect from `/Security/changepassword` back to `/Security/changepassword` when resetting your password if it's set as the "back URL"
* Separate some logic out of `LostPasswordHandler:: forgotPassword` into their own methods to allow child classes to modify behaviour more easily without overloading `forgotPassword` entirely
* Use the configurable `unique_identifier_field` property for the input form data as well as the lookup field on Member

### DIFF
--- a/src/Security/MemberAuthenticator/LostPasswordHandler.php
+++ b/src/Security/MemberAuthenticator/LostPasswordHandler.php
@@ -165,23 +165,14 @@ class LostPasswordHandler extends RequestHandler
      */
     public function forgotPassword($data, $form)
     {
-        // Ensure password is given
-        if (empty($data['Email'])) {
-            $form->sessionMessage(
-                _t(
-                    'SilverStripe\\Security\\Member.ENTEREMAIL',
-                    'Please enter an email address to get a password reset link.'
-                ),
-                'bad'
-            );
-
-            return $this->redirectToLostPassword();
+        // Run a first pass validation check on the data
+        $dataValidation = $this->validateForgotPasswordData($data, $form);
+        if ($dataValidation instanceof HTTPResponse) {
+            return $dataValidation;
         }
 
-        // Find existing member
-        $field = Member::config()->get('unique_identifier_field');
         /** @var Member $member */
-        $member = Member::get()->filter([$field => $data['Email']])->first();
+        $member = $this->getMemberFromData($data);
 
         // Allow vetoing forgot password requests
         $results = $this->extend('forgotPassword', $member);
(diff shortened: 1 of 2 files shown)


src/Security/MemberAuthenticator/LostPasswordHandler.php
```diff
@@ -197,16 +189,55 @@ public function forgotPassword($data, $form)
...
+     * implementation, but child classes can override this method to use another unique identifier field
+     * for validation.
+     *
+     * @param  array $data
+     * @param  LostPasswordForm $form
+     * @return HTTPResponse|null
+     */
+    protected function validateForgotPasswordData(array $data, LostPasswordForm $form)
+    {
+        if (empty($data['Email'])) {
```
tractorcow: We have hard-coded Email here but soft-coded it to unique_identifier_field elsewhere.

tractorcow: Also, LostPasswordForm.php has it hard-coded to `Email` as well.

tractorcow: If that's the case should you remove the soft-coding on the other part of the class and fix to `Email`?

src/Security/MemberAuthenticator/LostPasswordHandler.php
```diff
@@ -197,16 +189,55 @@ public function forgotPassword($data, $form)
...
+     * Load an existing Member from the provided data
+     *
+     * @param  array $data
+     * @param  string $uniqueIdentifier
+     * @return Member|null
+     */
+    protected function getMemberFromData(array $data, $uniqueIdentifier)
+    {
+        if (!empty($data[$uniqueIdentifier])) {
+            return Member::get()->filter([$uniqueIdentifier => $data[$uniqueIdentifier]])->first();
```
tractorcow: This is the $data[] array key access I was referring to.

src/Security/MemberAuthenticator/LostPasswordHandler.php
robbieaverill: Yep, sure. I think it's a confusion of trying to be abstract and concrete at the same time. I'll change `getMemberFromData` back to the original logic it had.

src/Security/MemberAuthenticator/LostPasswordHandler.php
robbieaverill: Ok I've switch this back to `$data['Email']` as it was originally

tractorcow:
Can be merged on green.

src/Security/MemberAuthenticator/LostPasswordHandler.php
```diff
@@ -197,16 +189,55 @@ public function forgotPassword($data, $form)
 
         // Avoid information disclosure by displaying the same status,
         // regardless wether the email address actually exists
+        $identifier = isset($data[$uniqueIdentifier]) ? $data[$uniqueIdentifier] : '';
```
tractorcow: Sorry I missed this one too. Also needs to be hard-coded to email. :)

robbieaverill: Ok, that is another part that would ideally be abstract. I'll push another commit that makes the use of `$data['Email']` more abstract, and see what you think

robbieaverill: @tractorcow let me know what you think of f3644484f8245c5089c04e799e0e000add466dc0

src/Security/MemberAuthenticator/LostPasswordHandler.php
```diff
@@ -227,4 +249,21 @@ protected function sendEmail($member, $token)
...
+    /**
+     * Avoid information disclosure by displaying the same status, regardless wether the email address actually exists
+     *
+     * @param array $data
+     * @return HTTPResponse
+     */
+    protected function redirectToSuccess(array $data)
+    {
+        $link = Controller::join_links(
+            $this->link('passwordsent'),
```
Firesphere: Should passwordsent be a configuration?

robbieaverill: The difficulty is that this class is a mixture of configuration and concrete implementation. By putting it into its own method we can overload it easily in children, which is what I opted for in the end

Firesphere: Hmmm, maybe both would be a better option? @tractorcow what do you think?

tractorcow: Seems ok as it is. It's hard-coded in the same class that implement that exact url handler, so it's not coupled incorrectly.

Perhaps if linking to Security/passwordsent then maybe. :)

src/Security/MemberAuthenticator/LostPasswordHandler.php
```diff
@@ -195,15 +187,45 @@ public function forgotPassword($data, $form)
...
+    }
+
+    /**
+     * Load an existing Member from the provided data
+     *
+     * @param  array $data
+     * @param  string $uniqueIdentifier
+     * @return Member|null
+     */
+    protected function getMemberFromData(array $data, $uniqueIdentifier)
```
tractorcow: I just realised this second arg isn't even needed. Just pull it in directly from config.

```
$uniqueIdentifier = Member::config()->get('unique_identifier_field');
```

robbieaverill: Cool, updated and rebased

Issue #7405, opened 2017-09-26
Convert core modules to silverstripe-vendormodule


chillu:
# Overview

We've allowed the placement of SilverStripe modules in the `vendor/` folder through https://github.com/silverstripe/silverstripe-framework/issues/7360. But currently, core modules don't declare themselves as vendormodules, because we haven't tested the impact of this change across all modules. This undermines our recommendation of vendormodules as the new best practice, and will hinder adoption.

# Acceptance Criteria

 * All SilverStripe core modules (packaged in `recipe-cms` and `recipe-core`) can be installed in the webroot

# Notes

 * Assumes this change is not an API breakage

# Pull requests

 * [x] https://github.com/silverstripe/webpack-config/pull/11
 * [x] https://github.com/silverstripe/recipe-core/pull/4
 * [x] https://github.com/silverstripe/silverstripe-campaign-admin/pull/66
 * [x] https://github.com/silverstripe/silverstripe-assets/pull/73
 * [x] https://github.com/silverstripe/silverstripe-graphql/pull/121
 * [x] https://github.com/silverstripe/silverstripe-admin/pull/280
 * [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/649
 * [x] https://github.com/silverstripe/silverstripe-versioned/pull/48
 * [x] https://github.com/silverstripe/silverstripe-reports/pull/84
 * [x] https://github.com/silverstripe/silverstripe-errorpage/pull/10
 * [x] https://github.com/silverstripe/silverstripe-cms/pull/1980
 * [x] https://github.com/silverstripe/silverstripe-siteconfig/pull/77
 * [x] https://github.com/silverstripe/silverstripe-postgresql/pull/76
 * [x] https://github.com/silverstripe/silverstripe-sqlite3/pull/32
 * [x] https://github.com/silverstripe/silverstripe-framework/pull/7418
 * [x] https://github.com/silverstripe/silverstripe-installer/pull/200
 * [x] https://github.com/silverstripe/silverstripe-assets/pull/78
 * [x] https://github.com/silverstripe/silverstripe-cms/pull/1982
 * [x] https://github.com/silverstripe/silverstripe-versioned/pull/50
 * [x] https://github.com/silverstripe/silverstripe-admin/pull/283
 * [x] https://github.com/silverstripe/silverstripe-admin/pull/285
 * [x] https://github.com/silverstripe/silverstripe-framework/pull/7427
[...]

PR #7429, opened 2017-10-04
Fix sake path, improve .htaccess upgrading instructions

chillu:
Follow up work from https://github.com/silverstripe/silverstripe-framework/issues/7405


dhensby:
Should we not add `sake` to the as a `bin` in `composer.json` so we can just reference `vendor/bin/sake` ?

tractorcow:
I'd also love to do that. :)

dhensby:
done in #7645

Issue #7430, opened 2017-10-04
Use index.php for routing instead of framework/main.php


chillu:
# Overview

Follow up after [moving modules into vendor folder](https://github.com/silverstripe/silverstripe-framework/issues/7405), and a precondition to [run webroot in public subfolder](https://github.com/silverstripe/silverstripe-framework/issues/7419). 

We've previously relied on a custom `url` query parameter for routing, which is clunky at best: It means you can't use this as a query parameter in your own payloads, and has security implications around route spoofing. Webservers seem to consistently set `REQUEST_URI` now, or at least there's various ways to get to this information without relying on mod_rewrite (see symfony's [Request.php](https://github.com/symfony/http-foundation/blob/master/Request.php#L1676) for the amount of variations - we should probably support more of them...)

# Acceptance Criteria

 * Serve SilverStripe with only a minimal index.php file in the webroot
 * Support operation on various webservers (incl. ones without rewrite rule support)
 * Upgrading is documented, and as automated as possible
 * Server installation docs are upgraded

# Tasks

 * [x] Fix framework tests
 * [x] Test installer in webroot and subfolder of webroot
 * [x] Fix `Installer.php` references to files
 * [x] Update `index.php` in installer
 * [x] Test and fix IIS operation
 * [x] Test installer and routing without rewrite support (`index.php` based routing)
 * [x] Test protected assets operation
 * [x] Server installation docs
 * [x] Upgrading docs

# Notes

 * Need to adjust staticpublisher module to new routing

# Pull Requests

 * [x] https://github.com/silverstripe/silverstripe-installer/pull/202
 * [x] https://github.com/silverstripe/silverstripe-framework/pull/7432
 * [x] https://github.com/silverstripe/silverstripe-assets/pull/79
 * [x] https://github.com/silverstripe/silverstripe-serve/pull/16
 * [x] https://github.com/silverstripe/recipe-core/pull/6

# Test

```
git clone -b pulls/4/request-uri https://github.com/open-sausages/silverstripe-installer.git 4.clean
cd 4.clean
composer config repositories.framework vcs https://github.com/open-sausages/silverstripe-framework.git
composer config repositories.assets vcs https://github.com/open-sausages/silverstripe-assets.git
composer config repositories.recipe-core vcs https://github.com/open-sausages/recipe-core.git
[...]

Issue #7419, opened 2017-10-02
Run webroot in public subfolder


chillu:
# Overview

Continued from work around [moving modules into the vendor folder](https://github.com/silverstripe/silverstripe-framework/issues/7405). Requires [use index.php for routing instead of framework/main.php](https://github.com/silverstripe/silverstripe-framework/issues/7430)

# Acceptance Criteria

 * Support operation on various webservers (incl. ones without rewrite rule support)
 * Support running through `php -S`
 * The `resources/` folder can be placed in the `public/` folder
 * The installer creates this project structure by default (to promote best practice for new projects)
 * The `recipe-core` module creates the `public/` folder
 * Upgrading to 4.1 does not change your project structure, you have to opt-in
 *  Opting out on newly created projects is possible (may require manual effort) but the default is to use public folder.
 * Folder is fixed to public folder. If the folder exists it will be used, and if it doesn't legacy non-public project is assumed.
 * Works with inlined themes (not composer dependencies) but these need to be manually installed to public folder if not using composer
 * Works with themes / non-vendor modules as composer dependencies (see silverstripe/vendor-plugin#4) . May require per-modules changes to support 4.1, but should be trivial to upgrade a module for 4.1 support.
 * Documentation for adding / removing a public folder from a website is available.
 * Theme usage docs are upgraded
 * Server installation docs are upgraded
 * There is a mechanism for treating `public` committed files as a themed resource.

# Tasks

 * [x] Update recipe-plugin to conditionally move public files into public / root depending on project configuration.
 * [x] Ensure that new projects with all recipes (recipe-core, recipe-cms, installer) create `public` folder by default.
 * [x] Shift all public resources for all recipes (including recipe-core/index.php) into the `public` folder on each repo).
 * [x] Update vendor-plugin Auto-move `resources/` into `public/resources` (detect `public` folder and use if present).
 * [x] Update vendor-plugin to expose resources for `silverstripe-module` and `silverstripe-theme` types as well.
 * [x] Update vendor-plugin to expose project resources not controlled by composer dependencies. E.g. to `expose` a custom-committed theme folder.
 * [x] Update module resource locator to work with files in public (e.g. public folder themes).
 * [x] Fix framework tests
 * [x] Test installer
 * [x] Test and fix IIS operation
 * [x] Adjust default gitignore paths
 * [x] Server installation docs
 * [x] Upgrading docs
 * [x] Check that #1384 has been addressed
[...]

PR #7432, opened 2017-10-04
Use index.php for serving content

chillu:
See https://github.com/silverstripe/silverstripe-framework/issues/7419 and https://github.com/silverstripe/silverstripe-framework/issues/7430


chillu:
Behat tests are failing until https://github.com/silverstripe/silverstripe-serve/pull/16 is merged. That change is backwards compatible with SS3 and the current SS4 (using `framework/main.php`), so can be merged prior to the other changes.

src/Control/HTTPRequestBuilder.php
```diff
@@ -17,10 +17,13 @@ public static function createFromEnvironment()
     {
         // Clean and update live global variables
         $variables = static::cleanEnvironment(Environment::getVariables());
-        Environment::setVariables($variables); // Currently necessary for SSViewer, etc to work
 
         // Health-check prior to creating environment
-        return static::createFromVariables($variables, @file_get_contents('php://input'));
+        $req = static::createFromVariables($variables, @file_get_contents('php://input'));
+
+        Environment::setVariables($variables); // Currently necessary for SSViewer, etc to work
```
tractorcow: Not sure why this order had to be switched. createFromVariables() should not be modifying $variables...

tractorcow:
This bit is incorrect;

```

**Comment by tractorcow** [comment]
// Remove query parameters (they're retained separately through $server['_GET']
$url = parse_url($variables['_SERVER']['REQUEST_URI'], PHP_URL_PATH);
 // Remove base folders from the URL if webroot is hosted in a subfolder
if (substr(strtolower($url), 0, strlen(BASE_URL)) === strtolower(BASE_URL)) {
    $url = substr($url, strlen(BASE_URL));
}
 // Normalise URI
$variables['_SERVER']['REQUEST_URI'] = $url;
```

chillu:
> $_SERVER['REQUEST_URI'] needs to be root-relative; E.g. /basedir/page not /page; Otherwise it breaks sites running in sub-folders.

Good catch, looks like you've fixed that already. I've checked operation in a subfolder, works fine (as long as you don't define `SS_BASE_URL` - which the installer happens to do, and that skewed my testing and sent me on a wild goose chase around core grrrr!)

tractorcow:
Yes I've done that before. :)

PR #7436, opened 2017-10-05
NEW Ensure changePassword is called by onBeforeWrite for a consistent API

### LINKED ISSUE #7435: Changing a member's password should use a consistent API
At the moment you can change a member's password in the following ways:

* Programmatically
* Via your profile in the CMS
* Directly on a member in the CMS (same logic as the above)
* Via "lost/change password" forms

The `Member::changePassword` API method should be used consistently to achieve this. Currently it's used via the lost/change password forms, and now (see #7424) provides some extra extension points for auxiliary modules to do things when this happens, but other parts of the framework rely on `Member::onBeforeWrite` to do password change logic (as does `::changePassword` to be fair, via `::write()`)

It would be great if the logic in `onBeforeWrite` could be adjusted to use `changePassword`, or moved to it altogether, so that changing a Member's password from any means would use the same logic.

---

I think if we were to add a `$write = true` argument to `changePassword` we could do this without too much of a refactor, then perhaps the majority of this logic could be moved to `changePassword` instead: https://github.com/silverstripe/silverstripe-framework/blob/6b52412693eaf13d0ae60bb807180069d9cb7eec/src/Security/Member.php#L905-L930. That would mean that `onBeforeWrite` would call `changePassword` if the password has been changed or added fresh, and would achieve some refactoring of the `onBeforeWrite` method at the same time.

robbieaverill:
This ensures that however you change a member's password, `Member::changePassword` is always called.

Resolves #7435 

**Update:** I've removed the `Member::SetPassword` property. It's unused from what I can tell. Was introduced in SS 2.2.2 and hasn't been touched since then.

### DIFF
--- a/docs/en/04_Changelogs/4.0.0.md
+++ b/docs/en/04_Changelogs/4.0.0.md
@@ -2507,6 +2507,7 @@ New `TimeField` methods replace `getConfig()` / `setConfig()`
 
  * `LoginForm` now has an abstract method `getAuthenticatorName()`. If you have made subclasses of this,
    you will need to define this method and return a short name describing the login method.
- * `MemberLoginForm` has a new constructor argument for the authenticator class, athough tis is usually
+ * `MemberLoginForm` has a new constructor argument for the authenticator class, although this is usually
    constructed by `MemberAuthenticator` and won't affect normal use.
- * `Authenticator` methods `register` and `unregister` are deprecated in favor of using `Config`
+ * `Authenticator` methods `register` and `unregister` are deprecated in favour of using `Config`
+ * Unused `SetPassword` property removed from `Member`. Use `Member::changePassword` or set `Password` directly.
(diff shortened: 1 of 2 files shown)


chillu:
Hmm, I would prefer if `changePassword()` was just not calling `write()` in the first place, but that'd be an API change. So the `passwordChangesToWrite` flag seems like the least obtrusive way to solve this.

kinglozzer:
@dhensby perhaps we just need to add [this](https://github.com/silverstripe/silverstripe-framework/blob/4/src/Forms/ConfirmedPasswordField.php#L330-L333) to `PasswordField`?

robbieaverill:
That sounds to me like a potential enhancement for the PasswordField API - thoughts?

dhensby:
> @dhensby perhaps we just need to add this to PasswordField?

I don't think we should ever be sending the password value *back* to the browser really... If that change has resolved the problem that `SetPassword` set out to achieve (which may well be more than just the case I've highlighted - for example it's more explicit that a change password event has happened), then maybe it's not a big deal...

Issue #7382, opened 2017-09-18
Performance recommendation: Singleton extensions


tractorcow:
Parent issue: https://github.com/silverstripe/silverstripe-admin/issues/199

At the moment, every time we construct a class that has extensions, we construct a number of extension instances for each of these instances (effectively o(n^2)). This is extremely expensive, given that most extensions provide behaviour only to the owned object and do not themselves contain instance-specific properties (although some do!).

As we currently rely on setOwner() before calling methods on extensions, it should be feasible to safely share instances across multiple parents (although the risk does increase if setOwners aren't updated down the call chain).

Extensible::getInstances() will be modified with the following change:

```diff
-$instance = Injector::inst()->create($extension);
+$instance = Injector::inst()->get($extension);
```

Extensions which must NOT be shared can still be declared as `type: prototype` via injector, which will cause get() to return a safely isolated new instance.

Referenc: https://blackfire.io/profiles/dea4e266-e469-4976-abf9-c7844503b04b/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=focused&settings%5BtabPane%5D=nodes&selected=SilverStripe%5CView%5CViewableData%3A%3AgetExtensionInstances&callname=SilverStripe%5CView%5CViewableData%3A%3AgetExtensionInstances

Note that getExtensionInstances() is called about 1100 on a normal /admin/pages url. Of these 1100 requests there are about 100 Injector::create() calls.

# PRs

 - [x] https://github.com/silverstripe/silverstripe-versioned/pull/52
 - [x] https://github.com/silverstripe/silverstripe-framework/pull/7437

tractorcow:
Effort set to medium for implementation: easy, testing: hard

flamerohr:
To clarify, you'll need to ensure that `setOwner()` is called everywhere before the extension class is used, this seems like a similar thing that is being done with the `Controller` class and `setRequest()`.
and the `type: prototype` that you mention will be a config on the Extension class right?

chillu:
Seems borderline as a late stage change, e.g. modules that have already upgraded will need to go back and check their extensions. But we allow opting out of this behaviour, so at least you don't need to rewrite custom code. It's a bit counter-intuitive to show devs `$this->owner` references in example code, but then have the caveat that other `$this` references won't work. We should just pass in `$owner` as a first arg to all extension methods to make these stateless, but that's a 5.x change. My assumption is there's non-trivial performance benefits which drive this change?

[...]

PR #7437, opened 2017-10-05
API Extensions are now stateless

tractorcow:
Parent story: https://github.com/silverstripe/silverstripe-framework/issues/7382

### DIFF
--- a/src/Core/ClassInfo.php
+++ b/src/Core/ClassInfo.php
@@ -197,7 +197,7 @@ class ClassInfo
         // Merge with descendants
         $descendants = ClassLoader::inst()->getManifest()->getDescendantsOf($className);
         return array_merge(
-            [ $lowerClassName => $className ],
+            [$lowerClassName => $className],
             $descendants
         );
     }
@@ -441,6 +441,10 @@ class ClassInfo
             } elseif (is_array($token) && $token[0] === T_NS_SEPARATOR) {
                 $class .= $token[1];
                 $hadNamespace = true;
+            } elseif ($token === '.') {
+                // Treat service name separator as NS separator
+                $class .= '.';
+                $hadNamespace = true;
             } elseif ($hadNamespace && is_array($token) && $token[0] === T_STRING) {
                 $class .= $token[1];
                 $hadNamespace = false;
@@ -454,7 +458,11 @@ class ClassInfo
                                 $result = stripcslashes(substr($argString, 1, -1));
                                 break;
                             case "'":
-                                $result = str_replace(array("\\\\", "\\'"), array("\\", "'"), substr($argString, 1, -1));
+                                $result = str_replace(
+                                    ["\\\\", "\\'"],
+                                    ["\\", "'"],
+                                    substr($argString, 1, -1)
+                                );
                                 break;
                             default:
                                 throw new Exception("Bad T_CONSTANT_ENCAPSED_STRING arg $argString");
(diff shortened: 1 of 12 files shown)


src/ORM/Hierarchy/Hierarchy.php
```diff
@@ -23,21 +23,6 @@
...
-     * @var int
-     */
-    protected $_cache_numChildren = null;
-
-    /**
-     * Cache for {@see Children()}
-     *
-     * @var SS_List
-     */
-    protected $_cache_children = null;
```
kinglozzer: I’ve done some provisional benchmarking of a blank install, replacing this with a static cache in `Hierarchy` (e.g. `if (isset(static::$_cache_children[$this->owner->ID]))`) and it seems to save a few calls to `->stageChildren()` when looking the `treeview` AJAX request. I can’t figure out why (perhaps individual pages are being garbage collected and then re-fetched?), do you want to look at that?

Usual caveat about memory usage may or may not apply... I’m struggling to think of the memory usage implications of prototype extension vs static cache vs cache on owner

src/Security/Group.php
```diff
@@ -618,9 +618,12 @@ public function AllChildrenIncludingDeleted()
...
         $extInstance = $this->getExtensionInstance(Hierarchy::class);
-        $extInstance->setOwner($this);
-        $children = $extInstance->AllChildrenIncludingDeleted();
-        $extInstance->clearOwner();
+        try {
+            $extInstance->setOwner($this);
+            $children = $extInstance->AllChildrenIncludingDeleted();
+        } finally {
+            $extInstance->clearOwner();
+        }
```
kinglozzer: Possibly a stupid question, and I know you didn’t write this, but could this be replaced with `$children = parent::AllChildrenIncludingDeleted()`? Won’t that invoke `CustomMethods::__call()`?

src/Core/ClassInfo.php
```diff
@@ -441,6 +441,10 @@ public static function parse_class_spec($classSpec)
             } elseif (is_array($token) && $token[0] === T_NS_SEPARATOR) {
                 $class .= $token[1];
                 $hadNamespace = true;
+            } elseif ($token === '.') {
+                // Treat service name separator as NS separator
+                $class .= '.';
```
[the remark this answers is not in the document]
tractorcow: Yes, if you have an extension with different constructor args it's necessary to use a service name, since otherwise you'll end up with multiple forms being injected with a single singleton.

src/Core/Extension.php
```diff
@@ -33,12 +33,12 @@
      * The base class that this extension was applied to; $this->owner must be one of these
      *
      * @var DataObject
+     * @deprecated 4.0...5.0 Use $this->owner->baseClass() instead
```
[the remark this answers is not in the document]
tractorcow: Yes I think so.

src/Core/Extension.php
```diff
@@ -63,24 +63,29 @@ public static function add_to_class($class, $extensionClass, $args = null)
...
+     * @param mixed $owner Owner to set
+     * @param callable $callback Callback to invoke
+     * @param array $args Args to pass to callback
+     * @return mixed
+     */
+    public function withOwner($owner, callable $callback, $args = [])
+    {
+        try {
+            $this->setOwner($owner);
+            return call_user_func_array($callback, $args);
```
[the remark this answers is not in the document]
tractorcow: Ah thanks for that it helps a lot. :)

src/Security/Group.php
tractorcow: Yes that shousd I think.

src/ORM/Hierarchy/Hierarchy.php
tractorcow: One problem is that it may be more than a simple ID to key the cache; It could also be stage.

kinglozzer: Good point

PR #7439, opened 2017-10-05
FIX Trim whitespace off names in Injector

### LINKED ISSUE #7421: DBField arguments are susceptible to whitespace
Not sure it's a bug, but juste in case, because I'm converting my SS3 scripts to SS4 and didn't see nothing special on https://docs.silverstripe.org/en/4/changelogs/4.0.0/#overview

In SS3 we got a class Decimal (http://api.silverstripe.org/en/3.5/class-Decimal.html)

In SS4, it failed during instalation. Example:

```php
class MyScript extends DataObject {
	private static $db = array(
		'CR' =>  #'Decimal (5,2)'
)}
```

robbieaverill:
Fixes #7421

### DIFF
--- a/src/Core/Injector/Injector.php
+++ b/src/Core/Injector/Injector.php
@@ -1024,7 +1024,8 @@ class Injector implements ContainerInterface
                 $args = $extraArgs;
             }
         }
-        return [ $name, $args ];
+        $name = trim($name);
+        return [$name, $args];
     }
 
     /**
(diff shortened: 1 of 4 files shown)


tractorcow:
This won't fix `Injector::inst()->get('Varchar (255)')` since the whitespace is actually IN the identifier.

Can you please put the trim into the `Injector::normaliseArguments` method instead, just before the return.

robbieaverill:
D'oh! Updated. I've also removed a couple of unnecessary namespace imports from the injector stubs.

PR #7448, opened 2017-10-08
BUG Fix missing property in requirements minificaction docs

### LINKED ISSUE #7441: SS4: Example implementation of minifier doesn't work
Minification of JavaScript / CSS as per [the docs](https://github.com/silverstripe/silverstripe-framework/blob/4/docs/en/02_Developer_Guides/01_Templates/03_Requirements.md#minification-of-css-and-js-files) for SilverStripe 4 does not appear to work (in the example the `MyMinifier` class does not seem to get called with flushes in live mode).

I am aware that the built-in minifier has been removed from SS4, but this is regarding the documented implementation of a minification process.

# PR

 - [ ] https://github.com/silverstripe/silverstripe-framework/pull/7448

tractorcow:
Fixes #7441


flamerohr:
it is strange that the flag also needs to be enabled for it to work, lol

PR #7449, opened 2017-10-08
Add updateBreadcrumbs hook to GridFieldDetailForm

wilr:
Requires: https://github.com/silverstripe/silverstripe-admin/pull/291

### DIFF
--- a/src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
+++ b/src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
@@ -587,6 +587,7 @@ class GridFieldDetailForm_ItemRequest extends RequestHandler
 
         /** @var ArrayList $items */
         $items = $this->popupController->Breadcrumbs($unlinked);
+
         if ($this->record && $this->record->ID) {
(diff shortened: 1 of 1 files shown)


src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -587,10 +587,11 @@ public function Breadcrumbs($unlinked = false)
 
         /** @var ArrayList $items */
         $items = $this->popupController->Breadcrumbs($unlinked);
+
         if ($this->record && $this->record->ID) {
             $title = ($this->record->Title) ? $this->record->Title : "#{$this->record->ID}";
             $items->push(new ArrayData(array(
-                'Title' => $title,
+                'Title' => DBField::create_field('Varchar', $title),
```
tractorcow: We don't need to explicitly cast anything that's the same as the default_cast.

tractorcow:
There are some PHPCS issues. I'm looking at behat.

wilr:
Linting issues? It's 1 line?

tractorcow:
That'll make it easy to fix. :)

PR #7455, opened 2017-10-09
FIX Dont use var_export for cache key generation

dhensby:
`var_export` doesn't handle circular references so shouldn't be used in this context to create a cache key

move to `serialize` for more resiliant cache key generation


tractorcow:
Behat failure unrelated.

Issue #7457, opened 2017-10-09
Ensure upgrade experience beta 3 -> beta4


tractorcow:
# Acceptance criteria:

- Users can upgrade from beta3 -> beta4 without unnecessary trouble
- Install is not blocked by either missing index.php nor missing install.php

# PRs:

- [x] https://github.com/silverstripe/silverstripe-framework/pull/7460
- [x] https://github.com/silverstripe/silverstripe-cms/pull/1986
- [x] https://github.com/silverstripe/silverstripe-upgrader/pull/43
- [x] https://github.com/silverstripe/silverstripe-installer/pull/203
- [x] https://github.com/silverstripe/recipe-core/pull/7

tractorcow:
Updated and ready for review again.

tractorcow:
@chillu can you please confirm this adequately provides an upgrade path re the index.php changes?

sminnee:
What about upgrading from 3 -> beta4?

tractorcow:
This change would also fix bootstrapping for sites upgrading from ss3.

sminnee:
:+1:

PR #7458, opened 2017-10-09
ENHANCEMENT Improve upgrade experience for beta3 -> beta4 upgrade

tractorcow:
Parent issue https://github.com/silverstripe/silverstripe-framework/issues/7457

### DIFF
--- a/main.php
+++ b/main.php
@@ -1,0 +1,24 @@
+<?php
+
+/**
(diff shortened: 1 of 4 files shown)


tractorcow:
I've decided that force-writing install.php is a risk we want to avoid.

PR #7464, opened 2017-10-10
adding getter for $tempDB

wernerkrauss:
Is used in  https://github.com/silverstripe/silverstripe-sqlite3/pull/34


tractorcow:
See comment on https://github.com/silverstripe/silverstripe-sqlite3/pull/34#pullrequestreview-68446716. The fix for this needs to be in sqlite.

PR #7467, opened 2017-10-10
Adjust FormField::onBeforeRendor so that it can influence both context and passed properties

### LINKED ISSUE #7466: FormField::Field sets properties in a way the extension cannot reset them
The FormField::Field function calls customise with passed properties and assigns results intoa local variable $conext.

$context is used to render the field with themplates.

The extension 'onBeforeRender' is called in the context of '$this' after that

So the 'onBeforeRender' cannot influence how the field has been set by passed properties

I am new to Silver STripe and newish to PHP and I cannot see an easy fix.   am not 100% sure how the 'extend' mechanism functions.

tractorcow:
Fixes #7466

### DIFF
--- a/src/Forms/FormField.php
+++ b/src/Forms/FormField.php
@@ -978,11 +978,11 @@ class FormField extends RequestHandler
     {
         $context = $this;
 
+        $this->extend('onBeforeRender', $context, $properties);
+
         if (count($properties)) {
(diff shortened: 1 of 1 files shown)


flamerohr:
DO NOT MERGE - until beta4 is released

Issue #7333, opened 2017-09-01
Missing resources and other minor issues causing server 500 responses


wilr:
In 4.0 `urlForResource` throws an Exception if a required file is not present. This causes production websites to completely die. Sure the errors exist and should be resolved but I think in the interest of production systems SilverStripe should take into account what is critical vs a warning.. 

Thoughts from everyone? I would prefer in these cases `trigger_error` or a switch for `isLive`.

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/7468

flamerohr:
I wonder what a good workaround would be...
You could replace `SimpleResourceURLGenerator` with a subclass which catches those exceptions and return a placeholder perhaps?
I think sometimes you do want it to throw the exception and stop the request right?

chillu:
So "resource" in the context of a module could be a template, image, stylesheet, JS, right? Some of which are added to the repo, others are dynamically generated (TinyMCE?). Why would those be missing? For images, the frontend rendering could behave gracefully. In all other cases, for all intents and purposes the frontend operation is broken - which seems "exception-throwing-worthy". Or is this a matter of networked filesystem syncs on deployment?

wilr:
In my case the file was missing only on live for whatever reason but the point was more around whether it was worthy enough of being a critical, site destroying, world ending exception or whether something like a warning for the developer would suffice in this case.

chillu:
In SS3, the [Requirements API](https://github.com/silverstripe/silverstripe-framework/blob/3/view/Requirements.php#L1030) handled this silently, without any form of error feedback - so that's what devs will be used to. 

`Module->getResourcePath()` (which uses `ResourceURLGenerator`) is currently mainly used for frontend assets (e.g. in `EmbedResource` and `Requirements`). Technically it could also be used for more critical server operation assets like YAML configs, `.htaccess` templates etc - which would require stricter error handling. 

Overall, I'd be fine with switching this to `trigger_error()`, and leaving it to the API consumer to raise criticality based on use case.

dhensby:
Yes, I'd rather a website responded without a stylesheet included than giving me a 500 instead

dhensby:
The error can just be logged rather than break everything?

PR #7468, opened 2017-10-11
BUG switch to `trigger_error()` when a resource is not found

flamerohr:
Rather than throw an exception.
This will now give a notice, which can be logged through a logger in production, and you can disable showing the notice through the normal php means `error_reporting()`

Fixes https://github.com/silverstripe/silverstripe-framework/issues/7333


unclecheese:
Have added a unit test.

PR #7477, opened 2017-10-12
ENHANCEMENT Allow extensions to intercept incorrect deletes on unpublish

tractorcow:
The issue with the existing behaviour is that some queried tables are NOT a part of the hierarchy, and the subsequent delete could unintentionally delete objects outside of this object.

For instance, fluent uses a localisation table as an inner join. This existing behaviour deletes these locales, but using the ID of the record, not the ID of the localisation (which is a separate ID itself).

With the given extension points extensions are able to intercept these queries and rewrite the delete safely without violating database integrity.

### DIFF
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -1453,8 +1453,11 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
             ->filter('ID', $this->ID)
             ->dataQuery()
             ->query();
-        foreach ($srcQuery->queriedTables() as $table) {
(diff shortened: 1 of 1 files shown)


tractorcow:
For instance, this is a page that has three failover locales. When deleted it performs these deletes:

```
"SiteTree_Live, ID = 3"
"SiteTree_Localised_Live, ID = 3"
"SiteTree_Localised_Live, ID = 3"
"SiteTree_Localised_Live, ID = 3"
```

Those following three deletes should be RecordID = 3 not ID = 3.

flamerohr:
Seems reasonable, and it doesn't change anything if there were no extensions :)

PR #7480, opened 2017-10-13
[FIX] Password reset message

### LINKED ISSUE #7339: Lost password's "password sent" action UX flow updates
Some minor things I've noticed about the change password handling:

### Don't show the confirmation page unless there's something to confirm

If I access `/Security/lostpassword/passwordsent` I get shown:

image

You can also add `/foo` etc to the end to get it to show any username, which is also not necessary or desirable.

image

### Don't include a period inside the username/email address

I assume that the purpose of putting the username/email address in quotes is to highlight it as a username/email address. If this is the case, the period should not be inside the quotes.

I suppose grammatically speaking that `Password reset link sent to 'foo@example.com.'` is correct, but `A reset link has been sent to 'foo@example.com.', provided...` is not.

My suggestion here would be to just remove the period entirely.

Firesphere:
Fixes #7339

### DIFF
--- a/src/Security/MemberAuthenticator/LostPasswordHandler.php
+++ b/src/Security/MemberAuthenticator/LostPasswordHandler.php
@@ -99,7 +99,10 @@ class LostPasswordHandler extends RequestHandler
     public function passwordsent()
     {
         $request = $this->getRequest();
-        $email = Convert::raw2xml(rawurldecode($request->param('EmailAddress')) . '.' . $request->getExtension());
+        $email = Convert::raw2xml(rawurldecode($request->param('EmailAddress')));
+        if ($request->getExtension()) {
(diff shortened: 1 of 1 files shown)


Firesphere:
I don't know how @tractorcow his commits got in here, I branched directly off origin/4

tractorcow:
Use 4.0 not 4 please. :)

Firesphere:
👍 Yep, will do :)

composer.json
```diff
@@ -61,6 +61,10 @@
         "psr/container-implementation": "1.0.0"
     },
     "extra": {
+        "branch-alias": {
```
tractorcow: This commit needs deleting from this PR.

Firesphere: I have no idea how it got in there. Dropped from PR

PR #7484, opened 2017-10-16
BUG Replace phpdotenv with thread-safe replacement

### LINKED ISSUE #7478: Race condition detecting ENV variables
PRs:

 - [ ] https://github.com/silverstripe/silverstripe-framework/pull/7484
 - [ ] https://github.com/silverstripe/silverstripe-assets/pull/87
 - [ ] https://github.com/silverstripe/silverstripe-behat-extension/pull/170
 - [ ] https://github.com/silverstripe/silverstripe-serve/pull/19

Replicate:

1. Install SS4 (framework only or full doesn't matter happens on both)
2. Ensure you have your .env configured correctly and mod_env enabled
3. Navigate to the homepage
4. (Optional) Remove install.php
5. Spam refresh as fast as possible letting the page load each time.

Inevitable Error:

If you've removed install.php
> SilverStripe Framework requires a $databaseConfig defined.

If you havn't removed it

> Redirected to install.php

Affecting: **Everyone**
* Easy to replicate on WIN/Linux
* Yet to replicate on OSX

image

tractorcow:
Fixes #7478 

Not strictly breaking, but any code which uses `getenv` php method will remain vulnerable until upgrading to the new `Environment::getenv` method.

In order to be as unobtrusive as possible the `getenv` and `putenv` methods on `Environment` class are designed as drop-in replacements for the global php methods.

Thanks to @NightJar for beginning this solution (see https://github.com/NightJar/sapphire/commit/e614edec61f396c7e8f29bd5ee3340f56cf0b1a6).

Other PRs for various modules incoming to replace `getenv` usages.

Other prs to merge after this:

- [ ] https://github.com/silverstripe/silverstripe-assets/pull/87
- [ ] https://github.com/silverstripe/silverstripe-behat-extension/pull/170
- [ ] https://github.com/silverstripe/silverstripe-serve/pull/19

References to this issue (notably laravel based)

- https://github.com/vlucas/phpdotenv/issues/163
- https://github.com/vlucas/phpdotenv/issues/76
- https://github.com/laravel/framework/pull/8187
- https://gist.github.com/progmars/1e545d96dd48676a2aa7

### DIFF
--- a/docs/en/00_Getting_Started/03_Environment_Management.md
+++ b/docs/en/00_Getting_Started/03_Environment_Management.md
@@ -7,8 +7,8 @@ server.
 For each of these environments we may require slightly different configurations for our servers. This could be our debug
 level, caching backends, or - of course - sensitive information such as database credentials.
 
-To solve this problem of setting variables per environment we use environment variables with the help of the 
-[PHPDotEnv](https://github.com/vlucas/phpdotenv) library by Vance Lucas.
+To manage environment variables, as well as other server globals, the [api:SilverStripe\Core\Environment] class
+provides a set of APIs and helpers.
 
 ## Security considerations
 
@@ -32,13 +32,18 @@ You can set "real" environment variables using Apache. Please
 
 ## How to access the environment variables
 
-Accessing the environment varaibles is easy and can be done using the `getenv` method or in the `$_ENV` and `$_SERVER`
-super-globals:
+Accessing the environment varaibles should be done via the `Environment::getEnv()` method
 
 ```php
-getenv('SS_DATABASE_CLASS');
-$_ENV['SS_DATABASE_CLASS'];
-$_SERVER['SS_DATABASE_CLASS'];
+use SilverStripe\Core\Environment;
+Environment::getEnv('SS_DATABASE_CLASS');
+```
+
+Individual settings can be assigned via `Environment::setEnv()` or `Environment::putEnv()` methods.
+
+```php
+use SilverStripe\Core\Environment;
+Environment::setEnv('API_KEY', 'AABBCCDDEEFF012345');
 ```
 
 ## Including an extra `.env` file
@@ -46,12 +51,14 @@ $_SERVER['SS_DATABASE_CLASS'];
 Sometimes it may be useful to include an extra `.env` file - on a shared local development environment where all
 database credentials could be the same. To do this, you can add this snippet to your `mysite/_config.php` file:
 
+Note that by default variables cannot be overloaded from this file; Existing values will be preferred
+over values in this file.
+
 ```php
-try {
-    (new \Dotenv\Dotenv('/path/to/env/'))->load();
-} catch (\Dotenv\Exception\InvalidPathException $e) {
-    // no file found
-}
+use SilverStripe\Core\EnvironmentLoader;
+$env = BASE_PATH . '/mysite/.env';
+$loader = new EnvironmentLoader();
+$loader->loadFile($env);
 ```
 
 ## Core environment variables
--- a/src/Core/Environment.php
+++ b/src/Core/Environment.php
@@ -6,6 +6,11 @@ namespace SilverStripe\Core;
  * Consolidates access and modification of PHP global variables and settings.
  * This class should be used sparingly, and only if information cannot be obtained
  * from a current {@link HTTPRequest} object.
+ *
+ * Acts as the primary store for environment variables, including those loaded
+ * from .env files. Applications should use Environment::getEnv() instead of php's
+ * `getenv` in order to include `.env` configuration, as the system's actual
+ * environment variables are treated immutably.
  */
 class Environment
 {
@@ -24,6 +29,13 @@ class Environment
     protected static $timeLimitMax = null;
 
     /**
+     * Local overrides for all $_ENV var protected from cross-process operations
+     *
+     * @var array
+     */
+    protected static $env = [];
+
+    /**
      * Extract env vars prior to modification
      *
      * @return array List of all super globals
@@ -31,7 +43,7 @@ class Environment
     public static function getVariables()
     {
         // Suppress return by-ref
-        return array_merge($GLOBALS, []);
+        return array_merge($GLOBALS, [ 'env' => static::$env ]);
     }
 
     /**
@@ -41,8 +53,14 @@ class Environment
      */
     public static function setVariables(array $vars)
     {
-        foreach ($vars as $key => $value) {
-            $GLOBALS[$key] = $value;
+        foreach ($vars as $varName => $varValue) {
+            if ($varName === 'env') {
+                continue;
+            }
+            $GLOBALS[$varName] = $varValue;
+        }
+        if (array_key_exists('env', $vars)) {
+            static::$env = $vars['env'];
         }
     }
 
@@ -150,5 +168,52 @@ class Environment
     public static function getTimeLimitMax()
     {
         return static::$timeLimitMax;
+    }
+
+    /**
+     * Get value of environment variable
+     *
+     * @param string $name
+     * @return mixed Value of the environment variable, or false if not set
+     */
+    public static function getEnv($name)
+    {
+        switch (true) {
+            case array_key_exists($name, static::$env):
+                return static::$env[$name];
+            case array_key_exists($name, $_ENV):
+                return $_ENV[$name];
+            case array_key_exists($name, $_SERVER):
+                return $_SERVER[$name];
+            default:
+                return getenv($name);
+        }
+    }
+
+    /**
+     * Set environment variable using php.ini syntax.
+     * Acts as a process-isolated version of putenv()
+     * Note: This will be parsed via parse_ini_string() which handles quoted values
+     *
+     * @param string $string Setting to assign in KEY=VALUE or KEY="VALUE" syntax
+     */
+    public static function putEnv($string)
+    {
+        // Parse name-value pairs
+        $envVars = parse_ini_string($string) ?: [];
+        foreach ($envVars as $name => $value) {
+            self::setEnv($name, $value);
+        }
+    }
+
+    /**
+     * Set environment variable via $name / $value pair
+     *
+     * @param string $name
+     * @param string $value
+     */
+    public static function setEnv($name, $value)
+    {
+        static::$env[$name] = $value;
     }
 }
--- a/src/includes/constants.php
+++ b/src/includes/constants.php
@@ -1,7 +1,7 @@
 <?php
 
-use Dotenv\Dotenv;
-use Dotenv\Exception\InvalidPathException;
+use SilverStripe\Core\Environment;
+use SilverStripe\Core\EnvironmentLoader;
 use SilverStripe\Core\TempFolder;
 
 /**
@@ -58,24 +58,23 @@ if (!defined('BASE_PATH')) {
 }
 
 // Allow a first class env var to be set that disables .env file loading
-if (!getenv('SS_IGNORE_DOT_ENV')) {
+if (!Environment::getEnv('SS_IGNORE_DOT_ENV')) {
     call_user_func(function () {
+        $loader = new EnvironmentLoader();
         foreach ([BASE_PATH, dirname(BASE_PATH)] as $path) {
-            try {
-                (new Dotenv($path))->load();
-            } catch (InvalidPathException $e) {
-                // no .env found - no big deal
-                continue;
+            // Stop searching after first `.env` file is loaded
+            $dotEnvFile = $path . DIRECTORY_SEPARATOR . '.env';
+            if ($loader->loadFile($dotEnvFile)) {
+                break;
             }
-            break;
         }
     });
 }
 
 // Validate SS_BASE_URL is absolute
-if (getenv('SS_BASE_URL') && !preg_match('#^(\w+:)?//.*#', getenv('SS_BASE_URL'))) {
+if (Environment::getEnv('SS_BASE_URL') && !preg_match('#^(\w+:)?//.*#', Environment::getEnv('SS_BASE_URL'))) {
     call_user_func(function () {
-        $base = getenv('SS_BASE_URL');
+        $base = Environment::getEnv('SS_BASE_URL');
         user_error(
             "SS_BASE_URL should be an absolute url with protocol "
             . "(http://$base) or without protocol (//$base)",
@@ -83,14 +82,14 @@ if (getenv('SS_BASE_URL') && !preg_match('#^(\w+:)?//.*#', getenv('SS_BASE_URL')
         );
         // Treat as protocol-less absolute url
         $base = '//' . $base;
-        putenv("SS_BASE_URL=$base");
+        Environment::setEnv('SS_BASE_URL', $base);
     });
 }
 
 if (!defined('BASE_URL')) {
     define('BASE_URL', call_user_func(function () {
         // Prefer explicitly provided SS_BASE_URL
-        $base = getenv('SS_BASE_URL');
+        $base = Environment::getEnv('SS_BASE_URL');
         if ($base) {
             // Strip relative path from SS_BASE_URL
             return rtrim(parse_url($base, PHP_URL_PATH), '/');
--- a/tests/php/Core/EnvironmentTest/test.env
+++ b/tests/php/Core/EnvironmentTest/test.env
@@ -1,0 +1,13 @@
+# Test file
+
+TEST_ENV_FIRST="first"
+TEST_ENV_SECOND="start \"#notcomment end"#Comment
+#Comment
+TEST_ENV_THIRD = 3 #comment
+
+TEST_ENV_FOURTH=true #comment
+TEST_ENV_FIFTH = "not#comment"#comment
+TEST_ENV_SIXTH="not#comment"
+
+TEST_ENV_SEVENTH=""   #comment
+
--- a/tests/php/Core/Manifest/ConfigManifestTest.php
+++ b/tests/php/Core/Manifest/ConfigManifestTest.php
@@ -2,9 +2,9 @@
 
 namespace SilverStripe\Core\Tests\Manifest;
 
-use Dotenv\Loader;
 use SilverStripe\Config\Collections\MemoryConfigCollection;
 use SilverStripe\Core\Config\CoreConfigFactory;
+use SilverStripe\Core\Environment;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Kernel;
 use SilverStripe\Core\Manifest\ModuleLoader;
@@ -100,8 +100,7 @@ class ConfigManifestTest extends SapphireTest
 
     public function testEnvVarSetRules()
     {
-        $loader = new Loader(null);
-        $loader->setEnvironmentVariable('ENVVARSET_FOO', 1);
+        Environment::setEnv('ENVVARSET_FOO', '1');
         $config = $this->getConfigFixtureValue('EnvVarSet');
 
         $this->assertEquals(
@@ -137,9 +136,7 @@ class ConfigManifestTest extends SapphireTest
 
     public function testEnvOrConstantMatchesValueRules()
     {
-        $loader = new Loader(null);
-
-        $loader->setEnvironmentVariable('CONSTANTMATCHESVALUE_FOO', 'Foo');
+        Environment::setEnv('CONSTANTMATCHESVALUE_FOO', 'Foo');
         define('CONSTANTMATCHESVALUE_BAR', 'Bar');
         $config = $this->getConfigFixtureValue('EnvOrConstantMatchesValue');
 
@@ -194,9 +191,7 @@ class ConfigManifestTest extends SapphireTest
 
     public function testMultipleRules()
     {
-        $loader = new Loader(null);
-
-        $loader->setEnvironmentVariable('MULTIPLERULES_ENVVARIABLESET', 1);
+        Environment::setEnv('MULTIPLERULES_ENVVARIABLESET', '1');
         define('MULTIPLERULES_DEFINEDCONSTANT', 'defined');
         $config = $this->getConfigFixtureValue('MultipleRules');
 
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
         "symfony/config": "^3.2",
         "symfony/translation": "^2.8",
         "symfony/yaml": "~3.2",
-        "vlucas/phpdotenv": "^2.4",
+        "m1/env": "^2.1",
         "php": ">=5.6.0",
         "ext-intl": "*",
         "ext-mbstring": "*",
--- a/docs/en/04_Changelogs/4.0.0.md
+++ b/docs/en/04_Changelogs/4.0.0.md
@@ -500,6 +500,8 @@ SS_BASE_URL="//localhost/"
 The global values `$database` and `$databaseConfig` have been deprecated, as has `ConfigureFromEnv.php`
 which is no longer necessary.
 
+To access environment variables you can use the `SilverStripe\Core\Environment::getEnv()` method.
+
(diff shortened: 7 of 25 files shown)


dhensby:
I'm not really behind this fix as the whole point in using an external library is so that we aren't maintaining this kind of code ourselves.

from my understanding `putenv` and `getenv` are not threadsafe methods and so this error is only happens on multi-threaded environments (IIS, HHVM and multithreaded apache on windows).

Have we investigated using `Dotenv\Loader::setEnvironmentVariable` and `Dotenv\Loader::getEnvironmentVariable` instead as they rely on the super globals first...

zanderwar:
The external library is not intended for production, so using it in the first place only satisfies the environment when in development albeit is not in `require-dev` and as such is adding unnecessary bloat. 

IMHO the dotenv dependency should be dropped entirely for methodology like this. Even if I am being biased here because this resolves my issue of wanting to use the .env file like laravel/symfony allow flawlessly

tractorcow:
Our external library is `parse_ini_string` :)

src/Core/Environment.php
```diff
@@ -151,4 +166,92 @@ public static function getTimeLimitMax()
...
+                },
+                explode(PHP_EOL, file_get_contents($path))
+            ))
+        );
+
+        // Pass to parse_ini_string
+        $variables = parse_ini_string($contents) ?: [];
+        foreach ($variables as $name => $value) {
+            // Don't overload vars
+            if (static::getenv($name) === false) {
```
flamerohr: could we have overloading vars as a second param for this method?

tractorcow: Yes that's a good idea thank you. :)

src/Core/Environment.php
```diff
@@ -151,4 +166,92 @@ public static function getTimeLimitMax()
...
+    }
+
+    /**
+     * Set environment variable using php.ini syntax.
+     * Acts as a process-isolated version of putenv()
+     * Note: This will be parsed via parse_ini_string() which handles quoted values
+     *
+     * @param string $string Setting to assign in KEY=VALUE or KEY="VALUE" syntax
+     */
+    public static function putenv($string)
```
sminnee: Do we need this to be our API rather than `setEnv($k, $v)`? Who would you recommend calls putenv?

src/Core/Environment.php
```diff
@@ -151,4 +166,92 @@ public static function getTimeLimitMax()
...
         return static::$timeLimitMax;
     }
+
+    /**
+     * Get value of environment variable
+     *
+     * @param string $name
+     * @return mixed Value of the environment variable, or null if not set
+     */
+    public static function getenv($name)
```
sminnee: Can we make this camel case, so getEnv()?

src/Core/Environment.php
```diff
@@ -151,4 +166,92 @@ public static function getTimeLimitMax()
...
+        }
+    }
+
+    /**
+     * Load environment variables from .env file
+     *
+     * @param string $path Path to the file
+     * @return array List of values parsed as an associative array
+     */
+    public static function putenvFromFile($path)
```
sminnee: Why have we reimplemented dotenv?

src/includes/constants.php
```diff
@@ -58,39 +57,39 @@
 }
 
 // Allow a first class env var to be set that disables .env file loading
-if (!getenv('SS_IGNORE_DOT_ENV')) {
+if (!Environment::getenv('SS_IGNORE_DOT_ENV')) {
```
sminnee: This should probably still be getenv, otherwise it's a bit of a circular dependency, and this var can't be set by .env

src/Core/Environment.php
sminnee: Having looked at the other code that calls it, it seems making the setEnv API public and removing putenv would be more appropriate to developers.

tractorcow: I've ensure that putenv() promises to be a drop-in replacement for the core `putenv`. Mirroring the API means that users don't have to do any string-manipulation in their usercode.

src/Core/Environment.php
tractorcow: Pretty much!

src/Core/Environment.php
tractorcow: This api mirrors the old global `getenv`, so I've opted to maintain the semantics as close as possible, including case name.

src/Core/Environment.php
sminnee: I don't see much value in that. People will still have to change their code, and if they're going to change their code we may as well give them the API they're more likely to want, which is setEnv.

sminnee: If you look at all our putenv calls we're doing string concatenation in them now, so providing `setEnv()` isn't going to force people to do string manipulation.

If we had 5 years of code built around putenv there might be a stronger argument for it, but we don't.

src/includes/constants.php
tractorcow: This behaviour is required in case user-specific bootstrap files set environment vars.

Note that `Environment::getenv` doesn't actually require `.env` files to be loaded; It also acts as an accessor to the underlying system environment. This just maintains consistency between all getenv accessors.

src/Core/Environment.php
sminnee: Wouldn't it be better to retain a dotenv dependency (either vlucas or symfony) and make reference to its loader within `Env`?

src/includes/constants.php
sminnee: Fair enough

src/Core/Environment.php
tractorcow: I'll have a look into symfony.

src/Core/Environment.php
tractorcow: ok, fair enough.

Shall I make `putEnv` and `setEnv` both public?

src/Core/Environment.php
tractorcow: We can make it getEnv

src/Core/Environment.php
sminnee: Seems like a reasonable compromise.

src/Core/Environment.php
tractorcow: Symfony seems to do exactly the same thing; symfony dotenv simply uses core-php `getenv`. Given our implementation is small and works for us, I suggest to leave it as is.

sminnee: ok

tractorcow:
Ok, all feedback has been implemented and also on all other repos.

src/includes/constants.php
```diff
@@ -58,39 +57,39 @@
 }
 
 // Allow a first class env var to be set that disables .env file loading
-if (!getenv('SS_IGNORE_DOT_ENV')) {
+if (!Environment::getEnv('SS_IGNORE_DOT_ENV')) {
```
dhensby: Just revisiting what @sminnee said - this should only ever be set by a global bonefido environment variable that won't change from execution to execution. As the comment says "Allow a first class env var to be set that disables .env file loading" but this allows second-class / psuedo environment variables to manipulate/unset this value which is the opposite of what's intended here.

src/Core/Environment.php
```diff
@@ -151,4 +166,93 @@ public static function getTimeLimitMax()
...
+     *
+     * @param string $name
+     * @param string $value
+     */
+    public static function setEnv($name, $value)
+    {
+        // If PHP is running as an Apache module and an existing
+        // Apache environment variable exists, overwrite it
+        if (function_exists('apache_getenv') && function_exists('apache_setenv') && apache_getenv($name)) {
+            apache_setenv($name, $value);
```
dhensby: I don't understand the preference for handrolling our own dotenv work when this is practically identical to using `Dotenv\Loader::setEnvironmentVariable` and `Dotenv\Loader::getEnvironmentVariable` and means we don't have to maintain this code :/

dhensby: This is what the `setEnvironmentVariable` is in the library: https://github.com/vlucas/phpdotenv/blob/828d19e597052ddeee65890bb2b1a0912d79fea8/src/Loader.php#L353-L375

It's super close to what we have just we also have an extra layer of the local static `$env` too.

tests/php/Core/EnvironmentTest/test.env
```diff
@@ -0,0 +1,13 @@
+# Test file
```
dhensby: There are no tests for nested variables, eg:

```
TEST="value"
TEST_AGAIN="$TEST"
```

dhensby:
@zanderwar 

> The external library is not intended for production, so using it in the first place only satisfies the environment when in development albeit is not in require-dev and as such is adding unnecessary bloat to production.

I don't see how lifting the code from the library and putting it in our codebase makes this any more suitable for production. If we feel that this PR solves the problem, then using the built in methods (instead of php's `getenv` and `putenv`)  will solve the problem too.
 
> IMHO the dotenv dependency should be dropped entirely for methodology like this. Even if I am being biased here because this resolves my issue of wanting to use the .env file like laravel/symfony allow flawlessly

Laravel uses the same library for their `.env` so if you're happy with how laravel works then there should be no complaints. We could move to symfony's if it's more robust/resilient.

tractorcow:
> I don't see how lifting the code from the library and putting it in our codebase makes this any more suitable for production.

We aren't lifting it, we have made a completely unique implementation. We store the vars in a static $env to protected it from cross-process pollution (which apparently $_ENV and both $_SERVER were subject to in testing).

tests/php/Core/EnvironmentTest/test.env
tractorcow: I didn't write any code to support it either.

src/Core/Environment.php
tractorcow: > just we also have an extra layer of the local static $env too.

That is the critical point of this PR. That's not a cache, that's the key feature that solves the process isolation issue.

src/includes/constants.php
tractorcow: > but this allows second-class / psuedo environment variables to manipulate/unset this value which is the opposite of what's intended here.

`getenv` allows exactly the same if you use `putenv`, so what's the introduced risk that didn't exist before?

tractorcow:
Maybe someone who has an environment which can reproduce this issue (e.g. @zanderwar ) could take over development of this fix and see if they can minimise the fix to one that still solves the issue, but maybe without removal of the existing backend dotenv library?

tractorcow:
Maybe we should request feedback from one of the laravel devs; Maybe they can provide some insight that I haven't seen.

tests/php/Core/EnvironmentTest/test.env
NightJar: Are nested variables a thing? :o

tractorcow:
Well, I've tracked down the appropriate laravel discussion.

https://github.com/laravel/framework/issues/8191#issuecomment-300462507

> You should avoid calling env()in your application.

The way this is addressed in laravel is that the result of the combined config is cached, and thus there is no race condition when setting / loading `env()` vars, because getenv() is never called, and loading `.env` files is only done to populate this cache.

In the case of SilverStripe, what we could do is use `.env` to populate (and cache) config, and we use the config API.

We already use the back-tick operator to interpolate env / constants into config. What we could do is actually cache these within the config manifest (rather than lazy-evaluating on each request).

For example, this is how the default cache factory sets the temp path (cache.yml)

```yaml
---
Name: corecache
---
  SilverStripe\Core\Cache\CacheFactory:
    class: 'SilverStripe\Core\Cache\DefaultCacheFactory'
    constructor:
      args:
        directory: '`TEMP_PATH`'
```

This could be solved in a number of ways:

## Option 1

 - Move all used env vars to a config var on some relevant class
 - Update config module to interpolate these values during flush / cache generation
 - Replace all getenv() calls to the appropriate config calls

# Option 2

Alternatively, we could not interpolate these variables, but instead create a separate config transformer which populates config directly from the environment into a holder class (e.g. SilverStripe\Core\Environment). Thus `getenv('SS_DATABASE_CLASS')` becomes `Environment::config()->get('SS_DATABASE_CLASS')`.

tractorcow:
@dhensby what do you think about these other options?

sminnee:
We're close to an RC1 now is not the time to be radically re-thinking APIs.

Are we all :+1: on `Environment::setEnv($key, $value)` and `Environment::getEnv($key)` as the public API for this?

I love the idea of relying on the config's backtick system as much as possible, but that's a styleguide question, we'd still need the public API.

I, like Dan, questioned why we rewrote the dotenv package instead of just including it as a dependency, but I don't think that an argument over that should have anything to do with the public API we expose. And, because of that, we could change this in 4.1 if we want.

I *don't* think that we should expose `Dotenv\Loader::setEnvironmentVariable` and `Dotenv\Loader::getEnvironmentVariable` as our public API as it ties us to that specific implementation, and if we realise (like we did a few days ago) that we have a problem with their API then we're stuffed until SS5. However, I think that `Environment::setEnv()` calling these methods behind the scenes would make a lot of sense.

tractorcow:
@sminnee pro-tip; If we switch to Environment::getEnv() now, we can still switch the back-end cache at some later point (e.g. 4.1).

docs/en/00_Getting_Started/03_Environment_Management.md
```diff
@@ -32,13 +32,18 @@ You can set "real" environment variables using Apache. Please
...
-Accessing the environment varaibles is easy and can be done using the `getenv` method or in the `$_ENV` and `$_SERVER`
-super-globals:
+Accessing the environment varaibles should be done via the `Environment::getenv()` method
 
 ```php
-getenv('SS_DATABASE_CLASS');
-$_ENV['SS_DATABASE_CLASS'];
-$_SERVER['SS_DATABASE_CLASS'];
+use SilverStripe\Core\Environment;
+Environment::getenv('SS_DATABASE_CLASS');
```
sminnee: getEnv

docs/en/00_Getting_Started/03_Environment_Management.md
```diff
@@ -32,13 +32,18 @@ You can set "real" environment variables using Apache. Please
...
+```
+
+Individual settings can be assigned via `Environment::putenv()` method.
+
+```php
+use SilverStripe\Core\Environment;
+Environment::putenv('API_KEY="AABBCCDDEEFF012345"');
```
sminnee: setEnv('API_KEY', 'AABBCCDDEEFF012345')

docs/en/00_Getting_Started/03_Environment_Management.md
```diff
@@ -47,11 +52,9 @@ Sometimes it may be useful to include an extra `.env` file - on a shared local d
...
 
 ```php
-try {
-    (new \Dotenv\Dotenv('/path/to/env/'))->load();
-} catch (\Dotenv\Exception\InvalidPathException $e) {
-    // no file found
-}
+use SilverStripe\Core\Environment;
+$env = BASE_PATH . '/mysite/.env';
+Environment::putenvFromFile($env);
```
sminnee: putEnvFromFile

tractorcow:
I'll fix up the docs now. :)

tests/php/Core/Manifest/ConfigManifestTest.php
```diff
@@ -100,8 +100,7 @@ public function testModuleRules()
 
     public function testEnvVarSetRules()
     {
-        $loader = new Loader(null);
-        $loader->setEnvironmentVariable('ENVVARSET_FOO', 1);
+        Environment::putEnv('ENVVARSET_FOO=1');
```
sminnee: setEnv('ENVVARSET_FOO', 1)

tests/php/Core/Manifest/ConfigManifestTest.php
```diff
@@ -137,9 +136,7 @@ public function testConstantDefinedRules()
 
     public function testEnvOrConstantMatchesValueRules()
     {
-        $loader = new Loader(null);
-
-        $loader->setEnvironmentVariable('CONSTANTMATCHESVALUE_FOO', 'Foo');
+        Environment::putEnv('CONSTANTMATCHESVALUE_FOO=Foo');
```
sminnee: setEnv('CONSTANTMATCHESVALUE_FOO', 'Foo')

tests/php/Core/Manifest/ConfigManifestTest.php
```diff
@@ -194,9 +191,7 @@ public function testEnvironmentRules()
 
     public function testMultipleRules()
     {
-        $loader = new Loader(null);
-
-        $loader->setEnvironmentVariable('MULTIPLERULES_ENVVARIABLESET', 1);
+        Environment::putEnv('MULTIPLERULES_ENVVARIABLESET=1');
```
sminnee: setEnv('MULTIPLERULES_ENVVARIABLESET', 1);

tractorcow:
Docs and inconsistencies fixed up.

tests/php/Core/EnvironmentTest/test.env
flamerohr: please don't let this be a thing... at least not here :P

NightJar: My thoughts exactly.

src/Core/Environment.php
```diff
@@ -151,4 +166,93 @@ public static function getTimeLimitMax()
...
+                    // Strip all # comments, including start or end of string, but not in quoted strings
+                    // '#comment', 'var=thing#comment', or 'var="not#comment"#comment
+                    return preg_replace('/((^[^"]*)|(^[^"]+("[^"]*){2}))#.*$/', '$1', $line);
+                },
+                explode(PHP_EOL, file_get_contents($path))
+            ))
+        );
+
+        // Pass to parse_ini_string
+        $variables = parse_ini_string($contents) ?: [];
```
NightJar: Actually, this is not ideal, as `#` is not a comment marker; `;` is, as per ini style.
As we discussed elsewhere, I'd leave inline # behaviour undefined, and only modify beginning of line comments as per https://github.com/NightJar/sapphire/blob/e614edec61f396c7e8f29bd5ee3340f56cf0b1a6/src/Core/Environment.php#L35-L36

This would reduce complexity, maintenance, and 'regex soup' that others have also raised.

NightJar:
Bad news team, I can still replicate.

tractorcow:
Yep, this issue isn't as easy as I thought it would be.

With this fix I attempted to break as little as possible (i.e. support getenv() legacy calls).

It may be necessary to act in a more breaking way, and stop relying on $_ENV / getenv at all for our environment variables, other than perhaps an initial "inherit from environment" preload prior to `.env` loading.

sminnee:
`getenv()` is fine – and a critical feature. It's `putenv()` that's the problem.

tractorcow:
Since laravel caches their config, it may be necessary for us to cache inherited env to a similar cache, cutting down on `getenv` calls on every request.

E.g. pre-cache $_ENV to `.env.global`, cleared on flush, and load from this instead of $_ENV.

getenv() is actually still a problem. I'd rather not use any on non-initial requests and just use the internal store.

sminnee:
> getenv() is actually still a problem. I'd rather not use any on non-initial requests and just use the internal store

What's the problem with `getenv()`? The problem with caching getenv is that's it's hard to know when to invalidate that cache.

tractorcow:
@sminnee it goes back to that laravel thread at https://github.com/laravel/framework/issues/8191#issuecomment-300462507 

They now only use config instead of env when loading config on each subsequent request.

sminnee:
It feels like it's calling the whole https://12factor.net/  standard into question. It's like finding out that you shouldn't really use PSR-4.

sminnee:
It's possible that it's better to deprecate whatever specific aspect of the platform Nightjar is running that still causes the issue.

NightJar:
To reiterate, @sminnee is correct, there is _no_ issue with PHP's `getenv`.
The issue purely springs from using `putenv` per request.

tractorcow:
I'll defer to your better experience in this case. :)

sminnee:
So in summary:

 * Environment::getEnv() uses `getenv()` as its starting point
 * .env files are loaded into static properties, and don't rely on `putenv()`
 * We recommend system-defined environment variables and not .env in prod environments

Right?

Perhaps we ditch the `putenv()` call to make it obvious that you should use `Environment::getEnv()` in your application code and not `getenv()`. In particular, if you're running with a .env file that means it will fail consistently, instead of only failing in a race condition, which seems like better developer experience.

tests/php/Core/EnvironmentTest/test.env
dhensby: This is a feature of the dotenv lib so if we're replacing it like-for-like it needs to be here.

dhensby:
> I don't think that we should expose `Dotenv\Loader::setEnvironmentVariable` and `Dotenv\Loader::getEnvironmentVariable` as our public API as it ties us to that specific implementation, and if we realise (like we did a few days ago) that we have a problem with their API then we're stuffed until SS5. However, I think that Environment::setEnv() calling these methods behind the scenes would make a lot of sense.

This is my view. We can use our own API which wraps the `DotEnv` library and even adds the local caching layer, but we shouldn't be removing it as a dependency at this time.

It also seems that this fix doesn't actually solve the problem anyway as we still rely on  `putenv` (and `getenv` to an extent) which aren't thread-safe...

tractorcow:
I may update so that Environment::setEnv() simply sets the $env static, and we don't modify the environment at all.

tractorcow:
One more iteration, with these changes:

 - Move env file loading into EnvironmentLoader. It's <120 lines so I don't feel too bad about replicating dotenv library logic.
 - Implement variable interpolation at @dhensby's request.
 - No longer uses putenv(). Environment::setEnv() now exclusively sets a protected static var, which is used as the authoritative source for vars loaded from `.env`.
 - getenv() no longer loads variables loaded in this request; Environment::getEnv() must be used by ss applications.

tests/php/Core/EnvironmentTest/test.env
tractorcow: I've restored this. ;)

sminnee:
Final version looks good to me; let's merge this.

sminnee:
@NightJar does the current solution fix the bugs that you experienced?

tractorcow:
I think supporting escaped quotes can be a later release fix, if necessary.

tractorcow:
Shall we escew dotenv / parse_ini_string() for say, one of the below:

- https://github.com/josegonzalez/php-dotenv/blob/master/src/josegonzalez/Dotenv/Loader.php
- https://github.com/m1/Env

Unlike the prior library, these support the ability for us to parse and extract, but not populate `putenv`.

@dhensby @sminnee opinions?

I'm not really interested in complicating my regexp any more.

tractorcow:
Switched out env parser for m1/env. @dhensby see, no internal-env-parser implementation anymore. :) And feature parity.

NightJar:
Attempting to test.

NightJar:
This patch; good news.
Can replicate without.
Cannot replicate with.

sminnee:
Merged, and merged 4.0 -> 4 as well.

dhensby:
Yay

GrahamCampbell:
The current version of phpdotenv is robust and thread-safe. The following example will work in a multithreaded environment. This avoids using the adapter that would have called `putenv` and `getenv`, which are not threadsafe.

```php
<?php

use Dotenv\Environment\Adapter\EnvConstAdapter;
use Dotenv\Environment\Adapter\ServerConstAdapter;
use Dotenv\Environment\DotenvFactory;
use Dotenv\Dotenv;

$factory = new DotenvFactory([new EnvConstAdapter(), new ServerConstAdapter()]);

Dotenv::create($path, null, $factory)->load();
```

tractorcow:
That's excellent, thank you @GrahamCampbell. I think the approach we've gone with is the best for us, however, but we appreciate your getting back to us.

PR #7485, opened 2017-10-17
BUG Ensure sake works when called from any directory

tractorcow:
Credit to the kind users of silverstripe-users slack for testing this out.

Checks have been simplified and cleaned up, checking:
 - relative path from vendor/bin/sake. (simlink dir)
 - relative path from vendor/silverstripe/framework/sake (real dir)
 - base relative search for vendor/silvestripe/framework (for global sake install)
 - base relative search for legacy framework (in case run on an old framework installation, for global sake install)

### DIFF
--- a/sake
+++ b/sake
@@ -9,39 +9,29 @@ Executes a SilverStripe command"
 	exit 1
 fi
 
-# find the silverstripe installation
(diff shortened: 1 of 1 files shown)


mikenz:
Thanks, works perfectly in all my scenarios!

Issue #7470, opened 2017-10-11
SS4: Email class doesn't find template of own Email implementation


wernerkrauss:
In my subclass of Email, the template isn't set to `MyEmail::class` but by default to `Email::class`

I suggest to leave `$HTMLTemplate` empty and use `SSViewer::get_templates_by_class(static::class, '', self::class);` if it's not overwritten manually.

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/7490

tractorcow:
Seems easy enough!

PR #7490, opened 2017-10-18
BUG Fix enable email subclasses to use their respective templates

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/7470

### DIFF
--- a/src/View/ThemeResourceLoader.php
+++ b/src/View/ThemeResourceLoader.php
@@ -103,7 +103,7 @@ class ThemeResourceLoader
             if (count($parts) > 1) {
                 throw new InvalidArgumentException("Invalid theme identifier {$identifier}");
             }
-            return substr($identifier, 1);
+            return ltrim($identifier, '/');
         }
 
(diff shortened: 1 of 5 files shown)


tractorcow:
I fixed this and updated tests AND I fixed all those old warnings on EmailTest. :)

tractorcow:
Oh, the test wasn't the one you wanted @kinglozzer. Sorry I'll add it now. :)

src/View/ThemeResourceLoader.php
```diff
@@ -148,7 +148,7 @@ public function getPath($identifier)
         }
 
         // Join module with subpath
-        return $modulePath . $subpath;
+        return ltrim($modulePath . $subpath, '/');
```
tractorcow: Fixes a bug where $modulePath is empty but $subPath has a leading `/`

flamerohr:
Nice, thanks for the additional fixes and unit tests @tractorcow 
they look good to me

kinglozzer:
Nice job, thanks both!

PR #7491, opened 2017-10-18
DOC Add example for using resource location for pages / modeladmin

SilbinaryWolf:
DOC Add example for using resource location for pages / modeladmin


dhensby:
thanks!

PR #7493, opened 2017-10-18
FIX: Use isolated scope when requiring files for module activation

kinglozzer:
Prevents access to `$this` inside `_config.php` files, and removes the need for [hacks to avoid exposing variables](https://github.com/silverstripe/silverstripe-asset-admin/blob/1.0.0-beta4/_config.php#L7) to other scopes.

Same approach used by [Composer](https://github.com/composer/composer/blob/e42e1156d51ca5494b058b1a7b480bc703c4f57a/src/Composer/Autoload/ClassLoader.php#L437-L445).

### DIFF
--- a/src/Core/Manifest/Module.php
+++ b/src/Core/Manifest/Module.php
@@ -158,7 +158,7 @@ class Module implements Serializable
     {
         $config = "{$this->path}/_config.php";
         if (file_exists($config)) {
-            require_once $config;
+            requireFile($config);
         }
     }
 
@@ -246,4 +246,16 @@ class Module implements Serializable
             ->getResource($path)
             ->exists();
     }
+}
+
+/**
+ * Scope isolated require - prevents access to $this, and prevents module _config.php
+ * files potentially leaking variables. Required argument $file is commented out
+ * to avoid leaking that into _config.php
+ *
+ * @param string $file
+ */
+function requireFile()
+{
+    require_once func_get_arg(0);
 }


dhensby:
love it - do we have any thoughts on whether this is a bug fix or not ;)

kinglozzer:
I’ll argue that being able to access `$this` (`SilverStripe\Core\Manifest\Module`) in `_config.php` is a bug just so I don’t have to amend my commit message 😅

flamerohr:
yay, does this mean we can remove that hack? :D

kinglozzer:
> yay, does this mean we can remove that hack? :D

Yep, unless there’s an edge case I can’t find! Removing the `call_user_func`, I could only access the “exposed” variables inside `Module::activate()`, and that no longer happens after this change. I couldn’t access/influence them from anywhere else before or after this change!

src/Core/Manifest/Module.php
```diff
@@ -247,3 +247,14 @@ public function hasResource($path)
...
+
+/**
+ * Scope isolated require - prevents access to $this, and prevents module _config.php
+ * files potentially leaking variables
+ *
+ * @param string $file
+ */
+function requireFile($file)
+{
+    require_once $file;
```
flamerohr: I was doing a bit of light reading, and have found that `$file` leaks into `_config.php` this way. The way suggested was to use `require func_get_arg(0);`

This is quite minor I guess because `$file` isn't such a big deal to leak in to.

kinglozzer: I like it, if we’re overkilling this then let’s do it properly 😉

PR #7494, opened 2017-10-18
Small fix to nginx.conf

martinduparc:
Source: https://github.com/silverstripe/silverstripe-framework/blob/4.0/docs/en/00_Getting_Started/01_Installation/How_To/Setup_Nginx_and_HHVM.md


dhensby:
thanks

PR #7495, opened 2017-10-18
WIP: Adding 308 HTTP Response as an redirect code

nathanbrauer:
https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml

Often times, 308 are preferred over 301 redirects, as they tell the requester to, for example, re-POST a form submission on the final URL.


nathanbrauer:
Attn: @brandon32

nathanbrauer:
@tractorcow - I've come to learn that this comes with a minimum requirement of Apache 2.4 since 308 response codes, in Apache 2.2 are swapped out for 500.

PR #7499, opened 2017-10-19
ENHANCEMENT Don't fail immediately on imagick install

tractorcow:


.travis.yml
```diff
@@ -58,7 +58,7 @@ matrix:
 before_script:
 # Init PHP
   - pecl channel-update pecl.php.net
-  - printf "\n" | pecl install imagick
+  - NEXT_WAIT_TIME=0 && until (printf "\n" | pecl install imagick) || [[ $NEXT_WAIT_TIME == "4" ]]; do sleep $(( NEXT_WAIT_TIME++ )); done
```
dhensby: use `travis_retry`? or even just add `|| true` like we do for other non-essential build steps?

flamerohr: `travis_retry` sounds like a better idea for this :)

dhensby:
I just saw a build where the `travis_retry` didn't seem to  have any effect - perhaps the pipe means that they are being interpreted as two separate commands rather than 1?

flamerohr:
yeah, we've noticed that too

haven't had the chance to look into it, but exploring the idea of just removing imagick if possible? by mocking the APIs being called

tractorcow:
Yes it's not working; let's remove it, and consider it sufficient to test that interventionimage works; Testing individual backends should be the role of intervention image (third party) and not framework anymore.

PR #7503, opened 2017-10-20
Spelling in DataQueryTest

andrewandante:


dhensby:
@kinglozzer do we think this is safe (regarding SemVer)?

tractorcow:
I'd go ahead and say that the names of specific tests don't really count as "public API" so I'll give it my tick. ;)

Issue #6187, opened 2016-10-16
Document Changesets for Developers


chillu:
# Acceptance Criteria
- As a developer new to SilverStripe I can understand how it versions content (and the underlying data model and database schema)
- As an existing SilverStripe developer upgrading to 4.0 I can make an informed decision if I should add "ownership" and versioned to my existing models
- The existing [Versioned docs](https://github.com/silverstripe/silverstripe-framework/blob/master/docs/en/02_Developer_Guides/00_Model/10_Versioning.md) are accurate
# Notes
- Could be a condensed version of the [RFC](https://github.com/silverstripe/silverstripe-framework/issues/4932)
- Should either improve the existing [4.0 changelog](https://github.com/silverstripe/silverstripe-framework/blob/master/docs/en/04_Changelogs/4.0.0.md#changeset-batch-publishing), or link to more permanent docs
- An existing [blog post](https://www.silverstripe.org/blog/campaigns-in-silverstripe-4/) might help to simplify the language of this

## Pull Requests
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/7505

PR #7505, opened 2017-10-23
Document changesets

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-framework/issues/6187

### DIFF
--- a/docs/en/02_Developer_Guides/00_Model/10_Versioning.md
+++ b/docs/en/02_Developer_Guides/00_Model/10_Versioning.md
@@ -122,16 +122,49 @@ automatically joined as required:
  * `MyRecordSubclass_Live` table: Contains only live data for subclass columns
  * `MyRecordSubclass_Versions` table: Contains only version history for subclass columns
 
-While not explicit DataObjects, `many_many` relationships create their own sets of records on their own tables.
-These records represent content changes to a DataObject, and are therefore versioned.
-If you have, for instance, a versioned `Product` DataObject with `many_many` categories, the following tables will be created:
+Because `many_many` relationships create their own sets of records on their own tables, representing content changes to a DataObject, they can therefore be versioned. This is done using the ["through" setting](https://docs.silverstripe.org/en/4/developer_guides/model/relations/#many-many-through-relationship-joined-on-a-separate-dataobject) on a `many_many` definition. This setting allows you to specify a custom DataObject through which to map the `many_many` relation. As such, it is possible to version your `many_many` data by versioning a "through" dataobject. For example:
 
-* `Product`
-* `Product_Live`
-* `Product_versions`
-* `Product_Categories`
-* `Product_Categories_Live`
-* `Product_Categories_versions`
+```php
+use SilverStripe\ORM\DataObject;
+
+class Product extends DataObject
+{
+    private static $db = [
+        'Title' => 'Varchar(100)',
+        'Price' => 'Currency',
+    ];
+    
+    private static $many_many = [
+        'Categories' => [
+            'through' => 'ProductCategory',
+            'from' => 'Product',
+            'to' => 'Category',
+        ],
+    ];
+}
+```
+
+```php
+use SilverStripe\ORM\DataObject;
+use SilverStripe\Versioned\Versioned;
+
+class ProductCategory extends DataObject
+{
+    private static $db = [
+        'SortOrder' => 'Int',
+    ];
+    
+    private static $has_one = [
+        'Product' => Product::class,
+        'Category'=> Category::class,
+    ];
+    
+    private static $extensions = [
+        Versioned::class,
+    ];
+}
+```
+
 
 ## Usage
 
@@ -331,7 +364,7 @@ the shortcode references the database identifier of the `Image` object.
 
 ### Changesets, a.k.a "Campaigns"
 
-Changes to many DataObjects can grouped together using the `ChangeSet` object, better known by its frontend name, "Campaign" (provided the `campaign-admin` module is installed). By grouping a series of content changes together as one cohesive unit, content editors can bulk publish or revert an entire body of content all at once, which affords them much more power and control over interdependent content types.
+Changes to many DataObjects can grouped together using the `ChangeSet` [api:SilverStripe\Versioning\ChangeSet] object, better known by its frontend name, "Campaign" (provided the `campaign-admin` module is installed). By grouping a series of content changes together as one cohesive unit, content editors can bulk publish an entire body of content all at once, which affords them much more power and control over interdependent content types.
 
 Records can be added to a changeset in the CMS by using the "Add to campaign" button
 that is available on the edit forms of all pages and files. Programmatically, this is done by creating a `SilverStripe\Versioned\ChangeSet` object and invoking its `addObject(DataObject $record)` method.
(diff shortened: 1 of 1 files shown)


docs/en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -122,6 +122,17 @@ automatically joined as required:
  * `MyRecordSubclass_Live` table: Contains only live data for subclass columns
  * `MyRecordSubclass_Versions` table: Contains only version history for subclass columns
 
+While not explicit DataObjects, `many_many` relationships create their own sets of records on their own tables.
```
tractorcow: That's not entirely true; `many_many` supports `through` syntax which does use an explicit DataObject. You should make the distinction here and describe how it differs from the traditional syntax with auto-scaffolded table.

docs/en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -122,6 +122,17 @@ automatically joined as required:
...
+These records represent content changes to a DataObject, and are therefore versioned.
+If you have, for instance, a versioned `Product` DataObject with `many_many` categories, the following tables will be created:
+
+* `Product`
+* `Product_Live`
+* `Product_versions`
+* `Product_Categories`
+* `Product_Categories_Live`
+* `Product_Categories_versions`
```
tractorcow: This is the opposite to what is true.... many_many is NOT automatically versioned. To get many_many versioning you MUST use the `through` form of many_many and apply versioning to that intermediary dataobject class.

docs/en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -318,6 +329,47 @@ The ownership relationship is tracked through an `[image]` [shortcode](/develope
 which is automatically transformed into an `<img>` tag at render time. In addition to storing the image path,
 the shortcode references the database identifier of the `Image` object.
 
+### Changesets, a.k.a "Campaigns"
+
+Changes to many DataObjects can grouped together using the `ChangeSet` object, better known by its frontend name, "Campaign" (provided the `campaign-admin` module is installed). By grouping a series of content changes together as one cohesive unit, content editors can bulk publish or revert an entire body of content all at once, which affords them much more power and control over interdependent content types.
```
tractorcow: We shouldn't be describing "revert" here since that functionality isn't built yet.

tractorcow: Please add an `[api:SilverStripe\Versioning\ChangeSet]` link here to the class in question in place of `ChangeSet`

docs/en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -318,6 +329,47 @@ The ownership relationship is tracked through an `[image]` [shortcode](/develope
...
+
+By including an object in a changeset, the user is implicitly including all of that object's owned records (as declared by the `$owns` setting). This ensures that when a changeset is published (or reverted), the action cascades through not only all of the items explicitly added to the changeset, but also all of the records that each of those items owns.
+
+#### ChangeSet actions
+
+Actions available on the frontend, i.e. those which are intended to be triggered by an end user, include:
+
+* `$myChangeSet->addObject(DataObject $record)`: Add a record and all of its owned records to the changeset (`canEdit()` dependent)
+* `$myChangeSet->removeObject(DataObject $record)`: Removes a record and all of its owned records from the changeset (`canEdit()` dependent)
+* `$myChangeSet->::publish()`: Publishes all items in the changeset that have modifications, along with all their owned records (`canPublish()` dependent). Closes the changeset on completion.
```
tractorcow: Inconsistent syntax `class->::method()` with the above.

docs/en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -318,6 +329,47 @@ The ownership relationship is tracked through an `[image]` [shortcode](/develope
...
+* `$myChangeSet->::revert()`: Reverts all items in the changeset that have modifications, along with all their owned records (`canRevert()` dependent). Closes the changeset on completion.
+ 
+ <div class="info" markdown="1">
+ Changesets cannot be "unpublished." Instead, a new changeset should be created to reverse the actions of the prior one.
```
tractorcow: It's not really clear at the moment how this should be done; We should err in favour of not documenting behaviour we have no recommendation for at this stage. At 5.0 we may have a proper campaign history / revert UI and API, at which stage we should describe that behaviour instead.

docs/en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -318,6 +329,47 @@ The ownership relationship is tracked through an `[image]` [shortcode](/develope
...
+ 
+ #### ChangeSet states
+ 
+ Changesets can exists in five different states:
+ 
+* `open` No action has been taked on the changeset. Resolves to `publishing` or `reverting`.
+* `publishing`:  The changeset is in the process of publishing its items. Resolves to to `published`.
+* `reverting`: The changeset is in the process of reverting changes to its items. Resolves to `reverted`.
+* `published`: The changeset has published changes to all of its items and its now closed.
+* `reverted`: The changeset has reverted changes to all of its items and its now closed.
```
tractorcow: I'd note that `reverted` and `reverting` are placeholders for future states that currently cannot be set.

tractorcow: `publishing` and `reverting` aren't actually states, after checking. Please remove those.

docs/en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -318,6 +329,47 @@ The ownership relationship is tracked through an `[image]` [shortcode](/develope
 which is automatically transformed into an `<img>` tag at render time. In addition to storing the image path,
 the shortcode references the database identifier of the `Image` object.
 
+### Changesets, a.k.a "Campaigns"
```
tractorcow: Consistent capitalisation; I.e. `ChangeSet` vs `Changeset`

docs/en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -318,6 +329,47 @@ The ownership relationship is tracked through an `[image]` [shortcode](/develope
...
+ 
+ Changesets can exists in five different states:
+ 
+* `open` No action has been taked on the changeset. Resolves to `publishing` or `reverting`.
+* `publishing`:  The changeset is in the process of publishing its items. Resolves to to `published`.
+* `reverting`: The changeset is in the process of reverting changes to its items. Resolves to `reverted`.
+* `published`: The changeset has published changes to all of its items and its now closed.
+* `reverted`: The changeset has reverted changes to all of its items and its now closed.
+ 
 ### Custom SQL
```
tractorcow: We should also document the changetype, and describe what each of them means in the campaigns view. I.e.:

- created
- modified
- deleted
- none

docs/en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -318,6 +329,47 @@ The ownership relationship is tracked through an `[image]` [shortcode](/develope
...
+
+Records can be added to a changeset in the CMS by using the "Add to campaign" button
+that is available on the edit forms of all pages and files. Programmatically, this is done by creating a `SilverStripe\Versioned\ChangeSet` object and invoking its `addObject(DataObject $record)` method.
+
+<div class="info" markdown="1">
+Any DataObject can exist in any number of changesets, and even added to a changeset in advance of being published. While a record need not have modifications to be part of a changeset, for practical purposes, changesets are only concerned with records that have modifications.
+</div>
+
+By including an object in a changeset, the user is implicitly including all of that object's owned records (as declared by the `$owns` setting). This ensures that when a changeset is published (or reverted), the action cascades through not only all of the items explicitly added to the changeset, but also all of the records that each of those items owns.
```
tractorcow: It's worth adding a distinction and full explanation of the difference between "implicit" and "explicit" added items, and that all objects are either of the two flags. Items that are both explicitly and implicitly added are treated as explicit, which isn't obvious to the user.

PR #7507, opened 2017-10-24
Improve changelog

chillu:
See detail commit messages. Ideally this makes it into rc1 (at least the upgrade.yml aspects, since they're bound to the actually released code, so can't go into the `4` branch)

### DIFF
--- a/.upgrade.yml
+++ b/.upgrade.yml
@@ -1008,6 +1008,12 @@ warnings:
       message: 'Removed'
     'Image_Cached':
       message: 'Removed'
+    'CMSFileAddController':
+      message: 'Removed'
+    'CMSBatchAction_Delete':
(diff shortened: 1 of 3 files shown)


chillu:
Regarding the composer.json instructions: The idea with separate `@dev` constraints is that we can remove `minimum-stability: dev` once core goes stable, and just instruct people to remove `@dev` once the modules in question go stable. 

I'm not sure why `"silverstripe/recipe-cms": "~1.1"` or `"silverstripe/recipe-cms": "~1.0"` with `prefer-stable: true` installs ` 1.x-dev` instead of the `1.0.0-beta4` tag - which then successively installs `-dev` branches of all core modules. @tractorcow Any ideas?

PR #7508, opened 2017-10-24
ENHANCEMENT Ensure DBVarchar scaffolds text field with TextField with appropriate max length

### LINKED ISSUE #1413: [2011-08-17] Member unique identifier field allows duplicates through db column truncation
created by: @chillu (ischommer)
created at: 2011-08-17
original ticket: http://open.silverstripe.org/ticket/6703

---

Member::get_unique_identifier_field() is set to Email by default. If somebody signs up through the forum module, or changes their profile info, they can create duplicates in the database by entering a string thats longer than the database field length (by default, a varchar 50).

Issue 1: The RFC for emails states a max length of 255, so our Member defaults are too short - see
http://tools.ietf.org/html/rfc5321#section-4.5.3.1 (4.5.3.1.3.). Thats easily increased, but doesn't solve the truncation issue.

Issue 2: The length of input isn't validated in PHP automatically - basically, no data-level validation takes place even when using a RequiredFields validator (which isn't named very well). Thats a broader scope than we can solve here.

Issue 3: Member->php() and Member->onBeforeWrite() don't artificially truncate the field before checking for valid identifiers. Ideally a generic data validation from Issue 2 would catch this, but until thats implemented we'll need this fallback.

PR https://github.com/silverstripe/silverstripe-framework/pull/7508

tractorcow:
Fixes #1413

Appropriate enough fix for a 4 year old issue. :)

Technically an enhancement, but it helps avoid errors in user applications.

### DIFF
--- a/src/ORM/FieldType/DBVarchar.php
+++ b/src/ORM/FieldType/DBVarchar.php
@@ -2,10 +2,11 @@
 
 namespace SilverStripe\ORM\FieldType;
 
-use SilverStripe\ORM\DB;
 use SilverStripe\Core\Config\Config;
-use SilverStripe\Forms\TextField;
 use SilverStripe\Forms\NullableField;
+use SilverStripe\Forms\TextField;
+use SilverStripe\ORM\Connect\MySQLDatabase;
+use SilverStripe\ORM\DB;
 
 /**
  * Class Varchar represents a variable-length string of up to 255 characters, designed to store raw text
@@ -22,6 +23,11 @@ class DBVarchar extends DBString
         "URL" => "Text",
     );
 
+    /**
+     * Max size of this field
+     *
+     * @var int
+     */
     protected $size;
 
     /**
@@ -58,8 +64,8 @@ class DBVarchar extends DBString
      */
     public function requireField()
     {
-        $charset = Config::inst()->get('SilverStripe\ORM\Connect\MySQLDatabase', 'charset');
-        $collation = Config::inst()->get('SilverStripe\ORM\Connect\MySQLDatabase', 'collation');
+        $charset = Config::inst()->get(MySQLDatabase::class, 'charset');
+        $collation = Config::inst()->get(MySQLDatabase::class, 'collation');
 
         $parts = array(
             'datatype'=>'varchar',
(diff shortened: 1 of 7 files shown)


dhensby:
seems legit - do we want this patch in 3.x too?

src/ORM/FieldType/DBVarchar.php
```diff
@@ -117,12 +123,14 @@ public function RTF()
...
-        } else {
-            // Automatically determine null (empty string)
-            return parent::scaffoldFormField($title);
+        // Set field with appropriate size
+        $field = TextField::create($this->name, $title);
+        $field->setMaxLength($this->getSize());
+
+        // Allow the user to select if it's null instead of automatically assuming empty string is
+        if (!$this->getNullifyEmpty()) {
+            $field = NullableField::create($field);
```
flamerohr: this if condition could be at the top and return early

src/ORM/FieldType/DBVarchar.php
```diff
@@ -117,12 +123,14 @@ public function RTF()
...
     public function scaffoldFormField($title = null, $params = null)
     {
-        if (!$this->nullifyEmpty) {
-            // Allow the user to select if it's null instead of automatically assuming empty string is
-            return new NullableField(new TextField($this->name, $title));
-        } else {
-            // Automatically determine null (empty string)
-            return parent::scaffoldFormField($title);
+        // Set field with appropriate size
+        $field = TextField::create($this->name, $title);
```
flamerohr: this is preferable to redefine the behaviour than to use `parent::scaffoldFormField($title, $params)`?

src/ORM/FieldType/DBVarchar.php
tractorcow: It can't be at the top because it needs the pre-constructed $field nested inside it.

src/ORM/FieldType/DBVarchar.php
tractorcow: I use this because it's a definite TextField, but scaffoldFormField only hints FormField.

tractorcow:
Tweaked code based on suggestions and fixed.

PR #7511, opened 2017-10-25
Warn if env isn't writable

tractorcow:


src/Dev/Install/InstallRequirements.php
```diff
@@ -259,6 +259,8 @@ public function check()
 
         $this->requireWriteable('index.php', array("File permissions", "Is the index.php file writeable?", null));
 
+        $this->requireWriteable('.env', ["File permissions", "Is the .env file writeable?"], false, false);
```
flamerohr: missing `null` for third param, other than that it works fine :)

tractorcow: Oops right. Fixed.

PR #7512, opened 2017-10-25
FIX Do not escape the readonly values since they get escaped when rendered

### LINKED ISSUE #7474: SS4: TreeDropdownField_Readonly / LookupField returns escaped html
SS4-beta4: The TreeDropdownField_Readonly / LookupField returns escaped(? .. html-encoded) html code in the read-only fields. Below a sample of a non-editable RedirectorPage:

screen shot 2017-10-12 at 15 37 21-fullpage

robbieaverill:
Since, in SS4, HTMLFragment is escaped by default in the template, this line of code results in double escaping.

Fixes #7474

### DIFF
--- a/src/Forms/LookupField.php
+++ b/src/Forms/LookupField.php
@@ -53,7 +53,6 @@ class LookupField extends MultiSelectField
         if ($mapped) {
             $attrValue = implode(', ', array_values($mapped));
 
-            $attrValue = Convert::raw2xml($attrValue);
             $inputValue = implode(', ', array_values($values));
         } else {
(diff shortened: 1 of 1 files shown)


tractorcow:
Wait, HTMLFragment is not escaped... that's the point of marking it as HTML to prevent another encoding. @robbieaverill can you show me your reproduction of this issue?

robbieaverill:
Discussed via Slack - the line removed here should be put back, and the original input data should be fixed instead

tractorcow:
Yeah the bug seems to be in treedropdownfield_readonly. We should be using getTitleField()  not getLabelField() now, since our refactor.

tractorcow:
Better fix at https://github.com/silverstripe/silverstripe-framework/pull/7517

PR #7517, opened 2017-10-26
BUG Ensure readonly tree dropdown is safely encoded

tractorcow:
Removed legacy entwine dead code
Added soft-deprecation to label field

Replaces prior fix https://github.com/silverstripe/silverstripe-framework/pull/7512

Fixes https://github.com/silverstripe/silverstripe-framework/issues/7474

### DIFF
--- a/src/Forms/TreeDropdownField.php
+++ b/src/Forms/TreeDropdownField.php
@@ -2,18 +2,16 @@
 
 namespace SilverStripe\Forms;
 
+use Exception;
+use InvalidArgumentException;
 use SilverStripe\Assets\Folder;
 use SilverStripe\Control\HTTPRequest;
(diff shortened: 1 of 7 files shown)


src/Forms/TreeDropdownField.php
```diff
@@ -633,6 +598,9 @@ public function getAttributes()
     }
 
     /**
+     * HTML-encoded label for this node, including css classes and other markup.
+     *
+     * @deprecated 4.0...5.0 Use setTitleField()
```
flamerohr: would a deprecation notice for these two methods be worth it?

tractorcow: We still need to actually call this method, so that would trigger errors even in code that doesn't call it.

Issue #7304, opened 2017-08-23
BUG SS361: User can accidentally remove self from the own Group


alessandromarotta:
## Acceptance Criteria

 * As a logged in user with access to the "Security" section, I can remove myself from groups
 * As a logged in user with access to the "Security" section, I can't remove myself from any group that would deny me further access to that section (groups with ADMIN permission)
 * In the "Group" edit view, I can't unlink from an ADMIN group in the "Members" GridField (either by hiding the action, or through a validation error)
 * In the "Member" edit view, I get a validation error when removing from an ADMIN group

## Notes

* No need for confirmation dialogs (too complex)
* Users without ADMIN, but CMS_ACCESS_SecurityAdmin permissions are an edge case that we can ignore. It means there's likely other admins left in the system to restore access if needed.
* Long term, the ListboxField should enforce this rather than relying on validation (`clearableValue: false` see https://github.com/silverstripe/silverstripe-admin/issues/52 and https://github.com/JedWatson/react-select#multiselect-options)

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/7518
- [x] https://github.com/silverstripe/silverstripe-testsession/pull/54
- [ ] https://github.com/silverstripe/silverstripe-behat-extension/pull/171

## Original post
Hi there!
I found out a bug (IMHO) about the /admin/security/EditForm/field/Groups/ mask.

Every user can remove self from the own Group clicking on the unlink button.
I think that it's a bad idea.

![image](https://user-images.githubusercontent.com/17084152/29611410-b6c08c38-87fd-11e7-8a85-13414135dd91.png)

Bye,
Alessandro

## Note
ACs and details can also be found here https://github.com/silverstripe/silverstripe-cms/issues/730

chillu:
Confirmed that's the case - it's a regression from 3.x (not possible there).

tractorcow:
> Confirmed that's the case - it's a regression from 3.x (not possible there).

It seems to be possible in 3.x.

flamerohr:
older issue here: https://github.com/silverstripe/silverstripe-cms/issues/730

PR #7518, opened 2017-10-26
Feature disable current user from modifying their own groups

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/7304
Fixes https://github.com/silverstripe/silverstripe-cms/issues/730

### DIFF
--- a/src/Forms/GridField/GridFieldGroupDeleteAction.php
+++ b/src/Forms/GridField/GridFieldGroupDeleteAction.php
@@ -12,14 +12,11 @@ use SilverStripe\Security\Security;
  * Adds a delete action for the gridfield to remove a relationship from group.
  * This is a special case where it captures whether the current user is the record being removed and
  * prevents removal from happening.
- *
- * Class GroupGridFieldDeleteAction
- * @package SilverStripe\Security
  */
 class GridFieldGroupDeleteAction extends GridFieldDeleteAction
 {
     /**
-     * @var integer
+     * @var int
      */
     protected $groupID;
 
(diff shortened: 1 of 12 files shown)


tractorcow:
There was a commit (never merged) that did something similar.

https://github.com/silverstripe/silverstripe-framework/pull/2782/commits/eaf966523713be8e558a41768e9039e482075774

Except that it was a warning, not an error, and it only applied to admins.

There are times it is necessary for a user to remove their own group, so I'm not sure this feature is strictly correct. Also it should probably only apply to admin users / groups.

tractorcow:
My other concern is the switch from a combo box to a full gridfield. It makes the UX more complicated to use.

Reassigning back to @chillu to provide some product owner oversight.

src/Forms/GridField/GridFieldGroupDeleteAction.php
```diff
@@ -0,0 +1,79 @@
...
+        $currentUser = Security::getCurrentUser();
+
+        if ($actionName == 'unlinkrelation' && (
+            $this->userID === $currentUser->ID || (
+                $gridField->getList()->first() instanceof Member &&
+                $currentUser->ID === $arguments['RecordID']
+            )
+        )) {
+            throw new ValidationException(
+                _t(self::class . '.DeleteSelfFailure', 'Cannot remove yourself from groups')
```
dhensby: my understanding is the text collector cant handle `::class` and we have to use `__CLASS__` or strings

flamerohr: yeah true, I forgot about `::class` not being supported

I'll include that in my rework

flamerohr:
Have updated pull request and added behat tests

tractorcow:
Pushed up some fixes and code cleanup.

Issue #7492, opened 2017-10-18
Director::forceSSL does not force SSL when using 


wilr:
Might just be something to note in the `.env` documentation but this is something I caught in production today:

Defined a base URL in `.env`

`SS_BASE_URL="https://www.site.co.nz/"`

In `app/_config.php`: `Director::forceSSL();`.

Result:

Accessing http://www.site.co.nz incorrectly works, no redirection takes place as the check in Director::is_https() fails with the provided SS_BASE_URL.

Fix is to use the non-ssl version in the `.env` file but I think that's a simple issue for users to trip up on and could result in users being able to access pages over HTTP.

PR:

 - [ ] https://github.com/silverstripe/silverstripe-framework/pull/7520
 - [ ] https://github.com/silverstripe/silverstripe-cms/pull/2007

dhensby:
My view is we stop recommending using forceSSL - instead use a middleware to force SSL or serverside config.

However, the problem is `Director::is_https()` check needs changing/ignoring in the `forceSSL` method as it's looking up whether the base URL is https rather than the current request.

dhensby:
OK - so I've looked into this - calling `forceSSL()` from the `_config.php` is too early in the bootstrap process because the current `HTTPRequest` object hasn't been built yet (or isn't available for some reason).

This means that you can't make changes based on request state at this point in SS4.

So the solution is either create an SSL middleware or add this to a controller init method.

See https://github.com/silverstripe/silverstripe-framework/pull/7215#issuecomment-318360678 for some more info

dhensby:
both forceSSL and forceWWW should be moved to middlewares or just fully deprecated in favour of server side configuration IMO

dhensby:
I'm going to close this as "not a bug" if we have docs that recommend putting the `forceWWW` or `forceSSL` in the `_config.php` files then we need to amend those.

wilr:
Well my 2c should be we either remove the function because it doesn't do what it says on the tin, or b we keep the function, detect that HTTPRequest hasn't been built and throw an exception as currently it's a change from 3.6. Forum is littered with use of `force_ssl`. I'm happy with either approach but we may not be able to remove for 4.0

dhensby:
@silverstripe/core-team thoughts please?

tractorcow:
I am in favour of throwing exception to prevent silent failures.

sminnee:
Not keen to break APIs just now.

[...]

PR #7520, opened 2017-10-26
API / BUG - Introduce new request resolver middleware and fix broken forceWWW / forceSSL

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/7492

Requires https://github.com/silverstripe/silverstripe-cms/pull/2007 to fix regressions in Director::makeRelative

### DIFF
--- a/src/Control/Middleware/CanonicalURLMiddleware.php
+++ b/src/Control/Middleware/CanonicalURLMiddleware.php
@@ -1,0 +1,331 @@
+<?php
+
+namespace SilverStripe\Control\Middleware;
+
+use SilverStripe\Control\Controller;
+use SilverStripe\Control\Director;
+use SilverStripe\Control\HTTP;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
+use SilverStripe\Control\HTTPResponse_Exception;
+use SilverStripe\Core\CoreKernel;
+use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\Core\Injector\Injector;
+
+/**
+ * Allows events to be registered and passed through middleware.
+ * Useful for event registered prior to the beginning of a middleware chain.
+ */
+class CanonicalURLMiddleware implements HTTPMiddleware
+{
+    use Injectable;
+
+    /**
+     * Set if we should redirect to WWW
+     *
+     * @var bool
+     */
+    protected $forceWWW = false;
+
+    /**
+     * Set if we should force SSL
+     *
+     * @var bool
+     */
+    protected $forceSSL = false;
+
+    /**
+     * Redirect type
+     *
+     * @var int
+     */
+    protected $redirectType = 301;
+
+    /**
+     * Environment variables this middleware is enabled in, or a fixed boolean flag to
+     * apply to all environments
+     *
+     * @var array|bool
+     */
+    protected $enabledEnvs = [
+        CoreKernel::LIVE
+    ];
+
+    /**
+     * If forceSSL is enabled, this is the list of patterns that the url must match (at least one)
+     *
+     * @var array Array of regexps to match against relative url
+     */
+    protected $forceSSLPatterns = [];
+
+    /**
+     * SSL Domain to use
+     *
+     * @var string
+     */
+    protected $forceSSLDomain = null;
+
+    /**
+     * @return array
+     */
+    public function getForceSSLPatterns()
+    {
+        return $this->forceSSLPatterns ?: [];
+    }
+
+    /**
+     * @param array $forceSSLPatterns
+     * @return $this
+     */
+    public function setForceSSLPatterns($forceSSLPatterns)
+    {
+        $this->forceSSLPatterns = $forceSSLPatterns;
+        return $this;
+    }
+
+    /**
+     * @return string
+     */
+    public function getForceSSLDomain()
+    {
+        return $this->forceSSLDomain;
+    }
+
+    /**
+     * @param string $forceSSLDomain
+     * @return $this
+     */
+    public function setForceSSLDomain($forceSSLDomain)
+    {
+        $this->forceSSLDomain = $forceSSLDomain;
+        return $this;
+    }
+
+    /**
+     * @return bool
+     */
+    public function getForceWWW()
+    {
+        return $this->forceWWW;
+    }
+
+    /**
+     * @param bool $forceWWW
+     * @return $this
+     */
+    public function setForceWWW($forceWWW)
+    {
+        $this->forceWWW = $forceWWW;
+        return $this;
+    }
+
+    /**
+     * @return bool
+     */
+    public function getForceSSL()
+    {
+        return $this->forceSSL;
+    }
+
+    /**
+     * @param bool $forceSSL
+     * @return $this
+     */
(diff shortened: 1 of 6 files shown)


src/Control/Middleware/RegisteredEventsMiddleware.php
```diff
@@ -0,0 +1,80 @@
...
+use InvalidArgumentException;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
+use SilverStripe\Core\Injector\Injectable;
+
+/**
+ * Allows events to be registered and passed through middleware.
+ * Useful for event registered prior to the beginning of a middleware chain.
+ */
+class RegisteredEventsMiddleware implements HTTPMiddleware
```
sminnee: I would have expected a DomainCanoncialisationMiddleware instead of this more abstract one?

sminnee: And then a `DomainCanoncialisationMiddleware.force_www` config option or something...

sminnee: The thing that weirds me out about this a bit is that a "registered event" and a "middleware" are two different abstractions that do very different things, in addition to the legacy "request filter" abstraction.

sminnee: The alternative approach would be to turn the forceWWW logic into its own middleware, and then add that middleware to Director instead of registering an event.

tractorcow:
@chillu do you have any thoughts on this pr?

tractorcow:
I think if we were to move to a less-generic middleware, I'd probably just have a separate WWW and SSL redirect middleware, rather than trying to find a middle ground. What do you think about this @sminnee ?

If we were to cater to some yet-unknown "third" redirect type, it would be hard to pre-code this into such a middleware, thus the over-generalisation of the initial solution.

chillu:
I agree with Sam, this adds unnecessary complexity. In my opinion, the benefits of middleware and the SS4 bootstrap is the relative simplicity and predictability of the logic flow (request in, request out). Modifying this behaviour with procedural logic (like calling `Director::force_ssl()` from `_config.php`) should be considered legacy behaviour, rather than encouraged through some kind of middleware events system. If you need influence over the request that early, you can add custom middleware in your `index.php`.

Laravel has three ways you can do this: `App::before()` (which looks like inlined middleware), global route filtering, or route specific "before" settings. If you need SSL redirection in your logic flow, you can use `Redirect::secure()` directly and return a HTTPresponse. This makes a lot of sense to me. https://stackoverflow.com/questions/19967788/laravel-redirect-all-requests-to-https

So, as a short term fix for keeping support for procedural invocation of `Director::force_ssl()`, I'd follow Sam's lead and suggest to set a config value in a domain canonicalisation middleware singleton. That's assuming injector is already booted at the time we parse/exec `_config.php`.

tractorcow:
Ok, let's go with the DomainCanonicalisationMiddleware

sminnee:
In terms of API for the middleware, I would suggest:

 * Force WWW can be set to "always" (maybe a `"*"` or `true`) or "if the hostname is in the following list"
 * Force SSL can be set to "always" (maybe a `"*"` or `true`) or "if the hostname is in the following list"

Beyond the requirements of the issue at hand I would also suggest:

 * A canonical hostname to redirect to
 * A list of other allowed hostnames that won't redirect to the canonical hostname

In terms of processing order I would do:
 * Auto-add www 
 * Check canonical hostname
 * Auto-add HTTPS

Outside the scope of this, but for subsites I'd probably look at a whole separate implementation of this.

chillu:
I'd prefer to limit the implementation (at least for 4.0) to something that allows use of the `Director::forceSSL()` API, which means implementing the `$patterns` argument (for regex path checks), and the `$secureDomain` argument. If domain whitelists and maps are little implementation overhead, go ahead. But it's creating an API where I'm not sure we need one.

Overall I agree with Dan that this stuff is better handled by the web server layer (`.htaccess` or nginx rules). `forceSSL()` made sense at a time when it was considered a special case (e.g. ensuring a form displayed on SSL for a secure submission). Best practice across the web now is to auto-redirect to SSL for every request, which makes those initial non-SSL requests to the homepage slower if you're doing them in a PHP layer. And conditional logic (only some paths, only some domains) is becoming less common over time as SSL adoption grows.

tractorcow:
Director::forceSSL() would still need to work after middleware has been initialised... e.g. in Controller::init(), so we would need it still, even if we implement it at the middleware level. My intention is to retain the "if request is available" condition to determine which is activated.

tractorcow:
Updated and pushed... another rather major refactor, so look at it with a fresh mind from the last solution. :)

src/Control/Middleware/CanonicalURLMiddleware.php
```diff
@@ -0,0 +1,347 @@
...
+     */
+    protected function getRedirectForceSSL(HTTPRequest $request)
+    {
+        // Check if force SSL is enabled
+        if (!$this->getForceSSL()) {
+            return null;
+        }
+
+        // Already on SSL
+        if (Director::is_https($request)) {
```
dhensby: As reported in #7492 this will provide a false positive if the `SS_BASE_URL` is set to an https scheme even if the request isn't. I recommend checking against the request only.

src/Control/Middleware/CanonicalURLMiddleware.php
```diff
@@ -0,0 +1,347 @@
...
+    protected function getRedirectForceWWW(HTTPRequest $request)
+    {
+        // Check if force www is enabled
+        if (!$this->getForceWWW()) {
+            return null;
+        }
+
+        // Check if already on www
+        $host = Director::host($request);
+        if (strpos($host, 'www') === 0) {
```
dhensby: should we check for `www.`?

tractorcow: Ah true, will update.

src/Control/Middleware/CanonicalURLMiddleware.php
tractorcow: It's nothing to do with SS_BASE_URL; `Director::is_https()` prefers the Request object before checking SS_BASE_URL.

When this line is called in CanonicalURLMiddleware, the $request will always be populated (since this middleware only activates when a request is available).

tractorcow: I think in that case, it shouldn't be functionally incorrect to do as you suggest, and just inline the small part of Director::is_https(). I think I see your point. :)

tractorcow: Ok, this identified a painful flaw in Director::makeRelative(). For sites with multiple domains this method wouldn't work, meaning that Director::mockRequest() was failing in tests.

After a bit of work I've fixed and tested this, however. Now we can directly use the request instead of needing Director for getting request components.

tractorcow:
All updated. :)

tractorcow:
Fixed up some regressions in Director::makeRelative(). I've added a note that user code should check is_site_url() BEFORE calling makeRelative(), if using un-trusted urls.

Added https://github.com/silverstripe/silverstripe-cms/pull/2007

PR #7526, opened 2017-10-30
Traversing searchcontext fix

andreaspiening:
I am using ModelAdmin to manage a nested model (A many many B many_many C). When I set up $searchable_fields in A to search A for a property in C (A.B.C.Name) I get an Error saying:

    [Emergency] Uncaught BadMethodCallException: Object->__call(): the method 'relation' does not exist on 'SilverStripe\ORM\UnsavedRelationList'

because DataObject::searchableFields() is traversing into the relation, trying to find a default_search_filter_class for C.Name to be merged with the specific filter spec, specified in the config or like in my case private static.

This can be worked around by not defining the filter in the config but by overloading DataObject::searchableFields() and adding the filter there.

There are various ways to fix the issue and my favourite is to return null on unresolvable traversal paths in DataObject::relObject(), but in my request I went with catching the exception, because it's least prone to cause side effects.

Here is my setup:

```
class A extends SilverStripe\ORM\DataObject
{
	private static $db = [
		'Name' => 'Varchar',
	];

	private static $many_many = [
		'Bs' => B::class,
	];

	private static $searchable_fields = [
		'Name' => [
			'filter' => 'PartialMatchFilter',
			'field' => SilverStripe\Forms\TextField::class,
			'title' => 'Name',
		],
		'Bs.Cs.Name' => [
			'filter' => 'PartialMatchFilter',
			'field' => SilverStripe\Forms\TextField::class,
			'title' => 'Name of Bs of Cs',
		],
	];
}

class B extends SilverStripe\ORM\DataObject
{
	private static $db = [
		'Name' => 'Varchar',
	];

	private static $belongs_many_many = [
		'As' => A::class,
	];

	private static $many_many = [
		'Cs' => C::class,
	];
}

class C extends SilverStripe\ORM\DataObject
{
	private static $db = [
		'Name' => 'Varchar',
	];

	private static $belongs_many_many = [
		'Bs' => B::class,
	];
}
```

The other issue with traversing SearchContexts is that they are not showing in the search summary: E.g. if you search for A.B.C.Name in the ModelAdmin, the parameterName = searchValue pair doesn't show at the to of the GridField to indicate that you are looking at a filtered list, consequently the "clear filter" button is missing also.

This is due to the fact that SearchContext::getSummary() is failing to find a filter A__B__C__Name (because it's called A.B.C.Name) and later failing to find a field called q[A.B.C.Name] (because it is called q[A__B__C__Name]).

I am not entirely sure how to best implement the translation between field and filter names. In order to demonstrate the issue the pull request shows an ad hoc translation.

Sorry for basing this off of 4.0.0-rc1. That's all I had for testing.

Thanks for the many cool new features in ss4 which allowed me to pull of a project, that would have been impossible in the previous version!

### DIFF
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -3215,10 +3215,14 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
                 //   'field' => 'NumericField', // optional
                 //   'title' => 'My Title', // optional
                 // ))
-                $rewrite[$identifer] = array_merge(
-                    array('filter' => $this->relObject($identifer)->config()->get('default_search_filter_class')),
-                    (array)$specOrName
-                );
+                try {
+                    $rewrite[$identifer] = array_merge(
+                        array('filter' => $this->relObject($identifer)->config()->get('default_search_filter_class')),
+                        (array)$specOrName
+                    );
+                } catch (Exception $exception) {
+                    $rewrite[$identifer] = (array)$specOrName;
+                }
             } else {
                 // Format: array('MyFieldName' => 'ExactMatchFilter')
                 $rewrite[$identifer] = array(
--- a/src/ORM/Search/SearchContext.php
+++ b/src/ORM/Search/SearchContext.php
@@ -356,12 +356,12 @@ class SearchContext
             if (empty($searchValue)) {
                 continue;
             }
-            $filter = $this->getFilter($searchField);
+            $filter = $this->getFilter(str_replace('__', '.', $searchField));
             if (!$filter) {
                 continue;
             }
 
-            $field = $this->fields->fieldByName($filter->getFullName());
+            $field = $this->fields->fieldByName(str_replace('.', '__', $filter->getFullName()));
             if (!$field) {
                 continue;
             }
--- a/tests/php/ORM/Search/TraversingSearchContextTest.php
+++ b/tests/php/ORM/Search/TraversingSearchContextTest.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace SilverStripe\ORM\Tests\Search;
+
+use SilverStripe\Dev\SapphireTest;
+
+class TraversingSearchContextTest extends SapphireTest
+{
+
+    protected static $fixture_file = 'TraversingSearchContextTest.yml';
+
+    protected static $extra_dataobjects = array(
+        TraversingSearchContextTest\Student::class,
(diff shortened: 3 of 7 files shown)


dhensby:
This for this fix - it looks good but adding a test to show that this solves a problem would be the cherry on top. Do you think you'd be able to write one?

Glad to hear that SS4 is doing well for you.

andreaspiening:
That test should cover the fix. What do you think?

tractorcow:
> There are various ways to fix the issue and my favourite is to return null on unresolvable traversal paths in DataObject::relObject(), but in my request I went with catching the exception, because it's least prone to cause side effects.

The problem with exceptions is that you are now catching errors which may be legitimate problems, but the issue is smothered and replaced with something the developer may never (but perhaps should) see.

May I suggest instead a fix in DataObject::relObject() so that it behaves like relField(); This will type-check the object before calling ->relation() on it.

E.g.

```
if (ClassInfo::methodExists($component, $relation)) {
	$component = $component->relation();
} elseif ($component instanceof DataList) {
	$component = $component->relation($relation);
} else {
	// some other logic, I guess? Maybe $component->dbObject($relation).
	return null;
}
```

You could also define relation() method on UnsavedRelationList. :)

tractorcow:
Related: https://github.com/silverstripe/silverstripe-framework/pull/6211

andreaspiening:
Absolutely. As I said, catching the exception was not my favourite fix. I just didn't feel comfortable to request a more invasive change so close to the release, congrats by the way. I'll have a crack, just not today.

tractorcow:
I got busy and had a cleanup of relObject / relField methods.

https://github.com/silverstripe/silverstripe-framework/pull/7583/files#diff-5bda7c74e3842ae1fa276619f19d7babR3205

andreaspiening:
On the first glance that change of yours should fix the searchbleFields() part of my issue. How do we proceed? Do you want me to remove it from my request and merge the rest?

tractorcow:
What I haven't done yet is update the tests; You wrote some really good tests that I haven't covered in my change. Do you think we need to rewrite these tests, or are they not necessary?

dhensby:
bump - let's not let this fall away :)

tractorcow:
Let's just merge https://github.com/silverstripe/silverstripe-framework/pull/7583 and we can update it with another PR if it goes haywire. :)

PR #7539, opened 2017-10-31
Fixed array/object mismatch bug in PaginatedList

colintucker:
The last commit to `PaginatedList` throws errors because `$this->request` can sometimes be an array and not an `HTTPRequest` object.  This fix checks if `$this->request` is an instance of `HTTPRequest` before calling `getURL()`.

Also, the `&` separator parameter passed to `HTTP::setGetVar()` has been removed, as that method already uses an encoded `&amp;` to build the query string.  This would result in pagination links with some encoded ampersands, and some unencoded ampersands.

The `PaginatedListTest` class was also updated to correctly parse the encoded links generated by `PaginatedList`.

### DIFF
--- a/src/Control/HTTP.php
+++ b/src/Control/HTTP.php
@@ -167,7 +167,7 @@ class HTTP
      * @param string $separator Separator for http_build_query().
      * @return string
      */
-    public static function setGetVar($varname, $varvalue, $currentURL = null, $separator = '&amp;')
+    public static function setGetVar($varname, $varvalue, $currentURL = null, $separator = '&')
     {
         if (!isset($currentURL)) {
             $request = Controller::curr()->getRequest();
--- a/src/ORM/PaginatedList.php
+++ b/src/ORM/PaginatedList.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\ORM;
 
 use SilverStripe\Control\HTTP;
+use SilverStripe\Control\HTTPRequest;
 use SilverStripe\ORM\Queries\SQLSelect;
 use SilverStripe\View\ArrayData;
 use ArrayAccess;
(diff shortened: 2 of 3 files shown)


tractorcow:
request should ONLY be a HTTPRequest. When is it assigned a non-request?

tractorcow:
I've answered my own question... the constructor.

colintucker:
@tractorcow it's actually specified as `ArrayAccess` [in the source](https://github.com/silverstripe/silverstripe-framework/blob/d876e36424918909b65c0c53d4d682d50aadcf7b/src/ORM/PaginatedList.php#L26-L43), and indeed the constructor default is an `array`. Perhaps this is legacy from when `PaginatedList` was passed `$_GET`?

tractorcow:
Requested changes temporarily; We need to look at the cause of the poor template behaviour before we merge this, as I have a feeling this is highly regression prone.

tractorcow:
Actually it may be 4.0.1 before we patch this.

tractorcow:
Ok, so this is the code I'm using to test.

https://gist.github.com/tractorcow/4b2f624b37c4eb515a39c5e11142ef41

I'm getting double-escaped `&amp;`s in my html output.

I think the default needs to be `&` not `&amp;`. Can you update to fix?

tractorcow:
```html
                <div class="pagination">
                    
                        <a class="prev" href="home?start=60&amp;queryarg=stehou">Prev</a>
                    
                    
                        
                            
                                <a href="home?start=0&amp;amp;queryarg=stehou">1</a>
                            
                        
                    
                        
                            
                                ...
                            
                        
                    
                        
                            
                                <a href="home?start=50&amp;amp;queryarg=stehou">6</a>
                            
                        
                    
                        
                            
                                <a href="home?start=60&amp;amp;queryarg=stehou">7</a>
                            
                        
                    
                        
                            8
                        
                    
                        
                            
                                <a href="home?start=80&amp;amp;queryarg=stehou">9</a>
                            
                        
                    
                    
                        <a class="next" href="home?start=80&amp;queryarg=stehou">Next</a>
                    
                </div>

```

colintucker:
Yep so it seems I had already a workaround in my standard `Pagination` include for this, by using `.RAW` for the links. I think then @tractorcow we need instead to change `HTTP::setGetVar()` to have only `&` separators and not `&amp;`.  All escaping should be handled by casting in the template layer.

colintucker:
Oops, forgot about `HTTPTest`, just made the changes and pushed.

dhensby:
@tractorcow I'll leave you to merge this as I'm not sure what the 4.0 branch is for now (ie: 4.0.1 or 4.0.0) and not sure whether you want this is 4.0.0 or 4.0.1

tractorcow:
It's for 4.0.0 for now; Yes please hold off until we tag stable (tomorrow).

PR #7544, opened 2017-11-01
BUG Remove usage of deprecated each() and use a helper method instead

tractorcow:
Prevents deprecation errors when running PHP7.

### DIFF
--- a/src/ORM/Hierarchy/MarkedSet.php
+++ b/src/ORM/Hierarchy/MarkedSet.php
@@ -459,18 +459,38 @@ class MarkedSet
 
         // Build markedNodes for this subtree until we reach the threshold
         // foreach can't handle an ever-growing $nodes list
-        while (list(, $node) = each($this->markedNodes)) {
-            $children = $this->markChildren($node);
-            if ($nodeCountThreshold && sizeof($this->markedNodes) > $nodeCountThreshold) {
-                // Undo marking children as opened since they're lazy loaded
-                /** @var DataObject|Hierarchy $child */
-                foreach ($children as $child) {
-                    $this->markClosed($child);
+        $done = [];
+        while ($nodes = $this->iterateNewItems($this->markedNodes, $done)) {
+            foreach ($nodes as $node) {
+                $children = $this->markChildren($node);
+                if ($nodeCountThreshold && sizeof($this->markedNodes) > $nodeCountThreshold) {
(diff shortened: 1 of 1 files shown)


src/ORM/Hierarchy/MarkedSet.php
```diff
@@ -459,21 +459,41 @@ public function markPartialTree()
...
+     * @param array $list Global list of items which must be iterated
+     * @param array $iterated Keys of already iterated items to skip, which is updated by calling this method
+     * @return array List of items yet to iterate
+     */
+    protected function iterateNewItems(array $list, array &$iterated)
+    {
+        $done = array_combine($iterated, $iterated);
+        $items = array_diff_key($list, $done);
+        // Mark these keys to prevent them being re-iterated again
+        $iterated = array_merge($iterated, array_keys($items));
```
flamerohr: it seems like you are doing `array_combine()` to make sure the items are unique rather than use `array_unique()`?

src/ORM/Hierarchy/MarkedSet.php
```diff
@@ -459,21 +459,41 @@ public function markPartialTree()
...
-            if ($nodeCountThreshold && sizeof($this->markedNodes) > $nodeCountThreshold) {
-                // Undo marking children as opened since they're lazy loaded
-                /** @var DataObject|Hierarchy $child */
-                foreach ($children as $child) {
-                    $this->markClosed($child);
+        $done = [];
+        while ($nodes = $this->iterateNewItems($this->markedNodes, $done)) {
+            foreach ($nodes as $node) {
+                $children = $this->markChildren($node);
+                if ($nodeCountThreshold && sizeof($this->markedNodes) > $nodeCountThreshold) {
```
flamerohr: Would this `sizeof()` cause warnings as well?

tractorcow: No because $this->markedNodes isn't a callable.

src/ORM/Hierarchy/MarkedSet.php
tractorcow: It's not to ensure uniqueness, it's to make the result work with the subsequent array_diff_key

tractorcow:
Don't merge, I'm going to move into ArrayLib and replace other each() calls.

PR #7545, opened 2017-11-01
BUG Remove usage of deprecated each() and use a helper method instead

tractorcow:
Replaces https://github.com/silverstripe/silverstripe-framework/pull/7544

There is still one reference to the `each()` method, but it's in a thirdparty folder library that I don't want to touch.

CMS PR at https://github.com/silverstripe/silverstripe-cms/pull/2010 (doesn't need to be merged at the same time)

### DIFF
--- a/src/Forms/GridField/GridField.php
+++ b/src/Forms/GridField/GridField.php
@@ -106,6 +106,11 @@ class GridField extends FormField
     protected $name = '';
 
     /**
+     * Pattern used for looking up
+     */
+    const FRAGMENT_REGEX = '/\$DefineFragment\(([a-z0-9\-_]+)\)/i';
+
+    /**
      * @param string $name
      * @param string $title
      * @param SS_List $dataList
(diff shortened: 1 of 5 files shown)


tractorcow:
@chillu should this be 4.0.0 or 4.0.1?

src/Forms/GridField/GridField.php
```diff
@@ -362,11 +367,19 @@ public function FieldHolder($properties = array())
...
             'after' => true,
         );
+        $fragmentDeferred = [];
 
-        reset($content);
+        // TODO: Break the below into separate reducer methods
 
-        while (list($contentKey, $contentValue) = each($content)) {
-            if (preg_match_all('/\$DefineFragment\(([a-z0-9\-_]+)\)/i', $contentValue, $matches)) {
+        // Continue looping if any placeholders exist
```
tractorcow: This block was a lot of no-fun to fix. =( Unlike other each() usages, it requires re-iteration of a single key multiple times.

tractorcow:
Note: Cost of not implementing this is errors in php 7.2 :D

chillu:
7.2 is in RC mode, and scheduled for end of Nov - let’s do this in 4.0.1.

sminnee:
> There is still one reference to the each() method, but it's in a thirdparty folder library that I don't want to touch.

I think you need to fix the thirdparty code as well, if we're still using it. Otherwise this PR won't actually solve the issue.

tractorcow:
Since it's 4.0.1 I'll look at it later. :)

src/Forms/GridField/GridField.php
flamerohr: would the `array_filter()` pick up on the newly added items, would it be worth using `iterateVolatile` in the top `while` loop?

flamerohr: Lets leave this for another time, there's a big top up top and tests are green

PR #7548, opened 2017-11-01
BUG Use parse_str in place of guzzle library

### LINKED ISSUE #7540: guzzlehttp/psr7 should be explicitly required in framework
It’s currently only used in one place: https://github.com/silverstripe/silverstripe-framework/blob/4a94dfe55bf579649d5185a8d3f4139956fb9c5a/src/Core/Startup/ParameterConfirmationToken.php#L5

It works because `silverstripe/framework` requires `silverstripe/assets`, which requires `invervention/image`, which requires `guzzlehttp/psr7`. That’s a pretty delicate chain of dependencies, so we should either explicitly require `psr7` or refactor that use of it.

Marked as high impact as it’s potentially breaking.

tractorcow:
Fixes #7540

### DIFF
--- a/src/Core/Startup/ParameterConfirmationToken.php
+++ b/src/Core/Startup/ParameterConfirmationToken.php
@@ -2,7 +2,6 @@
 
 namespace SilverStripe\Core\Startup;
 
-use function GuzzleHttp\Psr7\parse_query;
 use SilverStripe\Control\Controller;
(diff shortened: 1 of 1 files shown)


tractorcow:
Well, this is deprecated in 7.2.... we'll need a new fix anyway soon. =(
	
	http://php.net/manual/en/function.parse-str.php

PR #6829, opened 2017-04-22
[WIP] Refactor Authenticators

sminnee:
I've been working with @Firesphere on a pretty substantial refactor of the authenticator classes. This is the work in progress. It would be good to get feedback on the approach before it's finalised.

Main changes:

 * Authenticator is now an interface, not an abstract base class, and the static methods have become instance methods.
 * The authenticator provides extension points for lost password & change password as well as log-in, so that these systems can be authenticator-specific also
 * Instead of just having a Form as the extension point, we have a RequestHandler. This means that authenticators can provide multiple forms for a multi-step process, action handlers for OAuth, etc. @chillu @Firesphere and I discussed "what's the best abstraction for an arbitrary multi-step workflow?" and came to the conclusion that RequestHandler gives us most of what we need rather than needing some overblown new abstraction.
   * The handlers can return array key/values to substitute content into the default controller provided by Security. This requires a bit of extra code in Security but follows patterns already present in RequestHandler.
 * Authenticators can indicate the services that they support via the `supportedServices` method.

## Tasks

- [x] Make supportedServices() actually do something
- [x] Finish refactoring changepassword
- [x] Get unit tests working
- [x] Complete manual testing
- [x] Determine best way of providing addBackURLParam to the handlers - right now it's copypasta
- [x] Review and potentially remove the LoginForm base. -- I would prefer to keep it
- [x] Review and potentially move getAuthenticatorName() back into the authenticator -- I found it better fitting on the Form, due to the nature of the template rendering
- [x] Get support for multiple authenticators working again
- [x] Try upgrading @Firesphere's MFA module to validate that this API is useful
- [x] Try upgrading @dhensby's MFA module to do the same as above
- [ ] Try upgrading a facebook login module, for the same reason -- Seems to work, but it doesn't accept `localhost:8080` as a valid domain

### Verifying status of these tasks

- [x] Determine best way of handling messages from `authenticate()` in `LoginHandler::performLogin()`
- [ ] Potentially add configuration to the login handler so that the writing together of steps can be more easily modified on subclasses (e.g. to subclass `MemberAuthenticator\LoginHandler` and then wire in a 2nd step)

## Pull requests

- [ ] https://github.com/silverstripe/silverstripe-behat-extension/pull/159
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/6829
- [ ] https://github.com/silverstripe/silverstripe-admin/pull/98
- [ ] https://github.com/silverstripe/silverstripe-versioned/pull/15
[...]

Issue #6320, opened 2016-11-17
Does Member::logOut() need to destroy the entire session?


kinglozzer:
Is destroying the entire session actually necessary? I’ve found it an annoyance in the past, where things which aren’t dependent on the current member are stored in the session are wiped out if they log out. https://github.com/silverstripe/silverstripe-framework/blob/7a10c194bdd751f74669c07836befecc1fde400e/src/Security/Member.php#L609

tractorcow:
I can see why, although I disagree with it. It's acting this way in case of user-specific information present in the session.

dhensby:
Yes, it needs to be destroyed, we can't know what is in the users session and what potential information could be leaked if it was retained.

However, perhaps we could move it to after the `memberLoggedOut` hook to allow devs to inspect the session before it's destroyed.

dhensby:
see https://www.owasp.org/index.php/Session_Management_Cheat_Sheet#Manual_Session_Expiration

When users logout their session should be destroyed.

tractorcow:
That's as closed as we need I guess!

PR #7552, opened 2017-11-02
FIX Sessions must be destroyed on logout

dhensby:
Sessions must be destroyed on log out. Currently they are retained and just the "logged in user" identifier is removed.

See: https://www.owasp.org/index.php/Session_Management_Cheat_Sheet#Logout_Button

This is a regression from the new auth work (#6829)

### DIFF
--- a/src/Control/HTTPRequestBuilder.php
+++ b/src/Control/HTTPRequestBuilder.php
@@ -99,6 +99,23 @@ class HTTPRequestBuilder
             $headers['Content-Length'] = $server['CONTENT_LENGTH'];
         }
 
+        // Ensure basic auth is available via headers
+        if (isset($server['PHP_AUTH_USER']) && isset($server['PHP_AUTH_PW'])) {
+            // Shift PHP_AUTH_* into headers so they are available via request
+            $headers['PHP_AUTH_USER'] = $server['PHP_AUTH_USER'];
+            $headers['PHP_AUTH_PW'] = $server['PHP_AUTH_PW'];
+        } elseif (!empty($headers['Authorization']) && preg_match('/Basic\s+(.*)$/i', $headers['Authorization'], $matches)) {
+            // Enable HTTP Basic authentication workaround for PHP running in CGI mode with Apache
+            // Depending on server configuration the auth header may be in HTTP_AUTHORIZATION or
+            // REDIRECT_HTTP_AUTHORIZATION
+            //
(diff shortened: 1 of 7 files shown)


kinglozzer:
See also https://github.com/silverstripe/silverstripe-framework/issues/6320. I’m not sure it should be up to `SessionAuthenticationHandler` to decide that _everything_ in the session _must_ be destroyed with no way of opting out of that behaviour - the session can be used as a store for non-sensitive data.

It could be argued that if you add something sensitive to the session manually then you also need to make sure it’s removed when appropriate, rather than assuming that it will disappear when the user logs out. Of course we need to weigh that against the “secure by default” argument, which is about as heavy an argument as it’s possible to get.

I’m leaning towards merging but keen to hear what others think.

dhensby:
I think nothing has changed regarding the comments in #6320

If you need to retain some information after hitting logout (you shouldn't - logging out is analogous to starting a brand new session), then you can preserve this through an extension by making use of onbefore/after logout hooks.

---

@kinglozzer regarding your points, I understand your thinking but it's not quite right. It is exactly the responsibility of `SessionAuthenticationHandler` to assume the entire session is sensitive and not to be shared. It doesn't make much sense that if I log out that the session state I built up should then be passed onto the next user to login or use the site after me. Just because this data isn't sensitive doesn't mean it should be shared with others.

I don't agree that if you add something sensitive to the session you're responsible for its removal, standard practice is to destroy sessions when users logout and devs should be able to rely upon that behaviour.

dhensby:
I'm wondering here... the session should be destroyed no matter what `AuthenticationHandler` is being used... maybe there's somewhere more central we should be destroying the session on logout...

tractorcow:
Whoops, those tests are broken.

tractorcow:
@dhensby ,move it to RequestAuthenticationHandler instead; It's the parent to SessionAuthenticationHandler. That is the main entry point to "logout" actions.

tractorcow:
Oh, let's use `restart`; It destroys the session, and restarts a new one (so the application can continue).

tractorcow:
Oh look I found out that BasicAuth hasn't been working properly, due to bad test design. =(

tractorcow:
Shifted ->destry() to ->restart(), and fixed up BasicAuth at the same time (since those all regressed).

tractorcow:
Rebased

Issue #1396, opened 2013-04-03
[2011-04-11] ContextSummary handles german Umlauts (and probably other UTF-8 chars) wrong


silverstripe-issues:
created by: omarkohl
created at: 2011-04-11
original ticket: http://open.silverstripe.org/ticket/6594

---

In the Text class (sapphire/core/model/fieldtypes/Text.php) the function ContextSummary handles UTF-8 characters such as german Umlauts wrongly when those characters belong to words that get cut in half (because of the length of the context). 

For example with $characters = 10 and searching for string 'abc' in this content 'abc def ghijkl' the result is 'abc def gh?'.

My fix is changing line 292 from:

`$summary = substr($text, $position, $characters);`

to:

`$summary = utf8_encode(substr(utf8_decode($text), $position, $characters));`

tractorcow:
Easy to fix; Add a unit test to DBTextTest::providerContextSummary() which covers the above unit case.

sminnee:
Using mb_substr is probably also a good fix.

PR #7557, opened 2017-11-05
FIX: Fix ContextSummary behaviour with UTF8 chars

sminnee:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/1396

Ensure that DBTextTest has UTF8 test cases too.


sminnee:
This is pretty minor and can possibly be left until 4.0.1?

PR #7559, opened 2017-11-06
NEW Add sort columns to DB index automatically

### LINKED ISSUE #7272: default_sort fields should be added as indexes
Sorting on an indexed field is faster: https://dev.mysql.com/doc/refman/5.7/en/order-by-optimization.html#order-by-index-use. 

Therefore, I would like to suggest that, by default, database fields listed in `default_sort` should be added as indexes. 

Not sure if it would work for `default_sorts` with multiple fields, but, for example, for `SiteTree` we should definitely add an Index on the `Sort` Field.

dhensby:
fixes #7272

Automatically add `default_sort` columns to DB index

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -3405,25 +3405,23 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 		}
 
 		if ($sort && is_string($sort)) {
-		    $sort = preg_split('/,(?![^()]*+\\))/', $sort);
+			$sort = preg_split('/,(?![^()]*+\\))/', $sort);
 
 			foreach ($sort as $value) {
-				$value = trim($value);
-                if(strpos($value, ' ') !== false) {
-                    $parts = explode(' ', $value, 2);
-                    $column = $parts[0];
-				} else {
-					$column = $value;
-				}
-				if (substr($column, 0, 1) === '"') {
-					$column = substr($column, 1, strlen($column)-1);
-				}
-				if (substr($column, -1, 1) === '"') {
-					$column = substr($column, 0, -1);
-				}
-				if ($this->hasOwnTableDatabaseField($column) && !array_key_exists($column, $indexes)) {
-					$indexes[$column] = true;
-				}
+				try {
+					list ($table, $column) = $this->parseSortColumn(trim($value));
+
+					$table = trim($table, '"');
+					$column = trim($column, '"');
+
+					if ($table && strtolower($table) !== strtolower($this->class)) {
+						continue;
+					}
+
+					if ($this->hasOwnTableDatabaseField($column) && !array_key_exists($column, $indexes)) {
+						$indexes[$column] = true;
+					}
+				} catch (InvalidArgumentException $e) { }
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


dhensby:
Failures are behat suites that aren't working at the moment

tractorcow:
Can you please add some unit tests?

model/DataObject.php
```diff
@@ -3403,6 +3404,29 @@ public function databaseIndexes() {
 			}
 		}
 
+		if ($sort && is_string($sort)) {
+		    $sort = preg_split('/,(?![^()]*+\\))/', $sort);
```
tractorcow: What's the significance of `*+` here?

model/DataObject.php
```diff
@@ -3403,6 +3404,29 @@ public function databaseIndexes() {
...
+
+			foreach ($sort as $value) {
```
tractorcow: Please check out ArrayList::parseSortColumn(). This has pre-existing tests, and covers the range of various sort syntaxes a bit more robustly. For instance `$default_sort = '"SiteTree"."Order" DESC';` doesn't seem covered by this code.

Maybe we could extract that helper somewhere that index generation could access it?

tractorcow: It looks like that method could also be extended to support multi-columns specifications; There's a chance to improve both ArrayList::sort() as well as index generation. :)

model/DataObject.php
dhensby: It's copied from `SQLselect::getsortorder` so I don't know :)

model/DataObject.php
dhensby: I copied code from SQLSelect, happy to copy from somewhere else if it's more robust :)

model/DataObject.php
tractorcow: Oh! I see... yeah that code itself isn't so easy to understand either. Apologies for implying it was your issue. :)

That code exists as-is in 4.0 too.

I think the SQLSelect::getOrderBy() only needs to return the SQL fragments that'll be sent to the final query; What we want is more heavily parsed, so I guess the ArrayList code is probably closer to our needs in this case?.

Oh well, once you step in a poop you're walking around in it all day aren't you. :)

tractorcow:
Maybe it's time to deprecate string-only sort, and shift to array format like we did with indexes? That turned out really successfully didn't it?

We could shift away from "write sort as ansi-sql strings" to "field lists that the ORM will sql-ify".

dhensby:
> Maybe it's time to deprecate string-only sort, and shift to array format like we did with indexes? That turned out really successfully didn't it?
> 
> We could shift away from "write sort as ansi-sql strings" to "field lists that the ORM will sql-ify".

I'd really like that but I don't think it's realistic for 3 or 4, unfortunately.

I'll look into the other code we have for parsing sort strings and add the `"SiteTree"."Sort"` case to the tests

dhensby:
@tractorcow ok, I've copied (and stripped down) the `parseSortColumn` method from `ArrayList`. It's not really made a big difference to how much code I've had to hand roll because it doesn't handle multiple sorts nor does it strip ANSI quotes.

tractorcow:
> I've had to hand roll because it doesn't handle multiple sorts nor does it strip ANSI quotes.

Just one question before I merge... what happens if you have code like this?

```php
private static $default_sort = '"SiteTree"."Title" ASC';
```

It doesn't cause an error does it?

tests/model/DataObjectSchemaGenerationTest.php
```diff
@@ -212,3 +257,12 @@ class DataObjectSchemaGenerationTest_IndexDO extends DataObjectSchemaGenerationT
...
 	);
 }
+
+class DataObjectSchemaGenerationTest_Sorted extends DataObject implements TestOnly {
+    private static $db = array(
+        'Title' => 'Varchar',
+        'Sort' => 'Int',
+    );
+
+    private static $default_sort = "Sort";
```
tractorcow: Can you add a test using ansi sql quotes to show it doesn't break those?

dhensby:
I have test cases for ANSI quoted sort and table named sorts.

model/DataObject.php
```diff
@@ -3403,13 +3404,58 @@ public function databaseIndexes() {
...
+						continue;
+					}
+
+					if (substr($column, 0, 1) === '"') {
+						$column = substr($column, 1, strlen($column)-1);
+					}
+
+					if (substr($column, -1, 1) === '"') {
+						$column = substr($column, 0, -1);
+					}
```
kinglozzer: Would `trim($column, '"')` have the same effect as these two `substr` blocks?

dhensby: hah, that would be a touch more elegant, wouldn't it!

PR #7578, opened 2017-11-09
BUG Don't redirect in force_redirect() in CLI

tractorcow:
Replaces #4025


dhensby:
do we not want this in 3.5 as well?

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/pull/7632

PR #7579, opened 2017-11-09
Sapphiretest stateisnull

phptek:
Errors in 3rd party packages or other project code is "masked" in `SapphireTest` when running unit tests, due to those errors causing `$state` to be `null`.


phptek:
Oops, wrong base-branch.

PR #7583, opened 2017-11-09
BUG Ensure relObject() safely bails on empty objects

tractorcow:
Against 4 (4.1) branch, as this includes a bunch of API changes designed to address weaknesses in the relation API, in as much as I can get away with without breakages.

Replaces #6211,  #7526 (partly)

Marked WIP because I'm still testing and gathering feedback on whether or not the general approach is appropriate.

getComponents() / getManyManyComponents() can now take $ids as a parameter, which is a better solution than modifying properties on a singleton in order to suppress UnsavedRelationList being created.

relField() and relObject() now use the same internal loop for simplicity and consistency. E_USER_ERROR has been replaced with a catchable exception.

Two proposed methods have been added to the RelationField phpdoc, but not added directly in order to respect SemVer.

### DIFF
--- a/src/ORM/DataList.php
+++ b/src/ORM/DataList.php
@@ -1017,7 +1017,10 @@ class DataList extends ViewableData implements SS_List, Filterable, Sortable, Li
     public function relation($relationName)
     {
         $ids = $this->column('ID');
-        return singleton($this->dataClass)->$relationName()->forForeignID($ids);
+        $singleton = DataObject::singleton($this->dataClass);
+        /** @var HasManyList|ManyManyList $relation */
+        $relation = $singleton->$relationName($ids);
+        return $relation;
     }
 
     public function dbObject($fieldName)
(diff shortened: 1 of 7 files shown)


dhensby:
@tractorcow this looks good - will it be getting any more attention soon? 👼

tractorcow:
I think once 4.0.1 is done I'll be giving 4.1 more attention.

I'm not 100% happy with the new getManyManyComponents() parameter changes, but it's probably good enough.

tractorcow:
Removing WIP. I think we can merge if everyone is happy with it.

kinglozzer:
Just a heads up: the changes to the method signatures for `getComponents()` and `getManyManyComponents()` apparently caused issues in the SilverShop module which has a model that overrides one or both of those methods. Thankfully adding the extra `$id = null` param will make it compatible with both 4.0 and 4.1.

Not sure what counts as part of our public API here, and it’s a bit of an edge-case, but just thought it worth mentioning :)

dhensby:
Aren't changes to method signatures strict errors? so they wouldn't be a problem in production?

kinglozzer:
Yes that’s true, but adding extra args to a method could still be considered a BC-breaking change (e.g. if someone has overridden it and added their own extra arg(s)).

I don’t think this is worth reverting, it’s very low risk and an edge-case as I said, just something to bear in mind

tractorcow:
It was strictly a semver violation sorry.

PR #7591, opened 2017-11-13
"Discarded changes" message is no longer displayed onCreate().

mfendeksilverstripe:
Related to:

https://github.com/dnadesign/silverstripe-elemental/pull/149
https://github.com/symbiote/silverstripe-gridfieldextensions/issues/226


tractorcow:
The problem with this change is that, what if you DO modify the form for the new record, and accidentally click away? I would expect it to then notify that you have unsaved changes on your new record. After this change, your form would be lost.

dhensby:
I've noticed that you've created this pull request from one of the "core branches" rather than a branch specifically intended for this pull request. This can cause problems if we ask you to target this PR at a different branch. No action is needed at the moment, this is for your information.

dhensby:
closing due to lack of feedback

PR #7594, opened 2017-11-14
BUG Prevent ConfirmPasswordField from showing value on validation failure

### LINKED ISSUE #2496: [FORMS] BUG: ConfirmedPasswordField inserts first password into confirm field
When using `ConfirmedPasswordField`, if you submit without filling in the second confirm field, it automatically adds the value of the initial password into the confirm password field when failing validation

E.G:

New Password: XYZ
Confirm New Password: [left blank]

When the user submits a form with this data, an error is thrown and they're redirected back to the page, but the Confirm New Password has been auto filled in with the New Password's value.

tractorcow:
Fixes #2496 for 4.1

Pulling against 4 branch since it introduces new API for controlling password value visibility.

Will probably need a more-nasty fix for 4.0.1 / 3.x, although we may need to break some expected behaviour (e.g. just never set the value if they don't match). I don't think that negates the ability to fix this for 4.1 with new API.

### DIFF
--- a/src/Forms/ConfirmedPasswordField.php
+++ b/src/Forms/ConfirmedPasswordField.php
@@ -150,6 +150,9 @@ class ConfirmedPasswordField extends FormField
             )
         );
 
+        // Ensure passwords are never sent to the frontend
+        $this->setDisplaysSetValue(false);
+
         // has to be called in constructor because Field() isn't triggered upon saving the instance
         if ($showOnClick) {
             $this->children->push($this->hiddenField = new HiddenField("{$name}[_PasswordFieldVisible]"));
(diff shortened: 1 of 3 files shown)


dhensby:
Do we need to introduce an API at all? Is there ever a case where we should be sending the password back to the frontend form?

I get it's annoying when the form fails for some other validation reason and you have to reenter the value, but that's the price you pay for "security"

tractorcow:
> Do we need to introduce an API at all? Is there ever a case where we should be sending the password back to the frontend form?

Well, we are doing that, so if we REMOVE that feature, it becomes a major release change, which we don't want to do.

maxime-rainville:
I'm not quite sure what we are trying to fix here. I tried running the current code base and when I get a validation failure, my confirmed password field always comes back to me blank. Which sounds like the end goal this PR is trying to achieve. Am I missing something?

PR #7597, opened 2017-11-14
Moved security process from JIRA to Github

chillu:
JIRA isn't fully under the OSS team's control, and played up in the past (Dan couldn't move issues). Since Github has project boards now, and we're paying for private repos on github.com/silverstripe-security already anyway, there's no reason to introduce another tool (JIRA) into our workflows.

No need to move existing issues, the JIRA board hasn't been used in a while. Which leads to unclear ownership and status of security issues, and is exactly the reason for this change ;)

Tasks:

 - [x] Migrate open issues (and un-created issues on security list) to https://github.com/silverstripe-security/security-issues
 - [x] Triage all issues (type/security)
 - [ ] Update "making a core release" docs (this PR)
 - [ ] Close old jira board


chillu:
@dhensby @tractorcow @sminnee I think we should get rid of JIRA for OSS security workflow - any objections?

tractorcow:
Sounds good to me.

sminnee:
Agreed - using github issues is better for all.

dhensby:
agreed

PR #7599, opened 2017-11-14
WIP - BUG Fix don't treat zero-date as invalid

flamerohr:
Partially fixes https://github.com/silverstripe/silverstripe-cms/issues/2022

There is an issue where dates aren't saving properly as well (the Pages are created with 0000-00-00 00:00:00 LastEdited and Created value, which is incorrect as well)

There are still weird issues around numbers in Persian locale which needs a deeper investigation for, may be linked to the dates issue above. 
screen shot 2017-11-14 at 11 27 23 pm

### DIFF
--- a/src/ORM/FieldType/DBDate.php
+++ b/src/ORM/FieldType/DBDate.php
@@ -67,6 +67,9 @@ class DBDate extends DBField
         } else {
             // Convert US date -> iso, fix y2k, etc
             $value = $this->fixInputDate($value);
+            if (is_null($value)) {
+                return null;
+            }
             $source = strtotime($value); // convert string to timestamp
         }
         if ($value === false) {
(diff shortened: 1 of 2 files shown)


src/ORM/FieldType/DBDate.php
```diff
@@ -67,6 +67,9 @@ protected function parseDate($value)
         } else {
             // Convert US date -> iso, fix y2k, etc
             $value = $this->fixInputDate($value);
+            if (is_null($value)) {
```
tractorcow: Can you please add a test that `0-0-0` creates a null value?

flamerohr: I've added a couple of tests for this case :)

tractorcow:
Looks good.

tractorcow:
Behat issues fixed with https://github.com/silverstripe/silverstripe-cms/pull/2028

PR #7602, opened 2017-11-15
FIX ManyMany link table joined with LEFT JOIN

### LINKED ISSUE #7588: filterAny is ignoring records with empty field values in many_many relationship
**filterAny** is ignoring records that don't have many_many relationship, for example:

### **Database:**

Joes

ID | FirstName | Surname
-- | -- | --
1 | Joe | Bloggs
2 | Joe | Pesci
3 | Joe | DeMaggio

Cars

ID | Title
-- | --
1 | Porche
2 | Ferrari
3 | VW

Joes_Cars

ID | JoesID | CarsID
-- | -- | --
1 | 1 | 1
2 | 1 | 3
3 | 3 | 2

 

### **Filter:**
`return = Joes::get()
        	->filterAny(array(
	        	'FirstName:PartialMatch'	=>	'Joe',
	        	'Cars.Title'	=>	'Joe'
        	))`

This only returns 2 records (Bloggs and DeMaggio) and ignores Pesci due to him not having a many_many relationship with Cars specified.

I am trying to perform a "search anything" type of search that will search across all tables for a specified term. The above filter should search for "Joe" whether it appears in the **Joes** table or the **Cars** table but is unfortunately only searching the records where a relationship has been set up.

dhensby:
fixes #7588 

No tests yet because:

1. It's late
2. ~~I want to see if this has caused any regressions via travis first~~ No regressions, it seems

---

## Todo:
- [x] Add test

### DIFF
--- a/model/DataQuery.php
+++ b/model/DataQuery.php
@@ -726,7 +726,7 @@ class DataQuery {
 				list($parentClass, $componentClass, $parentField, $componentField, $relationTable) = $component;
 				$parentBaseClass = ClassInfo::baseDataClass($parentClass);
 				$componentBaseClass = ClassInfo::baseDataClass($componentClass);
-				$this->query->addInnerJoin($relationTable,
+				$this->query->addLeftJoin($relationTable,
 					"\"$relationTable\".\"$parentField\" = \"$parentBaseClass\".\"ID\"");
 				if (!$this->query->isJoinedTo($componentBaseClass)) {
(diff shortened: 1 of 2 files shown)


tractorcow:
Agree with this PR it seems ok to me. Just need to add the test. :)

4.0 seems to have the same problem / fix.

dhensby:
OK - test added. Locally the test fails and the patch fixes it. waiting for travis

PR #7605, opened 2017-11-16
BUG Fix DBEnum ignoring empty defaults

### LINKED ISSUE #7582: Enum without default is selecting a default in DropdownField that has an empty default
I have 

```php
    private static $db = [
        'FooterSection' => 'Enum("Navigate, About, Legal")'
    ];
```

and

```php
        $fields->addFieldsToTab('Root.Main', [
            DropdownField::create('FooterSection', 'Footer section', [
                'Navigate' => 'Navigate',
                'About'    => 'About',
                'Legal'    => 'Legal'
            ])->setEmptyString('Not displayed in footer')
        ], 'Metadata');
```

When i create a new page, "Navigate" is always selected, instead of "Not displayed in footer"

Adding:

```php
    private static $defaults = [
        'FooterSection' => 0
    ];
```

Fixes it for me

# PRs

 - [ ] https://github.com/silverstripe/silverstripe-framework/pull/7605
 - [x] https://github.com/silverstripe/silverstripe-postgresql/pull/78

tractorcow:
Fixes #7582

DBEnum was ignoring `null` as default option, meaning that you couldn't set an empty default.

I also discovered a potential issue with DBField::create_field() where field specifications and $object arg would silently create a mal-formed DBField. I've added a warning in this case, as a separate commit, without modifying the behaviour.

In 5.x we can make that warning an InvalidArgumentException.

### DIFF
--- a/src/ORM/Connect/PDOConnector.php
+++ b/src/ORM/Connect/PDOConnector.php
@@ -163,7 +163,7 @@ class PDOConnector extends DBConnector
         $connCollation = Config::inst()->get('SilverStripe\ORM\Connect\MySQLDatabase', 'connection_collation');
 
         // Set charset if given and not null. Can explicitly set to empty string to omit
-        if ($parameters['driver'] !== 'sqlsrv') {
+        if (!in_array($parameters['driver'], ['sqlsrv', 'pgsql'])) {
             $charset = isset($parameters['charset'])
                     ? $parameters['charset']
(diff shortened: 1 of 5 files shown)


tractorcow:
@flamerohr they were always the first item by default.

tractorcow:
PostgreSQL fix https://github.com/silverstripe/silverstripe-postgresql/pull/78

Issue #7606, opened 2017-11-16
BUG: Page fails to save when you have an SVG file in an <img> tag.


patricknelson:
I noticed that if I were to have an image tag such as the following in the content of my page, it would fail to save: `<img src="assets/Uploads/logo.svg" width="100" height="50" />` Note that I believe use of SVG as an `<img>` source like this is [entirely valid](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Adding_vector_graphics_to_the_Web) and should be expected from time to time.

It appears to be due to the fact that while this is inside an image tag, SilverStripe is assuming the resulting `File` instance returned by `File::find` is actually a `Image` instance which contains additional resize/resampling code in it. This is called in `HtmlEditorField` (where the bug resides) and gets called when the form in the CMS is submitted.

The simple fix here should be to just check that the resulting `$image instanceof Image` before attempting to call `Image`-specific methods on it 😄

PR #7607, opened 2017-11-16
FIX: Prevent crash when saving page with <img> that has an SVG source.

patricknelson:
Fix that addresses #7606. Ensure the object we're handling is actually an Image instance before calling methods specific to that class (e.g. in case of using SVG's in `` tag which may be File instances).  This works even if (or especially if) you haven't configured `Image` to [treat SVG files as images](https://stackoverflow.com/questions/28491842/treat-svg-as-an-image-in-silverstripe) (which is unrelated).


dhensby:
Looks like a bug to me that probably affects 3.5, so into 3.5 please :)

patricknelson:
I was going to say, this can be easily edited to merge into 3.5 since thankfully the file didn't look like it changed much (if at all) since even 3.3.

patricknelson:
Thanks! 👍

PR #7608, opened 2017-11-16
Fix HTTP::get_mime_type with uppercase filenames.

bummzack:
The fallback of `HTTP::get_mime_type` (that uses a lookup instead of `finfo`, see #441) doesn't ensure the extension is converted to lowercase before the lookup. A file named `Image.JPG` will return `'application/unknown'`.
This change fixes this issue.


dhensby:
This looks like a bugfix so I'm going to move it over to the 3.5 branch

bummzack:
@dhensby You know best :) It affects v4 as well though.

dhensby:
@bummzack ok, thanks

PR #7612, opened 2017-11-17
Update form template directory

mr-macedawg:
As of SS4.0, template files for forms need to be in the root template directory as the core forms folder no longer exists.


dhensby:
Hmmm - I think this is actually incorrect - they need to match your name space (if you've got one) otherwise this is correct...

Would you mind clarifying that, please?

tractorcow:
Actually, it needs to match the param passed into setTemplate, as in the example.

E.g. `setTemplate('MyCustomTextField')` means the file should be in `mysite/templates/MyCustomTextField.ss`

PR #7614, opened 2017-11-17
Update 01_Extensions.md

MichaelJJames:
If you did the has_one relationship like the docs it would result in

Uncaught InvalidArgumentException: has_one relation myClass.Image references class Image which doesn't exist


dhensby:
thanks

PR #7617, opened 2017-11-20
Make DataDifferencer injectable

chillu:
Requested by Steve Boyd


chillu:
Note that there's some known issue with diff viewing: https://github.com/silverstripe/silverstripe-cms/issues/687

Issue #6904, opened 2017-05-10
No warnings when table name exceeds allowed length


chillu:
This has been an issue in 3.x, but is made much more common in 4.x with namespaces included in table names by default - particularly in tests. MySQL caps out at 64 chars: https://dev.mysql.com/doc/refman/5.7/en/identifiers.html

At the moment you just get a non-descript database error:

> Incorrect table name 'SilverStripeAustralia_MultiValueField_Tests_MultiValueFieldTest_DataObject'' in /Users/ingo/Sites/master/framework/src/ORM/Connect/DBConnector.php:61

Make this a helpful error, mentioning the `table_name` property.

Note this was mentioned in https://github.com/silverstripe/silverstripe-framework/issues/5591

## Pull requests
- [x] https://github.com/silverstripe/silverstripe-framework/pull/7621
- [x] https://github.com/silverstripe/silverstripe-framework/pull/7620

dhensby:
Any plans to get this in before stable?

chillu:
Hmm, I think we've got higher impact tickets in rc2 already. I'd say it's something for after stable - you get an error message already, it's just not as useful as it could be

tractorcow:
You could just adjust the default scaffolder not to use the WHOLE namespace (maybe just vendor and base class) as a quick win.

flamerohr:
I've started work on this:
https://github.com/open-sausages/silverstripe-framework/tree/pulls/4.0/its-too-big-to-fit

two parts to it so far:
- Shortened automatic name from full namespace to the first token (vendor) and last token (base class)
- Added legacy full namespace check, which will rename to the shortened version if necessary.

dhensby:
> @tractorcow mentioned that this may be more suited to 4.1 as it raises some breaking compatibility concerns

If there are breaking compatibility concerns it'll have to be 5.x

tractorcow:
New table scaffolder can be 5.x then, and for 4.0.1 we can just echo a warning in the default table scaffolder. Sound ok @flamerohr ?

tractorcow:
@flamerohr there's feedback on https://github.com/silverstripe/silverstripe-framework/pull/7620 you need to fix.

PR #7620, opened 2017-11-21
BUG Added warning for auto-generated table_name

flamerohr:
As discussed https://github.com/silverstripe/silverstripe-framework/issues/6904
a warning will be added for 4.0.1

The new shorter table_name generated will be available in 5.0 (https://github.com/silverstripe/silverstripe-framework/pull/7621)

Related https://github.com/silverstripe/silverstripe-frameworktest/pull/44

### DIFF
--- a/src/ORM/DataObjectSchema.php
+++ b/src/ORM/DataObjectSchema.php
@@ -6,6 +6,7 @@ use Exception;
 use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\Core\Config\Configurable;
 use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Dev\TestOnly;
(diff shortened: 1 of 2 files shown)


src/ORM/DataObjectSchema.php
```diff
@@ -302,6 +302,10 @@ protected function buildTableName($class)
 
         // Generate default table name
         if (!$table) {
+            trigger_error(
```
tractorcow: We don't need this warning for non-namespaced classes (e.g. `Page`) so can you add a check for this first?

flamerohr: Good point, will add that in

tractorcow:
Smart fix @flamerohr !

PR #7621, opened 2017-11-21
WIP shorten auto-generated table names

flamerohr:
Fix add check for legacy long table name and rename to short version

Fixes https://github.com/silverstripe/silverstripe-framework/issues/6904 and warning added for 4.0.1 https://github.com/silverstripe/silverstripe-framework/pull/7620

### DIFF
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -3059,11 +3059,32 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
         $fields = $schema->databaseFields(static::class, false);
         $indexes = $schema->databaseIndexes(static::class, false);
         $extensions = self::database_extensions(static::class);
+        $legacyTables = $schema->getLegacyTableNames(static::class);
 
         if (empty($table)) {
             throw new LogicException(
                 "Class " . static::class . " not loaded by manifest, or no database table configured"
             );
+        }
+
+        if ($legacyTables) {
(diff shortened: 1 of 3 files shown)


src/ORM/DataObject.php
```diff
@@ -3064,13 +3064,31 @@ public function requireTable()
...
         }
 
+        if ($legacyTables) {
+            $ignore = Config::inst()->get(static::class, 'ignored_legacy_tables') ?: [];
+            $renameTables = array_diff(
+                array_intersect($legacyTables, DB::table_list()),
+                $ignore
+            );
+            if (count($renameTables) > 1) {
+                throw new LogicException(
```
tractorcow: Maybe just make this a warning and keep doing the update, otherwise users may end up with an inconsistent database.

flamerohr: hmm, I guess the issue I have is if there is more than one legacy table - which one would be the better one to use?

tractorcow: Don't do any rename; Create a brand new table in that case? It's equally likely that none of the tables are valid as is one of them is.

tractorcow:
@flamerohr I've done a bit of work on master to fix up the build. Rebasing in hope that this addresses the broken build. :)

tractorcow:
Nope, I messed up the constraints... needs to be 2.x-dev not 2.0.x-dev /sigh

tractorcow:
Ready to merge on green.

PR #7623, opened 2017-11-21
check for apc.enable_cli when running from CLI

andrewandante:
Fixes an issue when running PHPUnit via CLI where it spews out warnings about not being able to save values.

IMO the problem is actually at Symfony's end - https://github.com/symfony/symfony/pull/25080 - but they say it will break their cache warming so I've applied the fix here as well.

### DIFF
--- a/src/Core/Cache/DefaultCacheFactory.php
+++ b/src/Core/Cache/DefaultCacheFactory.php
@@ -5,6 +5,7 @@ namespace SilverStripe\Core\Cache;
 use Psr\Log\LoggerAwareInterface;
 use Psr\Log\LoggerInterface;
 use Psr\SimpleCache\CacheInterface;
+use SilverStripe\Control\Director;
 use SilverStripe\Core\Injector\Injector;
 use Symfony\Component\Cache\Simple\FilesystemCache;
 use Symfony\Component\Cache\Simple\ApcuCache;
@@ -85,7 +86,8 @@ class DefaultCacheFactory implements CacheFactory
     {
         static $apcuSupported = null;
         if (null === $apcuSupported) {
-            $apcuSupported = ApcuAdapter::isSupported();
+            // Need to check for CLI because Symfony won't: https://github.com/symfony/symfony/pull/25080
+            $apcuSupported = Director::is_cli() ? ini_get('apc.enable_cli') && ApcuAdapter::isSupported() : ApcuAdapter::isSupported();
         }
         return $apcuSupported;
     }


tractorcow:
Travis failures not your fault; This seems fine.

I don't think your change is going to make it into symfony.

src/Core/Cache/DefaultCacheFactory.php
```diff
@@ -85,7 +85,8 @@ protected function isAPCUSupported()
     {
         static $apcuSupported = null;
         if (null === $apcuSupported) {
-            $apcuSupported = ApcuAdapter::isSupported();
+            // Need to check for CLI because Symfony won't: https://github.com/symfony/symfony/pull/25080
+            $apcuSupported = PHP_SAPI === 'cli' ? ini_get('apc.enable_cli') && function_exists('apcu_fetch') : ApcuAdapter::isSupported();
```
tractorcow: Can you please use `Director::is_cli()`. You should probably always check `ApcuAdaptor::isSupported()` in either case.

andrewandante: Had a look around and I can't see a way to enable for CLI-only (and also that would be super weird) so I'm happy to do that.

Wanted to use `Director::is_cli()` in the Symfony PR, grumbled, forgot to use it here 😅

andrewandante: Also, is there a reason this is done is this roundabout manner? Why can't we simply return the ternary? Same with the `PHPFiles` check below

src/Core/Cache/DefaultCacheFactory.php
```diff
@@ -85,7 +86,8 @@ protected function isAPCUSupported()
     {
         static $apcuSupported = null;
         if (null === $apcuSupported) {
-            $apcuSupported = ApcuAdapter::isSupported();
+            // Need to check for CLI because Symfony won't: https://github.com/symfony/symfony/pull/25080
+            $apcuSupported = Director::is_cli() ? ini_get('apc.enable_cli') && ApcuAdapter::isSupported() : ApcuAdapter::isSupported();
```
dhensby: Can't/shouldn't the `ApcuAdapter::isSupported()` method do the `ini` check?

andrewandante: That's the Symfony part

andrewandante: We could make our own `ApcuAdapter` I guess, but that seems like over-engineering

dhensby: > We could make our own ApcuAdapter I guess, but that seems like over-engineering

for sure, don't do that :)

dhensby:
> Travis failures not your fault; This seems fine.

What's causing the failures? I don't see them on the current 4.0 branch :/

dhensby:
ah it's https://github.com/silverstripe/silverstripe-framework/pull/7627

dhensby:
green means go

PR #7626, opened 2017-11-22
Restore legacy $ThemeDir support

### LINKED ISSUE #6030: `ViewableData::ThemeDir()` broken
## Acceptance Criteria

 * `$ThemeDir` is restored to templates
 * The value of ThemeDir depends on the template it is being used within
 * The value will the root-relative URL (i.e. starts with "/") of the theme folder, as processed by ResourceURLGenerator.
 * A template string such as `` will reference images/something.png within the theme.

Exclusions:

 * Direct access to ThemeDir from ViewableData isn't necessary (so it can be implemented as a template global instead if preferred)

----

`$ThemeDir` is used in templates to pull in assets. This is currently used by https://github.com/silverstripe-themes/silverstripe-simple to pull in a JS file.

However this is now broken after moving to cascading themes.

# PRs

 - [ ] https://github.com/silverstripe/silverstripe-framework/pull/7626

tractorcow:
Fixes #6030

Although there is no longer the concept of a single theme with the advent of multiple cascading themes, I've restored support of the approximate behaviour along with a deprecation notice.

### DIFF
--- a/src/View/ViewableData.php
+++ b/src/View/ViewableData.php
@@ -3,9 +3,12 @@
 namespace SilverStripe\View;
 
 use Exception;
+use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Config\Configurable;
 use SilverStripe\Core\Extensible;
 use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\Core\Manifest\ModuleResourceLoader;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\ORM\ArrayLib;
 use SilverStripe\ORM\FieldType\DBField;
 use SilverStripe\ORM\FieldType\DBHTMLText;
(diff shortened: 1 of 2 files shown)


src/View/ViewableData.php
```diff
@@ -586,6 +589,35 @@ public function Me()
...
+     * @return string URL to the current theme
+     * @deprecated 4.0.0..5.0.0 Use $resourcePath or $resourceURL template helpers instead
+     */
+    public function ThemeDir()
+    {
+        Deprecation::notice('5.0', 'Use $resourcePath or $resourceURL template helpers instead');
+        $themes = SSViewer::get_themes();
+        foreach ($themes as $theme) {
+            // Skip theme sets
+            if (strpos($theme, '$') !== false) {
```
flamerohr: this would mean you could have a `$` in the middle of your path and it would skip, could this have the check at the beginning of the name `strpos() === 0` instead?

I guess this is edge-case so not a dealbreaker for me

tractorcow: If you have a `$` in the middle of your theme then it'd probably be skipped because it's a bad theme.

tractorcow:
Fixed @flamerohr

PR #7627, opened 2017-11-22
BUG Fix regressions in asset resize behaviour change

tractorcow:
Fixes regressions from https://github.com/silverstripe/silverstripe-assets/pull/101


tractorcow:
Green :D

PR #7628, opened 2017-11-22
Travis retry on imagick install

dhensby:


dhensby:
it works:

https://travis-ci.org/silverstripe/silverstripe-framework/jobs/305802929#L219

PR #7630, opened 2017-11-22
[Don’t merge] More attempts at fixing Behat/Travis/3.x

kinglozzer:


dhensby:
this stuff is such a pain

kinglozzer:
And with that, I give up for today. I tested a few Firefox versions/ESR versions, newer versions won’t connect and time out, and old versions (like those recommended [here](https://github.com/SeleniumHQ/selenium/blob/master/java/CHANGELOG#L1071-L1081)) tend to instead error with:

> `[Behat\Behat\Exception\ErrorException]
                                                                                             
> Exception has been thrown in "beforeScenario" hook, defined in SilverStripe\Framework\Test\Behaviour\FeatureContext::before()

> Notice: Undefined index: size in /home/travis/builds/ss/vendor/instaclick/php-webdriver/lib/WebDriver/AbstractWebDriver.php line 204`

Not sure if that means they’re connecting successfully and then failing on something else...

Maybe a version somewhere between 31 and 35 will work... who knows?

tractorcow:
Yeah I've been through all of this... long story short, firefox / geckodriver isn't ready to update, stick with 31ESR for now.

kinglozzer:
@dhensby tried Firefox 31 but [that failed too](https://travis-ci.org/silverstripe/silverstripe-framework/jobs/295689065) 😢. I’ll try pushing a build later with 31 ESR specified just in case it’s any different

kinglozzer:
> Chrome works fine though.

If we can’t get Firefox working, perhaps that’s worth a shot? I’d rather that than lose Behat builds for 3.x entirely

dhensby:
How easy is that?

dhensby:
We are green due to upstream dependency! https://github.com/instaclick/php-webdriver/issues/82#thread-subscription-status

PR #7635, opened 2017-11-23
FIX Update meber passwordencryption to default on password change

### LINKED ISSUE #6770: Update Member.PasswordEncryption on password update
After doing some work to implement a custom `PassswordEncryptor ` I noticed that for any user in the database once their *PasswordEncryption* property has been set, it is never updated.

This, imo, poses a little bit of a security concern as if I have an encryption method that gets compromise during the website's lifecycle and I decide to change the security setting to use a different encryption method it will only be applied to new users...

As seen in [`Member.php` lines 939-953](https://github.com/silverstripe/silverstripe-framework/blob/3.5.3/security/Member.php#L939-L953) this could possibly be changed very easily by just removing the *inline if* check and always use the Security set encryptor instead of the Database one for password changes, ie a new password is being encrypted and saved to the database with no bearings to the old one.

I may be able to provide this change as a PR in the next couple of days, but thought I'd bring the issue up first to see what others have to say (and potentially highlight any drawbacks from my proposed change).

Cheers.

dhensby:
fixes #6770

### DIFF
--- a/security/Member.php
+++ b/security/Member.php
@@ -975,8 +975,7 @@ class Member extends DataObject implements TemplateGlobalProvider {
 			$encryption_details = Security::encrypt_password(
 				$this->Password, // this is assumed to be cleartext
 				$this->Salt,
-				($this->PasswordEncryption) ?
-					$this->PasswordEncryption : Security::config()->password_encryption_algorithm,
+				$this->isChanged('PasswordEncryption') ? $this->PasswordEncryption : null,
 				$this
 			);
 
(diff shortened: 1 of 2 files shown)


dhensby:
OK - so there's a test that specifically checks that the encryption algorithm doesn't change when passwords are changed. I removed this test as it's clearly a contradiction to what I'm trying to do.

Happy to take thoughts on whether this is better or not.

tractorcow:
Seems ok to me. Please merge on green (excluding known-breaking behat tests).

tractorcow:
Technically an enhancement, but it improves security so allowing in a patch release.

Issue #7636, opened 2017-11-23
Login


MichaelJJames:
Hi,

When logging into an SS 4.0.0 site I get directed to /Security/login/default instead of /admin/pages is this the correct behaviour?

Thanks
Michael

MichaelJJames:
This is when I click logout in the admin to switch user would it be an idea to have a default redirect to /admin/pages when there isn't any backurl in the URL

dhensby:
`/Security/login/default` might be incorrect and we should look into why that's happening, but if you're being sent to the wrong page after login I'd recommend setting the `Security.default_login_dest` config to what you want. I don't think we should assume the CMS interface is the default login destination for all users

tractorcow:
It's correct, in that if you don't specify a default destination, it remains within the security section.

I've posted a suggestion on #7637 on improving the overall experience.

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-admin/pull/368 thanks @dhensby and thanks for the extra explanation. :)

tractorcow:
Also the other thing I was asking for implemented with https://github.com/silverstripe/silverstripe-framework/pull/7649, thanks @kinglozzer

PR #7637, opened 2017-11-24
Login Redirect

MichaelJJames:
Issue I found yesterday with the login not redirecting when no backURL is set #7636

### DIFF
--- a/src/Security/MemberAuthenticator/LoginHandler.php
+++ b/src/Security/MemberAuthenticator/LoginHandler.php
@@ -205,8 +205,8 @@ class LoginHandler extends RequestHandler
             Security::singleton()->setSessionMessage($message, ValidationResult::TYPE_GOOD);
         }
 
-        // Redirect back
-        return $this->redirectBack();
+        // Redirect to /admin/pages/ as a default if no backURL or default login dest is set
+        return $this->redirect('/admin/pages/');
     }
 
     /**


tractorcow:
My own experience is that when attempting to access the CMS, and framework requests the user to log out of one member, and login as another, the "BackURL" is lost. There are quite a few steps involved where we COULD possibly persist the BackURL a bit better:

- Login as non-cms user
- Access /admin
- Shows message `I'm sorry, but you can't access that part of the CMS. If you want to log in as someone else, do so below.`
- Click "Log in as someone else"
- Log in as member with CMS permissions
- Redirect to home page (when I would expect to be sent back to admin)

It's not so much a bug, but it is a necessary enhancement that would slightly improve the re-authentication experience.

tractorcow:
@MichaelJJames I would see if adding the BackURL querystring to the "login as someone else" action might fix this.

dhensby:
@tractorcow the specific issue being described is logging out of the CMS then logging in again.

1. Logging out of the CMS takes you to a login page, should it? Should you just be taken to the home page after logging out of the CMS? It seems strange to assume that you'd be wanting to log back into an area you just logged out of
2. Can we add a `BackURL` to the logout link in the cms? - just tested, the answer is yes - we can add `?BackURL=admin` to the logout url and it'll set up the back url goodness

dhensby:
> I would see if adding the BackURL querystring to the "login as someone else" action might fix this.

This should also be done, too. If there's a permissions failure and you've been redirected to the login form (but are logged in) then the BackURL should be persisted

kinglozzer:
For reference, the 3.x behaviour is to just redirect back after logging out. E.g:

- Log in
- Visit `/admin/pages/edit/show/1`
- Click log out button (href is just `/Security/logout`)
- Get redirected “back”, based on referer header, which because of a permission failure will redirect you to `/Security/login?BackURL=%2Fadmin%2Fpages%2Fedit%2Fshow%2F1`

I think this may be something that I missed when adding the CSRF token to the logout action (I added it to the [logout form](https://github.com/silverstripe/silverstripe-framework/blob/4.0/src/Security/LogoutForm.php#L54-L63) for when the token is missing). We should be able to populate the back URL hidden field from the referer header I think?

dhensby:
see https://github.com/silverstripe/silverstripe-admin/pull/368

I prefer being explicit than relying on a referer header. We could add the referer if it exists, but I'm not convinced that's a good idea.

kinglozzer:
I’ll submit a PR to re-add the referer behaviour to `LogoutHandler`: that’s something that was missed by me, and hitting the logout URL without a valid CSRF token already falls back to the referer header as per the 3.x behaviour.

The two PRs aren’t mutually exclusive, though, so we can still add the base admin URL as `BackURL` if it’s deemed a better default

dhensby:
ok - I'm closing this as it's not a solution we are happy with and we've got a couple of PRs that shuuld solve the initial bug

MichaelJJames:
Great thanks guys for sorting it :)

PR #7641, opened 2017-11-26
Reintroduce checking against original php.ini values.

Sil3ntStorm:
Seems to have been accidentally removed in 806ffb934e32644ddefc0dfb470d76218bd39334

Fixes https://github.com/silverstripe/silverstripe-installer/issues/211

### DIFF
--- a/src/Dev/Install/InstallRequirements.php
+++ b/src/Dev/Install/InstallRequirements.php
@@ -227,8 +227,9 @@ class InstallRequirements
     /**
      * Check everything except the database
      */
-    public function check()
+    public function check($originalIni)
     {
+        $this->originalIni = $originalIni;
         $this->errors = [];
         $isApache = $this->isApache();
         $isIIS = $this->isIIS();
--- a/src/Dev/Install/install5.php
+++ b/src/Dev/Install/install5.php
@@ -66,7 +66,7 @@ $theme = $config->getTheme($_REQUEST);
 
 // Check requirements
 $req = new InstallRequirements();
-$req->check();
+$req->check($originalIni);
 
 if ($req->isIIS()) {
     $webserverConfigFile = 'web.config';


src/Dev/Install/InstallRequirements.php
```diff
@@ -227,8 +227,9 @@ public function findWebserver()
     /**
      * Check everything except the database
      */
-    public function check()
+    public function check($originalIni)
```
tractorcow: It's still not going to be strictly semver, but to make this non-breaking, can you assign this in a separate setter instead of adding a method argument? E.g. `setOriginalIni($ini)`.

dhensby: Or just give it a default value? `$originalIni = true`?

dhensby: derp, ignore me, it's not a bool

dhensby:
I've messed this up. @tractorcow wrote a comment I accidentally deleted. The sentiment was "Please add as setter `setOriginalINI($originalINI)`.

My thoughts where that if this was removed by mistake from a beta release then I don't mind adding it back as a parameter. I feel that's less of a semver issue compared to adding a brand new method.

Sil3ntStorm:
The way I see it there are three options:
- add a second constructor parameter for it to InstallRequirements and just construct the object with it in  install5.php
- add a new setter function to InstallRequirements and call it before calling check()
- make the added parameter to check() optional

I don't really care either way but I don't claim to understand the potential semver issues. However adding to an API (that isn't strictly public?) should be fine I guess?
Whatever gets the job done is fine by me. If you come to a conclusion on which solution you prefer the most let me know and I will update the branch accordingly.

dhensby:
I'll defer to @tractorcow

tractorcow:
I think I should avoid bike shedding and get this appropriate fix in.

PR #7643, opened 2017-11-27
ENHANCEMENT Promote portugese (portugal) as primary locale

### LINKED ISSUE #6561: Portuguese translation confusion
Interface language cannot be set to Portuguese (Portugal) in Security section, Member profile. However all modules have both *pt* and *pt_BR* translations.

According to Unicode CLDR, *pt* is aliased with *pt_BR*, so it does in `i18n.likely_subtags` config.
so, to enable *pt* as *pt_PT* translation we need to change `i18n.likely_subtags` alias or rename translation files into pt_PT.yml

tractorcow:
Fixes #6561

### DIFF
--- a/src/i18n/Data/Intl/IntlLocales.php
+++ b/src/i18n/Data/Intl/IntlLocales.php
@@ -793,7 +793,7 @@ class IntlLocales implements Locales, Resettable
         'pl' => 'pl_PL',
         'pon' => 'pon_FM',
         'ps' => 'ps_AF',
-        'pt' => 'pt_BR',
+        'pt' => 'pt_PT',
         'qu' => 'qu_PE',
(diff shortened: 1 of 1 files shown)


sminnee:
Does this break anything for Brazillians?

tractorcow:
No, both locales are independently available.

image

robbieaverill:
In my opinion and my understanding of semver, if it's an enhancement we should be targeting the `4` branch. On the issue @tractorcow you said "Treating as patch enhancement", but IMO this isn't something that semver caters for. Having said that, reading the issue and this PR it seems like a bugfix to me anyway.

tractorcow:
In this context, I'm not treating data as an API. :)

PR #7644, opened 2017-11-27
MINOR Text fixes in docs

michalkleiner:
Remove value from the name of param (flush), fix pipe symbol (showqueries)


robbieaverill:
Thanks @michalkleiner!

Issue #7647, opened 2017-11-27
[I18N] BUG i18nTextCollector should support SomeClass::class syntax


robbieaverill:
Currently the following would cause an error in the text collector:

```php
return _t(SomeClass::class . '.Title', 'Some title');
```

As a result, all modules need to use FCQN in strings when referencing translations in other files.

We should fix the parser to support `::class` syntax.

For reference, `__CLASS__` currently works.

## Pull requests

- [x] Support for `self::class`: #7648 
- [x] Support for `Some::class`: https://github.com/silverstripe/silverstripe-framework/pull/10577

unclecheese:
The text collector works on static analysis, so this would be a big change. Even just getting `__CLASS__` to parse was a significant amount of work as I remember.

For strings that cannot be resolved statically, you should use `i18nEntityProvider`.

Relevant: https://github.com/silverstripe/silverstripe-framework/blob/4.0/src/i18n/TextCollection/i18nTextCollector.php#L619

unclecheese:
Just having a quick look -- it appears as if the intention is to skip entities that cannot be statically analysed. https://github.com/silverstripe/silverstripe-framework/blob/4.0/src/i18n/TextCollection/i18nTextCollector.php#L619

What was the error you were getting?

robbieaverill:
I've added a PR to enable `self::class` at #7648.

As @unclecheese noted, due to the fact that this is using static code analysis, it's not realistic to allow `parent::class` (without reflection) or `static::class` at this stage.

I've got a WIP to enable `SomeObject::class` support, which can be done with the following assumptions:

* SomeObject has been imported via a `use` statement, or;
* SomeObject exists in the same namespace (global or otherwise) as the current class

I think with the two assumptions above that the first one is safe, because the class has a `use` for the class, so it's not that important if it exists. The second assumption is "safe" in that it's how PHP would treat the class too.

I'll chip away at this for a few days - I facepalmed pretty hard when I looked at this code.

robbieaverill:
Yeah @unclecheese - there is some code that's using `::class` at the moment (it's already been noted that it's an issue, although not actioned yet) - see `SiteConfigLeftAndMain` for an example which has a set of `LeftAndMain::class` calls in it.

unclecheese:
Nice work on this, @robbieaverill .

tractorcow:
Instead of implementing this ourselves we should use the nikic/php-parser that's now a part of core.

[...]

PR #7648, opened 2017-11-27
FIX Support self::class text collection

robbieaverill:
Issue: #7647

This enables `_t(self::class . '.Foo', 'Foo Bar')` style translations.

### DIFF
--- a/src/i18n/TextCollection/i18nTextCollector.php
+++ b/src/i18n/TextCollection/i18nTextCollector.php
@@ -558,6 +558,7 @@ class i18nTextCollector
         $inNamespace = false;
         $inClass = false; // after `class` but before `{`
         $inArrayClosedBy = false; // Set to the expected closing token, or false if not in array
+        $inSelf = false; // Tracks progress of collecting self::class
         $currentEntity = array();
         $currentClass = []; // Class components
         $previousToken = null;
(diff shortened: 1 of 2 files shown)


dhensby:
Personally, I'd interpret `static` as `self`... It's a really bad idea to use late static binding on translation keys because there's no guarantee that translation will exist..

I think `parent::class` is equally risky to use as well, so can probably be ignored

robbieaverill:
I think it would be better to let the collector skip or trigger a user error for static and parent rather than treat static as self

dhensby:
Well, `static` definitely includes `self` so you could collect it as `self` and display a warning too?

tractorcow:
I think collecting as `self` is fine; we really don't want to maintain this old system for very long anyway, and build a non-self-rolled parser some time soonish.

tractorcow:
@robbieaverill feel free to PR a `static` patch as another PR if you think it's still good to have, but not a blocker. :)

tractorcow:
Wow this fix is going to make me SO happy come release time.

Another fix we need is `Page::class` style support, but that MAY require a php parser change. It was a lot of work in upgrader to get the namespace class context.

robbieaverill:
Yeah I'm not that concerned about `static` support to be honest. As @dhensby it'd be a bad idea to use it.

I've got a WIP for `Page::class` support, some notes added to #7647 about this. It involves tracking `use` imports and the current namespace. Nearly there :)

kinglozzer:
Oh wow that looks like a definite candidate for a php-parser refactor! I wonder if that can correctly resolve namespaces @dhensby?

dhensby:
> I wonder if that can correctly resolve namespaces @dhensby?

Oh, it can. it's out of the box :)

dhensby:
@robbieaverill I built my own namespace resolving logic into 3.x (https://github.com/silverstripe/silverstripe-framework/blob/3/core/manifest/ClassManifest.php)

There's the `get_imported_namespace_parser()` method and `getImportsFromTokens()`

robbieaverill:
Nice! I'll try and get a PR up for `Some::class` tonight

PR #7649, opened 2017-11-27
FIX: Restore BackURL preservation on log out (closes #7636)

### LINKED ISSUE #7636: Login

kinglozzer:
I overlooked this when adding the CSRF token to the logout action. It’s the same behaviour from the user’s point-of-view as 3.x and is in-place in the [logout form](https://github.com/silverstripe/silverstripe-framework/blob/4.0.0/src/Security/LogoutForm.php#L54-L63) (which people will hit if the token is missing/invalid), just not in `LogoutHandler` itself, which is why I’ve submitted against `4.0`. Happy to change branch if `4` is more appropriate.

This comes _after_ `RequestHandler::getBackURL()` is checked, so appending a `?BackURL` param to the logout action (like in https://github.com/silverstripe/silverstripe-admin/pull/368/) will take priority over the referer header.

### DIFF
--- a/src/Security/MemberAuthenticator/LogoutHandler.php
+++ b/src/Security/MemberAuthenticator/LogoutHandler.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\Security\MemberAuthenticator;
 
+use SilverStripe\Control\Controller;
 use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\RequestHandler;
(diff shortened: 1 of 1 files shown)


tractorcow:
Wow thanks @kinglozzer I think that's exactly what I was after! Glad you found the time to track it down and fix it.

PR #7654, opened 2017-11-28
BUG Fix table name test

flamerohr:


flamerohr:
I need more coffee - not sure why I created a whole branch and pull request when there already is one :P

PR #7655, opened 2017-11-28
BUG Fix broken scrutinizer

tractorcow:


tractorcow:
I've gone a bit risky and re-enabled scrutinizer. Previously it was ONLY enabled for 3 and master branches, so a bit of disclosure there. :)

@robbieaverill can you confirm this is safe? (https://scrutinizer-ci.com/g/silverstripe/silverstripe-framework/settings is settings page to confirm)

robbieaverill:
As far as I'm concerned it's safe to enable on all branches

PR #7658, opened 2017-11-29
TinyMCE lang is fr_FR for all French languages

lekoala:
"fr" is not a valid tiny mce lang

https://www.tinymce.com/docs/configure/localization/

having "fr" as it is currently prevents loading the language file


src/Forms/HTMLEditor/TinyMCEConfig.php
```diff
@@ -74,49 +74,49 @@ class TinyMCEConfig extends HTMLEditorConfig
...
-        'fr_VU' => 'fr',
-        'fr_WF' => 'fr',
-        'fr_YT' => 'fr',
-        'fr_GB' => 'fr',
-        'fr_US' => 'fr',
+        'fr_BE' => 'fr_FR',
+        'fr_BF' => 'fr_FR',
+        'fr_BI' => 'fr_FR',
+        'fr_BJ' => 'fr_FR',
+        'fr_CA' => 'fr_FR',
```
dhensby: Does this not need to be `fr_ca` as currently or does that not work?

lekoala: I didn't test, but there is no reason why it should work because there is only one js file which is fr_FR.js

dhensby: ok

MohammaddReza:
Persian language has similar problems (fa_IR)

tractorcow:
@MohammaddReza fixed at https://github.com/silverstripe/silverstripe-framework/pull/7667

PR #7661, opened 2017-11-30
FIX Use baseDataClass for allVersions as with other methods

dhensby:
The other methods in `Versioned` use `ClassInfo::baseDataClass()` to find base classes to grab versions from and I feel it should be used here too.

At the moment it's not possible to use `$this->allVersions()` with mocked objects because the mock class doesn't exist as a table/registered `DataObject`

### DIFF
--- a/model/Versioned.php
+++ b/model/Versioned.php
@@ -1050,7 +1050,7 @@ class Versioned extends DataExtension implements TemplateGlobalProvider {
 		$oldMode = self::get_reading_mode();
 		self::reading_stage('Stage');
 
-		$list = DataObject::get(get_class($this->owner), $filter, $sort, $join, $limit);
+		$list = DataObject::get(ClassInfo::baseDataClass($this->owner), $filter, $sort, $join, $limit);
 		if($having) $having = $list->having($having);
 
 		$query = $list->dataQuery()->query();


model/Versioned.php
```diff
@@ -1050,7 +1050,7 @@ public function allVersions($filter = "", $sort = "", $limit = "", $join = "", $
 		$oldMode = self::get_reading_mode();
 		self::reading_stage('Stage');
 
-		$list = DataObject::get(get_class($this->owner), $filter, $sort, $join, $limit);
+		$list = DataObject::get(ClassInfo::baseDataClass($this->owner), $filter, $sort, $join, $limit);
```
kinglozzer: >At the moment it's not possible to use `$this->allVersions()` with mocked objects because the mock class doesn't exist as a table/registered `DataObject`

Can you add a unit test for this? If `MyMockDataObject extends DataObject`, I can’t see why `DataObject::get('MyMockDataObject')` would fail 😕 

Won’t this now result in `DataObject::get('SiteTree')` rather than `DataObject::get('MyPageType')`? That could potentially result in versions with the wrong `ClassName` being returned couldn’t it?

dhensby: By mocked databojects I don't mean stubs like your example I mean mocked objects like `$mock = $this->createMock('SiteTree');` in PHPUnit tests.

This will result in getting *all* versions for this record regardless of its class, yes, which I think is expected behaviour... If I've changed it from Page to HomePage, getting the `Page` objects is still part of what you'd expect to see in `allVersions()`, right?

dhensby: `baseDataClass` is used several times in the Versioned class to make sure all versions of a record are accounted for (including being used in `get_latest_version`):

- https://github.com/silverstripe/silverstripe-framework/blob/3.5/model/Versioned.php#L935
- https://github.com/silverstripe/silverstripe-framework/blob/3.5/model/Versioned.php#L1293
- https://github.com/silverstripe/silverstripe-framework/blob/3.5/model/Versioned.php#L1346
- https://github.com/silverstripe/silverstripe-framework/blob/3.5/model/Versioned.php#L1392
- https://github.com/silverstripe/silverstripe-framework/blob/3.5/model/Versioned.php#L1440
- https://github.com/silverstripe/silverstripe-framework/blob/3.5/model/Versioned.php#L1496
- https://github.com/silverstripe/silverstripe-framework/blob/3.5/model/Versioned.php#L1514

tractorcow:
Fine with me, just updating branch before merge.

PR #7662, opened 2017-11-30
Add php 7.2 to test matrix

dhensby:
PHP 7.2 was released today! yay


dhensby:
we can expect this to fail (http://php.net/manual/en/migration72.incompatible.php#migration72.incompatible.object-reserved-word)

PR #7664, opened 2017-11-30
Add PHP nightly builds to travis

dhensby:


robbieaverill:
Nice, builds were cancelled though. Wanna restart them?

dhensby:
Sure, started them again - I cancelled the others so I could see if the nightly was running/working/passing (which it is, yay)

PR #7665, opened 2017-11-30
NEW Add php 7.2 support

dhensby:
Minimal effort attempt to see if we can get this working...

### DIFF
--- a/model/fieldtypes/compat/autoload.php
+++ b/model/fieldtypes/compat/autoload.php
@@ -2,16 +2,24 @@
 
 if (PHP_MAJOR_VERSION < 7) {
 	spl_autoload_register('php5_compat_autoloader');
+	spl_autoload_register('php7_compat_autoloader');
+} elseif (PHP_MAJOR_VERSION === 7 && PHP_MINOR_VERSION < 2) {
+	spl_autoload_register('php7_compat_autoloader');
 }
 
 function php5_compat_autoloader($classname) {
-	$classMap = array(
-		"int" => "/framework/model/fieldtypes/compat/Int.php",
-		"float" => "/framework/model/fieldtypes/compat/Float.php",
-	);
+	switch (strtolower($classname)) {
+		case 'int':
+			require_once BASE_PATH . '/framework/model/fieldtypes/compat/Int.php';
+			break;
+		case 'float':
+			require_once BASE_PATH . '/framework/model/fieldtypes/compat/Float.php';
+			break;
+	}
+}
 
-	$classname = strtolower($classname);
-	if(isset($classMap[$classname])) {
-		require_once BASE_PATH . $classMap[$classname];
+function php7_compat_autoloader($classname) {
+	if (strcasecmp($classname, 'object') === 0) {
+		require_once BASE_PATH . '/framework/core/compat/Object.php';
 	}
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ addons:
 
 matrix:
   include:
+    - php: 7.2
+      env: DB=MYSQL PDO=1
     - php: 5.3
       env: DB=MYSQL
     - php: 5.4
--- a/_config.php
+++ b/_config.php
@@ -23,8 +23,8 @@
  *
  * Overload both of these to support legacy code.
  */
-Object::useCustomClass('SSDatetime', 'SS_Datetime', true);
-Object::useCustomClass('Datetime',   'SS_Datetime', true);
+SS_Object::useCustomClass('SSDatetime', 'SS_Datetime', true);
+SS_Object::useCustomClass('Datetime',   'SS_Datetime', true);
 
 /**
  * The root directory of TinyMCE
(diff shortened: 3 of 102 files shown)


robbieaverill:
I think it's difficult to justify renaming `Object` in a minor version since it's a common class for modules and user code to extend from. I know that puts a road block in the way of SS3 and PHP 7.2.

kinglozzer:
Just to clarify: due to the conditional autoloading of `class Object extends SS_Object`, this shouldn’t break any PHP <7.2 sites. That is, if you’re on PHP <7.2 you won’t need to change anything. You’ll only need to replace instances of `Object` with `SS_Object` in your code if you want to run PHP >=7.2.

I’m not sure there’s a huge amount of value to be gained here. PHP 7.1 is supported for security patches beyond our support for SS 3 anyway, and 7.2 doesn’t offer any night/day performance improvements or features. Plus, we want to encourage people to upgrade to 4 rather than keep “effectively” extending the lifetime of 3.x.

What does everyone else think? @silverstripe/core-team

model/fieldtypes/compat/autoload.php
```diff
@@ -8,6 +8,7 @@ function php5_compat_autoloader($classname) {
 	$classMap = array(
 		"int" => "/framework/model/fieldtypes/compat/Int.php",
 		"float" => "/framework/model/fieldtypes/compat/Float.php",
+        "object" => "/framework/core/compat/Object.php",
```
kinglozzer: I think we need another `if (PHP_MINOR_VERSION...) {` to check specifically for PHP 7.2 for this class. Otherwise this will unnecessarily break things for people using 7.0 and 7.1

dhensby: good point

dhensby:
Whilst I'm not particularly committed to seeing this get over the line, as it was pretty low effort (<1hr work) to get done (thanks PHPStorm).

The same logic applies to this fix as applies to the 7.1 fix - user code that wouldn't run on 7.2 code before won't run now but if their code *could* have run on 7.2, it will. And that's without BC breakages.

model/fieldtypes/compat/autoload.php
dhensby: good point

sminnee:
I see a couple of things:

Compared to PHP7 Support this is less valuable: we’re not gaining the 200%+ speedup of php5.6 -> 7, and we’re not using this to extend the ss3 supportable lifetime.

Compared to renaming Int and Float, this is more invasive: it was rare that developers made direct reference to those classes, whereas it’s reasonably common to make Object subclasses and possibly instanceof checks.

For both of these reasons I’m less enthusiastic about this change, although I’m not dead set against it.

dhensby:
How do we feel about branching off 3.7 so this doesn't make it into a 3.7 release but then get it into 3.x-dev afterwards meaning that 3.8 (if it ever happens) would have it?

> For both of these reasons I’m less enthusiastic about this change, although I’m not dead set against it.

I'm also not that fussed, like I said above it was a really small amount of time/effort to get this working and I'm not too fussed either way.

Happy to close if there's no consensus

kinglozzer:
Maybe close for now and just keep an ear to the ground to see if demand for this crops up? At least now we’ve got a clear plan for how we can solve this

flamerohr:
At least those who are really determined to keep the ball rolling can use/refer to this pull request as well.

I'd be keen to wait and see how much demand there is to get 7.2 support out for 3.x

tractorcow:
I feel like a safer alternative is capping the max PHP version to `< 7.2.0`, at least on 3.6.

What was specific to php 7.2 that changed `object` class that wasn't a problem in php 7.1 / 7.0?

sminnee:
@tractorcow PHP 7.2 introduces `object` as a reserved word, to use a type-hint.

PHP 7 added "object" as a "soft reserved" word which basically said "don't use these they might become reserved in the future". It was that statement that drove the removal of the Object class in SS4.

chillu:
@tractorcow Can you please create a 4.x focused issue for this if there's any work left?

dhensby:
> Can you please create a 4.x focused issue for this if there's any work left?

I don't know of any 7.2 issues with 4.x - tests builds pass on 7.2 and nightly (last time I checked)

tractorcow:
@dhensby the difflib library in third party still has deprecation issues I think.

IsaacInsoll:
Is this going to make it into a SS3 release?
Our production servers are currently 5.6 and I just updated my local dev environment to 7.2 (from 7.0) to test then stumbled into this 'Object' issue. The reason I ask is because we are going to update our production servers to PHP7 for performance reasons and I thought we should go with the latest version of PHP7. 

We will eventually migrate to SS4 but it would have been nice to be able to set up the new production servers with 7.2 now and be done with it. Obviously we will have to use 7.1 which is no problem at all but just wanted to explain a potential reason why we might want PHP7.2 support in SS3.

tractorcow:
3.7 is coming out really soon.

PR #7667, opened 2017-12-01
BUG Ensure that all tinymce_lang mappings are valid

tractorcow:
I've added a test to prove that all locale codes are valid (if admin is installed).

Addresses remaining feedback from https://github.com/silverstripe/silverstripe-framework/pull/7658#issuecomment-348346248

@MohammaddReza I hope this fixes your issue. :)

Unfortunately I've had to remove `mi_NZ` since we don't have localised tinymce for that. I'm drawing these from https://www.tinymce.com/download/language-packages/


robbieaverill:
>Unfortunately I've had to remove mi_NZ since we don't have localised tinymce for that

That's a shame :( oh well

phalkunz:
@chillu, @dhensby already reviewed and approved it. I guess I don't need to re-review it.

flamerohr:
@phalkunz worth a re-review sometimes :)

PR #7668, opened 2017-12-01
NEW: Add ViewableData::getViewerTemplates()

kinglozzer:
Forms part of the fix for silverstripe/silverstripe-cms#2039, merge before https://github.com/silverstripe/silverstripe-cms/pull/2041

### DIFF
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -2,35 +2,36 @@
 
 namespace SilverStripe\ORM;
 
+use BadMethodCallException;
+use Exception;
+use InvalidArgumentException;
+use LogicException;
+use SilverStripe\Control\HTTP;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Resettable;
-use SilverStripe\Dev\Deprecation;
 use SilverStripe\Dev\Debug;
-use SilverStripe\Control\HTTP;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\FormField;
 use SilverStripe\Forms\FormScaffolder;
 use SilverStripe\i18n\i18n;
 use SilverStripe\i18n\i18nEntityProvider;
 use SilverStripe\ORM\Connect\MySQLSchemaManager;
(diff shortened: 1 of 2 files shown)


kinglozzer:
Uhh, so my editor decided to re-order those namespace imports... I can remove them if they’ll make merges a pain

src/ORM/DataObject.php
```diff
@@ -2146,6 +2147,14 @@ public function getFrontEndFields($params = null)
         return $untabbedFields;
     }
 
+    /**
+     * {@inheritdoc}
+     */
```
robbieaverill: AFAIK this would be the same result as not having a doc block at all, so may as well omit it

kinglozzer:
I’ve pointed this to the wrong branch as the new API is required for https://github.com/silverstripe/silverstripe-cms/pull/2041 to be merged. So it’s kinda both a bugfix and new API... 😛

Rationale: https://github.com/silverstripe/silverstripe-cms/pull/2041#issuecomment-348520027

src/ORM/DataObject.php
kinglozzer: Good point, will remove 😄

src/View/ViewableData.php
```diff
@@ -578,6 +579,17 @@ public function getIterator()
...
 
+    /**
+     * Find appropriate templates for SSViewer to use to render this object
+     *
+     * @param string $suffix
+     * @return array
+     */
+    public function getViewerTemplates($suffix = '')
+    {
+        return SSViewer::get_templates_by_class(static::class, $suffix, self::class);
```
tractorcow: 👍

tractorcow:
We have a rather unique case here, where we want to add a new API to fix a bug. @kinglozzer feels it should be 4.0, @robbieaverill feels it should be 4. Personally I'm leaning towards 4 as well, but I want to invite others to include their view on this.

Any of the other @silverstripe/core-team want to vote on whether this is 4.0 or 4? 👍  this comment for 4.0, 👎  for 4.

robbieaverill:
We're as early as possible in the 4.x release cycle, so now would be the time to do this in 4.0 if we need to. This change looks super useful.

sminnee:
If it's code that supports a bugfix then I think it's okay. If you wanted to be really strict you could mark it as "internal" in 4.0 and remove that when you merge to 4.

Issue #7384, opened 2017-09-18
Performance recommendation: DataObjectSchema caching


tractorcow:
Parent issue https://github.com/silverstripe/silverstripe-admin/issues/199

Many methods on DataObjectSchema require a significant amount of bootstrapping to populate from the config and underlying class model; It should be possible to apply a simple cache on top of this which will ensure that built caches are persisted across multiple requests.

These properties could be cached:

 - compositeFields
 - databaseFields
 - databaseIndexes
 - defaultDatabaseIndexes
 - tableNames
 
In addition we could move additional ClassInfo methods into DataObjectSchema which rely on schema:
 - ClassInfo::dataClassesFor
 
These methods are not well defined and could probably just be deprecated / inlined

 - ClassInfo::getValidSubClasses (deprecate / move to siteconfig)
 
 
Using a similar model to what we use for config, we could write the cache on __destruct / register_shutdown_function, and read from cache on DataObjectSchema::__construct.

chillu:
Removing this from the milestone since it's the least "cost/benefit ratio" of all the suggestions (effort/medium and impact/medium).

tractorcow:
I'm not sure how much of this work is necessary; We have probably over-estimated the impact. Most of these fields are recalculated each request, but from cached config. It may be more complex than we actually get in value from this.

@chillu I suggest we just close this and focus on feature development / bugfixes.

sb-relaxt-at:
We have done some profiling in order to improve performance in the backend. Although the calls are cached on a lower level there is is an enormous amount of calls to these methods (10-20k) that sum up to several milliseconds of execution time even for simple methods. Our tests show that in-memory caching of these methods can improve performance by 10-20%. Especially the rendering of the treeview is about 20% faster (going down from about 180ms to 150ms). But even the frontend can benefit. We haven't done any testing on other website instances but we can gladly share implementation and detailed testing results.

chillu:
Hey Stephan, that sounds very interesting - keen to see implementation and testing results :)

PR #7670, opened 2017-12-01
FIX: Remove some unnecessary ClassInfo calls in DataObjectSchema

kinglozzer:
`ClassInfo::ancestry()` will include `ViewableData` and `DataObject` in its results. Subsequent calls to `DataObjectSchema::databaseFields(ViewableData::class)` are a waste of time of course, as there are no database fields on those classes.

Below profile is for a homepage load on a blank install, but the performance difference will only grow the more `DataObject` classes are referenced during a given request.

screen shot 2017-12-05 at 12 23 47

### DIFF
--- a/src/ORM/DataObjectSchema.php
+++ b/src/ORM/DataObjectSchema.php
@@ -138,11 +138,11 @@ class DataObjectSchema
      */
     public function baseDataClass($class)
     {
-        $class = ClassInfo::class_name($class);
         $current = $class;
         while ($next = get_parent_class($current)) {
             if ($next === DataObject::class) {
-                return $current;
+                // Only use ClassInfo::class_name() to format the class if we've not used get_parent_class()
+                return ($current === $class) ? ClassInfo::class_name($current) : $current;
             }
             $current = $next;
         }
@@ -201,6 +201,11 @@ class DataObjectSchema
         $db = [];
         $classes = $uninherited ? [$class] : ClassInfo::ancestry($class);
         foreach ($classes as $tableClass) {
+            // Skip irrelevant parent classes
+            if (!is_subclass_of($tableClass, DataObject::class)) {
+                continue;
+            }
+
             // Find all fields on this class
             $fields = $this->databaseFields($tableClass, false);
             // Merge with composite fields
@@ -394,6 +399,10 @@ class DataObjectSchema
      */
     public function classHasTable($class)
     {
+        if (!is_subclass_of($class, DataObject::class)) {
+            return false;
+        }
+
         $fields = $this->databaseFields($class, false);
         return !empty($fields);
     }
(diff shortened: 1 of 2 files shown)


src/ORM/DataObjectSchema.php
```diff
@@ -201,6 +201,11 @@ public function fieldSpecs($classOrInstance, $options = 0)
         $db = [];
         $classes = $uninherited ? [$class] : ClassInfo::ancestry($class);
         foreach ($classes as $tableClass) {
+            // Skip irrelevant parent classes
+            if ($tableClass !== DataObject::class && !is_subclass_of($tableClass, DataObject::class)) {
```
kinglozzer: @tractorcow Omitting the `$tableClass !== DataObject::class` saves around another 500 method calls. As far as I can tell `$this->databaseFields(DataObject::class);` doesn’t return anything, do you think that would be safe?

src/ORM/DataObjectSchema.php
```diff
@@ -394,6 +399,10 @@ public function databaseIndexes($class, $aggregated = true)
      */
     public function classHasTable($class)
     {
+        if (!is_subclass_of($class, DataObject::class)) {
+            return false;
+        }
```
kinglozzer: Somehow `ViewableData` gets passed in here, I haven’t yet tracked down why but this seems like a reasonable change anyway.

dhensby: I think removing the assertion is ok... It's kind of stupid IMO, we probably should have been catching it

dhensby:
This seems reasonable to me - the exception throwing is the only thing we need to take a view on regarding if that's an API issue as people could be relying on it...

kinglozzer:
I’ve removed the test assertion for an exception to be thrown in `classHasTable()`, and also prevented unnecessary calls to `DataObjectSchema::databaseFields(DataObject::class)`, so this is now a “kitchen sink, let me know if anything needs to come out” PR 😝.

Also, as this is targeting `ClassInfo` calls, this should still offer a performance improvement even if the caching outlined in https://github.com/silverstripe/silverstripe-framework/issues/7384 is implemented.

src/ORM/DataObjectSchema.php
```diff
@@ -138,11 +138,11 @@ public function tableName($class)
      */
     public function baseDataClass($class)
     {
-        $class = ClassInfo::class_name($class);
```
dhensby: Should we not add the `if (!is_subclass_of($tableClass, DataObject::class))` check here?

kinglozzer: It does handle that at the end of the method, but I suppose we could move it to the start before the `while` loop

src/ORM/DataObjectSchema.php
tractorcow: It's from dataClassesFor() not properly filtering is_subclass_of() the way you have been doing (correctly).

src/ORM/DataObjectSchema.php
tractorcow: Nice job. :)

PR #7672, opened 2017-12-02
Fix db autodiscover comment on loading behavior.

russmac:
Reading this comment it seems placing a module folder under vendor will cause it to load however it needs to be under vendor/*/


robbieaverill:
Thanks!

PR #7673, opened 2017-12-04
Add new InheritedPermissionFlusher extension, CacheFlusher service

unclecheese:
This work was originally done as part of  https://github.com/silverstripe/silverstripe-framework/pull/7674, but since caching adds complexity, and the implementation of this caching doesn't add a significant performance gain to the CMS `treeview` render, the foundational work for  a per-member cache flushing service has been broken out into a separate commit so that we can enjoy the benefit of such a service without necessarily merging in the `InheritedPermissions` caching.


unclecheese:
Wondering if `CacheFlusher` needs to be `MemberCacheFlusher`, and `InheritedPermissionsFlusher` needs to be `MemberCacheFlusherExtension`. This service could easily be used for any service that has member-specific caching.

dhensby:
This will need to go against the `4` branch as there are a lot of new APIs/classes

src/Security/InheritedPermissionFlusher.php
```diff
@@ -0,0 +1,104 @@
...
+                Member::class,
+                Group::class
+            ));
+        }
+
+        parent::setOwner($owner);
+    }
+
+    /**
+     * @param CacheFlusher[]
```
robbieaverill: Could we please add a `@throws` PHPDoc for the InvalidArgumentException, and maybe return $this for chaining?

src/Security/InheritedPermissionFlusher.php
```diff
@@ -0,0 +1,104 @@
...
+        return $this->services;
+    }
+
+    /**
+     * Flushes all registered CacheFlusher services
+     */
+    public function flushCache()
+    {
+        $ids = $this->getMemberIDList();
+        foreach ($this->services as $service) {
```
robbieaverill: May as well use `$this->getServices()` here

src/Security/InheritedPermissionFlusher.php
```diff
@@ -0,0 +1,104 @@
...
+     *
+     * @return array|null
+     */
+    protected function getMemberIDList()
+    {
+        if (!$this->owner) {
+            return null;
+        }
+
+        if (!$this->owner->exists()) {
```
robbieaverill: This condition and the one above it are related - do you think it might be worth combining them e.g. `if (!$this->owner || !$this->owner->exists()) {`?

tests/php/Security/InheritedPermissionsFlusherTest.php
```diff
@@ -0,0 +1,100 @@
...
+use SilverStripe\Security\Group;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Security\Tests\InheritedPermissionsFlusherTest\TestCacheFlusher;
+use SilverStripe\Core\Config\Config;
+
+class InheritedPermissionsFlusherTest extends SapphireTest
+{
+    protected static $fixture_file = 'InheritedPermissionsFlusherTest.yml';
+
+    public function setUp()
```
robbieaverill: `setUp()` is protected nowadays

src/Security/InheritedPermissionFlusher.php
unclecheese: In general, I'm not a big fan of using public getters internally. I think it just makes the code harder to read. Is there a good reason to do this?

robbieaverill: I see it as dogfooding the API, but I'm happy for you to leave it the way it is :-)

tests/php/Security/InheritedPermissionsFlusherTest.php
unclecheese: Good catch. I'm really surprised this didn't throw a strict warning.

tractorcow:
Rebased and cleaned up some code style issues. Let's see if this fixes the tests.

tractorcow:
@robbieaverill can you please check I've implemented your feedback?

robbieaverill:
They were pretty minor comments. I'm sure you've both done a fine job :-)

Issue #7442, opened 2017-10-06
RFC Cache inherited permissions


tractorcow:
Applies to assets + cms modules.

Inherited permissions are calculated and cached on a per-request basis, with all cached results stored in a protected variable which is rebuilt on each request.

It would be more prudent to cache this and persist across requests, flushing when models are modified.

# Acceptance criteria:

- On a cache-hit, no DB queries are required
- File / Page modification flushes the cache
- Member, Group or Permission modification flushes the cache
- The same cache is shared between members
- If new member permissions are checked, they're added to the existing shared cache
- The cache supports multi-server setups without creating inconsistencies
- Loading a 1000+ page tree (with three levels of nesting) with CMS_ACCESS_CMSMAIN permissions loads demonstrably faster than before

# Suggested implementation

 - InheritedPermissions takes a CacheInterface dependency, which it uses to cache and restore permissions. Each key would be cached by class and ID ~~, as well as `SSViewer.global_key config` arg.~~ and MemberID.

# Pull Requests

## Merge first
- [x] https://github.com/silverstripe/silverstripe-framework/pull/7673

## Merge next
- [x] https://github.com/silverstripe/silverstripe-framework/pull/7674
- [x] https://github.com/silverstripe/silverstripe-assets/pull/102
- [x] https://github.com/silverstripe/silverstripe-cms/pull/2045

tractorcow:
cc @chillu for review

chillu:
Agreed, but too late for 4.0 stable. In particular since this stuff tends to have high impact side effects and regressions. Removing milestone.

tractorcow:
Affect of this issue is minified slightly by https://github.com/silverstripe/silverstripe-assets/pull/83. Shall we merge this smaller change instead?

tractorcow:
The real challenge here will be cache-invalidation; When we modify a page, is it possible to flush the cache for only itself and it's children?

patricknelson:
Another concern on cache busting is something I noted in https://github.com/silverstripe/silverstripe-cms/issues/1977 - custom implementations of permission methods (e.g. `->canCreate(...)`) which rely on unknown external factors. Also, when it comes to caching, file system is fast, database is faster and memory is (by far) the fastest. I'm more concerned about constant repeated hits to a particular method than I am hits to the DB (unless it's hits to variations of the method, in which case the DB has to be avoided since every query will be unique). At that point, caching should probably happen at multiple levels:

[...]

PR #7674, opened 2017-12-04
ENHANCEMENT: Add caching to InheritedPermissions

unclecheese:
## Do not merge
Until https://github.com/silverstripe/silverstripe-framework/pull/7673 is merged!

Adds backend caching to `InheritedPermissions`, e.g. `canEditMultiple()`.

Resolves https://github.com/silverstripe/silverstripe-framework/issues/7442

### DIFF
--- a/_config/security.yml
+++ b/_config/security.yml
@@ -38,7 +38,3 @@ SilverStripe\Core\Injector\Injector:
       Authenticators:
         cms: '%$SilverStripe\Security\MemberAuthenticator\CMSMemberAuthenticator'
   SilverStripe\Security\IdentityStore: '%$SilverStripe\Security\AuthenticationHandler'
-  SilverStripe\Security\InheritedPermissionFlusher:
-    properties:
-      Services:
-        - '%$SilverStripe\Security\PermissionChecker.sitetree'
--- a/_config/security.yml
+++ b/_config/security.yml
@@ -38,3 +38,7 @@ SilverStripe\Core\Injector\Injector:
       Authenticators:
         cms: '%$SilverStripe\Security\MemberAuthenticator\CMSMemberAuthenticator'
   SilverStripe\Security\IdentityStore: '%$SilverStripe\Security\AuthenticationHandler'
+  SilverStripe\Security\InheritedPermissionFlusher:
+    properties:
+      Services:
+        - '%$SilverStripe\Security\PermissionChecker.sitetree'
--- a/src/Security/Group.php
+++ b/src/Security/Group.php
@@ -305,7 +305,7 @@ class Group extends DataObject
      * including all members which are "inherited" from children groups of this record.
      * See {@link DirectMembers()} for retrieving members without any inheritance.
      *
-     * @param String $filter
+     * @param string $filter
      * @return ManyManyList
      */
     public function Members($filter = '')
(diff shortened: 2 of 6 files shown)


dhensby:
`4` branch, please

_config/security.yml
```diff
@@ -38,3 +38,7 @@ SilverStripe\Core\Injector\Injector:
       Authenticators:
         cms: '%$SilverStripe\Security\MemberAuthenticator\CMSMemberAuthenticator'
   SilverStripe\Security\IdentityStore: '%$SilverStripe\Security\AuthenticationHandler'
+  SilverStripe\Security\InheritedPermissionFlusher:
+    properties:
+      Services:
+        - '%$SilverStripe\Security\PermissionChecker.sitetree'
```
tractorcow: This is in `cms` module already; I'll remove this dupe from here. :)

src/Security/Group.php
```diff
@@ -327,9 +327,16 @@ public function Members($filter = '')
                 $query->removeFilterOn('Group_Members');
             });
         }
+
         // Now set all children groups as a new foreign key
-        $groups = Group::get()->byIDs($this->collateFamilyIDs());
-        $result = $result->forForeignID($groups->column('ID'))->where($filter);
+        $familyIDs = $this->collateFamilyIDs();
+        if (!empty($familyIDs)) {
```
tractorcow: This will never be empty, familyIDs always includes self ID too.

src/Security/Group.php
```diff
@@ -327,9 +327,16 @@ public function Members($filter = '')
...
+            $groups = Group::get()->byIDs($familyIDs);
+            $groupIDs = $groups->column('ID');
+            if (!empty($groupIDs)) {
```
tractorcow: This question is related to existing code, but why do we query groups based on a list of ids, then just get the ID column from that result? Isn't that just another unnecessary SQL query that's as simple as `$groupIDs = $familyIDs`?

tractorcow: I'm going to find out why, by messing around with it. :)

src/Security/InheritedPermissions.php
```diff
@@ -610,4 +699,34 @@ protected function getViewerGroupsTable()
...
+     * @return mixed
+     */
+    protected function getCachePermissions($cacheKey)
+    {
+        if (isset($this->cachePermissions[$cacheKey])) {
+            return $this->cachePermissions[$cacheKey];
+        }
+
+        if ($this->cacheService) {
+            return $this->cacheService->get($cacheKey);
```
tractorcow: I think we should have the persistent cache warm the non-persistent cache here. I'll update. :)

tractorcow:
Rebased on 4 now that the prior PR is merged.

@unclecheese I've added a separate commit with what I think is PROBABLY an improvement. Feel free to comment if you want to have a say. :)

src/Security/InheritedPermissionFlusher.php
```diff
@@ -94,9 +94,7 @@ protected function getMemberIDList()
         }
 
         if ($this->owner instanceof Group) {
-            return $this->owner->Members()->exists()
-                ? $this->owner->Members()->column('ID')
-                : null;
+            return $this->owner->Members()->column('ID');
```
tractorcow: Cut down an unnecessary SQL query.

tractorcow:
Can be merged on green:  But don't forget https://github.com/silverstripe/silverstripe-assets/pull/102 and https://github.com/silverstripe/silverstripe-cms/pull/2045 if you merge this.

tractorcow:
Fixed linting issues.

Issue #7549, opened 2017-11-02
TextareaField lacking setMaxLength


zanderwar:
`TextField` has it, but `TextareaField` does not

https://www.w3schools.com/tags/att_textarea_maxlength.asp

Workaround:

```
TextareaField::create('Description', 'Description')->setAttribute('maxlength', 255)
```

but obviously lacking validation behind-the-scenes

## Pull requests

- ~https://github.com/silverstripe/silverstripe-admin/pull/373~
- ~https://github.com/silverstripe/silverstripe-framework/pull/7676~
- [x] https://github.com/silverstripe/silverstripe-admin/pull/374
- [x] https://github.com/silverstripe/silverstripe-framework/pull/7677

flamerohr:
Lowering impact because TextareaField is inherently different - it uses the `Text` type to be created.

This is different to having a `Varchar` type which requires a hard length limit.

tractorcow:
An example where this would be necessary is DBHTMLVarchar, which is intrinsically limited in character size, but is backed by a TextareaField (Which HTMLEditorField extends), so it probably should make use of this API in core.

However, the use case is minor, so we can put it off until 4.1.0.

PR #7677, opened 2017-12-04
Enable max length for textarea field

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/7549

### DIFF
--- a/src/Forms/TextField.php
+++ b/src/Forms/TextField.php
@@ -79,6 +79,13 @@ class TextField extends FormField
         );
     }
 
+    public function getSchemaDataDefaults()
+    {
+        $data = parent::getSchemaDataDefaults();
+        $data['data']['maxlength'] =  $this->getMaxLength();
+        return $data;
+    }
+
     /**
      * @return string
      */
--- a/src/Forms/TextareaField.php
+++ b/src/Forms/TextareaField.php
@@ -134,16 +134,22 @@ class TextareaField extends FormField
      */
     public function getAttributes()
     {
-        return array_merge(
+        $attributes = array_merge(
             parent::getAttributes(),
             array(
                 'rows' => $this->getRows(),
                 'cols' => $this->getColumns(),
                 'value' => null,
                 'type' => null,
-                'maxlength' => $this->getMaxLength(),
             )
         );
+
+        $maxLength = $this->getMaxLength();
+        if($maxLength) {
+            $attributes['maxlength'] = $maxLength;
+        }
+
+        return $attributes;
     }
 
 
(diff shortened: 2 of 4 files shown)


src/Forms/TextField.php
```diff
@@ -79,6 +79,17 @@ public function getAttributes()
         );
     }
 
+    public function getSchemaDataDefaults() {
```
robbieaverill: You'll need to drop the opening brace to a new line - see the linting failures on https://travis-ci.org/silverstripe/silverstripe-framework/jobs/311519327

robbieaverill:
I've fixed the Scrutinizer configuration here: https://github.com/silverstripe/silverstripe-framework/pull/7678

src/Forms/TextField.php
phalkunz: Thanks @robbieaverill. Will fix that.

src/Forms/TextareaField.php
```diff
@@ -45,19 +45,29 @@ class TextareaField extends FormField
...
         $data = parent::getSchemaDataDefaults();
-        
+
         $data['data']['rows'] = $this->getRows();
         $data['data']['columns'] = $this->getColumns();
-        
+
+        $maxLength = $this->getMaxLength();
+        if ($maxLength) {
+            $data['maxLength'] = $maxLength;
```
tractorcow: Top level form schema keys should only be for properties that apply to all fields; maxLength only makes sense to text-based fields.

This should go into `$data['attributes']['maxLength']` or `$data['data']['maxLength']`

src/Forms/TextareaField.php
```diff
@@ -106,6 +116,25 @@ public function getColumns()
...
+    {
+        $this->maxLength = $maxLength;
+
+        return $this;
+    }
+
+    /**
+     * @return null|int
+     */
+    public function getMaxLength()
```
tractorcow: TextField adds this to `getAttributes`. Shouldn't be do this in TextareaField also?

phalkunz:
@tractorcow Thanks for the feedbacks :) I've updated the pull request.

tractorcow:
@phalkunz can you please update the unit tests?

src/Forms/TextareaField.php
```diff
@@ -117,7 +140,8 @@ public function getAttributes()
                 'rows' => $this->getRows(),
                 'cols' => $this->getColumns(),
                 'value' => null,
-                'type' => null
+                'type' => null,
+                'maxlength' => $this->getMaxLength(),
```
tractorcow: We should only set this attribute if non empty.

See TextField.php for pre-existing logic:

```php
if ($maxLength) {
    $attributes['maxLength'] = $maxLength;
    $attributes['size'] = min($maxLength, 30);
}
```

phalkunz: Fixed.

tractorcow:
I've just pushed up a fix to linting.

PR #7678, opened 2017-12-05
 FIX Update branch alias in Scrutinizer config for 4 branch 6f2551e

robbieaverill:
This fix has been made already in 4.0, so cherry picking that commit in and updating the branch alias for the 4 branch.

If there were compatibility fixes made in the 4.0 branch other than this then we should close this PR and merge up instead at some stage.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,7 +1,17 @@
+
 inherit: true
+
+build:
+  environment:
+    variables:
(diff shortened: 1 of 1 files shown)


dhensby:
This is better done by merging 4.0 into 4

dhensby:
fixed with eb55c2712

PR #7682, opened 2017-12-06
Enhancement: Added Array2String function

gorriecoe:
Converts an array or associative array to a string.
Examples:
Array
```
Convert::array2string(array("1","2","3"),',')
```
Outputs "1,2,3"

Associative array
```
Convert::array2string(array("background-color" => "#fff","color" => "#000","font-size" => "12px",),';',':')
```
Outputs "background-color: #fff;color:#000;font-size:12px"

### DIFF
--- a/src/Core/Convert.php
+++ b/src/Core/Convert.php
@@ -173,6 +173,36 @@ class Convert
     }
 
     /**
+     * Convert an array or associative array to a string.
+     *
+     * Examples:
+     * - Convert::array2string(array("1","2","3"),',') => "1,2,3"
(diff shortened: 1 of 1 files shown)


tractorcow:
Hi @gorriecoe this is a good idea, but it feels like something that's more useful as a part of your user code. A good argument for including this in core is to demonstrate other core components that could be improved by using this code, but the use cases seem very specific to project-specific functionality (e.g. generating CSS in php).

Thanks for the contribution and hope you don't feel too put off by us rejecting this feature. =)

PR #7684, opened 2017-12-06
Correcting docs around summary field relabelling

pitchandtone:
Although this is weird that it doesn't use the same logic and searchable fields.

### DIFF
--- a/docs/en/02_Developer_Guides/00_Model/11_Scaffolding.md
+++ b/docs/en/02_Developer_Guides/00_Model/11_Scaffolding.md
@@ -196,9 +196,9 @@ class MyDataObject extends DataObject
     ];
     
     private static $summary_fields = [
-        'Name' => 'Name',
-        'Description.Summary' => 'Description (summary)',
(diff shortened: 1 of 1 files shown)


docs/en/02_Developer_Guides/00_Model/11_Scaffolding.md
```diff
@@ -196,9 +196,9 @@ class MyDataObject extends DataObject
     ];
     
     private static $summary_fields = [
-        'Name' => 'Name',
```
tractorcow: The old behaviour wasn't incorrect; You can use field_labels to override summary_fields, but it's not necessary to.

Did you find the previously documented API didn't actually work?

robbieaverill:
Ping @pitchandtone - keen to get this merged?

kinglozzer:
https://github.com/silverstripe/silverstripe-framework/commit/1fb4c6f39f5252850267e91bacde0b43ca0b3125.

PR #7685, opened 2017-12-08
BUG Using ports different than 80/443 within alternate_base_url

marcosbc:
Currently, specifying a URL with a port number in `alternate_base_url` different than 80 (http) or 443 (https), will make the `Director` class not to detect this and break most URLs in the application.

Note that in our case, the URL must be specified in most cases when SilverStripe is working on a URL prefix, e.g. `/silverstripe`.

Steps to reproduce:

- Change the frontend's HTTP port to something different than 80.
- Set `alternate_base_url` for SilverStripe to use this new port:

```
SilverStripe\Control\Director:
  alternate_base_url: http://HOSTNAME:PORT/silverstripe
```

Next, if you access SilverStripe, you will see that most URLs are wrong and stylesheets are not loading.

For instance, by setting the URL `http://127.0.0.1:8080/silverstripe`, the `<base>` tag appears as following:

```
<base href="http://127.0.0.1/silverstripe/"><!--[if lte IE 6]></base><![endif]-->
```

And no mentions of the port `8080`.

### DIFF
--- a/src/Control/Director.php
+++ b/src/Control/Director.php
@@ -468,8 +468,9 @@ class Director implements TemplateGlobalProvider
         if ($baseURL = self::config()->get('alternate_base_url')) {
             $baseURL = Injector::inst()->convertServiceProperty($baseURL);
             $host = parse_url($baseURL, PHP_URL_HOST);
+            $port = parse_url($baseURL, PHP_URL_PORT);
             if ($host) {
-                return $host;
+                return $port ? "$host:$port" : $host;
             }
         }
 
@@ -487,8 +488,9 @@ class Director implements TemplateGlobalProvider
         if ($baseURL = self::config()->uninherited('default_base_url')) {
             $baseURL = Injector::inst()->convertServiceProperty($baseURL);
             $host = parse_url($baseURL, PHP_URL_HOST);
+            $port = parse_url($baseURL, PHP_URL_PORT);
             if ($host) {
-                return $host;
+                return $port ? "$host:$port" : $host;
             }
         }
 


kinglozzer:
It seems reasonable to support this IMO. My initial thought is that perhaps the port should be set/resolved separately by a new method (e.g. `Director::port()`). `Director::host()` is designed to return the hostname, and the port number isn’t really part of that...

I’ll await input from those who deal with this class more often than I do before suggesting you make that change ;)

tractorcow:
Yes it seems to be a real problem:

 - Add `Director::port()`
 - Ensure `Director::protocolAndHost()` includes port. (seems to be more correct than omitting in > 50% of cases). This will ensure that `Director::absoluteURL` respects port.
 - Ensure `Director::absoluteBaseURLWithAuth()` includes port.

tractorcow:
To respond to this direct PR; I'm not sure we can safely include the port number in `host()` method return. It would break code which relied on the pure-hostname being returned from this as a low-level method.

marcosbc:
Hi, thanks for checking, I'll try to find some time in the next couple of days to look into improving this.

marcosbc:
After checking this, it seems that some refactoring might be needed, as `Director::host` currently follows this logic:

1. It tries to make use of the `alternate_base_url` metadata: OK, the host is extracted safely.
1. Fallback to the HTTP `Host` header if set: **Not OK**, it might include a port.
1. Fallback to `default_base_url` metadata if set: OK, the host is extracted safely.
1. Fallback to `$_SERVER['HTTP_HOST']` if set: **Not OK**, it might include a port.
1. Final fallback to `$_SERVER['SERVER_NAME']` if set, or `gethostname()` if not: OK, these don't include the port.

If the `port` method were to be implemented, we should ensure that `host` doesn't include the port in any of the above cases.
Also, the new method shouldn't return a port value if using the default port for HTTP or HTTPS (80 or 443, respectively).

tractorcow:
@marcosbc looking into this, it seems like more often than not, Director::host() includes the port number. The reason is that the `$_SERVER['HTTP_HOST']` includes it, so it will be present in most circumstances.

Given that, I've decided that host() should probably keep the port. Instead I'll add port() (to get port by itself) and hostName() to get the host without the port.

In addition, I'll tweak host() so that it consistently adds the port when being overridden.

Those should be the only tweaks needed. :)

tractorcow:
Replaced with https://github.com/silverstripe/silverstripe-framework/pull/7780

tractorcow:
yay fixed and merged.

Issue #6890, opened 2017-05-08
RFC: Exclude tests & docs from dist packages


sminnee:
## Overview

Ignoring `.git`, this is a breakdwon

```
 96K	_config
4.0K	_config.php
4.0K	_register_database.php
8.0K	cli-script.php
 36K	client
4.0K	composer.json
4.0K	CONTRIBUTING.md
8.7M	docs
632K	lang
4.0K	LICENSE
4.0K	main.css
 12K	main.php
4.0K	main.php5
4.0K	phpunit.xml.dist
4.0K	README.md
4.0K	sake
  0B	silverstripe_version
4.1M	src
272K	templates
4.7M	tests
360K	thirdparty
4.0K	web.config
 19M	total
```

70% of the files come from these two items, neither of which are necessary to run SilverStripe, nor to develop with it.

 * 8.7M	docs
 * 4.7M	tests

Furthermore, it encourages the practise of splitting heavyweight tests into other modules. It would be good to consider, for example, whether frameworktest should be test fixtures in the framework module. But forcing every user to download this module seems inappropriate. The underlying issue that i that 

Documentation is arguably of assistance to developers, but I question whether reading local markdown files from `framework/docs/en/` is a sufficiently superior experience to reading it on doc.silverstripe.org or github.com that we should double the download size of every user in order to provide it.

## Acceptance Criteria

 * Exclude all doc sources from core package via gitattributes
 * Exclude all test sources from core package via gitattributes
 * Ensure that combined test builds (such as what we're planning in installer) run a --prefer-source install
 * Ensure that our documentation generator runs a --prefer source install
 * Silverstripe.org/download follows these practices (does not contain extraneous files)

## Alternatives

The most common counter-proposal as been a post-install script that deletes tests and/or docs, although that only solves disk size, not download size, and it's an order-of-magnitude more difficult (compared to a few lines in .gitattributes)

## External discussion

Some discussion of these idea around the internet; it's controversial and it looks like Symfony started doing this but changed their mind. Seldaek (composer lead) initially spoke in support of leaving in tests and then later pulled back to a "there are arguments to both approaches, I don't want to promote one approach over the other"

 * https://www.reddit.com/r/PHP/comments/2jzp6k/i_dont_need_your_tests_in_my_production/
 * https://github.com/composer/composer/issues/1750#issuecomment-76601575
 * https://github.com/symfony/symfony/issues/6605#issuecomment-60309257
[...]

PR #7692, opened 2017-12-11
Feature Set .gitattributes to ignore docs and tests folder, and …

flamerohr:
some source asset files which are not needed

Parent issue: https://github.com/silverstripe/silverstripe-framework/issues/6890


tractorcow:
Can be merged on green.

PR #7693, opened 2017-12-12
Added after priority to replace default authenticator.

gorriecoe:
The documentation needs to include the after priority in order to override the default authenticator.  Currently there is no clear indication of this and without it, coresecurity will override your changes.


gorriecoe:
@robbieaverill Updated =)

PR #7694, opened 2017-12-12
FIX Use Injector to retrieve the current session

robbieaverill:
This PR switches from getting the session from the current controller to using one from an HTTPRequest service via the injector. This is important if this class is used by middleware since no controller is registered at this point, but `Director::handleRequest` registers an HTTPRequest which can be used instead.

Fixes https://github.com/silverstripe/silverstripe-spellcheck/issues/33

### DIFF
--- a/src/Security/SecurityToken.php
+++ b/src/Security/SecurityToken.php
@@ -2,11 +2,13 @@
 
 namespace SilverStripe\Security;
 
+use Exception;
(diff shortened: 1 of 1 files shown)


robbieaverill:
Update pushed to allow `Controller::curr()` to return a request as well, since a lot of the test suite is depending on it at the moment.

Issue #7359, opened 2017-09-11
Allow $owns on non-Versioned objects.


sminnee:
Although we introduced it for better versioning support, "owns" can have semantics beyond that, and I don't think we need to limit its application to objects using Versioned.

The main benefit is being able to handle the case of `SiteTree ---owns--> UnversionedObject ---owns--> File` can be configured to ensure that the File is published with the SiteTree. This is a serious bug that occurs frequently when upgrading SS3 -> SS4, and the only current fix is to add versioning to UnversionedObject, which I agree is the "best" fix, but it's quite invasive and versioning in 4.0 might not be _quite_ polished enough for this to be free of annoying side-effects.

In addition, owns can be used to duplicate records within has_many relations, and duplicate links to records specified in many_many relations, without special duplicate overrides. Note that has_ones already work, and the actual records in has_one / many_many shouldn't need to be duplicated, only the links to them.

# Acceptance Criteria

 - publishRecursive() is supported on all objects, and will publish all owned objects regardless of whether any are versioned. This necessitates that all objects can be added to a changeset (even if not versioned)
 - All config and methods related to ownership (e.g. `owns`, `owned_by`, `findOwned`, `findOwners`, `findOwnersRecursive`
 - Non-versioned objects are visible in the campaign admin area when added to a changeset
 - The default "Save" action for model admin and gridfield for un-versioned objects will publish all owned objects
 
# Exclusions:

- UX for interface of unversioned objects. See https://github.com/silverstripe/silverstripe-versioned/issues/71
- This functionality isn't necessary in absentia of the versioned module.

# PRs

 - [x] https://github.com/silverstripe/silverstripe-versioned/pull/74
 - [x] https://github.com/silverstripe/silverstripe-framework/pull/7695 (docs update)

tractorcow:
I think we need to be careful, since publishing a versioned object adds this object + all owned objects to a changeset.

We may need to allow non-versioned objects to be added to changesets to support this feature.

chillu:
This just came up on the `#ss4` Slack again, but in the variation of `UnversionedObject ---owns--> File`. In this case it was `TeamMember ----has_one/owns--> ProfilePhoto`, which isn't what I'd consider a "publishable" entity. That'll confuse the hell out of many devs (including myself). I think if you declare ownership, you're stating a preference for these dependencies to be published as a developer. 

[...]

PR #7695, opened 2017-12-12
Update framework docs / phpdoc for versioning changes

tractorcow:
Parent story: https://github.com/silverstripe/silverstripe-framework/issues/7359

Also adds a necessary extension point to GridFieldDetailForm_ItemRequest

### DIFF
--- a/docs/en/02_Developer_Guides/00_Model/10_Versioning.md
+++ b/docs/en/02_Developer_Guides/00_Model/10_Versioning.md
@@ -27,22 +27,23 @@ use SilverStripe\ORM\DataObject;
 class MyStagedModel extends DataObject 
 {
     private static $extensions = [
-        Versioned::class
+        Versioned::class,
     ];
 }
 ```
 
 Alternatively, staging can be disabled, so that only versioned changes are tracked for your model. This
-can be specified by setting the constructor argument to "Versioned"
-
+can be specified by using the `.versioned` service variant that provides only version history, and no
+staging.
 
 ```php
 use SilverStripe\ORM\DataObject;
+use SilverStripe\Versioned\Versioned;
 
 class VersionedModel extends DataObject 
 {
     private static $extensions = [
-        "SilverStripe\\ORM\\Versioning\\Versioned('Versioned')"
+        Versioned::class . '.versioned',
     ];
 }
 ```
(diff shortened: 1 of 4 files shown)


docs/en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -3,13 +3,13 @@ summary: Add versioning to your database content through the Versioned extension
 
 # Versioning
 
-Database content in SilverStripe can be "staged" before its publication, as well as track all changes through the 
+Database content in SilverStripe can be "staged" before its publication, as well as track all changes through the
```
flamerohr: I'd be careful of trimming trailing whitespaces in markdown - as two spaces translate to a `<br>` in html

docs/en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -308,7 +309,19 @@ class Banner extends Page
...
+
+Ownership can be used with non-versioned dataobjects, as the necessary functionality is included by default
+by the versioned object through the `[api:SilverStripe\Versioned\RecursivePublishable]` extension which is
+applied to all objects.
+
+However, it is important to note that even when saving un-versioned objects, it is necessary to use
+`->publishRecursive()` to trigger a recursive publish.
+
+As long as the `owns` config is specified on all elements in the tree, you can use any combination of
+versioned or unversioned dataobjects.
```
flamerohr: Does this mean that we don't need to worry about `->publishRecursive()` if we have an `owns` defined?

Maybe elaborate that the `owns` config behaves in similar ways to the `owns` config in a versioned object?

docs/en/02_Developer_Guides/00_Model/10_Versioning.md
tractorcow: Well we have no need for a `<br>` here so it should be ok.

flamerohr: yeah, I don't want to spend time going through all the whitespace changes to see if you did end up trimming a double space at the end :P

it's pretty much "please don't trim for markdown"

flamerohr:
Thanks for reverting the trimming in markdown

Can merge on green.

docs/en/02_Developer_Guides/00_Model/10_Versioning.md
dhensby: > I'd be careful of trimming trailing whitespaces in markdown - as two spaces translate to a <br> in html

TBH, if people want a break, they should be using a double new line not two spaces - MD should be readable in both it's raw form and it's rendered form.

flamerohr: that's true, but two new lines means a new paragraph block right?

PR #7698, opened 2017-12-13
API Shift basic auth checking into middleware

### LINKED ISSUE #7554: Move BasicAuth::protect_site_if_necessary() into Middleware
Currently Controller is responsible for invoking this action. Similar to forceSSL / forceWWW, this should be handled via middleware.

```php
protected function init()
    {
        if ($this->basicAuthEnabled) {
            BasicAuth::protect_site_if_necessary();
        }

        // This is used to test that subordinate controllers are actually calling parent::init() - a common bug
        $this->baseInitCalled = true;
    }
```

# PRs

 - [x] https://github.com/silverstripe/silverstripe-framework/pull/7698

tractorcow:
Fixes #7554

Per-controller basicauth configuration is replaced with a route-matching rule applied to middleware.

### DIFF
--- a/_config/security.yml
+++ b/_config/security.yml
@@ -26,9 +26,13 @@ SilverStripe\Core\Injector\Injector:
     properties:
       Middlewares:
         AuthenticationMiddleware: '%$SilverStripe\Security\AuthenticationMiddleware'
+        BasicAuthMiddleware: '%$SilverStripe\Security\BasicAuthMiddleware'
   SilverStripe\Security\AuthenticationMiddleware:
     properties:
       AuthenticationHandler: '%$SilverStripe\Security\AuthenticationHandler'
+  SilverStripe\Security\BasicAuthMiddleware:
+    properties:
+      URLPatterns: []
   SilverStripe\Security\Security:
     properties:
       Authenticators:
--- a/src/Control/Controller.php
+++ b/src/Control/Controller.php
@@ -4,8 +4,10 @@ namespace SilverStripe\Control;
 
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Dev\Debug;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\ORM\FieldType\DBHTMLText;
 use SilverStripe\Security\BasicAuth;
+use SilverStripe\Security\BasicAuthMiddleware;
 use SilverStripe\Security\Member;
 use SilverStripe\Security\Security;
 use SilverStripe\View\SSViewer;
@@ -61,6 +63,8 @@ class Controller extends RequestHandler implements TemplateGlobalProvider
     protected $templates = [];
 
     /**
+     * @deprecated 4.1.0...5.0.0 Add this controller's url to
+     * SilverStripe\Security\BasicAuthMiddleware.URLPatterns injected property instead of setting false
      * @var bool
      */
     protected $basicAuthEnabled = true;
@@ -98,6 +102,7 @@ class Controller extends RequestHandler implements TemplateGlobalProvider
      */
     protected function init()
     {
+        // @todo This will be removed in 5.0 and will be controlled by middleware instead
         if ($this->basicAuthEnabled) {
             BasicAuth::protect_site_if_necessary();
         }
@@ -528,9 +533,16 @@ class Controller extends RequestHandler implements TemplateGlobalProvider
      * Call this to disable site-wide basic authentication for a specific controller. This must be
      * called before Controller::init(). That is, you must call it in your controller's init method
      * before it calls parent::init().
+     *
+     * @deprecated 4.1.0...5.0.0 Add this controller's url to
+     * SilverStripe\Security\BasicAuthMiddleware.URLPatterns injected property instead of setting false
      */
     public function disableBasicAuth()
     {
+        Deprecation::notice(
+            '5.0',
+            'Add this controller\'s url to ' . BasicAuthMiddleware::class . '.URLPatterns injected property instead'
+        );
         $this->basicAuthEnabled = false;
     }
 
--- a/src/Security/BasicAuthMiddleware.php
+++ b/src/Security/BasicAuthMiddleware.php
@@ -1,0 +1,118 @@
+<?php
+
+namespace SilverStripe\Security;
+
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
+use SilverStripe\Control\HTTPResponse_Exception;
+use SilverStripe\Control\Middleware\HTTPMiddleware;
+use SilverStripe\Control\Middleware\CanonicalURLMiddleware;
+
+class BasicAuthMiddleware implements HTTPMiddleware
+{
+    /**
+     * URL Patterns for basic auth. Keys are the Regexp string to match, and the key can
+     * be one of the below:
+     *  - true (bool) - Enabled for this url
+     *  - false (bool) - Disabled for this url
+     *  - Any string / array - Enabled for this url, and require the given string as a permission code
+     *  - null (default) - Calls BasicAuth::protect_site_if_necessary(), which falls back to config setting
+     *
+     * E.g.
+     * [
+     *   '#^home#i' => false,
+     *   '#^secure#i' => true,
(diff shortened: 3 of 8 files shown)


dhensby:
How do you force basic auth on a controller who's URL is dynamic (ie: a content controller)?

tractorcow:
You manually put the basic auth call into your controller if you need to.

dhensby:
I think it'd be better to match basic auth against Controllers rather than URL patterns. Or at least both. but this seems a bit of a step backwards if you can't protect a controller through the middleware...

src/Control/Controller.php
```diff
@@ -528,9 +532,16 @@ public function render($params = null)
...
      * before it calls parent::init().
+     *
+     * @deprecated 4.0..5.0
      */
     public function disableBasicAuth()
     {
+        Deprecation::notice(
+            '5.0',
+            'Add this controller\'s url to SilverStripe\Security\BasicAuthMiddleware.ExcludeURLPatterns'
+            . ' property instead'
```
unclecheese: Can we avoid the use of a hardcoded class? I'd prefer `BasicAuthMiddleware::class . 'ExcludeURLPatterns'`

src/Security/BasicAuthMiddleware.php
```diff
@@ -0,0 +1,163 @@
...
+     * @see CanonicalURLMiddleware
+     * @var array
+     */
+    protected $excludeURLPatterns = [];
+
+    /**
+     * URL to include, even if entire_site_protected is false.
+     * The list can either be an array of patterns to check, or an associative array
+     * of patterns mapped to either a permission code, or array of codes to check.
+     *
```
unclecheese: This is a bit of a complex spec to describe with written docs only. As a dev using this for the first time, I'd much appreciate an example here, e.g.

```
[
  'my-secret-page' => 'ADMIN',
  '/^members-only\/*$',
]
```

_config/security.yml
```diff
@@ -26,9 +26,14 @@ SilverStripe\Core\Injector\Injector:
...
       Middlewares:
         AuthenticationMiddleware: '%$SilverStripe\Security\AuthenticationMiddleware'
+        BasicAuthMiddleware: '%$SilverStripe\Security\BasicAuthMiddleware'
   SilverStripe\Security\AuthenticationMiddleware:
     properties:
       AuthenticationHandler: '%$SilverStripe\Security\AuthenticationHandler'
+  SilverStripe\Security\BasicAuthMiddleware:
+    properties:
+      ExcludeURLPatterns: []
+      IncludeURLPatterns: []
```
unclecheese: Just curious, what if I want my project to explicitly include a path that is excluded by a module, or vice-versa? Should we enforce an associative array here instead?


_security.yml_
```
SilverStripe\Security\BasicAuthMiddleware:
  properties:
    ExcludeURLPatterns:
        'my-path': true
        'my-other-path': 'ADMIN'
```

_mymodule.yml_
```
SilverStripe\Security\BasicAuthMiddleware:
  properties:
    ExcludeURLPatterns:
        'my-path': false
```

tractorcow: ok, so we use the bool flag for "enabled" right? And if it's a string, it's a permission check (except for exclude, which doesn't make sense to check for permissions).

tractorcow: Hm, it's a bit ambiguous if both key and value are strings; Someone might try to put the pattern as the value, and use a named key instead (like we do for extensions sometimes).

tractorcow: Ok, let's combine into one value; Keys are patterns, and value are:
 - bool true - force on (no permissions)
 - string - force on (special permission)
 - bool false - force exclude
 - null - default to global setting (controlled via yml)

tractorcow:
```php
/**
     * URL Patterns for basic auth. Keys are the Regexp string to match, and the key can
     * be one of the below:
     *  - true (bool) - Enabled for this url
     *  - false (bool) - Disabled for this url
     *  - Any string / array - Enabled for this url, and require the given string as a permission code
     *  - null (default) - Calls BasicAuth::protect_site_if_necessary(), which falls back to config setting
     *
     * @see CanonicalURLMiddleware
     * @var array
     */
    protected $urlPatterns = [];
```

src/Security/BasicAuthMiddleware.php
tractorcow: I've simplified this to a single array mapping patterns to rules.

tractorcow:
> I think it'd be better to match basic auth against Controllers rather than URL patterns. Or at least both. but this seems a bit of a step backwards if you can't protect a controller through the middleware...

We can't do that in the middleware because there is no controller at this point. The controller is constructed, but it's not registered anywhere the middleware can access. Trying to make middleware handler-aware isn't really possible.

If you need this in your controller, just put it in your `MyController::init()` manually and skip middleware.

tractorcow:
Updated based on feedback @unclecheese . I think the single-array config is a lot simpler and easier to understand.

_config/security.yml
unclecheese: Brilliant. Love this change.

dhensby:
> We can't do that in the middleware because there is no controller at this point. The controller is constructed, but it's not registered anywhere the middleware can access. Trying to make middleware handler-aware isn't really possible.
> 
> If you need this in your controller, just put it in your MyController::init() manually and skip middleware.

ok

dhensby:
Behat tests are failing

src/Control/Controller.php
```diff
@@ -61,6 +63,8 @@ class Controller extends RequestHandler implements TemplateGlobalProvider
     protected $templates = [];
 
     /**
+     * @deprecated 4.1.0...5.0.0 Add this controller's url to
+     * SilverStripe\Security\BasicAuthMiddleware.ExcludeURLPatterns property instead of setting false
```
flamerohr: old property still suggested `urlPatterns` is the new one right?

src/Control/Controller.php
```diff
@@ -528,9 +533,15 @@ public function render($params = null)
...
      * called before Controller::init(). That is, you must call it in your controller's init method
      * before it calls parent::init().
+     *
+     * @deprecated 4.0..5.0
      */
     public function disableBasicAuth()
     {
+        Deprecation::notice(
+            '5.0',
+            'Add this controller\'s url to ' . BasicAuthMiddleware::class . '.URLPatterns property instead'
```
flamerohr: does casing matter? I see `urlPatterns` and `URLPatterns` used interchangeably, could we keep it to one case?

src/Security/BasicAuthMiddleware.php
```diff
@@ -0,0 +1,118 @@
...
+     * @param HTTPRequest $request
+     * @return bool|string|array|null boolean value if enabled/disabled explicitly for this request,
+     * or null if should fall back to config value. Can also provide an explicit string / array of permission
+     * codes to require for this requset.
+     */
+    protected function checkMatchingURL(HTTPRequest $request)
+    {
+        // Null if no permissions enabled
+        $patterns = $this->getUrlPatterns();
```
flamerohr: mixed casing for method names, try to keep to one :)

src/Control/Controller.php
tractorcow: oops sorry.

src/Control/Controller.php
tractorcow: `URLPatterns` is public setter `$middleware->URLPatterns`. `$urlPatterns` is the protected var. :)

src/Security/BasicAuthMiddleware.php
tractorcow: oops yes

tractorcow:
Clarified PHPDocs and fixed case.

src/Control/Controller.php
flamerohr: oh, you're right :D nice one

tractorcow:
Need to fix those behat tests sorry. =(

tractorcow:
I missed a really obvious bug... you can't call BasicAuth::protect_site_if_necessary() when a controller isn't available. The very same restriction I just described to @dhensby ...

PR #7699, opened 2017-12-13
ENHANCEMENT Allow html in security failure message

phalkunz:
Issue: https://docs.silverstripe.org/en/contributing/code/

PR #7702, opened 2017-12-14
BUG Fix message casting for html security messages

tractorcow:
Fixes message casting and adds some tests.

Follow up from https://github.com/silverstripe/silverstripe-framework/pull/7699


tractorcow:
I should have noted that the bug was passing in the DBField to the setMessage() method, without extracting the inner text value. :)

PR #7704, opened 2017-12-14
BUG Do database migrations before default records

### LINKED ISSUE #7703: SS3 → SS4 `requireDefaultRecords`
This issue is specific to **page class name mapping** in **4.0.0**.

When migrating from SS3 to SS4, if your page instances are now namespaced, you can't do a `MyPage::get()` from within `requireDefaultRecords` anymore (it will always return an empty list), since `ClassName` does not yet contain the namespace. There is an upgrade mapping that you can do, such as the following https://github.com/nglasl/silverstripe-extensible-search/blob/master/_config/upgrade.yml (which maps your existing class names to include the namespace), however this is done *after* the `requireDefaultRecords`.

So for example, if you're creating a page should one not exist, this can result in the same page being created twice (during a SS3 to SS4 migration). This is my workaround https://github.com/nglasl/silverstripe-extensible-search/blob/master/src/pages/ExtensibleSearchPage.php#L105.

It seems like the class name mapping should happen *before* the database records are generated.

Thoughts?

tractorcow:
Fixes #7703

### DIFF
--- a/src/ORM/DatabaseAdmin.php
+++ b/src/ORM/DatabaseAdmin.php
@@ -318,22 +318,6 @@ class DatabaseAdmin extends Controller
                 }
             }
 
-            foreach ($dataClasses as $dataClass) {
-                // Check if class exists before trying to instantiate - this sidesteps any manifest weirdness
-                // Test_ indicates that it's the data class is part of testing system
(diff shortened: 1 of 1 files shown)


tractorcow:
Error while getting log =(

tractorcow:
Rebased and checked linting.

PR #7705, opened 2017-12-14
BUG Fix incorrect BASE_DIR inferred in CLI

### LINKED ISSUE #7689: Updating tests is a mild goose chase when supplying Director::test with urls relative to base.
https://github.com/silverstripe/silverstripe-framework/blob/d57dea03184aec636ab7af5f0d81be2b7cb03693/src/Control/Director.php#L259

This line always prepends e.g. `/home/users/moduleupgrader/public_html/silverstripefour/testsite`, which then doesn't match any director rule, as they're all relative to the base (as is sensible). This results in every single call to `Director::test` fail with `404` (unless the test expects a 404).

This is worked around by defining an `Director.alternative_base_url` in the config during `setUp`, but even then issues seemed to arise between whether or not to include a trailing slash on that base, and requires a test/upgrading developer to prefix all URIs with the alternative base (if there are assertions to check `Location` header for example).

While this is ok, and can be mitigated, I certainly didn't find it very helpful, and ultimately spent far too long trying to figure out why every test failed where I couldn't see a reason for it (it seems in the past that relative URIs worked as expected). To this end I personally don't think that test URIs should be prepended with the base, as in the CLI (where tests now always run since the removal of the `dev/tests` controller) will always be relative to root, which seems a bit unnecessary really.

Or perhaps I'm missing the vision of some compelling reason? If so, please share.
At the least I think some extra docs might be nice, as it's a gotcha that certainly gotme.

# PR:

 - [x] https://github.com/silverstripe/silverstripe-framework/pull/7705

tractorcow:
Also fixes Director::mockRequest() mocking incorrect $url
Fixes #7689

### DIFF
--- a/src/Control/Director.php
+++ b/src/Control/Director.php
@@ -263,7 +263,7 @@ class Director implements TemplateGlobalProvider
         $newVars = HTTPRequestBuilder::cleanEnvironment($newVars);
 
         // Create new request
-        $request = HTTPRequestBuilder::createFromVariables($newVars, $body);
+        $request = HTTPRequestBuilder::createFromVariables($newVars, $body, ltrim($url, '/'));
         if ($headers) {
             foreach ($headers as $k => $v) {
                 $request->addHeader($k, $v);
--- a/src/Control/HTTPRequestBuilder.php
+++ b/src/Control/HTTPRequestBuilder.php
@@ -28,16 +28,20 @@ class HTTPRequestBuilder
      *
      * @param array $variables
      * @param string $input Request body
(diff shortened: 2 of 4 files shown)


dhensby:
I'd prefer this to go into the 4 branch given the new API and people suffering from this issue can fix it locally by defining SS_BASE_URL as they should be.

tractorcow:
Sounds good.

tractorcow:
I think those linting issues are from the updates to codestyle checker. I'll fix. :)

tractorcow:
Rebased and pushed.

PR #7714, opened 2017-12-17
Returns chainability in setValue from parent

zanderwar:
Returns the chainability to setValue() and removes unnecessary `else` statements

### DIFF
--- a/src/ORM/FieldType/DBCurrency.php
+++ b/src/ORM/FieldType/DBCurrency.php
@@ -17,7 +17,6 @@ namespace SilverStripe\ORM\FieldType;
  */
 class DBCurrency extends DBDecimal
 {
-
     /**
      * @config
      * @var string
@@ -38,9 +37,9 @@ class DBCurrency extends DBDecimal
         $val = $this->config()->currency_symbol . number_format(abs($this->value), 2);
         if ($this->value < 0) {
             return "($val)";
-        } else {
-            return $val;
         }
+
+        return $val;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/ORM/FieldType/DBCurrency.php
```diff
@@ -62,9 +61,11 @@ public function setValue($value, $record = null, $markChanged = true)
         if (is_numeric($value)) {
             $this->value = $value;
         } elseif (preg_match('/-?\$?[0-9,]+(.[0-9]+)?([Ee][0-9]+)?/', $value, $matches)) {
-            $this->value = str_replace(array('$',',',$this->config()->currency_symbol), '', $matches[0]);
-        } else {
-            $this->value = 0;
+            $this->value = str_replace(['$', ',', $this->config()->currency_symbol], '', $matches[0]);
         }
+
+        $this->value = 0;
```
robbieaverill: This method doesn't return inside each condition, so this would override every value with zero. If you could put the else clause back this looks like a good change (as `DBField::setValue` returns `$this` so it's a fix rather than an enhancement) 😃

robbieaverill:
If you felt like it, here are some other DBField subclasses that aren't chainable which could also be fixed:

* DBPolymorphicForeignKey
* DBForeignKey
* DBClassName
* DBPrimaryKey

src/ORM/FieldType/DBCurrency.php
zanderwar: Got me

tractorcow:
@zanderwar can you please fix this up and I'll merge.

dhensby:
Looks like the downstream repo has been deleted so I've merged this by hand

tractorcow:
Merged version looks good.

Issue #7721, opened 2017-12-19
ENHANCEMENT Ensure minimal bootstrap files for tests are included in --prefer-dist


tractorcow:
## Affected Version

SilverStripe 4.1.x-dev

## Description

Installing framework with `--prefer-dist` doesn't include bootstrap.php, meaning that all modules that run tests (even if they don't run framework tests) either need to inline bootstrap.php, or must install with `--prefer-source` which is slow.

## Steps to Reproduce

`composer require silverstripe/framework 4.1.x-dev --prefer-dist` in a module

Run the tests.

Notice that `bootstrap.php` is non-existant and fails.

robbieaverill:
Can we move the bootstrapping out into a separate module which is included in the installer via require-dev, or a require-dev'd development recipe?

tractorcow:
if we are going that far maybe we just inline it. it's not really a "standard" bootstrap, and is more a "default works for most cases" one.

robbieaverill:
I think it _should_ be considered the standard bootstrap for SS4 now, especially since we've been promoting it as "the" way to run unit tests both in [local tests](https://github.com/silverstripe/silverstripe-installer/blob/4.0/phpunit.xml.dist#L19) and in Travis configuration. 

We've moved away from including a whole package (travis support) to something which is almost just a composer install and go, but we can't feasibly inline this bootstrap script in every single module including the community.

The change will obviously speed up production deployments, but I think for developers and for CI systems they'll probably be installing from source anyway so this will just cause problems.

tractorcow:
CMS has a separate bootstrap that has page.php specific setup.

Maybe it's time to come up with a new bootstrap system?

robbieaverill:
`vendor/autoload.php` would be ideal, but we're not there yet

flamerohr:
The whole bootstrapping could be separated out to a testing recipe, I think, whether that's a part of the recipe or a module included by the recipe is up for debate :)

robbieaverill:
How about a compromise?

```diff
- /tests export-ignore
+ /tests/php export-ignore
+ /tests/behat export-ignore
```

tractorcow:
My recommendation is:
 - Simplify bootstrap.php greatly; I don't think many of the things that happen in that file are necessary. The things it does are already handled by other core functions:
   * let framework or installer bootstrap /assets instead
   * let CLIRequestBuilder setup globals instead
   * Let constants.php setup constants instead
[...]

PR #7722, opened 2017-12-19
FIX Amend rules to ignore test files only

flamerohr:
Short term fix for https://github.com/silverstripe/silverstripe-framework/issues/7721
Follow up to the work done in https://github.com/silverstripe/silverstripe-framework/pull/7692


tractorcow:
Merge on green.

PR #7723, opened 2017-12-20
BUG Fix basic auth in PHP-CGI

### LINKED ISSUE #7717: Authorization issues with PHP-CGI
This came about while trying to set BasicAuth in the GraphQL module, but applies to general BasicAuth (otherAuths?) - see the discussion on Slack#graphql with @robbieaverill, who suggested  I might create this issue. Core, documentation or nothing? :)

## Affected Version
3.x, 4.x

## Description
With PHP_CGI, the Authorization header isn't set properly: It is set to REDIRECT_HTTP_AUTHORIZATION instead of HTTP_AUTHORIZATION, and this means that PHP_AUTH_USER and PHP_AUTH_PW, that are calculated based on HTTP_AUTHORIZATION,  aren't set. Thes $_SERVER variables are used in class BasicAuth.

There are multiple solutions thinkable to this problem, some of which could live in core, and others coudn't but could possibly be documented? At least it took me a while to figure this out and find a solution that worked for me. For instance:

On MAMP this worked in index.php:

    if (!empty($_SERVER['REDIRECT_HTTP_AUTHORIZATION'])) {
        $_SERVER['HTTP_AUTHORIZATION'] = $_SERVER['REDIRECT_HTTP_AUTHORIZATION'];
    }

And this in httpd.conf (but not in htaccess):
   
    SetEnvIf Authorization “(.*)” HTTP_AUTHORIZATION=$1

## Steps to Reproduce
PHP_CGI :)

Martine

tractorcow:
Fixes #7717

### DIFF
--- a/tests/php/Control/HTTPRequestBuilderTest.php
+++ b/tests/php/Control/HTTPRequestBuilderTest.php
@@ -52,7 +52,6 @@ class HTTPRequestBuilderTest extends SapphireTest
         ];
         $this->assertEquals($headers, HTTPRequestBuilder::extractRequestHeaders($request));
 
-
         $request = [
             'PHP_AUTH_USER' => 'admin',
             'PHP_AUTH_PW' => 'password',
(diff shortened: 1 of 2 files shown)


tests/php/Control/HTTPRequestBuilderTest.php
```diff
@@ -62,5 +61,14 @@ public function testExtractRequestHeadersBasicAuth()
...
         $this->assertEquals($headers, HTTPRequestBuilder::extractRequestHeaders($request));
+        
+        $request = [
+            'REDIRECT_HTTP_AUTHORIZATION' => 'Basic YWRtaW46cGFzc3dvcmQ=',
+        ];
+        $headers = [
+            'PHP_AUTH_USER' => 'admin',
+            'PHP_AUTH_PW' => 'password',
+        ];
+        $this->assertEquals($headers, HTTPRequestBuilder::extractRequestHeaders($request));
```
flamerohr: can we add a case for when both `Authorization` and `REDIRECT_HTTP_AUTHORIZATION` are set?

I know it's not common, but sets expectations

tractorcow:
Ok done @flamerohr

PR #7725, opened 2017-12-21
FIX DBTime timestamps should return the time in seconds

dhensby:
To me it is unexpected that the `getTimestamp` method on `DBTime` returns the timestamp of that time as of "today" when the application is running, rather than the time represented in seconds.

### DIFF
--- a/src/ORM/FieldType/DBTime.php
+++ b/src/ORM/FieldType/DBTime.php
@@ -185,7 +185,8 @@ class DBTime extends DBField
     public function getTimestamp()
     {
         if ($this->value) {
-            return strtotime($this->value);
+            $now = DBDatetime::now()->getTimestamp();
+            return strtotime($this->value, $now) - $now;
         }
         return 0;
     }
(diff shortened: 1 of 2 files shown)


dhensby:
hmm - well this didn't work :(

tractorcow:
I don't understand. It's as of "seconds since 1970" not "as of today". Although I'm not sure what the "now" relevance for this method is.

dhensby:
> I don't understand. It's as of "seconds since 1970" not "as of today". Although I'm not sure what the "now" relevance for this method is.

At the moment it returns the time stamp for the time as of "today".

`strtotime('01:00:00');` this code does not return 3600 (1 hour in seconds) it returns the time stamp for 1am on the current day according to the server (1513922400).

IMO we should be returning the time in seconds because there is no reference to a date in the `DBTime` class.

If you (@tractorcow) agree, I can reopen and fix this, but I spoke with @kinglozzer on slack and he wasn't convinced it should be returning seconds instead of a timestamp.

tractorcow:
Oh I see, that makes sense. Sorry for initial misunderstanding.

tractorcow:
I'm not 100% convinced it should be definitely one or the other... it's possible this method isn't relevant to be called at all on DBTime (without a date).

PR #7726, opened 2017-12-21
FIX Allow HTML 5 input tags in FunctionalTest form submissions

dhensby:
At the moment if you write a `FunctionalTest` to submit a `Form` which has `HTML5` `<input>` tags the form will fail to submit because `simpletest/form` doesn't recognise the newer HTML5 input types.

This PR allows any non-recognised inputs to fallback to `text` type (as the HTML spec does).

### DIFF
--- a/thirdparty/simpletest/page.php
+++ b/thirdparty/simpletest/page.php
@@ -94,7 +94,7 @@ class SimpleTagBuilder {
             $tag_class = $map[$type];
             return new $tag_class($attributes);
         }
-        return false;
+        return new SimpleTextTag($attributes);
     }
 
     /**


thirdparty/simpletest/page.php
```diff
@@ -90,7 +90,7 @@ function _createInputTag($attributes) {
                 'hidden' => 'SimpleTextTag',
                 'password' => 'SimpleTextTag',
                 'file' => 'SimpleUploadTag');
-        if (array_key_exists($type, $map)) {
+        if (array_key_exists($type, $map) || $type = 'text') {
```
tractorcow: Um, `text` is IN the $map. The second condition will never be true when the first is false.

tractorcow: Oh right, assignment via condition.... this is why it's a bad practice. Can you rewrite as a pure assignment to $tag_class instead of a condition?

dhensby:
@tractorcow ok, changed

Issue #7729, opened 2017-12-22
Themed asset inclusion using ResourceLoader breaks manual template rendering


dhensby:
## Affected Version

4.x

## Description

With the new themed resources one now includes themed CSS, images and javascript using template helpers such as `<% themedCSS("css/main.css") %>`; this would have been `<link rel="stylesheet" href="$ThemeDir/css/main.css">` in SS 3.

When executing a template like so: `SSViewer::execute_template('Template');` the themed resources are not included. This is because `execute_template` disables the inclusion of requirements (which is helpful as we don't want requirements from the main request thread included in our one-off template rendering) but then means the `themedCSS` (or other themed resources that are brought in) are ignored.

I'd expect that the themed resource to be included in the rendered template.

## Steps to Reproduce

- Create a template with the themed resource included (eg: `<% themedCSS('cssfile.css') %>`)
- Render template using `SSViewer::execute_template()`
- Notice themed resource is not included

tractorcow:
I'm not 100% sure I agree with the conclusion, although I'll admit this is a problem. `<% require %>` calls in templates aren't instructions to include that specific requirement in that actual template, just in the final page call. That is why you can have a `Widget.ss` that adds a `<% require %>` which actually sends that requirement up into the page header, potentially several templates higher up the chain.

`<% themedCSS("css/main.css") %>` works the same way in 3.x, this is not a new change.

The problem I agree with is the fact that these resources aren't being emitted anywhere. This is due to the nature of Requirements_Backend being bound to a static member on Requirements object. That's a lot harder to solve.

Wouldn't a simpler fix for the time being be to allow `execute_template` to conditionally include requirements for the time being? A larger refactor of requirements persistence could be done later. This would probably be something similar in scope to the de-staticifying to Session object.

tractorcow:
Ok, I've slightly modified my feeling on this. I've documented that over at https://github.com/silverstripe/silverstripe-framework/pull/7730#issuecomment-355871510 with some recommendations. :)

tractorcow:
Fixed with #7730

PR #7730, opened 2017-12-22
FIX `SSViwer::execute_template` should allow requirements to be included

dhensby:
Potential fix to #7729 

This creates a `Requirements_Backend` instance for the rendering of template when using `SSViewer::execute_template` so that requirements can be safely included in templates that are rendered standalone.

### DIFF
--- a/src/View/SSTemplateParser.peg
+++ b/src/View/SSTemplateParser.peg
@@ -893,7 +893,7 @@ class SSTemplateParser extends Parser implements TemplateParser
 
         // Note: 'type' here is important to disable subTemplates in SSViewer::getSubtemplateFor()
         $res['php'] = '$val .= \\SilverStripe\\View\\SSViewer::execute_template([["type" => "Includes", '.$template.'], '.$template.'], $scope->getItem(), array(' .
-            implode(',', $arguments)."), \$scope);\n";
+            implode(',', $arguments)."), \$scope, true);\n";
 
         if ($this->includeDebuggingComments) { // Add include filename comments on dev sites
             $res['php'] =
--- a/src/View/SSTemplateParser.php
+++ b/src/View/SSTemplateParser.php
@@ -3501,7 +3501,7 @@ class SSTemplateParser extends Parser implements TemplateParser
 
         // Note: 'type' here is important to disable subTemplates in SSViewer::getSubtemplateFor()
         $res['php'] = '$val .= \\SilverStripe\\View\\SSViewer::execute_template([["type" => "Includes", '.$template.'], '.$template.'], $scope->getItem(), array(' .
-            implode(',', $arguments)."), \$scope);\n";
+            implode(',', $arguments)."), \$scope, true);\n";
 
         if ($this->includeDebuggingComments) { // Add include filename comments on dev sites
             $res['php'] =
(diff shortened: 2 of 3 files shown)


tractorcow:
If you are using `<% require SomeTemplatesWithIncludes %>` I would (as a developer) prefer that these requirements were added to the top level template, which may actually contain my `<head>` tag (rather than forcing these inline).

E.g.

```
<html>
<head>
</head>
<body>
<% Include SomeTemplateWithIncludes %>
</body>
</html>
```

Should probably allow those css / js required files to go into the `<head>`. See my comments on https://github.com/silverstripe/silverstripe-framework/issues/7729 about how I see those `<% require %>` calls best behaving.

At the moment, those files aren't added anywhere, so it's still broken as it is. :)

Might I suggest instead another flag added to execute_template() to toggle the behaviour of require.

```
/**
     * Execute the given template, passing it the given data.
     * Used by the <% include %> template tag to process templates.
     *
     * @param string $template Template name
     * @param mixed $data Data context
     * @param array $arguments Additional arguments
     * @param Object $scope
     * @param bool $inlineRequirements Include required resources in the result. If false, these will be added to the global Requirements_Backend object instead
     * @return string Evaluated result
     */
    public static function execute_template($template, $data, $arguments = null, $scope = null, $inlineRequirements = true)
```

Note that the default matches your suggested behaviour; Most of the times that `execute_template` is called by users, those users would expect those files being included.

however, SSTemplateParser would pass in `false` to this argument, meaning that all requirements for a single template call (containing many `<% includes %>` potentially) are collected into a single top level requirements.


```php
// Note: 'type' here is important to disable subTemplates in SSViewer::getSubtemplateFor()
// Pass in false to $inlineRequirements to ensure requirements are collected at a single top level
        $res['php'] = '$val .= \\SilverStripe\\View\\SSViewer::execute_template([["type" => "Includes", '.$template.'], '.$template.'], $scope->getItem(), array(' .
            implode(',', $arguments)."), \$scope, false);\n";
```

And thirdly, get rid of `$v->includeRequirements(false);` call in execute_template() completely (as you have done), since a `<% require %>` should never be ignored.

tractorcow:
Note that the other risk is that multiple templates may `<% require %>` the same file (e.g. jquery) so consolidating this into a single requirements captures that use case.

dhensby:
ah, I didn't realise that `execute_template` was used so extensively in the compiled templates!

The issue I'm wanting to resolve is the fact that when used as a standalone template executor, the requirements are not included and this is a larger problem now that module resource location is managed by using requirements instead of being hard coded into templates...

I'll look into your suggested solution @tractorcow, thanks :)

tractorcow:
I wonder if $scope should actually have a requirements object attached to it, and if not passed it it adds it to global requirements.

dhensby:
OK, I've added an extra parameter to the `execute_template` and `execute_string` methods which toggles this behaviour and I've updated the template parser to pass through the right argument.

dhensby:
Sure

dhensby:
rebased

tractorcow:
Um, you rebased and deleted all the new code @dhensby .

> OK, I've added an extra parameter to the execute_template and execute_string methods which toggles this behaviour and I've updated the template parser to pass through the right argument.

That code seems to have disappeared. =(

dhensby:
Well that wasn't clever... I'll sort it out

dhensby:
OK - back to how it should be :)

tractorcow:
Test failures unrelated, fixed with https://github.com/silverstripe/silverstripe-framework/pull/7824

PR #7734, opened 2018-01-02
WIP [CMS-2064] BUGFIX: getCMSActions should add actions when defined on the DataObject

elliot-sawyer:
I raised this issue on https://github.com/silverstripe/silverstripe-cms/issues/2064. This is not a complete fix - the new FormAction buttons appear, but the actions are not invoked if they are added via a DataExtension.

I found that when you create a method "getCMSActions" and attempt to add some new formactions to the record, the method is never invoked. @flamerohr discovered that getCMSActions is never called from GridfieldDetailForm_ItemRequest.

Again, this is a partial fix and it still needs some work.


elliot-sawyer:
closing due to https://github.com/silverstripe/silverstripe-cms/issues/2064#issuecomment-355869647

PR #7738, opened 2018-01-08
DOCS Correct docs on polymorphic dataobject usage

### LINKED ISSUE #7732: PolymorphicForeignKey only works with `DataObject::class` when declared in has_one
## Affected Version

4

## Description

- It appears that using "DataObject" (as per docs) won't create a polymorphic Foreign key
- If you use "PolymorphicForeignKey" as your relationship type it'll be overridden with "ForeignKey"

https://github.com/silverstripe/silverstripe-framework/blob/4.0/src/ORM/DataObjectSchema.php#L488-L496

tractorcow:
Fixes #7732


dhensby:
Are any changes to the upgrader required?

tractorcow:
No, we still use `"DataObject"` to `DataObject::class` as a migration so it should just work.

PR #7741, opened 2018-01-09
API Implement support for public/ webroot folder

tractorcow:
Implements https://github.com/silverstripe/silverstripe-framework/issues/7419


src/Control/Director.php
```diff
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\Control;
 
+use SilverStripe\Assets\Filesystem;
```
kinglozzer: Minor concern: increased coupling to `silverstripe/assets`. Looking at `SilverStripe\Assets\Filesystem`, a lot of the contents of that class aren’t really specific to our asset model, so it might be more appropriate to have that “helper” class in framework given how regularly we are/will be using it.

Can be separate to this PR and target 5.x of course :)

src/Control/SimpleResourceURLGenerator.php
```diff
@@ -65,40 +67,103 @@ public function setNonceStyle($nonceStyle)
...
+                } elseif (!$withPublic && file_exists($privatePath)) {
+                    // String is private but exposed to resources/
+                    $absolutePath = $privatePath;
+                    $exists = true;
+                    $relativePath = Filesystem::joinPaths(ManifestFileFinder::RESOURCES_DIR, $relativePath);
+                }
+            }
         }
         if (!$exists) {
             trigger_error("File {$relativePath} does not exist", E_USER_NOTICE);
```
kinglozzer: If I request an “exposed” file, but I’ve forgotten to run `composer vendor-expose`, is this the error I’d see? Do you think there’d be any value in adding a hint about running that command here?

src/View/PublicThemes.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace SilverStripe\View;
+
+class PublicThemes implements ThemeList
+{
+    public function getThemes()
+    {
+        return PUBLIC_DIR ? ['/' . PUBLIC_DIR] : [];
+    }
```
kinglozzer: So simple, that’s awesome

src/Control/SimpleResourceURLGenerator.php
```diff
@@ -3,7 +3,11 @@
 namespace SilverStripe\Control;
 
 use InvalidArgumentException;
+use PhpParser\Node\Scalar\MagicConst\Dir;
```
kinglozzer: Oops?

docs/en/04_Changelogs/4.1.0.md
```diff
@@ -0,0 +1,87 @@
...
+
+```json
+{
+  "extra": {
+    "expose": [
+      "mysite/client"
+    ]
+  }
+}
+```
```
kinglozzer: It should be pretty obvious for most, but can you explicitly mention that this goes in `composer.json`?

docs/en/04_Changelogs/4.1.0.md
```diff
@@ -0,0 +1,87 @@
+# 4.1.0
+
+## Introduction
+
+This version introduces many breaking changes, which in most projects can be managed through a combination
```
kinglozzer: > This version introduces many breaking changes...

... is a worrying way to start a minor release changelog 😛. Does this actually break anything if you `composer update` 4.0.x to 4.1.x _without_ opting into the `public/` directory?

tractorcow: Oh dear I've copied and pasted the 4.0.0.md changelog intro and didn't modify the intro again.

docs/en/04_Changelogs/4.1.0.md
```diff
@@ -0,0 +1,87 @@
...
+
+ * Support for public webroot folder `public/`
+ * Better support for cross-platform filesystem path manipulation
+
+## Upgrading
+
+### Upgrade `public/` folder
+
+By default all new projects installed from 4.1.0 will have a `public/` webroot folder. This folder name is
+not configurable, but is turned on by creating this file, and off by ensuring this folder doesn't exist.
```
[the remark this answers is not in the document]
tractorcow: folder.

docs/en/04_Changelogs/4.1.0.md
```diff
@@ -0,0 +1,87 @@
...
+  themes:
+    - '$public'
+    - 'simple'
+    - '$default'
+```
+
+This would allow `<% require themedCSS('style.css') %>` to find a file comitted to `public/css/style.css`.
+
+Note that `Requirements` calls will look in both the `public` folder (first) and then the base path when
+resolving css or javascript files.
```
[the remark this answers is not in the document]
tractorcow: It will look in `public/css/style.css` not in mysite.

Expose is never automatic; You must declare it in composer.json.

src/Control/Director.php
tractorcow: I could move it to a `SilverStripe\Core` class maybe.

src/Control/Director.php
```diff
@@ -589,7 +588,38 @@ public static function baseURL()
...
         $alternate = Director::config()->uninherited('alternate_base_folder');
-        return ($alternate) ? $alternate : BASE_PATH;
+        return $alternate ?: BASE_PATH;
+    }
+
+    /**
+     * Check if using a seperate public dir.
+     * This will be removed in 5.0 and forced to true
+     *
+     * @return bool
```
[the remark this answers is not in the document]
tractorcow: Oh, another case of copy-paste fail. That phpdoc isn't even for that method!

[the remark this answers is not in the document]
tractorcow: Oh right, we re-purposed this method but didn't update phpdoc.

src/Control/SimpleResourceURLGenerator.php
tractorcow: Haha , oh dear.

src/Control/SimpleResourceURLGenerator.php
```diff
@@ -65,40 +67,103 @@ public function setNonceStyle($nonceStyle)
...
+                // All resources mapped directly to resources/
+                $relativePath = Filesystem::joinPaths(ManifestFileFinder::RESOURCES_DIR, $relativePath);
+            } elseif (stripos($relativePath, ManifestFileFinder::VENDOR_DIR . DIRECTORY_SEPARATOR) === 0) {
+                // @todo Non-public dir support will be removed in 5.0, so remove this block there
+                // If there is no public folder, map to resources/ but trim leading vendor/ too (4.0 compat)
+                $relativePath = Filesystem::joinPaths(
+                    ManifestFileFinder::RESOURCES_DIR,
+                    substr($relativePath, strlen(ManifestFileFinder::VENDOR_DIR))
+                );
+            }
```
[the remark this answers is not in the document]
tractorcow: Good idea.

src/Control/SimpleResourceURLGenerator.php
tractorcow: No, that would occur if you generated a url for a file that didn't exist at all.

docs/en/04_Changelogs/4.1.0.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Expose root project files
```
chillu: * Describe how to detect modules which might state compatibility with 4.x, but are using public files without using `expose` yet. Also make it really clear that module authors need to switch to `silverstripe-vendormodule` now, particularly if they're exposing public files.

docs/en/04_Changelogs/4.1.0.md
```diff
@@ -0,0 +1,87 @@
...
+ webserver. E.g. move `mysite/javascript/script.js` to `public/javascript/script.js`.
+ - Ensure that the web-root configured for your server of choice points to the public/ folder instead of the base path.
+ 
+You may also need to add various changes to your code if you reference the BASE_PATH directly:
+
+ - You should use `Director::publicFolder()` instead of `Director::baseFolder()` if referring to the
+ public folder.
+ - You can check if a public folder exists with `Director::publicDir()`
+ 
+### Use new `$public` theme set
```
chillu: * Should contain a section on how to run your project without using themes, as an alternative or addition to "use new $public theme set"

docs/en/04_Changelogs/4.1.0.md
```diff
@@ -0,0 +1,87 @@
...
+not configurable, but is turned on by creating this file, and off by ensuring this folder doesn't exist.
+
+When separating the public webroot from the BASE_PATH it is necessary to move a few files during migration:
+
+ - Move `.htaccess` from base to `public/`
+ - Move `index.php` from base to `public/`
+ - Move `assets` folder (including the nested `assets/.protected` folder) into `public/`
+ - Ensure that the `public/resources` folder exists; If this folder already exists in root, you should
+   delete this, and re-generate it by running `composer vendor-expose` in your root path.
+ - Any public assets committed directly to your project intended to be served directly to the
```
chillu: * Should point to a clear description how to reference files in the public webroot programmatically (in PHP and SS files) in the permanent docs. Should also cross-link to the section in the 4.0 upgrading docs describing this change

docs/en/04_Changelogs/4.1.0.md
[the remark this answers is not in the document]
chillu: * I think we need to sell this upgrade effort more actively - highlight that it's a security improvement and simplifies webserver configuration.
 * This should have a before/after sample project structure, much easier to grasp on quick review.

docs/en/04_Changelogs/4.1.0.md
```diff
@@ -0,0 +1,87 @@
...
+ webserver. E.g. move `mysite/javascript/script.js` to `public/javascript/script.js`.
+ - Ensure that the web-root configured for your server of choice points to the public/ folder instead of the base path.
```
chillu: I think this should be split out to a section that can be reviewed by infrastructure folks. Point out which folders need write access now, and which should be marked readonly.

Also add a point about reviewing your webserver configuration, and removing unnecessary rules (point to the new installer `.htaccess`?)

docs/en/04_Changelogs/4.1.0.md
```diff
@@ -0,0 +1,87 @@
...
+ public folder.
+ - You can check if a public folder exists with `Director::publicDir()`
+ 
+### Use new `$public` theme set
+ 
+In addition there is a new helper pseudo-theme that you can configure to expose files in the `public/`
+folder to the themed css / javascript file lookup. For instance, this is how you can prioritise those
+files:
+
+```yaml
```
chillu: Is the `$public` thing described in the permanent docs as well?

docs/en/04_Changelogs/4.1.0.md
```diff
@@ -0,0 +1,87 @@
...
+ - Move `assets` folder (including the nested `assets/.protected` folder) into `public/`
+ - Ensure that the `public/resources` folder exists; If this folder already exists in root, you should
+   delete this, and re-generate it by running `composer vendor-expose` in your root path.
+ - Any public assets committed directly to your project intended to be served directly to the
+ webserver. E.g. move `mysite/javascript/script.js` to `public/javascript/script.js`.
+ - Ensure that the web-root configured for your server of choice points to the public/ folder instead of the base path.
+ 
+You may also need to add various changes to your code if you reference the BASE_PATH directly:
+
+ - You should use `Director::publicFolder()` instead of `Director::baseFolder()` if referring to the
```
chillu: Do we need to document `PUBLIC_PATH` and `PUBLIC_DIR` here, and where we list all available constants? I guess they should be treated on the same level as `BASE_PATH` and `BASE_DIR` - either all four can be referenced directly in code, or they should all be deprecated in favour of their `Director` API counterparts.

docs/en/04_Changelogs/4.1.0.md
tractorcow: We probably need to discuss this; Most sites will have at least one theme.

docs/en/04_Changelogs/4.1.0.md
tractorcow: Yes.

docs/en/04_Changelogs/4.1.0.md
tractorcow: I think we prefer users use the methods, rather than the constants, as it provides for better test mocking.

docs/en/04_Changelogs/4.1.0.md
```diff
@@ -75,13 +161,28 @@ to the web server, you can also use the composer `expose` directive to symlink /
 Then run the composer helper `composer vendor-expose` in your project root. This will symlink (or copy)
 the `mysite/client` directory to `public/resources/mysite/client`.
 
-### Filesystem manipulation helpers
+If you are using third party modules which may not have explic `expose` directives,
```
chillu: Typo ("explic")

docs/en/04_Changelogs/4.1.0.md
```diff
@@ -15,25 +9,114 @@ guide developers in preparing existing 3.x code for compatibility with 4.0
...
  - You should use `Director::publicFolder()` instead of `Director::baseFolder()` if referring to the
  public folder.
  - You can check if a public folder exists with `Director::publicDir()`
+
+### Example `public/` folder structure
+ 
+For example, this is an existing folder structure:
+
+```
+- assets/
```
chillu: Marking this up as a nested list in pseudo-Markdown seems a bit weird - I don't think we do that in other places around the docs. I find it harder to visually parse this way.

More importantly: Naming multiple files the same (`style.css`), where only *some* are symlinks of others is very confusing.

docs/en/04_Changelogs/4.1.0.md
```diff
@@ -75,13 +161,28 @@ to the web server, you can also use the composer `expose` directive to symlink /
 Then run the composer helper `composer vendor-expose` in your project root. This will symlink (or copy)
 the `mysite/client` directory to `public/resources/mysite/client`.
 
-### Filesystem manipulation helpers
+If you are using third party modules which may not have explic `expose` directives,
+you can also expose those assets manually by declaring the full path to the directory to expose.
+This works the same for `silverstripe-module` and `silverstripe-vendormodule` types.
```
chillu: Link to more detail about what a vendormodule is here.

docs/en/04_Changelogs/4.1.0.md
chillu: You've inlined some ways of doing this, which is a good start - but I'd still like to see a deep link into the docs.

chillu:
I think the docs need to be more active about encouraging this switch. Current version:

> By default all new projects installed from 4.1.0 will have a public/ webroot folder which separates all web-accessible files from protected project assets, providing an additional level of security.

Recommended new version:

> This release allows to maintain a public webroot folder which separates all web-accessible files from protected project files (like the vendor folder and all of your PHP files). This makes your web hosting more secure, and less likely to leak information through accidentally deployed files like a project README. New projects will default to this behaviour. Existing projects (updating from 3.x or 4.0) will continue working as-is, but we strongly recommend switching to the public webroot structure in order to get the security benefits.

docs/en/04_Changelogs/4.1.0.md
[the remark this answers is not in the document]
kinglozzer: “and then the base path” is what’s confusing me here, does that mean it’ll look one level above `public/`?

docs/en/04_Changelogs/4.1.0.md
```diff
@@ -0,0 +1,188 @@
...
+      "mysite/client"
+    ]
+  }
+}
+```
+
+Then run the composer helper `composer vendor-expose` in your project root. This will symlink (or copy)
+the `mysite/client` directory to `public/resources/mysite/client`.
+
+If you are using third party modules which may not have explic `expose` directives,
```
unclecheese: `explicit`

docs/en/04_Changelogs/4.1.0.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+### Path manipulation helpers
+
+The following filesystem helpers have been added in order to better support working with cross-platform
+path manipulation:
+
+ * `SilverStripe\Core\Path::join()` which will join one or more relative or absolute paths
+ * `SilverStripe\Core\Path::normalise()` which will normalise and trim directory separators in a relative or absolute path
+ * `SilverStripe\Core\Convert::slashes()` to convert directory separators to either `/` or `\`
```
unclecheese: These descriptions are a bit obtuse. Example inputs and outputs for each would really help.

src/View/PublicThemes.php
unclecheese: Could use `Path` here to be consistent. :)

tractorcow: @unclecheese how do you mean? Absolute path themes in config always use `/` prefix.

tractorcow:
Ok, I've checked the tests... you will need to merge assets first, and then restart framework tests to get a green.

PR #7744, opened 2018-01-10
CS Force spaces around string concat operators

dhensby:
String concatenation operators should have spaces around them; this change enforces them.


tractorcow:
That's not psr1/2 is it? I'm hesitant to introduce new requirements over psr-2 (which we even now aren't following properly).

robbieaverill:
No, it's not PSR-2, but I like the change anyway. We potentially have a little more freedom with framework over putting this into installer for example. I do think that sticking to PSR-2 is good enough though.

dhensby:
PSR2 says spaces around operators, this is an operator.

At the moment I don't think there's any enforcement of this either way which means we are inconsistent and I'd like it to be...

tractorcow:
I don't think that's the case; PSR-2 intentionally omits declaring operator coding styles.

https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#7-conclusion

I like the coding style but I think we should be true to PSR-2 and not try to extend it too much, lest we become too strict.

I would like a "cleanup only" rule that might normalise to include the space as a preference, but not a rule that fails linting.

Maybe we use `<severity>0</severity>` ? Would that work? (cleanup code with this rule, but not fail CI if it doesn't match).

dhensby:
Sure, I'm happy with a loose constraint

tractorcow:
This
```diff
-<rule ref="Squiz.Strings.ConcatenationSpacing">
+<rule phpcbf-only="true" ref="Squiz.Strings.ConcatenationSpacing">
    <properties>
        <property name="spacing" value="1"/>
    </properties>
</rule>
```

tractorcow:
Can be merged on green. Looks like travis is backed up though.

PR #7746, opened 2018-01-11
FIX Allow extension instances to be overridden by injector

dhensby:
If you override an Extension instance in injector with Yaml like so:

```yaml
  WidgetExtension:
    class: MyWidgetExtension
```

The checks on `$object->hasExtension('WidgetExtension')` will fail because `Extensible` adds the class of the actual extension instance (`MyWidgetExtension`) as the name of the extension. This can result in logic that uses `hasExtension` in modules to fail when `Injector` is used to override the `Extension` instance.

This change would allow the `$object->hasExtension` (and related methods) to return correctly when `Injector` has been used to redefine the class of an `Extension`.

### DIFF
--- a/src/Core/Extensible.php
+++ b/src/Core/Extensible.php
@@ -495,17 +495,22 @@ trait Extensible
     /**
      * Get an extension instance attached to this object by name.
      *
-     * @uses hasExtension()
-     *
      * @param string $extension
-     * @return Extension
+     * @return Extension|null
      */
     public function getExtensionInstance($extension)
     {
         $instances = $this->getExtensionInstances();
-        return isset($instances[$extension])
-            ? $instances[$extension]
-            : null;
+        if (array_key_exists($extension, $instances)) {
+            return $instances[$extension];
+        }
+        // in case Injector has been used to replace an extension
+        foreach ($instances as $instance) {
+            if (is_a($instance, $extension)) {
+                return $instance;
+            }
+        }
+        return null;
     }
 
     /**
@@ -524,8 +529,7 @@ trait Extensible
      */
     public function hasExtension($extension)
     {
-        $instances = $this->getExtensionInstances();
-        return isset($instances[$extension]);
+        return (bool) $this->getExtensionInstance($extension);
     }
 
     /**
@@ -553,8 +557,16 @@ trait Extensible
 
             if ($extensions) {
                 foreach ($extensions as $extension) {
-                    $instance = Injector::inst()->get($extension);
-                    $this->extension_instances[get_class($instance)] = $instance;
+                    $name = $extension;
+                    // Allow service names of the form "%$ServiceName"
+                    if (substr($name, 0, 2) == '%$') {
+                        $name = substr($name, 2);
+                    }
+                    $name = trim(strtok($name, '('));
+                    if (class_exists($name)) {
+                        $name = ClassInfo::class_name($name);
+                    }
+                    $this->extension_instances[$name] = Injector::inst()->get($extension);
                 }
             }
         }
--- a/src/Core/ClassInfo.php
+++ b/src/Core/ClassInfo.php
@@ -208,6 +208,7 @@ class ClassInfo
      * eg: self::class_name('dataobJEct'); //returns 'DataObject'
      *
      * @param string|object $nameOrObject The classname or object you want to normalise
+     * @throws \ReflectionException
      * @return string The normalised class name
      */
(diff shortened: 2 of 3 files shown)


dhensby:
The key reason I feel this is a bug is that it breaks the idea that `Injector` provides an abstraction layer between the real classes of objects and the interfaces/names of classes that are defined.

In my example above we should be able to continue to use `WidgetExtension` as our extension name in code and be ignorant to the real class name of the extension.

src/Core/Extensible.php
```diff
@@ -553,8 +567,12 @@ public function getExtensionInstances()
 
             if ($extensions) {
                 foreach ($extensions as $extension) {
-                    $instance = Injector::inst()->get($extension);
-                    $this->extension_instances[get_class($instance)] = $instance;
+                    list($extensionName) = ClassInfo::parse_class_spec($extension);
```
kinglozzer: Can you copy the approach used in `Injector::normaliseArguments()` to skip calling `ClassInfo::parse_class_spec()` if the definition doesn’t contain a `(`?

Might be safest to also include the `$name = trim($name);` from that method too, just in case...

dhensby: done

src/Core/Extensible.php
```diff
@@ -525,7 +530,16 @@ public function getExtensionInstance($extension)
...
+        if (array_key_exists($extension, $instances)) {
+            return true;
+        }
+        // Injector can be used to override the class used
+        foreach ($instances as $instance) {
+            if (is_a($instance, $extension)) {
+                return true;
+            }
+        }
+        return false;
```
robbieaverill: Can we check by reusing `getExtensionInstances()` here instead of duplicating its logic? It doesn't look like it would be any extra overhead

dhensby: Good call - done

src/Core/Extensible.php
```diff
@@ -553,8 +557,21 @@ public function getExtensionInstances()
...
-                    $instance = Injector::inst()->get($extension);
-                    $this->extension_instances[get_class($instance)] = $instance;
+                    $name = $extension;
+                    // Allow service names of the form "%$ServiceName"
+                    if (substr($name, 0, 2) == '%$') {
+                        $name = substr($name, 2);
+                    }
+
+                    if (strstr($name, '(')) {
+                        list($name) = ClassInfo::parse_class_spec($name);
```
tractorcow:
```php
$name = strtok($name, '(');
```

would be a lot faster; You can even skip the `strstr` as it's safe.

src/Core/Extensible.php
```diff
@@ -553,8 +557,21 @@ public function getExtensionInstances()
 
             if ($extensions) {
                 foreach ($extensions as $extension) {
-                    $instance = Injector::inst()->get($extension);
-                    $this->extension_instances[get_class($instance)] = $instance;
+                    $name = $extension;
+                    // Allow service names of the form "%$ServiceName"
+                    if (substr($name, 0, 2) == '%$') {
+                        $name = substr($name, 2);
```
tractorcow: `$name = ltrim($name, '%$');` faster :)

src/Core/Extensible.php
```diff
@@ -553,8 +557,21 @@ public function getExtensionInstances()
...
+                    }
+
+                    if (strstr($name, '(')) {
+                        list($name) = ClassInfo::parse_class_spec($name);
+                    }
+                    $name = trim($name);
+                    try {
```
tractorcow: An exception thrown would have a larger performance cost than a `class_exists` check. We should avoid using exceptions in non-error conditions.

tractorcow:
Overall approach is good; Since this is a performance-critical piece of code I've left some suggestions on how to optimise it.

src/Core/Extensible.php
dhensby: `ltrim($name, '%$')` would also turn strings starting with `$%` or `$` or `%` or multiple combinations of those as the second argument to `trim` functions is a list of chars to trim.

src/Core/Extensible.php
dhensby: done

src/Core/Extensible.php
dhensby: done

src/Core/Extensible.php
tractorcow: Yeah and none of those characters are in a class name :)

PR #7754, opened 2018-01-12
JSON.parse does not like empty strings

oetiker:
Add some protection for the case that `event.data` is empty or null. Why the backend chose not to send data, I do not know, but I guess having the code not to crash is good anyway.

### DIFF
--- a/admin/javascript/LeftAndMain.js
+++ b/admin/javascript/LeftAndMain.js
@@ -73,7 +73,7 @@ jQuery.noConflict();
 		$(window).on("message", function(e) {
 			var target,
 				event = e.originalEvent,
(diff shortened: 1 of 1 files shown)


dhensby:
I was thinking of porting this to silverstripe/admin but have decided not to because we have this wrapped in a try/catch - @tractorcow happy not to port it across?

oetiker:
my patch simply allows for there to be no data ... I do not know how this fits into the overall SS picture ... can it be in a correct case that there is no data? if so you should not rely on try/catch I think

dhensby:
it seems to be handled fine:

https://github.com/silverstripe/silverstripe-admin/blob/1.0/client/src/legacy/LeftAndMain.js#L69-L79

PR #7755, opened 2018-01-15
BUG Don't re-declare manually-declared variables for public folder

tractorcow:
This prevents issues where utilites such as silverstripe/serve declare these in advance. E.g. https://github.com/silverstripe/silverstripe-serve/pull/22

### DIFF
--- a/src/includes/constants.php
+++ b/src/includes/constants.php
@@ -59,8 +59,12 @@ if (!defined('BASE_PATH')) {
 }
 
 // Set public webroot dir / path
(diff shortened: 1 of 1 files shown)


flamerohr:
behat is failing

tractorcow:
Yes because it needs to be restarted with https://github.com/silverstripe/silverstripe-serve/releases/tag/2.1.1

tractorcow:
Ok restarted, let's see if that works.

tractorcow:
Seems to fix it.  ;)

PR #7760, opened 2018-01-16
BUG Fix themed editor.css url not being resolved properly

tractorcow:
Switch out `Director::absoluteURL` for correct resource-to-url generator.


robbieaverill:
If this is a patch, do we want it in 4.0.x?

dhensby:
Agreed, looks like a patch

tractorcow:
It's only a bug in 4.1 because of public webroot.

PR #7766, opened 2018-01-16
FIX Misnamed test namespaces

dhensby:
The test namespaces are `Tests` not `Test`


tractorcow:
Can be merged on green.

PR #7770, opened 2018-01-17
Add `PUBLIC_*` constants to 4.0.x for easier compatibility

dhensby:
4.x now has support for a separate public dir to be used as the webroot.

This fix back-ports the definition of the `PUBLIC_*` constants so that modules that want to support 4.0 and 4.1 don't have to either litter their code with `if (defined('PUBLIC_PATH'))` or mandate 4.1 as a dependency.

At the moment supporting the public webroot requires extra logic that would best be avoided if we truly want it to be opt-in.


kinglozzer:
Makes sense to me, there’s no bigger BC risk from this change than there was introducing the `public/` dir behaviour for 4.1 and it’s one less thing for module devs to worry about.

PR #7773, opened 2018-01-18
BUG Make GridFieldConfig less susceptible to error when versioned isn't installed

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/pull/7422

I've updated this code to fail more gracefully in case versioned isn't installed.


flamerohr:
could this be shifted to the `versioned` module and use the `updateConfig` extension point?

tractorcow:
Yes but I'll make that a 4.1 change.

tractorcow:
4.1 change is at https://github.com/silverstripe/silverstripe-versioned/issues/87, but that relies on another story before it can be merged.

PR #7774, opened 2018-01-18
API Deprecate GridFieldVersionedState

tractorcow:
Main story: https://github.com/silverstripe/silverstripe-versioned/issues/87


tractorcow:
Updated based on feedback and ready to peer review.
 - Removed `published` state
 - Remove border style from inline badge

PR #7776, opened 2018-01-19
Feature/gridfield filter header customisation

mfendeksilverstripe:
Add following customization options to the `GridFieldFilterHeader` component:

* field aliases (useful when custom methods are used to populate column content)
* omitted fields (sometimes we don't want a default filter to appear in the column header)
* custom fields (for example dropdowns and such, custom placeholder attributes...)
* custom filters (to cover edge cases when filtering)

### DIFF
--- a/src/Forms/GridField/GridFieldFilterHeader.php
+++ b/src/Forms/GridField/GridFieldFilterHeader.php
@@ -3,10 +3,17 @@
 namespace SilverStripe\Forms\GridField;
 
 use SilverStripe\Forms\FieldGroup;
+use SilverStripe\Forms\FormField;
+use SilverStripe\Forms\GridField\GridField;
+use SilverStripe\Forms\GridField\GridField_FormAction;
+use SilverStripe\Forms\GridField\GridState_Data;
 use SilverStripe\Forms\TextField;
+use SilverStripe\ORM\DataList;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\DataObjectSchema;
 use SilverStripe\ORM\Filterable;
-use SilverStripe\ORM\SS_List;
 use SilverStripe\ORM\ArrayList;
+use SilverStripe\ORM\SS_List;
 use SilverStripe\View\ArrayData;
 use SilverStripe\View\SSViewer;
 use LogicException;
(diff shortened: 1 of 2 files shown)


dhensby:
Please can you add some docs for these features explaining how/when to use them?

tractorcow:
This is going to take quite a bt of review.

src/Forms/GridField/GridFieldFilterHeader.php
```diff
@@ -27,6 +63,131 @@ class GridFieldFilterHeader implements GridField_HTMLProvider, GridField_DataMan
...
+    /**
+     * @param string $field
+     * @return bool
+     */
+    protected function isAliasField($field)
+    {
+        return array_key_exists($field, $this->alias_fields);
+    }
+
+    /**
```
tractorcow: PHPDoc descriptions on these methods would assist with peer review. :D

mfendeksilverstripe:
@dhensby @tractorcow I added a detailed documentation on the new features. Please have a look.

src/Forms/GridField/GridFieldFilterHeader.php
```diff
@@ -15,10 +22,150 @@
...
+ *
+ * private static $summary_fields = [
+ *   'getTitleSummary' => 'Title',
+ *   'City.Name' => 'City',
+ *   'Expires.Nice' => 'Expires',
+ * ];
+ *
+ * alias fields need to be specified to match the summary fields mapping
+ *
+ * ->setAliasFields([
```
tractorcow: If this matches the summary_fields exactly, why is it mandatory for the user to set the field aliases explicitly? Can't the component just infer from summary_fields directly?

src/Forms/GridField/GridFieldFilterHeader.php
```diff
@@ -15,10 +22,150 @@
...
+ *
+ * REQUIREMENTS:
+ *
+ * table header component needs to be present in the GridField (for example GridFieldSortableHeader)
+ * the last column of the table needs to have a vacant header so the filter widget could be displayed there
+ * for example you can't have the last column with sorting header and filter widget at the same time
+ *
+ * ALIAS FIELDS
+ *
+ * in the case that the column name of the table is different from matching field name via the summary_fields
```
tractorcow: In the SilverStripe ORM Column === Field name. I think you may need to clarify this.

A better example would be a summary_field with humanised label.

E.g.

```
private static $summary_fields = [
  'getTitleSummary' => 'Summary Title',
  'City.Name' => 'City Name',
  'Expires.Nice' => 'Expires on',
];
```

src/Forms/GridField/GridFieldFilterHeader.php
```diff
@@ -15,10 +22,150 @@
...
+ * ->setCustomFilters([
+ *   'MediaType' => GridFieldFilterHeader::FILTER_EXACT,
+ *   'City' => function (Filterable $list, $columnName, $value) {
+ *     return $list->filter('CityID:ExactMatch', $value);
+ *   },
+ * ])
+ *
+ * custom filter can be either a string which identifies one of the filters that are available
+ * or a custom Closure which is expected to filter the list
+ * this is a great way to cover edge cases as the implementation of the filter is up to the developer
```
tractorcow: A lot of the functionality provided by this change seems to be covered by SearchContext. Have you considered using an injectable context instead of replicating this functionality?

This includes:

 - Scaffolding form fields
 - Querying results from a search
 - Handling http requests

mfendeksilverstripe:
New changes added:

* documentation updated
* field placeholder is now customizable

@tractorcow I think we can't rely on having `DataObjects` of the same type in the `GridField` at all times. The list can be populated by an `ArrayList` with arbitrary items. So I think that is the reason why we need `setAliasFields()`. It's the same thing as in `GridFieldSortableHeader::setFieldSorting()`.

I reviewed the `SearchContext`, but again this is `DataObject` specific stuff which wouldn't work properly with `ArrayList` with items of different types. Additionally, I don't think you can search in other fields via the search context. For example, we have a filter that searches three other fields like this:

```
->setCustomFilters([
    'Title' => function (Filterable $list, $columnName, $value) {
        return $list->filterAny([
            'Title:PartialMatch' => $value,
            'Caption:PartialMatch' => $value,
            'Credit:PartialMatch' => $value,
        ]);
    },
```

tractorcow:
> @tractorcow I think we can't rely on having DataObjects of the same type in the GridField at all times. The list can be populated by an ArrayList with arbitrary items. So I think that is the reason why we need setAliasFields(). It's the same thing as in GridFieldSortableHeader::setFieldSorting().

GridField is built on this assumption though. You should assume that `$gridfield->getModelClass()` is the correct base class for your list, and infer summary fields from that.

mfendeksilverstripe:
Some of these changes made it into a separate module. Feel free to cherrypick if you find it useful.

https://github.com/silverstripe-terraformers/gridfield-rich-filter-header

PR #7779, opened 2018-01-22
BUG Fix double casting in login authenticator name

### LINKED ISSUE #7769: Authenticator title shows escaped &amp; on login screen
Affects 4.x

When adding a second authenticator, the `MemberAuthenticator` title displays `&amp;` instead of an ampersand.

screen shot 2018-01-17 at 14 58 56

Suggested solution: add the following to `MemberLoginForm`

```php
private static $casting = [
    'AuthenticatorName' => 'HTMLFragment'
];
```

# PRs:

 - [ ] https://github.com/silverstripe/silverstripe-framework/pull/7779

tractorcow:
Fixes #7769

Given that this field is only ever used in places where the casting ends up broken, I'm going to err in favour of securing the source string instead of raw-ifying the output casting.

### DIFF
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -271,7 +271,7 @@ en:
     ERRORWRONGCRED: 'The provided details don''t seem to be correct. Please try again.'
     NoPassword: 'There is no password on this member.'
   SilverStripe\Security\MemberAuthenticator\MemberLoginForm:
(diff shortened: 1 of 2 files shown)


kinglozzer:
Thanks @tractorcow :)

PR #4598, opened 2015-09-10
fix gridfield overflow

scott1702:
Before:
image
After:
image

PR #7784, opened 2018-01-23
FIX: Gridfields with dropdowns having lots of overflow

scott1702:
Due to the issue with how css overflows work [here](https://stackoverflow.com/a/6433475), the [original fix](https://github.com/silverstripe/silverstripe-framework/pull/5080) for overflowing gridfields causes scrollable gridfields when used in conjunction with dropdowns. This change targets more specifically the areas of gridfield which caused the issue shown [here](https://github.com/silverstripe/silverstripe-framework/pull/4598)

Before:
gridfield overflow issue

After (no scrollbar):
image

Tested on Chrome


robbieaverill:
You happy with this @dhensby?

dhensby:
We should really check this doesn't cause any regressions in our supported browsers:

- [ ] IE 8
- [ ] IE 9
- [ ] IE 10
- [ ] IE 11
- [x] Chrome
- [ ] Firefox

PR #7785, opened 2018-01-23
BUG Better upload error message

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/720

The issue was the literal PHP error integer was being used as the error message.

Using the in-built upload validator will turn this into a proper localised error message.


tractorcow:
Before:

image


After:

image

Issue #7788, opened 2018-01-23
DBForeignKey has_one Dropdown fields resort to a NumericField for >100 objects


brew-jay:
## Description
When using scaffolding to generate a has_one field using a model admin. 

The has_one field generates a nice dropdownfield. However if the count of objects goes above 100 the field just resorts to numeric Field. 

This becomes extremely frustrating from a UX Point of View where the CMS works and looks nice, but if the client adds more than 100 objects then it breaks the experience.

It should at least be configurable? 
Offending code.  

```
if ($list->count() < 100) {
            $field = new DropdownField($this->name, $title, $list->map('ID', $titleField));
            $field->setEmptyString(' ');
        } else {
            $field = new NumericField($this->name, $title);
        }
```

This isn't mentioned in the documentation either, so it's caught us out where we've put a site live and a customer has raised this as a bug and affected their day-day running as they can't select a has-one object.

## Steps to Reproduce
Create a data object with a has_one to another dataobject. Allow scaffolding to create the fields using a model admin. 

You'll see a nice dropdown box is used. If you then add over 100 dataobjects. That has_one field resorts to a Numeric Field.

tractorcow:
In this case it's necessary for you to declare a `getCMSFields` for your custom model, and override this specific field. The default scaffolder isn't intended as a primary source of UX, just a default baseline you can use to get started.

```
public function getCMSFields()
{
	$fields = parent::getCMSFields();
	$fields->replace('RelationID', new DropdownField('RelationID', 'Select Relation', MyObject::get()->map('ID', 'Title'));
	return $fields;
}
```

maxime-rainville:
I think this recent changes fixes this https://github.com/silverstripe/silverstripe-framework/pull/11071

PR #7789, opened 2018-01-23
make dropdownFieldThreshold configurable on DBForeignKey

andrewandante:
Slightly fixes https://github.com/silverstripe/silverstripe-framework/issues/7788 by making it a configurable option and adding a description as to why this has changed.

### DIFF
--- a/src/ORM/FieldType/DBForeignKey.php
+++ b/src/ORM/FieldType/DBForeignKey.php
@@ -29,6 +29,16 @@ class DBForeignKey extends DBInt
      */
     protected $object;
 
+    /**
+     * This represents the number of related objects to show in a dropdown before it reverts
+     * to a NumericField. If you are tweaking this value, you should also consider constructing
+     * your form field manually rather than allowing it to be scaffolded
+     *
+     * @config
+     * @var int
+     */
+    private static $dropdown_field_threshold = 100;
+
     private static $index = true;
 
     private static $default_search_filter_class = 'ExactMatchFilter';
@@ -63,11 +73,13 @@ class DBForeignKey extends DBInt
         $list = DataList::create($hasOneClass);
         // Don't scaffold a dropdown for large tables, as making the list concrete
         // might exceed the available PHP memory in creating too many DataObject instances
-        if ($list->count() < 100) {
+        $threshold = self::config()->get('dropdown_field_threshold');
+        if ($list->count() < $threshold) {
             $field = new DropdownField($this->name, $title, $list->map('ID', $titleField));
             $field->setEmptyString(' ');
         } else {
             $field = new NumericField($this->name, $title);
+            $field->setRightTitle('Too many related objects; fallback field in use');
         }
         return $field;
     }
--- a/src/ORM/FieldType/DBForeignKey.php
+++ b/src/ORM/FieldType/DBForeignKey.php
@@ -79,7 +79,7 @@ class DBForeignKey extends DBInt
             $field->setEmptyString(' ');
         } else {
             $field = new NumericField($this->name, $title);
-            $field->setRightTitle('Too many related objects; fallback field in use');
+            $field->setRightTitle(_t(self::class . '.DROPDOWN_THRESHOLD_FALLBACK_MESSAGE', 'Too many related objects; fallback field in use'));
         }
         return $field;
     }
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -161,6 +161,8 @@ en:
       other: '{count} years'
   SilverStripe\ORM\FieldType\DBEnum:
     ANY: Any
+  SilverStripe\ORM\FieldType\DBForeignKey:
+    DROPDOWN_THRESHOLD_FALLBACK_MESSAGE: 'Too many related objects; fallback field in use'
   SilverStripe\ORM\Hierarchy:
     LIMITED_TITLE: 'Too many children ({count})'
   SilverStripe\ORM\Hierarchy\Hierarchy:


src/ORM/FieldType/DBForeignKey.php
```diff
@@ -63,11 +72,17 @@ public function scaffoldFormField($title = null, $params = null)
...
+        if ($list->count() < $threshold) {
             $field = new DropdownField($this->name, $title, $list->map('ID', $titleField));
             $field->setEmptyString(' ');
         } else {
             $field = new NumericField($this->name, $title);
+            $field->setRightTitle(
+                'This field has been converted to a NumericField as there are over ' . $threshold
+                . ' related objects. This can be set with the config variable'
+                . ' SilverStripe\ORM\FieldTypeDBForeignKey::dropdownFieldThreshold'
+            );
```
kinglozzer: I don’t think we should show this kind of message to CMS authors

src/ORM/FieldType/DBForeignKey.php
```diff
@@ -29,6 +29,15 @@ class DBForeignKey extends DBInt
...
     protected $object;
 
+    /**
+     * This represents the number of related objects to show in a dropdown before it reverts
+     * to a NumericField
+     *
+     * @config
+     * @var int
+     */
+    private static $dropdownFieldThreshold = 100;
```
kinglozzer: Config vars should always be `$lower_case_underscore` (except for when they’re not, `many_many_extraFields`)

andrewandante:
Updated rightTitle because I noticed it was `<` not `<=`

src/ORM/FieldType/DBForeignKey.php
```diff
@@ -63,11 +72,17 @@ public function scaffoldFormField($title = null, $params = null)
...
         // Don't scaffold a dropdown for large tables, as making the list concrete
         // might exceed the available PHP memory in creating too many DataObject instances
-        if ($list->count() < 100) {
+        $threshold = self::config()->get('dropdownFieldThreshold');
+        if ($list->count() < $threshold) {
             $field = new DropdownField($this->name, $title, $list->map('ID', $titleField));
             $field->setEmptyString(' ');
         } else {
             $field = new NumericField($this->name, $title);
+            $field->setRightTitle(
```
kinglozzer: I don’t think we should show a message like this to content authors

andrewandante: How about admins? Part of the complaint was visibility over the issue. Happy for it to be wrapped in `if ($isAdmin)`?

andrewandante: Or maybe just something more generic - remove the bit about the config var maybe. It'll save a fair bit of time for both parties if there is an "error message"

src/ORM/FieldType/DBForeignKey.php
andrewandante: snek_case_like_a_good_boi

src/ORM/FieldType/DBForeignKey.php
```diff
@@ -63,11 +73,18 @@ public function scaffoldFormField($title = null, $params = null)
...
+        if ($list->count() < $threshold) {
             $field = new DropdownField($this->name, $title, $list->map('ID', $titleField));
             $field->setEmptyString(' ');
         } else {
             $field = new NumericField($this->name, $title);
+
+            $field->setRightTitle(
+                'This field has been converted to a NumericField as there are at least ' . $threshold
+                . ' related object(s). This can be set with the config variable'
+                . ' SilverStripe\ORM\FieldTypeDBForeignKey::dropdown_field_threshold'
```
robbieaverill: I don't think that this description belongs on the field, since it's for CMS users rather than developers. This message likely to confuse the user more than anything.  What do you think?

robbieaverill: Oh, I see it was raised by @kinglozzer too. My vote is to remove this and leave the configurable limit in place without this message.

tractorcow: It was intended for phpdoc not for the user, right.

andrewandante: Well, no, not really. The problem raised was that the issue was invisible:

>This becomes extremely frustrating from a UX Point of View where the CMS works and looks nice, but if the client adds more than 100 objects then it breaks the experience.

I agree that we shouldn't be exposing the config stuff straight to the user, but it was complicated and not necessarily intuitive to find this particular block of code to figure out what was going on, so I'd argue that _some_ sort of message leaves a much better experience for both the CMS user and the dev.

Even something like "Too many objects for dropdown, fallback used" would help enormously in debugging, and give the user something useful to pass to their dev, who in turn has something to search for when they are looking to fix it.

Otherwise you end up with a field that seems to have mysteriously changed for what is an invisible reason, which is frustrating and confusing.

The other options I see as viable:

- move this functionality to `DropdownField` itself - which is where one would assume that the config var is located if you were to guess
- add some specific documentation around the issue and the config var

Ultimately up to you, but I think making this less "magical" is better

dhensby:
Degrading to a numericfield is a UX nightmare. How many content authors know the IDs of the related objects they need off the top of their head? or are able to map the titles to IDs without having to navigate away / inspect URLs, etc?

When clients/users see what used to be a dropdown become a numeric field for no obvious reason the reaction (naturally) is to think something has broken or even someone has changed something (like a dev deploying bad code).

Of course a dropdown that's so large it causes memory exhaustion or locks up the CMS is a worse experience (completely broken vs unfriendly) but we should be looking at a more friendly mid-point solution.

100 rows seems like a very small number (I'd vote to increase the limit to 1000 in this PR) and I support the idea that the CMS shows *some* kind of message to inform the user why they are experiencing a degraded experience.

tractorcow:
Just make an autocomplete field with ajax endpoint for searching records. We already do this with ListBox / TreeDropdownField; We just need a basic single select, no hierarchy version.

andrewandante:
@tractorcow would you propose this change to `DropdownField` itself? Or would you prefer to have a `AutocompleteDropdownField`, or allow `ListboxField` to `setMultiple(false)`? Or to set something for just this use-case?

I have to say that the idea of a searchable `DropdownField` appeals 😄

kinglozzer:
Short of someone taking the time to build an autocomplete field for core, I think this as good as it’s gonna get right now. I still don’t like showing a dev-related message to CMS authors, but in this context it probably does more good than harm. 

@tractorcow Do you agree that this is an improvement? Even if hopefully some kind soul will eventually replace this with an autocomplete field...

src/ORM/FieldType/DBForeignKey.php
```diff
@@ -63,11 +73,13 @@ public function scaffoldFormField($title = null, $params = null)
...
         // Don't scaffold a dropdown for large tables, as making the list concrete
         // might exceed the available PHP memory in creating too many DataObject instances
-        if ($list->count() < 100) {
+        $threshold = self::config()->get('dropdown_field_threshold');
+        if ($list->count() < $threshold) {
             $field = new DropdownField($this->name, $title, $list->map('ID', $titleField));
             $field->setEmptyString(' ');
         } else {
             $field = new NumericField($this->name, $title);
+            $field->setRightTitle('Too many related objects; fallback field in use');
```
tractorcow: Should be localised string not fixed english sentence.

tractorcow: Should be localised string not fixed english sentence.

tractorcow:
Current behat failure fixed with https://github.com/silverstripe/silverstripe-framework/pull/7860, not the fault of this PR.

PR #7794, opened 2018-01-24
NEW: Add support for ordinals in DBDate::Format()

kinglozzer:
E.g. `{$LastEdited.Format('EEEE d{o} MMMM YYYY')}` would display `Wednesday 24th January 2018`.

This solution only really works for ordinal suffixes. Apparently there are some languages which will [display the ordinal as a prefix](https://en.wikipedia.org/wiki/Ordinal_indicator#Representation_as_prefix), but from what I’ve seen true internationalisation is only really possible by using `DBDate::Full()` so I think that limitation is acceptable.

### DIFF
--- a/src/ORM/FieldType/DBDate.php
+++ b/src/ORM/FieldType/DBDate.php
@@ -217,7 +217,8 @@ class DBDate extends DBField
     }
 
     /**
-     * Return the date using a particular formatting string.
+     * Return the date using a particular formatting string. Use {o} to include an ordinal representation
+     * for the day of the month ("1st", "2nd", "3rd" etc)
      *
      * @param string $format Format code string. See http://userguide.icu-project.org/formatparse/datetime
      * @return string The date in the requested format
@@ -227,6 +228,12 @@ class DBDate extends DBField
         if (!$this->value) {
             return null;
         }
+
+        // Replace {o} with ordinal representation of day of the month
+        if (strpos($format, '{o}') !== false) {
+            $format = str_replace('{o}', "'{$this->DayOfMonth(true)}'", $format);
+        }
+
         $formatter = $this->getFormatter();
         $formatter->setPattern($format);
         return $formatter->format($this->getTimestamp());
--- a/tests/php/ORM/DBDateTest.php
+++ b/tests/php/ORM/DBDateTest.php
@@ -238,6 +238,21 @@ class DBDateTest extends SapphireTest
         $this->assertEquals('10th - 20th Oct 2000', $range);
     }
 
+    public function testFormatReplacesOrdinals()
+    {
+        $this->assertEquals(
+            '20th October 2000',
+            DBField::create_field('Date', '2000-10-20')->Format('{o} MMMM YYYY'),
+            'Ordinal day of month was not injected'
+        );
+
+        $this->assertEquals(
+            '20th is the 20th day of the month',
+            DBField::create_field('Date', '2000-10-20')->Format("{o} 'is the' {o} 'day of the month'"),
+            'Failed to inject multiple ordinal values into one string'
+        );
+    }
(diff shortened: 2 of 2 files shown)


src/ORM/FieldType/DBDate.php
```diff
@@ -227,6 +243,11 @@ public function Format($format)
         if (!$this->value) {
             return null;
         }
+
+        $format = preg_replace_callback('/{o(rd(inal)?)?}/i', function ($matches) {
```
dhensby: this seems unduly complex - can we not just use `{o}`?

kinglozzer: We could do, though I was originally going to just support `{ordinal}`: as we’re extending the syntax, I thought verbosity is probably a good thing as it’s much more obvious at a glance.

tractorcow: Then you don't need regular expression replace, just str_replace.

src/ORM/FieldType/DBDate.php
```diff
@@ -217,7 +217,23 @@ public function getISOFormat()
...
-     * Return the date using a particular formatting string.
+     * @return string
+     */
+    public function getDayOfMonthOrdinal()
+    {
+        $number = $this->Format('d');
+        if ($number) {
+            $formatter = NumberFormatter::create(i18n::get_locale(), NumberFormatter::ORDINAL);
+            $ordinal = str_replace($number, '', $formatter->format($number)); // Remove number, leaving just ordinal
+            return "'{$ordinal}'";
```
tractorcow: Why is the ordinal single quoted?

tractorcow: Oh I see, it's to escape for text.

http://userguide.icu-project.org/formatparse/datetime

The problem is that if you use this function by itself, it'll be quoted incorrectly.

src/ORM/FieldType/DBDate.php
tractorcow: Then you don't need regular expression replace, just str_replace.

tractorcow: Add the quotes around your string replacement here instead of in getDayOfMonthOrdinal().

src/ORM/FieldType/DBDate.php
```diff
@@ -217,7 +217,23 @@ public function getISOFormat()
     }
 
     /**
-     * Return the date using a particular formatting string.
+     * @return string
+     */
+    public function getDayOfMonthOrdinal()
+    {
+        $number = $this->Format('d');
```
tractorcow: Why is this method needed instead of just calling DayOfMonth(true) ?

tests/php/ORM/DBDateTest.php
```diff
@@ -238,6 +238,19 @@ public function testDayOfMonth()
         $this->assertEquals('10th - 20th Oct 2000', $range);
     }
 
+    public function testFormatReplacesOrdinals()
+    {
+        $this->assertEquals('20th', DBField::create_field('Date', '2000-10-20')->Format('d{o}'));
```
tractorcow: If you make {o} include the number itself you don't need any new code, just use DayOfMonth(true).

kinglozzer: I’m trying that, and getting segfaults 😕. Every time I call a method inside `Format()` that _isn’t_ prefixed with `get` I get a segfault...

kinglozzer: Oh of course, recursion. `DayOfMonth` calls `Format()`. I can probably guard against that with a `strpos()`

kinglozzer:
Okay, switched to a simple `str_replace` that now only replaces `{o}`

PR #7795, opened 2018-01-24
FIX: Fix UploadField cuts off “Save” button (closes #2862)

### LINKED ISSUE #2862: UploadField iframe incorrect height
Spotted this issue when using the ‘focuspoint’ module, doesn’t seem to happen with just the usual form fields - only when fields are added (see screenshot).

Note that this can only be partially observed post 3.1.2 - `fitHeight()` is triggered twice, the second time calculates the correct height (#2863).

The issue is caused by setting the form’s width to be 98% [here](https://github.com/silverstripe/silverstripe-framework/blob/3.1/javascript/UploadField.js#L499). We’re setting this width _after_ we’ve calculated the height - meaning that the calculated height may be (and is) wrong. Closing and re-opening the edit form will fix it, as it can then calculate the height with the width already set to 98%.

I’m not really sure why this width is set at all, but either removing it or moving it to before we find the form’s height fixes the issue.

screen shot 2014-02-18 at 16 33 53

kinglozzer:


dhensby:
#2862 - wow, that's quite a low number..

Issue #7724, opened 2017-12-20
Start with default password requirements


blueo:
## Affected Version
v4

## Description

It appears that there are no password requirements set by default in ss4. I'd suggest having at least a minimum character length set. It would also be great to be able to do this from a .yml, currently PasswordValidator's `$minLength, $minScore, $testNames, $historicalPasswordCount` properties aren't configurable from yaml. 

## Steps to Reproduce
install SS4, try to set a password and notice it can be set to a single character.

## Acceptance Criteria
- In 4.x, it only works on new projects (that don't have their own mysite/_config.php already).
- In 5.x, we'd move all of this out of PHP into YAML, at which point it wouldn't matter if you upgraded with an existing mysite/_config.php - you'll get that new default behaviour anyway, unless you opt out.
- default should be simple (e.g. min length/historical passwords)

## Pull requests
- [x] https://github.com/silverstripe/silverstripe-versioned/pull/102 (bugfix for upstream regression)
- [x] https://github.com/silverstripe/recipe-core/pull/17
- [x] https://github.com/silverstripe/silverstripe-framework/pull/7800
- [x] https://github.com/silverstripe/silverstripe-framework/pull/7801

robbieaverill:
Yeah, this is something that CWP does via `_config.php`. It'd be nice to have an option in YAML too though.

chillu:
Agreed that we should make defaults safer here, not just a subset of installations (like the CWP installer). The question is what is considered a common baseline. In New Zealand we have the NZISM requirements, which CWP is based on:

```php
$pwdValidator = new PasswordValidator();
$pwdValidator->minLength(8);
$pwdValidator->checkHistoricalPasswords(6);
$pwdValidator->characterStrength(3, array("lowercase", "uppercase", "digits", "punctuation"));
Member::set_password_validator($pwdValidator);
```
So in natural language, it means "at least 8 characters, satisfying at least three of these four rules: lowercase, uppercase, digits, punctuation. You can't reuse any of your last six passwords. There's no expiry time for a password.".
https://gitlab.cwp.govt.nz/cwp/cwp-core/blob/master/_config.php#L113

It won't lock out existing users, but kick in when they either create or change their password. You could argue that this doesn't make it an API change in the strictest sense. I assume we can reliably execute this PHP before the "project" `_config.php` (usually `mysite/_config.php`) kicks in though, so you can always override these defaults.

kinglozzer:
[...]

PR #7800, opened 2018-01-24
Enhancement Updated PasswordValidator to fallback to config options (ss4)

flamerohr:
This pull request goes to the `4` branch and still retains instance variables to not break semver.

For more details see: https://github.com/silverstripe/silverstripe-framework/issues/7724

### DIFF
--- a/src/Security/PasswordValidator.php
+++ b/src/Security/PasswordValidator.php
@@ -3,7 +3,9 @@
 namespace SilverStripe\Security;
 
 use SilverStripe\Core\Config\Configurable;
+use SilverStripe\Core\Extensible;
 use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\ORM\ValidationResult;
 
 /**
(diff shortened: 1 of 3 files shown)


src/Security/PasswordValidator.php
```diff
@@ -22,33 +24,73 @@ class PasswordValidator
...
+    /**
+     * @deprecated 5.0
      * Minimum password length
      *
      * @param int $minLength
      * @return $this
      */
     public function minLength($minLength)
     {
+        Deprecation::notice('5.0', 'Use ->config()->set(\'min_length\', $value) instead.');
```
tractorcow: I don't think we should deprecate the instance setters; Remember that config modifications have a cost, and it's entirely possible that you could create multiple instances of PasswordValidator.

robbieaverill: I'm in favour of deprecating _this one_ though, in favour of `setMinLength()` replacement

flamerohr: yeah true, I'm also in favour of deprecating **this** for `setMinLength()` and other more semantic method names.

tractorcow:
Rebasing to see if it fixes behat tests.

tractorcow:
Versioned needs a patch https://github.com/silverstripe/silverstripe-versioned/pull/102

tractorcow:
Behat is broken upstream, not the fault of this PR.

PR #7801, opened 2018-01-24
Enhancement Updated PasswordValidator to fallback to config options (ss5)

flamerohr:
This pull request goes to the `master` (pseudo-`5`) branch and removes the instance variables retained in https://github.com/silverstripe/silverstripe-framework/pull/7800

For more details see: https://github.com/silverstripe/silverstripe-framework/issues/7724

### DIFF
--- a/src/Security/PasswordValidator.php
+++ b/src/Security/PasswordValidator.php
@@ -3,40 +3,51 @@
 namespace SilverStripe\Security;
 
 use SilverStripe\Core\Config\Configurable;
+use SilverStripe\Core\Extensible;
 use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\ORM\ValidationResult;
 
 /**
  * This class represents a validator for member passwords.
- *
- * <code>
- * $pwdVal = new PasswordValidator();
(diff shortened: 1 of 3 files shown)


dhensby:
test failures need addresing

src/Security/PasswordValidator.php
```diff
@@ -3,52 +3,65 @@
...
+     * @deprecated 5.0
      * Minimum password length
      *
      * @param int $minLength
      * @return $this
      */
     public function minLength($minLength)
     {
-        $this->minLength = $minLength;
+        Deprecation::notice('5.0', 'Use ->config()->set(\'min_length\', $value) instead.');
```
robbieaverill: You've deprecated these in 4.x via https://github.com/silverstripe/silverstripe-framework/pull/7800 - should these be removals in the master branch instead?

flamerohr:
Closing in favour of keeping the instance variables

src/Security/PasswordValidator.php
flamerohr: I'd remove them in 5.1 as a preference - 5.0 means that the method is highlighted with a notice
But yeah could swing either way.

PR #7804, opened 2018-01-26
BUG $this->tableName in CompositeDBField->addToQuery is blank.

beerbohmdo:
`$this->tableName` is currently just blank in `addToQuery()` method of a CompositeDBField, because the `setTable()`-method is not called.

This is already fixed in 4.0 branch but not in 3.6.


dhensby:
What about 3.5?

beerbohmdo:
Yes it is also affected.
Should I change the pull request against 3.5?

dhensby:
Please :) I'll rebase it up into 3.6 after

tractorcow:
Forcing a rebuild since the branch was changed.

tractorcow:
Behat was passing but stalled half way for some reason. Restarting.

PR #7805, opened 2018-01-26
Make sure format for the string matches DB format

pitchandtone:
Currently the code attempts to insert '["Thurs","Fri","Sat"]' whereas the format should be "Thurs,Fri,Sat"

### DIFF
--- a/src/ORM/FieldType/DBMultiEnum.php
+++ b/src/ORM/FieldType/DBMultiEnum.php
@@ -76,4 +76,12 @@ class DBMultiEnum extends DBEnum
 
         return $field;
     }
+    
+    public function prepValueForDB($value)
+    {
+        $value = parent::prepValueForDB($value);
+        if ($value) {
+            return str_replace(array('[',']','"'), '', $value);
+        }
+    }
 }


pitchandtone:
This field needs more testing, these changes make sure the field goes into the DB, but it still needs work coming back out of the DB and pre-filling a checkboxlist.

tractorcow:
It should join the values together with `,` shouldn't it?

pitchandtone:
Yeah it uses a MYSQL SET which is “Value1,Value2,Value3”

While the code i submitted gets the data in i wasn’t sure if i’d handled it in the right way.

Next issue was to either transform it as it comes back which I couldn’t find a pattern for, or make checkbox field handle a comma separated value which seemed a fragile route.

So keen for input.

pitchandtone:
Should I be using this instead:
https://github.com/symbiote/silverstripe-multivaluefield ?

tractorcow:
This field should work.

maxime-rainville:
This got fixed independently by https://github.com/silverstripe/silverstripe-framework/pull/8499

PR #7808, opened 2018-01-29
FIX Allow cleanup marker regex to handle self closing HTML5 tags

robbieaverill:
When using self closing HTML5 tags, the regex to find left over marker classes in content and replace them with valid tags is looking for at least one character between `data-tagid="123"` and `>`, which isn't the case when using something like [silverstripe/html5](https://github.com/silverstripe/silverstripe-html5).

The result, for example in CWP 2.0, is that images embedded in content disappear when saving the page in the CMS, and are not rendered on the frontend. Instead you see `` instead.

The change is "at least one" to "any" for the regex quantifier.

### DIFF
--- a/src/View/Parsers/ShortcodeParser.php
+++ b/src/View/Parsers/ShortcodeParser.php
@@ -726,7 +726,7 @@ class ShortcodeParser
             $content = preg_replace_callback(
                 // Not a general-case parser; assumes that the HTML generated in replaceElementTagsWithMarkers()
                 // hasn't been heavily modified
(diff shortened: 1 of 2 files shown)


robbieaverill:
It’s a nasty method to test, but I can add a pseudo functional test perhaps :-)

tractorcow:
You'd need a shortcode that adds `` right?

robbieaverill:
Ok, added a test which fails in current 4.0 (with HTML5 module installed) but passes with the patch in place.Ok, added a test which fails in current 4.0 (with HTML5 module installed) but passes with the patch in place.

PR #7812, opened 2018-01-29
DOCS: Update nginx configuration for SilverStripe 4

oddnoc:
Port nginx configuration changes from 3.

### DIFF
--- a/docs/en/00_Getting_Started/01_Installation/How_To/Configure_Nginx.md
+++ b/docs/en/00_Getting_Started/01_Installation/How_To/Configure_Nginx.md
@@ -44,7 +44,7 @@ server {
     try_files $uri =404;
   }
 
-  location ~ /framework/.*(main|rpc|tiny_mce_gzip)\.php$ {
+  location /index.php {
     fastcgi_buffer_size 32k;
     fastcgi_busy_buffers_size 64k;
     fastcgi_buffers 4 32k;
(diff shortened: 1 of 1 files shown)


dhensby:
We'll get these merged up so we don't need a PR for it

oddnoc:
@dhensby There’s a crucial difference for SilverStripe 4: the try_files directive uses index.php rather than main.php.

dhensby:
ok, @oddnoc have you also updated the nginx wiki: https://www.nginx.com/resources/wiki/start/topics/recipes/silverstripe/ ?

dhensby:
OK, I've cherry-picked the commit and put it on the 4.0 branch so we don't get all those unrelated commits in this PR.

Please can you recheck this as I had to resolve some conflicts and they looked quite major and I just blindly accepted your changes

oddnoc:
@dhensby I think I see part of the problem. I meant to use 4, not 4.0, as the base, and now I've updated the base. But I don't see where you cherry-picked a commit into 4.0 (or 4).

Hadn't thought about the wiki, so thanks for the reminder. But I'll be happy to update the nginx wiki once the SilverStripe project has settled on the sample config.

dhensby:
> @dhensby I think I see part of the problem. I meant to use 4, not 4.0, as the base, and now I've updated the base. But I don't see where you cherry-picked a commit into 4.0 (or 4).

Right, but what's wrong with putting it on 4.0, is the change not applicable to 4.0 and only to 4.x?

Was there something wrong with the changes I had made to the PR?

>Hadn't thought about the wiki, so thanks for the reminder. But I'll be happy to update the nginx wiki once the SilverStripe project has settled on the sample config.

Ok

oddnoc:
Would you rather have the PR against the 4.0 branch? Happy to accommodate. The new try_files would apply equally to any 4.x, including 4.0.

robbieaverill:
For reference, we often ask for docs changes to go into the 4 branch because that's the branch that we use for generating docs.silverstripe.org. If you merge into 4.0 it will still end up in the 4 branch eventually, it'll just take longer to get there

PR #7816, opened 2018-01-30
Update phpdoc for FormField::setRightTitle regarding escaped HTML

### LINKED ISSUE #7813: FormField::setRightTitle expects escaped HTML but escapes it anyway
## Affected Version

4.1.x-dev

## Description

The PHPDoc for `FormField::setRightTitle` indicates that the input data should already be escaped. This would indicate that I could pass HTML into it at my own risk and have it render correctly (without being escaped again).

I'm not sure whether this is a PHPDoc issue, in that we _do_ want the input to be escaped in the template, or a bug in that the output is escaped when it shouldn't be.

## Steps to Reproduce

Failing unit test:

```php
public function testSetRightTitle()
{
    $field = new FormField('Test');
    $field->setRightTitle('Some <strong>escaped</strong> text');
    $result = $field->FieldHolder()->forTemplate();
    $this->assertContains('<strong>escaped</strong>', $result);
}
```

## Fix

We can cast as HTMLFragment instead of Text (which escapes itself):

```diff
diff --git a/src/Forms/FormField.php b/src/Forms/FormField.php
index 7a9efd5d4..c31d9684d 100644
--- a/src/Forms/FormField.php
+++ b/src/Forms/FormField.php
@@ -277,7 +277,7 @@ class FormField extends RequestHandler
         'isReadOnly' => 'Boolean',
         'HolderID' => 'Text',
         'Title' => 'Text',
-        'RightTitle' => 'Text',
+        'RightTitle' => 'HTMLFragment',
         'Description' => 'HTMLFragment',
     );
 
```

robbieaverill:
Resolves #7813

### DIFF
--- a/src/Forms/FormField.php
+++ b/src/Forms/FormField.php
@@ -490,7 +490,7 @@ class FormField extends RequestHandler
      * Gets the contextual label than can be used for additional field description.
      * Can be shown to the right or under the field in question.
      *
-     * @return string Contextual label text.
+     * @return string Contextual label text
      */
     public function RightTitle()
     {
@@ -499,9 +499,8 @@ class FormField extends RequestHandler
 
     /**
      * Sets the right title for this formfield
-     * Note: This expects escaped HTML.
      *
-     * @param string $rightTitle Escaped HTML for title
+     * @param string|DBField Plain text string, or a DBField with appropriately escaped HTML
      * @return $this
      */
     public function setRightTitle($rightTitle)


src/Forms/FormField.php
```diff
@@ -490,18 +490,17 @@ public function setTitle($title)
      * Gets the contextual label than can be used for additional field description.
      * Can be shown to the right or under the field in question.
      *
-     * @return string Contextual label text.
+     * @return string Contextual label text (HTML escaped)
```
tractorcow: The getter doesn't return the HTML escaped value, so that's not correct either.

src/Forms/FormField.php
```diff
@@ -490,18 +490,17 @@ public function setTitle($title)
...
      */
     public function RightTitle()
     {
         return $this->rightTitle;
     }
 
     /**
-     * Sets the right title for this formfield
-     * Note: This expects escaped HTML.
+     * Sets the right title for this formfield. Note: Any HTML passed in will be escaped.
```
tractorcow: Rather, document that you can pass in a DBField instance of the appropriate escape type to override the default escape.

```
@param string|DBField $rightTitle Plain text right title, or a DBField with the appropriate escape type
```

robbieaverill:
Thanks @tractorcow - updated. How does that look instead?

Issue #7818, opened 2018-01-30
RFC: Improve behaviour of unsaved has_one relations


tractorcow:
This RFC proposes improvements and simplification to the has_one component behaviour on dataobjects, especially in relation to unsaved records.

Note: The below more or less apply to `belongs_to` as well.

# Current state

At the moment saving of relations works this way, with the following notes

* `$obj->SomeRelation()`
  - Will return an object, even if SomeRelationID is 0 (unless polymorphic, when it's null)
  - User code must use `->exists()` / `->isInDB()` on this object; More expensive and error prone than simple falsey check.
  - Related object will not be saved on $obj->write() unless the 4th write() parameter is set from false to true.

* `$obj->SomeRelationID = $obj`
  - Will assign the object to the given relation, even if unsaved
  - Related object will not be saved on $obj->write() unless the 4th write() parameter is set from false to true.

* `return $obj->SomeRelation`
  - Does nothing;

* `$obj->SomeRelation = $obj`
  - Does nothing, would expect it to assign the new has_one

* `$obj->write()` doesn't save any related objects.

# Proposal

* `$obj->SomeRelation` / `$obj->SomeRelation()`
 - would act as a proper getter / setter for has_one / belongs_to relations. (code into getField / setField)
 - Don't need `ID` suffix, but would support it if provided
 - Not return empty objects on getter if SomeRelationID is 0

* `$obj->SomeRelationID`
  - Acts as getter / setter for ID
  - If provided an object, set the internal component instead (as well as ID if set)

* `$obj->write()`
  - Since we no longer create dummy objects in the relation getter, we no longer need to opt-in to writing all existing relations; All components are written on write. Deprecate 4th parameter.

# 5.0 breaking changes

* `$obj->write()`
  - Remove $showDebug
  - Remove $writeComponent
  - Use write flags instead of extra parameters

# Links

Related: https://github.com/silverstripe/silverstripe-framework/issues/3832

tractorcow:
This came up because I'm writing a recursive duplication mechanism for records. has_ones don't support unsaved records properly, so I've had to hack this in for 4.0.

```php
$item = $sourceObject->getComponent($relation);
if (!$item->isInDB()) {
    return;
}
$clonedItem = $item->duplicate(false);
$destinationObject->setField($relation.'ID', $clonedItem);

// After destination is written, assign this item the given ID
$destinationObject->afterExtending('onAfterWrite', function() use ($destinationObject, $clonedItem) {
[...]

PR #7819, opened 2018-01-31
API Implement cascade_duplications

tractorcow:
Parent story https://github.com/silverstripe/silverstripe-versioned/issues/28

Also partly implements fragments from https://github.com/silverstripe/silverstripe-framework/issues/7818 that I felt were necessary to support re-assignment of unary related objects.

Still need to write up docs.


tests/php/ORM/DataObjectDuplicationTest.php
```diff
@@ -2,29 +2,27 @@
 
 namespace SilverStripe\ORM\Tests;
 
-use SilverStripe\ORM\DataObject;
 use SilverStripe\Dev\SapphireTest;
-use SilverStripe\ORM\FieldType\DBDatetime;
 
 class DataObjectDuplicationTest extends SapphireTest
 {
+    protected static $fixture_file = 'DataObjectDuplicationtest.yml';
```
chillu: I find this test a lot harder to read with 37 lines of YAML fixtures for one big unit test. Mainly because the classes are just named sequentially (class1 to class7), rather than having a name that would imply their relation to each other. But that's just a continuation of the existing test, so I can't fault you for that :) Overall, think just creating the various associations in PHP based on YAML fixtures would improve readability quite a bit.

src/ORM/DataObject.php
```diff
@@ -279,6 +279,18 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
...
+     * List of relations that should be cascade duplicate.
+     * many_many duplications are shallow only.
+     *
+     * Note: If duplicating a many_many through you should refer to the
+     * has_many intermediary relation instead, otherwise extra fields
+     * will be omitted from the duplicated relation.
+     *
+     * @var array
+     */
+    private static $cascade_duplications = [];
```
chillu: Inconsistent naming? `cascade_deletes` vs. `cascade_deletions` vs. `cascade_duplications` vs. `cascade_duplicates`?

src/ORM/DataObject.php
```diff
@@ -444,27 +520,109 @@ protected function duplicateManyManyRelations($sourceObject, $destinationObject,
...
      * @param DataObject $destinationObject
-     * @param string $manyManyName
+     * @param string $relation
      */
-    protected function duplicateManyManyRelation($sourceObject, $destinationObject, $manyManyName)
+    protected function duplicateManyManyRelation($sourceObject, $destinationObject, $relation)
     {
         // Ensure this component exists on the destination side as well
-        if (!static::getSchema()->manyManyComponent(get_class($destinationObject), $manyManyName)) {
+        if (!static::getSchema()->manyManyComponent(get_class($destinationObject), $relation)) {
```
chillu: Shouldn't this throw an exception?

src/ORM/DataObject.php
```diff
@@ -444,27 +520,109 @@ protected function duplicateManyManyRelations($sourceObject, $destinationObject,
     }
 
     /**
-     * Duplicates a single many_many relation from one object to another
+     * Duplicates a single many_many relation from one object to another.
```
chillu: Point out that this doesn't copy many_many_extraFields? Needs to be mentioned in higher level docs as well

tests/php/ORM/DataObjectDuplicationTest.php
tractorcow: Ok now they are all animals.

Ask again and I'm making them chemical molecules. :)

src/ORM/DataObject.php
```diff
@@ -1582,6 +1741,52 @@ public function getComponent($componentName)
...
     }
 
+    /**
+     * Assign an item to the given component
+     *
+     * @param string $componentName
+     * @param DataObject|null $item
+     * @return $this
+     */
+    public function setComponent($componentName, $item)
```
chillu: Given the `getComponent/setComponent` symmetry, would it make sense to introduce `setComponents` in `getComponents/setComponents`? Probably not, just wondering if you had considered it.

src/ORM/DataObject.php
tractorcow: It's on the phpdoc of cascade_duplications

```php
    /**
     * List of relations that should be cascade duplicate.
     * many_many duplications are shallow only.
     *
     * Note: If duplicating a many_many through you should refer to the
     * has_many intermediary relation instead, otherwise extra fields
     * will be omitted from the duplicated relation.
     *
     * @var array
     */
    private static $cascade_duplications = [];
 ```

tractorcow: I'll add another note to the markdown docs too.

tractorcow: Hm, maybe I can fix the extra fields for non-through many_many.

src/ORM/DataObject.php
tractorcow: Good idea, I'd just copied it but let's be more strict. :)

src/ORM/DataObject.php
tractorcow: It's a bit harder, because we don't assign relations directly, we'd need to convert values from SS_List -> internal relation. It's out of scope of this change.

tractorcow:
Ok, things to do:

 - Have a go at duplicating extra_fields (if possible). It may be a lot of extra work if (for example) UnsavedRelations don't support it properly.
 - Rename cascade_duplications -> cascade_duplicates
 - Exception on relation duplication methods if not a real duplication (instead of no-op).
 
 I'm not sure about setComponents() for now.

tractorcow:
Actually that wasn't that hard... implemented. ;)

src/ORM/DataObject.php
tractorcow: I think we could implement a `Relation::copyTo(Relation $dest)` in 5.0

tractorcow:
Ah, found a regression in UploadField which I've just patched.

src/ORM/DataObject.php
chillu: All good, just wanted to hear your take on it

src/ORM/DataObject.php
chillu: No need to add the extra field duplications, but it's important to point out (as you've done in markdown)

PR #7821, opened 2018-01-31
Update reference from $databaseConfig to .env

tractorcow:
Because the message is so outdated

### DIFF
--- a/src/Core/CoreKernel.php
+++ b/src/Core/CoreKernel.php
@@ -308,7 +308,7 @@ class CoreKernel implements Kernel
         // Error if installer not available
         if (!file_exists($this->basePath . '/install.php')) {
             throw new HTTPResponse_Exception(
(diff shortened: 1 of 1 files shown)


src/Core/CoreKernel.php
```diff
@@ -308,7 +308,7 @@ protected function redirectToInstaller()
         // Error if installer not available
         if (!file_exists($this->basePath . '/install.php')) {
             throw new HTTPResponse_Exception(
-                'SilverStripe Framework requires a $databaseConfig defined.',
+                'SilverStripe Framework requires a .env configuration file',
```
flamerohr: maybe mention environment variables as an alternative as well?

flamerohr: preferred for dev environments right?
I remember production environments are against using it

tractorcow: No that was the old dotenv library; We don't use that anymore.

PR #7825, opened 2018-02-01
BUG Fix critical issue with incorrectly saved session data

tractorcow:
Removed / cleared information was not being correctly unset from $_SESSION on save. Data could be unset via the following ways:

 - Session::set() being called with a replacement array, with a set of keys which may not match existing data
 - Session::clear() being called on either a top level item or a nested item.
 
 This fix ensures that:
  - Partial unsets on arrays are properly completed
  - clear() actually removes these values from the resulting $_SESSION, and doesn't simply leave hanging array keys with null values
  
New methods are marked as `@internal` because otherwise it's not semver.

### DIFF
--- a/src/Control/Session.php
+++ b/src/Control/Session.php
@@ -4,6 +4,7 @@ namespace SilverStripe\Control;
 
 use BadMethodCallException;
 use SilverStripe\Core\Config\Configurable;
+use SilverStripe\Dev\Deprecation;
 
 /**
  * Handles all manipulation of the session.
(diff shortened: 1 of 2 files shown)


src/Control/Session.php
```diff
@@ -491,19 +481,22 @@ public function save(HTTPRequest $request)
...
      * Recursively apply the changes represented in $data to $dest.
      * Used to update $_SESSION
      *
+     * @deprecated 4.1...5.0 Will be removed in 5.0
      * @param array $data
      * @param array $dest
      */
     protected function recursivelyApply($data, &$dest)
     {
+        Deprecation::notice('5.0', 'Will be removed in 5.0');
```
flamerohr: Could you suggest `recursivelyApplyChanges` as the replacement/alternative?

tractorcow:
Ok done @flamerohr

dhensby:
is 3.x not affected?

flamerohr:
@tractorcow mentioned this dates back to 2.x days... so most likely will need to prod him to backport it next week :)

PR #7829, opened 2018-02-04
[bugfix] $request == null breaks

Firesphere:
The $request incoming as null was not properly detected by the if/elseif structure.

Observed behaviour:
- A call without a request is arriving at login throws a 500
Expected behaviour:
- The controller should attempt to read the request from itself before breaking

### DIFF
--- a/src/Security/Security.php
+++ b/src/Security/Security.php
@@ -651,7 +651,7 @@ class Security extends Controller implements TemplateGlobalProvider
     {
         if ($request) {
             $this->setRequest($request);
-        } elseif ($request) {
+        } elseif ($this->getRequest()) {
             $request = $this->getRequest();
         } else {
             throw new HTTPResponse_Exception("No request available", 500);


src/Security/Security.php
```diff
@@ -651,7 +651,7 @@ public function login($request = null, $service = Authenticator::LOGIN)
     {
         if ($request) {
             $this->setRequest($request);
-        } elseif ($request) {
+        } elseif (!$request) {
```
tractorcow: ` elseif ($this->getRequest())`

Firesphere: Would you agree it should be `elseif (!$request && $this->getRequest())`?

tractorcow: Why do you need to check if $request is falsey twice?

Firesphere: No, nevermind, `if ($request)` should be hit anyway, so yep, just a sec :)

tractorcow:
Merge on green.

PR #7833, opened 2018-02-05
Move CSV writing/reading to league/csv library

dhensby:
Moved the writing and reading of CSVs to the [`league/csv`](https://github.com/thephpleague/csv) library for better compatibility.

Improvements:
- Add UTF-8 BOM to start of CSV files
- Unified and standard way to read/write CSVs
- Push CSV code maintenance to external library

### DIFF
--- a/src/Dev/CsvBulkLoader.php
+++ b/src/Dev/CsvBulkLoader.php
@@ -74,9 +74,20 @@ class CsvBulkLoader extends BulkLoader
             $filepath = Director::getAbsFile($filepath);
             $csvReader = Reader::createFromPath($filepath, 'r');
 
+            $tabExtractor = function ($row, $rowOffset, $iterator) {
+                foreach ($row as &$item) {
+                    // [SS-2017-007] Ensure all cells with leading tab and then [@=+] have the tab removed on import
+                    if (preg_match("/^\t[\-@=\+]+.*/", $item)) {
+                        $item = ltrim($item, "\t");
+                    }
+                }
+                return $row;
+            };
+
             if ($this->columnMap) {
                 $headerMap = $this->getNormalisedColumnMap();
-                $remapper = function ($row, $rowOffset, $iterator) use ($headerMap) {
+                $remapper = function ($row, $rowOffset, $iterator) use ($headerMap, $tabExtractor) {
+                    $row = $tabExtractor($row, $rowOffset, $iterator);
                     foreach ($headerMap as $column => $renamedColumn) {
                         if ($column == $renamedColumn) {
                             continue;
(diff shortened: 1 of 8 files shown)


robbieaverill:
Yay! +1

src/Dev/CsvBulkLoader.php
```diff
@@ -67,37 +67,74 @@ public function preview($filepath)
...
+    {
+        $map = [];
+        foreach ($this->columnMap as $column => $newColumn) {
+            if (strpos($newColumn, "->") === 0) {
+                $map[$column] = $column;
+            } elseif (is_null($newColumn)) {
+                // the column map must consist of unique scalar values
+                // `null` can be present multiple times and is not scalar
+                // so we name it in a standard way so we can remove it later
+                $map[$column] = '_ignore_' . $column;
```
kinglozzer: Looks like this is new. If `null` isn’t supported, should it raise a notice of some sort?

kinglozzer:
Looking good 👍

src/Dev/CsvBulkLoader.php
dhensby: this is for compatibility with `league/csv`, rather than changing the API for SS users.

This keeps the existing SS API / behaviour.

tractorcow:
Can you confirm that SS-2017-007 is not a bug in the Reader library? The existing tests that cover this still run, but they only run on the old deprecated library, not the new reader. Looking at the code it doesn't seem to be considered.

To review, the fix was the add leading tabs to any risky cell value, and to also ensure that the reader trimmed those leading tabs.

We'll need to enforce the same in this library somehow.

tractorcow:
Relevant link https://github.com/thephpleague/csv/issues/253

dhensby:
We still have all the same sanitisation logic but I'll make sure that we cover it in tests

src/Forms/GridField/GridFieldExportButton.php
```diff
@@ -167,6 +167,17 @@ public function generateExportFileData($gridField)
         $csvWriter->setEnclosure($this->getCsvEnclosure());
         $csvWriter->setNewline("\r\n"); //use windows line endings for compatibility with some csv libraries
         $csvWriter->setOutputBOM(Writer::BOM_UTF8);
+        $csvWriter->addFormatter(function (array $row) {
+            if (!Config::inst()->get(get_class($this), 'xls_export_disabled')) {
```
tractorcow: You could put this config condition on the outside of the `addFormatter` call to cut down a large number of config api calls.

dhensby: Done

dhensby:
hell yes

Issue #6613, opened 2017-02-10
exclude acts like excludeAny not exclude"ALL" in 3.5.2


sunnysideup:
Here is the php:

```php
$list = $list
                ->filter(
                    array(
                        "CancelledByID" => 0
                    )
                )
                ->exclude(
                    array(
                        'ID' => array(1,2,3),
                        'StatusID' => array(2,4,5)
                    )
                );

```

the resulting mysql is:

```sql

SELECT DISTINCT Bla 
FROM "Order" 
WHERE 
    ("Order"."CancelledByID" = ?) AND 
    (
         ("Order"."ID" NOT IN (?, ?, ?)
    ) *** OR ****
         ("Order"."StatusID" NOT IN (?, ?, ?)
    )
) 
ORDER BY "Order"."LastEdited" DESC

```

From what I understand, and if I had to bet, I'd say I am wrong ;-)))))) 

 *** OR *** should be *** AND ***.

i.e. I want to select all Orders, excluding Orders with a certain `IDs` and certain `StatusIDs`.

dhensby:
@sunnysideup I think you're right - it's always confused me how we don't have `excludeAny`.

What happens if you do:

```php
$list->exclude([
    'ID' => [1, 2, 3],
])->exclude([
    'StatusID' => [1, 2, 3],
]);
```

sunnysideup:
Yeah - that is how I solved it...

Is there anyway we can fix this properly in 4.0?

dhensby:
This should be fixed, yes. I don't agree with deprecating `exclude` we should just make it work as expected.

sunnysideup:
can it still go into 4.0?  What will happen to all the people who know exclude as the way it works now.  It will be crazy confusing for them that it works the opposite way in 4.0???

sunnysideup:
If it can go into 4.0 then if you tell me which branch to use, etc... then I am happy to code it (should only be a few lines - famous last words ;-)))

dhensby:
Yes, it can go into 4. The master branch. 

We'll put something in the upgrade notes. But exclude is clearly working contrary to expectations.

tractorcow:
Could we just make `exclude` behave as `exclude all` (mirror of `filter`) and add `excludeAny` with the old behaviour of `exclude`? That way we have the same method naming and behaviour between filter/exclude methods.

sunnysideup:
This has actually caught me out a second time now in a project...

I made a start with a fix:

[...]

PR #7838, opened 2018-02-06
add excludeAny() and tests for complicated excludes/filters

andrewandante:
Proves #6613 is behaving correctly.

Also adds `excludeAny()` to counterpart `exclude()` in the same way as `filterAny()` and `filter()`. This would be the option that devs have tried to use previously, and had to chain `exclude()`s

### DIFF
--- a/src/ORM/DataList.php
+++ b/src/ORM/DataList.php
@@ -598,7 +598,9 @@ class DataList extends ViewableData implements SS_List, Filterable, Sortable, Li
      *
      * @todo extract the sql from this method into a SQLGenerator class
      *
-     * @param string|array Escaped SQL statement. If passed as array, all keys and values will be escaped internally
+     * @param string|array
+     * @param string [optional]
+     *
      * @return $this
      */
     public function exclude()
@@ -634,9 +636,9 @@ class DataList extends ViewableData implements SS_List, Filterable, Sortable, Li
      * @example $list = $list->excludeAny(array('Name'=>array('bob','phil'), 'Age'=>array(21, 43)));
      *          // bob, phil, 21 or 43 would be excluded
      *
-     * @todo extract the sql from this method into a SQLGenerator class
+     * @param string|array
+     * @param string [optional]
      *
-     * @param string|array Escaped SQL statement. If passed as array, all keys and values will be escaped internally
      * @return $this
      */
     public function excludeAny()
--- a/src/ORM/DataList.php
+++ b/src/ORM/DataList.php
@@ -653,14 +653,14 @@ class DataList extends ViewableData implements SS_List, Filterable, Sortable, Li
         } else {
             throw new InvalidArgumentException('Incorrect number of arguments passed to excludeAny()');
         }
-        $list = $this;
 
-        foreach ($whereArguments as $field => $value) {
-            $filter = $this->createSearchFilter($field, $value);
-            $list = $list->alterDataQuery(array($filter, 'exclude'));
-        }
-
-        return $list;
+        return $this->alterDataQuery(function (DataQuery $dataQuery) use ($whereArguments) {
+            foreach ($whereArguments as $field => $value) {
+                $filter = $this->createSearchFilter($field, $value);
+                $filter->exclude($dataQuery);
+            }
+            return $dataQuery;
+        });
     }
 
     /**
--- a/src/ORM/DataList.php
+++ b/src/ORM/DataList.php
@@ -587,9 +587,8 @@ class DataList extends ViewableData implements SS_List, Filterable, Sortable, Li
     }
 
     /**
-     * Return a copy of this list which does not contain any items with these charactaristics
+     * Return a copy of this list which does not contain any items that match all params
      *
-     * @see SS_List::exclude()
      * @example $list = $list->exclude('Name', 'bob'); // exclude bob from list
      * @example $list = $list->exclude('Name', array('aziz', 'bob'); // exclude aziz and bob from list
      * @example $list = $list->exclude(array('Name'=>'bob, 'Age'=>21)); // exclude bob that has Age 21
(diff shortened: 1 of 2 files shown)


tests/php/ORM/DataListTest.php
```diff
@@ -1509,14 +1509,61 @@ public function testMultipleExclude()
...
 
+    /**
+     * Test doesn't exclude if only matches one
+     * $list->exclude(array('Name'=>'bob, 'Age'=>21)); // exclude bob that has Age 21
+     */
+    public function testMultipleExcludeMultipleMatches()
+    {
+        $list = TeamComment::get();
+        $list = $list->exclude(array('Name'=>'Bob', 'Comment'=>'Phil is a unique guy, and comments on team2'));
+        $this->assertEquals(3, $list->count());
```
dhensby: `assertCount`!

tests/php/ORM/DataListTest.php
```diff
@@ -1509,14 +1509,61 @@ public function testMultipleExclude()
...
+    {
+        $list = TeamComment::get();
+        $list = $list->exclude(array(
+            'Name'=> array('Bob', 'Phil'),
+            'Comment'=> array(
+                'This is a team comment by Bob',
+                'Phil is a unique guy, and comments on team2'
+            )
+        ));
+        $this->assertEquals(1, $list->count());
```
dhensby: `assertCount`

tests/php/ORM/DataListTest.php
```diff
@@ -1509,14 +1509,61 @@ public function testMultipleExclude()
...
+        $this->assertEquals('Joe', $list->last()->Name);
```
dhensby: `assertListEquals([['Name' => 'Joe']])`

tests/php/ORM/DataListTest.php
```diff
@@ -1528,6 +1578,64 @@ public function testExcludeOnFilter()
...
+        $this->assertListEquals([['Name' => 'Phil']], $list);
+    }
+
+    /**
+     * Test that if a very complicated exclude() is applied to a filter(), the filter() is still preserved.
+     */
+    public function testVeryComplicatedExcludeOnFilter()
+    {
+        /**
+ * @var DataList $list
```
dhensby: indentation

src/ORM/DataList.php
```diff
@@ -625,6 +624,43 @@ public function exclude()
...
+     * Return a copy of this list which does not contain any items with any of these params
+     *
+     * @example $list = $list->excludeAny('Name', 'bob'); // exclude bob from list
+     * @example $list = $list->excludeAny('Name', array('aziz', 'bob'); // exclude aziz and bob from list
+     * @example $list = $list->excludeAny(array('Name'=>'bob, 'Age'=>21)); // exclude bob or Age 21
+     * @example $list = $list->excludeAny(array('Name'=>'bob, 'Age'=>array(21, 43))); // exclude bob or Age 21 or 43
+     * @example $list = $list->excludeAny(array('Name'=>array('bob','phil'), 'Age'=>array(21, 43)));
+     *          // bob, phil, 21 or 43 would be excluded
+     *
+     * @todo extract the sql from this method into a SQLGenerator class
```
dhensby: is this relevant?

src/ORM/DataList.php
```diff
@@ -625,6 +624,43 @@ public function exclude()
...
+     *
+     * @param string|array Escaped SQL statement. If passed as array, all keys and values will be escaped internally
```
dhensby: this is wrong?

tests/php/ORM/DataListTest.php
```diff
@@ -1528,6 +1578,64 @@ public function testExcludeOnFilter()
...
+        $this->assertListEquals([['Name' => 'Phil']], $list);
+    }
+
+    /**
+     * Test that if a complicated exclude() is applied to a filter(), the filter() is still preserved.
+     */
+    public function testComplicatedExcludeOnFilter()
+    {
+        /**
+         * @var DataList $list
```
dhensby: Can this be an inline comment like `/** @var DataList $list */` please

src/ORM/DataList.php
```diff
@@ -625,6 +626,43 @@ public function exclude()
...
+        } elseif ($numberFuncArgs == 2) {
+            $whereArguments[func_get_arg(0)] = func_get_arg(1);
+        } else {
+            throw new InvalidArgumentException('Incorrect number of arguments passed to excludeAny()');
+        }
+        $list = $this;
+
+        foreach ($whereArguments as $field => $value) {
+            $filter = $this->createSearchFilter($field, $value);
+            $list = $list->alterDataQuery(array($filter, 'exclude'));
```
tractorcow: Each alterDataQuery creates a new object each time, for performance reason invert this (alterDataQuery on the outside, foreach() on the inside).

tractorcow: If you use a closure `function ($query) use ($filter) { /* loop */ $filter->exclude($query); /* end loop */ }` instead of an array `[$filter , 'exclude']` then we can infer usage from other parts of the code.

tests/php/ORM/DataListTest.php
tractorcow: This is already a typehint on `DataObject::get()` isn't it?

src/ORM/DataList.php
```diff
@@ -625,6 +626,43 @@ public function exclude()
...
+        } elseif ($numberFuncArgs == 2) {
+            $whereArguments[func_get_arg(0)] = func_get_arg(1);
+        } else {
+            throw new InvalidArgumentException('Incorrect number of arguments passed to excludeAny()');
+        }
+
+        return $this->alterDataQuery(function (DataQuery $dataQuery) use ($whereArguments) {
+            foreach ($whereArguments as $field => $value) {
+                $filter = $this->createSearchFilter($field, $value);
+                $filter->exclude($dataQuery);
```
tractorcow: good job ;D

PR #7839, opened 2018-02-06
Moving tests to use transactions

dhensby:
Attempting to use transactions for tests to try to increase speed / reduce the need to rebuild all fixtures for each test

---

Depends on #7848

### DIFF
--- a/src/Dev/SapphireTest.php
+++ b/src/Dev/SapphireTest.php
@@ -23,6 +23,7 @@ use SilverStripe\Core\Manifest\ClassLoader;
 use SilverStripe\Dev\Constraint\SSListContains;
 use SilverStripe\Dev\Constraint\SSListContainsOnly;
 use SilverStripe\Dev\Constraint\SSListContainsOnlyMatchingItems;
+use SilverStripe\Dev\State\FixtureTestState;
 use SilverStripe\Dev\State\SapphireTestState;
 use SilverStripe\Dev\State\TestState;
 use SilverStripe\i18n\i18n;
@@ -66,7 +67,7 @@ class SapphireTest extends PHPUnit_Framework_TestCase implements TestOnly
     /**
      * @var FixtureFactory
      */
-    protected static $fixtureFactory;
+    protected $fixtureFactory;
 
     /**
      * @var Boolean If set to TRUE, this will force a test database to be generated
@@ -148,7 +149,7 @@ class SapphireTest extends PHPUnit_Framework_TestCase implements TestOnly
     /**
      * State management container for SapphireTest
      *
-     * @var TestState
+     * @var SapphireTestState
      */
     protected static $state = null;
 
@@ -158,6 +159,14 @@ class SapphireTest extends PHPUnit_Framework_TestCase implements TestOnly
      * @var TempDatabase
      */
     protected static $tempDB = null;
+
+    public static function tempDB()
+    {
+        if (!static::$tempDB) {
+            static::$tempDB = TempDatabase::create();
+        }
+        return static::$tempDB;
+    }
 
     /**
      * Gets illegal extensions for this class
@@ -208,6 +217,16 @@ class SapphireTest extends PHPUnit_Framework_TestCase implements TestOnly
         return static::$fixture_file;
     }
 
+    public function getUsesDatabase()
+    {
+        return $this->usesDatabase;
+    }
+
+    public function getRequireDefaultRecordsFrom()
+    {
+        return $this->requireDefaultRecordsFrom;
+    }
+
     /**
      * Setup  the test.
      * Always sets up in order:
@@ -254,33 +273,11 @@ class SapphireTest extends PHPUnit_Framework_TestCase implements TestOnly
 
         $fixtureFiles = $this->getFixturePaths();
 
-        // Set up fixture
         if ($this->shouldSetupDatabaseForCurrentTest($fixtureFiles)) {
-            if (!static::$tempDB->isUsed()) {
-                static::$tempDB->build();
-            }
-
-            DataObject::singleton()->flushCache();
-
-            if (!static::$tempDB->hasStarted()) {
-                foreach ($this->requireDefaultRecordsFrom as $className) {
-                    $instance = singleton($className);
-                    if (method_exists($instance, 'requireDefaultRecords')) {
-                        $instance->requireDefaultRecords();
-                    }
-                    if (method_exists($instance, 'augmentDefaultRecords')) {
-                        $instance->augmentDefaultRecords();
-                    }
-                }
-
-                foreach ($fixtureFiles as $fixtureFilePath) {
-                    $this->loadFixture($fixtureFilePath);
-                }
-            }
-
+            /** @var FixtureTestState $fixtureState */
+            $fixtureState = static::$state->getStateByName('fixtures');
+            $this->setFixtureFactory($fixtureState->getFixtureFactory(static::class));
             $this->logInWithPermission('ADMIN');
-
-            static::$tempDB->startTransaction();
         }
 
         // turn off template debugging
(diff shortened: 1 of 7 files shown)


tractorcow:
Whoops a lot of extra errors here. :)

src/Dev/SapphireTest.php
```diff
@@ -389,17 +391,19 @@ public static function tearDownAfterClass()
...
         static::resetDBSchema();
+
+        static::$fixtureFactory = null;
     }
 
     /**
      * @return FixtureFactory
      */
-    public function getFixtureFactory()
+    public static function getFixtureFactory()
```
tractorcow: Semver major change.

src/Dev/SapphireTest.php
```diff
@@ -389,17 +391,19 @@ public static function tearDownAfterClass()
...
     {
-        if (!$this->fixtureFactory) {
-            $this->fixtureFactory = Injector::inst()->create(FixtureFactory::class);
+        if (!static::$fixtureFactory) {
+            static::setFixtureFactory(Injector::inst()->create(FixtureFactory::class));
         }
-        return $this->fixtureFactory;
+        return static::$fixtureFactory;
```
tractorcow: Why does this need to be static now? the fixture factory belongs to the class instance. It runs the risk of cross-test pollution.

src/Dev/SapphireTest.php
```diff
@@ -546,6 +549,8 @@ protected function tearDown()
 
         // Call state helpers
         static::$state->tearDown($this);
+
+        static::$tempDB->rollbackTransaction();
```
tractorcow: The rollbackTransaction() needs to be in an if() statement that matches the if() on setUp().

`if ($this->shouldSetupDatabaseForCurrentTest($fixtureFiles)) {`

Otherwise you're rolling back more than you're starting transaction.

src/ORM/Connect/TempDatabase.php
```diff
@@ -23,6 +23,8 @@ class TempDatabase
      */
     protected $name = null;
 
+    private $hasStarted = false;
```
tractorcow: protected, and phpdoc. It's not clear that it's a transaction.

src/Dev/SapphireTest.php
dhensby: So something interesting that I found out when doing this work is that PHPUnit internally creates a new  object instance for every test method. That means the fixtureFactory instance is not shared from test-to-test on the same class. It has to be a static to allow it to be shared across tests.

The alternative is to have a way of loading a fixture without writing to the DB.

dhensby:
OK - I've got the tests passing locally. With this patch the framework test suite passes in 4.5 mins, without it takes 10.2mins - so we are seeing a 56% improvement in speed. It also saved 8MB of memory (on PHP 7.1 - savings on 5.x will be much higher).

Whilst I appreciate the changing of methods to `static` is a semver violation, we should try to explore how we can get this into 4

tractorcow:
Awesome... how much of it is due to the static fixture factory do you think?

tractorcow:
> Whilst I appreciate the changing of methods to static is a semver violation, we should try to explore how we can get this into 4

Just make the new static methods a different name, and deprecate the old ones. ;)

tractorcow:
How do you feel about moving the entire test fixture system into a TestState? I had originally wanted to do this, but it was going to be a huge amount of work. It would reduce the SapphireTest api surface area a great deal. :)

dhensby:
>how much of it is due to the static fixture factory do you think?

I'm not sure - I think it's mostly savings because we aren't trashing, and re-inserting, the fixtures to the DB. I'm sure it's more a DB bottleneck than a YAML parsing one (though it may well be both).

> Just make the new static methods a different name, and deprecate the old ones. ;)

that doesn't stop it being a breaking change. Any classes that have overridden the current method will no longer have the effect they used to, therefore breaking people's code.

> How do you feel about moving the entire test fixture system into a TestState? I had originally wanted to do this, but it was going to be a huge amount of work. It would reduce the SapphireTest api surface area a great deal. :)

This was what I was thinking too... but it does seem like a lot of work too :P - it's an interesting idea.

This PR at the moment was a semi-proof-of-concept to see just what kind of speed up we could get.

dhensby:
OK - moved to a `TestState` and made minimal API changes to `SapphireTest` - I've had to add a few APIs to both `SapphireTestState` and `SapphireTest` to allow the `FixtureTestState` to access the information it needs about tests.

src/ORM/Connect/TempDatabase.php
```diff
@@ -68,6 +73,46 @@ public function isUsed()
...
+            static::getConn()->transactionStart();
+        }
+    }
+
+    /**
+     * Rollback a transaction (or trash all data if the DB doesn't support databases
+     */
+    public function rollbackTransaction()
+    {
+        if (static::getConn()->supportsTransactions()) {
```
tractorcow: Why does this not set hasStarted to false if it supports transactions?

dhensby: So `$hasStarted` is really "do we have the fixtures in the DB" and the only time we don't have the fixtures in the DB is when we clear the DB... it's more "have the tests started" rather than "are we in a transaction"

dhensby: clearAllData actually resets hasStarted so we don't really need it in this method at all

src/Dev/State/FixtureTestState.php
```diff
@@ -0,0 +1,229 @@
...
+     * @param SapphireTest $test
+     *
+     * @return bool
+     */
+    protected function testNeedsDB(SapphireTest $test)
+    {
+        $annotations = $test->getAnnotations();
+
+        // annotation explicitly disables the DB
+        if (array_key_exists('useDatabase', $annotations['method'])
```
tractorcow: I never even knew this was a thing.

dhensby: I saw the PR adding it a while ago - we really should add the annotations to appropriate tests in core tests :)

src/Dev/SapphireTest.php
```diff
@@ -253,31 +281,10 @@ protected function setUp()
...
-            }
-
-            foreach ($fixtureFiles as $fixtureFilePath) {
-                $fixture = YamlFixture::create($fixtureFilePath);
-                $fixture->writeInto($this->getFixtureFactory());
-            }
-
+            /** @var FixtureTestState $fixtureState */
+            $fixtureState = static::$state->getStateByName('fixtures');
+            $this->setFixtureFactory($fixtureState->getFixtureFactory(static::class));
```
tractorcow: Is this needed still? Doesn't `this->setFixtureFactory()` just send the factory back to the teststate?

tractorcow:
I think I need to do more testing.

src/Dev/SapphireTest.php
dhensby: we need to set `$this->fixutureFactory` (property on `SapphireTest`) because some tests access it directly rather than through the helper methods or `getFixtureFactory` :(

src/ORM/Connect/TempDatabase.php
```diff
@@ -183,6 +229,11 @@ public function deleteAll()
      */
     public function resetDBSchema(array $extraDataObjects = [])
     {
+        // pgsql doesn't allow schema updates inside transactions
```
dhensby: TBH, I think this is most DBMS as in MySQL alter tables cause an implicit commit for transactions too...

dhensby:
@chillu can we get some input on this - it's a real shame that this has missed a large number of builds (especially CWP ones) that could have highlighted any edge cases with this fix (as well as saving a lot of build time)

src/Dev/State/FixtureTestState.php
```diff
@@ -0,0 +1,229 @@
...
+     * @param SapphireTest $test
+     *
+     * @return string
+     */
+    protected function resolveFixturePath($fixtureFilePath, SapphireTest $test)
+    {
+        // Support fixture paths relative to the test class, rather than relative to webroot
+        // String checking is faster than file_exists() calls.
+        $isRelativeToFile
+            = (strpos('/', $fixtureFilePath) === false)
```
mikenz: `int strpos ( string $haystack , mixed $needle [, int $offset = 0 ] )`
Around the wrong way?

tractorcow: Ah yes.

I also like strstr() for doing simple is_in_string :P

dhensby: Hah, this has been lifted directly from `SapphireTest` - so fixing this may have unintended consequences...

I've pushed up a fix, though

src/Dev/State/FixtureTestState.php
```diff
@@ -0,0 +1,229 @@
...
+
+    /**
+     * Called on tear down
+     *
+     * @param SapphireTest $test
+     */
+    public function tearDown(SapphireTest $test)
+    {
+        if ($this->testNeedsDB($test)) {
+            $test::tempDB()->rollbackTransaction();
```
tractorcow: Ok, follow up to https://github.com/silverstripe/silverstripe-framework/issues/8182, we need to check if this rollback failed.

If it failed, we need to destroy the database instead. That will cause the next setUp() to run `$tmpDB->build();`. 

Easy fix @dhensby ?

tractorcow: You'll need to rebuild fixture factories as well, since those normally only get rebuilt on setuponce.

PR #7841, opened 2018-02-07
BUG Fix incorrect BASE_URL when webroot is parent of public folder

tractorcow:
This issue comes up when installing a project for the first time, and before setting `.env` SS_BASE_URL or configuring webroot properly.

If you navigate to `http://localhost/somefolder` the BASE_URL will be set to `/somefolder/public` if the public folder exists. It SHOULD be set to `/somefolder`, otherwise the rewrite won't work properly, and the user will be redirected to the public folder on the first page navigation, which should not actually be in the URL.

It also sometimes breaks the home page, since silverstripe can't find the relative URL if the base doesn't match the request_uri.

### DIFF
--- a/src/includes/constants.php
+++ b/src/includes/constants.php
@@ -107,7 +107,7 @@ if (!defined('BASE_URL')) {
 
         // Unless specified, use empty string for base in CLI
         if (in_array(php_sapi_name(), ['cli', 'phpdbg'])) {
-            return "";
+            return '';
         }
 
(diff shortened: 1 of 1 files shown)


tractorcow:
To test: Make sure you have a project with a public folder, and ensure you navigate to the base folder to public.

flamerohr:
slightly unrelated issue:
installer File permission checks still references `sandbox.dev/index.php` rather than `sandbox.dev/public/index.php`

tractorcow:
That other issue is fixed with https://github.com/silverstripe/silverstripe-framework/pull/7842

PR #7844, opened 2018-02-07
NEW Allow GridFieldConfig::addComponents to accept an array

robbieaverill:
When I tried to use this method I expected that it would accept an array.
This pull request allows an array as well as an arbitrary number of arguments to be passed.

### DIFF
--- a/src/Forms/GridField/GridFieldConfig.php
+++ b/src/Forms/GridField/GridFieldConfig.php
@@ -68,12 +68,12 @@ class GridFieldConfig
     }
 
     /**
-     * @param GridFieldComponent $component,... One or more components
+     * @param GridFieldComponent|GridFieldComponent[] $component,... One or more components, or an array of components
      * @return $this
      */
     public function addComponents($component = null)
     {
-        $components = func_get_args();
+        $components = is_array($component) ? $component : func_get_args();
         foreach ($components as $component) {
             $this->addComponent($component);
         }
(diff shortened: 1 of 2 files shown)


dhensby:
To be honest, now we can use array expansion as a way of passing an array of arguments, these kinds of changes/core support are less and less needed.

You can call `$gridField->addComponents(...$arrayOfComponents)` instead of us having to support a somewhat outdated approach

Although I won't make a big fuss of it, I'd support reverting this

tractorcow:
Yeah I'm already in trouble for messing with method signatures... it's not strictly semver ;)

robbieaverill:
I'm not opposed to a revert, but this is backwards compatible

tractorcow:
Yes this is... but using variadic signature may not be.

dhensby:
> Yes this is... but using variadic signature may not be.

I'm talking about when the method is called, rather than the method definition

tractorcow:
Oh, sure. Is that ok to do in php 5.6?

dhensby:
> I'm not opposed to a revert, but this is backwards compatible

I accept there's not BC break - but there is an extra feature to maintain here. For devs who want to pass an array to the method they can use `...$array` instead of just `$array` and it reduces our maintenance burden, tests, etc :)

dhensby:
> Oh, sure. Is that ok to do in php 5.6?

Yes, was added in 5.6 src: http://php.net/manual/en/migration56.new-features.php#migration56.new-features.splat

dhensby:
I think, in summary, in 5.x I'd rather be removing the ability to pass an array and deprecating it from 4.x than adding more places where passing arrays is acceptable

PR #7848, opened 2018-02-08
FIX Add Nested DB transaction support

### LINKED ISSUE #7847: No support for nested transactions
## Affected Version

4.0+

## Description

MySQL does not support nested transactions:

> Transactions cannot be nested. This is a consequence of the implicit commit performed for any current transaction when you issue a START TRANSACTION statement or one of its synonyms.

https://dev.mysql.com/doc/refman/5.7/en/implicit-commit.html

This means that if a user starts a transaction in their code and then another module or core code starts a new transaction (eg: publishing `ChangeSet`s) then the first transaction is implicitly committed and no rollback of the original transaction is possible.

I suggest we add some kind of psuedo support for nested transactions where we don't call `START TRANSACTION` on the DB if we have already started one and instead add a save point which can be rolled back to.

## Steps to Reproduce

Write to the DB inside a transaction, publish a `ChangeSet` and then try to roll back your original changes

dhensby:
Fixes #7847

### DIFF
--- a/src/ORM/Connect/MySQLDatabase.php
+++ b/src/ORM/Connect/MySQLDatabase.php
@@ -57,6 +57,11 @@ class MySQLDatabase extends Database
      */
     private static $collation = 'utf8_general_ci';
 
+    /**
+     * @var bool
+     */
+    protected $transactionNesting = 0;
+
     public function connect($parameters)
     {
         // Ensure that driver is available (required by PDO)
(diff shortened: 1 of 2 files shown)


dhensby:
Looks like Postgres needs an update to support nested transactions too

dhensby:
Postgres fix at https://github.com/silverstripe/silverstripe-postgresql/pull/83

tractorcow:
Postgres fix merged.

tractorcow:
Oh wait, postgres was merged into 2 not 2.0.

tractorcow:
I've back ported the fix to the 2.0 branch and tagged 2.0.2

dhensby:
green 💚

PR #7850, opened 2018-02-08
FIX Singleton creation

jonom:
Singleton objects created by the Injector in 3.x always have the `populateDefaults()` method fired, despite a check in the DataObject constructor intended to prevent this.

The result is that populateDefaults() runs a lot of times that it shouldn't, including during the building of tables, which can easily break a build if you're trying to pull data from a table that isn't ready yet.

Additionally, infinite loops can occur if you try to access another object of the same class from within the populateDefaults() method. Example:

```php
public function populateDefaults()
{
    parent::populateDefaults();
    $first = self::get()->first();
    // Script will timeout
}
```

If you paste that code example in to the Page class, the CMS won't load at all.

Not sure if `in_array('DataObject', ClassInfo::ancestry($class))` or `is_subclass_of($class, 'DataObject')` would perform better, or if it matters?

### DIFF
--- a/control/injector/Injector.php
+++ b/control/injector/Injector.php
@@ -551,6 +551,11 @@ class Injector {
 			$constructorParams = $spec['constructor'];
 		}
 
+		// If we're dealing with a DataObject, pass through Singleton flag as second argument
+		if ($type != 'prototype' && empty($constructorParams) && is_subclass_of($class, 'DataObject')) {
+			$constructorParams = array(null, true);
+		}
+
 		$factory = isset($spec['factory']) ? $this->get($spec['factory']) : $this->getObjectCreator();
 		$object = $factory->create($class, $constructorParams);
 
--- a/tests/model/DataObjectTest.php
+++ b/tests/model/DataObjectTest.php
@@ -33,6 +33,51 @@ class DataObjectTest extends SapphireTest {
 		'ManyManyListTest_Category',
 	);
 
+	/**
+	 * @dataProvider provideSingletons
+	 */
+	public function testSingleton($inst, $defaultValue, $altDefaultValue)
+	{
+		$inst = $inst();
+		// Test that populateDefaults() isn't called on singletons
+		// which can lead to SQL errors during build, and endless loops
+		if ($defaultValue) {
+			$this->assertEquals($defaultValue, $inst->MyFieldWithDefault);
+		} else {
+			$this->assertEmpty($inst->MyFieldWithDefault);
+		}
+
+		if ($altDefaultValue) {
+			$this->assertEquals($altDefaultValue, $inst->MyFieldWithAltDefault);
+		} else {
+			$this->assertEmpty($inst->MyFieldWithAltDefault);
+		}
+	}
+
+	public function provideSingletons()
+	{
+		// because PHPUnit evalutes test providers *before* setUp methods
+		// any extensions added in the setUp methods won't be available
+		// we must return closures to generate the arguments at run time
+		return array(
+			array(function () {
+				return DataObjectTest_Fixture::create();
+			}, 'Default Value', 'Default Value'),
+			array(function () {
+				return new DataObjectTest_Fixture();
+			}, 'Default Value', 'Default Value'),
+			array(function () {
+				return singleton('DataObjectTest_Fixture');
+			}, null, null),
+			array(function () {
+				return DataObjectTest_Fixture::singleton();
+			}, null, null),
+			array(function () {
+				return new DataObjectTest_Fixture(null, true);
+			}, null, null),
+		);
+	}
+
 	public function testDb() {
 		$obj = new DataObjectTest_TeamComment();
 		$dbFields = $obj->db();


tests/model/DataObjectTest.php
```diff
@@ -33,6 +33,25 @@ class DataObjectTest extends SapphireTest {
 		'ManyManyListTest_Category',
 	);
 
+	public function testSingleton() {
+		// Test that populateDefaults() isn't called on singletons, which can lead to SQL errors during build, and endless loops
+		$inst = DataObjectTest_Fixture::create();
```
dhensby: might be better to use a dataProvider here

jonom: what would be the advantage?

tractorcow: If one test fails it won't block all the other tests.

control/injector/Injector.php
```diff
@@ -551,6 +551,14 @@ protected function instantiate($spec, $id=null, $type = null) {
 			$constructorParams = $spec['constructor'];
 		}
 
+		// If we're dealing with a DataObject, pass through Singleton flag as second argument
+		if ($type != 'prototype' && in_array('DataObject', ClassInfo::ancestry($class))) {
```
tractorcow: A lot of binding here to DataObject. How about looking into a DataObjectFactory configured via yaml?

jonom:
Switched to `is_subclass_of` to fix Injector test fail

tests/model/DataObjectTest.php
jonom: Do the other tests in this class use dataProviders? Or can you provide an example? I'm not really sure how to proceed.

control/injector/Injector.php
jonom: I don't really know what that means. I really want to see this ancient issue fixed but I don't have a lot of time to look further in to it or solve it more elegantly unfortunately. I had thought this would be only for 3.x too but actually glancing at Injector and DataObject in SS4 this is probably still an issue and would need to be merged up.

tests/model/DataObjectTest.php
dhensby: @jonom see https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers

I've made the change for you and pushed it up

jonom: > I've made the change for you and pushed it up

Thanks @dhensby! I didn't even realise you could do that. Were you able to push straight to my branch with your regular credentials or is it more complicated? I'll have to remember that option for the future.

Tests failed after your commit was introduced but [passed](https://travis-ci.org/silverstripe/silverstripe-framework/builds/339298668) before. I can't see anything wrong with your commit though so I'm quite confused.

jonom: From the docs you linked:

> All data providers are executed before both the call to the setUpBeforeClass static method and the first call to the setUp method. Because of that you can't access any variables you create there from within a data provider. This is required in order for PHPUnit to be able to compute the total number of tests.

I guess SapphireTest:setUp() is running after the objects are instantiated?

dhensby:
well, that's fun, using a dataprovider seems to make the test fail :/

jonom:
> well, that's fun, using a dataprovider seems to make the test fail :/

@dhensby not sure if you would have seen my comment since it's on an outdate diff:

From the docs you linked:

> All data providers are executed before both the call to the setUpBeforeClass static method and the first call to the setUp method. Because of that you can't access any variables you create there from within a data provider. This is required in order for PHPUnit to be able to compute the total number of tests.

I guess SapphireTest:setUp() is running after the objects are instantiated now and something critical happens in there?

control/injector/Injector.php
tractorcow: Kk well, fair enough. I'd probably not want this leaking up into 4.x branch though. 4 should be less concerned about singleton-specific construction.

tractorcow:
Just make the first argument a closure that creates the singleton. :)

control/injector/Injector.php
```diff
@@ -551,6 +551,14 @@ protected function instantiate($spec, $id=null, $type = null) {
 			$constructorParams = $spec['constructor'];
 		}
 
+		// If we're dealing with a DataObject, pass through Singleton flag as second argument
+		if ($type != 'prototype' && is_subclass_of($class, 'DataObject')) {
+			if (!count($constructorParams)) {
+				$constructorParams[0] = null;
+			}
+			$constructorParams[1] = true;
```
tractorcow: Sometimes constructor args are named.

```yaml
MyObject:
  class: MyObject
  constructor:
    record: []
    singleton: true
```

however that's super edge case. Maybe only apply this fix if $constructorParams are totally empty?

Not a review change request, because it feels edge case enough we can ignore it, but worth thinking about before we modify this in 4.x.

dhensby:
Also, @jonom does this affect 3.5 too? or just 3.6+?

dhensby:
@jonom 

>I didn't even realise you could do that. Were you able to push straight to my branch with your regular credentials or is it more complicated? I'll have to remember that option for the future.

It was a feature added a while ago: https://github.com/blog/2247-improving-collaboration-with-forks

By default users that open PRs allow repo owners / contributors the ability to push into that branch until permission is revoked or the PR is closed/merged.

dhensby:
@tractorcow we are green - thanks for the closure suggestion - can you decide if you're happy to merge, please?

control/injector/Injector.php
jonom: Could we just convert to numeric keys:
```php
$constructorParams = array_values($constructorParams);
```
The constructor args will be passed by position rather than name right? So I assume it should be safe to discard any non-numeric keys.

jonom:
> Also, @jonom does this affect 3.5 too? or just 3.6+?

I'm guessing this bug was introduced in 3.0 or whenever the Injector first became a thing.

control/injector/Injector.php
tractorcow: It's safer to only supply default constructor params, in this case based on `$type != 'prototype' && empty($constructorParams)`. Trying to merge in based on position / key is guesswork at best.

jonom: I don't think we can add a `empty($constructorParams)` check to this because `$constructorParams = [null]` would not be considered empty, would it?

tractorcow: Yes but when would we have `[null]`?

jonom: Don't know :) I'm just guessing so I'll trust you on this lol. Have added a commit with a `empty($constructorParams)` condition.

tractorcow:
Merge on green.

jonom:
I have a squashed version of this ready for 3.5 if you want me to push that up and change the base?

tractorcow:
We should merge into 3.5, better.

jonom:
Thanks for your help guys. @dhensby sorry I couldn't work out how to move this to 3.5 and keep a credit on your commit but I figured it was likely to be squashed in to mine anyway.

jonom:
Scrutinizer got a bit confused

dhensby:
OK - so I've merged this up to 3 but it's still an issue in 4 and @tractorcow has asked I don't merge the "fix" up into 4 as we want a less tightly coupled approach in `Injector`.

@tractorcow I looked into creating a `DataObject` factory, but the problem is that a factory has to be registered against every class (not just the common parent class) which means there's no consistent or easy way to create a factory for all `DataObject`s (that I can see). We'd also need to add a third param to the `Injector\Factory@create` interface to pass along the `$type` var (`SINGLETON` vs `PROTOTYPE`) so our `Factory` knows how it should be creating the objects.

So bad news is this PR isn't going to get merged up into 4 and all and we'll need a new PR for 4 if we want this fix to make its way up :(

tractorcow:
I really want to get away from singletons needing special arguments to work. I'm highly reluctant to keep supporting this feature going forward.

jonom:
> I really want to get away from singletons needing special arguments to work. I'm highly reluctant to keep supporting this feature going forward.

It seems the only use for the `$isSingleton` arg is to skip firing the `populateDefaults()` method. That seems pretty important for performance though, and also prevents build errors. If you want to take that argument out I guess you would have to not call `populateDefaults()` in the constructor. So what then? Do we make people call it manually? If we force people to use the Injector it's not an issue as we could call populateDefaults from there on new prototype DataObjects. But if you want to support the `new MyDataObject` syntax I can't see how we could handle this case without informing the constructor of whether or not this is a singleton.

Is dropping `new` support an option? Using the Injector is heavily encouraged, could it be mandated?

tractorcow:
> and also prevents build errors.

Only in cases where you have db-dependant populateDefaults() implementations.

We could just guard those with `catch (DatabaseException)`

tractorcow:
Sure it's an empty catch, but it's a specific guard against a known use case, rather than a global Exception suppressor.

```php
public function populateDefaults()
{
    try {
        $first = self::get()->first();
        if ($first) {
            $this->owner->Title = $first->Title . ' new';
            }
        } catch (DatabaseException $ex) {
        	// guard against errors during db bootstrapping
        }
}
```

jonom:
So that `catch` would allow the code to keep running would it? It sounds like a clever solution but it would be nice if devs didn't have to guard against this themselves. I guess if we update [the docs on this topic](https://docs.silverstripe.org/en/4/developer_guides/model/how_tos/dynamic_default_fields/) with instructions though that could do the trick.

I'm not actually sure what the reasoning is for DataObject Singletons to not have their defaults populated. Besides preventing build errors, I had assumed perhaps performance but if the whole point of a singleton is that only one for each class exists, executing a little extra code for each one shouldn't be a big concern, and this is what's been happening as long as this bug has existed anyway. I tried to look up the reasoning for the `$isSingleton` check in the DataObject constructor but it is prehistoric... was introduced in the [first commit](https://github.com/silverstripe/silverstripe-framework/commit/4a5d9b03f82d383cef40602aed7353dea17c31b4) in this repo.

Maybe the fix we implemented for 3 should be merged up to 4, but in master we could remove the $isSingleton arg from the DataObject constructor and let singletons have their defaults populated?

tractorcow:
Pretty much as you said it... it was more a performance / error prevention step. The solution in my mind is to simply avoid heavy lifting when populating defaults, as well as allowing it to silently ignore DB errors.

PR #7861, opened 2018-02-12
ENHANCEMENT composer.json missing notice

zanderwar:
This goes towards helping a dev resolve a "resource could not be found" error message when the lookup is of `vendor/package:resource` format but the composer.json file (or package directory) is missing.

### DIFF
--- a/src/Core/Manifest/ModuleResourceLoader.php
+++ b/src/Core/Manifest/ModuleResourceLoader.php
@@ -103,7 +103,7 @@ class ModuleResourceLoader implements TemplateGlobalProvider
         $resource = $matches['resource'];
         $moduleObj = ModuleLoader::getModule($module);
         if (!$moduleObj) {
-            throw new InvalidArgumentException("Can't find module '$module'");
+            throw new InvalidArgumentException("Can't find module '$module', the composer.json file may be missing from the modules installation directory");
         }
         $resourceObj = $moduleObj->getResource($resource);
         if (!$resourceObj->exists()) {


src/Core/Manifest/ModuleResourceLoader.php
```diff
@@ -103,12 +105,47 @@ public function resolveResource($resource)
         $resource = $matches['resource'];
         $moduleObj = ModuleLoader::getModule($module);
         if (!$moduleObj) {
-            throw new InvalidArgumentException("Can't find module '$module'");
+            $append = '';
+            if (!$this->hasModuleComposerConfig($module)) {
+                $append = ", the composer.json file appears to be missing.";
+            }
+            throw new InvalidArgumentException("Can't find module '$module'$append");
```
flamerohr: we could just extend the message with "make sure the composer.json file exists" without the extra check/guessing, what do you think?

zanderwar: Yeah, agree, lol.

robbieaverill: +1 but it's not necessarily _because_ the file doesn't exist - let's make sure we use wording that implies it might be missing and it's worth checking

zanderwar: Done

src/Core/Manifest/ModuleResourceLoader.php
```diff
@@ -3,6 +3,8 @@
 namespace SilverStripe\Core\Manifest;
 
 use InvalidArgumentException;
+use SilverStripe\Control\Controller;
+use SilverStripe\Control\Director;
```
flamerohr: imports no longer needed ;)

zanderwar: Ah, cheers, my bad - shouldn't code blind - resolved

PR #7864, opened 2018-02-13
BUG Fix incorrect convert slashes argument

tractorcow:
I am bad for not noticing a missing unit test earlier.


tests/php/Core/ConvertTest.php
```diff
@@ -613,4 +613,39 @@ public function bytes2MemStringProvider()
             [(512 * 1024 * 1024 * 1024 * 1024 * 1024), '512P']
         ];
     }
+
+    public function providerTestSlashes() {
```
robbieaverill: phpcs violation

PR #7871, opened 2018-02-15
FIX: Always write session cookie in Session::inst_start() (fixes #7870)

### LINKED ISSUE #7870: Unable to write anything to session after Session::destroy() is called
## Affected Version

3.6.5 (haven’t tested 4.x)

## Description

Calls to `Session::inst_destroy()` make it impossible to write any _new_ cookie data in the same request.

When `Session::inst_destroy()` is called, it calls `Cookie::force_expiry(session_name(), ...`. This means that any session data written after that point will be lost, because the cookie is forcibly removed.

Calling `Session::inst_start()` _after_ destroying the session doesn’t have any impact with the default configuration. `inst_start()` _can_ set a new cookie, but doesn’t if `Session.timeout` is `0`.

## Steps to Reproduce

Easiest way to trigger session destruction is by changing user agent:

1. Add `{$SecurityID}` to your template (or anything that’s enough to trigger a value to be written to the session)
2. Open Chrome web inspector in incognito mode
3. Visit the page, observe that a session cookie is set `Set-Cookie: PHPSESSID=12345`
4. Change user agent (device dropdown menu is enough in Chrome inspector)
5. Refresh the page. Session cookie is removed `Set-Cookie: PHPSESSID=deleted`. This is despite the fact that the session cookie is needed for security ID to work

kinglozzer:


kinglozzer:
Annoyingly I don’t think this can be unit tested, at least not without hacking in `$isUnitTest` checks into `Session` because headers are sent before the tests even start running 😕

dhensby:
@kinglozzer do we know how other libraries deal with this? Does symfony/http-foundation do something similar? I'm just worried that, as we don't have any abstraction layer for sessions, that this tightly couples to cookie based sessions when that might not be how sessions are persisted on all sites...

tractorcow:
So does timeout 0 mean "never times out" or "immediately times out"?

dhensby:
It means last as long as the session...

kinglozzer:
> So does timeout 0 mean "never times out" or "immediately times out"?

As Dan said, it’s equivalent to the default behaviour where it lasts until you close the browser

> @kinglozzer do we know how other libraries deal with this? Does symfony/http-foundation do something similar? I'm just worried that, as we don't have any abstraction layer for sessions, that this tightly couples to cookie based sessions when that might not be how sessions are persisted on all sites...

Yeah, I see what you mean. After looking at this again, I think the main problem here is how we “destroy” the session and forcibly set a cookie when we do so. Gleaning the http-foundation source, they don’t do that (in fact, they leave setting/removing cookies entirely to the PHP session handler):

- https://github.com/symfony/http-foundation/blob/master/Session/Session.php
- https://github.com/symfony/http-foundation/blob/master/Session/Storage/NativeSessionStorage.php

Back to the issue at hand, even if you set a non-cookie-based session handler the expiry cookie will still be set unless you remember to pass `false` for the `$removeCookie` argument (even then, core implementors of this method don’t pass that arg anyway). However `Session::destroy()` is only called on log out (or if UA has been altered between requests), so admittedly that is much rarer than forcing a cookie to be written on session start...

I could add an `&& ini_get('session.use_cookies')` to the if statement wrapping the `Cookie::set()` call?

dhensby:
> I think the main problem here is how we “destroy” the session and forcibly set a cookie when we do so. Gleaning the http-foundation source, they don’t do that (in fact, they leave setting/removing cookies entirely to the PHP session handler):

Does not forcibly setting the cookie solve the problem? Could we do that instead?

dhensby:
I don't like the idea of us interfering with session cookies at all :/

tractorcow:
The reasoning around us setting the cookie was probably to do with the desire to control the expiration programmatically. I'd also be happy with removing the code to manually modify that setting if we could control expiration via other means.

kinglozzer:
Okay, so an alternative solution might be to just remove the cookie logic entirely:

- Remove `Cookie::set(session_name()...` - we call `session_set_cookie_params($lifetime, ...`, so that should handle setting the lifetime of the cookie anyway (assuming that handler is in use)
- Remove `Cookie::force_expiry(session_name()...` from `Session::inst_destroy()`

Thoughts?

dhensby:
Agreed. We don't need to worry about the cookies at all. If someone says they have session id X via an old cookie, it's up to PHP (or or session store) to identify if that session is valid or not. 

If we've expired the session then it won't work even if they still have the cookie

kinglozzer:
Blocker - the reason the cookie is set manually is:

> Modify the timeout behaviour so it's the *inactive* time before the session expires. By default it's the total session lifetime

Basically if I set a timeout to 60 seconds, PHP will send a cookie with an expiry that’s 60 seconds in the future. Subsequent requests will not “refresh” the cookie, so it’ll expire 60 seconds after the first request. We altered that behaviour so that each subsequent request will set the cookie again with a new expiry time :/

For reference, other frameworks don’t bother altering PHP’s behaviour here. However, I don’t think we can change that in a patch release...

dhensby:
> For reference, other frameworks don’t bother altering PHP’s behaviour here. However, I don’t think we can change that in a patch release...

why not? ;) If our mangling with the session cookies is breaking things, then we should stop doing it if it doesn't break things

kinglozzer:
> why not? ;) If our mangling with the session cookies is breaking things, then we should stop doing it if it doesn't break things

If I’ve set my site’s session expiry to 5 minutes, sessions will now expire 5 minutes after the first request regardless of whether I’m still actively visiting the site. Pretty big behaviour change :/

dhensby:
Setting session lifetimes should be done via PHP's session settings - we shouldn't be touching the cookies at all :(

I'd say get rid of all of our cookie interference so that it works as PHP sessions handler intended

dhensby:
There is a big distinction between a session actually having a 5 min lifetime and a cookie expiring after 5 mins - the session will still exist even if the cookie doesn't

dhensby:
> It may still exist, but you’ll lose access to it so I still think it’s a problem

My point is that we currently rely on a cookie with a lifetime - well, that's bad because it doesn't actually cause the session to end once the cookie does; if we want to set session lifetimes then we should do that properly rather than via our own cookie expiration policy

tractorcow:
Someone want to create an RFC for removing session cookie handling? Maybe ss 5.0?

dhensby:
Right, I don't want this to languish with the bug not being fixed because we debate the best way to get this done. If we can remove our cookie handling and that fixes the problem, I'd advocate for that over this patch - otherwise we should get this merged

kinglozzer:
Okay so to summarise this PR/discussion on fixing in 3.x:

| Solution | Problems |
| -------- | ---------- |
| Current PR state - _always_ set session cookie | Duplicate `Set-Cookie` headers - no real impact, just an annoyance. Also means that non-cookie session handlers may get a cookie set where they previously wouldn’t |
| Remove cookie logic from session - let PHP handle it natively | Should be no practical implications, _unless_ a non-default session lifetime is set where there may be BC implications. See [this comment](https://github.com/silverstripe/silverstripe-framework/pull/7871#issuecomment-368853030) for explanation. |

I’ll try to find time to write it up properly into an RFC soon to address properly in 5.x, but essentially my desire would be to make our session helper class a wrapper around `symfony/http-foundation` 😉

dhensby:
@kinglozzer I'd love to get our sessions / request-response loop working with http-foundation but the problem is our nested controller resolution and URL matching is at odds with symfony and it's a bit of a road block - I had a look at it a while ago (https://github.com/silverstripe/silverstripe-framework/compare/4.0...dhensby:poc/4/symfony-http-foundation).

The alternative is to use the CMF dynamic router (which is what we'd need) - http://symfony.com/doc/master/cmf/bundles/routing/dynamic.html but it's a bit old.

I've also had murmorings of a pushback on non PSR-7 HTTPRequest/Response packages...

dhensby:
Personally, I'm not concerned about that BC issue you mention as it's really bad to do it this way as the session stays alive even though your cookie could be gone, which increases theoretical risk of session hijacking..

kinglozzer:
Closing this out because this PR targets 3.x and this isn’t a security issue. Still needs fixing in 4 so I’ll leave the original issue open

PR #7872, opened 2018-02-15
fall back to SSViewer::get_themes when using themeResourceLoaders

andrewandante:
When using a themeResourceLoader, you have to pass in a list of themes, recommending that you pass in `SSViewer::get_themes()` - we should default to that instead, as the assumption when using the method is that it will look through your themes for you. You've already declared them once!


dhensby:
perhaps this needed a test?

PR #7874, opened 2018-02-16
FIX Split HTML manipulation to onadd, so elements are not accidentally duplicated

flamerohr:
Fixes upload dialog problem intermittently duplicating the treedropdown html

grafik

NOTE: not the root cause of the issue - but I couldn't replicate the issue.

### DIFF
--- a/javascript/TreeDropdownField.js
+++ b/javascript/TreeDropdownField.js
@@ -47,13 +47,15 @@
 			// XMLHttpRequest
 			CurrentXhr: null,
 
-			onmatch: function() {
+			onadd: function() {
 				this.append(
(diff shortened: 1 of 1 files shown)


kinglozzer:
Some extra context on the issue: https://github.com/silverstripe/silverstripe-framework/pull/7345 seems to have been what introduced it. Clicking “From files” (or “Choose another file”) on an UploadField will _sometimes_ cause `.TreeDropdownField`’s `onmatch` handler to trigger twice. It’s super intermittent (1 in every ~15 attempts causes it for me), and I got absolutely nowhere trying to track down the cause.

PR #7875, opened 2018-02-17
Suggested feature: dev/build record count

sminnee:
This small piece of metadata is intended to expose record counts to
developers as an information radiator: from time to time, SilverStripe
tables can get very large, and this exposes this information without
the developer seeking it out. 

It’s reasonable to expect that count(*) calls aren’t too time consuming,
even on large tables.

On a small test run, dev/build execution went from 3.85s to 3.98s (a 3%
or 130ms increase). Given the small relative impact it should be okay.
Where this is inappropriate, it can be disabled with a config setting.

### DIFF
--- a/src/ORM/DatabaseAdmin.php
+++ b/src/ORM/DatabaseAdmin.php
@@ -48,6 +48,11 @@ class DatabaseAdmin extends Controller
         'RememberLoginHash' => 'SilverStripe\\Security\\RememberLoginHash',
     ];
 
+    /**
+     * Config setting to enabled/disable the display of record counts on the dev/build output
+     */
+    private static $show_record_counts = true;
+
     protected function init()
     {
         parent::init();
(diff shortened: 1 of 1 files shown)


src/ORM/DatabaseAdmin.php
```diff
@@ -292,10 +299,21 @@ public function doBuild($quiet = false, $populate = true, $testMode = false)
...
+                        } catch (Exception $e) {
+                            $countSuffix = " (error getting record count)";
+                        }
+                    } else {
+                        $countSuffix = "";
+                    }
+
                     if (Director::is_cli()) {
-                        echo " * $tableName\n";
+                        echo " * $tableName$countSuffix\n";
```
Firesphere: Wouldn't it make more sense to have the space here (and below), instead of in the variable?

dhensby: TBH, I'm happy with this - if you want to PR after it's merged to change it, that's fine, but I don't have a problem with it as is

PR #7879, opened 2018-02-20
BUG Fix extend() failing on protected extend-prefixed methods

tractorcow:
Also see https://github.com/silverstripe/silverstripe-cms/pull/2103

cc @dhensby for master bugfixing


mikenz:
Passes in your unit tests and our additonal ones in our CI system when built with https://github.com/silverstripe/silverstripe-cms/pull/2103
Thanks

PR #7883, opened 2018-02-22
php-shell-cmd script from pear/php_shell package

tony13tv:
[FIX] With this change we are allowed to use php-shell-cmd file from pear/php_shell package as that script now is located on scripts folder inside package.

### DIFF
--- a/dev/SapphireREPL.php
+++ b/dev/SapphireREPL.php
@@ -43,7 +43,7 @@ class SapphireREPL extends Controller {
 
 
 		/* Try using PHP_Shell if it exists */
-		@include 'php-shell-cmd.php' ;
+		@include 'scripts/php-shell-cmd.php' ;
 
 		/* Fall back to our simpler interface */
 		if( empty( $__shell ) ) {


tony13tv:
To test this you need to install pear/php_shell using composer but it's not registered on packagist so:
```
$ composer config repositories.php_shell git https://github.com/pear/PHP_Shell
$ composer require pear/php_shell
```
Using php 7.0.25 I had to install pear/pear_shell requirements (```$ composer require pear/pear pear/pear_exception```) before require php_shell as it would not install because of wildcard version for requirements.

dhensby:
So what chances does this have of causing a regression? I can see from that repo that the `php-shell-cmd.php` file has lived in `scripts` for a very long time.

What version of `php_shell` had it in the root (or, at least, at the root of an include path)?

@chillu @tractorcow - thoughts?

tony13tv:
I think ```php-shell-cmd.php``` lived in root pre-git (it was a svn repository), but, checking the [first commit with files](https://github.com/pear/PHP_Shell/tree/3f0968160eea57c86b18d1b4ddf58692d15af12b) on the repo found it in scripts.

kinglozzer:
Thanks for the pull request, but I’m afraid we’re no longer accepting non-critical/non-security fixes to 3.x.

PR #7884, opened 2018-02-22
Update 4.0.0.md

JorisDebonnet:
Fixed several small issues


docs/en/04_Changelogs/4.0.0.md
```diff
@@ -1325,7 +1325,7 @@ private static $owns = [
 ];
 ```
 
-This is done to ensure that the published state of linked components are consistent with their "owner."
+This is done to ensure that the published state of linked components are consistent with their "owner".
```
robbieaverill: Minor, but I think the original is grammatically correct

PR #7887, opened 2018-02-23
Allow PHPParser v4

dhensby:
Tested locally and still works a treat :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "embed/embed": "^3.0",
         "league/flysystem": "~1.0.12",
         "monolog/monolog": "~1.11",
-        "nikic/php-parser": "^2 || ^3",
+        "nikic/php-parser": "^2 || ^3 || ^4",
         "paragonie/random_compat": "^2.0",
         "silverstripe/config": "^2@dev",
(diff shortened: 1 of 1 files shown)


kinglozzer:
I can’t spot anything breaking in the [changelog](https://github.com/nikic/PHP-Parser/blob/master/CHANGELOG.md), but should we wait for a stable release before merging just in case anything else changes?

dhensby:
@kinglozzer it's been released as beta and they've said the next release will be stable - I don't expect any changes so radical from the beta that would cause our fairly basic use of the parser to break...

Though I suppose it doesn't hurt to wait

tractorcow:
How about disallowing the older ^2 while we're at it?

dhensby:
OK, 4.0.0 is stable now and I've restarted the test suites so they'll pull in 4.x

> How about disallowing the older ^2 while we're at it?

Personally, my view is that if it works, we should keep it. As a library we should be as loose as possible with constraints and leave it to the projects to be as strict as possible so that we aren't the cause of unresolvable dependency definitions. If we do work with v2 then we shouldn't be blocking it

PR #7893, opened 2018-02-27
Backport 4.0 implementation of numeric/assoc key merging for summary…

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-cms/issues/2006

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -3711,7 +3711,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	public function summaryFields() {
         $rawFields = $this->stat('summary_fields');
 
-        $fields = [];
+        $fields = array();
         // Merge associative / numeric keys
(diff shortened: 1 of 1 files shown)


model/DataObject.php
```diff
@@ -3709,13 +3709,18 @@ public function fieldLabel($name) {
...
-		$fields = $this->stat('summary_fields');
-
-		// if fields were passed in numeric array,
-		// convert to an associative array
-		if($fields && array_key_exists(0, $fields)) {
-			$fields = array_combine(array_values($fields), array_values($fields));
-		}
+		$rawFields = $this->stat('summary_fields');
+
+		$fields = [];
```
dhensby: We need to support 5.3 so no short array syntax please

dhensby:
The tabbing was still off (you're using spaces not tabs ;) ) but I merged this by hand and fixed it up :)

Issue #7888, opened 2018-02-25
Support get_by_id without $callerClass


JorisDebonnet:
I always get a bit annoyed when I have to write `MyClass::get_by_id(MyClass::class, $id)` because it seems like such a redundancy to write that second MyClass. 

I know how I can make it support `MyClass::get_by_id($id)` while maintaining full backwards compatibility, but before doing that and submitting a PR I thought I'd first hear whether people would be against it.

dhensby:
I think the general view is that we'd like to do away with `DataObject::get_by_id()` and replace it with `MyObject::get()->byID()` - the reason `get_by_id` has hung about is because it's got a cache backing it and no one has got round to caches to back `DataList`.

So... taking into account nothing has happened to properly deprecate `get_by_id`, I'd definitely consider a PR for adding the `MyObject::get_by_id()` syntax if it was BC safe

JorisDebonnet:
Being handled and discussed in PR #7894.

PR #7894, opened 2018-02-27
get_by_id: alternate signature to allow MyDataObject::get_by_id($id)

JorisDebonnet:
Implemented for #7888

Backwards compatible, although the API changes very slightly in the sense that `get_by_id` can be called with just one parameter, where two parameters were required in the past. A user warning is generated if this is attempted directly on the `DataObject` class.

### DIFF
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -3175,22 +3175,24 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
     /**
      * Return the given element, searching by ID
      *
-     * @param string $callerClass The class of the object to be returned (optional if called from extending class)
-     * @param int $id The id of the element
-     * @param boolean $cache See {@link get_one()}
+     * @param string|int $callerClassOrID The class of the object to be returned (optional if called from extending class)
+     * @param int|bool $id The id of the element or bool to represent cache if ID is passed as fist argument
+     * @param bool $cache See {@link get_one()}
      *
-     * @return DataObject The element
+     * @return static The element
      */
-    public static function get_by_id($callerClass, $id = null, $cache = true)
+    public static function get_by_id($callerClassOrID, $id = null, $cache = true)
     {
-        if(is_numeric($callerClass)) {
+        if (is_numeric($callerClassOrID)) {
             // Alternate signature to allow calling as MyDataObject::get_by_id($id)
-            $cache = is_null($id) ? true : !!$id;
-            $id = $callerClass;
+            $cache = is_null($id) ? true : (bool)$id;
+            $id = $callerClassOrID;
             $callerClass = get_called_class();
-            if (__CLASS__ === $callerClass) {
+            if (__CLASS__ === $callerClassOrID) {
                 user_error("DataObject::get_by_id called without providing object class", E_USER_WARNING);
             }
+        } else {
+            $callerClass = $callerClassOrID;
         }
 
         if (!is_numeric($id)) {
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -3039,29 +3039,33 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
         $limit = null,
         $containerClass = DataList::class
     ) {
-
+        // Validate arguments
         if ($callerClass == null) {
             $callerClass = get_called_class();
-            if ($callerClass == self::class) {
-                throw new \InvalidArgumentException('Call <classname>::get() instead of DataObject::get()');
+            if ($callerClass === self::class) {
+                throw new InvalidArgumentException('Call <classname>::get() instead of DataObject::get()');
             }
-
-            if ($filter || $sort || $join || $limit || ($containerClass != DataList::class)) {
-                throw new \InvalidArgumentException('If calling <classname>::get() then you shouldn\'t pass any other'
+            if ($filter || $sort || $join || $limit || ($containerClass !== DataList::class)) {
+                throw new InvalidArgumentException('If calling <classname>::get() then you shouldn\'t pass any other'
                     . ' arguments');
             }
-
-            return DataList::create(get_called_class());
+        } elseif ($callerClass === self::class) {
+            throw new InvalidArgumentException('DataObject::get() cannot query non-subclass DataObject directly');
         }
-
         if ($join) {
-            throw new \InvalidArgumentException(
+            throw new InvalidArgumentException(
                 'The $join argument has been removed. Use leftJoin($table, $joinClause) instead.'
             );
         }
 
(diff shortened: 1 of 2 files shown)


src/ORM/DataObject.php
```diff
@@ -3175,14 +3175,24 @@ public static function reset()
     /**
      * Return the given element, searching by ID
      *
-     * @param string $callerClass The class of the object to be returned
+     * @param string $callerClass The class of the object to be returned (optional if called from extending class)
```
tractorcow: Rename to `$callerClassOrID` to reflect the duality of the argument.

src/ORM/DataObject.php
```diff
@@ -3175,14 +3175,24 @@ public static function reset()
...
      * @param int $id The id of the element
      * @param boolean $cache See {@link get_one()}
```
tractorcow: Clarify if calling with `MyClass::get_by_id($id)`. Does this become second arg, or do you pass null for second arg and this stays third?

src/ORM/DataObject.php
```diff
@@ -3175,14 +3175,24 @@ public static function reset()
     /**
      * Return the given element, searching by ID
      *
-     * @param string $callerClass The class of the object to be returned
+     * @param string $callerClass The class of the object to be returned (optional if called from extending class)
      * @param int $id The id of the element
```
tractorcow: `, unless passed in the first argument`

src/ORM/DataObject.php
```diff
@@ -3175,14 +3175,24 @@ public static function reset()
...
      * @return DataObject The element
      */
-    public static function get_by_id($callerClass, $id, $cache = true)
+    public static function get_by_id($callerClass, $id = null, $cache = true)
     {
+        if(is_numeric($callerClass)) {
+            // Alternate signature to allow calling as MyDataObject::get_by_id($id)
+            $cache = is_null($id) ? true : !!$id;
+            $id = $callerClass;
+            $callerClass = get_called_class();
```
tractorcow: Best practice is create new variables instead of re-assigning method argument variables.

src/ORM/DataObject.php
```diff
@@ -3175,14 +3175,24 @@ public static function reset()
...
+            if (__CLASS__ === $callerClass) {
+                user_error("DataObject::get_by_id called without providing object class", E_USER_WARNING);
```
tractorcow: `InvalidArgumentException` is preferred.

tractorcow:
Speaking of linting, I came up with https://github.com/silverstripe/silverstripe-framework/issues/7899 as an idea. :)

src/ORM/DataObject.php
robbieaverill: We can't do this without breaking semver - unless you're just talking about the docblock?

dhensby: Renaming params are not semver violations

src/ORM/DataObject.php
```diff
@@ -3175,22 +3175,24 @@ public static function reset()
...
-    public static function get_by_id($callerClass, $id = null, $cache = true)
+    public static function get_by_id($callerClassOrID, $id = null, $cache = true)
     {
-        if(is_numeric($callerClass)) {
+        if (is_numeric($callerClassOrID)) {
             // Alternate signature to allow calling as MyDataObject::get_by_id($id)
-            $cache = is_null($id) ? true : !!$id;
-            $id = $callerClass;
+            $id = $callerClassOrID;
+            $cache = is_null($id) ? true : (bool)$id;
```
JorisDebonnet: @dhensby Isn't this in the wrong order now? You're first assigning a new value to $id, and then using $id to set $cache.

dhensby: good point - fixed

src/ORM/DataObject.php
JorisDebonnet: An Exception changes the signature because we'd have a @throws, but in practice it would still be a backwards compatible change because this position in the code can never be reached in any previously _working_ code (as no working code would ever send a numeric value as the first parameter).

src/ORM/DataObject.php
robbieaverill: True, you're right

src/ORM/DataObject.php
```diff
@@ -3175,14 +3175,26 @@ public static function reset()
...
      */
-    public static function get_by_id($callerClass, $id, $cache = true)
+    public static function get_by_id($callerClassOrID, $id = null, $cache = true)
     {
+        if (is_numeric($callerClassOrID)) {
+            // Alternate signature to allow calling as MyDataObject::get_by_id($id)
+            $cache = is_null($id) ? true : (bool)$id;
+            $id = $callerClassOrID;
+            $callerClass = get_called_class();
+            if (__CLASS__ === $callerClassOrID) {
```
tractorcow: This error is impossible if `is_numeric($callerClassOrID)`. Did you mean `if (__CLASS___ === $callerClass)`?

src/ORM/DataObject.php
tractorcow: We don't declare `@throws` in silverstripe. We could, but don't.

GIven that there's no way this error could be thrown unless you use the NEW form of the method it doesn't feel like a semver violation to me.

Also, it's treated as a warning, but it's a real error. Not sure a warning is sufficient in any case.

src/ORM/DataObject.php
tractorcow: Don't use argument re-assigment on $id. That's one of my original concerns (and demonstrates why this is bad practice). :P

src/ORM/DataObject.php
```diff
@@ -3175,14 +3175,26 @@ public static function reset()
...
+                user_error("DataObject::get_by_id called without providing object class", E_USER_WARNING);
```
tractorcow: You can't treat this as a warning because `DataObject::get_one(DataObject::class)` is a certain failure. A warning is for something that may be wrong, but isn't certainly an error.

src/ORM/DataObject.php
tractorcow: This error can be outside of the arg processing loop anyway.

src/ORM/DataObject.php
dhensby: It's what we do if the id is not numeric and I think we should keep that over adding an exception that is definitely more risky

tractorcow:
@JorisDebonnet see https://github.com/silverstripe/silverstripe-framework/pull/7901 I thought it'd be easier to just demonstrate what I mean. :)

JorisDebonnet:
@tractorcow Looks good, feel free to merge that one and we'll close this :)

tractorcow:
Well I can't merge my own PR. Maybe @dhensby can.

Issue #1903, opened 2013-05-13
[ORM] [2009-06-05] When you delete a DataObject that is has a many_many join, entries in the join table aren't deleted.


silverstripe-issues:
created by: @sminnee (sminnee)
created at: 2009-06-05
original ticket: http://open.silverstripe.org/ticket/4195

---

For example, delete a Group and it leaves orphans in Group_Members.

chillu:
Note that this is only a problem if AUTO_INCREMENT values are reset (at least on MySQL). 

- Create `Group ID#1`, give it author permissions
- Assign `Member ID#99` to the group
- Delete `Group ID#1`
- `Group_Members` still has an entry with `MemberID=99, GroupID=1`
- Create `Group ID#2`, give it admin permissions (with auto_increment)
- `Member ID#99` doesn't have admin permissions or is a member of the new group (different ID)

If that group was created with `ID#1` instead (because AUTO_INCREMENT was reset), we'd have a problem. This can happen when data is exported as SQL dumps without AUTO_INCREMENT, which raises the priority of this issue. It's already marked as `impact/high` (one of only 25 bugs across all core repos with that classification), but I've also added it to the 4.1.0 milestone to reflect the priority. Ideally we can back port this to 3.x - depending on how much extra work that is.

tractorcow:
A problem I've also been mulling over is how many_many mapping tables should relate to versioned records.

Would it be fair to say that, given the unversioned nature of this table, it should be bound to draft as the source of truth? If we were to trigger the removal of these fields (either by foreign keys or a separate SQL query) we could rely on this as the source of truth.

tractorcow:
If you use many_many through with cascade_deletes, then this issue disappears.

The issue remains only when using normal many_many.

How about:
 - By default, records delete their many_many relations (either through onAfterDelete or foreign key constraint)
 - Versioned records disable the above constraint check. In this case we document to use many_many through and cascade_deletes instead

sminnee:
> A problem I've also been mulling over is how many_many mapping tables should relate to versioned records. Would it be fair to say that, given the unversioned nature of this table, it should be bound to draft as the source of truth

Yes, that seems fair.

The reality is that it's not possible to use many_many with versioned data in a way that's safe. The solution for such case is to use many_many_through.

> By default, records delete their many_many relations (either through onAfterDelete or foreign key constraint)

Yes.

[...]

PR #7896, opened 2018-02-27
Delete orphaned Group_Members records after deleting a Member

JorisDebonnet:


dhensby:
I'm sure this was logged before, can you take a quick look for an old ticket or PR covering this?

lerni:
@dhensby related: https://github.com/silverstripe/silverstripe-framework/issues/1903#issuecomment-339520737

dhensby:
@lerni that's the one - thanks.

@tractorcow what do you think to this? is there likely to be movement on #1903 any time soon?

tractorcow:
I don't think #1903 is going to be implemented any time soon. Let's just merge this.

PR #7897, opened 2018-02-27
Clarify how to import JavaScript

worikgh:
The documentation as it stands is not clear enough for a complete novice.
I am not quite a complete novice so I decoded it.  But 12 weeks ago it would have stumped me
My style will not be right, I have not learnt well enough the documentation style, but I think I have made my point

### DIFF
--- a/docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/CMS_Alternating_Button.md
+++ b/docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/CMS_Alternating_Button.md
@@ -136,8 +136,12 @@ extras.
 Continuing our example let's add a "constructive" style to our *Clean-up* button. First you need to be able to add
 custom JS code into the CMS. You can do this by adding a new source file, here
 `mysite/javascript/CMSMain.CustomActionsExtension.js`, and requiring it
-through a YAML configuration value: `LeftAndMain.extra_requirements_javascript`.
-Set it to 'mysite/javascript/CMSMain.CustomActionsExtension.js'.
+through a YAML configuration value: 
+```
+<Name space>\<Class name>:
+  extra_requirements_javascript:
+    - mysite/javascript/<Name of file including including .js suffix
+```
 
 You can now add the styling in response to `afterrefreshalternate` event. Let's use entwine to avoid accidental memory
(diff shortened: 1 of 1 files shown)


docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/How_Tos/CMS_Alternating_Button.md
```diff
@@ -136,8 +136,12 @@ extras.
...
 custom JS code into the CMS. You can do this by adding a new source file, here
 `mysite/javascript/CMSMain.CustomActionsExtension.js`, and requiring it
-through a YAML configuration value: `LeftAndMain.extra_requirements_javascript`.
-Set it to 'mysite/javascript/CMSMain.CustomActionsExtension.js'.
+through a YAML configuration value: 
+```
+<Name space>\<Class name>:
+  extra_requirements_javascript:
+    - mysite/javascript/<Name of file including including .js suffix
+```
```
robbieaverill: I think it might be good to clarify that this is a LeftAndMain configuration option, but (if this is true) that you can also apply it to subclasses of LeftAndMain. I think keeping LeftAndMain in the example code would be good, so you're just adding the namespace to the class

worikgh: I did not know it is a LeftAndMain specific thing.  Makes sense that it is and it makes sense to clarify that  (I am not a complete novice, I have promoted myself to beginner...)
I am applying it to a LeftAndMain descendant (via ModelAdmin) it is working nicely.

flamerohr: Nice job! yes, this is LeftAndMain specific and it does apply to any classes that extends LeftAndMain :) Would be good to keep a mention of LeftAndMain (with the inheritance behaviour)

Perhaps an example for module files would be a good idea as well. Two extra lines like this:
```
# for module source files, you can do the following
    - silverstripe/blog:client/dist/js/main.bundle.js
```

kinglozzer:
Added in https://github.com/silverstripe/silverstripe-framework/commit/554c34eff73c48edab7961f24ae21887613f2b3a. Kept it simple, as this “how to” is an example with a specific file name

PR #7901, opened 2018-02-28
ENHANCEMENT get_by_id: alternate signature to allow MyDataObject::get_by_id($id)

tractorcow:
See #7894

Avoids variable re-assignment, and uses a single args assignment via `list` to demonstrate the difference between the two call forms.

The error has been consolidated and also added to dataobject::get(), which had an error path which wasn't covered.


dhensby:
duplicate of #7894 - I've merged these changes into that PR

PR #7909, opened 2018-03-05
Ensure errors are reported to users from cli-scripts

jakedaleweb:
Surfaces errors from the cli-scripts to the terminal, has the bonus side effect of outputting errors to deployment logs on the SilverStripe Platforms.

With change:
```
# sudo -u www-data framework/sake dev

Parse error: syntax error, unexpected 'global' (T_GLOBAL) in /sites/mediumtest-uat/releases/20180131205002/mysite/_config.php on line 3
```

Without change this fails silently (of course the error is logged).

If we are happy with this approach I can make a similar PR for 4.

### DIFF
--- a/cli-script.php
+++ b/cli-script.php
@@ -18,6 +18,13 @@ if(isset($_SERVER['HTTP_HOST'])) {
 	die();
 }
 
+
+/**
+ * Ensure errors are reported to users via CLI, as this script can only be run via CLI risk of surfacing
+ * errors to end users is minimised.
+ */
+ini_set("display_errors", 1);
+
 /**
  * Identify the cli-script.php file and change to its container directory, so that require_once() works
  */


tractorcow:
This feels like something that should be set at the php-cli.ini level. In some cases you may not want CLI tasks to emit errors (e.g. notices / warnings) if the raw output is consumed by other tasks.

It's not always obvious (or able to be assumed) that a CLI task is being run by a human. :)

Instead, I would recommend that you check your error logging and ensure you send errors there.

tractorcow:
@dhensby what do you think?

jakedaleweb:
@tractorcow 

> This feels like something that should be set at the php-cli.ini level.

The point being that `display_errors` is disabled on these environments, so in the event of a deployment failing at the `framework/sake dev` call no reason is given, as mentioned it is logged but it's a bad user experience.

>  In some cases you may not want CLI tasks to emit errors (e.g. notices / warnings) if the raw output is consumed by other tasks.

Yeah good point, _if_ we go this way we could look at tweaking the error level to ensure only fatal errors are show.

> Instead, I would recommend that you check your error logging and ensure you send errors there.

See above. They are, but it's less than optimal.

--

How would you feel about some sort of flag we can pass through to sake/cli-script to optionally enable this? i.e. verbosity level

tractorcow:
I think another flag would be a great idea. Normally CLI scripts use `-v`, `-vv` or `-vvv` to set varying levels of verbosity.

dhensby:
Are errors sent to the error output or the standard output? if they go to the error output then we don't need to worry about other scripts that rely on parsing output.

I'd lean towards being in favour of this change; but if we could enable it by flag, that could be cool - but note that flags in `sake` / our `cli-script.php` aren't really handled well or consistently... see

jakedaleweb:
@dhensby by default it is sent to the standard output but this can be changed with `ini_set("display_errors", 1);`. I'll look in to a flag for backwards compat at the very least and ping you two again once I have something worthwhile.

Thanks for your time to review (:

jakedaleweb:
Okay @dhensby @tractorcow updated to require a flag and output to stderr, let me know what you think of that approach (:

jakedaleweb:
So the usage is now 
```# framework/sake dev errorstocli
Parse error: syntax error, unexpected 'global' (T_GLOBAL) in /my/web/root/mysite/_config.php on line 3
# framework/sake dev
{no output}```

**Comment by dhensby** [comment]
> `framework/sake dev errorstocli`

So gross... I'll leave it @tractorcow to give some thoughts on that for now

**Comment by tractorcow** [comment]
I think we do need to solve this problem but I'm not 100% sure this is the solution we'd like.

See https://github.com/silverstripe/silverstripe-framework/issues/5542 for my preferred solution; However this involves a complete refactor of the CLI system.

**Comment by Mossman1215** [comment]
necro-ing this thread to provide tips for anyone else using framework and needing sake output to stderr
create a file like  this
`CLI.ini`
```
error_log=""
display_errors="stderr"
```
then when invoking sake add the `CLI.ini` to the beginning of your sake execution
`PHP_INI_SCAN_DIR=:/tmp/CLI.ini ./framework/sake dev/buid`
and errors will go to stderr instead of going to apache error log
Docs for php scan directory are [here](https://www.php.net/manual/en/configuration.file.php#configuration.file.scan)

PR #7910, opened 2018-03-05
API Add getContentCSS() / setContentCSS() to allow per-config customisation of content_css

### LINKED ISSUE #7873: WYSIWYG not loading content_css from _config.php and not able to have multiple configs
SS 4.0.1

In version 3.6.1 we were able to set different configurations for different WYSIWYG (HTMLText Fields) in the admin, for instance if I had a page on the site with a black background and white text, I would have a specific configuration for that page type content WYSIWYG with an editor.css file different than the default one.

In old projects in 3.6.1 I'd have something like this in my _config.php file:

```
HtmlEditorConfig::get(‘cms’)->setOption(‘content_css’,  ‘mysite/css/editor.css’); 

HtmlEditorConfig::get(‘my_other’)->setOption(‘content_css’,  ‘mysite/css/othereditorstyles.css’);
```

In SS 4.0.1  I see that, as per documentation, by default silverstripe tries to load any editor.css file that exists in your current theme or you can set up a different one in your yml file like:

```
---
name: MyCSS
---
SilverStripe\Forms\HTMLEditor\TinyMCEConfig:
  editor_css:
    - 'mysite/css/editor.css'
```

The yml file doesn't give me the option to have separate css files for different HTML Fields.

There are 2 pages on the current documentation on how to edit the html editor field or WYSIWYG:

https://docs.silverstripe.org/en/4/developer_guides/forms/field_types/htmleditorfield/
https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/typography/

So I believe that the the class SilverStripe\Forms\HTMLEditor\TinyMCEConfig is the one I should be using, but I expect that something like the following works:

```
TinyMCEConfig::get('cms')
->addButtonsToLine(1, 'styleselect')
->setOption( 'content_css', 'themes/next/css/customeditor.css');

TinyMCEConfig::get('my_other_config')
->addButtonsToLine(1, 'styleselect')
->setOption('content_css' ,'themes/next/css/my_other_editor.css');
```

And the do something like:

`new HTMLEditorField('OtherContent', 'Other content', $this->OtherContent, 'my_other_config');`

in the object as per section "Specify which configuration to use" in the documentation page: https://docs.silverstripe.org/en/4/developer_guides/forms/field_types/htmleditorfield/

# PRs:

 - [x] https://github.com/silverstripe/silverstripe-framework/pull/7910

tractorcow:
Fixes #7873

### DIFF
--- a/docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/05_Typography.md
+++ b/docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/05_Typography.md
@@ -17,6 +17,13 @@ SilverStripe\Forms\HTMLEditor\TinyMCEConfig:
 
 Will load the `mysite/css/editor.css` file.
 
+Alternatively, you can set this on a specific `TinyMCEConfig` instance via `setContentCSS` method.
+
+```php
+$config = new TinyMCEConfig();
+$config->setContentCSS([ '/mysite/client/css/editor.css' ]);
+```
+
 ## Custom style dropdown
 
 The custom style dropdown can be enabled via the `importcss` plugin bundled with admin module. ([Doc](https://www.tinymce.com/docs/plugins/importcss/))  
--- a/src/Forms/HTMLEditor/HTMLEditorConfig.php
+++ b/src/Forms/HTMLEditor/HTMLEditorConfig.php
@@ -20,7 +20,7 @@ use SilverStripe\Core\Injector\Injectable;
  *
  * Typically global config changes should set the active config.
  *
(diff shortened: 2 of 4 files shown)


diegopego86:
Hey guys, are you planing to add this to 4.1 anytime soon?

andrewandante:
@diegopego86 it will be in the next tagged release as it has been merged into the 4 branch. Until then you can either use `4.x-dev` or the specific commit sha in your composer dependencies to bring it in.

dhensby:
using `^4.2` as your dependency means you'll get the patch now but be linked to a stable release once it makes it into the wild

Issue #7866, opened 2018-02-13
Following lost password link a second time throws exception


robbieaverill:
## Affected Version

4.x-dev (probably 4.0.x too)

## Description

When I follow the link to reset my password from a "forgot my password" email, reset it, log out and follow the link in my email again I get this exception:

```
[Emergency] Uncaught BadMethodCallException: Object->__call(): the method 'forTemplate' does not exist on 'SilverStripe\View\ArrayData'
GET /Security/changepassword?m=2&t=76e780df6db70e2d7f5416e85af0973b7d1256e88ff374afbe8b4853706e04262707451bea157bb92c5f74c8671e5d929f1fb49331ef9b2706246aa24fdca043
```

Full stack trace: https://gist.github.com/robbieaverill/c3ba17e20a445a444d48ec22ef7f9227

## Steps to Reproduce

* Create a user account with your email address
* Log out of the CMS
* Click on "forgot my password"
* Enter your email address and submit
* Follow the "reset your password" link in the email you receive 
* Enter a new password and submit
* Log out
* Follow the "reset your password" link in the email again

HARVS1789UK:
@robbieaverill I can confirm that I am also seeing this issue in v 4.1.0-rc2 when following a framework generated password reset e-mail link.

What is more of an issue for me, is that I see this error the **first** time I follow a password reset link which is generated by my codebase!

I generate a password reset link on creation of a new member and e-mail them this link in an 'notification of new account creation' e-mail (so they can set themselves a password for, as they see it, the first time). I get this error when following that link for the first time, meaning users cannot set their own passwords.

Have you found any causes of this yet? I need to find a fix before this project can go live :-(

dhensby:
@HARVS1789UK have you done any debugging around this issue?

HARVS1789UK:
@dhensby - Yes some, I have taken note of the state of the Member record in the DB in these 3 scenarios.

### Framework Reset - Pre Reset
**TempIDHash**: null
**TempIDExpired**: null
**AutoLoginHash**: $2y$10$ae12a0a21f4ffda80dff2ud1HrolzkPdw.f.ARXLcZx13uO2/utQ6
**AutoLoginExpired**: 2018-03-09 11:38:44
**Note**: HTTP 302 redirect to strip query params

### Framework Reset - Post Reset
**TempIDHash**: 7ecf0cd09b96ee601c0dda361c20a99b2c4b2a5d
[...]

PR #7917, opened 2018-03-07
Bug Fix - Change Password Page

HARVS1789UK:
Bugfix - Correct duplicate nesting of 'Content' to be returned to template

In scenarios where:

- No member is logged in
- An 'AutoLoginHash' is provided via the 't' (token) query param
- The token isn't valid (determined by Member::validateAutoLoginToken())

The message which is intended to be returned to the end-user via $Content
in the template, is mistakenly double nested in ['Content' => ['Content' => 'Message']]
this leads to "The method forTemplate() doesn't exist on ArrayData" errors.

See - [silverstripe#7866](https://github.com/silverstripe/silverstripe-framework/issues/7866)

### DIFF
--- a/src/Security/MemberAuthenticator/ChangePasswordHandler.php
+++ b/src/Security/MemberAuthenticator/ChangePasswordHandler.php
@@ -112,21 +112,19 @@ class ChangePasswordHandler extends RequestHandler
         }
         // Show a friendly message saying the login token has expired
         if ($token !== null && $member && !$member->validateAutoLoginToken($token)) {
-            $message = [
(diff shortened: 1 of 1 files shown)


dhensby:
@HARVS1789UK you need to cherry-pick your commit onto 4.0 branch and force push that up to the PR.

```sh
$ git checkout bug-fix-change-password
$ git reset --hard 4.0
$ git pull https://github.com/silverstripe/silverstripe-framework
$ git cherry-pick 6138710
$ git push -f
```

PR #7922, opened 2018-03-11
Updated the DocBlock for ManyManyList's add() method

mooror:
This is a **very small** update to the DocBlock for ManyManyList's add() method. It took me a few hours (on a couple of occasions) to figure out how to update extra field data stored in already existing relationship records. Therefor I thought I would help my future self by updating the add() methods DocBlock which presumably will update the API documentation.

I apologize in advance if this is not the right format for a pull request. I am in the middle of a client project and just want to get this ball rolling.

### DIFF
--- a/model/ManyManyList.php
+++ b/model/ManyManyList.php
@@ -188,6 +188,13 @@ class ManyManyList extends RelationList {
 	 * Add an item to this many_many relationship
 	 * Does so by adding an entry to the joinTable.
 	 *
+	 * Can also be used to update an already existing joinTable entry
+	 *
+	 * Example:
+	 *
+	 *     $manyManyList->add($recordID,["ExtraField"=>"value"]);
+	 *
+	 *
 	 * @throws InvalidArgumentException
 	 * @throws Exception
 	 *


mooror:
I can also right up something a bit more substantial for the 3.6 Developer Guide documentation if needed.

mooror:
I can also right up something a bit more substantial for the 3.6 Developer Guide documentation if needed.

flamerohr:
ironically `extraFields` usage isn't even documented :P
https://docs.silverstripe.org/en/4/developer_guides/model/relations/#automatic-many-many-table

tractorcow:
Yeah I want to add it to the Relation interface in 5.x, and make it a standard relation API.

dhensby:
I've noticed that you've created this pull request from one of the "core branches" rather than a branch specifically intended for this pull request. This can cause problems if we ask you to target this PR at a different branch. No action is needed at the moment, this is for your information.

dhensby:
Thanks, I've pointed this PR to the 3.6 branch, as that appeared to be the intended target and I merged by hand so I could modify the short array syntax (3.6 supports PHP 5.3)

mooror:
Thank you @dhensby for taking the time to merge this. I apologize for the improper branching approach that I took. I will try to take more time when creating pull requests in the future so these problems can be avoided.

Thanks again,
Mooror

mooror:
Oh and can someone regenerate/refresh the API docs so that the updated Add() documentation will show up on the [ManyManyList API page](http://api.silverstripe.org/3/ManyManyList.html#method_add). I assume silverstripe uses some kind of DocBlock to HTML/Documentation generator. At least I hope so ;)

dhensby:
the docs will update once this is merged into the 3 or  4 branch, I think

PR #7925, opened 2018-03-13
Add Paul to core designers team

flamerohr:


flamerohr:
Agreed, lets keep it as one list for now :)

PR #7927, opened 2018-03-14
Fixing FormAction API documentation

3Dgoo:
The API documentation for FormAction is currently broken:
http://api.silverstripe.org/3/FormAction.html

This is because there are unescaped tags in the class description.

This fix changes the tag in the description to be escaped.

### DIFF
--- a/forms/FormAction.php
+++ b/forms/FormAction.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * The action buttons are <input type="submit"> as well as <button> tags.
+ * The action buttons are &lt;input type="submit"&gt; as well as &lt;button&gt; tags.
  *
  * Upon clicking the button below will redirect the user to doAction under the current controller.
  *
@@ -25,7 +25,7 @@ class FormAction extends FormField {
 	protected $action;
 
 	/**
-	 * Enables the use of <button> instead of <input>
+	 * Enables the use of &lt;button&gt; instead of &lt;input&gt;
 	 * in {@link Field()} - for more customizeable styling.
 	 *
 	 * @var boolean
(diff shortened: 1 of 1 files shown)


robbieaverill:
Note: I've cancelled the Travis builds manually to free up the queue for the 4.1 release backlog

tractorcow:
Those API docs have turned out to be a real problem. Sorry about that.

3Dgoo:
No problem. 

@robbieaverill I have updated the pull request to wrap code tags around the existing input and button tags. 

Does this look better now?

robbieaverill:
It does, but to be honest having read what I suggested again I actually don't know if it will fix the problem. We should probably run it locally and work out which works best. @tractorcow what do you think?

Reference issue: https://github.com/FriendsOfPHP/Sami/issues/261

3Dgoo:
No dramas. Let me know if there are any changes you want me to make.

robbieaverill:
Yep, will do - thanks and sorry for holding this up! I'll assign to me and do some tests

robbieaverill:
Ok I've done some testing and it's not too bad, you can use backticks to escape HTML in PHPDoc blocks like you do with markdown. If you wouldn't mind updating, I think that's a reasonable compromise between "the API site handles the doc blocks" and "I can read the doc blocks in an IDE".

3Dgoo:
Thanks for the info @robbieaverill . I have updated the pull request to wrap backticks around the existing input and button tags.

3Dgoo:
Is this change easy enough to push to 4 branch?
http://api.silverstripe.org/4/SilverStripe/Forms/FormAction.html

Or would you prefer a separate pull request?

robbieaverill:
It will get merged up to 4 at some stage, but you're welcome to make another PR if you'd like to

PR #7928, opened 2018-03-14
API Implement many_many through polymorphic (from only)

### LINKED ISSUE #3136: ManyManyList does not know about the object that is managing
Shouldn't `ManyManyList` know about the object whose relationship it is managing?

The logic at the moment is being passed in the `DataObject->getManyManyComponent` method here: https://github.com/silverstripe/silverstripe-framework/blob/master/model/DataObject.php#L1683

So there are no (clear) parameters like `$managingClass`, `$managingRelationship`, `$managingObjectID`.

This is perhaps something that is endemic across all `RelationList` classes.

### LINKED ISSUE #7911: belongs_many_many doesn’t work as documented with many-many-through relations
```php
// Project.php
class Project extends DataObject
{
    private static $belongs_many_many = [
        'Types' => Type::class
    ];
}

// Type.php
class Type extends DataObject
{
    private static $many_many = [
        'Projects' => [
            'through' => ProjectType::class,
            'from' => 'Type',
            'to' => 'Project'
        ]
    ];
}

// ProjectType.php
class ProjectType extends DataObject
{
    private static $has_one = [
        'Project' => Project::class,
        'Type' => Type::class
    ];
}
```

This setup works in that all the appropriate tables are created on dev/build, but as soon as a `Types` field is added to `Projects` it fails:

> Uncaught LogicException: belongs_many_many relation Project.Types points to Type without matching many_many

Triggered here: https://github.com/silverstripe/silverstripe-framework/blob/15410cb67a4744e0db7799e730b5b239fbdb99a4/src/ORM/DataObjectSchema.php#L795-L799 It appears that `DataObjectSchema::parseBelongsManyManyComponent()` doesn’t cater for many-many-through relations.

I can work around the issue by supplying dot-notation, but it seems to me that this shouldn’t be required:

```php
private static $belongs_many_many = [
    'Types' => Type::class . '.Projects'
];
```

Relevant documentation: https://docs.silverstripe.org/en/4/developer_guides/model/relations/#many-many-through-relationship-joined-on-a-separate-dataobject

tractorcow:
Parent story https://github.com/silverstripe/silverstripe-versioned/issues/75

This is a stand-alone feature that's ready to merge. :)

Fixes #3136
Fixes #7911

### DIFF
--- a/docs/en/02_Developer_Guides/00_Model/02_Relations.md
+++ b/docs/en/02_Developer_Guides/00_Model/02_Relations.md
@@ -339,6 +339,46 @@ $supporters = $team->Supporters()->where(['"TeamSupporter"."Ranking"' => 1]);
 Note: ->filter() currently does not support joined fields natively due to the fact that the
 query for the join table is isolated from the outer query controlled by DataList.
 
+### Polymorphic many_many (Experimental)
+
+Using many_many through, it is possible to support polymorphic relations on the mapping table.
+Note, that this feature is currently experimental, and has certain limitations:
+ - This feature only works with many_many through
+ - This feature will only allow polymorphic many_many, but not belongs_many_many. However,
+   you can have a has_many relation to the mapping table on this side, and iterate through this
+   to collate parent records.
+
+For instance, this is how you would link an arbitrary object to many_many tags.
+
+```php
+use SilverStripe\ORM\DataObject;
+
+class SomeObject extends DataObject 
+{
+    // This same many_many may also exist on other classes
+    private static $many_many = [
+        "Tags" => [
+            'through' => TagMapping::class,
+            'from' => 'Parent',
+            'to' => 'Tag',
+        ]
+    ];
+}
+class Tag extends DataObject 
+{
+    // has_many works, but belongs_many_many will not
+    private static $has_many = [
+        'TagMappings' => TagMapping::class,
+    ];
+}
+class TagMapping extends DataObject 
+{   
+    private static $has_one = [
+        'Parent' => DataObject::class, // Polymorphic has_one
+        'Tag' => Tag::class,
+    ];
+}
+```
 
 ### Using many_many in templates
 
--- a/docs/en/02_Developer_Guides/00_Model/02_Relations.md
+++ b/docs/en/02_Developer_Guides/00_Model/02_Relations.md
@@ -370,6 +370,19 @@ class Tag extends DataObject
     private static $has_many = [
         'TagMappings' => TagMapping::class,
     ];
+    
+    /**
+     * Example iterator placeholder for belongs_many_many.
+     * This is a list of arbitrary types of objects
+     * @return Generator|DataObject[]
+     */
+    public function TaggedObjects()
+    {
+        foreach ($this->TagMappings() as $mapping) {
+            yield $mapping->Parent();
+        }
+    }
+    
 }
 class TagMapping extends DataObject 
 {   
--- a/src/ORM/ManyManyThroughList.php
+++ b/src/ORM/ManyManyThroughList.php
@@ -4,6 +4,7 @@ namespace SilverStripe\ORM;
 
 use BadMethodCallException;
 use InvalidArgumentException;
+use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Injector\Injector;
 
 /**
@@ -216,5 +217,17 @@ class ManyManyThroughList extends RelationList
         $record = $hasManyList->createDataObject($extraFields ?: []);
         $record->$localKey = $itemID;
         $hasManyList->add($record);
+    }
+
+    /**
+     * Get extra fields used by this list
+     *
+     * @return array a map of field names to types
+     */
+    public function getExtraFields()
+    {
+        // Inherit config from join table
+        $joinClass = $this->manipulator->getJoinClass();
+        return Config::inst()->get($joinClass, 'db');
     }
 }
--- a/src/ORM/ManyManyThroughList.php
+++ b/src/ORM/ManyManyThroughList.php
@@ -25,14 +25,30 @@ class ManyManyThroughList extends RelationList
      * @param string $localKey The key in the join table that maps to the dataClass' PK.
      * @param string $foreignKey The key in the join table that maps to joined class' PK.
      *
+     * @param array $extraFields Ignored for ManyManyThroughList
+     * @param string $foreignClass 'from' class
+     * @param string $parentClass Parent class (should be subclass of 'from')
      * @example new ManyManyThroughList('Banner', 'PageBanner', 'BannerID', 'PageID');
      */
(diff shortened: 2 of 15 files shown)


src/ORM/ManyManyThroughList.php
```diff
@@ -25,14 +25,30 @@ class ManyManyThroughList extends RelationList
      * @param string $localKey The key in the join table that maps to the dataClass' PK.
      * @param string $foreignKey The key in the join table that maps to joined class' PK.
      *
+     * @param array $extraFields Ignored for ManyManyThroughList
```
kinglozzer: This makes me sad :(. I’m guessing it’s because `DataObject::inferReciprocalComponent()` and `DataObject::getManyManyComponents()` don’t know whether they’re constructing `ManyManyList` or `ManyManyThroughList`. Could we add an `if ($manyManyComponent['relationClass'] === ManyManyList::class)` where they’re constructed and pass different args?

src/ORM/ManyManyThroughList.php
```diff
@@ -145,14 +161,18 @@ public function add($item, $extraFields = [])
         if (is_numeric($item)) {
             $itemID = $item;
         } elseif ($item instanceof $this->dataClass) {
+            /** @var DataObject $item */
+            if (!$item->isInDB()) {
+                $item->write();
```
kinglozzer: Will this and `$hasManyList->add($record);` (L217) result in duplicate writes?

docs/en/02_Developer_Guides/00_Model/02_Relations.md
```diff
@@ -339,6 +339,46 @@ $supporters = $team->Supporters()->where(['"TeamSupporter"."Ranking"' => 1]);
...
 query for the join table is isolated from the outer query controlled by DataList.
 
+### Polymorphic many_many (Experimental)
+
+Using many_many through, it is possible to support polymorphic relations on the mapping table.
+Note, that this feature is currently experimental, and has certain limitations:
+ - This feature only works with many_many through
+ - This feature will only allow polymorphic many_many, but not belongs_many_many. However,
+   you can have a has_many relation to the mapping table on this side, and iterate through this
+   to collate parent records.
```
kinglozzer: > iterate through this to collate parent records.

Took me a couple of attempts to understand this, could you add an example further down and link this to it? Am I right in thinking it would work something like this?

```php
// Filtering by tags from the "some object" side
$taggedSomeObjects = SomeObject::get()->filter(['Tag.Name' => ['New', 'Featured']]);

// Or by collating IDs from the "tag" side
$tags = Tag::get()->filter(['Title' => ['New', 'Featured']]);
$taggedObjectIDs = $tags->relation('TagMappings')
    ->filter(['ParentClass' => SomeObject::class])
    ->column('ParentID');
$taggedSomeObjects = SomeObject::get()->byIDs($taggedObjectIDs);
```

src/ORM/ManyManyThroughQueryManipulator.php
```diff
@@ -36,19 +36,37 @@ class ManyManyThroughQueryManipulator implements DataQueryManipulator
...
      * will be infered at evaluation from query parameters set via the ManyManyThroughList
      *
      * @param string $joinClass Class name of the joined dataobject record
      * @param string $localKey The key in the join table that maps to the dataClass' PK.
      * @param string $foreignKey The key in the join table that maps to joined class' PK.
+     * @param string $foreignClass the 'from' class name
+     * @param string $parentClass Name of parent class. Subclass of $foreignClass
      */
-    public function __construct($joinClass, $localKey, $foreignKey)
+    public function __construct($joinClass, $localKey, $foreignKey, $foreignClass, $parentClass)
```
kinglozzer: Can we make these new arguments optional? This is a fairly internal API, but I don’t think it’s totally unreasonable to expect that someone could be using this class directly

docs/en/02_Developer_Guides/00_Model/02_Relations.md
tractorcow: Yeah I can add the generator in the example.

src/ORM/ManyManyThroughList.php
tractorcow: The arg is ignored because extraFields come from the join model. We aren't ignoring extraFields feature entirely, it's just that the constructor doesn't need it.

> Could we add an if ($manyManyComponent['relationClass'] === ManyManyList::class) where they’re constructed and pass different args?

I'd rather keep it consistent and agnostic to the underlying class constructed. That's the point of soft-coding the relation name.

In the future we may add a third type for PolymorphicManyManyThroughList (e.g.).

src/ORM/ManyManyThroughList.php
tractorcow: No because this writes the RHS record. The manyManyList writes the join record. Separate models ;)

src/ORM/ManyManyThroughQueryManipulator.php
tractorcow: They aren't optional though. ;) This class needs it to work.

src/ORM/ManyManyThroughList.php
tractorcow: In 5.x we could switch to named args (one constructor arg, array key in same form produced by DataObjectSchema::manyManyComponent().

kinglozzer: Okay

src/ORM/ManyManyThroughQueryManipulator.php
kinglozzer: I’m just being a militant semversionist. `... $foreignClass = null, $parentClass = null)` should work and be backwards compatible shouldn’t it? I can’t spot any code in this class that would break if those props are just `null`

tractorcow: null means unknown, so without those variables it may mistakenly not mock a polymorphic relation which is polymorphic. The ambiguity introduces a functional error, not a syntactic one.

tractorcow: In this case I could just make it a warning, so sure.

tractorcow:
@kinglozzer updated with the things you requested.

tractorcow:
1 bugfix coming :)

tractorcow:
Pushed.

docs/en/02_Developer_Guides/00_Model/02_Relations.md
tractorcow: Added this.

```php
    public function TaggedObjects()
    {
        foreach ($this->TagMappings() as $mapping) {
            yield $mapping->Parent();
        }
    }
```

tractorcow:
I've added another feature the parent story requires, although not strictly related to many_many.

tractorcow:
I found out manymanythrough wouldn't duplicate properly without a getExtraFields(). Added this in. :)

Issue #7929, opened 2018-03-14
Incorrect return type FieldList::removeByName


robbieaverill:
@worikgh commented on [Fri Feb 23 2018](https://github.com/silverstripe/api.silverstripe.org/issues/71)

It is listed at http://api.silverstripe.org/4/SilverStripe/Forms/FieldList.html#method_removeByName as returning $this
It does not.  FieldList::removeByname(..) returns null

---

@robbieaverill commented on [Fri Feb 23 2018](https://github.com/silverstripe/api.silverstripe.org/issues/71#issuecomment-367857349)

Thanks. Would you like to PR to the framework to fix the php doc return declaration?

robbieaverill:
Ok so in this example @worikgh, commit 91f1a58018 actually did make this method return itself. This is available in 4.1 onwards, so devs using 4.0 will still find this an issue.

That said - it still returns null in some conditions: https://github.com/silverstripe/silverstripe-framework/blob/4.1/src/Forms/FieldList.php#L354

I'll make a PR to fix that, then I think this issue would be resolved.

PR #7930, opened 2018-03-14
FIX Ensure consistent return type for FieldList::removeByName

robbieaverill:
This method was made chainable in https://github.com/silverstripe/silverstripe-framework/commit/91f1a58018b208b3677550dc4d22ac5d83f5124a but we missed a spot :)

Resolves https://github.com/silverstripe/silverstripe-framework/issues/7929


dhensby:
4.1.1 here we come

PR #7935, opened 2018-03-15
Fixing FormAction API documentation SS4

3Dgoo:
The API documentation for FormAction is currently broken:
https://api.silverstripe.org/4/SilverStripe/Forms/FormAction.html

This is because there are unescaped tags in the class description.

This fix wraps the tags in backticks.

This is the ss4 version of the #7927 pull request.


dhensby:
We'll get those changes merged up in time

PR #7939, opened 2018-03-15
Ensure tmpData exists

lekoala:
Otherwise you might get illegal string offset


tractorcow:
@lekoala can you give an example which would produce the array offset error?

tractorcow:
Seems ok logically. The whole code block is a bit hard to read but that's an existing issue. ;)

lekoala:
I'm not quite sure, I'm having some fields with lots of nesting like Data[Items][1][MyField] (that is serialized to json) and sometimes it happens. the changes has no side effects except preventing a php error so that cannot hurt :-)

tractorcow:
Yes I think so.

PR #7945, opened 2018-03-20
FIX: Don't auto-generate indexes for Text field types (fixes #7900)

### LINKED ISSUE #7900: Invalid indexes generated from default_sort
As reported by @wernerkrauss on Slack

## Affected Version

3.7 (unreleased), 4.1.0 (unreleased)

## Description

We automatically create indexes, if they’re not already defined, for fields set in `$default_sort` (introduced in https://github.com/silverstripe/silverstripe-framework/pull/7559). However, if the field is of type `Text`, the index is invalid in MySQL and causes an error on build:

```
[Emergency] Uncaught SilverStripe\ORM\Connect\DatabaseException:
Couldn't run query: ALTER TABLE "<table name>"
ADD index "Name" ("Name") BLOB/TEXT
column 'Name' used in key specification without a key length
```

MySQL `TEXT` indexes need a length specified - https://stackoverflow.com/questions/1827063/mysql-error-key-specification-without-a-key-length/1827099#1827099. I’ve not investigated whether the same applies to any other field types, or other database drivers.

I’d suggest we just update the logic to not bother creating an index for “unsafe” data types.

## Steps to Reproduce

```php
class MyDataObject extends DataObject
{
    private static $db = [
        'SomeField' => 'Text'
    ];

    private static $default_sort = 'SomeField ASC';
}
```

kinglozzer:


model/DataObject.php
```diff
@@ -3421,7 +3421,12 @@ public function databaseIndexes() {
 						continue;
 					}
 
-					if ($this->hasOwnTableDatabaseField($column) && !array_key_exists($column, $indexes)) {
+					list($fieldType) = Object::parse_class_spec($this->db($column));
+					if (
+						$this->hasOwnTableDatabaseField($column)
+						&& !array_key_exists($column, $indexes)
+						&& $fieldType !== 'Text' && !is_subclass_of($fieldType, 'Text')
```
dhensby: I'd prefer a less hard-coded approach here, where we can set a config list of blacklisted index types?

kinglozzer: Yeah sure. Do you think `DataObject` makes sense, or should it be a setting like `DBField.auto_indexable` and then set `private static $auto_indexable = false;` on `Text`?

dhensby: oooh, I like the latter idea 👍

kinglozzer:
Ping @dhensby

model/DataObject.php
```diff
@@ -3421,7 +3421,15 @@ public function databaseIndexes() {
 						continue;
 					}
 
-					if ($this->hasOwnTableDatabaseField($column) && !array_key_exists($column, $indexes)) {
+					list($fieldType) = Object::parse_class_spec($this->db($column));
```
dhensby: Needs to be `SS_Object`

kinglozzer: Fixed

dhensby:
merge on green

PR #7950, opened 2018-03-21
Update 4.1.0.md

andrewandante:
Add the `"type": "silverstripe-recipe"` line that is required as per https://github.com/silverstripe/vendor-plugin/issues/4


tractorcow:
True as of recipe 4.1.0 so will merge.

Issue #7948, opened 2018-03-21
Undefined class constant 'MODULE_FILTER'


ntd:
I'm upgrading a quite basic project from 4.0.1 to 4.1.0. This is the result after the first try:
```
$ composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 28 updates, 0 removals
  - Updating silverstripe/vendor-plugin (1.2.0 => 1.3.1): Downloading (100%)         
  - Updating symfony/yaml (v3.4.3 => v3.4.6): Downloading (100%)         
  - Updating symfony/polyfill-mbstring (v1.6.0 => v1.7.0): Downloading (100%)         
  - Updating symfony/translation (v2.8.33 => v2.8.36): Downloading (100%)         
  - Updating symfony/filesystem (v3.4.3 => v3.4.6): Downloading (100%)         
  - Updating symfony/config (v3.4.3 => v3.4.6): Downloading (100%)         
  - Updating symfony/polyfill-apcu (v1.6.0 => v1.7.0): Downloading (100%)         
  - Updating psr/simple-cache (1.0.0 => 1.0.1): Downloading (100%)         
  - Updating symfony/cache (v3.4.3 => v3.4.6): Downloading (100%)         
  - Updating swiftmailer/swiftmailer (v5.4.8 => v5.4.9): Downloading (100%)         
  - Updating symfony/finder (v3.4.3 => v3.4.6): Downloading (100%)         
  - Updating silverstripe/config (1.0.1 => 1.0.4): Downloading (100%)         
  - Updating silverstripe/framework (4.0.1 => 4.1.0): Downloading (100%)         
PHP Fatal error:  Undefined class constant 'MODULE_FILTER' in .../vendor/silverstripe/vendor-plugin/src/Library.php on line 230

Fatal error: Undefined class constant 'MODULE_FILTER' in .../vendor/silverstripe/vendor-plugin/src/Library.php on line 230
```

I searched through the [upgrading guide](https://docs.silverstripe.org/en/4/changelogs/4.1.0/#upgrading) but found nothing relevant. A second try solved the issue:
```
$ composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 16 updates, 0 removals
  - Updating silverstripe/framework (4.0.1 => 4.1.0): Loading from cache
Exposing web directories for silverstripe-vendormodule silverstripe/framework with method auto:
  - client/images
[...]

PR #7953, opened 2018-03-22
DOCS rerun composer on undefined MODULE_FILTER (#7948)

ntd:
Just a mention that you need to rerun composer when upgrading to 4.1.0. Not sure it is still relevant after @tractorcow silverstripe/vendor-plugin#18 ... feel free to reject.


tractorcow:
I fixed the root issue at https://github.com/silverstripe/vendor-plugin/pull/18 :D

tractorcow:
I think these docs can still be merged anyway.

PR #7956, opened 2018-03-26
Documentation for graphql customisation

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-graphql/issues/13

### DIFF
--- a/docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
+++ b/docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
@@ -867,7 +867,7 @@ The best thing you can do to make your code extensible is to use `Injector` earl
 
 First, let's break up the list into smaller components.
 
-*myapp/client/components/notesList.js*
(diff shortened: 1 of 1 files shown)


docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
```diff
@@ -795,4 +795,511 @@ You could manipulate the action called by the originalReducer, there isn't an ex
...
+    };
+  })
+};
+
+export default transformCreateNote;
+```
+
+All we've done here is overridden the `props` setting in the `CreateNote` apollo config. Recall from the previous section that it maps the `mutate` function to the `onAdd` prop. Since we've changed the signature of that function, we'll need to override the entire prop.
+
+Now we just need to register these transforms, and we're done!
```
flamerohr: would be a good idea to note the importance of spreading the `prevProps` at the end :)

easily missed and you lose previous transformations/properties if forgotten.

tractorcow: @unclecheese can you action this please?

PR #7957, opened 2018-03-26
DOCS: fix parsed base_tag code in alert note

rotassator:
It's a bit hard to fix the *alert* blocks, as the preview doesn't display them at all. Using HTML entities should do the trick though.


robbieaverill:
I disagree with this change, sorry. The issue with not parsing markdown inside the HTML blocks is one that lies with silverstripe/docsviewer, but I think this reduces readability of the docs themselves.

PR #7961, opened 2018-03-27
Use ModuleResourceLoader for fixtures path

wilr:
This adds the ability to refer to a fixture file path by the ModuleResourceLoader path (e.g  protected static $fixture_file = 'vendor/module:tests/fixtures.yml')


robbieaverill:
One issue I have with this is that we shouldn't have test dependencies between modules - is there a reason you'd need this other than that?

wilr:
Well for instance this is in the case of silvershop + children modules and other groups of modules which have a large amount of fixtures to get started be nice to say "Use the base fixtures from silvershop-core" which sets up page types etc rather than copying that to the `silvershop-discounts` module.

robbieaverill:
Right - true, could be useful for large app integration suites

Issue #7962, opened 2018-03-28
BUG in polymorphic ownership


tractorcow:
4.x-dev we introduced polymorphic many_many (partially). There are a couple of places where we need to soft-ignore polymorphic relationships when inferring ownership.

However, we should still be able to apply explicit ownership (via `owned_by => some polymorphic relationship`) which doesn't need to use this inference.

We need to fix framework to not error in those cases.

PR #7963, opened 2018-03-28
BUG Change inverse polymorphic inference to silent-empty instead of explicit error

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/7962


tractorcow:
Yep np.

tractorcow:
Done @kinglozzer thanks!

PR #7965, opened 2018-03-28
Fix for "too few parameters" error when using DBMultiEnum

andreaslindahl:
When adding a DBMultiEnum and doing a dev/build the following error message is displayed: 

"Too few arguments to function SilverStripe\ORM\FieldType\DBMultiEnum::__construct(), 0 passed and at least 1 expected". The field in by $db array looks like this: "Status" => "MultiEnum('Option1, Option2, Option3', 'Option2')""

This is a fix for that bug.

### DIFF
--- a/src/ORM/FieldType/DBMultiEnum.php
+++ b/src/ORM/FieldType/DBMultiEnum.php
@@ -11,7 +11,7 @@ use SilverStripe\Forms\CheckboxSetField;
  */
 class DBMultiEnum extends DBEnum
 {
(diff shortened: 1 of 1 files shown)


kinglozzer:
This is one of two possible fixes for this bug. The other is stopping `DataObjectSchema::cacheDatabaseFields()` from constructing singletons for each field.

I’m inclined to accept this fix, because all our other `DBField` classes use `= null` on constructor arguments to allow us to construct singletons for things like casting. Probably a sign that we should decouple our field types and casting 😁

PR #7971, opened 2018-03-29
FIX Config from extension should not override

### LINKED ISSUE #7970: Config definitions in Extension instances override YAML definitions
## Affected Version

^4

## Description

Config settings defined in yaml and `private statics` is overriden by `Extension` instances, rendering it impossible for user code to override default config values from modules.

Yaml and `private static`s should have preference over those defined on `Extension` classes.

## Steps to Reproduce

This can be reproduced by following the instructions given by the comments module as reported here: https://github.com/silverstripe/silverstripe-comments/issues/242

dhensby:
fixes #7970

### DIFF
--- a/docs/en/02_Developer_Guides/04_Configuration/00_Configuration.md
+++ b/docs/en/02_Developer_Guides/04_Configuration/00_Configuration.md
@@ -153,12 +153,12 @@ the result will be the higher priority false-ish value.
 
 The locations that configuration values are taken from in highest -> lowest priority order are:
 
-- Any values set via a call to Config#merge / Config#set
+- Runtime modifications, ie: any values set via a call to `Config::inst()->update()`
 - The configuration values taken from the YAML files in `_config/` directories (internally sorted in before / after 
 order, where the item that is latest is highest priority)
-- Any static set on an "additional static source" class (such as an extension) named the same as the name of the property
 - Any static set on the class named the same as the name of the property
 - The composite configuration value of the parent class of this class
+- Any static set on an "additional static source" class (such as an extension) named the same as the name of the property
 
 <div class="notice">
 It is an error to have mixed types of the same named property in different locations. An error will not necessarily
--- a/docs/en/04_Changelogs/4.0.5.md
+++ b/docs/en/04_Changelogs/4.0.5.md
@@ -1,0 +1,12 @@
+# 4.0.5
+
+## Notable changes
+
+Fix [#7971](https://github.com/silverstripe/silverstripe-framework/pull/7971) introduces a subtle change of behaviour
+to how `Config` settings are prioritised. In SilverStripe 4 there was a change where `Extension` objects took the
+highest importance when determining Config values; this was deemed to be unexpected and unintuitive as well as making it
+cumbersome and difficult for developers to override module defaults defined in `Extension`s. This change reverts
+behaviour to that of SilverStripe 3 where `Extension` instances are of lowest importance and are used only to set a
+default value. If you rely on your `Extension` or module providing an overriding config value, please move this to yaml.
+
+<!--- Changes below this line will be automatically regenerated -->
--- a/src/Core/Config/Middleware/ExtensionMiddleware.php
+++ b/src/Core/Config/Middleware/ExtensionMiddleware.php
@@ -39,7 +39,7 @@ class ExtensionMiddleware implements Middleware
         }
 
         foreach ($this->getExtraConfig($class, $config, $excludeMiddleware) as $extra) {
-            $config = Priority::mergeArray($extra, $config);
+            $config = Priority::mergeArray($config, $extra);
         }
         return $config;
     }
--- a/tests/php/ORM/DataObjectTest.php
+++ b/tests/php/ORM/DataObjectTest.php
@@ -1062,72 +1062,88 @@ class DataObjectTest extends SapphireTest
 
         // Test logical fields (including composite)
         $teamSpecifications = $schema->fieldSpecs(DataObjectTest\Team::class);
+        $expected = array(
+            'ID',
+            'ClassName',
+            'LastEdited',
+            'Created',
+            'Title',
+            'DatabaseField',
+            'ExtendedDatabaseField',
+            'CaptainID',
+            'FounderID',
+            'HasOneRelationshipID',
+            'ExtendedHasOneRelationshipID'
+        );
+        $actual = array_keys($teamSpecifications);
+        sort($expected);
+        sort($actual);
         $this->assertEquals(
-            array(
-                'ID',
-                'ClassName',
-                'LastEdited',
-                'Created',
-                'Title',
-                'DatabaseField',
-                'ExtendedDatabaseField',
-                'CaptainID',
-                'FounderID',
-                'HasOneRelationshipID',
-                'ExtendedHasOneRelationshipID'
-            ),
-            array_keys($teamSpecifications),
+            $expected,
+            $actual,
             'fieldSpecifications() contains all fields defined on instance: base, extended and foreign keys'
         );
 
         $teamFields = $schema->databaseFields(DataObjectTest\Team::class, false);
+        $expected = array(
+            'ID',
+            'ClassName',
+            'LastEdited',
+            'Created',
+            'Title',
+            'DatabaseField',
+            'ExtendedDatabaseField',
+            'CaptainID',
+            'FounderID',
+            'HasOneRelationshipID',
+            'ExtendedHasOneRelationshipID'
+        );
+        $actual = array_keys($teamFields);
+        sort($expected);
+        sort($actual);
         $this->assertEquals(
-            array(
-                'ID',
-                'ClassName',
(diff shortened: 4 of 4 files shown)


tractorcow:
Well, my feeling is that they should... that's what extensions are supposed to do. They are applied after both statics and yml config, and apply more "business level logic" over the top of that base config.

You'll find that this breaks as much as it solves.

This is because:
 - both private statics and yml config are "core" sources of config which are statically cached and stored against the root classes inside the config module.
 - extensions are dynamic decorators of config which are injected after the fact by framework.
 
I think the problem here is the assumption that yml config is "user overridden" config, where you can just as easily have user config overridden by an extension class as a yml file in `mysite/_config`, and core config in 'vendor/silverstripe/assets/_config` that you may need to modify via extensions.

The problem here is that this makes extensions less powerful; A core class (private statics + yml in a core module) can no longer be modified by an extension (even a user-code extension).

The reality is that both yml config or extensions could just as easily be treated as the base config, but with the status quo, it's easier for user extensions to dynamically decorate classes (e.g. the way that fluent does via get_extra_statics) than dynamic yml (which doesn't exist) could core extensions.

On top of all of these layers, by the way, is the "user modified" deltas, which are applied via `Config::modify()`, which is treated again as a more volatile source of config.

I respect that https://github.com/silverstripe/silverstripe-comments/issues/242 is an issue, but I don't think flipping the core convention at this point in time is a fair solution to this.

A better solution is to change the config provided by the extension to a separate key, e.g.

```php
private static $comments_defaults = [
]
```

And ensuring that `getCommentsOptions` did a recursive merge of `$config->get('comments')` and `$config_>get('comments_defaults')` to respect user-assigned values over the provided defaults.

dhensby:
OK - I've always thought that yaml / user statics were the source of truth and that extensions were just there for modules to provide a base of usability that should be easily override-able but not capable of changing a developer's explicitly defined behaviour.

I agree this change has the ability to break as much as it fixes in theory, though not sure how much in practice.

> The problem here is that this makes extensions less powerful; A core class (private statics + yml in a core module) can no longer be modified by an extension (even a user-code extension).

Sure, this is what yaml is for, I thought, and the `before` and `after` statements.

happy to close :)

tractorcow:
Ok, but please let's not merge this still. I'm against this as a fix, but I do want to see other options for allowing users to control merge order (e.g. opt in feature for extensions). You could set an extension as low priority (only adds config if not specified).

tractorcow:
@dhensby I've been thinking about this and I think I have a solution. One that solves MANY problems with extension.

How about we add `extensionOrder` config? In addition, extensions with negative order will be merged in before 'other' config (statics + yml), meaning that you can have a low priority extension that has it's values easily replaced.

So merge order is:
 - extensions with negative order (sorded by order asc)
 - statics + yml
 - extensions with non-negative order (including 0, sorted by order asc).

We can assign a default order of 100.

In addition, this can be used to set the order that ->extend() and ->augmentSQL() and so on are called. That's something people have been asking for for years.

How does this sound?

tractorcow:
I'll re-open if you are willing to continue working on this :)

dhensby:
I think that proposal sounds cool, but I do think that we should return the default behaviour to how it was in SS 3 so that we don't have a situation where modules need a work around for 4.0, 4.1 and 4.2.

We can then work on your proposal for SS 4.3 as a new feature / API

tractorcow:
> I think that proposal sounds cool, but I do think that we should return the default behaviour to how it was in SS 3 so that we don't have a situation where modules need a work around for 4.0, 4.1 and 4.2.

Neither way is the way it was in 3.x. The whole config API has been rewritten for 4 so there's no longer a direct parallel.

Changing it to the other direction would break more than it would fix. You would mess up all extensions for 4.2 that worked with 4.0/4.1, and then re-fix it for 4.3 if they opt in to a new option. It would be a mess.

tractorcow:
Closing with the intention that this should be fixed / amended in 4.3 or higher in a non-breaking way. We can't change it in 4.0 without regressions.

robbieaverill:
Hmmm, we had a stronger quorum voting in favour of changing it than not...

tractorcow:
The majority was established on the faulty assumption this was a fix to something "accidental" and not intentional.

I've ensured that my rejection included an option to resolve the issue in another way, which addresses @chillus's concern.

> My vote goes for revert not only because of code regressions, but also because I haven't heard a good way to deal with this accidental new behaviour going forward. If you voted against the revert, can you please also vote for one or more of the five proposed options I've collected?

Although this option is more a 6th feature, but it's something we can build into the extension api that makes sense and provides extra power to developers rather than locking them into one behaviour.

tractorcow:
If the discussion isn't closed I guess we can re-open this PR.

chillu:
> The majority was established on the faulty assumption this was a fix to something "accidental" and not intentional.

Can you point to any statement made during the implementation of this config refactoring which expresses intent on this change? Even the author of most of this refactoring (@micmania1) wants it reverted.

As far as I can tell, the [config docs](https://docs.silverstripe.org/en/4/developer_guides/configuration/configuration/) directly contradict this new behaviour:

image

Regarding your `extensionOrder` proposal, isn't the order of `augmentSQL` calls etc already determined by `before` and `after` rules in YAML based on where the `extends` config value is placed? Not sure if that's explicit behaviour (can't find docs or tests about it). Why do we need yet another layer of complexity in our config system?

Note that both `extensionOrder` and `before`/`after` would need to address how priority is applied procedurally through `add_extension()`, and declaratively through `Object::$extensions`. The `extensionOrder` solution would make this more explicit, we could assign "neutral" (zero) priority to those.

tractorcow:
> Regarding your extensionOrder proposal, isn't the order of augmentSQL calls etc already determined by before and after rules in YAML based on where the extends config value is placed? Not sure if that's explicit behaviour (can't find docs or tests about it). Why do we need yet another layer of complexity in our config system?

The problem is that sometimes users want to replace an extension, but have it execute first. You can use before / after rules in yml, but it doesn't let you override and prioritise at the same time; The replacements always end up last. There isn't a ticket (I believe) but I have been asked many times "how do I make sure my extension point is called first / last".

tractorcow:
> Can you point to any statement made during the implementation of this config refactoring which expresses intent on this change? Even the author of most of this refactoring (@micmania1) wants it reverted.

Yes that's a fair point, but it was my understanding (perhaps mistakenly) during de-development of @micmania1 's original work that extensions were "on top of" the base config. I did it intentionally, and admittedly I was opinionated at the time. Whereas @micmania1 's work around collecting and sorting configs was mostly unchanged, the middleware / extension work was my own and added after I picked the story up.

That being said, I'll try not to be too precious in defence of the status quo. I just wish we would be more careful with semver.

chillu:
> I did it intentionally, and admittedly I was opinionated at the time.

@tractorcow Your basis of argument would've been a lot easier if you had actually changed the docs as well when this was introduced. This would've signalled intent, made it a lot clearer to peer reviewers what's happening, and actually gave developers a chance to work their way through our complex config system. 

> The problem is that sometimes users want to replace an extension, but have it execute first. You can use before / after rules in yml, but it doesn't let you override and prioritise at the same time; The replacements always end up last

OK, that makes sense. You could argue that the priority is based on the extension method (e.g. call `onBeforeWrite()` first, call `onBeforeDelete()` last). But that's probably discussed in a bigger picture about what we're trying to achieve with `Extension` (vs. traits plus an event system). `extensionOrder` fixes a limitation of our config system separate from this semver discussion, and also happens to provide a workable workaround for the problem at hand. 

But: I do have a problem with the default order of 100, which retains the current (apparently intentional) behaviour. That's fine for extensions in your project code. But if a module contains extensions, it's config settings "win" by default, unless the module author specifically marks them as "configurable" via defining a lower extension order. That's no less counter-intuitive than the issue which @micmania1 outlined. And if we change that default order to -100, we're breaking semver just as much as a revert would. Maybe we fix this via documentation, and discourage modules to use config statics via extensions in the first place? Instructing users of your module to add YAML config to project code is much clearer to me, even though it's extra work for the user.

dhensby:
Right, we've had a chat off GitHub and we've tested existing 3.x behaviour and discussed what we think the expected and intuitive behaviour should be versus the intended behaviour when this was written. Whilst we accept that it was fully intended at the time of writing that the extension classes be most important, we now accept that this is likely not the expected behaviour for most developers.

Therefore we're happy to merge this change in (subject to testing) and release in in the next patch releases for 4.0, 4.1 and the soon to be 4.2.

Unfortunately a lot of this confusion has actually come out of trusting our docs which are wrong for both 3 and 4 because the merge priority is incorrectly listed.

robbieaverill:
Great result all round

dhensby:
I've tested with versioned to make sure:

1. The test suite passes locally
2. I can publish pages
3. I can save pages to draft site
4. I can switch between stages via URL and see the correct versions
5. I can upload assets
6. I can publish assets
7. I can insert assets into page content areas and publish

tractorcow:
Post merge approval!

PR #7972, opened 2018-03-29
API: Use symfony/http-foundation for sessions

kinglozzer:
I was going to do an RFC, but I got so far with my proof-of-concept that I ended up carrying on with it... so this is my RFC

An alternative proposal would be to remove the `SilverStripe\Control\Session` class entirely, and make `HTTPRequest::getSession()` return an instance of `Symfony\Component\HttpFoundation\Session\SessionInterface`. I didn’t go for that approach, because we still rely on the `Foo.Bar.Baz` session-name-to-array-traversal magic for things like form errors.

Pros are slightly less code to maintain (at least less complexity handling cookies on our end) and better support for/easier to configure save handlers.

### DIFF
--- a/src/Control/Session.php
+++ b/src/Control/Session.php
@@ -3,8 +3,9 @@
 namespace SilverStripe\Control;
 
 use BadMethodCallException;
-use SilverStripe\Core\Config\Configurable;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Dev\Deprecation;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
 
 /**
  * Handles all manipulation of the session.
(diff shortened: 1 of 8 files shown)


src/Control/Session.php
```diff
@@ -224,165 +182,69 @@ public function restart(HTTPRequest $request)
...
-     */
-    protected function markChanged($name)
-    {
-        $diffVar = &$this->changedData;
-        foreach (explode('.', $name) as $namePart) {
-            if (!isset($diffVar[$namePart])) {
-                $diffVar[$namePart] = [];
+        // Traverse down existing array (adding placeholders if needed) to set the value
+        $existing = $handler->get($key);
+        $var =& $existing;
```
dhensby: Can we do `$var = &$existing`, I think that's more "standard"

src/Control/Session.php
```diff
@@ -224,165 +182,69 @@ public function restart(HTTPRequest $request)
...
+        foreach ($pieces as $namePart) {
+            if (!isset($var) || !is_array($var)) {
```
dhensby: is this a change in behaviour, it feels dangerous as if I do `Session::set('Cart.Size', 3)` and then `Session::set('Card.Size.Total', 7)` then the 'Cart.Size' value is removed and replaced with an array...

I'm not sure how that should work now, but maybe it's a regression?

dhensby:
It looks like some of the tests are failing possibly because of session leakage in tests? or maybe because the session isn't being passed correctly to the mock requests to framework?

emteknetnz:
@kinglozzer I'm currently going through some old PR's to see what is still active and what makes sense to be closed - is this one something you're still planning on keeping active or would you prefer to now close this one?

kinglozzer:
Let’s close it. I’d still like this change to happen, but I think we really need to remove the “dot-to-array” magic and handle the array traversal in user-code, rather than re-implement it all on top of a different layer... that’s a bigger job and bigger upgrade pain though so I think it’d make more sense as part of some wider request/response work rather than on its own

sminnee:
My suggestions here would be to experiment with swapping out the whole of Control, after #9679 is done. In principle your View and ORM would be able to be retained.

What the dev experience is like for that is anyone's guess, but it's worth an experiment ;-)

PR #7977, opened 2018-04-03
add config var to skip confirm logout

andrewandante:
So that one can skip straight to auto-logout, should one desire.


dhensby:
hmm - what's the motivation for this change? this feature is technically a security feature and this is an opt-out of it without any clear warning of the risks and issues...

If you want to go straight to logout without an additional skip then you should be going to the link with the security token already added...

PR #7982, opened 2018-04-04
Fix invalid markup in SelectionGroup selected values

wilr:
When providing a selected value to a `SelectionGroup` it resulted in markup - `<li1>` since `$Selected` has been changed from a string to a simple boolean. This corrects the default template so that a class is output.


wilr:
Added visibility to methods

Issue #7991, opened 2018-04-11
SS5: Better PSR-2 conformance


tractorcow:
## Affected Version

5.x-dev

## Description

A set of PSR-2 rules are ignored for framework, specifically because fixing them would require breaking changes to code. These exclusions are necessary for 4.x but should all be fixed in 5.x.

## Steps to Reproduce

```
git clone https://github.com/silverstripe/silverstripe-framework.git -b=master
cd silverstripe-framework
phpcs --standard=PSR2 ./src/ ./tests/
```

many errors :)

## Todo-list

Eliminate all PSR* exclusions in https://github.com/silverstripe/silverstripe-framework/blob/4.0/phpcs.xml.dist for master branch.

robbieaverill:
I've made a start at #7993

tractorcow:
I've been doing quite a lot of class renaming, getting rid of all classes with `_` in them, as well as `SS` prefixed classes.

tractorcow:
Branches:

 - [ ] https://github.com/open-sausages/silverstripe-framework/tree/pulls/5.0/psr2-class-renaming

sminnee:
Yeah, PSR-12, Slevomat, PHP 7.2+ or even PHP 7.3+ support would get a :+1: from me in principle.

robbieaverill:
The good news is that a lot of the slevomat standards are "fixable" with phpcbf, e.g. PHP return type hinting etc

robbieaverill:
Hey @tractorcow, with context of your WIP branches above and [RFC: Rename all classes with underscores to be PSR-2 compliant](https://github.com/silverstripe/silverstripe-framework/issues/8636), do you want to revive those branches and make PRs for them? Or should whoever picks up #8636 pick your branches up and roll with those?

robbieaverill:
I've changed this to rfc/accepted, with me @sminnee and @tractorcow in support. I find it unlikely that the rest of the core team would be against this as well

tractorcow:
Sorry @robbieaverill I'm likely not going to be able to contribute the time needed to do such a large upgrade sorry. =( I suggest someone use my branches as inspiration, and just repeat it on the latest branches.

dhensby:
I'm happy for us to aim for PSR-12 compliance but it shouldn't be done at any cost. I can see that the PSR-12 standard is much more detailed and prescriptive than the previous standards.

sminnee:
[...]

PR #7993, opened 2018-04-11
PSR-2 improvements

robbieaverill:
This pull request tidies up some PSR-2 exclusions to bring framework a little closer to PSR-2 compliant (issue: #7991). Summary:

* Line lengths fixed
* Method visibility
* Switch cases shouldn't use braces

Also adds a phpcs exclusion to SSTemplateParser and removes some PHP 5.x checks from installer requirements.


src/Core/CustomMethods.php
```diff
@@ -60,10 +60,10 @@ public function __call($method, $arguments)
         }
 
         switch (true) {
-            case isset($config['callback']): {
+            case isset($config['callback']):
                 return $config['callback']($this, $arguments);
-            }
-            case isset($config['property']) : {
+                break;
```
tractorcow: Why does a return need a break?

robbieaverill: It’s a rule isn’t it? Every case must have a break?

PR #7995, opened 2018-04-11
PasswordValidator - Impossible to pass validation if min_test_score is set

HARVS1789UK:
The `SilverStripe\Security\PasswordValidator.php` class has configuration based settings for specifying a simple minimum password strength policy via `private static $min_test_score;`

If used, there is logic in the `PasswordValidate()` method to check the required minimum score against the actual score (based on the usage of lowercase, uppercase, digit and symbol characters in the password).

However, the check to calculate the actual score incorrectly references ```count($this->testNames)``` (which defaults to ```null```) instead of using ```count($testNames)``` (which contains an array of values returned by ```public function getTestNames()```

This equates to ```$score = count(null) - count([FAILED TESTS]);``` which, if it is set, will be less than `private static $min_test_score;` (e.g. 0 at best, or -1, -2 etc).

This PR fixes this mistake/typo.

### DIFF
--- a/src/Security/PasswordValidator.php
+++ b/src/Security/PasswordValidator.php
@@ -251,7 +251,7 @@ class PasswordValidator
                 );
             }
 
-            $score = count($this->testNames) - count($missedTests);
+            $score = count($testNames) - count($missedTests);
             if ($score < $minTestScore) {
                 $error = _t(
                     'SilverStripe\\Security\\PasswordValidator.LOWCHARSTRENGTH',


HARVS1789UK:
@maxime-rainville,

I am looking at `PasswordValidatorTest::testValidateMinScore()` at the moment, but I can't really see how it isn't already failing?

I am working in `4.1.0` and I have the below:

```
    public function testValidateMinScore()
    {
        $v = new PasswordValidator();
        $v->characterStrength(3, array("lowercase", "uppercase", "digits", "punctuation"));

        $r = $v->validate('aA', new Member());
        $this->assertFalse($r->isValid(), 'Passing too few tests');

        $r = $v->validate('aA1', new Member());
        $this->assertTrue($r->isValid(), 'Passing enough tests');
    }
```

The last line of that function, the assertTrue(), should already fail based on my findings shouldn't it? e.g. 3/4 character types are provided, but it's irrelevant because of the bug in `validate()` I identified, to the assertTrue on a falsey `ValidationResult` should fail?

Either way, I am not very familiure with the test classes and I don't really have time at the moment to learn about it an include that change in my PR as well. Would you mind doing so as you offered? If you could let me know when you do so that I can review what changes you have made in an effort to learn a bit more about the test classes?

dhensby:
@HARVS1789UK because the test calls `characterStrength` which calls `setTestNames` which populates `$this->testNames`  - I think there are two problems. the one you identified and the fact you're incorrectly implementing the class? you need to make sure you call `setTestNames` and `setMinTestScore` on your validator...

HARVS1789UK:
@dhensby - I think I see what you are saying, but that would mean I have to create a new instance of `PasswordValidator` and configure it, each time I wanted a form etc to use it to validate passwords (or at best, maybe do this in `_config.php` and call `Member::set_password_validator()` or a similar method to set it globally?)

What I am doing right now is making use of the YAML config based properties as so:

```
SilverStripe\Security\PasswordValidator:
  min_length: 8
  min_test_score: 3
```

Which means these settings are applied to all instances of `PasswordValidator` across my system.

I presume this is what these `@config` commented properties are for? I can also see that the public getter methods seem to make use of them, e.g. 
```
    public function getMinTestScore()
    {
        if ($this->minScore !== null) {
            return $this->minScore;
        }
        return $this->config()->get('min_test_score');
    }
```

I agree that your way would work (as per the if statement above) but I think my way should also work, no? (and is less intrusive/easier).

Is there any good reason that the `protected $testNames;` variable would/should be used over the `public function getTestNames()` (or at least the value that function outputs, stored in `$testNames`) in my example/changes?

dhensby:
hmm, ok - so maybe injector is creating the validator incorrectly?

HARVS1789UK:
@dhensby - Quite possibly, unfortunately I have no idea where to even start looking in order to validate that :-(

dhensby:
OK - I'll look into tests and maybe how the validator is created from config.

Is this a bug that only affects 4.x-dev or is it 4.0+?

maxime-rainville:
@HARVS1789UK @dhensby Just had a look.

I think the current test isn't failing because it calls `characterStrength()` which indirectly calls `setTestNames()`. So when the `validate` method is called later, `$this->testNames` will not be equal to `null`.

The following test fails for me with the current logic and works with @HARVS1789UK's fix.

```php
$v = new PasswordValidator();
$v->setMinTestScore(3);

$r = $v->validate('aA1', new Member());
$this->assertTrue($r->isValid(), 'Passing enough tests');
```

tractorcow:
This whole class is a mess, let's make a point to re-look at this soon and fix it up for 4.2 at the very least. I'm not happy in particular about the disjoint and independent behaviour of getTests() / getTestNames(), and I think the poor bounds checking of min score and total number of tests is the least of our problems.

In any case, don't feel pressured to address all of these issues just yet. I would rather put off merging this issue and address it properly in a separate PR.

dhensby:
> I think the current test isn't failing because it calls characterStrength() which indirectly calls setTestNames(). So when the validate method is called later, $this->testNames will not be equal to null.

That's right, yep :)

dhensby:
@tractorcow so you want to leave the bug as is without a patch in the meantime? I think it'd be good to get this in myself...

HARVS1789UK:
I am currently using my own GitHub hosted fork of silverstripe/framework,
with my patch applied, in order for the project I am working on to go live.
So I would agree it would be good to get a fix in, as I won't be able to
upgrade at all until that is done.

Kind regards,

[image: Quadra]   Joe Harvey
Lead Web Developer

t: +44 (0) 1534 491096
e:  <+441534491096>joe.harvey@quadradigital.co.uk
w: quadradigital.co.uk

   - [image: Follow us on Twitter] <https://twitter.com/Quadra_Digital>
   - [image: Find us on LinkedIn] <http://uk.linkedin.com/in/joeharvey2>



Quadra, 22 Hill Street, St. Helier, Jersey, JE2 4UAAn E-scape company

DISCLAIMER: This e-mail is confidential and may contain legally privileged
information. If you are not named above as an addressee it may be unlawful
for you to read, copy, distribute, disclose or otherwise use the
information in this e-mail. Any views or opinions presented are solely
those of the author, and do not necessarily represent those of Quadra
Digital. If you are not the intended recipient and have received this
e-mail in error please notify Quadra Digital on +44 (0) 1534 491 096
<+441534491096>.

Quadra and Quadra Digital are the trading names of Quadrahedron Limited.

dhensby:
> There's still bugs in the code even with this fix is my concern.

There are still bugs without it too, just less of them. We need to do that "is this an improvement" test rather than "is it perfect" for this PR IMO

tractorcow:
Check https://github.com/silverstripe/silverstripe-framework/pull/8013 which has another fix plus code quality cleanup.

tractorcow:
Replaced with https://github.com/silverstripe/silverstripe-framework/pull/8013

Issue #7791, opened 2018-01-23
Prepare for GDPR compliance


andrewandante:
On 25 May 2018, the [EU General Data Protection Regulation (GDPR)](https://www.eugdpr.org/) will come into enforcement. SilverStripe should decide to what extent it should change the core code to accommodate this change, to what extent it can provide advice about best practices, and to what extent it is left up to the individual site owners.

In addition, best practices can be established for what data is stored by default, what is done with that data, and how it is stored, processed, and disclosed.

At minimum a blogpost or docs update that addressed potential areas to look at or config settings to tweak would be a great bout of confidence for the community.

DorsetDigital:
As a starting point, it's probably worth considering what the core system actually stores that falls under the regulation.   From what I can see, the following are the only issues:

Member table: Name, Email, etc.
LoginAttempt table:  IP, Email

As I understand it, the critical things we need for GDPR are statements about the use of the data, and means to expire it, etc.

For the second of the two, the simplest way to deal with it would be to take it out of the equation altogether.  If it were possible to disable the loginattempt logging with a configuration option, that solves the issue.   If the loginattempt data is needed, then some mechanism to expire/purge it periodically (maybe an add-on) would help.

dhensby:
a good place to start: https://ico.org.uk/for-organisations/guide-to-the-general-data-protection-regulation-gdpr

chillu:
I agree that GDPR compliance itself is critical (we've started reviewing our own systems like SilverStripe Platform and silverstripe.com/org back in November). But in the context of our open source software, my understaning is that the software itself is neither a "Data Processor" nor a "Data Controller". It really depends on what you build with it. I agree that some documentation about potential privacy impacts emerging from our software would be nice. But unless somebody thinks we need to change our APIs or system behaviour, it's not `impact/critical`. I'll raise a GDPR blog post with our marketing team! :)

chillu:
[...]

PR #8003, opened 2018-04-13
Docs for personal data usage in core

chillu:
See https://github.com/silverstripe/silverstripe-framework/issues/7791

### DIFF
--- a/docs/en/02_Developer_Guides/09_Security/06_Personal_Data.md
+++ b/docs/en/02_Developer_Guides/09_Security/06_Personal_Data.md
@@ -1,0 +1,66 @@
+title: Personal Data
+summary: How the SilverStripe CMS deals with data privacy
+
+# Personal Data
+
+SilverStripe is an application framework which can be used to process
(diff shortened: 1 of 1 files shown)


chillu:
> I think making it more clear that login attempts are recorded if max_login_attempts is configured would be good.

I don't think that's the case, check `MemberAuthenticator->recordLoginAttempt()` which is the only place in core that creates a `LoginAttempt` record: https://github.com/silverstripe/silverstripe-framework/blob/4/src/Security/MemberAuthenticator/MemberAuthenticator.php#L167. It returns when `login_recording` isn't enabled.

chillu:
@dhensby Happy to merge? Or do you disagree with my last comment?

dhensby:
> @dhensby Happy to merge? Or do you disagree with my last comment?

Let's give this a few days ;)

chillu:
OK, given this a few days ;)

chillu:
Followed up with a correction at https://github.com/silverstripe/silverstripe-framework/pull/8079

PR #8008, opened 2018-04-16
Add .htaccess to base with public rewrite

blueo:
Upgrading from 4.0 to 4.1 we needed this extra file. It is created by default with the cms recipe eg: https://raw.githubusercontent.com/silverstripe/recipe-cms/1/.htaccess


tractorcow:
The important purpose of that file is not to simply enable base as root, it's to actually BLOCK requests to any root files outside of public. Recommending this increases the overall security.

tractorcow:
@chillu you are allowed to make pull requests too. :)

PR #8009, opened 2018-04-16
FIX: Duplicating many_many relationships looses the extra fields (fixes #7973)

### LINKED ISSUE #7973: Duplicating many_many relationships looses the extra fields in 3.6
## Affected Version

SilverStripe 3.6.5

## Description

When you duplicate a DataObject with a many_many relationship with extra fields defined for the relationship extra fields are lost due to how ``DataObject::duplicateManyManyRelations()`` and ``DataObject::duplicateRelations()`` function. In short when the new items are added to the many_many relationship the extra fields are not included or even retrieved when [DataObject::duplicateRelations()](https://github.com/silverstripe/silverstripe-framework/blob/3.6.5/model/DataObject.php#L589-L594) adds the item to the list. This issue also seems to affect 4.0 but appears to have been addressed in 4.1.

## Steps to Reproduce

1. Create a DataObject with a many_many relationship and many_many_extraFields for that relationship to another object
2. Add a few items to that relationship with the extra fields populated
3. Call duplicate on your original object which duplicates the relationship's items
4. The duplicated items in the ManyManyList will have lost all of their extra field values

UndefinedOffset:
Fixes issue when you duplicate a DataObject with a many_many relationship with extra fields defined for the relationship extra fields are lost due to how DataObject::duplicateManyManyRelations() and DataObject::duplicateRelations() function. In short when the new items are added to the many_many relationship the extra fields are not included or even retrieved when DataObject::duplicateRelations() adds the item to the list. This issue also seems to affect 4.0 but the fix required would be slightly different see #7973 for details.

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -585,7 +585,21 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 	private function duplicateRelations($sourceObject, $destinationObject, $name) {
 		$relations = $sourceObject->$name();
 		if ($relations) {
-			if ($relations instanceOf RelationList) {   //many-to-something relation
+            if ($relations instanceOf ManyManyList) { //many-to-many relation
+                $source = $sourceObject->getManyManyComponents($name);
+                $extraFieldNames = $source->getExtraFields();
+
+                if ($relations->Count() > 0) {  //with more than one thing it is related to
+					foreach($relations as $relation) {
+                        // Merge extra fields
+                        $extraFields = array();
+                        foreach ($extraFieldNames as $fieldName => $fieldType) {
+                            $extraFields[$fieldName] = $relation->getField($fieldName);
+                        }
+                        $destinationObject->$name()->add($relation, $extraFields);
+                    }
+                }
+            } else if ($relations instanceOf RelationList) {   //many-to-something relation
 				if ($relations->Count() > 0) {  //with more than one thing it is related to
 					foreach($relations as $relation) {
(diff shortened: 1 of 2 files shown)


dhensby:
This needs to go into 3, I'm afraid

dhensby:
If you can find a way of patching `duplicateRelations` without changing the API surface, then it could make a patch release.

dhensby:
I reckon you can add a new condition to `duplicateRelations` which was:

`if ($relations instanceof ManyManyList)` and then use your logic to populate extra data (line 589)

UndefinedOffset:
Ya that would make sense, let me give it a shot and I'll update in a bit.

UndefinedOffset:
@dhensby There hows that look? Also should we back port this to 3.5 too or just for 3.6 and another solution for 4.0?

UndefinedOffset:
Not sure why the PGSQL test failed... it should have worked my local copy with postgres 9.3 and SilverStripe 3.6.5 :S passed

kinglozzer:
@UndefinedOffset Try adding a sort, I seem to remember PostgreSQL having issues with inconsistent result ordering if you don’t provide one...

dhensby:
It looks good to me, let's rebase onto 3.5 :)

UndefinedOffset:
Will do and I have a branch for 4.0 I'll open a pull for that too https://github.com/webbuilders-group/silverstripe-framework/tree/duplicate-many-many-fix-4-0

PR #8013, opened 2018-04-17
BUG Prevent password validator min score producing false negatives

tractorcow:
Replaces #7995

Fixes min score failing if no tests are set
Fixes min score failing if there are no failing tests
Adds unit tests to cover the above
Fix up bad phpdoc, code style, linting, and upgrade deprecated API usage

### DIFF
--- a/src/Security/PasswordValidator.php
+++ b/src/Security/PasswordValidator.php
@@ -39,39 +39,39 @@ class PasswordValidator
 
     /**
      * @config
-     * @var integer
+     * @var int
      */
     private static $min_length = null;
 
     /**
      * @config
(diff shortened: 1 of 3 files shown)


dhensby:
OK - I assume this is a bug in 4.0 as well as 4.1, can we target 4.0 whilst it's still a supported version?

tractorcow:
Actually 4.0 is safe; We'd refactored PasswordValidator in 4.1, introducing the bug there.

chillu:
@dhensby Are you                        happy to merge this?

dhensby:
I am, but does it fully address the initial bug?

PR #8016, opened 2018-04-18
FIX: Duplicating many_many relationships looses the extra fields in 4.0

UndefinedOffset:
Same fix as #8009 but applied to SilverStripe 4.0, addresses the issue raised in #7973.


src/ORM/DataObject.php
```diff
@@ -459,8 +459,20 @@ protected function duplicateManyManyRelation($sourceObject, $destinationObject,
         // Copy all components from source to destination
         $source = $sourceObject->getManyManyComponents($manyManyName);
         $dest = $destinationObject->getManyManyComponents($manyManyName);
+
+        if ($source instanceof UnsavedRelationList) {
```
dhensby: I'd consider swapping this round so 

```php
if ($source instanceof ManyManyList) {
    $extraFieldNames = $source->getExtraFields();
} else { 
    $source = []; 
}
```

That way we are using the API we know exists on `ManyManyList` rather than *not* using it on  one we know doesn't...

UndefinedOffset: good point... commit incoming

PR #8018, opened 2018-04-19
BUG Fix incorrect installer base tag generation

### LINKED ISSUE #7926: 4.1 installer doesn't load styles by default
The 4.1 installer comes with a `public/` folder out of the box (which is great).
If you point your webserver to the base folder rather than the `public/` subfolder,
a `.htaccess` file redirects requests automatically (which is also great).
But: It sets `<base href="//">`, which breaks relative CSS loading at least in Chrome.

We should condense this to a single slash, but there's a lot of code in `constants.php` to determine the `BASE_URL` constant, so I'm a bit unsure on where to do this.

There's four variations which need testing:

 * `http://localhost/install.php`
 * `http://localhost/public/install.php`
 * `http://localhost/my-website/install.php`
 * `http://localhost/my-website/public/install.php`

Marking this as `impact/high` since it'll determine the default experience for anybody downloading SilverStripe once we release 4.1 (release is imminent). I don't think this holds up 4.1.0, but we should roll it into a 4.1.1 very soon.

# PRs:

 - [x] https://github.com/silverstripe/silverstripe-framework/pull/8018

tractorcow:
Fixes #7926

I didn't notice the `/` hard coded into the template.

Copies the logic from Director::baseURL() to strip and re-add the trailing `/`

### DIFF
--- a/src/Dev/Install/config-form.html
+++ b/src/Dev/Install/config-form.html
@@ -5,7 +5,7 @@
 <!--[if gt IE 8]><!--> <html class="no-js" lang="en"> <!--<![endif]-->
 	<head>
 		<title>SilverStripe CMS / Framework Installation</title>
-        <base href="<?php echo htmlentities($base); ?>/" />
+        <base href="<?php echo htmlentities($base); ?>" />
 		<meta http-equiv="Content-type" content="text/html; charset=utf-8">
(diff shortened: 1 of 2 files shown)


unclecheese:
Tested and confirms it resolves the issue.

PR #8019, opened 2018-04-19
ENHANCEMENT Only run coverage test as a cron

### LINKED ISSUE #8014: RFC Split out some of .travis.yml to make builds faster
The build matrix for framework is quite complex, and some of the tests being run do bottleneck the overall progress. "time to green" for individual builds can be up to an hour sometimes. For example:

image

In particular adding coverage adds at least 10 minutes to the overall build time. We don't need coverage on every PR (especially since it's pretty much ignored) but we do use this information.

Another issue is that framework tests running on our travis-ci.org do sometimes block other modules from being built. Especially when we are doing releases, builds can be backed up for quite a long time.

I have looked into possible options:

 - Switching framework entirely to CircleCI. This would prevent other travis / circleci tests from blocking each other, but it is a lot of work to migrate `.travis.yml` to circleci. This also involves choosing, upgrading, and maintaining a docker image, unless we can find a good standard third-party maintained one. Things like behat would need to be re-configured.
 - Switching only low-difficulty tests (non behat, possibly a single matrix with coverage enabled) to circleci. This would probably be a good medium between the above, and doing nothing at all, but we do have the added cost of splitting our maintenance.
 - Remain on travis, but switch sets of tests to daily cron job only. We could filter these jobs (e.g. coverage) to filter based on the `[[ $TRAVIS_EVENT_TYPE = cron ]]` condition. This would mean that certain features (e.g. coverage) wouldn't be available on pull requests anymore, but would be available on request from codecov.
 
I have also investigated paid travis, but it would appear that travis.com is only available for private repos.
 
# PRs:

- [x] https://github.com/silverstripe/silverstripe-framework/pull/8019

tractorcow:
Fixes #8014 

See docs at https://docs.travis-ci.com/user/conditional-builds-stages-jobs/#stq=&stp=0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ matrix:
         - DB=PGSQL
         - PHPUNIT_TEST=framework
     - php: 7.1
+      if: type IN (cron)
       env:
         - DB=MYSQL
         - PDO=1


tractorcow:
I've also setup a daily cron job on 4 branch

PR #8021, opened 2018-04-19
Add reference to the reorganise command

maxime-rainville:
# Related issue
https://github.com/silverstripe/silverstripe-upgrader/issues/60

# Waiting for
Wait for https://github.com/silverstripe/silverstripe-upgrader/pull/64 to be merge.

### DIFF
--- a/docs/en/03_Upgrading/index.md
+++ b/docs/en/03_Upgrading/index.md
@@ -114,6 +114,16 @@ cd ~/my-project-root
 ~/.composer/vendor/bin/upgrade-code upgrade ./app/src --write
 ```
 
+### Switching to new `app/src` structure
+
+The `reorganise` command can automatically rename the `mysite` and `code` folder for you. It will search your code and find any occurence of `mysite`. It won't replace those occurence with `app` however.
+
+After installing, run the upgrader reorganise command:
+```
+cd ~/my-project-root
+~/.composer/vendor/bin/upgrade-code reorganise --write
+```
+
 ## Migrating files
 
 Since the structure of `File` dataobjects has changed, a new task `MigrateFileTask`
--- a/docs/en/04_Changelogs/4.2.0.md
+++ b/docs/en/04_Changelogs/4.2.0.md
@@ -135,10 +135,16 @@ Code which relies on raw SQL queries to these tables will need to be updated.
 
 `SiteTreeFileExtension` is deprecated, and has it's functionality baked directly into `File` dataobject.
 
-### New upgrader command
-The SilverStripe upgrader tool can now be used to convert your `_ss_environment.php` file to an equivalent `.env` file when migrating a SilverStripe 3 project to SilverStripe 4.
+### New upgrader commands
+
+Two new commands have been added to the SilverStripe upgrader tool: `environment` and `reorganise`.
+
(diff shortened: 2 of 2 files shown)


docs/en/03_Upgrading/index.md
```diff
@@ -114,6 +114,16 @@ cd ~/my-project-root
...
 ```
 
+### Switching to new `app/src` structure
+
+The `reorganise` command can automatically rename the `mysite` and `code` folder for you. It will search your code and find any occurence of `mysite`. It won't replace those occurence with `app` however.
+
+After installing, run the upgrader reorganise command:
+```
+cd ~/my-project-root
+~/.composer/vendor/bin/upgrade-code reorganise --write
```
tractorcow: We should update our instructions to setup `upgrade-code` as a global command. Running the full path each time isn't very user friendly right? My preference is adding `~/.compsoer/vendor/bin` to $PATH env.

maxime-rainville: The point is taken, but that's more a side affect of the upgrading doc needing rewriting.

I'll merge it like that for now and discuss redrafting the upgrade doc with @chillu.

dhensby:
Hey @maxime-rainville, please can you avoid merging PRs that have changes requested, especially by a core committer. Unless you can get a general consensus from others that it's ok to dismiss the request you really shouldn't be ignoring (and self merging) your PRs without a consensus that this is OK.

chillu:
It looks like Maxime has addressed Damian's concern, which was somewhat separate from the pull request topic - the absolute composer path is used elsewhere in the docs already.

Regarding self-merges, we need to strike a balance between velocity, process and quality. The team doing most core work (Open Sourcerers) used to have two core committers (Aaron and Damian) and two contributors (Luke and Maxime). Damian is on leave for four weeks, and is not nearly as available after this. I'm not available for peer review most days either, and it doesn't sound like you want to be in a position to become the blocker for all of our team throughput Dan, right? ;) Robbie and the Creative Commoners are in the same position.

How about these rules for the two teams:
 * Complex or impactful changes need to be reviewed by one or more core committers. This includes any additions, removals or changes to commonly used APIs. If that's not possible in the team, ping `@silverstripe/core-team` to get other core committers involved.
 * More straightforward changes (documentation, styling) or areas which require quite specialised expertise (React) that's not available through most core committers can be merged by team members who aren't core committers
 * Self-merges should be avoided, but are preferable to having work go stale or forcing other team members to waste time by context switching into a complex review (e.g. because the original reviewer went on leave). Any self-merge should be accompanied by a comment why this couldn't be handled in another way, and a (preferably written) approval from another team member.

flamerohr:
I think the main point I'd like to bring up is to please give a sensible lead time so a chance is given to a core committer to be involved.
A link to the discussion if in a separate thread or an explanation is helpful with any outcome changes and dismissals.

I'm happy to help *when available* with regards to React and if Robbie and Aaron aren't able to.

maxime-rainville:
@dhensby Sorry ... I jumped the gun a bit here. I was going off the idea this was a simple doc change and that I had an offline discussion about this with @chillu. I'll be more careful next time.

dhensby:
thanks team

chillu:
I've captured those rules (in a modified way) for the open source docs, please continue discussion there :) https://github.com/silverstripe/silverstripe-framework/pull/8054

Issue #8022, opened 2018-04-19
LookupField unable to set value from DataObject


mfendeksilverstripe:
## Affected Version

`4.1.x-dev`, commit hash `5a8c5601a992fbc8b1531f8e0172940a30d48921`

## Description

This issue manifests as follows:

In some forms the `DropdownField` can be changed to `LookupField` via `performReadOnlyTransformation()`. These fields may be missing values.

The source of the problem is how the value is handled in the `LookupField`. First, note that `LookupField` extends `MultiSelectField`, not `SingleSelectField` for some reason.

`LookupField` inherits the following methods from `MultiSelectField`:

* `MultiSelectField::setValue()`
* `MultiSelectField::loadFrom()`

The issue is in the case were `MultiSelectField::setValue()` is called with a `DataObject` as a second param. This is done when `Form` is trying to populate all fields with data from `DataObject` via the `Form::loadDataFrom()` method.

In this case the `MultiSelectField::setValue()` method will call `MultiSelectField::loadFrom()` which will handle the value processing. We are interested in the case where a field present on the `DataObject` is providing the value. Within the `MultiSelectField::loadFrom()` function the following line of code gets executed:

`$value = $this->stringDecode($record->$fieldName);`

This internally does this:

`$result = json_decode($value, true);`

It clearly expects a composite value (valid `JSON`) but we are only providing a plain (atomic value) since this `DataObject` field was originally mapped to a `DropdownField`. `json_decode()` will produce a `null` value which will be set into the field resulting in the loss of original value.

## Steps to Reproduce

This is my testing setup:

I used `SilverStripe\Assets\Image` as `DataObject`. `Image` has an extra field `OwnerType` which is a `Enum`. The Field config looks like this:

```
/**
 * @var array
 */
private static $db = [
    'OwnerType' => 'Enum(array("TNZ", "OD", "My Uploads"), "My Uploads")',
];
``` 

Here is the code that reproduces this issue:

* set up a `DropdownField` that is mapped to the `OwnerType` DB field
* transform the field into a `LookupField` via `performReadonlyTransformation()`
* have an `Image` object ready that has the `OwnerType` value set to `'OD'` (value is arbitrary it just needs to be one of the enum values)
* first `var_dump()` will output `string(2) "OD"`
* second  `var_dump()` will output `NULL`

The desired outcome is that the second  `var_dump()` should output `string(2) "OD"`

```
/** @var LookupField $testField */
$testField = DropdownField::create(
[...]

Issue #8002, opened 2018-04-13
RFC Stop supporting multiple _Disabled / _Readonly classes


tractorcow:
We currently support both setDisabled(true) / setReadonly(true) as well as separate class implementations for these views. This promotes inconsistent (and non-accessible) views for form fields.

I suggest to ensure that setting one (or both properties) on any form field are properly respected, and that we remove these redundant classes.

maxime-rainville:
Off the top of my head, I would agree with @tractorcow.

However the HTML specs provides for [both readonly and disabled attributes](https://stackoverflow.com/questions/7730695/whats-the-difference-between-disabled-disabled-and-readonly-readonly-for-ht).

The main difference being that a readonly field's value will be submitted along with the rest of a form, while a disabled field's value will not.

In practice, there's no that many situations where that difference should matter. The only one I could think off is where you would like to display a value that would normally be tracked through an hidden field. e.g.: Let's say you wanted to display the ID of the record the user is editing.

tractorcow:
> However the HTML specs provides for both readonly and disabled attributes.

That's what I'm advocating for, rather than separate classes / templates to provide a different view. We should only need the html properties, not a whole new set of classes.

maxime-rainville:
@tractorcow Cool. Sorry, I read it the wrong way around.

tractorcow:
I just took it that you were agreeing with me. :)

maxime-rainville:
Let it be resolved that the RFC has been accepted.

sminnee:
Suits me. My only comment would be to deprecate, rather than delete, in 4.x for BC.

robbieaverill:
Ok sounds good. Deprecate in 4.x and start using the correct attributes from the base FormField class to handle this behaviour - I guess we still need to use the templates from the deprecated classes until 5.x though for B/C

PR #8023, opened 2018-04-19
LookupField value handling corrected

mfendeksilverstripe:
`LookupField` no longer throws away an atomic value.

Solution for https://github.com/silverstripe/silverstripe-framework/issues/8022

### DIFF
--- a/src/Forms/SingleLookupField.php
+++ b/src/Forms/SingleLookupField.php
@@ -1,0 +1,134 @@
+<?php
+
+namespace SilverStripe\Forms;
+
+use SilverStripe\Core\Convert;
+use SilverStripe\ORM\DataObjectInterface;
+use SilverStripe\ORM\Map;
+use SilverStripe\ORM\FieldType\DBField;
+
+/**
+ * Read-only complement of {@link DropdownField}.
+ *
+ * Shows the "human value" of the dropdown field for the currently selected
+ * value.
+ */
+class SingleLookupField extends SingleSelectField
+{
+    /**
+     * @var bool
+     */
+    protected $readonly = true;
+
+    /**
+     * @return mixed|null
+     */
+    protected function valueToLabel()
+    {
+        $value = $this->value;
+        $source = $this->getSource();
+        $source = ($source instanceof Map) ? $source->toArray() : $source;
+
+        if (array_key_exists($value, $source)) {
+            return $source[$value];
+        }
+
+        return null;
+    }
+
+    /**
+     * Ignore validation as the field is readonly
+     *
+     * @param Validator $validator
+     * @return bool
+     */
+    public function validate($validator)
+    {
+        return true;
+    }
+
+    /**
+     * Stubbed so invalid data doesn't save into the DB
+     *
+     * @param DataObjectInterface $record DataObject to save data into
+     */
+    public function saveInto(DataObjectInterface $record)
+    {
+    }
+
+    /**
+     * @return SingleLookupField
+     */
+    public function performReadonlyTransformation()
+    {
+        $clone = clone $this;
+
+        return $clone;
+    }
+
+    /**
+     * @return bool
+     */
+    public function getHasEmptyDefault()
+    {
+        return false;
+    }
+
+    /**
+     * @return string
+     */
+    public function Type()
+    {
+        return 'single-lookup readonly';
+    }
+
+    /**
+     * Note: we need to transform value in here becaue React fields do not use Field() to display value
+     *
+     * @return mixed
+     */
+    public function Value()
+    {
+        $label = $this->valueToLabel();
+        if (!is_null($label)) {
+            return $label;
+        }
+
+        return $value;
+    }
+
+    /**
+     * @return string
+     */
+    public function getTemplate()
+    {
+        // this field uses the same default template as LookupField
+        return parent::getTemplate() ?: LookupField::class;
+    }
+
+    /**
+     * Returns a readonly span containing the correct value.
+     *
+     * @param array $properties
+     *
+     * @return string
+     */
+    public function Field($properties = [])
+    {
+        $label = $this->valueToLabel();
+        if (!is_null($label)) {
+            $attrValue = Convert::raw2xml($label);
+            $inputValue = $this->value;
+        } else {
+            $attrValue = '<i>(' . _t('SilverStripe\\Forms\\FormField.NONE', 'none') . ')</i>';
+            $inputValue = '';
+        }
+
+        $properties = array_merge($properties, array(
+            'AttrValue' => DBField::create_field('HTMLFragment', $attrValue),
+            'InputValue' => $inputValue
+        ));
+
+        return parent::Field($properties);
+    }
+}
--- a/src/Forms/LookupField.php
+++ b/src/Forms/LookupField.php
@@ -8,9 +8,9 @@ use SilverStripe\ORM\DataObjectInterface;
 use SilverStripe\ORM\FieldType\DBField;
 
 /**
- * Read-only complement of {@link DropdownField}.
+ * Read-only complement of {@link MultiSelectField}.
  *
- * Shows the "human value" of the dropdown field for the currently selected
+ * Shows the "human value" of the MultiSelectField for the currently selected
  * value.
  */
 class LookupField extends MultiSelectField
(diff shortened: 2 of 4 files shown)


robbieaverill:
Also, if you'd like this in 4.x could you please rebase and change this PR's target branch?

mfendeksilverstripe:
@robbieaverill I rebased the branch on `4.1` and added a unit test. Please review.

robbieaverill:
Interesting catch. It looks almost to me like like `MultiSelectField::stringDecode` is expecting to throw an exception in a situation like this:

```php
        // If json deserialisation fails, then fallover to legacy format
        $result = json_decode($value, true);
        if ($result !== false) {
            return $result;
        }

        throw new \InvalidArgumentException("Invalid string encoded value for multi select field");
```

Given that `json_decode` returns null when deserialisation fails, this logic seems like it would never execute (the exception thrown).

I think this fix is fine, and looks like a safe change. I also feel like we need to fix `MultiSelectField::stringDecode` to correctly detect JSON deserialisation failures (I presume there's no unit test for this), but that's another story.

You're also right in that it seems odd for DropdownField > readonly > LookupField to extend from MultiSelectField. I guess it's probably the same case as a field that allows multiple selections going through readonly transformation extending from LookupField, so perhaps is used to handle both cases. In the case of DropdownField > readonly > LookupField, throwing an exception on failure to deserialise a JSON input value is wrong because we don't expect it.

The logic in this PR is good for that case, assuming that the parent logic is currently wrong and not detecting serialisation failures. I suspect if we "fix" `MultiSelectField::stringDecode` then this logic would no longer work, as an exception would be thrown early. This could be a bit risky in terms of future maintenance.

I also notice a disparity between `stringEncode` (public) and `stringDecode` (protected).

---

I think you've highlighted a valid problem (and fixed it) but I'm a little concerned with fixing it in this way with respect to the longevity of the change.

What do you think @tractorcow?

tractorcow:
The code in MultiSelectField is incorrectly checking for `false` as error condition, when `null` is actually the error condition.

I think you are right in that sometimes the object's field doesn't contain proper json content, but actually is the selected value to show.

How about this


```php
protected function stringDecode($value)
    {
        // Handle empty case
        if (empty($value)) {
            return [];
        }

        // Attempt to decode json format
        $result = json_decode($value, true);

        // If non-json value, assume single value
        if (json_last_error()) {
            return [ $value ];
        }

        return $result;
    }
```

Notice we retain the array value return result. ;)

Fix this in MultiSelectField, not just in a subclass.

tractorcow:
I also think dropdownfield readonly should just be setReadonly(true), not shifting to a separate form field class. See https://github.com/silverstripe/silverstripe-framework/issues/8002

mfendeksilverstripe:
@robbieaverill @tractorcow I implemented Damian's solution and it seems to be working fine. Please review.

tests/php/Forms/LookupFieldTest.php
```diff
@@ -124,4 +126,21 @@ public function testWithMultiDimensionalSource()
             trim($f->Field()->getValue())
         );
     }
+
+    public function testAtomicValueHandlingFromDataObject()
+    {
+        $member = $this->objFromFixture(Member::class, 'member1');
+
+        /** @var LookupField $testField */
```
tractorcow: Add an `$this->assertInstanceOf` as well

mfendeksilverstripe: Donzo.

mfendeksilverstripe: Donzo

src/Forms/MultiSelectField.php
```diff
@@ -160,13 +160,15 @@ protected function stringDecode($value)
...
 
-        // If json deserialisation fails, then fallover to legacy format
+        // Attempt to decode json format
         $result = json_decode($value, true);
-        if ($result !== false) {
-            return $result;
+
+        // If non-json value, assume single value
+        if (json_last_error()) {
+            return [$value];
```
robbieaverill: Why do we return an array here, rathe than the input value?

tractorcow:
> Hmmm... now the behaviour is to treat json deserialisation failures as atomic values rather than throw an exception. This would mean broken json strings are returned literally instead of throwing errors. I suspect this could end up causing data to be displayed to CMS users rather than a system error message


I think the REAL fix here is to stop using LookupField as a readonly dropdown? :P

robbieaverill:
Yeah totally, but we can't in 4.x, that's be an API change (`DropdownField::performReadonlyTransformation` no longer returns LookupField). This is a good intermediary until we can simplify the readonly/disabled form field methodology as discussed the other day!

mfendeksilverstripe:
@robbieaverill @tractorcow I agree that we probably don't want to change existing functionality of the `MultiSelectField`. I added a new `SingleLookupField` to represent the single value version of `LookupField`. Please review.

tractorcow:
I'm not sure this is better. You have a singleselect extends somefield extends multiselect. Shouldn't it extend SingleSelectField instead?

tractorcow:
This is the API of performReadonlyTransformation()

```php
/**
     * Returns a read-only version of this field.
     *
     * @return FormField
     */
    public function performReadonlyTransformation()
```

It doesn't require a specific subclass of FormField to be returned.

mfendeksilverstripe:
@tractorcow The reason why I decided to extend the `LookupField`  is that I can re-use functionality already present in that field. It seems to be more efficient than extending the `SingleSelectField `.

mfendeksilverstripe:
Also, it's more backward compatible because if some code checks for `instanceof LookupField`, my change will not break such code.

tractorcow:
> The reason why I decided to extend the LookupField is that I can re-use functionality already present in that field.

I understand that, but it's a mis-use of object inheritance. Inheritance isn't about code saving. There is much more appropriate logic to single-selection of fields in SingleSelectField that should be applied anyway.

mfendeksilverstripe:
@tractorcow Ok, I'll update the code, thanks for the feedback.

mfendeksilverstripe:
@robbieaverill @tractorcow New changes added, please review.

src/Forms/SelectField.php
```diff
@@ -241,7 +241,7 @@ public function isSelectedValue($dataValue, $userValue)
     public function performReadonlyTransformation()
     {
         /** @var LookupField $field */
-        $field = $this->castedCopy('SilverStripe\\Forms\\LookupField');
```
tractorcow: I'm going to be mean and state that you should have separate performReadonlyTransformation in both SingleSelectField and MultiSelectField (both direct subclasses of SelectField). You've switched the assumption from "all select fields use the multi select readonly" to "all select fields use the single select readonly".

src/Forms/SingleLookupField.php
```diff
@@ -0,0 +1,127 @@
...
+     * Returns a readonly span containing the correct value.
+     *
+     * @param array $properties
+     *
+     * @return string
+     */
+    public function Field($properties = [])
+    {
+        // this field uses the same template as LookupField
+        $this->setTemplate(LookupField::class);
```
tractorcow: as `Field()` is a renderer, try to make it as side-effect free as possible. Getters that modify object state are bug vectors, to be blunt. :)

try this instead.

```
public function getTemplate()
{
	return parent::getTemplate() ?: LookupField::class;
}
```

src/Forms/SingleLookupField.php
```diff
@@ -0,0 +1,127 @@
...
+
+        $label = $this->valueToLabel();
+        if (!is_null($label)) {
+            $attrValue = $label;
```
tractorcow: $attrValue is being cast as HTMLFragment below, but it's not secured up here. If it contains HTML it'll get raw-output.

robbieaverill:
Sorry, I dismissed the wrong review! I'll leave this to @tractorcow =)

tractorcow:
Implementation is ok with me, but @robbieaverill do you think we should merge to 4 branch not 4.1? It's a new class, seems like a new API rather than a pure bugfix.

robbieaverill:
I'm easy. Semver allows for bugfixes that are backwards compatible (this mostly is) as a patch

tractorcow:
In that case I'm making the call to make it a minor branch change.

tractorcow:
Bughancement

PR #8027, opened 2018-04-23
Add documentation on relating to core classes

blueskies79:
Adding a (has_one/has_many) relationship to core classes brings some extra syntax issues. I think it would be useful to note that core classes are related through Image::class (different syntax than 'MyDataObject') and that a has-one-relationship needed on a core class when relating them thru has_many can be set up through yml.

### DIFF
--- a/docs/en/02_Developer_Guides/00_Model/02_Relations.md
+++ b/docs/en/02_Developer_Guides/00_Model/02_Relations.md
@@ -39,6 +39,15 @@ class Player extends DataObject
 This defines a relationship called `Team` which links to a `Team` class. The `ORM` handles navigating the relationship
 and provides a short syntax for accessing the related object.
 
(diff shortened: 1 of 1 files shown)


docs/en/02_Developer_Guides/00_Model/02_Relations.md
```diff
@@ -39,6 +39,15 @@ class Player extends DataObject
 This defines a relationship called `Team` which links to a `Team` class. The `ORM` handles navigating the relationship
 and provides a short syntax for accessing the related object.
 
+To create a has_one/has_many relationship to core classes (File, Image, etc), reference the Classname::class, like below.
+
+```php
+    private static $has_many = [
```
dhensby: In our docs we usually add the import statement at the top of the example block here.

So please add the FQCN for `Image` and `File` in `use` statements before this line

PR #8028, opened 2018-04-23
unset http scheme on CLIRequestBuilder

andrewandante:
Encountered [a bug](https://github.com/silverstripe/silverstripe-staticpublishqueue/issues/77) when running StaticPublishQueue with `SS_BASE_URL` set to an https address.

When the full static cache job is run via CLI, the scheme will always be set to HTTP. This is because we [generate a dummy request](https://github.com/silverstripe/silverstripe-framework/blob/4.0/src/Control/Director.php#L969-L976), which when constructed [sets the scheme to HTTP by default](https://github.com/silverstripe/silverstripe-framework/blob/4.0/src/Control/HTTPRequest.php#L164).

Then when building the URL in the cache collection, when we use `Director::absoluteURL()`, `Director::is_https()` will always return `false` unless we set `Director::alternate_base_url()`, as the current request [takes priority](https://github.com/silverstripe/silverstripe-framework/blob/4.0/src/Control/Director.php#L533-L560) over `Director::default_base_url()`

In practice, I expect the CLI runner to return the protocol it has information for. In this case, it is blindly assuming that the dummy request has accurate information about the protocol the site expects. This PR unsets the set scheme to let it fall through to `SS_BASE_URL`.

An alternative idea was to check:

```php
        if ($request && $request->getScheme() === 'https') {
            return true;
        }
```

instead, to allow for deliberate overriding in `Director::is_https()`, but that means if you make an HTTP request, but the base tag is HTTPS, you end up with a pile of CORS errors.

### DIFF
--- a/src/Control/CLIRequestBuilder.php
+++ b/src/Control/CLIRequestBuilder.php
@@ -2,6 +2,8 @@
 
 namespace SilverStripe\Control;
 
+use SilverStripe\Core\Environment;
+
 /**
  * CLI specific request building logic
  */
@@ -65,5 +67,20 @@ class CLIRequestBuilder extends HTTPRequestBuilder
 
         // Parse rest of variables as standard
         return parent::cleanEnvironment($variables);
+    }
+
+    /**
+     * @param array $variables
+     * @param string $input
+     * @return HTTPRequest
+     */
+    public static function createFromVariables(array $variables, $input)
+    {
+        $request = parent::createFromVariables($variables, $input);
+        // unset scheme so that SS_BASE_URL can provide `is_https` information if required
+        $scheme = parse_url(Environment::getEnv('SS_BASE_URL'), PHP_URL_SCHEME);
+        $request->setScheme($scheme);
+
+        return $request;
     }
 }


src/Control/CLIRequestBuilder.php
```diff
@@ -66,4 +66,18 @@ public static function cleanEnvironment(array $variables)
...
+    /**
+     * @param array $variables
+     * @param string $input
+     * @return HTTPRequest
+     */
+    public static function createFromVariables(array $variables, $input)
+    {
+        $request = parent::createFromVariables($variables, $input);
+        // unset scheme so that SS_BASE_URL can provide `is_https` information if required
+        $request->setScheme(null);
```
tractorcow: How about we set the scheme from the `SS_BASE_URL` env var instead?

dhensby: that happens anyway, in a way, through director

dhensby: if we implement our own parser of `SS_BASE_URL` here rather than rely on `Director` then we're going to duplicate code. It'll also mean we still have the same problem where the `is_https` check never actually examines the `SS_BASE_URL` because the request object always has a scheme set.

tractorcow: Director doesn't setScheme(). HTTPRequest should be the source of truth and we need to stop relying on static state / helpers such as Director.

The only code duplication is `parse_url(Environment::getEnv('SS_BASE_URL'), PHP_URL_SCHEME)`

tractorcow: Director::is_https() doesn't even touch SS_BASE_URL so that's wrong to rely on too. It calls getScheme() on the request which is why you need to set it here.

tractorcow: I wouldn't say never, but it's certainly not the primary source of truth.

For cli my idea is that we shift to symfony console and have a separate application class. Cli would not have httprequests at all

andrewandante: So, just to make sure I've understood - we want to manually parse `SS_BASE_URL` for scheme here (if it's set)?

scott1702:
Can confirm this worked for me to resolve https://github.com/silverstripe/silverstripe-staticpublishqueue/issues/77, Although ^4.1 I had to include this [additional fix](https://github.com/scott1702/silverstripe-framework/commit/20ccc2ad0022b7c7d70a4895f1d2df88e21a320f) due to the changes [here](https://github.com/silverstripe/silverstripe-framework/pull/7705/commits/c5bd9bb424006a4cbce1c71806c323d9ca3328e1#diff-0db18c39f472bfa257235e8a90816756L33)

robbieaverill:
Thanks @scott1702, I've cherry picked your commit during the merge up

src/Control/CLIRequestBuilder.php
```diff
@@ -66,4 +68,19 @@ public static function cleanEnvironment(array $variables)
...
+     * @param array $variables
+     * @param string $input
+     * @return HTTPRequest
+     */
+    public static function createFromVariables(array $variables, $input)
+    {
+        $request = parent::createFromVariables($variables, $input);
+        // unset scheme so that SS_BASE_URL can provide `is_https` information if required
+        $scheme = parse_url(Environment::getEnv('SS_BASE_URL'), PHP_URL_SCHEME);
+        $request->setScheme($scheme);
```
tractorcow: This really should have an `if` around it. Not all base_urls have schema set; They could be protocol-less absolute url. (e.g. `//localhost`). Even that case notwithstanding, you should avoid setting a null scheme in case of error.

robbieaverill: Patch here: https://github.com/silverstripe/silverstripe-framework/pull/8358

PR #8034, opened 2018-04-25
Document requirements for Configurable trait

mobiusnz:


docs/en/02_Developer_Guides/04_Configuration/00_Configuration.md
```diff
@@ -60,6 +60,21 @@ Or through the `config()` object on the class.
 $config = $this->config()->get('property')';
 ```
 
+If your class does not inherit from a SilverStripe class, you must apply the [Configurable](api:SilverStripe\Core\Config\Configurable) trait in order to access the `config()` object.
```
robbieaverill: I think perhaps we should say "you may need to" rather than "you must", since not all SilverStripe classes have this trait

tractorcow: howa bout `the config() method` rather than `the config() object`?

mobiusnz:
@tractorcow I was just using the same wording from a couple lines up. 😆 I'll change both eh

PR #8037, opened 2018-04-27
Patch 2

azt3k:
Without this patch on PHP 7.2 you get the following warning if using `count` on a non-countable value:

`Warning: count(): Parameter must be an array or an object that implements Countable in /path/to/vendor/silverstripe/framework/src/Security/Member.php on line 1355`

Patch includes a couple of polyfills for helper functions introduced in newer 7.x Versions of php and a check on the `$groups` param prior to calling `count`

### DIFF
--- a/src/Security/Member.php
+++ b/src/Security/Member.php
@@ -1352,7 +1352,7 @@ class Member extends DataObject
             return ArrayList::create()->map();
         }
 
-        if (count($groups) == 0) {
+        if ($groups && count($groups) == 0) {
             $perms = array('ADMIN', 'CMS_ACCESS_AssetAdmin');
 
             if (class_exists(CMSMain::class)) {
--- a/src/Security/Member.php
+++ b/src/Security/Member.php
@@ -1352,7 +1352,7 @@ class Member extends DataObject
             return ArrayList::create()->map();
         }
 
-        if ($groups && count($groups) == 0) {
+        if (is_iterable($groups) && count($groups) == 0) {
             $perms = array('ADMIN', 'CMS_ACCESS_AssetAdmin');
 
             if (class_exists(CMSMain::class)) {
--- a/src/Security/Member.php
+++ b/src/Security/Member.php
@@ -1352,7 +1352,7 @@ class Member extends DataObject
             return ArrayList::create()->map();
         }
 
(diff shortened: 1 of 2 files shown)


src/Security/Member.php
```diff
@@ -1352,7 +1352,7 @@ public static function mapInCMSGroups($groups = null)
             return ArrayList::create()->map();
         }
 
-        if (count($groups) == 0) {
+        if (is_countable($groups) && count($groups) == 0) {
```
robbieaverill: To be honest, I'd rather just ensure that this variable is countable via logic rather than shiv in these polyfills. This is a common pattern that is widely used in the framework, so likely to occur in other places too.

I think you should just change the default argument value to an array in the function declaration. It's technically a semver break though, so for now maybe you could just add `if (!$group) { $group = []; }`, maybe even make a PR to update the method signature against the master branch for SS5 as well?

tractorcow:
@azt3k when is mapInCMSGroups called without an array? This fixes the warning, but it seems there's a bug yet remaining where a caller function is passing in the wrong value.

src/Security/Member.php
tractorcow: My feeling too. If it's not countable this should be an `InvalidArgumentException` shouldn't it?

tractorcow: Not countable or null that is. :)

azt3k: This warning is thrown with the default value of `null`

azt3k:
@tractorcow - as @robbieaverill  says the default value of `null` will trigger this.  There's a bunch of ways to fix it, but checking if the var is countable seems like the most correct way without changing the API.

tractorcow:
Fair enough :)

robbieaverill:
That’s true but also comes at the expense of polyfills

tractorcow:
If we are adding the polyfill maybe we should bump this to 4 branch.

robbieaverill:
An “if null make it empty array” check is preferable to me

robbieaverill:
We could also target this at 4.0 if it’s affecting that as well, it will get merged up into 4.1

tractorcow:
If we don't add the polyfill we can add to 4.0.

Issue #8038, opened 2018-04-27
Error thrown in Insert Link dialog


matthewntt:
## Affected Version

Framework 3.6.5

## Description

Within TinyMCE, adding a link to "page on the site" generates an error "Target page not found" under the Anchor field. The link can be saved successfully. 

## Steps to Reproduce

This issue occurs where there is a page type that extends SiteTree rather than Page. It appears that the Page selector queries all the SiteTree objects while the Anchor logic only considers the Page objects (which are a subset of SiteTree). Thus there is an inconsistency between these two fields in the same form. The simple fix is to get the Anchor logic to also reference the SiteTree objects. 

The reason I have page types extending SiteTree rather than Page is that I am updating a legacy site which already has a customised Page.php. I am developing a new template alongside the functional legacy pages.

PR #8039, opened 2018-04-27
Address issue #8038

matthewntt:
## Description
Within TinyMCE, adding a link to "page on the site" generates an error "Target page not found" under the Anchor field. The link can be saved successfully.

## Steps to Reproduce
This issue occurs where there is a page type that extends SiteTree rather than Page. It appears that the Page selector queries all the SiteTree objects while the Anchor logic only considers the Page objects (which are a subset of SiteTree). Thus there is an inconsistency between these two fields in the same form. The simple fix is to get the Anchor logic to also reference the SiteTree objects.

The reason I have page types extending SiteTree rather than Page is that I am updating a legacy site which already has a customised Page.php. I am developing a new template alongside the functional legacy pages.

## Resolution
This PR aligns the getAnchors() method to also query values from SiteTree rather than Page.


tractorcow:
Merge on green.

tractorcow:
Thanks @matthewntt for this fix it's really great. ;)

dhensby:
Is this only an issue in 3.6 or is 3.5 affected too?

dhensby:
I've noticed that you've created this pull request from one of the "core branches" (3.6) rather than a branch specifically intended for this pull request. This can cause problems if we ask you to target this PR at a different branch. If this issue affects 3.5 then your 3.6 branch will need to be rebased so that it is effectively 3.5, which may cause you problems.

dhensby:
> Is this only an issue in 3.6 or is 3.5 affected too?

OK - it looks like the code is the same in 3.5, so I'm going to rebase this and target 3.5

PR #8047, opened 2018-05-03
DOCS: Update incorrect documentation for Introduction

vincentneuts:
HTTPRequest is no longer passed in as a variable. Documentation should not have controller functions with the request variable.

### DIFF
--- a/docs/en/02_Developer_Guides/02_Controllers/01_Introduction.md
+++ b/docs/en/02_Developer_Guides/02_Controllers/01_Introduction.md
@@ -19,12 +19,12 @@ class TeamController extends Controller
         'index'
     ];
     
-    public function index(HTTPRequest $request) 
+    public function index() 
     {
         // ..
     }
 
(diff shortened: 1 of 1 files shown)


vincentneuts:
@robbieaverill following the instructions without modification on this page here: https://docs.silverstripe.org/en/4/developer_guides/controllers/introduction/ you got the following error on page load:
```
[Emergency] Uncaught TypeError: Argument 1 passed to TeamController::players() must be an instance of HTTPRequest, instance of SilverStripe\Control\HTTPRequest given, called in /var/www/vendor/silverstripe/framework/src/Control/RequestHandler.php on line 319
```

Both the index function and the players function suffer from the same issue.

Another solution would be to add the correct namespace to the top of the file:

```use SilverStripe\Control\HTTPRequest;```

robbieaverill:
Ah yes, your last suggestion is the correct solution. If you're keen to make a PR to add that to the code examples, that'd be welcomed =)

Issue #8029, opened 2018-04-23
Behat tests broken due to new modeladmin "owns" labels


unclecheese:
Now that all DataObjects have a polymorphic `has_many` files, `$owns` is universal, meaning every modeladmin, e.g. "Member" gets the versioned buttons ("Apply changes", "Publish", etc)

Relevant test failure: https://travis-ci.org/silverstripe/silverstripe-framework/jobs/369917097#L920

# PRs:

 - [x] https://github.com/silverstripe/silverstripe-versioned/pull/154
 - [X] https://github.com/silverstripe/silverstripe-admin/pull/523

robbieaverill:
I fixed this in the behat steps: https://github.com/silverstripe/silverstripe-framework/pull/8058

unclecheese:
This actually doesn't resolve the issue. I'm not sure @tractorcow 's assessment of the fix is accurate. The relevant tests failures were happening when editing, not creating. The appropriate fix, in my view, is to check that owned objects actually exist. `FileTracking` on a `Group`, for instance, is almost always empty, but just the definition of the ownership is enough to trigger the "Apply changes" button, which is a bad experience.

That said, the previous check for new records is also useful.

New PR: https://github.com/silverstripe/silverstripe-versioned/pull/154

unclecheese:
Behat tests also broken due to new graphql schema loading in silverstripe/admin that was merged without a framework behat test.

Fix is here: https://github.com/silverstripe/silverstripe-admin/pull/523

tractorcow:
Waiting until https://github.com/silverstripe/silverstripe-versioned/pull/154 is merged

PR #8058, opened 2018-05-07
FIX Update Behat test assertions for the "Save" button labels (make behat green again)

robbieaverill:
FIX Update Behat test assertions for the "Save" button labels


robbieaverill:
It seems that the test failures are actually showing a broken UI pattern, closing in favour of the solution linked in #8029

PR #8061, opened 2018-05-08
NEW: Switch default password encryptor to password_hash()

kinglozzer:
Submitted against 4 because there shouldn’t be any BC risk here: existing passwords will continue to work, `password_hash()` will be used for new members and when passwords are next changed

Updated the password column on `Member` to 255 characters just as a precaution - that’s what the [PHP manual](http://php.net/manual/en/function.password-hash.php) recommends as a safe default.

### DIFF
--- a/src/Security/PasswordEncryptor_PHPPasswordHash.php
+++ b/src/Security/PasswordEncryptor_PHPPasswordHash.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace SilverStripe\Security;
+
+use Exception;
+use SilverStripe\Core\Config\Config;
+
+/**
+ * Hashing using built-in password_hash()/password_verify() in PHP
+ */
+class PasswordEncryptor_PHPPasswordHash extends PasswordEncryptor
+{
+    /**
+     * @var int
+     * @config
+     */
+    private static $algorithm = PASSWORD_DEFAULT;
+
+    /**
+     * @var array
+     * @config
+     */
+    private static $options = [];
+
+    public function encrypt($password, $salt = null, $member = null)
+    {
+        $algorithm = Config::inst()->get(static::class, 'algorithm');
+        $options = Config::inst()->get(static::class, 'options');
+        return password_hash($password, $algorithm, $options);
+    }
+
+    public function check($hash, $password, $salt = null, $member = null)
+    {
+        return password_verify($password, $hash);
+    }
+
+    public function salt($password, $member = null)
+    {
+        return '';
+    }
+}
--- a/tests/php/Security/PasswordEncryptorTest.php
+++ b/tests/php/Security/PasswordEncryptorTest.php
@@ -165,4 +165,14 @@ class PasswordEncryptorTest extends SapphireTest
         $this->assertTrue($e->check($intelHash, "mypassword"));
         $this->assertFalse($e->check($wrongHash, "mypassword"));
     }
+
+    public function testPHPPasswordHashEncryptor()
+    {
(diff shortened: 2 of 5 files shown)


src/Security/PasswordEncryptor_PHPPasswordHash.php
```diff
@@ -0,0 +1,41 @@
...
+        $options = Config::inst()->get(static::class, 'options');
+        return password_hash($password, $algorithm, $options);
+    }
+
+    public function check($hash, $password, $salt = null, $member = null)
+    {
+        return password_verify($password, $hash);
+    }
+
+    public function salt($password, $member = null)
```
maxime-rainville: Maybe include a comment explaining why the salt is irrelevant in this context.

kinglozzer: Spoke to @dhensby briefly about this, we’re thinking it may be safer to keep the salt (it can inherit the method from `PasswordEncryptor::salt()`). As evidenced by the test failures, it appears the salt may be used in other locations 😕

kinglozzer:
Test failures are real, `Member::encryptWithUserSettings()` expects the following to be true:

```php
$string1 = Member::encryptWithUserSettings('test');
$string2 = Member::encryptWithUserSettings('test');
$string1 === $string2;
```

But if we’re using `password_hash()`, the values are different because a new salt is generated each time. You _can_ pass a salt to `password_hash()`, but it’s a bad idea:

- It’s documented as only working for the `PASSWORD_BCRYPT` algorithm
- It’s strongly recommended against
- It’s deprecated as of PHP 7.0.0

It looks like we may have to park this until 5.x 😞

dhensby:
If we get the application to store a new `Salt` against the `Member` object on password reset regardless of its use in the password, won't this solve the `encryptWithUserSettings` issue?

kinglozzer:
Nope, because `encryptWithUserSettings()` will use the member’s setting which is `PasswordEncryptor_PHPPasswordHash` which will ignore the salt set against the member

dhensby:
I see; well, that'll need to change if we want to be able to use this function, then ;)

To be honest, it's strange that `encryptWithUserSettings` acts quite this way; an encryptor should be able to manage it's own salt and that not cause these problems...

kinglozzer:
I can’t think of any nice way that doesn’t result in any breaking changes, I’ve also realised that we should probably make use of `password_needs_rehash()` too... I smell an RFC cooking 😬

dhensby:
I think making this change could be really useful. Removing `Salt` from the `Member` object (as it's now common practice to store the hash/salt/work/etc combined in one value) and replace it with some kind of security "token" that can be used to "encrypt with user settings" which would be managed separately to a password and allow things like expiring other login tokens independently of changing passwords

tests/php/Security/PasswordEncryptorTest.php
```diff
@@ -165,4 +165,14 @@ public function testEncryptorLegacyPHPHashCheck()
         $this->assertTrue($e->check($intelHash, "mypassword"));
         $this->assertFalse($e->check($wrongHash, "mypassword"));
     }
+
+    public function testPHPPasswordHashEncryptor()
+    {
+        $password = 'correcthorsebatterystaple';
```
flamerohr: yay, xkcd plug :D

chillu:
If you've been following along here, there's an RFC now: https://github.com/silverstripe/silverstripe-framework/issues/8526

PR #8065, opened 2018-05-11
NEW Add getJoinTable to MMTL

dhensby:
This is a small step to more closely aligning the API for `ManyManyList` and `ManyManyThroughList`

Ideally we'd add all the extra public methods that `ManyManyList` has to `ManyManyThroughList`

Missing methods:

- getLocalKey
- getForeignKey


tractorcow:
It seems you are adding getjointable not getjoinclass.

dhensby:
>It seems you are adding getjointable not getjoinclass

Whoops, that's right

tractorcow:
Regressions seem to be unrelated to the change.

PR #8075, opened 2018-05-18
FIX Polymorphic relationship class columns have obsolete class names remapped

### LINKED ISSUE #8052: SS4 upgrades: `classname_value_remapping` doesn't upgrade polymorphic relationships
Tested on SS 4.0

Initially raised in https://github.com/silverstripe/silverstripe-userforms/issues/761

The userforms EditableFormField has a polymorphic `Parent` relationship, which doesn't get updated with the dev/build command via `classname_value_remapping`. The logic that does that in DatabaseAdmin updates the ClassName column.

Should it look at polymorphic relationships as well? I'm probably going to patch something into userforms for now, but I feel like this would be a useful addition to `DatabaseAdmin::doBuild`, e.g. `// Remap obsolete class names` then `// Remap any polymophic relationships` - would need to find any DataObjects that have polymorphic relationships, then query for any of those that have obsolete class names in their record values. I'd expect this to slow down the dev/build process a little bit (a few seconds on an average site).

robbieaverill:
This repeats some of the logic for remapping legacy class names in the ClassName column of any DataObject to also apply to any DataObject that has polymorphic relationships, ensuring that those relationships are not corrupted while upgrading from SS 3.x to 4.x.

To test, add a polymorphic relationship to Page (for example) with some data in a SS3 site, import the DB into SS4 and run dev/build

Would be good to get some feedback on this. It does increase the number of DB queries run during a dev/build.

Resolves #8052

### DIFF
--- a/src/ORM/DatabaseAdmin.php
+++ b/src/ORM/DatabaseAdmin.php
@@ -407,7 +407,7 @@ class DatabaseAdmin extends Controller
 
         // Remap versioned table class name values as well
         /** @var Versioned|DataObject $class */
-        $class = DataObject::singleton($newClassName);
+        $class = DataObject::singleton($dataClass);
         if ($class->hasExtension(Versioned::class)) {
             if ($class->hasStages()) {
                 $updateQueries[] = sprintf($updateQuery, '_Live');
--- a/src/ORM/DatabaseAdmin.php
+++ b/src/ORM/DatabaseAdmin.php
@@ -7,9 +7,11 @@ use SilverStripe\Control\Controller;
 use SilverStripe\Control\Director;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Environment;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Manifest\ClassLoader;
 use SilverStripe\Dev\DevelopmentAdmin;
 use SilverStripe\Dev\TestOnly;
+use SilverStripe\ORM\FieldType\DBClassName;
 use SilverStripe\Security\Permission;
 use SilverStripe\Security\Security;
 use SilverStripe\Versioned\Versioned;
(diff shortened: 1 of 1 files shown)


src/ORM/DatabaseAdmin.php
```diff
@@ -390,6 +373,84 @@ public function doBuild($quiet = false, $populate = true, $testMode = false)
...
+     * @param string $dataClass    The data class to look up
+     * @param string $fieldName    The field name to look in for obsolete class names
+     * @param string $oldClassName The old class name
+     * @param string $newClassName The new class name
+     * @param bool $useBase        Whether to use the base data class for $dataClass
+     */
+    protected function updateLegacyClassNames($dataClass, $fieldName, $oldClassName, $newClassName, $useBase = true)
+    {
+        $schema = DataObject::getSchema();
+        $baseDataClass = $useBase ? $schema->baseDataClass($dataClass) : $dataClass;
```
kinglozzer: Could we work this out outside this method and pass it in as the `$dataClass` arg?

src/ORM/DatabaseAdmin.php
```diff
@@ -390,6 +373,84 @@ public function doBuild($quiet = false, $populate = true, $testMode = false)
...
+        if (!$badRecordCount) {
+            return;
+        }
+
+        if (Director::is_cli()) {
+            echo " * Correcting {$badRecordCount} obsolete {$fieldName} values for {$newClassName}\n";
+        } else {
+            echo "<li>Correcting {$badRecordCount} obsolete {$fieldName} values for {$newClassName}</li>\n";
+        }
+        $table = $useBase ? $schema->baseDataTable($dataClass) : $schema->tableName($dataClass);
```
kinglozzer: I think `$schema->tableName($baseDataClass)` should always be fine here?

robbieaverill: Yep you're right!

robbieaverill:
Updated with your comments, thanks @kinglozzer !

tractorcow:
Attacking polymorphic fields is not really the direct approach. The core issue is the sub-field of the composite PolymorphicForeignKey: The DBClassName field within it. Switch the filter to replace all DBClassName instead; This will cover BOTH the BaseTable.ClassName field, as well as the SomeTable.RelationClass fields from polymorphic relations, and prevent a double up of disjoint logic between the two.

robbieaverill:
That sounds sensible, thanks - I'll refactor and ping you when it's ready

robbieaverill:
Back from the dead!

@tractorcow I've updated the code to look for fields on DataObjects that are instances of DBClassName, and replace those instead. Thanks for the suggestion - it's much cleaner now!

cc @maxime-rainville

robbieaverill:
>Polymorphic relation WITH versioning ... the polymorphic column in the _live and _Version table didn't get updated.

Strange - there's logic in place for this. I'll have a look

maxime-rainville:
Just tried running multiple dev/builds under SS4 and I'm still seeing the problem using my example:
* In `ZFan` my SS3 records correctly have namespaced `FanOfClass` entry.
* In `ZFan_Live`, my SS3 records have a correctly namespaced `Class` column, but `FanOfClass` still reference the old class names.
* In `ZFan_Versions`, my SS3 records have a correctly namespaced `Class` column, but `FanOfClass` still reference the old class names.

maxime-rainville:
Raised https://github.com/silverstripe/silverstripe-framework/issues/8344 for the PostgreSQL thing.

PR #8090, opened 2018-05-23
FIX Prevent array to string conversion for default_sort on a fixed field

robbieaverill:
This shows a regression in **3.7.x-dev** where you can no longer define a `default_sort` that is a "fixed field" (e.g. Created from DataObject) without throwing an "array to string conversion" error.

One option is to cast empty arrays in `Object::parse_class_spec` to an empty string so this doesn't happen. I'm not sure where the regression happened, but this test shows it.

### DIFF
--- a/core/Object.php
+++ b/core/Object.php
@@ -213,6 +213,9 @@ abstract class SS_Object {
 	 * Returns a 2-elemnent array, with classname and arguments
 	 */
 	public static function parse_class_spec($classSpec) {
+		if ($classSpec === array()) {
+			$classSpec = '';
+		}
 		$tokens = token_get_all("<?php $classSpec");
 		$class = null;
 		$args = array();
--- a/tests/model/DataObjectTest.php
+++ b/tests/model/DataObjectTest.php
@@ -1772,8 +1772,8 @@ class DataObjectTest_SortOnFixed extends DataObject implements TestOnly {
 		'Name' => 'Varchar',
 	);
 
-	// Sorts on a "fixed field"
-	private static $default_sort = '"Created" DESC';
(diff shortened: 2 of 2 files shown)


kinglozzer:
I suspect this is related to https://github.com/silverstripe/silverstripe-framework/pull/7945/. It looks like `DataObject::db()` will always return an array for fixed fields 😕. Perhaps [this](https://github.com/kinglozzer/sapphire/blob/ebd3fb6526eb3ee9359111e548d9f6b6e0323e97/model/DataObject.php#L3424) could be refactored to something like

```php
// Outside the parent foreach()
$fields = DataObject::database_fields(get_class($this));

// ...

$columnSpec = $fields[$column];
list($fieldType) = SS_Object::parse_class_spec($columnSpec);
```

core/Object.php
```diff
@@ -213,6 +213,9 @@ public static function create_from_string($classSpec, $firstArg = null) {
 	 * Returns a 2-elemnent array, with classname and arguments
 	 */
 	public static function parse_class_spec($classSpec) {
+		if ($classSpec === array()) {
```
tractorcow: This is a little obscure, where the underlying reason is "sometimes code passes in an empty array in some edge case". It'd be better to never call this method with an invalid arg, rather than silently ignoring the error.

tractorcow:
I've fixed this with https://github.com/silverstripe/silverstripe-framework/pull/8146

The issue is nothing to do with parse_class_spec, it's a distraction.

database_fields() doesn't return anything for ID so you're out of luck there anyway.

The fix was to use hasOwnTableDatabaseField() instead. See https://github.com/silverstripe/silverstripe-framework/pull/8146

robbieaverill:
Thank you!

PR #8091, opened 2018-05-23
Adding a "clearfix" to gridfield-buttonrow

ScopeyNZ:
Currently the `GridFieldButtonRow` component has two floating groups (a left and right) but doesn't clear the float. This means that any subsequent blocks including subsequent `GridFieldButtonRow`s float weirdly:

Behaviour:
image
Dev Tools:
image

I have just added a typical `:after` clearfix and bundled the compiled CSS and map file.

I couldn't find any specific related issues.


robbieaverill:
We're only accepting critical bug and security fixes into 3.x now. Does this still affect v4? If so could we retarget it at that?

robbieaverill:
I don't think we'll be merging this into 3.x any more. Feel free to reopen against 4.x if you'd like to and it's still an issue there =) thanks for the PR though

PR #8094, opened 2018-05-24
FIX Remove use of deprecated each method from thirdparty

dhensby:
For php 7.2 compatibility we need to remove uses of `each` which is now deprecated


dhensby:
> It looks like the built failures might be related

@robbieaverill absolutely were, yes. I'd forgotten that `next` doesn't ever return the first element from an array so all my `while(...) { ... } do` needed to become `do { ... } while(...)`

Issue #8020, opened 2018-04-19
Incorporate reading mode in default object cache key


chillu:
In partial caching, we use `$CurrentReadingMode` and `$CurrentUser.ID` as a default value in the cache key. 

> An additional global key is incorporated in the cache lookup. The default value for this is $CurrentReadingMode, $CurrentUser.ID. This ensures that the current Versioned state and user ID are used. This may be configured by changing the config value of SSViewer.global_key. It is also necessary to flush the template caching when modifying this config, as this key is cached within the template itself.

https://docs.silverstripe.org/en/4/developer_guides/performance/partial_caching/#partial-caching

We don't do this for [object caching](https://docs.silverstripe.org/en/4/developer_guides/performance/caching/) through `CacheFactory`. This makes it far too easy for developers to forget they're caching an object representation tied to the current stage, which has the potential to leak draft content. In SilverStripe 4, the likelyhood of this will be higher, since more objects are versioned, and graphs of objects are more likely to unexpectedly contain versioned records as well.

```yml
SilverStripe\Core\Injector\Injector:
  Psr\SimpleCache\CacheInterface.myCache:
    factory: SilverStripe\Core\Cache\DefaultCacheFactory
    constructor:
      namespace: "myCache"
```

```php
$cache = Injector::inst()->get(CacheInterface::class . '.myCache');
$myValue = $cache->get('myCacheKey'); // gets myCacheKey + CurrentReadingMode + CurrentUser
```

In order to opt out, you'd use a different cache factory:

```yml
SilverStripe\Core\Injector\Injector:
  Psr\SimpleCache\CacheInterface.myCache:
    factory: SilverStripe\Core\Cache\CustomCacheFactory
    constructor:
      namespace: "myCache"
```

It's technically a semver violation to introduce this behaviour in the existing `CacheFactory`, but we could recommend using subclasses of it in our docs and upgrading guide. Not too happy with the naming, suggestions welcome.

# PRs:

 - [x] 4.x: https://github.com/silverstripe/silverstripe-versioned/pull/148
- [ ] 4.x https://github.com/silverstripe/silverstripe-framework/pull/8093
- [x] 3.x https://github.com/silverstripe/silverstripe-framework/pull/8095
- [x] 3.x https://github.com/silverstripe/silverstripe-cms/pull/2180

dhensby:
> It's technically a semver violation to introduce this behaviour in the existing CacheFactory

is it? Why?

[...]

PR #8093, opened 2018-05-24
Remove versioned cache from core caches

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-framework/issues/8020#issuecomment-391268602

PR #8095, opened 2018-05-25
ENHANCEMENT: Add new CacheProxy to segment versioned state

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-framework/issues/8020

### DIFF
--- a/cache/CacheProxy.php
+++ b/cache/CacheProxy.php
@@ -1,0 +1,149 @@
+<?php
+
+require_once 'Zend/Cache.php';
+
+class CacheProxy extends Zend_Cache_Core
+{
+    /**
+     * @var Zend_Cache_Backend|Zend_Cache_Backend_ExtendedInterface
+     */
+    protected $container;
+
+    /**
+     * CacheProxy constructor.
+     * @param Zend_Cache_Core $container
+     */
+    public function __construct(Zend_Cache_Core $container)
+    {
+        $this->container = $container;
+
+        parent::__construct();
+    }
+
+    /**
+     * @param array $directives
+     */
+    public function setDirectives($directives)
+    {
+        $this->container->setDirectives($directives);
+    }
+
+    public function setConfig(Zend_Config $config)
+    {
+        return $this->container->setConfig($config);
+    }
+
+    public function setBackend(Zend_Cache_Backend $backendObject)
+    {
+        return $this->container->setBackend($backendObject);
+    }
+
+    public function getBackend()
+    {
+        return $this->container->getBackend();
+    }
+
+    /**
+     * @param string $name
+     * @param mixed $value
+     */
+    public function setOption($name, $value)
+    {
+        $this->container->setOption($name, $value);
+    }
+
+    public function getOption($name)
+    {
+        return $this->container->getOption($name);
+    }
+
+    public function setLifetime($newLifetime)
+    {
+        return $this->container->setLifetime($newLifetime);
+    }
+
+    public function getIds()
+    {
+        return $this->container->getIds();
+    }
+
+    public function getTags()
+    {
+        return $this->container->getTags();
+    }
+
+    public function getIdsMatchingTags($tags = array())
+    {
+        return $this->container->getIdsMatchingTags($tags);
+    }
+
+    public function getIdsNotMatchingTags($tags = array())
+    {
+        return $this->container->getIdsNotMatchingTags($tags);
+    }
+
+    public function getIdsMatchingAnyTags($tags = array())
+    {
+        return $this->container->getIdsMatchingAnyTags($tags);
+    }
+
+    public function getFillingPercentage()
+    {
+        return $this->container->getFillingPercentage();
+    }
+
+    public function getMetadatas($id)
+    {
+        return $this->container->getMetadatas($this->createKey($id));
+    }
+
+    public function touch($id, $extraLifetime)
+    {
+        return $this->container->touch($this->createKey($id), $extraLifetime);
+    }
+
+    public function load($id, $doNotTestCacheValidity = false, $doNotUnserialize = false)
+    {
+        return $this->container->load($this->createKey($id), $doNotTestCacheValidity, $doNotUnserialize);
+    }
+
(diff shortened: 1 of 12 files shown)


cache/CacheProxy.php
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    public function clean($mode = Zend_Cache::CLEANING_MODE_ALL, $tags = array())
+    {
+        return $this->container->clean($mode, $tags);
+    }
+
+    /**
+     * Creates a dynamic key based on versioned state
+     * @param $key
```
robbieaverill: Missing type

cache/CacheProxy.php
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /**
+     * @param array $directives
+     */
+    public function setDirectives($directives) {
+        $this->container->setDirectives($directives);
+    }
+
+    public function setConfig(Zend_Config $config)
```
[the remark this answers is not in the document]
unclecheese: It's optional. You can use it if you want to mark something explicitly documented, but this is standards compliant as it is.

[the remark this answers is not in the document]
unclecheese: In fact, I think I'll just remove all the others, because it's inconsistent.

docs/en/04_Changelogs/3.7.0.md
```diff
@@ -0,0 +1,26 @@
+# 3.7.0
+
+### Versioned cache segmentation
+
+The cache API now maintains separate cache pools for each versioned stage. This prevents users from caching draft data and then exposing it on the live stage.
```
chillu: Call it `SS_Cache` API, to be clear about which cache you mean (e.g. not HTTP cache).

Also, I think the intent for this change should be more obvious. Suggested rewording:

> This prevents developers from caching draft data and then accidentally exposing it on the live stage without potentially required authorisation checks. Unless you rely on caching across stages, you don't need to change your own code for this change to take effect. Note that cache keys will be internally rewritten, causing any existing cache items to become invalid when this change is deployed.

docs/en/04_Changelogs/3.7.0.md
```diff
@@ -0,0 +1,26 @@
...
+
+```php
+// Before:
+$cache = SS_Cache::factory('myapp');
+Versioned::set_reading_mode('Stage.Live');
```
chillu: Shouldn't the reading mode be the other way around, first `Stage.Stage`, then `Stage.Live`?

chillu:
OK I'm happy with this, but would like a second opinion either from @dhensby or @sminnee

kinglozzer:
This captures all the methods defined in `Zend_Cache_Core`, but the `Zend_Cache_Frontend_<driver>` classes sometimes add extra methods that wouldn’t be covered by this. Could/should we add `__call()` for those cases? I imagine they’re scarcely used, but would be broken by this change

chillu:
@unclecheese Can you please go through the 3.x and 4.x default caches in core, and opt out where this behaviour isn't required? Otherwise we'll cause performance regressions, e.g. for inherited permissions caches. See Damian's comment on https://github.com/silverstripe/silverstripe-versioned/pull/148

unclecheese:
Sure. The 4.x ones are already done, though. Do you need any changes to that? https://github.com/silverstripe/silverstripe-framework/pull/8093#issuecomment-393313829

docs/en/04_Changelogs/3.7.0.md
```diff
@@ -0,0 +1,26 @@
+# 3.7.0
+
+### Versioned cache segmentation
```
chillu: @unclecheese Can you please document this in a persistent place in the docs as well? Also apply this to the 4.x PRs

chillu:
@unclecheese Can you please check the Travis failures? E.g. there's PHP 5.3 incompatible syntax.

docs/en/04_Changelogs/3.7.0.md
```diff
@@ -0,0 +1,26 @@
...
+$cache = SS_Cache::factory('myapp');
+Versioned::set_reading_mode('Stage.Stage');
+$cache->save('Some draft content. Not for public viewing yet.', 'my_key');
+Versioned::set_reading_mode('Stage.Live');
+$cache->load('my_key'); // null
+```
+Data that is not content sensitive can be cached across stages by simply opting out of the segmented cache with the `disable-container` argument.
+
+```php
+$cache = SS_Cache::factory('myapp', 'Output', array('disable-container' => true));
```
chillu: Should be `disable-segmentation`

chillu:
Have you checked how this affects `SSViewer` partial caching? Presumably this is also using `SS_Cache::factory()`, and now includes the stage twice in the cache key? Once from the `SS_Cache::factory()` logic, once from the custom `SSViewer` logic? That unnecessarily changes existing caches, and feels a bit confusing.

tractorcow:
Yes we can use the same disable-segmentation key in SSViewer to disable that. It's not necessary to support it where we already respect viewing mode.

tractorcow:
@ingo addressed above.

tractorcow:
Behat issues fixed over at https://github.com/silverstripe/silverstripe-behat-extension/pull/178

tractorcow:
Testing this PR locally with that behat PR looks good. Just need to get it running on travis.

robbieaverill:
PR to increase memory limit in Travis: https://github.com/silverstripe/silverstripe-framework/pull/8129

chillu:
@unclecheese Behat tests are still failing (just retriggered it a few mins ago), even though the memory usage PR Robbie referenced above got merged. Maybe we need to have a release for behat-extension?

docs/en/04_Changelogs/3.7.0.md
chillu: @unclecheese Still outstanding

tractorcow:
just tagged behat extension again. Let's see if it fixes it.

tractorcow:
Only remaining issue is a postgres issue that seems unrelated to this issue.

PR #8101, opened 2018-05-29
FIX PHP 5.3 compat for referencing $this in closures

robbieaverill:


tractorcow:
Regressions seem unrelated... merging as this blocks other fixes.

PR #8109, opened 2018-05-30
Object upgrade notice for 3.7.0 changelog

chillu:
See https://forum.silverstripe.org/t/php-7-2-backwards-compat-for-module-maintainers-on-ss-3-7/464/5


docs/en/04_Changelogs/3.7.0.md
```diff
@@ -0,0 +1,29 @@
...
+If you are publishing a 3.x-compatible module that directly references to the `Object` class,
+it's a bit more complicated. You can apply one of two fixes:
+
+ * Mark your module as not working with PHP 7.2, by putting `"php": ">=5.3, <7.2"` in your `composer.json`.
+ * Create a new version of your module that requires `"silverstripe/framework": "^3.7"`.
+   In this version, replace all references to `Object` with `SS_Object`.
+   This version won’t work on SilverStripe 3.6,
+   but it will work on SilverStripe 3.7+ on any version of PHP.
+   Note that in SilverStripe 4.x, the `Object` class was deleted so there isn’t an `SS_Object` class there either
+   (see [https://docs.silverstripe.org/en/4/changelogs/4.0.0/](changelog))
```
robbieaverill: Do you think there's a third option which is to avoid using the `Object` parent class altogether? This could be doable in some situations but not always, i.e. a utility class can be loaded via `Injector::inst()->get(MyClass::class)` rather than `MyClass::singleton()`

chillu: Can you list a few before/after cases which make `Object` obsolete?

robbieaverill: Perhaps it's not worth mentioning... you'd lose all the extensibility by not extending Object/SS_Object. Ignore me =)

docs/en/04_Changelogs/3.7.0.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+If you are publishing a 3.x-compatible module that directly references to the `Object` class,
+it's a bit more complicated. You can apply one of two fixes:
+
+ * Mark your module as not working with PHP 7.2, by putting `"php": ">=5.3, <7.2"` in your `composer.json`.
+ * Create a new version of your module that requires `"silverstripe/framework": "^3.7"`.
+   In this version, replace all references to `Object` with `SS_Object`.
+   This version won’t work on SilverStripe 3.6,
+   but it will work on SilverStripe 3.7+ on any version of PHP.
```
kinglozzer: If module devs added the following one-liner to `_config.php`, would that mean they could support both 3.6 and 3.7 without having to create new release lines/tags?

```php
if (!class_exists('SS_Object')) class_alias('Object', 'SS_Object');
```

They could then swap out `Object::whatever()` for `SS_Object::whatever()` while still supporting 3.6. It won’t break anything on PHP 7.2, because `SS_Object` would already exist

chillu:
Alright, I've rewritten this with @kinglozzer suggestion for the `class_alias()`, and made the decision path for project code a bit clearer.

PR #8113, opened 2018-05-31
BUGFIX: Allow buttons to opt out of display

unclecheese:
This bug is breaking Behat tests. https://travis-ci.org/silverstripe/silverstripe-framework/jobs/385964668#L896


src/Forms/GridField/GridField_ActionMenu.php
```diff
@@ -31,15 +31,20 @@ public function getColumnContent($gridField, $record, $columnName)
             return null;
         }
 
-        $schema = array_map(function (GridField_ActionMenuItem $item) use ($gridField, $record, $columnName) {
-            return [
+        $schema = [];
+        /* @var GridField_ActionMenuItem $item */
+        foreach ($items as $item) {
+            if (!$item->getGroup($gridField, $record, $columnName)) {
```
maxime-rainville: You're calling `getGroup()` twice. Looks like it would be better to assign the return value to a variable and work from that.

src/Forms/GridField/GridFieldGroupDeleteAction.php
```diff
@@ -41,6 +41,15 @@ public function getColumnContent($gridField, $record, $columnName)
         return null;
     }
 
+    public function getGroup($gridField, $record, $columnName)
```
maxime-rainville: Couldn't find a unit test for this class or any reference to `GridFieldGroupDeleteAction` under `tests/php/Forms/GridField`.

Looks like there's a bit of conditional logic in this class that should be tested.

robbieaverill:
Yay for green Behat though!

src/Forms/GridField/GridFieldGroupDeleteAction.php
unclecheese: Have added a unit test.

src/Forms/GridField/GridFieldGroupDeleteAction.php
```diff
@@ -41,6 +41,22 @@ public function getColumnContent($gridField, $record, $columnName)
...
+    /**
+     * Get the ActionMenu group (not related to Member group)
+     * @param GridField $gridField
+     * @param DataObject $record
+     * @param $columnName
+     * @return null|string
+     */
+    public function getGroup($gridField, $record, $columnName)
+    {
+        if (!$this->canUnlink($record)) {
```
tractorcow: Why are we only fixing this bug for groups and not GridFieldDeleteAction? You still need to check `canDelete` / 'canEdit' on that action also. Check the logic in GridFieldDeleteAction::getColumnContent() for the necessary legacy behaviour to replicate.

unclecheese: Have migrated this fix to `GridFieldDeleteAction` as well.

tractorcow: Thanks good job.

tractorcow:
Merge on green.

tractorcow:
Yep still merge on green.

tractorcow:
Also fixes https://app.zenhub.com/workspace/o/silverstripe/silverstripe-framework/issues/8123

Issue #1387, opened 2013-04-03
[ORM] [2010-10-18] Databases other than MySQL don't modify existing data for enumerated fields


silverstripe-issues:
created by: @halkyon (sharvey)
created at: 2010-10-18
original ticket: http://open.silverstripe.org/ticket/6117

---

When modifying the default value of an "Enum" DBField type in sapphire, the default value never gets fixed when dev/build is called. The reason for this is because MySQL uses enum data types, whereas MSSQL and PostgreSQL use constraints "CHECK" and "IN" keywords to check the values.

This code in particular not work with MSSQLDatabase, or PostgreSQLDatabase because it assumes the data type will always be an enum.

```
        } else if($fieldValue != $specValue) {
            // If enums/sets are being modified, then we need to fix existing data in the table.
            // Update any records where the enum is set to a legacy value to be set to the default.
            // One hard-coded exception is SiteTree - the default for this is Page.
            foreach(array('enum','set') as $enumtype) {
                if(preg_match("/^$enumtype/i",$specValue)) {
                    $newStr = preg_replace("/(^$enumtype\s*\(')|('$\).*)/i","",$spec_orig);
                    $new = preg_split("/'\s*,\s*'/", $newStr);

                    $oldStr = preg_replace("/(^$enumtype\s*\(')|('$\).*)/i","", $fieldValue);
                    $old = preg_split("/'\s*,\s*'/", $newStr);

                    $holder = array();
                    foreach($old as $check) {
                        if(!in_array($check, $new)) {
                            $holder[] = $check;
                        }
                    }
                    if(count($holder)) {
                        $default = explode('default ', $spec_orig);
                        $default = $default[1];
                        if($default == "'SiteTree'") $default = "'Page'";
                        $query = "UPDATE \"$table\" SET $field=$default WHERE $field IN (";
                        for($i=0;$i+1<count($holder);$i++) {
                            $query .= "'{$holder[$i]}', ";
                        }
                        $query .= "'{$holder[$i]}')";
                        DB::query($query);
                        $amount = DB::affectedRows();
                        $this->alterationMessage("Changed $amount rows to default value of field $field (Value: $default)");
                    }
                }
            }
            Profiler::mark('alterField');
            $this->transAlterField($table, $field, $spec_orig);
            Profiler::unmark('alterField');
            $this->alterationMessage("Field $table.$field: changed to $specValue <i style=\"color: #AAA\">(from {$fieldValue})</i>","changed");
        }
```

[...]

PR #8118, opened 2018-05-31
do not assume enum has default

lekoala:
I tried to override default enum behaviour and run into this issue where "default" might not be set
Adding a simple check on spec_orig fixes that

### DIFF
--- a/src/ORM/Connect/DBSchemaManager.php
+++ b/src/ORM/Connect/DBSchemaManager.php
@@ -733,45 +733,17 @@ MESSAGE
     }
 
     /**
-     * Determine if the provided column specification is for an enum and return a suitable SQL Query string to update
-     * legacy values if necessary.
+     * Determine if the provided column specification is for an enum and update invalidated values if need be.
      *
-     * @param string $specValue Specification for a column
-     * @return Query
+     * @param string $specValue Updated specification for the column.
+     * @param string $spec_orig Original specification for the column.
+     * @param string $fieldValue Current specification for the column.
+     * @param string $table Name of the table.
+     * @param string $field Name of the field.
+     * @return Query|null
      */
     protected function resetInvalidatedEnumValue($specValue, $spec_orig, $fieldValue, $table, $field)
     {
-        // If enums/sets are being modified, then we need to fix existing data in the table.
-        // Update any records where the enum is set to a legacy value to be set to the default.
-        foreach (array('enum', 'set') as $enumtype) {
-            if (preg_match("/^$enumtype/i", $specValue)) {
-                $newStr = preg_replace("/(^$enumtype\\s*\\(')|('\\).*)/i", "", $spec_orig);
-                $new = preg_split("/'\\s*,\\s*'/", $newStr);
-
-                $oldStr = preg_replace("/(^$enumtype\\s*\\(')|('\\).*)/i", "", $fieldValue);
-                $old = preg_split("/'\\s*,\\s*'/", $oldStr);
-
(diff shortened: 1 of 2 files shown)


maxime-rainville:
Thanks for submitting a PR @lekoala. Is this fixing a specific issue? If not, could you open one and explain what the desired behavior is and how it differs from the existing behavior.

Also, could you write some unit test?

lekoala:
if you have an enum without default, the current code will fail because exploding on "default " will result on an array with only 1 element and therefore looking for the element 1 will result in a undefined offset error.

The current code always assume that the array will have 2 elements, which is not safe to do.

It could be solved also by checking if the array has indeed 2 elements before doing $default = $default[1];

maybe something like, if (isset($default[1]) or something similar, but that adds an extra if.

I'm not planning on writing unit tests because it's not so easy to run the test suite locally. However, by reading the code, I believe it's quite obvious that a check should be mode on the existence of $default[1] before using it.

maxime-rainville:
I think I understand what you are gunning for. Would the following code sample be a good illustration of what you are trying to achieve?

```php

<?php
use SilverStripe\ORM\DataObject;
use SilverStripe\ORM\FieldType\DBEnum;

class Dummy extends DataObject {
    private static $db = [
        'MagicEnum' => "NullableEnum(array('Active', 'Injured', 'Retired'))"
        // 'MagicEnum' => "NullableEnum(array('Active', 'Injured'))"
    ];
}

class NullableEnum extends DBEnum {

    public function getDefault()
    {
        return null;
    }

        return parent::scaffoldFormField($title, $params)->setEmptyString('(null)');
    }
}
```

Basically, if you remove `Retired` from the list of available values and run a `dev/build`, you'll get an `Undefined offset` notice using our current production logic. Your fix addresses this edge case.

lekoala:
Yes exactly :)

lekoala:
so what do you think about merging this? (just bumping that again because I had the issue once again on one of my project). The build failure does not seem related to the patch.

maxime-rainville:
Sorry, it dropped off my radar. I still would like to see a unit test for it, however.

maxime-rainville:
I'll double check with some folks internally to see if there's an easy way to write test for something that only occurs after a multiple DB build.

lekoala:
any update on this?

lekoala:
the last test is a welcome addition ! 

looks good to me

maxime-rainville:
Raised a related SQLite issue https://github.com/silverstripe/silverstripe-sqlite3/issues/54

maxime-rainville:
Looks like PostgreSQL has the same issue as SQLite :frowning_face:

Probably need to move the bits of logic that decide whatever to update invalidated enum values to a separate method to a separate method so other DBSchemaManagers can override it.

@lekoala I can have a look at this later if you don't feel like patching every single one of our supported DB imlementations.

lekoala:
Well, I'm not sure there is a need to be able to change the value to default if you remove it, although it's nice to have, but if drivers don't support it, it probably don't make a lot of sense anyway.

However, having enums with null as a default should work out of the box on all drivers no?

maxime-rainville:
I got a PR for PostgreSQL https://github.com/silverstripe/silverstripe-postgresql/pull/92

maxime-rainville:
I think this is related to https://github.com/silverstripe/silverstripe-framework/pull/8118

maxime-rainville:
:man_facepalming: https://github.com/silverstripe/silverstripe-framework/issues/1387

PR #8122, opened 2018-06-01
BUG Prevent stage querystring args during dev/build

### LINKED ISSUE #8110: updateLink versioned extension regression on dev/build
## Affected Version

3.x-dev

## Description

When running a `dev/build` via the browser on an empty database (or specifically without any `ErrorPage` records when using the CMS module) the user is logged out and sent back to `/Security/login?BackURL=%2Fpage-not-found%2F%3Fstage%3DStage`

Expected behaviour: for `dev/build` to complete successfully.

I've narrowed this down to the following:

1. `DevelopmentAdmin::init()` calls `Versioned::reading_stage('Stage');` (https://github.com/silverstripe/silverstripe-framework/blob/47a9cdfd49146e769760e8d8db3f01925597de41/dev/DevelopmentAdmin.php#L76-L79)
2. `VersionedStateExtension::updateLink()` appends `stage=Stage` to links when the object was queried from the draft stage (https://github.com/silverstripe/silverstripe-framework/blob/47a9cdfd49146e769760e8d8db3f01925597de41/model/VersionedStateExtension.php#L23-L48), this means switching stages after the object has been fetched (and possibly published) won't fix the problem and `?stage=Stage` will still be appended to the URL.
3. `ErrorPage::requireDefaultRecords()` uses `Director::test()` to "statically publish" the error pages it generates (https://github.com/silverstripe/silverstripe-cms/blob/12826b2ace76808d29b5017e1f445514b7fcf62a/code/model/ErrorPage.php#L116)
4. `VersionedRequestFilter::preRequest()` takes outputting responses into it's own hands instead of returning a boolean response
5. `VersionedRequestFilter::preRequest()` saves the dummy session from `Director::test()` which causes the current session to be invalidated and the user logged out

## Steps to Reproduce

1. Create a vanilla 3.x-dev site or delete the error pages from the CMS
2. Go to `dev/build` in the browser

## Potential fix

I've fixed this locally by wrapping the `ErrorPage::requireDefaultRecords()` logic in `Versioned::reading_stage(Versioned::LIVE);` which fixes the specific error, but I feel this only resolves part of the issue.

Another fix could be to strip the `?stage` query param from `$page->Link()` in `ErrorPage::requireDefaultRecords()`

---

I've opened this against framework instead of CMS because most of the root issues are in framework and the PR to framework caused this regression

# PRs:

 - [x] https://github.com/silverstripe/silverstripe-framework/pull/8122
 - [x] https://github.com/silverstripe/silverstripe-cms/pull/2181

tractorcow:
Fixes #8110

### DIFF
--- a/dev/DevelopmentAdmin.php
+++ b/dev/DevelopmentAdmin.php
@@ -76,7 +76,8 @@ class DevelopmentAdmin extends Controller {
 		// Backwards compat: Default to "draft" stage, which is important
 		// for tasks like dev/build which call DataObject->requireDefaultRecords(),
 		// but also for other administrative tasks which have assumptions about the default stage.
-		Versioned::reading_stage('Stage');
+		Versioned::reading_stage(Versioned::DRAFT);
+		Versioned::set_default_reading_mode(Versioned::get_reading_mode());
 	}
 
 	public function index() {


dhensby:
Looks like there's a regression in behat?

tractorcow:
Fixed behat issue with https://github.com/silverstripe/silverstripe-behat-extension/pull/178

tractorcow:
Rebased; Waiting for it to rebuild.

tractorcow:
PG is broken but getting fixed in another PR. Hopefully here https://github.com/silverstripe/silverstripe-framework/pull/8095

dhensby:
well, it was fixed in #8132 too ;)

PR #8129, opened 2018-06-02
Increase memory limit to 2G in Travis builds

robbieaverill:
Framework Travis builds are exceeding their memory limit and failing (1.6GB) so increasing the memory limit to 2G


dhensby:
this is mental :(

We really need to find the source of all the memory usage and try to clear it up

robbieaverill:
Yeah! Weirdly it’s maxing our during the composer install on framework, not even getting to do anything

PR #8132, opened 2018-06-04
FIX Regression from #8009

dhensby:
fixes #8009 

This fixes the typical sort issue with fetching from postgres DBs

I've replaced the `$this->assertEquals($list->Count(), $listDuplicate->count())` with predefined numbers because this helps show what's going on to people reading it and also helps catch regressions.

It *is* expected that if you duplicate both sides of a `ManyMany` that you'd get 2 relations where you previously had only 1 and so these changed assertions are correct, the (intermittent) test failures reported in #8009 come from the fact that postgres has unreliable default sorting on objects so the first item wasn't always the one we were asserting against.


robbieaverill:
Related 4.0 fix: #8150

PR #8133, opened 2018-06-04
FIX PHP 5.3 compat for referencing $this in closures

dhensby:
backport of #8101


kinglozzer:
Behat failures unrelated

PR #8140, opened 2018-06-05
BUG Backport bugfix for belongs_many_many with many_many through.

### LINKED ISSUE #8131: belongs_many_many when though list used for many_many
## Affected Version

4.1.0

## Description

Although [developer guide](https://docs.silverstripe.org/en/4/developer_guides/model/relations/#many-many-through-relationship-joined-on-a-separate-dataobject) says you can use `$belongs_many_many` relationship for many_many_through lists as you used to, it looks like you can't. 

When I do it as in the documentation (adding/updating relation classes with FQNs) it always complaining `"belongs_many_many relation {$parentClass}.{$component} points to {$specification} without matching many_many"`

## Steps to Reproduce

If you set up an example like in the [documentation](https://docs.silverstripe.org/en/4/developer_guides/model/relations/#many-many-through-relationship-joined-on-a-separate-dataobject) but update relation classes with FQNs then you'll simply notice this issue.

## PRs:

 - [ ] https://github.com/silverstripe/silverstripe-framework/pull/8140

tractorcow:
Partial backport of #7928
Fixes #8131

### DIFF
--- a/src/ORM/DataObjectSchema.php
+++ b/src/ORM/DataObjectSchema.php
@@ -982,9 +982,17 @@ class DataObjectSchema
         if (!$otherManyMany) {
             return null;
         }
-        foreach ($otherManyMany as $inverseComponentName => $nextClass) {
-            if ($nextClass === $parentClass) {
-                return $inverseComponentName;
+        foreach ($otherManyMany as $inverseComponentName => $manyManySpec) {
+            // Normal many-many
+            if ($manyManySpec === $parentClass) {
+                 return $inverseComponentName;
+            }
+            // many-many through, inspect 'to' for the many_many
+            if (is_array($manyManySpec)) {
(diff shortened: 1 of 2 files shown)


priyashantha:
@dhensby I can see this fix is in 4.1 branch now. But just checking whether when this fix be added into a tag so we can use it via recipe-plugin.

tractorcow:
You can add a custom dependency to your root
```json
{
	"requirements": {
		"silverstripe/framework": "4.1.x-dev as 4.1.1"
	}
}
```

That will let you use untagged versions without modifying your recipe.

priyashantha:
@tractorcow Thanks for that. I'll use this. 👍

dhensby:
You don't even need to (and shouldn't) use the aliasing if you set up the minimum-stability and prefer-stable settings appropriately

tractorcow:
True I was thinking more for usage with cwp recipe (which only has stable tags).

PR #8143, opened 2018-06-06
The definitive upgrade guide to SilverStripe 4

maxime-rainville:
A comprehensive step-by-step guide to upgrading a SilverStripe 3 site to SilverStripe 4.

# Parent issue
https://github.com/silverstripe/silverstripe-upgrader/issues/72


docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+
+Basically, you'll need to go through the entire code base and update all references from the old non-namespaced SilverStripe classes to new namespaced equivalent. If you are referencing classes from third party packages that have been namespaced, you'll need to update those as well.
+
+#### Update explicit references to classes in your code
+
+Wherever your code explicitly references a SilverStripe class, it will need to be updated to the new namespaced equivalent. You can either update the reference to use the fully qualified name of the class or you can add a `use` statement to your file.
+
+For example take the following SilverStripe 3 class. `DataObject` and `FieldList` need to point to their namespace equivalents.
+
+```php
```
dhensby: Would using diff markdown be better for these kinds of before and after examples?

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+  SilverStripe\Security\Member:
+    SINGULARNAME: Coyote
+  Acme\Web\RoadRunner:
+    SALUTATION: Beep Beep
+```
+
+### Finalising namespace updates
+Whatever you upgraded your code automatically or manually you'll need to perform the two following steps manually.
+
+DataObject tables will default to use a namespaced name. For example, if you have class under `Acme\Web\Products\ExplosiveTennisBall` that extends `DataObject`, the matching table in your database will be called `Acme_Web_Products_ExplosiveTennisBall`.
```
dhensby: make it clear here than "tables" means "database tables".

Also, why doesn't the upgrader add BC table names to classes?

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+
+You can define a `private static $table_name` property on your DataObjects to use more convenient table names. For example, `private static $table_name = 'ExplosiveTennisBall';`.
+
+If your template files contain translatable strings, they also need to be updated to referenced the namespaced classes.
+
+For example, `<%t Member.SINGULARNAME 'Member' %>` would become `<%t SilverStripe\Security\Member.SINGULARNAME 'Member' %>`.
```
dhensby: You haven't mentioned here the fact that you can't use `static::class` or `parent::class` in when upgrading string representations in _t() usage.

I'd make it clear that in those cases either `__CLASS__` or `self::class` has to be used. I'd also mention this further up when talking about converting from strings to namespaces

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+
+Your codebase should now be referencing valid SilverStripe 4 classes. This means that your classes can be loaded at runtime. However, your codebase will still be using an outdated API.
+
+This is a good point to commit your changes to your source control system before moving on to the next step.
+
+## Step 5 - Updating your codebase to use SilverStripe 4 API {#step5}
+
+This is the most intricate and potentially time-consuming part of the upgrade. Basically, it involves going through your entire codebase to remove references to deprecated APIs and update your project logic.
```
dhensby: Remove "Basically,"

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+## Step 6 - Update your entry point {#step6}
+The location of SilverStripe's _entry file_ has changed. Your project and server environment will need
+to adjust the path to this file from `framework/main.php` to `index.php`.
+
+### Update your entry point
+You can get a copy of the SilverStripe 4 `index.php` file at:
+* `vendor/silverstripe/recipe-core/public/index.php` if you are upgrading to SilverStripe 4.1 or above
+* `vendor/silverstripe/recipe-core/index.php` if you are upgrading to SilverStripe 4.0
+
+If you've modified your SilverStripe 3 `index.php`, you'll need to reconcile those changes with the index.php file you got from `recipe-core`. Otherwise, just use the generic `index.php` file `recipe-core` provides.
```
dhensby: It's not really possible to modify your `index.php` file in SS 3 because the default "entry point" is `framework/main.php`... If someone has reworked their entry point to be an `index.php` file I think they are probably advanced enough to know about it and that this does not need mentioning. It also directly contradicts what you've said above about the old entry point and so could be confusing for readers

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+[Refer to the installation instruction for your platform](/getting_started/installation/) if your server configuration is not managed via a `.htaccess` or `web.config` file.
+
+### Finalising the entry point upgrade
+
+At this stage, you could in theory run your project in SilverStripe 4.
+
+This is a good point to commit your changes to your source control system before moving on to the next step.
+
+## Step 7 - Update project structure (optional) {#step7}
+SilverStripe 4.0 introduces a new recommended project structure. Adopting the recommended project structure is optional, but will become mandatory in SilverStripe 5.
```
dhensby: Was this 4.0 or 4.1?

docs/en/03_Upgrading/index.md
```diff
@@ -10,9 +10,61 @@ your SilverStripe installation means overwriting files, flushing the cache and u
 See our [upgrade notes and changelogs](/changelogs/4.0.0) for 4.0.0 specific information, bugfixes and API changes.
 </div>
 
-## Composer
+## Table of Content
```
maxime-rainville: Get ride of TOC. It's auto-generated when the doc is build.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+SS_DATABASE_CLASS="MySQLDatabase"
+SS_DATABASE_USERNAME="root"
+SS_DATABASE_PASSWORD=""
+SS_DATABASE_SERVER="127.0.0.1"
 ```
 
-### Switching to new `app/src` structure
+### Cleaning up `mysite/_config.php` after your environment configuration upgrade
+
+You'll need to clean up your `mysite/_config.php` file after upgrading your environment file.
```
chillu: Might be worth mentioning that any `Config->update()` calls will slow down your entire execution in the way the config system is implemented in 4.x (or rather speed it up if you don't use it) - as a motivator for fixing up this file.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
-The `reorganise` command can automatically rename the `mysite` and `code` folder for you. It will search your code and find any occurence of `mysite`. It won't replace those occurence with `app` however.
+The `add-namespace` command of the [upgrader tool](https://github.com/silverstripe/silverstripe-upgrader/) provides a feature
+to namespace your codebase and to automatic update references to those classes.
 
-After installing, run the upgrader reorganise command:
 ```
-cd ~/my-project-root
-~/.composer/vendor/bin/upgrade-code reorganise --write
+composer global require silverstripe/upgrader
+/var/www/SS_acme > upgrade-code add-namespace "Acme\\Web" ./mysite/code --recursive --write
```
chillu: Again, `MyProject` rather than `Acme\\Web`

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+    // ...
+}
 ```
 
-These hints only cover a part of the upgrade work, but can serve as a good indicator for where to start.
+### Enable PSR-4 auto-loading in your `composer.json` file
+If you have namespaced your project and followed the PSR-4 convention, you have the option to enable PSR-4 auto-loading in your composer.json file.
 
-If you've already had a look over the changelog, you will see that there are some fundamental changes that need to be implemented to upgrade from 3.x. Here's a couple of the most important ones to consider:
+Enabling PSR-4 auto-loading is optional. It will provide better auto-loading of your classes in your development environment and will future proof your code for SilverStripe 5.
```
chillu: Remove "for SilverStripe 5", we likely won't have time to remove our class loader for that.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+If you've opted to namespace your own code in the previous step, those references will need to be updated as well.
+
+### Automatically update namespaced references with the `upgrade` command
+
+If you've installed the upgrader, you can use the `upgrade` command to update references to namespaced classes.
+
+The `upgrade` command will update PHP file, YML configuration files, and YML language files.
+
+#### Before running the `upgrade` command
+Each core SilverStripe 4 package includes a `.upgrade.yml` that defines the equivalent fully qualified name of each class. Most third party SilverStripe modules that have been upgraded to be compatible with SilverStripe 4, also include a `.upgrade.yml`.
```
chillu: We call them "modules" rather than "packages"

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+
+
+    public function getShippingCost()
+    {
+        return Injector::inst('ProductService')->calculateCost($this);
+    }
+}
+```
+
+`Image`, `Blog`, and `ProductService` represent classes. Those strings need to be updated to specify the full namespace.
```
chillu: BlogPost

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+}
+```
 
-## Using the upgrader tool
+After switching to SilverStripe 4 recipes, the `composer.json` file should look like this.
+```json
+{
+    "name": "acme/cms-website",
+    "require": {
+        "silverstripe/recipe-cms": "~1.1.0"
```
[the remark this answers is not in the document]
dhensby: Mmmm - as always it comes down to whether we want to be definitive in our projects or we're happy to rely on our knowledge outside the system. We "know" that the recipes are meant to work like this, but I don't see that it makes it any better to rely on implied dependencies regardless that we "know" the recipe won't suddenly drop `versioned`.

I guess it's more a point of preference, but I wouldn't allow a project I'm working on to rely on implied deps; I'd rather sort out a dependency resolution conflict than have a dependent package disappear from my project.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+### Manually upgrading your dependencies
+The instruction in this section assumed you'll be editing your `composer.json` file in a text editor.
+
+#### Switching to _recipes_
+Where possible, we recommend you use recipes.
+
+If your SilverStripe 3 project requires the `silverstripe/cms` package, replace that dependency with `silverstripe/recipe-cms`. Set the version constraint for `silverstripe/recipe-cms` to:
+* `~1.0.0` to upgrade to SilverStripe 4.0
+* `~1.1.0` to upgrade to SilverStripe 4.1
+* `~1.2.0` to upgrade to SilverStripe 4.2
```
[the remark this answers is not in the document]
sminnee: This is completely unintuitive. "To upgrade to SilverStripe 4.0 install recipe 1.0". Find me **one* user who hasn't helped build or maintain the recipe system thinks that this makes sense.

👎 👎 👎 👎 👎 👎 👎 👎 👎

docs/en/03_Upgrading/index.md
maxime-rainville: Done

docs/en/03_Upgrading/index.md
```diff
@@ -10,9 +10,61 @@ your SilverStripe installation means overwriting files, flushing the cache and u
...
+* [Step 9 - Move away from hardcoded paths for referencing static assets](#step9)
+* [Step 10 - Running your upgraded site for the first time](#step10)
+* [List Major API Changes](#MajorAPIChange)
+
+
+## Understanding and planning your upgrade {#planning}
+
+How easy will it be to update my project? It's a fair question, and sometimes a difficult one to answer.
+
+* "Micro" releases (x.y.z) are explicitly backwards compatible, "minor" and "major" releases can deprecate features and change APIs (see our [release process](/contributing/release_process) for details)
```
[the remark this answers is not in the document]
maxime-rainville: That part was a carry over from the old doc.

I've split the first point in a sub-list with one entry for Major, Minor and Patch ... in that order.

docs/en/03_Upgrading/index.md
```diff
@@ -10,9 +10,61 @@ your SilverStripe installation means overwriting files, flushing the cache and u
...
+
+### Overview of changes
+If you've already had a look over the changelog, you will see that there are some fundamental changes that need to be implemented to upgrade from 3.x. Here's a couple of the most important ones to consider:
+
+* PHP 5.6 is now the minimum required version (and up to PHP 7.1.x is supported!).
+* CMS CSS has been re-developed using Bootstrap 4 as a base.
+* SilverStripe code _should_ now be [PSR-2 compliant](http://www.php-fig.org/psr/psr-2/). While this is not a requirement, we strongly suggest you switch over now. You can use tools such as [`phpcbf`](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Fixing-Errors-Automatically) to do most of it automatically.
+* We've also introduced some best practices for module development. [See the Modules article](/developer_guides/extending/modules) for more information.
+
+### Using recipes instead of requiring individual modules
```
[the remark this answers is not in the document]
maxime-rainville: Done.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
 For a description on how to handle issues with pre-existing composer installs or upgrading other modules, please read
 through the [Composer dependency update section](/changelogs/4.0.0#deps)
 
-## Manual upgrades
+### Install the upgrader tool (optional)
+Using the upgrader is not mandatory, but it can speed up the upgrade process. To install the upgrader globally run this command.
 
-* Check if any modules (e.g. `blog` or `forum`) in your installation are incompatible and need to be upgraded as well.
+```bash
+composer global require
```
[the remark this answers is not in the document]
maxime-rainville: Done.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+```
+
+Add your global composer bin directory to your path. On *nix system, this directory is normally located at `$HOME/.composer/vendor/bin`. On Windows system, this directory is normally located at `C:\Users\<COMPUTER NAME>\AppData\Roaming\Composer\vendor\bin`. You can find the exact location by running this command:
+```bash
+composer global config bin-dir
+```
+
+On *nix system, the following command will add your global composer bin directory to your path:
```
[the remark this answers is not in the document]
maxime-rainville: Yes, but this is already quite a detailed heavy guide. If I start explaining to people how to update their path on every possible combination of OS and shell, we'll be here until SS5.

Plus, I can't think of any major distro that doesn't ship with bash as its default shell.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+```bash
+echo 'export PATH=$PATH:~/.composer/vendor/bin/' >> ~/.bash_profile
+```
+
+Each command in the upgrader as somewhat different arguments. However must of them accept these two options:
```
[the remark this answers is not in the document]
maxime-rainville: Done

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
 
 <div class="warning" markdown="1">
 Never update a website on the live server without trying it on a development copy first!
 </div>
 
-## Environment variables file changed to dotenv
+### Branching your project
+Setting a dedicated branch in your source control system to track your upgrade work can help you manage your upgrade.
+
+If you're upgrading a big project, you should consider creating individual branches for each step.
```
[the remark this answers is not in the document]
maxime-rainville: My general preference is to have one distinct idea per paragraph, but I see what you mean.

I'll try to find a reasonable middle.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+
+### Running all the upgrader commands in this guide in on line
+The upgrader comes with an `all-in-one` command. This command will attempt to run all the upgrader commands in the same order as this guide. This is unlikely to work on your first try, but can be a good way to get started without going through this entire guide.
+
+```bash
+/var/www/SS_acme > upgrade-code recompose --recipe-core-constraint=1.1 --namespace="Acme\\Web"
```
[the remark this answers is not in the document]
maxime-rainville: I'll ditched the `/var/www/SS_acme >` and switch to App\Web for my default namespace.

FYI ... all my coyote, road runner and explosive tennis ball code examples are going to be nonsensical now :-P

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+        "silverstripe/reports": "3.6.5@stable",
+        "silverstripe/siteconfig": "3.6.5@stable",
+        "dnadesign/silverstripe-elemental": "^1.8.0"
+    }
+}
+```
 
+You can upgrade the `composer.json` file with this command:
+```bash
+/var/www/SS_acme > upgrade-code recompose --recipe-core-constraint=1.1
```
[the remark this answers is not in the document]
maxime-rainville: The decision for `--write` flag predates me. All things being equal, my view is that non-destructive options should be defaults.

I'll update the instruction to include the `--write` flag and alter the side notes to say _omit the --write flag to preview your changes_.

docs/en/03_Upgrading/index.md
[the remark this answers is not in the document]
maxime-rainville: @dhensby In this context, if we tell people to use `^1.0.0` to upgrade to 4.0, they're going to be upgraded to 4.2 when they run their `composer update`. 

On the broader point of whatever we should encourage `^` or `~`, it's not clear to me we recommend one or the other. My understanding is that government organisation and bigger business generally prefer using `~` to avoid being accidentally upgraded to a different minor release. In my agency days, I generally preferred to use `^` because I was usually dealing with smaller clients.

The upgrader right now will default to using `^`, unless you set the `--strict` flag.

@sminnee From the discussion on the other retagging issue, looks like we're going to retag. I'll update the instruction to use 4 version when require recipe. However this will block this issue until the retagging is done.

docs/en/03_Upgrading/index.md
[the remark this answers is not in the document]
chillu: Yeah, I see you're already doing that in later steps - so it's just consistent to use `--write` in your default instructions on this step as well.

> my view is that non-destructive options should be defaults

So you're using `cp --write` and `git commit --write`? ;) Non-destructive is a very uncommon default for CLI tools in the Unix ecosystem. It's much more common to have a `--dry-run`. But anyway, as you say - the API design decision predates you, let's just make sure there's useful default instructions here.

chillu:
> Shall we hold off merging this until we retag the recipes?

Let's get everything else in here across the line, it's a simple change once we re-tag. But yes, it blocks the merge.

docs/en/03_Upgrading/index.md
```diff
@@ -10,9 +10,61 @@ your SilverStripe installation means overwriting files, flushing the cache and u
...
+## Understanding and planning your upgrade {#planning}
+
+How easy will it be to update my project? It's a fair question, and sometimes a difficult one to answer.
+
+* "Micro" releases (x.y.z) are explicitly backwards compatible, "minor" and "major" releases can deprecate features and change APIs (see our [release process](/contributing/release_process) for details)
+*  If you've made custom branches of SilverStripe core, or any thirdparty module, it's going to be harder to upgrade.
+*  The more custom features you have, the harder it will be to upgrade. You will have to re-test all of those features, and adapt to API changes in core.
+*  Customizations of a well defined type - such as custom page types or custom blog widgets - are going to be easier to upgrade than customisations that modify deep system internals like rewriting SQL queries.
+
+### Overview of changes
```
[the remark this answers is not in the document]
chillu: The end of this section should link to the "list of major api changes" further down in the same document.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+```yaml
+en:
+  SilverStripe\Security\Member:
+    SINGULARNAME: Coyote
+  Acme\Web\RoadRunner:
+    SALUTATION: Beep Beep
+```
+
+### Finalising namespace updates
+Whatever you upgraded your code automatically or manually you'll need to perform the two following steps manually.
```
chillu:
```diff
-Whatever
+Whether
```

chillu: Or actually, more active wording: 

> You'll need to perform the following steps manually, even if you've used the automated rewrite of namespaces previously.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+```
+
+You can omit the `--write` flag if you just want to view the proposed changes without applying them. You can run the command on a specific subdirectory or file. This can be more manageable if you have a big project to go through.
+
+Like the `upgrade` command, `inspect` gets its list of API changes from `.upgrade.yml` files. So you may get upgrade suggestions and substitution from third party modules. You can even include your own project specific changes in your `.upgrade.yml` if you want.
+
+#### Sample output of the `inspect` command
+Here's some sample output of what you might get back the `inspect` command.
+
+```bash
```
chillu: Good idea to provide sample output!

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+Warnings for Coyote.php:
+ - Coyote.php:20 SS_Cache: Using symfony/cache now (https://docs.silverstripe.org/en/4/changelogs/4.0.0#cache)
+ - Coyote.php:42 SilverStripe\Control\Director::setUrlParams(): Method removed
+ - Coyote.php:71 SilverStripe\ORM\DataObject->manyManyComponent(): DataObject->manyManyComponent() moved to DataObjectSchema. Access through getSchema(). You must manually add static::class as the first argument to manyManyComponent()
+Changes not saved; Run with --write to commit to disk
+```
+
+### Manually update deprecated API references
+
+SilverStripe 4 introduces thousands of small and big API changes. To update deprecated API references manually, you have to go through each one of your project files.
```
chillu: Just say "many" rather than "thousands". Remember, the aim here is to make decision makers *more* confident in upgrades, not *less* confident.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+
+## Step 9 - Move away from hardcoded paths for referencing static assets {#step9}
+
+SilverStripe 4 introduce a new way to reference static assets like images and css. This enable innovations like moving SilverStripe module vendor folder or the public web root.
+
+This change is mandatory if you've completed either:
+* Step 7 - Update project structure
+* Step 8 - Switch to public web-root
+
+Otherwise, it is strongley recommended, but not mandatory.
```
chillu: "strongly"

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+
+If your SilverStripe 3 project requires the `silverstripe/framework` package without `silverstripe/cms`, replace `silverstripe/framework` with `silverstripe/recipe-core`. Set the version constraint for `silverstripe/recipe-core` to:
+* `~1.0.0` to upgrade to SilverStripe 4.0
+* `~1.1.0` to upgrade to SilverStripe 4.1
+* `~1.2.0` to upgrade to SilverStripe 4.2
+* and so on.
+
+The following packages are implicitly required by `silverstripe/recipe-core`. They can be removed from your `composer.json` dependencies if you are using `silverstripe/recipe-core` or `silverstripe/recipe-cms`.
+* `sivlerstripe/framework`
+* `silverstripe/config`
```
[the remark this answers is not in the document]
maxime-rainville: It's a lie that tells the truth. Basically I want to communicate that if you use the recipe, you don't need to explicitly install these packages. 

I don't really want to start explaining to people, "this package and this package didn't exist in SS3 because they were part of this other package that got split up, so you won't have them in you SS3 composer file, however if you want to manually define everything you need to include them".

Also, it's not strictly speaking a SS3 to SS4 upgrade guide. In theory you could follow this guide to upgrade from SS4.0 to SS4.2.

docs/en/03_Upgrading/index.md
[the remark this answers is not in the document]
maxime-rainville: The manual process here describes what the upgrader does ... the upgrader right now will try to find equivalent recipes to your package combinations and remove them.

If this part is wrong, then the upgrader is wrong.

docs/en/03_Upgrading/index.md
robbieaverill: Reference: https://github.com/silverstripe/silverstripe-framework/issues/7647

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+}
 ```
-cd ~/my-project-root
-upgrade-code doctor
+
+Looking at the [Packagist entry for `dnadesign/silverstripe-elemental`](https://packagist.org/packages/dnadesign/silverstripe-elemental#2.0.0), you can see that versions 2.0.0 and above of this package are compatible with SilverStripe 4. So you can update that constraint to `^2.0.0`.
+
+Alternatively, you can set a very permissive constraint and let composer find a SilverStripe 4 compatible version. After you're done updating your dependencies, make sure you adjust your constraints to be more specific.
+
+Once you've updated your third-party packages constraints, try updating your dependencies by running `composer update`. If composer can't find all your dependencies it will throw an error.
```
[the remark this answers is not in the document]
maxime-rainville: Done.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+
+### Resolving conflicts
+Whatever you have updated your dependencies with the upgrader or manually, you'll likely have some conflicts to resolve. Running a `composer update` will tell you which package are conflicted and suggested alternative combinations of package that might work.
```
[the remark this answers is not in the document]
maxime-rainville: Done.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+
+The most typical reason for a conflict is that the maintainer of a package has not release a version compatible with SilverStripe 4.
```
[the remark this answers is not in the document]
maxime-rainville: Done.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+
+If the maintainer of the package is in the process of upgrading to SilverStripe 4, a development version of the package might be available. In some cases, it can be worthwhile to look up the repository of the package or to reach out to the maintainer.
+
+To resolve a conflict you can either:
```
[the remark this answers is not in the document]
maxime-rainville: Done.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+* remove the package from your project, if it is not essential
+* fork the package and maintain it yourself.
+
+If you decide to fork a package, remove it from your `composer.json` file and from your `.gitignore` file. Then track the package's codebase in your project source control. Alternatively, you can track the fork as a private  package if you do not which to merge it in your code.
```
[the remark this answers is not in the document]
maxime-rainville: I'll need to think about that one a bit. I think a lot of modules that are not compatible with SS4 are probably abandoned. So doing a PR might not be an option in many cases.

Plus, I don't really want to turn this into a "create your first SS4 module" guide ... while it might be interesting to create that guide later on.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+
+<div class="info" markdown="1">
+If you're taking the time to upgrade a third party package, consider doing a pull request against the original project so other developers can benefit from your work.
+</div>
+
+### Finalising your dependency upgrade
+Once you've resolved all conflicts in your composer.json file, `composer update` will be able to run without errors.
```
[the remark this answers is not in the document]
maxime-rainville: Done.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+
+If you've decided to use recipes, some generic files will be copied from the recipe into your project. The `extra` attribute in your `composer.json` file will be updated to keep track of those new files.
+
+This is a good point to commit your changes to your source control system before moving on to the next step.  
+
+
+## Step 2 - Update your environment configuration {#env}{#step2}
+
+The php configuration `_ss_environment.php` file has been replaced in favour of a non-executable
+`.env` file, which follows a syntax similar to a `.ini` file for key/value pair assignment. Your `.env` file may be placed in either the web root, or your project root.
```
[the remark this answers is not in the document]
maxime-rainville: I think I got it from the changelog https://docs.silverstripe.org/en/4/changelogs/4.0.0/#env

> Like the old php file, .env may be placed in either the web root, or one level above.

I remember being confused when I came across this part as well. I took it to say that your `.env` file can be in your web root or your project root. But since this is specific to 4.0 the web root and your project root will always be the same.

I changed it to say:
> Your `.env` file may be placed in your project root, or one level above your project root.

docs/en/03_Upgrading/index.md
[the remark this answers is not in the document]
dhensby: Apologies, I'm not suggesting we list many ways to achieve the same thing; I'm suggesting the comment about "*nix" systems be a bit more accurate.

You could even link to the composer docs and/or the wikipedia article as they do (https://getcomposer.org/doc/00-intro.md#globally)

docs/en/03_Upgrading/index.md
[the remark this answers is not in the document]
dhensby: OK - then I just think the language needs to change; rather than using the word "forking" (which to me implies public maintaining) I'd just say to copy the code into their own codebase and work on it.

Still, I'd put this below recommending raising a PR to upgrade a 3rd party package

docs/en/03_Upgrading/index.md
unclecheese: > Whatever you upgraded

Whether you upgraded?

docs/en/03_Upgrading/index.md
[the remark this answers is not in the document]
maxime-rainville: I've updated it to say:
> On *nix system, the following command will add your global composer bin directory to your path if `bash` is your default shell environment

docs/en/03_Upgrading/index.md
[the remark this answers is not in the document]
maxime-rainville: Good suggestion. I've differentiated between forking a module and integrating a module into your own codebase.

I've linked to our _How to publish a module_ doc page.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+</div>
+
+### Finalising your dependency upgrade
+Once you've resolved all conflicts in your composer.json file, `composer update` will be able to run without errors.
+
+This will install your new dependencies. You'll notice many of the folders in the root of your project will disappear. That's because SilverStripe 4 packages can be installed in the vendor folder like generic PHP packages.
+
+If you've decided to use recipes, some generic files will be copied from the recipe into your project. The `extra` attribute in your `composer.json` file will be updated to keep track of those new files.
+
+This is a good point to commit your changes to your source control system before moving on to the next step.
```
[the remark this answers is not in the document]
maxime-rainville: I'd rather keep it has part of the main content. I think pull out are better for asides that aren't directly related to the upgrading process.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+### Cleaning up `mysite/_config.php` after your environment configuration upgrade
+
+You'll need to clean up your `mysite/_config.php` file after upgrading your environment file.
+
+The global values `$database` and `$databaseConfig` have been deprecated. Your database configuration details should be stored in your `.env` file. If you want to keep your database configuration in `_config.php`, you can use the new `DB::setConfig()` api, however this is discouraged.
+
+Requiring `conf/ConfigureFromEnv.php` is is no longer necessary. You should remove any references to it in `_config.php`.
+
+The removal of the `_ss_environment.php` file means that conditional logic is no longer available in the environment
+variable set-up process. This encouraged bad practice and should be avoided. If you still require conditional logic early in the bootstrap, this is best placed in the `_config.php` files.
```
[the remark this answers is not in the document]
maxime-rainville: I think that part got copied over from the existing doc.

I would rather discourage people editing their `index.php` file. If you need to high jack the bootstrapping logic you're probably doing something dodgy.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+* Read the [official Namespace PHP documentation](http://php.net/manual/en/language.namespaces.php)
+* Read the [PSR-4: Autoloader standard](https://www.php-fig.org/psr/psr-4/).
+
+### Before you start namespacing your codebase
+
+You need to choose a root namespace for your project. We recommend following the `Vendor\Package` pattern.
+
+The `Page` and `PageController` classes *must* be defined in the global namespace (or without a namespace).
+
+If you want your codebase to comply with the PSR-4 standard, make sure sub-directories of of your code folder are using the _upper case camel_ naming convention. For example, `mysite/code/page_types` should be renamed to `mysite/code/PageTypes`.
```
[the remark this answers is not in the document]
maxime-rainville: Upper Case Camel it is. 

I would avoid describing this as _title case_ to avoid confusion with the [heading writing style](http://www.grammar-monster.com/glossary/title_case.htm).

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+
+Note that applying a namespace to your project will also affect which template file SilverStripe tries to load when rendering certain objects.
+
+For example, pretend you have a `RoadRunnerPage` class that extends `Page`. In SilverStripe 3, you would define a template for this page in `themes/acme/templates/Layout/RoadRunnerPage.ss`. If you decide to move `RoadRunnerPage` to `Acme\Web\RoadRunnerPage`, you'll need to move the template to `themes/acme/templates/Acme/Web/Layout/RoadRunnerPage.ss`.
+
+This is a good point to commit your changes to your source control system before moving on to the next step.
+
+## Step 4 - Update codebase with references to newly namespaced classes {#step4}
+
+Nearly all core PHP classes in SilverStripe 4 have been namespaced. For example, `DataObject` is now called `SilverStripe\ORM\DataObject`. Your project codebase, config files and language files need be updated to reference those newly namespaced classes. This will include explicit references in your PHP code, but also string that contain the name of a class.
```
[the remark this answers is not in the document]
maxime-rainville: That bit got copied over, so not quite sure what nearly refers to in this context. 

There's some weird things under `thirdparty` that aren't namespaced, but I wouldn't call those core classes.

All drop the nearly to avoid confusing people.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+
+If you've opted to namespace your own code in the previous step, those references will need to be updated as well.
+
+### Automatically update namespaced references with the `upgrade` command
+
+If you've installed the upgrader, you can use the `upgrade` command to update references to namespaced classes.
+
+The `upgrade` command will update PHP file, YML configuration files, and YML language files.
```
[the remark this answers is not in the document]
maxime-rainville: Done

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+```
+
+If you omit the `--write` flag you will get a preview of what change the upgrader will apply to your codebase. This can be helpful if you if you are tweaking your `.upgrade.yml` or if you are trying to identify areas where you should add a `@skipUpgrade` statement,
+
+You can also tweak which rules to apply with the `--rule` flag. There's 3 options that can be provided: `code`, `config`, and `lang`. For example, the following command will only upgrade `lang` and `config` files:
+```bash
+/var/www/SS_acme > upgrade-code upgrade ./mysite/ --rule=config --rule=lang
+```
+
+The `upgrade` command can alter big chunk of your codebase. While it works reasonably well in most use case, you should not trust it blindly. You should take time to review all changes applied by the `upgrade` command and confirm you are happy with them.
```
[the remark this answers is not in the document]
maxime-rainville: Fix.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+
+### Manually update namespaced references
+
+Basically, you'll need to go through the entire code base and update all references from the old non-namespaced SilverStripe classes to new namespaced equivalent. If you are referencing classes from third party packages that have been namespaced, you'll need to update those as well.
```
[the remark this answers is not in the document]
maxime-rainville: Done.

docs/en/03_Upgrading/index.md
maxime-rainville: Done.

docs/en/03_Upgrading/index.md
maxime-rainville: Done.

docs/en/03_Upgrading/index.md
maxime-rainville: I've added the _database_ table reference.


@dhensby Your comment about the upgrader not adding the  `private static $table_name` reference is relevant. From memory SS4.0 wouldn't warn you if you didn't specify  a `private static $table_name`, but SS4.1 does.

Might be worthwhile adding it to the upgrade command.

maxime-rainville: Added an issue about the upgrader https://github.com/silverstripe/silverstripe-upgrader/issues/104

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +94,1028 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
+```
+
+### Any other script that needs running.
+
+Some third party modules may include their own migration tasks. Take a minute to consult the release notes of your third party dependencies to make sure you have miss anything.
+
+## List of Major API Changes {#MajorAPIChange}
+This is a list of the most common API changes that might affect you.
+
+(Read the SilverStripe 4.0.0 change logs)[/changelogs/4.0.0/] for a comprehensive list of eerything that his new in SilverStripe 4.
```
chillu: Should link to 4.0.0, 4.1.0 and 4.2.0 (only minor, not patch releases). All of those are noteworthy

docs/en/03_Upgrading/index.md
```diff
@@ -10,9 +10,66 @@ your SilverStripe installation means overwriting files, flushing the cache and u
...
+### Overview of changes
+
+If you've already had a look over the changelog, you will see that there are some fundamental changes that need to be implemented to upgrade from 3.x. Here's a couple of the most important ones to consider:
+
+* PHP 5.6 is now the minimum required version (and up to PHP 7.2.x is supported!).
+* CMS CSS has been re-developed using Bootstrap 4 as a base.
+* SilverStripe code _should_ now be [PSR-2 compliant](http://www.php-fig.org/psr/psr-2/). While this is not a requirement, we strongly suggest you switch over now. You can use tools such as [`phpcbf`](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Fixing-Errors-Automatically) to do most of it automatically.
+* We've introduced some best practices for module development. [See the Modules article](/developer_guides/extending/modules) for more information.
+
+[Read the SilverStripe 4.0.0 changelogs](/changelogs/4.0.0/) for a comprehensive list of changes introduced with SilverStripe 4.
```
chillu: Should link to 4.1.0 and 4.2.0 as well

docs/en/03_Upgrading/index.md
```diff
@@ -10,9 +10,66 @@ your SilverStripe installation means overwriting files, flushing the cache and u
 See our [upgrade notes and changelogs](/changelogs/4.0.0) for 4.0.0 specific information, bugfixes and API changes.
 </div>
 
-## Composer
+## Understanding and planning your upgrade {#planning}
```
chillu: The detailed changelogs (4.0.0, 4.1.0 and 4.2.0) should all point back to this upgrading guide with clear wording. Something like:

> This changelog describes API changes in detail. Use our [step-by-step guide for upgrading to SilverStripe 4](...) to perform an upgrade on your module or codebase.

docs/en/03_Upgrading/index.md
```diff
@@ -10,9 +10,66 @@ your SilverStripe installation means overwriting files, flushing the cache and u
...
+
+* SilverStripe follows _semantic versioning_ (see our [release process](/contributing/release_process) for details).
+  * "Major" releases introduces API change that may break your application.
+  * "Minor" releases (x.y) introduces API changes in a backward compatible way and can mark some API as deprecated.
+  * "Patch" releases (x.y.z) fix bugs without introducing any API changes.
+* If you've made custom branches of SilverStripe core, or any thirdparty module, it's going to be harder to upgrade.
+* The more custom features you have, the harder it will be to upgrade. You will have to re-test all of those features, and adapt to API changes in core.
+* Customizations of a well defined type - such as custom page types or custom blog widgets - are going to be easier to upgrade than customisations that modify deep system internals like rewriting SQL queries.
+
+### Overview of changes
```
chillu: I assume you're still working on consolidating this with the "list of major API changes" further down? There should be one list in the upgrading guide (for highlevel technical decision makers), and then a detailed list in the individual changelog documents.

docs/en/03_Upgrading/index.md
maxime-rainville: Not sure I'm familiar with this. Are config value cached?

docs/en/03_Upgrading/index.md
maxime-rainville: Done

docs/en/03_Upgrading/index.md
maxime-rainville: Done.

docs/en/03_Upgrading/index.md
maxime-rainville: I think it's the only place I provide an exact console output. Most other places, I did a before-and-after.

docs/en/03_Upgrading/index.md
maxime-rainville: Done

docs/en/03_Upgrading/index.md
maxime-rainville: The main reason why I mention the entry point has changed is so people who have hadcoded their rewrite rule into their VHOST config rather than rely on `.htaccess` no what they need to update.

While I agree it would be very unusual for someone to update their `index.php` file in SS3, it's a scenario that the webroot command handles, so it seems right to mention it in the manual process.

I've rename the header to _Update your index.php file_ to avoid implying that `index.php` was the entry point in SS3.

docs/en/03_Upgrading/index.md
maxime-rainville: Not sure if the official recommendation came with 4.0 or 4.1. In practice, it will work in either case.

I changed it to `SilverStripe 4 introduces ... `

docs/en/03_Upgrading/index.md
maxime-rainville: Fixed.

docs/en/03_Upgrading/index.md
maxime-rainville: Added references to 4.0.0, 4.1.0 and 4.2.0.

docs/en/03_Upgrading/index.md
```diff
@@ -10,9 +10,61 @@ your SilverStripe installation means overwriting files, flushing the cache and u
...
+It can also be use to upgrade your existing SilverStripe 4 to a newer minor release.
+
+## Step 0 - Pre-requisites and background work {#step0}
+
+Before you begin the upgrade process, make sure you meat these pre-requisites
+
+### Install composer
+Most of the steps in this guide assume you are tracking your project dependencies through composer.
+
+While it is possible to upgrade to SilverStripe 4 without using composer, you will save a lot of time if you do.
```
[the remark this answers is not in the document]
maxime-rainville: I'll need to confirm this with @tractorcow

docs/en/03_Upgrading/index.md
maxime-rainville: I'll call this one a nice to have.

docs/en/03_Upgrading/index.md
```diff
@@ -10,9 +10,61 @@ your SilverStripe installation means overwriting files, flushing the cache and u
...
+* "Micro" releases (x.y.z) are explicitly backwards compatible, "minor" and "major" releases can deprecate features and change APIs (see our [release process](/contributing/release_process) for details)
+*  If you've made custom branches of SilverStripe core, or any thirdparty module, it's going to be harder to upgrade.
+*  The more custom features you have, the harder it will be to upgrade. You will have to re-test all of those features, and adapt to API changes in core.
+*  Customizations of a well defined type - such as custom page types or custom blog widgets - are going to be easier to upgrade than customisations that modify deep system internals like rewriting SQL queries.
+
+### Overview of changes
+If you've already had a look over the changelog, you will see that there are some fundamental changes that need to be implemented to upgrade from 3.x. Here's a couple of the most important ones to consider:
+
+* PHP 5.6 is now the minimum required version (and up to PHP 7.1.x is supported!).
+* CMS CSS has been re-developed using Bootstrap 4 as a base.
```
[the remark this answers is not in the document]
maxime-rainville: I've updated it with a list of things that I think are more important. But that part should be reviewed.

docs/en/03_Upgrading/index.md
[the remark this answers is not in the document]
maxime-rainville: Added a link

docs/en/03_Upgrading/index.md
dhensby: >I've renamed the header to Update your index.php file to avoid implying that index.php was the entry point in SS3.

But it wasn't... I'm worried we'll confuse 95% of readers just to cater to 5% that have custom routing of requests...

docs/en/03_Upgrading/index.md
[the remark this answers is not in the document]
maxime-rainville: That's what I mean by _All things being equal_. If you have a command that is inherently destructive by its nature, there's no point requiring the `--write` flag.

I would argue that previewing your changes is an integral part of the upgrader.

docs/en/03_Upgrading/index.md
```diff
@@ -37,123 +89,1015 @@ and also checking the `composer.json` files in [recipe-core](https://github.com/
...
 
-Read the [upgrader `environment` command documentation](https://github.com/silverstripe/silverstripe-upgrader/blob/master/docs/en/environment.md)
-for more details.
+If the `recompose` command can't find a SilverStripe 4 compatible version for one of your package, it will keep this dependency in your `composer.json` file with its existing constraint.
+
+### Manually upgrading your dependencies
+The instruction in this section assumed you'll be editing your `composer.json` file in a text editor.
+
+#### Switching to _recipes_
+Where possible, we recommend you use recipes.
```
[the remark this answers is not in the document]
maxime-rainville: Couldn't find a neat article explaining what recipes are and it's not clear that the upgrade guide is the best place to do this.

docs/en/03_Upgrading/index.md
maxime-rainville: The upgrader doesn't like single NS space. Not sure if it's by design or if it's a bug.

I've switch to `App\Web` for now. And I'll discuss with @tractorcow tomorrow.

docs/en/03_Upgrading/index.md
maxime-rainville: I've removed most references of `package` except for those that directly reference non-SilverStripe package on packagist.

docs/en/03_Upgrading/index.md
maxime-rainville: I've linked to the list of important API changes at the end.

docs/en/03_Upgrading/index.md
maxime-rainville: That's been updated with what strikes me as the most important points.

docs/en/03_Upgrading/index.md
maxime-rainville: Done.

maxime-rainville:
I think I've got all the big points sorted. There's a few things I would like to clear:

* Need to confirm if we're merging with the 1.x recipe versions or the the 4.x recipe versions
* @chillu mentioned that should updating config value in `_config.php` will slow execution down ... need some clarification on that
* Need to find out if we have an article explaining recipes that I can link to
* Need to raise a bug on the upgrader failing when adding namespace with only a _vendor name_.

chillu:
Alright, @maxime-rainville I couldn't help myself and done some tweaks. Tried to keep them as isolated commits. Can you review please? I think we're pretty much there now, unless you have anything else.

docs/en/03_Upgrading/index.md
```diff
@@ -1026,22 +1020,19 @@ If you are upgrading from SilverStripe 4.0 to SilverStripe 4.1 (or above), you'l
...
 
 ## Step 9 - Move away from hardcoded paths for referencing static assets {#step9}
 
-SilverStripe 4 introduces a new way to reference static assets like images and css. This enable innovations like moving SilverStripe module vendor folder or the public web root.
-
-This change is mandatory if you've completed either:
-* Step 7 - Update project structure
-* Step 8 - Switch to public web-root
+SilverStripe 4 introduces a new way to reference static assets like images and css. 
+This enable innovations like moving SilverStripe module vendor folder or the public web root ([details]())
```
maxime-rainville: Missing link

docs/en/03_Upgrading/index.md
```diff
@@ -14,28 +14,31 @@ How easy will it be to update my project? It's a fair question, and sometimes a
   * "Major" releases introduces API change that may break your application.
   * "Minor" releases (x.y) introduces API changes in a backward compatible way and can mark some API as deprecated.
   * "Patch" releases (x.y.z) fix bugs without introducing any API changes.
-* If you've made custom branches of SilverStripe core, or any thirdparty module, it's going to be harder to upgrade.
-* The more custom features you have, the harder it will be to upgrade. You will have to re-test all of those features, and adapt to API changes in core.
-* Customizations of a well defined type - such as custom page types or custom blog widgets - are going to be easier to upgrade than customisations that modify deep system internals like rewriting SQL queries.
+* If you've made custom branches of SilverStripe core, or any thirdparty module, upgrades are going to be more complex.
+* More custom features mean more work to re-test all of those features, and adapt to API changes in core.
+* Customizations of a well defined type - such as custom page types or custom blog widgets -
```
maxime-rainville: Should be "Customisations"

docs/en/03_Upgrading/index.md
```diff
@@ -464,22 +458,19 @@ If you want your codebase to comply with the PSR-4 standard, make sure sub-direc
...
 
 This task will do the following:
 * Add the given namespace to all files in the code class, and subdirectories
 * All references to classes in any namespaced files will be safely retained with additional `use` directives added as necessary
-* A `mysite/.upgrade.yml` file be created/updated to record the new fully qualified name of each class.
-
-`.upgrade.yml` will be used in later steps to update references to the old non-namespaced classes.
-
-If you want to do a dry-run, omit the `--write` option to see a preview of all changed project files.
+* A `mysite/.upgrade.yml` file be created/updated to record the new fully qualified name of each class.
```
maxime-rainville: missing a `will` in "file will be created/updated"

docs/en/03_Upgrading/index.md
```diff
@@ -373,9 +369,9 @@ If you have installed the upgrader tool, you can use the `environment` command t
 upgrade-code environment --write
 ```
 
-If your `_ss_environment.php` file contains unusual logic (conditional statements or loops), you will get a warning. `upgrade-code` will still try to convert the file, but you should double-check the output.
+If your `_ss_environment.php` file contains unusual logic (conditional statements or loops), you will get a warning. `upgrade-code` will still try to convert the file, but you should double-check the output. Omit the `--write` flag to do a dry-run.
 
-Omit the `--write` flag to do a dry-run.
+[Continue to "Cleaning up `mysite/_config.php`"](env-config-cleanup)
```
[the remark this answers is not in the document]
chillu: Fixed

docs/en/03_Upgrading/index.md
```diff
@@ -923,9 +913,10 @@ At this stage, you could in theory run your project in SilverStripe 4.
 This is a good point to commit your changes to your source control system before moving on to the next step.
 
 ## Step 7 - Update project structure (optional) {#step7}
-SilverStripe 4 introduces a new recommended project structure. Adopting the recommended project structure is optional, but will become mandatory in SilverStripe 5.
+SilverStripe 4 introduces a new recommended project structure ([details](/changelogs/4.2.0##app-folder-name)).
```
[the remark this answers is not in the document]
chillu: Fixed

docs/en/03_Upgrading/index.md
chillu: Fixed

docs/en/03_Upgrading/index.md
chillu: Fixed

docs/en/03_Upgrading/index.md
chillu: Fixed

PR #8146, opened 2018-06-07
BUG Fix crash on ID in default_sort

tractorcow:
Not sure why this popped up recently.

Requesting a field from db('field') shouldn't return the entire list if it's not found.

ID fails because it's a fixed_fields, and is already indexed.

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -1959,7 +1959,10 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 				$items = isset($items) ? array_merge((array) $items, $dbItems) : $dbItems;
 			}
 		}
-
+		// If we requested a non-existant named field return null instead of all fields
+		if ($fieldName) {
+			return null;
+		}
 		return $items;
 	}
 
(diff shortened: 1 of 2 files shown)


robbieaverill:
Related: https://github.com/silverstripe/silverstripe-framework/pull/8090

model/DataObject.php
```diff
@@ -1959,7 +1959,9 @@ public function db($fieldName = null) {
 				$items = isset($items) ? array_merge((array) $items, $dbItems) : $dbItems;
 			}
 		}
-
+		if ($fieldName) {
```
chillu: Maybe add a comment here on *why* its returning? I had to read through the code twice to figure out that somewhere above it, there's a deeply nested `return` for this case heh

tractorcow: ok.

model/DataObject.php
```diff
@@ -3433,7 +3435,10 @@ public function databaseIndexes() {
 					if ($table && strtolower($table) !== strtolower($this->class)) {
 						continue;
 					}
-
+					// Skip primary key; Already indexed
+					if ($column === 'ID') {
```
chillu: I don't understand why "already indexed" is an exclusion criteria here. Wouldn't you expect `ID` to be part of the returned set?

tractorcow: Because you shouldn't add a second index when the ID is already the primary key.

tractorcow: Let me look into it more.

tractorcow: Initial investigation is that this is a pain to fix for all fixed_fields unless I add heaps more code, and we don't really need to return ID as an indexed field (it's hard-coded as an index in 3.x anyway).

tractorcow: ok did some more work. Should be ok now.

tests/model/DataObjectTest.php
```diff
@@ -1857,6 +1857,8 @@ class DataObjectTest_Fixture extends DataObject implements TestOnly {
 		'DateField.Nice' => 'Date'
 	);
 
+	private static $default_sort = '"DataObjectTest_Fixture"."ID" ASC';
```
tractorcow: No complex test, just "doesn't crash" is enough for this feature.

tractorcow:
It's green ready to merge. :)

PR #8147, opened 2018-06-07
FIX: Only set MYSQL_ATTR_INIT_COMMAND when using mysql driver (fixes #8103)

### LINKED ISSUE #8103: php_pdo_mysql extension must be enabled for PDOConnector to work
In order to use `PDOConnector` you _must_ enable the `php_pdo_mysql` extension - even if you’re not using a MySQL database:

https://github.com/silverstripe/silverstripe-framework/blob/db610aaf3bef897a1f3b5c60a307ee72a5d2a499/src/ORM/Connect/PDOConnector.php#L179-L181

If the extension isn’t enabled, `PDO::MYSQL_ATTR_INIT_COMMAND` isn’t defined.

Simplest fix is probably to wrap that in an `if (defined('PDO::MYSQL_ATTR_INIT_COMMAND'))`, but a much nicer fix would be to implement @tractorcow’s suggestion here: https://github.com/tractorcow/silverstripe-framework/issues/18

kinglozzer:
Looks like this was actually fixed in https://github.com/silverstripe/silverstripe-framework/pull/6612, but must have been lost during the merge up to 4.x

### DIFF
--- a/src/ORM/Connect/PDOConnector.php
+++ b/src/ORM/Connect/PDOConnector.php
@@ -176,9 +176,11 @@ class PDOConnector extends DBConnector
         if (!isset($charset)) {
             $charset = $connCharset;
         }
-        $options = array(
-            PDO::MYSQL_ATTR_INIT_COMMAND => 'SET NAMES ' . $charset . ' COLLATE ' . $connCollation
-        );
+
+        $options = [];
+        if ($parameters['driver'] === 'mysql') {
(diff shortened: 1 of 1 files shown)


dhensby:
Hmmm, there's a bunch of unexpected limiting issues... Has php_cs released another breaking change :P

tractorcow:
Maybe a bad merge somewhere?

tractorcow:
4.0 has been fixed for linting issues... maybe a re-open will fix the build also.

tractorcow:
Merge on green.

PR #8148, opened 2018-06-07
API: <% loop %> and <% with %> only ever create one new scope level (fixes #4015)

### LINKED ISSUE #4015: .Up not finding correct scope in certain circumstances

kinglozzer:
Removes up an age-old quirk where every `.` in `<% loop $Foo.Bar.Baz %>` would require an extra `$Up` to “escape” from. That was actually a feature, but I’m yet to see compelling use-case for it and it keeps causing confusion (see #4015).

Yes this will probably cause breakages for people upgrading, but they’re going to be easy ones to fix. I’ve changed the `user_error()` to an exception, because that way it’s at least possible to see the name of the template that’s affected (even if only the “compiled” line number is shown).

3.x and 4.x:
```html
{$Title}		<!-- Current page title -->
<% loop $Children.Limit(5) %>
    {$Title		<!-- Child page title -->
    {$Up.Title}		<!-- Nothing - equivalent to `$Children.Title` -->
    {$Up.Up.Title}	<!-- Current page title -->
<% end_loop %>
```

After this change:
```html
{$Title}		<!-- Current page title -->
<% loop $Children.Limit(5) %>
    {$Title		<!-- Child page title -->
    {$Up.Title}		<!-- Current page title -->
    {$Up.Up.Title}	<!-- Exception: Up called when we're already at the top of the scope -->
<% end_loop %>
```

### DIFF
--- a/src/View/SSViewer_Scope.php
+++ b/src/View/SSViewer_Scope.php
@@ -192,7 +192,7 @@ class SSViewer_Scope
         switch ($name) {
             case 'Up':
                 if ($this->upIndex === null) {
-                    user_error('Up called when we\'re already at the top of the scope', E_USER_ERROR);
+                    throw new \LogicException('Up called when we\'re already at the top of the scope');
                 }
 
                 list(
@@ -258,6 +258,9 @@ class SSViewer_Scope
         $this->popIndex = $this->itemStack[$newLocalIndex][SSViewer_Scope::POP_INDEX] = $this->localIndex;
         $this->localIndex = $newLocalIndex;
 
+        // $Up now becomes the parent scope - the parent of the current <% loop %> or <% with %>
+        $this->upIndex = $this->itemStack[$newLocalIndex][SSViewer_Scope::UP_INDEX] = $this->popIndex;
+
         // We normally keep any previous itemIterator around, so local $Up calls reference the right element. But
         // once we enter a new global scope, we need to make sure we use a new one
         $this->itemIterator = $this->itemStack[$newLocalIndex][SSViewer_Scope::ITEM_ITERATOR] = null;
@@ -291,7 +294,7 @@ class SSViewer_Scope
 
         if (!$this->itemIterator) {
             // Note: it is important that getIterator() is called before count() as implemenations may rely on
-            // this to efficiency get both the number of records and an iterator (e.g. DataList does this)
+            // this to efficiently get both the number of records and an iterator (e.g. DataList does this)
 
             // Item may be an array or a regular IteratorAggregate
             if (is_array($this->item)) {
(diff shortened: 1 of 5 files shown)


tractorcow:
I need to merge up fixes for master behat.

tractorcow:
I've merged up a couple fixes for both testsession + behat-extension. Let's see if that gets the tests running properly.

src/View/SSViewer_Scope.php
```diff
@@ -192,7 +192,7 @@ public function obj($name, $arguments = [], $cache = false, $cacheName = null)
         switch ($name) {
             case 'Up':
                 if ($this->upIndex === null) {
-                    user_error('Up called when we\'re already at the top of the scope', E_USER_ERROR);
+                    throw new \LogicException('Up called when we\'re already at the top of the scope', E_USER_ERROR);
```
kinglozzer: I should probably remove that `E_USER_ERROR`...

tractorcow:
I've merged up all the branches for core modules to master. Let's see if this gets the tests green.

tractorcow:
Tests are green and the code looks ok, but I really need to do some testing and research to check there are no regressions.

tractorcow:
FYI I used this template to test and it appeared to produce much more expected output.

```html
<% include SideBar %>
<div class="content-container unit size3of4 lastUnit">
	<article>
		<h1>$Title</h1>
		<div class="content">$Content</div>
        <% if $Parent %>
            <h2>all siblings</h2>
            <% loop Parent.Children %>
                <div>Menu: $Title, Up title: $Up.Title</div>

                <h3>Nested</h3>
                <% loop Parent.Children %>
                    <div>Menu: $Title, Up title: $Up.Title, Up up title: $Up.Up.Title</div>
                <% end_loop %>
                <h3>end nested</h3>
            <% end_loop %>
            <h2>first sibling</h2>
            <% loop Parent.Children.Limit(1) %>
                <div>Menu: $Title, Up title: $Up.Title</div>
            <% end_loop %>
        <% end_if %>
	</article>
</div>

```

tractorcow:
The logic goes over my head a little, but the effect seems to have resolved the usability issue. I'm inclined to merge this and hope that the future will be kind. ;)

PR #8152, opened 2018-06-08
API: Preserve variable types in template lookup args (closes #4822)

### LINKED ISSUE #4822: BUG: Boolean template variables evaluate true unless numeric
Apparently all method arguments in templates are passed as strings?
```
$Method(400, .5, false)
```
`func_get_args()` gives:
```
  0 => string '400' (length=3)
  1 => string '.5' (length=2)
  2 => string 'false' (length=5)
```
PHP's type juggling means the first 3 arguments work fine but the booleans will be evaluated as true.

Suggest that if template arguments are not quoted, we only wrap them in quote marks if they pass a test like `!(is_numeric($arg) || $arg == 'true' || $arg == 'false')`.

I think this needs to be implemented in `SSTemplateParser` but I don't know where or how. Can someone smarter than me have a look?

kinglozzer:


src/View/SSTemplateParser.peg
```diff
@@ -405,6 +406,14 @@ class SSTemplateParser extends Parser implements TemplateParser
 
     QuotedString: q:/['"]/   String:/ (\\\\ | \\. | [^$q\\])* /   '$q'
 
+    # Booleans
+
+    Boolean: /(true|false)/
+
+    # Integers and floats
+
+    IntOrFloat: / (\d+(\.\d+)?) /
```
kinglozzer: Doh. This needs to be updated to support passing things like `.5`, `-10` etc. There are other expressions that are valid numbers in PHP (exponents, hexadecimal, binary), not sure if the effort/reward will be worth it for those - I’ll look into it

src/View/SSTemplateParser.peg
```diff
@@ -286,7 +286,8 @@ class SSTemplateParser extends Parser implements TemplateParser
 
         $property = $sub['Call']['Method']['text'];
 
-        if (isset($sub['Call']['CallArguments']) && $arguments = $sub['Call']['CallArguments']['php']) {
+        if (isset($sub['Call']['CallArguments']) && isset($sub['Call']['CallArguments']['php'])) {
+            $arguments = $sub['Call']['CallArguments']['php'];
```
kinglozzer: Change required as sometimes the arguments might simply be an int `0` or boolean `false`

src/View/ViewableData.php
```diff
@@ -412,7 +412,7 @@ public function renderWith($template, $customFields = null)
     protected function objCacheName($fieldName, $arguments)
     {
         return $arguments
-            ? $fieldName . ":" . implode(',', $arguments)
+            ? $fieldName . ":" . var_export($arguments, true)
```
kinglozzer: Required as we can now pass actual booleans as arguments. Implode will cast `false` to an empty string, which would mean the same cache name between `$Foo` and `$Foo(false)`

dhensby:
wow - this is cool, I love it. So if you now want the string "true" it needs to be quoted in the template, right?

I guess there's a distinction between `"1"` and `1` too?

One other feature I'd like to see is the template throwing warnings or errors if there's some kind of unquoted string - would that be much work?

kinglozzer:
> So if you now want the string "true" it needs to be quoted in the template, right?
> I guess there's a distinction between "1" and 1 too?

Correct on both counts

> One other feature I'd like to see is the template throwing warnings or errors if there's some kind of unquoted string - would that be much work?

I think that should be pretty simple, we could trigger a deprecation notice or something. The counter argument is that as long as it’s not blocking any new features we want to add, why bother removing support for it? I personally prefer to be explicit about strings (and the same with remembering the `$` in `<% loop $Foo %>`), but I’m sure there are others out there that prefer the safety net of not having to worry about it.

dhensby:
I guess I like the idea of being able to define a strict template syntax and it feels like that's the way we are heading anyway...

tractorcow:
@kinglozzer do you want to be the new ss template engine sage? Because that's what it sounds like. ;) Somewhere a snowy mountain top is awaiting your presence.

PR #8154, opened 2018-06-09
fix Versioned::choose_site_stage() if no request given

xini:
Load request from current controller as a fallback if none is given.

Several modules break because previous version of choose_site_stage() didn't need the request.

### DIFF
--- a/model/Versioned.php
+++ b/model/Versioned.php
@@ -1216,7 +1216,7 @@ class Versioned extends DataExtension implements TemplateGlobalProvider {
 	 */
 	public static function choose_site_stage(SS_HTTPRequest $request = null) {
 		if (!$request) {
-			throw new InvalidArgumentException("Request not found");
+			$request = Controller::curr()->getRequest();
 		}
 		$mode = static::get_default_reading_mode();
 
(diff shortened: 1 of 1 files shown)


model/Versioned.php
```diff
@@ -1216,7 +1216,7 @@ public static function can_choose_site_stage($request) {
 	 */
 	public static function choose_site_stage(SS_HTTPRequest $request = null) {
 		if (!$request) {
-			throw new InvalidArgumentException("Request not found");
+			$request = Controller::curr()->getRequest();
```
tractorcow: We should check `Controller::has_curr()` first since there isn't always a current controller.

xini:
@tractorcow , I have added that check. Thanks!

tractorcow:
Not sure if test failure is related. Re-starting test on 3.7 branch at https://travis-ci.org/silverstripe/silverstripe-framework/builds/389503339 to confirm.

tractorcow:
It looks like, somehow, this broke something? I can't see how though.

tractorcow:
Restarting this build. Maybe it was just an occasional error.

dhensby:
Green

PR #8156, opened 2018-06-11
PATCH: removing duplicate key (SS_TRUSTED_PROXY_HOST_HEADER)

sunnysideup:
we are removing duplicates key from yml as this causes errors in symphony


tractorcow:
Merged in manually with https://github.com/silverstripe/silverstripe-framework/commit/8f03c7df050979c8ec65efb8d6a328fbd427c68f

PR #8157, opened 2018-06-11
Ensure no cli requests get forced to HTTPS

jakedaleweb:
This resolves an issue where it is possible to break CLI commands by attempting to route them to HTTPS. For example if you have a global `Director::ForceSSL()` you break dev/builds via CLI.

Notably this is causing issues in the CWP 2.0 release which forces SSL in specific places. This means that dev/builds _cannot_ complete during deployments resulting in broken sites.

### DIFF
--- a/src/Control/Middleware/CanonicalURLMiddleware.php
+++ b/src/Control/Middleware/CanonicalURLMiddleware.php
@@ -349,6 +349,10 @@ class CanonicalURLMiddleware implements HTTPMiddleware
      */
     protected function isEnabled()
     {
+        if (Director::is_cli()) {
+            return false;
+        }
+
         // At least one redirect must be enabled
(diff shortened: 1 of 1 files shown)


jakedaleweb:
cc @robbieaverill

tractorcow:
My suggested fix at https://github.com/silverstripe/silverstripe-framework/pull/8158

src/Control/Middleware/CanonicalURLMiddleware.php
```diff
@@ -349,6 +349,10 @@ public function setEnabledEnvs($enabledEnvs)
      */
     protected function isEnabled()
     {
+        if (Director::is_cli()) {
```
tractorcow: This will prevent unit tests from running the middleware. See https://github.com/silverstripe/silverstripe-framework/pull/8158 for a fix that works around this.

PR #8159, opened 2018-06-11
Fix obsolete naming

jakedaleweb:
Currently we try to rename obsolete tables without checking their case is correct. Therefore if, after upgrading 3->4 and running a dev/build for the first time, a misnamed table is marked as obsolete you get a failure.

For example
`ALTER TABLE "_obsolete_Page_Versions" RENAME "_obsolete__obsolete_Page_Versions"`

Where `_obsolete_Page_versions` exists but not `_obsolete_Page_Versions`.

This change updates the case of the table before running the rename to obsolete query.

### DIFF
--- a/src/ORM/Connect/DBSchemaManager.php
+++ b/src/ORM/Connect/DBSchemaManager.php
@@ -454,16 +454,22 @@ MESSAGE
      */
     public function dontRequireTable($table)
     {
-        if (isset($this->tableList[strtolower($table)])) {
-            $suffix = '';
-            while (isset($this->tableList[strtolower("_obsolete_{$table}$suffix")])) {
-                $suffix = $suffix
-                        ? ((int)$suffix + 1)
-                        : 2;
-            }
-            $this->renameTable($table, "_obsolete_{$table}$suffix");
-            $this->alterationMessage("Table $table: renamed to _obsolete_{$table}$suffix", "obsolete");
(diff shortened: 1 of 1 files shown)


src/ORM/Connect/DBSchemaManager.php
```diff
@@ -454,16 +454,22 @@ public function requireTable(
...
-                        : 2;
-            }
-            $this->renameTable($table, "_obsolete_{$table}$suffix");
-            $this->alterationMessage("Table $table: renamed to _obsolete_{$table}$suffix", "obsolete");
+        if (!isset($this->tableList[strtolower($table)])) {
+            return;
+        }
+
+        if (Config::inst()->get(static::class, 'fix_table_case_on_build')) {
+            $this->fixTableCase($table);
```
tractorcow: This will result in a large number of renames.

`SiteTree_versions -> somerandomtable -> SiteTree_Versions -> _obsolete_SiteTree_Versions`

Wouldn't a simpler fix be.

```php
$renameFrom = $this->tableList[strtolower($table)];
$renameTo = "_obsolete_{$table}$suffix";
$this->renameTable($renameFrom, $renameTo);
```

unclecheese:
Have revised per @tractorcow 's review.

tractorcow:
Amazing, merge on green.

PR #8160, opened 2018-06-11
BUG Error if invalid stage specified for get_by_stage

tractorcow:
With this PR, modules that set invalid stage when calling Versioned::get_by_stage() can actually be debugged. Without this, versioned will throw random errors later on in the execution stack which aren't easy to debug.


tractorcow:
Removing 3.7 alias from 3 branch fixed the build.

PR #8161, opened 2018-06-11
Simplify travis builds

tractorcow:
3 framework phpunit
1 framework behat
1 cms phpunit
1 cmsbehat

Related links:

 - https://github.com/silverstripe/silverstripe-framework/issues/8014
 - https://github.com/silverstripeltd/open-sourcerers/issues/30 (private repo sorry

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,25 +12,17 @@ addons:
 
 matrix:
   include:
-    - php: 7.2
-      env: DB=MYSQL PDO=1
-    - php: 5.3
-      env: DB=MYSQL
     - php: 5.4
-      env: DB=PGSQL
-    - php: 5.5
-      env: DB=SQLITE
-    - php: 5.6
       env: DB=MYSQL PDO=1
     - php: 7.0
-      env: DB=MYSQL
-    - php: 7.1
-      env: DB=MYSQL PDO=1
+      env: DB=SQLITE
+    - php: 7.2
+      env: DB=PGSQL
     - php: 5.6
       env: DB=MYSQL BEHAT_TEST=1
-    - php: 7.0
+    - php: 5.5
       env: DB=MYSQL CMS_TEST=1
-    - php: 7.0
+    - php: 7.1
       env: DB=MYSQL BEHAT_TEST=1 CMS_TEST=1
 
 before_script:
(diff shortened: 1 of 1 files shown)


tractorcow:
@sminnee any opinions here?

sminnee:
Looks like a step in the right direction. I disagree that it's not worth trimming down our builds. Failing to do this slows down our ability to work on 3.

Couple of suggestions:

* As I understand, PHP 5.3 testing is about to stop working altogether. If so, let's drop it from the test suite now. Use 5.4 instead, probably?
 * Maybe run one of the CMS tests on PHP 7.1 to improve the version coverage?

sminnee:
They [claimed last year](https://blog.travis-ci.com/2017-08-31-trusty-as-default-status) that support for precise will end 1 April 2018, which has passed. Presumably they could turn it off at any time and our builds will start breaking.

Although I wouldn't formally drop support for PHP 5.3, I think it's acceptable to stop testing it.

robbieaverill:
>Although I wouldn't formally drop support for PHP 5.3, I think it's acceptable to stop testing it.

Yeah, we'd need to maintain our own docker environment to test it in probably - not worth it. It did provide benefit of simple syntax checks against PHP 5.3 e.g. not using square brackets, but as long as we remain vigilant it shouldn't be too much of a problem

dhensby:
> Although I wouldn't formally drop support for PHP 5.3, I think it's acceptable to stop testing it.

Mmmm, maybe it would have been good to drop it from 3.7, but alas it's too late for that and probably too contentious whilst our policy on this area of semver is up in the air.

sminnee:
We could add "we drop testing of any PHP release that has been EOL'd for more than 3 years" to our policy. This will lead to a date for dropping 5.4 and 5.5 over the remainder of 3.x's lifetime, but no likely droppages for 4.x

dhensby:
Sorry, I meant official support rather than test runs. Happy to drop test runs if Travis has not updated their EoL guidance on precise builds

sminnee:
Yeah I think testrun only is an ok step right now.

robbieaverill:
We still need to remove `dist: precise` as soon as we can really, but this is a good start

Issue #8136, opened 2018-06-05
Using setItemRequestClass() on a GridFieldDetailForm has no effect, due to RecursivePublishable extension


colintucker:
## Affected Version

```
silverstripe/admin          1.1.1  SilverStripe admin interface
silverstripe/asset-admin    1.1.1  Asset management for the SilverStripe CMS
silverstripe/assets         1.1.1  SilverStripe Assets component
silverstripe/campaign-admin 1.1.1  SilverStripe campaign admin interface
silverstripe/cms            4.1.1  The SilverStripe Content Management System
silverstripe/config         1.0.4  SilverStripe configuration based on YAML and class statics
silverstripe/errorpage      1.1.1  ErrorPage component for SilverStripe CMS
silverstripe/framework      4.1.1  The SilverStripe framework
silverstripe/graphql        1.1.1  GraphQL server for SilverStripe models and other data
silverstripe/reports        4.1.1  Reports module for SilverStripe CMS
silverstripe/siteconfig     4.1.1  Site wide settings administration.
silverstripe/versioned      1.1.1  SilverStripe Versioned component
```

## Description

I have a `ModelAdmin` where I need to set a custom item request class using the following code:

```php
public function getEditForm($id = null, $fields = null)
{
    // Obtain Form Object (from parent):
    $form = parent::getEditForm($id, $fields);

    // Obtain Grid Field Object:
    $gridField = $form->Fields()->fieldByName($this->sanitiseClassName($this->modelClass));

    // Obtain Detail Form Component:
    $detailForm = $gridField->getConfig()->getComponentByType(GridFieldDetailForm::class);

    // Update Item Request Class:
    $detailForm->setItemRequestClass(MyCustomised_ItemRequest::class);

    // ...
}
```
However, upon reloading the admin, nothing changes.  What gives?  I started digging through the code.  We can see here in `GridFieldDetailForm` where the item request class is obtained:

```php
protected function getItemRequestHandler($gridField, $record, $requestHandler)
{
    $class = $this->getItemRequestClass();
    $this->extend('updateItemRequestClass', $class, $gridField, $record, $requestHandler);

    // ...
}
```

In the first line of the method, `$class` is indeed set to my custom class.  However, when it applies extensions in the following line, `$class` will *always* be changed to `VersionedGridFieldItemRequest`.  This is because the `VersionedGridFieldDetailForm` extension checks if the record has the extension `RecursivePublishable`.

[...]

PR #8164, opened 2018-06-11
ENHANCEMENT Ensure extensions are told the internal item request class for gridfield detail form

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/8136

Requires https://github.com/silverstripe/silverstripe-versioned/pull/159 to fix the above issue.

### DIFF
--- a/src/Forms/GridField/GridFieldDetailForm.php
+++ b/src/Forms/GridField/GridFieldDetailForm.php
@@ -129,7 +129,8 @@ class GridFieldDetailForm implements GridField_URLHandler
     protected function getItemRequestHandler($gridField, $record, $requestHandler)
     {
         $class = $this->getItemRequestClass();
-        $this->extend('updateItemRequestClass', $class, $gridField, $record, $requestHandler);
+        $assignedClass = $this->itemRequestClass;
+        $this->extend('updateItemRequestClass', $class, $gridField, $record, $requestHandler, $assignedClass);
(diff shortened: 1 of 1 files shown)


src/Forms/GridField/GridFieldDetailForm.php
```diff
@@ -129,7 +129,8 @@ public function handleItem($gridField, $request)
     protected function getItemRequestHandler($gridField, $record, $requestHandler)
     {
         $class = $this->getItemRequestClass();
-        $this->extend('updateItemRequestClass', $class, $gridField, $record, $requestHandler);
+        $assignedClass = $this->itemRequestClass;
+        $this->extend('updateItemRequestClass', $class, $gridField, $record, $requestHandler, $assignedClass);
```
maxime-rainville: I'm not super familiar with this class, but to me it feels a bit weird that we have a `updateItemRequestClass` hook in a method call `getItemRequestHandler()`. 

If we need to allow extension to alter the ItemRequestClass wouldn't that be better handled in `getItemRequestClass()`.

On a different topic, it's not clear at all here that `$class` is being passed by reference, which makes the code a bit confusing.

robbieaverill:
Thanks for fixing this! We've had to disable the default functionality completely in userforms because it wasn't extensible:

https://github.com/silverstripe/silverstripe-userforms/blob/5.1.1/code/Model/Recipient/EmailRecipient.php#L93-L98

PR #8166, opened 2018-06-12
NEW Add better HTTP cache-control manipulation (4.x branch)

tractorcow:
Parent story https://github.com/silverstripe/silverstripe-versioned/issues/146

Rewrite of https://github.com/silverstripe/silverstripe-framework/pull/8086 for 4.x branch.

In the original rewrite I wasn't happy with the "choose state" and "set directives" being so intercoupled so I've separated the concerns. You can now set a directive for a single (or set of) states, before you know what the actual state SHOULD be.

For example, I can set different caching ages for private and public caches before choosing between them.

HTTPCacheControl has been rewritten into a HTTPCacheControlMiddleware.

### DIFF
--- a/_config/requestprocessors.yml
+++ b/_config/requestprocessors.yml
@@ -11,6 +11,7 @@ SilverStripe\Core\Injector\Injector:
         SessionMiddleware: '%$SilverStripe\Control\Middleware\SessionMiddleware'
         RequestProcessorMiddleware: '%$SilverStripe\Control\RequestProcessor'
         FlushMiddleware: '%$SilverStripe\Control\Middleware\FlushMiddleware'
+        HTTPCacheControleMiddleware: '%$SilverStripe\Control\Middleware\HTTPCacheControlMiddleware'
         CanonicalURLMiddleware: '%$SilverStripe\Control\Middleware\CanonicalURLMiddleware'
   SilverStripe\Control\Middleware\AllowedHostsMiddleware:
     properties:
@@ -46,4 +47,12 @@ SilverStripe\Core\Injector\Injector:
     properties:
       ForceSSL: false
       ForceWWW: false
-
+---
+Name: httpcache-dev
+Only:
+  environment: dev
+---
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\Control\Middleware\HTTPCacheControlMiddleware:
+    Properties:
+      Enabled: false
--- a/_config/requestprocessors.yml
+++ b/_config/requestprocessors.yml
@@ -11,6 +11,7 @@ SilverStripe\Core\Injector\Injector:
         SessionMiddleware: '%$SilverStripe\Control\Middleware\SessionMiddleware'
         RequestProcessorMiddleware: '%$SilverStripe\Control\RequestProcessor'
         FlushMiddleware: '%$SilverStripe\Control\Middleware\FlushMiddleware'
+        ETagMiddleware: '%$SilverStripe\Control\Middleware\ETagMiddleware'
         HTTPCacheControleMiddleware: '%$SilverStripe\Control\Middleware\HTTPCacheControlMiddleware'
         CanonicalURLMiddleware: '%$SilverStripe\Control\Middleware\CanonicalURLMiddleware'
   SilverStripe\Control\Middleware\AllowedHostsMiddleware:
--- a/_config/requestprocessors.yml
+++ b/_config/requestprocessors.yml
@@ -47,12 +47,3 @@ SilverStripe\Core\Injector\Injector:
     properties:
       ForceSSL: false
       ForceWWW: false
----
-Name: httpcache-dev
-Only:
-  environment: dev
----
-SilverStripe\Core\Injector\Injector:
-  SilverStripe\Control\Middleware\HTTPCacheControlMiddleware:
-    Properties:
-      Enabled: false
--- a/_config/requestprocessors.yml
+++ b/_config/requestprocessors.yml
@@ -11,7 +11,7 @@ SilverStripe\Core\Injector\Injector:
         SessionMiddleware: '%$SilverStripe\Control\Middleware\SessionMiddleware'
         RequestProcessorMiddleware: '%$SilverStripe\Control\RequestProcessor'
         FlushMiddleware: '%$SilverStripe\Control\Middleware\FlushMiddleware'
-        ETagMiddleware: '%$SilverStripe\Control\Middleware\ETagMiddleware'
+        ChangeDetectionMiddleware: '%$SilverStripe\Control\Middleware\ChangeDetectionMiddleware'
         HTTPCacheControleMiddleware: '%$SilverStripe\Control\Middleware\HTTPCacheControlMiddleware'
         CanonicalURLMiddleware: '%$SilverStripe\Control\Middleware\CanonicalURLMiddleware'
   SilverStripe\Control\Middleware\AllowedHostsMiddleware:
--- a/docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
+++ b/docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
@@ -21,8 +21,8 @@ are a great way to learn about HTTP caching.
 
 In order to support developers in making safe choices around HTTP caching,
 we're using a `HTTPCacheControlMiddleware` class to control if a response
-should be considered public or private. This is an abstraction on existing
-lowlevel APIs like `HTTP::add_cache_headers()` and `HTTPResponse->addHeader()`.
+should be considered public or private. This is an abstraction on the
+`HTTPResponse->addHeader()` lowlevel API.
 
 The `HTTPCacheControlMiddleware` API makes it easier to express your caching preferences
 without running the risk of overriding essential core safety measures.
--- a/docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
+++ b/docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
@@ -1,7 +1,188 @@
 title: HTTP Cache Headers
 summary: Set the correct HTTP cache headers for your responses.
 
-# Caching Headers
+# HTTP Cache Headers
+
+## Overview
+
+By default, SilverStripe sends headers which signal to HTTP caches
+that the response should be not considered cacheable.
+HTTP caches can either be intermediary caches (e.g. CDNs and proxies), or clients (e.g. browsers).
+The cache headers sent are `Cache-Control: no-store, no-cache, must-revalidate`;
+
+HTTP caching can be a great way to speed up your website, but needs to be properly applied.
+Getting it wrong can accidentally expose draft pages or other protected content.
+The [Google Web Fundamentals](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#public_vs_private)
+are a great way to learn about HTTP caching.
+
+## Cache Control Headers
+
+### Overview
+
+In order to support developers in making safe choices around HTTP caching,
+we're using a `HTTPCacheControl` class to control if a response
+should be considered public or private. This is an abstraction on existing
+lowlevel APIs like `HTTP::add_cache_headers()` and `SS_HTTPResponse->addHeader()`.
+
+The `HTTPCacheControl` API makes it easier to express your caching preferences
+without running the risk of overriding essential core safety measures.
+Most commonly, these APIs will prevent HTTP caching of draft content.
+
+It will also prevent caching of content generated with an active session,
+since the system can't tell whether session data was used to vary the output.
+In this case, it's up to the developer to opt-in to caching,
+after ensuring that certain execution paths are safe despite of using sessions.
+
+The system behaviour does not guard against accidentally caching "private" content,
+since there are too many variations under which output could be considered private
+(e.g. a custom "approval" flag on a comment object). It is up to
+the developer to ensure caching is used appropriately there.
+
+The [api:SilverStripe\Control\Middleware\HTTPCacheControlMiddleware] class supplements the `HTTP` helper class.
+It comes with methods which let developers safely interact with the `Cache-Control` header.
+
+### disableCache()
+
+Simple way to set cache control header to a non-cacheable state.
+Use this method over `privateCache()` if you are unsure about caching details.
+Takes precendence over unforced `enableCache()`, `privateCache()` or `publicCache()` calls.
+
+Removes all state and replaces it with `no-cache, no-store, must-revalidate`. Although `no-store` is sufficient
+the others are added under [recommendation from Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Examples)
+
+Does not set `private` directive, use `privateCache()` if this is explicitly required
+([details](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#public_vs_private))
+
+### enableCache()
+
+Simple way to set cache control header to a cacheable state.
+Use this method over `publicCache()` if you are unsure about caching details.
+
+Removes `no-store` and `no-cache` directives; other directives will remain in place.
+Use alongside `setMaxAge()` to indicate caching.
+
+Does not set `public` directive. Usually, `setMaxAge()` is sufficient. Use `publicCache()` if this is explicitly required
+([details](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#public_vs_private))
+
+### privateCache()
+
+Advanced way to set cache control header to a non-cacheable state.
+Indicates that the response is intended for a single user and must not be stored by a shared cache.
+A private cache (e.g. Web Browser) may store the response. Also removes `public` as this is a contradictory directive.
+
+### publicCache()
+
+Advanced way to set cache control header to a cacheable state.
+Indicates that the response may be cached by any cache. (eg: CDNs, Proxies, Web browsers)
+Also removes `private` as this is a contradictory directive
+
+### Priority
+    
+Each of these highlevel methods has a boolean `$force` parameter which determines
+their application priority regardless of execution order.
+The priority order is as followed, sorted in descending order
+(earlier items will overrule later items): 
+
+ * `disableCache($force=true)`
+ * `privateCache($force=true)`
+ * `publicCache($force=true)`
+ * `enableCache($force=true)`
+ * `disableCache()`
+ * `privateCache()`
+ * `publicCache()`
+ * `enableCache()`
+
+## Cache Control Examples
+
+### Global opt-in for page content 
+
+Enable caching for all page content (through `Page_Controller`).
+
+```php
+<?php
+
+use SilverStripe\Control\Middleware\HTTPCacheControlMiddleware;
+use SilverStripe\CMS\Controllers\ContentController;
+
+class PageController extends ContentController
+{
+    public function init()
+    {
+        HTTPCacheControlMiddleware::singleton()
+           ->enableCache()
+           ->setMaxAge(60); // 1 minute
+        
+        parent::init();
+    }
+}
+```
+
+Note: SilverStripe will still override this preference when a session is active,
+a [CSRF token](/developer_guides/forms/form_security) token is present,
+or draft content has been requested.
+
+### Opt-out for a particular controller action
+
+If a controller output relies on session data, cookies,
+permission checks or other triggers for conditional output,
+you can disable caching either on a controller level
+(through `init()`) or for a particular action.
+
+```php
+<?php
+
+use SilverStripe\Control\Middleware\HTTPCacheControlMiddleware;
+use SilverStripe\CMS\Controllers\ContentController;
+
+class PageController extends ContentController
+{
+    public function myprivateaction($request)
+    {
+        HTTPCacheControlMiddleware::singleton()
+           ->disableCache();
+        
+        return $this->myPrivateResponse();;
+    }
+}
+```
+
+Note: SilverStripe will still override this preference when a session is active,
+a [CSRF token](/developer_guides/forms/form_security) token is present,
+or draft content has been requested. 
+
+### Global opt-in, ignoring session (advanced) 
+
+This can be helpful in situations where forms are embedded on the website.
+SilverStripe will still override this preference when draft content has been requested.
+CAUTION: This mode relies on a developer examining each execution path to ensure
+that no session data is used to vary output. 
+
+Use case: By default, forms include a [CSRF token](/developer_guides/forms/form_security)
+which starts a session with a value that's unique to the visitor, which makes the output uncacheable.
+But any subsequent requests by this visitor will also carry a session, leading to uncacheable output
+for this visitor. This is the case even if the output does not contain any forms,
+and does not vary for this particular visitor.
+
+```php
+<?php
+
+use SilverStripe\Control\Middleware\HTTPCacheControlMiddleware;
+use SilverStripe\CMS\Controllers\ContentController;
+
+class PageController extends ContentController
+{
+    public function init()
+    {
+        HTTPCacheControlMiddleware::inst()
+           ->enableCache($force=true) // DANGER ZONE
+           ->setMaxAge(60); // 1 minute
+        
+        parent::init();
+    }
+}
+```
+
+## Defaults
 
 By default, PHP adds caching headers that make the page appear purely dynamic. This isn't usually appropriate for most 
 sites, even ones that are updated reasonably frequently. SilverStripe overrides the default settings with the following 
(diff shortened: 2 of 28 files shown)


tractorcow:
todo: migrate tests

_config/requestprocessors.yml
```diff
@@ -46,4 +47,12 @@ SilverStripe\Core\Injector\Injector:
...
       ForceSSL: false
       ForceWWW: false
-
+---
+Name: httpcache-dev
+Only:
+  environment: dev
+---
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\Control\Middleware\HTTPCacheControlMiddleware:
```
tractorcow: I need to delete this; Keeping the enabled flag from HTTP class, not the new class.

src/Control/Middleware/HTTPCacheControlMiddleware.php
```diff
@@ -0,0 +1,557 @@
...
+     */
+    public function process(HTTPRequest $request, callable $delegate)
+    {
+        try {
+            $response = $delegate($request);
+        } catch (HTTPResponse_Exception $ex) {
+            $response = $ex->getResponse();
+        }
+
+		// If sessions exist we assume that the responses should not be cached by CDNs / proxies as we are
```
tractorcow: Typical from copy/pasting 3.x code... it seems to have copied the indentation too sorry.

dhensby:
Right, this looks pretty cool - thanks @tractorcow 

I've pushed up linting fixes, the tests and a bug I found with the ETag generation.

I've made the middleware throw the `HTTPResponse_Exception` that it catches because I don't think we should be swallowing them (??)

dhensby:
The tests aren't passing and I think that is because of the fact that the session is bootstrapped for every test and so there is never a time there's not a session running during a test...

dhensby:
It seems that `publicCache` calls aren't taking precedence over `enableCache` for some reason

src/Control/Middleware/HTTPCacheControlMiddleware.php
```diff
@@ -0,0 +1,561 @@
...
+use SilverStripe\Control\HTTP;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
+use SilverStripe\Control\HTTPResponse_Exception;
+use SilverStripe\Core\Config\Configurable;
+use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Core\Resettable;
+
+class HTTPCacheControlMiddleware implements HTTPMiddleware, Resettable
```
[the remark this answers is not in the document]
dhensby: I guess there are a couple points here

1. It's namespaced, so even if the classname is the same it will actually be different (though I accept upgrading much easier as you just need the class import once and that's it).
2. Our convention appears to dictate that we add the "thing" that the class is to the class name (and not only the namespace) so you know what it is just by looking at it.
3. If people really want to upgrade their code by *only* adding a class import they can just do `use ... as HTTPCacheControl`

tractorcow:
Quick rebase.

src/Control/Middleware/HTTPCacheControlMiddleware.php
```diff
@@ -50,6 +51,9 @@ public function process(HTTPRequest $request, callable $delegate)
         }
 
         HTTP::add_cache_headers($response);
+        if (isset($ex)) {
```
tractorcow: I wasn't sure if I should do that. Isn't returning the response safer?

src/Control/HTTP.php
```diff
@@ -571,7 +571,7 @@ protected static function generateETag($response)
             ? $response->getBody()
             : $response;
         if ($body) {
-            return sprintf('"%s"', md5($response));
+            return sprintf('"%s"', md5($body));
```
tractorcow: Oops my bad.

src/Forms/Form.php
```diff
@@ -884,7 +885,7 @@ public function getAttributesHTML($attrs = null)
 
         // If we need to disable cache, do it
         if ($needsCacheDisabled) {
-            HTTP::set_cache_age(0);
+            HTTPCacheControlMiddleware::singleton()->disableCache();
```
tractorcow: Good fix but really odd that we do this in getAttributesHTML lol.

tractorcow:
Oh @dhensby I had  a `if (!$this->applyChangeLevel(self::LEVEL_PUBLIC, $force)) {` not `if ($this->applyChangeLevel(self::LEVEL_PUBLIC, $force)) {`

tractorcow:
Yay tests are all green for me locally.

tests/php/Control/HTTPTest.php
```diff
@@ -18,31 +19,43 @@
...
 {
+    protected function setUp()
+    {
+        parent::setUp();
+        // Remove dev-only config
+        Config::modify()->remove(HTTP::class, 'disable_http_cache');
+        HTTPCacheControlMiddleware::reset();
+    }
 
     public function testAddCacheHeaders()
```
sminnee: Do we test deprecated methods?

src/Control/HTTP.php
```diff
@@ -376,156 +413,132 @@ public static function register_etag($etag)
      */
     public static function add_cache_headers($body = null)
```
sminnee: Add a deprecation warning and docblock note.

src/Control/HTTP.php
```diff
@@ -376,156 +413,132 @@ public static function register_etag($etag)
...
-        if (function_exists('apache_request_headers') && static::config()->uninherited('cache_ajax_requests')) {
-            $requestHeaders = array_change_key_case(apache_request_headers(), CASE_LOWER);
-            if (isset($requestHeaders['x-requested-with'])
-                && $requestHeaders['x-requested-with']=='XMLHttpRequest'
-            ) {
-                $cacheAge = 0;
-            }
+        // if http caching is disabled by config, disable it - used on dev environments due to frequently changing
+        // templates and other data. will be overridden by forced publicCache() or privateCache() calls
+        if ($config->get('disable_http_cache')) {
```
sminnee: Shift this to HTTPCacheControlMiddleware

src/Control/HTTP.php
```diff
@@ -376,156 +413,132 @@ public static function register_etag($etag)
...
+        // Populate $responseHeaders with all the headers that we want to build
+        $responseHeaders = [];
 
-            // Set empty pragma to avoid PHP's session_cache_limiter adding conflicting caching information,
-            // defaulting to "nocache" on most PHP configurations (see http://php.net/session_cache_limiter).
-            // Since it's a deprecated HTTP 1.0 option, all modern HTTP clients and proxies should
-            // prefer the caching information indicated through the "Cache-Control" header.
-            $responseHeaders["Pragma"] = "";
+        // if no caching ajax requests, disable ajax if is ajax request
+        if (!$config->get('cache_ajax_requests') && Director::is_ajax()) {
```
sminnee: Shift this to HTTPCacheControlMiddleware

src/Control/HTTP.php
```diff
@@ -376,156 +413,132 @@ public static function register_etag($etag)
...
-        } else {
-            $contentDisposition = null;
-            if ($body) {
-                // Grab header for checking. Unfortunately HTTPRequest uses a mistyped variant.
-                $contentDisposition = $body->getHeader('Content-disposition');
-                if (!$contentDisposition) {
-                    $contentDisposition = $body->getHeader('Content-Disposition');
-                }
-            }
+        // Errors disable cache (unless some errors are cached intentionally by usercode)
```
sminnee: Shift this to HTTPCacheControlMiddleware

src/Control/HTTP.php
```diff
@@ -376,156 +413,132 @@ public static function register_etag($etag)
...
-                // Set empty pragma to avoid PHP's session_cache_limiter adding conflicting caching information,
-                // defaulting to "nocache" on most PHP configurations (see http://php.net/session_cache_limiter).
-                // Since it's a deprecated HTTP 1.0 option, all modern HTTP clients and proxies should
-                // prefer the caching information indicated through the "Cache-Control" header.
-                $responseHeaders["Pragma"] = "";
-            } else {
-                $cacheControlHeaders['no-cache'] = "true";
-                $cacheControlHeaders['no-store'] = "true";
-            }
+        // split the current vary header into it's parts and merge it with the config settings
```
sminnee: Shift this to HTTPCacheControlMiddleware

src/Control/HTTP.php
```diff
@@ -376,156 +413,132 @@ public static function register_etag($etag)
...
+        // deal with IE6-IE8 problems with https and no-cache
+        $contentDisposition = null;
+        if ($body) {
+            // Grab header for checking. Unfortunately HTTPRequest uses a mistyped variant.
+            $contentDisposition = $body->getHeader('Content-Disposition');
         }
 
-        $responseHeaders['Cache-Control'] = implode(', ', $cacheControlHeaders);
-        unset($cacheControlHeaders, $header, $value);
+        if ($body &&
```
sminnee: Shift this to HTTPCacheControlMiddleware

tractorcow:
I've pushed up a huge refactor so that we no longer use HTTP class for cache control anymore; The middleware is now the source of truth.

I've also created an ETagMiddleware; But since it also handles If-Modified-Since, I'm thinking that perhaps it should be renamed to another class.

dhensby:
> I've also created an ETagMiddleware; But since it also handles If-Modified-Since, I'm thinking that perhaps it should be renamed to another class.

NotModifiedMiddleware?

src/Control/Middleware/ETagMiddleware.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * @param HTTPResponse|string $response
+     * @return string|false
+     */
+    protected function generateETag(HTTPResponse $response)
+    {
+        // Existing e-tag
+        if ($response instanceof HTTPResponse && $response->getHeader('ETag')) {
```
dhensby: Lots of this logic can be cleaned up, $response is always going to be passed and always will be a `HTTPResponse` object.

src/Forms/Form.php
dhensby: Yeah, which makes me worried - maybe it should really be in `forTemplate` ?

src/Control/Middleware/HTTPCacheControlMiddleware.php
```diff
@@ -0,0 +1,776 @@
...
+
+        // IE6-IE8 have problems saving files when https and no-cache/no-store are used
+        // (http://support.microsoft.com/kb/323308)
+        // Note: this is also fixable by ticking "Do not save encrypted pages to disk" in advanced options.
+        if ($request->getScheme() === 'https' &&
+            preg_match('/.+MSIE (7|8).+/', $request->getHeader('User-Agent')) &&
+            strstr($response->getHeader('Content-Disposition'), 'attachment;') == true &&
+            ($this->hasDirective('no-cache') || $this->hasDirective('no-store'))
+        ) {
+            $this->privateCache(true);
```
dhensby: We no longer support IE< 11, so should we keep this legacy code around? I'd vote to remove it.

Keeping it means we technically vary responses based on UA, which is not in our Vary header any longer. We'd probably want to add a max-age here too(?)

kinglozzer: We no longer support it in the CMS, but if this affects or can be used in the frontend then we should keep it IMO

dhensby: OK, for how long? When can we safely remove IE hacks from our codebase?

kinglozzer: Hmm good point. I was skimming and also missed the narrow range this actually applies to (only IE 7 & 8, while running over HTTPS, plus certain headers/directives). Devs can trigger this behaviour themselves too, so I’m convinced it’s enough of an edge case to discard in core code given how old those browser are.

dhensby: Yeah, someone can add their own middleware to deal with this edgecase ;)

dhensby:
OK - I've gone and addressed most of my feedback and left a few questions.

I've also removed the old <=IE8 attachment handling code as I don't think it's worth supporting anymore. People can add it with their own middlewares ;) - happy for others to veto this change, though.

I've also added a `__toString` method on `HTTPResponse` as I'm considering if `ETag`s should be generated from the full response (including headers) or not?

I'm also wondering if we should add the etag to the actual response object... what if we want our response to be a file attachment? It may be better for the asset module to add the ETag itself (the variant ID could be enough here) rather than us computing the md5 of the entire binary file (which could be a stream)... thoughts?

dhensby:
Oh, I'm also concerned about putting all the different headers into the single middleware; I feel like it takes control away from devs who may want to modify this behaviour for individual headers; plus the lastmodified stuff isn't really needed anymore if we have proper ETags..

I'd vote either to remove the LastModified stuff completely or move it to it's own middleware and have single purpose middlewares responsible for a header each.

tractorcow:
Yeah I wasn't sure about the other headers. They are important to caching, but aren't strictly caching only.

docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
```diff
@@ -1,7 +1,188 @@
...
+the developer to ensure caching is used appropriately there.
+
+The [api:SilverStripe\Control\Middleware\HTTPCacheControlMiddleware] class supplements the `HTTP` helper class.
+It comes with methods which let developers safely interact with the `Cache-Control` header.
+
+### disableCache()
+
+Simple way to set cache control header to a non-cacheable state.
+Use this method over `privateCache()` if you are unsure about caching details.
+Takes precendence over unforced `enableCache()`, `privateCache()` or `publicCache()` calls.
```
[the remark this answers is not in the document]
tractorcow: I just copied your docs. ;P

sminnee:
> I've also created an ETagMiddleware; But since it also handles If-Modified-Since, I'm thinking that perhaps it should be renamed to another class.

Frankly I think we can ditch both the last-modified header and interrogation of the IfModifiedSince. It's an inferior alternative to Etag. Since the site outcome is the same — we are providing an accurate caching header of the same purpose — I don't see this as an API breakage. However, we'd need register_modification_date() and register_modification_timestamps() to throw deprecation warnings rather than just deleting the functions.

Currently the last-modified date is auto-generated by any DataObject::__construct(), which means that the page only registers as changed if there was a DataObject instantiated that was changed. This is dangerously erroneous in cases such as partial caching and will likely lead to the "over-caching", where a 304 response is sent event though the page has changed. as part of deprecating register_modification_date() and register_modification_timestamps() we could remove this behaviour too. I think that changing broken and dangerous behaviour is allowed in a minor release ;-)

If we really wanted to provide it I'd have a LastModifiedMiddleware that handled both the application of the last_modified date, and the interrogation of if-modified-since. This, to me, is a good grouping of closely related concerns in a single class. I don't think this is necessary, but a more conservative approach might be decided upon by the group.

Then we have

ETagMiddleware:
 * Generates an Etag header
 * Returns 304 on an appropriate If-None-Match required

LastModifiedMiddleware:
 * Generates a LastModified header
 * Returns a 304 on an appropriate If-Modified-Since request

docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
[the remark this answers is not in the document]
dhensby: Whoops

docs/en/04_Changelogs/4.2.0.md
```diff
@@ -201,3 +201,147 @@ SilverStripe\Core\Injector\Injector:
...
++           ->setMaxAge(60); // 1 minute
+
+        parent::init();
+    }
+}
+```
+
+#### Detailed Cache-Control Changes:
+
+ * Added `Cache-Control: no-store` header to default responses,
```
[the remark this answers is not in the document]
dhensby: I think this is a big one that needs looking at, I think the default is and should remain no-cache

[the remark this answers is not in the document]
sminnee: OK, no-store seems excessive. It would be of most use in cases where the content was too sensitive to leave on a user's system, but that's overkill for our default.

src/Control/HTTP.php
```diff
@@ -323,41 +364,45 @@ public static function get_mime_type($filename)
...
+     * @deprecated 4.2..5.0 Use HTTPCacheControlMiddleware::registerModificationDate() instead
      */
     public static function register_modification_date($dateString)
     {
-        $timestamp = strtotime($dateString);
-        if ($timestamp > self::$modification_date) {
-            self::$modification_date = $timestamp;
-        }
+        Deprecation::notice('5.0', 'Use HTTPCacheControlMiddleware::registerModificationDate() instead');
+        HTTPCacheControlMiddleware::singleton()->registerModificationDate($dateString);
```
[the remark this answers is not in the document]
sminnee: Let's delete this functionality altogether, and leave the deprecated methods as a no-op.

src/Forms/Form.php
sminnee: That would probably be tidier.

sminnee:
Historically, setCacheAge(0) has been the way you disable caching. I wonder if we should have something where if you call setCacheAge(0) it throws a deprecation notice but also calls disableCache()?

docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
```diff
@@ -1,7 +1,188 @@
...
+since the system can't tell whether session data was used to vary the output.
+In this case, it's up to the developer to opt-in to caching,
+after ensuring that certain execution paths are safe despite of using sessions.
+
+The system behaviour does not guard against accidentally caching "private" content,
+since there are too many variations under which output could be considered private
+(e.g. a custom "approval" flag on a comment object). It is up to
+the developer to ensure caching is used appropriately there.
+
+The [api:SilverStripe\Control\Middleware\HTTPCacheControlMiddleware] class supplements the `HTTP` helper class.
```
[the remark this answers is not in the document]
chillu: Fixed

docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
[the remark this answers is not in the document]
chillu: This relates to the topic of the section:

> (disableCache) takes precendence over unforced `enableCache()`, `privateCache()` or `publicCache()` calls

docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
```diff
@@ -1,7 +1,188 @@
...
+           ->enableCache()
+           ->setMaxAge(60); // 1 minute
+        
+        parent::init();
+    }
+}
+```
+
+Note: SilverStripe will still override this preference when a session is active,
+a [CSRF token](/developer_guides/forms/form_security) token is present,
```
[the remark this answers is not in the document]
chillu: I've mentioned the first two cases (validator and tokens) consistently in both docs now. Didn't mention the GET vs. POST case because I'm not sure that something worth retaining. Forms can contain sensitive, user specific data regardless of GET or POST.

src/Control/HTTPResponse.php
```diff
@@ -109,12 +114,31 @@ class HTTPResponse
...
      */
-    public function __construct($body = null, $statusCode = null, $statusDescription = null)
+    public function __construct($body = null, $statusCode = null, $statusDescription = null, $version = null)
     {
         $this->setBody($body);
         if ($statusCode) {
             $this->setStatusCode($statusCode, $statusDescription);
         }
+        if (!$version) {
+            if (preg_match('/HTTP\/(\d+(\.\d+)?)/i', $_SERVER['SERVER_PROTOCOL'], $matches)) {
```
tractorcow: I'm not sure we really need this commit. It's not used by etag is it?

src/Control/HTTP.php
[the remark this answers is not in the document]
tractorcow: ChangeDetectionMiddleware

docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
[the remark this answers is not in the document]
tractorcow: Section heading is disableCache() not privateCache().

src/Control/HTTP.php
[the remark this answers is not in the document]
chillu: +1 on what Sam said. The whole idea around using `DataObject->LastEdited` for `register_modification_date()` is flawed, because it ignores any cached data which doesn't instanciate those objects. I'm aware of this leading to issues in a recent project. And rather than trying to make it more smart (tag caches with this metadata etc), we should remove it altogether. ETag is a better way to address this.

src/Control/HTTP.php
chillu: Agreed, its deprecated but if it only applies on a code path we're no longer using in core (by avoiding `add_cache_headers()`) then it needs to be moved.

src/Control/HTTPResponse.php
```diff
@@ -109,12 +114,31 @@ class HTTPResponse
...
      *  See {@link setStatusCode()} for more information.
      */
-    public function __construct($body = null, $statusCode = null, $statusDescription = null)
+    public function __construct($body = null, $statusCode = null, $statusDescription = null, $version = null)
     {
         $this->setBody($body);
         if ($statusCode) {
             $this->setStatusCode($statusCode, $statusDescription);
         }
+        if (!$version) {
```
chillu: Was this handled by `HTTP` before, or just an unrelated improvement? Can't see mentions of `SERVER_PROTOCOL` elsewhere in the diff

src/Control/Middleware/HTTPCacheControlMiddleware.php
```diff
@@ -0,0 +1,765 @@
...
+    }
+
+
+    /**
+     * Register a modification date. Used to calculate the Modification-Date http header
+     *
+     * @param string|int $date Date string or timestamp
+     * @return HTTPCacheControlMiddleware
+     */
+    public function registerModificationDate($date)
```
chillu: What value does this provide over `addHeader('Last-Modified")`?

src/Control/HTTPResponse.php
tractorcow: I don't think we need it, but it's PSR-7 so I'm inclined to let @dhensby's enhancement sneak in.

src/Control/Middleware/HTTPCacheControlMiddleware.php
tractorcow: It lets you call it multiple times and does the calculation to get only the latest modification time.

tractorcow:
@chillu 

> +1 on what Sam said. The whole idea around using DataObject->LastEdited for register_modification_date() is flawed, because it ignores any cached data which doesn't instanciate those objects. I'm aware of this leading to issues in a recent project. And rather than trying to make it more smart (tag caches with this metadata etc), we should remove it altogether. ETag is a better way to address this.

Ok to leave the API and as a usercode-accessible method? I think it still has value over a dumb header set.

tractorcow:
> Agreed, its deprecated but if it only applies on a code path we're no longer using in core (by avoiding add_cache_headers()) then it needs to be moved.


I can move to the middleware I guess.

tractorcow:
I'm shifting all the deprecated settings back into the critical path, but keeping them in a single method that's invoked by the middleware.

src/Control/Middleware/HTTPCacheControlMiddleware.php
chillu: Right, so it saves you a conditional in custom code. I guess its obvious enough to developers. I've clarified the PHPDoc, it talked about the "Modification-Date http header"

src/ORM/DataObject.php
```diff
@@ -379,12 +380,6 @@ public function __construct($record = null, $isSingleton = false, $queryParams =
 
         $this->original = $this->record;
 
-        // Keep track of the modification date of all the data sourced to make this page
-        // From this we create a Last-Modified HTTP header
-        if (isset($record['LastEdited'])) {
-            HTTP::register_modification_date($record['LastEdited']);
```
chillu: I'm going to mention the removal of this feature in the changelog. Don't think we need to retain it, but it's noteworthy

src/Forms/Form.php
```diff
@@ -1823,4 +1808,34 @@ protected function buildRequestHandler()
...
+    /**
+     * Can the body of this form be cached?
+     *
+     * Figure out if we can cache this form
+     * - forms with validation shouldn't be cached, because their error messages won't be shown
+     * - forms with security tokens shouldn't be cached because security tokens expire
+     *
+     * @return bool
+     */
+    public function canBeCached()
```
chillu: Can we make this `protected`? The `Form` API is already huge enough.

src/Forms/Form.php
```diff
@@ -1823,4 +1808,34 @@ protected function buildRequestHandler()
...
+    {
+        if ($this->getSecurityToken()->isEnabled()) {
+            return false;
+        }
+        if ($this->FormMethod() !== 'GET') {
```
chillu: This conditional is a bit weird (forms can display sensitive or session-specific data regardless of HTTP verb), but it's retained for backwards compat. It's probably a good idea not to mention it in the PHPDoc to avoid devs relying on this behaviour (it's not mentioned at the moment, and I don't think we need to fix that)

src/Control/RequestHandler.php
```diff
@@ -657,7 +658,7 @@ public function getReferer()
     public function redirectBack()
     {
         // Don't cache the redirect back ever
-        HTTP::set_cache_age(0);
+        HTTPCacheControlMiddleware::singleton()->disableCache(true);
```
chillu: I was previously [advocating](https://github.com/silverstripe/silverstripe-framework/pull/8086#issuecomment-395241798) for moving this logic into the middleware, since this decision can be made by inspecting the response. It's how the Laravel and Symfony HTTP caching middlewares work. The advantage here is that you can tailor behaviour through the middleware with high granularity, e.g. only applying redirect caching for a certain route. It's currently not possible to influence this here in `RequestHandler` unless we add some config flags which make the entire system more complex.

src/ORM/DataObject.php
tractorcow: Adding:

```
* DataObject no longer automatically invokes `HTTP::register_modification_date`
   with LastModified on construction. User code should instead invoke
   `HTTPCacheControlMiddleware::singleton()->registerModificationDate()` explicitly.
```

src/Forms/Form.php
tractorcow: Ok removing the phpdoc for now.

dhensby: Well, the URL changes if the form is submitted by GET and so HTTP Caching is mess relevant, right?

If someone goes to `http://example.com/home/Form?PersonalInfo=Dave`, they'll get that personal info regardless of whether the CDN has a cache or not...

The problem before was that errors / personal info was shown in the "vanilla" url `http://example.com/home/Form` because the submission was POSTed... right?

dhensby: I guess it's flawed logic if the data gets stored in the session and the user redirected back, though...

PR #8170, opened 2018-06-13
BUG Fix internal date formatting inheriting default locale

### LINKED ISSUE #8097: Publish error on Arabic locale caused by invalid date format
## Affected Version

4.1.0

## Description

When I set the locale to Arabic, I got the following internal error on publishing pages and files
`[Emergency] Uncaught InvalidArgumentException: Invalid date: '٢٠١٨-٠٥-٢٦ ٢٣،١٢،٢١'. Use y-MM-dd to prevent this error.
`

## Steps to Reproduce

- CMS -> My Profile
- Change Interface Language to: Arabic (Egypt)
- Try to publish a page

## PRs:

 - [ ] https://github.com/silverstripe/silverstripe-framework/pull/8170

tractorcow:
Fixes #8097

Marked as WIP because there are other issues yet to test, and I haven't updated unit tests to cover the new issues.

The issue came about because we were using setPattern() for localisation of internal date formats, but we WEREN'T setting locale. This ment that "now" time `2018-06-13 12:37:40` would instead be localised as `٢٠١٨-٠٦-١٣ ١٢:٣٧:٤٠`, for instance, due to arabic numeric localisation rules.

### DIFF
--- a/src/ORM/FieldType/DBDate.php
+++ b/src/ORM/FieldType/DBDate.php
@@ -37,6 +37,8 @@ class DBDate extends DBField
     /**
      * Fixed locale to use for ISO date formatting. This is necessary to prevent
      * locale-specific numeric localisation breaking internal date strings.
+     *
+     * @internal (remove internal in 4.2)
      */
     const ISO_LOCALE = 'en_US';
 
@@ -214,6 +216,8 @@ class DBDate extends DBField
     /**
      * Return formatter in a given locale. Useful if localising in a format other than the current locale.
      *
+     * @internal (Remove internal in 4.2)
+     *
      * @param string|null $locale The current locale, or null to use default
      * @param string|null $pattern Custom pattern to use for this, if required
      * @param int $dateLength
@@ -262,11 +266,14 @@ class DBDate extends DBField
      * for the day of the month ("1st", "2nd", "3rd" etc)
      *
      * @param string $format Format code string. See http://userguide.icu-project.org/formatparse/datetime
-     * @param string $locale Custom locale to use
+     * @param string $locale Custom locale to use (add to signature in 5.0)
      * @return string The date in the requested format
      */
-    public function Format($format, $locale = null)
+    public function Format($format)
     {
+        // Note: soft-arg uses func_get_args() to respect semver. Add to signature in 5.0
+        $locale = func_num_args() > 1 ? func_get_arg(1) : null;
+
         if (!$this->value) {
             return null;
         }
--- a/src/ORM/FieldType/DBDate.php
+++ b/src/ORM/FieldType/DBDate.php
@@ -34,6 +34,12 @@ class DBDate extends DBField
      */
     const ISO_DATE = 'y-MM-dd';
 
+    /**
+     * Fixed locale to use for ISO date formatting. This is necessary to prevent
+     * locale-specific numeric localisation breaking internal date strings.
+     */
+    const ISO_LOCALE = 'en_NZ';
+
     public function setValue($value, $record = null, $markChanged = true)
     {
         $value = $this->parseDate($value);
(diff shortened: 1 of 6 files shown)


tractorcow:
Probably should merge this into 4.0, but we did need to add new API to fix this. It's a pain to address without breaking semver.

@hudhaifas  I'd appreciate it if you could tell me if this fix addresses your issue. Testing locally it fixes it for me.

src/ORM/FieldType/DBDate.php
```diff
@@ -34,6 +34,12 @@ class DBDate extends DBField
      */
     const ISO_DATE = 'y-MM-dd';
 
+    /**
+     * Fixed locale to use for ISO date formatting. This is necessary to prevent
+     * locale-specific numeric localisation breaking internal date strings.
+     */
+    const ISO_LOCALE = 'en_NZ';
```
robbieaverill: Isn't the default i18n locale en_US? Maybe we should use that?

src/ORM/FieldType/DBDatetime.php
```diff
@@ -228,6 +228,38 @@ public function getFormatter($dateLength = IntlDateFormatter::MEDIUM, $timeLengt
         return new IntlDateFormatter(i18n::get_locale(), $dateLength, $timeLength);
```
hudhaifas: Probably should replace `i18n::get_locale()` with `DBDate::ISO_LOCALE`, here and `DBDate::getCustomFormatter()`

src/ORM/FieldType/DBDate.php
tractorcow: NZ is better. Also, we don't want a soft-coded default_locale since that can be changed.

src/ORM/FieldType/DBDatetime.php
tractorcow: The getFormatter() relies on the current locale otherwise it won't localise dates for the frontend.

hudhaifas: I see, but I have two points:
*calling $date->Format() for other purposes -below use case- will cause the same problem
`$date = DBDatetime::now();
$anotherJd = GregorianToJD($date->Format('M'), $date->Format('d'), $date->Format('Y'));`

* I think there should be a way to set the numerals system away from the locale, so in my case I want Arabic Numerals with Arabic locale, some may want Hindi Numerals with Arabic locale?

tractorcow: Yes I thought that would be the case.

For those specific methods we should allow a locale to be specified as well. That way you can use ->Format("someformat") (uses current locale) or ->Format("someformate", "en_NZ") (uses format in nz locale).

For methods that internally use Format() to output numeric values (e.g. day of month) we should similarly use the hard-coded ISO_LOCALE.

tractorcow:
I had some obvious bugs in my last PR, pushed a quick fix.

tractorcow:
@hudhaifas I've pushed up a fix to Format() that now allows a $locale to be conditionally provided, as well as including various other bugfixes.

tractorcow:
Oh great it's green. I think it's ok to merge but I should add a test for arabic date formats.

tractorcow:
This is a bit awkward, because it's a serious bugfix but we need to add new methods / args. I may need to soft-code new args with func_get_arg() instead.

tractorcow:
I've fixed semver violations, and added tests that cover the parent issue. It's ready to merge!

src/ORM/FieldType/DBDatetime.php
tractorcow: @hudhaifas you can now pass $locale (e.g. DBDate::ISO_LOCALE) as a second arg to ->Format(), so your use case is now covered. :)

Tests have been added to prove that the issues are fixed, both localising to hindi numerals, as well as internal storage when arabic is set as the parent locale.

robbieaverill:
Previous comment -> different default arguments -> all good

tractorcow:
Green. :D

robbieaverill:
Green here, green underneath the broken blog builds too. LGTM!

Misagh90:
It's happenning in persian number when locale set to Farsi (Persian) for pulish date & Failed Login Count

Misagh90:
4.2.2

MohammaddReza:
4.3

Issue #8112, opened 2018-05-31
Behat test suite is porous


unclecheese:
The behat tests we run on framework are all dependent on the `silverstripe-admin` module. It's possible to get changes merged into `admin`, get a clean CI, merge, and then unknowingly break the framework tests, which were never run.

The only foolproof way around this is to have `silverstripe-admin` run behat tests on all core modules, but that seems expensive and difficult to maintain. Plus, the last thing we need is a slower CI workflow.

Open for discussion...

## Pull requests

- [x] https://github.com/silverstripe/silverstripe-framework/pull/8172
- [x] https://github.com/silverstripe/silverstripe-admin/pull/540
- [x] https://github.com/silverstripe/silverstripe-admin/pull/547

robbieaverill:
I see the benefit in moving the tests, but I disapprove of renaming them. Any module that uses behat will probably use these contexts and will now break on this core version, which will require a new release in order to change the fixture names. I realise it’s not mission critical but it’s still a breaking change and is a pain. Perhaps we could move the tests but leave the namespaces as they were?

unclecheese:
Fair enough. I guess I don't consider the tests part of the API, but I take your point that it could cause pain for consumers of the core Behat contexts.

I do think we need to be mindful of decoupling and keeping things tidy. As an alternative, rather than deal with the dissonance of a `\Framework\` class in `Admin`, why not just move these classes back into `framework`, which will not run any behat tests itself, but be the provider of core behat features for modules that do want to consume them.

robbieaverill:
> I guess I don't consider the tests part of the API, but I take your point that it could cause pain for consumers of the core Behat contexts.

I don't really either, but yeah it will be a pain for module maintainers (like us)

I like your suggestion for keeping the context classes in framework but not running any tests on it!

maxime-rainville:
I'll hold off peer reviewing this until @unclecheese has done is magic.

unclecheese:
Closing per discussion in https://github.com/silverstripe/silverstripe-framework/issues/8112#issuecomment-397064004

maxime-rainville:
@tractorcow helpfully pointed out that most of our modules running behat test reference travis-upload-artifacts.php 5 seconds after I merged the PR.

They will call that file if the behat test fails.
```bash
[...]

PR #8172, opened 2018-06-13
Remove behat tests

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-framework/issues/8112


unclecheese:
Nevermind. That's configurable. Could potentially break others' tests, though.

robbieaverill:
I'm far less concerned about other modules relying on the test assets from this module's behat config - it's really just the context classes that I'd be worried about. Modules should be using their own assets :)

PR #8174, opened 2018-06-14
MINOR Replace use of deprecated getFixtureFactory() with FixtureTestState

robbieaverill:
I've left a couple of places that use getFixtureFactory() in methods that are also deprecated themselves

### DIFF
--- a/src/Dev/SapphireTest.php
+++ b/src/Dev/SapphireTest.php
@@ -283,6 +283,11 @@ class SapphireTest extends PHPUnit_Framework_TestCase implements TestOnly
         $fixtureFiles = $this->getFixturePaths();
 
         if ($this->shouldSetupDatabaseForCurrentTest($fixtureFiles)) {
+            // Assign fixture factory to deprecated prop in case old tests use it over the getter
+            /** @var FixtureTestState $fixtureState */
+            $fixtureState = static::$state->getStateByName('fixtures');
+            $this->fixtureFactory = $fixtureState->getFixtureFactory(static::class);
+
             $this->logInWithPermission('ADMIN');
         }
 
--- a/src/Dev/SapphireTest.php
+++ b/src/Dev/SapphireTest.php
@@ -20,12 +20,12 @@ use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Injector\InjectorLoader;
 use SilverStripe\Core\Manifest\ClassLoader;
+use SilverStripe\Core\Manifest\ModuleResourceLoader;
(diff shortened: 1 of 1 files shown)


src/Dev/SapphireTest.php
```diff
@@ -283,9 +283,6 @@ protected function setUp()
         $fixtureFiles = $this->getFixturePaths();
 
         if ($this->shouldSetupDatabaseForCurrentTest($fixtureFiles)) {
-            /** @var FixtureTestState $fixtureState */
-            $fixtureState = static::$state->getStateByName('fixtures');
-            $this->setFixtureFactory($fixtureState->getFixtureFactory(static::class));
```
tractorcow: We still need this line because some old tests use the prop not just the accessor.

robbieaverill: Can I make that more explicit by assigning `$this->fixtureFactory = $fixtureState->getFixtur...` rather than setting the getting?

tractorcow: Well, you could if you want, but that's what the setter does right.

robbieaverill: Yeah it does, but it also sets the fixture factory to the FixtureTestState which is why I got confused by this initially

robbieaverill: Updated with new commit =)

src/Dev/SapphireTest.php
```diff
@@ -436,7 +438,9 @@ public function setFixtureFactory(FixtureFactory $factory)
      */
     protected function idFromFixture($className, $identifier)
     {
-        $id = $this->getFixtureFactory()->getId($className, $identifier);
+        /** @var FixtureTestState $state */
+        $state = static::$state->getStateByName('fixtures');
```
tractorcow: I wonder if we should just un-deprecate getFixtureFactory(). This logic is repeated so much.

@dhensby what do you think?

dhensby: The problem is that it should be a static method... maybe we just need to change it to static?

robbieaverill: I can un-deprecate the method and make it static if we're happy to do that

dhensby: I don't think making it static is a good idea until 5 because (although very unlikely) any extension that overrides this will break if you make it static

robbieaverill: What would you guys like me to do from here?

robbieaverill:
Bump @tractorcow @dhensby - would be cool to get action points from here

dhensby:
I think it's fine as is

robbieaverill:
All good @tractorcow?

PR #8178, opened 2018-06-14
Revert "API Moving tests to use transactions"

robbieaverill:
This reverts commit ec956a682dce2876322a3bc7c169a8eccec558ec.

Unfortunately this commit breaks quite a few tests in the CWP recipe. Reverting it fixes them.

I haven't spent much time debugging it but it seems like fixtures are not correctly isolated between unit tests, or that they aren't correctly set up and torn down between each test. Some examples of test failures that are caused by this are [here](https://github.com/silverstripe/cwp/issues/105), and reverting this PR fixes them.

Related PRs https://github.com/silverstripe/silverstripe-framework/pull/7839 https://github.com/silverstripe/silverstripe-framework/pull/7848

If merged then we can close #8177 and close #8174


tractorcow:
Instead of reverting transactions we can just update tests where needed.

https://github.com/symbiote/silverstripe-advancedworkflow/pull/364

For instance, the test was incorrectly purging the internal test fixture factory (which is shared across all tests on that class) instead of the database record itself.

robbieaverill:
Recapping a convo about this we had in Slack just now:

I disagree with updating the tests in modules, because they're symptoms of a problem. It seems like test class level metadata is persisting between unit tests, which is not something that happened in SapphireTest with 4.0 or 4.1, and we could update the tests but in the case of https://github.com/symbiote/silverstripe-advancedworkflow/pull/364 I can't see logically why the original test method shouldn't work.

I'd much rather tweak the transactional SapphireTest changes so it doesn't persist fixture metadata between individual test runs. There are plenty of CWP module test failures that are being caused by this change and updating them all would be arduous and also require a minor version bump for each module, meaning we'd need to support multiple versions across framework versions.

robbieaverill:
Yeah, I get that it's a great performance benefit, but we've already identified regressions in behaviour against 4.0 and 4.1 and I don't want this to block us with CWP 2.1's release.

Keen to hear from @dhensby about this overnight =)

dhensby:
Is the blog now the only module failing?

I find it unlikely it's an issue with the core change if all other tests are passing but this one isn't... A state leak would have (and did) show up in core previously and I fixed it. It usually came down to the fact that `parent::tearDown()` was never called...

If @tractorcow can't replicate the failures locally, then I think that also shows it may not be the commit itself... it may just have exposed some other issue

robbieaverill:
We're re-running the CWP integration test suite now to see what the status is

tractorcow:
Yes there are lots of other issues with 4.2 besides test transactions. I only found one regression directly caused by it, and fixed it.

robbieaverill:
Cool, I'm going to close for now since we're preferring to move forward rather than backwards 👍

PR #8179, opened 2018-06-14
BUG Make regression in #7839 safer

tractorcow:
The regression is that clearFixtures() used to be safe in tests, but now it flushes the entire fixture factory for the lifetime of the test class. This will break all tests in the same class that are executed afterwards.

By making it harder to flush the metadata un-upgrade tests will no longer break.

Related https://github.com/symbiote/silverstripe-advancedworkflow/pull/364

Replaces https://github.com/silverstripe/silverstripe-framework/pull/8178

### DIFF
--- a/src/Dev/FixtureFactory.php
+++ b/src/Dev/FixtureFactory.php
@@ -201,15 +201,20 @@ class FixtureFactory
      * If the $class argument is set, limit clearing to items of this class.
      *
      * @param string $limitToClass
+     * @param bool $metadata Clear internal mapping as well as data.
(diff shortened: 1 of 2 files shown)


tractorcow:
We also discussed making FixtureState immutable, or at least restore from a cloned copy by FixtureTestState. That will isolate changes to the object between test methods.

robbieaverill:
I tested this against the cwp/cwp-recipe-search testsuite and it didn't fix the fixture data leaking regression

tractorcow:
Yeah it doesn't fix everything. One thing at a time. :)

PR #8188, opened 2018-06-17
Tweaks on the ArrayLib class

gelysis:
* Implemented improved associative array check using the approach suggested on php.net: http://nz.php.net/manual/en/function.is-array.php#121692
* Improved speed on the recursive array flattening
* Tweaked the other flattening method as well, allowing keys to be preserved


robbieaverill:
Thanks for the PR! Looks good to me

dhensby:
Thanks :)

Issue #8190, opened 2018-06-18
BUG DBText::Summary() Doesn't summarise non-latin based languages. (Chinese, Korean, Japanese) 


caffeineinc:
## Affected Version

Affects SilverStripe Framework 4.1.1 

```Running 1.6.4 (2018-04-13 12:04:24) with PHP 7.1.16 on Darwin / 17.5.0
cwp/agency-extensions               2.0.0                       Base module for the default CWP theme to add features
cwp/cwp-recipe-cms                  2.0.1                       CWP CMS requirements recipe
cwp/cwp-recipe-core                 2.0.1                       CWP core requirements recipe
cwp/cwp-recipe-search               2.0.1                       CWP search requirements recipe
cwp/starter-theme                   2.0.0                       CWP Theme
cwp/watea-theme                     dev-master cd1bdf8          CWP Watea theme
phpunit/phpunit                     5.7.27                      The PHP Unit Testing framework.
silverstripe/asset-admin            dev-bugfix/cwp-2168 7b81608 Asset management for the SilverStripe CMS
silverstripe/recipe-authoring-tools 1.0.0                       Extra tools for CMS authoring in SilverStripe
silverstripe/recipe-blog            1.0.0                       SilverStripe Blog Project Template
silverstripe/recipe-collaboration   1.0.0                       Add extra functionality to enhance CMS user collaboration
silverstripe/recipe-form-building   1.0.0                       A recipe of modules to help you build forms in SilverStripe
silverstripe/recipe-plugin          1.2.0                       Helper plugin to install SilverStripe recipes
silverstripe/recipe-reporting-tools 1.0.0                       Add extra CMS reporting tools to your SilverStripe project
silverstripe/recipe-services        1.0.0                       Add API and content service modules to your SilverStripe project
silverstripe/registry               dev-master d8976e6          Provide search and export interfaces for datasets.
silverstripe/subsites               2.0.2                       Run multiple sites from a single SilverStripe install.
squizlabs/php_codesniffer           3.3.0                       PHP_CodeSniffer tokenizes PHP, JavaScript and CSS files and detect...
tractorcow/silverstripe-fluent      4.0.2                       Simple localisation for Silverstripe
Mon Jun 18-11:58:05:~/Sites/ethnic```

## Description
`DBText::Summary()` doesn't summarise non latin based language (Chinese, Korean etc) paragraphs. East asian languages use a different kind of full stop, and word counts are based on characters instead of latin based "words".  

[...]

PR #8191, opened 2018-06-18
Bugfix #8190 - DBText::Summary() Doesn't summarise non-latin based languages

caffeineinc:
DBText::Summary() doesn't summarise non latin based language (Chinese, Korean etc) paragraphs. East asian languages use a different kind of full stop, and word counts are based on characters instead of latin based "words".

This changes DBText::Summary() to use str_word_count (php5.0) to get the specific word count for a paragraph instead of spliting on spaces and periods. This allows for a more accurate representation cross language.

https://github.com/silverstripe/silverstripe-framework/issues/8190

### DIFF
--- a/src/ORM/FieldType/DBText.php
+++ b/src/ORM/FieldType/DBText.php
@@ -129,12 +129,21 @@ class DBText extends DBString
         // Split on sentences (don't remove period)
         $sentences = array_filter(array_map(function ($str) {
             return trim($str);
-        }, preg_split('@(?<=\.)@', $value)));
-        $wordCount = count(preg_split('#\s+#u', $sentences[0]));
+        }, preg_split('@(?<=\.)|(?<=。)@', $value)));
+        $wordCount = str_word_count($sentences[0]);
 
         // if the first sentence is too long, show only the first $maxWords words
         if ($wordCount > $maxWords) {
-            return implode(' ', array_slice(explode(' ', $sentences[0]), 0, $maxWords)) . $add;
+            // Split the sentence by words (could be kanji, could be english words)
+            $words = str_word_count($sentences[0], $wordArray = 2);
+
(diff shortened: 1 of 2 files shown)


sminnee:
Rebased and added another commit to test my hypothesis that str_word_count() is returning the wrong value on Travis environemtn.

sminnee:
Hey @caffeineinc I've had a bit of a look at this and I'm not yet convinced that `str_word_count()` is a reliable mechanism for this.

 * It tends to treat any unrecognised character as a separate word, including, say Cyrillic or even é.

"Sam Minnée" - 3 words
"текст на кирилица" - 12 words

 * It's hard (i.e. I haven't figured it out yet) to get `setlocale()` to successfully shift to a mode where the text is correctly interpreted

 * Travis seems to think there are 0 words in 新西兰是一个拥有160余种语言的国家，这体现了我们国家的多元性。总结字段也应根据中文表意句号来分割

 * If we managed to solve that, we would need to choose which locale to use for processing Summary() on a per-record basis.

It's possible that we could build some notion of a record's locale, with the fluent module plugging into this, but I think we probably want to get the first 3 points addressed first.

sminnee:
@caffeineinc we probably want to look into this based on this class: http://php.net/manual/en/class.intlbreakiterator.php

php-intl is an SS4 dependency so it's okay to rely on it.

maxime-rainville:
Is there still interest in this? Is this PR worth reviewing or are we better closing it and starting a new one to address #8473?

sminnee:
Yeah I think we can close this unless Simon objects. It doesn't appear to be usable as-is.

maxime-rainville:
I'll lose for now. @caffeineinc if you want to have another go at it let us know and we can reopen.

PR #8193, opened 2018-06-18
ENHANCEMENT Ensure test DB is flushed on either DDL or transaction-disabled tests

### LINKED ISSUE #8182: BUG Issue with transactions polluting test state
## Affected Version

4.2.x-dev

## Description

If you have two unit tests, and one test makes a DDL query (e.g. truncate, rename table) then the FixtureTestState will not be able to rollback the transaction cleanly for the next unit test, resulting in a polluted test state. This pollution is silent.

## Steps to Reproduce

Run versioned unit tests at this commit.

https://github.com/silverstripe/silverstripe-versioned/blob/bb702731c0c8d0abdc193e6dbaf3deba8a2f47cb/tests/php/VersionedTest.php

VersionedTest::testGetIncludingDeleted() will fail. I've tracked the failure due to VersionedTest::testDeletingOrphanedVersions() which calls `$obj->extend('augmentDatabase', $dummy);`

After this test, all the fixtures created prior to that DDL are NOT reset by the transaction, and thus appear in the testGetIncludingDeleted().

## Suggested fixes:

We should detect failures to rollback in FixtureTestState, and trigger a forced rebuild.

Current code:

```php
public function tearDown(SapphireTest $test)
{
    if ($this->testNeedsDB($test)) {
        $test::tempDB()->rollbackTransaction();
    }
}
```

We should do something like

```php
public function tearDown(SapphireTest $test)
{
    if (!$this->testNeedsDB($test)) {
        return;
    }
    try {
        // Note: Rollbacks need to throw exceptions. Currently they silently do nothing
        $test->tempDB()->rollbackTransaction();
    } catch (DatabaseException $ex) {
        // Destroy database and create fresh factory
        $test->tempDB()->kill(); // or something that sufficiently resets the db
        $factory = Injector::inst()->create(FixtureFactory::class);
        $this->fixtureFactories[strtolower(get_class($test))] = $factory;
        $test->setFixtureFactory($factory);
    }
}
```

# PRs:

 - [x] https://github.com/silverstripe/silverstripe-postgresql/pull/86
- [x] https://github.com/silverstripe/silverstripe-framework/pull/8193
 - [ ] add usesTransactions = false to modules (e.g. asset-admin, fulltextsearch)

tractorcow:
Fixes #8182

### DIFF
--- a/src/ORM/Connect/Database.php
+++ b/src/ORM/Connect/Database.php
@@ -669,6 +669,20 @@ abstract class Database
     abstract public function transactionEnd($chain = false);
 
     /**
+     * Return depth of current transaction
+     *
+     * @return int Nesting level, or 0 if not in a transaction
+     */
+    public function transactionDepth()
+    {
+        // Placeholder error for transactional DBs that don't expose depth
+        if ($this->supportsTransactions()) {
+            user_error(get_class($this) . " does not support transactionDepth", E_USER_WARNING);
+        }
+        return 0;
+    }
+
+    /**
      * Determines if the used database supports application-level locks,
      * which is different from table- or row-level locking.
      * See {@link getLock()} for details.
--- a/src/ORM/Connect/Database.php
+++ b/src/ORM/Connect/Database.php
@@ -654,13 +654,17 @@ abstract class Database
      *
      * @param string|boolean $savepoint Name of savepoint, or leave empty to rollback
      * to last savepoint
+     * @return bool|null Boolean is returned if success state is known, or null if
+     * unknown. Note: For error checking purposes null should not be treated as error.
      */
     abstract public function transactionRollback($savepoint = false);
 
     /**
      * Commit everything inside this transaction so far
      *
-     * @param boolean $chain
+     * @param bool $chain
+     * @return bool|null Boolean is returned if success state is known, or null if
+     * unknown. Note: For error checking purposes null should not be treated as error.
      */
     abstract public function transactionEnd($chain = false);
 
@@ -669,7 +673,7 @@ abstract class Database
      * which is different from table- or row-level locking.
      * See {@link getLock()} for details.
      *
-     * @return boolean Flag indicating that locking is available
+     * @return bool Flag indicating that locking is available
      */
     public function supportsLocks()
     {
@@ -681,7 +685,7 @@ abstract class Database
      * See {@link supportsLocks()} to check if locking is generally supported.
      *
      * @param string $name Name of the lock
-     * @return boolean
+     * @return bool
      */
     public function canLock($name)
     {
@@ -703,7 +707,7 @@ abstract class Database
      *
      * @param string $name Name of lock
      * @param integer $timeout Timeout in seconds
-     * @return boolean
+     * @return bool
      */
     public function getLock($name, $timeout = 5)
     {
@@ -715,7 +719,7 @@ abstract class Database
      * (if the execution aborts (e.g. due to an error) all locks are automatically released).
      *
      * @param string $name Name of the lock
-     * @return boolean Flag indicating whether the lock was successfully released
+     * @return bool Flag indicating whether the lock was successfully released
      */
     public function releaseLock($name)
     {
@@ -756,7 +760,7 @@ abstract class Database
      * Determine if the database with the specified name exists
      *
      * @param string $name Name of the database to check for
-     * @return boolean Flag indicating whether this database exists
+     * @return bool Flag indicating whether this database exists
      */
     public function databaseExists($name)
     {
@@ -778,12 +782,12 @@ abstract class Database
      * database if it doesn't exist in the current schema.
      *
      * @param string $name Name of the database
-     * @param boolean $create Flag indicating whether the database should be created
+     * @param bool $create Flag indicating whether the database should be created
      * if it doesn't exist. If $create is false and the database doesn't exist
      * then an error will be raised
-     * @param int|boolean $errorLevel The level of error reporting to enable for the query, or false if no error
+     * @param int|bool $errorLevel The level of error reporting to enable for the query, or false if no error
      * should be raised
-     * @return boolean Flag indicating success
+     * @return bool Flag indicating success
      */
     public function selectDatabase($name, $create = false, $errorLevel = E_USER_ERROR)
     {
--- a/src/ORM/Connect/Database.php
+++ b/src/ORM/Connect/Database.php
@@ -676,7 +676,7 @@ abstract class Database
     public function transactionDepth()
     {
         // Placeholder error for transactional DBs that don't expose depth
-        if ($this->supportsTransactions()) {
+        if (!$this->supportsTransactions()) {
             user_error(get_class($this) . " does not support transactionDepth", E_USER_WARNING);
         }
         return 0;
--- a/src/ORM/Connect/MySQLDatabase.php
+++ b/src/ORM/Connect/MySQLDatabase.php
@@ -360,10 +360,7 @@ class MySQLDatabase extends Database
             return false;
         }
         --$this->transactionNesting;
-        if ($this->transactionNesting <= 0) {
-            $this->transactionNesting = 0;
-            $this->query('COMMIT AND ' . ($chain ? '' : 'NO ') . 'CHAIN');
-        }
+        $this->query('COMMIT AND ' . ($chain ? '' : 'NO ') . 'CHAIN');
         return true;
     }
 
--- a/src/ORM/Connect/MySQLDatabase.php
+++ b/src/ORM/Connect/MySQLDatabase.php
@@ -360,7 +360,10 @@ class MySQLDatabase extends Database
             return false;
         }
         --$this->transactionNesting;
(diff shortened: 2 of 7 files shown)


tractorcow:
Note: We'll need to PR to other modules `protected $usesTransactions = false;` for tests that were failing.

tractorcow:
ok, there's still a bug.

SapphireTest::setUpBeforeClass() has this code.

```
// Build DB if we have objects
        if (static::getExtraDataObjects()) {
            DataObject::reset();
            static::resetDBSchema(true, true);
        }
```

We need to occasionally do this on setUp() (not always setupBeforeClass) so it needs refactoring.

tractorcow:
Yay fixed it.

dhensby:
Postgres is failing for some reason :/

tractorcow:
Yeah that's ok. I'll test postgres locally.

src/ORM/Connect/TempDatabase.php
```diff
@@ -232,7 +245,12 @@ public function resetDBSchema(array $extraDataObjects = [])
         // pgsql doesn't allow schema updates inside transactions
         // so we need to rollback any transactions before commencing a schema reset
         if ($this->hasStarted()) {
-            $this->rollbackTransaction();
```
tractorcow: Ok it turns out this code doesn't work.

Trying

```php
public function resetDBSchema(array $extraDataObjects = [])
    {
        if (!$this->isUsed()) {
            return;
        }
        try {
            $this->rebuildTables($extraDataObjects);
        } catch (DatabaseException $ex) {
            // In case of error during build force a hard reset
            // e.g. pgsql doesn't allow schema updates inside transactions
            $this->kill();
            $this->build();
            $this->rebuildTables($extraDataObjects);
        }
    }
```

@dhensby seem ok?

tractorcow:
```
PHP Fatal error:  Uncaught SilverStripe\ORM\Connect\DatabaseException: Couldn't run query:

SELECT tablename
FROM pg_catalog.pg_tables
WHERE schemaname = $1 AND tablename NOT ILIKE 'pg\\_%' AND tablename NOT ILIKE 'sql\\_%'

ERROR:  current transaction is aborted, commands ignored until end of transaction block in /Users/dmooyman/Sites/ss42test/vendor/silverstripe/framework/src/ORM/Connect/DBConnector.php:64
Stack trace:
#0 /Users/dmooyman/Sites/ss42test/vendor/silverstripe/postgresql/code/PostgreSQLConnector.php(234): SilverStripe\ORM\Connect\DBConnector->databaseError('Couldn't run qu...', 256, 'SELECT tablenam...', Array)
#1 /Users/dmooyman/Sites/ss42test/vendor/silverstripe/framework/src/ORM/Connect/Database.php(168): SilverStripe\PostgreSQL\PostgreSQLConnector->preparedQuery('SELECT tablenam...', Array, 256)
#2 /Users/dmooyman/Sites/ss42test/vendor/silverstripe/framework/src/ORM/Connect/Database.php(222): SilverStripe\ORM\Connect\Database->SilverStripe\ORM\Connect\{closure}('SELECT tablenam...')
#3 /Users/dmooyman/Sites/s in /Users/dmooyman/Sites/ss42test/vendor/silverstripe/framework/src/ORM/Connect/DBConnector.php on line 64
```

WHY

tractorcow: I think we need rollbackRecursive() on database.

tractorcow: It's doing the rollback, but the rollback doesn't seem to actually rollback on postgres.

tractorcow: ok, there is like 6 levels of nesting at this error.

tractorcow: I was dumb. I was doing a `BEGIN` followed IMMEDIATELY by a `ROLLBACK` somehow.

tractorcow: ok the problem is hasStarted is true sometimes when there is no transaction. After a rollback hasStarted needs to stay true, even though the transaction is rolled back. It's a badly documented var.

tractorcow: I'm dropping hasStarted() and using isUsed() instead. Between runs, there will be two states:

 - The DB Is populated, assume that it's ready and safe to use.
 - The DB is killed, assume that it requires a rebuild.

There is no "there but needs a reset state".

tractorcow:
Local testing is VERY promising.

Only this test broke in postgres:

```
Time: 21.31 minutes, Memory: 124.00MB

There was 1 failure:

1) SilverStripe\ORM\Tests\DataListTest::testSortMixedCase
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => 'Bob'
-    1 => 'bonny'
-    2 => 'jane'
-    3 => 'John'
-    4 => 'sam'
-    5 => 'Steve'
+    1 => 'John'
+    2 => 'Steve'
+    3 => 'bonny'
+    4 => 'jane'
+    5 => 'sam'
     6 => 'steven'
 )

/Users/dmooyman/Sites/ss42test/vendor/silverstripe/framework/tests/php/ORM/DataListTest.php:627
/Users/dmooyman/Sites/ss42test/vendor/phpunit/phpunit/phpunit:52

FAILURES!
Tests: 2149, Assertions: 7858, Failures: 1, Skipped: 14, Incomplete: 2.
Damians-Macbook:ss42test dmooyman$ 

```

tractorcow:
Wow green.

This story was SO hard to get working properly... sorry in advance @dhensby if the new logic is hard to follow. =(

dhensby:
Do we know how or why we were getting to so many levels of transaction nesting? It seems like we really shouldn't get more than 2 or 3 levels of transactions deep during a test...

dhensby:
A point worth making, but not something I think needs specifically addressing as we've likely covered 99% of usecases, but the commands that cause "implicit commits" aren't *necassarily* just DDL queries - https://dev.mysql.com/doc/refman/5.7/en/implicit-commit.html

src/ORM/Connect/TempDatabase.php
```diff
@@ -73,14 +68,6 @@ public function isUsed()
         return $this->isDBTemp($selected);
     }
 
-    /**
-     * @return bool
-     */
-    public function hasStarted()
-    {
-        return $this->hasStarted;
-    }
```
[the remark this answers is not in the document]
dhensby: I'm not too concerned about this

dhensby:
I've addressed some of the feedback and made some changes I felt were appropriate

dhensby:
I'm wondering if the nested transaction changes will mean we need to go and update all our DB drivers to match

robbieaverill:
Ok we lost the 💚  😞

src/ORM/Connect/MySQLDatabase.php
```diff
@@ -360,10 +360,7 @@ public function transactionEnd($chain = false)
             return false;
         }
         --$this->transactionNesting;
-        if ($this->transactionNesting <= 0) {
-            $this->transactionNesting = 0;
-            $this->query('COMMIT AND ' . ($chain ? '' : 'NO ') . 'CHAIN');
-        }
```
robbieaverill: I feel like maybe this logic should remain, could explain the test breakages

robbieaverill: I added this back locally and re-tested and it fixes the new failures, pushed a fix to add this back

dhensby: derp

src/ORM/Connect/Database.php
```diff
@@ -654,22 +654,40 @@ public function supportsExtensions($extensions)
...
 
+    /**
+     * Return depth of current transaction
+     *
+     * @return int Nesting level, or 0 if not in a transaction
+     */
+    public function transactionDepth()
+    {
+        // Placeholder error for transactional DBs that don't expose depth
+        if ($this->supportsTransactions()) {
```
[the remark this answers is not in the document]
tractorcow: No, its an error because even though you support transactions, the depth is hard coded to 0.it's a placeholder for making this api abstract.

src/ORM/Connect/TempDatabase.php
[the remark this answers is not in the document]
tractorcow: It wasn't in a stable tag so I'm un-api ing it.

tractorcow:
@dhensby the logic on the nesting method was correct before you changed it. It's ment to error if not overridden eg sqlite. Can you revert.

tests/php/Security/GroupTest.php
```diff
@@ -20,6 +20,11 @@ class GroupTest extends FunctionalTest
         TestMember::class
     ];
 
+    protected function setUp()
+    {
+        parent::setUp();
+    }
```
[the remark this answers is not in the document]
tractorcow: Oops. That was my var dump holder lol

dhensby:
>@dhensby the logic on the nesting method was correct before you changed it. It's ment to error if not overridden eg sqlite. Can you revert.

ah, will do

dhensby:
@maxime-rainville did you use "rebase merge" on this PR? I can't find the merge commit anywhere...

PR #8195, opened 2018-06-18
FIX Default cache state should be `no-cache`

dhensby:
In SS <4.2 the default cache-control headers directives are: `max-age=0, must-revalidate, no-transform`

It was agreed that `no-transform` would be dropped from the directives list as it wasn't really relevant.

These remaining directives are equivalent to `no-cache` not `no-store`, we should set a default state to `no-cache` to keep this behaviour consistent and allow responses to have `ETag`s included.

### DIFF
--- a/src/Control/Middleware/HTTPCacheControlMiddleware.php
+++ b/src/Control/Middleware/HTTPCacheControlMiddleware.php
@@ -26,6 +26,8 @@ class HTTPCacheControlMiddleware implements HTTPMiddleware, Resettable
 
     const STATE_DISABLED = 'disabled';
 
+    const STATE_DEFAULT = 'default';
+
     /**
      * Generate response for the given request
      *
@@ -90,6 +92,9 @@ class HTTPCacheControlMiddleware implements HTTPMiddleware, Resettable
         self::STATE_ENABLED => [
             'must-revalidate' => true,
         ],
+        self::STATE_DEFAULT => [
+            'no-cache' => true,
+        ],
     ];
 
     /**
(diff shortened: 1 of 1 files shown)


chillu:
For context, here's some definitions from [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control): 

 * `no-store`: The cache should not store anything about the client request or server response.
 * `no-cache`: Forces caches to submit the request to the origin server for validation before releasing a cached copy.
 * `no-transform`: No transformations or conversions should be made to the resource. The Content-Encoding, Content-Range, Content-Type headers must not be modified by a proxy. A non- transparent proxy might, for example, convert between image formats in order to save cache space or to reduce the amount of traffic on a slow link. The no-transform directive disallows this.

From [Google](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#no-cache_and_no-store)

> "no-cache" indicates that the returned response can't be used to satisfy a subsequent request to the same URL without first checking with the server if the response has changed. As a result, if a proper validation token (ETag) is present, no-cache incurs a roundtrip to validate the cached response, but can eliminate the download if the resource has not changed.

> By contrast, "no-store" is much simpler. It simply disallows the browser and all intermediate caches from storing any version of the returned response—for example, one containing private personal or banking data. Every time the user requests this asset, a request is sent to the server and a full response is downloaded.

After reading this, I agree that `no-cache` is the better default because it potentially saves bandwidth (although not server computation time).

dhensby:
> After reading this, I agree that no-cache is the better default because it potentially saves bandwidth (although not server computation time).

yep, that's right; but most importantly, that's how it currently works, so we keep behaviour

dhensby:
> You still have $defaultState = self::STATE_DISABLED rather than the new state, so your setting doesn't have any effect as a default.

whoops

dhensby:
>- `no-cache`: Forces caches to submit the request to the origin server for validation before releasing a cached copy.

I really want to emphasise and make clear that this is the same as the `must-revalidate` directive; so this is not a change in behaviour

PR #8198, opened 2018-06-19
FIX Switch to Trusty in Travis

robbieaverill:
Follow on from #8161 to switch precise to trusty, to see what breaks. Composer won't install correctly in my fork's Travis set up for 3.x builds so couldn't test it outside of a PR (easily). _Edit:_ OK I worked out how to build in our fork. Behat tests are failing, will see what I can do quickly...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: php
 
-dist: precise
+dist: trusty
 
 sudo: false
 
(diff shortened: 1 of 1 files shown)


robbieaverill:
So looks like older versions of Selenium have a bug where duplicate or incorrectly ordered localhost definitions in /etc/hosts cause Selenium to break. I've added a workaround to .travis.yml so we can get it running, based on the workaround posted in https://github.com/travis-ci/travis-ci/issues/3054

Selenium issue: https://github.com/seleniumhq/selenium/issues/2074

dhensby:
nice - let's get it in :)

Issue #7966, opened 2018-03-28
[ORM] SS_DATABASE_SUFFIX doesn't work


kinglozzer:
Documented here: https://docs.silverstripe.org/en/4/getting_started/environment_management/#core-environment-variables, `CoreKernel` doesn’t take it into account at all

tractorcow:
It seems to do so here.

```php
   protected function getDatabasePrefix()
    {
        return Environment::getEnv('SS_DATABASE_PREFIX') ?: '';
    }
    
    protected function getDatabaseName()
    {
    	// ...

        // Check environment
        $database = Environment::getEnv('SS_DATABASE_NAME');

        if ($database) {
            return $this->getDatabasePrefix() . $database;
        }
```

Let me know if there's something wrong here and I can re-open this issue. ;)

tractorcow:
quick test
```
SS_DATABASE_PREFIX="superprefix_"
SS_DATABASE_NAME="myname"
```


dev/build:

```
Building database superprefix_myname using SilverStripe\ORM\Connect\MySQL 5.5.5-10.2.8-MariaDB

```

kinglozzer:
Hahaha, doh, I meant suffix!

tractorcow:
Oh....

tractorcow:
I can see that if we fixed this, it could possibly leave a lot of websites with a broken database, as the previously un-suffixed databases would suddenly work.

What if we changed this in 5.x to work again?

kinglozzer:
Yeah 5.x is fine by me

dhensby:
oh man, I'm not so sure about leaving this out of the 4.x line... if someone "fixed" it by keeping the suffix defined and added it to the DB name too... well, they are asking for trouble..

dhensby:
> Yes but if we fixed it in 4.x, it would result in an upgrade having a double-suffixed database (and apparent blank one would be created).

how, though? by someone doing what I said and double defining it?

tractorcow:
If you have
```
SS_DATABASE_NAME="mydb_suffux"
SS_DATABASE_SUFFIX="_suffix"
```

If we "fixed" this bug, the full DB name would switch from `mydb_suffux` to `mydb_suffux_suffix` wouldn't it?

dhensby:
yep, but that's pretty mental set up ;)

sminnee:
[...]

PR #8205, opened 2018-06-21
Update 03_Environment_Management.md

DorsetDigital:
Remove reference to DATABASE_SUFFIX - this doesn't appear to be used anywhere in the installer or core kernel.

### DIFF
--- a/docs/en/00_Getting_Started/03_Environment_Management.md
+++ b/docs/en/00_Getting_Started/03_Environment_Management.md
@@ -74,7 +74,6 @@ SilverStripe core environment variables are listed here, though you're free to d
 | `SS_DATABASE_USERNAME`| The database username (mandatory).|
 | `SS_DATABASE_PASSWORD`| The database password (mandatory).|
 | `SS_DATABASE_PORT`|     The database port.|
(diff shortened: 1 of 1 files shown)


kinglozzer:
Related https://github.com/silverstripe/silverstripe-framework/issues/7966

DorsetDigital:
That looks like a fun conversation :-)    I think we should probably make the docs match the current state of the code, even if the functionality is going back in at some point.

dhensby:
ergh :( this is why we should have fixed the bug when it was reported 3 months ago :(

Fixing something not working when it should have been is not a regression!

maxime-rainville:
I think Damian's [original point still stands](https://github.com/silverstripe/silverstripe-framework/issues/7966#issuecomment-378745244). I we fix the original issue, there's a good chance that somebody somewhere will have a terrible day because their live system suddenly expects their DB to be called `SS_something_suffix_suffix`.

In this context, I think it would make more sense to just remove the suffix option from the doc.

kinglozzer:
This _may_ be replaced by https://github.com/silverstripe/silverstripe-framework/pull/8470 if that PR is accepted

sminnee:
Recommend that we cancel this in favour of #8470

PR #8206, opened 2018-06-21
[WIP] Make PDOQuery::getIterator() a generator

kinglozzer:
The one-statement-at-a-time limitation mentioned [here](https://github.com/silverstripe/silverstripe-framework/blob/4.2.0-beta1/src/ORM/Connect/PDOQuery.php#L28-L30) is apparently only applicable when PHP is compiled with `libmysqlclient`. `mysqlnd` is highly recommended, and has apparently been the default since PHP 5.4: http://php.net/manual/en/mysqlinfo.library.choosing.php. ~I’m yet to get a `libmysqlclient` environment set up to test on, but if we really want to keep `libmysqlclient` support then hopefully there is some way to detect the library in use, and use the existing `fetchAll()` upfront if needed.~

I’ve not benchmarked before/after removing prepared statement caching yet, but I suspect the difference will be small. The number of repeated queries is small for most sites, and the roundtrip to prepare them is fast. The reason for that change is highlighted in one test specifically:

- https://github.com/silverstripe/silverstripe-framework/blob/4.2.0-beta1/tests/php/ORM/PDODatabaseTest.php#L29-L32 binds a parameter `0` into the query
- https://github.com/silverstripe/silverstripe-framework/blob/4.2.0-beta1/tests/php/ORM/PDODatabaseTest.php#L34-L37 has identical SQL, so re-uses the prepared statement but overrides the bound value to `2`

Currently, this only “works” accidentally because we’re calling `fetchAll()` as soon as the query is executed. You can’t clone a `PDOStatement` as a workaround either 😞.

The benefit of this change is memory savings, especially for large result sets, as currently every single row is loaded into memory at once, for every active query.

### DIFF
--- a/tests/php/Dev/CsvBulkLoaderTest.php
+++ b/tests/php/Dev/CsvBulkLoaderTest.php
@@ -216,8 +216,8 @@ class CsvBulkLoaderTest extends SapphireTest
         // Test nested setting of relation properties
         $contractAmount = DBField::create_field('Currency', $compareRow[5])->RAW();
         $this->assertEquals(
-            $testPlayer->Contract()->Amount,
             $contractAmount,
+            $testPlayer->Contract()->Amount,
             'Setting nested values in a relation works'
         );
 
--- a/tests/php/Dev/CsvBulkLoaderTest.php
+++ b/tests/php/Dev/CsvBulkLoaderTest.php
@@ -214,10 +214,10 @@ class CsvBulkLoaderTest extends SapphireTest
         $this->assertEquals($testPlayer->ContractID, $testContract->ID, 'Creating new has_one relation works');
 
         // Test nested setting of relation properties
-        $contractAmount = DBField::create_field('Currency', $compareRow[5])->RAW();
+        $contractAmount = DBField::create_field('Currency', $compareRow[5])->Nice();
         $this->assertEquals(
             $contractAmount,
-            $testPlayer->Contract()->Amount,
+            $testPlayer->Contract()->dbObject('Amount')->Nice(),
             'Setting nested values in a relation works'
         );
 
--- a/tests/php/ORM/DBMoneyTest.php
+++ b/tests/php/ORM/DBMoneyTest.php
@@ -292,7 +292,7 @@ class DBMoneyTest extends SapphireTest
             )->value()
         );
         $this->assertEquals(
-            '1.23',
+            '1.2300',
             DB::query(
                 sprintf(
                     'SELECT "MyMoneyAmount" FROM "MoneyTest_DataObject" WHERE "ID" = %d',
--- a/tests/php/ORM/DataObjectTest.yml
+++ b/tests/php/ORM/DataObjectTest.yml
@@ -49,7 +49,7 @@ SilverStripe\ORM\Tests\DataObjectTest\Player:
   captain1:
     FirstName: Captain
     Surname: Zookeeper
-    ShirtNumber: 007
+    ShirtNumber: '007'
     FavouriteTeam: =>SilverStripe\ORM\Tests\DataObjectTest\Team.team1
     Teams: =>SilverStripe\ORM\Tests\DataObjectTest\Team.team1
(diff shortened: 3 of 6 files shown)


tests/php/Dev/CsvBulkLoaderTest.php
```diff
@@ -214,10 +214,10 @@ public function testLoadWithCustomHeaderAndRelation()
         $this->assertEquals($testPlayer->ContractID, $testContract->ID, 'Creating new has_one relation works');
 
         // Test nested setting of relation properties
-        $contractAmount = DBField::create_field('Currency', $compareRow[5])->RAW();
+        $contractAmount = DBField::create_field('Currency', $compareRow[5])->Nice();
         $this->assertEquals(
-            $testPlayer->Contract()->Amount,
             $contractAmount,
+            $testPlayer->Contract()->dbObject('Amount')->Nice(),
```
kinglozzer: This [was failing](https://travis-ci.org/kinglozzer/sapphire/jobs/395006181) because the database query was returning `20000.00` while the manually constructed `DBField` instance was returning `20000`. I don’t know why, but that “regression” only shows up when the statement caching is removed. Given that the value is a `DECIMAL(9, 2)` I’d argue that `20000.00` is more correct than `20000` which is currently returned.

tests/php/ORM/DBMoneyTest.php
```diff
@@ -292,7 +292,7 @@ public function testWriteToDataObject()
             )->value()
         );
         $this->assertEquals(
-            '1.23',
+            '1.2300',
```
kinglozzer: As above, this only regressed once statement caching was removed. However, I’d argue that the new test is correct and the old one isn’t: `MyMoneyAmount` is a `DECIMAL(19, 4)`.

tests/php/ORM/DataObjectTest.yml
```diff
@@ -49,7 +49,7 @@ SilverStripe\ORM\Tests\DataObjectTest\Player:
   captain1:
     FirstName: Captain
     Surname: Zookeeper
-    ShirtNumber: 007
+    ShirtNumber: '007'
```
kinglozzer: Another “bug” exposed by removing statement caching - https://travis-ci.org/kinglozzer/sapphire/jobs/394634700#L810-L816

kinglozzer:
I’ve manually compiled PHP 7.2.7 on Ubuntu 16.04.4 with `libmysqlclient` and I can’t recreate the original issue with “nested” queries/multiple result sets.

tractorcow:
I forgot exactly how to reproduce the issue, but it caused segfaults in my original testing. That's why we had that comment on the statements being un-reusable. Maybe php 7 fixes it?

kinglozzer:
I did eventually find the original error, it rang a bell from when we first looked at using generators:

```
[Exception]                                                                                                                                                                                                                                 
SQLSTATE[HY000]: General error: 2014 Cannot execute queries while other unbuffered queries are active.  Consider using PDOStatement::fetchAll().  Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.
```

I’ve done some more digging on this, including through PHP’s source code. The reason I wasn’t encountering the error (with either `mysqlnd` or `libmysqlclient`) is because `PDO::MYSQL_ATTR_USE_BUFFERED_QUERY` defaults to true (it has done since [PHP 5.2](https://github.com/php/php-src/commit/e7c2bd6e4062ec6cb0f80f618a991a27b6de6959)).

That sounds fine, and we could force it to `true`, but it’s not portable for other non-MySQL drivers. It also doesn’t actually save any memory as in buffered mode the results are still stored in PHP’s memory internally, even if you don’t call `$statement->fetchAll();` - it just does it [silently in the background](https://github.com/php/php-src/blob/a89eb2f101bf0ca6f92ee3160fa46de0bd6f31b8/ext/pdo_mysql/mysql_statement.c#L140-L144).

I’ve scratched my itch here, and am satisfied we can’t “fix” this. I think the statement caching bugs are real, but I’ll open an issue/PR for those separately.

PR #8212, opened 2018-06-24
DOCS Consistent app/ folder and composer use

chillu:
- Stronger wording around "use composer"
- Consistent domain and email address naming
- Removed example for publishing non-composer modules (those shouldn't be encouraged)
- Removed instructions for installing modules from archives

[ci skip]


lukereative:
@chillu Is there any consideration around the fact that myapp.com (it redirects to android.myapp.com) is an actual domain whereas example.com is explicitly for use as a dummy domain. Personally I think example.com is good because it means if it somehow gets put in executed code it's not a "real" domain.

chillu:
Very good point Luke! Fixed now. https://en.wikipedia.org/wiki/Example.com

PR #8213, opened 2018-06-24
FIX FormField::Link works when no form is currently set

robbieaverill:
Most places where `$this->form` is used in FormField are checked for truthy first, this part of the code fails when requesting a form schema for a readonly form (e.g. history viewer) when an UploadField (that uses `$this->Link('upload')`) is cast as readonly with no form set.

This PR returns null from `FormField::Link` if there isn't a form available to the instance yet.

Related: https://github.com/silverstripe/cwp/issues/115

### DIFF
--- a/src/Forms/FormField.php
+++ b/src/Forms/FormField.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\Forms;
 
+use LogicException;
 use ReflectionClass;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\RequestHandler;
@@ -355,14 +356,20 @@ class FormField extends RequestHandler
     }
 
     /**
-     * Return a link to this field.
+     * Return a link to this field
      *
      * @param string $action
-     *
      * @return string
+     * @throws LogicException If no form is set yet
      */
     public function Link($action = null)
     {
+        if (!$this->form) {
+            throw new LogicException(
+                'Field must be associated with a form to call Link(). Please use $field->setForm($form);'
+            );
+        }
+
         $link = Controller::join_links($this->form->FormAction(), 'field/' . $this->name, $action);
         $this->extend('updateLink', $link, $action);
         return $link;
--- a/tests/php/Forms/FormFieldTest.php
+++ b/tests/php/Forms/FormFieldTest.php
@@ -384,4 +384,22 @@ class FormFieldTest extends SapphireTest
         $this->assertFalse($field->hasClass('banana'));
         $this->assertTrue($field->hasClass('cool-BAnana'));
     }
+
+    public function testLinkWithForm()
+    {
+        $field = new FormField('Test');
+        $form = new Form(null, 'Test', new FieldList, new FieldList);
+        $form->setFormAction('foo');
+        $field->setForm($form);
+        $this->assertSame('foo/field/Test/bar', $field->Link('bar'));
+    }
+
+    /**
+     * @expectedException \LogicException
+     */
+    public function testLinkWithoutForm()
+    {
+        $field = new FormField('Test');
+        $field->Link('bar');
+    }
 }


NightJar:
Seems good in that it stops fatal errors, and that's ideal. But I feel there should be a follow up issue opened as to what the effect is (if any) of having an invalid link (or is it somehow valid?) rendered out when no form is set.

kinglozzer:
I’m a little hesitant about this change, given it’ll return an obviously incorrect URL. This is usually solved by `$field->setForm($form);` or `$fieldList->setForm($form);`, is that not possible in this case?

dhensby:
I agree with @kinglozzer. under what circumstances would a working URL be given if there was no form?

robbieaverill:
Yeah, fair enough. You can still do `TextField:create('foo')->Link()` to cause this error, which I don't like, but I agree that it'd be better to fix it in a different place

robbieaverill:
>My instinct at this stage would be lean towards yes. So I don't think we should throw an exception. However we should be upfront about the fact that the returned link is meaningless by returning null.

Yeah, that was the intention of this PR - although returning invalid data is worse than returning null. I'll update this to return null if no form is available. I've fixed the original issue in the implementation rather than here :)

maxime-rainville:
I'll give a chance to @kinglozzer and @dhensby to disagree before merging.

robbieaverill:
Yeah great, thank you

kinglozzer:
Hmm, `null` is _better_, but I’m still not sold 😛. If someone uses `$field->Link()`, or `{$Link}` in a template, it’ll silently return nothing and they’ll have to dig through to try to find out why.

My preference would definitely be an exception. By calling `Link()` without a form set, you’re doing something that’s...



Sounds like a `LogicException` candidate to me. Given that’d currently result in a fatal error, I don’t see any BC issues by introducing it in a patch release :)

dhensby:
Yeah, I agree with @kinglozzer again.

At the moment this is an API change (return type is changing) and what would have been a fatal error is no longer an error at all. Even though we're becoming more permissive it's still a change in behaviour and one that I'm not sure is warranted.

In all honesty, if you've fixed the problem upstream and agree it was an implementation problem, then I don't see that this patch is warranted

robbieaverill:
Good feedback here, I'll switch it to an exception instead :)

maxime-rainville:
Looking at the rest of `FormField`, `securityTokenEnabled()` and `Required()` both need a form set to perform their task. They both check to make sure `$this->form` is set and return a non-value if it isn't.

I can't imagine anyone out there would rely on catching a `Call to a member function on null` fatal error to determine if a FormField is orphan or not. So I would consider it a bug fix as opposed to behavior change.

If we can't get a consensus on this specific PR, it might be best to have a separate RFC and take holistic view of how orphan FormField should behave.

robbieaverill:
State of play with this is that it's throwing an exception if the form isn't set yet. Are we happy to merge that, or shall I close this and leave it as it is?

src/Forms/FormField.php
```diff
@@ -355,14 +356,18 @@ protected function setupDefaultClasses()
...
      *
      * @param string $action
-     *
      * @return string
+     * @throws LogicException If no form is set yet
      */
     public function Link($action = null)
     {
+        if (!$this->form) {
+            throw new LogicException('No form is set yet');
```
kinglozzer: Happy with this, can you just add a little more detail to the message? Something like

`Field must be associated with a form to call Link(). Please use $field->setForm($form);`

dhensby:
OK - I'm concerned we are moving from a non-fatal error to a fatal error in a patch release here. This is something I really don't think we should be doing...

I'd prefer to leave it as is or replace it with a warning and skip the call to `$this->form->FormAction()`. It seems that the `updateLink` hook would still be called even if there was no form assigned at the moment(?)

robbieaverill:
It’s already fatal if you have no form, this is just more structured than allowing a php error.

Either way there’s no strong desire for this change and plenty of friction against it, so not worth it :-)

dhensby:
Ah, ok, if it's fatal, then this is ok - sorry, just looking at the error report I saw "alert" instead of "fatal" so made an assumption!

I'm happy with throwing an exception, then... let's get this over the line

robbieaverill:
Thanks all!

PR #8026, opened 2018-04-23
ENHANCEMENT Backport of querystring work to 3.x

tractorcow:
Parent issue https://github.com/silverstripe/silverstripe-versioned/issues/145

PR #8214, opened 2018-06-24
DOCS Upgrading 3.x docs on session stage param

chillu:
[ci skip]


chillu:
Forgotten as part of https://github.com/silverstripe/silverstripe-framework/pull/8026/

PR #8217, opened 2018-06-25
Allow thousands & decimals separator config values

### LINKED ISSUE #8216: Currency DBField doesn't allow setting different thousands & decimals separators
## Affected Version

3.*, 4.*

## Description

In some locales, currencies use different separators, eg in Dutch (Netherlands) we use '€1.000,00' (as opposed to '$1,000.00'). The Currency field internally uses number_format(), this allows setting both characters. Allowing config values for both on Currency class and feeding them as 3rd & 4th agument fixes this (PR on its way).

micschk:
FIX #8216

### DIFF
--- a/src/ORM/FieldType/DBCurrency.php
+++ b/src/ORM/FieldType/DBCurrency.php
@@ -22,6 +22,18 @@ class DBCurrency extends DBDecimal
      * @var string
      */
     private static $currency_symbol = '$';
+    
+    /**
+     * @config
+     * @var string
+     */
+    private static $dec_point = '.';
+    
+    /**
+     * @config
+     * @var string
+     */
+    private static $thousands_sep = ',';
 
     public function __construct($name = null, $wholeSize = 9, $decimalSize = 2, $defaultValue = 0)
     {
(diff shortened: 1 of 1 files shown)


dhensby:
Hmm, wouldn't it be better if there was some kind of locale formatting we could use like we do for datetime values? rather than just setting this as a config setting?

micschk:
Good point. Would be nice, but maybe should be implemented through a separate FormatForLocale() method (like SS_DateTime::FormatFromSettings)? 
Otherwise it seems you'd lose some of the control, eg to place the currency symbol in front or after, wether to separate thousands or not, and wether or not to display decimals.

Also, DateTime gets the format from the Member object. Currency should probably get it from the browser instead, which would make it a slightly bigger undertaking (or do we already have some kind of negotiated locale stored somewhere by default)?

micschk:
Some references: 
https://akrabat.com/using-phps-numberformatter-to-format-currencies/ 
http://php.net/manual/en/numberformatter.create.php 
http://php.net/manual/en/numberformatter.formatcurrency.php

micschk:
Something like this? (quick draft in SS3 format)
```php
    /**
     * Returns the number as a locale-formatted currency, eg nl_NL: € 1.000,00 / ru_RU: 1 000,00€
     * @param null $locale eg 'nl-NL' (default: Member's locale, i18n current locale)
     */
    public function FormatForLocale($locale = null)
    {
        if(!$locale) {
            // get Member's locale or else default/current i18n locale
            $locale = (Member::currentUserID() ? Member::currentUser()->Locale : i18n::get_locale());
        }

        $formatter = new \NumberFormatter($locale, \NumberFormatter::CURRENCY);
        $formatter->parseCurrency($this->config()->currency_symbol, $ISOcurrency);

        return $formatter->formatCurrency($this->value, $ISOcurrency);
    }
```

dhensby:
Perhaps, yes, but we wouldn't be accepting enhancements like this to 3.x anymore.

PR #8223, opened 2018-06-26
FIX remove personal information from password reset confirmation screen

dhensby:
Things to address:

1. What do we need to do about translations, those relying on `$email` being passed in will be broken / no longer make sense.
2. Is the wording of this acceptable? Do we want to make it clearer that a reset email will only be sent IF an account with that email exists?
3. Is this the right branch to target?


robbieaverill:
>What do we need to do about translations, those relying on $email being passed in will be broken / no longer make sense.

I'd suggest creating new ones and removing the old ones from the English file

>Do we want to make it clearer that a reset email will only be sent IF an account with that email exists?

I'd be for that

>Is this the right branch to target?

Depends if we want to treat this as a bug fix or not :) I'm happy for this to be a patch change/bug fix

dhensby:
>>What do we need to do about translations, those relying on $email being passed in will be broken / no longer make sense.
>
> I'd suggest creating new ones and removing the old ones from the English file

I'm not sure this really solves the problem at hand and just makes things a bit more noisy. We will still have translations that aren't right

>>Do we want to make it clearer that a reset email will only be sent IF an account with that email exists?
>
>I'd be for that

Any suggestions on wording? I can't think of anything eloquent :P

robbieaverill:
Well, by updating the original any translations will start to include `{email}` in them which will be v. wrong, unless Transifex detects that the source string has changed and invalidates the other translations for it (in which case it's fine!)

Re wording - the request password form says this:

>Thank you! A reset link has been sent to 'sdfds@example.com', provided an account exists for this email address.

So maybe we could use similar wording. Double up on "provided" then though! Argh - halp @sachajudd

sachajudd:
I think "Thank you! A reset link has been sent, provided an account exists for this email address." would be sufficient enough? @robbieaverill

dhensby:
> Well, by updating the original any translations will start to include {email} in them which will be v. wrong, unless Transifex detects that the source string has changed and invalidates the other translations for it (in which case it's fine!)

I don't know exactly how it works, so I've created a new key

I've also reworded the message :)

chillu:
> Is this the right branch to target?

I've moved this from the `4.0` to `4.2` branch. It's a bugfix (can be included in a patch release), but not high priority enough to backport it to older minor releases.

chillu:
`lostpassword.feature` moved to admin in 4.2. It probably shouldn't have been moved, @unclecheese ? For now, I've sent an accompanying PR there: https://github.com/silverstripe/silverstripe-admin/pull/561

dhensby:
thanks @chillu

PR #8231, opened 2018-07-02
MINOR: DataObject->fieldLabels() will only return an array

### LINKED ISSUE #3028: DataObject->fieldLabels() php docs wrong?
The docblock of `DataObject->fieldLabels()` wants to tell us that it returns array or string:

```
* @return array|string Array of all element labels if no argument given, otherwise the label of the field
```

however, in the code I can find no indication that would support that claim. am I missing something?

https://github.com/silverstripe/silverstripe-framework/blob/90ba514cda604cbd8e97537e7d95a8188dd45d63/model/DataObject.php#L3279

lukereative:
Fixes #3028

### DIFF
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -3495,7 +3495,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
      *
      * @param boolean $includerelations a boolean value to indicate if the labels returned include relation fields
      *
-     * @return array Array of all element labels if no argument given, otherwise the label of the field
+     * @return array Array of all element labels
      */
     public function fieldLabels($includerelations = true)
     {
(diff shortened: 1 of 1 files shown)


src/ORM/DataObject.php
```diff
@@ -3495,7 +3495,7 @@ public function searchableFields()
      *
      * @param boolean $includerelations a boolean value to indicate if the labels returned include relation fields
      *
-     * @return array|string Array of all element labels if no argument given, otherwise the label of the field
+     * @return array Array of all element labels if no argument given, otherwise the label of the field
```
robbieaverill: The "otherwise" part is the bit that was where it said it'd return a string right? Should we remove that too?

lukereative: Done

PR #8233, opened 2018-07-02
HTTP Caching fixes :/

dhensby:
This PR consists of:

1. Backport of ~#8165~ https://github.com/silverstripe/silverstripe-framework/pull/8195 (which should have gone into 4.2 not 4)
2. Making the defaultState config a `private static` (the `@config` tag is not working for me locally)

I've noticed during testing that if you have a session active, the default state changes to `private, must-revalidate` which doesn't seem right to me when it should probably hold the current state and simply add `private` and `must-revalidate` to the current state...


dhensby:
I'd also query how `privateCache()` works or, at least, how we make requests private when there's a session. I think we should probably only add the "private" directive instead of remove all of the developer's preferences ...

Issue #7739, opened 2018-01-08
[FORMS] ListboxField with enabled multiple attributes is not filtering properly in 3.5.6+


zemiacsik:
After upgrading from 3.5.5 to 3.5.6 (or 3.6.*) I think search filter is not working properly.
I have `ListboxField->setMultiple(true)` and when I 'Apply Filter' with selected multiple options only the last chosen is taken into account.

E.g. there is a `Status` field (enum of `DataObject`), I select `['New' 'Open']` and apply filter then in field will last only `['Open']` and in the url params is `?q[Status][]=Open`, before it was `?q[Status][]=New&q[Status][]=Open`.

Or am I missing something?

maxime-rainville:
I just double checked with a ListboxField on SS4 and this bug doesn't seem to exists there.

Unfortunately, [Silverstripe CMS 3 has entered limited support in June 2018](https://www.silverstripe.org/blog/update-on-silverstripe-5-x/). This means we'll only be fixing critical bugs and security issues for Silverstripe CMS 3 going forward.

You can read the [Silverstripe CMS Roadmap](https://www.silverstripe.org/software/roadmap/) for more information on our support commitments.

PR #8236, opened 2018-07-03
fix for handling arrays in url params when filter is used

zemiacsik:
when filter is used and there is an array in url query params (i.e. Listbox with multiple options selected), then is clicked on record (edit) and you got redirected, params in url will get broken
fix this by building query on first match in js object to serialize it correctly to prevent losing values

### DIFF
--- a/admin/javascript/lib.js
+++ b/admin/javascript/lib.js
@@ -183,7 +183,9 @@
 					parts = search ? search.split( '&' ) : [], i, tmp;
 				for(i=0; i < parts.length; i++) {
 					tmp = parts[i].split( '=' );
-					params[tmp[0]] = tmp[1];
+					// if key already exists, do not override its value but try to build url query instead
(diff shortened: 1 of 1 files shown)


zemiacsik:
This also solves [#7739](https://github.com/silverstripe/silverstripe-framework/issues/7739).

dhensby:
Hmm - how critical is this? Does it warrant 3.7 merge? I'd tentatively say yes...

zemiacsik:
I noticed this only when I was using ListboxField in ModelAdmin filter, so it depends who and how uses SS. It solves that "// TODO Handle repeating elements (e.g. arr[]=one&arr[]=two)".
I have tried it with SS 3.7.1 and it worked properly.

ScopeyNZ:
>  It solves that "// TODO Handle repeating elements (e.g. arr[]=one&arr[]=two)".

Would you mind updating the PR to remove that TODO given this solves that?

zemiacsik:
Yes, sure.

PR #8238, opened 2018-07-03
Make column work

alex-dna:
The ColumnCount is set on the CompositeField class, and not on its FieldList. So to get the ColumnCount, we need to use $Top to get out of the loop.


dhensby:
Thanks for spotting this!

Issue #6202, opened 2016-10-19
Global archive view ("Trashcan")


sminnee:
As a user, I want to be able to intuitive retrieve archived records.

## Acceptance Criteria

- I can see which records have been archived (deleted from both draft & published) in a central place
- I can choose the model to view archived records for
- All versioned models are available to choose from (works with dozens of models)
- I can page through long lists
- I have enough context to tell from the list where a record belongs to (breadcrumbs for pages and files, container page for a block) - unless the record is orphaned
- By default, items are shown sorted by the date they were deleted, most recent first
- I can see who archived a record, on overview and detail views
- I can see when a record was archived, on overview and detail views
- I can restore a record without clicking through to the edit view (some records won't have an edit view configured)
- I can understand why records can't be permanently deleted (no "empty the trash" feature)
- I can click through to a readonly "edit view" (e.g. in the "pages" section) where the record can be viewed and restored. Requires correct setup by devs (`DataObject->EditLink()`)
- If a record needs to be restored in a hierarchy that no longer exists, it will restore to the top level (mainly pages and files)
- If a record clashes on its primary identifier (e.g. filename, email), the process will auto-rename to avoid the clash and the user gets notified
- After restoring a record, it's clear to the author where it went (hierarchy location)
- I am notified when a restore is successful
- Optional: I can view the readonly "edit view" in context of the new archive section, and don't get redirected to a different section for it

## Notes

 * Designs at https://invis.io/FTG16IYV5KB#/299554599_Archive_-_Global_Tab_-_Menu
 * Excludes searching (separate card)
 * Excludes solving the "archive" vs. "delete" vs. "trash" discussion (see https://github.com/silverstripe/silverstripe-cms/issues/1575) - it's an easy technical fix if we decide to switching wording
 * Excludes bulk restore (separate card)
 * Should reuse as much of ModelAdmin's capabilities as possible to simplify implementation
 * Shouldn't show a combined list of multiple record types since that's too performance intensive to query
 * Centralising this view is an attempt to limit the complexity of different UI implementations
 * This is relatively high priority because we don't have a way to list archived records in the blocks or files UI at the moment. For blocks, you can go through the history of the page containing the block and restore from there, a file and any other records managed through GridField only are "lost" at this point.
* Excludes "Undo" action for after archiving or restoring

## Related

 * [View archived records in ModelAdmin and GridField](https://github.com/silverstripe/silverstripe-versioned/issues/91)
[...]

PR #8240, opened 2018-07-03
NEW Archive admin to manage archived records

lukereative:
Allows GridFieldViewButton to be added to action menu.

**Implements:** https://github.com/silverstripe/silverstripe-framework/issues/6202

**Sibling PRs:**
- https://github.com/silverstripe/silverstripe-admin/pull/559
- https://github.com/silverstripe/silverstripe-versioned/pull/168
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/29 #


src/Forms/RestoreAction.php
```diff
@@ -0,0 +1,155 @@
...
+        $archivedItem = Versioned::get_latest_version($classname, $id);
+
+        if (!$archivedItem) {
+            return new ValidationException($classname . " #$id not found", 400);
+        }
+
+        if (method_exists($archivedItem, 'doRestoreToStage')) {
+            $restoredItem = $archivedItem->doRestoreToStage();
+        } else {
+            $archivedItem->writeToStage(Versioned::DRAFT);
```
maxime-rainville: I think you can get away with just calling `write()`. That's [what I told people to do in the version doc](https://docs.silverstripe.org/en/4/developer_guides/model/versioning/#restoring-an-archived-version).

src/Forms/RestoreAction.php
```diff
@@ -0,0 +1,155 @@
...
+        $message = self::getRestoreMessage($item, $restoredItem, $changedLocation);
+
+        return $message;
+    }
+
+    /**
+     * Returns a message which notifies the user of a successful restoration
+     * and if anything has changed
+     *
+     * @param $record
```
maxime-rainville: Missing parameters

src/Forms/RestoreAction.php
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Determines whether this record can be restored to it's original location
+     *
+     * @param $record
+     * @return bool
+     */
+    public static function shouldRestoreToRoot($record)
+    {
+        if ($parentID = $record->ParentID) {
```
maxime-rainville: Would this logic only be relevant if the record use the `Hierarchy` extensions. I think it would be better to go off that other wise it might produce weird results for object that manually define a Parent attribute.

src/Forms/GridField/GridFieldRestoreAction.php
```diff
@@ -0,0 +1,198 @@
...
+    {
+        $isOnDraft = $record->isOnDraft();
+        $isPublished = $record->isPublished();
+        $canEdit = $record->canEdit();
+
+        if ($canEdit && !$isOnDraft && !$isPublished) {
+            $restoreToRoot = RestoreAction::shouldRestoreToRoot($record);
+
+            $title = $restoreToRoot
+                ? _t('SilverStripe\\Admin\\ArchiveAdmin.RESTORE_TO_ROOT', 'Restore draft at top level')
```
maxime-rainville: Not sure about referencing Admin translation in framework.

dhensby:
There are references to `Versioned` in this PR which would make it an explicit dependency of framework going forward, that needs to be avoided.

We should make the restore action plug-able so that versioned can add it's own restore actions rather than coding them into framework. Framework should have no knowledge of versioned or any other modules

Issue #8044, opened 2018-05-01
[CONTROL] Logging before httpError will cause HTML to be added and output as plaintext


Sil3ntStorm:
## Affected Version

SilverStripe 4.1
composer info --direct
```
silverstripe-themes/simple dev-master 4d546a4 The SilverStripe simple theme (default SilverStripe 3 theme)
silverstripe/recipe-cms    1.1.0              SilverStripe recipe for fully featured page and asset content editing
silverstripe/recipe-plugin 1.2.0              Helper plugin to install SilverStripe recipes
```

## Description

Calling httpError with an error message will (magically and at an unknown location) prefix the message with HTML Dom and then send the entire thing to the browser, which causes conforming browsers to display the complete HTML instead of only the error message, as the Content-Type Header is rightfully [set to text/plain](https://github.com/silverstripe/silverstripe-framework/blob/4.1/src/Control/HTTPResponse_Exception.php#L47) during the Exception handling for security reasons.

## Steps to Reproduce

Install Base SS 4.1 `composer create-project --no-dev silverstripe/installer . ^4.1` and add the following code to a custom page action and load the appropriate page.
```php
    Injector::inst()->get(LoggerInterface::class)->error("This causes bad stuff to happen"); // Removing this line will remove the HTML and just leave the browser outputting "foo"
    $this->httpError(400, "foo"); // Will raise an HTTPResponse_Exception, setting the content-type header to text/plain
```
For example edit mysite/code/PageController.php and add the above to a new function crash(), add 'crash' to the allowedActions array and add the necessary use lines at the top.

Complete PageController.php:
```php
use SilverStripe\CMS\Controllers\ContentController;
use SilverStripe\Core\Injector\Injector;
use Psr\Log\LoggerInterface;

class PageController extends ContentController
{
    private static $allowed_actions = ['crash'];

    protected function init()
    {
        parent::init();
    }

    public function crash() {
        Injector::inst()->get(LoggerInterface::class)->error("This causes bad stuff to happen"); // Removing this line will remove the HTML and just leave the browser outputting "foo"
        $this->httpError(400, "foo"); // Will raise an HTTPResponse_Exception, setting the content-type header to text/plain
    }
}
```
#### Result
Instead of displaying a text "foo", you will see a bunch of HTML source code and at the very end the "foo" we were expecting.
Comment out the Injector::inst() line, reload the page and see that you're only getting the expected "foo"

[...]

PR #8241, opened 2018-07-04
Separate core error logging from standard LoggerInterface

robbieaverill:
**Up for discussion!**

At a high level:

* Rename the existing error handler implementation
* Register the default handler implementation as `LoggerInterface` for general purpose use
* Allow MonologErrorHandler to accept multiple loggers (includes deprecation of `setHandler` in favour of `pushHandler`)
* Update documentation, fix some typos in related docs/code

I've tested the following situations:

* `Injector::inst()->get(LoggerInterface::class)->debug('Hello')` levels are all logged to whatever handlers are defined for `LoggerInterface` (default: none)
  * Tested with file stream logger configured
  * Tested with "Messages" tab log collection in lekoala/silverstripe-debugbar
* Throwing an exception is handled by the SilverStripe error handler (CLI/browser)
* Throwing an exception is sent to the defined loggers for `LoggerInterface` (e.g. if using a file/stream logger)
* Logging PSR-3 errors via `LoggerInterface` doesn't get included in SilverStripe's error handlers
* Explicit log entries via `LoggerInterface.errorhandler` do get included in SilverStripe's error handler (previously the default behaviour)

Considerations:

* Do we want to provide a `NoopLoggerTestState` test state in framework that is configured by default to replace any custom log handlers that user code has defined, e.g. file logs? We could push a `NullHandler` by default

Resolves https://github.com/silverstripe/silverstripe-framework/issues/8044

### DIFF
--- a/_config/logging.yml
+++ b/_config/logging.yml
@@ -1,21 +1,34 @@
 ---
 Name: logging
 ---
-# Logging is built up of a chain containing:
+# Core error logging is built up of a chain containing:
 # - A top level \ErrorHandler which registers the error service
 # - A \Logger which acts as the error service
 # - A \HandlerInterface which handles errors for the logger
 # - One or more \FormatterInterface which format errors for the handler
+#
+# Logging for use outside of core error handling also uses the same implementations,
+# but is available without the HandlerInterfaces attached
 SilverStripe\Core\Injector\Injector:
   SilverStripe\Logging\ErrorHandler:
     class: SilverStripe\Logging\MonologErrorHandler
-    properties:
-      Logger: '%$Psr\Log\LoggerInterface'
+    calls:
+      pushDefaultLogger: [ pushHandler, [ '%$Psr\Log\LoggerInterface' ] ]
+      pushErrorHandler: [ pushHandler, [ '%$Psr\Log\LoggerInterface.errorhandler' ] ]
+
+  # Default implementation for use as a standard logger. Up to developers to attach their own
+  # handlers
   Psr\Log\LoggerInterface:
     type: singleton
     class: Monolog\Logger
     constructor:
       - "error-log"
+  # Core error handling
+  Psr\Log\LoggerInterface.errorhandler:
+    type: singleton
+    class: Monolog\Logger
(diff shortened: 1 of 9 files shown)


dhensby:
> Do we want to provide a NoopLoggerTestState test state in framework that is configured by default to replace any custom log handlers that user code has defined, e.g. file logs? We could push a NullHandler by default

quite possibly, yes - but maybe it would be good to actually log the test logs somewhere?

robbieaverill:
>quite possibly, yes - but maybe it would be good to actually log the test logs somewhere?
>I'd probably just log it to standard out...

The problem with that is that your phpunit runs fill up with stdout log calls from various parts of the codebase.

How about we configure a NullHandler for unit tests by default, which devs can remove via config if they want to. We might even be able to use the `--verbose` flag from phpunit to control this as well...

robbieaverill:
Rebased on top of 4 since there were some changes underneath. I've also added a `->setLoggers(array $loggers)` method in case someone wants to clear them

robbieaverill:
Thanks for having a look. Remaining would be to add something for unit test logging config/handlers. Not sure it'd block review and merge of this though - it could be done as a follow up

sminnee:
Just going to flag that right now logging is broken if you use Raygun. Gonna fix that with some bandaids but we should get this into 4.4!

robbieaverill:
Thanks @sminnee. To recap, I think the follow up action required is to decide whether to automatically configure a null handler for user configured logging when running unit tests (override user configured handlers) or to log test output somewhere by default, e.g. a /var/tmp/silverstripe-tests.log file or something like that.

I don't think stdout logging is a good idea TBH because it'll pollute phpunit test results.

Can we get a vote on it from @silverstripe/core-team?

🍎 : disable test logging by default
🍌 : enable test logging to a file by default

robbieaverill:
Two weeks for a vote on that, seems like the consensus is to remove logging from tests by default.

ScopeyNZ:
Looks like there's general good thoughts towards this PR in 4.4. I'll merge it now then :)

PR #8242, opened 2018-07-04
FIX: Remove X-Requested-With from default Vary header.

sminnee:
The X-Requested-With header does modify the result of Director::is_ajax
and so this should strictly be in there. In practise, this can cause
issues with CDNs such as Incapsula, and LeftAndMain adds this vary
header itself, which is the principle place where Director::is_ajax
is used.

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -16,7 +16,7 @@ HTTP:
     no-cache: "true"
     no-store: "true"
     must-revalidate: "true"
-  vary: "X-Requested-With, X-Forwarded-Protocol"
+  vary: "X-Forwarded-Protocol"
 LeftAndMain:
   dependencies:
     versionProvider: %$SilverStripeVersionProvider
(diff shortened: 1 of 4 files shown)


chillu:
+1 on this, but should be accompanied by a PHPDoc note in `is_ajax()` as well as a note in our "HTTP headers" markdown docs.

sminnee:
How's that @chillu ?

chillu:
Note that this won't be part of the 3.7.1 and 4.2.0 releases (those have been tagged weeks ago for a security audit), we'll need to do follow up patch releases to get this out.

chillu:
Forward ported this to 4.2: https://github.com/silverstripe/silverstripe-framework/pull/8280

I've changed my mind about including this in 4.2.0: I think it's crucial in order to explain caching behaviour and API changes consistently in both the release notes and updated caching docs (CWP and docs.ss.org). To this end, I have added a task for @dhensby on your internal release checklists in https://github.com/silverstripeltd/cc-issues/issues/163 to cherry pick this for both the 3.7.0 and 4.2.0 releases.

PR #8244, opened 2018-07-05
Make column query not distinct

altwohill:
When selecting a column, it doesn't make sense to have it `DISTINCT`. Selecting `$person->column('Age')` would be a prime example.

As an alternative, the API could be changed to give the end user the option
(eg `->column($field, $distinct = false)`)
However if we did that, we would need to make sure we do something
similar with `SilverStripe\ORM\UnsavedRelationList` to ensure consistency.

### DIFF
--- a/docs/en/04_Changelogs/4.3.0.md
+++ b/docs/en/04_Changelogs/4.3.0.md
@@ -1,0 +1,13 @@
+# 4.2.0
+
+## Overview {#overview}
+
+ - `DataList::column()` now returns all values and not just "distinct" values from a column as per the API docs
+ - `DataList`, `ArrayList` and `UnsavedRalationList` all have `columnUnique()` method for fetching distinct column values
+
+## Upgrading {#upgrading}
+
+### Fetching distinct column values on DataList
+
+Prior to this release `DataList` would erroneously return a distinct list of values from a column on an object.
+If this behaviour is still required, please use `columnUnique()` instead.
--- a/src/ORM/ArrayList.php
+++ b/src/ORM/ArrayList.php
@@ -371,6 +371,17 @@ class ArrayList extends ViewableData implements SS_List, Filterable, Sortable, L
     }
 
     /**
+     * Returns a unique array of a single field value for all the items in the list
+     *
+     * @param string $colName
+     * @return array
+     */
+    public function columnUnique($colName = 'ID')
+    {
+        return array_unique($this->column($colName));
+    }
+
+    /**
      * You can always sort a ArrayList
      *
      * @param string $by
--- a/src/ORM/DataList.php
+++ b/src/ORM/DataList.php
@@ -990,7 +990,18 @@ class DataList extends ViewableData implements SS_List, Filterable, Sortable, Li
      */
     public function column($colName = "ID")
     {
-        return $this->dataQuery->column($colName);
+        return $this->dataQuery->distinct(false)->column($colName);
+    }
+
+    /**
+     * Returns a unque array of a single field value for all items in the list.
+     *
+     * @param string $colName
+     * @return array
+     */
+    public function columnUnique($colName = "ID")
+    {
+        return $this->dataQuery->distinct(true)->column($colName);
     }
 
     // Member altering methods
--- a/src/ORM/DataQuery.php
+++ b/src/ORM/DataQuery.php
@@ -1147,7 +1147,6 @@ class DataQuery
     {
         $fieldExpression = $this->expressionForField($field);
         $query = $this->getFinalisedQuery(array($field));
-        $query->setDistinct(false);
         $originalSelect = $query->getSelect();
         $query->setSelect(array());
         $query->selectField($fieldExpression, $field);
--- a/src/ORM/DataQuery.php
+++ b/src/ORM/DataQuery.php
@@ -1147,6 +1147,7 @@ class DataQuery
     {
         $fieldExpression = $this->expressionForField($field);
         $query = $this->getFinalisedQuery(array($field));
+        $query->setDistinct(false);
         $originalSelect = $query->getSelect();
         $query->setSelect(array());
         $query->selectField($fieldExpression, $field);
--- a/src/ORM/ListDecorator.php
+++ b/src/ORM/ListDecorator.php
@@ -142,6 +142,11 @@ abstract class ListDecorator extends ViewableData implements SS_List, Sortable, 
         return $this->list->column($value);
     }
 
+    public function columnUnique($value = "ID")
+    {
+        return $this->list->columnUnique($value);
+    }
+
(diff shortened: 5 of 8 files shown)


altwohill:
I noticed that after PRing. The full test suite wouldn't run on my computer for some reason. From a look at the test failures it looks like the regressions are in the tests themselves, but that is enough reason to assume that people are relying on this behaviour elsewhere.

The tricky thing is that `SilverStripe\ORM\ArrayList::column($field)` currently does _not_ return a distinct set, so you get different behaviours if you call `$myThing->MyHasMany()->column('Name')` depending on whether the relationship has been saved or not...

maxime-rainville:
I agree that the behavior of this method as it stands is bit counter-intuitive, but altering it would break semantic versioning.

I'm not oppose to the the flag idea. I'm just wandering what alternative options we could have. Off the top of my head I could think of those.

* Add a new method that doesn't use distinct (Laravel's equivalent method is called `pluck`, that could be a good name for it)
* Do nothing and developers to explicitly say if they want the query to be distinct (`ObjectA::get()->distinct(false)->column('Name')`). 
* Create an alternative method with a sensible behavior, deprecate `column` and alias the new method with some tweak to make it unique.

dhensby:
Hey @maxime-rainville could you explain your reasoning as to why you think it'd break SemVer?

---

Seperately, if we think that this method is acting correctly then we have to conclude that `ArrayList::column()` is wrong and that needs to be fixed. If we think `ArrayList::column()` is right then we need to fix this. One way or another, one of the methods is not working as it should.

sminnee:
Maybe deprecate `column()` in favour of `columnValues()` and `columnUnqiueValues()`? The latter would preserve the current behaviour...

maxime-rainville:
It breaks semver, because it changes the (admittedly awkward) behavior of `column()`. If you call the method with the expectation that all the values will be unique and you end up getting none-unique values your system will stop working. To me that's a breaking change.

altwohill:
I think the cleanest way would probably be to fix `column()`, keeping the same naming, but tag it to the next version. It's not an urgent fix.

maxime-rainville:
I think it's arguable whatever this is a bug. Wanting to get all the unique values from a column is a very common use case. I have difficulty imagining a common scenario where you would want to get all the values from a single column without an ID or some other reference.

In that context, breaking the code of a lot people who expect this method to return an array of unique values is a big downside.

I would vote for @sminnee suggestion.
> Maybe deprecate column() in favour of columnValues() and columnUniqueValues()? The latter would preserve the current behaviour...

dhensby:
@maxime-rainville ok, so you think that the behaviour *should* be distinct values, right?

So that means `UnsavedRelationList` and `ArrayList` (and possibly other implementations of `SS_List`) are broken and need fixing. As @altwohill points out this behaviour is not consistent and so there's a bug *somewhere*. It appears to be distinct in `DataList` and not distinct in any other implementation. Now, to me that implies `DataList` is wrong, but I could accept the other implementations are wrong, either way we have to accept that somewhere the implementation is wrong and needs fixing.

It becomes even clearer if we read the docs for `SS_List::column()`:

>Returns an array of a single field value for all items in the list.

Not "an array of distinct values" but an array [of values] for **all items** in the list

dhensby:
I also don't really agree with @sminnee's suggestion because it's effectively admitting there's a "bug as a feature" and just doing away with it by creating two more methods, but those could easily also crystallise some uncaught "bug as a feature" and then we go on this whack-a-mole of deprecating and creating new methods just to avoid fixing bugs in the fear that bugs are now features

sminnee:
I'm okay with declaring this as a bug, but I would strongly favour my approach over parking this 'til v5.

If we change column() as a bugfix, I think we should still  introduce a columnUniqueValues() method as well? What we should we call it? Let's vote on the colour of the bikeshed!

 * 😄 = `columnUniqueValues()`
 * 🎉 = `uniqueColumn()`
 * ❤️  =  `columnUnique()`

dhensby:
>If we change column() as a bugfix, I think we should still introduce a columnUniqueValues() method as well? What we should we call it? Let's vote on the colour of the bikeshed!

is `$uniqueColumns = array_unique($list->column('ClassName'));` not good enough ;)

If we really want to introduce a new API for it I'd probably call it `columnUnique` but not bothered enough to change it if someone goes to the trouble of a PR and makes a unilateral choice

sminnee:
> is $uniqueColumns = array_unique($list->column('ClassName')); not good enough ;)

I think the extra method is justified as it pushes the filtering to the DB. `$issuesFromAllTime->columnUnique('Status')` might pick up 5 values from a list of 50,000. Doing this in the database query would be highly advisable.

>  I'd probably call it columnUnique but not bothered enough to change it if someone goes to the trouble of a PR and makes a unilateral choice

Click the heart button, man!

dhensby:
❤️s all round

dhensby:
@altwohill OK - the test is failing because `TeamGroup` extends `Team` and the fixture file creates a `Team` and a `TeamGroup` for each city meaning the column doubles up.

If you change the test like so it should pass

```diff
     public function testGetManipulatedData()
     {
-        $list = new DataList(Team::class);
+        $list = Team::get()->filter([ 'ClassName' => Team::class ]);
         $config = new GridFieldConfig_RecordEditor();
         $gridField = new GridField('testfield', 'testfield', $list, $config);
```

src/ORM/SS_List.php
```diff
@@ -83,6 +83,14 @@ public function find($key, $value);
      */
     public function column($colName = "ID");
 
+    /**
+     * Returns a unique array of a single field value for all items in the list.
+     *
+     * @param  string $colName
+     * @return array
+     */
+    public function columnUnique($colName = "ID");
```
dhensby: I'm afraid we can't change the interface in a minor release because this will break any other classes that implement it. We'd have to add this in 5.x

tractorcow:
I'd really like to be able to set distinct to false by default. It creates a performance overhead on nearly every query (mysql is slower) when we only need it sparingly. Maybe 5.x?

src/ORM/SS_List.php
tractorcow: It's a new method not a change isn't it?

Personally I like `distinctColumn` instead of unique to be consistent, but I'm easy really.

altwohill:
@tractorcow distinct is set to false if you call the default `column` function after this change. Or are you referring to default for all `queries`?

src/ORM/SS_List.php
kinglozzer: Adding it to the interface would cause a fatal error for any classes that implement the interface until they also implement the new method

dhensby:
Added docs too

dhensby:
thanks @altwohill for this

PR #8245, opened 2018-07-05
HTTPCacheControlMiddleware fixes

### LINKED ISSUE #8243: CacheControl privateCache should hold current state rather than replacing it
Comment from @dhensby on https://github.com/silverstripe/silverstripe-framework/pull/8233

> I've noticed during testing that if you have a session active, the default state changes to private, must-revalidate which doesn't seem right to me when it should probably hold the current state and simply add private and must-revalidate to the current state...

> I'd also query how privateCache() works or, at least, how we make requests private when there's a session. I think we should probably only add the "private" directive instead of remove all of the developer's preferences ...

dhensby:
Fixes #8243

I'd like to add a set of comprehensive tests for these changes

### DIFF
--- a/src/Control/Middleware/HTTPCacheControlMiddleware.php
+++ b/src/Control/Middleware/HTTPCacheControlMiddleware.php
@@ -513,7 +513,7 @@ class HTTPCacheControlMiddleware implements HTTPMiddleware, Resettable
     public function setMaxAge($age)
     {
         // Affect all non-disabled states
-        $applyTo = [self::STATE_ENABLED, self::STATE_PRIVATE, self::STATE_PUBLIC];
+        $applyTo = [self::STATE_ENABLED, self::STATE_PRIVATE, self::STATE_PUBLIC, self::STATE_DEFAULT];
         $this->setStateDirective($applyTo, 'max-age', $age);
         return $this;
     }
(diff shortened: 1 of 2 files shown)


sminnee:
Yeah I'd agree with adding tests :-) overall approach to changing the code seems fine, but I'd say that adding tests blocks merge.

dhensby:
Started on tests, still not ready to merge; I'm going to find a nicer way of making sure that the disabled state is immutable

dhensby:
An important thing to note here is that `no-cache` is actually a kind of enabled state; ETags are generated and 304 responses are possible. So I've removed my "default" state which was just a soft enabled state and turned the "enabled" state to be what was the default.

This means by default we will allow 304 responses, but it's for a developer to set a max-age. The framework still protects them from caching during sessions and so on.

dhensby:
OK - I've finished adding tests and I think this is about there now

dhensby:
One thing I have thought whilst doing this is whether you should be able to directly call `setStateDirective` to set or unset `private` on the private state and `public` on the public state... I feel like those directives should not be editable, but that's a pretty specific action to be taking which I think would have to be deliberate by a developer, if they've chosen to do it, then on their head be it?

src/Control/Middleware/HTTPCacheControlMiddleware.php
```diff
@@ -103,7 +99,7 @@ public function process(HTTPRequest $request, callable $delegate)
      * @config
      * @var string
      */
-    private static $defaultState = self::STATE_DEFAULT;
+    private static $defaultState = self::STATE_ENABLED;
```
tractorcow: Sure, I never really liked "default" anyway. I only kept it in to maintain the old behaviour.

dhensby:
bump

dhensby:
<3

PR #8261, opened 2018-07-12
NEW Option for secure "remember me" cookie

### LINKED ISSUE #8234: alc_enc cookie does not include secure cookie flag
Reported by @madmatt:

SilverStripe does not set the secure flag on the alc_enc cookie, which means that even if user logins all happen over a secure connection (e.g. Director::forceSSL() on any secure content), any user who selects the 'Remember me' checkbox could leak their token unknowningly, and have someone take over their account.

I think we do this because some sites don't use HTTPS at all, so we can't just set it blindly, but we could potentially handle this in the same way that we already handle it for session cookies (we use PHPSESSID for insecure sessions, and SECSESSID - with the secure flag set - when the session is done over https).

Alternatively, we could remove the ability to 'Remember me' when login is done over http, and then we could reliable set the secure flag on the alc_enc cookie all the time, though that's quite a heavy-handed solution.

Most sites should be moving to HTTPS-everywhere-all-the-time, but this is still good practice to separate cookie usage, and ensure that we set the secure flag where-ever possible.

chillu:
Fixes #8234. I've tested that the cookie gets set locally
on a Vagrant box with SSL enabled (by inspecting headers),
but haven't gotten the browser to actually use it because
for some reason, this vagrant box sets a
`<base href="http://...">` (not HTTPS).

I really wanted to write a test for this,
but PHP doesn't let you inspect cookie headers on CLI.
Here's my attempt, any ideas welcome.

```php
<?php
namespace SilverStripe\Security\Tests\MemberAuthenticator;

use SilverStripe\Security\Member;
use SilverStripe\Security\MemberAuthenticator\CookieAuthenticationHandler;

class CookieAuthenticationHandlerTest extends SapphireTest
{

    protected $usesDatabase = true;

    public function testLoginOnlySetsTokenWithPersistent()
    {
        $member = new Member(['Email' => 'test@example.com']);
        $member->write();

        $auth = new CookieAuthenticationHandler();
        $auth->setTokenCookieName('testTokenCookie');

        $headers = $this->runWithHeaderIsolation(function () use ($auth, $member) {
            $auth->logIn($member, $persistent = true);
        });
        $cookieHeader = $this->getCookieHeader($headers);
        $this->assertNotContains("Secure", $cookieHeader);

        $auth->setTokenCookieSecure(true);
        $headers = $this->runWithHeaderIsolation(function () use ($auth, $member) {
            $auth->logIn($member, $persistent = true);
        });
        $cookieHeader = $this->getCookieHeader($headers);
        $this->assertContains("Secure", $cookieHeader);
    }

    protected function runWithHeaderIsolation($callback)
    {
        ob_start();
        $callback();
        // TODO Doesn't work on CLI
        $headers = headers_list();
        // TODO Doesn't actually clear headers
        ob_end_clean();

        return $headers;
    }

    protected function getCookieHeader($headers)
    {
        foreach($headers as $header) {
            if(preg_match('/^Set-Cookie:', $header)) {
                return $header;
            }
        }

        return null;
    }
}
```


src/Security/MemberAuthenticator/CookieAuthenticationHandler.php
```diff
@@ -189,13 +216,14 @@ public function logIn(Member $member, $persistent = false, HTTPRequest $request
             $rememberLoginHash = RememberLoginHash::generate($member);
             $tokenExpiryDays = RememberLoginHash::config()->uninherited('token_expiry_days');
             $deviceExpiryDays = RememberLoginHash::config()->uninherited('device_expiry_days');
+            $secure = $this->getTokenCookieSecure();
```
dhensby: Should this use the current request as well? like we do for the session cookie. "if the request is HTTPS && we use the secure cookie setting"?

chillu: At the moment this assumes that when you use this setting, your site unconditionally redirects to HTTPS for all routes. We could make it more friendly towards mixed mode (login form submitting to HTTP rather than HTTPS), but there's all kinds of other edge cases that we've discussed in the original ticket. I'd prefer if we document this as "only turn on option if your whole site is on HTTPS".

dhensby: OK, that makes sense

docs/en/02_Developer_Guides/09_Security/04_Secure_Coding.md
```diff
@@ -705,6 +705,15 @@ SilverStripe\Control\Session:
   cookie_secure: true
 ```
 
+The same treatment should be applied to the cookie responsible for remembering logins across sessions:
+
+```yml
```
dhensby: I think we should document how to do this only in Live mode or not on dev as I think that'll be quite a common usecase

PR #8263, opened 2018-07-14
FIX updateValidatePassword calls need to be masked from backtraces

dhensby:
#7800 should really have made sure that any arguments passed to the `updateValidatePassword` were masked.

Because we can't know what the class name will be, I've introduce the ability to use wildcards for class names when declaring what methods should be filtered.

Future enhancements would be to inspect if the class implements an interface or shares a common class ancestor when filtering


dhensby:
@chillu ideally this would make the 4.1 release as well

PR #8264, opened 2018-07-16
Add use statement to example

DorsetDigital:
Add the relevant 'use' statement to the sample code to remind people that it's needed!


dhensby:
thanks :)

PR #8265, opened 2018-07-16
Added install-requirement-check for php-intl

gbaumeister:
This PR adds a check that the php-intl extension is available, and prohibits the installation if missing. 

See: https://github.com/silverstripe/silverstripe-installer/issues/205 for details.


dhensby:
thanks

PR #8266, opened 2018-07-16
Update $odd logic, instanceof instead of is_a and remove SQLMap checks

robbieaverill:
SQLMap was removed early in 3.x, so no need to keep the checks around for it.
This commit also improves the logic in CheckboxSetField very slightly.

Some minor updates to the code.


dhensby:
I'm happy - just wondering if this is worthy of going into a sec-only release line?

robbieaverill:
It could go anywhere really, I can re-target at 4.x if you'd rather?

dhensby:
yes please

PR #8268, opened 2018-07-17
FIX sizeof doesnt work with null types

dhensby:
Getting a warning when using the installer

I've attempted to keep the return type / values the same even though it appears that the result is only used as a boolean


dhensby:
The warning is new in PHP 7.2 (http://php.net/manual/en/function.count.php#refsect1-function.count-changelog)

PR #8269, opened 2018-07-18
BUG Lazy session state (fixes #8267)

### LINKED ISSUE #8267: Unconditional call to Session::start() via SessionMiddleware
4.x introduced `SessionMiddleware`, which calls `$request->getSession()->init($request);`, which in turn calls `Session->start()`. This is done on every request, regardless if there's any session data to be written. `Session->start()` now calls `Cookie::set(session_name(), ...)`, which *always* sends a `Set-Cookie: PHPSESSID` header. This destroys HTTP cacheability of requests.

Tried this on 4.2.x-dev, but also checked out 4.0.3 with the same outcome. Using `curl -I` to ensure I'm not sending `PHPSESSID` cookies with the request to begin with, and have no further modifications or modules installed.

Am I going insane here? The `SessionMiddleware` behaviour has been the same for over a year (early beta days), I can only see a [Session.php data handling change commit](https://github.com/silverstripe/silverstripe-framework/commit/740c3326e9da7cb902778f46e209319e6dfd67a7#diff-d7f75870ba135b69dd1798e2f09f67ae) from @tractorcow back in Feb, but otherwise there's nothing notable in the [Session.php history](https://github.com/silverstripe/silverstripe-framework/commits/4/src/Control/Session.php).

As far as I'm concerned, this is blocking the 4.2.0 release, which has a big focus on HTTP cacheability.

Pull Requests:

 * [x] https://github.com/silverstripe/silverstripe-framework/pull/8269

chillu:
Context: https://github.com/silverstripe/silverstripe-framework/issues/8267

Fixes regression from 3.x, where sessions where lazy started as required:
Either because an existing session identifier was sent through with the request,
or because new session data needed to be persisted as part of the request execution.

Without this lazy starting, *every* request will get a session,
which makes all those responses uncacheable by HTTP layers.

Note that 4.x also changed the $data vs. $changedData payloads:
In 3.x, they both contained key/value pairs.
In 4.x, $data contains key/value, while $changedData contains key/boolean to declare isChanged.
While this reduces duplication in the class, it also surfaced a bug which was latent in 3.x:
When an existing session is lazily resumed via start(), $data is set back to an empty array.
In 3.x, any changed data before this point was *also* retained in $changedData,
ensuring it gets merged into existing $_SESSION data.
In 4.x, this clears out data - hence the need for a more complex merge logic.

Since isset($this->data) is no longer an accurate indicator of a started session,
we introduce a separate $this->started flag.

Note that I've chosen not to make lazy an opt-in (e.g. via start($request, $lazy=false)).
We already have a distinction between lazy starting via init(), and force starting via start().

I've tried to add unit tests for this, but it's nearly impossible to do because
of PHP's headers_sent() handling and PHPUnit, see https://github.com/silverstripe/silverstripe-framework/issues/8267#issuecomment-405766370

### DIFF
--- a/src/Control/Director.php
+++ b/src/Control/Director.php
@@ -206,7 +206,7 @@ class Director implements TemplateGlobalProvider
         if ($session instanceof Session) {
             // Note: If passing $session as object, ensure that changes are written back
             // This is important for classes such as FunctionalTest which emulate cross-request persistence
-            $newVars['_SESSION'] = $sessionArray = $session->getAll();
+            $newVars['_SESSION'] = $sessionArray = $session->getAll() ?: [];
             $finally[] = function () use ($session, $sessionArray) {
                 if (isset($_SESSION)) {
                     // Set new / updated keys
--- a/src/Control/Session.php
+++ b/src/Control/Session.php
@@ -262,8 +262,8 @@ class Session
     public function requestContainsSessionId(HTTPRequest $request)
     {
         $secure = Director::is_https($request) && $this->config()->get('cookie_secure');
-		$name = $secure ? $this->config()->get('cookie_name_secure') : session_name();
-		return (bool)Cookie::get($name);
+        $name = $secure ? $this->config()->get('cookie_name_secure') : session_name();
+        return (bool)Cookie::get($name);
     }
 
     /**
@@ -330,9 +330,9 @@ class Session
                 $this->recursivelyApply((array)$this->data, $data);
             } else {
                 // Use in-memory data if the session is lazy started
-                $data = isset($this->data) ? $this->data : [];
+                $data = $this->data;
             }
-            $this->data = $data;
+            $this->data = $data ?: [];
         } else {
             $this->data = [];
         }
@@ -615,6 +615,7 @@ class Session
      */
     protected function recursivelyApplyChanges($changes, $source, &$destination)
     {
+        $source = $source ?: [];
         foreach ($changes as $key => $changed) {
             if ($changed === true) {
                 // Determine if replacement or removal
--- a/src/Control/Session.php
+++ b/src/Control/Session.php
@@ -49,6 +49,9 @@ use SilverStripe\Dev\Deprecation;
  *
  * Once you have saved a value to the Session you can access it by using the {@link Session::get()} function.
  * Like the {@link Session::set()} function you can use this anywhere in your PHP files.
+ * Note that session data isn't persisted in PHP's own session store (via $_SESSION)
+ * until {@link Session::save()} is called, which happens automatically at the end of a standard request
+ * through {@link SilverStripe\Control\Middleware\SessionMiddleware}.
  *
  * The values in the comments are the values stored from the previous example.
  *
@@ -84,7 +87,6 @@ use SilverStripe\Dev\Deprecation;
  * </code>
  *
  * @see Cookie
- * @todo This class is currently really basic and could do with a more well-thought-out implementation.
  */
 class Session
 {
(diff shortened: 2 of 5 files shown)


robbieaverill:
I think the tests should be fixed with a patch to ensure that `$this->data` is always created as an array, as it was previously with this line:

```php
$this->data = isset($_SESSION) ? $_SESSION : array();
```

Patch:

```patch
diff --git a/src/Control/Session.php b/src/Control/Session.php
index 17a2c295a..92e48a040 100644
--- a/src/Control/Session.php
+++ b/src/Control/Session.php
@@ -325,7 +325,7 @@ class Session
 
             if (isset($_SESSION)) {
                 // Initialise data from session store if present
-                $data = $_SESSION;
+                $data = $_SESSION ?: [];
                 // Merge in existing in-memory data, taking priority over session store data
                 $this->recursivelyApply((array)$this->data, $data);
             } else {
@@ -615,6 +615,7 @@ class Session
      */
     protected function recursivelyApplyChanges($changes, $source, &$destination)
     {
+        $source = $source ?: [];
         foreach ($changes as $key => $changed) {
             if ($changed === true) {
                 // Determine if replacement or removal


```

chillu:
Thanks Robbie, I've incorporated your fixes. Also figured out how to run isolated tests in PHPUnit, so added some coverage (see [PHPUnit functions](https://phpunit.de/manual/5.7/en/appendixes.annotations.html) which are available in our minimum required version, 5.7).

I tried to test the tricky bit with merging existing `$_SESSION` data, but PHP's amazing architecture gets in the way (too many globals), see `testStartRetainsInMemoryData()`.

Note that the tests run a bit slower with process isolation (`SessionAuthenticationHandlerTest.php.php` takes 6s rather than 3s on my machine).

tests/php/Security/MemberAuthenticator/CookieAuthenticationHandlerTest.php
```diff
@@ -0,0 +1,57 @@
...
+        $auth = new CookieAuthenticationHandler();
+        $auth->setTokenCookieName('testTokenCookie');
+
+        $headers = $this->runWithHeaderIsolation(function () use ($auth, $member) {
+            $auth->logIn($member, $persistent = true);
+        });
+        $cookieHeader = $this->getCookieHeader($headers);
+        $this->assertNotContains("Secure", $cookieHeader);
+
+        $auth->setTokenCookieSecure(true);
```
robbieaverill: This method doesn't seem to exist anywhere

tests/php/Security/MemberAuthenticator/CookieAuthenticationHandlerTest.php
```diff
@@ -0,0 +1,57 @@
...
+        $member = new Member(['Email' => 'test@example.com']);
+        $member->write();
+
+        $auth = new CookieAuthenticationHandler();
+        $auth->setTokenCookieName('testTokenCookie');
+
+        $headers = $this->runWithHeaderIsolation(function () use ($auth, $member) {
+            $auth->logIn($member, $persistent = true);
+        });
+        $cookieHeader = $this->getCookieHeader($headers);
```
robbieaverill: This method can return null, which won't work in the assertion below. If you change it to return an empty string instead of null then the assertion passes.

tests/php/Security/MemberAuthenticator/CookieAuthenticationHandlerTest.php
chillu: PR crossover from https://github.com/silverstripe/silverstripe-framework/pull/8261, I'll remove the test here - it's the same area, but not really related to the PR at hand, my mistake.

tests/php/Security/MemberAuthenticator/CookieAuthenticationHandlerTest.php
chillu: Same here, not applicable to this PR

chillu:
OK all green now. Merging since Dan approved.

PR #8271, opened 2018-07-18
FIX Ensure tests pass by reducing to PHPUnit 7.1.3

robbieaverill:
PHPUnit 7.1.4 and higher introduces a bug with float comparisons in strings. The fix until this is resolved is to use PHPUnit 7.1.3.

Reference: https://github.com/sebastianbergmann/phpunit/issues/3185

Also updates the phpunit config file to match [the schema](https://phpunit.readthedocs.io/en/7.1/configuration.html) and imports a missing exception namespace.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,7 +13,7 @@ trim_trailing_whitespace = true
 [*.md]
 trim_trailing_whitespace = false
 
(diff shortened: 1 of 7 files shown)


dhensby:
Gosh, that's a painful regression that you'd think they'd fix a bit more quickly. :(

PR #8273, opened 2018-07-18
Update jquery-ui.min.js to fix minified syntax bug

AlphaCactus:
In two places in this file there is the syntax `+ ++` which throws an interpreter error and causes all javascripts in the same file (such as when combining this file with other files) to fail to load. This problem can easily be solved by wrapping the ++n and ++t with parentheses. which is what I've done in this pull request.

This bug is discussed at:
https://bugs.jqueryui.com/ticket/5576
and 
http://www.crockford.com/javascript/jsmin.html

I'm not sure why nobody else has ever encountered this issue. Till now I've been manually editing this one file after any framework updates.


dhensby:
> I'm not sure why nobody else has ever encountered this issue.

Indeed - I can only imagine the issue isn't actually that prevalent and because of that I'm not sure this will get merged. 3.x is security fixes only now so this is unlikely to make it in.

Are you using jquery-ui for something custom or is it used in the CMS out of the box? What browser do you experience this error in? Does it happen in a plain install with no other changes?

I'd expect the workaround we'd recommend is include jquery-ui in your own code and then use `requirements` to override the file that's imported

AlphaCactus:
Yep all of what you said makes perfect sense.
Using jQuery-ui on the front end so don't really need to be using the framework version and will do what you recommend.
Been occurring for a long time so been through lots of browsers, but error is currently experienced in Firefox 61.0.1 (64-bit) and Chrome Version 67.0.3396.99 (Official Build) (64-bit).

dhensby:
@AlphaCactus thanks for your understanding

PR #8274, opened 2018-07-19
Corrected caching docs and deprecation behaviour (fixes #8272)

### LINKED ISSUE #8272: enableCache() isn't actually enabling caching
Follow up on the recent [cache control changes](https://github.com/silverstripe/silverstripe-framework/pull/8166) slated for 4.2.0

# Overview

I believe there's a regression in there, or at least a misinterpretation of the spirit of the whole caching API. In short, the `enableCache()` meaning got changed halfway through the discussion, in a way which prevents both browsers and CDNs from *actually caching*. The headers instruct them to call back to the origin server with their `If-Modified-*` headers, and the server can choose to return a 304. That's not saving any server resources, since we still need to process the entire request execution. It just saves bandwidth. It's also [inconsistent with our 3.x implementation](https://github.com/silverstripe/silverstripe-framework/blob/3.7/control/HTTPCacheControl.php#L306).

**I think it's critical that we remove `no-cache` from `enableCache()` before we call 4.2.0 stable**

# Context

@dhensby submitted a [PR](https://github.com/silverstripe/silverstripe-framework/pull/8245) with a commit titled [State default should be state enabled (no-cache is an enabled state)](https://github.com/silverstripe/silverstripe-framework/commit/fd8448889c5170ad022c069856855b11fb6f57b6), which @tractorcow  merged into 4.2.

[HttpCacheControlMiddleware::stateDirectives](https://github.com/silverstripe/silverstripe-framework/blob/b83281fa5e1d063037cc8f6c8aa37e4d2daea7bf/src/Control/Middleware/HTTPCacheControlMiddleware.php#L76:L93)

```php
    protected $stateDirectives = [
        self::STATE_DISABLED => [
            'no-cache' => true,
            'no-store' => true,
            'must-revalidate' => true,
        ],
        self::STATE_PRIVATE => [
            'private' => true,
            'must-revalidate' => true,
        ],
        self::STATE_PUBLIC => [
            'public' => true,
            'must-revalidate' => true,
        ],
        self::STATE_ENABLED => [
            'no-cache' => true,
            'must-revalidate' => true,
        ]
    ];
```

In the PR, he [comments](https://github.com/silverstripe/silverstripe-framework/pull/8245#issuecomment-403000216):

> An important thing to note here is that no-cache is actually a kind of enabled state; ETags are generated and 304 responses are possible. So I've removed my "default" state which was just a soft enabled state and turned the "enabled" state to be what was the default. This means by default we will allow 304 responses, but it's for a developer to set a max-age. The framework still protects them from caching during sessions and so on.

Our [docs](https://github.com/silverstripe/silverstripe-framework/blob/4/docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md) say on `enableCache()`:

> Simple way to set cache control header to a cacheable state. Use this method over publicCache() if you are unsure about caching details. **Removes no-store and no-cache directives**; other directives will remain in place. Use alongside setMaxAge() to indicate caching.

Further down we use `enableCache()` in a code example described as:

> Global opt-in for page content
> Enable caching for all page content (through PageController).

# Notes

 * There was a discussion about [making the default state no-cache](https://github.com/silverstripe/silverstripe-framework/pull/8195#issuecomment-398180508). That's still OK, since the default state is disabled.

# Pull requests

 * [ ] https://github.com/silverstripe/silverstripe-framework/pull/8274

chillu:
See https://github.com/silverstripe/silverstripe-framework/issues/8272 for context

The HTTP.cache_control behaviour was removed in https://github.com/silverstripe/silverstripe-framework/pull/8166/files#diff-523aa2e152e10c465a85790ac13e30e4L402,
but not properly deprecated - it just silently stopped working.

This was a different approach to `HTTP.vary` deprecation, which I think why it's an omission (@dhensby couldn't remember).

### DIFF
--- a/docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
+++ b/docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
@@ -8,7 +8,7 @@ summary: Set the correct HTTP cache headers for your responses.
 By default, SilverStripe sends headers which signal to HTTP caches
 that the response should be not considered cacheable.
 HTTP caches can either be intermediary caches (e.g. CDNs and proxies), or clients (e.g. browsers).
-The cache headers sent are `Cache-Control: no-store, no-cache, must-revalidate`;
+The cache headers sent are `Cache-Control: no-cache, must-revalidate`;
 
 HTTP caching can be a great way to speed up your website, but needs to be properly applied.
 Getting it wrong can accidentally expose draft pages or other protected content.
@@ -59,8 +59,8 @@ Does not set `private` directive, use `privateCache()` if this is explicitly req
 Simple way to set cache control header to a cacheable state.
 Use this method over `publicCache()` if you are unsure about caching details.
 
-Removes `no-store` and `no-cache` directives; other directives will remain in place.
-Use alongside `setMaxAge()` to indicate caching.
+Removes the `no-store` directive unless a `max-age` is set; other directives will remain in place.
+Use alongside `setMaxAge()` to activate caching.
 
 Does not set `public` directive. Usually, `setMaxAge()` is sufficient. Use `publicCache()` if this is explicitly required
 ([details](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#public_vs_private))
@@ -184,24 +184,13 @@ class PageController extends ContentController
 }
 ```
 
-## Defaults
-
-By default, PHP adds caching headers that make the page appear purely dynamic. This isn't usually appropriate for most 
-sites, even ones that are updated reasonably frequently. SilverStripe overrides the default settings with the following 
-headers:
-
-  * The `Last-Modified` date is set to be most recent modification date of any database record queried in the generation 
-  of the page.
-  * The `Expiry` date is set by taking the age of the page and adding that to the current time.
-  * `Cache-Control` is set to `max-age=86400, must-revalidate`
-  * Since a visitor cookie is set, the site won't be cached by proxies.
-  * Ajax requests are never cached.
-
 ## Max Age
 
 The cache age determines the lifetime of your cache, in seconds.
 It only takes effect if you instruct the cache control
-that your response is cacheable in the first place (via `enableCache()` or via modifying the `HTTP.cache_control` defaults).
+that your response is cacheable in the first place
+(via `enableCache()`, `publicCache()` or `privateCache()`), 
+or via modifying the `HTTP.cache_control` defaults).
 
 ```php
 use SilverStripe\Control\Middleware\HTTPCacheControlMiddleware;
@@ -209,7 +198,8 @@ HTTPCacheControlMiddleware::singleton()
     ->setMaxAge(60)
 ```
 
-Note that `setMaxAge(0)` is NOT sufficient to disable caching in all cases.
+Note that `setMaxAge(0)` is NOT sufficient to disable caching in all cases,
+use `disableCache()` instead.
 
 ### Last Modified
 
--- a/src/Control/HTTP.php
+++ b/src/Control/HTTP.php
@@ -71,6 +71,7 @@ class HTTP
     /**
      * List of names to add to the Cache-Control header.
      *
+     * @deprecated 4.2..5.0 Handled by HTTPCacheControlMiddleware instead
      * @see HTTPCacheControlMiddleware::__construct()
      * @config
      * @var array Keys are cache control names, values are boolean flags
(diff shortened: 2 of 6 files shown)


docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
```diff
@@ -184,32 +184,22 @@ class PageController extends ContentController
 }
 ```
 
-## Defaults
```
chillu: I can't believe we still had this in there, it's completely out of date :(

src/Control/HTTP.php
```diff
@@ -509,6 +510,37 @@ public static function augmentState(HTTPRequest $request, HTTPResponse $response
...
         }
 
+        // Pass cache_control to middleware
+        $configCacheControl = $config->get('cache_control');
+        if ($configCacheControl) {
+            Deprecation::notice('5.0', 'Use HTTPCacheControlMiddleware API instead');
+
+            $supportedDirectives = ['max-age', 'no-cache', 'no-store', 'must-revalidate'];
+            if ($foundUnsupported = array_diff(array_keys($configCacheControl), $supportedDirectives)) {
+                throw new \LogicException(
```
robbieaverill: Could you add this to the PHPDoc please?

tests/php/Control/HTTPTest.php
```diff
@@ -112,6 +113,79 @@ public function testConfigVary()
         $this->assertEmpty($v);
     }
 
+    public function testDeprecatedVaryHandling()
+    {
+        /** @var Config */
+        Config::inst()->update(
```
robbieaverill:
```diff
- Config::inst()->update(
+ Config::modify()->set(
```

src/Control/HTTP.php
chillu: Done

tests/php/Control/HTTPTest.php
chillu: Oh is that what the cool kids are using these day? :D Updated.

tests/php/Control/Middleware/HTTPCacheControlMiddlewareTest.php
```diff
@@ -68,6 +68,65 @@ public function testSetMaxAge($state, $immutable)
         }
     }
 
+    public function testEnableCacheWithMaxAge()
```
dhensby: I think these tests aren't needed and we should just update the provider and the existing tests...

docs/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
```diff
@@ -59,8 +59,8 @@ Does not set `private` directive, use `privateCache()` if this is explicitly req
 Simple way to set cache control header to a cacheable state.
 Use this method over `publicCache()` if you are unsure about caching details.
 
-Removes `no-store` and `no-cache` directives; other directives will remain in place.
-Use alongside `setMaxAge()` to indicate caching.
+Removes the `no-store` directive; other directives will remain in place.
```
dhensby: This still isn't quire true; it doesn't remove the `no-store` directive unless a max-age is supplied.

chillu: Amended

tests/php/Control/Middleware/HTTPCacheControlMiddlewareTest.php
chillu: For this particular test (`testEnableCacheWithMaxAge`), I agree with you - it's already covered by `testSetMaxAge`.

But otherwise, I find this particular use of PHPUnit providers quite unreadable actually - it took me a while to figure out what the meaning of `$immutable` is in this context. I think providers are fine when they contain one dimension, since that's something you can keep in your head. But otherwise, DRY principles apply differently to unit tests: In my opinion, clarity (and sometimes verbosity) is more important than brevity. I can understand what `testEnableCacheWithMaxAgeAppliesWhenLevelDoesNot` tests. I couldn't understand this if it was mixed in as a third array element in `provideCacheStates`, even when using [named data provider sets](https://phpunit.de/manual/6.5/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers)

chillu: Update: The difference with `testEnableCacheWithMaxAge` (compared to `testSetMaxAge`) is the max-age setting via argument rather than setter. Not all methods in your data provider have this argument, because it doesn't make sense for all of them - so you'd either have to use a different data provider, or add more confusing dimensions to the existing one. And since the assertions change between `enableCache` and `publicCache` usage, I don't think a new data provider capturing both cases will be more readable.

PR #8275, opened 2018-07-19
FIX Make sure that we only send arrays on construct of HTTPCacheControl

dhensby:
If you unset the `HTTP.cache_control` config on your site, you get an error because we don't check it's actually got a value...

### DIFF
--- a/control/HTTPCacheControl.php
+++ b/control/HTTPCacheControl.php
@@ -81,7 +81,10 @@ class HTTPCacheControl extends SS_Object {
 
 		// If we've not been provided an initial state, then grab HTTP.cache_contrpl from config
 		if (!$this->state) {
-			$this->setDirectivesFromArray(Config::inst()->get('HTTP', 'cache_control'));
+			$defaultDirectives = Config::inst()->get('HTTP', 'cache_control');
+			if (!empty($defaultDirectives)) {
+				$this->setDirectivesFromArray($defaultDirectives);
+			}
 		}
 	}
 


control/HTTPCacheControl.php
```diff
@@ -81,7 +81,10 @@ public function __construct()
 
 		// If we've not been provided an initial state, then grab HTTP.cache_contrpl from config
 		if (!$this->state) {
-			$this->setDirectivesFromArray(Config::inst()->get('HTTP', 'cache_control'));
+			$defaultDirectives = Config::inst()->get('HTTP', 'cache_control');
+			if (!empty($defaultDirectives)) {
+				$this->setDirectivesFromArray(Config::inst()->get('HTTP', 'cache_control'));
```
robbieaverill: Could we re-use `$defaultDirectives` here? Or even cast the config value as an array and skip the if statement?

dhensby:
> Personally I have no qualms with merging non-security PRs into 3.x

It's about our policy (which is sec only) and the expectation we give the community around maintenance and also the potential hypocrisy this may expose ;)

@silverstripe/core-team thoughts, please

chillu:
I think the community trusts us to make judgement calls within the bounds of broad statements like "3.x is in security-only maintenance mode". The important bit here is that no new features get implemented in this release line. I think in terms of bug fixes, I think the comms should be "critical only, which mostly means security fixes". 

We might need a better definition of what the bug labels refer to, e.g. what criteria we apply when something is `impact/critical` vs. `impact/high` etc - that might help with community comms as well.

In this particular case, I'd say it's not a "high impact" bug, because it likely won't affect a lot of people - we recommended using controllerpolicy over `HTTP.cache_control` for a while now, and the new APIs presumably mean you don't get any errors either?

dhensby:
OK - I'll close. People who want to remove the default config need to replace it with an empty array.

robbieaverill:
To be fair, our support timeline on the website says we're in security only mode for 3.x so ignore me =)

PR #8281, opened 2018-07-24
Allow version 4 of nikic/php-parser

michalkleiner:
Allow version 4 of the php-parser library.

According to the GitHub repo, version 3 is no longer actively supported and other dependencies or tools like PHPStan may require version 4, creating an uninstallable set of conflicting dependencies.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "league/csv": "^8",
         "league/flysystem": "~1.0.12",
         "monolog/monolog": "~1.11",
-        "nikic/php-parser": "^2 || ^3",
+        "nikic/php-parser": "^2 || ^3 || ^4",
         "paragonie/random_compat": "^2.0",
         "silverstripe/config": "^1@dev",
         "silverstripe/assets": "^1@dev",


maxime-rainville:
I think we should double check version 2, 3, and 4 all actually work. We don't have logic right now in travis to test against all the versions of php-parser we ostensibly support.

michalkleiner:
Also, which version actually gets installed when it specifies ^2 or ^3?

dhensby:
> Also, which version actually gets installed when it specifies ^2 or ^3?

the highest that is compatible with your deps.

dhensby:
The work that php-parser was added for was written on 2.x, so that will "definitely" work (building class manifest info). I tested it with 3.x when it was released. Luckily out implementation of PHPParser for parsing classes is very simple as all we do is pull out the very high level info (namespace, classname, parent class, interfaces implemented) and so we don't really inspect many of the properties of the class

It looks like (as the tests are green) that 4.x works too. So I'm happy to take it in based on the green ticks.

As php-parser 3.x is no longer supported we're going to have to upgrade to 4.x anyway as new PHP 7.3 syntax is added and supported.

PR #8283, opened 2018-07-24
Allow removeDuplicates to be chainable

mspacemedia:
Added `return $this` to allow chaining of removeDuplicates. Same PR as one I added to SS4


robbieaverill:
Hi there, thank you for contributing! SilverStripe 3 is in security fix only state now, so we are no longer accepting non critical patches. If this chainability isn’t in 4.x we’d welcome a patch there though! Cheers

robbieaverill:
Just re read - thanks for contributing a SS4 patch 👌

PR #8285, opened 2018-07-25
Adds extension hooks to ManyManyList for add and remove operations.

ChrissiQ:
As a developer, I want to notify other parts of my application when relationships on objects are updated.  Without this hook, I can't, for example, update a field in the database when many-many relationships are modified.  This is especially useful for partial caching, so that I can update LastEdited on an object when objects are added or removed from its relation list.

### DIFF
--- a/src/ORM/ManyManyList.php
+++ b/src/ORM/ManyManyList.php
@@ -299,6 +299,8 @@ class ManyManyList extends RelationList
             $manipulation[$this->joinTable]['fields'][$this->localKey] = $itemID;
             $manipulation[$this->joinTable]['fields'][$this->foreignKey] = $foreignID;
 
+            $this->extend('onBeforeAdd', $manipulation, $itemID);
+
             DB::manipulate($manipulation);
         }
     }
@@ -345,6 +347,9 @@ class ManyManyList extends RelationList
         $query->addWhere(array(
             "\"{$this->joinTable}\".\"{$this->localKey}\"" => $itemID
         ));
+
+        $this->extend('onBeforeRemove', $query, $itemID);
+
         $query->execute();
     }
 


kinglozzer:
I think extension hooks here need to be carefully considered, it’s a unique use case. Thoughts that spring to my mind are:

- Should the hook be called on `ManyManyList`, or on the `DataObject` being added/removed? This will ultimately decide whether users implementing it will apply an extension to a specific object (how they do with `onBeforeWrite()`/`onAfterWrite()`) or use the same extension/hook for multiple object types.
- Is there any way we can provide more info on the relationship that’s being edited? Currently you’d have to extract the join table & model name and try to work backwards from there to work out which relationship you’re editing
- We should include `$extraField` in the data passed to the extensions
- There should be `after` as well as `before` hooks

maxime-rainville:
I'm having a difficulty imagining a scenario where I would want to be notified of every single relationship assignment on every single DataObject. I would probably want to be notified only of specific relationships on a specific DataObject class.

If we provide this kind of functionality it would be better if it was targeted.

I'm thinking it might be possible to provide this functionality by extending `ManyManyList` instead. `Member_GroupSet` is an example of where we do this for the `Member::Groups()` relation.

kinglozzer:
The cacheinclude module for SS3 used a similar approach, it replaced `ManyManyList` “globally” via injector: https://github.com/heyday/silverstripe-cacheinclude/blob/master/src/SilverStripe/ManyManyList.php

It’s going to be hard to create extension hooks that are:

- Flexible enough to cover most use cases
- Targeted enough to not be called on _every_ add/remove for every object type
- Performant enough to not add too much overhead (e.g. `$list->add(123)` would have to perform a query to fetch a DataObject instance to call an `extend` hook on it)

ChrissiQ:
Thanks for the feedback on this.  I do see your points, I wasn't considering the performance and flexibility of this addition.  I was also not aware of how expensive such a hook could be.  It looks like this PR isn't solving the problem I had in an effective way.  However, I am having a hard time imagining why detecting updates to objects through their many-many relationships is a "unique use case".  It seems more like a hole in the feature coverage between partial caching and many-many relationships.

The docs explain how to use partial caching, but this partial caching doesn't accommodate for relationships being added and removed from objects because the LastEdited field is not changed on any of the objects.  In our application, we are using partial caching on a page that would normally run several expensive queries to display lists of data from models, and their related objects through many-many relationships.  Being able to add partial caching to this page is a great feature of Silverstripe that we've taken advantage of here to achieve much better performance in general.

The problem is that changes to the relationships of the models does not trigger an update to the partial caching key.  It's confusing for our clients when they make a change to the data but do not see a change reflected on the front end of the website because the partial caching key for these objects hasn't changed.  I need a way to change the partial caching key when relational data is updated on these objects.

I am still searching for a solution to this problem that I can actually put into effect now on our SS4 site.  If I can't, it would be great to actually get something in to solve it in the future.  From my perspective it seems like a gap that needs to be filled.

kinglozzer:
> However, I am having a hard time imagining why detecting updates to objects through their many-many relationships is a "unique use case". It seems more like a hole in the feature coverage between partial caching and many-many relationships.

Sorry @ChrissiQ, I worded that poorly! I meant to say that in order to be as useful as possible, the extension hooks here need to expose a lot more detail than most other hooks in the framework do.

> From my perspective it seems like a gap that needs to be filled.

I agree that there’s a gap in the functionality here - both in terms of partial cache key generation and in an extension hook that would be nice to have.

As a stopgap solution, it should be possible to create a cache key manually that contains enough data from the relationship to guarantee the key will change when a relationship does. Exactly how that might work depends on your data structure, but if the [example](https://docs.silverstripe.org/en/4/developer_guides/performance/partial_caching/#cache-key-calculated-in-controller) in the docs doesn’t invalidate as expected or isn’t performant enough you might be able to query the join table directly and generate part of your key from that - something like:

```php
$query = SQLSelect::create('ID', 'Join_Table');
$joinIds = $query->execute()->column('ID');
$cacheKeyFragment = md5(implode('', $joinIds));
```

ChrissiQ:
Thanks @kinglozzer, that did the trick for my problem.  Not sure why I didn't think of that.

ChrissiQ:
I'll close this PR since obviously this requires a different approach, and I got a solution to the problem I was trying to fix anyway.

dhensby:
Another option that may be easier in the template (but also more error prone) would be to use the count of the relationship. `$Relationship.Count`

Issue #7660, opened 2017-11-29
[ORM] BUG Incorrect case on case sensitive DB servers is not being fixed


tractorcow:
Reported from https://github.com/silverstripe/silverstripe-framework/issues/4451

a table `sitetree_live` is not being renamed to `SiteTree_Live` on dev/build.

Report from @cansozeri

> The problem is still exists with ss4. I have copied my database to a linux server and I get same errors because of table names :=) Dev/build fix that except one table, I had to change it manually :=)

My follow up comment:

> In 4.0 we have a new config.

> `SilverStripe\ORM\Connect\DBSchemaManager.fix_table_case_on_build`.

> This should fix sitetree_live to SiteTree_Live on case sensitive filesystems. If it is not, there is a real bug.

# Related PRs
* [ ] https://github.com/silverstripe/silverstripe-framework/pull/8286

wilr:
This affecting deployments of CWP. In my case Page_versions exists, not Page_Versions

```
[2018-07-25 12:57:37]   * [1;31mERROR [Emergency]: Uncaught SilverStripe\ORM\Connect\DatabaseException: Couldn't run query:
  * 
  * ALTER TABLE "Page_Versions" RENAME "_obsolete_Page_Versions2"
  * 
  * Table 'SS_cwp.Page_Versions' doesn't exist
  * IN GET dev/build
```

tractorcow:
@wilr which framework version does it use? I recall we fixed _versions with a patch a while back, but I forgot if it was in 4.1 or 4.2.

wilr:
I'm using `4.2` and still seeing something related to that so that patch might not be quite enough.

Workaround for this is https://github.com/silverstripe/silverstripe-framework/pull/8286.

wilr:
@robbieaverill yes I haven't had this issue crop up since.

PR #8286, opened 2018-07-25
Check database has table before altering.

wilr:
Work around to prevent several issues with silverstripe-framework#4451 and cwp#0031726.

Preferred to fix it here globally as otherwise checks were needed for framework, versioned and CMS.

# Parent Issue
* https://github.com/silverstripe/silverstripe-framework/issues/7660

### DIFF
--- a/src/ORM/Connect/MySQLSchemaManager.php
+++ b/src/ORM/Connect/MySQLSchemaManager.php
@@ -4,6 +4,7 @@ namespace SilverStripe\ORM\Connect;
 
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Convert;
+use LogicException;
 
 /**
  * Represents schema management object for MySQL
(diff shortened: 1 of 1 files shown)


wilr:
As per the comment originally, I could check `hasTable` at the callee level but it means patches for several of the modules (and in the short term mantaining forks for 3+ repos I've found this issue in. 

This keeps the change clean and reduces the amount of boilerplate code each module needs to do (i.e I don't think you would ever not check before altering)

tractorcow:
> I would rather renameTable throws a LogicException.

I think I concur with this. You run the risk of the application silently ignoring what could be a real bug.

wilr:
Ok fair.. 2 v 1 democratic process and all. I've updated to throw a `LogicException` and will update the submodules individually to listen for it and update accordingly.

PR #8292, opened 2018-07-27
FIX Altering ID of authenticator tabs to resolve ID conflict

ScopeyNZ:
Currently when using multiple authenticators it uses a template called `Security_MultiAuthenticatorTabbedForms.ss`: https://github.com/silverstripe/silverstripe-framework/blob/4.2/templates/SilverStripe/Security/Security_MultiAuthenticatorTabbedForms.ss

The problem is the tab IDs are set to the form name, but the default rendering of the LoginForms also uses the form name as the ID of the form. This results in duplicate IDs (which makes an invalid DOM).

This PR alters the ID (and prior reference) of the tabs so they are suffixed with `_Tab`.

### DIFF
--- a/templates/SilverStripe/Security/Security_MultiAuthenticatorTabbedForms.ss
+++ b/templates/SilverStripe/Security/Security_MultiAuthenticatorTabbedForms.ss
@@ -1,12 +1,12 @@
 <ul>
 	<% loop $Forms %>
 	<li>
-		<a href="#{$FormName}">$AuthenticatorName</a>
+		<a href="#{$FormName}_Tab">$AuthenticatorName</a>
 	</li>
 	<% end_loop %>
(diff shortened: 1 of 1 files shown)


dhensby:
well... is this a breaking change? what's the impact on existing themes / users of this change?

ScopeyNZ:
Considering there was always and ID conflict I can't imagine anyone was actually using this ID to provide styling or tie in JS. Additionally, the multi-authenticator default styling is not very... feature complete... I highly doubt anyone would be affected by this change.

dhensby:
ok - deferring to @robbieaverill

robbieaverill:
I'm happy with the change, but I'll leave it for another core team member to agree with me before merging it =)

PR #8298, opened 2018-07-31
DOCS Update cow docs to reflect current process

dhensby:
The cow release process doesn't reflect the current state of play.

For better or worse, these docs have been updated to reflect the actual process as it currently is.

### DIFF
--- a/docs/en/05_Contributing/05_Making_A_SilverStripe_Core_Release.md
+++ b/docs/en/05_Contributing/05_Making_A_SilverStripe_Core_Release.md
@@ -28,10 +28,10 @@ As a core contributor it is necessary to have installed the following set of too
 * [cow release tool](https://github.com/silverstripe/cow#install). This should typically
   be installed in a global location via the below command. Please see the installation
   docs on the cow repo for more setup details.
-  `composer global require silverstripe/cow dev-master`
+  `composer global require silverstripe/cow ^2`
 * [satis repository tool](https://github.com/composer/satis). This should be installed
   globally for minimum maintenance.
-  `composer global require composer/satis ^1@dev` (or `^1@alpha` for php 5)
+  `composer global require composer/satis ^1`
 * [transifex client](http://docs.transifex.com/client/).
   `pip install transifex-client`
   If you're on OSX 10.10+, the standard Python installer is locked down.
@@ -149,9 +149,8 @@ this we use [satis](https://github.com/composer/satis).
 
 To setup a Satis project for a release:
 
-* Ensure Satis is installed globally: `composer global require composer/satis ^1@dev` (or `^1@alpha` if you
-  are running php 5)
-* `cd ~/Sites/` (or whereever your web-root is located)
+* Ensure Satis is installed globally: `composer global require composer/satis ^1` 
+* `cd ~/Sites/` (or wherever your web-root is located)
 * `mkdir satis-security && cd satis-security` (or some directory specific to your release)
 * Create a config file (e.g. config.json) of the given format (add only those repositories necessary).
 
@@ -200,11 +199,6 @@ The release process, at a high level, involves creating a release, publishing it
 reviewing the need for either another pre-release or a final stable tag within a short period
 (normally within 3-5 business days).
 
-During the pre-release cycle a temporary branch is created, and should only receive
-absolutely critical fixes during the cycle. Any changes to this branch should
-result in the requirement for a new release, thus a higher level of scrutiny is typically
-placed on any pull request to these branches.
-
 When creating a new pre-release or stable, the following process is broken down into two
 main sets of commands:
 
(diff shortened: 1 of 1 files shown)


docs/en/05_Contributing/05_Making_A_SilverStripe_Core_Release.md
```diff
@@ -137,11 +137,6 @@ The release process, at a high level, involves creating a release, publishing it
 reviewing the need for either another pre-release or a final stable tag within a short period
 (normally within 3-5 business days).
 
-During the pre-release cycle a temporary branch is created, and should only receive
```
dhensby: It's a shame we have to remove this block; but at the moment we just don't do this. Hopefully we will again soon.

docs/en/05_Contributing/05_Making_A_SilverStripe_Core_Release.md
```diff
@@ -213,14 +212,8 @@ Once the release task has completed, it may be ideal to manually test the site o
 by running it locally (e.g. `http://localhost/release-3.3.4`) to do some smoke-testing
 and make sure that there are no obvious issues missed.
 
-Since `cow` will only run the unit test suite, you'll need to check
-the build status of Behat end-to-end tests manually on travis-ci.org.
```
dhensby: Well, at the moment there's no way to push to travis without publishing the tag, so this is misleading

docs/en/05_Contributing/05_Making_A_SilverStripe_Core_Release.md
```diff
@@ -213,14 +212,8 @@ Once the release task has completed, it may be ideal to manually test the site o
...
-Check the badges on the various modules available on [github.com/silverstripe](http://github.com/silverstripe).
-
 It's also ideal to eyeball the git changes generated by the release tool, making sure
-that no translation strings were unintentionally lost, no malicious changes were
```
dhensby: I don't see how we can expect someone performing a release to know if there have been malicious changes to translation strings unless they speak 100+ languages

docs/en/05_Contributing/05_Making_A_SilverStripe_Core_Release.md
```diff
@@ -264,23 +257,23 @@ subtasks which are invoked in sequence:
 Once all of these commands have completed there are a couple of final tasks left that
 aren't strictly able to be automated:
 
-* If this is a stable release, it will be necessary to perform a post-release merge
```
dhensby: we should post release merge all our releases, we want RC logs be published on docs.silverstripe.org, don't we?

docs/en/05_Contributing/05_Making_A_SilverStripe_Core_Release.md
```diff
@@ -264,23 +257,23 @@ subtasks which are invoked in sequence:
...
   done the easier it is, but this can sometimes be left for when you have
   more free time. Branches not receiving regular stable versions anymore (e.g.
-  3.0 or 3.1) should usually be omitted.
+  3.0 or 3.1) can be omitted.
 * Set the github milestones to completed, and create placeholders for the next
   minor versions. It may be necessary to re-assign any issues assigned to the prior
   milestones to these new ones.
 * Make sure that the [releases page](https://github.com/silverstripe/silverstripe-installer/releases)
-  on github shows the new tag. 
+  on github shows the new tag. [what to do if it's not showing?!]
```
dhensby: It's all well and good saying that the release should show, but what should we do if it doesn't?

robbieaverill: I guess it's be a case of manually going into the module, checking the tag exists and pushing it. If not, re-run the release process? I assume it would do these checks again and re-tag if necessary

docs/en/05_Contributing/05_Making_A_SilverStripe_Core_Release.md
robbieaverill: Fair call. I think maybe this was more for obvious things like defacing (in English) or malicious HTML

docs/en/05_Contributing/05_Making_A_SilverStripe_Core_Release.md
robbieaverill: Maybe we mention that this would be checking the build status of the latest branch

docs/en/05_Contributing/05_Making_A_SilverStripe_Core_Release.md
chillu: Yeah, I wrote this originally, and the intent was to look for malicious HTML. We shouldn't really allow HTML in translation strings at all, but that's much tougher to fix. I vote for keeping this line in there, we've had no end of trouble with translations, so some form of "translation sanity checking" is required.

unclecheese:
Have rebased after the merge of the [Satis PR](https://github.com/silverstripe/silverstripe-framework/pull/7681) and retargeted to 4.0. 

This is looking much better. Thanks, everyone.

PR #8303, opened 2018-08-02
Implement __isset for ViewableData_Customised

### LINKED ISSUE #8302: Outer ViewableData_Customised returns null for fields contained by a nested ViewableData_Customised
## Affected Version
4.2.1

## Description
`ViewableData_Customised` returns `null` for properties set in a nested `ViewableData_Customised`.

## Steps to Reproduce
1. `$email = Email::create()->setData(instanceof ViewableData_Customised)`
2. Render/send `$email` as an HTML template.
3. Properties in the `ViewableData_Customised` instance are `null` in the template context, because it is nested in another `ViewableData_Customised` by `$this->renderWith()` in the `Email` class.

hchokshi:
Resolves #8302.

When a nested `ViewableData_Customised` provides some fields, an outer `ViewableData_Customised` will return `isset($outer->nestedField) === false` and so `$outer->nestedField === null` in contexts like a template.

Implementing `__isset` to check the nested objects resolves this by returning `isset === true` for nested fields that are available. This also allows for arbitrary levels of nesting in a `ViewableData_Customised` to be resolved recursively.

### DIFF
--- a/tests/php/View/ViewableDataCustomisedTest.php
+++ b/tests/php/View/ViewableDataCustomisedTest.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace SilverStripe\View\Tests;
+
+use SilverStripe\Dev\Constraint\ViewableDataContains;
+use SilverStripe\Dev\SapphireTest;
+use SilverStripe\View\ArrayData;
+use SilverStripe\View\ViewableData_Customised;
+
+/**
+ * Test for ViewableData_Customised.
+ */
+class ViewableDataCustomisedTest extends SapphireTest
+{
+    public function testNestedViewableDataCustomisedAsCustomised()
+    {
+        $outerCustomised = ViewableData_Customised::create($this->makeOuterOriginal(), $this->makeInnerViewableDataCustomised());
+        $this->assertThat($outerCustomised, $this->makeTestConstraint());
+    }
+
+    public function testNestedViewableDataCustomisedAsOriginal()
+    {
+        $outerCustomised = ViewableData_Customised::create($this->makeInnerViewableDataCustomised(), $this->makeOuterOriginal());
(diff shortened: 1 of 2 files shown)


tests/php/View/ViewableData_CustomisedTest.php
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Test for ViewableData_Customised.
+ */
+class ViewableData_CustomisedTest extends SapphireTest
+{
+    public function testNestedViewableData_CustomisedAsCustomised()
+    {
+        $outerCustomised = ViewableData_Customised::create($this->makeOuterOriginal(), $this->makeInnerViewableData_Customised());
+        self::assertThat($outerCustomised, $this->makeTestConstraint());
```
robbieaverill: We generally prefer to use `$this->assert*` rather than the static equivalent

hchokshi:
Hey @robbieaverill I've updated the `assertThat` calls, which also re-triggered the tests and the travis composer install issue with the `4.2` branch seems to be resolved now. Is this ready for a merge?

tests/php/View/ViewableDataCustomisedTest.php
```diff
@@ -0,0 +1,55 @@
...
+use SilverStripe\Dev\SapphireTest;
+use SilverStripe\View\ArrayData;
+use SilverStripe\View\ViewableData_Customised;
+
+/**
+ * Test for ViewableData_Customised.
+ */
+class ViewableDataCustomisedTest extends SapphireTest
+{
+    public function testNestedViewableData_CustomisedAsCustomised()
```
robbieaverill: I'm really nit picking, I know 😄 there's still two methods here with underscores

hchokshi:
@robbieaverill Whoops I only did the ones from the original review. Renamed all now.

PR #8304, opened 2018-08-06
BUGFIX Installer redirect to home/ (without domain)

micmania1:
This fixes a bug where the installer redirects to home/ ('home' as the domain) as the url as opposed to localhost/home.

To reproduce:
* Use chrome (untested in other browsers)
* composer create project silverstripe/installer
* Run through the installer

Expected:
* Silverstripe installs and it takes you to the success page

Actual:
* SilverStripe installs and takes you to a different domain (ie. 'http://home')

The problem seems to be, any URL starting with `//` (two slashes) is normalised by the browser to an absolute URL by removing both slashes meaning `//home` becomes just `home` (as in http://home)

### DIFF
--- a/src/Dev/Install/Installer.php
+++ b/src/Dev/Install/Installer.php
@@ -601,7 +601,7 @@ TEXT;
         $token = new ParameterConfirmationToken('flush', new HTTPRequest('GET', '/'));
         $params = http_build_query($token->params());
 
-        $destinationURL = BASE_URL . '/' . (
+        $destinationURL = rtrim(BASE_URL, '/') . '/' . (
             $this->checkModuleExists('cms')
                 ? "home/successfullyinstalled?$params"
                 : "?$params"


micmania1:
As this is the installer, I don't yet have a `.env`. However, post-install my SS_BASE_URL is being set to my hostname correctly.

Pre-install, it tries to work out a relative base url, which it correctly does as `/`. However, on the line I fixed, it also concatenates the BASE_URL with another `/` and then the URL segment 'home' (ie. `//home`).

The redirect to that page happens in the browser and chrome is treating `//home` as its own domain (ie. `http://home`) and i'm guessing other browsers would too.

The extra `/` which is being added on the fixed line needs to remain in case a user is installing silverstripe into a directory. eg. http://localhost/my-website where BASE_URL would become `/my-website`.

maxime-rainville:
I'm just thinking there might be a bug in `includes/constants.php`.

https://github.com/silverstripe/silverstripe-framework/blob/3b06385b7d4b7aab0408e21cf62668da35d3b210/src/includes/constants.php#L99-L153

That logic is meant to remove the trailing slash. If there's a problem in the way BASE_URL is set, it would be better to fix it there.

PR #8313, opened 2018-08-10
Readonlyfield return value broke history view

robbyahn:
Readonlyfield component  value() method update broke history view after upgrade from 3.4.1 -> 3.4.3.
Please review this is good to go.

### DIFF
--- a/forms/ReadonlyField.php
+++ b/forms/ReadonlyField.php
@@ -53,7 +53,7 @@ class ReadonlyField extends FormField {
 	}
 
 	public function Value() {
-		if($this->value) return $this->value;
+		if($this->value) return $this->dontEscape ? $this->value : Convert::raw2xml($this->value);
 		else return '<i>(' . _t('FormField.NONE', 'none') . ')</i>';
 	}
 


robbieaverill:
@robbyahn would you mind providing a little more information about the problem that this pull request fixes?

dhensby:
> Are we still accepting non-security-related bug fixes for 3.x?

Nope

dhensby:
The test failures also show that this has caused a functional regression

dhensby:
I think the correct fix is to udate the casting on your formfield

robbyahn:
Hi guys,

I've got the following markup related code.
After adding slash thing, it is no casting issue.
DBField::create_field('HTMLText', '<span class=\"icons--xl\"><i class=\"i i-section-$icon\"></i></span>');

Thanks, Leapfrognz
https://github.com/silverstripe/silverstripe-dms/pull/216

Is there any change escape things?

robbyahn:
I was misleading about this.
There was return value 

$this->getManyManyComponenetns('YOUR_DATA_OBJECT')->sort('SortOrder')

After removing this linke. It doesn't need to pass the object value to form anymore when history panel is loaded.

Issue #7237, opened 2017-08-02
SS4 Bug: Session is missing at Security controller


a2nt:
Trying to open /Security/login

`error-log.ERROR: Uncaught Exception BadMethodCallException: "No session available for this HTTPRequest" at /Volumes/Secondary/www/bids.local.pro/framework/src/Control/HTTPRequest.php line 885 {"exception":"[object] (BadMethodCallException(code: 0): No session available for this HTTPRequest at /Volumes/Secondary/www/bids.local.pro/framework/src/Control/HTTPRequest.php:885)"} []`

tractorcow:
Hi @a2nt can you please give us a full stacktrace for this issue? How can it be replicated on a clean install? Thank you. :)

a2nt:
The problem with PageController. There's a function called from template:
```
public function SiteWideMessage()
    {
        if (!isset($this->siteWideMessage)) {
            $session = $this->getRequest()->getSession();
            $this->siteWideMessage = unserialize($session->get('SiteWideMessage'));
            $session->clear('SiteWideMessage');
        }

        return  $this->siteWideMessage ? ArrayData::create($this->siteWideMessage) : false;
    }
```

$this->getRequest()->getSession() doesn't works which is weird, but Controller::curr()->getRequest()->getSession() works ok

tractorcow:
That makes sense... it constructs a new page and page controller, but doesn't set the request to that which is current. I think this can be fixed.

tractorcow:
Fix at https://github.com/silverstripe/silverstripe-framework/pull/7240

martinduparc:
This still seems to be an issue in v4.2.1. When accessing the CMS admin, `$this->getRequest()->getSession()` will generate an exception. The way I am getting passed it is to check the class of `$this->getRequest()` as shown below.

```php
class PageController extends ContentController {

    private static $allowed_actions = [];

    protected function init() {
        parent::init();

        if (get_class($this->getRequest()) == \SilverStripe\Control\HTTPRequest::class) {
            $session = $this->getRequest()->getSession();
        }
    }

    ...
}

robbieaverill:
@martinduparc I think your problem is because when Security creates a dummy page in order to use its controller, it calls doInit() before providing the request. I'll make a pull request to switch these around.

https://github.com/silverstripe/silverstripe-framework/blob/4.2.1/src/Security/Security.php#L579-L581

PR #8324, opened 2018-08-14
FIX Pass request to dummy controller before calling init

robbieaverill:
See https://github.com/silverstripe/silverstripe-framework/issues/7237#issuecomment-413008681


martinduparc:
Thanks, Robbie

PR #8325, opened 2018-08-14
FIX GridField delete button to offer archive action if possible

lukereative:
Fixes silverstripe/silverstripe-versioned#78

## Related PRs
- https://github.com/silverstripe/silverstripe-admin/pull/602
- https://github.com/silverstripe/silverstripe-versioned/pull/171
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/42

### DIFF
--- a/src/Forms/GridField/GridFieldDeleteAction.php
+++ b/src/Forms/GridField/GridFieldDeleteAction.php
@@ -200,9 +200,9 @@ class GridFieldDeleteAction implements GridField_ColumnProvider, GridField_Actio
                         _t(__CLASS__ . '.EditPermissionsFailure', "No permission to unlink record")
                     );
                 }
-            }
 
-            $gridField->getList()->remove($item);
+                $gridField->getList()->remove($item);
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/Forms/GridField/GridFieldDeleteAction.php
```diff
@@ -179,29 +179,51 @@ public function getColumnContent($gridField, $record, $columnName)
...
+            switch ($actionName) {
+                case 'archiverecord':
+                    if (!$item->canArchive()) {
+                        throw new ValidationException(
+                            _t(__CLASS__ . '.ArchivePermissionsFailure', "No archive permissions")
+                        );
+                    }
+
+                    $item->doArchive();
+                    break;
```
robbieaverill: I'm don't think I agree with putting this code here. It's specific to the versioned module

robbieaverill: Perhaps we can put this into a versioned extension on this class instead?

robbieaverill:
I'm all for making private things protected. This is a good example of why we shouldn't use it in our modules

unclecheese:
Actually, when I test this feature with the framework on the `4` branch, everything still works. 

I think, judging by the timeline on the commits and @robbieaverill 's feedback, @lukereative [moved this stuff over to Versioned](https://github.com/silverstripe/silverstripe-versioned/pull/171/commits/61c046c9cd6ce97456b6123a05438c7cd05d07cc), effectively deprecating this PR.

Can you confirm that, @lukereative ? It looks to me like this needs to be closed.

lukereative:
I had moved it but I realised some commits hadn't been pushed – so this PR is still relevant but with just very minor class changes

PR #8326, opened 2018-08-15
FIX Remove double escaping of HTML values in print views

robbieaverill:
Print view uses the SilverStripe templating to render values which means that
values are safely escaped by default. This can be tested by chaing `$CellString`
to `$CellString.RAW` in the GridField_print.ss template to see this escaping
being disabled.

This pull request removes double escaping of HTML in strings.

Related issue: https://github.com/silverstripe/silverstripe-userforms/issues/633


robbieaverill:
Self merging with @dhensby's approval

Issue #5594, opened 2016-05-25
text/json is not a content type


dhensby:
The correct content type for json is [`application/json`](https://en.wikipedia.org/wiki/JSON)

There are currently 11 references to `text/json` in framework (and likely more in cms). We even assert for them in tests.

tractorcow:
Ok, go ahead!

dhensby:
fine, I will

dhensby:
`text/json` is all over 4.x too :(

PR #8327, opened 2018-08-15
FIX text/json is not a valid mimetype

dhensby:
Partial fix of https://github.com/silverstripe/silverstripe-framework/issues/5594

### DIFF
--- a/src/Control/PjaxResponseNegotiator.php
+++ b/src/Control/PjaxResponseNegotiator.php
@@ -103,7 +103,7 @@ class PjaxResponseNegotiator
             }
         }
         $response->setBody(Convert::raw2json($responseParts));
-        $response->addHeader('Content-Type', 'text/json');
+        $response->addHeader('Content-Type', 'application/json');
 
         return $response;
     }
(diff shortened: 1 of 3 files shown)


robbieaverill:
Do you want to target a patch release?

lerni:
targeting a patch release... I wonder when "text/json" is replaced with "application/json" on php side, will the new mime-type be mainstream enough on production servers or does it need to be added to .htaccess & equivalent like 'AddType 'application/json; charset=UTF-8' json' or just mentioned in release notes?

kinglozzer:
@lerni I’d hope that people don’t have `text/json` in their `.htaccess` as it’s not a valid mime type. If they _do_ have that, but don’t have `application/json`, then in theory this could result in some caching changes - but it will also fix things in the opposite case!

dhensby:
> Do you want to target a patch release?

personally I don't think it should go in a patch release; I intended it at the minor release

dhensby:
@lerni `text/json` was never a valid mime-type so there shouldn't be any webserver support for it ever...

There was a draft `text/x-json` but that was never official

lerni:
@dhensby I've seen it on a server and I think it was used quite a bit. Since it was never official and application/json and is more with spread nowadays, sure this fixes more than it breaks.

PR #8337, opened 2018-08-20
Allow use of display names with SwiftMailer

martinduparc:
The syntax to allow display names with SwiftMailer is as follow:
`->setFrom(['john@doe.com' => 'John Doe'])`

Just like you set the `Email.admin_email`, you can set the default sender display name through the `Email.admin_email_displayname` configuration setting.

Refs:
- https://swiftmailer.symfony.com/docs/introduction.html#basic-usage
- https://docs.silverstripe.org/en/4/developer_guides/email/#administrator-emails

### DIFF
--- a/docs/en/02_Developer_Guides/10_Email/index.md
+++ b/docs/en/02_Developer_Guides/10_Email/index.md
@@ -133,6 +133,14 @@ SilverStripe\Control\Email\Email:
   admin_email: support@example.com
 ```
 
+To add a display name, set `admin_email` as follow.
+
+```yaml
+SilverStripe\Control\Email\Email:
+  admin_email:
+    support@example.com: 'Support team'
+```
+
 <div class="alert" markdown="1">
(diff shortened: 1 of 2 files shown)


dhensby:
Couldn't one just set the `Email.admin_email` option to be an associative array? Perhaps a docs update would be more appropriate?

eg:

```yml
Email:
  admin_email:
    me@example.com: SilverStripe Admin
```

dhensby:
there are also linting issues in this PR, please see the travis failure

martinduparc:
Hi Daniel,

Definitely, whatever makes the more sense to the SS core team. I will go ahead and make that change.

dhensby:
Less code to maintain is best IMO and if we can achieve it like this, that'd be great.

Can you update the docblock for the config option to say it can be an array and also update the email docs if there's somewhere appropriate to add this?

martinduparc:
Thanks for the feedback. I've made the changes

martinduparc:
Hi @maxime-rainville Feel free to squash-and-merge. Thanks!

Issue #3357, opened 2014-08-03
[GRIDFIELD] GridField needs a useful readonly view (currently just showing "none")


sunnysideup:
## Acceptance Criteria

 * I can click through GridField rows in readonly view (without action buttons)
 * I can perform standard GridField view operations (pagination, search, sorting)
 * I can click into a row of a readonly GridField to see it's edit view (which might also be readonly)
 * Edit views on rows in a readonly GridField are also marked readonly (regardless of `DataObject->canEdit()`) - this is important for history viewer
 * No state-altering actions are shown (create buttons, batch actions)

## Notes

 * Related: [Diff view for related records](https://github.com/silverstripe/silverstripe-versioned/issues/97)
 * If all records in the GridField aren't editable, it still shows an editable GridField
 * Components should be able to self determine if and how they're shown in readonly views
 * Track under ORB-175

## PR
- [x] https://github.com/silverstripe/silverstripe-framework/pull/8535

## Historic Notes

I would like to lodge the following bug just in case it is one ...

Hi,

I have a very strange situation.  I dont know if it is stupidity (80% sure it is!!!) or something else - I am just posting this here, because it may be of interest to some of you and because I am not sure if it is a bug or something else.

This code works fine on one data object and not on another:

``` php
function getCMSFields(){
        ....
        $config = GridFieldConfig_RecordViewer::create();
        $dataList = Member::get();
        $gridField = new GridField('AA','AA'.$dataList->count(), $dataList, $config);
        $fields->addFieldToTab("Root.BB", $gridField);
        ....
}
```

On DataObject 1 it returns a normal gridfield - its title is AA7 (i.e. there are seven members), but on dataobject 2 it returns as field that says: 

AA7 (none)

This is unexpected. 

What I know:

it only is "funny" on one dataobject

it does not matter if I use Member or Page or MyOtherDataObject as datalist source, it always return as "none"

the other cms fields are not causing the error (I removed all the other fields and I ended up with the same results). 

also with an actual relation (this is what I am working with e.g. Players on Team), then it still returns as "none", despite their being record available.

I am using framework 3.1.5 and I have not experienced this problem before.  

It is an extensive application with plenty of code so some magic word or unexpected far-flung bug may cause the error.

in the getCMSFields, I dont use parent::getCMSFields, but even if I do, I get the same error. 
[...]

PR #8348, opened 2018-08-29
Issue 3357 - Add GridField Readonly Transformation

caffeineinc:
GridField doesn't have a valid readonly state if it's value is set to an Object such as HasManyRelation
without the method `forTemplate()`. The default behaviour is to render GridField a ReadonlyField, but this calls this method, and raises an exception.

This bug fix provides a transformation method to GridField to render only components that are
whitelisted to provide a readonly state, while still rendering the contents of the GridField. (Action buttons, add new buttons etc are removed). 

@see #3357 - https://github.com/silverstripe/silverstripe-framework/issues/3357

### DIFF
--- a/src/Forms/GridField/GridField.php
+++ b/src/Forms/GridField/GridField.php
@@ -106,6 +106,21 @@ class GridField extends FormField
     protected $name = '';
 
     /**
+     * A whitelist of readonly component classes allowed if performReadonlyTransform is called.
+     *
+     * @var array
+     */
+    protected $readonlyComponents = array(
+        GridFieldDetailForm::class,
+        GridFieldDataColumns::class,
+        GridFieldConfig_RecordViewer::class,
+        GridFieldToolbarHeader::class,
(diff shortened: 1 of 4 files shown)


NightJar:
Looks pretty cool.

The only thing I might have to add to this would be that perhaps providing a `setValue` no-op type method on `GridField` could also help mitigate this issue (in that it makes no sense to set a value for a GridField - it manages relationships independently of the form state within which it exists.

NightJar:
I don't think state and value are equivalent. The GridField state tracks operational management things to do with the field, such as which sort column is active, etc. I perceive the value as a sort of `return` scalar of the field in question. In this manner a GridField returns `null` (to loosely equivilate it to code).

chillu:
@caffeineinc Is this still causing a fatal error, breaking the entire edit form? It's marked as `impact/critical` because of it. There's an entirely separate, lower priority issue for actually creating a readonly GridField, for now we just need to fix the fatal error.

PR #8351, opened 2018-08-30
BUG Use classes for TinyMCE alignment buttons

lukereative:
Fixes silverstripe/silverstripe-cms#2202

Rather than using inline-styles, the alignment buttons in the HTMLEditor field will now use classes which matches ones typically included in `editor.css`: 

For text specific elements i.e. `p` or `h1`:
- `text-left`
- `text-center`
- `text-right`
- `text-justify`

For block elements i.e. `img` or `div`:
- `left`
- `center`
- `right`

### DIFF
--- a/src/Forms/HTMLEditor/TinyMCEConfig.php
+++ b/src/Forms/HTMLEditor/TinyMCEConfig.php
@@ -216,8 +216,46 @@ class TinyMCEConfig extends HTMLEditorConfig
      *
      * @var array
      */
-    protected $settings = array(
+    protected $settings = [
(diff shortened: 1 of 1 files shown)


robbieaverill:
Nice - one thought though is that this has some coupling to Bootstrap right? Not everyone uses Bootstrap on their frontends

lukereative:
Sort of, those classes have been around in SilverStripe for a while (at least I've had them in sites I've built) – I think they used to be a part of the default editor config. At the end of the day users can customise those classes with their own tinymce config if they want different ones for their environment and I think those ones are the best logical default and lines up with the ones we apply via the file dialog anyway (`right`, `left` and `center`).

robbieaverill:
Lovely

PR #8359, opened 2018-09-04
Add schema information for HTMLEditorField

NightJar:
This allows React form builders (or other such view layer builders in a
headless environment) to obtain the details that would otherwise only be
rendered in a PHP side template. Some of the details are critical for
rendering correctly, and are necessary to be passed through -
particularly when moving toward replacing the Entwine initiator for
TinyMCE with a React component in `silverstripe/admin`.

Required as part of https://github.com/dnadesign/silverstripe-elemental/issues/327

### DIFF
--- a/src/Forms/HTMLEditor/HTMLEditorField.php
+++ b/src/Forms/HTMLEditor/HTMLEditorField.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\Forms\HTMLEditor;
 
 use SilverStripe\Assets\Shortcodes\ImageShortcodeProvider;
+use SilverStripe\Forms\FormField;
 use SilverStripe\Forms\TextareaField;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataObjectInterface;
@@ -22,6 +23,10 @@ class HTMLEditorField extends TextareaField
     private static $casting = [
         'Value' => 'HTMLText',
     ];
+
+    protected $schemaDataType = FormField::SCHEMA_DATA_TYPE_HTML;
+
+    protected $schemaComponent = 'HtmlEditorField';
 
     /**
      * Use TinyMCE's GZIP compressor
@@ -184,5 +189,13 @@ class HTMLEditorField extends TextareaField
         // Include requirements
         $this->getEditorConfig()->init();
         return parent::Field($properties);
+    }
+
+    public function getSchemaStateDefaults()
+    {
+        $stateDefaults = parent::getSchemaStateDefaults();
+        $config = $this->getEditorConfig();
+        $stateDefaults['data'] = $config->getConfigSchemaData();
+        return $stateDefaults;
     }
 }
--- a/src/Forms/HTMLEditor/HTMLEditorConfig.php
+++ b/src/Forms/HTMLEditor/HTMLEditorConfig.php
@@ -224,4 +224,17 @@ abstract class HTMLEditorConfig
      * Initialise the editor on the client side
      */
     abstract public function init();
+
+    /**
+     * Provide additional schema data for the field this object configures
+     *
+     * @return array
+     */
+    public function getConfigSchemaData()
+    {
+        return [
+            'attributes' => $this->getAttributes(),
+            'editorjs' => null,
+        ];
+    }
 }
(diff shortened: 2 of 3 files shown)


src/Forms/HTMLEditor/HTMLEditorField.php
```diff
@@ -185,4 +185,16 @@ public function Field($properties = array())
...
+
+    public function getSchemaStateDefaults()
+    {
+        $stateDefaults = parent::getSchemaStateDefaults();
+        $config = $this->getEditorConfig();
+        $stateDefaults['data']['config'] = $config->getAttributes();
+        $stateDefaults['data']['editorjs'] = null;
+        if (method_exists($config, 'getScriptURL')) {
+            $stateDefaults['data']['editorjs'] = $config->getScriptURL();
+        }
```
robbieaverill: Perhaps you could overload this method in `TinyMCEConfig` and add in the part that's specific to it there?

src/Forms/HTMLEditor/HTMLEditorField.php
```diff
@@ -185,4 +185,16 @@ public function Field($properties = array())
         $this->getEditorConfig()->init();
         return parent::Field($properties);
     }
+
+    public function getSchemaStateDefaults()
+    {
+        $stateDefaults = parent::getSchemaStateDefaults();
+        $config = $this->getEditorConfig();
+        $stateDefaults['data']['config'] = $config->getAttributes();
```
flamerohr: maybe this should be named more appropriately as `$stateDefaults['data']['attributes']`, looking from the otherside it's quite confusing

src/Forms/HTMLEditor/HTMLEditorField.php
NightJar: ooh, that's a much better idea 👍

src/Forms/HTMLEditor/HTMLEditorField.php
NightJar: Seems fair :)

src/Forms/HTMLEditor/HTMLEditorField.php
NightJar: oh, whoops... this is `HTMLEditorField`, not `HTMLEditorConfig` ;) There is no `TinyMCEField` to override the method in.

robbieaverill: I see. I think we can probably to this in a more structured way still though. Perhaps `HTMLEditorConfig` and `TinyMCEConfig` need a new method like `getSchemaData()` added?

robbieaverill: Yeah I guess the options are:

1) Config provides its own schema data, irrespective of the format of the HTMLEditorField's schema -> call it `getSchemaData` and return an isolated array
2) Config augments the existing HTMLEditorField's schema data (minor coupling) -> call it `augmentSchemaData` and pass the existing array by reference for modification

NightJar: Thinking about the way it's currently structured...
1) would require another level of data e.g. `this.props.data.config.attributes['data-whatever']` to keep things nicely separated.
2) sounds like a reasonably bad idea, unless it was a mixture of getting an independent array, then recursively merging the two results in some fashion (i.e. perhaps keeping the Field's data as immutable in a similar way to `getSchemaData` on `FormField`)

NightJar: `data` key on schema is freeform anyway, so I just left everything operating as it was (a true refactor).

src/Forms/HTMLEditor/HTMLEditorField.php
```diff
@@ -185,4 +190,12 @@ public function Field($properties = array())
         $this->getEditorConfig()->init();
         return parent::Field($properties);
     }
+
+    public function getSchemaStateDefaults()
+    {
+        $stateDefaults = parent::getSchemaStateDefaults();
+        $config = $this->getEditorConfig();
+        $stateDefaults['data'] = $config->getConfigSchemaData();
```
robbieaverill: May as well just write `$stateDefaults['data'] = $this->getEditorConfig()->getConfigSchemaData()`

Issue #7932, opened 2018-03-14
htaccess in "resources" folder denies page URL named "resources"


cwchong:
## Affected Version
4.01

## Description

After installation, the "resources/.htaccess" file contains rules that rewrites all requests that does not match a file to 404.
This will mean that the website cannot contain a urlsegment like "resources" e.g. http://website/resources

Propose to remove that 404 rewrite and let the system handle 404s instead of via htaccess

## Tasks

 * [x] Agree on new name
 * [ ] ~~Auto rename folder transparently (via vendor-expose? no need to opt-in)~~
 * [x] Investigate feasibility on customising folder name (separate out into new card?)
 * [x] Upgrade notes (call out static publishing, caches and any htaccess)
 * [x] Rewrite rule for htaccess to redirect if file exists
 * [x] Check with infrastructure about SSP compatibility
 * [x] Update dev docs

## Pull requests
* [x] ~~https://github.com/silverstripe/silverstripe-framework/pull/8364 (SS5 only)~~
* [x] ~~https://github.com/silverstripe/vendor-plugin/pull/25 (SS5 only)~~
* [x] https://github.com/silverstripe/vendor-plugin/pull/26
* [x] https://github.com/silverstripe/silverstripe-framework/pull/8519
* [x] https://github.com/silverstripe/silverstripe-installer/pull/242
* [x] https://github.com/silverstripe/recipe-core/pull/33
* [x] https://github.com/silverstripe/vendor-plugin-helper/pull/5
* [x] https://github.com/silverstripe/vendor-plugin/pull/27
* [x] https://github.com/silverstripe/silverstripe-cms/pull/2353
* [x] https://github.com/silverstripe/recipe-cms/pull/22

tractorcow:
Unfortunately, SiteTree::getExcludedURLSegments() intentionally excludes any real folders in the public webroot. It's not currently possible to have a page named 'resources' at the top level.

You can modify that .htaccess directly (add it to source control) and modify it if you need, but we'll be unlikely to make this a core change.

tractorcow:
In 5.0 maybe we can rename to something like _resources

chillu:
Also: Ideally that’s configurable, to allow people to mask their install. Any generic name we choose will clash, and non-generic name will be a marker that you’re running an SS install. Might need to be in the projects `composer.json`?

tractorcow:
[...]

PR #8364, opened 2018-09-07
API Rename resources to _resources

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/7932

Requires https://github.com/silverstripe/vendor-plugin/pull/25

Also all other modules will need to have their vendor-plugin dependency bumped to ^2 as well. My suggestion is to get the module merged first and migrate the modules one at a time.

### DIFF
--- a/src/Dev/Install/Installer.php
+++ b/src/Dev/Install/Installer.php
@@ -55,8 +55,8 @@ class Installer
     protected function installHeader()
     {
         $clientPath = PUBLIC_DIR
-            ? 'resources/vendor/silverstripe/framework/src/Dev/Install/client'
-            : 'resources/silverstripe/framework/src/Dev/Install/client';
+            ? '_resources/vendor/silverstripe/framework/src/Dev/Install/client'
+            : '_resources/silverstripe/framework/src/Dev/Install/client';
         ?>
         <html>
         <head>
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,7 @@
         "ext-session": "*",
         "psr/container-implementation": "1.0.0",
         "psr/container": "1.0.0",
-        "silverstripe/vendor-plugin": "^1.0"
+        "silverstripe/vendor-plugin": "^2.0"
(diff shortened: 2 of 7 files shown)


tractorcow:
To discuss:
 - 4.x or 5.x?
 - soft-code or hard-code?
 - _resources as a default or something else?

src/Dev/Install/Installer.php
```diff
@@ -55,8 +55,8 @@ protected function error($message = null)
     protected function installHeader()
     {
         $clientPath = PUBLIC_DIR
-            ? 'resources/vendor/silverstripe/framework/src/Dev/Install/client'
-            : 'resources/silverstripe/framework/src/Dev/Install/client';
+            ? '_resources/vendor/silverstripe/framework/src/Dev/Install/client'
```
ScopeyNZ: Why not use the constant for these bits?

ScopeyNZ:
> 4.x or 5.x?

I don't see how this wouldn't be a breaking API change :( . 

> soft-code or hard-code?

+1 for making this name configurable in 5.x. Can we do some composer.json configuration to make that work?

robbieaverill:
I tend to agree with @ScopeyNZ. This will be very disruptive to our module ecosystem. Can we not do this but just make it configurable at the project level from SS5 on?

src/Dev/Install/Installer.php
tractorcow: Laziness.

tractorcow:
Looks like we'll need to soft-code it. If it's not configured, it'll need to default to `resources` (per the constant), but we will probably need to do it in 5.x if existing 4.x code assumes the hard-coded value.

maxime-rainville:
https://github.com/silverstripe/silverstripe-framework/pull/8364 should supersede this PR I think. Unless someone has objections, I'll close this one.

sminnee:
Closing in favour of #8519

Issue #8366, opened 2018-09-08
Allow BASE_PATH to be set using en environment variable


jmfederico:
## Affected Version

"4.2.1"

## Description

Currently there is no reliable way to set BASE_PATH when for some reason SilverStripe is not able to detect it automatically. This might happen for instance when COMPOSER_VENDOR_DIR is not "vendor" (see #8077).

Setting it in index.php is a partial solution, but it won't work for CLI usage (sake).

A solution would be to allow it to be set via a Environment Variable, the same way BASE_URL is being set.

Is this something you would approve? If so, I could work on a PR.

tractorcow:
I disagree with allowing it to be configured. It should be auto-detected. Having to manually set it demonstrates the code is wrong if it doesn't detect it correctly. convention > configuration.

Issue #8077, opened 2018-05-18
custom COMPOSER_VENDOR_DIR breaks SilverStripe


jmfederico:
## Affected Version

silverstripe-themes/simple dev-master 4d546a4 The SilverStripe simple theme (default SilverStripe 3 theme)
silverstripe/recipe-cms    1.1.0              SilverStripe recipe for fully featured page and asset content editing
silverstripe/recipe-plugin 1.2.0              Helper plugin to install SilverStripe recipes

## Description

I am installing silverstripe using a custom `COMPOSER_VENDOR_DIR` and it breaks.
It seems like silverstripe expect `vendor` to be the `COMPOSER_VENDOR_DIR` and won't work otherwise.

Installation fails because it is not able to correctly detect the BASE_PATH, and from then on, everything is broken.

A workaround is to define "BASE_PATH" before initializing silverstripe (in install.php or index.php) and also create a symlink from `vendor` to the actual path of `COMPOSER_VENDOR_DIR`.

The symlink has to be created both for the `root` project and for `public/resources`.

## Steps to Reproduce

```
export COMPOSER_VENDOR_DIR=.dev_vendor
composer create-project silverstripe/installer test
cd test
php -S 0.0.0.0:8888
```

Now navigate to http://localhost:8888/public/install.php

robbieaverill:
Thanks for the report. You're right, we expect it to be "vendor", although other than the application bootstrapping this isn't mandated in many places. Would you like to submit a pull request to use the constant instead, assuming the constant has been around in composer for a while? :)

jmfederico:
Was about to do it, but it seems like it is not worth it. Pretty much every PHP package expects the directory to be "vendor".

Instead I was wondering if there is a way to set "BASE_PATH" before silverstripe is initialized, without modifying "index.php" and "install.php".

It seems like "index.php" is eventually overriden by SIlverStripe, so custom changes are lost.

I tried using "autoload > files" in composer.json, but those files are loaded AFTER silverstripe is initialized.

It seems like it should be possible, since "constants.php" checks if "BASE_PATH" is defined before defining it, but I can't find a way to do it.

robbieaverill:
>It seems like it should be possible, since "constants.php" checks if "BASE_PATH" is defined before defining it, but I can't find a way to do it.

[...]

PR #8367, opened 2018-09-08
Include custom bootstrap.php file if one exists.

jmfederico:
For background on why I created this read #8366 and #8077

I noticed that there is no way to set constants in SilverStripe before the constants.php file is loaded by autoload.php reliably. Altering index.php is not a good idea, since there are more entry points to SilverStripe (like [sake](https://docs.silverstripe.org/en/4/developer_guides/cli/#sake-silverstripe-make))

The simples solution with no backward compatibility issues that I could see is to load a custom file before loading autoload.php if it exists (bootstrap.php seemed like a good name to me).

This needs to be changed in two places. In this repo (which this PR does) and the in the [index.php](https://github.com/silverstripe/recipe-core/blob/4/public/index.php) file of [SilverStripe - Recipe Core](https://github.com/silverstripe/recipe-core/).

If you think this is useful, I will create a PR in [SilverStripe - Recipe Core](https://github.com/silverstripe/recipe-core/).

Also, documentation should be updated, but I am not sure where it should go.

### DIFF
--- a/src/Dev/Install/Installer.php
+++ b/src/Dev/Install/Installer.php
@@ -249,6 +249,12 @@ use SilverStripe\Control\HTTPRequestBuilder;
 use SilverStripe\Core\CoreKernel;
 use SilverStripe\Core\Startup\ErrorControlChainMiddleware;
 
+if (file_exists(__DIR__ . '/bootstrap.php')) {
+    require __DIR__ . '/bootstrap.php';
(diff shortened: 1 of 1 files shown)


jmfederico:
I do not know what was going on in my head, this PR does exactly what I said should not be done, altering index.php.

PR #8372, opened 2018-09-10
Removed SS_CONFIGSTATICMANIFEST

sb-relaxt-at:
Correct me if I'm wrong, but it seems as if the .env property `SS_CONFIGSTATICMANIFEST` is not used.


sb-relaxt-at:
You are welcome! Thanks for merging!

Issue #8379, opened 2018-09-17
Precache Hierarchy::numChildren() for performance


sminnee:
The CMS treeview makes heavy use of `Hierachy::numChildren()` and its caching feature should be extended to allow precaching in a similar manner to Versioned::prepopulate_versionnumber_cache.

There is a call to Hierarchy::numChildren() for every tree-item shown in the CMS. This accounts for the bulk of time calling treeview / getsubtree URL requests, at least in cases where https://github.com/silverstripe/silverstripe-versioned/pull/181 has been applied.

## Approach

 * Replace the cache variable in Hierarchy::numChildren with a static cache by stage, class and ID, like Versioned
 * Introduce a prepopulate_versionnumber_cache method based on Versioned::prepopulate_versionnumber_cache
 * Use a query similar to "SELECT ParentID AS ID, COUNT(*) AS NumChildren FROM SiteTree WHERE ParentID != ID" to cache the number of children in bulk.
 * Incorporate the exclusion filters of Hierarchy::stageChildren() into this query

## augmentStageChildren deprecation

It would be difficult to make this work with augmentStageChildren. I can't see many places where augmentStageChildren is used (it's not used in any of the supported modules, as far as I can see), and my recommendation is that we deprecate it in order to implement this performance features.

 * Detect if augmentStageChildren is used, If so, throw a warning that this disables the performance optimisation
 * Suppress the bulk prefetching of numChildren in this case

## CMS update

Once this change has been made, CMSMain::SiteTreeAsUL can make calls to `Hierarchy::prepopulate_numchildren_cache`

sminnee:
I'd be keen to get @tractorcow's thoughts on this.

sminnee:
Note that this might best be refactored as part of #8391.

sminnee:
I believe that we can close this one now, @maxime-rainville @ScopeyNZ?

ScopeyNZ:
I'll leave @maxime-rainville to confirm but I'm happy to close it.

maxime-rainville:
Yes ... we're all done here.

PR #8380, opened 2018-09-18
NEW: Add Hierarchy::prepopulate_numchildren_cache()

sminnee:
Adds the ability to prepopulate the cache for Hierarchy::numChildren()
in a batch.

This is still a WIP as it will give bad results if augmentStageChildren
is used in a way that alters the number of records.

This is an useful part of optimising the query count on tree generation.
See https://github.com/silverstripe/silverstripe-framework/issues/8379

# Parent issue
* silverstripe/silverstripe-cms#2250

### DIFF
--- a/src/ORM/Hierarchy/Hierarchy.php
+++ b/src/ORM/Hierarchy/Hierarchy.php
@@ -422,9 +422,12 @@ class Hierarchy extends DataExtension
         $hideFromHierarchy = $this->owner->config()->hide_from_hierarchy;
         $hideFromCMSTree = $this->owner->config()->hide_from_cms_tree;
         $baseClass = $this->owner->baseClass();
+        $baseTable = $this->owner->baseTable();
         $staged = DataObject::get($baseClass)->where(sprintf(
-            '%s <> %s',
+            '%s.%s <> %s.%s',
+            Convert::symbol2sql($baseTable),
             Convert::symbol2sql("ParentID"),
+            Convert::symbol2sql($baseTable),
             Convert::symbol2sql("ID")
         ));
 


sminnee:
OK there's an issue with this: it conflicts with the lumberjack module because the lumberjack module overrides stageChildren. This causes the issue that Max found.

I'm going to see if I can amend the prepopulation logic to use a call to stageChildren as the basis of its query.

sminnee:
Hrm, so I looked into this, and I wonder if we declare the problem to be with lumberjack: https://github.com/silverstripe/silverstripe-lumberjack/issues/86

sminnee:
You could for example do this instead:
```

**Comment by maxime-rainville** [comment]
Just tweaks Sam's logic to avoid counting the children of childless object when the cache is complete.

We're now getting a pretty solid 28% performance boost when rendering treeview. :sunglasses: 

image

**Comment by robbieaverill** [comment]
>You could for example do this instead:

The suspense is killing me

**Comment by maxime-rainville** [comment]
@robbieaverill Our key project will need to add this to their config:
```
SilverStripe\CMS\Model\SiteTree:
  hide_from_cms_tree:
  - App\Pages\RandomPageTypeOne
  - App\Pages\RandomPageTypeTwo
  - App\Pages\RandomPageTypeThree
```

**Comment by sminnee** [comment]
Your amendment looks good, Maxime. I think that there's one thing we want to note in the changelog:

`Take care with stageChildren() overrides. Hierarchy::numChildren() results will only make use of stageChildren() results that are applied to the base class and don't include record-specific behaviour.`

**Comment by sminnee** [comment]
@silverstripe/core-team can we get a sanity check on whether the change described in my most recent comment is okay for semver?

**Comment by chillu** [comment]
> @silverstripe/core-team can we get a sanity check on whether the change described in my most recent comment is okay for semver?

I've done a sanity check via Github for `stageChildren`, and it brings up a few use cases: https://github.com/search?l=PHP&q=%22function+stageChildren%22+dataobject&type=Code. Seems to be pretty much exclusively symbiote's various "connector" modules, and then a whole bunch of core code copied into other git repos.

The base module of this isn't even 4.x compatible (https://github.com/nyeholt/silverstripe-external-content), so I don't think it'll cause wider issues in practice - but looping in the maintainer @nyeholt in here as a courtesy either way.

**Comment by sminnee** [comment]
OK I've rebased this and added a note to the 4.3 changelog.

**Comment by maxime-rainville** [comment]
I've got the unit tests written, but there's an edge case that's being tripped that I didn't quite figure out how to resolve yet. I'll have another look at in the morning.

`prepopulate_numchildren_cache` doesn't honor the `hide_from_hierarchy` config.

There's also a bit of  an issue with unit testing the `hide_from_cms_tree` config flag because it only gets triggered when `Hierachy::showingCMSTree` returns true. This method is based on the current controller being `LeftAndMain` and the current action being one of "treeview", "listview", "getsubtree". FYI I'm very judgemental of this method!

File: src/ORM/Hierarchy/Hierarchy.php [hunk 1]
```diff
@@ -271,11 +291,18 @@ public function numHistoricalChildren()
      */
     public function numChildren($cache = true)
```
maxime-rainville: I'm wondering if we should add an option here to get `numChildren` for LIVE here? The way this is written right now, there's a strong assumption this will only be used for Draft stage.

`prepopulate_numchildren_cache` expect a `stage` parameter, but that doesn't actually affect which stage is pre-cached, just which key the cache is stored under ... which sound kind of useless since there's no way of accessing the cache unless you actually set that parameter to `Versioned::STAGE`.

tests/php/ORM/HierarchyCachingTest.php
```diff
@@ -0,0 +1,150 @@
...
+
+    public function cacheNumChildrenDataProvider()
+    {
+        return [
+            [TestObject::class, 'obj1', false, 0, 'childless object should have a numChildren of 0'],
+            [TestObject::class, 'obj1', true, 0, 'childless object should have a numChildren of 0 when cache'],
+            [TestObject::class, 'obj2', false, 2, 'Root object numChildren should count direct children'],
+            [TestObject::class, 'obj2', true, 2, 'Root object numChildren should count direct children when cache'],
+            [TestObject::class, 'obj3a', false, 2, 'Sub object numChildren should count direct children'],
+            [TestObject::class, 'obj3a', false, 2, 'Sub object numChildren should count direct children when cache'],
```
ScopeyNZ:
```diff
- false
+ true
```

sminnee:
@ScopeyNZ I agree with the concerns about the DB::replace_parameter implementation, but a couple of notes:

 * The implementation was copied from DB::inline_parameters(), rather than written from scratch.
 * The "real fix" to its necessity is probably an SS5 topic such as https://github.com/silverstripe/silverstripe-framework/issues/8397

src/ORM/Hierarchy/Hierarchy.php
sminnee: I've focused on preserving the existing behaviour; I got rid of the $stage parameter in my subsequent patch to this (the other PR); perhaps we should roll this in here?

sminnee:
> prepopulate_numchildren_cache doesn't honor the hide_from_hierarchy config.

This surprises me. Is the config applied to SiteTree or a subclass?

tests/php/ORM/HierarchyCachingTest.php
```diff
@@ -0,0 +1,150 @@
...
+    public function setUp()
+    {
+        parent::setUp();
+        TestObject::singleton()->flushCache();
+    }
+
+    public static function setUpBeforeClass()
+    {
+        parent::setUpBeforeClass();
+        HideTestObject::config()->update(
```
sminnee: Can you apply this to SiteTree instead?

sminnee:
@maxime-rainville do you think I should merge the #8395 changes into this one?

maxime-rainville:
Sorry, I bulked down and didn't see all the messages flying.

The problem was that some of the where statements in the loop where getting sent back as arrays and other were being sent back as `DataQuery_SubGroup`. The `DataQuery_SubGroup` were just being skipped altogether.

I just took some liberties with `prepopulate_numchildren_cache` logic. I added a flag to `stageChildren` that will suppress the ID/ParentID _where_ clauses when set to true. I'm using that instead of looping through all the where clause.

I think it makes the logic a bit easier to follow and a bit cleaner and the generate SQL statement looks legit.

```SQL
SELECT DISTINCT "ParentID", COUNT(DISTINCT "SiteTree"."ID") AS "NumChildren" FROM "SiteTree" LEFT JOIN "SiteTree_Localised" AS "SiteTree_Localised_en_NZ" ON "SiteTree"."ID" = "SiteTree_Localised_en_NZ"."RecordID" AND "SiteTree_Localised_en_NZ"."Locale" = ? WHERE (("SiteTree"."ClassName" NOT IN (?, ?, ?, ?) OR "SiteTree"."ClassName" IS NULL)) GROUP BY ParentID
```

However, it probably wouldn't hurt to have someone who actually knows what they're doing have another look at it.

On the topic https://github.com/silverstripe/silverstripe-framework/pull/8395, it probably would be better to fold it into this one, as rebasing it at this stage would probably be more trouble than its worth.

tests/php/ORM/HierarchyCachingTest.php
maxime-rainville: That's been fixed.

src/ORM/Hierarchy/Hierarchy.php
```diff
@@ -421,16 +394,23 @@ public function showingCMSTree()
      *
      * @param bool $showAll Include all of the elements, even those not shown in the menus. Only applicable when
      *                      extension is applied to {@link SiteTree}.
+     * @param bool $IDLess Set to true to supress the ParentID and ID where statements.
      * @return DataList
      */
-    public function stageChildren($showAll = false)
+    public function stageChildren($showAll = false, $IDLess = false)
```
sminnee: I would probably call this something like`$skipParentIDFilter`.

I would also note that this won't work if people have overridden stageChildren, but the original code only calls stageChildren on subclasses, so that's no worse than my implementation.

src/ORM/Hierarchy/Hierarchy.php
```diff
@@ -333,39 +333,13 @@ public static function prepopulate_numchildren_cache($baseClass, $stage, $idList
...
-                        } else {
-                            $paramNum++;
-                        }
-                    }
-
-                    $newWhere[] = [ $clause => $params ];
-                }
-            }
-        }
+        $query = $dummyObject->stageChildren(true, true)->dataQuery()->query();
```
sminnee: We can probably stop the implementation of DB::replace_parameter if we go with this, since it's a bit gross and was only added for this use-case.

sminnee: We need to add "Where ID != ParentID" as a filter here.

src/ORM/Hierarchy/Hierarchy.php
```diff
@@ -361,9 +361,7 @@ public static function prepopulate_numchildren_cache($baseClass, $stage, $idList
             '"ParentID"',
             "COUNT(DISTINCT $idColumn) AS \"NumChildren\"",
         ]);
-        $query->setGroupBy([
-            "ParentID
-        "]);
+        $query->setGroupBy(['"ParentID"']);
```
sminnee: It's probably best to use `Convert::symbol2sql("ParentID")` here.

sminnee:
@maxime-rainville I've left a few comments — do you want to address those, then I'll rebase/squash and maybe we can get Ingo to do the final merge?

tests/php/ORM/HierarchyCachingTest.php
```diff
@@ -0,0 +1,150 @@
...
+        $baseClass,
+        $stage,
+        $idList,
+        $className,
+        $identifier,
+        $cache,
+        $expected,
+        $message
+    ) {
+        Hierarchy::prepopulate_numchildren_cache($baseClass, $stage, $idList);
```
sminnee: We may want to test this instead, due to my most recent addition:

```
TestObject::singleton()->prepopulateTreeDataCache($idLIst, [ 'numChildrenMethod' => 'numChildren' ]);
```

Compare to how CMS is calling things in the relevant PR here: https://github.com/silverstripe/silverstripe-cms/pull/2266/files

src/ORM/Hierarchy/Hierarchy.php
maxime-rainville: Done.

tests/php/ORM/HierarchyCachingTest.php
maxime-rainville: Done.

src/ORM/Hierarchy/Hierarchy.php
maxime-rainville: I've added the `Where ID != ParentID` statement to stageChildren.

src/ORM/Hierarchy/Hierarchy.php
maxime-rainville: Done.

maxime-rainville:
All done.

tests/php/ORM/HierarchyCachingTest.php
maxime-rainville: I'd rather note as it ought to work with this and that's how the other Hierarchy test are using.

PR #8387, opened 2018-09-19
FIX: Ensure existing session can be accessed if headers_sent()

sminnee:
If a session already exists, and Session::start() isn’t called until
after a large enough block of content is output, then headers_sent()
will be false. The previous code prevented the session from being
started in this case. That might makes sense for the creation of a new
session, but it prevent legitimate access to an existing session.

This mostly manifested when running debugging tools such as showqueries,
which may output content before the session is started.

I noticed this when implementing ?showqueries=backtrace, which showed
a backtrace of each query, pushing the pre-session output above the output
threshold.

### DIFF
--- a/src/Control/Session.php
+++ b/src/Control/Session.php
@@ -298,11 +298,20 @@ class Session
             }
         }
 
-        if (!session_id() && !headers_sent()) {
-            if ($domain) {
-                session_set_cookie_params($timeout, $path, $domain, $secure, true);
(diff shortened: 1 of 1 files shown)


sminnee:
This needs to be tweaked: if `headers_sent()` then we need to `session_start()` only in cases where the session cookie already exists.

sminnee:
OK that's sorted, should be good for review now.

Issue #8389, opened 2018-09-19
Performance issues with BasicAuth and large number of LoginAttempts


caffeineinc:
## Affected Version

```
silverstripe/framework                                   4.2.1                              The SilverStripe framework
```

## Description
Two functions interact with the LoginAttempt object which when used in conjunction with BasicAuth result in significant performance degradation over time. (30-50sec page load times)

*Note:* this will only affect you if you're using email addresses and BasicAuth combined along with a large number of active users.

Basic auth calls `Member::isLockedOut` to determine if the member has been naughty and the last 5 attempts were unsuccessful. This results in the following query which due to the WHERE clause cannot be indexed: (OR) 

```
SELECT DISTINCT LoginAttempt.ClassName, LoginAttempt.LastEdited, LoginAttempt.Created, LoginAttempt.Email, LoginAttempt.EmailHashed, LoginAttempt.Status, LoginAttempt.IP, LoginAttempt.MemberID, LoginAttempt.ID, 
			CASE WHEN LoginAttempt.ClassName IS NOT NULL THEN LoginAttempt.ClassName
			ELSE 'SilverStripe\\Security\\LoginAttempt' END AS RecordClassName
 FROM LoginAttempt
 WHERE ((LoginAttempt.Email = "email")
 OR (LoginAttempt.EmailHashed = "email sha1 hash"))
 ORDER BY LoginAttempt.Created DESC
 LIMIT 5
```
This query is not optimised, and results in a full table scan and file sort (due to the order by). This becomes more of a problem when BasicAuth is enabled, as every request results in LoginAttempt object, which means the table can get _very_ large very quickly. 

We can resolve some of the performance issues by adding an index for email hashed, but because the query uses two filter clauses, it can't use any indexes we add :( 

`Member::getByEmail()`  We don't seem to use Email in this class (it's always null) and the setter prevents it's use. 

### `MemberAuthenticator::recordLoginAttempt()`
Also becomes slow due to the size of the table and lack of indexes. 

## Steps to Reproduce
this will only affect you if you're using email addresses and BasicAuth combined along with a large number of active users (~2 million rows in LoginAttempt) 

## Resolution
Add indexes across  `LoginAttempt->emailHash` and remove email from the filter clause on `Member::getByEmail()`. The query then uses the index and doesn't have a performance hit.

caffeineinc:
[...]

PR #8390, opened 2018-09-19
Performance issues with BasicAuth and LoginAttempts

caffeineinc:
Two functions interact with the LoginAttempt object which when used in conjunction with BasicAuth result in significant performance degradation over time, as the LoginAttempts Table fills.

This fix adds an index to the lookup column EmailHashed and removes the Email filter part of getByEmail() so it can use the index resulting in a much faster query.

For more information see https://github.com/silverstripe/silverstripe-framework/issues/8389


src/Security/LoginAttempt.php
```diff
@@ -36,7 +36,6 @@ class LoginAttempt extends DataObject
     const FAILURE = 'Failure';
 
     private static $db = array(
-        'Email' => 'Varchar(255)', // Remove in 5.0
```
robbieaverill: I suspect this is a semver break - can you keep it there but remove it from the filter?

robbieaverill:
Originally introduced in f1dd3d6f03eb1d94c29c495994a1da9176a758d9, it looks like the Email field was left in the filter (but semi marked as deprecated) to preserve some B/C in the database table. I think we can safely remove the implementation of the column in the filter but I think we need to leave it there so it's semver safe

caffeineinc:
Sweet, i've put the email references back.

ScopeyNZ:
It seems like `Email` was in the filter for backwards compat. Is that going to be an issue if this is merged?

robbieaverill:
Maybe, but it’s been around (as B/C) for long enough that hopefully not. The feature is also disabled by default too.

caffeineinc:
The other thing @ScopeyNZ is that `setEmail()` is overloaded so you can't actually provide a value for Email anyway. 

https://github.com/silverstripe/silverstripe-framework/pull/8390/commits/c269a987d5305bd5efc20a15a15b2f7867840dd4#diff-41b85a1c8dd9a73784e8b33811f10130R77

ScopeyNZ:
Yeah, the issue would be with people who have an attempts for an email set before the original patch. I agree with @robbieaverill that it's been around for long enough to not be an issue (probably). I'd merge but I have no merge access 😉

PR #8394, opened 2018-09-21
BUG Prevent error on valid response status codes

tractorcow:
There are a LOT of missing http response codes in this class. I've added the missing ones (using https://en.wikipedia.org/wiki/List_of_HTTP_status_codes as a reference) and added a few unit tests.


src/Control/HTTPResponse.php
```diff
@@ -53,14 +59,27 @@ class HTTPResponse
         415 => 'Unsupported Media Type',
         416 => 'Request Range Not Satisfiable',
         417 => 'Expectation Failed',
+        418 => 'I\'m a Teapot',
```
kinglozzer: The most important status code of all

tractorcow:
Awesome thanks.

PR #8395, opened 2018-09-21
Refactor prepopulation

### LINKED ISSUE #8391: Refactor tree-generation prepopulation
## Problem

Right now `CMSMain::SiteTreeAsUL()` calls some Versioned prepopulation in order to generate the site tree without an egregious number of queries. This works okay, but has a couple of challenges:

 * It has no extension point. Prepopulation of additional data would require manual adjustments to this method. This is a little awkward but workable for core changes (e.g. https://github.com/silverstripe/silverstripe-cms/pull/2266) but it's a showstopper for add-on modules that require prepopulation (e.g. https://github.com/tractorcow/silverstripe-fluent/issues/467)
 * There's no way of repurposing it for other uses, such as treedropdownfield.

This relates to https://github.com/silverstripe/silverstripe-cms/issues/2250.

## Recommended fix

This logic should be wrapped into a prepopulateTreeDataCache method:

`singleton('SiteTree')->prepopulateTreeDataCache(DataList $recordList = null, array $options)`

 * $recordList: An optional DataList of the records to include in the prepopulation. By default, all records will be included.
 * $options: A map of metadata to assist. None of this should be assumed to be supplied, but it can be used to optimise any prepopulation, where appropriate.
    * 'childrenMethod': The name of the method on the record used to get children
    * 'numChildrenMethod': The name of the method on the record used to get number of children

`Hierarchy` can provide this method, since it relates to trees, and other extensions can hook into the behaviour by defining `onPrepopulateTreeDataCache`. The existing behaviour can be shifted to Versioned::onPrepopulateTreeDataCache, and the fluent ticket above could be implemented by providing this.

sminnee:
NEW: Add Hierarchy::prepopulateTreeDataCache()

This provides a more extensible way of preopulating caches for optimised
tree generation.

Fixes #8391

### DIFF
--- a/src/ORM/DB.php
+++ b/src/ORM/DB.php
@@ -430,6 +430,69 @@ class DB
         return $joined;
     }
 
+
+    /**
+     * Replace 1 parameter with the given value
+     * For example, you can use this to replace a parameter with a column name instead of a literal
+     * @param string $sql The parameterised query
+     * @param int $paramIdx The 0-based position of the parameter
+     * @param array $replacement The value to insert into the queyr
+     * @param bool $skipEscaping Set to true to insert the value as-is, with no escaping. Use with caution!
+     *
+     * @return string
+     */
+    public static function replace_parameter($sql, $paramIdx, $replacement, $skipEscaping = false)
+    {
+        $segments = preg_split('/\?/', $sql);
+        $joined = '';
+        $inString = false;
+        $numSegments = count($segments);
+        $currentParamIdx = 0;
+
+        for ($i = 0; $i < $numSegments; $i++) {
+            $input = $segments[$i];
+            // Append next segment
+            $joined .= $segments[$i];
+            // Don't add placeholder after last segment
+            if ($i === $numSegments - 1) {
+                break;
+            }
+            // check string escape on previous fragment
+            // Remove escaped backslashes, count them!
+            $input = preg_replace('/\\\\\\\\/', '', $input);
+            // Count quotes
+            $totalQuotes = substr_count($input, "'"); // Includes double quote escaped quotes
+            $escapedQuotes = substr_count($input, "\\'");
+            if ((($totalQuotes - $escapedQuotes) % 2) !== 0) {
+                $inString = !$inString;
+            }
+
+            // Append placeholder replacement
+            if ($inString) {
+                // Literal question mark
+                $joined .= '?';
+                continue;
+            }
+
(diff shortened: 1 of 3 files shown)


src/ORM/DB.php
```diff
@@ -430,6 +430,69 @@ public static function inline_parameters($sql, $parameters)
...
+     * @param string $sql The parameterised query
+     * @param int $paramIdx The 0-based position of the parameter
+     * @param array $replacement The value to insert into the queyr
+     * @param bool $skipEscaping Set to true to insert the value as-is, with no escaping. Use with caution!
+     *
+     * @return string
+     */
+    public static function replace_parameter($sql, $paramIdx, $replacement, $skipEscaping = false)
+    {
+        $segments = preg_split('/\?/', $sql);
```
ScopeyNZ: `explode` would be faster here.

src/ORM/DB.php
```diff
@@ -430,6 +430,69 @@ public static function inline_parameters($sql, $parameters)
...
+        $inString = false;
+        $numSegments = count($segments);
+        $currentParamIdx = 0;
+
+        for ($i = 0; $i < $numSegments; $i++) {
+            $input = $segments[$i];
+            // Append next segment
+            $joined .= $segments[$i];
+            // Don't add placeholder after last segment
+            if ($i === $numSegments - 1) {
```
ScopeyNZ: You could adjust the loop condition to `$i < $numSegments - 1`

maxime-rainville:
@ScopeyNZ Parts of this PR are built on top of https://github.com/silverstripe/silverstripe-framework/pull/8380. When #8380 has been merged I'll have to rebase this one against it.

I'll take your feedback and implement it into the other one.

src/ORM/DB.php
maxime-rainville: Won't we miss the last segment if we do that?

ScopeyNZ: Yep, ignore me 😂

maxime-rainville: That's what I do most of the time any way :wink:

src/ORM/DB.php
maxime-rainville: Addressed on parent PR by https://github.com/silverstripe/silverstripe-framework/pull/8380/commits/07f9cdc262769a0e8b9689203c645aab42d6e6b7

sminnee:
I've incorporated this into #8380

PR #8398, opened 2018-09-24
FIX Don't flush manifests in tests by default

chillu:
It's massively slowing down test execution,
particularly for rapid execution through test-driven-development (TDD)

Regression since 4.0.0-beta1, see https://github.com/silverstripe/silverstripe-framework/commit/3873e4ba008cfc2af7e26ca86665affc289cd677#diff-e2f3520673dab1cd0127a469fe3a2a29R925

It replaced prior behaviour which respected the flush parameter: https://github.com/silverstripe/silverstripe-framework/commit/3873e4ba008cfc2af7e26ca86665affc289cd677#diff-e2f3520673dab1cd0127a469fe3a2a29R925

You could argue this is an API change, but I think it's not going to affect anyone negatively in production. CI based tests are quite unlikely to change files or YAML config between test runs which would require a manifest rebuild. And developers executing tests should follow our existing documentation which always said to use `flush` between test runs when you change stuff.


chillu:
Two core committer approvals, can one of you merge please? Damian has hundreds of notifications waiting, I'm keen not to make him an blocker unless it's really necessary (e.g. nobody else understands the code heh)

PR #8399, opened 2018-09-24
var_export does not handle circular references

lekoala:
this often happens in SilverStripe, a plain print_r is much safer


sminnee:
What about var_dump?

ScopeyNZ:
> What about var_dump?

I assume the choice is between `var_export` and `print_r` because they both can return (with `true` as the second param) 🙂

PR #8403, opened 2018-09-25
NEW: Add test for PHP 7.3 support

sminnee:
Just pushing this PR as an experiment, to see where the travis-CI support is at, and also to see whether 3.x-dev already supports it. With 3.x support likely in some form until late 2019 / early 2020, being able to run it on PHP 7.3 would be handy if feasible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,22 +8,50 @@ addons:
   apt:
     packages:
       - tidy
-  firefox: "31.0"
 
 matrix:
   include:
+    # Framework tests
     - php: 5.4
-      env: DB=MYSQL PDO=1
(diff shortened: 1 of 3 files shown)


robbieaverill:
We could put this in allowed failures and merge it. 7.3 support tracker: https://github.com/travis-ci/travis-ci/issues/9717

sminnee:
IMO allowed_failures have very little value and simply burn through our build quota. I'll close the PR and reopen it once that travis ticket is closed.

robbieaverill:
As with #8432 I think this is good to merge since PHP 7.3 is in RC now we shouldn't expect any breaking changes for new features to be introduced

PR #8405, opened 2018-09-25
Fix switching to recipes

mspacemedia:
change silverstripe/framework to silverstripe/recipe-cms


dhensby:
Thanks for this @mspacemedia but it doesn't appear to be a correct change; please can you provide a bit more information / justification about why you think this should be changed?

mspacemedia:
My mistake - please ignore ;)

dhensby:
No problem, thanks for taking the time to open a PR

PR #8408, opened 2018-09-25
Test cleanup

dhensby:
I'm going to spend some time going through tests and cleaning them up. As such this is a living PR.

If the builds are green and you like what you see, merge at will and I'll continue to work on it and open new PRs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/assets/
 .sass-cache
 .DS_Store
 npm-debug.log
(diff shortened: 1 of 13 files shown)


assets/.htaccess
```diff
@@ -0,0 +1,35 @@
...
+    # Turn the PHP engine off
+    php_flag engine off
+</IfModule>
+
+<IfModule mod_rewrite.c>
+    <IfModule mod_env.c>
+        SetEnv HTTP_MOD_REWRITE On
+    </IfModule>
+
+    RewriteEngine On
```
robbieaverill: These files didn't exist before, do they need to be there?

dhensby: whoops - maybe assets should be ignored?

PR #8409, opened 2018-09-26
MINOR Add caching to the TreeDropdownField tree action

maxime-rainville:
This is more of an experimental PR. I'm not actually sure it's a plus.

Basically we apply some of the same logic that got introduced by https://github.com/silverstripe/silverstripe-framework/pull/8380 to the `TreeDropdownField`.

My preliminary profiling on my local box seems to indicate its faster. It's speed things up by 12%. 

It increases the time spent waiting for SQL query results, but reduces the number of query ... basically we are running a smaller number of more intensive queries. My guess is that if you a using a remote DB Server on SSP your actual waiting for queries time will go down as well because of latency.

image

I'll try to run some test on SSP and on a site that does not have that has a smaller SiteTree.

### DIFF
--- a/src/Forms/TreeDropdownField.php
+++ b/src/Forms/TreeDropdownField.php
@@ -453,6 +453,8 @@ class TreeDropdownField extends FormField
         /** @var DataObject|Hierarchy $obj */
         $obj = null;
         $sourceObject = $this->getSourceObject();
+
+        Hierarchy::prepopulate_numchildren_cache($sourceObject);
         if ($id && !$request->requestVar('forceFullTree')) {
             $obj = DataObject::get_by_id($sourceObject, $id);
             $isSubTree = true;
--- a/src/Forms/TreeDropdownField.php
+++ b/src/Forms/TreeDropdownField.php
@@ -454,7 +454,13 @@ class TreeDropdownField extends FormField
         $obj = null;
         $sourceObject = $this->getSourceObject();
 
-        Hierarchy::prepopulate_numchildren_cache($sourceObject);
+        // Precache numChildren count if possible.
+        if ($this->getNumChildrenMethod() == 'numChildren') {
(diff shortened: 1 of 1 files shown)


maxime-rainville:
FYI I didn't used the `Hierarchy::prepopulateTreeDataCache()`, because adding the Versioned caching on top the Hierarchy caching actually seemed to slow things down.

ScopeyNZ:
> FYI I didn't used the `Hierarchy::prepopulateTreeDataCache()`, because adding the Versioned caching on top the Hierarchy caching actually seemed to slow things down.

It would probably be good to add that as a comment in-line 🙂

maxime-rainville:
I did some more local testing. I've having problem benchmarking this in SSP right now.

*Vanilla SS with small number of pages*
image

*Vanilla SS site with thousands of pages*
image

*CWP Kitchen wih thousands of pages*
image

maxime-rainville:
I added a in-code comment as requested by @ScopeyNZ.

src/Forms/TreeDropdownField.php
```diff
@@ -453,6 +453,14 @@ public function tree(HTTPRequest $request)
         /** @var DataObject|Hierarchy $obj */
         $obj = null;
         $sourceObject = $this->getSourceObject();
+
+        // Precache numChildren count if possible.
+        if ($this->getNumChildrenMethod() == 'numChildren') {
+            // We're not calling `Hierarchy::prepopulateTreeDataCache()` because we're not customising results based
+            // on version or Fluent locales. So there would be no performance gain from additional caching.
+            Hierarchy::prepopulate_numchildren_cache($sourceObject);
```
ScopeyNZ: Seems fine... It'd be nice having `prepopulatedTreeDataCache` only because that provides extension points for performance enhancements in the future but I understand that would probably make it slower in this case 😧 .

PR #8413, opened 2018-09-27
NEW Show file path on PHP parser exceptions

chillu:
How did it take so long to fix this? How did developers put up with now knowing *in which file* their parse error happened, and only get the line.

And to top it off, they get a misleading message by having them pointed at a PHP parser file instead.

This must've mystified hundreds of devs over the last two years (https://github.com/silverstripe/silverstripe-framework/commit/cdb4a86e1c8fe4a0b01ddff4fb3f32c2568685e8)

We need to get better at real world testing of our APIs :(

Before:

```
 Fatal error: Uncaught PhpParser\Error: Syntax error, unexpected T_STRING, expecting '{' on line 9 in /var/www/mysite/www/vendor/nikic/php-parser/lib/PhpParser/ParserAbstract.php on line 313

```

After:

```
 Fatal error: Uncaught PhpParser\Error: Syntax error, unexpected T_STRING, expecting '{' in /var/www/mysite/www/app/src/Page.php on line 9 in /var/www/mysite/www/vendor/nikic/php-parser/lib/PhpParser/ParserAbstract.php on line 313
 ```

### DIFF
--- a/src/Core/Manifest/ClassManifestErrorHandler.php
+++ b/src/Core/Manifest/ClassManifestErrorHandler.php
@@ -1,0 +1,33 @@
+<?php
+namespace SilverStripe\Core\Manifest;
+
+use PhpParser\Error;
+use PhpParser\ErrorHandler;
+
+/**
+
+ * Error handler which throws, but retains the original path context.
+ * For parsing errors, this is essential information to identify the issue.
+ */
+class ClassManifestErrorHandler implements ErrorHandler
+{
+    /**
+     * @var String
+     */
+    protected $pathname;
+
+    /**
+     * @param String $pathname
+     */
+    public function __construct($pathname)
+    {
+        $this->pathname = $pathname;
+    }
+
+    public function handleError(Error $error)
+    {
+        $newMessage = sprintf('%s in %s', $error->getRawMessage(), $this->pathname);
+        $error->setRawMessage($newMessage);
+        throw $error;
+    }
+}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/Core/Manifest/ClassManifestErrorHandler.php [hunk 1]
```diff
@@ -0,0 +1,33 @@
+<?php declare(strict_types=1);
```
ScopeyNZ: This is PHP 7 only? I believe this'll throw a warning in 5.6.

src/Core/Manifest/ClassManifestErrorHandlerTest.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+use SilverStripe\Dev\SapphireTest;
+use SilverStripe\Core\Manifest\ClassManifestErrorHandler;
+use PhpParser\Error;
+
+class ClassManifestErrorHandlerTest extends SapphireTest
```
kinglozzer: PHPCS is pointing out this class is missing a namespace

chillu: Fixed, thanks PHPCS! :)

src/Core/Manifest/ClassManifestErrorHandler.php
chillu: Derp derp, damn you copy paste. Fixed.

PR #8416, opened 2018-09-27
raw2att accepts array, but sprintf doesn't

lekoala:
it's not very likely to happen (it did in my case :-) ) but if the value is an array, sprintf will fail (because raw2att accepts array, but sprintf doesn't). i suggest to json encode any array data to ensure it's safely included in the html. Or we should throw proper exceptions to make sure invalid values do not result in a php error.

### DIFF
--- a/src/Forms/FormField.php
+++ b/src/Forms/FormField.php
@@ -730,7 +730,11 @@ class FormField extends RequestHandler
             if ($value === true) {
                 $parts[] = sprintf('%s="%s"', $name, $name);
             } else {
-                $parts[] = sprintf('%s="%s"', $name, Convert::raw2att($value));
+                $strValue = Convert::raw2att($value);
+                if (!is_string($strValue)) {
+                    $strValue = json_encode($strValue);
+                }
+                $parts[] = sprintf('%s="%s"', $name, $strValue);
             }
         }
 


src/Forms/FormField.php
```diff
@@ -730,7 +730,11 @@ public function getAttributesHTML($attributes = null)
             if ($value === true) {
                 $parts[] = sprintf('%s="%s"', $name, $name);
             } else {
-                $parts[] = sprintf('%s="%s"', $name, Convert::raw2att($value));
+                $strValue = Convert::raw2att($value);
+                if (!is_string($strValue)) {
+                    $strValue = json_encode($strValue);
```
robbieaverill: We could use `Convert::raw2json()` here

lekoala: what is the added value of calling raw2json since it does the same thing? If I scan the codebase, I see a lot of direct calls to json_encode already.

robbieaverill: Consistency of dogfooding our API really, otherwise what's the point in having it 😄

sminnee:
Note the recommended deprecation here: https://github.com/silverstripe/silverstripe-framework/issues/8424

PR #8417, opened 2018-09-27
Add more list sorting tests and fix many many through sorting

### LINKED ISSUE #7981: SS4: Cannot sort `many_many` result when using `through`
Unsure if this is expected behaviour, or missing functionality. Please let me know :)

## Setup

I have a class (`Locale`) that has a `many_many` relationship to itself. This is being run `through` a class called `FallbackLocale`.

`Locale`:
```php
class Locale extends DataObject
{
    private static $has_many = [
        'FallbackLocales' => FallbackLocale::class . '.Parent',
    ];

    private static $many_many = [
        'Fallbacks' => [
            'through' => FallbackLocale::class,
            'from' => 'Parent',
            'to' => 'Locale',
        ],
    ];
}
```

`FallbackLocale`:
```php
class FallbackLocale extends DataObject
{
    private static $has_one = [
        'Parent' => Locale::class,
        'Locale' => Locale::class,
    ];

    private static $db = [
        'Sort' => 'Int',
    ];

    private static $default_sort = 'Sort';
}
```

## Scenario

I create a new `Locale` through a very standard Model Admin with a Gridfield.

I add two `Fallback()` records, and I sort them.

Whenever I use `$locale->Fallbacks()` (or `$locale->Fallback()->sort('Sort')`), the records are always in `ID` order, not the Gridfield `Sort` order.

## Workaround

If I use `$locale->FallbackLocales()` instead, I get the correct Gridfield sorting - however, I am then required to go an extra step to get the `Object` I'm actually interested in.

EG:
```php
foreach ($locale->FallbackLocales() as $throughObject) {
    $throughObject->Locale();
}
```

Or: I need to add a join before I sort.

## Pull requests

- [ ] #8417

robbieaverill:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/7981

This separates the sorting assertions out of an existing test into its own method, and adds another sorting test that doesn't use table aliases in the sort field.

I've also added some tests for sorting HasManyLists

---

This now fixes #7981 - a ManyManyThroughList will now use its joinClass's default_sort if nothing else has been provided already

### DIFF
--- a/src/ORM/ManyManyThroughQueryManipulator.php
+++ b/src/ORM/ManyManyThroughQueryManipulator.php
@@ -3,6 +3,7 @@
 
 namespace SilverStripe\ORM;
 
+use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Dev\Deprecation;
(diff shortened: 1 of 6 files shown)


robbieaverill:
Self merging with @ScopeyNZ and @kinglozzer's approval

PR #8420, opened 2018-09-28
Add extension points for ModelAdmin search API

lukereative:
Resolves silverstripe/silverstripe-admin#662

**Dependent PR** https://github.com/silverstripe/silverstripe-admin/pull/665

### DIFF
--- a/src/Forms/GridField/GridFieldFilterHeader.php
+++ b/src/Forms/GridField/GridFieldFilterHeader.php
@@ -44,6 +44,26 @@ class GridFieldFilterHeader implements GridField_URLHandler, GridField_HTMLProvi
     public $useLegacyFilterHeader = false;
 
     /**
+     * @var \SilverStripe\ORM\Search\SearchContext
+     */
+    protected $searchContext = null;
+
+    /**
+     * @var Form
+     */
+    protected $searchForm = null;
+
+    /**
+     * @var callable
+     */
+    protected $updateSearchContext = null;
+
+    /**
+     * @var callable
+     */
+    protected $updateSearchForm = null;
+
+    /**
      * @inheritDoc
      */
     public function getURLHandlers($gridField)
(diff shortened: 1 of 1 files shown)


src/Forms/GridField/GridFieldFilterHeader.php
```diff
@@ -208,9 +234,15 @@ public function canFilterAnyColumns($gridField)
      */
     public function getSearchContext(GridField $gridField)
     {
-        $context = singleton($gridField->getModelClass())->getDefaultSearchContext();
+        if (!$this->searchContext) {
+            $this->searchContext = singleton($gridField->getModelClass())->getDefaultSearchContext();
+        }
 
-        return $context;
+        if ($this->updateSearchContextCallback) {
```
ScopeyNZ: I think it would be safer to move this condition into the prior one so this context is only likely to be called once when the context is first created.

lukereative: Yep, done

src/Forms/GridField/GridFieldFilterHeader.php
```diff
@@ -55,10 +75,16 @@ public function getURLHandlers($gridField)
...
-    public function __construct($useLegacy = false)
-    {
+    public function __construct(
+        $useLegacy = false,
+        callable $updateSearchContext = null,
+        callable $updateSearchForm = null
+    ) {
         $this->useLegacyFilterHeader = $useLegacy;
+        $this->updateSearchContextCallback = $updateSearchContext;
+        $this->updateSearchFormCallback = $updateSearchForm;
```
robbieaverill: I think it'd be good to see getters and setters for these props, then use them in this class

ScopeyNZ: I _think_ they were omitted because they're intended to be removed in 5.0. Best to reduce the API surface considering that?

robbieaverill: If that's the case we should mark them as deprecated

src/Forms/GridField/GridFieldFilterHeader.php
```diff
@@ -38,11 +38,33 @@ class GridFieldFilterHeader implements GridField_URLHandler, GridField_HTMLProvi
...
+    /**
+     * @var Form
+     */
+    protected $searchForm = null;
+
+    /**
+     * @var callable
+     * @deprecated 4.3.0:5.0.0 Will be removed in 5.0
+     */
+    protected $updateSearchContextCallback = null;
```
robbieaverill: These were incorrectly named, I've adjusted them

PR #8421, opened 2018-09-28
Update deprecation PHPDocs to be PSR-5 compliant

robbieaverill:
See: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#55-deprecated


robbieaverill:
Self merging with @kinglozzer's approval

PR #8423, opened 2018-09-28
FIX New members now receive the configured default locale, not the current locale

### LINKED ISSUE #8396: New Members receive the locale of the currently logged in user, instead of i18n::$default_locale
## Affected Version

Observed in SS 4.2.1

## Description

When a new `Member` is created, the Locale property is populated from `i18n::get_locale()`. This method returns the current locale if set, and falls back to the `i18n::$default_locale` config property. 

When a user logs in to the CMS, the current locale is set according to that user's locale (in `LeftAndMain::init()`).

This leads to new users inheriting the locale of the currently logged in CMS user, instead of the `i18n::$default_locale` property.

I don't think that the locale of the currently logged in member should influence the locale of new Members, especially if a developer has provided a default locale through yml config. The default locale should trump the locale of the current user.

## Steps to Reproduce

1. Note the locale of the currently logged in user (in Security section of CMS)
2. Set the default locale to something else through yml config and flush
```
SilverStripe\i18n\i18n:
  default_locale: 'en_GB'
```
3. Try to create a new Member in the CMS. The locale will be pre-selected to match the logged in user rather than the specified default.

## Why is this a problem

* I suspect CMS users don't pay a lot of attention to the locale field, so I think it's important that the default is respected as users may not correct a problem if present
* The workflow of setting up a new site may lead to en_US being the de-facto default locale, despite configuration otherwise*

*I discovered this issue because the New Member form always had en_US selected, even though I knew I'd set the default locale to en_GB. Once I dug in to how Members receive a locale I realised that my default administrator Member received an en_US locale during installation, before I changed the default locale through yml config. If I hadn't paid attention all new Members I created after that would have gotten the en_US locale too. (Note: Possibly the installer wizard asks you what locale to use for your default admin, but I use a parent .env file so skip this process.)

robbieaverill:
Fixes #8396

### DIFF
--- a/src/Security/Member.php
+++ b/src/Security/Member.php
@@ -268,7 +268,7 @@ class Member extends DataObject
     public function populateDefaults()
     {
         parent::populateDefaults();
-        $this->Locale = i18n::get_locale();
+        $this->Locale = i18n::config()->get('default_locale');
     }
 
     public function requireDefaultRecords()
@@ -929,7 +929,7 @@ class Member extends DataObject
 
         // save locale
         if (!$this->Locale) {
-            $this->Locale = i18n::get_locale();
+            $this->Locale = i18n::config()->get('default_locale');
         }
 
         parent::onBeforeWrite();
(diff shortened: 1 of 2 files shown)


robbieaverill:
Self merging with @kinglozzer's approval

PR #8425, opened 2018-10-01
ENHANCEMENT: Prevent lazy loading on setComponent

unclecheese:
The call to `setField` for the foreign key triggers a lazy load of the component. If you're setting a component explicitly, this query is unnecessary, from what I can see.

### DIFF
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -1755,6 +1755,9 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 
             // Update local ID
             $joinField = $componentName . 'ID';
+            if ($item) {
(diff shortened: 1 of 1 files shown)


tractorcow:
This breaks change detection; isChanged('RelationID') won't always report true if it doesn't know the DB value prior to being replaced. It is necessary in this case.

kinglozzer:
Just a heads up: you pushed your branch to the main repo instead of a fork 😉

sminnee:
@unclecheese I suspect that you've done this as part of your eager loading work. If so, the path that I would go down is to let the eager loading join to more subclass tables in the original query, so that lazy loading isn't required.

sminnee:
I'd say that pervasive eager loading is the best strategy to follow for query optimisation.

PR #8427, opened 2018-10-01
FIX: Append any fields that don’t match name in insertBefore/insertAfter

### LINKED ISSUE #1397: [2011-04-12] FieldSet::insertBefore() sometimes doesn't insert at all
created by: paradigmincarnate
created at: 2011-04-12
original ticket: http://open.silverstripe.org/ticket/6597

---

My code uses FieldSet::insertBefore() to add a field.

A third party changed their code so the other field no longer exists.

My field now isn't inserted at all.

FieldSet::insertBefore() takes 2 parameters. The first is a field to insert, the second is the name of another field before which to insert it.

If a second parameter is passed, and there's no existing field by that name, the first field will not be inserted.

I propose that FieldSet::insertBefore() will always add the field (either first or last) if the "existing field" doesn't exist.

In the alternative, an exception should be thrown to indicate that the "existing field" could not be found.

Any time insertBefore() is called, the developer expects that their field will be inserted. To silently not do so is unintuitive.

sminnee:
Previous behaviour was to return false, which has been described as
a confusing bug on https://github.com/silverstripe/silverstripe-framework/issues/1397
where the issue was identified.

Fixes #1397

### DIFF
--- a/src/Forms/FieldList.php
+++ b/src/Forms/FieldList.php
@@ -563,12 +563,14 @@ class FieldList extends ArrayList
 
     /**
      * Inserts a field before a particular field in a FieldList.
+     * Will traverse CompositeFields depth-first to find the maching $name, and insert before the first match
      *
      * @param string $name Name of the field to insert before
      * @param FormField $item The form field to insert
-     * @return FormField|false
+     * @param bool $appendIfMissing Append to the end of the list if $name isn't found
+     * @return FormField|false Field if it was successfully inserted, false if not inserted
      */
-    public function insertBefore($name, $item)
+    public function insertBefore($name, $item, $appendIfMissing = true)
     {
         // Backwards compatibility for order of arguments
         if ($name instanceof FormField) {
@@ -584,7 +586,7 @@ class FieldList extends ArrayList
                 array_splice($this->items, $i, 0, array($item));
                 return $item;
             } elseif ($child instanceof CompositeField) {
-                $ret = $child->insertBefore($name, $item);
+                $ret = $child->insertBefore($name, $item, false);
                 if ($ret) {
                     return $ret;
                 }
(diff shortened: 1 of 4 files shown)


src/Forms/FieldList.php
```diff
@@ -563,12 +563,14 @@ public function dataFieldByName($name)
 
     /**
      * Inserts a field before a particular field in a FieldList.
+     * Will traverse CompositeFields depth-frist to find the maching $name, and insert before the first match
      *
      * @param string $name Name of the field to insert before
      * @param FormField $item The form field to insert
-     * @return FormField|false
+     * @param bool $appendIfMissing Append to the end of th slit if $name isn't found
```
robbieaverill: There's some typos in this doc block

src/Forms/FieldList.php
```diff
@@ -584,25 +586,33 @@ public function insertBefore($name, $item)
...
     }
 
     /**
      * Inserts a field after a particular field in a FieldList.
+     * Will traverse CompositeFields depth-frist to find the maching $name, and insert after the first match
      *
      * @param string $name Name of the field to insert after
      * @param FormField $item The form field to insert
-     * @return FormField|false
+     * @param bool $appendIfMissing Append to the end of th slit if $name isn't found
```
robbieaverill: And this one ;)

src/Forms/FieldList.php
```diff
@@ -618,14 +628,21 @@ public function insertAfter($name, $item)
...
+                $ret = $child->insertAfter($name, $item, false);
                 if ($ret) {
                     return $ret;
                 }
             }
             $i++;
         }
 
+        // $name not found, append if needed
+        echo "$name not found, appendIfMissing = $appendIfMissing\n";
```
robbieaverill: Debug line?

sminnee:
> I kind of feel like we should just do this by default rather than have a flag for it. Devs could be safe and check whether the field exists before adding their own if they really didn't want this to happen. Thoughts?

You mean the appendIfMissing variable? The only reason I added this flag is so that I could set it to false on the nested call to CompositeField::insertBefore/After. Otherwise the logic is all wrong. The alternative would be to create a separate set of methods but that seems like a worse idea.

Make sense? I'd tidy my typos/echos.

sminnee:
Fixed with a squash.

PR #8429, opened 2018-10-01
FIX: Throw deprecation on limit=0

### LINKED ISSUE #2950: [ORM] SS_List->limit(0) returns all rows
currently:

```
 $list->limit(0) == $list->limit(SS_List->count())
```

however, if you ask me, when one tells a list to return 0 items, it should actually return 0 items.

the reason this happens can be found here:
- ArrayList: [ArrayList.php#L136](https://github.com/silverstripe/silverstripe-framework/blob/65d96e8d7ceb3a38af67b8dc6722e6dfd2a86643/model/ArrayList.php#L136)
- DataList: [SQLQuery.php#L483](https://github.com/silverstripe/silverstripe-framework/blob/65d96e8d7ceb3a38af67b8dc6722e6dfd2a86643/model/SQLQuery.php#L483) + [Database.php#L883](https://github.com/silverstripe/silverstripe-framework/blob/65d96e8d7ceb3a38af67b8dc6722e6dfd2a86643/model/Database.php#L883)  
  (you can see that 0 will be set directly to `$query->limit` in `SQLQuery` and then in `SS_Database` there is a `if ($query->getLimit)`)

When fixing this bug, we have to be careful though, `SS_List->limit(null)` should in fact return all rows, as this can be considered unsettling the limit.
Setting the limit to `null` is also the way that the `GridFieldExportButton` accomplishes exporting all rows instead of just the displayed ones.

related issue: #2949
related pull reuqest: #2294

sminnee:
It’s not clear what limit=0 should do; to clear the limit limit=null
is recommended. In addition, there’s a bit tighter type maintenance
in the internal limit data (ensure things are int)

Fixes #2950.

### DIFF
--- a/src/ORM/Queries/SQLSelect.php
+++ b/src/ORM/Queries/SQLSelect.php
@@ -238,8 +238,8 @@ class SQLSelect extends SQLConditionalExpression
      * Pass LIMIT clause either as SQL snippet or in array format.
      * Internally, limit will always be stored as a map containing the keys 'start' and 'limit'
      *
-     * @param int|string|array $limit If passed as a string or array, assumes SQL escaped data.
-     * Only applies for positive values, or if an $offset is set as well.
+     * @param int|string|array|null $limit If passed as a string or array, assumes SQL escaped data.
+     * Only applies for positive values.
      * @param int $offset
      * @throws InvalidArgumentException
      * @return $this Self reference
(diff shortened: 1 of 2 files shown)


kinglozzer:
My only concern: if using `.Limit()` in a template, is it possible to pass `null`?

sminnee:
We could allow for nulls in templates in addition to this change, I guess?

sminnee:
Having discussed with Zauberfisch further on the original ticket, perhaps we want to make this a deprecation notice, stating that in SS5 limit(0) will return no records?

kinglozzer:
> Having discussed with Zauberfisch further on the original ticket, perhaps we want to make this a deprecation notice, stating that in SS5 limit(0) will return no records?

Yeah I’d be much happier with that than raising an `E_USER_WARNING`.

We should definitely allow `null` in templates, I guess I just overlooked that in https://github.com/silverstripe/silverstripe-framework/pull/8152

src/ORM/Queries/SQLSelect.php
```diff
@@ -250,10 +250,17 @@ public function setLimit($limit, $offset = 0)
             throw new InvalidArgumentException("SQLSelect::setLimit() only takes positive values");
         }
 
+        if ($limit === 0) {
+            user_error(
+                "Query passed a limit of 0; this will return all records. To clear limit, set it to null",
+                E_USER_WARNING
```
robbieaverill: Since we're going to change this to return zero records in SS5, do you think we should mention that in the warning message?

sminnee: I'm going to replace with a deprecation notice.

sminnee:
How’s this now @kinglozzer?

PR #8430, opened 2018-10-01
FIX: Don’t use USE_FRM in MySQL repair. Fixes #6300.

### LINKED ISSUE #6300: `MySQLSchemaManager` should not use `USE_FRM`
Note docs from MySQL (http://dev.mysql.com/doc/refman/5.7/en/repair-table.html) state this is a very bad idea.

Side effects of using `USE_FRM`:

>- The current `AUTO_INCREMENT` value is lost.
>- The link to deleted records in the table is lost, which means that free space for deleted records will remain unoccupied thereafter.
>- The `.MYI` header indicates whether the table is compressed. If the server ignores this information, it cannot tell that a table is compressed and repair can cause change or loss of table contents. This means that `USE_FRM` should not be used with compressed tables. That should not be necessary, anyway: Compressed tables are read only, so they should not become corrupt.

>If you use `USE_FRM` for a table that was created by a different version of the MySQL server than the one you are currently running, `REPAIR TABLE` will not attempt to repair the table. In this case, the result set returned by `REPAIR TABLE` contains a line with a `Msg_type` value of error and a `Msg_text` value of Failed repairing incompatible `.FRM` file.

sminnee:
USE_FRM is a sufficiently dangerous option that it should require
manual DBA activity to carry out.

See http://dev.mysql.com/doc/refman/5.7/en/repair-table.html for more
info.

### DIFF
--- a/src/ORM/Connect/MySQLSchemaManager.php
+++ b/src/ORM/Connect/MySQLSchemaManager.php
@@ -167,7 +167,7 @@ class MySQLSchemaManager extends DBSchemaManager
             "Table $tableName: repaired",
             "repaired"
         );
-        return $this->runTableCheckCommand("REPAIR TABLE \"$tableName\" USE_FRM");
+        return $this->runTableCheckCommand("REPAIR TABLE \"$tableName\"");
     }
 
     /**


kinglozzer:
Whoops, I merged before I spotted you’d requested review from @dhensby... sorry Dan!

dhensby:
That's ok! I remember the PR that added this but I can't really recall having a good understanding of why it was justified.

From reading the MySQL docs I agree we shouldn't be using it.

dhensby:
I would suggest we put this into the patch releases too

PR #8432, opened 2018-10-02
NEW: Add test for PHP 7.3 support (SS4 version)

sminnee:
This also removed the allow_failures nightly build, reasoning that the
PHP 7.3 build is the primary test-case where, and allow_failures builds
are of limited value.


robbieaverill:
I think since PHP 7.3 is in the RC stage now that we shouldn't expect more breaking changes, so I'm going to merge this now 👍

PR #8433, opened 2018-10-02

sminnee:

Implemented in SS3 at 815da76b056a716c8831d22bbf93528912cbcb28

### DIFF
--- a/src/ORM/Connect/Database.php
+++ b/src/ORM/Connect/Database.php
@@ -403,7 +403,7 @@ abstract class Database
      */
     public function clearTable($table)
     {
-        $this->query("TRUNCATE \"$table\"");
+        $this->query("DELETE FROM \"$table\"");
     }
 
     /**


robbieaverill:
Ok we've got a bit of cleanup work to do with Behat tests now

robbieaverill:
Damn. Ok I think I'm going to revert this for now because we've got a bit going on at the moment and worrying about Behat failures (more than usual) isn't ideal. I like the change though for the record, just not the unintended side effects it brought. Maybe with #8434 implemented and tests updated elsewhere it would be safer

robbieaverill:
Reverted in 5a20bb5e2, sorry!

sminnee:
Yeah that makes sense. The important question for the @silverstripe/core-team is probably “are predictable IDs in tests part of our public API?”; it’s sloppy, but bear in mind that behat currently relies on predictable IDs?

We can fix our behat tests not to rely on IDs but is it okay to make this change after that?

robbieaverill:
>are predictable IDs in tests part of our public API?

Not IMO - we have `SapphireTest::idFromFixture()` for that.

The breakages I saw were in Behat tests with fixed IDs in them - we can always run this change through the kitchen sink and see what happens aside from Behat

chillu:
>  are predictable IDs in tests part of our public API?

I'd say they aren't. Your business logic shouldn't rely on hardcoded identifiers, and your tests shouldn't either.

sminnee:
It's okay we managed to have our cake & eat it too, speeding things up while also resetting auto-increments.

sminnee:
> Your tests shouldn't either.

But yours (well, ours) do! :trollface:

sminnee:
FYI to @chillu @wilr and others - this is where we had hardcoded IDs; it only appeared in behat tests due to limitations about our behat assertions:
https://github.com/silverstripe/silverstripe-cms/blob/4/tests/behat/features/insert-a-link.feature#L25

I was going to try and adjust this so you could pass a regex instead of a literal in your behat tests, but this is a good sign that we want to avoid breaking this until a major release. Luckily we found a way of getting the speedup in spite of that.

chillu:
Created a ticket, and continued discussion there (since the PR will get lost as its merged): https://github.com/silverstripe/silverstripe-framework/issues/8445

PR #8434, opened 2018-10-02
NEW: Allow regexes in behat tests of HTML content.

sminnee:
This will be necessary to adjust CMS tests to not hard-code specific
page IDs, which is necessary to speed up clearAllData()


robbieaverill:
Tests are broken =/

PR #8435, opened 2018-10-02

sminnee:

### DIFF
--- a/model/connect/MySQLDatabase.php
+++ b/model/connect/MySQLDatabase.php
@@ -362,4 +362,23 @@ class MySQLDatabase extends SS_Database {
 	public function random() {
 		return 'RAND()';
 	}
+
+	/**
+	 * Clear all data in a given table
+	 *
+	 * @param string $table Name of table
+	 */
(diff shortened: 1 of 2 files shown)


sminnee:
This is a massive speedup but it’s too risky for a patch release as it may break users’ test suites if they rely on predictable IDs

I might try a different way of speeding up MySQL tests

ScopeyNZ:
> I wonder whether we could `DELETE FROM` then reset the auto increment value? [stackoverflow.com/a/8923132/2812842](https://stackoverflow.com/a/8923132/2812842)

PostgreSQL doesn't have auto-increments and instead uses `SERIAL` field types which create a matching sequence and set the default value of the field to `nextval(SEQUENCE)`. The syntax would have to be different for PostgreSQL

sminnee:
That's easy enough - we can implement in MySQLDatabase instead of Database. Note that PGSQL already calls DELETE FROM.

sminnee:
Excellent, that worked and provided most of the original speed-up!

PR #8437, opened 2018-10-03

sminnee:

Cherry-pick of SS3 ae9ab22a8ff1b48c90f7dfe2899c09efaa65b161 provided in #8435


tractorcow:
Not well known but truncate counts as DDL and invalidates all cached prepared statements. delete from doesn't have this problem. :)

PR #8439, opened 2018-10-03
FIX: Make ArrayList::limit() consistent with DataList::limit()

### LINKED ISSUE #2949: [ORM] Inconsitent implementations of SS_Limitable
`DataList` and `ArrayList` implement the `SS_Limitable` interface. they however have inconsistent behavior in `$list->limit()` when passing anything other than positive integer values.

```
$arrayList->limit(-1); // returns all records except the last one, so its $l->limit($l->count()-1)
$dataList->limit(-1); // throws InvalidArgumentException

$arrayList->limit("some string"); // returns 0 records
$dataList->limit("some string"); // raises a '[Notice] Undefined offset: 1' but still returns all records
```

and then there is the fact that `->limit(0)` returns all items, see #2950 for the full issue.

----

## To do

 * [x] Deprecate in SS4 (#8439)
 * [ ] Change ArrayList and DataList behaviour in SS5

sminnee:
This makes it easier to swap one fo the other without code breaking.
Since it’s strictly a removed API, I’ve opted to throw a deprecation
note in SS4 rather than throwing an InvalidArgumentException.

Fixes #2949

### DIFF
--- a/src/ORM/ArrayList.php
+++ b/src/ORM/ArrayList.php
@@ -8,6 +8,7 @@ use SilverStripe\View\ViewableData;
 use ArrayIterator;
 use InvalidArgumentException;
 use LogicException;
+use SilverStripe\Dev\Deprecation;
 
 /**
  * A list object that wraps around an array of objects or arrays.
@@ -160,7 +161,29 @@ class ArrayList extends ViewableData implements SS_List, Filterable, Sortable, L
      */
     public function limit($length, $offset = 0)
     {
+        // Type checking: designed for consistency with DataList::limit()
+        if (!is_numeric($length) || !is_numeric($offset)) {
+            Deprecation::notice(
+                '4.3',
+                'Arguments to ArrayList::limit() should be numeric'
+            );
+        }
+
+        if ($length < 0 || $offset < 0) {
+            Deprecation::notice(
+                '4.3',
+                'Arguments to ArrayList::limit() should be positive'
+            );
+        }
+
         if (!$length) {
+            if ($length === 0) {
+                Deprecation::notice(
+                    '4.3',
+                    "limit(0) is deprecated in SS4. In SS5 a limit of 0 will instead return no records."
+                );
+            }
+
             $length = count($this->items);
         }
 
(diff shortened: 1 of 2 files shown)


Zauberfisch:
haven't tested it. but code looks good

sminnee:
There's a breakage in that `limit(null)` to clear limit still needs to be supported.

We should probably also handle `limit(0)` in the same way that I was envisioning `DataList::limit(0)` from being handled (deprecation notice that states in SS5 this will return no records).

src/ORM/ArrayList.php
```diff
@@ -159,6 +160,21 @@ public function toNestedArray()
...
+        if (!is_numeric($length) || !is_numeric($offset)) {
+            Deprecation::notice(
+                '4.0',
+                'Arguments to ArrayList::limit() should be numeric'
+            );
+        }
+
+        if ($length < 0 || $offset < 0) {
+            Deprecation::notice(
+                '4.0',
```
robbieaverill: We should probably deprecate from 4.3 on

kinglozzer: I thought the first arg was the target version for removal, in this case `'5.0'`?

robbieaverill: No it's the version the notice will start being raised from

sminnee:
To do:

 * [ ] change deprecation version to 4.3
 * [ ] ensure that null is allowed and both null and 0 are covered by tests as with DataList

sminnee:
I opted to leave special consideration for limit(null) because our current ArrayList implementation doesn't actually allow for this very naturally.

unclecheese:
Have retargeted to `4.3`

tests/php/ORM/ArrayListTest.php
```diff
@@ -133,6 +134,19 @@ public function testLimit()
         );
     }
 
+    /**
+     * @expectedException PHPUnit_Framework_Error
```
robbieaverill: A note for anyone who merged up this pull request - in PHPUnit 7 the equivalent of this class is `\PHPUnit\Framework\Error\Error` and it must be a FCQN, not an imported class when using it in `@expectedException`

PR #8444, opened 2018-10-03
FIX allow extensions to be applied via injector service definitions

NightJar:
Throughout the code there is the implication of being able to apply
extensions through an Injector service name - such as the stripping
of %$ before looking up an extensions name.

However this is not possible as class_exists is exclusively used to
ensure that the Extension is a valid subclass of Extension (or
DataExtension).

This commit seeks to rectify that by falling back to using Injector
to see if there is a service registered with the given name, and if
so using it via instantion before performing the sanity check of
the concrete class' inheritance heirarchy.

### DIFF
--- a/src/Core/Config/Middleware/ExtensionMiddleware.php
+++ b/src/Core/Config/Middleware/ExtensionMiddleware.php
@@ -10,6 +10,7 @@ use SilverStripe\Config\Middleware\MiddlewareCommon;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Extension;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\ORM\DataExtension;
 
 class ExtensionMiddleware implements Middleware
@@ -66,6 +67,9 @@ class ExtensionMiddleware implements Middleware
             list($extensionClass, $extensionArgs) = ClassInfo::parse_class_spec($extension);
             // Strip service name specifier
             $extensionClass = strtok($extensionClass, '.');
+            if ($extensionSpec = Injector::inst()->getServiceSpec($extensionClass)) {
+                $extensionClass = $extensionSpec['class'];
+            }
             if (!class_exists($extensionClass)) {
                 throw new InvalidArgumentException("$class references nonexistent $extensionClass in 'extensions'");
             }
--- a/tests/php/Core/ObjectTest.php
+++ b/tests/php/Core/ObjectTest.php
@@ -4,6 +4,7 @@ namespace SilverStripe\Core\Tests;
 
 use SilverStripe\Control\Controller;
 use SilverStripe\Core\ClassInfo;
+use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Extension;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Tests\ObjectTest\BaseObject;
@@ -12,6 +13,7 @@ use SilverStripe\Core\Tests\ObjectTest\ExtendTest2;
 use SilverStripe\Core\Tests\ObjectTest\ExtendTest3;
 use SilverStripe\Core\Tests\ObjectTest\ExtendTest4;
 use SilverStripe\Core\Tests\ObjectTest\ExtensionRemoveTest;
+use SilverStripe\Core\Tests\ObjectTest\ExtensionService;
 use SilverStripe\Core\Tests\ObjectTest\ExtensionTest;
 use SilverStripe\Core\Tests\ObjectTest\ExtensionTest2;
 use SilverStripe\Core\Tests\ObjectTest\ExtensionTest3;
@@ -537,5 +539,27 @@ class ObjectTest extends SapphireTest
         $this->assertInstanceOf($mockClass, $object->getExtensionInstance(TestExtension::class));
         $this->assertInstanceOf(TestExtension::class, $object->getExtensionInstance($mockClass));
         $this->assertInstanceOf($mockClass, $object->getExtensionInstance($mockClass));
+    }
+
+    /**
+     * @expectedException InvalidArgumentException
+     */
+    public function testExtensionsAppliedViaANonExistentServiceAsAControl()
+    {
+        Config::modify()->set(BaseObject::class, 'extensions', ['NotAService']);
+        BaseObject::create();
+    }
+
+    public function testExtensionsCanBeAppliedViaInjectorServices()
+    {
+        $config = Config::modify();
+        $config->set(Injector::class, 'ATestService', [
+            'class' => ExtensionService::class,
+            'constructor' => ['nonsense'],
+        ]);
+        $config->set(BaseObject::class, 'extensions', ['ATestService']);
+
+        $nonsense = BaseObject::create();
+        $this->assertEquals('This extension is nonsense', $nonsense->aMethod());
     }
 }
--- a/tests/php/Core/ObjectTest/ExtensionService.php
+++ b/tests/php/Core/ObjectTest/ExtensionService.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace SilverStripe\Core\Tests\ObjectTest;
+
+use SilverStripe\Core\Extension;
+use SilverStripe\Dev\TestOnly;
+
+class ExtensionService extends Extension implements TestOnly
+{
+    private $property;
+
+    public function __construct($property)
+    {
+        $this->property = $property;
+    }
+
+    public function aMethod()
+    {
+        return 'This extension is ' . $this->property;
+    }
+}
--- a/src/Core/Extensible.php
+++ b/src/Core/Extensible.php
@@ -178,10 +178,13 @@ trait Extensible
             $extension = $classOrExtension;
         }
 
-        if (!preg_match('/^([^(]*)/', $extension, $matches)) {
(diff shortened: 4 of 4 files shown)


src/Core/Config/Middleware/ExtensionMiddleware.php
```diff
@@ -66,10 +67,14 @@ protected function getExtraConfig($class, $classConfig, $excludeMiddleware)
             list($extensionClass, $extensionArgs) = ClassInfo::parse_class_spec($extension);
             // Strip service name specifier
             $extensionClass = strtok($extensionClass, '.');
-            if (!class_exists($extensionClass)) {
+            if (!class_exists($extensionClass) && !Injector::inst()->has($extensionClass)) {
                 throw new InvalidArgumentException("$class references nonexistent $extensionClass in 'extensions'");
             }
 
+            if (!class_exists($extensionClass)) {
```
robbieaverill: Maybe you could combine this into the previous if statement to reduce duplicating the condition

tests/php/Core/ObjectTest.php
```diff
@@ -538,4 +540,26 @@ public function testInjectedExtensions()
         $this->assertInstanceOf(TestExtension::class, $object->getExtensionInstance($mockClass));
         $this->assertInstanceOf($mockClass, $object->getExtensionInstance($mockClass));
     }
+
+    /**
+     * @expectedException InvalidArgumentException
+     */
+    public function testExtensionsAppliedViaANonExantServiceAsAControl()
```
robbieaverill:
```diff
- Exant
- Existant
```

tests/php/Core/ObjectTest.php
```diff
@@ -538,4 +540,26 @@ public function testInjectedExtensions()
...
+    {
+        Config::modify()->set(BaseObject::class, 'extensions', ['NotAService']);
+        $failure = BaseObject::create();
```
robbieaverill: This doesn't need to be an assignment - `$failure` is irrelevant

tests/php/Core/ObjectTest/ExtensionService.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace SilverStripe\Core\Tests\ObjectTest;
+
+use SilverStripe\Core\Extension;
+
+class ExtensionService extends Extension
```
robbieaverill: Can you please add `TestOnly`?

src/Core/Config/Middleware/ExtensionMiddleware.php
NightJar: I thought about it, but it gets difficult as the previous condition is dealing specifically with the lack of reference. It could be done, however the trade-off would be a touch of pyramid nesting I think.

tests/php/Core/ObjectTest.php
NightJar: That is true. But It is also of no consequence. It does serve as an indication. I could swap it out for a comment (or nothing) if you prefer.

tests/php/Core/ObjectTest/ExtensionService.php
NightJar: ooh, whoops!

tests/php/Core/ObjectTest.php
NightJar: It's a word :)

src/Core/Config/Middleware/ExtensionMiddleware.php
robbieaverill: How about something like this?

```php
$extensionInstance = $this->getExtensionInstance($extensionClass);

protected function getExtensionInstance($className)
{
    if (Injector::inst()->has($className)) {
        return Injector::inst()->get($className);
    }

    if (!class_exists($className)) {
        throw new Exception('...');
    }

    return $className;
}
```

Something like this might be good, you could implement it in the places where the logic is duplicated.

Also note that returning either an instance or a class name string from the logic seems odd (before this PR it only returns a class name string), so perhaps you could change it to always return an instance of the extension

tests/php/Core/ObjectTest.php
robbieaverill:
```diff
- $failure = BaseObject::create();
+ BaseObject::create();
```

👍 

IDEs will warn this as irrelevant code otherwise and eventually someone else will remove it 😄

NightJar: ok, fair enough :)

tests/php/Core/ObjectTest.php
robbieaverill: What benefit do you add by using it over "existant", which anyone would understand?

NightJar: It was shorter :P

robbieaverill: We have [some guidelines for writing style](https://docs.silverstripe.org/en/4/contributing/documentation/#writing-style), one of which is "Use simple language and words. Avoid uncommon jargon and overly long words."

As a native English speaker I don't know what "exant" means, nor can I find a definition for it (should it be "extant"? I still don't know what that means without looking it up). I can imagine this writing style being even harder to understand for a non-native English speaker =)

src/Core/Config/Middleware/ExtensionMiddleware.php
NightJar: Yeah sweet I'll do something like that :)

NightJar: So I looked at this - but in the end there are two issues:
1. The exceptions require `$class` (that the extension would belong to) which would not exist if I abstracted the method - passing it around seems silly. I had the new method setting on Extension because it seemed the only logical place for it to be, otherwise I'd introduce coupling between the `trait` and the `Middleware`. One of the calls actually results in an `error` as opposed to an `Exception` as well... so there'd need to be code to allow for that anyway, otherwise we'd be changing the API.
2. Adding functions frivolously for what essentially amounts to 'Injector::doOneThing()` also seemed silly at that point. I could add `getConcreteClassName` to `Injector`... `getServiceNamedSpec` already exists, but it's protected.

robbieaverill:
To clarify the problem, you're suggesting that you want to be able to do something like this?

```yaml
YourClass:
  extensions:
    - '%$YourExtension'
```

If that's the case, I'm not sure we need to support it. The `%$` syntax tells injector to instantiate the class name after it - extensions are still created with injector, and are handled as singletons by default (similar idea to a service I guess).

Can you clarify a little please?

NightJar:
Correct that extensions are still instantiated with injector, however they are not _tested_ in such a manner. This leads to a valid service name (such as `%$YourExtension`) resulting in either a `Fatal Error` or equally fatal uncaught `Invalid Argument Exception`. As testing whether or not an Extension is valid happens _before_ instantiation, this is less than ideal.

https://github.com/silverstripe/silverstripe-framework/blob/c7e341c67d62343574bc8b9ae0c82dd73714816d/src/Core/Extensible.php#L560-L564

robbieaverill:
Right i see, so you’re aliasing extensions and this is adding support for it

NightJar:
Is there anything outstanding on this?
I've just realised it's missed the 4.3 bus :(

robbieaverill:
ping @silverstripe/core-team for feedback on this

src/Core/Config/Middleware/ExtensionMiddleware.php
tractorcow: I was going to recommend exactly what @robbieaverill suggested. That is, using has() instead of the spec. You can do it inline though. It doesn't need to be a method.

NightJar: That does not consider the code outside this change, where `$extensionClass` is then used to look up the concrete parent chain via `ClassInfo` - which would severely break if it was not both set, or a valid concrete class.

tractorcow: If you just rely on injector you don't need to do class_exists() or other checks. Injector will internally validate the service is valid. Do you really need to do concrete class lookups outside of injector?

NightJar: Yes.
https://github.com/silverstripe/silverstripe-framework/blob/d75ce9b39ef4803a651b66f873fdbe5f48cd2ff2/src/Core/Config/Middleware/ExtensionMiddleware.php#L99-L109

NightJar:
Any more feedback? I think it would be nice to get in to 4.4.0

NightJar:
Hi @maxime-rainville, thanks for picking up. I'll update with some documentation adjustments.

I've targeted 4.0 in order to address this as a bug due to this code implying this should be a feature already, but it does not work (thus bug). I can re-target if you feel strongly about it not being so.

https://github.com/silverstripe/silverstripe-framework/blob/c7e341c67d62343574bc8b9ae0c82dd73714816d/src/Core/Extensible.php#L560-L564

maxime-rainville:
Even if it's a bug, I don't think it's a high impact one. It would probably be better to release it has part of the new minor release rather than in a patch release.

ScopeyNZ:
Hey @NightJar. Feel free to reopen this targeting a new branch!

unclecheese:
@NightJar This was automatically closed because it was targeted against an EOL branch. Can you please rebase against 4.2 and retarget? (4.2 is EOL in 11 days, FYI)

robbieaverill:
I've revived this PR and recreated it at https://github.com/silverstripe/silverstripe-framework/pull/9457

Issue #6709, opened 2017-03-15
[ORM] DataObject::write() causing unnecessary writes with integer like data type


UndefinedOffset:
In 3.5 (at least) when an integer like data type (for example a Boolean field or ForeignKey) is updated by ``Form::saveInto()`` it's updated as an actual integer. However when it's read from the database and stored in $record it is a string. Due to this when ``DataObject::write()`` calls ``DataObject::isChanged()`` it does not pass any arguments, which the default setting for the ``$changeLevel`` argument is ``self::CHANGE_STRICT``. This causes ``DataObject::isChanged()`` to flag the integer like fields as changed because of strict type checking. This is most noticeable with a versioned DataObject (other than SiteTree), every write with the below scenario causes a new versioned record. SiteTree gets around this I think because of [these lines](https://github.com/silverstripe/silverstripe-cms/blob/27f8acf3732ee755d784a79f17d6d04c97a43663/code/model/SiteTree.php#L1535-L1546) given the comments around that area it would appear to be a happy accident though maybe intended.

Given a DataObject with a Boolean field named ShowHeading and an Image relationship called BackgroundImage. Using a form that controls those 2 fields just before the manipulation is run for the object the ``$changed`` variable contains the following:

```
array(2) {
  ["ShowHeading"]=>
  array(3) {
    ["before"]=>
    string(1) "1"
    ["after"]=>
    int(1)
    ["level"]=>
    int(1)
  }
  ["BackgroundImageID"]=>
  array(3) {
    ["before"]=>
    string(1) "0"
    ["after"]=>
    int(0)
    ["level"]=>
    int(1)
  }
}
```

tractorcow:
Here's a good question, should we treat `(int)1` => `(string)"1"` as a strict change? I believe the intent of the change is more to detect changes such as `(null)` => `(string)""`. Numeric values are a special case, since different ORM backends will return these either as string values, or real integers, triggering the case you noted above.

One way to address this is to move the change detection into the DBField instance itself, and code type-specific change detection into each subclass.

> happy accident though

![](http://images.mentalfloss.com/sites/default/files/styles/article_640x430/public/bob_0.png)

dhensby:
The ORM should be dealing with casting the data correctly when it's loaded from the DB. Ideally in the DB layer but the DBFields/DataObject could do it too.

FormFields should cast their data properly when loading into dataobjects too

dhensby:
[...]

PR #8448, opened 2018-10-04
FIX: Ensure that types are preserved fetching from database

### LINKED ISSUE #7039: [ORM] Inconsistent DB ORM casting of int values

sminnee:
This ensures that numeric fields appear in PHP as int/float values
rather than strings, which allows the development of more type-safe PHP
code.

This doesn’t work on the legacy mysql driver and this will now throw
a notice-level error. It requires mysqlnd.

Fixes #7039, and other bugs such as #6709

Also requires https://github.com/silverstripe/silverstripe-postgresql/pull/91

Taken from https://github.com/open-sausages/silverstripe-framework/commit/f4e6aae411f2ffcfd52ff08e6bf66dbe4eeee464

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,20 +20,20 @@ matrix:
   include:
     - php: 5.6
       env:
-        - DB=MYSQL
+        - DB=PGSQL
         - PHPCS_TEST=1
         - PHPUNIT_TEST=framework
 
     - php: 7.0
       env:
         - DB=PGSQL
+        - PDO=1
         - PHPUNIT_TEST=framework
 
     - php: 7.1
       if: type IN (cron)
       env:
         - DB=MYSQL
-        - PDO=1
         - PHPUNIT_COVERAGE_TEST=framework
 
     - php: 7.2
@@ -50,7 +50,6 @@ matrix:
     - php: 7.3.0RC1
       env:
         - DB=MYSQL
-        - PDO=1
         - PHPUNIT_TEST=framework
       sudo: required
       dist: xenial
@@ -74,7 +73,7 @@ before_script:
 # Install composer dependencies
   - composer validate
   - mkdir ./public
-  - if [[ $DB == PGSQL ]]; then composer require silverstripe/postgresql:2.1.x-dev --no-update; fi
+  - if [[ $DB == PGSQL ]]; then composer require silverstripe/postgresql:2.2.x-dev --no-update; fi
   - if [[ $DB == SQLITE ]]; then composer require silverstripe/sqlite3:2.0.x-dev --no-update; fi
   - composer require silverstripe/recipe-testing:^1 silverstripe/recipe-core:4.4.x-dev silverstripe/admin:1.4.x-dev silverstripe/versioned:1.4.x-dev --no-update
   - if [[ $PHPUNIT_TEST == cms ]]; then composer require silverstripe/recipe-cms:4.4.x-dev --no-update; fi
--- a/src/ORM/Connect/MySQLDatabase.php
+++ b/src/ORM/Connect/MySQLDatabase.php
@@ -21,7 +21,7 @@ use Exception;
  * You are advised to backup your tables if changing settings on an existing database
  * `connection_charset` and `charset` should be equal, similarly so should `connection_collation` and `collation`
  */
-class MySQLDatabase extends Database
+class MySQLDatabase extends Database implements TransactionManager
 {
     use Configurable;
 
@@ -50,17 +50,19 @@ class MySQLDatabase extends Database
     private static $charset = 'utf8';
 
     /**
+     * Cache for getTransactionManager()
+     *
+     * @var TransactionManager
+     */
+    private $transactionManager = null;
+
+    /**
      * Default collation
      *
      * @config
      * @var string
      */
     private static $collation = 'utf8_general_ci';
-
-    /**
-     * @var bool
-     */
-    protected $transactionNesting = 0;
 
     public function connect($parameters)
     {
@@ -298,73 +300,64 @@ class MySQLDatabase extends Database
         return $list;
     }
 
+
+    /**
+     * Returns the TransactionManager to handle transactions for this database.
+     *
+     * @return TransactionManager
+     */
+    protected function getTransactionManager()
+    {
+        if (!$this->transactionManager) {
+            // PDOConnector providers this
+            if ($this->connector instanceof TransactionManager) {
+                $this->transactionManager = new NestedTransactionManager($this->connector);
+            // Direct database access does not
+            } else {
+                $this->transactionManager = new NestedTransactionManager(new MySQLTransactionManager($this));
+            }
+        }
+        return $this->transactionManager;
+    }
     public function supportsTransactions()
     {
         return true;
     }
+    public function supportsSavepoints()
+    {
+        return $this->getTransactionManager()->supportsSavepoints();
+    }
 
     public function transactionStart($transactionMode = false, $sessionCharacteristics = false)
     {
-        if ($this->transactionNesting > 0) {
-            $this->transactionSavepoint('NESTEDTRANSACTION' . $this->transactionNesting);
-        } else {
-            // This sets the isolation level for the NEXT transaction, not the current one.
-            if ($transactionMode) {
-                $this->query('SET TRANSACTION ' . $transactionMode);
-            }
-
-            $this->query('START TRANSACTION');
-
-            if ($sessionCharacteristics) {
-                $this->query('SET SESSION TRANSACTION ' . $sessionCharacteristics);
-            }
-        }
-        ++$this->transactionNesting;
+        $this->getTransactionManager()->transactionStart($transactionMode, $sessionCharacteristics);
     }
 
     public function transactionSavepoint($savepoint)
     {
-        $this->query("SAVEPOINT $savepoint");
+        $this->getTransactionManager()->transactionSavepoint($savepoint);
     }
 
     public function transactionRollback($savepoint = false)
     {
-        // Named transaction
-        if ($savepoint) {
-            $this->query('ROLLBACK TO ' . $savepoint);
-            return true;
-        }
-
-        // Fail if transaction isn't available
-        if (!$this->transactionNesting) {
-            return false;
-        }
-        --$this->transactionNesting;
-        if ($this->transactionNesting > 0) {
-            $this->transactionRollback('NESTEDTRANSACTION' . $this->transactionNesting);
-        } else {
-            $this->query('ROLLBACK');
-        }
-        return true;
+        return $this->getTransactionManager()->transactionRollback($savepoint);
     }
 
     public function transactionDepth()
     {
-        return $this->transactionNesting;
+        return $this->getTransactionManager()->transactionDepth();
     }
 
     public function transactionEnd($chain = false)
     {
-        // Fail if transaction isn't available
-        if (!$this->transactionNesting) {
-            return false;
+        $result = $this->getTransactionManager()->transactionEnd();
+
+        if ($chain) {
+            Deprecation::notice('4.4', '$chain argument is deprecated');
+            return $this->getTransactionManager()->transactionStart();
         }
-        --$this->transactionNesting;
-        if ($this->transactionNesting <= 0) {
-            $this->transactionNesting = 0;
-            $this->query('COMMIT AND ' . ($chain ? '' : 'NO ') . 'CHAIN');
-        }
-        return true;
+
+        return $result;
     }
 
     /**
@@ -372,6 +365,12 @@ class MySQLDatabase extends Database
      */
     protected function resetTransactionNesting()
     {
+        // Check whether to use a connector's built-in transaction methods
+        if ($this->connector instanceof TransactionalDBConnector) {
+            if ($this->transactionNesting > 0) {
+                $this->connector->transactionRollback();
+            }
+        }
         $this->transactionNesting = 0;
     }
 
--- a/src/ORM/Connect/MySQLTransactionManager.php
+++ b/src/ORM/Connect/MySQLTransactionManager.php
@@ -1,0 +1,100 @@
+<?php
+
+namespace SilverStripe\ORM\Connect;
+
+use SilverStripe\Dev\Deprecation;
+
+/**
+ * TransactionManager that executes MySQL-compatible transaction control queries
+ */
+class MySQLTransactionManager implements TransactionManager
+{
+    protected $dbConn;
+
+    protected $inTransaction = false;
+
+    public function __construct(Database $dbConn)
+    {
+        $this->dbConn = $dbConn;
+    }
+
+    public function transactionStart($transactionMode = false, $sessionCharacteristics = false)
+    {
+        if ($transactionMode || $sessionCharacteristics) {
+            Deprecation::notice(
+                '4.4',
+                '$transactionMode and $sessionCharacteristics are deprecated and will be removed in SS5'
+            );
+        }
+
+        if ($this->inTransaction) {
+            throw new DatabaseException(
+                "Already in transaction, can't start another. Consider decorating with NestedTransactionManager."
+            );
+        }
+
+        // This sets the isolation level for the NEXT transaction, not the current one.
+        if ($transactionMode) {
+            $this->dbConn->query('SET TRANSACTION ' . $transactionMode);
+        }
+
+        $this->dbConn->query('START TRANSACTION');
+
+        if ($sessionCharacteristics) {
+            $this->dbConn->query('SET SESSION TRANSACTION ' . $sessionCharacteristics);
+        }
+
+        $this->inTransaction = true;
+        return true;
+    }
+
+    public function transactionEnd($chain = false)
+    {
+        if (!$this->inTransaction) {
+            throw new DatabaseException("Not in transaction, can't end.");
+        }
+
+        if ($chain) {
+            user_error(
+                "transactionEnd() chain argument no longer implemented. Use NestedTransactionManager",
+                E_USER_WARNING
+            );
+        }
+
+        $this->dbConn->query('COMMIT');
+
+        $this->inTransaction = false;
+        return true;
+    }
+
+    public function transactionRollback($savepoint = null)
+    {
+        if (!$this->inTransaction) {
+            throw new DatabaseException("Not in transaction, can't roll back.");
+        }
+
+        if ($savepoint) {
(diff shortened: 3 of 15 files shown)


sminnee:
OK @dhensby @tractorcow I made quite a bit of progress with this once I realised that the underlying issue is that, when using PDO, you're expected to also use their transaction methods.

Unfortunately their transaction methods don't supported nested transactions. So, the current code "fakes" the nested transactions by making any internal transaction a no-op. Right now this is breaking the tests that do things with transactions.

So, there's a questions: how badly do we need nested transactions that allow you to roll-back only the inner transaction? I would probably advocate for a nested transaction model where if an inner transaction had an issue, the top-level transaction would also be rolled back. In real world code where transaction rollback is primarily to prevent data corruption in the case of failures, this should do.

We can then amend our tests to suit this expectation.

Related: does DBAL support nested transactions?

sminnee:
OK I've had a look at https://www.doctrine-project.org/projects/doctrine-dbal/en/2.8/reference/transactions.html#transaction-nesting about DBAL, and they do "fake" transaction nesting similar to what I have implemented here for PDODatabase. I'm going to tidy up this implementation so that it matches their semantics and then get our test suite working with that.

tractorcow:
Ah yes, I found my own proof of concept of such an issue a while back.

https://gist.github.com/tractorcow/390c9075020b6e2fdfa807c835403afa

Maybe you can fix it. :)

sminnee:
Yeah the big breakthrough was realising that doing transactions without the PDO's built-in transaction methods was a big idea.

I've solved the problem now — it depends on a postgresql PR to get the build passing, but the code is written.

I just need feedback on whether it's solved in the right way.

sminnee:
OK I've gone with formally declaring booleans to be `1` and `0`, following MySQL's lead (which doesn't have a native boolean type)

sminnee:
@robbieaverill @kinglozzer might be good to broaden the review pool for this? See also https://github.com/silverstripe/silverstripe-postgresql/pull/91

kinglozzer:
I’ll try to find time to do a proper code review later, immediate thoughts/queries:

- I’m not familiar with the finer points of transactions and what our API currently supports, but is this acceptable for a mid-4.x-release-line change? https://github.com/silverstripe/silverstripe-framework/pull/8448/files#diff-517cba5a1e63143651649368ef337664R333
- If we are going to go with this change, do we need to either:
  - Make it opt-in? I can’t imagine why anyone would be doing strict type checks on incorrect types, but you never know...
  - Somehow preserve the existing transaction behaviour - in a “legacy transaction manager” perhaps?

My preference would be for booleans to represented as actual booleans, but probably at a higher (likely DBField) level. i.e. an `SQLSelect` might return a `0` or `1`, but `$myDataObject->BooleanField` would return an actual boolean. That can always come after this PR of course :)

tests/php/ORM/DatabaseTest.php
```diff
@@ -270,4 +271,40 @@ public function testFieldTypes()
...
+            $success = DataObject::get(MyObject::class, "\"MyField\"='Read only success'");
+            $fail = DataObject::get(MyObject::class, "\"MyField\"='Read only page failed'");
+
+            //This page should be in the system
+            $this->assertTrue(is_object($success) && $success->exists());
+
+            //This page should NOT exist, we had 'read only' permissions
+            $this->assertFalse(is_object($fail) && $fail->exists());
+        } else {
+            $this->markTestSkipped('Current database is doesn\'t support transactions');
```
robbieaverill: You could probably do an early return instead to save all the indentation

tests/php/ORM/DatabaseTest.php
```diff
@@ -231,4 +237,74 @@ function () use (&$failed) {
...
+            DataObject::flush_and_destroy_cache();
+
+            $success = DataObject::get(MyObject::class, "\"MyField\"='Read only success'");
+            $fail = DataObject::get(MyObject::class, "\"MyField\"='Read only page failed'");
+
+            //This page should be in the system
+            $this->assertTrue(is_object($success) && $success->exists());
+
+            //This page should NOT exist, we had 'read only' permissions
+            $this->assertFalse(is_object($fail) && $fail->exists());
```
robbieaverill: What do you think about
```suggestion
            $this->assertNotInternalType('object', $fail);
            $this->assertFalse($fail->exists()); 
```

Note that second assertion will probably fail because `$fail` isn't an object, so maybe omit it. We could do the opposite assertions for the `assertTrue()` line above

src/ORM/Connect/PDOConnector.php
```diff
@@ -468,4 +496,48 @@ public function isActive()
...
+     */
+    public function transactionRollback()
+    {
+        $this->inTransaction = false;
+        return $this->pdoConnection->rollBack();
+    }
+
+    /**
+     * @inherit
+     */
```
robbieaverill: You can omit these, the `@inheritDoc` tag on its own is functionally equivalent to not having a doc block at all

src/ORM/Connect/PDOConnector.php
```diff
@@ -213,6 +231,16 @@ public function connect($parameters, $selectDB = false)
         }
     }
 
+
+    /**
+     * Return the driver for this connector
+     * E.g. 'mysql', 'sqlsrv', 'pgsql'
```
robbieaverill: Can we add `@return string`?

src/ORM/Connect/PDOConnector.php
```diff
@@ -64,6 +64,18 @@ class PDOConnector extends DBConnector
      */
     protected $cachedStatements = array();
 
+    /**
+     * Driver
+     * @var string
+     */
+    private $driver = null;
```
robbieaverill: I think we should use protected over private

sminnee:
> I’m not familiar with the finer points of transactions and what our API currently supports, but is this acceptable for a mid-4.x-release-line change? https://github.com/silverstripe/silverstripe-framework/pull/8448/files#diff-517cba5a1e63143651649368ef337664R333

The nature of this changes is that it moves away from _genuine_ support for nested transactions to simulated nested transactions, which means that child transactions can't be rolled back without rolling back the entire transaction, which is essentially the same behaviour as savepoints.

This is because PDO is dislikes running transaction-control statements as regular queries; it really wants you to use its own transaction control commands instead. Specifically, they cause bugs when running unbuffered queries, and unbuffered queries are necessary to get correct type data.

The two broad paths that I considered were:

 * Deprecate 'true' nested queries
 * Deprecate PDO

I'm not joking about the 2nd one. Although on the surface of it using a built-in abstraction library is "better", in practise, PDO has been a deeply disappointing system at many turns, providing underwhelming lowest-common-denominator functionality and confusing bugs and inconsistencies.

However, I've opted for the first solution in this PR, mainly because it matches what we would need to do in order to move to DBAL, which Dan has previously raised as an option for the future. So, PDO lives on. ;-)

> My preference would be for booleans to represented as actual booleans, but probably at a higher (likely DBField) level. i.e. an SQLSelect might return a 0 or 1, but $myDataObject->BooleanField would return an actual boolean. That can always come after this PR of course :)

Yeah, we can't easily do this as the database level because MySQL doesn't return "type = boolean" data. Creating a DataObject-level casting of values to booleans would require the instantiation of a whole lot of DBBoolean objects and I'm worried about the performance impact, so I've stayed away from that.

src/ORM/Connect/NestedTransactionManager.php
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        if ($chain) {
+            return $this->transactionStart();
+        }
+    }
+
+    /**
+     * @inherit
+     */
```
robbieaverill: Remove

src/ORM/Connect/MySQLTransactionManager.php
```diff
@@ -0,0 +1,98 @@
...
+
+        $this->dbConn->query('COMMIT');
+
+        $this->inTransaction = false;
+        return true;
+    }
+
+    /**
+     * @inherit
+     */
```
robbieaverill: Remove

src/ORM/Connect/MySQLDatabase.php
```diff
@@ -298,80 +299,77 @@ public function searchEngine(
...
-            return false;
-        }
-        --$this->transactionNesting;
-        if ($this->transactionNesting <= 0) {
-            $this->transactionNesting = 0;
-            $this->query('COMMIT AND ' . ($chain ? '' : 'NO ') . 'CHAIN');
+        $result = $this->getTransactionManager()->transactionEnd();
+
+        if ($chain) {
+            Deprecation::notice('4.3', '$chain argument is dereprecated');
```
robbieaverill: Type in deprecated

src/ORM/Connect/MySQLDatabase.php
```diff
@@ -298,80 +299,77 @@ public function searchEngine(
...
+        user_error('transactionSavepoint() no longer does anything', E_USER_WARNING);
     }
 
     public function transactionRollback($savepoint = false)
     {
-        // Named transaction
         if ($savepoint) {
-            $this->query('ROLLBACK TO ' . $savepoint);
-            return true;
+            user_error('transactionRollback() $savepont option no longer does anything', E_USER_WARNING);
```
robbieaverill: Type in variable name

src/ORM/Connect/MySQLDatabase.php
```diff
@@ -49,6 +49,12 @@ class MySQLDatabase extends Database
      */
     private static $charset = 'utf8';
 
+    /**
+     * Cache for getTransactionManager()
+     * @var TransactionManager
+     */
+    private $transactionManager = null;
```
robbieaverill: Protected over private?

.travis.yml
```diff
@@ -20,20 +20,20 @@ matrix:
...
       env:
-        - DB=MYSQL
+        - DB=PGSQL
         - PHPCS_TEST=1
         - PHPUNIT_TEST=framework
 
     - php: 7.0
       env:
         - DB=PGSQL
+        - PDO=1
```
robbieaverill: It'd probably be good to keep a PDO run on MySQL too

sminnee:
> Somehow preserve the existing transaction behaviour - in a “legacy transaction manager” perhaps?

Yeah, we could potentially provide support for savepoint-based transactions on
 * Non-PDO connectors
 * PDO connectors with a "legacy" flag enabled, which would break all the type-correctness stuff

I could do this with a SavepointManager interface as a companion to TransactionManager, or something.

src/ORM/Connect/PDOConnector.php
sminnee: I don't necessarily disagree, although I should point out a previous discussion where you suggested that protected variables are part of our semver-governed "Public API". I'm worried about doing both.

I think that we should either treat protected properties that aren't explicitly documented as public API as "internal", or we use privates more often.

src/ORM/Connect/PDOConnector.php
sminnee: :+1:

.travis.yml
sminnee: one of the builds is7.2 / PDO / MySQL

src/ORM/Connect/PDOConnector.php
sminnee: Does removing this cause more strict linters to bark, though?

sminnee:
@kinglozzer this c34c73f4ceea10a9e12d13745cebea1055c81b19 restores savepoint functionality, but if you want to use it on a PDO database, then you need to set this yaml config:

```
SilverStripe\ORM\Connect\PDOConnector:
  legacy_types: true
```

In effect it reverses the effect of this PR.

src/ORM/Connect/PDOConnector.php
robbieaverill: Yeah I guess it's a topic for another conversation, but at the moment our codebase is almost all protected over private so I guess we should keep it consistent?

.travis.yml
robbieaverill: Ah sweet, I looked but didn't see

src/ORM/Connect/PDOConnector.php
robbieaverill: It causes the slevomat standard to, yeah, but that'd be an issue to raise with them - this is fine in terms of the (proposed) PSR-5 spec =)

robbieaverill: Actually, sounds like that standard prevents you from doing this anyway: https://github.com/slevomat/coding-standard/issues/399

src/ORM/Connect/MySQLDatabase.php
sminnee: No, it's got a getter that subclasses should use.

kinglozzer:
Awesome, thanks @sminnee

src/ORM/Connect/PDOQuery.php
```diff
@@ -22,16 +22,52 @@ class PDOQuery extends Query
      * Hook the result-set given into a Query class, suitable for use by SilverStripe.
      * @param PDOStatement $statement The internal PDOStatement containing the results
      */
-    public function __construct(PDOStatement $statement)
+    public function __construct(PDOStatement $statement, PDOConnector $conn)
```
kinglozzer: Is `DB::get_conn()->getDriver()` an acceptable alternative to changing this method signature? I don’t feel too strongly about it given this is fairly internal

src/ORM/Connect/PDOQuery.php
```diff
@@ -22,16 +22,52 @@ class PDOQuery extends Query
...
+        }
+
+        // Re-map fetched data using columnMeta
+        return array_map(
+            function ($rowArray) use ($columnMeta) {
+                $row = [];
+                foreach ($columnMeta as $i => $meta) {
+                    // Coerce floats from string to float
+                    // PDO PostgreSQL fails to do this
+                    if (isset($meta['native_type']) && preg_match('/^float/', $meta['native_type'])) {
```
kinglozzer: Given this will be called on an unknown number of rows for each query, could/should we replace this with `strpos($meta['native_type'], 'float') === 0`?

src/ORM/Connect/PDOQuery.php
```diff
@@ -22,16 +22,52 @@ class PDOQuery extends Query
...
+        } else {
+            $this->results = $statement->fetchAll(PDO::FETCH_ASSOC);
+        }
         $statement->closeCursor();
     }
 
+    /**
+     * Fetch a record form the statementw ith its type data corrected
+     * Necessary to fix float data retrieved from PGSQL
+     */
```
kinglozzer: Missing param/return types

src/ORM/Connect/PDOQuery.php
sminnee: I'm pretty sure that preg_match is the fastest match. But we could also potentially get a list of all the options.

sminnee: I stand corrected:
```
php test-preg.php 
preg... 3.5830228328705
strpos... 1.2603948116302
in_array... 1.4446079730988
```

```php
<?php

define('LOTS', 10000000);

profile('preg', function() {
  for ($i=0; $i<LOTS; $i++) {
    preg_match('/^float/', 'float8');
    preg_match('/^float/', 'float16');
    preg_match('/^float/', 'varchar');
    preg_match('/^float/', 'int');
  }
});

profile('strpos', function() {
  for ($i=0; $i<LOTS; $i++) {
    strpos('float8', 'float');
    strpos('float16', 'float');
    strpos('varchar', 'float');
    strpos('int', 'float');
  }
});

profile('in_array', function() {
    $array = ['float8', 'float16', 'float4'];
  for ($i=0; $i<LOTS; $i++) {
    in_array('float8', $array);
    in_array('float16', $array);
    in_array('varchar', $array);
    in_array('int', $array);
  }
});

function profile($label, $callback) {
    echo $label . "... ";
    $t1 = microtime(true);
    $callback();
    $t2 = microtime(true);
    echo ($t2-$t1) . "\n";
}
```

sminnee: Still pretty close; i'm going with === 0 as it reduces the risk that i'm missing an entry in my array ;)

```
preg... 4.3494780063629
strpos... 1.7755749225616
in_array... 1.6038949489594
```

sminnee:
OK @kinglozzer I've rebased and addressed a couple of points of your feedback, but left the changes to PDOQuery::__construct()'s method signature in place, as I thought that avoiding access to global state was more important than preserving this method signature. Notably, doing DB::get_conn() would block the creation of more than on dbConn object which is currently possible.

sminnee:
It would be cool if we could merge https://github.com/silverstripe/silverstripe-postgresql/pull/91 so that I can retrigger these tests and get a green build before we merge.

kinglozzer:
Thanks @sminnee! I’ve merged https://github.com/silverstripe/silverstripe-postgresql/pull/91 and will restart the build, if you can write up some brief docs then I think we’re good to merge

sminnee:
Cool, so regarding docs, does this make sense?

 * DocBlock documentation for the TransactionManager interface
 * DocBlock documentation about the return types in the Query class
 * A "PHP Types" section in https://docs.silverstripe.org/en/4/developer_guides/model/sql_select/

Anything else?

kinglozzer:
I think that just about covers it @sminnee, maybe a quick note in the changelog linking to the “PHP Types” section would be good too. Will leave this a few days anyway to give others the chance to review 👍

sminnee:
OK @kinglozzer and @silverstripe/core-team - docs added, tests passing, ready to merge for 4.4.x-dev from my perspective!

kinglozzer:
🎉 nice work @sminnee!

PR #8451, opened 2018-10-04
NEW Add memory cache to ThemeResourceLoader::findTemplate()

robbieaverill:
For large sites, this reduces the number of file_exists, Path::join, getThemePaths etc calls and can contribute to 26% reduction in TTFB load time

For example, a comparison from Blackfire showing a large site's reduction in time for a 0.1% gain in memory use:

image

### DIFF
--- a/src/View/ThemeResourceLoader.php
+++ b/src/View/ThemeResourceLoader.php
@@ -350,4 +350,12 @@ class ThemeResourceLoader
         }
         return $paths;
     }
+
+    /**
+     * Clear any internally cached data from memory
+     */
+    public static function flushCache()
+    {
+        static::$cacheData = [];
+    }
 }
--- a/src/View/ThemeResourceLoader.php
+++ b/src/View/ThemeResourceLoader.php
@@ -3,13 +3,16 @@
 namespace SilverStripe\View;
 
 use InvalidArgumentException;
+use Psr\SimpleCache\CacheInterface;
+use SilverStripe\Core\Flushable;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Manifest\ModuleLoader;
 use SilverStripe\Core\Path;
 
 /**
  * Handles finding templates from a stack of template manifest objects.
  */
-class ThemeResourceLoader
+class ThemeResourceLoader implements Flushable
 {
 
     /**
@@ -38,6 +41,11 @@ class ThemeResourceLoader
      * @var ThemeList[]
      */
     protected $sets = [];
+
+    /**
+     * @var CacheInterface
+     */
+    protected $cache;
 
     /**
      * @return ThemeResourceLoader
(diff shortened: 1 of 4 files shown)


src/View/ThemeResourceLoader.php
```diff
@@ -174,6 +181,12 @@ public function getPath($identifier)
      */
     public function findTemplate($template, $themes = null)
     {
+        // Look for a cached result for this data set
+        $cacheKey = json_encode($template) . json_encode($themes);
+        if (isset(static::$cacheData[$cacheKey])) {
+            return static::$cacheData[$cacheKey];
+        }
```
micmania1: Instead of storing these in memory per session can we implement a proper caching interface? At the very least, if we have an in-memory cache then we get the same results as this and at best we can use APCu and share the same cache for all requests.

ScopeyNZ:
I think this only affects running in "dev" environments right?

src/View/ThemeResourceLoader.php
micmania1: Note: You might want to ensure the cache isn't being applied in dev mode

src/View/ThemeResourceLoader.php
```diff
@@ -174,6 +181,12 @@ public function getPath($identifier)
      */
     public function findTemplate($template, $themes = null)
```
maxime-rainville: Can we add a unit test scenario where we call `findTemplate` twice in a row with the same arguments to make sure the method works correctly when it hits the cache?

src/View/ThemeResourceLoader.php
```diff
@@ -174,6 +181,12 @@ public function getPath($identifier)
...
     {
+        // Look for a cached result for this data set
+        $cacheKey = json_encode($template) . json_encode($themes);
```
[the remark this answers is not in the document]
robbieaverill: Yeah I agree it's pretty edgey. Usually the `$themes` argument is only supplied internally, and it defaults to `SSViewer::get_themes()` so it's going to be the same 99% of the time

robbieaverill:
I'm going to:

* Check that this actually improves things in test/prod mode (@ScopeyNZ's request) otherwise close this PR if it's a dev mode only change
* Make this a proper cache implementation instead of a memory cache (@micmania1's request)
* Add a unit test to prove that its being cached (@maxime-rainville's request)

src/View/ThemeResourceLoader.php
robbieaverill: >Note: You might want to ensure the cache isn't being applied in dev mode

That'd probably be the same as the config/template/class manifests being cached, which devs are used to flushing when they change things. I don't think that's a problem here

robbieaverill: I've implemented the regular SS cache API instead of a static memory cache

maxime-rainville: Part of me desperately wants to find an MD5 collision, here!!!

src/View/ThemeResourceLoader.php
```diff
@@ -3,20 +3,30 @@
...
      * @var ThemeResourceLoader
      */
     private static $instance;
 
+    /**
+     * Internal memory cache for large sets of repeated calls
+     *
+     * @var array
+     */
+    protected static $cacheData = [];
```
micmania1: @robbieaverill should this have been removed?

robbieaverill: Yep, soz. PR incoming

robbieaverill: #8463

PR #8453, opened 2018-10-04
FIX: Update “original” DataObject data to be the content of the last write

### LINKED ISSUE #3821: [ORM] DataObject->getChangedFields not comparing values to original record
Upating a DataObject property triggers an update to to the changed record of that property.
However when checking on record changes using getChangedFields(), if that property has been restored to its original value it is still marked as having changed.  This triggers unnecessary database write actions.

Behaviour test:
https://gist.github.com/svandragt/42863fdc18e7062955c0

A practical use case is to null a value prior to a foreach loop that might set the value, in an implementation of a sync method. When the data source no longer contains the value the property should be set to null, otherwise to the value of that property. Because prior to the loop the property is nulled and then again set to the record's value, it is marked as changed unnecessary.

The solution would involve adding a conditional comparison that excludes any change where the value is now identical to the original record.

I've observed this behaviour in 3.0 and 3.1

### LINKED ISSUE #4561: [ORM] DataObject->original should be updated after write
Currently the `$original` Variable is only only populated once (in the constructor and when loading lazyfields).

The problem of this is, that the result of `getChangedFields()` will always be generated against the first value.

For example:

``` php
$obj = new MyObject();
$obj->Value = 'hi';

var_dump($obj->getChangedFields(false, 2)['Value']['before']); // NULL
$obj->write();

$obj->Value = 'hello';
var_dump($obj->getChangedFields(false, 2)['Value']['before']); // NULL, expected: hi
```

### LINKED ISSUE #8443: Original values are not updated after a write
# TL;DR
A `DataObject` should have the same state after a `write` that it would having been freshly `get`

## Affected Version

4.3.x-dev 3b7802bb51ee32432ed8b45a722275f5243ffcbf
Although probably affecting versions back near enough to the [beginning of time](https://github.com/silverstripe/silverstripe-framework/blob/2.4/core/model/DataObject.php).

## Description

Changes are tracked in two ways; one by updating a `changed` private array, and another via a copy of `record` kept in the field `original`, set at `__construct` - both ways can be accessed via the `getChangedFields` method.
After calling `write` one would expect to have the local instance of the object updated to match that which would be produced by a brand new query for it via the active record controls (`get`). However this is not the case, and one cannot, for example, look up the value a particular field had before it was changed in order to reset it in `onBeforeWrite`.

## Steps to Reproduce

1. Code a `DataObject` subclass. It could be empty (no extra DB fields) - it's not important the content.
2. `dev/build`
3. `create` one of these objects
4. `update` the `Title`
5. `write`
6. `update` the `Title`
7. `write`
8. `update` the `Title`
9. `write`
10. `update` the `Title`
11. call `getChangedFields('Title')`
12. **behold the `before` value of `null`.**
13. use that existing object's ID to `get` a new instance of it from the database.
14. repeat step 10, but on the new instance.
15. repeat step 11, but on the new instance.
16. behold the `before` value of that which was saved at step 9.

sminnee:
FIX: Compare to original when determining fields changes

This fixes a number of edge-case issues relating to change detection.

Fixes #8443
Fixes #3821
Fixes #4561

### DIFF
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -167,6 +167,11 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
     const CHANGE_NONE = 0;
 
     /**
+     * Represents a field that has had a change forced
+     */
+    const CHANGE_FORCED = 0.5;
+
+    /**
      * Represents a field that has changed type, although not the loosely defined value.
      * (before !== after && before == after)
      * E.g. change 1 to true or "true" to true, but not true to 0.
@@ -1126,7 +1131,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 
         foreach ($fieldNames as $fieldName) {
             if (!isset($this->changed[$fieldName])) {
-                $this->changed[$fieldName] = self::CHANGE_STRICT;
+                $this->changed[$fieldName] = self::CHANGE_FORCED;
             }
             // Populate the null values in record so that they actually get written
             if (!isset($this->record[$fieldName])) {
@@ -1520,7 +1525,11 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
             // If there's any relations that couldn't be saved before, save them now (we have an ID here)
             $this->writeRelations();
             $this->onAfterWrite();
-            $this->changed = array();
+
+            // Reset isChanged data
+            // DBComposites properly bound to the parent record will also have their isChanged value reset
+            $this->changed = [];
+            $this->original = $this->record;
         } else {
             if ($showDebug) {
                 Debug::message("no changes for DataObject");
@@ -2487,7 +2496,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
     }
 
     /**
-     * Return the fields that have changed.
+     * Return the fields that have changed since the last write.
      *
      * The change level affects what the functions defines as "changed":
      * - Level CHANGE_STRICT (integer 1) will return strict changes, even !== ones.
@@ -2628,22 +2637,25 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
             }
 
             // if a field is not existing or has strictly changed
-            if (!isset($this->record[$fieldName]) || $this->record[$fieldName] !== $val) {
+            if (!isset($this->original[$fieldName]) || $this->original[$fieldName] !== $val) {
                 // TODO Add check for php-level defaults which are not set in the db
                 // TODO Add check for hidden input-fields (readonly) which are not set in the db
                 // At the very least, the type has changed
                 $this->changed[$fieldName] = self::CHANGE_STRICT;
 
-                if ((!isset($this->record[$fieldName]) && $val)
-                    || (isset($this->record[$fieldName]) && $this->record[$fieldName] != $val)
+                if ((!isset($this->original[$fieldName]) && $val)
+                    || (isset($this->original[$fieldName]) && $this->original[$fieldName] != $val)
                 ) {
                     // Value has changed as well, not just the type
                     $this->changed[$fieldName] = self::CHANGE_VALUE;
                 }
-
-                // Value is always saved back when strict check succeeds.
-                $this->record[$fieldName] = $val;
+            // Value has been restored to its original, remove any record of the change
+            } elseif (isset($this->changed[$fieldName]) && $this->changed[$fieldName] !== self::CHANGE_FORCED) {
+                unset($this->changed[$fieldName]);
             }
+
+            // Value is saved regardless, since the change detection relates to the last write
+            $this->record[$fieldName] = $val;
         }
         return $this;
     }
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -167,11 +167,6 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
     const CHANGE_NONE = 0;
 
     /**
-     * Represents a field that has had a change forced
-     */
-    const CHANGE_FORCED = 0.5;
-
-    /**
      * Represents a field that has changed type, although not the loosely defined value.
      * (before !== after && before == after)
      * E.g. change 1 to true or "true" to true, but not true to 0.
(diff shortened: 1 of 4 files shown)


sminnee:
OK test-duration changes seem to be within Travis' variance (8% slower than most recent build on PHP <= 7.2, 3% faster on 7.3, but looking at previous 4 builds, it seems like the most recent build was faster than typical)

Final thing is to add some tests for the specific bugs fixed by this change.

sminnee:
@beerbohmdo @svandragt @NightJar @dhensby do you want to see if this fully addresses the issues raised in the 3 tickets linked in the ticket description?

sminnee:
Added Dan/Damian to the review request list; 1 of them would be good to get input from given the centrality/age of the relevant code.

NightJar:
Thanks @sminnee I'll have a play tonight and report back. Looks like you've covered all bases though :)

NightJar:
Gave it a run through, does as I expect :)
Solves both:
1. Setting `$original` after a successful `writeManipulation` operation
2. Resetting `$changed` values on the case of `$newValue === $oldValue`
And also solves a case I hadn't considered:
- Preserves force write... in the case of point 2 above occurring between giving the directive for a forcing a change, and actually writing (e.g. some onBeforeWrite modifies the value).

However, since you've got me thinking about forced changes...
Consider the following:
```php
<?php

namespace SilverStripe\ORM\Tests\DataObjectTest;

use SilverStripe\Dev\TestOnly;
use SilverStripe\ORM\DataObject;

class Enforcer extends DataObject implements TestOnly
{
    private static $db = ['Title' => 'Varchar'];
}
```
```yml
SilverStripe\ORM\Tests\DataObjectTest\Enforcer:
  tester:
    Title: Original
```
```php
<?php

use SilverStripe\ORM\DataObject;
use SilverStripe\ORM\Tests\DataObjectTest\Enforcer;

class DataObjectResetTest extends SapphireTest
{
    protected static $fixture_file = 'DataObjectResetTest.yml';

    protected static $extra_dataobjects = [
        Enforcer::class,
    ];

    private function expectChanges($from, $to, $type)
    {
        return [
            'Title' => [
                'before' => $from,
                'after' => $to,
                'level' => $type
            ]
        ];
    }
    
    public function testObjectChangesStartEmpty()
    {
        // Simply a set up & sanity check
        $object = $this->objFromFixture(Enforcer::class, 'tester');
        $this->assertEquals('Original', $object->Title);
        $this->assertSame(
            [],
            $object->getChangedFields(['Title']),
            'No changes to begin with'
        );
    }

    public function testForceChangesForceChanges()
    {
        $object = $this->objFromFixture(Enforcer::class, 'tester');

        $object->forceChange();

        $this->assertSame(
            $this->expectChanges('Original', 'Original', DataObject::CHANGE_FORCED),
            $object->getChangedFields(['Title']),
            'Title is set to forced change'
        );
    }

    public function testSettingTheOriginalValueOverTheOrignalValueDoesNotCancelAForcedChange()
    {
        $object = $this->objFromFixture(Enforcer::class, 'tester');

        $object->forceChange();
        $object->Title = 'Original';

        $this->assertSame(
            $this->expectChanges('Original', 'Original', DataObject::CHANGE_FORCED),
            $object->getChangedFields(['Title']),
            'Setting a value to the same value does not remove the forced change flag'
        );
    }

    // OK neat, that's all good up to here.

    // But... A challenger appears!

    // Challenger uses Update

    public function testForcingAChangeOnANewValueStillForcesAChange()
    {
        $object = $this->objFromFixture(Enforcer::class, 'tester');

        $object->Title = 'Updated!';
        $object->forceChange();

        $this->assertSame(
            $this->expectChanges('Original', 'Updated!', DataObject::CHANGE_FORCED),
            $object->getChangedFields(['Title']),
            'Forcing a change on a new value still forces change'
        );
    }

    // Object's DEFENSE greatly fell!

    // Challenger uses Shuffle

    public function testChangingAValueAfterForcingAChangeMaintainsTheForcedChangeFlag()
    {
        $object = $this->objFromFixture(Enforcer::class, 'tester');

        $object->forceChange();
        $object->Title = 'Updated!';

        $this->assertSame(
            $this->expectChanges('Original', 'Updated!', DataObject::CHANGE_FORCED),
            $object->getChangedFields(['Title']),
            'Changing a value after forcing a change remains flagged to be forced'
        );
    }

    // It's super effective!

    // Challenger uses Confuse

    public function testResettingAForcedChangeMaintainsTheForcedChangeFlag()
    {
        $object = $this->objFromFixture(Enforcer::class, 'tester');

        $object->Title = 'Updated!';
        $object->forceChange();
        $object->Title = 'Original';

        $this->assertSame(
            $this->expectChanges('Original', 'Original', DataObject::CHANGE_FORCED),
            $object->getChangedFields(['Title']),
            'Resetting a value after issuing the force directive does not negate the force flag'
        );
    }

    // Object is close to fainting!

    // Challenger uses Reset

    public function testForceChangeCannotBeCircumvented()
    {
        $object = $this->objFromFixture(Enforcer::class, 'tester');

        $object->forceChange();
        $object->Title = 'Updated!';
        $object->Title = 'Original';

        $this->assertSame(
            $this->expectChanges('Original', 'Original', DataObject::CHANGE_FORCED),
            $object->getChangedFields(['Title']),
            'Setting, then resetting a value does not circumvent a previously issued force change directive'
        );
    }

    // Object has fainted!
}
```

sminnee:
Thanks for the set of test cases, but could you clarify which of those tests still fail under this PR — just the last one?

I could point the PR at 4.2 instead but for a bug that's minor enough to have been left unfixed for years I think that merging it into 4.0 and merging forward 3 other versions is a bit of a waste of time.

NightJar:
Here is a commit you should be able to `cherry-pick` into your branch :) Forewarning: I've not tested it because I run SQLite locally & it doesn't play nice with `DataObjectTest` (but the code in my previous comment was fully working, and sat beside `DataObjectTest` in folder structure).
I was able to cherry pick your branch into `4.0`, so hopefully we can re-target the fix & merge it up through `4.1` & `4.2`.

Every test from `    // OK neat, that's all good up to here.` fails - the last 4.

sminnee:
In order to address the deeper test that NightJar raised, I think I'm going to create a separate private `$this->changesForced` boolean, which will alter the output of getChangedFields().

sminnee:
@kinglozzer since you're on a roll with PR-review... ;-)

NightJar:
I'll take another squizz tonight :)

NightJar:
Yep, covers all my cases, and also nicely maintains change level (which it didn't before). Nice :)
Click the green button to give it the :purple_heart: @kinglozzer

NightJar:
Next up: https://github.com/silverstripe/silverstripe-assets/issues/169

PR #8456, opened 2018-10-05
Add support for null values in template lookups

kinglozzer:


sminnee:
Should this be master or `4`, @kinglozzer? I don't think it's breaking?

kinglozzer:
@sminnee it builds upon #8152, which was merged into master. IIRC I took that decision mainly to “play it safe”, though it may be possible to backport one or both of these PRs to `4`... I definitely don’t think `SSViewer_DataPresenter` counts as part of our public API 😉

sminnee:
OK then I'll merge to master for now.

From my perspective both of these changes could be merged back to `master`; perhaps the two of them cherry-picked as a single PR?

PR #8458, opened 2018-10-05
FIX: Don’t break validation on selects without a source.

### LINKED ISSUE #4849: BUG Empty DropdownField can't be submitted if it has no values
DropdownField::validate() should validate true if the source is empty.

Spotted at https://github.com/silverstripe/silverstripe-subsites/issues/223

### LINKED ISSUE #7159: Dropdown field where "none" is the only option blocks form from ever submitting
If you have a dropdown with a source such as `[ "" => "(none)" ]`, it won't let the form ever submit. You'll get an error like:

"Please select a value within the list provided. (none) is not a valid option"

Now, it's probably better UX to hide the field in such circumstances, but I think it's still a bug for DropdownField to break in this way. It's also a regression from 3.x that will easily trip up developers as it can hide in the corners of upgraded projects.

sminnee:
Fixes cases where there are no valid values or an empty-string
value is manually added rather than using setEmptyString()

Fixes #4849
Fixes #7159

### DIFF
--- a/tests/php/Forms/DropdownFieldTest.php
+++ b/tests/php/Forms/DropdownFieldTest.php
@@ -505,7 +505,7 @@ class DropdownFieldTest extends SapphireTest
      */
     public function testRequiredDropdownHasEmptyDefault()
     {
-        $field = new \SilverStripe\Forms\DropdownField("RequiredField", "dropdown", ["item 1", "item 2"]);
+        $field = new DropdownField("RequiredField", "dropdown", ["item 1", "item 2"]);
 
         $form = new Form(
             Controller::curr(),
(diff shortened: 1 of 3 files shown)


tests/php/Forms/DropdownFieldTest.php
```diff
@@ -517,4 +517,33 @@ public function testRequiredDropdownHasEmptyDefault()
...
+        $v = new RequiredFields();
+        $field->validate($v);
+        $this->assertTrue($v->getResult()->isValid());
+
+        // Source with all items disabled
+        $field = new DropdownField("SourceWithBlankVal", "", [ "A" => "A", "B" => "B" ]);
+        $field->setDisabledItems([ 'A', 'B' ]);
+        $v = new RequiredFields();
+        $field->validate($v);
+        $this->assertTrue($v->getResult()->isValid());
```
robbieaverill: To be finicky, these could be separate test methods

PR #8460, opened 2018-10-08
API Add a new GridFieldLazyLoader component

maxime-rainville:
Add a new GridField component that can be applied to a gridfield to enable it to lazy load when it's in a tab set. You also need to be running https://github.com/silverstripe/silverstripe-admin/pull/679 for the frontend bits to work.

For this to take affect you just need to add the GridFieldLazyLoader component to your GridField config.

# Parent issue
* https://github.com/silverstripe/silverstripe-cms/issues/2276

### DIFF
--- a/src/Forms/GridField/GridFieldLazyLoader.php
+++ b/src/Forms/GridField/GridFieldLazyLoader.php
@@ -5,6 +5,7 @@ namespace SilverStripe\Forms\GridField;
 use SilverStripe\Forms\FormField;
 use SilverStripe\Forms\TabSet;
 use SilverStripe\ORM\ArrayList;
+use SilverStripe\ORM\Filterable;
 use SilverStripe\ORM\Limitable;
 use SilverStripe\ORM\SS_List;
 
@@ -22,11 +23,11 @@ class GridFieldLazyLoader implements GridField_DataManipulator, GridField_HTMLPr
      */
     public function getManipulatedData(GridField $gridField, SS_List $dataList)
     {
-        // If we are lazy loading, empty the list
+        // If we are lazy loading an empty the list
         if ($this->isLazy($gridField)) {
-            if ($dataList instanceof Limitable) {
-                // If our original list can be limited, set the limit to 0.
-                $dataList = $dataList->limit(0);
+            if ($dataList instanceof Filterable) {
+                // If our original list can be filtered, filter out all results.
+                $dataList = $dataList->byIDs([-1]);
             } else {
                 // If not, create an empty list instead.
                 $dataList = ArrayList::create([]);
--- a/src/Forms/GridField/GridFieldLazyLoader.php
+++ b/src/Forms/GridField/GridFieldLazyLoader.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace SilverStripe\Forms\GridField;
+
+use SilverStripe\ORM\ArrayList;
+use SilverStripe\ORM\SS_List;
+
+/**
+ * GridFieldLazyLoader alters the {@link GridField} behavior to delay rendering of rows until the tab containing the
+ * GridField is selected by the user.
+ *
+ * @see GridField
+ */
+class GridFieldLazyLoader implements GridField_DataManipulator, GridField_HTMLProvider
+{
+
+    /**
+     * @inheritDoc
+     */
+    public function getManipulatedData(GridField $gridField, SS_List $dataList)
+    {
+        return $this->isLazy($gridField) ?
+            ArrayList::create([]) :
+            $dataList;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getHTMLFragments($gridField)
+    {
+        $gridField->addExtraClass($this->isLazy($gridField) ?
+            'grid-field-lazy-loadable' :
+            'grid-field-lazy-loaded');
+        return [];
+    }
+
+    /**
+     * Detect if the current request should include results
+     * @param GridField $gridField
+     * @return bool
+     */
+    private function isLazy(GridField $gridField)
+    {
+        return $gridField->getRequest()->getHeader('X-Pjax') !== 'CurrentField';
+    }
+}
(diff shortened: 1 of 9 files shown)


src/Forms/GridField/GridFieldLazyLoader.php
```diff
@@ -0,0 +1,78 @@
...
+                // If not, create an empty list instead.
+                $dataList = ArrayList::create([]);
+            }
+        }
+        return $dataList;
+    }
+
+    /**
+     * @inheritDoc
+     */
```
robbieaverill: This is the same as not having a doc block at all - may as well save three lines =D

src/Forms/GridField/GridFieldLazyLoader.php
```diff
@@ -0,0 +1,78 @@
...
+    public function getHTMLFragments($gridField)
+    {
+        $gridField->addExtraClass($this->isLazy($gridField) ?
+            'grid-field-lazy-loadable' :
+            'grid-field-lazy-loaded');
```
robbieaverill: Should these be BEM convention since `loadable/loaded` seem like modifiers? `grid-field--lazy-loadable` `grid-field--lazy-loaded`

src/Forms/GridField/GridFieldLazyLoader.php
```diff
@@ -0,0 +1,78 @@
...
+     * @return bool
+     */
+    private function isInTabSet(FormField $field)
+    {
+        $list = $field->getContainerFieldList();
+        if ($list && $containerField = $list->getContainerField()) {
+            // Classes that extends TabSet might not have the expected JS to lazy load.
+            return get_class($containerField) === TabSet::class
+                ?: $this->isInTabSet($containerField);
+        } else {
```
robbieaverill: The else is redundant here

tests/php/Forms/GridField/GridFieldLazyLoaderTest.php
```diff
@@ -0,0 +1,153 @@
...
+        $gridFied = $this->getOutOfTabSetGridField();
+        $actual = $this->component->getHTMLFragments($gridFied);
+        $this->assertEmpty($actual, 'getHTMLFragments should always return an array');
+        $this->assertContains('grid-field-lazy-loaded', $gridFied->extraClass());
+        $this->assertNotContains('grid-field-lazy-loadable', $gridFied->extraClass());
+    }
+
+    public function testGetHTMLFragmentsNonLazy()
+    {
+        $gridFied = $this->getNonLazyGridField();
```
robbieaverill: Minor typo

src/Forms/GridField/GridFieldLazyLoader.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+namespace SilverStripe\Forms\GridField;
+
+use SilverStripe\Forms\FormField;
+use SilverStripe\Forms\TabSet;
+use SilverStripe\ORM\ArrayList;
+use SilverStripe\ORM\Filterable;
+use SilverStripe\ORM\Limitable;
```
robbieaverill: This doesn't look like it's used any more

src/Forms/GridField/GridFieldLazyLoader.php
```diff
@@ -0,0 +1,78 @@
...
+            $gridField->getRequest()->getHeader('X-Pjax') !== 'CurrentField' &&
+            $this->isInTabSet($gridField);
+    }
+
+    /**
+     * Recursively check if $field is inside a TabSet.
+     * @param FormField $field
+     * @return bool
+     */
+    private function isInTabSet(FormField $field)
```
robbieaverill: Can we please use protected over private?

maxime-rainville: I view this more as an implementation detail. 

This bit of logic is tightly coupled with the front-end JS in admin. If developers try to override `isInTabSet`, it's almost guaranteed to break the front end logic.

My preference would be to keep it private.

src/Forms/GridField/GridFieldLazyLoader.php
maxime-rainville: I've added insightful comments about this method.

src/Forms/GridField/GridFieldLazyLoader.php
maxime-rainville: Done. Matching changes have been made to `silverstripe/admin` as well.

src/Forms/GridField/GridFieldLazyLoader.php
maxime-rainville: Done.

tests/php/Forms/GridField/GridFieldLazyLoaderTest.php
maxime-rainville: Done.

src/Forms/GridField/GridFieldLazyLoader.php
maxime-rainville: Removed.

maxime-rainville:
Everything should be working now.

The problem was that when performing a readonly transformation on a form, the new FieldList for composite fields where not assigned a Container Field. I've updated `CompositeField::setChildren()` so it sets the `ContainerField` for its new FieldList.

src/Forms/GridField/GridFieldLazyLoader.php
robbieaverill: Suspicious, but ok

robbieaverill:
Needs tests fixed, will update now

robbieaverill:
This makes such a profound difference. Nice!

robbieaverill:
I found one bug, where if you toggle to a page's Dependent Pages tab, then switch to another tab, when it remembers that you were on the Dependent Pages tab and loads that tab again immediately, it shows no records. You have to toggle out of it and back into it again to get it to lazy load. I don't think this blocks merge, but I'll raise another issue for it to follow up as a lower impact bug.

PR #8461, opened 2018-10-08
NEW Adding a helper to find a form field by label content (Behat)

ScopeyNZ:
This allows step authors to use `$this->findInputByLabelContent("Label")` to find the related input. It searches for the label and uses the `for` attribute to find an ID selector for the field.

Additionally I've added this to the `getHtmlField` method for a fallback if the field was not found using the "name" of the field - this can be quite specfic when there are nested forms.

### DIFF
--- a/tests/behat/src/CmsFormsContext.php
+++ b/tests/behat/src/CmsFormsContext.php
@@ -310,9 +310,36 @@ JS;
     {
         $locator = $this->fixStepArgument($locator);
         $page = $this->getSession()->getPage();
+        
+        // Searching by name is usually good...
         $element = $page->find('css', 'textarea.htmleditor[name=\'' . $locator . '\']');
+        
+        if ($element === null) {
+            $element = $this->findInputByLabelContent($locator);
+        }
+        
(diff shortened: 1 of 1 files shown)


tests/behat/src/CmsFormsContext.php
```diff
@@ -310,11 +310,38 @@ protected function getHtmlField($locator)
...
+        $page = $this->getSession()->getPage();
+        $label = $page->findAll('xpath', sprintf('//label[contains(text(), \'%s\')]', $locator));
+
+        if (empty($label)) {
+            return null;
+        }
+
+        assertCount(1, $label, sprintf(
+            'Found more than one element containing the phrase "%s".',
+            $locator
```
robbieaverill: This seems restrictive. Can you just return the first match instead of failing?

ScopeyNZ: I considered that but:
- I thought it would just be confusing for the implementor if it was matching the wrong field, doing something like typing into it and your test was failing
- If a new field is added the problem would me immediately apparent.
- The fix would probably be to make the match more specific anyway.

PR #8462, opened 2018-10-08
FIX: Make all enums non-destructive, not just ClassName

sminnee:
This change also renders a portion of DBSchemaManager irrelevant, that
destructively “fixes” old values. This is in keeping with the
non-destructive principle of dev/build, and some suggestions to move
away from enum fields altogether.

Fixes https://github.com/silverstripe/silverstripe-framework/issues/1387

### DIFF
--- a/src/ORM/FieldType/DBEnum.php
+++ b/src/ORM/FieldType/DBEnum.php
@@ -20,7 +20,7 @@ class DBEnum extends DBString
      *
      * @var array
      */
-    protected $enum = array();
+    protected $enum = [];
 
     /**
      * Default value
@@ -30,6 +30,22 @@ class DBEnum extends DBString
     protected $default = null;
 
     private static $default_search_filter_class = 'ExactMatchFilter';
+
+    /**
+     * Internal cache for obsolete enum values. The top level keys are the table, each of which contains
+     * nested arrays with keys mapped to field names. The values of the lowest level array are the enum values
+     *
+     * @var array
+     */
+    protected static $enum_cache = [];
+
+    /**
+     * Clear all cached enum values.
+     */
+    public static function flushCache()
+    {
+        self::$enum_cache = [];
+    }
 
     /**
      * Create a new Enum field, which is a value within a defined set, with an optional default.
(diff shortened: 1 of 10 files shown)


sminnee:
I don't think we've tested schema migration in the past but I'll see if we can dig something up.

unclecheese:
Have retargeted to `4.3`

sminnee:
@unclecheese is it something that you're particularly keen to see in 4.3? I would have expected the merge to go into 4.4 so we can avoid rocking the boat for 4.3?

unclecheese:
I was optimistic about getting this into 4.3, but now that I have a better handle on the scope of the release, I don't think it's realistic.

As you were.

src/ORM/FieldType/DBEnum.php
```diff
@@ -173,6 +189,48 @@ public function getEnum()
...
+     * If table or name are not set, or if it is not a valid field on the given table,
+     * then only known enum values are returned.
+     *
+     * Values cached in this method can be cleared via `DBEnum::clear_classname_cache();`
+     *
+     * @return array
+     */
+    public function getEnumObsolete()
+    {
+        // Without a table or field specified, we can only retrieve known classes
```
dhensby: comments are still referring to "classes" instead "enum values"

sminnee:
> However, is this really non-destructive? Isn't the order the enum is defined important to it's internal DB value? I guess not if this is working with ClassName columns... but I just wanted to be sure that this was definitely safe.

I'll double check with a manual test but I'm pretty sure that changing option order doesn't destroy anything.

sminnee:
Fixed docs, and added a test, including a specific change to the value ordering to cover off @dhensby's worry.

kinglozzer:
Needs a rebase (docs), are you happy with this now @dhensby?

sminnee:
ping

sminnee:
@dhensby or @kinglozzer, any feedback?

kinglozzer:
I’m happy with the changes and the test looks to me like it covers @dhensby’s concerns, merging

PR #8466, opened 2018-10-09
Make the DBCurrency fields decimal_point and thousands_separator conf…

TheBnl:
…igurable

Some countries have the decimal point and thousands separator switched around like €1.000,00 instead of $1,000.00


robbieaverill:
Great, thanks for your understanding

PR #8469, opened 2018-10-11
DOCS: New nested fields, args for ApolloGraphQLManager

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-admin/issues/611


docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
```diff
@@ -1080,13 +1080,67 @@ Now, let's update the query to fetch our new field.
 *my-other-app/client/transformReadNotes.js*
 ```js
 const transformReadNotes = (manager) => {
-  manager.addField('Priority');
+  manager.addField('root', 'Priority');
```
robbieaverill: This is obviously an API break, and while I know we don't really treat our React API surface with semver yet, this does seem like it would break for anyone who had read these docs in the past and used the example code. Could you switch the argument order to make it B/C?

Issue #8467, opened 2018-10-09
Can not override FallbackLocale


jmfederico:
This is related to #6805 and should be solved if the solution proposed there is executed. But this is a major problem for multilingual sites that do not require english, so I am opening a new issue.

## Affected Version
4.2.1

## Description
When trying to override the fallback locale, one should be able to create a configuration setting like this (for instance to set Spanish as fallback):
```yml
SilverStripe\Core\Injector\Injector:
  Symfony\Component\Translation\TranslatorInterface:
    calls:
      FallbackLocales: [ setFallbackLocales, [['es']]]
```

But, since the setting is an array, instead of overriding, it add the parameters to the original setting, which ends up being:
```php
array:4 [
  0 => "setFallbackLocales"
  1 => array:1 [
    0 => array:1 [
      0 => "en"
    ]
  ]
  2 => "setFallbackLocales"
  3 => array:1 [
    0 => array:1 [
      0 => "es"
    ]
  ]
]
```

Which will then throw an `InvalidArgumentException` exception.

## Steps to Reproduce
On a SilverStripe project, add this to your config.yml file:

```yml
SilverStripe\Core\Injector\Injector:
  Symfony\Component\Translation\TranslatorInterface:
    calls:
      FallbackLocales: [ setFallbackLocales, [['es']]]
```

<hr>

in the meantime, is there any way to override the value of the fallback language?

kinglozzer:
Yeah that was the gist of it, I guess in this case it would be something like

```yml
---
Name: translatorreset
---
SilverStripe\Core\Injector\Injector:
  Symfony\Component\Translation\TranslatorInterface:
    calls: null
---
Name: translator
---
SilverStripe\Core\Injector\Injector:
  Symfony\Component\Translation\TranslatorInterface:
    calls:
      FallbackLocales: [ setFallbackLocales, [['es']]]
```

jmfederico:
Great, thanks

Final solution was:

```yml
---
Name: translatorreset
---
SilverStripe\Core\Injector\Injector:
  Symfony\Component\Translation\TranslatorInterface:
    calls:
      FallbackLocales: null
---
Name: translator
---
SilverStripe\Core\Injector\Injector:
  Symfony\Component\Translation\TranslatorInterface:
    calls:
      FallbackLocales: [ setFallbackLocales, [['es']]]
```

I've created a PR with an example for these situations: #8471

PR #8471, opened 2018-10-11
Example to override a value that is an array.

jmfederico:
Add an example for how to override a configuration value that has is array.


jmfederico:
I had a hard time overriding array configurations.

According to #8467 there is a workaround, but it is not documented.

Hope this helps.

PR #8475, opened 2018-10-12
WIP: NEW Use toArray() instead of $this->items where appropriate

robbieaverill:
Allows for toArray() to be modified in child classes and have it be respected in getIterator()

At the moment child classes must overload getIterator() as well to make it use `$this->toArray()` over `$this->items`.

### DIFF
--- a/src/ORM/ArrayList.php
+++ b/src/ORM/ArrayList.php
@@ -51,8 +51,9 @@ class ArrayList extends ViewableData implements SS_List, Filterable, Sortable, L
      */
     public function dataClass()
     {
-        if (count($this->items) > 0) {
-            return get_class($this->items[0]);
+        $items = $this->toArray();
+        if (count($items) > 0) {
+            return get_class($items[0]);
         }
         return null;
     }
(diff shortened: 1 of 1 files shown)


unclecheese:
Have retargeted to `4.3`.

robbieaverill:
Rebased

robbieaverill:
>Add a note to the toArray() docblock, "This method can be overridden in subclasses to change the contents of the list".

I would've thought that was inherent in the way we design our APIs, e.g. by making things protected rather than private

unclecheese:
I think it's more a question of the naming of that method. It's not very self-documenting. It isn't obvious that a method named `toArray` on a class that is basically an array is the place I would do filtering.

Another approach might be the addition of a protected method `getIteratorItems()` defaulting to `$this->items`.

robbieaverill:
I see what you're saying - yeah, `getItems()` or something would be clearer about its intent. How about I mark this as WIP and add a new method to do this instead of overloading `toArray()`?

sminnee:
> I would've thought that was inherent in the way we design our APIs, e.g. by making things protected rather than private

This is a historic bad habit that makes it very difficult to provide sem-ver. We need to get better about clarifying, documenting, and testing our public API more than "anything technically possible within the bounds of PHP"

robbieaverill:
I'll re-open another one later if I get around to it =)

PR #8477, opened 2018-10-15
Fix search filtering relations and clear filters

lukereative:
Fixes silverstripe/silverstripe-admin#668 and fixes silverstripe/silverstripe-admin#672


maxime-rainville:
I've got a PR for Framework test that makes those edge cases easier to test https://github.com/silverstripe/silverstripe-frameworktest/pull/55

PR #8479, opened 2018-10-15
allow dataobject get_one without passing a class

andrewandante:
Allows you to call `MyObject::get_one()` instead of `DataObject::get_one(MyObject::class)` - which is a bit weird

### DIFF
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -3090,8 +3090,11 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
      *
      * @return DataObject|null The first item matching the query
      */
-    public static function get_one($callerClass, $filter = "", $cache = true, $orderby = "")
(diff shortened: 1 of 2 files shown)


src/ORM/DataObject.php
```diff
@@ -3090,8 +3090,11 @@ public static function get(
      *
      * @return DataObject|null The first item matching the query
      */
-    public static function get_one($callerClass, $filter = "", $cache = true, $orderby = "")
+    public static function get_one($callerClass = null, $filter = "", $cache = true, $orderby = "")
```
robbieaverill: I wonder if this signature change would be breaking

kinglozzer: It is 😞

andrewandante: It _is_ backwards compatible though - is that enough?

andrewandante: True - sigh. So PR against 5 then?

andrewandante:
Moved to a PR against 5 because of BC issues

PR #8480, opened 2018-10-15
allow dataobject get_one without passing a class

andrewandante:
Moved #8479 because of non-BC

### DIFF
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -3074,24 +3074,28 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
      * Return the first item matching the given query.
      * All calls to get_one() are cached.
      *
-     * @param string $callerClass The class of objects to be returned
-     * @param string|array $filter A filter to be inserted into the WHERE clause.
-     * Supports parameterised queries. See SQLSelect::addWhere() for syntax examples.
-     * @param boolean $cache Use caching
-     * @param string $orderby A sort expression to be inserted into the ORDER BY clause.
+     * @param string|null  $callerClass The class of objects to be returned. Defaults to the class that calls the method
+     *                                  e.g. MyObject::get_one() will return a MyObject
+     * @param string|array $filter      A filter to be inserted into the WHERE clause.
+     *                                  Supports parameterised queries. See SQLSelect::addWhere() for syntax examples.
+     * @param boolean      $cache       Use caching
+     * @param string       $orderBy     A sort expression to be inserted into the ORDER BY clause.
      *
      * @return DataObject|null The first item matching the query
      */
-    public static function get_one($callerClass, $filter = "", $cache = true, $orderby = "")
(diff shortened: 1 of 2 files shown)


src/ORM/DataObject.php
```diff
@@ -3082,8 +3082,11 @@ public static function get(
      *
      * @return DataObject|null The first item matching the query
      */
-    public static function get_one($callerClass, $filter = "", $cache = true, $orderby = "")
+    public static function get_one($callerClass = null, $filter = "", $cache = true, $orderby = "")
```
robbieaverill: LGTM, the docblock for the callerClass argument could do with a little tweak tho

andrewandante: Updato

src/ORM/DataObject.php
```diff
@@ -3074,24 +3074,28 @@ public static function get(
...
+     * @param boolean      $cache       Use caching
+     * @param string       $orderBy     A sort expression to be inserted into the ORDER BY clause.
      *
      * @return DataObject|null The first item matching the query
      */
-    public static function get_one($callerClass, $filter = "", $cache = true, $orderby = "")
+    public static function get_one($callerClass = null, $filter = "", $cache = true, $orderBy = "")
     {
+        if ($callerClass === null) {
+            $callerClass = static::class;
```
tractorcow: There really should be a check to make sure `static::class !== self::class`, otherwise you are querying `DataObject` potentially.

andrewandante: So that would mean if you called `DataObject::get_one()` it would... throw an exception? We could enforce `DataObject::get_one(DataObject::class)`, but that seems a little clunky to me.

The more I think about it, the more I want to completely re-write the method - given that this is going into 5 now and not 4, I think we can probably straight remove the `$callerClass` parameter, don't you think? I see no reason why we need to leave it this way - are there any?

andrewandante: _I've started having bold ideas about removing the filter, sort, join etc from `DataObject::get()` entirely..._

Would it be so bad to entirely remove those and require chaining instead? You can see inside the `get_one()` method that chaining is used anyway :)

andrewandante: Oh - and I've realised that if you call `DataObject::get_one()`, it will just call `DataObject::get(DataObject::class)`, and hit the validation there anyway - do we need to catch it earlier than that?

robbieaverill: I think that's what @tractorcow is suggesting you catch early. Personally I'd be in favour of deprecating support for the argument based filters (in favour of DataList chaining) but I feel like maybe that should go through an RFC process separately

andrewandante: Most likely - I've just taken a quick look at the implications of it, but I think it's a good change. It essentially turns `MyObject::get()` into `DataList::create(MyObject::class)` - but I think that's ok. Anyway, that's an aside.

I'll add an earlier catch :)

PR #8485, opened 2018-10-16
Update admin example code

DorsetDigital:
Added an additional config block to the example code to set the AdminRootController.url_base config option as well and updated the description.


robbieaverill:
So did this fix your bug in https://github.com/silverstripe/silverstripe-admin/issues/689?

DorsetDigital:
No, this didn't fix that issue.  It did fix an unrelated problem I was having with the site I was working on (where an add-on wasn't correctly identifying the admin route directly from the AdminRootController config, rather than via `get_admin_route()` ), so I felt it was prudent to add it to the docs anyway.

PR #8490, opened 2018-10-17
NEW Some minor refactoring of the PDO and MySQLi connectors

robbieaverill:
Some small performance optimisations (not using callables in loops, switch strval for string
casting), Config call updates and replace call_user_func_array with direct variadic call.
Also removes some redundant else statements after returns.

EA Inspections report that `(string) $value` is up to 6x more performant than `strval($value)`. There are no functional changes in this PR.

### DIFF
--- a/src/ORM/Connect/PDOConnector.php
+++ b/src/ORM/Connect/PDOConnector.php
@@ -106,7 +106,7 @@ class PDOConnector extends DBConnector
      */
     public static function is_emulate_prepare()
     {
-        return Config::inst()->get('SilverStripe\ORM\Connect\PDOConnector', 'emulate_prepare');
+        return self::config()->get('emulate_prepare');
     }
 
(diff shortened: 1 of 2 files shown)


src/ORM/Connect/PDOConnector.php
```diff
@@ -338,7 +339,7 @@ public function bindParameters(PDOStatement $statement, $parameters)
             // Check type of parameter
             $type = $this->getPDOParamType($phpType);
             if ($type === PDO::PARAM_STR) {
-                $value = strval($value);
+                $value = (string) $value;
```
sminnee: Do we normally do `(string) $value` or `(string)$value`? i.e. should the space be there?

sminnee: ...the space, as is here, seems more common.

PR #8492, opened 2018-10-17
MINOR Correct implementation of single lookup

maxime-rainville:
`SingleLookupField::Value()` craps itself when `valueToLabel` returns `null`.


src/Forms/SingleLookupField.php
```diff
@@ -94,7 +94,7 @@ public function Value()
             return $label;
         }
 
-        return $value;
+        return $this->value;
```
robbieaverill: How about this?

```suggestion
        return parent::Value();
```

robbieaverill:
This could also probably target an older release if you wanted to

src/Forms/SingleLookupField.php
maxime-rainville: Good point. I just implemented your suggestion.

maxime-rainville:
@robbieaverill `SingleLookupField` only got introduced in 4.3, so there's nothing for me to fix in another other minor release.

Issue #1489, opened 2013-04-03
[FORMS] [2012-06-12] CheckboxSetField doesn't save into MultiEnum correctly


silverstripe-issues:
created by: @halkyon (sharvey)
created at: 2012-06-12
original ticket: http://open.silverstripe.org/ticket/7449

---

CheckboxSetField::saveInto() sets the CheckboxSetField by replacing commas with "{comma}"

When it comes to writing into the database, it never gets converted back to "," to save into a multi enum field and MySQL silently fails.

Shouldn't saveInto() just leave the commas unescaped if it's trying to saveInto a MultiEnum field?

chillu:
Confirmed this is still the case

PR #8499, opened 2018-10-18
FIX: CheckboxSetField can now save into DBMultiEnum

sminnee:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/1489

### DIFF
--- a/src/Forms/MultiSelectField.php
+++ b/src/Forms/MultiSelectField.php
@@ -4,6 +4,7 @@ namespace SilverStripe\Forms;
 
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataObjectInterface;
+use SilverStripe\ORM\FieldType\DBMultiEnum;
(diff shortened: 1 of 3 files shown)


src/Forms/MultiSelectField.php
```diff
@@ -169,6 +184,45 @@ protected function stringDecode($value)
...
+     * @param string $value
+     * @return array
+     */
+    protected function csvDecode($value)
+    {
+        if (!$value) {
+            return [];
+        }
+
+        return preg_split('/ *, */', trim($value));
```
ScopeyNZ:
```suggestion
        return preg_split('/\s*,\s*/', trim($value));
```

sminnee:
Looks like this will need to have the multienum test split out and disabled for PGSQL

sminnee:
@silverstripe/core-team FYI this is ready for merge

PR #8503, opened 2018-10-19
FIX: Fix occasional error in SQLite3 PDO usage, Add SQLite + MariaDB builds, ensure PDO builds for all, clarify database versions

sminnee:
This PR pulls together a few interrelated changes to the databases we support and the built matrix necessary to cover them:

 * It fixes a couple of final things blocking SQLite from working nicely.
 * The build matrix now covers
   * MySQL (5.6 and 5.7), MariaDB (10.0 and 10.3), PostgreSQL (9.2 and 9.4), and SQLite (3.8 and 3.11)
   * PDO and non for each database
   * 2 specific database versions for each database (as listed above)

The built matrix is bigger but a couple of the builds are a bit faster. The total build time goes from about 50m to about 1h20. However, the wall-clock time is similar at about 15m - 20m.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,37 +16,50 @@ env:
   global:
     - COMPOSER_ROOT_VERSION=4.4.x-dev
 
+# Note that the "DBV" variable below doesnt do anything other than improve travis display
 matrix:
   fast_finish: true
   include:
     - php: 5.6
       env:
         - DB=SQLITE
+        - DBV=3.8
         - PHPCS_TEST=1
         - PHPUNIT_TEST=framework
     - php: 7.0
       env:
         - DB=MYSQL
+        - DBV=5.7
         - PHPUNIT_TEST=framework
+      dist: xenial
+      services:
+        - mysql
     - php: 7.1
       env:
         - DB=PGSQL
+        - DBV=9.2
         - PHPUNIT_TEST=framework
     - php: 7.2
       env:
         - DB=SQLITE
+        - DBV=3.11
         - PDO=1
         - PHPUNIT_TEST=framework
+      dist: xenial
     - php: 7.3
       env:
         - DB=MYSQL
+        - DBV=5.6
         - PDO=1
         - PHPUNIT_TEST=framework
     - php: 7.2
       env:
         - DB=PGSQL
+        - DBV=9.4
         - PDO=1
         - PHPUNIT_TEST=framework
+      addons:
+        postgresql: "9.4"
 
     - php: 7.1
       if: type IN (cron)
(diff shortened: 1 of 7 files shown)


kinglozzer:
The change looks fine to me, but the PDO + SQLite build has timed out. I’ve restarted it in case it’s a one-time blip

dhensby:
> The change looks fine to me, but the PDO + SQLite build has timed out. I’ve restarted it in case it’s a one-time blip

It timed out again, i bumped again - if it fails again we may have to investigate some kind of infinite loop??

dhensby:
Same timeout, same location; indicates some kind of bug here. Can you test locally with 7.3 and see what's going on?

sminnee:
I'll see if it relates to the changed approach to 7.3 config I introduced - I'll try reverting that and seeing if it fixes.

sminnee:
OK it looks like this is caused by trying to start a transaction while `@runInSeparateProcess` is in use.

I'm pretty confident that the underlying issue is that we're doing transactions in PDO wrong, the same issue that we bumped into in #8448, and the "real fix" would be merging that PR.

However, as a quick fix, I'm setting $usesTransactions = false on the two tests that make use of `@runInSeparateProcess`

sminnee:
Ok I've added MariaDB in the mix since it's been discussed. We now have 9 builds that cover:
 * PDO and non
 * SQLite, MariaDB, PostgreSQL, and MySQL
 * 2 versions of each database
 * PHP 5.6 - 7.3
 * A CMS test

image

Seems okay, but I admit it's a lot of builds.

robbieaverill:
Pretty epic matrix there. Framework would be the most important repository to run a complex set of builds like this, there's a bunch of them but I'm OK with it.

maxime-rainville:
`SilverStripe\Security\Tests\MemberAuthenticator\SessionAuthenticationHandlerTest::testAuthenticateRequestStartsSessionWithSessionIdentifier` seems grumpy and caused a lot of the build to fail.

Is this related to this change or is it something else?

sminnee:
This bug already exists in PDO + SQLite in the 4 branch, but we’re not currently testing.

The root cause is PHPunit’s runInSeparateProcess combined with transaction use.

I hate PDO, but I’m looking to see if I can avoid the need for both runInSeparateProcess and database access by refactoring the tests

maxime-rainville:
@sminnee  Are you still working on this?

emteknetnz:
@sminnee  I'm currently going through pull-requests that haven't had much activity for a while to see what is still active and what now makes sense to be closed.

I don't believe Sqlite or MariaDB are supported in any other travis files, though we've recently done a bunch of PR's to all the supported modules to get PDO on MYSQL more consistently there.

Where do you see this one going?  Do you still want to do something further here or would you prefer to now close this one?

sminnee:
We can close this for now

PR #8505, opened 2018-10-19
FIX getExtensionInstance can return null, add a case to handle that

robbieaverill:
Parent issue: https://github.com/silverstripe/cwp-recipe-kitchen-sink/issues/18

### DIFF
--- a/src/Core/Extensible.php
+++ b/src/Core/Extensible.php
@@ -135,6 +135,10 @@ trait Extensible
                 $this->addCallbackMethod($method, function ($inst, $args) use ($method, $extensionClass) {
                     /** @var Extensible $inst */
                     $extension = $inst->getExtensionInstance($extensionClass);
+                    if (!$extension) {
+                        return null;
+                    }
+
                     try {
                         $extension->setOwner($inst);
                         return call_user_func_array([$extension, $method], $args);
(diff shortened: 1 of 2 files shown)


robbieaverill:
This issue is caused by a regression in https://github.com/silverstripe/silverstripe-framework/pull/8465 that only surfaces in some cases. At the moment I can only see it in `Symbiote\AdvancedWorkflow\Tests\WorkflowEmbargoExpiryTest::testDuplicateRemoveEmbargoExpiry` on 5.0.x-dev when queuedjobs 4.x-dev is installed too. I'm looking to see if I can patch the regression as well, but this chance is defensive and should be merged too.

robbieaverill:
Ok, patch fixed for that issue

src/Core/Extensible.php
```diff
@@ -135,6 +135,10 @@ protected function defineExtensionMethods()
                 $this->addCallbackMethod($method, function ($inst, $args) use ($method, $extensionClass) {
                     /** @var Extensible $inst */
                     $extension = $inst->getExtensionInstance($extensionClass);
+                    if (!$extension) {
```
dhensby: Under what circumstances would there not be an extension instance?

robbieaverill: You can reproduce it with the advancedworkflow comment above - the method can return null though, so this is a defensive patch

PR #8508, opened 2018-10-22
Add entry to change log about CMSPageHistoryController deprecation.

maxime-rainville:
Fix https://github.com/silverstripe/silverstripe-versioned-admin/issues/67 and https://github.com/silverstripe/silverstripe-cms/issues/2277

### DIFF
--- a/docs/en/04_Changelogs/4.3.0.md
+++ b/docs/en/04_Changelogs/4.3.0.md
@@ -7,6 +7,7 @@
  - Take care with `stageChildren()` overrides. `Hierarchy::numChildren() ` results will only make use of `stageChildren()` customisations that are applied to the base class and don't include record-specific behaviour.
  - New React-based search UI for the CMS, Asset-Admin, GridFields and ModelAdmins.
  - A new `GridFieldLazyLoader` component can be added to `GridField`. This will delay the fetching of data until the user access the container Tab of the GridField.
+ - `SilverStripe\VersionedAdmin\Controllers\CMSPageHistoryViewerController` is now the default CMS history controller and `SilverStripe\CMS\Controllers\CMSPageHistoryController` has been deprecated.
 
 ## Upgrading {#upgrading}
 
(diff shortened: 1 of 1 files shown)


docs/en/04_Changelogs/4.3.0.md
```diff
@@ -41,3 +42,47 @@ public function getCMSFields()
 }
 
 ```
+
+### Keep using the legacy `CMSPageHistoryController`
+
+To keep using the old CMS history controller for every page type, add the following entry to your your YML config.
```
robbieaverill: Your your

docs/en/04_Changelogs/4.3.0.md
```diff
@@ -41,3 +42,47 @@ public function getCMSFields()
...
+
+```yml
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\CMS\Controllers\CMSPageHistoryController:
+    class: SilverStripe\CMS\Controllers\CMSPageHistoryController
+```
+
+Alternatively, you can use `HistoryControllerFactory` to switch between both controllers.
```
robbieaverill: We deprecated this class so I don’t think we should document how to use it

maxime-rainville: I'm not especially attached to this bit. But if we remove it and someone decides they want to keep exactly the same history controller behavior they had in 4.2, they don't really have away of to find out without reading through our source code.

If you think that's an acceptable compromise, I'm good removing it.

docs/en/04_Changelogs/4.3.0.md
maxime-rainville: Fixed.

docs/en/04_Changelogs/4.3.0.md
ScopeyNZ: What about saying that you may use a factory to generate either of the controllers with your own custom logic:

```yml
  SilverStripe\CMS\Controllers\CMSPageHistoryController:
    factory:
      Vendor\Package\MyCustomControllerFactory
```

I agree that it's strange to add documentation about something that's being removed, and someone will have to come back and remove this documentation eventually.

maxime-rainville: Good idea. I've added your blurb and linked off to the Factory section on the Injector doc.

PR #8514, opened 2018-10-23
FIX use Injector for FormField::castedCopy

wilr:
Allows LookupField to be replaced with user specified classes.


dhensby:
Test failures?

robbieaverill:
Re-ran, green now =)

PR #8515, opened 2018-10-23
BUGFIX MySQLi has collation conflict issues when using COALESCE()

micmania1:
To Reproduce:
- Install fluent https://github.com/tractorcow-farm/silverstripe-fluent
- Ensure SS_DATABASE_CLASS is not set or set to "MySQLDatabase"
- Login to the CMS
- Go to pages
- Search, entering a date range for LastEdited

Expected:
- Shows pages last edited within the given date range

Actual:
- MySQL Error "Illegal mix of collations (utf8mb4_unicode_ci,COERCIBLE) and (utf8mb4_general_ci,COERCIBLE) for operation '>='"

This is a problem with the MySQLi connector not setting the character set correctly between the client/database: http://php.net/manual/en/mysqli.set-charset.php#121647

### DIFF
--- a/src/ORM/Connect/MySQLiConnector.php
+++ b/src/ORM/Connect/MySQLiConnector.php
@@ -121,6 +121,13 @@ class MySQLiConnector extends DBConnector
         if (!empty($collation)) {
             $this->dbConn->query("SET collation_connection = {$collation}");
         }
+
+        // We need to do this otherwise the client character set doesn't get applied correctly since we may be
+        // using a different collation to the default.
+        // @see http://php.net/manual/en/mysqli.set-charset.php#121647
+        if (!empty($collation) && !empty($charset)) {
+            $this->dbConn->query("SET NAMES {$charset} COLLATE {$collation}");
+        }
     }
 
     public function __destruct()


src/ORM/Connect/MySQLiConnector.php
```diff
@@ -121,6 +121,13 @@ public function connect($parameters, $selectDB = false)
         if (!empty($collation)) {
```
sminnee: Do we still need this query? I seems like a double-up, although I'm not sure how to test for it.

src/ORM/Connect/MySQLiConnector.php
```diff
@@ -121,6 +121,13 @@ public function connect($parameters, $selectDB = false)
...
             $this->dbConn->query("SET collation_connection = {$collation}");
         }
+
+        // We need to do this otherwise the client character set doesn't get applied correctly since we may be
+        // using a different collation to the default.
+        // @see http://php.net/manual/en/mysqli.set-charset.php#121647
+        if (!empty($collation) && !empty($charset)) {
+            $this->dbConn->query("SET NAMES {$charset} COLLATE {$collation}");
```
sminnee: This is called out as not recommended in the PHP docs

> This is the preferred way to change the charset. Using mysqli_query() to set it (such as SET NAMES utf8) is not recommended. See the MySQL character set concepts section for more information.

See 
http://nz2.php.net/manual/en/mysqlinfo.concepts.charset.php

Perhaps the fix here is to check if set_charset returns false and if so, throw an error?

sminnee: For whatever reason, `SET collation_connection` above isn't working. It would be good to understand why. Calling set_charset followed by SET NAMEs is probably no worse than calling set_charset followed by SET collation_connection, as long as the same charset is given to both. It's a little redundant, though,.

Still — the relevant thing that SET NAMES does is in turn call SET collation_connection, so there's something fishy going on.

sminnee:
Related question: why is a datetime comparison getting a string collation error?

sminnee:
@micmania1 did you ever get any time to dig into my comments above?

micmania1:
@sminnee nah, not yet. This got de-prioritised as we had a workaround.

maxime-rainville:
Anyone has any objections if close this? Speak now or hold your peace.

PR #8516, opened 2018-10-24
FIX DataObject singleton creation

### LINKED ISSUE #4878: populateDefaults not working on extensions

sminnee:
Ensure DataObject instances are aware they are singletons so functions like populateDefaults() can be skipped. (fixes #4878)

Correctly applies https://github.com/silverstripe/silverstripe-framework/pull/7850 to the 4.x line
This has already been fixed in 3.x

### DIFF
--- a/src/Core/Injector/Injector.php
+++ b/src/Core/Injector/Injector.php
@@ -13,6 +13,7 @@ use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Environment;
 use SilverStripe\Dev\Deprecation;
+use SilverStripe\ORM\DataObject;
 
 /**
  * A simple injection manager that manages creating objects and injecting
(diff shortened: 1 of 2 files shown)


src/Core/Injector/Injector.php
```diff
@@ -581,6 +582,11 @@ protected function instantiate($spec, $id = null, $type = null)
             $constructorParams = $spec['constructor'];
         }
 
+        // If we're dealing with a DataObject, pass through Singleton flag as second argument
+        if ($type != 'prototype' && empty($constructorParams) && is_subclass_of($class, DataObject::class)) {
```
unclecheese: Strict comparison, please. Also use the provided constant (`self::PROTOTYPE`)

chillu:
Tentatively marking as 4.3, assuming it's nearly over the line. This is the issue where every first dev build on a migrated data set fails in certain situations, making for a pretty crap onboarding experience to our platforms, right?

sminnee:
@unclecheese @chillu if this is merged to the 4.1 branch today, will it make it into the 4.3 release? Or should I rebase this to a different branch?

src/Core/Injector/Injector.php
sminnee: Fixed. Also checked for !$type, as that's interpreted as singleton elsewhere.

PR #8519, opened 2018-10-26
NEW Make resources dir configurable

maxime-rainville:
This pull request updates frameworks so the name of the folder where vendor-module assets are exposed can be configured.

You need https://github.com/silverstripe/vendor-plugin/pull/26 for this PR to work.

A new `RESOURCES_DIR` constant is defined and replaces and `ManifestFileFinder::RESOURCES_DIR`.

# Parent issue
* https://github.com/silverstripe/silverstripe-framework/issues/7932

### DIFF
--- a/docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
+++ b/docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
@@ -11,6 +11,68 @@ The examples below are using certain folder naming conventions (CSS files in `cs
 SilverStripe core modules like `cms` use a different naming convention (CSS and JavaScript files in `client/src/`).
 The `Requirements` class can work with arbitrary file paths.
 
+## Exposing static assets
+
+Before requiring a static assets file in PHP code or in a template, those assets need to be "exposed". This process allows SilverStripe projects and SilverStripe modules to make static asset files available via the web server without directly exposing PHP files.
+
+### Configuring your project "exposed" folders
+
+Exposed assets are made available in your web root in a dedicated "resources" directory. Prior to SilverStripe 4.4, the name of this directory was hardcoded to `resources`. In SilverStripe 4.4 and above, the name of the resources directory can be configured by defining the `extra.resources-dir` key in your `composer.json`. SilverStripe projects created from `silverstripe/installer` 4.4 and above will automatically be configured to use `_resources` as their resource directory.
+
+Each folder that needs to be exposed must be entered under the `extra.expose` key in your `composer.json` file. Module developers should use a path relative to the root of their module (don't include the "vendor/package-developer/package-name" path).
+
+This is a sample `composer.json` configured to expose some assets.
+
+```json
+{
+    "name": "app/myproject",
+    "type": "silverstripe-project",
+    "require": {
+        "silverstripe/recipe-cms": "4.4.x-dev"
+    },
+    "extra": {
+        "resources-dir": "_resources",
+        "expose": [
+            "app/client/dist",
+            "app/images"
+        ]
+    }
+}
+```
+
+Files contained inside the `app/client/dist` and `app/images` will be made publicly available under the `_resources` directory.
+
+SilverStripe projects should not track the "resources" directory in their source control system.
+
+### Exposing assets in the web root
+
+SilverStripe projects ship with `silverstripe/vendor-plugin`. This composer plugin automatically tries to expose assets of modules after installation or after an update.
+
+Developers can explicitly expose static assets by calling `composer vendor-expose`. This is necessary after updating your `resources-dir` or `expose` configuration in your `composer.json` file.
+
+`composer vendor-expose` accepts an optional `method` argument (e.g.: `composer vendor-expose auto`). This controls how the files are exposed in the "resources" directory:
+* `none` disables all symlink / copy,
+* `copy` copy the exposed files,
+* `symlink` create symlinks to the exposed folder,
+* `junction` uses a junction (windows only),
+* `auto` performs symlink (or junction on windows), but fail over to copy.
+
+### Referencing exposed assets
+
+When referencing exposed assets, you should not use their web server path. e.g.:
+```php
+// When referencing project files, use the same path defined in your `composer.json` file.
+Requirements::javascript('app/client/dist/bundle.js');
+
+// When referencing theme files, use a path relative to the root of your project
+Requirements::javascript('themes/simple/javascript/script.js');
+
+// When referencing files from a module, you need to prefix the path with the module name.
+Requirements::javascript('silverstripe/admin:client/dist/js/bundle.js');
+```
+
+When rendered in HTML code, these URLs will be rewritten to their matching path inside the "resources" directory.
+
 ## Template Requirements API
 
 **<my-module-dir>/templates/SomeTemplate.ss**
@@ -179,7 +241,7 @@ is not appropriate. Normally a single backend is used for all site assets, so a 
 replaced. For instance, the below will set a new set of dependencies to write to `app/javascript/combined`
 
 
-```yaml
+```yml
 ---
 Name: myrequirements
 ---
--- a/docs/en/03_Upgrading/index.md
+++ b/docs/en/03_Upgrading/index.md
@@ -1022,14 +1022,14 @@ If you are using a modified `index.php`, `.htaccess`, or `web.config`, you will 
   * `assets`
   * Any `favicon` files
   * Other common files that should be accssible in your project webroot (example: `robots.txt`)
-* Delete the root `resources` directory if present.
+* Delete the root `resources` or `_resources` directories if present.
 * Run the following command `composer vendor-expose` to make static assets files accessible via the `public` directory.
 
 If you are upgrading from SilverStripe 4.0 to SilverStripe 4.1 (or above), you'll need to update `index.php` before moving it to the public folder. You can get a copy of the generic `index.php` file from `vendor/silverstripe/recipe-core/public`. If you've made modifications to your `index.php` file, you'll need to replicate those into the new `public/index.php` file.
 
 ### Finalising the web root migration
 You'll need to update your server configuration to point to the public directory rather than the root of your project.
-Update your `.gitignore` file so `assets` and `resources` are still ignored when located under the `public` folder.
+Update your `.gitignore` file so `assets` and `_resources` (or `resources` if using a pre SilverStripe 4.3 release) are still ignored when located under the `public` folder.
 Your project should still be functional, although you may now be missing some static assets.
 
 This is a good point to commit your changes to your source control system before moving on to the next step.
--- a/docs/en/03_Upgrading/index.md
+++ b/docs/en/03_Upgrading/index.md
@@ -1029,7 +1029,7 @@ If you are upgrading from SilverStripe 4.0 to SilverStripe 4.1 (or above), you'l
 
 ### Finalising the web root migration
 You'll need to update your server configuration to point to the public directory rather than the root of your project.
-Update your `.gitignore` file so `assets` and `_resources` (or `resources` if using a pre SilverStripe 4.3 release) are still ignored when located under the `public` folder.
+Update your `.gitignore` file so `assets` and `_resources` (or `resources` if using a pre SilverStripe 4.4 release) are still ignored when located under the `public` folder.
 Your project should still be functional, although you may now be missing some static assets.
 
 This is a good point to commit your changes to your source control system before moving on to the next step.
--- a/docs/en/04_Changelogs/4.4.0.md
+++ b/docs/en/04_Changelogs/4.4.0.md
@@ -8,7 +8,15 @@
 
  - dev/build is now non-destructive for all Enums, not just ClassNames. This means your data won't be lost if you're switching between versions, but watch out for code that breaks when it sees an unrecognised value!
 
+### Adapting to new `_resources` directory
+
+Update your `.gitignore` file to ignore the new `_resources` directory. This file is typically located in the root of your project or in the `public` folder.
+
+You may also need to update your server configuration if you have applied special conditions to the `resources` path.
+
 ## Changes to internal APIs
 
  - `PDOQuery::__construct()` now has a 2nd argument. If you have subclassed PDOQuery and overridden __construct() 
    you may see an E_STRICT error
+ - The name of the directory where vendor module resources are exposed can now be configured via the `SS_RESOURCES_DIR` environment variables and defaults to `_resources` rather than `resources`. This allows you to use `resources` as a URL segment or a route.
+
--- a/docs/en/04_Changelogs/4.4.0.md
+++ b/docs/en/04_Changelogs/4.4.0.md
@@ -3,14 +3,28 @@
 ## Overview {#overview}
 
  - [Correct PHP types are now returned from database queries](/developer_guides/model/sql_select#data-types)
- 
-## Upgrading {#upgrading}
-
+ - The name of the directory where vendor module resources are exposed can now be configured by defining a `extra.resources-dir` key in your `composer.json` file. If the key is not set, it will automatically default to `resources`. New projects will be preset to `_resources`. This will avoid potential conflict with SiteTree URL Segments.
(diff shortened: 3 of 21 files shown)


maxime-rainville:
If we decide to re-target this to 4.4, the change log and some of the doc will need to be updated, because they explicitly reference 4.3

sminnee:
I believe that this renders #8364 obsolete?

robbieaverill:
Noting that the dependent PR has been merged now. I've re-run the Travis builds.

maxime-rainville:
I've rebased it.

robbieaverill:
>Alternatively, I could update the unit test to expect the old name.

That sounds like a good idea to me, but I'm not really over this change yet

maxime-rainville:
Just having another look. `installer`, `recipe-core` and `recipe-cms` should normally have a `resources-dir` extra composer key. I think they will all run the framework test. So if I adjust the test to expect `resources`, they will break.

I'll update the test to expect the constant. So it will work whatever is in the composer file.

docs/en/03_Upgrading/index.md
```diff
@@ -1022,14 +1022,14 @@ If you are using a modified `index.php`, `.htaccess`, or `web.config`, you will
   * `assets`
   * Any `favicon` files
   * Other common files that should be accssible in your project webroot (example: `robots.txt`)
-* Delete the root `resources` directory if present.
+* Delete the root `resources` or `_resources` directories if present.
```
robbieaverill: Worth mentioning here something like "or whatever your custom `resources-dir` value is in `composer.json`?

docs/en/04_Changelogs/4.4.0.md
```diff
@@ -3,12 +3,33 @@
...
- 
+ - The name of the directory where vendor module resources are exposed can now be configured by defining a `extra.resources-dir` key in your `composer.json` file. If the key is not set, it will automatically default to `resources`. New projects will be preset to `_resources`. This will avoid potential conflict with SiteTree URL Segments.
+ - dev/build is now non-destructive for all Enums, not just ClassNames. This means your data won't be lost if you're switching between versions, but watch out for code that breaks when it sees an unrecognised value!
+
+### Adapting to new `_resources` directory
+
 ## Upgrading {#upgrading}
 
- - dev/build is now non-destructive for all Enums, not just ClassNames. This means your data won't be lost if you're switching between versions, but watch out for code that breaks when it sees an unrecognised value!
+### Adopting to new `_resources` directory
```
robbieaverill: Duplicate headings?

src/Core/Manifest/Module.php
```diff
@@ -124,6 +128,20 @@ public function getShortName()
         return basename($this->path);
     }
 
+    /**
+     * Name of the resource directory where vendor resources should be exposed as defined by the `extra.resourcesDir`
```
robbieaverill:
```suggestion
     * Name of the resource directory where vendor resources should be exposed as defined by the `extra.resources-dir`
```

src/Core/Manifest/Module.php
```diff
@@ -124,6 +128,20 @@ public function getShortName()
...
+     * key in the composer file. A blank string will will be returned if the key is undefined.
+     *
+     * Only applicaple when reading the composer file for the main project.
```
robbieaverill:
```suggestion
     * Only applicable when reading the composer file for the main project.
```

src/includes/constants.php
```diff
@@ -198,3 +199,18 @@
 if (!defined('TEMP_FOLDER')) {
     define('TEMP_FOLDER', TEMP_PATH);
 }
+
+// Define the Ressource Dir constant that will be use to exposed vendor assets
+if (!defined('RESOURCES_DIR')) {
+    $project = new SilverStripe\Core\Manifest\Module(BASE_PATH, BASE_PATH);
+    $resourceDir = $project->getResourcesDir() ?: 'resources';
```
robbieaverill: Super minor, but maybe we could call it `$resourcesDir` for consistency

docs/en/04_Changelogs/4.4.0.md
```diff
@@ -3,12 +3,33 @@
...
+3. Expose your vendor assets by running `composer vendor-expose`.
+4. Remove the old `resources` folder. This folder will be located in the `public` folder if you have adopted the public web root, or in the root of your project if you haven't.
+
+You may also need to update your server configuration if you have applied special conditions to the `resources` path.
 
 ## Changes to internal APIs
 
  - `PDOQuery::__construct()` now has a 2nd argument. If you have subclassed PDOQuery and overridden __construct() 
    you may see an E_STRICT error
+ - The name of the directory where vendor module resources are exposed can now be configured via the `SS_RESOURCES_DIR` environment variables and defaults to `_resources` rather than `resources`. This allows you to use `resources` as a URL segment or a route.
```
robbieaverill: I can't see any evidence of this being true, am I missing something?

docs/en/03_Upgrading/index.md
maxime-rainville: This would only apply to someone using using S4.4 or greater without the public web-root who decided to update the public webroot. I think bringing in the _resources-dir value in composer.json_ here would make thing more confusing for users.

docs/en/04_Changelogs/4.4.0.md
maxime-rainville: Fixed.

src/Core/Manifest/Module.php
maxime-rainville: Fixed.

src/includes/constants.php
maxime-rainville: Fixed.

docs/en/04_Changelogs/4.4.0.md
maxime-rainville: That's a left over from when I was thinking of making it an environment variable. Fixed it now.

maxime-rainville:
I've added docs about the finer intricacies of exposing assets in the Requirement DOC article. 542842791ea8f0b18b37d39d15a26c143e44bc2b

I'm not sure this is the best place to have it. Might be worthwhile having this in a dedicated article. 

There's other part of this article that probably need to be updated/cleaned-up, but that's another job for another day.

docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
```diff
@@ -11,6 +11,68 @@ The examples below are using certain folder naming conventions (CSS files in `cs
 SilverStripe core modules like `cms` use a different naming convention (CSS and JavaScript files in `client/src/`).
 The `Requirements` class can work with arbitrary file paths.
 
+## Exposing static assets
+
+Before requiring a static assets file in PHP code or in a template, those assets need to be "exposed". This process allows SilverStripe projects and SilverStripe modules to make static asset files available via the web server without directly exposing PHP files.
```
robbieaverill:
```suggestion
Before requiring static asset files in PHP code or in a template, those assets need to be "exposed". This process allows SilverStripe projects and SilverStripe modules to make static asset files available via the web server without directly exposing PHP files.
```

Worth adding something like "... from locations that would otherwise be blocked from web server access, such as the `vendor` folder"?

docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
```diff
@@ -11,6 +11,68 @@ The examples below are using certain folder naming conventions (CSS files in `cs
...
+
+### Configuring your project "exposed" folders
+
+Exposed assets are made available in your web root in a dedicated "resources" directory. Prior to SilverStripe 4.4, the name of this directory was hardcoded to `resources`. In SilverStripe 4.4 and above, the name of the resources directory can be configured by defining the `extra.resources-dir` key in your `composer.json`. SilverStripe projects created from `silverstripe/installer` 4.4 and above will automatically be configured to use `_resources` as their resource directory.
+
+Each folder that needs to be exposed must be entered under the `extra.expose` key in your `composer.json` file. Module developers should use a path relative to the root of their module (don't include the "vendor/package-developer/package-name" path).
+
+This is a sample `composer.json` configured to expose some assets.
```
robbieaverill: Worth noting that this is a project composer.json file, rather than a module file? It's obvious from the `type` in the example, but maybe it couldn't hurt to add a word or two in this sentence to make it crystal clear?

docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
```diff
@@ -11,6 +11,68 @@ The examples below are using certain folder naming conventions (CSS files in `cs
...
+            "app/client/dist",
+            "app/images"
+        ]
+    }
+}
+```
+
+Files contained inside the `app/client/dist` and `app/images` will be made publicly available under the `_resources` directory.
+
+SilverStripe projects should not track the "resources" directory in their source control system.
```
robbieaverill: This is slightly opinionated to be honest - it depends on your deployment process. If you're deploying and running `composer install`, then you shouldn't track it in VCS, but if you're not using composer on the production server you're going to be committing in the vendor and resources folders. I agree that this is true for the majority of use cases though.

docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
```diff
@@ -11,6 +11,68 @@ The examples below are using certain folder naming conventions (CSS files in `cs
...
+
+### Exposing assets in the web root
+
+SilverStripe projects ship with `silverstripe/vendor-plugin`. This composer plugin automatically tries to expose assets of modules after installation or after an update.
```
robbieaverill:
```suggestion
SilverStripe projects ship with `silverstripe/vendor-plugin`. This Composer plugin automatically tries to expose assets from your project and installed modules after installation, or after an update.
```

docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
```diff
@@ -11,6 +11,68 @@ The examples below are using certain folder naming conventions (CSS files in `cs
...
+
+Developers can explicitly expose static assets by calling `composer vendor-expose`. This is necessary after updating your `resources-dir` or `expose` configuration in your `composer.json` file.
+
+`composer vendor-expose` accepts an optional `method` argument (e.g.: `composer vendor-expose auto`). This controls how the files are exposed in the "resources" directory:
+* `none` disables all symlink / copy,
+* `copy` copy the exposed files,
+* `symlink` create symlinks to the exposed folder,
+* `junction` uses a junction (windows only),
+* `auto` performs symlink (or junction on windows), but fail over to copy.
```
robbieaverill: Minor: we don't usually terminate list items with commas etc

docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
```diff
@@ -11,6 +11,68 @@ The examples below are using certain folder naming conventions (CSS files in `cs
...
+
+### Exposing assets in the web root
+
+SilverStripe projects ship with `silverstripe/vendor-plugin`. This composer plugin automatically tries to expose assets of modules after installation or after an update.
+
+Developers can explicitly expose static assets by calling `composer vendor-expose`. This is necessary after updating your `resources-dir` or `expose` configuration in your `composer.json` file.
+
+`composer vendor-expose` accepts an optional `method` argument (e.g.: `composer vendor-expose auto`). This controls how the files are exposed in the "resources" directory:
+* `none` disables all symlink / copy,
+* `copy` copy the exposed files,
```
robbieaverill: This line and the next line are using a different tense to the others

docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
```diff
@@ -11,6 +11,68 @@ The examples below are using certain folder naming conventions (CSS files in `cs
...
+* `symlink` create symlinks to the exposed folder,
```
robbieaverill:
```suggestion
* `symlink` create symbolic links to the exposed folder,
```

docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
```diff
@@ -11,6 +11,68 @@ The examples below are using certain folder naming conventions (CSS files in `cs
...
+* `junction` uses a junction (windows only),
```
robbieaverill:
```suggestion
* `junction` uses a junction (Windows only),
```

docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
robbieaverill:
```suggestion
* `auto` creates symbolic links (or junctions on Windows), but fails over to copy.
```

docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
```diff
@@ -11,6 +11,68 @@ The examples below are using certain folder naming conventions (CSS files in `cs
...
+* `auto` performs symlink (or junction on windows), but fail over to copy.
+
+### Referencing exposed assets
+
+When referencing exposed assets, you should not use their web server path. e.g.:
```
robbieaverill: I think we should document what developers should do rather than what they shouldn't, e.g.:

```suggestion
When referencing exposed static assets, use either the project file path (relative to the project root folder) or a module name and relative file path to that module's root folder. E.g.:
```

docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
```diff
@@ -179,7 +241,7 @@ is not appropriate. Normally a single backend is used for all site assets, so a
 replaced. For instance, the below will set a new set of dependencies to write to `app/javascript/combined`
 
 
-```yaml
+```yml
```
robbieaverill: FYI, these are synonymous in GitHub markdown rendering: https://github.com/github/linguist/blob/e761f9b013e5b61161481fcb898b59721ee40e3d/lib/linguist/languages.yml#L5480-L5484

docs/en/04_Changelogs/4.4.0.md
```diff
@@ -3,12 +3,31 @@
...
+}
+```
+3. Expose your vendor assets by running `composer vendor-expose`.
+4. Remove the old `resources` folder. This folder will be located in the `public` folder if you have adopted the public web root, or in the root of your project if you haven't.
+
+You may also need to update your server configuration if you have applied special conditions to the `resources` path.
 
 ## Changes to internal APIs
 
  - `PDOQuery::__construct()` now has a 2nd argument. If you have subclassed PDOQuery and overridden __construct()
```
robbieaverill:
```suggestion
 - `PDOQuery::__construct()` now has a 2nd argument. If you have subclassed PDOQuery and overridden `__construct()` 
```

To prevent flow on bold highlighting (see GitHub preview) - I don't think our docs site renders it this way though, just a minor thing

docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
maxime-rainville: There's nothing wrong with being opinionated when you are right! :stuck_out_tongue_winking_eye: 

The description will be appropriate for 90% of developers. I'm keen on avoiding going off on a tangent about "if you've got a weird GIT set up, you should do this ..."

If a developer decide to deviate from best practices, they ought to understand why and adjust their set up accordingly.

maxime-rainville:
@robbieaverill PR got merged before I got around to implement your feedback. I'll address your feedback in a separate PR.

docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
maxime-rainville: fixed.

docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
maxime-rainville: Fixed.

docs/en/02_Developer_Guides/01_Templates/03_Requirements.md
robbieaverill: Yeah fair enough!

Issue #8518, opened 2018-10-25
GridFieldFilterHeader legacy option


mfendeksilverstripe:
## Affected Version

`"silverstripe/recipe-cms": "4.3.x-dev" / 1577d6386b27922b825928811ce9bc0fa108f227`

## Description

`useLegacyFilterHeader` option is very appreciated, but unfortunately, a lot of SilverStripe modules do not use `Injector` so using this option doesn't seem very viable. A global legacy mode option is needed.

I recommend adding a simple static configuration for `useLegacyFilterHeader` default value. This default value would be used in the constructor.

```
public function __construct(
    $useLegacy = null,
    callable $updateSearchContext = null,
    callable $updateSearchForm = null
) {
    if ($useLegacy === null) {
        // fetch global default for this value
    }

    $this->useLegacyFilterHeader = $useLegacy;
    $this->updateSearchContextCallback = $updateSearchContext;
    $this->updateSearchFormCallback = $updateSearchForm;
}
```

Note that this option is very much needed because the new rendering mode of this component renders different scaffolded fields.

chillu:
In general, we should aim to update the components under our control which use this legacy mode. Which ones are causing this request?

mfendeksilverstripe:
@chillu I'm not sure where the problem is, however `4.3` introduced a breaking change which should have been an opt-in, not opt-out on the instance level. As of right now, all gridfields with filter header components are practically unusable.

The top priority should be to provide an opt-out of this breaking change and investigate the issue afterward.

mfendeksilverstripe:
@chillu I think the filter component changes the way how fields are scaffolded. I also raised a separate issue related to the scaffolding itself:

https://github.com/silverstripe/silverstripe-framework/issues/8517

lukereative:
In this instance it's because `QueuedJobDescriptor` has 
```
    private static $searchable_fields = [
        'JobTitle',
    ];
```

Take that away and you get:
![image](https://user-images.githubusercontent.com/1415622/47529235-bcc5a980-d903-11e8-95b5-3fc0c4da6f90.png)

mfendeksilverstripe:
Thanks for the update. I don't think we can afford to go through each gridfield and figure out why the display is different now. We need global opt-out of this feature as soon as possible.

bergice:
Merged https://github.com/silverstripe/silverstripe-framework/pull/8521

lienajp:
indeed the fix in 8827 is a workaround but it works.

PR #8521, opened 2018-10-26
Add config to force legacy filter header globally

lukereative:
Resolves silverstripe/silverstripe-framework#8518


mfendeksilverstripe:
Nice one! Thanks @lukereative :)

PR #8534, opened 2018-10-29
BUGFIX many many through not sorting by join table

### LINKED ISSUE #7981: SS4: Cannot sort `many_many` result when using `through`

micmania1:
Resolves #7981 

Todo:

- [x] Write new tests to cover default sort on main dataobject as well as join table
- [x] Add docs on sorting many many lists

`many_many` joins work by doing a join with two separate queries.

eg.
```
SELECT ...
 FROM "Fluent_Locale" INNER JOIN (
     SELECT ...
     FROM "Fluent_FallbackLocale"
     WHERE ("Fluent_FallbackLocale"."ParentID" = ?)
     ORDER BY "Fluent_FallbackLocale"."Sort" ASC
 ) AS "Fluent_FallbackLocale" ON "Fluent_FallbackLocale"."LocaleID" = "Fluent_Locale"."ID"
 ORDER BY "Fluent_Locale"."Locale" ASC
```

As you can see, the inner query is sorting correctly on the join table (by Sort ASC). However, the main query has its own sort which overrides that when you get the final set of results.

This inner query is built by SilverStripe using the standard ORM functions and assumes the default sort of the join DataObject (FluentFallbackLocale in this case). Likewise for the main query.

With @robbieaverill's [fix](https://github.com/silverstripe/silverstripe-framework/pull/8417/files#diff-d3a5863552455ed2232ba1082ea6e1fbR257) it will sort correctly in cases where no default sort is set on the main DataObject and default is set on the join table data object. However, if the default sort is set on the main Dataobject, the fix will essentially be bypassed since sort is already set which is what we're seeing with Fluent.

The fix i've added applies the sort order much earlier, when it first starts building the ORM query objects. It will check to see if there is config which defines a default sort on the join table (this should also work for standard many-many sort too) and applies that if so. If that has no value, it will fallback to whatever sort SilverStripe gives it (default sort on dataobject or natural database sort).

This happens early enough that `->sort()` will function as normal if the user decides to sort by something else and early enough that `updateManymanyComponents` can be hooked into.

### DIFF
--- a/docs/en/02_Developer_Guides/00_Model/02_Relations.md
+++ b/docs/en/02_Developer_Guides/00_Model/02_Relations.md
@@ -295,6 +295,15 @@ class Supporter extends DataObject
 }
 ```
 
+To ensure this `many_many` is sorted by "Ranking" by default you can add this to your config:
+
+```yaml
+Team_Supporters:
+  default_sort: '"Team_Supporter"."Ranking" ASC'
+```
+
+`Team_Supporters` is the table name automatically generated for the many_many relation in this case.
+
 ### many_many through relationship joined on a separate DataObject
 
 If necessary, a third DataObject class can instead be specified as the joining table,
@@ -311,6 +320,9 @@ This is declared via array syntax, with the following keys on the many_many:
  - `through` Class name of the mapping table
  - `from` Name of the has_one relationship pointing back at the object declaring many_many
  - `to` Name of the has_one relationship pointing to the object declaring belongs_many_many.
+
+Just like a any normal DataObject, you can apply a default sort which will be applied when
+accessing many many through relations.
 
 Note: The `through` class must not also be the name of any field or relation on the parent
 or child record.
@@ -348,6 +360,8 @@ class TeamSupporter extends DataObject
         'Team' => Team::class,
         'Supporter' => Supporter::class,
     ];
+
+    private static $default_sort = '"TeamSupporter"."Ranking" ASC'
 }
 ```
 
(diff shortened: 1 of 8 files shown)


tractorcow:
I had an idea once where you could declare the sort directly on the `many_many` array config, so you could override the model-specific sorts, but just for that relation.

e.g.

```
private static $many_many = [
    'Items' => [
        'through' => Relation::class,
        'from' => 'Parent',
        'to' => 'Child',
        'sort' => '"App_Relation"."Sort" ASC',
    ],
];
```

mfendeksilverstripe:
+1 for Damian's idea. However, at this point, we will be happy with whatever solution that works.

PR #8535, opened 2018-10-29
Persist GridField readonly state, add view button

### LINKED ISSUE #3357: [GRIDFIELD] GridField needs a useful readonly view (currently just showing "none")

lukereative:
Resolves #3357 (remaining ACs/bugs)

- Fixes an issue that when the GridField is called via pjax it loses it's readonly state i.e. when pagination buttons are pushed
- Add a couple missing components to readonly whitelist
- Add view button to re-enable access to view records as edit buttons will be removed

### DIFF
--- a/src/Forms/GridField/GridField.php
+++ b/src/Forms/GridField/GridField.php
@@ -255,7 +255,7 @@ class GridField extends FormField
         }
 
         // As the edit button may have been removed, add a view button if it doesn't have one
-        if ($copyConfig->getComponentsByType(GridFieldViewButton::class)->count() === 0) {
+        if (!$copyConfig->getComponentByType(GridFieldViewButton::class)) {
             $copyConfig->addComponent(new GridFieldViewButton);
         }
 
@@ -306,7 +306,7 @@ class GridField extends FormField
     public function setReadonly($readonly)
     {
         parent::setReadonly($readonly);
-        $this->getState()->Readonly($readonly);
+        $this->getState()->Readonly = $readonly;
         return $this;
     }
 
@@ -1029,7 +1029,7 @@ class GridField extends FormField
         }
 
         if ($request->getHeader('X-Pjax') === 'CurrentField') {
-            if ($this->getState(true)->Readonly) {
+            if ($this->getState()->Readonly) {
                 $this->performDisabledTransformation();
             }
             return $this->FieldHolder();
--- a/src/Forms/GridField/GridField.php
+++ b/src/Forms/GridField/GridField.php
@@ -113,11 +113,13 @@ class GridField extends FormField
     protected $readonlyComponents = [
         GridField_ActionMenu::class,
         GridFieldConfig_RecordViewer::class,
+        GridFieldButtonRow::class,
         GridFieldDataColumns::class,
         GridFieldDetailForm::class,
         GridFieldLazyLoader::class,
         GridFieldPageCount::class,
         GridFieldPaginator::class,
+        GridFieldFilterHeader::class,
         GridFieldSortableHeader::class,
         GridFieldToolbarHeader::class,
         GridFieldViewButton::class,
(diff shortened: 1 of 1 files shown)


src/Forms/GridField/GridField.php
```diff
@@ -1009,6 +1029,9 @@ public function gridFieldAlterAction($data, $form, HTTPRequest $request)
         }
 
         if ($request->getHeader('X-Pjax') === 'CurrentField') {
+            if ($this->getState(true)->Readonly) {
```
unclecheese: `true` is the default value. This can just be `getState()`

src/Forms/GridField/GridField.php
```diff
@@ -290,6 +298,18 @@ public function setConfig(GridFieldConfig $config)
...
 
+    /**
+     * @param bool $readonly
+     *
+     * @return $this
+     */
+    public function setReadonly($readonly)
+    {
+        parent::setReadonly($readonly);
+        $this->getState()->Readonly($readonly);
```
unclecheese: Wondering, only because this is so ugly and IDE unfriendly, is this functionally equivalent to just `getState()->Readonly = $readonly`? I think it's just a question of whether it uses `__call()` or `__set()`.

src/Forms/GridField/GridField.php
lukereative: Done

src/Forms/GridField/GridField.php
lukereative: Done

src/Forms/GridField/GridField.php
```diff
@@ -241,16 +243,22 @@ public function performReadonlyTransformation()
...
         foreach ($this->getConfig()->getComponents() as $component) {
             // if a component doesn't exist, remove it from the readonly version.
             if (!in_array(get_class($component), $allowedComponents)) {
-                $copy->getConfig()->removeComponent($component);
+                $copyConfig->removeComponent($component);
             }
         }
 
+        // As the edit button may have been removed, add a view button if it doesn't have one
+        if ($copyConfig->getComponentsByType(GridFieldViewButton::class)->count() === 0) {
```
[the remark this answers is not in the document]
lukereative: Got it mixed up with `removeComponentsByType` which only has a multiple selector method, I've now changed it to the singular.

Issue #8543, opened 2018-10-29
Duplicate Session Cookies


caffeineinc:
## Affected Version

4.x.x

## Description

SilverStripe 4 changes how php sessions are bootstrapped, but with the current configuration the session related cookies are duplicated, re-sent on every request. This causes issues with WAF's and cookie persistence as there is always a "Set-Cookie" header. 

Session should only be bootstrapped when/if there is modification (writes or changes) to the session data to be persisted. If there's a form on a page, this will create a session (to save the CSRF token) - The current behaviour creates the session,  but writes two cookies as [session_start()](https://github.com/silverstripe/silverstripe-framework/blob/4.0/src/Control/Session.php#L278) starts a cookie, and outputs a header. Then the [timeout](https://github.com/silverstripe/silverstripe-framework/blob/4.0/src/Control/Session.php#L288) code resets the header values. 

```
< Set-Cookie: PHPSESSID=<sessionid>; expires=Mon, 29-Oct-2018 21:38:08 GMT; Max-Age=3600; path=/; HttpOnly
< Set-Cookie: PHPSESSID=<sessionid>; expires=Mon, 29-Oct-2018 21:38:08 GMT; Max-Age=3600; path=/; HttpOnly
```

The problem here's is the second header is unneeded, but  also is being output even if we have a session bootstrapped elsewhere. This means you **always** get a Set-Cookie on every requests where session is used. this breaks WAFs that are trying to use persistence to direct sessions to the same origins when load-balancing.  

## Steps to Reproduce

- `SilverStripe\Control\Session` config parameter `timeout` must be `!= 0`
- Create a page in SilverStripe with a form with a CSRF token (default is enabled)
- Visit the page and note the headers with curl. `curl -svok /dev/null https://mysite.com/my-form-page`
- Note, duplicate headers in response
- Visit the page again with the headers and cookie `curl -svok --cookie "PHPSESSID=<PHPSESSID from-request-above>" /dev/null https://mysite.com/my-form-page` 
- Note `Set-Cookie` header with the same PHPSESSID value as your cookie. 

The second request shouldn't return another `Set-Cookie` unless the session has been destroyed, updated, or is no longer valid.

Related to https://github.com/silverstripe/silverstripe-hybridsessions/issues/41

# Pull Request

 * [x] https://github.com/silverstripe/silverstripe-framework/pull/8544
 * [x] https://github.com/silverstripe/silverstripe-framework/pull/8639

[...]

PR #8544, opened 2018-10-29
Resolve duplicate session headers #8543

caffeineinc:
Headers were being sent twice, once by session_start() and once by Cookie::set()

The second header was unneeded, and meant `Set-Cookie` was appended to every request which
breaks WAF behaviour for sticky sessions

@see https://github.com/silverstripe/silverstripe-framework/issues/8543

### DIFF
--- a/src/Control/Session.php
+++ b/src/Control/Session.php
@@ -299,19 +299,13 @@ class Session
         }
 
         if (!session_id() && !headers_sent()) {
-            if ($domain) {
-                session_set_cookie_params($timeout, $path, $domain, $secure, true);
-            } else {
-                session_set_cookie_params($timeout, $path, null, $secure, true);
-            }
-
             // Allow storing the session in a non standard location
             if ($session_path) {
                 session_save_path($session_path);
             }
 
-            // If we want a secure cookie for HTTPS, use a seperate session name. This lets us have a
-            // seperate (less secure) session for non-HTTPS requests
+            // If we want a secure cookie for HTTPS, use a separate session name. This lets us have a
+            // separate (less secure) session for non-HTTPS requests
             if ($secure) {
                 session_name($this->config()->get('cookie_name_secure'));
             }
@@ -321,27 +315,29 @@ class Session
                 session_cache_limiter($limiter);
             }
 
-            session_start();
+            $sessionParameters = [
+                "cookie_path" => $path,
+                "cookie_domain" => $domain ?: "",
+                "cookie_secure" => $secure,
+                "cookie_httponly" => true
+            ];
+
+            if ($timeout) {
+                $sessionParameters["cookie_lifetime"] = $timeout;
+            }
+
+            session_start($sessionParameters);
 
             if (isset($_SESSION)) {
                 // Initialise data from session store if present
                 $data = $_SESSION;
                 // Merge in existing in-memory data, taking priority over session store data
                 $this->recursivelyApply((array)$this->data, $data);
-            } else {
-                // Use in-memory data if the session is lazy started
-                $data = $this->data;
+                $this->data = $data;
             }
-            $this->data = $data ?: [];
         } else {
+            // Use in-memory data if the session is lazy started
             $this->data = [];
-        }
-
-        // Modify the timeout behaviour so it's the *inactive* time before the session expires.
-        // By default it's the total session lifetime
-        if ($timeout && !headers_sent()) {
-            Cookie::set(session_name(), session_id(), $timeout/86400, $path, $domain ? $domain
-                : null, $secure, true);
         }
 
         $this->started = true;
@@ -497,9 +493,7 @@ class Session
 
     /**
      * Set user agent key
-     *
-     * @param HTTPRequest $request
-     */
+*/
     public function finalize(HTTPRequest $request)
     {
         $this->set('HTTP_USER_AGENT', $this->userAgent($request));
--- a/src/Control/Session.php
+++ b/src/Control/Session.php
@@ -494,7 +494,7 @@ class Session
 
     /**
      * Set user agent key
-*/
+     */
     public function finalize(HTTPRequest $request)
     {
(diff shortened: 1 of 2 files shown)


src/Control/Session.php
```diff
@@ -321,29 +315,31 @@ public function start(HTTPRequest $request)
...
 
             if (isset($_SESSION)) {
                 // Initialise data from session store if present
                 $data = $_SESSION;
                 // Merge in existing in-memory data, taking priority over session store data
                 $this->recursivelyApply((array)$this->data, $data);
-            } else {
-                // Use in-memory data if the session is lazy started
-                $data = $this->data;
+                $this->data = $data;
```
madmatt: Is this change (swapping the order) intentional? Seems like it might be accidental?

Regardless, can we get a comment to describe what it's doing? It's not entirely clear at a glance what it does if session isn't set (vs. the exact same line 4 lines down on L340 but with a blank array).

src/Control/Session.php
```diff
@@ -497,9 +493,7 @@ public function getAll()
 
     /**
      * Set user agent key
-     *
-     * @param HTTPRequest $request
-     */
+*/
```
madmatt: Missing whitespace here (unless it's GitHub messing up the preview)

tests/php/Control/SessionTest.php
```diff
@@ -335,4 +335,35 @@ public function testSave()
             $_SESSION
         );
     }
+
+    /**
+     * @runInSeparateProcess
+     * @preserveGlobalState disabled
+     */
+    public function testSessionOutput()
```
madmatt: Isn't an additional issue here that the first request that requires session should set a cookie, and then subsequent requests should *not* re-set the same cookie unless the info in the cookie has changed?

This test only seems to check that a single cookie is sent one time.

src/Control/Session.php
caffeineinc: yeah, you can only see what i've done with a comparison against the original. Each if else block sets `$data` and then sets it back to `$this->data = $data` at the end. Instead i'm doing the assignment once

src/Control/Session.php
caffeineinc: done

tests/php/Control/SessionTest.php
caffeineinc: That's correct, i can only test the first scenario, as session_start emits the cookie, You can't actively test session resume in phpunit, as you'd need to start a php session, and then resume it in a new  thread?  - To my knowledge i don't think that's possible.

src/Control/Session.php
caffeineinc: I've pushed some changes to resolve this :)

tests/php/Control/SessionTest.php
madmatt: You can do functional tests though still, right? Extend `FunctionalTest`, then make multiple calls to `$this->get('/')` and inspect the responses each time to see what the header data is...

Having said that, I'm not sure if that will work as it doesn't do a full request/response send I don't think (so PHP won't necessarily force-send its own cookie headers via `session_start()`...

src/Control/Session.php
```diff
@@ -321,28 +316,30 @@ public function start(HTTPRequest $request)
...
+                "cookie_domain" => $domain ?: "",
+                "cookie_secure" => $secure,
+                "cookie_httponly" => true
+            ];
+
+            if ($timeout) {
+                $sessionParameters["cookie_lifetime"] = $timeout;
+            }
+
+            session_start($sessionParameters);
```
dnsl48: The `options` argument for `session_start` [has been introduced](http://php.net/manual/en/function.session-start.php) in `PHP 7.0`.
I believe we should keep compatibility with `PHP 5.6` for a while yet.

tests/php/Control/SessionTest.php
dnsl48: This feels like the server behaviour testing rather then unit. I'm not sure if it's worth it, but we potentially could consider testing it in behat tests.

src/Control/Session.php
ScopeyNZ: I _think_ that providing these params in 5.6 is just a no-op - meaning that this wouldn't break 5.6 but the bug won't be fixed either.

dnsl48: Yes, PHP allows passing extra/undeclared arguments into a function. What I meant is we should still keep using `session_set_cookie_params` for a while.

tests/php/Control/SessionTest.php
caffeineinc: Yes, there's little point in unit testing this, as we can only test one of the behaviours. We can't test session resume functionality as it's core PHP functionality. We can't test this from behat either, as the session flag `httponly` is set, so session cookies aren't readable from javascript (for security reasons).  I wouldn't recommend overloading that just for testings sake. 

The only way to really test the resume behaviour is make a request using curl or similar.

```
 // should set session cookie
curl -svo /dev/null <my-url>

// should reply without a cookie parameter as the session is resumed.
curl -svo /dev/null <my-url> --cookie "mycookiename=<mycookievalue>;"

```

Any other ideas? or do we just accept it as is? @dnsl48 @madmatt

dnsl48: I didn't dig into it yet, but I think we use `chromedriver` for behat, which may give us access to httponly cookies.

dnsl48: Ok, I've investigated it and it appears we don't actually use ChromeDriver, but rather our own implementation named `FacebookWebDriver` and facebook `RemoteWebDriver`, which seem to be custom implementations of `Selenium2` driver.

Here's the table with capabilities of drivers: http://mink.behat.org/en/latest/guides/drivers.html

Unless we migrate to ChromeDriver we won't be able to access headers. However, I think that's obviously not within the context of this issue.

caffeineinc:
OK I have to close this PR to open a new one from the correct base branch.

Issue #3080, opened 2014-04-29
BUG get_declared_permissions_list() always returns NULL


phptek:
`Permission::get_declared_permissions_list()` always returns null becuase it relies on `Permission::$declared_permissions` which by default is `null`. What's more no code in `framework` or `cms` seems to refer to either this method or to the property.

Has this method ever worked? I see the only place this would likely be used in the CMS is in the Security admin, but in fact the logic used to show permission descriptions is on `Member::getCMSFields()` and uses a `PermissionCheckboxSetField_Readonly` field to do the job.

# Related Pull Request
* [ ] https://github.com/silverstripe/silverstripe-framework/pull/8558

kinglozzer:
Marked as docs because if it is used anywhere, it’s certainly not documented well enough 😉

dhensby:
Is this actually going to be addressed?

robbieaverill:
I've only had a quick skim of this code but I think there's some minor refactoring that could be done (at least) here. E.g. using `$this->config()->get('...')` over `self::$config_prop_name` (SS 4.2.x)

phptek:
Wow. I was still firmly in my 30s when I wrote this...

maxime-rainville:
I added the `affects/4` tag because it's still relevant to SS4. It's not quite clear to me what this thing is meant to do.

There's the following methods and properties that only reference themselves:
* `private static $declared_permissions = null;`
* `private static $declared_permissions_list = null;`
* `public static function get_declared_permissions_list()`
* `public static function get_label_for_permission($perm)`
* `protected static function traverse_declared_permissions($declared, &$list)`

No other files anywhere else references those methods.

My vote would be to deprecate the whole lot.

maxime-rainville:
I did https://github.com/silverstripe/silverstripe-framework/pull/8558 to add deprecation notices.

GuySartorelli:
The deprecation PR was merged, and the deprecated code will be removed in the next major version.
This issue doesn't need to be open any more.

PR #8558, opened 2018-10-31
API Deprecating Permission::$declared_permissions and related methods/props

maxime-rainville:
The following methods and props on Permission don't appear to be used anywhere any more.
```
    private static $declared_permissions = null;
    private static $declared_permissions_list = null;
    public static function get_declared_permissions_list()
    public static function get_label_for_permission($perm)
    protected static function traverse_declared_permissions($declared, &$list)
```

# Parent issue
* https://github.com/silverstripe/silverstripe-framework/issues/3080


maxime-rainville:
@robbieaverill Updated the `@deprecated` comments to conform to PSR-19.

PR #8560, opened 2018-11-01
NEW Deprecate RandomGenerator::generateEntropy in favour of using random_bytes directly

### LINKED ISSUE #8527: RFC: Deprecate and remove RandomGenerator::generateEntropy
## Affected Version

4.3.x / 5.0

## Description

Currently this function takes the function `random_bytes` and uses it to generate 64 bytes of entropy and converts it into a 128 bytes of hexadecimal. Given that `random_bytes` generates the same errors as this method when there is not a genuine source of CSPRNG it seems this method is pretty useless - and also a little misleading as you get 128 bytes but only 64 bytes of entropy. Additionally the docblock indicates you get an ASCII string (implying base64?) but you actually only get base16.

## Proposal

Deprecate this method
Replace usages with direct calls to `random_bytes` (only used in core by password encryptors AFAICT)

If this is unsuitable we should at least remove the `bin2hex` and indicate that it will just return random bytes.

## PR
#8560

ScopeyNZ:
RFC at #8527 . This function is only used within the same class for tokens but seems to be a more ambiguous function than just using `random_bytes` directly.

Fixes #8527

### DIFF
--- a/src/Security/RandomGenerator.php
+++ b/src/Security/RandomGenerator.php
@@ -14,7 +14,7 @@ class RandomGenerator
     /**
      * @return string A 128-character, randomly generated ASCII string
      * @throws Exception If no suitable CSPRNG is installed
-     * @deprecated 4.4:5.0
+     * @deprecated 4.4.0:5.0.0
      */
     public function generateEntropy()
     {
--- a/src/Security/RandomGenerator.php
+++ b/src/Security/RandomGenerator.php
@@ -4,6 +4,7 @@ namespace SilverStripe\Security;
 
 use Error;
 use Exception;
+use SilverStripe\Dev\Deprecation;
 
 /**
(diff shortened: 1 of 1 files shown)


src/Security/RandomGenerator.php
```diff
@@ -13,9 +14,12 @@ class RandomGenerator
     /**
      * @return string A 128-character, randomly generated ASCII string
      * @throws Exception If no suitable CSPRNG is installed
+     * @deprecated 4.4:5.0
```
robbieaverill: I think these need to be full versions:

```suggestion
     * @deprecated 4.4.0:5.0.0
```

robbieaverill:
>Do we make use of generateEntropy in any supported modules?

@sminnee nope =)

Issue #8343, opened 2018-08-26
Tool to maintain repo and github metadata across all supported modules


chillu:
Create a script which syncs up standard metadata across our supported modules. We've gone past the point where this is feasible to achieve manually. See Robbie's suggestion in https://forum.silverstripe.org/t/improving-the-way-we-maintain-the-silverstripe-open-source-ecosystem/859.

# Acceptance Criteria

 * Uses list of supported modules from the common JSON format
 * Can choose to run specific commands only (e.g. only update labels)
 * Runs in a few minutes (e.g. does sparse Github checkouts?)
 * Has an optional list where you can opt-out of specific modules for specific commands (e.g. those with a different license)
 * Uses local user for Github commits (assumes commit rights)

# Script features

 * Github labels (see [existing script](https://gist.github.com/chillu/45f9e9b777016f6257a64026aa84e0b3))
 * Contributing.md file (see [framework example](https://github.com/silverstripe/silverstripe-framework/blob/4/CONTRIBUTING.md))
 * [Issue and pull request templates](https://blog.github.com/2016-02-17-issue-and-pull-request-templates/) (see [framework example](https://github.com/silverstripe/silverstripe-framework/tree/4/.github))
 * LICENSE

# Notes

 * `.travis.yml` files are too complex for now

robbieaverill:
I've made a start on this with a Symfony console application for syncing labels to all supported modules: https://github.com/creative-commoners/maintain. The logic is based on https://gist.github.com/chillu/45f9e9b777016f6257a64026aa84e0b3 and it will pull the modules list from https://raw.githubusercontent.com/silverstripe/supported-modules/gh-pages/modules.json. I'll change this to use https://addons.silverstripe.org/api/supported-addons at some point, although that's a composer package name and we really need the GitHub slug

robbieaverill:
Cross referencing:

* Sync labels command: https://github.com/silverstripe/cow/issues/112
* Sync other stuff command: https://github.com/silverstripe/cow/issues/113

robbieaverill:
I've made a start with the file sync command in https://github.com/silverstripe/cow/pull/114. It'll currently pull the license from GitHub: https://github.com/silverstripe/supported-modules/blob/gh-pages/templates/LICENSE.md, then write, stage, commit and push any changes for each supported module.

[...]

PR #8569, opened 2018-11-05
Port betterbuttons to framework

bergice:
This adds `Previous`, `Next` and `Add` buttons to the edit form. It should also retain the search filters from the grid field where it was opened from.

Fixes #436

### DIFF
--- a/_config/buttons.yml
+++ b/_config/buttons.yml
@@ -1,0 +1,8 @@
+---
+Name: buttons
+---
+SilverStripe\Forms\GridField\GridFieldDetailForm_ItemRequest:
+  formActions:
+    showPrevious: true
+    showNext: true
+    showAdd: true
--- a/_config/buttons.yml
+++ b/_config/buttons.yml
@@ -3,6 +3,5 @@ Name: buttons
 ---
 SilverStripe\Forms\GridField\GridFieldDetailForm_ItemRequest:
   formActions:
-    showPrevious: true
-    showNext: true
+    showPagination: true
     showAdd: true
--- a/src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
+++ b/src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
@@ -308,10 +308,13 @@ class GridFieldDetailForm_ItemRequest extends RequestHandler
             $previousAndNextGroup->addExtraClass('rounded');
             $previousAndNextGroup->setFieldHolderTemplate(get_class($previousAndNextGroup) . '_holder_buttongroup');
 
+
+            $component = $this->gridField->getConfig()->getComponentByType(GridFieldDetailForm::class);
             $formActionsConfig = $this->config()->get("formActions");
-            $showPrevious = $formActionsConfig["showPrevious"];
-            $showNext = $formActionsConfig["showNext"];
-            $showAdd = $formActionsConfig["showAdd"];
+
+            $showPrevious = is_bool($component->getShowPrevious()) ? $component->getShowPrevious() : $formActionsConfig["showPrevious"];
+            $showNext = is_bool($component->getShowNext()) ? $component->getShowNext() : $formActionsConfig["showNext"];
+            $showAdd = is_bool($component->getShowAdd()) ? $component->getShowAdd() : $formActionsConfig["showAdd"];
 
             if ($showPrevious) {
                 $previousAndNextGroup->push(FormAction::create('doPrevious')
@@ -515,7 +518,6 @@ class GridFieldDetailForm_ItemRequest extends RequestHandler
             $this->gridField->Link(),
             "item",
             $id,
-            // todo: use http header instead
             '?gridState=' . urlencode($this->gridField->getState(false)->Value())
         );
     }
--- a/src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
+++ b/src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
@@ -310,36 +310,28 @@ class GridFieldDetailForm_ItemRequest extends RequestHandler
 
 
             $component = $this->gridField->getConfig()->getComponentByType(GridFieldDetailForm::class);
-            $formActionsConfig = $this->config()->get("formActions");
 
-            $showPrevious = is_bool($component->getShowPrevious()) ? $component->getShowPrevious() : $formActionsConfig["showPrevious"];
-            $showNext = is_bool($component->getShowNext()) ? $component->getShowNext() : $formActionsConfig["showNext"];
-            $showAdd = is_bool($component->getShowAdd()) ? $component->getShowAdd() : $formActionsConfig["showAdd"];
-
-            if ($showPrevious) {
+            if ($component->getShowPagination()) {
                 $previousAndNextGroup->push(FormAction::create('doPrevious')
                     ->setUseButtonTag(true)
                     ->setAttribute('data-grid-state', $gridStateStr)
                     ->setDisabled(!$this->getPreviousRecordID())
-                    ->addExtraClass('btn btn-secondary font-icon-left-open action--previous cms-panel-link'));
-            }
+                    ->addExtraClass('btn btn-secondary font-icon-left-open action--previous discard-confirmation'));
 
-            if ($showNext) {
                 $previousAndNextGroup->push(FormAction::create('doNext')
                     ->setUseButtonTag(true)
                     ->setAttribute('data-grid-state', $gridStateStr)
                     ->setDisabled(!$this->getNextRecordID())
-                    ->addExtraClass('btn btn-secondary font-icon-right-open action--next cms-panel-link'));
+                    ->addExtraClass('btn btn-secondary font-icon-right-open action--next discard-confirmation'));
             }
 
             $rightGroup->push($previousAndNextGroup);
 
-
-            if ($showAdd) {
+            if ($component->getShowAdd()) {
                 $rightGroup->push(FormAction::create('doNew')
                     ->setUseButtonTag(true)
                     ->setAttribute('data-grid-state', $this->getRequest()->getVar('gridState'))
-                    ->addExtraClass('btn btn-primary font-icon-plus rounded cms-panel-link'));
+                    ->addExtraClass('btn btn-primary font-icon-plus rounded action--new discard-confirmation'));
             }
         } else { // adding new record
             //Change the Save label to 'Create'
@@ -475,9 +467,9 @@ class GridFieldDetailForm_ItemRequest extends RequestHandler
      */
     public function doPrevious($data, $form)
     {
-        Controller::curr()->getResponse()->addHeader("X-Pjax", "Content");
+        $this->getToplevelController()->getResponse()->addHeader("X-Pjax", "Content");
         $link = $this->getEditLink($this->getPreviousRecordID());
-        return Controller::curr()->redirect($link);
+        return $this->redirect($link);
     }
 
     /**
@@ -488,9 +480,9 @@ class GridFieldDetailForm_ItemRequest extends RequestHandler
      */
     public function doNext($data, $form)
     {
-        Controller::curr()->getResponse()->addHeader("X-Pjax", "Content");
+        $this->getToplevelController()->getResponse()->addHeader("X-Pjax", "Content");
         $link = $this->getEditLink($this->getNextRecordID());
-        return Controller::curr()->redirect($link);
+        return $this->redirect($link);
     }
 
     /**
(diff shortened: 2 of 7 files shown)


_config/buttons.yml
```diff
@@ -0,0 +1,7 @@
+---
+Name: buttons
+---
+FormActions:
```
sminnee: I feel like this should be put into a better namespace, or at least something that makes it clearer that we're talking about form actions on the detail form of a grid field.

Also, it's not clear how I would override these yml defaults on a specific gridfield. There should probably be properties of the relevant gridfield component that can be overridden.

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -290,6 +294,54 @@ protected function getFormActions()
                     ->setUseButtonTag(true)
                     ->addExtraClass('btn-outline-danger btn-hide-outline font-icon-trash-bin action--delete'));
             }
+
+            if ($gridStateStr = ($this->getRequest()->getVar('gridState')
+                || $this->getRequest()->postVar('gridState'))) {
```
unclecheese: `->requestVar('gridState')` is all-inclusive.

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -290,6 +294,54 @@ protected function getFormActions()
...
+                $this->gridField->getState(false)->setValue($gridStateStr);
+                $this->owner->gridField->getState(false)->setValue($gridStateStr);
```
unclecheese: What's the `$this->owner` for? This isn't an extension.

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -290,6 +294,54 @@ protected function getFormActions()
...
+            }
+
+            $actions->push(HiddenField::create('gridState', null, $this->getRequest()->getVar('gridState')));
+
```
unclecheese: The initial check is against POST or GET, but this is only GET. Shouldn't this just be using `$gridStateStr`?

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -290,6 +294,54 @@ protected function getFormActions()
...
+            $previousAndNextGroup->setFieldHolderTemplate(get_class($previousAndNextGroup) . '_holder_buttongroup');
+
+            $showPrevious = Config::inst()->get("FormActions", "showPrevious");
+            $showNext = Config::inst()->get("FormActions", "showNext");
+            $showAdd = Config::inst()->get("FormActions", "showAdd");
+
+            if ($showPrevious) {
+                $previousAndNextGroup->push(FormAction::create('doPrev')
+                    ->setUseButtonTag(true)
+                    ->setAttribute('data-grid-state', $this->getRequest()->getVar('gridState'))
```
unclecheese: Same

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -290,6 +294,54 @@ protected function getFormActions()
...
+                    ->setDisabled(!$this->getPreviousRecordID())
+                    ->setAttribute('data-text-alternate', 'Previous')
+                    ->addExtraClass('btn btn-secondary font-icon-left-open action--previous'));
+            }
+
+            if ($showNext) {
+                $previousAndNextGroup->push(FormAction::create('doNext')
+                    ->setUseButtonTag(true)
+                    ->setAttribute('data-grid-state', $this->getRequest()->getVar('gridState'))
```
unclecheese: Same here

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -290,6 +294,54 @@ protected function getFormActions()
...
+                    ->addExtraClass('btn btn-secondary font-icon-right-open action--next'));
+            }
+
+            $rightGroup->push($previousAndNextGroup);
+
+
+            if ($showAdd) {
+                $rightGroup->push(FormAction::create('doNew')
+                    ->setUseButtonTag(true)
+                    ->setAttribute('data-grid-state', $this->getRequest()->getVar('gridState'))
```
unclecheese: Same here

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -410,6 +468,94 @@ public function doSave($data, $form)
...
+     * Goes to the next record
+     * @param  array $data The form data
+     * @param  Form $form The Form object
+     * @return HTTPResponse
+     */
+    public function doNext($data, $form)
+    {
+        Controller::curr()->getResponse()->addHeader("X-Pjax", "Content");
+        $link = $this->getEditLink($this->getNextRecordID());
+        return Controller::curr()->redirect($link);
```
unclecheese: Shouldn't need `Controller::curr()` here as we're in a controller right now. `$this` should work fine, I would think?

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -410,6 +468,94 @@ public function doSave($data, $form)
...
+    /**
+     * Gets the edit link for a record
+     *
+     * @param  int $id The ID of the record in the GridField
+     * @return string
+     */
+    public function getEditLink($id)
+    {
+        return Controller::join_links(
+            $this->owner->gridField->Link(),
```
unclecheese: Another `owner` here.

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -410,6 +468,94 @@ public function doSave($data, $form)
...
+            "item",
+            $id,
+            // todo: use http header instead
```
unclecheese: Either do this or remove it. My preference would be to remove it and leave it as is. I think the tradeoffs for using a header far outweigh the benefits of a tidy URL.

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -410,6 +468,94 @@ public function doSave($data, $form)
...
+            '?gridState=' . urlencode($this->owner->gridField->getState(false)->Value())
```
unclecheese: Is there ever a case for `gridState` being in the request body? If not, I would remove your check to `POST` above.

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -410,6 +468,94 @@ public function doSave($data, $form)
...
+     * @return int
+     */
+    public function getPreviousRecordID()
+    {
+        $gridField = $this->owner->gridField;
+        $gridStateStr = $this->getRequest()->postVar('gridState');
+        $gridField->getState(false)->setValue($gridStateStr);
+
+        $map = $gridField->getManipulatedList()->limit(PHP_INT_MAX, 0)->column('ID');
+        $offset = array_search($this->owner->record->ID, $map);
```
unclecheese: Remove the `owner`

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -410,6 +468,94 @@ public function doSave($data, $form)
...
+     * Creates a new record. If you're already creating a new record,
+     * this forces the URL to change.
+     *
+     * @param  array $data The form data
+     * @param  Form $form The Form object
+     * @return HTTPResponse
+     */
+    public function doNew($data, $form)
+    {
+        return Controller::curr()->redirect($this->owner->Link('addnew'));
```
unclecheese: `owner`

src/Forms/GridField/GridFieldEditButton.php
```diff
@@ -62,7 +62,13 @@ public function getExtraData($gridField, $record, $columnName)
      */
     public function getUrl($gridField, $record, $columnName)
     {
-        return Controller::join_links($gridField->Link('item'), $record->ID, 'edit');
+        return Controller::join_links(
+            $gridField->Link('item'),
+            $record->ID,
+            'edit',
+            // todo: use http header instead
```
unclecheese: Remove `todo`

_config/buttons.yml
unclecheese: Agreed. Let's also pin these to a concrete class, rather than against the arbitrary `FormActions`, which is really ambiguous.

```
SilverStripe\Forms\GridField\GridFieldDetailForm_ItemRequest:
  formActions:
    ...
```

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -290,6 +294,54 @@ protected function getFormActions()
...
+            $rightGroup->addExtraClass('right');
+            $rightGroup->setFieldHolderTemplate(get_class($rightGroup) . '_holder_buttongroup');
+
+            $previousAndNextGroup = CompositeField::create()->setName('PreviousAndNextGroup');
+            $previousAndNextGroup->addExtraClass('rounded');
+            $previousAndNextGroup->setFieldHolderTemplate(get_class($previousAndNextGroup) . '_holder_buttongroup');
+
+            $showPrevious = Config::inst()->get("FormActions", "showPrevious");
+            $showNext = Config::inst()->get("FormActions", "showNext");
+            $showAdd = Config::inst()->get("FormActions", "showAdd");
```
unclecheese: If you accept the config changes above, these will need to be refactored as `$this->config()->get()`.

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -410,6 +467,93 @@ public function doSave($data, $form)
...
+     */
+    public function getPreviousRecordID()
+    {
+        $gridField = $this->gridField;
+        $gridStateStr = $this->getRequest()->postVar('gridState');
+        $gridField->getState(false)->setValue($gridStateStr);
+
+        $map = $gridField->getManipulatedList()->limit(PHP_INT_MAX, 0)->column('ID');
+        $offset = array_search($this->record->ID, $map);
+        return isset($map[$offset-1]) ? $map[$offset-1] : false;
```
unclecheese: OK, this amount of computation, while good enough for my module, isn't good enough for core. If the grid has 5,000 records in it, this falls on its face. Maybe even at 500 or 1,000.

Not sure what to do here, but here's a thought

* Add `itemsPerPage` to the `GridFieldPaginator` state (currently only stores `currentPage`)
* Instead of using an unlimited query on `getManipulatedList`, use `$itemsPerPage + 2` for the limit (to allow for the last record on the previous page and the first record on the next page), and use `($itemsPerPage * $currentPage) - 1` as the offset. In other words, with 10 records per page, on page 5, it should be looking at records 49-61.
* Search for the ID per the next couple lines, as is currently shown.

What do you think?

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -410,6 +467,93 @@ public function doSave($data, $form)
...
+     *
+     * @return int
+     */
+    public function getNextRecordID()
+    {
+        $gridField = $this->gridField;
+        $gridStateStr = $this->getRequest()->postVar('gridState');
+        $gridField->getState(false)->setValue($gridStateStr);
+
+        $map = $gridField->getManipulatedList()->limit(PHP_INT_MAX, 0)->column('ID');
```
unclecheese: Goes without saying, but same change applies here. I think you'll want to encapsulate it into a method, like `$this->getCurrentPageWithAdjacentRecords()`

unclecheese:
The changes to `GridFieldPaginator` are likely incompatible with the #1 installed module in the SS ecosystem, [gridfieldextensions](https://addons.silverstripe.org/add-ons/symbiote/silverstripe-gridfieldextensions).

We haven't broken any APIs, so this is safe to merge into a minor release, but I think we should at least check in with the maintainers to let them know about this change so they can issue a 4.3 compatible release. CC @robbieaverill @NightJar

unclecheese:
Awesome. Tests were failing because `$state->getData('GridFieldPaginator')` wasn't always defined. Have put a guard in against that.

Other minor points:
* Consistency with single/double quotes. Prefer single, unless doing variable interpolation
* Private members should be used sparingly, as they're antithetical to extensibility.
* Don't assume config vars are set (`$formConfig['showAdd']`), as it tightly couples your code to the config.

Merge on green.

robbieaverill:
>Consistency with single/double quotes. Prefer single, unless doing variable interpolation

We should add this to phpcs rules. Related: https://github.com/silverstripe/silverstripe-framework/issues/8343

PR #8571, opened 2018-11-05
API LoginForm::authentiator_class is now deprecated, use getters or setters instead

robbieaverill:
This pull request deprecates the protected `$authenticator_class` property (since it's not PSR-2 compliant) and replaces it with an `$authenticatorClass` property. It also adds a getter so that implementations don't need to set the property directly.

### DIFF
--- a/src/Security/LoginForm.php
+++ b/src/Security/LoginForm.php
@@ -15,24 +15,46 @@ use SilverStripe\Forms\Form;
  */
 abstract class LoginForm extends Form
 {
-
     /**
-     * Authenticator class to use with this login form
-     *
-     * Set this variable to the authenticator class to use with this login
-     * form.
+     * @deprecated 4.4.0:5.0.0 Use getAuthenticatorClass() or setAuthenticatorClass() instead
      * @var string
      */
     protected $authenticator_class;
 
+    /**
+     * Authenticator class to use with this login form
+     *
+     * Set this variable to the authenticator class to use with this login form.
+     *
+     * @var string
+     */
+    protected $authenticatorClass;
+
+    /**
+     * Set the authenticator class name to use
+     *
+     * @param string $class
+     * @return $this
+     */
     public function setAuthenticatorClass($class)
     {
-        $this->authenticator_class = $class;
+        $this->authenticatorClass = $class;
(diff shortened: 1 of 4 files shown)


src/Security/LoginForm.php
```diff
@@ -15,26 +15,56 @@
  */
 abstract class LoginForm extends Form
 {
+    /**
+     * @deprecated 4.4.0:5.0.0 Use getAuthenticatorClass() or setAuthenticatorClass() instead
+     * @var string
+     */
+    protected $authenticator_class;
```
maxime-rainville: If someone has extended `LoginForm` and access `$this->authenticator_class` directly within their new class, won't this change break their code?

robbieaverill: Yeah true, I guess we need to read from this prop still from the places that were using it, that way B/C would be preserved.

e.g. `HiddenField::create("AuthenticationMethod", null, $this->authenticator_class ?: $this->getAuthenticatorClass(), $this),`

Would that make sense?

If devs are setting and reading it from extended classes in their own modules or project code, it would continue to work anyway. The break would be if someone hadn't set the property but was reading from it - it's protected, so would need to be done in the same hierarchy as this class

maxime-rainville: The only problem would be if someone relies on `$this->authenticator_class` being set by the base class but then decided to get directly from another method. I'm not sure there's a nice elegant way around this ... short of keeping the existing name.

Scenarios like this are why I think we should define more methods and properties as private.

robbieaverill: I've always been more for protected over private, but as long as we provide transparent getters and setters I guess there's no reason we shouldn't have private properties. Agreed that in the end this replacement would be too risky. We could deprecate the property without providing an immediate replacement, that might at least encourage people not to use the property directly?

robbieaverill:
Ready for re-review @maxime-rainville =)

robbieaverill:
I’m not familiar with such modern wizardry but if you don’t mind I’ll learn from your commit

maxime-rainville:
@robbieaverill Wizardry has been applied. I've rebased with 4 as well.

robbieaverill:
Rebased again, I'll merge on green to stop it going stale

PR #8573, opened 2018-11-05
Convert::memstring2bytes should preserve -1

### LINKED ISSUE #8570: `Environment::increaseMemoryLimitTo()` cannot increase memory to "unlimited"
## Affected Version

4.2.1

## Description

I cannot increase memory (e.g. in a dev task) to "unlimited", cause "-1" get's convertet to "1".

## Steps to Reproduce

Step debugger on `Environment::increaseTimeLimitTo();`

after it calls `Convert::memstring2bytes()` the default string of "-1", which means "unlimited" is convertet to an integer "1".

This regex in `Convert::memstring2bytes()` removes the minus sign:

grafik

Questions: Who is repsonsible to preserve the "-1" ? Should the logic go to the convert method or should increaseMemoryLimitTo handle that?

I'd vote for convert, what do the core commiters think about it?

As most of the time I'm happy to create a PR for that small but annoying issue

wernerkrauss:
fixes #8570

### DIFF
--- a/src/Core/Convert.php
+++ b/src/Core/Convert.php
@@ -553,6 +553,7 @@ class Convert
 
     /**
      * Turn a memory string, such as 512M into an actual number of bytes.
+     * Preserves integer values like "1024" or "-1"
      *
      * @param string $memString A memory limit string, such as "64M"
      * @return float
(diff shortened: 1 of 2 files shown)


unclecheese:
This looks like a bugfix. Could we target it to 4.0?

wernerkrauss:
Hmm, @unclecheese changing the base in the PR doesn't seem to do the trick. Do I have to cherry pick my commit in 4.0 or how can I acchieve that?

robbieaverill:
@wernerkrauss it just needs a rebase, I'll do it now for you

PR #8581, opened 2018-11-07
FIX Reverting ExtensionTestState and Extensible extra methods modifications to prevent PHP 5.6 segfault

robbieaverill:
~~This is a partial revert of #8465 which allows the original test it added to pass while preventing segmentation faults on PHP 5.6 in some cases (see https://github.com/silverstripe/recipe-content-blocks/pull/8).~~

~~I thought about making hasMethod() recurse once, but it would've involved a signature change. Since the addition of flush_extra_methods_cache() is unreleased, I'm calling this a patch change.~~

~~The logic wasn't quite correct - flushing all methods (or even those just for the given class in tearDownOnce()) is not sufficient. This change means that if a previously defined and cached list of extra methods returns false for some method, it will redefine and re-check once, allowing the cached values to be reset one time. I don't see this as having a significant performance hit, because hasMethod() returning false should be an edge case anyway.~~

~~Updated: switched to call flush_extra_methods_cache() on Extensible instead of DataObject~~ ~~Reverted back~~ Reverted back again

~~add/remove_extension calls reset the extra_methods cache for the object you're applying the extension to. The original bug was from modifying extension class config after those had been generated, so extra methods on the overloaded class were not available.~~

I've done some testing, and I think we should revert all of these ~~and change the unit test to call Member::add_extension(SomeExtension::class) after it changes the config for that extension~~. This fixes the issue (I'll also pass on the message to the key project that had this problem).


kinglozzer:
I’ve been trying to recreate this with `silverstripe/recipe-content-blocks` locally, and can fix the segfault in PHP 5.6 by changing the calls to `DataObject::flush_extra_methods_cache();` to `Extensible::flush_extra_methods_cache();`. Could you check if that fixes the issue for you before we merge this?

robbieaverill:
Yep, I'll test that now. That'd be a weird PHP bug if so, but I guess most PHP version specific segfaults are weird 😆thanks for reviewing!

robbieaverill:
Confirmed - I'll adjust the PR

robbieaverill:
Ah, but this doesn't work with the original test that was added with #8465 - `testExtendedExtensions`. I'm going to restore the PR to where it was =/

src/Core/CustomMethods.php
```diff
@@ -145,7 +145,15 @@ protected function registerExtraMethodCallback($name, $callback)
      */
     public function hasMethod($method)
     {
-        return method_exists($this, $method) || $this->getExtraMethodConfig($method);
+        $hasMethod = method_exists($this, $method) || $this->getExtraMethodConfig($method);
+
+        if (!$hasMethod) {
```
sminnee: If hasMethod is called on a method that doesn't exist, we're adding some expensive code here. hasMethod returning false could be designed behaviour on a site, and this change risks materially slowing that site down.

I'd treat this with extreme caution. I'm not sure that this is the right fix, and I think that this would be too risky for a late-stage 4.3 change, and too risky for a feature that merely tidies up tests. :-/

sminnee: I'm a little confused as to why this is necessary.

 * Extensible::add_extension() and Extensible::remove_extension() reset the custom methods for the named class and all subclasses, forcing a defineMethods() call the next time getExtraMethodConfig() is called.

robbieaverill: Yeah ok, fair enough. The background here is this:

* #8465 allows for config for extensions to be overloaded with injector after the extra_methods have been calculated for the original. This was fixing a bug (shown in the unit tests added in that PR) where an overloaded extension was not having its extra methods registered in a key project's unit test suite - it was global state leakage between tests - this came down to the fact that the extra methods for the original Extension class had already been defined in setUpOnce (or similar)
* #8505 fixes a bug in advanced workflow that was caused by the changes to ExtensionTestState in #8465. It was a defensive change, but also added the extra call to reset the extra methods between tests.
* This PR was fixing a seg fault in PHP 5.6 unit tests, and ensuring that the original test added in #8465 would still pass by doing a second call to defineMethods during the hasMethod() check (this is what you're pointing out might be risky)

You're right that add/remove_extension calls reset the extra_methods cache for the object you're applying the extension to. The original bug was from modifying extension class config after those had been generated, so extra methods on the overloaded class were not available.

I've done some testing, and I think we should revert all of these and change the unit test to call `Member::add_extension(SomeExtension::class)` after it changes the config for that extension. This fixes the issue (I'll also pass on the message to the key project that had this problem).

I'm actually not confident that the changes I made are correctly fixing the problem, there's still the odd test state leak happening and it's hard to reproduce in isolation.

Anyway, new commit incoming. Thanks for reviewing!

sminnee: Yeah, this makes sense, I think the number of follow-on PRs suggests that a revised approach is needed. As you've hinted, perhaps we should initially revert _all_ these PRs and set up a new one with tests that capture all you've learned from the previous PRs? In particular, we probably want to get back to a consistent state before 4.3 goes stable.

> This allows for config for extensions to be overloaded with injector after the extra_methods have been calculated for the original. This was fixing a bug (shown in the unit tests added in that PR) where an overloaded extension was not having its extra methods registered in a key project's unit test suite - it was global state leakage between tests - this came down to the fact that the extra methods for the original Extension class had already been defined in setUpOnce (or similar)

I assume that we're talking specifically about within tests, right? Was the test using $required_extensions, and was that broken? Or was the issue that you wanted to apply the extension for a single test as opposed to a whole test class.

I would recommend that we clearly document the limitations of setting the `extensions` configuration property; this will mitigate the risk of us not being able to make it magically do everything. ;-)

robbieaverill: >As you've hinted, perhaps we should initially revert all these PRs and set up a new one with tests that capture all you've learned from the previous PRs

Yep totally - this PR now does that (it leaves the existing test that was added to show how overloaded extensions can be used _after_ extra methods are defined for the class it is applied to.

>In particular, we probably want to get back to a consistent state before 4.3 goes stable.

100%

>I assume that we're talking specifically about within tests, right? Was the test using $required_extensions, and was that broken? Or was the issue that you wanted to apply the extension for a single test as opposed to a whole test class.

Yep that's right. The test was _not_ using $required_extensions - the report was that the test was passing in 4.2 and started to break in 4.3. Adding $required_extensions fixed the problem, but the key project was concerned about why it changed since it previously worked, and the extension in question was common in a lot of tests. I've passed a message on to the project that this change is likely to be reverted and that they should use $required_extensions to ensure their tests keep passing.

>It could, for example, be a case of saying "if you've modified the extensions property for a class, you should call SomeClass::someResetStaticMethod()"

Yeah so that's what I'd originally added - `SomeClass::flush_extra_methods_cache()`, but it's ended up causing PHP 5.6 segfaults for some reason (in tests). It's quite a complicated architecture under there, with DataObject using a trait which uses another trait. As you said, we want a consistent state for 4.3 so I think it's safer to revert back to the way it was before and maybe re-assess it as a wider SapphireTest refactor at some point in the future.

We could add some docs saying something like "if you're changing an extension class with Injector after the config manifest has been loaded and after the class you're extending has already been loaded, ensure you call `SomeClass::add_extension(TheExtensionYoureExtending::class)` afterwards to regenerate the extra methods cache" - that's the problem we're trying to fix here

sminnee: It is curious that 4.3 caused a regression here, if it was working in 4.2 - do we know what change from 4.2 originally introduced this problem?

robbieaverill: I’d like to find some time to investigate for sure

robbieaverill:
Bump - I'll spend a little bit of time tomorrow trying to work out what changed between 4.2 and 4.3 in this area (if anything - reports were second hand), but I think this PR now does what was agreed on as the best approach; reverting the changes to ExtensionTestState and retaining the original test

tests/php/Core/Injector/InjectorTest.php
```diff
@@ -1062,6 +1062,11 @@ public function testExtendedExtensions()
                 SomeExtension::class,
             ]);
 
+        // Since we've changed config for the extension's class name after it has been loaded (via
```
sminnee: This change essentially invalidates this test. My read is that the feature ostensibly tested by this test isn't actually available. I'm surprised that it ever was but apparently it _was_ available in 4.1 (according to @robbieaverill if I understood a slack convo with him correctly).

I would recommend:
 * Removing this test altogether and noting in the docs that dynamically changing the extensions config option doesn't work, AND/OR
 * Putting this test into 4.1 and, if it passes, trying to work out what change broke it.

Essentially I'm recommending a _full_ revert of #8465.

Alternatively, we retain Extensible::flush_extra_methods_cache(), document that method as necessary to be called after dynamic modification of the extensions property, and call DataObject::flush_extra_methods_cache() here instead.

To be clear, I would retain the add-extensions / remove-extensions material that this PR re-introduces, which means that it's _not_ necessary to call DataObject::flush_extra_methods_cache in the setUp(). Rather, it only necessary to call if it you're making changes _between_ tests.

sminnee: I've had a look at this. I think the pass in 4.0, 4.1, 4.2 is bogus.

In particular, I think that it's only working if `Config::modify()` on Member extensions is called before any method-exist checks are called on Member.

Notably, this test fails (in 4.1 at least)
```diff
    /**
     * Tests that overloaded extensions work, see {@link Extensible::getExtensionInstance()}
     */
    public function testExtendedExtensions()
    {
+        $member = new Member();
+        $this->assertFalse($member->hasExtension(SomeExtension::class));
+
        Config::modify()
            ->set(Injector::class, SomeExtension::class, [
                'class' => SomeCustomisedExtension::class,
            ])
            ->merge(Member::class, 'extensions', [
                SomeExtension::class,
            ]);

        /** @var Member|SomeExtension $member */
        $member = new Member();
        $this->assertTrue($member->hasExtension(SomeExtension::class));
        $this->assertSame('bar', $member->someMethod());
    }
```

I haven't validated which change to SapphireTest causes this to fail – it's probably something about initialising Members – but my view is that this feature has therefore always been unreliable in 4.x and it's fine to simply document that it doesn't work.

Put another way, I think this might break tests going from 4.2 to 4.3, but those tests that were passing in 4.2 were relying on an unreliable code-path.

robbieaverill: @sminnee OK, thanks for following up - sounds like we're OK about reverting all of this and making it a documentation issue?

robbieaverill:
Ok, thanks to everyone for all the feedback on this issue. Now this PR:

* Reverts #8465 
* Reverts half of #8505 (leaving the check for null extension lookups, which is a valid API return value)
* Updates the Extending docs to explain how to extend extensions correctly (with Injector config in YAML)
* Adds a note to avoid using `Config::modify()->set()` to do this in unit tests to the 4.3.0 changelog

tractorcow:
Ah, yeah I can see how this would be a problem. :)

It's not just enough to add extensions via config changes. It has to be done via add_extension(), if not registered in the "at boot" yml config.

You could make it work that way though, but the whole system would be a lot slower if dynamically resolving extensions each time.

robbieaverill:
Thanks everyone

PR #8583, opened 2018-11-07
Convert::memstring2bytes should return integer value

### LINKED ISSUE #8572: Convert::memstring2bytes returns a float
## Affected Version

4.x

## Description

`Convert::memstring2bytes` returns a float value due to the usage of `round()`. IMHO that does not make sense, as bytes are by nature integer.

If I provide a patch, against which branch should I commit the PR?

wernerkrauss:
bytes are by nature an integer

fixes #8572

### DIFF
--- a/src/Core/Convert.php
+++ b/src/Core/Convert.php
@@ -556,7 +556,7 @@ class Convert
      * Preserves integer values like "1024" or "-1"
      *
      * @param string $memString A memory limit string, such as "64M"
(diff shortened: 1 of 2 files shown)


unclecheese:
FYI This broke the `silverstripe/assets` test, which is expecting a float. https://github.com/silverstripe/silverstripe-assets/blob/1/tests/php/FileTest.php#L701

PR #8584, opened 2018-11-07
Cleanup Convert

wernerkrauss:
- remove unneeded parenthesis and double quotes
- simplify flow / remove unneeded else


tests/php/Core/ConvertTest.php
```diff
@@ -224,7 +224,7 @@ public function testArray2JSON()
           'Complicated' => 'Structure'
          )
         );
-        $encoded = Convert::array2json($val);
+        $encoded = json_encode($val);
```
robbieaverill: These are testing the methods you've replaced, they are deprecated but should stay as they were

wernerkrauss: good catch, sorry

PR #8587, opened 2018-11-08
FIX Move password complexity requirements into framework

robbieaverill:
The contents of this file are not extensible, so I'm proposing we move it into framework YAML config

Issue: https://github.com/silverstripe/cwp-core/issues/52

Requires recipe-core: https://github.com/silverstripe/recipe-core/pull/34

### DIFF
--- a/_config/passwords.yml
+++ b/_config/passwords.yml
@@ -1,0 +1,13 @@
+---
+Name: corepasswords
+---
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\Security\PasswordValidator:
+    properties:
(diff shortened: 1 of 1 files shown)


kinglozzer:
I think we originally introduced this in the recipe so it wouldn’t affect existing projects, is storing it in YAML in the recipe an option?

robbieaverill:
Sort of:

* Yes, it can be in a YAML file
* The name for it could be `#mypasswords`
* Module code can't reliably say it wants to run after this

We could add `After: '*'` but we've had problems with that in the past

Issue #8426, opened 2018-10-01
Eager loading for DataObject relationships


unclecheese:
`DataList` does not allow caching or joining of relationships, particularly `has_many` and `many_many` relationships, which leads to an [N+1 problem](https://stackoverflow.com/questions/97197/what-is-the-n1-select-query-issue)
in many contexts.
```
<% loop $Teams %>
$Title
	<% with $Schedule %>
		$Title
		<% loop $Matches %>
			$Score
			Attendees
			<% loop $Attendees %>
				$Name
			<% end_loop %>
		<% end_loop %>
	<% end_with %>
<% end_loop %>
```

If there are 50 teams and each team has 10 matches, this results in **601 queries**.

While DataObjects do a good job of caching `has_one` relationships via `getComponent`, plural relationships are not cached, and joining their respective tables on to the query doesn't help. Each time one of those methods is called, it results in a new query.

[EDIT]
This has a massive impact on GraphQL, where nested queries are encouraged.

```
query readTeams {
  Name
  Schedule {
    Title
    Matches {
        Score
        Attendees {
            Email
        } 
    }
  }
}
```

## Proposal

Allow eager loading of relationships and stuff them into memory somewhere, and apply them to dataobjects when they're created.

## Proof of concept

https://github.com/unclecheese/silverstripe-eager-loader

## For discussion

How are other frameworks/ORMs solving this issue?

## Acceptance criteria
- You have the ability to eager load relationships
- Changelog highlights this new feature.
- ORM doc is updated with a new eager loading section.
- Lower level APIs are marked as `@internal` where appropriate.
- 
## Notes
- There's probably some scope to improve CMS using this new feature. This should be done in a follow up card and not part of this card.

## PRs
- https://github.com/silverstripe/silverstripe-framework/pull/8687
- https://github.com/silverstripe/silverstripe-framework/pull/10783
- https://github.com/silverstripe/developer-docs/pull/277

sminnee:
I really like this, but given the importance of GraphQL to the current but certainly the future performance of the CMS, I think we'd want to consider how any kind of N+1 avoidance also fits with GraphQL.

From what I understand from `http://webonyx.github.io/graphql-php/data-fetching/#solving-n1-problem` this would need to fit with the use of `GraphQL\Deferred` somehow. It's possible that eager loading specifically isn't the right solution here, but how much would the solution overlap with eager loading?

unclecheese:
[...]

PR #8588, opened 2018-11-09
FIX: Allow filtering of a DataList by subclass fields

### LINKED ISSUE #1683: [ORM] [2013-02-20] Children can't be filtered by non SiteTree fields
created by: @dhensby (Pigeon)
created at: 2013-02-20
original ticket: http://open.silverstripe.org/ticket/8272

---

When accessing children on a page like so:

```
$this->Children()
```

The RelationList can't be filtered on a DB field that is not on the SiteTree table as the SQL builder only looks at ancestors of SiteTree (the defined class in the has_many) for fields.

Example:

```
$this->Children()->Filter('ShowInFooter',1);
```

throws error:

```
[User Error] Couldn't find field ShowInFooter in any of SiteTree's tables.
```

Using setDataModel('Page') didn't help :(

sminnee:
FIX: Allow generating DataList aggregates of subclass fields
NEW: Add DataQuery::joinSubclassTableForField()
NEW: Add SearchFilter::getDbNameAndAddToQuery()

Fixes #1683

### DIFF
--- a/src/ORM/DataObjectSchema.php
+++ b/src/ORM/DataObjectSchema.php
@@ -669,7 +669,7 @@ class DataObjectSchema
     }
 
     /**
-     * Returns the table name in the class hierarchy which contains a given
+     * Returns the table name in the class ancestry which contains a given
      * field column for a {@link DataObject}. If the field does not exist, this
      * will return null.
      *
@@ -688,7 +688,30 @@ class DataObjectSchema
     }
 
     /**
-     * Returns the class name in the class hierarchy which contains a given
+     * Returns the table name(s) in the class ancestry and descendants which contain a given
+     * field column for a {@link DataObject}. If the field does not exist, this will return null.
+     *
+     * @param string $candidateClass
+     * @param string $fieldName
+     *
+     * @return string
+     */
+    public function tablesForField($candidateClass, $fieldName)
+    {
+        $classes = $this->classesForField($candidateClass, $fieldName);
+        if ($classes) {
+            return array_map(
+                function ($class) {
+                    return $this->tableName($class);
+                },
+                $classes
+            );
+        }
+        return null;
+    }
+
+    /**
+     * Returns the class name in the class ancestors which contains a given
      * field column for a {@link DataObject}. If the field does not exist, this
      * will return null.
      *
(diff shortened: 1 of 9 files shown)


sminnee:
Paging @tractorcow and @dhensby since it's an ORM change

src/ORM/DataObjectSchema.php
```diff
@@ -688,7 +688,30 @@ public function tableForField($candidateClass, $fieldName)
...
-     * Returns the class name in the class hierarchy which contains a given
+     * Returns the table name(s) in the class ancestry and descendants which contain a given
+     * field column for a {@link DataObject}. If the field does not exist, this will return null.
+     *
+     * @param string $candidateClass
+     * @param string $fieldName
+     *
+     * @return string
+     */
+    public function tablesForField($candidateClass, $fieldName)
```
sminnee: This duplicates a lot of tableForField, which
 * Only looks at the current class and its ancestors
 * Can more confidently return a single field

Whereas this
 * Looks at the descendants as well as ancestors
 * Returns an array of fields, although quite likely an array of 1

We may want to look at deprecating the original version?

src/ORM/Filters/SearchFilter.php
```diff
@@ -309,6 +309,32 @@ public function getDbName()
         return '"' . implode('"."', $parts) . '"';
     }
 
+    /**
+     * Return getDbName() and ensures that the passed query has all the joins necessary to reference it.
+     *
+     * @param $query The DataQuery to add joins to
+     * @return string A table-qualified, quoted field name
+     */
+    public function getDbNameAndAddToQuery(DataQuery $query)
```
sminnee: This name is a mouthful, it was intended to clarify that it both modifying the query and then returning a reference to that field on the modified query.

Another approach to this would be to refactor this to live more on the DataQuery object itself, e.g.

```
$query->provideAccessToField($this->name)
```

```
$query->provideAccessToAggregate($this->aggregate['function'], $this->aggregate['column'])
```

This might also mean that joinSubclassTableForField doesn't need to be added to the public API anymore. It would mean that the SearchFilters need more refactoring, and it might mean that we want to formally deprecate getDbName() as well.

tractorcow:
I have another way I would solve this problem. Comment on the other issue.

src/ORM/Filters/SearchFilter.php
tractorcow: Everything I've added has been overly verbose, so it stays in character. :)

src/ORM/DataObjectSchema.php
tractorcow: Note, DataQuery (not datalist) already had (still has?) logic in it to cover querying of subfields of tables.

This actually caused issues where you had multiple sub-tables that had the same field names as siblings. E.g. `Banner`, `HeaderBanner extends Banner has_one Image`, `FooterBanner extends Banner has_one Image`.

Look up `collidingFields` in DataQuery.

I don't think it's impossible to solve this, but I was frustrated with the complexity and how convoluted the code needed to be to handle these situations.

In order for the ORM to get better it probably needs to be better at managing this complexity by being picky about what problems we want to solve. :) It's purely opinion but I think my `->ofType()` solution is a lot simpler.

tractorcow: https://github.com/silverstripe/silverstripe-framework/issues/1683#issuecomment-439180349

sminnee: > Note, DataQuery (not datalist) already had (still has?) logic in it to cover querying of subfields of tables.

True, I might look at refactoring it into here.

> It's purely opinion but I think my ->ofType() solution is a lot simpler.

Simpler for us, harder for users.

src/ORM/Filters/SearchFilter.php
sminnee: What do you think about shifting more of this responsibility from SearchFilter to DataQuery?

src/ORM/DataObjectSchema.php
sminnee: > This actually caused issues where you had multiple sub-tables that had the same field names as siblings.

That's why it's tablesForField() and not tableForField(). ;) Thankfully(?) our test suite actually contains these cases so I identified it relatively early. This PR hands such changes.

Although, now that I mention it, my test cases were only for colliding _fields_ not colliding _relations_.

IMO we should probably throw a warning/deprecation when you're using this annoying capability, but that's a separate ticket.

sminnee:
I want to flag with @unclecheese that this PR is all about providing more tools for query manipulation, and as such might overlap with eager loading work. It might be useful to ask "what additional APIs does DataQuery need for both use-cases?"

Cheddam:
@sminnee Does this still make sense to pursue, or should we close it out? (Eager loading still hasn't shipped 😅)

sminnee:
I'm going to close this as rendered obsolete by #8426 - let's aim to get that merged instead!

Issue #3473, opened 2014-09-07
[ORM] currency field that is set to zero evaluates to TRUE 


sunnysideup:
currency field that is set to zero evaluates to TRUE.  This is confusing to say the least - as it goes against all the basics in PHP.   To replicate
- set up dataobject

``` php

class MyProduct extends DataObject {
    private static $db = array(
        "MyPrice" => "Currency",
        "MyDouble" => "Double"
    );
}
```
-  setup test 

``` php

    function test(){
        $MyProduct = MyProduct::get()->last();
        $MyProduct->MyDouble = 0;
        $MyProduct->MyPrice = 0;
        $MyProduct->write();
        if($MyProduct->MyDouble) {
            db::alteration_message("MyDouble with new DataObject: ERROR", "deleted");
        }
        else {
            db::alteration_message("MyDouble with new DataObject: OK", "created");
        }
        $MyProduct = MyProduct::get()->last();
        if($MyProduct->MyDouble) {
            db::alteration_message("MyDouble with saved DataObject: ERROR", "deleted");
        }
        else {
            db::alteration_message("MyDouble with saved DataObject: OK", "created");
        }
        $MyProduct->MyDouble = 0;
        $MyProduct->MyPrice = 0;
        $MyProduct->write();
        if($MyProduct->MyPrice) {
            db::alteration_message("MyCurrency with new DataObject: ERROR", "deleted");
        }
        else {
            db::alteration_message("MyCurrency with new DataObject: OK", "created");
        }
        $MyProduct = MyProduct::get()->last();
        if($MyProduct->MyPrice) {
            db::alteration_message("MyCurrency with saved DataObject: ERROR", "deleted");
        }
        else {
            db::alteration_message("MyCurrency with saved DataObject: OK", "created");
        }
    }

```

This returns the following 

![image](https://cloud.githubusercontent.com/assets/167154/4180039/b54b41fa-36e2-11e4-8d4f-3774c1b3f644.png)

kinglozzer:
I’ve encountered this before in templates too:

```html
<% if $Price %>
    {$Price}
<% end_if %>
```

Will return `0.00`

tractorcow:
We should override `exists` method in `DBCurrency`, so that at least in templates the bool-ness of the value can be corrected.

jonom:
[...]

PR #8590, opened 2018-11-09
Ensure that decimal DB values are fetched as floats

sminnee:
It turns out that the work from #8448 was incomplete in to important ways:

 * Decimal values (rather than Float) values are still being treated as strings
 * SQLite still had a number of bugs relating to this functionality (wasn't part of our test suite at the time of testing)

In order to fix this, most Query subclasses are going to end up with some value processing a la the code currently reserved for MySQL. Since I've added the tests for these issues to this branch I'll continue the work here.

---

Original ticket:

Validates that https://github.com/silverstripe/silverstripe-framework/issues/3473 has been fixed

The bug was fixed in #8448

I'm not sure how important it is to merge this test, but the more the merrier, I guess? ;-)

### DIFF
--- a/src/ORM/Connect/MySQLQuery.php
+++ b/src/ORM/Connect/MySQLQuery.php
@@ -4,6 +4,7 @@ namespace SilverStripe\ORM\Connect;
 
 /**
  * A result-set from a MySQL database (using MySQLiConnector)
+ * Note that this class is only used for the results of non-prepared statements
  */
 class MySQLQuery extends Query
 {
@@ -18,6 +19,11 @@ class MySQLQuery extends Query
     protected $handle;
 
     /**
+     * Metadata about the columns of this query
+     */
+    protected $columns;
+
+    /**
      * Hook the result-set given into a Query class, suitable for use by SilverStripe.
      *
      * @param MySQLiConnector $database The database object that created this query.
@@ -27,6 +33,9 @@ class MySQLQuery extends Query
     public function __construct($database, $handle)
     {
         $this->handle = $handle;
+        if (is_object($this->handle)) {
+            $this->columns = $this->handle->fetch_fields();
+        }
     }
 
     public function __destruct()
@@ -40,7 +49,7 @@ class MySQLQuery extends Query
     {
         if (is_object($this->handle)) {
             $this->handle->data_seek($row);
-            return $this->handle->fetch_assoc();
+            return $this->nextRecord();
         }
         return null;
     }
@@ -55,7 +64,19 @@ class MySQLQuery extends Query
 
     public function nextRecord()
     {
-        if (is_object($this->handle) && ($data = $this->handle->fetch_assoc())) {
+        $floatTypes = [MYSQLI_TYPE_FLOAT, MYSQLI_TYPE_DOUBLE, MYSQLI_TYPE_DECIMAL, MYSQLI_TYPE_NEWDECIMAL];
+
+        if (is_object($this->handle) && ($row = $this->handle->fetch_array(MYSQLI_NUM))) {
+            $data = [];
+            foreach ($row as $i => $value) {
+                if (!isset($this->columns[$i])) {
+                    throw new DatabaseException("Can't get metadata for column $i");
+                }
+                if (in_array($this->columns[$i]->type, $floatTypes)) {
+                    $value = (float)$value;
+                }
+                $data[$this->columns[$i]->name] = $value;
+            }
             return $data;
         } else {
             return false;
--- a/src/ORM/Connect/MySQLStatement.php
+++ b/src/ORM/Connect/MySQLStatement.php
@@ -6,7 +6,7 @@ use mysqli_result;
 use mysqli_stmt;
 
 /**
- * Provides a record-view for mysqli statements
+ * Provides a record-view for mysqli prepared statements
  *
  * By default streams unbuffered data, but seek(), rewind(), or numRecords() will force the statement to
  * buffer itself and sacrifice any potential performance benefit.
@@ -43,6 +43,13 @@ class MySQLStatement extends Query
     protected $columns = array();
 
     /**
+     * Map of column types, keyed by column name
+     *
+     * @var array
+     */
+    protected $types = array();
+
+    /**
      * List of bound variables in the current row
      *
      * @var array
@@ -59,6 +66,7 @@ class MySQLStatement extends Query
         // Bind each field
         while ($field = $this->metadata->fetch_field()) {
             $this->columns[] = $field->name;
+            $this->types[$field->name] = $field->type;
             // Note that while boundValues isn't initialised at this point,
             // later calls to $this->statement->fetch() Will populate
             // $this->boundValues later with the next result.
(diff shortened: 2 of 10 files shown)


tests/php/ORM/DataObjectTest.php
```diff
@@ -1915,6 +1915,24 @@ public function testFieldTypes()
...
+        $obj->write();
+
+        $this->assertFalse((bool)$obj->MyInt);
+        $this->assertFalse((bool)$obj->MyDecimal);
+        $this->assertFalse((bool)$obj->MyCurrency);
+
+        $obj2 = DataObjectTest\Fixture::get()->byId($obj->ID);
+        $this->assertFalse((bool)$obj2->MyInt); // return false as hoped
+        $this->assertFalse((bool)$obj2->MyDecimal); // returns true, bug
+        $this->assertFalse((bool)$obj2->MyCurrency); // returned true, bug
```
ScopeyNZ: Tests are failing. Would you mind tossing in some comments on each of these `assertFalse` so it's more obvious what behaviour is failing?

sminnee:
The remaining issue here is that PDO Prepared statements that are cached and used a 2nd time don't retain their native_type metadata on PostgreSQL.

 * The quick-n-dirty fix would be to avoid caching prepared statements for PDO + PostgreSQL
 * The better fix would be to introduce a SilverStripe\ORM\Connect\PDOStatement class that caches both the column metadata and the PDOStatement object, and perhaps we put the typeCorrectedFetchAll() implementation there too. This would change the constructor signature of PDOQuery, but we'd already accepted that in my previous work in this area.

I raised #8598 out of repeated frustration with PDO.

sminnee:
> The better fix would be to introduce a SilverStripe\ORM\Connect\PDOStatement class that caches both the column metadata and the PDOStatement object, and perhaps we put the typeCorrectedFetchAll() implementation there too. This would change the constructor signature of PDOQuery, but we'd already accepted that in my previous work in this area.

Any :+1: / :-1: feedback on this approach

sminnee:
I'm attaching this to the 4.4 milestone as it's kind of the 2nd half of some work already done post-4.3.

sminnee:
OK this is ready for review now :-)

sminnee:
Have requested reviews from a few probably-relevant people. If 1 or 2 people give this a 👍I'd say it's good to merge.

src/ORM/Connect/MySQLQuery.php
```diff
@@ -55,7 +64,17 @@ public function numRecords()
...
     public function nextRecord()
     {
-        if (is_object($this->handle) && ($data = $this->handle->fetch_assoc())) {
+        if (is_object($this->handle) && ($row = $this->handle->fetch_array(MYSQLI_NUM))) {
+            $data = [];
+            foreach ($row as $i => $value) {
+                if (!isset($this->columns[$i])) {
+                    throw new DatabaseException("Can't get metadata for column $i");
+                }
+                if ($this->columns[$i]->type == 246) {
```
kinglozzer:
```suggestion
                $floatTypes = [MYSQLI_TYPE_FLOAT, MYSQLI_TYPE_DOUBLE, MYSQLI_TYPE_DECIMAL, MYSQLI_TYPE_NEWDECIMAL];
                if (in_array($this->columns[$i]->type, $floatTypes)) {
```

src/ORM/Connect/MySQLStatement.php
```diff
@@ -116,6 +124,9 @@ public function nextRecord()
         // Dereferenced row
         $row = array();
         foreach ($this->boundValues as $key => $value) {
+            if ($this->types[$key] === 246) {
```
kinglozzer:
```suggestion
            $floatTypes = [MYSQLI_TYPE_FLOAT, MYSQLI_TYPE_DOUBLE, MYSQLI_TYPE_DECIMAL, MYSQLI_TYPE_NEWDECIMAL];
            if (in_array($this->types[$key], $floatTypes)) {
```

src/ORM/Connect/PDOConnector.php
```diff
@@ -113,11 +113,14 @@ public function getOrPrepareStatement($sql)
...
+        // Wrap in a PDOStatementHandle, to cache column metadata
+        $statementHandle = ($statement === false) ? false : new PDOStatementHandle($statement);
+
         // Only cache select statements
         if (preg_match('/^(\s*)select\b/i', $sql)) {
-            $this->cachedStatements[$sql] = $statement;
+            $this->cachedStatements[$sql] = $statementHandle;
         }
-        return $statement;
+        return $statementHandle;
```
kinglozzer: PHPDoc `@return` should be updated

src/ORM/Connect/PDOConnector.php
```diff
@@ -395,17 +405,20 @@ public function preparedQuery($sql, $parameters, $errorLevel = E_USER_ERROR)
     {
         $this->beforeQuery($sql);
 
-        // Prepare statement
+        // Fetch cached statement, or create it
         $statement = $this->getOrPrepareStatement($sql);
```
kinglozzer: Sorry to be a pain, can you rename to `$statementHandle`?

src/ORM/Connect/PDOConnector.php
```diff
@@ -395,17 +405,20 @@ public function preparedQuery($sql, $parameters, $errorLevel = E_USER_ERROR)
...
 
-        // Bind and invoke statement safely
-        if ($statement) {
-            $this->bindParameters($statement, $parameters);
+        // Error handling
+        if ($statement === false) {
+            $this->databaseError($this->getLastError(), $errorLevel, $sql, $this->parameterValues($parameters));
```
kinglozzer: `return null` and remove the `else {`?

src/ORM/Connect/PDOConnector.php
```diff
@@ -418,29 +431,22 @@ public function preparedQuery($sql, $parameters, $errorLevel = E_USER_ERROR)
      * @param array $parameters
      * @return PDOQuery
      */
-    protected function prepareResults($statement, $errorLevel, $sql, $parameters = array())
+    protected function prepareResults(PDOStatementHandle $statement, $errorLevel, $sql, $parameters = array())
```
kinglozzer: `@param` needs to be updated

sminnee:
OK your feedback has been addressed here https://github.com/sminnee/silverstripe-framework/commit/935f693937f94da22810bad5a779c724c131a7ba which I'll now squash back into the other commits ready for merge.

sminnee:
...fixed the phpcs failure that broke the previous build.

src/ORM/Connect/PDOConnector.php
sminnee: not a pain, that's what peer review is for ;-)

kinglozzer:
Looks good @sminnee, there’s just one review comment outstanding (swapping `=== 246` for a list of consts) that I think was missed?

sminnee:
OK reviewing and rebasing this now.

src/ORM/Connect/MySQLStatement.php
sminnee: Squahsed this into https://github.com/silverstripe/silverstripe-framework/pull/8590/commits/45e1fcaf309bc3d3a6ae9da9a716cf4ff935b1ad

sminnee:
@kinglozzer @ScopeyNZ waddaya reckon?

kinglozzer:
I reckon this is good to go 👍

PR #8591, opened 2018-11-10
NEW: Allow for user-created objects to have values passed in the constructor

sminnee:
This change extends isSingleton into a tri-state creationType, which
recognises a new option, CREATE_HYDRATED. This is used to create records
from database content, which bypasses both setters and defaults.

Which that in place, the default, non-hydrated call will use setters
via DataObject::update(), and is therefore allowable for user-code
instantiating new objects.

Some quirks of existing behaviour are retained, notably that for new
objects data created by populateDefaults() will register as a change in
$this->original vs $this->record, but not $this->changed.

# Parent issue
* #6460

# To do

 * [x] Remove stray var_dump
 * [x] Fix type-hint on creationType: boolean|int
 * [x] Add another test case for all 3 creation modes

### DIFF
--- a/src/Core/Injector/Injector.php
+++ b/src/Core/Injector/Injector.php
@@ -587,7 +587,7 @@ class Injector implements ContainerInterface
         if ((!$type || $type !== self::PROTOTYPE)
             && empty($constructorParams)
             && is_subclass_of($class, DataObject::class)) {
-            $constructorParams = [null, true];
+            $constructorParams = [null, DataObject::CREATE_SINGLETON];
         }
 
         $factory = isset($spec['factory']) ? $this->get($spec['factory']) : $this->getObjectCreator();
--- a/src/ORM/DataList.php
+++ b/src/ORM/DataList.php
@@ -819,6 +819,7 @@ class DataList extends ViewableData implements SS_List, Filterable, Sortable, Li
 
     /**
      * Create a DataObject from the given SQL row
+     * If called without $row['ID'] set, then a new object will be created rather than rehydrated.
      *
      * @param array $row
      * @return DataObject
@@ -841,7 +842,9 @@ class DataList extends ViewableData implements SS_List, Filterable, Sortable, Li
             $class = $row['RecordClassName'];
         }
 
-        $item = Injector::inst()->create($class, $row, false, $this->getQueryParams());
+        $creationType = empty($row['ID']) ? DataObject::CREATE_OBJECT : DataObject::CREATE_HYDRATED;
+
+        $item = Injector::inst()->create($class, $row, $creationType, $this->getQueryParams());
 
         return $item;
     }
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -966,13 +966,18 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 
     /**
      * Convert this object to a map.
+     * Note that it has the following quirks:
+     *  - custom getters, including those that adjust the result of database fields, won't be executed
+     *  - NULL values won't be returned.
      *
      * @return array The data as a map.
      */
     public function toMap()
     {
         $this->loadLazyFields();
-        return $this->record;
+        return array_filter($this->record, function ($val) {
+            return $val !== null;
+        });
     }
 
     /**
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -184,6 +184,25 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
     const CHANGE_VALUE = 2;
 
     /**
+     * Value for 2nd argument to constructor, indicating that a new record is being created
+     * Setters will be called on fields passed, and defaults will be populated
+     */
+    const CREATE_OBJECT = 0;
+
+    /**
+     * Value for 2nd argument to constructor, indicating that a record is a singleton representing the whole type,
+     * e.g. to call requireTable() in dev/build
+     * Defaults will not be populated and data passed will be ignored
+     */
+    const CREATE_SINGLETON = 1;
+
+    /**
+     * Value for 2nd argument to constructor, indicating that a record is being hydrated from the database
+     * Neither setters and nor default population will be called
+     */
+    const CREATE_HYDRATED = 2;
+
+    /**
      * An array indexed by fieldname, true if the field has been changed.
      * Use {@link getChangedFields()} and {@link isChanged()} to inspect
      * the changed state.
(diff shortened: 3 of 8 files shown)


sminnee:
> Some quirks of existing behaviour are retained, notably that for new
> objects data created by populateDefaults() will register as a change in
> $this->original vs $this->record, but not $this->changed.

I might actually revisit this — it looks like both original _and_ changed should reflect the content currently in the database, which in the case of new records should have none of the content derived from populateDefaults(). However, I'll get the tests passing without touching this before diving into that, and I'll see if there are tests that rely on this quirky behaviour, and if so, why...

src/Core/Injector/Injector.php
```diff
@@ -587,7 +587,7 @@ protected function instantiate($spec, $id = null, $type = null)
         if ((!$type || $type !== self::PROTOTYPE)
             && empty($constructorParams)
             && is_subclass_of($class, DataObject::class)) {
-            $constructorParams = array(null, true);
+            $constructorParams = array(null, DataObject::CREATE_SINGLETON);
```
tractorcow: I have a feeling this should really be a part of a DataObjectFactory. We'll have other places where similar concerns require logic in a similar fashion. This ties injector to ORM more than it should IMO.

Maybe allowing us to specify a factory for a class hierarchy would be good to look into?

tractorcow:
```yaml
  SilverStripe\ORM\DataObject:
    factory: SilverStripe\ORM\DataObjectFactory
    inherits: true
```

php: 

```php
public function getFactory($spec) {
    // Get factory for this class
    if (isset($spec['factory'])) {
        return $this->get($spec['factory']);
    }
    // Look for any factory on parent class with `inherits: true`
    $next = $spec['class'];
    while ($next = get_parent_class($next)) {
        $nextConfig = $this->configLocator->locateConfigFor($next);
        if (isset($nextConfig['factory']) && !empty($nextConfig['inherits'])) {
            return $this->get($nextConfig['factory']));
        }
     }
    // Default constructor
    return $this->getObjectCreator();
}
```

sminnee: I don't necessarily disagree with any of this, but it's unrelated to this PR. This PR just removed the use of a now-deprecated value in the constructor.

Perhaps you want to pick this up as a separate issue.

Briefly, the reasons I haven't dug into this are:
 * Concerns about adding complexity for one weird edge-case
 * Performance
 * Whether the "best fix" would be an API breakage and as such best left 'til SS5

None of those are insurmountable, but they've meant that I've left this one alone myself.

src/ORM/DataList.php
```diff
@@ -817,7 +818,9 @@ public function createDataObject($row)
             $class = $row['RecordClassName'];
         }
 
-        $item = Injector::inst()->create($class, $row, false, $this->getQueryParams());
+        $creationType = empty($row['ID']) ? DataObject::CREATE_OBJECT : DataObject::CREATE_HYDRATED;
+
+        $item = Injector::inst()->create($class, $row, $creationType, $this->getQueryParams());
```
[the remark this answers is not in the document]
sminnee: > I'm worried about the tri-state for dataobjects, bool state for all other objects and coupling of injector to dataobject mostly.

AFAIK DataObject is the only class that makes use of this 2nd argument, so what we're talking about here is turning a bi-state argument tri-state argument.

Hyrdation, Creation, and Singletons are different cases. In the status-quo code, Creation and Hydration are coupled with the assumption that the Creation case will have an empty record. Frankly, Hydration and Singleton have more in common so you could argue "maybe hydration should set isSingleton = true". But I think that this will risk more messy edge-cases and also risk semver breakages. So I think a tri-state field is better.

We could arguably use the presence of an ID value to distinguish hydration (ID is present) from creation (ID isn't present), but that will cause bugs in two cases:
 * We attempt to hydrate a record without an ID
 * We attempt to create a record with a predefined ID

The former case seems unlikely unless we change our ORM, but the latter could conceivably occur in edge-cases such as copying content from stage to draft. I can't guarantee that it will introduce bugs, but it seems a risky proposition — I think a tri-state is cleaner.

*Why is this conditional needed?*
This condition was added to make ManyManyThroughList::add() work, which creates a _new_ object by calling `$hasManyList->createDataObject()` on line 217.

So, if we wanted to reduce the use of the conditional here, we could potentially deprecate `createDataObject()` in favour of `hydrateDataObject()` and `createNewDataObject()`. I'd probably still leave this code as-is, but at least the messiness would be contain in a deprecated API.

sminnee:
Damian gave some feedback here that I don't really think should be incorporated into this PR. I feel it pushes towards more like "possible bigger refactor in SS5", when here I have an SS4-appropriate fix. 

There's one suggestion I had for partially addressing his concern:

> So, if we wanted to reduce the use of the conditional here, we could potentially deprecate createDataObject() in favour of hydrateDataObject() and createNewDataObject(). I'd probably still leave this code as-is, but at least the messiness would be contain in a deprecated API.

Should I do this? If I do that, shall we merge it?

Cc @dhensby @chillu @kinglozzer @tractorcow for good measure.

src/ORM/DataObject.php
```diff
@@ -316,88 +335,107 @@ public static function getSchema()
...
+                        E_USER_WARNING
+                    );
+                    $record = [];
                 }
-            }
-        }
 
-        $this->original = $this->record;
+                // Default columns
+                $this->record['ID'] = empty($record['ID']) ? 0 : $record['ID'];
```
chillu: This used to cast to `(int)` - did you remove that on purpose?

tests/php/ORM/DataObjectTest.php
```diff
@@ -188,8 +188,8 @@ public function testConstructAcceptsValues()
         $this->assertEquals('John', $player->FirstName);
         $this->assertEquals('Doe', $player->Surname);
 
-        // IDs should be stored as integers, not strings
-        $player = new DataObjectTest\Player(array('ID' => '5'));
+        // Note that automatic conversion of IDs to integer no longer happens as the DB layer does that for us now
+        $player = new DataObjectTest\Player(array('ID' => 5));
```
chillu: Feels a bit light on unit tests - we've gone from accepting two states to three states. Surely that should require another test case?

src/ORM/DataList.php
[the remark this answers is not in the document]
chillu: I agree with Sam, in that it's safer to make this an explicit decision of the API caller (tri-state), rather than keep the current signature and infer from the various constructor argument combinatoins (bi-state)

sminnee:
Cool I’ll address feedback on hackday.

src/ORM/DataObject.php
```diff
@@ -316,88 +335,107 @@ public static function getSchema()
     /**
      * Construct a new DataObject.
      *
-     * @param array|null $record Used internally for rehydrating an object from database content.
-     *                           Bypasses setters on this class, and hence should not be used
-     *                           for populating data on new records.
-     * @param boolean $isSingleton This this to true if this is a singleton() object, a stub for calling methods.
-     *                             Singletons don't have their defaults set.
+     * @param array $record Initial record content, or rehydrated record content, depending on $creationType
+     * @param boolean $creationType Set to DataObject::CREATE_OBJECT, DataObject::CREATE_HYDRATED, or DataObject::CREATE_SINGLETON. Used by SilverStripe internals as best left as the default by regular users.
```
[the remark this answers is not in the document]
sminnee: Can be `boolean|int` for backwards compat, will fix

src/ORM/DataObject.php
sminnee: I think my rationale was No longer needed thanks to my strict-typing work

sminnee:
Ok I've squashed the 3 agreed bits of feedback into the 
  * [Remove stray var_dump](https://github.com/silverstripe/silverstripe-framework/pull/8591/commits/25499087c28961f0cbef4ee2b01c780cc0296204#diff-5bda7c74e3842ae1fa276619f19d7babR367)
  * [Fix type-hint on creationType: boolean|int](https://github.com/silverstripe/silverstripe-framework/pull/8591/commits/25499087c28961f0cbef4ee2b01c780cc0296204#diff-5bda7c74e3842ae1fa276619f19d7babR340)
  * [Add another test case for all 3 creation modes](https://github.com/silverstripe/silverstripe-framework/pull/8591/commits/25499087c28961f0cbef4ee2b01c780cc0296204#diff-d018d15e90dda91af06810b9870e8145R2417)

Should be good to merge now?

sminnee:
OK got a few weird failures here I think when I rebased... a couple of fixes dropping in now.

sminnee:
A side-effect of this change is that NULL column values are now included in toMap(), whereas previously they were excluded. This is because the code previously had an "if <x> !== null" check when populating $this->record.

I can't tell where this is a bugfix or a breaking change, although it has caused some issues with calculated field comparison when assertListEquals is passed the result of toMap() for a DataObject with calculated fields.

In particular:

 * MigrateSiteTreeLinkingTaskTest passes the result of toMap to assertListEquals, to test SiteTree records.
 * MenuTitle shows up as NULL in the toMap() result, but on the object a `getMenuTitle()` method returns the value of Title.
 * This means the record doesn't match and the test fails.

So, what's our conclusion? Should toMap() exclude NULL fields (the current behaviour on `4`) or should MigrateSiteTreeLinkingTaskTest not rely so much on toMap() results (it's a bit of a nasty coding pattern at the mo)

@silverstripe/core-team is the restoration of NULL values to toMap() results a bugfix or an API change?

sminnee:
The CMS PR https://github.com/silverstripe/silverstripe-cms/pull/2441 will fix the failing test on this, but it would be good to confirm whether what I've fixed is, in fact, an API change before merging this.

robbieaverill:
What was the behaviour in SilverStripe 3? I think having null values present would be making the API honest about its intention, but I also recognise that it might break someone's code somewhere

sminnee:
> So, what's our conclusion? Should toMap() exclude NULL fields (the current behaviour on 4) or should MigrateSiteTreeLinkingTaskTest not rely so much on toMap() results (it's a bit of a nasty coding pattern at the mo)

I've looked at `3`. The behaviour of `4` matches `3`, so there's no regression there. This is just one of those tightening up activities that is likely easier to achieve because of the type-integrity work I added in 4.4.

robbieaverill:
Would it be silly to suggest we change it but also provide a configuration option to revert back if people find it breaks something?

sminnee:
> Would it be silly to suggest we change it but also provide a configuration option to revert back if people find it breaks something?

I don't think this will help much. I think the hard thing will be identifying that this is what's happening; once it's identified it's pretty easy to fix. And config flags for such deep behaviour is likely to make the system less reliable overall.

If we do change it, we should list it in the upgrade notes though.

I might break this out to a separate bug card.

maxime-rainville:
If this breaks someone's code, the proper fix would probably be to add a null check in their loop, not tweak some random configuration flag that they will never have heard off

If it's actually breaking our own existing test, it sounds like a sensible expectation that this will be breaking someone's code somewhere. So I would err on the side of not returning nulls.

sminnee:
In my view it broke a test that abused assertListEquals by pushing the result of toMap into it, and https://github.com/silverstripe/silverstripe-cms/pull/2441/files shows what was the right way to do it. But, yeah, a bit of a value-judgement there.

sminnee:
Broke out a separate bug card here #9021

sminnee:
Context for the test that broke:

> The specific issue came with a field whose database value is a null but whose calculated value isn't. This happened in the case of MenuTitle, which defaults to return the value of Title if the database MenuTitle is null. assertListEquals was being passed the result of a toMap() in one badly written test, and because the null MenuTitle wasn't included in toMap(), it side-stepped this issue. So fixing this bug broke a test, but it was a very brittle situation to begin with.

sminnee:
Following up form my comment on #9021, I'm going to amend this PR to do the following:

 * Fix toMap() so that it doesn't return NULL values
 * Update the docblock of toMap() so that it says "neither NULL values nor results of custom getters will be included"
 * Add a test to highlight this odd behaviour.

Cheddam:
@sminnee Do you have time to work on those amendments, or shall we close this out for the time being?

sminnee:
Thanks for the nudge @Cheddam I've addressed the final work on this now

src/ORM/DataObject.php
```diff
@@ -318,88 +337,107 @@ public static function getSchema()
...
-            ];
+        // Legacy $record default
+        if ($record === null) {
+            $record = [];
         }
 
-        if ($record instanceof stdClass) {
-            $record = (array)$record;
+        // Legacy $isSingleton boolean
+        if (is_bool($creationType)) {
```
maxime-rainville, line 354: Maybe I'm overthinking this, but if some legacy code calls this method with a non-boolean truthy/falsy value weird things will happen. I'd be tempted to have some strong type guards here, to handle cases where  a value different from `false`, `true`, `CREATE_SINGLETON`, `CREATE_OBJECT` or `CREATE_HYDRATED` is passed.

src/ORM/DataObject.php
```diff
@@ -318,88 +337,107 @@ public static function getSchema()
...
+                        "DataObject::__construct passed $passed.  It's supposed to be passed an array,"
+                        . " taken straight from the database.  Perhaps you should use DataList::create()->First(); instead?",
+                        E_USER_WARNING
+                    );
+                    $record = [];
+                }
+
+                // Default columns
+                $this->record['ID'] = empty($record['ID']) ? 0 : $record['ID'];
+                $this->record['ClassName'] = $this->record['RecordClassName'] = static::class;
```
maxime-rainville, line 415:
```suggestion
                $this->record['ClassName'] = static::class;
                $this->record['RecordClassName'] = static::class;
```

src/ORM/DataObject.php
```diff
@@ -318,88 +337,107 @@ public static function getSchema()
...
+                }
+                break;
 
-        // prevent populateDefaults() and setField() from marking overwritten defaults as changed
-        $this->changed = [];
-        $this->changeForced = false;
+            case self::CREATE_SINGLETON:
+                // No setting happens for a singleton
+                $this->record['ID'] = 0;
+                $this->record['ClassName'] = $this->record['RecordClassName'] = static::class;
```
maxime-rainville, line 435:
```suggestion
                $this->record['ClassName'] = static::class;
                $this->record['RecordClassName'] = static::class;
```

src/ORM/DataObject.php
sminnee: It's an edge-case but easily addressed 👍

sminnee: I've opted for
 * Check !is_int rather than is_bool
 * Throw a warning if it's neither int nor bool
 * Throw a fatal error if it's int but not a proper value

If someone has passed $isSingle = 1042 their code will throw a fatal, but that's technically violating the original API and it's a rare that this API is directly used regardless.

sminnee:
okay this is ready for another review thanks

Issue #8332, opened 2018-08-17
Add file folder DatabaseException with PostgreSQL


jacobaarnold:
## Affected Version

phpunit/phpunit            5.7.27
silverstripe-themes/simple 3.2.0 
silverstripe/postgresql    2.0.2 
silverstripe/recipe-cms    4.2.1 
silverstripe/recipe-plugin 1.3.0 

## Description

When running a clean base install of Silverstripe 4.2.1 against PostgreSQL 10.0 using using pg_connect, clicking the "Add Folder" button under Files in the CMS admin results in an error:

[...]

PR #8592, opened 2018-11-12
FIX TreeMultiselectField passes value 'unchanged' as null to ORM

dnsl48:
Fix for https://github.com/silverstripe/silverstripe-framework/issues/8332

### DIFF
--- a/tests/php/Forms/TreeMultiselectFieldTest.php
+++ b/tests/php/Forms/TreeMultiselectFieldTest.php
@@ -15,6 +15,41 @@ class TreeMultiselectFieldTest extends SapphireTest
     protected $formId = 'TheFormID';
     protected $fieldName = 'TestTree';
 
+    /**
+     * Mock object of a generic form
+     *
+     * @var Form
+     */
+    protected $form;
+
+    /**
+     * Instance of the TreeMultiselectField
+     *
+     * @var TreeMultiselectField
+     */
+    protected $field;
+
+    /**
+     * The File objects of folders loaded from the fixture
+     *
+     * @var File[]
+     */
+    protected $folders;
+
+    /**
+     * The array of folder ids
+     *
+     * @var int[]
+     */
+    protected $folderIds;
+
+    /**
+     * Concatenated folder ids for use as a value for the field
+     *
+     * @var string
+     */
+    protected $fieldValue;
+
     protected function setUp()
     {
         parent::setUp();


robbieaverill:
There's also a missing space after `===` (not covered by phpcs) and would be good to add a unit test for it if you have time =)

dnsl48:
there's even more about `@deprecate`. We're not deprecating the method itself, but rather will have to remove this particular implementation of it in the next major. I'm gonna change it to internal.

Regarding tests - fair enough, I'll add some coverage for the class.

dnsl48:
@maxime-rainville 
> Silly questions, but what is actually preventing us from fixing the core issue that we're assigning unchanged to the value? Is it a semver thing that we're afraid that other people might be relying on the value being unchanged?

Yes, exactly. The string `'unchanged'` is the default empty value of the field. That contract is propagating to some other libraries as well in both back-end and front-end sides. At this point we don't have guarantees that no 3rd party libraries rely on this particular behaviour and changing that default value would be a breaking change in the framework API.

> If there's a good reason for keeping the current behavior in 4, I think we should do a PR to fix the issue in 5 right away, because that's the kind of thing we'll forget about later on.

I believe we should keep the current behaviour for 4 in sake of backward compatibility.
Regarding the fix for the next major (5.0), yes that's the current plan. However, due to avoiding conflicts when merging 4 into master I want this PR to go into the 4 first, then we merge it into master and then I do another PR that overwrites this change and removes `'unchanged'` string, leaving `null` value in its place.

maxime-rainville:
Sounds sensible. We should be able to merge once we got some unit test.

tests/php/Forms/TreeMultiselectFieldTest.php
```diff
@@ -10,22 +10,57 @@ class TreeMultiselectFieldTest extends SapphireTest
...
+        $field = new TreeMultiselectField('TestTree', 'Test tree', File::class);
+
+        $asdf = $this->objFromFixture(File::class, 'asdf');
+        $subfolderfile1 = $this->objFromFixture(File::class, 'subfolderfile1');
+
+        $schemaStateDefaults = $field->getSchemaStateDefaults();
+        $this->assertArraySubset(['id' => 'TestTree', 'name' => 'TestTree', 'value' => 'unchanged'], $schemaStateDefaults, $strict = true);
+
+        $items = $field->getItems();
+        $this->assertCount(0, $items, $message = 'there must be no items selected');
```
robbieaverill: `$strict` and `$message` appear to be unused

tests/php/Forms/TreeMultiselectFieldTest.php
```diff
@@ -10,22 +10,57 @@ class TreeMultiselectFieldTest extends SapphireTest
...
+
+        $html = $field->performReadonlyTransformation()->Field();
+
+        $this->assertEquals(
+            <<<"HTML"
+<span id="TestTree_ReadonlyValue" class="readonly">
+	<i>('none')</i>
+</span><input type="hidden" name="TestTree" class="hidden" id="TestTree" />
```
robbieaverill: Sorry, one last thing - this is weak since it's prone to break if you override the templates that this form field uses to render with. Examples of this are in the CWP recipes, where we modify some of the core form field templates to increase their accessibility compliance. I'd suggest doing partial assertions instead, e.g. `assertContains("('none')", $html)`, `assertNotContains('value="', $html)` etc

tests/php/Forms/TreeMultiselectFieldTest.php
```diff
@@ -10,22 +10,57 @@ class TreeMultiselectFieldTest extends SapphireTest
...
         $subfolderfile1 = $this->objFromFixture(File::class, 'subfolderfile1');
+
         $field->setValue(implode(',', [$asdf->ID, $subfolderfile1->ID]));
 
-        $readonlyField = $field->performReadonlyTransformation();
+        $schemaStateDefaults = $field->getSchemaStateDefaults();
+        $this->assertArraySubset(['id' => 'TestTree', 'name' => 'TestTree', 'value' => [$asdf->ID, $subfolderfile1->ID]], $schemaStateDefaults, $strict = true);
+
+        $items = $field->getItems();
+        $this->assertCount(2, $items, $message = 'there must be exactly 2 items selected');
```
robbieaverill: `$strict` and `$message` appear to be unused

tests/php/Forms/TreeMultiselectFieldTest.php
dnsl48: oh, yeah, didn't notice the leftovers of autocompletion...

tests/php/Forms/TreeMultiselectFieldTest.php
dnsl48: well, since it's supposed to be a unit test, I'd rather not to test the rendering at all and only test the class own and inherited methods. I'll try to rewrite it in a more generic way, without rendering and templates involved

P.S. even if we test the partials of the resulting string, there are no guarantees it's gonna always contain all the things we are checking for.

tests/php/Forms/TreeMultiselectFieldTest.php
```diff
@@ -10,22 +10,57 @@ class TreeMultiselectFieldTest extends SapphireTest
...
+        $schemaStateDefaults = $field->getSchemaStateDefaults();
+        $this->assertArraySubset(['id' => 'TestTree', 'name' => 'TestTree', 'value' => 'unchanged'], $schemaStateDefaults, $strict = true);
+
+        $items = $field->getItems();
+        $this->assertCount(0, $items, $message = 'there must be no items selected');
+
+        $html = $field->performReadonlyTransformation()->Field();
+
+        $this->assertEquals(
+            <<<"HTML"
```
ScopeyNZ: Minor: The double-quotes here are unnecessary (same as no quotes) - Perhaps you meant for a nowdoc, but that's specifically single-quotes.

dnsl48: well, not really meant anything, just copy pasted this bit from the original code :)
I'm rewriting this test to be more unit, so will remove the html block.

dnsl48:
@robbieaverill I've managed to rewrite the tests. Now we're not checking the rendering anymore, but rather what data the object passes to the template engine.

tests/php/Forms/TreeMultiselectFieldTest.php
```diff
@@ -4,28 +4,260 @@
...
     protected static $fixture_file = 'TreeDropdownFieldTest.yml';
 
-    public function testReadonly()
+    protected $formId = 'TheFormID';
+    protected $fieldName = 'TestTree';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setUp()
```
robbieaverill: Two very minor things - `setUp()` is usually protected in PHPUnit 5, and the `@inheritDoc` above doesn't add any value unless you augment it. PHPDoc will inherit by default

tests/php/Forms/TreeMultiselectFieldTest.php
```diff
@@ -4,28 +4,260 @@
...
+    {
+        parent::setUp();
+
+        $this->form = $this->buildFormMock();
+        $this->field = $this->buildField($this->form);
+        $this->folders = $this->loadFolders();
+
+        $this->folderIds = array_map(
+            static function ($f) {
```
robbieaverill: I've never seen this before - is the static part needed? Just curious, not a request to change

dnsl48: That works in the same way as with static variables.
Since we don't need to ditch and recreate this closure on every setup invocation, we can just make it static and reuse the static instance across all tests.

tests/php/Forms/TreeMultiselectFieldTest.php
dnsl48: ok, didn't know that, thank you :)

maxime-rainville: Me learning new things.

dnsl48:
Ok, everything's been fixed. Good to be merged.

PR #8609, opened 2018-11-15
API Add a GenericList interface to encapsulate a list that is fitlerable, sortable, limitable

maxime-rainville:
This is admittedly a somewhat academic point, but there's quite a few places where we say we expect/return an `SS_List` when what we really mean is an `SS_List` that is `Filterable`, `Sortable` and/or `Limitable`.

I think we should be explicit and provide an interface that encapsulate all of those other interfaces.

Aside from the additional clarity it will also help with code completion.

I'm not overly fuss about that `GenericList` name. If someone can think of a more descriptive name, go for it.

### DIFF
--- a/src/ORM/ArrayList.php
+++ b/src/ORM/ArrayList.php
@@ -24,7 +24,7 @@ use SilverStripe\Dev\Deprecation;
  *   - filter
  *   - exclude
  */
-class ArrayList extends ViewableData implements SS_List, Filterable, Sortable, Limitable
+class ArrayList extends ViewableData implements GenericList
 {
 
     /**
--- a/src/ORM/DataList.php
+++ b/src/ORM/DataList.php
@@ -32,7 +32,7 @@ use LogicException;
  *
  * Subclasses of DataList may add other methods that have the same effect.
  */
-class DataList extends ViewableData implements SS_List, Filterable, Sortable, Limitable
+class DataList extends ViewableData implements GenericList
 {
 
     /**
--- a/src/ORM/GenericList.php
+++ b/src/ORM/GenericList.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace SilverStripe\ORM;
+
+/**
+ * A Generic implementation of `SS_List` with all related interface to enable filtering, sorting and limiting.
+ *
+ * @see SS_List, Sortable, Limitable, Filterable
+ */
+interface GenericList extends SS_List, Filterable, Sortable, Limitable
+{
+}
--- a/src/ORM/Relation.php
+++ b/src/ORM/Relation.php
@@ -13,7 +13,7 @@ use SilverStripe\ORM\FieldType\DBField;
  * @method Relation forForeignID($id)
  * @method string dataClass()
  */
-interface Relation extends SS_List, Filterable, Sortable, Limitable
+interface Relation extends GenericList
 {
 
     /**


robbieaverill:
It's been argued previously that adding interfaces can be a breaking change, we'd need to be careful with this

ping @silverstripe/core-team

unclecheese:
@robbieaverill Can you articulate why this would be a breaking change?

robbieaverill:
I'm voicing a comment I remember from @dhensby, perhaps he can help :)

robbieaverill:
I guess if the interfaces don't have any new methods in them then it should be fine

sminnee:
It's more "CompleteList" or "FullFeaturedList" than "GenericList", but I agree that having a "union interface" probably makes sense.

I'd see this as a minor change but I struggle to see what it would break. Although changing a method that currently accepts an SS_List to accept a full-featured list _would_ be a breaking change.

unclecheese:
+1. I only see this as a breaking change if we start consuming it in methods.

tractorcow:
It's only breaking if the new interface doesn't extend the old one. This does so it should be fine.

I dislike the name "Generic" also, because it doesn't convey any meaning. The `SS_List` has the same problem. It tries to be "the" list type for silverstripe. :)

We had a similar problem where we ended up adding `Relation` interface for relations, since not all relations could extend `RelationList` class.

We also have some ambiguity around item types used for these lists. DataList of course only works with DataObjects, but ArrayList is used for all viewabledata types, and even simple types.

I would prefer the name `ObjectList` and clarify that this list implies it contains ViewableData items only, and thus supports other list types that would work on objects only (e.g. Filterable, which requires objects).

unclecheese:
`ManipulableList`?

Just thinking of `getManipulatedData()` and `DataManipulator` in `GridField`.

tractorcow:
I'd be forever spelling that wrong. :P

unclecheese:
Fortunately, all you need to know are the first three letters, and the IDE will take it from there. :)

robbieaverill:
I already read it three times - is it a real word? =P

maxime-rainville:
I'll throw it in the backend dev channel and see what the masses comes back with.

sminnee:
“Collection” means something In pre-existing patterns. It differs from “List” in that it doesn’t have an order, so you can’t iterate it. I think this is misleading - like how “DataObjectDecorator” wasn’t actually a decorator.

Calling it “CollectionList” is like calling something “CircleSquare”.

Can anyone elaborate on what problems this change actually solves? Do we have a lot of objects that explicitly require a list that has all 3 features?

Another option: FLSList - for filteable, limitable, sortable ;-)

maxime-rainville:
The specific examples that were bugging me were the GridField components and interfaces that proclaim they expect an SS_List object but then calls method specific to Limitable/Sortable/Filterable.

But in practice, pretty much every where we expect an `SS_List` we implicitly expect them to implement those 3 interfaces.

maxime-rainville:
As I mentioned in the original description, the point is somewhat academic ... I doubt there's many people out there creating their own list implementation from scratch.

The main problem is we don't have a way of saying: "this method needs a list that is sortable and filterable".

robbieaverill:
>The specific examples that were bugging me were the GridField components and interfaces that proclaim they expect an SS_List object but then calls method specific to Limitable/Sortable/Filterable.

Yeah that bugs me too

robbieaverill:
If it helps, PSR-5 is talking about introducing a union type for phpdocs, but that doesn't help if you can't type hint it in the PHP argument for the method

sminnee:
I think that the API-breaking fix to this would be to push the features of these 3 interfaces into SS_List and find a different way of doing feature-detection.

Related, "Collection" would more closely match the current SS_List and "List" the full-featured version.

The original reason for this pattern was to allow for the implementation of other List implementations (e.g. providing integration with 3rd party APIs) that didn't necessarily provide sort, filter, and limit options. In other words, a form of feature-detection.

However, this could be done with a method such as `$list->supports(SS_List::SORT_FEATURE) // boolean` instead. Furthermore, it's rare that this capability is actually used.

These are all API breakages. So, if we were to go down this path, it would change in SS5.

So perhaps you could call it SS5List? ;)

maxime-rainville:
I've raised a matching issue for SS5 https://github.com/silverstripe/silverstripe-framework/issues/8778.
I'll close this for now.

PR #8616, opened 2018-11-19
DOCS Updating HistoryViewer documentation to specify the rollback mutation instead of copyToStage

ScopeyNZ:
This PR updates the HistoryViewer docs to specify using a `rollback` mutation instead of the lower level `copyToStage` API. Rollback uses `rollbackRecursive` and is more likely to provide the expected functionality.

Relies on silverstripe/silverstripe-versioned#192


robbieaverill:
I guess we need to update the cms and elemental modules to use the new mutation for a SilverStripe ^4.3 only version?

PR #8619, opened 2018-11-19
Use new plus icon

bergice:


src/Forms/GridField/GridFieldConfig_RecordEditor.php
```diff
@@ -10,8 +10,10 @@ class GridFieldConfig_RecordEditor extends GridFieldConfig
     /**
      *
      * @param int $itemsPerPage - How many items per page should show up
+     * @param bool $showPagination Whether the `Previous` and `Next` buttons should display or not, leave as null to use default
+     * @param bool $showAdd Whether the `Add` button should display or not, leave as null to use default
      */
-    public function __construct($itemsPerPage = null)
+    public function __construct($itemsPerPage = null, $showPagination = null, $showAdd = null)
```
robbieaverill: I think it'd be better to make this configuration rather than new constructor arguments

src/Forms/GridField/GridFieldDetailForm.php
```diff
@@ -79,10 +88,14 @@ public function getURLHandlers($gridField)
      * controller who wants to display the getCMSFields
      *
      * @param string $name The name of the edit form to place into the pop-up form
+     * @param bool $showPagination Whether the `Previous` and `Next` buttons should display or not, leave as null to use default
+     * @param bool $showAdd Whether the `Add` button should display or not, leave as null to use default
      */
-    public function __construct($name = 'DetailForm')
+    public function __construct($name = null, $showPagination = null, $showAdd = null)
```
robbieaverill: Changing null to DetailForm would be a semver break, no?

src/Forms/GridField/GridFieldDetailForm.php
```diff
@@ -179,6 +196,68 @@ public function getName()
...
+    /**
+     * @return bool
+     */
+    public function getShowAdd()
+    {
+        if ($this->showAdd === null) {
+            return $this->getDefaultShowAdd();
+        }
+
+        return (boolean) $this->showAdd;
```
robbieaverill: Stylistic: prefer `(bool)` over `(boolean)`

src/Forms/GridField/GridFieldDetailForm.php
```diff
@@ -232,8 +311,8 @@ public function getItemRequestClass()
     {
         if ($this->itemRequestClass) {
             return $this->itemRequestClass;
-        } elseif (ClassInfo::exists(static::class . "_ItemRequest")) {
-            return static::class . "_ItemRequest";
+        } elseif (ClassInfo::exists(static::class . '_ItemRequest')) {
+            return static::class . '_ItemRequest';
         } else {
```
robbieaverill: Minor: the `else` here is irrelevant, just the return inside it would be fine

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -290,6 +294,45 @@ protected function getFormActions()
                     ->setUseButtonTag(true)
                     ->addExtraClass('btn-outline-danger btn-hide-outline font-icon-trash-bin action--delete'));
             }
+
+            $gridStateStr = $this->getRequest()->requestVar('gridState');
```
robbieaverill: Could you use a different variable name, or expand `Str` to whatever it's longer form should be?

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -290,6 +294,45 @@ protected function getFormActions()
...
+            }
+
+            $rightGroup->push($previousAndNextGroup);
+
+            if ($component->getShowAdd()) {
+                $rightGroup->push(FormAction::create('doNew')
+                    ->setUseButtonTag(true)
+                    ->setAttribute('data-grid-state', $this->getRequest()->getVar('gridState'))
+                    ->addExtraClass('btn btn-primary font-icon-plus rounded action--new discard-confirmation'));
+            }
```
robbieaverill: There's a lot of code here, are you able to break it up a bit into individual methods?

src/Forms/GridField/GridFieldConfig_RecordEditor.php
bergice: It is in the constructor so that this can be customised per instance, rather than statically.

Let me know if you know a better way to approach this.

src/Forms/GridField/GridFieldDetailForm.php
bergice: This should still yield the same results with the below line:

```
$this->setName($name ?: 'DetailForm');
```

src/Forms/GridField/GridFieldConfig_RecordEditor.php
robbieaverill: Oh I'm sorry!

>Unlike with other methods, PHP will not generate an E_STRICT level error message when __construct() is overridden with different parameters than the parent __construct() method has.

Ignore me

bergice: Haha, no problem. I started thinking you were correct too!

src/Forms/GridField/GridFieldDetailForm_ItemRequest.php
```diff
@@ -271,73 +271,79 @@ public function ItemEditForm()
         return $form;
     }
 
+    /**
+     * @return CompositeField Returns the right aligned toolbar group field along with its FormAction's
+     */
+    private function getRightGroupField()
```
robbieaverill: Please, we prefer protected over private.

There's also some other code looking for `$rightGroup` which is redundant now that this code is separated.

PR incoming

robbieaverill: It'd be also cool to add some more details to the PR description - either a summary of what you've changed, and/or a reference issue for what this change is related to

PR #8624, opened 2018-11-20
FIX Extensions inherited from parent classes should be retained when removing subclass extensions

### LINKED ISSUE #8613: Extensions behave inconsistent
## Affected Version
Tested on current 4.2.2

## Description

Extensions behave weird and inconsistent. 

After removing extension by calling `remove_extension`, it will not preserve inherited extensions.
Affected method is `get_extensions` which returns before calling `remove_extension` all extensions (including inherited), but after removing single extension, it will probably save wrong configuration. The result is only extensions on called class.

I was able to discover this issue by getting error in CMS. Error was that Hierarchy extension is not added to Page. There was also [opened issue](https://github.com/silverstripe/silverstripe-cms/issues/2117) in CMS repo, but it cannot be reproduced due to not calling `remove_extension`.

A picture is worth a thousand words:

screenshot_20

## Steps to Reproduce

For instance try to add extension on Page class (which extends SiteTree) and then remove_extension.

robbieaverill:
The test added in this PR shows the bug is valid. From my testing I could see that by the time the delta config change is applied to the subclass to remove some extension, the inherited list of extensions including parent extensions has been merged and normalised on the class config. This means that when we only ask for the uninherited extensions in `remove_extension()` and set the new config array back again, we overwrite any parent extensions.

Fixes #8613

### DIFF
--- a/src/Core/Extensible.php
+++ b/src/Core/Extensible.php
@@ -255,7 +255,6 @@ trait Extensible
             ) {
                 $found = true;
                 unset($config[$key]);
-                break;
             }
         }
         // Don't dirty cache if no changes
--- a/src/Core/Extensible.php
+++ b/src/Core/Extensible.php
@@ -245,9 +245,9 @@ trait Extensible
     {
         $class = get_called_class();
 
-        // Build filtered extension list
+        // Build extension list
         $found = false;
-        $config = Config::inst()->get($class, 'extensions', Config::EXCLUDE_EXTRA_SOURCES | Config::UNINHERITED) ?: [];
+        $config = Config::inst()->get($class, 'extensions') ?: [];
         foreach ($config as $key => $candidate) {
             // extensions with parameters will be stored in config as ExtensionName("Param").
             if (strcasecmp($candidate, $extension) === 0 ||
@@ -255,6 +255,7 @@ trait Extensible
             ) {
                 $found = true;
                 unset($config[$key]);
+                break;
             }
         }
         // Don't dirty cache if no changes
--- a/tests/php/Core/ObjectTest.php
+++ b/tests/php/Core/ObjectTest.php
@@ -329,6 +329,39 @@ class ObjectTest extends SapphireTest
         );
     }
 
+    public function testRemoveExtensionPreservesParentExtensions()
+    {
+        ObjectTest\ExtensionRemoveTestSubclass::add_extension(ExtendTest2::class);
+        $this->assertTrue(
+            ObjectTest\ExtensionRemoveTestSubclass::has_extension(ExtendTest1::class),
+            'Extensions should be inherited from parent classes'
+        );
+        $this->assertTrue(
+            ObjectTest\ExtensionRemoveTestSubclass::has_extension(ExtendTest2::class),
+            'Extensions added to a subclass should be registered to the subclass'
+        );
+
+        ObjectTest\ExtensionRemoveTestSubclass::remove_extension(ExtendTest2::class);
+        $this->assertTrue(
+            ObjectTest\ExtensionRemoveTestSubclass::has_extension(ExtendTest1::class),
+            'Non-specified extensions inherited from a parent class should not be removed from subclass'
+        );
+        $this->assertFalse(
+            ObjectTest\ExtensionRemoveTestSubclass::has_extension(ExtendTest2::class),
+            'Specified extensions should be removed from specified class'
+        );
+
+        ObjectTest\ExtensionRemoveTestSubclass::remove_extension(ExtendTest1::class);
+        $this->assertFalse(
+            ObjectTest\ExtensionRemoveTestSubclass::has_extension(ExtendTest1::class),
+            'Specified extensions inherited from parent class should be removed from specified class'
+        );
+        $this->assertTrue(
+            ObjectTest\ExtensionRemoveTest::has_extension(ExtendTest1::class),
+            'Parent class should still have an extension that is removed from its children'
+        );
+    }
+
     public function testRemoveExtensionWithParameters()
     {
         ObjectTest\ExtensionRemoveTest::add_extension(ExtendTest2::class . '("MyParam")');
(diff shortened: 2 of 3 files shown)


robbieaverill:
cc @silverstripe/core-team

sminnee:
Does this relate to #8465 / #8581 at all? i.e. is it a better fix for the underlying issue? Or are we just coincidentally addressing 2 extension bugs at the same time? ;-)

My main concern with this change is that it only works as intended if `Config::modify()` has been called prior to `Object::remove_extension()`. This is potentially the case in the entirety of the test system, but not the entirety of all SS runtimes.

The question is: is `Config::modify()` clear about whether it should be passed extra / inherited values or not?

The implementation of MemoryConfigCollection suggests that `set()` should be passed the entire value, factoring in the effects of any config middleware (which is how inheritance and extra extensions are provided). So this fix would be correct.

However, I wonder if we need to make this explicit in the MutableConfigCollectionInterface docs, and maybe add some tests to ensure that this behaviour is retained in MemoryConfigCollection (this would be a silverstripe/config change).

I think that treating this as a but in MemoryConfigCollection, while it would arguably be more intuitive for developers, would lead us down a config-complicating rabbit-hole, and so locking in the current behaviour (with docs and tests) is probably best.

@tractorcow or @micmania1 may have a view, they built a lot of this...

robbieaverill:
Yeah would be good to get weigh in from others.

This can be replicated in `app/_config.php` as well, without mutating config manifests (voluntarily). I guess adding and removing extensions with PHP will do that behind the scenes though.

As for whether this is the same issue as the other one we discussed, my gut feeling is that it's coincidental - this bug exists in all 4.x release lines, but I'll certainly run some tests on the other one tomorrow =)

maxime-rainville:
It's on the OS Peer Review board. We just didn't get a chance to look at it yet. I'll try to have a look at it this afternoon.

tractorcow:
This seems to be attempting to resolve an inconsistency by explicitly allowing it to remain inconsistent. As an analogy, this is similar to trying to remove an inherited method from a subclass, without removing it from the parent class.

If an extension is added to a parent class, and you attempt to remove it from a subclass, it should be an error IMO. Otherwise you run the risk of an inconsistent application.

That aside there are other issues in the code.

src/Core/Extensible.php
```diff
@@ -245,9 +245,9 @@ public static function remove_extension($extension)
     {
         $class = get_called_class();
 
-        // Build filtered extension list
+        // Build extension list
         $found = false;
-        $config = Config::inst()->get($class, 'extensions', Config::EXCLUDE_EXTRA_SOURCES | Config::UNINHERITED) ?: [];
+        $config = Config::inst()->get($class, 'extensions') ?: [];
```
tractorcow: If you don't have EXCLUDE_EXTRA_SOURCES it'll call extensions when calculating the extensions. Seems risky.

tractorcow: Instead of removing UNINHERITED isn't it worth finding out the specific class the extension is added to?

src/Core/Extensible.php
```diff
@@ -245,9 +245,9 @@ public static function remove_extension($extension)
     {
         $class = get_called_class();
 
-        // Build filtered extension list
+        // Build extension list
         $found = false;
-        $config = Config::inst()->get($class, 'extensions', Config::EXCLUDE_EXTRA_SOURCES | Config::UNINHERITED) ?: [];
```
tractorcow: This is a big change in behaviour for 4.0 that isn't strictly a bugfix.

tests/php/Core/ObjectTest.php
```diff
@@ -329,6 +329,39 @@ public function testRemoveExtension()
...
+        );
+        $this->assertFalse(
+            ObjectTest\ExtensionRemoveTestSubclass::has_extension(ExtendTest2::class),
+            'Specified extensions should be removed from specified class'
+        );
+
+        ObjectTest\ExtensionRemoveTestSubclass::remove_extension(ExtendTest1::class);
+        $this->assertFalse(
+            ObjectTest\ExtensionRemoveTestSubclass::has_extension(ExtendTest1::class),
+            'Specified extensions inherited from parent class should be removed from specified class'
```
tractorcow: As per my comment in the discussion, I'm not sure I agree with this chain. It should be an error IMO to try to remove a parent class's extension from a subclass.

tractorcow:
@sminnee as far as I recall, using Extension::add/remove is always as destructive to config cache as config::modify() directly. We should really be encouraging YML based solutions instead of PHP.

src/Core/Extensible.php
```diff
@@ -255,7 +255,6 @@ public static function remove_extension($extension)
             ) {
                 $found = true;
                 unset($config[$key]);
-                break;
```
tractorcow: Good catch.

src/Core/Extensible.php
robbieaverill: We can add EXCLUDE_EXTRA_SOURCES back in again no problem I think.

This logic was actually fine the way it was, it's the fact that what comes out of it is the canonical list that gets set back to the config for the class, excluding parent extensions that were there originally. Perhaps the parent extensions should never have been there, and `has_extension()` should factor in parent class config when it checks for them

tests/php/Core/ObjectTest.php
robbieaverill: Yep this is fair enough, I can remove this from the tests

chillu:
@robbieaverill Are you planning to peer review this?

robbieaverill:
@chillu yeah I want to but haven't had time yet. Will get back to it as soon as I can

robbieaverill:
@tractorcow:

>as far as I recall, using Extension::add/remove is always as destructive to config cache as config::modify() directly. We should really be encouraging YML based solutions instead of PHP.

Do you think we should deprecate the PHP methods instead, update docs and not fix this bug?

tractorcow:
We can't really deprecate based on performance alone, but we should build a better yml mutation system that users can use instead. A custom transformer would be a good idea; One that takes specific commands (provided by yml) and can do things like lookup and remove / replace values by keys and nested values.

robbieaverill:
Sure, I'll close this for now

PR #8627, opened 2018-11-21
FIX Provide alternatives to session for storing GridField_FormAction state

### LINKED ISSUE #8589: GridField actions cause heavy sessions, data is never cleaned up
## Affected Version

Tested against SilverStripe 4.3.x-dev

## Description

GridField makes use of the session to store parameters required to trigger each action available on the grid (including a set of keys for the actions on each grid item). This works simply and effectively in basic use-cases, and if you use a traditional single-server hosting environment, you'd likely never notice any problems.

However, with the combination of two increasingly common factors, we quickly see issues with this approach:

- The [Queued Jobs module](https://github.com/symbiote/silverstripe-queuedjobs), which has a ModelAdmin that contains a _significant_ number of GridField Actions
- A cloud-based hosting environment with multiple web servers, utilising DynamoDB for session storage - which has limited Read / Write capacity that is measured both in the number _and size_ of requests.

Loading the Queued Jobs ModelAdmin when at least a page worth of jobs exist stores approximately _40_ session keys in my testing, weighing roughly 5kb. This may seem small, but it means a user who visits this interface once during their session will consume ~5 write units in DynamoDB on every request until they log out. With more complex projects that utilise a range of GridFields, this number will rise further.

We could improve this in one of two ways:

- Ensure keys are purged once they are no longer relevant (i.e. once the user navigates to another interface)
- Shift away from use of the session to store this data entirely (any reason we can't send it down to the user? CSRF concerns?)

The former is lower effort, and would be a big improvement - not significantly improving peak usage, but at least mitigating the general impact.

## Steps to Reproduce

- Install `symbiote/silverstripe-queuedjobs`
- Create enough jobs to fill a page of the QueuedJobs ModelAdmin interface
- Visit the ModelAdmin and observe excessive session use
- Navigate around the CMS and observe the `gf_` session keys are never cleared

## Related PRs
* [x] https://github.com/silverstripe/silverstripe-framework/pull/8627
* [x] https://github.com/silverstripe/silverstripe-admin/pull/757

ScopeyNZ:
Fixes #8589

Currently the session is used to store GridField action state. This PR doesn't change that _by default_. I've abstracted the storage method into an interface and provided an alternate that will just pass down the state to the DOM and then route it back when a request is built to perform an action.

I tried to go down the path of invalidating this action state but it's hard to tell exactly when the user navigates away from the GridField - especially as multiple GridFields can exist on one page, and some are loaded through AJAX requests.

Relies on silverstripe/silverstripe-admin#757

### DIFF
--- a/src/Forms/GridField/FormAction/AttributeStore.php
+++ b/src/Forms/GridField/FormAction/AttributeStore.php
@@ -1,12 +1,10 @@
 <?php
 namespace SilverStripe\Forms\GridField\FormAction;
 
-use SilverStripe\Control\HTTPRequest;
-
 /**
  * Stores GridField action state on an attribute on the action and then analyses request parameters to load it back
  */
-class AttributeStore extends AbstractRequestAwareStore implements StateStore
+class AttributeStore extends AbstractRequestAwareStore
 {
     /**
      * Save the given state against the given ID returning an associative array to be added as attributes on the form
--- a/src/Forms/GridField/FormAction/AttributeStore.php
+++ b/src/Forms/GridField/FormAction/AttributeStore.php
@@ -1,0 +1,51 @@
+<?php
+namespace SilverStripe\Forms\GridField\FormAction;
+
+use SilverStripe\Control\HTTPRequest;
+
+/**
+ * Stores GridField action state on an attribute on the action and then analyses request parameters to load it back
+ */
+class AttributeStore implements StateStore
+{
+    /**
+     * @var HTTPRequest
+     */
+    protected $request;
+
+    /**
+     * @param HTTPRequest $request
+     */
+    public function __construct(HTTPRequest $request)
+    {
+        $this->request = $request;
+    }
+
+    /**
+     * Save the given state against the given ID returning an associative array to be added as attributes on the form
+     * action
+     *
+     * @param string $id
+     * @param array $state
+     * @return array
+     */
+    public function save($id, array $state)
+    {
+        // Just save the state in the attributes of the action
+        return [
+            'data-action-state' => json_encode($state),
+        ];
+    }
+
+    /**
+     * Load state for a given ID
+     *
+     * @param string $id
+     * @return mixed
+     */
+    public function load($id)
+    {
+        // Check the request
+        return json_decode($this->request->requestVar('ActionState'), true);
+    }
+}
--- a/src/Forms/GridField/FormAction/AttributeStore.php
+++ b/src/Forms/GridField/FormAction/AttributeStore.php
@@ -6,21 +6,8 @@ use SilverStripe\Control\HTTPRequest;
 /**
  * Stores GridField action state on an attribute on the action and then analyses request parameters to load it back
  */
-class AttributeStore implements StateStore
+class AttributeStore extends AbstractRequestAwareStore implements StateStore
 {
-    /**
-     * @var HTTPRequest
-     */
-    protected $request;
-
-    /**
-     * @param HTTPRequest $request
-     */
-    public function __construct(HTTPRequest $request)
-    {
-        $this->request = $request;
-    }
-
     /**
      * Save the given state against the given ID returning an associative array to be added as attributes on the form
      * action
@@ -41,11 +28,11 @@ class AttributeStore implements StateStore
      * Load state for a given ID
      *
      * @param string $id
-     * @return mixed
+     * @return array
      */
     public function load($id)
     {
         // Check the request
-        return json_decode($this->request->requestVar('ActionState'), true);
+        return (array) json_decode((string) $this->getRequest()->requestVar('ActionState'), true);
     }
 }
--- a/src/Forms/GridField/FormAction/SessionStore.php
+++ b/src/Forms/GridField/FormAction/SessionStore.php
@@ -1,0 +1,50 @@
+<?php
+namespace SilverStripe\Forms\GridField\FormAction;
+
+use SilverStripe\Control\HTTPRequest;
+
+/**
+ * Stores GridField action state in the session in exactly the same way it has in the past
+ */
+class SessionStore implements StateStore
+{
+    /**
+     * @var HTTPRequest
+     */
+    protected $request;
+
+    /**
+     * @param HTTPRequest $request
+     */
+    public function __construct(HTTPRequest $request)
+    {
+        $this->request = $request;
+    }
+
+    /**
+     * Save the given state against the given ID returning an associative array to be added as attributes on the form
+     * action
+     *
+     * @param string $id
+     * @param array $state
+     * @return array
+     */
+    public function save($id, array $state)
+    {
+        $this->request->getSession()->set($id, $state);
+
+        // This adapter does not require any additional attributes...
+        return [];
+    }
+
+    /**
+     * Load state for a given ID
+     *
+     * @param string $id
+     * @return mixed
+     */
+    public function load($id)
+    {
+        return $this->request->getSession()->get($id);
+    }
+}
--- a/src/Forms/GridField/FormAction/SessionStore.php
+++ b/src/Forms/GridField/FormAction/SessionStore.php
@@ -6,21 +6,8 @@ use SilverStripe\Control\HTTPRequest;
 /**
  * Stores GridField action state in the session in exactly the same way it has in the past
  */
-class SessionStore implements StateStore
+class SessionStore extends AbstractRequestAwareStore implements StateStore
 {
-    /**
-     * @var HTTPRequest
-     */
-    protected $request;
-
-    /**
-     * @param HTTPRequest $request
-     */
-    public function __construct(HTTPRequest $request)
-    {
-        $this->request = $request;
-    }
-
     /**
      * Save the given state against the given ID returning an associative array to be added as attributes on the form
      * action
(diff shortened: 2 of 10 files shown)


src/Forms/GridField/FormAction/AttributeStore.php
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Load state for a given ID
+     *
+     * @param string $id
+     * @return mixed
+     */
+    public function load($id)
+    {
+        // Check the request
+        return json_decode($this->request->requestVar('ActionState'), true);
```
robbieaverill: It might be worth casting the requestVar as a string, since `json_decode(true, true)` would return `int(1)`. Maybe also update the PHPdoc to be explicit about what it can return instead of "mixed"

src/Forms/GridField/FormAction/StateStore.php
```diff
@@ -0,0 +1,30 @@
...
+namespace SilverStripe\Forms\GridField\FormAction;
+
+use SilverStripe\Control\HTTPRequest;
+
+interface StateStore
+{
+    /**
+     * @param HTTPRequest $request
+     */
+    public function __construct(HTTPRequest $request);
```
robbieaverill: I'm in two minds about whether it's a good idea to have the constructor in this interface

src/Forms/GridField/FormAction/StateStore.php
```diff
@@ -0,0 +1,30 @@
...
+     * @param array $state
+     * @return array
+     */
+    public function save($id, array $state);
+
+    /**
+     * Load state for a given ID
+     *
+     * @param string $id
+     * @return mixed
```
robbieaverill: We should be more specific than this - are we expecting an array?

src/Forms/GridField/FormAction/StateStore.php
robbieaverill: I've just read this and now believe we should remove the constructor: https://www.alainschlesser.com/including-constructor-interface/

ScopeyNZ: Yeah at the time I wasn't really happy with this code either. I need the request in both implementations of the interface and I don't think it's a good pattern to fetch the request from global state (`Controller::curr` or `Injector`). I'll just suck it up though; remove the constructor and use `Controller::curr`.

robbieaverill: You could always give it a setter and use injector calls to provide it

robbieaverill: Or put the constructor in an abstract parent class - separating the implementation from the interface entirely

robbieaverill:
There are also some minimal docs that could use a nudge too: https://docs.silverstripe.org/en/4/developer_guides/forms/field_types/gridfield/#saving-the-gridfield-state

ScopeyNZ:
Thanks for the heads up. The `GridField` state is distinct form `GridField_FormAction` state. Something that took me a few moments to realise when I started this work 😅 .

I'll add a section about configuring it to use attributes/request params instead of session.

Otherwise I'm done with implementation now.

ScopeyNZ:
Docs pushed 🙂

src/Forms/GridField/FormAction/AttributeStore.php
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Load state for a given ID
+     *
+     * @param string $id
+     * @return array
+     */
+    public function load($id)
+    {
+        // Check the request
+        return (array) json_decode((string) $this->getRequest()->requestVar('ActionState'), true);
```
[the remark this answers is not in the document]
ScopeyNZ: To assert that `null` comes through as `[]`. I could omit the second argument but I'd imagine getting an object and then casting to array would be a performance hit...

src/Forms/GridField/FormAction/AttributeStore.php
```diff
@@ -0,0 +1,38 @@
+<?php
+namespace SilverStripe\Forms\GridField\FormAction;
+
+use SilverStripe\Control\HTTPRequest;
+
+/**
+ * Stores GridField action state on an attribute on the action and then analyses request parameters to load it back
+ */
+class AttributeStore extends AbstractRequestAwareStore implements StateStore
```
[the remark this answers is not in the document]
ScopeyNZ: Fair call. I'll change this and add some tests hopefully tomorrow.

src/Forms/GridField/GridField.php
```diff
@@ -1009,9 +1012,14 @@ public function gridFieldAlterAction($data, $form, HTTPRequest $request)
...
+        // Fetch the store for the "state" of actions (not the GridField)
+        /** @var StateStore $store */
+        $store = Injector::inst()->create(StateStore::class . '.' . $this->getName());
+
         foreach ($data as $dataKey => $dataValue) {
             if (preg_match('/^action_gridFieldAlterAction\?StateID=(.*)/', $dataKey, $matches)) {
-                $stateChange = $request->getSession()->get($matches[1]);
+                $stateChange = $store->load($matches[1]);
+
                 $actionName = $stateChange['actionName'];
```
maxime-rainville: I get a 500 thrown back in my face on this line if I set `AttributeStore` as my default `SilverStripe\Forms\GridField\FormAction\StateStore`. To replicate:
1. Put the following in your YML config
```yml
---
Name: myproject
After: '#gridfieldconfig'
---
  SilverStripe\Forms\GridField\FormAction\StateStore:
    class: SilverStripe\Forms\GridField\FormAction\AttributeStore
```
2. Access a gridifield in a model admin (I tried the Company and Employee ones from frameworktest)
3. Try Deleting/Archiving an entry via the GridField action.

I get the following error returned to me.
```
[Notice] Undefined index: actionName
POST /admin/test/SilverStripe-FrameworkTest-Model-Employee/EditForm/field/SilverStripe-FrameworkTest-Model-Employee

Line 1023 in /home/maxime/Projects/ss43/vendor/silverstripe/framework/src/Forms/GridField/GridField.php
```

ScopeyNZ: Just double checking that you have the admin dependant PR also checked out?

maxime-rainville: Just had a look. I'm running the `pulls/4.3/abstr-action-state` branch of framework and the `pulls/1.3/may-the-state-be-with-you` branch of admin. I made sure I had the latest changed pulled.

Here's the exact request that is returning a 500 for me.

```
POST http://ss43.max.silverstripe.com/admin/test/SilverStripe-FrameworkTest-Model-Company/EditForm/field/SilverStripe-FrameworkTest-Model-Company

Host: ss43.max.silverstripe.com
User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:63.0) Gecko/20100101 Firefox/63.0
Accept: text/html, */*; q=0.01
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate
Referer: http://ss43.max.silverstripe.com/admin/test/
Content-Type: application/x-www-form-urlencoded; charset=UTF-8
X-Pjax: CurrentField
X-Requested-With: XMLHttpRequest
Content-Length: 346
Connection: keep-alive
Cookie: cms-panel-collapsed-cms-menu=false; PHPSESSID=tbpq0ipneqq2m9du0sdak79gke
Pragma: no-cache
Cache-Control: no-cache

action_gridFieldAlterAction%3FStateID%3Dgf_b8fd577d=&SilverStripe-FrameworkTest-Model-Company%5BGridState%5D=%7B%22GridFieldSortableHeader%22%3A%7B%22SortColumn%22%3A%5B%5D%7D%2C%22GridFieldFilterHeader%22%3A%7B%22Columns%22%3Anull%7D%2C%22GridFieldPaginator%22%3A%7B%22currentPage%22%3A1%7D%7D&SecurityID=75e9240c599dcd6b180a4510ee73bf40fa13071b
```

ScopeyNZ: Thanks I'll take a look 🙂

ScopeyNZ: I updated the admin PR. There was a change that I remember making but I must've `git checkout --`'d the wrong thing

maxime-rainville: Confirmed fixed.

src/Forms/GridField/FormAction/SessionStore.php
```diff
@@ -0,0 +1,37 @@
+<?php
+namespace SilverStripe\Forms\GridField\FormAction;
+
+use SilverStripe\Control\HTTPRequest;
+
+/**
+ * Stores GridField action state in the session in exactly the same way it has in the past
+ */
+class SessionStore extends AbstractRequestAwareStore implements StateStore
```
maxime-rainville: `AbstractRequestAwareStore` is implementing `StateStore` now. This implements statement is kind of redundant.

maxime-rainville:
@ScopeyNZ The Travis fail isn't related to behat. It's something about the History Viewer controller. Is this related to your PR?

ScopeyNZ:
I think this was _just_ resolved in CMS. Rerunning the job.

maxime-rainville:
Cool ... I'll merge on green.

PR #8633, opened 2018-11-25
Referencing env vars in configs

falnyr:
As mentioned here: https://stackoverflow.com/a/53459699/2111274

### DIFF
--- a/docs/en/00_Getting_Started/03_Environment_Management.md
+++ b/docs/en/00_Getting_Started/03_Environment_Management.md
@@ -48,6 +48,17 @@ use SilverStripe\Core\Environment;
 Environment::setEnv('API_KEY', 'AABBCCDDEEFF012345');
 ```
 
+### Using environment variables in config
+
+To use environment variables in `.yaml` configs you can reference them using backticks.
+
+```yaml
+SilverStripe\Core\Injector\Injector:
+  MyServiceClass:
+    properties:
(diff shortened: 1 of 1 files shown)


docs/en/00_Getting_Started/03_Environment_Management.md
```diff
@@ -48,6 +48,17 @@ use SilverStripe\Core\Environment;
...
+### Using environment variables in config
+
+To use environment variables in `.yaml` configs you can reference them using backticks.
+
+```yaml
+SilverStripe\Core\Injector\Injector:
+  MyServiceClass:
+    properties:
+      MyProperty: '`ENV_VAR_HERE`'
+```
```
robbieaverill: It might be worth noting that this only works for injector configuration

ScopeyNZ:
I wonder if this is better suited in the [configuration docs](https://docs.silverstripe.org/en/4/developer_guides/configuration/). I am a little surprised it's missing.

falnyr:
> I wonder if this is better suited in the [configuration docs](https://docs.silverstripe.org/en/4/developer_guides/configuration/). I am a little surprised it's missing.

I think it can be in both (mainly here). Usually I look for environment variables and how I can use them in a framework without needing to know there is a "Configuration API".

robbieaverill:
I agree

ScopeyNZ:
Thanks @falnyr ! The change should propagate to the docs over the next few days

PR #8634, opened 2018-11-26
FIX allow for Injector service dot notation to fall-back to classes

NightJar:
A feature was introduced to Injector with SilverStripe 4 allowing
developers to define services in the format
`ServiceName.SubAlias`
allowing for ServiceName to have contextual information to allow it
to be used in a modified way for a particular situation where that
context is relevant. The idea being that if a particular contextual
alias of a service is undefined, instantiation will fall back to
the service proper.

Injector also allows for an undefined service to be treated as a
classname, and attempt to instantiate it as such. This has always
been a feature so far as I know.

However the two features are incompatible, a contextual service will
only fall back to another pre-defined service, which is contrary to
expectation if the fallback is a valid classname, and an undefined
service.

So we can fix this by treating the section of the service definition
preceding a dot as we would with any service definition without a
dot - that is if the service is undefined, to treat the name as a
fully qualified class name.

### DIFF
--- a/src/Core/Injector/Injector.php
+++ b/src/Core/Injector/Injector.php
@@ -1037,7 +1037,7 @@ class Injector implements ContainerInterface
         } else {
             // Late-generate config spec for non-configured spec
             $spec = [
-                'class' => $name,
+                'class' => strtok($name, '.'),
                 'constructor' => $constructorArgs,
             ];
         }
--- a/tests/php/Core/Injector/InjectorTest.php
+++ b/tests/php/Core/Injector/InjectorTest.php
@@ -1047,4 +1047,20 @@ class InjectorTest extends SapphireTest
         Injector::unnest();
         $this->nestingLevel--;
     }
+
+    public function testServiceConfigFallback()
+    {
+        $config = [
+            SomeService::class . '.Another' => AnotherService::class,
+        ];
+        $injector = new Injector($config);
+
+        // test 'sub-service'
+        $subservice = $injector->create(SomeService::class . '.Another');
+        $this->assertInstanceOf(AnotherService::class, $subservice, 'dot notation service should work as normal');
+
+        // test falling back
+        $fallback = $injector->create(AnotherService::class . '.Fallback');
+        $this->assertInstanceOf(AnotherService::class, $fallback, 'undefined dot notation should fall back');
+    }
 }


NightJar:
**PR currently WIP**, tests are expected to fail (I want to see them fail, and testing framework outside of a project is too hard - change was made this way because 4.0 target)

robbieaverill:
>The idea being that if a particular contextual
>alias of a service is undefined, instantiation will fall back to
>the service proper.

I'm not familiar with this to be honest. I know we allow "named services" by grouping similar services by an interface that they implement, but I don't remember hearing that we expect it to fallback to the unnamed service definition if the named definition doesn't exist.

I think this would be a risky change to introduce as a patch change.

If you take cache definitions for example, each individual cache service is registered as a named service definition under CacheInterface:

```yaml
  Psr\SimpleCache\CacheInterface.myapp:
    factory: SilverStripe\Core\Cache\CacheFactory     
    constructor:
      namespace: "MyInsensitiveData"
      args:
        disable-container: true 
```

In this case, we would probably not want this to fall back to a generic cache implementation. I think cache definitions are an example of why this behaviour might not be desired.

Another point worth bringing up - if we've written code that uses `MyService::class . '.my_specific_implementation'` as a dependency, presumably we've also written the service definition for injector, so why would we expect it not to resolve correctly?

chillu:
Yeah I'm inclined to agree with Robbie. If you instantiate a named service scoped with dot notation, and there is no service registered under that full string, it should throw rather than silently fall back to a potentially generic implementation. What's the use case for an "optionally available service with dot notation"?

chillu:
No further feedback on this, so I'm closing as "won't fix"

NightJar:
Haven't had time sorry, still planning to come back to it.

NightJar:
OK, lets step through this.

# Injector

An [undocumented](https://docs.silverstripe.org/en/4/developer_guides/extending/injector/) feature exists in Injector since [3.0.0-beta3](https://github.com/silverstripe/silverstripe-framework/blob/3.0.0-beta3/control/injector/Injector.php#L603-L612) allowing a service to be suffixed with a `.` seemingly in order to be "extended" of sorts.

This feature does not actually allow extension in that no properties or configuration are inherited, rather the purpose is to provide a _fallback_ for the case that a full service name does not exist. This allows an author to provide a type of _specificity_ not unlike CSS, and is useful in situations where someone attempting to inject a service wishes to do so in a particular use-case and not globally throughout their project.

This feature is directly analogous to the concept of [`context`](https://github.com/silverstripe/silverstripe-framework/blob/4.0.0-beta1/docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md#injector-context) in SilverStripe 4's JavaScript Injector.

```yml
SomeService: DefaultClass
SomeService.InThisParticularUseCase:
  class: SpecialtyClass
  properties:
    # etc.
```

## SilverStripe Lookup

The ability to ask Injector for the service by an existing class name without first defining a service of that class name is a long standing feature of [Injector and SilverStripe](https://github.com/silverstripe/silverstripe-framework/blob/3.0.0-beta3/control/injector/Injector.php#L720).

>The following sums up the simplest usage of the Injector it creates a new object of type MyClassName through create
>```
>use SilverStripe\Core\Injector\Injector;
>
>$object = Injector::inst()->create('MyClassName');
>```
>The benefit of constructing objects through this syntax is ClassName can be swapped out using the Configuration API by developers.

\- _The [docs](https://docs.silverstripe.org/en/4/developer_guides/extending/injector/) implicitly stating `create`ing a true class name without service definition is OK._

## Instantiation

This leaves a (developing) user with the facts:
 - Services names are looked up as many times there are sections separated by `.`
 - Sections of the name are popped after each unsuccessful lookup.
 - The first match to a registered service is instantiated and returned.
 - An error is thrown if no services are found.
 - **If a service matches a class name, it is _always_ registered**
 - Writing services with [Configuration](https://docs.silverstripe.org/en/4/developer_guides/configuration/) is the way to override default services provided by SilverStripe

# Customisation

If a user wishes to override a default service in their specific use case only, then they should be able to write configuration YML in the following form:
`SilverStripe\Core\Example.Context: Vendor\Module\CustomExample`

## Use case

It may be because the user is trying to configure a core class used throughout the system
Or a third party dependency class, perhaps leaving an extension point for other users to be nice.

It could be that the user is content with the defaults for a class, but can foresee that other users may wish to tweak this configuration, and it is entirely impossible to know if someone's project may be using, and how - in due to this global reconfiguration would be less than ideal (c.f. the case for namespaces).

## Problem

This proves false, as `hasService` is false, even when querying actively injected services. This is bad because it breaks the assumptions and provides an inconstant API to the user.

```php
// `use` statements here
// [...]
GridField::create(); // A GridField
$injector = Injector::inst();
$injector->create(GridField::class); // Another GridField
$injector->hasService(GridField::class); // false
```

Because of this, we end up with the following disparity:

```php
// `use` statements here
// [...]
$injector = Injector::inst();

// 1. An injected class as defined in YML
$injector->create('ConfiguredService');
// 2. An injected class configured differently to the above, but still as defined in YML
$injector->create('ConfiguredService.DefiendInConfig');
// 3. Effectively the same as the first example, as it is not defined in YML
$injector->create('ConfiguredService.Undefined');

// 4. An injected class, which may or may not be defined in YML
$injector->create('GridField');
// 5. An injected class configured differently to the above, as it is defined in YML
$injector->create('GridField.DefiendInConfig');
// 6. An error, although expectation is set that it would be the same as the 4th example
$injector->create('GridField.Undefined');
```

## Work around

Subclassing in order to alter configuration independently defeats the point of injection, as does instantiating and configuring in code.

## Fix

Configure the automated fallback to instantiate the service name as a class, even when context is supplied. The context at the point will have already been found to not exist, so this is safe, and results in already present behaviour consistent with not supplying context.

# Objection

## Failing over sounds dangerous

### Interfaces

> [...] by grouping similar services by an interface that they implement [...]
> [...] we would probably not want this to fall back to a generic cache implementation.

This is safe. An interface cannot be instantiated, and will always cause an error.
```php
<?php
interface AnInterface {}
class AClass implements AnInterface {}
$reflector = new ReflectionClass(AnInterface::class);
var_dump(
  class_exists(AnInterface::class), // false
  class_exists(AClass::class), // true
  $reflector->newInstanceArgs([]) // PHP Fatal error:  Uncaught Error: Cannot instantiate interface AnInterface
);
```

### Unexpected results

> [...] it should throw rather than silently fall back to a potentially generic implementation.

The definitions are not generic. The _requestor_ of the service is specifically asking for "This one if it exists, or the default is OK too". Someone must ask directly for a contextual service - there is no silent fail-over.

This change will not make already set up service instantiation of any kind magically resolve to something new and unexpected. The same is true for `Interface` 'grouped' style definitions as above, the requestor must be asking for it specifically. **The control sits with the consumer, not the supplier.**

If a user wishes to have their own private service that will not fail over to anything ever, then they can ommit the `.`, or the entire first segment: `SilverStripe\Forms\TextField_ThatICustomised` will never fail-over, and still provides a reasonable request interface (1 `char` difference in code), and keeps a clear note of its usage intention and intended interface within the YML, if that should be a concern.

## Why?

> presumably we've also written the service definition for injector, so why would we expect it not to resolve correctly?

>What's the use case for an "optionally available service with dot notation"?

Namespaces are good, they provide enough context that we can be reasonably sure that we are configuring exactly the thing we want to, but not always. It is feasible and reasonable to wish to configure third party dependencies via Injector. We know that [global scope is not ideal](http://wiki.c2.com/?GlobalVariablesAreBad), however this is what Injector with YML configuration introduces, especially as singletons. A user may be consuming this interface in a single use case, and be happy with out of the box solution provided with zero configuration. However they may wish to use this third party code in several places, or it may be already in use within a large project that they're introducing a new feature to. Greedily and selfishly reconfiguring the context used in all cases throughout the code base is not ideal, so as the module author we can [provide an extension point](https://github.com/silverstripe/silverstripe-framework/blob/4.0.0/src/ORM/DataObject.php#L1041-L1053) to allow modification in a single case via contextual selection for injection.

**TL;DR; Implementor is happy with the default, but others may not be. They are writing extensible code.**

E.g. Configuring logging.
Writing code to be used by many others. Logging to the default stream is just fine, for us, as a default usage. Perhaps others may wish to split their streams by business domain. We're using an interface implementation we didn't write, and it is used used by many many other services. We cannot reconfigure that service, nor should we extend it unnecessarily.

robbieaverill:
Wow, epic write up! Here are some thoughts off the top of my head:

>This feature does not actually allow extension in that no properties or configuration are inherited, rather the purpose is to provide a fallback for the case that a full service name does not exist. This allows an author to provide a type of specificity not unlike CSS, and is useful in situations where someone attempting to inject a service wishes to do so in a particular use-case and not globally throughout their project.

I think this is an assumption that you've made about how it should work, not how it does work or is intended to work.

>This proves false, as hasService is false, even when querying actively injected services. This is bad because it breaks the assumptions and provides an inconstant API to the user.

It's important to note that your examples are not instantiating services - the `create()` method is used to instantiate new instances, not get (or create) service classes (singletons). By using `get()` you'd be getting (or creating) service instances, after which `->hasService()` would return true. `Injector::inst()->create('Foo)` is equivalent to `new Foo()` or `Foo::create()`, which should not register that instance as a service with Injector.

> [Fix] Configure the automated fallback to instantiate the service name as a class, even when context is supplied.

I don't think you should treat dot notated Injector definitions as extra context, they are standalone definitions - only grouped with dots for the sake of semantically keeping similar definitions grouped together.

> [Interfaces] This is safe. An interface cannot be instantiated, and will always cause an error.

You can instantiate interfaces with Injector, provided that Injector has a concrete definition for the class to use when instantiating an interface. We've used this pattern recently [in the ckan-registry module](https://github.com/silverstripe/silverstripe-ckan-registry/blob/20717b7f31d1ccaeb28ef8696adb06d72e2baacb/src/Model/Resource.php#L48-L50).

>If a user wishes to have their own private service that will not fail over to anything ever, then they can ommit the ., or the entire first segment: SilverStripe\Forms\TextField_ThatICustomised will never fail-over, and still provides a reasonable request interface (1 char difference in code), and keeps a clear note of its usage intention and intended interface within the YML, if that should be a concern.

I think you're just shifting the way the dot works at the moment into another arbitrary character (_) in order to change the way dot notation behaves - I think it's a risky change to introduce in 4.x at least.

---

I think the idea of allowing specificity based Injector configuration via dot notation is a good one, but I don't think it's a safe change to make in 4.x (especially not as a patch change) and could be raised as an RFC separately if you'd like to!

NightJar:
> Wow, epic write up! Here are some thoughts off the top of my head:

Thanks, however it seems I've still not communicated very well. To be clear I'm going to sound very blunt, so sorry in advance.

The dot notation is _NOT_ simply a naming convention.
**Nothing is being introduced with this proposed change**
I am proposing to make existing functionality more consistent.
**Existing core behaviour will _not_ change**

> It's important to note that your examples are not instantiating services - the `create()` method is used to instantiate new instances, not get (or create) service classes (singletons). By using `get()` you'd be getting (or creating) service instances, after which `->hasService()` would return true. `Injector::inst()->create('Foo)` is equivalent to `new Foo()` or `Foo::create()`, which should not register that instance as a service with Injector.

Yes that's true, however `get` and `create` are functionally equivalent excepting three important differences:

*create* will obey constructor parameters every time.
*get* will register a service if one does not exist, iff the service name is a valid class name.
*get* will store the first instance it creates as a named singleton.

`create` is simply a proxy for `get`, with the `isSingleton` argument always set to `false`. The important part is that because of this, they both search for specifications in the same manner. A call to `create` still uses a service specification to instantiate the object the same as `get` does - and one can define a service that is equivalent to `create` by using the `type: prototype` property of the spec.

However that did dilute my point a little, sorry for dangling a red herring.

> I don't think you should treat dot notated Injector definitions as extra context, they are standalone definitions - only grouped with dots for the sake of semantically keeping similar definitions grouped together.

It is true that each dot notated Injector definition is its own standalone definition - nothing is inherited as I mentioned above. However this is not simply a naming convention, regardless of how it appears to be used.

> You can instantiate interfaces with Injector, provided that Injector has a concrete definition for the class to use when instantiating an interface. We've used this pattern recently [in the ckan-registry module](https://github.com/silverstripe/silverstripe-ckan-registry/blob/20717b7f31d1ccaeb28ef8696adb06d72e2baacb/src/Model/Resource.php#L48-L50).

This is a bit of a straw man - one can never instantiate interfaces. One has the ability to request a defined service with a name that matches a valid interface, yes. But it is irrelevant to the point, and equally relies on a developer making this definition.
If someone were to do this however, it would then serve as the default for all dot notated 'sub'/'contextual' definitions extending from that interface name.

> I think you're just shifting the way the dot works at the moment into another arbitrary character (_) in order to change the way dot notation behaves - I think it's a risky change to introduce in 4.x at least.

This is not a change. My intention here was to highlight that such a naming convention could preserve your preexisting understanding of what the dot is - that is purely as a naming convention.
Doing this has a single benefit in that when the PHP code requests `$class . '_Extra'` it will _never_ 'fail over' to the base definition. It would simply be naming convention, presumably used to ease recognition in YML.

Although equally - continuing to use the dot is safe. As above, nothing is inherited, existing code will not have to change; Only in the _absence_ of a dot notated context definition does the 'fall back' occur.

> I think the idea of allowing specificity based Injector configuration via dot notation is a good one, but I don't think it's a safe change to make in 4.x (especially not as a patch change) and could be raised as an RFC separately if you'd like to!

I'm glad you think it's a good idea, because it's already a part of 4.x :)

And 3.x for that matter. This functionality has existed since the [introduction of Injector](https://github.com/nyeholt/injector/blob/8264775a85af19af687d00f7d93c9e39b629ed5c/src/November/Injection/Injector.php#L678-L684), however it's only since 4.x that I've seen it begin to be used more commonly - to the point that I thought it was a new feature at 4.0.0 - but upon investigation I was wrong.


Here is a bash script you can run to see this all working as I describe it in 4.3.0:

In summary the steps here are:
1. Install 4.3.0 with installer
2. Add a simple class with an injectable property, and a way to access it
3. Create 2 service definitions for this class, one with and one without a dot suffix
4. Alter `PageController` to use the dot suffix to get the service & output the injected property
5. Test - see the output of the dot suffixed definition
6. Delete the dot suffixed service definition - but do not change the instantiation in `PageController`!
7. Test - see the output of the 'base' service definition
8. Change `PageController` to issue a `notice` level `user_error`
9. Test - see SilverStripe formatted error message
10. Remove the logging definitions in `framework/_config/logging.yml`
11. Test - see a PHP level fatal error - showing that 'default services' of Interfaces are safe.

```sh
#!/bin/sh
composer create-project silverstripe/installer injector.test 4.3.0
cd injector.test/
cat > .env << ENV
SS_ENVIRONMENT_TYPE=dev
SS_DATABASE_USERNAME=root
SS_DATABASE_PASSWORD=root
SS_DATABASE_CHOOSE_NAME=1
ENV
cat > app/src/Talk.php << 'DOT'
<?php

class Talk
{
    private $phrase = 'Default';

    public function setPhrase($phrase)
    {
        $this->phrase = $phrase;
    }

    public function speak()
    {
        echo $this->phrase . PHP_EOL;
    }
}
DOT
cat > app/_config/dot.yml << 'YML'
  Talk.test:
    class: Talk
    properties:
      Phrase: "Dot context"
  Talk:
    class: Talk
    properties:
      Phrase: "Fall back"
YML
sed -i '30c            die(SilverStripe\\Core\\Injector\\Injector::inst()->get("Talk.test")->speak());' app/src/PageController.php
vendor/bin/sake dev/build '' quiet=1
echo "There are 2 services defined. 'Talk.test' & 'Talk'"
echo "The code always asks and prints output from 'Talk.test'"
echo
echo "Test as a control:"
vendor/bin/sake / '' flush=1
echo "Output is you expect - 'Dot context'"
echo '[enter] to continue...'
read
sed -i '2,5d' app/_config/dot.yml
echo
echo "With 'Talk.test' undefined:"
vendor/bin/sake / '' flush=1
echo "Output is as I expect - 'Fall back'"
echo '[enter] to continue...'
read
rm app/_config/dot.yml
echo
echo "With neither 'Talk.test' nor 'Talk' defined:"
vendor/bin/sake / '' flush=1 2>1 | head -n1
echo "Output should be 'Default' - Instead there is a fatal error. This is what I'm trying to fix."
echo '[enter] to continue...'
read
rm app/src/Talk.php
echo
echo "So now lets look at interfaces, using logging examples (framework/_config/logging.yml):"
sed -i '30c            user_error("dottest");' app/src/PageController.php
vendor/bin/sake / '' flush=1 2>1 | head -n3
echo "Output is a normal SilverStripe formatted notice error report"
echo '[enter] to continue...'
read
sed -i '21,33s/^/#/' vendor/silverstripe/framework/_config/logging.yml
echo "And with the logging definitions commented out:"
vendor/bin/sake / '' flush=1
echo "Output is a PHP fatal error from Injector - thus Interfaces are safe, fall backs are still fatal errors."
sed -i '21,33s/^#//' vendor/silverstripe/framework/_config/logging.yml

echo "This is all existing functionality, we've used 4.3.0 stable"
```

robbieaverill:
Thanks for the example @NightJar, I think I understand what you mean now. Would you mind converting your example into some basic unit tests that show that loading `Talk` without any injector configuration will _not_ output `Default` when you call `speak()` on it? I think that would be the best way to proceed with this. I'll reopen it for now anyway.

robbieaverill:
Here are those tests:

```php
<?php

use SilverStripe\Core\Injector\Injector;

class NightJarTest extends SapphireTest
{
    public function testDependencyWithDotNotation()
    {
        Injector::inst()->load([Talk::class . '.test' => [
            'class' => Talk::class,
            'properties' => ['Phrase' => 'Dot context'],
        ]]);
        Injector::inst()->load([Talk::class => [
            'class' => Talk::class,
            'properties' => ['Phrase' => 'Fallback'],
        ]]);

        $instance = Injector::inst()->get(Talk::class . '.test');
        $this->assertSame(
            'Dot context',
            $instance->speak(),
            'Dot notated alias should be loaded'
        );
    }

    public function testDependencyWithoutDotNotation()
    {
        Injector::inst()->load([Talk::class => [
            'class' => Talk::class,
            'properties' => ['Phrase' => 'Fallback'],
        ]]);

        $instance = Injector::inst()->get(Talk::class . '.test');
        $this->assertSame(
            'Fallback',
            $instance->speak(),
            'Fallback should be used without defined dot alias'
        );
    }

    public function testDependencyWithNoInjectorRegistration()
    {
        $instance = Injector::inst()->get(Talk::class);
        $this->assertSame(
            'Default',
            $instance->speak(),
            'Default should be used without any injector definitions'
        );
    }

    public function testDotNotatedDependencyWithNoInjectorRegistration()
    {
        $instance = Injector::inst()->get(Talk::class . '.test');
        $this->assertSame(
            'Default',
            $instance->speak(),
            'Default should be used without any injector definitions'
        );
    }
}

class Talk
{
    private $phrase = 'Default';

    public function setPhrase($phrase)
    {
        $this->phrase = $phrase;
    }

    public function speak()
    {
        return $this->phrase;
    }
}

```

So the current behaviour is that asking Injector for `Talk.test` will fall back to `Talk` if it's defined and `Talk.test` isn't, but it won't load `Talk` unless it has injector configuration already - that's what this PR is changing. In this case I'm less opposed to the change. The tests still need to be green though =)

NightJar:
Thanks for reopening, I'll get to fixing those tests :)

maxime-rainville:
@NightJar Did you have a chance to have another look at this?

NightJar:
@maxime-rainville Not yet, but we do have a hackday coming up :)

unclecheese:
@NightJar Please rebase and retarget if you're still keen to get this merged.

PR #8638, opened 2018-11-27
NEW: Add visibility updates to `upgrade.yml`

bergice:
This will make the `inspect` upgrader command ensure these visibilities are used


robbieaverill:
Is there any reason this shouldn't go into 4.0?

bergice:
> Is there any reason this shouldn't go into 4.0?

Nope. I'll fix this up in a second.

PR #8639, opened 2018-11-27
Stop duplicate session cookies on every request.

caffeineinc:
Ensure only a single Set-Cookie header is returned from Session once
we have data to save. Include backwards compatibility for PHP56

Headers were being sent twice, once by session_start() and once by Cookie::set()

The second header was unneeded, and meant Set-Cookie was appended to every request which
breaks WAF behaviour for sticky sessions

@see #8543

### DIFF
--- a/src/Control/Session.php
+++ b/src/Control/Session.php
@@ -300,6 +300,7 @@ class Session
 
         // If the session cookie is already set, then the session can be read even if headers_sent() = true
         // This helps with edge-case such as debugging.
+        $data = [];
         if (!session_id() && (!headers_sent() || !empty($_COOKIE[ini_get('session.name')]))) {
             if (!headers_sent()) {
                 session_set_cookie_params($timeout, $path, $domain ?: null, $secure, true);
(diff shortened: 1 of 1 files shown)


caffeineinc:
@madmatt  @dnsl48  I've created this one from 4 so it merges better, but replaced the php56 code, so it should be backwards compatible. 

https://github.com/silverstripe/silverstripe-framework/pull/8544 for reference

src/Control/Session.php
```diff
@@ -300,6 +300,7 @@ public function start(HTTPRequest $request)
 
         // If the session cookie is already set, then the session can be read even if headers_sent() = true
         // This helps with edge-case such as debugging.
+        $data = [];
         if (!session_id() && (!headers_sent() || !empty($_COOKIE[ini_get('session.name')]))) {
             if (!headers_sent()) {
```
dnsl48: It looks like `session_start` should be in this block as if the headers have already been sent, we can't start the session and will get a PHP warning alike
```
Warning: session_start(): Cannot start session when headers already sent in ...
```

src/Control/Session.php
```diff
@@ -319,36 +320,38 @@ public function start(HTTPRequest $request)
...
 
-            session_start();
+            $sessionParameters = [
+                "cookie_path" => $path,
+                "cookie_domain" => $domain ?: "",
+                "cookie_secure" => $secure,
+                "cookie_httponly" => true
+            ];
+
+            if ($timeout) {
```
dnsl48: this `if` block feels to be rather redundant. also, for consistency we may want to move `cookie_lifetime` parameter definition into the array with other params. E.g.
```php
$sessionParameters = [
  ...
  "cookie_lifetime" => $timeout ?: "0"
  ...
];
```

src/Control/Session.php
```diff
@@ -319,36 +320,38 @@ public function start(HTTPRequest $request)
...
+                $sessionParameters["cookie_lifetime"] = $timeout;
+            }
+
+            session_start($sessionParameters);
```
dnsl48: I'm not sure if we need to overwrite the cookie pararmeters in here at all, since we already pass them all into `session_set_cookie_params` above in the line 306?

caffeineinc: Yeah, it's true, it's redundant, it's the PHP7 way of setting session parameters, so we can either use the new and old or just retain the old -> `session_set_cookie_params` I didn't initially call this, but it was suggested for backwards compatibility. (5.6)

src/Control/Session.php
caffeineinc: Huh, this got lost in the merge. I'll chuck that back in :)

src/Control/Session.php
caffeineinc: Ok, I think i've resolved this issue too, although I noticed the `$_COOKIE[ini_get('session.name')]` wasn't checking for the secure cookie, so i've amended that by calling the helper function `requestContainsSessionId in the class instead. -  I think that's what you intended @sminnee  ?

dnsl48:
@caffeineinc would you like to make the linter happy? :)

caffeineinc:
haha yeah sure. @dnsl48  There ya go 📦

chillu:
@dnsl48 Please rebase from `4` to `4.3` yourself, to get this over the line. My assumption is that you can push to Simon's PR, otherwise just do it locally

PR #8643, opened 2018-11-28
NEW: Add visibility updates to `upgrade.yml`

bergice:
This will make the inspect upgrader command ensure these visibilities are used

### DIFF
--- a/.upgrade.yml
+++ b/.upgrade.yml
@@ -1314,6 +1314,132 @@ warnings:
       replacement: 'writeJavascriptToBody'
     'SilverStripe\Forms\Formfield->dontEscape':
       message: 'FormField::$dontEscape has been removed. Escaping is now managed on a class by class basis.'
+    'SilverStripe\ORM\DataObject::db':
+      visibility: private
+    'SilverStripe\ORM\DataObject::singular_name':
+      visibility: private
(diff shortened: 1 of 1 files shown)


maxime-rainville:
I'm just thinking here that if you are running an older version of the upgrader that doesn't know about visibility you'll get a boat load of warnings when running.

image

Do you think there's an easy way around this to avoid displaying all those warnings?

@chillu Would you consider this an acceptable down side?

bergice:
> I'm just thinking here that if you are running an older version of the upgrader that doesn't know about visibility you'll get a boat load of warnings when running.
> 
> image
> 
> Do you think there's an easy way around this to avoid displaying all those warnings?
> 
> @chillu Would you consider this an acceptable down side?

Composer require? :D

maxime-rainville:
The problem is that until now, the upgrader always expected the `warning` entries in `.upgrade.yml` to have a `message` key. So if someones gets the updated framework `.upgrade.yml` with the old upgrader installed, they'll get that big list of PHP notice.

chillu:
I don't think this is appropriate for the `warnings` key. Warnings are things we can't fix automatically, but still tell users about. The config stuff *can* be fixed automatically, and just causes upgrader log output like all other rewrites. I'd recommend creating a new toplevel key in `.upgrade.yml`.

PR #8650, opened 2018-11-30
FIX: Manifest should ignore vendor folders within packages contained in vendor

sminnee:
Without this change vendor/silverstripe/framework/vendor/silverstripe/config
will be pick up by the manifest, which is inappropriate.

Although this doesn’t happen often, it can occur if you have run
“composer install” within vendor/silverstripe/framework, which can be
done either accidentally or (in my case) as part of running the
framework tests isolated from the rest of your project (which is closer
to the execution model on Travis)

Note that the presence of the ‘nestedvendor.txt’ file tests that this
works without any explicit changes to the PHP of the tests, since it’s
merely confirming that such a file is *not* picked up.

### DIFF
--- a/src/Core/Manifest/ManifestFileFinder.php
+++ b/src/Core/Manifest/ManifestFileFinder.php
@@ -41,6 +41,11 @@ class ManifestFileFinder extends FileFinder
         // Keep searching inside vendor
         $inVendor = $this->isInsideVendor($basename, $pathname, $depth);
         if ($inVendor) {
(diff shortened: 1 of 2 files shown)


robbieaverill:
I haven't looked into this much at all, but off the top of my head - when you do this you get lots of errors for duplicate class names in the autoloader, etc. If we ignore that in SilverStripe's class manifest, do you think we might be hiding a very visible error in a way that may surface in a less visible and harder to debug way further down the line? I kind of feel like it's a good thing that the errors are obvious when you do this

PR #8658, opened 2018-12-08
MAJOR: improved logic on form scaffolding inclusion

sunnysideup:
1. Has One is a relationship so should only be included if relationships are added
2. other relationships should have the whitelist applied as well (restrictFields)
3. belongsManyMany should also be added.

# Parent issue
* https://github.com/silverstripe/silverstripe-framework/issues/8659

### DIFF
--- a/src/Forms/FormScaffolder.php
+++ b/src/Forms/FormScaffolder.php
@@ -110,7 +110,9 @@ class FormScaffolder
         }
 
         // add has_one relation fields
-        if ($this->obj->hasOne()) {
+        if ($this->obj->hasOne()
+           && ($this->includeRelations === true || ! empty($this->includeRelations['has_one']))
+           ) {
             foreach ($this->obj->hasOne() as $relationship => $component) {
                 if ($this->restrictFields && !in_array($relationship, $this->restrictFields)) {
                     continue;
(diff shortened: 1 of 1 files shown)


src/Forms/FormScaffolder.php
```diff
@@ -110,7 +110,9 @@ public function getFieldList()
         }
 
         // add has_one relation fields
-        if ($this->obj->hasOne()) {
+        if ($this->obj->hasOne()
```
maxime-rainville: There's 4 variations of this condition check. I think it would probably be worthwhile moving it to a small private method. Something like :
```php
private function includeRelation($type) {
    return 
        $this->obj->$type() && 
        ($this->includeRelations === true || !empty($this->includeRelations[$type]);
}
```

maxime-rainville:
@sunnysideup Did you have a chance to have another look at this?

Also, I just noticed you've targeted your PR at the `master` branch, which is for SilverStripe 5. If you want the change to be available in SilverStripe 4, just retarget your PR to the `4` branch.

emteknetnz:
It seems like there's going to be no further activity on this pull request, so we've closed it for now. Please open a new pull-request if you want to re-approach this work, or for anything else you think could be fixed or improved.

sunnysideup:
I will have a look at this in the future.

PR #8661, opened 2018-12-09
FIX escaping blackslashes when using the new attribute store for gridfield actions

ScopeyNZ:
Currently it is possible for action state for gridfield actions to contain backslashes. This makes jQuery (silently) mad when it tries to parse objects from data attributes and in turn results in double escaped JSON objects being sent back server-side

We encountered this issue with userforms which is placing class names in the "action state" for the add button. You can see the issue without this patch by attempting to add fields to a userform in SS4.

### DIFF
--- a/src/Forms/GridField/FormAction/AttributeStore.php
+++ b/src/Forms/GridField/FormAction/AttributeStore.php
@@ -18,7 +18,7 @@ class AttributeStore extends AbstractRequestAwareStore
     {
         // Just save the state in the attributes of the action
         return [
-            'data-action-state' => json_encode($state),
+            'data-action-state' => addcslashes(json_encode($state), '\\'),
         ];
     }
 


ScopeyNZ:
The problem isn't so much with jQuery - it's with correctly escaping backslashes in attributes. In our case we _want_ double slashes in the state when stored as an attribute because it's escaping unicode references.

maxime-rainville:
My main concern is to make sure that whatever goes into the `save` method comes back the same way when the `load` is called.

I would be OK with the `load` method removing the double slashes. But ideally we would address the underlying issue that prevents us from outputting valid JSON in HTML attributes.

ScopeyNZ:
Yeah okay. There's a reasonably wide-impact change that I'd consider a solution to this then... Adding `addcslashes($text, '\\')` to whatever `.ATTR` uses on templates.

maxime-rainville:
Cool. Shall we close this PR for now?

ScopeyNZ:
I'd prefer to keep it open for now...

maxime-rainville:
I'll move it to blocked on the OS board.

emteknetnz:
@ScopeyNZ what would you like to do with this pull-request? Are you going to be doing further work on it or would you prefer to now close this one?

ScopeyNZ:
Well it's fixing a real bug as it stands, but realistically I'm probably not going to address this for years to come. So you can close it. It should probably be converted into an issue though - as `.ATTR` on ArrayData doesn't escape attributes properly.

emteknetnz:
OK thanks.  I've created an issue for this https://github.com/silverstripe/silverstripe-framework/issues/9567 - feel free to modify the issue as you see fit

PR #8668, opened 2018-12-12
FIX: tableList() function slow

smpettit:
Observed behaviour: After upgrading our site to 3.7 our pages were slow to load. It seemed the "SHOW FULL TABLES WHERE Table_Type != 'VIEW'" query was taking 2s+ in production.

What I'm fixing:

Researching why this query is so slow, it seems to relate to MySQL (we're using 5.7) permission checks when the query is written this way. There is an open bug that has gone nowhere: https://bugs.mysql.com/bug.php?id=81865

If I change:

SHOW FULL TABLES WHERE Table_Type != 'VIEW';
to
SELECT table_name, table_type FROM information_schema.tables WHERE table_schema = database();

We get the same result however the latter query takes 0.093s.

### DIFF
--- a/model/connect/MySQLSchemaManager.php
+++ b/model/connect/MySQLSchemaManager.php
@@ -336,7 +336,7 @@ class MySQLSchemaManager extends DBSchemaManager {
 
 	public function tableList() {
 		$tables = array();
-		foreach ($this->query("SHOW FULL TABLES WHERE Table_Type != 'VIEW'") as $record) {
+		foreach ($this->query("SELECT table_name, table_type FROM information_schema.tables WHERE table_schema = database()") as $record) {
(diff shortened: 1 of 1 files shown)


maxime-rainville:
Hi @smpettit, Thank for taking the time to do this PR.

It looks like the [query you are updating is still present in SilverStripe 4](https://github.com/silverstripe/silverstripe-framework/blob/e18549beb08b47f63c87df3f1d727313e380bb95/src/ORM/Connect/MySQLSchemaManager.php#L382). Would you mind re-targeting your PR against the 4 branch?

maxime-rainville:
I'll close this for now. Just post a comment if there's a need for further discussion.

Issue #8644, opened 2018-11-28
Session class docs need upgrade


wernerkrauss:
## Affected Version

4.x

## Description

https://github.com/silverstripe/silverstripe-framework/blob/4d14e9b6b1c7ede1c9a52efd366cafe8d6cc3cdf/src/Control/Session.php#L34-L88 has outdated docs, cause the method's `get()` and `set()` are now class methods that should not be called statically. PHP7.2 errors on this.

# Pull Request
* [x] https://github.com/silverstripe/silverstripe-framework/pull/8671

maxime-rainville:
Just had a quick glance at the code. Looks like the best way to access a session object is now to do it via `HTTPRequest::getSession()`. Can someone second that?

The upgrade guidance in `.upgrade.yml` right now is kind of useless, so we probably should update that as well.

ScopeyNZ:
Yep `$request->getSession` is the way, and it makes sense because the request determines the state of the session (with a session ID cookie parameter).

maxime-rainville:
Added a PR at https://github.com/silverstripe/silverstripe-framework/pull/8671

PR #8671, opened 2018-12-12
DOCS Updating Session doc to reflect that you need to operation on an instance

maxime-rainville:
Simply updates the PHPDoc for the session class to be more inline with the usage in SS4. Some related `.upgrade.yml` changes as well.

There's a few bits in there about how to use session in unit test. I left them in there because I'm sure if they're still true/relevant. If you're more knowledgeable than me about that, just jump in there.

# Parent Issue
* #8644

### DIFF
--- a/.upgrade.yml
+++ b/.upgrade.yml
@@ -1285,9 +1285,9 @@ warnings:
     'days_between()':
       message: 'Removed SilverStripe\ORM\FieldType\DBDate->days_between()'
     'SilverStripe\Control\Session::get_all()':
-      message: 'Session can not be access statically and `get_all()` is now called `getAll()'
+      message: 'Session can not be accessed statically and `get_all()` is now called `getAll()'
     'SilverStripe\Control\Session::clear_all()':
-      message: 'Session can not be access statically and `clear_all()` is now called `clearAll()'
+      message: 'Session can not be accessed statically and `clear_all()` is now called `clearAll()'
   props:
     'class':
       message: '$this->class access has been removed'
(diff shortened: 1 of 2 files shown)


.upgrade.yml
```diff
@@ -1284,6 +1284,10 @@ warnings:
       message: 'Removed SilverStripe\ORM\FieldType\DBDate->day_before()'
     'days_between()':
       message: 'Removed SilverStripe\ORM\FieldType\DBDate->days_between()'
+    'SilverStripe\Control\Session::get_all()':
+      message: 'Session can not be access statically and `get_all()` is now called `getAll()'
```
ScopeyNZ:
```suggestion
      message: 'Session can not be accessed statically and `get_all()` is now called `getAll()'
```

.upgrade.yml
```diff
@@ -1284,6 +1284,10 @@ warnings:
...
+    'SilverStripe\Control\Session::clear_all()':
+      message: 'Session can not be access statically and `clear_all()` is now called `clearAll()'
```
ScopeyNZ:
```suggestion
      message: 'Session can not be accessed statically and `clear_all()` is now called `clearAll()'
```

src/Control/Session.php
```diff
@@ -9,46 +9,43 @@
 /**
  * Handles all manipulation of the session.
  *
- * The static methods are used to manipulate the currently active controller's session.
- * The instance methods are used to manipulate a particular session.  There can be more than one of these created.
+ * An instance of `Session` object can be retrieve via an `HTTPRequest` by calling the `getSession()` method.
```
ScopeyNZ:
```suggestion
 * An instance of a `Session` object can be retrieved via an `HTTPRequest` by calling the `getSession()` method.
```

ScopeyNZ:
A few recommended typo corrections 🙂 Otherwise LGTM.

PR #8681, opened 2018-12-18
FIX: Fixes #8459 Missing conditionals for optional constructor args

### LINKED ISSUE #8459: `Form` define `$fields` and `$actions` as optional but then it throws an error if you don't define them

# Pull Request

phptek:


phptek:
Feature branch came off branch "4" as this issue exists in 4.2 and 4.3 (Possibly elsewhere in the 4.x series)

phptek:
Closing in favour of updated PR.

PR #8683, opened 2018-12-18
FIX: Fixes #8459

### LINKED ISSUE #8459: `Form` define `$fields` and `$actions` as optional but then it throws an error if you don't define them

phptek:
Missing conditionals for optional constructor args

### DIFF
--- a/src/Forms/Form.php
+++ b/src/Forms/Form.php
@@ -286,8 +286,13 @@ class Form extends ViewableData implements HasRequestHandler
     ) {
         parent::__construct();
 
-        $fields->setForm($this);
-        $actions->setForm($this);
+        if ($fields) {
+            $fields->setForm($this);
+        }
+
+        if ($actions) {
+            $actions->setForm($this);
+        }
 
         $this->fields = $fields;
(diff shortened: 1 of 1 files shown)


src/Forms/Form.php
```diff
@@ -1405,9 +1413,10 @@ public function loadDataFrom($data, $mergeStrategy = 0, $fieldList = null)
             $submitted = true;
         }
 
-        // dont include fields without data
-        $dataFields = $this->Fields()->dataFields();
-        if (!$dataFields) {
+        // Don't include fields without data
+        $fields = $this->Fields();
+
+        if (!$fields || !$dataFields = $fields->dataFields()) {
```
robbieaverill: I guess the variable assignment isn't required any more...

phptek: Mate, it was late. I'm cannng this one, and re-opening a new (the original) one. I'm conflating too many different issues in the one PR. Sorry.

maxime-rainville:
@phptek Did you mean to close this PR?

PR #8684, opened 2018-12-18
FIX: Fixes #8459 Missing conditionals for optional constructor args

### LINKED ISSUE #8459: `Form` define `$fields` and `$actions` as optional but then it throws an error if you don't define them

phptek:
Fixes #8459
- Missing conditionals for optional constructor args
- Missing calls to `FieldList::setForm()`
- Missing guards around naive calls to `Form::Fields()->foo()`

### DIFF
--- a/src/Forms/Form.php
+++ b/src/Forms/Form.php
@@ -286,6 +286,9 @@ class Form extends ViewableData implements HasRequestHandler
     ) {
         parent::__construct();
 
+        $fields = $fields ? $fields : FieldList::create();
+        $actions = $actions ? $actions : FieldList::create();
+
         $fields->setForm($this);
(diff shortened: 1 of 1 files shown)


kinglozzer:
Would it be easier, rather than scatter around extra null conditional checks, to just add something to the constructor like:

```php
$fields = $fields ? $fields : FieldList::create();
$actions = $actions ? $actions : FieldList::create();
```

maxime-rainville:
@kinglozzer's suggestion makes more sense.

phptek:
Fixed. Force-pushed. Please excuse the unrelated changes, I need newlines to be there and typos not to be.

PR #8685, opened 2018-12-20
BUG: Fix switching to errors inside nested tabs.

patricknelson:
Currently, if an error occurs in a tab other than the main tab, the tab will be switched so the user can see the error. However, this will not work if the error is present on a field that is nested beyond the first level. This will address that problem and more, specifically:

1. If an error occurs at any tab depth, it will be revealed.
2. If there are **multiple** errors on **different** tabs, the user will be notified so that they're aware of the presence of multiple errors to ensure they don't have to waste time on a round-trip form submission only to reveal the next error in a series of errors.
    -  **Note:** This changes prior functionality where the error dialog (in the top right) was only rendered if no tabs could be found.
3. In the case of multiple errors nested in different tabs, the errors will be immediately visible with **minimal** clicks when the user progresses to the next top level tab with errors. 

**Other notes:**
- This relates to previous work done on issue https://github.com/silverstripe/silverstripe-framework/issues/5410#issuecomment-215303233 (point about nested errors) and PR #5412
- This bug exists both in SS 3.x and SS 4.x. The same fix will work unmodified (after switching to using jQuery UI tab CSS selectors).

### Demo 1: Automatically selects the correct tabs at-depth.

auto-select depth

### Demo 2 + 3: Multi-error notification + auto selecting nested tabs at-depth (for secondary error[s]).

muliple errors auto-select depth

**TODO:**
- [x] Show gifs/screenshots with examples so this PR is easier to understand 😄 
- [ ] PR for SS 4.x in `silverstripe-admin` in `LeftAndMain.EditForm.js` here: https://github.com/silverstripe/silverstripe-admin/blob/1/client/src/legacy/LeftAndMain.EditForm.js#L112

### DIFF
--- a/admin/javascript/LeftAndMain.EditForm.js
+++ b/admin/javascript/LeftAndMain.EditForm.js
@@ -99,27 +99,54 @@
 			},
 			'from .cms-tabset': {
 				onafterredrawtabs: function () {
-					// Show validation errors if necessary
-					if(this.hasClass('validationerror')) {
-						// Ensure the first validation error is visible
-						var tabError = this.find('.message.validation, .message.required').first().closest('.tab');
-						$('.cms-container').clearCurrentTabState(); // clear state to avoid override later on
+					// Show validation errors if necessary (and if not already done).
+					if(this.hasClass('validationerror') && !this.getValidationErrorShown()) {
+						this.setValidationErrorShown(true);
+						var $allErrors = this.find('.message.validation, .message.required');
 
-						// Attempt #1: Look for nearest .ss-tabset (usually nested deeper underneath a .cms-tabset).
-						var $tabSet = tabError.closest('.ss-tabset');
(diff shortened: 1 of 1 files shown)


patricknelson:
@tractorcow, is 3.7 the right branch for bug fixes still?

maxime-rainville:

I'll have a look at the SS4 PR.

patricknelson:
Working on an upgrade now (yes, 8 years coming) and I'm wondering: Was this ever really addressed in SS v4+? i.e. Does SSv4 switch to the tab to show the field with the active error after form submission?

patricknelson:
Note mainly to self: https://github.com/silverstripe/silverstripe-admin/pull/1111 appears to help remedy this in a superior way (i.e. flagging the actual tabs that contain errors with a ❗), it unfortunately doesn't propagate that to grandparent tabs, such as when the error is deeply nested beyond the first level.

PR #8689, opened 2018-12-31
Fix for CheckboxSetField not disabling items correctly

jinjie:
Adds a new variable `$disabledItems` to get disabled items from `getDisabledItems()`

Bug was caused by checking $itemValue against $defaultItems to check for disabled items.


jinjie:
This is my first pull request for silverstripe. Not sure if I'm doing it correctly. Hope this PR gets through.

PR #8692, opened 2019-01-04
do not assume anything about arriving messages

oetiker:
do not assume that messages arriving from an iframe contain 'nice' data

this prevents LeftAndMain from dying when 'odd' messages arrive at the handler.


kinglozzer:
Hi @oetiker, thanks for the pull request. Unfortunately, [SilverStripe 3 entered limited support in June 2018](https://www.silverstripe.org/blog/update-on-silverstripe-5-x/). This means we'll only be fixing critical bugs and security issues for SilverStripe 3 going forward.

Issue #8696, opened 2019-01-07
Default scaffolded field titles should only uppercase the first letter


robbieaverill:
## Affected Version

* 4.3.0

## Description

If I have a DB field called `MyFieldName` then the scaffolded field title will be "My Field Name". It should really be "My field name" (only uppercasing the first letter).

cc @sachajudd @clarkepaul

kinglozzer:
Yes!

<img width="210" alt="screenshot 2019-01-07 at 15 54 32" src="https://user-images.githubusercontent.com/1655548/50777887-a1c4cd80-1294-11e9-8715-6e8b3d60f634.png">

clarkepaul:
Yes please and now the rest of the CMS ;)

PR #8698, opened 2019-01-07
NEW Scaffolded field labels now only have an uppercased first word

robbieaverill:
See the unit test for examples of how this now behaves.

Resolves https://github.com/silverstripe/silverstripe-framework/issues/8696

### DIFF
--- a/tests/php/Forms/FormFieldTest.php
+++ b/tests/php/Forms/FormFieldTest.php
@@ -402,4 +402,31 @@ class FormFieldTest extends SapphireTest
         $field = new FormField('Test');
         $field->Link('bar');
     }
+
+    /**
+     * @param string $name
+     * @param string $expected
+     * @dataProvider nameToLabelProvider
+     */
+    public function testNameToLabel($name, $expected)
+    {
+        $this->assertSame($expected, FormField::name_to_label($name));
+    }
+
+    /**
+     * @return array[]
+     */
+    public function nameToLabelProvider()
+    {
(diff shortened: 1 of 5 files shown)


robbieaverill:
Uh oh, I broke things!

tests/php/Forms/FormFieldTest.php
```diff
@@ -402,4 +402,30 @@ public function testLinkWithoutForm()
...
+     * @return array[]
+     */
+    public function nameToLabelProvider()
+    {
+        return [
+            ['TotalAmount', 'Total amount'],
+            ['Organisation.ZipCode', 'Organisation zip code'],
+            ['Organisation.zipCode', 'Organisation zip code'],
+            ['FooBarBaz', 'Foo bar baz'],
+            ['URLSegment', 'URL segment'],
```
kinglozzer: ❤️

robbieaverill: The `testSchemaValidation` error is real - `SecurityID` should become `Security ID` - yay for tests!

robbieaverill:
Ok I've updated the tests that had labels like "My Field Name" in them to "My field name", fixed a couple of assertion argument orders and added handling for field names like `SecurityID` or `SpecialURL` to still be converted to `Security ID` and `Special URL` - rebased and pushed

robbieaverill:
I think that Scrutinizer failure is unrelated

PR #8700, opened 2019-01-08
DOCS clarified release EOL process

chillu:
Important that we remember this if we ever want to stop supporting a dozen active branches ;)


robbieaverill:
I think we should merge this into the 4 branch. I'll merge by hand

PR #8701, opened 2019-01-08
Update Travis to use PHP 7.3 stable

robbieaverill:


robbieaverill:
Self merging since it's a low impact Travis change. I wanted to ensure the builds passed before pushing it directly to origin though.

PR #8707, opened 2019-01-09
Check array keys existence prior to their usage when removing methods

michalkleiner:
The updated piece of code throws a Notice when trying to use the failover mechanism with a custom object other than the standard SiteTree via a ContentController.


kinglozzer:
This looks very similar to https://github.com/silverstripe/silverstripe-framework/pull/8703

michalkleiner:
I think in our case it would get past that added condition and still fail on checking unset array keys.
Also, our problem is with 4.2.1, perhaps I should retarget the PR to 4.2?

robbieaverill:
`SilverStripe\View\Tests\ViewableDataTest::testSetFailover` is failing, looks relevant. Thoughts @michalkleiner?

lhalaa:
Any updates on this? We had a similar issue when using failover. 
```
// CustomMethods
$methodInfo = self::$extra_methods[$class][$method];
if ($methodInfo['property'] === $property && $methodInfo['index'] === $index) {
    unset(self::$extra_methods[$class][$method]);
}
```
throws a notice
`[Notice] Undefined index: property`
where the returned `$methodInfo` is 
```
array (
  'callback' => 
  Closure::__set_state(array(
  )),
)
```

Basically, what we're trying to do is render a UDF page which has a unique generated URL like shared draft, but for forms; and set the dataRecord as if creating a new controller instance, and setting failover for the that particular record.
Either we're using the `setFailover` incorrectly or that this code just needs some checking for the existence of an array key. Appreciate to hear some suggestions.

michalkleiner:
I'll need to get my head around it once again and see what the failure means and what the test is testing to assess what needs changing where.

robbieaverill:
I think this PR needs a rebase

michalkleiner:
Will look into it

michalkleiner:
For a couple minutes I was super confused why there are no changes, but a newer PR #9271 has already been merged.

PR #8709, opened 2019-01-10
Add documentation for additional file types

jinjie:
Documentation for silverstripe/silverstripe-assets#203

# Parent issue
* https://github.com/silverstripe/silverstripe-assets/issues/202


jinjie:
> This probably should be its own top level section. It's pretty far from being the most important topic on this page so I would move it down just before _Related Lessons_.

If `getMimeType()` is already in this section, shouldn't `getFileType()` also ends up in the same section?

Like you said, this is not of high importance, don't think it needs it's own section. 🤔

maxime-rainville:
My thinking was that the section its been put in is more of a generic list of method, while the bit you added is more about configuration. However, that's one of my many pet peeve.

Merging now.

PR #8717, opened 2019-01-10
BUG Fix issue with assertListEquals() ignoring field getters

tractorcow:
This is mostly an issue with DataObjects, since in order to respect both getField() and getSomeField(), you need to access it via __get(). I've created a test class to replicate this.

The issue I came across was where I had a non-db field getter that I couldn't test with assertListContains().


robbieaverill:
Hmm, maybe it's not an unrelated test failure... what do you think @tractorcow?

tractorcow:
It's because I made the PR to 4 branch not 4.2 by accident. I've restarted the tests.

PR #8726, opened 2019-01-15
MINOR Update add_i18n_javascript to load lowercase-dash language files

maxime-rainville:
We need a way to load MomentJS local files because of this issue https://github.com/silverstripe/silverstripe-admin/pull/791.

MomentJS file formats are like `de-de`, while our current format is `de_DE.js`.

This update just add a few extra format to check for.


src/View/Requirements_Backend.php
```diff
@@ -1029,6 +1038,17 @@ public function add_i18n_javascript($langDir, $return = false)
         return null;
     }
 
+    /**
```
maxime-rainville: I'll remove that.

src/View/Requirements_Backend.php
```diff
@@ -1004,13 +1005,21 @@ public function add_i18n_javascript($langDir, $return = false)
...
+            i18n::getData()->langFromLocale(i18n::config()->get('default_locale')),
+            i18n::config()->get('default_locale'),
+            i18n::getData()->langFromLocale(i18n::get_locale()),
+            i18n::get_locale(),
+            strtolower(DBField::create_field('Locale', i18n::get_locale())->RFC1766()),
+            strtolower(DBField::create_field('Locale', i18n::config()->get('default_locale'))->RFC1766())
         );
+
+        $candidates = array_map(
+            function ($candiate) { return $candiate . '.js'; },
```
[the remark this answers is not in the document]
maxime-rainville: Fixed.

maxime-rainville:
Unit tests have been added and typo has been corrected.

src/View/Requirements_Backend.php
```diff
@@ -1004,13 +1005,23 @@ public function add_i18n_javascript($langDir, $return = false)
...
-            i18n::config()->get('default_locale') . '.js',
-            i18n::getData()->langFromLocale(i18n::get_locale()) . '.js',
-            i18n::get_locale() . '.js',
+            'en',
+            'en_US',
+            i18n::getData()->langFromLocale(i18n::config()->get('default_locale')),
+            i18n::config()->get('default_locale'),
+            i18n::getData()->langFromLocale(i18n::get_locale()),
+            i18n::get_locale(),
+            strtolower(DBField::create_field('Locale', i18n::get_locale())->RFC1766()),
```
bergice: Should `strtolower` be encapsulated inside `RFC1766`?

tests/php/View/RequirementsTest.php
```diff
@@ -1109,4 +1110,77 @@ protected function getBackendFiles($backend, $type)
...
+        $backend = Injector::inst()->create(Requirements_Backend::class);
+        $this->setupRequirements($backend);
+        $backend->add_i18n_javascript('i18n');
+
+        $actual = $backend->getJavascript();
+
+        // English and English US should always be loaded no matter what
+        $this->assertArrayHasKey('i18n/en.js', $actual);
+        $this->assertArrayHasKey('i18n/en_US.js', $actual);
+        $this->assertArrayHasKey('i18n/en-us.js', $actual);
```
bergice: Why is there 2 of these?

tests/php/View/RequirementsTest.php
```diff
@@ -1109,4 +1110,77 @@ protected function getBackendFiles($backend, $type)
...
+    {
+        i18n::config()->set('default_locale', 'fr_CA');
+
+        /** @var Requirements_Backend $backend */
+        $backend = Injector::inst()->create(Requirements_Backend::class);
+        $this->setupRequirements($backend);
+        $backend->add_i18n_javascript('i18n');
+
+        $actual = $backend->getJavascript();
+
```
bergice: Please remove duplicate blank lines.

tests/php/View/RequirementsTest.php
```diff
@@ -1109,4 +1110,77 @@ protected function getBackendFiles($backend, $type)
         }
         return array();
     }
+
```
bergice: It might pay to check that we don't load locales that are irrelevant.

src/View/Requirements_Backend.php
maxime-rainville: It's not clear. On the [W3C site](https://www.w3.org/International/articles/language-tags/) they code the region in upper-case, but Microsoft doesn't in their doc (https://docs.microsoft.com/en-us/windows/desktop/wmformat/language-strings).

The region bit should confirm to the RFC3066 standard. The IETF site says it doesn't matter. https://tools.ietf.org/html/rfc3066#section-2.1

In any case, changing to lower case everywhere could break other things, so we're probably best not to touch it right now.

tests/php/View/RequirementsTest.php
maxime-rainville: Are regular translations are in the `en_US.js` format. MomentJS Locale files use the `en-us.js` format.

PR #8728, opened 2019-01-15
FIX: Fixes #4376 partially added missing doc blocks.

### LINKED ISSUE #4376: update DataExtension class to match all extension points
There are some undocumented extensions points in DataObject that are quite unknown, cause DataExtension class doesn't reflect it. E.g. `updateSearchableFields()`  (see http://api.silverstripe.org/3.1/source-class-DataObject.html#3260)

TylahBucktonPereira:
added doc blocks

### DIFF
--- a/src/ORM/DataExtension.php
+++ b/src/ORM/DataExtension.php
@@ -60,22 +60,52 @@ abstract class DataExtension extends Extension
     {
     }
 
+    /**
+     * Optional extension hook called by SilverStripe directly before a database write operation.
+     * Any logic expressed here, will be executed by SilverStripe.
+     *
+     * @return void
+     */
     public function onBeforeWrite()
     {
     }
 
+    /**
+     * Optional extension hook called by SilverStripe directly after a database write operation.
+     * Any logic expressed here, will be executed by SilverStripe.
+     *
+     * @return void
+     */
     public function onAfterWrite()
     {
     }
 
+    /**
+     * Optional extension hook called by SilverStripe directly before deleting a database write operation.
+     * Any logic expressed here, will be executed by SilverStripe.
+     *
+     * @return void
+     */
     public function onBeforeDelete()
     {
     }
 
+    /**
+     * Optional extension hook called by SilverStripe directly after deleting a database write operation.
+     * Any logic expressed here, will be executed by SilverStripe.
+     *
+     * @return void
+     */
     public function onAfterDelete()
     {
     }
 
+    /**
+     * Optional extension hook called by SilverStripe directly during a dev/build operation.
+     * Any logic expressed here, will be executed by SilverStripe.
+     *
+     * @return void
+     */
     public function requireDefaultRecords()
     {
     }
@@ -84,18 +114,37 @@ abstract class DataExtension extends Extension
     {
     }
 
+
     public function can($member)
     {
     }
 
+    /**
+     * Whether or not the given $member is able to edit this decorated object.
+     *
+     * @param  Member $member
+     * @return bool
+     */
     public function canEdit($member)
     {
     }
 
+    /**
+     * Whether or not the given $member is able to delete this decorated object.
+     *
+     * @param  Member $member
+     * @return bool
+     */
     public function canDelete($member)
     {
     }
 
+    /**
+     * Whether or not the given $member is able to create this decorated object.
+     *
+     * @param  Member $member
+     * @return bool
+     */
     public function canCreate($member)
     {
     }


src/ORM/DataExtension.php
```diff
@@ -60,22 +60,52 @@ public function augmentWrite(&$manipulation)
...
+     * Any logic expressed here, will be executed by SilverStripe.
+     *
+     * @return void
+     */
     public function onAfterWrite()
     {
     }
 
+    /**
+     * Optional extension hook called by SilverStripe directly before deleting a database write operation.
```
robbieaverill: >directly before deleting a database write operation

I think this might be a result of a copy/paste, would you mind updating it to say something like "directly before deleting a DataObject" or something like that?

src/ORM/DataExtension.php
```diff
@@ -60,22 +60,52 @@ public function augmentWrite(&$manipulation)
     {
     }
 
+    /**
+     * Optional extension hook called by SilverStripe directly before a database write operation.
+     * Any logic expressed here, will be executed by SilverStripe.
```
robbieaverill: I'm not sure we need this sentence to be honest

src/ORM/DataExtension.php
```diff
@@ -60,22 +60,52 @@ public function augmentWrite(&$manipulation)
...
+     *
+     * @return void
+     */
     public function onBeforeDelete()
     {
     }
 
+    /**
+     * Optional extension hook called by SilverStripe directly after deleting a database write operation.
```
robbieaverill: As with onBeforeDelete, I think the end of this sentence could be adjusted slightly

src/ORM/DataExtension.php
```diff
@@ -60,22 +60,52 @@ public function augmentWrite(&$manipulation)
...
+     *
+     * @return void
+     */
     public function onAfterDelete()
     {
     }
 
+    /**
+     * Optional extension hook called by SilverStripe directly during a dev/build operation.
+     * Any logic expressed here, will be executed by SilverStripe.
```
robbieaverill: Here's the doc block from `DataObject::requireDefaultRecords()`:

```php
    /**
     * Add default records to database. This function is called whenever the
     * database is built, after the database tables have all been created. Overload
     * this to add default records when the database is built, but make sure you
     * call parent::requireDefaultRecords().
     *
     * @uses DataExtension->requireDefaultRecords()
     */
```

Perhaps you could use that and adjust it to communicate that it's run in an extension as opposed to directly on the object?

src/ORM/DataExtension.php
```diff
@@ -84,18 +114,37 @@ public function populateDefaults()
     {
     }
 
+
```
robbieaverill:
```suggestion
    /**
     * Returns true if the member is allowed to do the given action.
     * See {@link extendedCan()} for a more versatile tri-state permission control.
     *
     * @param Member @member
     */
```

src/ORM/DataExtension.php
```diff
@@ -84,18 +114,37 @@ public function populateDefaults()
...
     public function can($member)
     {
     }
 
+    /**
+     * Whether or not the given $member is able to edit this decorated object.
+     *
+     * @param  Member $member
```
robbieaverill: We need to import the namespace for Member as well, up the top of the file: `use SilverStripe\Security\Member;`

src/ORM/DataExtension.php
```diff
@@ -84,18 +114,37 @@ public function populateDefaults()
...
+     */
     public function canEdit($member)
     {
     }
 
+    /**
+     * Whether or not the given $member is able to delete this decorated object.
+     *
+     * @param  Member $member
+     * @return bool
```
robbieaverill: Technically these methods return `@return int|bool` (see `Permission::check()`). The DataObject equivalents of these methods don't indicate that though, which is why it's a bit misleading

src/ORM/DataExtension.php
```diff
@@ -84,18 +114,37 @@ public function populateDefaults()
...
+     */
     public function canEdit($member)
     {
     }
 
+    /**
+     * Whether or not the given $member is able to delete this decorated object.
```
robbieaverill:
```suggestion
     * Whether or not the given {@link Member} is able to delete the decorated object.
```

src/ORM/DataExtension.php
```diff
@@ -84,18 +114,37 @@ public function populateDefaults()
     {
     }
 
+
     public function can($member)
     {
     }
 
+    /**
+     * Whether or not the given $member is able to edit this decorated object.
```
robbieaverill:
```suggestion
     * Whether or not the given {@link Member} is able to edit the decorated object.
```

src/ORM/DataExtension.php
```diff
@@ -84,18 +114,37 @@ public function populateDefaults()
...
+     *
+     * @param  Member $member
+     * @return bool
+     */
     public function canDelete($member)
     {
     }
 
+    /**
+     * Whether or not the given $member is able to create this decorated object.
```
robbieaverill:
```suggestion
     * Whether or not the given {@link Member} is able to create the decorated object.
```

PR #8729, opened 2019-01-16
Ensure document is not falsey before attempting to clone it

emteknetnz:
If !$this->valid, then getDocument() will return false.  

$doc = clone false; will throw a php error

Can get into this state if a user puts invalid html into a userforms EditableLiteralField e.g. '<p<Some text'


ScopeyNZ:
Thanks for the PR @emteknetnz . Do you want to target this patch against 4.3 (so that it might be released in 4.3.1)?

PR #8732, opened 2019-01-16
Update index.md

TylahBucktonPereira:
formatting


robbieaverill:
Hi @TylahBucktonPereira, thanks for taking the time to make this pull request.

The `title: ` and `description: ` syntax is a kind of metadata that we use for our SilverStripe documentation. While it's not pure markdown, the docs.silverstripe.org parser detects it and uses it for various parts of the page. In this example you see that they're displayed as a page title instead of in the markdown area itself:

image

Changing this would probably be detrimental to the page layout when rendered in the documentation site, so I'm going to close this, but again thank you for contributing!

Issue #8664, opened 2018-12-10
[ASSETS] MigrateFileTask runs out of memory 


torleif:
## Affected Version

Framework / assets 4.3.0-RC1

## Description

I have a site with gigabytes of files, and running ```vendor/silverstripe/framework/sake dev/tasks/MigrateFileTask``` causes out of memory errors. Running it returns errors such as ```Fatal Error (E_ERROR): Allowed memory size of 4487905280 bytes exhausted (tried to allocate 28800001 bytes)```

I've upped the memory limit to 4 gigs to try and get through the task.

It seems it is progressing each time it's run, as the byte allocation is different.

Ideally there wouldn't be a memory leak so the entire task can run without running out of memory.

## Steps to Reproduce

Run ```vendor/silverstripe/framework/sake dev/tasks/MigrateFileTask ``` with a large asset base

chillu:
We actually have an issue to determine rough maximums for this task: https://github.com/silverstripeltd/open-sourcerers/issues/45

Out of interest, do you have very large images (>4000px dimensions) in your site? How many files do you have in total, and how many are processed before it runs out of memory? Just trying to narrow down the use case.  4GB should definitely be enough to run this, I think 512MB is a reasonable max that should allow most environments to run this task.

torleif:
There's 23.0GB of files, and 197,667 files in total. I'm imagining there's a few large images over 4000px in there. 

It seems to fault on the second part of the task, generating thumbnails. it gets to about 68k in the list of files.

wernerkrauss:
@torleif could this be related to https://github.com/silverstripe/silverstripe-framework/issues/8570 ? Dunno if that patch is available in 4.3RC-1 or was merged later.

What if you increase your memory limit manually in `FileMigrationHelper::run()`? 

```
ini_set('memory_limit', -1); 
```

helped for me in SS4.2.

torleif:
I'm overwrote the memory limit in the Envornment file, to see if it would work with that. It looks like it was setting the limit to the default PHP amount. 

after running that I'm getting a new error: 

```
VirtualAlloc() failed: [0x000005af] The paging file is too small for this operation to complete.


VirtualAlloc() failed: [0x000005af] The paging file is too small for this operation to complete.


Fatal error: Out of memory (allocated 16114515968) (tried to allocate 24576 bytes) in C:\Users\torleifw\op\vendor\intervention\image\src\Intervention\Image\Gd\Decoder.php on line 136

[...]

PR #8733, opened 2019-01-17
NEW Add new execmetric debur URL parameter to print out exection time and peak memory usage

maxime-rainville:
This is a quick something I put together to help debug https://github.com/silverstripe/silverstripe-framework/issues/8664

Basically, if you add `execmetrics` as a URL Get parameter, you'll get a print out of how long the request took and the peak memory usage.

Figure it might be helpful going forward.

### DIFF
--- a/src/Control/Middleware/ExecMetricMiddleware.php
+++ b/src/Control/Middleware/ExecMetricMiddleware.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace SilverStripe\Control\Middleware;
+
+use SilverStripe\Assets\File;
+use SilverStripe\Control\Director;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
+use SilverStripe\Dev\Debug;
+
+/**
+ * Secures requests by only allowing a whitelist of Host values
+ */
+class ExecMetricMiddleware implements HTTPMiddleware
+{
+    /**
+     * @inheritdoc
+     */
+    public function process(HTTPRequest $request, callable $delegate)
+    {
+        if (!$this->showMetric($request)) {
+            return $delegate($request);
+        }
+
+        $start = microtime(true);
+        try {
+            return $delegate($request);
+        } finally {
+            $end = microtime(true);
+            Debug::message(
+                sprintf(
+                    "Execution time: %s, Peak memory usage: %s\n",
+                    $this->formatExecutionTime($start, $end),
+                    $this->formatPeakMemoryUsage()
+                ),
+                false
+            );
+        }
+    }
+
+    private function showMetric(HTTPRequest $request)
+    {
+        return Director::isDev() && array_key_exists('execmetric', $request->getVars());
+    }
+
+    /**
+     * Convert the provided start and end time to a interval in secs.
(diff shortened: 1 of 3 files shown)


maxime-rainville:
I witch to using middlewares as suggested by @ScopeyNZ. Seems to be working pretty well.

A minor concern is the order in which the middle ware get added will potentially impact the actual memory usage. There's not really a way of giving those a priority right now. Probably not a big deal, because you would hope your middlewares are pretty light weight any way.

src/Control/Middleware/ExecMetricMiddleware.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        $diff = round($end - $start, 4);
+        return $diff . ' seconds';
+    }
+
+    /**
+     * Get the peak memory usage formatted has a string and a meaningful unit.
+     * @return string
+     */
+    private function formatPeakMemoryUsage()
```
robbieaverill: We prefer protected over private

src/Control/Middleware/ExecMetricMiddleware.php
```diff
@@ -0,0 +1,67 @@
...
+use SilverStripe\Dev\Debug;
+
+/**
+ * Secures requests by only allowing a whitelist of Host values
+ */
+class ExecMetricMiddleware implements HTTPMiddleware
+{
+    /**
+     * @inheritdoc
+     */
```
robbieaverill: This is equivalent to not having a doc block at all

src/Control/Middleware/ExecMetricMiddleware.php
maxime-rainville: Those 3 methods are implementations details and are not meant to be reuse. In theory, I could have implemented all of those directly in `process`, but I wanted to split them out for readability. In this context, I don't think it's worthwhile increasing our public API surface.

src/Control/Middleware/ExecMetricMiddleware.php
```diff
@@ -0,0 +1,67 @@
...
+namespace SilverStripe\Control\Middleware;
+
+use SilverStripe\Assets\File;
+use SilverStripe\Control\Director;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
+use SilverStripe\Dev\Debug;
+
+/**
+ * Secures requests by only allowing a whitelist of Host values
```
ScopeyNZ: Is that really what this middleware does?

ScopeyNZ: Oh whoops that's a retrospective review. Explains why I didn't see all the options when submitting it.

maxime-rainville: :man_facepalming: I copied over one of the other files. Forgot to update the comment.

maxime-rainville: I'll do a follow up PR.

maxime-rainville: https://github.com/silverstripe/silverstripe-framework/pull/8770

Issue #8724, opened 2019-01-14
4.3.1 regression / Session timeout


caffeineinc:
## Affected Version
4.3

## Description
session_start() creates a session or resumes the current one based on a session identifier passed with a request. Duplicate Session Cookies · Issue #8543 highlighted that two headers were emitted during the session start and the second header was removed. This means that sessions no longer refreshed the cookie expiry. 

The regression was debated in the previous issue, but subsequent requests _should_ respond with an updated cookie of the same name so the session can be persisted for the cookie lifetime _after_ the last request. Otherwise the session will only be valid from the point of the initial request or session start, and expire regardless of subsequent requests. 

## Steps to Reproduce

* Load a page which uses session e.g the CSRF token on  `/Security/login` 
* Note the headers returned using `curl -svok /dev/null http://localhost:8000/Security/login` 
* Including a session token on the next request `curl -svok /dev/null  http://localhost:8000/Security/login` --cookie "PHPSESSID=<yourtokenhere>" 
* The second request should have a new cookie with the current GMT + cookie expiry, if it's not present, then the cookie token on the first request will time out. 

## PRs
 - [x] #8736

caffeineinc:
Testing resolution: https://github.com/caffeineinc/silverstripe-framework/blob/bugfix/4.2.1-duplicate-session-headers-8543/src/Control/Session.php#L334

ScopeyNZ:
I discussed this with @caffeineinc but I'm confused as to what exactly is preventing PHP from adding a cooking once the session is started. I thought that was default behaviour in PHP and I can't seem to find any specific code that's preventing it.

I'd love somebody to point it out to me.

caffeineinc:
I looked into Laravel to see how they handled similar core behaviour, and they actually roll their own, emitting cookies rather than relying on native PHP code `session_start()`... it's really bizarre

dnsl48:
> Otherwise the session will only be valid from the point of the initial request or session start, and expire regardless of subsequent requests.

I think that's the default PHP way to implement short living sessions, which forces reauthentication once in a while.

It looks like we switched to it without acknowledging, which may be a breakage of backwards compatibility between 4.3 and 4.3.1 and we might want to address it shortly as the bugfix window closes this week for 4.3.1 release (@chillu)

In general I guess that's 2 different features to consider : 
 - `fixed time living sessions` (accidentally switched to this since 4.3.1)
[...]

PR #8736, opened 2019-01-17
Existing sessions need to set a new cookie on each request, if the

caffeineinc:
Existing sessions need to set a new cookie on each request, if the
session exists, otherwise our expiry is never updated and sessions
can't roll on every request.

This resolves the regression that the existing behaviour: remember sessions for inactive users which was in place before 4.3.1 and https://github.com/silverstripe/silverstripe-framework/issues/8724#issuecomment-454586145

See issue #8724

### DIFF
--- a/src/Control/Session.php
+++ b/src/Control/Session.php
@@ -325,9 +325,9 @@ class Session
 
                 session_start();
 
-                // Session start emits a cookie, but only if there's no existing session. If there is a session tied to
-                // this request, make sure the session is held for the entire timeout by refreshing the cookie age.
-                if ($this->requestContainsSessionId($request)) {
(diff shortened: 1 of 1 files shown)


src/Control/Session.php
```diff
@@ -324,6 +324,12 @@ public function start(HTTPRequest $request)
                 }
 
                 session_start();
+
+                // Session start emits a cookie, but only if there's no existing session. If there is a session tied to
+                // this request, make sure the session is held for the entire timeout by refreshing the cookie age.
+                if ($this->requestContainsSessionId($request)) {
```
dnsl48: We don't really have to send it here if the `$timeout` is undefined or `0`

caffeineinc: Good catch, have amended.

PR #8743, opened 2019-01-22
fix user feedback when jquery could not be loaded

BenediktSeidl:
### expected 
if code.jquery.com can not reached, display error message and execute fallback

### observed behaviour
fallback can not be executed because of typo


ScopeyNZ:
Hey @BenediktSeidl thanks for the PR. This is a patch fix for a bug that seems to affect version 4 since its inception. Would you mind targeting this fix at the `4.0` branch?

You can do this by changing the target branch of this PR (you will probably have to `git rebase`) or closing the PR and re-raising it with the patch made on the correct branch.

ScopeyNZ:
Thanks @BenediktSeidl!

PR #8747, opened 2019-01-23
FIX: Injector may instantiate prototypes as if they're singletons (fixes #8567)

### LINKED ISSUE #8567: BUG: defaults are not inherited in SS3
## Affected Version

3.7.1

## Description

Basically we have this:

```php
ParentClass extends DataObject
{
    private static $db = [
        'FieldA' => 'Boolean'
    ];
    private static $defaults = [
         'FieldA' = 1;
    ];

}
```
and
```php

class OtherClass extends ParentClass
{

}
```
When we create an object of the  `OtherClass` type then FieldA is set to FALSE. We would expect this to be TRUE because that is the default.  We tried both "true" and "1" as default values. 

From what I know about Silverstripe this is not correct AND worked differently before AND if correct, has a massive impact.

kinglozzer:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/8567.

This is a bit of a tricky one to explain, I’ve done my best to simplify it below. Currently usercode and unit tests exhibit different behaviour, because Injector’s `$specs` cache is warmed by the test database scaffolding.

---

There’s a mismatch between how Injector behaves if a `type` (`prototype` or `singleton`) isn’t set for a service, based on whether the `$this->specs[$service]` cache is empty or not. If the spec cache is set (without a `type` key) Injector [assumes `prototype`](https://github.com/silverstripe/silverstripe-framework/blob/3.6.6/control/injector/Injector.php#L831), but if the cache isn’t set it [doesn’t make any assumption](https://github.com/silverstripe/silverstripe-framework/blob/3.6.6/control/injector/Injector.php#L864).

```php
// Current behaviour
MyDataObject::create(); // Treated as a singleton - populateDefaults() not called

// Current behaviour (unit tests call singleton() during database scaffolding)
MyDataObject::singleton(); // Warms the service cache
MyDataObject::create(); // Treated as a prototype - populateDefaults() is called

// Expected behaviour
MyDataObject::create(); // *Always* treated as a prototype
```

### DIFF
--- a/tests/model/DataObjectTest.php
+++ b/tests/model/DataObjectTest.php
@@ -29,6 +29,7 @@ class DataObjectTest extends SapphireTest {
 		'DataObjectTest_Play',
 		'DataObjectTest_Ploy',
 		'DataObjectTest_Bogey',
+		'DataObjectTest_Sortable',
 		'ManyManyListTest_Product',
 		'ManyManyListTest_Category',
 	);
@@ -38,6 +39,13 @@ class DataObjectTest extends SapphireTest {
 	 */
 	public function testSingleton($inst, $defaultValue, $altDefaultValue)
 	{
+		// Calls to scaffold the test database may have cached service specs for DataObjects
+		// with the incorrect 'type' set (singleton instead of prototype)
+		Injector::nest();
+		$reflectionProp = new ReflectionProperty('Injector', 'specs');
+		$reflectionProp->setAccessible(true);
+		$reflectionProp->setValue(Injector::inst(), array());
+
 		$inst = $inst();
 		// Test that populateDefaults() isn't called on singletons
 		// which can lead to SQL errors during build, and endless loops
@@ -52,6 +60,8 @@ class DataObjectTest extends SapphireTest {
 		} else {
 			$this->assertEmpty($inst->MyFieldWithAltDefault);
 		}
+
+		Injector::unnest();
 	}
 
 	public function provideSingletons()
--- a/control/injector/Injector.php
+++ b/control/injector/Injector.php
@@ -866,7 +866,7 @@ class Injector {
 			return $this->instantiate($spec, $name);
 		}
 
-		return $this->instantiate($spec);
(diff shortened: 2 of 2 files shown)


tests/model/DataObjectTest.php
```diff
@@ -38,6 +39,13 @@ class DataObjectTest extends SapphireTest {
 	 */
 	public function testSingleton($inst, $defaultValue, $altDefaultValue)
 	{
+		// Calls to scaffold the test database may have cached service specs for DataObjects
+		// with the incorrect 'type' set (singleton instead of prototype)
+		Injector::nest();
+		$reflectionProp = new ReflectionProperty('Injector', 'specs');
+		$reflectionProp->setAccessible(true);
+		$reflectionProp->setValue(Injector::inst(), array());
```
kinglozzer: This smells, but without adding extra API there’s no way to clear the service spec cache :(

maxime-rainville:
We're only accepting critical bug fixes against SS3 now. I'm not sure this would qualify, but I'm open to changing my mind.

sminnee:
I think that given this is a regression in 3.5 of a pretty fundamental part of SilverStripe it counts as a sufficiently important fix to merge in to 3.x.

I've had a review and I think that this has minimal change of regression – the type argument being passed is clearly in error, and the behaviours it changes are:
 * adding the object to the service cache (which shouldn't be done with a non-singleton)
 * setting the DataObject isSingleton args (which is precisely the bug being fixed)

If we apply this to 3.6 we should ensure that 3.7 gets the fix as well. Also I don't know if Loz wants to pick up the duty for tagging appropriate 3.x releases if we do fix this?

kinglozzer:
> Also I don't know if Loz wants to pick up the duty for tagging appropriate 3.x releases if we do fix this?

I’m happy to merge up and push tags, I’ve never done a full release though with cow so I suspect I’d need some guidance!

ScopeyNZ:
How about we just get this merged and expect that it might get released soon anyway...

Was there any complaints? There seems to have been plenty of discussion but no approvals yet?

PR #8751, opened 2019-01-24
Update 03_Form_Templates.md - form template locations

DorsetDigital:
Update guidance on form template location.  They don't necessarily have to be placed in /app/templates, they'll work in the theme directory too.    The current text also seems to suggest that they can be placed in the core directory - something which I don't believe should be advised,

### DIFF
--- a/docs/en/02_Developer_Guides/03_Forms/03_Form_Templates.md
+++ b/docs/en/02_Developer_Guides/03_Forms/03_Form_Templates.md
@@ -16,7 +16,7 @@ $field = new TextField(..);
 $field->setTemplate('MyCustomTextField');
 ```
 
-Both `MyCustomTemplate.ss` and `MyCustomTextField.ss` should be located in **app/templates/** or the same directory as the core.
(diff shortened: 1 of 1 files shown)


DorsetDigital:
Fair point, I was only considering front-end forms.  

I think people will understand that they would need to have their theme enabled for it to work in the front-end; if they don't, then none of their site would be using the theme anyway.  I'd say that particular situation is probably outside the scope of this page.

How about something like:

> To override the template for CMS forms, the templates should be located in /app/templates.  Front-end form templates can be located in /app/templates or in the active theme's /templates directory.

robbieaverill:
How about just the theme's templates dir for frontend form templates? app/templates would override the CMS as well

DorsetDigital:
That's where it all gets a bit confusing.  The lessons talk about using /app/templates rather than the /themes directory for the frontend, which is at odds with some of the other documentation where themes are assumed.  I was trying to be as broad as possible :-)

robbieaverill:
Ah ok, I think the lessons are misleading in that case. Either way, if you could update that sentence slightly that would be great! =)

PR #8755, opened 2019-01-24
Fix tests not loading fixtures and loading themes in CWP recipe kitchen sink

robbieaverill:
Issue: https://github.com/silverstripe/cwp-recipe-kitchen-sink/issues/20


robbieaverill:
Self merging since this is a low impact change to unit tests only

PR #8757, opened 2019-01-25
FIX: Misconfiguration for versioned cache segmentation (fixes #8754)

### LINKED ISSUE #8754: Non-versioned-segmented caches are configured incorrectly
https://docs.silverstripe.org/en/4/changelogs/4.2.0/#versioned-cache-segmentation states that to disable versioned cache segmentation, caches should be configured like:

```yml
  Psr\SimpleCache\CacheInterface.myapp:
    factory: SilverStripe\Core\Cache\CacheFactory     
    constructor:
      namespace: "MyInsensitiveData"
      args:
        disable-container: true
```

This doesn’t work, the caches are still segmented - the following config _does_ prevent segmentation:

```yml
  Psr\SimpleCache\CacheInterface.myapp:
    factory: SilverStripe\Core\Cache\CacheFactory     
    constructor:
      namespace: "MyInsensitiveData"
      disable-container: true 
```

This means that most of the core caches are misconfigured: https://github.com/silverstripe/silverstripe-framework/blob/4.3.0/_config/cache.yml

kinglozzer:
Fixes #8754

No unit tests exist for this in framework (as the cache proxy exists in `silverstripe/versioned`)

### DIFF
--- a/_config/cache.yml
+++ b/_config/cache.yml
@@ -18,17 +18,14 @@ SilverStripe\Core\Injector\Injector:
       factory: SilverStripe\Core\Cache\CacheFactory
       constructor:
         namespace: "VersionProvider_composerlock"
-        args:
-          disable-container: true
+        disable-container: true
   Psr\SimpleCache\CacheInterface.RateLimiter:
       factory: SilverStripe\Core\Cache\CacheFactory
       constructor:
         namespace: 'ratelimiter'
-        args:
-          disable-container: true
+        disable-container: true
   Psr\SimpleCache\CacheInterface.InheritedPermissions:
     factory: SilverStripe\Core\Cache\CacheFactory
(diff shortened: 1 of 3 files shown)


kinglozzer:
I just spotted something I missed initially that means the bug may be elsewhere... I’ll close this PR for now and update the issue

kinglozzer:
Okay I’ve re-tested and I’m happy that this is the best solution

ScopeyNZ:
I had no idea we proxied the cache in `Versioned`. This patch could break a bunch of code if people rely on an incorrectly configured cache. To be honest it wouldn't surprise me - it feels a lot more normal for a cache to just be a simple key/value store.

Anyway, LGTM, just flagging a concern.

PR #8765, opened 2019-01-28
WIP: FIX Remove coupling to DB when querying a pages CMS fields

robbieaverill:
Fixes a bug where running unit tests on a Page instance (that isn't written to the DB) in a test class that doesn't use a DB connection will fail when it tries to query the non-existent DB.

Steps to reproduce:

* `composer create-project silverstrip/installer ./43test 4.3.x-dev`
* Add test class (below) into `app/tests/ExampleTest.php`
* `vendor/bin/phpunit app/tests`

Expected: should pass
Actual: "No database selected" exception thrown

```php
<?php

use SilverStripe\Forms\FieldList;

class ExampleTest extends SapphireTest
{
    public function testExample()
    {
        $page = new Page();
        $this->assertInstanceOf(FieldList::class, $page->getCMSFields());
    }
}

```

### DIFF
--- a/src/ORM/Hierarchy/Hierarchy.php
+++ b/src/ORM/Hierarchy/Hierarchy.php
@@ -296,6 +296,10 @@ class Hierarchy extends DataExtension
         $cacheType = 'numChildren';
         $id = $this->owner->ID;
 
+        if (!$this->owner->isInDB()) {
+            return 0;
+        }
+
         // cached call
         if ($cache) {
             if (isset(self::$cache_numChildren[$baseClass][$cacheType][$id])) {


maxime-rainville:
Shouldn't we be adding the `ExampleTest` to the test suite? Or is this edge case tested by something else?

robbieaverill:
I can do if you want, I think I didn’t because it sort of feels like an expectation that it should work rather than a specific logical test. I’m happy to though, will add

robbieaverill:
Let's make this WIP for now - writing this test with a stub object, or with Group which ships with framework (rather than SiteTree) is finding other places where a DB connection is expected.

robbieaverill:
I'll re-address this another time if I get around to it =)

PR #8766, opened 2019-01-28
MINOR: Change “Choose Page” to “Search or choose Page”

sminnee:
Based on discussions with James Ford, it wasn’t clear that you could
search.

### DIFF
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -48,7 +48,7 @@ en:
     VALIDDATETIMEMINDATE: 'Your date has to be newer or matching the minimum allowed date and time ({datetime})'
   SilverStripe\Forms\DropdownField:
     CHOOSE: (Choose)
-    CHOOSE_MODEL: '(Choose {name})'
+    CHOOSE_MODEL: '(Search or choose {name})'
     SOURCE_VALIDATION: 'Please select a value within the list provided. {value} is not a valid option'
   SilverStripe\Forms\EmailField:
     VALIDATION: 'Please enter an email address'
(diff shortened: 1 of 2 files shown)


sminnee:
Behat test fix here: https://github.com/silverstripe/silverstripe-cms/pull/2373

sminnee:
Note that I have opted _not_ to change the translation key as I thought the change was subtle enough for it to be reasonable to retain the old text for other languages, until such time as translators opt to translate it.

sminnee:
OK so what do we do here? Do we require a new entity code for every wording change, and accept that we lose the old translations?

robbieaverill:
For minor changes, we could update existing translation keys and sync with Transifex on that branch. This would mean we don't pull translations into previous patch releases any more, because we'd risk losing translations, or introducing translations incorrect for the old context.

I think the best approach for minor changes would be to add new keys. This way they can safely be pulled into patch releases as well without corrupting old translations.

I also think that whenever possible we should make translation changes in the earliest minor release branch we have and merge them up.

In this case "Choose page" and "Search or choose page" are two different sentences, so I think separating them is the right thing to do.

maxime-rainville:
So what's the conclusion here? Rename `CHOOSE_MODEL` to `SEARCH_OR_CHOOSE_MODEL` in our code and do the same thing in Transifex so we retain existing translation?

unclecheese:
Have updated to new translation key. This is the best solution IMO.

unclecheese:
@robbieaverill If you can state your approval, I'll merge as we're not going to get a green button on this one with the inter-dependent PRs.

PR #8772, opened 2019-01-30
FIX: Block Manifest of the compatibility class Object and lean on the compatibility auto loader

UndefinedOffset:
On PHP 7.2 initial dev/build of a SilverStripe 3.7.1 site can throw a `Fatal error: Cannot use 'Object' as class name as it is reserved`, I experienced this on SilverStripe Platform causing the deployment to fail and rollback.

This pull request simply adds a manifest exclude file in `core/compat` which causes the framework to lean on the auto loader defined in `model/fieldtypes/compat/autoload.php` instead.


UndefinedOffset:
Will this make it into the upcoming 3.7.3 release? or will it need to wait until 3.7.4? Trying to plan ahead with one of our projects that's moving to SilverStripe Platform on a stack that's currently configured for PHP 7.2, but without this pull being merged would need to be downgraded to PHP 7.1.

PR #8775, opened 2019-01-31
FIX: Prevent null->null being flagged as a value change (fixes #8774)

### LINKED ISSUE #8774: DataObject::setField() records NULL -> NULL as a changed field
## Affected Version

4.3.0

## Description

`DataObject::setField()` records a value change from `NULL` to `NULL` as a strict change, even though the value hasn’t actually changed.

## Steps to Reproduce

```php
$member = Member::create()->update(['FirstName' => 'Test', 'Surname' => null]);
$member->write();
$member->update(['Surname' => null]); // or $member->Surname = null;
var_dump($member->getChangedFields());
```

Results in

```
array (size=1)
  'Surname' => 
    array (size=3)
      'before' => null
      'after' => null
      'level' => int 1
```

## Solution

Probably changing `isset()` to `array_key_exists()` in a few places here: https://github.com/silverstripe/silverstripe-framework/blob/4.3.0/src/ORM/DataObject.php#L2630-L2646

kinglozzer:
Fixes #8774

### DIFF
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -2650,14 +2650,14 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
             }
 
             // if a field is not existing or has strictly changed
-            if (!isset($this->original[$fieldName]) || $this->original[$fieldName] !== $val) {
+            if (!array_key_exists($fieldName, $this->original) || $this->original[$fieldName] !== $val) {
                 // TODO Add check for php-level defaults which are not set in the db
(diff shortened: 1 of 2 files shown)


robbieaverill:
I think this change is good, and I think that the unit tests should've caught any unexpected side effects, but I'd still like to get someone else from the @silverstripe/core-team to review before merging.

sminnee:
I'd be more comfortable putting this into the `4` branch since I did a lot of changes to type handing post-4.3.

PR #8777, opened 2019-02-01
4.2 / fix behat

dnsl48:
add behat.yml back (cherry-picking from 4.3)

### DIFF
--- a/behat.yml
+++ b/behat.yml
@@ -1,0 +1,34 @@
+# Run framework behat tests with this command (installed with silverstripe/installer)
+# Note that framework behat tests require CMS module
+# ========================================================================= #
+# vendor/bin/selenium-server-standalone -Dwebdriver.firefox.bin="/Applications/Firefox31.app/Contents/MacOS/firefox-bin"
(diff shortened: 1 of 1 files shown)


behat.yml
```diff
@@ -0,0 +1,34 @@
...
+# ========================================================================= #
+# vendor/bin/selenium-server-standalone -Dwebdriver.firefox.bin="/Applications/Firefox31.app/Contents/MacOS/firefox-bin"
+# vendor/bin/serve --bootstrap-file vendor/silverstripe/framework/tests/behat/serve-bootstrap.php
+# vendor/bin/behat @framework
+# ========================================================================= #
+default:
+  suites:
+    framework:
+      paths:
+        - %paths.modules.admin%/tests/behat/features
```
robbieaverill: You're specifying admin here - should this be framework? Admin is not a dependency of framework

dnsl48: I guess we could refactor that, but I'm not writing it now, that's exactly what we've got in 4.3 right now. I'm merely cherry-picking it down to 4.2 from 4.3.

robbieaverill: Oh right, ok

PR #8781, opened 2019-02-03
FIX Filter hostname in TrustedProxyMiddleware

chillu:
In depth defense, rather than a security vulnerability.
The trusted proxy IP list is empty by default, meaning no request can set this information.
Systems need to opt in to this behaviour by setting specific proxy IPs.
While there is an "all IPs" option, it's not documented on
https://docs.silverstripe.org/en/4/developer_guides/security/secure_coding/#request-hostname-forgery.
If you find and configure this option, you're also expected to sanitise these layers (e.g. a load balancer).

This issue was originally referred to as SS-2018-024

### DIFF
--- a/src/Control/Middleware/TrustedProxyMiddleware.php
+++ b/src/Control/Middleware/TrustedProxyMiddleware.php
@@ -150,7 +150,7 @@ class TrustedProxyMiddleware implements HTTPMiddleware
                 $host = $hostList ? strtok($hostList, ',') : null;
 
                 // Ensure the host name is valid
-                $host = filter_var($host, FILTER_VALIDATE_DOMAIN, ['flags' => FILTER_FLAG_HOSTNAME]);
+                $host = $this->filterHostname($host);
 
                 if ($host) {
                     $request->addHeader('Host', $host);
@@ -181,6 +181,22 @@ class TrustedProxyMiddleware implements HTTPMiddleware
         }
 
         return $delegate($request);
+    }
+
+    /**
+     * Make sure the provided value is a valid hostname
+     * @param string $hostname
+     * @return bool|string
+     */
+    private function filterHostname($hostname)
+    {
+        if (defined('FILTER_VALIDATE_DOMAIN') && defined('FILTER_FLAG_HOSTNAME')) {
+            // Only works on PHP 7
+            return filter_var($hostname, FILTER_VALIDATE_DOMAIN, ['flags' => FILTER_FLAG_HOSTNAME]);
+        } else {
+            $regex = "#^(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])#i";
+            return preg_match($regex, $hostname) ? $hostname : false;
+        }
     }
 
     /**
--- a/src/Control/Middleware/TrustedProxyMiddleware.php
+++ b/src/Control/Middleware/TrustedProxyMiddleware.php
@@ -145,8 +145,15 @@ class TrustedProxyMiddleware implements HTTPMiddleware
             // Replace host
             foreach ($this->getProxyHostHeaders() as $header) {
                 $hostList = $request->getHeader($header);
-                if ($hostList) {
-                    $request->addHeader('Host', strtok($hostList, ','));
+
+                // Get first host (if passed through multiple proxies)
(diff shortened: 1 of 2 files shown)


src/Control/Middleware/TrustedProxyMiddleware.php
```diff
@@ -145,8 +145,15 @@ public function process(HTTPRequest $request, callable $delegate)
...
             foreach ($this->getProxyHostHeaders() as $header) {
                 $hostList = $request->getHeader($header);
-                if ($hostList) {
-                    $request->addHeader('Host', strtok($hostList, ','));
+
+                // Get first host (if passed through multiple proxies)
+                $host = $hostList ? strtok($hostList, ',') : null;
+
+                // Ensure the host name is valid
+                $host = filter_var($host, FILTER_VALIDATE_DOMAIN, ['flags' => FILTER_FLAG_HOSTNAME]);
```
robbieaverill: These two constants were added in PHP 7.0, so won't work in 5.6

tests/php/Control/Middleware/TrustedProxyMiddlewareTest.php
```diff
@@ -0,0 +1,47 @@
...
+        $request = (new HTTPRequest('GET', ''))
+            ->addHeader('X-Forwarded-Host', '">invalid-test.com');
+
+        $response = $m->process($request, $this->getIdentityFn());
+        $this->assertEquals('', $response->getHeader('Host'));
+    }
+
+    protected function getIdentityFn()
+    {
+        return function ($response) { return $response; };
```
robbieaverill: phpcs violations here

src/Control/Middleware/TrustedProxyMiddleware.php
```diff
@@ -176,6 +183,22 @@ public function process(HTTPRequest $request, callable $delegate)
...
+     * @param string $hostname
+     * @return bool|string
+     */
+    private function filterHostname($hostname)
+    {
+        if (defined('FILTER_VALIDATE_DOMAIN') && defined('FILTER_FLAG_HOSTNAME')) {
+            // Only works on PHP 7
+            return filter_var($hostname, FILTER_VALIDATE_DOMAIN, ['flags' => FILTER_FLAG_HOSTNAME]);
+        } else {
+            $regex = "#^(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])#i";
```
robbieaverill: Do you want to add a unit test covering this regex?

maxime-rainville: Just added a test for all sort of hostnames. It's failing right now. I'll have to confirm what we want to do with `localhost:8080` and punny code.

It's Monday job.

src/Control/Middleware/TrustedProxyMiddleware.php
chillu: Actually, we've waited so long with this PR that we can now require PHP 7 in the `4` branch :D I've rebased and removed Max PHP 5 work again

chillu:
> I'll have to confirm what we want to do with localhost:8080 and punny code.

Thanks for adding more test coverage! Looking at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host, https://tools.ietf.org/html/rfc7230#section-5.4, https://tools.ietf.org/html/rfc3986#section-3.2.2. IPv6 can be covered by `FILTER_VALIDATE_IP`. Any host can have a port, which isn't covered by `FILTER_VALIDATE_DOMAIN` or `FILTER_VALIDATE_IP`. We can strip it, but that gets complicated with IPv6 which also uses double colons as delimiters. If I'm reading https://www.rfc-editor.org/rfc/rfc3986.html#section-3.2.2 correctly, this is solved by marking the host in square brackets:

> A host identified by an IPv6 literal address is represented inside the square brackets without a preceding version flag.

Now, I could parse out the port, account for IPv4 and IPv6. But this is getting into diminishing returns for an in-depth defense. In the end, it's up to devs in complex setups (with reverse proxies etc) to secure their setups, and we already go through great lengths to describe the ins and outs of that in https://docs.silverstripe.org/en/4/developer_guides/security/secure_coding/#request-hostname-forgery. I don't see us making any insecure recommendations there, in particular this sentence is accurate:

> If only trusted servers will make requests then you can use '*' to trust all clients.

Closing this.

Issue #8792, opened 2019-02-10
BadMethodCallException when scafolding search


maxime-rainville:
## Affected Version

SilverStripe 4.3

## Description

Originally reported by @purplespider on [SilverStripe forum](https://forum.silverstripe.org/t/upgrading-to-4-3-the-method-scaffoldsearchfield-does-not-exist/1619)

```
[Emergency] Uncaught BadMethodCallException: Object->__call(): the method 'scaffoldSearchField' does not exist on 'PurpleSpider\MSG\SalesCategory'
GET /admin/pages/edit/show/44

Line 54 in /vendor/silverstripe/framework/src/Core/CustomMethods.php
Source

45      * @throws BadMethodCallException
46      */
47     public function __call($method, $arguments)
48     {
49         // If the method cache was cleared by an an Object::add_extension() / Object::remove_extension()
50         // call, then we should rebuild it.
51         $class = static::class;
52         $config = $this->getExtraMethodConfig($method);
53         if (empty($config)) {
54             throw new BadMethodCallException(
55                 "Object->__call(): the method '$method' does not exist on '$class'"
56             );
57         }
58 
59         switch (true) {
60             case isset($config['callback']): {

Trace

    SilverStripe\View\ViewableData->__call(scaffoldSearchField, Array)
    DataObject.php:2234
    SilverStripe\ORM\DataObject->scaffoldSearchFields()
    DataObject.php:2175
    SilverStripe\ORM\DataObject->getDefaultSearchContext()
    GridFieldFilterHeader.php:249
    SilverStripe\Forms\GridField\GridFieldFilterHeader->getSearchContext(SilverStripe\Forms\GridField\GridField)
    GridFieldFilterHeader.php:269
    SilverStripe\Forms\GridField\GridFieldFilterHeader->getSearchFieldSchema(SilverStripe\Forms\GridField\GridField)
    GridFieldFilterHeader.php:491
    SilverStripe\Forms\GridField\GridFieldFilterHeader->getHTMLFragments(SilverStripe\Forms\GridField\GridField)
    GridField.php:434
    SilverStripe\Forms\GridField\GridField->FieldHolder()
    call_user_func_array(Array, Array)
    ViewableData.php:485
    SilverStripe\View\ViewableData->obj(FieldHolder, , 1)
    ViewableData.php:547
    SilverStripe\View\ViewableData->XML_val(FieldHolder, , 1)
    SSViewer_Scope.php:323
```

[...]

PR #8793, opened 2019-02-11
Throw error message when searchable_field does not return a object

wilr:
While upgrading 4.2 to 4.3 the behaviour of GridFieldFilterHeader has changed (example code: https://gist.github.com/wilr/7c59f577e1d072eb66e4c0c84cba6732).

This patch doesn’t affect the behaviour (since the explict object is a better API) but it makes upgrading easier to identify the field.

### DIFF
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -2241,7 +2241,15 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
 
             // Otherwise, use the database field's scaffolder
             } elseif ($object = $this->relObject($fieldName)) {
-                $field = $object->scaffoldSearchField();
+                if (is_object($object) && $object->hasMethod('scaffoldSearchField')) {
+                    $field = $object->scaffoldSearchField();
+                } else {
+                    throw new Exception(sprintf(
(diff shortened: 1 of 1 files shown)


src/ORM/DataObject.php
```diff
@@ -2241,7 +2241,15 @@ public function scaffoldSearchFields($_params = null)
...
-                $field = $object->scaffoldSearchField();
+                if (is_object($object) && $object->hasMethod('scaffoldSearchField')) {
+                    $field = $object->scaffoldSearchField();
+                } else {
+                    throw new Exception(sprintf(
+                        "SearchField '%s' on '%s' does not return a valid DBField instance.",
+                        $fieldName,
+                        get_class($this)
+                    ));
+                }
```
ScopeyNZ: Super minor semantics request:

Can we flip the condition and throw the exception in the `if` statement? Then we can drop the `else`.

maxime-rainville:
Might be related to https://github.com/silverstripe/silverstripe-framework/issues/8792

ScopeyNZ:
My request is way too small to block any merge and there's been some more chatter about this footgun tripping up some devs. Merging

PR #8795, opened 2019-02-12
FIX Caching the result of counting a foreign list for performance

ScopeyNZ:
This PR caches the results of counting and sometimes mapping a full list of all object which is done per DataObject class in `DBForeignKey::scaffoldFormField`. This appeared to have a pretty substantial performance improvements on a large key project when rendering history for a page utilising elemental.

### DIFF
--- a/src/ORM/FieldType/DBForeignKey.php
+++ b/src/ORM/FieldType/DBForeignKey.php
@@ -43,6 +43,13 @@ class DBForeignKey extends DBInt
 
     private static $default_search_filter_class = 'ExactMatchFilter';
 
+    /**
+     * Cache for multiple subsequent calls to scaffold form fields with the same foreign key object
+     *
+     * @var array
+     */
+    protected static $foreignListCache = [];
+
     public function __construct($name, $object = null)
     {
         $this->object = $object;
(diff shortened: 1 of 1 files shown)


src/ORM/FieldType/DBForeignKey.php
```diff
@@ -74,8 +81,21 @@ public function scaffoldFormField($title = null, $params = null)
         // Don't scaffold a dropdown for large tables, as making the list concrete
         // might exceed the available PHP memory in creating too many DataObject instances
         $threshold = self::config()->get('dropdown_field_threshold');
-        if ($list->count() < $threshold) {
-            $field = new DropdownField($this->name, $title, $list->map('ID', $titleField));
+
+        // Add the count of the list to a cache for subsequent calls
+        if (!isset(static::$foreignListCache[$hasOneClass])) {
```
robbieaverill: How would this work between model calls? Would you get cache bleed between them?

ScopeyNZ: The calls are literally `DataList::create($hasOneClass)->count()`. I don't think it matters where it comes from, the result should be the same regardless AFAICS.

I think the only possible avenue for trouble comes if you have something that adds objects after the first call...

robbieaverill: Right, but what about HasManyLists for example, e.g. `$blog->Categories()->count()` and `$blogPost->Categories()->count()`. I guess that kind of logic isn't affected by this change since it would be internally filtered, but just spitballing to check

ScopeyNZ: Yeah this is only for "hasOne"s. Line 65 to 68 fetch the hasOne relation with `$schema->hasOneComponent` and returning if it doesn't exist.

robbieaverill: Cool. Tests are green too which would've caught something if my concern was valid. LGTM.

Issue #8804, opened 2019-02-17
Preview caching in CMS when using HTTPCacheControlMiddleware


ichaber:
## Affected Version

Every version having HTTPCacheControlMiddleware, which was introduced with SilverStripe 4.2.0 will have this issue.

## Description

After editing and saving a page in the CMS, the preview will not reflect the changes after the second edit, if HTTP caching was configured.
This is the behaviour when following the performance suggestions for HTTP headers outlined in the documentation (https://docs.silverstripe.org/en/4/developer_guides/performance/http_cache_headers/). 
The page is served with  _cache-control_ header set to "private, must-revalidate, max-age=300" and the _expires_ header to current time plus whatever is set as max-age in HTTPCacheControlMiddleware. That causes the browser so respond with a Status Code _200 OK (from disk cache)_ and serving the local copy in it's cache. (One way to prevent that is to disable the caching in a developer toolbar)

## Steps to Reproduce

- In the CMS go to a page that uses a controller with caching enabled according to: https://docs.silverstripe.org/en/4/changelogs/4.2.0/#example-usage.
- Wait until the page renders in the preview pane
- Change content on the page in the edit pane
- Save the changes on the page by clicking the _Save_ button
- The preview pane shows the cached version of the page although the changes were successfully saved

**Alternative way to reproduce:**
- Open the developer toolbar in a browser
- Open a page in the CMS and let the preview render
- Reload the whole page
- Search for the request for the document (usually containing the params ?stage=Stage&CMSPreview=1)
- Check that the Status Code in the response is _200 OK (from disk cache)_

## Pull Requests

 * [ ] https://github.com/silverstripe/silverstripe-cms/pull/2471

chillu:
@clarkepaul The impact is that preview appears broken to authors, particularly on high profile sites which tend to use HTTP caching. This was raised by a large SS Ltd bespoke project through Charlie. Maybe add it to the `Content Editing` Experience Debt epic?

PR #8805, opened 2019-02-17
Fix for #8804 which can cause pages in CMS preview to be cached.

ichaber:
To solve the issue with cached responses after an edit in the CMS (described in issue #8804) augmentState in _HTTPCacheControlMiddleware_ set the HTTP headers to disable caching, if the request originates from the CMS preview.
The criteria to distinguish a request in the preview pane from any other request is the URL param **CMSPreview**. This will only be executed, if there is an active session.


src/Control/Middleware/HTTPCacheControlMiddleware.php
```diff
@@ -796,11 +796,17 @@ protected function augmentState(HTTPRequest $request, HTTPResponse $response)
...
         } elseif ($request->getSession()->getAll()) {
-            // If sessions exist we assume that the responses should not be cached by CDNs / proxies as we are
-            // likely to be supplying information relevant to the current user only
 
-            // Don't force in case user code chooses to opt in to public caching
-            $this->privateCache();
+            if ($request->getVar("CMSPreview") === "1") {
+                //Disable caching when rendering the page in the CMS preview pane
+                //this is to prevent browsers from serving the local cache version after a page edit
+                $this->disableCache("true");
```
ScopeyNZ:
```suggestion
                $this->disableCache(true);
```

ScopeyNZ:
I'm not completely up to speed with our caching implementation but it seems the existing behaviour of setting cache to private should be correct?

ScopeyNZ:
After some discussion I understand the reason for the change. `private` cache means the browser will assume it's cache-able where-as normally the CDN would cache it.

Still I'll defer to someone who has more context on the caching work in 4.2.

maxime-rainville:
Is this fix meant to be against SilverStripe 5? If not, can you re-target to a `4` branch?

sminnee:
Since this is a bug recommend `4.3`.

However, I'm not totally sold on the approach. This creates a CMS-specific dependency in the framework module. I think that this check should probably reside within `ContentController::init()` in the CMS module instead. You can access that via:

```php
HTTPCacheControlMiddleware::singleton()->disableCache(true)
```

ichaber:
I'll create a new PR once I've changed the approach.

Issue #8526, opened 2018-10-28
RFC: Introduce a "Needs rehash" API to MemberPassword


ScopeyNZ:
## Affected Version

4.x

## Description

In order to easily facilitate upgrading password hashes as they evolve (and computers get better at generating rainbow tables) we should add an API to check if password hashes need to be rehashed and use this new interface.

This will closely replicate the API provided by PHP for password hashing: http://php.net/manual/en/function.password-needs-rehash.php . Once this is done we should implement a `PasswordEncryptor` that uses the `password_hash` API from PHP and trust PHP core team to stay up to date - which they seems to be capable of doing: https://wiki.php.net/rfc/argon2_password_hash, https://wiki.php.net/rfc/argon2_password_hash_enhancements

## Proposal

- Add a method to `MemberPassword` called `needsRehash($password)` or something similar.
- Add a method to the `PasswordEncryptor` interface along the same lines
- After checking a password in `MemberAuthenticator` we use the `needsRehash` and perform the rehash if necessary
- Add a new default `PasswordEncryptor` which delegates the provided `password_hash` API - This could be a follow up I guess.

ScopeyNZ:
Yes. Or even if they're in bcrypt and you decide to upgrade to argon2i then it would happen on the next user login (because you need to rehash their plain-text password). It's transparent to the user but you're upgrading the password hash.

ScopeyNZ:
Ok I've been looking at this, also taking into account @madmatt's comments on #8525 about renaming these classes too. I'm trying to come up with an plan to essentially rewrite this hashing API to be password/member inspecific, and then have some nice drop in code so the existing API still works. Of course we'd still deprecate the whole existing API slated for removal in 5.0

## Goals:
- Drop "Encrypt" terminology
- Hashing should be an API that is available irrespective of passwords or members
- Rehashing should be part of the API, and taken care of __by default__
- Audit/drop/replace use of existing "`PasswordEncryptor`" API usage

## Write a new API

Here's my proposal for a new API:

### CryptographicHashService
- `hash($plaintext): string`
- `verify($plaintext, $hash): boolean`
- `needsRehash($hash): boolean`
- `setHashMethod(HashMethodInterface $hashMethod): $this`

By default on construct of this service it will use a configured hash method. More on this later. It's pretty commonplace to store the salt on the result of the hash these days, that's what I recommend here. Validating a hash using this API will use a timing safe comparison function (not currently used).

[...]

Issue #8525, opened 2018-10-28
RFC: Deprecate and warn against using bad PasswordEncyptors


ScopeyNZ:
## Affected Version

4.3.x / 5.0

## Description

Currently we have several terrible options for password hashing:
- MySQLOldPassword
- MySQLPassword
- None
- PhpHash
- PhpLegacyHash

The MySQL "old" option is awful, and can be reversed with ~50 lines of C code, and the "new" version doesn't seem to hold up to much scrutiny either.

I have no idea why we offer a "none" option. We should never encourage this.

PHPHash (and the extension LegacyHash) uses PHPs `hash` function. This is not suitable to password hashing purposes - these hashes are designed for efficiency - the opposite of what you want for passwords.

All of those are insufficient methods for hashing passwords.

## Proposal

We mark all these "encryptor"s (hashers) as deprecated and show a warning (in dev mode? Only while logged in as admin?) if they are being used.

ScopeyNZ:
I thought that might be the case but:

https://github.com/silverstripe/silverstripe-framework/blob/beec0c0d47d542c6d5b7aa1d32f49cd6fec326cc/src/Security/MemberAuthenticator/MemberAuthenticator.php#L140

I think regardless we should at least put more of a disclaimer than "Not recommended" in the classes docblock.

Firesphere:
I'm unsure where exactly, but I know that if you use DefaultAdmin, at some point in the process, it attempts to hash with `None`.... I should know this, but I can't find it at the moment, it's been quite a while since I encountered it.

Firesphere:
Not a valid counter-argument against deprecating bad hashing methods though! I fully agree it should be done. Just saying we should be a bit cautious and not break DefaultAdmin-alike services in the process :)

madmatt:
While we're at it - please name `encryptor` to `hasher` or similar to indicate what it's actually doing :D

chillu:
Haha, mea culpa, I named them "encryptor" back in the days. Can I blame Sam for missing that in peer review back in 2012ish? ;)

Firesphere:
See https://github.com/silverstripe/silverstripe-framework/issues/8525 :)

Firesphere:
How can this have the label `affects/v3`? You want to change it for V3 as well @chillu ?

chillu:
Nope, 3.x is in security and critical bugfix maintenance mode.

ScopeyNZ:
[...]

PR #8806, opened 2019-02-18
WIP Adding a new hashing API

ScopeyNZ:
This is a WIP PR. I still need to get tests green, do some more testing, and write some more tests. I'm raising the PR welcoming people to have a look, give suggestions, or constructive feedback on the approach.

New:
- Hashing is now called "hashing" instead of "encryption"
- There's a new interface for hash methods:
"SilverStripe\Security\HashMethod\HashMethodInterface"
- The new interface supports rehashing over time
- Created a new _default_ hash method that just proxies through to the PHP
password hashing API
- Implemented a HashMethod which proxies through to the old method
- Implemented a new simple service for consuming this API outside of a
"Member" sense: `CryptographicHashService`
- Implemented a service for using it with a member: `PasswordHashService`

Implementation detail:
- Detail about the passwords is stored on the same field as the password
- The hash service being used is configured via config
- Changing the hash using this method is non-breaking and passwords will be
converted over time (when the hash is verified)

Related issues:

- #8526 (RFC)
- #8525 (RFC)


src/Security/Member.php
```diff
@@ -254,14 +254,6 @@ class Member extends DataObject
      */
     private static $auto_login_token_lifetime = 172800;
 
-    /**
-     * Used to track whether {@link Member::changePassword} has made changed that need to be written. Used to prevent
-     * the write from calling changePassword again.
-     *
-     * @var bool
-     */
-    protected $passwordChangesToWrite = false;
```
ScopeyNZ: Potentially I need to restore this. I'd consider this not part of public API and it's no longer used in the new implementation

src/Security/Member.php
```diff
@@ -921,11 +928,20 @@ public function onBeforeWrite()
                 ->send();
         }
 
-        // The test on $this->ID is used for when records are initially created. Note that this only works with
-        // cleartext passwords, as we can't rehash existing passwords. Checking passwordChangesToWrite prevents
-        // recursion between changePassword and this method.
-        if (!$this->ID || ($this->isChanged('Password') && !$this->passwordChangesToWrite)) {
-            $this->changePassword($this->Password, false);
+
+        if (!$this->isChanged('PasswordExpiry') && (!$this->isInDB() || $passwordIsChanged)) {
```
ScopeyNZ: This is moved from the `changePassword` method.

src/Security/Member.php
```diff
@@ -1713,63 +1719,63 @@ public function validate()
...
+//        if (!$suppressNotice) {
+//            user_error(
+//                'Setting a Password on a member should no longer be plaintext. To update a password for a member use ' .
+//                    'the changePassword method or use PasswordHashService instead',
+//                E_USER_NOTICE
+//            );
+//        }
 
-        return $valid;
+        $this->changePassword($password, false, true);
```
ScopeyNZ: This bit could be improved - I think. I've altered the functionality so that passwords are never set _unhashed_ on a Member object. This is capturing calls to `$member->Password = 'thing'`. I'll at least like to add in a deprecation notice - but I need to update places that set a password to use `PasswordHashService::setForMember` first. The challenge here is `Field::saveInto`...

src/Security/Member.php
```diff
@@ -1713,63 +1719,63 @@ public function validate()
...
     }
 
     /**
      * Takes a plaintext password (on the Member object) and encrypts it
      *
+     * This method is deprecated from 4.4.0. Member passwords should be hashed when set on the member. Use
+     * PasswordHashService to set passwords for members. Note that existing calls to update a password
+     * (`$member->Password = 'thing'`) will automatically hash now.
+     *
+     * @deprecated 4.4.0
```
ScopeyNZ: I might just be able to remove this API if we don't consider it "public".

src/Security/MemberAuthenticator/ChangePasswordHandler.php
```diff
@@ -158,8 +158,10 @@ protected function setSessionToken($member, $token)
             Injector::inst()->get(IdentityStore::class)->logOut();
         }
 
-        // Store the hash for the change password form. Will be unset after reload within the ChangePasswordForm.
-        $this->getRequest()->getSession()->set('AutoLoginHash', $member->encryptWithUserSettings($token));
+        // Store the token for the change password form. Will be unset after reload within the ChangePasswordForm.
+        $session = $this->getRequest()->getSession();
+        $session->set('AutoLoginHash', $token);
+        $session->set('PasswordResetMemberID', $member->ID);
```
ScopeyNZ: The "Hash" is now a token - arguably the randomness of the token is reduced when "hashed with user info". Also now it's tracking the ID of the member in the session too, meaning that we can fetch the member and their reset tokens rather than scanning for all members.

mooror:
I look forward to this new API @ScopeyNZ. Keep up the good work.

src/Security/Member.php
```diff
@@ -351,7 +343,8 @@ public function validateCanLogin(ValidationResult &$result = null)
             $result->addError(
                 _t(
                     __CLASS__ . '.ERRORLOCKEDOUT2',
-                    'Your account has been temporarily disabled because of too many failed attempts at ' . 'logging in. Please try again in {count} minutes.',
+                    'Your account has been temporarily disabled because of too many failed attempts at ' .
+                    'logging in. Please try again in {count} minutes.',
                     null,
                     array('count' => static::config()->get('lock_out_delay_mins'))
```
Firesphere: Switch to shorthand array notation while you're at it?

src/Security/MemberAuthenticator/ChangePasswordHandler.php
```diff
@@ -219,14 +221,15 @@ public function doChangePassword(array $data, $form)
 
         $session = $this->getRequest()->getSession();
         if (!$member) {
-            if ($session->get('AutoLoginHash')) {
-                $member = Member::member_from_autologinhash($session->get('AutoLoginHash'));
+            if ($session->get('AutoLoginHash') && ($memberId = $session->get('PasswordResetMemberID'))) {
```
Firesphere: Wouldn't this break if the session of the member has expired?

Cheddam:
@ScopeyNZ Still keen to move forward with this PR?

ScopeyNZ:
Sure.

emteknetnz:

ScopeyNZ:
This is a bit of a pain to close - as I think it has massive value for SilverStripe. Realistically I'm not sure I'll get the enthusiasm to tackle this again, as Security is a bit of a untamed beast in SilverStripe.

I'll close for now, but if anyone comes across this, feel free to reach out here or in Slack for some assistance if they want to take over.

PR #8809, opened 2019-02-18
Fix makePartialTree call with $nodeCountThreshold from config

shot131:
Change hardcoded number nodeCountThreshold to config variable.


robbieaverill:
Thanks for taking the time to create a pull request for this. Unfortunately, [SilverStripe 3 has entered limited support in June 2018](https://www.silverstripe.org/blog/update-on-silverstripe-5-x/). This means we'll only be fixing critical bugs and security issues for SilverStripe 3 going forward.

This change would be an enhancement, which we're not introducing any more of in SilverStripe 3. I've checked, and this change has already been made in SilverStripe 4.

PR #8812, opened 2019-02-20
BUG Explicitely define a dependency conflict with phpunit 6 and above

maxime-rainville:
This PR updates `composer.json` to explicitly state that SilverStripe won't work with any version of phpunit 6 or above.

The issue here is that if you add phpunit with this command, composer will try to install version 7.5, which will not work at all: `composer require phpunit/phpunit --dev`.

The impetus for this is that it will make my life a bit easier for fixing https://github.com/silverstripe/silverstripe-upgrader/issues/86. They are other ways I could address this, but this one is probably the most elegant.

I've set the lower limit to phpunit 5.3. I ran some quick tests with 5.3 and it appears to work reasonably well, although I didn't throw the kitchen sink at it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -101,5 +101,8 @@
         "php-peg": "php thirdparty/php-peg/cli.php src/View/SSTemplateParser.peg > src/View/SSTemplateParser.php"
     },
     "minimum-stability": "dev",
-    "prefer-stable": true
+    "prefer-stable": true,
+    "conflict": {
+        "phpunit/phpunit": "<5.3 || >=6"
+    }
 }


ScopeyNZ:
Doesn't it make more sense to limit the dependency on `phpunit` rather than declare a conflict? Technically this package is still installable with an unsupported version of phpunit if you're not planning on running the tests right?

maxime-rainville:
Problem is that it's not a dependency on framework. You can create a project without unit test. The root composer project ignores dev dependencies.

ScopeyNZ:
Okay if my project is using framework right now (`@stable`) and in my project I'm using the latest bleeding edge version of PHPUnit. I've got all my weird PHPUnit 7 specific greatness going and it's all hunky dory.

Now I see there's a new version of framework - I'd better update. Oh but I can't because it declares a conflict with my version of PHPUnit that I've been using for months.

I feel like that'll be a genuine problem that might occur from this. Perhaps I'm misunderstanding or assuming something incorrectly?

maxime-rainville:
Can you get phpunit 7 working with SapphireTest?

robbieaverill:
We use PHPUnit 7 already in SilverStripe 5, and our recipes and modules all declare PHPUnit ^5.7 as their constraints. How are you expecting people to add ^6 to their projects given that it's always defined as a dev dependency?

maxime-rainville:
This would only be relevant for someone who started their composer file from scratch and/or decide to upgrade to a later phpunit version afterwards.

The reason why I would like to add this is so I can tell the upgrader to require phpunit without having to tell it a specific version. This would avoid me having to hard-code a specific rule for phpunit. If we decide to support a newer version of phpunit in a later release, the upgrader would still work because it would be driven by the `conflict` rule.

At least, that was the plan. Just ran some tests locally against the PR and it looks like composer is not as clever as I'm giving it credit for. If you try running `composer require phpunit/phpunit --dev`, it will still try to install PHPUnit 7.5. Composer doesn't seem to be smart enough to look for a version that doesn't match the conflict constraint. It just does a hard fail without even telling you about the conflict rule.

This won't do for my purposes.

maxime-rainville:
Actually, it can work, but you have to use a wild card `composer require phpunit/phpunit:* --dev`. When you don't specify a constraint it seems to assume you're always after the latest stable release.

robbieaverill:
I think I'd rather document the phpunit versions that work with each major release more clearly to be honest

maxime-rainville:
I'll close since there's no consensus ... I ended addressing my issue by hard-coding it the upgrader.

If we decide to upgrade to a later version of PHP unit we'll need to update the upgrager.

PR #8816, opened 2019-02-21
FIX: Ensure DBComposite::writeToManipulation() is called on writes (fixes #8800)

### LINKED ISSUE #8800: DBComposite::writeToManipulation() is never called
## Affected Version

4.0.0 - 4.3.0

## Description

Noticed this when implementing a custom `DBComposite` field type - before a DataObject is written, `DBField::writeToManipulation()` should be called against all database fields, but it isn’t called on `DBComposite` fields.

Looks like there was a subtle behavioural change in commit https://github.com/silverstripe/silverstripe-framework/commit/11bbed4f76b2dab35ed0c1db50114eef4a29c4b7.

Before: https://github.com/silverstripe/silverstripe-framework/blob/f0dd9af699cd03e3e315ff6e8297efdc0a9cfd23/ORM/DataObject.php#L1255-L1263
After: https://github.com/silverstripe/silverstripe-framework/blob/11bbed4f76b2dab35ed0c1db50114eef4a29c4b7/ORM/DataObject.php#L1190-L1200

The config option `'dbOnly'` (since replaced by the bitmask `DataObjectSchema::DB_ONLY`) shouldn’t be set here - as that explicitly excludes “virtual” (i.e. composite) fields.

Marking as low impact as the composite values are still saved - by virtue of `DBComposite::setField()` calling `$record->setField()` for each of the composite values

# Related PRs
* https://github.com/silverstripe/silverstripe-framework/pull/8816
* https://github.com/silverstripe/silverstripe-framework/pull/10913

kinglozzer:
Fixes #8800

### DIFF
--- a/tests/php/ORM/DBCompositeTest/DBDoubleMoney.php
+++ b/tests/php/ORM/DBCompositeTest/DBDoubleMoney.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace SilverStripe\ORM\Tests\DBCompositeTest;
+
+use SilverStripe\Dev\TestOnly;
+use SilverStripe\ORM\FieldType\DBMoney;
+
+class DBDoubleMoney extends DBMoney implements TestOnly
+{
+    public function writeToManipulation(&$manipulation)
+    {
+        // Duplicate the amount before writing
+        $this->setAmount($this->getAmount() * 2);
+
+        parent::writeToManipulation($manipulation);
+    }
+}
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -1243,7 +1243,7 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
             }
 
             // Ensure this field pertains to this table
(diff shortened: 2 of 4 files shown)


tests/php/ORM/DBCompositeTest/DBDoubleMoney.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace SilverStripe\ORM\Tests\DBCompositeTest;
+
+use SilverStripe\ORM\FieldType\DBMoney;
+
+class DBDoubleMoney extends DBMoney
```
robbieaverill: TestOnly?

kinglozzer: Oops, added 😄

michalkleiner:
@kinglozzer would you have any recollection of why this was closed mere 4 years ago? :-D

At a guess, it feels like the team was just closing stuff that were not to make 4.x or something along the lines 🤷.

PR #8830, opened 2019-02-26
Fix #8829: mention get_one does not escape field names

### LINKED ISSUE #8829: The filter argument on `get_one` is not properly escaped
Follow up of  this [forum thread](https://forum.silverstripe.org/t/postgresql-unit-test-fails-unknown-field/1617).

It seems that the optional filter argument of `get_one()` is not escaped in the same way of `filter()`. This could give surprising results (e.g. on PostgreSQL, where quoted fields are case sensitive and unquoted are case insensitives).

    // It does not work: it becomes ... WHERE (FirstName = 'John') ...
    $obj = DataObject::get_one(Player::class, [ 'FirstName' => 'John' ]);
    // It works
    $obj = Player::get()->filter('FirstName', 'John' )->first();

If this is expected behavior I can provide a patch that at least mention it in the `get_one` docblock.

CC: @mlewis-everley

# Related PRs
* [ ] https://github.com/silverstripe/silverstripe-framework/pull/8830

ntd:
Documentation improvement because of #8829.

### DIFF
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -3154,9 +3154,14 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
      * Return the first item matching the given query.
      * All calls to get_one() are cached.
      *
+     * The filter argument supports parameterised queries (see SQLSelect::addWhere() for syntax examples). Because
+     * of that (and differently from e.g. DataList::filter()) you need to manually escape the field names:
+     * <code>
+     * $member = DataObject::get_one('Member', [ '"FirstName"' => 'John' ]);
+     * </code>
+     *
      * @param string $callerClass The class of objects to be returned
      * @param string|array $filter A filter to be inserted into the WHERE clause.
-     * Supports parameterised queries. See SQLSelect::addWhere() for syntax examples.
      * @param boolean $cache Use caching
      * @param string $orderby A sort expression to be inserted into the ORDER BY clause.
      *


src/ORM/DataObject.php
```diff
@@ -3102,10 +3102,15 @@ public static function get(
...
+     * The filter argument supports parameterised queries (see SQLSelect::addWhere() for syntax examples). Because of
+     * that (and differently from e.g. DataList::filter()) you need to manually escape the field names:
+     * <code>
+     * $member = DataObject::get_one('Member', [ '"FirstName"' => 'John' ]);
+     * </code>
+     *
      * @param string|null  $callerClass The class of objects to be returned. Defaults to the class that calls the method
      *                                  e.g. MyObject::get_one() will return a MyObject
      * @param string|array $filter      A filter to be inserted into the WHERE clause.
-     *                                  Supports parameterised queries. See SQLSelect::addWhere() for syntax examples.
```
ScopeyNZ: Did you mean to remove this line? It's still relevant?

ntd: Yes to both ;) I just moved it in the body to be able to append paragraph and example mentioning the quoting issue.

maxime-rainville:
Is this meant to be targeted at SilverStripe 5? I think it would make more sense in SilverStripe 4.

sminnee:
Agree that this should go into the the `4` branch but it will need to be re-based by the submitter before it's ready for merge.

ntd:
Sorry for the mess, I was thinking it was already against the SS4.

Issue #8411, opened 2018-09-26
RFC - DataObject ID Decoupling


dhensby:
Proposal to remove dependency on auto-incrementing Database driven IDs

### Motivations

- Remove race conditions on inserts / ID dependencies
Inserting and pulling out auto-incrementing IDs can cause race conditions when using distributed architecture, especially with distributed databases
- Pre-determined IDs before writing
At the moment, to get an ID you have to write the item to the DB first, this is bad practice as all Models should have an ID at all times; this can also allow us to simplify "unsaved relations" because we'll be able to set a relation on an ID that's not been written yet, allowing us to depend on eventual consistency.
- Security: Don't implicitly expose existing IDs
Incrementing IDs are guessable and easy to determine if an ID will exist. As IDs are often used in URLs, they should not be easily guessable
- Better database compatibility
We decouple from DB integrations and allow people to define their own ID scheme; our Model shouldn't *depend* upon the database to work

### Impact

- Have to either upgrade existing IDs or work with both numeric and UUIDs
- Have to fix any `is_numeric` logic from ID validation / make ID validation a bit more abstract
- Stop ordering items by ID to get the inserted order (use Created instead)
- Modify `isInDB()` / `exists()` methods to not just assume a model exists if it's got an ID (either we need to ping the DB or we need to set a flag on loading from DB / writing) - Perhaps replace `isInDB()` with `isPersisted()`?

### Approach

- Create an "ID Generator" interface
This would have the responsibility of generating and validation IDs so that this is abstracted away and we no longer need to do `is_numeric($this->ID)` to validate IDs. 
- Create ID Generators
By default we'd have a UUID generator but we could also have a "legacy" generator. This could be a hybrid that accepts numeric IDs and UUIDs or just a numeric auto-incrementing ID generator.

robbieaverill:
I like the idea for a few reasons:

* a UUID will always be unique, which is good for highly concurrent writes
* we can stop having to write models in order to get the ID for it (then do something and write it again later)

One concern would be that SilverStripe URLs will become much longer and less human friendly

dhensby:
> One concern would be that SilverStripe URLs will become much longer and less human friendly

Well, an integer is almost as meaningless to a human as a UUID, but URLs in the CMS will become longer, but that's an aesthetics issue and UUIDs are not there for aesthetics, they are there for uniqueness.

---

[...]

PR #8831, opened 2019-02-27
BUG Fix writeBaseRecord with unique indexes

### LINKED ISSUE #6819: unique index / indexes are error prone
**Summary:**

Unique indexes are not such a good idea within the current Silverstripe set up (3.5.3), AFAIK, because they are prone to race conditions and other errors, making all future inserts impossible.  The reason for this is that `DataObject::write` carries out its writing in two steps. The first step writes an empty record to obtain an ID, the second (and further writes), add the data.  If the first write is performed and the second write crashes or whatever else can go wrong then the table can not be written to ever again (until the empty record is manually removed). 

----

When creating a unique index, you are likely to encounter problems due to the way SS inserts new records.  Please forgive me if I got it all wrong.  I have limited time to raise this ticket and I am not coding pro, but it may be worth noting. I have already raised this on the dev group (https://groups.google.com/forum/#!topic/silverstripe-dev/n-FZ8pSsWD8)

AFAIK Records (database rows) are written, 
1. firstly with creation date and ID only. 
2. with second write for any data.

This means that no records can be written to a table while any write process is between 1 and 2.  Sometimes this can be because of race-like conditions but it could also be the case that 2 does not happen because of faulty data, server failure, or whatever else you can imagine.

Here is the relevant code (from 3.5.3) in DataObject:

image

image

Line 1313 ... 1316 above executes a query:

image

call to DB::prepared_query:

image

and then to the connector:

image

The way I read this code is that `writeBaseRecord` actually inserts an empty record, after which further write(s) add the additional data.

Options (some of them a little crazy!):

* make this limitation clear in documentation OR
* add roll-back procedure to delete new entry in case of error OR
* disallow unique indexes as they do not work so well with Silvertripe ORM OR
* lock database row until full insert has been done (this will prevent race conditions I guess) OR
* refactor the whole system ;-) so that we insert the data with the original insert OR
* build a silverstripe specific way to deal with unique indexes that does not involve the DB itself
* ...

tractorcow:
Fixes #6819

I've had this bug in mind for years. This will fix the below bug once and for all:

```
Couldn't run query:

INSERT INTO "App_MemberRole"
 (\"Created\")
 VALUES
 (?)

Duplicate entry '0-0' for key 'OneMemberPerProperty'
```

This often came up when writing ChangesetItem due to the primary key being a composite on a couple of columns.

### DIFF
--- a/tests/php/ORM/DataObjectSchemaTest.php
+++ b/tests/php/ORM/DataObjectSchemaTest.php
@@ -343,4 +343,35 @@ class DataObjectSchemaTest extends SapphireTest
             'columns' => ['IndexedMoneyCurrency', 'IndexedMoneyAmount']
         ], $indexes['IndexedMoney']);
     }
+
+    /**
+     * Ensure that records with unique indexes can be written
+     */
+    public function testWriteUniqueIndexes()
+    {
+        // Create default object
+        $zeroObject = new AllIndexes();
+        $zeroObject->Number = 0;
+        $zeroObject->write();
+
+        $this->assertListEquals(
+            [
+                ['Number' => 0],
+            ],
+            AllIndexes::get()
+        );
+
+        // Test a new record can be created without clashing with default value
+        $validObject = new AllIndexes();
+        $validObject->Number = 1;
+        $validObject->write();
+
+        $this->assertListEquals(
+            [
+                ['Number' => 0],
+                ['Number' => 1],
+            ],
+            AllIndexes::get()
+        );
+    }
 }
--- a/src/ORM/DataObject.php
+++ b/src/ORM/DataObject.php
@@ -1413,8 +1413,10 @@ class DataObject extends ViewableData implements DataObjectInterface, i18nEntity
         // Inserts done one the base table are performed in another step, so the manipulation should instead
         // attempt an update, as though it were a normal update.
         $manipulation[$table]['command'] = $isNewRecord ? 'insert' : 'update';
-        $manipulation[$table]['id'] = $this->record['ID'];
         $manipulation[$table]['class'] = $class;
+        if ($this->isInDB()) {
+            $manipulation[$table]['id'] = $this->record['ID'];
+        }
     }
 
     /**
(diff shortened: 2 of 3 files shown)


tests/php/ORM/DataObjectSchemaTest.php
```diff
@@ -343,4 +343,35 @@ public function testFieldsCanBeIndexedFromFieldSpecs()
...
+            [
+                ['Number' => 0],
+            ],
+            AllIndexes::get()
+        );
+
+        // Test a new record can be created without clashing with default value
+        $validObjcet = new AllIndexes();
+        $validObjcet->Number = 1;
+        $validObjcet->write();
```
tractorcow: Eh, I'll fix this in a second.

tractorcow:
Interestingly this bug has been around as long as SilverStripe has.

https://github.com/silverstripe/silverstripe-framework/blob/2.1/core/model/DataObject.php#L432-L436

tractorcow:
Oops, prepareManipulationTable assumes ID was set. A small tweak fixed that. :)

dhensby:
Nice one @tractorcow! 🎉

patricknelson:
I'm guessing in the example above `App_MemberRole` is actually a descendant of another `DataObject` correct? If so, then the reason why this fix works (i.e. checking `$this->isInDB() === true` first) is because you're actually writing a new record in a _child object's_ table, so the child object (who's `ID` must be synchronized with the parent object's table)  doesn't actually exist in the DB, which is **why** you're assigning the ID. As the code is written, it looks backwards, but it works because the `->ID` is indeed assigned because the _base_ record has in fact already been written and (naturally) an artifact of that is the the ID will not be zero and thus no duplicate errors.

It's more of a point on succinct semantics and reducing tech debt/legacy. Maybe a comment can be introduced here, like:


```php
        // Ensure we have an ID first and then synchronize ID's across child object records.
        if ($this->isInDB()) {
            $manipulation[$table]['id'] = $this->record['ID'];
        }
```

Or similar. I've proposed an explanation of an API level disambiguation on that nuance in https://github.com/silverstripe/silverstripe-framework/issues/8411#issuecomment-469422545

tractorcow:
I would write something like `// Provide ID if generated for this record, otherwise leave blank to trigger a new auto-increment value`.

robbieaverill:
@tractorcow the kitchen sink found some bugs that this introduced, I've made a tiny patch at https://github.com/silverstripe/silverstripe-framework/pull/8839 to fix that

robbieaverill:
@tractorcow ignore me, actually some dodgy logic in the auditor module

Issue #8823, opened 2019-02-22
PaginatedList should return relative links that don't rely on meta base


axllent:
Currently `PaginatedList` (SilverStripe 3 & 4) returns pagination URL variables that rely on the meta `base` tag to work (eg `blog?start=10`).

For example pagination on `/blog/` generates URLs that, when no base tag is used, result in `/blog/blog?start=10` etc.

# Pull Request
* [x] https://github.com/silverstripe/silverstripe-framework/pull/8834

chillu:
Good point, we should reduce our reliance on the base tag. Wanna create a pull request for this?

sminnee:
`Controller::join_links(Director::baseUrl(), $url)` is the usual approach for this; many `Link()` methods do this already.

axllent:
That works if you are creating the `Link()` functions, however this is regarding the links returned by `PaginatedList` for templating, such as `$PaginatedPages.PrevLink`, `$PaginatedPages.NextLink` and `$PaginatedPages.Pages`. I provided a simple [pull request](https://github.com/silverstripe/silverstripe-framework/pull/8834) which addresses this and hopefully doesn't have negative implications for other things.

robbieaverill:
The PR has been merged and is available in 4.3.x-dev. [SilverStripe 3 has entered limited support in June 2018](https://www.silverstripe.org/blog/update-on-silverstripe-5-x/). This means we'll only be fixing critical bugs and security issues for SilverStripe 3 going forward. If someone wants to backport the PR for SilverStripe 3.x they'd be welcome to do so, but for now I'm going to close this issue.

Thanks @axllent for the PR!

axllent:
Thanks @robbieaverill - that patch for v3 has been on my TODO since this was merged, however considering the limited support and the fact that new projects would use 4 anyway, I just didn't get around to it. It's probably best to leave v3 alone anyway, else risk "legacy" websites that got manually templated to include a leading `/` ending up with a leading `//`.

michalkleiner:
We ended up extending the class, changing the behaviour by allowing for custom prefix and then using instances of the extending class instead.

PR #8834, opened 2019-02-28
Include baseURL with relative setGetVar() links

axllent:
Ensures relative URLs returned with `HTTP::setGetVar()` include `Director::baseURL()`, and thus no reliance on the meta `base` tag. One example where this is an issue is with all URLs generated from `PaginatedList` as reported in #8823 

This should also be fixed in the SilverStripe 3 branch. Assuming this solution is accepted, should I create a separate pull request for that branch too?

### DIFF
--- a/src/Control/HTTP.php
+++ b/src/Control/HTTP.php
@@ -268,7 +268,7 @@ class HTTP
         $newUri = $scheme . '://' . $user . $host . $port . $path . $params . $fragment;
 
         if ($isRelative) {
-            return Director::makeRelative($newUri);
+            return Director::baseURL() . Director::makeRelative($newUri);
         }
 
         return $newUri;
--- a/tests/php/Control/HTTPTest.php
+++ b/tests/php/Control/HTTPTest.php
@@ -246,13 +246,13 @@ class HTTPTest extends FunctionalTest
         }, 'relative/url/');
 
         $this->assertEquals(
-            'relative/url?foo=bar',
+            '/relative/url?foo=bar',
             HTTP::setGetVar('foo', 'bar', 'relative/url'),
             'Relative URL without existing query params'
         );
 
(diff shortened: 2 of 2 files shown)


src/Control/HTTP.php
```diff
@@ -268,7 +268,7 @@ public static function setGetVar($varname, $varvalue, $currentURL = null, $separ
         $newUri = $scheme . '://' . $user . $host . $port . $path . $params . $fragment;
 
         if ($isRelative) {
-            return Director::makeRelative($newUri);
+            return Director::baseURL() . Director::makeRelative($newUri);
```
robbieaverill: By definition this is no longer returning a relative link. I have a conceptual issue with that change

axllent: > By definition this is no longer returning a relative link. I have a conceptual issue with that change

There is no mention of this function returning a relative link, in fact the docblock clearly states:
```
* CAUTION: If the URL is determined to be relative, it is prepended with Director::absoluteBaseURL().
* This method will always return an absolute URL because Director::makeRelative() can lead to
* inconsistent results.
```
The thing is it doesn't "always return an absolute URL".

Anyway, this was an attempt at the least invasive approach to fix the issue. It sounds to me like you are proposing a different core change, so I'm not sure what to do next (or even if I should). This is I think best left to you guys.

robbieaverill: That's fair enough =) I'll remove my concern on that issue

axllent:
Thanks @unclecheese, this solves the issue for SilverStripe 3.*, but not 4.*

Is a more acceptable solution for 4 being worked on (along the lines of what @robbieaverill suggested), or do you need another similar MR for 4?

robbieaverill:
@axllent pull requests are always welcome! This isn't on our radar at the moment

PR #8838, opened 2019-03-04
Use '/' as an alternative designation for root in routing

Cheddam:
Currently, SilverStripe's router logic uses an empty string (`''`) as the sole designation for a 'root' rule in Director config and in URL Handlers in controllers. This generally makes sense and works fine, until you try to specify an HTTP method in your URL Handler:

```
private static $url_handler = [
    'GET' => 'getMethod' // Matches /GET as a URL
    'GET /' => 'getMethod' // Calls getMethod, but _also_ throws a 'cannot handle sub-URLs' error
    'GET ' => 'getMethod' // Works, but looks bizarre and is undocumented
];
```

Specifying HTTP methods is arguably an important security practice, and avoids adding boilerplate in the top of Controller methods in order to fail early when the wrong method is used. It also allows different controller methods to be used for different HTTP methods against the same URL, e.g. separating 'GET the things' from 'POST a new thing'.

This PR adds `'/'` as a valid alternative to `''` for designating a 'root' rule, which results in more sensible config:

```
private static $url_handler = [
    'GET /' => 'getMethod' // Now works as expected!
    '/' => 'anyMethod' // Also works
];
```

This will also apply to Director rules, though it's less relevant there. I opted not to push for the empty string matching to be deprecated at this stage, but it might make sense to minimise inconsistencies - keen to get feedback on the new behaviour first.

### DIFF
--- a/tests/php/Control/RequestHandlingTest/HTTPMethodTestController.php
+++ b/tests/php/Control/RequestHandlingTest/HTTPMethodTestController.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace SilverStripe\Control\Tests\RequestHandlingTest;
+
+use SilverStripe\Control\Controller;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Dev\TestOnly;
+
+class HTTPMethodTestController extends Controller implements TestOnly
+{
+    private static $url_segment = 'HTTPMethodTestController';
+
+    private static $url_handlers = [
+        'GET /' => 'getRoot',
+        'POST ' => 'postLegacyRoot',
+    ];
+
+    private static $allowed_actions = [
+        'getRoot',
+        'postLegacyRoot',
+    ];
+
+    public function getRoot(HTTPRequest $request)
+    {
+        return "Routed to getRoot";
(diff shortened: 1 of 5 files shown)


Cheddam:
Fair point! I'll knock some out before merge.

Cheddam:
50 weeks later, I've followed through on adding some test coverage 😅

tests/php/Control/RequestHandlingTest/HTTPMethodTestController.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace SilverStripe\Control\Tests\RequestHandlingTest;
+
+use SilverStripe\Control\Controller;
+use SilverStripe\Control\HTTPRequest;
+
+class HTTPMethodTestController extends Controller
```
robbieaverill: It might be worth adding `implements TestOnly`, even though it's not a DataObject it still makes it clear

Cheddam: Nice catch! Fixed.

PR #8839, opened 2019-03-05
FIX Manipulation ID is now always present although it can be null

robbieaverill:
A follow on from https://github.com/silverstripe/silverstripe-framework/pull/8831 that still defines the manipulation ID but with a null value instead. Removing it entirely seems to break some behaviour in manipulation proxy modules that we use (fulltextsearch, auditor, etc). See https://travis-ci.org/silverstripe/cwp-recipe-kitchen-sink/jobs/500785254.

Fixes https://github.com/silverstripeltd/cc-issues/issues/252


robbieaverill:
This is breaking spectacularly in postgres

robbieaverill:
Whoops, this is actually showing a bug in the auditor module. Closing!

PR #8840, opened 2019-03-05
ENHANCEMENT: has_extension() should allow injector overrides

unclecheese:
`DataObject::has_extension()` allows subclasses, but not overrides.

`MyObj::has_extension(Versioned::class) // true`

`MyObj::has_extension(VersionedSubclass::class) // true`

```
  SilverStripe\Versioned\Versioned:
    class: MyProject\MyVersionedSubclass
```

`MyObj::has_extension(VersionedSubclass::class) // false`

### DIFF
--- a/src/Core/Extensible.php
+++ b/src/Core/Extensible.php
@@ -390,7 +390,7 @@ trait Extensible
             }
             $inst = Injector::inst()->get($extension);
             if ($inst instanceof $requiredExtension) {
-                return $strict ? get_class($inst) === $requiredExtension : true;
+                return $strict ? strcasecmp(get_class($inst), $requiredExtension) === 0 : true;
             }
         }
 
--- a/src/Core/Extensible.php
+++ b/src/Core/Extensible.php
@@ -388,6 +388,9 @@ trait Extensible
             if (!$strict && is_subclass_of($extension, $requiredExtension)) {
                 return true;
             }
+            if (Injector::inst()->get($extension) instanceof $requiredExtension) {
+                return true;
+            }
         }
 
         return false;
--- a/src/Core/Extensible.php
+++ b/src/Core/Extensible.php
@@ -388,8 +388,9 @@ trait Extensible
             if (!$strict && is_subclass_of($extension, $requiredExtension)) {
                 return true;
             }
-            if (Injector::inst()->get($extension) instanceof $requiredExtension) {
-                return true;
+            $inst = Injector::inst()->get($extension);
+            if ($inst instanceof $requiredExtension) {
+                return $strict ? get_class($inst) === $requiredExtension : true;
             }
         }
 
--- a/src/includes/constants.php
+++ b/src/includes/constants.php
@@ -139,11 +139,11 @@ if (!defined('BASE_URL')) {
         ) {
             $requestURI = $_SERVER['REQUEST_URI'];
             // Check if /base/public or /base are in the request
-            foreach ([$baseURL, dirname($baseURL)] as $candidate) {
-                if (stripos($requestURI, $candidate) === 0) {
-                    return $candidate;
-                }
+        foreach ([$baseURL, dirname($baseURL)] as $candidate) {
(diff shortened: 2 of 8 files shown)


src/Core/Extensible.php
```diff
@@ -388,6 +388,9 @@ public static function has_extension($classOrExtension, $requiredExtension = nul
             if (!$strict && is_subclass_of($extension, $requiredExtension)) {
                 return true;
             }
+            if (Injector::inst()->get($extension) instanceof $requiredExtension) {
```
ScopeyNZ:
```suggestion
            if (!$strict && Injector::inst()->get($extension) instanceof $requiredExtension) {
```

ScopeyNZ: That should fix the tests. There was a test specifically against this type of checking when `strict` mode is on...

unclecheese: It will fix the test, but I'm not sure it's correct. Maybe it's worthy of discussion.

In my mind, `has_extension()` should check for a named implementation, not necessarily a concretion.

I can see the case where "not strict" would allow subclasses, but if `strict` is on, and the service name matches, that to me seems correct.

tractorcow: If $strict is true, you should probably do a get_class() comparison on the result of Injector::inst()->get().

ScopeyNZ: The docblock of the strict parameter reads:
> if the extension has to match the required extension and not be a subclass
The current patch breaks this functionality. @tractorcow has a good suggestion, we can't just `return true`. Instead we can `return !$strict || get_class([...]) === Injector::inst()->get([...])`.

unclecheese: Agreed on @tractorcow 's strict check.

unclecheese:
Have fixed the tests and added the `strict` check.

unclecheese:
Can't figure out what's wrong with the `phpcs` test. That comes back clean in my local copy (same version), and the file itself looks fine. The changes it's suggesting would break the formatting. Any suggestions would be appreciated.

src/includes/constants.php
```diff
@@ -139,11 +139,11 @@
         ) {
             $requestURI = $_SERVER['REQUEST_URI'];
             // Check if /base/public or /base are in the request
-            foreach ([$baseURL, dirname($baseURL)] as $candidate) {
-                if (stripos($requestURI, $candidate) === 0) {
-                    return $candidate;
-                }
+        foreach ([$baseURL, dirname($baseURL)] as $candidate) {
```
kinglozzer: Appreciate you did this to fix php-cs, but yuck. One thing I’d check (if you haven’t already) is whether the failure is caused by the first `if` condition above being on the same line. I.e. change
```php
if ($baseURL
    && PUBLIC_DIR
```

to 

```php
if (
    $baseURL
    && PUBLIC_DIR
```

silverstripe-accounts: Yeah, thanks @kinglozzer , that makes sense. I was having a hard time figuring out why it wanted that. I tried several reformats, but I'm not sure that was one of them.

It's only in the newer release of php-cs. Bizarre.

unclecheese: Yeah, thanks, @kinglozzer . That makes sense. I could not figure out why it wanted that change. I tried several reformats, but I'm not sure that was one of them. Only in the newest release of php-cs, too. Bizarre.

src/Core/Extensible.php
```diff
@@ -388,6 +388,10 @@ public static function has_extension($classOrExtension, $requiredExtension = nul
             if (!$strict && is_subclass_of($extension, $requiredExtension)) {
                 return true;
             }
+            $inst = Injector::inst()->get($extension);
+            if ($inst instanceof $requiredExtension) {
+                return $strict ? get_class($inst) === $requiredExtension : true;
```
tractorcow: Nice!

tractorcow: Maybe we should do `strcasecmp` just, in case. (pun intended)

Issue #8845, opened 2019-03-07
Poor (missing) error handling for mismatched braces in templates


kinglozzer:
## Affected Version

4.x

## Description

Given the following template:

```
<h2>Title 1</h2>

<% if $Something %>
	<h2>Title 2</h2>
	<li><a href="{$Link">{$Title}</a></li>
<% end_if %>

<h2>Title 3</h2>
```

“Title 2” and “Title 3” will never be output, but nor will any errors. The missing `}` in `{$Link}` causes the issue, but silently.

The parsed template simply stops after `<h2>Title 1</h2>`. Removing the `<% if` statement lets it get a little further, but it still ends abruptly:

```php
$val .= '
<h2>Title 1</h2>

	<h2>Title 2</h2>
	<li><a href="';
```

sminnee:
There's a few ways to approach fixing this:

 * Stricter parsing of the content within `{...}`. I'm not sure how many different variations that needs to support and whether it's just passing the buck to PHP, which might make that difficult.
 * Show a parse error if it hits EOL or another { when it's processing those.

Do you know your way around php-peg enough to figure that out?

kinglozzer:
PR: https://github.com/silverstripe/silverstripe-framework/pull/8846

Issue #8742, opened 2019-01-22
Template parser cannot handle empty template comment


wernerkrauss:
## Affected Version
SS4.3, maybe earlier versions affected

## Description

If you have for some reason an empty template comment `<%----%>` in your template the rest of the template will be treated as commented out. Everything above gets rendererd, below not. 

While a dev is not supposed to insert this manually, unluckily PHPStorm's "comment" method (on Windows that's STRG + / on Numpad) comments an empty line like that.

## Steps to Reproduce

In a template:

```
This will be parsed
<%-- This is commented out --%>
<%----%>
This is not commented out but won't be printed
```

sminnee:
@kinglozzer if you end up looking at that other parser issue you could try your hand at this one too. This might be a simpler one to get your feet wet with ;-)

kinglozzer:
PR: https://github.com/silverstripe/silverstripe-framework/pull/8846

PR #8846, opened 2019-03-08
Template parser bugs

kinglozzer:
Fixes:

- https://github.com/silverstripe/silverstripe-framework/issues/8742
- https://github.com/silverstripe/silverstripe-framework/issues/8845

### DIFF
--- a/src/View/SSTemplateParser.peg
+++ b/src/View/SSTemplateParser.peg
@@ -216,7 +216,7 @@ class SSTemplateParser extends Parser implements TemplateParser
     # Any new template elements need to be included in this list, if they are to work.
 
     Template: (Comment | Translate | If | Require | CacheBlock | UncachedBlock | OldI18NTag | Include | ClosedBlock |
-        OpenBlock | MalformedBlock | Injection | Text)+
+        OpenBlock | MalformedBlock | MalformedBracketInjection | Injection | Text)+
     */
     function Template_STR(&$res, $sub)
     {
@@ -367,6 +367,18 @@ class SSTemplateParser extends Parser implements TemplateParser
         $res['php'] .= ']';
     }
 
+    /*!*
+
+    # This is used to detect a malformed bracket injection - where the closing '}' is missing
+
+    MalformedBracketInjection: "{$" :Lookup !( "}" )
+    */
+    function MalformedBracketInjection__finalise(&$res)
+    {
+        $lookup = $res['text'];
+        throw new SSTemplateParseException("Malformed bracket injection $lookup. Perhaps you have forgotten the " .
+            "closing bracket (})?", $this);
+    }
 
     /*!*
 
--- a/src/View/SSTemplateParser.peg
+++ b/src/View/SSTemplateParser.peg
@@ -1156,7 +1156,9 @@ class SSTemplateParser extends Parser implements TemplateParser
 
     # This is used to remove template comments
 
-    Comment: "<%--" (!"--%>" /(?s)./)+ "--%>"
+    CommentWithContent: '<%--' ( !"--%>" /(?s)./ )+ '--%>'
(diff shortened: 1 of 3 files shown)


src/View/SSTemplateParser.peg
```diff
@@ -1156,7 +1168,9 @@ class SSTemplateParser extends Parser implements TemplateParser
 
     # This is used to remove template comments
 
-    Comment: "<%--" (!"--%>" /(?s)./)+ "--%>"
+    CommentWithContent: '<%--' ( !"--%>" /(?s)./ )+ '--%>'
+    EmptyComment: '<%----%>'
```
kinglozzer: I tried _a lot_ of different things to get this to work as just one rule, but failed :(

sminnee:
Adding a rule for invalid syntax mystified me a bit, although I see we've done that for blocks as well, so I don't think that necessarily blocks merge.

Even with this rule in place, we'd still have issues when you have `{$` and then a letter, but not something that matches `Lookup`. Perhaps `{$broken[` would trigger this situation as a test-case? This PR is no worse than the status quo in that regard, but should we try and incorporate that one too?

It mystified me that the template parser wasn't throwing an error. I wonder if we can get to the bottom of that? I *think* what might be happening is that the top-level `Template` rule is, in fact, not forced to process the _whole_ template. The malformed tag doesn't match any of `(Comment | Translate | If | Require | CacheBlock | UncachedBlock | OldI18NTag | Include | ClosedBlock |
        OpenBlock | MalformedBlock | Injection | Text)` and so parsing simply stops.

If we could sort that out, then the specific "malformed" rules would be principally about improving the error messages.

The comment change looks fine :-)

kinglozzer:
> Even with this rule in place, we'd still have issues when you have `{$` and then a letter, but not something that matches Lookup. Perhaps `{$broken[` would trigger this situation as a test-case? This PR is no worse than the status quo in that regard, but should we try and incorporate that one too?

`{$broken[` will trigger the parse error I’ve added, as `broken` is a valid injection (it just ignores the trailing square bracket):

> [Emergency] Uncaught SilverStripe\View\SSTemplateParseException: Parse error in template on line 32. Error was: Malformed bracket injection {$broken. Perhaps you have forgotten the closing bracket (})?

Other attempts to break this:
- `{$[` - outputs `{$[` in the template
- `{$123` - outputs `{$123` in the template
- `{$$broken` - outputs `{$` in the template (`$broken` is processed separately as a lookup)

> It mystified me that the template parser wasn't throwing an error. I wonder if we can get to the bottom of that? I think what might be happening is that the top-level Template rule is, in fact, not forced to process the whole template. The malformed tag doesn't match any of (Comment | Translate | If | Require | CacheBlock | UncachedBlock | OldI18NTag | Include | ClosedBlock | OpenBlock | MalformedBlock | Injection | Text) and so parsing simply stops.
> 
>If we could sort that out, then the specific "malformed" rules would be principally about improving the error messages.

I think that sounds like a good idea, I’m not sure how to approach it though. Are you suggesting this could replace existing logic (like the new rule here and/or the mismatched open tags rules) or complement them as a more generic “something went wrong with this statement: `X`” handler?

emteknetnz:

kinglozzer:
I think this is ready. It’s not perfect, but it’s an improvement on the status quo

dhensby (approved):

dhensby:
As @sminnee has said, this is an improvement and the bit's that could be improved further are not from a regression.

sminnee:
I think we can merge into 4.6 or 4

dhensby:
Rebased and green 💚

sminnee (approved):

kinglozzer:
Thanks for sorting Dan

PR #8847, opened 2019-03-10
Add option to leave Temp DB when exiting

blueo:
Please describe expected and observed behaviour, and what you're fixing:

TempDatabase will always tear down the database on exit via the register_shutdown_function [on line 198](https://github.com/silverstripe/silverstripe-framework/blob/4/src/ORM/Connect/TempDatabase.php#L198)

This is problematic if you have an external process that needs the DB to stay in place. For example an e2e test that isn't run via PHP. If you use the testsession module and run /dev/testsession/start via http if will happily create a db and session but then immediately remove the DB once the request finishes.

This PR adds a config option to skip the teardown.

### DIFF
--- a/src/ORM/Connect/TempDatabase.php
+++ b/src/ORM/Connect/TempDatabase.php
@@ -4,6 +4,7 @@ namespace SilverStripe\ORM\Connect;
 
 use Exception;
 use SilverStripe\Core\ClassInfo;
+use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Environment;
 use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\Core\Injector\Injector;
@@ -22,6 +23,13 @@ class TempDatabase
      * @var string
      */
     protected $name = null;
+
+    /**
+     * Optionally remove the test DB when the PHP process exits
+     *
+     * @var boolean
+     */
+    private static $teardown_on_exit = true;
 
     /**
      * Create a new temp database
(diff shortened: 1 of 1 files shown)


src/ORM/Connect/TempDatabase.php
```diff
@@ -23,6 +24,13 @@ class TempDatabase
      */
     protected $name = null;
 
+    /**
+     * Optionally remove the test DB when the PHP process exits
+     *
+     * @var boolean
+     */
+    protected static $teardown_on_exit = true;
```
robbieaverill: If you want this to be a config prop it'll need to be private

blueo: whoops thanks :)

blueo: just looping back in @robbieaverill changes made :)

src/ORM/Connect/TempDatabase.php
```diff
@@ -195,13 +203,16 @@ public function build()
...
 
         // Ensure test db is killed on exit
-        register_shutdown_function(function () {
-            try {
-                $this->kill();
-            } catch (Exception $ex) {
-                // An exception thrown while trying to remove a test database shouldn't fail a build, ignore
-            }
-        });
+        $teardownOnExit = Config::inst()->get(static::class, 'teardown_on_exit');
```
robbieaverill: I think I'd rather see `$this->config()->get('teardown_on_exit')`, thoughts?

blueo: ah yes - that's a bit more consistent + less verbose 👍 done.

robbieaverill: My bad, this class doesn't `use Configurable`. It does have config props on it, but rather than adding the trait, maybe revert back to what you had before. Sorry!

blueo: haha oops - thanks travis :)

PR #8849, opened 2019-03-11
FIX: RequestHandler::__construct() should run after middlewares (fixes #8848)

### LINKED ISSUE #8848: Middlewares are called after RequestHandler::__construct() has already run
PR:

https://github.com/silverstripe/silverstripe-framework/pull/8849

---

Tentatively labelling as a bug, though I’m not sure if this is a bug, an enhancement, or as-designed.

Given the following code:
```php
class MyController extends Controller
{
    public function __construct()
    {
        parent::__construct();
        var_dump(Security::getCurrentUser());
    }
}
```

`Security::getCurrentUser()` will always return `null`, even if the user is logged in, as the authentication middleware hasn’t been called yet. Easily rectified in this case by moving to `protected function init()` instead, however...

Current code is essentially:
```php
$requestHandler = Injector::inst()->get($class)
HTTPMiddleware::process(function() {
    $requestHandler->handleRequest();
});
```

I’m suggesting we patch this and change it to (pseudo-code):
```php
HTTPMiddleware::process(function() {
    $requestHandler = Injector::inst()->get($class);
    $requestHandler->handleRequest();
});
```

The change is a relatively simple patch in `Director`. The only potential BC issue I can spot is if a middleware relies on some logic that’s built as part of `RequestHandler::__construct()`, though given that middlewares can’t directly access controllers I don’t see how that’s possible without some weird static implementation.

kinglozzer:


tractorcow:
It's possible that this could have side effects, but I do agree with the change.

I think it was more important back in the day when we had business logic in ContentController::__construct(), but that's not there anymore.

robbieaverill:
We're merging into an unstable branch, so should give time to catch any side effects. Cheers @kinglozzer.

Issue #8598, opened 2018-11-13
RFC: Deprecate/drop PDO


sminnee:
This is essentially the opposite of #6502.

PDO doesn't provide any benefit to us. It's a leaky, lowest-common-denominator abstraction that introduces a lot of issues that are very difficult to debug, and in return saves us from having to mainly 3 pieces of code that are small and pretty easy to maintain (SQLite, MySQL, and PostgreSQL native connectors).

Here area  few of the places where PDO has failed us:

 * Can't seek on query results
 * Doesn't support nested transactions, when doing transactions "properly". Not doing transactions properly causes other undocumented bugs.
 * Segfaulting in SQLite when using runInSeparateProcess on a test
 * Not retaining native_type metadata on PostgreSQL when using a cached prepared statement second time
 * It has no notion of float/decimal values. It's type system just works with ints and strings.

In terms of its impact on bugfixing-effort, PDO is the IE6 of database connectors.

There's a possibility that we're just "using PDO wrong". There have certainly been some bugs introduced by our quirky implementation. But that's not really the point. It's full of undocumented, database-specific limitations, and improving the implementation just seems to introduce other bugs or limitations. Our users gain almost nothing from our PDO support - native drivers are at least as likely to be available as PDO. And our dev team doesn't either.

I recognise that successfully moving to DBAL would render this unnecessary. However, moving to DBAL seems like a _lot_ of work, and I'm not sure that I'd want to park other improvements because this idea has been proposed.

So my recommendations would be:

 * Flag PDO as deprecated in a 4.x minor release
 * Remove PDO support in 5.0
 * Change the installer to use MySQL native by default
 * Allow for some performance or edge-case-feature limitations of PDO use, via judicious skipTest. For example, don't expect real nested transactions / savepoint rollback to work on PDO.

sminnee:
Pinging @dhensby and @tractorcow about this in particular.

sminnee:
> To be fair, my main preference for PDO is just not having to maintain a separate connector for each db type, although we end up having to do it anyway for all the other classes.

Yeah and that seems plausible on paper, but my experience is that the maintenance effort of PDO is greater than the maintenance effort of MySQLConnector + PostgreSQLConnector + SQLiteConnector. Hence this RFC

> How about switching it out to a module instead? I use PDO for a few projects (e.g. where we use MS SQL Server as a secondary data source).

[...]

PR #8851, opened 2019-03-12
FIX Default database connector to MySQLPDODatabase

ichaber:
Set the default database type to MySQLPDODatabase if config `SS_DATABASE_CLASS` is not defined.
That means a PDO connector is used instead of a MySQLi connector.

## Reasoning:
As per changelog of SilverStripe 4.0 (https://docs.silverstripe.org/en/4/changelogs/4.0.0/) the DB Driver **defaults** to **PDO**. This change was introduced in commit https://github.com/silverstripe/silverstripe-framework/commit/9a0e01d4a083514c70992d8c7625c6cfedd6e54a
This has is also highlighted in the documentation of the core environment variables: https://docs.silverstripe.org/en/4/getting_started/environment_management/#core-environment-variables

_However_, when refactoring the API of bootstrap and request handling the behaviour prior to commit 9a0e01d4a was reintroduced.
That means the database connection does  **no**  longer default to **PDO** after commit  https://github.com/silverstripe/silverstripe-framework/commit/3873e4ba008cfc2af7e26ca86665affc289cd677, but uses the previous default MySQLi connection.

### DIFF
--- a/src/Core/CoreKernel.php
+++ b/src/Core/CoreKernel.php
@@ -289,7 +289,7 @@ class CoreKernel implements Kernel
     {
         /** @skipUpgrade */
         $databaseConfig = [
-            "type" => Environment::getEnv('SS_DATABASE_CLASS') ?: 'MySQLDatabase',
+            "type" => Environment::getEnv('SS_DATABASE_CLASS') ?: 'MySQLPDODatabase',
             "server" => Environment::getEnv('SS_DATABASE_SERVER') ?: 'localhost',
             "username" => Environment::getEnv('SS_DATABASE_USERNAME') ?: null,
             "password" => Environment::getEnv('SS_DATABASE_PASSWORD') ?: null,


ichaber:
Once thing to note: Using PDO by default potentially changes the return type for fields.

E.g. A field in a data object, that is returned as an **integer** with MySQLi would be returned as a **string** after this change.

That would mean checks for identical values would fail, which passed before.

maxime-rainville:
We currently have an RFC for dropping support for PDO altogether. https://github.com/silverstripe/silverstripe-framework/issues/8598

In this context, I think it doesn't make much sense to switch the default back to PDO.

ScopeyNZ:
/cc @sminnee 

There's been a bunch of work trying to make return types (as you mentioned) consistent but I'm not sure what branch is targetted for that work.

In any regard developers should explicitly set their DB adapter with `.env` settings so that changes to the default adapter don't affect your existing site.

sminnee:
My recommendation here would be to ensure that there are no vestiges that still default to PDO.

sminnee:
Return type consistency is coming in 4.4. It applies to all backends.

sminnee:
Defaulting to PDO is definitely something I oppose given my RFC ;)

ScopeyNZ:
Given the RFC targetting 5.0, this PR is targetting 5.0 and the relatively low impact of "if you don't have environment config set up" I'm going to close this issue. I think it's unlikely to be merged. If you feel strongly please feel free to continue the discussion on this issue 🙂 

Thanks for the contribution!

PR #8852, opened 2019-03-12
Added an option to create an empty DataList.

mfendeksilverstripe:
Added an option to create an empty `DataList` via a helper function. This is useful when handling some special cases where:

* empty results should be returned and the logic that shapes the `DataList` is quite complex
* we are interested to keep function signature unchanged (mostly for PHP 7)
* don't want to deal with `null` values

### DIFF
--- a/src/ORM/DataList.php
+++ b/src/ORM/DataList.php
@@ -1185,6 +1185,25 @@ class DataList extends ViewableData implements SS_List, Filterable, Sortable, Li
     }
 
     /**
+     * Returns a DataList of the same dataClass which will always yield empty results
+     *
+     * @return DataList
+     */
+    public function getEmptyList()
+    {
+        // clear any existing query conditions to make the query as lightweight as possible
+        $list = DataList::create($this->dataClass);
+
+        // set an impossible condition which targets indexed DB column
+        $list = $list->byIDs([0]);
+
+        // limit further minimizes the query (can't do limit 0 though T_T)
+        $list = $list->limit(1);
+
+        return $list;
+    }
+
+    /**
      * Reverses a list of items.
      *
      * @return static
--- a/tests/php/ORM/DataListTest.php
+++ b/tests/php/ORM/DataListTest.php
@@ -1815,4 +1815,13 @@ class DataListTest extends SapphireTest
             $list->column("Title")
         );
     }
+
+    public function testGetEmptyList()
+    {
+        $players = Player::get();
+        $this->assertGreaterThan(0, (int) $players->count());
+
+        $players = $players->getEmptyList();
+        $this->assertEquals(0, (int) $players->count());
+    }
 }


kinglozzer:
I’m a bit confused what the “real world” use-case is for this...

> - empty results should be returned and the logic that shapes the `DataList` is quite complex

That sounds like the implementing code should return before building complex filtering logic, or just `return ArrayList::create();`

> - we are interested to keep function signature unchanged (mostly for PHP 7)

If this is about type-hinting, it should use `SS_List` instead - that would still be a backward compatible change if this refers to a module

mfendeksilverstripe:
@kinglozzer Thanks for the feedback. Here are some more notes to explain:

* we couldn't use `ArrayList` at all due to performance reasons
* also because some of the processing requires `DataList`, we're working with a big DB (`1GB` compressed snapshot)
* we really didn't want to compromise function signature to `SS_List` because of that reason
* the complex functionality that shapes the `DataList` is a set of classes (normal classes, parents classes, and extensions)
* this is because the code is quite `DRY`, the problem is that handling such edge cases is quite a pain
* setting the `DataList` to be empty is a really elegant solution to the problem and it's also fine from the performance perspective

This PR is basically trying to go with the code reuse theme and pushing some project code to modules.

tractorcow:
> don't want to deal with null values

Can you elaborate on this point?

> we couldn't use ArrayList at all due to performance reasons

What's the performance cost of an empty arraylist?

> we really didn't want to compromise function signature to SS_List because of that reason

Which functions are missing that you depend on? 

> setting the DataList to be empty is a really elegant solution to the problem and it's also fine from the performance perspective

I disagree that this is elegant; You are executing a sql query you know will be empty for the sake of a function signature. That's a redundant performance cost there.

You can add this method to DataList via an extension. This shouldn't be a part of core, given there's no demonstrated use case in core itself.

tractorcow:
If you just need a typed dataclass for ArrayList (e.g. gridfield with empty list) I could support adding a dataClass() typehint-setter to ArrayList. That's about the only use case where a DataList is needed over an ArrayList.

```php
    /**
     * Underlying type class for this list
     *
     * @var string
     */
    protected $dataClass = null;

    /**
     * Return the class of items in this list, by looking at the first item inside it.
     *
     * @return string
     */
    public function dataClass()
    {
        if ($this->dataClass) {
            return $this->dataClass;
        }
        if (count($this->items) > 0) {
            return get_class($this->items[0]);
        }
        return null;
    }

    /**
     * Hint this list to a specific type
     *
     * @param $class
     * @return $this
     */
    public function setDataClass($class)
    {
        $this->dataClass = $class;
        return $this;
    }
```

tractorcow:
I'm guessing you're running into this issue, correct?

```
    /**
     * Returns a data class that is a DataObject type that this GridField should look like.
     *
     * @return string
     *
     * @throws LogicException
     */
    public function getModelClass()
    {
        if ($this->modelClassName) {
            return $this->modelClassName;
        }

        /** @var DataList|ArrayList $list */
        $list = $this->list;
        if ($list && $list->hasMethod('dataClass')) {
            $class = $list->dataClass();

            if ($class) {
                return $class;
            }
        }

        throw new LogicException(
            'GridField doesn\'t have a modelClassName, so it doesn\'t know the columns of this grid.'
        );
    }
```

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/pull/8866

mfendeksilverstripe:
@tractorcow Thanks for the feedback, will have a look at this during next Hackday (or my PDP time if time allows).

mfendeksilverstripe:
The situation is as follows:

The `DataList` undergoes multiple transformations before outputting results during runtime. Something like this:

`DataList` -> `Class1` -> `Class2` -> `Class3` -> `Extension1` -> `Extension2` -> `Result`

Note that there are many other classes that are not in this specific chain.

Every class potentially alters the data list by applying stuff like filtering, sorting, grouping, subtraction and also it can decide to return an empty result.

We could alter the result signature to allow nulls like this: `: ?DataList`, but that would mean that we would have to copy-paste an early exit case to every function that alters the data list something like this:

```
if ($list === null) {
    return $list;
}
```

Ideally, we wanted to avoid that. Returning a `DataList` that would yield empty result set would be a much-preferred option. It's quite common to return an empty `ArrayList` in such cases, but we don't want to allow `ArrayList` in this transformation chain as our DB operates on large data sets. We initially used `ArrayList` which resulted in a terrible performance, we want to enforce `DataList` in our code base.

I agree that my current implementation is not ideal as it produces a useless query. This is still a preferred option on a large scale project compared to a potential performance problem with ArrayLists.

I did a timeboxed investigation into allowing the `DataList` to have the `dataClass` to be initially `null`. This would allow an early exit like this:

```
if (some_condition) {
    return DataList::create();
}
```

The problem is that the current code base relies on the dataClass to be present at all times and even uses nested constructors. This change would require significant effort to be implemented which is beyond scope of this PR.

I guess we'll use `DataList::byId(0)` for now...

PR #8853, opened 2019-03-12
NEW Avoid repeat readonly transformations in FieldList and Form

chillu:
This is an important performance optimisation to avoid repeatedly transforming those objects.
Since field sets can be deeply nested, and return clones of all fields on readonly transformation,
that's a significant processing overhead. It can also confuse fields which assume they're already transformed.

In particular, both LeftAndMain->getEditForm() and CMSMain->getEditForm() call $fields->makeReadonly(),
with slightly different criteria:
- LeftAndMain checks for $record->canEdit()
- CMSMain checks for !$record->canEdit() || $deletedFromStage
- If you have workflow installed, it transforms to readonly a *third time* in AdvancedWorkflowExtension->updateEditForm() based on $record->canEditWorkflow()

Admittedly this is increasing the surface of an already inconsistent API:
- Form->makeReadonly(), FieldList->makeReadonly(), but FormField->performReadonlyTransformation()
- setReadonly() is an internal helper, but needs to be public since it's called on cloned instances
- FormField->isDisabled(), but no higher level functions on Form or FieldList to transform all fields to disabled
- A readonly form can still process submissions (I think, haven't tested if it checks for FormAction->isReadonly())

I still think it's worthwhile due to the large performance impact of doing the same work repeatedly.
It relies on the code invoking readonly transformations to check, rather than making the readonly state setting
part of the transformation itself. I could call setReadonly() in ReadonlyTransformation, but that would only apply to FormField.
Neither Form nor FieldList use ReadonlyTransformation in their makeReadonly() implementations.

### DIFF
--- a/src/Forms/FieldList.php
+++ b/src/Forms/FieldList.php
@@ -35,6 +35,14 @@ class FieldList extends ArrayList
      */
     protected $containerField;
 
+    /**
+     * Determines if the entire field list was transformed to readonly state.
+     * Even if this is false, the list might still contain readonly fields.
+     *
+     * @var bool
+     */
+    protected $readonly = false;
+
     public function __construct($items = array())
     {
         if (!is_array($items) || func_num_args() > 1) {
@@ -808,13 +816,43 @@ class FieldList extends ArrayList
     }
 
     /**
-     * Transforms this FieldList instance to readonly.
+     * @return bool
+     */
+    public function isReadonly()
+    {
+        return $this->readonly;
+    }
+
+    /**
+     * Sets a read-only flag.
+     * Use makeReadonly() to transform this instance.
+     * Setting this to false has no effect.
+     *
+     * @param bool $readonly
+     *
+     * @return $this
+     */
+    public function setReadonly($readonly)
+    {
+        $this->readonly = $readonly;
+        return $this;
+    }
+
+    /**
+     * Transforms this FieldList instance to readonly, and returns a new instance.
+     * Returns the original instance if it's already marked as readonly.
      *
      * @return FieldList
      */
     public function makeReadonly()
     {
-        return $this->transform(new ReadonlyTransformation());
+        if ($this->isReadonly()) {
+            return $this;
+        }
+
+        $clone = $this->transform(new ReadonlyTransformation());
+        $clone->setReadonly(true);
+        return $clone;
     }
 
     /**
--- a/src/Forms/Form.php
+++ b/src/Forms/Form.php
@@ -227,6 +227,15 @@ class Form extends ViewableData implements HasRequestHandler
     protected $validationExemptActions = array();
 
     /**
+     * Determines if the entire formwas transformed to readonly state.
+     * Even if this is false, fields or actions contained in this form
+     * might still contain readonly fields.
+     *
+     * @var bool
+     */
+    protected $readonly = false;
+
+    /**
      * @config
      * @var array
      */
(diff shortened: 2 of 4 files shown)


kinglozzer:
> Admittedly this is increasing the surface of an already inconsistent API

I think this is fine, it’s a pretty modest increase. My only query really is around the functionality of `isReadonly()` - if I manually transform every form field to readonly, then call `FieldList::isReadonly()` it would return false when I might expect it to return true. However given the performance implications of recursively inspecting every field, and how rare it is to _manually_ transform every single field to readonly, I think storing it as a flag instead is acceptable

chillu:
>  if I manually transform every form field to readonly, then call FieldList::isReadonly() it would return false when I might expect it to return true.

Yeah, that's an existing inconsistency further down already - `CompositeField->isReadonly()` already has that problem (same with `TabSet`). It just increases the surface of that inconsistency.

tractorcow:
I'm not 100% sure the isReadonly() implementation on Form / FieldList is complete. As a converse example to @kinglozzer's one, you can add a non-readonly field after the fact, and the list still falsely reports it as being readonly.

Should adding a field to a readonly fieldlist auto-transform that field to readonly? That would maintain consistency.

I'm a little concerned about (existing) inconsistencies with transform() method. Sometimes it returns a cloned modified version, sometimes it modifies the base record.

@chillu do you think the isReadonly() need to be public on fieldlist / form? Given we aren't maintaining internal integrity, maybe we just need to record a flag to prevent multiple transformations occurring, instead of making public declarations?

chillu:
When merging this, please also re-merge https://github.com/symbiote/silverstripe-advancedworkflow/pull/388 - @robbieaverill (rightfully) reverted this in the workflow module for now, due to the dependency here

chillu:
>  do you think the isReadonly() need to be public on fieldlist / form? Given we aren't maintaining internal integrity, maybe we just need to record a flag to prevent multiple transformations occurring, instead of making public declarations?

@tractorcow Thanks for pitching in! :)

Option 1: Add `Form->isReadonly()` and `Form->setReadonly()`. Allows consumers of this API fine-grained control, for example re-transforming a form to readonly after extensions might have added non-readonly fields.

Option 2: Keep `Form::$isReadonly` as internal state. After `Form->makeReadonly()` is called, it can't be called again. If non-readonly fields are added to a "readonly" form, they won't be converted later on. Relies on consumers to replicate conditions which triggered the readonly transformation in the first place (often not possible?)

Option 3: Auto-convert newly added fields internally. This might not always be desireable, for example adding a "version selector" dropdown to an otherwise readonly form representing this version.

Example:

```php
<?php
class CoreController
{
	public function Form()
	{
		$form = new Form();
		// ...

		if (!$this->canEdit()) {
			$form = $form->makeReadonly();
		}

		return $form;
	}
}

class MyController extends CoreController
{
	public function Form()
	{
		$form = parent::Form();
		$newField = new TextField('Editable');

		// Option 1a: Transform field conditionally on form state
		if ($form->isReadonly()) {
			$newField = $newField->performReadonlyTransformation();
		}
		$form->Fields()->push($newField);

		// Option 1b: Transform field conditionally on same permission checks
		if (!$this->canEdit) {
			$newField = $newField->performReadonlyTransformation();
		}
		$form->Fields()->push($newField);

		// Option 2a: Transform form based on form state
		$form->Fields()->push($newField);
		if ($form->isReadonly()) {
			$form = $form->makeReadonly();
		}

		// Option 2b: Transform form based on same permission checks
		$form->Fields()->push($newField);
		if (!$this->canEdit()) {
			$form = $form->makeReadonly();
		}

		// Option 3: Form auto-converts field
		$form->Fields()->push($newField);

		return $form;
	}
}
```

As you can see, `Option 2` isn't really feasible here, particularly in controllers extending `LeftAndMain` and `CMSMain` with non-trivial permission checks triggering readonly conversion. So I think either we do `Option 1` (my current pull request), and live with broadening the existing API inconsistencies on `CompositeField->isReadonly()`. Or we keep the logic as-is, and live with the performance impact of transforming each field three times (in the case of advancedworkflow). I'm not a fan of `Option 3` because it creates different ambiguities.

GuySartorelli:
Would an additional option not be to add functionality to the `add*` methods check check if the field(s) being added are read-only, and if not, set read-only to false? Or even just assume they're not read-only. So whenever new fields are added, we turn the internal flag to false which allows calling `makeReadOnly()` again.

chillu:
I'm not going to be able to finish this, sorry

PR #8854, opened 2019-03-12
NEW: Add PHP 7.4’s daily snapshot to the travis suite.

sminnee:
This will help avoid any inadvertent 7.4 failures; IMO the sooner we
add new releases to the test mix the better.

If this ends up creating intermittent failures outside of our control
I would recommend rolling back entirely rather than adding to
allowed_failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,6 +52,15 @@ matrix:
         - DB=MYSQL
         - PHPUNIT_TEST=framework
 
+    - php: 7.4snapshot
+      env:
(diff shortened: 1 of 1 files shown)


sminnee:
This is blocked by GD's absence due to: https://github.com/travis-ci/php-src-builder/pull/29#issuecomment-472242280

sminnee:
OK this should be unblocked now and assuming it's green, is ready from my perspective for merge.

PR #8856, opened 2019-03-13
applyRelation versioning fix, ORM enhancements

mfendeksilverstripe:
ORM functionality fixes / enhancements:

* `DataQuery::applyRelation` now supports versioned tables
* improved documentation on how to use `DataList::applyRelation`
* `column` function now accepts table prefix format of DB fields, which allows querying data of related tables (via join)
* column function now throws `InvalidArgumentException` if an invalid field is provided instead of failing on the query later

Example use:

* pages `<- many many ->` taxonomy terms
* fetch all pages that have a specific taxonomy term
* fetch IDs of all taxonomy terms which are attached to specific pages

### DIFF
--- a/src/ORM/DataList.php
+++ b/src/ORM/DataList.php
@@ -501,6 +501,13 @@ class DataList extends ViewableData implements SS_List, Filterable, Sortable, Li
      * Unlike getRelationName, this is immutable and will fallback to the quoted field
      * name if not a relation.
      *
+     * Example use (simple WHERE condition on data sitting in a related table):
+     *
+     * <code>
+     *  $columnName = null;
+     *  $list = Page::get()->applyRelation('TaxonomyTerms.ID', $columnName)->where([$columnName => <some_value>]);
+     * </code>
+     *
      * @param string $field Name of field or relation to apply
      * @param string &$columnName Quoted column name
      * @param bool $linearOnly Set to true to restrict to linear relations only. Set this
--- a/src/ORM/DataQuery.php
+++ b/src/ORM/DataQuery.php
@@ -2,13 +2,15 @@
 
 namespace SilverStripe\ORM;
 
+use InvalidArgumentException;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Convert;
+use SilverStripe\Core\Extensible;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\ORM\Connect\Query;
 use SilverStripe\ORM\Queries\SQLConditionGroup;
 use SilverStripe\ORM\Queries\SQLSelect;
-use InvalidArgumentException;
+use SilverStripe\Versioned\Versioned;
 
 /**
  * An object representing a query of data from the DataObject's supporting database.
@@ -915,6 +917,19 @@ class DataQuery
             return;
         }
 
+        if (class_exists(Versioned::class)) {
+            /** @var Versioned $versioned */
+            $versioned = Injector::inst()->get(Versioned::class);
+            $stage = $versioned->get_stage();
+        } else {
+            $versioned = null;
+            $stage = null;
+        }
+
+        $foreignTableStaged = ($versioned && Extensible::has_extension($foreignClass, Versioned::class))
+            ? $versioned->stageTable($foreignTable, $stage)
+            : $foreignTable;
+
         // Join table with associated has_one
         /** @var DataObject $model */
         $foreignKey = $schema->getRemoteJoinField($localClass, $localField, $type, $polymorphic);
@@ -930,7 +945,7 @@ class DataQuery
             $joinExpression = "{$foreignKeyIDColumn} = {$localIDColumn}";
         }
         $this->query->addLeftJoin(
-            $foreignTable,
+            $foreignTableStaged,
             $joinExpression,
             $foreignTableAliased
         );
@@ -943,8 +958,12 @@ class DataQuery
             $ancestorTable = $schema->tableName($ancestor);
             if ($ancestorTable !== $foreignTable) {
                 $ancestorTableAliased = $foreignPrefix . $ancestorTable;
+                $ancestorTableStaged = ($versioned && Extensible::has_extension($ancestor, Versioned::class))
+                    ? $versioned->stageTable($ancestorTable, $stage)
+                    : $ancestorTable;
+
                 $this->query->addLeftJoin(
-                    $ancestorTable,
+                    $ancestorTableStaged,
                     "\"{$foreignTableAliased}\".\"ID\" = \"{$ancestorTableAliased}\".\"ID\"",
                     $ancestorTableAliased
                 );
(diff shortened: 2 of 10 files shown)


robbieaverill:
Epic contribution! Will take some time for the @silverstripe/core-team to review. In the meantime, if you wouldn't mind rebasing and switching onto the `4` branch (as an enhancement) that'd be great.

mfendeksilverstripe:
@robbieaverill Rebase on `4` complete :)

src/ORM/DataList.php
```diff
@@ -501,6 +501,13 @@ public function filterByCallback($callback)
      * Unlike getRelationName, this is immutable and will fallback to the quoted field
      * name if not a relation.
      *
+     * Example use (simple WHERE condition on data sitting in a related table):
+     *
+     * <code>
+     *  $columnName = null;
+     *  $list = Page::get()->applyRelation('TaxonomyTerms.ID', $columnName)->where([$columnName => <some_value>]);
```
tractorcow: Sorry to jump on the first line I saw, but is this a real usage example? Why is `$columnName` null?

src/ORM/DataList.php
```diff
@@ -501,6 +501,13 @@ public function filterByCallback($callback)
...
+     * </code>
+     *
```
tractorcow: Oh no, I wrote this method didn't I?

I should have done a better job of documenting this sorry.

tractorcow:
Question, what's the difference between.

```php
Category::get()->relationWhere('Products.ID', ['"Products.ID"' => $product->ID])
```

and 

```php
Category::get()->filter('Products.ID', $product->ID);
```

tractorcow:
Some user-docs changes in `*.md` files might help me understand. :)

maxime-rainville:
@chillu I'm moving this to the next column because it's a big chunk of work that got done internally, so I think it should be prioritise.

mfendeksilverstripe:
@tractorcow There is no difference in your example, however in general `relationWhere` allows you to build the condition clause yourself so it allows greater flexibility. `relationWhere` is really just a shorthand for `where` function in case the relation data is queried. The idea is to remove the need for juggling with the `$columnName` in the `applyRelation`.

Overall, the `relationWhere` function is just a nice to have, but I think the rest of the diff is actually more important.

tractorcow:
> in case the relation is versioned (uses many many through) you will always get the draft content as the filter internally uses the applyRelation which currently doesn't support versioning.


Sounds like that should be fixed instead right?

mfendeksilverstripe:
@tractorcow Fortunately, the versioning issue is fixed as a part of this PR as well. `relationWhere` is not supposed to be used as a workaround for the versioning issues. It's just a shorthand method which provides an easier way to make relation based `where` conditions.

tractorcow:
I suggest this PR focus on fixing the bug rather than adding extraneous API.

mfendeksilverstripe:
@tractorcow `relationWhere` function removed.

src/ORM/DataQuery.php
```diff
@@ -912,6 +914,14 @@ protected function joinHasManyRelation(
             return;
         }
 
+        /** @var Versioned $versioned */
+        $versioned = Injector::inst()->get(Versioned::class);
```
ScopeyNZ: `Versioned` is technically a dev requirement so we can't assume that it's installed. Preferably this would actually live in Versioned code too. Perhaps we can just put

```
$this->extend('updateForeignTable', $foreignTable, $foreignClass);
```

And then have this code in `Versioned`?

mfendeksilverstripe: I think that's fine, will try to pick it up on the next hack day, but I don't know if I have time. Do you think this requires someone else to pick it up and fix the issue earlier?

ScopeyNZ: I will need to get more familiar with the bug to give you a good answer. I'm happy to help out when you next pick it up though :)

chillu:
@tractorcow Could you take another look and merge if you're happy?

mfendeksilverstripe:
Squashed and rebased.

src/ORM/DataQuery.php
```diff
@@ -915,6 +917,19 @@ protected function joinHasManyRelation(
             return;
         }
 
+        if (class_exists(Versioned::class)) {
```
mfendeksilverstripe: @ScopeyNZ I decided to go for a different approach as extension points seems like overkill here. This code now checks if `Versioned` is available.

src/ORM/DataQuery.php
```diff
@@ -1038,13 +1069,29 @@ protected function joinManyManyRelationship(
     ) {
         $schema = DataObject::getSchema();
 
+        /** @var Versioned $versioned */
+        $versioned = Injector::inst()->get(Versioned::class);
```
ScopeyNZ: Sorry @mfendeksilverstripe - this bit probably needs the same guards you did for the other bit of code.

mfendeksilverstripe: Good spotting, thanks :)

mfendeksilverstripe:
@ScopeyNZ `Versioned` guard added to all other instances where it's used.

mfendeksilverstripe:
@chillu Any updates on this PR?

maxime-rainville:
@brynwhyman Can we prioritise this one for peer review? It's a lot of work and it would be very sad to see it go stale.

mfendeksilverstripe:
@maxime-rainville What's the status on this one? If we can't get this whole thing through maybe it's worth separating out the Versioned related fix as it's the most important of the change set. Until this is fixed the `ManyManyThroughList` will serve draft content from the through object table. It also affects instances where `filter` and `applyRelation` is used with through objects.

src/ORM/DataQuery.php
```diff
@@ -1142,16 +1203,51 @@ public function addSelectFromTable($table, $fields)
...
      * Query the given field column from the database and return as an array.
+     * querying DB columns of related tables is supported but you need to make sure that the related table
+     * is already available in join
+     *
+     * @see DataList::applyRelation()
+     *
+     * example use:
+     *
+     * <code>
+     *  column("<table_name>"."<column_name>")
```
robbieaverill: I think that this code example should prioritise one like `->column('Title')` since it's probably a more common use case

src/ORM/DataQuery.php
```diff
@@ -1142,16 +1203,51 @@ public function addSelectFromTable($table, $fields)
...
+     * </code>
      *
      * @param string $field See {@link expressionForField()}.
      * @return array List of column values for the specified column
+     * @throws InvalidArgumentException
```
robbieaverill: This could be considered a breaking change. Currently this method does already throw an InvalidArgumentException via `DataObjectSchema::sqlColumnForField()`, but perhaps not in the same context as this - although the logic does look fairly similar. Would you mind confirming?

src/ORM/DataQuery.php
```diff
@@ -1142,16 +1203,51 @@ public function addSelectFromTable($table, $fields)
...
+            $tablePrefix = str_replace('"', '', $fieldData[0]);
+
+            // check if related table is available
+            if (!$query->isJoinedTo($tablePrefix)) {
+                throw $exception;
+            }
+
+            $fieldExpression = $field;
+            $field = null;
+        }
```
robbieaverill: I think I'd prefer to see this in its own method, maybe something like `validateFieldExpression()`

tests/php/ORM/DataListTest.php
```diff
@@ -4,34 +4,58 @@
...
+            Versioned::class,
+        ],
+        HouseVisit::class => [
+            Versioned::class,
+        ],
+        Visitor::class => [
+            Versioned::class,
+        ],
+        Roof::class => [
+            Versioned::class,
```
robbieaverill: Maybe these tests belong in the versioned module? If we do keep them here then these tests will fail without versioned installed (logically, since the class not existing as a string literal is translucent but acceptable) - we may want to mark these tests as skipped if the module is not installed. It could even become its own test class with this in the setup method.

src/ORM/DataQuery.php
```diff
@@ -915,6 +917,19 @@ protected function joinHasManyRelation(
...
         }
 
+        if (class_exists(Versioned::class)) {
+            /** @var Versioned $versioned */
+            $versioned = Injector::inst()->get(Versioned::class);
+            $stage = $versioned->get_stage();
+        } else {
+            $versioned = null;
+            $stage = null;
+        }
```
robbieaverill: I'd really prefer to see this in an extension in the versioned module. If we do keep it here, I think there's some code duplication in this file that could be mitigated with a private method to apply these rules.

mfendeksilverstripe:
Split into

https://github.com/silverstripe/silverstripe-framework/pull/9386
https://github.com/silverstripe/silverstripe-versioned/pull/253

PR #8860, opened 2019-03-15
Added an option to fetch one row at a time.

mfendeksilverstripe:
* this is a Proof of Concept PR, feedback is welcome
* added a fetch method option to `Query` to support `fetch one row at a time` feature
* this is great for maintenance tasks that are expected to process a lot of data and do not need to more than one record at a time

### DIFF
--- a/src/ORM/Connect/PDOQuery.php
+++ b/src/ORM/Connect/PDOQuery.php
@@ -2,14 +2,43 @@
 
 namespace SilverStripe\ORM\Connect;
 
-use PDOStatement;
 use PDO;
+use PDOStatement;
 
 /**
  * A result-set from a PDO database.
  */
+
+/**
+ * Class PDOQuery
+ * A result-set from a PDO database.
+ *
+ * Fetch method will be set to FETCH_ALL by default
+ * this means ALL results are fetched immediately which makes the handling of data convenient
+ * the problem of this approach is that all results are stored in memory which can be a problem with large data sets
+ * FETCH_ROW method should be used for cases where large data sets are used and only one row needs to be in memory
+ * at one time
+ * note that this option disables some features like counting the rows and seeking specific rows
+ *
+ * <code>
+ * $results = PDOQuery::withFetchMethodFlag(function (): Query {
+ *     PDOQuery::setFetchMethodFlag(PDOQuery::FETCH_METHOD_ROW);
+ *
+ *     return DB::query('SELECT `ID`, `Title` FROM `SiteTree` ORDER BY `ID` ASC LIMIT 10');
+ * });
+ * while ($row = $results->next()) {
+ *    // one row at a time
+ *   print_r($row);
+ * }
+ * </code>
+ *
+ * @package SilverStripe\ORM\Connect
+ */
 class PDOQuery extends Query
 {
+    const FETCH_METHOD_ALL = 'FETCH_ALL';
+    const FETCH_METHOD_ROW = 'FETCH_ROW';
+
     /**
      * The internal MySQL handle that points to the result set.
      * @var PDOStatement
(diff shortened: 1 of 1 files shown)


src/ORM/Connect/PDOQuery.php
```diff
@@ -18,13 +47,70 @@ class PDOQuery extends Query
 
     protected $results = null;
 
+    /**
+     * @var string
+     */
+    protected $fetchMedthod = self::FETCH_METHOD_ALL;
```
ScopeyNZ: Typo 😉 

```suggestion
    protected $fetchMethod = self::FETCH_METHOD_ALL;
```

Obviously don't accept the suggestion without updating all the other bits of code.

src/ORM/Connect/PDOQuery.php
```diff
@@ -18,13 +47,70 @@ class PDOQuery extends Query
...
+    /**
+     * Invoke a callback which may modify fetch method flag, but ensures this flag is restored
+     * after completion, without modifying global state.
+     *
+     * The desired fetch method flag should be set by the callback directly
+     *
+     * @param callable $callback
+     * @return mixed Result of $callback
+     */
+    public static function withFetchMethodFlag($callback)
```
ScopeyNZ: Wouldn't it make sense to also ask for the intended fetch method as a parameter so this method can handle both changing to a new method and back to the existing method?

src/ORM/Connect/PDOQuery.php
```diff
@@ -34,17 +120,39 @@ public function __construct(PDOStatement $statement)
 
     public function seek($row)
     {
+        // seek is not available for fetch row
+        if ($this->fetchMedthod === static::FETCH_METHOD_ROW) {
+            return [];
```
ScopeyNZ: I believe that throwing an Exception here would actually be more convenient for a developer. Debugging why your row contains no values when you don't realise that seek shouldn't be used with this fetch method would be a footgun. This goes for the other "unsupported" API.

mfendeksilverstripe:
@ScopeyNZ Thanks for feedback. Looking at this code now, I'm thinking that a new class would be a better approach. Will try to revisit this during my PDP time or the next hackday.

kinglozzer:
@mfendeksilverstripe I’m not sure if you’re looking to add this to other drivers too, but just a friendly heads up that we have an RFC open about dropping PDO support entirely https://github.com/silverstripe/silverstripe-framework/issues/8598

kinglozzer:
Actually, it looks like @sminnee already switched non-PDO drivers to use generators for SS5, which should achieve the same goal as this feature: https://github.com/silverstripe/silverstripe-framework/pull/6518. That doesn’t exclude this as a possible SS4 feature, just noting it should be possible in SS5 without any extra flags :D

mfendeksilverstripe:
@kinglozzer Thanks for the update, these are great news. I guess having this option just for SS 4 would still be nice.

sminnee:
Do you know if the problems that this PR fixes occur with the non-PDO driver, @mfendeksilverstripe? Doesn’t block merge, just curious. It would be another validation point for my proposed depreciation of PDO in 4.5

mfendeksilverstripe:
@sminnee Not sure will need to investigate further.

mfendeksilverstripe:
@sminnee @ScopeyNZ I had a look at the https://github.com/silverstripe/silverstripe-framework/pull/6518

I don't that PR fixes the issue with the `PDO` as this driver will still fetch all data at once via this call `$this->statement->fetchAll(PDO::FETCH_NUM)`.

I had a look at other driver - `MySQLi` and that one looks fine. It seems to fetch one row at a time.

I think long term solution is to switch to `MySQLi` so I'm going to close this PR as this was just a `PoC` anyway.

PR #8866, opened 2019-03-19
API Allow empty arraylists to be typed

tractorcow:
Replaces https://github.com/silverstripe/silverstripe-framework/pull/8852

Allows an empty typed list to be created. Useful for GridFields or other areas where dataclass is essential, even for empty sets.

I've run the phpcs linter over framework and it caught a few issues on the way.

### DIFF
--- a/src/ORM/ArrayList.php
+++ b/src/ORM/ArrayList.php
@@ -2,13 +2,13 @@
 
 namespace SilverStripe\ORM;
 
-use SilverStripe\Dev\Debug;
-use SilverStripe\View\ArrayData;
-use SilverStripe\View\ViewableData;
 use ArrayIterator;
 use InvalidArgumentException;
 use LogicException;
+use SilverStripe\Dev\Debug;
 use SilverStripe\Dev\Deprecation;
+use SilverStripe\View\ArrayData;
+use SilverStripe\View\ViewableData;
 
 /**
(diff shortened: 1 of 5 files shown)


unclecheese:
I won't miss the days of having to hack this in.

robbieaverill:
Re: the linting, have you used PSR-12? We may want to update the phpcs ruleset

tractorcow:
No I used the rules in the framework root; `phpcs.xml.dist` https://github.com/silverstripe/silverstripe-framework/blob/4/phpcs.xml.dist

robbieaverill:
Strange - I wonder why the builds didn't fail previously

tractorcow:
maybe there's a discrepency between phpcs and phpcbf?

tractorcow:
```
<rule phpcbf-only="true" ref="Squiz.Strings.ConcatenationSpacing">
		<properties>
			<property name="spacing" value="1" />
			<property name="ignoreNewlines" value="true"/>
		</properties>
	</rule>
```

robbieaverill:
Right, makes sense. PSR-12 includes these rules anyway, we can remove that when we switch in the 4.x branch (if we do that)

ScopeyNZ:
I've done this in a custom subclass of `ArrayList` so it makes a lot of sense to me!

